PR #988, opened 2021-11-19
update Scheduled conditon when failed scheduling

mrlihanbo:
Signed-off-by: lihanbo <lihanbo2@huawei.com>

**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**:
1. update `Scheduled` conditon when failed scheduling
2. remove `FirstSchedule`, reconcile schedule once applied policy is different from the latest propogation policy.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:
```
  conditions:
  - lastTransitionTime: "2021-11-19T07:34:38Z"
    message: 'failed to assignReplicas: failed to scaleUp: clusters resources are
      not enough to schedule, max 6 replicas are support'
    reason: BindingFailedScheduling
    status: "False"
    type: Scheduled
```

**Does this PR introduce a user-facing change?**:
"NONE"

PR #1051, opened 2021-11-30
karmada-scheduler: merge scalescheduling with normal scheduling

Garrybest:
**What type of PR is this?**
/kind feature

Now scale scheduling has a particular process without share the same logic with normal scaleduling. Here shows some problems:

1. The code is too complicated.
2. If a new cluster has added, the scale scheduling could not make the new one as a candidate because the process does not contain a replica estimation step. The result of candidate clusters are based on the previous scheduling result. See [here](https://github.com/karmada-io/karmada/blob/763a50ef3c3edda7f561a3ab88b3a079a5b35f61/pkg/scheduler/core/division_algorithm.go#L143-L149)

Now the two steps have been merged. Actually, the first scheduling is a special kind of `scaling up`. There is no need to separate the two kinds.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/core/division_algorithm.go
+++ b/pkg/scheduler/core/division_algorithm.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"sort"
 
+	"k8s.io/apimachinery/pkg/util/sets"
+
 	clusterv1alpha1 "github.com/karmada-io/karmada/pkg/apis/cluster/v1alpha1"
 	policyv1alpha1 "github.com/karmada-io/karmada/pkg/apis/policy/v1alpha1"
 	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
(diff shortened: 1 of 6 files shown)


Garrybest:
The PR has been divided into 3 commits.

1. Merge scale scheduling with normal scheduling. This is the most important change.
2. Delete scale scheduling. I just delete the scale scheduling logic from scheduler, it is almost full of deletion codes.
3. Unit test change for function parameters change. This is not that important.

pkg/scheduler/core/generic_scheduler.go
```diff
@@ -207,23 +210,29 @@ func (g *genericScheduler) assignReplicas(clusters []*clusterv1alpha1.Cluster, r
...
 				if replicaSchedulingStrategy.WeightPreference == nil {
 					replicaSchedulingStrategy.WeightPreference = getDefaultWeightPreference(clusters)
 				}
+				// 2.1.1 Dynamic Weighted Scheduling (by resource)
 				if len(replicaSchedulingStrategy.WeightPreference.DynamicWeight) != 0 {
 					return divideReplicasByDynamicWeight(clusters, replicaSchedulingStrategy.WeightPreference.DynamicWeight, object)
 				}
+				// 2.1.2 Static Weighted Scheduling
 				return divideReplicasByStaticWeight(clusters, replicaSchedulingStrategy.WeightPreference.StaticWeightList, object.Replicas)
+			// 2.1 Aggregated scheduling (by resource)
```
mrlihanbo, line 235: should be 2.2

RainbowMango:
> Now the two steps have been merged. Actually, the first scheduling is a special kind of scaling up. There is no need to separate the two kinds.

Same idea with https://github.com/karmada-io/karmada/pull/988#discussion_r760015979.

RainbowMango:
cc @dddddai

qianjun1993:
As we merged the logic of first schedule and scale schedule, Maybe we do not need retain the cluster with 0 replicas in scheudlue result. https://github.com/karmada-io/karmada/blob/bbf113adeb0f32ed1831363f61f007716240557f/pkg/scheduler/core/division_algorithm.go#L146

pkg/scheduler/core/division_algorithm.go
```diff
@@ -140,13 +168,6 @@ func divideReplicasByAvailableReplica(clusterAvailableReplicas []workv1alpha2.Ta
 	}
 	divideRemainingReplicas(int(replicas-allocatedReplicas), desireReplicaInfos, clusterNames)
 
-	// For scaling up
-	for _, cluster := range unusedClusters {
-		if _, exist := desireReplicaInfos[cluster]; !exist {
-			desireReplicaInfos[cluster] = 0
```
Garrybest, line 146: That's right, I made a deletion here. @qianjun1993

qianjun1993: my mistake

pkg/scheduler/core/division_algorithm.go
```diff
@@ -179,45 +200,45 @@ func divideRemainingReplicas(remainingReplicas int, desiredReplicaInfos map[stri
...
-	} else {
-		return spec.Clusters, nil
-	}
+func scaleDownScheduleByReplicaDivisionPreference(
+	spec *workv1alpha2.ResourceBindingSpec,
+	preference policyv1alpha1.ReplicaDivisionPreference,
+) ([]workv1alpha2.TargetCluster, error) {
+	// The previous scheduling result will be the weight reference of scaling down.
+	// In other words, we scale down the replicas proportionally by their scheduled replicas.
+	return divideReplicasByPreference(spec.Clusters, spec.Replicas, preference, emptyClusterNames)
```
RainbowMango, line 209:
```suggestion
	return divideReplicasByPreference(spec.Clusters, spec.Replicas, preference, sets.NewString())
```
Just a suggestion. And just a feeling that the `emptyClusterNames` should be removed as looks weird.

RainbowMango (commented):
Review still on-going.

We are lacking proper abstractions, the code becomes difficult to understand, now.  :(

Garrybest:
> We are lacking proper abstractions, the code becomes difficult to understand, now. :(

Emmm, that's right, too complicated😵‍💫. But this PR could help make it better because we merge the two cases.

RainbowMango:
Yes, this PR is a relief. 
By the way, can you solve the conflicts, we just merged #988, and I wish this PR is the next.

PR #1053, opened 2021-12-01
abort deleting workload when the member cluster is unready

### LINKED ISSUE #438: Cluster not ready, work delete should retry

Cluster not ready, `work` delete should retry.

What need to do:

1) add an field forceCleanup to struct clusterspce (pkg/apis/cluster/v1alpha1/types.go)

2) add a flag --force to command karmadactl unjoin
If flag --force was added, we will modify clusterspce's forceCleanup to false, and then delete the cluster.

3) extend clustercontroller
If cluster was deleted and field forceCleanup was true, we cloud ignore err when cleanup fail.

**Why is this needed**:

The detail: https://github.com/karmada-io/karmada/pull/387

mrlihanbo:

**What type of PR is this?**
/kind cleanup

Abort deleting workload when the member cluster is unready, otherwise we can't unjoin the member cluster when the member cluster is unready

Fixes #438

"NONE"

### DIFF
--- a/pkg/controllers/execution/execution_controller.go
+++ b/pkg/controllers/execution/execution_controller.go
@@ -34,12 +34,13 @@ const (
 
 // Controller is to sync Work.
 type Controller struct {
-	client.Client   // used to operate Work resources.
-	EventRecorder   record.EventRecorder
-	RESTMapper      meta.RESTMapper
-	ObjectWatcher   objectwatcher.ObjectWatcher
-	PredicateFunc   predicate.Predicate
-	InformerManager informermanager.MultiClusterInformerManager
+	client.Client        // used to operate Work resources.
+	EventRecorder        record.EventRecorder
+	RESTMapper           meta.RESTMapper
+	ObjectWatcher        objectwatcher.ObjectWatcher
+	PredicateFunc        predicate.Predicate
+	InformerManager      informermanager.MultiClusterInformerManager
+	ClusterClientSetFunc func(clusterName string, client client.Client) (*util.DynamicClusterClient, error)
 }
 
 // Reconcile performs a full reconciliation for the object referred to by the Request.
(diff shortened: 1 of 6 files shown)


pkg/controllers/execution/execution_controller.go
```diff
@@ -86,6 +84,11 @@ func (c *Controller) Reconcile(ctx context.Context, req controllerruntime.Reques
 		return c.removeFinalizer(work)
 	}
 
+	if !isClusterReady {
+		klog.Errorf("Stop sync work(%s/%s) for cluster(%s) as cluster not ready.", work.Namespace, work.Name, cluster.Name)
+		return controllerruntime.Result{Requeue: true}, fmt.Errorf("cluster(%s) not ready", cluster.Name)
```
RainbowMango, line 89: Endless re-queue in case of the cluster not ready?

RainbowMango:
cc @pigletfly 

cc @XiShanYongYe-Chang 
since we need an E2E to cover: `not ready cluster could be unjoined`

RainbowMango:
@mrlihanbo Please create an issue to track the E2E case if it not included in this PR.

pkg/controllers/execution/execution_controller.go
mrlihanbo: > 

Need to work with failover scheduler. For execution controller, it will always re-queue in case of the cluster not ready. But when the cluster turns into not ready status, scheduler will reschedule and remove the cluster from schedule result.

mrlihanbo:
@RainbowMango @XiShanYongYe-Chang found another bug, we used to give up deleting workloads in member cluster if `Applied` condition is false. But it is not correct, the condition will set to false if `tryUpdateWorkload` failed. Under this scenario, `Applied` condition is false but we have actually dispatched the resource to member clusters. So, I try to get resource first, and check if we need to delete it. Any better ideas?

PR #1062, opened 2021-12-03
Correct the use of lock in singleClusterInformerManagerImpl.ForResource

cvvz:
**What type of PR is this?**
/kind bug

1. Correct the use of lock in `singleClusterInformerManagerImpl.ForResource`
2. Trim the unnecessary conditional judgement
3. Change the type of `singleClusterInformerManagerImpl.syncedInformers` to `mapset.Set` to avoid use the same lock as `singleClusterInformerManagerImpl.handlers` explicitly

The `lock` in `singleClusterInformerManagerImpl` is just for the sake of map's concurrency safety, but it can't promise the atomicity of `ForResource` method.  `AddEventHandler` will still get to execute repeatedly when multiple goroutines enter `ForResource` simultaneously.

```release-note
NONE
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/karmada-io/karmada
 go 1.16
 
 require (
+	github.com/deckarep/golang-set v1.7.1
 	github.com/distribution/distribution/v3 v3.0.0-20210507173845-9329f6a62b67
 	github.com/evanphx/json-patch/v5 v5.2.0
 	github.com/gogo/protobuf v1.3.2
--- a/go.sum
+++ b/go.sum
@@ -169,6 +169,8 @@ github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSs
 github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c=
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/daviddengcn/go-colortext v0.0.0-20160507010035-511bcaf42ccd/go.mod h1:dv4zxwHi5C/8AeI+4gX4dCWOIvNi7I6JCSX0HvlKPgE=
+github.com/deckarep/golang-set v1.7.1 h1:SCQV0S6gTtp6itiFrTqI+pfmJ4LN85S1YzhDf9rTHJQ=
+github.com/deckarep/golang-set v1.7.1/go.mod h1:93vsz/8Wt4joVM7c2AVqh+YRMiUSc14yDtF28KmMOgQ=
 github.com/denverdino/aliyungo v0.0.0-20190125010748-a747050bb1ba/go.mod h1:dV8lFg6daOBZbT6/BDGIz6Y3WFGn8juu6G+CQ6LHtl0=
 github.com/dgrijalva/jwt-go v0.0.0-20170104182250-a601269ab70c/go.mod h1:E3ru+11k8xSBh+hMPgOLZmtrrCbhqsmaPHjLKYnJCaQ=
 github.com/dgrijalva/jwt-go v3.2.0+incompatible/go.mod h1:E3ru+11k8xSBh+hMPgOLZmtrrCbhqsmaPHjLKYnJCaQ=
--- a/pkg/clusterdiscovery/clusterapi/clusterapi.go
+++ b/pkg/clusterdiscovery/clusterapi/clusterapi.go
@@ -67,10 +67,8 @@ func (d *ClusterDetector) Start(ctx context.Context) error {
 
 func (d *ClusterDetector) discoveryCluster() {
 	for _, gvr := range clusterGVRs {
-		if !d.InformerManager.IsHandlerExist(gvr, d.EventHandler) {
-			klog.Infof("Setup informer fo %s", gvr.String())
-			d.InformerManager.ForResource(gvr, d.EventHandler)
-		}
+		klog.Infof("Setup informer fo %s", gvr.String())
+		d.InformerManager.ForResource(gvr, d.EventHandler)
 	}
 
 	d.InformerManager.Start()
--- a/pkg/controllers/mcs/service_export_controller.go
+++ b/pkg/controllers/mcs/service_export_controller.go
@@ -170,7 +170,7 @@ func (c *ServiceExportController) registerInformersAndStart(cluster *clusterv1al
 
 	allSynced := true
 	for _, gvr := range gvrTargets {
-		if !singleClusterInformerManager.IsInformerSynced(gvr) || !singleClusterInformerManager.IsHandlerExist(gvr, c.getEventHandler(cluster.Name)) {
+		if !singleClusterInformerManager.IsInformerSynced(gvr) {
 			allSynced = false
 			singleClusterInformerManager.ForResource(gvr, c.getEventHandler(cluster.Name))
 		}
--- a/pkg/controllers/status/workstatus_controller.go
+++ b/pkg/controllers/status/workstatus_controller.go
@@ -375,7 +375,7 @@ func (c *WorkStatusController) registerInformersAndStart(cluster *clusterv1alpha
 
 	allSynced := true
 	for gvr := range gvrTargets {
-		if !singleClusterInformerManager.IsInformerSynced(gvr) || !singleClusterInformerManager.IsHandlerExist(gvr, c.getEventHandler()) {
+		if !singleClusterInformerManager.IsInformerSynced(gvr) {
 			allSynced = false
 			singleClusterInformerManager.ForResource(gvr, c.getEventHandler())
 		}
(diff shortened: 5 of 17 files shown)


algebra2k:
Thank you for this pull request, But I think that calling `ForResource` is generally when the program starts, which means it will determine in advance which resources are needed, so there may be no concurrency problems. But then, if the program has a `ForResource` competition problem in a concurrent environment, this pull request It makes sense, so can you help confirm it?

RainbowMango:
cc @Garrybest @XiShanYongYe-Chang

cvvz:
> 

Sure, I'll confirm it when I'm not occupied.

RainbowMango:
I think we should prevent this kind of issue even though no risk for now.
But after looking into the code, I found another lock in `dynamicSharedInformerFactory `, which might be possible to avoid this issue.

https://github.com/karmada-io/karmada/blob/c7aeba3481f1b26b5a898736b42bfcadc650a22a/vendor/k8s.io/client-go/dynamic/dynamicinformer/informer.go#L68-L70

cvvz:
hi，这个问题搞清楚了，但是由于问题比较的复杂，所以我只好用中文来描述了，抱歉。

首先这个地方锁的用法的确是有问题的，因为这里的锁是用来保证map读写的线程安全的，但是它并不能保证读map、“其他操作”、写map三个操作的原子性。这样当发生并发时，就会存在并发执行“其他操作”的可能性，就可能导致问题。

我尝试从以下几种场景来寻找并发执行同一个ForResource（**resource和handler都要相同**）的可能性：
1. 在manager启动执行start的时候，会并发的执行Runnable的start，有没有可能执行到同一个ForResource？看了一下应该是没有的。
2. 多个controller在并发reconcile的时候，有没有可能执行同一个ForResource的情况？看了一下应该也没有
3. 但是每个controller的worker个数有可能大于1。在这种情况下同一个controller的多个worker会并发执行同一个reconcile，也就存在并发执行执行同一个ForResource的可能。虽然现在默认都是1个worker，但是通过设置[MaxConcurrentReconciles](https://github.com/karmada-io/karmada/blob/master/vendor/sigs.k8s.io/controller-runtime/pkg/controller/controller.go#L38)是可以进行调整的。

另外，看了一下  @RainbowMango  指出的[dynamicSharedInformerFactory.ForResource](https://github.com/karmada-io/karmada/blob/c7aeba3481f1b26b5a898736b42bfcadc650a22a/vendor/k8s.io/client-go/dynamic/dynamicinformer/informer.go#L68-L70)中的锁的用法，这里锁的用法是正确的，而且这也是一个很好的正面例子：这里的锁锁住的就是读map、“其他操作”、写map这三个操作，保证了他们的原子性。但是正因为这里锁的用法是正确的，反而导致这个问题的出现概率变高了：

假设现在有两个goroutine恰好同时执行到了singleClusterInformerManagerImpl.ForResource中的s.informerFactory.ForResource，然后先执行的goroutine进入dynamicSharedInformerFactory.ForResource，[发现f.informers中不存在指定gvr的informer，就创建一个新的informer并存入f.informers](https://github.com/karmada-io/karmada/blob/c7aeba3481f1b26b5a898736b42bfcadc650a22a/vendor/k8s.io/client-go/dynamic/dynamicinformer/informer.go#L78-L79)；接着后执行的goroutine进入dynamicSharedInformerFactory.ForResource，[发现f.informers中存在informer了，就直接把现有的informer返回回去](https://github.com/karmada-io/karmada/blob/c7aeba3481f1b26b5a898736b42bfcadc650a22a/vendor/k8s.io/client-go/dynamic/dynamicinformer/informer.go#L73-L76)；这样，在singleClusterInformerManagerImpl.ForResource中，就必然会[对同一个informer执行两次AddEventHandler](https://github.com/karmada-io/karmada/blob/c7aeba3481f1b26b5a898736b42bfcadc650a22a/pkg/util/informermanager/single-cluster-manager.go#L88)。

试想，要是dynamicSharedInformerFactory.ForResource中的锁的用法不正确，也就是如果只对map的读写进行加锁，那么反而有可能就避免了这个问题 :)


另外，代码中还有一处锁的用法也有问题。就是MultiClusterInformerManager的[ForCluster](https://github.com/karmada-io/karmada/blob/c7aeba3481f1b26b5a898736b42bfcadc650a22a/pkg/util/informermanager/multi-cluster-manager.go#L87-L98)方法，也是一样的没有对读map、“其他操作”、写map同时加锁，导致[NewSingleClusterInformerManager](https://github.com/karmada-io/karmada/blob/c7aeba3481f1b26b5a898736b42bfcadc650a22a/pkg/util/informermanager/multi-cluster-manager.go#L95)可能会被并发执行到。而且这个在代码里是存在被并发执行的可能性的：WorkStatusController和ServiceExportController都是watch的work对象，所以他们的reconcile是可能会同时执行ForCluster，即为成员集群创建informerFactory的。虽然看起来应该不会造成什么实质性问题，因为只有存入了[multiClusterInformerManagerImpl.managers](https://github.com/karmada-io/karmada/blob/c7aeba3481f1b26b5a898736b42bfcadc650a22a/pkg/util/informermanager/multi-cluster-manager.go#L75)中的informerFactory才会被start，多创建的那个informerFactory应该会被gc掉。但是还是应该修复一下。

cc @algebra2k @RainbowMango

cvvz:
https://github.com/karmada-io/karmada/pull/1066 fix the use of lock in `MultiClusterInformerManager`

RainbowMango:
@cvvz 感谢精彩和深入的分析，昨晚回复完我也意识到`dynamicSharedInformerFactory`中的锁并不能解决你提到的问题。

RainbowMango:
I'll review the patch as soon as possible.

RainbowMango:
I see you introduced [golang-set](https://github.com/deckarep/golang-set) which provides a thread-safe set, and it looks great.
The problem is the `golang-set` looks inactivity(the latest commit happened a year ago).  We usually don't depend on inactivity packages.

So, can we use the current [read-write lock](https://github.com/karmada-io/karmada/blob/f7b2a91eed08c65c2d534b31068dd19dffdbca30/pkg/util/informermanager/single-cluster-manager.go#L79) instead?

cvvz:
I think [golang-set](https://github.com/deckarep/golang-set) is reliable since it is used by more than 4.7k projects, including some famous projects such as [docker](https://github.com/moby/moby/tree/master/vendor/github.com/deckarep/golang-set)  
 and [ethereum](https://github.com/ethereum/go-ethereum/blob/master/go.mod#L20) as well. The [lock](https://github.com/karmada-io/karmada/blob/f7b2a91eed08c65c2d534b31068dd19dffdbca30/pkg/util/informermanager/single-cluster-manager.go#L79) should not be used both by [syncedInformers](https://github.com/karmada-io/karmada/blob/f7b2a91eed08c65c2d534b31068dd19dffdbca30/pkg/util/informermanager/single-cluster-manager.go#L79) and [handlers](https://github.com/karmada-io/karmada/blob/f7b2a91eed08c65c2d534b31068dd19dffdbca30/pkg/util/informermanager/single-cluster-manager.go#L77).

RainbowMango:
> I think golang-set is reliable since it is used by more than 4.7k projects, including some famous projects such as docker
and ethereum as well.

I know it's well adopted, but once it stops maintenance it'll become a risk for projects depending on it.
That means no bugfix and no regression testing against on following Golang releases.

In addition, I know some companies will not allow using inactive utils.

RainbowMango:
> The lock should not be used both by syncedInformers and handlers.

Why? The lock is used to protect the whole struct(`singleClusterInformerManagerImpl `), we should ensure the lock before making changes on any fields of the struct.

cvvz:
That would be ok use map to implement set by ourselves. But I think the [lock](https://github.com/karmada-io/karmada/blob/f7b2a91eed08c65c2d534b31068dd19dffdbca30/pkg/util/informermanager/single-cluster-manager.go#L79) is not designed to be a struct-lock, in fact it is performed as a read-write lock for two maps in the same struct. Let use following situation as an example, I lock before using `syncedInformers` and then before do I unlock, I lock again because of using `handlers`, that would be a dead lock, right? May be we can use sync.Map to implement `syncedInformers`.

cvvz:
> In addition, I know some companies will not allow using inactive utils.

BTW, I think it is inactive because it is mature.

RainbowMango:
> That would be ok use map to implement set by ourselves.

Don't get me wrong, I didn't mean to implement a set for it. :)

> Let use following situation as an example, I lock before using syncedInformers and then before do I unlock, I lock again because of using handlers, that would be a dead lock, right?

Yes, but we won't lock twice.
Maybe we can introduce some internal methods like `xxxNolock()`.

> May be we can use sync.Map to implement syncedInformers.

That's an option.

cvvz:
> Don't get me wrong, I didn't mean to implement a set for it. :)

But [syncedInformers](https://github.com/karmada-io/karmada/blob/f7b2a91eed08c65c2d534b31068dd19dffdbca30/pkg/util/informermanager/single-cluster-manager.go#L75) is a set and the lock is actually acting as a read-write lock for the two maps in [singleClusterInformerManagerImpl](https://github.com/karmada-io/karmada/blob/f7b2a91eed08c65c2d534b31068dd19dffdbca30/pkg/util/informermanager/single-cluster-manager.go#L69) at the same time.

> Yes, but we won't lock twice.

I think this is just the situation for now, it is still a risk that two maps share the same lock may lead to dead lock.

cvvz:
The Go standard library was kept minimal on purpose and not implement `set`, so we can either use open source solutions directly or implement `set` by ourselves, just like [what kubernenetes did](https://github.com/kubernetes/apimachinery/tree/master/pkg/util/sets).

RainbowMango:
Hi @cvvz the informer manager package changed a lot recently. Do you think this patch is still needed?

cvvz:
I think that it is ok to close.

RainbowMango:
Thanks
/close
I'll close another similar PR #1066 as well.

PR #1073, opened 2021-12-06
Print pointer variables correctly in the scheduler log

iawia002:
**What type of PR is this?**

/kind cleanup

https://github.com/karmada-io/karmada/blob/85920e7ef49afc66325b85a37de97a662a7913c1/pkg/scheduler/core/generic_scheduler.go#L62

Currently we printed the `feasibleClusters` information in the generic_scheduler, but the `feasibleClusters` is an array of pointers, which makes the printed log completely unreadable:

```
I1206 06:22:54.370778       1 generic_scheduler.go:62] feasible clusters found: [0xc000858000 0xc000858200]
```

So I added a `String()` method to the `Cluster` object to make the log readable:

```
I1206 06:51:59.916405       1 generic_scheduler.go:62] feasible clusters found: [member1 member2]
```

Fixes #

```release-note
NONE
```

/assign @RainbowMango

### DIFF
--- a/pkg/apis/cluster/v1alpha1/cluster_utils.go
+++ b/pkg/apis/cluster/v1alpha1/cluster_utils.go
@@ -1,0 +1,6 @@
+package v1alpha1
+
+// String returns a well-formatted string for the Cluster object.
+func (c *Cluster) String() string {
+	return c.Name
+}
--- a/pkg/apis/cluster/v1alpha1/cluster_utils_test.go
+++ b/pkg/apis/cluster/v1alpha1/cluster_utils_test.go
@@ -1,0 +1,57 @@
+package v1alpha1
+
+import (
+	"fmt"
+	"testing"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestString(t *testing.T) {
+	clusterName := "cluster1"
+	cluster1 := &Cluster{
+		ObjectMeta: metav1.ObjectMeta{Name: clusterName},
+	}
+
+	tests := []struct {
+		name    string
+		fmtFunc func() string
+		want    string
+	}{
+		{
+			name: "%s pointer test",
+			fmtFunc: func() string {
+				return fmt.Sprintf("%s", cluster1) // nolint
+			},
+			want: clusterName,
+		},
+		{
+			name: "%v variable test",
+			fmtFunc: func() string {
+				return fmt.Sprintf("%v", *cluster1)
+			},
+			want: "{{ } {cluster1      0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] []  []} {  <nil> false     []} { [] [] <nil> <nil>}}",
+		},
+		{
+			name: "%v pointer test",
+			fmtFunc: func() string {
+				return fmt.Sprintf("%v", cluster1)
+			},
+			want: clusterName,
+		},
+		{
+			name: "%v pointer array test",
+			fmtFunc: func() string {
+				return fmt.Sprintf("%v", []*Cluster{cluster1})
+			},
+			want: fmt.Sprintf("[%s]", clusterName),
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


gy95:
will not this patch  also print like this?
`I1206 06:22:54.370778       1 generic_scheduler.go:62] feasible clusters found: [0xc000858000 0xc000858200]`

iawia002:
> will not this patch also print like this? `I1206 06:22:54.370778 1 generic_scheduler.go:62] feasible clusters found: [0xc000858000 0xc000858200]`

Nope, as I mentioned in the description, after this patch, it will print like this:

```
I1206 06:51:59.916405       1 generic_scheduler.go:62] feasible clusters found: [member1 member2]
```

And we can customize the information we want to print by updating the content returned by the `String` method.

gy95:
Is this function that is implemented by package `klog`?
I use `fmt` for an example, but it still print the address, but not the struct infos.
```
package main

import (
	"fmt"
)

type cluster struct {
	name string
}

func main() {
	slice := []*cluster{
		&cluster{
			name: "a",
		},
		&cluster{
			name: "b",
		},
	}
	fmt.Printf("%v", slice)
}

func (c *cluster) string() string {
	return c.name
}
```

iawia002:
> Is this function that is implemented by package `klog`? I use `fmt` for an example, but it still print the address, but not the struct infos.
> 
> ```
> package main
> 
> import (
> 	"fmt"
> )
> 
> type cluster struct {
> 	name string
> }
> 
> func main() {
> 	slice := []*cluster{
> 		&cluster{
> 			name: "a",
> 		},
> 		&cluster{
> 			name: "b",
> 		},
> 	}
> 	fmt.Printf("%v", slice)
> }
> 
> func (c *cluster) string() string {
> 	return c.name
> }
> ```

The `String` method needs to be exported:

```diff
-func (c *cluster) string() string {
+func (c *cluster) String() string {
```

gy95:
I got this usage for the first time, I thought it was only for `%s` if we define the `String()` function 😂

RainbowMango:
I was thinking to add some methods for our APIs(structs), just you did in this PR. 
But I don't know if this is a good practice, do you know any examples from other projects?

iawia002:
> I was thinking to add some methods for our APIs(structs), just you did in this PR. But I don't know if this is a good practice, do you know any examples from other projects?

I know that chaos-mesh is doing this, eg: https://github.com/chaos-mesh/chaos-mesh/blob/master/api/v1alpha1/awschaos_types.go#L101

I think we can add some public util functions when we encounter specific scenarios in the future (such as this one now).

RainbowMango:
Thanks. All the methods added to the APIs should be considered as a part of the API, which means we need to deal with caution.
When we make changes to the methods should follow consistent compatibility guarantees.

RainbowMango:
There might be quite a lot of methods in the future, do you think we should separate them from the `*_types.go`? 
I mean to introduce a new file under the same package.

iawia002:
> There might be quite a lot of methods in the future, do you think we should separate them from the `*_types.go`? I mean to introduce a new file under the same package.

Yes, for better readability, we should separate them, but what should the file name be, `*_utils.go`?

RainbowMango:
> we should separate them, but what should the file name be, *_utils.go?

That's an option. No preference yet.

iawia002:
> > we should separate them, but what should the file name be, *_utils.go?
> 
> That's an option. No preference yet.

I'm using this for now, we can change it if we have a better idea.

RainbowMango:
OK to me. But should we use `cluster_utils.go`? We can follow the naming rule in the other packages(e.g. policy) too.

And I guess a test file is necessary. We can test it by formats:
| format | receiver | output |
| --- | --- | --- |
| %s | variable | |
| %s | pointer | |
| %v | variable | |
| %v | pointer | |

iawia002:
> OK to me. But should we use cluster_utils.go? We can follow the naming rule in the other packages(e.g. policy) too.

Reasonable, I have updated.

> And I guess a test file is necessary.

I have added some tests, but `%s variable` is not testable because the `%s` placeholder does not match the type `Cluster`.

PR #1081, opened 2021-12-08
Bump kindest/node version to v1.22.4

iawia002:

**What type of PR is this?**

/kind cleanup

When I run `hack/local-up-karmada.sh` locally(macOS), I always encounter the problem that the `karmada-webhook` pod cannot be started. The error message is as follows:

```
E1208 06:08:16.367549       1 webhook.go:99] webhook server exits unexpectedly: too many open files
```

I tried many methods, such as increasing the ulimits configuration of the docker engine, but the problem still exists. When I tried to upgrade the kindest/node version, the problem was solved (I think this is a problem with node image configuration, but I'm not sure), so I submitted a PR to update it, I wonder if other macOS users have encountered this problem?

Fixes #

```release-note
NONE
```


lfbear:
It's probably the problem with the 'kindest/node' version, can I have your macOS's version and the chip type(intel/m1)?
I'll try to verify it.

iawia002:
> It's probably the problem with the 'kindest/node' version, can I have your macOS's version and the chip type(intel/m1)? I'll try to verify it.

Thanks, I'm using macOS 12.0.1 with intel chip + Docker 4.3.0

lfbear:
> > It's probably the problem with the 'kindest/node' version, can I have your macOS's version and the chip type(intel/m1)? I'll try to verify it.
> 
> Thanks, I'm using macOS 12.0.1 with intel chip + Docker 4.3.0

I met the same problem, it is certain that Docker Desktop 4.3.0 will trigger this problem(Docker Desktop 4.2.0 is ok), but upgrading the 'kindest/node' version does not work for me (macOS 12.0.1 with intel chip + Docker Desktop 4.3.0).

iawia002:
> I met the same problem, it is certain that Docker Desktop 4.3.0 will trigger this problem(Docker Desktop 4.2.0 is ok), but upgrading the 'kindest/node' version does not work for me (macOS 12.0.1 with intel chip + Docker Desktop 4.3.0).

That's really weird, I will try to downgrade my docker.

/close

PR #1087, opened 2021-12-10
Introduce ResourceConflictResolution annotation to control the strategy for managing resources with the same name

iawia002:

**What type of PR is this?**

/kind feature

This is part of https://github.com/karmada-io/karmada/discussions/1039, We finally decided to use the annotation solution, we will introduce the following annotation to control the takeover strategy when encountering resources with the same name:

```go
// ResourceConflictResolutionAnnotation is added to the resource template to specify how to resolve the conflict
// in case of resource already existing in member clusters.
// The valid value is:
//   - overwrite: always overwrite the resource if already exist. The resource will be overwritten with the
//     configuration from control plane.
// Note: Propagation of the resource template without this annotation will fail in case of already exists.
ResourceConflictResolutionAnnotation = "work.karmada.io/conflict-resolution"
```

Before this patch, Karmada will not process resources with the same name that already exist in the target cluster. Now we support adding a custom strategy to control this behavior. If you want Karmada to automatically take over existing resources with the same name, you can add a `work.karmada.io/conflict-resolution: overwrite` annotation to the resource template, eg:

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx
  labels:
    app: nginx
  annotations:
    work.karmada.io/conflict-resolution: overwrite
spec:
  replicas: 1
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - image: nginx
        name: nginx
```

In this way Karmada will directly take over existing resources with the same name.

Fixes #

I have tested it with the Nginx sample locally, feel free to comment if I miss something.

```release-note
Introduce ResourceConflictResolution annotation to control the strategy for managing resources with the same name.
```

/cc @RainbowMango @lonelyCZ @dddddai

### DIFF
--- a/pkg/apis/work/v1alpha2/well_known_labels.go
+++ b/pkg/apis/work/v1alpha2/well_known_labels.go
@@ -26,3 +26,17 @@ const (
 	// WorkNameLabel is added to objects to specify associated Work's name.
 	WorkNameLabel = "work.karmada.io/name"
 )
+
+// Define resource conflict resolution
+const (
+	// ResourceConflictResolutionAnnotation is added to the resource template to specify how to resolve the conflict
+	// in case of resource already existing in member clusters.
+	// The valid value is:
+	//   - overwrite: always overwrite the resource if already exist. The resource will be overwritten with the
+	//     configuration from control plane.
+	// Note: Propagation of the resource template without this annotation will fail in case of already exists.
+	ResourceConflictResolutionAnnotation = "work.karmada.io/conflict-resolution"
+
+	// ResourceConflictResolutionOverwrite is the value of ResourceConflictResolutionAnnotation, indicating the overwrite strategy.
+	ResourceConflictResolutionOverwrite = "overwrite"
+)
--- a/pkg/util/annotation.go
+++ b/pkg/util/annotation.go
@@ -21,3 +21,11 @@ func MergeAnnotations(dst *unstructured.Unstructured, src *unstructured.Unstruct
 		MergeAnnotation(dst, key, value)
 	}
 }
+
+// GetAnnotationValue retrieves the value via 'annotationKey' if exist, otherwise returns an empty string.
+func GetAnnotationValue(annotations map[string]string, annotationKey string) string {
+	if annotations == nil {
+		return ""
+	}
+	return annotations[annotationKey]
+}
--- a/pkg/util/objectwatcher/objectwatcher.go
+++ b/pkg/util/objectwatcher/objectwatcher.go
@@ -16,6 +16,7 @@ import (
 
 	configv1alpha1 "github.com/karmada-io/karmada/pkg/apis/config/v1alpha1"
 	workv1alpha1 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha1"
+	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
 	"github.com/karmada-io/karmada/pkg/resourceinterpreter"
 	"github.com/karmada-io/karmada/pkg/util"
 	"github.com/karmada-io/karmada/pkg/util/restmapper"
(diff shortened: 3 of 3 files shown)


pkg/util/annotation.go
```diff
@@ -21,3 +21,11 @@ func MergeAnnotations(dst *unstructured.Unstructured, src *unstructured.Unstruct
 		MergeAnnotation(dst, key, value)
 	}
 }
+
+// GetAnnotationValue retrieves the value via 'annotationKey' if exist, otherwise returns an empty string.
+func GetAnnotationValue(annotations map[string]string, annotationKey string) string {
```
dddddai, line 26: Why is this needed? if key doesn't exist, `map[key]` will be "" naturally

iawia002: The annotations may be nil.

dddddai: Sorry, I misread it

pkg/apis/work/v1alpha2/well_known_labels.go
```diff
@@ -25,4 +25,15 @@ const (
...
+	// ResourceConflictResolutionAnnotation is added to the resource template to specify how to resolve the conflict
+	// in case of resource already existing in member clusters.
+	// The valid value is:
+	//   - overwrite: always overwrite the resource if already exist. The resource will be overwritten with the
+	//     configuration from control plane.
+	// Note: Propagation of the resource template without this annotation will fail in case of already exists.
+	ResourceConflictResolutionAnnotation = "work.karmada.io/conflict-resolution"
+
+	// ResourceConflictResolutionOverwrite is the value of ResourceConflictResolutionAnnotation, indicating the overwrite strategy.
+	ResourceConflictResolutionOverwrite = "overwrite"
```
dddddai, line 38: I guess we should not put a "value" here, maybe put it in `pkg\util\constants.go`?

RainbowMango: It's part of the API, I don't think we should separate them.

iawia002: > It's part of the API, I don't think we should separate them.

Yes, but do we need to put these values in a separate file in the API directory?

RainbowMango: But we can group the resolution by a separated `const`:
```golang
// Define resource conflict resolution
const (
	// ResourceConflictResolutionAnnotation is added to the resource template to specify how to resolve the conflict
	// in case of resource already existing in member clusters.
	// The valid value is:
	//   - overwrite: always overwrite the resource if already exist. The resource will be overwritten with the
	//     configuration from control plane.
	// Note: Propagation of the resource template without this annotation will fail in case of already exists.
	ResourceConflictResolutionAnnotation = "work.karmada.io/conflict-resolution"

	// ResourceConflictResolutionOverwrite is the value of ResourceConflictResolutionAnnotation, indicating the overwrite strategy.
	ResourceConflictResolutionOverwrite = "overwrite"
)
```

RainbowMango (commented):
Only some nits, otherwise LGTM.

RainbowMango:
> I have tested it with the Nginx sample locally, feel free to comment if I miss something.

Does the container re-started after Karmada overwrites the `deployment`?

iawia002:
> > I have tested it with the Nginx sample locally, feel free to comment if I miss something.
> 
> Does the container re-started after Karmada overwrites the `deployment`?

It depends on whether the resource template and the existing deployment template are the same. If they are the same, only the replicas will change and containers will not restart. If they are different, the Pod will be regenerated and started.

pkg/util/objectwatcher/objectwatcher.go
```diff
@@ -84,9 +85,19 @@ func (o *objectWatcherImpl) Create(clusterName string, desireObj *unstructured.U
...
+				switch util.GetAnnotationValue(desireObj.GetAnnotations(), workv1alpha2.ResourceConflictResolutionAnnotation) {
+				case workv1alpha2.ResourceConflictResolutionOverwrite:
+					// if the strategy value is overwrite, we will update the existing resource, and we do nothing else in this case.
+				default:
+					// if the strategy value is empty, we will avoid updating the existing resource, so we directly return an error here.
+					return fmt.Errorf(
+						"resource(kind=%s, %s/%s) already exist in cluster %v and the %s strategy value is empty, karmada will not manage this resource",
+						desireObj.GetKind(), desireObj.GetNamespace(), desireObj.GetName(), clusterName, workv1alpha2.ResourceConflictResolutionAnnotation,
+					)
+				}
```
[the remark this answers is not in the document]
iawia002: I adjusted it to a more understandable structure, a little different from yours, PTAL

pkg/apis/work/v1alpha2/well_known_labels.go
iawia002: Done, PTAL

RainbowMango:
/approve
Leave LGTM to @lonelyCZ or @dddddai 

@iawia002 Have re-tested it after fixing the comments? by the way.

iawia002:
> @iawia002 Have re-tested it after fixing the comments? by the way.

Yes, everything works as expected

PR #1099, opened 2021-12-13
Cleanup names package

iawia002:

**What type of PR is this?**

/kind cleanup

Clean up the `names` package and modify the comments format.

Fixes #

```release-note
NONE
```

/cc @RainbowMango

### DIFF
--- a/pkg/util/names/names.go
+++ b/pkg/util/names/names.go
@@ -16,11 +16,11 @@ const (
 	// - kube-public
 	// - kube-node-lease
 	KubernetesReservedNSPrefix = "kube-"
-	//NamespaceKarmadaSystem is reserved namespace
+	// NamespaceKarmadaSystem is reserved namespace
 	NamespaceKarmadaSystem = "karmada-system"
-	//NamespaceKarmadaCluster is reserved namespace
+	// NamespaceKarmadaCluster is reserved namespace
 	NamespaceKarmadaCluster = "karmada-cluster"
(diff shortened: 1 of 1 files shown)


pkg/util/names/names.go
```diff
@@ -16,12 +16,10 @@ const (
 	// - kube-public
 	// - kube-node-lease
 	KubernetesReservedNSPrefix = "kube-"
-	//NamespaceKarmadaSystem is reserved namespace
+	// NamespaceKarmadaSystem is reserved namespace
 	NamespaceKarmadaSystem = "karmada-system"
-	//NamespaceKarmadaCluster is reserved namespace
+	// NamespaceKarmadaCluster is reserved namespace
```
RainbowMango, line 21: Thanks. I've noticed this issue too.

pkg/util/names/names.go
```diff
@@ -16,12 +16,10 @@ const (
...
 	NamespaceKarmadaCluster = "karmada-cluster"
-	//NamespaceDefault is reserved namespace
-	NamespaceDefault = "default"
```
RainbowMango, line 24: The `const()` block used to define reserved namespaces,  people could get the whole list from here when they read code.
So, I don't think it's worth to remove `default` from here and replace by a reference.

iawia002: You're right, I hadn't thought about that.

RainbowMango (commented):
/lgtm
/approve

PR #1105, opened 2021-12-14
update scheduler log

snowplayfire:
**What type of PR is this?**

Add one of the following kinds:

/kind cleanup

log before update:
```
generic_scheduler.go:61] feasible clusters found: [0xc000890a00 0xc000890c00]
```
I'm confused about the result because `Cluster` has `String()`. 
```
package v1alpha1

// String returns a well-formatted string for the Cluster object.
func (c *Cluster) String() string {
	return c.Name
}

```

log after update:
```
generic_scheduler.go:61] feasible clusters found: [cluster1 cluster2]
```

My solution may be not the best, but the problem exists.

Fixes #

```release-note

```


RainbowMango:
@snowplayfire We added the `String()` method last week(by #1073 @iawia002 ), are you using the master?

snowplayfire:
I'm sorry I got mixed up about branch......

PR #1098, opened 2021-12-13
Add e2e test for managing existing resources scenario

iawia002:
Signed-off-by: Xinzhao Xu <z2d@jifangcheng.com>

**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

Add e2e test for #1087, we create a deployment on the target cluster (member1) first, then create a resource template and a propagation policy, and check the deployment's replicas, at this time, the number of Pods is equal to 2 * the original replicas, then update the deployment, add the ResourceConflictResolution annotation, and then check the deployment's replicas again, now the number of Pods is equal to 3 * the original replicas(the existing deployment of another cluster has been managed).

```
STEP: Creating Namespace(karmadatest-lpw)
STEP: Creating Deployment(karmadatest-lpw/deploy-8s2)
STEP: Creating PropagationPolicy(karmadatest-lpw/deploy-8s2)
STEP: Creating Deployment(karmadatest-lpw/deploy-8s2)
STEP: check deployment's replicas
I1213 15:56:42.395730   81587 manage_existing_resources_test.go:51] Waiting for deployment(karmadatest-lpw/deploy-8s2) collecting status
I1213 15:56:42.411077   81587 manage_existing_resources_test.go:56] deployment(karmadatest-lpw/deploy-8s2) readyReplicas: 0, wanted replicas: 6
I1213 15:56:47.416422   81587 manage_existing_resources_test.go:56] deployment(karmadatest-lpw/deploy-8s2) readyReplicas: 0, wanted replicas: 6
I1213 15:56:52.418734   81587 manage_existing_resources_test.go:56] deployment(karmadatest-lpw/deploy-8s2) readyReplicas: 0, wanted replicas: 6
I1213 15:56:57.419794   81587 manage_existing_resources_test.go:56] deployment(karmadatest-lpw/deploy-8s2) readyReplicas: 0, wanted replicas: 6
I1213 15:57:02.761739   81587 manage_existing_resources_test.go:56] deployment(karmadatest-lpw/deploy-8s2) readyReplicas: 0, wanted replicas: 6
I1213 15:57:07.424440   81587 manage_existing_resources_test.go:56] deployment(karmadatest-lpw/deploy-8s2) readyReplicas: 1, wanted replicas: 6
I1213 15:57:12.418399   81587 manage_existing_resources_test.go:56] deployment(karmadatest-lpw/deploy-8s2) readyReplicas: 4, wanted replicas: 6
[...]

PR #1108, opened 2021-12-15
Fix namespace issue in the porting workloads e2e test

iawia002:

**What type of PR is this?**

/kind failing-test

See our previous discussion at https://github.com/karmada-io/karmada/pull/1098#discussion_r768253161

ref: https://github.com/karmada-io/karmada/pull/1098#issuecomment-993458453

```release-note
NONE
```

/assign @XiShanYongYe-Chang

### DIFF
--- a/test/e2e/porting_workloads_test.go
+++ b/test/e2e/porting_workloads_test.go
@@ -44,6 +44,11 @@ var _ = ginkgo.Describe("porting workloads testing", func() {
 				"Creating deployment(%s/%s) on the member cluster %s first to simulate a scenario where the target cluster already has a deployment with the same name",
 				deploymentNamespace, deploymentName, member1,
 			)
+			gomega.Eventually(func() bool {
+				klog.Infof("Waiting for namespace(%s) to exist on cluster %s", testNamespace, member1)
+				_, err := member1Client.CoreV1().Namespaces().Get(context.TODO(), testNamespace, metav1.GetOptions{})
+				return err == nil
+			}, pollTimeout, pollInterval).Should(gomega.Equal(true))
 			framework.CreateDeployment(member1Client, deployment)
 
 			framework.CreatePropagationPolicy(karmadaClient, policy)


test/e2e/porting_workloads_test.go
```diff
@@ -44,6 +45,9 @@ var _ = ginkgo.Describe("porting workloads testing", func() {
 				"Creating deployment(%s/%s) on the member cluster %s first to simulate a scenario where the target cluster already has a deployment with the same name",
 				deploymentNamespace, deploymentName, member1,
 			)
+			if _, err := member1Client.CoreV1().Namespaces().Get(context.TODO(), testNamespace, metav1.GetOptions{}); apierrors.IsNotFound(err) {
```
XiShanYongYe-Chang, line 48: Is there a case where the namespace is not found during the query but conflicts with the namespace during creation?

iawia002: Almost never, the namespace will exist or not depending on which e2e test executes first, `porting_workloads_test` or `resource_test`

iawia002: Actually I got a little confused about this, or to be safe we can just create the namespace without using the `framework.CreateNamespace` method and not check for errors, we just need it to exist.

XiShanYongYe-Chang: My thoughts are with you.

XiShanYongYe-Chang:
It's not clear what this FOSSA error is: https://github.com/karmada-io/karmada/runs/4529383533?check_suite_focus=true

iawia002:
> It's not clear what this FOSSA error is: https://github.com/karmada-io/karmada/runs/4529383533?check_suite_focus=true

Looks like their server is not stable, maybe we just need to rerun this check.

RainbowMango:
Yes. retriggered.

test/e2e/porting_workloads_test.go
```diff
@@ -44,6 +44,8 @@ var _ = ginkgo.Describe("porting workloads testing", func() {
 				"Creating deployment(%s/%s) on the member cluster %s first to simulate a scenario where the target cluster already has a deployment with the same name",
 				deploymentNamespace, deploymentName, member1,
 			)
+			// We just need to make sure that `testNamespace` exists on the member cluster, no need to check for errors
+			member1Client.CoreV1().Namespaces().Create(context.TODO(), helper.NewNamespace(testNamespace), metav1.CreateOptions{}) // nolint
```
XiShanYongYe-Chang, line 48: What does this `// nolint` comment mean? Record for there is an unhandled err?

iawia002: Yes, otherwise the static check will throw an error:

```
test/e2e/porting_workloads_test.go:48:46: Error return value of `` is not checked (errcheck)
			member1Client.CoreV1().Namespaces().Create(context.TODO(), helper.NewNamespace(testNamespace), metav1.CreateOptions{})
```

RainbowMango: > We just need to make sure that `testNamespace` exists on the member cluster, no need to check for errors

Why? Seems weird.

XiShanYongYe-Chang: thanks, got it.

iawia002: > Why? Seems weird.

Because in some cases the namespace will be created in other cases first, we will get an error if we create it again, but we can ignore that, as long as the namespace exists, we can continue to create the deployment

RainbowMango: I know, but the namespace should be propagated to member cluster by `namespace` controller, it'll be a conflict if we create ns here.

RainbowMango: Why not wait for the ns ready?

XiShanYongYe-Chang: If we need to wait ns ready, we can check like this:
```
gomega.Eventually(func() bool {
	_, err := member1Client.CoreV1().Namespaces().Get(context.TODO(), testNamespace, metav1.GetOptions{})
	if err != nil {
		return false
	}
	return true
}, pollTimeout, pollInterval).Should(gomega.Equal(true))
```

iawia002: > Why not wait for the ns ready?

Reasonable, I have updated

PR #1116, opened 2021-12-16
retain ownerReferences when update member cluster object

snowplayfire:
**What type of PR is this?**

/kind bug
/kind cleanup

Retain ownerReferences since they will typically be set by controllers in a member cluster.

Fixes #

```release-note

```

### DIFF
--- a/pkg/util/objectwatcher/objectwatcher.go
+++ b/pkg/util/objectwatcher/objectwatcher.go
@@ -121,6 +121,10 @@ func (o *objectWatcherImpl) retainClusterFields(desired, observed *unstructured.
 	// controllers in a member cluster.  It is still possible to set the fields
 	// via overrides.
 	desired.SetFinalizers(observed.GetFinalizers())
+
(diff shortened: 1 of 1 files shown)


pkg/util/objectwatcher/objectwatcher.go
```diff
@@ -121,6 +121,10 @@ func (o *objectWatcherImpl) retainClusterFields(desired, observed *unstructured.
 	// controllers in a member cluster.  It is still possible to set the fields
 	// via overrides.
 	desired.SetFinalizers(observed.GetFinalizers())
+
+	// Retain ownerReferences since they will typically be set by controllers in a member cluster.
+	desired.SetOwnerReferences(observed.GetOwnerReferences())
```
RainbowMango, line 126: That makes sense. Here we don't `merge` the owner reference because we usually won't set owner reference in the resource template.

@mrlihanbo Could you please double confirm it?

PR #1093, opened 2021-12-11
Enhancement --controllers flag of karmada-controller-manager

jameszhangyukun:
**What type of PR is this?**

/kind feature
**What this PR does / why we need it**:
Enhancement --controllers flag of karmada-controller-manager
**Which issue(s) this PR fixes**:
Part of  #1085

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:
```release-note
NONE
```

PR #1123, opened 2021-12-17
controller-manager: automatically generate all controllers list in flag

### LINKED ISSUE #1085: Enhancement `--controllers` flag of `karmada-controller-manager`
**Background**
#1083(by @snowplayfire) introduced `--controllers` flag to `karmada-controller-manager` for customize controllers:
```
--controllers string                            A list of controllers to enable. '*' enables all on-by-default controllers, 'foo' enables the controller named 'foo', '-foo' disables the controller named 'foo'. All controllers: cluster, clusterStatus, hpa, binding, execution, workStatus, namespace, serviceExport, endpointSlice, serviceImport. (default "*")
```

The default enabled controller list now is `hard-coded`:
https://github.com/karmada-io/karmada/blob/43d75741e84ef31e1ce7b06f90f0ef8a31220a4c/cmd/controller-manager/app/options/options.go#L81

It'd be great if we can generate the controller list automatically, otherwise, if we introduce a new controller in the future, we might forget to update.

**Why is this needed**:

iawia002:

**What type of PR is this?**

/kind cleanup

This patch updates the comments I mentioned in https://github.com/karmada-io/karmada/pull/1093#pullrequestreview-829967849 and now the all controllers list in flag is not hard-coded:

```
      --controllers strings                           A list of controllers to enable. '*' enables all on-by-default controllers, 'foo' enables the controller named 'foo', '-foo' disables the controller named 'foo'. All controllers: binding, cluster, clusterStatus, endpointSlice, execution, hpa, namespace, serviceExport, serviceImport, workStatus. (default [*])
```

Fixes #1085

```release-note
NONE
```

/cc @RainbowMango

### DIFF
--- a/cmd/controller-manager/app/controllermanager.go
+++ b/cmd/controller-manager/app/controllermanager.go
@@ -8,6 +8,7 @@ import (
 	"strconv"
 
 	"github.com/spf13/cobra"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/dynamic"
 	kubeclientset "k8s.io/client-go/kubernetes"
@@ -60,7 +61,7 @@ func NewControllerManagerCommand(ctx context.Context) *cobra.Command {
 
 	cmd.Flags().AddGoFlagSet(flag.CommandLine)
 	cmd.AddCommand(sharedcommand.NewCmdVersion(os.Stdout, "karmada-controller-manager"))
-	opts.AddFlags(cmd.Flags())
+	opts.AddFlags(cmd.Flags(), KnownControllers())
 	return cmd
 }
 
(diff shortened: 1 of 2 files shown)


cmd/controller-manager/app/controllermanager.go
```diff
@@ -135,6 +153,11 @@ func NewControllerInitializers() map[string]InitFunc {
 	return controllers
 }
 
+// KnownControllers returns all known controllers's name
+func KnownControllers() []string {
+	return sets.StringKeySet(NewControllerInitializers()).List()
```
RainbowMango, line 158: Should we sort the controllers? I'm afraid the list is unordered(the map is unordered).

cmd/controller-manager/app/controllermanager.go
```diff
@@ -146,7 +169,6 @@ func startClusterController(ctx ControllerContext) (enabled bool, err error) {
 		ClusterStartupGracePeriod: opts.ClusterStartupGracePeriod.Duration,
 	}
 	if err := clusterController.SetupWithManager(mgr); err != nil {
-		klog.Fatalf("Failed to setup cluster controller: %v", err)
```
RainbowMango, line 149: Why do you think we should remove these logs?

RainbowMango (commented):
Generally looks good to me.
Just some questions.

cmd/controller-manager/app/controllermanager.go
iawia002: The `List` method will sort the results.

RainbowMango: Oh, thanks. Good to know.

cmd/controller-manager/app/controllermanager.go
iawia002: This is really unnecessary cause we can do this uniformly in `StartControllers` like I mentioned in https://github.com/karmada-io/karmada/pull/1093#discussion_r768272821

RainbowMango: Yeah, I'm sold.

PR #1104, opened 2021-12-14
add impersonation config sync controller

mrlihanbo:
Signed-off-by: lihanbo <lihanbo2@huawei.com>

**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
1. create a service account for impersonation when joining clusters.
2. add impersonation config sync controller

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:
self test:
step1: create a clusterrole in karmada control plane
```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: karmada-cluster-proxy
rules:
- apiGroups:
  - '*'
  resources:
  - cluster/proxy
  resourceNames:
  - member1
  - member2
  verbs:
  - '*'
- nonResourceURLs:
  - '*'
  verbs:
  - get
```
step2: create a clusterrolebinding in karmada control plane
```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: karmada-cluster-proxy
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: karmada-cluster-proxy
subjects:
  - kind: User
    name: karmada-users
  - kind: ServiceAccount
    name: default
    namespace: karmada-cluster
```
step3: check clusterrole and clusterrolebinding in member cluster
```
kubectl --kubeconfig=/root/.kube/members.config --context=member1 get clusterrole karmada-impersonator -oyaml 
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  creationTimestamp: "2021-12-14T08:56:33Z"
  labels:
    work.karmada.io/name: karmada-impersonator-7cbb6bd5c9
    work.karmada.io/namespace: karmada-es-member1
  name: karmada-impersonator
  resourceVersion: "742"
  uid: a037eede-7893-4be9-8182-e968844541f3
rules:
- apiGroups:
  - ""
  resourceNames:
  - karmada-users
  resources:
  - users
  verbs:
  - impersonate
- apiGroups:
  - ""
  resourceNames:
  - system:serviceaccount:karmada-cluster:default
  resources:
  - serviceaccounts
  verbs:
  - impersonate
```

**Does this PR introduce a user-facing change?**:
"NONE"

PR #1131, opened 2021-12-20
Add --controllers flag to karmada-agent and move ControllerContext to a separate package

### LINKED ISSUE #1127: Add `--controllers` flag to `karmada-agent`
Add a `--controllers` flag to specify enabled controllers for `karmada-agent`. Just like we did for `karmada-controller-manager` at #1083/#1084/#1123/#1093. 

**Why is this needed**:

iawia002:

**What type of PR is this?**

/kind feature

Add `--controllers` flag to karmada-agent and move `ControllerContext` to a separate package.

```
      --controllers strings                           A list of controllers to enable. '*' enables all on-by-default controllers, 'foo' enables the controller named 'foo', '-foo' disables the controller named 'foo'. All controllers: clusterStatus, execution, serviceExport, workStatus. (default [*])
```

close #1127

```release-note
Add --controllers flag to karmada-agent.
```

/cc @RainbowMango

### DIFF
--- a/cmd/agent/app/agent.go
+++ b/cmd/agent/app/agent.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/karmada-io/karmada/cmd/agent/app/options"
 	clusterv1alpha1 "github.com/karmada-io/karmada/pkg/apis/cluster/v1alpha1"
+	controllerscontext "github.com/karmada-io/karmada/pkg/controllers/context"
 	"github.com/karmada-io/karmada/pkg/controllers/execution"
 	"github.com/karmada-io/karmada/pkg/controllers/mcs"
 	"github.com/karmada-io/karmada/pkg/controllers/status"
@@ -60,10 +61,19 @@ func NewAgentCommand(ctx context.Context) *cobra.Command {
 		},
 	}
 
-	opts.AddFlags(cmd.Flags())
+	opts.AddFlags(cmd.Flags(), controllers.ControllerNames())
 	cmd.AddCommand(sharedcommand.NewCmdVersion(os.Stdout, "karmada-agent"))
 	cmd.Flags().AddGoFlagSet(flag.CommandLine)
 	return cmd
+}
+
+var controllers = make(controllerscontext.Initializers)
+
+func init() {
+	controllers["clusterStatus"] = startClusterStatusController
+	controllers["execution"] = startExecutionController
+	controllers["workStatus"] = startWorkStatusController
+	controllers["serviceExport"] = startServiceExportController
 }
 
 func run(ctx context.Context, karmadaConfig karmadactl.KarmadaConfig, opts *options.Options) error {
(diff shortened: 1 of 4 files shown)


RainbowMango:
/assign @Garrybest 
Can you help review it?

pkg/controllers/context/context.go
```diff
@@ -0,0 +1,103 @@
...
+// Context defines the context object for controller.
+type Context struct {
+	Mgr                         controllerruntime.Manager
+	ObjectWatcher               objectwatcher.ObjectWatcher
+	Opts                        Options
+	StopChan                    <-chan struct{}
+	DynamicClientSet            dynamic.Interface
+	OverrideManager             overridemanager.OverrideManager
+	ControlPlaneInformerManager informermanager.SingleClusterInformerManager
+}
```
RainbowMango, line 59: The controllers running in `karmada-agent` and `karmada-controller-manager` might not always share the same `Context`. 
So, I'm thinking if it is worth abstracting a common `Context`.

iawia002: I have also considered this issue, but the agent is also a collection of several specific controllers, it's the same with controller-manager. With this abstraction, we can reuse part of the code

RainbowMango: > With this abstraction, we can reuse part of the code

Yes, That's the benefit. We also have to consider the maintenance effort that brings. Nothing right or wrong, just the tradeoff.

As far as I am concerned, if the benefits are not obvious I'd hesitate to do it.

Garrybest:
Thanks, I'll check it later.

cmd/agent/app/agent.go
```diff
@@ -60,12 +62,27 @@ func NewAgentCommand(ctx context.Context) *cobra.Command {
...
 
-	opts.AddFlags(cmd.Flags())
+	opts.AddFlags(cmd.Flags(), KnownControllers())
 	cmd.AddCommand(sharedcommand.NewCmdVersion(os.Stdout, "karmada-agent"))
 	cmd.Flags().AddGoFlagSet(flag.CommandLine)
 	return cmd
 }
 
+// NewControllerInitializers is a public map of named controller groups (you can start more than one in an init func)
+func NewControllerInitializers() map[string]controllerscontext.InitFunc {
```
Garrybest, line 72: Well, why not make the Initializers as a global variable? It seems we don't need to modify this struct. I prefer this kind of code.

```go
var (
	controllers = make(map[string]controllerscontext.InitFunc)
)

func init() {
	controllers["clusterStatus"] = startClusterStatusController
	controllers["execution"] = startExecutionController
	controllers["workStatus"] = startWorkStatusController
	controllers["serviceExport"] = startServiceExportController
}

type map[string]controllerscontext.InitFunc ControllerInitializers

func (c *ControllerInitializers) ControllerNames() []string {
    return sets.StringKeySet(c).List()
}

func (c *ControllerInitializers) StartControllers(ctx Context) error {
    ......
}
```

iawia002: > Well, why not make the Initializers as a global variable?

Similar to what I mentioned before: https://github.com/karmada-io/karmada/pull/1093#discussion_r767553986, and I'm just following the way the controller-manager, I will update them all.

Garrybest: Cool, would you update in this PR or next?

iawia002: Just in this PR, I will update it soon.

iawia002: Updated, PTAL @Garrybest

Garrybest:
Nice work and I like it! Now the code is clean.
/lgtm

iawia002:
> Now the code is clean.

Thanks for your review!

/assign @RainbowMango 
Can you take another look?

pkg/controllers/context/context.go
RainbowMango: OK, then let's do it since you both insist.

RainbowMango:
@iawia002 The #1104 introduced another controller, you might have to solve the conflicts. Thanks.

pkg/controllers/context/context.go
```diff
@@ -0,0 +1,112 @@
...
+
+// InitFunc is used to launch a particular controller.
+// Any error returned will cause the controller process to `Fatal`
+// The bool indicates whether the controller was enabled.
+type InitFunc func(ctx Context) (enabled bool, err error)
+
+// Initializers is a public map of named controller groups
+type Initializers map[string]InitFunc
+
+// ControllerNames returns all known controllers's name
```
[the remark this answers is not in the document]
iawia002: Updated.

RainbowMango (approved):
/approve
Leave LGTM to @Garrybest for final confirmation.

PR #1146, opened 2021-12-22
changed the installation mode of k8s kind

### LINKED ISSUE #1141: install with local-up-karmada.sh, cannot be deleted with delete-cluster.sh
**What happened**:
I use local-up-karma.sh to test, then I want to recycle resources, I found delete-cluste.sh.
When I use it to delete a cluster. such as member1, I use it to delete a cluster, an error is returned.
```shell
hack/delete-cluster.sh member1 /root/.kube/karmada.config
>>> error: cannot rename the context "member1", it's not in /root/.kube/karmada.config
```
when i used "kubectl config use-context member1" first
hack/delete-cluster.sh member1 /root/.kube/karmada.config is work.
also
https://github.com/karmada-io/karmada/blob/b57b7235044fea6ecec4b5ed7907ae0dcef582ab/hack/util.sh#L137
the next time I come in, I find that kind can't find it.

First I can delete the cluster correctly
Second I can use kind correctly, or guide me for adding a directory to my PATH
```shell
git clone https://github.com/karmada-io/karmada
cd karmada
hack/local-up-karmada.sh
```
Execute the kind command
Execute hack/delete-cluster.sh

**Environment**:
- Karmada version:
- Others:

leonharetd:
**What type of PR is this?**
/kind bug
Changed the installation mode of kind

Fixes #1141 
@lfbear @RainbowMango

```release-note
NONE
```

### DIFF
--- a/hack/util.sh
+++ b/hack/util.sh
@@ -25,7 +25,7 @@ MIN_Go_VERSION=go1.16.0
 function util::install_tools() {
 	local package="$1"
 	local version="$2"
-
+	echo "go install ${package}@${version}"
 	GO111MODULE=on go install "${package}"@"${version}"
 	GOPATH=$(go env GOPATH | awk -F ':' '{print $1}')
 	export PATH=$PATH:$GOPATH/bin
@@ -117,28 +117,6 @@ function util::install_kubectl {
         echo "Failed to install kubectl, can not download the binary file at https://dl.k8s.io/release/$KUBECTL_VERSION/bin/$OS/$ARCH/kubectl"
         exit 1
     fi
-}
-
-# util::install_kind will install the given version kind
-function util::install_kind {
-  local kind_version=${1}
-  echo "Installing 'kind ${kind_version}' for you"
-  local os_name
-  os_name=$(go env GOOS)
-  local arch_name
-  arch_name=$(go env GOARCH)
-  curl --retry 5 -sSLo ./kind -w "%{http_code}" "https://kind.sigs.k8s.io/dl/${kind_version}/kind-${os_name:-linux}-${arch_name:-amd64}" | grep '200' > /dev/null
-  ret=$?
-  if [ ${ret} -eq 0 ]; then
-      chmod +x ./kind
-      mkdir -p ~/.local/bin/
-      mv ./kind ~/.local/bin/kind
-
-      export PATH=$PATH:~/.local/bin
-  else
-      echo "Failed to install kind, can not download the binary file at https://kind.sigs.k8s.io/dl/${kind_version}/kind-${os_name:-linux}-${arch_name:-amd64}"
-      exit 1
-  fi
 }
 
 # util::create_signing_certkey creates a CA, args are sudo, dest-dir, ca-id, purpose
--- a/hack/README.md
+++ b/hack/README.md
@@ -51,6 +51,9 @@ ensures development quality.
   please use it like this: `hack/undeploy-karmada.sh <KUBECONFIG> <CONTEXT_NAME>`, the same parameters as you input at
   the installing step.
 
+- [`delete-cluster.sh`](delete-cluster.sh) This script delete a kube cluster by kind,
+  please use it like this: `hack/delete-cluster.sh.sh <CLUSTER_NAME> <KUBECONFIG>`
+
 ## For CI pipeline
 - [`local-up-karmada.sh`](local-up-karmada.sh) This script also used for testing.
 
--- a/hack/local-up-karmada.sh
+++ b/hack/local-up-karmada.sh
@@ -65,7 +65,7 @@ if util::cmd_exist kind; then
   echo "passed"
 else
   echo "not pass"
-  util::install_kind $kind_version
+  util::install_tools "sigs.k8s.io/kind" $kind_version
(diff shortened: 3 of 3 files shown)


hack/util.sh
```diff
@@ -25,7 +25,7 @@ MIN_Go_VERSION=go1.16.0
 function util::install_tools() {
 	local package="$1"
 	local version="$2"
-
+  echo "go install ${package}@${version}"
```
lfbear, line 28: Maybe you can help to keep the space style the same. Seems some bad `tab`s here:)

lfbear:
Thanks for your first contribution. And it looks generally good to me.
For uninstalling the Karmada, I just have an idea, shall we have a `one line` uninstall as we installed. For example, when I run `hack/delete-cluster.sh all` it deletes all clusters created by our script and is free from finding clusters' names and kubeconfig files. That's just an idea.

leonharetd:
> 
It's a good idea. how about local-up-karmada.sh install/uninstall/add/delete? Then change the name of local-up-karmada.sh to local-karmada.sh. Use local-karmada.sh to manage locally created clusters.

lfbear:
> 

afaik, the responsibility of the script `local-up-karmada.sh` is that start up a local Karmada environment with `kind` quickly for development. And the prefix of the name `local-up` gradually become a  conventional name for quickly installing tools, just like `setup`, I have never seen the way like `local-up-karmada.sh install/uninstall`, that maybe not compliance with public habits I think.
And the uninstalling operation most likely possibility is for debugging the installation, you know, install and uninstall all over. So I do not suggest integrating into one shell script for these different user purposes.

leonharetd:
> > 
> 
> afaik, the responsibility of the script `local-up-karmada.sh` is that start up a local Karmada environment with `kind` quickly for development. And the prefix of the name `local-up` gradually become a conventional name for quickly installing tools, just like `setup`, I have never seen the way like `local-up-karmada.sh install/uninstall`, that maybe not compliance with public habits I think. And the uninstalling operation most likely possibility is for debugging the installation, you know, install and uninstall all over. So I do not suggest integrating into one shell script for these different user purposes.

Let me see.

hack/delete-cluster.sh
```diff
@@ -26,9 +26,14 @@ if [[ -z "${KUBECONFIG}" ]]; then
...
 
+REPO_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
+source "${REPO_ROOT}"/hack/util.sh
+# befor delete cluster, check kind command exist, Ensure that it can be deleted directly after renaming
+util::cmd_must_exist "kind"
+
 # The context name has been changed when creating clusters by 'create-cluster.sh'.
 # This will result in the context can't be removed by kind when deleting a cluster.
 # So, we need to change context name back and let kind take care about it.
+kubectl config use-context "${CLUSTER_NAME}" --kubeconfig="${KUBECONFIG}"
```
lfbear, line 37: why need `use-context` first?

README.md
```diff
@@ -146,9 +146,21 @@ There are two contexts about karmada:
 
 The `karmada-apiserver` is the **main kubeconfig** to be used when interacting with karmada control plane, while `karmada-host` is only used for debugging karmada installation with the host cluster. You can check all clusters at any time by running: `kubectl config view`. To switch cluster contexts, run `kubectl config use-context [CONTEXT_NAME]`
 
+### Uninstall karmada control plane
```
lfbear, line 149: Those docs should be in `hack/README.md` not in `/README.md`, it will take users more puzzle at getting started.

hack/delete-cluster.sh
```diff
@@ -6,29 +6,36 @@ set -o pipefail
...
+# befor delete cluster, check kind command exist, Ensure that it can be deleted directly after renaming
+util::cmd_must_exist "kind"
 
-# The context name has been changed when creating clusters by 'create-cluster.sh'.
-# This will result in the context can't be removed by kind when deleting a cluster.
-# So, we need to change context name back and let kind take care about it.
-kubectl config rename-context "${CLUSTER_NAME}" "kind-${CLUSTER_NAME}" --kubeconfig="${KUBECONFIG}"
+case $1 in
+    all)
+        kind delete clusters --all
```
lfbear, line 21: That's great to provide the `all` parameter, I think lots of people will use it.
But I am a little worried about deleting all clusters if I have other 'kind clusters' unrelated to Karmada, those will delete too. Shall we just delete the clusters that we will create?

lfbear: BTW, the others changes seem all right.

leonharetd: I filtered by name，Maybe there's something better.

lfbear:
It seems more puzzled for me than the previous version, because I feel deleting clusters should be more direct. Just give me the name of the cluster to be deleted, then we'll delete it.
Let's review this requirement together, `hack/delete-cluster.sh` is a tool script for deleting some cluster created by `hack/local-up-karmada.sh`, actually deleting container created by `Kubernetes-sigs/kind`, so could we directly delete cluster by name, because the `host-cluster` has the fixed name and `member-clusters` have the same prefix in the name. 
In a word, we need a more easy tool to use.

PR #1145, opened 2021-12-22
fix ensureWork cyclomatic complexity

kerthcet:
Signed-off-by: kerthcet <kerthcet@gmail.com>

**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
fix cyclomatic complexity

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #1149, opened 2021-12-22
Implementing the ReviseReplica hook

### LINKED ISSUE #1111: Add interpreter.ReviseReplica hook when create work
Recently, I tried to work with ResourceInterpreterWebhook framework which is the new feature in karmada v0.10.0 .
I tried `InterpretReplica` hook, and it work fine for me , But I  found a problem with `ReviseReplica`. 
I cannt find any hook point to call `ReviseReplica`, both in karmada v0.10.0. & v0.10.1 , which like `InterpretReplica ` hook point 
 `ResourceInterpreter.GetReplicas(object) ... ` 

**Why is this needed**:
Now , I interperter my CR's replica and scheduled well, but Karmada Work.workload.manifests replicas cannt be revised so that replica cannot be divided in member cluster. 

Should add  `ReviseReplica` hook point in function `ensureWork` ?

iawia002:

**What type of PR is this?**

/kind feature

This patch implements the `ReviseReplica` hook, check [resource interpreter webhook proposal](https://github.com/karmada-io/karmada/tree/master/docs/proposals/resource-interpreter-webhook) for more info.

The original code to modify the replica of `Deployment` and `Job` has been changed to use default interpreters, for custom resources, users only need to implement the corresponding webhook and return the patch that updates the replica, take our `workload` resource for example:

```go
func (e *workloadInterpreter) responseWithExploreReviseReplica(workload *workloadv1alpha1.Workload, req interpreter.Request) interpreter.Response {
	wantedWorkload := workload.DeepCopy()
	// update the replica
	wantedWorkload.Spec.Replicas = req.DesiredReplicas
	marshaledBytes, err := json.Marshal(wantedWorkload)
	if err != nil {
		return interpreter.Errored(http.StatusInternalServerError, err)
	}
	// return the patch response
	return interpreter.PatchResponseFromRaw(req.Object.Raw, marshaledBytes)
}
```

I've tested it locally with [the workload example](https://github.com/karmada-io/karmada/tree/master/examples/customresourceinterpreter), everything works as excepted, my test files:

workload:

```yaml
apiVersion: workload.example.io/v1alpha1
kind: Workload
metadata:
  name: nginx
  labels:
    app: nginx
spec:
  replicas: 3
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - image: nginx
        name: nginx
```

propagation policy:

```yaml
kind: PropagationPolicy
metadata:
  name: nginx-workload-propagation
spec:
  resourceSelectors:
    - apiVersion: workload.example.io/v1alpha1
      kind: Workload
      name: nginx
  placement:
    clusterAffinity:
      clusterNames:
        - member1
        - member2
    replicaScheduling:
      weightPreference:
        staticWeightList:
          - targetCluster:
              clusterNames:
                - member1
            weight: 2
          - targetCluster:
              clusterNames:
                - member2
            weight: 1
```

And in the cluster member1, the workload detail:

```yaml
apiVersion: workload.example.io/v1alpha1
kind: Workload
metadata:
  annotations:
    resourcebinding.karmada.io/name: nginx-workload
    resourcebinding.karmada.io/namespace: default
  creationTimestamp: "2021-12-22T09:16:02Z"
  generation: 1
  labels:
    app: nginx
    propagationpolicy.karmada.io/name: nginx-workload-propagation
    resourcebinding.karmada.io/key: 7f8d449df9
    work.karmada.io/name: nginx-7f8d449df9
    work.karmada.io/namespace: karmada-es-member1
  name: nginx
  namespace: default
  resourceVersion: "5448"
  uid: 8ebe4ee5-d6df-4e62-956c-29e1a3da613b
spec:
  replicas: 2
  template:
    metadata: {}
    spec:
      containers:
      - image: nginx
        name: nginx
        resources: {}
```

In the cluster member2, the workload detail:

```yaml
apiVersion: workload.example.io/v1alpha1
kind: Workload
metadata:
  annotations:
    resourcebinding.karmada.io/name: nginx-workload
    resourcebinding.karmada.io/namespace: default
  creationTimestamp: "2021-12-22T09:16:02Z"
  generation: 1
  labels:
    app: nginx
    propagationpolicy.karmada.io/name: nginx-workload-propagation
    resourcebinding.karmada.io/key: 7f8d449df9
    work.karmada.io/name: nginx-7f8d449df9
    work.karmada.io/namespace: karmada-es-member2
  name: nginx
  namespace: default
  resourceVersion: "5465"
  uid: dd92ed92-8ec9-41a6-a2b2-089173f76085
spec:
  replicas: 1
  template:
    metadata: {}
    spec:
      containers:
      - image: nginx
        name: nginx
        resources: {}
```

Fixes #1111

```release-note
Implementing the ReviseReplica hook.
```

/cc @RainbowMango @Garrybest @XiShanYongYe-Chang

### DIFF
--- a/pkg/controllers/binding/binding_controller.go
+++ b/pkg/controllers/binding/binding_controller.go
@@ -23,6 +23,7 @@ import (
 	policyv1alpha1 "github.com/karmada-io/karmada/pkg/apis/policy/v1alpha1"
 	workv1alpha1 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha1"
 	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
+	"github.com/karmada-io/karmada/pkg/resourceinterpreter"
 	"github.com/karmada-io/karmada/pkg/util"
 	"github.com/karmada-io/karmada/pkg/util/helper"
 	"github.com/karmada-io/karmada/pkg/util/informermanager"
@@ -34,12 +35,13 @@ const ControllerName = "binding-controller"
 
 // ResourceBindingController is to sync ResourceBinding.
 type ResourceBindingController struct {
-	client.Client                                                // used to operate ClusterResourceBinding resources.
-	DynamicClient   dynamic.Interface                            // used to fetch arbitrary resources from api server.
-	InformerManager informermanager.SingleClusterInformerManager // used to fetch arbitrary resources from cache.
-	EventRecorder   record.EventRecorder
-	RESTMapper      meta.RESTMapper
-	OverrideManager overridemanager.OverrideManager
+	client.Client                                                    // used to operate ClusterResourceBinding resources.
+	DynamicClient       dynamic.Interface                            // used to fetch arbitrary resources from api server.
+	InformerManager     informermanager.SingleClusterInformerManager // used to fetch arbitrary resources from cache.
+	EventRecorder       record.EventRecorder
+	RESTMapper          meta.RESTMapper
+	OverrideManager     overridemanager.OverrideManager
+	ResourceInterpreter resourceinterpreter.ResourceInterpreter
 }
 
 // Reconcile performs a full reconciliation for the object referred to by the Request.
@@ -110,7 +112,7 @@ func (c *ResourceBindingController) syncBinding(binding *workv1alpha2.ResourceBi
 		return controllerruntime.Result{Requeue: true}, err
 	}
 	var errs []error
-	err = ensureWork(c.Client, workload, c.OverrideManager, binding, apiextensionsv1.NamespaceScoped)
+	err = ensureWork(c.Client, c.ResourceInterpreter, workload, c.OverrideManager, binding, apiextensionsv1.NamespaceScoped)
 	if err != nil {
 		klog.Errorf("Failed to transform resourceBinding(%s/%s) to works. Error: %v.",
 			binding.GetNamespace(), binding.GetName(), err)
--- a/pkg/controllers/binding/common.go
+++ b/pkg/controllers/binding/common.go
@@ -15,9 +15,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	clusterv1alpha1 "github.com/karmada-io/karmada/pkg/apis/cluster/v1alpha1"
+	configv1alpha1 "github.com/karmada-io/karmada/pkg/apis/config/v1alpha1"
 	policyv1alpha1 "github.com/karmada-io/karmada/pkg/apis/policy/v1alpha1"
 	workv1alpha1 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha1"
 	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
+	"github.com/karmada-io/karmada/pkg/resourceinterpreter"
 	"github.com/karmada-io/karmada/pkg/util"
 	"github.com/karmada-io/karmada/pkg/util/helper"
 	"github.com/karmada-io/karmada/pkg/util/names"
@@ -58,7 +60,10 @@ var workPredicateFn = builder.WithPredicates(predicate.Funcs{
 // ensureWork ensure Work to be created or updated.
 // TODO(Garrybest): clean up the code to fix cyclomatic complexity
 //nolint:gocyclo
-func ensureWork(c client.Client, workload *unstructured.Unstructured, overrideManager overridemanager.OverrideManager, binding metav1.Object, scope apiextensionsv1.ResourceScope) error {
+func ensureWork(
+	c client.Client, resourceInterpreter resourceinterpreter.ResourceInterpreter, workload *unstructured.Unstructured,
+	overrideManager overridemanager.OverrideManager, binding metav1.Object, scope apiextensionsv1.ResourceScope,
+) error {
 	var targetClusters []workv1alpha2.TargetCluster
 	switch scope {
 	case apiextensionsv1.NamespaceScoped:
@@ -69,7 +74,7 @@ func ensureWork(c client.Client, workload *unstructured.Unstructured, overrideMa
 		targetClusters = bindingObj.Spec.Clusters
 	}
 
-	hasScheduledReplica, referenceRSP, desireReplicaInfos, err := getRSPAndReplicaInfos(c, workload, targetClusters)
+	hasScheduledReplica, _, desireReplicaInfos, err := getRSPAndReplicaInfos(c, workload, targetClusters)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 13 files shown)


RainbowMango:
Pretty good after a quick look at the testing report. I'll review it ASAP.

@GodLoveMinch Is this the feature that you are asking at #1111.

pkg/controllers/binding/binding_controller.go
```diff
@@ -110,7 +112,7 @@ func (c *ResourceBindingController) syncBinding(binding *workv1alpha2.ResourceBi
 		return controllerruntime.Result{Requeue: true}, err
 	}
 	var errs []error
-	err = ensureWork(c.Client, workload, c.OverrideManager, binding, apiextensionsv1.NamespaceScoped)
+	err = ensureWork(c.Client, c.ResourceInterpreter, workload, c.OverrideManager, binding, apiextensionsv1.NamespaceScoped)
```
RainbowMango, line 115: Too much parameters required by `ensureWork` now, maybe you can help review the #1145.

pkg/resourceinterpreter/interpreter.go
```diff
@@ -95,6 +98,25 @@ func (i *customResourceInterpreterImpl) GetReplicas(object *unstructured.Unstruc
 	return
 }
 
+// ReviseReplica revises the replica of the given object.
+func (i *customResourceInterpreterImpl) ReviseReplica(object *unstructured.Unstructured, replica int64) (*unstructured.Unstructured, error) {
+	klog.V(4).Infof("Begin to revise replicas for object: %v %s/%s.", object.GroupVersionKind(), object.GetNamespace(), object.GetName())
+
+	patch, patchType, hookEnabled, err := i.customizedInterpreter.Retain(context.TODO(), &webhook.RequestAttributes{
```
RainbowMango, line 105: Why use `customizedInterpreter.Retain` here? Do you want to share the same logic?

RainbowMango (commented):
Review still on-going, post my comments first as I see the base changed.

pkg/resourceinterpreter/interpreter.go
iawia002: Actually the logic is the same, I could add a new `customizedInterpreter.ReviseReplica` method but it will be just the same as `customizedInterpreter.Retain`

pkg/controllers/binding/binding_controller.go
iawia002: Yes but these are all needed, IMHO #1145 just dived one function into several parts, it doesn't solve the too much parameters problem, I think this function will be clearer and more unified after we remove the deprecated RSP API.

iawia002:
> as I see the base changed.

Oh just typo and code format changed, no logic changed.

pkg/resourceinterpreter/interpreter.go
iawia002: Maybe we can rename the `customizedInterpreter.Retain` to `Patch`, `ReviseReplica` and `Retain` can share this common method.

pkg/resourceinterpreter/utils/utils.go
```diff
@@ -0,0 +1,22 @@
+package utils
+
+import (
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	"github.com/karmada-io/karmada/pkg/util"
+)
+
+// ApplyReplica applies the Replica value for the specific field.
+func ApplyReplica(workload *unstructured.Unstructured, desireReplica int64, field string) error {
```
RainbowMango, line 10: This function just takes `desireReplica` as the common value, it cloud be used to set any fields with `int64` type, right?
I mean the common function is misleading.

By the way, I don't think we should add a `utils` package here, too many `utils` in the same repo is also confusing, we have to use ugly import alias to distinguish them.
Not sure if we need this function yet, if it does, it should be `https://github.com/karmada-io/karmada/blob/master/pkg/util/helper/unstructured.go`.

pkg/controllers/binding/common.go
```diff
@@ -104,26 +110,18 @@ func ensureWork(c client.Client, workload *unstructured.Unstructured, overrideMa
 
 		workLabel := mergeLabel(clonedWorkload, workNamespace, binding, scope)
 
-		if clonedWorkload.GetKind() == util.DeploymentKind && (referenceRSP != nil || hasScheduledReplica) {
-			err = applyReplicaSchedulingPolicy(clonedWorkload, desireReplicaInfos[targetCluster.Name], util.ReplicasField)
+		if (hasScheduledReplica || referenceRSP != nil) && resourceInterpreter.HookEnabled(clonedWorkload, configv1alpha1.InterpreterOperationReviseReplica) {
```
RainbowMango, line 113: The `RSP` has been deprecated and should be cleaned up. I'll open an issue to track it.

pkg/resourceinterpreter/interpreter.go
RainbowMango: Not rename, we can abstract a common method shared by `Retain()` and `ReviseReplica()`, for code readability.

pkg/resourceinterpreter/utils/utils.go
iawia002: > it cloud be used to set any fields with int64 type, right?

Yes any fields under `spec`, so should we rename it? Like `ApplySpecInt64Value`, it's weird and there are no others who will use this function.

> it should be https://github.com/karmada-io/karmada/blob/master/pkg/util/helper/unstructured.go.

Sure.

pkg/controllers/binding/common.go
iawia002: I have not used the `referenceRSP` check logic, this part of the code will be cleaned up separately after the PR merge, #1153

iawia002:
All comments addressed and solved the conflicts, PTAL @RainbowMango

RainbowMango:
got it.
/assign

pkg/controllers/binding/binding_controller.go
RainbowMango: OK.

RainbowMango (approved):

RainbowMango:
/lgtm
/hold
have you tested again after you fixed the comments? I'm asking as the E2E test hasn't covered this feature yet.
Feel free to cancel the hold once you confirmed.

@Garrybest 
Just a notice to you, the `Job` related functionality and behavior just work as before.

iawia002:
> have you tested again after you fixed the comments? I'm asking as the E2E test hasn't covered this feature yet.

Will test the deployment case and workload case again.

iawia002:
> > have you tested again after you fixed the comments? I'm asking as the E2E test hasn't covered this feature yet.
> 
> Will test the deployment case and workload case again.

I did find a small problem, the original code panic in case of webhook error because I refer to a nil object, now it has been fixed, others work as expected:

```diff
--- a/pkg/controllers/binding/common.go
+++ b/pkg/controllers/binding/common.go
@@ -113,7 +113,7 @@ func ensureWork(
                        clonedWorkload, err = resourceInterpreter.ReviseReplica(clonedWorkload, desireReplicaInfos[targetCluster.Name])
                        if err != nil {
                                klog.Errorf(“failed to revise replica for %s/%s/%s in cluster %s, err is: %v”,
-                                       clonedWorkload.GetKind(), clonedWorkload.GetNamespace(), clonedWorkload.GetName(), targetCluster.Name, err)
+                                       workload.GetKind(), workload.GetNamespace(), workload.GetName(), targetCluster.Name, err)
                                return err
                        }
```

PATL @RainbowMango

RainbowMango:
Good Job @iawia002

PR #1152, opened 2021-12-23
add validation for cluster api and remove validating webhook

carlory:
**What type of PR is this?**

/kind feature

```release-note
add validation for cluster API
```


XiShanYongYe-Chang:
Hi @carlory , thanks for your contribution. :)

RainbowMango:
Thanks @carlory I'll review it ASAP.

pkg/registry/cluster/strategy.go
```diff
@@ -76,8 +77,8 @@ func (Strategy) PrepareForUpdate(ctx context.Context, obj, old runtime.Object) {
 
 // Validate returns an ErrorList with validation errors or nil.
 func (Strategy) Validate(ctx context.Context, obj runtime.Object) field.ErrorList {
-	// TODO: add validation for Cluster
-	return field.ErrorList{}
+	cluster := obj.(*clusterapis.Cluster)
```
RainbowMango, line 80: This obj will always be `clusterapis.Cluster`? not `clusterv1alpha1.Cluster`? I'm not sure about it.

carlory: @RainbowMango Yes, the versioned object is always converted into the internal object in the chain of apiserver hander.

https://github.com/karmada-io/karmada/blob/master/vendor/k8s.io/apiserver/pkg/endpoints/handlers/create.go#L120

```go
        decoder := scope.Serializer.DecoderToVersion(s.Serializer, scope.HubGroupVersion)
        original := r.New()
        // defaultGVK = {Group: cluster, Version: v1alpha1, Kind: Cluster}
        // originalGVK = {Group: cluster, Version: __internal, Kind: Cluster}
	obj, gvk, err := decoder.Decode(body, &defaultGVK, original)
        r.Create(ctx, name,obj, est.AdmissionToValidateObjectFunc(admit, admissionAttributes, scope), options)
```

RainbowMango:
Got it @carlory.

carlory:
add validation for impersonatorSecretRef
/cc @RainbowMango @XiShanYongYe-Chang

carlory:
/cc @RainbowMango @XiShanYongYe-Chang 

remove validating webhook

XiShanYongYe-Chang:
Hi @carlory,  I'll review it now.

XiShanYongYe-Chang:
Thanks, @carlory.  It looks good to me.

RainbowMango:
Not an objection. Just a question.

The validation functionality in `webhook` still works as expected, right?
If it is the case, I'd suggest doing this after the coming new release.

RainbowMango:
We should remove the configuration from `karmada init`:
https://github.com/karmada-io/karmada/blob/14487548dd9c4970dad003da7d253c25811f2515/pkg/karmadactl/cmdinit/karmada/webhook_configuration.go#L83-L98

cc @lonelyCZ @prodanlabs

lonelyCZ:
> We should remove the configuration from karmada init

Ok, I will do it.

lonelyCZ:
This has been removed in this pr. Thanks @carlory

carlory:
/cc @RainbowMango 

the validation in `pkg/apis/cluster/validation` is more complete than in `pkg/util/validation`. 

In my opinion, there is no need to keep two verification processes at the same time

pkg/apis/cluster/validation/validation.go
```diff
@@ -0,0 +1,172 @@
...
+		allErrors = append(allErrors, unversionedvalidation.ValidateLabelName(currTaint.Key, idxPath.Child("key"))...)
+		// validate the taint value
+		if errs := validation.IsValidLabelValue(currTaint.Value); len(errs) != 0 {
+			allErrors = append(allErrors, field.Invalid(idxPath.Child("value"), currTaint.Value, strings.Join(errs, ";")))
+		}
+		// validate the taint effect
+		allErrors = append(allErrors, validateTaintEffect(&currTaint.Effect, false, idxPath.Child("effect"))...)
+
+		// validate if taint is unique by <key, effect>
+		if len(uniqueTaints[currTaint.Effect]) > 0 && uniqueTaints[currTaint.Effect].Has(currTaint.Key) {
```
RainbowMango, line 137:
```suggestion
		if uniqueTaints[currTaint.Effect].Has(currTaint.Key) {
```
This will be enough, right? 
Don't know if `len(sets.String)` will work.

RainbowMango (commented):
Generally looks good to me.

1. I suggest just putting `validation.go` to `pkg/apis/cluster` directory as it is only for `internel` types.
2. One main effect for moving the validation from webhook to AA is if there is a bug, it'll block people registering cluster(They can't disable the validation anymore). So, I suggest more unit test for the validations.

pkg/apis/cluster/validation/validation.go
```diff
@@ -0,0 +1,172 @@
...
+
+	api "github.com/karmada-io/karmada/pkg/apis/cluster"
+	utilvalidation "github.com/karmada-io/karmada/pkg/util/validation"
+)
+
+// ValidateClusterName can be used to check whether the given cluster name is valid.
+// Prefix indicates this name will be used as part of generation, in which case
+// trailing dashes are allowed.
+func ValidateClusterName(name string, prefix bool) []string {
+	return utilvalidation.ValidateClusterName(name)
```
RainbowMango, line 22: Can we move `utilvalidation.ValidateClusterName` to here?

pkg/apis/cluster/validation/validation.go
carlory: @RainbowMango it is necessary to avoid panic

RainbowMango: Really? Do you mean if we invoke `Has()` method on an empty `sets.String{}` will panic?

carlory: > 1. I suggest just putting `validation.go` to `pkg/apis/cluster` directory as it is only for `internel` types.

@RainbowMango

In [kubernetes/kubernetes](https://github.com/kubernetes/kubernetes/tree/master/pkg/apis/apps) and [openshift/origin](https://github.com/openshift/origin/tree/v3.11.0/pkg/apps/apis/apps) repo, the validation files are placed in a separate directory. 

I recommend keeping the same code style as them, what do you think?

RainbowMango: > I recommend keeping the same code style as them, what do you think?

OK for me.

RainbowMango:
@carlory Please cc me again if it is ready for review. Thanks for doing this, appreciate it!!

pkg/apis/cluster/validation/validation.go
carlory: > Really? Do you mean if we invoke `Has()` method on an empty `sets.String{}` will panic?

It may not have been initialized when we invoke `Has()` method.

```go
// ValidateClusterTaints tests if given taints have valid data.
func ValidateClusterTaints(taints []corev1.Taint, fldPath *field.Path) field.ErrorList {
	allErrors := field.ErrorList{}

	uniqueTaints := map[corev1.TaintEffect]sets.String{}
	for i, currTaint := range taints {
             ...
		// validate if taint is unique by <key, effect>
		if len(uniqueTaints[currTaint.Effect]) > 0 && uniqueTaints[currTaint.Effect].Has(currTaint.Key) {
                      ....
		}

		// add taint to existingTaints for uniqueness check
		if len(uniqueTaints[currTaint.Effect]) == 0 {
			uniqueTaints[currTaint.Effect] = sets.String{}
		}
		uniqueTaints[currTaint.Effect].Insert(currTaint.Key)
	}
	return allErrors
}
```

carlory: @RainbowMango 

`ValidateClusterTaints`  is lifted from [kubernetes validateNodeTaints](https://github.com/kubernetes/kubernetes/blob/885961cda8875cce192700fe6a32ff57ba706106/pkg/apis/core/validation/validation.go#L5003)

RainbowMango: But, I'm still interest if the `len(uniqueTaints[currTaint.Effect]) > 0` could be removed, though. 
Because, the `sets` has been initilized by `uniqueTaints := map[corev1.TaintEffect]sets.String{}`.

XiShanYongYe-Chang (commented):
LGTM, can you squash some commits.

carlory:
@XiShanYongYe-Chang done.

RainbowMango (approved):
/approve
Leave LGTM to @XiShanYongYe-Chang

PR #1154, opened 2021-12-23
fix division bugs when pod request zero scalar resource

Garrybest:

**What type of PR is this?**
/kind bug

If a pod request zero scalar resource, like `alpha.kubernetes.io/nvidia-gpu: '0'`, but the node does not have a resource called `alpha.kubernetes.io/nvidia-gpu`, scheduler-estimator will return 0 max replicas. It does not match what kubernetes does. In kubernetes, the zero scalar resource would be ignored even the node does not have this type of resource.

Fixes #

```release-note
NONE
```


RainbowMango (commented):
/lgtm
/approve

PR #1162, opened 2021-12-24
Implement the `detector` to watch the event of the `overridePolicy`

BDXGD:
Thanks for the help of @dddddai 

**What type of PR is this?**

/kind bug

Consider this situation: when the `deployment` and `propagation` are successfully created, it is found that there is no dependent `override`. After the `override` is created, the user finds that `rb` is still not generated, which will cause confusion. Finally found that because of too many attempts, deployment has been dropped from the workqueue.

First, currently Karmada `detector` does not watch to `override`, which is inappropriate. **The detector should watch the change of override**. 

Secondly, when the `deployment` matches the `propagation`, when the `override` dependent on `propagation` has not been created yet , it is meaningless to continue the requeue of the object's key, because the creation of the `override` may not be so fast. We can directly put the object's key into the `waiting list` and wait for the `override` creation time to wake up. 

image

```release-note
NONE
```

### DIFF
--- a/pkg/detector/detector.go
+++ b/pkg/detector/detector.go
@@ -63,6 +63,11 @@ type ResourceDetector struct {
 	clusterPolicyReconcileWorker   util.AsyncWorker
 	clusterPropagationPolicyLister cache.GenericLister
 
+	// overrideReconcileWorker maintains a rate limited queue which used to store OverridePolicy's key and
+	// a reconcile function to consume the items in queue.
+	overrideReconcileWorker util.AsyncWorker
+	overridePolicyLister    cache.GenericLister
+
 	// bindingReconcileWorker maintains a rate limited queue which used to store ResourceBinding's key and
 	// a reconcile function to consume the items in queue.
 	bindingReconcileWorker util.AsyncWorker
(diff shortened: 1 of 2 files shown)


pkg/detector/detector.go
```diff
@@ -226,10 +245,11 @@ func (d *ResourceDetector) Reconcile(key util.QueueKey) error {
 		return err
 	}
 	if propagationPolicy != nil {
-		// return err when dependents not present, that we can retry at next reconcile.
+		// Add clusterWideKey to waiting list when dependents not present, waiting for override to be created.
 		if present, err := helper.IsDependentOverridesPresent(d.Client, propagationPolicy); err != nil || !present {
-			klog.Infof("Waiting for dependent overrides present for policy(%s/%s)", propagationPolicy.Namespace, propagationPolicy.Name)
-			return fmt.Errorf("waiting for dependent overrides")
+			d.AddWaiting(clusterWideKey)
+			klog.Infof("object's key has been added to waiting list")
```
dddddai, line 251: I think we should keep the original log here, so that we are able to know the reason

And we should separate `err` and `present` here, for `err!=nil` we just return the error

dddddai: I think it should do the same check for `ClusterPropagationPolicy`, but it didn't, I guess it's an oversight, right? @RainbowMango 

If yes, we should watch `ClusterPropagationPolicy` as well

dddddai:
We should also update https://github.com/karmada-io/karmada/blob/9725a92af65d3baf6a693f222fd3b0795d36661a/pkg/detector/detector.go#L1025

and https://github.com/karmada-io/karmada/blob/9725a92af65d3baf6a693f222fd3b0795d36661a/pkg/detector/detector.go#L1062

If not present, just return nil

RainbowMango:
> After the override is created, the user finds that rb is still not generated, which will cause confusion. Finally found that because of too many attempts, deployment has been dropped from the workqueue.

I don't get why the deployment will be dropped from the work queue. 

Cloud you please reproduce it if you are still interested in this issue?

Given this topic was inactive for a long time, I'm going to close it for now, but feel free to reopen it if you think it's still needed.

PR #978, opened 2021-11-17
Logic patch for dividing remaining replicas to clusters

BDXGD:
**What type of PR is this?**

/kind cleanup

**What this PR does / why we need it**:

There are a lot of code logic to be optimized, the following is an example:

1. This makes the program not run the way as we want 

Change the location of the code block will solve the problem 

2. Furthermore, the running time of the program written in this way will depend on the value of `remainReplicas`. But if `remainReplicas` is much larger than the length of `desireReplicaInfos`, the first method runs  obviously not as fast as the second method.

First method:

Second method:

3. So we can choose the faster method based on the size relationship between the value of `remainReplicas` and the length of `desireReplicaInfos`:

**Which issue(s) this PR fixes**:

A part of issue #972

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #1168, opened 2021-12-25
A logic patch and optimization 

BDXGD:
**What type of PR is this?**

/kind cleanup

The same as #978.

### DIFF
--- a/pkg/util/binding.go
+++ b/pkg/util/binding.go
@@ -76,16 +76,33 @@ func DivideReplicasByTargetCluster(clusters []workv1alpha2.TargetCluster, sum in
 		}
 		allocatedReplicas += res[i].Replicas
 	}
-	if remainReplicas := sum - allocatedReplicas; remainReplicas > 0 {
-		for i := 0; remainReplicas > 0; i++ {
-			if i == len(res) {
-				i = 0
-			}
+
+	divideRemainingReplicas(int(sum-allocatedReplicas), res)
+
+	return res
+}
+
+// divideRemainingReplicas divide remaining Replicas to clusters
+func divideRemainingReplicas(remainingReplicas int, res []workv1alpha2.TargetCluster) {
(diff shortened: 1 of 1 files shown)


RainbowMango:
Sorry for letting this sit so long.

@XiShanYongYe-Chang Please help look at it.

pkg/util/binding.go
```diff
@@ -76,16 +76,33 @@ func DivideReplicasByTargetCluster(clusters []workv1alpha2.TargetCluster, sum in
...
+}
+
+// divideRemainingReplicas divide remaining Replicas to clusters
+func divideRemainingReplicas(remainingReplicas int, res []workv1alpha2.TargetCluster) {
+	if remainingReplicas <= 0 {
+		return
+	}
+
+	clusterSize := len(res)
+	if remainingReplicas < clusterSize {
```
XiShanYongYe-Chang, line 92: When the value of `remainingReplicas` is several times the value of `clusterSize`, some calculations can be omitted.

Maybe we can remove judgment. If match the case `remainingReplicas < clusterSize`, the `avg` value is 0.

pkg/util/binding.go
```diff
@@ -76,16 +76,33 @@ func DivideReplicasByTargetCluster(clusters []workv1alpha2.TargetCluster, sum in
...
+		for i := 0; i < remainingReplicas; i++ {
 			res[i].Replicas++
-			remainReplicas--
+		}
+	} else {
+		avg, residue := remainingReplicas/clusterSize, remainingReplicas%clusterSize
```
XiShanYongYe-Chang, line 97: We can convert the int type to int32 here, so as not to convert one by one during the loop.

XiShanYongYe-Chang:
kindly ping @BDXGD

RainbowMango:
/close
Feel free to re-open if you think this is still needed.

Issue #1136, opened 2021-12-21
Run karmada-interpreter-webhook-example in CI


RainbowMango:
**What would you like to be added**:
The [karmada-interpreter-webhook-example](https://github.com/karmada-io/karmada/tree/master/examples) was used to give an example about how to implement webhooks, and on the other hand, we also want it running in CI testing.

But no ideas about how to deploy it in our E2E testing yet.

**Why is this needed**:
We have implemented `InterpretReplica` and `Retain` [operations ](https://github.com/karmada-io/karmada/blob/98390f5615e0263da00a84b16588319752b893dc/pkg/apis/config/v1alpha1/resourceinterpreterwebhook_types.go#L71-L108)at v0.10 release, but the testing still missing for now, without the E2E it will slow down the iterate frequency.

lonelyCZ:
I am just interested in `interpreter` workflow, I would like to do it.

/assign

lonelyCZ:
I consider the test processes is below:

1. In `hack/run-e2e.sh`, creating `work` crd in `karmada-apiserver` and all member clusters, deploying `karmada-interpreter-webhook-example` in host cluster.
2. In `test/e2e/resourceinterpreter_test.go`, testing `InterpretReplica`, `Retain` and `ReviseReplica` by distributing `workload` CR. And adding tests case if there will be new interpret functions.

How do you think? @RainbowMango

RainbowMango:
I haven't really thought about it.

Things what we have to do for the environment might include:
- Deploy the [Workload CRD](https://github.com/karmada-io/karmada/blob/master/examples/customresourceinterpreter/apis/workload.example.io_workloads.yaml) to karmada-apiserver and propagate it to all member clusters(by a ClusterPropagationPolicy).
- Deploy the [karmada-interpreter-webhook-example](https://github.com/karmada-io/karmada/blob/master/examples/customresourceinterpreter/karmada-interpreter-webhook-example.yaml) to host cluster
- Deploy the [webhook-configuration](https://github.com/karmada-io/karmada/blob/master/examples/customresourceinterpreter/webhook-configuration.yaml) to karmada-apiserver

Things what we have to do for the test might include:
- Create the `Workload CR` and propagate it to member clusters
- For the `retain`, we might have to update the `.spec.paused` fields and expect the fielded can be `retained`.
- For the `InterpretReplica`, we might have to check if the replicas resource requirement be `interpreted` correctly? (Not sure).
[...]

PR #1172, opened 2021-12-27
Add e2e test for the ReviseReplica hook

iawia002:

**What type of PR is this?**

/kind feature

Add e2e test case for #1149

Fixes #

```release-note
NONE
```

### DIFF
--- a/test/e2e/resourceinterpreter_test.go
+++ b/test/e2e/resourceinterpreter_test.go
@@ -10,7 +10,9 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/rand"
 	"k8s.io/klog/v2"
+	"k8s.io/utils/pointer"
 
+	workloadv1alpha1 "github.com/karmada-io/karmada/examples/customresourceinterpreter/apis/workload/v1alpha1"
 	policyv1alpha1 "github.com/karmada-io/karmada/pkg/apis/policy/v1alpha1"
 	"github.com/karmada-io/karmada/pkg/util/names"
 	"github.com/karmada-io/karmada/test/e2e/framework"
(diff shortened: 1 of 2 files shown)


RainbowMango:
Do you think the `karmada-interpreter-webhook-example` should be deployed separately as described at #1136 ? @lonelyCZ is working on it.

lonelyCZ:
Thanks @iawia002 ,
The general idea is this, but it isn't suggested to create directly resource about `workload example`  in `local-up-karmada.sh`, `workload example` is only belong to e2e. I'm working on it that is expected to be finished today.

iawia002:
OK, I can wait for you and update the PR (just keep my e2e case part).

test/e2e/resourceinterpreter_test.go
```diff
@@ -121,4 +123,74 @@ var _ = ginkgo.Describe("Resource interpreter webhook testing", func() {
...
+								ClusterNames: []string{"member2"},
+							},
+							Weight: 1,
+						},
+						{
+							TargetCluster: policyv1alpha1.ClusterAffinity{
+								ClusterNames: []string{"member3"},
+							},
+							Weight: 1,
+						},
```
XiShanYongYe-Chang, line 167: How about calculating the replicas like this:

https://github.com/karmada-io/karmada/blob/14487548dd9c4970dad003da7d253c25811f2515/test/e2e/scheduling_test.go#L535-L546

In this way, we can make the logic is independent of the specific cluster name.

I also plan to gradually change other E2E parts that involve specific cluster names to those that are not related to cluster names.

test/e2e/resourceinterpreter_test.go
```diff
@@ -121,4 +123,74 @@ var _ = ginkgo.Describe("Resource interpreter webhook testing", func() {
 			framework.RemovePropagationPolicy(karmadaClient, policy.Namespace, policy.Name)
 		})
 	})
+
+	ginkgo.Context("Workload resource ReviseReplica hook testing", func() {
```
XiShanYongYe-Chang, line 127: How about rename this context to `InterpreterOperation ReviseReplica testing`. That way, it's consistent with the above.

XiShanYongYe-Chang (commented):
Thanks!

test/e2e/resourceinterpreter_test.go
iawia002: Great advice, I was wondering how to avoid using specific cluster names myself, updated, PTAL

test/e2e/resourceinterpreter_test.go
iawia002: Updated.

test/e2e/resourceinterpreter_test.go
XiShanYongYe-Chang: I don't have a specific idea yet. It may be determined based on the cluster type or traverse the existing cluster.
If you have a good idea, we can discuss it :)

iawia002: For my case now, I think the way in the scheduling test is just fine.

XiShanYongYe-Chang:
Uhum... Something wrong.

iawia002:
> Uhum... Something wrong.

Yeah, I'm testing it locally.

XiShanYongYe-Chang:
The `framework.ClusterNames()` will get an empty list outside the `ginkgo.It` block, so the sum weight will be 0.

We need to move the calculate of sum weight into  the `ginkgo.It` block.

XiShanYongYe-Chang:
The reason why the `framework.ClusterNames(https://github.com/karmada-io/karmada/blob/14487548dd9c4970dad003da7d253c25811f2515/test/e2e/framework/cluster.go#L32)` will get an empty, I'm guessing the value of [clusterNames]() is determined at Context init-time, and it hasn't been processed yet, so it's going to be nil.

When running `ginkgo.It`, the `clusterNames` will be calculated, so it will be ok.

iawia002:
OK, I have updated the code, PTAL @XiShanYongYe-Chang

PR #1178, opened 2021-12-28
add filter for binding informer in karmada scheduler

mrlihanbo:

**What type of PR is this?**
/kind cleanup

I found that the resource binding without reference propagation policy label will be scheduled to all member clusters. See [here](https://github.com/karmada-io/karmada/blob/24f586062e0cd7c9d8e6911e52ce399106f489aa/pkg/scheduler/scheduler.go#L338)
So I add a filter in binding informer. The resource binding without reference propagation policy label will be dropped.

Fixes #

"NONE"

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -139,18 +139,24 @@ func NewScheduler(dynamicClient dynamic.Interface, karmadaClient karmadaclientse
 
 	metrics.Register()
 
-	bindingInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
-		AddFunc:    sched.onResourceBindingAdd,
-		UpdateFunc: sched.onResourceBindingUpdate,
+	bindingInformer.AddEventHandler(cache.FilteringResourceEventHandler{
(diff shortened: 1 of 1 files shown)


XiShanYongYe-Chang:
Just a question: under what circumstances would `resource binding without reference propagation policy label` occur?

mrlihanbo:
> Just a question: under what circumstances would `resource binding without reference propagation policy label` occur?

Binding created due to auto propagete dependencies or created by users.

qianjun1993:
I think Resource binding without reference propagation policy label rarely appears, So maybe It will be better to add a interrupt here https://github.com/karmada-io/karmada/blob/24f586062e0cd7c9d8e6911e52ce399106f489aa/pkg/scheduler/scheduler.go#L343 ？

mrlihanbo:
> I think Resource binding without reference propagation policy label rarely appears, So maybe It will be better to add a interrupt here？
> 
> https://github.com/karmada-io/karmada/blob/24f586062e0cd7c9d8e6911e52ce399106f489aa/pkg/scheduler/scheduler.go#L343

The pr works with #1182 and the incoming auto propagate deps feature.

PR #1199, opened 2021-12-31
bugfix: resource binding is not created when creating resources and propagation policies at the same time

### LINKED ISSUE #1195: Resourcebinding NotFound.
#### Which jobs are flaking:

https://github.com/karmada-io/karmada/runs/4663848781?check_suite_focus=true

```log
propagation with label and group constraints testing Deployment propagation testing 
  deployment propagation with label and group constraints testing
  /home/runner/work/karmada/karmada/test/e2e/scheduling_test.go:70
STEP: Creating PropagationPolicy(karmadatest-mv8/deploy-z8w)
STEP: Creating Deployment(karmadatest-mv8/deploy-z8w)
STEP: collect the target clusters in resource binding

• Failure [300.127 seconds]
propagation with label and group constraints testing
/home/runner/work/karmada/karmada/test/e2e/scheduling_test.go:36
  Deployment propagation testing
  /home/runner/work/karmada/karmada/test/e2e/scheduling_test.go:37
    deployment propagation with label and group constraints testing [It]
    /home/runner/work/karmada/karmada/test/e2e/scheduling_test.go:70

    Timed out after 300.001s.
    Error: Assertion in callback at /home/runner/work/karmada/karmada/test/e2e/framework/deployment.go:111 failed:
    Unexpected error:
        <*errors.StatusError | 0xc0005fc3c0>: {
            ErrStatus: {
                TypeMeta: {Kind: "", APIVersion: ""},
                ListMeta: {
                    SelfLink: "",
                    ResourceVersion: "",
                    Continue: "",
                    RemainingItemCount: nil,
                },
                Status: "Failure",
                Message: "resourcebindings.work.karmada.io \"deploy-z8w-deployment\" not found",
                Reason: "NotFound",
                Details: {
                    Name: "deploy-z8w-deployment",
                    Group: "work.karmada.io",
                    Kind: "resourcebindings",
                    UID: "",
                    Causes: nil,
                    RetryAfterSeconds: 0,
                },
                Code: 404,
            },
        }
        resourcebindings.work.karmada.io "deploy-z8w-deployment" not found
    occurred

    /home/runner/work/karmada/karmada/test/e2e/framework/deployment.go:118
```

https://github.com/karmada-io/karmada/blob/6f4b65a99f2617f467de47509ee8d6a44b11c3be/test/e2e/scheduling_test.go#L70

#### Anything else we need to know:

Key logs:

```log
2021-12-30T03:49:20.818645777Z stderr F I1230 03:49:20.818201       1 detector.go:206] Reconciling object: apps/v1, kind=Deployment, karmadatest-mv8/deploy-z8w
2021-12-30T03:49:20.820743343Z stderr F I1230 03:49:20.819080       1 detector.go:339] attempts to match policy for resource(apps/v1, kind=Deployment, karmadatest-mv8/deploy-z8w)
2021-12-30T03:49:20.821621071Z stderr F I1230 03:49:20.821444       1 detector.go:346] no propagationpolicy find in namespace(karmadatest-mv8).
2021-12-30T03:49:20.841408099Z stderr F I1230 03:49:20.841283       1 detector.go:751] Create PropagationPolicy(policy.karmada.io/v1alpha1, kind=PropagationPolicy, karmadatest-mv8/deploy-z8w)
2021-12-30T03:49:20.8414287Z stderr F I1230 03:49:20.841330       1 detector.go:793] PropagationPolicy(karmadatest-mv8/deploy-z8w) has been added.
2021-12-30T03:49:20.842470833Z stderr F I1230 03:49:20.842353       1 detector.go:377] attempts to match cluster policy for resource(apps/v1, kind=Deployment, karmadatest-mv8/deploy-z8w)
2021-12-30T03:49:20.842894446Z stderr F I1230 03:49:20.842766       1 detector.go:384] no propagationpolicy find.
2021-12-30T03:49:20.861188527Z stderr F I1230 03:49:20.857122       1 detector.go:946] Matched 0 resources by policy(karmadatest-mv8/deploy-z8w)
2021-12-30T03:49:20.87985912Z stderr F I1230 03:49:20.878372       1 detector.go:707] Add object(apps/v1, kind=Deployment, karmadatest-mv8/deploy-z8w) to waiting list, length of list is: 194
```

dddddai:
**What type of PR is this?**
/kind bug
/kind flake

The root cause is as the follow process:
image

We must **add the resource to waiting list before looking up the matched propagation policy**, so that the matched resource can always be found when reconciling propagation policies

Fixes #1195

```release-note
NONE
```

### DIFF
--- a/pkg/detector/detector.go
+++ b/pkg/detector/detector.go
@@ -54,14 +54,8 @@ type ResourceDetector struct {
 	// ResourceInterpreter knows the details of resource structure.
 	ResourceInterpreter resourceinterpreter.ResourceInterpreter
 	EventRecorder       record.EventRecorder
-	// policyReconcileWorker maintains a rate limited queue which used to store PropagationPolicy's key and
-	// a reconcile function to consume the items in queue.
-	policyReconcileWorker   util.AsyncWorker
-	propagationPolicyLister cache.GenericLister
 
-	// clusterPolicyReconcileWorker maintains a rate limited queue which used to store ClusterPropagationPolicy's key and
-	// a reconcile function to consume the items in queue.
-	clusterPolicyReconcileWorker   util.AsyncWorker
+	propagationPolicyLister        cache.GenericLister
 	clusterPropagationPolicyLister cache.GenericLister
 
 	// bindingReconcileWorker maintains a rate limited queue which used to store ResourceBinding's key and
@@ -85,30 +79,17 @@ func (d *ResourceDetector) Start(ctx context.Context) error {
 	d.waitingObjects = make(map[keys.ClusterWideKey]struct{})
 	d.stopCh = ctx.Done()
 
-	// setup policy reconcile worker
-	d.policyReconcileWorker = util.NewAsyncWorker("propagationPolicy reconciler", ClusterWideKeyFunc, d.ReconcilePropagationPolicy)
-	d.policyReconcileWorker.Run(1, d.stopCh)
-	d.clusterPolicyReconcileWorker = util.NewAsyncWorker("clusterPropagationPolicy reconciler", ClusterWideKeyFunc, d.ReconcileClusterPropagationPolicy)
-	d.clusterPolicyReconcileWorker.Run(1, d.stopCh)
-
-	// watch and enqueue PropagationPolicy changes.
 	propagationPolicyGVR := schema.GroupVersionResource{
 		Group:    policyv1alpha1.GroupVersion.Group,
 		Version:  policyv1alpha1.GroupVersion.Version,
 		Resource: "propagationpolicies",
 	}
-	policyHandler := informermanager.NewHandlerOnEvents(d.OnPropagationPolicyAdd, d.OnPropagationPolicyUpdate, d.OnPropagationPolicyDelete)
-	d.InformerManager.ForResource(propagationPolicyGVR, policyHandler)
-	d.propagationPolicyLister = d.InformerManager.Lister(propagationPolicyGVR)
-
-	// watch and enqueue ClusterPropagationPolicy changes.
 	clusterPropagationPolicyGVR := schema.GroupVersionResource{
 		Group:    policyv1alpha1.GroupVersion.Group,
 		Version:  policyv1alpha1.GroupVersion.Version,
 		Resource: "clusterpropagationpolicies",
 	}
-	clusterPolicyHandler := informermanager.NewHandlerOnEvents(d.OnClusterPropagationPolicyAdd, d.OnClusterPropagationPolicyUpdate, d.OnClusterPropagationPolicyDelete)
-	d.InformerManager.ForResource(clusterPropagationPolicyGVR, clusterPolicyHandler)
+	d.propagationPolicyLister = d.InformerManager.Lister(propagationPolicyGVR)
 	d.clusterPropagationPolicyLister = d.InformerManager.Lister(clusterPropagationPolicyGVR)
 
 	// setup binding reconcile worker
@@ -134,7 +115,9 @@ func (d *ResourceDetector) Start(ctx context.Context) error {
 	clusterBindingHandler := informermanager.NewHandlerOnEvents(d.OnClusterResourceBindingAdd, d.OnClusterResourceBindingUpdate, nil)
 	d.InformerManager.ForResource(clusterResourceBindingGVR, clusterBindingHandler)
 
-	d.EventHandler = informermanager.NewFilteringHandlerOnAllEvents(d.EventFilter, d.OnAdd, d.OnUpdate, d.OnDelete)
+	d.EventHandler = informermanager.NewHandlerOnEvents(d.OnAdd, d.OnUpdate, d.OnDelete)
+	d.InformerManager.ForResource(propagationPolicyGVR, d.EventHandler)
+	d.InformerManager.ForResource(clusterPropagationPolicyGVR, d.EventHandler)
 	d.Processor = util.NewAsyncWorker("resource detector", ClusterWideKeyFunc, d.Reconcile)
 	d.Processor.Run(1, d.stopCh)
 	go d.discoverResources(30 * time.Second)
@@ -145,8 +128,10 @@ func (d *ResourceDetector) Start(ctx context.Context) error {
 }
 
 // Check if our ResourceDetector implements necessary interfaces
-var _ manager.Runnable = &ResourceDetector{}
-var _ manager.LeaderElectionRunnable = &ResourceDetector{}
+var (
+	_ manager.Runnable               = &ResourceDetector{}
+	_ manager.LeaderElectionRunnable = &ResourceDetector{}
+)
 
 func (d *ResourceDetector) discoverResources(period time.Duration) {
 	wait.Until(func() {
@@ -204,8 +189,20 @@ func (d *ResourceDetector) Reconcile(key util.QueueKey) error {
 		klog.Error("invalid key")
 		return fmt.Errorf("invalid key")
 	}
-	klog.Infof("Reconciling object: %s", clusterWideKey)
 
+	if d.SkippedFromPropagating(clusterWideKey) {
+		if clusterWideKey.Group == policyv1alpha1.GroupName {
+			switch clusterWideKey.Kind {
+			case "PropagationPolicy":
+				return d.ReconcilePropagationPolicy(key)
+			case "ClusterPropagationPolicy":
+				return d.ReconcileClusterPropagationPolicy(key)
+			}
+		}
+		return nil
+	}
+
+	klog.Infof("Reconciling object: %s", clusterWideKey)
 	object, err := d.GetUnstructuredObject(clusterWideKey)
 	if err != nil {
 		if apierrors.IsNotFound(err) {
@@ -253,7 +250,7 @@ func (d *ResourceDetector) Reconcile(key util.QueueKey) error {
 	return nil
 }
 
-// EventFilter tells if an object should be take care of.
+// SkippedFromPropagating tells if an object should be propagated.
 //
 // All objects under Kubernetes reserved namespace should be ignored:
 // - kube-system
@@ -266,6 +263,8 @@ func (d *ResourceDetector) Reconcile(key util.QueueKey) error {
 // All objects which API group defined by Karmada should be ignored:
 // - cluster.karmada.io
 // - policy.karmada.io
+// - work.karmada.io
+// - config.karmada.io
 //
 // The api objects listed above will be ignored by default, as we don't want users to manually input the things
 // they don't care when trying to skip something else.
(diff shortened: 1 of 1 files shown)


pkg/detector/detector.go
```diff
@@ -231,7 +235,7 @@ func (d *ResourceDetector) Reconcile(key util.QueueKey) error {
 			klog.Infof("Waiting for dependent overrides present for policy(%s/%s)", propagationPolicy.Namespace, propagationPolicy.Name)
 			return fmt.Errorf("waiting for dependent overrides")
 		}
-
+		d.RemoveWaiting(clusterWideKey)
```
XiShanYongYe-Chang, line 238: What if `AppliPolicy` failed?

pkg/detector/detector.go
```diff
@@ -955,7 +958,6 @@ func (d *ResourceDetector) HandlePropagationPolicyCreation(policy *policyv1alpha
 	}
 
 	for _, key := range matchedKeys {
-		d.RemoveWaiting(key)
```
XiShanYongYe-Chang, line 958: I didn't understand why this line was deleted.

pkg/detector/detector.go
dddddai: Then it would requeue and reconcile again, right?

pkg/detector/detector.go
dddddai: Because the waiting obj would be removed when reconciling, as above shows

pkg/detector/detector.go
XiShanYongYe-Chang: get it.

pkg/detector/detector.go
XiShanYongYe-Chang: Reconciling will add the key again, keeping it here feels fine and easier to understand.

dddddai: > Reconciling will add the key again, keeping it here feels fine and easier to understand.

https://github.com/karmada-io/karmada/blob/7224234a742924d78b25cbcc46eaeaccd442fbb8/pkg/detector/detector.go#L74

The `waitingObject` is a map, so it won't matter if an obj is added multiple times

If you mean it's for better understanding, I will revert it though

XiShanYongYe-Chang:
image

The current processing logic can cover the scenario.

dddddai:
@XiShanYongYe-Chang Sorry... I didn't get the meaning of the picture

XiShanYongYe-Chang:
> @XiShanYongYe-Chang Sorry... I didn't get the meaning of the picture

I didn't draw it well. I'm just trying to explain that the current way can overwrite the previous error.

RainbowMango:
Good job @dddddai .
This patch definitely can solve the corner case. The side-effect I can tell is the `resource template` might be got in and out the `waitting list` many times.

Can you run a test to observe the logs:
https://github.com/karmada-io/karmada/blob/8b3aefdb6f5d5ee0016d96d4353a9d822491e358/pkg/detector/detector.go#L707

Let's see how many times a resource template would be put in `waitting list` in a propagation process.

dddddai:
Thanks @RainbowMango, here's the log
If creating resource before the propagation policy:
```sh
I0105 10:29:38.611012       1 detector.go:710] Add object(apps/v1, kind=Deployment, default/nginx) to waiting list, length of list is: 160
I0105 10:29:46.133486       1 detector.go:710] Add object(apps/v1, kind=Deployment, default/nginx) to waiting list, length of list is: 160
I0105 10:29:46.177437       1 detector.go:710] Add object(apps/v1, kind=Deployment, default/nginx) to waiting list, length of list is: 160
I0105 10:29:46.208280       1 detector.go:710] Add object(apps/v1, kind=Deployment, default/nginx) to waiting list, length of list is: 160
I0105 10:29:46.618257       1 detector.go:710] Add object(apps/v1, kind=Deployment, default/nginx) to waiting list, length of list is: 160
I0105 10:29:46.633569       1 detector.go:710] Add object(apps/v1, kind=Deployment, default/nginx) to waiting list, length of list is: 160
I0105 10:29:46.733210       1 detector.go:710] Add object(apps/v1, kind=Deployment, default/nginx) to waiting list, length of list is: 160
I0105 10:29:48.428089       1 detector.go:710] Add object(apps/v1, kind=Deployment, default/nginx) to waiting list, length of list is: 160
```
The resource was added to waiting list **8** times

If creating the propagation policy before resource:
```sh
I0105 10:32:25.143445       1 detector.go:710] Add object(apps/v1, kind=Deployment, default/nginx) to waiting list, length of list is: 160
I0105 10:32:25.165382       1 detector.go:710] Add object(apps/v1, kind=Deployment, default/nginx) to waiting list, length of list is: 160
I0105 10:32:25.202162       1 detector.go:710] Add object(apps/v1, kind=Deployment, default/nginx) to waiting list, length of list is: 160
I0105 10:32:25.576123       1 detector.go:710] Add object(apps/v1, kind=Deployment, default/nginx) to waiting list, length of list is: 160
I0105 10:32:25.816949       1 detector.go:710] Add object(apps/v1, kind=Deployment, default/nginx) to waiting list, length of list is: 160
I0105 10:32:25.930986       1 detector.go:710] Add object(apps/v1, kind=Deployment, default/nginx) to waiting list, length of list is: 160
I0105 10:32:26.822290       1 detector.go:710] Add object(apps/v1, kind=Deployment, default/nginx) to waiting list, length of list is: 160
```
The resource was added to waiting list **7** times

> The side-effect I can tell is the resource template might be got in and out the waitting list many times.

Yes, but this is the simplest way I could figure out to fix the bug, I'd appreciate if there is any better idea

RainbowMango:
I've no idea yet. Since it's the corner case, so let's do it slowly and correctly.

@iawia002 any comments?

iawia002:
> I've no idea yet. Since it's the corner case, so let's do it slowly and correctly.
> 
> @iawia002 any comments?

Actually I have no better idea, maybe we can make a judgment before adding keys to the waiting list:

```diff
+	if _, ok := d.waitingObjects[objectKey]; ok {
+		return
+	}
	d.waitingObjects[objectKey] = struct{}{}
	klog.V(1).Infof("Add object(%s) to waiting list, length of list is: %d", objectKey.String(), len(d.waitingObjects))
```

dddddai:
@iawia002 , Thanks for your suggestion, but in most cases it's unlikely for objects to stay in waiting list before adding, and the judgment can make `AddWaiting` slower

XiShanYongYe-Chang (commented):
What's going on with this pull request?

RainbowMango:
Looking forward to a better solution.
/priority important-soon

dddddai:
To be honest I don't see any better fixes, maybe we can merge it for quick fix(when we are gonna release 1.1) and update this once there's a better solution.

If [the logs](https://github.com/karmada-io/karmada/pull/1199#issuecomment-1005332649) are too noisy , we may change them from `info` to `debug`

RainbowMango:
I'm thinking if we can merge the two queues into one. But haven't time to look at it yet.

It's not an urgent bug, so we can hold for a while.  I think it will be included in v1.1.

RainbowMango:
This problem is simply that two routines(`resource template reconciler` and `PP/CPP reconciler`) depend on the same mutex(`waiting list`).

So, how about merging `PP/CPP reconciler` to `resource template reconciler`?  Thus, the `waiting list` will not be shared.
I mean the following reconcile could be removed.
https://github.com/karmada-io/karmada/blob/e3c8d3998a903a271a8a19168f669857e0eb9596/pkg/detector/detector.go#L57-L65

When reconciling there will be two routing path:
- For `PP/CPP`, routing to `ReconcilePropagationPolicy` or `ReconcileClusterPropagationPolicy`.
- For `resource template`, routing to `Reconcile`

There will be a little bit complex when dealing with the [EventFilter](https://github.com/karmada-io/karmada/blob/e3c8d3998a903a271a8a19168f669857e0eb9596/pkg/detector/detector.go#L277:28).

dddddai:
> So, how about merging `PP/CPP reconciler` to `resource template reconciler`? Thus, the `waiting list` will not be shared.

Do you mean reconciling both PP/CPPs and resource templates **in one goroutine**?

XiShanYongYe-Chang:
> Do you mean reconciling both PP/CPPs and resource templates **in one goroutine**?

@dddddai I think it should be. Can you try to modify it? Today has two cases failed on the master branch for this reason.

RainbowMango:
> Do you mean reconciling both PP/CPPs and resource templates in one goroutine?

Yes, that's the idea.

pkg/detector/detector.go
```diff
@@ -323,7 +324,15 @@ func (d *ResourceDetector) OnAdd(obj interface{}) {
 
 // OnUpdate handles object update event and push the object to queue.
 func (d *ResourceDetector) OnUpdate(oldObj, newObj interface{}) {
-	d.OnAdd(newObj)
+	switch newObj.(type) {
+	case *policyv1alpha1.PropagationPolicy:
+		// currently do nothing, since a policy's resource selector can not be updated.
+		return
+	case *policyv1alpha1.ClusterPropagationPolicy:
+		return
```
RainbowMango, line 332: If the update operation is not allowed, there is no chance to be here, right? 

In addition, it will block PP/CPP update on other fields(like `.spec.placement`).

pkg/detector/detector.go
```diff
@@ -286,6 +281,12 @@ func (d *ResourceDetector) EventFilter(obj interface{}) bool {
 		return false
 	}
 
+	if clusterWideKey.Group == policyv1alpha1.GroupName {
+		if clusterWideKey.Kind == "PropagationPolicy" || clusterWideKey.Kind == "ClusterPropagationPolicy" {
+			return true
+		}
+	}
```
RainbowMango, line 288: > There will be a little bit complex when dealing with the EventFilter.

This is what I am worried about. It seems like a `glue`, I believe it can solve the problem, but the code becomes hard to understand.

Let's think about how to deal with the `event handler` thing better.

pkg/detector/detector.go
dddddai: Oh yes I forgot the update is rejected by the webhook

pkg/detector/detector.go
dddddai: I see, how about [moving the event filter to `Reconcile`](https://github.com/karmada-io/karmada/pull/1199/commits/19694c3db4fd9b198615492d6b3c79458298ba6e#diff-fdb6430a46e82ace9f449b971619671eac711f7a82d6b862c130236a65cd5101R203)?

pkg/detector/detector.go
```diff
@@ -206,6 +191,19 @@ func (d *ResourceDetector) Reconcile(key util.QueueKey) error {
 	}
 	klog.Infof("Reconciling object: %s", clusterWideKey)
```
RainbowMango, line 192: Can we move the log down? I don't want the ignored resources present in the logs by default.

pkg/detector/detector.go
```diff
@@ -206,6 +191,19 @@ func (d *ResourceDetector) Reconcile(key util.QueueKey) error {
...
+			return d.ReconcilePropagationPolicy(key)
+		case "ClusterPropagationPolicy":
+			return d.ReconcileClusterPropagationPolicy(key)
+		}
+	}
+
+	if !d.EventFilter(clusterWideKey) {
+		return nil
+	}
+
```
RainbowMango, line 206: The `if clusterWideKey.Group == policyv1alpha1.GroupName` block better to moved into the `if !d.EventFilter(clusterWideKey)`, like:
```
	if !d.EventFilter(clusterWideKey) {
		if clusterWideKey.Group == policyv1alpha1.GroupName {
			switch clusterWideKey.Kind {
			case "PropagationPolicy":
				return d.ReconcilePropagationPolicy(key)
			case "ClusterPropagationPolicy":
				return d.ReconcileClusterPropagationPolicy(key)
			}
		}
		klog.V(4).Infof("Ignored object(%s) from reconciling.", clusterWideKey.String())
		return nil
	}
	klog.Infof("Reconciling object: %s", clusterWideKey)
```

pkg/detector/detector.go
```diff
@@ -274,18 +274,7 @@ func (d *ResourceDetector) Reconcile(key util.QueueKey) error {
...
-		return false
-	}
-
-	clusterWideKey, ok := key.(keys.ClusterWideKey)
-	if !ok {
-		klog.Errorf("Invalid key")
-		return false
-	}
-
+func (d *ResourceDetector) EventFilter(clusterWideKey keys.ClusterWideKey) bool {
```
RainbowMango, line 277: I guess the `EventFilter` should be renamed now, as it not handling the `event` anymore, but checks if the specified object should be prevented from propagating. So, how about `SkippedFromPropagating`? 

Remember to update the comments and take care of the function body.

RainbowMango (commented):
Pretty good now.

pkg/detector/detector.go
```diff
@@ -274,18 +273,7 @@ func (d *ResourceDetector) Reconcile(key util.QueueKey) error {
...
-		return false
-	}
-
-	clusterWideKey, ok := key.(keys.ClusterWideKey)
-	if !ok {
-		klog.Errorf("Invalid key")
-		return false
-	}
-
+func (d *ResourceDetector) SkippedFromPropagating(clusterWideKey keys.ClusterWideKey) bool {
```
RainbowMango, line 276: After renaming the function, the implementation should be updated as well, when a resource is skipped, it should return `true` now.

dddddai: Fixed, thanks for elaborating!

RainbowMango (approved):
/lgtm
/approve

Now we just started [one worker](https://github.com/karmada-io/karmada/blob/0b9a5d330d41f34db7bfb3192e7b9a72fe10bf2e/pkg/detector/detector.go#L139), this risk still exists if there is more than one worker.

No other better solution yet, and this issue happens several times in our CI, merging this patch first.

dddddai:
> this risk still exists if there is more than one worker

That's exactly what I'm worried about, so honestly I'd prefer the original solution

RainbowMango:
>  so honestly I'd lean towards the original solution

...

The original solution(throw into waiting list) does not 100% solve the problem but reduces the probability significantly.

dddddai:
Why? Could you please give an example?

Anyway I'm fine with both solutions

pigletfly:
> > so honestly I'd lean towards the original solution
> 
> ...
> 
> The original solution(throw into waiting list) does not 100% solve the problem but reduces the probability significantly.

we should start a seperate goroutine to consume waiting, if resource template and pp are created at the same time and miss the match(resource template is created and has no more event),then goroutine keeps find the match.

RainbowMango:
> we should start a seperate goroutine to consume waiting, if resource template and pp are created at the same time and miss the match(resource template is created and has no more event),then goroutine keeps find the match.

I'm also thinking of this approach.

PR #1224, opened 2022-01-07
add error message when wait file exist timeout

weilaaa:
**What type of PR is this?**
/kind cleanup

What: add error message when wait file exist timeout 

Why: call util::wait_file_exist() timeout but had no message return that causes confused.


RainbowMango:
LGTM
/assign @lfbear

PR #1225, opened 2022-01-08
fix bug of resource computing

Garrybest:

**What type of PR is this?**
/kind bug

1. If pod limits are specified, but requests are not, default requests to limits.
2. If Overhead is being utilized, add to the total requests for the pod.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/util/resource.go
+++ b/pkg/util/resource.go
@@ -253,6 +253,41 @@ func (r *Resource) LessEqual(rr *Resource) bool {
 	return true
 }
 
+// AddPodTemplateRequest add the effective request resource of a pod template to the origin resource.
+// If pod container limits are specified, but requests are not, default requests to limits.
+// The code logic is almost the same as kubernetes.
+// https://github.com/kubernetes/kubernetes/blob/f7cdbe2c96cc12101226686df9e9819b4b007c5c/pkg/apis/core/v1/defaults.go#L147-L181
+func (r *Resource) AddPodTemplateRequest(podSpec *corev1.PodSpec) *Resource {
+	// DeepCopy first because we may modify the Resources.Requests field.
+	podSpec = podSpec.DeepCopy()
+	for i := range podSpec.Containers {
+		// set requests to limits if requests are not specified, but limits are
+		if podSpec.Containers[i].Resources.Limits != nil {
+			if podSpec.Containers[i].Resources.Requests == nil {
+				podSpec.Containers[i].Resources.Requests = make(corev1.ResourceList)
+			}
+			for key, value := range podSpec.Containers[i].Resources.Limits {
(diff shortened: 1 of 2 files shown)


RainbowMango:
cc @Huone1
Could you please look at this PR?

RainbowMango (commented):
I wonder how and when the [SetDefaults_Pod](https://github.com/kubernetes/kubernetes/blob/f7cdbe2c96cc12101226686df9e9819b4b007c5c/pkg/apis/core/v1/defaults.go#L147-L181) executed? Does the `karmada-apiserver` also set the defaults for Pod?

pkg/util/resource.go
```diff
@@ -265,6 +300,11 @@ func (r *Resource) AddPodRequest(podSpec *corev1.PodSpec) *Resource {
 	for _, container := range podSpec.InitContainers {
 		r.SetMaxResource(container.Resources.Requests)
 	}
+	// If Overhead is being utilized, add to the total requests for the pod.
+	// We assume the EnablePodOverhead feature gate of member cluster is set (it is on by default since 1.18).
+	if podSpec.Overhead != nil {
+		r.Add(podSpec.Overhead)
+	}
```
RainbowMango, line 307: The field `.spec.overhead` was introduced by K/K from v1.16. 

Are there risks if a cluster(<v1.16) joined Karmada?

Garrybest:
> I wonder how and when the [SetDefaults_Pod](https://github.com/kubernetes/kubernetes/blob/f7cdbe2c96cc12101226686df9e9819b4b007c5c/pkg/apis/core/v1/defaults.go#L147-L181) executed? Does the `karmada-apiserver` also set the defaults for Pod?

The api-server mutating webhook will set the default values. See [here](https://github.com/kubernetes/kubernetes/blob/f7cdbe2c96cc12101226686df9e9819b4b007c5c/staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/mutating/dispatcher.go#L344-L345)

pkg/util/resource.go
Garrybest: I think it's ok. It may lead some resource difference larger than before. The estimator will give a more strict result of available replicas. However, actually the estimator itself is not 100% accurate.

huone1:
the overhead resource data  should be obtained through by RuntimeClass; the podtemplate shouldn't have the overhead info

RainbowMango:
Discussed this with @huone1 locally, I get his concern now.

Since the `karmada-apiserver` only set `overhead` fields for `Pod`, but not for other workloads like `Deployment`, `Job`, etc.
So, the `overhead` will be always empty for these workloads, which means we can't get precise resource requirements.

PR #1227, opened 2022-01-08
Fix stop cluster informer when a cluster is removed

pigletfly:

**What type of PR is this?**
/kind cleanup

Fixes #

```release-note
None

```

### DIFF
--- a/pkg/controllers/status/workstatus_controller.go
+++ b/pkg/controllers/status/workstatus_controller.go
@@ -432,8 +432,6 @@ func (c *WorkStatusController) getGVRsFromWork(work *workv1alpha1.Work) (map[sch
 // getSingleClusterManager gets singleClusterInformerManager with clusterName.
 // If manager is not exist, create it, otherwise gets it from map.
 func (c *WorkStatusController) getSingleClusterManager(cluster *clusterv1alpha1.Cluster) (informermanager.SingleClusterInformerManager, error) {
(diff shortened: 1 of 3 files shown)


pkg/controllers/status/workstatus_controller.go
```diff
@@ -432,8 +432,6 @@ func (c *WorkStatusController) getGVRsFromWork(work *workv1alpha1.Work) (map[sch
 // getSingleClusterManager gets singleClusterInformerManager with clusterName.
 // If manager is not exist, create it, otherwise gets it from map.
 func (c *WorkStatusController) getSingleClusterManager(cluster *clusterv1alpha1.Cluster) (informermanager.SingleClusterInformerManager, error) {
-	// TODO(chenxianpao): If cluster A is removed, then a new cluster that name also is A joins karmada,
-	//  the cache in informer manager should be updated.
```
RainbowMango, line 436: Seems the informer for `unregistered-cluster` will be cleaned up at [cluster status controller](https://github.com/karmada-io/karmada/blob/27f5bc55c1a150b929ac22337772d2ba7f78b9b0/pkg/controllers/status/cluster_status_controller.go#L95).

This is outdated comments I guess.

pigletfly: yes

RainbowMango: Maybe I don't think so. Why do you think that? 
They are sharing the same `informer manager`, stop once would be enough.

PR #1239, opened 2022-01-11
Fix replicas in values.yaml

pigletfly:
Signed-off-by: pigletfly wangbing.adam@gmail.com

**What type of PR is this?**
/kind cleanup

Fixes #

```release-note

```

### DIFF
--- a/charts/values.yaml
+++ b/charts/values.yaml
@@ -82,7 +82,7 @@ scheduler:
   labels:
     app: karmada-scheduler
   ## @param scheduler.replicaCount target replicas
-  replicaCount: 1
+  replicaCount: 3
   ## @param scheduler.podAnnotations
   podAnnotations: { }
   ## @param scheduler.podLabels
@@ -121,7 +121,7 @@ webhook:
   labels:
     app: karmada-webhook
   ## @param webhook.replicaCount target replicas
-  replicaCount: 1
+  replicaCount: 3
   ## @param webhook.podAnnotations
   podAnnotations: { }
   ## @param webhook.podLabels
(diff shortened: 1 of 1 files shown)


RainbowMango:
Can you explain why we need to enlarge the default replicas?

pigletfly:
for high availability

RainbowMango:
The default value should cover most use cases I think.
@algebra2k @prodanlabs How do you think?

algebra2k:
> The default value should cover most use cases I think. @algebra2k @prodanlabs How do you think?

I think very good, three replicas  is more suitable for production environment.

RainbowMango:
> I think very good, three replicas is more suitable for production environment.

I agree. Just a concern, if it will bring inconvenience to general cases?  Such as a user just want to try it on a very small cluster, like Kind which is usually one node by default.

pigletfly:
> 

we shouldn't provide a test helm charts values,the charts should be used for production.If users install in dev environment,they can modify the replicas.

RainbowMango:
> we shouldn't provide a test helm charts values

I didn't mean to provide a `test` charts. I mean default value should suit the basic production use case. (Not all users need HA right?)

> For the number of replicas to be changed to three, can we add some explanations in charts README of table?

+1. 
We can document it for people who want to enable HA.

prodanlabs:
> The default value should cover most use cases I think. @algebra2k @prodanlabs How do you think?

Yes, I think so too. The default is 1, we provide options for the user to choose.

pigletfly:
xref https://kubernetes.io/docs/setup/production-environment/

jrkeen:
> > we shouldn't provide a test helm charts values
> 
> I didn't mean to provide a `test` charts. I mean default value should suit the basic production use case. (Not all users need HA right?)
> 
> > For the number of replicas to be changed to three, can we add some explanations in charts README of table?
> 
> +1. We can document it for people who want to enable HA.

I think so.
Most project default replicas will be set to 1 so that to meet the minimum startup environment. 
xref https://github.com/bitnami/charts/tree/master/bitnami

PR #1248, opened 2022-01-12
bump go version to 1.17

pigletfly:

**What type of PR is this?**
/kind cleanup

Fixes #
follow up upstream go version

```release-note
NONE
```


RainbowMango:
Why do you think we need to update Go version?

We usually select Go version according to which k/k we depend on. E.g, for k/k@v1.22 requires Go 1.16, and for k/k@v1.23 requires Go 1.17.

Issue #1237, opened 2022-01-11
[Feature] ResourceInterpreter framework supports `AggregateStatus`


RainbowMango:
**What would you like to be added**:
Enable [AggregateStatus](https://github.com/karmada-io/karmada/blob/eb4b3e27bc9d080377b899ce251f49ce3f67fbd5/pkg/apis/config/v1alpha1/resourceinterpreterwebhook_types.go#L96-L98) operation for ResourceInterpreter framework.

Tasks:
- [x] Implements default interpreter (#1253)
- [x] Implements customized interpreter (#1253)
- [ ] E2E testing (#1286)

**Why is this needed**:

/assign @iawia002

PR #1253, opened 2022-01-13
Implementing the AggregateStatus hook

iawia002:

**What type of PR is this?**

/kind feature

Part of #1237

This patch implements the `AggregateStatus` hook, check [resource interpreter webhook proposal](https://github.com/karmada-io/karmada/tree/master/docs/proposals/resource-interpreter-webhook) for more info.

The original code to modify the status of core resources such as `Deployment` and `Job` has been changed to use default interpreters. After the update, the status collection of the sample deployment is the same as before:

```console
$ kubectl get deploy

nginx   2/2     2            2           56s
```

```yaml
status:
  availableReplicas: 2
  observedGeneration: 1
  readyReplicas: 2
  replicas: 2
  updatedReplicas: 2
```

for custom resources, users only need to implement the corresponding webhook and return the patch data:

```go
func (e *workloadInterpreter) responseWithExploreAggregateStatus(workload *workloadv1alpha1.Workload, req interpreter.Request) interpreter.Response {
	wantedWorkload := workload.DeepCopy()
	var readyReplicas int32
	for _, item := range req.AggregatedStatus {
		// NOTE: this is just an example, you need to change the method of updating status according to the actual situation
		if !item.Applied {
			continue
		}
		readyReplicas += 1
	}
	wantedWorkload.Status.ReadyReplicas = readyReplicas
	marshaledBytes, err := json.Marshal(wantedWorkload)
	if err != nil {
		return interpreter.Errored(http.StatusInternalServerError, err)
	}
	return interpreter.PatchResponseFromRaw(req.Object.Raw, marshaledBytes)
}
```

```yaml
status:
  readyReplicas: 2
```

Fixes #

```release-note
Implementing the AggregateStatus hook.
```


RainbowMango:
cc @XiShanYongYe-Chang 
/assign

pkg/resourceinterpreter/defaultinterpreter/aggregatestatus.go
```diff
@@ -0,0 +1,209 @@
...
+	return s
+}
+
+func toUnstructured(obj interface{}) (*unstructured.Unstructured, error) {
+	uncastObj, err := runtime.DefaultUnstructuredConverter.ToUnstructured(obj)
+	if err != nil {
+		return nil, err
+	}
+	return &unstructured.Unstructured{Object: uncastObj}, nil
+}
```
XiShanYongYe-Chang, line 40: This is a more common method, I think we can move it to `pkg/util/helper/unstructured.go` file.

pkg/resourceinterpreter/defaultinterpreter/aggregatestatus.go
```diff
@@ -0,0 +1,209 @@
...
+	object *unstructured.Unstructured, aggregatedStatusItems []workv1alpha2.AggregatedStatusItem, _ []workv1alpha2.TargetCluster,
+) (*unstructured.Unstructured, error) {
+	deploy, err := helper.ConvertToDeployment(object)
+	if err != nil {
+		return nil, err
+	}
+
+	if deploy.APIVersion != "apps/v1" {
+		return nil, nil
+	}
```
XiShanYongYe-Chang, line 52: This judgment may be redundant.

pkg/resourceinterpreter/defaultinterpreter/aggregatestatus.go
```diff
@@ -0,0 +1,209 @@
...
+	object *unstructured.Unstructured, aggregatedStatusItems []workv1alpha2.AggregatedStatusItem, _ []workv1alpha2.TargetCluster,
+) (*unstructured.Unstructured, error) {
+	service, err := helper.ConvertToService(object)
+	if err != nil {
+		return nil, err
+	}
+
+	if service.APIVersion != "v1" {
+		return nil, nil
+	}
```
XiShanYongYe-Chang, line 104: This judgment may be redundant.

pkg/resourceinterpreter/defaultinterpreter/aggregatestatus.go
```diff
@@ -0,0 +1,209 @@
...
+	object *unstructured.Unstructured, aggregatedStatusItems []workv1alpha2.AggregatedStatusItem, _ []workv1alpha2.TargetCluster,
+) (*unstructured.Unstructured, error) {
+	ingress, err := helper.ConvertToIngress(object)
+	if err != nil {
+		return nil, err
+	}
+
+	if ingress.APIVersion != "extensions/v1beta1" {
+		return nil, nil
+	}
```
XiShanYongYe-Chang, line 153: This judgment may be redundant.

pkg/resourceinterpreter/defaultinterpreter/aggregatestatus.go
```diff
@@ -0,0 +1,209 @@
...
+
+func aggregateJobStatus(object *unstructured.Unstructured, aggregatedStatusItems []workv1alpha2.AggregatedStatusItem, clusters []workv1alpha2.TargetCluster) (*unstructured.Unstructured, error) {
+	job, err := helper.ConvertToJob(object)
+	if err != nil {
+		return nil, err
+	}
+
+	if job.APIVersion != "batch/v1" {
+		return nil, nil
+	}
```
XiShanYongYe-Chang, line 195: This judgment may be redundant.

pkg/detector/detector.go
```diff
@@ -1026,19 +1026,34 @@ func (d *ResourceDetector) ReconcileResourceBinding(key util.QueueKey) error {
...
+	newObj, err := d.ResourceInterpreter.AggregateStatus(obj, binding.Status.AggregatedStatus, binding.Spec.Clusters)
+	if err != nil {
+		klog.Errorf("AggregateStatus for resource(%s/%s/%s) failed: %v", resource.Kind, resource.Namespace, resource.Name, err)
+		return err
+	}
+	if newObj == nil {
 		return nil
 	}
+
+	if _, err = d.DynamicClient.Resource(gvr).Namespace(resource.Namespace).UpdateStatus(context.TODO(), newObj, metav1.UpdateOptions{}); err != nil {
```
XiShanYongYe-Chang, line 1052: The update may be failed, we can add `retry.RetryOnConflict`.

XiShanYongYe-Chang (commented):
Nice job! Thanks 👍

pkg/detector/detector.go
```diff
@@ -1026,19 +1026,34 @@ func (d *ResourceDetector) ReconcileResourceBinding(key util.QueueKey) error {
...
+		klog.Errorf("Failed to get GVR from GVK %s %s. Error: %v", resource.APIVersion, resource.Kind, err)
+		return err
+	}
+	obj, err := d.DynamicClient.Resource(gvr).Namespace(resource.Namespace).Get(context.TODO(), resource.Name, metav1.GetOptions{})
+	if err != nil {
+		klog.Errorf("Failed to get resource(%s/%s/%s), Error: %v", resource.Kind, resource.Namespace, resource.Name, err)
+		return err
+	}
+
+	newObj, err := d.ResourceInterpreter.AggregateStatus(obj, binding.Status.AggregatedStatus, binding.Spec.Clusters)
```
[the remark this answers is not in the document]
iawia002: Oh, I forgot that, updated.

pkg/resourceinterpreter/defaultinterpreter/aggregatestatus.go
iawia002: OK, I plan to clean up the rest of the code that converts Unstructured objects in the next PR.

pkg/detector/detector.go
iawia002: It's a little tricky to use `retry` here, we need to retrieve the object, execute `AggregateStatus` and update it again, which would be a lot of redundant code, perhaps it would be better to return the error directly to trigger its resync.

pkg/resourceinterpreter/defaultinterpreter/aggregatestatus.go
XiShanYongYe-Chang: Thanks, I have the same idea.

pkg/resourceinterpreter/defaultinterpreter/aggregatestatus.go
```diff
@@ -0,0 +1,209 @@
...
+
+type aggregateStatusInterpreter func(
+	object *unstructured.Unstructured, aggregatedStatusItems []workv1alpha2.AggregatedStatusItem, clusters []workv1alpha2.TargetCluster,
+) (*unstructured.Unstructured, error)
+
+func getAllDefaultAggregateStatusInterpreter() map[schema.GroupVersionKind]aggregateStatusInterpreter {
+	s := make(map[schema.GroupVersionKind]aggregateStatusInterpreter)
+	s[appsv1.SchemeGroupVersion.WithKind(util.DeploymentKind)] = aggregateDeploymentStatus
+	s[corev1.SchemeGroupVersion.WithKind(util.ServiceKind)] = aggregateServiceStatus
+	s[extensionsv1beta1.SchemeGroupVersion.WithKind(util.IngressKind)] = aggregateIngressStatus
```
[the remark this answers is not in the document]
XiShanYongYe-Chang: I agree with you.

pkg/detector/detector.go
XiShanYongYe-Chang: Perhaps we can directly copy the status part with the get newest object?

pkg/resourceinterpreter/interpreter.go
```diff
@@ -31,6 +31,9 @@ type ResourceInterpreter interface {
 	// Retain returns the objects that based on the "desired" object but with values retained from the "observed" object.
 	Retain(desired *unstructured.Unstructured, observed *unstructured.Unstructured) (retained *unstructured.Unstructured, err error)
 
+	// AggregateStatus returns aggregated statuses of the object.
+	AggregateStatus(object *unstructured.Unstructured, aggregatedStatusItems []workv1alpha2.AggregatedStatusItem, clusters []workv1alpha2.TargetCluster) (*unstructured.Unstructured, error)
```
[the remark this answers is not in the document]
RainbowMango: > If the returned structure remains unchanged, the original value instead of nil should be returned.
What do you think?

+1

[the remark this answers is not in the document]
iawia002: > The length of `[]workv1alpha2.AggregatedStatusItem` is always equal to `len(clusters)`, why do we need the `clusters` parameter?

I was misled by the function `ParsingJobStatus`, the `clusters` parameter is required by it:

```go
func ParsingJobStatus(obj *batchv1.Job, status []workv1alpha2.AggregatedStatusItem, clusters []workv1alpha2.TargetCluster) (*batchv1.JobStatus, error)
```

iawia002:
All comments addressed, PTAL @RainbowMango @XiShanYongYe-Chang 56cd871

/hold
NOTE: I will squash the commits into one after you finished the review.

pkg/detector/detector.go
```diff
@@ -1026,19 +1028,55 @@ func (d *ResourceDetector) ReconcileResourceBinding(key util.QueueKey) error {
...
+
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		obj, err := d.DynamicClient.Resource(gvr).Namespace(resource.Namespace).Get(context.TODO(), resource.Name, metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+		if err = unstructured.SetNestedField(obj.Object, newStatus, "status"); err != nil {
+			return err
+		}
+		if _, err = d.DynamicClient.Resource(gvr).Namespace(resource.Namespace).UpdateStatus(context.TODO(), obj, metav1.UpdateOptions{}); err != nil {
```
XiShanYongYe-Chang, line 1074: Update first to save a get op.

iawia002: Updated, PTAL 9e85728

pkg/detector/detector.go
```diff
@@ -1063,7 +1064,11 @@ func (d *ResourceDetector) ReconcileResourceBinding(key util.QueueKey) error {
 		return nil
 	}
 
-	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
+	if err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
XiShanYongYe-Chang, line 1067: I'm sorry, i didn't describe it clearly, we can do an update first and then get in one action, just as:
https://github.com/karmada-io/karmada/blob/361978a3bf339db62b27757eacb45cf8194135ac/pkg/scheduler/scheduler.go#L820-L836

iawia002: Thanks for pointing out, didn't expect this method

XiShanYongYe-Chang (commented):
Thanks, LGTM

/cc @RainbowMango

pkg/detector/detector.go
```diff
@@ -1026,19 +1029,61 @@ func (d *ResourceDetector) ReconcileResourceBinding(key util.QueueKey) error {
...
+		return err
+	}
+	newStatus, _, err := unstructured.NestedFieldNoCopy(newObj.Object, "status")
+	if err != nil {
+		return err
+	}
+	oldStatus, _, err := unstructured.NestedFieldNoCopy(obj.Object, "status")
+	if err != nil {
+		return err
+	}
```
RainbowMango, line 1061: We don't need to parse the `.status`, we can just compare the `obj` and `newObj`, since the `Interpreter` won't change anything but `.status`, so if `obj` != `newObj` means the status has been `aggregated`, then we do `update` operation.

In addition, we seem can't assume there is a `.status` field. See [InterpretStatus](https://github.com/karmada-io/karmada/blob/8ceb9df2b9672d3e50b0d52d6b4e7d8c000eff81/pkg/apis/config/v1alpha1/resourceinterpreterwebhook_types.go#L85-L87).

RainbowMango: Something like this:
```golang
	if reflect.DeepEqual(obj, newObj) {
		klog.V(3).Infof("ignore update resource(%s/%s/%s) status as up to date", resource.Kind, resource.Namespace, resource.Name)
		return nil
	}
```

Maybe we can't use `retry.RetryOnConflict` for this case,

examples/customresourceinterpreter/webhook/app/workloadwebhook.go
```diff
@@ -86,3 +88,21 @@ func (e *workloadInterpreter) responseWithExploreRetaining(desiredWorkload *work
...
+
+func (e *workloadInterpreter) responseWithExploreAggregateStatus(workload *workloadv1alpha1.Workload, req interpreter.Request) interpreter.Response {
+	wantedWorkload := workload.DeepCopy()
+	var readyReplicas int32
+	for _, item := range req.AggregatedStatus {
+		// NOTE: this is just an example, you need to change the method of updating status according to the actual situation
+		if !item.Applied {
+			continue
+		}
+		readyReplicas += 1
```
RainbowMango, line 100: Why not sum the `.status.readyReplicas`? 
We can add a comment here to show how we aggregate the status for the example `Workload`, but this is `not an example`, as we'll add E2E tests for this.

RainbowMango (commented):
Good Job! @iawia002 
Generally looks good to me. Just two nits.

examples/customresourceinterpreter/webhook/app/workloadwebhook.go
iawia002: It seems karmada didn't collect status info for the example workload resource, I didn't see any `readyReplicas` data from resource bindings.

RainbowMango: You are right, we don't have the `example controller` to update the status, we simulate the controller behavior in E2E tests.

pkg/detector/detector.go
iawia002: > Maybe we can't use retry.RetryOnConflict for this case

What is this for? I can compare two objects directly, but it doesn't seem to affect the use of `RetryOnConflict`, is it because the newObj may also not have status?

examples/customresourceinterpreter/webhook/app/workloadwebhook.go
iawia002: So do I need to update this now? Maybe I can update the collection method when I add the E2E test, so I can debug if the method is correct.

RainbowMango: I think we should do it in this PR as we can confirm if all the information passed to webhook is enough.

pkg/detector/detector.go
RainbowMango: > but it doesn't seem to affect the use of RetryOnConflict, is it because the newObj may also not have status?

As after we `get` a new copy of the object, we have to set the status, do you mean to call the `webhook` again in `retry`?

iawia002: Nope, Is there a problem with the current retry method?

RainbowMango: The thing is 
```
if err = unstructured.SetNestedField(obj.Object, newStatus, "status"); err != nil {
			return err
		}
```

Personally, I'm hesitant to assume this is a `.status` filed(Maybe no problem for 99.9% cases).

iawia002: I am not so familiar with `SetNestedField`, but by looking at the source code, if the original object does not have a status field, it should not report an error when setting for it, it will only report an error if the value is not compliant.

The workload object initially had no status field, but `SetNestedField` added it, as expected.

RainbowMango: I mean we can't assume all resource's status represent by `.status`, that might be `.status-xxx` or something else. That's why we design `InterpretStatus` operation for.

iawia002: I get your point now, it seems we can't use retry in this case, let's return the error directly to trigger its resync.

examples/customresourceinterpreter/webhook/app/workloadwebhook.go
iawia002: I can update the code, but I can barely test it because I can't use kubectl to update the status of the workload object or the resource binding object, and I can't simulate the situation where the status of the workload object is updated.

RainbowMango: That's true. We can test it in E2E.

pkg/detector/detector.go
```diff
@@ -1004,6 +1005,7 @@ func (d *ResourceDetector) OnResourceBindingUpdate(_, newObj interface{}) {
 // ReconcileResourceBinding handles ResourceBinding object changes.
 // For each ResourceBinding changes, we will try to calculate the summary status and update to original object
 // that the ResourceBinding refer to.
+//nolint:gocyclo
```
RainbowMango, line 1008: Do we still need to ignore this checking? As the complexity is decreased.

RainbowMango (commented):
/lgtm

Just the `nolint` need to be clarified.

pkg/detector/detector.go
iawia002: Oh I forgot to delete this, we don't need it now.

RainbowMango (approved):
/lgtm
/approve

PR #1254, opened 2022-01-13
[scheduler]fix a defer execution sequence problem

huone1:

**What type of PR is this?**
/kind bug

when **updateBindingScheduledConditionIfNeeded** return error, we should retry it in next loop according to the comments
but the error will not be used as a return value. 
https://github.com/karmada-io/karmada/blob/9167f9bce4fac02634bc3568977f24d0ae4b3213/pkg/scheduler/scheduler.go#L418-L423

just return nil even if **updateBindingScheduledConditionIfNeeded** return error
https://github.com/karmada-io/karmada/blob/9167f9bce4fac02634bc3568977f24d0ae4b3213/pkg/scheduler/scheduler.go#L456-L458

NONE

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -396,7 +396,7 @@ func (s *Scheduler) doSchedule(key string) error {
 	return s.doScheduleClusterBinding(name)
 }
 
-func (s *Scheduler) doScheduleBinding(namespace, name string) error {
+func (s *Scheduler) doScheduleBinding(namespace, name string) (err error) {
 	rb, err := s.bindingLister.ResourceBindings(namespace).Get(name)
 	if err != nil {
 		if apierrors.IsNotFound(err) {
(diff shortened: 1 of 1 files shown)


RainbowMango:
/assign @mrlihanbo 
Please take a look.

pkg/scheduler/scheduler.go
```diff
@@ -396,8 +396,9 @@ func (s *Scheduler) doSchedule(key string) error {
 	return s.doScheduleClusterBinding(name)
 }
 
-func (s *Scheduler) doScheduleBinding(namespace, name string) error {
-	rb, err := s.bindingLister.ResourceBindings(namespace).Get(name)
+func (s *Scheduler) doScheduleBinding(namespace, name string) (err error) {
```
RainbowMango, line 399: Yeah, the named return value is the key point.

pkg/scheduler/scheduler.go
```diff
@@ -396,8 +396,9 @@ func (s *Scheduler) doSchedule(key string) error {
...
+	var rb *workv1alpha2.ResourceBinding
```
dddddai, line 400: lgtm but I'd prefer `:=` than `var`, which seems kinda tedious

RainbowMango: +1
The named return value already defined the `err` variable, `:=` won't redefine it again.

huone1: I looked up the grammar, and you're right. I will revert the modification about it

RainbowMango (approved):
/lgtm
/approve

PR #1280, opened 2022-01-19
don't block the IsInformerSynced func when waiting for cache sync

carlory:

**What type of PR is this?**

/kind bug

```release-note
NONE
```

### DIFF
--- a/pkg/util/informermanager/single-cluster-manager.go
+++ b/pkg/util/informermanager/single-cluster-manager.go
@@ -150,9 +150,9 @@ func (s *singleClusterInformerManagerImpl) WaitForCacheSyncWithTimeout(cacheSync
 }
 
 func (s *singleClusterInformerManagerImpl) waitForCacheSync(ctx context.Context) map[schema.GroupVersionResource]bool {
+	res := s.informerFactory.WaitForCacheSync(ctx.Done())
 	s.lock.Lock()
 	defer s.lock.Unlock()
-	res := s.informerFactory.WaitForCacheSync(ctx.Done())
 	for resource, synced := range res {
 		if _, exist := s.syncedInformers[resource]; !exist && synced {
 			s.syncedInformers[resource] = struct{}{}


pkg/util/informermanager/single-cluster-manager.go
```diff
@@ -150,9 +150,9 @@ func (s *singleClusterInformerManagerImpl) WaitForCacheSyncWithTimeout(cacheSync
 }
 
 func (s *singleClusterInformerManagerImpl) waitForCacheSync(ctx context.Context) map[schema.GroupVersionResource]bool {
+	res := s.informerFactory.WaitForCacheSync(ctx.Done())
```
RainbowMango, line 153: Is there any evidence for the performance improving?

Seems no bad effects, but looks strange that here operates the internal data but without the lock.

pigletfly: I don't think we should remote `WaitForCacheSync` here.It's used in https://github.com/karmada-io/karmada/blob/846b1aeb986fe02249d0b3b5b9545bfb2facb59a/pkg/controllers/status/cluster_status_controller.go#L259,https://github.com/karmada-io/karmada/blob/5eefcd6a49585f2503c3d98646945503ba1f9b42/pkg/controllers/mcs/service_export_controller.go#L187 and other places.The controller should follow `InformerFactory.Start` `InformerFactory.WaitForCacheSync` then `controller.Run`.

pigletfly (changes requested):

pkg/util/informermanager/single-cluster-manager.go
pigletfly: I don't think we should remote WaitForCacheSync here.It's used in [https://github.com/karmada-io/karmada/blob/846b1aeb986fe02249d0b3b5b9545bfb2facb59a/pkg/controllers/status/cluster_status_controller.go#L259,https://github.com/karmada-io/karmada/blob/5eefcd6a49585f2503c3d98646945503ba1f9b42/pkg/controllers/mcs/service_export_controller.go#L187]() and other places.The controller should follow `InformerFactory.Start ` `InformerFactory.WaitForCacheSync` then `controller.Run`

RainbowMango:
/close
Please re-open it if this is still needed.

PR #1289, opened 2022-01-20
os.Exit will not wait for the defer function

kerthcet:

**What type of PR is this?**
/kind bug

os.Exit will not wait for the defer function

Fixes #

```release-note
Fix failed flushing logs when scheduler os.Exit(1).
```

### DIFF
--- a/cmd/scheduler/main.go
+++ b/cmd/scheduler/main.go
@@ -10,12 +10,20 @@ import (
 )
 
 func main() {
+	if err := runSchedulerCmd(); err != nil {
+		os.Exit(1)
+	}
+}
+
+func runSchedulerCmd() error {
 	logs.InitLogs()
 	defer logs.FlushLogs()
 
 	stopChan := apiserver.SetupSignalHandler()
-
-	if err := app.NewSchedulerCommand(stopChan).Execute(); err != nil {
-		os.Exit(1)
+	command := app.NewSchedulerCommand(stopChan)
+	if err := command.Execute(); err != nil {
+		return err
 	}
+
+	return nil
 }


lonelyCZ:
Thank you pointed out. I just have a question if there are some errors occurring,  can the logs be saved and conveniently checked? Because I find there are same usages in other projects.

kerthcet:
> Thank you pointed out. I just have a question if there are some errors occurring, can the logs be saved and conveniently checked? Because I find there are same usages in other projects.

yes, but you should configure the klog flags, like setting `-log_file=xxx.log` and `--logtostderr=false`.

lonelyCZ:
LGTM, there are same discussions in `kube-scheduler` at [https://github.com/kubernetes/kubernetes/pull/104503](https://github.com/kubernetes/kubernetes/pull/104503).

And we also need to fix other karmada components.

/cc @RainbowMango

cmd/scheduler/main.go
```diff
@@ -1,21 +1,31 @@
...
 )
 
 func main() {
+	stopChan := apiserver.SetupSignalHandler()
+	command := app.NewSchedulerCommand(stopChan)
+	code := run(command)
+	os.Exit(code)
+}
+
+func run(cmd *cobra.Command) int {
```
lonelyCZ, line 22: Hi, @kerthcet , do you fix it like this [https://github.com/kubernetes/kubernetes/pull/104503](https://github.com/kubernetes/kubernetes/pull/104503) to replace function `run` with `runSchedulerCmd`, and set `SilenceErrors: true` and `SilenceUsage:  true` to avoid repeatedly print Errors.

RainbowMango: Are you requesting changes? I leave LGTM to @lonelyCZ 
/approve

RainbowMango:
> And we also need to fix other karmada components.

It's not necessary to fix this in this PR.

cmd/scheduler/main.go
kerthcet: As you say, we set `SilenceErrors: true` to avoid repeat error printing, but we didn't print errors here. So it's not needed.WDYT?

kerthcet:
> > And we also need to fix other karmada components.
> 
> It's not necessary to fix this in this PR.

Follow up with others components.

cmd/scheduler/main.go
lonelyCZ: Sorry, I just saw this reply.

I just test it, and it seems to print error twice.
```
[root@master67 karmada]# ./karmada-scheduler
I0124 19:20:16.021823   16059 scheduler.go:64] karmada-scheduler version: version.Info{GitVersion:"v0.7.0-932-g4089758-dirty", GitCommit:"4089758d3029be2b1b813337582714b196868a24", GitTreeState:"dirty", BuildDate:"2022-01-20T20:33:27Z", GoVersion:"go1.16.8", Compiler:"gc", Platform:"linux/amd64"}
W0124 19:20:16.022031   16059 client_config.go:615] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
W0124 19:20:16.022052   16059 client_config.go:620] error creating inClusterConfig, falling back to default config: unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined
Error: error building kubeconfig: invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable
Usage:
  karmada-scheduler [flags]
  ......
Use "karmada-scheduler [command] --help" for more information about a command.

command failed: error building kubeconfig: invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable
```

lonelyCZ: I suggest to change it to

```
func main() {
	if err := runSchedulerCmd(); err != nil {
		// fmt.Fprintf(os.Stderr, "%v\n", err)
		os.Exit(1)
	}
}

func runSchedulerCmd() error {
	logs.InitLogs()
	defer logs.FlushLogs()

	stopChan := apiserver.SetupSignalHandler()

	command := app.NewSchedulerCommand(stopChan)

	if err := command.Execute(); err != nil {
		return err
	}

	return nil
}
```

This looks a little clearer. How do you think?

kerthcet: The duplicate error is printed by `fmt.Println` here, removed, thanks for caching.
I have no strong preference of the code implement, updated already. Thanks.

lonelyCZ:
Thanks.

/lgtm

Issue #1262, opened 2022-01-15
KEP-697: Descheduler TODO List


Garrybest:
The PR #726  has given a new proposal about adding Descheduler for evict replicas in member clusters. Here is the TODO List about the proposal.

- [x] pb: proto struct definition. #1263
- [x] pb: add mock client for test. #1299
- [x] descheduler: core. #1392
- [x] estimator: add new interface. #1380
- [x] makefile and image workflow. #1435
- [x] deploy script. #1435
- [x] test: more test cases #1392
- [x] docs: README #1479 
- [x] helm: support helm install

Garrybest:
Hey guys, would you please do me a favour? I need some help for the chart part. @jrkeen @pidb 

`karmada-descheduler` is a component which works in the controller-plane and helps to evict replicas from ResourceBinding.
Here is the [karmada-descheduler.yaml](https://github.com/karmada-io/karmada/blob/master/artifacts/deploy/karmada-descheduler.yaml).

RainbowMango:
@Poor12 Could you please take a look helm thing? I guess @jrkeen and @pidb are pretty busy these days.

RainbowMango:
Thanks @Garrybest @Poor12 for your hard work. The `karmada-descheduler` is officially on board now!!!

PR #1299, opened 2022-01-22
descheduler: add mock client for decoupling test

Garrybest:
**What type of PR is this?**
/kind feature

According to #1262, add a mock grpc client so the descheduler could run the test by calling mock client.

I have moved service proto files into a new directory so that the mock client will not be handled by `go-to-proto` generation.

Fixes part of #1262

```release-note
NONE
```


RainbowMango:
@huone1 Could you please help take a look?

huone1:
> @huone1 Could you please help take a look?

OK， I will review it as soon as possible in this week

hack/tools/tools.go
```diff
@@ -1,10 +1,12 @@
+//go:build tools
```
RainbowMango, line 1: You added this or the `gofmt` added this?

PS: If you are using Go v1.17 which will automatically add this tag.

pkg/estimator/service/doc.go
```diff
@@ -0,0 +1,4 @@
+package service
+
+//go:generate protoc --gogo_opt=paths=source_relative -I . -I ../../.. -I ../../../vendor --gogo_out=plugins=grpc:. --gogo_opt=Mpkg/estimator/pb/generated.proto=github.com/karmada-io/karmada/pkg/estimator/pb service.proto
+//go:generate mockery --name=EstimatorClient --inpackage
```
RainbowMango, line 4: We just generate the mock code for EstimatorClient, but have not used it yet, right?

RainbowMango (commented):
/lgtm

pkg/estimator/service/doc.go
Garrybest: Right, not used yet until the descheduler is contributed.

Issue #1313, opened 2022-01-25
os.Exit() will broke the effect of defer function


kerthcet:
**What happened**:
```golang
func main() {
	logs.InitLogs()
	defer logs.FlushLogs()

	if err := xxx.Execute(); err != nil {
		os.Exit(1)
	}
}
```
`FlushLogs()`will not call if `os.Exit(1)`

**What you expected to happen**:
`FlushLogs()` should take effect even `Execute()` returns error.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:
Similar to https://github.com/karmada-io/karmada/pull/1289#event-5943184872. Following up all related prs regarding to other components in this issue. 
Besides, each pr also solves other problems:
1. Repetitive printing error message due to cobra will also print error messages: https://github.com/karmada-io/karmada/blob/85554ff18796bbf3866ffe310cea8d8557b96105/vendor/github.com/spf13/cobra/command.go#L985-L987
2. Wrapping an error with %w makes it available to `errors.Is` and `errors.As``
3. Remove `os.Exit()` if possible

**Environment**:
- Karmada version:
- kubectl-karmada or karmadactl version (the result of `kubectl-karmada version` or `karmadactl version`):
- Others:

[edit by @RainbowMango]
- [x] karmada-scheduler (#1289)
- [x] karmada-agent(#1314)
- [x] karmada-aggregated-apiserver(#1315)
- [x] karmada-controller-manager(#1316)
- [x] karmadactl(#1317)
- [x] kubectl-karmada(#1318)
- [x] karmada-scheduler-estimator(#1319)
- [x] karmada-webhook(#1320)

RainbowMango:
@lonelyCZ Could you please help review the relevant PRs of this issue?

lonelyCZ:
> @lonelyCZ Could you please help review the relevant PRs of this issue?

Ok, thanks.

PR #1314, opened 2022-01-25
bug: agent's os.Exit will not wait for the defer function

kerthcet:

**What type of PR is this?**
/kind bug

Part of # https://github.com/karmada-io/karmada/issues/1313

```release-note
None
```

### DIFF
--- a/cmd/agent/app/agent.go
+++ b/cmd/agent/app/agent.go
@@ -82,13 +82,13 @@ func run(ctx context.Context, karmadaConfig karmadactl.KarmadaConfig, opts *opti
 	klog.Infof("karmada-agent version: %s", version.Get())
 	controlPlaneRestConfig, err := karmadaConfig.GetRestConfig(opts.KarmadaContext, opts.KarmadaKubeConfig)
 	if err != nil {
-		return fmt.Errorf("error building kubeconfig of karmada control plane: %s", err.Error())
+		return fmt.Errorf("error building kubeconfig of karmada control plane: %w", err)
 	}
 	controlPlaneRestConfig.QPS, controlPlaneRestConfig.Burst = opts.KubeAPIQPS, opts.KubeAPIBurst
 
 	clusterConfig, err := controllerruntime.GetConfig()
 	if err != nil {
-		return fmt.Errorf("error building kubeconfig of member cluster: %s", err.Error())
+		return fmt.Errorf("error building kubeconfig of member cluster: %w", err)
 	}
 	clusterKubeClient := kubeclientset.NewForConfigOrDie(clusterConfig)
 	// ensure namespace where the impersonation secret to be stored in member cluster.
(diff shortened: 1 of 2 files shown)


cmd/agent/app/agent.go
```diff
@@ -106,13 +106,12 @@ func run(ctx context.Context, karmadaConfig karmadactl.KarmadaConfig, opts *opti
...
 	if err != nil {
-		return fmt.Errorf("failed to register with karmada control plane: %s", err.Error())
+		return fmt.Errorf("failed to register with karmada control plane: %w", err)
 	}
 
 	executionSpace, err := names.GenerateExecutionSpaceName(opts.ClusterName)
 	if err != nil {
-		klog.Errorf("Failed to generate execution space name for member cluster %s, err is %v", opts.ClusterName, err)
-		return err
+		return fmt.Errorf("failed to generate execution space name for member cluster %s, err is %v", opts.ClusterName, err)
```
lonelyCZ, line 114: What is the difference between `%w` and `%v`? And Should `%w` be used here?

lonelyCZ:
I suggest to change the label of this pr to `kind/cleanup`.

/cc @RainbowMango

RainbowMango:
Yeah, it's more like a cleanup, even though it works not as expected a little bit.
/remove-kind bug
/kind cleanup

@lonelyCZ I guess you can use these [robot commands](https://prow.k8s.io/command-help?repo=kubernetes%2Fkubernetes) too.

cmd/agent/app/agent.go
kerthcet: We can refer to https://go.dev/blog/go1.13-errors for details, Chapter: `Wrapping errors with %w` says:
```
Wrapping an error with %w makes it available to errors.Is and errors.As:
```
Except this, they are the same. Just refactor it by the way.

lonelyCZ: Got it, thanks! 

Do we need to change 
```
return fmt.Errorf("failed to generate execution space name for member cluster %s, err is %v", opts.ClusterName, err)
```
to
```
return fmt.Errorf("failed to generate execution space name for member cluster %s, err is %w", opts.ClusterName, err)
```

lonelyCZ:
> I guess you can use these [robot commands](https://prow.k8s.io/command-help?repo=kubernetes%2Fkubernetes) too.

Thanks, I try it next time.

cmd/agent/app/agent.go
kerthcet: Nope, `%w` only works when there's one error, or `%w` is identical to `%v`.

lonelyCZ:
Thanks, have you compiled and tested it? @kerthcet

lonelyCZ:
I guess it is reasonable.

/lgtm

kerthcet:
So I think it's a small bug somehow. Whatever.

RainbowMango (commented):
/approve

PR #1315, opened 2022-01-25
bug: aggregated-apiserver's os.Exit will not wait for the defer function

kerthcet:

**What type of PR is this?**
/kind bug

Part of #https://github.com/karmada-io/karmada/issues/1313

```release-note
None
```

### DIFF
--- a/cmd/aggregated-apiserver/app/options/options.go
+++ b/cmd/aggregated-apiserver/app/options/options.go
@@ -84,7 +84,7 @@ func (o *Options) Run(ctx context.Context) error {
 
 	restConfig, err := clientcmd.BuildConfigFromFlags(o.Master, o.karmadaConfig)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


cmd/aggregated-apiserver/app/options/options.go
```diff
@@ -84,7 +84,7 @@ func (o *Options) Run(ctx context.Context) error {
 
 	restConfig, err := clientcmd.BuildConfigFromFlags(o.Master, o.karmadaConfig)
 	if err != nil {
-		return fmt.Errorf("error building kubeconfig: %s", err.Error())
+		return fmt.Errorf("error building kubeconfig: %w", err)
```
RainbowMango, line 87: I'm ok with this(wrap the error). But I want to clarify that it's not necessary to do that as we won't check the error types.

RainbowMango (commented):
/approve

PR #1316, opened 2022-01-25
bug: controller-manager's os.Exit will not wait for the defer function

kerthcet:

**What type of PR is this?**
/kind bug

Part of # https://github.com/karmada-io/karmada/issues/1313

```release-note
None
```

### DIFF
--- a/cmd/controller-manager/controller-manager.go
+++ b/cmd/controller-manager/controller-manager.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"fmt"
 	"os"
 
 	// Note that Kubernetes registers workqueue metrics to default prometheus Registry. And the registry will be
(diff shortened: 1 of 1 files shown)


cmd/controller-manager/app/controllermanager.go
```diff
@@ -85,21 +87,18 @@ func Run(ctx context.Context, opts *options.Options) error {
 		LivenessEndpointName:       "/healthz",
 	})
 	if err != nil {
-		klog.Errorf("failed to build controller manager: %v", err)
-		return err
+		return fmt.Errorf("failed to build controller manager: %w", err)
```
RainbowMango, line 90: I can't see any benefits from the wrapped error here, but if the error happens it becomes a little complicated to figure out where the problem is.

I suggest letting this PR focus on what it meant to do.

RainbowMango: cc @lonelyCZ

kerthcet: thanks, that also makes sense to me. Updated.

RainbowMango (commented):
/lgtm
/approve

kerthcet:
Would like to apply this change to others.

PR #1336, opened 2022-02-03
deprecate --master and --karmada-config flags for aggregrated-apiserver

carlory:
Signed-off-by: carlory <baofa.fan@daocloud.io>

**What type of PR is this?**

/kind deprecation
/kind feature

**What this PR does / why we need it**:

informerfactory uses --kubeconfig to communicate with coreapiserver, we should reuse the same configuration.

```go
func (o *CoreAPIOptions) ApplyTo(config *server.RecommendedConfig) error {
	if o == nil {
		return nil
	}

	// create shared informer for Kubernetes APIs
	var kubeconfig *rest.Config
	var err error
	if len(o.CoreAPIKubeconfigPath) > 0 {
		loadingRules := &clientcmd.ClientConfigLoadingRules{ExplicitPath: o.CoreAPIKubeconfigPath}
		loader := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, &clientcmd.ConfigOverrides{})
		kubeconfig, err = loader.ClientConfig()
		if err != nil {
			return fmt.Errorf("failed to load kubeconfig at %q: %v", o.CoreAPIKubeconfigPath, err)
		}
	} else {
		kubeconfig, err = rest.InClusterConfig()
		if err != nil {
			return err
		}
	}
	if feature.DefaultFeatureGate.Enabled(features.APIServerTracing) {
		kubeconfig.Wrap(traces.WrapperFor(config.TracerProvider))
	}
	clientgoExternalClient, err := clientgoclientset.NewForConfig(kubeconfig)
	if err != nil {
		return fmt.Errorf("failed to create Kubernetes clientset: %v", err)
	}
	config.ClientConfig = kubeconfig
	config.SharedInformerFactory = clientgoinformers.NewSharedInformerFactory(clientgoExternalClient, 10*time.Minute)

	return nil
}
```

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
deprecate --master and --karmada-config flags for karmada-aggregrated-apiserver
```

PR #1321, opened 2022-01-26
introduce concurrent flags to controllers

pigletfly:
Signed-off-by: pigletfly <wangbing.adam@gmail.com>

**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:

Make controller reconcile concurrent configurable

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
`karmada-controller-manager`: Introduced several flags to specify controller's capacities:
- --concurrent-work-syncs
- --concurrent-namespace-syncs
- --concurrent-resource-template-syncs
- --concurrent-cluster-syncs
- --concurrent-clusterresourcebinding-syncs
- --concurrent-resourcebinding-syncs

`karmada-agent`: Introduced several flags to specify controller's capacities:
- --concurrent-work-syncs
- --concurrent-cluster-syncs

The larger number of these flags means faster controller reconciles, but more CPU (and network) load.
```

PR #1342, opened 2022-02-08
improve the usage description of `--kubeconfig` flag

XiShanYongYe-Chang:

**What type of PR is this?**

/kind cleanup

As discussed in https://github.com/karmada-io/karmada/pull/1336#pullrequestreview-875652444, we need to modify the usage description of `kubeconfig` flag to the correct meaning.

Fixes #

```release-note
Modify the kubeconfig usage description of the `karmada-controller-manager/karmada-scheduler/karmada-webhook/karmada-aggregated-apiserver` executable file.
```

### DIFF
--- a/cmd/aggregated-apiserver/app/options/options.go
+++ b/cmd/aggregated-apiserver/app/options/options.go
@@ -54,6 +54,7 @@ func NewOptions() *Options {
 // AddFlags adds flags to the specified FlagSet.
 func (o *Options) AddFlags(flags *pflag.FlagSet) {
 	o.RecommendedOptions.AddFlags(flags)
+	flags.Lookup("kubeconfig").Usage = "Path to karmada control plane kubeconfig file."
 
 	flags.StringVar(&o.karmadaConfig, "karmada-config", o.karmadaConfig, "Path to a karmada-apiserver KubeConfig.")
 	// Remove it when we are in v1.2(+).
--- a/cmd/controller-manager/app/options/options.go
+++ b/cmd/controller-manager/app/options/options.go
@@ -95,6 +95,8 @@ func NewOptions() *Options {
 
 // AddFlags adds flags to the specified FlagSet.
 func (o *Options) AddFlags(flags *pflag.FlagSet, allControllers []string) {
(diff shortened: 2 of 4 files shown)


pkg/util/constants.go
```diff
@@ -141,3 +141,11 @@ const (
...
 	CacheSyncTimeout = 30 * time.Second
 )
+
+const (
+	// Kubeconfig indicates the flag name of the kubeconfig
+	Kubeconfig = "kubeconfig"
+
+	// KubeconfigUsage show the usage of kubeconfig flag
+	KubeconfigUsage = "Path to karmada control plane kubeconfig file."
+)
```
RainbowMango, line 151: /assign @lonelyCZ 

Just a feeling that these const variables don't so `common`. 

In addition, we might be going to add more shared flags in the future(see #1321), I'm thinking if we need a common package to manage the shared flags.(like the [sharedcommand](https://github.com/karmada-io/karmada/tree/master/pkg/version/sharedcommand)).

lonelyCZ: The meaning of `kubeconfig` is similar in components of Karmada. I think adding a command package of  shared flags is ok.

XiShanYongYe-Chang: Do any package and file names recommend?

lonelyCZ: Like this?

https://github.com/karmada-io/karmada/blob/ca3bc35f685991d6604e2af10210d49fbf40bdcc/pkg/karmadactl/options/global.go#L16-L22

`karmada/pkg/options/sharedflag.go`?

XiShanYongYe-Chang: Prefer `karmada/pkg/options/sharedflag.go`.

How do you think @RainbowMango.

RainbowMango: package name `options` is a little bit too common that we have to use an alias. 

No better idea yet.

XiShanYongYe-Chang: How about moving it to `cmd/sharedconstant/sharedflags.go` file.

RainbowMango:
/retitle improve the usage description of `--kubeconfig` flag

RainbowMango (approved):
/approve

Leave LGTM to @lonelyCZ

PR #1348, opened 2022-02-09
Chart: make `--max-requests-inflight` and `--max-mutating-requests-inflight` configurable

pigletfly:

**What type of PR is this?**
/kind cleanup

Fixes #

```release-note

```

### DIFF
--- a/charts/templates/karmada_apiserver.yaml
+++ b/charts/templates/karmada_apiserver.yaml
@@ -70,6 +70,8 @@ spec:
             - --requestheader-username-headers=X-Remote-User
             - --tls-cert-file=/etc/kubernetes/pki/karmada.crt
             - --tls-private-key-file=/etc/kubernetes/pki/karmada.key
+            - --max-requests-inflight={{ .Values.apiServer.maxRequestsInflight }}
+            - --max-mutating-requests-inflight={{ .Values.apiServer.maxMutatingRequestsInflight }}
           ports:
             - name: http
               containerPort: 5443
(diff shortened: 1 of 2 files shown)


charts/templates/karmada_apiserver.yaml
```diff
@@ -70,6 +70,8 @@ spec:
             - --requestheader-username-headers=X-Remote-User
             - --tls-cert-file=/etc/kubernetes/pki/karmada.crt
             - --tls-private-key-file=/etc/kubernetes/pki/karmada.key
+            - --max-requests-inflight=1500
+            - --max-mutating-requests-inflight=500
```
RainbowMango, line 74: Just a question, can users configure this parameter?

leave to owners @jrkeen @algebra2k

pigletfly: ok,we can have a default value,then make them configurable

jrkeen: > ok,we can have a default value,then make them configurable

I agree with you

RainbowMango (commented):
/retitle Chart: make `--max-requests-inflight` and `--max-mutating-requests-inflight` configurable

pigletfly:
@jrkeen PTAL

jrkeen:
> @jrkeen PTAL

image

@pigletfly Do you think the `--enable-priority-and-fairness` configuration needs to be configured?

Configuration comment:
image

pigletfly:
I don't think so.Because it's not enabled by default,and if it's enabled, the `--max-requests-inflight` and `--max-mutating-requests-inflight` flags are still useful,ref :https://kubernetes.io/docs/concepts/cluster-administration/flow-control/#priority-levels

Issue #1347, opened 2022-02-09
Replica set by overridepolicy will be reset by ReviseReplica interpreterhook


likakuli:
**What happened**:
Replica set by overridepolicy in its `Plaintext` will be reset by ReviseReplica interperterhook

**What you expected to happen**:
Use the replica value after applied override policy as the replica passed to ReviseReplica interpreterhook request, not the original value from targetclusters.

**How to reproduce it (as minimally and precisely as possible)**:
Set an overridepolicy for an deploy which contains replica info in its `Plaintext`, the replica in override policy will not effect.

**Anything else we need to know?**:
If it is a bug, I can work on it.

**Environment**:
- Karmada version: 1.0.1
- kubectl-karmada or karmadactl version (the result of `kubectl-karmada version` or `karmadactl version`):
- Others:

RainbowMango:
> Set an overridepolicy for an deploy which contains replica info in its Plaintext, the replica in override policy will not effect.

Can you provide the steps to reproduce it? E.g. the yamls.

cc @XiShanYongYe-Chang @iawia002 

/assign @likakuli 
Feel free to take it once get confirmed by the owners. Thanks!

likakuli:
deployment.yaml

```
apiVersion: apps/v1
kind: Deployment
metadata:
  annotations:
    test: "test"
    test2: "test2"
  name: nginx
  labels:
    app: nginx
    test: test
spec:
  replicas: 3
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
        test: test
    spec:
      containers:
      - image: nginx
        name: nginx
```

override.yaml
```
apiVersion: policy.karmada.io/v1alpha1
kind: OverridePolicy
metadata:
  name: example-override
  namespace: default
spec:
  resourceSelectors:
    - apiVersion: apps/v1
      kind: Deployment
      name: nginx
  targetCluster:
    clusterNames:
      - member1
      - member2
  overriders:
    plaintext:
      - path: "/metadata/annotations"
        operator: add
        value:
          foo: foo
      - path: "/spec"
        operator: replace
        value:
          replicas: 1
          selector:
            matchLabels:
              app: nginx
          template:
            metadata:
              labels:
                app: nginx
                test: test
            spec:
              containers:
              - image: nginx
                name: nginx
```
propagetion.yaml
```
apiVersion: policy.karmada.io/v1alpha1
kind: PropagationPolicy
metadata:
  name: nginx-propagation
[...]

PR #1352, opened 2022-02-10
fix override result will be ignored by ReviseReplica issue

likakuli:
**What type of PR is this?**
/kind bug

Replica set by overridepolicy in its Plaintext will be reset by ReviseReplica interperterhook

Fixes https://github.com/karmada-io/karmada/issues/1347

NONE

```release-note
fix a bug that Replica set by overridepolicy will be reset by ReviseReplica interpreterhook
```

### DIFF
--- a/pkg/controllers/binding/common.go
+++ b/pkg/controllers/binding/common.go
@@ -81,11 +81,6 @@ func ensureWork(
 	for i := range targetClusters {
 		targetCluster := targetClusters[i]
 		clonedWorkload := workload.DeepCopy()
-		cops, ops, err := overrideManager.ApplyOverridePolicies(clonedWorkload, targetCluster.Name)
-		if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/controllers/binding/common.go
```diff
@@ -116,6 +111,13 @@ func ensureWork(
 			}
 		}
 
+		// Call ReviseReplica func first, as ApplyOverridePolicy has the highest priority
```
iawia002, line 114: This comment is a bit ambiguous, as it refers to the `ReviseReplica` function after it has already been called, I suggest adjusting it:

```suggestion
		// We should call ApplyOverridePolicies last, as override rules have the highest priority
```

likakuli: updated

XiShanYongYe-Chang (commented):
/lgtm

RainbowMango:
/retitle fix override result will be ignored by ReviseReplica issue

PR #1356, opened 2022-02-12
Optimize prune unused fields

pigletfly:

**What type of PR is this?**
/kind cleanup

Fixes #

```release-note
    NONE
```


pkg/resourceinterpreter/defaultinterpreter/prune/prune.go
```diff
@@ -14,38 +14,10 @@ import (
...
-	unstructured.RemoveNestedField(workload.Object, "metadata", "generation")
-
-	// This is mostly for internal housekeeping, and users typically shouldn't need to set or understand this field.
-	// Remove this field to keep 'Work' clean and tidy.
-	unstructured.RemoveNestedField(workload.Object, "metadata", "managedFields")
-
-	// populated by the kubernetes.
-	unstructured.RemoveNestedField(workload.Object, "metadata", "resourceVersion")
-
-	// populated by the kubernetes and has been deprecated by kubernetes.
```
RainbowMango, line 40: To be honest, I can't see much benefit from this change. In addition, I hope this kind of comment is still useful and should not be removed.

lonelyCZ: I think so, these comments is useful for me to understand why to remove these fields.

pigletfly: Comments can left above getIrrelevantFields func.The benefit is the removal of redundant code,if we want to add more fields to ignore,we have to copy `unstructured.RemoveNestedField(workload.Object, "metadata", "uid")` and paste.

RainbowMango:
sorry, we need more evidence for the benefits.
/close

PR #1360, opened 2022-02-15
add hpa proposal draft

YueHonghui:
**What type of PR is this?**

This PR introduces a new component FederatedHPA for Karmada.
Fixes #

```release-note
NONE
```


docs/proposals/hpa/federated-hpa.md
```diff
@@ -0,0 +1,135 @@
...
+-->
+
+## Design Details
+### Architecture
+![Karmada FederatedHPA Architecture](statics/karmadafederatedhpa-arch.drawio.png)
+
+1. The Kubernetes related HPA components are still used in the member cluster and can work standalone.
+1. The FederatedHPAController is responsible for purposes
+   1. Distribute 'some' fields of the HPA CRs between member clusters based on the `Propagation`.
+   2. Redistribute 'some' fields of the HPA CRs after disaster of member clusters detected or `Propagation` is changed.
```
kerthcet, line 103: How to detect the member cluster‘s fault in our design? Is it real-time?

kerthcet:
Hi @YueHonghui , I took an overall review and left some comments about this KEP for I'm quite curious about this feature. Correct me if I mis-understood your design, Thanks.

docs/proposals/hpa/federated-hpa.md
```diff
@@ -0,0 +1,135 @@
...
+This section is for explicitly listing the motivation, goals, and non-goals of
+this KEP.  Describe why the change is important and the benefits to users.
+-->
+
+### Goals
+1. Bring HPA from single cluster to multiple clusters.
+1. Keep the HPA related CRDs of multi-cluster version the same as the single cluster version.
+1. Tolerate the disaster of member cluster or karmada control plane.
+1. It is better to integrate well with the scenarios such as workloads shifting and cloud burst.
+1. It is better to support different implementations of HPA.
```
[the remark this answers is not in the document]
snowplayfire: Different implementations of HPA include k8s community HPA and other self-developed HPA.

docs/proposals/hpa/federated-hpa.md
```diff
@@ -0,0 +1,135 @@
...
+## Motivation
+
+<!--
+This section is for explicitly listing the motivation, goals, and non-goals of
+this KEP.  Describe why the change is important and the benefits to users.
+-->
+
+### Goals
+1. Bring HPA from single cluster to multiple clusters.
+1. Keep the HPA related CRDs of multi-cluster version the same as the single cluster version.
```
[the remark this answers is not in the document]
snowplayfire: It means zero change upgrade from single-cluster HPA to multi-cluster HPA with the help of karmada which is k8s native api compatible.

docs/proposals/hpa/federated-hpa.md
```diff
@@ -0,0 +1,135 @@
...
+For a platform developer using Kubernetes, I have implemented CD system depending on the single cluster version of HPA. Now I want to use Karmada to run apps on multiclusters. Because the Karmada version of FederatedHPA CRDs are the same as the single version, so I don't need to pay efforts to modify the code of ecosystem. HPA CRs created for the single cluster version still work for the Karmada version of FederatedHPA.
+
+#### Story 2
+For an application developer, I create a HPA CR for the application running on Karmada with FederatedHPA enabled.
+```
+target cpu util 30%
+min replica 3
+max replica 100
+```
+Suddenly, one of the member clusters which my application running on stops to work and can't scale up new pods. Unfortunately, a request burst is comming into the application. The CPU util of pods becomes higher than 30%. It will need 100 Pods totally to take the request burst. Thanks to the good design of the Karmada FederatedHPA, the FederatedHPAController detects failure and modify the HPA CR in another healthy cluster to increase the `max` value. Finally, new pods are scaled on another healthy member cluster successfully.
```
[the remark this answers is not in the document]
snowplayfire: Under normal circumstances,  FederatedHPAController distributes `max` into member clusters when custom resource `HPA` is created or updated.
But if a certain cluster breaks down and the cluster's HPAController can not scale,  FederatedHPAController should redistribute the cluster's share into other healthy member clusters.

docs/proposals/hpa/federated-hpa.md
```diff
@@ -0,0 +1,135 @@
...
+```
+Suddenly, one of the member clusters which my application running on stops to work and can't scale up new pods. Unfortunately, a request burst is comming into the application. The CPU util of pods becomes higher than 30%. It will need 100 Pods totally to take the request burst. Thanks to the good design of the Karmada FederatedHPA, the FederatedHPAController detects failure and modify the HPA CR in another healthy cluster to increase the `max` value. Finally, new pods are scaled on another healthy member cluster successfully. 
+
+#### Story 3
+As an administrator of the Karmada&Kubernetes platform, I receive an alert that the Karmada control plane stops to work and any requests to the Karmada control plane are failed. There are many applications running on the platform heavilly depend on the HPA to handle the unpredictable burst of requests. The chance of RCA occurred becomes really high if the system can't tolerate the failure of federation control plane.
+Thanks to the design of Karmada FederatedHPA, the single cluster version of HPA is reused and still works as expected in the member cluster without the help of Karmada control plane or any other federated controller.
```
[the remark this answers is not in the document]
snowplayfire: Maybe the description is not very accurate, but it is really needed.

Thanks to the design of the FederatedHPA, member clusters HPA capacity should not be affected when fed control plane breaks down, because each member cluster's HPAController can work without fed control plane.

docs/proposals/hpa/federated-hpa.md
snowplayfire: It depends on `Karmada`, which has already solved your question.

snowplayfire:
@kerthcet  Thanks for your comments, I hope I have answered your questions.

docs/proposals/hpa/federated-hpa.md
```diff
@@ -0,0 +1,135 @@
...
+### User Stories (Optional)
+
+<!--
+Detail the things that people will be able to do if this KEP is implemented.
+Include as much detail as possible so that people can understand the "how" of
+the system. The goal here is to make this feel real for users without getting
+bogged down.
+-->
+
+#### Story 1
```
merryzhou, line 58: i have  a story：

2 member clusters C1 and C2, a deployment D deployed in C1, and the hpa configuration is :

```
target cpu util 30%
min replica 1
max replica 50
```
i hope when the deployment‘s cpu util exceed 30%, the deployment can be scaled up in cluster C1。But if the cluster C1 don't have enough resources for D to scale,  new pods(or deployment) of deployment D can be created in Cluster C2.

Does the design cover this story?

docs/proposals/hpa/federated-hpa.md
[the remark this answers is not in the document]
YueHonghui: @kerthcet Now we have a different design and implementation of HPA widely used in our company. But we also have received the suggestion that it is better to support original version of HPA in K8s. So, this goal means it is better to support original K8s HPA and other implementation of HPA.

docs/proposals/hpa/federated-hpa.md
YueHonghui: @merryzhou Yes, this story is also taken into account in this design. But we may need another component `node autoscaler` to handle this scenario. If cluster C1 don't have enough resources, `node autoscaler` should join more nodes into C1. There is another option that FedHPAController can detect this and redistribute the `min/max`, but I prefer the `node autoscaler` option.

docs/proposals/hpa/federated-hpa.md
```diff
@@ -0,0 +1,135 @@
...
+- "@robot"
+- TBD
+
+creation-date: 2022-02-14
+
+---
+
+# Federated HPA
+## Summary
+HPA is a widely used approach to improve the stablity of apps dealing with burst of request and resource utility. We can run our apps on multiple clusters as easy as a single cluster with Karmada, but Karmada lacks the HPA support now.
```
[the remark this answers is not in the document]
YueHonghui: Thanks, I've corrected it in the new commit.

docs/proposals/hpa/federated-hpa.md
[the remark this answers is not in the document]
YueHonghui: @kerthcet Ah...my poor English. @snowplayfire 's comment is what I want to say. I've improved the sentence here in the new commit.

docs/proposals/hpa/federated-hpa.md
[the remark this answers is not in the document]
YueHonghui: @kerthcet I've modified the story in the last commit.

docs/proposals/hpa/federated-hpa.md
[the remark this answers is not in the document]
YueHonghui: @kerthcet Thanks for your suggestion. I've corrected in the last commit.

docs/proposals/hpa/federated-hpa.md
merryzhou: Hi， I  have another story：
2 member clusters C1 and C2, a deployment D deployed in C1, and the hpa configuration is :

```
target cpu util 30%
min replica 1
max replica 50
```

When the scaled up conditions are met, I expect there is an hpa strategy to scale the first 30 instances in cluster C1, and the last 20 instances in cluster C2。

Is this story covered？

YueHonghui: No, this is not covered by this design(FederatedHPA). I think the story is very similar to scenario `cloud burst`, which is that we prefer to use resources in our private cloud than public cloud unless out of capacity. We plan to cover this story in other components not the FederatedHPA. This is exactly what the goal "It is better to integrate well with the scenarios such as workloads shifting and cloud burst" means.

snowplayfire:
1. The FederatedHPAController is responsible for purposes
   1. Distribute 'some' fields of the HPA CRs/resources between member clusters based on the `Propagation`.
   2. Redistribute 'some' fields of the HPA CRs/resources after disaster of member clusters detected or `Propagation` is changed.

Let me give an example to explain the above content.

Suppose 7 replicas need to be propagated to cluster1, cluster2 and cluster3 equally by FederatedHPAController . 3 replicas are distributed to cluster1,  2 replicas are distributed to cluster2,  2 replicas are distributed to cluster3 as the following  second column. 

If cluster1 has a disaster,  FederatedHPAController redistributes cluster1's 3 replicas to other healthy member clusters as the following third column.

|             |replicas before cluster1 breaks down| replicas after cluster1 breaks down |
|-------------|-------|-------|
| fed         |  7    |  7    |
| cluster1  |  3    |  3    |
| cluster2    |  2    |  4    |
| cluster3    |  2    |  3    |

docs/proposals/hpa/federated-hpa.md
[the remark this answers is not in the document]
RainbowMango: We can update this item to `It is better to support both Kubernetes HPA and customized HPA.`

RainbowMango:
Hi @snowplayfire As we discussed at the meeting, we need to add more details about the design(such as examples). 
Are you going to update on this PR or by following PR, what's your preference?

YueHonghui:
@RainbowMango I plan to update the PR at this weekend.

docs/proposals/hpa/federated-hpa.md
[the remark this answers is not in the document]
kerthcet: > @kerthcet Now we have a different design and implementation of HPA widely used in our company. But we also have received the suggestion that it is better to support original version of HPA in K8s. So, this goal means it is better to support original K8s HPA and other implementation of HPA.

Thanks for explaining.

> 

Thanks for explaining.

docs/proposals/hpa/federated-hpa.md
kerthcet: Maybe we can add some details in case someone not familiar with this.

> @kerthcet Thanks for your comments, I hope I have answered your questions.

Sorry for the delayed reply, Looking forward to your detailed design.

kerthcet:
> @kerthcet Thanks for your comments, I hope I have answered your questions.

Sorry for the delayed reply, Looking forward to your detailed design

RainbowMango:
cc @GitHubxsy @kevin-wangzefeng

docs/proposals/hpa/federated-hpa.md
```diff
@@ -0,0 +1,333 @@
...
+When a new HPA resource created or changed, the `FederatedHPAController` should know the propagation and weight information of the corresponding `Workload`. How does the `FederatedHPAController` know it? The `FederatedHPAController` will easily find the corresponding `Workload` based on the field `ScaleTargetRef` and then will find the `PropagationPolicy` resource based on the matching of `ResourceSelectors`. 
+
+For the weight information, because the karmada scheduler already plays the role to schedule the replicas, the `FederatedHPAController` can simplely reuse the scheduling result to learn the weight. The `HPAController` in the member cluster scale the `Workload` in the member cluster directly, it will conflict between the karmada scheduler. Maybe we can introduce `autoscaling/manualscaling` mode represented by the annotation `policy.karmada.io/scalingmode=[autoscaling|manualscaling]` in the `ResourceBinding` consuming by karmada scheduler and binding controller. In the `autoscaling` mode, the karmada scheduler should not schedule the replicas directly but calculate the weight information instead and the binding controller should ignore the replicas override.
+
```
riverzhang, line 111: the binding controller should ignore the replicas override
it should be updated, then binding controller will update work resources. Only in this way, replicas of the work is the same as replicas of the member cluster. Otherwise, the execution controller will update the old replicas to the member cluster. the workload will keep scale up，then scale down in memeber cluster. it will repeat these steps.

YueHonghui: @riverzhang Thanks for your comment. You are right, it wouldn't work if we simply remove the replicas override in the binding controller. After discussion with @snowplayfire , we find that we can use [resource interpreter webhook](https://github.com/karmada-io/karmada/tree/master/docs/proposals/resource-interpreter-webhook) to retain the replicas in the member cluster. I will update the proposal later.

RainbowMango:
Hi @YueHonghui Will you present at [tomorrow's meeting](https://docs.google.com/document/d/1y6YLVC-v7cmVAdbjedoyR5WL0-q45DBRXTvz5_I7bkA/edit)? If yes, please add an agenda to it. And after joining the [mail group](https://groups.google.com/forum/#!forum/karmada) you'll be able to edit the notes.

YueHonghui:
@RainbowMango I will. Done with the agenda.

docs/proposals/hpa/federated-hpa.md
```diff
@@ -0,0 +1,286 @@
...
+-->
+
+## Design Details
+### Architecture
+![Karmada FederatedHPA Architecture](statics/karmadafederatedhpa-arch.drawio.png)
+
+There are no new CRDs or resources introduced in this design. All the core functions are implemented in the `FederatedHPAController`.
+1. The Kubernetes HPA components are still used in the member cluster and can work standalone.
+1. The FederatedHPAController is responsible for the purposes
+   1. Watch the HPA resource and `PropagationPolicy/ResourceBinding` corresponding to the `Workload`, to learn which clusters the HPA resource should propagated to and what weight the workloads should be spread between clusters.
```
RainbowMango, line 102:
```suggestion
   1. Watch the HPA resource and `ResourceBinding` corresponding to the `Workload`, to learn which clusters the HPA resource should propagated to and what weight the workloads should be spread between clusters.
```
I think the `FederatedHPAController` don't need to watch `PropagationPolicy`, it can get the workload's metadata and target cluster from the `ResourceBinding`.

docs/proposals/hpa/federated-hpa.md
```diff
@@ -0,0 +1,286 @@
...
+   1. Create `Work` corresponding to HPA resource to spread the HPA to clusters. Distribute `min/max` fields of the HPA resources between member clusters based on the weight learned.
+   1. Redistribute 'some' fields of the HPA resources after `PropagationPolicy/ResourceBinding` corresponding to `Workload` is changed.
```
RainbowMango, line 104:
```suggestion
   1. Redistribute 'some' fields of the HPA resources after `ResourceBinding` corresponding to `Workload` is changed.
```
Same as above.

docs/proposals/hpa/federated-hpa.md
```diff
@@ -0,0 +1,286 @@
...
+In the original Kubernetes HPA and Workload(Deployment, for example), HPA scale the workload through the `scale` subresource of Workload. Then, the field `replicas` will be modified to the desired number. But in this design, HPAController in the member cluster work standalone and don't scale workloads through control plane so that the actual number of pods in the member clusters don't match the `spec.replicas` of Workload in the control plane. This mismatch would cause incident when `spec.replicas` in control plane is much smaller than in member clusters and user delete the HPA resource. To solve this problem, `FederatedHPAController` can collect the sum of `spec.replicas` values from member clusters and set it to the `scale` subresource of Workload in the control plane.
+
+Every time the `spec.replicas` of Workload in the control plane is modified, the replicas distribution in the `Work` will be re-calculated by Karmada scheduler. The replicas distribution calculated by Karmada scheduler most probably don't match the actual distribution in the member clusters. The mismatch need to be solved to get rid of the risk of incident mentioned above.
```
RainbowMango, line 115: > The mismatch need to be solved to get rid of the risk of incident mentioned above.

It is the issue we talked about at the last meeting, any idea about this?

docs/proposals/hpa/federated-hpa.md
```diff
@@ -0,0 +1,286 @@
...
+kind: ConfigMap
+metadata:
+  name: federated-hpa
+  namespace: karmada-system
+data:
+# ...
+  enabledByDefault: true/false
+```
+
+If the global `enabledByDefault` is set to `true`, `FederatedHPAController` will take control of all the HPA resources except which is labeled `hpa.karmada.io/enabled=false`.
```
RainbowMango, line 138: I suggest just introducing the `federatedhpa.karmada.io/enabled` label for alpha.
- federatedhpa.karmada.io/enabled=true means the HPA is FederatedHPA and it'll be processed by FederatedHPAController.
- federatedhpa.karmada.io/enabled=false means the HPA is `normal` HPA and it'll be propagated by PP/CPP.

YueHonghui: @RainbowMango  Do you mean remove the `enabledByDefault` global configuration? If so, what if HPA resources without label `federatedhpa.karmada.io/enabled` ? Is it regarded as a normal HPA or federated HPA resource? For scenario which only uses FederatedHPA, it would be too tedious to label every HPA resources instead of using global `enabledByDefault`.

RainbowMango: > Do you mean remove the enabledByDefault global configuration?

Yes. 

>  For scenario which only uses FederatedHPA, it would be too tedious to label every HPA resources instead of using global enabledByDefault.

Yes, I agree with you. Can we tolerate it at the `alpha` stage?

I'm hesitant to introduce a configuration by `Secret`, as the configuration essentially is a kind of API. And the API group [config.karmada.io](https://github.com/karmada-io/karmada/tree/master/pkg/apis/config) is the appropriate place.

Yeah, I also agree to start to design the config API, but I'm afraid it'll slow down the progress.

YueHonghui: @RainbowMango It makes sense, I agree with you.

RainbowMango: By the way, are you going to present at the community meeting today? I guess we can talk about how to dev plan now.

RainbowMango:
Hi @YueHonghui thanks for your update. 

Would you mind squashing your commits and then we are ready to go.

YueHonghui:
@RainbowMango Done

RainbowMango (approved):
/lgtm
/approve

PR #1358, opened 2022-02-14
fix default image tags for `<karmadactl/kubectl-karmada> init`

zach593:
Signed-off-by: zach593 <zach_li@outlook.com>

**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:

- kubectl-karmada/karmadactl from releases will refer to other karmada components with static version when `<bin> init` by default

**Which issue(s) this PR fixes**:

- versioned karmadactl/kubectl-karmada still ref to latest images, it causes https://github.com/karmada-io/karmada/issues/1291#issuecomment-1018307337

**Special notes for your reviewer**:

these changes are split from https://github.com/karmada-io/karmada/pull/1345

**Does this PR introduce a user-facing change?**:

```release-note
kubectl-karmada/karmadactl from release will pull karmada components with the same version instead of "latest"
```

PR #1365, opened 2022-02-15
build and upload images when publishing

zach593:
Signed-off-by: zach593 <zach_li@outlook.com>

**What type of PR is this?**

/kind feature

build and upload images with static tags when release published

NONE

split from https://github.com/karmada-io/karmada/pull/1358

```release-note
NONE
```

### DIFF
--- a/.github/workflows/release-image.yml
+++ b/.github/workflows/release-image.yml
@@ -1,0 +1,30 @@
+name: Release Images
+on:
+  release:
+    types:
+      - published
+jobs:
+  release-image:
+    name: release images
+    # prevent job running from forked repository, otherwise
+    # 1. running on the forked repository would fail as missing necessary secret.
+    # 2. running on the forked repository would use unnecessary GitHub Action time.
+    if: ${{ github.repository == 'karmada-io/karmada' }}
+    runs-on: ubuntu-18.04
+    steps:
+      - name: checkout code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: install Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.16.x
+      - name: build images
+        env:
+          REGISTRY: ${{secrets.SWR_REGISTRY}}
+          REGISTRY_USER_NAME: ${{secrets.SWR_REGISTRY_USER_NAME}}
+          REGISTRY_PASSWORD: ${{secrets.SWR_REGISTRY_PASSWORD}}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -44,3 +44,27 @@ jobs:
     - uses: actions/checkout@master
     - name: Update new version in krew-index
       uses: rajatjindal/krew-release-bot@v0.0.40
+  publish-image:
+    name: publish images
```
RainbowMango, line 48:
```suggestion
  release-image:
    name: release image
```

.github/workflows/release.yml
```diff
@@ -44,3 +44,27 @@ jobs:
...
+    # prevent job running from forked repository, otherwise
+    # 1. running on the forked repository would fail as missing necessary secret.
+    # 2. running on the forked repository would use unnecessary GitHub Action time.
+    if: ${{ github.repository == 'karmada-io/karmada' }}
+    runs-on: ubuntu-18.04
+    steps:
+    - name: checkout code
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
```
RainbowMango, line 58: Just a question, it defaults to check out the code from the master branch, right?
What if we cut a release from the release branch? like cut `v1.0.2` from branch `release-1.0`?

RainbowMango (commented):
I suggest moving this job to a standalone workflow file, e.g. `release-image.yml`.

zach593:
quote from https://github.com/actions/checkout#usage
```
    # The branch, tag or SHA to checkout. When checking out the repository that
    # triggered a workflow, this defaults to the reference or SHA for that event.
    # Otherwise, uses the default branch.
    ref: ''
```

.github/workflows/release.yml
zach593: should it always be "publish a release"?

RainbowMango: ?
This workflow takes the responsibility of releasing the images when we cut a release.

RainbowMango:
> quote from https://github.com/actions/checkout#usage
> 
> ```
>     # The branch, tag or SHA to checkout. When checking out the repository that
>     # triggered a workflow, this defaults to the reference or SHA for that event.
>     # Otherwise, uses the default branch.
>     ref: ''
> ```

Good, that's fine.

.github/workflows/release.yml
zach593: forget it, doesn't matter, just focused on "release" should be a noun or a verb

RainbowMango (approved):
/lgtm
/approve

PR #1370, opened 2022-02-16
feat: quick return in filter phase when unsuccessful state returned

kerthcet:

**What type of PR is this?**
/kind feature

Fixes #

```release-note
None
```

### DIFF
--- a/pkg/scheduler/core/generic_scheduler.go
+++ b/pkg/scheduler/core/generic_scheduler.go
@@ -89,9 +89,7 @@ func (g *genericScheduler) findClustersThatFit(
 	var out []*clusterv1alpha1.Cluster
 	clusters := clusterInfo.GetReadyClusters()
 	for _, c := range clusters {
-		resMap := fwk.RunFilterPlugins(ctx, placement, resource, c.Cluster())
-		res := resMap.Merge()
-		if !res.IsSuccess() {
+		if result := fwk.RunFilterPlugins(ctx, placement, resource, c.Cluster()); !result.IsSuccess() {
 			klog.V(4).Infof("cluster %q is not fit", c.Cluster().Name)
 		} else {
 			out = append(out, c.Cluster())
@@ -264,9 +262,7 @@ func (g *genericScheduler) FailoverSchedule(ctx context.Context, placement *poli
 			return result, fmt.Errorf("failed to get clusterObj by clusterName: %s", clusterName)
 		}
 
-		resMap := g.scheduleFramework.RunFilterPlugins(ctx, placement, &spec.Resource, clusterObj.Cluster())
-		res := resMap.Merge()
-		if !res.IsSuccess() {
+		if result := g.scheduleFramework.RunFilterPlugins(ctx, placement, &spec.Resource, clusterObj.Cluster()); !result.IsSuccess() {
 			klog.V(4).Infof("cluster %q is not fit", clusterName)
 		} else {
 			candidateClusters.Insert(clusterName)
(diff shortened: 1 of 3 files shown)


RainbowMango:
cc @huone1

RainbowMango (commented):
I'm ok with this change, this could return faster once a plugin denies the option.
But, how to deal with the `merge` process?
https://github.com/karmada-io/karmada/blob/cd80e2f3c273aaea541457c7465f603c0fe760f2/pkg/scheduler/core/generic_scheduler.go#L92-L93

With this change, we don't have to return a `map`, should we change the interface as well?

pkg/scheduler/framework/runtime/framework.go
```diff
@@ -53,6 +53,9 @@ func (frw *frameworkImpl) RunFilterPlugins(ctx context.Context, placement *polic
 	for _, p := range frw.filterPlugins {
 		pluginResult := p.Filter(ctx, placement, resource, cluster)
 		result[p.Name()] = pluginResult
```
huone1, line 55: resultmap just save the result about part of filterPlugin. it is better to return the failure result

kerthcet:
> I'm ok with this change, this could return faster once a plugin denies the option. But, how to deal with the `merge` process?
> 
> https://github.com/karmada-io/karmada/blob/cd80e2f3c273aaea541457c7465f603c0fe760f2/pkg/scheduler/core/generic_scheduler.go#L92-L93
> 
> With this change, we don't have to return a `map`, should we change the interface as well?

reasonable for now.

pkg/scheduler/framework/runtime/framework.go
kerthcet: Failure reasons are already recorded in `Filter` result, do I understood your words correctly?

huone1: I think it is better to just record the failure reasons and remove the success record if it would return once one plugin filtered the cluster

kerthcet: Yes, it was updated as you advised. Now, I will only return a `*Result` instead of a `Map` struct. PTAL thanks a lot.

kerthcet:
updated guys.

kerthcet:
kindly ping @huone1

kerthcet:
ping @RainbowMango

pkg/scheduler/framework/runtime/framework.go
```diff
@@ -48,14 +48,13 @@ func NewFramework(plugins []string) framework.Framework {
...
 	for _, p := range frw.filterPlugins {
-		pluginResult := p.Filter(ctx, placement, resource, cluster)
-		result[p.Name()] = pluginResult
+		if result := p.Filter(ctx, placement, resource, cluster); !result.IsSuccess() {
+			return result
+		}
 	}
-
-	return result
+	return nil
```
RainbowMango, line 57: It looks weird when all plugins pass but returns a `nil` result.

pkg/scheduler/core/generic_scheduler.go
```diff
@@ -89,9 +89,7 @@ func (g *genericScheduler) findClustersThatFit(
 	var out []*clusterv1alpha1.Cluster
 	clusters := clusterInfo.GetReadyClusters()
 	for _, c := range clusters {
-		resMap := fwk.RunFilterPlugins(ctx, placement, resource, c.Cluster())
-		res := resMap.Merge()
-		if !res.IsSuccess() {
+		if result := fwk.RunFilterPlugins(ctx, placement, resource, c.Cluster()); !result.IsSuccess() {
```
RainbowMango, line 92: I know it is allowed by Golang that involve methods like `IsSuccess` by a `nil` receiver, but it looks creepy(nil.IsSuccess()).

kerthcet: Updated as make sense to newcomers when reading codes.

kerthcet:
Squashed also. Thanks guys.

RainbowMango (approved):
/lgtm
/approve

Thanks.

PR #1378, opened 2022-02-19
delete some invalid logic in static weight strategy

huone1:

**What type of PR is this?**
/kind flake

delete some invalid logic in static weight strategy and it is no impact on functions. 
https://github.com/karmada-io/karmada/blob/5833252422306f62a69d6066aab0802d10ca7354/pkg/scheduler/core/division_algorithm.go#L100-L104
Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/core/division_algorithm.go
+++ b/pkg/scheduler/core/division_algorithm.go
@@ -97,12 +97,6 @@ func divideReplicasByStaticWeight(clusters []*clusterv1alpha1.Cluster, weightLis
 
 	divideRemainingReplicas(int(replicas-allocatedReplicas), desireReplicaInfos, clusterNames)
 
-	for _, cluster := range clusters {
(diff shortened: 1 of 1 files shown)


RainbowMango:
@XiShanYongYe-Chang 
Can you confirm if the `static weight` feature is covered by E2E? If not, we need to add one.

Garrybest:
/lgtm
Right, no need to keep the empty result.

XiShanYongYe-Chang:
> Can you confirm if the `static weight` feature is covered by E2E? If not, we need to add one.

There is one E2E case that covered it: https://github.com/karmada-io/karmada/blob/2d48dd495c1165c3bc59c1250867d94079ec8dd2/test/e2e/scheduling_test.go#L509

RainbowMango:
OK

RainbowMango (approved):
/approve

PR #1383, opened 2022-02-20
[feature]support rescheduling when deleting a cluster

### LINKED ISSUE #829: Reschedule bindings on cluster change
**What happened**:
Unjoined clusters still remain in `binding.spec.clusters`

Unjoined clusters should be deleted from `binding.spec.clusters`

1.Set up environment(script v0.8)
```sh
root@myserver:~/karmada# hack/local-up-karmada.sh

root@myserver:~/karmada# hack/create-cluster.sh member1 $HOME/.kube/karmada.config

root@myserver:~/karmada# kubectl config use-context karmada-apiserver

root@myserver:~/karmada# karmadactl join member1 --cluster-kubeconfig=$HOME/.kube/karmada.config

root@myserver:~/karmada# kubectl apply -f samples/nginx

root@myserver:~/karmada# kubectl get deploy
nginx   1/1     1            1           47h
```

2.Unjoin member1
```sh
root@myserver:~/karmada# karmadactl unjoin member1

root@myserver:~/karmada# kubectl get clusters
No resources found
```

3.Check `binding.spec.clusters`
```sh
root@myserver:~/karmada# kubectl describe rb
......
Spec:
  Clusters:
    Name:  member1
......
```

Is it an expected behavior? If not, who is supposed to take the responsibility to delete unjoined clusters from binding? Scheduler or other controllers (like cluster controller)?

**Environment**:
- Karmada version:v0.8.0
- Others:

### LINKED ISSUE #1411: cannot reschedule when propagationpolicy changed
**What happened**:
1. two clusters: member1 and member2, both in ready status
2. a deployment with 2 replicas, and the propagationpolicy spec is:
```
spec:
  placement:
    clusterAffinity:
      clusterNames:
      - member1
    replicaScheduling:
      weightPreference:
```
3. wait pods of deployment deployed to member1. then update propagationpolicy to:
```
spec:
  placement:
    clusterAffinity:
      clusterNames:
      - member2                               # this line changed
    replicaScheduling:
      weightPreference:
```

i hope the deployment can be moved to member2 cluster, but nothing happend.
i think the problem may be here: https://github.com/karmada-io/karmada/blob/master/pkg/scheduler/core/division_algorithm.go#L36

Accoding to this logic, in the above case, assignedReplicas == spec.Replicas ==>  the target clusters not changed ==> no need to reschedule

**Environment**:
- Karmada version: v1.0.1
- Others:

huone1:

**What type of PR is this?**

/kind feature

support rescheduling when deleting a cluster
Fixes #829
Fixes #1411

This PR does not consider the spreadconstraints and it will be considered in the scheduler refactoring .

```release-note
`karmada-scheduler`: Workload is now able to reschedule after the cluster be unregistered.
```

### DIFF
--- a/pkg/scheduler/core/division_algorithm.go
+++ b/pkg/scheduler/core/division_algorithm.go
@@ -34,10 +34,13 @@ func divideReplicasByResource(
 	spec *workv1alpha2.ResourceBindingSpec,
 	preference policyv1alpha1.ReplicaDivisionPreference,
 ) ([]workv1alpha2.TargetCluster, error) {
-	// Step 1: Get previous total sum of replicas.
-	assignedReplicas := util.GetSumOfReplicas(spec.Clusters)
+	// Step 1: Find the ready clusters that have old replicas
+	scheduledClusters := findOutScheduledCluster(spec.Clusters, clusters)
 
-	// Step 2: Check the scale type (up or down).
+	// Step 2: calculate the assigned Replicas in scheduledClusters
+	assignedReplicas := util.GetSumOfReplicas(scheduledClusters)
+
+	// Step 3: Check the scale type (up or down).
 	if assignedReplicas > spec.Replicas {
 		// We need to reduce the replicas in terms of the previous result.
 		newTargetClusters, err := scaleDownScheduleByReplicaDivisionPreference(spec, preference)
@@ -48,7 +51,7 @@ func divideReplicasByResource(
 	} else if assignedReplicas < spec.Replicas {
 		// We need to enlarge the replicas in terms of the previous result (if exists).
 		// First scheduling is considered as a special kind of scaling up.
-		newTargetClusters, err := scaleUpScheduleByReplicaDivisionPreference(clusters, spec, preference)
+		newTargetClusters, err := scaleUpScheduleByReplicaDivisionPreference(clusters, spec, preference, scheduledClusters, assignedReplicas)
 		if err != nil {
 			return nil, fmt.Errorf("failed to scaleUp: %v", err)
 		}
@@ -202,24 +205,20 @@ func scaleUpScheduleByReplicaDivisionPreference(
 	clusters []*clusterv1alpha1.Cluster,
 	spec *workv1alpha2.ResourceBindingSpec,
 	preference policyv1alpha1.ReplicaDivisionPreference,
+	scheduledClusters []workv1alpha2.TargetCluster,
+	assignedReplicas int32,
 ) ([]workv1alpha2.TargetCluster, error) {
-	// Step 1: Find the clusters that have old replicas, so we can prefer to assign new replicas towards them.
-	scheduledClusters := findOutScheduledCluster(spec.Clusters, clusters)
-
-	// Step 2: calculate the assigned Replicas in scheduledClusters
-	assignedReplicas := util.GetSumOfReplicas(scheduledClusters)
-
-	// Step 3: Get how many replicas should be scheduled in this cycle and construct a new object if necessary
+	// Step 1: Get how many replicas should be scheduled in this cycle and construct a new object if necessary
 	newSpec := spec
 	if assignedReplicas > 0 {
 		newSpec = spec.DeepCopy()
 		newSpec.Replicas = spec.Replicas - assignedReplicas
 	}
 
-	// Step 4: Calculate available replicas of all candidates
+	// Step 2: Calculate available replicas of all candidates
 	clusterAvailableReplicas := calAvailableReplicas(clusters, newSpec)
 
-	// Step 5: Begin dividing.
+	// Step 3: Begin dividing.
 	// Only the new replicas are considered during this scheduler, the old replicas will not be moved.
 	// If not, the old replicas may be recreated which is not expected during scaling up.
 	// The parameter `scheduledClusterNames` is used to make sure that we assign new replicas to them preferentially
@@ -230,6 +229,6 @@ func scaleUpScheduleByReplicaDivisionPreference(
 		return result, err
 	}
 
-	// Step 6: Merge the result of previous and new results.
+	// Step 4: Merge the result of previous and new results.
 	return util.MergeTargetClusters(scheduledClusters, result), nil
 }
--- a/pkg/scheduler/core/generic_scheduler.go
+++ b/pkg/scheduler/core/generic_scheduler.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"time"
 
-	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog/v2"
 
 	clusterv1alpha1 "github.com/karmada-io/karmada/pkg/apis/cluster/v1alpha1"
@@ -21,7 +20,6 @@ import (
 // ScheduleAlgorithm is the interface that should be implemented to schedule a resource to the target clusters.
 type ScheduleAlgorithm interface {
 	Schedule(context.Context, *policyv1alpha1.Placement, *workv1alpha2.ResourceBindingSpec) (scheduleResult ScheduleResult, err error)
-	FailoverSchedule(context.Context, *policyv1alpha1.Placement, *workv1alpha2.ResourceBindingSpec) (scheduleResult ScheduleResult, err error)
 }
 
 // ScheduleResult includes the clusters selected.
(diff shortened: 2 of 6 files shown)


mrlihanbo:
notes: This PR does not consider the spreadconstraints.
/lgtm

RainbowMango:
Please solve the conflict. @huone1

huone1:
issue #1411 is also fixed in the PR

pkg/scheduler/core/division_algorithm.go
```diff
@@ -48,7 +51,7 @@ func divideReplicasByResource(
 	} else if assignedReplicas < spec.Replicas {
 		// We need to enlarge the replicas in terms of the previous result (if exists).
 		// First scheduling is considered as a special kind of scaling up.
-		newTargetClusters, err := scaleUpScheduleByReplicaDivisionPreference(clusters, spec, preference)
+		newTargetClusters, err := scaleUpScheduleByReplicaDivisionPreference(clusters, spec, preference, scheduledClusters, assignedReplicas)
 		if err != nil {
 			return nil, fmt.Errorf("failed to scaleUp: %v", err)
 		}
```
merryzhou, line 57: what if assignedReplicas == spec.Replicas, but len(scheduledClusters) < len(spec.clusters)

huone1: I didn't think of a similar rescheduling scenario

huone1: > what if assignedReplicas == spec.Replicas, but len(scheduledClusters) < len(spec.clusters)

I didn't think of a similar rescheduling scenario

huone1:
> Please solve the conflict. @huone1

ok, it is done

RainbowMango:
cc @dddddai @Garrybest 
Can you take a look?

RainbowMango (approved):
/approve

Leave LGTM to @Garrybest or @dddddai

pkg/scheduler/core/division_algorithm.go
dddddai: Could you please give an example? It seems not possible to happen "in one event"

dddddai (commented):
Just one concern, would leave lgtm to @Garrybest

pkg/scheduler/core/generic_scheduler.go
```diff
@@ -246,53 +244,3 @@ func (g *genericScheduler) assignReplicas(
...
-	klog.V(4).Infof("Reserved bindingClusters : %v", reservedClusters.List())
-	klog.V(4).Infof("Candidate bindingClusters: %v", candidateClusters.List())
-
-	// TODO: should schedule as much as possible?
-	deltaLen := len(spec.Clusters) - len(reservedClusters)
-	if len(candidateClusters) < deltaLen {
-		// for ReplicaSchedulingTypeDivided, we will try to migrate replicas to the other health clusters
-		if placement.ReplicaScheduling == nil || placement.ReplicaScheduling.ReplicaSchedulingType == policyv1alpha1.ReplicaSchedulingTypeDuplicated {
-			klog.Warningf("ignore reschedule binding as insufficient available cluster")
-			return ScheduleResult{}, nil
```
dddddai, line 281: Let's say we have 2 clusters [A、B] (with `Failover` enabled)

1. a rb was scheduled to [A、B] (duplicated)
2. cluster A becomes not ready
3. the orignal behavior: scheduled to [A、B] while the current behavior: scheduled to [B]

I'm not sure if we want to keep the original behavior, if not then lgtm

pkg/scheduler/core/division_algorithm.go
dddddai: Could you please give an example? It seems not possible to happen "in one event"

pkg/scheduler/core/division_algorithm.go
```diff
@@ -34,10 +34,13 @@ func divideReplicasByResource(
 	spec *workv1alpha2.ResourceBindingSpec,
 	preference policyv1alpha1.ReplicaDivisionPreference,
 ) ([]workv1alpha2.TargetCluster, error) {
-	// Step 1: Get previous total sum of replicas.
-	assignedReplicas := util.GetSumOfReplicas(spec.Clusters)
+	// Step 1: Find the ready clusters that have old replicas
+	scheduledClusters := findOutScheduledCluster(spec.Clusters, clusters)
```
Garrybest, line 38: What if we just delete the unjoined clusters from `spec.clusters`?

Then the replicas in unjoined clusters will be considered as a special kind of scaling up.

Garrybest: Seems we don't need more changes, am I right?

pkg/scheduler/core/division_algorithm.go
merryzhou: Let's say we have 2 clusters [A、B] , both with allocatable resources: 4C8G
a deployment requests: 2C4G, 3 replicas
and the propagationPolicy is:
```
placement:
    clusterAffinity:
      clusterNames:
      - A
      - B
    replicaScheduling:
      weightPreference:
        dynamicWeight: AvailableReplic
```
so the deployment was placed in both A and B, let's assume A has 2 replicas and B has 1。

1. update propagationPolicy placement.clusterAffinity.clusterNames=[A], now no enough resources for this deployment to reschedule, so nothing happened;
2. update deployment.spec.replicas=2, with this pr, nothing happened again..

huone1: you are right that it should return the scheduledClusters as the result

pkg/scheduler/core/division_algorithm.go
huone1: it is a  special kind of scaling up

pkg/scheduler/core/generic_scheduler.go
huone1: @mrlihanbo Please see if this change is reasonable.

Garrybest:
I think there is a bug here. Imagine: 

1. `Failover` is disabled.
2. A cluster is not ready.

We should not erase the scheduling replicas of the not-ready cluster in RB because `Failover` is disabled. However, if it happens to scale up, e.g. the user scale up desired replicas. This PR will delete all replicas in this cluster which is dangerous. PTAL, @huone1

huone1:
when Failover is disabled, the rescheduling for  not-ready cluster is not triggered

Garrybest:
I'm afraid not. When scaling up, rescheduling will be triggered.

huone1:
> I'm afraid not. When scaling up, rescheduling will be triggered.

what is wrong with it

RainbowMango:
> However, if it happens to scale up, e.g. the user scale up desired replicas. This PR will delete all replicas in this cluster which is dangerous.

Why is it dangerous? 

What I'm thinking about is how to postpone the deletion operation until the desired replicas are all in the available state. That guarantees there are always sufficient replicas running at any time.

Garrybest:
`Failover` is disabled, but under this circumstance all replicas will be removed. It does not match the expectation.

Garrybest:
`Failover` is disabled because the user does not want to remove all replicas when a cluster is not ready. If the api-server of a member cluster is temporarily down and `Failover` is disabled, we potentially do not want the replicas in this member cluster to be evicted when the user triggers scaling up. However, now it does not match the expectation.

RainbowMango:
Agree with @Garrybest.
We should consider the `false positive` cluster failure `seriously`.  Can we hold the replicas(for the un-healthy cluster) unchanged even in the case of scaling up and decreasing the replicas in the case of scaling down.

huone1:
it is different between deleting a cluster and the change from healthy to unhealthy。it is reasonable to migrate the replicas  to Other Clusters when deleting a cluster

RainbowMango:
Yeah, deleting a cluster is another story.

Garrybest:
Let me describe an example.

1. We have a deployment with a 5 desired replicas which are propagated to member cluster A.
2. The cluster is a little bit busy so the connection between karmada-agent and kube-apiserber of member A is lost for about 1 minute. So the cluster status is unhealthy.
3. It happens that the deployment has been scaled up to 10 replicas at this time.
4. Even if we disable `Failover`, the scaling up triggers the scheduling procedure, now the karmada-scheduler will remove replicas in cluster A which may be dangerous. It is possible that there is nothing wrong with the 5 replicas in member cluster A.

RainbowMango:
Thanks @Garrybest for the details. Can you help file an issue to track this?

Garrybest:
Sure.

PR #1388, opened 2022-02-22
fix bugs of resource computing

Garrybest:

**What type of PR is this?**
/kind bug

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/util/resource.go
+++ b/pkg/util/resource.go
@@ -105,7 +105,7 @@ func (r *Resource) Sub(rl corev1.ResourceList) error {
 			if resourcehelper.IsScalarResourceName(rName) {
 				rScalar, ok := r.ScalarResources[rName]
 				scalar := rQuant.Value()
-				if !ok {
+				if !ok && scalar > 0 {
 					return fmt.Errorf("scalar resources %s does not exist, got %d", rName, scalar)
 				}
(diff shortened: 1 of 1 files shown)


pkg/util/resource.go
```diff
@@ -105,7 +105,7 @@ func (r *Resource) Sub(rl corev1.ResourceList) error {
 			if resourcehelper.IsScalarResourceName(rName) {
 				rScalar, ok := r.ScalarResources[rName]
 				scalar := rQuant.Value()
-				if !ok {
+				if !ok && scalar > 0 {
```
RainbowMango, line 108: Is there a case that `scalar==0`?

RainbowMango: In case of the scalar resource doesn't exist, then `rScalar == 0`:
- If scalar > 0, it won't pass the following check
- If scalar == 0, no-opts.

So, what's wrong?

Garrybest: Scalar could be zero. It comes from podSpec, e.g., "nvidia.com/gpu = 0". If the user request empty resource even if nodes does not contains, the pod could still be scheduled by kube-scheduler.

RainbowMango (commented):
/lgtm
/approve

PR #1395, opened 2022-02-24
add default dependenciesInterpreter framework

mrlihanbo:

**What type of PR is this?**
/kind feature

add default dependenciesInterpreter framework

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/resourceinterpreter/defaultinterpreter/default.go
+++ b/pkg/resourceinterpreter/defaultinterpreter/default.go
@@ -18,6 +18,7 @@ type DefaultInterpreter struct {
 	reviseReplicaHandlers   map[schema.GroupVersionKind]reviseReplicaInterpreter
 	retentionHandlers       map[schema.GroupVersionKind]retentionInterpreter
 	aggregateStatusHandlers map[schema.GroupVersionKind]aggregateStatusInterpreter
+	dependenciesHandlers    map[schema.GroupVersionKind]dependenciesInterpreter
 }
 
 // NewDefaultInterpreter return a new DefaultInterpreter.
@@ -27,6 +28,7 @@ func NewDefaultInterpreter() *DefaultInterpreter {
 		reviseReplicaHandlers:   getAllDefaultReviseReplicaInterpreter(),
 		retentionHandlers:       getAllDefaultRetentionInterpreter(),
 		aggregateStatusHandlers: getAllDefaultAggregateStatusInterpreter(),
+		dependenciesHandlers:    getAllDefaultDependenciesInterpreter(),
 	}
 }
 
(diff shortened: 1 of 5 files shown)


pkg/resourceinterpreter/defaultinterpreter/default.go
```diff
@@ -18,6 +18,7 @@ type DefaultInterpreter struct {
 	reviseReplicaHandlers   map[schema.GroupVersionKind]reviseReplicaInterpreter
 	retentionHandlers       map[schema.GroupVersionKind]retentionInterpreter
 	aggregateStatusHandlers map[schema.GroupVersionKind]aggregateStatusInterpreter
+	dependenciesHandler     map[schema.GroupVersionKind]dependenciesInterpreter
```
iawia002, line 21:
```suggestion
	dependenciesHandlers    map[schema.GroupVersionKind]dependenciesInterpreter
```

iawia002 (commented):
Basically LGTM, just one comment

pkg/resourceinterpreter/defaultinterpreter/default.go
RainbowMango: Nice catch!!
+1

mrlihanbo: > 
Done! Thanks for your review.

pkg/resourceinterpreter/interpreter.go
```diff
@@ -153,3 +155,22 @@ func (i *customResourceInterpreterImpl) AggregateStatus(object *unstructured.Uns
 
 	return i.defaultInterpreter.AggregateStatus(object, aggregatedStatusItems)
 }
+
+// GetDependencies returns the dependencies' info of the given object.
```
RainbowMango, line 159: comments.

pkg/resourceinterpreter/defaultinterpreter/default.go
```diff
@@ -92,3 +97,12 @@ func (e *DefaultInterpreter) AggregateStatus(object *unstructured.Unstructured,
 	}
 	return handler(object, aggregatedStatusItems)
 }
+
+// GetDependencies returns the dependencies' info of the given object.
```
RainbowMango, line 101: comments...

pigletfly:
do we want to support customized depend interperter? For example, we have a crd's cr which will create another crd's cr.
```
apiVersion: xxx
kind: XXX
metadata:
  name: abc
spec:
  disk:
     - size: 10
```
CRD XXX has a disk in spec, which is another CRD that should be created when XXX is created.

mrlihanbo:
> do we want to support customized depend interperter? For example, we have a crd's cr which will create another crd's cr.
> 
> ```
> apiVersion: xxx
> kind: XXX
> metadata:
>   name: abc
> spec:
>   disk:
>      - size: 10
> ```
> 
> CRD XXX has a disk in spec, which is another CRD that should be created when XXX is created.

For now, dependenciesInterpreter will be used in `dependncies distributor`. CRD XXX has a disk in spec, if XXX is propagated to cluster A, B, C, the disk will be automatically propagated to cluster A, B, C.

XiShanYongYe-Chang (commented):
/lgtm

RainbowMango (approved):
/approve

PR #1418, opened 2022-03-01
Code optimization and add unit test 

AllenZMC:
**What type of PR is this?**

Code optimization and add unit test 
Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/util/binding.go
+++ b/pkg/util/binding.go
@@ -30,10 +30,7 @@ func IsBindingReplicasChanged(bindingSpec *workv1alpha2.ResourceBindingSpec, str
 		return false
 	}
 	if strategy.ReplicaSchedulingType == policyv1alpha1.ReplicaSchedulingTypeDivided {
-		replicasSum := int32(0)
-		for _, targetCluster := range bindingSpec.Clusters {
-			replicasSum += targetCluster.Replicas
-		}
+		replicasSum := GetSumOfReplicas(bindingSpec.Clusters)
 		return replicasSum != bindingSpec.Replicas
 	}
 	return false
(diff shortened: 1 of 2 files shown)


RainbowMango:
```
pkg/util/binding_test.go:6: File is not `goimports`-ed with -local github.com/karmada-io/karmada (goimports)
```

pkg/util/binding.go
```diff
@@ -42,8 +39,8 @@ func IsBindingReplicasChanged(bindingSpec *workv1alpha2.ResourceBindingSpec, str
 // GetSumOfReplicas will get the sum of replicas in target clusters
 func GetSumOfReplicas(clusters []workv1alpha2.TargetCluster) int32 {
 	replicasSum := int32(0)
-	for i := range clusters {
-		replicasSum += clusters[i].Replicas
+	for _, targetCluster := range clusters {
+		replicasSum += targetCluster.Replicas
```
Garrybest, line 43: We use `for i := range clusters` in order to avoid copying the `workv1alpha2.TargetCluster` in this loop.

XiShanYongYe-Chang: The problem may not occur in all cases. For details, can see here: https://github.com/golang/go/wiki/CommonMistakes#using-reference-to-loop-iterator-variable

AllenZMC:
> When you squash commit occurs problem, you can do as:
> 
> ```
> 
> git reset HEAD~1 # 1 indicates how many commits you want to reback
> 
> git add .
> 
> git commit -s -m "<commit message>"
> 
> git push origin <branch-name> -f
> 
> ```

OK，Thanks

PR #1433, opened 2022-03-04
[helm chart] make rolling update strategy configurable

pigletfly:

**What type of PR is this?**
/kind cleanup

It's better to set `maxUnavailable` to 0

Fixes #

```release-note
    NONE
```

### DIFF
--- a/charts/templates/karmada-aggregated-apiserver.yaml
+++ b/charts/templates/karmada-aggregated-apiserver.yaml
@@ -13,6 +13,10 @@ spec:
     matchLabels:
       {{- include "karmada.aggregatedApiserver.labels" . | nindent 6}}
   replicas: {{ .Values.aggregatedApiServer.replicaCount }}
+  {{- with .Values.aggregatedApiServer.strategy }}
+  strategy:
+  {{- toYaml . | nindent 8 }}
+  {{- end }}
   template:
     metadata:
       {{- with .Values.aggregatedApiServer.podAnnotations }}
--- a/charts/templates/karmada_agent.yaml
+++ b/charts/templates/karmada_agent.yaml
@@ -70,6 +70,10 @@ spec:
   selector:
     matchLabels:
       {{- include "karmada.agent.labels" . | indent 6}}
+  {{- with .Values.agent.strategy }}
+  strategy:
+  {{- toYaml . | nindent 8 }}
+  {{- end }}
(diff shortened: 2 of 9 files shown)


RainbowMango:
> It's better to set maxUnavailable to 0

Why?

I'd like to invite @fleeto for comments. (I don't have much experience  :( ）

fleeto:
/lgtm

This configuration will require rolling updates by creating a new version of the Pod first and deleting the old one later, which should improve availability during rolling updates

RainbowMango:
Thanks @fleeto .
I agree that this will improve availability during rolling updates. But my question is if this is suitable to be the default value?

My concern is if it will block rolling updates in case pods has some `anti-affinity` rules. E.g.
There is a small cluster with 3 nodes and 3 pods running on the cluster, each node running a pod, the pods have the `podAntiAffinity` specification as follows:
https://github.com/karmada-io/karmada/blob/4f9c3ca90cb9ebbfbf21c0a7c7548ddb472609cf/artifacts/deploy/karmada-apiserver.yaml#L23-L32

I'm afraid the new pods can't be scheduled as the `podAntiAffinity`.

It's not a big deal, how do you say guys?

fleeto:
> 

O I've got the point.

pigletfly:
> Thanks @fleeto . I agree that this will improve availability during rolling updates. But my question is if this is suitable to be the default value?
> 
> My concern is if it will block rolling updates in case pods has some `anti-affinity` rules. E.g. There is a small cluster with 3 nodes and 3 pods running on the cluster, each node running a pod, the pods have the `podAntiAffinity` specification as follows:
> 
> https://github.com/karmada-io/karmada/blob/4f9c3ca90cb9ebbfbf21c0a7c7548ddb472609cf/artifacts/deploy/karmada-apiserver.yaml#L23-L32
> 
> I'm afraid the new pods can't be scheduled as the `podAntiAffinity`.
> 
> It's not a big deal, how do you say guys?

in karmada Helm charts, there is no default `affinity`.The  `karmada/artifacts/` folder is used by installing via binary.Or we can provide this as an option in values.yaml, then users can modify the rollingUpdate strategy.

RainbowMango:
/retitle [helm chart] make rolling update strategy configurable

RainbowMango (approved):
/approve

PR #1447, opened 2022-03-08
Don't retain service nodeport when updating service

### LINKED ISSUE #1445: Ensure that the user can modify the service nodeport.
When users create a service with an assigned nodeport:
```
apiVersion: v1
kind: Service
metadata:
  name: my-service
spec:
  type: NodePort
  selector:
    app: MyApp
  ports:
    - port: 80
      targetPort: 80
      nodePort: 30007
```
And then users want to update service nodeport, this action will be failed, because karmada retain service nodeport:
https://github.com/karmada-io/karmada/blob/4f9c3ca90cb9ebbfbf21c0a7c7548ddb472609cf/pkg/resourceinterpreter/defaultinterpreter/retain.go#L118-L154

And when users create a service without an assigned nodeport, `karmada-apiserver` will assign a nodeport to the service automatically. Then if users want to add a node port to the service, this action will also be failed.

So how about removing `retainServiceNodePort` function when updating the service.

XiShanYongYe-Chang:

**What type of PR is this?**

/kind feature

Ensure that the user can update the service nodeport, so don't retain service nodeport.

Fixes #1445 

```release-note
`karmada-controller-manager`: Don't retain service nodeport when updating service.
```

### DIFF
--- a/pkg/resourceinterpreter/defaultinterpreter/retain.go
+++ b/pkg/resourceinterpreter/defaultinterpreter/retain.go
@@ -75,12 +75,8 @@ func retainServiceFields(desired, observed *unstructured.Unstructured) (*unstruc
 		return nil, err
 	}
 
-	// ClusterIP and NodePort are allocated to Service by cluster, so retain the same if any while updating
+	// ClusterIP is allocated to Service by cluster, so retain the same if any while updating
 	if err := retainServiceClusterIP(desired, observed); err != nil {
-		return nil, err
-	}
-
-	if err := retainServiceNodePort(desired, observed); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


pigletfly:
is there a specific reason why we retain Service NodePort ?

XiShanYongYe-Chang:
> is there a specific reason why we retain Service NodePort ?

Before we lifted from the kubefed retain logic:

https://github.com/kubernetes-sigs/kubefed/blob/9631566a2451dab3d4fc6a9901b1cdebd2511323/pkg/controller/sync/dispatch/retain.go#L60

RainbowMango:
cc @mrlihanbo

RainbowMango (approved):
/approve

Issue #112, opened 2021-01-07
Use github package to host container images


kevin-wangzefeng:
**What would you like to be added**:
Github package now offers container images hosting, we can upload container images of karmada components to github package

**Why is this needed**:

RainbowMango:
Refer to https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry

RainbowMango:
/close
As we now use DockerHub to host both the `latest` and `released` images.

PR #1459, opened 2022-03-09
1. build images for arm64

fleeto:
1. build images for arm64

2. push images into dockerhub

Signed-off-by: Vincent Cui <dustise@live.com>

Secrets: 

REPO:  Prefix for image name
REPO_USERNAME: dockerhub username
REPO_PASSWORD: dockerhub password

Fixes https://github.com/karmada-io/karmada/issues/112

```release-note
NONE
```

### DIFF
--- a/.github/workflows/release-image.yml
+++ b/.github/workflows/release-image.yml
@@ -6,25 +6,40 @@ on:
 jobs:
   release-image:
     name: release images
-    # prevent job running from forked repository, otherwise
-    # 1. running on the forked repository would fail as missing necessary secret.
-    # 2. running on the forked repository would use unnecessary GitHub Action time.
-    if: ${{ github.repository == 'karmada-io/karmada' }}
+    strategy:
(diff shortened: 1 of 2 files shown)


fleeto:
/hold

For the dockerhub and secrets.

btw, manifests should be updated accordingly

RainbowMango:
I think we should keep uploading images to SWR for a while(for compatibility), and start uploading images to Docker Hub.

So, I suggest creating another workflow, like `release-image-to-dockerhub.yml` and after about 2 releases, we can remove the original workflow.

RainbowMango:
I think we should keep uploading images to SWR for a while(for compatibility), and start uploading images to Docker Hub.

So, I suggest creating another workflow, like `release-image-to-dockerhub.yml` and after about 2 releases, we can remove the original workflow.

fleeto:
That’s a great idea, I will work on it later

发自我的iPhone

PR #1462, opened 2022-03-09
accelerate merge clusters during schedule and improve test coverage

AllenZMC:
**What type of PR is this?**
/kind cleanup

reduce unnecessary code operations in some cases and improve test coverage
Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/util/binding.go
+++ b/pkg/util/binding.go
@@ -87,6 +87,12 @@ func DivideReplicasByTargetCluster(clusters []workv1alpha2.TargetCluster, sum in
 
 // MergeTargetClusters will merge the replicas in two TargetCluster
 func MergeTargetClusters(old, new []workv1alpha2.TargetCluster) []workv1alpha2.TargetCluster {
+	switch {
+	case len(old) == 0:
+		return new
+	case len(new) == 0:
+		return old
+	}
 	// oldMap is a map of the result for the old replicas so that it can be merged with the new result easily
 	oldMap := make(map[string]int32)
 	for _, cluster := range old {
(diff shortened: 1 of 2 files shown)


pkg/util/binding.go
```diff
@@ -87,6 +87,12 @@ func DivideReplicasByTargetCluster(clusters []workv1alpha2.TargetCluster, sum in
 
 // MergeTargetClusters will merge the replicas in two TargetCluster
 func MergeTargetClusters(old, new []workv1alpha2.TargetCluster) []workv1alpha2.TargetCluster {
+	switch {
```
XiShanYongYe-Chang, line 90: IMOP, this logic may seem a little redundant, after all, the following logic already overrides this judgment.

XiShanYongYe-Chang (commented):
Other parts LGTM

pkg/util/binding.go
RainbowMango: This is the short path.
If we need a short patch depends on :
1. How often does the short patch hit? 
2. The following logic is complicated and most cases don't need to be involved.

cc @Garrybest for comment

Garrybest: When first scheduled, the `old` cluster must be empty. So I think this patch is useful.

RainbowMango:
/retitle accelerate merge clusters during schedule and improve test coverage
/lgtm
/approve

pkg/util/binding.go
RainbowMango: ok. here we go.

PR #1477, opened 2022-03-11
replace update method with patch method when updating schedule result

mrlihanbo:
**What type of PR is this?**
/kind cleanup

Replace update method with patch method when updating schedule result. 
Conflict error will happen in updating schedule result due to the resource binding controller may update binding's status frequently.

Fixes #

```release-note
karmada-scheduler: replace update method with patch method when updating schedule result
```

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -6,10 +6,12 @@ import (
 	"fmt"
 	"time"
 
+	jsonpatch "github.com/evanphx/json-patch/v5"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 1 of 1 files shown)


Garrybest:
Well, if conflicts occur frequently, I think it's better to use `patch` instead of `update`. When api-server receives a `patch` request, it will get the latest object from etcd and patch the object. So the possibility of conflicts will decrease.

RainbowMango:
That makes sense.

mrlihanbo:
> Well, if conflicts occur frequently, I think it's better to use `patch` instead of `update`. When api-server receives a `patch` request, it will get the latest object from etcd and patch the object. So the possibility of conflicts will decrease.

Sounds good to me!

Garrybest (approved):
Well done.
/lgtm

RainbowMango (approved):
/approve

I wonder if we can use the patch to update `.status` ?

PR #1107, opened 2021-12-14
update kube-apiserver and kube-controller-manager default version

Shike-Ada:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:
Upgrade k8s version to v1.21.7
**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:
The latest version of K8s image with v1.21.x is 7.
**Does this PR introduce a user-facing change?**:

```release-note

```

PR #1486, opened 2022-03-15
[CI] Support run CI with multi version kubernetes

XiShanYongYe-Chang:

**What type of PR is this?**

/kind feature

Support run CI with multi version kubernetes.

Fixes #

Current kubernetes versions:
- v1.21
- v1.22
- v1.23

```release-note
Support run CI with multi version kubernetes
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,8 +80,9 @@ jobs:
     name: e2e test
     needs: build
     runs-on: ubuntu-18.04
-    env:
-      ARTIFACTS_PATH: ${{ github.workspace }}/karmada-e2e-logs/
+    strategy:
+      matrix:
+        k8s: [v1.21.10, v1.22.7, v1.23.4]
     steps:
       - name: checkout code
         uses: actions/checkout@v2
@@ -90,12 +91,16 @@ jobs:
         with:
           go-version: 1.17.x
       - name: setup e2e test environment
-        run: hack/local-up-karmada.sh
+        run: |
+          export CLUSTER_VERSION=kindest/node:${{ matrix.k8s }}
+          hack/local-up-karmada.sh
       - name: run e2e
-        run: hack/run-e2e.sh
+        run: |
+          export ARTIFACTS_PATH=${{ github.workspace }}/karmada-e2e-logs/${{ matrix.k8s }}/
+          hack/run-e2e.sh
       - name: upload logs
         if: always()
         uses: actions/upload-artifact@v2
         with:
-          name: karmada_e2e_log
-          path: ${{ github.workspace }}/karmada-e2e-logs/
+          name: karmada_e2e_log_${{ matrix.k8s }}
+          path: ${{ github.workspace }}/karmada-e2e-logs/${{ matrix.k8s }}/


.github/workflows/ci.yml
```diff
@@ -76,12 +76,13 @@ jobs:
...
-    name: e2e test
+  kubernetes:
+    name: Kubernetes
     needs: build
     runs-on: ubuntu-18.04
-    env:
-      ARTIFACTS_PATH: ${{ github.workspace }}/karmada-e2e-logs/
+    strategy:
+      matrix:
+        k8s: [v1.21.2, v1.22.0, v1.23.0]
```
pigletfly, line 85: the minimum kubernetes version in compat matrix is 1.15, should we add it as well ?

pigletfly (changes requested):

.github/workflows/ci.yml
XiShanYongYe-Chang: Yes, plan to do so, the current three k8s versions test ok, the others I will test and add later.

XiShanYongYe-Chang:
Kubernetes v1.20 doesn't use `v1beta1 EndpointSlice`, for detail: https://github.com/karmada-io/karmada/pull/1107#issuecomment-997159415.

So we may need to disable MCS-related E2E under kubernetes v1.21. We can do this in the later pull request.

.github/workflows/ci.yml
RainbowMango: That depends, we've done testing manually against v1.15, but it will be complicated to run it on CI, because if we can't select v1.15 as host cluster (no CRD@v1 support). 

So, my opinion is, to support as much as we can for now with limited effort.

RainbowMango:
I'm still thinking about how to select the version of Kubernetes for the matrix,  take v1.23 as an example, v1.23.0 should be OK and makes little difference between the following patch versions. Do we need to catch up with the latest version?

RainbowMango:
> should we update the compat matrix ?

Documenting this limitation is good, but how and where?

.github/workflows/ci.yml
```diff
@@ -76,12 +76,13 @@ jobs:
           go-version: 1.17.x
       - name: make test
         run: make test
-  e2e:
-    name: e2e test
+  kubernetes:
+    name: Kubernetes
```
RainbowMango, line 80: Why change the job name?
I'd prefer the previous name, like:
- CI Workflow / e2e test(vx.y.z) (pull_request)

.github/workflows/ci.yml
```diff
@@ -76,12 +76,13 @@ jobs:
...
     needs: build
     runs-on: ubuntu-18.04
-    env:
-      ARTIFACTS_PATH: ${{ github.workspace }}/karmada-e2e-logs/
+    strategy:
+      matrix:
+        k8s: [v1.21.10, v1.22.7, v1.23.4]
```
RainbowMango, line 85: Should we always use the latest patch release?

XiShanYongYe-Chang: Personally, I don't think this is necessary. As Kind keeps building new image versions of k8s, updates will be fast. For us, we need to make sure that the Kubernetes minor version number is feasible.

RainbowMango (approved):
/lgtm
/approve

PR #1512, opened 2022-03-21
Fixed the `FullyApplied` condition of `RB/CRB` mislabeling issue in case of non-scheduled.

### LINKED ISSUE #1511: the status FULLYAPPLIED in the resource binding  is true while the resource obj is not scheduled successfully
**What happened**:
the rb status FULLYAPPLIED is true while the resource obj is not scheduled successfully
the status FULLYAPPLIED in the resource binding  is  false
it can apply the following yaml to reproduce it

the workload yaml
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx
  labels:
    app: nginx
spec:
  replicas: 10
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - image: nginx
        name: nginx
```

the  pp yaml select a cluster which is no existed
```
kind: PropagationPolicy
metadata:
  name: nginx-propagation
spec:
  resourceSelectors:
    - apiVersion: apps/v1
      kind: Deployment
      name: nginx
  placement:
    clusterAffinity:
      clusterNames:
        - none
```

the result is 
image

**Environment**:
- Karmada version:
- Others:

huone1:

**What type of PR is this?**
/kind bug

add the Judgment Condition **len(clusterNames) != 0** to fix #1511
Fixes #1511

```release-note
`karmada-controller-manager`: Fixed the `FullyApplied` condition of `ResourceBinding/ClusterResourceBinding` mislabeling issue in case of non-scheduled. 
```

### DIFF
--- a/pkg/util/helper/workstatus.go
+++ b/pkg/util/helper/workstatus.go
@@ -12,6 +12,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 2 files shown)


pigletfly:
can we add a e2e test for this ?

RainbowMango:
Yes, we can. But I think a unit test is also just fine.

XiShanYongYe-Chang:
> Yes, we can. But I think a unit test is also just fine.

+1, I think we can move all judgment into `areWorksFullyApplied` function and add a UT for `areWorksFullyApplied`.

XiShanYongYe-Chang (commented):
/lgtm

XiShanYongYe-Chang (commented):
/lgtm

RainbowMango (approved):
/approve

/hold
for @pigletfly to confirm.

pigletfly:
good catch!

PR #1562, opened 2022-03-30
[fix] Add a log reminder for creating a cluster

my-git9:
**What type of PR is this?**
/kind cleanup

When I used hack/local-up-karmada.sh to install karmada for the first time in local, the k8s cluster was not created due to the image. Since there was no reminder and I didn’t know where the log file was, I had to look at the docker log for query reasons. I think it is necessary to remind the user in the script where the cluster creation log is located


RainbowMango (approved):
/lgtm
/approve

Well done!!! I also noticed the issue but don't have time to do it.
@my-git9 Thanks and welcome!!!

Issue #1330, opened 2022-01-28
[lfx-mentorship] refactor the scheduler framework


RainbowMango:
**What would you like to be added**:
Refactor the framework of `karmada-scheduler` to make it easier to extend and adopt more scheduling policies.

**Why is this needed**:

Refer to:
- https://github.com/karmada-io/karmada/issues/1195

kerthcet:
I'm glad to follow this feature, should I have to apply for the mentorship? @RainbowMango

halfrost:
I applied for the mentorship program of this project in LFX, clicked the apply button and it was successful. I thought I needed to fill in some information. So I suspected that the application was not successful, I withdrew the application, but when I reapplied, the LFX system told me "You already applied to this program.". I'm just curious whether my application is successful or not? @kevin-wangzefeng

huone1:
it is the schedule refactor proposal : https://docs.google.com/document/d/1yZfw50F7uu9WG_dZ0M7-jbsmlidPIlMmDSI3AA4eWRg/edit?usp=sharing
it is just a draft version, so  you can give any suggestions and ideas.

kerthcet:
It's a great start. I read the document overall and left several  comments there. Maybe not mature but make sense to me. Thanks @huone1

RainbowMango:
/close
as finished.

PR #1563, opened 2022-03-30
[optimize]Consider available resources when selecting clusters

huone1:

**What type of PR is this?**

/kind feature

consider available resources when selecting clusters.
when spread constraint is setted, the available resources may be insufficient for workload currently.

For example

cluster constraint minGroup=maxGroup=2   needReplicas = 80，
the cluster information is following as
|clusterName|member1|member2|member3|
|-----|----|-----|-----|
| score| 60 | 50| 40|
| AvailableReplicas| 40 | 30|60

Although member1 and member2 have higher scores，the sum of AvailableReplicas is less than needReplicas，
so the result is member1 and member3
a part of https://github.com/karmada-io/karmada/issues/1330

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/core/spreadconstraint/select_clusters.go
+++ b/pkg/scheduler/core/spreadconstraint/select_clusters.go
@@ -3,53 +3,64 @@ package spreadconstraint
 import (
 	"fmt"
 
+	"k8s.io/klog/v2"
+
 	clusterv1alpha1 "github.com/karmada-io/karmada/pkg/apis/cluster/v1alpha1"
 	policyv1alpha1 "github.com/karmada-io/karmada/pkg/apis/policy/v1alpha1"
 )
 
 // SelectBestClusters selects the cluster set based the GroupClustersInfo and placement
-func SelectBestClusters(placement *policyv1alpha1.Placement, groupClustersInfo *GroupClustersInfo) ([]*clusterv1alpha1.Cluster, error) {
-	if len(placement.SpreadConstraints) != 0 {
-		return selectBestClustersBySpreadConstraints(placement.SpreadConstraints, groupClustersInfo)
+func SelectBestClusters(placement *policyv1alpha1.Placement, groupClustersInfo *GroupClustersInfo, needReplicas int32) ([]*clusterv1alpha1.Cluster, error) {
+	if len(placement.SpreadConstraints) == 0 || shouldIgnoreSpreadConstraint(placement) {
+		var clusters []*clusterv1alpha1.Cluster
+		for _, cluster := range groupClustersInfo.Clusters {
+			clusters = append(clusters, cluster.Cluster)
+		}
+		klog.V(4).Infof("select all clusters")
+		return clusters, nil
 	}
 
-	var clusters []*clusterv1alpha1.Cluster
-	for _, cluster := range groupClustersInfo.Clusters {
-		clusters = append(clusters, cluster.Cluster)
+	if shouldIgnoreAvailableResource(placement) {
+		needReplicas = InvalidReplicas
 	}
 
-	return clusters, nil
+	return selectBestClustersBySpreadConstraints(placement.SpreadConstraints, groupClustersInfo, needReplicas)
 }
 
 func selectBestClustersBySpreadConstraints(spreadConstraints []policyv1alpha1.SpreadConstraint,
-	groupClustersInfo *GroupClustersInfo) ([]*clusterv1alpha1.Cluster, error) {
+	groupClustersInfo *GroupClustersInfo, needReplicas int32) ([]*clusterv1alpha1.Cluster, error) {
 	if len(spreadConstraints) > 1 {
 		return nil, fmt.Errorf("just support single spread constraint")
 	}
 
 	spreadConstraint := spreadConstraints[0]
 	if spreadConstraint.SpreadByField == policyv1alpha1.SpreadByFieldCluster {
-		return selectBestClustersByCluster(spreadConstraint, groupClustersInfo)
+		return selectBestClustersByCluster(spreadConstraint, groupClustersInfo, needReplicas)
 	}
 
 	return nil, fmt.Errorf("just support cluster spread constraint")
 }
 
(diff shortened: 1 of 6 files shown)


RainbowMango:
assign it to me once it ready for review.

pkg/scheduler/core/spreadconstraint/select_clusters.go
```diff
@@ -46,10 +52,99 @@ func selectBestClustersByCluster(spreadConstraint policyv1alpha1.SpreadConstrain
...
+
+// if needClusterCount = 2, needReplicas = 80, member1 and member3 will be selected finally.
+// because the total resource of member1 and member2 is less than needReplicas although their scores is highest
+// --------------------------------------------------
+// | clusterName      | member1 | member2 | member3 |
+// |-------------------------------------------------
+// | score            |   60    |    50   |    40   |
+// |------------------------------------------------|
+// |AvailableReplicas |   40    |    30   |    60   |
+// |------------------------------------------------|
```
RainbowMango, line 76: This is an algorithm problem, can we abstract a common function to implement it and decoupling with the business logic?
@halfrost Would you like comment on it?

halfrost: First, let's discuss an abstract model for this problem. We want to find the maximum score that satisfies the following 2 conditions:
- 1. total availableReplicas >= needReplicas, and the smaller the total availableReplicas, the better, the less remaining. eg, needReplicas = 100, total availableReplicas = 101 is better than total availableReplicas = 110.
- 2. Only needClusterCount servers can be selected.

Your current solution does find a solution, but your solution is an approximate optimal solution, not a globally optimal solution. E.g, needClusterCount = 2, needReplicas = 80

```
-----------------------------------------------------------------------
| clusterName      | member1 | member2 | member3 |  member4 | member5 |
|----------------------------------------------------------------------
| score            |   60    |    50   |    40   |    40   |    40    |
|---------------------------------------------------------------------|
|AvailableReplicas |   40    |    30   |    35   |    100   |    40   |
|---------------------------------------------------------------------|
```

In this case, your code logic will find member 1 and member 4. score = 100, total availableReplicas = 140, but this is not a globally optimal solution, the optimal solution is member 1 and member 5 . score = 100, total availableReplicas = 80.

Then consider the time complexity of your solution. For your solution, the time complexity is O(m * n^2), where m is the number of times that selectClustersByAvailableResource function is called. n is the total number of candidateClusters.

I don't know whether you need to find an approximate optimal solution, or you must find a globally optimal solution.

If you must want to find the global optimal solution, this question should actually be answered with a two-dimensional 01 knapsack. The normal 01 backpack background you should already know. Here I will talk about the corresponding deformation of this question.

There are 2 conditions that need to be met here, the first condition is to find the minimum value larger than V. It can be known from the title condition that the value range larger than V is [v, total availableReplicas sum], and the second condition is to take needClusterCount items. We can use a dp array, dp[0...total sum][total count of cluster] to count all states.

Regarding time complexity, assuming that the number of clusters is a constant variable. At the beginning, calculate the dp data once when it is initialized, and the time complexity is O(n^2). After that, each query is to perform a for loop on the rows in this two-dimensional array. Start from v, find the first solution larger than v, and column = needClusterCount, which is the optimal solution. The final time complexity is O(n^2 + m*n), where m is the number of times that selectClustersByAvailableResource function is called. n is the total number of candidateClusters.

If the number of cluster does not change frequently, the above method is the best solution. If the number of cluster changes frequently, maybe need other optimization solution.

huone1: here needs to find an approximate optimal solution.
1. the sum of selected availableReplicas >= needReplicas, but don't consider whether the total availableReplicas is small
   so E.g, needClusterCount = 2, needReplicas = 80
   the {member1, member4} is equal to {member1, member5}

halfrost: If you are only looking for approximate optimal solutions. Your solution can achieve this goal.

RainbowMango: My concern is the implementation is hard to understand.  Is there room to optimize?

pkg/scheduler/core/spreadconstraint/select_clusters.go
```diff
@@ -3,53 +3,64 @@ package spreadconstraint
...
 	clusterv1alpha1 "github.com/karmada-io/karmada/pkg/apis/cluster/v1alpha1"
 	policyv1alpha1 "github.com/karmada-io/karmada/pkg/apis/policy/v1alpha1"
 )
 
 // SelectBestClusters selects the cluster set based the GroupClustersInfo and placement
-func SelectBestClusters(placement *policyv1alpha1.Placement, groupClustersInfo *GroupClustersInfo) ([]*clusterv1alpha1.Cluster, error) {
-	if len(placement.SpreadConstraints) != 0 {
-		return selectBestClustersBySpreadConstraints(placement.SpreadConstraints, groupClustersInfo)
+func SelectBestClusters(placement *policyv1alpha1.Placement, groupClustersInfo *GroupClustersInfo, needReplicas int32) ([]*clusterv1alpha1.Cluster, error) {
+	if len(placement.SpreadConstraints) == 0 || isSpreadConstraintIgnored(placement) {
```
RainbowMango, line 14:
```suggestion
	if len(placement.SpreadConstraints) == 0 || shouldIgnoreSpreadConstraint(placement) {
```

pkg/scheduler/core/spreadconstraint/select_clusters.go
```diff
@@ -3,53 +3,64 @@ package spreadconstraint
...
+			clusters = append(clusters, cluster.Cluster)
+		}
+		klog.V(4).Infof("select all clusters")
+		return clusters, nil
 	}
 
-	var clusters []*clusterv1alpha1.Cluster
-	for _, cluster := range groupClustersInfo.Clusters {
-		clusters = append(clusters, cluster.Cluster)
+	if isAvailableResourceIgnored(placement) {
```
RainbowMango, line 23:
```suggestion
	if shouldIgnoreAvaiableResource(placement) {
```

huone1: done

RainbowMango (approved):
/lgtm
/approve

PR #1589, opened 2022-04-08
add UT for aggregatestatus

Poor12:
**What type of PR is this?**
/kind feature

add UT for aggregatestatus

None

### DIFF
--- a/pkg/resourceinterpreter/defaultinterpreter/aggregatestatus_test.go
+++ b/pkg/resourceinterpreter/defaultinterpreter/aggregatestatus_test.go
@@ -1,0 +1,335 @@
+package defaultinterpreter
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	appsv1 "k8s.io/api/apps/v1"
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	extensionsv1beta1 "k8s.io/api/extensions/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
+	"github.com/karmada-io/karmada/pkg/util/helper"
+)
+
+func TestAggregateDeploymentStatus(t *testing.T) {
+	statusMap := map[string]interface{}{
+		"replicas":            1,
+		"readyReplicas":       1,
+		"updatedReplicas":     1,
+		"availableReplicas":   1,
+		"unavailableReplicas": 0,
+	}
+	raw, _ := helper.BuildStatusRawExtension(statusMap)
+	aggregatedStatusItems := []workv1alpha2.AggregatedStatusItem{
+		{ClusterName: "member1", Status: raw, Applied: true},
+		{ClusterName: "member2", Status: raw, Applied: true},
+	}
+
+	oldDeploy := &appsv1.Deployment{}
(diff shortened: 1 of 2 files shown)


pkg/resourceinterpreter/defaultinterpreter/aggregatestatus_test.go
```diff
@@ -0,0 +1,336 @@
...
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	extensionsv1beta1 "k8s.io/api/extensions/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
+	"github.com/karmada-io/karmada/pkg/util/helper"
+
+	"github.com/stretchr/testify/assert"
```
RainbowMango, line 16: Move this line to above.

pkg/resourceinterpreter/defaultinterpreter/aggregatestatus_test.go
```diff
@@ -0,0 +1,335 @@
...
+		oldObj                *unstructured.Unstructured
+		aggregatedStatusItems []workv1alpha2.AggregatedStatusItem
+		newObj                *unstructured.Unstructured
+		wantErr               error
+	}{
+		{
+			oldObj:                objServiceClusterIP,
+			aggregatedStatusItems: aggregatedStatusItemsNotLB,
+			newObj:                objServiceClusterIP,
+			wantErr:               nil,
```
XiShanYongYe-Chang, line 101: Without a test case name, the purpose of the test is not clear at a glance.

Poor12: Thanks for advice. Add testcase name.

XiShanYongYe-Chang (commented):
lgtm

pkg/resourceinterpreter/defaultinterpreter/aggregatestatus_test.go
```diff
@@ -0,0 +1,363 @@
...
+	oldDeploy := &appsv1.Deployment{}
+	newDeploy := &appsv1.Deployment{Status: appsv1.DeploymentStatus{Replicas: 2, ReadyReplicas: 2, UpdatedReplicas: 2, AvailableReplicas: 2, UnavailableReplicas: 0}}
+	oldObj, _ := helper.ToUnstructured(oldDeploy)
+	newObj, _ := helper.ToUnstructured(newDeploy)
+
+	tests := []struct {
+		name                  string
+		oldObj                *unstructured.Unstructured
+		aggregatedStatusItems []workv1alpha2.AggregatedStatusItem
+		newObj                *unstructured.Unstructured
```
RainbowMango, line 41: Does it makes better readability if we rename like?:
```diff
-               oldObj                *unstructured.Unstructured
+               curObj                *unstructured.Unstructured
                aggregatedStatusItems []workv1alpha2.AggregatedStatusItem
-               newObj                *unstructured.Unstructured
+               expectedObj           *unstructured.Unstructured
```

pkg/resourceinterpreter/defaultinterpreter/aggregatestatus_test.go
```diff
@@ -0,0 +1,363 @@
...
+		wantErr               error
```
RainbowMango, line 42: Should we cover the case that expects an error?

If you don't want to cover the `abnormal` case, then the `wantErr` should not be there, which means all your test cases below do not allow `err`.

pkg/resourceinterpreter/defaultinterpreter/aggregatestatus_test.go
Poor12: seems better.

PR #1622, opened 2022-04-18
check if cluster secretRef is nil before using it

### LINKED ISSUE #1621: controller manager crashes when cluster secretRef is empty
**What happened**:

When cluster secretRef is nil, controller manager crashes. 

Controller manager should not crash at any time.

1. Create a cluster object with secretRef is nil
2. Controller manager crashes

**Environment**:
- Karmada version:
1.1.1
kubectl karmada version: version.Info{GitVersion:"v1.1.1", GitCommit:"8ea4c4d8819dc4c4b5fa885e3b24bb2fef910dbb", GitTreeState:"clean", BuildDate:"2022-03-18T01:34:59Z", GoVersion:"go1.17.7", Compiler:"gc", Platform:"linux/amd64"}
- Others:

stingshen:
**What type of PR is this?**

Do nil check before using cluster secretRef

Fixes #1621 

None

```release-note

```

### DIFF
--- a/pkg/util/membercluster_client.go
+++ b/pkg/util/membercluster_client.go
@@ -121,11 +121,12 @@ func buildClusterConfig(clusterName string, client client.Client) (*rest.Config,
 		return nil, fmt.Errorf("the api endpoint of cluster %s is empty", clusterName)
 	}
 
+	if cluster.Spec.SecretRef == nil {
+		return nil, fmt.Errorf("cluster %s does not have a secret", clusterName)
+	}
+
 	secretNamespace := cluster.Spec.SecretRef.Namespace
 	secretName := cluster.Spec.SecretRef.Name
-	if secretName == "" {
-		return nil, fmt.Errorf("cluster %s does not have a secret name", clusterName)
-	}
 
 	secret := &corev1.Secret{}
 	if err := client.Get(context.TODO(), types.NamespacedName{Namespace: secretNamespace, Name: secretName}, secret); err != nil {
--- a/cmd/controller-manager/app/controllermanager.go
+++ b/cmd/controller-manager/app/controllermanager.go
@@ -191,14 +191,29 @@ func startClusterStatusController(ctx controllerscontext.Context) (enabled bool,
 	clusterPredicateFunc := predicate.Funcs{
 		CreateFunc: func(createEvent event.CreateEvent) bool {
 			obj := createEvent.Object.(*clusterv1alpha1.Cluster)
+
+			if obj.Spec.SecretRef == nil {
+				return false
+			}
+
 			return obj.Spec.SyncMode == clusterv1alpha1.Push
 		},
 		UpdateFunc: func(updateEvent event.UpdateEvent) bool {
 			obj := updateEvent.ObjectNew.(*clusterv1alpha1.Cluster)
+
+			if obj.Spec.SecretRef == nil {
+				return false
+			}
+
 			return obj.Spec.SyncMode == clusterv1alpha1.Push
 		},
 		DeleteFunc: func(deleteEvent event.DeleteEvent) bool {
 			obj := deleteEvent.Object.(*clusterv1alpha1.Cluster)
+
+			if obj.Spec.SecretRef == nil {
+				return false
+			}
+
 			return obj.Spec.SyncMode == clusterv1alpha1.Push
 		},
 		GenericFunc: func(genericEvent event.GenericEvent) bool {


pkg/util/membercluster_client.go
```diff
@@ -121,12 +121,13 @@ func buildClusterConfig(clusterName string, client client.Client) (*rest.Config,
 		return nil, fmt.Errorf("the api endpoint of cluster %s is empty", clusterName)
 	}
 
-	secretNamespace := cluster.Spec.SecretRef.Namespace
-	secretName := cluster.Spec.SecretRef.Name
-	if secretName == "" {
+	if cluster.Spec.SecretRef == nil || cluster.Spec.SecretRef.Name == "" {
```
RainbowMango, line 124: Do you think we should check the `Namespace` here?

RainbowMango:
Echo your use case from https://github.com/karmada-io/karmada/issues/1621#issuecomment-1101943380 here:
> Users call the api server to register cluster. The api server creates a cluster object without secretRef, sa, impersonate sa, and doesn't wait for the sa secret to be generated. The controller manager watches for the cluster object and fill in the sa secret if available.

This patch prevents panic from nil reference. But before your `controller manager` fills the `secret` the `cluster status controller` might keep failing as it can't build client
https://github.com/karmada-io/karmada/blob/082c3e182d1a8aa19b59b0d23ec672e2aba6471f/pkg/controllers/status/cluster_status_controller.go#L127-L132

Have you ever found this issue? Can you live with that? I'm thinking about how to deal with it in an elegant way.

pkg/util/membercluster_client.go
stingshen: Actually, `Name` and `Namespace` fields are required in `SecretRef` schema. So, maybe we don't need check `Name` nor `Namespace` here?

stingshen:
> This patch prevents panic from nil reference. But before your `controller manager` fills the `secret` the `cluster status controller` might keep failing as it can't build client
> 
> https://github.com/karmada-io/karmada/blob/082c3e182d1a8aa19b59b0d23ec672e2aba6471f/pkg/controllers/status/cluster_status_controller.go#L127-L132
> 
> Have you ever found this issue? Can you live with that? I'm thinking about how to deal with it in an elegant way.

Yes, I found this issue and can live with that. I think it is the expected behavior that controller throws error and retries in such circumstances...

pkg/util/membercluster_client.go
RainbowMango: Oh, you are right.

RainbowMango:
> Yes, I found this issue and can live with that. I think it is the expected behavior that controller throws error and retries in such circumstances...

Hold on for a while. Let me see if there is an elegant way. 

By the way, are you using Karmada in production?

stingshen:
> By the way, are you using Karmada in production?

No. We are just doing research and creating a POC of multi-cluster.

RainbowMango:
I guess we can ignore the Cluster that `SecretRef` hasn't present, like:
```diff
diff --git a/cmd/controller-manager/app/controllermanager.go b/cmd/controller-manager/app/controllermanager.go
index e0653716..bf43bf45 100644                                                                               
--- a/cmd/controller-manager/app/controllermanager.go                                                         
+++ b/cmd/controller-manager/app/controllermanager.go                                                         
@@ -191,14 +191,29 @@ func startClusterStatusController(ctx controllerscontext.Context) (enabled bool,        
        clusterPredicateFunc := predicate.Funcs{                                                              
                CreateFunc: func(createEvent event.CreateEvent) bool {
                        obj := createEvent.Object.(*clusterv1alpha1.Cluster)
+
+                       if obj.Spec.SecretRef == nil {
+                               return false
+                       }
+
                        return obj.Spec.SyncMode == clusterv1alpha1.Push
                },
                UpdateFunc: func(updateEvent event.UpdateEvent) bool {
                        obj := updateEvent.ObjectNew.(*clusterv1alpha1.Cluster)
+
+                       if obj.Spec.SecretRef == nil {
+                               return false
+                       }
+
                        return obj.Spec.SyncMode == clusterv1alpha1.Push
                },
                DeleteFunc: func(deleteEvent event.DeleteEvent) bool {
                        obj := deleteEvent.Object.(*clusterv1alpha1.Cluster)
+
+                       if obj.Spec.SecretRef == nil {
+                               return false
+                       }
+
                        return obj.Spec.SyncMode == clusterv1alpha1.Push
                },
                GenericFunc: func(genericEvent event.GenericEvent) bool {
```

How do you think?

stingshen:
> I guess we can ignore the Cluster that `SecretRef` hasn't present, like:
> 
> ```diff
> diff --git a/cmd/controller-manager/app/controllermanager.go b/cmd/controller-manager/app/controllermanager.go
> index e0653716..bf43bf45 100644                                                                               
> --- a/cmd/controller-manager/app/controllermanager.go                                                         
> +++ b/cmd/controller-manager/app/controllermanager.go                                                         
> @@ -191,14 +191,29 @@ func startClusterStatusController(ctx controllerscontext.Context) (enabled bool,        
>         clusterPredicateFunc := predicate.Funcs{                                                              
>                 CreateFunc: func(createEvent event.CreateEvent) bool {
>                         obj := createEvent.Object.(*clusterv1alpha1.Cluster)
> +
> +                       if obj.Spec.SecretRef == nil {
> +                               return false
> +                       }
> +
>                         return obj.Spec.SyncMode == clusterv1alpha1.Push
>                 },
>                 UpdateFunc: func(updateEvent event.UpdateEvent) bool {
>                         obj := updateEvent.ObjectNew.(*clusterv1alpha1.Cluster)
> +
> +                       if obj.Spec.SecretRef == nil {
> +                               return false
> +                       }
> +
>                         return obj.Spec.SyncMode == clusterv1alpha1.Push
>                 },
>                 DeleteFunc: func(deleteEvent event.DeleteEvent) bool {
>                         obj := deleteEvent.Object.(*clusterv1alpha1.Cluster)
> +
> +                       if obj.Spec.SecretRef == nil {
> +                               return false
> +                       }
> +
>                         return obj.Spec.SyncMode == clusterv1alpha1.Push
>                 },
>                 GenericFunc: func(genericEvent event.GenericEvent) bool {
> ```
> 
> How do you think?

There may be a corner case that cluster object cannot be deleted? The following is the procedure:

1. Create a cluster object with SecretRef
2. Cluster controller adds a `finalizer` to the cluster
3. Somehow, delete the SecretRef
4. Delete the cluster object
5. The cluster cannot be deleted because no controller deletes the `finalizer`

I can live with the behavior, but I don't know whether Karmada community can accept that.

RainbowMango:
@stingshen I'm OK with this patch. Just want to confirm if we can eliminate the surprising error logs.

> The cluster cannot be deleted because no controller deletes the finalizer

Why do you think no controller deletes the finalizer?  The [cluster controller](https://github.com/karmada-io/karmada/blob/5b3ef7d93f63daac0b396ef50fdacb4ab0f5ca48/pkg/controllers/cluster/cluster_controller.go#L81-L83) will remove the finalizer(`karmada.io/cluster-controller`). 

The above patch(https://github.com/karmada-io/karmada/pull/1622#issuecomment-1102170751) adds a filter to `cluster status` controller, it's kind of a short path.

stingshen:
> Why do you think no controller deletes the finalizer? The [cluster controller](https://github.com/karmada-io/karmada/blob/5b3ef7d93f63daac0b396ef50fdacb4ab0f5ca48/pkg/controllers/cluster/cluster_controller.go#L81-L83) will remove the finalizer(`karmada.io/cluster-controller`).

Oh, this patch is fine. The cluster can be deleted correctly on the scenario I just provided. I mistakenly thought the filter is on cluster controller:)

I tested the patch on my enviornment. It works fine except that the error still occurs in the following circumstances:

1. Register a cluster with secretRef. Cluster status controller starts updating cluster status
2. Delete the secretRef from cluster. The cluster status controller still reconciles the cluster because the previous reconcile result is: `controllerruntime.Result{Requeue: true}`
3. The reconcile fails with `cluster xxx does not have a secret name` periodically

I think this is an extremely corner case and the behavior is acceptable. What do you think?

RainbowMango:
/lgtm

> Delete the secretRef from cluster. The cluster status controller still reconciles the cluster because the previous reconcile result is: controllerruntime.Result{Requeue: true}

Do you mean delete the `Secret` object or set `.spec.SecretRef` of `Cluster` object to `nil`? That's different.

If you delete the `Secret` object, yes the cluster status controller will continue the reconciliation and fails when it tries to get the secret.

If you set `.spec.SecretRef` of `Cluster` object to `nil`, the cluster `update` event will be ignored from the queue, thus the status controller won't reconcile it (No more error logs). 
But there is a case where, the `update` event is already in the queue, but hasn't been consumed by `cluster status controller`. If you set the `.spec.SecretRef` of the `Cluster` object to `nil`, the controller will reconcile it and continually failing(re-try).

RainbowMango:
This patch is pretty good now. Hold on to confirm the corner case you mentioned above. Thanks, @stingshen in advance.

stingshen:
> /lgtm
> 
> > Delete the secretRef from cluster. The cluster status controller still reconciles the cluster because the previous reconcile result is: controllerruntime.Result{Requeue: true}
> 
> Do you mean delete the `Secret` object or set `.spec.SecretRef` of `Cluster` object to `nil`? That's different.
> 
> If you delete the `Secret` object, yes the cluster status controller will continue the reconciliation and fails when it tries to get the secret.
> 
> If you set `.spec.SecretRef` of `Cluster` object to `nil`, the cluster `update` event will be ignored from the queue, thus the status controller won't reconcile it (No more error logs). But there is a case where, the `update` event is already in the queue, but hasn't been consumed by `cluster status controller`. If you set the `.spec.SecretRef` of the `Cluster` object to `nil`, the controller will reconcile it and continually failing(re-try).

I mean set `.spec.SecretRef` of `Cluster` object to `nil`. From my observation, the controller reconciles the cluster and continually failing in all circumstances following the procedure I provided. No matter whether the reconcile succeeds, the cluster status controller always requeue the reconcile after ClusterStatusUpdateFrequency.

RainbowMango:
> No matter whether the reconcile succeeds, the cluster status controller always requeue the reconcile after ClusterStatusUpdateFrequency.

Yeah. right.

RainbowMango (approved):
/approve

Looking forward to hearing your feedback. You can DM me if you need any help. Thanks.

Issue #1371, opened 2022-02-17
Extension for out-of-tree plugins


kerthcet:
**What would you like to be added**:
We have some filter plugins customized, and we want to add them to scheduler easily, e.g.
1. initialization
```golang
	command := app.NewSchedulerCommand(stopChan, customizedPlugins...)
```
2. Locate our plugins' source code to the plugins folder. 

**Why is this needed**:
Extension for out-of-tree plugins.

TODOs:

- [x] introduce registry to store plugins in framework initialization: https://github.com/karmada-io/karmada/pull/1382
- [ ] add out-of-tree options to initialize scheduler

RainbowMango:
I'd like to have a chat with you at the next community meeting.
If it is accepted, it can be included in the next release(release 1.2).

RainbowMango:
I think we are ready to start the discussion now, right? @huone1

huone1:
the first part  has been merged， I am looking forward that you can start the development of the second port

kerthcet:
I have completed the second part but will follow after https://github.com/karmada-io/karmada/pull/1627 merged for some extra works.

PR #1627, opened 2022-04-20
introduce schedulerOptions to optimize scheduler instantiation

kerthcet:

**What type of PR is this?**
/kind feature

1. Currently  when we instantce `NewScheduler`, we passed in a named `Options` struct which contains too much useless fields, this PR made the parameters passed in when needed.  It's also easy to mock scheduler.
2. Add scheduler unit tests.

Fixes #

```release-note
None
```

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -21,7 +21,6 @@ import (
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog/v2"
 
-	"github.com/karmada-io/karmada/cmd/scheduler/app/options"
 	clusterv1alpha1 "github.com/karmada-io/karmada/pkg/apis/cluster/v1alpha1"
 	policyv1alpha1 "github.com/karmada-io/karmada/pkg/apis/policy/v1alpha1"
 	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
@@ -86,8 +85,41 @@ type Scheduler struct {
 	schedulerEstimatorWorker util.AsyncWorker
 }
 
+type schedulerOptions struct {
+	// enableSchedulerEstimator represents whether the accurate scheduler estimator should be enabled.
+	enableSchedulerEstimator bool
+	// schedulerEstimatorTimeout specifies the timeout period of calling the accurate scheduler estimator service.
+	schedulerEstimatorTimeout metav1.Duration
+	// schedulerEstimatorPort is the port that the accurate scheduler estimator server serves at.
+	schedulerEstimatorPort int
+}
+
+// Option configures a Scheduler
+type Option func(*schedulerOptions)
+
+// WithEnableSchedulerEstimator sets the enableSchedulerEstimator for scheduler
+func WithEnableSchedulerEstimator(enableSchedulerEstimator bool) Option {
+	return func(o *schedulerOptions) {
+		o.enableSchedulerEstimator = enableSchedulerEstimator
+	}
+}
+
+// WithSchedulerEstimatorTimeout sets the schedulerEstimatorTimeout for scheduler
+func WithSchedulerEstimatorTimeout(schedulerEstimatorTimeout metav1.Duration) Option {
+	return func(o *schedulerOptions) {
+		o.schedulerEstimatorTimeout = schedulerEstimatorTimeout
+	}
+}
+
+// WithSchedulerEstimatorPort sets the schedulerEstimatorPort for scheduler
+func WithSchedulerEstimatorPort(schedulerEstimatorPort int) Option {
+	return func(o *schedulerOptions) {
+		o.schedulerEstimatorPort = schedulerEstimatorPort
+	}
+}
+
 // NewScheduler instantiates a scheduler
-func NewScheduler(dynamicClient dynamic.Interface, karmadaClient karmadaclientset.Interface, kubeClient kubernetes.Interface, opts *options.Options) (*Scheduler, error) {
+func NewScheduler(dynamicClient dynamic.Interface, karmadaClient karmadaclientset.Interface, kubeClient kubernetes.Interface, opts ...Option) (*Scheduler, error) {
 	factory := informerfactory.NewSharedInformerFactory(karmadaClient, 0)
 	bindingLister := factory.Work().V1alpha2().ResourceBindings().Lister()
 	policyLister := factory.Policy().V1alpha1().PropagationPolicies().Lister()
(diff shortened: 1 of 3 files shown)


kerthcet:
cc @huone1

RainbowMango:
Is this part of #1371 ?

pkg/scheduler/scheduler_test.go
```diff
@@ -0,0 +1,49 @@
...
+		},
+	}
+
+	for _, tc := range testcases {
+		t.Run(tc.name, func(t *testing.T) {
+			sche, err := NewScheduler(dynamicClient, karmadaClient, kubeClient, tc.opts...)
+			if err != nil {
+				t.Errorf("create scheduler error: %s", err)
+			}
+			if diff := cmp.Diff(tc.enableSchedulerEstimator, sche.enableSchedulerEstimator); diff != "" {
```
RainbowMango, line 44: Is it the right place to introduce `cmp` here? 
I tried to hack this test and make it fail. It doesn't looks better:
```
=== RUN   TestCreateScheduler/scheduler_with_enableSchedulerEstimator_enabled
    scheduler_test.go:45: unexpected enableSchedulerEstimator (-want, +got):
          bool(
        -       false,
        +       true,
          )
```

pkg/scheduler/scheduler.go
```diff
@@ -105,30 +142,31 @@ func NewScheduler(dynamicClient dynamic.Interface, karmadaClient karmadaclientse
...
-		bindingLister:            bindingLister,
-		policyLister:             policyLister,
-		clusterBindingLister:     clusterBindingLister,
-		clusterPolicyLister:      clusterPolicyLister,
-		clusterLister:            clusterLister,
-		informerFactory:          factory,
-		queue:                    queue,
-		Algorithm:                algorithm,
-		schedulerCache:           schedulerCache,
-		enableSchedulerEstimator: opts.EnableSchedulerEstimator,
```
RainbowMango, line 120: nice finding.

pkg/scheduler/scheduler.go
```diff
@@ -86,8 +85,41 @@ type Scheduler struct {
...
+	// enableSchedulerEstimator represents whether the accurate scheduler estimator should be enabled.
+	enableSchedulerEstimator bool
+	// schedulerEstimatorTimeout specifies the timeout period of calling the accurate scheduler estimator service.
+	schedulerEstimatorTimeout metav1.Duration
+	// schedulerEstimatorPort is the port that the accurate scheduler estimator server serves at.
+	schedulerEstimatorPort int
+}
+
+// Option configures a Scheduler
+type Option func(*schedulerOptions)
```
RainbowMango, line 98: Would it be a little bit confusing for this Option `function` and [struct](https://github.com/karmada-io/karmada/blob/28a6c2c7672a5efd504c9c264bf2c6f53e5e89b4/cmd/scheduler/app/options/options.go#L28)?

RainbowMango (commented):
/assign @lonelyCZ @huone1

kerthcet:
> Is this part of #1371 ?

Somehow related.

pkg/scheduler/scheduler_test.go
kerthcet: got it.

pkg/scheduler/scheduler.go
kerthcet: They belong to different packages, looks good to me. Or change to `type Configurator func(*schedulerConfigurators)` ?

huone1:
it is similar to the k8s framework，and  is the ""cmp""  necessary ?

pkg/scheduler/scheduler.go
RainbowMango: I'm thinking about why we need to introduce the `Option`, can we reuse the legacy option? https://github.com/karmada-io/karmada/blob/28a6c2c7672a5efd504c9c264bf2c6f53e5e89b4/cmd/scheduler/app/options/options.go#L28

I know, there is a lot of `option`s in the Kubernetes scheduler, but my concern is the readability.

huone1: it is better to describe the benefits to import the **Option**

kerthcet: Yes, it's not necessary but I think meaningful. [karmada/cmd/scheduler/app/options/options.go](https://github.com/karmada-io/karmada/blob/28a6c2c7672a5efd504c9c264bf2c6f53e5e89b4/cmd/scheduler/app/options/options.go#L28) holds lots of options we don't need to instantiate scheduler, as time goes on, there will be more options introduced. It's hard to maintain as we have to check the code to see which options we need and which one is required by other components, so we should only keep the necessary ones. Also it's easy to mock the `NewScheduler` tests then.

kerthcet:
> it is similar to the k8s framework，and is the ""cmp"" necessary ?

Well, it's also not necessary, but I prefer to introduce this lib to our tests as we don't have to think about the proper words like `We expected xxxx, but got yyy', we can unified the test like 
```
			if diff := cmp.Diff(tc.want, got); diff != "" {
				t.Errorf("Unexpected xxx (-want,+got):\n%s", diff)
			}
```
It also reduces the burden of reviewers.  But I'm ok with introduce this lib or not, in the pr or that pr.  WDYT?

kerthcet:
I'll revert the `cmp.Diff` change then.

huone1:
> I'll revert the `cmp.Diff` change then.

@RainbowMango  what do you think about the lib **cmp** ?

kerthcet:
> > I'll revert the `cmp.Diff` change then.
> 
> @RainbowMango what do you think about the lib **cmp** ?

Removed the cmp package.

pkg/scheduler/scheduler_test.go
```diff
@@ -0,0 +1,49 @@
...
+	}{
+		{
+			name:                     "scheduler with default configuration",
+			opts:                     nil,
+			enableSchedulerEstimator: false,
+		},
+		{
+			name: "scheduler with enableSchedulerEstimator enabled",
+			opts: []Option{
+				WithEnableSchedulerEstimator(true),
```
RainbowMango, line 31: How about the `WithSchedulerEstimatorTimeout` and `WithSchedulerEstimatorPort` ? should we test them as well?

kerthcet: I'll add tests for `WithSchedulerEstimatorPort`, but `Timeout` is hard to test unless we add a getter method. Considering `WithSchedulerEstimatorPort` somehow behaves the same as `WithSchedulerEstimatorTimeout`, I think it's ok.

RainbowMango (approved):
/lgtm
/approve

PR #1629, opened 2022-04-21
Introduce AssignedReplicasForCluster for ResourceBinding

RainbowMango:
**What type of PR is this?**
/kind cleanup

Introduce `AssignedReplicasForCluster()` method to `ResourceBinding`/`ClusterResourceBinding`.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/core/spreadconstraint/group_clusters.go
+++ b/pkg/scheduler/core/spreadconstraint/group_clusters.go
@@ -117,7 +117,7 @@ func (info *GroupClustersInfo) generateClustersInfo(clustersScore framework.Clus
 	clustersReplicas := calAvailableReplicas(clusters, rbSpec)
 	for i, clustersReplica := range clustersReplicas {
 		info.Clusters[i].AvailableReplicas = int64(clustersReplica.Replicas)
-		info.Clusters[i].AvailableReplicas += int64(getScheduledReplicas(rbSpec, clustersReplica.Name))
+		info.Clusters[i].AvailableReplicas += int64(rbSpec.AssignedReplicasForCluster(clustersReplica.Name))
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/scheduler/core/spreadconstraint/group_clusters.go
```diff
@@ -117,7 +117,7 @@ func (info *GroupClustersInfo) generateClustersInfo(clustersScore framework.Clus
 	clustersReplicas := calAvailableReplicas(clusters, rbSpec)
 	for i, clustersReplica := range clustersReplicas {
 		info.Clusters[i].AvailableReplicas = int64(clustersReplica.Replicas)
-		info.Clusters[i].AvailableReplicas += int64(getScheduledReplicas(rbSpec, clustersReplica.Name))
+		info.Clusters[i].AvailableReplicas += int64(rbSpec.AssignedReplicasForCluster(clustersReplica.Name))
```
RainbowMango, line 120: @huone1 Can we be sure that `rbSpec` is not nil here? Otherwise, it will panic.

RainbowMango: The failing unit test is due to `nil` `speck` at :
https://github.com/karmada-io/karmada/blob/28a6c2c7672a5efd504c9c264bf2c6f53e5e89b4/pkg/scheduler/core/spreadconstraint/group_clusters_test.go#L54-L57

huone1: It will not be a null pointer  because the spec is structure ， not a pointer in the definition of resource binding API

RainbowMango:
/assign @huone1 
Solved the failing unit test, hopefully.

PR #1633, opened 2022-04-21
Add ReflectStatus ResourceInterperter interface to grab workload status

XiShanYongYe-Chang:

**What type of PR is this?**

/kind feature

Add ReflectStatus ResourceInterperter interface to grab workload status, so as to support granularity control of workload status collection.

Fixes #

The default implementation of some resources kind(such as `deployment`) is not added. 

Add default reflectstatus implementation with:
- Deployment kind
- Service kind
- Ingress kind
- Job kind
- DaemonSet kind
- StatefulSet kind

```release-note
Supports granularity control of workload status collection with InterpretStatus InterpreterOperation.
```

### DIFF
--- a/pkg/controllers/status/workstatus_controller.go
+++ b/pkg/controllers/status/workstatus_controller.go
@@ -21,6 +21,7 @@ import (
 
 	clusterv1alpha1 "github.com/karmada-io/karmada/pkg/apis/cluster/v1alpha1"
 	workv1alpha1 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha1"
+	"github.com/karmada-io/karmada/pkg/resourceinterpreter"
 	"github.com/karmada-io/karmada/pkg/sharedcli/ratelimiterflag"
 	"github.com/karmada-io/karmada/pkg/util"
 	"github.com/karmada-io/karmada/pkg/util/helper"
@@ -50,6 +51,7 @@ type WorkStatusController struct {
 	ClusterClientSetFunc      func(clusterName string, client client.Client) (*util.DynamicClusterClient, error)
 	ClusterCacheSyncTimeout   metav1.Duration
 	RateLimiterOptions        ratelimiterflag.Options
+	ResourceInterpreter       resourceinterpreter.ResourceInterpreter
 }
 
 // Reconcile performs a full reconciliation for the object referred to by the Request.
@@ -275,10 +277,10 @@ func (c *WorkStatusController) recreateResourceIfNeeded(work *workv1alpha1.Work,
 
 // reflectStatus grabs cluster object's running status then updates to its owner object(Work).
 func (c *WorkStatusController) reflectStatus(work *workv1alpha1.Work, clusterObj *unstructured.Unstructured) error {
-	statusMap, _, err := unstructured.NestedMap(clusterObj.Object, "status")
+	statusRaw, err := c.ResourceInterpreter.ReflectStatus(clusterObj)
 	if err != nil {
-		klog.Errorf("Failed to get status field from %s(%s/%s), error: %v", clusterObj.GetKind(), clusterObj.GetNamespace(), clusterObj.GetName(), err)
-		return err
+		klog.Errorf("Failed to reflect status for object(%s/%s/%s) with resourceInterpreter.",
+			clusterObj.GetKind(), clusterObj.GetNamespace(), clusterObj.GetName(), err)
 	}
 
 	identifier, err := c.buildStatusIdentifier(work, clusterObj)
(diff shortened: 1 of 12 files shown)


pkg/resourceinterpreter/defaultinterpreter/default.go
```diff
@@ -106,3 +112,23 @@ func (e *DefaultInterpreter) GetDependencies(object *unstructured.Unstructured)
...
+
+// ReflectStatus returns the cluster object's running status after the grab.
+func (e *DefaultInterpreter) ReflectStatus(object *unstructured.Unstructured) (status *runtime.RawExtension, err error) {
+	handler, exist := e.reflectStatusHandlers[object.GroupVersionKind()]
+	if exist {
+		return handler(object)
+	}
+
+	// for no default build-in resource kind, directly collect the entire status.
+	statusMap, _, err := unstructured.NestedMap(object.Object, "status")
```
RainbowMango, line 124: Please add a unit test for this function. I want to see what happened if no `.status` field.
Maybe we can use the second return value to suppress error logs.

XiShanYongYe-Chang: Done.

pkg/util/helper/workstatus.go
```diff
@@ -283,7 +283,7 @@ func IsWorkContains(workStatus *workv1alpha1.WorkStatus, targetResource schema.G
 }
 
 // BuildStatusRawExtension builds raw JSON by a status map.
-func BuildStatusRawExtension(status map[string]interface{}) (*runtime.RawExtension, error) {
+func BuildStatusRawExtension(status interface{}) (*runtime.RawExtension, error) {
```
RainbowMango, line 286: Just a question, why change the parameter here?

pkg/resourceinterpreter/customizedinterpreter/customized.go
```diff
@@ -293,3 +294,18 @@ func (e *CustomizedInterpreter) GetDependencies(ctx context.Context, attributes
 
 	return response.Dependencies, matched, nil
 }
+
+// ReflectStatus returns the cluster object's running status after the grab.
```
RainbowMango, line 298: Update the comments.

pkg/controllers/status/workstatus_controller.go
```diff
@@ -275,10 +277,10 @@ func (c *WorkStatusController) recreateResourceIfNeeded(work *workv1alpha1.Work,
 
 // reflectStatus grabs cluster object's running status then updates to its owner object(Work).
 func (c *WorkStatusController) reflectStatus(work *workv1alpha1.Work, clusterObj *unstructured.Unstructured) error {
-	statusMap, _, err := unstructured.NestedMap(clusterObj.Object, "status")
+	statusRaw, err := c.ResourceInterpreter.ReflectStatus(clusterObj)
```
RainbowMango, line 280: If `statusRaw` == nil and the status in `Work` is nil too, then we can save one update operation.

There is a corner case that, the status in `Work` is not nil, but this time we get `nil` from the interpreter, then, we have to update the `Work`.

pkg/resourceinterpreter/defaultinterpreter/reflectstatus.go
```diff
@@ -0,0 +1,50 @@
...
+
+func reflectDeploymentStatus(object *unstructured.Unstructured) (*runtime.RawExtension, error) {
+	statusMap, exist, err := unstructured.NestedMap(object.Object, "status")
+	if err != nil {
+		klog.Errorf("Failed to get status field from %s(%s/%s), error: %v",
+			object.GetKind(), object.GetNamespace(), object.GetName(), err)
+		return nil, err
+	}
+	if !exist {
+		klog.Warningf("Obtain no status from %s(%s/%s).",
```
RainbowMango, line 32:
```suggestion
		klog.Errorf("Failed to grab status from xxx which should have status field.",
```

pkg/util/helper/workstatus.go
XiShanYongYe-Chang: So we can directly use this function when the input parameter is `appsv1.DeploymentStatus` or something else.

In fact, we will marshal the status inside this method, the input parameter of the `json.Marshal` is not restrict with `map[string]interface{}` type.

This change can makes this function more general.

pkg/controllers/status/workstatus_controller.go
XiShanYongYe-Chang: How about adding an equal judgment when updating.

RainbowMango (approved):
/lgtm
/approve

PR #1635, opened 2022-04-21
docs: add FAQ about reuse kube-apiserver

pigletfly:
Signed-off-by: Wang Bing <wangbing.adam@gmail.com>

**What type of PR is this?**
/kind documentation

add faq "install Karmada in a existing kubernetes cluster"

Fixes #

```release-note
     NONE
```

### DIFF
--- a/docs/frequently-asked-questions.md
+++ b/docs/frequently-asked-questions.md
@@ -20,3 +20,21 @@ to keep a consistent user experience and behavior.
 
 It's worth noting that not all controllers are needed by Karmada, for the recommended controllers please
 refer to [Recommended Controllers](./userguide/configure-controllers.md#recommended-controllers).
+
+
+## Can I install Karmada in a Kubernetes cluster and reuse the kube-apiserver as Karmada apiserver?
+
+The quick answer is `yes`. In that case, you can save the effort to deploy
+[karmada-apiserver](https://github.com/karmada-io/karmada/blob/master/artifacts/deploy/karmada-apiserver.yaml) and just
+share the APIServer between Kubernetes and Karmada. In addition, the high availability capabilities in the origin clusters
+can be inherited seamlessly. We do have some users using Karmada in this way.
+
+There are some things you should consider before doing so:
+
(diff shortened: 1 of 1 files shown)


docs/frequently-asked-questions.md
```diff
@@ -20,3 +20,10 @@ to keep a consistent user experience and behavior.
 
 It's worth noting that not all controllers are needed by Karmada, for the recommended controllers please
 refer to [Recommended Controllers](./userguide/configure-controllers.md#recommended-controllers).
+
+
+## Can I install Karmada in a existing kubernetes cluster as host cluster?
```
RainbowMango, line 25: I guess you want to explain `Can I install Karmada in an existing cluster and reuse the `kube-apiserver` as Karmada apiserver?` Am I right?

pigletfly: yes

RainbowMango:
I think it's worth noting. How about this?
```
## Can I install Karmada in a Kubernetes cluster and reuse the kube-apiserver as Karmada apiserver?

The quick answer is `yes`. In that case, you can save the effort to deploy
[karmada-apiserver](https://github.com/karmada-io/karmada/blob/master/artifacts/deploy/karmada-apiserver.yaml) and just
share the APIServer between Kubernetes and Karmada. In addition, the high availability capabilities in the origin clusters
can be inherited seamlessly. We do have some users using Karmada in this way.

There are some things you should consider before doing so:
- This approach hasn't been fully tested by the Karmada community and no plan for it yet.
- This approach will increase computation costs for the Karmada system. E.g.
  After you apply a `resource template`, take `Deployment` as an example, the `kube-controller` will create `Pods` for the
  Deployment and update the status persistently, Karmada system will reconcile these changes too, so there might be 
  conflicts.

TODO: Link to adoption use case once it gets on board.
```

RainbowMango:
cc @pigletfly

RainbowMango (approved):
/retitle docs: add FAQ about reuse kube-apiserver
/lgtm
/approve

PR #1636, opened 2022-04-21
skip reconciling loop quickly after workload gone

snowplayfire:
**What type of PR is this?**

/kind cleanup

Fixes #

```release-note

```

### DIFF
--- a/pkg/controllers/binding/binding_controller.go
+++ b/pkg/controllers/binding/binding_controller.go
@@ -115,6 +115,9 @@ func (c *ResourceBindingController) syncBinding(binding *workv1alpha2.ResourceBi
 
 	workload, err := helper.FetchWorkload(c.DynamicClient, c.InformerManager, c.RESTMapper, binding.Spec.Resource)
 	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return controllerruntime.Result{}, nil
+		}
 		klog.Errorf("Failed to fetch workload for resourceBinding(%s/%s). Error: %v.",
 			binding.GetNamespace(), binding.GetName(), err)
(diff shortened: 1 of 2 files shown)


pkg/controllers/binding/binding_controller.go
```diff
@@ -115,6 +115,9 @@ func (c *ResourceBindingController) syncBinding(binding *workv1alpha2.ResourceBi
 
 	workload, err := helper.FetchWorkload(c.DynamicClient, c.InformerManager, c.RESTMapper, binding.Spec.Resource)
 	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return controllerruntime.Result{}, nil
+		}
```
RainbowMango, line 120: If the workload has been removed, then the `ResourceBinding` will be removed subsequently(by the garbage collector in favor of ownerReference). That means even we retry(return `controllerruntime.Result{Requeue: true}`) here, the retry loop will be ended eventually(at [here](https://github.com/karmada-io/karmada/blob/28a6c2c7672a5efd504c9c264bf2c6f53e5e89b4/pkg/controllers/binding/binding_controller.go#L63-L66)) . 

So, I think the benefit of this patch is we can save the retry loop and prevent annoying error logs. Am I right?

RainbowMango:
/retitle skip reconciling loop quickly after workload gone

pkg/controllers/binding/binding_controller.go
RainbowMango: I talked to @snowplayfire and her use case is a little bit complicated. But what's in common is that we can use this patch to break the loop if the resource template has gone.

RainbowMango (approved):
/lgtm
/approve

PR #1656, opened 2022-04-25
Add time zone interpretation for local-up-karmada.sh

my-git9:
/kind cleanup

Add mainland China time zone judgment to automatically confirm whether to use a proxy.
Is this better for Chinese users?

### DIFF
--- a/hack/local-up-karmada.sh
+++ b/hack/local-up-karmada.sh
@@ -28,6 +28,7 @@ KIND_LOG_FILE=${KIND_LOG_FILE:-"/tmp/karmada"}
 
 #step0: prepare
 # proxy setting in China mainland
+ping -c 3 -i 0.2 -W 3 k8s.gcr.io &>/dev/null || CHINA_MAINLAND=yes
 
 if [[ -n ${CHINA_MAINLAND:-} ]]; then
 export GOPROXY=https://goproxy.cn,direct # set domestic go proxy


RainbowMango:
/assign @mrlihanbo @lfbear 
Could you please help with this?

mrlihanbo:
This pr is to automatically confirm whether to use a proxy which will be better for Chinese users. 
/lgtm

RainbowMango:
I personally usually work on a machine located in Hong Kong. The timezone matches the condition in this patch:
```
# date +"%Z %z"
CST +0800
```

When I run `hack/local-up-karmada.sh`, it'll fetch images from a mirror:
```
# kubectl get deployments.apps -n karmada-system karmada-apiserver -o=jsonpath='{.spec.template.spec.containers[0].image}'
registry.aliyuncs.com/google_containers/kube-apiserver:v1.21.7-
```
Have you notice this? @mrlihanbo

my-git9:
> I personally usually work on a machine located in Hong Kong. The timezone matches the condition in this patch:
> 
> ```
> # date +"%Z %z"
> CST +0800
> ```
> 
> When I run `hack/local-up-karmada.sh`, it'll fetch images from a mirror:
> 
> ```
> # kubectl get deployments.apps -n karmada-system karmada-apiserver -o=jsonpath='{.spec.template.spec.containers[0].image}'
> registry.aliyuncs.com/google_containers/kube-apiserver:v1.21.7
> ```
> 
> Have you notice this? @mrlihanbo
Would it be better？
```
timedatectl |grep 'Asia/Shanghai' && CHINA_MAINLAND=yes
```

RainbowMango:
Is the `timedatectl` common enough that it is always available in system(Linux/MacOS)? 

Show the output on my side, the effect is the same as the previous one:
```
-bash-4.2# timedatectl 
      Local time: Thu 2022-04-28 08:56:40 CST
  Universal time: Thu 2022-04-28 00:56:40 UTC
        RTC time: Thu 2022-04-28 00:56:37
       Time zone: Asia/Shanghai (CST, +0800)
     NTP enabled: yes
NTP synchronized: yes
 RTC in local TZ: no
      DST active: n/a
```

RainbowMango:
I think if people in China can set the `CHINA_MAINLAND=y` to their bash profile.

my-git9:
> I think if people in China can set the `CHINA_MAINLAND=y` to their bash profile.

Do we need to alert the user in some way?

lfbear:
I think most will be settled by this solution, however, still miss out, only tolerable.
What I want to say is that the original of `CHINA_MAINLAND` is just a way to solve the GFW, why don't we just try the networks to decide the value of  `CHINA_MAINLAND`. For instance, connect the 'k8s.gcr.io' or other with a timeout. 
Taking control of the time well will not bring any trouble for the users.

my-git9:
I personally think that this PR is not intended to benefit everyone.
I just hope that most new users in China can deploy a set of Karmada more quickly without affecting the use of users in other regions. 
As a new user of karmada, I spend a lot of time dealing with proxy and mirroring issues. And when I looked at the script I found that what I was doing most of the time was not needed.

RainbowMango:
> why don't we just try the networks to decide the value of CHINA_MAINLAND. For instance, connect the 'k8s.gcr.io' or other with a timeout.

I like the idea.

my-git9:
> > why don't we just try the networks to decide the value of CHINA_MAINLAND. For instance, connect the 'k8s.gcr.io' or other with a timeout.
> 
> I like the idea.
Is this OK ?
```
ping -c 3 -i 0.2 -W 3 k8s.gcr.io &>/dev/null || CHINA_MAINLAND=yes
```

RainbowMango:
LGTM
@lfbear It's your call.

RainbowMango:
/lgtm
/approve

On behalf of @lfbear

PR #1658, opened 2022-04-25
remove karmada-system in chart

my-git9:
I don't think the 'karmada-system' namespace should be hard-coded in the helm chart

/kind cleanup

### DIFF
--- a/charts/templates/_karmada_webhook_configuration.tpl
+++ b/charts/templates/_karmada_webhook_configuration.tpl
@@ -59,7 +59,7 @@ webhooks:
         resources: ["clusteroverridepolicies"]
         scope: "Cluster"
     clientConfig:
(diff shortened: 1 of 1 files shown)


charts/templates/_karmada_webhook_configuration.tpl
```diff
@@ -59,7 +59,7 @@ webhooks:
         resources: ["clusteroverridepolicies"]
         scope: "Cluster"
     clientConfig:
-      url: https://karmada-webhook.karmada-system.svc:443/validate-clusteroverridepolicy
+      url: https://karmada-webhook.{{ $namespace }}.svc:443/validate-clusteroverridepolicy
```
Poor12, line 62: Thanks！Why not change to https://{{ $name }}-webhook.{{ $namespace }}

RainbowMango (approved):
/lgtm
/approve

PR #1670, opened 2022-04-26
Update judge work contains serviceExport with .spec.workload.manifests

XiShanYongYe-Chang:

**What type of PR is this?**

/kind cleanup

The `ServiceExport` resource does not have the status filed. Therefore, the work status of the ServiceExport type should not be collected during work status collection. In this case, the information in work status is not suitable to determine whether the work contains ServiceExport.

So I change to use `.spec.workload.manifests` to judge whether the work contains `ServiceExport`.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/mcs/service_export_controller.go
+++ b/pkg/controllers/mcs/service_export_controller.go
@@ -81,7 +81,7 @@ func (c *ServiceExportController) Reconcile(ctx context.Context, req controllerr
 		return controllerruntime.Result{}, nil
 	}
 
-	if !helper.IsWorkContains(&work.Status, serviceExportGVK) {
+	if !isWorkContains(work.Spec.Workload.Manifests, serviceExportGVK) {
 		return controllerruntime.Result{}, nil
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/controllers/mcs/service_export_controller.go
```diff
@@ -81,7 +81,7 @@ func (c *ServiceExportController) Reconcile(ctx context.Context, req controllerr
 		return controllerruntime.Result{}, nil
 	}
 
-	if !helper.IsWorkContains(&work.Status, serviceExportGVK) {
```
RainbowMango, line 84: > Therefore, the work status of the ServiceExport type should not be collected during work status collection. 

Do you mean here `helper.IsWorkContains` always return `false`? 

By the way, should we add `ServiceExport`/`ServiceImport` to interpreter framework?

XiShanYongYe-Chang: > Do you mean here helper.IsWorkContains always return false?

For now, even if the resource doesn't have status, we still will collect resource `ManifestStatuses.ResourceIdentifier`, `helper.IsWorkContains` is judging by  `ManifestStatuses.ResourceIdentifier`.

RainbowMango (approved):
/lgtm
/approve

PR #1386, opened 2022-02-22
[proposal] Caching member cluster resources for Karmada

huntsman-li:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

Caching member cluster resources for Karmada

**Which issue(s) this PR fixes**:
Fixes #1245 

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #1674, opened 2022-04-27
Add a CRD for the cached member cluster and resource scope.

### LINKED ISSUE #1245: [Proposal] caching member cluster resources

Caching member cluster resources

Scenarios:
- Global resource access/search (multi-cluster/cross-regional cluster)
- Global application view (multi-cluster application)
- Unified multi-cluster resource entry (for dashboard/cli)

**Why is this needed**:

- Accelerate (cross-regional) resource access
- High-frequency kubernetes resource access
- Aggregate multi-cluster resources

liys87x:
Co-authored-by: liys87x <liyasong1987x@gmail.com>
Signed-off-by: huntsman_ly <huntsman_ly@sina.com>

**What type of PR is this?**

/kind feature

Caching member cluster resources for Karmada

Fixes #1245 #1386 

proposal https://github.com/karmada-io/karmada/pull/1386

```release-note

```


RainbowMango:
Thanks @huntsman-li  @liys87x I'll look at the today.

pkg/apis/query/v1alpha1/doc.go
```diff
@@ -0,0 +1,5 @@
+// Package v1alpha1 is the v1alpha1 version of the API.
+// +k8s:deepcopy-gen=package,register
+// +k8s:openapi-gen=true
+// +groupName=query.karmada.io
```
RainbowMango, line 4: @kevin-wangzefeng @GitHubxsy Are you OK with this group?

pkg/apis/query/v1alpha1/cache_types.go
```diff
@@ -0,0 +1,151 @@
...
+	// ResourcePluralClusterCache is plural name of ClusterCache.
+	ResourcePluralClusterCache = "clusterCaches"
+	// ResourceNamespaceScopedClusterCache indicates if ClusterCache is not NamespaceScoped.
+	ResourceNamespaceScopedClusterCache = false
+)
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:shortName=cc
```
RainbowMango, line 22: Is it a `namespace-scoped` resource or `cluster-scoped`? 
If no specific instructions, it takes `namespace-scoped` by default.

I think it should be a `cluster-scoped` resource.
```suggestion
// +kubebuilder:resource:scope="Cluster"
```
Suggest add the `scope` and removes shortName (We can add it back later. )

pkg/apis/query/v1alpha1/cache_types.go
```diff
@@ -0,0 +1,151 @@
...
+
+// ClusterCacheSpec defines the desired state of ClusterCache
+type ClusterCacheSpec struct {
+	// ClusterSelectors represents the filter to select clusters.
+	// +required
+	ClusterSelectors []ClusterSelector `json:"clusterSelectors"`
+
+	// ResourceSelectors used to select resources.
+	// +required
+	ResourceSelectors []ResourceSelector `json:"resourceSelectors"`
```
RainbowMango, line 45: I'm not familiar with the index thing. I wonder to know do we usually set up `index`s on specific `kinds` or not?
In other words, is there any difference between the index for `Pods` and `Secret`? If yes, the `ResourceSelector` more like a `rules` about how to select resources and how to set up the index.

pkg/apis/query/v1alpha1/cache_types.go
```diff
@@ -0,0 +1,151 @@
...
+	ResourceNamespaceScopedClusterCache = false
+)
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:shortName=cc
+
+// ClusterCache defines a list of member cluster to be cached.
+type ClusterCache struct {
```
RainbowMango, line 25: @kevin-wangzefeng Could you please help to propose a resource name?

pkg/apis/query/v1alpha1/cache_types.go
liys87x: ok, add the scope and remove shortName.

pkg/apis/query/v1alpha1/cache_types.go
liys87x: I'm not sure if the user is required to decide the resource scope.

pkg/apis/query/v1alpha1/cache_types.go
```diff
@@ -0,0 +1,151 @@
...
+// CachePhase is the current state of the cache
+// +enum
+type CachePhase string
+
+// These are the valid statuses of cache.
+const (
+	CacheRunning CachePhase = "Running"
+	CacheFailed  CachePhase = "Failed"
+	CacheUnknown CachePhase = "Unknown"
+)
```
RainbowMango, line 131: I suppose we can remove the `CachePhase` filed for now, because we are not sure if we really need it yet.
And on the other hand, we do need another filed to explain the failed reason for `Failed` and `Unknown`.

In addition, according to [Kubernetes API convention](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status), the pattern of using `phase` is deprecated: 
> Some resources in the v1 API contain fields called phase, and associated message, reason, and other status fields. The pattern of using phase is deprecated. Newer API types should use conditions instead. Phase was essentially a state-machine enumeration field, that contradicted [system-design principles](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/architecture/principles.md#control-logic) and hampered evolution, since [adding new enum values breaks backward compatibility](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md).

We can add more `Conditions` once there is a need.

pkg/apis/query/v1alpha1/cache_types.go
```diff
@@ -0,0 +1,151 @@
...
+	// +optional
+	Namespace string `json:"namespace,omitempty"`
+}
+
+// ClusterCacheStatus defines the observed state of ClusterCache
+type ClusterCacheStatus struct {
+	// +optional
+	ResourceStatuses []ResourceStatusRef `json:"resources,omitempty"`
+	// +optional
+	StartTime *metav1.Time `json:"startTime,omitempty"`
```
RainbowMango, line 101: I'm not so confident with the status filed, how about leaving it empty for now, just add an empty `Condition`?, like:
```golang
// ClusterCacheStatus defines the observed state of ClusterCache
type ClusterCacheStatus struct {
	// Conditions contain the different condition statuses.
	// +optional
	Conditions []metav1.Condition `json:"conditions,omitempty"`
}
```

RainbowMango (commented):
Generally looks good to me except:
- 1. The name of the Kind (ClusterCache) (I'm still thinking if there are any better alternatives)
- 2. The status fields(I suggest leaving it empty for now, and adding it back once there is a need).

pkg/apis/query/v1alpha1/cache_types.go
liys87x: Good suggestion

RainbowMango: Yeah, thanks. It's easy to add more fields to it, but not to remove(deprecate) them.

RainbowMango:
Just discussed this API with @kevin-wangzefeng and @GitHubxsy locally, we propose following changes:
- Change the group name to `search.karmada.io`, the reason why select `search` is because it's more close to `open-search`.
- Change the kind name to `ResourceRegistry`, which means the user registers resources to the cache.

And, we also proposed the component name `karmada-search` which @XiShanYongYe-Chang will help build the draft structure.

RainbowMango:
@liys87x How do you think?

liys87x:
Good job, I will modify

RainbowMango (approved):
/lgtm
/approve

PR #1682, opened 2022-04-28
add update-lifted and verify-lifted

### LINKED ISSUE #1652: Automatic update lifted/doc.go
`lifted/doc.go` seems to be updated manually. It is better to do it automatically.

We can add lifted info in comments, like:
```
//+source: https://github.com/kubernetes/kubernetes/blob/release-1.23/pkg/apis/core/helper/helpers.go#L57-L61
//+changed: no

// IsQuotaHugePageResourceName returns true if the resource name has the quota
// related huge page resource prefix.
func IsQuotaHugePageResourceName(name core.ResourceName) bool {
	return strings.HasPrefix(string(name), core.ResourceHugePagesPrefix) || strings.HasPrefix(string(name), core.ResourceRequestsHugePagesPrefix)
}
```

Then we can parse it with `AST` and generate `doc.go`. 
`update-lifted.sh` and `verify-lifted.sh` can be added in hack.

**Why is this needed**:
Avoid `lifted/doc.go` out of date.

ikaven1024:

**What type of PR is this?**
/kind feature

Update and verify `lifted/doc.go` automatically.

Fixes #1652

```release-note
NONE
```


ikaven1024:
@XiShanYongYe-Chang @mrlihanbo @lonelyCZ

hack/verify-lifted.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+SCRIPT_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
+
+doc_file="pkg/util/lifted/doc.go"
+"$SCRIPT_ROOT/hack/update-lifted.sh"
```
XiShanYongYe-Chang, line 10: Is it possible to create a temporary file like any other verify script and then restore it after the comparison?

XiShanYongYe-Chang (commented):
This looks great, thanks for your contribution.

hack/verify-lifted.sh
ikaven1024: ok~

lonelyCZ:
LGTM, this is great. I think whether we could add a usage example or document for this feature later, which can help contributors to learn about this specification.

ikaven1024:
> I think whether we could add a usage example or document for this feature later,

add it at another pr.

ikaven1024:
@XiShanYongYe-Chang Plz take a look again.

XiShanYongYe-Chang (commented):
/lgtm

ikaven1024:
@RainbowMango could you approve it?

ikaven1024:
@RainbowMango  PTAL

RainbowMango:
OK. Try my best to look at it today. Sorry for the delay.

hack/tools/lifted/lifted.go
```diff
@@ -0,0 +1,243 @@
+package main
```
RainbowMango, line 1: What do you think about the name `lifted-gen`? Just like `client-gen`/`deepcopy-gen`? :)

`hack/tools/lifted-gen/main.go`.

hack/tools/lifted/lifted.go
```diff
@@ -0,0 +1,243 @@
...
+	"path/filepath"
+	"sort"
+	"strings"
+
+	"github.com/olekukonko/tablewriter"
+)
+
+const (
+	liftedCommentPrefix  = "//+lifted:"
+	liftedCommentChanged = "changed"
```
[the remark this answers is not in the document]
RainbowMango: No better ideas. I'd prefer the second one between the two.

ikaven1024:
@RainbowMango updated, and PTAL.

ikaven1024:
shall we moving this forward now 😂 @RainbowMango?

RainbowMango:
Yeah. Will look at it today.(try my best :))

RainbowMango:
We are ready to go now. Just a question, should we enable the verification at workflow?
https://github.com/karmada-io/karmada/blob/2a9d10628819a47a6da6ea4f637c24a1c6847066/.github/workflows/ci.yml#L24-L50

ikaven1024:
> should we enable the verification at workflow?

lgtm.    
I thought `verify-all.sh` is called in ci.

RainbowMango (approved):
/lgtm
/approve

PR #1706, opened 2022-05-01
fix empty resource selectors which result to policy will not be consi…

likakuli:
…dered matched

**What type of PR is this?**

/kind bug

Override policy with empty resourceselectors should effective as nil which means match every resource.

Fixes #

```release-note
override policy with empty resourceselectors will be considered to match all resource like with nil resourceselectors
```


RainbowMango:
@huone1 Please help with it.

likakuli:
Any problem with it?

RainbowMango:
Sorry for let this sit. I'll try my best to look at it today.

RainbowMango:
I guess we can't apply an empty `ResourceSelector`. E.g.
```yaml
kind: PropagationPolicy
metadata:
  name: nginx-propagation
spec:
  resourceSelectors:
  placement:
    clusterAffinity:
      clusterNames:
        - member1
        - member2
```
Set ResourceSelector to empty, and this is not allowed:
```
error: error validating "samples/nginx/propagationpolicy.yaml": error validating data: ValidationError(PropagationPolicy.spec): missing required field "resourceSelectors" in io.karmada.policy.v1alpha1.PropagationPolicy.spec; if you choose to ignore these errors, turn validation off with --validate=false
```

I'm not saying this patch is not acceptable, just want to know your cases? Am I missing something?

RainbowMango (approved):
/lgtm
/approve

PR #1713, opened 2022-05-02
fix typo in workload.example.io_workloads.yaml description

my-git9:
/kind cleanup

fix typo in workload.example.io_workloads.yaml description

### DIFF
--- a/examples/customresourceinterpreter/apis/workload.example.io_workloads.yaml
+++ b/examples/customresourceinterpreter/apis/workload.example.io_workloads.yaml
@@ -6424,7 +6424,7 @@ spec:
                               type: string
                             nfs:
                               description: 'NFS represents an NFS mount on the host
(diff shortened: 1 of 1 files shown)


RainbowMango:
cc @XiShanYongYe-Chang

XiShanYongYe-Chang:
`hack/verify-crdgen.sh` ignores the files in the `examples/customresourceinterpreter` directory.

RainbowMango:
Hi @my-git9 
We probably can't just edit the generated file. I agree with you that there is a typo. Once it gets fixed from Kubernetes, we can sync it back when we update dependencies.

my-git9:
> Hi @my-git9 We probably can't just edit the generated file. I agree with you that there is a typo. Once it gets fixed from Kubernetes, we can sync it back when we update dependencies.

OK,thanks for your review

Issue #1573, opened 2022-04-01
How to add or remove targetCluster in  propagationpolicy dynamically?


CharlesQQ:
**Please provide an in-depth description of the question you have**:
Some deployment's replicas be managed by cronhpa, but podtemplates should be managed by karmada; When  cronhpa object is created,  the cronhpa cluster should be added to targetCluster in propagationpolicy dynamically, vice versa; Only in this way, When deployment is updated By karmada, deployment in cronhpa cluster will also be updated
**What do you think about this question?**:
Is it possible to add a new interpreter webhook type that allows users to dynamically add or remove targetCluster in propagationpolicy, and webhook should be called when karmada deployment be updated
**Environment**:
- Karmada version:
- Kubernetes version:
- Others:

GitHubxsy:
When the number of deployment replicas is managed by hpa or cronhpa, Karmada should not update the replicas.When updating other fields such as image, Karmada needs to update it.
Is my understanding correct?

CharlesQQ:
@GitHubxsy correct

CharlesQQ:
@GitHubxsy @RainbowMango 
it seems that workload can only be propagated to  member cluster in replicaScheduling field; like following propagationpolicy spec,workload only can be  propagated to a and b; however, workload should be updated in cluster c when it is updated, Karmada should not update the replicas in cluster c, replicas can be retain by interpreter webhook
```
spec:
  placement:
    clusterAffinity:
      clusterNames:
      - a
      - b
      - c
    replicaScheduling:
      replicaDivisionPreference: Weighted
      replicaSchedulingType: Divided
      weightPreference:
        staticWeightList:
        - targetCluster:
            clusterNames:
            - a
            - b
          weight: 1
```

RainbowMango:
Both #1715 and #1720 are trying to solve this issue?

RainbowMango:
What's the `cronhpa` thing?

CharlesQQ:
`Both https://github.com/karmada-io/karmada/pull/1715 and https://github.com/karmada-io/karmada/pull/1720 are trying to solve this issue?`
yes

CharlesQQ:
when cronhpa object cteated, cronhpa targetCluster will be added to PP targetCluster list
![image](https://user-images.githubusercontent.com/18431718/167561981-2bc26052-a048-4012-8fdb-d8a972ff4096.png)

PR #1715, opened 2022-05-02
feat: add `clusterName` parameter for interperter webhook `retain`

CharlesQQ:
Signed-off-by: charlesQQ <charles_ali@qq.com>

**What type of PR is this?**
/kind api-change
/kind feature

interterter webhook server should know cluster name , and decide to  retain spec.replicas field
Fixes  https://github.com/karmada-io/karmada/issues/1573

```
interperter-webhook server

```

### DIFF
--- a/pkg/apis/config/v1alpha1/interpretercontext_types.go
+++ b/pkg/apis/config/v1alpha1/interpretercontext_types.go
@@ -64,6 +64,10 @@ type ResourceInterpreterRequest struct {
 	// AggregatedStatus represents status list of the resource running in each member cluster.
 	// +optional
 	AggregatedStatus []workv1alpha2.AggregatedStatusItem `json:"aggregatedStatus,omitempty"`
+
+	// ClusterName represents member cluster incoming request
+	// +optional
+	ClusterName string `json:"clusterName,omitempty"`
 }
 
 // ResourceInterpreterResponse describes an interpreter response.
--- a/pkg/generated/openapi/zz_generated.openapi.go
+++ b/pkg/generated/openapi/zz_generated.openapi.go
@@ -1096,6 +1096,13 @@ func schema_pkg_apis_config_v1alpha1_ResourceInterpreterRequest(ref common.Refer
 							},
 						},
 					},
+					"clusterName": {
+						SchemaProps: spec.SchemaProps{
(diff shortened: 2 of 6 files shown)


CharlesQQ:
I use the command `bash hack/update-codegen.sh` to update openapi, but file not changed all the time，where is the problem?

lonelyCZ:
> I use the command `bash hack/update-codegen.sh` to update openapi, but file not changed all the time，where is the problem?

I guess `deepcopy-gen` depend on `gopath`, you can find it at your `gopath`. I haven't found a solution to it yet, so I only copy it from `gopath`.

RainbowMango:
You can clone karmada at `$GOPATH/src/github.com/karmada-io/`.

CharlesQQ:
`You can clone karmada at $GOPATH/src/github.com/karmada-io/`
I do it and file not changed, but lots of generated files changed, i don't find problems

lonelyCZ:
You could try to run `hack/update-all.sh`. And do you use go version `v1.17`?

CharlesQQ:
@lonelyCZ 
`And do you use go version v1.17`
go version go1.17.1 darwin/arm64

CharlesQQ:
/assign @RainbowMango 
Hi, Can you check pr?  thanks

RainbowMango:
Sure. and sorry for the delay.

RainbowMango:
I guess you can set a `ClusterOverridePolicy` to add annotations, like:
```yaml
kind: ClusterOverridePolicy
metadata:
  name: foo
spec:
  resourceSelectors:
    - apiVersion: apps/v1
      kind: Deployment
  overrideRules:
    - targetCluster:
        clusterNames:
          - member1
      overriders:
        plaintext:
          - path: "/metadata/annotations"
            operator: add
            value:
              cluster.example.com: member1
```

RainbowMango:
This patch tries to extend the API to add some extra info to the webhook, it will work but not a good idea. What if the webhook also needs to know which region the workload belongs to? Should we extend the API again?

For the `Retain` interface, we have passed the whole object to the webhook, so if you want to transmit any information, you can set it to the `annotation`.

CharlesQQ:
``` so if you want to transmit any information, you can set it to the annotation.```
that a good idea, i can try it. thanks~

PR #1731, opened 2022-05-06
codegen will update in repository

### LINKED ISSUE #1730: update-codegen.sh will generate file out of repository
**What happened**:
When we run `hack/update-codegen.sh`, updated files will generated on `GOPATH`, not in repository.

files in repository are updated

change types and run`hack/update-codegen.sh`

**Environment**:
- Karmada version:
- Others:

ikaven1024:

**What type of PR is this?**
/kind bug

Fix file will not updated after run `hack/update-codegen.sh`

Fixes #1730

```release-note
NONE
```

### DIFF
--- a/hack/update-codegen.sh
+++ b/hack/update-codegen.sh
@@ -8,11 +8,29 @@ set -o pipefail
 REPO_ROOT=$(git rev-parse --show-toplevel)
 cd "${REPO_ROOT}"
 
-echo "Generating with deepcopy-gen"
 GO111MODULE=on go install k8s.io/code-generator/cmd/deepcopy-gen
+GO111MODULE=on go install k8s.io/code-generator/cmd/register-gen
+GO111MODULE=on go install k8s.io/code-generator/cmd/conversion-gen
+GO111MODULE=on go install k8s.io/code-generator/cmd/client-gen
+GO111MODULE=on go install k8s.io/code-generator/cmd/lister-gen
+GO111MODULE=on go install k8s.io/code-generator/cmd/informer-gen
+GO111MODULE=on go install k8s.io/code-generator/cmd/openapi-gen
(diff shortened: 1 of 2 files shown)


hack/update-codegen.sh
```diff
@@ -8,11 +8,29 @@ set -o pipefail
...
 export GOPATH=$(go env GOPATH | awk -F ':' '{print $1}')
 export PATH=$PATH:$GOPATH/bin
 
+go_path="${REPO_ROOT}/_go"
+cleanup() {
+  rm -rf "${go_path}"
+}
+trap "cleanup" EXIT SIGINT
+
+cleanup
```
RainbowMango, line 27: Why keep this since we already put the `cleanup` in `trap`?

RainbowMango:
That's indeed a relief for guys who don't clone Karmada in `GOPATH`. It always creates a `dummy`/`temp` GOPATH and ignores whether it is in GOPATH already.

hack/update-codegen.sh
ikaven1024: Copy it from: 
https://github.com/karmada-io/karmada/blob/2974563c6a14c9fd4b85bacc8ae44ec6013bf99b/hack/verify-codegen.sh#L15-L20

It's cannot be careful.

RainbowMango: Can you help confirm it? I don't think `cleanup` is needed here.

ikaven1024: `go_path` maybe contaminated by other programs/scripts, it is safe to clean it before generate files.

RainbowMango: Yeah, that makes sense to me.

RainbowMango (approved):
/lgtm
/approve

PR #1743, opened 2022-05-09
[fix] delete work status 'Resource' field

lvyanru8200:
Signed-off-by: lv <yanru.lv@daocloud.io>

**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #1746, opened 2022-05-09
Update crd, codegen and swagger of work status

AllenZMC:
**What type of PR is this?**
/kind feature

Since the resource field in work status is not used and deleted, we need to update work api, crd and codegen. See #1743
Fixes #

```release-note
rm  `Resource` field in work status
```

### DIFF
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -14381,7 +14381,6 @@
         "ordinal",
         "version",
         "kind",
-        "resource",
         "name"
       ],
       "properties": {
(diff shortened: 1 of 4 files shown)


RainbowMango (commented):
The `Resource` field is inherited from [kubernetes work-api](https://github.com/kubernetes-sigs/work-api/blob/586d707fdb2cdbdc0a98b3b5ee3dfbc5ca51cef0/pkg/apis/v1alpha1/work_types.go#L76-L77), even though we are not using this field, but we'd better keep it unchanged just as it is.

I'm sorry for I made a misleading `TODO` there:
```
TODO(RainbowMango): Consider remove Resource from Work API.
```
that means we should consider move this filed from Kubernetes work-api first.

AllenZMC:
OK, close it? @RainbowMango

RainbowMango:
Yeah, we can re-open it once the upstream removed this field.

PR #1788, opened 2022-05-13
fix CRDs with multi-versions cannot match pp

Poor12:

**What type of PR is this?**
/kind feature

For example，if users apply a CRD with multi-versions which has a storaged version named v1beta1 and served version named v1alpha1，when users apply v1alpha1 and create pp with v1alpha1 CR, this will not make sense due to we only create informers for CRD with preferred version(v1beta1) and resource selector is about v1alpha1.

Fixes #

None

### DIFF
--- a/pkg/util/lifted/discovery.go
+++ b/pkg/util/lifted/discovery.go
@@ -28,27 +28,27 @@ import (
 
 // GetDeletableResources returns all resources from discoveryClient that the
 // garbage collector should recognize and work with. More specifically, all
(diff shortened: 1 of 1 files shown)


RainbowMango:
@Poor12 Could you share the details about how to reproduce it? And share the test output here?

Poor12:
For example, we do have a CRD has two versions, one of them is storaged version:
image
Meanwhile, I create a CR with v1beta1, which is not preferred version and bind it with a pp(off course, resource selector is v1beta1). Now this CR will not propogate to member clusters because there is not a corresponding informer.

PR #1793, opened 2022-05-13
Set timeout configs flexible when run e2e before

wuyingjun-lucky:
Signed-off-by: wuyingjun <wuyingjun_yewu@cmss.chinamobile.com>

**What type of PR is this?**
/kind cleanup

When I run e2e test locally , I always failed  because out of time 
before :
image
image
We should set the config polltimeout  flexible
After 
image

Fixes #

```release-note
NONE
```

### DIFF
--- a/test/e2e/suite_test.go
+++ b/test/e2e/suite_test.go
@@ -2,6 +2,7 @@ package e2e
 
 import (
 	"context"
+	"flag"
 	"fmt"
 	"io"
 	"os"
@@ -31,11 +32,6 @@ import (
 )
 
 const (
-	// pollInterval defines the interval time for a poll operation.
-	pollInterval = 5 * time.Second
-	// pollTimeout defines the time after which the poll operation times out.
-	pollTimeout = 300 * time.Second
-
 	// RandomStrLength represents the random string length to combine names.
 	RandomStrLength = 3
 )
@@ -64,6 +60,13 @@ const (
 )
 
 var (
+	// pollInterval defines the interval time for a poll operation.
+	pollInterval time.Duration
+	// pollTimeout defines the time after which the poll operation times out.
+	pollTimeout time.Duration
+)
+
+var (
 	kubeconfig            string
 	restConfig            *rest.Config
 	karmadaHost           string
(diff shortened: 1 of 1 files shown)


XiShanYongYe-Chang:
Hi @wuyingjun-lucky.

IMOP, the timeout period of 5 minutes is long enough to check the matching conditions. If the timeout error persists during E2E running, a check maybe not be met. That is, an error does occur.

In addition, I am in favor of setting the timeout to be configurable. How about using the command line flags? Can refer to:
https://onsi.github.io/ginkgo/#supporting-custom-configuration-custom-command-line-flags

wuyingjun-lucky:
> https://onsi.github.io/ginkgo/#supporting-custom-configuration-custom-command-line-flags

Yes but every time I modify the pollTimeout It will sucess； If I do not modify the value， It always failed. 
Do you suggest close the pr or modfiy  the pr by using command line flags

XiShanYongYe-Chang:
> Yes but every time I modify the pollTimeout It will sucess； If I do not modify the value， It always failed.

For the wrong e2e case, can you reproduce it and then check the environment with the five minutes.

> Do you suggest close the pr or modfiy the pr by using command line flags

You can modify it on the current PR.

If possible, modify the label of the `kind bug`.

test/e2e/suite_test.go
```diff
@@ -77,6 +80,16 @@ var (
 	pushModeClusterLabels = map[string]string{"sync-mode": "Push"}
 )
 
+func init() {
+	var pollIntervalValue, pollTimeoutValue int
+
+	flag.IntVar(&pollIntervalValue, "poll-interval-value", 300, "Interval value of seconds for poll operation")
+	flag.IntVar(&pollTimeoutValue, "poll-timeout-value", 5, "Value of seconds which the poll operation times out")
```
XiShanYongYe-Chang, line 87: The default value of those two flags may be mixed up.

hack/run-e2e.sh
```diff
@@ -37,8 +37,11 @@ REPO_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
 export KUBECONFIG=${KARMADA_APISERVER_KUBECONFIG}
 export PULL_BASED_CLUSTERS=${PULL_BASED_CLUSTERS}
 
+POLL_INTERVAl_VALUE=${POLL_INTERVAl_VALUE:-"5"}
+POLL_TIMEOUT_VALUE=${POLL_TIMEOUT_VALUE:-"300"}
+
 set +e
-ginkgo -v --race --trace --fail-fast -p --randomize-all ./test/e2e/
+ginkgo -v --race --trace --fail-fast -p --randomize-all ./test/e2e/ -- --poll-interval-value=${POLL_INTERVAl_VALUE} --poll-timeout-value=${POLL_TIMEOUT_VALUE}
```
XiShanYongYe-Chang, line 44: The two flags may not need to be added for now.

If needed, we can add it later.

XiShanYongYe-Chang (commented):
👍

test/e2e/suite_test.go
wuyingjun-lucky: I am sorry I has not finished testing and the pull request was not ready at first. 
Now I finish testing including the value passed from command to program .
Besides I try to use `/kind cleanup`to replace the label, But the origin label `kind bug` has not removed. Can you help to modify it ?

XiShanYongYe-Chang (commented):
/lgtm
/approve

test/e2e/suite_test.go
XiShanYongYe-Chang: Updated.

PR #1814, opened 2022-05-17
[refactor] refactor make build

ikaven1024:

**What type of PR is this?**
/kind flake

This is subsequent to #1740. In this pr:
1. Add `release` in Makefile
2. Simplify `make upload-images`
3. Merge target list in `Makefile` and `hack/util.sh`
4. Adjust make output format.

Fixes #

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,5 @@
 GOOS ?= $(shell go env GOOS)
 GOARCH ?= $(shell go env GOARCH)
-SOURCES := $(shell find . -type f  -name '*.go')
 LDFLAGS='$(shell hack/version.sh)'
 
 # Images management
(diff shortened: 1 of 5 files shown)


ikaven1024:
@RainbowMango PTAL

RainbowMango:
> Add release in Makefile

I like this approach. Is it possible to split this function into a separate PR. 
except `kubectl-karmada`, can we also make the release for `karmadactl`?

If we can generate the checksum(sha256) for the `tar` ball would be a wonderful thing.

ikaven1024:
> I like this approach. Is it possible to split this function into a separate PR. except `kubectl-karmada`, can we also make the release for `karmadactl`?
> 
> If we can generate the checksum(sha256) for the `tar` ball would be a wonderful thing.

ikaven1024:
> Is it possible to split this function into a separate PR. except `kubectl-karmada`, can we also make the release for `karmadactl`?
> 
> If we can generate the checksum(sha256) for the `tar` ball would be a wonderful thing.

Good idea!

PR #1818, opened 2022-05-18
Update default version

my-git9:
/kind cleanup

Update default version about image in doc


RainbowMango:
Thanks for doing this. Can you post the test result here?
When updating the image version, some parameters might have to update as well.

my-git9:
> Thanks for doing this. Can you post the test result here? When updating the image version, some parameters might have to update as well.

This is just to modify the default value in the document table and synchronize it with 'value.yaml' in the chart

RainbowMango:
Oh, right. Sorry for the misunderstanding.

/lgtm
/approve

PR #1822, opened 2022-05-18
fixed a bug that the service could not be promoted and e2e test

### LINKED ISSUE #1758: Kubectl Karmada Promete can not take over some GVRs
**What happened**:
Some panics or problems will happens if we promete any GVRs in k8s.Such as promete a service will generate a panic,promete a ns will generate a error too.
图片

limit GVR (deployments,job,crd and  so on ) when excute kubectl karmada promete. Provide a flag to propagate dependencies(secrets and configmap).Or resovle the problems one by one to support any resources?

**Environment**:
- Karmada version:
- Others:

duanmengkk:

**What type of PR is this?**

/kind bug

Fixes #1758

```release-note
NONE
```


pkg/util/objectwatcher/objectwatcher.go
```diff
@@ -69,32 +69,31 @@ func (o *objectWatcherImpl) Create(clusterName string, desireObj *unstructured.U
...
 		// - 2. The resource already exist in the member cluster but it's not created by karmada.
 		if apierrors.IsAlreadyExists(err) {
-			existObj, err := dynamicClusterClient.DynamicClientSet.Resource(gvr).Namespace(desireObj.GetNamespace()).Get(context.TODO(), desireObj.GetName(), metav1.GetOptions{})
-			if err != nil {
-				return fmt.Errorf("failed to get exist resource(kind=%s, %s/%s) in cluster %v: %v", desireObj.GetKind(), desireObj.GetNamespace(), desireObj.GetName(), clusterName, err)
-			}
-
 			// If the existing resource is managed by Karmada, then just update it.
 			if util.GetLabelValue(desireObj.GetLabels(), workv1alpha1.WorkNameLabel) == util.GetLabelValue(existObj.GetLabels(), workv1alpha1.WorkNameLabel) {
 				return o.Update(clusterName, desireObj, existObj)
```
chaunceyjiang, line 83: `apierrors.IsAlreadyExists(err)`  will it be false forever?

when will this code run?

duanmengkk: > `apierrors.IsAlreadyExists(err)` will it be false forever?
> 
> when will this code run?

if promote a nodeport service from member cluster to control plane,`apierrors.IsAlreadyExists(err)` is true

pkg/util/objectwatcher/objectwatcher.go
```diff
@@ -69,32 +69,31 @@ func (o *objectWatcherImpl) Create(clusterName string, desireObj *unstructured.U
...
 	if err != nil {
+		existObj, err := dynamicClusterClient.DynamicClientSet.Resource(gvr).Namespace(desireObj.GetNamespace()).Get(context.TODO(), desireObj.GetName(), metav1.GetOptions{})
+		if err != nil {
+			return fmt.Errorf("failed to get exist resource(kind=%s, %s/%s) in cluster %v: %v", desireObj.GetKind(), desireObj.GetNamespace(), desireObj.GetName(), clusterName, err)
+		}
 		// The 'IsAlreadyExists' conflict may happen in following known scenarios:
 		// - 1. In a reconcile process, the execution controller successfully applied resource to member cluster but failed to update the work conditions(Applied=True),
 		//   when reconcile again, the controller will try to apply(by create) the resource again.
 		// - 2. The resource already exist in the member cluster but it's not created by karmada.
 		if apierrors.IsAlreadyExists(err) {
```
chaunceyjiang, line 80: > if promote a nodeport service from member cluster to control plane,apierrors.IsAlreadyExists(err) is true

Ah?    line 73  `err != nil` has returned ,  line 76 `err` it must be `nil` ,therefore `apierrors.IsAlreadyExists(err)` must be false. 

right ?

duanmengkk: > > if promote a nodeport service from member cluster to control plane,apierrors.IsAlreadyExists(err) is true
> 
> Ah? line 73 `err != nil` has returned , line 76 `err` it must be `nil` ,therefore `apierrors.IsAlreadyExists(err)` must be false.
> 
> right ?

sorry,I misunderstand before.Excution_controller will create a resource in member cluster instead of update it by the manifest in works .So if a reosurce is exist in member cluster,`apierrors.IsAlreadyExists(err)` will return true.However, if a reource is a nodeport service,create if will return notInvalid err.

chaunceyjiang: yes ,l also found this problems , if a reource is a nodeport service,create it will return [provided port is already allocated](https://github.com/kubernetes/kubernetes/blob/master/pkg/registry/core/service/portallocator/allocator.go#L43)

pkg/util/objectwatcher/objectwatcher.go
```diff
@@ -107,6 +112,22 @@ func (o *objectWatcherImpl) Create(clusterName string, desireObj *unstructured.U
 	return nil
 }
 
+// promoteResource update the resource if a conflict resolution instruction is existed in annotation.
+// The existing resource is not managed by Karmada, then we should consult conflict resolution instruction in annotation.
+func (o *objectWatcherImpl) promoteResource(clusterName string, desireObj *unstructured.Unstructured, existObj *unstructured.Unstructured) error {
```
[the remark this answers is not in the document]
duanmengkk: How about `resourceConflictOverwrite`?

RainbowMango:
cc @lonelyCZ @XiShanYongYe-Chang 
Please help take a look.

pkg/util/objectwatcher/objectwatcher.go
```diff
@@ -78,23 +78,28 @@ func (o *objectWatcherImpl) Create(clusterName string, desireObj *unstructured.U
...
-					desireObj.GetKind(), desireObj.GetNamespace(), desireObj.GetName(), clusterName, workv1alpha2.ResourceConflictResolutionAnnotation,
-				)
+			// update the resource if a conflict resolution instruction is existed in annotation.
+			err = o.promoteResource(clusterName, desireObj, existObj)
+			if err != nil {
+				return fmt.Errorf("failed to update exist resource(kind=%s, %s/%s) in cluster %v: %v", desireObj.GetKind(), desireObj.GetNamespace(), desireObj.GetName(), clusterName, err)
+			}
+		}
+		//  The 'IsInvalid' conflict will happen when a resource is created if the resource kind is 'Service'
+		if apierrors.IsInvalid(err) {
```
XiShanYongYe-Chang, line 92: How about merge those two if branch:

```go
apierrors.IsAlreadyExists(err) ||
			/* The 'IsInvalid' conflict will happen when a resource is created if the resource kind is 'Service'*/ (apierrors.IsInvalid(err) && desireObj.GetKind() == util.ServiceKind) {
```

test/helper/resource.go
```diff
@@ -152,6 +152,32 @@ func NewService(namespace string, name string) *corev1.Service {
 	}
 }
 
+// NewNodePortService will build a nodeport service object.
+func NewNodePortService(namespace string, name string) *corev1.Service {
```
XiShanYongYe-Chang, line 156: We can add a new input parameter to the `NewService` method to indicate the service type.

test/helper/resource.go
```diff
@@ -128,27 +128,45 @@ func NewStatefulSet(namespace string, name string) *appsv1.StatefulSet {
...
-		TypeMeta: metav1.TypeMeta{
-			APIVersion: "v1",
-			Kind:       "Service",
-		},
-		ObjectMeta: metav1.ObjectMeta{
-			Namespace: namespace,
-			Name:      name,
-		},
-		Spec: corev1.ServiceSpec{
+func NewService(namespace string, name string, svcType corev1.ServiceType) *corev1.Service {
```
XiShanYongYe-Chang, line 131: How about this:

```diff
--- a/test/helper/resource.go
+++ b/test/helper/resource.go
@@ -128,7 +128,7 @@ func NewStatefulSet(namespace string, name string) *appsv1.StatefulSet {
 }

 // NewService will build a service object.
-func NewService(namespace string, name string) *corev1.Service {
+func NewService(namespace string, name string, svcType corev1.ServiceType) *corev1.Service {
        return &corev1.Service{
                TypeMeta: metav1.TypeMeta{
                        APIVersion: "v1",
@@ -147,7 +147,7 @@ func NewService(namespace string, name string) *corev1.Service {
                                        TargetPort: intstr.IntOrString{IntVal: 8080},
                                },
                        },
-                       Type: corev1.ServiceTypeClusterIP,
+                       Type: svcType,
                },
        }
```

duanmengkk: I have a test in my env.
A nodeport should be defined clearly,or isInvalid error will not happen 
 
image

duanmengkk:
Hi,ask you a question,the error output in lint is not same with my code,what is the problem? @XiShanYongYe-Chang 
image
image

test/helper/resource.go
XiShanYongYe-Chang: A random port will be assigned to nodePort, refer to https://kubernetes.io/docs/concepts/services-networking/service/#type-nodeport

I think this error should not occur.

XiShanYongYe-Chang:
> Hi,ask you a question,the error output in lint is not same with my code,what is the problem? @XiShanYongYe-Chang

This maybe refer to the lint error in line 199.

test/helper/resource.go
duanmengkk: > A random port will be assigned to nodePort, refer to https://kubernetes.io/docs/concepts/services-networking/service/#type-nodeport
> 
> I think this error should not occur.

yes,but the e2e_test is to test a fixed nodeport service cannot be promoted,so the nodeport args is needed.

duanmengkk:
> > Hi,ask you a question,the error output in lint is not same with my code,what is the problem? @XiShanYongYe-Chang
> 
> This maybe refer to the lint error in line 199.

there is no change in line 199, `./hack/verify-staticcheck.sh` works fine in my env. I'll look into it

test/helper/resource.go
XiShanYongYe-Chang: I probably understand what you mean, but I see that there is no testing for this in e2e.

duanmengkk: > but I see that there is no testing for this in e2e.

I added in karmactl_test.go

XiShanYongYe-Chang: New added e2e test run failed:
```log
 Unexpected error:
      <*errors.errorString | 0xc0005573c0>: {
          s: "failed to create resource \"/v1, Resource=services\"(karmadatest-wp7/service-jtg) in control plane: Service \"service-jtg\" is invalid: spec.ports[0].nodePort: Invalid value: 30002: provided port is already allocated",
      }
      failed to create resource "/v1, Resource=services"(karmadatest-wp7/service-jtg) in control plane: Service "service-jtg" is invalid: spec.ports[0].nodePort: Invalid value: 30002: provided port is already allocated
  occurred
  In [It] at: /home/runner/work/karmada/karmada/test/e2e/karmadactl_test.go:251
```

test/helper/resource.go
```diff
@@ -128,7 +128,21 @@ func NewStatefulSet(namespace string, name string) *appsv1.StatefulSet {
...
+func NewService(namespace string, name string, svcType corev1.ServiceType, nodePort int32) *corev1.Service {
+	port := corev1.ServicePort{
+		Name:       "http",
+		Protocol:   corev1.ProtocolTCP,
+		Port:       80,
+		TargetPort: intstr.IntOrString{IntVal: 8080},
+	}
+	// specify a random nodePort explicitly rather than handing it over to K8S
+	// as create same a explicit nodePort service, a invalid error will occur
+	if svcType == corev1.ServiceTypeNodePort && nodePort != 0 {
```
XiShanYongYe-Chang, line 141: Hi @duanmengkk, IMOP, the `NewService` function is just a template and should not have too complex logic in it.
We can set a nodePort for NodePort type Service after we created a service.

duanmengkk: ok.i will fix it

duanmengkk: /cc @XiShanYongYe-Chang take a look,all reviews are fixed

XiShanYongYe-Chang: ok~

XiShanYongYe-Chang (commented):
Thanks!

/lgtm

pkg/util/objectwatcher/objectwatcher.go
```diff
@@ -73,28 +73,20 @@ func (o *objectWatcherImpl) Create(clusterName string, desireObj *unstructured.U
 		// - 1. In a reconcile process, the execution controller successfully applied resource to member cluster but failed to update the work conditions(Applied=True),
 		//   when reconcile again, the controller will try to apply(by create) the resource again.
 		// - 2. The resource already exist in the member cluster but it's not created by karmada.
-		if apierrors.IsAlreadyExists(err) {
+		//  The 'IsInvalid' conflict will happen when a resource is created if the resource kind is 'Service'
```
RainbowMango, line 76:
```suggestion
		// The 'IsInvalid' conflict will happen when a resource is created if the resource kind is 'Service'
```

RainbowMango: Can you provide the reproduce steps for the `Invalid` error?

XiShanYongYe-Chang: +1

When I recreate a Service same as before created, the err info is:

```
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {},
  "status": "Failure",
  "message": "error when creating \"nginx-svc.yaml\": services \"my-nginx\" already exists",
  "reason": "AlreadyExists",
  "details": {
    "name": "my-nginx",
    "kind": "services"
  },
  "code": 409
```

chaunceyjiang: reproduce  https://github.com/karmada-io/karmada/issues/1942

XiShanYongYe-Chang: I think there may be a situation that can't be handled properly: when a `NodePort` type Service is created and an allocated nodePort number is specified.

duanmengkk: > when a `NodePort` type Service is created and an allocated nodePort number is specified.

Yes,it is right,invalid err only happens when nodePort number is specified. `karmadactl promote service` will create a service that nodeport number is specified in member cluster,so I ignored it before.I will fix it soon.

RainbowMango:
The reproduce steps on my side:

S1: create a `service` on `member1` cluster:
```yaml
# nodeport-svc.yaml 
apiVersion: v1
kind: Service
metadata:
  name: serve
spec:
  ports:
  - port: 80
    targetPort: 8080
  selector:
    app: serve
  type: NodePort
```
```bash
$ export KUBECONFIG=/root/.kube/members.config;kubectl config use-context member1
$ kubectl apply -f nodeport-svc.yaml
```

S2: promote dry-run 
```bash
karmadactl promote service serve -c member1 -o yaml 
apiVersion: v1
kind: Service
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"v1","kind":"Service","metadata":{"annotations":{},"name":"serve","namespace":"default"},"spec":{"ports":[{"port":80,"targetPort":8080}],"selector":{"app":"serve"},"type":"NodePort"}}
    work.karmada.io/conflict-resolution: overwrite
  name: serve
  namespace: default
spec:
  externalTrafficPolicy: Cluster
  internalTrafficPolicy: Cluster
  ipFamilies:
  - IPv4
  ipFamilyPolicy: SingleStack
  ports:
  - nodePort: 32238
    port: 80
    protocol: TCP
    targetPort: 8080
  selector:
    app: serve
  sessionAffinity: None
  type: NodePort
---
kind: PropagationPolicy
metadata:
  creationTimestamp: null
  name: serve-propagation
  namespace: default
spec:
  placement:
    clusterAffinity:
      clusterNames:
      - member1
  resourceSelectors:
  - apiVersion: v1
    kind: Service
    name: serve

```

S3: promote the service to Karmda control plan
```
$ export KUBECONFIG=/root/.kube/karmada.config
$ kubectl config use-context karmada-apiserver
$ kubectl config use-context karmada-apiserver
$ karmadactl promote service serve -c member1
```
S4: check the log of `karmada-controller-manager`:
```
E0607 12:37:27.516812       1 objectwatcher.go:100] Failed to create resource(kind=Service, default/serve) in cluster member1, err is Service "serve" is invalid: spec.ports[0].nodePort: Invalid value: 32238: provided port is already allocated 
E0607 12:37:27.516830       1 execution_controller.go:206] Failed to create resource(default/serve) in the given member cluster member1, err is Service "serve" is invalid: spec.ports[0].nodePort: Invalid value: 32238: provided port is already allocated
```

RainbowMango:
The root cause is that when creating the `service` to `member1`, we get the `Invalid` error but the expected `AlreadyExists`.

RainbowMango:
Personally, I don't want to do anything special to the `Service` resources. 

I guess we can optimize the current [apply logic](https://github.com/karmada-io/karmada/blob/6ceb11d475bce73f4db6ddca49e1fa7decadec52/pkg/controllers/execution/execution_controller.go#L194-L214). But no idea about how to do it yet.

duanmengkk:
> Personally, I don't want to do anything special to the `Service` resources.

Agree,I am afraid of that `service `is not only one resource that return other error besides  `AlreadyExists` when creating it.

duanmengkk:
> I guess we can optimize the current [apply logic](https://github.com/karmada-io/karmada/blob/6ceb11d475bce73f4db6ddca49e1fa7decadec52/pkg/controllers/execution/execution_controller.go#L194-L214). But no idea about how to do it yet.

I will research on it ASAP.

duanmengkk:
> Personally, I don't want to do anything special to the `Service` resources.
> 
> I guess we can optimize the current [apply logic](https://github.com/karmada-io/karmada/blob/6ceb11d475bce73f4db6ddca49e1fa7decadec52/pkg/controllers/execution/execution_controller.go#L194-L214). But no idea about how to do it yet.

Why not get the resource first  to check whether the resource exists,instead of check whether the resource exists by `apierrors.IsAlreadyExists(err)`, Is it because it leads to more performance loss  @RainbowMango 

for  example
```
	clusterObj, err := dynamicClusterClient.DynamicClientSet.Resource(gvr).Namespace(desireObj.GetNamespace()).Get(context.TODO(), desireObj.GetName(), metav1.GetOptions{})
	if clusterObj != nil {
		// Decide if we need to update it
		.....
	}
```

XiShanYongYe-Chang:
@duanmengkk I have the same idea. We can get the resource from the cache, just as:

```
clusterObj, err := helper.GetObjectFromCache(c.RESTMapper, c.InformerManager, fedKey)
```

RainbowMango:
> Why not get the resource first to check whether the resource exists,instead of check whether the resource exists by apierrors.IsAlreadyExists(err), Is it because it leads to more performance loss?

We are in a context of `creating` an object, it'll be weird to `get` the object first, in addition, in most cases we don't need to do that.

duanmengkk:
> > Why not get the resource first to check whether the resource exists,instead of check whether the resource exists by apierrors.IsAlreadyExists(err), Is it because it leads to more performance loss?
> 
> We are in a context of `creating` an object, it'll be weird to `get` the object first, in addition, in most cases we don't need to do that.

In my hand，if a GVK is determined,such as deployment, propagationpolicy,configmap...  We can creating it directly to check if `apierrors.IsAlreadyExists(err)`,which can get a short path.   but if a resource is not determined,we can't. see https://github.com/kubernetes/kubernetes/blob/7c127b33dafc530f7ca0c165ddb47db86eb45880/staging/src/k8s.io/apiserver/pkg/registry/generic/registry/store.go#L396-L422 , `isInvalid `error is always before `isAlready` err ,Any CRD may cause a `IsInvalid `error instead of `isAlready` err besides Service.

duanmengkk:
> @duanmengkk I have the same idea. We can get the resource from the cache, just as:
> 
> ```
> clusterObj, err := helper.GetObjectFromCache(c.RESTMapper, c.InformerManager, fedKey)
> ```

What if the cache is not synchronized in time? @XiShanYongYe-Chang

XiShanYongYe-Chang:
@duanmengkk How about updating like this:

image

duanmengkk:
> @duanmengkk How about updating like this:
> 
> image

This is what i thought,but @RainbowMango seems have some reservations about get GVR first to check wether it exists.Do you have some other suggestion? 

And, what if the cache is not synchronized in time? @XiShanYongYe-Chang

XiShanYongYe-Chang:
> This is what i thought,but @RainbowMango seems have some reservations about get GVR first to check wether it exists.Do you have some other suggestion?

I have just a face-to-face discussion with him, For the moment, this method is relatively simple. We can modify it in the way.

> And, what if the cache is not synchronized in time?

If the cache is not synchronized in time, that means we may choose to create the resource. In fact, the resource exists, so the creation will fail and return with an error, then the logic will reconcile the resource the next time.

duanmengkk:
> > This is what i thought,but @RainbowMango seems have some reservations about get GVR first to check wether it exists.Do you have some other suggestion?
> 
> I have just a face-to-face discussion with him, For the moment, this method is relatively simple. We can modify it in the way.
> 
> > And, what if the cache is not synchronized in time?
> 
> If the cache is not synchronized in time, that means we may choose to create the resource. In fact, the resource exists, so the creation will fail and return with an error, then the logic will reconcile the resource the next time.

ok

XiShanYongYe-Chang (commented):
@duanmengkk, Sorry for letting this pr hang for so long. Thanks for your contribution!

/lgtm

/cc @RainbowMango

PR #1829, opened 2022-05-18
Add cluster failure threshold

dddddai:
**What type of PR is this?**
/kind feature

Currently cluster status will be set to NotReady immediately if the check failed
To make it more robust, this pr adds cluster failure threshold(default to 30s), which is the duration of failure for the cluster to be considered unhealthy.

Fixes #

Should the flag `cluster-failure-threshold` be the number of failures(default to 3) or the time(default to 30s)? Which is better?

```release-note
`karmada-controller-manager/karmada-agent`: Introduced `--cluster-failure-threshold` flag to specify cluster failure threshold.
```


RainbowMango:
> Should the flag cluster-failure-threshold be the number of failures(default to 3) or the time(default to 30s)? Which is better?

What's your option?

dddddai:
Either is fine with me, time is probably more intuitive but users need to keep `cluster-failure-threshold` a multiple of `cluster-status-update-frequency`

pkg/controllers/status/cluster_status_controller.go
```diff
@@ -309,24 +300,42 @@ func (c *ClusterStatusController) initLeaseController(ctx context.Context, clust
 	}()
 }
 
-func getClusterHealthStatus(clusterClient *util.ClusterClient) (online, healthy bool) {
```
RainbowMango, line 312: I don't think we need to put more logic into this function. 
This function just tells the result of this time, and don't have to know how we use this result.

pkg/controllers/status/cluster_status_controller.go
```diff
@@ -156,31 +157,30 @@ func (c *ClusterStatusController) syncClusterStatus(cluster *clusterv1alpha1.Clu
...
 		klog.Errorf("Failed to get or create informer for Cluster %s. Error: %v.", cluster.GetName(), err)
-		return c.setStatusCollectionFailedCondition(cluster, currentClusterStatus, fmt.Sprintf("failed to get or create informer: %v", err))
 	}
 
 	// init the lease controller for every cluster
 	c.initLeaseController(clusterInformerManager.Context(), cluster)
 
 	clusterVersion, err := getKubernetesVersion(clusterClient)
 	if err != nil {
-		return c.setStatusCollectionFailedCondition(cluster, currentClusterStatus, fmt.Sprintf("failed to get kubernetes version: %v", err))
```
RainbowMango, line 167: Why change this behavior? I think it's out of this PR's scope.

cmd/agent/app/agent.go
```diff
@@ -223,6 +224,8 @@ func startClusterStatusController(ctx controllerscontext.Context) (bool, error)
 		ClusterStatusUpdateFrequency:      ctx.Opts.ClusterStatusUpdateFrequency,
 		ClusterLeaseDuration:              ctx.Opts.ClusterLeaseDuration,
 		ClusterLeaseRenewIntervalFraction: ctx.Opts.ClusterLeaseRenewIntervalFraction,
+		ClusterFailureThreshold:           ctx.Opts.ClusterFailureThreshold,
+		ClusterFailureCounter:             make(map[string]int),
```
RainbowMango, line 228: We can initialize it inside the controller. See here: https://github.com/karmada-io/karmada/pull/1781#discussion_r874274821

pkg/controllers/status/cluster_status_controller.go
```diff
@@ -131,20 +130,22 @@ func (c *ClusterStatusController) syncClusterStatus(cluster *clusterv1alpha1.Clu
...
+	c.Lock()
+	if online && healthy {
+		c.ClusterFailureCounter[cluster.Name] = 0
+	} else if c.ClusterFailureCounter[cluster.Name] < c.ClusterFailureThreshold {
+		// retry a few times to avoid network unstable problems.
+		c.ClusterFailureCounter[cluster.Name]++
+		c.Unlock()
+		return controllerruntime.Result{RequeueAfter: c.ClusterStatusUpdateFrequency.Duration}, nil
+	}
+	c.Unlock()
```
RainbowMango, line 143: Much better than the last commit.
Can we finish it by this PR? Please consider adding a unit test for it.

dddddai:
Thank you, I've addressed your comments

pkg/controllers/status/cluster_status_controller.go
```diff
@@ -82,6 +78,10 @@ type ClusterStatusController struct {
 	ClusterLeaseRenewIntervalFraction float64
 	// ClusterLeaseControllers store clusters and their corresponding lease controllers.
 	ClusterLeaseControllers sync.Map
+	// ClusterFailureThreshold is the duration of failure for the cluster to be considered unhealthy.
+	ClusterFailureThreshold metav1.Duration
+	// ClusterFailureTimestamp is the timestamp of the first failed cluster health check.
+	ClusterFailureTimestamp map[string]*time.Time
```
RainbowMango, line 84: I guess we'll add a success threshold too, then we need to introduce another map? as well as a lock?

RainbowMango:
I really want something like [ClusterData](https://github.com/kubernetes-sigs/kubefed/blob/ba0bf5273832cbf616c763c6887bbb2b1e41968a/pkg/controller/kubefedcluster/controller.go#L46) which caches each cluster's last prob time and status. How do you think?

RainbowMango:
By the way, we should clean up the data when deleting the cluster. 
https://github.com/karmada-io/karmada/blob/92a134b00fb77936921a595d14dd9a68be91e18a/pkg/controllers/status/cluster_status_controller.go#L98-L102

dddddai:
> I really want something like [ClusterData](https://github.com/kubernetes-sigs/kubefed/blob/ba0bf5273832cbf616c763c6887bbb2b1e41968a/pkg/controller/kubefedcluster/controller.go#L46) which caches each cluster's last prob time and status. How do you think?

For "last prob time", kubefed has [its own Condition](https://github.com/kubernetes-sigs/kubefed/blob/ba0bf5273832cbf616c763c6887bbb2b1e41968a/pkg/apis/core/v1beta1/kubefedcluster_types.go#L110) but we are using `metav1.Condition` where we don't have "last prob time"

For "status", kubefed stores it because it will [restore the old status](https://github.com/kubernetes-sigs/kubefed/blob/ba0bf5273832cbf616c763c6887bbb2b1e41968a/pkg/controller/kubefedcluster/controller.go#L282), we just return and retry in the next health check, I think they take the same effect

So I don't see where we are gonna use it? Please correct me if I'm wrong

pkg/controllers/status/cluster_status_controller.go
dddddai: Not really, we can change the map value from `time` to `{successTime, failureTime}`

pkg/controllers/status/cluster_status_controller.go
```diff
@@ -61,6 +56,7 @@ var (
 
 // ClusterStatusController is to sync status of Cluster.
 type ClusterStatusController struct {
+	sync.Mutex                  // used to lock the ClusterFailureCounter.
```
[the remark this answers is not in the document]
dddddai: Sorry, fixed

RainbowMango:
What I mean `something like` ClusterData, learn the idea, not the whole struct.
And `last prob time` means last success or last failure time the time will be changed only when the condition change.

dddddai:
OK, added

pkg/controllers/status/cluster_status_controller.go
```diff
@@ -195,6 +201,30 @@ func (c *ClusterStatusController) syncClusterStatus(cluster *clusterv1alpha1.Clu
 	return c.updateStatusIfNeeded(cluster, currentClusterStatus)
 }
 
+func (c *ClusterStatusController) retryOnFailure(online, healthy bool, cluster string) bool {
```
RainbowMango, line 204: I can't tell but seems something doesn't feel right here.
Let me think about it.

dddddai: Have modified as you suggested

pkg/controllers/status/cluster_condition_cache_test.go
```diff
@@ -0,0 +1,119 @@
...
+			healthy:       false,
+			expectedReady: metav1.ConditionTrue,
+		},
+		{
+			// the third retry
+			// cluster is still unhealthy after the threshold, set cluster status to not ready
+			online:        false,
+			healthy:       false,
+			expectedReady: metav1.ConditionFalse,
+		},
```
RainbowMango, line 98: Seems these tests interact with each other?

dddddai: Yes thery are observed statuses ordered by time, because `thresholdAdjustedReadyCondition` is stateful, it depends on previous statuses

RainbowMango: In this test, we are testing `thresholdAdjustedReadyCondition`, we should set up `cache`, `current condition`, and `observed condition` for each case. They should not depend on each other.

dddddai: That would be kind of tricky..How to set up the the `cache`?(`clusterDataMap` and `thresholdStartTime`)

dddddai: We are testing `clusterConditionStore.thresholdAdjustedReadyCondition` where `clusterConditionStore` is stateful

RainbowMango:
```golang
func TestThresholdAdjustedReadyCondition2(t *testing.T) {
	tests := []struct {
		name              string
		clusterData       *clusterData
		currentCondition  *metav1.Condition
		observedCondition *metav1.Condition
		expectedCondition *metav1.Condition
	}{
		{
			name:              "cluster just joined in ready state",
			clusterData:       nil, // no cache yet
			currentCondition:  nil, // no condition was set on cluster object yet
			observedCondition: &metav1.Condition{Status: metav1.ConditionTrue},
			expectedCondition: &metav1.Condition{Status: metav1.ConditionTrue},
		},
		{
			name:              "cluster just joined in not-ready state",
			clusterData:       nil, // no cache yet
			currentCondition:  nil, // no condition was set on cluster object yet
			observedCondition: &metav1.Condition{Status: metav1.ConditionFalse},
			expectedCondition: &metav1.Condition{Status: metav1.ConditionFalse},
		},
		{
			name: "cluster becomes not ready but still not reach threshold",
			clusterData: &clusterData{
				readyCondition:     metav1.ConditionFalse,
				thresholdStartTime: time.Now(),
			},
			currentCondition:  &metav1.Condition{Status: metav1.ConditionTrue},
			observedCondition: &metav1.Condition{Status: metav1.ConditionFalse},
			expectedCondition: &metav1.Condition{Status: metav1.ConditionTrue},
		},
	}

	for _, test := range tests {
		t.Run(test.name, func(t *testing.T) {
			cache := clusterConditionStore{
				clusterDataMap:   sync.Map{},
				failureThreshold: 10,
			}

			if test.clusterData != nil {
				cache.update("foo", test.clusterData)
			}

			cluster := &clusterv1alpha1.Cluster{}
			cluster.Name = "foo"
			if test.currentCondition != nil {
				meta.SetStatusCondition(&cluster.Status.Conditions, *test.currentCondition)
			}

			thresholdReadyCondition := cache.thresholdAdjustedReadyCondition(cluster, test.observedCondition)

			if test.expectedCondition.Status != thresholdReadyCondition.Status {
				t.Fatalf("expected: %s, but got: %s", test.expectedCondition.Status, thresholdReadyCondition.Status)
			}
		})
	}
}

```

RainbowMango:
Just tried add some cases(not finished).

RainbowMango:
I tested it on my side, I think we are ready to move further. Thanks @dddddai

RainbowMango (approved):
/lgtm
/approve

PR #1720, opened 2022-05-04
feat: allow propagate workload to cluster but replicas is zero

CharlesQQ:
Signed-off-by: charlesQQ <charles_ali@qq.com>

**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
allow propagate workload to cluster but replicas is zero; if don't want to do that, cluster should not exist in PP spec.clusterAffinity.clusterNames field
**Which issue(s) this PR fixes**:
Fixes #1573

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```
rb's spec.clusters have cluster that replicas is zero

scheduler  add a  new command line  option  enable-propagate-empty-workLoad
```

PR #1842, opened 2022-05-20
supply unit test to cover zero replica case

RainbowMango:
**What type of PR is this?**
/kind cleanup

Add more unit tests.
Clarify the behavior with zero replicas. Refer to #1720 

Fixes #

```release-note
NONE
```


RainbowMango:
@Garrybest here is an extreme example that propagates deployments with `0` replica to clusters by static weight.
Seems it doesn't obey the rules we discussed at #1720.

What's the expected behavior in your opinion ?

Garrybest:
Well, I think this case meets our expectation. There is no need to add more logics in to `StaticWeight`. We just divide replicas by a fixed proportion. https://github.com/karmada-io/karmada/pull/1720 focuses more about dynamic dividing.

RainbowMango:
@Garrybest I added comments to the function and add one more unit test for it. Please help take another look.

CharlesQQ:
So In case of not enough replica for specific cluster which should be ignored?

Garrybest:
That's it.
/lgtm

PR #1851, opened 2022-05-20
remove the source for logo

my-git9:
/kind bug

When I try to install a Karmada using this script, I get the following error：
```
# ./hack/remote-up-karmada.sh ~/.kube/config local
go install github.com/cloudflare/cfssl/cmd/...@v1.5.0
./hack/../hack/../hack/util.sh:行71: go: 未找到命令
```
Golang needs to be installed, but I found that this script does not need to be compiled by golang, and the reason why this util.sh script is sourced is only because a karmada logo is required. I think it's a bit unreasonable for users to have to install a golang on the client just because of a logo.

About the commit: https://github.com/karmada-io/karmada/commit/003717a0c6b9df6bc1b662c0ac03868bb9fb5e18

### DIFF
--- a/hack/remote-up-karmada.sh
+++ b/hack/remote-up-karmada.sh
@@ -4,6 +4,20 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
+#https://textkool.com/en/ascii-art-generator?hl=default&vl=default&font=DOS%20Rebel&text=KARMADA
+KARMADA_GREETING='
+------------------------------------------------------------------------------------------------------
+ █████   ████   █████████   ███████████   ██████   ██████   █████████   ██████████     █████████
+░░███   ███░   ███░░░░░███ ░░███░░░░░███ ░░██████ ██████   ███░░░░░███ ░░███░░░░███   ███░░░░░███
+ ░███  ███    ░███    ░███  ░███    ░███  ░███░█████░███  ░███    ░███  ░███   ░░███ ░███    ░███
+ ░███████     ░███████████  ░██████████   ░███░░███ ░███  ░███████████  ░███    ░███ ░███████████
+ ░███░░███    ░███░░░░░███  ░███░░░░░███  ░███ ░░░  ░███  ░███░░░░░███  ░███    ░███ ░███░░░░░███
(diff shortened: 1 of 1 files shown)


RainbowMango:
Maybe it's not the best approach. The `KARMADA_GREETING` is shared by scripts. We shouldn't put it into each script.

my-git9:
> Maybe it's not the best approach. The `KARMADA_GREETING` is shared by scripts. We shouldn't put it into each script.

Do you have a better idea, I still think it's unacceptable to install golang because of a logo

my-git9:
> Maybe it's not the best approach. The `KARMADA_GREETING` is shared by scripts. We shouldn't put it into each script.

Currently it looks like only two scripts will print this logo

my-git9:
Do we need to use a separate script like `karmada-logo.sh` to store this logo

RainbowMango:
> and the reason why this util.sh script is sourced is only because a karmada logo is required.

Is that true? Only `source` util.sh shouldn't check the `go` command, I guess. Remind me if I'm wrong.

> go install github.com/cloudflare/cfssl/cmd/...@v1.5.0

`deploy-karmada.sh` is required by `remote-up-karmada.sh` and it requires `go` command.

my-git9:
Sorry, I only looked at this commit and util.sh, didn't check deploy-karmada.sh

my-git9:
> > and the reason why this util.sh script is sourced is only because a karmada logo is required.
> 
> Is that true? Only `source` util.sh shouldn't check the `go` command, I guess. Remind me if I'm wrong.
> 
> > go install github.com/cloudflare/cfssl/cmd/...@v1.5.0
> 
> `deploy-karmada.sh` is required by `remote-up-karmada.sh` and it requires `go` command.

Do you think we can check the relevant commands before installing like in local-up-karmada.sh, like `kubectl`, `go`？

RainbowMango:
> Do you think we can check the relevant commands before installing like in local-up-karmada.sh, like kubectl, go？

I'm not sure I fully understand what you mean. 
As a general answer, I'd say each script just like a `modle` has its responsibility. Let's see.

> Sorry, I only looked at this commit and util.sh, didn't check deploy-karmada.sh

No worries. Thanks all the same.

Issue #1853, opened 2022-05-20
Use taint instead of condition in karmada scheduler


likakuli:
**What would you like to be added**:

use taint to filter cluster instead of using cluster ready condition.

**Why is this needed**:

User story:

Support set not ready clusters to rb.  For now, not ready cluster will be filtered.

for example:

pp's cluster affinity has tow cluster: A and B, and B is not ready for some reason now.

if i add another ready cluster C to pp's cluster affinity, the rb.spec.cluster will be A and C，which will result to resources already dispatched to cluster B be deleted. This is what we not expected to happen. Just let rb.spec.clusters be A and B and C even if cluster B is not ready, karmada-controller-manager will retry to make resource dispatched to cluster B when it becomes ready.

Use taint instead of condition could make it easy to be controlled by user who has the requirement. The only thing need to do is to set pp's clustertolerations as actual requirement.

duanmengkk:
refer to https://github.com/karmada-io/karmada/issues/1762,  They are working for it now.

likakuli:
> refer to #1762, They are working for it now.

thx，I found the taint has already been added. so i just make a pr to update karmada scheduler to make it not filter cluster by condition.

likakuli:
https://github.com/karmada-io/karmada/issues/1762 is for evict, use `TaintEffectNoExecute`, taint is added when failover eviction timeout, but what i need is to add `TaintEffectNoSchedule` immediately when cluster not ready.

Garrybest:
Hi @likakuli, we'are going to stop evicting already-scheduled result in scheduler. So I think this problem may be fixed after https://github.com/karmada-io/karmada/issues/1762.

likakuli:
I don't have any requirement for eviction. In my case, i wish the resource  can be scheduled to clusters specified in pp's placement whether the cluster is ready or not and whether the api  is installed in the cluster or not.

Maybe i can directly operate the resoucebinding to skip the karmada scheduler?

Garrybest:
It seems like you want to disable some plugins, right?

Now all plugins are enabled by default and cannot be configurable via config file. There is a plan to optimize config files. https://github.com/karmada-io/karmada/issues/1605#issuecomment-1098848260 may be useful.

likakuli:
https://github.com/karmada-io/karmada/blob/9a72c70085db4b15eef05c881c7f1cead1896941/pkg/scheduler/core/generic_scheduler.go#L96

[...]

PR #1854, opened 2022-05-20
feat: use taint instead of condition to filter cluster

likakuli:

**What type of PR is this?**
/kind feature

let users to make decision whether to use unhealthy cluster in rb.

Fixes https://github.com/karmada-io/karmada/issues/1853

```release-note
None
```

### DIFF
--- a/pkg/controllers/cluster/cluster_controller.go
+++ b/pkg/controllers/cluster/cluster_controller.go
@@ -39,16 +39,31 @@ const (
 
 var (
 	// UnreachableTaintTemplate is the taint for when a cluster becomes unreachable.
+	// Used for taint based eviction.
 	UnreachableTaintTemplate = &corev1.Taint{
 		Key:    clusterv1alpha1.TaintClusterUnreachable,
 		Effect: corev1.TaintEffectNoExecute,
 	}
 
-	// NotReadyTaintTemplate is the taint for when a cluster is not ready for
-	// executing resources.
+	// UnreachableTaintTemplateForSched is the taint for when a cluster becomes unreachable.
+	// Used for taint based schedule.
+	UnreachableTaintTemplateForSched = &corev1.Taint{
+		Key:    clusterv1alpha1.TaintClusterUnreachable,
+		Effect: corev1.TaintEffectNoSchedule,
+	}
+
+	// NotReadyTaintTemplate is the taint for when a cluster is not ready for executing resources.
+	// Used for taint based eviction.
 	NotReadyTaintTemplate = &corev1.Taint{
 		Key:    clusterv1alpha1.TaintClusterNotReady,
 		Effect: corev1.TaintEffectNoExecute,
+	}
+
+	// NotReadyTaintTemplateForSched is the taint for when a cluster is not ready for executing resources.
+	// Used for taint based schedule.
+	NotReadyTaintTemplateForSched = &corev1.Taint{
+		Key:    clusterv1alpha1.TaintClusterNotReady,
+		Effect: corev1.TaintEffectNoSchedule,
 	}
 )
 
@@ -144,7 +159,7 @@ func (c *Controller) Reconcile(ctx context.Context, req controllerruntime.Reques
 		return c.removeCluster(cluster)
 	}
 
-	return c.syncCluster(cluster)
+	return c.syncCluster(ctx, cluster)
 }
 
 // Start starts an asynchronous loop that monitors the status of cluster.
@@ -172,11 +187,18 @@ func (c *Controller) SetupWithManager(mgr controllerruntime.Manager) error {
 	})
 }
 
-func (c *Controller) syncCluster(cluster *clusterv1alpha1.Cluster) (controllerruntime.Result, error) {
+func (c *Controller) syncCluster(ctx context.Context, cluster *clusterv1alpha1.Cluster) (controllerruntime.Result, error) {
 	// create execution space
 	err := c.createExecutionSpace(cluster)
 	if err != nil {
 		c.EventRecorder.Event(cluster, corev1.EventTypeWarning, fmt.Sprintf("Failed %s", clusterv1alpha1.EventReasonCreateExecutionSpaceFailed), err.Error())
+		return controllerruntime.Result{Requeue: true}, err
+	}
+
+	// taint cluster by condition
+	err = c.taintClusterByCondition(ctx, cluster)
+	if err != nil {
+		c.EventRecorder.Event(cluster, corev1.EventTypeWarning, fmt.Sprintf("Failed %s", clusterv1alpha1.EventReasonTaintClusterByConditionFailed), err.Error())
 		return controllerruntime.Result{Requeue: true}, err
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/scheduler/core/generic_scheduler.go
```diff
@@ -93,7 +93,8 @@ func (g *genericScheduler) findClustersThatFit(
 	defer metrics.ScheduleStep(metrics.ScheduleStepFilter, time.Now())
 
 	var out []*clusterv1alpha1.Cluster
-	clusters := clusterInfo.GetReadyClusters()
+	// DO NOT filter unhealthy cluster, let users make decisions by using ClusterTolerations of Placement.
```
Garrybest, line 96: Hi @likakuli, I think we still need to filter ready clusters. The taint is not added into a cluster immediately when a cluster becomes unhealthy until `FailoverEvictionTimeout` is up, default to 5 minutes.  So in this period of time, we'd better not consider these clusters as schedulable.

likakuli: > The taint is not added into a cluster immediately

I see, so the code to add taint when cluster not healthy is added above the `FailoverEvictionTimeout ` judge.

Garrybest: That's right. You could add this taint when reconciling.

likakuli: yeah, make sense, let me do it

likakuli: done. Add a method `taintClusterByCondition` in `syncCluster`.

pkg/controllers/cluster/cluster_controller.go
```diff
@@ -511,3 +533,37 @@ func (c *Controller) processTaintBaseEviction(ctx context.Context, cluster *clus
 	}
 	return nil
 }
+
+func (c *Controller) taintClusterByCondition(ctx context.Context, cluster *clusterv1alpha1.Cluster) error {
+	// Cluster object may be changed in this function.
+	observedReadyCondition, currentReadyCondition, err := c.tryUpdateClusterHealth(ctx, cluster)
```
Garrybest, line 539: There is no need to update cluster health when reconciling because it is a little bit expensive. I suggest retrieve conditions from cluster status directly. BTW, [here](https://github.com/kubernetes/kubernetes/blob/1868a3a2eb9945f26239f700fc491fb37232db5a/pkg/controller/nodelifecycle/node_lifecycle_controller.go#L611-L665) is some references.

Garrybest:
Thanks👍
/lgtm

pkg/controllers/cluster/cluster_controller.go
likakuli: done

RainbowMango:
I'm thinking should we put the logic of managing taint together?

https://github.com/karmada-io/karmada/blob/d7ad34afa60779803aa8625baaeed129b4dcfc7b/pkg/controllers/cluster/cluster_controller.go#L345-L350

@Garrybest How do you think?

RainbowMango:
I'm thinking should we put the logic of managing taint together?

https://github.com/karmada-io/karmada/blob/d7ad34afa60779803aa8625baaeed129b4dcfc7b/pkg/controllers/cluster/cluster_controller.go#L345-L350

@Garrybest How do you think?

Garrybest:
> I'm thinking should we put the logic of managing taint together?
> 
> https://github.com/karmada-io/karmada/blob/d7ad34afa60779803aa8625baaeed129b4dcfc7b/pkg/controllers/cluster/cluster_controller.go#L345-L350
> 
> @Garrybest How do you think?

Well, I'm afraid not. We monitor cluster health every 5s, but we want to add this taint immediately when cluster become not ready. Otherwise the scheduler will consider this cluster schedulable in this period.

pkg/controllers/cluster/cluster_controller.go
```diff
@@ -511,3 +533,28 @@ func (c *Controller) processTaintBaseEviction(ctx context.Context, cluster *clus
 	}
 	return nil
 }
+
+func (c *Controller) taintClusterByCondition(ctx context.Context, cluster *clusterv1alpha1.Cluster) error {
+	currentReadyCondition := meta.FindStatusCondition(cluster.Status.Conditions, clusterv1alpha1.ClusterConditionReady)
+
+	if currentReadyCondition != nil {
```
RainbowMango, line 540: probably the last question.
Why skip the `nil` condition here?

likakuli: Ready condition nil represent that cluster_status_controller never update cluster status. so we can't determine how to deal with the taint.
Other operation like eviction also has this judgement too.

RainbowMango: > Other operation like eviction also has this judgement too.

Share a link?

likakuli: https://github.com/karmada-io/karmada/blob/a42c819c101e850ec89f15f8fd7cb8f2b9bf94e9/pkg/controllers/cluster/cluster_controller.go#L367-L373

likakuli: Is this ok?

RainbowMango: Sorry I missed this. 
I'm thinking if we should add a taint to the cluster when it is registering. cc @Garrybest

Garrybest: Should we? It seems that K8s does not add a taint when a node is registering.

RainbowMango: cc @prodanlabs @pidb who heavily uses the scheduler.

pkg/scheduler/core/generic_scheduler.go
```diff
@@ -93,7 +93,8 @@ func (g *genericScheduler) findClustersThatFit(
 	defer metrics.ScheduleStep(metrics.ScheduleStepFilter, time.Now())
 
 	var out []*clusterv1alpha1.Cluster
-	clusters := clusterInfo.GetReadyClusters()
```
RainbowMango, line 96: According to https://github.com/karmada-io/karmada/issues/1853#issuecomment-1141593401,

You mentioned that your cluster  `is default cordoned`, does that means you will set the taint(key: `cluster.karmada.io/unschedulable`, effect=`NoSchedule`) to the cluster? Just similar thing like `karmadactl taint`  do. 

My question here is even if a cluster has taints, it won't be filtered by `GetReadyClusters()` as the taint won't lead to condition change, right? 

I'm still confused about the real use case.

likakuli: User story:

There are k8s resources which need to be dispatched to member clusters filtered by selector before member cluster can serve. Just like daemonset for single cluster.

There are two limit in karmada

1. `karmada-scheduler` will not trigger a schedule cycle if the object has been scheduled successfully once even if there are new clusters which satisfied the pp's labelSelector added. 
2. `karmada-scheduler` will not trigger a schedule cycle if the object has been scheduled successfully once even if  there are some clusters whose status changed from notready to ready. 

So the object can't be dispatched to clusters new added or status changed from notready to ready. 

For case 1: we added a custom controller to update rb directly.
For case 2: we use this pr to let the notready clusters can also be set to `rb.spec.clusters` and work will be created. When the cluster's status changes from notready to ready, the work can be executed successfully.

RainbowMango (approved):
This patch looks good from Karmada's point of view. (though, I didn't so sure about your use case.)
/approve

PR #1858, opened 2022-05-21
make buildInformerForCluster configurable

snowplayfire:
**What type of PR is this?**

/kind feature

When member cluster is large(3k nodes, 10w pods), karmada-controller-manager is easy to be OOMkilled and retry to sync nodes and pods again and again, the member cluster also stays in `unknown`.

Whether to execute `buildInformerForCluster()` to sync nodes and pods should be configurable  especially  for large cluster.

```release-note
A new flag `build-informer-for-cluster` for controller-manager and agent.
```

### DIFF
--- a/cmd/agent/app/options/options.go
+++ b/cmd/agent/app/options/options.go
@@ -107,6 +107,8 @@ type Options struct {
 
 	// ClusterRegion represents the region of the cluster locate in.
 	ClusterRegion string
+
+	EnableClusterResourceModeling bool
 }
 
 // NewOptions builds an default scheduler options.
@@ -174,6 +176,9 @@ func (o *Options) AddFlags(fs *pflag.FlagSet, allControllers []string) {
 	fs.StringVar(&o.MetricsBindAddress, "metrics-bind-address", ":8080", "The TCP address that the controller should bind to for serving prometheus metrics(e.g. 127.0.0.1:8088, :8088)")
 	fs.StringVar(&o.ClusterProvider, "cluster-provider", "", "Provider of the joining cluster. The Karmada scheduler can use this information to spread workloads across providers for higher availability.")
 	fs.StringVar(&o.ClusterRegion, "cluster-region", "", "The region of the joining cluster. The Karmada scheduler can use this information to spread workloads across regions for higher availability.")
+	fs.BoolVar(&o.EnableClusterResourceModeling, "enable-cluster-resource-modeling", true, "Enable means controller would build resource modeling for each cluster by syncing Nodes and Pods resources.\n"+
+		"The resource modeling might be used by the scheduler to make scheduling decisions in scenario of dynamic replica assignment based on cluster free resources.\n"+
+		"Disable if it does not fit your cases for better performance.")
 	o.RateLimiterOpts.AddFlags(fs)
 	o.ProfileOpts.AddFlags(fs)
 }
--- a/pkg/controllers/status/cluster_status_controller.go
+++ b/pkg/controllers/status/cluster_status_controller.go
@@ -86,6 +86,8 @@ type ClusterStatusController struct {
 
 	ClusterCacheSyncTimeout metav1.Duration
 	RateLimiterOptions      ratelimiterflag.Options
+
+	EnableClusterResourceModeling bool
 }
 
 // Reconcile syncs status of the given member cluster.
@@ -154,7 +156,7 @@ func (c *ClusterStatusController) syncClusterStatus(cluster *clusterv1alpha1.Clu
 	}
 
 	// skip collecting cluster status if not ready
-	if online && healthy && readyCondition.Status == metav1.ConditionTrue {
+	if online && healthy && readyCondition.Status == metav1.ConditionTrue && c.EnableClusterResourceModeling {
 		// get or create informer for pods and nodes in member cluster
 		clusterInformerManager, err := c.buildInformerForCluster(cluster)
 		if err != nil {
--- a/cmd/agent/app/agent.go
+++ b/cmd/agent/app/agent.go
@@ -253,6 +253,7 @@ func startClusterStatusController(ctx controllerscontext.Context) (bool, error) 
 		ClusterFailureThreshold:           ctx.Opts.ClusterFailureThreshold,
 		ClusterCacheSyncTimeout:           ctx.Opts.ClusterCacheSyncTimeout,
 		RateLimiterOptions:                ctx.Opts.RateLimiterOptions,
+		EnableClusterResourceModeling:     ctx.Opts.EnableClusterResourceModeling,
 	}
 	if err := clusterStatusController.SetupWithManager(ctx.Mgr); err != nil {
 		return false, err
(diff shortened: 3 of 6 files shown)


RainbowMango:
> When member cluster is large(3k nodes, 10w pods), karmada-controller-manager is easy to be OOMkilled

That's very bad. 
I wonder to know how many memory reserved for `karmada-controller-manager`?

dddddai:
Yeah I have the same concern, karmada stores all pods and nodes in the informer cache, which consumes lots of memory in large scale scenarios

IMO we can [transform objects in informers](https://github.com/kubernetes/kubernetes/pull/107507), this is a new feature in client-go 0.24 while we are using 0.23 now

RainbowMango:
@dddddai Great idea, I like this feature! Just created #1866 to track this. Let's do it at release-1.3.

Before that feature, I still want to know more details about how bad the situation is now. It can also evaluate the effect of this feature in the future.

snowplayfire:
> > When member cluster is large(3k nodes, 10w pods), karmada-controller-manager is easy to be OOMkilled
> 
> That's very bad. I wonder to know how many memory reserved for `karmada-controller-manager`?

One cluster with 3k nodes and 10w pods needs rough100G for `kube-controller-manager`,  so `karmada-controller-manager` may need 100G * n, n is the number of clusters.

RainbowMango:
> One cluster with 3k nodes and 10w pods needs rough100G( far exceed 100G when cluster failure occurs) for kube-controller-manager, so karmada-controller-manager may need more than 100G * n, n is the number of clusters.

I'm very surprised! @Poor12 is leading the effort to do performance testing and improvement. He might confirm it later.

But if we don't sync node and pods here, we can't get the resource summary, then the general estimator will always deny workloads as `no resources left` .  
https://github.com/karmada-io/karmada/blob/d9b043bedd4507b9f81196f55ff1bd3fd5df6ab0/pkg/estimator/client/general.go#L36-L39

snowplayfire:
@RainbowMango  member cluster probably can  calculate resource summary and report to fed cluster.

RainbowMango:
> @RainbowMango member cluster probably can calculate resource summary and report to fed cluster.

I'm not sure if you are saying that `karmada-agent` now takes the responsibility to report resource summary for clusters in `Pull` mode.

RainbowMango:
I want to let you know that @Poor12 is trying to figure out how serious the issue is.

Let's wait for a while. cc guys who met this situation here.
cc @likakuli

likakuli:
copy that, thx

RainbowMango:
@snowplayfire 
We have been doing performance tests on this for the last few weeks. Now we can confirm that there is a performance issue but seems not so bad as mentioned on https://github.com/karmada-io/karmada/pull/1858#issuecomment-1134170421.

There are several improvement ideas:
- 1. change the `generic informer` to `typed informer` which is addressed by #2318 
- 2. Add the cache then we don't need to list all things out. it is addressed by #2034 
- 3. Adopt informer transform func to save memory. It is addressed by #2089 

I think it's fair enough to make a flag for people to enable/disable the `resource-modeling` thing.
Since we are going to make changes to the code that this PR also touching, I'd like to move this forward first to avoid conflict.

I'm sorry for let this sit so long, now there has conflicts that need to be resolved, please let me know if you can solve them and push again.

cmd/agent/app/options/options.go
```diff
@@ -108,5 +110,6 @@ func (o *Options) AddFlags(fs *pflag.FlagSet, allControllers []string) {
 	fs.DurationVar(&o.ResyncPeriod.Duration, "resync-period", 0, "Base frequency the informers are resynced.")
 	fs.IntVar(&o.ConcurrentClusterSyncs, "concurrent-cluster-syncs", 5, "The number of Clusters that are allowed to sync concurrently.")
 	fs.IntVar(&o.ConcurrentWorkSyncs, "concurrent-work-syncs", 5, "The number of Works that are allowed to sync concurrently.")
+	fs.BoolVar(&o.BuildInformerForCluster, "build-informer-for-cluster", true, "Enable this when need cluster informer.")
```
RainbowMango, line 113: I suggest change the flag to `--enable-cluster-resource-modeling`:
```suggestion
	fs.BoolVar(&o.EnableClusterResourceModeling, "enable-cluster-resource-modeling", true,
		"Enable means controller would build resource modeling for each cluster by syncing Nodes and Pods resources.\n"+
			"The resource modeling might be used by the scheduler to make scheduling decisions in scenario of dynamic replica assignment based on cluster free resources.\n"+
			"Disable if it does not fit your cases for better performance.")
```

Also, I'm thinking if we should add a field to `Cluster` API.

pkg/controllers/status/cluster_status_controller.go
```diff
@@ -152,15 +154,30 @@ func (c *ClusterStatusController) syncClusterStatus(cluster *clusterv1alpha1.Clu
...
+			return c.setStatusCollectionFailedCondition(cluster, currentClusterStatus, fmt.Sprintf("failed to list pods: %v", err))
+		}
+
+		currentClusterStatus.NodeSummary = getNodeSummary(nodes)
+		currentClusterStatus.ResourceSummary = getResourceSummary(nodes, pods)
 	}
 
 	// init the lease controller for every cluster
-	c.initLeaseController(clusterInformerManager.Context(), cluster)
+	c.initLeaseController(context.TODO(), cluster)
```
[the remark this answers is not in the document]
RainbowMango: Yes, you are right. We expect the lease controller could be stopped after the cluster is gone.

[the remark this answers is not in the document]
snowplayfire: done

cmd/agent/app/options/options.go
snowplayfire: change the flag to --enable-cluster-resource-modeling, done

snowplayfire: `add a field to Cluster API`, a field to declare synced resources?

RainbowMango: > add a field to Cluster API, a field to declare synced resources?

never mind. Let's finish this first.

RainbowMango: > change the flag to --enable-cluster-resource-modeling, done

I also proposed the flag `usage`, how do you think?
```
"Enable means controller would build resource modeling for each cluster by syncing Nodes and Pods resources.\n"+
			"The resource modeling might be used by the scheduler to make scheduling decisions in scenario of dynamic replica assignment based on cluster free resources.\n"+
			"Disable if it does not fit your cases for better performance."
```

RainbowMango:
Hi @snowplayfire Your branch has conflicts with the master, you might need to rebase your branch.

By the way, I see you are using your Chinese name in commit `author`:
```
Author: ljx李静雪 <xxx@trip.com>
```

cmd/agent/app/options/options.go
snowplayfire: aggree

snowplayfire:
> Hi @snowplayfire Your branch has conflicts with the master, you might need to rebase your branch.
> 
> By the way, I see you are using your Chinese name in commit `author`:
> 
> ```
> Author: ljx李静雪 <xxx@trip.com>
> ```

done

RainbowMango:
Hi @snowplayfire I picked your commit and fixed the conflicts as well as my comments at #2387.

PR #1881, opened 2022-05-25
[feature]Support helm chart propagation

Poor12:
**What type of PR is this?**
/kind feature

Support helm chart propagation.

Fixes #

None


hack/tools/swagger/lib/render.go
```diff
@@ -115,6 +115,7 @@ func RenderOpenAPISpec(cfg Config) (string, error) {
 	}
 
 	// Create Swagger Spec.
+	//nolint:staticcheck
```
RainbowMango, line 118: What's the problem here?

PR #1883, opened 2022-05-25
add Error message when unjoining cluster with a non-existent clustername

xyz2277:
Signed-off-by: bruce <zhangyongxi_yewu@cmss.chinamobile.com>

**What type of PR is this?**
/kind flake

when unjoining a cluster with a non-existent clustername, there is no any output on the console. It's better to output a Error message, I think...
unjoin
Fixes #

none

### DIFF
--- a/pkg/karmadactl/unjoin.go
+++ b/pkg/karmadactl/unjoin.go
@@ -271,7 +271,7 @@ func deleteClusterObject(controlPlaneKarmadaClient *karmadaclientset.Clientset, 
 
 	err := controlPlaneKarmadaClient.ClusterV1alpha1().Clusters().Delete(context.TODO(), opts.ClusterName, metav1.DeleteOptions{})
 	if apierrors.IsNotFound(err) {
-		return nil
+		return fmt.Errorf("no cluster object %s found in karmada control Plane", opts.ClusterName)
 	}
 	if err != nil {
 		klog.Errorf("Failed to delete cluster object. cluster name: %s, error: %v", opts.ClusterName, err)


pkg/karmadactl/unjoin.go
```diff
@@ -271,6 +271,7 @@ func deleteClusterObject(controlPlaneKarmadaClient *karmadaclientset.Clientset,
 
 	err := controlPlaneKarmadaClient.ClusterV1alpha1().Clusters().Delete(context.TODO(), opts.ClusterName, metav1.DeleteOptions{})
 	if apierrors.IsNotFound(err) {
+		fmt.Printf("Error:No cluster object %s found in karmada control Plane\n", opts.ClusterName)
```
lonelyCZ, line 274: We could directly delete these error check `apierrors.IsNotFound(err)`.

xyz2277: > We could directly delete these error check `apierrors.IsNotFound(err)`.

I think, deleting this error check apierrors.IsNotFound(err) would skip the stage of deleting the cluster role, cluster rolebindings and service account from the unjoining cluster. Only cluster object in karmada control Plane has been  deleted, in some situation, it's needed to re-execute unjoin command to delete the above-mentioned resources of unjoining cluster.

lonelyCZ: >  Only cluster object in karmada control Plane has been deleted, in some situation, it's needed to re-execute unjoin command to delete the above-mentioned resources of unjoining cluster.

Ok, we can replace `fmt.Printf` with `klog`?

xyz2277: > > Only cluster object in karmada control Plane has been deleted, in some situation, it's needed to re-execute unjoin command to delete the above-mentioned resources of unjoining cluster.
> 
> Ok, we can replace `fmt.Printf` with `klog`?

I prefer "fmt.Printf" , because it is more official and eye-catching than "klog".

lonelyCZ: Mainly it is not a actural error, only a promption. Generally, if report `Error: `, the command is stoped immidetely.

xyz2277: > 

I agree with you. Should I remove the "Error:" and use "fmt.Printf"?

xyz2277: > Mainly it is not a actural error, only a promption. Generally, if report `Error: `, the command is stoped immidetely.

I agree with you. Should I remove the "Error:" and use "fmt.Printf"?

xyz2277: > Mainly it is not a actural error, only a promption. Generally, if report `Error: `, the command is stoped immidetely.

I agree with you.

lonelyCZ: I was going to replace all the `klogs` in cmd with `fmt` before, but it wasn't all done. If we use `fmt` only at this point in `unjoin cmd`, the format would be strange.

xyz2277: > 

yeah, I've replace it with klog.

lonelyCZ:
Could you please provide your test results? I would like to see the print content. @xyz2277  Thanks~

xyz2277:
> Could you please provide your test results? I would like to see the print content. @xyz2277 Thanks~

lonelyCZ:
I  was considering if we should give more accurate prompt. Acturally we want to prompt this satituation:

> Only cluster object in karmada control Plane has been deleted, in some situation, it's needed to re-execute unjoin command to delete the above-mentioned resources of unjoining cluster.

The Error reported only seems very confused to users.

xyz2277:
> I was considering if we should give more accurate prompt. Acturally we want to prompt this satituation:
> 
> > Only cluster object in karmada control Plane has been deleted, in some situation, it's needed to re-execute unjoin command to delete the above-mentioned resources of unjoining cluster.
> 
> The Error reported only seems very confused to users.

When rejoining cluster member1,  it is what happens:
join

I think we can just refer it.
unjoin

lonelyCZ:
> Only cluster object in karmada control Plane has been deleted, in some situation, it's needed to re-execute unjoin command to delete the above-mentioned resources of unjoining cluster.

I think this behavior seems not our expection. We should not delete the remain resource of member cluster that has been removed from control plane through executing `karmadactl unjoin` again.

mrlihanbo:
> /cc @mrlihanbo



> delete the remain resource of member cluster that has been removed from control plane through executing `karmadactl unjoin` again.

The disagreement is should we allow user to delete the remain resource of member cluster that has been removed from control plane through executing karmadactl unjoin again?

xyz2277:
> > /cc @mrlihanbo
> 
> > delete the remain resource of member cluster that has been removed from control plane through executing `karmadactl unjoin` again.
> 
> The disagreement is should we allow user to delete the remain resource of member cluster that has been removed from control plane through executing karmadactl unjoin again?

I think so. @lonelyCZ

xyz2277:
Hey guys, what should I do now? @mrlihanbo @lonelyCZ

mrlihanbo:
> Hey guys, what should I do now? @mrlihanbo @lonelyCZ

Maybe we can add a flag to let users to determine if we should continue to delete RBAC resources.

xyz2277:
> > Hey guys, what should I do now? @mrlihanbo @lonelyCZ
> 
> Maybe we can add a flag to let users to determine if we should continue to delete RBAC resources.

xyz2277:
> > Hey guys, what should I do now? @mrlihanbo @lonelyCZ
> 
> Maybe we can add a flag to let users to determine if we should continue to delete RBAC resources.

Maybe, we should hide the details from users.

lonelyCZ:
> > > Hey guys, what should I do now? @mrlihanbo @lonelyCZ
> > 
> > 
> > Maybe we can add a flag to let users to determine if we should continue to delete RBAC resources.
> 
> Maybe, we should hide the details from users.

Generally, only when user provider `member cluster kubeconfig`, the remain RBAC resources can be deleted. 

I don't suggest that it is to use `karmadactl unjoin` again to delete the remain resources in the `Cluster` that has been moved from karmada. However, currently it can be,  I think it is only convenient but it's not recommended.

xyz2277:
> > > > Hey guys, what should I do now? @mrlihanbo @lonelyCZ
> > > 
> > > 
> > > Maybe we can add a flag to let users to determine if we should continue to delete RBAC resources.
> > 
> > 
> > Maybe, we should hide the details from users.
> 
> Generally, only when user provider `member cluster kubeconfig`, the remain RBAC resources can be deleted.
> 
> I don't suggest that it is to use `karmadactl unjoin` again to delete the remain resources in the `Cluster` that has been moved from karmada. However, currently it can be, I think it is only convenient but it's not recommended.

so, should I directly delete these error check apierrors.IsNotFound(err)？

pkg/karmadactl/unjoin.go
```diff
@@ -271,6 +271,7 @@ func deleteClusterObject(controlPlaneKarmadaClient *karmadaclientset.Clientset,
 
 	err := controlPlaneKarmadaClient.ClusterV1alpha1().Clusters().Delete(context.TODO(), opts.ClusterName, metav1.DeleteOptions{})
 	if apierrors.IsNotFound(err) {
+		klog.Errorf("No cluster object %s found in karmada control Plane", opts.ClusterName)
```
lonelyCZ, line 274:
```suggestion
		return fmt.Errorf("No cluster object %s found in karmada control Plane", opts.ClusterName)
```

xyz2277: got it~

lonelyCZ:
Thanks @xyz2277 

/lgtm

RainbowMango (approved):
/approve
Thanks.

PR #1884, opened 2022-05-25
Add cluster success threshold

dddddai:
**What type of PR is this?**
/kind feature

Cluster success threshold is the duration of successes for the cluster to be considered healthy after recovery.

Fixes #

We should be careful about the default value as it could slow down the cluster recovery, personally I'd prefer set the default value to 0 for keeping user experience as before, if users do need this threshold, they could set it.

```release-note
`karmada-controller-manager/karmada-agent`: Introduced `--cluster-success-threshold` flag to specify cluster success threshold, default to 30s.
```


pkg/controllers/status/cluster_condition_cache.go
```diff
@@ -46,9 +48,16 @@ func (c *clusterConditionStore) thresholdAdjustedReadyCondition(cluster *cluster
...
-		now.Before(saved.thresholdStartTime.Add(c.failureThreshold)) {
+
+	var threshold time.Duration
+	if isConditionTrue(observedReadyCondition) {
+		threshold = c.successThreshold
+	} else {
+		threshold = c.failureThreshold
+	}
+
+	if isConditionTrue(observedReadyCondition) != isConditionTrue(curReadyCondition) &&
```
RainbowMango, line 59:
```suggestion
	if observedReadyCondition != curReadyCondition &&
```
I can't see what's the difference between this, and the unit test seems ok with the change I'm suggesting.

pkg/controllers/status/cluster_status_controller.go
```diff
@@ -139,7 +142,7 @@ func (c *ClusterStatusController) syncClusterStatus(cluster *clusterv1alpha1.Clu
 	readyCondition := c.clusterConditionCache.thresholdAdjustedReadyCondition(cluster, &observedReadyCondition)
 
 	// cluster is offline after retry timeout, update cluster status immediately and return.
-	if !online && readyCondition.Status != metav1.ConditionTrue {
+	if !online && !isConditionTrue(readyCondition) {
```
RainbowMango, line 145:
```suggestion
	if !online && readyCondition.Status != metav1.ConditionFalse {
```
Does that make sense to you?
And, I don't think `isConditionTrue()` would help, can we remove it?

pkg/controllers/status/cluster_status_controller.go
```diff
@@ -149,7 +152,7 @@ func (c *ClusterStatusController) syncClusterStatus(cluster *clusterv1alpha1.Clu
 	}
 
 	// skip collecting cluster status if not ready
-	if online && healthy {
+	if online && healthy && isConditionTrue(readyCondition) {
```
RainbowMango, line 155: ditto.

RainbowMango:
@dddddai 
Now we cached the ready status in memory, if it is possible to remove the cache and turn to the [LastTransitionTime](https://github.com/kubernetes/kubernetes/blob/549e1b4fb6c524f60f2133bd28534ff10c9501d4/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L1490) ?

For a similar idea, I also post at https://github.com/karmada-io/karmada/pull/1945#discussion_r894273562.

dddddai:
Sounds great, let me have a try

pkg/controllers/status/cluster_condition_cache.go
dddddai: Because the condition can be `True/False/Unknown`, here we should only distinguish "true" and "not true" I think

For example, if the condition is changed from `False` to `Unknown`, we should update to `Unknown` immediately rather than retaining `False`

dddddai:
> @dddddai Now we cached the ready status in memory, if it is possible to remove the cache and turn to the [LastTransitionTime](https://github.com/kubernetes/kubernetes/blob/549e1b4fb6c524f60f2133bd28534ff10c9501d4/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L1490) ?
> 
> For a similar idea, I also post at [#1945 (comment)](https://github.com/karmada-io/karmada/pull/1945#discussion_r894273562).

Not really, we can't tell when the "real" cluster status changed through `LastTransitionTime`

pkg/controllers/status/cluster_condition_cache.go
RainbowMango: > For example, if the condition is changed from False to Unknown,

The condition should never change from `true`/`false` to `unknown`. Any other examples?

RainbowMango:
> Not really, we can't tell when the "observed" cluster status changed through LastTransitionTime

Give an example?

RainbowMango (approved):
/lgtm
/approve

PR #1902, opened 2022-05-28
add syncEvent from execution_controller to resource

### LINKED ISSUE #1898: SyncEvent from execution_controller is not be bind with workload
Now SyncEvent from execution_controller is not be bind with workload. For example, when you apply a nginx deployment to member clusters. You can see the sync event like below:
image
But it is not bind with actual workload:
image

Mainly because the workload in the work object is just a template, some values like resourceVersion and UID is missing. I wonder if there is a way to improve it.

**Why is this needed**:
Improve user experience.

hanweisen:
Signed-off-by: hanweisen <hanweisen_yewu@cmss.chinamobile.com>

**What type of PR is this?**
/kind feature

add syncEvent from execution_controller to resource.
For example if we describe deployment we can see event from execution-controller
```
#  kubectl --kubeconfig .kube/karmada.config describe deployment nginx
...
Events:
  Type     Reason                  Age                 From                  Message
  ----     ------                  ----                ----                  -------
  Warning  ApplyPolicyFailed       22s                 resource-detector     No policy match for resource
  Normal   ApplyPolicySucceed      22s (x3 over 22s)   resource-detector     Apply policy(default/nginx-propagation) succeed
  Normal   SyncWorkSucceed         22s (x9 over 22s)   binding-controller    Sync work of resourceBinding(default/nginx-deployment) successful.
  Normal   AggregateStatusSucceed  22s (x9 over 22s)   binding-controller    Update resourceBinding(default/nginx-deployment) with AggregatedStatus successfully.
  Normal   ScheduleBindingSucceed  22s (x11 over 22s)  karmada-scheduler     Binding has been scheduled
  Normal   SyncSucceed             22s                 execution-controller  Successfully applied resource(default/nginx) to cluster member1
  Normal   SyncSucceed             22s                 execution-controller  Successfully applied resource(default/nginx) to cluster member2
```
Fixes #1898

```release-note
NONE
```

### DIFF
--- a/pkg/apis/work/v1alpha2/well_known_labels.go
+++ b/pkg/apis/work/v1alpha2/well_known_labels.go
@@ -40,3 +40,13 @@ const (
 	// ResourceConflictResolutionOverwrite is the value of ResourceConflictResolutionAnnotation, indicating the overwrite strategy.
 	ResourceConflictResolutionOverwrite = "overwrite"
 )
+
+// Define annotations that are added to the resource template.
+const (
+	// ResourceTemplateUIDAnnotation is the annotation that is added to the manifest in the Work object.
+	// The annotation is used to identify the resource template which the manifest is derived from.
+	// The annotation can also be used to fire events when syncing Work to member clusters.
+	// For more details about UID, please refer to:
+	// https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#uids
+	ResourceTemplateUIDAnnotation = "resourcetemplate.karmada.io/uid"
+)
--- a/pkg/controllers/execution/execution_controller.go
+++ b/pkg/controllers/execution/execution_controller.go
@@ -196,7 +196,7 @@ func (c *Controller) syncToClusters(clusterName string, work *workv1alpha1.Work)
 			err = c.tryUpdateWorkload(clusterName, workload)
 			if err != nil {
 				klog.Errorf("Failed to update resource(%v/%v) in the given member cluster %s, err is %v", workload.GetNamespace(), workload.GetName(), clusterName, err)
-				c.EventRecorder.Eventf(workload, corev1.EventTypeWarning, workv1alpha1.EventReasonSyncWorkFailed, "Failed to update resource(%v/%v) in the given member cluster %s, err is %v", workload.GetNamespace(), workload.GetName(), clusterName, err)
+				c.eventf(workload, corev1.EventTypeWarning, workv1alpha1.EventReasonSyncWorkFailed, "Failed to update resource(%v/%v) in the given member cluster %s, err is %v", workload.GetNamespace(), workload.GetName(), clusterName, err)
 				errs = append(errs, err)
 				continue
 			}
@@ -204,12 +204,12 @@ func (c *Controller) syncToClusters(clusterName string, work *workv1alpha1.Work)
 			err = c.tryCreateWorkload(clusterName, workload)
 			if err != nil {
 				klog.Errorf("Failed to create resource(%v/%v) in the given member cluster %s, err is %v", workload.GetNamespace(), workload.GetName(), clusterName, err)
-				c.EventRecorder.Eventf(workload, corev1.EventTypeWarning, workv1alpha1.EventReasonSyncWorkFailed, "Failed to create resource(%v/%v) in the given member cluster %s, err is %v", workload.GetNamespace(), workload.GetName(), clusterName, err)
+				c.eventf(workload, corev1.EventTypeWarning, workv1alpha1.EventReasonSyncWorkFailed, "Failed to create resource(%v/%v) in the given member cluster %s, err is %v", workload.GetNamespace(), workload.GetName(), clusterName, err)
 				errs = append(errs, err)
 				continue
 			}
 		}
-		c.EventRecorder.Eventf(workload, corev1.EventTypeNormal, workv1alpha1.EventReasonSyncWorkSucceed, "Successfully applied resource(%v/%v) to cluster %s", workload.GetNamespace(), workload.GetName(), clusterName)
+		c.eventf(workload, corev1.EventTypeNormal, workv1alpha1.EventReasonSyncWorkSucceed, "Successfully applied resource(%v/%v) to cluster %s", workload.GetNamespace(), workload.GetName(), clusterName)
 		syncSucceedNum++
 	}
 
@@ -291,4 +291,13 @@ func (c *Controller) updateAppliedCondition(work *workv1alpha1.Work, status meta
 		}
 		return updateErr
 	})
+}
+
+func (c *Controller) eventf(object *unstructured.Unstructured, eventtype, reason, messageFmt string, args ...interface{}) {
+	ref, err := helper.GenEventRef(object)
+	if err != nil {
+		klog.Errorf("ignore event(%s) as failed to build event reference for: kind=%s, %s/%s due to %v", reason, object.GetKind(), object.GetNamespace(), object.GetName(), err)
+		return
+	}
+	c.EventRecorder.Eventf(ref, eventtype, reason, messageFmt, args...)
 }
--- a/pkg/util/helper/work.go
+++ b/pkg/util/helper/work.go
@@ -2,22 +2,28 @@ package helper
 
 import (
 	"context"
+	"fmt"
 
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	workv1alpha1 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha1"
+	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
+	"github.com/karmada-io/karmada/pkg/util"
 )
 
 // CreateOrUpdateWork creates a Work object if not exist, or updates if it already exist.
 func CreateOrUpdateWork(client client.Client, workMeta metav1.ObjectMeta, resource *unstructured.Unstructured) error {
 	workload := resource.DeepCopy()
+	util.MergeAnnotation(workload, workv1alpha2.ResourceTemplateUIDAnnotation, string(workload.GetUID()))
 	workloadJSON, err := workload.MarshalJSON()
 	if err != nil {
 		klog.Errorf("Failed to marshal workload(%s/%s), Error: %v", workload.GetNamespace(), workload.GetName(), err)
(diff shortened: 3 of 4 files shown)


hanweisen:
Currently only supported resource propagate with `PropagationPolicy`.`ClusterPropagationPolicy` is not supported.

pkg/controllers/execution/execution_controller.go
```diff
@@ -111,18 +116,46 @@ func (c *Controller) SetupWithManager(mgr controllerruntime.Manager) error {
 
 func (c *Controller) syncWork(clusterName string, work *workv1alpha1.Work) (controllerruntime.Result, error) {
 	err := c.syncToClusters(clusterName, work)
+	resource := c.tryGetResource(work)
```
RainbowMango, line 119: My concern is if it costs too much to add an event? We need two requests to get the resource.

RainbowMango (commented):
cc @Poor12 
Can you help with it?

hanweisen:
Sent an event to resource, only need information in `k8s.io/api/core/v1ObjectReference`, Maybe I can use annotation to store this information in the work object  when the work object is created

Poor12:
Compare the two events：involvedObject of events created by execution-controller is missing some paras. 
image
others:
image
We need complete object references. Guess add an annotation is a good way.

RainbowMango:
> We need complete object references. Guess add an annotation is a good way.

I doubt it. Because maintaining the annotation needs some effort too.

The effort depends on how much information we need. I can see `resourceVersion` and `uid` are missing from the description above.  I'm not sure if the `resourceVersion` is the mandatory information for an event. 
If yes, then we need to maintain the `resourceVersion` in the annotation, that might be the biggest concern.

RainbowMango:
So, let's clarify what informations we really need to fire an event. And we should start talking about the solution after that?

hanweisen:
Before request api server, The event sinker will filter events based on rate limits, and then aggregate events. Sinker will map event to key, The filter key is 
```
// k8s.io/client-go/tools/record/events_cache.go

func getSpamKey(event *v1.Event) string {
	return strings.Join([]string{
		event.Source.Component,
		event.Source.Host,
		event.InvolvedObject.Kind,
		event.InvolvedObject.Namespace,
		event.InvolvedObject.Name,
		string(event.InvolvedObject.UID),
		event.InvolvedObject.APIVersion,
	},
		"")
}
```

and the aggregate key is
```
func EventAggregatorByReasonFunc(event *v1.Event) (string, string) {
	return strings.Join([]string{
		event.Source.Component,
		event.Source.Host,
		event.InvolvedObject.Kind,
		event.InvolvedObject.Namespace,
		event.InvolvedObject.Name,
		string(event.InvolvedObject.UID),
		event.InvolvedObject.APIVersion,
		event.Type,
		event.Reason,
		event.ReportingController,
		event.ReportingInstance,
	},
		""), event.Message
}
```
I think we need at least `Kind`, `Name`, `Namespace`, `UID` and `APIVersion`

hanweisen:
I also tested that sending an event to the deployment, the minimum field that can be described by deployment is  `Kind`, `Name`, `Namespace`, `UID`.

hanweisen:
`kubectl describe` search event for resource defined in `describerMap` function in `k8s.io/kubectl/pkg/describe/describe.go`. Use `EventExpansion.GetFieldSelector` build field selector
```
// The EventExpansion interface allows manually adding extra methods to the EventInterface.
type EventExpansion interface {
        .....
	// Returns the appropriate field selector based on the API version being used to communicate with the server.
	// The returned field selector can be used with List and Watch to filter desired events.
	GetFieldSelector(involvedObjectName, involvedObjectNamespace, involvedObjectKind, involvedObjectUID *string) fields.Selector
}
```
So for the query side, I think only `Kind`, `Name`, `Namespace`, and `UID` is enough.

pkg/controllers/execution/execution_controller.go
```diff
@@ -185,6 +188,13 @@ func (c *Controller) syncToClusters(clusterName string, work *workv1alpha1.Work)
 	for _, manifest := range work.Spec.Workload.Manifests {
 		workload := &unstructured.Unstructured{}
 		err := workload.UnmarshalJSON(manifest.Raw)
+		reference, _ := ref.GetReference(c.Client.Scheme(), workload)
+		annotations := workload.GetAnnotations()
+		if annotations != nil && reference != nil {
+			if uid, uidExist := annotations[workv1alpha2.ObjectReferenceUIDLabel]; uidExist {
+				reference.UID = types.UID(uid)
```
hanweisen, line 195: Currently the event sent by the `execution_controller` already contains `Kind`, `Name`, `Namespace`, just add the `UID` and the event will be bound to the workload. We can refer to the `searchEvents` method of `k8s.io/kubectl/pkg/describe/describe.go`. I think `resourceVerstion` is not necessary for events. I can not find 
 any relevant usage in k8s
@RainbowMango @Poor12

hanweisen: It is possible that a third party event consumer may use `ResourceVersion`，we can add it later

Poor12: Thanks for confirming. Have you tested whether sync events are bound to workloads?

hanweisen: Here is the result
image

pkg/apis/work/v1alpha2/well_known_labels.go
```diff
@@ -25,6 +25,9 @@ const (
 
 	// WorkNameLabel is added to objects to specify associated Work's name.
 	WorkNameLabel = "work.karmada.io/name"
+
+	// ObjectReferenceUIDLabel is added to objects to specify object's uid
+	ObjectReferenceUIDLabel = "objectference.k8s.io/uid"
```
RainbowMango, line 30: It's not a label I guess.

RainbowMango, line 30:
```golang
// Define annotations that are added to the resource template.
const (
	// ResourceTemplateUIDAnnotation is the annotation that is added to the manifest in the Work object.
	// The annotation is used to identify the resource template which the manifest is derived from.
	// The annotation can also be used to fire events when syncing Work to member clusters.
	// For more details about UID, please refer to:
	// https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#uids
	ResourceTemplateUIDAnnotation = "resourcetemplate.karmada.io/uid"
)
```
How about this annotation?

pkg/controllers/execution/execution_controller.go
```diff
@@ -185,6 +188,13 @@ func (c *Controller) syncToClusters(clusterName string, work *workv1alpha1.Work)
 	for _, manifest := range work.Spec.Workload.Manifests {
 		workload := &unstructured.Unstructured{}
 		err := workload.UnmarshalJSON(manifest.Raw)
+		reference, _ := ref.GetReference(c.Client.Scheme(), workload)
```
RainbowMango, line 191: I suppose we can add a helper function at such as `pkg/util/helper/work.go`, or just to the controller, like:
```golang
// GenEventRefObj sets the UID(.metadata.uid) that might be missing for fire events.
// Do nothing if the UID already exist, otherwise set the UID from annotation.
func GenEventRefObj(resource *unstructured.Unstructured) *unstructured.Unstructured {
	if len(resource.GetUID()) > 0 {
		return resource
	}

	// set UID according to ResourceTemplateUIDAnnotation.
	// TODO

	return resource
}
```
or 
```golang
// GenEventRef returns the event reference.
func GenEventRef(resource *unstructured.Unstructured) *corev1.ObjectReference {

}
```

It will be good to re-use this function if we need to fire an event in another place.
And we can add a unit test for it too.

RainbowMango (commented):
I like this approach, thanks @hanweisen take this.

pkg/apis/work/v1alpha2/well_known_labels.go
hanweisen: Yes. It is not good.
`ClusterResourceBindingLabel`, `ResourceBindingNamespaceLabel` and `ResourceBindingNameLabel` is annotation in this file

hanweisen: That's great. I'm not very good at naming

RainbowMango: Thanks for reminding me.  Yeah, I'll optimize it by a separate PR.

RainbowMango: No worries, we need to know that the annotation/label is a kind of API, that's we have to take care of very carefully.

RainbowMango: opened #1979 for this.

pkg/controllers/execution/execution_controller.go
hanweisen: I prefer `GenEventRef `, `recorder` converts the object to an ObjectReference and no longer uses the object later. `GenEventRef` is enough and Avoid deep copy of the object。

pkg/controllers/execution/execution_controller.go
```diff
@@ -185,6 +185,7 @@ func (c *Controller) syncToClusters(clusterName string, work *workv1alpha1.Work)
 	for _, manifest := range work.Spec.Workload.Manifests {
 		workload := &unstructured.Unstructured{}
 		err := workload.UnmarshalJSON(manifest.Raw)
+		reference, _ := helper.GenEventRef(c.Client.Scheme(), workload)
```
RainbowMango, line 188: Probably we shouldn't ignore the error here. 

Maybe we can introduce a method to the `execution controller`, like:
```golang
func (c *Controller) Eventf(object *unstructured.Unstructured, eventtype, reason, messageFmt string, args ...interface{}) {
	ref, err := helper.GenEventRef(...)
	if err != nil {
		fmt.Errorf("ignored event(%s) as failed to build event reference: %s", reason, err)
	}
	
	c.EventRecorder.Eventf(ref, eventtype, reason, messageFmt, args...)
}
```

pkg/util/helper/work.go
```diff
@@ -76,3 +86,19 @@ func GetWorksByLabelsSet(c client.Client, ls labels.Set) (*workv1alpha1.WorkList
 
 	return workList, c.List(context.TODO(), workList, listOpt)
 }
+
+// GenEventRef returns the event reference. sets the UID(.spec.uid) that might be missing for fire events.
+// Do nothing if the UID already exist, otherwise set the UID from annotation.
+func GenEventRef(schema *runtime.Scheme, resource *unstructured.Unstructured) (*corev1.ObjectReference, error) {
+	reference, err := ref.GetReference(schema, resource)
```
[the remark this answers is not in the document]
hanweisen: Cluster-level resources don't have `namespace`, I can check `namespace`, `name`, and `kind`.

pkg/controllers/execution/execution_controller.go
hanweisen: Okay

hanweisen:
All solved, thanks for reviewing

pkg/controllers/execution/execution_controller.go
```diff
@@ -292,3 +292,12 @@ func (c *Controller) updateAppliedCondition(work *workv1alpha1.Work, status meta
 		return updateErr
 	})
 }
+
+func (c *Controller) eventf(object *unstructured.Unstructured, eventtype, reason, messageFmt string, args ...interface{}) {
+	ref, err := helper.GenEventRef(object)
+	if err != nil {
+		klog.Errorf("ignored event(%s) as failed to build event reference to: '%#v' due to %v", reason, object, err)
```
RainbowMango, line 299:
```suggestion
		klog.Errorf("ignore event(%s) as failed to build event reference for: kind=%s, %s/%s due to %v", reason, object.GetKind(), object.GetNamespace(), object.GetName(), err)
```

log the whole object might be verbose.

RainbowMango (commented):
Generally looks good.
Just one nit.

RainbowMango:
/lgtm
/approve

Thanks @hanweisen for the quick response.

PR #1911, opened 2022-05-30
fix implicit issue for event filter

Garrybest:

**What type of PR is this?**
/kind bug

Some controller uses `SetupWithManager` with event handlers. However, this handler takes effect in global, i.e., other watched kinds will also be filtered by this global handler. This is not a correct usage because in most cases we only want to filter objects in `For` function (Please see [here](https://github.com/karmada-io/karmada/blob/master/vendor/sigs.k8s.io/controller-runtime/pkg/builder/controller.go#L136-L144)). So these predicated functions should be moved into `ForOption`.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/execution/execution_controller.go
+++ b/pkg/controllers/execution/execution_controller.go
@@ -14,6 +14,7 @@ import (
 	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
 	controllerruntime "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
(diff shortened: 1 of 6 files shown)


XiShanYongYe-Chang (commented):
Thanks~

/lgtm

/cc @mrlihanbo @RainbowMango

pkg/controllers/execution/execution_controller.go
```diff
@@ -100,9 +101,8 @@ func (c *Controller) Reconcile(ctx context.Context, req controllerruntime.Reques
 // SetupWithManager creates a controller and register to controller manager.
 func (c *Controller) SetupWithManager(mgr controllerruntime.Manager) error {
 	return controllerruntime.NewControllerManagedBy(mgr).
-		For(&workv1alpha1.Work{}).
+		For(&workv1alpha1.Work{}, builder.WithPredicates(c.PredicateFunc)).
 		WithEventFilter(predicate.GenerationChangedPredicate{}).
```
RainbowMango, line 105: Should we do the same thing for  `predicate.GenerationChangedPredicate{}`?

RainbowMango (commented):
Thanks @Garrybest for pointing it out!

pkg/controllers/execution/execution_controller.go
Garrybest: `GenerationChangedPredicate` could be a global predicate function. It can take effect when any `watched` resource changed. Well, it does not matter here cause we don't have any other resources to watch.

RainbowMango: Yes, no big deal.

PR #1919, opened 2022-05-30
adjust etcd port

my-git9:
/kind bug

The etcd port seems  wrong.


mrlihanbo:
Nice finding.
/lgtm

RainbowMango:
Why we didn't catch it during our testing?

my-git9:
> Why we didn't catch it during our testing?

containerPort is just a statement, it will not affect the operation of the application, maybe for this reason

RainbowMango:
Thanks, @AllenZMC for the explanation. 

/approve

PR #1933, opened 2022-05-31
modify output for deinit

my-git9:
/kind cleanup

When deinit for karmada:`kubectl karmada deinit`, the output for yes or no is:

image

Mybe it shoud be:
image


my-git9:
> may be this is not a problem. This is my kubebuilder output image

 Maybe it would be more comfortable for them to be on the same line？

RainbowMango:
I tend to this approach. (It's not about good or bad, just a feeling)

/approve
/lgtm

Issue #1762, opened 2022-05-10
TODO List: Enhance Cluster Lifecycle Management


Garrybest:
**What would you like to be added**:
According to https://github.com/karmada-io/karmada/issues/1640#issuecomment-1119377312, I'd like to enhance cluster lifecycle management. Here is the todo-list.

- [x] Add a taint for the clusters which are unhealthy for a period of time.#1781
- [x] Add taint manager. #1945
- [x] Remove cluster scheduling result in `spec.clusters` when the cluster is terminating. #2373
- [ ] Remove `Failover` from scheduler. #2391

**Why is this needed**:
Move some functions away from scheduler and make the scheduler more concise.

RainbowMango:
Hi @Garrybest 
I guess the task(`Remove cluster scheduling result in spec.clusters when the cluster is terminating.`) is included in #1945.

Should we update the task list?

Garrybest:
> Hi @Garrybest I guess the task(`Remove cluster scheduling result in spec.clusters when the cluster is terminating.`) is included in #1945.
> 
> Should we update the task list?

Well, I'm afraid not. This task is to drain a cluster by using NoExecute taint when we decide to unjoin it, while #1945 focuses on how taint manager works.

Garrybest:
I have an idea. Maybe we could implement this in cmd `karmadactl unjoin`. Just like what `kubectl drain` does, when we want to unjoin a cluster, we cordon this cluster first and then remove RB's cluster scheduling result in spec.clusters.

RainbowMango:
Great thanks to @Garrybest !!!

PR #1945, opened 2022-06-01
add NoExecute taint manager

Garrybest:

**What type of PR is this?**
/kind feature

Add taint manager for NoExecute manager.

Fixes part of #1762

```release-note
`karmada-controller-manager`: Introduced `taint manager` to cluster controller to evict workloads from clusters with `no-execute` taint. 
```

### DIFF
--- a/cmd/controller-manager/app/controllermanager.go
+++ b/cmd/controller-manager/app/controllermanager.go
@@ -5,6 +5,7 @@ import (
 	"flag"
 	"net"
 	"strconv"
+	"time"
 
 	"github.com/spf13/cobra"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -178,6 +179,11 @@ func init() {
 func startClusterController(ctx controllerscontext.Context) (enabled bool, err error) {
 	mgr := ctx.Mgr
 	opts := ctx.Opts
+
+	if err := cluster.IndexField(mgr); err != nil {
+		return false, err
+	}
+
 	clusterController := &cluster.Controller{
 		Client:                    mgr.GetClient(),
 		EventRecorder:             mgr.GetEventRecorderFor(cluster.ControllerName),
@@ -189,6 +195,17 @@ func startClusterController(ctx controllerscontext.Context) (enabled bool, err e
 	if err := clusterController.SetupWithManager(mgr); err != nil {
 		return false, err
 	}
+
+	taintManager := &cluster.NoExecuteTaintManager{
+		Client:                             mgr.GetClient(),
+		EventRecorder:                      mgr.GetEventRecorderFor(cluster.TaintManagerName),
+		ClusterTaintEvictionRetryFrequency: 10 * time.Second,
+		ConcurrentReconciles:               3,
+	}
+	if err := taintManager.SetupWithManager(mgr); err != nil {
+		return false, err
+	}
+
 	return true, nil
 }
 
--- a/cmd/controller-manager/app/controllermanager.go
+++ b/cmd/controller-manager/app/controllermanager.go
@@ -180,10 +180,6 @@ func startClusterController(ctx controllerscontext.Context) (enabled bool, err e
 	mgr := ctx.Mgr
 	opts := ctx.Opts
 
-	if err := cluster.IndexField(mgr); err != nil {
-		return false, err
-	}
-
 	clusterController := &cluster.Controller{
 		Client:                    mgr.GetClient(),
 		EventRecorder:             mgr.GetEventRecorderFor(cluster.ControllerName),
@@ -196,14 +192,20 @@ func startClusterController(ctx controllerscontext.Context) (enabled bool, err e
 		return false, err
 	}
 
-	taintManager := &cluster.NoExecuteTaintManager{
-		Client:                             mgr.GetClient(),
-		EventRecorder:                      mgr.GetEventRecorderFor(cluster.TaintManagerName),
-		ClusterTaintEvictionRetryFrequency: 10 * time.Second,
-		ConcurrentReconciles:               3,
-	}
-	if err := taintManager.SetupWithManager(mgr); err != nil {
-		return false, err
+	if ctx.Opts.EnableTaintManager {
+		if err := cluster.IndexField(mgr); err != nil {
+			return false, err
+		}
+
+		taintManager := &cluster.NoExecuteTaintManager{
+			Client:                             mgr.GetClient(),
+			EventRecorder:                      mgr.GetEventRecorderFor(cluster.TaintManagerName),
+			ClusterTaintEvictionRetryFrequency: 10 * time.Second,
+			ConcurrentReconciles:               3,
+		}
+		if err := taintManager.SetupWithManager(mgr); err != nil {
+			return false, err
+		}
 	}
 
 	return true, nil
@@ -524,6 +526,7 @@ func setupControllers(mgr controllerruntime.Manager, opts *options.Options, stop
 			ClusterAPIBurst:                   opts.ClusterAPIBurst,
 			SkippedPropagatingNamespaces:      opts.SkippedPropagatingNamespaces,
 			ConcurrentWorkSyncs:               opts.ConcurrentWorkSyncs,
+			EnableTaintManager:                opts.EnableTaintManager,
 			RateLimiterOptions:                opts.RateLimiterOpts,
 		},
 		StopChan:                    stopChan,
--- a/pkg/controllers/cluster/taint_manager.go
+++ b/pkg/controllers/cluster/taint_manager.go
@@ -152,15 +152,13 @@ func (tc *NoExecuteTaintManager) syncBindingEviction(key util.QueueKey) error {
 	// Case 2: Need eviction after toleration time. If time is up, do eviction right now.
 	// Case 3: Tolerate forever, we do nothing.
 	if needEviction || tolerationTime == 0 {
-		if result := util.RemoveTargetCluster(binding.Spec.Clusters, cluster); len(result) < len(binding.Spec.Clusters) {
-			// update final result to evict the target cluster
-			binding.Spec.Clusters = result
-			if err = tc.Update(context.TODO(), binding); err != nil {
-				klog.ErrorS(err, "Failed to update binding", "binding", klog.KObj(binding))
-				return err
-			}
-			tc.emitClusterEvictionEventForResourceBinding(binding, cluster)
+		// update final result to evict the target cluster
+		binding.Spec.Clusters = util.RemoveTargetCluster(binding.Spec.Clusters, cluster)
+		if err = tc.Update(context.TODO(), binding); err != nil {
+			klog.ErrorS(err, "Failed to update binding", "binding", klog.KObj(binding))
+			return err
 		}
+		tc.emitClusterEvictionEventForResourceBinding(binding, cluster)
 	} else if tolerationTime > 0 {
 		tc.bindingEvictionWorker.AddAfter(fedKey, tolerationTime)
 	}
@@ -199,15 +197,13 @@ func (tc *NoExecuteTaintManager) syncClusterBindingEviction(key util.QueueKey) e
 	// Case 2: Need eviction after toleration time. If time is up, do eviction right now.
 	// Case 3: Tolerate forever, we do nothing.
 	if needEviction || tolerationTime == 0 {
-		if result := util.RemoveTargetCluster(binding.Spec.Clusters, cluster); len(result) < len(binding.Spec.Clusters) {
-			// update final result to evict the target cluster
-			binding.Spec.Clusters = result
-			if err = tc.Update(context.TODO(), binding); err != nil {
-				klog.ErrorS(err, "Failed to update cluster binding", "binding", binding.Name)
-				return err
-			}
-			tc.emitClusterEvictionEventForClusterResourceBinding(binding, cluster)
+		// update final result to evict the target cluster
+		binding.Spec.Clusters = util.RemoveTargetCluster(binding.Spec.Clusters, cluster)
+		if err = tc.Update(context.TODO(), binding); err != nil {
+			klog.ErrorS(err, "Failed to update cluster binding", "binding", binding.Name)
+			return err
 		}
+		tc.emitClusterEvictionEventForClusterResourceBinding(binding, cluster)
 	} else if tolerationTime > 0 {
 		tc.clusterBindingEvictionWorker.AddAfter(fedKey, tolerationTime)
 		return nil
--- a/pkg/controllers/cluster/taint_manager.go
+++ b/pkg/controllers/cluster/taint_manager.go
@@ -1,0 +1,289 @@
+package cluster
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/types"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/client-go/tools/record"
+	"k8s.io/klog/v2"
+	controllerruntime "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	clusterv1alpha1 "github.com/karmada-io/karmada/pkg/apis/cluster/v1alpha1"
+	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
+	"github.com/karmada-io/karmada/pkg/util"
+	"github.com/karmada-io/karmada/pkg/util/helper"
+	"github.com/karmada-io/karmada/pkg/util/informermanager/keys"
+)
+
+// TaintManagerName is the controller name that will be used for taint management.
+const TaintManagerName = "taint-manager"
+
+// NoExecuteTaintManager listens to Taint/Toleration changes and is responsible for removing objects
+// from Clusters tainted with NoExecute Taints.
+type NoExecuteTaintManager struct {
+	client.Client // used to operate Cluster resources.
+	EventRecorder record.EventRecorder
+
+	ClusterTaintEvictionRetryFrequency time.Duration
+	ConcurrentReconciles               int
+
+	bindingEvictionWorker        util.AsyncWorker
+	clusterBindingEvictionWorker util.AsyncWorker
+}
+
+// Reconcile performs a full reconciliation for the object referred to by the Request.
+// The Controller will requeue the Request to be processed again if an error is non-nil or
+// Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
+func (tc *NoExecuteTaintManager) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	klog.V(4).Infof("Reconciling cluster %s for taint manager", req.NamespacedName.Name)
+
+	cluster := &clusterv1alpha1.Cluster{}
+	if err := tc.Client.Get(context.TODO(), req.NamespacedName, cluster); err != nil {
+		// The resource may no longer exist, in which case we stop processing.
+		if apierrors.IsNotFound(err) {
+			return controllerruntime.Result{}, nil
+		}
+		return controllerruntime.Result{Requeue: true}, err
+	}
+
+	// Check whether the target cluster has no execute taints.
+	if !helper.HasNoExecuteTaints(cluster.Spec.Taints) {
+		return controllerruntime.Result{}, nil
+	}
+
+	return tc.syncCluster(ctx, cluster)
+}
+
+func (tc *NoExecuteTaintManager) syncCluster(ctx context.Context, cluster *clusterv1alpha1.Cluster) (reconcile.Result, error) {
+	// List all ResourceBindings which are assigned to this cluster.
+	rbList := &workv1alpha2.ResourceBindingList{}
+	if err := tc.List(ctx, rbList, client.MatchingFieldsSelector{
+		Selector: fields.OneTermEqualSelector(rbClusterKeyIndex, cluster.Name),
+	}); err != nil {
+		klog.ErrorS(err, "Failed to list ResourceBindings", "cluster", cluster.Name)
+		return controllerruntime.Result{Requeue: true}, err
+	}
+	for i := range rbList.Items {
+		key, err := keys.FederatedKeyFunc(cluster.Name, &rbList.Items[i])
+		if err != nil {
+			klog.Warningf("Failed to generate key for obj: %s", rbList.Items[i].GetObjectKind().GroupVersionKind())
+			continue
+		}
+		tc.bindingEvictionWorker.Add(key)
+	}
+
+	// List all ClusterResourceBindings which are assigned to this cluster.
+	crbList := &workv1alpha2.ClusterResourceBindingList{}
+	if err := tc.List(ctx, crbList, client.MatchingFieldsSelector{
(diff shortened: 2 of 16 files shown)


RainbowMango:
/assign
But I'm afraid don't have enough time to look at it.

pkg/controllers/cluster/taint_manager.go
```diff
@@ -0,0 +1,347 @@
...
+		} else {
+			tc.evictionStartTimeMap.set(keyStr, now)
+			tc.bindingEvictionWorker.AddAfter(fedKey, tolerationTime)
+		}
+	}
+
+	if evictNow {
+		if result := util.RemoveTargetCluster(binding.Spec.Clusters, cluster); len(result) < len(binding.Spec.Clusters) {
+			// update final result to evict the target cluster
+			binding.Spec.Clusters = result
```
dddddai, line 205: I'm not sure if we should directly modify the obj in the cache

pkg/controllers/cluster/taint_manager.go
```diff
@@ -0,0 +1,347 @@
...
+		return fmt.Errorf("invalid key")
+	}
+	cluster := fedKey.Cluster
+	keyStr := fedKey.String()
+
+	binding := &workv1alpha2.ResourceBinding{}
+	if err := tc.Client.Get(context.TODO(), types.NamespacedName{Namespace: fedKey.Namespace, Name: fedKey.Name}, binding); err != nil {
+		// The resource no longer exist, in which case we stop processing.
+		if apierrors.IsNotFound(err) {
+			return nil
```
dddddai, line 169: Should call `tc.evictionStartTimeMap.delete(keyStr)` here?

pkg/controllers/cluster/taint_manager.go
```diff
@@ -0,0 +1,347 @@
...
+	if needEviction || tolerationTime == 0 {
+		evictNow = true
+	} else if tolerationTime > 0 {
+		now := time.Now()
+		if startTime, exist := tc.evictionStartTimeMap.get(keyStr); exist {
+			if startTime.Add(tolerationTime).Before(now) {
+				evictNow = true
+			}
+		} else {
+			tc.evictionStartTimeMap.set(keyStr, now)
```
dddddai, line 197: I think we should check if the key exists, if not then set the start time

pkg/controllers/cluster/taint_manager.go
```diff
@@ -0,0 +1,347 @@
...
+			binding.Spec.Clusters = result
+			if err = tc.Update(context.TODO(), binding); err != nil {
+				klog.ErrorS(err, "Failed to update binding", "binding", klog.KObj(binding))
+				return err
+			}
+			tc.emitClusterEvictionEventForResourceBinding(binding, cluster)
+			tc.evictionStartTimeMap.delete(keyStr)
+		}
+	}
+	return nil
```
dddddai, line 214: Should call `tc.evictionStartTimeMap.delete(keyStr)` if don't need to evict? (the cluster recovered and the taint is removed)

pkg/controllers/cluster/taint_manager.go
```diff
@@ -0,0 +1,347 @@
...
+	if err := tc.Client.Get(context.TODO(), types.NamespacedName{Namespace: fedKey.Namespace, Name: fedKey.Name}, binding); err != nil {
+		// The resource no longer exist, in which case we stop processing.
+		if apierrors.IsNotFound(err) {
+			return nil
+		}
+		return fmt.Errorf("failed to get binding %s: %v", fedKey.NamespaceKey(), err)
+	}
+
+	if !binding.DeletionTimestamp.IsZero() || !util.IsTargetClusterExist(binding.Spec.Clusters, cluster) {
+		return nil
```
dddddai, line 175: Should call `tc.evictionStartTimeMap.delete(keyStr)` here?

dddddai (commented):
Cool!

pkg/controllers/cluster/taint_manager.go
Garrybest: I think it's OK. First, when we get object from informer cache, controller-runtime will deep copy this object first and inject this object into our object, see [here](https://github.com/karmada-io/karmada/blob/master/vendor/sigs.k8s.io/controller-runtime/pkg/cache/internal/cache_reader.go#L84-L91). And the option `UnsafeDisableDeepCopyByObject` is default to be empty, so every object has been deep copied internally, see [here](https://github.com/karmada-io/karmada/blob/master/vendor/sigs.k8s.io/controller-runtime/pkg/cache/cache.go#L126-L131).

pkg/controllers/cluster/taint_manager.go
Garrybest: Indeed, the check is done first. 
```go
		if startTime, exist := tc.evictionStartTimeMap.get(keyStr); exist {
			...
		} else {
			tc.evictionStartTimeMap.set(keyStr, now)
			...
		}
```

dddddai: Oh yes I missed it

pkg/controllers/cluster/taint_manager.go
Garrybest: That's right, I have made some changes. If trigger time has not been reached, return first.

pkg/controllers/cluster/taint_manager.go
```diff
@@ -0,0 +1,360 @@
...
+	if !binding.DeletionTimestamp.IsZero() || !util.IsTargetClusterExist(binding.Spec.Clusters, cluster) {
+		tc.evictionStartTimeMap.delete(keyStr)
+		return nil
+	}
+
+	needEviction, tolerationTime, err := tc.needEviction(cluster, binding.Annotations)
+	if err != nil {
+		klog.ErrorS(err, "Failed to check if cluster binding needs eviction", "binding", binding.Name)
+		return err
+	}
```
RainbowMango, line 252: I guess we can add a short path in case of `needEviction==false`, like:

```golang
	if !needEviction {
		tc.evictionStartTimeMap.delete(keyStr)
		return nil
	}
```

pkg/controllers/cluster/taint_manager.go
```diff
@@ -0,0 +1,360 @@
...
+		} else {
+			// If first, we set the start time and add this object after tolerationTime.
+			tc.evictionStartTimeMap.set(keyStr, now)
+			tc.bindingEvictionWorker.AddAfter(fedKey, tolerationTime)
+			return nil
+		}
+	}
+
+	if evictNow {
+		if result := util.RemoveTargetCluster(binding.Spec.Clusters, cluster); len(result) < len(binding.Spec.Clusters) {
```
RainbowMango, line 210: Note: This is the thing I want to optimize. 
I hope the `evicted` cluster could be exist in somewhere until the scheduler finds a new cluster for the resource referencing by rb/crb.

I'll start working on it.

RainbowMango (commented):
It's really enjoyable to look at code like this.

Thanks @Garrybest and sorry for let this sit.

pkg/controllers/cluster/taint_manager.go
```diff
@@ -0,0 +1,360 @@
...
+	}
+	for i := range crbList.Items {
+		key, err := keys.FederatedKeyFunc(cluster.Name, &crbList.Items[i])
+		if err != nil {
+			klog.Warningf("Failed to generate key for obj: %s", crbList.Items[i].GetObjectKind().GroupVersionKind())
+			continue
+		}
+		tc.clusterBindingEvictionWorker.Add(key)
+	}
+	return controllerruntime.Result{RequeueAfter: tc.ClusterTaintEvictionRetryFrequency}, nil
```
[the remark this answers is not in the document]
Garrybest: Well, the logic of taint manager is to list cluster which has NoExecute taint periodically and add all respective RBs and CRBs into eviction worker. So if something happens to a RB, e.g. it has an NoExecute toleration before but the toleration changes so that it should be evicted, we could not receive an event. Thus, we check the cluster periodically, it's a litlle bit like what `cluster_status_controller` does.

pkg/controllers/cluster/taint_manager.go
```diff
@@ -0,0 +1,360 @@
...
+}
+
+// Start starts an asynchronous loop that handle evictions.
+func (tc *NoExecuteTaintManager) Start(ctx context.Context) error {
+	bindingEvictionWorkerOptions := util.Options{
+		Name:          "binding-eviction",
+		KeyFunc:       nil,
```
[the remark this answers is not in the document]
Garrybest: Do you mean why using a `nil KeyFunc` but generated `FederatedKey` in `syncCluster`? 

The item in the eviction worker should inform the worker which cluster it should be evicted from. However, `FederatedKeyFunc: func FederatedKeyFunc(cluster string, obj interface{}) (FederatedKey, error)` is not a kind of `KeyFunc: func(obj interface{}) (QueueKey, error)`. So we can't use `FederatedKeyFunc` as a `KeyFunc`.

I refer to [this part of the code](https://github.com/karmada-io/karmada/blob/main/pkg/controllers/mcs/service_export_controller.go#L132-L141) in `service_export_controller.go`.

pkg/controllers/cluster/taint_manager.go
Garrybest: I'm afraid not. If `needEviction==false`, maybe the toleration time is not zero, which means  it should be evicted after some period.

pkg/controllers/cluster/taint_manager.go
```diff
@@ -0,0 +1,360 @@
...
+	// Case 2: need eviction after toleration time. If time is up, do eviction right now.
+	// Case 3: Tolerate forever, we do nothing.
+	evictNow := false
+	if needEviction || tolerationTime == 0 {
+		evictNow = true
+	} else if tolerationTime > 0 {
+		now := time.Now()
+		if startTime, exist := tc.evictionStartTimeMap.get(keyStr); exist {
+			// If trigger time has not been reached, just return.
+			if startTime.Add(tolerationTime).After(now) {
```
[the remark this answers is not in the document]
Garrybest: Thanks for reminding! I didn't realize this field. I think we don't need a eviction map now. Let me make a change.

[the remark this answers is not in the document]
Garrybest: I guess we could automatically add this field in cluster mutating.

pkg/controllers/cluster/taint_manager.go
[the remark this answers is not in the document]
RainbowMango: got it. thanks.

pkg/controllers/cluster/taint_manager.go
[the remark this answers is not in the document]
RainbowMango: Yes, exactly. Got it.

But, the mismatched type is not the key problem, no `cluster` attribute in `RB/CRB`, so we can't use `FederatedKeyFunc` here. right?

pkg/controllers/cluster/taint_manager.go
```diff
@@ -0,0 +1,360 @@
...
+	}
+
+	if !binding.DeletionTimestamp.IsZero() || !util.IsTargetClusterExist(binding.Spec.Clusters, cluster) {
+		tc.evictionStartTimeMap.delete(keyStr)
+		return nil
+	}
+
+	needEviction, tolerationTime, err := tc.needEviction(cluster, binding.Annotations)
+	if err != nil {
+		klog.ErrorS(err, "Failed to check if binding needs eviction", "binding", klog.KObj(binding))
```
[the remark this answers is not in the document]
RainbowMango: If I remember right, this is first time we use the `structured` logs, are you using it in your production environment?

Do you think it helpful enough for use to migrate the legacy logs?

pkg/controllers/cluster/taint_manager.go
RainbowMango: right.

pkg/controllers/cluster/taint_manager.go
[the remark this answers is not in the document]
Garrybest: That's right, we must know the RB/CRB along with need-to-evicted cluster.

pkg/controllers/cluster/taint_manager.go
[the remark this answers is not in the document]
Garrybest: Yes, I prefer to use `structured` logs. We could see many similar samples in `k/k`. As a result, the log could be more readable.

Actually, this is not the first in karmada. Structured logs are used in scheduler and descheduler.

pkg/controllers/cluster/taint_manager.go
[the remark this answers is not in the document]
Garrybest: I have made a change, now the added time will be recorded in this field. We don't need to handle the eviction start time, which really annoys me before🤣.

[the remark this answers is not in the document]
RainbowMango: > I have made a change, now the added time will be recorded in this field.

Where?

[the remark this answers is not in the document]
Garrybest: pkg/registry/cluster/strategy.go

Garrybest:
The e2e error does not make any sense. Maybe we need a retest.

RainbowMango:
Right.

Echo here for @XiShanYongYe-Chang to check.
https://github.com/karmada-io/karmada/runs/6858581736?check_suite_focus=true
```
[namespace auto-provision] namespace auto-provision testing joining new cluster
namespace should be propagated to new created clusters[needCreateCluster]
/home/runner/work/karmada/karmada/test/e2e/namespace_test.go:131
------------------------------
• [FAILED] [105.817 seconds]
[namespace auto-provision] namespace auto-provision testing
/home/runner/work/karmada/karmada/test/e2e/namespace_test.go:24
joining new cluster [BeforeEach][needCreateCluster]
/home/runner/work/karmada/karmada/test/e2e/namespace_test.go:84
    namespace should be propagated to new created clusters
/home/runner/work/karmada/karmada/test/e2e/namespace_test.go:131
Begin Captured StdOut/StdErr Output >>
    E0613 09:56:15.873001   42505 unjoin.go:277] Failed to delete cluster object. cluster name: member-e2e-blx, error: Delete "https://172.18.0.5:5443/apis/cluster.karmada.io/v1alpha1/clusters/member-e2e-blx": dial tcp 172.18.0.5:5443: connect: connection refused
    E0613 09:56:15.873109   42505 unjoin.go:166] Failed to delete cluster object. cluster name: member-e2e-blx, error: Delete "https://172.18.0.5:5443/apis/cluster.karmada.io/v1alpha1/clusters/member-e2e-blx": dial tcp 172.18.0.5:5443: connect: connection refused
```

XiShanYongYe-Chang:
Logs are lost. Access to `karmada-apiserver` is denied based on existing error information:

```
Delete "https://172.18.0.5:5443/apis/cluster.karmada.io/v1alpha1/clusters/member-e2e-blx": dial tcp 172.18.0.5:5443: connect: connection refused
Get \"[https://172.18.0.5:5443/api/v1/namespaces/karmada-cluster\](https://172.18.0.5:5443/api/v1/namespaces/karmada-cluster/)": dial tcp 172.18.0.5:5443: connect: connection refused
Delete "https://172.18.0.5:5443/api/v1/namespaces/karmadatest-zzk": dial tcp 172.18.0.5:5443: connect: connection refused
```

Garrybest:
No big deal, maybe just some occasional errors.

dddddai:
Looks pretty good!

pkg/controllers/cluster/taint_manager.go
RainbowMango: cc @lonelyCZ who might be interested in this feature I'm talking about.
(Sorry, for the delay, I'm afraid I don't have enough time for it this week, I'll try to get back on this next week.)

Garrybest: OK, Thanks for your hard work.

Garrybest:
Could we move forward?

RainbowMango:
I just talked to @XiShanYongYe-Chang about this, he will work on the issue I mentioned above(https://github.com/karmada-io/karmada/pull/1945#discussion_r894278454).

I don't want the `replica/cluster` to be removed directly from `binding.spec.clusters` that will lead to the replicas being removed immediately. 

So, can we just hold for a while? I guess we can include this in the next release(v1.3).

Garrybest:
OK.

cmd/controller-manager/app/controllermanager.go
```diff
@@ -181,6 +187,17 @@ func startClusterController(ctx controllerscontext.Context) (enabled bool, err e
...
 		return false, err
 	}
+
+	taintManager := &cluster.NoExecuteTaintManager{
+		Client:                             mgr.GetClient(),
+		EventRecorder:                      mgr.GetEventRecorderFor(cluster.TaintManagerName),
+		ClusterTaintEvictionRetryFrequency: 10 * time.Second,
+		ConcurrentReconciles:               3,
+	}
+	if err := taintManager.SetupWithManager(mgr); err != nil {
```
RainbowMango, line 197: The NoExecuteTaintManager is essentially a controller, do you think we should reserve the capacity of disabling it?

Garrybest: I think NoExecuteTaintManager is a part of cluster controller. 

Kubernetes takes `TaintManager` as a part of NodeLifecycleController. It provides an option `--enable-taint-manager` in `NodeLifecycleController` to enable taint manager.

So what about adding an option `--enable-taint-manager` as well?

pkg/controllers/cluster/taint_manager.go
RainbowMango: Another question, I guess the `len(result) < len(binding.Spec.Clusters)` want to confirm if the cluster is still in the `target`, if yes, can we use something like:
```golang
		if binding.Spec.TargetContains(cluster) {
			binding.Spec.Clusters = util.RemoveTargetCluster(binding.Spec.Clusters, cluster)
			...
		}
```
would it looks better?

cmd/controller-manager/app/controllermanager.go
RainbowMango: I agree. the --enable-taint-manager defaults to `true`. But, I think the feature is part of `Failover`, I suggest checking both the [feature gate](https://github.com/karmada-io/karmada/blob/f46d90f6b4ab04e5263aa142c3b6ae6078b645da/pkg/features/features.go#L10) and the flags. 
By the way, the feature gate will be removed eventually.

RainbowMango (commented):
Generally looks good to me.

Even though I hope [this issue](https://github.com/karmada-io/karmada/pull/1945#discussion_r894278454) could be solved before this PR, but at the meantime, I don't want this to be delied so much, so I suggest moving this forward now since we have a lot of things that addressed by #1945 to do.

Does that makes sense to you? @Garrybest

cmd/controller-manager/app/controllermanager.go
Garrybest: > But, I think the feature is part of Failover.

Actually, I think this feature is independent. We implement NoExecute taint manager, so another effect of taint is introduced into karmada. Users may taint the cluster by themselves manually. Meanwhile, `Failover` does depend on this feature, right?

pkg/controllers/cluster/taint_manager.go
Garrybest: I have already checked here first, see L141.

```go
	if !binding.DeletionTimestamp.IsZero() || !binding.Spec.TargetContains(cluster) {
		return nil
	}
```

So `len(result) < len(binding.Spec.Clusters)` is just a double check, no meaning here. I could remove this check.

cmd/controller-manager/app/controllermanager.go
Garrybest: Oh, the feature gate should be added in #1781. Let me make a change here.

Garrybest: Add a feature gate check in `cluster_controller`, it should  have been added in https://github.com/karmada-io/karmada/pull/1781. Thanks for reminding!

RainbowMango: > Actually, I think this feature is independent. We implement NoExecute taint manager, so another effect of taint is introduced into karmada. Users may taint the cluster by themselves manually. Meanwhile, Failover does depend on this feature, right?

After the eviction, the scheduler would select another cluster to fill the slot(if possible), which is exactly the scenario of Failover. For the case users taint the cluster, I think it's a special scenario that users initiate the `failover` by themselves. 

By the way, how do think about the `failover` feature? What are the use cases?

RainbowMango: > Oh, the feature gate should be added in https://github.com/karmada-io/karmada/pull/1781. Let me make a change here.

I'm ok with it. The feature gate means the feature is just introduced, maybe not mature enough to default enable, right?

cmd/controller-manager/app/controllermanager.go
```diff
@@ -170,6 +171,11 @@ func init() {
 func startClusterController(ctx controllerscontext.Context) (enabled bool, err error) {
 	mgr := ctx.Mgr
 	opts := ctx.Opts
+
+	if err := cluster.IndexField(mgr); err != nil {
+		return false, err
+	}
```
RainbowMango, line 177: The indexer is only used by the taint manager, right? 
So it should be set only when `--enable-taint-manager=true`.

pkg/controllers/cluster/taint_manager.go
RainbowMango: Yeah, I agree to remove the redundant check.

RainbowMango:
Please rebase your branch as we are fixing the fake testing recently.

cmd/controller-manager/app/controllermanager.go
Garrybest: > By the way, how do think about the failover feature? What are the use cases?

1. Cluster has something wrong, ready condition becomes not ready or unknown.
2. Cluster controller tries to add NoExecute taint after 5 minutes, #1781.
3. Taint manager removes scheduling result of failed cluster in `rb.spec.clsuters`, #1945. It's a little bit like what `descheduler` does.
4. Scheduler tries to scale up these evicted replicas.

> I'm ok with it. The feature gate means the feature is just introduced, maybe not mature enough to default enable, right?

Not exactly, we just try to move `Failover` from scheduler to controller-manager.

cmd/controller-manager/app/controllermanager.go
Garrybest: Agree, now only taint manager need this index field. If needed by other controllers in the future, we could remove this check.

cmd/controller-manager/app/controllermanager.go
RainbowMango: > Not exactly, we just try to move Failover from scheduler to controller-manager.

I see the plan at #1762. I think removing `failover` from the scheduler means relieving some responsibilities from the scheduler and letting the scheduler focus on `select` cluster for workload according to PropgationPolicy. Are we on the same page?

Anyway, let's have a talk at next week's meeting, are you available?

Garrybest: Cool. No problem.

cmd/controller-manager/app/controllermanager.go
```diff
@@ -189,6 +198,19 @@ func startClusterController(ctx controllerscontext.Context) (enabled bool, err e
 	if err := clusterController.SetupWithManager(mgr); err != nil {
 		return false, err
 	}
+
+	if ctx.Opts.EnableTaintManager {
+		taintManager := &cluster.NoExecuteTaintManager{
```
RainbowMango, line 203: Why not move the indexer thing here?

XiShanYongYe-Chang (commented):
Looks good to me.

RainbowMango (commented):
/lgtm
/approve

/hold
for 
no release notes?

RainbowMango:
/hold cancel
PS: I updated the release notes to the PR description.

Garrybest:
OK.

PR #1960, opened 2022-06-05
rename cache resource annotation

calvin0327:

**What type of PR is this?**
/kind cleanup

Minor modification for code specification.

Fixes #

I've been learning karmada's code, please correct me if not.

```release-note
`karmada-search`: The annotation `cluster.karmada.io/name` which is used to represent the source of cache now has been changed to `resource.karmada.io/cached-from-cluster`.
```

### DIFF
--- a/pkg/registry/search/storage/cache.go
+++ b/pkg/registry/search/storage/cache.go
@@ -140,8 +140,8 @@ func addAnnotationWithClusterName(resourceObjects []runtime.Object, clusterName 
 		if annotations == nil {
 			annotations = make(map[string]string)
 		}
-		// TODO: move this annotation key `cluster.karmada.io/name` to the Cluster API.
-		annotations["cluster.karmada.io/name"] = clusterName
+
+		annotations[clusterv1alpha1.CacheSourceAnnotationKey] = clusterName
 
 		resource.SetAnnotations(annotations)
 		resources = append(resources, resource)
--- a/pkg/apis/cluster/v1alpha1/well_known_constants.go
+++ b/pkg/apis/cluster/v1alpha1/well_known_constants.go
@@ -4,7 +4,6 @@ const (
 	// TaintClusterUnscheduler will be added when cluster becomes unschedulable
 	// and removed when cluster becomes scheduable.
 	TaintClusterUnscheduler = "cluster.karmada.io/unschedulable"
-
 	// TaintClusterNotReady will be added when cluster is not ready
 	// and removed when cluster becomes ready.
 	TaintClusterNotReady = "cluster.karmada.io/not-ready"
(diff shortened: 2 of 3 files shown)


calvin0327:
cc @XiShanYongYe-Chang  thanks

XiShanYongYe-Chang:
@calvin0327 Thanks, it's my to-do list. Let me take a review.

pkg/apis/cluster/v1alpha1/types.go
```diff
@@ -14,6 +14,8 @@ const (
 	ResourcePluralCluster = "clusters"
 	// ResourceNamespaceScopedCluster indicates if Cluster is NamespaceScoped.
 	ResourceNamespaceScopedCluster = false
+	// ClusterNameAnnotation is annotation key of cluster name.
+	ClusterNameAnnotation = "cluster.karmada.io/name"
```
XiShanYongYe-Chang, line 18: We can move this key to the `well_known_annotations.go` file.

@RainbowMango How do you think this `ClusterNameAnnotation` name?

RainbowMango: That's exactly what I'm thinking about.
Wait for a while, let me see if there is a better name.

calvin0327: 😄，naming is harder than coding.

calvin0327: @RainbowMango @XiShanYongYe-Chang  how about `ClusterAnnotationKeyName`?

XiShanYongYe-Chang: Better than the previous :)

pkg/registry/search/storage/cache.go
```diff
@@ -140,8 +140,8 @@ func addAnnotationWithClusterName(resourceObjects []runtime.Object, clusterName
 		if annotations == nil {
 			annotations = make(map[string]string)
 		}
-		// TODO: move this annotation key `cluster.karmada.io/name` to the Cluster API.
-		annotations["cluster.karmada.io/name"] = clusterName
+
+		annotations[clusterv1alpha1.ClusterAnnotationKeyName] = clusterName
```
RainbowMango, line 144: This annotation indicates which cluster a `cached resource` comes from.
So, I think the sub-domain should not be `cluster`, but `resource`.

My suggestion is:
```golang
	// CacheSourceAnnotationKey is the annotation that added to a resource to
	// represent which cluster it cached from.
	CacheSourceAnnotationKey = "resource.karmada.io/cached-from-cluster"
```

PS: `cluster` API might not be the most appropriate place to put this annotation, but since we don't have any other better place. So, it's a tradeoff here.

calvin0327: ok, thanks point it.  `CacheSourceAnnotationKey = "resource.karmada.io/cached-from-cluster"` is reasonable.

RainbowMango:
/retitle rename cache resource annotation

/approve

Leave LGTM to @XiShanYongYe-Chang or @liys87x to confirm.

XiShanYongYe-Chang (commented):
/lgtm

PR #1968, opened 2022-06-06
fix the incorrect use of ServerGroupsAndResources() in client-go

duanmengkk:

**What type of PR is this?**

/kind feature

Fixes #

```release-note
`karmada-controller-manager/karmada-agent`: cluster status controller now able to collect partial API list in case of discovery failure.
```

### DIFF
--- a/pkg/controllers/status/cluster_status_controller.go
+++ b/pkg/controllers/status/cluster_status_controller.go
@@ -169,8 +169,10 @@ func (c *ClusterStatusController) syncClusterStatus(cluster *clusterv1alpha1.Clu
 
 		// get the list of APIs installed in the member cluster
 		apiEnables, err := getAPIEnablements(clusterClient)
-		if err != nil {
-			klog.Errorf("Failed to get APIs installed in Cluster %s. Error: %v.", cluster.GetName(), err)
+		if len(apiEnables) == 0 {
+			klog.Errorf("Failed to get any APIs installed in Cluster %s. Error: %v.", cluster.GetName(), err)
+		} else if err != nil {
+			klog.Warningf("Maybe get partial(%d) APIs installed in Cluster %s. Error: %v.", len(apiEnables), cluster.GetName(), err)
 		}
 
 		nodes, err := listNodes(clusterInformerManager)
@@ -364,11 +366,14 @@ func getKubernetesVersion(clusterClient *util.ClusterClient) (string, error) {
 	return clusterVersion.GitVersion, nil
 }
 
+// getAPIEnablements returns the list of API enablement(supported groups and resources).
+// The returned lists might be non-nil with partial results even in the case of non-nil error.
 func getAPIEnablements(clusterClient *util.ClusterClient) ([]clusterv1alpha1.APIEnablement, error) {
 	_, apiResourceList, err := clusterClient.KubeClient.Discovery().ServerGroupsAndResources()
-	if err != nil {
+	if len(apiResourceList) == 0 {
 		return nil, err
 	}
+
 	var apiEnablements []clusterv1alpha1.APIEnablement
 	for _, list := range apiResourceList {
 		var apiResources []clusterv1alpha1.APIResource
@@ -391,7 +396,7 @@ func getAPIEnablements(clusterClient *util.ClusterClient) ([]clusterv1alpha1.API
 	sort.SliceStable(apiEnablements, func(i, j int) bool {
 		return apiEnablements[i].GroupVersion < apiEnablements[j].GroupVersion
 	})
-	return apiEnablements, nil
+	return apiEnablements, err
 }
 
 // listPods returns the Pod list from the informerManager cache.


pkg/controllers/status/cluster_status_controller.go
```diff
@@ -364,11 +366,16 @@ func getKubernetesVersion(clusterClient *util.ClusterClient) (string, error) {
 	return clusterVersion.GitVersion, nil
 }
 
+// getAPIEnablements get the list of APIs installed in the member cluster
```
RainbowMango, line 369:
```suggestion
// getAPIEnablements returns the list of API enablement(supported groups and resources).
// The returned lists might be non-nil with partial results even in the case of non-nil error.
```

pkg/controllers/status/cluster_status_controller.go
```diff
@@ -364,11 +366,16 @@ func getKubernetesVersion(clusterClient *util.ClusterClient) (string, error) {
...
 func getAPIEnablements(clusterClient *util.ClusterClient) ([]clusterv1alpha1.APIEnablement, error) {
+	// This can return an error *and* the results it was able to find.
+	// If the length of api resources is not zero,we don't need to fail on the error.
```
RainbowMango, line 372:
```suggestion
```

pkg/controllers/status/cluster_status_controller.go
```diff
@@ -169,8 +169,10 @@ func (c *ClusterStatusController) syncClusterStatus(cluster *clusterv1alpha1.Clu
 
 		// get the list of APIs installed in the member cluster
 		apiEnables, err := getAPIEnablements(clusterClient)
-		if err != nil {
-			klog.Errorf("Failed to get APIs installed in Cluster %s. Error: %v.", cluster.GetName(), err)
+		if err != nil && apiEnables == nil {
+			klog.Errorf("Failed to get any APIs installed in Cluster %s. Error: %v.", cluster.GetName(), err)
+		} else if err != nil && apiEnables != nil {
+			klog.Warningf("Failed to get all APIs installed in Cluster %s. Error: %v.", cluster.GetName(), err)
 		}
```
RainbowMango, line 176:
```suggestion
		if len(apiEnables) == 0 {
			klog.Errorf("Failed to get any APIs installed in Cluster %s. Error: %v.", cluster.GetName(), err)
		} else if err != nil {
			klog.Warningf("Maybe get partial(%d) APIs installed in Cluster %s. Error: %v.", len(apiEnables), cluster.GetName(), err)
		}
```
Does this looks better?

pkg/controllers/status/cluster_status_controller.go
```diff
@@ -364,11 +366,16 @@ func getKubernetesVersion(clusterClient *util.ClusterClient) (string, error) {
...
 }
 
+// getAPIEnablements get the list of APIs installed in the member cluster
 func getAPIEnablements(clusterClient *util.ClusterClient) ([]clusterv1alpha1.APIEnablement, error) {
+	// This can return an error *and* the results it was able to find.
+	// If the length of api resources is not zero,we don't need to fail on the error.
 	_, apiResourceList, err := clusterClient.KubeClient.Discovery().ServerGroupsAndResources()
-	if err != nil {
+	if len(apiResourceList) == 0 {
+		klog.Errorf("unable to get any supported resources from server. Error: %v.", err)
```
RainbowMango, line 375:
```suggestion
```
We could remove this log as it missing `cluster` name, and is redundant with the outside logs.

RainbowMango:
By the way, have you run some tests for the case we can only get partial results? Please share the testing report.

pkg/controllers/status/cluster_status_controller.go
duanmengkk: I have some reservations. ` len(apiEnables) == 0` does not means a Error,because it is a result a filtered by the result of `_, apiResourceList, err := clusterClient.KubeClient.Discovery().ServerGroupsAndResources()`,see line 379 ~ 340 （apiEnables是经过过滤处理后得到的结果，如果apiResourceList 不为nil，但是经过过滤处理后apiEnables为0，这个情况是正常的，虽然这种情况k8s不太可能出现)

duanmengkk:
> By the way, have you run some tests for the case we can only get partial results? Please share the testing report.

ok

duanmengkk:
> By the way, have you run some tests for the case we can only get partial results? Please share the testing report.

1.apply a  random apiservice in member1.`kubectl --kubeconfig /root/.kube/members.config --context member1 apply -f apiservice.yaml `,The output is 
image
2.The condition of cluster1 ` kubectl --kubeconfig /root/.kube/karmada.config --context karmada-apiserver get cluster member1 -o yaml`  is 
image
3.The Warn message in controller manager is 
image
4. schduler work fine 
image

pkg/controllers/status/cluster_status_controller.go
RainbowMango: > ApiEnables is the result of filtering. If the apiResourceList is not nil, but the lenth of apiEnables is 0 after filtering, this is normal, although the situation in k8S cannot happen, but it doesn't make sense logically

We just filtered sub-resources, it `should` returns a non-nil `apiEnables`, otherwise, I can assume there must be an error.

RainbowMango:
> apply a random apiservice in member1.kubectl --kubeconfig /root/.kube/members.config --context member1 apply -f apiservice.yaml ,

What's the `partial` API enablement look like?

pkg/controllers/status/cluster_status_controller.go
duanmengkk: > it `should` returns a non-nil `apiEnables`, otherwise, I can assume there must be an error.

ok,if so,line 373-375 should be deleted,and pass the result and err to the outside directly,right?
image

RainbowMango: We can keep it as a short path. It's not a big deal. Both are ok to me.

RainbowMango:
Could you please squash the commits and we are ready to go?

Thanks.

RainbowMango:
I mean is the partial API list collected to `Cluster` object as expected, you can check that by:
```
kubectl get clusters.cluster.karmada.io member1 -o=jsonpath='{.status.apiEnablements}'
```

duanmengkk:
> What's the `partial` API enablement look like?



> I mean is the partial API list collected to `Cluster` object as expected, you can check that by:
> 
> ```
> kubectl get clusters.cluster.karmada.io member1 -o=jsonpath='{.status.apiEnablements}'
> ```

sorry,the output is 
image

RainbowMango:
/lgtm
/approve

Thanks @duanmengkk

RainbowMango:
/retitle cluster status controller should be able to collect partial API list in case of discovery failure

PR #1982, opened 2022-06-09
update crd version

my-git9:
update crd version in doc

/kind cleanup


RainbowMango (approved):
Thanks. 
Maybe we don't need to do this, because here is just an example.

/lgtm
/approve

PR #1998, opened 2022-06-14
Clean-up unused code for upgrade scene

XiShanYongYe-Chang:

**What type of PR is this?**

/kind cleanup

In Karmada v1.0, considering the upgrade scene, we ensure impersonation secret for each cluster at the start of the controller,  it's unneeded for now(after Karmada v1.2).

Fixes #

```release-note
NONE
```


XiShanYongYe-Chang:
> The 'default' service account should not be propagated automatically. As the default sa will be created by kube-controller-manager when namespace created.

I think you are not commenting on the current PR.

PR #2008, opened 2022-06-16
Add pprof for karmada components

Poor12:

**What type of PR is this?**
/kind feature

Since we need to do some performance tests for karmada components, we found that there are still some components which do not support pprof.

Because each components have different implementing logic, some are based on genericAPIServer, some are base on controller-runtime(controller-runtime now do not support pprof https://github.com/kubernetes-sigs/controller-runtime/pull/1943). In this PR, we unify the config to enable pprof.

Fixes #

```Karmada-agent, Karmada-aggregated-apiserver, Karmada-controller-manager, Karmada-descheduler, Karmada-search, Karmada-scheduler, Karmada-scheduler-estimator, Karmada-webhook now support --enable-pprof to enable pprof and --profiling-bind-address to config address for pprof web server.```

### DIFF
--- a/cmd/agent/app/agent.go
+++ b/cmd/agent/app/agent.go
@@ -34,6 +34,7 @@ import (
 	"github.com/karmada-io/karmada/pkg/resourceinterpreter"
 	"github.com/karmada-io/karmada/pkg/sharedcli"
 	"github.com/karmada-io/karmada/pkg/sharedcli/klogflag"
+	"github.com/karmada-io/karmada/pkg/sharedcli/profileflag"
 	"github.com/karmada-io/karmada/pkg/util"
 	"github.com/karmada-io/karmada/pkg/util/gclient"
 	"github.com/karmada-io/karmada/pkg/util/helper"
@@ -104,6 +105,9 @@ func init() {
 
 func run(ctx context.Context, karmadaConfig karmadactl.KarmadaConfig, opts *options.Options) error {
 	klog.Infof("karmada-agent version: %s", version.Get())
+
+	profileflag.ListenAndServe(opts.ProfileOpts)
+
 	controlPlaneRestConfig, err := karmadaConfig.GetRestConfig(opts.KarmadaContext, opts.KarmadaKubeConfig)
 	if err != nil {
 		return fmt.Errorf("error building kubeconfig of karmada control plane: %w", err)
--- a/cmd/agent/app/options/options.go
+++ b/cmd/agent/app/options/options.go
@@ -10,6 +10,7 @@ import (
 	"k8s.io/client-go/tools/leaderelection/resourcelock"
 	componentbaseconfig "k8s.io/component-base/config"
 
+	"github.com/karmada-io/karmada/pkg/sharedcli/profileflag"
 	"github.com/karmada-io/karmada/pkg/sharedcli/ratelimiterflag"
 	"github.com/karmada-io/karmada/pkg/util"
 )
@@ -88,6 +89,7 @@ type Options struct {
 	MetricsBindAddress string
 
 	RateLimiterOpts ratelimiterflag.Options
+	ProfileOpts     profileflag.Options
 }
 
 // NewOptions builds an default scheduler options.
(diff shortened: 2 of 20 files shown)

### CODE REVIEW

**Comment by RainbowMango** [comment]
Please share the test reports and remember to cc the students who are doing the performance thing.

**Comment by Poor12** [comment]
1. First enable the flag.
image
2. Curl PodIP to get pprof file.
image

File: cmd/agent/app/agent.go [hunk 1]
```diff
@@ -4,6 +4,9 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"net/http"
+	_ "net/http/pprof" // #nosec
```
RainbowMango, line 8: Why set `nosec` here? 
Please add a comment for the reason.

cmd/agent/app/options/options.go
```diff
@@ -117,5 +122,7 @@ func (o *Options) AddFlags(fs *pflag.FlagSet, allControllers []string) {
 	fs.IntVar(&o.ConcurrentClusterSyncs, "concurrent-cluster-syncs", 5, "The number of Clusters that are allowed to sync concurrently.")
 	fs.IntVar(&o.ConcurrentWorkSyncs, "concurrent-work-syncs", 5, "The number of Works that are allowed to sync concurrently.")
 	fs.StringVar(&o.MetricsBindAddress, "metrics-bind-address", ":8080", "The TCP address that the controller should bind to for serving prometheus metrics(e.g. 127.0.0.1:8088, :8088)")
+	fs.BoolVar(&o.EnableProfile, "enable-pprof", false, "The flag about whether to enable pprof profiling.")
+	fs.IntVar(&o.ProfilePort, "pprof-port", 6060, "The port for pprof profiling.")
```
RainbowMango, line 126: I'm thinking if we should support specify the address. Not sure yet.

by the way, how to validate the ports?

lonelyCZ:
> I'm thinking if we need to introduce the shared flags, just like what we did at [klog-flag](https://github.com/karmada-io/karmada/tree/master/pkg/sharedcli/klogflag).

I think it is a good idea. These flags apply to `karmada-agent` and `karmada-controller-manager`. Such as `pkg/sharedcli/pprofflag/pprofflag.go`

RainbowMango:
Maybe the reason why `karmada-schedule` enabled `pprof` is:
- imports `k8s.io/apiserver/pkg/server` at [here](https://github.com/karmada-io/karmada/blob/67996bb5c881949f737d74835dfe690bee70657f/cmd/scheduler/main.go#L6) -->
- `k8s.io/apiserver/pkg/server` imports `k8s.io/apiserver/pkg/server/routes` at [here](https://github.com/karmada-io/karmada/blob/67996bb5c881949f737d74835dfe690bee70657f/vendor/k8s.io/apiserver/pkg/server/config.go#L62)
- `k8s.io/apiserver/pkg/server/routes` imports `net/http/pprof` at [here](https://github.com/karmada-io/karmada/blob/67996bb5c881949f737d74835dfe690bee70657f/vendor/k8s.io/apiserver/pkg/server/routes/profiling.go#L21)

And finally,  `karmada-scheduler` [starts the `http`-server](https://github.com/karmada-io/karmada/blob/67996bb5c881949f737d74835dfe690bee70657f/cmd/scheduler/app/scheduler.go#L179).

@Poor12 
Please confirm.
If we are going to introduce flags for profiling, we should ensure the profiling feature could be disabled. 
So, I think we need another PR before this.

By the way, the similar issue happens at `karmada-descheduler`.

cmd/agent/app/agent.go
```diff
@@ -104,6 +106,18 @@ func init() {
 
 func run(ctx context.Context, karmadaConfig karmadactl.KarmadaConfig, opts *options.Options) error {
 	klog.Infof("karmada-agent version: %s", version.Get())
+
+	if opts.ProfileOpts.EnableProfile {
+		addr := fmt.Sprintf(":%d", opts.ProfileOpts.ProfilePort)
+		klog.Infof("Starting profiling on port %s", opts.ProfileOpts.ProfilePort)
+		go func() {
+			if err := http.ListenAndServe(addr, nil); err != nil {
```
RainbowMango, line 114: 1. we shouldn't use default `DefaultServeMux`
2. we can abstract a common function to start the server in `pkg/sharedcli/profileflag/`, like:
```
func ListenAndServe(opts Options)
```

RainbowMango (commented):
Generally looks good to me.

RainbowMango:
Please describe how to grab the pprof data from Karmada build by `hack/karmada-local-up.sh`.

Poor12:
cc @lonelyCZ to review. For further usage, you can base on https://github.com/karmada-io/karmada/pull/2141

lonelyCZ:
I just tested it that worked fine. Thanks~

/lgtm

PR #2017, opened 2022-06-17
[Proposal] Adoption of ClusterID

XiShanYongYe-Chang:

**What type of PR is this?**

/kind feature

Add proposal of adoption of ClusterID.

Fixes #

```release-note
NONE
```

### DIFF
--- a/docs/proposals/clusterid/README.md
+++ b/docs/proposals/clusterid/README.md
@@ -1,0 +1,122 @@
+---
+title: Adoption of ClusterID
+
+authors:
+- "@XiShanYongYe-Chang"
+- "@RainbowMango"
+
+reviewers:
+- TBD
+
+approvers:
+- TBD
+
+creation-date: 2022-06-16
+
+---
+
+# Adoption of ClusterID
+
+## Summary
+
+In the progress of registering a cluster into Karmada, we need to confirm the cluster identity to prevent the cluster from being added to Karmada repeatedly. [Kubernetes sig-multicluster](https://github.com/kubernetes/community/tree/master/sig-multicluster) have made a KEP: [ClusterId for ClusterSet identification](https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/2149-clusterid#kep-2149-clusterid-for-clusterset-identification) to identify a cluster in a Kubernetes-native way. We want to use this [ClusterID](https://github.com/kubernetes-sigs/about-api) to confirm the cluster identifier. 
+
+## Motivation
+
+Kubernetes' enhancements has proposed a [KEP-2149](https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/2149-clusterid). This KEP proposes an API called `ClusterProperty` which provides a way to store identification related, cluster scoped information. We can adopt this API to identify each cluster that joins Karmada.
+
+### Goals
+
+- Set the cluster identifier for the cluster that has been joined to Karmada.
+
+### Non-Goals
+
+- The API describing cluster properties upgrade.
+
+## Proposal
+
+### User Stories (Optional)
+
+#### Story 1
+
+As a cluster administrator, I want each cluster could be assigned with a unique ID, so that the registering process can be rejected if a cluster tries to register with another name.
+
+[Issue](https://github.com/karmada-io/karmada/issues/880) raises this situation.
+
+#### Story 2
+
+As a cluster administrator, when I create a `id.k8s.io ClusterProperty` in a kubernetes cluster, I want to use it as its cluster ID after joining Karmada, so that to uniquely identify the cluster.
+
+### Notes/Constraints/Caveats (Optional)
+
+### Risks and Mitigations
+
+## Design Details
+
+We can divide the implementation into two phases:
+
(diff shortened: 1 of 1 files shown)


calvin0327:
@XiShanYongYe-Chang Thanks, I think this proposal is feasible. as our company has a product with similar logic and many of our company partners (@wawa0210 @zgfh @chaunceyjiang ...) also contribute to Karmada. so after the proposal is discussed clearly, I think we can work on this feature together?

XiShanYongYe-Chang:
Okay, thanks.

docs/proposals/clusterid/README.md
```diff
@@ -0,0 +1,118 @@
...
+type ClusterStatus struct {
+    // ClusterProperty represents the cluster identity information.
+    // +optional
+    ClusterProperty map[string]string `json:"clusterProperty,omitempty"`
+  
+	...
+}
+```
+
+We provide a flag (such as `create-clusterproperty-not-exist`) to tell Karmada whether to create `id.k8s.io ClusterProperty` when it does not exist in the Kubernetes Cluster. During the cluster joining process, Karmada will look for if the kubernetes cluster hase defined the cluster identifier by `id.k8s.io ClusterProperty`. If so, we just take the `id.k8s.io ClusterProperty` value as cluster id; otherwise, when the `create-clusterproperty-not-exist` flag is true, we'll just use kubernetes cluster's `kube-system` namespace uuid to create `id.k8s.io ClusterProperty`, when the `create-clusterproperty-not-exist` flag is false, we refuse to join the cluster into Karmada.
```
Garrybest, line 98: hase -> has

docs/proposals/clusterid/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+We propose to introduce a `ClusterProperty` filed to the Cluster API:
+
+```go
+// ClusterStatus contains information about the current status of a
+// cluster updated periodically by cluster controller.
+type ClusterStatus struct {
+    // ClusterProperty represents the cluster identity information.
+    // +optional
+    ClusterProperty map[string]string `json:"clusterProperty,omitempty"`
```
Garrybest, line 92: Why is `map` ? I assume a cluster only has one property, right?

Garrybest (commented):
Good job👍

calvin0327:
@XiShanYongYe-Chang  Is the `cluster` resource creation phase referred to in controller or admission?

docs/proposals/clusterid/README.md
calvin0327: there should be a string

XiShanYongYe-Chang: There may be multiple ClusterProperty with different names(such as `id.k8s.io`, `xx.karmada.io`, etc), we may need to collect them all.

XiShanYongYe-Chang:
> @XiShanYongYe-Chang Is the `cluster` resource creation phase referred to in controller or admission? Which of the two is better？

My personal understanding is `admission`, I'm sorry that I don't know `In controller` means where place?

calvin0327:
> > @XiShanYongYe-Chang Is the `cluster` resource creation phase referred to in controller or admission? Which of the two is better？
> 
> My personal understanding is `admission`, I'm sorry that I don't know `In controller` means where place?

😄，The `in cluster` indicated cluster_controller.go, however, I think it would be better to use adminssion to validate the cluster ID.

XiShanYongYe-Chang:
@calvin0327 Thanks for your explanation 😄.

docs/proposals/clusterid/README.md
XiShanYongYe-Chang: Thanks, update.

docs/proposals/clusterid/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Proposal
+
+### User Stories (Optional)
+
+#### Story 1
+
+In Karmada, a cluster can be joined many times with different name. For a cluster joined Karmada repeatedly, once we remove one of them, all the clusters will turn into unready. More detail info can refer to [issue](https://github.com/karmada-io/karmada/issues/880).
+
+So we need to verify the cluster uniques when cluster joining.
```
RainbowMango, line 44: The `User Story` is used to capture a description of a feature from a `user's perspective`. The user story describes the type of user, what they want and why. 
See [What's user story](https://www.techtarget.com/searchsoftwarequality/definition/user-story#:~:text=A%20user%20story%20is%20a,simplified%20description%20of%20a%20requirement.)

RainbowMango: As a cluster administrator, I want each cluster could be assigned with a unique ID, so that the registering process can be rejected if a cluster tries to register with another name.

For example.

docs/proposals/clusterid/README.md
RainbowMango: I'm not sure if we need to show cluster property in `Cluster` object, do you need to sync the changes made from the member cluster? 
I can't see much evidence for that, at least for now.

RainbowMango:
Generally looks good to me.
I'll make some changes based on this version.
/lgtm
/approve

PR #1989, opened 2022-06-10
dump controllerruntime's logs to klog

ikaven1024:
Signed-off-by: yingjinhui <yingjinhui@didiglobal.com>

**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
The default logs of controllerruntime is dummy, and errors will not be printed. It is unfriendly for developing.
https://github.com/karmada-io/karmada/blob/04365bd9f6219e211db0d647fe5614ee8d0fa6f3/vendor/sigs.k8s.io/controller-runtime/pkg/log/log.go#L77-L85

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:
update klog to `v2.60.1`, as same as [kubernetes v1.24.1](https://github.com/kubernetes/kubernetes/blob/v1.24.1/go.mod#L123). Which implements `logr.Logger`

**Does this PR introduce a user-facing change?**:
```release-note
NONE
```

Issue #1866, opened 2022-05-23
Transform objects in informer


RainbowMango:
**What would you like to be added**:
Adopt `SetTransform` feature of `client-go`.

**Why is this needed**:
Save memory usage for the informers.

### References
background: https://github.com/karmada-io/karmada/pull/1858#issuecomment-1134114741
SetTransform feature: https://github.com/kubernetes/kubernetes/pull/107507

### Dependencies 
- [ ] Bump Kubernetes dependencies to v1.24 (do it after release-1.2)

dddddai:
Transforming pods would prevent users from propagating pods directly, personally I don't think users would do it though... but anyway it's a breaking change, do you think it's ok to ignore it?

RainbowMango:
> Transforming pods would prevent users from propagating pods directly, 

Why do you think that?  I haven't looked at this feature carefully but seems we can `null out` unused fields to save ram cost.

PR #2032, opened 2022-06-20
Update dependencies Go to 1.18.3

RainbowMango:
**What type of PR is this?**
/kind feature

We are going to update Kubernetes dependencies to `v1.24.2` which is built with `Go 1.18.3`. PRs/Issues that requires the dependencies are:
- #1866 
- #1989

This PR focus on updating the Golang version.

And, the golangci-lint needs an update seems current `v1.42.2` can't work with go 1.18:
- https://github.com/RainbowMango/karmada/runs/6960119555?check_suite_focus=true

Fixes #

```release-note
Karmada is now built with Golang 1.18.3. 
```


XiShanYongYe-Chang (commented):
/lgtm

PR #2033, opened 2022-06-20
Skip updating lease for push mode clusters

dddddai:
**What type of PR is this?**
/kind cleanup

It's pointless to update lease for push mode clusters since the cluster status controller is at the control-plane
Meanwhile it's consuming goroutines

Fixes #

```release-note
`karmada-controller-manager`: Stopped reporting and refreshing `lease` for clusters with `Push` mode.
```

### DIFF
--- a/pkg/controllers/cluster/cluster_controller.go
+++ b/pkg/controllers/cluster/cluster_controller.go
@@ -420,6 +420,10 @@ func (c *Controller) tryUpdateClusterHealth(ctx context.Context, cluster *cluste
 		}
 	}
 
+	if cluster.Spec.SyncMode == clusterv1alpha1.Push {
+		return observedReadyCondition, currentReadyCondition, nil
+	}
+
 	// Always update the probe time if cluster lease is renewed.
 	// Note: If cluster-status-controller never posted the cluster status, but continues renewing the
 	// heartbeat leases, the cluster controller will assume the cluster is healthy and take no action.
--- a/pkg/controllers/status/cluster_status_controller.go
+++ b/pkg/controllers/status/cluster_status_controller.go
@@ -156,8 +156,10 @@ func (c *ClusterStatusController) syncClusterStatus(cluster *clusterv1alpha1.Clu
 			klog.Errorf("Failed to get or create informer for Cluster %s. Error: %v.", cluster.GetName(), err)
 		}
 
-		// init the lease controller for every cluster
-		c.initLeaseController(clusterInformerManager.Context(), cluster)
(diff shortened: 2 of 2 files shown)


Garrybest:
What if `karmada-controller-manager` is crashed? The condition of push mode clusters would not changed.

Garrybest:
Oh I see. If a lease of a pushed cluster is no longer renewed, it means the controller-manager is done. So the status will never change.

/lgtm

pkg/controllers/cluster/cluster_controller.go
```diff
@@ -420,6 +420,10 @@ func (c *Controller) tryUpdateClusterHealth(ctx context.Context, cluster *cluste
 		}
 	}
 
+	if cluster.Spec.SyncMode == clusterv1alpha1.Push {
+		return observedReadyCondition, currentReadyCondition, nil
+	}
+
```
RainbowMango, line 426: Why not check it earlier? At least we can do it before `Step 3`

Can we do it at the head of this function(tryUpdateClusterHealth)?

RainbowMango:
> Oh I see. If a lease of a pushed cluster is no longer renewed, it means the controller-manager is done. So the status will never change.

Is it possible that `cluster-status-controller` hanging(maybe just slow) there thus can't report lease?

pkg/controllers/cluster/cluster_controller.go
dddddai: But we still need to maintain `clusterHealth.readyTransitionTimestamp` for [processTaintBaseEviction](https://github.com/karmada-io/karmada/blob/v1.2.0/pkg/controllers/cluster/cluster_controller.go#L492)

dddddai:
> Is it possible that `cluster-status-controller` hanging(maybe just slow) there thus can't report lease?

Then I don't think we should [consider the cluster status as `Unknown`](https://github.com/karmada-io/karmada/blob/v1.2.0/pkg/controllers/cluster/cluster_controller.go#L456)(due to lease timeout)

pkg/controllers/cluster/cluster_controller.go
RainbowMango: oh.. right

RainbowMango:
@dddddai I think we need a release note for this patch.

dddddai:
Thanks, added

RainbowMango:
```
`karmada-controller-manager`: Stoped report and refresh `lease` for clusters with `Push` mode.
```
Does this make sense to you?

RainbowMango:

/approve

PR #2034, opened 2022-06-20
Performance improve for karmada-controller-manager

Poor12:

**What type of PR is this?**
/kind feature

As is talked in #1858，when member cluster is large(3k nodes, 10w pods), karmada-controller-manager is easy to be OOMkilled and retry to sync nodes and pods again and again.

Based on #2008, I tested the two clusters with 100 nodes and almost 300 pods. We can easily see that in cluster status controllers, list nodes and pods periodicly cost a lot when the cluster is large(a lot of unmarshal works)  according to the frame graph.

So I modified the mechanism from full amount list to `watch based on event` to make the memory increase more stable.

Fixes #

None

### DIFF
--- a/pkg/controllers/status/cluster_resource_summary_cache.go
+++ b/pkg/controllers/status/cluster_resource_summary_cache.go
@@ -1,0 +1,190 @@
+package status
+
+import (
+	"fmt"
+	"sync"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	clusterv1alpha1 "github.com/karmada-io/karmada/pkg/apis/cluster/v1alpha1"
+	"github.com/karmada-io/karmada/pkg/util"
+	"github.com/karmada-io/karmada/pkg/util/helper"
+	"github.com/karmada-io/karmada/pkg/util/informermanager/keys"
+)
+
+type nodeSummary struct {
+	isReady         bool
+	nodeAllocatable corev1.ResourceList
+}
+
+type podSummary struct {
+	isAllocating bool
+	isAllocated  bool
+	podResource  *util.Resource
+}
+
+type singleClusterSummaryCache struct {
+	nodeSummaryMap map[keys.ClusterWideKey]nodeSummary
+	podSummaryMap  map[keys.ClusterWideKey]podSummary
+}
+
+func newSingleClusterSummaryCache() *singleClusterSummaryCache {
+	return &singleClusterSummaryCache{
+		nodeSummaryMap: make(map[keys.ClusterWideKey]nodeSummary),
+		podSummaryMap:  make(map[keys.ClusterWideKey]podSummary),
+	}
+}
+
+func (s *singleClusterSummaryCache) set(clusterWideKey keys.ClusterWideKey, object runtime.Object) error {
+	switch clusterWideKey.Kind {
+	case util.NodeKind:
+		node, err := helper.ConvertToNode(object.(*unstructured.Unstructured))
+		if err != nil {
+			return fmt.Errorf("failed to convert unstructured to typed object: %v", err)
+		}
+
+		s.nodeSummaryMap[clusterWideKey] = nodeSummary{isReady: helper.NodeReady(node), nodeAllocatable: node.Status.Allocatable}
+	case util.PodKind:
+		pod, err := helper.ConvertToPod(object.(*unstructured.Unstructured))
+		if err != nil {
+			return fmt.Errorf("failed to convert unstructured to typed object: %v", err)
+		}
+		s.podSummaryMap[clusterWideKey] = podSummary{isAllocating: isPodAllocating(pod), isAllocated: isPodAllocated(pod), podResource: getPodResource(pod)}
+	}
+
+	return nil
+}
+
+func (s *singleClusterSummaryCache) delete(clusterWideKey keys.ClusterWideKey) {
+	switch clusterWideKey.Kind {
+	case util.NodeKind:
+		delete(s.nodeSummaryMap, clusterWideKey)
+	case util.PodKind:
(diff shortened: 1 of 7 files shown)


Poor12:
cc @RainbowMango

pkg/controllers/status/cluster_status_controller.go
```diff
@@ -402,132 +413,104 @@ func getAPIEnablements(clusterClient *util.ClusterClient) ([]clusterv1alpha1.API
...
-
-	resourceSummary := &clusterv1alpha1.ResourceSummary{}
-	resourceSummary.Allocatable = allocatable
-	resourceSummary.Allocating = allocating
-	resourceSummary.Allocated = allocated
+func (c *ClusterStatusController) handleEvent(fedKey keys.FederatedKey) error {
+	runtimeObj, err := helper.GetObjectFromCache(c.RESTMapper, c.InformerManager, fedKey)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			c.ClusterSummaryCache.delete(fedKey)
```
ikaven1024, line 449: NotFound  error shall be returned?

pkg/controllers/status/cluster_status_controller.go
```diff
@@ -402,132 +413,104 @@ func getAPIEnablements(clusterClient *util.ClusterClient) ([]clusterv1alpha1.API
...
-		for key, val := range node.Status.Allocatable {
-			tmpCap, ok := allocatable[key]
-			if ok {
-				tmpCap.Add(val)
-			} else {
-				tmpCap = val
+func (c *ClusterStatusController) genHandlerUpdateFunc(clusterName string) func(oldObj, newObj interface{}) {
+	return func(oldObj, newObj interface{}) {
+		curObj := newObj.(runtime.Object)
+		if !reflect.DeepEqual(oldObj, newObj) {
```
ikaven1024, line 488: How do you think compare with resource version?

pkg/controllers/status/cluster_resource_summary_cache.go
```diff
@@ -0,0 +1,181 @@
...
+	defer n.Unlock()
+
+	delete(n.clusterSummary, clusterName)
+}
+
+func (n *ClusterSummaryCache) getNodeSummary(clusterName string) (*clusterv1alpha1.NodeSummary, corev1.ResourceList) {
+	n.Lock()
+	defer n.Unlock()
+
+	totalNum := len(n.clusterSummary[clusterName].nodeSummaryMap)
```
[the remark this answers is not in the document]
Poor12: Thanks for pointing out.

pkg/controllers/status/cluster_status_controller.go
Poor12: miss return statement

pkg/controllers/status/cluster_status_controller.go
Poor12: I think it's ok to comparing with obj, because we don't know the type of obj. If comparing with resource version, may add the cost of converting type.

ikaven1024: I test it on my side with go 1.17.5
```go

var (
	obj1 interface{}
	obj2 interface{}
)

func init() {
        pod1Yaml := `<skipped>`
        pod2Yaml := `<skipped>`

	unstructuredObj1 := &unstructured.Unstructured{}
	unstructuredObj2 := &unstructured.Unstructured{}
	unstructured.UnstructuredJSONScheme.Decode([]byte(pod1Yaml), nil, unstructuredObj1)
	unstructured.UnstructuredJSONScheme.Decode([]byte(pod2Yaml), nil, unstructuredObj2)

        obj1, obj2 = unstructuredObj1, unstructuredObj2
}

func BenchmarkDeepEqual(b *testing.B) {
	for i := 0; i < b.N; i++ {
		_ = reflect.DeepEqual(obj1, obj2)
	}
}

func BenchmarkCast(b *testing.B) {
	for i := 0; i < b.N; i++ {
		o1, _ := meta.Accessor(obj1)
		o2, _ := meta.Accessor(obj2)
		_ = o1.GetResourceVersion() == o2.GetResourceVersion()
	}
}
```

```
goos: darwin
goarch: arm64
pkg: github.com/ikaven1024/gists/k8s/benchmark
BenchmarkDeepEqual
BenchmarkDeepEqual-8   	 7717909	       145.2 ns/op
BenchmarkCast
BenchmarkCast-8        	33085802	        36.39 ns/op
PASS
```

Type cast is more faster than `reflect.DeepEqual`

pkg/controllers/status/cluster_status_controller.go
```diff
@@ -254,9 +264,9 @@ func (c *ClusterStatusController) buildInformerForCluster(cluster *clusterv1alph
 	// create the informer for pods and nodes
 	allSynced := true
 	for _, gvr := range gvrs {
-		if !singleClusterInformerManager.IsInformerSynced(gvr) {
+		if !singleClusterInformerManager.IsInformerSynced(gvr) || !singleClusterInformerManager.IsHandlerExist(gvr, c.getEventHandler(cluster.Name)) {
```
RainbowMango, line 267: Informer synced for `gvr` doesn't mean the `eventHandler` exit. 
Since all controllers share the `MultiClusterInformerManager`, the informer might be created(and synced from another place).

pkg/util/helper/unstructured.go
```diff
@@ -228,3 +229,20 @@ func ToUnstructured(obj interface{}) (*unstructured.Unstructured, error) {
...
+		return false
+	}
+	newMetaInfo, err := meta.Accessor(newObj)
+	if err != nil { //should not happen
+		return false
+	}
+
+	if oldMetaInfo.GetResourceVersion() != newMetaInfo.GetResourceVersion() {
+		return true
+	}
```
RainbowMango, line 246: Looks weird that `.metadata.ResourceVersion` is not changed in an update event.
I doubt it.

PR #2042, opened 2022-06-21
feature delete default etcd data in deinit

Momeaking:
Signed-off-by: Momeaking <1026415175@qq.com>

**What type of PR is this?**

kind feature

Delete ETCD data during deinit

Fixes #

```release-note
karmadactl deinit --de-etcd
```

### DIFF
--- a/pkg/karmadactl/deinit.go
+++ b/pkg/karmadactl/deinit.go
@@ -320,7 +320,10 @@ func (o *CommandDeInitOption) Run() error {
 	}
 
 	if o.DeDefaultEtcd {
-		cleanDir(etcdDefaultPath)
+		err := cleanDir(etcdDefaultPath)
+		if err != nil {
+			return err
+		}
 		fmt.Println("remove Karmada from Kubernetes successfully.\n" +
 			"\n delete etcd data  in the default path,default path is : /var/lib/karmada-etcd .")
 	} else {
--- a/pkg/karmadactl/deinit.go
+++ b/pkg/karmadactl/deinit.go
@@ -325,10 +325,10 @@ func (o *CommandDeInitOption) Run() error {
 			return err
 		}
 		fmt.Println("remove Karmada from Kubernetes successfully.\n" +
-			"\n delete etcd data  in the default path,default path is : /var/lib/karmada-etcd .")
+			"\ndelete etcd data in the default path, default path is: /var/lib/karmada-etcd.")
 	} else {
 		fmt.Println("remove Karmada from Kubernetes successfully.\n" +
-			"\ndeinit will not delete etcd data, if the etcd data is persistent, please delete it yourself.")
+			"\ndeinit will not delete etcd data, if the etcd data is persistent, please delete it by yourself.")
 	}
 	return nil
 }
--- a/pkg/karmadactl/deinit.go
+++ b/pkg/karmadactl/deinit.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


pkg/karmadactl/deinit.go
```diff
@@ -294,7 +319,16 @@ func (o *CommandDeInitOption) Run() error {
...
 
-	fmt.Println("remove Karmada from Kubernetes successfully.\n" +
-		"\ndeinit will not delete etcd data, if the etcd data is persistent, please delete it yourself.")
+	if o.DeDefaultEtcd {
+		err := cleanDir(etcdDefaultPath)
+		if err != nil {
+			return err
+		}
+		fmt.Println("remove Karmada from Kubernetes successfully.\n" +
+			"\n delete etcd data  in the default path,default path is : /var/lib/karmada-etcd .")
```
Poor12, line 328:
```suggestion
			"\ndelete etcd data in the default path, default path is: /var/lib/karmada-etcd.")
```

pkg/karmadactl/deinit.go
```diff
@@ -294,7 +319,16 @@ func (o *CommandDeInitOption) Run() error {
...
+	} else {
+		fmt.Println("remove Karmada from Kubernetes successfully.\n" +
+			"\ndeinit will not delete etcd data, if the etcd data is persistent, please delete it yourself.")
```
Poor12, line 331:
```suggestion
			"\ndeinit will not delete etcd data, if the etcd data is persistent, please delete it by yourself.")
```

prodanlabs:
There is a problem, if the host cluster has multiple nodes, and etcd is not running the current node, then this command is invalid.

prodanlabs:
By the way, if the persistence of `ectd` uses `pvc`, because after deleting the `namespace`, `pvc` will be deleted in cascade, so it is wrong to say that `deinit` will not delete `etcd` data.

Momeaking:
> By the way, if the persistence of `ectd` uses `pvc`, because after deleting the `namespace`, `pvc` will be deleted in cascade, so it is wrong to say that `deinit` will not delete `etcd` data.

First, the --t parameters are optional, and the user can choose whether to configure them at uninstall time based on how etCD data is persisted
Second,if the persistence of `ectd` uses `pvc`, you can leave this parameter out,delete it by yourself

prodanlabs:
> First, the --t parameters are optional, and the user can choose whether to configure them at uninstall time based on how etCD data is persisted Second,if the persistence of `ectd` uses `pvc`, you can leave this parameter out,delete it by yourself

agree.

Another small question, what if `etcd` is not on the current server.

Momeaking:
> delete it by yourself



> > First, the --t parameters are optional, and the user can choose whether to configure them at uninstall time based on how etCD data is persisted Second,if the persistence of `ectd` uses `pvc`, you can leave this parameter out,delete it by yourself
> 
> agree.
> 
> Another small question, what if `etcd` is not on the current server.

If this is the case, you can only delete it by yourself. In addition,This parameter is designed for installation using karmada init, similar to kubeadm init and kubeadm reset

prodanlabs:
> If this is the case, you can only delete it by yourself. In addition,This parameter is designed for installation using karmada init, similar to kubeadm init and kubeadm reset

Or output some logs, prompting the user to delete it manually.

RainbowMango:
I'm going to close this PR as it has been inactive for a long time. If you think this still needed, please feel free to re-open it.

In my opinion, there are plenty of ways people use ETCD, and the default persistence patch is unlikely to be used in production. 
So, it's complicated to consider the situation when and how to remove the ETCD data.
I want to hear more feedback from real-world use cases, and then consider how to enhance it.

Anyway, thanks for bring this. @Momeaking 

/close

PR #2052, opened 2022-06-23
karmada-search: short circuit event hander for resourceRegistry updation

calvin0327:

**What type of PR is this?**
/kind bug

If  spec of resourceRegistry has not changed, we should not to registry reource informer  once again.  in the case, short circuit resourceRegistry update event.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/search/controller.go
+++ b/pkg/search/controller.go
@@ -264,11 +264,16 @@ func (c *Controller) updateResourceRegistry(oldObj, newObj interface{}) {
 	oldRR := oldObj.(*v1alpha1.ResourceRegistry)
 	newRR := newObj.(*v1alpha1.ResourceRegistry)
 
+	if reflect.DeepEqual(oldRR.Spec, newRR.Spec) {
+		klog.V(4).Infof("Ignore ResourceRegistry(%s) update event as spec not changed", newRR.Name)
+		return
+	}
+
 	// TODO: stop resource informers if it is not in the new resource registry
 	resources := c.getResources(newRR.Spec.ResourceSelectors)
 
 	clusters := c.getClusters(newRR.Spec.TargetCluster)
-	clusterSets := make(map[string]struct{})
+	clusterSets := make(map[string]struct{}, len(clusters))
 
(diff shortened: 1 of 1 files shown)


RainbowMango:
/assign @liys87x 

@liys87x By the way, we need an E2E test for `karmada-search` basic functionalities.

liys87x:
> 

ok

liys87x:
Thank you very much @calvin0327, I have tried it.

calvin0327:
> Thank you very much @calvin0327, I will try it.

@liys87x thanks for review

pkg/search/controller.go
```diff
@@ -264,11 +264,16 @@ func (c *Controller) updateResourceRegistry(oldObj, newObj interface{}) {
 	oldRR := oldObj.(*v1alpha1.ResourceRegistry)
 	newRR := newObj.(*v1alpha1.ResourceRegistry)
 
+	if reflect.DeepEqual(oldRR.Spec, newRR.Spec) {
```
liys87x, line 267: make sense

calvin0327: okay.

calvin0327:
@RainbowMango PTAL

pkg/search/controller.go
```diff
@@ -264,11 +264,16 @@ func (c *Controller) updateResourceRegistry(oldObj, newObj interface{}) {
...
+		klog.Infof("spec of resourceRegistry[%s] has not changed", newRR.Name)
```
RainbowMango, line 268:
```suggestion
		klog.V(4).Infof("Ignore ResourceRegistry(%s) update event as spec not changed", newRR.Name) 
```

This log can be suppressed or ignored I think. I can't see why we need it.

calvin0327: @RainbowMango That's right.  the update event message doesn't make sense. 
I hadn't thought of that，it's good.
Done

RainbowMango (commented):
/lgtm
/approve

PR #2053, opened 2022-06-23
karmada-search: optimize backend store lock

calvin0327:

**What type of PR is this?**
/kind cleanup

In most cases it's reading, a small optimization to separate reading and writing. 

Fixes #

please correct me if not.

```release-note
NONE
```

### DIFF
--- a/pkg/search/backendstore/store.go
+++ b/pkg/search/backendstore/store.go
@@ -17,7 +17,7 @@ type BackendStore interface {
 }
 
 var (
-	backendLock sync.Mutex
+	backendLock sync.RWMutex
 	backends    map[string]BackendStore
 	k8sClient   *kubernetes.Clientset
 )
@@ -62,12 +62,19 @@ func DeleteBackend(cluster string) {
 
 // GetBackend get backend store
 func GetBackend(cluster string) BackendStore {
+	if bs, ok := func() (BackendStore, bool) {
+		backendLock.RLock()
+		defer backendLock.RUnlock()
+
+		bs, ok := backends[cluster]
+		return bs, ok
+	}(); ok {
+		return bs
+	}
+
 	backendLock.Lock()
 	defer backendLock.Unlock()
-	bs, ok := backends[cluster]
-	if !ok {
-		klog.Errorf("cannot find backend store %s, use defult backend", cluster)
-		return NewDefaultBackend(cluster)
-	}
-	return bs
+
+	klog.Errorf("cannot find backend store %s, use defult backend", cluster)
+	return NewDefaultBackend(cluster)
 }


RainbowMango:
Is it a bug? (I see you tagged the PR with `kind/bug`)

RainbowMango:
I guess it is more like a cleanup or improvement. (We don't have a `kind/improvement` label though)

calvin0327:
> I guess it is more like a cleanup or improvement. (We don't have a `kind/improvement` label though)

👌

pkg/search/backendstore/store.go
```diff
@@ -62,12 +62,19 @@ func DeleteBackend(cluster string) {
...
 	backendLock.Lock()
 	defer backendLock.Unlock()
-	bs, ok := backends[cluster]
-	if !ok {
-		klog.Errorf("cannot find backend store %s, use defult backend", cluster)
-		return NewDefaultBackend(cluster)
-	}
-	return bs
+
+	klog.Errorf("cannot find backend store %s, use defult backend", cluster)
```
ikaven1024, line 78: is it necessary lock for `NewDefaultBackend`?

pkg/search/backendstore/store.go
```diff
@@ -62,12 +62,19 @@ func DeleteBackend(cluster string) {
...
+		backendLock.RLock()
+		defer backendLock.RUnlock()
+
+		bs, ok := backends[cluster]
+		return bs, ok
+	}(); ok {
+		return bs
+	}
+
 	backendLock.Lock()
```
ikaven1024, line 75: is it necessary lock for NewDefaultBackend?

liys87x: unnecessary

calvin0327: If there are other goroutines writing to the backends map, `GetBackend` should be locked.  but is it necessary to use read/write locks? please corrent me if not

pkg/search/backendstore/store.go
chaunceyjiang, line 78: klog.Errorf -> klog.Warning ?

liys87x: This path does not meet user expectations, So the error log was selected

liys87x:
The `default` backend store does nothing, but the `singleClusterInformerManagerImpl.ForResource` method must provide a handler.
In the future, we will remove the `default` backen store if we can.

liys87x:
This is a low frequency operation and returning the default backend does not meet user expectations. We will not deal with this for the time being, and we will adjust it together in the future. 
What do you think? @calvin0327

calvin0327:
@liys87x okay.

PR #2058, opened 2022-06-24
Add proxy proposal

ikaven1024:
Co-authored-by: raymondmiaochaoyue <raymondmiaochaoyue@didiglobal.com>

**What type of PR is this?**
/kind design

Support resource aggregation proxy for member clusters.

Fixes #

```release-note
NONE
```


ikaven1024:
@RainbowMango

RainbowMango:
Thanks @ikaven1024 !!!

I'll start to look at it ASAP.

docs/proposals/resource-aggregation-proxy/README.md
```diff
@@ -0,0 +1,455 @@
...
+
+Apparently, use the proxy approach will greatly simplify external component's access to kubernetes cluster, and hide multiple cluster's implementation details from them. It's what we propose.
+
+
+
+
+Why not use :
+
+- **[Aggregated Kubernetes API Endpoint](https://github.com/karmada-io/karmada/blob/master/docs/userguide/aggregated-api-endpoint.md)** This component is built for Unified authentication, the restful api client still need to enumerate all the member clusters to get the right information.
+- **[Caching](https://github.com/karmada-io/karmada/tree/master/docs/proposals/caching) (aka karmada-search)** This component can only support get/list data, it cannot watch or update data.
```
Tingtal, line 44:
```suggestion
- **[Caching](https://github.com/karmada-io/karmada/tree/master/docs/proposals/caching) (aka karmada-search)** This component can only get/list, not watch or update data. 
```

docs/proposals/resource-aggregation-proxy/README.md
```diff
@@ -0,0 +1,455 @@
...
+
+
+Why not use :
+
+- **[Aggregated Kubernetes API Endpoint](https://github.com/karmada-io/karmada/blob/master/docs/userguide/aggregated-api-endpoint.md)** This component is built for Unified authentication, the restful api client still need to enumerate all the member clusters to get the right information.
+- **[Caching](https://github.com/karmada-io/karmada/tree/master/docs/proposals/caching) (aka karmada-search)** This component can only support get/list data, it cannot watch or update data. 
+
+# Motivation
+
+Write a Resource Aggregation Proxy for Karmada. This proxy provide native kubernetes api for components outside kubernetes. When client requests (include read & write operation) proxy, proxy will figure out the correct downstream kubenetes api servers (including Karmada API server and member clusters), request all of the selected servers, then merge the responses into one single response, then return the result to the client. External components can treat Karmada managed kubernetes clusters as one single cluster.
```
Tingtal, line 48:
```suggestion
Write a Resource Aggregation Proxy for Karmada. This proxy provides native kubernetes API for components outside kubernetes. When the client requests operations (including reads & writes), the proxy will figure out the correct downstream kubernetes API servers (including Karmada API server and member clusters), request all the selected servers, and merge the responses into one before responding to the client. External components can treat Karmada-managed kubernetes clusters as one single cluster.
```

docs/proposals/resource-aggregation-proxy/README.md
```diff
@@ -0,0 +1,455 @@
...
+
+## Goals
+
+Allow kubernetes api clients (kubectl, client-go, and other client use kubernetes api) read write following resources:
```
Tingtal, line 52:
```suggestion
Allow kubernetes API clients (kubectl, client-go, and other clients that use kubernetes APIs) to read/write the following resources:
```

docs/proposals/resource-aggregation-proxy/README.md
```diff
@@ -0,0 +1,455 @@
...
+
+- workload: create/get/list/watch/update/patch/delete
+- node: get/list/watch/update/patch
+- pod: get/list/watch/update/patch/logs/exec
+
+## Non-Goals
+
+- read write following resources:
```
Tingtal, line 60:
```suggestion
- read/write resources:
```

docs/proposals/resource-aggregation-proxy/README.md
```diff
@@ -0,0 +1,455 @@
...
+  - service account token
+  - event
+
+- Provide a solution for unique resource names across all member clusters.
+
+# Proposal
+
+Add a Resource Aggregation Proxy. When client try to list pod by label, this proxy will aggregate pods information from all member clusters, then return to client as a single list. When client try to get workload resource, proxy will get the information from Karmada API server directly.
```
Tingtal, line 68:
```suggestion
Add a Resource Aggregation Proxy. When the client tries to list pod by label, this proxy will aggregate pods information from all member clusters, and then return to the client as a single list. When the client tries to query workload resources, the proxy will get the information from Karmada API server directly.
```

docs/proposals/resource-aggregation-proxy/README.md
```diff
@@ -0,0 +1,455 @@
...
+
+## User Stories
+
+### Story 1: Service Discovery
+
+Not everyone use kubernetes Service for service discovery. We use a private service discovery architecture to accommodate both containerized application and traditional application deployed on physical machines.
```
Tingtal, line 74:
```suggestion
Not everyone uses kubernetes Services for service discovery. We use a private service discovery architecture to accommodate both containerized and traditional applications deployed on physical machines. 
```

docs/proposals/resource-aggregation-proxy/README.md
```diff
@@ -0,0 +1,455 @@
...
+
+## User Stories
+
+### Story 1: Service Discovery
+
+Not everyone use kubernetes Service for service discovery. We use a private service discovery architecture to accommodate both containerized application and traditional application deployed on physical machines. 
+
+![user story - service discovery](user story - service discovery.svg)
+
+Its main workflow can be summaried as following:
```
Tingtal, line 78:
```suggestion
Its main workflow can be summarized as follows:
```

docs/proposals/resource-aggregation-proxy/README.md
```diff
@@ -0,0 +1,455 @@
...
+
+1. list/watch pods.
```
Tingtal, line 80:
```suggestion
1. List/watch pods.
```

docs/proposals/resource-aggregation-proxy/README.md
```diff
@@ -0,0 +1,455 @@
...
+2. When pod's ip changed, update the corresponding application's routing metadata.
```
Tingtal, line 81:
```suggestion
2. Update the corresponding application's routing metadata when the pod IP changes.
```

docs/proposals/resource-aggregation-proxy/README.md
```diff
@@ -0,0 +1,455 @@
...
+3. Patch the pod, add a custom annotation to mark the pod as already processed.
```
Tingtal, line 82:
```suggestion
3. Patch the pod and mark it as processed by adding a custom annotation. 
```

docs/proposals/resource-aggregation-proxy/README.md
```diff
@@ -0,0 +1,455 @@
...
+
+![user story - service discovery](user story - service discovery.svg)
+
+Its main workflow can be summaried as following:
+
+1. list/watch pods.
+2. When pod's ip changed, update the corresponding application's routing metadata.
+3. Patch the pod, add a custom annotation to mark the pod as already processed. 
+
+The service discovery controller ignore pod with this annotation, to avoid redundant routing metadata change. Workload redeployment or other operation caused pod to be deleted/replaced will automatically remove the added annotation, and make the pod eligible for next processing.
```
Tingtal, line 84:
```suggestion
The service discovery controller ignores pods with this annotation to avoid redundant routing metadata change. Workload redeployment or other operations that cause pods to be deleted/replaced will automatically remove the added annotation, and make the pods eligible for the next steps of processing. 
```

docs/proposals/resource-aggregation-proxy/README.md
```diff
@@ -0,0 +1,455 @@
...
+
+Client needs to add prefix to the API server url. In our test, Client-go can accept such url during its initiation, and automatically rewrite all request's url with the prefix. Users of client-go do not need to change code in their program. Of course, kubectl can accept the prefixed version of url, too.
+
+In order to accelerate access speed of member cluster resources, we choose to cache resources in proxy.
+
+
+
+![Design-Details-Architecture](Design-Details-Architecture.svg)
+
+The main workflow can be summarized as following:
```
Tingtal, line 125:
```suggestion
The main workflow can be summarized as follows:
```

docs/proposals/resource-aggregation-proxy/README.md
```diff
@@ -0,0 +1,455 @@
...
+
+1. Client calls kubernetes native api, the request is sent to Karmada API server with prefixed url, Karmada API server redirect the request to Resource Aggregation Proxy.
```
Tingtal, line 127:
```suggestion
1. Client calls kubernetes native APIs. The request is sent to Karmada API server with the prefixed URL and redirected to Resource Aggregation Proxy.
```

docs/proposals/resource-aggregation-proxy/README.md
```diff
@@ -0,0 +1,455 @@
...
+2. Resource Aggregation Proxy removes prefix from url, routes the request by resource type:
```
Tingtal, line 128:
```suggestion
2. Resource Aggregation Proxy removes the prefix from the URL and routes the request by resource type:
```

docs/proposals/resource-aggregation-proxy/README.md
```diff
@@ -0,0 +1,455 @@
...
+   1. Resource in Karmada control plane, like deployment: Redirect the request to Karmada API server.
+   2. Resource in member clusters:
+      1. Read operation, like get/list/watch: Read data from cache, merge the results then return.
```
Tingtal, line 131:
```suggestion
      1. Read operation, like get/list/watch: Read data from cache, merge the results, and then return them.
```

docs/proposals/resource-aggregation-proxy/README.md
```diff
@@ -0,0 +1,455 @@
...
+
+## How to Merge Result
+
+- Get request: return first result.
+- List request: concat the results, and return to client as a single list.
+- Watch request: return result one by one.
+
+## Resource Version
+
+Kubernetes use ResourceVersion to uniquely identify every version of resource. We merge all clusters' ResourceVersion into one json struct, then `base64.RawURLEncoding` it, get the final string.
```
Tingtal, line 260:
```suggestion
Kubernetes uses ResourceVersion to uniquely identify every version of a resource. We merge all clusters' ResourceVersion into one JSON struct, `base64.RawURLEncoding` it, and get the final string.
```

docs/proposals/resource-aggregation-proxy/README.md
```diff
@@ -0,0 +1,455 @@
...
+| supported Authentication strategies          | X509 Client Certs                                            | ⭐️ All k8s supported Authentication strategies: X509 Client Certs, Static Token, Bootstrap Token, Service Account Token, OpenID Connect Token, Webhook Token Authentication, Authenticating Proxy |
+| performance                                  | ⭐️ Higher. Much shorter api request path: client => proxy => kube-apiserver | Lower. Longer api request path: client => kube-apiserver => proxy => kube-apiserver |
+| developerment complexity                     | High. In order to support all kubernetes Authentication strategies, we need to include kubernetes Authentication module's code, and update proxy when kubernetes updates. | ⭐️ Low. Only need to write an aggregated API server.          |
+
+## Conclusion
+
+Some users may need to use Authentication strategies other than X509 Client Certs, but implement it in proxy itself will be too complex. So we consider put proxy behind API server will be a better option.
```
Tingtal, line 455:
```suggestion
Some users may need to use Authentication strategies other than X509 Client Certs, but implement it in proxy itself will be too complex. So we consider putting the proxy behind API server a better option.
```

Tingtal (changes requested):
Some touches on the expressions.

ikaven1024:
> Some touches on the expressions.

Thank you for your corrections.

docs/proposals/resource-aggregation-proxy/README.md
```diff
@@ -0,0 +1,455 @@
...
+
+Add a Resource Aggregation Proxy. When the client tries to list pod by label, this proxy will aggregate pods information from all member clusters, and then return to the client as a single list. When the client tries to query workload resources, the proxy will get the information from Karmada API server directly.
+
+## User Stories
+
+### Story 1: Service Discovery
+
+Not everyone uses kubernetes Services for service discovery. We use a private service discovery architecture to accommodate both containerized and traditional applications deployed on physical machines. 
+
+![user story - service discovery](user story - service discovery.svg)
```
XiShanYongYe-Chang, line 76: It doesn't seem to be displayed properly here.

image

ikaven1024, line 76:
```suggestion
![user story - service discovery](user-story-service discovery.svg)
```

ikaven1024: Updated

docs/proposals/resource-aggregation-proxy/README.md
```diff
@@ -0,0 +1,455 @@
...
+
+1\. Client sends request with prefixed url.
+
+```
+GET /apis/proxy.karmada.io/v1alpha1/globalresources/dummy/proxy/api/v1/namespaces/default/pods?resourceVersion=eyJjbHVzdGVyMSI6IjEyMzQiLCJjbHVzdGVyMiI6IjU2NzgifQ&watch=true
+```
+
+2\. Proxy watches cluster1 and cluster2's cache.
+
+3\. When there is change, proxy merges the event, modifies the resourceVersion field, then returns response to client.
```
wuyingjun-lucky, line 365: 1. In the watch situation. The proposal will collect the member clusters events and re-allocate global resouceVersion ？
2. Accutally we now use clusterpedia to do `Get` operation for collect member clusters and use  aa to router clusterpedia and karmada.  And we plan to impl watch by `1`.  I think it is same with your proposal. 
Maybe we can help to join your proposal.

ikaven1024: > re-allocate global resouceVersion ？

you mean the RV is independent from member RV? In fact, we do not want to maintain this RV. The global RV is built with member RVs in a json. See it in `Resource Version` section.

> 2\. Maybe we can help to join your proposal.

Thanks~

docs/proposals/resource-aggregation-proxy/README.md
```diff
@@ -0,0 +1,455 @@
...
+2. Update the corresponding application's routing metadata when the pod IP changes.
+3. Patch the pod and mark it as processed by adding a custom annotation. 
+
+The service discovery controller ignores pods with this annotation to avoid redundant routing metadata change. Workload redeployment or other operations that cause pods to be deleted/replaced will automatically remove the added annotation, and make the pods eligible for the next steps of processing. 
+
+
+
+We need Karmada to provide following functionality:
+
+1. list/watch/patch pod
```
RainbowMango, line 90: Are the Pods restricted to the one that propagated by Karmada? 
In other words, if the user can list pods that they created directly on member clusters?

docs/proposals/resource-aggregation-proxy/README.md
```diff
@@ -0,0 +1,455 @@
...
+- node: get/list/watch/update/patch
+- pod: get/list/watch/update/patch/logs/exec
+
+## Non-Goals
+
+- read/write resources:
+  - service account token
+  - event
+
+- Provide a solution for unique resource names across all member clusters.
```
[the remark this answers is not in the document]
RainbowMango: Yeah, I get it from below section:

> Resource names need to be unique across all member clusters. Otherwise, the proxy will not be able to differentiate which resource needs to be accessed.

But how to assured that the resource name is unique?

docs/proposals/resource-aggregation-proxy/README.md
```diff
@@ -0,0 +1,455 @@
...
+base64.RawURLEncoding({"cluster1":"1234","cluster2":"5678"}) = "eyJjbHVzdGVyMSI6IjEyMzQiLCJjbHVzdGVyMiI6IjU2NzgifQ"
+```
+
+## Example
+
+There are two member clusters: cluster1 and cluster2.
+
+|                 | cluster1         | cluster2         |
+| --------------- | ---------------- | ---------------- |
+| resourceVersion | 1234             | 5678             |
```
RainbowMango, line 272: Is the `1234` the `.metadata.resourceversion` of `Cluster` object?

docs/proposals/resource-aggregation-proxy/README.md
```diff
@@ -0,0 +1,455 @@
...
+
+## How to Merge Result
+
+- Get request: return first result.
+- List request: concat the results, and return to client as a single list.
+- Watch request: return result one by one.
+
+## Resource Version
+
+Kubernetes uses ResourceVersion to uniquely identify every version of a resource. We merge all clusters' ResourceVersion into one JSON struct, `base64.RawURLEncoding` it, and get the final string.
```
RainbowMango, line 260: Why user or client needs the `resource version`?

Kubernetes could guarantee the `ResourceVersion` comparable and incremental, can we? or should we?

docs/proposals/resource-aggregation-proxy/README.md
```diff
@@ -0,0 +1,455 @@
...
+
+
+
+1\. Client sends request with prefixed url.
+
+```
+GET /apis/proxy.karmada.io/v1alpha1/globalresources/dummy/proxy/api/v1/namespaces/default/pods/nginx-2
+```
+
+2\. Proxy queries all caches, and finds out the required Pod is in cluster1's cache.
```
RainbowMango, line 335: I know we assume that each pod is unique, but what if two pods with the same name, what's the behavior here?
returns one by random or panic?

docs/proposals/resource-aggregation-proxy/README.md
```diff
@@ -0,0 +1,455 @@
...
+| supported Authentication strategies          | X509 Client Certs                                            | ⭐️ All k8s supported Authentication strategies: X509 Client Certs, Static Token, Bootstrap Token, Service Account Token, OpenID Connect Token, Webhook Token Authentication, Authenticating Proxy |
+| performance                                  | ⭐️ Higher. Much shorter api request path: client => proxy => kube-apiserver | Lower. Longer api request path: client => kube-apiserver => proxy => kube-apiserver |
+| developerment complexity                     | High. In order to support all kubernetes Authentication strategies, we need to include kubernetes Authentication module's code, and update proxy when kubernetes updates. | ⭐️ Low. Only need to write an aggregated API server.          |
+
+## Conclusion
+
+Some users may need to use Authentication strategies other than X509 Client Certs, but implement it in proxy itself will be too complex. So we consider putting the proxy behind API server a better option.
```
RainbowMango, line 455: > So we consider putting the proxy behind API server a better option.

Does that means some changes are needed from `PaaS components`?

RainbowMango (commented):
Awesome, thanks @ikaven1024 

Generally looks good to me, just some questions.
(I haven't looked into the API part, as well as the component name)

docs/proposals/resource-aggregation-proxy/README.md
ikaven1024: All the pods.

docs/proposals/resource-aggregation-proxy/README.md
ikaven1024: > Why user or client needs the `resource version`?

client use `resource version` to start a watch.

```
kubectl get po -w --v 6
...
https://192.168.50.83:8443/api/v1/namespaces/default/pods?limit=500 200 OK in 13 milliseconds
...
https://192.168.50.83:8443/api/v1/namespaces/default/pods?resourceVersion=13446488&watch=true 200 OK in 1 milliseconds
...
```

> Kubernetes could guarantee the `ResourceVersion` comparable and incremental, can we? or should we?

Kubernetes's `ResourceVersion` actually is the  ResourceVersion of `etcd`, it is comparable and incremental.

docs/proposals/resource-aggregation-proxy/README.md
ikaven1024: Good question. I prefer to return the first one (order by the cluster name). For these reasons:
1. panic is two serious.
2. the returns is stable for client.

docs/proposals/resource-aggregation-proxy/README.md
ikaven1024: The only changes for `PaaS components` is change the url of `apiserver`.
Originally, it may be
```
server=http://10.10.10.10:5443
```

Then  change it to
```
server=http://10.10.10.10:5443/apis/proxy.karmada.io/v1alpha1/globalresources/dummy/proxy
```

docs/proposals/resource-aggregation-proxy/README.md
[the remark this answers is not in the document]
ikaven1024: It is out the scope of this proposal. But i have a rough idea.
We can add a webhook in member clusters, used to validate the resource creations. When validating creations, webhook will search the name from `proxy` component in controller panel.

docs/proposals/resource-aggregation-proxy/README.md
RainbowMango: Got it. It makes sense to me.

docs/proposals/resource-aggregation-proxy/README.md
RainbowMango: In this table, there are two Pods (`nginx-1` and `nginx-2`), how do you know who is the `latest`?

docs/proposals/resource-aggregation-proxy/README.md
RainbowMango: I agree.
I think ensuring each pod is unique is the biggest challenge for this feature. Do you have any idea to solve it?

docs/proposals/resource-aggregation-proxy/README.md
RainbowMango: em, both the API and implementation are similar to karmada-search now. 
Have you ever thought about putting this feature into karmada search?

docs/proposals/resource-aggregation-proxy/README.md
ikaven1024: Indeed, we needn't to care which pod is latest. Response from apiserver will tell it, the field `.metadata.resourceversion` of `PodList`.
```
GET /api/v1/namespaces/test/pods
---
200 OK
Content-Type: application/json

{
  "kind": "PodList",
  "apiVersion": "v1",
  "metadata": {"resourceVersion":"1234"},
  "items": [...]
}
```

RainbowMango: Oh, good to know. I didn't notice the `List` also has the `resourceVersion`. Thanks for the clarification. 
Do you think we should add it to the proposal? I think it would get easy for other people to understand.

docs/proposals/resource-aggregation-proxy/README.md
ikaven1024: I have read the `karmada search` codes, it uses `Informer` to cache resource. It is difficult and complex to implement
[Semantics of resourveVersion](https://kubernetes.io/docs/reference/using-api/api-concepts/#semantics-for-get-and-list), [bookmarks](https://kubernetes.io/docs/reference/using-api/api-concepts/#watch-bookmarks), [resourceVersionMatch](https://kubernetes.io/docs/reference/using-api/api-concepts/#semantics-for-get-and-list)

In `proxy` component, i will use [k8s.io/apiserver/pkg/storage/cacher.Cacher](https://github.com/kubernetes/kubernetes/blob/340555476a4926e10c9ffac8b283cd77ca088c5d/staging/src/k8s.io/apiserver/pkg/storage/cacher/cacher.go#L226-L296), it is native to support all scenes of get/list/watch

RainbowMango: It's not my preference, just thinking about it.

For `proxy` purposes, we also use `informer` to collect resources from member clusters, right? Is it easy for karmada-search using the `k8s.io/apiserver/pkg/storage/cacher.Cacher`?
cc @liys87x here

ikaven1024: > For `proxy` purposes, we also use `informer` to collect resources from member clusters, right? I

No. `cacher.Cacher` also list/watch from member clusters, collecting resources.

If `karmada-search` use it, is shall start `informer` from the `cacher.Cacher`, instead of member cluster apiserver.

@liys87x please evaluate it.

liys87x: I found that `cacher.Cacher` is embedded in restOption and I am learning how it is used.

liys87x: Then I will evaluate it.

docs/proposals/resource-aggregation-proxy/README.md
```diff
@@ -0,0 +1,455 @@
...
+
+# Motivation
+
+Write a Resource Aggregation Proxy for Karmada. This proxy provides native kubernetes API for components outside kubernetes. When the client requests operations (including reads & writes), the proxy will figure out the correct downstream kubernetes API servers (including Karmada API server and member clusters), request all the selected servers, and merge the responses into one before responding to the client. External components can treat Karmada-managed kubernetes clusters as one single cluster.
+
+## Goals
+
+Allow kubernetes API clients (kubectl, client-go, and other clients that use kubernetes APIs) to read/write the following resources:
+
+- workload: create/get/list/watch/update/patch/delete
```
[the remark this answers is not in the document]
ikaven1024: updated

docs/proposals/resource-aggregation-proxy/README.md
```diff
@@ -0,0 +1,455 @@
...
+- node: get/list/watch/update/patch
+- pod: get/list/watch/update/patch/logs/exec
+
+## Non-Goals
+
+- read/write resources:
+  - service account token
+  - event
```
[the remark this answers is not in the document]
ikaven1024: updated

docs/proposals/resource-aggregation-proxy/README.md
ikaven1024: updated

docs/proposals/resource-aggregation-proxy/README.md
teou: so the karmada aggregation proxy will :
1. start informers listening to each member cluster and watch registered resources from member clusters
2. watched resources and theire changes from the member cluster will be stored into some etcd storage
3. start a cacher.Cacher with backend storage pointing to etcd storage above and serve list/watch requests

is my understanding correct?  @ikaven1024

ikaven1024: There is no informers and etcd in proxy. Indeed, the backend storage of `Cacher` is `kube-apiserver` in member clusters. Cacher list & watch from member apiserver, store member resources in memory.

ikaven1024: Store all resources from member cluster, it is a big challenge to etcd. So we don't use etcd in proxy.

teou: > 

thank you for your kind explanation @ikaven1024 
two more question please :  
    if cacher.Cacher is used , we will have n cacher.Cacher object for n member clusters ? 
    when a watch request is served, we will have a aggregation proxy to start watch from each cacher.Cacher object, then merge the events from each cacher.Cacher, modify resourceVersion, finally return values to the user?

ikaven1024: > we will have n cacher.Cacher object for n member clusters

Right



> when a watch request is served, we will have a aggregation proxy to start watch from each cacher.Cacher object, then merge the events from each cacher.Cacher, modify resourceVersion, finally return values to the user?

Right

docs/proposals/resource-aggregation-proxy/README.md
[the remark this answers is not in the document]
RainbowMango: > We can add a webhook in member clusters, used to validate the resource creations. When validating creations, webhook will search the name from proxy component in controller panel.

Personally, I don't think people would like it. a little bit complicated I guess.

ikaven1024:
> > It tells no collisions occur.
> 
> This depends on how many replicas are in the Deployment. No collision would happen between different Deployments.

Deployments create replicasets, and replicasets creates pods. So just guaranteeing no collision between replicasets.

ikaven1024:
I just look into the generator of pod. Deployment create replicaset with a suffix name [genrated by hash of its template](https://github.com/kubernetes/kubernetes/blob/release-1.24/pkg/controller/deployment/sync.go#L189), so it will create replicasets with the same name accross clusters.
Then replicaset creates pods with random suffix name. [The random chars pool size is 27](https://github.com/kubernetes/kubernetes/blob/release-1.24/staging/src/k8s.io/apimachinery/pkg/util/rand/rand.go#L83), and [suffix length is 5](https://github.com/kubernetes/kubernetes/blob/release-1.24/staging/src/k8s.io/apiserver/pkg/storage/names/generate.go#L45).

According to this [article](https://www.cnblogs.com/yinzhengjie/p/12203765.html), we can estimate the collision probability by this program:
```go
func main() {
	var k = math.Pow(27, 5)
	fmt.Println("K =", int64(k))
	fmt.Println("pods\tno_collision_probability")
	for _, n := range []float64{10, 100, 200, 400, 600, 800, 1000, 2000, 4000, 6000, 8000, 10000} {
		fmt.Println(n, "\t", math.Exp(-n*(n-1)/2.0/k))
	}
}
```

It prints the amount of pods, and the probability of no-collision
```
K = 14348907
pods    no_collision_probability
10       0.9999968638775457
100      0.9996550854857799
200      0.998614096037548
400      0.9944540367459195
600      0.9875544919854364
800      0.9779727488347782
1000     0.9657879195731809
2000     0.8699550398230972
4000     0.5726993631642694
6000     0.28529205369204497
8000     0.10754388811853854
10000    0.03067714052002095
```

It looks well when pods is less than 600.

ikaven1024:
The usage of cacher in proxy:
 https://github.com/ikaven1024/karmada/blob/feature-proxy/pkg/proxy/cache/util.go#L213

ikaven1024:
Compare between Cacher and Informer:
|  | Cacher                | Informer |
|--|-----------------------|----------|
| how to sync from datasource | List/Watch            | List/Watch |
| Get/List with resourceVersion | [support semantics of resourceVersion](https://kubernetes.io/docs/reference/using-api/api-concepts/#semantics-for-get-and-list) | equivalent to `resourceVersion="0"` |
| List with resourceVersionMatch | suported | not suported |
| List with labelSelector | suported | suported |
| List with filedSelector | suported | not suported |
| watch/ watch with allowWatchBookmarks | suported | not suported |

ikaven1024:
### About the question of long pod name
The generated pod name is limited to be no more than 63 chars.  @RainbowMango 

```go
const (
	// TODO: make this flexible for non-core resources with alternate naming rules.
	maxNameLength          = 63
	randomLength           = 5
	maxGeneratedNameLength = maxNameLength - randomLength
)

func (simpleNameGenerator) GenerateName(base string) string {
	if len(base) > maxGeneratedNameLength {
		base = base[:maxGeneratedNameLength]
	}
	return fmt.Sprintf("%s%s", base, utilrand.String(randomLength))
}
```
FYI: https://github.com/kubernetes/kubernetes/blob/a3e71b2cefc755bfc83884f00ce11f2b6bdc41c4/staging/src/k8s.io/apiserver/pkg/storage/names/generate.go#L42-L54


### Test 
1. create a statefulSet with long name in member cluster.
```
➜  ~ kubectl get deploy
NAME                                                                                                                                                                                                                                            READY   UP-TO-DATE   AVAILABLE   AGE
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa   3/3     3            3           58m
➜  ~ kubectl get po
NAME                                                              READY   STATUS    RESTARTS   AGE
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa88bqq   1/1     Running   0          58m
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa8fnhs   1/1     Running   0          58m
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaajwckc   1/1     Running   0          58m
```

2. get and edit pod with proxy
```
➜  kubectl get po aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa8fnhs
NAME                                                                    READY   STATUS    RESTARTS   AGE
yjh-2-aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa8fnhs   1/1     Running   0          59m

➜  kubectl edit po aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa8fnhs
pod/yjh-2-aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa8fnhs edited
```

RainbowMango:
> The generated pod name is limited to be no more than 63 chars. 

Great! 
I just tried on my side, we can create a deployment with name length > 63, but the Pods name indeed <= 63.

```
deployment.apps/nginx123456789123456789123456789123456789123456789123456789123456789123456789nginx123456789123456789123456789123456789123456789123456789123456789123456789 created
[yangyong@ecs-d8b6 ~]$ kubectl get pods 
NAME                                                              READY   STATUS    RESTARTS   AGE
nginx12345678912345678912345678912345678912345678912345678n9nr6   1/1     Running   0          8s
nginx12345678912345678912345678912345678912345678912345678ptgkz   1/1     Running   0          8s

```

ikaven1024:
### About separator in name

I prefer to use an invalid char as separator, such as `_`. It exactly split the cluster name and resource name.
@RainbowMango 

### Test

```shell
➜ kubectl get po
NAME                READY   STATUS    RESTARTS   AGE
sts-nginx-0_yjh-1   1/1     Running   0          12m
sts-nginx-1_yjh-1   1/1     Running   0          12m
sts-nginx-2_yjh-1   1/1     Running   0          11m
sts-nginx-0_yjh-2   1/1     Running   0          12m
sts-nginx-1_yjh-2   1/1     Running   0          12m
sts-nginx-2_yjh-2   1/1     Running   0          12m

➜ kubectl get po sts-nginx-0_yjh-1
NAME                READY   STATUS    RESTARTS   AGE
sts-nginx-0_yjh-1   1/1     Running   0          12m

➜ kubectl edit po sts-nginx-0_yjh-1
pod/sts-nginx-0_yjh-1 edited
```

RainbowMango:
> I prefer to use an invalid char as separator, such as _. It exactly split the cluster name and resource name.

Yeah, that's a great idea, since `_` is not a valid DNS character, and it can't be in the name.

ikaven1024:
@RainbowMango `generateName` don't help to solve the conflict of names.

I create a deploy and set both `metadata.generateName` and `spec.template.metadata.generateName`
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx
  generateName: cluster1-nginx
spec:
  replicas: 3
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      generateName: cluster1-nginx
      labels:
        app: nginx
    spec:
      containers:
      - image: nginx:alpine
        name: nginx
```

Then create it, and pods are created without cluster prefix
```
➜ kubectl get po
NAME                     READY   STATUS    RESTARTS   AGE
nginx-547676d98c-btgc7   1/1     Running   0          5s
nginx-547676d98c-nvlsz   1/1     Running   0          5s
nginx-547676d98c-x4ltm   1/1     Running   0          5s
```

----
Looking into the k8s source, replicaSet's name is generated with `deploy.name + hash` 
```go
	newRS := apps.ReplicaSet{
			// Make the name deterministic, to ensure idempotence
			Name:            d.Name + "-" + podTemplateSpecHash,
			Namespace:       d.Namespace,
			...
		}
	}
```
https://github.com/kubernetes/kubernetes/blob/7d8aae51df4f318b71dc3b790dd01d0bcb79c5dd/pkg/controller/deployment/sync.go#L195-L209

Then pod's `GenerateName` is set to be `rs.name-`
```go
func GetPodFromTemplate(template *v1.PodTemplateSpec, parentObject runtime.Object, controllerRef *metav1.OwnerReference) (*v1.Pod, error) {
        ...
	prefix := getPodsPrefix(accessor.GetName())

	pod := &v1.Pod{
			GenerateName: prefix,
                         ...
		},
	}
}

func getPodsPrefix(controllerName string) string {
	// use the dash (if the name isn't too long) to make the pod name a bit prettier
	prefix := fmt.Sprintf("%s-", controllerName)
	if len(validation.ValidatePodName(prefix, true)) != 0 {
		prefix = controllerName
	}
	return prefix
}
```
https://github.com/kubernetes/kubernetes/blob/78318c7a26928bdabe2491cf022cf9e61be32188/pkg/controller/controller_utils.go#L544-L567


So `generateName` we set in yaml  is not used.

RainbowMango:
Well, according to the [GenerateName comments](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L121-L134), seems it will take effect only when `Name` has not been provided. Just echo the words here:

> 	// GenerateName is an optional prefix, used by the server, to generate a unique
	// name ONLY IF the Name field has not been provided.

I tried on my side, with the `cluster1` as the `generatedName`:
```yaml
# cat deployment.yaml 
apiVersion: apps/v1
kind: Deployment
metadata:
  generateName: cluster1
  labels:
    app: nginx
spec:
  replicas: 2
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - image: nginx
        name: nginx
```

And I can see the name of `Deployment`/`ReplicaSet`/`Pod` are all starts with `generateName`(`cluster1`):
```bash
NAME            READY   UP-TO-DATE   AVAILABLE   AGE
cluster15t6xh   2/2     2            2           7s

# kubectl get replicasets.apps 
NAME                       DESIRED   CURRENT   READY   AGE
cluster15t6xh-85b98978db   2         2         2       22s

# kubectl get pods 
NAME                             READY   STATUS    RESTARTS   AGE
cluster15t6xh-85b98978db-smdhh   1/1     Running   0          25s
cluster15t6xh-85b98978db-wrdch   1/1     Running   0          25s
```

ikaven1024:
> `Name` has not been provided

It seems to be a challenge to karmada. As i know, `resourceBinding` name is created by kind+name.
https://github.com/karmada-io/karmada/blob/98ef6281a6731253b68a94551562d1d0c34ab035/pkg/util/names/names.go#L56-L59

ikaven1024:
@RainbowMango updated

RainbowMango (approved):
/lgtm
/approve

PR #2080, opened 2022-06-28
Fix checking if scheduler has a replica assigned to each cluster

AllenZMC:
**What type of PR is this?**
/kind bug

Fix checking if scheduler has a replica assigned to each cluster
Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/util/helper/binding_test.go
+++ b/pkg/util/helper/binding_test.go
@@ -1,0 +1,75 @@
+package helper
+
+import (
+	"testing"
+
+	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
+)
+
+func TestHasScheduledReplica(t *testing.T) {
+	tests := []struct {
+		name           string
+		scheduleResult []workv1alpha2.TargetCluster
+		want           bool
+	}{
+		{
+			name: "all targetCluster have replicas",
+			scheduleResult: []workv1alpha2.TargetCluster{
+				{
+					Name:     "foo",
+					Replicas: 1,
+				},
+				{
+					Name:     "bar",
+					Replicas: 2,
+				},
+			},
+			want: true,
+		},
+		{
+			name: "a targetCluster has replicas",
+			scheduleResult: []workv1alpha2.TargetCluster{
+				{
+					Name:     "foo",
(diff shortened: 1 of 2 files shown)


RainbowMango:
Can you explain how to reproduce this bug?

AllenZMC:
@RainbowMango  There is no such problem yet, but logically this code has loopholes, look at my test case.

XiShanYongYe-Chang (commented):
Hi @AllenZMC, the comment of this method may mislead you. If the number of replicas of a cluster is greater than 0, it indicates that the scheduler has allocated replicas to these clusters, even if the number of replicas of some clusters is 0.

AllenZMC:
@XiShanYongYe-Chang Thank you for the explanation. That is to say, as long as the number of replicas in a cluster is greater than 0, it indicates that the cluster has been scheduled?

XiShanYongYe-Chang:
> @XiShanYongYe-Chang Thanks for the explanation. That is to say, as long as the number of replicas in a cluster is greater than 0, it indicates that the cluster has been scheduled?
> 
> We need to change the description of the comment, it really misleads me.

Yes, you can update the comment and add the ut for the test.

XiShanYongYe-Chang:
Hi @AllenZMC, maybe you need to rebase the master branch and push it again.

AllenZMC:
@XiShanYongYe-Chang  rebased.

XiShanYongYe-Chang (commented):
Thanks!

/lgtm

pkg/util/helper/binding_test.go
```diff
@@ -0,0 +1,75 @@
...
+)
+
+func TestHasScheduledReplica(t *testing.T) {
+	tests := []struct {
+		name           string
+		scheduleResult []workv1alpha2.TargetCluster
+		want           bool
+	}{
+		{
+			name: "assigned replicas for a cluster",
```
RainbowMango, line 16: The name of sub-test should keep unique for distinguishing them from each other.

AllenZMC: Fixed. @RainbowMango

RainbowMango (approved):
/lgtm
/approve

Issue #2105, opened 2022-07-01
performance improvement for karmada-controller-manager to use `DeferredDiscoveryRESTMapper`


likakuli:
**What would you like to be added**:

use `DeferredDiscoveryRESTMapper` instead of `DynamicRESTMapper` to improve karmada-controller-manager performance

**Why is this needed**:
https://github.com/karmada-io/karmada/blob/c1b5aef905f39e56fd48120c9ff4fcc8e0cda637/pkg/detector/detector.go#L772-L773
https://github.com/karmada-io/karmada/blob/c1b5aef905f39e56fd48120c9ff4fcc8e0cda637/pkg/detector/detector.go#L591-L593

`DynamicRESTMapper` will visit cluster every time to get specific gvr by gvk. it's a waste of performance when there are many object but with less gvk.

RainbowMango:
Thanks for pointing it out. 
In addition to `DynamicRESTMapper` and `DeferredDiscoveryRESTMapper`, I just noticed there also other alternatives:
- [DefaultRESTMapper](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/api/meta/restmapper.go#L57)
- [FirstHitRESTMapper](https://github.com/kubernetes/kubernetes/blob/89aaf7c02dd33415ac994cecfddca7c02ed75834/staging/src/k8s.io/apimachinery/pkg/api/meta/firsthit_restmapper.go#L32)
- [MultiRESTMapper](https://github.com/kubernetes/kubernetes/blob/89aaf7c02dd33415ac994cecfddca7c02ed75834/staging/src/k8s.io/apimachinery/pkg/api/meta/multirestmapper.go#L32)
- [PriorityRESTMapper](https://github.com/kubernetes/kubernetes/blob/89aaf7c02dd33415ac994cecfddca7c02ed75834/staging/src/k8s.io/apimachinery/pkg/api/meta/priority.go#L38)
- [ResettableRESTMapper](https://github.com/kubernetes/kubernetes/blob/89aaf7c02dd33415ac994cecfddca7c02ed75834/staging/src/k8s.io/apimachinery/pkg/api/meta/interfaces.go#L140)

I guess you have evaluated them and I wonder why you finally chose `DeferredDiscoveryRESTMapper` .

likakuli:
[...]

PR #2106, opened 2022-07-01
feat: use DeferredDiscoveryRESTMapper instead of DynamicRESTMapper

likakuli:

**What type of PR is this?**
/kind feature

karmada-controller-manager performance improvement when there are many object but with less gvk.

Fixes # https://github.com/karmada-io/karmada/issues/2105

```release-note
NONE
```

### DIFF
--- a/cmd/controller-manager/app/controllermanager.go
+++ b/cmd/controller-manager/app/controllermanager.go
@@ -5,13 +5,19 @@ import (
 	"flag"
 	"net"
 	"strconv"
+	"time"
 
 	"github.com/spf13/cobra"
+	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/discovery"
+	cacheddiscovery "k8s.io/client-go/discovery/cached/memory"
 	"k8s.io/client-go/dynamic"
 	kubeclientset "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/restmapper"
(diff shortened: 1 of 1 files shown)


RainbowMango:
@XiShanYongYe-Chang Could you please look at the failing case:
https://github.com/karmada-io/karmada/runs/7149212797?check_suite_focus=true
Seems not due to this patch.

XiShanYongYe-Chang:
I increased the cluster unjoin waiting time. Hi @likakuli, can you rebase the master branch and push it again.

likakuli:
@XiShanYongYe-Chang Could you please help look at the failing case. Seems not due to this patch. thx

RainbowMango:
re-triggered.
The CI is in an unstable state recently.

likakuli:
I make a benchmark test with three kind of method to get gvr from gvk.

First one is use a global variable, code like below
```
var gvkToGVRMap sync.Map

// GetGroupVersionResource is a helper to map GVK(schema.GroupVersionKind) to GVR(schema.GroupVersionResource).
func GetGroupVersionResource(restMapper meta.RESTMapper, gvk schema.GroupVersionKind) (schema.GroupVersionResource, error) {
	value, ok := gvkToGVRMap.Load(gvk)
	if !ok {
		restMapping, err := restMapper.RESTMapping(gvk.GroupKind(), gvk.Version)
		if err != nil {
			return schema.GroupVersionResource{}, err
		}

		gvkToGVRMap.Store(gvk, restMapping.Resource)
		value = restMapping.Resource
	}

	return value.(schema.GroupVersionResource), nil
}
```

The second is just as now. The third is use the code of this PR.

The final result is that the first one is the fastest. The second and the third is almost the same cost. The first is about 100 times faster than the others.

RainbowMango:
A benchmark test? But how initialize the RESTMapper? Can you share the benchmark?

> The first is about 100 times faster than the others.

Awesome!

PR #2118, opened 2022-07-03
propagate dependencies support propagate pvc

### LINKED ISSUE #1887: Propagate dependencies add support propagate pvc and service
Propagate dependencies add support propagate `pvc` and `service,` and give a option to user configure which resources need to dependend propagate
**Why is this needed**:
Propagate dependencies support propagate `configmap` and `secret` #1684

hanweisen:
Signed-off-by: hanweisen <hanweisen_yewu@cmss.chinamobile.com>

**What type of PR is this?**
/kind feature

propagate dependencies support propagate pvc
Fixes #1887

```release-note
NONE
```

### DIFF
--- a/pkg/resourceinterpreter/defaultinterpreter/dependencies.go
+++ b/pkg/resourceinterpreter/defaultinterpreter/dependencies.go
@@ -117,6 +117,7 @@ func getDependenciesFromPodTemplate(podObj *corev1.Pod) ([]configv1alpha1.Depend
 	dependentConfigMaps := getConfigMapNames(podObj)
 	dependentSecrets := getSecretNames(podObj)
 	dependentSas := getServiceAccountNames(podObj)
+	dependentPVCs := getPVCNames(podObj)
 	var dependentObjectRefs []configv1alpha1.DependentObjectReference
 	for cm := range dependentConfigMaps {
 		dependentObjectRefs = append(dependentObjectRefs, configv1alpha1.DependentObjectReference{
@@ -143,6 +144,16 @@ func getDependenciesFromPodTemplate(podObj *corev1.Pod) ([]configv1alpha1.Depend
 			Name:       sa,
 		})
 	}
+
+	for pvc := range dependentPVCs {
+		dependentObjectRefs = append(dependentObjectRefs, configv1alpha1.DependentObjectReference{
+			APIVersion: "v1",
+			Kind:       "PersistentVolumeClaim",
+			Namespace:  podObj.Namespace,
+			Name:       pvc,
+		})
+	}
+
 	return dependentObjectRefs, nil
 }
 
@@ -169,5 +180,19 @@ func getConfigMapNames(pod *corev1.Pod) sets.String {
 		result.Insert(name)
 		return true
 	})
+	return result
+}
+
+func getPVCNames(pod *corev1.Pod) sets.String {
+	result := sets.NewString()
+	for i := range pod.Spec.Volumes {
+		volume := pod.Spec.Volumes[i]
+		if volume.PersistentVolumeClaim != nil {
+			claimName := volume.PersistentVolumeClaim.ClaimName
+			if len(claimName) != 0 {
+				result.Insert(claimName)
+			}
+		}
+	}
 	return result
 }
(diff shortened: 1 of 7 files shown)


test/e2e/framework/persistentvolumeclaim.go
```diff
@@ -0,0 +1,83 @@
...
+		if apierrors.IsNotFound(err) {
+			return true
+		}
+
+		klog.Errorf("Failed to get pvc(%s/%s) on cluster(%s), err: %v", namespace, name, cluster, err)
+		return false
+	}, pollTimeout, pollInterval).Should(gomega.Equal(true))
+}
+
+// WaitPVCDisappearOnClusters wait service disappear on member clusters until timeout.
```
Tingtal, line 76:
```suggestion
// WaitPVCDisappearOnClusters: Wait for the service to disappear on member clusters until timeout.
```

Not sure whether my suggestion makes sense. It looks like a function description to me. Another thing is, why do we say "wait **service** disappear" in the description while the function name is "Wait**PVC**"? Shouldn't we say "Wait for the **PVC** to disappear"?

test/e2e/framework/persistentvolumeclaim.go
```diff
@@ -0,0 +1,83 @@
...
+func WaitPVCDisappearOnClusters(clusters []string, namespace, name string) {
+	ginkgo.By(fmt.Sprintf("Check if pvc(%s/%s) diappeare on member clusters", namespace, name), func() {
```
Tingtal, line 78:
```suggestion
	ginkgo.By(fmt.Sprintf("Check if pvc(%s/%s) diappears on member clusters", namespace, name), func() {
```

test/helper/resource.go
```diff
@@ -589,6 +589,17 @@ func NewConfigMap(namespace string, name string, data map[string]string) *corev1
 	}
 }
 
+// NewPVC will build a new PersistentVolumeClaim:.
```
Tingtal, line 592:
```suggestion
// NewPVC will build a new PersistentVolumeClaim.
```

Tingtal (changes requested):
Small touches on the language. Pls check whether my suggestions make sense. Thanks.

test/e2e/framework/persistentvolumeclaim.go
hanweisen: Yes, that's an error, `wait for the PVC to disappear` is right

hanweisen: I have to use `WaitPVCDisappearOnClusters` instead of `WaitPVCDisappearOnClusters:`, because use `WaitPVCDisappearOnClusters:` my goland will give me a warning, may this is a convention in `go`

pkg/resourceinterpreter/defaultinterpreter/dependencies.go
```diff
@@ -117,6 +117,7 @@ func getDependenciesFromPodTemplate(podObj *corev1.Pod) ([]configv1alpha1.Depend
 	dependentConfigMaps := getConfigMapNames(podObj)
 	dependentSecrets := getSecretNames(podObj)
 	dependentSas := getServiceAccountNames(podObj)
+	dependPVCs := getPVCNames(podObj)
```
mrlihanbo, line 120: dependentPVCs may be better?

mrlihanbo:
Generally looks good to me. We need to clarify that the PVC will be propagated or migrated automatically，but the data in PV won't be migrated automatically. In other worlds, we just create a new pv in new clusters.

pkg/resourceinterpreter/defaultinterpreter/dependencies.go
hanweisen: yes

hanweisen:
> We need to clarify that the PVC will be propagated or migrated automatically，but the data in PV won't be migrated automatically. In other worlds, we just create a new pv in new clusters.
I think I can add this in user guide, after this pr is merged

RainbowMango (approved):
/approve
/hold
for @XiShanYongYe-Chang confirm

XiShanYongYe-Chang (commented):
/lgtm
/hold cancel

Issue #2110, opened 2022-07-02
the kube controller manager of karmada keep restarting after installed the master helm chart due to liveness probe


brucexyz1234:
**What happened**:
the kube controller manager of karmada keep restarting after installed the master chart due to liveness probe
**What you expected to happen**:
should not keep restating
**How to reproduce it (as minimally and precisely as possible)**:
install the master helm chart 
**Anything else we need to know?**:
try it
**Environment**:
- Karmada version: master
- kubectl-karmada or karmadactl version (the result of `kubectl-karmada version` or `karmadactl version`):
- Others:

calvin0327:
Is there some message with the kube-controller livenessProbe?

calvin0327:
@brucexyz1234 can you support more message about the issue?

brucexyz1234:
![image](https://user-images.githubusercontent.com/26198722/177071046-aaa09435-8318-4baa-bce0-049a46c2e21a.png)

brucexyz1234:
![image](https://user-images.githubusercontent.com/26198722/177071207-54306d63-3bf6-4b50-bbd3-32faa15a48bb.png)

brucexyz1234:
@calvin0327 
fyi

brucexyz1234:
----     ------        ----                    ----                  -------
  Warning  Unhealthy     24m (x4251 over 35h)    kubelet, xxx  Liveness probe failed: Get "https://127.0.0.1:10257/healthz": dial tcp 127.0.0.1:10257: connect: connection refused
  Normal   Pulled        9m29s (x538 over 35h)   kubelet, xxx Container image "xxxxxxx/kube-controller-manager:v1.22.10" already present on machine
  Warning  BackOffStart  4m20s (x6552 over 34h)  kubelet, xxx  the failed container exited with ExitCode: 2

brucexyz1234:
@calvin0327

calvin0327:
I think we should add `Startup probe` work before liveness probe.

calvin0327:
It maybe kube controller manager server is not running really and the liveness probe start worl already.

calvin0327:
<img width="1263" alt="image" src="https://user-images.githubusercontent.com/45745657/177084621-2367a211-0585-4bb0-a1f4-845689e4a38c.png">

Restart 2 time and running normal.

Poor12:
Hello,  I wonder if kube-controller-manager just restarts serveral times finally start or it will never start successfully.

[...]

PR #2121, opened 2022-07-04
chart: add startup probe to kube controller

calvin0327:

**What type of PR is this?**
/kind bug

Kube controller manager server is not running really and the liveness probe start work already.

Fixes # https://github.com/karmada-io/karmada/issues/2110

There are test result:
Before:

Now:
image

```release-note

```

### DIFF
--- a/charts/karmada/templates/kube-controller-manager.yaml
+++ b/charts/karmada/templates/kube-controller-manager.yaml
@@ -73,17 +73,22 @@ spec:
             - --use-service-account-credentials=true
             - --v=5
           image: {{ template "karmada.kubeControllerManager.image" . }}
-          livenessProbe:
-            failureThreshold: 8
+          {{- if .Values.kubeControllerManager.livenessProbe.enabled }}
+          livenessProbe: {{- include "common.tplvalues.render" (dict "value" (omit .Values.kubeControllerManager.livenessProbe "enabled") "context" $) | nindent 12 }}
             httpGet:
               host: 127.0.0.1
               path: /healthz
               port: 10257
               scheme: HTTPS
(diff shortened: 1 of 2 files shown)


calvin0327:
@Poor12 PTAL

Poor12:
I am not sure that the problem is just restart serveral times finally start or it will never start successfully.

calvin0327:
@Poor12 If there is not startup probe, it will running after serveral restart. but the startup probe can prevents liveness probe from working early.

Other components also restart due to service dependencies. we can add initContainer to resolve the problem. I will submit a new PR to deal with it.

Poor12:
Yeah. Two kinds of question: 1. livenessprobe before pod started. I wonder if it's a must-have problem. If restart too often, I prefer to enlarge `initialDelaySeconds`. Cause we don't want to make chart too complex. 2. component dependencies. e.g. some components may depend on Karmada-AA. Sadly, helm seems only support `weight` for chart hooks.

calvin0327:
Thanks for your comments.
Question 1: I don't think add probe will cause our chart to be complex. Expose the probe parameter in chart  value.yaml is a basic feature. And If use kubeadm to install the cluster, `kube-controller` also provides startup and liveness probes.

image

Question 2: we can use [initcontainer](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/) to check the dependecy server is running. like this: https://github.com/clusterpedia-io/clusterpedia/blob/main/charts/templates/clustersynchro-manager-deployment.yaml#L32

calvin0327:
@Poor12 I found the cause of the problem, I will reopen new pr to resolve the issue.

PR #2132, opened 2022-07-05
karmadactl:modify the usage of cluster-namespace

Fish-pro:
**What type of PR is this?**
/kind cleanup

The same parameter（--cluster-namespace） has different usage in different command execution, and the parameter usage in the join command is difficult to understand
Run `karmadactl join -h` to get the usage of --cluster-namespace
```
➜  karmada git:(master) go run cmd/karmadactl/karmadactl.go join -h
Flags:
      --cluster-namespace string    Namespace in the control plane where member secrets cluster are stored. (default "karmada-cluster")
```
Run `karmadactl unjoin -h` to get the usage of --cluster-namespace
```
➜  karmada git:(master) go run cmd/karmadactl/karmadactl.go unjoin -h
Flags:
      --cluster-namespace string    Namespace in the control plane where member cluster are stored. (default "karmada-cluster")
```
Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/karmadactl/get.go
+++ b/pkg/karmadactl/get.go
@@ -90,7 +90,7 @@ func NewCmdGet(karmadaConfig KarmadaConfig, parentCommand string) *cobra.Command
 	cmd.Flags().StringVarP(&o.Namespace, "namespace", "n", "default", "-n=namespace or -n namespace")
 	cmd.Flags().StringVarP(&o.LabelSelector, "labels", "l", "", "-l=label or -l label")
 	cmd.Flags().StringSliceVarP(&o.Clusters, "clusters", "C", []string{}, "-C=member1,member2")
-	cmd.Flags().StringVar(&o.ClusterNamespace, "cluster-namespace", options.DefaultKarmadaClusterNamespace, "Namespace in the control plane where member cluster are stored.")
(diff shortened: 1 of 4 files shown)


lonelyCZ:
Thanks @Fish-pro 

/lgtm

pkg/karmadactl/get.go
```diff
@@ -90,7 +90,7 @@ func NewCmdGet(karmadaConfig KarmadaConfig, parentCommand string) *cobra.Command
 	cmd.Flags().StringVarP(&o.Namespace, "namespace", "n", "default", "-n=namespace or -n namespace")
 	cmd.Flags().StringVarP(&o.LabelSelector, "labels", "l", "", "-l=label or -l label")
 	cmd.Flags().StringSliceVarP(&o.Clusters, "clusters", "C", []string{}, "-C=member1,member2")
-	cmd.Flags().StringVar(&o.ClusterNamespace, "cluster-namespace", options.DefaultKarmadaClusterNamespace, "Namespace in the control plane where member cluster are stored.")
+	cmd.Flags().StringVar(&o.ClusterNamespace, "cluster-namespace", options.DefaultKarmadaClusterNamespace, "Namespace in the control plane where member cluster secret are stored.")
```
RainbowMango, line 93: We have more than one secret, so I think here should use secret`s`.

Fish-pro: Thanks for your guidance. It has been updated

Issue #1703, opened 2022-05-01
interpreter webhook response's patch is nil, workload sync to member  cluster unsuccessfully


CharlesQQ:
**What happened**:
deployment not synced to member cluster, log print following message:
```
E0501 09:24:02.017878      20 common.go:102] failed to revise replica for Deployment/default/nginx in cluster xxx, err is: Internal error occurred: failed calling webhook "interpreter-webhook/deployment.interpreter.webhook": reveived invalid webhook response: webhook returned empty response.patch

Failed to transform resourceBinding(nginx-deployment), Error: Internal error occurred: failed calling webhook "interpreter-webhook/deployment.interpreter.webhook": reveived invalid webhook response: webhook returned empty response.patch
```

**What you expected to happen**:
response'patch is nil, that is expected, workload should be synced successfully

**How to reproduce it (as minimally and precisely as possible)**:
interpreter webhook  type `ReviseReplica`  response object's patch is nil

**Anything else we need to know?**:
response with func verifyResourceInterpreterContextWithPatch, the following code is unnecessary and cause problems
https://github.com/karmada-io/karmada/blob/46dacea8972e061c1b5de108c776b214c7504848/pkg/resourceinterpreter/customizedinterpreter/webhook/resourceinterpretercontext.go#L138-L140

**Environment**:
- Karmada version:
v1.1.2
- kubectl-karmada or karmadactl version (the result of `kubectl-karmada version` or `karmadactl version`):
- Others:

XiShanYongYe-Chang:
Hi @CharlesQQ , when response object's patch is nil, the response object's patchType also needs to be nil. 

In your case, is patchType also nil?

RainbowMango:
> response'patch is nil, that is expected,

Why do you think the `nil patch` is expected? Any more detailed infos?

CharlesQQ:
`In your case, is patchType also nil?`
patchType is JSONPatch, it's similar example, logic enter following code
https://github.com/karmada-io/karmada/blob/772e6fa6f33dc6aa5e803b23e2b1e684dc35c2ca/pkg/webhook/interpreter/response.go#L26-L45

CharlesQQ:
`Why do you think the nil patch is expected?`
original and  current is same, so response'patch is nil is expected
https://github.com/karmada-io/karmada/blob/772e6fa6f33dc6aa5e803b23e2b1e684dc35c2ca/pkg/webhook/interpreter/response.go#L48-L56

[...]

PR #2161, opened 2022-07-10
fix: fix applyPatch when patch and patchType is nil

CharlesQQ:
**What type of PR is this?**
/kind bug

Fixes https://github.com/karmada-io/karmada/issues/1703

```
`karmada-controller-manager`: Fixed return err in case of interpreter webhook returns nil patch and nil patchType.

```

### DIFF
--- a/pkg/resourceinterpreter/customizedinterpreter/customized.go
+++ b/pkg/resourceinterpreter/customizedinterpreter/customized.go
@@ -251,6 +251,10 @@ func (e *CustomizedInterpreter) callHook(ctx context.Context, hook configmanager
 
 // applyPatch uses patchType mode to patch object.
 func applyPatch(object *unstructured.Unstructured, patch []byte, patchType configv1alpha1.PatchType) (*unstructured.Unstructured, error) {
+	if len(patch) == 0 && len(patchType) == 0 {
+		klog.Infof("Skip apply patch for object(%s: %s) as patch and patchType is nil", object.GroupVersionKind().String(), object.GetName())
+		return object, nil
+	}
 	switch patchType {
 	case configv1alpha1.PatchTypeJSONPatch:
 		if len(patch) == 0 {


XiShanYongYe-Chang:
Hi @CharlesQQ, do you mean when `patch` and `patchType` are nil, we shouldn't treat it as an error?

CharlesQQ:
> do you mean when patch and patchType are nil, we shouldn't treat it as an error?

because following code not treat is as an error

https://github.com/karmada-io/karmada/blob/778b2eaa3b300d80b30d22cb598cd09dca0ea84d/pkg/resourceinterpreter/customizedinterpreter/webhook/resourceinterpretercontext.go#L132-L134

https://github.com/karmada-io/karmada/blob/778b2eaa3b300d80b30d22cb598cd09dca0ea84d/pkg/webhook/interpreter/response.go#L26-L45

XiShanYongYe-Chang (commented):
Ok, I agree with you.

/lgtm

XiShanYongYe-Chang:
@CharlesQQ, can you help add the release notes.

The affects conponent should be `karmada-controller-manager`.

CharlesQQ:
> can you help add the release notes?

Is that ok?

XiShanYongYe-Chang:
`The affects conponent should be karmada-controller-manager.`  Sorry, I didn't make it clear that this sentence is only an explanation, not the release-note.

CharlesQQ:
> Sorry, I didn't make it clear that this sentence is only an explanation, not the release-note.

got it!

XiShanYongYe-Chang:
Thanks~

pkg/resourceinterpreter/customizedinterpreter/customized.go
```diff
@@ -251,6 +251,10 @@ func (e *CustomizedInterpreter) callHook(ctx context.Context, hook configmanager
 
 // applyPatch uses patchType mode to patch object.
 func applyPatch(object *unstructured.Unstructured, patch []byte, patchType configv1alpha1.PatchType) (*unstructured.Unstructured, error) {
+	if len(patch) == 0 && len(patchType) == 0 {
+		klog.Infof(" (%s: %s), patch and patchType is nil, give up apply patch ", object.GroupVersionKind().String(), object.GetName())
+		return object, nil
+	}
```
RainbowMango, line 257:
```diff
diff --git a/pkg/resourceinterpreter/customizedinterpreter/customized.go b/pkg/resourceinterpreter/customizedinterpreter/customized.go
index 35af4a5b..05e9aed2 100644
--- a/pkg/resourceinterpreter/customizedinterpreter/customized.go
+++ b/pkg/resourceinterpreter/customizedinterpreter/customized.go
@@ -252,7 +252,7 @@ func (e *CustomizedInterpreter) callHook(ctx context.Context, hook configmanager
 // applyPatch uses patchType mode to patch object.
 func applyPatch(object *unstructured.Unstructured, patch []byte, patchType configv1alpha1.PatchType) (*unstructured.Unstructured, error) {
        if len(patch) == 0 && len(patchType) == 0 {
-               klog.Infof(" (%s: %s), patch and patchType is nil, give up apply patch ", object.GroupVersionKind().String(), object.GetName())
+               klog.Infof("Skip apply patch for object(%s: %s) as patch and patchType is nil", object.GroupVersionKind().String(), object.GetName())
                return object, nil
        }
        switch patchType {

```
I think `skip` would be better than `give up`

CharlesQQ: > I think skip would be better than give up

good suggestion

RainbowMango (approved):
/lgtm
/approve

PR #2163, opened 2022-07-10
improve karmadactl command line description

AllenZMC:
**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**:
Unify the style of command line descriptions
```
➜  ~ kubectl karmada -h     
karmadactl controls a Kubernetes Cluster Federation.

Usage:
  karmadactl [flags]
  karmadactl [command]

Available Commands:
  completion  Generate the autocompletion script for the specified shell
  cordon      Mark cluster as unschedulable
  deinit      Removes Karmada from Kubernetes
  describe    Show details of a specific resource or group of resources in a cluster
  exec        Execute a command in a container in a cluster
  get         Display one or many resources
  help        Help about any command
  init        Install karmada in kubernetes.
  join        Register a cluster to control plane
  logs        Print the logs for a container in a pod in a cluster
  promote     Promote resources from legacy clusters to karmada control plane.
  taint       Update the taints on one or more clusters.
  uncordon    Mark cluster as schedulable
  unjoin      Remove the registration of a cluster from control plane
  version     Print the version information.
```

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #2167, opened 2022-07-11
karmadactl:print long describe by templates like kubectl

Fish-pro:
**What type of PR is this?**

/kind cleanup

The existing uses of short and long, include const, string, var, without uniform formatting rules, the code is hard to read,now fix it

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/karmadactl/get.go
+++ b/pkg/karmadactl/get.go
@@ -59,8 +59,8 @@ var (
 	}
 	eventColumn = metav1.TableColumnDefinition{Name: "EVENT", Type: "string", Format: "", Priority: 0}
 
-	getShort = `Display one or many resources`
-
+	getLong = templates.LongDesc(`
+		Display one or many resources.`)
 	getExample = templates.Examples(`
 		# List all pods in ps output format
 		%[1]s get pods
@@ -93,7 +93,8 @@ func NewCmdGet(karmadaConfig KarmadaConfig, parentCommand string) *cobra.Command
 	o := NewCommandGetOptions(ioStreams)
 	cmd := &cobra.Command{
 		Use:          "get [NAME | -l label | -n namespace]  [flags]",
-		Short:        getShort,
+		Short:        `Display one or many resources`,
+		Long:         getLong,
 		SilenceUsage: true,
 		Example:      fmt.Sprintf(getExample, parentCommand),
 		RunE: func(cmd *cobra.Command, args []string) error {
--- a/pkg/karmadactl/cmdinit/cmdinit.go
+++ b/pkg/karmadactl/cmdinit/cmdinit.go
@@ -12,12 +12,9 @@ import (
 	"github.com/karmada-io/karmada/pkg/version"
 )
 
-const (
-	initShort = `Install karmada in kubernetes`
-	initLong  = `Install karmada in kubernetes.`
-)
-
 var (
+	initLong = templates.LongDesc(`
+		Install karmada in kubernetes.`)
 	initExamples = templates.Examples(`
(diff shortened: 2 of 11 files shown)


pkg/karmadactl/get.go
```diff
@@ -58,6 +58,7 @@ var (
 	eventColumn = metav1.TableColumnDefinition{Name: "EVENT", Type: "string", Format: "", Priority: 0}
 
 	getShort = `Display one or many resources`
+	getLog   = `Display one or many resources.`
```
lonelyCZ, line 61: `getLog` --> `getLong`

prodanlabs:
Part of the job seems to conflict with #2163

pkg/karmadactl/get.go
Fish-pro: @lonelyCZ Thank you for pointing out the error, it has been modified

Fish-pro:
> #2163

@prodanlabs @lonelyCZ Oh, I just see that，this pr changes are more than #2163, I will deal with the conflict after #2163 merge

lonelyCZ:
Can you describe the advantages of introducing `templates.LongDesc`？

Fish-pro:
> Can you describe the advantages of introducing `templates.LongDesc`？

@lonelyCZ I found that some modules already introduced templates.LongDesc, so this pr holds all the short long formats together. I understand that as Karmadactl gets better, longer command descriptions are necessary. In the meantime, I think Karmadactl should have its own format for subcommands in the future, what's your opinion?
for example:
```go
// Long command description and examples.
var (
  mineLong = templates.LongDesc(`
    mine which is described here
    with lots of details.`)

  mineExample = templates.Examples(`
    # Run my command's first action
    %[1]s mine first_action

    # Run my command's second action on latest stuff
    %[1]s mine second_action --flag`)
)

// MineOptions contains all the options for running the mine cli command.
type MineOptions struct {
  mineLatest bool

  genericclioptions.IOStreams
}

// NewCmdMine implements the kubectl mine command.
func NewCmdMine(karmadaConfig KarmadaConfig, parentCommand string, streams genericclioptions.IOStreams) *cobra.Command {
  opts := &MineOptions{
    IOStreams: streams,
  }

  cmd := &cobra.Command{
    Use:     "mine",
    Short:   "Run my command",
    Long:    mineLong,
    Example: fmt.Sprintf(mineExample, parentCommand),
    Run: func(cmd *cobra.Command, args []string) {
      if err := opts.Complete(karmadaConfig, cmd, args); err != nil {
        return err
      }
      if err := opts.Validate(cmd, args); err != nil {
        return err
      }
      if err := opts.Run(); err != nil {
        return err
      }
    },
  }

  cmd.Flags().BoolVar(&options.mineLatest, "latest", false, "Use latest stuff")
  return cmd
}

// Complete completes all the required options for mine.
func (o *MineOptions) Complete(f *cmdutil.Factory, cmd *cobra.Command, args []string, out io.Writer) error {
  return nil
}

// Validate validates all the required options for mine.
func (o MineOptions) Validate() error {
  return nil
}

// Run implements all the necessary functionality for mine.
func (o MineOptions) Run() error {
  return nil
}
```

lonelyCZ:
Because there are some commands that have a short `Long describe`, introducing `templates.LongDesc` seems not necessary. I tend to use it when we have a real Long description.

Fish-pro:
> Because there are some commands that have a short `Long describe`, introducing `templates.LongDesc` seems not necessary. I tend to use it when we have a real Long description.

@lonelyCZ Ok, so now short descriptions, some are defined by var,some passed in as string arguments, does that need to be uniform?

PR #2168, opened 2022-07-11
feature: add name length check for fedresourcequota

likakuli:

**What type of PR is this?**
/kind feature

FedResourceQuota is a user-facing resource and its name is used in work's label. So the length should be no more than 63 characters like other resource. e.g. `PropagationPolicy`

Fixes #

```release-note
`karmada-webhook`: The length of FederatedResourceQuota now has been restricted to no more than 63 chars.
```

### DIFF
--- a/pkg/webhook/federatedresourcequota/validating.go
+++ b/pkg/webhook/federatedresourcequota/validating.go
@@ -18,11 +18,13 @@ package federatedresourcequota
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"strings"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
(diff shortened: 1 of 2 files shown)


pkg/webhook/federatedresourcequota/validating_test.go
```diff
@@ -156,3 +158,30 @@ func Test_validateOverallAndAssignments(t *testing.T) {
 		})
 	}
 }
+
+func Test_validateFederatedResourceQuotaName(t *testing.T) {
+	invalidName := "test-name-more-than-63-characters-test-name-more-than-63-characters-test-name-more-than-63-characters-test-name-more-than-63-characters"
```
carlory, line 163: use `strings.Repeat("a", 64)` instead

likakuli: done

XiShanYongYe-Chang (commented):
I think it makes sense to add this validation. However, it may be not a bug fix but only a verification enhancement.

likakuli:
> I think it makes sense to add this validation. However, it may be not a bug fix but only a verification enhancement.

OK，let me update the kind.

XiShanYongYe-Chang:
This failing e2e case will be fixed by #2177

RainbowMango:
> FedResourceQuota is a user-facing resource and its name is used in work's label. 

I'm afraid this is the wrong decision that directly takes `name` as the label.
We already have lessons learned about that. We should use the hash value, just like what we did for ResourceBinding.

By the way, restricting the name length definitely a cheaper way, but we should avoid doing that as we don't know exactly the use case and should not make the assumption that the name length never exceeds 63.

RainbowMango:
Hi @likakuli Long time no see, hope you are doing well.
As I mentioned earlier, using resource names as labels is not a recommended practice. However, this approach was once widely used in Karmada. Fortunately, we have fixed most of these issues, such as replacing it with the UUID of PropagationPolicy. Unfortunately, FederatedResourceQuota still is an exception.

Currently, we are enhancing the FederatedResourceQuota, see #6350, For now, I think we can adopt this `cheaper` solution to prevent people from doing wrong and fix it later.

RainbowMango (commented):
/assign

pkg/webhook/federatedresourcequota/validating.go
```diff
@@ -164,3 +173,13 @@ func validateResourceList(resourceList corev1.ResourceList, fld *field.Path) fie
 
 	return errs
 }
+
+func validateFederatedResourceQuotaName(name string, fld *field.Path) field.ErrorList {
+	errs := field.ErrorList{}
+
+	if len(name) > labelValueMaxLength {
```
RainbowMango, line 180: Do you want to use this const `DNS1123LabelMaxLength`?
https://github.com/kubernetes/kubernetes/blob/8e3adc4df64d5b382c8916610313ce25e0df8e28/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L183

likakuli: already update to use this parameter. the above pkg already have been imported.

PR #2098, opened 2022-06-30
use kubectl logs to implement karmadactl logs

carlory:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

use kubectl logs to implement karmadactl logs in order to improve code reuse

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #2181, opened 2022-07-13
remove unused lifted codes

carlory:
**What type of PR is this?**

/kind feature

After #2098 is merged, they are not being used.

```release-note
NONE
```


RainbowMango:
I wonder how many unused lifted codes left? Can we remove it along with the PR that leads them to unused?

carlory:
@RainbowMango I'll check it later.

carlory:
@RainbowMango 

When I submitted the PR, I didn't notice it used lifted code before. When I was familiar with the code, I found out that there was a hack/update-lifted script, and then I realized this problem. I will avoid this situation in the future. I have checked the lifted The remaining files under the file are in use.

RainbowMango:
Thanks @carlory, 
/lgtm
/approve

PR #2182, opened 2022-07-13
Update util.sh

ihuibin:
**What type of PR is this?**
/kind bug

Signed-off-by: bin.hui <bin.hui@daocloud.io>


RainbowMango:
/close
Due to conflict.

PR #1478, opened 2022-03-12
Enable `karmadactl get` command to leverage aggregated API

lonelyCZ:
Signed-off-by: lonelyCZ <531187475@qq.com>

**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
step1: Use `cluster/proxy` to access resources of member clusters.

step2: Authorize `system:admin` to use `cluster/proxy` when join a member cluster to control plane.

When users don't have right to use `cluster/proxy` resource:
```
[root@master67 karmada]# ./karmadactl get pod
Error: users "system:admin" is forbidden: User "system:serviceaccount:karmada-cluster:karmada-impersonator" cannot impersonate resource "users" in API group "" at the cluster scope
```

**Which issue(s) this PR fixes**:
Fixes #1329 

**Special notes for your reviewer**:
There are some key points needing to resolve.

**Does this PR introduce a user-facing change?**:

```release-note
Enable `karmadactl get` command to leverage aggregated API
```

PR #2190, opened 2022-07-14
remove unused flag for karmadactl get

carlory:
**What type of PR is this?**

/kind bug

After #1478 is merged,  the `--cluster-namespace` flag should be removed.

```release-note
`karmadactl`: remove the `--cluster-namespace` flag for `get` command.
```


lonelyCZ:
Nice finding, thanks!

/lgtm

RainbowMango:
> After https://github.com/karmada-io/karmada/pull/1478 is merged, the --cluster-namespace flag should be removed.

why? I didn't get it, can you give a more detail description. 
Does that mean `--namespace` would be enough?

lonelyCZ:
> why? I didn't get it, can you give a more detail description.

Now, `karmadactl get` directly access resources of member clusters by `aggregated-apiserver`, not need to read member clusters auth secret stored in`cluster-namespace`, so it doesn't need `-cluster-namespace` flag.

RainbowMango:
oh, right.  Thanks for explaining. 
/approve

PR #2196, opened 2022-07-15
Use helm to install karmada-agent

Poor12:

**What type of PR is this?**
/kind cleanup

Unify the config between helm chart and artifacts，and we can use helm to generate yaml of karmada-agent，to reduce maintenance of artifacts

Fixes #

None


RainbowMango:
Just thought a question. What would I do if I want to check the yamls we are using in CI?

Poor12:
> Just thought a question. What would I do if I want to check the yamls we are using in CI?

If we want to see yamls, we can output the yamls in the scripts by echo it.  Or output the yamls to a temp dir.

PR #2207, opened 2022-07-17
Check file path before Initialize crd

my-git9:
**What type of PR is this?**
/kind bug

1.When  we run `karmadactl init --kube-image-registry=registry.cn-hangzhou.aliyuncs.com/google_containers --crds /root/crds.tar.gz`
2.If the crds dir do not has patches or bases, like
```
$ ls crds
crd.yaml  net.yaml
```
3. The output:
image

4. What do we expect:
Command output file does not exist, and exit

Fixes #

The test result:
image

```release-note
Check if the file path exists before Initialize crd.
```

### DIFF
--- a/pkg/karmadactl/cmdinit/karmada/deploy.go
+++ b/pkg/karmadactl/cmdinit/karmada/deploy.go
@@ -64,26 +64,8 @@ func InitKarmadaResources(dir, caBase64, systemNamespace string) error {
 	}
 
 	// Initialize crd
-	basesCRDsFiles := utils.ListFiles(dir + "/crds/bases")
-	klog.Infof("Initialize karmada bases crd resource `%s/crds/bases`", dir)
-	for _, v := range basesCRDsFiles {
-		if path.Ext(v) != ".yaml" {
-			continue
-		}
-		if err := createCRDs(crdClient, v); err != nil {
-			return err
-		}
-	}
-
-	patchesCRDsFiles := utils.ListFiles(dir + "/crds/patches")
-	klog.Infof("Initialize karmada patches crd resource `%s/crds/patches`", dir)
-	for _, v := range patchesCRDsFiles {
-		if path.Ext(v) != ".yaml" {
-			continue
-		}
-		if err := patchCRDs(crdClient, caBase64, v); err != nil {
-			return err
-		}
+	if err = crdInitialize(crdClient, dir, caBase64); err != nil {
+		klog.Exitln(err)
 	}
 
 	// create webhook configuration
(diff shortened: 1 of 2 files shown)


RainbowMango:
Sorry for letting this sit.
I think this is a reasonable improvement.  @my-git9 would you like to rebase and re-open this PR? You can just assign it to me.

my-git9:
> Sorry for letting this sit. I think this is a reasonable improvement. @my-git9 would you like to rebase and re-open this PR? You can just assign it to me.

Sure, thanks

RainbowMango (commented):
I'll try to get back on this by this week.
@chaosi-zju Would you like to take a look?

pkg/karmadactl/cmdinit/utils/util.go
```diff
@@ -126,6 +126,18 @@ func ioCopyN(outFile *os.File, tr *tar.Reader) error {
...
+// PathExists check if the file path exists
+func PathExists(path string) error {
+	_, err := os.Stat(path)
+	if err == nil {
+		return nil
+	}
+	if os.IsNotExist(err) {
+		return err
+	}
+	return nil
```
chaosi-zju, line 138: Hello, I wanna why if `err != nil && !os.IsNotExist(err)`, we not return the err but return nil?

how about like this:

```
func PathExist(path string) error {
    _, err := os.Stat(path)
    if err != nil && !os.IsExist(err) {
        return err
    }
    return nil
}
```

pkg/karmadactl/cmdinit/karmada/deploy.go
```diff
@@ -178,6 +160,38 @@ func createExtralResources(clientSet *kubernetes.Clientset, dir string) error {
 	return nil
 }
 
+// crdInitialize initialize crd
+func crdInitialize(crdClient *clientset.Clientset, dir, caBase64 string) error {
+	if err := utils.PathExists(dir + "/crds/bases"); err != nil {
+		return err
+	}
+	basesCRDsFiles := utils.ListFiles(dir + "/crds/bases")
```
chaosi-zju, line 168: Hi, have you tested if this dir is just a empty dir, if there is any other unexpected failure?

chaosi-zju (commented):
Hello, thank you for your important commit! Review is ongoing, and my viewpoint just for reference only.

RainbowMango (commented):
Just looked into it.
The root cause seems that the [filepath.Walk](https://github.com/karmada-io/karmada/blob/ef87bd2264f1e9ec7354579794527221bec07911/pkg/karmadactl/cmdinit/utils/util.go#L132) might arise an error(like `no such file or directory`) when visiting the path(in this case it is the non-exist `/crds/bases`), this error should be handled but accidentally be ignored at here:
https://github.com/karmada-io/karmada/blob/ef87bd2264f1e9ec7354579794527221bec07911/pkg/karmadactl/cmdinit/utils/util.go#L133.

Currently, we lack an efficient way to validate a specified `crd.tar.gz`, this approach just focuses on checking if `crds/base` and `crds/patches` exist or not,  it's a little bit restrictive. E.g. what if these paths exist but don't have the expected files? (I didn't test it, just a guess, it will make things worse, because it may mask the issue and allow the entire installation process to continue, giving the user a false impression of successful installation, which I mean it is worse.)

So, my opinion is either to establish a comprehensive validation mechanism for the `crds.tar.gz` based on this issue or temporarily postpone this issue.

my-git9:
Yes, it seems that this PR has a lot of limitations

PR #2209, opened 2022-07-17
improve waitAPIServiceReady func

carlory:

**What type of PR is this?**

/kind cleanup

```release-note
NONE
```


RainbowMango (commented):
Can you remind me what's the benefit?

carlory:
@RainbowMango 

The logic of checking `err != nil` is redundant, we can call and return it directly without doing anything

RainbowMango:
I think it's a kind of code style. Both are OK to me.
Thanks for pointing it out, but we shouldn't focus on this.
/close

PR #2213, opened 2022-07-18
working with istio on non flat network when your cluster don't have LB

### LINKED ISSUE #2212: working with istio on non flat network when your cluster don't have LB
working with istio on non flat network when your cluster don't have LB
**Why is this needed**:
Most of clusters don't have LB.

suwliang3:
Signed-off-by: suwanliang_yewu <suwanliang_yewu@cmss.chinamobile.com>

**What type of PR is this?**
/kind feature

Most of cluster don't have LB.
Fixes #2212

```release-note
NONE
```

### DIFF
--- a/docs/working-with-istio-on-non-flat-network.md
+++ b/docs/working-with-istio-on-non-flat-network.md
@@ -11,9 +11,9 @@ not direct connectivity between the pods in all clusters.
 
 ***
 The reason for deploying `istiod` on the `member1` is that `kiali` needs to be deployed on the same cluster as `istiod`
-. If `istiod` and `kiali` are deployed on the `karmada-host`,`kiali` will not find the namespace created by `karmada`. It
-cannot implement the function of service topology for application deployed by `karmada`. I will continue to provide a new
-solution later that deploys `istiod` on the `karmada-host`.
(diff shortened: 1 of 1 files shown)


suwliang3:
@RainbowMango Can you approve to run the workflows?

suwliang3:
> IMHO, it's not a good practice to use `NodePort` in prod env; for dev env, it's easy to install MetaLB(all tests in istio repo run in this way) or something else to provider a public IP.

You're right. I'm just providing a way to use nodePort. Of course you can use MetaLB for dev env.

suwliang3:
> > IMHO, it's not a good practice to use `NodePort` in prod env; for dev env, it's easy to install MetaLB(all tests in istio repo run in this way) or something else to provider a public IP.
> 
> You're right. I'm just providing a way to use nodePort. Of course you can use MetaLB for dev env.

@XiShanYongYe-Chang @zirain What do you think ?

RainbowMango:
> IMHO, it's not a good practice to use NodePort in prod env; for dev env, it's easy to install MetaLB(all tests in istio repo run in this way) or something else to provider a public IP.

I guess @suwliang3 has concerns about the NodePort way.
Given we are going to move the documents to `karmada-io/website` repo, so I'm going to close it for now. 
Thanks all the same.
/close

PR #2214, opened 2022-07-18
use the existing utilerrors.AggregateGoroutines func to collect each cluster's replicas concurrently

carlory:
**What type of PR is this?**

/kind cleanup

Improve code reuse

```release-note
NONE
```


pkg/estimator/client/accurate.go
```diff
@@ -127,22 +127,18 @@ func getClusterReplicasConcurrently(parentCtx context.Context, clusters []string
...
-	var wg sync.WaitGroup
-	errChan := make(chan error, len(clusters))
-	for i := range clusters {
-		wg.Add(1)
-		go func(idx int, cluster string) {
-			defer wg.Done()
+	clusterReplicas := make([]workv1alpha2.TargetCluster, len(clusters))
+	funcs := make([]func() error, len(clusters))
+	for i, cluster := range clusters {
+		i, cluster := i, cluster
```
Garrybest, line 133: I'd like to change the name in order to prevent confusion.

Garrybest, line 133: I'd like to change the name in order to prevent confusion.

Garrybest (commented):
Generally LGTM.

pkg/estimator/client/accurate.go
Garrybest: Rename the variables `i, cluster` in inner loop may be better.

carlory: @Garrybest updated

RainbowMango (approved):
/approve

PR #2230, opened 2022-07-21
Reduce the number of reconciliations on update in detector

zhuwint:
**What type of PR is this?**

/kind feature

Reduce the number of reconciliations on update in detector

Fixes #

Test detector using the deployment and propagationpolicy in {repo}/samples/nginx, and change the parmaters as the follow steps:
* create nginx-propagation with weightPreference member1:1, menber2:2
* create nginx-deployment with replicas 1
* change nginx-deployment replicas to 3
* change nginx-propagation weightPreference to member1:2, member2:1
* change nginx-deployment replicas to 5
* change nginx-propagation weightPreference to member1:2, member2:3
* delete nginx-deployment
* delete nginx-propagation

the count number of reconciliations and interceptiosn as follows: 

Before:
* Reconciling: 18

image

Now:
* Reconciling: 4
* Intercept: 14

image
image

None

### DIFF
--- a/pkg/detector/detector.go
+++ b/pkg/detector/detector.go
@@ -320,6 +320,23 @@ func (d *ResourceDetector) OnAdd(obj interface{}) {
 
 // OnUpdate handles object update event and push the object to queue.
 func (d *ResourceDetector) OnUpdate(oldObj, newObj interface{}) {
+	unstructuredOldObj, err := helper.ToUnstructured(oldObj)
+	if err != nil {
+		klog.Errorf("Failed to transform oldObj, error: %v", err)
+		return
+	}
+
+	unstructuredNewObj, err := helper.ToUnstructured(newObj)
+	if err != nil {
+		klog.Errorf("Failed to transform newObj, error: %v", err)
+		return
+	}
+
+	if !SpecificationChanged(unstructuredOldObj, unstructuredNewObj) {
+		klog.V(4).Infof("Ignore update event of object (kind=%s, %s/%s) as specification no change", unstructuredOldObj.GetKind(), unstructuredOldObj.GetNamespace(), unstructuredOldObj.GetName())
+		return
+	}
+
 	d.OnAdd(newObj)
 }
 
--- a/pkg/detector/eventfilter.go
+++ b/pkg/detector/eventfilter.go
@@ -1,0 +1,23 @@
+package detector
+
+import (
+	"reflect"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+// SpecificationChanged check if the specification of the given object change or not
+func SpecificationChanged(oldObj, newObj *unstructured.Unstructured) bool {
+	oldBackup := oldObj.DeepCopy()
+	newBackup := newObj.DeepCopy()
+
+	// Remove the status and some system defined mutable fields in metadata, including managedFields and resourceVersion.
+	// Refer to https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/object-meta/#ObjectMeta for more details.
+	removedFields := [][]string{{"status"}, {"metadata", "managedFields"}, {"metadata", "resourceVersion"}}
+	for _, r := range removedFields {
+		unstructured.RemoveNestedField(oldBackup.Object, r...)
+		unstructured.RemoveNestedField(newBackup.Object, r...)
+	}
+
+	return !reflect.DeepEqual(oldBackup, newBackup)
+}
--- a/pkg/detector/eventfilter_test.go
+++ b/pkg/detector/eventfilter_test.go
@@ -1,0 +1,205 @@
+package detector
+
+import (
+	"testing"
+
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog/v2"
+	"k8s.io/utils/pointer"
+
+	workloadv1alpha1 "github.com/karmada-io/karmada/examples/customresourceinterpreter/apis/workload/v1alpha1"
+	"github.com/karmada-io/karmada/pkg/util/helper"
+)
+
+func TestSpecificationChanged(t *testing.T) {
+	tests := []struct {
+		name       string
+		oldObj     interface{}
+		newObj     interface{}
+		wantChange bool
+	}{
+		{
+			name: "Only user defined fields changed(Deployment)",
+			oldObj: &appsv1.Deployment{
+				ObjectMeta: metav1.ObjectMeta{
+					ResourceVersion: "123456",
+					ManagedFields:   []metav1.ManagedFieldsEntry{},
+				},
+				Spec: appsv1.DeploymentSpec{
+					Selector: &metav1.LabelSelector{
+						MatchLabels: map[string]string{"app": "nginx_v1"},
+					},
+				},
+				Status: appsv1.DeploymentStatus{},
+			},
+			newObj: &appsv1.Deployment{
+				ObjectMeta: metav1.ObjectMeta{
+					ResourceVersion: "123456",
+					ManagedFields:   []metav1.ManagedFieldsEntry{},
+				},
+				Spec: appsv1.DeploymentSpec{
+					Selector: &metav1.LabelSelector{
+						MatchLabels: map[string]string{"app": "nginx_v2"},
+					},
+				},
+				Status: appsv1.DeploymentStatus{},
+			},
+			wantChange: true,
+		},
+		{
+			name: "Only status fields changed(Deployment)",
+			oldObj: &appsv1.Deployment{
+				ObjectMeta: metav1.ObjectMeta{
+					ResourceVersion: "123456",
+					ManagedFields:   []metav1.ManagedFieldsEntry{{}},
+				},
+				Spec:   appsv1.DeploymentSpec{},
+				Status: appsv1.DeploymentStatus{Replicas: 3, UpdatedReplicas: 3, ReadyReplicas: 0},
+			},
+			newObj: &appsv1.Deployment{
+				ObjectMeta: metav1.ObjectMeta{
+					ResourceVersion: "123498",
+					ManagedFields:   []metav1.ManagedFieldsEntry{{}, {}},
+				},
+				Spec:   appsv1.DeploymentSpec{},
+				Status: appsv1.DeploymentStatus{Replicas: 3, UpdatedReplicas: 1, ReadyReplicas: 2},
+			},
+			wantChange: false,
+		},
+		{
+			name: "Both user defined fields and status fields changed(Deployment)",
+			oldObj: &appsv1.Deployment{
+				ObjectMeta: metav1.ObjectMeta{
+					ResourceVersion: "123456",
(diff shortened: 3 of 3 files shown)


pkg/detector/detector_test.go
```diff
@@ -0,0 +1,301 @@
...
+	tests := []struct {
+		name    string
+		oldObj  interface{}
+		newObj  interface{}
+		wantAdd bool
+	}{
+		{
+			name: "Only user defined fields changed(Deployment)",
+			oldObj: &appsv1.Deployment{
+				TypeMeta: metav1.TypeMeta{Kind: "Deployment", APIVersion: "v1"},
```
Poor12, line 46: A little advise.  We have declared the struct type. I think we can omit the `TypeMeta` field.

zhuwint: > A little advise. We have declared the struct type. I think we can omit the `TypeMeta` field.

Thanks for point out, it's modified in my new commit

pkg/detector/detector.go
```diff
@@ -320,6 +321,36 @@ func (d *ResourceDetector) OnAdd(obj interface{}) {
 
 // OnUpdate handles object update event and push the object to queue.
 func (d *ResourceDetector) OnUpdate(oldObj, newObj interface{}) {
+	unstructuredOldObj, err := helper.ToUnstructured(oldObj)
+	if err != nil {
+		klog.Errorf("Failed to transform oldObj, error: %v", err)
+		d.OnAdd(newObj)
```
XiShanYongYe-Chang, line 327: Have you ever occurs a converting error?

XiShanYongYe-Chang (commented):
Good job, it's pretty obvious.

pkg/detector/detector.go
zhuwint: > Have you ever occurs a converting error?

No, but I referred to pkg\clusterdiscovery\clusterapi.go:153.

XiShanYongYe-Chang: If so, when we occur err, maybe we need not to call `OnAdd` method.

zhuwint: Ok, I was thinking about not changing the original processing logic when the transform fails because it might be processed elsewhere, but if it's not necessnary, I'll remove it.

XiShanYongYe-Chang:
/lgtm
Ask @Poor12 @RainbowMango for helping review.

Poor12:
Thanks.
/lgtm

RainbowMango:
@likakuli This is the issue that you mentioned before.

likakuli:
This pr can reduce the number of reconciliations on update in detector when `karmada-controller-manager` is running. 

When we restart `karmada-controller-manager`, all object will be reconciled once. For object already propagated to member cluster and not changed after propagated, the reconciliation should be ignored too.

RainbowMango:
> This pr can reduce the number of reconciliations on update in detector when karmada-controller-manager is running.

Yes, exactly.

> When we restart karmada-controller-manager, all object will be reconciled once. For object already propagated to member cluster and not changed after propagated, the reconciliation should be ignored too.

Is there a bad effect for the redundant reconciliation in the secnario of restart?
And any idea about it? I don't know how to improve it(not sure if we need to, either). As far as I know, every controller would reconcile once during restarting.

likakuli:
> Is there a bad effect for the redundant reconciliation in the secnario of restart?

The restart time is too long when there are to many resources exist. 


> And any idea about it? I don't know how to improve it(not sure if we need to, either). As far as I know, every controller would reconcile once during restarting.
We add hash to original object label and add a filter handler to check whether the actual hash and the old hash save in labels is same in our internal version.

RainbowMango:
> The restart time is too long when there are to many resources exist.

How long? By the way, have you configured `--concurrent-resource-template-syncs`(The number of resource templates that are allowed to sync concurrently.) which defaults to `5`.

> We add hash to original object label and add a filter handler to check whether the actual hash and the old hash save in labels is same in our internal version.

I'm confused because in the start progress, the event we received actually the `Add` event, not the `Update` event.

likakuli:
It cost more than 10 minute for our case when restart. 

We implemented this by using custom `FilteringResourceEventHandler`

XiShanYongYe-Chang:
If labels need to be added, does that mean that labels need to be added for objects that the controller cares about, such as binding, work, and something else? Are there too many objects to be processed?

pkg/detector/detector.go
```diff
@@ -320,6 +321,34 @@ func (d *ResourceDetector) OnAdd(obj interface{}) {
...
+	// Remove the status and some system defined mutable fields in metadata, including managedFields and resourceVersion.
+	// Refer to https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/object-meta/#ObjectMeta for more details.
+	removedFields := [][]string{{"status"}, {"metadata", "managedFields"}, {"metadata", "resourceVersion"}}
+	for _, r := range removedFields {
+		unstructured.RemoveNestedField(oldBackup.Object, r...)
+		unstructured.RemoveNestedField(newBackup.Object, r...)
+	}
+
+	if reflect.DeepEqual(oldBackup, newBackup) {
+		klog.Infof("Object (kind=%s, %s/%s) no need to reconcile, intercept the event adding to worker queue", oldBackup.GetKind(), oldBackup.GetNamespace(), oldBackup.GetName())
```
RainbowMango, line 348:
```suggestion
		klog.V(4).Infof("Ignore update event of object (kind=%s, %s/%s) as specification no change", oldBackup.GetKind(), oldBackup.GetNamespace(), oldBackup.GetName())
```

RainbowMango, line 348: I suggest creating a new file(like `eventfilter.go`) and putting this logic into a new function, like
```golang
func SpecificationChanged(oldObj, newObj *unstructured.Unstructured) bool
```
After this, you don't need to mock anymore, as the function would be easier to test.

RainbowMango (commented):
Generally looks good.

RainbowMango:
> We add hash to original object label and add a filter handler to check whether the actual hash and the old hash save in labels is same in our internal version.

Let me try to understand this.
You set a hash to the object label, something like:
```
object.metadata.labels = hash(object) 
```
and in the `FilteringResourceEventHandler`:
- get the hash from the label, let's say `lastObservedHash`
- generate the hash again(maybe you need to remove the hash from label first), let's say `currObservedHash`

Then compare the two hash:
if `lastObservedHash` == `currObservedHash` ==> ignore

Am I right?

pkg/detector/detector.go
zhuwint: Thanks for your advice, I've updated in my current commit

pkg/detector/eventfilter.go
```diff
@@ -0,0 +1,28 @@
...
+	// Remove the status and some system defined mutable fields in metadata, including managedFields and resourceVersion.
+	// Refer to https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/object-meta/#ObjectMeta for more details.
+	removedFields := [][]string{{"status"}, {"metadata", "managedFields"}, {"metadata", "resourceVersion"}}
+	for _, r := range removedFields {
+		unstructured.RemoveNestedField(oldBackup.Object, r...)
+		unstructured.RemoveNestedField(newBackup.Object, r...)
+	}
+
+	if reflect.DeepEqual(oldBackup, newBackup) {
+		klog.V(4).Infof("Ignore update event of object (kind=%s, %s/%s) as specification no change", oldBackup.GetKind(), oldBackup.GetNamespace(), oldBackup.GetName())
```
RainbowMango, line 24: We don't need to print this log in this function, we can move this log to the place where use it. 
Since from this function of view, it is not sure it will be used in `update event`.

pkg/detector/detector.go
```diff
@@ -320,7 +320,20 @@ func (d *ResourceDetector) OnAdd(obj interface{}) {
...
+	}
+
+	unstructuredNewObj, err := helper.ToUnstructured(newObj)
+	if err != nil {
+		klog.Errorf("Failed to transform newObj, error: %v", err)
+		return
+	}
+	if SpecificationChanged(unstructuredOldObj, unstructuredNewObj) {
+		d.OnAdd(newObj)
+	}
```
RainbowMango, line 336:
```
	if !SpecificationChanged(unstructuredOldObj, unstructuredNewObj) {
		// log here
		return
	}

	d.OnAdd(newObj)
```

pkg/detector/eventfilter_test.go
```diff
@@ -0,0 +1,274 @@
...
+		},
+		{
+			name: "Only user defined fields changed(CRD_WorkLoad)",
+			oldObj: &workloadv1alpha1.Workload{
+				ObjectMeta: metav1.ObjectMeta{
+					ResourceVersion: "123456",
+					ManagedFields:   []metav1.ManagedFieldsEntry{{}},
+				},
+				Spec: workloadv1alpha1.WorkloadSpec{
+					Replicas: ptrInt32(3),
```
RainbowMango, line 176:
```suggestion
					Replicas: pointer.Int32(3),
```
import `"k8s.io/utils/pointer"`

pkg/detector/eventfilter_test.go
```diff
@@ -0,0 +1,274 @@
...
+					ManagedFields:   []metav1.ManagedFieldsEntry{{}},
+				},
+				Immutable:  nil,
+				Data:       map[string]string{"color": "green"},
+				BinaryData: nil,
+			},
+			wantChange: true,
+		},
+		{
+			name: "Only status fields changed(ConfigMap)",
```
RainbowMango, line 125: I can get that you tested `Deployment`(as k8s native resource) and `Workload`(as CRD), don't know why test `ConfigMap`, it's a special resource that doesn't have a `.status`?

RainbowMango (commented):
Generally looks good. Thanks.

pkg/detector/eventfilter_test.go
zhuwint: Yes, I think `ConfigMap` is a resource of a `storage type`, it's different from the `running type` such as `Deployment`

RainbowMango: `storage type`? I don't get it.

likakuli:
> If labels need to be added, does that mean that labels need to be added for objects that the controller cares about, such as binding, work, and something else? Are there too many objects to be processed?

`PropagationPolicy` is just like schedule policy as single cluster, `kube-scheduler` will treat scheduled and unscheduled pod differently and not each property change will cause a schedule cycle. 

Other crd, like `Work` and `rb`, we also need to add some filter for different controller but may not be hash. For examples, some controller only care about resource spec change event but status related controller may only care about status change event. 

It's ok for our scene even if there are some unnecessary event but the reconcile logic is lightweight enough.

likakuli:
> Then compare the two hash: if `lastObservedHash` == `currObservedHash` ==> ignore
> 
> Am I right?

yes, just like you said.

We add hash check only in detector's reconcile logic because it will always call update api to update original object's label and `resoureVersion` in rb.spec.resource. Thus will trigger other controller's reconcile logic. But all these reconcile is unnecessary for me.

RainbowMango (approved):
/lgtm
/approve

RainbowMango:
@likakuli I get it.
@Poor12 I guess you can take an eye on it, to see if there is a performance issue in the scenario of controller restart.

PR #2240, opened 2022-07-22
Apply ClusterClientOption to controllers

Poor12:
**What type of PR is this?**
/kind bug

Now ClusterAPIQPS and ClusterAPIBurst are not fully applied to controllers which need to connect with kube-apiservers in member clusters.

Fixes #

```release-note
karmada-search introduces `cluster-api-qps` to indicate qps to use while talking with cluster kube-apiserver and `cluster-api-burst` to indicate burst to use while talking with cluster kube-apiserver.
`cluster-api-qps` and `cluster-api-burst` in karmada-controller-manager and karmada-agent now are valid for all controllers.
```

### DIFF
--- a/cmd/karmada-search/app/options/options.go
+++ b/cmd/karmada-search/app/options/options.go
@@ -23,6 +23,10 @@ type Options struct {
 	KubeAPIQPS float32
 	// KubeAPIBurst is the burst to allow while talking with karmada-search.
 	KubeAPIBurst int
+	// ClusterAPIQPS is the QPS to use while talking with cluster kube-apiserver.
+	ClusterAPIQPS float32
+	// ClusterAPIBurst is the burst to allow while talking with cluster kube-apiserver.
+	ClusterAPIBurst int
 
 	ProfileOpts profileflag.Options
 
@@ -52,7 +56,8 @@ func (o *Options) AddFlags(flags *pflag.FlagSet) {
 	flags.IntVar(&o.KubeAPIBurst, "kube-api-burst", 60, "Burst to use while talking with karmada-apiserver. Doesn't cover events and node heartbeat apis which rate limiting is controlled by a different set of flags.")
 	flags.BoolVar(&o.DisableSearch, "disable-search", false, "Disable search feature that would save memory usage significantly.")
 	flags.BoolVar(&o.DisableProxy, "disable-proxy", false, "Disable proxy feature that would save memory usage significantly.")
-
+	flags.Float32Var(&o.ClusterAPIQPS, "cluster-api-qps", 40.0, "QPS to use while talking with cluster kube-apiserver.")
+	flags.IntVar(&o.ClusterAPIBurst, "cluster-api-burst", 60, "Burst to use while talking with cluster kube-apiserver.")
 	utilfeature.DefaultMutableFeatureGate.AddFlag(flags)
 	o.ProfileOpts.AddFlags(flags)
 }
--- a/pkg/controllers/status/cluster_status_controller.go
+++ b/pkg/controllers/status/cluster_status_controller.go
@@ -68,7 +68,7 @@ type ClusterStatusController struct {
 	GenericInformerManager      genericmanager.MultiClusterInformerManager
 	StopChan                    <-chan struct{}
 	ClusterClientSetFunc        func(string, client.Client, *util.ClientOption) (*util.ClusterClient, error)
-	ClusterDynamicClientSetFunc func(clusterName string, client client.Client) (*util.DynamicClusterClient, error)
+	ClusterDynamicClientSetFunc func(clusterName string, client client.Client, clientOption *util.ClientOption) (*util.DynamicClusterClient, error)
 	// ClusterClientOption holds the attributes that should be injected to a Kubernetes client.
 	ClusterClientOption *util.ClientOption
 
@@ -293,7 +293,7 @@ func (c *ClusterStatusController) initializeGenericInformerManagerForCluster(clu
 		return
 	}
 
-	dynamicClient, err := c.ClusterDynamicClientSetFunc(clusterClient.ClusterName, c.Client)
+	dynamicClient, err := c.ClusterDynamicClientSetFunc(clusterClient.ClusterName, c.Client, c.ClusterClientOption)
 	if err != nil {
 		klog.Errorf("Failed to build dynamic cluster client for cluster %s.", clusterClient.ClusterName)
 		return
--- a/pkg/util/membercluster_client.go
+++ b/pkg/util/membercluster_client.go
@@ -45,6 +45,14 @@ type ClientOption struct {
 	Burst int
 }
 
+// NewClientOption returns a ClientOption with qps and burst.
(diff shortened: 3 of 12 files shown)


RainbowMango:
Please update the release note according to the code touches by this PR.
The code touches not only `karmada-search`.

Poor12:
> Please update the release note according to the code touches by this PR. The code touches not only `karmada-search`.

Done

cmd/karmada-search/app/options/options.go
```diff
@@ -69,7 +74,8 @@ func (o *Options) AddFlags(flags *pflag.FlagSet) {
 
 	flags.Float32Var(&o.KubeAPIQPS, "kube-api-qps", 40.0, "QPS to use while talking with karmada-apiserver. Doesn't cover events and node heartbeat apis which rate limiting is controlled by a different set of flags.")
 	flags.IntVar(&o.KubeAPIBurst, "kube-api-burst", 60, "Burst to use while talking with karmada-apiserver. Doesn't cover events and node heartbeat apis which rate limiting is controlled by a different set of flags.")
-
+	flags.Float32Var(&o.ClusterAPIQPS, "cluster-api-qps", 40.0, "QPS to use while talking with cluster kube-apiserver. Doesn't cover events and node heartbeat apis which rate limiting is controlled by a different set of flags.")
+	flags.IntVar(&o.ClusterAPIBurst, "cluster-api-burst", 60, "Burst to use while talking with cluster kube-apiserver. Doesn't cover events and node heartbeat apis which rate limiting is controlled by a different set of flags.")
```
RainbowMango, line 78: In `karmada-search`, do we have the events and node heartbeat?

pkg/controllers/status/cluster_status_controller.go
```diff
@@ -156,7 +156,7 @@ func (c *ClusterStatusController) syncClusterStatus(cluster *clusterv1alpha1.Clu
 	currentClusterStatus := *cluster.Status.DeepCopy()
 
 	// create a ClusterClient for the given member cluster
-	clusterClient, err := c.ClusterClientSetFunc(cluster.Name, c.Client, c.ClusterClientOption)
+	clusterClient, err := c.ClusterClientSetFunc(cluster.Name, c.Client, c.ClusterClientOptions)
```
RainbowMango, line 159: For the usage of [`--cluster-api-qps` and `cluster-api-burst`](https://github.com/karmada-io/karmada/blob/654f146c39f1cf27cb6885358e91e00453638da4/cmd/controller-manager/app/options/options.go#L191-L192):

> Doesn't cover events and node heartbeat apis which rate limiting is controlled by a different set of flags.

I'm not sure the description about `events` and `node heartbeat` is specific to here(cluster status refresh)?

RainbowMango: @Garrybest Can you help to confirm it? If it is the case, this PR might break the rule.

RainbowMango: Yeah, I can see [the two flags in kubelet](https://github.com/kubernetes/kubernetes/blob/bcea98234f0fdc5ce4e976758977eec72b13bfd4/cmd/kubelet/app/options/options.go#L501-L502).

Given the `--cluster-api-qps` and `cluster-api-burst` in Karmada are used to specify the limit against `member clusters`, we won't deliver events to member cluster(but we do need to collect health states), so I think we need to update the usage description for disambiguation. How do you say?

By the way, do you think the limit should be applied in case of collecting cluster health states?

pkg/util/membercluster_client.go
```diff
@@ -84,28 +92,36 @@ func NewClusterClientSetForAgent(clusterName string, client client.Client, clien
 }
 
 // NewClusterDynamicClientSet returns a dynamic client for the given member cluster.
-func NewClusterDynamicClientSet(clusterName string, client client.Client) (*DynamicClusterClient, error) {
+func NewClusterDynamicClientSet(clusterName string, client client.Client, clientOption *ClientOption) (*DynamicClusterClient, error) {
```
RainbowMango, line 95: @Poor12 I have a question. 
Assume we have two clients that are built by the `NewClusterDynamicClientSet`, they don't affect each other, right?

A := NewClusterDynamicClientSet(xxx)
B := NewClusterDynamicClientSet(xxx)

If both A and B are used to access the sam cluster, the maximum request could be 2 * QPS, right?

pkg/controllers/status/cluster_status_controller.go
Garrybest: > So I think we need to update the usage description for disambiguation. How do you say?

Agreed. Actually controllers in karmada does not care about any events and node heartbeat in member cluster.

> By the way, do you think the limit should be applied in case of collecting cluster health states?

Do you mean cluster heartbeat lease? If so, I have the same idea too.

RainbowMango:
Please re-open it if you think it is still needed.
/close

Poor12:
/reopen
I think this PR is still needed.

pkg/util/objectwatcher/objectwatcher.go
```diff
@@ -43,22 +43,24 @@ type objectWatcherImpl struct {
 	ClusterClientSetFunc ClientSetFunc
 	resourceInterpreter  resourceinterpreter.ResourceInterpreter
 	InformerManager      genericmanager.MultiClusterInformerManager
+	ClientOption         *util.ClientOption
 }
 
 // NewObjectWatcher returns an instance of ObjectWatcher
-func NewObjectWatcher(kubeClientSet client.Client, restMapper meta.RESTMapper, clusterClientSetFunc ClientSetFunc, interpreter resourceinterpreter.ResourceInterpreter) ObjectWatcher {
+func NewObjectWatcher(kubeClientSet client.Client, restMapper meta.RESTMapper, clusterClientSetFunc ClientSetFunc, interpreter resourceinterpreter.ResourceInterpreter, clientOption *util.ClientOption) ObjectWatcher {
```
RainbowMango, line 50: This `clientOption` might be confusing with `kubeClientSet`. I think we can change the signature.

pkg/util/membercluster_client.go
RainbowMango: ping. I still want to know the answer.

RainbowMango (commented):
I wonder if it is possible to split this PR.

Poor12:
There are two places to set qps and burst.
1. ObjectWatcher will use dynamicClient to create or upgrade workloads in member clusters. Since we do not share the client for each clusters, we will create the client when we need to create a specific resource. So it will not work in this situation.
2. SingleClusterInformerManager will use the client to create informer manager mainly for watching the resources in member clusters. However, `watch` will not be limited. So this will not work, too.
```
// Watch attempts to begin watching the requested location.
// Returns a watch.Interface, or an error.
func (r *Request) Watch(ctx context.Context) (watch.Interface, error) {
	// We specifically don't want to rate limit watches, so we
	// don't use r.rateLimiter here.
```

Poor12:
Thanks for @RainbowMango to point out.
/close

Issue #2239, opened 2022-07-22
When the member cluster is removed, the deployment status does not change


chaunceyjiang:
**What happened**:
![image](https://user-images.githubusercontent.com/17962021/180376942-d86a3bb9-9977-4338-88ed-4dc051484389.png)

**What you expected to happen**:

Currently, no member cluster has nginx deployment. So , i think the nginx deployment should display `0/2`  in karmada

**How to reproduce it (as minimally and precisely as possible)**:

1. create a nginx deployment
2. create a pp
3. remove the member cluster
4. check nginx deployment status

nginx.yaml
``` yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx
  labels:
    app: nginx
spec:
  replicas: 2
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - image: docker.m.daocloud.io/nginx:1.14.2
        name: nginx
````

pp.yaml
``` yaml
apiVersion: policy.karmada.io/v1alpha1
kind: PropagationPolicy
metadata:
  name: nginx-propagation
spec:
  resourceSelectors:
    - apiVersion: apps/v1
      kind: Deployment
      name: nginx
  placement:
    clusterAffinity:
      clusterNames:
        - member1
```

**Anything else we need to know?**:

**Environment**:
- Karmada version:  master
- kubectl-karmada or karmadactl version (the result of `kubectl-karmada version` or `karmadactl version`): master
- Others:

dddddai:
I remember there was the same issue https://github.com/karmada-io/karmada/issues/768 and that was fixed
Looks like the cause is https://github.com/karmada-io/karmada/pull/1366

/assign

XiShanYongYe-Chang:
Hi @chaunceyjiang, the bugfix has been merged, you can have a retry.

/close

RainbowMango:
> Hi @chaunceyjiang, the bugfix has been merged, you can have a retry.

If afraid #2249 didn't solve this problem. This issue probably will be fixed along with #1995.

chaunceyjiang:
OK, I'll try again

Issue #1995, opened 2022-06-13
feature: unjoin a cluster gracefully


Garrybest:
**What would you like to be added**:
Now when using `karmadactl unjoin`, we delete the cluster directly. The scheduling result in resource binding still remains. I would like to make unjoin more gracefully. The procedure could refer to what `kubectl drain` does.

1. Cordon the target cluster.
2. Delete the scheduling result in `RB/CRB` which has been scheduled to the target cluster.
3. Delete the cluster.

**Why is this needed**:
Make unjoin a cluster more gracefully.

lonelyCZ:
> Delete the scheduling result in RB/CRB which has been scheduled to the target cluster.

Should we execute all these actions in `cli` tool? Or is there a `controller` to clear the scheduling result in RB/CRB?

Garrybest:
I think so, or we may add a NoExecute taint in this cluster and waint for the eviction. Which one do you prefer?

FYI, in `k/k`, `kubectl drain` delete pods in CLI tool.

carlory:
@lonelyCZ @Garrybest Can I pick up it ?

lonelyCZ:
@carlory Sure, thanks for your interest to it., you can `/assign` to youself. Look forward to your contribution!

Garrybest:
/help
/assign @carlory 

Thanks, glad to hear that. FYI, https://github.com/kubernetes/kubernetes/blob/050f930f8968874855eb215f0c0f0877bcdaa0e8/staging/src/k8s.io/kubectl/pkg/cmd/drain/drain.go#L293-L334

Garrybest:
Hi @carlory, any progress here?

carlory:
So sorry, I forget it.

I will do it this week, is still ok ? 

@Garrybest

Garrybest:
Thanks @carlory, take your time😄

RainbowMango:
This is another scenario of eviction. 
Do you know where can I find the PDB/evict-sub-resource thing? share a link? @Garrybest

Garrybest:
Eviction is a subresource of Pod.

https://kubernetes.io/docs/concepts/workloads/pods/disruptions/#pod-disruption-budgets

Cluster managers and hosting providers should use tools which respect PodDisruptionBudgets by calling the [Eviction API](https://kubernetes.io/docs/tasks/administer-cluster/safely-drain-node/#eviction-api) instead of directly deleting pods or deployments.

RainbowMango:
[...]

PR #2249, opened 2022-07-25
Make sure to update resource status

dddddai:
**What type of PR is this?**
/kind bug

This is a regression during the development where we added resource interpreter, this pr fixes it

For https://github.com/karmada-io/karmada/issues/2239

```release-note
NONE
```

### DIFF
--- a/pkg/resourceinterpreter/defaultinterpreter/default.go
+++ b/pkg/resourceinterpreter/defaultinterpreter/default.go
@@ -72,7 +72,7 @@ func (e *DefaultInterpreter) HookEnabled(kind schema.GroupVersionKind, operation
 func (e *DefaultInterpreter) GetReplicas(object *unstructured.Unstructured) (int32, *workv1alpha2.ReplicaRequirements, error) {
 	handler, exist := e.replicaHandlers[object.GroupVersionKind()]
 	if !exist {
-		return 0, &workv1alpha2.ReplicaRequirements{}, fmt.Errorf("defalut %s interpreter for %q not found", configv1alpha1.InterpreterOperationInterpretReplica, object.GroupVersionKind())
+		return 0, &workv1alpha2.ReplicaRequirements{}, fmt.Errorf("default %s interpreter for %q not found", configv1alpha1.InterpreterOperationInterpretReplica, object.GroupVersionKind())
 	}
 	return handler(object)
 }
@@ -81,7 +81,7 @@ func (e *DefaultInterpreter) GetReplicas(object *unstructured.Unstructured) (int
 func (e *DefaultInterpreter) ReviseReplica(object *unstructured.Unstructured, replica int64) (*unstructured.Unstructured, error) {
 	handler, exist := e.reviseReplicaHandlers[object.GroupVersionKind()]
 	if !exist {
-		return nil, fmt.Errorf("defalut %s interpreter for %q not found", configv1alpha1.InterpreterOperationReviseReplica, object.GroupVersionKind())
+		return nil, fmt.Errorf("default %s interpreter for %q not found", configv1alpha1.InterpreterOperationReviseReplica, object.GroupVersionKind())
 	}
 	return handler(object, replica)
 }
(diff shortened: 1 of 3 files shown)


pkg/controllers/binding/binding_controller.go
```diff
@@ -152,15 +152,14 @@ func (c *ResourceBindingController) syncBinding(binding *workv1alpha2.ResourceBi
 		c.EventRecorder.Event(binding, corev1.EventTypeNormal, workv1alpha2.EventReasonAggregateStatusSucceed, msg)
 		c.EventRecorder.Event(workload, corev1.EventTypeNormal, workv1alpha2.EventReasonAggregateStatusSucceed, msg)
 	}
-	if len(errs) > 0 {
-		return controllerruntime.Result{Requeue: true}, errors.NewAggregate(errs)
-	}
```
RainbowMango, line 157: I don't think we should continue to update the status in an error situation.

RainbowMango (commented):
#1995 addresses an issue that tries to cleanup the `RB/CRB` after a cluster is gone. 
I think it will be helpful for the issue this PR wants to fix.

pkg/resourceinterpreter/interpreter.go
```diff
@@ -146,11 +146,6 @@ func (i *customResourceInterpreterImpl) Retain(desired *unstructured.Unstructure
 func (i *customResourceInterpreterImpl) AggregateStatus(object *unstructured.Unstructured, aggregatedStatusItems []workv1alpha2.AggregatedStatusItem) (*unstructured.Unstructured, error) {
 	klog.V(4).Infof("Begin to aggregate status for object: %v %s/%s.", object.GroupVersionKind(), object.GetNamespace(), object.GetName())
 
-	// If status has not been collected, there is no need to aggregate.
-	if len(aggregatedStatusItems) == 0 {
```
dddddai, line 150: But I still think we should update status when the aggregated status is empty?

dddddai (commented):
Ah I get it

pkg/resourceinterpreter/interpreter.go
RainbowMango: good point!
@XiShanYongYe-Chang Please confirm.

XiShanYongYe-Chang: When a `job` resource is propagated, the status slice is empty for a short time when the status collection starts. In this case, the job status is set to `Completed`. We need to figure out a way to be compatible.

dddddai: > When a job resource is propagated, the status slice is empty for a short time when the status collection starts.

Could you please explain why? I'm asking because I want to ensure my fix is correct

XiShanYongYe-Chang: At the beginning of the status collection, the resource status remains empty for a short period of time.

dddddai: Yeah I think I get it, this happens when the binding is not scheduled yet, thanks

pkg/util/helper/job.go
```diff
@@ -67,7 +67,7 @@ func ParsingJobStatus(obj *batchv1.Job, status []workv1alpha2.AggregatedStatusIt
 		})
 	}
 
-	if successfulJobs == len(status) {
+	if successfulJobs == len(status) && successfulJobs > 0 {
```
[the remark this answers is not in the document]
dddddai: Added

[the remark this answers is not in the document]
XiShanYongYe-Chang: Thanks! 
@Garrybest can you help take a look?

pkg/resourceinterpreter/defaultinterpreter/default.go
```diff
@@ -99,7 +99,7 @@ func (e *DefaultInterpreter) Retain(desired *unstructured.Unstructured, observed
 func (e *DefaultInterpreter) AggregateStatus(object *unstructured.Unstructured, aggregatedStatusItems []workv1alpha2.AggregatedStatusItem) (*unstructured.Unstructured, error) {
 	handler, exist := e.aggregateStatusHandlers[object.GroupVersionKind()]
 	if !exist {
-		return nil, fmt.Errorf("defalut %s interpreter for %q not found", configv1alpha1.InterpreterOperationAggregateStatus, object.GroupVersionKind())
+		return nil, fmt.Errorf("default %s interpreter for %q not found", configv1alpha1.InterpreterOperationAggregateStatus, object.GroupVersionKind())
```
RainbowMango, line 102: nice finding.
I found that there are `4` places that use the `defalut`, could you please fix them all?

RainbowMango: Yeah, in the same file.

dddddai: Fixed

RainbowMango: Thanks.

Garrybest:
Nice patch👍
/lgtm

RainbowMango (approved):
/approve
/hold
for @XiShanYongYe-Chang to confirm

PR #2250, opened 2022-07-25
Comment should start with a whitespace and Uppercase

dapengJacky:
Signed-off-by: jacky <peng.chen@daocloud.io>

**What type of PR is this?**
/kind cleanup

Fixes #
Comment should start with a whitespace and Uppercase

```release-note
NONE
```

### DIFF
--- a/cmd/controller-manager/app/options/options.go
+++ b/cmd/controller-manager/app/options/options.go
@@ -94,7 +94,7 @@ type Options struct {
 	// It can be set to "0" to disable the metrics serving.
 	// Defaults to ":8080".
 	MetricsBindAddress string
-	// concurrentClusterSyncs is the number of cluster objects that are
+	// ConcurrentClusterSyncs is the number of cluster objects that are
 	// allowed to sync concurrently.
 	ConcurrentClusterSyncs int
 	// ConcurrentClusterResourceBindingSyncs is the number of clusterresourcebinding objects that are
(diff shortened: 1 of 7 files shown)


RainbowMango:
I agree that comments should start with a while space but with uppercase only when it's the comment for `exported` variable/function/methods, etc.

dapengJacky:
> I agree that comments should start with a while space but with uppercase only when it's the comment for exported variable/function/methods, etc.
okay,I have updated,thanks for your advice

cmd/controller-manager/app/options/options.go
```diff
@@ -94,7 +94,7 @@ type Options struct {
 	// It can be set to "0" to disable the metrics serving.
 	// Defaults to ":8080".
 	MetricsBindAddress string
-	// concurrentClusterSyncs is the number of cluster objects that are
+	// ConcurrentClusterSyncs is the number of cluster objects that are
```
RainbowMango, line 97: nice catch!

But for others, I'm sorry I didn't make it clear.
For comments not for exported variables/functions/methods, there is no rule for should comment start with uppercase or not, so I guess no needs to update them.

dapengJacky: I agree with u,I didn't update them

RainbowMango (approved):
/lgtm
/approve

Issue #1185, opened 2021-12-29
Default PropagationPolicy


iawia002:
**What would you like to be added**:

Introduce a default `PropagationPolicy` that matches all resources, and use it directly if the resource does not have a `PropagationPolicy` configured.

Or maybe we can also introduce `PropagationPolicy` priority.

**Why is this needed**:

The scenario is as follows: in many cases I need to distribute my resources to all clusters, but to configure `PropagationPolicy` every time is very tedious, if there is a default `PropagationPolicy` will be very convenient, this may also be able to achieve direct distribution of a helm application.

RainbowMango:
>  in many cases I need to distribute my resources to all clusters, 

What's the exact use case we are talking about? Why users need that?

RainbowMango:
We have considered this issue before, we named it `fully propagation` mode.
Haven't do it because:
- Not so clear the user cases are.
- Need a way to identify the resources between users and karmada-apiserver auto-generated?

iawia002:
> > in many cases I need to distribute my resources to all clusters,
> 
> What's the exact use case we are talking about? Why users need that?

Not necessarily all clusters, but just need a default `PropagationPolicy` to describe which clusters to distribute to, many of our users will use the same policy to distribute resources to several clusters (just as today I create an application and distribute it to cluster A B with the same number of replicas, tomorrow I create a new application and need to use the same strategy), this time the default `PropagationPolicy` is very convenient, if when there is a separate need, you can create a separate `PropagationPolicy`.

> karmada-apiserver auto-generated?

Do we have such resources, like what? We can label such resources and make the default `PropagationPolicy` ignores such resources.

GitHubxsy:
You want a propagation policy that can match any resource?

iawia002:
> You want a propagation policy that can match any resource? For example：
> 
> ```
> apiVersion: policy.karmada.io/v1alpha1
> kind: PropagationPolicy
> metadata:
>   name: example-policy
>   namespace: default
> spec:
>   resourceSelectors:
>     - apiVersion: *
>       kind: *
>       name: *
>   placement:
>     clusterAffinity:
>       clusterNames:
>         - cluster1
>         - cluster2
> ```

It's not just global matching, it's about the priority, If I configure a single policy for a resource, which policy will be used? Can I know and control this clearly?

GitHubxsy:
> 

If there are multiple policies that can match, we use lexicographic sorting. Priority is a good idea.

[...]

PR #2267, opened 2022-07-27
add implicit priority for PP

Garrybest:

**What type of PR is this?**
/kind feature

If more than one PP matches the object, we shuold choose the most suitable one. However, now we select PP by alphabetical order, it does not make any sense.

There are 4 types of implicit priority here, ordered by descend.
1. Name: ResourceSelector contains object Name and matches the object.
2. LabelSelector: ResourceSelector contains a LabelSelector and matches the object.
3. All: ResourceSelector has no Name or LabelSelector, which selects all target objects in the namespace.
4. MisMatch: ResourceSelector does not match the object.

For example:

```yaml
kind: PropagationPolicy
metadata:
  name: nginx-propagation-1
spec:
  resourceSelectors:
    - apiVersion: apps/v1
      kind: Deployment
      name: nginx
```

```yaml
kind: PropagationPolicy
metadata:
  name: nginx-propagation-2
spec:
  resourceSelectors:
    - apiVersion: apps/v1
      kind: Deployment
```

In this case, `nginx-propagation-1` should be privileged compared to `nginx-propagation-2` because it's more precise.

Fixes #

```release-note
add implicit priority for PropagationPolicy
```

### DIFF
--- a/pkg/detector/detector.go
+++ b/pkg/detector/detector.go
@@ -373,9 +373,16 @@ func (d *ResourceDetector) LookForMatchedPolicy(object *unstructured.Unstructure
 		policyList = append(policyList, policy)
 	}
 
-	var matchedPolicy *policyv1alpha1.PropagationPolicy
+	var (
+		matchedPolicy         *policyv1alpha1.PropagationPolicy
+		matchedPolicyPriority = util.PriorityMisMatch
+	)
+
 	for _, policy := range policyList {
-		if util.ResourceMatchSelectors(object, policy.Spec.ResourceSelectors...) {
+		if p := util.ResourceMatchSelectorsPriority(object, policy.Spec.ResourceSelectors...); p > matchedPolicyPriority {
+			matchedPolicy = policy
+			matchedPolicyPriority = p
+		} else if p > util.PriorityMisMatch && p == matchedPolicyPriority {
 			matchedPolicy = GetHigherPriorityPropagationPolicy(matchedPolicy, policy)
 		}
 	}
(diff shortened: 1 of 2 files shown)


RainbowMango:
I'm also thinking to introduce an explicit priority just like #1185 proposed.
You prefer to `explicit` or `implicit`?

Garrybest:
Well, I prefer implicit priority, it's easier for users. And I think if there are a lot pps, maybe it's a burden for users to care about priority of every PP/CPP.

RainbowMango:
I suppose the `implicit priority` and `explicit priority` don't conflict with each other. The `explicit priority` will have a default priority, but with a higher priority than the implicit one.

Garrybest:
> I suppose the implicit priority and explicit priority don't conflict with each other. The explicit priority will have a default priority, but with a higher priority than the implicit one.

Agree. They could both exist.

chaunceyjiang:
I suggest that the OP also add a rule like this.



> There are 4 types of implicit priority here, ordered by descend.
> 
> 1. Name: ResourceSelector contains object Name and matches the object.
> 2. LabelSelector: ResourceSelector contains a LabelSelector and matches the object.
> 3. All: ResourceSelector has no Name or LabelSelector, which selects all target objects in the namespace.
> 4. MisMatch: ResourceSelector does not match the object.

RainbowMango:
Hey guys, I hope I can get back to this next week.

RainbowMango:
@Garrybest I'm ok with the implicit priority, for now, it won't conflict with `explicit priority` that we are probably going to do.

Just a question, what's the use case that they deploy multiple PPs, especially when they might have intersections?

Garrybest:
In my scenario, I create a default pp for users to select all deployments in a namespace called `common-pp`.

```
  resourceSelectors:
  - apiVersion: apps/v1
    kind: Deployment
    namespace: ns1
```

So users could create a deployment and propagate them quickly. But some deployments may not want to use this default pp, e.g., they request GPUs which are not included in `common-pp`. So we create a new pp and use labels to match resource and new pp.

RainbowMango:
Got it. It is reasonable.
/assign

RainbowMango:
Can you help to rebase this commit? I suppose it'd be better to review and run tests against the master branch.

Garrybest:
Sure.

RainbowMango:
> So should I made a change in this PR too?

No, I didn't mean that. 

I'm thinking if we can make some common code because the process is probably common:
1. get the list of PP(or OP) that matches the workload.
2. sort the list by priority (I assume the list won't be so long, so performance won't be a concern here.)
3. for PP, we just pick the first item from the list, for OP, we apply the whole list by order.

RainbowMango:
/lgtm

@chaunceyjiang Would you like to take a look?
We can make some changes, like sharing code,  when implementing the OverridePolicy part.

Garrybest:
Got it, the patch will be added in next PR.

chaunceyjiang:
> /lgtm
> 
> @chaunceyjiang Would you like to take a look? We can make some changes, like sharing code, when implementing the OverridePolicy part.

of course

RainbowMango:
@Garrybest 
I tried making some changes based on your commit([see this commit](https://github.com/RainbowMango/karmada/commit/2be17917997331c987cc22abcc5f1a3efa6f0245)), do you think it reasonable?

Garrybest:
> @Garrybest I tried making some changes based on your commit([see this commit](https://github.com/RainbowMango/karmada/commit/2be17917997331c987cc22abcc5f1a3efa6f0245)), do you think it reasonable?

Cool, I will take this one 😊

RainbowMango (approved):
/lgtm
/approve

PR #2269, opened 2022-07-27
Adjust the order of kubeconfig loadingRules

zhuwint:
**What type of PR is this?**

/kind bug

If we use the same `karmadaConfig` object, and call method `GetClientConfig` more than once. The first time it would set the `ExplicitPath` and `Precedence` successfully. But the second time it would set `Precedence` with the value of `ExplicitPath` the first time set.

Fixes #

NONE

### DIFF
--- a/pkg/karmadactl/cmdinit/utils/kubeclient.go
+++ b/pkg/karmadactl/cmdinit/utils/kubeclient.go
@@ -13,8 +13,8 @@ func RestConfig(context, kubeconfigPath string) (*rest.Config, error) {
 	pathOptions := clientcmd.NewDefaultPathOptions()
 
 	loadingRules := *pathOptions.LoadingRules
-	loadingRules.Precedence = pathOptions.GetLoadingPrecedence()
 	loadingRules.ExplicitPath = kubeconfigPath
+	loadingRules.Precedence = pathOptions.GetLoadingPrecedence()
 	overrides := &clientcmd.ConfigOverrides{
 		CurrentContext: context,
 	}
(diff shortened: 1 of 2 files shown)


lonelyCZ:
It makes sense.

https://github.com/karmada-io/karmada/blob/dabdbea0fb89bdce97accf231dafafe97b0d7861/vendor/k8s.io/client-go/tools/clientcmd/config.go#L85-L88

https://github.com/karmada-io/karmada/blob/dabdbea0fb89bdce97accf231dafafe97b0d7861/vendor/k8s.io/client-go/tools/clientcmd/config.go#L137-L139

https://github.com/karmada-io/karmada/blob/dabdbea0fb89bdce97accf231dafafe97b0d7861/vendor/k8s.io/client-go/tools/clientcmd/config.go#L141-L143

zhuwint:
Used here
> https://github.com/karmada-io/karmada/blob/dabdbea0fb89bdce97accf231dafafe97b0d7861/pkg/karmadactl/join.go#L140-L151

lonelyCZ:
https://github.com/karmada-io/karmada/blob/f8e4c47acc9b19674230ec61fcaa5cb325e92038/pkg/karmadactl/cmdinit/utils/kubeclient.go#L16-L17

Does it also need to be adjusted here?

zhuwint:
> https://github.com/karmada-io/karmada/blob/f8e4c47acc9b19674230ec61fcaa5cb325e92038/pkg/karmadactl/cmdinit/utils/kubeclient.go#L16-L17
> 
> Does it also need to be adjusted here?

Thansks for pointing out. 

The `PathOptions` will be recreated on every function call, so is not necessnary to change the order. But if we provide `kubeconfigPath` in command line, it's no need to read the `$KUBECONFIG`. For this reason, I think we should adjust the order here.

I've updated in the current commit

lonelyCZ:
Thanks @zhuwint 

/lgtm

/assign @RainbowMango

RainbowMango (commented):
/approve

PR #2273, opened 2022-07-28
Extend Binding API for graceful eviction

RainbowMango:
**What type of PR is this?**
/kind api-change
/kind feature

This PR proposed the API change which is used to perform graceful eviction.

Part of #2281

```release-note
NONE
```


pkg/apis/work/v1alpha2/binding_types.go
```diff
@@ -142,6 +158,47 @@ type TargetCluster struct {
...
+	// Reason contains a programmatic identifier indicating the reason for the eviction.
+	// Producers may define expected values and meanings for this field,
+	// and whether the values are considered a guaranteed API.
+	// The value should be a CamelCase string.
+	// This field may not be empty.
+	// +required
+	// +kubebuilder:validation:MaxLength=32
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:Pattern=`^[A-Za-z]([A-Za-z0-9_,:]*[A-Za-z0-9_])?$`
+	Reason string `json:"reason" protobuf:"bytes,5,opt,name=reason"`
```
kevin-wangzefeng, line 175: To avoid hard coding strings for `Reason` in multiple locations, you may need to enumerate valid values, or at least find a unified place to define consts.

pkg/apis/work/v1alpha2/binding_types.go
```diff
@@ -142,6 +158,47 @@ type TargetCluster struct {
...
+
+	// Message is a human-readable message indicating details about the eviction.
+	// This may be an empty string.
+	// +optional
+	// +kubebuilder:validation:MaxLength=1024
+	Message string `json:"message,omitempty"`
+
+	// Producers indicates the controller who triggered the eviction.
+	// +required
+	Producers string `json:"producers"`
```
kevin-wangzefeng, line 185: 1. Should it be `Producer` or `Producers` here? If it's `Producers`, you'd better define delimiter for spliting each producer. And clarifing the behavior of "mutiple producers generating the same `GracefulEvictionTask` would be very helpful."
2. To avoid hard coding strings for producers in multiple locations, you may need to enumerate valid producers, or at least find a unified place to define consts.

kevin-wangzefeng (commented):
I don't see any relevant proposals or issues linked to this PR, it would be very helpful to create one, describing background/motivation and also system behaviours especially when multiple evictors working together.

pkg/apis/work/v1alpha2/binding_types.go
RainbowMango: Done. Added a const in new push.

pkg/apis/work/v1alpha2/binding_types.go
RainbowMango: Done. Added a const in new push.

RainbowMango:
cc @XiShanYongYe-Chang @Garrybest @CharlesQQ

pkg/apis/work/v1alpha2/binding_types.go
```diff
@@ -72,6 +72,22 @@ type ResourceBindingSpec struct {
 	// +optional
 	Clusters []TargetCluster `json:"clusters,omitempty"`
 
+	// GracefulEvictionTasks holds the eviction tasks that are expected to perform
+	// the eviction in a graceful way.
+	// The intended workflow is:
+	// 1. Once the controller(such as 'taint-manager') decided to evict the resource that
+	//    is referenced by current ResourceBinding or ClusterResourceBinding from a target
+	//    cluster, it removes(or scale down the replicas) the target from Clusters(.spec.Clusters)
```
XiShanYongYe-Chang, line 80: > (or scale down the replicas)

Do we need to move the scale down replicas into `GracefulEvictionTasks`?

pkg/apis/work/v1alpha2/binding_types.go
```diff
@@ -72,6 +72,22 @@ type ResourceBindingSpec struct {
...
+	//    and builds a graceful eviction task.
+	// 2. The scheduler may perform a re-scheduler and probably select a substitute cluster
+	//    to take over the evicting workload(resource).
+	// 3. The graceful eviction controller takes care of the graceful eviction tasks and
+	//    performs the final removal after the workload(resource) is available on the substitute
+	//    cluster or exceed the grace termination period(defaults to 10 minutes).
```
XiShanYongYe-Chang, line 86: > the grace termination period(defaults to 10 minutes).

Is this a configurable parameter?

pkg/apis/work/v1alpha2/binding_types.go
```diff
@@ -142,6 +158,47 @@ type TargetCluster struct {
 	Replicas int32 `json:"replicas,omitempty"`
 }
 
+// GracefulEvictionTask represents an graceful eviction task.
+type GracefulEvictionTask struct {
+	// Name of the cluster where the eviction perform from.
+	Name string `json:"name"`
```
XiShanYongYe-Chang, line 164: Would it be better to use `ClusterName`? Because there is no cluster in the current context.

pkg/apis/work/v1alpha2/binding_types.go
RainbowMango: Yes, e.g.
```yaml
spec:
  clusters:
  - name: member2
    replicas: 10
  - name: member1
    replicas: 10
```
evict `5` replicas from `member1`, then
```yaml
spec:
  clusters:
  - name: member2
    replicas: 10
  - name: member1
    replicas: 5
  gracefulEvictionTasks:
  - name: member1
    replicas: 5
    ...
```

pkg/apis/work/v1alpha2/binding_types.go
RainbowMango: Yes, probably. And a new field might be needed to hold the configurable parameter.

pkg/apis/work/v1alpha2/binding_types.go
RainbowMango: Good point. This field should represent which cluster the referencing resource is evicted from.

pkg/apis/work/v1alpha2/binding_types.go
XiShanYongYe-Chang: This may cause problems. `.spec.clusters` describes the current cluster replicas distribution status. When the number of  member2 replicas changes from 10 to 5, the number of workload replicas will be directly changed when `ensuringWork` in binding-controller. This will conflict with the eviction behavior.

RainbowMango: > When the number of member2 replicas changes from 10 to 5,

`member2` do you mean `member1`?

XiShanYongYe-Chang: Clerical error. It does refer to `member1`.

RainbowMango: Yeah, that would be too complicated, and given the real use case is uncertain yet, I suppose we can remove the `replicas` from `gracefulEvictionTasks`. Now only focus on evicting the entire resource from clusters. 

@Garrybest What do you say?

pkg/apis/work/v1alpha2/binding_types.go
```diff
@@ -142,6 +158,47 @@ type TargetCluster struct {
...
+	// It is represented in RFC3339 form(like '2021-04-25T10:02:10Z') and is in UTC.
+	//
+	// Populated by the system. Read-only.
+	// +optional
+	CreationTimestamp metav1.Time `json:"creationTimestamp,omitempty"`
+
+	// Replicas indicates the number of replicas should be evicted.
+	// Should be ignored for resource type that doesn't have replica.
+	// +optional
+	Replicas int32 `json:"replicas,omitempty"`
```
Garrybest, line 199: What about `*int32`? Non-workload could set it to nil.

pkg/apis/work/v1alpha2/well_known_constants.go
```diff
@@ -59,3 +59,16 @@ const (
 	// https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#uids
 	ResourceTemplateUIDAnnotation = "resourcetemplate.karmada.io/uid"
 )
+
+// Define eviction reasons.
+const (
+	// EvictionReasonExceedTaintToleration describes the eviction is triggered
+	// because can not tolerate taint or exceed toleration period of time.
+	EvictionReasonExceedTaintToleration = "ExceedTaintToleration"
```
Garrybest, line 67: What about `EvictionReasonTaintUntolerated`?

pkg/apis/work/v1alpha2/binding_types.go
```diff
@@ -161,6 +218,11 @@ type BindingSnapshot struct {
 
 // ResourceBindingStatus represents the overall status of the strategy as well as the referenced resources.
 type ResourceBindingStatus struct {
+	// SchedulerObservedGeneration is the generation(.metadata.generation) observed by the scheduler.
+	// If SchedulerObservedGeneration is less than the generation in metadata means the scheduler hasn't confirmed
+	// the scheduling result or hasn't done the schedule yet.
+	// +optional
+	SchedulerObservedGeneration int64 `json:"schedulerObservedGeneration,omitempty"`
```
Garrybest, line 225: So what if the evicition task has not been done yet but there comes a scaling-up event, how to deal with this situation by using `SchedulerObservedGeneration`?

pkg/apis/work/v1alpha2/well_known_constants.go
RainbowMango: +1

pkg/apis/work/v1alpha2/binding_types.go
RainbowMango: I'm not sure if I understand you correctly. 

A general idea in my mind is that `scheduler` shouldn't schedule the resource(in binding) to a cluster which already in one of the `graceful eviction tasks`.

pkg/apis/work/v1alpha2/well_known_constants.go
RainbowMango: done.

pkg/apis/work/v1alpha2/binding_types.go
RainbowMango: done.

pkg/apis/work/v1alpha2/binding_types.go
RainbowMango: Done. `Name` --> `FromCluster`.

PR #2284, opened 2022-07-29
add default tolerations for pp and cpp

Garrybest:
**What type of PR is this?**
/kind feature

Add default tolerations, defaultNotReadyTolerationSeconds and defaultUnreachableTolerationSeconds, for pp and cpp.

Refer to https://github.com/kubernetes/kubernetes/blob/24753aa8a4df8d10bfd6330e0f29186000c018be/plugin/pkg/admission/defaulttolerationseconds/admission.go

Fixes #

```release-note
`karmada-webhook`: Added default tolerations, defaultNotReadyTolerationSeconds and defaultUnreachableTolerationSeconds, for (Cluster)PropagationPolicy.
```

### DIFF
--- a/cmd/webhook/app/options/options.go
+++ b/cmd/webhook/app/options/options.go
@@ -47,6 +47,9 @@ type Options struct {
 	// Defaults to ":8000".
 	HealthProbeBindAddress string
 
+	DefaultNotReadyTolerationSeconds    int64
+	DefaultUnreachableTolerationSeconds int64
+
 	ProfileOpts profileflag.Options
 }
 
@@ -72,6 +75,10 @@ func (o *Options) AddFlags(flags *pflag.FlagSet) {
 	flags.IntVar(&o.KubeAPIBurst, "kube-api-burst", 60, "Burst to use while talking with karmada-apiserver. Doesn't cover events and node heartbeat apis which rate limiting is controlled by a different set of flags.")
 	flags.StringVar(&o.MetricsBindAddress, "metrics-bind-address", ":8080", "The TCP address that the controller should bind to for serving prometheus metrics(e.g. 127.0.0.1:8088, :8088)")
 	flags.StringVar(&o.HealthProbeBindAddress, "health-probe-bind-address", ":8000", "The TCP address that the controller should bind to for serving health probes(e.g. 127.0.0.1:8000, :8000)")
+
+	// webhook flags
+	flags.Int64Var(&o.DefaultNotReadyTolerationSeconds, "default-not-ready-toleration-seconds", 300, "Indicates the tolerationSeconds of the propagation policy toleration for notReady:NoExecute that is added by default to every propagation policy that does not already have such a toleration.")
+	flags.Int64Var(&o.DefaultUnreachableTolerationSeconds, "default-unreachable-toleration-seconds", 300, "Indicates the tolerationSeconds of the propagation policy toleration for unreachable:NoExecute that is added by default to every propagation policy that does not already have such a toleration.")
 
 	o.ProfileOpts.AddFlags(flags)
 }
--- a/cmd/webhook/app/webhook.go
+++ b/cmd/webhook/app/webhook.go
@@ -111,9 +111,11 @@ func Run(ctx context.Context, opts *options.Options) error {
 
 	klog.Info("registering webhooks to the webhook server")
 	hookServer := hookManager.GetWebhookServer()
-	hookServer.Register("/mutate-propagationpolicy", &webhook.Admission{Handler: &propagationpolicy.MutatingAdmission{}})
+	hookServer.Register("/mutate-propagationpolicy", &webhook.Admission{Handler: propagationpolicy.NewMutatingHandler(
+		opts.DefaultNotReadyTolerationSeconds, opts.DefaultUnreachableTolerationSeconds)})
 	hookServer.Register("/validate-propagationpolicy", &webhook.Admission{Handler: &propagationpolicy.ValidatingAdmission{}})
-	hookServer.Register("/mutate-clusterpropagationpolicy", &webhook.Admission{Handler: &clusterpropagationpolicy.MutatingAdmission{}})
+	hookServer.Register("/mutate-clusterpropagationpolicy", &webhook.Admission{Handler: clusterpropagationpolicy.NewMutatingHandler(
+		opts.DefaultNotReadyTolerationSeconds, opts.DefaultUnreachableTolerationSeconds)})
 	hookServer.Register("/validate-clusterpropagationpolicy", &webhook.Admission{Handler: &clusterpropagationpolicy.ValidatingAdmission{}})
 	hookServer.Register("/mutate-overridepolicy", &webhook.Admission{Handler: &overridepolicy.MutatingAdmission{}})
 	hookServer.Register("/validate-overridepolicy", &webhook.Admission{Handler: &overridepolicy.ValidatingAdmission{}})
--- a/pkg/util/helper/taint.go
+++ b/pkg/util/helper/taint.go
@@ -9,12 +9,23 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	clusterv1alpha1 "github.com/karmada-io/karmada/pkg/apis/cluster/v1alpha1"
+	policyv1alpha1 "github.com/karmada-io/karmada/pkg/apis/policy/v1alpha1"
 )
 
 // TaintExists checks if the given taint exists in list of taints. Returns true if exists false otherwise.
 func TaintExists(taints []corev1.Taint, taintToFind *corev1.Taint) bool {
 	for _, taint := range taints {
 		if taint.MatchTaint(taintToFind) {
+			return true
+		}
+	}
+	return false
+}
+
+// TolerationExists checks if the given toleration exists in list of tolerations. Returns true if exists false otherwise.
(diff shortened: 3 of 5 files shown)


pkg/util/helper/policy.go
```diff
@@ -33,6 +35,33 @@ func SetDefaultSpreadConstraints(spreadConstraints []policyv1alpha1.SpreadConstr
 	}
 }
 
+// SetDefaultClusterTolerations set default notReadyToleration and unreachableToleration if not set.
+func SetDefaultClusterTolerations(placement *policyv1alpha1.Placement, notReadyToleration, unreachableToleration corev1.Toleration) {
```
XiShanYongYe-Chang, line 39: This function is a little redundant. We can ignore which `corev1.Toleration` is set, suggest:

```
func SetDefaultClusterTolerations(placement *policyv1alpha1.Placement, tolerations ...corev1.Toleration)
```
or
```
func SetDefaultClusterToleration(placement *policyv1alpha1.Placement, toleration corev1.Toleration)
```

cmd/webhook/app/options/options.go
```diff
@@ -75,3 +79,15 @@ func (o *Options) AddFlags(flags *pflag.FlagSet) {
...
+// AddCustomGlobalFlags explicitly registers flags that internal packages register
+// against the global flagsets from "flag".
+func (o *Options) AddCustomGlobalFlags(flags *pflag.FlagSet) {
+	// Lookup flags in global flag set and re-register the values with our flagset.
+
+	// Adds flags from k8s.io/apiserver/pkg/admission.
+	globalflag.Register(flags, "default-propagation-policy-not-ready-toleration-seconds")
+	globalflag.Register(flags, "default-propagation-policy-unreachable-toleration-seconds")
+	globalflag.Register(flags, "default-cluster-propagation-policy-not-ready-toleration-seconds")
+	globalflag.Register(flags, "default-cluster-propagation-policy-unreachable-toleration-seconds")
```
XiShanYongYe-Chang, line 92: IMOP, can we not distinguish between `PropagationPolicy` and `ClusterPropagationPolicy`?

Garrybest: Good idea. I also think distinguishing may be redundant.

cmd/webhook/app/options/options.go
```diff
@@ -2,8 +2,12 @@ package options
 
 import (
 	"github.com/spf13/pflag"
+	"k8s.io/component-base/cli/globalflag"
 
 	"github.com/karmada-io/karmada/pkg/sharedcli/profileflag"
+	// ensure libs have a chance to globally register their flags
+	_ "github.com/karmada-io/karmada/pkg/webhook/clusterpropagationpolicy"
+	_ "github.com/karmada-io/karmada/pkg/webhook/propagationpolicy"
```
XiShanYongYe-Chang, line 10: I have a doubt about whether the import path needs to be changed to `github.com/karmada-io/karmada/pkg/webhook/defaulttolerationseconds`.

Garrybest: Sorry, my mistake here 🤣

XiShanYongYe-Chang (commented):
/lgtm

cmd/webhook/app/options/options.go
```diff
@@ -75,3 +78,13 @@ func (o *Options) AddFlags(flags *pflag.FlagSet) {
 
 	o.ProfileOpts.AddFlags(flags)
 }
+
+// AddCustomGlobalFlags explicitly registers flags that internal packages register
+// against the global flagsets from "flag".
+func (o *Options) AddCustomGlobalFlags(flags *pflag.FlagSet) {
```
RainbowMango, line 84: Who will use this function?

cmd/webhook/app/options/options.go
```diff
@@ -2,8 +2,11 @@ package options
 
 import (
 	"github.com/spf13/pflag"
+	"k8s.io/component-base/cli/globalflag"
 
 	"github.com/karmada-io/karmada/pkg/sharedcli/profileflag"
+	// ensure libs have a chance to globally register their flags
+	_ "github.com/karmada-io/karmada/pkg/webhook/defaulttolerationseconds"
```
RainbowMango, line 9: Here, we don't use anything defined at `pkg/webhook/defaulttolerationseconds`. right?

cmd/webhook/app/options/options.go
Garrybest: In `NewWebhookCommand`, I forgot it.

cmd/webhook/app/options/options.go
Garrybest: We want to initialize `defaultNotReadyTolerationSeconds` and `defaultUnreachableTolerationSeconds` first. So before `AddCustomGlobalFlags`, the `flag.CommandLine` now already has these two flags. When we call `AddCustomGlobalFlags`, these two flags are retrieved from `flag.CommandLine` and inject into `pflag.FlagSet`, please refer to `k8s.io/component-base/cli/globalflag`.

RainbowMango: I know, when importing `"github.com/karmada-io/karmada/pkg/webhook/defaulttolerationseconds"`, the two flags will be added to `commandline`, then `globalflag.Register(flags,...)` moves the two flags to our FlagSet, right?

It's a little bit tricky, isn't it? Why not just add the two flags in this file?

Garrybest: Because it's ugly to pass these parameters into the struct `MutatingAdmission`. We use these two flags to initialize variables`notReadyToleration` and `unreachableToleration`. If we pass these the flags, we could not new the variables in advance and we must set this field every time when mutating function is called. I think this is why `k/k` also does.

RainbowMango: >  If we pass these the flags, we could not new the variables in advance and we must set this field every time when mutating function is called.

I don't understand why you think `set this field every time`, do you mean the [webhook.Admission](https://github.com/karmada-io/karmada/blob/e1536144771e5a67b9f1043feea97a225e9edd88/cmd/webhook/app/webhook.go#L114) will be build each HTTP call?

> Because it's ugly to pass these parameters into the struct MutatingAdmission.

I made a try based on your commit(https://github.com/karmada-io/karmada/commit/9a4821db6ba34844170b27d74a8c78719efa740e), is it the way(pass parameters) you think ugly?

Garrybest: > I don't understand why you think set this field every time, do you mean the [webhook.Admission](https://github.com/karmada-io/karmada/blob/e1536144771e5a67b9f1043feea97a225e9edd88/cmd/webhook/app/webhook.go#L114) will be build each HTTP call?

Right, https://github.com/karmada-io/karmada/commit/9a4821db6ba34844170b27d74a8c78719efa740e generate default toleration every time when handle each HTTP request. Acutually, this toleration will never change since we have set the flag.

I guess the flag is more like a kind of global ones. Once it is set, the toleration is determined. So pp and cpp mutating webhook could use these tolerations directly.

RainbowMango: > I guess the flag is more like a kind of global ones. Once it is set, the toleration is determined. 

Yes, they are initialized only once.

> Right, https://github.com/karmada-io/karmada/commit/9a4821db6ba34844170b27d74a8c78719efa740e generate default toleration every time when handle each HTTP request.

You are right, I just made another patch at https://github.com/karmada-io/karmada/commit/b25b6e74b784379723ff1a6ab02105340866fb74, i guess the handler will be shared by all HTTP call.

By the way, the reason why I don't like initializing flags by `init` is when people import something(like functions) from the package they don't know the flags are added. I think it's a bad programming model, I have some lessons learned before.
For example, the controller-runtime framework will add `--kubeconfig` flag by default(we ever don't know it since we just import some functions from the package), the flag is useful for most cases, but doesn't make sense to us, we have to modify it, like:
https://github.com/karmada-io/karmada/blob/e1536144771e5a67b9f1043feea97a225e9edd88/cmd/controller-manager/app/controllermanager.go#L80-L84

> I think this is why k/k also does.

In addition, you mentioned Kubernetes did the same thing, I think the reason is the flag is defined and used by a specific admission, so they put the flags in the admission for maintainability. Another reason is the `Plugin` is shared by all admissions, so introducing fields into it doesn't make sense to other admissions.

Garrybest: Thanks very much, I got it. I agree with patch https://github.com/karmada-io/karmada/commit/b25b6e74b784379723ff1a6ab02105340866fb74. How can I patch the commit to this PR?

RainbowMango: This patch is a draft one only used for explaining my idea, only for your reference. (Didn't solve golint issues.)

You can fetch it from my forked repo and cherry-pick it if you want, like:
```
$ git remote add hongcai https://github.com/RainbowMango/karmada.git
$ git fetch hongcai
$ git cherry-pick b25b6e74b784379723ff1a6ab02105340866fb74
```

cmd/webhook/app/options/options.go
```diff
@@ -57,6 +62,8 @@ func NewOptions() *Options {
 
 // AddFlags adds flags to the specified FlagSet.
 func (o *Options) AddFlags(flags *pflag.FlagSet) {
+	flags.AddGoFlagSet(flag.CommandLine)
```
RainbowMango, line 65: Forgot to remove it?

RainbowMango (commented):
Just a nit. Otherwise LGTM.

RainbowMango (approved):
/lgtm
/approve

Issue #2234, opened 2022-07-21
performance improvement for list request with labelSelector


likakuli:
**What would you like to be added**:

Use index to improve the performance of list request with labelSelector

**Why is this needed**:

LabelSelector is used in many places in `karmada-controller` to list object. `controller-runtime` not setup index for label and not support to setup index for label too. So the list request cost O(n). When there are many object, it will cost too much time.

controller-runtime code: https://github.com/kubernetes-sigs/controller-runtime/blob/f46919744bee01060c9084a285e049afffd38c9d/pkg/cache/internal/cache_reader.go#L115-L138

likakuli:
I don't know whether this function should be implemented in `controller-runtime` or just in `karmada-controller`.  
There are two points related to this.

1. `controller-runtime` not support index when to use labelSelector in list request.
2. `controller-runtime` not support complicated fieldSelector just as the comment. 

In our case, we implemented this by two steps.

1. We add custom manager and cacher used to convert labelSelector to fieldSelector.
2. We use custom `client-go` and `controller-runtime` package to support complicated selector. refer to:  
https://github.com/kubernetes/kubernetes/pull/109334 
https://github.com/kubernetes-sigs/controller-runtime/pull/1838

RainbowMango:
cc @yy158775 Please take a look this issue.

yy158775:
From my point of view,I think we can convet labelSelector to fieldSelector with a custom IndexFunc.The IndexFunc can compute the set of indexed values for an object's labels.

Eg: If we want to select resourcebindings with label key "clusterpropagationpolicy.karmada.io/name" and specific value,we can construct a IndexFunc which return "key=value" pair when the "clusterpropagationpolicy.karmada.io/name" exists in object labels.

Such as:
indices["labels"] = {"clusterpropagationpolicy.karmada.io/name=$var":[key1,key2,key3,......],......}
indexers["labels"] = indexfunc

If we need to select objects by other labels,we can edit the indexfunc to support.

likakuli:
If no one is working on it, i can open a pr to support this issue.

RainbowMango:
Sure thing! (anyone is welcome to pick a task without an assignment)

Have you done it on your side?

I know @yy158775 started looking at it and working on the demo.

This issue is much more involved. And @yy158775 can cooperate with you.

likakuli:
Our internal version has implemented this function.

yy158775:
[...]

PR #2287, opened 2022-07-30
feat: support use index when to list by labelSelector

likakuli:

**What type of PR is this?**
/kind feature

Support use index when to list object by labelSelector

Fixes # https://github.com/karmada-io/karmada/issues/2234

```release-note
NONE
```

### DIFF
--- a/cmd/agent/app/agent.go
+++ b/cmd/agent/app/agent.go
@@ -37,6 +37,7 @@ import (
 	"github.com/karmada-io/karmada/pkg/util/gclient"
 	"github.com/karmada-io/karmada/pkg/util/helper"
 	"github.com/karmada-io/karmada/pkg/util/informermanager"
+	"github.com/karmada-io/karmada/pkg/util/manager"
 	"github.com/karmada-io/karmada/pkg/util/names"
 	"github.com/karmada-io/karmada/pkg/util/objectwatcher"
 	"github.com/karmada-io/karmada/pkg/util/restmapper"
@@ -148,7 +149,7 @@ func run(ctx context.Context, karmadaConfig karmadactl.KarmadaConfig, opts *opti
 		return fmt.Errorf("failed to generate execution space name for member cluster %s, err is %v", opts.ClusterName, err)
 	}
 
-	controllerManager, err := controllerruntime.NewManager(controlPlaneRestConfig, controllerruntime.Options{
+	controllerManager, err := manager.NewLabelIndexerManager(controlPlaneRestConfig, controllerruntime.Options{
 		Scheme:                     gclient.NewSchema(),
 		SyncPeriod:                 &opts.ResyncPeriod.Duration,
 		Namespace:                  executionSpace,
--- a/pkg/util/manager/manager.go
+++ b/pkg/util/manager/manager.go
@@ -1,0 +1,106 @@
+package manager
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+
+	policyv1alpha1 "github.com/karmada-io/karmada/pkg/apis/policy/v1alpha1"
+	workv1alpha1 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha1"
+	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
+	"github.com/karmada-io/karmada/pkg/util"
+)
+
+// typeIndexerFuncMap contains internal type and its indexerFunc used to set up indexer.
+var typeIndexerFuncMap = map[client.Object]client.IndexerFunc{
+	&workv1alpha2.ResourceBinding{}:        indexForResourceBinding(),
+	&workv1alpha1.ResourceBinding{}:        indexForResourceBinding(),
+	&workv1alpha2.ClusterResourceBinding{}: indexForClusterResourceBinding(),
+	&workv1alpha1.ClusterResourceBinding{}: indexForClusterResourceBinding(),
+	&workv1alpha1.Work{}:                   indexForWork(),
+}
+
+type labelIndexerManager struct {
+	manager.Manager
+}
+
+// NewLabelIndexerManager returns a labelIndexerManager which set up field indexer for all needed internal type using typeIndexerFuncMap.
+// This file need to be refactored when controller-runtime support labelIndexer and client-go support multiIndexer in the future.
+// refer to:
+//		https://github.com/karmada-io/karmada/issues/2234
+//		https://github.com/kubernetes/kubernetes/pull/109334
+//      https://github.com/kubernetes-sigs/controller-runtime/pull/1838
+func NewLabelIndexerManager(config *rest.Config, options manager.Options) (manager.Manager, error) {
+	manager, err := manager.New(config, options)
+	if err != nil {
+		return nil, err
+	}
+
+	lim := &labelIndexerManager{
+		Manager: manager,
+	}
+
+	// register label indexer
+	ctx := context.Background()
+	for obj, indexerFunc := range typeIndexerFuncMap {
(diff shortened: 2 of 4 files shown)


pkg/util/manager/manager.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	lim := &labelIndexerManager{
+		Manager: manager,
+	}
+
+	// register label indexer
+	ctx := context.Background()
+	for obj, indexerFunc := range typeIndexerFuncMap {
+		err := lim.Manager.GetFieldIndexer().IndexField(ctx, obj, "metadata.labels", indexerFunc)
```
Garrybest, line 50: transfer `metadata.labels` to a constant.

pkg/util/manager/manager.go
```diff
@@ -0,0 +1,106 @@
...
+	policyv1alpha1 "github.com/karmada-io/karmada/pkg/apis/policy/v1alpha1"
+	workv1alpha1 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha1"
+	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
+	"github.com/karmada-io/karmada/pkg/util"
+)
+
+// typeIndexerFuncMap contains internal type and its indexerFunc used to set up indexer.
+var typeIndexerFuncMap = map[client.Object]client.IndexerFunc{
+	&workv1alpha2.ResourceBinding{}:        indexForResourceBinding(),
+	&workv1alpha1.ResourceBinding{}:        indexForResourceBinding(),
```
Garrybest, line 21: I think there is no need to index `workv1alpha1.ResourceBinding`, `workv1alpha1.ClusterResourceBinding` either.

cmd/agent/app/agent.go
```diff
@@ -148,7 +149,7 @@ func run(ctx context.Context, karmadaConfig karmadactl.KarmadaConfig, opts *opti
 		return fmt.Errorf("failed to generate execution space name for member cluster %s, err is %v", opts.ClusterName, err)
 	}
 
-	controllerManager, err := controllerruntime.NewManager(controlPlaneRestConfig, controllerruntime.Options{
+	controllerManager, err := manager.NewLabelIndexerManager(controlPlaneRestConfig, controllerruntime.Options{
```
Garrybest, line 152: I don't know why we establish our own manager and cache in `util/manager`. I think we only need to add indexer to this controllerManager. And I have implement the respective indexer in controllerManager in https://github.com/karmada-io/karmada/blob/master/pkg/controllers/cluster/common.go

Garrybest: And here is the list function. The field `client.Client` already use a cache here. I think there is no need to establish our own cache, right? It could make the code cleaner.

https://github.com/karmada-io/karmada/blob/616033fc0e7bb5bc68d5932ed74194582c00dad6/pkg/controllers/cluster/taint_manager.go#L68-L74

likakuli: thanks for your suggestion. 

I think there are three reasons why add custom manager and cache.
There are many callers to call list func with labelSelector and the first reason is just don't want to modify too many callers. 

The second is that when controller-runtime and client-go support use multi-indexer for labelSelector then what we 
 need to do is just use original manager and cache instead of our custom types. 

The third is that it's a non-business logic. User just do what they need to do in its controller.

pkg/util/manager/manager.go
```diff
@@ -0,0 +1,106 @@
...
+	&workv1alpha2.ClusterResourceBinding{}: indexForClusterResourceBinding(),
+	&workv1alpha1.ClusterResourceBinding{}: indexForClusterResourceBinding(),
+	&workv1alpha1.Work{}:                   indexForWork(),
+}
+
+type labelIndexerManager struct {
+	manager.Manager
```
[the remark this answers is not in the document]
likakuli: got it, thx

pkg/util/manager/manager.go
```diff
@@ -0,0 +1,106 @@
...
+
+func indexForClusterResourceBinding() client.IndexerFunc {
+	return func(object client.Object) []string {
+		return indexKeys(object, []string{
+			policyv1alpha1.ClusterPropagationPolicyLabel,
+		})
+	}
+}
+
+func indexForWork() client.IndexerFunc {
```
yy158775, line 77: When List work objects with label FederatedResourceQuotaNamespaceLabel and FederatedResourceQuotaNameLabel
https://github.com/karmada-io/karmada/blob/ad94a94dce24b98e59cdd9d5018a983ef0fa64ac/pkg/controllers/federatedresourcequota/federated_resource_quota_status_controller.go#L65-L72

According to your index function,if the work object has three labels it may be cause failure.
Eg:
workobject:
[
labels:
    "FederatedResourceQuotaNamespaceLabel":1,
    "FederatedResourceQuotaNameLabel":2,
    "ServiceNameLabel":3,
]
the index value is "FederatedResourceQuotaNamespaceLabel=1&FederatedResourceQuotaNameLabel=2&ServiceNameLabel=3",
but List function calculate the index value is "FederatedResourceQuotaNamespaceLabel=1&FederatedResourceQuotaNameLabel=2".
I'am afraid this will cause can't find the objects L in such sitaution.

cmd/agent/app/agent.go
yy158775: Combine your views, I think we don't need a new manager and only need to establish a new cache to meet @likakuli's three reasons.Because code about cache index only exists in controller-runtime/pkg/cache, don't need to edit manager.

  We can add new cache's constructor in here
https://github.com/karmada-io/karmada/blob/7cce95f96aeb1101b70190fd1744bbefc4dd42b0/cmd/controller-manager/app/controllermanager.go#L111-L118

pkg/util/manager/manager.go
likakuli: Yes. Is this possible that work both have the three labels above?

cmd/agent/app/agent.go
likakuli: Or we can directly add indexer to controllerManager using https://github.com/karmada-io/karmada/blob/master/pkg/controllers/cluster/common.go.  It's ok for me.

yy158775: But the List method still need to rewrite to avoid changes in other spaces.

likakuli: You mean we just need a custom cache to override the List func to support list by labelSelector and don't need to call `IndexFunc` to set up indexer?

yy158775: We can call IndexFunc using cache's method.Compared to calling it on manager,it does the same thing .
https://github.com/kubernetes-sigs/controller-runtime/blob/f0351217e9e026533aece74d054b23cae5441659/pkg/cache/informer_cache.go#L172-L173

Garrybest: Got it👌. We don't want to modify so many `List` callers, so we build our own cache which implement the `List` function.

Actually, I'm not sure about how much performace it will improve? I think all objects are in cache, and it seems like there have not been any perform issue so far. So is it worthy to implement our own cache and manager here? Just some questions.

yy158775: Our own cache will add index of some specific labels to optimize the List from O(n) to O(1).You can see https://github.com/kubernetes-sigs/controller-runtime/blob/f46919744bee01060c9084a285e049afffd38c9d/pkg/cache/internal/cache_reader.go#L135-L162
When List with labelselector,it use range and costs O(n).We now convert labelselector to filedselector for optimization.

yy158775: @Garrybest Do you think it is worth optimizing now? I want to finish it.

likakuli: I think it's necessary. What we need to do is thinking about when and how to do it. Also i think we should provide some best practice for users just like k8s.  How many clusters we can support? How many work we can support? e.g. Optimizing may help to increase the cap.

Garrybest: Glad to see this feature. Meanwhile, I hope we could change the basic cache or manager as little as possible. 

cc @RainbowMango who may have some ideas about this patch.

xigang:
mark

PR #2289, opened 2022-07-30
disabled search APIService chart resource by default

calvin0327:

**What type of PR is this?**
/kind bug

When use the helm cmd install karmada without any component,  but the search `APIService` is be deployed togother. 
if we want to delete a namespace in karmada cluster, the following error message will be report:

image

Because when we delete namespace, the k8s will delete all of the resource with delete verb in the namespace, incloude of  AA resource. But the search conponent is not be installed by default,  the apiserver not find the server of search.

Fixes #

Please correct me if not.

```release-note
NONE
```


RainbowMango (approved):
/approve
Thanks.

PR #2300, opened 2022-08-01
chart: add descheduler name suffix of chart deployment manifest

calvin0327:

**What type of PR is this?**
/kind bug

the descheduler missing name suffix after install with helm.

Fixes #

```release-note
NONE
```

### DIFF
--- a/charts/karmada/README.md
+++ b/charts/karmada/README.md
@@ -7,6 +7,7 @@ Karmada aims to provide turnkey automation for multi-cluster application managem
 ## TL;DR
 
 Switch to the `root` directory of the repo.
+
 ```console
 helm install karmada -n karmada-system --create-namespace --dependency-update ./charts/karmada
 ```
@@ -23,6 +24,7 @@ To install the chart with the release name `karmada` in namespace `karmada-syste
 - local installation
 
 Switch to the `root` directory of the repo.
+
 ```console
 helm install karmada -n karmada-system --create-namespace --dependency-update ./charts/karmada
 ```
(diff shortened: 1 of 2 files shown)


Poor12:
Actually, it's an issue about user habits. When users use `descheduler` as release name, it's ok. But when users use `karmada-descheduler` the issue will occur, too. I perfer to use complete name as release name, based on https://github.com/karmada-io/karmada/blob/master/charts/karmada/README.md.

calvin0327:
https://github.com/karmada-io/karmada/blob/master/charts/karmada/templates/karmada-search.yaml#L7
https://github.com/karmada-io/karmada/blob/master/charts/karmada/templates/karmada-scheduler-estimator.yaml#L8
https://github.com/karmada-io/karmada/blob/master/charts/karmada/templates/karmada-descheduler.yaml#L6

ok, I got your means. but the three components name is not consistent.

Poor12:
For this issue, I think the name in karmada-search should change to ${name}. Karmada-scheduler-estimator is a little special. It's ok to maintain the status. Could you modify the corresponding yaml？Then we can go on.

calvin0327:
https://github.com/karmada-io/karmada/pull/2507
in the pr we install karmada and karmada component with hostMode, if we delete suffix of components, the following will happen, we cannot konw what the pod is, unless we get the pod yaml.
image

Poor12:
I mean that in fact, we can't guarantee what the user enters, but we prefer the user to enter the full name when installing optional components.

Poor12:
> I mean that in fact, we can't guarantee what the user enters, but we prefer the user to enter the full name when installing optional components.
Using hostmode to install optional components does cause problems. In that case, use {name}-suffix may be a better choice.

calvin0327:
@Poor12 yes,  it's seem not problem we install optional components in component mode with `{name}-suffix`

Poor12:
Since #2507 is merged, I think we can move this pr on. Could you please update the docs by the way?
And I think we can delete the section about installing karmada-descheduler and add descriptions for installing components with host mode instead.

calvin0327:
> Since #2507 is merged, I think we can move this pr on. Could you please update the docs by the way? And I think we can delete the section about installing karmada-descheduler and add descriptions for installing components with host mode instead.

OKey,  Agree with you.

RainbowMango (approved):
/approve

PR #2302, opened 2022-08-01
add scheduling diagnosis

Garrybest:

**What type of PR is this?**
/kind feature

Add scheduling diagnosis when no clusters fit. Here is the example.

Warning  ScheduleBindingFailed   2s (x10 over 4s)  karmada-scheduler  0/3 clusters are available: 1 cluster(s) had untolerated taint {dedicated: special-user}, 2 cluster(s) didn't have the API resource.

Fixes #

```release-note
add scheduling diagnosis
```

### DIFF
--- a/pkg/scheduler/core/generic_scheduler.go
+++ b/pkg/scheduler/core/generic_scheduler.go
@@ -58,12 +58,16 @@ func (g *genericScheduler) Schedule(ctx context.Context, placement *policyv1alph
 		return result, fmt.Errorf("no clusters available to schedule")
 	}
 
-	feasibleClusters, err := g.findClustersThatFit(ctx, g.scheduleFramework, placement, spec, clusterInfoSnapshot)
+	feasibleClusters, diagnosis, err := g.findClustersThatFit(ctx, g.scheduleFramework, placement, spec, clusterInfoSnapshot)
 	if err != nil {
 		return result, fmt.Errorf("failed to findClustersThatFit: %v", err)
 	}
+
 	if len(feasibleClusters) == 0 {
-		return result, fmt.Errorf("no clusters fit")
+		return result, &framework.FitError{
+			NumAllClusters: clusterInfoSnapshot.NumOfClusters(),
+			Diagnosis:      diagnosis,
+		}
 	}
 	klog.V(4).Infof("feasible clusters found: %v", feasibleClusters)
 
@@ -97,8 +101,13 @@ func (g *genericScheduler) findClustersThatFit(
 	fwk framework.Framework,
 	placement *policyv1alpha1.Placement,
 	bindingSpec *workv1alpha2.ResourceBindingSpec,
-	clusterInfo *cache.Snapshot) ([]*clusterv1alpha1.Cluster, error) {
+	clusterInfo *cache.Snapshot,
+) ([]*clusterv1alpha1.Cluster, framework.Diagnosis, error) {
 	defer metrics.ScheduleStep(metrics.ScheduleStepFilter, time.Now())
+
+	diagnosis := framework.Diagnosis{
+		ClusterToResultMap: make(framework.ClusterToResultMap),
+	}
 
 	var out []*clusterv1alpha1.Cluster
 	// DO NOT filter unhealthy cluster, let users make decisions by using ClusterTolerations of Placement.
@@ -106,12 +115,13 @@ func (g *genericScheduler) findClustersThatFit(
 	for _, c := range clusters {
 		if result := fwk.RunFilterPlugins(ctx, placement, bindingSpec, c.Cluster()); !result.IsSuccess() {
 			klog.V(4).Infof("cluster %q is not fit, reason: %v", c.Cluster().Name, result.AsError())
+			diagnosis.ClusterToResultMap[c.Cluster().Name] = result
 		} else {
 			out = append(out, c.Cluster())
 		}
 	}
 
-	return out, nil
+	return out, diagnosis, nil
 }
 
 // prioritizeClusters prioritize the clusters by running the score plugins.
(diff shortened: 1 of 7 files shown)


pkg/scheduler/core/generic_scheduler.go
```diff
@@ -58,12 +58,16 @@ func (g *genericScheduler) Schedule(ctx context.Context, placement *policyv1alph
...
 
-	feasibleClusters, err := g.findClustersThatFit(ctx, g.scheduleFramework, placement, &spec.Resource, clusterInfoSnapshot)
+	feasibleClusters, diagnosis, err := g.findClustersThatFit(ctx, g.scheduleFramework, placement, &spec.Resource, clusterInfoSnapshot)
 	if err != nil {
 		return result, fmt.Errorf("failed to findClustersThatFit: %v", err)
 	}
+
 	if len(feasibleClusters) == 0 {
-		return result, fmt.Errorf("no clusters fit")
+		return result, &framework.FitError{
```
XiShanYongYe-Chang, line 67: I was wondering if we could also print diagnostics in general.

pkg/scheduler/framework/types.go
```diff
@@ -1,9 +1,21 @@
...
 	clusterv1alpha1 "github.com/karmada-io/karmada/pkg/apis/cluster/v1alpha1"
 )
 
+const (
+	// NoClusterAvailableMsg is used to format message when no clusters available.
+	NoClusterAvailableMsg = "0/%v clusters are available"
+)
+
+// ClusterToResultMap declares map from cluster name to its Result.
+type ClusterToResultMap map[string]*Result
```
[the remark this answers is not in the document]
Garrybest: Just a alias here, make the diagnosis more distinct when it is built.

pkg/scheduler/framework/types.go
```diff
@@ -24,3 +36,35 @@ func (n *ClusterInfo) Cluster() *clusterv1alpha1.Cluster {
...
+	NumAllClusters int
+	Diagnosis      Diagnosis
+}
+
+// Error returns detailed information of why the object failed to fit on each cluster
+func (f *FitError) Error() string {
+	reasons := make(map[string]int)
+	for _, result := range f.Diagnosis.ClusterToResultMap {
+		for _, reason := range result.Reasons() {
+			reasons[reason]++
```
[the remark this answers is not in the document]
Garrybest: Why constant? These reasons are corresponding with scheduler plugin errors.

Actually, this file is copied from https://github.com/kubernetes/kubernetes/blob/24753aa8a4df8d10bfd6330e0f29186000c018be/pkg/scheduler/framework/types.go#L214-L258.

pkg/scheduler/core/generic_scheduler.go
Garrybest: This `FitError` is also similar with https://github.com/kubernetes/kubernetes/blob/24753aa8a4df8d10bfd6330e0f29186000c018be/pkg/scheduler/schedule_one.go#L330-L337, it is only used for filter plugins. If errors happen in other steps, e.g. assigning, I think we should build and return a new type of error.

Garrybest:
PTAL, diagnosis is significant to show details when fail scheduling.

RainbowMango:
/assign 
I'm sorry for missing this.

pkg/scheduler/framework/plugins/apienablement/api_enablement.go
```diff
@@ -37,7 +37,7 @@ func (p *APIEnablement) Filter(ctx context.Context, placement *policyv1alpha1.Pl
 	bindingSpec *workv1alpha2.ResourceBindingSpec, cluster *clusterv1alpha1.Cluster) *framework.Result {
 	if !helper.IsAPIEnabled(cluster.Status.APIEnablements, bindingSpec.Resource.APIVersion, bindingSpec.Resource.Kind) {
 		klog.V(2).Infof("Cluster(%s) not fit as missing API(%s, kind=%s)", cluster.Name, bindingSpec.Resource.APIVersion, bindingSpec.Resource.Kind)
-		return framework.NewResult(framework.Unschedulable, "no such API resource")
+		return framework.NewResult(framework.Unschedulable, "cluster(s) didn't have the API resource")
```
RainbowMango, line 40: why use `(s)` here? I don't understand. It should be always singular.

pkg/scheduler/framework/plugins/clusteraffinity/cluster_affinity.go
```diff
@@ -39,7 +39,7 @@ func (p *ClusterAffinity) Filter(ctx context.Context, placement *policyv1alpha1.
 		if util.ClusterMatches(cluster, *affinity) {
 			return framework.NewResult(framework.Success)
 		}
-		return framework.NewResult(framework.Unschedulable, "cluster is not matched the placement cluster affinity constraint")
+		return framework.NewResult(framework.Unschedulable, "cluster(s) didn't match the placement cluster affinity constraint")
```
RainbowMango, line 42: ditto.

pkg/scheduler/framework/plugins/apienablement/api_enablement.go
Garrybest: For constructing the diagnosis event, like what `k/k` does. For example:

Warning ScheduleBindingFailed 2s (x10 over 4s) karmada-scheduler 0/3 clusters are available: 1 cluster(s) had untolerated taint {dedicated: special-user}, 2 cluster(s) didn't have the API resource.

Garrybest: The number of clusters could be singular or plural.

RainbowMango:
So sorry for let this sit. It is still on my TODO list. I'll try to focus on it this week.
PS:
When I look at it, I can't help to think about how to show more scheduler statuses directly to Binding objects, not just log it out.

RainbowMango (approved):
/lgtm
/approve

PR #2303, opened 2022-08-01
add ut for binding/common

AllenZMC:
**What type of PR is this?**
/kind cleanup

add ut for binding/common
Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/binding/common_test.go
+++ b/pkg/controllers/binding/common_test.go
@@ -4,7 +4,12 @@ import (
 	"reflect"
 	"testing"
 
+	v1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


XiShanYongYe-Chang (commented):
Actually, I think these two methods `mergeLabel`, `mergeAnnotations` are not easy to write ut because they both do two things, one of which is to modify the workload and the other is to return the merged result. So far, we've only tested the latter one.

RainbowMango (approved):
/lgtm
/approve

RainbowMango:
> Actually, I think these two methods `mergeLabel`, `mergeAnnotations` are not easy to write ut because they both do two things, one of which is to modify the workload and the other is to return the merged result. So far, we've only tested the latter one.

Sorry, I missed your comments. Yes, indeed, actually I'm looking at it, let's how to improve it.

XiShanYongYe-Chang:
> Sorry, I missed your comments. Yes, indeed, actually I'm looking at it, let's how to improve it.

It's okay~

PR #2311, opened 2022-08-02
optimize unstructured convert methods

weilaaa:
**What type of PR is this?**
/kind cleanup

Instead of adding a conversion function for each resource, we should replace it with a generic conversion function.

### DIFF
--- a/pkg/dependenciesdistributor/dependencies_distributor.go
+++ b/pkg/dependenciesdistributor/dependencies_distributor.go
@@ -236,23 +236,15 @@ func dependentObjectReferenceMatches(objectKey keys.ClusterWideKey, referenceBin
 
 // OnResourceBindingUpdate handles object update event and push the object to queue.
 func (d *DependenciesDistributor) OnResourceBindingUpdate(oldObj, newObj interface{}) {
-	unstructuredOldObj, ok := oldObj.(*unstructured.Unstructured)
-	if !ok {
+	oldBindingObject := &workv1alpha2.ResourceBinding{}
+	if err := helper.ConvertToTypedObject(oldObj, oldBindingObject); err != nil {
+		klog.Warningf("convert to resource binding failed: %v", err)
 		return
 	}
 
-	oldBindingObject, err := helper.ConvertToResourceBinding(unstructuredOldObj)
-	if err != nil {
-		return
-	}
-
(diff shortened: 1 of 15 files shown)


RainbowMango:
Generally looks good. We didn't use the `Golang generic feature` in this patch, right? (I didn't see it, just a confirmation).

weilaaa:
> Generally looks good. We didn't use the `Golang generic feature` in this patch, right? (I didn't see it, just a confirmation).

Not yet, when upgrading to go 1.18, consider using generics.

RainbowMango:
> Not yet, when upgrading to go 1.18, consider using generics.

I'm hesitant to introduce generics even though we are already in Go1.18. Because technically we can use generics in master(v1.3), but can't cherry-pick patches to release branches(<=release-1.2).
Basically, I agree with [Kubernetes Generics Policy](https://github.com/kubernetes/community/blob/master/sig-architecture/generics.md#generics-policy).

weilaaa:
> > Not yet, when upgrading to go 1.18, consider using generics.
> 
> I'm hesitant to introduce generics even though we are already in Go1.18. Because technically we can use generics in master(v1.3), but can't cherry-pick patches to release branches(<=release-1.2). Basically, I agree with [Kubernetes Generics Policy](https://github.com/kubernetes/community/blob/master/sig-architecture/generics.md#generics-policy).

Thanks for sharing, I read k8s generics policy and I agreed. We need to use generics sparingly.

pkg/dependenciesdistributor/dependencies_distributor.go
```diff
@@ -235,24 +235,15 @@ func dependentObjectReferenceMatches(objectKey keys.ClusterWideKey, referenceBin
 }
 
 // OnResourceBindingUpdate handles object update event and push the object to queue.
+// todo:(weilaaa) if should care about error occurs in event handler?
```
RainbowMango, line 238: We can't do anything except log the error here.
Yeah, please add a log.

RainbowMango: Same as below.

RainbowMango (commented):
Looks great! 
Let's fix the new added todo with this patch.

weilaaa:
> /lgtm
> 
> You might need to squash your commits.

PTAL again

RainbowMango (approved):
/lgtm
/approve

PR #2313, opened 2022-08-03
add ID the unique identifier for push and pull mode's cluster

yy158775:
**What type of PR is this?**
/kind feature

To add spec.id in karmada control plane's Cluster object.
refer to #2017

Fixes #

```release-note
Collect cluster's kube-system namespace uuid into Cluster object.
```

Here are Verification Results,
When use hack/local-up-karmada to start,you can find the result:
1.push mode

the same as kube-system's Namespace in Cluster. 

2.pull mode
The result is the same as above.

When upgrading,the existed Cluster objects will have spec.id field.
 

Push and Pull mode both have the same result.

### DIFF
--- a/cmd/agent/app/agent.go
+++ b/cmd/agent/app/agent.go
@@ -132,6 +132,13 @@ func run(ctx context.Context, karmadaConfig karmadactl.KarmadaConfig, opts *opti
 		ControlPlaneConfig: controlPlaneRestConfig,
 		ClusterConfig:      clusterConfig,
 	}
+
+	id, err := util.ObtainClusterID(clusterKubeClient)
+	if err != nil {
+		return err
+	}
+	registerOption.ClusterID = id
+
 	clusterSecret, impersonatorSecret, err := util.ObtainCredentialsFromMemberCluster(clusterKubeClient, registerOption)
 	if err != nil {
 		return err
@@ -323,6 +330,7 @@ func generateClusterInControllerPlane(opts util.ClusterRegisterOption) (*cluster
 		cluster.Spec.SyncMode = clusterv1alpha1.Pull
 		cluster.Spec.APIEndpoint = opts.ClusterAPIEndpoint
 		cluster.Spec.ProxyURL = opts.ProxyServerAddress
+		cluster.Spec.ID = opts.ClusterID
 		if opts.ClusterProvider != "" {
 			cluster.Spec.Provider = opts.ClusterProvider
 		}
--- a/pkg/controllers/cluster/cluster_controller.go
+++ b/pkg/controllers/cluster/cluster_controller.go
@@ -165,6 +165,18 @@ func (c *Controller) Reconcile(ctx context.Context, req controllerruntime.Reques
 
 // Start starts an asynchronous loop that monitors the status of cluster.
 func (c *Controller) Start(ctx context.Context) error {
+	// starts a goroutine to collect existed clusters' ID for upgrade scenario
+	// And can be removed in next version
+	klog.Infof("Starting collect ID of already existed clusters")
+	go func() {
+		err := c.collectIDForClusterObjectIfNeeded(ctx)
+		if err != nil {
+			klog.Errorf("Error collecting clusters' ID: %v", err)
+		}
+
+		klog.Infof("Finishing collect ID of already existed clusters")
+	}()
+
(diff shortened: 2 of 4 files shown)


XiShanYongYe-Chang:
Hi @yy158775, please sign off refer to https://github.com/karmada-io/karmada/pull/2313/checks?check_run_id=7647358240

pkg/controllers/cluster/cluster_controller.go
```diff
@@ -165,8 +165,18 @@ func (c *Controller) Reconcile(ctx context.Context, req controllerruntime.Reques
 
 // Start starts an asynchronous loop that monitors the status of cluster.
 func (c *Controller) Start(ctx context.Context) error {
+	klog.Infof("Starting collect ID of already existed clusters")
```
RainbowMango, line 168: I'd prefer to move this logic to `Reconcile`.

pkg/controllers/cluster/cluster_controller.go
```diff
@@ -563,3 +574,32 @@ func (c *Controller) taintClusterByCondition(ctx context.Context, cluster *clust
...
+		if err != nil {
+			return err
+		}
+		ns, err := clusterClient.KubeClient.CoreV1().Namespaces().Get(ctx, "kube-system", metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+		if cluster.Spec.ID == "" {
+			cluster.Spec.ID = string(ns.GetUID())
+		}
```
RainbowMango, line 598: We should check this before getting `ns`.

pkg/controllers/cluster/cluster_controller.go
RainbowMango: ignore me. We just need to update once.

RainbowMango:
/assign @XiShanYongYe-Chang 
Generally looks good, we need some comments and function rename.

cmd/agent/app/agent.go
```diff
@@ -132,6 +132,12 @@ func run(ctx context.Context, karmadaConfig karmadactl.KarmadaConfig, opts *opti
 		ControlPlaneConfig: controlPlaneRestConfig,
 		ClusterConfig:      clusterConfig,
 	}
+	ns, err := clusterKubeClient.CoreV1().Namespaces().Get(ctx, "kube-system", metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	registerOption.ClusterID = string(ns.UID)
```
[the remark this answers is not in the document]
yy158775: I'm not sure this is a reasonable change,
https://github.com/karmada-io/karmada/blob/ec84d649a9975435f58f690493e943204f9fbff6/cmd/agent/app/agent.go#L135-L140
https://github.com/karmada-io/karmada/blob/ec84d649a9975435f58f690493e943204f9fbff6/pkg/karmadactl/join.go#L180-L189
This both can also move into RegisterClusterInControllerPlane.But I don't know which will be better.
The function RegisterClusterInControllerPlane and others' ability is not very clear.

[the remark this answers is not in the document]
XiShanYongYe-Chang: I think `RegisterClusterInControllerPlane` can do all thing that needs to be done for registering the cluster so that the login in the current `run` function will be clear. This cleanup may need to be done in a separate pr.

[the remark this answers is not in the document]
yy158775: clusterKubeClient is created by clusterConfig but RegisterClusterInControllerPlane don't have clusterConfig.So it will result in a bigger change if do so.

XiShanYongYe-Chang (commented):
/lgtm

pkg/util/cluster.go
```diff
@@ -169,3 +171,12 @@ func updateCluster(controlPlaneClient karmadaclientset.Interface, cluster *clust
 
 	return newCluster, nil
 }
+
+// GetClusterID returns the given Cluster ID
+func GetClusterID(clusterKubeClient *kubernetes.Clientset) (string, error) {
```
XiShanYongYe-Chang, line 176: Recommend: `ObtainClusterID`

```
ObtainClusterID returns the cluster ID property with clusterKubeClient
```

XiShanYongYe-Chang:
/lgtm
cc @RainbowMango

RainbowMango (approved):
/approve

PR #2314, opened 2022-08-03
cleanup event in execution_controller

Garrybest:

**What type of PR is this?**
/kind cleanup

Some resource's scope is `Cluster`, so we use `klog.KObj`

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/execution/execution_controller.go
+++ b/pkg/controllers/execution/execution_controller.go
@@ -195,7 +195,7 @@ func (c *Controller) syncToClusters(clusterName string, work *workv1alpha1.Work)
 			err = c.tryUpdateWorkload(clusterName, workload)
 			if err != nil {
 				klog.Errorf("Failed to update resource(%v/%v) in the given member cluster %s, err is %v", workload.GetNamespace(), workload.GetName(), clusterName, err)
-				c.eventf(workload, corev1.EventTypeWarning, workv1alpha1.EventReasonSyncWorkFailed, "Failed to update resource(%v/%v) in the given member cluster %s, err is %v", workload.GetNamespace(), workload.GetName(), clusterName, err)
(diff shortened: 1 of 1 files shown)


pkg/controllers/execution/execution_controller.go
```diff
@@ -292,11 +292,11 @@ func (c *Controller) updateAppliedCondition(work *workv1alpha1.Work, status meta
 	})
 }
 
-func (c *Controller) eventf(object *unstructured.Unstructured, eventtype, reason, messageFmt string, args ...interface{}) {
+func (c *Controller) eventf(object *unstructured.Unstructured, eventType, reason, messageFmt string, args ...interface{}) {
 	ref, err := helper.GenEventRef(object)
 	if err != nil {
-		klog.Errorf("ignore event(%s) as failed to build event reference for: kind=%s, %s/%s due to %v", reason, object.GetKind(), object.GetNamespace(), object.GetName(), err)
+		klog.Errorf("ignore event(%s) as failed to build event reference for: kind=%s, %s due to %v", reason, object.GetKind(), klog.KObj(object), err)
```
duanmengkk, line 298: Should we do the same way in other places?

XiShanYongYe-Chang: I have the same doubts and often see this message in the log: `(/xxx) `. Perhaps we can create a work issue that records the modification of similar problems in all code modules.

RainbowMango (approved):
/lgtm
/approve

pkg/controllers/execution/execution_controller.go
RainbowMango: Good question.
I have an idea in my mind to implement one function for generating the `key` just for logging.
I'll try to do this and make a benchmark with `Klog`.

PR #2315, opened 2022-08-04
stop update status if Work object in terminating state

yy158775:
Signed-off-by: yy158775 <1584616775@qq.com>

**What type of PR is this?**
/kind feature

When the object's status changed,it will not collect status if the Work object being deleted. 

Fixes #

NONE
```release-note
NONE
```
Here are my tests:
Step 1:
kubectl apply -f samples/nginx/propagationpolicy.yaml
kubectl apply -f samples/nginx/deployment.yaml
Step 2:
Add finalizer to Work object in order to avoid being deleted.
step-1
Step 3:
Add finalizer to Deployment object in order to avoid being deleted.
step-3
Step 4:
Delete Work.Because of finalizer Work and Deployment will not  be deleted.
Step 5:
In order to make status changed,we can delete Pod.Because Pod is in charge of ReplicaSet,the Pod will restart immediately.
Now we look up logs in controller-manager,we can find the log about update.
代码
result

### DIFF
--- a/pkg/controllers/status/workstatus_controller.go
+++ b/pkg/controllers/status/workstatus_controller.go
@@ -194,6 +194,11 @@ func (c *WorkStatusController) syncWorkStatus(key util.QueueKey) error {
 		return err
 	}
 
+	// stop update status if Work object in terminating state.
+	if !workObject.DeletionTimestamp.IsZero() {
+		return nil
+	}
+
 	desiredObj, err := c.getRawManifest(workObject.Spec.Workload.Manifests, observedObj)
 	if err != nil {
 		return err


pkg/controllers/status/workstatus_controller.go
```diff
@@ -194,6 +194,11 @@ func (c *WorkStatusController) syncWorkStatus(key util.QueueKey) error {
 		return err
 	}
 
+	// stop update status as the work object being deleting.
```
RainbowMango, line 197:
```suggestion
	// stop update status if Work object in terminating state
```

RainbowMango:
/retitle stop update status if Work object in terminating state

RainbowMango (commented):
/lgtm
/approve

PR #2318, opened 2022-08-04
reactor: use the version informer instead of dynamic

Charlie17Li:
Signed-off-by: charlie <qianglin98@qq.com>

**What type of PR is this?**
/kind cleanup

use the versiond informer instead of dynamic when sync pod and node from member cluster, because when using dynamic client, there are more resource(cpu,mem) to serialization 

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/util/informermanager/single-cluster-manager.go
+++ b/pkg/util/informermanager/single-cluster-manager.go
@@ -2,12 +2,16 @@ package informermanager
 
 import (
 	"context"
+	"fmt"
 	"sync"
 	"time"
 
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/dynamic/dynamicinformer"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	v1 "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
 
 	"github.com/karmada-io/karmada/pkg/util"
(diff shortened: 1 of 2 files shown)


Charlie17Li:
cc @Poor12

pkg/util/informermanager/single-cluster-manager.go
```diff
@@ -54,6 +58,12 @@ type SingleClusterInformerManager interface {
 
 	// GetClient returns the dynamic client.
 	GetClient() dynamic.Interface
+
+	ListerWithNode() v1.NodeLister
```
duanmengkk, line 62: Should we put the lister of node and pod to `single-cluster-manager.go` ,It is a util for dynamic informer. This seems a little strange.

Charlie17Li: Yes, I agree, but In a large-scale scenes (e.g. 50w Nodes & 200w Pods), The Dynamic informer Sync the Resource (especially Nodes and Pods) will consume a lot of memory and CPU, I have tried my best not to modify the original interface and method(because many controllers use the `single-cluster-manager`), but it does look strange now. Could you please give me some advice, thx.

duanmengkk: Just a questiton and I have no objection to it, may be we can ask @Poor12 for suggestion

duanmengkk: If we only need lister of node and pod in `cluster_status_controller`,we can put listers in it

Poor12:
This PR is completed by #2350. Thanks @Charlie17Li.
/close

PR #2326, opened 2022-08-04
Fix "namespace \"karmada-cluster"\ not found" when cluster is being registered with "pull" mode

nuclearwu:
Fix "namespace \"karmada-cluster"\ not found" when cluster is being registered with "pull" mode

**What type of PR is this?**
/kind bug

Fixes #

NONE


nuclearwu:
https://github.com/karmada-io/karmada/pull/2327

PR #2327, opened 2022-08-04
Fix "namespace \"karmada-cluster"\ not found" when cluster is being registered with "pull" mode

nuclearwu:
Fix "namespace \"karmada-cluster"\ not found" when cluster is being registered with "pull" mode

Signed-off-by: wuzhongjian <wuzhongjian_yewu@cmss.chinamobile.com>

**What type of PR is this?**
/kind bug

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/karmadactl/cmdinit/karmada/deploy.go
+++ b/pkg/karmadactl/cmdinit/karmada/deploy.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	options2 "github.com/karmada-io/karmada/pkg/karmadactl/options"
 	"io/ioutil"
 	"path"
 	"path/filepath"
@@ -24,6 +25,7 @@ import (
 
 	"github.com/karmada-io/karmada/pkg/karmadactl/cmdinit/options"
 	"github.com/karmada-io/karmada/pkg/karmadactl/cmdinit/utils"
+	options2 "github.com/karmada-io/karmada/pkg/karmadactl/options"
 )
 
 const (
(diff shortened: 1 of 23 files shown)


nuclearwu:
Most people use karmada to add clusters through push mode, and then consider adding clusters through pull mode, but the karmada-cluster namespace is created by default only when clusters are added through push mode, so we have not found this problem. Including the official documents of karmada, it is also said that two clusters are added through the push mode first, and then a cluster is added through the pull mode. So when you add a cluster directly through pull mode, you will find that the karmada-cluster namespace is not created by default, which will cause this error.

pkg/karmadactl/cmdinit/karmada/deploy.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	options2 "github.com/karmada-io/karmada/pkg/karmadactl/options"
```
wuyingjun-lucky, line 7: We organize the import package in three parts:
`go package

third-party package

current project package`

wuyingjun-lucky:
@RainbowMango  help to active the pipeline

wuyingjun-lucky:
@nuclearwu 
You can use 
[squash](https://github.com/karmada-io/karmada/blob/master/docs/contributors/guide/github-workflow.md#squash-commits) to squash commits

nuclearwu:
https://github.com/karmada-io/karmada/pull/2328 @wuyingjun-lucky

RainbowMango:
@nuclearwu You don't need to close and reopen a new one. Just following the GitHub workflow would be fine.
https://karmada.io/docs/next/contributor/github-workflow

PR #2328, opened 2022-08-04
fix karmada-agent can not ensure default namespace issue

nuclearwu:
Fix "namespace "karmada-cluster" not found" when cluster is being registered with "pull" mode

Signed-off-by: wuzhongjian <wuzhongjian_yewu@cmss.chinamobile.com>

**What type of PR is this?**
/kind bug

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/karmadactl/join.go
+++ b/pkg/karmadactl/join.go
@@ -160,11 +160,6 @@ func JoinCluster(controlPlaneRestConfig, clusterConfig *rest.Config, opts Comman
 
 	klog.V(1).Infof("joining cluster config. endpoint: %s", clusterConfig.Host)
 
-	// ensure namespace where the cluster object be stored exists in control plane.
-	if _, err = util.EnsureNamespaceExist(controlPlaneKubeClient, opts.ClusterNamespace, opts.DryRun); err != nil {
-		return err
-	}
-
 	registerOption := util.ClusterRegisterOption{
 		ClusterNamespace:   opts.ClusterNamespace,
 		ClusterName:        opts.ClusterName,
--- a/pkg/util/credential.go
+++ b/pkg/util/credential.go
@@ -98,6 +98,11 @@ func ObtainCredentialsFromMemberCluster(clusterKubeClient kubeclient.Interface, 
 
 // RegisterClusterInControllerPlane represents register cluster in controller plane
 func RegisterClusterInControllerPlane(opts ClusterRegisterOption, controlPlaneKubeClient kubeclient.Interface, generateClusterInControllerPlane generateClusterInControllerPlaneFunc) error {
+	// ensure namespace where the cluster object be stored exists in control plane.
+	if _, err := EnsureNamespaceExist(controlPlaneKubeClient, opts.ClusterNamespace, opts.DryRun); err != nil {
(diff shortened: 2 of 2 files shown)


nuclearwu:
Most people use karmada to add clusters through push mode, and then consider adding clusters through pull mode, but the karmada-cluster namespace is created by default only when clusters are added through push mode, so we have not found this problem. Including the official documents of karmada, it is also said that two clusters are added through the push mode first, and then a cluster is added through the pull mode. So when you add a cluster directly through pull mode, you will find that the karmada-cluster namespace is not created by default, which will cause this error.

XiShanYongYe-Chang (commented):
Thanks for pointing out the error. The current solution may be can not solve the problem at all. We can modify the logic in the function `run` of the file `cmd/agent/app/agent.go`, just like: https://github.com/karmada-io/karmada/blob/54752ed6260825e76b7ee6b56aad616174dcd0a6/pkg/karmadactl/join.go#L163-L166

nuclearwu:
> Thanks for pointing out the error. The current solution may be can not solve the problem at all. We can modify the logic in the function `run` of the file `cmd/agent/app/agent.go`, just like:
> 
> https://github.com/karmada-io/karmada/blob/54752ed6260825e76b7ee6b56aad616174dcd0a6/pkg/karmadactl/join.go#L163-L166

Ok, let me modify it

nuclearwu:
> Thanks for pointing out the error. The current solution may be can not solve the problem at all. We can modify the logic in the function `run` of the file `cmd/agent/app/agent.go`, just like:
> 
> https://github.com/karmada-io/karmada/blob/54752ed6260825e76b7ee6b56aad616174dcd0a6/pkg/karmadactl/join.go#L163-L166

I have done, please review

XiShanYongYe-Chang (commented):
Thanks for your quick updateion! I have a idea: we can move this ensure logic into the `RegisterClusterInControllerPlane` function, if so, we are able to combine the two codes into one.

By the way, have you ever encountered the same errors that you had before since the modification?

nuclearwu:
> Thanks for your quick updateion! I have a idea: we can move this ensure logic into the `RegisterClusterInControllerPlane` function, if so, we are able to combine the two codes into one.
> 
> By the way, have you ever encountered the same errors that you had before since the modification?

Yes, I rebuilt the karmada-agent image and registered the member cluster through pull mode, this error did not happen again

XiShanYongYe-Chang:
> Yes, I rebuilt the karmada-agent image and registered the member cluster through pull mode, this error did not happen again

Ok, thanks~

RainbowMango (approved):
/retitle fix karmada-agent can not ensure default namespace issue
/approve

XiShanYongYe-Chang:
This pr may need to cherry-pick before release-branch.

nuclearwu:
> This pr may need to cherry-pick before release-branch.

ok. I will do it

nuclearwu:
> This pr may need to cherry-pick before release-branch.

I checked the release-1.2.0 branch and it doesn't have this problem

XiShanYongYe-Chang:
Thanks~

PR #2334, opened 2022-08-05
stop publish duplicated event

Garrybest:

**What type of PR is this?**
/kind cleanup

Stop publish duplicated event when there is nothing to change.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/util/helper/workstatus.go
+++ b/pkg/util/helper/workstatus.go
@@ -7,12 +7,14 @@ import (
 	"reflect"
 	"sort"
 
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -36,10 +38,9 @@ const (
 
 // AggregateResourceBindingWorkStatus will collect all work statuses with current ResourceBinding objects,
 // then aggregate status info to current ResourceBinding status.
-func AggregateResourceBindingWorkStatus(c client.Client, binding *workv1alpha2.ResourceBinding, workload *unstructured.Unstructured) error {
+func AggregateResourceBindingWorkStatus(c client.Client, binding *workv1alpha2.ResourceBinding, workload *unstructured.Unstructured, eventRecorder record.EventRecorder) error {
 	workList, err := GetWorksByBindingNamespaceName(c, binding.Namespace, binding.Name)
 	if err != nil {
-		klog.Errorf("Failed to get works by ResourceBinding(%s/%s): %v", binding.Namespace, binding.Name, err)
 		return err
 	}
 
@@ -58,7 +59,7 @@ func AggregateResourceBindingWorkStatus(c client.Client, binding *workv1alpha2.R
 		return nil
 	}
 
-	return retry.RetryOnConflict(retry.DefaultRetry, func() (err error) {
+	err = retry.RetryOnConflict(retry.DefaultRetry, func() (err error) {
 		binding.Status = *currentBindingStatus
 		updateErr := c.Status().Update(context.TODO(), binding)
 		if updateErr == nil {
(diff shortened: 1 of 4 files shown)


RainbowMango:
/assign @Poor12 
Please take a look.

pkg/util/helper/workstatus.go
```diff
@@ -36,16 +37,16 @@ const (
 
 // AggregateResourceBindingWorkStatus will collect all work statuses with current ResourceBinding objects,
 // then aggregate status info to current ResourceBinding status.
-func AggregateResourceBindingWorkStatus(c client.Client, binding *workv1alpha2.ResourceBinding, workload *unstructured.Unstructured) error {
+func AggregateResourceBindingWorkStatus(c client.Client, binding *workv1alpha2.ResourceBinding, workload *unstructured.Unstructured) (controllerutil.OperationResult, error) {
```
RainbowMango, line 40: I think the `OperationResult` is dedicated to presenting the result of a `CreateOrUpdate call`, I'm afraid it would lead to confusing. 
Let me see how we can improve it here.

Poor12: At present, the main distinction is the following three states, one is that the update fails, the second is that the state has not changed, and finally the state is changed and the update is successful. Based on this, should we introduce a bool variable to identify it?

Garrybest: Any ideas?

Garrybest: Actually I think the `controllerutil.OperationResult` is distinct instead of confusing. If return bool, it's more confusing from my point of view😅

Garrybest: > My concern is the controllerutil.OperationResult is specific to CreateOrUpdate.

Got it. So should we create our own types here? I'm OK with the suggestion.

RainbowMango: Hold on for a while. I'm looking at it.

RainbowMango: Maybe we can pass the `` to `AggregateResourceBindingWorkStatus` which knows when and how to fire the event:
```diff
-       err = helper.AggregateResourceBindingWorkStatus(c.Client, binding, workload)                                 
+       err = helper.AggregateResourceBindingWorkStatus(c.Client, binding, workload, c.EventRecorder)
```

By the way, the two places don't need to fire events but reply on retry.
https://github.com/karmada-io/karmada/blob/10990cd96c1d1dd20ee3e2a44dbd58145f529a1b/pkg/util/helper/workstatus.go#L40-L49

Garrybest: Do you mean pass event recorder to this function? I'm afraid there are more functions need to be updated too, like `ensureWork` and `ApplyPolicy`.

> By the way, the two places don't need to fire events but reply on retry.

Return `controllerutil.OperationResultNone` does not change the events behavior in this PR. When err is not nil, we still fire events in `binding_controller.go`.

```go
	operationResult, err := helper.AggregateResourceBindingWorkStatus(c.Client, binding, workload)
	if err != nil {
		klog.Errorf("Failed to aggregate workStatuses to resourceBinding(%s/%s). Error: %v.",
			binding.GetNamespace(), binding.GetName(), err)
		c.EventRecorder.Event(binding, corev1.EventTypeWarning, workv1alpha2.EventReasonAggregateStatusFailed, err.Error())
		c.EventRecorder.Event(workload, corev1.EventTypeWarning, workv1alpha2.EventReasonAggregateStatusFailed, err.Error())
		errs = append(errs, err)
	} else if operationResult != controllerutil.OperationResultNone {
		msg := fmt.Sprintf("Update resourceBinding(%s/%s) with AggregatedStatus successfully.", binding.Namespace, binding.Name)
		klog.V(4).Infof(msg)
		c.EventRecorder.Event(binding, corev1.EventTypeNormal, workv1alpha2.EventReasonAggregateStatusSucceed, msg)
		c.EventRecorder.Event(workload, corev1.EventTypeNormal, workv1alpha2.EventReasonAggregateStatusSucceed, msg)
	}
```

Garrybest: It makes sense. I have made a change, left `[ensureWork](https://github.com/karmada-io/karmada/blob/2a0610eb0053dac9aa2688935fde3926c2b7db8a/pkg/controllers/binding/binding_controller.go#L116)` for refactoring in the future.

pkg/util/helper/workstatus.go
```diff
@@ -75,14 +76,23 @@ func AggregateResourceBindingWorkStatus(c client.Client, binding *workv1alpha2.R
...
+	return nil
 }
 
 // AggregateClusterResourceBindingWorkStatus will collect all work statuses with current ClusterResourceBinding objects,
 // then aggregate status info to current ClusterResourceBinding status.
-func AggregateClusterResourceBindingWorkStatus(c client.Client, binding *workv1alpha2.ClusterResourceBinding, workload *unstructured.Unstructured) error {
+func AggregateClusterResourceBindingWorkStatus(c client.Client, binding *workv1alpha2.ClusterResourceBinding, workload *unstructured.Unstructured, eventRecorder record.EventRecorder) error {
 	workList, err := GetWorksByBindingNamespaceName(c, "", binding.Name)
 	if err != nil {
-		klog.Errorf("Failed to get works by ClusterResourceBinding(%s): %v", binding.Name, err)
```
RainbowMango, line 85: ditto.

pkg/util/helper/workstatus.go
```diff
@@ -36,10 +38,9 @@ const (
 
 // AggregateResourceBindingWorkStatus will collect all work statuses with current ResourceBinding objects,
 // then aggregate status info to current ResourceBinding status.
-func AggregateResourceBindingWorkStatus(c client.Client, binding *workv1alpha2.ResourceBinding, workload *unstructured.Unstructured) error {
+func AggregateResourceBindingWorkStatus(c client.Client, binding *workv1alpha2.ResourceBinding, workload *unstructured.Unstructured, eventRecorder record.EventRecorder) error {
 	workList, err := GetWorksByBindingNamespaceName(c, binding.Namespace, binding.Name)
 	if err != nil {
-		klog.Errorf("Failed to get works by ResourceBinding(%s/%s): %v", binding.Namespace, binding.Name, err)
```
RainbowMango, line 42: Is there any specific reason you removed this log?
I suppose a whole failure path would help debug.

RainbowMango (commented):
Looks good to me, just one question.

pkg/util/helper/workstatus.go
Garrybest: This log is duplicated because the return err will be printed by `syncBinding`. BTW, other errors are not printed in `AggregateResourceBindingWorkStatus`.

RainbowMango: In my experience, sometimes when I want to trace an error log, I want to find the underlying function which first raised it, it will be difficult, especially in many layers. 

Never mind. no big deal.

RainbowMango (approved):
/lgtm
/approve

PR #2343, opened 2022-08-07
add ut for cmdinit

AllenZMC:
**What type of PR is this?**
/kind cleanup

add ut for cmdinit
Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/karmadactl/cmdinit/utils/util_test.go
+++ b/pkg/karmadactl/cmdinit/utils/util_test.go
@@ -2,6 +2,7 @@ package utils
 
 import (
 	"os"
+	"reflect"
 	"testing"
 )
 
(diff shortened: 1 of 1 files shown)


AllenZMC:
@XiShanYongYe-Chang @RainbowMango

pkg/karmadactl/cmdinit/utils/util_test.go
```diff
@@ -24,3 +25,24 @@ func TestDownloadFile(t *testing.T) {
 		panic(err)
 	}
 }
+
+func TestListFiles(t *testing.T) {
+	tests := []struct {
+		name string
+		path string
+		want []string
+	}{
```
XiShanYongYe-Chang, line 34: What happens to input a path that does not exist?

pkg/karmadactl/cmdinit/utils/util_test.go
```diff
@@ -24,3 +25,24 @@ func TestDownloadFile(t *testing.T) {
...
+		{
+			name: "files",
+			path: "../options",
+			want: []string{"../options/global.go"},
```
XiShanYongYe-Chang, line 38: The result of this test case will depend on the file in the `../options` path.

AllenZMC: what good advice do you have @XiShanYongYe-Chang

pkg/karmadactl/cmdinit/utils/util_test.go
AllenZMC: Empty path will panic

pkg/karmadactl/cmdinit/utils/util_test.go
XiShanYongYe-Chang: I don't have any ideas yet.

pkg/karmadactl/cmdinit/utils/util_test.go
XiShanYongYe-Chang: Does that have to be taken into account in the method?

PR #2345, opened 2022-08-08
 patch resourcebinding instead of update and update observed generation.

yy158775:
**What type of PR is this?**
/kind feature

Fixes #

```release-note
patch resourcebinding and clusterresourcebinding instead of update and update observed generation.
add test for them
```

### DIFF
--- a/pkg/scheduler/scheduler_test.go
+++ b/pkg/scheduler/scheduler_test.go
@@ -1,13 +1,19 @@
 package scheduler
 
 import (
+	"context"
+	"reflect"
 	"testing"
+	"time"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	dynamicfake "k8s.io/client-go/dynamic/fake"
 	"k8s.io/client-go/kubernetes/fake"
 
+	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
 	karmadafake "github.com/karmada-io/karmada/pkg/generated/clientset/versioned/fake"
+	"github.com/karmada-io/karmada/pkg/util"
 )
 
 func TestCreateScheduler(t *testing.T) {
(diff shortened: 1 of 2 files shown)


pkg/scheduler/scheduler_test.go
```diff
@@ -63,3 +69,211 @@ func TestCreateScheduler(t *testing.T) {
...
+			err = scheduler.patchBindingScheduleStatus(test.binding, test.newScheduledCondition)
+			if err != nil {
+				t.Error(err)
+			}
+			res, err := karmadaClient.WorkV1alpha2().ResourceBindings(test.binding.Namespace).Get(context.TODO(), test.binding.Name, metav1.GetOptions{})
+			if err != nil {
+				t.Error(err)
+			}
+			if !reflect.DeepEqual(res, test.expected) {
+				t.Errorf("Test %s Not Equal", test.name)
```
RainbowMango, line 171: We don't need to log the test name as the name is already present in the result.
For example(I hacked the tests and let it fails) :
```
$ go test ./pkg/scheduler/... -run=Test_patchBindingScheduleStatus -v
=== RUN   Test_patchBindingScheduleStatus
=== RUN   Test_patchBindingScheduleStatus/add_success_condition
=== RUN   Test_patchBindingScheduleStatus/add_failure_condition
=== RUN   Test_patchBindingScheduleStatus/replace_to_success_condition
=== RUN   Test_patchBindingScheduleStatus/replace_failure_condition
    scheduler_test.go:171: Test replace failure condition Not Equal
```
The `Test_patchBindingScheduleStatus/replace_failure_condition` is exactly the test name.

pkg/scheduler/scheduler_test.go
```diff
@@ -63,3 +69,211 @@ func TestCreateScheduler(t *testing.T) {
...
+			}
```
RainbowMango, line 172:
```suggestion
			if !reflect.DeepEqual(res.Status, test.expected.Status) {
				t.Errorf("expected status: %v, but got: %v", test.expected.Status, res.Status)
			}
```

pkg/scheduler/scheduler_test.go
```diff
@@ -63,3 +69,211 @@ func TestCreateScheduler(t *testing.T) {
...
+			err = scheduler.patchBindingScheduleStatus(test.binding, test.newScheduledCondition)
+			if err != nil {
+				t.Error(err)
+			}
+			res, err := karmadaClient.WorkV1alpha2().ResourceBindings(test.binding.Namespace).Get(context.TODO(), test.binding.Name, metav1.GetOptions{})
+			if err != nil {
+				t.Error(err)
+			}
+			if !reflect.DeepEqual(res.Status, test.expected.Status) {
+				t.Errorf("Not Equal")
```
RainbowMango, line 171: https://github.com/karmada-io/karmada/pull/2345#discussion_r939993743

pkg/scheduler/scheduler_test.go
```diff
@@ -63,3 +69,211 @@ func TestCreateScheduler(t *testing.T) {
...
+				Status:     workv1alpha2.ResourceBindingStatus{Conditions: []metav1.Condition{failureCondition}},
+			},
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			_, err := karmadaClient.WorkV1alpha2().ResourceBindings(test.binding.Namespace).Create(context.TODO(), test.binding, metav1.CreateOptions{})
+			if err != nil {
+				t.Error(err)
```
RainbowMango, line 160: We should use `Fatal` or `Fatalf` here since if there is an error, we can skip following checks.

RainbowMango (commented):
/lgtm

RainbowMango (approved):
/approve

PR #1802, opened 2022-05-16
Add karmada-search fabric

XiShanYongYe-Chang:
Signed-off-by: changzhen <changzhen5@huawei.com>

**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

Add `karmada-search` component base structure.

**Which issue(s) this PR fixes**:
Part of #1794 

**Special notes for your reviewer**:

This PR changes the implementation of `ResourceRegistry` from CRD to AA.

The search API need to in such format: `/apis/search.karmada.io/v1alpha1/search/xxx`, I'm using `search` for now, we need to think about what word to use for this location.

@GitHubxsy @RainbowMango 

Currently, the `helm` and `karmadactl init` installation modes are not adapted.

**Does this PR introduce a user-facing change?**:

```release-note
Add karmada-search fabric.
```

**How to test it**:

We have deployed `karmada-search` with `hack/local-up-karmada.sh` for current PR. 

We can apply the following yaml:

```yaml
apiVersion: search.karmada.io/v1alpha1
kind: ResourceRegistry
metadata:
  name: test
spec:
  clusterSelectors:
    - clusterNames:
      - member1
  resourceSelectors:
    - apiVersion: apps/v1
      kind: Deployment
```

For the search URL, you can run the command `kubectl get --raw /apis/search.karmada.io/v1alpha1/search/xxx`.

PR #2366, opened 2022-08-11
add proxy skeleton

ikaven1024:

**What type of PR is this?**
/kind api-change
/kind feature

Implement for proxy

Fixes #

```release-note
users can request global resource by api: /apis/search.karmada.io/v1alpha1/search/proxying/karmada/proxy/...

e.g. list all the pods across clusters:
kubectl get --raw /apis/search.karmada.io/v1alpha1/search/proxying/karmada/proxy/api/v1/pods
```


ikaven1024:
@RainbowMango is the api is appropriate?

RainbowMango:
cc @XiShanYongYe-Chang 
I'll look at it later.

XiShanYongYe-Chang:
ok

pkg/proxy/controller.go
```diff
@@ -0,0 +1,132 @@
...
+	workerOptions := util.Options{
+		Name:          "proxy-controller",
+		KeyFunc:       nil,
+		ReconcileFunc: ctl.reconcile,
+	}
+	ctl.worker = util.NewAsyncWorker(workerOptions)
+
+	resourceEventHandler := cache.ResourceEventHandlerFuncs{
+		AddFunc: func(interface{}) {
+			ctl.worker.Add(workKey)
```
RainbowMango, line 67: I don't get why just add a `key` here, since the reconcile function is empty now, I can't see how to reconcile the key.

pkg/registry/search/storage/search.go
```diff
@@ -50,7 +54,7 @@ func (r *SearchREST) NamespaceScoped() bool {
 
 // ConnectMethods returns the list of HTTP methods handled by Connect.
 func (r *SearchREST) ConnectMethods() []string {
-	return []string{"GET"}
+	return proxyMethods
```
RainbowMango, line 57: Does this mean `search` api starts to support `"POST", "PUT", "PATCH", "DELETE", "HEAD", "OPTIONS"`?

ikaven1024: Proxy supports all the methods, so I add them here.
If the original search not support these methods,  shall add logic
```go
if sets.NewString("POST", "PUT", "PATCH", "DELETE", "HEAD", "OPTIONS").Has(method) {
    return apierrors.NewMethodNotSupported(gvr, method)
}
```

pkg/proxy/controller.go
ikaven1024: I will show you the code in next pr.

pkg/registry/search/storage/search.go
```diff
@@ -79,6 +83,14 @@ func (r *SearchREST) Connect(ctx context.Context, id string, _ runtime.Object, r
 		klog.Errorf("Failed to parse k8s native RequestInfo, err: %v", err)
 		return nil, err
 	}
+
+	if id == "proxy" {
```
XiShanYongYe-Chang, line 87: It the last URL is in this format: `xxxx/proxy/apis/v1/…`, do we need to create a `proxy` resource API like: https://github.com/karmada-io/karmada/blob/c5ce2af608bcae8f0d97f36d0d9755d7986dbf04/pkg/apis/search/searchregistry_types.go#L104-L107

Otherwise, as currently written, the final URL will look like this: `xxxx/search/proxy/apis/v1/…`.

ikaven1024: Yeah, current url is `xxx/search/<id>/apis/v1/...`

When `id` is:
- proxy: direct to proxy
- cache: search resource from cache
- opensearch: search resources from opensearch

I prefers this.
But the `search` name is not appropriate, we may change it later.

---------
While creating above API, url will be `xxx/proxy/<id>/apis/v1/...`. Then what the id is?

---------

What do you think, @RainbowMango @XiShanYongYe-Chang

ikaven1024: TBD:
- API Group: rename `search.karmada.io`?
- Resource: reuse `search` or add a new resource API

ikaven1024: > If the last URL is in this format: `xxxx/proxy/apis/v1/…`

Hi, @XiShanYongYe-Chang how to create an API, generating above url, discarding the `id`?

XiShanYongYe-Chang: - Add a Proxy struct in the APIs files `pkg/apis/search`, and then run `make update` to generate the related file.
- Add the registry strategy file like search `pkg/registry/search`.
- Register the proxy resource in here: https://github.com/karmada-io/karmada/blob/3ddee004adb21bf20b9a0807c59c015c2e28ecf5/pkg/search/apiserver.go#L86

You can look at the `Search` API creation here: https://github.com/karmada-io/karmada/pull/1802

ikaven1024: I created `Proxy` API.
But when I type
```
kubectl get --raw "/apis/search.karmada.io/v1alpha1/proxy/api/v1/nodes" --v=6
I0816 18:28:30.018509   55471 helpers.go:222] server response object: [{
  "metadata": {},
  "status": "Failure",
  "message": "the server could not find the requested resource",
  "reason": "NotFound",
  "details": {
    "causes": [
      {
        "reason": "UnexpectedServerResponse",
        "message": "404 page not found"
      }
    ]
  },
  "code": 404
}]
```

When i insert a name `foo`, it works:
```
kubectl get --raw "/apis/search.karmada.io/v1alpha1/proxy/foo/api/v1/nodes"
{"kind":"NodeList","apiVersion":"v1","metadata":{"resourceVersion":"1032221"},"items":[]}
```

It seems can not discarding the name. Any wrong with it @XiShanYongYe-Chang ? I have pushed this code.

XiShanYongYe-Chang: Dose discarding the id work? I received your reply in the email, but I didn't find it in the issue.

ikaven1024: I created Proxy API (code is pushed)
Now i can request with: `/apis/search.karmada.io/v1alpha1/proxy/api/v1/nodes`

But when requesting `/apis/search.karmada.io/v1alpha1/proxy/api` returns error:
```
➜  ~ kubectl get --raw "/apis/search.karmada.io/v1alpha1/proxy/api" --v=6
I0816 19:26:15.147357   59415 round_trippers.go:553] GET https://127.0.0.1:5443/apis/search.karmada.io/v1alpha1/proxy/api 400 Bad Request in 61 milliseconds
I0816 19:26:15.149290   59415 helpers.go:222] server response object: [{
  "metadata": {},
  "status": "Failure",
  "message": "name must be provided",
  "reason": "BadRequest",
  "code": 400
}]
Error from server (BadRequest): name must be provided
```

This error is threw here 
https://github.com/karmada-io/karmada/blob/614e28508336d6c03a938ce1bf0678dafef034f0/vendor/k8s.io/apiserver/pkg/endpoints/handlers/namer.go#L63-L65

ikaven1024: kubectl get no will call `/api` before get nodes

XiShanYongYe-Chang: How about change the method `NewConnectOptions`:
```
func (r *ProxyREST) NewConnectOptions() (runtime.Object, bool, string) {
	return nil, false, ""
}
```

ikaven1024: I solved this question by rename `proxy` to `proxying`. As `proxy` is a special verb.
https://github.com/karmada-io/karmada/blob/b0c6c2253c1099d8c1c84b91f3c430924a75ee91/vendor/k8s.io/apiserver/pkg/endpoints/request/requestinfo.go#L71

XiShanYongYe-Chang: ok 👍

XiShanYongYe-Chang (commented):
Thanks~
/lgtm
/approve

XiShanYongYe-Chang:
The release-note may need to be updated.

RainbowMango (approved):
/approve

Issue #772, opened 2021-09-28
[lfx-mentorship-2022-summer]Cluster Resource modeling


RainbowMango:
**What would you like to be added**:
We don't want to collect and store each node's resources in detail(That's a burden for Karmada to maintain the information), but we want to build a `resource model` for each cluster, something like:
```yaml
resourceModel:
  - grade:
      cpu: "1"
      memory: 2Gi
    count: 10
  - grade:
      cpu: "2"
      memory: 4Gi
    count: 6
  - grade:
      cpu: "4"
      memory: 8Gi
    count: 2
  - grade:
      cpu: "8"
      memory: 16Gi
    count: 1
```

**Why is this needed**:
In the scheduling progress, the `karmada-scheduler` makes decisions as per a bunch of factors, one of the factors is the resource details of the cluster.

We introduced [ResourceSummary](https://github.com/karmada-io/karmada/blob/b5af525b1969ee6040c27e0daefcac3c59c84d48/pkg/apis/cluster/types.go#L141) to the [Cluster API](https://github.com/karmada-io/karmada/blob/b5af525b1969ee6040c27e0daefcac3c59c84d48/pkg/apis/cluster/types.go#L15). For example:
```yaml
  resourceSummary:
    allocatable:
      cpu: "4"
      ephemeral-storage: 206291924Ki
      hugepages-1Gi: "0"
      hugepages-2Mi: "0"
      memory: 16265856Ki
      pods: "110"
    allocated:
      cpu: 950m
      memory: 290Mi
      pods: "11"
```
But the `ResourceSummary` is not precise enough, it mechanically counts the resources on all nodes, but ignores the fragment resources.(For example, a cluster with 2000 node, 1 core cpu left on each node, from the `ResourceSummary`, we get there are 2000 core CPU left for the cluster, that's not correct.)

## References
- LFX Mentorship: https://github.com/cncf/mentoring/tree/main/lfx-mentorship
- timeline: https://github.com/cncf/mentoring/tree/main/lfx-mentorship/2022/02-Summer#timeline

anutosh491:
Hello @RainbowMango  , hope you're doing well , this is Anutosh here from India . I'm an open source enthusiast and I'm currently involved with communities based on numerical and symbolic computations/algorithms in math and physics like numpy, sympy, networkx.

[...]

PR #2367, opened 2022-08-12
Cluster resource modeling

halfrost:
What type of PR is this?
/kind feature

This PR introduced a feature of cluster resource modeling

Which issue(s) this PR fixes:
Part of https://github.com/karmada-io/karmada/issues/772

This pr is for issue https://github.com/karmada-io/karmada/issues/772 cluster resource modeling code implementation part.

None

My solution is the picture below:

The system defaults to 11 grade levels. The grading of each grade is as shown above. For example, the resource configuration of grade =2 model is as follows

```
resourceModel:
  - grade:
      cpu: "2"
      memory: 4Gi
    count: 34
```

grade = 2, which means cpu belongs to (2C, 4C], memory belongs to (2G, 4G], left open and right closed.

Users can also customize modeling. The user passes in a custom modeling. During initialization, it will be initialized according to the user's configuration. Users can customize multiple levels. Such as cpu, memory, gpu and so on. The system will sort them in order of priority. For example, if the user defines the order of cpu, memoryu, gpu. Then when comparing, the CPU will be compared first. If the cpu is the same, then compare the memory, and finally compare the gpu. modeling in the order of this comparison

Implementation:

The data structure shown in the figure is maintained in a one-dimensional array. There are 3 values ​​stored in item in the array, one value is count. This value indicates how many there are in this model. In addition, 2 pointers are stored. They point to a doubly linked list and a red-black tree, respectively.

> In a paper, someone tested that when the number of nodes is less than 6, it is faster to look up nodes in a doubly linked list than in a red-black tree. So less than 6 nodes use double linked list. Red-black trees are used when there are more than 6 nodes. In the implementation of java's red-black tree, there is also such a design.

When the number of nodes is less than 6, the doubly linked list inserts nodes from small to large according to the size of the model. When the node is greater than 6 hours, first convert the double-linked list into a red-black tree, and then insert nodes into the red-black tree. A red-black tree is an AVL tree, which guarantees ordering. So the time complexity of insert, update and delete operations is O(logn)

When adding, deleting, and updating a resource model, insert nodes into this data structure. The data structure will count and maintain the count value of resourceModel.

### DIFF
--- a/pkg/modeling/modeling.go
+++ b/pkg/modeling/modeling.go
@@ -1,0 +1,327 @@
+package modeling
+
+import (
+	"container/list"
+	"errors"
+	"math"
+	"sync"
+
+	rbt "github.com/emirpasic/gods/trees/redblacktree"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/klog/v2"
+
+	clusterapis "github.com/karmada-io/karmada/pkg/apis/cluster/v1alpha1"
+)
+
+var (
+	mu                  sync.Mutex
+	modelSortings       [][]resource.Quantity
+	defaultModelSorting []clusterapis.ResourceName
+)
+
+// ResourceSummary records the list of resourceModels
+type ResourceSummary []resourceModels
+
+// resourceModels records the number of each allocatable resource models.
+type resourceModels struct {
+	// quantity is the total number of each allocatable resource models
+	// +required
+	Quantity int
+
+	// when the number of node is less than or equal to six, it will be sorted by linkedlist,
+	// when the number of node is more than six, it will be sorted by red-black tree.
+
+	// when the data structure is linkedlist,
+	// each item will store clusterResourceNode.
+	// +required
+	linkedlist *list.List
+
+	// when the data structure is redblack tree,
+	// each item will store a key-value pair,
+	// key is ResourceList, the value is quantity of this ResourceList
+	// +optional
+	redblackTree *rbt.Tree
+}
+
+// ClusterResourceNode represents the each raw resource entity without modeling.
+type ClusterResourceNode struct {
+	// quantity is the the number of this node
+	// Only when the resourceLists are exactly the same can they be counted as the same node.
+	// +required
+	quantity int
+
+	// resourceList records the resource list of this node.
+	// It maybe contain cpu, mrmory, gpu...
+	// User can specify which parameters need to be included before the cluster starts
+	// +required
+	resourceList ResourceList
+}
+
+// ResourceList is a set of (resource name, quantity) pairs.
+type ResourceList map[clusterapis.ResourceName]resource.Quantity
+
+// InitSummary is the init function of modeling data structure
+func InitSummary(resourceModels []clusterapis.ResourceModel) (ResourceSummary, error) {
+	rsName, rsList := []clusterapis.ResourceName{}, []ResourceList{}
+	for _, rm := range resourceModels {
+		tmp := map[clusterapis.ResourceName]resource.Quantity{}
+		for _, rmItem := range rm.Ranges {
+			if len(rsName) != len(rm.Ranges) {
+				rsName = append(rsName, rmItem.Name)
+			}
+			tmp[rmItem.Name] = rmItem.Min
+		}
+		rsList = append(rsList, tmp)
+	}
+
+	if len(rsName) != 0 && len(rsList) != 0 && (len(rsName) != len(rsList[0])) {
+		return nil, errors.New("the number of resourceName is not equal the number of resourceList")
+	}
+	var rs ResourceSummary
+	if len(rsName) != 0 {
+		defaultModelSorting = rsName
+	}
+	rs = make(ResourceSummary, len(rsList))
+	// generate a sorted array by first priority of ResourceName
+	modelSortings = make([][]resource.Quantity, len(rsList))
+	for index := 0; index < len(rsList); index++ {
+		for i, name := range rsName {
+			modelSortings[i] = append(modelSortings[i], rsList[index][name])
+		}
+	}
+	return rs, nil
+}
+
+// NewClusterResourceNode create new cluster resource node
+func NewClusterResourceNode(resourceList corev1.ResourceList) ClusterResourceNode {
+	return ClusterResourceNode{
+		quantity:     1,
+		resourceList: ConvertToResourceList(resourceList),
+	}
+}
+
+func (rs *ResourceSummary) getIndex(crn ClusterResourceNode) int {
+	index := math.MaxInt
+	for i, m := range defaultModelSorting {
+		tmpIndex := searchLastLessElement(modelSortings[i], crn.resourceList[m])
+		if tmpIndex < index {
+			index = tmpIndex
+		}
+	}
+	return index
+}
+
+func searchLastLessElement(nums []resource.Quantity, target resource.Quantity) int {
+	low, high := 0, len(nums)-1
+	for low <= high {
+		mid := low + ((high - low) >> 1)
+		diff1 := nums[mid].Cmp(target)
+		var diff2 int
+		if mid != len(nums)-1 {
+			diff2 = nums[mid+1].Cmp(target)
+		}
+		// diff < 1 means nums[mid] <= target
+		// diff == 1 means nums[mid+1] > target
+		if diff1 < 1 {
+			if (mid == len(nums)-1) || (diff2 == 1) {
+				// find the last less element that equal to target element
+				return mid
+			}
+			low = mid + 1
+		} else {
+			high = mid - 1
+		}
+	}
+	return -1
+}
+
+// clusterResourceNodeComparator provides a fast comparison on clusterResourceNodes
+func clusterResourceNodeComparator(a, b interface{}) int {
+	s1 := a.(ClusterResourceNode)
+	s2 := b.(ClusterResourceNode)
+	for index := 0; index < len(defaultModelSorting); index++ {
+		tmp1, tmp2 := s1.resourceList[defaultModelSorting[index]], s2.resourceList[defaultModelSorting[index]]
+
+		diff := tmp1.Cmp(tmp2)
+		if diff != 0 {
+			return diff
+		}
+	}
+	return 0
+}
+
+func safeChangeNum(num *int, change int) {
+	mu.Lock()
+	defer mu.Unlock()
+
+	(*num) += change
+}
+
+// AddToResourceSummary add resource node into modeling summary
+func (rs *ResourceSummary) AddToResourceSummary(crn ClusterResourceNode) {
+	index := rs.getIndex(crn)
+	modeling := &(*rs)[index]
+	if rs.GetNodeNumFromModel(modeling) <= 5 {
+		root := modeling.linkedlist
+		if root == nil {
+			root = list.New()
+		}
+		found := false
+		// traverse linkedlist to add quantity of recourse modeling
+		for element := root.Front(); element != nil; element = element.Next() {
+			switch clusterResourceNodeComparator(element.Value, crn) {
+			case 0:
+				{
+					tmpCrn := element.Value.(ClusterResourceNode)
+					safeChangeNum(&tmpCrn.quantity, crn.quantity)
+					element.Value = tmpCrn
+					found = true
+					break
+				}
+			case 1:
+				{
+					root.InsertBefore(crn, element)
+					found = true
+					break
+				}
+			case -1:
+				{
+					continue
+				}
+			}
+			if found {
+				break
+			}
+		}
+		if !found {
+			root.PushBack(crn)
+		}
+		modeling.linkedlist = root
+	} else {
+		root := modeling.redblackTree
+		if root == nil {
+			root = llConvertToRbt(modeling.linkedlist)
+			modeling.linkedlist = nil
+		}
+		tmpNode := root.GetNode(crn)
+		if tmpNode != nil {
+			node := tmpNode.Key.(ClusterResourceNode)
+			safeChangeNum(&node.quantity, crn.quantity)
+			tmpNode.Key = node
+		} else {
+			root.Put(crn, crn.quantity)
+		}
+		modeling.redblackTree = root
+	}
+	safeChangeNum(&modeling.Quantity, crn.quantity)
+}
+
+func llConvertToRbt(list *list.List) *rbt.Tree {
+	root := rbt.NewWith(clusterResourceNodeComparator)
+	for element := list.Front(); element != nil; element = element.Next() {
+		tmpCrn := element.Value.(ClusterResourceNode)
+		root.Put(tmpCrn, tmpCrn.quantity)
+	}
+	return root
+}
+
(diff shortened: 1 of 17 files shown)


Poor12:
Hi, @halfrost, please check DCO and other CI failure.

Poor12:
cc @RainbowMango to start CI

pkg/modeling/modeling.go
```diff
@@ -0,0 +1,340 @@
...
+
+// resourceModels records the number of each allocatable resource models.
+// models is a pointer, it points to the address of the first model
+// You don't need to care about the data structure behind the first model.
+type resourceModels struct {
+	// quantity is the total number of each allocatable resource models
+	// +required
+	Quantity int
+
+	// when the the number of node is less than or equal to six, it will be sorted by linkedlist,
```
Poor12, line 79: when the number

pkg/modeling/modeling.go
```diff
@@ -0,0 +1,340 @@
...
+	defaultModelLevel   = 10
+	modelSorting        []int64
+	DefaultModelSorting = []corev1.ResourceName{
+		corev1.ResourceCPU,
+		corev1.ResourceMemory,
+		corev1.ResourceStorage,
+		corev1.ResourceEphemeralStorage,
+	}
+
+	DefaultModel = []corev1.ResourceList{
```
Poor12, line 25: 10 default model？ Why index from cluster API is from 0 to 11

pkg/apis/cluster/types.go
```diff
@@ -104,6 +105,65 @@ type ClusterSpec struct {
 	// any resource that does not tolerate the Taint.
 	// +optional
 	Taints []corev1.Taint
+
+	// The model list of resource modeling in this cluster. Each modeling name and quota can be customized by the user.
```
RainbowMango, line 109: Comments should be started with `ResourceModels xxx`

RainbowMango: > Each modeling name and quota can be customized by the user.

Does the modeling `name` can be customized? I guess the name should be allowed `enumerated values`. 
Otherwise, for the name `my-customized-resource`, how do we know what it is?

pkg/apis/cluster/types.go
```diff
@@ -216,6 +276,19 @@ type ResourceSummary struct {
...
+	// The Grade of AllocatableModelingItem is the index of ResourceModel,
+	// The Count of AllocatableModelingItem is the number of nodes that own the resources delineated by this modeling.
+	// For example, AllocatableModelingItem{Grade: 2, Count: 10}, which means there are 10 such resource models of ResourceModel[2].
+	// +optional
+	AllocatableModeling []AllocatableModelingItem
+}
+
+type AllocatableModelingItem struct {
+	Grade int
+	Count int
```
RainbowMango, line 291: Need comments here.

RainbowMango (commented):
I just looked at the API part.

We also need to add the API to `pkg/apis/cluster/v1alpha1`.

RainbowMango:
Hi @halfrost
@Poor12 created an issue(#2379) to track the whole task, he will work with you to finish it.

pkg/modeling/modeling.go
```diff
@@ -0,0 +1,340 @@
...
+		}
+		quantity2, ok := tmp2.AsInt64()
+		if !ok {
+			klog.Infof("ModelComparator: Unable to parse the values of %v's quantity2 in the cluster", DefaultModelSorting[index])
+		}
+		diff = quantity1 - quantity2
+		if diff < 0 {
+			return -1
+		}
+		if diff > 0 {
```
Poor12, line 186: There is something wrong with the comparison function. For example, if a's limit is 2C 5G, b is 1C 8G, that will be CPU diff > 0, memory diff < 0. So I think should not  return directly.

pkg/apis/cluster/types.go
```diff
@@ -104,6 +105,65 @@ type ClusterSpec struct {
...
+
+// ResourceModel describes the modeling that you want to statistics.
+type ResourceModel struct {
+	// Grade is the index for the resource modeling.
+	// +optional
+	Grade int
+
+	// Ranges describes the resource quota ranges.
+	// +optional
+	Ranges []ResourceModelItem
```
Poor12, line 151: One question：whether the order of the range represents the order of the following sort？

pkg/apis/cluster/types.go
```diff
@@ -104,6 +105,65 @@ type ClusterSpec struct {
...
+}
+
+// ResourceModelItem describes the detail of each modeling quota that ranges from min to max.
+type ResourceModelItem struct {
+	// Name is the name for the resource that you want to categorize.
+	// +optional
+	Name string
```
Poor12, line 158: Name should be enum type

pkg/apis/cluster/types.go
```diff
@@ -104,6 +105,65 @@ type ClusterSpec struct {
...
+	// - index: 11
+	// range:
+	//   - name: CPU
+	//     min: 1024 C
+	//     max: MAXINT
+	//   - name: memory
+	//     min: 1024 GB
+	//     max: MAXINT
+	// +optional
+	ResourceModels []ResourceModel
```
Poor12, line 140: types.go in apis/cluster/v1alpha1 also needs to be modified.

pkg/modeling/modeling.go
halfrost: I also want to discuss this part with you. I'm comparing the quota by priority. Assuming that the priority of the cpu is high, the cpu is small means that this resource quota's priority is low.

But now there is a situation, as you said. If cpu is small, mem is large. So should this be considered big or small? How do you think about this situation?

pkg/apis/cluster/types.go
halfrost: Ranges is the quota for different modeling resources. For example, assume that grade = 2, the resource corresponding to this modeling has 3 indicators, cpu, gpu, mem. Then Ranges is the value range of these three indicators.

pkg/apis/cluster/types.go
halfrost: Let's discuss this line. For now, this structure can support 4 kinds of resources. In the actual production environment, are there more resource names? If there is, I will customize an enum structure to support more resource names.

```go
const (
	// CPU, in cores. (500m = .5 cores)
	ResourceCPU ResourceName = "cpu"
	// Memory, in bytes. (500Gi = 500GiB = 500 * 1024 * 1024 * 1024)
	ResourceMemory ResourceName = "memory"
	// Volume size, in bytes (e,g. 5Gi = 5GiB = 5 * 1024 * 1024 * 1024)
	ResourceStorage ResourceName = "storage"
	// Local ephemeral storage, in bytes. (500Gi = 500GiB = 500 * 1024 * 1024 * 1024)
	// The resource name for ResourceEphemeralStorage is alpha and it can change across releases.
	ResourceEphemeralStorage ResourceName = "ephemeral-storage"
)
```

pkg/modeling/modeling.go
Poor12: Could you share your slack with me? we can talk about it on slack.

pkg/apis/cluster/types.go
halfrost: Done.

pkg/apis/cluster/types.go
halfrost: Done.

pkg/apis/cluster/types.go
halfrost: You are right. What resource names are needed in the current production environment? In this version, I only support 4 kinds of resources. If there needs more resource name, I will customize an enum structure to support more resource names.

```go
const (
	// CPU, in cores. (500m = .5 cores)
	ResourceCPU ResourceName = "cpu"
	// Memory, in bytes. (500Gi = 500GiB = 500 * 1024 * 1024 * 1024)
	ResourceMemory ResourceName = "memory"
	// Volume size, in bytes (e,g. 5Gi = 5GiB = 5 * 1024 * 1024 * 1024)
	ResourceStorage ResourceName = "storage"
	// Local ephemeral storage, in bytes. (500Gi = 500GiB = 500 * 1024 * 1024 * 1024)
	// The resource name for ResourceEphemeralStorage is alpha and it can change across releases.
	ResourceEphemeralStorage ResourceName = "ephemeral-storage"
)
```

pkg/modeling/modeling.go
halfrost: Sorry, it's done.

halfrost:
> Hi @halfrost @Poor12 created an issue(#2379) to track the whole task, he will work with you to finish it.

Sounds great. We will try our best to complete this task together.

halfrost:
@Poor12 I have already split API part out of this pr. Please review API part in this pr https://github.com/karmada-io/karmada/pull/2386

pkg/modeling/modeling.go
```diff
@@ -0,0 +1,403 @@
...
+		ResourceCPU,
+		ResourceMemory,
+		ResourceStorage,
+		ResourceEphemeralStorage,
+	}
+
+	// Although the quota of each resource modeling is an interval. But the right boundary of each interval never coincides with the left boundary of the next interval.
+	// If the two overlap, it will cause ambiguity, and the modeling in the overlapping interval will belong to multiple intervals, which will cause an error.
+	// Then we can mark the interval only with the left boundary of each interval.
+	DefaultModel = []ResourceList{
```
Poor12, line 76: unit of Default model is uncorrect.

pkg/modeling/modeling.go
```diff
@@ -0,0 +1,403 @@
...
+	Name ResourceName
+
+	// Min is the minimum amount of this resource represented by resource name。
+	// +optional
+	Min resource.Quantity
+
+	// Max is the maximum amount of this resource represented by resource name。
+	// +optional
+	Max resource.Quantity
+}
```
Poor12, line 60: These values should come from cluster.spec.ResourceModel.

pkg/modeling/modeling.go
```diff
@@ -0,0 +1,403 @@
...
+	// +optional
+	redblackTree *rbt.Tree
+}
+
+// clusterResourceNode represents the each raw resource entity without modeling.
+type clusterResourceNode struct {
+	// quantity is the the number of this node
+	// Only when the resourceLists are exactly the same can they be counted as the same node.
+	// +required
+	quantity int
```
Poor12, line 150: Quantity may unneccassary

pkg/modeling/modeling.go
```diff
@@ -0,0 +1,403 @@
...
+
+	// resourceList records the resource list of this node.
+	// It maybe contain cpu, mrmory, gpu...
+	// User can specify which parameters need to be included before the cluster starts
+	// +required
+	resourceList ResourceList
```
Poor12, line 156: corev1.ResourceList

pkg/modeling/modeling.go
halfrost: Although it is impossible to insert multiple resource nodes at once. But if the same resource node is inserted, it needs to be accumulated on this node. For example, if I insert a resource that has cpu 2c memory 4G 2 times, I need to accumulate it in clusterResourceNode.

pkg/modeling/modeling.go
halfrost: Here need to use a custom ResourceList data structure. Because ResourceName is an enum maintained by API. The key of ResourceList is ResourceName, which is a custom data structure.

pkg/modeling/modeling.go
halfrost: Done.

pkg/modeling/modeling.go
halfrost: Done.

pkg/modeling/modeling.go
Poor12: Then we need to provide a method to convert corev1.resourcelist to Resourcelist.

pkg/generated/clientset/versioned/fake/register.go
```diff
@@ -33,14 +33,14 @@ var localSchemeBuilder = runtime.SchemeBuilder{
...
+//	import (
+//	  "k8s.io/client-go/kubernetes"
+//	  clientsetscheme "k8s.io/client-go/kubernetes/scheme"
+//	  aggregatorclientsetscheme "k8s.io/kube-aggregator/pkg/client/clientset_generated/clientset/scheme"
+//	)
 //
-//   kclientset, _ := kubernetes.NewForConfig(c)
-//   _ = aggregatorclientsetscheme.AddToScheme(clientsetscheme.Scheme)
+//	kclientset, _ := kubernetes.NewForConfig(c)
+//	_ = aggregatorclientsetscheme.AddToScheme(clientsetscheme.Scheme)
```
RainbowMango, line 43: This change is not expected, and this might be the reason why CI fails.

pkg/modeling/modeling.go
```diff
@@ -0,0 +1,341 @@
...
+	if len(rsName) != 0 {
+		defaultModelSorting = rsName
+	}
+	rs = make(ResourceSummary, len(rsList))
+	// generate a sorted array by first priority of ResourceName
+	modelSortings = make([][]int64, len(rsList))
+	for index := 0; index < len(rsList); index++ {
+		for i, name := range rsName {
+			tmpQuantity := rsList[index][name]
+			quantityNum, ok := tmpQuantity.AsInt64()
```
Poor12, line 91: This method needs to consider the milliquantity of the CPU.

pkg/modeling/modeling.go
```diff
@@ -0,0 +1,341 @@
...
+		quantity1, ok := tmp1.AsInt64()
+		if !ok {
+			klog.Infof("ModelComparator: Unable to parse the values of %v's quantity1 in the cluster", defaultModelSorting[index])
+		}
+		quantity2, ok := tmp2.AsInt64()
+		if !ok {
+			klog.Infof("ModelComparator: Unable to parse the values of %v's quantity2 in the cluster", defaultModelSorting[index])
+		}
+		diff = quantity1 - quantity2
+		if diff < 0 {
```
Poor12, line 159: Now the value judgment mainly depends on the resource with the highest priority, which may lead to some misjudgments.

Poor12 (commented):
Two questions remaining.

RainbowMango:
@halfrost I tried to fix the CI failure and pushed the commits to your branch on your forked GitHub repo.

Now the CI failure is due to #2404 and is not available in the period `August 22, 12:00 UTC - August 22, 16:00 UTC`. 
You can rebase your branch and push again, or I'll retrigger it tomorrow.

pkg/modeling/modeling.go
halfrost: You said internal sorting rule can be ignored.

Poor12:
Generallly look good to me，cc @RainbowMango

RainbowMango (approved):
Thanks. Generally looks good to me.
Given we have a lot of work to do, I'll move this forward and fix nits by following PRs.
/lgtm
/approve

pkg/apis/cluster/types.go
chaosi-zju: the extension name is defined by user, diffierent user may has different custom names, we can not predict the users' extension name

PR #2373, opened 2022-08-13
remove TargetCluster when cluster is terminating

### LINKED ISSUE #1995: feature: unjoin a cluster gracefully

Garrybest:

**What type of PR is this?**
/kind feature

Remove cluster scheduling result in spec.clusters when the cluster is terminating.

part of #1762
Fixes #1995

```release-note
NONE
```

### DIFF
--- a/cmd/controller-manager/app/controllermanager.go
+++ b/cmd/controller-manager/app/controllermanager.go
@@ -184,12 +184,14 @@ func startClusterController(ctx controllerscontext.Context) (enabled bool, err e
 	opts := ctx.Opts
 
 	clusterController := &cluster.Controller{
-		Client:                    mgr.GetClient(),
-		EventRecorder:             mgr.GetEventRecorderFor(cluster.ControllerName),
-		ClusterMonitorPeriod:      opts.ClusterMonitorPeriod.Duration,
-		ClusterMonitorGracePeriod: opts.ClusterMonitorGracePeriod.Duration,
-		ClusterStartupGracePeriod: opts.ClusterStartupGracePeriod.Duration,
-		FailoverEvictionTimeout:   opts.FailoverEvictionTimeout.Duration,
+		Client:                             mgr.GetClient(),
+		EventRecorder:                      mgr.GetEventRecorderFor(cluster.ControllerName),
+		ClusterMonitorPeriod:               opts.ClusterMonitorPeriod.Duration,
+		ClusterMonitorGracePeriod:          opts.ClusterMonitorGracePeriod.Duration,
+		ClusterStartupGracePeriod:          opts.ClusterStartupGracePeriod.Duration,
+		FailoverEvictionTimeout:            opts.FailoverEvictionTimeout.Duration,
+		EnableTaintManager:                 ctx.Opts.EnableTaintManager,
+		ClusterTaintEvictionRetryFrequency: 10 * time.Second,
 	}
 	if err := clusterController.SetupWithManager(mgr); err != nil {
 		return false, err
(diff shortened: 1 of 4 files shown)


pkg/controllers/cluster/cluster_controller.go
```diff
@@ -219,16 +230,21 @@ func (c *Controller) syncCluster(ctx context.Context, cluster *clusterv1alpha1.C
...
-func (c *Controller) removeCluster(cluster *clusterv1alpha1.Cluster) (controllerruntime.Result, error) {
-	err := c.removeExecutionSpace(cluster)
-	if err != nil {
-		klog.Errorf("Failed to remove execution space %v, err is %v", cluster.Name, err)
-		c.EventRecorder.Event(cluster, corev1.EventTypeWarning, fmt.Sprintf("Failed %s", clusterv1alpha1.EventReasonCreateExecutionSpaceFailed), err.Error())
+func (c *Controller) removeCluster(ctx context.Context, cluster *clusterv1alpha1.Cluster) (controllerruntime.Result, error) {
+	// add terminating taint before cluster is deleted
+	if err := utilhelper.UpdateClusterControllerTaint(ctx, c.Client, []*corev1.Taint{TerminatingTaintTemplate}, nil, cluster); err != nil {
+		c.EventRecorder.Event(cluster, corev1.EventTypeWarning, clusterv1alpha1.EventReasonRemoveTargetClusterFailed, err.Error())
+		klog.ErrorS(err, "Failed to update terminating taint", "cluster", cluster.Name)
```
RainbowMango, line 237: Here we assume there is no toleration for `cluster.karmada.io/terminating`, so the `taint manager` would finish the cleanup soon without waiting, am I right?

cmd/controller-manager/app/controllermanager.go
```diff
@@ -183,23 +183,25 @@ func startClusterController(ctx controllerscontext.Context) (enabled bool, err e
 	mgr := ctx.Mgr
 	opts := ctx.Opts
 
+	if err := cluster.IndexField(mgr); err != nil {
```
RainbowMango, line 186: Seems we still don't need to move this out, since the only place uses the index is also controlled by `EnableTaintManager`:
```
// check if target cluster is removed from all bindings.
	if c.EnableTaintManager {
```

Garrybest: I'm afraid not, we need this indexer in `isTargetClusterRemoved`.

pkg/controllers/cluster/cluster_controller.go
Garrybest: That's right.

pkg/controllers/cluster/cluster_controller.go
```diff
@@ -238,9 +254,46 @@ func (c *Controller) removeCluster(cluster *clusterv1alpha1.Cluster) (controller
 	// delete the health data from the map explicitly after we removing the cluster.
 	c.clusterHealthMap.delete(cluster.Name)
 
+	// check if target cluster is removed from all bindings.
+	if c.EnableTaintManager {
+		if done, err := c.isTargetClusterRemoved(ctx, cluster); err != nil {
+			klog.ErrorS(err, "Failed to check whether target cluster is removed from bindings", "cluster", cluster.Name)
```
RainbowMango, line 260: https://github.com/karmada-io/karmada/pull/2373/files/d468101a705ce83fd6ffc3022fba2a8199b1c7ca#r945591288

I mean here we checked the `c.EnableTaintManager`. right?

Garrybest: Oh, I see.

RainbowMango (approved):
/lgtm
/approve

PR #2374, opened 2022-08-14
improve ut for binding

AllenZMC:
**What type of PR is this?**
/kind cleanup

improve ut for binding
Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/util/helper/binding_test.go
+++ b/pkg/util/helper/binding_test.go
@@ -4,8 +4,10 @@ import (
 	"reflect"
 	"testing"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 
+	clusterapis "github.com/karmada-io/karmada/pkg/apis/cluster"
 	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
 )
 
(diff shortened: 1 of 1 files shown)


AllenZMC:
@XiShanYongYe-Chang

XiShanYongYe-Chang (commented):
thanks~
/lgtm

AllenZMC:
@RainbowMango

RainbowMango:
/assign
Will look at it later. Thanks.

pkg/util/helper/binding_test.go
```diff
@@ -180,3 +182,96 @@ func TestObtainBindingSpecExistingClusters(t *testing.T) {
...
+			name: "condition type is not Scheduled",
+			status: &workv1alpha2.ResourceBindingStatus{
+				Conditions: []metav1.Condition{
+					{Type: clusterapis.ClusterConditionReady, Status: metav1.ConditionTrue},
+				},
+			},
+			want: false,
+		},
+		{
+			name: "condition status is not true",
```
RainbowMango, line 202: `not true`?

pkg/util/helper/binding_test.go
```diff
@@ -180,3 +182,96 @@ func TestObtainBindingSpecExistingClusters(t *testing.T) {
...
+			want: []ClusterWeightInfo{
+				{
+					ClusterName: "a",
+					Weight:      1,
+				},
+				{
+					ClusterName: "b",
+					Weight:      1,
+				},
+			},
```
RainbowMango, line 267: The `map` is unordered, which means we can't assume always get `a` then get `b`.  
I don't know how to test the `same weight` scenario.

RainbowMango, line 267: The `map` is unordered, which means we can't assume always get `a` then get `b`.  
I don't know how to test the `same weight` scenario.

PR #2383, opened 2022-08-16
Use transform func to save memory when large cluster

Poor12:

**What type of PR is this?**
/kind feature

Use transform func to save memory for node and pod when large cluster.

Fixes #

None


Poor12:
cc @RainbowMango

pkg/util/fedinformer/transform.go
```diff
@@ -0,0 +1,60 @@
...
+// it removes unused fields from pods before they are stored in the cache to save memory.
+func StripUnusedFieldsForPod(obj interface{}) (interface{}, error) {
+	pod, ok := obj.(*corev1.Pod)
+	if !ok {
+		// shouldn't happen
+		return obj, nil
+	}
+
+	aggregatedPod := &corev1.Pod{
+		Spec: corev1.PodSpec{
```
dddddai, line 49: 👍Nice, I guess this is why we are using another **typed informer** so that it won't impact other places where we fetch pods from cache

RainbowMango: Yes, but not exactly. `typed informer` has better performance than `generic informer`. Yes, we still benefit from that because here we can safely use the transform function.

pkg/util/fedinformer/typedmanager/single-cluster-manager.go
```diff
@@ -206,3 +209,11 @@ func (s *singleClusterInformerManagerImpl) Context() context.Context {
 func (s *singleClusterInformerManagerImpl) GetClient() kubernetes.Interface {
 	return s.client
 }
+
+func (s *singleClusterInformerManagerImpl) SetTransformFunc(resource schema.GroupVersionResource, handler cache.TransformFunc) error {
+	resourceInformer, err := s.informerFactory.ForResource(resource)
+	if err != nil {
+		return err
+	}
+	return resourceInformer.Informer().SetTransform(handler)
```
RainbowMango, line 218: > Transform function must be set before starting the informer.

This function can't guarantee that.

Poor12: How about move TransformFunc as a ForResource/Lister parameter？

pkg/util/fedinformer/typedmanager/multi-cluster-manager.go
```diff
@@ -20,9 +21,9 @@ func init() {
 }
 
 // GetInstance returns a shared MultiClusterInformerManager instance.
-func GetInstance() MultiClusterInformerManager {
+func GetInstance(transformFuncs map[schema.GroupVersionResource]cache.TransformFunc) MultiClusterInformerManager {
```
RainbowMango, line 24: The `GetInstance` actually just initializes at the first time it is called, which means only the first input takes effect, but the parameters by following calls will be ignored.

That's probably unacceptable.

pkg/util/fedinformer/transform.go
```diff
@@ -0,0 +1,64 @@
...
+		return obj, nil
+	}
+	// ManagedFields is large and we never use it
+	accessor.SetManagedFields(nil)
+	return obj, nil
+}
+
+// StripUnusedFieldsForNode is the transform function for shared informers,
+// it removes unused fields from nodes before they are stored in the cache to save memory.
+func StripUnusedFieldsForNode(obj interface{}) (interface{}, error) {
```
RainbowMango, line 24:
```suggestion
// NodeTransformFunc is the dedicated transform function for Node objects.
// It cleans up some parts of the object before it will be put into the controller
// cache to reduce memory usage.
//
// Note: this function removes most of the fields, please make sure your controller
// doesn't care for the removed fields, especially when use in shared informers.
func NodeTransformFunc(obj interface{}) (interface{}, error) {
```

pkg/util/fedinformer/transform.go
```diff
@@ -0,0 +1,64 @@
...
+			Allocatable: node.Status.Allocatable,
+			Conditions:  node.Status.Conditions,
+		},
+	}
+	return aggregatedNode, nil
+}
+
+// StripUnusedFieldsForPod is the transform function for shared informers,
+// it removes unused fields from pods before they are stored in the cache to save memory.
+func StripUnusedFieldsForPod(obj interface{}) (interface{}, error) {
```
RainbowMango, line 45:
```suggestion
// PodTransformFunc is the dedicated transform function for Pod objects.
// It cleans up some parts of the object before it will be put into the controller
// cache to reduce memory usage.
//
// Note: this function removes most of the fields, please make sure your controller
// doesn't care for the removed fields, especially when use in shared informers.
func PodTransformFunc(obj interface{}) (interface{}, error) {
```
Rename function name and add more comments.

RainbowMango (commented):
review still on-going.

pkg/util/fedinformer/typedmanager/multi-cluster-manager.go
```diff
@@ -22,7 +25,12 @@ func init() {
 // GetInstance returns a shared MultiClusterInformerManager instance.
 func GetInstance() MultiClusterInformerManager {
 	once.Do(func() {
-		instance = NewMultiClusterInformerManager(stopCh)
+		// transforms records Record the mapping relationship between gvr and transformFunc.
```
RainbowMango, line 28: I think we don't need the comment, it's straightforward enough.

pkg/util/fedinformer/typedmanager/single-cluster-manager.go
```diff
@@ -69,15 +69,17 @@ type SingleClusterInformerManager interface {
...
-		informerFactory: informers.NewSharedInformerFactory(client, defaultResync),
-		handlers:        make(map[schema.GroupVersionResource][]cache.ResourceEventHandler),
-		syncedInformers: make(map[schema.GroupVersionResource]struct{}),
-		ctx:             ctx,
-		cancel:          cancel,
-		client:          client,
+		informerFactory:  informers.NewSharedInformerFactory(client, defaultResync),
+		handlers:         make(map[schema.GroupVersionResource][]cache.ResourceEventHandler),
+		syncedInformers:  make(map[schema.GroupVersionResource]struct{}),
+		startedInformers: make(map[schema.GroupVersionResource]bool),
```
RainbowMango, line 78: It's ambiguous between the field name(`startedInformers`) and its type.
From the name `startedInformers` I can get this is the map for storing `started` informers only, but why it's type allows a `bool`?

pkg/util/fedinformer/typedmanager/single-cluster-manager.go
```diff
@@ -89,14 +91,22 @@ type singleClusterInformerManagerImpl struct {
...
+
 	client kubernetes.Interface
 
 	lock sync.RWMutex
 }
 
 func (s *singleClusterInformerManagerImpl) ForResource(resource schema.GroupVersionResource, handler cache.ResourceEventHandler) error {
+	if _, exist := s.startedInformers[resource]; !exist {
+		s.startedInformers[resource] = false
+	}
```
RainbowMango, line 108: Should use lock the map.

pkg/util/fedinformer/typedmanager/single-cluster-manager.go
```diff
@@ -139,17 +163,30 @@ func (s *singleClusterInformerManagerImpl) IsHandlerExist(resource schema.GroupV
 }
 
 func (s *singleClusterInformerManagerImpl) Lister(resource schema.GroupVersionResource) (interface{}, error) {
+	if _, exist := s.startedInformers[resource]; !exist {
+		s.startedInformers[resource] = false
+	}
```
RainbowMango, line 168: ditto

pkg/util/fedinformer/transform.go
```diff
@@ -0,0 +1,72 @@
...
+// It cleans up some parts of the object before it will be put into the controller
+// cache to reduce memory usage.
+//
+// Note: this function removes most of the fields, please make sure your controller
+// doesn't care for the removed fields, especially when use in shared informers.
+func NodeTransformFunc(obj interface{}) (interface{}, error) {
+	node, ok := obj.(*corev1.Node)
+	if !ok {
+		// shouldn't happen
+		return obj, nil
```
RainbowMango, line 32:
```suggestion
		return obj, fmt.Errorf("expect resource Node but got %v", reflect.TypeOf(obj))
```

pkg/util/fedinformer/transform.go
```diff
@@ -0,0 +1,72 @@
...
+// It cleans up some parts of the object before it will be put into the controller
+// cache to reduce memory usage.
+//
+// Note: this function removes most of the fields, please make sure your controller
+// doesn't care for the removed fields, especially when use in shared informers.
+func PodTransformFunc(obj interface{}) (interface{}, error) {
+	pod, ok := obj.(*corev1.Pod)
+	if !ok {
+		// shouldn't happen
+		return obj, nil
```
RainbowMango, line 57: ditto.

pkg/util/fedinformer/transform_test.go
```diff
@@ -0,0 +1,236 @@
...
+					Phase: corev1.PodRunning,
+				},
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, _ := PodTransformFunc(tt.obj)
+			if !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("StripUnusedFieldsForPod: got %v, want %v", got, tt.want)
```
RainbowMango, line 232: `StripUnusedFieldsForPod`?

RainbowMango (approved):
/lgtm
/approve

/retitle adopt transform func to reduce memory usage

PR #2385, opened 2022-08-16
Start lease controller with separated context.

RainbowMango:
**What type of PR is this?**
/kind cleanup

This PR aims to decouple the context between `informer manager` and `lease controller`.

As we are going to introduce a flag to disable cluster resource modeling(#1858), but the lease controller shouldn't be disabled by this flag.

Fixes #

```release-note
NONE
```


RainbowMango:
/assign @Garrybest 
cc @snowplayfire 
Would like to take a look?

PR #2386, opened 2022-08-16
Add cluster resource modeling api

halfrost:
What type of PR is this?
/kind feature

This PR introduced a feature of cluster resource modeling

Which issue(s) this PR fixes:
Part of https://github.com/karmada-io/karmada/issues/772

This pr is for issue https://github.com/karmada-io/karmada/issues/772 cluster resource modeling API part.

YES


halfrost:
@Poor12 This is API part of cluster resource modeling. Please help me review it. Thx.

halfrost:
@Poor12 I have already changed the resource name into an enum. We can maintain by ourselves in the future. We can add more resource names in the enum. Please review it.

RainbowMango:
@halfrost Please squash your commits and use [PR templates](https://github.com/karmada-io/karmada/blob/master/.github/PULL_REQUEST_TEMPLATE.md).

pkg/apis/cluster/v1alpha1/types.go
```diff
@@ -214,15 +298,15 @@ type ResourceSummary struct {
...
 	// Allocating represents the resources of a cluster that are pending for scheduling.
 	// Total amount of required resources of all Pods that are waiting for scheduling.
 	// +optional
-	Allocating corev1.ResourceList `json:"allocating,omitempty"`
+	Allocating ResourceList `json:"allocating,omitempty"`
 	// Allocated represents the resources of a cluster that have been scheduled.
 	// Total amount of required resources of all Pods that have been scheduled to nodes.
 	// +optional
-	Allocated corev1.ResourceList `json:"allocated,omitempty"`
+	Allocated ResourceList `json:"allocated,omitempty"`
```
chaunceyjiang, line 309: Why modify this field?

Will this change break compatibility?

halfrost: This change has been reverted. I didn't consider compatibility issues. So sorry for changing the wrong place.

pkg/apis/cluster/v1alpha1/types.go
```diff
@@ -16,6 +17,29 @@ const (
...
+	ResourceMemory ResourceName = "memory"
+	// Volume size, in bytes (e,g. 5Gi = 5GiB = 5 * 1024 * 1024 * 1024)
+	ResourceStorage ResourceName = "storage"
+	// Local ephemeral storage, in bytes. (e,g. 500Gi = 500GiB = 500 * 1024 * 1024 * 1024)
+	// The resource name for ResourceEphemeralStorage is alpha and it can change across releases.
+	ResourceEphemeralStorage ResourceName = "ephemeral-storage"
+)
+
+// ResourceList is a set of (resource name, quantity) pairs.
+type ResourceList map[ResourceName]resource.Quantity
```
Poor12, line 41: Delete ResourceList

halfrost: Done.

Poor12:
/lgtm
cc @RainbowMango

pkg/apis/cluster/v1alpha1/types.go
```diff
@@ -116,6 +137,66 @@ type ClusterSpec struct {
...
 	Taints []corev1.Taint `json:"taints,omitempty"`
+
+	// ResourceModels is the list of resource modeling in this cluster. Each modeling quota can be customized by the user.
+	// Modeling name must be one of the following: cpu, memory, storage, ephemeral-storage.
+	// If the user does not define the modeling name and modeling quota, it will be the default model.
+	// The default model index from 0 to 11. Each default model's cpu quota is (2^index-1, 2^index], index >= 0
+	// Each default model's memory quota is also (2^index-1, 2^index], index >= 0, for example, model0 is like this:
+	// - index: 0
+	// ranges:
+	//   - name: CPU
```
RainbowMango, line 148: `CPU`, should be lower-case, right?

pkg/apis/cluster/v1alpha1/types.go
```diff
@@ -116,6 +137,66 @@ type ClusterSpec struct {
 	// any resource that does not tolerate the Taint.
 	// +optional
 	Taints []corev1.Taint `json:"taints,omitempty"`
+
+	// ResourceModels is the list of resource modeling in this cluster. Each modeling quota can be customized by the user.
+	// Modeling name must be one of the following: cpu, memory, storage, ephemeral-storage.
+	// If the user does not define the modeling name and modeling quota, it will be the default model.
+	// The default model index from 0 to 11. Each default model's cpu quota is (2^index-1, 2^index], index >= 0
```
RainbowMango, line 144: I don't think we really need so many models by default.
How about 6? Then, the maximum model is (32, 64)?

@kevin-wangzefeng How do you think?

pkg/apis/cluster/v1alpha1/types.go
```diff
@@ -16,6 +17,26 @@ const (
...
+const (
+	// CPU, in cores. (e,g. 500m = .5 cores)
+	ResourceCPU ResourceName = "cpu"
+	// Memory, in bytes. (e,g. 500Gi = 500GiB = 500 * 1024 * 1024 * 1024)
+	ResourceMemory ResourceName = "memory"
+	// Volume size, in bytes (e,g. 5Gi = 5GiB = 5 * 1024 * 1024 * 1024)
+	ResourceStorage ResourceName = "storage"
+	// Local ephemeral storage, in bytes. (e,g. 500Gi = 500GiB = 500 * 1024 * 1024 * 1024)
+	// The resource name for ResourceEphemeralStorage is alpha and it can change across releases.
+	ResourceEphemeralStorage ResourceName = "ephemeral-storage"
```
RainbowMango, line 37: I know these constants are inherited from K/K, but please let the comments start with the constant name.

pkg/apis/cluster/v1alpha1/types.go
```diff
@@ -116,6 +137,66 @@ type ClusterSpec struct {
...
+	// +optional
+	Name ResourceName `json:"name,omitempty"`
+
+	// Min is the minimum amount of this resource represented by resource name。
+	// +optional
+	Min resource.Quantity `json:"min,omitempty"`
+
+	// Max is the maximum amount of this resource represented by resource name。
+	// +optional
+	Max resource.Quantity `json:"max,omitempty"`
```
RainbowMango, line 199: Do these 3 fields really `optional`? 
What if the user does not set them(or one of them)? 

In addition, how to set the last model range? For example, the user wants to set 2 models,
The first one is used to target the small nodes, like(1c, 10c], and the second one is used to target the larger nodes which have more than `10c`, but don't know the exact `Max` value.

pkg/apis/cluster/v1alpha1/types.go
```diff
@@ -215,14 +296,35 @@ type ResourceSummary struct {
...
 	// Total amount of required resources of all Pods that are waiting for scheduling.
 	// +optional
 	Allocating corev1.ResourceList `json:"allocating,omitempty"`
+
 	// Allocated represents the resources of a cluster that have been scheduled.
 	// Total amount of required resources of all Pods that have been scheduled to nodes.
 	// +optional
 	Allocated corev1.ResourceList `json:"allocated,omitempty"`
+
+	// AllocatableModeling represents the number of each resources modeling in a cluster that are available for scheduling.
```
RainbowMango, line 310: typo: resources --> resource

> represents the number of each resources modeling

The number of nodes xxxx?

pkg/apis/cluster/v1alpha1/types.go
```diff
@@ -215,14 +296,35 @@ type ResourceSummary struct {
...
+	// Total amount of allocatable resources on all nodes.
+	// The Grade of AllocatableModelingItem is the index of ResourceModel,
+	// The Count of AllocatableModelingItem is the number of nodes that own the resources delineated by this modeling.
+	// For example, AllocatableModelingItem{Grade: 2, Count: 10}, which means there are 10 such resource models of ResourceModel[2].
```
RainbowMango, line 314: Move the comments to `type AllocatableModelingItem struct`.

> which means there are 10 such resource models of ResourceModel[2]

By the way, this description is not accurate.

RainbowMango (commented):
Generally looks good to me.
Some small touches. 
/assign @kevin-wangzefeng

pkg/apis/cluster/v1alpha1/types.go
chaunceyjiang, line 37: What is the purpose of these const?

I feel very similar to [this](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/core/v1/types.go#L5256-L5274)

chaunceyjiang: https://github.com/karmada-io/karmada/pull/2367#discussion_r946298844

pkg/apis/cluster/v1alpha1/types.go
halfrost: Yes, it should be lower-case. I have fixed this comment.

pkg/apis/cluster/v1alpha1/types.go
halfrost: Done.

pkg/apis/cluster/v1alpha1/types.go
halfrost: The intention of set optional in there is that if the user does not want to set any value, the system will select the default modeling.

For the question2, we can make on a rule: the minimum value here can be obtained, and the maximum value cannot be obtained. In other words, the interval is closed on the left and open on the right. So in your example, the first one will be [1c,10c), the second one will be [10c, ∞).

For the right side of the last interval, which defaults to no matter what value you pass, anything is infinite to me. Because the quota of resources is larger than the left value of the last interval will be classified into the last interval. I have added some comments on this part.

pkg/apis/cluster/v1alpha1/types.go
halfrost: the number of each resource modeling in a cluster. Isn't it right?

pkg/apis/cluster/v1alpha1/types.go
halfrost: Done. I have modified this description. Thanks for such a rigorous inspection.👍🏻

pkg/apis/cluster/v1alpha1/types.go
halfrost: I also want to discuss this part. It is determined here, and the code of my implementation part should also be changed.

pkg/apis/cluster/v1alpha1/types.go
```diff
@@ -16,6 +17,26 @@ const (
...
+// Resource names must be not more than 63 characters, consisting of upper- or lower-case alphanumeric characters,
+// with the -, _, and . characters allowed anywhere, except the first or last character.
+// The default convention, matching that for annotations, is to use lower-case names, with dashes, rather than
+// camel case, separating compound words.
+// Fully-qualified resource typenames are constructed from a DNS-style subdomain, followed by a slash `/` and a name.
+const (
+	// ResourceCPU, in cores. (e,g. 500m = .5 cores)
+	ResourceCPU ResourceName = "cpu"
+	// ResourceMemory, in bytes. (e,g. 500Gi = 500GiB = 500 * 1024 * 1024 * 1024)
+	ResourceMemory ResourceName = "memory"
```
RainbowMango, line 32:
```suggestion
	// ResourceCPU in cores. (e,g. 500m = .5 cores)
	ResourceCPU ResourceName = "cpu"
	// ResourceMemory in bytes. (e,g. 500Gi = 500GiB = 500 * 1024 * 1024 * 1024)
	ResourceMemory ResourceName = "memory"
```

pkg/apis/cluster/v1alpha1/types.go
```diff
@@ -116,6 +137,79 @@ type ClusterSpec struct {
...
+// This rule ensure that all intervals have the same meaning. If the last interval is +∞,
+// it is definitely unreachable. Therefore, we define the right interval as the open interval.
+// For a valid interval, the value on the right is greater than the value on the left,
+// in other words, max must be greater than min.
+type ResourceModelItem struct {
+	// Name is the name for the resource that you want to categorize.
+	// +optional
+	Name ResourceName `json:"name,omitempty"`
+
+	// Min is the minimum amount of this resource represented by resource name。
```
RainbowMango, line 201: `。`--> `.`

pkg/apis/cluster/v1alpha1/types.go
```diff
@@ -116,6 +137,79 @@ type ClusterSpec struct {
...
+	// +optional
+	Min resource.Quantity `json:"min,omitempty"`
+
+	// Max is the maximum amount of this resource represented by resource name。
```
RainbowMango, line 205: 。--> `.`

pkg/apis/cluster/v1alpha1/types.go
RainbowMango: Yes, the default model should be close to real production practice, thus most case users don't need to customize it.
@Poor12 is investigating it, let's wait for a while.

pkg/apis/cluster/v1alpha1/types.go
```diff
@@ -116,6 +137,79 @@ type ClusterSpec struct {
...
+	Ranges []ResourceModelItem `json:"ranges,omitempty"`
+}
+
+// ResourceModelItem describes the detail of each modeling quota that ranges from min to max.
+// Please pay attention, by default, the value of min can be obtained, and the value of max cannot be obtained.
+// For example, in an interval, min = 2, max =10 is set, which means the interval [2,10).
+// This rule ensure that all intervals have the same meaning. If the last interval is +∞,
+// it is definitely unreachable. Therefore, we define the right interval as the open interval.
+// For a valid interval, the value on the right is greater than the value on the left,
+// in other words, max must be greater than min.
```
RainbowMango, line 195: I'm confused by the description.
> by default, the value of min can be obtained, and the value of max cannot be obtained.

How we can `obtain` the min value? Why we can't obtain the `max` value?

By the way, you are using `[2, 10)`(left open, right close) which violates the description on `ResourceModels`.

pkg/apis/cluster/v1alpha1/types.go
Poor12: // - index: 0
	// ranges:
	//   - name: CPU
	//     min: 0 C
	//     max: 1 C
	//   - name: memory
	//     min: 0 GB
	//     max: 4 GB
        // - index: 1
	// ranges:
	//   - name: CPU
	//     min: 1 C
	//     max: 2 C
	//   - name: memory
	//     min: 4 GB
	//     max: 16 GB
        // - index: 2
	// ranges:
	//   - name: CPU
	//     min: 2 C
	//     max: 4 C
	//   - name: memory
	//     min: 16 GB
	//     max: 32 GB
       // - index: 3
	// ranges:
	//   - name: CPU
	//     min: 4 C
	//     max: 8 C
	//   - name: memory
	//     min: 32 GB
	//     max: 64 GB
         // - index: 10
	// ranges:
	//   - name: CPU
	//     min: 64 C
	//     max: 128 C
	//   - name: memory
	//     min: 512 GB
	//     max: 1024 GB

pkg/apis/cluster/v1alpha1/types.go
```diff
@@ -116,6 +137,79 @@ type ClusterSpec struct {
...
+	Name ResourceName `json:"name,omitempty"`
+
+	// Min is the minimum amount of this resource represented by resource name。
+	// +optional
+	Min resource.Quantity `json:"min,omitempty"`
+
+	// Max is the maximum amount of this resource represented by resource name。
+	// Special Instructions, for the last ResourceModelItem, which no matter what Max value you pass,
+	// the meaning is infinite. Because for the last item,
+	// any ResourceModelItem's quota larger than Min will be classified to the last one.
```
RainbowMango, line 208: I basically agree with this that put all notes into the `last`(largest) model.
My concern is, from API point of view it might be counter-intuitive, for example:
I specified the last model is `(10c, 20c}`, why the nodes with `30c` are counted in this model ?

cc @kevin-wangzefeng for comments.

pkg/apis/cluster/v1alpha1/types.go
halfrost: The `obtain` here may be inaccurate. Using `include` is more appropriate.

`[2, 10)` is (left close, right open).

The reason why makes a rule that the left interval is inclusive and the right interval is not inclusive. It's because of wanting to unify all situations. If the last interval contains +∞. The right interval is definitely not included. Therefore, it makes a rule that all intervals on the right interval are not included. This makes the program logic more unified.

pkg/apis/cluster/v1alpha1/types.go
halfrost: Done.

pkg/apis/cluster/v1alpha1/types.go
halfrost: Done.

pkg/apis/cluster/v1alpha1/types.go
halfrost: Done.

pkg/apis/cluster/v1alpha1/types.go
```diff
@@ -215,14 +309,33 @@ type ResourceSummary struct {
...
+
 	// Allocated represents the resources of a cluster that have been scheduled.
 	// Total amount of required resources of all Pods that have been scheduled to nodes.
 	// +optional
 	Allocated corev1.ResourceList `json:"allocated,omitempty"`
+
+	// AllocatableModeling represents the number of each resource modeling in a cluster that are available for scheduling.
+	// Total amount of allocatable resources on all nodes.
+	// +optional
+	AllocatableModeling []AllocatableModelingItem `json:"allocatableModeling,omitempty"`
```
RainbowMango, line 326:
```suggestion
	// AllocatableModeling represents the number of each resource modeling in a cluster that are available for scheduling.
	// AllocatableModelings represents the statistical resource modeling.
	// +optional
	AllocatableModelings []AllocatableModeling `json:"allocatableModelings,omitempty"`
```
How about this?
Rename `ResourceModelItem` --> `ResourceModelRange`

pkg/apis/cluster/v1alpha1/types.go
```diff
@@ -215,14 +309,33 @@ type ResourceSummary struct {
...
+}
+
+// AllocatableModelingItem represents the number of this resources modeling in a cluster that are available for scheduling.
+// For example, AllocatableModelingItem{Grade: 2, Count: 10}, which means there are 10 such resource models of grade 2 ResourceModel.
+type AllocatableModelingItem struct {
```
RainbowMango, line 331:
```suggestion
+// AllocatableModeling represents the number of nodes in which allocatable resources in a specific resource model grade.
+// E.g. AllocatableModeling{Grade: 2, Count: 10} means 10 nodes belong to resource model in grade 2.
+type AllocatableModeling struct {
```

Rename `AllocatableModelingItem` --> `AllocatableModeling`?

pkg/apis/cluster/v1alpha1/types.go
halfrost: OK, I modify the comment in the API pr, then I will modify my Implementation code. I just change the default value. However, my test cases need to change a few groups case.

pkg/apis/cluster/v1alpha1/types.go
RainbowMango: I haven't seen it yet. Waiting for your next push.

pkg/apis/cluster/v1alpha1/types.go
halfrost: I understand that users will be confused in this case. But in this case, if `30c` should not be assigned to any group, it is weird too. This is also not reasonable. Then `30c` is discarded and not counted? I think it might be better to join the last model in statistical calculations. What's your opinion? @kevin-wangzefeng

pkg/apis/cluster/v1alpha1/types.go
halfrost: ResourceModelRange sounds better.

pkg/apis/cluster/v1alpha1/types.go
halfrost: Done.

pkg/apis/cluster/types.go
```diff
@@ -104,6 +125,97 @@ type ClusterSpec struct {
 	// any resource that does not tolerate the Taint.
 	// +optional
 	Taints []corev1.Taint
+
+	// ResourceModels is the list of resource modeling in this cluster. Each modeling quota can be customized by the user.
+	// Modeling name must be one of the following: cpu, memory, storage, ephemeral-storage.
+	// If the user does not define the modeling name and modeling quota, it will be the default model.
+	// The default model index from 0 to 8. Each default model's cpu quota is [2^(index-1), 2^index), 2 <= index <= 8
```
Poor12, line 132: when index =0， 2^(0-1)?， and memory quota doesn't satisfy your formula.

pkg/apis/cluster/v1alpha1/types.go
```diff
@@ -116,6 +137,97 @@ type ClusterSpec struct {
 	// any resource that does not tolerate the Taint.
 	// +optional
 	Taints []corev1.Taint `json:"taints,omitempty"`
+
+	// ResourceModels is the list of resource modeling in this cluster. Each modeling quota can be customized by the user.
+	// Modeling name must be one of the following: cpu, memory, storage, ephemeral-storage.
+	// If the user does not define the modeling name and modeling quota, it will be the default model.
+	// The default model index from 0 to 8. Each default model's cpu quota is [2^(index-1), 2^index), 2 <= index <= 8
+	// Each default model's memory quota is also [2^(index + 2), 2^(index + 3)), 2 <= index <= 8, E.g. model 0 is like this:
```
RainbowMango, line 145: The description might not be so clear and accurate I guess.

By the way, since the memory model is different from cpu's, should no `also`.

pkg/apis/cluster/v1alpha1/types.go
```diff
@@ -116,6 +137,97 @@ type ClusterSpec struct {
...
+	// - index: 0
+	// ranges:
+	//   - name: "cpu"
+	//     min: 0 C
+	//     max: 1 C
+	//   - name: "memory"
+	//     min: 0 GB
+	//     max: 4 GB
```
RainbowMango, line 153:
```suggestion
	// - grade: 0
	//   ranges:
	//   - name: "cpu"
	//     min: 0 C
	//     max: 1 C
	//   - name: "memory"
	//     min: 0 GB
	//     max: 4 GB
```

1. rename `index` to `grade`.
2. represent in yaml format

RainbowMango (commented):
This is probably my last comment. 
Thanks @halfrost for your hard and excellent work.

pkg/apis/cluster/types.go
halfrost: When index = 0 or index = 1, it does not have a uniform formula. So the formula starts with index = 2.

pkg/apis/cluster/v1alpha1/types.go
halfrost: Done.

pkg/apis/cluster/v1alpha1/types.go
halfrost: Done.

pkg/apis/cluster/v1alpha1/types.go
```diff
@@ -116,6 +137,100 @@ type ClusterSpec struct {
...
 	Taints []corev1.Taint `json:"taints,omitempty"`
+
+	// ResourceModels is the list of resource modeling in this cluster. Each modeling quota can be customized by the user.
+	// Modeling name must be one of the following: cpu, memory, storage, ephemeral-storage.
+	// If the user does not define the modeling name and modeling quota, it will be the default model.
+	// The default model grade from 0 to 8.
+	// When grade = 0 or grade = 1, the default model's cpu quota and memory quota is a fix value.
+	// When grade greater than or equal to 2, each default model's cpu quota is [2^(grade-1), 2^grade), 2 <= grade <= 7
+	// Each default model's memory quota is [2^(grade + 2), 2^(grade + 3)), 2 <= grade <= 7
+	// E.g. model 0 is like this:
```
RainbowMango, line 148: Much clear now. Should we also describe the last grade, since we are using `MAXINT`?

> E.g. model 0 is like this:

`model 0`?

pkg/apis/cluster/types.go
```diff
@@ -216,6 +331,23 @@ type ResourceSummary struct {
...
+	// AllocatableModelings represents the statistical resource modeling.
+	// +optional
+	AllocatableModelings []AllocatableModeling
+}
+
+// AllocatableModeling represents the number of nodes in which allocatable resources in a specific resource model grade.
+// E.g. AllocatableModeling{Grade: 2, Count: 10} means 10 nodes belong to resource model in grade 2.
+type AllocatableModeling struct {
+
+	// Grade is the index of ResourceModel.
```
RainbowMango, line 344: A redundant empty line.

pkg/apis/cluster/v1alpha1/types.go
halfrost: I modify to grade now. Done.

pkg/apis/cluster/types.go
halfrost: Delete it. Done.

pkg/apis/cluster/types.go
```diff
@@ -104,6 +125,101 @@ type ClusterSpec struct {
...
+	Min resource.Quantity
+
+	// Max is the maximum amount of this resource represented by resource name.
+	// Special Instructions, for the last ResourceModelRange, which no matter what Max value you pass,
+	// the meaning is infinite. Because for the last item,
+	// any ResourceModelRange's quota larger than Min will be classified to the last one.
+	// Of course, the value of the Max field is always greater than the value of the Min field.
+	// It should be true in any case.
+	// +optional
+	Max resource.Quantity
```
kevin-wangzefeng, line 222: > Special Instructions, for the last ResourceModelRange, which no matter what Max value you pass,
> the meaning is infinite. Because for the last item
Shall we make the last grade ignore user-input max value, or, add a computed last/largest grade according to last/largest user-input max value?

pkg/apis/cluster/types.go
```diff
@@ -104,6 +125,101 @@ type ClusterSpec struct {
...
+// For a valid interval, the value on the right is greater than the value on the left,
+// in other words, max must be greater than min.
+type ResourceModelRange struct {
+	// Name is the name for the resource that you want to categorize.
+	// +optional
+	Name ResourceName
+
+	// Min is the minimum amount of this resource represented by resource name.
+	// +optional
+	Min resource.Quantity
```
kevin-wangzefeng, line 213: Describing the default value would be very helpful here, e.g. Min of grade 0 defaults to zero

pkg/apis/cluster/types.go
```diff
@@ -104,6 +125,101 @@ type ClusterSpec struct {
...
+	//
+	// +optional
+	ResourceModels []ResourceModel
+}
+
+// ResourceModel describes the modeling that you want to statistics.
+type ResourceModel struct {
+	// Grade is the index for the resource modeling.
+	// +optional
+	Grade int
```
kevin-wangzefeng, line 192: Nit: what's the expected behavior when user sets ResourceModels with `grade 1:CPU[2,4)`, `grade 3:CPU[8,16)`, `grade 4:CPU[16,32)`?

kevin-wangzefeng: BTW, IIRC we currently don't define the output of `kubectl describe cluster`, but it might be great if we can define a human readable format of ResourceModels like:


```
ResourceModels:
	Grade 1:	CPU [2,4)	Memory [4G,8G)
	Grade 2:	CPU [4,8)	Memory [8G,16G)
	Grade 3:	CPU [8,16)	Memory [16G,32G)
```

A method like `ResourceModel.Stringify()` shall be helpful in this feature.

kevin-wangzefeng (commented):
Minor comments, thanks

pkg/apis/cluster/types.go
halfrost: My current implementation is that whatever value the last max is, those greater than min will be placed in the last group. This ensures that there will be no missing ResourceModelRange. Do you think my solution is reasonable?

pkg/apis/cluster/types.go
halfrost: Done.

pkg/apis/cluster/types.go
halfrost: when user sets ResourceModels with `grade 1:CPU[2,4)`, `grade 3:CPU[8,16)`, `grade 4:CPU[16,32)`, if some resource's CPU is 6, this resource will be assigned to a higher grade. In other words, it will belong to `grade 3:CPU[8,16)`. Does it make sense? 

By the way, in this case, grade2 doesn't exist? Or the user forgot to enter it?

halfrost: This `ResourceModel.Stringify()` feature is very nice. If time is enough, I can do it together in this version.

pkg/apis/cluster/types.go
RainbowMango: I agree with the implementation behavior, I guess what @kevin-wangzefeng mean is we should clearly note the behavior on API comments.

RainbowMango: By the way, the behavior description should be added to `ResourceModels`(.spec.ResourceModels), not on `ResourceModelRange` which just represents the range. How we use/treat ranges is decided at `ResourceModels`.

pkg/apis/cluster/v1alpha1/types.go
```diff
@@ -116,6 +137,102 @@ type ClusterSpec struct {
...
+// it is definitely unreachable. Therefore, we define the right interval as the open interval.
+// For a valid interval, the value on the right is greater than the value on the left,
+// in other words, max must be greater than min.
+type ResourceModelRange struct {
+	// Name is the name for the resource that you want to categorize.
+	// +optional
+	Name ResourceName `json:"name,omitempty"`
+
+	// Min is the minimum amount of this resource represented by resource name.
+	// E.g. Min of grade 0 defaults to zero.
```
RainbowMango, line 224: // Note: The Min value of first grade(usually 0) always acts as zero.
	// E.g. [1,2) equal to [0,2).

pkg/apis/cluster/types.go
RainbowMango: > when user sets ResourceModels with grade 1:CPU[2,4), grade 3:CPU[8,16), grade 4:CPU[16,32), if some resource's CPU is 6, this resource will be assigned to a higher grade.

+1. We can take this as the default behavior. But I don't think we should note this case.

> By the way, in this case, grade2 doesn't exist? Or the user forgot to enter it?

This is an abnormal case, probably due to a mistake. 
This usage is ambiguous, my suggestion is to validate the input at webhook(Validation Plugin) to avoid discontinuous ranges.

pkg/apis/cluster/v1alpha1/types.go
halfrost: Done.

RainbowMango (approved):
/lgtm
/approve

imageslr:
This is a good case for learning code review :)

PR #2387, opened 2022-08-16
Introduce `--enable-cluster-resource-modeling` to customize resource modeling feature

RainbowMango:
**What type of PR is this?**
/kind feature

This PR actually takes over #1858, fixed comments.

Fixes #1858

```release-note
`karmada-controller-manager`/`karmada-agent`: Introduced `--enable-cluster-resource-modeling` flag to enable or disable cluster modeling feature.
```

### DIFF
--- a/pkg/controllers/status/cluster_status_controller.go
+++ b/pkg/controllers/status/cluster_status_controller.go
@@ -91,6 +91,8 @@ type ClusterStatusController struct {
 
 	ClusterCacheSyncTimeout metav1.Duration
 	RateLimiterOptions      ratelimiterflag.Options
+
+	EnableClusterResourceModeling bool
 }
 
 // Reconcile syncs status of the given member cluster.
@@ -170,7 +172,7 @@ func (c *ClusterStatusController) syncClusterStatus(cluster *clusterv1alpha1.Clu
 	}
 
 	// skip collecting cluster status if not ready
-	if online && healthy && readyCondition.Status == metav1.ConditionTrue {
+	if online && healthy && readyCondition.Status == metav1.ConditionTrue && c.EnableClusterResourceModeling {
 		// get or create informer for pods and nodes in member cluster
 		clusterInformerManager, err := c.buildInformerForCluster(clusterClient)
 		if err != nil {
--- a/pkg/controllers/status/cluster_status_controller.go
+++ b/pkg/controllers/status/cluster_status_controller.go
@@ -92,6 +92,10 @@ type ClusterStatusController struct {
 	ClusterCacheSyncTimeout metav1.Duration
 	RateLimiterOptions      ratelimiterflag.Options
 
+	// EnableClusterResourceModeling indicates if enable cluster resource modeling.
+	// The resource modeling might be used by the scheduler to make scheduling decisions
+	// in scenario of dynamic replica assignment based on cluster free resources.
+	// Disable if it does not fit your cases for better performance.
 	EnableClusterResourceModeling bool
 }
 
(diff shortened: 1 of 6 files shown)


RainbowMango:
@dddddai Could you please take a look?

cc guys who might be interested in it.
@garfcat @likakuli

dddddai (commented):
Looks good!

/lgtm

RainbowMango:
I forgot to add the controller parameters at [here](https://github.com/karmada-io/karmada/blob/6faf2285849a1ace7fc7d4724d260a5e6758f740/cmd/agent/app/agent.go#L227), that's the root cause of the CI failure.

pkg/controllers/status/cluster_status_controller.go
```diff
@@ -197,21 +195,29 @@ func (c *ClusterStatusController) syncClusterStatus(cluster *clusterv1alpha1.Clu
...
 		}
+		currentClusterStatus.APIEnablements = apiEnables
 
-		nodes, err := listNodes(clusterInformerManager)
-		if err != nil {
-			klog.Errorf("Failed to list nodes for Cluster %s. Error: %v.", cluster.GetName(), err)
-		}
+		if c.EnableClusterResourceModeling {
+			// get or create informer for pods and nodes in member cluster
+			clusterInformerManager, err := c.buildInformerForCluster(clusterClient)
```
RainbowMango, line 202: The `buildInformerForCluster` now helps to initialize the informer for clusters, the informer will be used by other controllers. 
We must de-coupling it before moving this forward.
/hold

XiShanYongYe-Chang: +1
The current control coupling is a little severe.

dddddai: Maybe we can build the `generic informer` as usual but not build the `typed informer` in `buildInformerForCluster`?

RainbowMango: That's definitely a workable solution. 

But the annoying thing is it's a bad program paradigm. Too much coupling leads to hard to maintain.
Now, the `multi-cluster-informer-manager` is initialized by `cluster-status-controller` but used by `execution-controller` , and `work-status-controller`. :(

I'm trying to decouple them.

RainbowMango:
I just pushed two more commits, the first one is used to address the temporary solution, and the last one is just for verification and it will be removed before moving forward.

RainbowMango:
ready for review again :)  @dddddai

pkg/controllers/status/cluster_status_controller.go
```diff
@@ -256,22 +269,22 @@ func (c *ClusterStatusController) updateStatusIfNeeded(cluster *clusterv1alpha1.
...
 
+func (c *ClusterStatusController) initializeGenericInformerManagerForCluster(clusterClient *util.ClusterClient) {
+	if c.GenericInformerManager.IsManagerExist(clusterClient.ClusterName) {
+		return
+	}
+
+	dynamicClient, err := c.ClusterDynamicClientSetFunc(clusterClient.ClusterName, c.Client)
+	if err != nil {
+		klog.Errorf("Failed to build dynamic cluster client for cluster %s.", clusterClient.ClusterName)
+		return
```
dddddai, line 280: Should we return err? (keep the behavior as before)

RainbowMango: I think just one log is enough. 
Even in case of error, what else can the cluster status controller do except log again the error? 
The error shouldn't block the normal process.

Issue #2282, opened 2022-07-29
[Umbrella] Develop a new cluster registration method based on TLS/SSL


lonelyCZ:
**What would you like to be added**:

We will introduce  `bootstarp tokens` registration method, refering https://kubernetes.io/docs/reference/access-authn-authz/bootstrap-tokens

Iterated task:

- [x] Certificate generation normalization. #2097 #2297
- [x] Set open cluster-info to distribute root CA certificates. #2312 
- [x] Set the required `ClusterRole` and `ClusterRoleBinding` to enable the certificate signing request. #2357 
- [x] Develop the `karmadactl token` command to manage bootstrap tokens. #2399 
- [x] Develop the `karmadactl register` command to support the use of `token` to register the `PULL mode cluster`. #2388 
- [x] Print the new cluster register way after finishing `karmadactl init`. #2409 
- [x] Add user guide about `karmdactl register` command. https://github.com/karmada-io/website/pull/142 https://github.com/karmada-io/website/pull/143
- [x] Realize the certificate rotation function. #2596 

**Why is this needed**:

Currently, to register a member cluster in PULL mode with the Karmada control plane, it is necessary to provide the administrator certificate of the Karmada control plane. This behavior brings a certain risk of certificate leakage, and it is impossible to distinguish the permissions of multiple cluster administrators from single cluster administrators. We hope to enhance Karmada's certificate management capability to achieve the following goals:

1、Implement the bootstrap token authentication mode to register member clusters and automatically issue certificates.

2、Realize automatic certificate rotation function for karmada-agent, reduce operation and maintenance pressure and enhance its security.

Relating issues #1363 #1648

RainbowMango:
@lonelyCZ 
> Realize the certificate rotation function.

I have a question about the last task, does this task should be included in v1.3?

PR #2399, opened 2022-08-20
Implement karmadactl token command to manage bootstrap tokens

lonelyCZ:
Signed-off-by: lonelyCZ <531187475@qq.com>

**What type of PR is this?**
/kind feature

```
[root@master67 karmada]# ./karmadactl token -h
This command manages bootstrap tokens. It is optional and needed only for advanced use cases.

 In short, bootstrap tokens are used for establishing bidirectional trust between a client and a server. A bootstrap
token can be used when a client (for example a member cluster that is about to join control plane) needs to trust the
server it is talking to. Then a bootstrap token with the "signing" usage can be used. bootstrap tokens can also function
as a way to allow short-lived authentication to the API Server (the token serves as a way for the API Server to trust
the client), for example for doing the TLS Bootstrap.

 What is a bootstrap token more exactly? - It is a Secret in the kube-system namespace of type
"bootstrap.kubernetes.io/token". - A bootstrap token must be of the form " [a-z0-9]{6}. [a-z0-9]{16}". The former part
is the public token ID, while the latter is the Token Secret and it must be kept private at all circumstances! - The
name of the Secret must be named "bootstrap-token-(token-id)".

 This command is same as 'kubeadm token', but it will create tokens that are used by member clusters.

Examples:
  # Create a token and print the full 'karmadactl register' flag needed to join the cluster using the token.
  karmadactl token create --print-register-command

Available Commands:
  create        Create bootstrap tokens on the server
  delete        Delete bootstrap tokens on the server
  list          List bootstrap tokens on the server
```

```
[root@master67 karmada]# ./karmadactl token create -h
This command will create a bootstrap token for you. You can specify the usages for this token, the "time to live" and an
optional human friendly description.

 This should be a securely generated random token of the form " [a-z0-9]{6}. [a-z0-9]{16}".

Options:
    --description='':
        A human friendly description of how this token is used.

    --groups=[system:bootstrappers:karmada:default-cluster-token]:
        Extra groups that this token will authenticate as when used for authentication. Must match
        "\\Asystem:bootstrappers:[a-z0-9:-]{0,255}[a-z0-9]\\z"

    --karmada-context='':
        Name of the cluster context in control plane kubeconfig file.

    --print-register-command=false:
        Instead of printing only the token, print the full 'karmadactl join' flag needed to join the member cluster
        using the token.

    --ttl=24h0m0s:
        The duration before the token is automatically deleted (e.g. 1s, 2m, 3h). If set to '0', the token will never
        expire

    --usages=[signing,authentication]:
        Describes the ways in which this token can be used. You can pass --usages multiple times or provide a comma
        separated list of options. Valid options: [signing,authentication]

Usage:
  karmadactl token create [options]
```

```
[root@master67 karmada]# ./karmadactl token create --print-register-command
karmadactl register 10.10.103.67:6443 --token nmw5kz.39hiaey0sb6oh4yq --discovery-token-ca-cert-hash sha256:75f0484137761fbd345a21aff57c8fecbe0a6953da1012089371452f3dfd013d
[root@master67 karmada]# ./karmadactl token list
TOKEN                     TTL   EXPIRES                USAGES                   DESCRIPTION   EXTRA GROUPS
nmw5kz.39hiaey0sb6oh4yq   23h   2022-08-22T11:06:01Z   authentication,signing   <none>        system:bootstrappers:karmada:default-cluster-token
```

Part of #2282

```release-note
karmadactl: implement `token` command to manage bootstrap tokens
```

### DIFF
--- a/pkg/karmadactl/token.go
+++ b/pkg/karmadactl/token.go
@@ -1,0 +1,378 @@
+package karmadactl
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/spf13/cobra"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/duration"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	kubeclient "k8s.io/client-go/kubernetes"
+	bootstrapapi "k8s.io/cluster-bootstrap/token/api"
+	bootstraputil "k8s.io/cluster-bootstrap/token/util"
+	"k8s.io/klog/v2"
+	"k8s.io/kubectl/pkg/cmd/get"
+	"k8s.io/kubectl/pkg/util/templates"
+
+	"github.com/karmada-io/karmada/pkg/karmadactl/options"
+	tokenutil "github.com/karmada-io/karmada/pkg/karmadactl/util/bootstraptoken"
+)
+
+var (
+	tokenLong = templates.LongDesc(`
+	This command manages bootstrap tokens. It is optional and needed only for advanced use cases.
+
+	In short, bootstrap tokens are used for establishing bidirectional trust between a client and a server.
+	A bootstrap token can be used when a client (for example a member cluster that is about to join control plane) needs
+	to trust the server it is talking to. Then a bootstrap token with the "signing" usage can be used.
+	bootstrap tokens can also function as a way to allow short-lived authentication to the API Server
+	(the token serves as a way for the API Server to trust the client), for example for doing the TLS Bootstrap.
+
+	What is a bootstrap token more exactly?
+	 - It is a Secret in the kube-system namespace of type "bootstrap.kubernetes.io/token".
+	 - A bootstrap token must be of the form "[a-z0-9]{6}.[a-z0-9]{16}". The former part is the public token ID,
+	   while the latter is the Token Secret and it must be kept private at all circumstances!
+	 - The name of the Secret must be named "bootstrap-token-(token-id)".
+
+	This command is same as 'kubeadm token', but it will create tokens that are used by member clusters.`)
+
+	tokenExamples = templates.Examples(`
+	# Create a token and print the full '%[1]s register' flag needed to join the cluster using the token.
+	%[1]s token create --print-register-command
+	`)
+)
+
+// NewCmdToken returns cobra.Command for token management
+func NewCmdToken(karmadaConfig KarmadaConfig, parentCommand string, streams genericclioptions.IOStreams) *cobra.Command {
+	opts := &CommandTokenOptions{
+		parentCommand: parentCommand,
+		TTL: &metav1.Duration{
+			Duration: 0,
+		},
+	}
+
+	cmd := &cobra.Command{
+		Use:     "token",
+		Short:   "Manage bootstrap tokens",
+		Long:    tokenLong,
+		Example: fmt.Sprintf(tokenExamples, parentCommand),
+	}
+
+	cmd.AddCommand(NewCmdTokenCreate(karmadaConfig, streams.Out, opts))
+	cmd.AddCommand(NewCmdTokenList(karmadaConfig, streams.Out, streams.ErrOut, opts))
+	cmd.AddCommand(NewCmdTokenDelete(karmadaConfig, streams.Out, opts))
+
+	return cmd
+}
+
+// CommandTokenOptions holds all command options for token
+type CommandTokenOptions struct {
+	// global flags
+	options.GlobalCommandOptions
+
+	TTL         *metav1.Duration
+	Description string
+	Groups      []string
+	Usages      []string
+
+	PrintRegisterCommand bool
+
+	parentCommand string
+}
+
+// NewCmdTokenCreate returns cobra.Command to create token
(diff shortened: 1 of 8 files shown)


RainbowMango:
retest not working. 
Manually re-triggered.

pkg/karmadactl/token.go
```diff
@@ -0,0 +1,725 @@
...
+	This command manages bootstrap tokens. It is optional and needed only for advanced use cases.
+
+	In short, bootstrap tokens are used for establishing bidirectional trust between a client and a server.
+	A bootstrap token can be used when a client (for example a member cluster that is about to join control plane) needs
+	to trust the server it is talking to. Then a bootstrap token with the "signing" usage can be used.
+	bootstrap tokens can also function as a way to allow short-lived authentication to the API Server
+	(the token serves as a way for the API Server to trust the client), for example for doing the TLS Bootstrap.
+
+	What is a bootstrap token more exactly?
+	 - It is a Secret in the kube-system namespace of type "bootstrap.kubernetes.io/token".
```
RainbowMango, line 54: Can you confirm if the Secret will be created in `kube-system` namespace?

pkg/karmadactl/util/pubkeypin/pubkeypin.go
```diff
@@ -0,0 +1,122 @@
+/*
+Copyright 2017 The Kubernetes Authors.
```
RainbowMango, line 2: The reason why you copy the files is due to the concern of huge dependencies, right?

Can we put it under `pkg/util/lifted`?

pkg/karmadactl/token.go
lonelyCZ: Yes, the `tokencleaner` controller only delete expired tokens in `kube-system` namespace.

pkg/karmadactl/util/pubkeypin/pubkeypin.go
lonelyCZ: > The reason why you copy the files is due to the concern of huge dependencies, right?

Yes, `kubeadm` doesn't provide a standard dependent package, only can import it with `k8s.io/kubernetes` that is not recommended.

> Can we put it under pkg/util/lifted?

Sure, I only think the work of putting it to `pkg/util/lifted` is a lit redious~, Hhh.

lonelyCZ:
Please continue to review this pr, if there are some problem, I can fix them at the same time. @RainbowMango

pkg/karmadactl/util/pubkeypin/pubkeypin.go
RainbowMango: You can move the whole package there.

RainbowMango:
> Please continue to review this pr, if there are some problem, I can fix them at the same time. @RainbowMango

Generally looks fine now.

lonelyCZ:
/unhold

PTAL @RainbowMango

pkg/karmadactl/util/bootstraptoken/bootstraptoken.go
```diff
@@ -0,0 +1,353 @@
...
+	// used for authentication
+	// +optional
+	Groups []string
+}
+
+// BootstrapTokenString is a token of the format abcdef.abcdef0123456789 that is used
+// for both validation of the practically of the API server from a joining cluster's point
+// of view and as an authentication method for the cluster in the bootstrap phase of
+// "karmadactl join". This token is and should be short-lived
+type BootstrapTokenString struct {
```
RainbowMango, line 81: Maybe we can rename it to `TokenString` or `String`.

RainbowMango: By the way, the comment seems hard to understand...

pkg/karmadactl/util/bootstraptoken/bootstraptoken.go
```diff
@@ -0,0 +1,353 @@
...
+	if len(substrs) != validatedSubstringsSize {
+		return nil, fmt.Errorf("the bootstrap token %q was not of the form %q", token, bootstrapapi.BootstrapTokenPattern)
+	}
+
+	return &BootstrapTokenString{ID: substrs[1], Secret: substrs[2]}, nil
+}
+
+// BootstrapTokenToSecret converts the given BootstrapToken object to its Secret representation that
+// may be submitted to the API Server in order to be stored.
+func BootstrapTokenToSecret(bt *BootstrapToken) *corev1.Secret {
```
RainbowMango, line 178: Maybe we can rename to `TokenToSecret` or `ToSecret`.

pkg/karmadactl/util/bootstraptoken/bootstraptoken.go
```diff
@@ -0,0 +1,353 @@
...
+}
+
+// NewBootstrapTokenStringFromIDAndSecret is a wrapper around NewBootstrapTokenString
+// that allows the caller to specify the ID and Secret separately
+func NewBootstrapTokenStringFromIDAndSecret(id, secret string) (*BootstrapTokenString, error) {
+	return NewBootstrapTokenString(bootstraputil.TokenFromIDAndSecret(id, secret))
+}
+
+// BootstrapTokenFromSecret returns a BootstrapToken object from the given Secret
+func BootstrapTokenFromSecret(secret *corev1.Secret) (*BootstrapToken, error) {
```
RainbowMango, line 233:
```suggestion
func FromSecret(secret *corev1.Secret) (*BootstrapToken, error) {
```

lonelyCZ: Thanks for your suggestions, I have updated them, hope to succeed.

RainbowMango (approved):
/lgtm
/approve

PR #2413, opened 2022-08-24
fix some problems in proxy

ikaven1024:
**What type of PR is this?**

/kind api-change
/kind bug

1. Fix proxy get request will modify data in cache
2. change the error when creating member cluster resource.

Fixes #

```release-note
When creating resource defined in ResourceRegistry, `MethodNotSupported` error will be responsed
```

### DIFF
--- a/pkg/search/proxy/store/multi_cluster_cache.go
+++ b/pkg/search/proxy/store/multi_cluster_cache.go
@@ -146,7 +146,8 @@ func (c *MultiClusterCache) Get(ctx context.Context, gvr schema.GroupVersionReso
 		return nil, err
 	}
 
-	accessor, err := meta.Accessor(obj)
+	cloneObj := obj.DeepCopyObject()
+	accessor, err := meta.Accessor(cloneObj)
 	if err != nil {
 		return nil, err
 	}
@@ -154,7 +155,7 @@ func (c *MultiClusterCache) Get(ctx context.Context, gvr schema.GroupVersionReso
 	mrv := newMultiClusterResourceVersionWithCapacity(1)
 	mrv.set(clusterName, accessor.GetResourceVersion())
 	accessor.SetResourceVersion(mrv.String())
-	return obj, err
+	return cloneObj, err
 }
 
 // List returns the object list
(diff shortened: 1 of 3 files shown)


XiShanYongYe-Chang (commented):
Thanks!
/lgtm

In fact, we don't change the API in this pr.
/remove-kind api-change

Website docs is updated by https://github.com/karmada-io/website/pull/140 at the same time.

RainbowMango (approved):
/lgtm
/approve

pkg/search/proxy/store/multi_cluster_cache.go
```diff
@@ -146,15 +146,16 @@ func (c *MultiClusterCache) Get(ctx context.Context, gvr schema.GroupVersionReso
...
+	accessor, err := meta.Accessor(cloneObj)
 	if err != nil {
 		return nil, err
 	}
 
 	mrv := newMultiClusterResourceVersionWithCapacity(1)
 	mrv.set(clusterName, accessor.GetResourceVersion())
 	accessor.SetResourceVersion(mrv.String())
-	return obj, err
+	return cloneObj, err
```
liblikewhen, line 158: I think you don't need copy it every time when call the get method,  just copy it whenever need to modify the object

ikaven1024: `accessor.SetResourceVersion(mrv.String())`  it is just modifying it.

liblikewhen: so, why not deepcopy the object when call the accessor.SetResourceVersion(mrv.String())

ikaven1024: > so, why not deepcopy the object when call the accessor.SetResourceVersion(mrv.String())

Do you mean as below? I don't think it has too much different. 
```go
	mrv := newMultiClusterResourceVersionWithCapacity(1)
	mrv.set(clusterName, accessor.GetResourceVersion())

	cloneObj := obj.DeepCopyObject()
	accessor, err := meta.Accessor(cloneObj)
	if err != nil {
		return nil, err
	}
	accessor.SetResourceVersion(mrv.String())
	return cloneObj, err
```

liblikewhen: yeah. I think it will reduce the memory copy and save some memory

ikaven1024: In fact, err will not occur normally, and copy will be always done

PR #2097, opened 2022-06-30
Standardize the generation and management of certificates

lonelyCZ:
Signed-off-by: lonelyCZ <531187475@qq.com>

**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

Standardize the generation and management of certificates to prepare for certificate authentication and certificate rotation.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #2423, opened 2022-08-24
fix karmada chart kube-controller cluster-signing config error

wawa0210:
Signed-off-by: ndx-robot <xiao.zhang@daocloud.io>

**What type of PR is this?**

/kind bug

At present, in the karmada charts, the kube-controller configuration is incorrect, which causes the certificate to not work properly after it is issued.

Fixes #

```release-note

```

### DIFF
--- a/charts/karmada/templates/kube-controller-manager.yaml
+++ b/charts/karmada/templates/kube-controller-manager.yaml
@@ -57,8 +57,8 @@ spec:
             - --client-ca-file=/etc/karmada/pki/server-ca.crt
             - --cluster-cidr=10.244.0.0/16
             - --cluster-name=karmada
-            - --cluster-signing-cert-file=/etc/karmada/pki/karmada.crt
-            - --cluster-signing-key-file=/etc/karmada/pki/karmada.key
+            - --cluster-signing-cert-file=/etc/karmada/pki/server-ca.crt
+            - --cluster-signing-key-file=/etc/karmada/pki/server-ca.key
             - --controllers=namespace,garbagecollector,serviceaccount-token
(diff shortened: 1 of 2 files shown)


RainbowMango:
Does the CI hasn't cover this case? @Poor12

Poor12:
> Does the CI hasn't cover this case? @Poor12

Now for the ct tool provided by Helm, it regards the success of helm install as whether it is successful or not, and cannot use the status of a Pod in helm as the basis for CI to pass.  By the way, chart-lint will only run when files in chart directory have changed.

charts/karmada/templates/kube-controller-manager.yaml
```diff
@@ -57,8 +57,8 @@ spec:
             - --client-ca-file=/etc/karmada/pki/server-ca.crt
             - --cluster-cidr=10.244.0.0/16
             - --cluster-name=karmada
-            - --cluster-signing-cert-file=/etc/karmada/pki/karmada.crt
-            - --cluster-signing-key-file=/etc/karmada/pki/karmada.key
+            - --cluster-signing-cert-file=/etc/karmada/pki/server-ca.crt
+            - --cluster-signing-key-file=/etc/karmada/pki/server-ca.key
```
lonelyCZ, line 61: Nice work, it refer https://github.com/karmada-io/karmada/pull/2097#discussion_r914557504

wawa0210: We found this bug in testing

RainbowMango (approved):
/lgtm
/approve

Sorry for let this sit.

PR #2424, opened 2022-08-24
karmada controller-plane supports the ability to issue certificates via csr

wawa0210:
**What type of PR is this?**

/kind feature

At present, karmada controller-plane does not support the ability to issue certificates through csr, but we can access karmada controller-plane by issuing certificates with different permissions through csr, which is very meaningful

Fixes #

```release-note

```

### DIFF
--- a/artifacts/deploy/kube-controller-manager.yaml
+++ b/artifacts/deploy/kube-controller-manager.yaml
@@ -42,7 +42,7 @@ spec:
             - --cluster-name=karmada
             - --cluster-signing-cert-file=/etc/karmada/pki/ca.crt
             - --cluster-signing-key-file=/etc/karmada/pki/ca.key
-            - --controllers=namespace,garbagecollector,serviceaccount-token,ttl-after-finished
+            - --controllers=namespace,garbagecollector,serviceaccount-token,ttl-after-finished,bootstrapsigner,csrapproving,csrcleaner,csrsigning
             - --kubeconfig=/etc/kubeconfig
             - --leader-elect=true
             - --node-cidr-mask-size=24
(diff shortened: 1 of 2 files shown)


RainbowMango:
cc @lonelyCZ

artifacts/deploy/kube-controller-manager.yaml
```diff
@@ -42,7 +42,7 @@ spec:
             - --cluster-name=karmada
             - --cluster-signing-cert-file=/etc/karmada/pki/ca.crt
             - --cluster-signing-key-file=/etc/karmada/pki/ca.key
-            - --controllers=namespace,garbagecollector,serviceaccount-token,ttl-after-finished
+            - --controllers=namespace,garbagecollector,serviceaccount-token,ttl-after-finished,csrapproving,csrcleaner,csrsigning
```
lonelyCZ, line 45: It is no problem to only support `csr` now, and I add a document about `car` at https://github.com/karmada-io/website/pull/142.

Perhaps we can enable `bootstrapsigner` and `tokencleaer` as well in the  future. Like https://github.com/karmada-io/karmada/blob/23ec2afa8b4e09edcf0c7352b11bd477eab7593d/pkg/karmadactl/cmdinit/kubernetes/deployments.go#L270

But currently, only karmada control palne that is created by `karmadactl init` can use `karmadactl register` command.

wawa0210: It would be better if the cluster configuration installed in the karamda helm chart mode and the cluster configuration installed in the karmadactl mode are consistent

Poor12: It quite makes sense. Now the community has several different ways. If we need to add new features in the future, we may need to modify the yaml in multiple places, which greatly increases our maintenance cost and also leads to the fork of each version. We may need to develop a good strategy for maintaining source files for different installations.

wawa0210:
> Hi @wawa0210 This branch has a conflict after #2423. Could you please help to resolve it?

done

RainbowMango (approved):
/lgtm
/approve

PR #2449, opened 2022-08-30
[karmada-scheduler] don't consider the older cluster when filtering with taint_toleration

XiShanYongYe-Chang:

**What type of PR is this?**

/kind feature

When we use `taint_toleration` plug-in in the scheduler to filter clusters, we should ignore those clusters that have been scheduled and leave them to the `taint-manager` for processing.  This way, the cluster failover eviction will be performed only in the `taint-manager`.

Fixes #

```release-note
karmada-scheduler: don't consider the older cluster when filtering with taint_toleration
```

### DIFF
--- a/pkg/scheduler/framework/plugins/tainttoleration/taint_toleration.go
+++ b/pkg/scheduler/framework/plugins/tainttoleration/taint_toleration.go
@@ -34,7 +34,15 @@ func (p *TaintToleration) Name() string {
 }
 
 // Filter checks if the given tolerations in placement tolerate cluster's taints.
-func (p *TaintToleration) Filter(ctx context.Context, placement *policyv1alpha1.Placement, resource *workv1alpha2.ObjectReference, cluster *clusterv1alpha1.Cluster) *framework.Result {
+func (p *TaintToleration) Filter(ctx context.Context, placement *policyv1alpha1.Placement,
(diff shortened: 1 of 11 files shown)


pkg/scheduler/framework/plugins/tainttoleration/taint_toleration.go
```diff
@@ -34,7 +34,17 @@ func (p *TaintToleration) Name() string {
...
+func (p *TaintToleration) Filter(ctx context.Context, placement *policyv1alpha1.Placement,
+	bindingSpec *workv1alpha2.ResourceBindingSpec, cluster *clusterv1alpha1.Cluster) *framework.Result {
+	// The current plug-in does not determine whether to filter
+	// target cluster that already exist in the scheduling result.
+	// Instead, the taint-manager processes the clusters.
+	for _, c := range bindingSpec.Clusters {
+		if c.Name == cluster.Name {
+			return framework.NewResult(framework.Success)
+		}
+	}
```
RainbowMango, line 46:
```suggestion
	// skip the filter if the cluster is already in the list of scheduling results,
	// if the workload referencing by the binding can't tolerate the taint,
	// the taint-manager will evict it after a graceful period.
	if bindingSpec.TargetContains(cluster.Name) {
		return framework.NewResult(framework.Success)
	}
```

RainbowMango (approved):
/lgtm
/approve

I'm not sure it's the best choice but for now, it should be a workable solution.

PR #2450, opened 2022-08-30
fix karmadactl init pending when k8s is in ipv6 mode

duanmengkk:

**What type of PR is this?**

Fixes #

If k8s is in ipv6 mode, then the apiserverUrl of karmada should in `http://[ipv6_address]:32443`  format.

```release-note
NONE
```

### DIFF
--- a/pkg/karmadactl/cmdinit/kubernetes/deploy.go
+++ b/pkg/karmadactl/cmdinit/kubernetes/deploy.go
@@ -432,15 +432,12 @@ func (i *CommandInitOption) RunInit(parentCommand string) error {
 	}
 
 	// Create karmada kubeconfig
-	serverURL := fmt.Sprintf("https://%s:%v", i.KarmadaAPIServerIP[0].String(), i.KarmadaAPIServerNodePort)
-	if err := utils.WriteKubeConfigFromSpec(serverURL, options.UserName, options.ClusterName, i.KarmadaDataPath, options.KarmadaKubeConfigName,
-		i.CertAndKeyFileData[fmt.Sprintf("%s.crt", options.CaCertAndKeyName)], i.CertAndKeyFileData[fmt.Sprintf("%s.key", options.KarmadaCertAndKeyName)],
-		i.CertAndKeyFileData[fmt.Sprintf("%s.crt", options.KarmadaCertAndKeyName)]); err != nil {
-		return fmt.Errorf("failed to create karmada kubeconfig file. %v", err)
+	err := i.createKarmadaConfig()
+	if err != nil {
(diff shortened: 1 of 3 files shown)


duanmengkk:
Before
image
image

After
image
image

and test ok 
image

RainbowMango:
> If k8s is in ipv6 mode

How do we know whether K8s in IPv6 or IPv4? Just from the kubeconfig file?

duanmengkk:
> > If k8s is in ipv6 mode
> 
> How do we know whether K8s in IPv6 or IPv4? Just from the kubeconfig file?

`karmadactl init` get `masterIP` from clientset rather than kubeconfig [https://github.com/karmada-io/karmada/blob/12e8f01d01571932e6fe45cb7f0d1bffd2e40fd9/pkg/karmadactl/cmdinit/kubernetes/node.go#L33](url). And use it to generate kubeconfig of karmada control plane(karmada-apiserver.config). If the `masterIP` is ipv6 ,it should be in the format of `https://[masterIP]:443` ranther than` https://masterIP:443`

lonelyCZ:
Sorry, I am not familar with ipv6 and it is difficult to test it in my env.

/assign @RainbowMango

RainbowMango:
No worries, I'll look at it.

RainbowMango:
Echo from [wikiversity](https://en.wikiversity.org/wiki/IPv6#:~:text=digits%20de52%3A32fd.-,Use%20in%20URLs,db8%3A%3A1%5D%3A80.): 

> Because IPv6 addresses contain colons, they cannot be directly used in URLs because the colons would conflict with both the protocol declaration (http:// or https://) and port numbers. Therefore, when a literal IPv6 address is used, it is encased in a bracket, like [http://[2001:db8::1]:80](http://[2001:db8::1]/)

This PR transforms an IPv6 address to the format that the URL requires.

RainbowMango (approved):
/lgtm
/approve

RainbowMango:
When looking into the code, I found now we just take `the first address` from the node. 
https://github.com/karmada-io/karmada/blob/1af078cd4a97b01924e158a93246eb82a5f3af32/pkg/karmadactl/cmdinit/kubernetes/node.go#L48

I wonder if we should get the most available address, `ExternalIP` > `InternalIP`, I guess taking an external IP is better than an internal IP. @prodanlabs what do you think?

duanmengkk:
Should cherry pick it to other releases ?currently we use 1.2 @RainbowMango

RainbowMango:
Sure, please help do to release-1.1/1.2/1.3.
By the way, please add a release-note to this PR, it will be copied to cherry-pick PRs.

I think this patch is more like a bug fix.
/kind bug

Issue #1847, opened 2022-05-20
The PULL mode cluster cannot be deleted immediately


lonelyCZ:
**What happened**:
When the status of PULL mode cluster is `Unknow`

```
[root@master67 karmada]# kubectl get clusters
NAME      VERSION   MODE   READY     AGE
member1   v1.22.3   Pull   Unknown   5d18h
member2   v1.19.1   Push   True      14d
```

The deleting process is stuck.
```
[root@master67 karmada]# kubectl delete clusters member1
cluster.cluster.karmada.io "member1" deleted
^C
```

Then needing to remove manually `finalizers` filed of `cluster` and all `work` objects, the cluster can be deleted.
```
[root@master67 karmada]# kubectl get work -A
NAMESPACE            NAME                              APPLIED   AGE
karmada-es-member1   cert-manager-674b866f4c           True      5d18h
karmada-es-member1   karmada-impersonator-7cbb6bd5c9   True      5d18h
karmada-es-member1   karmada-impersonator-84f8c8f8c6   True      5d18h
karmada-es-member1   nginx-687f7fb96f                  True      5d18h
```

```
apiVersion: cluster.karmada.io/v1alpha1
kind: Cluster
metadata:
  creationTimestamp: "2022-05-05T04:12:42Z"
  finalizers:
  - karmada.io/cluster-controller
```

```
apiVersion: work.karmada.io/v1alpha1
kind: Work
metadata:
  creationTimestamp: "2022-05-05T04:12:42Z"
  finalizers:
  - karmada.io/execution-controller
```

**What you expected to happen**:
The PULL mode cluster can be deleted immediately, not need to remove `finalizers` manually.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Karmada version: master
- kubectl-karmada or karmadactl version (the result of `kubectl-karmada version` or `karmadactl version`): master
- kubectl version: 1.23.5
- Others:

RainbowMango:
> The PULL mode cluster can be deleted immediately, not need to remove finalizers manually.

I doubt it. Let me guess, when you are deleting the cluster by `kubectl delete clusters member1` what you really want to do is `unregister` the cluster from Karmada, then I'd say it's not the recommended way to do that. You should turn to `karmadactl unjoin` command.

lonelyCZ:
> I'd say it's not the recommended way to do that.

I think so, but this way is shown  in the usage document.

[...]

PR #2468, opened 2022-09-02
Update work status when a cluster becomes (not) ready

dddddai:
**What type of PR is this?**
/kind bug

Now when a cluster goes down, the status of resources will stay the same as before, instead of being correctly reflected in the control plane
This pr tries making use of [WorkAvailable](https://github.com/karmada-io/karmada/blob/v1.3.0/pkg/apis/work/v1alpha1/work_types.go#L128) to solve the problem

Fixes #

Test in my env:
1. A deploy is running on member1
```
[root@test karmada]# kubectl get cluster --context karmada-apiserver
NAME      VERSION   MODE   READY   AGE
member1   v1.24.0   Push   True    3h56m

[root@test karmada]# kubectl get deploy --context karmada-apiserver
nginx   1/1     1            1           2m11s
```

2. Then member1 becomes not ready, the deploy status will be correctly reflected in the control plane
```
[root@test karmada]# kubectl get cluster --context karmada-apiserver
NAME      VERSION   MODE   READY   AGE
member1   v1.24.0   Push   False   4h3m

[root@test karmada]# kubectl get deploy --context karmada-apiserver
nginx   0/1     0            0           9m34s

[root@test karmada]# kubectl get work -A --context karmada-apiserver
NAMESPACE            NAME               APPLIED   AVAILABLE   AGE
karmada-es-member1   nginx-687f7fb96f   True      Unknown     9m33s
```

```release-note
NONE
```

### DIFF
--- a/charts/karmada/_crds/bases/work.karmada.io_works.yaml
+++ b/charts/karmada/_crds/bases/work.karmada.io_works.yaml
@@ -21,6 +21,9 @@ spec:
     - jsonPath: .status.conditions[?(@.type=="Applied")].status
       name: Applied
       type: string
+    - jsonPath: .status.conditions[?(@.type=="Available")].status
+      name: Available
+      type: string
     - jsonPath: .metadata.creationTimestamp
       name: Age
       type: date
--- a/pkg/apis/work/v1alpha1/work_types.go
+++ b/pkg/apis/work/v1alpha1/work_types.go
@@ -21,6 +21,7 @@ const (
 // +kubebuilder:subresource:status
 // +kubebuilder:resource:categories={karmada-io}
 // +kubebuilder:printcolumn:JSONPath=`.status.conditions[?(@.type=="Applied")].status`,name="Applied",type=string
+// +kubebuilder:printcolumn:JSONPath=`.status.conditions[?(@.type=="Available")].status`,name="Available",type=string
 // +kubebuilder:printcolumn:JSONPath=`.metadata.creationTimestamp`,name="Age",type=date
 
 // Work defines a list of resources to be deployed on the member cluster.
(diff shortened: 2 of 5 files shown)


lonelyCZ:
It seemingly can resolve #1847 if we add a `work_controller` in control plane to reconcile the `Available` status of the works.

RainbowMango:
> Now when a cluster goes down, the status of resources will stay the same as before, 

I think we need more discussion about the behavior in case of the cluster down. If we can't get state updates from member clusters, and then we `clean up` the legacy status, does this is what users expected? 

By the way, if the fail-over feature is enabled, the legacy status will be cleaned up automatically as the failure cluster will be removed from RB/CRB, then, the legacy status will be removed subsequently.

lonelyCZ:
> If we can't get state updates from member clusters, and then we clean up the legacy status, does this is what users expected?

I think it seems no problem to set the `works` status to `Unknown` that only represent it is not updated by `execution_controller`.

dddddai:
> If we can't get state updates from member clusters, and then we `clean up` the legacy status, does this is what users expected?

This is also what I've been worried about, I do agree we should be careful

> I think it seems no problem to set the `works` status to `Unknown` that only represent it is not updated by `execution_controller`.

Would be great if this can help #1847, so do you think it's worth introducing `WorkAvailable` regardless of the resource status stuff?

lonelyCZ:
> so do you think it's worth introducing WorkAvailable regardless of the resource status stuff?

Yes, otherwise we can't get any status of the works if the cluster is not ready and the status controller should be executed in control plane for both `PUSH` and `PULL` mode. Certainly, more detail need to be discussed.

PR #2482, opened 2022-09-07
fix deploy-scheduler-estimator.sh cannot generate correct kubeconfig file problem

cmicat:
fix deploy-scheduler-estimator.sh cannot generate correct kubeconfig file problem

Signed-off-by: raymondmiaochaoyue <raymondmiaochaoyue@didichuxing.com>

**What type of PR is this?**

/kind bug

When used as [Cluster Accurate Scheduler Estimator For Rescheduling](https://karmada.io/docs/userguide/scheduling/scheduler-estimator) guided, the original `hack/deploy-scheduler-estimator.sh` have two problems:

(1) The <MEMBER_CLUSTER_KUBECONFIG> file is not flattened, and the `kubectl create secret generic` command will copy the <MEMBER_CLUSTER_KUBECONFIG> file's content directly without changing anything. So if the <MEMBER_CLUSTER_KUBECONFIG> contains none embedded cert reference like the following, karmada-scheduler-estimator will not be able to read the certificate file. Because the certificate file is not mounted by the container.

```
client-certificate: /home/raymondmiaochaoyue/.minikube/profiles/ray-karmada-member2/client.crt
```

(2) <MEMBER_CLUSTER_KUBECONFIG> is used as secret's content, it will not be changed by `--context=` parameter in `kubectl create secret generic` command. We need change it first, then save to temporary file.

None

```release-note
NONE
```

### DIFF
--- a/hack/deploy-scheduler-estimator.sh
+++ b/hack/deploy-scheduler-estimator.sh
@@ -49,11 +49,23 @@ then
 fi
 MEMBER_CLUSTER_NAME=$4
 
+TEMP_PATH="$(mktemp -d)"
+MEMBER_CLUSTER_KUBECONFIG_NAME="$(basename "${MEMBER_CLUSTER_KUBECONFIG}")"
+# --context & --minify will generate minified kubeconfig file with required context
+# --flatten will embed certificate
+kubectl config view --kubeconfig "${MEMBER_CLUSTER_KUBECONFIG}" \
+	--context "${MEMBER_CLUSTER_NAME}" --minify  --flatten \
+	> "${TEMP_PATH}/${MEMBER_CLUSTER_KUBECONFIG_NAME}"
+
 # check whether the kubeconfig secret has been created before
 if ! kubectl --kubeconfig="${HOST_CLUSTER_KUBECONFIG}" --context="${HOST_CLUSTER_NAME}" get secrets -n karmada-system | grep "${MEMBER_CLUSTER_NAME}-kubeconfig"; then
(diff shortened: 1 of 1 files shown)


hack/deploy-scheduler-estimator.sh
```diff
@@ -49,11 +49,23 @@ then
 fi
 MEMBER_CLUSTER_NAME=$4
 
+TEMP_PATH=$(mktemp -d)
+MEMBER_CLUSTER_KUBECONFIG_NAME=$(basename ${MEMBER_CLUSTER_KUBECONFIG})
```
ikaven1024, line 53:
```suggestion
MEMBER_CLUSTER_KUBECONFIG_NAME=$(basename "${MEMBER_CLUSTER_KUBECONFIG}")
```

ikaven1024 (changes requested):

hack/deploy-scheduler-estimator.sh
cmicat: > fixed

RainbowMango:
/assign
Thanks in advance. I'll look at it tomorrow.

RainbowMango (approved):
Fantastic!!!
Thanks for doing this.

I don't know these flags before.
```
    --flatten=false:
	Flatten the resulting kubeconfig file into self-contained output (useful for creating portable kubeconfig
	files)
    --minify=false:
	Remove all information not used by current-context from the output
```

/lgtm
/approve

PR #2489, opened 2022-09-08
Fixed pod can not be collected issue when building resource summary

Poor12:

**What type of PR is this?**
/kind bug

When transformFunc ignores name and namespace of Pod，informer will only save one Pod.
Test Result After this PR:
```
nodeSummary:
    readyNum: 2
    totalNum: 2
....
 allocated:
      cpu: 1050m
      memory: 340Mi
      pods: "12"
```

Fixes #

```
`karmada-controller-manager`/`karmada-agent`: Fixed pod information can not be collected issue when building resource summary.
```

### DIFF
--- a/pkg/util/fedinformer/transform.go
+++ b/pkg/util/fedinformer/transform.go
@@ -59,6 +59,10 @@ func PodTransformFunc(obj interface{}) (interface{}, error) {
 	}
 
 	aggregatedPod := &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      pod.Name,
+			Namespace: pod.Namespace,
+		},
 		Spec: corev1.PodSpec{
 			NodeName:       pod.Spec.NodeName,
 			InitContainers: pod.Spec.InitContainers,
--- a/pkg/util/fedinformer/transform_test.go
+++ b/pkg/util/fedinformer/transform_test.go
@@ -169,7 +169,7 @@ func TestPodTransformFunc(t *testing.T) {
 		want interface{}
 	}{
 		{
-			name: "transform pods without needed",
+			name: "transform pods without status",
 			obj: &corev1.Pod{
 				ObjectMeta: metav1.ObjectMeta{
 					Namespace:   "foo",
@@ -183,10 +183,15 @@ func TestPodTransformFunc(t *testing.T) {
 					},
 				},
 			},
-			want: &corev1.Pod{},
+			want: &corev1.Pod{
+				ObjectMeta: metav1.ObjectMeta{
+					Namespace: "foo",
+					Name:      "bar",
+				},
+			},
 		},
 		{
-			name: "transform pods with needed",
+			name: "transform pods with status",
 			obj: &corev1.Pod{
 				ObjectMeta: metav1.ObjectMeta{
 					Namespace: "foo",
@@ -208,6 +213,10 @@ func TestPodTransformFunc(t *testing.T) {
 				},
 			},
 			want: &corev1.Pod{
+				ObjectMeta: metav1.ObjectMeta{
+					Namespace: "foo",
+					Name:      "bar",
+				},
 				Spec: corev1.PodSpec{
 					NodeName:       "test",
 					InitContainers: []corev1.Container{{Name: "test"}},


pkg/util/fedinformer/transform_test.go
```diff
@@ -196,6 +201,7 @@ func TestPodTransformFunc(t *testing.T) {
 					NodeName:       "test",
 					InitContainers: []corev1.Container{{Name: "test"}},
 					Containers:     []corev1.Container{{Name: "test"}},
+					RestartPolicy:  corev1.RestartPolicyAlways,
```
RainbowMango, line 204: Why need this?

pkg/util/fedinformer/transform.go
```diff
@@ -59,6 +59,10 @@ func PodTransformFunc(obj interface{}) (interface{}, error) {
 	}
 
 	aggregatedPod := &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
```
RainbowMango, line 62: Are you sure we don't need `APIVersion` and `Kind` here? Why, can you show me the code in Kubernetes about how to store the object in the cache?

pkg/util/fedinformer/transform_test.go
Poor12: The former test case mainly want to test whether this method can filter metadata information. When I add the metadata, I found that the expected and wanted are the same. Adding this field mainly wants to verify that this method works.

pkg/util/fedinformer/transform.go
Poor12: First, let's find where to use this transformFunc:
```
func processDeltas(
	// Object which receives event notifications from the given deltas
	handler ResourceEventHandler,
	clientState Store,
	transformer TransformFunc,
	deltas Deltas,
) error {
	// from oldest to newest
	for _, d := range deltas {
		obj := d.Object
		if transformer != nil {
			var err error
			obj, err = transformer(obj)
			if err != nil {
				return err
			}
		}

		switch d.Type {
		case Sync, Replaced, Added, Updated:
			if old, exists, err := clientState.Get(obj); err == nil && exists {
				if err := clientState.Update(obj); err != nil {
					return err
				}
				handler.OnUpdate(old, obj)
			} else {
				if err := clientState.Add(obj); err != nil {
					return err
				}
				handler.OnAdd(obj)
			}
		case Deleted:
			if err := clientState.Delete(obj); err != nil {
				return err
			}
			handler.OnDelete(obj)
		}
	}
	return nil
}
```
Simply put, when obtaining obj information, first filter the field through transformFunc, and then store it in the cache according to the Delta information.

```
// `*cache` implements Indexer in terms of a ThreadSafeStore and an
// associated KeyFunc.
type cache struct {
	// cacheStorage bears the burden of thread safety for the cache
	cacheStorage ThreadSafeStore
	// keyFunc is used to make the key for objects stored in and retrieved from items, and
	// should be deterministic.
	keyFunc KeyFunc
}
// Add inserts an item into the cache.
func (c *cache) Add(obj interface{}) error {
	key, err := c.keyFunc(obj)
	if err != nil {
		return KeyError{obj, err}
	}
	c.cacheStorage.Add(key, obj)
	return nil
}
```
Cache is simply a key-value map. KeyFunc is mainly defined by `namespace-name`.
```
// MetaNamespaceKeyFunc is a convenient default KeyFunc which knows how to make
// keys for API objects which implement meta.Interface.
// The key uses the format <namespace>/<name> unless <namespace> is empty, then
// it's just <name>
func MetaNamespaceKeyFunc(obj interface{}) (string, error) {
	if key, ok := obj.(ExplicitKey); ok {
		return string(key), nil
	}
	meta, err := meta.Accessor(obj)
	if err != nil {
		return "", fmt.Errorf("object has no meta: %v", err)
	}
	if len(meta.GetNamespace()) > 0 {
		return meta.GetNamespace() + "/" + meta.GetName(), nil
	}
	return meta.GetName(), nil
}
```
So from the test result, when metadata is only a namespace, key is only defined by namespace,  pods in the same namespace will be overrided to the same space, therefore two namespaces will only have two pods. When metadata is empty, key can be regarded as "" and will occupy the the same space. But it's strange there's no error throwing.

Poor12:
cc @RainbowMango

RainbowMango (approved):
/lgtm
/approve

I updated the release note.
Please help to cherry-pick this to release-1.3.

Issue #2504, opened 2022-09-11
init failed and report `namespaces "karmada-system" already exists`


cleverhu:
**What happened**:
I install karmada controller-plane by `karmadactl init --kube-image-registry=registry.cn-hangzhou.aliyuncs.com/google_containers` , before execute this, I have tried `karmadactl deinit`.
The log of install is:
```
 karmadactl init --kube-image-registry=registry.cn-hangzhou.aliyuncs.com/google_containers  [13:36:11 09/11]
I0911 13:36:41.544986 3505120 deploy.go:145] kubeconfig file: /root/.kube/config, kubernetes: https://139.198.167.220:6443
W0911 13:36:41.986808 3505120 node.go:30] the kubernetes cluster does not have a Master role.
I0911 13:36:41.986827 3505120 node.go:38] randomly select 3 Node IPs in the kubernetes cluster.
I0911 13:36:42.080548 3505120 deploy.go:165] karmada apiserver ip: [10.130.59.18 10.130.58.25]
I0911 13:36:42.770834 3505120 cert.go:229] Generate ca certificate success.
I0911 13:36:42.951945 3505120 cert.go:229] Generate karmada certificate success.
I0911 13:36:43.031271 3505120 cert.go:229] Generate apiserver certificate success.
I0911 13:36:43.118985 3505120 cert.go:229] Generate front-proxy-ca certificate success.
I0911 13:36:43.393367 3505120 cert.go:229] Generate front-proxy-client certificate success.
I0911 13:36:43.571701 3505120 cert.go:229] Generate etcd-ca certificate success.
I0911 13:36:43.798231 3505120 cert.go:229] Generate etcd-server certificate success.
I0911 13:36:43.966271 3505120 cert.go:229] Generate etcd-client certificate success.
I0911 13:36:43.966398 3505120 deploy.go:251] download crds file name: /etc/karmada/crds.tar.gz
Downloading...[ 100.00% ]
Downloading...[ 100.00% ]
Download complete.I0911 13:36:45.129184 3505120 deploy.go:441] Create karmada kubeconfig success.
I0911 13:36:45.201297 3505120 namespace.go:36] Create Namespace 'karmada-system' successfully.
I0911 13:36:45.267352 3505120 rbac.go:60] CreateClusterRole karmada-controller-manager success.
I0911 13:36:45.304952 3505120 rbac.go:81] CreateClusterRoleBinding karmada-controller-manager success.
[...]

PR #2466, opened 2022-09-01
Add default path prompt for deinit to delete etcd data

nuclearwu:
Signed-off-by: wuzhongjian <wuzhongjian_yewu@cmss.chinamobile.com>

**What type of PR is this?**
/kind documentation

**What this PR does / why we need it**:
Prompt the user to delete the default path of etcd data to deinit karmada

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:
```release-note
NONE
```

PR #2505, opened 2022-09-11
fix namespace already exist

cleverhu:
**What type of PR is this?**
Fix namespace already exists when use `karmadactl init`

Fixes https://github.com/karmada-io/karmada/issues/2504

```release-note
None
```


pkg/karmadactl/cmdinit/karmada/deploy.go
```diff
@@ -46,12 +47,13 @@ func InitKarmadaResources(dir, caBase64, systemNamespace string) error {
...
-	// create namespace
-	if _, err := clientSet.CoreV1().Namespaces().Create(context.TODO(), &corev1.Namespace{
+	ns := &corev1.Namespace{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: systemNamespace,
 		},
-	}, metav1.CreateOptions{}); err != nil {
+	}
+	// create namespace
+	if _, err := clientSet.CoreV1().Namespaces().Create(context.TODO(), ns, metav1.CreateOptions{}); err != nil && !apierrors.IsAlreadyExists(err) {
```
RainbowMango, line 56: I'm sure this patch could fix this issue. 
This patch blinds the error, I hope we can figure out the root cause before we make a fix.

cleverhu: I test locally. It works. The root cause for it is it has been saved in etcd.

RainbowMango: > The root cause for it is it has been saved in etcd.

Do you mean the `deinit` forgot to remove the `karmada-system` namespace?  Is it still in the etcd?

cleverhu: It is right.

cleverhu: However, `deinit` not only doesn't remove ns `karmada-system`, there are some crds installed also not have been removed, which I also checks for.

lonelyCZ: It is in `karmada-etcd`. There is exactly a problem that it can't be executed secondly after the first execution fail. I suggest to use `CreateOrUpdate()` function to create these resources, which is clearer. Like

https://github.com/karmada-io/karmada/blob/1af078cd4a97b01924e158a93246eb82a5f3af32/pkg/karmadactl/cmdinit/utils/rbac.go#L91-L105

cleverhu: It seems better, but the handler of `etcd data` is still a problem. Which we use the data before or clean up and generadate data for it or someway others.

RainbowMango: Yeah, I get it. The root cause is clear now.

I agree with @lonelyCZ, which seems cleaner. But, can we install karmada on a `legacy etcd`?  What if there are also other resources in it? like `PP`/`Deployments`..., they will be propagated to registered clusters that might be surprised by users.

I wonder if we can clean up the default host-data-path(`/var/lib/karmada-etcd`) in `karmada init`?

lonelyCZ: > Which we use the data before or clean up and generadate data for it or someway others.

It is enough to manage the resources of `init` phase. Ohter resources should be decided by users.

lonelyCZ: > I wonder if we can clean up the default host-data-path(/var/lib/karmada-etcd) in karmada init?

Actually, for `PUSH` mode cluster, it can continue to work, which is benefit to upgrade karmada components.

lonelyCZ: To install by scripts can demostrate. It can be reinstall by scripts without removing `/var/lib/karmada-etcd`, I tested it before.

RainbowMango: sorry, I'm still confused. 


From another perspective, does the user expected to `re-use` the legacy data in ETCD? I guess not.

By the way, do you know what's the behavior in `kubeadm`? What can we learn from it?

RainbowMango: > Perhaps, we also make deinit to delete local ETCD by default

Do you know why we didn't remove the etcd data before? In addition, we added a prompt at #2466 :)

> And for init, add a pre-flight phase.

Seems we don't have the `pre-fflight` thing yet, do you think we should make a quick fix? just like:
https://github.com/karmada-io/karmada/blob/1af078cd4a97b01924e158a93246eb82a5f3af32/pkg/karmadactl/cmdinit/kubernetes/deploy.go#L173-L178

Yeah, I agree, the pre-flight would be great.

lonelyCZ:
> Do you know why we didn't remove the etcd data before?

Because we couldn't certain if schedule the `etcd` to `master node` of karmada-host. It could be scheduled to other node, but we don't execute `deinit` on this node.

RainbowMango:
> Because we couldn't certain if schedule the etcd to master node of karmada-host. It could be scheduled to other node, but we don't execute deinit on this node.

Oh, I see. Then we also have the same problem when we try to delete the legacy etcd. Isn't? Because we are not sure that the etcd was running on the current node before.

lonelyCZ:
> Then we also have the same problem when we try to delete the legacy etcd. Isn't? Because we are not sure that the etcd was running on the current node before.

Yes.

pkg/karmadactl/cmdinit/karmada/deploy.go
cleverhu: I am willing to do my best to make it better and I would appreciate it if you could provide some guidance.

RainbowMango:
Now we have two approaches to solve this issue:
1. Try to delete the etcd data during `deinit` or initialize the etcd data path during `init`.
2. Try to blind the error by using the `create or update`

The disadvantage of the first approach is we can't guarantee the etcd data can be cleaned as we don't know if the data on the node where running the command.

The disadvantage of the second approach is it allows people to reuse the legacy etcd data which is probably not expected.

I prefer the first approach, even though it has its limitation. 
@cleverhu @lonelyCZ  how do you say?

Also, ask @prodanlabs  for comments.

cleverhu:
> Now we have two approaches to solve this issue:
> 
> 1. Try to delete the etcd data during `deinit` or initialize the etcd data path during `init`.
> 
> 2. Try to blind the error by using the `create or update`
> 
> 
> 
> The disadvantage of the first approach is we can't guarantee the etcd data can be cleaned as we don't know if the data on the node where running the command.
> 
> 
> 
> The disadvantage of the second approach is it allows people to reuse the legacy etcd data which is probably not expected.
> 
> 
> 
> I prefer the first approach, even though it has its limitation. 
> 
> @cleverhu @lonelyCZ  how do you say?
> 
> 
> 
> Also, ask @prodanlabs  for comments.

I'd like the first too.Such like "kubeadm reset".

RainbowMango:
> I'd like the first too.Such like "kubeadm reset".

Yes, just make a clean base for the next `init`.

lonelyCZ:
Sorry for delay @cleverhu , I will review ASAP.

/assign

pkg/karmadactl/cmdinit/karmada/deploy.go
```diff
@@ -207,10 +208,17 @@ func createCRDs(crdClient *clientset.Clientset, filename string) error {
 		return err
 	}
 
+	klog.Infof("Attempting to create resource")
```
lonelyCZ, line 211:
```suggestion
	klog.Infof("Attempting to create CRD.")
```

pkg/util/namespace.go
```diff
@@ -72,3 +73,28 @@ func EnsureNamespaceExist(client kubeclient.Interface, namespace string, dryRun
 
 	return createdObj, nil
 }
+
+// CreateOrUpdateNamespace creates a Namespaces if the target resource doesn't exist. If the resource exists already, this function will update the resource instead.
+func CreateOrUpdateNamespace(client kubeclient.Interface, ns *corev1.Namespace) error {
```
lonelyCZ, line 78: Suggest to move this function to `pkg/karmadactl/cmdinit/utils` for the moment.

cleverhu: Ok,I will do it.

cleverhu:
@lonelyCZ I have updated the code, /PTAL.

lonelyCZ:
Hi, @cleverhu , could you please help to resolve the conflicts? BTW, we have moved all `CreateOrUpdateXXX` functions to `pkg/karmadactl/util/idempotency.go`

cleverhu:
> Hi, @cleverhu , could you please help to resolve the conflicts? BTW, we have moved all `CreateOrUpdateXXX` functions to `pkg/karmadactl/util/idempotency.go`

Ok,I will update the code later.

cleverhu:
@lonelyCZ I have updated the code,PTAL.

pkg/karmadactl/cmdinit/kubernetes/namespace.go
```diff
@@ -1,38 +1,11 @@
...
-		},
-	}
-
-	_, err = namespaceClient.Create(context.TODO(), n, metav1.CreateOptions{})
-	if err != nil {
-		return err
-	}
-	klog.Infof("Create Namespace '%s' successfully.", i.Namespace)
-	return nil
+	ns := util.NewNamespace(i.Namespace)
```
lonelyCZ, line 9: It seems that `func (i *CommandInitOption) CreateNamespace() error` is redundancy.

pkg/karmadactl/util/idempotency.go
```diff
@@ -158,3 +159,44 @@ func CreateOrUpdateConfigMap(client *kubeclient.Clientset, cm *corev1.ConfigMap)
...
+}
+
+// CreateOrUpdateNamespace creates a Namespaces if the target resource doesn't exist. If the resource exists already, this function will update the resource instead.
+func CreateOrUpdateNamespace(client kubeclient.Interface, ns *corev1.Namespace) error {
+	if _, err := client.CoreV1().Namespaces().Create(context.TODO(), ns, metav1.CreateOptions{}); err != nil {
+		if !apierrors.IsAlreadyExists(err) {
+			return fmt.Errorf("unable to create Namespace: %v", err)
+		}
+
+		if _, err := client.CoreV1().Namespaces().Update(context.TODO(), ns, metav1.UpdateOptions{}); err != nil {
```
lonelyCZ, line 179: We need to check the `resourceVersion`

like

```
// CreateOrUpdateConfigMap creates a ConfigMap if the target resource doesn't exist. If the resource exists already, this function will update the resource instead.
func CreateOrUpdateConfigMap(client *kubeclient.Clientset, cm *corev1.ConfigMap) error {
	if _, err := client.CoreV1().ConfigMaps(cm.ObjectMeta.Namespace).Create(context.TODO(), cm, metav1.CreateOptions{}); err != nil {
		if !apierrors.IsAlreadyExists(err) {
			return fmt.Errorf("unable to create ConfigMap: %v", err)
		}

		existCm, err := client.AppsV1().Deployments(cm.Namespace).Get(context.TODO(), cm.Name, metav1.GetOptions{})
		if err != nil {
			return err
		}

		cm.ResourceVersion = existCm.ResourceVersion

		if _, err := client.CoreV1().ConfigMaps(cm.ObjectMeta.Namespace).Update(context.TODO(), cm, metav1.UpdateOptions{}); err != nil {
			return fmt.Errorf("unable to update ConfigMap: %v", err)
		}
	}
	klog.V(2).Infof("ConfigMap %s/%s has been created or updated.", cm.ObjectMeta.Namespace, cm.ObjectMeta.Name)

	return nil
}
```

cleverhu: Thank you, I will update the code later.

pkg/karmadactl/util/idempotency.go
```diff
@@ -158,3 +159,58 @@ func CreateOrUpdateConfigMap(client *kubeclient.Clientset, cm *corev1.ConfigMap)
...
+}
+
+// CreateOrUpdateService creates a Service if the target resource doesn't exist. If the resource exists already, this function will update the resource instead.
+func CreateOrUpdateService(client kubernetes.Interface, svc *corev1.Service) error {
+	if _, err := client.CoreV1().Services(svc.Namespace).Create(context.TODO(), svc, metav1.CreateOptions{}); err != nil {
+		if !apierrors.IsAlreadyExists(err) {
+			return fmt.Errorf("unable to create Service: %v", err)
+		}
+
+		existSvc, err := client.CoreV1().Secrets(svc.Namespace).Get(context.TODO(), svc.Name, metav1.GetOptions{})
```
lonelyCZ, line 202:
```suggestion
		existSvc, err := client.CoreV1().Services(svc.Namespace).Get(context.TODO(), svc.Name, metav1.GetOptions{})
```

lonelyCZ:
It works well in my env.

/lgtm
/approve

xu1257:
Before init I removed `/root/.kube/cache/` , but  it still show error
`I1114 17:13:02.749986  121527 services.go:66] service karmada-aggregated-apiserver create successfully.
I1114 17:13:05.786757  121527 check.go:49] pod: karmada-aggregated-apiserver-595dd77955-rcsg2 is ready. status: Running
F1114 17:13:05.797635  121527 deploy.go:55] namespaces "karmada-system" already exists`
And I can't found `/var/lib/karmada-*` in Ubuntu18.04 , how can i do it?

PR #2511, opened 2022-09-13
karmada-controller-manager removes unnecessary permissions

carlory:

**What type of PR is this?**

/kind feature

Karmada-controller manger should not depend on any permission of a host cluster. It only talks with karmada-apiserver.

manifests:

```yaml
      containers:
        - name: karmada-controller-manager
          image: swr.ap-southeast-1.myhuaweicloud.com/karmada/karmada-controller-manager:latest
          imagePullPolicy: IfNotPresent
          command:
            - /bin/karmada-controller-manager
            - --kubeconfig=/etc/kubeconfig
            - --bind-address=0.0.0.0
            - --secure-port=10357
            - --feature-gates=PropagateDeps=true,Failover=true,GracefulEviction=true,CustomizedClusterResourceModeling=true
            - --failover-eviction-timeout=30s
            - --v=4
```

controllerruntime.GetConfig():

```
	// GetConfig creates a *rest.Config for talking to a Kubernetes apiserver.
	// If --kubeconfig is set, will use the kubeconfig file at that location.  Otherwise will assume running
	// in cluster and use the cluster provided kubeconfig.
	//
	// Config precedence
	//
	// * --kubeconfig flag pointing at a file
	//
	// * KUBECONFIG environment variable pointing at a file
	//
	// * In-cluster config if running in cluster
	//
	// * $HOME/.kube/config if exists.
```

```release-note
NONE
```


hack/deploy-karmada.sh
```diff
@@ -149,8 +149,6 @@ kubectl --context="${HOST_CLUSTER_NAME}" apply -f "${REPO_ROOT}/artifacts/deploy
 
 # create service account, cluster role for controller-manager
 kubectl --context="${HOST_CLUSTER_NAME}" apply -f "${REPO_ROOT}/artifacts/deploy/serviceaccount.yaml"
-kubectl --context="${HOST_CLUSTER_NAME}" apply -f "${REPO_ROOT}/artifacts/deploy/clusterrole.yaml"
-kubectl --context="${HOST_CLUSTER_NAME}" apply -f "${REPO_ROOT}/artifacts/deploy/clusterrolebinding.yaml"
```
RainbowMango, line 153: Yes, controller-manager won't talk to `host` cluster. But why keep the `ServiceAccount` here? (just a question)

RainbowMango:
cc @lonelyCZ

hack/deploy-karmada.sh
carlory: @RainbowMango No need to create serviceaccounts for karmada-components, I will remove them in next PR.

https://github.com/kubernetes/api/blob/master/core/v1/types.go#L3153

Do we need to set the value of the AutomountServiceAccountToken field to false for the karmada-components?

```
// AutomountServiceAccountToken indicates whether a service account token should be automatically mounted.
// +optional
```

XiShanYongYe-Chang: > No need to create serviceaccounts for [karmada-components](https://github.com/karmada-io/karmada/blob/master/artifacts/deploy/serviceaccount.yaml), I will remove them in next PR.

I agree with it.

> Do we need to set the value of the AutomountServiceAccountToken field to false for the karmada-components?

What's the impact if we don't set this field? I tried using a v1.23 k8s cluster, and when I did not set this field, the `spec.serviceAccountName` field is not set as a result.

RainbowMango: > What's the impact if we don't set this field?

The `AutomountServiceAccountToken`(acts like defaults to `true`) determines if the token of ServiceAccount(referenced by `.spec.serviceAccountName`) should be mounted on `/var/run/secrets/kubernetes.io/serviceaccount/token`. 

The reason why @carlory is asking is we don't need `ServiceAccount` at all, then we don't need to mount the token either. 

> I tried using a v1.23 k8s cluster, and when I did not set this field, the spec.serviceAccountName field is not set as a result.

I guess you mean the `.spec.serviceAccountName` still is `karmada-controller-manager` not the `default`.

The reason is we redundantly set the `.spec.serviceAccountName`... See:
https://github.com/karmada-io/karmada/blob/1af078cd4a97b01924e158a93246eb82a5f3af32/artifacts/deploy/karmada-controller-manager.yaml#L18

RainbowMango: > Do we need to set the value of the AutomountServiceAccountToken field to false for the karmada-components?

My opinion is yes, we can do it in the next PR when you are going to clean up the service account.

lonelyCZ:
Thanks. The test was fine.

/lgtm

PR #2501, opened 2022-09-09
[E2E] fix flaking test: search proxy get node report not found

XiShanYongYe-Chang:
Signed-off-by: changzhen <changzhen5@huawei.com>

**What type of PR is this?**

/kind failing-test

**What this PR does / why we need it**:

I found a new flaking test in master branch:

https://github.com/karmada-io/karmada/runs/8267105862?check_suite_focus=true

It still report:

```log
"nodes \"member1-control-plane\" not found",
```

We need to use `g` for assert in `gomega.Eventually(func(g gomega.Gomega) {}` block.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

/cc @ikaven1024 @RainbowMango Can you help review it?

PR #2490, opened 2022-09-08
[e2e] add wait for proxy ready

ikaven1024:
Signed-off-by: yingjinhui <yingjinhui@didiglobal.com>

**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
As say in https://github.com/karmada-io/karmada/issues/2488#issuecomment-1240452424, add wait proxy ready before test.

**Which issue(s) this PR fixes**:
Fixes #2488

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:
```release-note
NONE
```

PR #2483, opened 2022-09-07
fix concurrent map writes panic while list via proxy

ikaven1024:
Signed-off-by: yingjinhui <yingjinhui@didiglobal.com>

**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
fix concurrent map writes panic while listing via proxy
```
fatal error: concurrent map writes

goroutine 1414 [running]:
runtime.throw({0x105907244?, 0x140017c8270?})
        /Users/didi/sdk/go1.18.3/src/runtime/panic.go:992 +0x50 fp=0x14000a931e0 sp=0x14000a931b0 pc=0x10414cab0
runtime.mapassign_faststr(0x105f2fb40?, 0x14001a7eea0?, {0x1058f6675, 0xb})
        /Users/didi/sdk/go1.18.3/src/runtime/map_faststr.go:212 +0x3bc fp=0x14000a93250 sp=0x14000a931e0 pc=0x10412819c
k8s.io/apimachinery/pkg/apis/meta/v1/unstructured.setNestedFieldNoCopy(0x14000a93348?, {0x105f2fb40?, 0x140017fb290}, {0x14000a933f8?, 0x140009b3de0?, 0x2?})
        /Users/didi/OpenSouorce/karmada/vendor/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured/helpers.go:228 +0x78 fp=0x14000a932e0 sp=0x14000a93250 pc=0x1046338b8
k8s.io/apimachinery/pkg/apis/meta/v1/unstructured.SetNestedStringMap(0x0?, 0x140017fb260, {0x14000a933f8, 0x2, 0x2})
        /Users/didi/OpenSouorce/karmada/vendor/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured/helpers.go:255 +0x12c fp=0x14000a933b0 sp=0x14000a932e0 pc=0x104633dcc
k8s.io/apimachinery/pkg/apis/meta/v1/unstructured.(*Unstructured).setNestedMap(...)
        /Users/didi/OpenSouorce/karmada/vendor/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured/unstructured.go:175
k8s.io/apimachinery/pkg/apis/meta/v1/unstructured.(*Unstructured).SetAnnotations(0x105f31520?, 0x140017fb260?)
[...]

PR #2514, opened 2022-09-14
Automated cherry pick of #2469: add cache source annotation for objects returned by proxy
#2483: fix concurrent map writes panic while list via proxy
#2490: add wait for proxy ready
#2501: fix flaking test: search proxy get node report not found

ikaven1024:
Cherry pick of #2469 #2483 #2490 #2501 on release-1.3.
#2469: add cache source annotation for objects returned by proxy
#2483: fix concurrent map writes panic while list via proxy
#2490: add wait for proxy ready
#2501: fix flaking test: search proxy get node report not found
```release-note
`karmada-search`: fix concurrent map writes panic while list via proxy; And add cache source annotation to returned objects.
```

### DIFF
--- a/pkg/search/proxy/cluster_proxy.go
+++ b/pkg/search/proxy/cluster_proxy.go
@@ -1,13 +1,17 @@
 package proxy
 
 import (
+	"bytes"
 	"context"
 	"fmt"
+	"io"
+	"io/ioutil"
 	"net/http"
 	"path"
 
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apiserver/pkg/endpoints/request"
 	"k8s.io/apiserver/pkg/registry/rest"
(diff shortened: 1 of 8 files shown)


XiShanYongYe-Chang:
I was wondering if we need to cherry-pick #2469 as well, which is actually a feature. @RainbowMango

RainbowMango:
Yes, I noticed, It's more like a bug because that's the thing we planned but missed from implementation.

ikaven1024:
> I was wondering if we need to cherry-pick #2469 as well, which is actually a feature.

#2469 also fix a bug: updating objects return invalid resource version error.

ikaven1024:
> By the way, we usually pick one bug fix each time, because it's hard to organize the release notes and often conflicts.
> I looked at the patches, seems #2483 and #2469 need release notes.

#2483 introduces a bug, and #2469 fix it. Shall I cherry pick them separately?

RainbowMango:
> https://github.com/karmada-io/karmada/pull/2483 introduces a bug, and https://github.com/karmada-io/karmada/pull/2469 fix it. Shall I cherry pick them separately?

For this case, #2469 and #2483 should be picked together and they fixed the same bug. But we can ignore the #2469's release note as the `bug` is transparent for users using v1.3.0.

XiShanYongYe-Chang:
> #2483 introduces a bug, and #2469 fix it. Shall I cherry pick them separately?

I think what you're trying to say is: #2469 introduces a bug, and #2483 fix it. :)

XiShanYongYe-Chang (commented):
Thanks
/lgtm

Can you help add a release-note for this cherry-pick?

PR #2523, opened 2022-09-17
remove service accounts for karmada components except for karmada-agent

carlory:

**What type of PR is this?**

/kind feature

See https://github.com/karmada-io/karmada/pull/2511#discussion_r972541003

Part of #2511

```release-note
`karmadactl`/`chart`: The `init` no longer creates redundant ServiceAccounts for components except `karmada-agent`.
```


RainbowMango:
Thanks @carlory , I'll look at it next week.

RainbowMango:
I just remember that there is an unusual case where they install Karmada on a cluster and take the `kube-apiserver` as the `karmada-apiserver`. In that case, does the component needs RBAC?

carlory:
@RainbowMango 

I don't think we should encourage such usage, because two controllers controlling the same resource, kube-controller-manager and karmada-controller-manager, will conflict when updating state.

carlory:
> In that case, does the component needs RBAC?

yes. but we should not support this case, right?

lonelyCZ:
> In that case, does the component needs RBAC?

This should be a custom requirement, we can add a doc to describe it. I think that the standard installation method currently doesn't need them.

RainbowMango:
> This should be a custom requirement, we can add a doc to describe it. I think that the standard installation method currently doesn't need them.

Good point.

RainbowMango (approved):
/lgtm
/approve

Note: I updated the release note a little bit.

PR #2540, opened 2022-09-19
`karmadactl`: Fixed options of `deinit` can not be shown issue.

helen-frank:
Signed-off-by: helen <helenfrank@protonmail.com>

**What type of PR is this?**
/kind feature

`deinit -h` need print options
Fixes #

A novice who has just come into contact with karmada thinks it is very important for `deinit -h` to have options.
it`s me ^-^

```release-note
`karmadactl`: Fixed options of `deinit` can not be shown issue.
```

### DIFF
--- a/pkg/karmadactl/deinit.go
+++ b/pkg/karmadactl/deinit.go
@@ -61,7 +61,7 @@ func NewCmdDeInit(parentCommand string) *cobra.Command {
 		},
 	}
 
-	flags := cmd.PersistentFlags()
+	flags := cmd.Flags()
 	flags.StringVarP(&opts.Namespace, "namespace", "n", "karmada-system", "namespace where Karmada components are installed.")
 	flags.StringVar(&opts.KubeConfig, "kubeconfig", "", "Path to the host cluster kubeconfig file.")
 	flags.StringVar(&opts.Context, "context", "", "The name of the kubeconfig context to use")


pkg/karmadactl/deinit.go
```diff
@@ -61,7 +61,7 @@ func NewCmdDeInit(parentCommand string) *cobra.Command {
 		},
 	}
 
-	flags := cmd.PersistentFlags()
```
RainbowMango, line 64: I can't tell the exact difference between `PersistentFlags` and `Flags`, can you remind me @carlory ?

lonelyCZ: The `PersistentFlags` will be passed to all `subcommands` of the current command. We have imported `karmadactl option` command, it seems to capture all `PersistentFlags`, so we can't use `PersistentFlags` in specific subcommand.

lonelyCZ:
Hi, @helen-frank. Could you please make the `Does this PR introduce a user-facing change?:` shorter and clearer(generally in a word)?

helen-frank:
> Hi, @helen-frank. Could you please make the `Does this PR introduce a user-facing change?:` shorter and clearer(generally in a word)?

`karmadactl deinit -h` print Options
Is this reasonable?

pkg/karmadactl/deinit.go
RainbowMango: Do you mean `PersistentFlags` `used` to add the flags that are intended used by `subcommands`, but the flags can't be shown in the usage of the current command?

Take here, for example, we added following `--namespace`/`--kubeconfig`/`--context` to the `deinit` as `PersistentFlags`, so these flags can not be shown.  right?

helen-frank: @RainbowMango Yes，reference [https://stackoverflow.com/questions/63495992/what-is-the-difference-between-go-cobra-persistentflags-and-flags](url)

RainbowMango: Great thanks @helen-frank. I got it.

RainbowMango (approved):
/approve

RainbowMango:
@helen-frank Would you like to check if following release branches need this patch, we probably need to [cherry-pick this patch](https://karmada.io/docs/contributor/cherry-picks).
- release-1.3
- release-1.2
- release-1.1

RainbowMango:
I just checked the code, we don't need to cherry-pick this patch to `release-1.1`, as we started to support `deinit` at `release-1.2`.

helen-frank:
OK, I'll do it later.

PR #2542, opened 2022-09-19
Add helm variables to choose controllers of karmada and kubernetes ma…

### LINKED ISSUE #2512: Namespaces should not be synced in all clusters
The cluster controller of karmada should not automatically create a new namespace in all the clusters. It should process like others resources.

**Why is this needed**:
Sometimes we want that some namspaces be created only in some clusters.  
The namespaces should be created only if there is a `ClusterPropagationPolicy` referencing it, and the target clusters should be determined from it.

fredgate:
**What type of PR is this?**
/kind feature

Allow to choose controllers executed by karmada and kubernetes manager of control plane when deploying with the Helm chart.

Fixes #2512 

```release-note
Add variables `controllerManager.config.controllers` and `kubeControllerManager.config.controllers` to choose controllers executed by karmada and kubernetes managers
```

### DIFF
--- a/charts/karmada/values.yaml
+++ b/charts/karmada/values.yaml
@@ -261,6 +261,9 @@ controllerManager:
     ##   - myRegistryKeySecretName
     ##
     pullSecrets: []
+  ## @param controllerManager.config.controllers the controllers launched by the manager.
+  config:
+    controllers: ""
   ## @param controllerManager.resources resource quota of the karmada-controller-manager
   resources: {}
     # If you do want to specify resources, uncomment the following
@@ -458,6 +461,9 @@ kubeControllerManager:
     ##   - myRegistryKeySecretName
     ##
     pullSecrets: []
+  ## @param kubeControllerManager.config.controllers the controllers launched by the kubernetes manager.
+  config:
+    controllers: namespace,garbagecollector,serviceaccount-token,ttl-after-finished,bootstrapsigner,csrapproving,csrcleaner,csrsigning
   ## @param kubeControllerManager.resources resource quota of the kube-controller-manager
   resources:
     # If you do want to specify resources, uncomment the following
--- a/charts/karmada/templates/karmada-controller-manager.yaml
+++ b/charts/karmada/templates/karmada-controller-manager.yaml
@@ -50,6 +50,9 @@ spec:
             - /bin/karmada-controller-manager
             - --kubeconfig=/etc/kubeconfig
             - --bind-address=0.0.0.0
+            {{- if .Values.controllerManager.config.controllers }}
+            - --controllers={{ .Values.controllerManager.config.controllers }}
+            {{- end }}
             - --cluster-status-update-frequency=10s
             - --secure-port=10357
             - --leader-elect-resource-namespace={{ include "karmada.namespace" . }}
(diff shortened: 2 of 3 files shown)


charts/karmada/values.yaml
```diff
@@ -261,6 +261,9 @@ controllerManager:
     ##   - myRegistryKeySecretName
     ##
     pullSecrets: []
+  ## @param controllerManager.config.controllers the controllers launched by the manager.
+  config:
```
Poor12, line 265: I prefer to use `controllers` directly. `config` looks a bit redundant.

Poor12 (commented):
Thanks @fredgate. Would you like to update the document [here](https://github.com/karmada-io/karmada/blob/master/charts/karmada/README.md#configuration) synchronously? :)

charts/karmada/values.yaml
fredgate: I can use `controllers` directly instead of `config.controllers`, but I preferred the latter because it will be more organized if some additional parameters need to be added such as `cluster-status-update-frequency` or `v`...  
All these values will be clearly under `controllerManager.config` rather than directly at the root of `controllerManager`, which I find more readable.

What do you think about it ?

Poor12: I got what you mean. I prefer to put some related configuration items under the same group. It's ok to use `controllers` directly when there are currently no other options.
BTW, I don't think `config` is a good group name. All of them in `controllerManager` belong to the category of config.
```
## controller manager config
controllerManager:
```

RainbowMango: > I can use controllers directly instead of config.controllers, but I preferred the latter because it will be more organized if some additional parameters need to be added such as cluster-status-update-frequency or v...

I agree that we may need to customize other parameters later, they are organized together would be great!

> I don't think config is a good group name. All of them in controllerManager belong to the category of config.

How about `args`?

fredgate: Yes `args` seems like a better name to me.  
Should we use it as
- an object with predefined properties (like I did in the PR)
- an array where as many elements can be added freely. We could fill this variable like this during a release
```
controllerManager:
  args:
    - --controllers=-namespace,*
    - --v=3
```

The latter makes it possible to better cover all the possibilities without modification.

Poor12: I think the former is better. The latter may confuse users about which configuration items can be configured.

fredgate: The latter option allows to configure all args of `karmada-controller-manager`, that allows the user to configure anything he wants, not just what we imagine : there can always be unimagined scenarios, like for example mine. And  both current and future. And the chart will be compatible without change not only with current arguments but also with future ones.
 @RainbowMango What do you think about this two solutions ?

RainbowMango: > The latter option allows to configure all args of karmada-controller-manager, that allows the user to configure anything he wants, not just what we imagine

The idea is great, but my concern is the user experience. The users might need to specify all args at [here](https://github.com/karmada-io/karmada/blob/00b574f91c6ae3b030f90788e4f30f417bdabd21/charts/karmada/templates/karmada-controller-manager.yaml#L51-L56), even they just want to customize the `--controllers`.

RainbowMango: Kindly ping @fredgate .

Poor12:
Hi, @fredgate, I guess you are busy now. I put the remaining work of this pr in #3240. Looking forward to your reply.

PR #2549, opened 2022-09-20
replace colon with point in the rbac resource name

### LINKED ISSUE #2392: resource with specific name not sync to member cluster
**What happened**:

I'm trying to install [OLM](https://olm.operatorframework.io/) with karmada, but got following error message:
```
time="2022-08-19T00:11:31Z" level=fatal msg="couldn't clean previous release" error="catalogsources.operators.coreos.com \"olm-operators\" is forbidden: User \"system:serviceaccount:olm:olm-operator-serviceaccount\" cannot delete resource \"catalogsources\" in API group \"operators.coreos.com\" in the namespace \"olm\": RBAC: clusterrole.rbac.authorization.k8s.io \"system:controller:operator-lifecycle-manager\" not found"
```

seems `system:controller:operator-lifecycle-manager` not sync to member cluster?

cc @XiShanYongYe-Chang 

**Environment**:
- Karmada version:
- Others:

XiShanYongYe-Chang:

**What type of PR is this?**

/kind bug
/kind feature

As talked about in #2392, when we propagate RBAC resources, whose names contain colon characters, it will report err.

The reason is that colons are allowed in names in validations for these resources, but this is different from name validations for CRD resources:

https://github.com/kubernetes/kubernetes/blob/f0823c0f59d6ea8e2ad0dc4e3fe0f2b396c9185f/pkg/apis/rbac/validation/validation.go#L124

https://github.com/kubernetes/kubernetes/blob/f0823c0f59d6ea8e2ad0dc4e3fe0f2b396c9185f/staging/src/k8s.io/apiextensions-apiserver/pkg/registry/customresource/validator.go#L60

Fixes #2392 

Replace colon with point in the RBAC resource name.

```release-note
karmada-controller-manager: fix the error of resources whose name contains colons failing to be created.
```

### DIFF
--- a/pkg/util/names/names.go
+++ b/pkg/util/names/names.go
@@ -55,6 +55,17 @@ func GetClusterName(executionSpaceName string) (string, error) {
 
 // GenerateBindingName will generate binding name by kind and name
 func GenerateBindingName(kind, name string) string {
+	// The name of resources, like 'Role'/'ClusterRole'/'RoleBinding'/'ClusterRoleBinding',
+	// may contain symbols(like ':') that are not allowed by CRD resources which require the
+	// name can be used as a DNS subdomain name. So, we need to replace it.
+	// These resources may also allow for other characters(like '&','$') that are not allowed
+	// by CRD resources, we only handle the most common ones now for performance concerns.
+	// For more information about the DNS subdomain name, please refer to
+	// https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names.
+	if strings.Contains(name, ":") {
+		name = strings.ReplaceAll(name, ":", ".")
+	}
+
 	return strings.ToLower(name + "-" + kind)
 }
 
@@ -73,6 +84,17 @@ func GenerateBindingReferenceKey(namespace, name string) string {
 
 // GenerateWorkName will generate work name by its name and the hash of its namespace, kind and name.
 func GenerateWorkName(kind, name, namespace string) string {
+	// The name of resources, like 'Role'/'ClusterRole'/'RoleBinding'/'ClusterRoleBinding',
+	// may contain symbols(like ':') that are not allowed by CRD resources which require the
+	// name can be used as a DNS subdomain name. So, we need to replace it.
+	// These resources may also allow for other characters(like '&','$') that are not allowed
+	// by CRD resources, we only handle the most common ones now for performance concerns.
+	// For more information about the DNS subdomain name, please refer to
+	// https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names.
+	if strings.Contains(name, ":") {
+		name = strings.ReplaceAll(name, ":", ".")
+	}
+
 	var workName string
 	if len(namespace) == 0 {
 		workName = strings.ToLower(name + "-" + kind)
--- a/pkg/util/names/names_test.go
+++ b/pkg/util/names/names_test.go
@@ -140,6 +140,31 @@ func TestGenerateBindingName(t *testing.T) {
 			name:     "Pod",
 			expect:   "pod-pods",
 		},
+		// RBAC kind resource test with colon character in the name
+		{
+			testCase: "role kind resource",
+			kind:     "Role",
+			name:     "system:controller:tom",
+			expect:   "system.controller.tom-role",
+		},
+		{
+			testCase: "clusterRole kind resource",
+			kind:     "ClusterRole",
+			name:     "system::tom",
+			expect:   "system..tom-clusterrole",
+		},
+		{
+			testCase: "roleBinding kind resource",
+			kind:     "RoleBinding",
+			name:     "system::tt:tom",
+			expect:   "system..tt.tom-rolebinding",
+		},
+		{
+			testCase: "clusterRoleBinding kind resource",
+			kind:     "ClusterRoleBinding",
+			name:     "system:tt:tom",
+			expect:   "system.tt.tom-clusterrolebinding",
+		},
 	}
 	for _, test := range tests {
 		if result := GenerateBindingName(test.kind, test.name); result != test.expect {
--- a/test/e2e/clusterpropagationpolicy_test.go
+++ b/test/e2e/clusterpropagationpolicy_test.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	"github.com/onsi/ginkgo/v2"
+	rbacv1 "k8s.io/api/rbac/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/util/rand"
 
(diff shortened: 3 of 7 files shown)


pkg/util/names/names.go
```diff
@@ -134,3 +137,12 @@ func GeneratePolicyName(namespace, name, gvk string) string {
 
 	return strings.ToLower(fmt.Sprintf("%s-%s", name, rand.SafeEncodeString(fmt.Sprint(hash.Sum32()))))
 }
+
+var rbacKindStrings = sets.NewString("Role", "ClusterRole", "RoleBinding", "ClusterRoleBinding")
+
+func replaceColonWithPointInNameIfNeeded(kind, name string) string {
```
jwcesign, line 143: add some comment like other functions?

XiShanYongYe-Chang: ok, I will update it, thanks~.

zirain (commented):
lgtm

pkg/util/names/names.go
```diff
@@ -134,3 +137,14 @@ func GeneratePolicyName(namespace, name, gvk string) string {
 
 	return strings.ToLower(fmt.Sprintf("%s-%s", name, rand.SafeEncodeString(fmt.Sprint(hash.Sum32()))))
 }
+
+var rbacKindStrings = sets.NewString("Role", "ClusterRole", "RoleBinding", "ClusterRoleBinding")
```
RainbowMango, line 141: Can you explain why we need the preset list?
It looks weird to me because for those resource types that don't support `:` in name, they never get a chance to be applied in karmada-apiserver, and for resource types that support `:` in the name(other than RBAC resources, if there is any), why don't we support them?

pkg/util/names/names.go
```diff
@@ -55,6 +56,7 @@ func GetClusterName(executionSpaceName string) (string, error) {
 
 // GenerateBindingName will generate binding name by kind and name
 func GenerateBindingName(kind, name string) string {
+	name = replaceColonWithPointInNameIfNeeded(kind, name)
```
RainbowMango, line 59: We should comment on the reason.

pkg/util/names/names.go
XiShanYongYe-Chang: It seems superfluous, I remove this list.

pkg/util/names/names.go
XiShanYongYe-Chang: Updated.

pkg/util/names/names.go
```diff
@@ -55,6 +55,13 @@ func GetClusterName(executionSpaceName string) (string, error) {
 
 // GenerateBindingName will generate binding name by kind and name
 func GenerateBindingName(kind, name string) string {
+	// For some resources (resources currently known are Role, ClusterRole, RoleBinding,
+	// ClusterRoleBinding), their names may contain symbols other than "." and "-", which is
+	// not allowed during the creation of CustomResource, so we need to replace them.
+	// After investigation, ":" is used in these rbac resources, such as "system:node", so we
+	// replace it first and then process other symbols if they are found later.
```
RainbowMango, line 62:
```suggestion
	// The name of resources, like 'Role'/'ClusterRole'/'RoleBinding'/'ClusterRoleBinding',
	// may contain symbols(like ':') that are not allowed by CRD resources which require the
	// name can be used as a DNS subdomain name. So, we need to replace it.
	// These resources may also allow for other characters(like '&','$') that are not allowed
	// by CRD resources, we only handle the most common ones now for performance concerns.
	// For more information about the DNS subdomain name, please refer to
	// https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names.
```

XiShanYongYe-Chang:
Test with benchmark test with the following code:
```go
func GetName01(name string) string {
	return strings.ReplaceAll(name, ":", ".")
}

func GetName02(name string) string {
	if strings.Contains(name, ":") {
		return strings.ReplaceAll(name, ":", ".")
	}
	return name
}

func BenchmarkGetName01(b *testing.B) {
	names := []string{"aaa-111", "bbb-222", "ccc-333", "ddd-444", "eee-555", "ttt:000", "ooo.666", "ppp.777", "qqq.888", "mmm.999"}

	for i := 0; i < b.N; i++ {
		for _, name := range names {
			GetName01(name)
		}
	}
}

func BenchmarkGetName02(b *testing.B) {
	names := []string{"aaa-111", "bbb-222", "ccc-333", "ddd-444", "eee-555", "ttt:000", "ooo.666", "ppp.777", "qqq.888", "mmm.999"}

	for i := 0; i < b.N; i++ {
		for _, name := range names {
			GetName02(name)
		}
	}
}
```

Test result:
```
goos: windows
goarch: amd64
pkg: github.com/karmada-io/karmada/pkg/util/names
cpu: Intel(R) Core(TM) i7-8700 CPU @ 3.20GHz
BenchmarkGetName01
BenchmarkGetName01-12    	 6945460	       174.8 ns/op
PASS

goos: windows
goarch: amd64
pkg: github.com/karmada-io/karmada/pkg/util/names
cpu: Intel(R) Core(TM) i7-8700 CPU @ 3.20GHz
BenchmarkGetName02
BenchmarkGetName02-12    	 9404542	       113.5 ns/op
PASS
```

XiShanYongYe-Chang:
Test 10 times:
```
$ go test -bench="." -count 10
goos: windows
goarch: amd64
pkg: github.com/karmada-io/karmada/pkg/util/names
cpu: Intel(R) Core(TM) i7-8700 CPU @ 3.20GHz
BenchmarkGetName01-12            7352548               155.4 ns/op
BenchmarkGetName01-12            7366518               158.6 ns/op
BenchmarkGetName01-12            7736853               157.9 ns/op
BenchmarkGetName01-12            8215824               161.6 ns/op
BenchmarkGetName01-12            8245869               161.6 ns/op
BenchmarkGetName01-12            7543009               156.5 ns/op
BenchmarkGetName01-12            8124388               158.6 ns/op
BenchmarkGetName01-12            8274423               159.8 ns/op
BenchmarkGetName01-12            7917421               158.1 ns/op
BenchmarkGetName01-12            8086982               161.0 ns/op
BenchmarkGetName02-12           10591256               108.7 ns/op
BenchmarkGetName02-12           11280982               108.9 ns/op
BenchmarkGetName02-12           11021712               109.8 ns/op
BenchmarkGetName02-12           11333691               107.8 ns/op
BenchmarkGetName02-12           10991103               110.0 ns/op
BenchmarkGetName02-12           10892484               104.0 ns/op
BenchmarkGetName02-12           10644733               108.0 ns/op
BenchmarkGetName02-12           11716118               105.5 ns/op
BenchmarkGetName02-12           11356237               109.9 ns/op
BenchmarkGetName02-12           11076076               114.6 ns/op
PASS
ok      github.com/karmada-io/karmada/pkg/util/names    27.423s
```

pkg/util/names/names.go
XiShanYongYe-Chang: Update.

RainbowMango:
LGTM.
@Garrybest would you like to take a look?

Garrybest:
LGTM, just one question. What if a RBAC have both colon and point?

RainbowMango:
> What if a RBAC have both colon and point?

Good question. If I remember correctly, we don't use the name of `RB` or `Work` name to find the `resource template`, so  there is no problem. Please @XiShanYongYe-Chang confirm.  A test for this would be awsome.

XiShanYongYe-Chang:
> LGTM, just one question. What if a RBAC have both colon and point?

Just as @RainbowMango, this updation only changes the rb and work names of related resources, and does not change the names propagated to member clusters.

> A test for this would be awsome.

An E2E test？

XiShanYongYe-Chang:
I update the resource name as follows in E2E:
```
Role: system:test:%s
RoleBinding: system:test.%s
ClusterRole: system:test-%s
ClusterRoleBinding: system:test.tt:%s
```

Garrybest:
> We don't use the name of RB or Work name to find the resource template.

I don't remember exactly. Do we use labels or annotations to target the corresponding RB and original resource? If so, there is no problem here.

XiShanYongYe-Chang:
The rb finds the resource template based on the `.spec.resource` field. The label is used between the rb and the work.

RainbowMango (approved):
/approve

PR #2552, opened 2022-09-20
[E2E] replace gomeme.Eventually with wait when logic has assertions

XiShanYongYe-Chang:

**What type of PR is this?**

/kind failing-test
/kind cleanup

Using assertions in `wait.PollImmediate` breaks the cyclic waiting process, resulting in failure.

Meanwhile, as the error message suggests:

```
      When you, or your assertion library, calls Ginkgo's Fail(),
      Ginkgo panics to prevent subsequent assertions from running.

      Normally Ginkgo rescues this panic so you shouldn't see it.

      However, if you make an assertion in a goroutine, Ginkgo can't capture the
      panic.
      To circumvent this, you should call

      	defer GinkgoRecover()

      at the top of the goroutine that caused this panic.

      Alternatively, you may have made an assertion outside of a Ginkgo
      leaf node (e.g. in a container node or some out-of-band function) - please
      move your assertion to
      an appropriate Ginkgo node (e.g. a BeforeSuite, BeforeEach, It, etc...).
```

Error info has printed the panic, as a result, we can see the normal error info, so I use `gomega.Eventually` to replace `wait.PollImmediate`.

Fixes #

https://github.com/karmada-io/karmada/actions/runs/3088395768/jobs/4994950606

```release-note
NONE
```

### DIFF
--- a/test/e2e/failover_test.go
+++ b/test/e2e/failover_test.go
@@ -104,20 +104,17 @@ var _ = framework.SerialDescribe("failover testing", func() {
 			})
 
 			ginkgo.By("check whether deployment of failed cluster is rescheduled to other available cluster", func() {
-				err := wait.PollImmediate(pollInterval, pollTimeout, func() (bool, error) {
+				gomega.Eventually(func() int {
 					targetClusterNames = framework.ExtractTargetClustersFrom(controlPlaneClient, deployment)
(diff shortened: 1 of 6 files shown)


RainbowMango (approved):
/lgtm
/approve

PR #2555, opened 2022-09-21
Fix the issue that incomingBindingMetrics ignores the ClusterChanged event

Poor12:

**What type of PR is this?**
/kind cleanup

Add metrics test for scheduler.

Fixes #

None

### DIFF
--- a/pkg/scheduler/event_handler.go
+++ b/pkg/scheduler/event_handler.go
@@ -12,6 +12,7 @@ import (
 
 	clusterv1alpha1 "github.com/karmada-io/karmada/pkg/apis/cluster/v1alpha1"
 	policyv1alpha1 "github.com/karmada-io/karmada/pkg/apis/policy/v1alpha1"
+	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
 	"github.com/karmada-io/karmada/pkg/scheduler/metrics"
 	"github.com/karmada-io/karmada/pkg/util"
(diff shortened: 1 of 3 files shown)


pkg/scheduler/event_handler.go
```diff
@@ -131,14 +154,7 @@ func (s *Scheduler) requeueClusterResourceBindings(selector labels.Selector) err
...
 
 	for _, clusterResourceBinding := range referenceClusterResourceBindings {
-		key, err := cache.MetaNamespaceKeyFunc(clusterResourceBinding)
-		if err != nil {
-			klog.Errorf("couldn't get key for ClusterResourceBinding(%s): %v", clusterResourceBinding.Name, err)
-			continue
-		}
-		klog.Infof("Requeue ClusterResourceBinding(%s) as placement changed.", clusterResourceBinding.Name)
-		s.queue.Add(key)
-		metrics.CountSchedulerBindings(metrics.PolicyChanged)
```
RainbowMango, line 141: The `requeueClusterResourceBindings` is called when:
1. Policy changed  `onClusterPropagationPolicyUpdate`
2. Cluster changed `enqueueAffectedClusterPolicy`

So, the metric is inaccurate, we should fix it before adding a unit test for it because the code would change accordingly.

Poor12: Done.

RainbowMango (approved):
/lgtm
/approve

PR #2562, opened 2022-09-22
allow to update the resourceSelector of PropagationPolicy/ClusterPropagationPolicy

XiShanYongYe-Chang:

**What type of PR is this?**

/kind feature

Allow users to update the `.spec.resourceSelectors` filed of `PropagationPolicy/ClusterPropagationPolicy`.

Fixes #

```release-note
karmada-controller-manager: allow users to update the `.spec.resourceSelectors` filed of `PropagationPolicy/ClusterPropagationPolicy`.
```

### DIFF
--- a/pkg/detector/detector.go
+++ b/pkg/detector/detector.go
@@ -734,7 +734,13 @@ func (d *ResourceDetector) OnPropagationPolicyAdd(obj interface{}) {
 
 // OnPropagationPolicyUpdate handles object update event and push the object to queue.
 func (d *ResourceDetector) OnPropagationPolicyUpdate(oldObj, newObj interface{}) {
-	// currently do nothing, since a policy's resource selector can not be updated.
+	key, err := ClusterWideKeyFunc(newObj)
+	if err != nil {
+		return
+	}
+
+	klog.V(2).Infof("Update PropagationPolicy(%s) resourceSelectors", key)
+	d.policyReconcileWorker.Add(key)
 }
 
 // OnPropagationPolicyDelete handles object delete event and push the object to queue.
@@ -776,7 +782,7 @@ func (d *ResourceDetector) ReconcilePropagationPolicy(key util.QueueKey) error {
 		klog.Errorf("Failed to convert PropagationPolicy(%s) from unstructured object: %v", ckey.NamespaceKey(), err)
 		return err
 	}
-	return d.HandlePropagationPolicyCreation(propagationObject)
+	return d.HandlePropagationPolicyCreationOrUpdate(propagationObject)
 }
 
 // OnClusterPropagationPolicyAdd handles object add event and push the object to queue.
@@ -792,7 +798,13 @@ func (d *ResourceDetector) OnClusterPropagationPolicyAdd(obj interface{}) {
 
 // OnClusterPropagationPolicyUpdate handles object update event and push the object to queue.
 func (d *ResourceDetector) OnClusterPropagationPolicyUpdate(oldObj, newObj interface{}) {
-	// currently do nothing, since a policy's resource selector can not be updated.
+	key, err := ClusterWideKeyFunc(newObj)
+	if err != nil {
+		return
+	}
+
+	klog.V(2).Infof("Update ClusterPropagationPolicy(%s) resourceSelectors", key)
+	d.clusterPolicyReconcileWorker.Add(key)
 }
 
 // OnClusterPropagationPolicyDelete handles object delete event and push the object to queue.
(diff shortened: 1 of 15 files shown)


pkg/detector/detector.go
```diff
@@ -794,7 +794,17 @@ func (d *ResourceDetector) OnPropagationPolicyAdd(obj interface{}) {
 
 // OnPropagationPolicyUpdate handles object update event and push the object to queue.
 func (d *ResourceDetector) OnPropagationPolicyUpdate(oldObj, newObj interface{}) {
-	// currently do nothing, since a policy's resource selector can not be updated.
+	if resourceSelectorsDeepEqual(oldObj, newObj) {
+		return
+	}
```
RainbowMango, line 799: Since the `detector` only focuses on the `resource` match thing instead of `cluster` match. 
That gives an illusion that here we just care about the `ResourceSelector` would be enough, but we still have other fields that need to be updated when building ResourceBinding, like ([PropagateDeps](https://github.com/karmada-io/karmada/blob/07ddc07997224c1ea4df21ce65b0ac8b2f0e22b5/pkg/detector/detector.go#L723)), so my opinion is to remove this check, just put all PP into the queue.

RainbowMango (commented):
cc @Garrybest for review, i remember you asked why `resourceSelector` is not allowed to update.

pkg/detector/policy.go
```diff
@@ -0,0 +1,36 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/klog/v2"
+
+	"github.com/karmada-io/karmada/pkg/util/helper"
+)
+
+func resourceSelectorsDeepEqual(oldObj, newObj interface{}) bool {
+	oldResourceSelectors, err := getPolicyResourceSelectors(oldObj)
+	if err != nil {
+		klog.Errorf("Failed to get oldObj resourceSelectors, error: %v", err)
```
Garrybest, line 15: Log the obj ns and name here would be better to diagnose.

pkg/detector/detector.go
```diff
@@ -794,7 +794,17 @@ func (d *ResourceDetector) OnPropagationPolicyAdd(obj interface{}) {
...
+
+	key, err := ClusterWideKeyFunc(newObj)
+	if err != nil {
+		return
+	}
+
+	klog.V(2).Infof("Update PropagationPolicy(%s) resourceSelectors", key)
+	d.policyReconcileWorker.Add(key)
```
Garrybest, line 807: It seems like we only add this pp into reconcile queue. However, the object which is selected by this pp but not matched yet after updating will not be reconciled right? `ReconcilePropagationPolicy` only adds all matching objects into `Processor`. I guess the old object should also be handled.

RainbowMango: > However, the object which is selected by this pp but not matched yet after updating will not be reconciled right? 

No. If the update adds a new `ResourceSelector` which aims to match a new resource, and if the resource exists in `waiting list` it will be [put into reconciling queue](https://github.com/karmada-io/karmada/blob/c0349fe42931e9de3eb18671b6c028750660a0be/pkg/detector/detector.go#L779-L793).

> ReconcilePropagationPolicy only adds all matching objects into Processor. I guess the old object should also be handled.

Why do you think the old object should be handled? I guess the `handled` means `re-concile`. 
This reminds me that if an old object is matched by `previous PP`, but after the update, it doesn't match the `new PP`, it equals to remove the PP, we should clear the labels from the old object, just like https://github.com/karmada-io/karmada/blob/c0349fe42931e9de3eb18671b6c028750660a0be/pkg/detector/detector.go#L915-L949

XiShanYongYe-Chang: When there is an object which is selected by this pp but not matched yet after updating, I think we can ignore it. The life cycle of the rb is consistent with the life cycle of the resource,  and it will be deleted as the resource is deleted.

Therefore, in general, when modifying the `ResourceSelector` of the pp, we encourage the users to add some items rather than delete the old items at will. In this way, although some flexibility is lost, the behavior is more secure. In my opinion, the action of deleting the old item can be replaced by deleting the entire pp.

pkg/detector/detector.go
XiShanYongYe-Chang: Update.

pkg/detector/detector.go
Garrybest: It makes sense, I agree with you.

XiShanYongYe-Chang: >  I think the label in RBs will not be removed, which is not expected.

Hi @Garrybest , as you said, we need to remove the labels in the old RBs, otherwise, changes to the PP placement will still take effect on the old PRs.

XiShanYongYe-Chang:
/hold
We need to clean the labels on the old RBs and resource templates.

RainbowMango:
Please assign it to me once it ready for review.

pkg/detector/detector.go
```diff
@@ -934,10 +946,17 @@ func (d *ResourceDetector) HandleClusterPropagationPolicyDeletion(policyName str
 	return errors.NewAggregate(errs)
 }
 
-// HandlePropagationPolicyCreation handles PropagationPolicy add event.
-// When a new policy arrives, should check if object in waiting list matches the policy, if yes remove the object
+// HandlePropagationPolicyCreationOrUpdate handles PropagationPolicy add event.
```
RainbowMango, line 949: not just `add` events anymore.

RainbowMango (approved):
/lgtm
/approve

PR #2565, opened 2022-09-22
Add a new env var `DOCKER_BUILD_ARGS`

loicalbertin:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

This env var is used to provide additional arguments to the docker build command.

This typically allow to set HTTP proxy settings in the container only at build time. This was the original need described in #2553

**Which issue(s) this PR fixes**:

Fixes #2553

**Special notes for your reviewer**:

An alternative specific to HTTP proxies is proposed in #2568

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #2568, opened 2022-09-22
Automatically handle `http(s)_proxy` env vars for docker builds

### LINKED ISSUE #2553: run hack/local-up-karmada.sh behind an HTTP proxy
:wave: Hi all!

I'm evaluating Karmada in an environment that access to the internet through an HTTP proxy.

When running `hack/local-up-karmada.sh` I'm facing an error when building docker images. Basically `apk add` fails when trying to download packages.

To workaround this I added a `DOCKER_BUILD_ARGS` variable to [`hack/docker.sh`](https://github.com/karmada-io/karmada/blob/ab889889faa84b5a1c1ad89241862b642b3cfb4c/hack/docker.sh#L57)

```bash
  docker build --build-arg BINARY="${target}" \
          ${DOCKER_BUILD_ARGS} \                # <<<< ADDED THIS LINE
          --tag "${image_name}" \
          --file "${REPO_ROOT}/cluster/images/Dockerfile" \
          "${REPO_ROOT}/_output/bin/${platform}"
```

And I now run the `hack/local-up-karmada.sh` this way:

```bash
DOCKER_BUILD_ARGS="--build-arg https_proxy=${https_proxy}" hack/local-up-karmada.sh
```

Is there a better way to work with proxies?

Thanks & have a nice day!
Loïc

loicalbertin:
**What type of PR is this?**

/kind feature

This PR allows to build images in a environment that only allow access to internet through proxies

Fixes #2553

An alternative allowing to provide more arguments to the docker build is proposed in #2565

```release-note
NONE
```

### DIFF
--- a/hack/docker.sh
+++ b/hack/docker.sh
@@ -30,6 +30,10 @@ source "${REPO_ROOT}/hack/util.sh"
 REGISTRY=${REGISTRY:-"swr.ap-southeast-1.myhuaweicloud.com/karmada"}
 VERSION=${VERSION:="unknown"}
 
+export http_proxy="${HTTP_PROXY:-${http_proxy:-}}"
+export https_proxy="${HTTPS_PROXY:-${https_proxy:-}}"
+export no_proxy="${NO_PROXY:-${no_proxy:-}}"
+
 function build_images() {
   local -r target=$1
   local -r output_type=${OUTPUT_TYPE:-docker}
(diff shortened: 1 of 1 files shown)


hack/docker.sh
```diff
@@ -55,6 +59,7 @@ function build_local_image() {
   echo "Building image for ${platform}: ${image_name}"
   set -x
   docker build --build-arg BINARY="${target}" \
+          --build-arg http_proxy --build-arg https_proxy --build-arg no_proxy \
```
jwcesign, line 62: Did u try build without setting `http_proxy/https_proxyno_proxy`?

loicalbertin: It always depends on the implementation of the underlying program but usually the consensus is to not take this into account when the variable is defined but empty.

ikaven1024: shall we add `HTTP_PROXY`? Some programs only recognize the upper case variables.

loicalbertin: #2565 was preferred to this solution.
Are you ok to close this PR?

RainbowMango: /close

Issue #2349, opened 2022-08-09
[Umbrella] karmadactl uses factory instead of karmadaConfig


carlory:
**What would you like to be added**:

Iterated task:

- [x] karmadactl logs uses factory #2202
- [x] karmadactl apply uses factory
- [x] karmadactl describe uses factory #2471
- [x] karmadactl exec uses factory #2470
- [ ] karmadactl get uses factory 
- [ ] karmadactl promote uses factory
- [ ] karmadactl taint uses factory
- [ ] all sub command accepts factory as input parameter

**Why is this needed**:

At present, there are many places in karmadactl, which are building clients and factories for member cluster repeatedly, such as init, get, log, exec, etc. 

#2202  introduces factory interface, and try to resolve it.

/assign

liangyuanpeng:
I would working for `karmadactl promote`.

lonelyCZ:
> I would working for karmadactl promote.

Hi @liangyuanpeng , have it moved forward? 

And I consider to optimize the code style of karmadactl at #2664 , let's push ahead this issue.

carlory:
/close
All tasks has been done

PR #2570, opened 2022-09-22
karmadactl apply uses factory to access member cluster

helen-frank:
**What type of PR is this?**
/kind feature

Fixes #
Part of [#2349](https://github.com/karmada-io/karmada/issues/2349)
Test
```shell
go run ./cmd/karmadactl/karmadactl.go --kubeconfig /etc/karmada/karmada-apiserver.config apply -f ./_tmp/nginx.yaml
namespace/test1 created
deployment.apps/nginx1 created
```

```release-note
NONE
```

### DIFF
--- a/pkg/karmadactl/apply.go
+++ b/pkg/karmadactl/apply.go
@@ -12,14 +12,12 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/cli-runtime/pkg/resource"
-	restclient "k8s.io/client-go/rest"
 	kubectlapply "k8s.io/kubectl/pkg/cmd/apply"
-	cmdutil "k8s.io/kubectl/pkg/cmd/util"
 	"k8s.io/kubectl/pkg/util/templates"
 
 	policyv1alpha1 "github.com/karmada-io/karmada/pkg/apis/policy/v1alpha1"
 	karmadaclientset "github.com/karmada-io/karmada/pkg/generated/clientset/versioned"
-	"github.com/karmada-io/karmada/pkg/karmadactl/options"
+	"github.com/karmada-io/karmada/pkg/karmadactl/util"
 	"github.com/karmada-io/karmada/pkg/util/names"
 )
 
@@ -27,11 +25,8 @@ var metadataAccessor = meta.NewAccessor()
 
 // CommandApplyOptions contains the input to the apply command.
 type CommandApplyOptions struct {
-	// global flags
-	options.GlobalCommandOptions
 	// apply flags
 	KubectlApplyFlags *kubectlapply.ApplyFlags
-	Namespace         string
 	AllClusters       bool
 	Clusters          []string
 
@@ -48,8 +43,8 @@ var (
 		JSON and YAML formats are accepted.
 
 		Alpha Disclaimer: the --prune functionality is not yet complete. Do not use unless you are aware of what the current state is. See https://issues.k8s.io/34274.
-		
-		Note: It implements the function of 'kubectl apply' by default. 
+
+		Note: It implements the function of 'kubectl apply' by default.
 		If you want to propagate them into member clusters, please use %[1]s apply --all-clusters'.`)
 
 	applyExample = templates.Examples(`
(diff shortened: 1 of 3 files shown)


RainbowMango:
@XiShanYongYe-Chang Seems this change has nothing to do with E2E.

XiShanYongYe-Chang:
> @XiShanYongYe-Chang Seems this change has nothing to do with E2E.

Let me take a look.

carlory:
I'll review ASAP.

lonelyCZ:
Hi, @helen-frank , It seems a problem.

```
[root@master67 karmada]# ./karmadactl apply -f deployment.yaml --kubeconfig /etc/karmada/karmada-apiserver.config -C member68 -v=4
deployment.apps/nginx created
error: no kind "PropagationPolicy" is registered for version "policy.karmada.io/v1alpha1" in scheme "k8s.io/kubectl/pkg/scheme/scheme.go:28"
```

But the pp has been created.

```
[root@master67 karmada]# kapi get pp
NAME               AGE
nginx-6d7f8d5f5b   3m45s
```

helen-frank:
@lonelyCZ fixed

lonelyCZ:
Hi, @helen-frank , have you tested it in your env? It seems still the problem in my env.

helen-frank:
> Hi, @helen-frank , have you tested it in your env? It seems still the problem in my env.

Yes, I have tested it with pp.yaml below samples, but I am trying to solve this problem by modifying Scheme, please wait a moment

helen-frank:
@lonelyCZ I added the karmada scheme to `k8s.io/kubectl/pkg/scheme`, `karmadactl apply` the create capability of apply is complete, but patch will have problems `Error from server (UnsupportedMediaType): the body of the request was in an unknown format - accepted media types include: application/json-patch+json, application/merge-patch+json, application/apply-patch+yaml`

`vendor/k8s.io/kubectl/pkg/cmd/apply/patcher.go`
```go
// Create the versioned struct from the type defined in the restmapping
	// (which is the API version we'll be submitting the patch to)
	versionedObject, err := scheme.Scheme.New(p.Mapping.GroupVersionKind)
	switch {
	case runtime.IsNotRegisteredError(err):
		// fall back to generic JSON merge patch
		patchType = types.MergePatchType
		...
	case err != nil:
		return nil, nil, cmdutil.AddSourceToErr(fmt.Sprintf("getting instance of versioned object for %v:", p.Mapping.GroupVersionKind), source, err)
	case err == nil:
		// Compute a three way strategic merge patch to send to server.
		patchType = types.StrategicMergePatchType
                ...
	}
```

When the karmada scheme is not added to' k8s.io/kubectl/pkg/scheme', use
```go
// Complete completes all the required options
func (o *CommandApplyOptions) Complete(f util.Factory, cmd *cobra.Command, parentCommand string, args []string) error {
	karmadaClient, err := f.KarmadaClientSet()
	if err != nil {
		return err
	}
	o.karmadaClient = karmadaClient

	restConfig, err := f.ToRESTConfig()
	if err != nil {
		return err
	}
	defaultConfigFlags.WrapConfigFn = func(config *rest.Config) *rest.Config { return restConfig }

	o.KubectlApplyFlags.Factory = cmdutil.NewFactory(defaultConfigFlags)
        ...
}
```
It can solve the problem, but I don't quite understand this logic in the code.

helen-frank:
I tried to solve this problem by FactoryForKarmada()

lonelyCZ:
> And I think these codes are unreasonable, can you give me some tips about the solution here?

I guess the reason could be that the `Factory` is not  instantiated.

lonelyCZ:
I just tested it that worked also fine in my env. 

Thanks for your hard work! @helen-frank 

/lgtm

/cc @carlory

carlory:
sorry for too late reply, I will review it tomorrow.

carlory:
@helen-frank 

we don't need define the `FactoryForKarmada` func because the factoryImpl has implemented the cmdutil.Factory interface. 

The root cause is https://github.com/karmada-io/karmada/pull/2570/files#diff-844a05f6fc2de604697f89ecba199616552950457c69505fa6b97f87e3a28942L182

we should use UnstructuredClientForMapping instead of ClientForMapping.

helen-frank:
> @helen-frank
> 
> we don't need define the `FactoryForKarmada` func because the factoryImpl has implemented the cmdutil.Factory interface.
> 
> The root cause is
> 
> https://github.com/karmada-io/karmada/blob/d9e3bc860629202aa3c8629f743080d61afc2297/pkg/karmadactl/apply.go#L177
> 
> we should use UnstructuredClientForMapping instead of ClientForMapping.

OK, let me try

helen-frank:
I have just tested it and can work normally. Thank you @carlory

helen-frank:
Look at my other pr if you have time.

carlory:
> Look at my other pr if you have time.

I will review it this week.

lonelyCZ:
Thanks!

/approve

PR #2574, opened 2022-09-23
Support Deploy karmada-scheduler-estimator in Physical Machine

### LINKED ISSUE #2487: Support Deploy karmada-scheduler-estimator in Physical Machine
# What Would You Like to Be Added

In the code of karmada-scheduler establish connection to karmada-scheduler-estimator, add the logic of handling ExternalName type Service. This will allow the connection establishment to work without kube-proxy.

# Why is This Needed

There should be two different ways to deploy Karmada:

1. Deploy in Physical Machine or Virtual Machine directly.
2. Deploy in Host Kubernetes Cluster.

Currently, the connection establishment rely on kube-proxy to do the network traffic reforwarding. But in karmada-scheduler deployed directly in physical machine without host kubernetes cluster, there is no kube-proxy.

# Current Logic

karmada-scheduler needs host and port information to connect to karmada-scheduler-estimator.

Host: It's generated from cluster name. For example, if the cluster name is "ray-karmada-member1", the host will be "karmada-scheduler-estimator-ray-karmada-member1".

Port: "--scheduler-estimator-port" parameter of "karmada-scheduler"

Relevant code: <https://github.com/karmada-io/karmada/blob/665af6916a733da4cae2ce1dfb9321c264f114ff/pkg/estimator/client/cache.go#L82>

[hack/deploy-scheduler-estimator.sh](https://github.com/karmada-io/karmada/blob/665af6916a733da4cae2ce1dfb9321c264f114ff/hack/deploy-scheduler-estimator.sh) will deploy [a Service named "karmada-scheduler-estimator-{{member_cluster_name}}"](https://github.com/karmada-io/karmada/blob/665af6916a733da4cae2ce1dfb9321c264f114ff/artifacts/deploy/karmada-scheduler-estimator.yaml#L47). So if Karmada is deployed inside host kubernetes cluster, the kube-proxy will implement a form of virtual IP for the Service.

```yaml
apiVersion: v1
kind: Service
metadata:
  name: karmada-scheduler-estimator-{{member_cluster_name}}
  namespace: karmada-system
  labels:
    cluster: {{member_cluster_name}}
spec:
  selector:
    app: karmada-scheduler-estimator-{{member_cluster_name}}
  ports:
    - protocol: TCP
      port: 10352
      targetPort: 10352
  type: ClusterIP
```

# How to Do It

In physical machine deployment, we don't have kube-proxy. We use another kind of Virual IP to forward traffic, for example Linux Virtual Server.

But LVS will not install iptables rules in the machine deploying karmada-scheduler. So we need karmada-scheduler itself to interpret "Service" content to connect to the required Virtual IP and Port provided by LVS server.

There are existing logic used by Kubernetes API Server to connect to Web Hook: <https://github.com/kubernetes/kubernetes/blob/8287e21228e9b1c945d031581e2fc63920f46d56/staging/src/k8s.io/apiserver/pkg/util/proxy/proxy.go#L92>

We will use similar but different logic to support Physical Machine Deployment.

# Design Details

The Service in Physical Machine Deployment will be ExternalName type like the following.

```yaml
apiVersion: v1
kind: Service
metadata:
  name: karmada-scheduler-estimator-{{member_cluster_name}}
  namespace: karmada-system
  labels:
    cluster: {{member_cluster_name}}
spec:
  ports:
    - protocol: TCP
      port: 10352
      targetPort: 12345
  type: ExternalName
  externalName: 10.0.1.2
```

`spec.ports[0].port`: This is the same as "--scheduler-estimator-port" parameter of "karmada-scheduler".

`spec.ports[0].targetPort`: This is the Virtual Port provided by LVS.

`spec.externalName`: This is the Virtual IP provided by LVS.

Note: If this is a test cluster deployed on your Development Machine, you can also assign real IP and port to avoid LVS.

cmicat:

**What type of PR is this?**

/kind feature

See <https://github.com/karmada-io/karmada/issues/2487>

Fixes #2487

NONE

```release-note
NONE
```

### DIFF
--- a/pkg/estimator/client/service.go
+++ b/pkg/estimator/client/service.go
@@ -1,0 +1,59 @@
+package client
+
+import (
+	"context"
+	"fmt"
+	"net"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/client-go/kubernetes"
+)
+
+// ResolveCluster parses Service resource content by itself.
+// Fixes Issue https://github.com/karmada-io/karmada/issues/2487
+// Modified from "k8s.io/apiserver/pkg/util/proxy/proxy.go:92 => func ResolveCluster"
+func resolveCluster(kubeClient kubernetes.Interface, namespace, id string, port int32) (string, error) {
+	svc, err := kubeClient.CoreV1().Services(namespace).Get(context.TODO(), id, metav1.GetOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			/*
+			 * When Deploying Karmada in Host Kubernetes Cluster, the kubeClient will connect kube-apiserver
+			 * of Karmada Control Plane, rather than of host cluster.
+			 * But the Service resource is defined in Host Kubernetes Cluster. So we cannot get its content here.
+			 * The best thing we can do is just glue host:port together, and try to connect to it.
+			 */
+			return net.JoinHostPort(id, fmt.Sprintf("%d", port)), nil
+		}
+
+		return "", err
+	}
+
+	if svc.Spec.Type != corev1.ServiceTypeExternalName {
+		// We only support ExternalName type here.
+		// See discussions in PR: https://github.com/karmada-io/karmada/pull/2574#discussion_r979539389
+		return "", fmt.Errorf("unsupported service type %q", svc.Spec.Type)
+	}
+
+	svcPort, err := findServicePort(svc, port)
+	if err != nil {
+		return "", err
+	}
+	if svcPort.TargetPort.Type != intstr.Int {
+		return "", fmt.Errorf("ExternalName service type should have int target port, "+
+			"current target port: %v", svcPort.TargetPort)
+	}
+	return net.JoinHostPort(svc.Spec.ExternalName, fmt.Sprintf("%d", svcPort.TargetPort.IntVal)), nil
+}
+
+// findServicePort finds the service port by name or numerically.
+func findServicePort(svc *corev1.Service, port int32) (*corev1.ServicePort, error) {
+	for _, svcPort := range svc.Spec.Ports {
+		if svcPort.Port == port {
+			return &svcPort, nil
+		}
+	}
+	return nil, apierrors.NewServiceUnavailable(fmt.Sprintf("no service port %d found for service %q", port, svc.Name))
+}
--- a/hack/util.sh
+++ b/hack/util.sh
@@ -503,7 +503,7 @@ function util::fill_cabundle() {
   local ca_file=$1
   local conf=$2
 
-  local ca_string=$(sudo cat ${ca_file} | base64 | tr "\n" " "|sed s/[[:space:]]//g)
+  local ca_string=$(cat "${ca_file}" | base64 | tr "\n" " "|sed s/[[:space:]]//g)
   sed -i'' -e "s/{{caBundle}}/${ca_string}/g" "${conf}"
 }
 
--- a/pkg/descheduler/descheduler.go
+++ b/pkg/descheduler/descheduler.go
@@ -250,7 +250,7 @@ func (d *Descheduler) establishEstimatorConnections() {
 		return
 	}
 	for i := range clusterList.Items {
-		if err = estimatorclient.EstablishConnection(clusterList.Items[i].Name, d.schedulerEstimatorCache, d.schedulerEstimatorPort); err != nil {
+		if err = estimatorclient.EstablishConnection(d.KubeClient, clusterList.Items[i].Name, d.schedulerEstimatorCache, d.schedulerEstimatorPort); err != nil {
 			klog.Error(err)
 		}
 	}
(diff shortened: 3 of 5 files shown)


RainbowMango:
I'm not sure if you and the estimator author @Garrybest reached a consensus at #2487.
Here I just assign it to @Garrybest 
/assign @Garrybest

pkg/estimator/client/service.go
```diff
@@ -0,0 +1,69 @@
...
+	svc, err := kubeClient.CoreV1().Services(namespace).Get(context.TODO(), id, metav1.GetOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			/*
+			 * When Deploying Karmada in Host Kubernetes Cluster, the kubeClient will connect kube-apiserver
+			 * of Karmada Control Plane, rather than of host cluster.
+			 * But the Service resource is defined in Host Kubernetes Cluster. So we cannot get its content here.
+			 * The best thing we can do is just glue host:port together, and try to connect to it.
+			 */
+			return net.JoinHostPort(id, fmt.Sprintf("%d", port)), nil
```
Garrybest, line 29: Seems like now we resolve the DNS by ourselves instead of CoreDNS, right? For some scenarios without deploying CoreDNS, this patch may be helpful. I still have some questions.

The parameter `kubeClient` is corresponding with karmada-apiserver instead of host cluster. So normally we can't find any service here. Did you create a external service in karmada control plane?

BTW, if so, I think only external service will take effect. A `clusterIP` or `nodePort` service created in karmada control plane will never be connected cause the IP is fake.

pkg/estimator/client/service.go
```diff
@@ -0,0 +1,69 @@
+package client
+
+import (
+	"context"
+	"fmt"
+	"net"
+
+	"k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
```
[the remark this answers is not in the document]
cmicat: Fixed

pkg/estimator/client/service.go
cmicat: (1) 

> The parameter kubeClient is corresponding with karmada-apiserver instead of host cluster. So normally we can't find any service here. Did you create a external service in karmada control plane?

Yes, when I need to deploy karmada-scheduler without a host cluster,  I create a Service in karamda-apiserver.

(2) 

> BTW, if so, I think only external service will take effect. A clusterIP or nodePort service created in karmada control plane will never be connected cause the IP is fake.

Yes, I use create a Service with ExternalName type.

(3) 

The commented logic is for scheduler-estimator deployed in host cluster. Because karamda-apiserver don't have the required `Service`, I cannot parse `Service` content by myself. I fallback to the original logic, aka just try to connect to `host:port`.

pkg/estimator/client/service.go
```diff
@@ -0,0 +1,68 @@
...
+		}
+
+		return "", err
+	}
+
+	switch {
+	case svc.Spec.Type == v1.ServiceTypeClusterIP && svc.Spec.ClusterIP == v1.ClusterIPNone:
+		return "", fmt.Errorf(`cannot route to service with ClusterIP "None"`)
+	// use IP from a clusterIP for these service types
+	case svc.Spec.Type == v1.ServiceTypeClusterIP, svc.Spec.Type == v1.ServiceTypeLoadBalancer,
```
Garrybest, line 38: > BTW, if so, I think only external service will take effect. A clusterIP or nodePort service created in karmada control plane will never be connected cause the IP is fake.

Got it, I mean here, the svc is created in karmada-apiserver, so the cluster IP and node port is fake, right? We don't have any kube-proxy in karmada control plane.

pkg/estimator/client/service.go
```diff
@@ -0,0 +1,68 @@
+package client
+
+import (
+	"context"
+	"fmt"
+	"net"
+
+	v1 "k8s.io/api/core/v1"
```
Garrybest, line 8: using corev1 here

pkg/estimator/client/service.go
cmicat: I copied the logic from "k8s.io/apiserver/pkg/util/proxy/proxy.go:92 => func ResolveCluster", and didn't delete this logic.

I think ClusterIP, LoadBalancer, NodePort will not be used.

Maybe I should delete them?

pkg/estimator/client/service.go
cmicat: Is there any code style file I can use to configure GoLand? Or any tool to do code style check? The goimports tools have different style then `corev1 `.

Garrybest: Please check `hack/verify-import-aliases.sh` or `hack/.import-aliases`. corev1 is the alias for "k8s.io/api/core/v1"

pkg/estimator/client/service.go
Garrybest: Yes, these types will never be used. I think we could delete them.

pkg/estimator/client/service.go
cmicat: Eh, I found out `hack/verify-import-aliases.sh` `-include-path` doesn't include `pkg/estimator/client`, so `make verify` will not check go files under this path.

cmicat: Done

RainbowMango (approved):
/approve

PR #2582, opened 2022-09-27
Add scripts to install cli tools

Poor12:

**What type of PR is this?**
/kind feature

Add scripts to support one-click installation of cli tools.
```
curl -s https://raw.githubusercontent.com/Poor12/karmada/add-scripts-to-install-karmadactl/hack/install.sh | sudo bash
[INFO]  Downloading metadata https://api.github.com/repos/karmada-io/karmada/releases/latest
[INFO]  Using 1.3.0 as release
[INFO]  Downloading hash https://github.com/karmada-io/karmada/releases/download/v1.3.0/karmadactl-linux-amd64.tgz.sha256
[INFO]  Downloading binary https://github.com/karmada-io/karmada/releases/download/v1.3.0/karmadactl-linux-amd64.tgz
[INFO]  Verifying binary download
[INFO]  Installing karmadactl to /usr/local/bin/karmadactl
```

Fixes #

```
Now users can use ./hack/install.sh to support one-click installation of cli tools.
```

### DIFF
--- a/hack/install-cli.sh
+++ b/hack/install-cli.sh
@@ -1,0 +1,260 @@
+#!/usr/bin/env bash
+set -o errexit
+set -o nounset
+set -o pipefail
+
+# This script installs karmada cli tools to the install location you want.
+# Usage: ./install-cli.sh <install_cli_type>
+# Example:
+
+# ./install-cli.sh
+# will install the latest karmadactl to /usr/local/bin
+
+# ./install-cli.sh kubectl-karmada
+# will install the latest kubectl-karmada to /usr/local/bin
+
+# export INSTALL_CLI_VERSION=1.3.0
+# ./install-cli.sh
+# will install the 1.3.0 karmadactl to /usr/local/bin
+
+# INSTALL_CLI_TYPE supports 'karmadactl' or 'kubectl-karmada'.
+INSTALL_CLI_TYPE=${1:-"karmadactl"}
+# INSTALL_CLI_VERSION represents which version should be installed.
+# Defaults to 'latest' means latest release version.
+INSTALL_CLI_VERSION=${INSTALL_CLI_VERSION:-"latest"}
+# INSTALL_LOCATION represents where to put the CLI.
+INSTALL_LOCATION=${INSTALL_LOCATION:-"/usr/local/bin"}
+GITHUB_REPO="karmada-io/karmada"
+
+# Helper functions for logs
+info() {
+  echo '[INFO] ' "$@"
+}
+
+warn() {
+  echo '[WARN] ' "$@" >&2
+}
+
+fatal() {
+  echo '[ERROR] ' "$@" >&2
+  exit 1
+}
+
+# Set os, fatal if operating system not supported
+setup_verify_os() {
+  if [[ -z "${OS-}" ]]; then
+    OS=$(uname)
+  fi
+  case ${OS} in
+    Darwin)
+      OS=darwin
+      ;;
+    Linux)
+      OS=linux
+      ;;
+    *)
+      fatal "Unsupported operating system ${OS}"
+  esac
+}
+
+# Set arch, fatal if architecture not supported
+setup_verify_arch() {
+  if [[ -z "${ARCH-}" ]]; then
+    ARCH=$(uname -m)
+  fi
+  case ${ARCH} in
+    arm64|aarch64|armv8l)
+      ARCH=arm64
+      ;;
+    amd64|x86_64)
+      ARCH=amd64
+      ;;
+    *)
+      fatal "Unsupported architecture ${ARCH}"
+  esac
+}
+
+# Set cli tools, install karmadactl by default, fatal if cli tool not supported
+setup_install_cli() {
+  case ${INSTALL_CLI_TYPE} in
+    karmadactl|kubectl-karmada)
+      ;;
+    *)
+      fatal "Unsupported cli tool ${INSTALL_CLI_TYPE}"
+  esac
+}
+
+# Verify existence of downloader executable
+verify_downloader() {
+  # Return failure if it doesn't exist or is no executable
+  [[ -x "$(which "$1")" ]] || return 1
+
+  # Set verified executable as our downloader program and return success
+  DOWNLOADER=$1
+  return 0
+}
+
+# Create tempory directory and cleanup when done
+setup_tmp() {
+  TMP_DIR=$(mktemp -d -t "${INSTALL_CLI_TYPE}"-install.XXXXXXXXXX)
+  TMP_METADATA="${TMP_DIR}/${INSTALL_CLI_TYPE}.json"
+  TMP_HASH="${TMP_DIR}/${INSTALL_CLI_TYPE}.hash"
+  TMP_BIN="${TMP_DIR}/${INSTALL_CLI_TYPE}.tgz"
+  cleanup() {
+    local code=$?
+    rm -rf "${TMP_DIR}" || true
+    exit ${code}
+  }
+  trap cleanup INT EXIT
+}
+
+# Find version from Github metadata
+get_release_version() {
+  if [[ "${INSTALL_CLI_VERSION-}" != "latest" ]]; then
+    SUFFIX_URL="tags/v${INSTALL_CLI_VERSION}"
+  else
+    SUFFIX_URL="latest"
+  fi
+
+  METADATA_URL="https://api.github.com/repos/${GITHUB_REPO}/releases/${SUFFIX_URL}"
+
+  info "Downloading metadata ${METADATA_URL}"
+  download "${TMP_METADATA}" "${METADATA_URL}"
+
+  VERSION_KARMADA=$(grep '"tag_name":' "${TMP_METADATA}" | sed -E 's/.*"([^"]+)".*/\1/' | cut -c 2-)
+  if [[ -n "${VERSION_KARMADA}" ]]; then
+    info "Using ${VERSION_KARMADA} as release"
+  else
+    fatal "Unable to determine release version"
+  fi
+
+  ret=0
+  vercomp ${VERSION_KARMADA} 1.2.0 || ret=$?
+  if [[ $ret -eq 2 ]]; then
+    fatal "Installation of cli tools before version 1.2.0 is not supported"
+  fi
+}
+
+# Download from file from URL
+download() {
+  [[ $# -eq 2 ]] || fatal 'download needs exactly 2 arguments'
+
+  ret=0
+  case $DOWNLOADER in
+    curl)
+      curl -o "$1" -sfL "$2" || ret=$?
+      ;;
+    wget)
+      wget -qO "$1" "$2" || ret=$?
+      ;;
+    *)
+      fatal "Incorrect executable '${DOWNLOADER}'"
+      ;;
+    esac
+
+  # Abort if download command failed
+  [[ $ret -eq 0 ]] || fatal 'Download failed'
+}
+
+# Version comparison
+# Returns 0 on '=', 1 on '>', and 2 on '<'.
+# Ref: https://stackoverflow.com/a/4025065
+vercomp () {
+  if [[ $1 == $2 ]]
+  then
+    return 0
+  fi
+  local IFS=.
+  local i ver1=($1) ver2=($2)
+  # fill empty fields in ver1 with zeros
+  for ((i=${#ver1[@]}; i<${#ver2[@]}; i++))
+  do
+    ver1[i]=0
+  done
+  for ((i=0; i<${#ver1[@]}; i++))
+  do
+    if [[ -z ${ver2[i]} ]]
+    then
+      # fill empty fields in ver2 with zeros
+      ver2[i]=0
+    fi
+    if ((10#${ver1[i]} > 10#${ver2[i]}))
+    then
+      return 1
+    fi
+    if ((10#${ver1[i]} < 10#${ver2[i]}))
+    then
+      return 2
+    fi
+  done
+  return 0
+}
+
+# Download hash from Github URL
+download_hash() {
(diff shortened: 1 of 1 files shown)


hack/install.sh
```diff
@@ -0,0 +1,253 @@
...
+    HASH_BIN=$(compute_sha256sum "${TMP_BIN}")
+    HASH_BIN=${HASH_BIN%%[[:blank:]]*}
+    if [[ "${HASH_EXPECTED}" != "${HASH_BIN}" ]]; then
+        fatal "Download sha256 does not match ${HASH_EXPECTED}, got ${HASH_BIN}"
+    fi
+}
+
+# Setup permissions and move binary
+setup_binary() {
+    chmod 755 "${TMP_BIN}"
```
ikaven1024, line 229: Why shall we chmod it?

ikaven1024:
I prefer this usage of the script:
```
./install.sh  karmadactl
```

hack/install.sh
```diff
@@ -0,0 +1,253 @@
...
+
+    VERSION_KARMADA=$(grep '"tag_name":' "${TMP_METADATA}" | sed -E 's/.*"([^"]+)".*/\1/' | cut -c 2-)
+    if [[ -n "${VERSION_KARMADA}" ]]; then
+        info "Using ${VERSION_KARMADA} as release"
+    else
+        fatal "Unable to determine release version"
+    fi
+
+    set +e
+    vercomp ${VERSION_KARMADA} 1.2.0
```
ikaven1024, line 124: Only support above 1.2.0 because these have checksum?
Could we support below version without checksum verify?

hack/install.sh
```diff
@@ -0,0 +1,253 @@
...
+    if [[ $? -eq 2 ]]; then
```
ikaven1024, line 125: If it returns non 2 code, this script will not exit.
```
ret=0
vercomp ${VERSION_KARMADA} 1.2.0 || ret=$?
if [[ $ret -eq 0 ]]; then
  return
else ...
```

hack/install.sh
```diff
@@ -0,0 +1,253 @@
...
+}
+
+warn() {
+    echo '[WARN] ' "$@" >&2
+}
+
+fatal() {
+    echo '[ERROR] ' "$@" >&2
+    exit 1
+}
```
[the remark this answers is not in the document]
Poor12: It is hoped that users can install cli users without downloading the source code. Moving to util may not achieve the purpose.

hack/install.sh
```diff
@@ -0,0 +1,253 @@
...
+
+# Verify downloaded binary hash
+verify_binary() {
+    info "Verifying binary download"
+    HASH_BIN=$(compute_sha256sum "${TMP_BIN}")
+    HASH_BIN=${HASH_BIN%%[[:blank:]]*}
+    if [[ "${HASH_EXPECTED}" != "${HASH_BIN}" ]]; then
+        fatal "Download sha256 does not match ${HASH_EXPECTED}, got ${HASH_BIN}"
+    fi
```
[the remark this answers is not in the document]
Poor12: There is a problem here. If the user downloads the binary of 1.3.0, but the sha256 file is of 1.2.0, this command may not be recognized.

hack/install.sh
Poor12: The original intention is to ensure the permissions of karmadactl, which does not seem to have any effect. Delete it.

hack/install.sh
Poor12: There is one of a reason. Another is that the release before 1.2.0, we only output the kubectl-karmada file, which means the scripts cannot install karmdactl before v1.2.0.

hack/install.sh
Poor12: If return code is 2, it means version below 1.2.0. The script will exit. Otherwise, it will continue and install cli tools normally.

hack/install.sh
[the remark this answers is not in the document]
ikaven1024: Oh, you want users to install with
```
curl https://karmada-io/install.sh | bash
```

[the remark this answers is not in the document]
Poor12: Yeah.

hack/install.sh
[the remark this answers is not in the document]
ikaven1024: Binary and checksum files are downloaded in `download_binary` and `download_hash`, why they are unmatched?

hack/install.sh
ikaven1024: But users expect to install v1.2.0  kubectl-karmada with
```
export CLI=kubectl-karmada
export KARMADA_VERSION=kubectl-karmada
./install.sh 
```

Poor12: Yes, we can offer the installation of kubectl-karmada before v1.2.0. I guess it's not security enough to use. cc @RainbowMango for opinions.

hack/install.sh
ikaven1024: It's no recommended to turn off `errexit`. Any errors will be ignored, it's dangerous. Instead:
```
ret=0
vercomp ${VERSION_KARMADA} 1.2.0 || ret=$?
if [[ $ret -eq 2 ]]; then
 ...
```

hack/install.sh
```diff
@@ -0,0 +1,254 @@
...
+            ARCH=amd64
+            ;;
+        *)
+            fatal "Unsupported architecture ${ARCH}"
+    esac
+}
+
+# Set cli tools, install karmadactl by default, fatal if cli tool not supported
+setup_install_cli() {
+    if [[ -z "${CLI}" ]]; then
```
ikaven1024, line 63: How do you think put the cli variable to $1,
```
./install.sh karmadactl
```
I think it is more reasonable.

Poor12: Done

RainbowMango:
@Poor12 Please rebase your code and assign it to @ikaven1024 who can help review and approve this now.

hack/install.sh
```diff
@@ -0,0 +1,250 @@
...
+# Create tempory directory and cleanup when done
+setup_tmp() {
+    TMP_DIR=$(mktemp -d -t "${CLI}"-install.XXXXXXXXXX)
+    TMP_METADATA="${TMP_DIR}/${CLI}.json"
+    TMP_HASH="${TMP_DIR}/${CLI}.hash"
+    TMP_BIN="${TMP_DIR}/${CLI}.tgz"
+    cleanup() {
+        local code=$?
+        set +e
+        trap - EXIT
```
ikaven1024, line 95: Could you explain what's this statement mean.

hack/install.sh
```diff
@@ -0,0 +1,250 @@
...
+        rm -rf "${TMP_DIR}"
+        exit ${code}
+    }
+    trap cleanup INT EXIT
+}
+
+# Find version from Github metadata
+get_release_version() {
+    if [[ -n "${KARMADA_VERSION}" ]]; then
```
ikaven1024, line 104:
```suggestion
    if [[ -n "${KARMADA_VERSION-}" ]]; then
```

Or define this variable in the head this script, people can easily get all the variables they can use.

hack/install.sh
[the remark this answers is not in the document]
ikaven1024: > If the user downloads the binary of 1.3.0, but the sha256 file is of 1.2.0

I don't get why they are not same.

hack/install.sh
```diff
@@ -0,0 +1,250 @@
+#!/usr/bin/env bash
+set -o errexit
+set -o pipefail
```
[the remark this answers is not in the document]
Poor12: This is because we want the script to work when environment variables are not set, like {os}, {arch}.

hack/install.sh
Poor12: This is mainly to ensure that the temporary directory can be deleted when the script exits abnormally.

hack/install.sh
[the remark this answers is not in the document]
Poor12: Although this is unlikely to happen, I think it's safer to use the comparison method to ensure that the file and the sha256 file match and are safe.

hack/install.sh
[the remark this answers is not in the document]
ikaven1024: I prefer use `{os:-}`, rather remove remove `set -o nounset`

hack/install.sh
ikaven1024:
```
trap - EXIT
```

What its affect?

ikaven1024: > This is mainly to ensure that the temporary directory can be deleted when the script exits abnormally.

I prefer use below, rather than close the `errexit`
```bash
rm -rf "${TMP_DIR}" || true
```

Poor12: All has done.

hack/install.sh
```diff
@@ -0,0 +1,246 @@
...
+        fatal "Download sha256 does not match ${HASH_EXPECTED}, got ${HASH_BIN}"
+    fi
+}
+
+# Setup permissions and move binary
+setup_binary() {
+    info "Installing ${CLI} to ${BIN_DIR}/${CLI}"
+    tar -xzof "${TMP_BIN}" -C "${TMP_DIR}" --no-same-owner
+
+    local CMD_MOVE="mv -f \"${TMP_DIR}/${CLI}\" \"${BIN_DIR}\""
```
ikaven1024, line 226: shall we use `-i` instead of `-f`? I think override the existing file is dangerous.

hack/install.sh
```diff
@@ -0,0 +1,246 @@
...
+
+DEFAULT_BIN_DIR="/usr/local/bin"
+DEFAULT_CLI="karmadactl"
+CLI=${1:-"${DEFAULT_CLI}"}
+BIN_DIR=${2:-"${DEFAULT_BIN_DIR}"}
+GITHUB_REPO="karmada-io/karmada"
+
+# Helper functions for logs
+info() {
+    echo '[INFO] ' "$@"
```
ikaven1024, line 14: I see other scripts in out project, they are indented with two space. Could you modify it?

hack/install.sh
```diff
@@ -0,0 +1,246 @@
+#!/usr/bin/env bash
+set -o errexit
+set -o nounset
+set -o pipefail
+
```
ikaven1024, line 5: Add a usage of this script please.

Poor12: usage will add in our docs：https://github.com/karmada-io/website/pull/205

ikaven1024:
I test this script on my side, it downloads fail with bad network. But no error information prompts. Could you add this prompt?

```
➜  ./install.sh
[INFO]  Downloading metadata https://api.github.com/repos/karmada-io/karmada/releases/latest
[INFO]  Using 1.3.1 as release
[INFO]  Downloading hash https://github.com/karmada-io/karmada/releases/download/v1.3.1/karmadactl-darwin-arm64.tgz.sha256
```

Poor12:
> I test this script on my side, it downloads fail with bad network. But no error information prompts. Could you add this prompt?
> 
> ```
> ➜  ./install.sh
> [INFO]  Downloading metadata https://api.github.com/repos/karmada-io/karmada/releases/latest
> [INFO]  Using 1.3.1 as release
> [INFO]  Downloading hash https://github.com/karmada-io/karmada/releases/download/v1.3.1/karmadactl-darwin-arm64.tgz.sha256
> ```
I have added it at line 142. Does it simply exit the script if it fails to install?

ikaven1024:
> I have added it at line 142.

curl command execute fail and scripts exit immediately. Below fatal statement will not execute.

Maybe you shall change it to
```bash
  ret=0
  case $DOWNLOADER in
    curl)
      curl -o "$1" -sfL "$2" || ret=$?
      ;;
    wget)
      wget -qO "$1" "$2" || ret=$?
      ;;
    *)
      fatal "Incorrect executable '${DOWNLOADER}'"
      ;;
    esac

  # Abort if download command failed
  [[ $ret -eq 0 ]] || fatal 'Download failed'
```

ikaven1024:
/lgtm

@RainbowMango  will you take a look?

RainbowMango:
/hold

I think `hack/install.sh` should grant `755`, just like other scirpts:
```
-bash-4.2# hack/install.sh
-bash: hack/install.sh: Permission denied
```

And, the name should be more precise, like `hack/install-karmadactl.sh`.

ikaven1024:
> And, the name should be more precise, like `hack/install-karmadactl.sh`, `hack/install-cli.sh`.

I prefer `install-cli.sh`, these script install both karmadactl and kubectl-karmada.

```
hack/install-cli.sh  kubectl-karmama
```
It is human readable.

hack/install.sh
```diff
@@ -0,0 +1,247 @@
...
+DEFAULT_BIN_DIR="/usr/local/bin"
+DEFAULT_CLI="karmadactl"
+CLI=${1:-"${DEFAULT_CLI}"}
+BIN_DIR=${2:-"${DEFAULT_BIN_DIR}"}
```
RainbowMango, line 9:
```bash
# INSTALL_CLI_TYPE supports 'karmadactl' or 'kubectl-karmada'.
INSTALL_CLI_TYPE=${INSTALL_CLI_TYPE:-"karmadactl"}
# INSTALL_CLI_VERSION represents which version should be installed.
# Defaults to 'latest' means latest release version.
INSTALL_CLI_VERSION=${INSTALL_CLI_VERSION:-"latest"}
# INSTALL_LOCATION represents where to put the CLI.
INSTALL_LOCATION=${INSTALL_LOCATION:-"/usr/local/bin"}
```

We can put all variables that support customized at the head of the script.

In addition, we'd better describe the usage at the head of the script.

RainbowMango:
> I prefer install-cli.sh, these script install both karmadactl and kubectl-karmada.

+1, I just looked into the script. `install-cli.sh` is better.

hack/install.sh
ikaven1024: @RainbowMango what do you think about it?

RainbowMango: I'm ok that this script just supports v1.2.0+. For versions before v1.2, seems we just have `kubectl-karmada` in releases([like v1.1.0](https://github.com/karmada-io/karmada/releases/tag/v1.1.0)).

PR #2496, opened 2022-09-09
Add metrics for cluster

Poor12:
Signed-off-by: Poor12 <shentiecheng@huawei.com>

**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
Add metrics for cluster status, metrics name according to https://prometheus.io/docs/practices/naming/#metric-and-label-naming.
```
# HELP cluster_cpu_allocatable_number Number of allocatable CPU in the cluster.
# TYPE cluster_cpu_allocatable_number gauge
cluster_cpu_allocatable_number{cluster_name="member1"} 8
cluster_cpu_allocatable_number{cluster_name="member2"} 8
# HELP cluster_cpu_allocated_number Number of allocated CPU in the cluster.
# TYPE cluster_cpu_allocated_number gauge
cluster_cpu_allocated_number{cluster_name="member1"} 1.95
cluster_cpu_allocated_number{cluster_name="member2"} 0.9500000000000001
# HELP cluster_memory_allocatable_bytes Allocatable cluster memory resource in bytes.
# TYPE cluster_memory_allocatable_bytes gauge
cluster_memory_allocatable_bytes{cluster_name="member1"} 1.6817958912e+10
cluster_memory_allocatable_bytes{cluster_name="member2"} 1.6817958912e+10
# HELP cluster_memory_allocated_bytes Allocated cluster memory resource in bytes.
# TYPE cluster_memory_allocated_bytes gauge
cluster_memory_allocated_bytes{cluster_name="member1"} 3.0408704e+08
cluster_memory_allocated_bytes{cluster_name="member2"} 3.0408704e+08
# HELP cluster_node_number Number of nodes in the cluster.
# TYPE cluster_node_number gauge
cluster_node_number{cluster_name="member1"} 1
cluster_node_number{cluster_name="member2"} 1
# HELP cluster_pod_allocatable_number Number of allocatable pods in the cluster.
# TYPE cluster_pod_allocatable_number gauge
cluster_pod_allocatable_number{cluster_name="member1"} 110
cluster_pod_allocatable_number{cluster_name="member2"} 110
# HELP cluster_pod_allocated_number Number of allocated pods in the cluster.
# TYPE cluster_pod_allocated_number gauge
cluster_pod_allocated_number{cluster_name="member1"} 11
cluster_pod_allocated_number{cluster_name="member2"} 10
# HELP cluster_ready_node_number Number of ready nodes in the cluster.
# TYPE cluster_ready_node_number gauge
[...]

PR #2594, opened 2022-09-28
set ReadHeaderTimeout for http server

RainbowMango:
**What type of PR is this?**
/kind cleanup

This PR adds `ReadHeaderTimeout` to HTTP server to solve security concerns.

Reference: 
- https://deepsource.io/directory/analyzers/go/issues/GO-S2114
- https://en.wikipedia.org/wiki/Slowloris_(computer_security)

Fixes #

Required by #2591 

```release-note
NONE
```


RainbowMango:
/assign @Garrybest @Poor12 
I'm thinking if we need to introduce a common function `serveHealthzAndMetrics` for sharing by components. But not sure where to put it.

Poor12:
> /assign @Garrybest @Poor12 I'm thinking if we need to introduce a common function `serveHealthzAndMetrics` for sharing by components. But not sure where to put it.

For some components, we can. But for other controllers builded by controller-runtime, the way the metrics are implemented may vary. https://github.com/karmada-io/karmada/pull/2496 is an example.

Garrybest:
> For some components, we can. But for other controllers builded by controller-runtime, the way the metrics are implemented may vary. #2496 is an example.

I found `controller-runtime` sets `ReadHeaderTimeout` to 32s, so I think we don't need to care about controllers.
https://github.com/karmada-io/karmada/blob/main/vendor/sigs.k8s.io/controller-runtime/pkg/internal/httpserver/server.go

BTW, is 3s too short here because `controller-runtime` sets it to 32s? I'm not sure.

RainbowMango:
> BTW, is 3s too short here because controller-runtime sets it to 32s? I'm not sure.

Nice finding! I found an article [`Reasonable default timeouts for HTTP servers?`](https://www.reddit.com/r/golang/comments/cz7m9b/reasonable_default_timeouts_for_http_servers/). A larger timeout is used to deal with a large request, especially in cases request from a proxy.

I don't think we really need so long timeout here, because the HTTP server touched by this patch won't accept large requests. But given the [kube-controller-manager](https://github.com/kubernetes/kubernetes/blob/b4b67c9ca52b7baa9a4c97f6d5cd6e61df5ee541/staging/src/k8s.io/apiserver/pkg/server/secure_serving.go#L172) also set it to 32s for `/healthz`. So, no big deal, we can set it to 32s too :).

PR #2598, opened 2022-09-29
Fix a problem with the API interface for creating a propagationpolicy when neither the namespace of the propagationpolicy nor the resourceselectors is set

### LINKED ISSUE #2597: There is a problem with the API interface for creating a propagationpolicy when neither the namespace of the propagationpolicy nor the resourceselectors is set
**What happened**:
When I use the API interface to create a **PropagationPolicy**, the namespace of the PropagationPolicy is not set, nor the namespace in the resourceSeletors. The template of the PropagationPolicy is as follows:
```
{
    "apiVersion": "policy.karmada.io/v1alpha1",
    "kind": "PropagationPolicy",
    "metadata": {
        "name": "example-policy"
    },
    "spec": {
        "resourceSelectors": [
            {
            "apiVersion": "apps/v1",
            "kind": "Deployment",
            "name": "nginx"
            }
        ],
        "placement": {
            "clusterAffinity": {
                "clusterNames": ["test72-k8s"]
            }
        }
    }
}
```
Although the deployment is successfully distributed and the namespace of the PropagationPolicy is set to "default" by default, the namespace in the resourceSeletors does not exist, see below:

image

Next, when I edit and modify this PropagationPolicy:

image

The following error occurs:

image

I hope the deployment can be distributed normally after modifying the PropagationPolicy

**Environment**:
- Karmada version: 1.2
- kubectl-karmada or karmadactl version (the result of `kubectl-karmada version` or `karmadactl version`): 1.2
- Others:

ranzhouol:
**What type of PR is this?**
/kind cleanup

When the API interface is called to create a PropagationPolicy, and the namespaces of both the PropagationPolicy and resourceSelectors are not set, we automatically set the namespace in resourceSelectors to default. You can based on https://github.com/karmada-io/karmada/issues/2597
Fixes #2597

```release-note
NONE
```

### DIFF
--- a/pkg/webhook/propagationpolicy/mutating.go
+++ b/pkg/webhook/propagationpolicy/mutating.go
@@ -38,8 +38,13 @@ func (a *MutatingAdmission) Handle(ctx context.Context, req admission.Request) a
 	// Set default namespace for all resource selector if not set.
 	for i := range policy.Spec.ResourceSelectors {
 		if len(policy.Spec.ResourceSelectors[i].Namespace) == 0 {
-			klog.Infof("Setting resource selector default namespace for policy: %s/%s", policy.Namespace, policy.Name)
-			policy.Spec.ResourceSelectors[i].Namespace = policy.Namespace
+			if len(policy.Namespace) == 0 {
+				klog.Infof("Setting resource selector default namespace for policy: %s/%s", policy.Namespace, policy.Name)
+				policy.Spec.ResourceSelectors[i].Namespace = defaultNameSpace
+			} else {
+				klog.Infof("Setting resource selector default namespace for policy: %s/%s", policy.Namespace, policy.Name)
+				policy.Spec.ResourceSelectors[i].Namespace = policy.Namespace
+			}
 		}
 	}
 
--- a/pkg/webhook/propagationpolicy/mutating.go
+++ b/pkg/webhook/propagationpolicy/mutating.go
@@ -19,6 +19,8 @@ type MutatingAdmission struct {
 	decoder *admission.Decoder
 }
 
+const defaultNameSpace = "default"
+
 // Check if our MutatingAdmission implements necessary interface
 var _ admission.Handler = &MutatingAdmission{}
 var _ admission.DecoderInjector = &MutatingAdmission{}


pkg/webhook/propagationpolicy/mutating.go
```diff
@@ -36,8 +38,13 @@ func (a *MutatingAdmission) Handle(ctx context.Context, req admission.Request) a
 	// Set default namespace for all resource selector if not set.
 	for i := range policy.Spec.ResourceSelectors {
 		if len(policy.Spec.ResourceSelectors[i].Namespace) == 0 {
-			klog.Infof("Setting resource selector default namespace for policy: %s/%s", policy.Namespace, policy.Name)
-			policy.Spec.ResourceSelectors[i].Namespace = policy.Namespace
```
RainbowMango, line 40: Wow, I just realized the policy doesn't have a default namespace yet by this time.

RainbowMango: Do you know when the `kube-apiserver` would set the default namespace for CR resources?

ranzhouol:
I don't know when the kube apiserver would set the default namespace for CR resources, but in this case, it is obviously after Karmada's webhook. If we can put Karmada's webhook behind Kubernetes's webhook, this problem can be solved.




------------------&nbsp;原始邮件&nbsp;------------------
发件人:                                                                                                                        "karmada-io/karmada"                                                                                    ***@***.***&gt;;
发送时间:&nbsp;2022年9月29日(星期四) 晚上6:02
***@***.***&gt;;
***@***.******@***.***&gt;;
主题:&nbsp;Re: [karmada-io/karmada] Fix a problem with the API interface for creating a propagationpolicy when neither the namespace of the propagationpolicy nor the resourceselectors is set (PR #2598)





 
@RainbowMango commented on this pull request.
 
 
In pkg/webhook/propagationpolicy/mutating.go:
 &gt; @@ -36,8 +38,13 @@ func (a *MutatingAdmission) Handle(ctx context.Context, req admission.Request) a  	// Set default namespace for all resource selector if not set.  	for i := range policy.Spec.ResourceSelectors {  		if len(policy.Spec.ResourceSelectors[i].Namespace) == 0 { -			klog.Infof("Setting resource selector default namespace for policy: %s/%s", policy.Namespace, policy.Name) -			policy.Spec.ResourceSelectors[i].Namespace = policy.Namespace  
Wow, I just realized the policy doesn't have a default namespace yet by this time.
 
 
In pkg/webhook/propagationpolicy/mutating.go:
 &gt; @@ -36,8 +38,13 @@ func (a *MutatingAdmission) Handle(ctx context.Context, req admission.Request) a  	// Set default namespace for all resource selector if not set.  	for i := range policy.Spec.ResourceSelectors {  		if len(policy.Spec.ResourceSelectors[i].Namespace) == 0 { -			klog.Infof("Setting resource selector default namespace for policy: %s/%s", policy.Namespace, policy.Name) -			policy.Spec.ResourceSelectors[i].Namespace = policy.Namespace  
Do you know when the kube-apiserver would set the default namespace for CR resources?

RainbowMango:
> but in this case, it is obviously after Karmada's webhook.

Yes. Let's wait for a while, let me try to find where Kubernetes set the default namespace.

I'm ok with this patch, but I'm also thinking of another approach:
We might don't need to set the default namespace for `ResourceSelector`, because we always try matching resources with the PropagationPolicy from the same namespace. See,
https://github.com/karmada-io/karmada/blob/c817ace443bc32f06ee091fc2bd1b35cfe942343/pkg/detector/detector.go#L356
and
https://github.com/karmada-io/karmada/blob/c817ace443bc32f06ee091fc2bd1b35cfe942343/pkg/util/selector.go#L17

In addition, I guess this is more like a cleanup.
/remove-kind bug
/kind cleanup

RainbowMango:
Hi @ranzhouol ,
Thank you for sending this patch.

Since @carlory found this issue either and sent a patch(#2858) for this. 
And I think the approach in #2858 would be more reasonable, we can't just use `default` as a default namespace name, because  Kubernetes supports defining the default namespace name. I'm not quite sure about that. But getting the namespace from the request is a safe choice.

So, this PR as well as the relevant issue will be closed by #2858. 

Sorry for let this sit. And thanks again for your contribution.

ranzhouol:
It's ok&nbsp;


一颗糖
***@***.***



&nbsp;




------------------&nbsp;原始邮件&nbsp;------------------
发件人: "Hongcai ***@***.***&gt;; 
发送时间: 2022年11月26日(星期六) 下午5:58
收件人: ***@***.***&gt;; 
抄送: ***@***.***&gt;; ***@***.***&gt;; 
主题: Re: [karmada-io/karmada] Fix a problem with the API interface for creating a propagationpolicy when neither the namespace of the propagationpolicy nor the resourceselectors is set (PR #2598)





 
Hi @ranzhouol ,
 Thank you for sending this patch.
 
Since @carlory found this issue either and sent a patch(#2858) for this.
 And I think the approach in #2858 would be more reasonable, we can't just use default as a default namespace name, because  Kubernetes supports defining the default namespace name. I'm not quite sure about that. But getting the namespace from the request is a safe choice.
 
So, this PR as well as the relevant issue will be closed by #2858.
 
Sorry for let this sit. And thanks again for your contribution.

Issue #2579, opened 2022-09-26
[umbrella] UT Supplemental Plan


XiShanYongYe-Chang:
To improve the reliability of the Karmada project, verify the implementation and usability of functions, and improve user experience, we plan to supplement the UT case. 

We'll list all the tasks in the google execel: https://docs.google.com/spreadsheets/d/1691AczZtovlF8t_UzGjEHCfSde5GHLPsjREGvOpmGDA/edit?hl=zh-cn#gid=0.

You can pick up any tasks you're interested in, thanks all!

> The task list is based on files. We need to improve the code coverage of the current file as much as possible.

> You can also let me know if you think there are some files that need to be added to UT but aren't listed in the task list, thanks~

RainbowMango:
@XiShanYongYe-Chang I unpinned this issue, Is there a gate to mark this issue complete?

XiShanYongYe-Chang:
> @XiShanYongYe-Chang I unpinned this issue, Is there a gate to mark this issue complete?

My initial idea was to supplement the UT of the previously identified files https://docs.google.com/spreadsheets/d/1691AczZtovlF8t_UzGjEHCfSde5GHLPsjREGvOpmGDA/edit?hl=zh-cn#gid=0. Considering that we will recently start an [lfx project](https://github.com/karmada-io/karmada/issues/3086) to enhance testing coverage, I think we can close the current issue first.

Since this issue was raised, our coverage rate has increased by [26%](https://app.codecov.io/gh/karmada-io/karmada?search=&trend=6%20months), and now the coverage rate has reached 49%. Thank you for all your contributions.

RainbowMango:
That makes sense to me.

Closing his doesn't means we are not accepting UT contributions anymore.
/close

PR #2607, opened 2022-10-01
add ut for util/metrics

my-git9:
**What type of PR is this?**
add ut for util/metrics

https://github.com/karmada-io/karmada/issues/2579

```release-note

```

### DIFF
--- a/pkg/util/metrics/metrics_test.go
+++ b/pkg/util/metrics/metrics_test.go
@@ -1,0 +1,32 @@
+package metrics
+
+import (
+	"errors"
+	"testing"
+)
+
+func TestGetResultByError(t *testing.T) {
+	tests := []struct {
+		name  string
+		error error
+		want  string
+	}{
+		{
+			name:  "error is nil",
+			error: nil,
+			want:  "success",
+		},
+		{
+			name:  "error is not nil",
+			error: errors.New("hello,error"),
+			want:  "error",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := GetResultByError(tt.error); got != tt.want {
+				t.Errorf("GetResultByError() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
--- a/pkg/util/metrics/metrics.go
+++ b/pkg/util/metrics/metrics.go
@@ -19,16 +19,6 @@ func GetResultByError(err error) string {
 	return ResultSuccess
 }
 
-// DurationInMicroseconds gets the time in microseconds.
-func DurationInMicroseconds(start time.Time) float64 {
-	return float64(time.Since(start).Nanoseconds()) / float64(time.Microsecond.Nanoseconds())
-}
-
-// DurationInMilliseconds gets the time in milliseconds.
-func DurationInMilliseconds(start time.Time) float64 {
-	return float64(time.Since(start).Nanoseconds()) / float64(time.Millisecond.Nanoseconds())
-}
-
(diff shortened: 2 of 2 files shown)


pkg/util/metrics/metrics_test.go
```diff
@@ -0,0 +1,88 @@
...
+		start time.Time
+	}{
+		{
+			name:  "testMicroseconds",
+			start: time.Date(2022, 10, 1, 7, 8, 9, 0, time.Local),
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := DurationInMicroseconds(tt.start)
```
Poor12, line 48: I'm not sure that for these functions about time, such a test method is sufficient.

pkg/util/metrics/metrics_test.go
```diff
@@ -0,0 +1,88 @@
...
+)
+
+func TestGetResultByError(t *testing.T) {
+	tests := []struct {
+		name  string
+		error error
+		want  string
+	}{
+		{
+			name:  "testnoerr",
```
Poor12, line 17: Test case names can be more specific.

pkg/util/metrics/metrics_test.go
my-git9: I have the same thought, but this function can't get a definite value for comparison. Can you give me some advice? Thanks!

XiShanYongYe-Chang:
ping @Poor12

pkg/util/metrics/metrics_test.go
Poor12: /cc @XiShanYongYe-Chang for opinion.

XiShanYongYe-Chang: Personally, I think it makes no sense to test this function. I suggest removing unused functions.
Hi @my-git9 How do you think?

my-git9: Do you mean we want to delete unused functions from "pkg/util/metrics/metrics.go"? It seems that none of the functions inside were used

XiShanYongYe-Chang: > Do you mean we want to delete unused functions from "pkg/util/metrics/metrics.go"? 

Yes.

Those two methods seem not used:

https://github.com/karmada-io/karmada/blob/594ad9f44b67d199354ce375c9df43adcb6187e9/pkg/util/metrics/metrics.go#L22-L30

my-git9: Has beed updated, please review again, thx

XiShanYongYe-Chang: Thanks~ There is a lint error.

pkg/util/metrics/metrics_test.go
```diff
@@ -0,0 +1,52 @@
...
+			start: time.Date(2022, 10, 1, 7, 8, 9, 0, time.Local),
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := DurationInSeconds(tt.start)
+			fmt.Printf("DurationInSeconds() = %v\n", got)
+		})
+	}
+}
```
XiShanYongYe-Chang, line 52: Hi @my-git9, This test does not judge the test result as expected and can be removed.

my-git9: Updated

XiShanYongYe-Chang (commented):
Thanks~
/lgtm

RainbowMango (approved):
/approve
Thanks.

PR #2624, opened 2022-10-08
chart: support controller manager feature gate

calvin0327:

**What type of PR is this?**
/kind feature

we want to use the feature gate `Propagate dependent`, we have not a swith to open the feature.

image

Fixes #

please correct me if not

```release-note
NONE
```

### DIFF
--- a/charts/karmada/templates/_helpers.tpl
+++ b/charts/karmada/templates/_helpers.tpl
@@ -390,4 +390,18 @@ Return the proper karmada kubectl image name
 {{ include "common.images.image" (dict "imageRoot" .Values.kubectl.image "global" .Values.global) }}
 {{- end -}}
 
+{{- define "karmada.controllerManager.featureGates" -}}
+     {{- if (not (empty .Values.controllerManager.featureGates)) }}
+          {{- $featureGatesFlag := "" -}}
+          {{- range $key, $value := .Values.controllerManager.featureGates -}}
+               {{- if not (empty (toString $value)) }}
+                    {{- $featureGatesFlag = cat $featureGatesFlag $key "=" $value ","  -}}
+               {{- end -}}
+          {{- end -}}
 
+          {{- if gt (len $featureGatesFlag) 0 }}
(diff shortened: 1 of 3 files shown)


calvin0327:
@Poor12 PTAL

charts/karmada/templates/_helpers.tpl
```diff
@@ -390,4 +390,18 @@ Return the proper karmada kubectl image name
...
+                    {{- $featureGatesFlag = cat $featureGatesFlag $key "=" $value ","  -}}
+               {{- end -}}
+          {{- end -}}
+
+          {{- if gt (len $featureGatesFlag) 0 }}
+               {{- $featureGatesFlag := trimSuffix "," $featureGatesFlag  | nospace -}}
+               {{- printf "%s=%s" "--feature-gates" $featureGatesFlag -}}
+          {{- end -}}
+     {{- end -}}
+{{- end -}}
```
Poor12, line 407: Add a new line here.

calvin0327: done

charts/karmada/values.yaml
```diff
@@ -285,6 +285,10 @@ controllerManager:
     rollingUpdate:
       maxUnavailable: 0
       maxSurge: 50%
+  ## @param featureGate to controllerManager
+  featureGates:
+    ## @param PropagateDeps is a feature gate for the controllerManager to allow propagate dependent respurce to workloads.
+    PropagateDeps: false
```
Poor12, line 291: I am thinking that propagateDeps=false is a reasonable default value. How about setting the featureGates to empty by default?

calvin0327: Empty is same effect with false by default. I think it's better to set a exact value to controller manager. 
the user can yet know the current value whart is.

Poor12: It seems make sense.

RainbowMango (approved):
/approve

Issue #2618, opened 2022-10-08
PP's ResourceSelector is empty


chaunceyjiang:
**Please provide an in-depth description of the question you have**:

When I use the following yaml file to create PP, it can be created successfully, but when I look at the api definition of PP, [ ResourceSelector cannot be empty](https://github.com/karmada-io/karmada/blob/master/pkg/apis/policy/v1alpha1/propagation_types.go#L42-L46). Is this normal?

``` yaml
apiVersion: policy.karmada.io/v1alpha1
kind: PropagationPolicy
metadata:
  name: nginx-propagation
spec:
  resourceSelectors: []
  placement:
    clusterAffinity:
      clusterNames:
        - member1
        - member2
    replicaScheduling:
      replicaDivisionPreference: Weighted
      replicaSchedulingType: Divided
      weightPreference:
        staticWeightList:
          - targetCluster:
              clusterNames:
                - member1
            weight: 1
          - targetCluster:
              clusterNames:
                - member2
            weight: 1

```

<img width="605" alt="image" src="https://user-images.githubusercontent.com/17962021/194683183-12d13d9f-8cc9-4dc2-9741-771a1c8cb897.png">

**What do you think about this question?**:

If this is not the expected behavior, should the user be prevented from creating a PP with an empty ResourceSelector?

**Environment**:
- Karmada version: v1.3.0
- Kubernetes version: v1.24.0
- Others:

RainbowMango:
Good question, it works as expected.
An empty `ResourceSelector` matches all resources across different resource types.

> resourceSelectors: []

`ResourceSelector` not defaults to `empty`, and this is the way to set it to `empty` explicitly.

chaunceyjiang:
> An empty ResourceSelector matches all resources across different resource types.

Does this mean that all resources across different resource types will be propagated to the member cluster?

chaunceyjiang:
The reason why I asked this question is that I found that when I created the above-mentioned PP, the resources would not be propagated to the member clusters. 

The behaviors of the PP and the OP are not consistent (the OP with an empty ResourceSelector can be applied, but the PP cannot be applied)

RainbowMango:
> Does this mean that all resources across different resource types will be propagated to the member cluster?

[...]

Issue #1618, opened 2022-04-15
cluster/proxy 下长链接问题


ikaven1024:
**What happened**:
使用cluster/proxy来测试watch、logs -f、exec操作，均有失败结果。

0. 准备
```
$ cat ~/.kube/karmada-cluster-2.config
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: xxx
    server: https://192.168.50.183:32443/apis/cluster.karmada.io/v1alpha1/clusters/cluster2/proxy
```

2. Watch
```
$ kubectl --kubeconfig ~/.kube/karmada-cluster-2.config get po -w
NAME                     READY   STATUS    RESTARTS   AGE
nginx-6d8dbbb794-lrp9f   1/1     Running   0          9m15s
nginx-6d8dbbb794-r4sq6   1/1     Running   0          9m15s
STATUS                   REASON          MESSAGE
Failure                  InternalError   an error on the server ("unable to decode an event from the watch stream: stream error: stream ID 69; INTERNAL_ERROR; received from peer") has prevented the request from succeeding
```

3. logs
```
$ kubectl --kubeconfig ~/.kube/karmada-cluster-2.config logs -f nginx-6d8dbbb794-lrp9f
/docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration
/docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/
/docker-entrypoint.sh: Launching /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh
10-listen-on-ipv6-by-default.sh: info: Getting the checksum of /etc/nginx/conf.d/default.conf
10-listen-on-ipv6-by-default.sh: info: Enabled listen on IPv6 in /etc/nginx/conf.d/default.conf
/docker-entrypoint.sh: Launching /docker-entrypoint.d/20-envsubst-on-templates.sh
/docker-entrypoint.sh: Launching /docker-entrypoint.d/30-tune-worker-processes.sh
/docker-entrypoint.sh: Configuration complete; ready for start up
2022/04/15 10:19:52 [notice] 1#1: using the "epoll" event method
2022/04/15 10:19:52 [notice] 1#1: nginx/1.21.6
2022/04/15 10:19:52 [notice] 1#1: built by gcc 10.3.1 20211027 (Alpine 10.3.1_git20211027)
2022/04/15 10:19:52 [notice] 1#1: OS: Linux 4.19.202
[...]

PR #2625, opened 2022-10-09
Update comments for ResourceSelectors in PropagationPolicy to eliminate ambiguity

RainbowMango:
**What type of PR is this?**
/kind cleanup

People might be confused about why the behavior of `ResourceSelectors` is different between `PropagationPolicy` and `OverridePolicy`.(see #2618)

This PR adds a clear comment on it to eliminate ambiguity according to [the discussion](https://github.com/karmada-io/karmada/pull/422#issuecomment-858222351).

Fixes #

With this patch, when applying with an empty resource selector as follow:
```bash
-bash-4.2# cat samples/nginx/propagationpolicy.yaml 
kind: PropagationPolicy
metadata:
  name: nginx-propagation
spec:
  resourceSelectors: []
#    - apiVersion: apps/v1
#      kind: Deployment
#      name: nginx
...
```
An error would raise up:
```bash
The PropagationPolicy "nginx-propagation" is invalid: spec.resourceSelectors: Invalid value: 0: spec.resourceSelectors in body should have at least 1 items
```

```release-note
NONE
```

### DIFF
--- a/pkg/apis/policy/v1alpha1/propagation_types.go
+++ b/pkg/apis/policy/v1alpha1/propagation_types.go
@@ -42,7 +42,11 @@ type PropagationPolicy struct {
 // PropagationSpec represents the desired behavior of PropagationPolicy.
 type PropagationSpec struct {
 	// ResourceSelectors used to select resources.
+	// Nil or empty selector is not allowed and doesn't mean match all kinds
+	// of resources for security concerns that sensitive resources(like Secret)
+	// might be accidentally propagated.
 	// +required
+	// +kubebuilder:validation:MinItems=1
 	ResourceSelectors []ResourceSelector `json:"resourceSelectors"`
 
 	// Association tells if relevant resources should be selected automatically.
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -15328,7 +15328,7 @@
           "type": "boolean"
         },
         "resourceSelectors": {
-          "description": "ResourceSelectors used to select resources.",
+          "description": "ResourceSelectors used to select resources. Nil or empty selector is not allowed and doesn't mean match all kinds of resources for security concerns that sensitive resources(like Secret) might be accidentally propagated.",
           "type": "array",
           "items": {
             "default": {},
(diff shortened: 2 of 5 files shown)


chaunceyjiang:
> .(see https://github.com/karmada-io/karmada/issues/1618)

Maybe is #2618 ?

RainbowMango:
Yeah, right.

pkg/apis/policy/v1alpha1/propagation_types.go
```diff
@@ -42,6 +42,9 @@ type PropagationPolicy struct {
 // PropagationSpec represents the desired behavior of PropagationPolicy.
 type PropagationSpec struct {
 	// ResourceSelectors used to select resources.
+	// Nil selector is not allowed and an empty selector selects nothing for
```
chaunceyjiang, line 45: As far as I know, `resourceSelectors` are not allowed to be modified, so it is meaningless for users to create empty resource selectors. If users are not prevented from creating empty resource selectors, will the empty resource selectors accidentally created by users become technical debt?

RainbowMango: Good questions.

> As far as I know, resourceSelectors are not allowed to be modified,

We are going to enable it. See #2562.

> so it is meaningless for users to create empty resource selectors.

Yeah, I agree with that. a PropagationPolicy with an empty `ResourceSelector` looks weird. 
I don't know if the `kubebuilder:` markers can help to restrict the length.(restrict to len(resourceSelectors) > 0). If not, we can add the restriction logic to `karmada-webhook`. How do you think?

RainbowMango: Just checked the marker, maybe `+kubebuilder:validation:MinItems:=1` would help.
https://book.kubebuilder.io/reference/markers/crd-validation.html

chaunceyjiang: > +kubebuilder:validation:MinItems:=1

+1

RainbowMango: Done. and I put the test report into to PR description.

PR #2631, opened 2022-10-12
Unit Test registry.go

cmicat:

**What type of PR is this?**

/kind cleanup

Add Unit Test for registry.go

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/framework/runtime/registry_test.go
+++ b/pkg/scheduler/framework/runtime/registry_test.go
@@ -7,13 +7,15 @@ import (
 	"github.com/karmada-io/karmada/pkg/scheduler/framework"
 )
 
+func mockPluginFactory() (framework.Plugin, error) {
+	return nil, nil
+}
+
 func TestRegistry_Filter(t *testing.T) {
 	plugins := []string{"bar", "foo", "fuzz"}
 	var r = make(Registry)
 	for _, name := range plugins {
-		_ = r.Register(name, func() (framework.Plugin, error) {
-			return nil, nil
-		})
+		_ = r.Register(name, mockPluginFactory)
 	}
 
 	tests := []struct {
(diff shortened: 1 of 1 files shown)


pkg/scheduler/framework/runtime/registry_test.go
```diff
@@ -56,3 +58,108 @@ func TestRegistry_Filter(t *testing.T) {
...
+		{
+			name:              "p1-2",
+			pluginFactoryName: "p1",
+			wantErr:           true,
+			expectedPlugins:   []string{"p1", "p2"},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := r.Register(tt.pluginFactoryName, mockPluginFactory)
```
RainbowMango, line 92: I think each sub-test shouldn't be coupled with the other. We can initialize the registry in every sub-test.
Maybe we can orginze the tests by :
```golang
	tests := []struct {
		name              string
		initialPlugins    []string
		registeringPlugin string
		wantErr           bool
		expectedPlugins   []string
	}{
		{
			name:              "plugin registered to a empty registry",
			initialPlugins:    nil,
			registeringPlugin: "foo",
			wantErr:           false,
			expectedPlugins:   []string{"foo"},
		},
		{
			name:              "plugin registered to a none empty registry",
			initialPlugins:    []string{"foo"},
			registeringPlugin: "bar",
			wantErr:           false,
			expectedPlugins:   []string{"foo", "bar"},
		},
		{
			name:              "plugin already exists",
			initialPlugins:    []string{"foo", "bar"},
			registeringPlugin: "foo",
			wantErr:           true,
			expectedPlugins:   []string{"foo", "bar"},
		},
	}
```

pkg/scheduler/framework/runtime/registry_test.go
```diff
@@ -56,3 +58,108 @@ func TestRegistry_Filter(t *testing.T) {
...
+
+			if (err != nil) != tt.wantErr {
+				t.Errorf("Register() error = %v, wantErr %v", err, tt.wantErr)
+			}
+
+			if !reflect.DeepEqual(r.FactoryNames(), tt.expectedPlugins) {
```
RainbowMango, line 98: We can't just `deepEqual` two slices here, because the `first slice` is dumped from a map, and the item is unordered.

pkg/scheduler/framework/runtime/registry_test.go
cmicat: Good point

pkg/scheduler/framework/runtime/registry_test.go
cmicat: Eh, I copied this logic from original test. I will fix this.

cmicat: I found out the `FactoryNames()` already did the sort operation, so there is no need to sort by ourselves.

RainbowMango: Fine. But a little bit tricky because we mush ensure the test data is orderd.

RainbowMango (approved):
/lgtm
/approve

PR #2633, opened 2022-10-13
[fix] close result chan when watchMux stopped

ikaven1024:

**What type of PR is this?**
/kind bug

fix result chan in watchMux not closed when stop.
https://github.com/karmada-io/karmada/blob/e7d4f46ea1cff5570bd993b98f21351619dfd7d9/pkg/search/proxy/store/util.go#L143-L151

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/search/proxy/store/util.go
+++ b/pkg/search/proxy/store/util.go
@@ -104,7 +104,7 @@ type decoratedWatcher struct {
 }
 
 type watchMux struct {
-	lock    sync.Mutex
+	lock    sync.RWMutex
 	sources []decoratedWatcher
 	result  chan watch.Event
 	done    chan struct{}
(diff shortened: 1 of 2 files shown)


ikaven1024:
> Does this need to be cherry-picked?

These matter is not seriousness, the consumer goroutine will quick by `done` and `timeoutCh`. So I think no need to cherry pick it.
https://github.com/karmada-io/karmada/blob/614e28508336d6c03a938ce1bf0678dafef034f0/vendor/k8s.io/apiserver/pkg/endpoints/handlers/watch.go#L233-L244

> and do we need a release note?

This doesn't  introduce a user-facing change, so I didn't add release note.

RainbowMango:
ok.
/lgtm
/approve

Issue #2371, opened 2022-08-12
[lfx-mentorship-2022-Sept-Nov] Configurable Interpreter


RainbowMango:
**What would you like to be added**:
Now in [Resource Interpreter](https://karmada.io/docs/userguide/globalview/customizing-resource-interpreter) framework, we enabled both `built-in` and `customized` interpreter. 

We'd like to enable the third interpreter, which is `configurable` interpreter.

In addition, for the `Retain` interpreter operation, we have [the proposal ](https://github.com/karmada-io/karmada/tree/master/docs/proposals/configurable-local-value-retention). 

**Why is this needed**:

PradyumnaKrishna:
Hey @RainbowMango I am interested in this project for LFX-Mentorship, will you provide me more information to get started?

RainbowMango:
I updated a document list to the issue description.

basit9958:
Hey @RainbowMango, I have read the proposal,documents and the struct  and believe we need to build a webhook and write a lua script which contains the logic to retain the configured field by the user to sync changes so the there is no endless loop in karmada control plane. After we need to register it here `cmd/webhook/app/webhook.go` . 
I am interested in solving this issue ,Get me on board for this one.

chaunceyjiang:
Friendly ping @jameszhangyukun Any progress？

RainbowMango:
Hi @chaunceyjiang 
@jameszhangyukun [made a demo](https://github.com/jameszhangyukun/karmada/pull/1) for this. 

The next step I think is `Resource Interpreter framework cache and reorganize configurations`.

XiShanYongYe-Chang:
Hi @jameszhangyukun, I'll be involved with you in the development of this work, thanks~

chaunceyjiang:
Hi, @jameszhangyukun @RainbowMango @XiShanYongYe-Chang  I am also very interested in this, can I participate in the development together?

XiShanYongYe-Chang:
> Hi, @jameszhangyukun @RainbowMango @XiShanYongYe-Chang I am also very interested in this, can I participate in the development together?

Of course, thank you very much.

RainbowMango:
Thank you @chaunceyjiang @ikaven1024 @jameszhangyukun @XiShanYongYe-Chang for all your effort in this feature.
It's a big success for all of us!

Most of the planned features have been done and will be delivered with the coming v1.4.0.
For the rest of the improvement task, let's do it in the next release.
/close

PR #2636, opened 2022-10-13
Take into account MetalLB in Resource Interpreter webhook example

### LINKED ISSUE #2614: Custom Resource Interpreter example README should be updated with LoadBalancer/MetalLB support
:wave: Hi :wave: 

**What happened**:

Currently if you follow the README on the Custom Resource Interpreter example you will get an error [at that point](https://github.com/karmada-io/karmada/blob/master/examples/README.md#step2-deploy-webhook-configuration-in-karmada-apiserver).

This is because the webhook url should be replaced but this is not done in `webhook-configuration.sh`:

https://github.com/karmada-io/karmada/blob/75a0305a8c2bbbb91823bd0d000331ca7db91b6d/examples/customresourceinterpreter/webhook-configuration.yaml#L14 

Based on `hack/pre-run-e2e.sh` I think that:

* MetalLB should be deployed (could be described in tutorial or through `hack/local-up-karmada.sh`).
* `webhook-configuration.sh` should replace the webhook IP address.

Follow the tutorial

loicalbertin:
**What type of PR is this?**

/kind bug

While required to run the RI webhook example, MetalLB is not installed by `hack/local-up-karmada.sh`.

So this PR adds a Prequisites section that describe how to install and configure MetalLB.

It also change a script to  automatically replace the webhook @IP in webhook configuration.

Fixes #2614

I did the minimum to make it work.
To do it I mainly copied corresponding parts of `hack/pre-run-e2e.sh`.

But I think you should consider installing and configuring MetalLB with `hack/local-up-karmada.sh` as it will reduce duplicated shell code / config stuff.

Also all the TLS & certificate generation is pretty complicated and hardcoded (see for instance the @IP of the webhook already reserved at certificate generation in `hack/deploy-karmada.sh` whereas it does not even deploy MetalLB).
Did you ever considered using [cert-manager](https://cert-manager.io/) to handle this?

```release-note
NONE
```

Signed-off-by: Albertin Loic <loic.albertin@atos.net>

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -6,7 +6,44 @@ This example implements a new CustomResourceDefinition(CRD), `Workload`, and cre
 
 ### Install
 
-> For karmada deployed using `hack/local-up-karmada.sh`, there are `karmada-host`, `karmada-apiserver` and three member clusters named `member1`, `member2` and `member3`.
+### Prerequisites
+
+For karmada deploy using `hack/local-up-karmada.sh`, there are `karmada-host`, `karmada-apiserver` and three member clusters named `member1`, `member2` and `member3`.
+
+Then you need to deploy MetalLB as a Load Balancer to expose the webhook.
+
+```bash
+kubectl --context="karmada-host" get configmap kube-proxy -n kube-system -o yaml | \
+  sed -e "s/strictARP: false/strictARP: true/" | \
+  kubectl --context="karmada-host" apply -n kube-system -f -
+
+curl https://raw.githubusercontent.com/metallb/metallb/v0.13.5/config/manifests/metallb-native.yaml -k | \
+  sed '0,/args:/s//args:\n        - --webhook-mode=disabled/' | \
+  sed '/apiVersion: admissionregistration/,$d' | \
+  kubectl --context="karmada-host" apply -f -
+
+export interpreter_webhook_example_service_external_ip_address=$(kubectl config view --template='{{range $_, $value := .clusters }}{{if eq $value.name "karmada-apiserver"}}{{$value.cluster.server}}{{end}}{{end}}' | \
+  awk -F/ '{print $3}' | \
+  sed 's/:.*//' | \
+  awk -F. '{printf "%s.%s.%s.8",$1,$2,$3}')
+
+cat <<EOF | kubectl --context="karmada-host" apply -f -
+apiVersion: metallb.io/v1beta1
+kind: IPAddressPool
+metadata:
+  name: metallb-config
+  namespace: metallb-system
+spec:
+  addresses:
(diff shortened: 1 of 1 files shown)


RainbowMango:
> But I think you should consider installing and configuring MetalLB with hack/local-up-karmada.sh as it will reduce duplicated shell code / config stuff.

Yes, that makes sense to me. Once we installed and configured MetalLB in `hack/local-up-karmada.sh`, it makes easy for people to experience the webhook thing. 

> Also all the TLS & certificate generation is pretty complicated and hardcoded (see for instance the @ip of the webhook already reserved at certificate generation in hack/deploy-karmada.sh whereas it does not even deploy MetalLB).
Did you ever considered using [cert-manager](https://cert-manager.io/) to handle this?

Hmm, I have the same feeling that the certificate generation is complicated. But I didn't really think about the cert-manager yet,  I'm glad to introduce it if it can simplify the certificate issue.

RainbowMango:
This patch LGTM. But, you might need to amend or squash your commits to deal with the `invalid commit messages` issue mentioned above.
> [Keywords](https://help.github.com/articles/closing-issues-using-keywords) which can automatically close issues and at(@) or hashtag(#) mentions are not allowed in commit messages.

RainbowMango:
By the way, @loicalbertin , we are going to introduce another resource interpreter that is tracked by #2371. We hope to provide a configurable interpreter for CRDs, and it won't rely on an exposed service(webhook) anymore. 
If everything goes well, this feature will be released at v1.4(by the end of November 2022) 
Any comments are welcome~~

RainbowMango (approved):
/lgtm
/approve
Thanks @loicalbertin

loicalbertin:
> By the way, @loicalbertin , we are going to introduce another resource interpreter that is tracked by #2371. We hope to provide a configurable interpreter for CRDs, and it won't rely on an exposed service(webhook) anymore. If everything goes well, this feature will be released at v1.4(by the end of November 2022) Any comments are welcome~~

Hey @RainbowMango, that looks very promising. I will have a deeper look on this soon.
I was actually thinking about something similar for handling replicas because from my understanding most of the `InterpretReplica` and `ReviseReplica` operations could be handled by a simple jsonpath expression and sounds a good candidate for a generic and configurable interpreter, do you agree?

RainbowMango:
> InterpretReplica and ReviseReplica operations could be handled by a simple jsonpath expression and sounds a good candidate for a generic and configurable interpreter, do you agree?

I guess jsonpath expression suitable for simple scenarios(like `InterpretReplica`/`ReviseReplica`), but for a little bit more complex   scenarios, like [`AggregateStatus` or `ReflectStatus`](https://github.com/karmada-io/karmada/blob/69ad816e57d6adf9b8612266068f663cf3c31c75/pkg/resourceinterpreter/interpreter.go#L37-L43), I doubt the jsonpath can help.

PR #2639, opened 2022-10-13
Add configurable resource interpreter API

RainbowMango:
**What type of PR is this?**
/kind feature
/kind api-change

This PR proposes the configurable interpreter API and takes the `retention` as a start.

Basically according to [this proposal](https://github.com/karmada-io/karmada/tree/master/docs/proposals/configurable-local-value-retention).

Part of #2371 
Fixes #

```release-note
NONE(Part feature)
```


pkg/apis/config/v1alpha1/resourceinterpretercustomization_types.go
```diff
@@ -0,0 +1,70 @@
...
+
+// +genclient
+// +genclient:nonNamespaced
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:resource:scope="Cluster"
+// +kubebuilder:storageversion
+
+// ResourceInterpreterCustomization describes the configuration of a specific
+// resource for Karmada to get the structure.
+// It has higher precedence than the default interpreter and the interpreter
```
chaunceyjiang, line 13: Can you explain it more clearly?

Why does `ResourceInterpreterCustomization` take precedence over `Resource Interpreter Webhook`?

RainbowMango: Sure, I can add more specific comment on it later.
For a quick answer, that is because the configurable interpreter would be more effective than webhook, that don't need to fire an HTTP request.

pkg/apis/config/v1alpha1/resourceinterpretercustomization_types.go
```diff
@@ -0,0 +1,70 @@
...
+
+	// Retention describes the rules about how to honor the changes on member clusters.
+	// This ensures that Karmada doesn't continually clear the generated fields
+	// from a resource, to avoid continual regeneration by the resource controller
+	// in the member cluster.
+	// For example, the replica of Deployment might be set by the HPA controller
+	// on member cluster, in this case, Karmada can retain the replicas with this
+	// configuration and doesn't try to correct it.
+	// +optional
+	Retention LocalValueRetention `json:"retention,omitempty"`
```
kevin-wangzefeng, line 42:
```suggestion
	// Retention describes the desired behavior that karmada should react on
	// the changes made by member cluster components. This avoids system
	// running into a meaningless loop that Karmada resource controller and
	// the member cluster component continually applying opposite values of a field. 
	// For example, the "replicas" of Deployment might be changed by the HPA
	// controller on member cluster. In this case, Karmada should retain the "replicas"
	// and not try to change it.
	// +optional
	Retention LocalValueRetention `json:"retention,omitempty"`
```

pkg/apis/config/v1alpha1/resourceinterpretercustomization_types.go
```diff
@@ -0,0 +1,70 @@
...
+	//
+	// function Retain(desiredObj, observedObj)
+	//     desiredObj.spec.fieldFoo = observedObj.spec.fieldFoo
+	//     return desiredObj
+	// end
+	//
+	// RetentionLua only holds the function body part, take the `desiredObj` and `observedObj` as the function
+	// parameters or global variables, and finally returns a retained specification.
+	// +required
+	RetentionLua string `json:"retentionLua"`
```
kevin-wangzefeng, line 59:
```suggestion
	// RetentionLua holds the Lua script that is used to retain runtime values to the desired specification.
	// The script should implement a function as follows:
	//     retentionLua: >
	//         function Retain(desiredObj, observedObj)
	//             desiredObj.spec.fieldFoo = observedObj.spec.fieldFoo
	//             return desiredObj
	//         end
	//
	// RetentionLua only holds the function body part, take the `desiredObj` and `observedObj` as the function
	// parameters or global variables, and finally returns a retained specification.
	// +required
	RetentionLua string `json:"retentionLua"`
```

Issue #2432, opened 2022-08-26
[umbrella]Karmada search iterated tasks


liys87x:
**What would you like to be added**:
We are going to implememt karmada search.

I'd like to list all the iterated tasks here, anyone wants to join or help on it, please feel free to leave your comment on it.

**Why is this needed**:

- [ ] User guide for karmada search
- [ ] Search api supports query through backend
- [ ] Implement resource registry status
- [ ] Provides an example of an opensearch based dashboard
- [ ] Add election logic to let the leader write to the backend storage

ikaven1024:
Proxy has been added in search. Shall we refactor search, fetch data from proxy cache, instead of informing from member clusters?

liys87x:
Yes, we should store unified. But cache is more complex than informer and I needs a more comprehensive understanding.

liys87x:
@ikaven1024 Currently cacheproxy is private, should it be exposed first?

ikaven1024:
> @ikaven1024 Currently cacheproxy is private, should it be exposed first?

You shall use this interface, not cacheProxy:

https://github.com/karmada-io/karmada/blob/2b7ed24875f3914a68b3bb618799fa6d1e4fba37/pkg/search/proxy/store/multi_cluster_cache.go#L32-L37

liys87x:
> 

How do I get its implementation, or rather, how do I get the MultiClusterCache

ikaven1024:
> How do I get its implementation, or rather, how do I get the MultiClusterCache

You can add a method in proxy controller to export it.

ikaven1024:
@liys87x proxy benchmark is done, what do you think about it? https://github.com/karmada-io/karmada/issues/2518#issue-1375349836

PR #2650, opened 2022-10-18
add option to disable search and proxy

ikaven1024:

**What type of PR is this?**
/kind cleanup

In `karmada-search` component, both `search` and `proxy` will cache the sub cluster's resource in memory, it will take much physical resource. While not both features are used for users. So I add the options to enable/disable these.

Fixes #

This shall be merged into release branches.

```release-note
`karmada-search`: users now can use `--disable-search` and `--disable-proxy` options to disable search and proxy feature (default both are enabled).
```

### DIFF
--- a/cmd/karmada-search/app/options/options.go
+++ b/cmd/karmada-search/app/options/options.go
@@ -47,6 +47,9 @@ type Options struct {
 	KubeAPIBurst int
 
 	ProfileOpts profileflag.Options
+
+	DisableSearch bool
+	DisableProxy  bool
 }
 
 // NewOptions returns a new Options.
@@ -69,6 +72,8 @@ func (o *Options) AddFlags(flags *pflag.FlagSet) {
 
 	flags.Float32Var(&o.KubeAPIQPS, "kube-api-qps", 40.0, "QPS to use while talking with karmada-apiserver. Doesn't cover events and node heartbeat apis which rate limiting is controlled by a different set of flags.")
 	flags.IntVar(&o.KubeAPIBurst, "kube-api-burst", 60, "Burst to use while talking with karmada-apiserver. Doesn't cover events and node heartbeat apis which rate limiting is controlled by a different set of flags.")
+	flags.BoolVar(&o.DisableSearch, "disable-search", false, "Disable search feature that would save memory usage significantly.")
+	flags.BoolVar(&o.DisableProxy, "disable-proxy", false, "Disable proxy feature that would save memory usage significantly.")
 
 	utilfeature.DefaultMutableFeatureGate.AddFlag(flags)
 	o.ProfileOpts.AddFlags(flags)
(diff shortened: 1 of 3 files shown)


RainbowMango:
Seems you want a switch to control if `search`/`proxy` is enabled or not. 
I'm afraid the feature gate is inappropriate as a feature gate used to set up a switch that is mainly for experiment features, after the feature becomes mature, the relevant feature gate will be removed eventually. 

We can consider doing it by providing a command flag.

ikaven1024:
@liys87x @XiShanYongYe-Chang @RainbowMango please take a look.

RainbowMango:
/assign
note: @XiShanYongYe-Chang is on vocation and will get back next Monday.

RainbowMango:
@ikaven1024 Please update the PR title, descriptions, and release note, since the current approach doesn't feature gate anymore.

RainbowMango:
> In karmada-search component, both search and proxy will cache the sub cluster's resource in memory, it will take much physical resource. 

Is there a possibility that combining the two caches together?

ikaven1024:
> Is there a possibility that combining the two caches together?

It's not easy to do this. @liys87x is studying it. https://github.com/karmada-io/karmada/issues/2432#issuecomment-1229853832.

Before this realized, we have to add options to disable them.

cmd/karmada-search/app/options/options.go
```diff
@@ -69,6 +72,8 @@ func (o *Options) AddFlags(flags *pflag.FlagSet) {
 
 	flags.Float32Var(&o.KubeAPIQPS, "kube-api-qps", 40.0, "QPS to use while talking with karmada-apiserver. Doesn't cover events and node heartbeat apis which rate limiting is controlled by a different set of flags.")
 	flags.IntVar(&o.KubeAPIBurst, "kube-api-burst", 60, "Burst to use while talking with karmada-apiserver. Doesn't cover events and node heartbeat apis which rate limiting is controlled by a different set of flags.")
+	flags.BoolVar(&o.DisableSearch, "disable-search", o.DisableSearch, "turn off the search feature.")
+	flags.BoolVar(&o.DisableProxy, "disable-proxy", o.DisableProxy, "turn off the proxy feature.")
```
RainbowMango, line 76:
```suggestion
	flags.BoolVar(&o.DisableSearch, "disable-search", false, "Disable search feature that would save memory usage significantly.")
	flags.BoolVar(&o.DisableProxy, "disable-proxy", false, "Disable proxy feature that would save memory usage significantly.")
```
Just give my comments here, if we can't combine the two flags together for now.

By the way, what if both flags are `true`? What's the behavior?

ikaven1024: > By the way, what if both flags are `true`? What's the behavior?

Both search and proxy are working, they have two independent cache. For some large scale cluster, `karmada-search` may OOM.

pkg/search/controller.go
```diff
@@ -115,7 +106,6 @@ func (c *Controller) Start(stopCh <-chan struct{}) {
 
 	defer runtime.HandleCrash()
 
-	c.informerFactory.Start(stopCh)
```
XiShanYongYe-Chang, line 118: Just a question: why do we need to delete this line?

ikaven1024: karmadaInformerFactory is shared between searchController and proxyController. It's start is moved to
https://github.com/karmada-io/karmada/blob/b7554264ed07be3da20646d6e822f651415814dc/cmd/karmada-search/app/options/options.go#L108-L110

XiShanYongYe-Chang: ok, thanks for your explain~
/lgtm

ikaven1024:
@RainbowMango shall we move forward it?

RainbowMango:
Sure. I thought @XiShanYongYe-Chang has approved this.
/approve

PR #2652, opened 2022-10-18
Unit Test worker.go

cmicat:

**What type of PR is this?**

/kind cleanup

Add Unit Test for worker.go

```release-note
NONE
```

### DIFF
--- a/pkg/util/worker_test.go
+++ b/pkg/util/worker_test.go
@@ -1,0 +1,211 @@
+package util
+
+import (
+	"errors"
+	"fmt"
+	"sync"
+	"testing"
+	"time"
+
+	"go.uber.org/atomic"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/karmada-io/karmada/pkg/sharedcli/ratelimiterflag"
+)
+
+func newTestAsyncWorker(reconcileFunc ReconcileFunc) *asyncWorker {
+	options := Options{
+		Name:          "test_async_worker",
+		KeyFunc:       MetaNamespaceKeyFunc,
+		ReconcileFunc: reconcileFunc,
+		RateLimiterOptions: ratelimiterflag.Options{
+			RateLimiterBaseDelay:  time.Millisecond,
+			RateLimiterMaxDelay:   10 * time.Millisecond,
+			RateLimiterQPS:        5000,
+			RateLimiterBucketSize: 100,
+		},
+	}
+	worker := NewAsyncWorker(options)
+
+	return worker.(*asyncWorker)
+}
+
+func Test_asyncWorker_Enqueue(t *testing.T) {
+	const name = "fake_node"
+
+	worker := newTestAsyncWorker(nil)
+
+	node := &corev1.Node{
+		ObjectMeta: metav1.ObjectMeta{Name: name},
+	}
+
+	worker.Enqueue(node)
+
+	item, _ := worker.queue.Get()
+
+	if name != item {
+		t.Errorf("Added Item: %v, want: %v", item, name)
+	}
+}
+
+func Test_asyncWorker_AddAfter(t *testing.T) {
+	const name = "fake_node"
+	const duration = 1 * time.Second
+
+	worker := newTestAsyncWorker(nil)
+
+	start := time.Now()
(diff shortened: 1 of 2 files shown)


pkg/util/worker_test.go
```diff
@@ -0,0 +1,174 @@
...
+	stopChan := make(chan struct{})
+	worker.Run(5, stopChan)
+
+	for i := 0; i < cnt; i++ {
+		worker.Add(i)
+	}
+
+	time.Sleep(time.Second)
+
+	close(stopChan)
```
jwcesign, line 127: defer close(stopChan)

pkg/util/worker_test.go
```diff
@@ -0,0 +1,174 @@
...
+	processed := reconcile.getProcessed()
+	if len(processed) < cnt {
+		t.Errorf("processed item not equal to input, len() = %v, processed item is %v", len(processed), processed)
+		return
+	}
+
+	sort.Ints(processed)
+
+	for i := 0; i < cnt; i++ {
+		if processed[i] != i {
```
jwcesign, line 138: Map is better?  use array is little tricky

jwcesign:
other is LGTM

pkg/util/worker_test.go
cmicat: I intend to stop the worker after 1s. defer will close it on method exit.

pkg/util/worker_test.go
cmicat: OK. I can reuse `asyncWorkerReconciler.shouldPass` map.

pkg/util/worker_test.go
```diff
@@ -0,0 +1,171 @@
...
+	reconcile := newAsyncWorkerReconciler()
+	worker := newTestAsyncWorker(reconcile.ReconcileFunc)
+	stopChan := make(chan struct{})
+	worker.Run(5, stopChan)
+
+	for i := 0; i < cnt; i++ {
+		worker.Add(i)
+	}
+
+	time.Sleep(time.Second)
```
RainbowMango, line 124: We can wait for `len(processed) == cnt` for a period of time, we don't know if 1s is enough or not, you know the test will be launched on various environments.

pkg/util/worker_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+	close(stopChan)
+
+	processed := reconcile.ProcessedItem()
+	if len(processed) < cnt {
+		t.Errorf("processed item not equal to input, len() = %v, processed item is %v", len(processed), processed)
```
chaunceyjiang, line 130:
``` diff
        if len(processed) < cnt {
-               t.Errorf("processed item not equal to input, len() = %v, processed item is %v", len(processed), processed)
-               return
+               t.Fatalf("processed item not equal to input, len() = %v, processed item is %v", len(processed), processed)
        }

```

pkg/util/worker_test.go
```diff
@@ -0,0 +1,171 @@
...
+		return
+	}
+
+	for i := 0; i < cnt; i++ {
+		if _, ok := processed[i]; !ok {
+			t.Errorf("expected item not processed, expected: %v, all processed item: %v", i, processed)
```
chaunceyjiang, line 136: ditto.

pkg/util/worker_test.go
cmicat: I already set ratelimiter QPS & delay, so ratelimiter will not delay the execution. I think 1s is enough for most such short code to finish.

cmicat: @RainbowMango

cmicat: I think I can write a polling logic: check the status periodically, fail after certain seconds. For example, check every 50 ms, fail after 10s. This way, bigger wait duration won't result in bigger test execution time.

cmicat: > I think I can write a polling logic: check the status periodically, fail after certain seconds. For example, check every 50 ms, fail after 10s. This way, bigger wait duration won't result in bigger test execution time.

Done

cmicat:
@RainbowMango Please take a look again

cmicat:
@jwcesign @chaunceyjiang 
The problems you mentioned is all fixed.

RainbowMango (approved):
/approve

PR #2654, opened 2022-10-18
UT: import gomock to make write ut easier

jwcesign:

**What type of PR is this?**

/kind cleanup

Give a demo how to use gomock to make write ut easier

```release-note
none
```


hack/update-mocks.sh
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env bash
+
```
chaunceyjiang, line 2: image

chaunceyjiang: In my local environment.

hack/update-mocks.sh
```diff
@@ -0,0 +1,52 @@
...
+        -o -wholename '*/bindata.go' \
+      \) -prune \
+    \) -name '*.go'
+}
+
+cd "${KARMADA_ROOT}"
+
+echo 'executing go generate command on below files'
+for IFILE in $(find_files | xargs grep --files-with-matches -e '//go:generate mockgen'); do
+    go generate -v "$IFILE"
```
chaunceyjiang, line 51: image


Can we check the `mockgen` command in this script?

hack/update-mocks.sh
jwcesign: Ops, I will chmod it

hack/update-mocks.sh
jwcesign: Sure, it's better idea.

jwcesign: Fix, try again

chaunceyjiang:
lint failed.

https://github.com/karmada-io/karmada/actions/runs/3287561163/jobs/5416867607

jwcesign:
> lint failed.
> 
> https://github.com/karmada-io/karmada/actions/runs/3287561163/jobs/5416867607

If this idea is received, should change link check to ignore mock gen  files.

hack/update-mocks.sh
```diff
@@ -0,0 +1,57 @@
...
+
+KARMADA_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
+# Explicitly opt into go modules, even though we're inside a GOPATH directory
+export GO111MODULE=on
+export PATH=$PATH:$(go env GOPATH)/bin
+
+trap EXIT
+
+echo 'installing mockgen'
+go install github.com/golang/mock/mockgen
```
chaunceyjiang, line 32:
```suggestion
source "${KARMADA_ROOT}"/hack/util.sh
echo -n "Preparing: 'mockgen' existence check - "
if util::cmd_exist mockgen; then
  echo "passed"
else
  echo "not pass"
  util::install_tools "github.com/golang/mock/mockgen" "v1.6.0"
fi
```

chaunceyjiang:
> If this idea is received, should change link check to ignore mock gen files.

Get it.

pkg/scheduler/framework/runtime/framework_test.go
```diff
@@ -0,0 +1,147 @@
...
+			result := frameWork.RunFilterPlugins(ctx, nil, nil, nil)
+			if result.IsSuccess() != tt.isSuccess {
+				t.Errorf("want %v, but get:%v", tt.isSuccess, result.IsSuccess())
+			}
+		})
+	}
+}
+
+func createAndRegisterFactory(plugin framework.Plugin) (Registry, error) {
+	filterPluginFactory := PluginFactory(func() (framework.Plugin, error) {
```
RainbowMango, line 66: Redundant type conversion

RainbowMango: Maybe we don't need to abstract the function.

jwcesign:
cc @RainbowMango PTAL

jwcesign:
> > If this idea is received, should change link check to ignore mock gen files.
> 
> Get it.

I change the link check tools, please take a loook, @chaunceyjiang

hack/tools/preferredimports/preferredimports.go
```diff
@@ -42,12 +42,13 @@ import (
...
-	importAliases = flag.String("import-aliases", "hack/.import-aliases", "json file with import aliases")
-	confirm       = flag.Bool("confirm", false, "update file with the preferred aliases for imports")
-	regex         = flag.String("include-path", "(test/e2e/|test/e2e_node)", "only files with paths matching this regex is touched")
-	isTerminal    = term.IsTerminal(int(os.Stdout.Fd()))
-	logPrefix     = ""
-	aliases       map[string]string
+	importAliases    = flag.String("import-aliases", "hack/.import-aliases", "json file with import aliases")
+	confirm          = flag.Bool("confirm", false, "update file with the preferred aliases for imports")
+	includePathRegex = flag.String("include-path", "(test/e2e/|test/e2e_node)", "only files with paths matching this regex is touched")
+	excludePathRegex = flag.String("exclude-path", "(testing)", "exclude path for check alias")
```
RainbowMango, line 48:
```suggestion
	excludePathRegex = flag.String("exclude-path", "(testing)", "files with paths matching this regex is ignored")
```

hack/update-mocks.sh
```diff
@@ -0,0 +1,60 @@
...
+export GO111MODULE=on
+export PATH=$PATH:$(go env GOPATH)/bin
+
+trap EXIT
+
+echo 'installing mockgen'
+source "${KARMADA_ROOT}"/hack/util.sh
+echo -n "Preparing: 'mockgen' existence check - "
+if [ ! $(util::cmd_exist mockgen) ]; then
+  util::install_tools "github.com/golang/mock/mockgen" "latest"
```
RainbowMango, line 35: We can't install `latest`, the generated file may change over `mockgen` versions.
```suggestion
  GO111MODULE=on go install "github.com/golang/mock/mockgen"
```
This way installing `mockgen` with the version in the vendor. For CI we need a consistent version.

RainbowMango (commented):
Generally looks good to me.

hack/update-mocks.sh
chaunceyjiang: +1

pkg/scheduler/framework/runtime/framework_test.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			frameWork, err := NewFramework(registry)
+			if err != nil {
+				t.Errorf("create frame work error:%v", err)
+			}
+
+			tt.mockFunc()
```
cmicat, line 140: I think every test case should be independent. So it's better not share mock object between test cases. Instead, we can create a new mock object in every test case.

pkg/scheduler/framework/runtime/framework_test.go
```diff
@@ -0,0 +1,147 @@
...
+			},
+			isSuccess: true,
+		},
+		{
+			name: "Test filter func error",
+			mockFunc: func() {
+				mockFilterPlugin.EXPECT().Filter(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).
+					Return(framework.NewResult(framework.Error, "foo error"))
+			},
+			isSuccess: false,
```
ikaven1024, line 40: It only test two cases: with one error or error plugin. Other cases are not covered: no plugin, mix plugins.

How about these:
```go
func Test_frameworkImpl_RunFilterPlugins(t *testing.T) {
	ctx := context.Background()
	mockCtrl := gomock.NewController(t)
	defer mockCtrl.Finish()

	alwaysError := frameworktesting.NewMockFilterPlugin(mockCtrl)
	alwaysError.EXPECT().Filter(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Return(framework.NewResult(framework.Error))

	alwaysSuccess := frameworktesting.NewMockFilterPlugin(mockCtrl)
	alwaysSuccess.EXPECT().Filter(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Return(framework.NewResult(framework.Success))

	tests := []struct {
		name      string
		framework framework.Framework
		isSuccess bool
	}{
		{
			name: "no plugin",
			framework: &frameworkImpl{
				filterPlugins: []framework.FilterPlugin{},
			},
			isSuccess: true,
		},
		{
			name: "error plugin",
			framework: &frameworkImpl{
				filterPlugins: []framework.FilterPlugin{alwaysError},
			},
			isSuccess: false,
		},
		{
			name: "success plugin",
			framework: &frameworkImpl{
				filterPlugins: []framework.FilterPlugin{alwaysSuccess},
			},
			isSuccess: true,
		},
		{
			name: "error and success plugins",
			framework: &frameworkImpl{
				filterPlugins: []framework.FilterPlugin{alwaysError, alwaysSuccess},
			},
			isSuccess: false,
		},
		{
			name: "success and error plugins",
			framework: &frameworkImpl{
				filterPlugins: []framework.FilterPlugin{alwaysSuccess, alwaysError},
			},
			isSuccess: true,
		},
	}

	for _, tt := range tests {
			result := tt.framework.RunFilterPlugins(ctx, nil, nil, nil)
			if result.IsSuccess() != tt.isSuccess {
				t.Errorf("want %v, but get:%v", tt.isSuccess, result.IsSuccess())
			}
		})
	}
}
```

hack/update-mocks.sh
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env bash
```
ikaven1024, line 1: please add `verify-mocks.sh`, add add these files to `update-all.sh` and `verify-all.sh`.

pkg/scheduler/framework/runtime/framework_test.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+}
+
+func createAndRegisterFactory(plugins []framework.Plugin) (Registry, error) {
+	registry := Registry{}
+	for _, plugin := range plugins {
+		filterPluginFactory := func() (framework.Plugin, error) {
+			return plugin, nil
+		}
+		if err := registry.Register("foo", filterPluginFactory); err != nil {
```
ikaven1024, line 84: we can not register plugins with the same name, they are override.

hack/verify-all.sh
```diff
@@ -19,3 +19,4 @@ bash "$REPO_ROOT/hack/verify-vendor.sh"
 bash "$REPO_ROOT/hack/verify-swagger-docs.sh"
 bash "$REPO_ROOT/hack/verify-crdgen.sh"
 bash "$REPO_ROOT/hack/verify-codegen.sh"
+bash "$REPO_ROOT/hack/verify-mocks.sh"
```
ikaven1024, line 22: place the script in order with:
https://github.com/karmada-io/karmada/blob/5a1addf6936d7c6604fb26c4ae725ce7c716db3a/hack/verify-all.sh#L12-L14

pkg/scheduler/framework/runtime/framework_test.go
jwcesign: done for filter plugin, but not for score plugin, it does not increase code coverage and get few benefits.

pkg/scheduler/framework/runtime/framework_test.go
jwcesign: Every time createAndRegisterFactory is called, the registry is created new, so there is no override.

pkg/scheduler/framework/runtime/framework_test.go
jwcesign: It make sense, but I prefer to keep it as case global, it makes code looks more simple

jwcesign:
cc @RainbowMango @ikaven1024 @chaunceyjiang @cmicat PTAL

pkg/scheduler/framework/runtime/framework_test.go
ikaven1024: mock is polluted between cases, It doesn't seem a good thing.

pkg/scheduler/framework/runtime/framework_test.go
```diff
@@ -0,0 +1,170 @@
...
+	for _, plugin := range plugins {
+		pluginCopy := plugin
+		filterPluginFactory := func() (framework.Plugin, error) {
+			return pluginCopy, nil
+		}
+		if err := registry.Register("foo"+strconv.Itoa(nameIndex), filterPluginFactory); err != nil {
+			return nil, err
+		}
+		nameIndex++
+	}
```
ikaven1024, line 95:
```suggestion
	for i, plugin := range plugins {
		pluginCopy := plugin
		filterPluginFactory := func() (framework.Plugin, error) {
			return pluginCopy, nil
		}
		if err := registry.Register("foo"+strconv.Itoa(i), filterPluginFactory); err != nil {
			return nil, err
		}
	}
```

jwcesign:
cc @ikaven1024 , PTAL

RainbowMango (approved):
/approve

RainbowMango:
We shouldn't install `mockgen` with a version suffix, because we have to maintain to the version in two places(go.mod and the script):
https://github.com/karmada-io/karmada/blob/4521a0ddd2d46dcbfdeca1d172467faf6736dcf3/hack/update-mocks.sh#L35
Just like I mentioned above: https://github.com/karmada-io/karmada/pull/2654#discussion_r1001603344

(It's too late to hold this, I'll send another PR for this.)

PR #2679, opened 2022-10-24
Use kwok in e2e instead of kind

### LINKED ISSUE #2661: Use Kwok instead of Kind without actually running a container

https://github.com/kubernetes-sigs/kwok

1. Use Kwok instead of Kind with the non-pull mode clusters in the example
2. ~~Update the document to use Kwok instead of [fake-kubelet] (https://github.com/wzshiming/fake-kubelet)
    https://karmada.io/docs/next/developers/performance-test-setup-for-karmada~~

**Why is this needed**:

Reduce CI resource requirements

wzshiming:
**What type of PR is this?**

/kind feature

Fixes #2661

```release-note
NONE
```


hack/local-up-karmada.sh
```diff
@@ -48,6 +49,16 @@ else
   echo "not pass"
   util::install_tools "sigs.k8s.io/kind" $kind_version
 fi
+
+echo -n "Preparing: 'kwokctl' existence check - "
+if util::cmd_exist kwokctl; then
+  echo "passed"
+else
+  echo "not pass"
+  wget -O /usr/local/bin/kwokctl "https://github.com/kubernetes-sigs/kwok/releases/download/v0.0.1/kwokctl-$(go env GOOS)-$(go env GOARCH)" && \
```
ikaven1024, line 58: `/usr/local/bin` directory is not always permitted. Some guys run without `root`

hack/local-up-karmada.sh
```diff
@@ -48,6 +49,16 @@ else
   echo "not pass"
   util::install_tools "sigs.k8s.io/kind" $kind_version
 fi
+
+echo -n "Preparing: 'kwokctl' existence check - "
+if util::cmd_exist kwokctl; then
+  echo "passed"
+else
+  echo "not pass"
```
ikaven1024, line 57: `no pass` means script run fail and will exit. Change to `kwokctl  is not found, and will install it...`

wzshiming: Updated, But everything else seems to be echo 'not pass'

wzshiming:
/cc @Poor12
/cc @RainbowMango
/cc @ikaven1024

The PR is almost ready, and just by replacing Kind, the average test time of e2e dropped by about 10 minutes, which could be improved even more if we continue to optimize the logic of e2e.

hack/local-up-karmada.sh
```diff
@@ -51,6 +51,19 @@ fi
...
+# install kwok
+kwok_version=v0.1.0
+echo -n "Preparing: 'kwokctl' existence check - "
+if util::cmd_exist kwokctl; then
+  echo "passed"
+else
+  echo "not pass, kwokctl is not found, and will install it..."
+  wget -O "${BS_PATH}/bin/kwokctl" "https://github.com/kubernetes-sigs/kwok/releases/download/${kwok_version}/kwokctl-${BS_OS}-${BS_ARCH}" && \
+    chmod +x "${BS_PATH}/bin/kwokctl"
+fi
```
ikaven1024, line 65: It seems `kwok` is only used in e2e test. Shall we move this check to e2e.sh? Most guys use this script just setup a karmada environment, `kwok` is not requisite.

test/e2e/suite_test.go
```diff
@@ -169,6 +174,7 @@ func cleanupTestNamespace(namespace string, kubeClient kubernetes.Interface) err
 	return nil
 }
 
+//nolint:unused
```
ikaven1024, line 177: shall we remove the unused code?

test/e2e/suite_test.go
```diff
@@ -202,10 +208,144 @@ func createCluster(clusterName, kubeConfigPath, controlPlane, clusterContext str
...
+	err = os.MkdirAll(filepath.Dir(kubeConfigPath), 0755)
+	if err != nil {
+		return "", err
+	}
+
+	config := buf.Bytes()
+
+	// For Docker running on Linux used 172.17.0.1 which is the Docker-host in Docker’s default-network.
+	ip := "172.17.0.1"
+	if runtime.GOOS != "linux" {
```
ikaven1024, line 309: In line 258, we set the `binary` runtime for linux,  I am confused why use docker here?

test/e2e/suite_test.go
wzshiming: Too much code needed to be removed and I committed it in a new commit.

test/e2e/suite_test.go
wzshiming: This is actually just to get the ip of the host accessing from the container, now it's really easy to misunderstand, I'll change it.

test/e2e/suite_test.go
```diff
@@ -169,41 +169,132 @@ func cleanupTestNamespace(namespace string, kubeClient kubernetes.Interface) err
...
+    machineID: ""
+    operatingSystem: linux
+    osImage: ""
+    systemUUID: ""
+  phase: Running
+`
+
+	kwokRuntime := "docker"
+	hostAddress := "host.docker.internal" // Require Docker 18.03
+	if runtime.GOOS == "linux" {
```
ikaven1024, line 214: Could you explain why use binary runtime only on linux?

wzshiming: https://www.downloadkubernetes.com/

Kubernetes only provides binary files for Linux.

RainbowMango:
@wzshiming I appreciate all your efforts and I am very impressed with kwork's ability. This PR proved that kwok could be a perfect option to run specific tests. But for Karmada E2E testing, we need to run containers in some tests, so I suggest postponing this replacement work.

In addition, I believe kwok will be useful for Karmada's performance benchmark testing. I'll talk to Karmada folks to see if we can run the performance testing again this year. Last year, we run the performance testing with 100 kind clusters, I hope we can run against more kwok clusters, like 500 or 1000.

PR #2681, opened 2022-10-25
Add configurable interpreter api

RainbowMango:
**What type of PR is this?**
/kind feature
/kind api-change

This PR proposes the new API of the `Configurable Interpreter`.

Part of #2371

```release-note
NONE
```


pkg/apis/config/v1alpha1/resourceinterpretercustomization_types.go
```diff
@@ -40,23 +42,195 @@ type ResourceInterpreterCustomizationSpec struct {
...
+
+	// ReplicaResource describes the rules for Karmada to discover the resource's
+	// replica as well as resource requirements.
+	// It would be useful for those CRD resources that declare workload types like
+	// Deployment.
+	// It is usually not needed for Kubernetes native resources(Deployment, Job) as
+	// Karmada knows how to discover info from them. But if it is set, the default
+	// discovery rule will be ignored.
+	// +optional
+	ReplicaResource ReplicaResourceRequirement `json:"replicaResource,omitempty"`
```
chaunceyjiang, line 54: Just a question. Why not use struct pointer?

RainbowMango: Right. Good point. optional filed should use pointer.

pkg/apis/config/v1alpha1/resourceinterpretercustomization_types.go
```diff
@@ -40,23 +42,195 @@ type ResourceInterpreterCustomizationSpec struct {
...
+	// It is usually not needed for Kubernetes native resources(Deployment, Job) as
+	// Karmada knows how to revise replicas for them. But if it is set, the default
+	// revision rule will be ignored.
+	// +optional
+	ReplicaRevision ReplicaRevision `json:"replicaRevision,omitempty"`
+
+	// StatusReflection describes the rules for Karmada to pick the resource's status.
+	// Karmada provides built-in rules for several standard Kubernetes types, see:
+	// https://karmada.io/docs/userguide/globalview/customizing-resource-interpreter/#interpretstatus
+	// If StatusReflection is set, the default rule will be ignored.
```
chaunceyjiang, line 68:
```suggestion
	// If StatusReflection is set, the build-in rule will be ignored.
```

chaunceyjiang:
I'm mainly worried that the lua code is too long, which makes the yaml file of the `ResourceInterpreterCustomization` CR resource unreadable.

Can you share the yaml  file of the `ResourceInterpreterCustomization ` CR resource  ?

pkg/apis/config/v1alpha1/resourceinterpretercustomization_types.go
RainbowMango: done.

pkg/apis/config/v1alpha1/resourceinterpretercustomization_types.go
```diff
@@ -31,32 +33,204 @@ type ResourceInterpreterCustomizationSpec struct {
 	// +required
 	Kind string `json:"kind"`
```
ikaven1024, line 34: I don't think put `APIVersion` and `Kind` in this struct is a good thing, for the reasons:
1. `APIVersion` and `Kind` are not customization rulers
2. The spec will be very large, and users will be difficult to find the these two field. These fields are important for users.

I prefer the format:
```yaml
apiVersion: config.karmada.io/v1alpah1
kind: ResourceInterpreterCustomization
metadata:
  name: foo
target:
  apiVersion: v1
  kind: Pod
customizations:
  ...
```

pkg/apis/config/v1alpha1/resourceinterpretercustomization_types.go
```diff
@@ -31,32 +33,204 @@ type ResourceInterpreterCustomizationSpec struct {
...
+	//     luaScript: >
+	//         function GetReplicas(desiredObj)
+	//             nodeClaim = {}
+	//             resourceRequest = {}
+	//             result = {}
+	//
+	//             result.replica = desiredObj.spec.replicas
+	//             result.resourceRequest = desiredObj.spec.template.spec.containers[0].resources.limits
+	//
+	//			   nodeClaim.nodeSelector = desiredObj.spec.template.spec.nodeSelector
```
ikaven1024, line 126:
```suggestion
	//             nodeClaim.nodeSelector = desiredObj.spec.template.spec.nodeSelector
```

pkg/apis/config/v1alpha1/resourceinterpretercustomization_types.go
chaunceyjiang: Good idea.

ikaven1024:
> I'm mainly worried that the lua code is too long, which makes the yaml file of the `ResourceInterpreterCustomization` CR resource unreadable.

I don't think it s a big matter. People can read and edit it in IDE.
`CustomResourceDe` is always verify large :)

I either think defining the rules of one resource in different ResourceInterpreterCustomization has little benefit, but introducing too much matters: rule conflicting, unreadable for user... @RainbowMango

XiShanYongYe-Chang:
> I either think defining the rules of one resource in different ResourceInterpreterCustomization has little benefit, but introducing too much matters: rule conflicting, unreadable for user...

+1

pkg/apis/config/v1alpha1/resourceinterpretercustomization_types.go
RainbowMango: Good point. But I'm afraid the `target` might be present always after `customizations`(`Alphabetical order`).
Is that OK? Do we have other alternatives for `target`?
```yaml
customizations:
target:
  apiVersion: v1
  kind: Pod
```

RainbowMango: accept.

RainbowMango:
@ikaven1024 @chaunceyjiang Please take look.

pkg/apis/config/v1alpha1/resourceinterpretercustomization_types.go
```diff
@@ -23,40 +25,226 @@ type ResourceInterpreterCustomization struct {
...
+	LuaScript string `json:"luaScript"`
+}
+
+// ReplicaResourceRequirement holds the scripts for getting the desired replicas
+// as well as the resource requirement of each replica.
+type ReplicaResourceRequirement struct {
+	// LuaScript holds the Lua script that is used to discover the resource's
+	// replica as well as resource requirements
+	// The script should implement a function as follows:
+	//     luaScript: >
```
XiShanYongYe-Chang, line 131: This line is necessary? It seems redundant:
```yaml
...
  retention:
    luaScript:
      luaScript: >
       ...
```

RainbowMango:
```yaml
  retention:
    luaScript: >
      desiredObj.spec.fieldFoo = observedObj.spec.fieldFoo
      return desiredObj
```
This is expected. 
`LuaScript only holds the function body part`

ikaven1024:
Just a question: If I add a script with bad grammar, where and when this mistake will be reported? Do it in apiserver validating, or just throw error in runtime. And how to report these errors, in `ResourceInterpreterCustomization.Status`, or in events?

pkg/apis/config/v1alpha1/resourceinterpretercustomization_types.go
XiShanYongYe-Chang: Got it.

RainbowMango:
Good question. 

> If I add a script with bad grammar

I don't if there is a kind of tool that can verify the grammar issue. If so we can validate it by admission webhook.
@jameszhangyukun any comments?

I'm thinking to introduce a tool to help user to verify/testing their customizations, like:
```
karmadactl interpret --type=retention --desired=desired.yaml --observed=observed.yaml --customization=<ResourceInterpreterCustomization> configuration
```
People can test their scripts before applying.

XiShanYongYe-Chang:
> I'm thinking to introduce a tool to help user to verify/testing their customizations,

If this works, I think it's going to be a new and meaningful job.

ikaven1024:
LGTM

chaunceyjiang:
https://argo-cd.readthedocs.io/en/stable/operator-manual/health/#way-2-contribute-a-custom-health-check

Like argo-cd, we can also provide a directory to customize lua scripts for well-known projects.

RainbowMango:
> Like argo-cd, we can also provide a directory to customize lua scripts for well-known projects.

+1

chaunceyjiang:
> I either think defining the rules of one resource in different ResourceInterpreterCustomization has little benefit, but introducing too much matters: rule conflicting, unreadable for user...

Do we need to introduce an implicit priority for this?

chaunceyjiang:
As @ikaven1024  said, if the user applies the following two yaml files, which one will work?

```yaml
apiVersion: config.karmada.io/v1alpah1
kind: ResourceInterpreterCustomization
metadata:
  name: foo
target:
  apiVersion: v1
  kind: Pod
customizations:
  retention: 
    luaScript: >
        function Retain(desiredObj, observedObj)
            desiredObj.spec.fieldFoo = observedObj.spec.fieldFoo
            return desiredObj
        end
  replicaResource:
   ...
````


```yaml
apiVersion: config.karmada.io/v1alpah1
kind: ResourceInterpreterCustomization
metadata:
  name: bar
target:
  apiVersion: v1
  kind: Pod
customizations:
  retention: 
    luaScript: >
        function Retain(desiredObj, observedObj)
            desiredObj.spec.fieldFoo = desiredObj.spec.fieldFoo
            return desiredObj
        end
  replicaResource:
   ...
````

File: pkg/apis/config/v1alpha1/resourceinterpretercustomization_types.go [hunk 6]
```diff
@@ -23,40 +25,226 @@ type ResourceInterpreterCustomization struct {
...
+type StatusReflection struct {
+	// LuaScript holds the Lua script that is used to get the status from the observed specification.
+	// The script should implement a function as follows:
+	//     luaScript: >
+	//         function ReflectStatus(observedObj)
+	//             status = {}
+	//             status.readyReplicas = observedObj.status.observedObj
+	//             return status
+	//         end
+	//
```
chaunceyjiang, line 179: Example A
``` lua
luaScript: >
 function ReflectStatus(observedObj)
     status = {}
     status.readyReplicas = observedObj.status.observedObj
     return status
 end
```
Example B
``` lua
luaScript: >
     status = {}
     status.readyReplicas = observedObj.status.observedObj
     return status

```

Which one is right?

pkg/apis/config/v1alpha1/resourceinterpretercustomization_types.go
```diff
@@ -23,40 +25,226 @@ type ResourceInterpreterCustomization struct {
...
+	// LuaScript only holds the function body part, take the `observedObj` as the function
```
chaunceyjiang, line 180: Example A
``` lua
luaScript: >
 function ReflectStatus(observedObj)
     status = {}
     status.readyReplicas = observedObj.status.observedObj
     return status
 end
```
Example B
``` lua
luaScript: >
   status = {}
   status.readyReplicas = observedObj.status.observedObj
   return status

```

Which one is right?

RainbowMango: `Example` B is expected because it has been proved by @jameszhangyukun's demo.

But I'd prefer `Example A` which is easier to understand and straightforward.  @jameszhangyukun Can we load a whole function to Lua runtime instead of a piece of code?

RainbowMango:
> As @ikaven1024 said, if the user applies the following two yaml files, which one will work?

It's obviously an ambiguous situation. I think we can add an admission webhook to deny the second configuration.

pkg/apis/config/v1alpha1/resourceinterpretercustomization_types.go
XiShanYongYe-Chang: I think is Example B.

ikaven1024: > But I'd prefer `Example A` which is easier to understand and straightforward.

+1

chaunceyjiang: Maybe we can refer to nginx lua. 

https://github.com/openresty/lua-nginx-module#data-sharing-within-an-nginx-worker

RainbowMango: Can you make a demo with it?

RainbowMango:
Let's do it as per the previous plan and iterate it once we have a better solution.

pkg/apis/config/v1alpha1/resourceinterpretercustomization_types.go
RainbowMango: Great!, Let's do it by the function way, I'll update the API description later.

PR #2684, opened 2022-10-25
New Factory global flags merge function

helen-frank:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
These code comes multiple times, we may extract them as one function.
**Which issue(s) this PR fixes**:
Fixes #
Port #2611 
**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #2694, opened 2022-10-28
karmadactl promote uses factory to access cluster

lonelyCZ:

**What type of PR is this?**
/kind feature
/kind cleanup

Not only introduce factory, but cleanup the code of promote.

Part of #2349 

We need to improve the code of karmadactl ASAP, then add unit test for it.

/cc @carlory @XiShanYongYe-Chang 

```release-note
NONE
```

### DIFF
--- a/pkg/karmadactl/get.go
+++ b/pkg/karmadactl/get.go
@@ -33,7 +33,6 @@ import (
 	"k8s.io/kubectl/pkg/util/templates"
 	utilpointer "k8s.io/utils/pointer"
 
-	clusterv1alpha1 "github.com/karmada-io/karmada/pkg/apis/cluster/v1alpha1"
 	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
 	karmadaclientset "github.com/karmada-io/karmada/pkg/generated/clientset/versioned"
 	"github.com/karmada-io/karmada/pkg/karmadactl/util"
@@ -295,18 +294,18 @@ func (g *CommandGetOptions) Run(f util.Factory, cmd *cobra.Command, args []strin
 		return err
 	}
 
-	clusterInfos, err := getClusterInKarmadaForClient(gclient)
-	if err != nil {
-		return fmt.Errorf("method getClusterInKarmadaForClient get cluster info in karmada failed, err is: %w", err)
-	}
-
 	if err := g.getRBInKarmada(gclient); err != nil {
 		return err
 	}
 
 	if len(g.Clusters) <= 0 {
-		for c := range clusterInfos {
-			g.Clusters = append(g.Clusters, c)
+		clusterList, err := gclient.ClusterV1alpha1().Clusters().List(context.TODO(), metav1.ListOptions{})
+		if err != nil {
+			return fmt.Errorf("failed to list all member clusters in control plane, err: %w", err)
+		}
+
+		for i := range clusterList.Items {
+			g.Clusters = append(g.Clusters, clusterList.Items[i].Name)
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


XiShanYongYe-Chang (commented):
LGTM
/assign @carlory @helen-frank

pkg/karmadactl/get.go
```diff
@@ -961,8 +943,8 @@ func (g *CommandGetOptions) getRBInKarmada(gclient karmadaclientset.Interface) e
 	return nil
 }
 
-// getClusterInKarmadaForClient get cluster info in karmada cluster
-func getClusterInKarmadaForClient(gclient karmadaclientset.Interface) (map[string]*ClusterInfo, error) {
+// GetClusterInKarmadaForClient get cluster info in karmada cluster
+func GetClusterInKarmadaForClient(gclient karmadaclientset.Interface) (map[string]*ClusterInfo, error) {
```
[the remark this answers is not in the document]
lonelyCZ: Got it, thanks!

pkg/karmadactl/promote.go
```diff
@@ -121,11 +118,12 @@ type CommandPromoteOption struct {
 
 // AddFlags adds flags to the specified FlagSet.
 func (o *CommandPromoteOption) AddFlags(flags *pflag.FlagSet) {
-	o.GlobalCommandOptions.AddFlags(flags)
+	flags.StringVar(defaultConfigFlags.KubeConfig, "kubeconfig", *defaultConfigFlags.KubeConfig, "Path to the kubeconfig file to use for CLI requests.")
```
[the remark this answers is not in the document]
lonelyCZ: Right, we can use it after #2684 merged.

pkg/karmadactl/get.go
```diff
@@ -961,8 +943,8 @@ func (g *CommandGetOptions) getRBInKarmada(gclient karmadaclientset.Interface) e
 	return nil
 }
 
-// getClusterInKarmadaForClient get cluster info in karmada cluster
-func getClusterInKarmadaForClient(gclient karmadaclientset.Interface) (map[string]*ClusterInfo, error) {
+// GetClusterInKarmada get cluster info in karmada cluster
+func GetClusterInKarmada(gclient karmadaclientset.Interface) (map[string]*ClusterInfo, error) {
```
carlory, line 947: Why we need the function and the ClusterInfo struct. 

we can use directly  karmadaclientset.Interface to get cluster list in the Run function

lonelyCZ: Yes, just looked it, we don't need this function and struct now.

pkg/karmadactl/promote.go
```diff
@@ -172,81 +184,81 @@ func (o *CommandPromoteOption) Validate() error {
...
-		return fmt.Errorf("failed to get cluster info in karmada control plane. err: %v", err)
+	clusterInfos, err := GetClusterInKarmada(karmadaClient)
+	if err != nil {
+		return fmt.Errorf("failed to get cluster info in karmada control plane. err: %w", err)
 	}
 
-	if _, exist := clusterInfos[opts.Cluster]; !exist {
-		return fmt.Errorf("cluster(%s) does't exist in karmada control plane", opts.Cluster)
+	if _, exist := clusterInfos[o.Cluster]; !exist {
+		return fmt.Errorf("cluster(%s) does't exist in karmada control plane", o.Cluster)
```
carlory, line 204: the code logic of L198 - L204 is conflicted with the L209 - L219.

I think the  L198 - L204  is redundant.

pkg/karmadactl/promote.go
```diff
@@ -159,6 +159,18 @@ func (o *CommandPromoteOption) Complete(args []string) error {
 		}
 	}
 
+	if o.Namespace == "" {
```
carlory, line 162: this check is redundant.

FYI https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/apply/apply.go#L292

lonelyCZ: We don't use the `flags.Factory.ToRawKubeConfigLoader().Namespace()` directly, it is the default namespace in kubeconfig file, if we need to set other namespace, we need to speficy it.

https://github.com/karmada-io/karmada/blob/9b70cf2873a1dbd5d3ed45094e85e2b317b1ba44/pkg/karmadactl/promote.go#L324

lonelyCZ: Because, the namespace is not necessary for all subcommands, so we could extract the global flags `kubeconfig` and `karmada-context`. refer https://github.com/karmada-io/karmada/pull/2684#discussion_r1007801205

How do you think? @carlory

carlory: There is a significant difference between karmadactl and kubectl, kubectl operates on resources and is not responsible for cluster creation, while karmadactl is not. For deployment-related commands, it is indeed not suitable to introduce this parameter.

howerver, on the other hand, openshift has a command binary named oc, which can not only create a simple openshift cluster, but also operate the cluster. oc takes namespace as a global parameter.

IMO, I can accept them both. If the installation logic is splited from the karmadactl in the feature, I intend to takes namespace as a global parameter.

@lonelyCZ

lonelyCZ: > IMO, I can accept them both. If the installation logic is splited from the karmadactl in the feature, I intend to takes namespace as a global parameter.

Yes,  at that point, we can do it. It is clearer to separate temporarily and avoid invalid options.

lonelyCZ:
Any other suggestions, please? @carlory @helen-frank

helen-frank:
lgtm

lonelyCZ:
PTAL

/assign @RainbowMango

RainbowMango (approved):
/approve

RainbowMango (commented):
/lgtm

PR #2709, opened 2022-10-31
improve predicate.go

ikaven1024:

**What type of PR is this?**
/kind cleanup

When we predicate for update event, we shall predicate both the old and new object.
For example: https://github.com/karmada-io/karmada/blob/12ed0204436291c32645fa2148a041e609f38213/pkg/util/helper/predicate.go#L101-L107

When a work is updated to add `Suppressed` label, this predication will filter out this event, and client will never known the work is updated.

Part of #2579

```release-note
NONE
```

### DIFF
--- a/pkg/util/helper/predicate.go
+++ b/pkg/util/helper/predicate.go
@@ -3,6 +3,7 @@ package helper
 import (
 	"k8s.io/klog/v2"
 	controllerruntime "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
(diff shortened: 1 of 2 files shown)


XiShanYongYe-Chang:
> When a work is updated to Suppressed, these predication will filter out this event, and client will never known the work is updated.

I think what you said is that a work is updated to remove the `Suppressed` label, am I right?

ikaven1024:
> I think what you said is that a work is updated to remove the `Suppressed` label, am I right?

A work is updated to **add** the `Suppressed` label,

XiShanYongYe-Chang:
Actually, the label can be used only when the work is created. Therefore, the previous idea is to judge only one object in the update. The modification now makes the logic more rigorous.

ikaven1024:
> Actually, the label can be used only when the work is created. Therefore, the previous idea is to judge only one object in the update. The modification now makes the logic more rigorous.

Yes, it's more rigorous.
While the new logic will judge the old and new  object, and will invoke the client getting cluster twice. I am wonder if it’s worthy.

https://github.com/karmada-io/karmada/blob/0d0e7c333e6eb2b715e910dcd418d8793093f08e/pkg/util/helper/predicate.go#L115

XiShanYongYe-Chang:
Your concern is reasonable because the filtering of work is a high-frequency operation, and I suggest removing one judgment(Given the before logic, this is not a problem).

ikaven1024:
> Your concern is reasonable because the filtering of work is a high-frequency operation, and I suggest removing one judgment(Given the before logic, this is not a problem).

Why don't fetch clusters from cache?

XiShanYongYe-Chang:
The controller-runtime' client seems use the cache:

https://github.com/karmada-io/karmada/blob/bd8d734054f55f44786ffebd2acff015abea2963/vendor/sigs.k8s.io/controller-runtime/pkg/cluster/cluster.go#L51-L55

XiShanYongYe-Chang:
> Yes, I tested it.

Test with benchmark?

ikaven1024:
> > Yes, I tested it.
> 
> Test with benchmark?

I debug the code, and see it read from informer cache.

pkg/util/helper/predicate.go
```diff
@@ -17,74 +18,39 @@ import (
...
-			}
-
-			clusterObj, err := util.GetCluster(mgr.GetClient(), clusterName)
-			if err != nil {
-				klog.Errorf("Failed to get the given member cluster %s", clusterName)
-				return false
-			}
-
-			return clusterObj.Spec.SyncMode == clusterv1alpha1.Push
+			return predFunc("update", updateEvent.ObjectNew) || predFunc("update", updateEvent.ObjectOld)
```
chaunceyjiang, line 50: I think the original logic is easier to understand.

ikaven1024: The code in `UpdateFunc` and `DeleteFunc` are same, so I extract it.

XiShanYongYe-Chang (commented):
/lgtm

PR #2711, opened 2022-10-31
fix chart lint failure

Poor12:
**What type of PR is this?**
/kind failing-test

Recently chart-lint sometimes failed, such as https://github.com/karmada-io/karmada/actions/runs/3350829274/jobs/5551890843.

There are mainly two reasons:
1. Timeout. The installation of the release could not be completed within the given time(400s).
2. Post-install job failed. This is mainly because karmada-apiserver is always not ready due to readinessProbe. The job will install the crd but it cannot visit karmada-apiserver. After backoffLimit, it will fail in the end.

Fixes #

None


RainbowMango (approved):
/lgtm
/approve

Issue #2671, opened 2022-10-22
go test doesn't successfully on Goland


ikaven1024:
**What happened**:
When I run `go test` the `pkg/karmadactl/cmdinit/utils/util_test.go` on Goland,
<img width="454" alt="image" src="https://user-images.githubusercontent.com/17907604/197346325-9b1f57e6-914f-40e4-933f-c6389f384437.png">

In the console shows `Process finished with the exit code 0`, while Goland don't display the test tree green.
<img width="788" alt="image" src="https://user-images.githubusercontent.com/17907604/197346376-7116dde7-dd4a-461c-9178-8fc3bf7f4715.png">

**What you expected to happen**:
Test case shall be green in Goland window
<img width="624" alt="image" src="https://user-images.githubusercontent.com/17907604/197346746-ddcf003f-3e14-433c-b0ed-d64104092433.png">

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Karmada version: master
- kubectl-karmada or karmadactl version (the result of `kubectl-karmada version` or `karmadactl version`):
- Others: `go version go1.18.3 darwin/arm64`  
<img width="444" alt="image" src="https://user-images.githubusercontent.com/17907604/197346989-1774f8e8-302c-4934-940d-9549f75aaf21.png">

ikaven1024:
cd pkg/karmadactl/cmdinit/utils
go test -json  -race ./...
```
{"Time":"2022-10-22T22:52:31.64132+08:00","Action":"output","Package":"github.com/karmada-io/karmada/pkg/karmadactl/cmdinit/utils","Test":"TestDownloadFile","Output":"Download complete.--- PASS: TestDownloadFile (0.01s)\n"}
```

Thes output `Download complete.--- PASS: TestDownloadFile (0.01s)\n` mix the go test output and `util.go`
https://github.com/karmada-io/karmada/blob/875d2ef6c16780d5ef38e41d288be766f04ae93f/pkg/karmadactl/cmdinit/utils/util.go#L22-L30

Goland didn't parse the log correctly.

---
When I append the '\n' to the print,
```diff
--- a/pkg/karmadactl/cmdinit/utils/util.go
+++ b/pkg/karmadactl/cmdinit/utils/util.go
@@ -22,9 +22,9 @@ type Downloader struct {
 func (d *Downloader) Read(p []byte) (n int, err error) {
        n, err = d.Reader.Read(p)
[...]

PR #2713, opened 2022-10-31
Fix Test Not Success in GoLand

cmicat:

**What type of PR is this?**

/kind cleanup

The original test used `fmt.Printf` without `\n` to print log, this will cause stdout be polluted. Polluted stdout will cause GoLand parse test result failure.

This is similar to this issue: <https://github.com/karmada-io/karmada/issues/2671>

```release-note
NONE
```

### DIFF
--- a/pkg/search/proxy/controller_test.go
+++ b/pkg/search/proxy/controller_test.go
@@ -418,7 +418,6 @@ func TestController_Connect(t *testing.T) {
 			conn.ServeHTTP(recorder, req)
 
 			response := recorder.Result()
-			fmt.Printf("response: %v", response)
 
 			if (response.StatusCode != 200) != tt.wantErr {
 				t.Errorf("http request returned status code = %v, want error = %v",


pkg/search/proxy/controller_test.go
```diff
@@ -418,7 +418,7 @@ func TestController_Connect(t *testing.T) {
 			conn.ServeHTTP(recorder, req)
 
 			response := recorder.Result()
-			fmt.Printf("response: %v", response)
+			t.Logf("response: %v\n", response)
```
RainbowMango, line 421: Why log this out?

RainbowMango: In addition, we don't need `\n` in `t.Logf`, since `A final newline is added if not provided.`

cmicat: Eh, you are right, delete it is better.

ikaven1024 (approved):

XiShanYongYe-Chang (commented):
/lgtm

PR #2715, opened 2022-10-31
clean up proxy repeating code

ikaven1024:

**What type of PR is this?**
/kind cleanup

merge some repeating code

Fixes #

```release-note
NONE
```


pkg/registry/cluster/storage/proxy_test.go
```diff
@@ -0,0 +1,148 @@
...
+			}
+			h, err := r.Connect(req.Context(), tt.args.id, tt.args.options, utiltest.NewResponder(resp))
+			if (err != nil) != tt.wantErr {
+				t.Errorf("Connect() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+
+			if err != nil {
+				return
+			}
```
cmicat, line 135: why check err again with different logic?

XiShanYongYe-Chang (commented):
Thanks~
/lgtm

RainbowMango (approved):
/approve

pkg/registry/cluster/storage/proxy_test.go
XiShanYongYe-Chang: I think the reason is that when the error is not nill, there is no need to execute subsequent logic.

ikaven1024: yes

PR #2724, opened 2022-11-01
Fix and unify `CreateOrUpdateXXX` functions for karmadactl

### LINKED ISSUE #2687: CreateOrUpdateSecret function not works as expected
The `CreateOrUpdateSecret` doesn't work as expected as the `Update` operation would fail due to not refresh the latest `ResourceVersion`:
https://github.com/karmada-io/karmada/blob/5d42e82b7b92812a04aaade4b77e66556aa20d66/pkg/util/secret.go#L49-L61

For more details please refer to: https://github.com/karmada-io/karmada/pull/2596#discussion_r1006333298

Since this function is used by `karmadactl init` and `karmadactl register`, so we need to verify the two functionalities after we make changes.

**Why is this needed**:

lonelyCZ:

**What type of PR is this?**
/kind cleanup

Fixes #2687 

/cc @RainbowMango 

```release-note
NONE
```


RainbowMango:
/assign 
I'll look at it tomorrow.

pkg/karmadactl/util/idempotency.go
```diff
@@ -0,0 +1,160 @@
+package util
```
RainbowMango, line 1: What's the meaning of the file name `idempotency.go`? 
Should we move these functions to `pkg/util`?

lonelyCZ: This should be from https://github.com/kubernetes/kubernetes/blob/20936e51b80255bbb28a4eed5a814121dac6149e/cmd/kubeadm/app/util/apiclient/idempotency.go

I think it is more applied to `karmadactl` that doesn't call `update` frequently.

RainbowMango: that's ok. no big deal.

RainbowMango (approved):
/lgtm
/approve

RainbowMango:
As you know, the related function hasn't been covered by E2E, I assume you have verified it locally.
> Since this function is used by karmadactl init and karmadactl register, so we need to verify the two functionalities after we make changes.

lonelyCZ:
> As you know, the related function hasn't been covered by E2E, I assume you have verified it locally.

Right, it works fine in my env.

PR #2668, opened 2022-10-21
karmadactl token uses factory to access cluster

lonelyCZ:
Signed-off-by: lonelyCZ <531187475@qq.com>

**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:
Part of #2349 

**Special notes for your reviewer**:
/cc @carlory @helen-frank 

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #2730, opened 2022-11-03
Fix karmadactl token that can't set `--kubeconfig`

lonelyCZ:

**What type of PR is this?**

/kind bug

Fix that setting `--kubeconfig` and `--karmada-context` flags is invalid.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/karmadactl/token.go
+++ b/pkg/karmadactl/token.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"os"
 	"strings"
 	"time"
 
@@ -21,7 +20,6 @@ import (
 	"k8s.io/kubectl/pkg/cmd/get"
 	"k8s.io/kubectl/pkg/util/templates"
 
-	"github.com/karmada-io/karmada/pkg/karmadactl/options"
 	"github.com/karmada-io/karmada/pkg/karmadactl/util"
 	tokenutil "github.com/karmada-io/karmada/pkg/karmadactl/util/bootstraptoken"
 )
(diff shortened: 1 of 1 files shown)


RainbowMango:
How to reproduce it?

Which PR introduced this bug? We should figure out which release also needs this patch.

lonelyCZ:
> Which PR introduced this bug? We should figure out which release also needs this patch.

#2668 , it doesn't need patch.

> How to reproduce it?

```
[root@master67 karmada]# _output/bin/linux/amd64/karmadactl token create --kubeconfig /etc/karmada/karmada-apiserver.config --karmada-context karmada-apiserver
```

`--kubeconfig` and `--karmada-context` is invalid, it still uses `/root/.kube/config`.

RainbowMango:
I don't get why the previous flags not working. We do add the flags at :
https://github.com/karmada-io/karmada/blob/bcd7814bc1ea171d15c8477afd486593b681153b/pkg/karmadactl/token.go#L119

lonelyCZ:
> I don't get why the previous flags not working. We do add the flags at :
> 
> https://github.com/karmada-io/karmada/blob/bcd7814bc1ea171d15c8477afd486593b681153b/pkg/karmadactl/token.go#L119

Because we will directly use `defaultConfigFlags`, and not need to checkout the `kubeconfig` in subcommands.

https://github.com/karmada-io/karmada/blob/bcd7814bc1ea171d15c8477afd486593b681153b/pkg/karmadactl/karmadactl.go#L27-L28

RainbowMango (approved):
/lgtm
/approve

PR #2758, opened 2022-11-08
Introduce priority to PropagationPolicy

RainbowMango:
**What type of PR is this?**
/kind api-change
/kind feature

TBD.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/apis/policy/v1alpha1/propagation_types.go
+++ b/pkg/apis/policy/v1alpha1/propagation_types.go
@@ -70,6 +70,27 @@ type PropagationSpec struct {
 	// +optional
 	Placement Placement `json:"placement,omitempty"`
 
+	// Priority indicates the importance of a policy(PropagationPolicy or ClusterPropagationPolicy).
+	// A policy will be applied for the matched resource templates if there is
+	// no other policies with higher priority at the point of the resource
+	// template be processed.
+	// Once a resource template has been claimed by a policy, by default it will
+	// not be preempted by following policies even with a higher priority.
+	//
+	// In case of two policies have the same priority, the one with a more precise
+	// matching rules in ResourceSelectors wins:
+	// - matching by name(resourceSelector.name) has higher priority than
+	//   by selector(resourceSelector.labelSelector)
+	// - matching by selector(resourceSelector.labelSelector) has higher priority
+	//   than by APIVersion(resourceSelector.apiVersion) and Kind(resourceSelector.kind).
+	// If there is still no winner at this point, the one with the lower alphabetic
+	// order wins, e.g. policy 'bar' has higher priority than 'foo'.
+	//
+	// The higher the value, the higher the priority. Defaults to zero.
+	// +optional
+	// +kubebuilder:default=0
+	Priority *int32 `json:"priority,omitempty"`
+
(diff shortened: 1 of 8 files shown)


pkg/apis/policy/v1alpha1/propagation_types.go
```diff
@@ -70,6 +70,16 @@ type PropagationSpec struct {
...
 
+	// Priority indicates the importance of a policy(PropagationPolicy or ClusterPropagationPolicy).
+	// A policy will be applied for the matched resource templates if there is
+	// no other policies with higher priority at the point of the resource
+	// template be observed.
+	// Once a resource template has been claimed by a policy, it can not be
+	// preempted by following polices even with a higher priority.
+	// The higher the value, the higher the priority. Defaults to zero.
+	// +optional
+	Priority *int32 `json:"priority,omitempty"`
```
jwcesign, line 81: Why is `*int32` type used? Are there any restrictions on using `int32`?

RainbowMango: Echo from [API-conversion]():
> They(optinal fileds) are a pointer type in the Go definition (e.g. AwesomeFlag *SomeFlag) or have a built-in nil value (e.g. maps and slices).

But I also realized each time when access this field, we should check if it is nil, that's pretty annoying, especially when we can definitely be sure it can't be nil(since it has a default value).

So, I think we can use `int32` for this case.

RainbowMango:
cc @Garrybest

pkg/apis/policy/v1alpha1/propagation_types.go
RainbowMango: @jwcesign Please start to implement it based on this version. Let's see the type is an annoying thing.

RainbowMango:
> Do we need to define a range of priority value? like [0-1000) ?

Why? What's the benefit.

pkg/apis/policy/v1alpha1/propagation_types.go
Garrybest: So what about `Priority int32 `json:"priority"`?

RainbowMango:
The higher the value, the higher the priority.

pkg/apis/policy/v1alpha1/propagation_types.go
RainbowMango: After going through the implementation(#2767), looks not so bad. 

In addition, we can add some helpers function for the API, like:
```go
func (p *PropagationSpec) ExplicitPriority() int32 {
	if p.Priority == nil {
		return 0
	}
	return *p.Priority
}
```
So, @Garrybest do you have other comments?

Garrybest: I'm ok with this.

XiShanYongYe-Chang:
> > Do we need to define a range of priority value? like [0-1000) ?
> 
> Why? What's the benefit.

IMOP, if the user does not define a priority, the default priority will be 0, and it becomes meaningless to define a pp with a plural priority(Because once there is a pp with other default priorities, the pp will never be applied.). The creation of pp may not be artificial, so it is possible to create a plural priority, so it makes sense to limit the priority to a non-plural one.

PR #2772, opened 2022-11-10
Fix random errors occur during binary compilation

XiShanYongYe-Chang:
**What type of PR is this?**

/kind failing-test

Recently, we have frequently encountered CI failures:

https://github.com/karmada-io/karmada/actions/runs/3417843605/jobs/5689574998

This is obviously an occasional problem.

For the root cause of the problem, see @ikaven1024 's comments.

Fixes #

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,5 @@
 GOOS ?= $(shell go env GOOS)
 GOARCH ?= $(shell go env GOARCH)
-SOURCES := $(shell find . -type f  -name '*.go')
 LDFLAGS='$(shell hack/version.sh)'
 
 # Images management
@@ -50,7 +49,7 @@ CMD_TARGET=$(TARGETS) $(CTL_TARGETS)
 all: $(CMD_TARGET)
 
 .PHONY: $(CMD_TARGET)
-$(CMD_TARGET): $(SOURCES)
+$(CMD_TARGET):
 	LDFLAGS=$(LDFLAGS) BUILD_PLATFORMS=$(GOOS)/$(GOARCH) hack/build.sh $@
 
 # Build image.


RainbowMango:
We should try to find the root cause.

ikaven1024:
I think I got how to trigger this bug. Just type:
```bash
touch vendor/github.com/yuin/gopher-lua/parse/parser.go.y
```

As makefile rule, the prerequisites are newer than target, the command will be executed.
https://github.com/karmada-io/karmada/blob/3698c92bcfc96b757c0006e84a206db63836cbbc/vendor/github.com/yuin/gopher-lua/parse/Makefile#L3-L4

ikaven1024:
Our `make xxx` depends on `vendor/github.com/yuin/gopher-lua/parse/parser.go`, due to
https://github.com/karmada-io/karmada/blob/3698c92bcfc96b757c0006e84a206db63836cbbc/Makefile#L53

and .`../parser.go` depends on `.../parser.go.y` due to 
https://github.com/karmada-io/karmada/blob/3698c92bcfc96b757c0006e84a206db63836cbbc/vendor/github.com/yuin/gopher-lua/parse/Makefile#L3-L4

When timestamp of parser.go.y is later than parser.go (If these two file are checkout in different timestamp), above command will be executed. So bug appears.

ikaven1024:
We don't need to update the `parser.go`. So we remove it from our prerequisites. This fix works.

By the way, I think we can remove `$(SOURCES)`. Because we can accelerate our compiling with `GOCACHE`.

And the prerequisites may cause unexpected result.
For example, we have Makefile as follower:
```makefile
foo: $(SOURCES)
	go build -o foo
```

Then we type `GOOS=linux make foo` and build the executable `foo` for linux os.

Then we type `GOOS=darwin make foo`. Because the sources are older than the `foo` executable, so this build will not run.

ikaven1024:
> Traditionally, file timestamps had 1-second resolution, and make used those timestamps to determine whether one file was newer than the other.
-- from: https://www.gnu.org/software/autoconf/manual/autoconf-2.63/html_node/Timestamps-and-Make.html

Mostly, these files are checked out in same seconds, so don't have this error.

XiShanYongYe-Chang:
@ikaven1024 Thanks a lot.

/assign @RainbowMango

RainbowMango (approved):
/approve

Leave LGTM to @ikaven1024

RainbowMango (commented):
/lgtm

PR #2777, opened 2022-11-11
chart: optimize pod startup order

calvin0327:

**What type of PR is this?**

In our chart install workflow, use `post-job` to install all karmada crds. there is a problem, the karmada-controller-manager  is be installed before the job due to it always not running. becuase the karmada resource is not be `informer`:

wecom-temp-8ad6d41dcbe2b869cd65ae7833470986

so, we can install the karmada controller manager after the job have done.

Fixes #

please correct me if not.

```release-note
NONE
```


calvin0327:
@Poor12 PTAL

XiShanYongYe-Chang:
Change the label to the correct one.
/kind cleanup

PR #2780, opened 2022-11-12
karmadactl init uses factory to access cluster

helen-frank:
**What type of PR is this?**
/kind feature

see https://github.com/karmada-io/karmada/issues/2349
Fixes #

Self-test without abnormality
/cc @lonelyCZ @carlory

```release-note
NONE
```

### DIFF
--- a/pkg/karmadactl/cmdinit/cmdinit.go
+++ b/pkg/karmadactl/cmdinit/cmdinit.go
@@ -9,6 +9,7 @@ import (
 	"k8s.io/kubectl/pkg/util/templates"
 
 	"github.com/karmada-io/karmada/pkg/karmadactl/cmdinit/kubernetes"
+	"github.com/karmada-io/karmada/pkg/karmadactl/options"
 	"github.com/karmada-io/karmada/pkg/karmadactl/util"
 	"github.com/karmada-io/karmada/pkg/version"
 )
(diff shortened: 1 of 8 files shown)


lonelyCZ:
It seems that `Factory` only apply to interact with karmada control plane. `Init` command is for deploying Karmada in kubernetes host cluster. `Factory` doesn't play its advantages.

I think `apiclient` tool is more appropriate. How do you think? @helen-frank

helen-frank:
> It seems that `Factory` only apply to interact with karmada control plane. `Init` command is for deploying Karmada in kubernetes host cluster. `Factory` doesn't play its advantages.
>
> I think `apiclient` tool is more appropriate. How do you think? @helen-frank

The default operation object for karmadactl is local kube (~/.kube/config)

Perhaps we should distinguish between deployment commands and daily commands by setting different default operation objects, rather than by using `apiclient` or `Factory` to build client.

`Factory` can replace `apiclient`, use kube's methods instead of building its own wheels

What do you think 😀 @lonelyCZ

helen-frank:
Can I invite you to help me review the pr that speeds up the init process? @lonelyCZ

lonelyCZ:
https://github.com/karmada-io/karmada/blob/3cbae4043460ae28f72b1b4e82ad0ac7b17402b5/pkg/karmadactl/options/global.go#L22

Current `Factory` is not appropriate. Its main design goal is for `memberFactory`. We can refer `kubectl` and `kubeadm`, because we could split them to `amd` and `ctl` commands in future.

helen-frank:
> https://github.com/karmada-io/karmada/blob/3cbae4043460ae28f72b1b4e82ad0ac7b17402b5/pkg/karmadactl/options/global.go#L22
> 
> Current `Factory` is not appropriate. Its main design goal is for `memberFactory`. We can refer `kubectl` and `kubeadm`, because we could split them to `amd` and `ctl` commands in future.

ok, Agree

PR #2782, opened 2022-11-12
Karmadactl init: optimize getKarmadaAPIServerIP

helen-frank:
**What type of PR is this?**
/kind feature

getKarmadaAPIServerIP do not need to list node multiple times
Fixes #

Speed up for init

```release-note
NONE
```

### DIFF
--- a/pkg/karmadactl/cmdinit/kubernetes/node.go
+++ b/pkg/karmadactl/cmdinit/kubernetes/node.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net"
 	"strings"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


helen-frank:
@carlory @lonelyCZ cc

pkg/karmadactl/cmdinit/kubernetes/node.go
```diff
@@ -26,43 +26,34 @@ func (i *CommandInitOption) getKarmadaAPIServerIP() error {
...
 	klog.Info("randomly select 3 Node IPs in the kubernetes cluster.")
-	nodes, err := nodeClient.List(context.TODO(), metav1.ListOptions{})
-	if err != nil {
-		return err
-	}
-
 	for number := 0; number < 3; number++ {
 		if number >= len(nodes.Items) {
-			break
+			if len(i.KarmadaAPIServerIP) == 0 {
```
lonelyCZ, line 50: This code seems not necessary to embed here. Putting it the end could be clearer.

helen-frank: I apologize for not replying to your message for such a long time (I am infected with COVID-19 and have just improved recently😀)

I adjusted the original code logic

RainbowMango: :) I'm infected with COVID-19 as well...

RainbowMango:
Just a reminder: This branch has conflicts that must be resolved.

RainbowMango:
/assign

RainbowMango:
I'm sorry for letting this sit so long. After going through the code, listing nodes more than once is probably not a problem. 

Since in most cases, a cluster would have the nodes with `node-role.kubernetes.io/control-plane` (or `node-role.kubernetes.io/master`), listing with a label selector could limit the requested data. It would be helpful for clusters with a large number of nodes.

In other words, we usually don't need to list all nodes.

helen-frank:
> I'm sorry for letting this sit so long. After going through the code, listing nodes more than once is probably not a problem.
> 
> Since in most cases, a cluster would have the nodes with `node-role.kubernetes.io/control-plane` (or `node-role.kubernetes.io/master`), listing with a label selector could limit the requested data. It would be helpful for clusters with a large number of nodes.
> 
> In other words, we usually don't need to list all nodes.

This is indeed the case. My more consideration is that I can take out all the nodes at once and then filter the nodes of `node-role.kubernetes.io/control-plane` or `node-role.kubernetes.io/master`. You can consider this angle. If it is not acceptable, I will close the pr.

RainbowMango:
> You can consider this angle. 

Yeah, I looked at the patch. And my concern is that maybe we usually don't need to fetch all nodes. You know, node configuration is pretty large.

helen-frank:
got it and /close

PR #2765, opened 2022-11-09
Add limitReader to io.ReadAll

Poor12:
Signed-off-by: Poor12 <shentiecheng@huawei.com>

**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
`io.ReadAll` has no restrictions on memory request, which can easily lead to DoS attacks. Add limitReader can limit the memory.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:
```
`Security`: Add limitReader to `io.ReadAll` which could limit the memory request and avoid DoS attacks.
```

PR #2786, opened 2022-11-13
[bugfix] ignore resource that do not match with policy before apply policy

### LINKED ISSUE #2779: [E2E] `IngressStatus collection testing` run failed randomly
#### Which jobs are flaking:

https://github.com/karmada-io/karmada/actions/runs/3442118199/jobs/5744439084

```
[resource-status collection] resource status collection testing
/home/runner/work/karmada/karmada/test/e2e/resource_test.go:28
  IngressStatus collection testing
  /home/runner/work/karmada/karmada/test/e2e/resource_test.go:259
    [It] ingress status collection testing
```

#### Reason for failure:

After analysis, it is found that the problem is caused by the #2734 merged.

In the k8s cluster earlier than v1.22, the ingress `extensions/v1beta1` version has not been decrypted. So `detector` will setup both of them:

image

When a `networking.k8s.io/v1` version ingress is created, `detector` will receive two related event:

image

When reconcile `networking.k8s.io/v1` first and then reconcile `extensions/v1beta1`, the error will occur.

A ResoureceBinding is created for the `networking.k8s.io/v1` ingress, then reconcile the `extensions/v1beta1` ingress, the following logic is executed:

https://github.com/karmada-io/karmada/blob/3698c92bcfc96b757c0006e84a206db63836cbbc/pkg/detector/policy.go#L23-L30

In this way, the previous ResoureceBinding is overwritten and resources in ResoureceBinding are associated with `extensions/v1beta1` ingress.

The member k8s cluster version is 1.23. As a result, resource propagate fails for the reason:

```
server could not find the requested resource 
```

XiShanYongYe-Chang:

**What type of PR is this?**

/kind bug

For detail see #2779 

Fixes #2779 

```release-note
karmada-controlle-managerr: ignore resource that do not match with policy before apply policy
```

### DIFF
--- a/pkg/detector/policy.go
+++ b/pkg/detector/policy.go
@@ -93,6 +93,17 @@ func (d *ResourceDetector) getAndApplyPolicy(object *unstructured.Unstructured, 
 		return err
 	}
 
+	// Some resources are available in more than one group in the same kubernetes version.
+	// Therefore, the following scenarios occurs:
+	// In v1.21 kubernetes cluster, Ingress are available in both networking.k8s.io and extensions groups.
+	// When user creates an Ingress(networking.k8s.io/v1) and specifies a PropagationPolicy to propagate it
+	// to the member clusters, the detector will listen two resource creation events:
+	// Ingress(networking.k8s.io/v1) and Ingress(extensions/v1beta1). In order to prevent
+	// Ingress(extensions/v1beta1) from being propagated, we need to ignore it.
+	if !util.ResourceMatchSelectors(object, matchedPropagationPolicy.Spec.ResourceSelectors...) {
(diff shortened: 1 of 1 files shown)


RainbowMango (approved):
/lgtm
/approve

Seems we don't need to cherry-pick this. Correct me if I'm wrong.

XiShanYongYe-Chang:
> Seems we don't need to cherry-pick this. Correct me if I'm wrong.

It's still needed, The problem is found when #2765 is patched to release-1.2, in which release the `karmada-apiserver` version is v1.21.

RainbowMango:
OK. Then, pick it.

PR #2787, opened 2022-11-13
karmadactl get: add validate cluster exist

helen-frank:

**What type of PR is this?**

add validate cluster exist
Fixes #

```release-note
karmadactl get: When entering a cluster that does not exist, it will report `error: cluster xxx does not exist`
```


pkg/karmadactl/get/get.go
```diff
@@ -244,6 +251,24 @@ func (g *CommandGetOptions) Validate(cmd *cobra.Command) error {
...
+			return err
+		}
+		clusterSet := sets.NewString()
+		for _, cluster := range clusters.Items {
+			clusterSet.Insert(cluster.Name)
+		}
+		errs := []error{}
+		for _, cluster := range g.Clusters {
+			if !clusterSet.Has(cluster) {
+				errs = append(errs, fmt.Errorf("cluster %s does not exist", cluster))
```
jwcesign, line 267: the error message will be like: `[cluster does not exist:0, cluster does not exist:1, cluster does not exist:2]`, which is not a good format, can we make it better?

helen-frank: I am a reference https://github.com/karmada-io/karmada/blob/aeac17b77dc98c5dc74435fad9b5c8dc40b65be8/pkg/karmadactl/apply/apply.go#L134

helen-frank: I have considered it. I agree with your opinion and will revise apply after that.

helen-frank:
```shell
# _output/bin/linux/amd64/karmadactl --kubeconfig /etc/karmada/karmada-apiserver.config get pod -C me1,meqe -A                                                                          [0]
error: [ cluster does not exist: me1, cluster does not exist: meqe ]
```

pkg/karmadactl/get/get.go
```diff
@@ -244,6 +251,25 @@ func (g *CommandGetOptions) Validate(cmd *cobra.Command) error {
...
+		}
+
+		errs := []string{}
+		for _, cluster := range g.Clusters {
+			if !clusterSet.Has(cluster) {
+				errs = append(errs, "cluster does not exist: "+cluster)
+			}
+		}
+		return fmt.Errorf("[ " + strings.Join(errs, ", ") + " ]")
+	}
```
jwcesign, line 272: How about this:
```go
	noneExistClusters := []string{}
	for _, cluster := range g.Clusters {
		if !clusterSet.Has(cluster) {
			noneExistClusters = append(noneExistClusters, cluster)
		}
	}
	if len(noneExistClusters) != 0 {
		return fmt.Errorf("cluster does not exist:" + strings.Join(noneExistClusters, ","))
	}
```

jwcesign:
other LGTM
cc @XiShanYongYe-Chang

pkg/karmadactl/get/get.go
helen-frank: ok

pkg/karmadactl/get/get.go
```diff
@@ -244,6 +251,27 @@ func (g *CommandGetOptions) Validate(cmd *cobra.Command) error {
...
+		}
+
+		noneExistClusters := []string{}
+		for _, cluster := range g.Clusters {
+			if !clusterSet.Has(cluster) {
+				noneExistClusters = append(noneExistClusters, cluster)
+			}
+		}
+		if len(noneExistClusters) != 0 {
+			return fmt.Errorf("clusters does not exist: " + strings.Join(noneExistClusters, ","))
```
lonelyCZ, line 272:
```suggestion
			return fmt.Errorf("clusters don't exist: " + strings.Join(noneExistClusters, ","))
```

PR #2799, opened 2022-11-14
clean up rbac items when install successfully

cleverhu:
**What type of PR is this?**

When install by helm, there are some redundant rbac resources which can be cleaned up after successful installation.

Fixes #

```release-note

```

### DIFF
--- a/charts/karmada/templates/pre-install-job.yaml
+++ b/charts/karmada/templates/pre-install-job.yaml
@@ -228,6 +228,7 @@ metadata:
   annotations:
     "helm.sh/hook": pre-install
     "helm.sh/hook-weight": "1"
+    "helm.sh/hook-delete-policy": hook-succeeded
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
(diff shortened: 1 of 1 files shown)


cleverhu:
> Have you tested installing Karmada by helm chart? I'm afraid it's not ok. To run pre-install-job, it will install rbac resources first and then install job. This change may clean rbac resources before installing job.

This might result in errors, the resources will be deleted after helm install successfully. In fact, in our enterprise use, we set the deletion policy as ` "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded`, which didn't lead to errors. I'm not sure whether it is appropriate to add it to karmada.

Poor12:
I prefer not. In addition to the reasons I said above, it will also be used in post-install-job to clean up leftover resources.

cleverhu:
> post-install-job 

Post-install-job is also included throughout the installation process, so there will be no problem. Helm will delete resources after the operation is completed.

RainbowMango:
@chaosi-zju Can you help to confirm if this issue still exists on master?

RainbowMango:
OK. Thanks for the confirmation. Then we don't need this PR anymore.
/close

PR #2655, opened 2022-10-18
introduce `--private-image-registry` flag to `karmadactl init`

helen-frank:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
At present, it is mainly used for offline installation karmada to simplify installation commands.
**Which issue(s) this PR fixes**:
Fixes #
Port of #2541 
**Special notes for your reviewer**:
Test
```shell
# go run cmd/karmadactl/karmadactl.go init --image-registry=xxx/xxx --crds /root/crds.tar.gz
I1019 00:34:22.710879 1942695 deploy.go:188] kubeconfig file: /root/.kube/config, kubernetes: https://10.29.12.229:44230
I1019 00:34:22.732691 1942695 deploy.go:208] karmada apiserver ip: [172.18.0.4]
I1019 00:34:23.492457 1942695 cert.go:229] Generate ca certificate success.
I1019 00:34:23.695980 1942695 cert.go:229] Generate karmada certificate success.
I1019 00:34:24.011098 1942695 cert.go:229] Generate apiserver certificate success.
I1019 00:34:24.497476 1942695 cert.go:229] Generate front-proxy-ca certificate success.
I1019 00:34:24.798200 1942695 cert.go:229] Generate front-proxy-client certificate success.
I1019 00:34:25.060485 1942695 cert.go:229] Generate etcd-ca certificate success.
I1019 00:34:25.309735 1942695 cert.go:229] Generate etcd-server certificate success.
I1019 00:34:25.391543 1942695 cert.go:229] Generate etcd-client certificate success.
...
------------------------------------------------------------------------------------------------------
 █████   ████   █████████   ███████████   ██████   ██████   █████████   ██████████     █████████
░░███   ███░   ███░░░░░███ ░░███░░░░░███ ░░██████ ██████   ███░░░░░███ ░░███░░░░███   ███░░░░░███
 ░███  ███    ░███    ░███  ░███    ░███  ░███░█████░███  ░███    ░███  ░███   ░░███ ░███    ░███
 ░███████     ░███████████  ░██████████   ░███░░███ ░███  ░███████████  ░███    ░███ ░███████████
 ░███░░███    ░███░░░░░███  ░███░░░░░███  ░███ ░░░  ░███  ░███░░░░░███  ░███    ░███ ░███░░░░░███
 ░███ ░░███   ░███    ░███  ░███    ░███  ░███      ░███  ░███    ░███  ░███    ███  ░███    ░███
[...]

PR #2840, opened 2022-11-20
introduce kube-image-tag flag to karmadactl init

helen-frank:
**What type of PR is this?**
/kind feature

Choose a specific Kubernetes version for the control plane.
Fixes #
see https://github.com/karmada-io/karmada/pull/2541 https://github.com/karmada-io/karmada/pull/2655
TEST
```
./karmadactl init --private-image-registry=daocloud.io/atsctoo --kube-image-tag=v1.18.20 --crds=/root/karmada/crds.tar.gz
```

```release-note
`karmadactl`: Introduced `--kube-image-tag` flag to the `init` command to specify the Kubernetes image version.
```

### DIFF
--- a/pkg/karmadactl/cmdinit/cmdinit.go
+++ b/pkg/karmadactl/cmdinit/cmdinit.go
@@ -97,6 +97,7 @@ func NewCmdInit(parentCommand string) *cobra.Command {
 	// kube image registry
 	flags.StringVarP(&opts.KubeImageMirrorCountry, "kube-image-mirror-country", "", "", "Country code of the kube image registry to be used. For Chinese mainland users, set it to cn")
 	flags.StringVarP(&opts.KubeImageRegistry, "kube-image-registry", "", "", "Kube image registry. For Chinese mainland users, you may use local gcr.io mirrors such as registry.cn-hangzhou.aliyuncs.com/google_containers to override default kube image registry")
+	flags.StringVar(&opts.KubeImageTag, "kube-image-tag", "v1.25.2", "Choose a specific Kubernetes version for the control plane.")
 	// cert
 	flags.StringVar(&opts.ExternalIP, "cert-external-ip", "", "the external IP of Karmada certificate (e.g 192.168.1.2,172.16.1.2)")
 	flags.StringVar(&opts.ExternalDNS, "cert-external-dns", "", "the external DNS of Karmada certificate (e.g localhost,localhost.com)")
--- a/pkg/karmadactl/cmdinit/kubernetes/deploy.go
+++ b/pkg/karmadactl/cmdinit/kubernetes/deploy.go
@@ -47,9 +47,7 @@ var (
 
 	karmadaRelease string
 
-	defaultEtcdImage                  = "etcd:3.5.3-0"
-	defaultKubeAPIServerImage         = "kube-apiserver:v1.25.2"
-	defaultKubeControllerManagerImage = "kube-controller-manager:v1.25.2"
+	defaultEtcdImage = "etcd:3.5.3-0"
 
 	// DefaultCrdURL Karmada crds resource
 	DefaultCrdURL string
@@ -92,6 +90,7 @@ type CommandInitOption struct {
 	ImageRegistry                      string
 	KubeImageRegistry                  string
 	KubeImageMirrorCountry             string
+	KubeImageTag                       string
 	EtcdImage                          string
 	EtcdReplicas                       int32
 	EtcdInitImage                      string
@@ -552,7 +551,8 @@ func (i *CommandInitOption) kubeAPIServerImage() string {
 	if i.KarmadaAPIServerImage != "" {
 		return i.KarmadaAPIServerImage
 	}
-	return i.kubeRegistry() + "/" + defaultKubeAPIServerImage
+
+	return i.kubeRegistry() + "/kube-apiserver:" + i.KubeImageTag
 }
 
 // get kube-controller-manager image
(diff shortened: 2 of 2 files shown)


helen-frank:
According to our current application scenario, we are more concerned about the version of kube (some problems of interface compatibility), so I limited the scope of tag parameters discussed earlier to kube. 
What do you think ? @RainbowMango @carlory @lonelyCZ

carlory:
LGTM

pkg/karmadactl/cmdinit/cmdinit.go
```diff
@@ -97,6 +97,7 @@ func NewCmdInit(parentCommand string) *cobra.Command {
 	// kube image registry
 	flags.StringVarP(&opts.KubeImageMirrorCountry, "kube-image-mirror-country", "", "", "Country code of the kube image registry to be used. For Chinese mainland users, set it to cn")
 	flags.StringVarP(&opts.KubeImageRegistry, "kube-image-registry", "", "", "Kube image registry. For Chinese mainland users, you may use local gcr.io mirrors such as registry.cn-hangzhou.aliyuncs.com/google_containers to override default kube image registry")
+	flags.StringVar(&opts.KubeVersion, "kube-version", "v1.25.2", "Choose a specific Kubernetes version for the control plane.")
```
lonelyCZ, line 100: How about `kube-image-tag`? It seems to account for the relation with `kube-image-registry`.

helen-frank: I refer to https://github.com/kubernetes/kubernetes/blob/57eb5d631ccd615cd161b6da36afc759af004b93/cmd/kubeadm/app/cmd/options/generic.go#L79

I think these two can be a definition @lonelyCZ

RainbowMango: > kubeadm is specially to install Kubernetes that is a little different from my scenario. 

Totally agree!

@helen-frank Can you help to describe why we need a new flag? I'm kind of lost from it.

We already have the following flags:
- --karmada-kube-controller-manager-image=''
- --etcd-image=''
- --etcd-init-image='docker.io/alpine:3.15.1'
- --karmada-apiserver-image=''

helen-frank:
```shell
./karmadactl init --private-image-registry=daocloud.io/atsctoo --kube-version=v1.18.20 --crds=/root/karmada/crds.tar.gz`
```
replace
```shell
./karmadactl init --private-image-registry=daocloud.io/atsctoo --karmada-apiserver-image=daocloud.io/atsctoo/kube-apiserver:v1.18.20 --karmada-kube-controller-manager-image=daocloud.io/atsctoo/kube-controller-manager:v1.18.20 --crds=/root/karmada/crds.tar.gz`
```
For common application scenarios, kube's version is more concerned, and etcd may be neglected, so the scope of tag is limited to kube. @RainbowMango

RainbowMango: Thanks for your explanation.
I can see there is a slight benefit.
My concern is introducing a new flag(no matter `kube-image-tag` or `--kube-version`) also brings confusion.
It no clear specification of which components would effect by this flag.

RainbowMango: My opinion is to postpone this feature until we have clear evidence or we have a better alternative.

helen-frank: Yes, this parameter is very helpful for me to provide offline installation karmada commands (currently, the platform cannot upgrade kube versions with karmada to adapt, so different versions of kube will be installed in different scenarios)
So, let's take a look at everyone's opinions.  @RainbowMango , can you help me @ some active partners?

RainbowMango: > currently, the platform cannot upgrade kube versions with karmada to adapt

What do you mean?

helen-frank: Some of our existing business platforms have different requirements for kube versions, so there is often a requirement for setting up kube versions when using karmadactl to install karmada, while we are not particularly concerned about others. For the Karmada itself, we use the latest stable version.

helen-frank: I think this is a common situation. What do you think

RainbowMango: Yes, I agree it's a reasonable case. 
Do we have any other alternatives in addition to `kube-image-tag and --kube-version`?

helen-frank: At present, there is no better idea. Individuals are more inclined to `kube-version`. `kube-image-tag` is also acceptable.

helen-frank:
I want to use kube-version function now. Can I continue this topic? @lonelyCZ @RainbowMango

lonelyCZ:
I think it can be move forward. Now the main question is the name of the option. Perhaps, we can vote for deciding it.

I perfer `kube-image-tag` that is more match to `kube-image-registry` and can express more semantics .

carlory:
kube-image-tag +1

RainbowMango:
I'm ok with it.

XiShanYongYe-Chang:
In terms of contextual consistency, it seems that `kube-image-tag` looks a little more readable.

helen-frank:
OK, everyone seems to approve `kube-image-tag` more

pkg/karmadactl/cmdinit/kubernetes/deploy.go
```diff
@@ -92,6 +90,7 @@ type CommandInitOption struct {
 	ImageRegistry                      string
 	KubeImageRegistry                  string
 	KubeImageMirrorCountry             string
+	KubeVersion                        string
```
lonelyCZ, line 93:
```suggestion
	KubeImageTag                        string
```

helen-frank: fixed

lonelyCZ:
/lgtm

Thanks @helen-frank , and please update the PR description.

RainbowMango (approved):
/approve

I updated the release notes, by the way.

helen-frank:
Thank you😂

Issue #2834, opened 2022-11-18
Scheduler: Assignment Algorithm Refactor TODO List


Garrybest:
Now the assginment algorithm and code structure is complex, I'm trying to refactor the algorithms.

- [x] refactor assign replicas strategy #2827
- [x] refactor static weight dividing strategy #2845
- [x] refactor dynamic weight dividing strategy #2946
- [x] refactor aggregated dividing strategy #2946
- [ ] refactor job completion dividing algorithm #2964

PR #2845, opened 2022-11-22
refactor static weight strategy

Garrybest:

**What type of PR is this?**
/kind cleanup

Fixes part of #2834

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/core/assignment.go
+++ b/pkg/scheduler/core/assignment.go
@@ -48,14 +48,28 @@ func assignByAggregatedStrategy(state *assignState) ([]workv1alpha2.TargetCluste
 	return divideReplicasByResource(state.candidates, state.object, policyv1alpha1.ReplicaDivisionPreferenceAggregated)
 }
 
-// assignByStaticWeightStrategy assigns replicas by StaticWeightStrategy.
+/*
+* assignByStaticWeightStrategy assigns a total number of replicas to the selected clusters by the weight list.
+* For example, we want to assign replicas to two clusters named A and B.
+* | Total | Weight(A:B) | Assignment(A:B) |
+* |   9   |   1:2       |     3:6         |
+* |   9   |   1:3       |     2:7         | Approximate assignment
+* Note:
+* 1. If any selected cluster which not present on the weight list will be ignored(different with '0' replica).
+* 2. In case of not enough replica for specific cluster which will get '0' replica.
+ */
 func assignByStaticWeightStrategy(state *assignState) ([]workv1alpha2.TargetCluster, error) {
 	// If ReplicaDivisionPreference is set to "Weighted" and WeightPreference is not set,
 	// scheduler will weight all clusters averagely.
 	if state.strategy.WeightPreference == nil {
 		state.strategy.WeightPreference = getDefaultWeightPreference(state.candidates)
 	}
-	return divideReplicasByStaticWeight(state.candidates, state.strategy.WeightPreference.StaticWeightList, state.object.Replicas)
+	weightList := getStaticWeightInfoList(state.candidates, state.strategy.WeightPreference.StaticWeightList)
+
+	acc := newDispenser(state.object.Replicas, nil)
+	acc.takeByWeight(weightList)
+
+	return acc.result, nil
 }
 
 // assignByDynamicWeightStrategy assigns replicas by assignByDynamicWeightStrategy.
--- a/pkg/scheduler/core/division_algorithm.go
+++ b/pkg/scheduler/core/division_algorithm.go
@@ -75,7 +75,7 @@ func getStaticWeightInfoList(clusters []*clusterv1alpha1.Cluster, weightList []p
 		var weight int64
 		for _, staticWeightRule := range weightList {
 			if util.ClusterMatches(cluster, staticWeightRule.TargetCluster) {
-				weight += staticWeightRule.Weight
+				weight = util.MaxInt64(weight, staticWeightRule.Weight)
 			}
 		}
 		if weight > 0 {
--- a/pkg/scheduler/core/division_algorithm.go
+++ b/pkg/scheduler/core/division_algorithm.go
@@ -20,6 +20,82 @@ func (a TargetClustersList) Len() int           { return len(a) }
 func (a TargetClustersList) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }
 func (a TargetClustersList) Less(i, j int) bool { return a[i].Replicas > a[j].Replicas }
 
+type dispenser struct {
+	numReplicas int32
+	result      []workv1alpha2.TargetCluster
+}
+
+func newDispenser(numReplicas int32, init []workv1alpha2.TargetCluster) *dispenser {
+	cp := make([]workv1alpha2.TargetCluster, len(init))
+	copy(cp, init)
+	return &dispenser{numReplicas: numReplicas, result: cp}
+}
+
+func (a *dispenser) done() bool {
+	return a.numReplicas == 0 && len(a.result) != 0
+}
+
+func (a *dispenser) takeByWeight(w helper.ClusterWeightInfoList) {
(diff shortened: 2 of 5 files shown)


pkg/scheduler/core/division_algorithm.go
```diff
@@ -20,6 +20,82 @@ func (a TargetClustersList) Len() int           { return len(a) }
...
+	a.result = util.MergeTargetClusters(a.result, result)
+}
+
+func getStaticWeightInfoList(clusters []*clusterv1alpha1.Cluster, weightList []policyv1alpha1.StaticClusterWeight) helper.ClusterWeightInfoList {
+	list := make(helper.ClusterWeightInfoList, 0)
+	for _, cluster := range clusters {
+		var weight int64
+		for _, staticWeightRule := range weightList {
+			if util.ClusterMatches(cluster, staticWeightRule.TargetCluster) {
+				weight += staticWeightRule.Weight
```
RainbowMango, line 78: What if a cluster matches by more than one `weightRule`? 
Should we get the maximum weight instead of summing it?

RainbowMango (commented):
Generally looks good. 
Please @jwcesign take a look

pkg/scheduler/core/division_algorithm.go
Garrybest: Good idea! Sum may be a confusion.

pkg/scheduler/core/division_algorithm.go
```diff
@@ -20,6 +20,82 @@ func (a TargetClustersList) Len() int           { return len(a) }
...
+	// TODO(Garrybest): take rest replicas by fraction part
+	for i := range result {
+		if remain == 0 {
+			break
+		}
+		result[i].Replicas++
+		remain--
+	}
+
+	a.numReplicas = remain
```
jwcesign, line 68: What's the function of step `a.numReplicas = remain` and `if a.done() {`? In waht kind of scenario, it will be used

pkg/scheduler/core/division_algorithm.go
```diff
@@ -20,6 +20,82 @@ func (a TargetClustersList) Len() int           { return len(a) }
...
+	remain := a.numReplicas
+	for _, info := range w {
+		replicas := int32(info.Weight * int64(a.numReplicas) / sum)
+		result = append(result, workv1alpha2.TargetCluster{
+			Name:     info.ClusterName,
+			Replicas: replicas,
+		})
+		remain -= replicas
+	}
+	// TODO(Garrybest): take rest replicas by fraction part
```
jwcesign, line 59: I think this doesnt need more works, it looks fine

pkg/scheduler/core/assignment.go
```diff
@@ -48,14 +48,28 @@ func assignByAggregatedStrategy(state *assignState) ([]workv1alpha2.TargetCluste
...
+* 2. In case of not enough replica for specific cluster which will get '0' replica.
+ */
 func assignByStaticWeightStrategy(state *assignState) ([]workv1alpha2.TargetCluster, error) {
 	// If ReplicaDivisionPreference is set to "Weighted" and WeightPreference is not set,
 	// scheduler will weight all clusters averagely.
 	if state.strategy.WeightPreference == nil {
 		state.strategy.WeightPreference = getDefaultWeightPreference(state.candidates)
 	}
-	return divideReplicasByStaticWeight(state.candidates, state.strategy.WeightPreference.StaticWeightList, state.object.Replicas)
+	weightList := getStaticWeightInfoList(state.candidates, state.strategy.WeightPreference.StaticWeightList)
```
jwcesign, line 67: According to your ut, if `state.object.Replicas` is 0, just give a waring log and return it.

Garrybest: Acutually, it never happens. Assignment is only revoked when replicas are non-zero.

https://github.com/karmada-io/karmada/blob/master/pkg/scheduler/core/generic_scheduler.go#L181

pkg/scheduler/core/division_algorithm.go
Garrybest: Well, this assignment may be a little unfair in some scenario. Let me give a example here.

1. We have 3 clusters which have a weight of [A:1, B:2, C:3].
2. The desired replicas is 2.

According to this alg, we may have a result of [A:1, B:1], which is not fair because C got a higher weight.

pkg/scheduler/core/division_algorithm.go
Garrybest: It's a reservation here. The `dispenser` is designed for dispensing in multiple times considering constraints in the future. If we have some contraints, one-time dispensing may not finish yet. So we need this function here to conclude whether the dispensing is done.

pkg/scheduler/core/division_algorithm.go
jwcesign: I did the test, looks like we need to think more about this:
```go
		{
			name:        "tmp",
			numReplicas: 2,
			result: []workv1alpha2.TargetCluster{
				{Name: "A", Replicas: 0},
				{Name: "B", Replicas: 0},
				{Name: "C", Replicas: 0},
			},
			weightList: []utilhelper.ClusterWeightInfo{
				{ClusterName: "A", Weight: 1},
				{ClusterName: "B", Weight: 2},
				{ClusterName: "C", Weight: 3},
			},
			desired: []workv1alpha2.TargetCluster{
				{Name: "A", Replicas: 0},
				{Name: "B", Replicas: 1},
				{Name: "C", Replicas: 1},
			},
			done: true,
		},
```
The result:
image

pkg/scheduler/core/division_algorithm.go
jwcesign: I prefer to delete it, if we need it, then we add it back.

pkg/scheduler/core/division_algorithm.go
Garrybest: Sorry, this case is not proper because now we have sort the cluster by weight in descent order(I didn't notice before). Consider we have a weight of [A:5, B:4, C:1] and desired replicas is 9, the result is [A:5, B:4, C:0]. 

Actually, the weighted result is [A:4.5, B:3.6, C:0.9], I think [A:4, B:4, C:1] should be fair because 0.9 > 0.6 > 0.5, i.e., the rest replicas of C is closer to 1. This case demonstrates the TODO list, I think regarding the fraction part of weighted result and sorting them by descent order could be more accurate mathematically.🤣

pkg/scheduler/core/division_algorithm.go
Garrybest: All right.

pkg/scheduler/core/division_algorithm.go
jwcesign: Completely agree.

pkg/scheduler/core/division_algorithm.go
Garrybest: Hi @jwcesign, I have reconsidered this again. From the aspect of a go `struct`, if the invoker invokes function `takeByWeight` repeatedly, the result may be a mess because the dispensing action is done before unless invoker set a new `numReplicas`. So I think in terms of object-oriented concept, this function `done()` is vital and we never do it twice if `numReplicas` has been dispensed. How do you think? 😊

jwcesign: What I'm thinking about is how to use this function. I see that every time this function is called, a new structure will be created, so there is no concurrency problem(or assign two times).

Garrybest: You are right, from the viewpoint of invoker. Nevertheless, for the design of this struct, it does not know how the invoker will use the struct. So `done` is just an assurance for check here, I still consider it as a reasonable desgin here. 🤣

jwcesign:
/lgtm
cc @RainbowMango

PR #2846, opened 2022-11-22
fix strategy chart format

cleverhu:
**What type of PR is this?**

/kind cleanup

The strategy format is wrong, so I fixed it to the correct format.

Fixes #

```release-note

```


Poor12:
A small question: how do you discover this problem?

cleverhu:
> A small question: how do you discover this problem?

I found it by 'helm template./'.

Issue #2587, opened 2022-09-28
Status of PVC out of sync


lts0609:
**What happened**:
I created a multi cluster helm application，but the pvc status out of sync
1. create helm application on the karmada controller plane by argocd
2. synchronize resources
3. apply propagation

```
[root@node1 ~]# kubectl get pvc --kubeconfig /etc/karmada/karmada-apiserver.config
NAME      STATUS    VOLUME   CAPACITY   ACCESS MODES   STORAGECLASS   AGE
grafana   Pending                                                     106s
```

```
[root@node1 ~]# kubectl describe pvc grafana --kubeconfig /etc/karmada/karmada-apiserver.config
Name:          grafana
Namespace:     
StorageClass:
Status:        Pending
Events:
  Type     Reason                  Age                    From                  Message
  ----     ------                  ----                   ----                  -------
  Warning  ApplyPolicyFailed       2m33s                  resource-detector     No policy match for resource
  Normal   ApplyPolicySucceed      2m12s                  resource-detector     Apply policy(lts/grafana) succeed
  Normal   SyncSucceed             2m12s                  execution-controller  Successfully applied resource(lts/grafana) to cluster host
  Normal   SyncSucceed             2m12s                  execution-controller  Successfully applied resource(lts/grafana) to cluster member
  Normal   SyncWorkSucceed         2m11s (x6 over 2m12s)  binding-controller    Sync work of resourceBinding(lts/grafana-persistentvolumeclaim) successful.
  Normal   AggregateStatusSucceed  2m11s (x6 over 2m12s)  binding-controller    Update resourceBinding(lts/grafana-persistentvolumeclaim) with AggregatedStatus successfully.
  Normal   ScheduleBindingSucceed  2m11s (x3 over 2m12s)  karmada-scheduler     Binding has been scheduled
```

**Environment**:
- Karmada version: 1.3.0
- Kubernetes version: 1.22.9
- Others: helm application

RainbowMango:
@Poor12 Please help to reproduce and investigate it with v1.3.0.

lts0609:
I found that this can make the pvc state normal
1. Delete pvc in the karmada controller plane
2. Synchronize again resource by argocd
```
[root@node1 ~]# kubectl get pvc -n lts --kubeconfig /etc/karmada/karmada-apiserver.config
NAME      STATUS   VOLUME   CAPACITY   ACCESS MODES   STORAGECLASS   AGE
grafana   Bound                                                      16s
```
[...]

PR #2849, opened 2022-11-23
validate cluster fields

carlory:

**What type of PR is this?**

/kind api-change
/kind feature

when we use the fieldSelector field of the clusterAffinity to select clusters, some cluster won't be selected and we will get this scheduler result that there're no fit clusters. 

for example, the provider of cluster contains chinese characters, the scheduler will not select this cluster. 

because the fieldSelector is converted to a labelSelector, the labelSelector will check the value.

FYI: 

- https://github.com/karmada-io/karmada/blob/master/pkg/util/selector.go#L110
- https://github.com/karmada-io/karmada/blob/23db2ac9d66b87033b855bd9f694defadc3d5b1c/vendor/k8s.io/apimachinery/pkg/labels/selector.go#L192

```release-note
validate cluster fields: provider, region and zone.
```

### DIFF
--- a/pkg/apis/cluster/validation/validation.go
+++ b/pkg/apis/cluster/validation/validation.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"math"
 	"net/url"
+	"strings"
 
 	apimachineryvalidation "k8s.io/apimachinery/pkg/api/validation"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -22,6 +23,7 @@ const clusterNameMaxLength int = 48
 // - Must be a valid label value as per RFC1123.
 //   * An alphanumeric (a-z, and 0-9) string, with a maximum length of 63 characters,
 //     with the '-' character allowed anywhere except the first or last character.
+//
 // - Length must be less than 48 characters.
 //   * Since cluster name used to generate execution namespace by adding a prefix, so reserve 15 characters for the prefix.
 func ValidateClusterName(name string) []string {
@@ -90,6 +92,8 @@ func ValidateClusterSpec(spec *api.ClusterSpec, fldPath *field.Path) field.Error
 		allErrs = append(allErrs, ValidateClusterProxyURL(fldPath.Child("proxyURL"), spec.ProxyURL)...)
 	}
 
+	allErrs = append(allErrs, validateClusterSpecForFieldSelector(spec, fldPath)...)
+
 	if len(spec.Taints) > 0 {
 		allErrs = append(allErrs, lifted.ValidateClusterTaints(spec.Taints, fldPath.Child("taints"))...)
 	}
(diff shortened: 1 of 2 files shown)


pkg/apis/cluster/validation/validation.go
```diff
@@ -90,6 +94,24 @@ func ValidateClusterSpec(spec *api.ClusterSpec, fldPath *field.Path) field.Error
 		allErrs = append(allErrs, ValidateClusterProxyURL(fldPath.Child("proxyURL"), spec.ProxyURL)...)
 	}
 
+	if spec.Provider != "" {
+		if errs := kubevalidation.IsValidLabelValue(spec.Provider); len(errs) != 0 {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("provider"), spec.Provider, strings.Join(errs, "; ")))
+		}
+	}
```
RainbowMango, line 101: Maybe we can add a new function to eliminate the `gocyclo` warning here, take this `Provider` as an example, we don't need to check if the `provider` is empty or not outside. Something like:

```golang
allErrs = append(allErrs, ValidateClusterProvider(spec.Provider)...)
```

I'm saying this because I know some companies that don't use `gocyclo` but check the code complexity with other tools, so the `nolint:gocyclo` can't suppress the warning.

pkg/apis/cluster/validation/validation.go
```diff
@@ -20,10 +21,11 @@ const clusterNameMaxLength int = 48
 // If the cluster name is not valid, a list of error strings is returned. Otherwise an empty list (or nil) is returned.
 // Rules of a valid cluster name:
 // - Must be a valid label value as per RFC1123.
-//   * An alphanumeric (a-z, and 0-9) string, with a maximum length of 63 characters,
+//   - An alphanumeric (a-z, and 0-9) string, with a maximum length of 63 characters,
```
RainbowMango, line 24: Why change this?

RainbowMango: I see. It is the gofmt of Go1.19. 
It's included #2587.

carlory:
@RainbowMango can help me to rerun it ?

RainbowMango:
> W1123 16:10:47.522402   27264 cluster.go:153] failed to create cluster(member1). error: the server is currently unable to handle the request (post clusters.cluster.karmada.io)
W1123 16:10:47.522452   27264 cluster.go:98] failed to create cluster(member1). error: the server is currently unable to handle the request (post clusters.cluster.karmada.io)
error: failed to create cluster(member1) object. error: the server is currently unable to handle the request (post clusters.cluster.karmada.io)

OK. Just echo the error msg here.
https://github.com/karmada-io/karmada/actions/runs/3533606594/jobs/5929600480

pkg/apis/cluster/validation/validation.go
```diff
@@ -90,6 +92,10 @@ func ValidateClusterSpec(spec *api.ClusterSpec, fldPath *field.Path) field.Error
 		allErrs = append(allErrs, ValidateClusterProxyURL(fldPath.Child("proxyURL"), spec.ProxyURL)...)
 	}
 
+	allErrs = append(allErrs, validateClusterForSelector(fldPath.Child("provider"), spec.Provider))
+	allErrs = append(allErrs, validateClusterForSelector(fldPath.Child("region"), spec.Region))
+	allErrs = append(allErrs, validateClusterForSelector(fldPath.Child("zone"), spec.Zone))
```
RainbowMango, line 97: I'm afraid this will append a `nil` error to `allErrs`.

RainbowMango: CI failed again. I guess this is the reason.

RainbowMango: Sorry for I didn't clear it at my previous comment: https://github.com/karmada-io/karmada/pull/2849#discussion_r1030088832

The `validateClusterForSelector` can return a error slice, so that we can append it outside. append(x, y`...`).

carlory:
@RainbowMango updated

RainbowMango (approved):
/lgtm
/approve

PR #2866, opened 2022-11-24
fix dirty is not added in docker image tag

ikaven1024:

**What type of PR is this?**
/kind bug

```bash
$ echo >> README.md

$ make image-karmada-controller-manager
Building image for linux/amd64: docker.io/karmada/karmada-controller-manager:v1.3.0-452-g63a67d7c
...
```

Expect image: `docker.io/karmada/karmada-controller-manager:v1.3.0-452-g63a67d7c-dirty`

Fixes #

```release-note
fix dirty is not added in docker image tag
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,24 +1,12 @@
 GOOS ?= $(shell go env GOOS)
 GOARCH ?= $(shell go env GOARCH)
-LDFLAGS='$(shell hack/version.sh)'
+VERSION ?= '$(shell hack/version.sh)'
 
 # Images management
 REGISTRY?="docker.io/karmada"
 REGISTRY_USER_NAME?=""
 REGISTRY_PASSWORD?=""
 REGISTRY_SERVER_ADDRESS?=""
-
-# Set your version by env or using latest tags from git
-VERSION?=""
-ifeq ($(VERSION), "")
-    LATEST_TAG=$(shell git describe --tags)
-    ifeq ($(LATEST_TAG),)
-        # Forked repo may not sync tags from upstream, so give it a default tag to make CI happy.
-        VERSION="unknown"
-    else
-        VERSION=$(LATEST_TAG)
-    endif
-endif
 
 TARGETS := karmada-aggregated-apiserver \
 			karmada-controller-manager \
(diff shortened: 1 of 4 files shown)


RainbowMango (commented):
LGTM.

@gy95 Could you please take a look?

Makefile
```diff
@@ -1,25 +1,13 @@
...
 # Images management
 REGISTRY?="docker.io/karmada"
 REGISTRY_USER_NAME?=""
 REGISTRY_PASSWORD?=""
 REGISTRY_SERVER_ADDRESS?=""
 
-# Set your version by env or using latest tags from git
-VERSION?=""
-ifeq ($(VERSION), "")
-    LATEST_TAG=$(shell git describe --tags)
```
gy95, line 14: binary version and image tag should be kept the same.
For image tag, it comes here.
we just need to add a `--dirty` flag `LATEST_TAG=$(shell git describe --tags --dirty)`?

ikaven1024: As you say `binary version and image tag should be kept the same.`  It's best way to use the same code. So I refactor it to get from `util::get_version` function.

If you keep the two ways to get version. It's easy to make different results, just like this bug.

hack/build.sh
```diff
@@ -23,6 +23,8 @@ set -o pipefail
 REPO_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
 source "${REPO_ROOT}/hack/util.sh"
 
+LDFLAGS="$(util::version_ldflags) ${LDFLAGS:-}"
```
gy95, line 26: this is confusing

gy95: LDFLAGS="${LDFLAGS}:-$(util::version_ldflags)}" ? should be like this?

ikaven1024: It's not a mistake. Users can set custom `ldflags` by `LDFLAGS` variable., not only version info. FYI:
https://github.com/kubernetes/kubernetes/blob/a2408f39c7233fc7ceb06795ffd24b84ca471243/hack/lib/golang.sh#L847

gy95: /lgtm

RainbowMango (approved):
/approve

PR #2869, opened 2022-11-26
Update ResourceInterpreterCustomization comments according to implementation

RainbowMango:
**What type of PR is this?**
/kind cleanup

This PR updated ResourceInterpreterCustomization's comments as per the implementation.

Fixes #

```release-note
NONE
```

cc @XiShanYongYe-Chang @chaunceyjiang @ikaven1024 @jameszhangyukun


XiShanYongYe-Chang (commented):
lgtm

pkg/apis/config/v1alpha1/resourceinterpretercustomization_types.go
```diff
@@ -108,16 +108,27 @@ type CustomizationRules struct {
...
 	//             return desiredObj
 	//         end
 	//
-	// LuaScript only holds the function body part, take the `desiredObj` and `observedObj` as the function
-	// parameters or global variables, and finally returns a retained specification.
+	// The content of the LuaScript needs to be a whole function including both
+	// declaration and implementation.
+	//
+	// The parameters will be supplied by the system:
+	//   - desiredObj: the object represents the configuration to be applied
```
ikaven1024, line 125: In fact, the parameters's name is not fixed. It also accept:
```lua
function Retain(d, o)
    d.spec.fieldFoo = o.spec.fieldFoo
    return d
end
```

The only requirements are:
- function name must be `Retain`
- two input args, delivery desired and observed objects
- one return value, delivery the retained object

pkg/apis/config/v1alpha1/resourceinterpretercustomization_types.go
```diff
@@ -241,8 +287,14 @@ type DependencyInterpretation struct {
 	//             return dependencies
 	//         end
 	//
-	// LuaScript only holds the function body part, take the `desiredObj` as the function
-	// parameters or global variables, and finally returns the dependent resources.
+	// The content of the LuaScript needs to be a whole function including both
```
ikaven1024, line 290: I think it shall be ` dependencies[1] = ...`

https://github.com/karmada-io/karmada/blob/e7b9a11fbd934be2a75b449e67f634ef4052a68e/pkg/apis/config/v1alpha1/resourceinterpretercustomization_types.go#L284-L285

charts/karmada/_crds/bases/config.karmada.io_resourceinterpretercustomizations.yaml
```diff
@@ -57,10 +57,12 @@ spec:
                           = \"v1\" dependency.kind = \"ServiceAccount\" dependency.name
                           = desiredObj.spec.serviceAccountName dependency.namespace
                           = desiredObj.namespace dependencies[0] = {} dependencies[0]
```
jameszhangyukun, line 59: int lua, the array index starts from 1, so I think it shall be `dependencies[1]=dependency`

pkg/apis/config/v1alpha1/resourceinterpretercustomization_types.go
```diff
@@ -127,25 +138,31 @@ type LocalValueRetention struct {
...
-	//         end
+	// The parameters will be supplied by the system:
+	//   - desiredObj: the object represents the configuration to be applied
+	//       to the member cluster.
 	//
-	// LuaScript only holds the function body part, take the `desiredObj` as the function
-	// parameters or global variable, and finally returns the replica number and required resources.
+	// The function expects two return values:
+	//   - replica: the declared replica number
+	//   - requirement: the resource required by each replica expressed with a
```
chaunceyjiang, line 163: It is recommended to add a type description, such as `replica` is a `number` type and `requirement` is a `table` type.

pkg/apis/config/v1alpha1/resourceinterpretercustomization_types.go
RainbowMango: Yes, I think it's common sense that we don't need to explain more. :)

pkg/apis/config/v1alpha1/resourceinterpretercustomization_types.go
RainbowMango: +1

RainbowMango: Can you give an example?

pkg/apis/config/v1alpha1/resourceinterpretercustomization_types.go
RainbowMango: I agree, but it also makes the example hard to read.

pkg/apis/config/v1alpha1/resourceinterpretercustomization_types.go
RainbowMango: I totally agree with the idea. But after going through the Lua documents about [Values and Types](https://www.lua.org/manual/5.1/manual.html#2.2) I feel hard to describe the types. Maybe the example is clear enough? We can add more descriptions to documents.

charts/karmada/_crds/bases/config.karmada.io_resourceinterpretercustomizations.yaml
RainbowMango: Fixed.

pkg/apis/config/v1alpha1/resourceinterpretercustomization_types.go
chaunceyjiang: > We can add more descriptions to documents.

+1

PR #2871, opened 2022-11-26
add kube lib for lua

ikaven1024:

**What type of PR is this?**
/kind feature

When users use customization to implement `GetReplicas`, it's difficult to accurate the sum of resource quantity. This PR load the `kube` library to help do this work.

See the demo: https://github.com/karmada-io/karmada/blob/f6a9956c078c4886e88a3d98952d4ff3ecd319ba/pkg/resourceinterpreter/configurableinterpreter/luavm/lua_test.go#L65-L70

part of #2371

```release-note
NONE
```

### DIFF
--- a/pkg/resourceinterpreter/configurableinterpreter/luavm/kube.go
+++ b/pkg/resourceinterpreter/configurableinterpreter/luavm/kube.go
@@ -1,0 +1,93 @@
+package luavm
+
+import (
+	lua "github.com/yuin/gopher-lua"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+
+	"github.com/karmada-io/karmada/pkg/util/helper"
+)
+
+const (
+	// KubeLibName is the name of the kube library.
+	KubeLibName = "kube"
+)
+
+// KubeLoader loads the kube library. Import this library before your script by:
+//
+//	local kube = require("kube")
+//
+// Then you can call functions in this library by:
+//
+//	kube.xxx()
+//
+// This library Contains:
+//   - function resourceAdd(r1, r2, ...)
+//     accruing the quantity of resources. Example:
+//     cpu = kube.resourceAdd(r1.cpu, r2.cpu, r3.cpu)
+//   - function accuratePodRequirements(pod) requirements
+//     accurate total resource requirements for pod. Example:
+//     requirements = kube.accuratePodRequirements(pod)
+func KubeLoader(ls *lua.LState) int {
+	mod := ls.SetFuncs(ls.NewTable(), kubeFuncs)
+	ls.Push(mod)
+	return 1
+}
+
+var kubeFuncs = map[string]lua.LGFunction{
+	"resourceAdd":             resourceAdd,
+	"accuratePodRequirements": accuratePodRequirements,
+}
+
+func resourceAdd(ls *lua.LState) int {
+	res := resource.Quantity{}
+	n := ls.GetTop()
+	for i := 1; i <= n; i++ {
+		q := checkResourceQuantity(ls, i)
+		res.Add(q)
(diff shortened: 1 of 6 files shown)


RainbowMango:
I like the idea. Sounds like it makes easier for users to write the scripts.

pkg/resourceinterpreter/configurableinterpreter/luavm/lua_test.go
```diff
@@ -41,31 +41,55 @@ func TestGetReplicas(t *testing.T) {
...
+		nodeClaim = {
+			nodeSelector = desiredObj.spec.template.spec.nodeSelector,
+			tolerations = desiredObj.spec.template.spec.tolerations
+    	},
+		resourceRequest = {},
+	}
+	for i = 1, #desiredObj.spec.template.spec.containers do
+		requires.resourceRequest.cpu = kube.resourceAdd(requires.resourceRequest.cpu, desiredObj.spec.template.spec.containers[i].resources.limits.cpu)
+		requires.resourceRequest.memory = kube.resourceAdd(requires.resourceRequest.memory, desiredObj.spec.template.spec.containers[i].resources.limits.memory)
+	end
```
RainbowMango, line 78: Can we have a higher abstract function/lib to get the requirements from a `pod template`?

XiShanYongYe-Chang (commented):
LGTM

Do we need to update the `getReplicas` example in the `examples` at the same time?

ikaven1024:
> Do we need to update the `getReplicas` example in the `examples` at the same time?

OK. I will do it.

pkg/resourceinterpreter/configurableinterpreter/luavm/kube.go
```diff
@@ -0,0 +1,93 @@
...
+
+	"github.com/karmada-io/karmada/pkg/util/helper"
+)
+
+const (
+	// KubeLibName is the name of the kube library.
+	KubeLibName = "kube"
+)
+
+// KubeLoader loads the kube library. Import this library before you script by:
```
RainbowMango, line 16:
```suggestion
// KubeLoader loads the kube library. Import this library before your script by:
```
Is it a typo?

pkg/resourceinterpreter/configurableinterpreter/luavm/kube.go
```diff
@@ -0,0 +1,93 @@
...
+	if err != nil {
+		ls.RaiseError("fail to convert %#v to Lua value: %v", requirements, err)
+		return 0
+	}
+
+	ls.Push(retValue)
+	return 1
+}
+
+func checkResourceQuantity(ls *lua.LState, n int) resource.Quantity {
```
RainbowMango, line 81: This function isn't about to `check`, but `parse` a resource to resource.Quantity.

Rename to `ParseResourceQuantity` or one more accurate?

pkg/resourceinterpreter/configurableinterpreter/luavm/kube.go
```diff
@@ -0,0 +1,93 @@
...
+	n := ls.GetTop()
+	for i := 1; i <= n; i++ {
+		q := checkResourceQuantity(ls, i)
+		res.Add(q)
+	}
+
+	s := res.String()
+	ls.Push(lua.LString(s))
+	return 1
+}
```
RainbowMango, line 53: When this function would be used?

RainbowMango (commented):
Good job!
Only some nits and questions.

pkg/resourceinterpreter/configurableinterpreter/luavm/kube.go
ikaven1024: Users may want to accurate resource quantity with their logic, rather `accuratePodRequirements`.

PR #2878, opened 2022-11-28
karmada-agent: move all default values to NewOptions

Zhuzhenghao:
**What type of PR is this?**

/kind cleanup

I move all default options to function NewOptions ，so we can get all default values easily

Fixes #

```release-note
NONE
```

### DIFF
--- a/cmd/agent/app/options/options.go
+++ b/cmd/agent/app/options/options.go
@@ -11,22 +11,10 @@ import (
 	componentbaseconfig "k8s.io/component-base/config"
 
 	"github.com/karmada-io/karmada/pkg/features"
+	"github.com/karmada-io/karmada/pkg/karmadactl/options"
 	"github.com/karmada-io/karmada/pkg/sharedcli/profileflag"
 	"github.com/karmada-io/karmada/pkg/sharedcli/ratelimiterflag"
 	"github.com/karmada-io/karmada/pkg/util"
-)
-
-const (
-	// DefaultKarmadaClusterNamespace defines the default namespace where the member cluster secrets are stored.
-	DefaultKarmadaClusterNamespace = "karmada-cluster"
-	defaultBindAddress             = "0.0.0.0"
-	defaultPort                    = 10357
-)
-
-var (
-	defaultElectionLeaseDuration = metav1.Duration{Duration: 15 * time.Second}
-	defaultElectionRenewDeadline = metav1.Duration{Duration: 10 * time.Second}
-	defaultElectionRetryPeriod   = metav1.Duration{Duration: 2 * time.Second}
 )
 
 // Options contains everything necessary to create and run controller-manager.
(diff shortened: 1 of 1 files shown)


XiShanYongYe-Chang:
Hi @Zhuzhenghao Welcome to join the karmada. You can sign off refer to https://github.com/karmada-io/karmada/pull/2878/checks?check_run_id=9742553922

Zhuzhenghao:
@XiShanYongYe-Chang Done. Can you approve running workflows?

Zhuzhenghao:
@XiShanYongYe-Chang I think this change is reasonable.

FYI: https://github.com/kubernetes/kubernetes/blob/master/cmd/kube-apiserver/app/options/options.go

XiShanYongYe-Chang (commented):
/cc @lonelyCZ @RainbowMango

lonelyCZ:
Look good. We can find default options quickly by setting them in `NewOption` and make the code clearer. I also see this usage in other packages.

cmd/agent/app/options/options.go
```diff
@@ -127,11 +115,37 @@ type Options struct {
...
 		LeaderElection: componentbaseconfig.LeaderElectionConfiguration{
 			LeaderElect:       true,
 			ResourceLock:      resourcelock.LeasesResourceLock,
 			ResourceNamespace: util.NamespaceKarmadaSystem,
+			LeaseDuration:     metav1.Duration{Duration: 15 * time.Second},
+			RenewDeadline:     metav1.Duration{Duration: 10 * time.Second},
+			RetryPeriod:       metav1.Duration{Duration: 2 * time.Second},
 		},
+		BindAddress:                        "0.0.0.0",
+		SecurePort:                         10357,
```
lonelyCZ, line 128: Perhaps, we can rank them by ordered alphabets.

Zhuzhenghao: ok

lonelyCZ:
Thanks @Zhuzhenghao 

/lgtm

carlory:
cc @RainbowMango

RainbowMango:
> I move all default options to function NewOptions ，so we can get all default values easily

Do we have trouble getting default values?

Sorry, I can't see many benefits from it. In addition,
> fs.StringVar(&o.BindAddress, "bind-address", o.BindAddress,

From this, I feel confused about taking `o.BindAddress` as the default value.

carlory:
@RainbowMango 

In karmada, we can see the setting of the default value in two places, one in the NewOptions function, one in AddFlags, this is very confusing, but also to maintain the consistency of the two settings. 

We should only Set the default value in the NewOptions function, and set the parameter name and parameter description in AddFlags. This will also benefit the reading of the code. The name of the parameter can usually be seen from the field name. Compared with the parameter name, More often than not, we care more about the setting of parameters, and NewOptions is easier to retrieve and maintain than AddFlags.

FYI (kubernetes): 

https://github.com/kubernetes/kubernetes/blob/master/cmd/kube-apiserver/app/options/options.go#L93

FYI (karmada): 

- https://github.com/karmada-io/karmada/blob/HEAD/cmd/scheduler/app/options/options.go#L74
- https://github.com/karmada-io/karmada/blob/HEAD/cmd/scheduler/app/options/options.go#L90

lonelyCZ:
I just found that it seems not convenient for new code readers. If they want to know the meanings of options, they still need to see the AddFlags.

Currently, we don't set more default values in NewOptions() that seems only  for initialization, it seems ok to maintain current uniformal usages that put the defalut in AddFlags().

RainbowMango:
> I just found that it seems not convenient for new code readers. If they want to know the meanings of options, they still need to see the AddFlags.

Yes, that's exactly my concern at https://github.com/karmada-io/karmada/pull/2878#issuecomment-1358742669.

Zhuzhenghao:
Thank you for your feedback on my pull request. I understand the reason for refusal and agree with the decision.Thank you for your help and guidance.

PR #2883, opened 2022-11-29
add metrics for pool

ikaven1024:

**What type of PR is this?**
/kind cleanup

Add metrics for `FixedPool`, helps us to choose an optimal pool size.

Fixes #

```release-note
NONE
```

### DIFF
--- a/cmd/controller-manager/app/controllermanager.go
+++ b/cmd/controller-manager/app/controllermanager.go
@@ -158,6 +158,7 @@ func Run(ctx context.Context, opts *options.Options) error {
 
 	crtlmetrics.Registry.MustRegister(metrics.ClusterCollectors()...)
 	crtlmetrics.Registry.MustRegister(metrics.ResourceCollectors()...)
+	crtlmetrics.Registry.MustRegister(metrics.PoolCollectors()...)
 
 	setupControllers(controllerManager, opts, ctx.Done())
 
--- a/pkg/metrics/pool.go
+++ b/pkg/metrics/pool.go
@@ -1,0 +1,48 @@
+package metrics
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	poolGetCounterMetricsName = "pool_get_operation_total"
+	poolPutCounterMetricsName = "pool_put_operation_total"
+)
+
+var (
+	poolGetCounter = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: poolGetCounterMetricsName,
+		Help: "Total times of getting from pool",
+	}, []string{"name", "from"})
+
+	poolPutCounter = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: poolPutCounterMetricsName,
+		Help: "Total times of putting from pool",
+	}, []string{"name", "to"})
+)
+
+// RecordPoolGet records the times of getting from pool
+func RecordPoolGet(name string, created bool) {
+	from := "pool"
+	if created {
+		from = "new"
+	}
+	poolGetCounter.WithLabelValues(name, from).Inc()
+}
+
+// RecordPoolPut records the times of putting from pool
+func RecordPoolPut(name string, destroyed bool) {
+	to := "pool"
+	if destroyed {
+		to = "destroyed"
+	}
+	poolPutCounter.WithLabelValues(name, to).Inc()
+}
+
+// PoolCollectors returns the collectors about pool.
+func PoolCollectors() []prometheus.Collector {
+	return []prometheus.Collector{
+		poolGetCounter,
+		poolPutCounter,
+	}
+}
--- a/pkg/util/fixedpool/fixedpool.go
+++ b/pkg/util/fixedpool/fixedpool.go
@@ -2,11 +2,14 @@ package fixedpool
 
 import (
 	"sync"
+
+	"github.com/karmada-io/karmada/pkg/metrics"
 )
 
 // A FixedPool like sync.Pool. But it's limited capacity.
 // When pool is full, Put will abandon and call destroyFunc to destroy the object.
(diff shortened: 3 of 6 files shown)


pkg/metrics/pool.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+const (
+	poolGetCounterMetricsName = "pool_get_total"
+	poolPutCounterMetricsName = "pool_put_total"
+	poolSizeMetricsName       = "pool_put_size"
+)
+
+var (
+	poolGetCounter = prometheus.NewCounterVec(prometheus.CounterOpts{
```
Poor12, line 14: We should add UT for metrics exclude histogram.

cmd/controller-manager/app/controllermanager.go
```diff
@@ -158,6 +158,7 @@ func Run(ctx context.Context, opts *options.Options) error {
 
 	crtlmetrics.Registry.MustRegister(metrics.ClusterCollectors()...)
 	crtlmetrics.Registry.MustRegister(metrics.ResourceCollectors()...)
+	crtlmetrics.Registry.MustRegister(metrics.PoolCollectors()...)
```
Poor12, line 161: As the number of collectors increases, I think a general `initCollectors` method can be added.

ikaven1024: Greate idea. I think we'd better do it in another PR.

pkg/metrics/pool.go
ikaven1024: Added

XiShanYongYe-Chang (commented):
/assign @Poor12

pkg/metrics/pool.go
```diff
@@ -0,0 +1,60 @@
+package metrics
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	poolGetCounterMetricsName = "pool_get_total"
```
Poor12, line 8: BTW, I found that metrics name is not specific. Users cannnot get it directly.
1. what is pool used to do?
2. get or put what?

ikaven1024: It records the times of `Get` and `Put` method called. Do you have better suggestion?

XiShanYongYe-Chang: How about adding a `lua_vm` prefix?

ikaven1024: This pool can be used not only in lua vm. And `luavm` is added in the pool name.
```
pool_get_total{from="pool",name="luavm",type="fixedpool"} 1
```

pkg/util/fixedpool/fixedpool.go
```diff
@@ -2,11 +2,18 @@ package fixedpool
 
 import (
 	"sync"
+
+	"github.com/karmada-io/karmada/pkg/metrics"
+)
+
+const (
+	poolType = "fixedPool"
```
XiShanYongYe-Chang, line 10: What dose this type mean? Are there any other values of types? Is this a necessary value?

pkg/metrics/pool.go
Poor12: I think it's hard to use one metrics covering multiple scenes. For example, the labels that need to be used will not be exactly the same.

pkg/util/fixedpool/fixedpool.go
ikaven1024: Maybe we have anther pool type. For example we may use a `dynamic pool` in the future. It will scale the pool capacity according the history of pool size. So here use this label to distinguish.

pkg/metrics/pool.go
ikaven1024: > I think it's hard to use one metrics covering multiple scenes. For example, the labels that need to be used will not be exactly the same.

This metric is recording the operation of pool, not the luavm. This metric covers all the operations of pool, I think it's enough.

pkg/util/fixedpool/fixedpool.go
XiShanYongYe-Chang: But the pool we're using is the `FixedPool`.

ikaven1024: I mean all the pool use this metric. As thus, when it changes to another type of pool, monitor server (such as grafana) don't need to change.

pkg/metrics/pool.go
ikaven1024: How about `pool_get_operation_total`?

pkg/metrics/pool.go
```diff
@@ -0,0 +1,60 @@
...
+	poolPutCounterMetricsName = "pool_put_total"
+	poolSizeMetricsName       = "pool_size"
```
XiShanYongYe-Chang, line 10: I don't understand the function of this metric. I think we can judge whether a pool is sufficient based on the first two indicators,

pkg/metrics/pool.go
XiShanYongYe-Chang: @Poor12 How do you think?

pkg/metrics/pool.go
ikaven1024: Um, it's redundant. I will remove it.

pkg/util/fixedpool/fixedpool.go
XiShanYongYe-Chang: How about adding this field after a new type of pool is designed?

ikaven1024: ok

RainbowMango (approved):
/approve

Poor12:
BTW, could you please update the release note and document https://karmada.io/docs/next/reference/instrumentation/metrics here in the same time?

RainbowMango:
+1

Issue #2862, opened 2022-11-24
Add labels on the namespace created by karmada


Rains6:
**What happened**:
When I add a cluster to karmada, karmada will create a namespace on my cluster. I hope there will be a label indicating that the namespace was created by karmada to avoid accidental deletion.

**What you expected to happen**:
There will be a label indicating that the namespace was created by karmada to avoid accidental deletion.

**How to reproduce it (as minimally and precisely as possible)**:
Add a cluster to karmada 

**Anything else we need to know?**:
none

**Environment**:
- Karmada version: v1.3.0

![image](https://user-images.githubusercontent.com/49423798/203715857-b02171da-8091-44ea-bdd4-b1b495051864.png)

Poor12:
I wonder if you want a label for ns which stored the secret of the cluster or all the ns that karmada created, such as `karmada-es-member1`, `karmada-system`. This question is a little similar to https://github.com/karmada-io/karmada/issues/1350.

RainbowMango:
I've tested it on my side:
```yaml
# kubectl get ns karmada-cluster -o yaml 
apiVersion: v1
kind: Namespace
metadata:
  creationTimestamp: "2022-11-15T07:16:00Z"
  labels:
    kubernetes.io/metadata.name: karmada-cluster
  name: karmada-cluster
  resourceVersion: "966"
  uid: 95bf6df1-a5c3-481a-a0b8-71ff1b20d3dc
spec:
  finalizers:
  - kubernetes
status:
  phase: Active

```

I wonder who and where added this label `kubernetes.io/metadata.name: karmada-cluster`?

RainbowMango:
I get it. The label is added by Kubernetes.

> There will be a label indicating that the namespace was created by karmada

That makes sense, but if we don't use it for searching, we can introduce an annotation. Can you propose one?

Rains6:
> I get it. The label is added by Kubernetes.
> 
> > There will be a label indicating that the namespace was created by karmada
> 
> That makes sense, but if we don't use it for searching, we can introduce an annotation. Can you propose one?

For example, karmada.io/metadata.name: karmada-cluster ?

Zhuzhenghao:
@RainbowMango Can I add  karmada.io/metadata.name: karmada-cluster  to namespace?

RainbowMango:
Thanks for your contribution. I see the #2888.
But for now, I don't think the name(`karmada.io/metadata.name`) is the final decision yet.
[...]

PR #2888, opened 2022-11-30
Add annotation on the namespace created by karmada

Zhuzhenghao:
Signed-off-by: zhuzhenghao <zhenghao.zhu@daocloud.io>

**What type of PR is this?**
/kind feature

There will be a label indicating that the namespace was created by karmada to avoid accidental deletion.

Fixes #
https://github.com/karmada-io/karmada/issues/2862

```release-note
NONE
```

### DIFF
--- a/pkg/apis/work/v1alpha1/well_known_constants.go
+++ b/pkg/apis/work/v1alpha1/well_known_constants.go
@@ -15,4 +15,8 @@ const (
 
 	// WorkNameLabel is added to objects to specify associated Work's name.
 	WorkNameLabel = "work.karmada.io/name"
+
+	// AnnotationMetadataName is the annotation name which is used to automatically annotation namespaces
+	// so they can indicate that the namespace was created by karmada to avoid accidental deletion.
+	AnnotationMetadataName = "karmada.io/metadata.name"
 )
--- a/pkg/controllers/namespace/namespace_sync_controller.go
+++ b/pkg/controllers/namespace/namespace_sync_controller.go
@@ -144,6 +144,7 @@ func (c *Controller) buildWorks(namespace *corev1.Namespace, clusters []clusterv
 
 			util.MergeLabel(clonedNamespaced, workv1alpha1.WorkNamespaceLabel, workNamespace)
 			util.MergeLabel(clonedNamespaced, workv1alpha1.WorkNameLabel, workName)
+			util.MergeAnnotation(clonedNamespaced, workv1alpha1.AnnotationMetadataName, namespaceObj.GetName())
 
 			if err = helper.CreateOrUpdateWork(c.Client, objectMeta, clonedNamespaced); err != nil {
 				ch <- fmt.Errorf("sync namespace(%s) to cluster(%s) failed due to: %v", clonedNamespaced.GetName(), cluster.GetName(), err)


Zhuzhenghao:
TEST:
A namespace named demo already exists on cluster A and not exists on cluster B.
When I add a namespace which is named demo to karmada, karmada will create a namespace on cluster B with an annotation which is karmada.io/metadata.name: demo.

`
apiVersion: v1
kind: Namespace
metadata:
  name: demo
  uid: f984bc88-90ab-4ed8-8d68-df81ca265a0d
  resourceVersion: '6480'
  creationTimestamp: '2022-11-30T06:52:39Z'
  labels:
    kubernetes.io/metadata.name: demo
    work.karmada.io/name: demo-6654f4c4f6
    work.karmada.io/namespace: karmada-es-member3
  annotations:
    karmada.io/metadata.name: demo
    resourcetemplate.karmada.io/uid: 2b909830-1ea4-476b-a5b6-984df43d54e8
  managedFields:
    - manager: karmada-agent
      operation: Update
      apiVersion: v1
      time: '2022-11-30T06:52:39Z'
      fieldsType: FieldsV1
      fieldsV1:
        f:metadata:
          f:annotations:
            .: {}
            f:karmada.io/metadata.name: {}
            f:resourcetemplate.karmada.io/uid: {}
          f:labels:
            .: {}
            f:kubernetes.io/metadata.name: {}
            f:work.karmada.io/name: {}
            f:work.karmada.io/namespace: {}
        f:spec:
          f:finalizers: {}
  selfLink: /api/v1/namespaces/demo
status:
  phase: Active
spec:
  finalizers:
    - kubernetes
`

By the way，there is no changes in namespace demo which is on cluster A.

jwcesign:
How about `app.kubernetes.io/name: karmada`? (Refer to other google project)
Karmada like a app of kubenetes.

jwcesign:
cc @RainbowMango , maybe let's push this forward?

jwcesign:
I am thinking add a unified annotation or label to indicate all the resource created by Karmada.

1. For none-system-ns, I see there is labels: `work.karmada.io/name: xxx` and `work.karmada.io/namespace: xxx`.
2. For system-ns, like `karmada-cluster`, it hasn't any labels or annotations.
3. For workloads, like deployment, it also has labels: `work.karmada.io/name: xxx` and `work.karmada.io/namespace: xxx`
deployment

I think for 1/3, the labels are already enough to indicate Karmada created it(or need more info like the annotation here?).
For 2, we need a label/annotation to indicate.

What do you think? @XiShanYongYe-Chang

XiShanYongYe-Chang:
Yes, we just have to deal with scenario two.

jwcesign:
Referring to: https://kubernetes.io/zh-cn/docs/reference/labels-annotations-taints/#app-kubernetes-io-manged-by
I prefer to use annotation `app.kubernetes.io/managed-by: "karmada"`.

cc @RainbowMango, how about this solution? only add for scenario two.

RainbowMango:
I agree that this PR can focus on `case 2`, but I don't think `work.karmada.io/name` should take the responsibility to present a resource managed by Karmada.

We need an annotation for resources, `in member clusters`, to represent they are `managed` by Karmada.

We also need an annotation for resources, `in host cluster` to represent they are `created or used` by Karmada.

PR #2898, opened 2022-12-01
Remove redundant pointer use of atomic types

RainbowMango:
**What type of PR is this?**

/kind cleanup

This PR cleanup redundant pointer usage of atomic types for both `atomic.Value` and `atomic.Bool`,
they are safe to use the zero value.

echo from `atomic.Value`'s comment:
> The zero value for a Value returns nil from Load.

echo from `atomic.Bool`'s comment:
> The zero value is false.

Fixes #

```release-note
NONE
```


XiShanYongYe-Chang (commented):
/lgtm
/approve

PR #2899, opened 2022-12-02
clean up metrics for scheduler, descheduler and estimator

Garrybest:

**What type of PR is this?**
/kind cleanup
/kind feature

Note that scheduler, descheduler and estimator does not use `controller-runtime`, so we'd better use metrics pkg of `k8s.io/component-base/metrics` instead of primitive prometheus pkg. In kubernetes, all `workqueue` metrics is registered in `k8s.io/component-base/metrics`, so that we could get extra `workqueue` metrics in our metric http handler.

Fixes #

```release-note
NONE
```


Garrybest:
There is something wrong with the e2e environment again.

RainbowMango:
> There is something wrong with the e2e environment again.

Yeah. Just re-run it. 

echo from https://github.com/karmada-io/karmada/actions/runs/3603267239/jobs/6071407672:
> failed to delete nodes: command "docker rm -f -v member-e2e-4hc7h-control-plane" failed with error: exit status 1

RainbowMango:
All Kubernetes metrics are emitted from [Metrics Stability Framework](https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/1209-metrics-stability), and each metric will be assigned a `stable` status, like `ALPHA`/`BETA`/`STABLE`. Following is the example grabbed from `karmada-scheduler`:

```
# HELP apiserver_audit_event_total [ALPHA] Counter of audit events generated and sent to the audit backend.
# TYPE apiserver_audit_event_total counter
apiserver_audit_event_total 0
```

> so we'd better use metrics pkg of k8s.io/component-base/metrics instead of primitive prometheus pkg. In kubernetes, all workqueue metrics is registered in k8s.io/component-base/metrics, so that we could get extra workqueue metrics in our metric http handler.

Do you want to get more metrics from the `k8s.io/component-base/metrics` package or want to adopt the `Metrics Stability Framework`?

Garrybest:
> Do you want to get more metrics from the k8s.io/component-base/metrics package or want to adopt the Metrics Stability Framework?

Yes, now we cannot get the metrics of workqueue in scheduler. This is important, I think. The pkg `k8s.io/component-base/metrics` could help us report metrics automatically.

RainbowMango:
Using `k8s.io/component-base/metrics` means adopting the [Metrics Stability Framework](https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/1209-metrics-stability), I'm not sure if the framework ready for use by other projects yet. I'll figure it out.

Garrybest:
Thanks for reminding, I believe I should find another way to report the workqueue metrics.

Garrybest:
I'm trying to set all metrics with registry in `controller-runtime`. Now we could report the internal workqueue metrics. This is like what components using `controller-runtime` does, i.e., karmada-agent and karmada-controller-runtime.

Garrybest:
This lint should be ignored since we must insure `sigs.k8s.io/controller-runtime/pkg/metrics` is imported before `k8s.io/apiserver/pkg/server`.

RainbowMango:
cc @Poor12 for help.

Poor12:
It works well at my side. It mainly solves this problem by uniformly registering indicators to the controller-runtime registry.
If lint should be ignored, we can add a comment to skip it.

RainbowMango:
Hmm, I'm ok with that disable the checks temporarily.
Can we make a list of the changes that should be made against the upstream(controller-runtime, k/k) to get rid of the exceptions?

Garrybest:
That's right. So there are 2 things we need to do:

1. Register all metrics to `ctrlmetrics.Registry` instead of default registry in `component-base/metrics` or `promhttp`. We only need one registry.
2. Initialize pkg `sigs.k8s.io/controller-runtime/pkg/metrics` before `k8s.io/apiserver/pkg/server` because workqueue metric could only be registered to one registry which is initialized first.

I think this PR fix these two steps.

RainbowMango (approved):
/lgtm
/approve

Issue #2906, opened 2022-12-06
Should a label mismatch in PropagationPolicy placement delete resources from member clusters?


cebernardi:
**Please provide an in-depth description of the question you have**:
If a PropagationPolicy targets a set of member clusters by labels:

```
  placement:
    clusterAffinity:
      labelSelector:
        matchLabels:
          phase: setup
```

and a the label value is changed on the cluster resource:

```
kubectl --context=karmada-api-server label cluster my-cluster phase=ready
```
the resources targeted by the `PropagationPolicy` are not removed from the member cluster `my-cluster`.

Is this the expected behavior or am I missing something?

**What do you think about this question?**:
I was expecting that Karmada would reconcile the placement and remove the resource from the member cluster that doesn't match the placement anymore.

I'm aware of [this](https://karmada.io/docs/userguide/scheduling/resource-propagating#delete-a-propagationpolicy), but it feels to me that labels placement mismatch is a different scenario.

**Environment**:
- Karmada version: 1.4.0
- Kubernetes version: 1.24
- Others:

XiShanYongYe-Chang:
@cebernardi Thank you for your discovery. This is a stable reproduction of the problem. The reason is that label deletion is not considered when the cluster update event is processed:

https://github.com/karmada-io/karmada/blob/e6ba531a739668af1459e4f7d4c2a5cd80fb163e/pkg/scheduler/event_handler.go#L228-L234

We will fix it as soon as possible.

/kind bug

RainbowMango:
/priority important-soon
@cebernardi Thanks for spotting and sorry for the inconvenience. 

This issue will be fixed on the master branch, and a cherry-pick will be applied to release-1.4 after that. So, the patch will be available in v1.4.1.

RainbowMango:
@cebernardi 
Just a notice, we just fixed this issue on the master branch, and we are going to tag a new patch release for release-1.4 this week. So, this patch should be available in v1.4.1.

cebernardi:
Thank you, trying it asap!

cebernardi:
thanks, works like a charm!

RainbowMango:
Hi @cebernardi, It's been a while since the last time I heard from you. Hope you are doing well :)
Just wanted to check in: Are you using Karmada now? If so, have you run into any further issues or any feedback for us?

PS: I'm currently looking into another related issue and traced it back to this thread.

Issue #2261, opened 2022-07-26
Karmada not propagate resources to new member cluster


liuchintao:
**What happened**:

I just want to propagate my deployment to all member clusters automatically, when they join.

```yaml
apiVersion: policy.karmada.io/v1alpha1
kind: PropagationPolicy
metadata:
  name: nginx
spec:
  resourceSelectors:
  - apiVersion: apps/v1
    kind: Deployment
    name: nginx
```

After creating this PP, I find nginx deployment in cluster-1 and cluster-2.

But when I join a new member cluster-3, I find nginx deployment is not propagated to cluster-3

**What you expected to happen**:

Deployment should be propagated to new member cluster.

```golang
type Placement struct {
	// ClusterAffinity represents scheduling restrictions to a certain set of clusters.
	// If not set, any cluster can be scheduling candidate.
	// +optional
	ClusterAffinity *ClusterAffinity `json:"clusterAffinity,omitempty"`
}
```

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Karmada version:
- kubectl-karmada or karmadactl version (the result of `kubectl-karmada version` or `karmadactl version`):
- Others:

RainbowMango:
> I just want to propagate my deployment to all member clusters automatically, when they join.

Yes, I think this is a reasonable use case. But now the scheduler doesn't re-schedule in case of cluster joining or removal.

@dddddai I can see there is `TODO`, do you mean to cover this case?
https://github.com/karmada-io/karmada/blob/ed9b838056ee45720abf51b2d603548e6ce922bf/pkg/scheduler/scheduler.go#L384

dddddai:
Yeah I was thinking so, there were similar issues
https://github.com/karmada-io/karmada/issues/1644
https://github.com/karmada-io/karmada/issues/829#issuecomment-1107735202

I think the descheduler should be responsible for this case?

RainbowMango:
> I think the descheduler should be responsible for this case?

cc @Garrybest 

I'll pay attention to this feature. @dddddai can you help to lead the effort?

Garrybest:
Descheduler's duty is to evict replicas only. This issue focuses on scheduling a object to new joined clusters.

I think we could always `s.scheduleResourceBinding(rb)` when the placement is `Duplicated`. It is ok?

liuchintao:
hello, is there anything new?

RainbowMango:
[...]

PR #2912, opened 2022-12-07
fix a corner case that re-schedule be skipped in case of the cluster becomes not fit.

jwcesign:

**What type of PR is this?**
/kind bug

* Fix resource hangover in member cluster when propagating missmatch

Fixes https://github.com/karmada-io/karmada/issues/2906

none

```release-note
`karmada-scheduler`: Fixed a corner case that re-schedule be skipped in case of the cluster becomes not fit.
```

### DIFF
--- a/pkg/scheduler/core/generic_scheduler.go
+++ b/pkg/scheduler/core/generic_scheduler.go
@@ -63,6 +63,7 @@ func (g *genericScheduler) Schedule(ctx context.Context, placement *policyv1alph
 		return result, fmt.Errorf("failed to findClustersThatFit: %v", err)
 	}
 
+	// Short path for case no cluster fit.
 	if len(feasibleClusters) == 0 {
 		return result, &framework.FitError{
 			NumAllClusters: clusterInfoSnapshot.NumOfClusters(),
--- a/pkg/scheduler/event_handler.go
+++ b/pkg/scheduler/event_handler.go
@@ -229,13 +229,13 @@ func (s *Scheduler) updateCluster(oldObj, newObj interface{}) {
 	case !equality.Semantic.DeepEqual(oldCluster.Labels, newCluster.Labels):
 		fallthrough
 	case !equality.Semantic.DeepEqual(oldCluster.Spec, newCluster.Spec):
-		s.enqueueAffectedPolicy(newCluster)
-		s.enqueueAffectedClusterPolicy(newCluster)
+		s.enqueueAffectedPolicy(oldCluster, newCluster)
+		s.enqueueAffectedClusterPolicy(oldCluster, newCluster)
 	}
 }
 
 // enqueueAffectedPolicy find all propagation policies related to the cluster and reschedule the RBs
-func (s *Scheduler) enqueueAffectedPolicy(newCluster *clusterv1alpha1.Cluster) {
+func (s *Scheduler) enqueueAffectedPolicy(oldCluster, newCluster *clusterv1alpha1.Cluster) {
 	policies, _ := s.policyLister.List(labels.Everything())
 	for _, policy := range policies {
 		selector := labels.SelectorFromSet(labels.Set{
@@ -245,10 +245,13 @@ func (s *Scheduler) enqueueAffectedPolicy(newCluster *clusterv1alpha1.Cluster) {
 		affinity := policy.Spec.Placement.ClusterAffinity
 		switch {
 		case affinity == nil:
-			// If no clusters specified, add it in queue
+			// If no clusters specified, add it to the queue
 			fallthrough
 		case util.ClusterMatches(newCluster, *affinity):
-			// If specific cluster matches the affinity. add it in queue
+			// If the new cluster manifest match the affinity, add it to the queue, trigger rescheduling
+			fallthrough
+		case util.ClusterMatches(oldCluster, *affinity):
+			// If the old cluster manifest match the affinity, add it to the queue, trigger rescheduling
 			err := s.requeueResourceBindings(selector, metrics.ClusterChanged)
 			if err != nil {
 				klog.Errorf("Failed to requeue ResourceBinding, error: %v", err)
@@ -258,7 +261,7 @@ func (s *Scheduler) enqueueAffectedPolicy(newCluster *clusterv1alpha1.Cluster) {
 }
 
 // enqueueAffectedClusterPolicy find all cluster propagation policies related to the cluster and reschedule the RBs/CRBs
-func (s *Scheduler) enqueueAffectedClusterPolicy(newCluster *clusterv1alpha1.Cluster) {
+func (s *Scheduler) enqueueAffectedClusterPolicy(oldCluster, newCluster *clusterv1alpha1.Cluster) {
 	clusterPolicies, _ := s.clusterPolicyLister.List(labels.Everything())
 	for _, policy := range clusterPolicies {
 		selector := labels.SelectorFromSet(labels.Set{
@@ -267,10 +270,13 @@ func (s *Scheduler) enqueueAffectedClusterPolicy(newCluster *clusterv1alpha1.Clu
 		affinity := policy.Spec.Placement.ClusterAffinity
 		switch {
 		case affinity == nil:
-			// If no clusters specified, add it in queue
+			// If no clusters specified, add it to the queue
 			fallthrough
 		case util.ClusterMatches(newCluster, *affinity):
-			// If specific cluster matches the affinity. add it in queue
+			// If the new cluster manifest match the affinity, add it to the queue, trigger rescheduling
+			fallthrough
+		case util.ClusterMatches(oldCluster, *affinity):
+			// If the old cluster manifest match the affinity, add it to the queue, trigger rescheduling
 			err := s.requeueClusterResourceBindings(selector, metrics.ClusterChanged)
 			if err != nil {
 				klog.Errorf("Failed to requeue ClusterResourceBinding, error: %v", err)
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -3,6 +3,7 @@ package scheduler
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"reflect"
 	"time"
@@ -13,6 +14,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
@@ -33,6 +35,7 @@ import (
 	worklister "github.com/karmada-io/karmada/pkg/generated/listers/work/v1alpha2"
 	schedulercache "github.com/karmada-io/karmada/pkg/scheduler/cache"
 	"github.com/karmada-io/karmada/pkg/scheduler/core"
+	"github.com/karmada-io/karmada/pkg/scheduler/framework"
 	frameworkplugins "github.com/karmada-io/karmada/pkg/scheduler/framework/plugins"
 	"github.com/karmada-io/karmada/pkg/scheduler/framework/runtime"
 	"github.com/karmada-io/karmada/pkg/scheduler/metrics"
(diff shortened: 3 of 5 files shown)


jwcesign:
cc @XiShanYongYe-Chang

XiShanYongYe-Chang:
Thanks~
/assign

RainbowMango:
/retitle fix a corner case that re-schedule be skipped in case of the cluster becomes not fit.

jwcesign:
cc @chaunceyjiang , Please take a look, I listen there is other background.

pkg/scheduler/core/generic_scheduler.go
```diff
@@ -64,10 +64,9 @@ func (g *genericScheduler) Schedule(ctx context.Context, placement *policyv1alph
 	}
 
 	if len(feasibleClusters) == 0 {
-		return result, &framework.FitError{
-			NumAllClusters: clusterInfoSnapshot.NumOfClusters(),
-			Diagnosis:      diagnosis,
-		}
+		warning := &framework.FitError{NumAllClusters: clusterInfoSnapshot.NumOfClusters(), Diagnosis: diagnosis}
+		klog.Warningf("Couldn't finding feasible clusters: %v", warning)
```
XiShanYongYe-Chang, line 68: What is the log output form of this line? Do we need to use `warning.String()`?

test/e2e/propagationpolicy_test.go
```diff
@@ -939,3 +939,59 @@ var _ = ginkgo.Describe("[AdvancedPropagation] propagation testing", func() {
 		})
 	})
 })
+
+// LabeChangePropagation does propagation testing when label changes.
+var _ = ginkgo.Describe("[LabeChangePropagation] propagation testing", func() {
```
XiShanYongYe-Chang, line 944: How about moving this test into `rescheduling_test.go`?

pkg/scheduler/core/generic_scheduler.go
jwcesign: The outlog is:
```
W1207 08:46:59.576229       1 generic_scheduler.go:68] Couldn't finding feasible clusters: 0/3 clusters are available: 3 cluster(s) didn't match the placement cluster affinity constraint.
```

XiShanYongYe-Chang: ok, thx

test/e2e/framework/cluster.go
```diff
@@ -230,6 +230,44 @@ func setClusterLabel(c client.Client, clusterName string) error {
...
 
+// UpdateClusterLabels update cluster labels.
+func UpdateClusterLabels(client karmada.Interface, clusterName string, labels map[string]string) {
+	gomega.Eventually(func() (bool, error) {
+		cluster, err := client.ClusterV1alpha1().Clusters().Get(context.TODO(), clusterName, metav1.GetOptions{})
+		if err != nil {
+			return false, err
+		}
+
+		for key, value := range labels {
```
XiShanYongYe-Chang, line 241: Maybe here need to check whether the `cluster.Labels` is nil first.

test/e2e/rescheduling_test.go
```diff
@@ -319,3 +319,114 @@ var _ = ginkgo.Describe("[cluster joined] reschedule testing", func() {
...
+			labelsUpdate = map[string]string{"cluster": "ok"}
+			framework.UpdateClusterLabels(karmadaClient, targetMember, labelsUpdate)
+			framework.WaitDeploymentDisappearOnClusters([]string{targetMember}, deployment.Namespace, deployment.Name)
+		})
+	})
+})
+
+// LabeChangePropagation does propagation testing when label changes.
+var _ = ginkgo.Describe("[LabeChangePropagation] reschedule testing while cpp matches", func() {
+	ginkgo.Context("Changes cluster labels to test reschedule while cpp matches", func() {
```
XiShanYongYe-Chang, line 381: We can move the `Context` case to the previous `Describe`.

pkg/scheduler/event_handler.go
```diff
@@ -247,7 +247,7 @@ func (s *Scheduler) enqueueAffectedPolicy(newCluster *clusterv1alpha1.Cluster) {
 		case affinity == nil:
 			// If no clusters specified, add it in queue
 			fallthrough
-		case util.ClusterMatches(newCluster, *affinity):
+		case util.ClusterMatches(oldCluster, *affinity), util.ClusterMatches(newCluster, *affinity):
```
RainbowMango, line 250: Why not use another `fallthrough`?

pkg/scheduler/core/generic_scheduler.go
```diff
@@ -64,10 +64,9 @@ func (g *genericScheduler) Schedule(ctx context.Context, placement *policyv1alph
 	}
 
 	if len(feasibleClusters) == 0 {
-		return result, &framework.FitError{
-			NumAllClusters: clusterInfoSnapshot.NumOfClusters(),
-			Diagnosis:      diagnosis,
-		}
+		warning := &framework.FitError{NumAllClusters: clusterInfoSnapshot.NumOfClusters(), Diagnosis: diagnosis}
```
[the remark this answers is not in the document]
RainbowMango: agree, we can remove the `FitEroor`.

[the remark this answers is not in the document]
RainbowMango: Please add a comment ahead of `len(feasibleClusters)`, like:
```golang
	// Short path for case no cluster fit.
	// No cluster fit also means success from the scheduler's point of view.
	if len(feasibleClusters) == 0 {
```

chaunceyjiang:
Oh, I missed this pr.

> cc @chaunceyjiang , Please take a look, I've heard that this question has other context.

yes ,refer to https://github.com/karmada-io/karmada/issues/2261

pkg/scheduler/event_handler.go
chaunceyjiang: +1 
This is a new case.

test/e2e/rescheduling_test.go
```diff
@@ -319,3 +319,114 @@ var _ = ginkgo.Describe("[cluster joined] reschedule testing", func() {
 		})
 	})
 })
+
+// LabeChangeReschedule does reschedule testing when label changes.
+var _ = ginkgo.Describe("[LabeChangePropagation] reschedule testing while pp matches", func() {
```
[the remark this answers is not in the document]
jwcesign: fixed

pkg/scheduler/event_handler.go
```diff
@@ -229,26 +229,21 @@ func (s *Scheduler) updateCluster(oldObj, newObj interface{}) {
...
 			policyv1alpha1.PropagationPolicyNameLabel:      policy.Name,
 		})
 		affinity := policy.Spec.Placement.ClusterAffinity
-		switch {
-		case affinity == nil:
-			// If no clusters specified, add it in queue
-			fallthrough
-		case util.ClusterMatches(newCluster, *affinity):
-			// If specific cluster matches the affinity. add it in queue
+		if affinity == nil || util.ClusterMatches(oldCluster, *affinity) || util.ClusterMatches(newCluster, *affinity) {
```
chaunceyjiang, line 246: 1. I think we need to add some comments.
2. I think we should match the newCluster first, then the oldCluster.

jwcesign: PTAL

test/e2e/framework/cluster.go
```diff
@@ -230,6 +230,50 @@ func setClusterLabel(c client.Client, clusterName string) error {
 	return err
 }
 
+// UpdateClusterLabels update cluster labels.
```
XiShanYongYe-Chang, line 233: update -> updates

test/e2e/framework/cluster.go
```diff
@@ -230,6 +230,50 @@ func setClusterLabel(c client.Client, clusterName string) error {
...
+		}
+		_, err = client.ClusterV1alpha1().Clusters().Update(context.TODO(), cluster, metav1.UpdateOptions{})
+		if err != nil {
+			return false, err
+		}
+		return true, nil
+	}, pollTimeout, pollInterval).Should(gomega.Equal(true))
+}
+
+// DeleteClusterLabels delete cluster labels if it exists.
```
XiShanYongYe-Chang, line 255: delete -> deletes

test/e2e/rescheduling_test.go
```diff
@@ -319,3 +319,112 @@ var _ = ginkgo.Describe("[cluster joined] reschedule testing", func() {
 		})
 	})
 })
+
+// reschedule testing while policy matches, triggered by label changes.
+var _ = ginkgo.Describe("[cluster labels changed] reschedule testing while policy matches", func() {
+	var deployment *appsv1.Deployment
+	var targetMember string
+	var labeKey string
```
XiShanYongYe-Chang, line 327: labeKey -> labelKey

XiShanYongYe-Chang (commented):
Just some typo.
Other parts LTGM

pkg/scheduler/core/generic_scheduler.go
```diff
@@ -58,16 +58,16 @@ func (g *genericScheduler) Schedule(ctx context.Context, placement *policyv1alph
 		return result, fmt.Errorf("no clusters available to schedule")
 	}
 
-	feasibleClusters, diagnosis, err := g.findClustersThatFit(ctx, g.scheduleFramework, placement, spec, &clusterInfoSnapshot)
```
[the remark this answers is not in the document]
Garrybest: Why removing? Users are not able to know why they could not propagate their resources now and just receive an error event like 'Oh, we could not find a feasible cluster for you, please figure it out by yourself' 🤣. It's not user-friendly.

Actually, diagnosis is inspired by kube-scheduler which has a complete diagnosis detail that tells users why a bunch of nodes have no qualification to produce a pod.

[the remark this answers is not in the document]
RainbowMango: I agree with you, that we need to provide a way to notify users other than logging. 
But for now, the `diagnossis` is used only for logging, do we have any plans to improve it?

> Actually, diagnosis is inspired by kube-scheduler which has a complete diagnosis detail that tells users why a bunch of nodes have no qualification to produce a pod.

Can you show me a link about how kube-scheduler deals with the diagnosis?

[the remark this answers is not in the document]
Garrybest: > But for now, the diagnossis is used only for logging, do we have any plans to improve it?

Well, the diagnose is return as an error and scheduler record the error as a form of event. So users could receive a respective event for their workloads. Refer to https://github.com/karmada-io/karmada/blob/941637e805b267e00ca5bd2d979f98c4916be358/pkg/scheduler/scheduler.go#L357.

The event is like: `Warning ScheduleBindingFailed 2s (x10 over 4s) karmada-scheduler 0/3 clusters are available: 1 cluster(s) had untolerated taint {dedicated: special-user}, 2 cluster(s) didn't have the API resource.`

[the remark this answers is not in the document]
RainbowMango: Oh, thanks. I didn't notice it. Yes, you are right.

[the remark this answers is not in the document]
jwcesign: @Garrybest PTAL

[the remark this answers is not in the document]
Garrybest: OK, I will review ASAP.

pkg/scheduler/event_handler.go
```diff
@@ -245,10 +245,13 @@ func (s *Scheduler) enqueueAffectedPolicy(newCluster *clusterv1alpha1.Cluster) {
...
 		switch {
 		case affinity == nil:
-			// If no clusters specified, add it in queue
+			// If no clusters specified, add it to the queue
 			fallthrough
 		case util.ClusterMatches(newCluster, *affinity):
-			// If specific cluster matches the affinity. add it in queue
+			// If the new cluster manifest match the affinity, add it to the queue, trigger rescheduling
+			fallthrough
+		case util.ClusterMatches(oldCluster, *affinity):
```
Garrybest, line 253: I guess it's could be better to merge old and new cluster match.

```go
case util.ClusterMatches(oldCluster, *affinity) != util.ClusterMatches(newCluster, *affinity)
```

Am I right?

pkg/scheduler/scheduler.go
```diff
@@ -476,13 +478,14 @@ func (s *Scheduler) scheduleResourceBinding(resourceBinding *workv1alpha2.Resour
 	}
 
 	scheduleResult, err := s.Algorithm.Schedule(context.TODO(), &placement, &resourceBinding.Spec, &core.ScheduleAlgorithmOption{EnableEmptyWorkloadPropagation: s.enableEmptyWorkloadPropagation})
-	if err != nil {
+	if err != nil && reflect.TypeOf(err) != reflect.TypeOf(&framework.FitError{}) {
```
Garrybest, line 481: I'm afraid we have a situation here. When a cluster is not ready for a while, the cluster will be added a NoSchedule taint. And before, a workload is running on this cluster and happens to scale up(may be triggered by users or hpa-controller), however, no cluster is fit due to the taint. I guess the previous replicas will be evicted here because now fit error is skipped and result will be patched to RB.

pkg/scheduler/event_handler.go
jwcesign: I think it may not be right. It only triggers rescheduling when the affinity changes, but not when the taints are modified.

pkg/scheduler/scheduler.go
Garrybest: Oh, I see. I didn't realize taint plugin is updated here. Thanks.

RainbowMango:
@XiShanYongYe-Chang Have you confirmed the E2E part?

pkg/scheduler/scheduler.go
RainbowMango: Now the `FitError` will be dedicated used in the situation that `no cluster fit`, right?

XiShanYongYe-Chang:
> @XiShanYongYe-Chang Have you confirmed the E2E part?

Yes, I've confirmed it.

pkg/scheduler/scheduler.go
jwcesign: yes, then the `fitError` information will be shown in the event and RB's status, the same behavior as k8s.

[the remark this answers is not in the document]
jwcesign: fixed

jwcesign:
> Have you figured out the root cause of the failing tests? I didn't see any change from the force push.

it's my mistake, should be `if err != nil && !errors.As(err, &noClusterFit) {` instead of `if err != nil && errors.As(err, &noClusterFit) {`.

RainbowMango (approved):
/lgtm
/approve

Please cherry-pick this patch to release branches.
And we are going to tag a new release after that.

PR #2930, opened 2022-12-08
Fix work status not sync to control plane when apiserver failed and recovered soon

### LINKED ISSUE #2863: Workload not synced to control plane version
**What happened**:
When I send the deployment to multiple clusters, everything is ok. After one of the clusters fails and then recovers, I modify the delivered deployment on the cluster, and it is not synchronized back to the version of the karmada control plane.

I hope to directly modify the distributed deployment on the cluster, which can be automatically synchronized back to the version of the karmada control plane

Send the deployment to multiple clusters, After one of the clusters fails and then recovers, modify the delivered deployment on the cluster, and it is not synchronized back to the version of the karmada control plane.

**Environment**:
- Karmada version: v1.3.0

Poor12:

**What type of PR is this?**
/kind bug

When cluster status is unhealthy, we should not stop informer manager because when apiserver recovered, workload can not sync to karmada control plane.

Reproduction：
1.Failover is close.
2.Deploy a deployment to m1 and m2.
3.m1 apiserver failed. The cluster status of m1 became false. Soon it recovered. The cluster status of m1 became true again.
4.Update the deployment in m1. But the control plane cannot perceive the change of deployment in m1.

Reason:
Informer of m1 didn't work and could not collect update events.

Fixes #2863 

```release-note
karmada-controller-manager/karmada-agent: fix work status not sync to control plane when apiserver failed and recovered soon.
```

### DIFF
--- a/pkg/controllers/status/cluster_status_controller.go
+++ b/pkg/controllers/status/cluster_status_controller.go
@@ -177,8 +177,6 @@ func (c *ClusterStatusController) syncClusterStatus(cluster *clusterv1alpha1.Clu
 	if !online && readyCondition.Status != metav1.ConditionTrue {
 		klog.V(2).Infof("Cluster(%s) still offline after %s, ensuring offline is set.",
 			cluster.Name, c.ClusterFailureThreshold.Duration)
-		c.GenericInformerManager.Stop(cluster.Name)
-		c.TypedInformerManager.Stop(cluster.Name)
 		setTransitionTime(cluster.Status.Conditions, readyCondition)
 		meta.SetStatusCondition(&currentClusterStatus.Conditions, *readyCondition)
 		return c.updateStatusIfNeeded(cluster, currentClusterStatus)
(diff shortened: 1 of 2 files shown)


XiShanYongYe-Chang:
Good job, can we add an e2e test to cover this case?

test/e2e/failover_test.go
```diff
@@ -141,6 +141,117 @@ var _ = framework.SerialDescribe("failover testing", func() {
 			})
 		})
 	})
+
+	ginkgo.Context("Deployment status synchronization when cluster failed and recovered soon", func() {
```
XiShanYongYe-Chang, line 145: This e2e test doesn't involve failover. It's more like a test case for resource state collection. How about moving it to `resource_test.go`?

test/e2e/failover_test.go
```diff
@@ -141,6 +141,117 @@ var _ = framework.SerialDescribe("failover testing", func() {
...
+		ginkgo.BeforeEach(func() {
+			policyNamespace = testNamespace
+			policyName = deploymentNamePrefix + rand.String(RandomStrLength)
+			deploymentNamespace = testNamespace
+			deploymentName = policyName
+			deployment = testhelper.NewDeployment(deploymentNamespace, deploymentName)
+			numOfFailedClusters = 1
+			originalReplicas = 3
+
+			// set MaxGroups=MinGroups=1, label is location=CHN.
```
XiShanYongYe-Chang, line 161: This comment seems unrelated.

XiShanYongYe-Chang (commented):
/lgtm

PR #2936, opened 2022-12-09
Stop taint-manager when failover is close

Poor12:

**What type of PR is this?**
/kind cleanup

Stop taint-manager when failover is close. I delete one of the cluster which is propagated a workload.
Then cluster-controller will add a Terminating taint. Taint-manager watches the change and add a gracefulEvictionTask since gracefulEviction feature gate is default enabled. 

However, rb_graceful_eviction_controller is close when failover is close. So this gracefulEvictionTask will never be dealed.

Fixes #

None


XiShanYongYe-Chang (commented):
LGTM
Ask for @Garrybest  to help take a look.

/cc @Garrybest

Garrybest:
Actually, taint manager runs independently. Any other controller does not have an impact on it. When a user add a no-execute taint to a cluster, taint manager will try to remove the result from binding.

This issue arises from the coordination of `GracefulEviction` and its controller, which I suppose we should cope with.

XiShanYongYe-Chang:
> Actually, taint manager runs independently. Any other controller does not have an impact on it. When a user adds a no-execute taint to a cluster, taint manager will try to remove the result from binding.
> 
> This issue arises from the coordination of `GracefulEviction` and its controller, which I suppose we should cope with.

It makes sense. Perhaps we need to modify the judgment conditions for starting the `graceful-eviction-controller` and remove the judgment of failover feature gate:

https://github.com/karmada-io/karmada/blob/db4e2abf1a3b4d8f6fbd714d70848c5be683f85f/cmd/controller-manager/app/controllermanager.go#L475

Garrybest:
>  Perhaps we need to modify the judgment conditions for starting the graceful-eviction-controller and remove the judgment of failover feature gate.

+1

Poor12:
You are right. Taint-manager is mainly to deal with no-execute taints. And no-execute taints are mainly from 

1. Failover
2. Others, such as cluster removed and user added.

So gracefulEviction controller does not depend on the `Failover` feature gate, only cares about `GracefulEviction`.

RainbowMango (approved):
/approve

> Actually, taint manager runs independently. Any other controller does not have an impact on it.

+1

PR #2947, opened 2022-12-12
fix failing test of editing policy resourceSelectors

XiShanYongYe-Chang:

**What type of PR is this?**

/kind failing-test

We found such a failing test on the master branch:
https://github.com/karmada-io/karmada/actions/runs/3671907442/jobs/6207636506

After analysis, it is found that the primary `karmada-controller-manager` instance is faulty when the `resourceSelector` of `PropagationPolicy` is changed but not successfully processed during the test. After the standby instance takes over the work from the primary instance, the informer of the `detector` is rebuilt.

However, for the `detector` informer, the order fo `PropagationPolicy` is earlier than the order of `Deployment`. As a result, the related `Deployment` cannot be found during `PropagationPolicy` processing. So the labels clean is missing:

https://github.com/karmada-io/karmada/blob/37cca1cd8dabc62435a276c63c6460d35d97bbcd/pkg/detector/policy.go#L254-L256

Therefore, the purpose of this modification is to use the dynamic client to obtain the target object again when the target object cannot be found in the informer cache.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/detector/detector.go
+++ b/pkg/detector/detector.go
@@ -537,9 +537,17 @@ func (d *ResourceDetector) GetUnstructuredObject(objectKey keys.ClusterWideKey) 
 
 	object, err := d.InformerManager.Lister(objectGVR).Get(objectKey.NamespaceKey())
 	if err != nil {
-		if !apierrors.IsNotFound(err) {
-			klog.Errorf("Failed to get object(%s), error: %v", objectKey, err)
+		if apierrors.IsNotFound(err) {
+			// If the target object is not found in the informer cache,
+			// use the DynamicClient to get the target object again.
+			var object *unstructured.Unstructured
+			object, err = d.DynamicClient.Resource(objectGVR).Namespace(objectKey.Namespace).
+				Get(context.TODO(), objectKey.Name, metav1.GetOptions{})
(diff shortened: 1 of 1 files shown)


pkg/detector/detector.go
```diff
@@ -537,9 +537,20 @@ func (d *ResourceDetector) GetUnstructuredObject(objectKey keys.ClusterWideKey)
...
-			klog.Errorf("Failed to get object(%s), error: %v", objectKey, err)
+		if apierrors.IsNotFound(err) {
+			// If the target object is not found in the informer cache,
+			// use the DynamicClient to get the target object again.
+			object, err := d.DynamicClient.Resource(objectGVR).Namespace(objectKey.Namespace).
+				Get(context.TODO(), objectKey.Name, metav1.GetOptions{})
+			if err == nil {
+				return object, nil
+			}
+			if apierrors.IsNotFound(err) {
```
jwcesign, line 548: how about deleting this part and letting line 553 print the error?

XiShanYongYe-Chang: Update

RainbowMango (approved):
/lgtm
/approve

PR #2949, opened 2022-12-12
CommandUnjoinOption wait must be a positive duration

yanfeng1992:

**What type of PR is this?**
/kind bug

CommandUnjoinOption wait must be a positive duration， but now allow wait = 0
if wait = 0,  when use  func poller,  timeout=0

A timeout of 0 is interpreted as an infinity, and in such a case
it would be the caller's responsibility to close the done channel.
Failure to do so would result in a leaked goroutine.

image

Fixes #

```release-note
CommandUnjoinOption wait must be a positive duration
```


yanfeng1992:
@RainbowMango Does anyone have a look at this question? Or think it's not a problem

carlory:
sorry for too late reply.

/lgtm

lonelyCZ:
Thanks @yanfeng1992 

/approve

PR #2954, opened 2022-12-13
fix post-delete job of chart

calvin0327:

**What type of PR is this?**
/kind bug

if the resources associated with cert is failed to apply in pre-install job, and the whole process of installtion is failed, then while we delete the release of  karmada,  the err log is be throwed in post-delete job. due to those resource is not found in the cluster.

image

Fixes #

So, in the case, we should ignore err  when delete non-exist resource.

```release-note
NONE
```

### DIFF
--- a/charts/karmada/templates/post-delete-job.yaml
+++ b/charts/karmada/templates/post-delete-job.yaml
@@ -35,8 +35,8 @@ spec:
             - |
               bash <<'EOF'
               set -ex
-              kubectl delete -f /opt/mount/
-              kubectl delete cm/{{ $name }}-config -n {{ $namespace }}
+              kubectl delete -f /opt/mount/ --ignore-not-found=true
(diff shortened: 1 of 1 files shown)


charts/karmada/templates/post-delete-job.yaml
```diff
@@ -35,7 +35,7 @@ spec:
             - |
               bash <<'EOF'
               set -ex
-              kubectl delete -f /opt/mount/
+              kubectl delete -f /opt/mount/ --ignore-not-found=true
               kubectl delete cm/{{ $name }}-config -n {{ $namespace }}
```
jwcesign, line 39: how about line 39? should we add this also?

Poor12:
A little strange that CI did not run. How about re-push the PR again?

charts/karmada/templates/post-delete-job.yaml
calvin0327: Although the configmap will exist at 99% of the time, but we can add the param alse.

PR #2956, opened 2022-12-13
Karmadactl join: default use current context

helen-frank:

**What type of PR is this?**
/kind feature

The current context should be used instead of the clustername when the `--cluster-context` parameter is not set.
Fixes #

When I don't want to provide the`--cluster-context` parameter and encounter obstacles customizing the clustername, clustername become the `--cluster-context`, but I don't have such a context.

```release-note
`karmadactl`: The `--cluster-context` flag of `init` command now takes `current-context` by default.
```

### DIFF
--- a/pkg/karmadactl/join/join.go
+++ b/pkg/karmadactl/join/join.go
@@ -98,11 +98,6 @@ func (j *CommandJoinOption) Complete(args []string) error {
 		j.ClusterName = args[0]
 	}
 
-	// If '--cluster-context' not specified, take the cluster name as the context.
-	if len(j.ClusterContext) == 0 {
-		j.ClusterContext = j.ClusterName
-	}
-
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


helen-frank:
A little idea, maybe we can use the name of the current context to join the karmada when the clustername is not filled in.

pkg/karmadactl/join/join.go
```diff
@@ -98,9 +99,13 @@ func (j *CommandJoinOption) Complete(args []string) error {
 		j.ClusterName = args[0]
 	}
 
-	// If '--cluster-context' not specified, take the cluster name as the context.
+	// If '--cluster-context' not specified, use current context.
 	if len(j.ClusterContext) == 0 {
-		j.ClusterContext = j.ClusterName
+		apiconfig, err := clientcmd.LoadFromFile(j.ClusterKubeConfig)
```
lonelyCZ, line 104: It seems that the default context is exactly current context when `j.ClusterContext = ""`.

helen-frank: I just tested and looked at the following code. When the context is empty, use the current context.

pkg/karmadactl/join/join.go
```diff
@@ -127,9 +122,8 @@ func (j *CommandJoinOption) Validate(args []string) error {
 // AddFlags adds flags to the specified FlagSet.
 func (j *CommandJoinOption) AddFlags(flags *pflag.FlagSet) {
 	flags.StringVar(&j.ClusterNamespace, "cluster-namespace", options.DefaultKarmadaClusterNamespace, "Namespace in the control plane where member cluster secrets are stored.")
-
 	flags.StringVar(&j.ClusterContext, "cluster-context", "",
-		"Context name of cluster in kubeconfig. Only works when there are multiple contexts in the kubeconfig.")
+		"Context name of cluster in kubeconfig. Please make sure this context exists in the cluster, default use current context.")
```
lonelyCZ, line 126:
```suggestion
		"Name of cluster context in kubeconfig. The current context is used by default.")
```

helen-frank: OK, but my e2e cannot pass. I'll check it later.

lonelyCZ: It seems we also need to update the e2e, because it using the old process.

lonelyCZ:
/lgtm

It could resolve some confusion of new users that don't unfamiliar the relationship between `cluster name` and `cluster context`.

/cc @RainbowMango

RainbowMango (approved):
/approve

I updated the release-note.

lonelyCZ:
> I updated the release-note.

Hi @RainbowMango , it should be `join` command.

> `karmadactl`: The `--cluster-context` flag of `init` command now takes `current-context` by default.

to
> `karmadactl`: The `--cluster-context` flag of `join` command now takes `current-context` by default.

RainbowMango:
oh, right.

PR #2958, opened 2022-12-14
Introduce generic hashset to simplify code

RainbowMango:
**What type of PR is this?**

/kind cleanup

This PR introduced a generic `set` that can be widely used by `comparable` types.

Fixes #

This is our first time adopting Golang generics, I didn't add many abilities to the `hashset`, let's see if it can simplify the code.

```release-note
NONE
```


XiShanYongYe-Chang (commented):
/lgtm
Thanks~

Garrybest:
nit: I think `k/k` already has a hashset code [here](https://github.com/kubernetes/kubernetes/blob/release-1.26/staging/src/k8s.io/apimachinery/pkg/util/sets/set.go) if we update to v1.26.

RainbowMango:
Good to know. That's great!

Let's do it when v1.26.1 comes out. And I think we have a lot of adaptation work to do since we are heavily using the sets everywhere.

Issue #2099, opened 2022-07-01
rb、work name collision


likakuli:
**What happened**:

Name collision happens both for rb and work.

For rb, its name is generated by original resource object name and kind, so if there are two resources with same name and kind but different group version info in same namespace, the rb's name will be same too.

For work, its name is generated by original resource object namespace, name and rb's hash. Namespace, Name and hash can be same at the same time for different original resource object, so the work's name can be same too.

All the places to create or update rb or work is called by CreateOrUpdate function which will result to override when name collision happens. It's dangerous.  

**What you expected to happen**:
rb、work name unique for each resource
**How to reproduce it (as minimally and precisely as possible)**:

1. create two resources in same ns with same name, one for advanced statefulset in openkruise and another is statefulset in k8s. 
2. create two pp resource with same cluster affinity for the two resource created above.

**Anything else we need to know?**:

**Environment**:
- Karmada version: master
- kubectl-karmada or karmadactl version (the result of `kubectl-karmada version` or `karmadactl version`):
- Others:

RainbowMango:
Do you have any idea to fix this?

likakuli:
RB and Work are both internal resource which user should not access them directly. But there may some user user use them directly and user-facing change is inevitable for those users. 

I think we should change RB and Work name specification and offer a guarantee that user will not need to redeploy their original resource like deployment and pp and user's resource already dispatched to member clusters can work normally when upgrading RB and Work name. 

I think we can provide new sub-command in kubectl-karmada whose function is to add some labels in old rb and work's metadata and use it when reconcile rb and work to determine whether the old rb, work and work's manifest should to be deleted.

RainbowMango:
> I think we should change RB and Work name specification and offer a guarantee that user will not need to redeploy their original resource like deployment and pp and user's resource already dispatched to member clusters can work normally when upgrading RB and Work name. 

+1

> I think we can provide new sub-command in kubectl-karmada whose function is to add some labels in old rb and work's metadata and karmada-controller-manager can use it when reconcile rb and work to determine whether the old rb, work and work's manifest should to be deleted.

likakuli:
I can work on this and #2090  if there is nobody working on them /cc @RainbowMango

RainbowMango:
Thanks, But, seems we haven't made the final solution yet.

RainbowMango:
/assign @dddddai 
Since @dddddai is leading the effort for this now. We hope to solve this issue in release-1.5.

[...]

PR #2966, opened 2022-12-16
fix RB may not be created when delete and create the same object immediately

### LINKED ISSUE #2090: RB may not be created when delete and create the same object immediately
**What happened**:
1. Create a deployment and wait it until running.
```bash
# kubectl create -f samples/nginx/
```

```bash
nginx   2/2     2            2           3m4s
```

2. Delete this deployment and create it again immediately, we could use `kubectl replace`

```bash
# kubectl replace --force -f samples/nginx/deployment.yaml                         
deployment.apps "nginx" deleted
deployment.apps/nginx replaced
```

3. RB will not be created

```bash
# kubectl get rb                    
NAME            SCHEDULED   FULLYAPPLIED   AGE
```

Recreate RB.

Procedure:
1. We delete deployment.
2. RB tends to be `GC` but not yet deleted.
3. We create deployment, controller could just update the former RB.
4. RB is deleted by `GC`.
5. The controller could not create RB because no more events received.

I think we could watch the deletion event for RB and enqueue the object into detector.

/cc @XiShanYongYe-Chang @pigletfly @RainbowMango 

**Environment**:
- Karmada version:
- Others:

Garrybest:

**What type of PR is this?**
/kind bug

Fixes #2090

```release-note
NONE
```

### DIFF
--- a/pkg/detector/detector.go
+++ b/pkg/detector/detector.go
@@ -400,6 +400,12 @@ func (d *ResourceDetector) ApplyPolicy(object *unstructured.Unstructured, object
 	bindingCopy := binding.DeepCopy()
 	err = retry.RetryOnConflict(retry.DefaultRetry, func() (err error) {
 		operationResult, err = controllerutil.CreateOrUpdate(context.TODO(), d.Client, bindingCopy, func() error {
+			// If this binding exists and its owner is not the input object, return error and let garbage collector
+			// delete this binding and try again later. See https://github.com/karmada-io/karmada/issues/2090.
+			if ownerRef := metav1.GetControllerOfNoCopy(bindingCopy); ownerRef != nil && ownerRef.UID != object.GetUID() {
+				return fmt.Errorf("failed to update binding due to different owner reference UID, will " +
+					"try again later after binding is garbage collected, see https://github.com/karmada-io/karmada/issues/2090")
+			}
 			// Just update necessary fields, especially avoid modifying Spec.Clusters which is scheduling result, if already exists.
 			bindingCopy.Labels = util.DedupeAndMergeLabels(bindingCopy.Labels, binding.Labels)
 			bindingCopy.OwnerReferences = binding.OwnerReferences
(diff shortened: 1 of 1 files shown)


dddddai:
Oh sorry for the delay actually I'm working on another solution, let me file a PR now
I think this fix makes sense, we can keep it util we _enforce_ the new naming rule

Garrybest:
No problem, this PR could be a temporary fix until your final solution.

RainbowMango:
Oh, I just saw you sent another PR https://github.com/karmada-io/karmada/pull/2969.

dddddai:
> Do you mean to move this forward?

Yes

> Your solution actually solves https://github.com/karmada-io/karmada/issues/2099.

It solves #2090 as well, because we will create a new binding for the new object instead of updating the old one

pkg/detector/detector.go
```diff
@@ -466,6 +473,12 @@ func (d *ResourceDetector) ApplyClusterPolicy(object *unstructured.Unstructured,
 		bindingCopy := binding.DeepCopy()
 		err = retry.RetryOnConflict(retry.DefaultRetry, func() (err error) {
 			operationResult, err = controllerutil.CreateOrUpdate(context.TODO(), d.Client, bindingCopy, func() error {
+				// If this binding exists and its owner is not the input object, return error and let garbage collector
+				// delete this binding and try again later. See https://github.com/karmada-io/karmada/issues/2090.
+				if ownerRef := metav1.GetControllerOfNoCopy(bindingCopy); ownerRef != nil && ownerRef.UID != object.GetUID() {
```
dddddai, line 478: I think we should add this for L516 as well?

nit: maybe `bindingCopy.Spec.Resource.UID != object.GetUID()` is shorter?

RainbowMango: > nit: maybe bindingCopy.Spec.Resource.UID != object.GetUID() is shorter?

It is. But I think we should use `ownerRef` because it is designed for representing the dependency.

RainbowMango: > I think we should add this for L516 as well?

Please @Garrybest check.

Garrybest: > I think we should add this for L516 as well?
Indeed, thanks for reminding.

> nit: maybe bindingCopy.Spec.Resource.UID != object.GetUID() is shorter?
Nice finding👍. Both are OK but maybe using ownerRef is more distinct literally.

RainbowMango (approved):
/lgtm
/approve

PR #2968, opened 2022-12-16
fix the condition that revise replicas

### LINKED ISSUE #2967: Workload may be propagated unexpectedly
**What happened**:
Workload may be propagated unexpectedly in some prescribed circumstances.

```bash
kubectl apply -f - << EOF
kind: PropagationPolicy
metadata:
  name: nginx
spec:
  resourceSelectors:
    - apiVersion: apps/v1
      kind: Deployment
  placement:
    clusterAffinity:
      clusterNames:
        - member1
        - member2
        - member3
    replicaScheduling:
      weightPreference:
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-dp
spec:
  replicas: 10
  selector:
    matchLabels:
      app: nginx-dp
  template:
    metadata:
      labels:
        app: nginx-dp
    spec:
      containers:
        - name: nginx-dp
          image: nginx
          resources:
            limits:
              cpu: "1"
EOF
```

This workload will be divided into member clusters. In my environment, it's 
```
Spec:
  Clusters:
    Name:      member1
    Replicas:  4
    Name:      member2
    Replicas:  3
    Name:      member3
    Replicas:  3
```

We scale down this deployment to 0.
```
kubectl scale --replicas 0 deployment nginx-dp
```

And we scale up this deployment to 100.
```
kubectl scale --replicas 100 deployment nginx-dp
```

We don't have so many resoures in member1, member2 and member3. However, this deployment is reproduced into 3 member clusters with 100 replicas each, which is not expected.
```
Spec:
  Clusters:
    Name:  member1
    Name:  member2
    Name:  member3
  Replica Requirements:
    Resource Request:
      Cpu:   1
  Replicas:  100
  Resource:
    API Version:       apps/v1
    Kind:              Deployment
    Name:              nginx-dp
    Namespace:         default
    Resource Version:  11557
    UID:               0c90b1ab-1752-4d53-87dc-d15352e5ee0e
Status:
  Aggregated Status:
    Applied:       true
    Cluster Name:  member1
    Health:        Unhealthy
    Status:
      Unavailable Replicas:  100
    Applied:                 true
    Cluster Name:            member2
    Health:                  Unhealthy
    Status:
      Unavailable Replicas:  100
    Applied:                 true
    Cluster Name:            member3
    Health:                  Unhealthy
    Status:
      Unavailable Replicas:  100
  Conditions:
    Last Transition Time:         2022-12-16T08:57:59Z
    Message:                      All works have been successfully applied
    Reason:                       FullyAppliedSuccess
    Status:                       True
    Type:                         FullyApplied
    Last Transition Time:         2022-12-16T08:59:43Z
    Message:                      failed to assignReplicas: failed to scale up: clusters resources are not enough to schedule, max 0 replicas are support
    Reason:                       BindingFailedScheduling
    Status:                       False
    Type:                         Scheduled
```

**Environment**:
- Karmada version:
- Others:

Garrybest:

**What type of PR is this?**
/kind bug

Fixes #2967

```release-note
NONE
```


pkg/controllers/binding/common.go
```diff
@@ -165,13 +176,6 @@ func mergeTargetClusters(targetClusters []workv1alpha2.TargetCluster, requiredBy
 	return targetClusters
 }
 
-func getReplicaInfos(targetClusters []workv1alpha2.TargetCluster) (bool, map[string]int64) {
-	if helper.HasScheduledReplica(targetClusters) {
```
XiShanYongYe-Chang, line 169: The func `HasScheduledReplica` can be removed at the same time,

pkg/controllers/binding/common.go
```diff
@@ -252,3 +256,14 @@ func divideReplicasByJobCompletions(workload *unstructured.Unstructured, cluster
 
 	return targetClusters, nil
 }
+
+func hasReplicaSchedulingPolicy(appliedPlacement map[string]string) (bool, error) {
+	placement, err := helper.GetAppliedPlacement(appliedPlacement)
+	if err != nil {
+		return false, err
+	}
+	if placement == nil {
```
XiShanYongYe-Chang, line 265: When `placement` is nil,  it indicates that the scheduler has not scheduled the scheduler. In this case, the scheduling result is empty, so I think we don't need to return an error.

pkg/controllers/binding/common.go
```diff
@@ -81,7 +83,16 @@ func ensureWork(
...
+
+	// Check if we should revise replicas, only bindings who have ReplicaScheduling policy could be revised.
+	var desireReplicaInfos map[string]int64
+	shouldReviseReplica, err := hasReplicaSchedulingPolicy(binding.GetAnnotations())
+	if err != nil {
+		return err
+	}
+	if shouldReviseReplica {
+		desireReplicaInfos = transScheduleResultToMap(targetClusters)
+	}
```
XiShanYongYe-Chang, line 95: Moving this logic into line107 will be ok.

pkg/controllers/binding/common.go
Garrybest: OK, it may be better to add a judgement first. If `len(targetClusters) == 0`, function `ensureWork` would return. So if placement is nil here, which also indicates `len(targetClusters) != 0`, there must be something wrong here.

pkg/controllers/binding/common.go
```diff
@@ -54,34 +55,54 @@ var workPredicateFn = builder.WithPredicates(predicate.Funcs{
...
+	// If binding has replicas, those who have ReplicaScheduling policy should be revised.
+	if needReviseReplica {
+		needReviseReplica, err = hasReplicaSchedulingPolicy(binding.GetAnnotations())
+		if err != nil {
+			return err
+		}
+		if needReviseReplica {
+			desireReplicaInfos = transScheduleResultToMap(targetClusters)
+		}
+	}
```
XiShanYongYe-Chang, line 97: I think we can simply remove this logic and determine whether to continue the execution based on whether the rb scheduling is complete:

https://github.com/karmada-io/karmada/blob/c4428d6f224b432a47bfa6aa2bdd6b3a1ccfa275/pkg/apis/work/v1alpha2/binding_types.go#L222-L226

And `ResourceInterpreter.HookEnabled` will determine whether to revise.

Garrybest: > I think we can simply remove this logic and determine whether to continue the execution based on whether the rb scheduling is complete.

By determining `SchedulerObservedGeneration` > 0 ? I'm afraid if you delete `applied-placement` manually(which I have tried this trick before🤣 to fully reschedule a deployment), there could be some problems here.

XiShanYongYe-Chang: > By determining SchedulerObservedGeneration > 0 ?

By determining `rb.Status.SchedulerObservedGeneration == rb.Generation`, this indicates that the current rb is the latest result processed by the scheduler.

If this method is used for judgment, we need to change the previous usage habit. For the pp whose `ReplicaScheduling` is nil, and the matched resource is workload, the replicas field in `Spec.Clusters` needs to be filled scheduling is complete.

Garrybest: 1. If `rb.Status.SchedulerObservedGeneration != rb.Generation`, just return nil.
2. And then if `ReplicaScheduling` is nil, we don't revise replicas.

Right?

XiShanYongYe-Chang: Just the first point is enough, and it's up to `enable` function to decide whether to revise.

Garrybest: I don't think it gonna works. If `ReplicaScheduling` is nil, scheduler consider this workload as a kind of 'common' resources, like configMap or secret, so the result could be `[member1, member2, member3]` without replica assigned. But here if we don't care about `ReplicaScheduling`and let `enable` function to decide, replicas with all member clusters will be zero by revision which cause an expected result.

Garrybest: Oh, I think we have conclued a perfect solution in private😄.

Garrybest:
still have some problems here /hold

XiShanYongYe-Chang:
Hi @Garrybest please cc me when it's ready to review.

Garrybest:
cc @XiShanYongYe-Chang

pkg/resourceinterpreter/defaultinterpreter/retain.go
```diff
@@ -111,5 +112,17 @@ func retainJobSelectorFields(desired, observed *unstructured.Unstructured) (*uns
...
+	// retain job completions: spec.completions is immutable if job was created before.
+	completions, ok, err := unstructured.NestedInt64(observed.Object, "spec", "completions")
+	if err != nil {
+		return nil, fmt.Errorf("failed to retrieve completions from job: %v", err)
+	}
+	if ok {
+		if err = unstructured.SetNestedField(desired.Object, completions, "spec", "completions"); err != nil {
+			return nil, fmt.Errorf("failed to set completions for job: %v", err)
+		}
+	}
```
XiShanYongYe-Chang, line 125: Does this change involve problems that existed before?

Garrybest: I think so, we had this problem before because job completion should never be updated.

PR #2969, opened 2022-12-16
Revise binding/work naming

### LINKED ISSUE #2090: RB may not be created when delete and create the same object immediately

dddddai:
**What type of PR is this?**
/kind bug
/kind design

The current binding/work naming has some defects as mentioned in #2090 and #2099, this pr revises the naming rule to solve them

If name collision happens, we will append a random suffix to the name and retry again

Fixes https://github.com/karmada-io/karmada/issues/2090 and https://github.com/karmada-io/karmada/issues/2099

```release-note
Fixed binding/work name collision
```


pkg/features/features.go
```diff
@@ -18,6 +18,10 @@ const (
 
 	// CustomizedClusterResourceModeling indicates if enable cluster resource custom modeling.
 	CustomizedClusterResourceModeling featuregate.Feature = "CustomizedClusterResourceModeling"
+
+	// AdvancedResourceBinding indicates if names of bindings and works should be generated with parents' uid hash.
+	// TODO: remove this when legacy names are no longer supported
+	AdvancedResourceBinding featuregate.Feature = "AdvancedResourceBinding"
```
dddddai, line 24: We need a reasonable name for it

pkg/controllers/federatedresourcequota/federated_resource_quota_sync_controller.go
```diff
@@ -150,7 +150,7 @@ func (c *SyncController) buildWorks(quota *policyv1alpha1.FederatedResourceQuota
 			errs = append(errs, err)
 			continue
 		}
-		workName := names.GenerateWorkName("ResourceQuota", quota.Name, quota.Namespace)
+		workName := names.GenerateWorkNameLegacy("ResourceQuota", quota.Name, quota.Namespace)
```
dddddai, line 153: For works created by dedicated controllers, they are still using legacy names, I think it's fine since name collisions would not happen

We can use new naming for them, that would add some boilerplate code in those controllers though

dddddai:
Hold, let me think if there's an easier way..

RainbowMango:
Thanks @dddddai, I'll get back on this later, maybe after release 1.5.

dddddai:
OK, no rush :)

pkg/util/names/names.go
```diff
@@ -49,8 +61,65 @@ func GetClusterName(executionSpaceName string) (string, error) {
 	return strings.TrimPrefix(executionSpaceName, ExecutionSpacePrefix), nil
 }
 
-// GenerateBindingName will generate binding name by kind and name
-func GenerateBindingName(kind, name string) string {
+// GenerateBindingName will generate binding name for the template.
+func GenerateBindingName(ctx context.Context, karmadaClient client.Client, dynamicClient dynamic.Interface,
```
RainbowMango, line 65: Here(and the `GenerateWorkName` below) tells the main idea that identifying the conflicts and making a random name, right?

dddddai: Yes, we will append a random suffix if there's a conlict, and record the name in resource template annotations

RainbowMango: One of the advantages is that the mechanism can be shared with three controllers:
1. build ResourceBinding by [detector](https://github.com/karmada-io/karmada/blob/83c7862289ac3884985cee2a10f023ffb8f925b2/pkg/detector/detector.go#L392)
2. build ResourceBinding by [dependenciesdistributor](https://github.com/karmada-io/karmada/blob/83c7862289ac3884985cee2a10f023ffb8f925b2/pkg/dependenciesdistributor/dependencies_distributor.go#L551)
3. build ResourceBinding by [hpa controller](https://github.com/karmada-io/karmada/blob/83c7862289ac3884985cee2a10f023ffb8f925b2/pkg/controllers/hpa/hpa_controller.go#L138)

But the disadvantage seems to be that `GenerateBindingName` is coupled with too much logic that beyond a function should do.  And after we generated the ResourceBinding, we usually try to create or update it by `controllerutil.CreateOrUpdate` which will `get` the ResourceBinding one more time in addition to `GenerateBindingName`, that seems redundant. 

In general, I like the idea that we can introduce an `annotation` to the resource template to record the conflict condition(just like  [Deployment CollisionCount](https://github.com/kubernetes/kubernetes/blob/c6f3007071d3212417ddab088adc5cc2e388b167/staging/src/k8s.io/api/apps/v1/types.go#L515)) that could be a factor to build ResourceBinding name.

So, I'm thinking if there is a more elegant way to do this. I don't have a clue yet.
I'd like to invite @Garrybest @likakuli for comment.

Garrybest: 100% agree. This function is much more complex than it should be. Introduce an anotation or an API field could be better.

likakuli: agree. I think a new field like CollisionCount could be better.

RainbowMango: @likakuli We can't change the API of any resource template, that's the reason why we are considering annotation.

RainbowMango:
I guess we don't have enough time for this feature, so I'm moving this to v1.8.

PR #2991, opened 2022-12-26
keep collecting status when apply failed

Garrybest:

**What type of PR is this?**
/kind bug

If there is something wrong with applied condition, we should not erase the status in binding status. Applied information is from execution controller whereas status is from work status controller.

Fixes #

```release-note
NONE
```


jwcesign:
Can we add ut for this? Also, we need release notes.

RainbowMango:
> If there is something wrong with applied condition, we should not erase the status in binding status.

Can you explain it in more detail? Where do we `erase` the status? and why we shouldn't do that?

Garrybest:
> Where do we erase the status?

If failed to apply, aggregated status will not be set even though status exists in work status and this function just returns.
https://github.com/karmada-io/karmada/blob/3d06bb84a339c099803f5165822a5703b42fe97b/pkg/util/helper/workstatus.go#L186-L196

> Why we shouldn't do that?

Actually, apply is sort of synchronization from 'karmada' to 'member clusters', it only take effects on resource metadata and spec. However, status collection is a synchronization from 'member clusters' to 'karmada'. So it is unresonable to skip status aggregation when failed to apply because these two actions have no identical relationship.

XiShanYongYe-Chang (commented):
Hi @Garrybest, I think the modification is reasonable. 

Do you encounter this problem? After a resource fails to be applied, the status of the work exists.

In addition, we have a similar logic. Please confirm whether to modify it:
https://github.com/karmada-io/karmada/blob/a20dd6e6303b279107a990ec0720cf9ca606d786/pkg/controllers/federatedresourcequota/federated_resource_quota_status_controller.go#L183

XiShanYongYe-Chang:
This one looks more like a cleanup, what do you think @Garrybest?

Garrybest:
> Do you encounter this problem? After a resource fails to be applied, the status of the work exists.

Yes, I got a validation webhook in member cluster which rejects the modification whereas it had no effects on status collection.

> In addition, we have a similar logic.

Cool.

> This one looks more like a cleanup, what do you think @Garrybest?

I'm not sure about the category, cc @RainbowMango for help😄.

RainbowMango:
>  So it is unresonable to skip status aggregation when failed to apply because these two actions have no identical relationship.

The reason why `skip status aggregation` is the status is probably an ambiguous one. It's not the right status of `current` resource template.
For instance:
1. We deployed a Deployment(v1), and get the status back, it works as expected.
2. Then, we update the Deployment to V2, if we can't apply it to member clusters, then we still get the V1 status. It will bring ambiguity.

Garrybest:
> For instance:
> 
> 1. We deployed a Deployment(v1), and get the status back, it works as expected.
> 2. Then, we update the Deployment to V2, if we can't apply it to member clusters, then we still get the V1 status. It will bring ambiguity.

Got it, it makes sense. Well, in this case, deployment of status will be erased here. So users will see a deployment with 0 ready replicas and 0 available replicas from status, maybe this situation brings a panic for users.

These two different logic represents different consideration and scenarios, which one is more sensible? Glad to here more opinions😊.

RainbowMango:
> So users will see a deployment with 0 ready replicas and 0 available replicas from status, maybe this situation brings a panic for users.

Yes, I agree that just erasing the previous status is not ideal behavior, but I think we have sent the event to the resource template, isn't it? it should be a relief.

Garrybest:
Actually I'm just concerned that users are likely to have their own controllers in karmada control plane which depend on object status to reconcile and the handling process could be disrupted since status does not collect correctly under this circumstance.

Well, it's just some ideas, sort of design tradeoff😄. We may need more evidences and scenario about this patch.

RainbowMango:
> Well, it's just some ideas, sort of design tradeoff😄. We may need more evidences and scenario about this patch.

Yes. kind of tradeoff here.

RainbowMango:
I guess we don't have enough time for this feature, so I'm moving this to v1.8.

@XiShanYongYe-Chang Could you please confirm if we still facing this issue? I feel we solved a similar issue somewhere. Not sure about that.

PR #2997, opened 2022-12-27
support PodDisruptionBudget resource in default interpreter

a7i:
Signed-off-by: Amir Alavi <amiralavi7@gmail.com>

**What type of PR is this?**

/kind feature
Fixes #

```release-note
support PodDisruptionBudget resource in default interpreter
```


jwcesign:
Thanks for your PR @a7i , I have one question, for pdb, there is `spec.minAvailable` like follows:
```yaml
apiVersion: policy/v1
kind: PodDisruptionBudget
metadata:
  name: zk-pdb
spec:
  minAvailable: 10
  selector:
    matchLabels:
      app: zookeeper
```
So this pdp will be created in all member clusters with `minAvailable: 10`? or will be split into different numbers for different member clusters?

RainbowMango:
Hi @a7i, thanks for doing this.
Can you help to share your use case here?

a7i:
> Thanks for your PR @a7i , I have one question, for pdb, there is `spec.minAvailable` like follows:
> 
> ```yaml
> apiVersion: policy/v1
> kind: PodDisruptionBudget
> metadata:
>   name: zk-pdb
> spec:
>   minAvailable: 10
>   selector:
>     matchLabels:
>       app: zookeeper
> ```
> 
> So this pdp will be created in all member clusters with `minAvailable: 10`? or will be split into different numbers for different member clusters?

@jwcesign  As far as I know, that is how it is today. With or without this change. We are doing more testing on PDBs this week and I can circle back and let you know how it works with different propagation policies.

a7i:
> Hi @a7i, thanks for doing this. Can you help to share your use case here?

@RainbowMango We are conducting experiments on how Karmada treats PDBs. First blocker was that PDB status is not aggregated. A `kubectl get pdb ..` will yield missing or N/A results today.

chaunceyjiang:
> So this pdp will be created in all member clusters with `minAvailable: 10`? or will be split into different numbers for different member clusters?

At present, kamrada can only create this pdp in all member clusters, right? @jwcesign

jwcesign:
> At present, kamrada can only create this pdp(minAvailable: 10) in all member clusters, right? @jwcesign

I think yes, if want to make pdp work perfectly, `minAvailable` needs to be scheduled based on pp, the logic is like what we talk about FederationHPA. @chaunceyjiang , what u think?

a7i:
> > At present, kamrada can only create this pdp(minAvailable: 10) in all member clusters, right? @jwcesign
> 
> I think yes, if want to make pdp work perfectly, `minAvailable` needs to be scheduled based on pp, the logic is like what we talk about FederationHPA. @chaunceyjiang , what u think?

Confirming that this is the case using [v1.3.3](https://github.com/karmada-io/karmada/tree/v1.3.3)
Currently, the percentage or number (min or max) basically creates the same resource in all member clusters, regardless of the weighted/divided configuration of pp.

jwcesign:
Hi @a7i , in your use case, which behavior is expected? keep the same with the original pdb, or it should be scheduled based on the pp configuration?

chaunceyjiang:
> I think yes, if want to make pdp work perfectly, minAvailable needs to be scheduled based on pp, the logic is like what we talk about FederationHPA. @chaunceyjiang , what u think?

Good idea, agree with you.

a7i:
@jwcesign My organization almost exclusively uses `maxUnavailable` with a percentage (e.g. `20%`). Minus a few exceptions like Kafka where replica count is always fixed.

I think that for percentages, keeping the same value is valid but there are edge cases to that even:

- member-1: 2 replica Deployment
- member-2: 2 replica Deployment
- PDB `maxUnavailable: 50%`

Deleting a single Pod from member-1 will signal no more disruption allowed (even though at a global level, one more is still tolerable). We could report `DisruptionAllowed=True` but need to take into account what that means for operations that still happen at the member cluster level (e.g. draining a node and how it could get blocked).

I agree that handling PDB requires its own proposal (separate from HPA) but still feel that this PR can be reviewed since it's only aggregating status.

jwcesign:
Yeah, it should work fine with percentages.

I am ok with this PR(only aggregating status), cc @XiShanYongYe-Chang for checking

pkg/resourceinterpreter/defaultinterpreter/aggregatestatus.go
```diff
@@ -455,3 +458,49 @@ func aggregatePersistentVolumeClaimStatus(object *unstructured.Unstructured, agg
...
+		}
+
+		temp := &policyv1.PodDisruptionBudgetStatus{}
+		if err = json.Unmarshal(item.Status.Raw, temp); err != nil {
+			return nil, err
+		}
+		klog.V(3).Infof(
+			"Grab pdb(%s/%s) status from cluster(%s), desired healthy: %d, current healthy: %d, disrupted: %d, expected: %s",
+			pdb.Namespace, pdb.Name, item.ClusterName,
+			temp.DesiredHealthy, temp.CurrentHealthy, temp.DisruptedPods, temp.ExpectedPods,
```
XiShanYongYe-Chang, line 484: `temp.DisruptedPods` is a slice, It's inappropriate to use `%d` to indicate it;
and why not use `%d` to format the output `temp.ExpectedPods`?

pkg/resourceinterpreter/defaultinterpreter/aggregatestatus.go
```diff
@@ -455,3 +458,49 @@ func aggregatePersistentVolumeClaimStatus(object *unstructured.Unstructured, agg
...
+		)
+
+		newStatus.CurrentHealthy += temp.CurrentHealthy
+		newStatus.DesiredHealthy += temp.DesiredHealthy
+		newStatus.ExpectedPods += temp.ExpectedPods
+		newStatus.DisruptionsAllowed += temp.DisruptionsAllowed
+		newStatus.Conditions = append(newStatus.Conditions, temp.Conditions...)
```
XiShanYongYe-Chang, line 491: What if the same `reason` condition occurs in the different clusters?

pkg/resourceinterpreter/defaultinterpreter/aggregatestatus.go
```diff
@@ -455,3 +458,49 @@ func aggregatePersistentVolumeClaimStatus(object *unstructured.Unstructured, agg
...
+		if temp.DisruptedPods != nil {
+			for k, v := range temp.DisruptedPods {
```
XiShanYongYe-Chang, line 493: Is it possible that different clusters have the same pod name?

XiShanYongYe-Chang (commented):
Hi @a7i, thanks for your contribution!
Can you help add an E2E test in the `https://github.com/karmada-io/karmada/blob/master/test/e2e/resource_test.go` file for the `PodDisruptionBudget` resource?

pkg/resourceinterpreter/defaultinterpreter/aggregatestatus.go
a7i: I would want them all still aggregated and combined. Ideally, it would contain the member cluster name but there are no fields to squeeze that in.

Here is an example of a condition struct:
```
  conditions:
  - lastTransitionTime: "2023-01-04T17:17:54Z"
    message: ""
    observedGeneration: 1
    reason: SufficientPods
    status: "True"
    type: DisruptionAllowed
```

pkg/resourceinterpreter/defaultinterpreter/aggregatestatus.go
a7i: UID is guaranteed to be unqiue across a whole cluster. To have a collision on a replicaset UID and a pod UID is highly highly unlikely.

pod name: `{deployment-name}-{replicaset-uid}-{pod-uid}`

I'll go ahead and include the clusterName in there as well

pkg/resourceinterpreter/defaultinterpreter/aggregatestatus.go
a7i: We don't plan on using the conditions for any monitors or metrics. It can, however, be helpful for live debugging.

I'm not certain what you're asking to change here, can you elaborate?

a7i: We don't plan on using the conditions for any monitors or metrics. It can, however, be helpful for live debugging.

I'm not certain what you're asking to change here, can you elaborate?

XiShanYongYe-Chang: I think it's not useful to collect the condition, due to the existence of redundant information.

pkg/resourceinterpreter/defaultinterpreter/aggregatestatus.go
XiShanYongYe-Chang: ok, thanks~

a7i:
Just requested another review from @XiShanYongYe-Chang , didn't intend to remove.

cc: @RainbowMango

test/e2e/framework/poddisruptionbudget.go
```diff
@@ -0,0 +1,28 @@
...
+	"fmt"
+
+	"github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
+	policyv1 "k8s.io/api/policy/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+// CreatePodDisruptionBudget create PodDisruptionBudget.
```
XiShanYongYe-Chang, line 14: create -> creates

test/e2e/framework/poddisruptionbudget.go
```diff
@@ -0,0 +1,28 @@
...
+func CreatePodDisruptionBudget(client kubernetes.Interface, pdb *policyv1.PodDisruptionBudget) {
+	ginkgo.By(fmt.Sprintf("Creating PodDisruptionBudget(%s/%s)", pdb.Namespace, pdb.Name), func() {
+		_, err := client.PolicyV1().PodDisruptionBudgets(pdb.Namespace).Create(context.TODO(), pdb, metav1.CreateOptions{})
+		gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+	})
+}
+
+// RemovePodDisruptionBudget delete PodDisruptionBudget.
```
XiShanYongYe-Chang, line 22: delete -> deletes

XiShanYongYe-Chang (commented):
Just some nits, other parts look good to me, thanks a lot~

pkg/resourceinterpreter/defaultinterpreter/aggregatestatus.go
```diff
@@ -455,3 +458,48 @@ func aggregatePersistentVolumeClaimStatus(object *unstructured.Unstructured, agg
...
+			"Grab pdb(%s/%s) status from cluster(%s), desired healthy: %d, current healthy: %d, disrupted allowed: %d, expected: %d",
+			pdb.Namespace, pdb.Name, item.ClusterName,
+			temp.DesiredHealthy, temp.CurrentHealthy, temp.DisruptionsAllowed, temp.ExpectedPods,
+		)
+
+		newStatus.CurrentHealthy += temp.CurrentHealthy
+		newStatus.DesiredHealthy += temp.DesiredHealthy
+		newStatus.ExpectedPods += temp.ExpectedPods
+		newStatus.DisruptionsAllowed += temp.DisruptionsAllowed
+		if temp.DisruptedPods != nil {
```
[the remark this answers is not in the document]
a7i: How so? maps can be nil

test/e2e/framework/poddisruptionbudget.go
a7i: FWIW, this is following convention:

[ConfigMap](https://github.com/karmada-io/karmada/blob/9b8a2b5a184d459af5e189ca12176e994da9297d/test/e2e/framework/configmap.go#L17)
[Ingress](https://github.com/karmada-io/karmada/blob/9b8a2b5a184d459af5e189ca12176e994da9297d/test/e2e/framework/ingress.go#L14)
[Daemonset](https://github.com/karmada-io/karmada/blob/9b8a2b5a184d459af5e189ca12176e994da9297d/test/e2e/framework/daemonSet.go#L14)
[Job](https://github.com/karmada-io/karmada/blob/9b8a2b5a184d459af5e189ca12176e994da9297d/test/e2e/framework/job.go#L17)

test/e2e/framework/poddisruptionbudget.go
a7i: Similar [commit](https://github.com/karmada-io/karmada/commit/695c29fe61ee172574ffd15776052e571acc28c3#diff-c71f375f8dd4601747f87c05674dcb8ef01b1b071063127d24082cd01692ac73R22) following such convention

pkg/util/constants.go
```diff
@@ -98,8 +98,10 @@ const (
 	PersistentVolumeClaimKind = "PersistentVolumeClaim"
 	// PersistentVolumeKind indicates the target resource is a persistentvolume
 	PersistentVolumeKind = "PersistentVolume"
-	// HorizontalPodAutoscalerKind indicated the target resource is a horizontalpodautoscaler
+	// HorizontalPodAutoscalerKind indicates the target resource is a horizontalpodautoscaler
```
RainbowMango, line 101: nice finding.

RainbowMango (approved):
/approve

RainbowMango:
I just opened a task(https://github.com/karmada-io/website/issues/287) to track the document updation.

PR #3000, opened 2022-12-28
Uniform klog output starts with uppercase

Fish-pro:
**What type of PR is this?**

/kind cleanup

Fixes #

In the existing error log, parts begin in lower case，this pr uniform error log output starts with uppercase

```release-note
NONE
```

### DIFF
--- a/cmd/agent/app/agent.go
+++ b/cmd/agent/app/agent.go
@@ -208,7 +208,7 @@ func run(ctx context.Context, opts *options.Options) error {
 	}
 
 	if err := controllerManager.AddHealthzCheck("ping", healthz.Ping); err != nil {
-		klog.Errorf("failed to add health check endpoint: %v", err)
+		klog.Errorf("Failed to add health check endpoint: %v", err)
 		return err
 	}
 
--- a/cmd/controller-manager/app/controllermanager.go
+++ b/cmd/controller-manager/app/controllermanager.go
@@ -147,12 +147,12 @@ func Run(ctx context.Context, opts *options.Options) error {
 		}),
 	})
 	if err != nil {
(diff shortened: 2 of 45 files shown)


Garrybest:
I like this patch, thanks a lot👍.

pkg/util/lifted/pubkeypin/pubkeypin_test.go
```diff
@@ -160,7 +160,7 @@ func TestHash(t *testing.T) {
 	actualHash := Hash(testCert(t, testCertPEM))
 	if actualHash != expectedHash {
 		t.Errorf(
```
Garrybest, line 162: I guess we don't need to change test files because the log file does not using klog.

Fish-pro: @Garrybest Thanks for pointing out that it has been revised

Garrybest (changes requested):
I think we still have other codes need to be modify like https://github.com/karmada-io/karmada/blob/eda7814ffc9f69db7f1e066ad24544fc7dcc19a7/cmd/controller-manager/app/controllermanager.go#L167

Fish-pro:
> I think we still have other codes need to be modify like
> 
> https://github.com/karmada-io/karmada/blob/eda7814ffc9f69db7f1e066ad24544fc7dcc19a7/cmd/controller-manager/app/controllermanager.go#L167

@Garrybest Due to the large number of modified files, I did not modify all of them. Now I have searched for modifications globally

Garrybest:
Good job👍.

/lgtm

RainbowMango (approved):
/approve

PR #3015, opened 2023-01-03
Check for err judgments by `==` and fix it

Fish-pro:

**What type of PR is this?**

/kind cleanup

Comparing with `==` will fail on wrapped errors, use `errors.Is` to check for a specific error，select the unmodified part in the following way
```sh
➜  karmada git:(master) git grep 'err ==' | grep -v ^vendor | sed 's/.*err == \([^ :),;.]\+\).*/\1/g' | grep Err | grep -v "err == nil" | sort -u
pkg/karmadactl/util/genericresource/builder.go:         if err == filepath.ErrBadPattern {

```

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/karmadactl/util/genericresource/builder.go
+++ b/pkg/karmadactl/util/genericresource/builder.go
@@ -1,6 +1,7 @@
 package genericresource
 
 import (
+	"errors"
 	"fmt"
 	"net/url"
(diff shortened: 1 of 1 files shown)


RainbowMango (approved):
/approve

Issue #2979, opened 2022-12-20
[Umbrella] Karmada Operator Plan (First Version) 


carlory:
- [x] Karmada CRD #2953
- [x] Operator  Skeleton #2973 
- [ ] Update scripts, makefile and workflows
- [ ] Operator Webhooks ( It is built into the operator)
       - possible solution 1: #2992 
       - possible solution 2: #3023
- [ ] operator implementation (draft) 
       - possible solution 1: #2990 
       - possible solution 2 (recommend https://github.com/karmada-io/karmada/pull/3102#issuecomment-1414662967): #3102
       We need to discuss, which implementation to use
- [ ] Etcd HA

/assign

carlory:
- What is the relationship between Karmada Operator and Karmadactl?

Refactor Karmadactl, use Karmada Operator to manage and create Karmada instances. Consider init/deinit for now

- Karmada Operator’s early contribution and collaboration methods, regular community meetings, or Karmada Operator’s community meetings for independent issues?

git issue + regular community meeting

- Is the management of the Proposal in the Karmada main library or in the Karmada Operator? The Proposal in the Karmada community has no way to determine the actual progress of its associated functions based on the Proposal document.

Early stage No Proposal.

- What is the scope of responsibility of the Karmada operator? For dependencies such as etcd.opensearch, use the same support granularity to develop an etcd-operator within the karmada-operator?

Built-in, simple and highly available etcd cluster

- Karmada operator certificate management, self-signed or support community solutions such as cert-manager at the same time?

The first version is self-signed, and the follow-up integration

- What version range of Karmada does Karmada Operator support to install? < 1.0 Is it in the supported list?

The first version of Mouding v1.4, each version of Karmada Operator has a supported Karmada version range. Try to support a large version space

- When Karmada Operator itself is installed and deployed, what are the requirements for the host cluster version? 1.16 and above?

The first version supports 1.16, CI added

- When Karmada Operator exposes the service address to the outside world, should it natively support solutions such as MetalLB?

The first version supports external exposure of services, and the first version and the next version support external load balancing

- How does Karmada Operator support the creation of Karmada instances on remote clusters? When the cluster where Karmada Operator is located hangs (unrecoverable), how does the Karmada instance on the remote cluster ensure that its update and daily operation and maintenance are not affected?

To be discussed, first covered in the current cluster

[...]

PR #2990, opened 2022-12-26
[WIP] implement karmada operator

carlory:
Signed-off-by: carlory <baofa.fan@daocloud.io>

**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:
Part of #2979 

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #3023, opened 2023-01-05
add karmada operator crd validations

calvin0327:

**What type of PR is this?**
/kind feature

Before, I want to implement a build-in webhook for karmada operator and I think it's very complex and troublesome for both developers and users. 

The main reason is that the webhook need the tls to connect with apiserver, the means that before running karmada-operator pod, we need generate a certificate and mount to the pod. the pod  and apiserver can be used for authentication each other. It's very unfriendly to users.
In our scenario is every simple, we only need populate defaults to some fields, so, It,s too heavy for us to develop a webhook.

In k8s community, there are two ways replace webhook in simple scenario:
1. replace mutaing webhook with the way add `+kubebuilder:default` annotations to required fields and generate schama.
https://kubernetes.io/zh-cn/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#efaulting

2. replace validation webhook with CEL https://kubernetes.io/blog/2022/09/29/enforce-immutability-using-cel/#basics-of-validation-rules 

Part of # https://github.com/karmada-io/karmada/issues/2979

```release-note
NONE
```

### DIFF
--- a/operator/pkg/apis/operator/v1alpha1/type.go
+++ b/operator/pkg/apis/operator/v1alpha1/type.go
@@ -37,6 +37,7 @@ type Karmada struct {
 
 	// Spec defines the desired behavior of the Karmada.
 	// +optional
+	// +kubebuilder:default:={spec: {}}
 	Spec KarmadaSpec `json:"spec,omitempty"`
 
 	// Most recently observed status of the Karmada.
@@ -49,7 +50,9 @@ type KarmadaSpec struct {
 	// HostCluster represents the cluster where to install the Karmada control plane.
 	// If not set, the control plane will be installed on the cluster where
 	// running the operator.
-	// +optional
+	// +required
+	// +kubebuilder:default:={hostCluster: {}}
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf", message="The value is immutable"
 	HostCluster *HostCluster `json:"hostCluster,omitempty"`
 
 	// PrivateRegistry is the global image registry.
@@ -61,7 +64,8 @@ type KarmadaSpec struct {
 
 	// Components define all of karmada components.
 	// not all of these components need to be installed.
-	// +optional
+	// +required
+	// +kubebuilder:default={components: {}}
 	Components *KarmadaComponents `json:"components,omitempty"`
 
 	// FeatureGates enabled by the user.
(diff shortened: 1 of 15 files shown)


calvin0327:
The tests:

the origin yaml file:
```yaml
kind: Karmada
metadata:
  name: karmada-demo
```

after creating the karmada resource with the yaml file:

```yaml
kind: Karmada
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"operator.karmada.io/v1alpha1","kind":"Karmada","metadata":{"annotations":{},"name":"karmada-demo","namespace":"default"}}
  creationTimestamp: "2023-01-05T09:14:52Z"
  generation: 1
  name: karmada-demo
  namespace: default
  resourceVersion: "6213058"
  uid: e1e578ac-1aa9-4241-9b47-7c55f2d7a12b
spec:
  components:
    etcd:
      local:
        imageRepository: registry.k8s.io
        imageTag: 3.5.3-0
        volumeData:
          emptyDir: {}
    karmadaAPIServer:
      imageRepository: registry.k8s.io
      imageTag: v1.25.2
      replicas: 1
      serviceSubnet: 10.96.0.0/12
    karmadaAggregratedAPIServer:
      imageRepository: docker.io
      imageTag: v1.4.0
      replicas: 1
      serviceType: NodePort
    karmadaControllerManager:
      imageRepository: docker.io
      imageTag: v1.4.0
      replicas: 1
    karmadaScheduler:
      imageRepository: docker.io
      imageTag: v1.4.0
      replicas: 1
    karmadaWebhook:
      imageRepository: docker.io
      imageTag: v1.4.0
      replicas: 1
    kubeControllerManager:
      imageRepository: registry.k8s.io
      imageTag: v1.25.2
      replicas: 1
  hostCluster:
    networking:
      dnsDomain: cluster.local
```

calvin0327:
@lonelyCZ @RainbowMango PTAL

operator/pkg/apis/operator/v1alpha1/type.go
```diff
@@ -89,31 +93,38 @@ type ImageRegistry struct {
...
 	KarmadaAggregratedAPIServer *KarmadaAggregratedAPIServer `json:"karmadaAggregratedAPIServer,omitempty"`
 
 	// KubeControllerManager holds settings to kube-controller-manager component of the karmada.
 	// +optional
+	// +kubebuilder:default:={imageRepository: "registry.k8s.io", imageTag: "v1.25.2", replicas: 1}
 	KubeControllerManager *KubeControllerManager `json:"kubeControllerManager,omitempty"`
 
 	// KarmadaControllerManager holds settings to karmada-controller-manager component of the karmada.
 	// +optional
+	// +kubebuilder:default:={imageRepository: "docker.io", imageTag: "v1.4.0", replicas: 1}
```
lonelyCZ, line 117: It seems inconvenient that we need to replace the default version with every new Karmada version released. Is there a better way to do it? How about using cmd options like `karmadactl init`？

calvin0327: I think we seem not need to add the imageTag field to every component, for user, they only care about the karmada version. 

whether we support global fields for karmada version?  and we calculate the corresponding to k8s version. It,s convenient for upgrading for users and us.  they only need to update one field, like this:

```golang
type KarmadaSpec struct {
  KarmadaVersion *string `json:"karmadaVersion,omitempty"`
}
```

calvin0327: oh.. It,s seem two questions. and I agress with your suggest use `option` to set default version for karmada. we can specify the default installtion version with a flag.

jwcesign:
> The main reason is that the webhook is need the tls to connect with apiserver, the means that before running karmada-operator pod, we need generate a certificate and mount to the pod. 

Why not use CSR(certificateSignRequest) to get a certificate?

calvin0327:
> > The main reason is that the webhook is need the tls to connect with apiserver, the means that before running karmada-operator pod, we need generate a certificate and mount to the pod.
> 
> Why not use CSR(certificateSignRequest) to get a certificate?

We must have a certificate before running the karmada operator pod,  I don't know  how to create CSR resource? could you talk about your approach?  thanks

jwcesign:
From my point of view, karmada operator webhook pod is deployed for validating operator crd, so we can create a csr when the webhook starts like: https://medium.com/trendyol-tech/tls-certificates-for-kubernetes-admission-webhooks-made-easy-with-certificator-and-helm-hook-89ece42fa193
img
The client-go will be responsible for certificate rotation. So, we don't need to generate a certificate in advance and mount it.

calvin0327:
@jwcesign ok, I got your means. there is a problem if the kubeconfig without  authority to `Apporve` the CRS, the webhook can not get a certification.

jwcesign:
> there is a problem if the kubeconfig without authority to Apporve the CSR, the webhook can not get a certification.
So the kubeconfig mount into webhhook hasn't this authority ?

calvin0327:
> 

It's possible to happen, because we don't know what a kubeconfig with authority that the user support to us.

jwcesign:
> It's possible to happen, because we don't know what a kubeconfig with authority that users support to us.

Make sense

lonelyCZ:
Let's move it forward.

/assign

lonelyCZ:
The default values can be  filled by operator controller, we will back this PR after #2990

Poor12:
Hi @calvin0327, is this pr still needed? Could we push it forward?

calvin0327:
@lonelyCZ @Poor12 
I want to close the pr, I still think it's a good idea to set the initial value in controller manager. Please see: https://github.com/karmada-io/karmada/pull/3374

PR #3025, opened 2023-01-06
Update protoc setup token

RainbowMango:
**What type of PR is this?**

/kind cleanup

Use the automatic token, so that the task(`Install Protoc`) can be run in the forked repo.

https://docs.github.com/en/actions/security-guides/automatic-token-authentication

```release-note
NONE
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,7 +44,9 @@ jobs:
         uses: arduino/setup-protoc@v1
         with:
           version: '3.x'
-          repo-token: ${{ secrets.GITHUB_TOKEN_FOR_SETUP_PROTOC }}
+          # Use the automatic token, so that this task can be run in the forked repo.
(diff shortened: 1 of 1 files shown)


XiShanYongYe-Chang (commented):
/lgtm

> The secret name GITHUB_TOKEN_FOR_SETUP_PROTOC that started with a GITHUB_ prefix is not recommended.

By the way, is this suggested document guidance?

RainbowMango:
It seems works after changing the secret to `secrets.GITHUB_TOKEN`.
https://github.com/karmada-io/karmada/actions/runs/3853748542/jobs/6567028412#step:4:4
```bash
Run arduino/setup-protoc@v1
  with:
    version: [3](https://github.com/karmada-io/karmada/actions/runs/3853748542/jobs/6567028412#step:4:3).x
    repo-token: ***
    include-pre-releases: false
```
Still don't know why the previous token not working...

PR #3027, opened 2023-01-06
stop proxy cache asynchronously

ikaven1024:

**What type of PR is this?**
/kind bug

When a member cluster is down, stopping the proxy cache for this cluster takes a few seconds(as below log, stop the pods cache takes 9s)

```
I0106 17:25:07.764975   62644 resource_cache.go:35] Stop store for yjh-1 /v1, Resource=pods
I0106 17:25:16.609796   62644 resource_cache.go:35] Stop store for yjh-1 /v1, Resource=nodes
```

During stopping, it hold the lock of `MultiClusterCache`. This results in client requests will be blocked. Because request handlers also need this lock:

https://github.com/karmada-io/karmada/blob/7b4c541bb818fb1e2677311aa34d8ff17b73d119/pkg/search/proxy/store/multi_cluster_cache.go#L162-L168

https://github.com/karmada-io/karmada/blob/7b4c541bb818fb1e2677311aa34d8ff17b73d119/pkg/search/proxy/store/multi_cluster_cache.go#L319-L321

Fixes #

```release-note
`karmada-search`: avoid proxy request block when member cluster down.
```

### DIFF
--- a/pkg/search/proxy/store/multi_cluster_cache.go
+++ b/pkg/search/proxy/store/multi_cluster_cache.go
@@ -7,6 +7,7 @@ import (
 	"sort"
 	"strings"
 	"sync"
+	"time"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
@@ -57,6 +58,13 @@ func NewMultiClusterCache(newClientFunc func(string) (dynamic.Interface, error),
 
 // UpdateCache update cache for multi clusters
 func (c *MultiClusterCache) UpdateCache(resourcesByCluster map[string]map[schema.GroupVersionResource]struct{}) error {
+	if klog.V(3).Enabled() {
+		start := time.Now()
+		defer func() {
+			klog.Infof("MultiClusterCache update cache takes %v", time.Since(start))
+		}()
+	}
+
 	c.lock.Lock()
 	defer c.lock.Unlock()
 
--- a/pkg/search/proxy/store/resource_cache.go
+++ b/pkg/search/proxy/store/resource_cache.go
@@ -33,7 +33,7 @@ type resourceCache struct {
 
 func (c *resourceCache) stop() {
 	klog.Infof("Stop store for %s %s", c.clusterName, c.resource)
-	c.Store.DestroyFunc()
+	go c.Store.DestroyFunc()
 }
 
 func newResourceCache(clusterName string, gvr schema.GroupVersionResource, gvk schema.GroupVersionKind,


jwcesign:
How long it takes now?

jwcesign:
There is RLock in `c.cache[resource].stop()`, so even with `go c.cache[resource].stop()`, it still runs serially. How it improves performance? Can you explain a little bit?

XiShanYongYe-Chang:
Hi @ikaven1024, does this change affect the access of other client requests to the cache resources of the cluster? 

I have another question, why not do it here:

https://github.com/karmada-io/karmada/blob/master/pkg/search/proxy/store/multi_cluster_cache.go#L71

ikaven1024:
> How long it takes now?

I add logs to measure the spending time:

```
I0109 12:12:00.792441   87527 multi_cluster_cache.go:64] MultiClusterCache update cache takes 162.416µs
```

ikaven1024:
> I have another question, why not do it here:
> 
> https://github.com/karmada-io/karmada/blob/master/pkg/search/proxy/store/multi_cluster_cache.go#L71

You are right. But when users remove resource from `ResourceRegistry` also will stop the cache via `clusterCache.updateCache`. So I move the `go` to the place `Cacher` really stopping:

```go
func (c *resourceCache) stop() {
	klog.Infof("Stop store for %s %s", c.clusterName, c.resource)
	go c.Store.DestroyFunc()
}
```

jwcesign:
> > How long it takes now?
> 
> I add logs to measure the spending time:
> 
> ```
> I0109 12:12:00.792441   87527 multi_cluster_cache.go:64] MultiClusterCache update cache takes 162.416µs
> ```

Great! 
/lgtm

pkg/search/proxy/store/multi_cluster_cache.go
```diff
@@ -57,6 +58,13 @@ func NewMultiClusterCache(newClientFunc func(string) (dynamic.Interface, error),
 
 // UpdateCache update cache for multi clusters
 func (c *MultiClusterCache) UpdateCache(resourcesByCluster map[string]map[schema.GroupVersionResource]struct{}) error {
+	if klog.V(3).Enabled() {
+		start := time.Now()
+		defer func() {
+			klog.V(3).Infof("MultiClusterCache update cache takes %v", time.Since(start))
+		}()
+	}
```
XiShanYongYe-Chang, line 66: Isn't it more concise:
```go
defer klog.V(3).Infof("MultiClusterCache update cache takes %v", time.Since(time.Now()))
```

XiShanYongYe-Chang: That doesn't seem to be correct.

XiShanYongYe-Chang:
/approve
Ask for @jwcesign to have a review again.

Hi @ikaven1024, do we need to cherry-pick this path to the previous release?

PR #3040, opened 2023-01-10
UpdateOrCreateToken get secrets err handling optimization

helen-frank:

**What type of PR is this?**
/kind feature

when UpdateOrCreateToken get secrets, if the error is not a NotFound, the error is thrown in time
Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/karmadactl/util/bootstraptoken/bootstraptoken.go
+++ b/pkg/karmadactl/util/bootstraptoken/bootstraptoken.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	kubeclient "k8s.io/client-go/kubernetes"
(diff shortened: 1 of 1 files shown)


pkg/karmadactl/util/bootstraptoken/bootstraptoken.go
```diff
@@ -320,6 +321,8 @@ func UpdateOrCreateToken(client kubeclient.Interface, failIfExists bool, token *
 	secret, err := client.CoreV1().Secrets(metav1.NamespaceSystem).Get(context.TODO(), secretName, metav1.GetOptions{})
 	if secret != nil && err == nil && failIfExists {
 		return fmt.Errorf("a token with id %q already exists", token.Token.ID)
+	} else if err != nil && !apierrors.IsNotFound(err) {
```
lonelyCZ, line 324: The other errors will be checked in `cmdutil.CreateOrUpdateSecret()`, it seems only preflight for `karmadactl token create` command and not to influence the CreateOrUpdateSecret phase.

And it refers to https://github.com/kubernetes/kubernetes/blob/cde45fb161c5a4bfa7cfe45dfd814f6cc95433f7/cmd/kubeadm/app/phases/bootstraptoken/node/token.go#L38

helen-frank: Yes, so I think this is an optimization feature rather than a bug.

helen-frank: @lonelyCZ https://github.com/kubernetes/kubernetes/pull/115000 kubeadm someone approved of my idea

pkg/karmadactl/util/bootstraptoken/bootstraptoken.go
```diff
@@ -318,7 +319,10 @@ func CreateNewToken(client kubeclient.Interface, token *BootstrapToken) error {
 func UpdateOrCreateToken(client kubeclient.Interface, failIfExists bool, token *BootstrapToken) error {
 	secretName := bootstraputil.BootstrapTokenSecretName(token.Token.ID)
 	secret, err := client.CoreV1().Secrets(metav1.NamespaceSystem).Get(context.TODO(), secretName, metav1.GetOptions{})
-	if secret != nil && err == nil && failIfExists {
+	if err != nil && !apierrors.IsNotFound(err) {
+		return err
+	}
```
lonelyCZ, line 324: Great, this is also my point, it is clearer.

pkg/karmadactl/util/bootstraptoken/bootstraptoken.go
lonelyCZ: Right, if the err is not nil, the `already exist` error will be skipped that could cause the unexpected result. This patch make it safer.

RainbowMango (approved):
/remove-kind feature
/kind cleanup
/approve

PR #3044, opened 2023-01-12
fix dependencies distributor buildAttachedBinding Namespace

yanfeng1992:

**What type of PR is this?**
/kind bug

When creating a dependent resourceBinding, the namespace should use the namespace of the object itself to support cross-namespace references to dependent resources.

For example, we have a crd virtualmachine, create virtualmachine cr test in ns defalut, webhook get virtualmachine cr test dependece configmap test-cm in ns test-2。

Fixes #
dependencies distributor buildAttachedBinding namespace need 

```release-note

`karmada-controller-manager`:  fix build attached resourceBinding namespace

```

### DIFF
--- a/pkg/dependenciesdistributor/dependencies_distributor.go
+++ b/pkg/dependenciesdistributor/dependencies_distributor.go
@@ -2,10 +2,12 @@ package dependenciesdistributor
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
+	"hash/fnv"
+	"strings"
 	"time"
 
+	"github.com/google/go-cmp/cmp"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
@@ -15,6 +17,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/rand"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/dynamic"
@@ -34,6 +37,7 @@ import (
 	"github.com/karmada-io/karmada/pkg/util/fedinformer"
 	"github.com/karmada-io/karmada/pkg/util/fedinformer/genericmanager"
 	"github.com/karmada-io/karmada/pkg/util/fedinformer/keys"
+	hashutil "github.com/karmada-io/karmada/pkg/util/hash"
 	"github.com/karmada-io/karmada/pkg/util/helper"
 	"github.com/karmada-io/karmada/pkg/util/names"
 )
@@ -42,9 +46,9 @@ const (
 	// bindingDependedByLabelKeyPrefix is the prefix to a label key specifying an attached binding referred by which independent binding.
 	// the key is in the label of an attached binding which should be unique, because resource like secret can be referred by multiple deployments.
 	bindingDependedByLabelKeyPrefix = "resourcebinding.karmada.io/depended-by-"
-	// bindingDependenciesAnnotationKey represents the key of dependencies data (json serialized)
-	// in the annotations of an independent binding.
-	bindingDependenciesAnnotationKey = "resourcebinding.karmada.io/dependencies"
+	//bindingDependedOnLabelKeyPrefix is the prefix to a label key specifying a binding depended on one object
+	// the key is in the label of a binding which should be unique
+	bindingDependedOnLabelKeyPrefix = "resourcebinding.karmada.io/depended-on-"
 )
 
 var supportedTypes = []schema.GroupVersionResource{
@@ -170,7 +174,9 @@ func (d *DependenciesDistributor) Reconcile(key util.QueueKey) error {
 	}
 	klog.V(4).Infof("DependenciesDistributor start to reconcile object: %s", clusterWideKey)
 
-	bindingObjectList, err := d.resourceBindingLister.ByNamespace(clusterWideKey.Namespace).List(labels.Everything())
+	bindingDependedLabelKey := generateBindingDependedLabelKey(clusterWideKey.Kind, clusterWideKey.Version, clusterWideKey.Name, clusterWideKey.Namespace)
+	selector := labels.SelectorFromSet(map[string]string{bindingDependedLabelKey: "true"})
+	bindingObjectList, err := d.resourceBindingLister.List(selector)
 	if err != nil {
 		return err
 	}
@@ -185,17 +191,6 @@ func (d *DependenciesDistributor) Reconcile(key util.QueueKey) error {
 		if !binding.DeletionTimestamp.IsZero() {
 			continue
 		}
-
-		matched, err := dependentObjectReferenceMatches(clusterWideKey, binding)
-		if err != nil {
-			klog.Errorf("Failed to evaluate if binding(%s/%s) need to sync dependencies: %v", binding.Namespace, binding.Name, err)
-			errs = append(errs, err)
-			continue
-		} else if !matched {
-			klog.V(4).Infof("No need to sync binding(%s/%s)", binding.Namespace, binding.Name)
-			continue
-		}
-
 		klog.V(4).Infof("Resource binding(%s/%s) is matched for resource(%s/%s)", binding.Namespace, binding.Name, clusterWideKey.Namespace, clusterWideKey.Name)
 		bindingKey, err := detector.ClusterWideKeyFunc(binding)
 		if err != nil {
@@ -208,35 +203,6 @@ func (d *DependenciesDistributor) Reconcile(key util.QueueKey) error {
 	}
 
 	return utilerrors.NewAggregate(errs)
-}
-
-// dependentObjectReferenceMatches tells if the given object is referred by current resource binding.
-func dependentObjectReferenceMatches(objectKey keys.ClusterWideKey, referenceBinding *workv1alpha2.ResourceBinding) (bool, error) {
-	dependencies, exist := referenceBinding.Annotations[bindingDependenciesAnnotationKey]
-	if !exist {
-		return false, nil
-	}
-
-	var dependenciesSlice []configv1alpha1.DependentObjectReference
-	err := json.Unmarshal([]byte(dependencies), &dependenciesSlice)
-	if err != nil {
-		return false, err
-	}
-
-	if len(dependenciesSlice) == 0 {
-		return false, nil
-	}
-
-	for _, dependence := range dependenciesSlice {
-		if objectKey.Version == dependence.APIVersion &&
-			objectKey.Kind == dependence.Kind &&
-			objectKey.Namespace == dependence.Namespace &&
-			objectKey.Name == dependence.Name {
-			return true, nil
-		}
-	}
-
-	return false, nil
 }
 
 // OnResourceBindingUpdate handles object update event and push the object to queue.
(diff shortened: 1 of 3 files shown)


jwcesign:
Thanks!
So for one crd in default ns, it depends on the configmap in ns-2.
The configmap's rb will be created in default ns for the previous implementation.
After this patch, the configmap's rb will be created in ns-2 ns, Do I understand correctly?

yanfeng1992:
> Thanks! So for one crd in default ns, it depends on the configmap in ns-2. The configmap's rb will be created in default ns for the previous implementation. After this patch, the configmap's rb will be created in ns-2 ns, Do I understand correctly?

@jwcesign yes, that's right。 For the previous implementation,  configmap's rb be created in default ns ,  config map ceate in ns-2 ns,  configmap's rb will be a Orphan binding.  https://github.com/yanfeng1992/karmada/blob/85fcf2f627dd12cbf9849b49090451dcaa7b2a29/pkg/dependenciesdistributor/dependencies_distributor.go#L453  Do I understand correctly?

yanfeng1992:
> Hi @yanfeng1992, I have a question, why do you need to use resources across ns?

for example，we need a pv, pv is provided by another company， pv provider company configmap in specific ns。

Doesn't karmada plan to support  use resources across ns?

XiShanYongYe-Chang:
> for example，we need a pv, pv is provided by another company， pv provider company configmap in specific ns。

Thanks for your explanation!

> Doesn't karmada plan to support use resources across ns? @XiShanYongYe-Chang

No, I just wanted to know the scene.

XiShanYongYe-Chang:
Hi @yanfeng1992, can you help add an E2E to cover this case?

yanfeng1992:
i add an E2E case cover this case, but  but ran into a problem. After the configmap is updated, it cannot be synchronized to the member cluster. need your help. @XiShanYongYe-Chang 


image

XiShanYongYe-Chang:
> i add an E2E case cover this case, but but ran into a problem. After the configmap is updated, it cannot be synchronized to the member cluster. need your help. @XiShanYongYe-Chang

Let me take a look.

yanfeng1992:
https://github.com/yanfeng1992/karmada/blob/69a1b993415fe13633c41bdcd95ca6621d67b8b5/pkg/dependenciesdistributor/dependencies_distributor.go#L173  @XiShanYongYe-Chang I found the problem, can we list all namespace?

pkg/dependenciesdistributor/dependencies_distributor.go
```diff
@@ -170,7 +170,7 @@ func (d *DependenciesDistributor) Reconcile(key util.QueueKey) error {
 	}
 	klog.V(4).Infof("DependenciesDistributor start to reconcile object: %s", clusterWideKey)
 
-	bindingObjectList, err := d.resourceBindingLister.ByNamespace(clusterWideKey.Namespace).List(labels.Everything())
+	bindingObjectList, err := d.resourceBindingLister.List(labels.Everything())
```
XiShanYongYe-Chang, line 173: Hi @yanfeng1992, I think you have found the root cause of the problem. 

But this change can have a big impact, because the total number of `ResourceBinding` may be large, which makes me a little hesitant to support this feature (Yes, I think the current patch is a new feature, not a bug fix), or we need to add a flag to enable this feature.

Ask more people for more ideas. @RainbowMango @mrlihanbo @jwcesign

test/helper/resource.go
```diff
@@ -302,6 +302,29 @@ func NewCustomResource(apiVersion, kind, namespace, name string) *unstructured.U
 	}
 }
 
+// NewCustomResourceAnother will build a CR object depended on across ns ConfigMap.
+func NewCustomResourceAnother(apiVersion, kind, namespace, name, configMapName string) *unstructured.Unstructured {
```
jwcesign, line 306: NewCustomResourceWithConfigMap better?

pkg/dependenciesdistributor/dependencies_distributor.go
jwcesign: I more like giving some labels to the dependency resource's rb, then using label selector to select the bindingObject list.

yanfeng1992:
> Hi @yanfeng1992, can you help add an E2E to cover this case?

E2E done @XiShanYongYe-Chang

pkg/dependenciesdistributor/dependencies_distributor.go
XiShanYongYe-Chang: Hi @yanfeng1992, how about finding the target `ResourceBinding` through the following path?

resource template -> ResourceBinding of resource template -> dependent ResourceBinding

RainbowMango:
Could you please [squash your commits](https://karmada.io/docs/contributor/github-workflow#squash-commits)?

yanfeng1992:
> Could you please [squash your commits](https://karmada.io/docs/contributor/github-workflow#squash-commits)?

done @RainbowMango

pkg/dependenciesdistributor/dependencies_distributor.go
yanfeng1992: > Hi @yanfeng1992, how about finding the target `ResourceBinding` through the following path?
> 
> resource template -> ResourceBinding of resource template -> dependent ResourceBinding

i try to find bindobject.Find dependent ResourceBinding according to bindingObject.Spec.RequiredBy, is there anything wrong? @XiShanYongYe-Chang

XiShanYongYe-Chang: The newly added e2e seems to run failed.

XiShanYongYe-Chang: You can get the log from here: https://github.com/karmada-io/karmada/suites/10409837217/artifacts/514337324

At the bottom of this page: https://github.com/karmada-io/karmada/actions/runs/3937200810

XiShanYongYe-Chang:
> Hi @yanfeng1992, how about finding the target ResourceBinding through the following path?
>
> resource template -> ResourceBinding of resource template -> dependent ResourceBinding

Hi @yanfeng1992, first of all, I apologize for my wrong advice. I tried to solve the problem using the above thinking, but I found that there was one situation that I couldn't solve: the resource creation situation. In this case, the target `ResourceBinding` cannot be found. 
Solving this problem may still require you to go through all the `ResourceBinding` objects the way you did before.

But back to the previous question, is it worth it? Ask @RainbowMango for help.

yanfeng1992:
Hi @XiShanYongYe-Chang @RainbowMango @jwcesign 
I think I have solved this problem, through the way of increasing lable mentioned by jwcesign. 

There are two advantages of adding labels for resourceBinding in this way
The first, can support support use resources across ns
The second,reduces the number of rb processing

XiShanYongYe-Chang:
Hi @yanfeng1992, I feel that this solution is still a little incomplete, and if there are a large number of `ResourceBinding` that meet the requirements, the time consumption is still large.

In addition, the problem itself is that cross-namespace resource distribution is likely to encounter permissions issues, because for some users, he may not have permissions on all namespaces. In this case, distributing resources across namespaces may not work.

> for example，we need a pv, pv is provided by another company， pv provider company configmap in specific ns。

For this application scenario, ask for @RainbowMango to take a look.

yanfeng1992:
> > Hi @yanfeng1992, how about finding the target ResourceBinding through the following path?
> > resource template -> ResourceBinding of resource template -> dependent ResourceBinding
> 
> Hi @yanfeng1992, first of all, I apologize for my wrong advice. I tried to solve the problem using the above thinking, but I found that there was one situation that I couldn't solve: the resource creation situation. In this case, the target `ResourceBinding` cannot be found. Solving this problem may still require you to go through all the `ResourceBinding` objects the way you did before.
> 
> But back to the previous question, is it worth it? Ask @RainbowMango for help.



> > Hi @yanfeng1992, how about finding the target ResourceBinding through the following path?
> > resource template -> ResourceBinding of resource template -> dependent ResourceBinding
> 
> Hi @yanfeng1992, first of all, I apologize for my wrong advice. I tried to solve the problem using the above thinking, but I found that there was one situation that I couldn't solve: the resource creation situation. In this case, the target `ResourceBinding` cannot be found. Solving this problem may still require you to go through all the `ResourceBinding` objects the way you did before.
> 
> But back to the previous question, is it worth it? Ask @RainbowMango for help.



> > Hi @yanfeng1992, how about finding the target ResourceBinding through the following path?
> > resource template -> ResourceBinding of resource template -> dependent ResourceBinding
> 
> Hi @yanfeng1992, first of all, I apologize for my wrong advice. I tried to solve the problem using the above thinking, but I found that there was one situation that I couldn't solve: the resource creation situation. In this case, the target `ResourceBinding` cannot be found. Solving this problem may still require you to go through all the `ResourceBinding` objects the way you did before.
> 
> But back to the previous question, is it worth it? Ask @RainbowMango for help.

@XiShanYongYe-Chang

yanfeng1992:
> > Hi @yanfeng1992, how about finding the target ResourceBinding through the following path?
> > resource template -> ResourceBinding of resource template -> dependent ResourceBinding
> 
> Hi @yanfeng1992, first of all, I apologize for my wrong advice. I tried to solve the problem using the above thinking, but I found that there was one situation that I couldn't solve: the resource creation situation. In this case, the target `ResourceBinding` cannot be found. Solving this problem may still require you to go through all the `ResourceBinding` objects the way you did before.
> 
> But back to the previous question, is it worth it? Ask @RainbowMango for help.

hi @XiShanYongYe-Chang @RainbowMango 


> > Hi @yanfeng1992, how about finding the target ResourceBinding through the following path?
> > resource template -> ResourceBinding of resource template -> dependent ResourceBinding

Your previous suggestion is correct. There was a bit of a problem with my previous implementation, which I have fixed.
the resource creation situation. In this case, the target `ResourceBinding` cannot be found.
And you mentioned this question: ‘the resource creation situation. In this case, the target `ResourceBinding` cannot be found.’, I don't think it has any effect.

I think there are two advantages to doing this now

The first, can support support use resources across ns.
The second,reduces the number of rb processing. There is no need to judge whether there is a reference relationship based on the annotation.

In addition, why do we need to use resources across namespaces. When PV is mounted, xsky needs a cm. This cm is in a fixed ns. One pv corresponds to one cm, and other vendors’ csi implementations don’t seem to have this problem

yanfeng1992:
Do we need to continue discussing this issue? Should it be considered a new feature rather than a bug? @XiShanYongYe-Chang

XiShanYongYe-Chang:
> Do we need to continue discussing this issue? Should it be considered a new feature rather than a bug? @XiShanYongYe-Chang

I think it's a new feature. How about adding a talk in the today's community meeting: https://docs.google.com/document/d/1y6YLVC-v7cmVAdbjedoyR5WL0-q45DBRXTvz5_I7bkA/edit#heading=h.g61sgp7w0d0c

yanfeng1992:
> > Do we need to continue discussing this issue? Should it be considered a new feature rather than a bug? @XiShanYongYe-Chang
> 
> I think it's a new feature. How about adding a talk in the today's community meeting: https://docs.google.com/document/d/1y6YLVC-v7cmVAdbjedoyR5WL0-q45DBRXTvz5_I7bkA/edit#heading=h.g61sgp7w0d0c

OK, we can join the meeting, do you want me to edit this document?

XiShanYongYe-Chang:
> OK, we can join the meeting, do you want me to edit this document?

Yes, you can edit it directly.

XiShanYongYe-Chang:
> And you mentioned this question: ‘the resource creation situation. In this case, the target `ResourceBinding` cannot be found.’, I don't think it has any effect.

I just looked at the implementation, and I feel that this problem has not been solved.

yanfeng1992:
> > And you mentioned this question: ‘the resource creation situation. In this case, the target `ResourceBinding` cannot be found.’, I don't think it has any effect.
> 
> I just looked at the implementation, and I feel that this problem has not been solved.
I feel that in the previous implementation，the resource creation situation the target `ResourceBinding` aslo cannot be found.

XiShanYongYe-Chang:
> I feel that in the previous implementation，the resource creation situation the target `ResourceBinding` aslo cannot be found. @XiShanYongYe-Chang

When a resource is created, it queues all matched rbs.

yanfeng1992:
I understand what you mean. 
For example,we created a deployment that depends on configmap.
My implementation can't cover the following situation:
deployment  rb created, configmap does not exist. when configmap is created, the target ResourceBinding cannot be found.
Do i understand correctly?

XiShanYongYe-Chang:
> I understand what you mean.
> For example,we created a deployment that depends on configmap.
> My implementation can't cover the following situation:
> deployment rb created, configmap does not exist. when configmap is created, the target ResourceBinding cannot be found.
> Do i understand correctly?

Yes, it is.

XiShanYongYe-Chang:
Hi @yanfeng1992, what are your plans for the issue now? Pack the whole application?

yanfeng1992:
> Hi @yanfeng1992, what are your plans for the issue now? Pack the whole application?

There is currently no plan to package it into an application. I am now trying to solve the problem by adding a dependent resource label in rb @XiShanYongYe-Chang

yanfeng1992:
Please take a look at my latest commit. I feel that the resource creation situation not find the target ResourceBinding has been solved.@XiShanYongYe-Chang

XiShanYongYe-Chang:
Hi @yanfeng1992, this patch is functionally satisfactory. However, this solution may be different from the one we discussed at the meeting on Tuesday ([2023-01-31](https://docs.google.com/document/d/1y6YLVC-v7cmVAdbjedoyR5WL0-q45DBRXTvz5_I7bkA/edit)). One of the considerations is the authority, which I think should also be taken into account.

yanfeng1992:
Thank you very much for the help of the Karmada community. I think packaging applications is a better way. Using resources across namespaces is not a reasonable and common scenario. I think this issue can be closed. What are your opinions?
@XiShanYongYe-Chang @RainbowMango

XiShanYongYe-Chang:
@yanfeng1992 Thanks for your hard trying! 👍

RainbowMango:
> Using resources across namespaces is not a reasonable and common scenario. 

My primary concern must be security.  Karmada and K8S's multi-tenant concepts are based on namespaces, if a resource depends on a resource in another namespace, it'll cause a potential security risk. 

Anyway, thanks for spotting and this is definitely a valuable discussion. 

For your case, my suggestion would be `try to feed your workload with a  config in the same namespace`, or declare the config in PropagationPolicy explicitly.

PR #3053, opened 2023-01-16
Adds errorlint checks to golangci-lint, standardizing the error format and error judgments

Fish-pro:

**What type of PR is this?**

/kind feature

Add an errorlint check to make it easier to provide standard error output

Fixes #

```release-note
NONE
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -35,6 +35,7 @@ linters:
   - govet
   # linters default enabled by golangci-lint .
   - errcheck
+  - errorlint
   - gosimple
   - ineffassign
   - staticcheck
--- a/pkg/util/lifted/discovery.go
+++ b/pkg/util/lifted/discovery.go
@@ -38,6 +38,8 @@ import (
 func GetDeletableResources(discoveryClient discovery.ServerResourcesInterface) map[schema.GroupVersionResource]struct{} {
 	preferredResources, err := discoveryClient.ServerPreferredResources()
 	if err != nil {
+		//nolint:errorlint
+		// disable `error format` check for backward compatibility.
 		if discovery.IsGroupDiscoveryFailedError(err) {
 			klog.Warningf("Failed to discover some groups: %v", err.(*discovery.ErrGroupDiscoveryFailed).Groups)
 		} else {
--- a/cmd/aggregated-apiserver/app/options/options.go
+++ b/cmd/aggregated-apiserver/app/options/options.go
@@ -110,7 +110,7 @@ func (o *Options) Run(ctx context.Context) error {
 func (o *Options) Config() (*aggregatedapiserver.Config, error) {
 	// TODO have a "real" external address
 	if err := o.RecommendedOptions.SecureServing.MaybeDefaultWithSelfSignedCerts("localhost", nil, []net.IP{netutils.ParseIPSloppy("127.0.0.1")}); err != nil {
-		return nil, fmt.Errorf("error creating self-signed certificates: %v", err)
+		return nil, fmt.Errorf("error creating self-signed certificates: %w", err)
 	}
 
 	o.RecommendedOptions.Etcd.StorageConfig.Paging = utilfeature.DefaultFeatureGate.Enabled(features.APIListChunking)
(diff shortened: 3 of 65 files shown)


XiShanYongYe-Chang (commented):
Hi @Fish-pro, these modifications are after the `errorlint` is added, according to the error prompt modified?

Fish-pro:
> Hi @Fish-pro, these modifications are after the `errorlint` is added, according to the error prompt modified?

@XiShanYongYe-Chang Yes, you can see that all of the changes are in accordance with the `errlint` specification

pkg/util/lifted/discovery.go
```diff
@@ -39,7 +41,10 @@ func GetDeletableResources(discoveryClient discovery.ServerResourcesInterface) m
 	preferredResources, err := discoveryClient.ServerPreferredResources()
 	if err != nil {
 		if discovery.IsGroupDiscoveryFailedError(err) {
-			klog.Warningf("Failed to discover some groups: %v", err.(*discovery.ErrGroupDiscoveryFailed).Groups)
+			var tErr *discovery.ErrGroupDiscoveryFailed
```
RainbowMango, line 44: This is the `lifted` code, I'm hesitant to change it unless there is a good reason.
Can we contribute to the upstream first?

.golangci.yml
```diff
@@ -35,6 +35,7 @@ linters:
   - govet
   # linters default enabled by golangci-lint .
   - errcheck
+  - errorlint
```
RainbowMango, line 38: This linter is not enabled by default as per https://golangci-lint.run/usage/linters/#enabled-by-default.

pkg/util/lifted/discovery.go
Fish-pro: Thanks for pointing out the question of whether the lifted code can be interpreted to mean that it will never be modified

.golangci.yml
Fish-pro: `errorlint` is disabled by default. I have seen both `fmt.Errorf("%s",err.Error)` and `fmt.Errorf("%v",err)` exist, so it is considered for better code specification. If the community feels it is not necessary, I will close the merge request

pkg/util/lifted/discovery.go
RainbowMango: The lifted code can be touched if there is a need(like fixing a bug). 
For the errorlint check, I'm not sure if worth to do it according to the benefits we can get from this.

.golangci.yml
RainbowMango: I mean `errorlint` should be grouped into [other linters supported by golangci-lint](https://github.com/karmada-io/karmada/blob/2211ce63e4dd331dcda119f845dc8d3585794219/.golangci.yml#L43).

RainbowMango (commented):
I'm not sure if we should introduce the `errorlint` thing. 

My concern is it is probably not a common code specification, why do we do that restriction? More restrictions would block contributors and makes them feel sad. 

But I'm not saying no to the approach, I'd like to invite you to the community meeting to talk about this because I'm not quite similar about the `errorlint` as well as the benefits from it.

Fish-pro:
> I'm not sure if we should introduce the `errorlint` thing.
> 
> My concern is it is probably not a common code specification, why do we do that restriction? More restrictions would block contributors and makes them feel sad.
> 
> But I'm not saying no to the approach, I'd like to invite you to the community meeting to talk about this because I'm not quite similar about the `errorlint` as well as the benefits from it.

@RainbowMango Thank you. I see what you mean

Fish-pro:
Disable `error format` check for backward compatibility in lifted

Fish-pro:
@RainbowMango 
errorlint consists of three main types of code specifications:
1. format error in `fmt`
  fmt.Errorf("this is error: %s", err)-----> fmt.Errorf("this is error: %w", err)
  fmt.Errorf("this is error: %v", err)-----> fmt.Errorf("this is error: %w", err)
  fmt.Errorf("this is error: %s", err.Error())-----> fmt.Errorf("this is error: %w", err)
2. Error comparison uses `errors.Is()`
   err == io.EOF -------> errors.Is(err, io.EOF)
3. Error type conversion uses `errors.As()`
   err.(*tt.WantErr) -----> errors.As(err, &tt.WantErr{})

PR #3058, opened 2023-01-17
Add devcontainer

lgarcia11:
**What type of PR is this?**
/kind feature

There are multiple issues when running hack/local-up-karmada.sh. Running it from inside a controlled environment will make easier to get started. 

May need to add more tools for actual development.

NONE


RainbowMango:
@lgarcia11 Thanks for doing this. Can you explain it in more detail? How to use it?

PR #3079, opened 2023-01-29
ci: Cancel when there is a new change

wzshiming:
**What type of PR is this?**

/kind cleanup

Fixes #

Running a full test every time push is unnecessary, and this patch will cancel CI when new changes are made.

```release-note
NONE
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,10 @@ on:
   # Run workflow on fork repository will help contributors find and resolve issues before sending a PR.
   push:
   pull_request:
+# https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency
+concurrency:
+  group: ${{ github.workflow }}-${{ github.actor }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
 jobs:
   golangci:
     name: lint


.github/workflows/ci.yml
```diff
@@ -24,6 +24,10 @@ jobs:
   codegen:
     name: codegen
     runs-on: ubuntu-20.04
+    # https://docs.github.com/en/actions/using-jobs/using-concurrency
+    concurrency:
+      group: codegen-${{ github.workflow }}-${{ github.head_ref || github.run_id }}
```
jwcesign, line 29: in what situation value ${{ github.run_id }} will be selected?

wzshiming: https://docs.github.com/en/actions/using-jobs/using-concurrency#example-using-a-fallback-value

The `github.head_ref` is only defined on pull_request events. if `github.head_ref` is undefined, the concurrency group will fallback to the `github.run_id`, which is guaranteed to be both unique and defined for the run.

jwcesign:
So the core benefit is that it will terminate the old changes' CI workflow and give more resources to other PRs(making pending time shorter).

Very nice work, cc @RainbowMango for checking.

XiShanYongYe-Chang (commented):
Hi @wzshiming, thank you for finding this great way.
And I have a question, why not consider setting `concurrency` at the workflow level directly? https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency

It seems more effective to operate like this.

wzshiming:
You make a lot of sense, and I can't believe I didn't think of that. 😄

wzshiming:
Updated, I tested that work.

https://github.com/karmada-io/karmada/actions/runs/4035746991

XiShanYongYe-Chang (commented):
Thanks a lot~
/lgtm

.github/workflows/ci.yml
```diff
@@ -4,6 +4,10 @@ on:
   # Run workflow on fork repository will help contributors find and resolve issues before sending a PR.
   push:
   pull_request:
+# https://docs.github.com/en/actions/using-jobs/using-concurrency
```
RainbowMango, line 7: The link should be https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency since we are setting the concurrency on the workflow level(not jobs.)
```suggestion
# https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency
```

.github/workflows/ci.yml
```diff
@@ -4,6 +4,10 @@ on:
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
```
RainbowMango, line 9: What's the value of `github.head_ref` exactly? I didn't get it from the [github-context](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context):
> The head_ref or source branch of the pull request in a workflow run.

And I can't get an example from the documents.

wzshiming: https://github.com/karmada-io/karmada/pull/3079#discussion_r1089875933

> https://docs.github.com/en/actions/using-jobs/using-concurrency#example-using-a-fallback-value
> The github.head_ref is only defined on pull_request events. if github.head_ref is undefined, the concurrency group will fallback to the github.run_id, which is guaranteed to be both unique and defined for the run.
> Each PR has a unique github.head_ref, which ensures that it will be cancelled correctly, and for non-PR triggers ensures that they will not be cancelled.

RainbowMango: Can you give the example of the group name for both `push` and `pull_request` events?
We should ensure that each `pull request` uses a unique group name so that it won't cancel the workflows that are triggered by `push` event.

wzshiming: It is in the [github-context](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context).

```
github.head_ref | string | The head_ref or source branch of the pull request in a workflow run. This property is only available when the event that triggers a workflow run is either pull_request or pull_request_target.
```

wzshiming: echo ${{ github.workflow }}-${{ github.head_ref }}-${{ github.run_id }}

```
CI Workflow-ci/cancel-4071108226
```

wzshiming: github.workflow: is the `.name` of the `ci.yml`
github.head_ref: is the branch of the pr
github.run_id: is the job run id

wzshiming: I added `github.actor` to avoid wrong cancellation due to different users but same branch name

RainbowMango: > https://github.com/karmada-io/karmada/actions/runs/4071108226/jobs/7012597471
> 
> `echo ${{ github.workflow }}-${{ github.head_ref }}-${{ github.run_id }}`
> 
> ```
> CI Workflow-ci/cancel-4071108226
> ```

Do you mean we can get the group name from logs?

wzshiming: Yeash

RainbowMango:
Before moving this PR forward, I still want to clarify what the group name exactly looks like for both `PR event` and `push event`.
My concerns are:
- Does a PR's workflow would cancel the workflow triggered by a `push event`?
- Does a push event workflow would cancel the workflow triggered by another `push event`?

wzshiming:
> Does a PR's workflow would cancel the workflow triggered by a push event?

that will cancel the last workflow

> Does a push event workflow would cancel the workflow triggered by another push event?

Same behavior as the original, that not cancel any workflow.

XiShanYongYe-Chang:
image

XiShanYongYe-Chang:
I think we can move this pr on.

I test this pr on my own karmada registry, it is found that:

for `push event`, the group is like this:
```log
# master ci workflow log
Run echo CI Workflow-XiShanYongYe-Chang-4313695314

# after pr test01 is merged, ci workflow log
Run echo CI Workflow-XiShanYongYe-Chang-4313795672

# after pr test02 is merged, ci workflow log
Run echo CI Workflow-XiShanYongYe-Chang-4313798490
```

for `PR event`, the group is like this:
```log
# when pr test01 is make a pull request
Run echo CI Workflow-XiShanYongYe-Chang-debug-echo-group-test01

# when pr test02 is make a pull request
Run echo CI Workflow-XiShanYongYe-Chang-debug-echo-group-test02
```

According to the above log, `head_ref` is the source branch name of pr, `github.head_ref` is only defined on `pull_request` events, for the `push` event, it will be empty. And `github.run_id` is unique for every running workflow.

RainbowMango:
Thanks @XiShanYongYe-Chang for the demo that removed all my concerns.

/approve

And thanks @wzshiming for spotting, it must be very useful!

PR #3081, opened 2023-01-30
add UT for GetMinTolerationTime and GetMatchingTolerations

wongearl:
Signed-off-by: wongearl <vamliang@163.com>

**What type of PR is this?**

/kind cleanup
add UT for GetMinTolerationTime and GetMatchingTolerations
Subtask of https://github.com/karmada-io/karmada/issues/2579

```release-note
NONE
```

### DIFF
--- a/pkg/util/helper/taint_test.go
+++ b/pkg/util/helper/taint_test.go
@@ -4,9 +4,11 @@ import (
 	"context"
 	"reflect"
 	"testing"
+	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	fakeclient "sigs.k8s.io/controller-runtime/pkg/client/fake"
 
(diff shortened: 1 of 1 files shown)


pkg/util/helper/taint_test.go
```diff
@@ -357,3 +359,209 @@ func TestGetNoExecuteTaints(t *testing.T) {
...
+						TimeAdded: nil,
+					},
+				},
+				tolerations: nil,
+			},
+			want:  false,
+			want1: []corev1.Toleration{},
+		},
+		{
+			name: "test return tolerations",
```
XiShanYongYe-Chang, line 517: How about changing the test case name to `don't exist mismatch taint`?

And Adding a case `exist mismatch taint` case will be better.

pkg/util/helper/taint_test.go
```diff
@@ -357,3 +359,209 @@ func TestGetNoExecuteTaints(t *testing.T) {
...
+		noExecuteTaints []corev1.Taint
+		usedTolerations []corev1.Toleration
+	}
+	tests := []struct {
+		name string
+		args args
+		want time.Duration
+	}{
+		{
+			name: "test no noExecute taints",
```
XiShanYongYe-Chang, line 376: How about changing the test case name to `the length of taints is 0`? Same as the other case names.

pkg/util/helper/taint_test.go
```diff
@@ -357,3 +359,209 @@ func TestGetNoExecuteTaints(t *testing.T) {
...
+						Effect:            "NoExecute",
+						TolerationSeconds: pointer.Int64(10000),
+					},
+					{
+						Key:               "key",
+						Operator:          "Equal",
+						Value:             "value",
+						Effect:            "NoExecute",
+						TolerationSeconds: pointer.Int64(20000),
+					},
```
XiShanYongYe-Chang, line 450: Those taints and tolerations have the same key, value and operator, how about making them a little different?

pkg/util/helper/taint_test.go
```diff
@@ -357,3 +359,209 @@ func TestGetNoExecuteTaints(t *testing.T) {
...
+				},
+			},
+			want: 9999999999999,
```
XiShanYongYe-Chang, line 453: I test the case failed on my local side:
```
taint_test.go:459: GetMinTolerationTime() = 10000000, want 9999999
```

I think it's best that we don't use the current time when we use the time: 

```go
# line 364
now := metav1.Now()
```

pkg/util/helper/taint_test.go
```diff
@@ -357,3 +359,209 @@ func TestGetNoExecuteTaints(t *testing.T) {
...
 }
+
+func TestGetMinTolerationTime(t *testing.T) {
+	now := metav1.Now()
+	later := metav1.NewTime(time.Now().Add(5000))
+	type args struct {
+		noExecuteTaints []corev1.Taint
+		usedTolerations []corev1.Toleration
+	}
+	tests := []struct {
```
XiShanYongYe-Chang, line 370: I think it's necessary to add a use case to this test `exist mismatch taint`.

The current method may not take this into account in its implementation, and it assumes that all the taints are matched, which is actually the logic in the program, but as a tool method, this assumption is not reasonable, so the implementation of the method may need to be modified.

XiShanYongYe-Chang (commented):
Hi @wongearl, thanks a lot~

XiShanYongYe-Chang:
Kindly ping @wongearl

pkg/util/helper/taint_test.go
wongearl: ok

pkg/util/helper/taint_test.go
wongearl: ok

pkg/util/helper/taint_test.go
wongearl: ok

pkg/util/helper/taint_test.go
wongearl: ok

pkg/util/helper/taint_test.go
```diff
@@ -357,3 +359,283 @@ func TestGetNoExecuteTaints(t *testing.T) {
...
+					{
+						Key:               "key5",
+						Operator:          "Equal",
+						Value:             "value5",
+						Effect:            "NoExecute",
+						TolerationSeconds: pointer.Int64(20000),
+					},
+				},
+			},
+			want: -1,
```
XiShanYongYe-Chang, line 499: This is not the expected value. The expected value should be 0. We need to modify the test method to meet this test case.

XiShanYongYe-Chang (commented):

wongearl:
sorry，I close it

PR #3090, opened 2023-02-01
Change score logic from serial to parallel

Poor12:

**What type of PR is this?**
/kind cleanup

1. Incorrect uasge of variables:
pluginToClusterScores := make(framework.PluginToClusterScores, len(frw.filterPlugins)) => pluginToClusterScores := make(framework.PluginToClusterScores, len(frw.scorePlugins))

2. Change score logic from serial to parallel. 
It will improve performance when the number of clusters is large or there are many custom scoring plugins.

Fixes #

None


pkg/scheduler/framework/runtime/framework.go
```diff
@@ -102,38 +115,70 @@ func (frw *frameworkImpl) RunScorePlugins(ctx context.Context, placement *policy
...
+
+	// Run Score method for each cluster in parallel
+	frw.parallelizer.Until(ctx, len(clusters), func(index int) {
+		for _, p := range frw.scorePlugins {
+			cluster := clusters[index]
 			s, res := frw.runScorePlugin(ctx, p, placement, spec, cluster)
 			if !res.IsSuccess() {
-				return nil, framework.AsResult(fmt.Errorf("plugin %q failed with: %w", p.Name(), res.AsError()))
+				err := fmt.Errorf("plugin %q failed with: %w", p.Name(), res.AsError())
+				errCh.SendErrorWithCancel(err, cancel)
```
jwcesign, line 132: errCh imports new code, how about just using a channel, like:
```go
	errCh := make(chan error, 1)
	...
	errCh <- err
	...
	if len(errCh) > 0 {
		xxx
	}
```

pkg/scheduler/framework/runtime/framework.go
```diff
@@ -102,38 +115,70 @@ func (frw *frameworkImpl) RunScorePlugins(ctx context.Context, placement *policy
...
+	}
 
-		weight, ok := frw.scorePluginsWeightMap[p.Name()]
-		if !ok {
-			pluginToClusterScores[p.Name()] = scoreList
-			continue
+	// Run NormalizeScore method for each ScorePlugin in parallel.
+	frw.Parallelizer().Until(ctx, len(frw.scorePlugins), func(index int) {
+		p := frw.scorePlugins[index]
+		clusterScoreList := pluginToClusterScores[p.Name()]
```
jwcesign, line 148: do we need to check whether the key exists?

pkg/scheduler/framework/runtime/framework.go
```diff
@@ -28,20 +29,31 @@ type frameworkImpl struct {
...
 type frameworkOptions struct {
 	metricsRecorder *metricsRecorder
+	parallelizer    parallelize.Parallelizer
 }
 
 // Option for the frameworkImpl.
 type Option func(*frameworkOptions)
 
+// WithParallelism sets parallelism for the scheduling frameworkImpl.
+func WithParallelism(parallelism int) Option {
```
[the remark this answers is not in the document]
Poor12: I think parallelesium is a configurable parameter. So I set aside this scaffolding method.

pkg/scheduler/framework/runtime/framework.go
Poor12: In fact, Kubernetes is also implemented in this way.

Garrybest (approved):
Cool, almost identical as kubernetes codes.

/lgtm

Issue #3024, opened 2023-01-06
[umbrella]Support multiple schedulers


Poor12:
Karmada ships with a default scheduler. If the default scheduler does not suit users' needs, it's recommended to implement their own scheduler. Now pp has `schedulerName` field but Karmada has not fully adopted it.
```
	// SchedulerName represents which scheduler to proceed the scheduling.
	// If specified, the policy will be dispatched by specified scheduler.
	// If not specified, the policy will be dispatched by default scheduler.
	// +optional
	SchedulerName string `json:"schedulerName,omitempty"`
```

- [x] set default scheduler name for propagationPolicy. linked issue: https://github.com/karmada-io/karmada/issues/279 #3026 
- [x] add `schedulerName` for ResourceBinding #3026 
- [x] add `schedulerName` option for karmada-scheduler, karmada-scheduler can deal with the scheduling task by the `scheduleName` #3026 
- [x] add validation for `schedulerName` option and field. #3093 
- [x] events about scheduler add `schedulerName` label, docs update
- [x] docs about customizing specific scheduler https://github.com/karmada-io/website/pull/286

Poor12:
The overall granularity of this feature is not very large. If you are interested, you can try to study the scenario about multiple schedulers with descheduler and scheduler-estimator.

chaunceyjiang:
> Support multiple schedulers

Will we introduce configuration files like `KubeSchedulerConfiguration`?

Poor12:
It's basicly similar to `--plugins`. Just as https://github.com/karmada-io/karmada/pull/2135#issuecomment-1179646681 mentioned, it might be a transitional strategy at the stage where our component functions are still in rapid iteration.

PR #3093, opened 2023-02-01
Add schedulerName validation

Poor12:

**What type of PR is this?**
/kind feature

Add validation for `--scheduler-name` option in `karmada-scheduler` and `schedulerName` field in policy.

Part of #3024 

None

### DIFF
--- a/cmd/scheduler/app/options/validation.go
+++ b/cmd/scheduler/app/options/validation.go
@@ -27,5 +27,9 @@ func (o *Options) Validate() field.ErrorList {
 		errs = append(errs, field.Invalid(newPath.Child("SchedulerEstimatorTimeout"), o.SchedulerEstimatorTimeout, "must be greater than or equal to 0"))
 	}
 
+	if o.SchedulerName == "" {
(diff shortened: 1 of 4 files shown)


cmd/scheduler/app/options/validation.go
```diff
@@ -27,5 +27,9 @@ func (o *Options) Validate() field.ErrorList {
 		errs = append(errs, field.Invalid(newPath.Child("SchedulerEstimatorTimeout"), o.SchedulerEstimatorTimeout, "must be greater than or equal to 0"))
 	}
 
+	if o.SchedulerName == "" {
```
XiShanYongYe-Chang, line 30: We have set the default value, will the value be left blank?

Poor12: No, it will be the default value. This logic may avoid the case `--scheduler-name=`.

XiShanYongYe-Chang: ok, got it.

XiShanYongYe-Chang (commented):
/lgtm

RainbowMango (approved):
/approve

PR #3094, opened 2023-02-01
Add schedulerName for events and metrics

Poor12:

**What type of PR is this?**
/kind feature

Add schedulerName for events and metrics

Part of #3024 

None

### DIFF
--- a/pkg/scheduler/core/generic_scheduler.go
+++ b/pkg/scheduler/core/generic_scheduler.go
@@ -41,8 +41,9 @@ type genericScheduler struct {
 func NewGenericScheduler(
 	schedCache cache.Cache,
 	registry runtime.Registry,
+	schedulerName string,
 ) (ScheduleAlgorithm, error) {
-	f, err := runtime.NewFramework(registry)
+	f, err := runtime.NewFramework(registry, schedulerName)
 	if err != nil {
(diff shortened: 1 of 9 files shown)


RainbowMango:
Please think about do we really need this.

Poor12:
I think I got it. Whether to add profile to metrics depends on whether the metrics differs in different schedulers. For example, 
`e2e_scheduling_duration_seconds` is the whole scheduling time. It differs between a default scheduler and a custom scheduler with 100+ plugins. So we need `profile` label to distinguish them. Otherwise when we aggregate metrics from different schedulers, the metrics data is not trustworthy

But `plugin_execution_duration_seconds` is the plugin execution time. It's the same for different schedulers. Similarly, metrics such as k8s workqueue are the same for different components, so there is no need to add a profile.

RainbowMango:
> Otherwise when we aggregate metrics from different schedulers, the metrics data is not trustworthy.

When aggregating metrics from multi-components by tools like Prometheus, Prometheus can help to add a label to identify the source of the metrics. 

Please provide more evidence of why we need it or postpone it for now.

Poor12:
Sorry, I forgot one thing. Event is needed indicating which scheduler is scheduling the workload.
/reopen

PR #3095, opened 2023-02-01
Fix fail of ResourceModeling when init informer failed

jwcesign:

**What type of PR is this?**
/kind bug

With the original logic, if initializeGenericInformerManagerForCluster/ClusterDynamicClientSetFunc fails, it returns nil.
But setCurrentClusterStatus will use the informer(when ResourceModeling is enabled), so we should return fail and retry it.

Fixes #none

none

```release-note

```

### DIFF
--- a/pkg/controllers/status/cluster_status_controller.go
+++ b/pkg/controllers/status/cluster_status_controller.go
@@ -194,7 +194,9 @@ func (c *ClusterStatusController) syncClusterStatus(cluster *clusterv1alpha1.Clu
 		// 	We should move this logic to both `execution-controller` and `work-status-controller`.
 		//  After that the 'initializeGenericInformerManagerForCluster' function as well as 'c.GenericInformerManager'
 		//  can be safely removed from current controller.
-		c.initializeGenericInformerManagerForCluster(clusterClient)
+		if err := c.initializeGenericInformerManagerForCluster(clusterClient); err != nil {
+			return controllerruntime.Result{Requeue: true}, err
+		}
 
 		err := c.setCurrentClusterStatus(clusterClient, cluster, &currentClusterStatus)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


jwcesign:
cc @XiShanYongYe-Chang

RainbowMango:
Is this patch for bugfix or cleanup?

chaunceyjiang:
How to reproduce?

jwcesign:
> How to reproduce?

code review

RainbowMango:
What's the impact of the problem? Please try to give more info.

pkg/controllers/execution/execution_controller.go
```diff
@@ -206,7 +206,7 @@ func (c *Controller) syncToClusters(clusterName string, work *workv1alpha1.Work)
 			}
 		} else {
 			err = c.tryCreateWorkload(clusterName, workload)
-			if err != nil {
+			if err != nil && !apierrors.IsAlreadyExists(err) {
```
jwcesign, line 209: at line pkg/controllers/execution/execution_controller.go:189, if there are two Manifests(A and B).
1. creating the work for the first time, create A succeed, B failed, so the applied will be false.
2. create the work for the second time, create A failed(already exists), and create B succeed.
3. Then create the work continually, A will be failed always(always an error of already exists).

@RainbowMango

pkg/controllers/status/cluster_status_controller.go
```diff
@@ -194,7 +194,9 @@ func (c *ClusterStatusController) syncClusterStatus(cluster *clusterv1alpha1.Clu
 		// 	We should move this logic to both `execution-controller` and `work-status-controller`.
 		//  After that the 'initializeGenericInformerManagerForCluster' function as well as 'c.GenericInformerManager'
 		//  can be safely removed from current controller.
-		c.initializeGenericInformerManagerForCluster(clusterClient)
+		if err := c.initializeGenericInformerManagerForCluster(clusterClient); err != nil {
```
jwcesign, line 197: initializeGenericInformerManagerForCluster could return an error, should requeue it.

pkg/controllers/status/cluster_status_controller.go
```diff
@@ -288,17 +290,18 @@ func (c *ClusterStatusController) updateStatusIfNeeded(cluster *clusterv1alpha1.
...
-		return
+		return nil
 	}
 
 	dynamicClient, err := c.ClusterDynamicClientSetFunc(clusterClient.ClusterName, c.Client)
 	if err != nil {
-		klog.Errorf("Failed to build dynamic cluster client for cluster %s.", clusterClient.ClusterName)
-		return
+		klog.Errorf("Failed to build dynamic cluster client for cluster %s: %v", clusterClient.ClusterName, err)
+		return err
```
jwcesign, line 301: ClusterDynamicClientSetFunc could return an error, we shouldn't ignore it, as it will cause the resource module to not work(informer not created)

jwcesign:
after discussion, there is no problem because the ResourceModeling doesn't use the generic informer.
/close

PR #3101, opened 2023-02-02
Fix multi-manifest creation keeps failing

jwcesign:

**What type of PR is this?**
/kind bug

at line pkg/controllers/execution/execution_controller.go:189, if there are two Manifests(A and B).

1. creating the work for the first time, create A succeed, B failed, so the applied will be false.
2. create the work for the second time, create A failed(already exists), and create B succeed.
3. Then create the work continually, A will be failed always(always an error of already exists).

Fixes #none

```release-note
```

### DIFF
--- a/pkg/controllers/execution/execution_controller.go
+++ b/pkg/controllers/execution/execution_controller.go
@@ -206,7 +206,7 @@ func (c *Controller) syncToClusters(clusterName string, work *workv1alpha1.Work)
 			}
 		} else {
 			err = c.tryCreateWorkload(clusterName, workload)
-			if err != nil {
+			if err != nil && !apierrors.IsAlreadyExists(err) {
 				klog.Errorf("Failed to create resource(%v/%v) in the given member cluster %s, err is %v", workload.GetNamespace(), workload.GetName(), clusterName, err)
 				c.eventf(workload, corev1.EventTypeWarning, events.EventReasonSyncWorkloadFailed, "Failed to create resource(%s) in member cluster(%s): %v", klog.KObj(workload), clusterName, err)
 				errs = append(errs, err)


jwcesign:
cc @XiShanYongYe-Chang

XiShanYongYe-Chang (commented):
/lgtm
According to the current logic, there are no more than two manifest files in the same `Work`. Therefore, I think this patch does not need to be synchronized to the previous release.

RainbowMango:
Does the `already exist error` comes out from here?:
https://github.com/karmada-io/karmada/blob/8d924efc23a8efb7f9977535d28328e440a6285a/pkg/util/objectwatcher/objectwatcher.go#L85-L89

jwcesign:
> Does the `already exist error` comes out from here?:

From here: https://github.com/karmada-io/karmada/blob/8d924efc23a8efb7f9977535d28328e440a6285a/pkg/controllers/execution/execution_controller.go#L270

jwcesign:
> Does the already exist error comes out from here?:

Sorry, it's here. 
### Step 1
https://github.com/karmada-io/karmada/blob/8a694133c46b65a5b4a4478474f873da59bee225/pkg/util/objectwatcher/objectwatcher.go#L85
### Step 2
https://github.com/karmada-io/karmada/blob/8a694133c46b65a5b4a4478474f873da59bee225/vendor/k8s.io/client-go/dynamic/simple.go#L112
### Step 3
https://github.com/karmada-io/karmada/blob/8a694133c46b65a5b4a4478474f873da59bee225/vendor/k8s.io/client-go/rest/request.go#L1020
### Step 4
https://github.com/karmada-io/karmada/blob/8a694133c46b65a5b4a4478474f873da59bee225/vendor/k8s.io/client-go/rest/request.go#L1054
### Step 5
https://github.com/karmada-io/karmada/blob/8a694133c46b65a5b4a4478474f873da59bee225/vendor/k8s.io/client-go/rest/request.go#L1104
### Step 6 - final
https://github.com/karmada-io/karmada/blob/8a694133c46b65a5b4a4478474f873da59bee225/vendor/k8s.io/apimachinery/pkg/api/errors/errors.go#L435

jwcesign:
> Does the already exist error comes out from here?

From the logic, if it already exists, there will be no error return, so this PR may haven't got any benefits.
https://github.com/karmada-io/karmada/blob/8d924efc23a8efb7f9977535d28328e440a6285a/pkg/util/objectwatcher/objectwatcher.go#L79  

But from the caller function, it's reasonable to check where it's an already-exits error(because it calls a very low-level function)

XiShanYongYe-Chang:
Can we close this update after the related refactoring?

PR #2702, opened 2022-10-29
Add placement for bindingSpec

Poor12:
Signed-off-by: Poor12 <shentiecheng@huawei.com>

**What type of PR is this?**
/kind api-change

**What this PR does / why we need it**:
Add placement for BindingSpec can reduce the dependency of policy and make API more clear. Now the component above will check the placement of bindingSpec first and switch to get placement by label if placement of bindSpec is empty.

It can be smoothly upgraded from a lower version to a higher version. If karmada-controller-manager upgrades, spec.placement will occur according to the placement of policy.

Upgrade path: upgrade crd --> upgrade karmada-controller-manager --> upgrade karmada-scheduler.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:
None

PR #3114, opened 2023-02-06
Detector watch pp fields changed other than resource selector

Poor12:

**What type of PR is this?**
/kind feature

Now detector will reconcile only when pp's resourceSelector changes. So when other fields updates, resourceBindings will not synchronize this update of the policy.

For example, if you have propagated a policy(which propagateDeps's false) and deployment, now you want to update policy's propagateDeps to true. However, resourceBinding will not update this field after policy has changed.
It will also be an issue when we add `placement` to resourceBinding #2702.

The main reason is that detector will get obj from `waitingObjects` queue, then add it to the workqueue for reconciling. However, in the above case, the binding relationship of the object will not change, so it will not be added to the `waitingObjects` queue.
```
func (d *ResourceDetector) GetMatching(resourceSelectors []policyv1alpha1.ResourceSelector) []keys.ClusterWideKey {
	d.waitingLock.RLock()
	defer d.waitingLock.RUnlock()

	var matchedResult []keys.ClusterWideKey

	for waitKey := range d.waitingObjects {
		waitObj, err := d.GetUnstructuredObject(waitKey)
		if err != nil {
			// all object in waiting list should exist. Just print a log to trace.
			klog.Errorf("Failed to get object(%s), error: %v", waitKey.String(), err)
			continue
		}

		for _, rs := range resourceSelectors {
			if util.ResourceMatches(waitObj, rs) {
				matchedResult = append(matchedResult, waitKey)
				break
			}
		}
	}

	return matchedResult
}
```

Fixes #

None

### DIFF
--- a/pkg/detector/detector.go
+++ b/pkg/detector/detector.go
@@ -368,7 +368,7 @@ func (d *ResourceDetector) LookForMatchedClusterPolicy(object *unstructured.Unst
 // ApplyPolicy starts propagate the object referenced by object key according to PropagationPolicy.
 func (d *ResourceDetector) ApplyPolicy(object *unstructured.Unstructured, objectKey keys.ClusterWideKey, policy *policyv1alpha1.PropagationPolicy) (err error) {
 	start := time.Now()
-	klog.Infof("Applying policy(%s%s) for object: %s", policy.Namespace, policy.Name, objectKey)
+	klog.Infof("Applying policy(%s/%s) for object: %s", policy.Namespace, policy.Name, objectKey)
 	var operationResult controllerutil.OperationResult
 	defer func() {
 		metrics.ObserveApplyPolicyAttemptAndLatency(object, policy.ObjectMeta, err, start)
@@ -756,7 +756,7 @@ func (d *ResourceDetector) OnPropagationPolicyUpdate(oldObj, newObj interface{})
 		return
 	}
 
-	klog.V(2).Infof("Update PropagationPolicy(%s) resourceSelectors", key)
+	klog.V(2).Infof("Update PropagationPolicy(%s)", key)
 	d.policyReconcileWorker.Add(key)
 }
 
@@ -820,7 +820,7 @@ func (d *ResourceDetector) OnClusterPropagationPolicyUpdate(oldObj, newObj inter
 		return
 	}
 
-	klog.V(2).Infof("Update ClusterPropagationPolicy(%s) resourceSelectors", key)
+	klog.V(2).Infof("Update ClusterPropagationPolicy(%s)", key)
 	d.clusterPolicyReconcileWorker.Add(key)
 }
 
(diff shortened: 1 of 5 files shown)


Poor12:
UT failed due to the codecov thing.

RainbowMango:
> However, in the above case, the binding relationship of the object will not change, so it will not be added to the waitingObjects queue.

I don't get it.  Do you mean the resource now not in `waitingObjects`, so `GetMatching` would get nothing(empty list)?

Poor12:
> > However, in the above case, the binding relationship of the object will not change, so it will not be added to the waitingObjects queue.
> 
> I don't get it. Do you mean the resource now not in `waitingObjects`, so `GetMatching` would get nothing(empty list)?

Right.

pkg/detector/detector.go
```diff
@@ -974,6 +974,26 @@ func (d *ResourceDetector) HandlePropagationPolicyCreationOrUpdate(policy *polic
 		return err
 	}
 
+	resourceBindings, err := d.getMatchedResourceBindings(policy.Namespace, policy.Name)
```
XiShanYongYe-Chang, line 977: Perhaps some comments could be added to explain why this is done.

pkg/detector/detector.go
```diff
@@ -974,6 +974,26 @@ func (d *ResourceDetector) HandlePropagationPolicyCreationOrUpdate(policy *polic
...
+	if err != nil {
+		return err
+	}
+	if len(resourceBindings.Items) != 0 {
```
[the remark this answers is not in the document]
Poor12: Done.

pkg/detector/policy.go
```diff
@@ -270,3 +270,40 @@ func (d *ResourceDetector) removeResourceLabelsIfNotMatch(objectReference workv1
...
+func (d *ResourceDetector) getMatchedResourceBindings(policyNamespace, policyName string) (*workv1alpha2.ResourceBindingList, error) {
+	var ls labels.Set
+	if len(policyNamespace) == 0 {
+		ls = labels.Set{policyv1alpha1.ClusterPropagationPolicyLabel: policyName}
+	} else {
+		ls = labels.Set{
+			policyv1alpha1.PropagationPolicyNamespaceLabel: policyNamespace,
+			policyv1alpha1.PropagationPolicyNameLabel:      policyName,
+		}
+	}
```
RainbowMango, line 283: Please think about how to test it.
I'm thinking if we can have functions like `ListPPDerivedRB` `ListCPPDerivedCRB`.

RainbowMango (commented):
Basically looks good to me. Please consider add some tests.

pkg/detector/policy.go
XiShanYongYe-Chang: +1

Poor12: Done

Poor12:
UT about the method of detector raises a PR separately.

pkg/detector/policy.go
```diff
@@ -270,3 +270,49 @@ func (d *ResourceDetector) removeResourceLabelsIfNotMatch(objectReference workv1
...
 	return true, nil
 }
+
+func (d *ResourceDetector) listPPDerivedRB(policyNamespace, policyName string) (*workv1alpha2.ResourceBindingList, error) {
+	bindings := &workv1alpha2.ResourceBindingList{}
+	listOpt := &client.ListOptions{
+		LabelSelector: labels.SelectorFromSet(labels.Set{
+			policyv1alpha1.PropagationPolicyNamespaceLabel: policyNamespace,
+			policyv1alpha1.PropagationPolicyNameLabel:      policyName,
+		})}
```
RainbowMango, line 280: Can we add the `namespace` option to list RBs from the `policyNamespace`? Like:
```diff
diff --git a/pkg/detector/policy.go b/pkg/detector/policy.go
index c07e4a4a..3e979b92 100644
--- a/pkg/detector/policy.go
+++ b/pkg/detector/policy.go
@@ -277,7 +277,9 @@ func (d *ResourceDetector) listPPDerivedRB(policyNamespace, policyName string) (
                LabelSelector: labels.SelectorFromSet(labels.Set{
                        policyv1alpha1.PropagationPolicyNamespaceLabel: policyNamespace,
                        policyv1alpha1.PropagationPolicyNameLabel:      policyName,
-               })}
+               }),
+               Namespace: policyNamespace,
+       }
        err := d.Client.List(context.TODO(), bindings, listOpt)
        if err != nil {
                klog.Errorf("Failed to list ResourceBinding with policy(%s/%s), error: %v", policyNamespace, policyName, err)
```

test/e2e/clusterpropagationpolicy_test.go
```diff
@@ -321,6 +324,143 @@ var _ = ginkgo.Describe("[AdvancedClusterPropagation] propagation testing", func
...
+				gomega.Eventually(func() bool {
+					bindings, err := karmadaClient.WorkV1alpha2().ResourceBindings(testNamespace).List(context.TODO(), metav1.ListOptions{
+						LabelSelector: labels.SelectorFromSet(labels.Set{
+							policyv1alpha1.ClusterPropagationPolicyLabel: policy.Name,
+						}).String(),
+					})
+					if err != nil {
+						return false
+					}
+					return bindings.Items[0].Spec.PropagateDeps == true
```
XiShanYongYe-Chang, line 392: The length of bindings items maybe 0, same as other places.

RainbowMango (commented):
/approve

Leave the E2E part to @XiShanYongYe-Chang

XiShanYongYe-Chang (commented):
/lgtm

Issue #780, opened 2021-09-30
Discussion about adding propagate priority during scheduling


Garrybest:
**What would you like to be added**:

Add a new field `PropagatePriority` into `ClusterAffinity` so that user could have a propagating preference with multiple levels of priority. 

**User Story**:

1. Story 1

Image that we have our own private cloud clusters which are managed by karmada. When the load of clusters reaches the peak, we must scale our workload to public cloud clusters because the resources in our private cloud is limited. 

However, the cost in public clusters are much larger than private clusters. We would like to propagate replicas as many as possible between our private clusters, the rest ones are going to be propagated between public clusters. 

2. https://github.com/karmada-io/karmada/issues/2085

**Why is this needed**:

It makes multi-cluster scheduling more flexible. Clusters are descending sorted by weight and scheduler will try to propagate to the preferred clusters under the premise of constraints, e.g., SpreadConstraints, AvailableReplicas, etc.

RainbowMango:
@Garrybest Is there a use case that can indicate the benefits to Karmada users?

Do you propose both `preferred` and `required` affinity? Or, just `preferred`?  (I can see both from the issue title, but just one from #842)

Garrybest:
It is proposed long time ago. I have updated this issue description, please take a look.

RainbowMango:
Thanks, I see it. it makes sense(just looked at story 1).

This feature belongs to the scope of `dividing` replicas, I'm not sure if we can put it into the `ReplicaSchedulingStrategy`. Just a second thought, I'll look at #2085.

Garrybest:
I think it's not just for dividing replicas. Let me explain this field.

When propagating, we have some `constraints`, e.g., `spreadConstraints`, `availableReplicas`, etc. What we are going to do is try to find a solution under the `constraints`. 

`PropagatePriority` aims to tell us in which set we could look for solutions. Imagine we get a priority list with set A, set B and set C by descending order. 

1. We try to find solutions in set A under the `constraints`.
2. If failed, maybe due to `spreadConstraints` or `availableReplicas`, we will try to propagate between set A+B. Note that dividing will have a preference with set A.
3. Once the `constraints` is met, we could return the solution.

jwcesign:
Have a questison, why not define the `PropagatePriority` in `ReplicaScheduling`?
If define in `ClusterAffinity`, if it's static weighte scheduling strategy, the priority will works?

Garrybest:
[...]

PR #3126, opened 2023-02-07
Add multiple scheduling group proposal

RainbowMango:
**What type of PR is this?**
/kind documentation
/kind feature

This PR provides a proposal for addressing the issue https://github.com/karmada-io/karmada/issues/780.

Fixes #

```release-note
NONE
```

### DIFF
--- a/docs/proposals/scheduling/multi-scheduling-group/README.md
+++ b/docs/proposals/scheduling/multi-scheduling-group/README.md
@@ -1,0 +1,371 @@
+---
+title: Multiple scheduling group
+
+authors:
+- "@RainbowMango"
+
+reviewers:
+- "@chaunceyjiang"
+- "@Garrybest"
+- "@lonelyCZ"
+- "@Poor12"
+- "@XiShanYongYe-Chang"
+
+approvers:
+- "@kevin-wangzefeng"
+
+creation-date: 2023-02-06
+
+---
+
+# Multiple scheduling group
+
+## Summary
+
+The current PropagationPolicy supports declaring `only one` group of clusters, that
+is the `.spec.placement.clusterAffinity`, e.g.
+```yaml
+apiVersion: policy.karmada.io/v1alpha1
+kind: PropagationPolicy
+metadata:
+  name: foo
+spec:
+  resourceSelectors:
+    - apiVersion: apps/v1
+      kind: Deployment
+      name: foo
+  placement:
+    clusterAffinity:
+      clusterNames:
+        - member1
+        - member2
+```
+The `clusterAffinity` supplies a group of candidate clusters to the `karmada-scheduler`.
+The `karmada-scheduler` makes schedule decisions among the candidate clusters
+according to relevant restrictions(like spreadConstraint, filter plugins, and so on),
+and the scheduling results are either successful or failure.
+- successful: successfully selected a group of clusters(probably a subset of candidates) for the referencing resource.
+- failure: failed to select a group of clusters that satisfied the restrictions.
+
+This proposal proposes a strategy that multiple `clusterAffinity` could be declared,
+and the `karmada-scheduler` could make the decision by evaluating each `clusterAffinity`
+in a specified order.
+
+## Motivation
+
+The cluster administrators usually classify their clusters by categories(like,
+by provider, usage, and so on) into different groups, and hoping to deploy
+workload to the preferred groups and leave a backup group in case of preferred
+group doesn't satisfy scheduling restrictions like lack of resources.
+
+The Karmada community received a lot of feedback about that from end users, like
+issue [#780](https://github.com/karmada-io/karmada/issues/780) and
+[#2085](https://github.com/karmada-io/karmada/issues/2085).
+
+### Goals
+
+- Extend the API of PropagationPolicy to hold multiple affinity groups declaration.
+- Extend the API of ResourceBinding to annotate the affinity group that the scheduler is inspecting.
+- Propose the implementation ideas for involved components, including `karmada-controller-manager`, `karmada-webhook` and `karmada-scheduler`.
+
+### Non-Goals
+
+- **The relative priority of clusters in the same group**
+
+This proposal focuses on introducing multiple affinity groups, for the relative
+priority of clusters in the same group topic is out of scope. And such requirements
+could be done by `weightPreference`(.spec.placement.replicaScheduling.weightPreference) or
+define a strategy for `karmada-scheduler` to score specified clusters.
+
+- **Scheduling re-balance**
+
+Indeed, oftentimes people want the `karmada-scheduler` to perform an extra schedule
+for different purposes, like the question addressed by the discussion
+[#3069](https://github.com/karmada-io/karmada/discussions/3069), these kind of
+issues should be tracked by another proposal.
+
+## Proposal
+
+### User Stories (Optional)
+
+#### As a user, I prefer to deploy my applications on clusters in local data center to save costs.
+
+I have some clusters in my data center as well as some managed clusters from cloud
+providers(like AWS and Google Cloud). Since the cost of the managed clusters is higher
+than private clusters, so I prefer to deploy applications on private clusters and
+take the managed clusters as a backup.
+
+#### As a user, I want to deploy applications on the primary cluster and leave a backup cluster for disaster recovery cases.
+
+I have two clusters, the primary, and the backup cluster, I want Karmada to deploy
+my applications on the primary cluster and migrate applications on the backup cluster
+when the primary cluster becomes unavailable due to like data center power off or
+maintenance task.
+
+### Notes/Constraints/Caveats (Optional)
+
+This proposal mainly focuses on the changes in `PropagationPolicy` and `ResourceBinding`,
+but it will also be applied to `ClusterPropagationPolicy` and `ClusterResourceBinding`.
+
+### Risks and Mitigations
+
+This proposal maintains the backward compatibility, the system built with previous
+versions of Karmada can be seamlessly migrated to the new version.
+The previous configurations(yamls) could be applied to the new version of Karmada and
+without any behavior change.
+
+## Design Details
+
+### API change
+
+#### PropagationPolicy API change
+
+This proposal proposes a new field `ClusterAffinities` for declaring
+multiple affinity terms in `.spec.placement` of `PropagationPolicy`.
+```go
+// Placement represents the rule for select clusters.
+type Placement struct {
+	// ClusterAffinity represents scheduling restrictions to a certain set of clusters.
+	// If not set, any cluster can be scheduling candidate.
+	// +optional
+	ClusterAffinity *ClusterAffinity `json:"clusterAffinity,omitempty"`
+
+    // ClusterAffinities represents scheduling restrictions to multiple cluster
+    // groups that indicated by ClusterAffinityTerm.
+    //
+    // The scheduler will evaluate these groups one by one in the order they
+    // appear in the spec, the group that does not satisfy scheduling restrictions
+    // will be ignored which means all clusters in this group will not be selected
+    // unless it also belongs to the next group(a cluster could belong to multiple
+    // groups).
+    //
+    // If none of the groups satisfy the scheduling restrictions, then scheduling
+    // fails, which means no cluster will be selected.
+    //
+    // Note:
+    //   1. ClusterAffinities can not co-exist with ClusterAffinity.
+    //   2. If both ClusterAffinity and ClusterAffinities are not set, any cluster
+    //      can be scheduling candidates.
+    //
+    // Potential use case 1:
+    // The private clusters in the local data center could be the main group, and
+    // the managed clusters provided by cluster providers could be the secondary
+    // group. So that the Karmada scheduler would prefer to schedule workloads
+    // to the main group and the second group will only be considered in case of
+    // the main group does not satisfy restrictions(like, lack of resources).
+    //
+    // Potential use case 2:
+    // For the disaster recovery scenario, the clusters could be organized to
+    // primary and backup groups, the workloads would be scheduled to primary
+    // clusters firstly, and when primary cluster fails(like data center power off),
+    // Karmada scheduler could migrate workloads to the backup clusters.
+    //
+    // +optional
+    ClusterAffinities []ClusterAffinityTerm `json:"clusterAffinities,omitempty"`
+
+	// ClusterTolerations represents the tolerations.
+	// +optional
+	ClusterTolerations []corev1.Toleration `json:"clusterTolerations,omitempty"`
+
+	// SpreadConstraints represents a list of the scheduling constraints.
+	// +optional
+	SpreadConstraints []SpreadConstraint `json:"spreadConstraints,omitempty"`
+
+	// ReplicaScheduling represents the scheduling policy on dealing with the number of replicas
+	// when propagating resources that have replicas in spec (e.g. deployments, statefulsets) to member clusters.
+	// +optional
+	ReplicaScheduling *ReplicaSchedulingStrategy `json:"replicaScheduling,omitempty"`
+}
+
+// ClusterAffinityTerm selects a set of cluster.
+type ClusterAffinityTerm struct {
+	// AffinityName is the name of the cluster group.
+	// +required
+	AffinityName string `json:"affinityName"`
+
+	ClusterAffinity `json:",inline"`
+}
+```
+Each affinity term essentially is the named `ClusterAffinity`. During the scheduling
+phase, the scheduler evaluates terms in the order they appear in the spec one by
+one, and turns to the next term if the term does not satisfy restrictions.
+
+The following configuration sample declares 3 affinity terms:
+```yaml
+apiVersion: policy.karmada.io/v1alpha1
+kind: PropagationPolicy
+metadata:
+  name: nginx
+spec:
+  resourceSelectors:
+    - apiVersion: apps/v1
+      kind: Deployment
+      name: nginx
+  placement:
+    clusterAffinities:
+      - affinityName: dc-shanghai
+        clusterNames:
+          - unavailable
+      - affinityName: dc-beijing
+        clusterNames:
+          - member1
+      - affinityName: dc-hongkong
+        clusterNames:
+          - member2
+```
+
+During the scheduling phase, the scheduler will first look at the affinity
+named `dc-shanghai` and try to select a feasible cluster set from it.
+
+If no feasible cluster is found from the term, the scheduler will try against the
+next term which is the one named `dc-beijing`. And once the scheduler successfully
+selected a feasible cluster set from this affinity term, the scheduler will not
+continue to look at the following terms.
+
+And, in case of cluster `member1` becomes unavailable, by leveraging the
+[Failover feature](https://karmada.io/docs/userguide/failover/failover-overview)
+the scheduler will start looking for alternatives `in current affinity term`, and
+if that fails, it will look at the term named `dc-hongkong`.
+
+**Note:** Each affinity term is completely independent, and the scheduler will
+only select one term at a time during scheduling. But it is allowed for the same
+cluster present in multiple affinity terms.
+
+#### ResourceBinding API change
+
+In order to track the affinity term that the scheduler last evaluated,
+the proposal proposes a new field named `SchedulerObservedAffinityName` to
+`.status` of `ResourceBinding`, so that the scheduler can continue the previous
+scheduling cycle in any cases of re-schedule.
+
+```go
+// ResourceBindingStatus represents the overall status of the strategy as well as the referenced resources.
+type ResourceBindingStatus struct {
+	// SchedulerObservedGeneration is the generation(.metadata.generation) observed by the scheduler.
+	// If SchedulerObservedGeneration is less than the generation in metadata means the scheduler hasn't confirmed
+	// the scheduling result or hasn't done the schedule yet.
+	// +optional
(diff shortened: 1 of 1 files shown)


RainbowMango:
Any comments are welcome, just cc guys who might be interested in the feature.
@kevin-wangzefeng @Garrybest @lonelyCZ @gf457832386 @jwcesign @Poor12 @XiShanYongYe-Chang 

I'll add the reviewer and approver and put names on the head of the proposal once it is accepted.

docs/proposals/scheduling/multi-scheduling-group/README.md
```diff
@@ -0,0 +1,366 @@
...
+        - member2
+```
+The `clusterAffinity` supplies a group of candidate clusters to the `karmada-scheduler`.
+The `karmada-scheduler` makes schedule decisions among the candidate clusters
+according to relevant restrictions(like spreadConstraint, filter plugins, and so on),
+and the scheduling results are either successful or failure.
+- successful: successfully selected a group of clusters(probably a subset of candidates) for the referencing resource.
+- failure: failed to select a group of clusters that satisfied the restrictions.
+
+This proposal proposes a strategy that multiple `clusterAffinity` that could be declared,
```
lonelyCZ, line 43:
```suggestion
This proposal proposes a strategy that multiple `clusterAffinity` could be declared,
```

docs/proposals/scheduling/multi-scheduling-group/README.md
```diff
@@ -0,0 +1,366 @@
...
+group doesn't satisfy scheduling restrictions like lack of resources.
+
+The Karmada community received a lot of feedback about that from end users, like
+issue [#780](https://github.com/karmada-io/karmada/issues/780) and
+[#2085](https://github.com/karmada-io/karmada/issues/780).
+
+### Goals
+
+- Define the API of PropagationPolicy to hold multiple affinity groups declaration.
+- Define the API of ResourceBinding to annotate the affinity group that the scheduler inspecting on.
```
lonelyCZ, line 61:
```suggestion
- Define the API of ResourceBinding to annotate the affinity group that the scheduler is inspecting on.
```

docs/proposals/scheduling/multi-scheduling-group/README.md
```diff
@@ -0,0 +1,366 @@
...
+workload to the preferred groups and leave a backup group in case of preferred
+group doesn't satisfy scheduling restrictions like lack of resources.
+
+The Karmada community received a lot of feedback about that from end users, like
+issue [#780](https://github.com/karmada-io/karmada/issues/780) and
+[#2085](https://github.com/karmada-io/karmada/issues/780).
+
+### Goals
+
+- Define the API of PropagationPolicy to hold multiple affinity groups declaration.
```
lonelyCZ, line 60:
```suggestion
- Extend the API of PropagationPolicy to hold multiple affinity groups declaration.
```

How aboud using "Extend"?

lonelyCZ, line 60:
```suggestion
- Extend the API of PropagationPolicy to hold multiple affinity groups declaration.
```

How aboud using "Extend"?

docs/proposals/scheduling/multi-scheduling-group/README.md
```diff
@@ -0,0 +1,366 @@
...
+## Proposal
+
+### User Stories (Optional)
+
+#### As a user, I prefer to deploy my applications on clusters in my data center to save costs.
+
+I have some clusters in my data center as well as some managed clusters from cloud
+providers(like AWS and Google Cloud). Since the cost of the managed clusters is higher
+than private clusters, so I prefer to deploy applications on private clusters and
+take managed clusters as a backup.
```
lonelyCZ, line 89:
```suggestion
take the managed clusters as a backup.
```

docs/proposals/scheduling/multi-scheduling-group/README.md
```diff
@@ -0,0 +1,366 @@
...
+Indeed, oftentimes people want the `karmada-scheduler` to perform an extra schedule
+for different purposes, like the question addressed by the discussion
+[#3069](https://github.com/karmada-io/karmada/discussions/3069), these kind of
+issues should be tracked by another proposal.
+
+## Proposal
+
+### User Stories (Optional)
+
+#### As a user, I prefer to deploy my applications on clusters in my data center to save costs.
```
lonelyCZ, line 84:
```suggestion
#### As a user, I prefer to deploy my applications on clusters in local data center to save costs.
```

docs/proposals/scheduling/multi-scheduling-group/README.md
RainbowMango: done.

docs/proposals/scheduling/multi-scheduling-group/README.md
```diff
@@ -0,0 +1,366 @@
...
+## Motivation
+
+The cluster administrators usually classify their clusters by categories(like,
+by provider, usage, and so on) into different groups, and hoping to deploy
+workload to the preferred groups and leave a backup group in case of preferred
+group doesn't satisfy scheduling restrictions like lack of resources.
+
+The Karmada community received a lot of feedback about that from end users, like
+issue [#780](https://github.com/karmada-io/karmada/issues/780) and
+[#2085](https://github.com/karmada-io/karmada/issues/780).
```
Poor12, line 56: Link error.

docs/proposals/scheduling/multi-scheduling-group/README.md
```diff
@@ -0,0 +1,366 @@
...
+
+## Design Details
+
+### API change
+
+#### PropagationPolicy API change
+
+This proposal proposes a new field `OrderedClusterAffinities` for declaring
+multiple affinity terms in `.spec.placement` of `PropagationPolicy`.
+```yaml
```
Garrybest, line 118: `yaml` -> `go`

docs/proposals/scheduling/multi-scheduling-group/README.md
```diff
@@ -0,0 +1,366 @@
...
+      - affinityName: dc-hongkong
+        order: 3
+        clusterNames:
+          - member2
+```
+
+During the scheduling phase, the scheduler will first look at the affinity
+term with `order 1` and try to select a feasible cluster set from it.
+
+If no feasible cluster is found from the term, the scheduler will try against the
```
Garrybest, line 217: It seems like all affinity terms are independent. How could we cope with the case below?

Goal: I would like to deploy 10 replicas among `dc-shanghai` as many as possible, `dc-beijing` is my second choice.
Situation: Only 8 replicas could `dc-shanghai` produce, the other 2 replicas must be deployed among `dc-beijing` which has sufficient resources.

If the affinities are independent, it seems like `dc-shanghai` does not fit so that we could only choose `dc-beijing` to produce all 10 replicas which may not meet the initial requirement.

docs/proposals/scheduling/multi-scheduling-group/README.md
```diff
@@ -0,0 +1,366 @@
...
+When creating or updating `ResourceBidning`/`ClusterResourceBinding`, the added
+`OrderedClusterAffinities` in `PropagationPolicy`/`ClusterPropagationPolicy` should
+be synced.
+
+#### karmada-scheduler
+
+Currently, the karmada-scheduler only runs a single loop by accepting an affinity
+term, that is [ScheduleAlgorithm interface](https://github.com/karmada-io/karmada/blob/32b8f21b79017e2f4154fbe0677cb63fb18b120c/pkg/scheduler/core/generic_scheduler.go#L22).
+
+With this proposal, the `ScheduleAlgorithm interface` probably be invoked multi
```
Garrybest, line 307: +1

docs/proposals/scheduling/multi-scheduling-group/README.md
```diff
@@ -0,0 +1,366 @@
...
+if that fails, it will look at the term with `order 3`.
+
+**Note:** Each affinity term is completely independent, and the scheduler will
+only select one term at a time during scheduling. But it is allowed for the same
+cluster present in multiple affinity terms.
+
+#### ResourceBinding API change
+
+In order to track the affinity term that the scheduler last evaluated,
+the proposal proposes a new field named `SchedulerObservedAffinityIndex` to
```
Garrybest, line 234: Does any other component need this field status, or just for a record?

docs/proposals/scheduling/multi-scheduling-group/README.md
```diff
@@ -0,0 +1,366 @@
...
+`.status` of `ResourceBinding`, so that the scheduler can continue the previous
+scheduling cycle.
```
Garrybest, line 236: I'm sure about the effect of this field, could you give a case here about how scheduler leverage this field for a second scheduling?

docs/proposals/scheduling/multi-scheduling-group/README.md
RainbowMango: > So I suggest we could make a concept of hierarchical relationship, i.e., if dc-shanghai does not fit, we add dc-beijing with dc-shanghai as an input.

For this case, I think it's not a good idea to group clusters by category `data center`. We can combine two data centers into a new group, E.g.:

```yaml
  placement:
    orderedClusterAffinities:
      - affinityName: primary
        order: 1
        clusterNames:
          - cluster-in-dc-beijing
      - affinityName: backup
        order: 2
        clusterNames:
          - cluster-in-dc-beijing  # still in the candidate list
          - cluster-in-dc-shanghai
```

This proposal does not help to assign more replicas to `cluster-in-dc-beijing`, my idea is we can do that by `weightPreference` or something else. The current `ClusterAffinity` also has this problem.
Please see `Non-Goals` section, that is `The relative priority of clusters in the same group`.

docs/proposals/scheduling/multi-scheduling-group/README.md
RainbowMango: Re-schedule after an eviction or fail-over. The scheduler could continue the schedule from the last `affinity term`.

In addition, we also need to know if a re-schedule should be triggered in case of placement changes on PropagationPolicy.

docs/proposals/scheduling/multi-scheduling-group/README.md
```diff
@@ -0,0 +1,366 @@
...
+
+And, in case of cluster `member1` becomes unavailable, by leveraging the
+[Failover feature](https://karmada.io/docs/userguide/failover/failover-overview)
+the scheduler will start looking for alternatives `in current affinity term`, and
+if that fails, it will look at the term with `order 3`.
+
+**Note:** Each affinity term is completely independent, and the scheduler will
+only select one term at a time during scheduling. But it is allowed for the same
+cluster present in multiple affinity terms.
+
```
chaunceyjiang, line 230:
```yaml
kind: PropagationPolicy
metadata:
  name: nginx
spec:
  resourceSelectors:
    - apiVersion: apps/v1
      kind: Deployment
      name: nginx
  placement:
    orderedClusterAffinities:
      - affinityName: dc-shanghai
        order: 1
        clusterNames:
          - unavailable
      - affinityName: dc-beijing
        order: 2
        clusterNames:
          - member1
      - affinityName: dc-hongkong
        order: 10   # order is not sequential. 
        clusterNames:
          - member2
```

Is this usage correct?

docs/proposals/scheduling/multi-scheduling-group/README.md
```diff
@@ -0,0 +1,366 @@
...
+	// SchedulerObservedGeneration is the generation(.metadata.generation) observed by the scheduler.
+	// If SchedulerObservedGeneration is less than the generation in metadata means the scheduler hasn't confirmed
+	// the scheduling result or hasn't done the schedule yet.
+	// +optional
+	SchedulerObservedGeneration int64 `json:"schedulerObservedGeneration,omitempty"`
+
+	// SchedulerObservedAffinityName is the affinity terms that
+	// scheduler looking at.
+	// +optional
+	SchedulerObservedAffinityName int32 `json:"schedulerObservingAffinityName,omitempty"`
```
chaunceyjiang, line 250: Why use `int32` instead of `string`?

docs/proposals/scheduling/multi-scheduling-group/README.md
RainbowMango: Good questions. What's your opinion?

docs/proposals/scheduling/multi-scheduling-group/README.md
RainbowMango: Nice catch. It should be a string. 
(I just renamed the filed from SchedulerObservedAffinity`Index`).

docs/proposals/scheduling/multi-scheduling-group/README.md
```diff
@@ -0,0 +1,366 @@
...
+
+This proposal proposes a new field `OrderedClusterAffinities` for declaring
+multiple affinity terms in `.spec.placement` of `PropagationPolicy`.
+```yaml
+// Placement represents the rule for select clusters.
+type Placement struct {
+	// ClusterAffinity represents scheduling restrictions to a certain set of clusters.
+	// If not set, any cluster can be scheduling candidate.
+	// +optional
+	ClusterAffinity *ClusterAffinity `json:"clusterAffinity,omitempty"`
```
XiShanYongYe-Chang, line 124: The newly added `OrderedClusterAffinities` field has overwritten the function of this field(When the length of the new field is 1), will this field be deprecated?

docs/proposals/scheduling/multi-scheduling-group/README.md
```diff
@@ -0,0 +1,366 @@
...
+	// when propagating resources that have replicas in spec (e.g. deployments, statefulsets) to member clusters.
+	// +optional
+	ReplicaScheduling *ReplicaSchedulingStrategy `json:"replicaScheduling,omitempty"`
+}
+// OrderedClusterAffinityTerm selects a set of clusters and indicates the order
+// that the Karmada scheduler should inspect.
+type OrderedClusterAffinityTerm struct {
+	ClusterAffinity `json:",inline"`
+
+	// Order of the cluster group that the Karmada scheduler should inspect with, in the range 1-100.
```
XiShanYongYe-Chang, line 174: A higher value indicates a higher priority or a lower priority? Shouldn't this be explained here?

And can users set the same value for different terms?

docs/proposals/scheduling/multi-scheduling-group/README.md
RainbowMango: done

docs/proposals/scheduling/multi-scheduling-group/README.md
RainbowMango: done.

docs/proposals/scheduling/multi-scheduling-group/README.md
RainbowMango: done

docs/proposals/scheduling/multi-scheduling-group/README.md
XiShanYongYe-Chang: Does this mean that we need to distinguish the scenarios that trigger rescheduling?

docs/proposals/scheduling/multi-scheduling-group/README.md
```diff
@@ -0,0 +1,366 @@
...
+cluster present in multiple affinity terms.
+
+#### ResourceBinding API change
+
+In order to track the affinity term that the scheduler last evaluated,
+the proposal proposes a new field named `SchedulerObservedAffinityIndex` to
+`.status` of `ResourceBinding`, so that the scheduler can continue the previous
+scheduling cycle.
+
+```yaml
```
chaunceyjiang, line 238: `yaml` ->`go`

docs/proposals/scheduling/multi-scheduling-group/README.md
chaunceyjiang: > In addition, we also need to know if a re-schedule should be triggered in case of placement changes on PropagationPolicy.

If the user deletes `ds-hongkong` in `orderedClusterAffinities`, will re-schedule?

docs/proposals/scheduling/multi-scheduling-group/README.md
```diff
@@ -0,0 +1,366 @@
...
+if that fails, it will look at the term with `order 3`.
+
+**Note:** Each affinity term is completely independent, and the scheduler will
+only select one term at a time during scheduling. But it is allowed for the same
+cluster present in multiple affinity terms.
+
+#### ResourceBinding API change
+
+In order to track the affinity term that the scheduler last evaluated,
+the proposal proposes a new field named `SchedulerObservedAffinityIndex` to
```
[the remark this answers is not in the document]
RainbowMango: Yes. It's a typo.

docs/proposals/scheduling/multi-scheduling-group/README.md
```diff
@@ -0,0 +1,370 @@
...
+	// +required
+	Order int32 `json:"order"`
+	// AffinityName is the name of the cluster group.
+	// +required
+    AffinityName string `json:"affinityName"`
+}
+```
+Each affinity term essentially is the current `ClusterAffinity` but has an
+ordinal number named `order` against which the scheduler evaluates each term individually.
+That means the scheduler always evaluates the first term(with `order: 1`), and turns
```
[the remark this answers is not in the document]
Garrybest: Actually, order could be removed if we order these affinity items by indexes in slice🤔.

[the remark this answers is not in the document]
chaunceyjiang: > Actually, order could be removed if we order these affinity items by indexes in slice🤔.

Perhaps the index of affinity displayed would be more understandable to the user?

docs/proposals/scheduling/multi-scheduling-group/README.md
```diff
@@ -0,0 +1,370 @@
...
+
+If no feasible cluster is found from the term, the scheduler will try against the
+next term which is the one with `order 2`. And once the scheduler successfully
+selected a feasible cluster set from this affinity term, the scheduler will not
+continue to look at the following terms.
+
+And, in case of cluster `member1` becomes unavailable, by leveraging the
+[Failover feature](https://karmada.io/docs/userguide/failover/failover-overview)
+the scheduler will start looking for alternatives `in current affinity term`, and
+if that fails, it will look at the term with `order 3`.
```
XiShanYongYe-Chang, line 225: In this case, terms will always increase. When will they start with the lowest-priority term?

docs/proposals/scheduling/multi-scheduling-group/README.md
RainbowMango: > If the user deletes ds-hongkong in orderedClusterAffinities, will re-schedule?

That depends on if the `ds-hongkong` is in use, that is I mentioned on https://github.com/karmada-io/karmada/issues/3153#issuecomment-1433007295.

docs/proposals/scheduling/multi-scheduling-group/README.md
RainbowMango: I think we can tolerate this case, that is non-sequential order is allowed.
I also mentioned this case in the proposal.

docs/proposals/scheduling/multi-scheduling-group/README.md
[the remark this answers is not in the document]
RainbowMango: I'm not sure the index of slice in yaml is constant after it is stored in ETCD. Are you sure about it?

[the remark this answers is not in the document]
Garrybest: > I'm not sure the index of slice in yaml is constant after it is stored in ETCD. Are you sure about it?

I think so, slice does have a order. Take init containers as an example, kubelet ensures that all init containers start by order, the latter one must not start until the former one finishes.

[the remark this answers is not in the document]
RainbowMango: Yeah, thanks for reminding me. I get it from https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#detailed-behavior.
> Then the kubelet runs the Pod's init containers in the order they appear in the Pod's spec.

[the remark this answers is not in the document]
RainbowMango: In that case, we don't need this order either. I'll send a PR to remove the order. Meanwell, the `OrderedClusterAffinities` could be shortened to `Affinities`, how do you think?

XiShanYongYe-Chang:
I think it is necessary to explain where both `ClusterAffinify` and `OrderedClusterAffinities` are nil. In this case, `ClusterAffinify` is nill indicates that all clusters can be selected, or `OrderedClusterAffinities` is nil indicates that no cluster group can be selected.

docs/proposals/scheduling/multi-scheduling-group/README.md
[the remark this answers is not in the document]
Garrybest: > Meanwell, the OrderedClusterAffinities could be shortened to Affinities, how do you think?

+1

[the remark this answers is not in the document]
RainbowMango: Please see https://github.com/karmada-io/karmada/pull/3172 for the updates.

docs/proposals/scheduling/multi-scheduling-group/README.md
RainbowMango: I'll update the proposal according to latest APIs in #3172.

RainbowMango:
Updated the proposal to address the comments above.
Please let me know If I missed any.

PR #3132, opened 2023-02-09
Modify unit tests that never fail

Fish-pro:

**What type of PR is this?**

/kind cleanup

Change will never happen wrong unit tests, from the function implementation, the need to focus only on `corev1. TaintEffectNoExecute` types of `timeAdd` isn't empty
```golang
// MutateClusterTaints add TimeAdded field for cluster NoExecute taints only if TimeAdded not set.
func MutateClusterTaints(taints []corev1.Taint) {
	for i := range taints {
		if taints[i].Effect == corev1.TaintEffectNoExecute && taints[i].TimeAdded == nil {
			now := metav1.Now()
			taints[i].TimeAdded = &now
		}
	}
}

```

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/registry/cluster/strategy_test.go
+++ b/pkg/registry/cluster/strategy_test.go
@@ -277,23 +277,30 @@ func TestStrategy_AllowUnconditionalUpdate(t *testing.T) {
 
 func TestStrategy_Canonicalize(t *testing.T) {
 	clusterStrategy := NewStrategy(clusterscheme.Scheme)
-	clusterBefore := getValidCluster("cluster")
-	clusterAfter := getValidCluster("cluster")
-	clusterAfter.Spec.Taints = []corev1.Taint{
+	cluster := getValidCluster("cluster")
+	cluster.Spec.Taints = []corev1.Taint{
 		{
 			Key:    "foo",
-			Value:  "abc",
+			Value:  "bar",
 			Effect: corev1.TaintEffectNoSchedule,
 		},
 		{
-			Key:    "bar",
+			Key:    "foo1",
+			Value:  "bar1",
 			Effect: corev1.TaintEffectNoExecute,
 		},
 	}
 
-	clusterStrategy.Canonicalize(clusterBefore)
-	if !reflect.DeepEqual(clusterAfter, clusterAfter) {
-		t.Errorf("Object mismatch! Excepted: \n%#v \ngot: \n%#v", clusterAfter, clusterAfter)
+	clusterStrategy.Canonicalize(cluster)
+	success := true
+	for _, taint := range cluster.Spec.Taints {
+		if taint.Effect == corev1.TaintEffectNoExecute && taint.TimeAdded == nil {
+			success = false
+			break
+		}
+	}
+	if !success {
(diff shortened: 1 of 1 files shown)


pkg/registry/cluster/strategy_test.go
```diff
@@ -277,23 +277,30 @@ func TestStrategy_AllowUnconditionalUpdate(t *testing.T) {
...
 		{
-			Key:    "bar",
+			Key:    "foo1",
+			Value:  "bar1",
 			Effect: corev1.TaintEffectNoExecute,
 		},
 	}
 
-	clusterStrategy.Canonicalize(clusterBefore)
-	if !reflect.DeepEqual(clusterAfter, clusterAfter) {
```
Fish-pro, line 295: The original comparison is meaningless

pkg/registry/cluster/strategy_test.go
```diff
@@ -277,23 +277,30 @@ func TestStrategy_AllowUnconditionalUpdate(t *testing.T) {
...
-		t.Errorf("Object mismatch! Excepted: \n%#v \ngot: \n%#v", clusterAfter, clusterAfter)
+	clusterStrategy.Canonicalize(cluster)
+	success := false
+	for _, taint := range cluster.Spec.Taints {
+		if taint.Key == "foo1" && taint.TimeAdded != nil {
```
XiShanYongYe-Chang, line 297: Is it better to judge the `Effect` value?

XiShanYongYe-Chang (commented):
Thanks for the correction. I wrote it wrong.

pkg/registry/cluster/strategy_test.go
Fish-pro:
```go
// MutateClusterTaints add TimeAdded field for cluster NoExecute taints only if TimeAdded not set.
func MutateClusterTaints(taints []corev1.Taint) {
	for i := range taints {
		if taints[i].Effect == corev1.TaintEffectNoExecute && taints[i].TimeAdded == nil {
			now := metav1.Now()
			taints[i].TimeAdded = &now
		}
	}
}
```
I understand from Canonicalize implementation, it adds `timeAdd`, so I judged `timeAdd`, but we can still add a type judgment, what do you think?
```go
	success := false
	for _, taint := range cluster.Spec.Taints {
		if taint.Effect != corev1.TaintEffectNoExecute {
			continue
		}
		if taint.Key == "foo1" && taint.TimeAdded != nil {
			success = true
			break
		}
	}
```

XiShanYongYe-Chang: My thought is like this:
```go
success := true
for _, taint := range cluster.Spec.Taints {
  if taint.Effect == corev1.TaintEffectNoExecute && taint.TimeAdded == nil {
    success = false
    break
  }
}
```

Fish-pro: I got it

XiShanYongYe-Chang (commented):
/lgtm

RainbowMango (approved):
/approve

PR #3133, opened 2023-02-09
Add “Application failover mechanism“ proposal

Poor12:

**What type of PR is this?**
/kind documentation
/kind design

Add “Rescheduling mechanism for unhealthy application state“ proposal

Fixes #

None

### DIFF
--- a/docs/proposals/failover/application-failover.md
+++ b/docs/proposals/failover/application-failover.md
@@ -1,0 +1,658 @@
+---
+title: Application failover mechanism 
+authors:
+  - "@Poor12"
+reviewers:
+  - "@RainbowMango"
+  - "@Garrybest"
+  - "@chaunceyjiang"
+  - "@XiShanYongYe-Chang"
+  - "@kevin-wangzefeng"
+  - "@GitHubxsy"
+approvers:
+  - "@RainbowMango"
+  - "@kevin-wangzefeng"
+  - "@GitHubxsy"
+
+create-date: 2023-02-10
+---
+# Application failover mechanism
+
+## Summary
+
+Kubernetes has fault tolerance and self-healing capabilities, which can automatically migrate workloads to other normal nodes when nodes fail.
+When kubelet detects that the node resources are insufficient and reaches the eviction threshold, kubelet will evict the low-priority Pod from the node and reschedule it to another node with sufficient resources.
+Also, when a node is abnormal or a user shuts down a node for maintenance, Kubernetes ensures the HA operation of Pods based on the taint eviction mechanism.
+
+However, the above solutions still cannot solve some scenarios.
+When there are not enough resources in the cluster to run an application, the evicted application will not be able to be scheduled and end up in a suspended state for a long time.
+At this time, users hope to provide a multi-cluster failover mechanism to schedule applications to another cluster with sufficient resources.
+
+The same situation will occur in clusters with resource preemption. if no node is found that satisfies all the specified requirements of the Pod, preemption logic tries to find a node where removal of one or more Pods with lower priority than scheduling Pod would enable it to be scheduled on that node.
+If such a node is found, one or more lower priority Pods get evicted from the node. After the Pods are gone, the Pod with high priority can be scheduled.
+In that case, Pods with low priority may be evicted frequently and have fewer scheduling opportunities.
+The reason for preemption may be due to resource shortage, or it may be due to affinity between Pods.
+At this time, single-cluster failover cannot effectively solve the problem. Multi-cluster failover may be a reasonable way.
+
+In the multi-cluster scenario, user workloads may be deployed in multiple clusters to improve service high availability.
+Karmada already supports multi-cluster failover when detecting a cluster fault. It's a consideration from a cluster perspective.
+However, the application may still be unavailable when the control plane of the cluster is in a healthy state.
+Therefore, Karmada needs to provide a means of fault migration from an application perspective.
+
+## Motivation
+
+The cluster administrators want to ensure that the application can run normally, regardless of which cluster the application is running. When the application is unavailable and it cannot self-heal within a single cluster,
+they want Karmada to reschedule the application and avoid clusters that failed to run before.
+
+## Goals
+
+- Extend the API of PropagationPolicy to provide users with the rescheduled configuration.
+- Extend the API of ResourceBinding to record previously evicted clusters.
+- Propose the implementation ideas for involved components, including `karmada-controller-manager`, `karmada-webhook` and `karmada-scheduler`.
+  For example, `karmada-controller-manager` need to add a component to observe the health status of the application. `Karmada-scheduler` need to
+  add a plugin to filter the evicted cluster.
+
+## Non-Goals
+
+## Proposal
+
+### User Stories (Optional)
+
+#### As a user, I deploy an application in multiple clusters with preemptive scheduling. When business peaks, low-priority applications cannot run normally for a long time.
+
+Supposing I deploy an application in a cluster with preemptive scheduling. When cluster resources are in short supply, low-priority applications that were running normally are preempted and cannot run normally for a long time.
+At this time, applications cannot self-heal within a cluster. Users wants to try to schedule it to another cluster to ensure that the service is continuously served.
+
+#### As a user, I use the cloud vendor's spot instance to deploy my application. As a result, the application may fail due to insufficient resources.
+
+Spot instances are a mode of instances that feature a discounted price and a system interruption mechanism, which means that you can purchase a spot instance at a discounted price, but the system may automatically repossess it.
+When users use spot instances to deploy applications, the application may fail to run due to resources being recycled. In this scenario, the amount of resources perceived by the scheduler is the size of the resource quota, not the actual available resources.
+At this time, users want to schedule the application to a cluster other than the one that failed previously.
+
+Link:
+
+* en: [Use preemptible instances(Alibaba Cloud)](https://www.alibabacloud.com/help/en/container-service-for-kubernetes/latest/use-preemptible-instances), [Spot Mode(Tencent Cloud)](https://www.tencentcloud.com/document/product/457/46967), [Amazon EC2 Spot(AWS)](https://aws.amazon.com/ec2/spot/details/)
+* cn: [使用抢占式实例（阿里云）](https://help.aliyun.com/document_detail/165053.html), [竞价模式说明（腾讯云）](https://cloud.tencent.com/document/product/457/59364), [Amazon EC2 Spot功能（AWS）](https://aws.amazon.com/cn/ec2/spot/details/?nc1=h_ls)
+
+### Dependency
+
+Karmada's `Resource Interpreter Framework` is designed for interpreting resource structure. It provides users with a interpreter operation to tell Karmada
+how to figure out the health status of a specific object. It is up to users to decide when to reschedule.
+
+For example:
+
+```yaml
+apiVersion: config.karmada.io/v1alpha1
+kind: ResourceInterpreterCustomization
+metadata:
+  name: declarative-configuration-example
+spec:
+  target:
+    apiVersion: apps/v1
+    kind: Deployment
+  healthInterpretation:
+    luaScript: >
+      function InterpretHealth(observedObj)
+        return observedObj.status.readyReplicas == observedObj.spec.replicas
+      end
+```
+
+**Note**: In the application failover scenario, the `Health` field not only represents the health status of the application,
+but also one of the conditions to determine whether the application is migrated. Please configure this field carefully when using failover.
+
+### Risks and Mitigations
+
+This proposal maintains the backward compatibility, the system built with previous versions of Karmada can be seamlessly migrated to the new version.
+The previous configurations(yamls) could be applied to the new version of Karmada and without any behavior change.
+
+## Design Details
+
+Users are able to define the behavior of failover by defining policies.
+The configuration of failover behavior includes:
+
+* When to trigger application failover
+* How to migrate from one cluster to another
+* Time at which the migrated cluster can be recovered
+
+Sometimes, you have to deal with legacy applications that might require an additional startup time on their first initialization. Or you want to
+avoid invalid failovers due to misconfigurations. In such cases, we need some preconditions to start the failover behavior.
+Optional configuration items include:
+
+* Healthy state. We just want to migrate those applications that have run successfully. For those applications that failed at the beginning, we prefer to solve them in the single cluster.
+* A period of time. We can wait a period of time before the application failover is initiated.
+
+Also, there are some decision conditions of performing the failover process.
+
+* Healthy state. We can define in what state the application triggers the migration. Valid options are "Unhealthy", "Unknown" or both.
+* A period of time. For some short-lived exceptions(which can be recovered after a period of time), we don't want to immediately migrate the application to other clusters which may make the migration very frequent.
+  Therefore, we need a parameter to tolerate the unhealthy application.
+
+### API change
+
+#### PropagationPolicy API change
+
+```golang
+
+// PurgeMode represents that how to deal with the legacy applications on the
+// cluster from which the application is migrated.
+type PurgeMode string
+
+const (
+    // Immediately represents that Karmada will immediately evict the legacy
+    // application.
+    Immediately PurgeMode = "Immediately"
+    // Graciously represents that Karmada will wait for the application to 
+    // come back to healthy on the new cluster or after a timeout is reached 
+    // before evicting the application.
+    Graciously PurgeMode = "Graciously"
+    // Never represents that Karmada will not evict the application and
+    // users manually confirms how to clean up redundant copies.
+    Never PurgeMode = "Never"
+)
+
+
+// PropagationSpec represents the desired behavior of PropagationPolicy.
+type PropagationSpec struct {
+    // ....
+  
+    PropagateDeps bool `json:"propagateDeps,omitempty"`
+  
+    // Failover indicates how Karmada migrates applications in case of failures.
+    // If this value is nil, failover is disabled.
+    // +optional
+    Failover *FailoverBehavior `json:"failover,omitempty"`
+  
+    // ...
+}
+
+// FailoverBehavior indicates failover behaviors in case of an application or
+// cluster failure.
+type FailoverBehavior struct {
+    // Application indicates failover behaviors in case of application failure.
+    // If this value is nil, failover is disabled.
+    // If set, the PropagateDeps should be true so that the dependencies could
+    // be migrated along with the application.
+    // +optional
+    Application *ApplicationFailoverBehavior `json:"application,omitempty"`
+  
+    // Cluster indicates failover behaviors in case of cluster failure.
+    // If this value is nil, failover is disabled.
+    // +optional
+    // Cluster *ClusterFailoverBehavior `json:"cluster,omitempty"`
+}
+
+// ApplicationFailoverBehavior indicates application failover behaviors.
+type ApplicationFailoverBehavior struct {
+    // PreConditions indicates the preconditions of the failover process.
+    // If specified, only when all conditions are met can the failover process be started.
+    // Currently, PreConditions includes several conditions:
+    // - DelaySeconds (optional)
+    // - HealthyState (optional)
+    // +optional
+    PreConditions *PreConditions `json:"preConditions,omitempty"`
+  
+    // DecisionConditions indicates the decision conditions of performing the failover process.
+    // Only when all conditions are met can the failover process be performed.
+    // Currently, DecisionConditions includes several conditions:
+    // - TolerationSeconds (optional)
+    // - HealthyState (mandatory)
+    // +required
+    DecisionConditions *DecisionConditions `json:"decisionConditions,omitempty"`
+  
+    // PurgeMode represents how to deal with the legacy applications on the
+    // cluster from which the application is migrated.
+    // Valid options are "Immediately", "Graciously" and "Never".
+    // Defaults to "Graciously".
+    // +kubebuilder:default=Graciously
+    // +optional
+    PurgeMode PurgeMode `json:"purgeMode,omitempty"`
+  
+    // BlockPredecessorSeconds represents the period of time the cluster from which the
+    // application was migrated from can be schedulable again.
+    // During the period of BlockPredecessorSeconds, clusters are forcibly filtered out by the scheduler. 
+	// If not specified, the scheduler may still schedule the application to the evicted cluster when rescheduling.
+    // Defaults to 600s. Zero means the cluster will never be schedulable.
+    // +kubebuilder:default=600
+    // +optional
+    BlockPredecessorSeconds *int32 `json:"blockPredecessorSeconds,omitempty"`
+}
+
+// PreConditions represents the preconditions of the failover process.
+type PreConditions struct {
+    // DelaySeconds refers to a period of time after the control plane collects
+    // the status of the application for the first time.
+    // If specified, the failover process will be started after DelaySeconds is reached.
+    // It can be used simultaneously with HealthyState and does not affect each other.
+    // +optional
+    DelaySeconds *int32 `json:"delaySeconds,omitempty"`
+  
+    // HealthyState refers to the healthy status reported by the Karmada resource
+    // interpreter.
+    // Valid options are "Healthy".
+    // If specified, the failover process will be started when the application reaches the healthy state.
+    // It can be used simultaneously with DelaySeconds and does not affect each other.
+    // +optional
+    HealthyState ResourceHealth `json:"healthyState,omitempty"`
+}
+
+// DecisionConditions represents the decision conditions of performing the failover process.
+type DecisionConditions struct {
+    // TolerationSeconds represents the period of time Karmada should wait
+    // after reaching the desired state before performing failover process.
+    // If not specified, Karmada will immediately perform failover process.
+    // Defaults to 10s.
+    // +kubebuilder:default=10
+    // +optional
+    TolerationSeconds *int32 `json:"tolerationSeconds,omitempty"`
+  
+    // HealthyState refers to the healthy status reported by the Karmada resource
+    // interpreter.
+    // Valid options are "Unhealthy", "Unknown", or both.
+    // When the application reaches the desired HealthyState, Karmada will perform failover process
+    // after TolerationSeconds is reached.
+    // +kubebuilder:validation:MinItems=1
+    // +required
+    HealthyState []ResourceHealth `json:"healthyState,omitempty"`
+}
+```
+
+For example:
+
+```yaml
+apiVersion: policy.karmada.io/v1alpha1
+kind: PropagationPolicy
+metadata:
+  name: nginx-propagation
+spec:
+  resourceSelectors:
+    - apiVersion: apps/v1
+      kind: Deployment
+      name: nginx
+    - apiVersion: apps/v1
+      kind: StatefulSet
+      name: mysql
+  failover:
+    application:
+      preConditions:
+        delaySeconds: 5000
+      decisionConditions:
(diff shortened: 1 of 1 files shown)


RainbowMango:
We must clearify the realationship with karmada-descheduler. 
cc @Garrybest here.

Poor12:
In my opinion, descheduler focuses more on `Replicas`, whereas this proposal focuses more on `Applications deployed in one specific cluster`. For example, if one deployment is unhealthy in a cluster(2 desired replicas, 1 is pending), the descheduler perfers to keep the healthy replica and reschedule that pending replica to other clusters. And this proposal perfers to remove all replicas and reschedule the deploy to a new clusters. Off course, it will inevitably lead to some duplication of functions in some scenarios.

docs/proposals/rescheduling/README.md
```diff
@@ -0,0 +1,220 @@
...
+    * Custom Configuration for specific clusters
+
+#### PropagationPolicy API change
+
+```
+type PropagationSpec struct {
+    ...
+    // ReschedulingTriggers represents the rescheduling config for unhealthy application state.
+    // +optional
+    ReschedulingTriggers []ReschedulingTrigger `json:"reschedulingTriggers,omitempty"`
```
Poor12, line 73: I am not sure about the default behavior which users do not set rescheduling config.
It can be `never reschedule` or `reschedule all workloads tolerating a default period of time`.
/cc @RainbowMango for opinions.

docs/proposals/rescheduling/README.md
```diff
@@ -0,0 +1,220 @@
...
+
+## Non-Goals
+
+## Proposal
+
+### User Stories (Optional)
+
+#### As a user, I use the cloud vendor's spot instance to deploy my application. As a result, the application may fail due to insufficient resources.
+
+Spot instances are a mode of instances that feature a discounted price and a system interruption mechanism, which means that you can purchase a spot instance ar a discounted price, but the system may automatically repossess it.
```
Garrybest, line 39: I'm also confused about this story, it seems like descheduler could solve this problem by rescheduling replicas from cluster with insufficient resources. What's the difference?

docs/proposals/rescheduling/README.md
```diff
@@ -0,0 +1,220 @@
...
+When users use spot instances to deploy applictions, the probability of an application failing in a vendor's cluster increases significantly. This can also occur in clusters with resource preemption and reclamation.
+At this time, users hope that Karmada provides a rescheduling mechanism.
+
+### As a user, I deploy an application which depends on cluster infrastructure(such as network or storage).
+
+Supposing I deploy an application which depends on cluster infrastructure. Due to the sudden failure of the storage of member1, the application located in member1 suddenly fails.
```
Garrybest, line 45: If there is something wrong with the storage infra in a cluster, how could we know if the application could do well in other clusters? Because maybe the user has provided a wrong config which cannot establish the storage.

The keypoint is how to determine whether the failure could be solved by rescheduling? You know, now descheduler only cares about the insufficient problem because it could never be solved by Kubernetes itself.

docs/proposals/rescheduling/README.md
Poor12: The biggest difference is that the system will allocate resources to you only after you have scheduled it.

docs/proposals/rescheduling/README.md
Poor12: Yeah, it's definitely a multi-cluser issue. Only users can know whether the failure could be solved by rescheduling. So the definition of health status, whether to reschedule and the maximum number of rescheduling should be left to the user to decide.

Poor12:
A configuration item needs to be added to configure the maximum number of rescheduling allowed.

docs/proposals/rescheduling/README.md
Garrybest: Well, could you enlighten me how could a user define the cluster health state?

Poor12: By interpretHealth in `Resource Interpret Framework`.

docs/proposals/rescheduling/README.md
RainbowMango: I understand that this case is dedicated to `serverless cluster` users who bought instances at a limited price, but the cluster will hang the application if other users request for the instance with a higher price.

@Poor12 Could you please add links to some cloud providers' services? like AWS, Google.

RainbowMango: > a spot instance ar a discounted price

What's the `ar` thing?

docs/proposals/rescheduling/README.md
```diff
@@ -0,0 +1,273 @@
...
+## Non-Goals
+
+## Proposal
+
+### User Stories (Optional)
+
+#### As a user, I use the cloud vendor's spot instance to deploy my application. As a result, the application may fail due to insufficient resources.
+
+Spot instances are a mode of instances that feature a discounted price and a system interruption mechanism, which means that you can purchase a spot instance ar a discounted price, but the system may automatically repossess it.
+When users use spot instances to deploy applictions, the probability of an application failing in a vendor's cluster increases significantly. This can also occur in clusters with resource preemption and reclamation.
```
RainbowMango, line 43: > When users use spot instances to deploy applictions, the probability of an application failing in a vendor's cluster increases significantly.

That doesn't make sense, why use a spot instance would `increase` failing probability?

RainbowMango: > This can also occur in clusters with resource preemption and reclamation.

This is another use case, not `spot instance`, right? Then tell the story in another story.

docs/proposals/rescheduling/README.md
RainbowMango: Good point, I think you should add an example of `InterpretHealth` in this proposal.

docs/proposals/rescheduling/README.md
```diff
@@ -0,0 +1,273 @@
...
+### As a user, I deploy an application which depends on cluster infrastructure(such as network or storage).
+
+Supposing I deploy an application which depends on cluster infrastructure. Due to the sudden failure of the storage of member1, the application located in member1 suddenly fails.
+However, member1 is still available in Karmada'view and other applications which do not depend on storage work well.
+At this time, I want to be able to failover this single application.
+
+### Dependency
+
+Karmada's `Resource Interpreter Framework` is designed for interpreting resource structure. It provides users with a interpreter operation to tell Karmada
+how to figure out the health status of a specific object. It is up to users to decide when to reschedule.
```
RainbowMango, line 55: We need an example here.

docs/proposals/rescheduling/README.md
```diff
@@ -0,0 +1,273 @@
...
+require an additional startup time on their first initialization. In such cases, we need a parameter to cover the worse case startup time.
+    * Global Configuration
+    * Custom Configuration for specific resource types
+- Flag for whether to reschedule when the application is unhealthy
+    * Global Configuration
+    * Custom Configuration for specific resource types
+
+#### PropagationPolicy API change
+
+```
```
RainbowMango, line 83:
```suggestion
```golang
```

docs/proposals/rescheduling/README.md
```diff
@@ -0,0 +1,273 @@
...
+    // UnhealthyTolerationSeconds represents the period of time how soon the unhealthy state of applications can be tolerated.
+    // By default, the value is 10 seconds. Minimum value is 1.
+    // +optional
+    UnhealthyTolerationSeconds int32 `json:"unhealthyTolerationSeconds,omitempty"`
+}
+```
+
+For example:
+
+```
```
RainbowMango, line 110:
```suggestion
```yaml
```

docs/proposals/rescheduling/README.md
Poor12: > I understand that this case is dedicated to `serverless cluster` users who bought instances at a limited price, but the cluster will hang the application if other users request for the instance with a higher price.
> 
> @Poor12 Could you please add links to some cloud providers' services? like AWS, Google.

updated.

docs/proposals/rescheduling/README.md
Poor12: added.

docs/proposals/rescheduling/README.md
Poor12: > > This can also occur in clusters with resource preemption and reclamation.
> 
> This is another use case, not `spot instance`, right? Then tell the story in another story.

This is a mistake. The key is clusters with a dynamically resource pool. If the resource pool is stable, estimator will get the accurate resource margin.

docs/proposals/rescheduling/rescheduling.md
```diff
@@ -0,0 +1,312 @@
...
+}
+
+// ReschedulingTrigger represents the rescheduling trigger for matched resource objects.
+type ReschedulingTrigger struct {
+    // Index represents the corresponding index of ResourceSelectors.
+    // If this value is not set, it means that all resources hit by ResourceSectors use the same configuration.
+    // If the resource meets the conditions of multiple ResourceSelectors at the same time,
+    // it will use the matching configuration with the highest priority.
+    // +optional
+    Index *int32 `json:"index,omitempty"`
```
chaunceyjiang, line 115: Can we mimic the `downward API`?

First, we could rename the field to `ResourceSelectorsPath  json:"resourceSelectorsPath,omitempty"`,

and then allow the user to set it like this `resourceSelectorsPath: .spec.resourceSelectors` or `resourceSelectorsPath: .spec.resourceSelectors[0]`. 

`resourceSelectorsPath: .spec. resourceSelectors`  means that all resources hit by ResourceSectors use the same configuration.

RainbowMango:
Propose a very primitive alternative here.

The main idea is that we should allow users to customize the behavior for both `application failover` and `cluster failover`.

**Introduce a filed to PropagationPolicy to declare the behavior of failover**

```golang
type PropagationSpec struct {
	// ....

	PropagateDeps bool `json:"propagateDeps,omitempty"`

	// Failover indicates how Karmada migrates applications in case of failures.
	// If this value is nil, failover is disabled.
	// +optional
	Failover *FailoverBehavior `json:"failover,omitempty"`

	// ...
}
```

**The failover behavior should consist of both application and cluster**

```golang
// FailoverBehavior indicates failover behaviors in case of an application or
// cluster failure.
type FailoverBehavior struct {
	// Application indicates failover behaviors in case of application failure.
	// If this value is nil, failover is disabled.
	// If set, the PropagateDeps should be true so that the dependencies could
	// be migrated along with the application.
	// +optional
	Application *ApplicationFailoverBehavior `json:"application,omitempty"`

	// Cluster indicates failover behaviors in case of cluster failure.
	// If this value is nil, failover is disabled.
	// +optional
	Cluster *ClusterFailoverBehavior `json:"cluster,omitempty"`
}
```

**application failover behavior**

```golang
// ApplicationFailoverBehavior indicates application failover behaviors.
type ApplicationFailoverBehavior struct {
	// InitiateGates indicates the preconditions of the failover process.
	// If specified, all gates will be evaluated for prerequisites.
	// +optional
	InitiateGates []FailoverInitiateGate `json:"initiateGates,omitempty"`

	// ActiveGates indicates the conditions of performing the failover process.
	//
	// +kubebuilder:validation:MinItems=1
	// +required
	ActiveGates []FailoverActiveGate `json:"activeGates"`

	// TolerationSeconds represents the period of time the Karmada should wait
	// before performing failover. Defaults to 10s.
	//
	// +kubebuilder:default=10
	// +optional
	TolerationSeconds int32

	// PurgeMode represents how to deal with the legacy applications on the
	// cluster from which the application is migrated.
	// Valid options are "Immediately", "Graciously" and "Never".
	// Defaults to "Graciously".
	//
	// +kubebuilder:validation:Enum=Immediately;Graciously;Never
	// +optional
	PurgeMode string `json:"purgeMode,omitempty"`

	// EscapeSeconds represents the period of time the cluster from which the
	// application was migrated from should be escaped.
	// Defaults to 600s.
	//
	// +kubebuilder:default=600
	// +optional
	EscapeSeconds int32 `json:"escapeSeconds,omitempty"`
}


// FailoverInitiateGate represents the preconditions of the failover process.
type FailoverInitiateGate struct {
	// ConditionType refers to a condition in the object's condition list with
	// matching type. If the status of specified condition is "true" means pass.
	// +optional
	ConditionType string `json:"conditionType,omitempty"`

	// HealthyState refers to the healthy status reported by the Karmada resource
	// interpreter.
	// Valid options are "Unhealthy" and "Unknown".
	// +optional
	HealthyState string `json:"healthyState,omitempty"`
}

// FailoverActiveGate represents the conditions of performing the failover process.
type FailoverActiveGate struct {
	// ConditionType refers to a condition in the object's condition list with
	// matching type. If the status of the specified condition is "true" means pass.
	// +optional
	ConditionType string `json:"conditionType,omitempty"`

	// HealthyState refers to the healthy status reported by the Karmada resource
	// interpreter.
	// Valid options are "Unhealthy" and "Unknown".
	// +optional
	HealthyState string `json:"healthyState,omitempty"`
}
```

**cluster failover behavior should be reserved even though this is not the scope of this proposal**

```golang
// ClusterFailoverBehavior indicates application failover behaviors in case of
// cluster failure.
type ClusterFailoverBehavior struct {
	// PurgeMode represents how to deal with the legacy applications on the
	// cluster from which the application is migrated.
	// Valid options are "Immediately", "Graciously" and "Never".
	// Defaults to "Graciously".
	//
	// +kubebuilder:validation:Enum=Immediately;Graciously;Never
	// +optional
	PurgeMode string `json:"purgeMode,omitempty"`

	// EscapeSeconds represents the period of time the cluster from which the
	// application was migrated from should be escaped.
	// Defaults to 600s.
	//
	// +kubebuilder:default=600
	// +optional
	EscapeSeconds int32 `json:"escapeSeconds,omitempty"`
}
```

docs/proposals/failover/application-failover.md
```diff
@@ -0,0 +1,373 @@
...
+  resourceSelectors:
+    - apiVersion: apps/v1
+      kind: Deployment
+      name: nginx
+    - apiVersion: apps/v1
+      kind: StatefulSet
+      name: mysql
+  failover:
+    - apiVersion: apps/v1
+      kind: Deployment
```
chaunceyjiang, line 230: defined here `.spec.failover[0]. apiVersion` and `.spec.failover[0]. kind` mean?

What happens if I have two pp and apiVersion and kind are set to the same for both pp, but `behavior.application.activeGates` is not the same?

Poor12: Sorry for the slow reply. For these two pp, one resource will only apply one of above policies. So it will use the failover behavior in the hit policy.

docs/proposals/failover/application-failover.md
```diff
@@ -0,0 +1,373 @@
...
+// FailoverBehavior indicates failover behaviors in case of an application or
+// cluster failure.
+type FailoverBehavior struct {
+    // APIVersion represents the API version of the target resources.
+    // +optional
+    APIVersion string `json:"apiVersion,omitempty"`
+  
+    // Kind represents the Kind of the target resources.
+    // +optional
+    Kind string `json:"kind,omitempty"`
```
RainbowMango, line 125: What's the behavior of the two fields(`apiVersion` and `kind`) not set?

docs/proposals/failover/application-failover.md
RainbowMango: Personality, I don't like to introduce an extra layer to declare the `apiVersion/kind` that the failover rules applied to.

My concern is for most cases, there is usually only one application in the PropagatioPolicy, but this design forces people to declare the `apiVersion/kind` one more time, a little bit redundant.

But, on the other hand, we might need to support declaring different behaviors for different applications in the same PropagationPolicy. 

I don't have a better idea yet.

RainbowMango: > (1) When users set different applications in the same pp and config the behavior by condition, they may need to live with the configuration of one of the applications not taking effect.

I probably don't buy it if I'm the user.

docs/proposals/failover/application-failover.md
```diff
@@ -0,0 +1,506 @@
...
+// FailoverActiveGate represents the conditions of performing the failover process.
+type FailoverActiveGate struct {
+    // ConditionType refers to a condition in the object's condition list with
+    // matching type. If the status of the specified condition is "true" means pass.
+    // +optional
+    ConditionType string `json:"conditionType,omitempty"`
+  
+    // HealthyState refers to the healthy status reported by the Karmada resource
+    // interpreter.
+    // Valid options are "Unhealthy" and "Unknown".
```
Poor12, line 340: Added two optional API designs. Meanwhile, I think `Unknown` may not be suitable as an option. As far as I know, it will only occur when users do not config the `InterpretHealth` or the config is wrong.

Poor12:
Summarizes all cases where all copies are kept.
image

I think we may provide some strategy with users about which copy will be keep.

docs/proposals/failover/application-failover.md
```diff
@@ -0,0 +1,566 @@
...
+## Proposal
+
+### User Stories (Optional)
+
+#### As a user, I deploy an application in multiple clusters with preemptive scheduling. When business peaks, low-priority applications cannot run normally for a long time.
+
+Supposing I deploy an application in a cluster with preemptive scheduling. When cluster resources are in short supply, low-priority applications that were running normally are preempted and cannot run normally for a long time.
+At this time, applications cannot self-heal within a cluster. Users wants to try to schedule it to another cluster to ensure that the service is continuously served and the replicas does not drop to zero.
+
+#### As a user, I deploy an application in multiple clusters for HA. I want to upgrade one of the clusters and migrate to the original cluster when a single point of failure occurs in the higher version cluster.
```
RainbowMango, line 60: I didn't get this story, i think if the cluster administrators want to upgrade the cluster, they might try to `drain` a cluster.

> and migrate to the original cluster

I didn't get what the `original cluster` means.

docs/proposals/failover/application-failover.md
```diff
@@ -0,0 +1,566 @@
...
+    Health ResourceHealth `json:"health,omitempty"`
+
+    // CreationTimestamp is a timestamp representing the server time when this object was
+    // created.
+    // Clients should not set this value to avoid the time inconsistency issue.
+    // It is represented in RFC3339 form(like '2021-04-25T10:02:10Z') and is in UTC.
+    //
+    // Populated by the system. Read-only.
+    // +optional
+    CreationTimestamp metav1.Time `json:"creationTimestamp,omitempty"`
```
RainbowMango, line 300: I can't get the meaning and how it works from this comment.
Could you please explain more?

docs/proposals/failover/application-failover.md
```diff
@@ -0,0 +1,566 @@
...
+```
+
+#### ResourceBinding API change
+
+```golang
+type ResourceBindingSpec struct {
+    ...
+    // EvictedClusters represents the evicted clusters.
+    // +optional
+    EvictedClusters []EvictedCluster `json:"evictedClusters,omitempty"`
```
RainbowMango, line 258: I don't think this field should be put in `spec`, because it's not a declaration, actually, it is a kind of intermediate that is used by controllers(scheduler).

docs/proposals/failover/application-failover.md
```diff
@@ -0,0 +1,566 @@
...
+    Name string `json:"name,omitempty"`
+    //
+    // CreationTimestamp is a timestamp representing the server time when this object was
+    // created.
+    // Clients should not set this value to avoid the time inconsistency issue.
+    // It is represented in RFC3339 form(like '2021-04-25T10:02:10Z') and is in UTC.
+    //
+    // Populated by the system. Read-only.
+    // +optional
+    CreationTimestamp metav1.Time `json:"creationTimestamp,omitempty"`
```
RainbowMango, line 315: Same as above, we need more clear comments for it.

By the way, it will be used to record the eviction timestamp, but for field naming, it deserves a more meaningful name.

docs/proposals/failover/application-failover.md
```diff
@@ -0,0 +1,566 @@
...
+* A Period of time. We can wait a period of time before the application failover is initiated.
+
+For some short-lived exceptions(which can be recovered after a period of time), we don't want to immediately migrate the application to other clusters which may make the migration very frequent.
+Therefore, we need a parameter to tolerate the unhealthy application.
+
+### API change
+
+#### PropagationPolicy API change
+
+Option One:
```
RainbowMango, line 126: We don't need to highlight this is `option one`.  The design present here means this is the proposing option.

docs/proposals/failover/application-failover.md
Poor12: I think it's a bit like `GracefulEvictionTasks`. It may also has a scene. The user manually fixed the failure within the recovery time set by the cluster. Users may manually make a cluster schedulable.
BTW, I have also considered using the taint mechanism to achieve this function, but it will introduce `cluster` object, which may make the function more complicated.

docs/proposals/failover/application-failover.md
Poor12: `Drain` is a solution. However, the current mainstream cluster upgrade method does not need to migrate to existing services. For example, [CCE](https://support.huaweicloud.com/intl/zh-cn/usermanual-cce/cce_01_0197.html).

docs/proposals/failover/application-failover.md
```diff
@@ -0,0 +1,566 @@
...
+## Non-Goals
+
+## Proposal
+
+### User Stories (Optional)
+
+#### As a user, I deploy an application in multiple clusters with preemptive scheduling. When business peaks, low-priority applications cannot run normally for a long time.
+
+Supposing I deploy an application in a cluster with preemptive scheduling. When cluster resources are in short supply, low-priority applications that were running normally are preempted and cannot run normally for a long time.
+At this time, applications cannot self-heal within a cluster. Users wants to try to schedule it to another cluster to ensure that the service is continuously served and the replicas does not drop to zero.
```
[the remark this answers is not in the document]
Poor12: Make sense.

docs/proposals/failover/application-failover.md
Poor12: After I thought about it, I'd like to pick the name `EvictionHistory` or `ActiveEvictionHistory`.

docs/proposals/failover/application-failover.md
```diff
@@ -0,0 +1,562 @@
...
+// ApplicationFailoverBehavior indicates application failover behaviors.
+type ApplicationFailoverBehavior struct {
+    // InitiateGates indicates the preconditions of the failover process.
+    // If specified, all gates will be evaluated for prerequisites.
+    // +optional
+    InitiateGates *FailoverInitiateGate `json:"initiateGates,omitempty"`
+  
+    // ActiveGates indicates the conditions of performing the failover process.
+    //
+    // +kubebuilder:validation:MinItems=1
```
XiShanYongYe-Chang, line 161: Is this item limit unnecessary?

docs/proposals/failover/application-failover.md
```diff
@@ -0,0 +1,562 @@
...
+    ActiveGates *FailoverActiveGate `json:"activeGates"`
+  
+    // PurgeMode represents how to deal with the legacy applications on the
+    // cluster from which the application is migrated.
+    // Valid options are "Immediately", "Graciously" and "Never".
+    // Defaults to "Graciously".
+    //
+    // +kubebuilder:validation:Enum=Immediately;Graciously;Never
+    // +optional
+    PurgeMode string `json:"purgeMode,omitempty"`
```
XiShanYongYe-Chang, line 172: Would you consider creating a type and explaining what each value means?

docs/proposals/failover/application-failover.md
```diff
@@ -0,0 +1,562 @@
...
+  
+    // EscapeSeconds represents the period of time the cluster from which the
+    // application was migrated from should be escaped.
+    // Defaults to 600s.
+    //
+    // +kubebuilder:default=600
+    // +optional
+    EscapeSeconds int32 `json:"escapeSeconds,omitempty"`
```
XiShanYongYe-Chang, line 180: int32 -> *int32 ?

docs/proposals/failover/application-failover.md
```diff
@@ -0,0 +1,562 @@
...
+    // Valid options are "Immediately", "Graciously" and "Never".
+    // Defaults to "Graciously".
+    //
+    // +kubebuilder:validation:Enum=Immediately;Graciously;Never
+    // +optional
+    PurgeMode string `json:"purgeMode,omitempty"`
+  
+    // EscapeSeconds represents the period of time the cluster from which the
+    // application was migrated from should be escaped.
+    // Defaults to 600s.
```
XiShanYongYe-Chang, line 176: Is the default value too long? Can it be considered to be the same as the default value with failover?

docs/proposals/failover/application-failover.md
```diff
@@ -0,0 +1,562 @@
...
+
+// FailoverInitiateGate represents the preconditions of the failover process.
+type FailoverInitiateGate struct {
+    // DelaySeconds refers to a period of time after the application has started before
+    // the application failover is initiated.
+    // +optional
+    DelaySeconds int32 `json:"delaySeconds,omitempty"`
+  
+    // HealthyState refers to the healthy status reported by the Karmada resource
+    // interpreter.
```
XiShanYongYe-Chang, line 192: I think there is no need to specifically point out the resource interpreter in this comment.

docs/proposals/failover/application-failover.md
```diff
@@ -0,0 +1,562 @@
...
+}
+
+// FailoverActiveGate represents the conditions of performing the failover process.
+type FailoverActiveGate struct {
+    // TolerationSeconds represents the period of time the Karmada should wait
+    // before performing failover. Defaults to 10s.
+    //
+    // +kubebuilder:default=10
+    // +optional
+    TolerationSeconds int32
```
XiShanYongYe-Chang, line 205: int32 -> *int32 ?

docs/proposals/failover/application-failover.md
```diff
@@ -0,0 +1,562 @@
...
+
+    // HealthyState refers to the healthy status reported by the Karmada resource
+    // interpreter.
+    // Valid options are "Unhealthy" and "Unknown".
+    // +optional
+    HealthyState []string `json:"healthyState,omitempty"`
```
XiShanYongYe-Chang, line 211: Why is this a slice?

docs/proposals/failover/application-failover.md
```diff
@@ -0,0 +1,562 @@
...
+    // If this value is nil, failover is disabled.
+    // If set, the PropagateDeps should be true so that the dependencies could
+    // be migrated along with the application.
+    // +optional
+    Application *ApplicationFailoverBehavior `json:"application,omitempty"`
+  
+    // Cluster indicates failover behaviors in case of cluster failure.
+    // If this value is nil, failover is disabled.
+    // +optional
+    Cluster *ClusterFailoverBehavior `json:"cluster,omitempty"`
```
XiShanYongYe-Chang, line 149: What does the `ClusterFailoverBehavior` looks like?

docs/proposals/failover/application-failover.md
XiShanYongYe-Chang: What is the difference between this field and `GracefulEvictionTasks`?

docs/proposals/failover/application-failover.md
Poor12: There is a difference in the default value here. If you use int32 directly, the default value may be a certain time. If you use a pointer, the default value will be never escaped. What do you think is better?

docs/proposals/failover/application-failover.md
Poor12: I think filling it in allows users to understand how this field is configured.

docs/proposals/failover/application-failover.md
Poor12: Ditto.

RainbowMango (approved):
/lgtm
/approve

Issue #2757, opened 2022-11-08
What's the version relationship of kubernetes between karmada control plane and member?


carlory:
**Please provide an in-depth description of the question you have**:

I have two member clusters, one is kubernetes v1.25 and the other is kubernetes v1.18. I want to propagate an ingress to these two clusters. Those two member clusters don't share a possible ingress version. I don't known how to choose the right kuberentes version for karmada control plane.

```sh
# v1.25
~ kubectl api-resources | grep ingress
ingresses                         ing                  networking.k8s.io/v1               true         Ingress

# v1.18
~ kubectl api-resources | grep ingress
ingresses                         ing                  extensions/v1beta1                      true         Ingress
ingresses                         ing                  networking.k8s.io/v1beta1               true         Ingress
```

How does karmada control plane deal with the different versions of kubernetes when users want to propagate a ingress resource? I cannot find the answer in the office documentation.

when I want to propagate a ingress (v1beta1) into all the member clusters, the karmada-scheduler won't schedule it to the v1.25 cluster due to the AIPEnablement plugin limitation.

If I want to propagate a ingress (v1) into all the member clusters, the karmada-apiserver must contain the v1 ingress API according to the current implementation.

In this case, I cannot find a suitable kubernetes version for karmada control plane.

This [link](https://github.com/karmada-io/karmada#kubernetes-compatibility) says the kubernetes-comaptibility of karmada. but I don't know that this  kubernetes-comaptibility table is designed for karmada control plane or member cluster? 

for example, the mcs feature requires that the karmada control plane and member clusters must be v1.21 or higher. However, according to the kubernetes-comaptibility table, the range of v1.16 - v1.20 is supported. 

In summary, the core question is that the karmada how to handle the different versions of kubernetes between karmada control plane and member clusters?

**What do you think about this question?**:

Karmada should support the different versions of kubernetes between karmada control plane and member clusters. As karmada users, they should not care about the resoure version of member clusters. They just need to propagate the resource to all the member clusters that supports the resource.

RainbowMango:
Good question. I think it'd be better to have a document to clarify this question. 
/assign

RainbowMango:
As discussed at the community meeting, I'm going to add a document to explain the `Kubernetes compatibility` thing.
Sorry for the delay, but it is still on my agenda and I hope to get back to this next month.

dg424:
Is there a final answer to this question ?

RainbowMango:
[...]

PR #3135, opened 2023-02-09
Update default kind version and testing against v1.26.0

RainbowMango:
**What type of PR is this?**
/kind cleanup

The default kind version is too old now when using the latest `kubectl` it rise a warning, like:
```
WARNING: version difference between client (1.26) and server (1.23) exceeds the supported minor version skew of +/-1
```

This PR also adds Kubernetes `v1.26.0` to the E2E matrix.

Fixes #

1. I leave the [Kubernetes compatibility](https://github.com/karmada-io/karmada#kubernetes-compatibility) un-updated because we need to explain more about the compatibility and it is tracked by #2757.
2. We now got 6 Kubernetes versions in E2E matrix, we can't afford all versions, we need the rule to get rid of the burden. (no idea about that yet)

```release-note
NONE
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -107,7 +107,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        k8s: [ v1.21.10, v1.22.7, v1.23.4, v1.24.2, v1.25.0 ]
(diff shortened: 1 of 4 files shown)


XiShanYongYe-Chang (commented):
/lgtm

The tag of the documentation should be removed. We have not modified the document.

PR #3136, opened 2023-02-09
[bug-fix] update clueter-cluster shell

XiShanYongYe-Chang:

**What type of PR is this?**

/kind bug

image

The cluster fails to be created using the `hack/create-cluster.sh` script, and the installation of the weave fails because the installation address of the weave is changed. We installed the weave to use the submarier, but now the submarier does not use the weave. For details, see [here](https://github.com/submariner-io/shipyard/issues/858). Therefore, we also modify the weave accordingly.

Fixes #

```release-note
NONE
```


RainbowMango (approved):
/lgtm
/approve

PR #3142, opened 2023-02-11
Add bindingStatus controller to aggregate work status and update the status of resource template

XiShanYongYe-Chang:

**What type of PR is this?**

/kind cleanup
/kind feature

Currently, the synchronization and state collection of binding are both processed in one controller, and there will be redundant calculations, because sometimes only synchronization is required, and sometimes only state collection is required, so I moved the logic of state collection from the binding-controller isolated.

Fixes #

```release-note
`karmada-controller`: add `bindingStatus` controller to aggregate work status and update the status of resource template.
```


XiShanYongYe-Chang:
/hold

Before pushing it forward, we need to resolve the below problem:
After analyzing the CI error, we need to consider the following problem during graceful eviction, after the faulty cluster recovers, the cluster is not removed from the graceful eviction queue, and the new scheduling result contains the cluster. The cluster appears in both `spec.gracefulEvictionTasks` and `spec.clusters`, it's a scene of conflict.

Poor12:
> /hold
> 
> Before pushing it forward, we need to resolve the below problem: After analyzing the CI error, we need to consider the following problem during graceful eviction, after the faulty cluster recovers, the cluster is not removed from the graceful eviction queue, and the new scheduling result contains the cluster. The cluster appears in both `spec.gracefulEvictionTasks` and `spec.clusters`, it's a scene of conflict.

The key problem is not “the cluster apprears in both spec.gracefulEvictionTasks and spec.clusters. When a cluster fails and recovers quickly, we will remove the orphan work in the failed cluster and rebuild a new one when it recovers. However, the period is too short. So we add the deletetimestamp to the work but the work is not deleted now. Then `controllerruntime.CreateOrUpdate` will update the work but it will soon be deleted by `execution_controller`. 
This will prevent new work from being created.

Poor12:
Add judgment logic after executing the CreateOrUpdate method:
```
key := pkgclient.ObjectKeyFromObject(runtimeObject)
err = client.Get(context.TODO(), key, runtimeObject)
if err != nil || !runtimeObject.DeletionTimestamp.IsZero() {
    return fmt.Errorf("failed to get work %s/%s or it's being deleted", runtimeObject.GetNamespace(), runtimeObject.GetName())
}
```

XiShanYongYe-Chang:
This task is planned to be completed by #3202 
/close

PR #3149, opened 2023-02-13
singleClusterInformerManager != nil return early

helen-frank:
**What type of PR is this?**
/kind feature

singleClusterInformerManager != nil return early
Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/status/cluster_status_controller.go
+++ b/pkg/controllers/status/cluster_status_controller.go
@@ -305,9 +305,11 @@ func (c *ClusterStatusController) initializeGenericInformerManagerForCluster(clu
 // and pod and start it. If the informer manager exist, return it.
 func (c *ClusterStatusController) buildInformerForCluster(clusterClient *util.ClusterClient) (typedmanager.SingleClusterInformerManager, error) {
 	singleClusterInformerManager := c.TypedInformerManager.GetSingleClusterManager(clusterClient.ClusterName)
-	if singleClusterInformerManager == nil {
-		singleClusterInformerManager = c.TypedInformerManager.ForCluster(clusterClient.ClusterName, clusterClient.KubeClient, 0)
+	if singleClusterInformerManager != nil {
+		return singleClusterInformerManager, nil
 	}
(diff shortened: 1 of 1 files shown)


Poor12:
Hi @helen-frank, can you explain the problem briefly?

helen-frank:
> Hi @helen-frank, can you explain the problem briefly?

Make the implementation conform to the description of the comment
https://github.com/karmada-io/karmada/blob/5d9f0ec49b0befb64d07ef8fc06c44b6a21dc8d6/pkg/controllers/status/cluster_status_controller.go#L304-L306

pkg/controllers/status/cluster_status_controller.go
```diff
@@ -305,10 +305,12 @@ func (c *ClusterStatusController) initializeGenericInformerManagerForCluster(clu
 // and pod and start it. If the informer manager exist, return it.
 func (c *ClusterStatusController) buildInformerForCluster(clusterClient *util.ClusterClient) (typedmanager.SingleClusterInformerManager, error) {
 	singleClusterInformerManager := c.TypedInformerManager.GetSingleClusterManager(clusterClient.ClusterName)
-	if singleClusterInformerManager == nil {
-		singleClusterInformerManager = c.TypedInformerManager.ForCluster(clusterClient.ClusterName, clusterClient.KubeClient, 0)
+	if singleClusterInformerManager != nil {
+		return singleClusterInformerManager, nil
```
RainbowMango, line 309: I don't think we can just return here because we must ensure the informer for `Node and Pod` has been set up and started.

RainbowMango (commented):
/hold

pkg/controllers/status/cluster_status_controller.go
helen-frank: Got it, it seems that the comments are somewhat distorted.

helen-frank: Got it, it seems that my understanding has some errors.
/close

PR #2987, opened 2022-12-22
The expiration time of cert is configurable。

hexiaofeng:
The expiration time of cert by kubectl-karmada is configurable。

the default value is 10 years

PR #3156, opened 2023-02-14
Make the validity period of cert configurable

### LINKED ISSUE #2985: The expiration time of the cert
The expiration time of the cert by  kubectl-karmada is 1 year

``` golang
func (i *CommandInitOption) genCerts() error {
	notAfter := time.Now().Add(cert.Duration365d).UTC()
        ......
}
```

The expiration time of the cert by  deploy-karmada.sh is 10 years

```  shell
function util::create_signing_certkey {
    ......
    ${OPENSSL_BIN} req -x509 -sha256 -new -nodes -days 3650 -newkey rsa:2048 -keyout "${dest_dir}/${id}.key" -out "${dest_dir}/${id}.crt" -subj "/CN=${cn}/"
    ......
}
``` 

Expectations are 10 years

lonelyCZ:

**What type of PR is this?**
/kind feature

Fixes #2985 

This is continuing develoption of #2987 that has been hold too long. And it is useful for users.

What do you think? Solfly ping @hexiaofeng 

```release-note
karmadactl: import `--cert-validity-period` for `init` to make the validity period of cert configurable
```

### DIFF
--- a/pkg/karmadactl/cmdinit/cmdinit.go
+++ b/pkg/karmadactl/cmdinit/cmdinit.go
@@ -8,6 +8,7 @@ import (
 	"k8s.io/klog/v2"
 	"k8s.io/kubectl/pkg/util/templates"
 
+	"github.com/karmada-io/karmada/pkg/karmadactl/cmdinit/cert"
 	"github.com/karmada-io/karmada/pkg/karmadactl/cmdinit/kubernetes"
 	"github.com/karmada-io/karmada/pkg/karmadactl/util"
 	"github.com/karmada-io/karmada/pkg/version"
@@ -102,6 +103,7 @@ func NewCmdInit(parentCommand string) *cobra.Command {
 	// cert
 	flags.StringVar(&opts.ExternalIP, "cert-external-ip", "", "the external IP of Karmada certificate (e.g 192.168.1.2,172.16.1.2)")
 	flags.StringVar(&opts.ExternalDNS, "cert-external-dns", "", "the external DNS of Karmada certificate (e.g localhost,localhost.com)")
+	flags.DurationVar(&opts.CertValidity, "cert-validity-period", cert.Duration365d, "the validity period of Karmada certificate (e.g 8760h0m0s, that is 365 days)")
 	// Kubernetes
 	flags.StringVarP(&opts.Namespace, "namespace", "n", "karmada-system", "Kubernetes namespace")
 	flags.StringVar(&opts.StorageClassesName, "storage-classes-name", "", "Kubernetes StorageClasses Name")
--- a/pkg/karmadactl/cmdinit/kubernetes/deploy.go
+++ b/pkg/karmadactl/cmdinit/kubernetes/deploy.go
@@ -121,6 +121,7 @@ type CommandInitOption struct {
 	CRDs                               string
 	ExternalIP                         string
 	ExternalDNS                        string
+	CertValidity                       time.Duration
 	KubeClientSet                      kubernetes.Interface
 	CertAndKeyFileData                 map[string][]byte
 	RestConfig                         *rest.Config
@@ -218,7 +219,7 @@ func initializeDirectory(path string) error {
 
 // genCerts create ca etcd karmada cert
 func (i *CommandInitOption) genCerts() error {
-	notAfter := time.Now().Add(cert.Duration365d).UTC()
+	notAfter := time.Now().Add(i.CertValidity).UTC()
 
 	etcdServerCertDNS := []string{
 		"localhost",


pkg/karmadactl/cmdinit/cmdinit.go
```diff
@@ -102,6 +103,7 @@ func NewCmdInit(parentCommand string) *cobra.Command {
 	// cert
 	flags.StringVar(&opts.ExternalIP, "cert-external-ip", "", "the external IP of Karmada certificate (e.g 192.168.1.2,172.16.1.2)")
 	flags.StringVar(&opts.ExternalDNS, "cert-external-dns", "", "the external DNS of Karmada certificate (e.g localhost,localhost.com)")
+	flags.DurationVar(&opts.ValidityPeriod, "cert-validity-period", cert.Duration365d, "the validity period of Karmada certificate (e.g 8760h0m0s: default 365d)")
```
RainbowMango, line 106:
```suggestion
	flags.DurationVar(&opts.ValidityPeriod, "cert-validity-period", cert.Duration365d, "The validity period of Karmada certificate (e.g 8760h0m0s: default 365d)")
```
Does `Karmada cerificate` means all signed certificates, including both `client` and `server`?

By the way, we don't need to specify the default value again after we set the default value. The framework would set the default value.

RainbowMango: > e.g 8760h0m0s
Can people set the flag by `year` /`month`/`day`?

RainbowMango: I suggest renaming the `cert.Duration365d` to `cert.DefaultCertificateValidity`.

pkg/karmadactl/cmdinit/kubernetes/deploy.go
```diff
@@ -121,6 +121,7 @@ type CommandInitOption struct {
 	CRDs                               string
 	ExternalIP                         string
 	ExternalDNS                        string
+	ValidityPeriod                     time.Duration
```
RainbowMango, line 124:
```suggestion
	CertificateValidity                     time.Duration
```
How about this?

pkg/karmadactl/cmdinit/kubernetes/deploy.go
```diff
@@ -218,7 +219,7 @@ func initializeDirectory(path string) error {
 
 // genCerts create ca etcd karmada cert
 func (i *CommandInitOption) genCerts() error {
-	notAfter := time.Now().Add(cert.Duration365d).UTC()
```
RainbowMango, line 221: There is more than one place referencing this default variable,  please check if we get it from the option.

pkg/karmadactl/cmdinit/cmdinit.go
lonelyCZ: > Does Karmada cerificate means all signed certificates, including both client and server?

Yes, include certs of all components.

> Can people set the flag by `year` /`month`/`day`?

It seems not working, it is only recognizing the default format(hour/minute/second) of time.Duration. It should be more flexible about fine grained time.

> By the way, we don't need to specify the default value again after we set the default value. The framework would set the default value.

Yes, but `(e.g 8760h0m0s: default 365d)` is only for commenting `8760h0m0s == 365d`.

pkg/karmadactl/cmdinit/kubernetes/deploy.go
lonelyCZ: Thanks for pointing that, it has some problems and I am going to fix it.

pkg/karmadactl/cmdinit/cmdinit.go
RainbowMango: OK. get it. Please echo the output of `karmadactl init --help`, especially for this flag part.
Maybe we can comment it like `e.g 8760h0m0s, that is 365d`.

lonelyCZ:
```
Options:
    --cert-external-dns='':
        the external DNS of Karmada certificate (e.g localhost,localhost.com)

    --cert-external-ip='':
        the external IP of Karmada certificate (e.g 192.168.1.2,172.16.1.2)

    --cert-validity-period=8760h0m0s:
        the validity period of Karmada certificate (e.g 8760h0m0s, that is 365 days)
```

pkg/karmadactl/cmdinit/kubernetes/deploy.go
lonelyCZ: `Duration365d` is used here, and it only work when `cfg.NotAfter == nil`, but the `cfg.NotAfter` has been filled when `NewCertConfig`.

https://github.com/karmada-io/karmada/blob/b1f8b09d073d9f856aef90297e5ae3f7c780ee00/pkg/karmadactl/cmdinit/cert/cert.go#L110-L113

https://github.com/karmada-io/karmada/blob/b1f8b09d073d9f856aef90297e5ae3f7c780ee00/pkg/karmadactl/cmdinit/cert/cert.go#L234-L244

https://github.com/karmada-io/karmada/blob/b1f8b09d073d9f856aef90297e5ae3f7c780ee00/pkg/karmadactl/cmdinit/kubernetes/deploy.go#L233-L234

RainbowMango (approved):
/lgtm
/approve

Issue #3143, opened 2023-02-11
unexcepted behavior of `names.GenerateBindingReferenceKey`


zirain:
```
left := names.GenerateBindingReferenceKey("a","b-c") 
right := names.GenerateBindingReferenceKey("a-b","c") 
```

left should not equal to right

cc @XiShanYongYe-Chang

jwcesign:
I tested, and for K8s resource, '/' is not allowed, so how about changing `GenerateBindingReferenceKey` to:
![image](https://user-images.githubusercontent.com/19745536/218926755-f65fb798-415d-4ec6-a948-6b6708254832.png)

jwcesign:
cc @XiShanYongYe-Chang

PR #3157, opened 2023-02-15
Make the result unique for GenerateBindingReferenceKey function

jwcesign:
**What type of PR is this?**
Use `/` as the delimiter for  `GenerateBindingReferenceKey` function

In the following case, the key is not unique:
```go
left := names.GenerateBindingReferenceKey("a","b-c") 
right := names.GenerateBindingReferenceKey("a-b","c") 
```

Fixes https://github.com/karmada-io/karmada/issues/3143

```release-note
`karmada-controller-manager`: Make the result unique for `GenerateBindingReferenceKey` function.
```

### DIFF
--- a/pkg/util/names/names.go
+++ b/pkg/util/names/names.go
@@ -69,7 +69,7 @@ func GenerateBindingName(kind, name string) string {
 func GenerateBindingReferenceKey(namespace, name string) string {
 	var bindingName string
 	if len(namespace) > 0 {
-		bindingName = namespace + "-" + name
+		bindingName = namespace + "/" + name
 	} else {
 		bindingName = name
 	}
--- a/pkg/util/names/names_test.go
+++ b/pkg/util/names/names_test.go
@@ -92,6 +92,14 @@ func TestGenerateBindingReferenceKey(t *testing.T) {
 			name:      "mytest-deployment",
 			namespace: "",
 		},
+		{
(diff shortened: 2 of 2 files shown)


XiShanYongYe-Chang:
Thanks~
/assign

jwcesign:
### Test result
namespace `a`, deployment `b-c`, with the old version(delimiter is `-`):


namespace `a-b`, deployment `c`, with the new version(delimiter is `/`):


After delete pp/rb for deployment `c`, it hasn't affection to deployment `b-c` :

jwcesign:
Now, this function will generate the same key, which will cause the label same for different deployments:
ns/deployment: a/b-c
image
ns/deployment: a-b/c
image

But there is a function to check whether it's the same key and the same deployment name, so it solves this conflict.
https://github.com/karmada-io/karmada/blob/e523977cb0b8eb0367dd7bfb9ac5820c235f2a47/pkg/util/helper/work.go#L107

So does we need this PR?

XiShanYongYe-Chang:
Ask for @RainbowMango to take a look.
/cc @RainbowMango

RainbowMango:
Please update the unit test to cover this case.

RainbowMango (approved):
/lgtm
/approve

By the way, I don't think we need to cherry-pick this patch to release branches.

PR #3158, opened 2023-02-15
update cluster condition to unknow when cluster is offline

XiShanYongYe-Chang:
**What type of PR is this?**

/kind cleanup

When the cluster is offline, I think its Ready status condition should be `Unknow`, not `False`.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/status/cluster_status_controller.go
+++ b/pkg/controllers/status/cluster_status_controller.go
@@ -411,7 +411,7 @@ func healthEndpointCheck(client *clientset.Clientset, path string) (int, error) 
 
 func generateReadyCondition(online, healthy bool) metav1.Condition {
 	if !online {
-		return util.NewCondition(clusterv1alpha1.ClusterConditionReady, clusterNotReachableReason, clusterNotReachableMsg, metav1.ConditionFalse)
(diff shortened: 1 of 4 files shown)


jwcesign:
I create a k8s and delete the node(same with offline), the status will be:
```yaml
  - lastHeartbeatTime: "2023-02-16T02:59:04Z"
    lastTransitionTime: "2023-02-16T03:00:20Z"
    message: Kubelet stopped posting node status.
    reason: NodeStatusUnknown
    status: Unknown
    type: Ready
```

So it's reasonable update.

RainbowMango (approved):
/approve

XiShanYongYe-Chang:
/hold
I need to update the related E2E

RainbowMango:
/hold

`Unknown` means Karmada can't decide if a cluster is in the `ready` condition or not. But for this case, Karmada does know that the cluster is `offline`. 

The evidence provided by @jwcesign on the above can't apply to this case, as that is the `controller-manager` didn't receive a report from kubelet for a while.

XiShanYongYe-Chang:
More pieces of evidence are required to prove the difference between offline and not-ready, close it first.
/close

PR #3161, opened 2023-02-16
Add proposal for federatedhpa

jwcesign:
**What type of PR is this?**
/kind documentation
/kind feature

The proposal of FederatedHPA, which meets:
1. Bring HPA from single cluster to multi-clusters.
1. Compatible with the HPA-related resources in a single cluster.
1. Tolerate the disaster of member clusters or Karmada control plane.
1. It is better to integrate well with scenarios such as cloud burst.
1. It is better to support both Kubernetes HPA and customized HPA.
1. Minimize control plane overhead when scaling workloads across clusters, and maximize the utilization of member clusters' scaling capabilities. 
1. Having rich scaling policies, users can enable different scenarios with different configurations.
1. Consider cluster-autoscaler, scaling across clusters could be delayed after some period.

Fixes #none

none

```release-note

```


jwcesign:
Any comments are welcome, just cc guys who might be interested in the feature.

cc @RainbowMango @XiShanYongYe-Chang @Poor12 @chaunceyjiang @YueHonghui @Garrybest 

I'll add the reviewer and approver and put names on the head of the proposal once it is accepted.

docs/proposals/hpa/federated-hpa-v2.md
```diff
@@ -0,0 +1,809 @@
...
+
+    // Status of the current information about the federated HPA.
+    // +optional
+    Status FederatedHPAStatus `json:"status,omitempty"`
+}
+
+type FederatedHPASpec struct {
+    // ScaleTargetRef points to the target resource to scale, and is used to the pods for which metrics
+    // should be collected, as well as to actually change the replica count.
+    ScaleTargetRef autoscalingv2.CrossVersionObjectReference `json:"scaleTargetRef"`
```
chaunceyjiang, line 136: `FederatedHPASpec `has all the fields of `HorizontalPodAutoscalerSpec`.
Why not inline the `HorizontalPodAutoscalerSpec` instead?

jwcesign: Good idea, I will update it. Same with the status spec.

docs/proposals/hpa/federated-hpa-v2.md
```diff
@@ -0,0 +1,809 @@
...
+<!--
+Detail the things that people will be able to do if this KEP is implemented.
+Include as much detail as possible so that people can understand the "how" of
+the system. The goal here is to make this feel real for users without getting
+bogged down.
+-->
+
+#### Migration HPA from single cluster to multi-clusters with minimal efforts
+As a platform developer using Kubernetes, now I want to use Karmada to run apps on multi clusters. But the CD ecosystem is built based on a single cluster and the original HPA is heavily used. So I want to migrate the HPA resources to multi-clusters without too much effort. It is better to be compatible with the schema of HPA used in a single cluster.
+
```
chaunceyjiang, line 63: Is the migration here automatic? 
Automatically convert user's HPA to Federated-HPA.

jwcesign: No, you should do it manually.  But the spec is almost the same, so it will not cost much effort

docs/proposals/hpa/federated-hpa-v2.md
```diff
@@ -0,0 +1,809 @@
...
+    // Conditions is the set of conditions required for this autoscaler to scale its target,
+    // and indicates whether or not those conditions are met.
+    Conditions []autoscalingv2.HorizontalPodAutoscalerCondition `json:"conditions,omitempty"`
+}
+
+```
+
+### Prerequisites
+1. Once the FederatedHPA is applied for the workload, Karmada scheduler should not work for this workload anymore.
+1. Once the FederatedHPA is applied for workload, the replicas changes in member clusters should be retained.
```
Garrybest, line 275: According to this prerequisites, it seems that `rb.spec.clusters` does not take effect. But the policy case below shows that RB is updated by FederatedHPA controller. I'm confused about why we should handle this field `rb.spec.clusters`?

docs/proposals/hpa/federated-hpa-v2.md
```diff
@@ -0,0 +1,809 @@
...
+
+Consider including folks who also work outside the SIG or subproject.
+-->
+
+1. This proposal doesn't consider cluster disaster migration for HPA, which should be done in another proposal.
+1. When assigning the FederatedHPA, we use availableReplicas as the key factor, it may cause inaccurate HPA assignments.
+
+## Design Details
+### Architecture
+![Karmada Federated HPA Architecture](./statics/karmadafederatedhpa-arch-v2.drawio.png)
```
Garrybest, line 110: According to this image, I guess scheduler does not take any effect on Federated HPA workload, and all scaling result is based on the decision made by HPA controller in member clusters, right?

docs/proposals/hpa/federated-hpa-v2.md
jwcesign: To implement the initial assignment, we still need `rb` to generate `works`, which will be used to create/update the workloads in member clusters with specific replicas. So field `rb.spec.clusters` is still used.

docs/proposals/hpa/federated-hpa-v2.md
jwcesign: Correct, the core concept in this proposal is: when a member cluster could scale up the workloads, FederatedHPA controller should not care about it, but when it can't(reached maxReplicas or pending, or other reason), FederatedHPA controller should select one and scale it up.

docs/proposals/hpa/federated-hpa-v2.md
Garrybest: > o implement the initial assignment, we still need rb to generate works

Do you mean we should need the initial result to generate HPA objects in member clusters? After that, do we still need these results?

jwcesign: > Do you mean we should need the initial result to generate HPA objects in member clusters?

For instance, the initial assignment result is:
```
#cluster 1
minReplicas: 3
maxReplicas: 5
#cluster 2
minReplicas: 1
maxReplicas: 2

# the current ResourceBinding is:
- name: cluster1
  replicas: 3
```
So initial assignment needs to update `ResourceBinding` to:
```
- name: cluster1
  replicas: 3
- name: cluster2
  replicas: 1
```
If not, the deployment/replicas in cluster2 will be 0, and couldn't be scaled up. `ResourceBinding` is just like a tool used to manage the initial replicas when FederatedHPA applied for the first time.

Garrybest: Got it.

docs/proposals/hpa/federated-hpa-v2.md
```diff
@@ -0,0 +1,809 @@
...
+
+cluster 3:
+  current replicas: 8
+  maxReplicas: 100
+  current Utilization: 50
+  currentAvailableReplicas: 50
+
+```
+
+So, FederatedHPA controller will directly scale 2 replicas up in cluster 2, and 2 replicas in cluster 3(If there are pending pods, the behavior will be the same).
```
Garrybest, line 605: Now we need to scale up 4 replicas in member1, but it does not have more available resources. So we assign these 4 replicas to member2 and member3 by weight of their currentAvailableReplicas, right?

How does FederatedHPA controller do this scaling? By calling scale subresource of  workload in member clusters directly? Does it mean FederatedHPA controller should watch all member clusters?

docs/proposals/hpa/federated-hpa-v2.md
```diff
@@ -0,0 +1,809 @@
...
+cluster 1:
+  current replicas: 4
+  maxReplicas: 4
+  current Utilization: 100
+
+cluster 2:
+  current replicas: 8
+  maxReplicas: 100
+  current Utilization: 50
+  currentAvailableReplicas: 50
```
Garrybest, line 595: How do we know the currentAvailableReplicas? Will FederatedHPA controller call estimator?

And how does FederatedHPA controller find out that member1 does not have available resource and need to be adjusted by FederatedHPA controller? By checking if there are any pending pods of this workload?

docs/proposals/hpa/federated-hpa-v2.md
jwcesign: > Now we need to scale up 4 replicas in member1, but it does not have more available resources. So we assign these 4 replicas to member2 and member3 by weight of their currentAvailableReplicas, right?

Right.

> By calling scale subresource of workload in member clusters directly?

Yes

> Does it mean FederatedHPA controller should watch all member clusters?

I think FederatedHPA controller could find all the information needed in the work resources.
For updating the scale subresource, we don't need to watch it, just update it directly.

Garrybest: > I think FederatedHPA controller could find all the information needed in the work resources.

I guess By fetching the available replicas and spec replicas of this workload? If not equal, we may need to trigger FederatedHPA controller to adjust replicas in member clusters.

> For updating the scale subresource, we don't need to watch it, just update it directly.

It seems like FederatedHPA controller should also have all kubeconfig files because it needs to update the scale subresources.

docs/proposals/hpa/federated-hpa-v2.md
jwcesign: > How do we know the currentAvailableReplicas? Will FederatedHPA controller call estimator?

In cluster CR, we can find the allocatable replicas. No, it will not call now, but I think we should in the future. `AvailableReplicas` is inaccurate, I describe it in the `Risks and Mitigations`.

> And how does FederatedHPA controller find out that member1 does not have available resource and need to be adjusted by FederatedHPA controller? By checking if there are any pending pods of this workload?

Yes, the pending workload triggers the operation, Or periodic.

Garrybest: > In cluster CR, we can find the allocatable replicas.

Interesting, but this field is quite inaccurate. Workload could contain a node affinity based on which we could not determine `AvailableReplicas` by allocatable replicas.

> Yes, the pending workload triggers the operation, Or maxReplicas reached.

Could you enlighten me how to know if there is pending pods of this workload?

docs/proposals/hpa/federated-hpa-v2.md
jwcesign: > I guess By fetching the available replicas and spec replicas of this workload?

more specifically, the status of FederatedHPA(contains the utilization of CPU), and the status of the deployment. When one cluster reaches the maxReplicas and utilization is above the target, or the unavailable replicas in deployment status(long time pending), FederatedHPA controller should scale in other clusters

> It seems like FederatedHPA controller should also have all kubeconfig files because it needs to update the scale subresources.

Yes

docs/proposals/hpa/federated-hpa-v2.md
jwcesign: > Interesting, but this field is quite inaccurate. Workload could contain a node affinity based on which we could not determine AvailableReplicas by allocatable replicas.

Yes, it's quite inaccurate, we need a better way to solve this. Do you have any good ideas?

> Could you enlighten me how to know if there is pending pods of this workload?

 the unavailable replicas in deployment status(long time pending), could get it from the control plane.

docs/proposals/hpa/federated-hpa-v2.md
jwcesign: HorizontalPodAutoscalerSpec has `minReplicas/maxReplicas`, but in FederatedHPA, it's `federatedMinReplicas/federatedMaxReplicas`, so can't inline `HorizontalPodAutoscalerSpec`,  but the status field can.

docs/proposals/hpa/federated-hpa-v2.md
```diff
@@ -0,0 +1,785 @@
...
+1. This proposal doesn't consider cluster disaster migration for HPA, which should be done in another proposal.
+1. When assigning the FederatedHPA, we use availableReplicas as the key factor, it may cause inaccurate HPA assignments.
+
+## Design Details
+### Architecture
+![Karmada Federated HPA Architecture](./statics/karmadafederatedhpa-arch-v2.drawio.png)
+
+From the above architecture, FederatedHPA controller should be responsible for the following things:
+* Based on the FederatedHPA configuration, assign the minReplicas/maxReplicas of HPA to member clusters by creating works.
+* When HPA initial assignment, the controller should update the workload's ResourceBinding based on the assignment results.
```
chaunceyjiang, line 114: what mean is `assignment results` ?

docs/proposals/hpa/federated-hpa-v2.md
```diff
@@ -0,0 +1,785 @@
...
+    // behavior configures the scaling behavior of the target
+    // in both Up and Down directions (scaleUp and scaleDown fields respectively).
+    // If not set, the default HPAScalingRules for scale up and scale down are used.
+    // +optional
+    Behavior *autoscalingv2.HorizontalPodAutoscalerBehavior `json:"behavior,omitempty"`
+
+    // ClusterAffinity represents scheduling restrictions to a certain set of clusters.
+    // If not set, any cluster can be scheduling candidate.
+    // +optional
+    ClusterAffinity *policyv1alpha1.ClusterAffinity `json:"clusterAffinity,omitempty"`
```
chaunceyjiang, line 157: We are introducing `multiple scheduling group`, will you introduce it here too?

chaunceyjiang, line 157: What does this field mean?

docs/proposals/hpa/federated-hpa-v2.md
```diff
@@ -0,0 +1,785 @@
...
+    replicas: 3
+  - name: member2
+    replicas: 4 # To avoid overloading, the replicas will not be changed.
+  - name: member3
+    replicas: 10
+  - name: member5
+    replicas: 3
+```
+
+We can see member4 is deleted from ResourceBinding, which means once the FederatedHPA is applied, the selection result of Karmada scheduler will be ignored.
```
chaunceyjiang, line 303: If `member5` is filtered out because it does not meet the scheduling conditions, but you can add `member5` directly here, is that expected?

jwcesign: Yes, it's expected. Once HPA is applied, the selection of Karmada scheduler means nothing.

docs/proposals/hpa/federated-hpa-v2.md
jwcesign: Same with PP, FederatedHPA should do HPA propagation based on ClusterAffinity(select part of the clusters).

docs/proposals/hpa/federated-hpa-v2.md
jwcesign: For instance, FederatedHPA's `federatedMinReplicas=4/federatedMaxReplicas=8`, cluster-1's weight is 1, cluster-2's weight is 3, so the assignment result is: cluster-1's HPA `minReplicas=1/maxReplicas=2`,  cluster-2's HPA `minReplicas=3/maxReplicas=6`.

docs/proposals/hpa/federated-hpa-v2.md
Garrybest: What about talking to estimator and get the unschedulable pods of this workload?

docs/proposals/hpa/federated-hpa-v2.md
```diff
@@ -0,0 +1,785 @@
...
+* When HPA initial assignment, the controller should update the workload's ResourceBinding based on the assignment results.
+* When the member cluster could scale up the workload, FederatedHPA controller should not scale up the workload across clusters, but if the member cluster can not, FederatedHPA controller should scale the workload in other member clusters with different policies.
+* If there is a priority policy, the FederatedHPA controller should scale down the low-priority cluster first, if there is no priority policy, the FederatedHPA controller should not care about the scaling down operation.
+* The HPA resource version will be `autoscaling/v2`, because the new fields introduced in autoscaling/v2 are preserved as annotations when working with autoscaling/v1. [Referring](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/)
+
+To isolate from PP/CPP and avoid PP changes that make it difficult for FederatedHPA to evolve, so we introduce FedertedHPA API to implement the scaling across clusters with rich policies.
+```go
+type FedertedHAP struct {
```
jameszhangyukun, line 121: a typo `FedertedHAP` may be `FedertedHPA`

docs/proposals/hpa/federated-hpa-v2.md
jwcesign: But estimator is an optional component. HPA should work ok by default.

docs/proposals/hpa/federated-hpa-v2.md
chaunceyjiang: > Once HPA is applied, the selection of Karmada scheduler means nothing.

If the scheduler schedules workloads to `member1`, `member2` and `member3`, and `FederatedHPA` has only `member3` set, and suddenly `member3` fails, what will happen to the workloads at that time?

docs/proposals/hpa/federated-hpa-v2.md
```diff
@@ -0,0 +1,785 @@
...
+    // +optional
+    ScalingToZero bool `json:"scalingToZero,omitempty"`
+
+	// FederatedHPAAssignment is the assignment policy of FederatedMinReplicas/FederatedMaxReplicas
+	// +required
+    FederatedHPAAssignment FederatedHPAAssignment `json:"federatedHPAAssignment,omitempty"`
+}
+
+// FederatedHPAAssignmentType describes assignment methods for FederatedHPA FederatedMinReplicas/FederatedMaxReplicas
+type FederatedHPAAssignmentType string
```
chaunceyjiang, line 183: I'm curious, is the `assignment` here done by the Karmada scheduler or by the FederatedHP controller?

docs/proposals/hpa/federated-hpa-v2.md
```diff
@@ -0,0 +1,785 @@
...
+If users deploy the workloads in multi-clusters, they may want to scale the workloads in different clusters with different policies. For example, the workloads in the cluster with high priority may need to be scaled first before the clusters with low priority.
+
+### Goals
+1. Bring HPA from single cluster to multi-clusters.
+1. Compatible with the HPA-related resources in a single cluster.
+1. Tolerate the disaster of member clusters or Karmada control plane.
+1. It is better to integrate well with scenarios such as cloud burst.
+1. It is better to support both Kubernetes HPA and customized HPA.
+1. Minimize control plane overhead when scaling workloads across clusters, and maximize the utilization of member clusters' scaling capabilities. 
+1. Having rich scaling policies, users can enable different scenarios with different configurations.
```
chaunceyjiang, line 35: > Having rich scaling policies

I'm curious, in what scenario would a user not use a custom HPA directly? Instead, they want to use the policies in Federated-HPA?

docs/proposals/hpa/federated-hpa-v2.md
chaunceyjiang: On the other hand, does this field implicitly contain the meaning of "scheduling"? Does it make the user wonder why there are two places where a scheduling policy needs to be set at the same time?

jwcesign: > I'm curious, is the assignment here done by the Karmada scheduler or by the FederatedHP controller?

FederatedHPA controller

> On the other hand, does this field implicitly contain the meaning of "scheduling"?

Yes

> Does it make the user wonder why there are two places(PP and FederatedHPA) where a scheduling policy needs to be set at the same time?

Once FederatedHPA is applied, PP is ignored, and the police could be not the same. The reason why we do this is:
1. If we continue to use PP, it will make the FederatedHPA's evolution difficult.
2. PP lacks some policies, and it's not meet the HPA’s requirements fully.

docs/proposals/hpa/federated-hpa-v2.md
jwcesign: For example, scaling up the replicas in the clusters with high priority.

docs/proposals/hpa/federated-hpa-v2.md
chaunceyjiang: Can we implement the scheduling strategy in FederatedHPA by developing a scheduler plugin?


I am worried that implementing scheduling in FederatedHPA will complicate the overall architecture of Karmada.

jwcesign: Yeah, this is what I worry too.

From my point of view, the scheduler only cares about scheduling, but FederatedHPA needs other operations, like creating HPA works and updating scaling gvr.

Or FederatedHPA controller could call the scheduler to do assignments?

jwcesign:
cc @Garrybest @chaunceyjiang @XiShanYongYe-Chang for checking

docs/proposals/hpa/federated-hpa-v2.md
```diff
@@ -0,0 +1,703 @@
...
+
+<!--
+List the specific goals of the KEP. What is it trying to achieve? How will we
+know that this has succeeded?
+-->
+
+1. Bring HPA from single cluster to multi-clusters.
+1. Compatible with the HPA-related resources in a single cluster.
+1. Tolerate the disaster of member clusters or Karmada control plane.
+1. It is better to support both Kubernetes HPA and customized HPA.
```
RainbowMango, line 47: `It is better to`?  yes, or no?

jwcesign: fixed

Garrybest:
No more questions for me, cc @RainbowMango @chaunceyjiang @XiShanYongYe-Chang for checking.

jwcesign:
also cc @YueHonghui, please check when you are available

docs/proposals/hpa/federated-hpa-v2.md
```diff
@@ -0,0 +1,699 @@
...
+    // Spec defines the desired behavior of federated HPA.
+    Spec FederatedHPASpec `json:"spec,omitempty"`
+
+    // Status of the current information about the federated HPA.
+    // +optional
+    Status FederatedHPAStatus `json:"status,omitempty"`
+}
+
+type FederatedHPASpec struct {
+    autoscalingv2.HorizontalPodAutoscalerSpec
```
chaunceyjiang, line 145:
```suggestion
    autoscalingv2.HorizontalPodAutoscalerSpec `json:",inline"`
```

docs/proposals/hpa/federated-hpa-v2.md
chaunceyjiang: Currently, there is a field defined `Placement` in our RB. Can we transfer the configuration related to scheduling strategy in `FederatedHPA` to RB's Placement through `FederatedHPA-controller`?

docs/proposals/hpa/federated-hpa-v2.md
```diff
@@ -0,0 +1,699 @@
...
+
+From the above architecture, FederatedHPA controller should be responsible for the following things:
+* Based on the FederatedHPA configuration, assign the minReplicas/maxReplicas of HPA to member clusters by creating works.
+* When HPA initial assignment, the controller should update the workload's ResourceBinding based on the assignment results.
+* If there is a priority policy, the FederatedHPA controller should scale down the low-priority cluster first, if there is no priority policy, the FederatedHPA controller should not care about the scaling down operation.
+* If the cluster's K8s version under 1.23, HPA resource version will be `autoscaling/v2beta2`, if not, the version will be `autoscaling/v2`. (`autoscaling/v2beta2` is supported after 1.11, `autoscaling/v2` is supported after 1.22)
+
+FederatedHPA has its own characteristics, such as cross-cluster scaling execution delay. To make Federated HPA more flexible and easy to extend, we have introduced the Federated HPA API, which enables the implementation of scaling with a wide range of policies.
+```go
+type FedertedHPA struct {
```
chaunceyjiang, line 132:
```suggestion
type FederatedHPA struct {
```

docs/proposals/hpa/federated-hpa-v2.md
```diff
@@ -0,0 +1,699 @@
...
+## Summary
+
+<!--
+This section is incredibly important for producing high-quality, user-focused
+documentation such as release notes or a development roadmap. 
+
+A good summary is probably at least a paragraph in length.
+-->
+
+Horizontal Pod Autoscaling (HPA) is a popular method for stabilizing applications by adjusting the number of pods based on different metrics. While Karmada simplifies running applications on multiple clusters, it currently lacks HPA support.
```
RainbowMango, line 24: The reason why Karmada is going to introduce FederatedHPA there are some specific use cases that HPA can't handle. (not the `lacks HPA support`).

RainbowMango: It would be better to briefly describe the pain points of using HPA in a single cluster or the limitations in multi-cluster scenarios.

jwcesign: modified, please check again

docs/proposals/hpa/federated-hpa-v2.md
```diff
@@ -0,0 +1,699 @@
...
+This proposal aims to introduce FederatedHPA to Karmada, allowing it to scale workloads across clusters with different policies. Furthermore, the proposal seeks to minimize user experience discrepancies between HPA in a single cluster and multi-cluster environments. With FederatedHPA, users can easily scale their applications across multiple clusters and improve the overall efficiency of their Kubernetes clusters.
+
+## Motivation
+
+<!--
+This section is for explicitly listing the motivation, goals, and non-goals of
+this KEP.  Describe why the change is important and the benefits to users.
+-->
+
+In the current era of single clusters, Horizontal Pod Autoscaling (HPA) is commonly used to dynamically scale workload replicas to handle incoming requests and enhance resource utilization. Therefore, in the era of multi-clusters, HPA should still work and be capable of scaling the workloads in different member clusters, to improved disaster recovery capabilities for the clusters.
```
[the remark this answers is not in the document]
jwcesign: modified, please check again

Issue #3153, opened 2023-02-14
[umbrella] Add multiple scheduling group


XiShanYongYe-Chang:
Refer to #3126 

Task breakdown list:
- [x] Add multiple scheduling group proposal #3126 
- [x] API modification adaptation #3105 #3163 #3172 
- [x] `karmada-controller-manager`: when creating or updating rb/crb, the added OrderedClusterAffinities in pp/cpp should be synced (The adaptation has been modified in #2702 .)
- [x] `karmada-scheduler`: invoke multi times `ScheduleAlgorithm interface` until the schedule succeeds (Adapt #3165 modification, add an input parameter to indicate the affinityName that is being processed.)
    - [x] add binding status in the schedule algorithm interface #3209 
    - [x] modify judgement of placement changed and consider clusteraffinities #3211
    - [x] invoke multi time with schedule algorithm #3181 
    - [x] update event handler to consider clusteraffinities #3181 
- [x] `karmada-webhook`: New field validation #3164 
- [x] E2E
  - [x] Workload propagating with scheduling type Duplicated #3219 
  - [ ] Workload propagating with scheduling type Divided
  - [x] Failover scenario #3219

RainbowMango:
Please consider if we need to update the logic of policy changed or not at:
https://github.com/karmada-io/karmada/blob/e523977cb0b8eb0367dd7bfb9ac5820c235f2a47/pkg/scheduler/scheduler.go#L368-L374

and 
https://github.com/karmada-io/karmada/blob/e523977cb0b8eb0367dd7bfb9ac5820c235f2a47/pkg/scheduler/scheduler.go#L430-L435

XiShanYongYe-Chang:
We have completed the development and testing of this feature,
/close

RainbowMango:
> Workload propagating with scheduling type Divided

How to track this issue?

XiShanYongYe-Chang:
> > Workload propagating with scheduling type Divided
> 
> How to track this issue?

Tracked by #3223

PR #3223, opened 2023-02-28
[WIP] Add e2e for schedule with clusterAffinities when replica scheduling type is Divided

XiShanYongYe-Chang:
**What type of PR is this?**

/kind feature

Part of https://github.com/karmada-io/karmada/issues/3153

When the cluster label changes, resources whose scheduling policy is `Divided` cannot be scheduled again.

To solve this problem, we need to update the triggering conditions of the scheduler.

```release-note
NONE
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -112,7 +112,7 @@ jobs:
         # Here support the latest three minor releases of Kubernetes, this can be considered to be roughly
         # the same as the End of Life of the Kubernetes release: https://kubernetes.io/releases/
         # Please remember to update the CI Schedule Workflow when we add a new version.
-        k8s: [ v1.24.2, v1.25.0, v1.26.0 ]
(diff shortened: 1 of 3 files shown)


XiShanYongYe-Chang:
> Really? Can it be consistently reproduced?

If the problem can be reproduced, use the resources in the e2e to perform the test.

chaunceyjiang:
This e2e I feel very confused, modify the cluster label also to trigger the rescheduler ?

XiShanYongYe-Chang:
> This e2e I feel very confused, modify the cluster label also to trigger the rescheduled ?

Yes, when the label of the cluster changes, rescheduling also needs to be triggered, because we can use the label to select the cluster in pp/cpp.

XiShanYongYe-Chang:
Since there has been no errors on this case for a long time, close this pr first.
/close

PR #3229, opened 2023-03-01
add unit tests for dependency interpretation

whitewindmills:
**What type of PR is this?**
/kind failing-test

add unit tests for dependency interpretation
Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/resourceinterpreter/defaultinterpreter/dependencies_test.go
+++ b/pkg/resourceinterpreter/defaultinterpreter/dependencies_test.go
@@ -11,22 +11,193 @@ import (
 	"github.com/karmada-io/karmada/test/helper"
 )
 
+var (
+	namespace = "karmada-test"
+	podName   = "fake-pod"
+)
+
 func TestGetSecretNames(t *testing.T) {
-	fakePod := helper.NewPod("foo", "bar")
-	fakePod.Spec.Volumes = []corev1.Volume{
+	emptySecretsPod := helper.NewPod(namespace, podName)
+
+	secretsPod := helper.NewPod(namespace, podName)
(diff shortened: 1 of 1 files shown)


RainbowMango:
The failing test is due to unstable CodeCov services, I'll re-trigger that.

whitewindmills:
@RainbowMango Hi, can you approve it?

RainbowMango:
/assign @XiShanYongYe-Chang 
Please help to take a look~

pkg/resourceinterpreter/defaultinterpreter/dependencies_test.go
```diff
@@ -205,6 +503,9 @@ func TestGetDependenciesFromPodTemplate(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
 			res, _ := getDependenciesFromPodTemplate(tt.pod)
+			if len(res) == 0 {
+				res = nil
+			}
```
XiShanYongYe-Chang, line 508: Why add this judgment? I remove this check it runs well.

XiShanYongYe-Chang (commented):
Thanks~, the other looks good to me.

pkg/resourceinterpreter/defaultinterpreter/dependencies_test.go
whitewindmills: > Why add this judgment? I remove this check it runs well.

Because the func `reflect.DeepEqual` will return false if nil and empty slice compare. I do not know `res` will be empty slice or nil in the future, but I'm sure int the test case they should be all equal.

XiShanYongYe-Chang: ok
/lgtm
/approve

PR #3232, opened 2023-03-02
fix bug that spreadconstraints are not validated correctly

### LINKED ISSUE #3231: pp/cpp's spreadconstraints are not validated correctly
**What happened**:
Error pp/cpp's spreadconstraints can be create and affect the scheduling process.
0880c4c80710fad0c20c

0880c4c80710fad0a20c

`minGroups` and `maxGroup` should not be less that zero, and `maxGroup` should also get default value(I think it should be `minGroups`)
Create pp or cpp with incorrect `minGroups` and `maxGroup`.

**Environment**:
- Karmada version:
- Others:

whitewindmills:
**What type of PR is this?**
/kind feature
/kind cleanup

fix bug that spreadconstraints are not validated correctly
Fixes #3231

```release-note
`karmada-webhook`: Introduced validation to ensure the `.spec.placement.spreadConstraints.maxGroups/minGroups` in PropagationPolicy is declared with a reasonable value.
```

### DIFF
--- a/pkg/util/validation/validation.go
+++ b/pkg/util/validation/validation.go
@@ -102,6 +102,16 @@ func ValidateSpreadConstraint(spreadConstraints []policyv1alpha1.SpreadConstrain
 			allErrs = append(allErrs, field.Invalid(fldPath.Index(index), constraint, "spreadByLabel should not co-exist with spreadByField"))
 		}
 
+		// If MinGroups provided, it should not be lower than 0.
+		if constraint.MinGroups < 0 {
+			allErrs = append(allErrs, field.Invalid(fldPath.Index(index), constraint, "minGroups lower than 0 is not allowed"))
+		}
+
+		// If MaxGroups provided, it should not be lower than 0.
+		if constraint.MaxGroups < 0 {
+			allErrs = append(allErrs, field.Invalid(fldPath.Index(index), constraint, "maxGroups lower than 0 is not allowed"))
+		}
+
 		// If MaxGroups provided, it should greater or equal than MinGroups.
 		if constraint.MaxGroups > 0 && constraint.MaxGroups < constraint.MinGroups {
 			allErrs = append(allErrs, field.Invalid(fldPath.Index(index), constraint, "maxGroups lower than minGroups is not allowed"))
--- a/pkg/util/validation/validation_test.go
+++ b/pkg/util/validation/validation_test.go
@@ -444,6 +444,32 @@ func TestValidatePropagationSpec(t *testing.T) {
 			expectedErr: "maxGroups lower than minGroups is not allowed",
 		},
 		{
+			name: "spreadConstraint maxGroups lower than 0",
+			spec: policyv1alpha1.PropagationSpec{
+				Placement: policyv1alpha1.Placement{
+					SpreadConstraints: []policyv1alpha1.SpreadConstraint{
+						{
+							MaxGroups: -1,
+							MinGroups: 1,
+						},
+					},
+				}},
+			expectedErr: "maxGroups lower than 0 is not allowed",
+		},
+		{
+			name: "spreadConstraint minGroups lower than 0",
+			spec: policyv1alpha1.PropagationSpec{
+				Placement: policyv1alpha1.Placement{
+					SpreadConstraints: []policyv1alpha1.SpreadConstraint{
+						{
+							MaxGroups: 2,
+							MinGroups: -2,
+						},
+					},
+				}},
+			expectedErr: "minGroups lower than 0 is not allowed",
+		},
+		{
 			name: "spreadConstraint spreadByFieldCluster must be included if using spreadByField",
 			spec: policyv1alpha1.PropagationSpec{
 				Placement: policyv1alpha1.Placement{


pkg/util/validation/validation.go
```diff
@@ -102,6 +102,16 @@ func ValidateSpreadConstraint(spreadConstraints []policyv1alpha1.SpreadConstrain
...
+		// If MinGroups provided, it should not be lower than 0.
+		if constraint.MinGroups < 0 {
+			allErrs = append(allErrs, field.Invalid(fldPath.Index(index), constraint, "minGroups lower than 0 is not allowed"))
+		}
+
+		// If MaxGroups provided, it should not be lower than 0.
+		if constraint.MaxGroups < 0 {
+			allErrs = append(allErrs, field.Invalid(fldPath.Index(index), constraint, "maxGroups lower than 0 is not allowed"))
+		}
+
```
XiShanYongYe-Chang, line 114: Do we need to check that `MaxGroups` cannot be less than `MinGroups`?

XiShanYongYe-Chang (commented):
Perhaps we need to specify in the API declaration as well:

https://github.com/karmada-io/karmada/blob/0652cfbe2c87946e2c6ca8ab3a6d5f22ff9a3740/pkg/apis/policy/v1alpha1/propagation_types.go#L233-L240

pkg/util/helper/policy.go
```diff
@@ -25,6 +25,11 @@ func SetDefaultSpreadConstraints(spreadConstraints []policyv1alpha1.SpreadConstr
 			klog.Infof("Setting default MinGroups to 1")
 			spreadConstraints[i].MinGroups = 1
 		}
+
+		if spreadConstraints[i].MaxGroups == 0 && spreadConstraints[i].MinGroups > 0 {
```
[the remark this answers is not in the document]
whitewindmills: > 

If minGroups is less than zero? This part of the judgment is necessary.

whitewindmills:
> Perhaps we need to specify in the API declaration as well:
> 
> https://github.com/karmada-io/karmada/blob/0652cfbe2c87946e2c6ca8ab3a6d5f22ff9a3740/pkg/apis/policy/v1alpha1/propagation_types.go#L233-L240

You're right.

pkg/util/helper/policy.go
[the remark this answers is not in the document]
XiShanYongYe-Chang: Isn't this a situation that has been intercepted by validate?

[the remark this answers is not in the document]
whitewindmills: > intercepted

No, mutating webhook works first!

pkg/util/validation/validation.go
XiShanYongYe-Chang: Sorry, I missed it. There's already a judgment in the logic.

pkg/util/helper/policy.go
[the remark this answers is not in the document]
chaunceyjiang: I recommend using `+kubebuilder:validation:Minimum=1`.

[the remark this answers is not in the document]
XiShanYongYe-Chang: > No, mutating webhook works first!

Ok

> I recommend using `+kubebuilder:validation:Minimum=1`.

Does this conflict with `// +optional`?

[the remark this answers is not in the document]
whitewindmills: > I recommend using `+kubebuilder:validation:Minimum=1` and `+kubebuilder:default=1`.

I don't see any difference in performance, because the webhook will always be called, but this part of the verification is done by the apiserver.

whitewindmills:
Sorry, I accidentally clicked the close button.

pkg/util/validation/validation.go
jwcesign: `MaxGroups` could equal 0? If so, it means what? i.e., HPA minReplicas could only be >=1. I also think `MinGroups` should >=1

XiShanYongYe-Chang:
> // Defaults to MinGroups.

IMOP, it will be enough.

pkg/util/validation/validation.go
whitewindmills: No, after mutating, `MaxGroups` could be more than 0 or less than 0, never be 0.

pkg/apis/policy/v1alpha1/propagation_types.go
```diff
@@ -231,6 +231,7 @@ type SpreadConstraint struct {
 	SpreadByLabel string `json:"spreadByLabel,omitempty"`
 
 	// MaxGroups restricts the maximum number of cluster groups to be selected.
+	// Defaults to MinGroups.
```
RainbowMango, line 234: I don't think the `MaxGroups` deserves a default value.
In general, the default value should be a common one that meets most of the use cases.
Here, the valid range of `MinGroups` is [1, max(int)], in order to let the scheduler works out(or, try to) a result we set the defaults to `1`, it seems reasonable, but for `MaxGroups` which is essentially a limitation to the scheduler that can't spread the group that much, we don't have enough evidence to show that users always set `MaxGroups` with the same value as `MinGroups`.

RainbowMango (commented):
The validation part sounds good to me. And I believe this PR is more like focus on the validation instead of the default value:)

pkg/apis/policy/v1alpha1/propagation_types.go
whitewindmills: @RainbowMango Hi, thanks for your replies!
I agree with most of your thoughts, but `MaxGroups` would be zero which is an incomprehensible value for ` karmada-scheduler` without a default value. In other words, how should we(or ` karmada-scheduler` does) understand `MaxGroups` without a default value?

RainbowMango: > how should we(or  karmada-scheduler does) understand MaxGroups without a default value?

It perhaps can't happen because there is a validation to ensure `MaxGroups` can't be lower than `MinGroups`(1).
right?

whitewindmills:
@RainbowMango I don't know if you have any other suggestions, if not, I think this PR can be closed.

RainbowMango:
I'm still thinking if the `MaxGroups` should be set with a default value.
Can we focus on the validation part for this PR?

whitewindmills:
> I'm still thinking if the `MaxGroups` should be set with a default value.
> Can we focus on the validation part for this PR?

Okay, I'll only retain the validation part for this PR due to the difference of opinion for `MaxGroups`.

RainbowMango:
OK, thanks.

RainbowMango (approved):
/lgtm
/approve

Thanks.

PR #3237, opened 2023-03-05
karmadactl: add image pull secret flags for karmadactl init

my-git9:
**What type of PR is this?**

/kind feature

In a production environment, registry  are generally non-public. If the registry we get from the administrator is private, we need to configure the imagepullsecret in the application's yaml file.
So I think it would be better if `karmadactl init` could specify imagepullsecret.

Fixes #

```release-note
karmadactl: Add image pull secret flags for `karmadactl init`
```

### DIFF
--- a/pkg/karmadactl/cmdinit/cmdinit.go
+++ b/pkg/karmadactl/cmdinit/cmdinit.go
@@ -96,6 +96,7 @@ func NewCmdInit(parentCommand string) *cobra.Command {
 	}
 	flags := cmd.Flags()
 	flags.StringVarP(&opts.ImageRegistry, "private-image-registry", "", "", "Private image registry where pull images from. If set, all required images will be downloaded from it, it would be useful in offline installation scenarios.  In addition, you still can use --kube-image-registry to specify the registry for Kubernetes's images.")
+	flags.StringSliceVar(&opts.PullSecrets, "image-pull-secrets", nil, "Image pull secrets are used to pull images from the private registry, could be secret list separated by comma (e.g '--image-pull-secrets PullSecret1,PullSecret2', the secrets should be pre-settled in the namespace declared by '--namespace')")
 	// kube image registry
 	flags.StringVarP(&opts.KubeImageMirrorCountry, "kube-image-mirror-country", "", "", "Country code of the kube image registry to be used. For Chinese mainland users, set it to cn")
 	flags.StringVarP(&opts.KubeImageRegistry, "kube-image-registry", "", "", "Kube image registry. For Chinese mainland users, you may use local gcr.io mirrors such as registry.cn-hangzhou.aliyuncs.com/google_containers to override default kube image registry")
--- a/pkg/karmadactl/cmdinit/kubernetes/deploy.go
+++ b/pkg/karmadactl/cmdinit/kubernetes/deploy.go
@@ -121,6 +121,7 @@ type CommandInitOption struct {
 	CRDs                               string
 	ExternalIP                         string
 	ExternalDNS                        string
+	PullSecrets                        []string
 	CertValidity                       time.Duration
 	KubeClientSet                      kubernetes.Interface
 	CertAndKeyFileData                 map[string][]byte
@@ -611,6 +612,18 @@ func (i *CommandInitOption) karmadaAggregatedAPIServerImage() string {
 		return i.ImageRegistry + "/karmada-aggregated-apiserver:" + karmadaRelease
 	}
 	return i.KarmadaAggregatedAPIServerImage
+}
+
+// get image pull secret
+func (i *CommandInitOption) getImagePullSecrets() []corev1.LocalObjectReference {
+	var imagePullSecrets []corev1.LocalObjectReference
+	for _, val := range i.PullSecrets {
+		secret := corev1.LocalObjectReference{
+			Name: val,
+		}
+		imagePullSecrets = append(imagePullSecrets, secret)
+	}
+	return imagePullSecrets
 }
 
 func generateServerURL(serverIP string, nodePort int32) (string, error) {
--- a/pkg/karmadactl/cmdinit/kubernetes/deployments.go
+++ b/pkg/karmadactl/cmdinit/kubernetes/deployments.go
@@ -138,6 +138,7 @@ func (i *CommandInitOption) makeKarmadaAPIServerDeployment() *appsv1.Deployment 
 	}
 
 	podSpec := corev1.PodSpec{
+		ImagePullSecrets: i.getImagePullSecrets(),
 		Affinity: &corev1.Affinity{
 			PodAntiAffinity: &corev1.PodAntiAffinity{
 				RequiredDuringSchedulingIgnoredDuringExecution: []corev1.PodAffinityTerm{
@@ -251,6 +252,7 @@ func (i *CommandInitOption) makeKarmadaKubeControllerManagerDeployment() *appsv1
 	}
 
 	podSpec := corev1.PodSpec{
+		ImagePullSecrets: i.getImagePullSecrets(),
 		Affinity: &corev1.Affinity{
 			PodAntiAffinity: &corev1.PodAntiAffinity{
 				RequiredDuringSchedulingIgnoredDuringExecution: []corev1.PodAffinityTerm{
@@ -396,6 +398,7 @@ func (i *CommandInitOption) makeKarmadaSchedulerDeployment() *appsv1.Deployment 
 	}
 
 	podSpec := corev1.PodSpec{
+		ImagePullSecrets: i.getImagePullSecrets(),
 		Affinity: &corev1.Affinity{
 			PodAntiAffinity: &corev1.PodAntiAffinity{
 				RequiredDuringSchedulingIgnoredDuringExecution: []corev1.PodAffinityTerm{
@@ -510,6 +513,7 @@ func (i *CommandInitOption) makeKarmadaControllerManagerDeployment() *appsv1.Dep
 	}
 
 	podSpec := corev1.PodSpec{
+		ImagePullSecrets: i.getImagePullSecrets(),
 		Affinity: &corev1.Affinity{
 			PodAntiAffinity: &corev1.PodAntiAffinity{
 				RequiredDuringSchedulingIgnoredDuringExecution: []corev1.PodAffinityTerm{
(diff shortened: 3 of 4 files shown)


RainbowMango:
Hi @my-git9 Can you explain more about why we need this flag in the PR description?

my-git9:
> description

The description has been updated

my-git9:
> Hi @my-git9 Can you explain more about why we need this flag in the PR description?

The description has been updated, thanks

RainbowMango:
I'd like to invite @lonelyCZ for comments.

lonelyCZ:
> If both `--private-image-registry` and `--kube-image-registry` are specified and from different registry, we can specify two imagepullsecret. What you think?

I think it seems rare for using `--private-image-registry` and `--kube-image-registry` togather. 

`--kube-image-registry` is mainly  designed  for solving that some people can't access `k8s.gcr.io` registry.

We should only consider a common case, otherwise it even need to set different `pull secret` for each image. So I this it only set one flag `pull secrets` for all components. In my opinion, if the registry need secret, it will found a matched secret in `ImagePullSecrets`, otherwise it don't need a secret even if already set this field.

pkg/karmadactl/cmdinit/cmdinit.go
```diff
@@ -96,6 +96,7 @@ func NewCmdInit(parentCommand string) *cobra.Command {
 	}
 	flags := cmd.Flags()
 	flags.StringVarP(&opts.ImageRegistry, "private-image-registry", "", "", "Private image registry where pull images from. If set, all required images will be downloaded from it, it would be useful in offline installation scenarios.  In addition, you still can use --kube-image-registry to specify the registry for Kubernetes's images.")
+	flags.StringVar(&opts.PullSecrets, "pull-secrets", "", "Pull Secrets are used to pull images from the private registry, could be secret list separated by comma ('--pull-secrets PullSecret1,PullSecret2')")
```
lonelyCZ, line 99:
```suggestion
	flags.StringSliceVar(&opts.PullSecrets, "pull-secrets", "", "Pull Secrets are used to pull images from the private registry, could be secret list separated by comma ('--pull-secrets PullSecret1,PullSecret2')")
```

pkg/karmadactl/cmdinit/kubernetes/deploy.go
```diff
@@ -121,6 +121,8 @@ type CommandInitOption struct {
 	CRDs                               string
 	ExternalIP                         string
 	ExternalDNS                        string
+	PullSecrets                        string
```
lonelyCZ, line 124:
```suggestion
	PullSecrets                        []string
```

pkg/karmadactl/cmdinit/kubernetes/deploy.go
```diff
@@ -121,6 +121,8 @@ type CommandInitOption struct {
...
+	imagePullSecrets                   []corev1.LocalObjectReference
```
lonelyCZ, line 125: Suggest to move private common variates to end. like

```
// CommandInitOption holds all flags options for init.
type CommandInitOption struct {
	ImageRegistry                      string
	KubeImageRegistry                  string
	KubeImageMirrorCountry             string
	KubeImageTag                       string
        ......
        CertAndKeyFileData                 map[string][]byte
	RestConfig                         *rest.Config
	KarmadaAPIServerIP                 []net.IP

        imagePullSecrets                   []corev1.LocalObjectReference
}
```

RainbowMango:
> In my opinion, if the registry need secret, it will found a matched secret in ImagePullSecrets, otherwise it don't need a secret even if already set this field.

Do you mean the kubelet knows if the specified registry requires a secret? For public registry, the `ImagePullSecrets` will not be used even it is set, right?

jwcesign:
> Do you mean the kubelet knows if the specified registry requires a secret? For public registry, the ImagePullSecrets will not be used even it is set, right?  

Kubelet will try to select the secret by matching the registry. If not match, it will try without secrets.

https://github.com/kubernetes/kubernetes/blob/a55bd631728590045b51a4f65bba31aed1415571/pkg/kubelet/kuberuntime/kuberuntime_image.go#L45

lonelyCZ:
> Do you mean the kubelet knows if the specified registry requires a secret? For public registry, the ImagePullSecrets will not be used even it is set, right?

Yes, I just tried it that was ok.

RainbowMango:
Thanks, I get it. Then, we don't need two flags for the pull secrets.

pkg/karmadactl/cmdinit/cmdinit.go
RainbowMango: Should we use a more clear flag, like `--image-pull-secrets`?

pkg/karmadactl/cmdinit/kubernetes/deploy.go
```diff
@@ -613,6 +615,19 @@ func (i *CommandInitOption) karmadaAggregatedAPIServerImage() string {
 	return i.KarmadaAggregatedAPIServerImage
 }
 
+// get image pull secret
+func (i *CommandInitOption) getImagepullsecret() []corev1.LocalObjectReference {
```
RainbowMango, line 619:
```suggestion
func (i *CommandInitOption) getImagepullsecrets() []corev1.LocalObjectReference {
```

pkg/karmadactl/cmdinit/cmdinit.go
```diff
@@ -96,6 +96,7 @@ func NewCmdInit(parentCommand string) *cobra.Command {
 	}
 	flags := cmd.Flags()
 	flags.StringVarP(&opts.ImageRegistry, "private-image-registry", "", "", "Private image registry where pull images from. If set, all required images will be downloaded from it, it would be useful in offline installation scenarios.  In addition, you still can use --kube-image-registry to specify the registry for Kubernetes's images.")
+	flags.StringSliceVar(&opts.PullSecrets, "image-pull-secrets", nil, "Image pull secrets are used to pull images from the private registry, could be secret list separated by comma ('--image-pull-secrets PullSecret1,PullSecret2')")
```
RainbowMango, line 99: Here only specifies the secrets name, which namespace they should be in?

RainbowMango (commented):
Please share a test report here.
I wonder how to use it, especially how to prepare the secrets.

my-git9:
If we want to use this parameter, we should have the following steps:
1. Create a namespace in advance, such as `karmada-system`
2. Create the required pullsecret in the namespace in advance, such as `pullsecret1`
3. Execute the installation command: `karmactl init --image-pull-secrets=pullsecret1`

The result:
```
kubectl -nkarmada-system get po -oyaml |grep -A 1 imagePullSecrets
    imagePullSecrets:
    - name: pullsecret1
--
    imagePullSecrets:
    - name: pullsecret1
```
@RainbowMango

RainbowMango:
OK, I get it. I guess we can explain it in the flag usage.

my-git9:
> OK, I get it. I guess we can explain it in the flag usage.

Updated, thanks

pkg/karmadactl/cmdinit/cmdinit.go
```diff
@@ -96,6 +96,7 @@ func NewCmdInit(parentCommand string) *cobra.Command {
 	}
 	flags := cmd.Flags()
 	flags.StringVarP(&opts.ImageRegistry, "private-image-registry", "", "", "Private image registry where pull images from. If set, all required images will be downloaded from it, it would be useful in offline installation scenarios.  In addition, you still can use --kube-image-registry to specify the registry for Kubernetes's images.")
+	flags.StringSliceVar(&opts.PullSecrets, "image-pull-secrets", nil, "Image pull secrets are used to pull images from the private registry, could be secret list separated by comma (e.g '--image-pull-secrets PullSecret1,PullSecret2', the image pull secrets and their namespace need to be created in advance)")
```
RainbowMango, line 99:
```suggestion
	flags.StringSliceVar(&opts.PullSecrets, "image-pull-secrets", nil, "Image pull secrets are used to pull images from the private registry, could be secret list separated by comma (e.g '--image-pull-secrets PullSecret1,PullSecret2', the secrets should be pre-settled in the namespace declared by '--namespace')")
```
How about this?

RainbowMango (commented):
Generally looks good to me.

Please @lonelyCZ take a look.
And consider if we need to add validation for checking the existence of the secret.

pkg/karmadactl/cmdinit/kubernetes/deploy.go
```diff
@@ -613,6 +615,19 @@ func (i *CommandInitOption) karmadaAggregatedAPIServerImage() string {
 	return i.KarmadaAggregatedAPIServerImage
 }
 
+// get image pull secret
+func (i *CommandInitOption) getImagepullsecrets() []corev1.LocalObjectReference {
```
lonelyCZ, line 619:
```suggestion
func (i *CommandInitOption) getImagePullSecrets() []corev1.LocalObjectReference {
```
It seems clearer for camel case.

pkg/karmadactl/cmdinit/kubernetes/deploy.go
```diff
@@ -121,11 +121,13 @@ type CommandInitOption struct {
 	CRDs                               string
 	ExternalIP                         string
 	ExternalDNS                        string
+	PullSecrets                        []string
 	CertValidity                       time.Duration
 	KubeClientSet                      kubernetes.Interface
 	CertAndKeyFileData                 map[string][]byte
 	RestConfig                         *rest.Config
 	KarmadaAPIServerIP                 []net.IP
+	imagePullSecrets                   []corev1.LocalObjectReference
```
RainbowMango, line 130: Sorry I didn't notice this variable.

I think we don't need the `imagePullSecrets` as it's essentially an `internal variable` of `getImagePullSecrets`.

pkg/karmadactl/cmdinit/kubernetes/deploy.go
```diff
@@ -613,6 +614,20 @@ func (i *CommandInitOption) karmadaAggregatedAPIServerImage() string {
 	return i.KarmadaAggregatedAPIServerImage
 }
 
+// get image pull secret
+func (i *CommandInitOption) getImagePullSecrets() []corev1.LocalObjectReference {
+	var imagePullSecrets []corev1.LocalObjectReference
+	if i.PullSecrets != nil {
```
RainbowMango, line 620: This check is actually redundant because the `for .. range` can work with nil slice.

RainbowMango (commented):
Generally looks good to me now.
/assign @lonelyCZ

RainbowMango (commented):
/lgtm
/assign @lonelyCZ 

Thanks.

lonelyCZ:
I just tested it in my env that worked fine!

/approve

PR #3239, opened 2023-03-06
merge labels when building RB/CRB by ClusterPropagationPolicy

### LINKED ISSUE #3238: When Modifying (Cluster)ResourceBinding, Labels Was Merged Only When Binding with PropagationPolicy
**What happened**:  
When the `resource-detector` modify the `(cluster)resourcebinding`, whose labels would be merged only when binding with `propagationpolicy`:   

https://github.com/karmada-io/karmada/blob/master/pkg/detector/detector.go#L407  

but not when binding with `clusterpropagationpolicy`:  

https://github.com/karmada-io/karmada/blob/master/pkg/detector/detector.go#L482  

The labels should be merged no matter which kind of propagationpolicy is binding with.

1. Build a `clusterpropagationpolicy` and a `propagationpolicy` for different workload, then apply them.  
2. Add a test label in both bindings of the two workload manually.
3. Scale the two workload and get their bindings, you will see the test label in bindings with `propagationpolicy` is still exists, while which in bindings with `clusterpropagationpolicy` has been already removed.

I saw some changes in commit history, which seems that the method to modify the labels in bindings with `propagationpolicy` was changed from replacing to merging for adapting dependencies propagation. But in the later commits, I saw only some changes to modify the field `.spec.propagateDeps`, but no more changes to modify the labels. I don't know exactly if it is quite a bug or it is reasonable. If you have some other considering, please tell me.

<strike>

By the way, I found that the modifing of `propagateDeps` was missed in `clusterresourcebinding`:   
https://github.com/karmada-io/karmada/blob/master/pkg/detector/detector.go#L512-#L533

was it also a bug? If it was, I could add a commit to fix it. :)

</strike>

**Environment**:
- Karmada version:
- Others:

lxtywypc:
**What type of PR is this?**

  
/kind bug  

Merge labels in (cluster)resourcebinding when the binding is modifying no matter which kind of propagationpolicy is bind with.

Fixes #3238 

```release-note
`karmada-controller-manager`: Fixed the issue that RB/CRB labels are not merged when syncing new changes. 
```

### DIFF
--- a/pkg/detector/detector.go
+++ b/pkg/detector/detector.go
@@ -479,7 +479,7 @@ func (d *ResourceDetector) ApplyClusterPolicy(object *unstructured.Unstructured,
 						"try again later after binding is garbage collected, see https://github.com/karmada-io/karmada/issues/2090")
 				}
 				// Just update necessary fields, especially avoid modifying Spec.Clusters which is scheduling result, if already exists.
-				bindingCopy.Labels = binding.Labels
+				bindingCopy.Labels = util.DedupeAndMergeLabels(bindingCopy.Labels, binding.Labels)
 				bindingCopy.OwnerReferences = binding.OwnerReferences
 				bindingCopy.Finalizers = binding.Finalizers
 				bindingCopy.Spec.Resource = binding.Spec.Resource
@@ -523,7 +523,7 @@ func (d *ResourceDetector) ApplyClusterPolicy(object *unstructured.Unstructured,
 					"try again later after binding is garbage collected, see https://github.com/karmada-io/karmada/issues/2090")
 			}
 			// Just update necessary fields, especially avoid modifying Spec.Clusters which is scheduling result, if already exists.
-			bindingCopy.Labels = binding.Labels
+			bindingCopy.Labels = util.DedupeAndMergeLabels(bindingCopy.Labels, binding.Labels)
 			bindingCopy.OwnerReferences = binding.OwnerReferences
 			bindingCopy.Finalizers = binding.Finalizers
 			bindingCopy.Spec.Resource = binding.Spec.Resource


pkg/detector/detector.go
```diff
@@ -523,7 +523,7 @@ func (d *ResourceDetector) ApplyClusterPolicy(object *unstructured.Unstructured,
 					"try again later after binding is garbage collected, see https://github.com/karmada-io/karmada/issues/2090")
 			}
 			// Just update necessary fields, especially avoid modifying Spec.Clusters which is scheduling result, if already exists.
-			bindingCopy.Labels = binding.Labels
+			bindingCopy.Labels = util.DedupeAndMergeLabels(bindingCopy.Labels, binding.Labels)
```
XiShanYongYe-Chang, line 526: Maybe we don't need to modify this line, because we don't currently support cluster scope resources for the dependency distribution feature.

lxtywypc: In my opinion, this is not just for the dependency distribution feature, but customized labels as well.  
For example, if we have a webhook which would patch some custmized info into the labels of resourcebindings to group them, after I scale the workload once, the info would be lost.  
This might not be a good example, but I think even if it is a label for temporary usage, we shouldn't abandon them.

XiShanYongYe-Chang: Do you think we should add some E2E to protect this case?

lxtywypc: Humm......I think it is nice to have.  
I hope to help add some.  
May I commit them in this PR? Or open a new one?

RainbowMango (commented):
/assign

RainbowMango:
/retitle merge labels when building RB/CRB by ClusterPropagationPolicy

RainbowMango:
Hi @XiShanYongYe-Chang 
Could you please open another issue to track the E2E tasks? Given @lxtywypc is a first-time contributor who might feel hard to involve the E2E, I'd like to move this forward because seems no much risk here.

RainbowMango:
@lxtywypc Is there anything bad that happened due to this issue?  After going through the analysis in #3238, I think this is more like an improvement in stead of a bug, which means we don't need to cherry-pick it to previous releases.

pkg/detector/detector.go
RainbowMango: Sounds good to me.

XiShanYongYe-Chang: Thanks~

XiShanYongYe-Chang:
> Hi @XiShanYongYe-Chang
> Could you please open another issue to track the E2E tasks? Given @lxtywypc is a first-time contributor who might feel hard to involve the E2E, I'd like to move this forward because seems no much risk here.

Done, refer to #3265 

Hi @lxtywypc, you can sign it to you by the command `/assign` and send a new pr to do it, thanks~

RainbowMango (approved):
/lgtm
/approve

PR #3240, opened 2023-03-06
Add controllers config for karmada-controller-manager

Poor12:
**What type of PR is this?**
/kind cleanup

Add controllers config to karmada-controller-manager.
linked: #2542 
Fixes #2542

```
chart: introduce `controllers` config for karmada-controller-manager
```


charts/karmada/README.md
```diff
@@ -174,195 +186,199 @@ helm install karmada-scheduler-estimator -n karmada-system ./charts/karmada
...
-| `search.certs`                            | Certs of the search                                                                                                                                                                                                                           | `karmada-cert`                                                                                                                                            |
-| `search.kubeconfig`                       | Kubeconfig of the search                                                                                                                                                                                                                      | `karmada-kubeconfig`                                                                                                                                       |
+
```
Poor12, line 197: Formatted the README file.

RainbowMango:
Please link to the previous PR so that people can get the background as well as the discussions there.

Poor12:
> Please link to the previous PR so that people can get the background as well as the discussions there.

Updated.

charts/karmada/values.yaml
```diff
@@ -296,6 +296,7 @@ controllerManager:
   featureGates:
     ## @param PropagateDeps is a feature gate for the controllerManager to allow propagate dependent respurce to workloads.
     PropagateDeps: false
+  controllers: ""
```
calvin0327, line 299: Use a array to define the controllers feild?  like: `controllers: [ "a", "b" ]`

Poor12: But for the value of `--controllers` flag, it's a string.

RainbowMango:
@calvin0327 Do you have any other comments?

charts/karmada/values.yaml
calvin0327: We can spport a template to parse the array as a string. I don't know whether it is necessary to do it

Poor12: For the users, there is no significant improvement in using arrays. I think it's fine to keep the status quo.

RainbowMango: @Poor12 Can you share the test result here? I want to know how to use this flag.

and @calvin0327 can you give an example for the array way? Let's see which one is better.

Poor12: For example:
If you want to disable hpa controller:
```
featureGates:
    ## @param PropagateDeps is a feature gate for the controllerManager to allow propagate dependent respurce to workloads.
    PropagateDeps: false
  controllers: "*,-hpa"
```

Then use `helm template` to render yamls:
```
 command:
            - /bin/karmada-controller-manager
            - --kubeconfig=/etc/kubeconfig
            - --bind-address=0.0.0.0
            - --secure-port=10357
            - --leader-elect-resource-namespace=default
            - --v=2
            - --controllers=*,-hpa
            - --feature-gates=PropagateDeps=false
```

calvin0327: value.yaml:
```yaml
controllers: [ "a", "b", "-c", "*" ]
```

template:
```yaml
{{- define "karmada.controllerManager.controllers" -}}
  {{- if (not (empty .controllers)) }}
    {{- $controllersFlag := "" -}}
    {{- range .controllers }}
      {{- if not (empty (toString .)) }}
        {{- $controllersFlag = append $controllersFlag . -}}
      {{- end }}
    {{- end }}

    {{- if gt (len $controllers) 0 }}
      {{- $controllersFlag := trimSuffix "," $controllersFlag  | nospace -}}
      {{- printf "%s=%s" "--controllers" $controllersFlag -}}
    {{- end }}
  {{- end }}
{{- end -}}
```

uasage:
```yaml
- --kubeconfig=/etc/kubeconfig
- --bind-address=0.0.0.0
{{- if .Values.controllers }}
- {{ include "karmada.controllerManager.controllers" (dict "controllers" .Values.controllers) }}
{{- end }}
```

@Poor12 @RainbowMango That's the probable idea, How do you think?

RainbowMango: I‘d prefer to string one instead of the slice. Because it is more straightforward that we don't need to involve much conversion work.

calvin0327: indeed, The way is complicated

RainbowMango: Thanks. I guess we can move forward now.

RainbowMango (approved):
/lgtm
/approve

yanggangtony:
I have a question , how about the user define ```controllerManager.config.controllers , controllerManager.controllers ```
at the same time .

And why retention the 2 parms  ?

Poor12:
> I have a question , how about the user define `controllerManager.config.controllers , controllerManager.controllers ` at the same time .
> 
> And why retention the 2 parms ?
> 
> @Poor12 @RainbowMango @calvin0327

Sorry, I made a mistake. I forgot to delete the config.controllers things when I cherry-pick #2542. Would you like to fix it?

RainbowMango:
Thanks @yanggangtony for spotting this, that's my bad for missing it from the review.

yanggangtony:
thanks all for the re-check .
:smiley::smiley:

PR #3241, opened 2023-03-06
Increase `.metadata.generation` once desired state of the `Cluster` object is changed.

XiShanYongYe-Chang:
**What type of PR is this?**

/kind feature

We should set the `generation` field of the Cluster object. When the `spec` part of the Cluster object changes, the `generation` value should increase. Its value begins with 1.

Fixes #

Do we need to pay attention to the changes in labels and annotations of the Cluster object?

```release-note
`karmada-aggregated-apiserver`: Increase `.metadata.generation` once desired state of the `Cluster` object is changed.
```

### DIFF
--- a/pkg/registry/cluster/strategy.go
+++ b/pkg/registry/cluster/strategy.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -73,6 +74,10 @@ func (Strategy) GetResetFields() map[fieldpath.APIVersion]*fieldpath.Set {
 // PrepareForCreate is invoked on create before validation to normalize the object.
 func (Strategy) PrepareForCreate(ctx context.Context, obj runtime.Object) {
 	cluster := obj.(*clusterapis.Cluster)
+	cluster.Status = clusterapis.ClusterStatus{}
+
+	cluster.Generation = 1
+
 	if utilfeature.DefaultMutableFeatureGate.Enabled(features.CustomizedClusterResourceModeling) {
 		if len(cluster.Spec.ResourceModels) == 0 {
 			mutation.SetDefaultClusterResourceModels(cluster)
@@ -84,10 +89,18 @@ func (Strategy) PrepareForCreate(ctx context.Context, obj runtime.Object) {
 
 // PrepareForUpdate is invoked on update before validation to normalize the object.
 func (Strategy) PrepareForUpdate(ctx context.Context, obj, old runtime.Object) {
-	cluster := obj.(*clusterapis.Cluster)
+	newCluster := obj.(*clusterapis.Cluster)
+	oldCluster := old.(*clusterapis.Cluster)
+	newCluster.Status = oldCluster.Status
+
+	// Any changes to the spec increases the generation number.
+	if !apiequality.Semantic.DeepEqual(newCluster.Spec, oldCluster.Spec) {
+		newCluster.Generation = oldCluster.Generation + 1
+	}
+
 	if utilfeature.DefaultMutableFeatureGate.Enabled(features.CustomizedClusterResourceModeling) {
-		if len(cluster.Spec.ResourceModels) != 0 {
-			mutation.StandardizeClusterResourceModels(cluster.Spec.ResourceModels)
+		if len(newCluster.Spec.ResourceModels) != 0 {
+			mutation.StandardizeClusterResourceModels(newCluster.Spec.ResourceModels)
 		}
 	}
 }
--- a/pkg/scheduler/event_handler.go
+++ b/pkg/scheduler/event_handler.go
@@ -167,7 +167,7 @@ func (s *Scheduler) updateCluster(oldObj, newObj interface{}) {
 	switch {
 	case !equality.Semantic.DeepEqual(oldCluster.Labels, newCluster.Labels):
 		fallthrough
-	case !equality.Semantic.DeepEqual(oldCluster.Spec, newCluster.Spec):
+	case oldCluster.Generation != newCluster.Generation:
 		s.enqueueAffectedBindings(oldCluster, newCluster)
 	}
 }
--- a/pkg/registry/cluster/strategy_test.go
+++ b/pkg/registry/cluster/strategy_test.go
@@ -34,6 +34,12 @@ func getValidCluster(name string) *clusterapis.Cluster {
 	}
 }
 
+func getValidClusterWithGeneration(name string, generation int64) *clusterapis.Cluster {
+	cluster := getValidCluster(name)
+	cluster.Generation = generation
+	return cluster
+}
+
 func setFeatureGateDuringTest(tb testing.TB, gate featuregate.FeatureGate, f featuregate.Feature, value bool) func() {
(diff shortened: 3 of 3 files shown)


pkg/registry/cluster/strategy.go
```diff
@@ -73,6 +74,10 @@ func (Strategy) GetResetFields() map[fieldpath.APIVersion]*fieldpath.Set {
 // PrepareForCreate is invoked on create before validation to normalize the object.
 func (Strategy) PrepareForCreate(ctx context.Context, obj runtime.Object) {
 	cluster := obj.(*clusterapis.Cluster)
+	cluster.Status = clusterapis.ClusterStatus{}
```
Garrybest, line 77: I think this line is useless, why we initialize` cluster.Status` here? It is not a pointer.

pkg/registry/cluster/strategy.go
```diff
@@ -84,10 +89,17 @@ func (Strategy) PrepareForCreate(ctx context.Context, obj runtime.Object) {
 
 // PrepareForUpdate is invoked on update before validation to normalize the object.
 func (Strategy) PrepareForUpdate(ctx context.Context, obj, old runtime.Object) {
-	cluster := obj.(*clusterapis.Cluster)
+	oldCluster := old.(*clusterapis.Cluster)
+	newCluster := obj.(*clusterapis.Cluster)
+
+	// Any changes to the spec increment the generation number.
```
Garrybest, line 95: increment -> increases

pkg/registry/cluster/strategy.go
XiShanYongYe-Chang: Just like k8s native resource like this:

https://github.com/kubernetes/kubernetes/blob/f461527491effcdbe23814b7481f46723cddda9e/pkg/registry/apps/deployment/strategy.go#L79

https://github.com/kubernetes/kubernetes/blob/master/pkg/registry/apps/daemonset/strategy.go#L76

etc.

pkg/registry/cluster/strategy.go
```diff
@@ -84,10 +89,17 @@ func (Strategy) PrepareForCreate(ctx context.Context, obj runtime.Object) {
 
 // PrepareForUpdate is invoked on update before validation to normalize the object.
 func (Strategy) PrepareForUpdate(ctx context.Context, obj, old runtime.Object) {
-	cluster := obj.(*clusterapis.Cluster)
+	oldCluster := old.(*clusterapis.Cluster)
+	newCluster := obj.(*clusterapis.Cluster)
+
+	// Any changes to the spec increases the generation number.
+	if !apiequality.Semantic.DeepEqual(newCluster.Spec, oldCluster.Spec) {
+		newCluster.Generation = oldCluster.Generation + 1
```
jwcesign, line 97: If the label changes, will it trigger rescheduling?

pkg/registry/cluster/strategy.go
XiShanYongYe-Chang: updated.

pkg/registry/cluster/strategy.go
XiShanYongYe-Chang: As I explained, we need to discuss this. Same as `annotation`.

chaunceyjiang: https://github.com/karmada-io/karmada/blob/master/pkg/scheduler/event_handler.go#L170

Do we need to change it here as well? We should use `Generation` to determine this.

XiShanYongYe-Chang: Thanks for the reminder, and as @jwcesign said, combined with the usage in the scheduler event handler, I think we can take the label changes into account.

RainbowMango: Seems there is no guideline about if a `generation` should be increased in case of `label` updation, but `no resource in k8s will do that`.

The only benefit we can get from that is the scheduler doesn't have to compare the labels here:
https://github.com/karmada-io/karmada/blob/2c834fe94eb2eff8a9c6954a4c2b0ad567d6caf1/pkg/scheduler/event_handler.go#L168-L172

Please remind me if I miss any benefits. 
It's hard to make the choice, I'd like your opinion. @XiShanYongYe-Chang @chaunceyjiang @Garrybest

pkg/scheduler/event_handler.go
```diff
@@ -164,10 +163,7 @@ func (s *Scheduler) updateCluster(oldObj, newObj interface{}) {
 		s.schedulerEstimatorWorker.Add(newCluster.Name)
 	}
 
-	switch {
-	case !equality.Semantic.DeepEqual(oldCluster.Labels, newCluster.Labels):
-		fallthrough
-	case !equality.Semantic.DeepEqual(oldCluster.Spec, newCluster.Spec):
+	if oldCluster.Generation != newCluster.Generation {
```
RainbowMango, line 166: If we are going to do this, make sure to add comments to explain the generation change probably caused by label change.

pkg/registry/cluster/strategy.go
XiShanYongYe-Chang: It's really hard to choose. From an unknown future expansion point of view, addition is better than subtraction, so I think it might be better not to consider labels here.

pkg/registry/cluster/strategy.go
XiShanYongYe-Chang: I think the initialization status here can filters the situation where the status is specified during creation.

I feel that the changes in the status section can be done in another pr, so I will remove this change.

pkg/registry/cluster/strategy.go
chaunceyjiang: I prefer to compare only the spec section here, and then use `generation` and `label` in the event_handler.go.

RainbowMango: Thanks, then please @XiShanYongYe-Chang make changes.

RainbowMango:
@Garrybest @chaunceyjiang Do you have any other comments?

> Add generation filed for Cluster object.

You are not `adding` a field to Cluster.

RainbowMango:
@XiShanYongYe-Chang Have you tested it? Please share a test report here.

chaunceyjiang:
> @Garrybest @chaunceyjiang Do you have any other comments?

No, LGTM

XiShanYongYe-Chang:
Before update:

```yaml
apiVersion: cluster.karmada.io/v1alpha1
kind: Cluster
metadata:
  creationTimestamp: "2023-03-06T12:24:41Z"
  finalizers:
  - karmada.io/cluster-controller
  generation: 5
  name: member1
  resourceVersion: "804"
  uid: c2b697d0-d9cc-4e32-982b-eb7c3f50082d
spec:
  apiEndpoint: https://172.18.0.4:6443
  id: 668b5146-f8f3-4592-847a-a79c34106193
  impersonatorSecretRef:
    name: member1-impersonator
    namespace: karmada-cluster
  ...
```

When I update the cluster apiEndpoint field, after:

```yaml
apiVersion: cluster.karmada.io/v1alpha1
kind: Cluster
metadata:
  creationTimestamp: "2023-03-06T12:24:41Z"
  finalizers:
  - karmada.io/cluster-controller
  generation: 6
  name: member1
  resourceVersion: "844527"
  uid: c2b697d0-d9cc-4e32-982b-eb7c3f50082d
spec:
  apiEndpoint: https://172.18.0.4:6444
  id: 668b5146-f8f3-4592-847a-a79c34106193
  impersonatorSecretRef:
    name: member1-impersonator
    namespace: karmada-cluster
  ...
```

RainbowMango (approved):
/lgtm
/approve

PR #2973, opened 2022-12-18
add operator skeleton

carlory:
**What type of PR is this?**

/kind feature

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

/cc @lonelyCZ @RainbowMango

PR #3245, opened 2023-03-07
karmada operator: add deinit workflow

calvin0327:
**What type of PR is this?**
/kind feature

add karmada operator deInit workflow to remove a karmada instance. the workflow only have three task: `RemoveComponent` => `CleanupCert` => `CleanupKubeconfig`.

Fixes #

Test result:

image

```release-note
NONE
```


calvin0327:
@lonelyCZ @Poor12 PTAL

operator/pkg/controller/karmada/controller.go
```diff
@@ -58,41 +58,31 @@ func (ctrl *Controller) Reconcile(ctx context.Context, req controllerruntime.Req
 		if !controllerutil.ContainsFinalizer(karmada, ControllerFinalizerName) {
 			controllerutil.AddFinalizer(karmada, ControllerFinalizerName)
 			if err := ctrl.Update(ctx, karmada); err != nil {
-				return controllerruntime.Result{}, err
+				return controllerruntime.Result{Requeue: true}, err
```
Poor12, line 61: I remember the controller will requeue the request to be processed if err is not nil https://github.com/karmada-io/karmada/pull/2973#discussion_r1052886320. I feel like add or not-add is both ok. The key is that we might need to unify the code style in one module which may not confuse others.

operator/pkg/tasks/deint/component.go
```diff
@@ -0,0 +1,95 @@
...
+)
+
+// NewRemoveComponentTask init a remove karmada components task
+func NewRemoveComponentTask() workflow.Task {
+	return workflow.Task{
+		Name:        "remove-component",
+		Run:         runRemoveComponent,
+		RunSubTasks: true,
+		Tasks: []workflow.Task{
+			newRemoveComponentSubTask(constants.KarmadaWebhookComponent, util.KarmadaWebhookName),
```
Poor12, line 22: I remember there is some svc installed. Do we need to uninstall them when deinit?

operator/pkg/controller/karmada/controller.go
calvin0327: I agree to that

operator/pkg/tasks/deint/component.go
calvin0327: I want to set the service OwnReference to their workflow. so, we only delete the worklfow.
@Poor12  How do you think?

Poor12: For now I think it's okay. I wonder if there is a scenario where we need to consider deployment and svc separately.

calvin0327: Ok, If the service resource exists alone, it doesn't to make sense. I seem not encounter the scenario.

lonelyCZ:
Thanks @calvin0327 , I will test it in my env ASAP.

/assign

Poor12:
Basicly LGTM
Leave comments for @lonelyCZ .

operator/pkg/controller/karmada/controller.go
```diff
@@ -61,38 +61,28 @@ func (ctrl *Controller) Reconcile(ctx context.Context, req controllerruntime.Req
...
 		return controllerruntime.Result{}, err
 	}
-	return planner.Execute()
-}
+	if err := planner.Execute(); err != nil {
+		return controllerruntime.Result{}, err
+	}
+
+	// The object is being deleted
+	if !karmada.DeletionTimestamp.IsZero() && controllerutil.ContainsFinalizer(karmada, ControllerFinalizerName) {
```
lonelyCZ, line 76: Suggest to move this forward to keep the controller code style. Firstly check `!karmada.DeletionTimestamp.IsZero()` and then sync cr, which can reduce the number of judgments.

like 

https://github.com/karmada-io/karmada/blob/a4d3eced85ede213c9c6cf6813ea564dde1b6583/pkg/controllers/cluster/cluster_controller.go#L171-L175

operator/pkg/controller/karmada/controller.go
```diff
@@ -61,38 +61,28 @@ func (ctrl *Controller) Reconcile(ctx context.Context, req controllerruntime.Req
...
+		// remove our finalizer from the list and update it.
```
lonelyCZ, line 77:
```suggestion
		// remove the finalizer from the list and update it.
```

operator/pkg/controller/karmada/controller.go
```diff
@@ -61,38 +61,28 @@ func (ctrl *Controller) Reconcile(ctx context.Context, req controllerruntime.Req
...
+		controllerutil.RemoveFinalizer(karmada, ControllerFinalizerName)
```
lonelyCZ, line 78: Suggest to encapsulate it as a function.

like

https://github.com/karmada-io/karmada/blob/a4d3eced85ede213c9c6cf6813ea564dde1b6583/pkg/controllers/execution/execution_controller.go#L171-L183

calvin0327: ok

operator/pkg/controller/karmada/controller.go
calvin0327: @lonelyCZ  Ok, I got it, but in this scenario, must put the check `!cluster.DeletionTimestamp.IsZero()`  behind the process of `reconciliation` due to both Init and Deinit workflow are to excute the following logic:

```golang
klog.V(2).InfoS("Reconciling karmada", "name", req.Name)
	planner, err := NewPlannerFor(karmada, ctrl.Client, ctrl.Config)
	if err != nil {
	}
	if err := planner.Execute(); err != nil {
	}
```

lonelyCZ:
Hi @calvin0327 , could you provide `operator/config/deploy/karmada-operator.yaml` again, I want to test it in my env.

calvin0327:
> Hi @calvin0327 , could you provide `operator/config/deploy/karmada-operator.yaml` again, I want to test it in my env.

Of cource, here:

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: karmada-operator
  namespace: karmada-system
  labels:
    karmada-app: karmada-operator
spec:
  replicas: 1
  selector:
    matchLabels:
      karmada-app: karmada-operator
  template:
    metadata:
      labels:
        karmada-app: karmada-operator
    spec:
      containers:
      - name: karmada-operator
        image: release.daocloud.io/karmada/karmada-operator:v1.1.0-1965-g20c7316c
        imagePullPolicy: IfNotPresent
        command:
        - /bin/karmada-operator
        - --kubeconfig=/etc/config
        - --v=4
        volumeMounts:
        - name: kubeconfig
          mountPath: /etc/config
          subPath: config
      volumes:
      - name: kubeconfig
        secret:
          secretName: my-kubeconfig
```

lonelyCZ:
Hi @calvin0327 , Is `release.daocloud.io/karmada/karmada-operator:v1.1.0-1965-g10323eac` a newest image?

calvin0327:
> 

The image release.daocloud.io/karmada/karmada-operator:v1.1.0-1965-g10323eac is latest image.

operator/pkg/tasks/deint/cert.go
```diff
@@ -0,0 +1,65 @@
+package tasks
```
lonelyCZ, line 1: Please update `deint/cert.go` to `deinit/cert.go`

lonelyCZ (commented):
I just tested it in my env that worked fine! 

There is only a spell-error here.

Poor12:
> > 
> 
> @lonelyCZ The image release.daocloud.io/karmada/karmada-operator:v1.1.0-1965-g10323eac is latest image.

BTW, we can make image locally. I guess it's the same.

lonelyCZ:
> BTW, we can make image locally. I guess it's the same.

Yes, we can add it to Makefile in next PR.

Poor12:
> > BTW, we can make image locally. I guess it's the same.
> 
> Yes, we can add it to Makefile in next PR.

It's alreay done. You can use `make image-karmada-operator` to make the image.

RainbowMango:
Seems we don't have the image registry on the Docker Hub yet.  
Can we create the registry now?

RainbowMango:
Oh, i see #3270.
I'll get back to #3270 once I created the repo.

operator/pkg/tasks/deint/cert.go
calvin0327: updated. @lonelyCZ

calvin0327:
@Poor12 @lonelyCZ @RainbowMango any others coments?

operator/pkg/constants/constants.go
```diff
@@ -76,4 +82,14 @@ const (
 	KarmadaControllerManagerComponent = "KarmadaControllerManager"
 	// KarmadaSchedulerComponent defines the name of karmada scheduler component
 	KarmadaSchedulerComponent = "KarmadaScheduler"
+	// KarmadaWebhookComponent defines the name of the karmada-webhook component
+	KarmadaWebhookComponent = "karmadaWebhook"
```
lonelyCZ, line 86: Is it right?
```suggestion
	KarmadaWebhookComponent = "KarmadaWebhook"
```

calvin0327: @lonelyCZ Thanks. done

RainbowMango (approved):
/approve
Thanks.

PR #3259, opened 2023-03-10
region selection algorithm optimization

### LINKED ISSUE #3253: Region selection algorithm optimization
There are some problems with the current region selection algorithm, and I think it should be further optimized.

**Why is this needed**:
I will give the following two examples to illustrate why the region selection algorithm needs to be optimized.

**Example-1: those regions that meet the cluster's the quantity requirements(`[minGroups, maxGroups]`) and have higher score should be selected**
Assuming that there is the following region distribution:

regions-1

and we use those spread constraints:
```
    spreadConstraints:
    - maxGroups: 3
      minGroups: 2
      spreadByField: region
    - maxGroups: 6
      minGroups: 4
      spreadByField: cluster
```
they will be sorted by score first:
```
R3 -> R1 -> R2 -> R4
```
then according to the existing karmada algorithm, `R3` and `R4` will eventually be selected. But `R2` and `R3` are the best choice because the score of `R2` is higher than `R4` and they can both meet the cluster's the quantity requirements.

**Example-2: if the quantity of any `minGroups` regions’ clusters cannot meet the cluster's the quantity requirements(`[minGroups, maxGroups]`), we should try to increase the number of regions to meet the requirements instead of giving up**
Assuming that there is the following region distribution:

region-2

and we use those spread constraints:
```
    spreadConstraints:
    - maxGroups: 3
      minGroups: 1
      spreadByField: region
    - maxGroups: 8
      minGroups: 6
      spreadByField: cluster
```
they will be sorted by score first:
```
R4 -> R1 -> R3 -> R5 -> R2
```
then according to the existing karmada algorithm, no region will eventually be selected. But `R2`, `R3` and `R4` are the best choice because the score of `R4` is higher than `R1` and `R5`, `R2` and `R3` are required.

whitewindmills:
**What type of PR is this?**
/kind feature

Fixes #3253

See the above issue for details.

```release-note
`karmada-scheduler`: optimize the region selection algorithm
```

### DIFF
--- a/pkg/scheduler/core/spreadconstraint/select_clusters_by_region.go
+++ b/pkg/scheduler/core/spreadconstraint/select_clusters_by_region.go
@@ -2,7 +2,8 @@ package spreadconstraint
 
 import (
 	"fmt"
-	"sort"
+
+	"k8s.io/utils/pointer"
 
 	clusterv1alpha1 "github.com/karmada-io/karmada/pkg/apis/cluster/v1alpha1"
 	policyv1alpha1 "github.com/karmada-io/karmada/pkg/apis/policy/v1alpha1"
@@ -18,7 +19,8 @@ func selectBestClustersByRegion(spreadConstraintMap map[policyv1alpha1.SpreadFie
 	}
 
 	// firstly, select regions which have enough clusters to satisfy the cluster and region propagation constraints
-	regions := selectRegions(groupClustersInfo.Regions, spreadConstraintMap[policyv1alpha1.SpreadByFieldRegion], spreadConstraintMap[policyv1alpha1.SpreadByFieldCluster])
+	regions := selectRegions(groupClustersInfo.Regions, spreadConstraintMap[policyv1alpha1.SpreadByFieldRegion],
+		spreadConstraintMap[policyv1alpha1.SpreadByFieldCluster])
 	if len(regions) == 0 {
 		return nil, fmt.Errorf("the number of clusters is less than the cluster spreadConstraint.MinGroups")
 	}
@@ -35,68 +37,67 @@ func selectBestClustersByRegion(spreadConstraintMap map[policyv1alpha1.SpreadFie
 	}
 
 	// thirdly, select the remaining Clusters based cluster.Score
-	sortClusters(candidateClusters)
 	restCnt := needCnt - len(clusters)
-	for i := 0; i < restCnt; i++ {
-		clusters = append(clusters, candidateClusters[i].Cluster)
+	if restCnt > 0 {
+		sortClusters(candidateClusters, func(i *ClusterDetailInfo, j *ClusterDetailInfo) *bool {
+			if i.AvailableReplicas != j.AvailableReplicas {
+				return pointer.Bool(i.AvailableReplicas > j.AvailableReplicas)
+			}
+			return nil
+		})
+		for i := 0; i < restCnt; i++ {
+			clusters = append(clusters, candidateClusters[i].Cluster)
+		}
 	}
 
 	return clusters, nil
 }
 
-func selectRegions(RegionInfos map[string]RegionInfo, regionConstraint, clusterConstraint policyv1alpha1.SpreadConstraint) []RegionInfo {
-	var regions []RegionInfo
-	for i := range RegionInfos {
-		regions = append(regions, RegionInfos[i])
+// selectRegions is an implementation of the region selection algorithm, the purpose of
+// the region selection algorithm is to try to find those regions that satisfy all spread constraints.
+//
+// First, it needs to be clear how many regions we need. of course, it needs to meet the
+// region spread constraints([minGroups, maxGroups]), and starts from region's minGroups.
+//
+// Second, we need to find the combination with the highest score when there are multiple
+// sets of choices that can satisfy the cluster spread constraint.
+// For example:
+//
+//	R1 -> R2 -> R3 -> R4
+//
+// We assume that those regions have already done the above descending order according to
+// the score. Now those combinations([R1,R3] and [R1,R4]) can all satisfy all spread constraints.
+// Who is the best choice?
+// Obviously [R1,R3].
+//
+// Finally, the number of regions that we need should be auto-incremented when there is no
+// combination to meet cluster spread constraint. Of course, it can never exceed region's
+// maxGroups or the total number of regions. Let's use the example from the second statement above:
+//
+//	R1 -> R2 -> R3 -> R4
+//
+// When any combination of two regions cannot satisfy cluster spread constraint, we should try
+// to increase the number of regions. It is very likely that we will find the best choice by trying
+// to choose a combination of three regions.
+func selectRegions(regionMap map[string]RegionInfo, regionConstraint, clusterConstraint policyv1alpha1.SpreadConstraint) []RegionInfo {
+	groups := make([]*GroupInfo, 0, len(regionMap))
+	for _, region := range regionMap {
+		group := &GroupInfo{
+			name:   region.Name,
+			value:  len(region.Clusters),
+			weight: region.Score,
+		}
+		groups = append(groups, group)
 	}
 
-	sort.Slice(regions, func(i, j int) bool {
-		if regions[i].Score != regions[j].Score {
-			return regions[i].Score > regions[j].Score
-		}
-
-		return regions[i].Name < regions[j].Name
-	})
-
-	retRegions := regions[:regionConstraint.MinGroups]
-	candidateRegions := regions[regionConstraint.MinGroups:]
-	var replaceID = len(retRegions) - 1
-	for !checkClusterTotalForRegion(retRegions, clusterConstraint) && replaceID >= 0 {
-		regionID := getRegionWithMaxClusters(candidateRegions, len(retRegions[replaceID].Clusters))
-		if regionID == InvalidRegionID {
-			replaceID--
-			continue
-		}
-
-		retRegions[replaceID], candidateRegions[regionID] = candidateRegions[regionID], retRegions[replaceID]
-		replaceID--
-	}
-
-	if !checkClusterTotalForRegion(retRegions, clusterConstraint) {
+	groups = selectGroups(groups, regionConstraint.MinGroups, regionConstraint.MaxGroups, clusterConstraint.MinGroups)
+	if len(groups) == 0 {
 		return nil
 	}
 
-	return retRegions
-}
-
-func checkClusterTotalForRegion(regions []RegionInfo, clusterConstraint policyv1alpha1.SpreadConstraint) bool {
-	var sum int
-	for i := range regions {
-		sum += len(regions[i].Clusters)
+	result := make([]RegionInfo, 0, len(groups))
+	for _, group := range groups {
+		result = append(result, regionMap[group.name])
 	}
-
-	return sum >= clusterConstraint.MinGroups
-}
-
-func getRegionWithMaxClusters(candidateRegions []RegionInfo, originClusters int) int {
-	var maxClusters = originClusters
-	var regionID = -1
-	for i := range candidateRegions {
-		if maxClusters < len(candidateRegions[i].Clusters) {
-			regionID = i
-			maxClusters = len(candidateRegions[i].Clusters)
-		}
-	}
-
-	return regionID
+	return result
 }
--- a/pkg/scheduler/core/spreadconstraint/select_clusters_by_region_test.go
+++ b/pkg/scheduler/core/spreadconstraint/select_clusters_by_region_test.go
@@ -37,9 +37,8 @@ func Test_selectBestClustersByRegion(t *testing.T) {
 				groupClustersInfo: &GroupClustersInfo{
 					Regions: map[string]RegionInfo{
 						"R1": {
-							Name:              "R1",
-							Score:             100,
-							AvailableReplicas: 100,
+							Name:  "R1",
+							Score: 100,
 							Clusters: []ClusterDetailInfo{
 								{
 									Name:              "member1",
@@ -56,9 +55,8 @@ func Test_selectBestClustersByRegion(t *testing.T) {
 							},
 						},
 						"R2": {
-							Name:              "R2",
-							Score:             80,
-							AvailableReplicas: 100,
+							Name:  "R2",
+							Score: 80,
 							Clusters: []ClusterDetailInfo{
 								{
 									Name:              "member3",
@@ -107,9 +105,8 @@ func Test_selectBestClustersByRegion(t *testing.T) {
 				groupClustersInfo: &GroupClustersInfo{
 					Regions: map[string]RegionInfo{
 						"R1": {
-							Name:              "R1",
-							Score:             100,
-							AvailableReplicas: 100,
+							Name:  "R1",
+							Score: 100,
 							Clusters: []ClusterDetailInfo{
 								{
 									Name:              "member1",
@@ -126,9 +123,8 @@ func Test_selectBestClustersByRegion(t *testing.T) {
 							},
 						},
 						"R2": {
-							Name:              "R2",
-							Score:             80,
-							AvailableReplicas: 100,
+							Name:  "R2",
+							Score: 80,
 							Clusters: []ClusterDetailInfo{
 								{
 									Name:              "member3",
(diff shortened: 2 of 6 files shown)


RainbowMango:
Thanks @whitewindmills, I'd like to invite @jwcesign to take a look.
I'm looing at it too.

pkg/scheduler/core/spreadconstraint/select_clusters_by_region.go
```diff
@@ -35,66 +36,128 @@ func selectBestClustersByRegion(spreadConstraintMap map[policyv1alpha1.SpreadFie
...
 		}
+	}
+
+	return nil
+}
 
-		retRegions[replaceID], candidateRegions[regionID] = candidateRegions[regionID], retRegions[replaceID]
-		replaceID--
+// doSelectRegions selects feasible regions that meet the cluster spread constraint and have higher score.
+func doSelectRegions(regions []RegionInfo, wantedGroupCnt, minClusterCnt int) []RegionInfo {
```
jwcesign, line 111: The last parameter is MinGroups, but the meaning changes to cluster count here.
`getClusterDeltaForRegion` also uses it like cluster count instead of region group(one group contains multiple clusters).

Can you explain more about this?

whitewindmills: we need to find any `wantedGroupCnt` regions to meet the min cluster constraint(that’s `minClusterCnt`). just select those regions with highest score if there are multiple choices.
About `getClusterDeltaForRegion` func, his duty is to calculate the delta between  the cluster number of current regions and `minClusterCnt`. The delta will be used to find who is the best.

pkg/scheduler/core/spreadconstraint/util.go
```diff
@@ -32,6 +32,7 @@ func sortClusters(infos []ClusterDetailInfo) {
 			return infos[i].Score > infos[j].Score
 		}
 
-		return infos[i].Name < infos[j].Name
+		// It makes more sense to compare the available replicas of the cluster.
+		return infos[i].AvailableReplicas > infos[j].AvailableReplicas
```
jwcesign, line 36: If the AvailableReplicas is the same, I prefer to keep it in name order, or it's unpredictable.

whitewindmills: Yeah, I will adjust it.

pkg/scheduler/core/spreadconstraint/select_clusters_by_region_test.go
```diff
@@ -160,6 +160,235 @@ func Test_selectBestClustersByRegion(t *testing.T) {
...
+						"R1": {
+							Name:              "R1",
+							Score:             90,
+							AvailableReplicas: 40,
+							Clusters: []ClusterDetailInfo{
+								{
+									Name:              "member1",
+									Score:             90,
+									AvailableReplicas: 40,
+									Cluster:           NewClusterWithTopology("member1", "P1", "R1", "Z1"),
```
jwcesign, line 189: sum of clusters' AvailableReplicas should be same with region group's AvailableReplicas

jwcesign:
cc @chaunceyjiang, is it better to do this after your refactoring?

pkg/scheduler/core/spreadconstraint/select_clusters_by_region.go
```diff
@@ -35,66 +36,131 @@ func selectBestClustersByRegion(spreadConstraintMap map[policyv1alpha1.SpreadFie
...
+			result[replaceID], candidateRegions[regionID] = candidateRegions[regionID], result[replaceID]
+			return result
+		}
+	}
+
+	// When the number of regions that we need is equal to 1,
+	// we do not need to recalculate the region selection by shifting the candidates
+	// because there is no region to meet the cluster spread constraint.
+	if wantedGroupCnt > 1 && wantedGroupCnt <= len(candidateRegions) {
+		return doSelectRegions(candidateRegions, wantedGroupCnt, minClusterCnt)
```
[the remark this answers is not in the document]
whitewindmills: Thanks for the algorithm error you've found. I guess I've found the problem.

pkg/scheduler/core/spreadconstraint/select_clusters_by_region_test.go
whitewindmills: Yeah, but I don't see any problem.

jwcesign: Just for making people not confused when they see the ut.

jwcesign: Sorry, you already fixed it in your newest commit.

jwcesign: for the case `no enough region`, `R1` has two clusters(40/40 AvailableReplicas), but this group has 100 AvailableReplicas, can you help fix the related problem?

whitewindmills: I get it. I thought those test cases that I recently added had this problem. I'll fix them together in my next commit.

pkg/scheduler/core/spreadconstraint/select_clusters_by_region.go
```diff
@@ -35,66 +36,131 @@ func selectBestClustersByRegion(spreadConstraintMap map[policyv1alpha1.SpreadFie
...
+		regions = append(regions, region)
 	}
 
 	sort.Slice(regions, func(i, j int) bool {
 		if regions[i].Score != regions[j].Score {
 			return regions[i].Score > regions[j].Score
 		}
 
+		if regions[i].AvailableReplicas != regions[j].AvailableReplicas {
+			return regions[i].AvailableReplicas > regions[j].AvailableReplicas
```
chaunceyjiang, line 63: Can you develop a `score` scheduler plugin based on cluster `AvailableReplicas`? This way we don't need to deal with AvailableReplicas here.

whitewindmills: It's a good idea! I think this is a new feature and I'd love to work on it, but I don't think it should be included in this PR.
@RainbowMango How do you think of this?

RainbowMango: I'll get back to this tomorrow.

whitewindmills: I think the code is a bit redundant in https://github.com/karmada-io/karmada/pull/3287, and the score result is best kept in the [0,100] interval. Maybe we need to discuss it further.

pkg/scheduler/core/spreadconstraint/select_clusters_by_region.go
```diff
@@ -35,66 +36,131 @@ func selectBestClustersByRegion(spreadConstraintMap map[policyv1alpha1.SpreadFie
...
+// getSortedRegions returns sorted region slice by score and available replicas.
+func getSortedRegions(regionInfoMap map[string]RegionInfo) []RegionInfo {
+	regions := make([]RegionInfo, 0, len(regionInfoMap))
+	for _, region := range regionInfoMap {
+		regions = append(regions, region)
 	}
 
 	sort.Slice(regions, func(i, j int) bool {
 		if regions[i].Score != regions[j].Score {
 			return regions[i].Score > regions[j].Score
```
chaunceyjiang, line 59: By the way, currently the region score is the highest score within the cluster. I think this algorithm may not reflect the overall situation of this regional cluster very well. Would it be more reasonable to use median, average or some other algorithm instead?

whitewindmills: Yeah, I've done this [optimization,](https://github.com/karmada-io/karmada/pull/3259/files#diff-d71362f4d2ac9a200a70291dd5117b7c6e3bc9de9880a23dc536361606ad3ad5R136) it does not conflict with your PR(https://github.com/karmada-io/karmada/pull/3287). Can you take a look?

pkg/scheduler/core/spreadconstraint/select_clusters_by_region.go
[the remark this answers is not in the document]
whitewindmills: Hi, @jwcesign
I refactored the previous algorithm, can you help to continue the review?
In addition, based on the configuration you gave above, I think the best choice should be `R1 R2 R3 R4`.

/cc @RainbowMango

whitewindmills:
Hi @RainbowMango 
I see a failed test, but only from the e2e logs, I can't understand why it failed, can you help to see the log of Karmada?

RainbowMango:
> propagation with label and group constraints testing Deployment propagation testing [It] deployment propagation with label and group constraints testing
/home/runner/work/karmada/karmada/test/e2e/scheduling_test.go:86

Seems this is a relevant test. You can download the logs from [here](https://github.com/karmada-io/karmada/suites/11670293745/logs?attempt=1).

pkg/scheduler/core/spreadconstraint/select_groups.go
```diff
@@ -0,0 +1,196 @@
...
+	"k8s.io/klog/v2"
+)
+
+// GroupInfo indicate the group info.
+type GroupInfo struct {
+	name   string
+	value  int
+	weight int64
+}
+
```
jwcesign, line 17: People may be confused by the terms 'value' and 'weight'. Perhaps using 'clustersNum' for value and 'groupScore' for weight would be clearer.

pkg/scheduler/core/spreadconstraint/group_clusters.go
```diff
@@ -130,7 +133,42 @@ func (info *GroupClustersInfo) generateClustersInfo(clustersScore framework.Clus
 	sortClusters(info.Clusters)
 }
 
-func (info *GroupClustersInfo) generateZoneInfo(spreadConstraints []policyv1alpha1.SpreadConstraint) {
+func (info *GroupClustersInfo) calcGroupScore(clusters []ClusterDetailInfo, rbSpec *workv1alpha2.ResourceBindingSpec) int64 {
+	// First, calculate the base score.
+	var highScoreSum int64
+	var selectedNum int
```
jwcesign, line 139: Can you explain the reason why this algorithm is done in this way?

pkg/scheduler/core/spreadconstraint/select_groups.go
```diff
@@ -0,0 +1,196 @@
...
+			}
+
+			return groups[i].name < groups[j].name
+		})
+	}
+
+	rootPath := new(dfsPath)
+	var dfsFunc func(int, int)
+	dfsFunc = func(sum, begin int) {
+		if sum >= target && rootPath.length() >= minConstraint && rootPath.length() <= maxConstraint {
```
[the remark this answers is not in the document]
jwcesign: Thank you for pointing out this scene, I didn't find it before. 
And how about this situation? rootPath.length() is bigger than maxConstraint, but sum is smaller than target?
I believe it should be returned directly in the following manner.
```
	dfsFunc = func(sum, begin int) {
		if rootPath.length() > maxConstraint {
			return
		}
		if sum >= target && rootPath.length() >= minConstraint {
			paths = append(paths, rootPath.next())
			return
		}
```

pkg/scheduler/core/spreadconstraint/group_clusters.go
XiShanYongYe-Chang: Hi @whitewindmills, thanks for your hard work, can we divide the improvement of the calculation score into a separate pr? Let's focus on the reconstruction selection algorithm part so that the code review and integration can be accelerated.

whitewindmills: Cool! It's a good idea. I'll do that.

pkg/scheduler/core/spreadconstraint/select_groups.go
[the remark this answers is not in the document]
whitewindmills: >And how about this situation? rootPath.length() is bigger than maxConstraint, but sum is smaller than target? I think it should be returned directly in the following manner.
> 
> ```
> 	dfsFunc = func(sum, begin int) {
> 		if rootPath.length() > maxConstraint {
> 			return
> 		}
> 		if sum >= target && rootPath.length() >= minConstraint {
> 			paths = append(paths, rootPath.next())
> 			return
> 		}
> ```
Cool! But I think it's better like this:
```
dfsFunc = func(sum, begin int) {
	if sum >= target && rootPath.length() >= minConstraint && rootPath.length() <= maxConstraint {
		paths = append(paths, rootPath.next())
		return
	}
	
	if rootPath.length() >= maxConstraint {
		return
	}
```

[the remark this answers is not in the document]
whitewindmills: @jwcesign It's an excellent pruning!
Thanks.

pkg/scheduler/core/spreadconstraint/select_groups.go
whitewindmills: Yeah, they should have identical meanings. `value` is not only about the number of clusters but `regions`, `zone`, etc.
eg: some zones in a region; some region in a provider...
And 'weight' is just for 'weight', not 'score', `select_groups` should be completely out of the concept of `region`, `zone` and `provider` like a tool.

jwcesign (commented):
other generally looks good to me.
cc @chaunceyjiang  @RainbowMango @XiShanYongYe-Chang for checking

pkg/scheduler/core/spreadconstraint/select_groups.go
```diff
@@ -0,0 +1,202 @@
...
+	for i, group := range subPath.groups {
+		if path.groups[i].name != group.name {
+			return false
+		}
+	}
+
+	return true
+}
+
+func (path *dfsPath) String() string {
```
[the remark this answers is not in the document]
whitewindmills: `String()` function is implicitly used by `klog`.
See [line#95](https://github.com/karmada-io/karmada/pull/3259/files#diff-44ca813838df40549f4e3bfb5bdca3bb52da32731b3fd0e7a31bd78c5461634dR95)

chaunceyjiang:
This is really great work! 👏

RainbowMango (commented):
/assign
I wonder do @jwcesign @chaunceyjiang have any other comments?

pkg/scheduler/core/spreadconstraint/select_clusters_by_region.go
```diff
@@ -35,68 +37,67 @@ func selectBestClustersByRegion(spreadConstraintMap map[policyv1alpha1.SpreadFie
...
+//
+//	R1 -> R2 -> R3 -> R4
+//
+// We assume that those regions have already done the above descending order according to
+// the score. Now those combinations([R1,R3] and [R1,R4]) can all satisfy all spread constraints.
+// Who is the best choice?
+// Obviously [R1,R3].
+//
+// Finally, the number of regions that we need should be auto-incremented when there is no
+// combination to meet cluster spread constraint. Of course it can never exceed region's
```
RainbowMango, line 74:
```suggestion
// combination to meet cluster spread constraint. Of course, it can never exceed region's
```

pkg/scheduler/core/spreadconstraint/select_groups.go
```diff
@@ -0,0 +1,202 @@
...
+	var groupNames []string
+	for _, group := range path.groups {
+		groupNames = append(groupNames, group.name)
+	}
+	return fmt.Sprintf("[%s]", strings.Join(groupNames, "->"))
+}
+
+// selectGroups select best groups in all groups.
+func selectGroups(groups []*GroupInfo, minConstraint, maxConstraint, target int) []*GroupInfo {
```
RainbowMango, line 86: The signature `target` might be not so clear enough from the code readability point of view.

XiShanYongYe-Chang (commented):
LGTM

jwcesign:
/lgtm

As discussed in the group, we need to consider DFS when dealing with multiple spread constraint types. This can be addressed in the next PR. We can proceed with merging this one for now.

RainbowMango (commented):
Looks awesome!
Only some nits.

chaunceyjiang (commented):
LGTM

RainbowMango:
@whitewindmills You might need to [Squash commits](https://karmada.io/docs/contributor/github-workflow#squash-commits).

RainbowMango (approved):
/lgtm
/approve

PR #3262, opened 2023-03-10
add managed-by label to resources created by karmada controllers

a7i:
**What type of PR is this?**
/kind feature

Adds the well-known `"app.kubernetes.io/managed-by"` label to resources created by karmada controllers

We need a way to bypass creation of some ClusterRole/Rolebindings by our OPA Gatekeeper and we don't have a consistent label to select on. Not all the resources in this PR are needed by us (mostly just needed unified auth) but wanted to keep it consistent across controllers.

N/A

```release-note
Add the reserved label `karmada.io/managed` to resources created by karmada controllers.
```

### DIFF
--- a/pkg/util/constants.go
+++ b/pkg/util/constants.go
@@ -22,6 +22,9 @@ const (
 
 	// FederatedResourceQuotaNameLabel is added to Work to specify associated FederatedResourceQuota's name.
 	FederatedResourceQuotaNameLabel = "federatedresourcequota.karmada.io/name"
+
+	// ManagedByWellKnownLabel is a well known label to specify the controller that manages the resource
+	ManagedByWellKnownLabel = "app.kubernetes.io/managed-by"
 )
 
 // Define annotations used by karmada system.
@@ -141,6 +144,9 @@ const (
 const (
 	// NamespaceKarmadaSystem is the karmada system namespace.
 	NamespaceKarmadaSystem = "karmada-system"
+
+	// Karmada is the value for managed-by label.
+	Karmada = "karmada"
 )
 
 // ContextKey is the key of context.
--- a/pkg/util/constants.go
+++ b/pkg/util/constants.go
@@ -23,8 +23,11 @@ const (
 	// FederatedResourceQuotaNameLabel is added to Work to specify associated FederatedResourceQuota's name.
 	FederatedResourceQuotaNameLabel = "federatedresourcequota.karmada.io/name"
 
-	// ManagedByWellKnownLabel is a well known label to specify the controller that manages the resource
-	ManagedByWellKnownLabel = "app.kubernetes.io/managed-by"
+	// ManagedByKarmadaLabel is a reserved karmada label to indicate whether resources are managed by karmada controllers.
+	ManagedByKarmadaLabel = "karmada.io/managed"
+
+	// ManagedByKarmadaLabelValue indicates that resources are managed by karmada controllers.
+	ManagedByKarmadaLabelValue = "true"
 )
 
 // Define annotations used by karmada system.
@@ -144,9 +147,6 @@ const (
 const (
 	// NamespaceKarmadaSystem is the karmada system namespace.
 	NamespaceKarmadaSystem = "karmada-system"
-
-	// Karmada is the value for managed-by label.
-	Karmada = "karmada"
 )
 
 // ContextKey is the key of context.
--- a/pkg/controllers/binding/common.go
+++ b/pkg/controllers/binding/common.go
@@ -137,7 +137,7 @@ func mergeLabel(workload *unstructured.Unstructured, workNamespace string, bindi
 	var workLabel = make(map[string]string)
 	util.MergeLabel(workload, workv1alpha1.WorkNamespaceLabel, workNamespace)
 	util.MergeLabel(workload, workv1alpha1.WorkNameLabel, names.GenerateWorkName(workload.GetKind(), workload.GetName(), workload.GetNamespace()))
-	util.MergeLabel(workload, util.ManagedByWellKnownLabel, util.Karmada)
+	util.MergeLabel(workload, util.ManagedByKarmadaLabel, util.ManagedByKarmadaLabelValue)
 	if scope == apiextensionsv1.NamespaceScoped {
 		util.MergeLabel(workload, workv1alpha2.ResourceBindingReferenceKey, names.GenerateBindingReferenceKey(binding.GetNamespace(), binding.GetName()))
 		workLabel[workv1alpha2.ResourceBindingReferenceKey] = names.GenerateBindingReferenceKey(binding.GetNamespace(), binding.GetName())
--- a/pkg/controllers/binding/common.go
+++ b/pkg/controllers/binding/common.go
@@ -137,6 +137,7 @@ func mergeLabel(workload *unstructured.Unstructured, workNamespace string, bindi
 	var workLabel = make(map[string]string)
 	util.MergeLabel(workload, workv1alpha1.WorkNamespaceLabel, workNamespace)
 	util.MergeLabel(workload, workv1alpha1.WorkNameLabel, names.GenerateWorkName(workload.GetKind(), workload.GetName(), workload.GetNamespace()))
+	util.MergeLabel(workload, util.ManagedByWellKnownLabel, util.Karmada)
 	if scope == apiextensionsv1.NamespaceScoped {
 		util.MergeLabel(workload, workv1alpha2.ResourceBindingReferenceKey, names.GenerateBindingReferenceKey(binding.GetNamespace(), binding.GetName()))
 		workLabel[workv1alpha2.ResourceBindingReferenceKey] = names.GenerateBindingReferenceKey(binding.GetNamespace(), binding.GetName())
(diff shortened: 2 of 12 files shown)


RainbowMango:
Firstly, it makes 100% sense to me to have a label(or annotation) on resources that are managed by Karmada.

> We need a way to bypass creation of some ClusterRole/Rolebindings by our OPA Gatekeeper and we don't have a consistent label to select on.

I agree that the label(or annotation) should be added to all types of resources, but could you please explain more about your use case? How your `OPA Gatekeeper` could bypass the creation of some RBAC resources?

RainbowMango:
Just a question, why is it not an `annotation`? I'm asking because compared to annotation, the label must satisfy some restrictions like value length can't exceed 63 characters.

chaunceyjiang:
Just one question, according to [the description ](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/)in the k8s official website documentation.


| Key | Description | Example| Type 
| -- | -- | -- | --
| app.kubernetes.io/managed-by | The tool being used to manage the operation of an application | helm | string

Why not set it like this `app.kubernetes.io/managed-by: karmada` ?

a7i:
Thanks for the feedback @RainbowMango and @chaunceyjiang ! Those are great questions/suggestions, please find my response below:

--------------

>could you please explain more about your use case?

My organization has 100+ OPA policies to ensure that the resource being created match our security, compliance, and reliability requirements. For example, one of those policies is to discourage the usage of ClusterRole/ClusterRoleBinding and instead encourage application teams to use Role/RoleBinding instead. This is a good general rule for applications as they're almost always namespace bound. 

--------------

>How your OPA Gatekeeper could bypass the creation of some RBAC resources?

However, there are cluster-level/node-level capabilities that require permissions beyond a single namespace (e.g. aws-vpc-cni, operator, and even Karmada) so we add them to an "allow-list" by using OPA's [match field](https://open-policy-agent.github.io/gatekeeper/website/docs/howto/#the-match-field) of `labelSelector` or `namespaceSelector`.

--------------

>why is it not an annotation?

Unfortunately, OPA Gatekeeper or any ValidatingWebhooks do not allow match against annotations. See [kubernetes](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-objectselector) docs.

--------------

>Why not set it uniformly like this `app.kubernetes.io/managed-by: karmada` ?

Yes, this makes perfect sense to me and would honestly be much easier for us to add a single match to our allow list. From our perspective, it doesn't really matter if it was created by unified-auth controller or service-import controller. I can go ahead and change it to karmada if you suggest.

RainbowMango:
> We may pivot from this and only run OPA Gatekeeper on the "host" cluster (per [karmada docs](https://karmada.io/docs/userguide/security-governance/working-with-gatekeeper/)). However, for the time being (and throughout the whole migration phase), we will continue to run OPA Gatekeeper at each member cluster.

I'd say, we are looking forward to hearing more feedback about running no matter OPA Gatekeeper or Kyverno, just let us know if you figured out anything that we can do. We are glad to help.

> Why not set it uniformly like this app.kubernetes.io/managed-by: karmada ?

That makes sense to me too.

RainbowMango:
Good question. 

There are two possible ways to use helm on Karmada.
1. Deploy charts on the Karmada control plane then propagate the workloads by PropagationPolicy.
2. Deploy charts with cd tools like `Flux`.  [Here is the documentation for the use case](https://karmada.io/docs/userguide/cicd/working-with-flux#helm-release-propagation)

For `1`, after the charts be installed on the Karmada control plane, the label will be `app.kubernetes.io/managed-by: helm`, but during the propagation, Karmada probably overrides the label to `app.kubernetes.io/managed-by: karmada`. So, what you see in the member cluster is `app.kubernetes.io/managed-by: karmada`.

For `2`, take Fulx as an example, the resource template targeted by PropagationPolicy is `HelmRelease`, Karmada will label the `HelmRelease` with `app.kubernetes.io/managed-by: karmada`, but the chart installation happened on member cluster. So, you will see two labels in the member cluster:
- `HelmRelease`'s label is `app.kubernetes.io/managed-by: karmada`.
- The resources extracted from the chart will be labeled by `app.kubernetes.io/managed-by: helm`. 

That reminds me to consider if we need to define the label by ourselves, like `app.karmada.io/managed-by`.
Any ideas?

chaunceyjiang:
> For 1, after the charts be installed on the Karmada control plane, the label will be app.kubernetes.io/managed-by: helm, but during the propagation, Karmada probably overrides the label to app.kubernetes.io/managed-by: karmada. So, what you see in the member cluster is app.kubernetes.io/managed-by: karmada.



Yes, I don't think this is a reasonable behavior. Karmada should not override this label.



> That reminds me to consider if we need to define the label by ourselves, like app.karmada.io/managed-by.

+1

pkg/util/constants.go
```diff
@@ -22,6 +22,9 @@ const (
 
 	// FederatedResourceQuotaNameLabel is added to Work to specify associated FederatedResourceQuota's name.
 	FederatedResourceQuotaNameLabel = "federatedresourcequota.karmada.io/name"
+
+	// ManagedByWellKnownLabel is a well known label to specify the controller that manages the resource
+	ManagedByWellKnownLabel = "app.kubernetes.io/managed-by"
```
RainbowMango, line 27:
```suggestion
	ManagedByWellKnownLabel = "app.karmada.io/managed-by"
```
I'd prefer to use the "app.`karmada`.io" instead of "app.`kubernetes`.io" as per the discussion at https://github.com/karmada-io/karmada/pull/3262#issuecomment-1489905627 and https://github.com/karmada-io/karmada/pull/3262#issuecomment-1501284793.

RainbowMango (commented):
Generally looks good to me. 
Just hope to change the label name a little bit.

pkg/util/constants.go
a7i: I was honestly hoping to go with the [Kubernetes well-known labels](https://kubernetes.io/docs/reference/labels-annotations-taints/) but for my use-case, any identifier will work.

I will have to also rename the variable/comments to no longer suggest it's a well-known label.

a7i: Also, if the key has karmada in it, "app.karmada.io/managed-b", perhaps the value being the actual controller (rather than `karmada`) may have been better.

With that being said, I have applied the changes as you requested.

i.e. `app.karmada.io/managed-by: karmada`

RainbowMango: If there is no side effect, I would love to adopt Kubernete's well-known labels, because that will be easier for users to understand their meaning.  But there is a risk that Karmada might unexpectedly overwrite the label (see https://github.com/karmada-io/karmada/pull/3262#issuecomment-1489905627).  That's the concern that drives me to go with another label.

> I will have to also rename the variable/comments to no longer suggest it's a well-known label.

Sounds good to me. The label essentially is a `reserved` label by Karmada.

> Also, if the key has karmada in it, "app.karmada.io/managed-b", perhaps the value being the actual controller (rather than karmada) may have been better.

That's definitely a good point! 
I remember the Kubefed project also [reserved a similar label](https://github.com/kubernetes-retired/kubefed/blob/d6f10d29c3785dc55abc954dec0afeddce4893ef/pkg/controller/util/managedlabel.go#L23-L27), inspire of it, maybe we can have a similar label, like:
```golang
const (
	ManagedByKarmadaLabelKey     = "karmada.io/managed"
	ManagedByKarmadaLabelValue   = "true"
	UnmanagedByKarmadaLabelValue = "false"
)
```
If we need to label a resource in the future we are still able to introduce another label(managed-by).

What do you say? @a7i @chaunceyjiang

a7i: Sounds good to me. As far as this PR is concerned, is it correct to say that I will be using the folllowing?

```
const (
	ManagedByKarmadaLabelKey     = "karmada.io/managed"
	ManagedByKarmadaLabelValue   = "true"
)
```

RainbowMango: Yes, for now, we'll probably just use the `true`.

chaunceyjiang: Currently, we are developing `karmada-operator`. With `karmada-operator`, we can declaratively create multiple karmadas.

If we have multiple karmadas, can we use this label to indicate which one manages the resources?

Like:
```
karmada.io/managed: kamrada1
karmada.io/managed: kamrada2
```

RainbowMango:
/label tide/merge-method-squash

I'm curious why the bot didn't respond to this command.  let me try.

RainbowMango (approved):
/lgtm
/approve

Thanks @a7i. 

And I'll continue to figure out the bot issue.

RainbowMango (approved):
/lgtm
/approve

Thanks @a7i. 

And I'll continue to figure out the bot issue.

pkg/util/constants.go
RainbowMango: Sorry, I missed your comments.
> Currently, we are developing karmada-operator. With karmada-operator, we can declaratively create multiple karmadas.

But I guess we can have another more straightforward label for the scenario. `karmada.io/managed` only indicates a resource is managed or not.

RainbowMango:
At first, I thought the same way. However, I believe both myself and the robot should have sufficient permissions. Let me investigate it carefully when I have time.

PR #3283, opened 2023-03-16
Avoid the same cluster appearing in both spec.clusters and spec.gracefulEvictionTasks

Poor12:
**What type of PR is this?**
/kind bug

Avoid the same cluster appearing in both spec.clusters and spec.gracefulEvictionTasks.
There are basicly two solutions:
1. Adjust the scheduler. Make sure that scheduler will not schedule resource templates to those cluster in gracefulEvictionTasks.
2. Adjust the controller. For those clusters that appear at the same time, keep the clusters in spec.clusters and delete the clusters in spec.gracefulEvictionTask.

When the scheduler can dispatch to a cluster in gracefulEvictionTasks, it means that the cluster has a high probability of recovering from the abnormal state when it was evicted. From this point of view, option 2 is a better choise.

Fixes #

None

### DIFF
--- a/pkg/controllers/gracefuleviction/evictiontask.go
+++ b/pkg/controllers/gracefuleviction/evictiontask.go
@@ -48,6 +48,13 @@ func assessEvictionTasks(bindingSpec workv1alpha2.ResourceBindingSpec,
 }
 
 func assessSingleTask(task workv1alpha2.GracefulEvictionTask, opt assessmentOption) *workv1alpha2.GracefulEvictionTask {
+	// avoid the same cluster appearing in both spec.clusters and spec.gracefulEvictionTasks
+	for _, targetCluster := range opt.scheduleResult {
+		if task.FromCluster == targetCluster.Name {
+			return nil
+		}
+	}
+
 	if task.SuppressDeletion != nil {
 		if *task.SuppressDeletion {
 			return &task
--- a/pkg/controllers/gracefuleviction/evictiontask_test.go
+++ b/pkg/controllers/gracefuleviction/evictiontask_test.go
@@ -210,6 +210,25 @@ func Test_assessSingleTask(t *testing.T) {
 			},
 			want: nil,
 		},
+		{
+			name: "cluster appears both in targetClusters and gracefulEvictionTasks",
+			args: args{
+				task: workv1alpha2.GracefulEvictionTask{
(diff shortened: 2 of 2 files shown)


XiShanYongYe-Chang (commented):
How about adding a ut to cover this case?

I think it's more like a cleanup.

RainbowMango:
> Avoid the same cluster appearing in both spec.clusters and spec.gracefulEvictionTasks.

Could you please explain under what circumstances would cause this issue before talking about the solution?

Poor12:
I agree. It's more like a cleanup, not a bug.
At present, it mainly occurs when the cluster fails and recovers quickly, and the application cannot quickly reach the Healthy state. It will also be needed for failover of the application.
/remove-kind bug
/kind cleanup

Poor12:
When we use application failover and we do not set blockPredecessorSeconds, there is a high probability that when a cluster is evicted, the application is still scheduled to the evicted cluster. In that case, we need to avoid the same cluster appearing in both spec.clusters and spec.gracefulEvictionTasks.

RainbowMango (commented):
> Could you please help trigger the e2e again?

Done. 
/assign @XiShanYongYe-Chang

XiShanYongYe-Chang (commented):
/lgtm
Thanks~

RainbowMango (approved):
/approve

RainbowMango:
> When the scheduler can dispatch to a cluster in gracefulEvictionTasks, it means that the cluster has a high probability of recovering from the abnormal state when it was evicted. From this point of view, option 2 is a better choise.

Given the eviction process is based on `taint-toleration`, the scheduler selecting a cluster doesn't mean the cluster is recovered from the abnormal state, it is also possible that the resource referenced by the RB tolerates the taint of the cluster, in that case, there will be an endless loop, like:
1. The scheduler selects a cluster(A) because the RB tolerates the taint
2. The taint-manager evicts cluster A from RB after `TolerationSeconds` and build a graceful eviction task
3. The scheduler adds cluster A back to RB
4. The `resource-binding-graceful-eviction-controller` remove the graceful eviction task added in step 2.

I tend to let the scheduler skip the cluster if it appears in eviction tasks, just as @jwcesign mentioned at https://github.com/karmada-io/karmada/pull/3456#issuecomment-1522660388.

PR #3288, opened 2023-03-16
bugfix: runtime.Object apiVersion/kind is empty 

### LINKED ISSUE #3275: ClusterWideKeyFunc don't get runtime.Object GVK
**What happened**:
call ClusterWideKeyFunc() method don't get runtime.Object GVK information.

code: https://github.com/karmada-io/karmada/blob/master/pkg/util/fedinformer/keys/keys.go#L63
```
// ClusterWideKeyFunc generates a ClusterWideKey for object.
func ClusterWideKeyFunc(obj interface{}) (ClusterWideKey, error) {
	key := ClusterWideKey{}

	runtimeObject, ok := obj.(runtime.Object)
	if !ok {
		klog.Errorf("Invalid object")
		return key, fmt.Errorf("not runtime object")
	}

	metaInfo, err := meta.Accessor(obj)
	if err != nil { // should not happen
		return key, fmt.Errorf("object has no meta: %v", err)
	}

	gvk := runtimeObject.GetObjectKind().GroupVersionKind() // this gvk might be empty.
	key.Group = gvk.Group
	key.Version = gvk.Version
	key.Kind = gvk.Kind
	key.Namespace = metaInfo.GetNamespace()
	key.Name = metaInfo.GetName()

	return key, nil
}
```

Test code output：
image

Expect to get GVK information from runtime.Object

Test code:
```
package main

import (
	"context"
	"encoding/json"
	"flag"
	"fmt"
	"path/filepath"

	apiv1 "k8s.io/api/core/v1"
	"k8s.io/apimachinery/pkg/api/meta"
	"k8s.io/apimachinery/pkg/runtime"
	"k8s.io/client-go/kubernetes"
	"k8s.io/client-go/tools/clientcmd"
	"k8s.io/client-go/util/homedir"
	"k8s.io/klog/v2"
)

// ClusterWideKey is the object key which is a unique identifier under a cluster, across all resources.
type ClusterWideKey struct {
	// Group is the API Group of resource being referenced.
	Group string

	// Version is the API Version of the resource being referenced.
	Version string

	// Kind is the type of resource being referenced.
	Kind string

	// Namespace is the name of a namespace.
	Namespace string

	// Name is the name of resource being referenced.
	Name string
}

// ClusterWideKeyFunc generates a ClusterWideKey for object.
func ClusterWideKeyFunc(obj interface{}) (ClusterWideKey, error) {
	key := ClusterWideKey{}

	runtimeObject, ok := obj.(runtime.Object)
	if !ok {
		klog.Errorf("Invalid object")
		return key, fmt.Errorf("not runtime object")
	}

	metaInfo, err := meta.Accessor(obj)
	if err != nil { // should not happen
		return key, fmt.Errorf("object has no meta: %v", err)
	}

	gvk := runtimeObject.GetObjectKind().GroupVersionKind()
	key.Group = gvk.Group
	key.Version = gvk.Version
	key.Kind = gvk.Kind
	key.Namespace = metaInfo.GetNamespace()
	key.Name = metaInfo.GetName()

	return key, nil
}

func main() {
	var kubeconfig *string
	if home := homedir.HomeDir(); home != "" {
		kubeconfig = flag.String("kubeconfig", filepath.Join(home, ".kube", "config"), "(optional) absolute path to the kubeconfig file")
	} else {
		kubeconfig = flag.String("kubeconfig", "", "absolute path to the kubeconfig file")
	}
	flag.Parse()

	config, err := clientcmd.BuildConfigFromFlags("", *kubeconfig)
	if err != nil {
		panic(err)
	}
	clientset, err := kubernetes.NewForConfig(config)
	if err != nil {
		panic(err)
	}

	deploymentsClient := clientset.AppsV1().Deployments(apiv1.NamespaceDefault)

	// List Deployments
	fmt.Printf("Listing deployments in namespace %q:\n", apiv1.NamespaceDefault)
	list, err := deploymentsClient.List(context.TODO(), metav1.ListOptions{})
	if err != nil {
		panic(err)
	}
	for _, d := range list.Items {
		objectKey, err := ClusterWideKeyFunc(&d)
		if err != nil {
			fmt.Println(err)
			return
		}

		data, _ := json.MarshalIndent(&objectKey, " ", "   ")
		fmt.Printf("object name: %v key: %v\n", d.Name, string(data))
	}

}
```

xigang:
**What type of PR is this?**
/kind cleanup

When using a typed client, decoding to a versioned struct (not an internal API type), the apiVersion/kind information will be dropped. Therefore, the APIVersion/Kind information of runtime.Object needs to be verified.
See issue: https://github.com/kubernetes/kubernetes/issues/80609

Fixes #3275

@RainbowMango @whitewindmills 

```
NONE
```

### DIFF
--- a/pkg/util/fedinformer/keys/keys.go
+++ b/pkg/util/fedinformer/keys/keys.go
@@ -74,7 +74,18 @@ func ClusterWideKeyFunc(obj interface{}) (ClusterWideKey, error) {
 		return key, fmt.Errorf("object has no meta: %v", err)
 	}
 
+	// When using a typed client, decoding to a versioned struct (not an internal API type), the apiVersion/kind
+	// information will be dropped. Therefore, the APIVersion/Kind information of runtime.Object needs to be verified.
+	// See issue: https://github.com/kubernetes/kubernetes/issues/80609
 	gvk := runtimeObject.GetObjectKind().GroupVersionKind()
+	if len(gvk.Kind) == 0 {
+		return key, fmt.Errorf("runtime object has no kind")
+	}
+
+	if len(gvk.Version) == 0 {
+		return key, fmt.Errorf("runtime object has no version")
+	}
+
 	key.Group = gvk.Group
 	key.Version = gvk.Version
 	key.Kind = gvk.Kind
--- a/pkg/util/fedinformer/keys/keys_test.go
+++ b/pkg/util/fedinformer/keys/keys_test.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"testing"
 
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -50,6 +51,12 @@ var (
 			Name:      "bar",
 		},
 	}
+	deploymentObj = appsv1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: "foo",
+			Name:      "bar",
+		},
+	}
 )
 
 func TestClusterWideKeyFunc(t *testing.T) {
@@ -91,6 +98,11 @@ func TestClusterWideKeyFunc(t *testing.T) {
 		{
 			name:      "nil object should be error",
 			object:    nil,
+			expectErr: true,
+		},
+		{
+			name:      "non APIVersion and kind runtime object should be error",
+			object:    deploymentObj,
 			expectErr: true,
 		},
 	}


chaunceyjiang:
Just a question,will this change break compatibility?

RainbowMango:
> Just a question,will this change break compatibility?

Good question.
My understanding is we don't provide compatibility guarantees at `function level` right now. 
Speaking of this change, given the `ClusterWideKeyFunc` aims to generate cluster-wide-key, the `apiversion` and `kind` are both necessary filed. So I think this patch is a kind of protection, it doesn't change the interface.

RainbowMango (commented):
cc @chaunceyjiang Do you have any other comments?

/lgtm

pkg/util/fedinformer/keys/keys.go
```diff
@@ -74,7 +74,18 @@ func ClusterWideKeyFunc(obj interface{}) (ClusterWideKey, error) {
 		return key, fmt.Errorf("object has no meta: %v", err)
 	}
 
+	// When using a typed client, decoding to a versioned struct (not an internal API type), the apiVersion/kind
```
Garrybest, line 77: Is APIVersion/Kind information of `runtime.Object` filled in when we using dynamic client?

RainbowMango (approved):
/approve

Thanks, @xigang and welcome to Karmada.

PR #3290, opened 2023-03-17
Fix cluster_locality plugin when replicaScheduling is nil

### LINKED ISSUE #3289: Scheduling behavior is inconsistent when replicaScheduling is not filled
Assume there are two policies which the meaning is exactly the same:
```
placement:
    clusterAffinity:
      clusterNames:
      - member1
      - member2
    replicaScheduling:
      replicaSchedulingType: Duplicated
   spreadConstraints:
      - maxGroups: 1
         minGroups: 1
```

```
placement:
    clusterAffinity:
      clusterNames:
      - member1
      - member2
     spreadConstraints:
      - maxGroups: 1
         minGroups: 1
```

Now the scheduler schedules the workload to member1. But I edit the spec.clusters of resourceBinding and hope it will schedule to member2. For the first policy, it will schedule to member2. For the second policy, it will not.

Through analysis, we find that `clusterLocality` does not work for the second policy because replicas == 0.
```
func (p *ClusterLocality) Score(ctx context.Context, placement *policyv1alpha1.Placement,
	spec *workv1alpha2.ResourceBindingSpec, cluster *clusterv1alpha1.Cluster) (int64, *framework.Result) {
	if len(spec.Clusters) == 0 {
		return framework.MinClusterScore, framework.NewResult(framework.Success)
	}

	replicas := util.GetSumOfReplicas(spec.Clusters)
	if replicas <= 0 {
		return framework.MinClusterScore, framework.NewResult(framework.Success)
	}

	if spec.TargetContains(cluster.Name) {
		return framework.MaxClusterScore, framework.NewResult(framework.Success)
	}

	return framework.MinClusterScore, framework.NewResult(framework.Success)
}
```

I think when I don't fill in the strategy it's implied that we are using duplicated strategy, so the number of replicas is necessary if the workload has.

Poor12:
**What type of PR is this?**
/kind cleanup

Cluster_locality can work normally for placement without replicaSchedulingStrategy.

~~Do not fill replicas in bindings when replicaScheduling is duplicated. Keep the schedule behavior consistent with replicaScheduling is nil.~~

Fixes #3289 

None

### DIFF
--- a/pkg/scheduler/framework/plugins/clusterlocality/cluster_locality.go
+++ b/pkg/scheduler/framework/plugins/clusterlocality/cluster_locality.go
@@ -6,7 +6,6 @@ import (
 	clusterv1alpha1 "github.com/karmada-io/karmada/pkg/apis/cluster/v1alpha1"
 	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
 	"github.com/karmada-io/karmada/pkg/scheduler/framework"
-	"github.com/karmada-io/karmada/pkg/util"
 )
 
 const (
@@ -35,11 +34,6 @@ func (p *ClusterLocality) Name() string {
 func (p *ClusterLocality) Score(ctx context.Context,
 	spec *workv1alpha2.ResourceBindingSpec, cluster *clusterv1alpha1.Cluster) (int64, *framework.Result) {
 	if len(spec.Clusters) == 0 {
-		return framework.MinClusterScore, framework.NewResult(framework.Success)
-	}
-
-	replicas := util.GetSumOfReplicas(spec.Clusters)
-	if replicas <= 0 {
 		return framework.MinClusterScore, framework.NewResult(framework.Success)
 	}
 


RainbowMango (commented):
/assign

pkg/scheduler/core/assignment.go
```diff
@@ -54,6 +54,9 @@ type assignState struct {
 func newAssignState(candidates []*clusterv1alpha1.Cluster, strategy *policyv1alpha1.ReplicaSchedulingStrategy, obj *workv1alpha2.ResourceBindingSpec) *assignState {
 	var strategyType string
 
+	if strategy == nil {
+		strategy = &policyv1alpha1.ReplicaSchedulingStrategy{ReplicaSchedulingType: DuplicatedStrategy}
+	}
```
RainbowMango, line 59: Hi @Garrybest 

Do you remember why we introduce `Duplicated` for `replicaSchedulingType`?
https://github.com/karmada-io/karmada/blob/a14d64f055e1011cd774a76d607c4696fb75bc87/pkg/apis/policy/v1alpha1/propagation_types.go#L300-L308

I don't know when people will set the replicaSchedulingType with `Duplicated`, can you remind me?

Garrybest: If `Duplicated` is set, a workload must be dispatched to all member clusters. The effect is as same as that `strategy` is set to `nil`. 

BTW, I don't think this design is decent, now we have a lot of if-else judgement in many places. I would like to add a bool field to replace `ReplicaSchedulingType` to represent whether we should divide replicas.

RainbowMango: Yeah, I'm thinking if we can make some improvements from the API.

Now I suppose if `ReplicaSchedulingStrategy` is set means we are going to `split` replicas, conversely(not set) means we are going to `delicates` replicas. Am I right?

If so, actually we can remove the `ReplicaSchedulingType` from `ReplicaSchedulingStrategy` I think. But we should consider giving `ReplicaSchedulingStrategy` a more clear name.

Garrybest: Exactly, just as I thought.

RainbowMango: Given the fact that we probably not going to promote the API version for a single field of optimization, like this.

I think we need a place to address the changes gonna make for the next version. My idea is to set up the next version API in the codebase, like `pkg/apis/policy/v1alpha2`, but we don't switch to a newer version until enough optimizations are accumulated. What do you think?

Garrybest: It's great. We'll switch the API when the time is right.

RainbowMango: OK. I can do that. And will talk about it at the community meeting before promoting it.

pkg/scheduler/framework/plugins/clusterlocality/cluster_locality.go
```diff
@@ -38,11 +37,6 @@ func (p *ClusterLocality) Score(ctx context.Context,
 		return framework.MinClusterScore, framework.NewResult(framework.Success)
 	}
 
-	replicas := util.GetSumOfReplicas(spec.Clusters)
-	if replicas <= 0 {
-		return framework.MinClusterScore, framework.NewResult(framework.Success)
-	}
```
RainbowMango, line 44: I think removing this check would fix the issue tranced by https://github.com/karmada-io/karmada/pull/3290, right?

RainbowMango: I suggest letting this PR focus on this issue.

Poor12: > I think removing this check would fix the issue tranced by #3290, right?

Yes, do we have a plan to unify the form whether duplicated is explict or implicit?

RainbowMango: Yes we can do that, but please clarify the benefits in the next PR and we can discuss it there.
My idea is don't mix too many things in a PR, that will complicate things.

RainbowMango (approved):
/lgtm
/approve

PR #3309, opened 2023-03-22
Using schema.GroupVersionKind instead of Kind when determining object types

### LINKED ISSUE #3011: Using `schema.GroupKind` instead of Kind when determining object types
**What happened**:
Now we determine the object types only by their kind, but sometimes different groups shared with same kind could have a conflict. Here is some reference code here:

https://github.com/karmada-io/karmada/blob/4d9b23ff729a5f8b41263d24a389c48c59a6d69a/pkg/resourceinterpreter/defaultinterpreter/prune/prune.go#L62

https://github.com/karmada-io/karmada/blob/4d9b23ff729a5f8b41263d24a389c48c59a6d69a/pkg/estimator/server/replica/replica.go#L35-L36

https://github.com/karmada-io/karmada/blob/4d9b23ff729a5f8b41263d24a389c48c59a6d69a/pkg/resourceinterpreter/defaultinterpreter/prune/prune.go#L53

I suppose that it's better to use `GroupVersionKind().GroupKind()` to replace these judgement.

**Environment**:
- Karmada version:
- Others:

Poor12:
**What type of PR is this?**
/kind cleanup

Using schema.GroupVersionKind instead of Kind when determining object types.
I'd like to user GroupVersionKind rather than GroupKind because we finally use the resource objects with specific version.

Fixes #3011 

None

### DIFF
--- a/pkg/controllers/binding/common.go
+++ b/pkg/controllers/binding/common.go
@@ -1,6 +1,7 @@
 package binding
 
 import (
+	batchv1 "k8s.io/api/batch/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -39,7 +40,7 @@ func ensureWork(
 
 	var jobCompletions []workv1alpha2.TargetCluster
 	var err error
-	if workload.GetKind() == util.JobKind {
+	if workload.GroupVersionKind() == batchv1.SchemeGroupVersion.WithKind(util.JobKind) {
 		jobCompletions, err = divideReplicasByJobCompletions(workload, targetClusters)
 		if err != nil {
 			return err
--- a/pkg/controllers/mcs/service_export_controller.go
+++ b/pkg/controllers/mcs/service_export_controller.go
@@ -149,14 +149,14 @@ func (c *ServiceExportController) syncServiceExportOrEndpointSlice(key util.Queu
 
 	klog.V(4).Infof("Begin to sync %s %s.", fedKey.Kind, fedKey.NamespaceKey())
 
-	switch fedKey.Kind {
-	case util.ServiceExportKind:
+	switch fedKey.GroupVersionKind() {
+	case serviceExportGVK:
 		if err := c.handleServiceExportEvent(fedKey); err != nil {
 			klog.Errorf("Failed to handle serviceExport(%s) event, Error: %v",
 				fedKey.NamespaceKey(), err)
 			return err
 		}
-	case util.EndpointSliceKind:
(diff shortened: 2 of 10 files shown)


pkg/controllers/binding/common.go
```diff
@@ -39,7 +40,7 @@ func ensureWork(
 
 	var jobCompletions []workv1alpha2.TargetCluster
 	var err error
-	if workload.GetKind() == util.JobKind {
+	if workload.GroupVersionKind() == batchv1.SchemeGroupVersion.WithKind(util.JobKind) {
```
RainbowMango, line 43: `GroupVersionKind` or `GroupKind`?

> I suppose that it's better to use GroupVersionKind().GroupKind() to replace these judgement.

Echo from https://github.com/karmada-io/karmada/issues/3011, I think what @Garrybest means is `GroupKind`.

Poor12: I'd like to user GroupVersionKind rather than GroupKind because we finally use the resource objects with specific version.
I'm afraid there will also be previous problems regardless of the version.  /cc @Garrybest for opinions.

Garrybest: > Echo from #3011, I think what @Garrybest means is `GroupKind`.

+1, should we care about version? I thought that an object has different versions but we only care about group kind here. I guess we should handle both `batch/v1beta1` and `batch/v1` here.

Poor12: Take an extreme example:
```
if workload.GroupVersionKind() == batchv1.SchemeGroupVersion.WithKind(util.JobKind) {
        job := &batchv1.Job{}
	err := helper.ConvertToTypedObject(workload, job)
        ....
}
```
In many cases, we judge the kind of workload and convert it into specific version of resource(e.g. batchv1/job). If the two versions of a resource have obvious field changes, does it mean that the converted Job has missing content? We cannot guarantee that these missing fields will not affect subsequent logical judgments.

RainbowMango: > I guess we should handle both batch/v1beta1 and batch/v1 here.

> In many cases, we judge the kind of workload and convert it into specific version of resource

Both points make sense.
Changing `Kind` to `GroupKind` doesn't change anything, but changing to `GroupVersionKind` would reduce the range of supported API.

> If the two versions of a resource have obvious field changes, does it mean that the converted Job has missing content?

We should carefully evaluate every change. Please @Poor12 check if there are obvious field changes.

Garrybest: Interesting, I realize the `GroupVersionKind` of workload comes from [preferred resources of detector](https://github.com/karmada-io/karmada/blob/d349f001db4551ffa814fcf78086798b270209c2/pkg/util/lifted/discovery.go#L39). So I guess if a preferred version is used here, e.g. `batch/v1` rather than `batch/v1beta1`, determination by GVK is fine.

Poor12: Yes, I remembered that for multi-version resources, now we will only create the preferred version of the Informer.

PR #3314, opened 2023-03-22
Implement group score calculation

whitewindmills:
**What type of PR is this?**
/kind feature

separated from https://github.com/karmada-io/karmada/pull/3259
Fixes #

```release-note
```

### DIFF
--- a/pkg/scheduler/core/spreadconstraint/group_clusters.go
+++ b/pkg/scheduler/core/spreadconstraint/group_clusters.go
@@ -16,14 +16,16 @@ type GroupClustersInfo struct {
 	// Clusters from global view, sorted by cluster.Score descending.
 	Clusters []ClusterDetailInfo
 
+	// The average score of all cluster.
+	averageScore int64
+
 	calAvailableReplicasFunc func(clusters []*clusterv1alpha1.Cluster, spec *workv1alpha2.ResourceBindingSpec) []workv1alpha2.TargetCluster
 }
 
 // ProviderInfo indicate the provider information
 type ProviderInfo struct {
-	Name              string
-	Score             int64 // the highest score in all clusters of the provider
-	AvailableReplicas int64
+	Name  string
+	Score int64 // the comprehensive score in all clusters of the provider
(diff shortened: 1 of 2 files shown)


pkg/scheduler/core/spreadconstraint/group_clusters.go
```diff
@@ -130,7 +133,42 @@ func (info *GroupClustersInfo) generateClustersInfo(clustersScore framework.Clus
...
+	var selectedNum int
+	for _, cluster := range clusters {
+		if cluster.Score > info.averageScore {
+			highScoreSum += cluster.Score
+			selectedNum++
+		}
+	}
+	var baseScore int64
+	if selectedNum == 0 {
+		baseScore = clusters[0].Score
```
chaunceyjiang, line 148: Will there be a situation where len(clusters) equals 0 here?

whitewindmills: Never.
If there is no cluster, no region/zone/provider does exist.

pkg/scheduler/core/spreadconstraint/group_clusters.go
```diff
@@ -130,7 +133,42 @@ func (info *GroupClustersInfo) generateClustersInfo(clustersScore framework.Clus
...
+	} else {
+		baseScore = highScoreSum / int64(selectedNum)
+	}
+
+	// Second, calculate the bonus score.
+	var bonusScore int64
+	if rbSpec.Placement == nil ||
```
chaunceyjiang, line 155: Why does the calculation of Group's score still depend on the scheduling type?

whitewindmills: Avaiable replicas of every cluster still affects the score of a group, especially `Duplicated` strategy.
eg: The group `G1` has the highest score but low avaiable replicas so that no cluster can meet `rbSepc.replicas` when using `Duplicated` strategy. Should `G1` still keep the highest score?

PR #3315, opened 2023-03-22
parseSingle: add test

helen-frank:
**What type of PR is this?**
/kind feature

The logic in the switch statement is changed to the number of parts (split into multiple strings) to make the code clearer and clearer.
When dealing with the situation "<group>/<version>", the situation with the version number of "v1" is handled separately from other situations to avoid too many nested statements and duplicate codes.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/util/apigroup_test.go
+++ b/pkg/util/apigroup_test.go
@@ -1,6 +1,7 @@
 package util
 
 import (
+	"fmt"
 	"testing"
 
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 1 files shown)


RainbowMango:
Can you help to add a unit test for it?

helen-frank:
> Can you help to add a unit test for it?

I saw that the Parse function has tests, so I didn't add unit tests to the parseSingle.
Okay, I'll add unit tests

pkg/util/apigroup.go
```diff
@@ -71,64 +71,26 @@ func (r *SkippedResourceConfig) Parse(c string) error {
...
-		if strings.HasPrefix(token, "v1") {
-			var kinds []string
-			for _, k := range strings.Split(token, ",") {
-				if strings.Contains(k, "/") { // "v1/<kind>"
-					s := strings.Split(k, "/")
-					kinds = append(kinds, s[1])
-				} else {
-					kinds = append(kinds, k)
-				}
+		r.GroupVersionKinds[schema.GroupVersionKind{Group: token}] = struct{}{}
```
XiShanYongYe-Chang, line 78: Compared with the original logic, this line is newly added. What is the reason?

pkg/util/apigroup_test.go
```diff
@@ -240,3 +241,134 @@ func TestDefaultSkippedResourceConfigGroupVersionKindParse(t *testing.T) {
...
 	}
 }
+
+func TestParseSingle(t *testing.T) {
+	tests := []struct {
+		input           string
+		expectedErr     error
+		expectedGroups  []string
+		expectedGroupVs []schema.GroupVersion
+		expectedGvks    []schema.GroupVersionKind
```
XiShanYongYe-Chang, line 251: Suggest renaming: `expectedGVKs`

pkg/util/apigroup_test.go
```diff
@@ -240,3 +241,134 @@ func TestDefaultSkippedResourceConfigGroupVersionKindParse(t *testing.T) {
...
+
+	for i, test := range tests {
+		t.Run(fmt.Sprintf("Test case %d", i+1), func(t *testing.T) {
+			// Create a new SkippedResourceConfig
+			r := NewSkippedResourceConfig()
+
+			// Call parseSingle with the input
+			err := r.parseSingle(test.input)
+
+			if !isErrorMsgEqual(err, test.expectedErr) {
```
XiShanYongYe-Chang, line 338: Check whether the err exists will be ok, just like other test case.

pkg/util/apigroup.go
helen-frank: Sorry, I missed the GroupVersionKinds note.

pkg/util/apigroup_test.go
helen-frank: ok

pkg/util/apigroup_test.go
helen-frank: I designed a situation for it that might make mistakes...

XiShanYongYe-Chang:
Thanks~
/lgtm

I think it's more like a cleanup.
/kind cleanup
/remove-kind feature

RainbowMango (commented):
/assign

pkg/util/apigroup_test.go
RainbowMango: > I designed a situation for it that might make mistakes...

What?

I agree with @XiShanYongYe-Chang that a `boolean` flag should be ok since the `parseSingle` only returns one error, so we don't need to check if it is the expected one.

RainbowMango (commented):
I can get the idea of this patch that merges the expectation case of `core` group into one branch. That indeed can simplify the code, but from the readability point of view, I tend to the previous one.

Please reminds me if I missed other benefits.  Thanks for the unit test part, we can keep the test.

helen-frank:
Error checking is necessary, unless the previous error example is deleted, it is not enough to simply determine whether an error has occurred.  @RainbowMango 
(I'm sorry, the recovery has been slow due to the busy work recently.

helen-frank:
For my optimized code, I think I made it faster, after all, it reduced a lot of unnecessary loops

RainbowMango (commented):
No worries, take your time.

> Error checking is necessary, unless the previous error example is deleted, it is not enough to simply determine whether an error has occurred.

Check or not is no big deal, my feeling is the `isErrorMsgEqual` seems unusual, as I mentioned above, just checking if there is an error happened would be more clear.

> For my optimized code, I think I made it faster, after all, it reduced a lot of unnecessary loops

Given this function only be called during startup, how much performance improved from it? 
I tend to code readability more than performance in this case.

helen-frank:
0 is optimized, 1 is original, it seems that there is little change, it is up to you to decide
```bash
go test -benchmem  -bench 'BenchmarkTestParseSingle*' github.com/karmada-io/karmada/pkg/util -benchtime=5s
goos: darwin
goarch: arm64
pkg: github.com/karmada-io/karmada/pkg/util
BenchmarkTestParseSingle0/Group_version_and_kind-8              44846637               120.0 ns/op            64 B/op          2 allocs/op
BenchmarkTestParseSingle0/Multiple_kinds-8                      35999512               166.9 ns/op            80 B/op          2 allocs/op
BenchmarkTestParseSingle0/Only_kind-8                           63702043                94.78 ns/op           48 B/op          2 allocs/op
BenchmarkTestParseSingle0/Multiple_kinds_without_group-8        44153196               135.9 ns/op            64 B/op          2 allocs/op
BenchmarkTestParseSingle1/Group_version_and_kind-8              40744158               147.4 ns/op            80 B/op          3 allocs/op
BenchmarkTestParseSingle1/Multiple_kinds-8                      26034458               231.5 ns/op           128 B/op          4 allocs/op
BenchmarkTestParseSingle1/Only_kind-8                           48660369               122.6 ns/op            64 B/op          3 allocs/op
BenchmarkTestParseSingle1/Multiple_kinds_without_group-8        29428203               203.6 ns/op           112 B/op          4 allocs/op
PASS
ok      github.com/karmada-io/karmada/pkg/util  58.296s
```

helen-frank:
> Check or not is no big deal, my feeling is the `isErrorMsgEqual` seems unusual, as I mentioned above, just checking if there is an error happened would be more clear.

got it, I will remove error handling

RainbowMango:
I don't think performance should be the motivation for this optimization, I'm more focused on readability.
By the way, the error [handling issue](https://github.com/karmada-io/karmada/pull/3315#issuecomment-1491265362) hasn't been addressed.

helen-frank:
> I don't think performance should be the motivation for this optimization, I'm more focused on readability. By the way, the error [handling issue](https://github.com/karmada-io/karmada/pull/3315#issuecomment-1491265362) hasn't been addressed.

ok, please wait a moment

RainbowMango (approved):
/lgtm
/approve
Thanks.

Issue #3313, opened 2023-03-22
No need to call setTransitionTime before calling the SetStatusCondition method in syncClusterStatus.


xigang:
**What happened**:
There is no need to call setTransitionTime before calling the SetStatusCondition method. From the code logic, call setTransitionTime is not required.

https://github.com/karmada-io/karmada/blob/master/pkg/controllers/status/cluster_status_controller.go#L180 and https://github.com/karmada-io/karmada/blob/master/pkg/controllers/status/cluster_status_controller.go#L205

**What you expected to happen**:

The setTransitionTime method should be removed.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:
```
NONE
```

dddddai:
Yes I think so

XiShanYongYe-Chang:
Hi @xigang, do you mean that calling `setTransitionTime ` is redundant because it has been handled in `SetStatusCondition`?

If I understand correctly, this is a cleanup, not a bug.

xigang:
> Hi @xigang, do you mean that calling `setTransitionTime ` is redundant because it has been handled in `SetStatusCondition`?
> 
> If I understand correctly, this is a cleanup, not a bug.

@XiShanYongYe-Chang  Yes, calling setTransitionTime is redundant. It's not a bug.

XiShanYongYe-Chang:
Thanks~ Would you like to send a pr to remove it?

/kind cleanup
/remove-kind bug

xigang:
> Thanks~ Would you like to send a pr to remove it?
> 
> /kind cleanup /remove-kind bug

okay, I submit a PR.

Garrybest:
+1

RainbowMango:
+1
/assign @xigang 
Thanks.

PR #3316, opened 2023-03-22
Fix: remove the setTransitionTime from cluster status controller.

xigang:
**What type of PR is this?**

/kind cleanup

There is no need to call setTransitionTime before calling the SetStatusCondition method. From the code logic, call setTransitionTime is not required.

https://github.com/karmada-io/karmada/blob/master/pkg/controllers/status/cluster_status_controller.go#L180 and https://github.com/karmada-io/karmada/blob/master/pkg/controllers/status/cluster_status_controller.go#L205

Fixes #https://github.com/karmada-io/karmada/issues/3313

@XiShanYongYe-Chang 

```
NONE
```

### DIFF
--- a/pkg/controllers/status/cluster_status_controller.go
+++ b/pkg/controllers/status/cluster_status_controller.go
@@ -177,7 +177,6 @@ func (c *ClusterStatusController) syncClusterStatus(cluster *clusterv1alpha1.Clu
 	if !online && readyCondition.Status != metav1.ConditionTrue {
 		klog.V(2).Infof("Cluster(%s) still offline after %s, ensuring offline is set.",
 			cluster.Name, c.ClusterFailureThreshold.Duration)
-		setTransitionTime(cluster.Status.Conditions, readyCondition)
 		meta.SetStatusCondition(&currentClusterStatus.Conditions, *readyCondition)
 		return c.updateStatusIfNeeded(cluster, currentClusterStatus)
 	}
@@ -202,7 +201,6 @@ func (c *ClusterStatusController) syncClusterStatus(cluster *clusterv1alpha1.Clu
 		}
 	}
 
-	setTransitionTime(currentClusterStatus.Conditions, readyCondition)
 	meta.SetStatusCondition(&currentClusterStatus.Conditions, *readyCondition)
 
 	return c.updateStatusIfNeeded(cluster, currentClusterStatus)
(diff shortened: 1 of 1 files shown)


Poor12 (commented):
Could you delete `setTransitionTime` together? Seems this function is non-used.

xigang:
> Could you delete `setTransitionTime` together? Seems this function is non-used.

`setTransitionTime` has been deleted

RainbowMango (commented):
Can you have a test on it to see if the condition transition time is set as expected?

It should be ok, but one more test would be better.

XiShanYongYe-Chang (commented):
Thanks~
/lgtm

XiShanYongYe-Chang:
Have you make a test on your local site?

xigang:
> Have you made a test on your local site?

I add some unit test cases

> Have you made a test on your local site?
I'll add some unit test cases a little later.

XiShanYongYe-Chang:
Hi, @xigang thanks for your work. I think we don't need to add ut to the method in the third-party library, but judging from your new ut context, what @RainbowMango says is satisfied.

Can you help remove the newly added ut, and we can continue to approve?

pkg/util/condition_test.go
```diff
@@ -56,3 +58,102 @@ func TestIsConditionsEqual(t *testing.T) {
...
+					LastTransitionTime: metav1.Unix(int64(1679496434), int64(0)),
+				},
+			},
+			expected: true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			meta.SetStatusCondition(tt.ouputConditions, tt.inputCondition)
```
RainbowMango, line 153: We don't need this unit test, as this function is maintained by Kubernetes and there is already a unit test at:
https://github.com/kubernetes/kubernetes/blob/d2be69ac11346d2a0fab8c3c168c4255db99c56f/staging/src/k8s.io/apimachinery/pkg/api/meta/conditions_test.go#L27

RainbowMango (changes requested):
Sorry, I didn't make it clear.

> Can you have a test on it to see if the condition transition time is set as expected?

I mean we can test it manually would be fine.  No worries, I can help do it.
But, please remove the unit test.

xigang:
@XiShanYongYe-Chang @RainbowMango The newly added unit tests have been removed. I only have the local environment, can you help verify it in your test environment?

RainbowMango (approved):
It works. Thanks.
/lgtm
/approve

Ready condition:
```yaml
  - lastTransitionTime: "2023-03-23T02:01:33Z"
    message: cluster is healthy and ready to accept workloads
    reason: ClusterReady
    status: "True"
    type: Ready
```

Not Ready condition:
```yaml
  - lastTransitionTime: "2023-03-23T02:06:53Z" ## We can see the transitionTime is updated as expected.
    message: cluster is not reachable
    reason: ClusterNotReachable
    status: "False"
    type: Ready
```

xigang:
> It works. Thanks. /lgtm /approve
> 
> Ready condition:
> 
> ```yaml
>   - lastTransitionTime: "2023-03-23T02:01:33Z"
>     message: cluster is healthy and ready to accept workloads
>     reason: ClusterReady
>     status: "True"
>     type: Ready
> ```
> 
> Not Ready condition:
> 
> ```yaml
>   - lastTransitionTime: "2023-03-23T02:06:53Z" ## We can see the transitionTime is updated as expected.
>     message: cluster is not reachable
>     reason: ClusterNotReachable
>     status: "False"
>     type: Ready
> ```

@RainbowMango  Thanks for helping verify.

PR #3321, opened 2023-03-23
Do not revise replicas when strategy is duplicated

Poor12:
**What type of PR is this?**
/kind cleanup

Do not revise replicas when strategy is duplicated, because the replicas is exactly the same as the original workload.
Cleanup some redundant logic to make it clearer.

Refer to https://github.com/karmada-io/karmada/issues/3289#issuecomment-1476070082 and https://github.com/karmada-io/karmada/pull/3290#discussion_r1145618919

Fixes #

None


pkg/scheduler/core/assignment.go
RainbowMango, line 1: Does this part change mandatory to the purpose of  `Do not revise replicas when strategy is duplicated`?

pkg/controllers/binding/common.go
```diff
@@ -106,6 +110,13 @@ func ensureWork(
 	return nil
 }
 
+// If and only if the scheduler has scheduled replicas for the target cluster,
+// and the replica scheduling policy is not duplicated(exactly the same as the origin workload),
+// then run the following logic
+func needReviseReplicas(hasScheduledReplica bool, placement *policyv1alpha1.Placement) bool {
```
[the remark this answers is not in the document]
Poor12: Not exactly. `hasScheduledReplica` also requires scheduling results(targetCluster is not nil) in addition to requiring `RB.spec.replicas > 0`

pkg/scheduler/core/assignment.go
Poor12: This part of the code is mainly to make the logic clearer, first judge whether there are replicas, and if so, allocate replicas according to the scheduling policy.

pkg/controllers/binding/common.go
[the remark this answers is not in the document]
RainbowMango: What I'm thinking is to allow `karmada-scheduler` to assign the replicas no matter whether the replica scheduling type is `Duplicated` or `Divided`.  We only revise the replicas when the replica scheduling type is `Divided`.

Does that make sense?

pkg/util/helper/policy.go
```diff
@@ -118,6 +118,15 @@ func IsReplicaDynamicDivided(strategy *policyv1alpha1.ReplicaSchedulingStrategy)
...
 }
 
+// IsReplicaDuplicated checks if a PropagationPolicy schedules replicas as duplicated
+func IsReplicaDuplicated(strategy *policyv1alpha1.ReplicaSchedulingStrategy) bool {
+	if strategy == nil || strategy.ReplicaSchedulingType == policyv1alpha1.ReplicaSchedulingTypeDuplicated {
+		return true
+	}
+
+	return false
+}
```
RainbowMango, line 128: I guess we can have a helper function at `pkg/apis/policy/v1alpha1/propagation_helper.go` like:
```golang
func (p *Placement) ReplicaSchedulingType() ReplicaSchedulingType {
	if p.ReplicaScheduling == nil {
		return ReplicaSchedulingTypeDuplicated
	}

	return p.ReplicaScheduling.ReplicaSchedulingType
}
```

RainbowMango: I'll open an issue for this.

RainbowMango: Done. https://github.com/karmada-io/karmada/issues/3347

pkg/controllers/binding/common.go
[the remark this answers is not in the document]
Poor12: Yes. You reminded me that `replicasInfo` is really unnecessary for duplicated. We can simplify it more.

RainbowMango (approved):
/lgtm
/approve

PR #3326, opened 2023-03-24
feat: add --purge-namespace flags for `karmadactl deinit`, and default not delete namespace

my-git9:
**What type of PR is this?**

/kind feature

Fixes #

`karmadactl deinit` will delete the namespace where karmda components are located. 
In some cases, this processing is too rough, which may delete other resources created by the user under the namespace (the user does not want to be cleaned up together), such as imagepullsecret, Qos settings, etc. 
If the user is given the opportunity to choose whether to clean up the entire namespace, it will have a better experience.

```release-note
karmadactl: Add `--purge-namespace` flags for `karmadactl deinit`.
```

### DIFF
--- a/pkg/karmadactl/deinit/deinit.go
+++ b/pkg/karmadactl/deinit/deinit.go
@@ -37,8 +37,9 @@ type CommandDeInitOption struct {
 	Namespace  string
 
 	// DryRun tells if run the command in dry-run mode, without making any server requests.
-	DryRun bool
-	Force  bool
+	DryRun         bool
+	Force          bool
+	PurgeNamespace bool
 
 	KubeClientSet *kubernetes.Clientset
 }
@@ -81,6 +82,7 @@ func NewCmdDeInit(parentCommand string) *cobra.Command {
 	flags.StringVar(&opts.Context, "context", "", "The name of the kubeconfig context to use")
 	flags.BoolVar(&opts.DryRun, "dry-run", false, "Run the command in dry-run mode, without making any server requests.")
 	flags.BoolVarP(&opts.Force, "force", "f", false, "Reset cluster without prompting for confirmation.")
+	flags.BoolVar(&opts.PurgeNamespace, "purge-namespace", false, "Run the command with purge-namespace, the namespace which Karmada components were installed will be deleted.")
 	return cmd
 }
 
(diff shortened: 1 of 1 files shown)


my-git9:
The test result:
- Init Karmada:
image

- Create secret in karmada-system ns:
image

- Deinit karmada with `--ignore-namespace`
image

- Check
image

pkg/karmadactl/deinit/deinit.go
```diff
@@ -81,6 +82,7 @@ func NewCmdDeInit(parentCommand string) *cobra.Command {
 	flags.StringVar(&opts.Context, "context", "", "The name of the kubeconfig context to use")
 	flags.BoolVar(&opts.DryRun, "dry-run", false, "Run the command in dry-run mode, without making any server requests.")
 	flags.BoolVarP(&opts.Force, "force", "f", false, "Reset cluster without prompting for confirmation.")
+	flags.BoolVar(&opts.IgnoreNamespace, "ignore-namespace", false, "Reset cluster without delete namespace.")
```
RainbowMango, line 85: The flag name `ignore-namespace` might be ambiguous, since `deinit` is going to uninstall/remove `components` from a cluster. `ignore-namespace` can be misinterpreted as `if the component in the namespace, then ignore(cancel the process)`.  What do you think?

my-git9: This is indeed a question, what do you think of the flag name `no-delete-namespace`?

RainbowMango: Before talking about the flag name, I'd like to invite owners(@carlory @lonelyCZ ) to comment if this is a reasonable feature.

calvin0327:
If there are other workloads in the namespace `karmada-system` before installing karmada with karmadactl, It,s dangerous to delete the namespace.

pkg/karmadactl/deinit/deinit.go
jwcesign: --preserve-resources ?

my-git9: Mybe we can not delete namespace by default, and  delete namespace by specify options `--purge-namespace` ?

XiShanYongYe-Chang: kindly ping @lonelyCZ @my-git9, do we have an executive decision?

hzxuzhonghu: >For security and more intuitive, I think we could let users determine if remove the namespace. Like

+1, istio does not delete `istio-system` namespace when it is uninstalled by istioctl

lonelyCZ: > Mybe we can not delete namespace by default, and delete namespace by specify options --purge-namespace ?

I think it is ok.

hzxuzhonghu:
/lgtm

@lonelyCZ can you approve

hzxuzhonghu (approved):

lonelyCZ:
Thanks @my-git9 

/approve

Issue #3331, opened 2023-03-25
Add third-party resources into the Resource Interpreter framework


yike21:
**What would you like to be added**:
Karmada's Resource Interpreter Framework is designed for interpreting resource structure. It consists of built-in and customized interpreters. Karmada could bundle some popular and open-sourced resources so that users can save the effort to customize them.

**Tasks**:
1. Argo Workflow  (deadline: April 23, 2023)
    group: argoproj.io
    version: v1alpha1
    - [x] Implement resoucecustomization for  Workflow, including `InterpretReplica`, `ReviseReplica`, `Retain`,  `InterpretDependency`,`InterpretHealthy`. #3438 

2. Flux CD (deadline: April 30, 2023)
    group: source.toolkit.fluxcd.io
    version: v1
    - [x] implement resoucecustomization for  GitRepository, including `Retain`, `AggregateStatus`, `InterpretStatus`, `InterpretDependency`,`InterpretHealthy`. #3481 

    group: source.toolkit.fluxcd.io
    version: v1beta2
    - [x] implement resoucecustomization for OCIRepository, including `Retain`, `AggregateStatus`, `InterpretStatus`, `InterpretDependency`,`InterpretHealthy`. #3482
    - [x] implement resourcecustomization for HelmRepository, including `Retain`, `AggregateStatus`, `InterpretStatus`, `InterpretDependency`,`InterpretHealthy`. #3484
    - [x] implement resoucecustomization for Bucket, including `Retain`, `AggregateStatus`, `InterpretStatus`, `InterpretDependency`,`InterpretHealthy`. #3485 
    - [x] implement resourcecustomization for HelmChart, including `Retain`, `AggregateStatus`, `InterpretStatus`, `InterpretDependency`,`InterpretHealthy`. #3486 

    group: kustomize.toolkit.fluxcd.io
    version: v1
    - [x] implement resoucecustomization for Kustomization, including `Retain`, `AggregateStatus`, `InterpretStatus`, `InterpretDependency`,`InterpretHealthy`. #3487 

    group: helm.toolkit.fluxcd.io
    version: v2beta1
    - [x] implement resoucecustomization for HelmRelease, including `Retain`, `AggregateStatus`, `InterpretStatus`, `InterpretDependency`,`InterpretHealthy`. #3488
    
3. Kyverno (deadline: May 10, 2023)
    group: kyverno.io    
    version: v1
[...]

PR #3352, opened 2023-04-01
add customizations for apps.krusie.io/v1beta1/StatefulSet

yike21:
**What type of PR is this?**
/kind cleanup

Add third-party resources `apps.krusie.io/v1beta1/StatefulSet` into Resource Interpreter framework.
According to [Note](https://openkruise.io/docs/user-manuals/advancedstatefulset), we can find that since Kruise v0.7.0, Advanced StatefulSet has been promoted to v1beta1, which is compatible with v1alpha1.

Part of #3331 

@Poor12 

NONE

```release-note
`Resource Interpreter`: Support `apps.krusie.io/v1beta1/StatefulSet`
```

### DIFF
--- a/pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/apps.kruise.io/v1beta1/StatefulSet/customizations.yaml
+++ b/pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/apps.kruise.io/v1beta1/StatefulSet/customizations.yaml
@@ -1,0 +1,125 @@
+apiVersion: config.karmada.io/v1alpha1
+kind: ResourceInterpreterCustomization
+metadata:
+  name: declarative-configuration-statefulset
+spec:
+  target:
+    apiVersion: apps.kruise.io/v1beta1
+    kind: StatefulSet
+  customizations:
+    replicaResource:
+      luaScript: >
+        local kube = require("kube")
+        function GetReplicas(obj)
+          replica = obj.spec.replicas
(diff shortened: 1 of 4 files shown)


pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/apps.kruise.io/v1beta1/StatefulSet/customizations.yaml
```diff
@@ -0,0 +1,121 @@
...
+          status.readyReplicas = observedObj.status.readyReplicas
+          status.currentReplicas = observedObj.status.currentReplicas
+          status.updatedReplicas = observedObj.status.updatedReplicas
+          status.availableReplicas = observedObj.status.availableReplicas
+          status.updateRevision = observedObj.status.updateRevision
+          status.currentRevision = observedObj.status.currentRevision
+          status.updatedReadyReplicas = observedObj.status.updatedReadyReplicas
+          return status
+        end
+    healthInterpretation:
```
yike21, line 99: `apps.kruise.io` statefulset is a reference to the code in `pkg/resourceinterpreter/default/native/healthy.go` for the health part of native statefulset. Compared to cloneset, statefulset health is more relaxed, the idea is changed from "determine what is healthy" to "determine what is unhealthy/illegal". It allows a member cluster to be considered healthy when the pod is not fully ready, and to be aggregated.

PR #3363, opened 2023-04-04
fix an issue where DeepEqual becomes invalid in aggregate binding status

XiShanYongYe-Chang:
**What type of PR is this?**

/kind cleanup

When the binding status is collected from the work, we check whether the binding status changes to reduce the number of updated. However, the result is always false when the status pointer is compared with bindingStatus, so we need to change the pointer to the struce.

Fixes #

```release-note
NONE
```


RainbowMango (approved):
/approve

How did you find this?

XiShanYongYe-Chang:
> How did you find this?

One person reminded me, but he couldn't submit the code inside the company.

PR #3364, opened 2023-04-04
use patch helper instead of jsonpatch

whitewindmills:
**What type of PR is this?**
/kind cleanup

We should use patch helper instead of `jsonpatch`, it's reusable code for easy maintenance.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -8,7 +8,6 @@ import (
 	"reflect"
 	"time"
 
-	jsonpatch "github.com/evanphx/json-patch/v5"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
@@ -505,19 +504,11 @@ func (s *Scheduler) patchScheduleResultForResourceBinding(oldBinding *workv1alph
 	newBinding.Annotations[util.PolicyPlacementAnnotation] = placement
 	newBinding.Spec.Clusters = scheduleResult
 
-	oldData, err := json.Marshal(oldBinding)
+	patchBytes, err := helper.GenMergePatch(oldBinding, newBinding)
 	if err != nil {
-		return fmt.Errorf("failed to marshal the existing resource binding(%s/%s): %v", oldBinding.Namespace, oldBinding.Name, err)
+		return err
 	}
-	newData, err := json.Marshal(newBinding)
-	if err != nil {
-		return fmt.Errorf("failed to marshal the new resource binding(%s/%s): %v", newBinding.Namespace, newBinding.Name, err)
-	}
-	patchBytes, err := jsonpatch.CreateMergePatch(oldData, newData)
-	if err != nil {
-		return fmt.Errorf("failed to create a merge patch: %v", err)
-	}
-	if "{}" == string(patchBytes) {
+	if len(patchBytes) == 0 {
 		return nil
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/scheduler/scheduler.go
```diff
@@ -771,26 +759,32 @@ func patchClusterBindingStatusCondition(karmadaClient karmadaclientset.Interface
...
-		return nil
-	}
-
-	patchBytes, err := helper.GenMergePatch(crb, updateCRB)
+	patchBytes, ok, err := helper.GenMergePatch(crb, updateCRB)
 	if err != nil {
-		return fmt.Errorf("failed to create a merge patch: %v", err)
+		return err
+	}
+	if !ok {
```
RainbowMango, line 786: If we want to check whether a patch is empty or not, why not just use `if len(patchBytes) == 0`? It's more straightforward I guess.

whitewindmills: Thanks for your suggestion.

pkg/util/helper/patch_test.go
```diff
@@ -40,7 +40,6 @@ func TestGenMergePatch(t *testing.T) {
 				return modified
 			},
 			expectedPatch: `{"spec":{"clusters":[{"name":"m1","replicas":5},{"name":"m2","replicas":5}],"replicas":10}}`,
-			expectErr:     false,
```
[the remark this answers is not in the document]
whitewindmills: You're right, I'll add it.

pkg/scheduler/scheduler.go
```diff
@@ -727,27 +710,32 @@ func patchBindingStatusCondition(karmadaClient karmadaclientset.Interface, rb *w
...
+	}
 	return patchBindingStatus(karmadaClient, rb, updateRB)
 }
 
 // patchBindingStatusWithAffinityName patches schedule status with affinityName of ResourceBinding when necessary.
 func patchBindingStatusWithAffinityName(karmadaClient karmadaclientset.Interface, rb *workv1alpha2.ResourceBinding, affinityName string) error {
-	klog.V(4).Infof("Begin to patch status with affinityName(%s) to ResourceBinding(%s/%s).", affinityName, rb.Namespace, rb.Name)
+	if rb.Status.SchedulerObservedAffinityName == affinityName {
+		return nil
+	}
```
XiShanYongYe-Chang, line 724: It make sense.

pkg/scheduler/scheduler.go
```diff
@@ -727,27 +710,32 @@ func patchBindingStatusCondition(karmadaClient karmadaclientset.Interface, rb *w
 	if newScheduledCondition.Status == metav1.ConditionTrue {
 		updateRB.Status.SchedulerObservedGeneration = rb.Generation
 	}
+
+	if reflect.DeepEqual(rb.Status, updateRB.Status) {
+		return nil
+	}
```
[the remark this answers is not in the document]
XiShanYongYe-Chang: I don't get it, +1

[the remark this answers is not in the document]
whitewindmills: Maybe, some simple checks do not need it.
```
if rb.Status.SchedulerObservedAffinityName == affinityName {
		return nil
	}
```

pkg/util/helper/patch_test.go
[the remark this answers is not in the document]
RainbowMango: Yeah, thanks. Let's do it.

pkg/scheduler/scheduler.go
```diff
@@ -505,19 +504,11 @@ func (s *Scheduler) patchScheduleResultForResourceBinding(oldBinding *workv1alph
 	newBinding.Annotations[util.PolicyPlacementAnnotation] = placement
 	newBinding.Spec.Clusters = scheduleResult
 
-	oldData, err := json.Marshal(oldBinding)
+	patchBytes, err := helper.GenMergePatch(oldBinding, newBinding)
```
jwcesign, line 507: Why did we use `application/merge-patch+json` to patch the resource instead of using `application/json-patch+json`? 

`application/json-patch+json` has a smaller data exchange.

whitewindmills: It's not absolute. For example:
data: `{"name":"foo", "class":"A", "count":3}`
We want to update `name` and `class` fields.
* Use _JSON Patch_: 
```
[
	{
		"op" : "replace" ,
		"path" : "/name" ,
		"value" : "bar"
	},
	{
		"op" : "replace" ,
		"path" : "/class" ,
		"value": "B"
        }
]
```

* Use _JSON Merge Patch_: 
```
{"name":"foo", "class":"A"}
```
Both have advantages and disadvantages, and none of them will fit everybody’s usecases, but I think _JSON Merge 
 Patch_ is enough here and more simple.

RainbowMango:
@XiShanYongYe-Chang @jwcesign Do you have any other comments?

RainbowMango (approved):
/approve

PR #3371, opened 2023-04-06
Unify the assign replicas when duplicated

Poor12:
**What type of PR is this?**
/kind cleanup

1. Unify the `assignReplicas` function when `placement.ReplicaSchedulingType()` is equal to `Duplicated`.
2. Refactor some codes to make it clearer.

Fixes #

None

### DIFF
--- a/pkg/scheduler/core/generic_scheduler.go
+++ b/pkg/scheduler/core/generic_scheduler.go
@@ -89,7 +89,7 @@ func (g *genericScheduler) Schedule(
 	}
 	klog.V(4).Infof("Selected clusters: %v", clusters)
 
-	clustersWithReplicas, err := g.assignReplicas(clusters, spec.Placement.ReplicaScheduling, spec)
+	clustersWithReplicas, err := g.assignReplicas(clusters, spec.Placement, spec)
 	if err != nil {
 		return result, fmt.Errorf("failed to assignReplicas: %v", err)
 	}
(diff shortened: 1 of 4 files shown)


Poor12:
/hold
This PR depend on #3321.

Garrybest:
Good job👍.

/lgtm

RainbowMango (approved):
/approve

pkg/scheduler/core/generic_scheduler.go
```diff
@@ -182,13 +182,13 @@ func (g *genericScheduler) assignReplicas(
 		return nil, fmt.Errorf("no clusters available to schedule")
 	}
 
-	if object.Replicas > 0 && replicaSchedulingStrategy != nil {
```
whitewindmills, line 185: In fact, it's a bug. I think it needs to be cherry-picked to the previous branches. When `replicaSchedulingStrategy` is undefined, the workload will not be assigned replicas.

Poor12: Before this PR, when `replicaSchedulingStrategy` is undefined, it implies that `replicaSchedulingStrategy` is `duplicated`. Though `resourceBinding` do not show `replicas`, Karmada will propagate the workload with the whole replicas to member clusters. I wonder if it will affect your business.

whitewindmills: Thanks your reminder.

PR #3393, opened 2023-04-11
Avoid accidental impact for WeightPreference on the cache

### LINKED ISSUE #3361: The logic of karmada-scheduler judging whether the placement has changed could be optimized

A while ago we had a problem when updating karmada. We restarted karmada-agent in member cluster and karmada-controller-manager in master cluster, then all resourceBindings were re-enqueued in karmada-scheduler (I understand this was normal behavior). But we found that **some of**(not all) our deployments were rescheduled, and karmada-scheduler will have a corresponding log "Start to schedule ResourceBinding(xxx/xxx) as placement changed". In addition, the status of some clusters changed during the change process, resulting in changes in the scheduling results, and some Pods were rebuilt. 

Because we actually did not enable the feature gates of Failover, we thought that the rescheduling was actually not as expected. So we checked the log and karmada-scheduler, and found that a judgment in karmada-scheduler was not very reasonable.

Our PropagationPolicy is:

```yaml
kind: PropagationPolicy
metadata:
  name: deploy
  namespace: test-test
spec:
  placement:
    clusterAffinity:
      clusterNames:
      - clusterA
      - clusterB
    replicaScheduling:
  resourceSelectors:
  - apiVersion: apps/v1
    kind: Deployment
```

but the placement in the resourceBinding's annotations is

```json
{
  "clusterAffinity": {
    "clusterNames": [
      "clusterA",
      "clusterB"
    ]
  },
  "replicaScheduling": {
    "replicaSchedulingType": "Divided",
    "replicaDivisionPreference": "Weighted",
    "weightPreference": {
      "staticWeightList": [
        {
          "targetCluster": {
            "clusterNames": [
              "clusterA"
            ]
          },
          "weight": 1
        },
        {
          "targetCluster": {
            "clusterNames": [
              "clusterB"
            ]
          },
          "weight": 1
        }
      ]
    }
  }
}
```

The karmada-scheduler compares the strings of the placement of pp and the placement in the annotation. In this case, they must be unequal and will be rescheduled.

```go
// file: pkg/scheduler/scheduler.go
	policyPlacement, policyPlacementStr, err := s.getPlacement(rb)
	if err != nil {
		return err
	}
	appliedPlacement := util.GetLabelValue(rb.Annotations, util.PolicyPlacementAnnotation)
	if policyPlacementStr != appliedPlacement {
		// policy placement changed, need schedule
		klog.Infof("Start to schedule ResourceBinding(%s/%s) as placement changed", namespace, name)
		err = s.scheduleResourceBinding(rb)
		metrics.BindingSchedule(string(ReconcileSchedule), utilmetrics.DurationInSeconds(start), err)
		return err
	}
```

I'm not sure if this can be considered a bug, but at least it's an optimization. 

**Why is this needed**:

whitewindmills:
**What type of PR is this?**
/kind bug

Fixes #3361

```release-note
karmada-scheduler: Fixed unexpected re-scheduling due to mutating informer cache issue.
```

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -298,6 +298,7 @@ func (s *Scheduler) doScheduleBinding(namespace, name string) (err error) {
 		}
 		return err
 	}
+	rb = rb.DeepCopy()
 
 	if rb.Spec.Placement == nil {
 		// never reach here
@@ -345,6 +346,7 @@ func (s *Scheduler) doScheduleClusterBinding(name string) (err error) {
 		}
 		return err
 	}
+	crb = crb.DeepCopy()
 
 	if crb.Spec.Placement == nil {
 		// never reach here


Poor12:
Nice job!
/lgtm

whitewindmills:
> Hi @whitewindmills thanks, can you help add a release-note, maybe we need to cherry-pick this patch to the previous branch.

Thanks for your reminder.

RainbowMango (commented):
I believe this patch fixes the bug exactly, but what I'm thinking is if a similar issue(mutating cache) exists somewhere else, how can we eliminate this problem fundamentally?

Just revisited the relevant code, the whole scheduling process uses a pointer to the cache, that's very scary.

So, I tend to `deep copy` it immediately after RB/CRB is listed from the cache. I know this approach might be more expensive since we are going to copy a whole object instead of one field, but I still think we should do it.

@whitewindmills @Garrybest What's your opinion?

The following pieces of code just explain my idea.
```diff
diff --git a/pkg/scheduler/scheduler.go b/pkg/scheduler/scheduler.go
index e372c695..e2561bce 100644
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -298,6 +298,7 @@ func (s *Scheduler) doScheduleBinding(namespace, name string) (err error) {
                }
                return err
        }
+       rb = rb.DeepCopy()

        if rb.Spec.Placement == nil {
                // never reach here
@@ -345,6 +346,7 @@ func (s *Scheduler) doScheduleClusterBinding(name string) (err error) {
                }
                return err
        }
+       crb = crb.DeepCopy()

        if crb.Spec.Placement == nil {
                // never reach here
```

whitewindmills:
@RainbowMango 
I agree with your concerns. But I don't have a better way for now, if we do this, I think we need to test the memory consumption.

Poor12:
+1, I agree with it.

whitewindmills:
Now that we all agree with this approach, I will update this PR.

RainbowMango:
I updated the release note, by the way.

RainbowMango (approved):
/lgtm
/approve

@whitewindmills Can you help to cherry-pick this to release branches(release-1.5, release-1.4, release-1.3)?

PR #3415, opened 2023-04-17
WaitAPIServiceReady use watch

helen-frank:
**What type of PR is this?**
/kind feature

Avoid too many logs
Fixes #

```release-note
NONE
```


RainbowMango:
How many logs `avoided` by this patch?

helen-frank:
> How many logs `avoided` by this patch?

The new cluster installed karmada, and measured the init before and after the change, the measured log volume dropped from 77 to 18

helen-frank:
> > the measured log volume dropped from 77 to 18
> 
> What kind of log is avoided? I believe you have tested it before sending the PR, it'd be helpful if you can give more details.

The original code will print this log every 1s.
```bash
I0418 06:33:32.788899  375506 check.go:26] Waiting for APIService(v1alpha1.cluster.karmada.io) condition(Available), will try
```

helen-frank:
@RainbowMango  @lonelyCZ hi, After troubleshooting errors for https://github.com/karmada-io/karmada/pull/3418, I think it is necessary to reduce the useless duplicate logs of WaitAPIServiceReady and request merge.

RainbowMango:
Do you mean #3418 rely on this PR?

pkg/karmadactl/cmdinit/karmada/check.go
```diff
@@ -14,19 +15,39 @@ import (
...
 // WaitAPIServiceReady wait the api service condition true
 func WaitAPIServiceReady(c *aggregator.Clientset, name string, timeout time.Duration) error {
-	if err := wait.PollImmediate(time.Second, timeout, func() (done bool, err error) {
-		apiService, e := c.ApiregistrationV1().APIServices().Get(context.TODO(), name, metav1.GetOptions{})
-		if e != nil {
-			return false, nil
+	apiService, err := c.ApiregistrationV1().APIServices().Get(context.TODO(), name, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
```
RainbowMango, line 21: What if the APIService not exist at that time?

helen-frank: I think when this situation occurs, the normal error can be thrown, WaitAPIServiceReady premise is that there is already APIService

helen-frank:
> Do you mean #3418 rely on this PR?

Not interdependent, just found useful when troubleshooting errors and reducing duplicate error logs

pkg/karmadactl/cmdinit/karmada/check.go
RainbowMango: Please @lonelyCZ check if the behavior change is acceptable.

pkg/karmadactl/cmdinit/karmada/check.go
```diff
@@ -14,19 +15,39 @@ import (
...
+		updatedAPIService := &apiregistrationv1.APIService{}
+		if err := runtime.DefaultUnstructuredConverter.FromUnstructured(u, updatedAPIService); err != nil {
+			return err
+		}
+
+		if apiregistrationv1helper.IsAPIServiceConditionTrue(updatedAPIService, apiregistrationv1.Available) {
+			return nil
 		}
 
 		klog.Infof("Waiting for APIService(%s) condition(%s), will try", name, apiregistrationv1.Available)
```
RainbowMango, line 51: Here will dump `info` logs, does it seems too verbose?

helen-frank: The length looks ok, for example `I0601 05:59:14.497568 20024 check.go:26] Waiting for APIService(v1alpha1.cluster.karmada.io) condition(Available), will try`

Maybe we shouldn't print this log, it just tells the user that karmadactl is executing and the current shell is not disconnected, so it's not really that useful

PS: I prefer to use kwokctl circle waiting method, but this seems to be a bit of a big change, I don't know what you think about the evolution of karmadactl later

lonelyCZ:
Thanks @helen-frank , I will look it ASAP

/assign

pkg/karmadactl/cmdinit/karmada/check.go
lonelyCZ: Before executing `WaitAPIServiceReady`, we firstly `CreateOrUpdateAPIService` in general.

https://github.com/karmada-io/karmada/blob/360412399b494d903576e0dba7afe78ea3dbf37a/pkg/karmadactl/cmdinit/karmada/deploy.go#L302-L308

pkg/karmadactl/cmdinit/karmada/check.go
lonelyCZ: The `kworkctl circle` is cool, many project CLI use this way, like [subctl of submariner](https://submariner.io/getting-started/quickstart/kind/#install-subctl) that are interesting and clear. I think we also use this way in the future.

image

lonelyCZ:
Hi, @helen-frank , Could you paste your test result, how many logs do we reduce?

helen-frank:
> Hi, @helen-frank , Could you paste your test result, how many logs do we reduce?

The log volume was reduced from 77 to 18 when karmada-aggregated-apiserver failed to start properly

helen-frank:
ping @lonelyCZ

PR #3416, opened 2023-04-17
feat: Proposal for FederatedHPA

### LINKED ISSUE #3379: Centralized HPA architecture
#### Decentralized HPA:
The concept of [Decentralized HPA](https://github.com/karmada-io/karmada/blob/33a0593c6984e7d35c34ec9576fa5448d81593ca/docs/proposals/hpa/federated-hpa-v2.md) can be complex to comprehend and configure, leading to confusion among users. Additionally, it may result in duplicate jobs with the scheduler.
But it does have one advantage, though: if the control plane goes down, it can still handle burst requests.

#### Centralized HPA
Centralized HPA is quite easy to understand and configure, there are three kinds of architecture:
K8s native metrics server(CPU/Memory):
image

Customer metrics server:
image

image

By implementing this architecture, the HPA will solely focus on scaling deployments while the scheduler schedules replicas based on priority policies.   
However, a drawback of this approach is that in case of control plane failure, it cannot manage sudden surge requests.

**Why is this needed**:
In the current era of single clusters, Horizontal Pod Autoscaling (HPA) is commonly used to dynamically scale workload replicas to handle incoming requests and enhance resource utilization. Therefore, in the era of multi-clusters, HPA should still work and be capable of scaling the workloads in different member clusters, to resolve resource limitations for a single cluster and improve scaling fail recovery capabilities.

jwcesign:
**What type of PR is this?**
/kind feature

1. Manage the HPA resource in a unified manner across clusters. 
1. Scaling workloads across multiple clusters with FederatedHPA and rich policies in PropagationPolicy.
1. Support both Kubernetes HPA(CPU/Memory) and customized HPA(customized metrics).
1. Propose the implementation ideas for involved components, including `karmada-controller-manager` and `karmada-metrics-adaptor`.

Part of #3379 

@XiShanYongYe-Chang @Garrybest @chaunceyjiang 

```release-note

```


docs/proposals/hpa/centralized-hpa.md
```diff
@@ -0,0 +1,181 @@
...
+-->
+
+1. If the HPA is propagated to member clusters, there will be conflicts with Karmada control plane. So users couldn't propagate HPA from Karmada control plane.
+2. The HPA used in a single cluster can be seamlessly migrated to multiple clusters' centralized HPA.
+
+## Design Details
+### Architecture
+For different scenarios, the architecture may be different:  
+HPA with CPU/Memory metrics:  
+![K8s native metrics](./statics/centralizedhpa-native-metrics.png)
```
Garrybest, line 124: Does nativew metrics server support multiple clusters? How does metrics server know which apiserver to request when HPA controller start a new request?

jwcesign: > Does nativew metrics server support multiple clusters?
From the code of K8s native HPA controller: https://github.com/kubernetes/kubernetes/blob/9db01b6eb32995b3d0bc76e3693ebcee80ef0a82/pkg/controller/podautoscaler/replica_calculator.go#L66

The responsibility of Karmada-metrics-server is to query the metrics from member clusters and do aggregation.
So it hasn't got connection with native metrics server.

> How does metrics server know which apiserver to request when HPA controller start a new request?   

From RB

docs/proposals/hpa/centralized-hpa.md
```diff
@@ -0,0 +1,183 @@
...
+```yaml
+- command:
+  - kube-controller-manager
+  - ...
+  - --controllers=xxx,horizontalpodautoscaling
+  - ...
+```
+
+#### karmada-metrics-adaptor
+To query metrics from member clusters, Karmada introduces a new component called `karmada-metrics-adaptor` which is responsible for the following tasks:
```
chaunceyjiang, line 155: Karmada now has ten components.

Can we coupling this component with the existing components?

jwcesign: I prefer to keep the optional component separate since it has its own tasks.

docs/proposals/hpa/centralized-hpa.md
```diff
@@ -0,0 +1,183 @@
...
+* `karmada-metrics-adaptor` is responsible for query metrics from member clusters and do aggregation.
+* Centralized HPA only responsible for scaling instances but doesn't care where the instances will be scheduled.
+* The HPA version should be compatible with `karmada-kube-controller-manager` version.
+
+For better user experience, centralized HPA API is same with original HPA API.
+
+### Components change
+
+#### karmada-kube-controller-manager  
+In order to make HPA scaling work, HPA controller should be enablded in `karmada-kube-controller-manager` as follows:
```
Poor12, line 145: We may introduce a new API to distinguish kubernetes native hpa and      
the multi-cluster hpa. So we might not use the original API though they may have the same fields.

docs/proposals/hpa/centralized-hpa.md
```diff
@@ -0,0 +1,183 @@
...
+  - ...
+```
+
+#### karmada-metrics-adaptor
+To query metrics from member clusters, Karmada introduces a new component called `karmada-metrics-adaptor` which is responsible for the following tasks:  
+* Registering the metrics API service including metrics.k8s.io, custom.metrics.k8s.io, and external.metrics.k8s.io.
+* Listing and watching ResourceBindings to know where workloads are scheduled. When a query is received from apiserver, the `karmada-metrics-adaptor` will query the member clusters and perform aggregation.
+
+For different PropagationPolicy, the aggregation will be different:
+* For divided method, `karmada-metrics-adaptor` will return the metrics found directly from the member clusters.
```
Poor12, line 160: For divide method, we regard the distributed replicas as one application. So we might do some calculation after we get all metrics from member clusters. Here can be explained in detail.

Poor12 (commented):
Generally LGTM

docs/proposals/hpa/federatedhpa-v2.md
```diff
@@ -0,0 +1,479 @@
...
+3. `karmada-metrics-adapter` always list&watch the member clusters' pods, when the requests come, it could query the metrics from the target clusters(where the pod exists). After the metrics are collected, it will aggregate them and return it.
+4. HPA controller will calculate the desired replicas based on metrics and scale the workload directly.
+
+
+**HPA with custom metrics**:  
+![Custom metrics](./statics/federatedhpa-custom-metrics.png)  
+The steps are described as follows:
+1. HPA controller query metrics via metrics API `custom.metrics.k8s.io` with label selector periodically.
+2. `karmada-apiserver` get the metrics API query, and it will route to `karmada-metrics-adapter` via API service registration.
+3. `karmada-metrics-adapter` always list&watch the member clusters' pods, when the requests come, it could query the metrics from the target clusters(where the pod exists). After the metrics are collected, it will aggregate them and return it.
```
chaunceyjiang, line 140: Do you list&watch the `nodes` of the member cluster?

If there are pods with the same name in a member cluster, how should they be handled?

jwcesign: > Do you list&watch the nodes of the member cluster?

Yes

> If there are pods with the same name in a member cluster, how should they be handled?

They are in different Lister cache, I think there are no conflicts.

chaunceyjiang: ok,

I think you should provide some explanations in the proposal.

jwcesign: Got it

jwcesign: Fixed, please take another look

RainbowMango:
This proposal is not the end, we still need to continue iterating. For example, how to solve the installation, the proposal should be updated during the feature iteration.

RainbowMango (approved):
/lgtm
/approve

PR #3418, opened 2023-04-17
Optimization WaitForStatefulSetRollout and WaitForDeploymentRollout

helen-frank:
**What type of PR is this?**
/kind feature

1. use watch 
2. use kubernetes-sigs/application judge status

Fixes #

```release-note
NONE
```


pkg/karmadactl/util/check.go
```diff
@@ -6,72 +6,135 @@ import (
...
+
+	sts := &appsv1.StatefulSet{}
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(u, sts); err != nil {
+		return StatusUnknown, err
+	}
+
+	if sts.Status.ObservedGeneration == sts.Generation &&
+		sts.Status.Replicas == *sts.Spec.Replicas &&
+		sts.Status.ReadyReplicas == *sts.Spec.Replicas &&
+		sts.Status.CurrentReplicas == *sts.Spec.Replicas {
```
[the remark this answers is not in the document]
helen-frank: Yes, when you update the sts.Spec.Replicas field to a higher value, Kubernetes will start creating new replicas of the StatefulSet. This process can take some time depending on the size of the StatefulSet and the available resources in the cluster.

During this process, the sts.Status.CurrentReplicas field will gradually increase as each new replica is created and becomes ready. However, it may take some time for all replicas to become ready, so there may be a period of time where sts.Spec.Replicas is greater than sts.Status.CurrentReplicas.

Once all replicas have been created and become ready, sts.Status.CurrentReplicas should equal sts.Spec.Replicas. If there are any issues during the creation of new replicas, the sts.Status.CurrentReplicas field may be lower than sts.Spec.Replicas, indicating that there are some replicas that are not yet ready or have failed to start.

( However, your screenshot does not seem to be related to the problem.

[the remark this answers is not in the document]
zishen: ok，i got it. Thanks fo your explanation.

RainbowMango (commented):
Can you explain why we need to do this a little bit?

helen-frank:
> Can you explain why we need to do this a little bit?

More accurate status determination and event-dependent waiting instead of violent polling

pkg/karmadactl/util/check.go
```diff
@@ -6,72 +6,138 @@ import (
...
+		return StatusReady, nil
+	}
+	return StatusInProgress, nil
+}
+
+// Reference https://github.com/kubernetes-sigs/application/blob/e5329b1b083ece8abb4b4efaf738ab9277fbb2ce/controllers/status.go#L94-#L132
+//
+//nolint:gocyclo
```
RainbowMango, line 100: Personally, I'm hesitant to introduce `nolint`. That's my concern.

helen-frank: This is my reference to the kubernetes-sigs/application code, only adding ToUnstructured, probably because the logic of if is more complicated to determine

yanfeng1992:
> More accurate status determination and event-dependent waiting instead of violent polling

Use the watch method to check, the real-time performance is better. And through the watch k8s event, some unnecessary queries can be reduced. 

/lgtm

pkg/karmadactl/util/check.go
```diff
@@ -6,72 +6,138 @@ import (
...
-				d.Generation, d.Status.ObservedGeneration)
-			return false, nil
-		}
-		if (d.Spec.Replicas != nil) && (d.Status.UpdatedReplicas < *d.Spec.Replicas) {
-			lastErr = fmt.Errorf("the number of pods targeted by the deployment (%d pods) is different "+
-				"from the number of pods targeted by the deployment that have the desired template spec (%d pods)",
-				*d.Spec.Replicas, d.Status.UpdatedReplicas)
-			return false, nil
+func WaitForDeploymentRollout(c kubernetes.Interface, dep *appsv1.Deployment, timeoutSeconds time.Duration) error {
+	depWatcher, err := c.AppsV1().Deployments(dep.GetNamespace()).Watch(context.Background(), metav1.SingleObject(dep.ObjectMeta))
```
carlory, line 52: It forgot to call `.Stop()` function. A better choice is  using `k8s.io/client-go/tools/watch` which provides some helper funcs.

pkg/karmadactl/util/check.go
```diff
@@ -6,72 +6,138 @@ import (
...
-			lastErr = fmt.Errorf("expected %d replicas, got %d updated replicas",
-				*s.Spec.Replicas, s.Status.UpdatedReplicas)
-			return false, nil
-		}
-		if s.Status.AvailableReplicas < s.Status.UpdatedReplicas {
-			lastErr = fmt.Errorf("expected %d replicas, got %d available replicas",
-				s.Status.UpdatedReplicas, s.Status.AvailableReplicas)
-			return false, nil
+func WaitForStatefulSetRollout(c kubernetes.Interface, sts *appsv1.StatefulSet, timeoutSeconds time.Duration) error {
+	stsWatcher, err := c.AppsV1().StatefulSets(sts.GetNamespace()).Watch(context.Background(), metav1.SingleObject(sts.ObjectMeta))
```
carlory, line 25: ditto

pkg/karmadactl/util/check.go
helen-frank: ok, got it

RainbowMango:
I'm looking at the CLI/init tests. Don't worry about that.

helen-frank:
> I'm looking at the CLI/init tests. Don't worry about that.

I previously looked at the updated code is relatively simple, did not do the test in their own environment
I self-tested and found that it would report an error "error: prepare karmada failed. failed download file. url: https://github.com/karmada-io/karmada/releases/download/v1.7.0/crds. tar.gz code: 404"
The reason is
https://github.com/karmada-io/karmada/blob/cd9ca28f26fa266c2977a30df64ecdd7eb1046a4/pkg/version/release.go#L14-L27

But I see that cicd is using the local crds.tar.gz

RainbowMango:
> I self-tested and found that it would report an error "error: prepare karmada failed. failed download file. 

Yes, because I tagged a pre-release `[v1.7.0-alpha.1](https://github.com/karmada-io/karmada/releases/tag/v1.7.0-alpha.1)` this morning, `init` will use the `v1.7.0` as the default version.  I'm thinking about how to deal with this.  
cc @jwcesign who also looking at this issue.

By the way. I just disabled the `CLI/init` workflow temporarily, so that it won't block the following PRs.

helen-frank:
> > I self-tested and found that it would report an error "error: prepare karmada failed. failed download file.
> 
> Yes, because I tagged a pre-release `[v1.7.0-alpha.1](https://github.com/karmada-io/karmada/releases/tag/v1.7.0-alpha.1)` this morning, `init` will use the `v1.7.0` as the default version. I'm thinking about how to deal with this. cc @jwcesign who also looking at this issue.
> 
> By the way. I just disabled the `CLI/init` workflow temporarily, so that it won't block the following PRs.

I found out that the reason for waiting for APIService is that karmada-aggregated-apiserver uses v1.7.0, so it can't pull the image and apiservice `v1alpha1.cluster.karmada.io` is not ready.

By the way request to merge https://github.com/karmada-io/karmada/pull/3415 to reduce the log volume

jwcesign:
> I found out that the reason for waiting for APIService is that karmada-aggregated-apiserver uses v1.7.0, so it can't pull the image and apiservice v1alpha1.cluster.karmada.io is not ready.

The main reason is that when building karmadactl, the latest tag is used as the image version. However, when parsing the `v1.7.0-alpha.1` tag, it results in a parsed version of `v1.7.0`, which is not what was expected.

helen-frank:
An idea: shouldn't it be tested with a mirror built by the submitted code and crds.tar.gz?

jwcesign:
> An idea: use the committed code to build images and crds.tar.gz for testing? 

It makes sense, now the crd is from the committed code, but images are not.

helen-frank:
ping @RainbowMango

RainbowMango (commented):
I guess this will conflict with #3665, I'll look at it after #3665 get merged.

RainbowMango (commented):
I'm hesitant to move this forward because it lacks enough benefits.

helen-frank:
> I'm hesitant to move this forward because it lacks enough benefits.

My idea is to use as many kubernetes mechanisms as possible to make code more cloud-native and faster. And using watch is cooler than wait.

PR #3426, opened 2023-04-18
Automated cherry pick of #3393: Avoid accidental impact for rb/crb's pointer on the cache

whitewindmills:
Cherry pick of #3393 on release-1.3.
#3393: Avoid accidental impact for rb/crb's pointer on the cache
```release-note
karmada-scheduler: Fixed unexpected re-scheduling due to mutating informer cache issue.
```

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -245,8 +245,8 @@ func (s *Scheduler) worker() {
 	}
 }
 
-func (s *Scheduler) getPlacement(resourceBinding *workv1alpha2.ResourceBinding) (policyv1alpha1.Placement, string, error) {
-	var placement policyv1alpha1.Placement
+func (s *Scheduler) getPlacement(resourceBinding *workv1alpha2.ResourceBinding) (*policyv1alpha1.Placement, string, error) {
+	var placement *policyv1alpha1.Placement
 	var clusterPolicyName string
 	var policyName string
 	var policyNamespace string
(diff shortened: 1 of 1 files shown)


pkg/scheduler/scheduler.go
```diff
@@ -369,14 +368,14 @@ func (s *Scheduler) doScheduleBinding(namespace, name string) (err error) {
 	if appliedPlacement := util.GetLabelValue(rb.Annotations, util.PolicyPlacementAnnotation); policyPlacementStr != appliedPlacement {
 		// policy placement changed, need schedule
 		klog.Infof("Start to schedule ResourceBinding(%s/%s) as placement changed", namespace, name)
-		err = s.scheduleResourceBinding(rb)
+		err = s.scheduleResourceBinding(rb, policyPlacement, policyPlacementStr)
```
RainbowMango, line 371: Is this change mandatory for the cherry-pick? Can we have a minimum patch?

whitewindmills: > Is this change mandatory for the cherry-pick? Can we have a minimum patch?

it is related, I don't want to call `DeepCopy` in multiple places to copy Placement.

RainbowMango: The first thing we need to figure out is if the same bug exists in the release branch, and we should always do minimum updates against the release branch to avoid introducing additional bugs.

RainbowMango (commented):
Hi, I wonder can we have a minimum patch for release-1.3, or we can skip the patch for release 1.3.

whitewindmills:
> or we can skip the patch for release 1.3.

Maybe this is not a serious problem, I think it can be skipped.

PR #3437, opened 2023-04-20
Add GracePeriodSeconds for GracefulEvictionTask

### LINKED ISSUE #3430: Introduce deleteOption for GracefulEvictionTask
For application migration, we provide three ways, one is `Never` which means that Karmada will not delete the application in the failed clusters. But for the current GracefulEvictionTask, we cannot distinguish that GracefulEvictionTask should be kept or deleted when timeout is reached. In that case, we may need an flag to mark it.

**Why is this needed**:
Distinguish that GracefulEvictionTask should be kept or deleted when timeout is reached.

### LINKED ISSUE #3436: How to handle dynamic parameters which need to be changed every time before generating works for member cluster?
Below is a basic scenario in our release process:

image

We've implement a `deploy CRD called Rollout` based on [openkruise](https://github.com/openkruise/kruise) to handle the whole process. Part of rollout's definition for above case is as follows:
```yaml
apiVersion:
kind: Rollout
metadata:
  name:
  namespace:
spec:
  fortPull: false
  replicas: 4
  selector:
    matchLabels:
      app: 
  strategy:
    canary:
      steps:
      - properties:  # fort
          delay: "-1"
          pull: "false"
        type: Fort
      - properties:  # canary
          delay: "-1"
          pull: "true"
        type: Fort
      - properties:  # first batch of rolling
          delay: "-1"
          maxUnavailable: "0"
          replicas: 50%
          timeout: "5"
        type: Batch
      - properties:  # second batch of rolling
          delay: "-1"
          maxUnavailable: "0"
          replicas: 100%
          timeout: "5"
        type: Batch
  subsetType: CloneSet
  template:
    ...
```

So in multi-cluster case, what we need to do every time are:
1. Generate `canary.steps` for all member clusters when users start their releases.
2. Allocate a `fort pod` to one of member clusters (values in steps are different between fort and non-fort clusters).

For example, above rollout need to be updated in two clusters: `cluster-x` and `cluster-z`. The differences between two cluster are as follows:
```yaml
# cluster-x.yaml, fort pod in this cluster
strategy:
    canary:
      steps:
      - properties:  # fort
          delay: "-1"
          pull: "false"  # pull out fort
        type: Fort
      - properties:  # canary
          delay: "-1"
          pull: "true"  # pull out fort
        type: Fort
      ...

# cluster-z.yaml, no fort pod
strategy:
    canary:
      steps:
      - properties:  # fort
          delay: "-1"
          pull: nil  # skip
        type: Fort
      - properties:  # canary
          delay: "-1"
          pull: nil  # skip
        type: Fort
      ...
```

So, is there any suitable ways to implement such differentiated configuration? And where is the appropriate place to put the code? 

Related issue is #3375

**Environment**:
- Karmada version: v1.4.1-with-lua
- Kubernetes version: 1.19
- Others:

Poor12:
**What type of PR is this?**
/kind api-change
/kind design

Add GracePeriodSeconds for GracefulEvictionTask.

Fixes #3430 

```
Add GracePeriodSeconds for GracefulEvictionTask.
```

### DIFF
--- a/pkg/apis/policy/v1alpha1/propagation_types.go
+++ b/pkg/apis/policy/v1alpha1/propagation_types.go
@@ -221,6 +221,15 @@ type ApplicationFailoverBehavior struct {
 	// +optional
 	PurgeMode PurgeMode `json:"purgeMode,omitempty"`
 
+	// GracePeriodSeconds is the maximum waiting duration in seconds before
+	// application on the migrated cluster should be deleted.
+	// Required only when PurgeMode is "Graciously" and defaults to 600s.
+	// If the application on the new cluster cannot reach a Healthy state,
+	// Karmada will delete the application after GracePeriodSeconds is reached.
+	// Value must be positive integer.
+	// +optional
+	GracePeriodSeconds *int32 `json:"gracePeriodSeconds,omitempty"`
+
 	// BlockPredecessorSeconds represents the period of time the cluster from which the
 	// application was migrated from can be schedulable again.
 	// During the period of BlockPredecessorSeconds, clusters are forcibly filtered out by the scheduler.
--- a/pkg/apis/work/v1alpha2/binding_types.go
+++ b/pkg/apis/work/v1alpha2/binding_types.go
@@ -206,6 +206,20 @@ type GracefulEvictionTask struct {
 	// +required
 	Producer string `json:"producer"`
 
+	// GracePeriodSeconds is the maximum waiting duration in seconds before the item
+	// should be deleted. If the application on the new cluster cannot reach a Healthy state,
+	// Karmada will delete the item after GracePeriodSeconds is reached.
+	// Value must be positive integer.
+	// It can not co-exist with SuppressDeletion.
+	// +optional
+	GracePeriodSeconds *int32 `json:"gracePeriodSeconds,omitempty"`
+
+	// SuppressDeletion represents the grace period will be persistent until
+	// the tools or human intervention stops it.
+	// It can not co-exist with GracePeriodSeconds.
+	// +optional
+	SuppressDeletion *bool `json:"suppressDeletion,omitempty"`
+
 	// CreationTimestamp is a timestamp representing the server time when this object was
 	// created.
 	// Clients should not set this value to avoid the time inconsistency issue.
--- a/pkg/controllers/gracefuleviction/evictiontask.go
+++ b/pkg/controllers/gracefuleviction/evictiontask.go
@@ -48,8 +48,22 @@ func assessEvictionTasks(bindingSpec workv1alpha2.ResourceBindingSpec,
 }
 
 func assessSingleTask(task workv1alpha2.GracefulEvictionTask, opt assessmentOption) *workv1alpha2.GracefulEvictionTask {
+	if task.SuppressDeletion != nil {
+		if *task.SuppressDeletion {
+			return &task
+		}
+		// If *task.SuppressDeletion is equal to false,
+		// it means users have confirmed that they want to delete the redundant copy.
+		// In that case, we will delete the task immediately.
+		return nil
+	}
+
+	timeout := opt.timeout
+	if task.GracePeriodSeconds != nil {
+		timeout = time.Duration(*task.GracePeriodSeconds) * time.Second
+	}
 	// task exceeds timeout
-	if metav1.Now().After(task.CreationTimestamp.Add(opt.timeout)) {
+	if metav1.Now().After(task.CreationTimestamp.Add(timeout)) {
 		return nil
 	}
 
@@ -94,6 +108,9 @@ func nextRetry(tasks []workv1alpha2.GracefulEvictionTask, timeout time.Duration,
 	retryInterval := timeout / 10
 
 	for i := range tasks {
+		if tasks[i].GracePeriodSeconds != nil {
+			timeout = time.Duration(*tasks[i].GracePeriodSeconds) * time.Second
+		}
 		next := tasks[i].CreationTimestamp.Add(timeout).Sub(timeNow)
 		if next < retryInterval {
 			retryInterval = next
--- a/pkg/util/helper/policy.go
+++ b/pkg/util/helper/policy.go
@@ -6,6 +6,7 @@ import (
 	discoveryv1 "k8s.io/api/discovery/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/klog/v2"
+	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	policyv1alpha1 "github.com/karmada-io/karmada/pkg/apis/policy/v1alpha1"
@@ -132,10 +133,17 @@ func GetAppliedPlacement(annotations map[string]string) (*policyv1alpha1.Placeme
 	return placement, nil
 }
 
-// SetReplicaDivisionPreferenceWeighted Set the default value of ReplicaDivisionPreference to Weighted
+// SetReplicaDivisionPreferenceWeighted sets the default value of ReplicaDivisionPreference to Weighted
 func SetReplicaDivisionPreferenceWeighted(placement *policyv1alpha1.Placement) {
 	if placement.ReplicaSchedulingType() != policyv1alpha1.ReplicaSchedulingTypeDivided || placement.ReplicaScheduling.ReplicaDivisionPreference != "" {
 		return
 	}
 	placement.ReplicaScheduling.ReplicaDivisionPreference = policyv1alpha1.ReplicaDivisionPreferenceWeighted
+}
+
+// SetDefaultGracePeriodSeconds sets the default value of GracePeriodSeconds in ApplicationFailoverBehavior
+func SetDefaultGracePeriodSeconds(behavior *policyv1alpha1.ApplicationFailoverBehavior) {
+	if behavior.PurgeMode == policyv1alpha1.Graciously && behavior.GracePeriodSeconds == nil {
+		behavior.GracePeriodSeconds = pointer.Int32(600)
+	}
 }
--- a/pkg/util/validation/validation_test.go
+++ b/pkg/util/validation/validation_test.go
@@ -602,7 +602,42 @@ func TestValidateApplicationFailover(t *testing.T) {
 			expectedErr: "spec.failover.application.blockPredecessorSeconds: Invalid value: -100: must be greater than or equal to 0",
 		},
 		{
-			name: "application behavior is correctly defined",
+			name: "the gracePeriodSeconds is declared when purgeMode is not graciously",
+			applicationFailoverBehavior: &policyv1alpha1.ApplicationFailoverBehavior{
+				BlockPredecessorSeconds: pointer.Int32(100),
+				DecisionConditions: policyv1alpha1.DecisionConditions{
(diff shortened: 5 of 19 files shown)


pkg/apis/work/v1alpha2/binding_types.go
```diff
@@ -206,6 +217,12 @@ type GracefulEvictionTask struct {
 	// +required
 	Producer string `json:"producer"`
 
+	// DeleteOption represents that how to deal with the legacy GracefulEvictionTask.
+	// Defaults to "Graciously".
+	// +kubebuilder:default=Graciously
+	// +optional
+	DeleteOption DeleteOption `json:"deleteOption,omitempty"`
```
RainbowMango, line 224:
```golang
	// GracePeriodSeconds is the duration in seconds before the item should be
	// deleted. The value must be a non-negative integer.
	// The value zero indicates delete immediately.
	// It can not co-exist with SuppressDeletion.
	// +optional
	GracePeriodSeconds *int64 `json:"gracePeriodSeconds,omitempty"`

	// SuppressDeletion represents the grace period that will be persistent until
	// the tools or human intervention stops it.
	// It can not co-exist with GracePeriodSeconds.
	SuppressDeletion *bool `json:"suppressDeletion,omitempty"`
```

pkg/apis/policy/v1alpha1/propagation_types.go
```diff
@@ -221,6 +221,15 @@ type ApplicationFailoverBehavior struct {
 	// +optional
 	PurgeMode PurgeMode `json:"purgeMode,omitempty"`
 
+	// GracePeriodSeconds is the maximum waiting duration in seconds before application
+	// on the migrated cluster should be deleted if when PurgeMode is "Graciously".
```
Poor12, line 225: A typo：if when PurgeMode is "Graciously".

RainbowMango (commented):
/assign

pkg/util/validation/validation_test.go
```diff
@@ -601,6 +601,30 @@ func TestValidateApplicationFailover(t *testing.T) {
 			},
 			expectedErr: "spec.failover.application.blockPredecessorSeconds: Invalid value: -100: must be greater than or equal to 0",
 		},
+		{
+			name: "the gracePeriodSeconds is defined when purgeMode is not graciously",
```
RainbowMango, line 605:
```suggestion
			name: "the gracePeriodSeconds is declared when purgeMode is not graciously",
```

RainbowMango (commented):
Please @XiShanYongYe-Chang take a look at the eviction controller part.

pkg/apis/policy/v1alpha1/propagation_types.go
```diff
@@ -221,6 +221,15 @@ type ApplicationFailoverBehavior struct {
 	// +optional
 	PurgeMode PurgeMode `json:"purgeMode,omitempty"`
 
+	// GracePeriodSeconds is the maximum waiting duration in seconds before application
+	// on the migrated cluster should be deleted if PurgeMode is "Graciously".
+	// If the application on the new cluster cannot reach a Healthy state,
+	// Karmada will delete the application after GracePeriodSeconds is reached.
+	// Value must be non-negative integer.
+	// The value zero indicates delete immediately.
```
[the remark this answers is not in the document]
Poor12: Currently I have add a validation to ensure users set `GracePeriodSeconds` when PurgeMode is not `Graciously`. 
I think it's a optional field. If users do not set this field, we will use the global timeout config by eviction-controller.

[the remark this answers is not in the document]
Poor12: I have a question, do we need to keep the implementation of `Immediately` and `Graciously but GracePeriodSeconds = 0`.
From the user's point of view, they may be the same. But `Immediately` will not generate `GracefulEvictionTask` while the latter will generate and this task will be deleted soon due to short periods.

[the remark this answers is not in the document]
RainbowMango:
```golang
	// GracePeriodSeconds is the maximum waiting duration in seconds before
	// application on the migrated cluster should be deleted.
	// Required only when PurgeMode is "Graciously" and defaults to 600s.
	// If the application on the new cluster cannot reach a Healthy state,
	// Karmada will delete the application after GracePeriodSeconds is reached.
	// Value must be non-negative integer.
	// +optional
	GracePeriodSeconds *int32 `json:"gracePeriodSeconds,omitempty"`
```
I'd prefer to have a default value `only when`  PurgeMode is "Graciously".

pkg/apis/work/v1alpha2/binding_types.go
```diff
@@ -206,6 +206,20 @@ type GracefulEvictionTask struct {
...
+	// Value must be positive integer.
+	// It can not co-exist with SuppressDeletion.
+	// +optional
+	GracePeriodSeconds *int32 `json:"gracePeriodSeconds,omitempty"`
+
+	// SuppressDeletion represents the grace period will be persistent until
+	// the tools or human intervention stops it.
+	// It can not co-exist with GracePeriodSeconds.
+	// +optional
+	SuppressDeletion *bool `json:"suppressDeletion,omitempty"`
```
XiShanYongYe-Chang, line 221: Can we just set its type to bool?

pkg/controllers/gracefuleviction/evictiontask.go
```diff
@@ -48,8 +48,19 @@ func assessEvictionTasks(bindingSpec workv1alpha2.ResourceBindingSpec,
 }
 
 func assessSingleTask(task workv1alpha2.GracefulEvictionTask, opt assessmentOption) *workv1alpha2.GracefulEvictionTask {
+	if task.SuppressDeletion != nil {
+		if *task.SuppressDeletion {
+			return &task
+		}
+		return nil
```
XiShanYongYe-Chang, line 55: In this case, do not need to judge the timeout?

pkg/apis/work/v1alpha2/binding_types.go
```diff
@@ -206,6 +206,20 @@ type GracefulEvictionTask struct {
...
+	// GracePeriodSeconds is the maximum waiting duration in seconds before the item
+	// should be deleted. If the application on the new cluster cannot reach a Healthy state,
+	// Karmada will delete the item after GracePeriodSeconds is reached.
+	// Value must be positive integer.
+	// It can not co-exist with SuppressDeletion.
+	// +optional
+	GracePeriodSeconds *int32 `json:"gracePeriodSeconds,omitempty"`
+
+	// SuppressDeletion represents the grace period will be persistent until
+	// the tools or human intervention stops it.
```
XiShanYongYe-Chang, line 218: What are the `tools or human intervention`?

pkg/util/helper/policy.go
```diff
@@ -139,3 +140,10 @@ func SetReplicaDivisionPreferenceWeighted(placement *policyv1alpha1.Placement) {
 	}
 	placement.ReplicaScheduling.ReplicaDivisionPreference = policyv1alpha1.ReplicaDivisionPreferenceWeighted
 }
+
+// SetDefaultGracePeriodSeconds Set the default value of GracePeriodSeconds in ApplicationFailoverBehavior
+func SetDefaultGracePeriodSeconds(behavior *policyv1alpha1.ApplicationFailoverBehavior) {
```
XiShanYongYe-Chang, line 145: Can we directly use `kubebuilder` to set the default value in the API definition?

pkg/apis/work/v1alpha2/binding_types.go
```diff
@@ -206,6 +206,20 @@ type GracefulEvictionTask struct {
 	// +required
 	Producer string `json:"producer"`
 
+	// GracePeriodSeconds is the maximum waiting duration in seconds before the item
+	// should be deleted. If the application on the new cluster cannot reach a Healthy state,
+	// Karmada will delete the item after GracePeriodSeconds is reached.
+	// Value must be positive integer.
+	// It can not co-exist with SuppressDeletion.
+	// +optional
+	GracePeriodSeconds *int32 `json:"gracePeriodSeconds,omitempty"`
```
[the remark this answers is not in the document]
Poor12: If a GracefulEvictionTask has GracePeriodSeconds, then we use it, otherwise will use the parameter of the karmada-controller-manager.

pkg/apis/work/v1alpha2/binding_types.go
Poor12: We should distinguish between `==nil` and `false` these two cases. if `SuppressDeletion` is nil, we will use `Graciously` mode to deal with tasks. If `SuppressDeletion` == false, it means user want to delete the task now.

pkg/controllers/gracefuleviction/evictiontask.go
Poor12: If users set `SuppressDeletion`, then whether to delete the task depends on this field.

pkg/apis/work/v1alpha2/binding_types.go
Poor12: For example, we might offer a command-line tool which may allow users to set this field.

pkg/util/helper/policy.go
Poor12: This field only takes effect when purge == `Graciously`. I'm afraid `kubebuilder` cannot judge this condition.

pkg/util/helper/policy.go
```diff
@@ -139,3 +140,10 @@ func SetReplicaDivisionPreferenceWeighted(placement *policyv1alpha1.Placement) {
 	}
 	placement.ReplicaScheduling.ReplicaDivisionPreference = policyv1alpha1.ReplicaDivisionPreferenceWeighted
 }
+
+// SetDefaultGracePeriodSeconds Set the default value of GracePeriodSeconds in ApplicationFailoverBehavior
```
RainbowMango, line 144:
```suggestion
// SetDefaultGracePeriodSeconds sets the default value of GracePeriodSeconds in ApplicationFailoverBehavior
```
Please help update the comment of avove function as well.
I mean this one:
```
// SetReplicaDivisionPreferenceWeighted Set the default value of ReplicaDivisionPreference to Weighted
func SetReplicaDivisionPreferenceWeighted(placement *policyv1alpha1.Placement) {
```

RainbowMango (commented):
Generally looks good to me. Only one nit.

pkg/util/helper/policy.go
Poor12: Done.

pkg/apis/work/v1alpha2/binding_types.go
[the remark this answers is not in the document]
XiShanYongYe-Chang: For cluster failover, can we directly set this field based on `GracefulEvictionTimeout`?

pkg/apis/work/v1alpha2/binding_types.go
XiShanYongYe-Chang: ok, I get it.

pkg/controllers/gracefuleviction/evictiontask.go
XiShanYongYe-Chang: thx, i get it.

pkg/apis/work/v1alpha2/binding_types.go
XiShanYongYe-Chang: ok

PR #3446, opened 2023-04-21
cleanup pkg/descheduler/core/filter.go

wlp1153468871:
**What type of PR is this?**

/kind cleanup

```release-note
None
```


Poor12:
Hi @wlp1153468871, there are some differences between binding.Placement and binding.AppliedAnnotation. The former is desired scheduling strategy whereas the latter is the applied scheduling strategy currently.

RainbowMango:
@wlp1153468871 Just a kindly reminder, it'd be great to describe a little bit for PRs as per the [template](https://github.com/karmada-io/karmada/blob/master/.github/PULL_REQUEST_TEMPLATE.md).

wlp1153468871:
@Poor12 @RainbowMango Is there any problem that needs to be modified？

Poor12:
> Hi @wlp1153468871, there are some differences between binding.Placement and binding.AppliedAnnotation. The former is desired scheduling strategy whereas the latter is the applied scheduling strategy currently.
Just as I mentioned, these two fields differ. I do not think we should simply replace both. I prefer to keep the status quo.

RainbowMango (commented):
I see the `FilterBinding()` is filtering the RB by `applied placement`:
https://github.com/karmada-io/karmada/blob/17eae7d32c75817e8defe03c745a3340f9fe9e94/pkg/descheduler/core/filter.go#L17-L27

The applied placement might not always same as the placement in RB(which is `declared` placement). 
I want to invite the code owner @Garrybest to take a look.

By the way, @Garrybest might be on vacation recently, it might take a little while for him to respond.

Garrybest:
> The applied placement might not always same as the placement in RB(which is declared placement).
I want to invite the code owner @Garrybest to take a look.

Exactly. This funtion aims to filter bindings with **divided applied placement policy** which is different from **spec**.

RainbowMango:
Thanks @Garrybest for the clarification.  
Please @wlp1153468871 take a look.

PR #3455, opened 2023-04-23
Ensure the latest schedulerObservedGeneration if do not need to schedule

### LINKED ISSUE #3454: After workload is scheduled, binding.status.schedulerObservedGeneration and generation are inconsistent
**What happened**:
When I deploy a workload to multiple clusters with `divided` type and the workload is successfully scheduled, I found that binding.status.schedulerObservedGeneration and generation are inconsistent. Generation is 2 whereas schedulerObservedGeneration is 1. Because binding.Spec has undergone two changes, the binding is generated for the first time, and the scheduling result is written into the binding for the second. However, the second change will not be scheduled because the following three rules are not met.

https://github.com/karmada-io/karmada/blob/48e93dccb6b6ecce984175a37e06bc674424f24a/pkg/scheduler/scheduler.go#L312-L334

Binding.status.schedulerObservedGeneration equal to generation is a access requirement for eviction controller. This may influence failover process.

https://github.com/karmada-io/karmada/blob/48e93dccb6b6ecce984175a37e06bc674424f24a/pkg/controllers/gracefuleviction/rb_graceful_eviction_controller.go#L95

After workload is scheduled, binding.status.schedulerObservedGeneration and generation should be the same.

Just use the sample in the Karmada repo.

This may also exist in the latest three version.

**Environment**:
- Karmada version: v1.5
- Others:

### LINKED ISSUE #3467: graceful-eviction does not work.
**What happened**:
graceful-eviction does not work.

depoly.yaml
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx
  labels:
    app: nginx
spec:
  replicas: 4
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - image: docker.m.daocloud.io/nginx
        name: nginx
        resources:
          limits:
            cpu: 250m
            memory: 512Mi
          requests:
            cpu: 250m
            memory: 512Mi
```

pp.yaml

```
kind: PropagationPolicy
metadata:
  name: nginx-propagation
spec:
  resourceSelectors:
    - apiVersion: apps/v1
      kind: Deployment
      name: nginx
  placement:
    clusterAffinities:
      - affinityName: default
        clusterNames:
          - member1
          - member2
    replicaScheduling:
      weightPreference:

```

Add taints.
``` shell
./_output/bin/darwin/arm64/karmadactl taint clusters member1 key1=value1:NoExecute 
```

RB.yaml
```
spec:
  clusters:
  - name: member2
    replicas: 4
  gracefulEvictionTasks:
  - fromCluster: member1
    producer: TaintManager
    reason: TaintUntolerated
    replicas: 2
  placement:
    clusterAffinities:
    - affinityName: default
      clusterNames:
      - member1
      - member2
    clusterTolerations:
    - effect: NoExecute
      key: cluster.karmada.io/not-ready
      operator: Exists
      tolerationSeconds: 30
    - effect: NoExecute
      key: cluster.karmada.io/unreachable
      operator: Exists
      tolerationSeconds: 30
    replicaScheduling:
      weightPreference:
  replicaRequirements:
    resourceRequest:
      cpu: 250m
      memory: 512Mi
  replicas: 4
  resource:
    apiVersion: apps/v1
    kind: Deployment
    name: nginx
    namespace: default
    resourceVersion: "6942"
    uid: 78aff1f4-ddf4-40f9-9880-5d5f8ea23f33
status:
  aggregatedStatus:
  - applied: true
    clusterName: member1
    health: Healthy
    status:
      availableReplicas: 2
      readyReplicas: 2
      replicas: 2
      updatedReplicas: 2
  - applied: true
    clusterName: member2
    health: Healthy
    status:
      availableReplicas: 4
      readyReplicas: 4
      replicas: 4
      updatedReplicas: 4
  conditions:
  - lastTransitionTime: "2023-04-26T03:20:45Z"
    message: Binding has been scheduled
    reason: BindingScheduled
    status: "True"
    type: Scheduled
  - lastTransitionTime: "2023-04-26T03:20:46Z"
    message: All works have been successfully applied
    reason: FullyAppliedSuccess
    status: "True"
    type: FullyApplied
  schedulerObservedGeneration: 3
  schedulerObservingAffinityName: default

```

```
  gracefulEvictionTasks:
  - fromCluster: member1
    producer: TaintManager
    reason: TaintUntolerated
    replicas: 2

```

The `gracefulEvictionTasks` queue is never empty.

Restart `karmada-controller-manager` doesn't work either.

graceful-eviction can work.

**Environment**:
- Karmada version: release-1.5
- Others:

Poor12:
**What type of PR is this?**
/kind bug

Ensure the latest schedulerObservedGeneration if do not need to schedule.

Fixes #3454 
Fixes #3467 

This PR need to cherry-pick.

```
`karmada-scheduler`: Fixed the issue of inconsistent Generation and SchedulerObservedGeneration.
```

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -334,6 +334,15 @@ func (s *Scheduler) doScheduleBinding(namespace, name string) (err error) {
 	}
 	// TODO(dddddai): reschedule bindings on cluster change
 	klog.V(3).Infof("Don't need to schedule ResourceBinding(%s/%s)", rb.Namespace, rb.Name)
+
+	// If no scheduling is required, we need to ensure that binding.Generation is equal to
+	// binding.Status.SchedulerObservedGeneration which means the current status of binding
+	// is the latest status of successful scheduling.
+	if rb.Generation != rb.Status.SchedulerObservedGeneration {
+		updateRB := rb.DeepCopy()
+		updateRB.Status.SchedulerObservedGeneration = updateRB.Generation
+		return patchBindingStatus(s.KarmadaClient, rb, updateRB)
+	}
 	return nil
 }
 
@@ -382,6 +391,15 @@ func (s *Scheduler) doScheduleClusterBinding(name string) (err error) {
 	}
 	// TODO(dddddai): reschedule bindings on cluster change
 	klog.Infof("Don't need to schedule ClusterResourceBinding(%s)", name)
+
+	// If no scheduling is required, we need to ensure that binding.Generation is equal to
+	// binding.Status.SchedulerObservedGeneration which means the current status of binding
+	// is the latest status of successful scheduling.
+	if crb.Generation != crb.Status.SchedulerObservedGeneration {
+		updateCRB := crb.DeepCopy()
+		updateCRB.Status.SchedulerObservedGeneration = updateCRB.Generation
+		return patchClusterResourceBindingStatus(s.KarmadaClient, crb, updateCRB)
+	}
 	return nil
 }
 
@@ -515,6 +533,7 @@ func (s *Scheduler) patchScheduleResultForResourceBinding(oldBinding *workv1alph
 	_, err = s.KarmadaClient.WorkV1alpha2().ResourceBindings(newBinding.Namespace).Patch(context.TODO(), newBinding.Name, types.MergePatchType, patchBytes, metav1.PatchOptions{})
 	if err != nil {
 		klog.Errorf("Failed to patch schedule to ResourceBinding(%s/%s): %v", oldBinding.Namespace, oldBinding.Name, err)
+		return err
 	}
 
 	klog.V(4).Infof("Patch schedule to ResourceBinding(%s/%s) succeed", oldBinding.Namespace, oldBinding.Name)
@@ -649,7 +668,13 @@ func (s *Scheduler) patchScheduleResultForClusterResourceBinding(oldBinding *wor
 	}
 
 	_, err = s.KarmadaClient.WorkV1alpha2().ClusterResourceBindings().Patch(context.TODO(), newBinding.Name, types.MergePatchType, patchBytes, metav1.PatchOptions{})
-	return err
+	if err != nil {
+		klog.Errorf("Failed to patch schedule to ClusterResourceBinding(%s): %v", oldBinding.Name, err)
+		return err
+	}
+
+	klog.V(4).Infof("Patch schedule to ClusterResourceBinding(%s) succeed", oldBinding.Name)
+	return nil
 }
 
 func (s *Scheduler) handleErr(err error, key interface{}) {


XiShanYongYe-Chang (commented):
Thanks~
/lgtm

XiShanYongYe-Chang:
Maybe we need to describe the components in the release-note.

Poor12:
> Maybe we need to describe the components in the release-note.

Done.

RainbowMango (commented):
The root cause of this issue is that we are [using patch](https://github.com/karmada-io/karmada/blob/2ef01942f17d51216decd97c72749c05dcfa7b3e/pkg/scheduler/scheduler.go#L702-L718) to update the status `after we updated the `.spec` part which will lead generation increasing`. 

I believe this patch could work around this problem, but it looks like a redundancy design. I wonder if we have another alternative solution. cc @Garrybest for help.

Poor12:
> The root cause of this issue is that we are [using patch](https://github.com/karmada-io/karmada/blob/2ef01942f17d51216decd97c72749c05dcfa7b3e/pkg/scheduler/scheduler.go#L702-L718) to update the status `after we updated the `.spec` part which will lead generation increasing`.
> 
> I believe this patch could work around this problem, but it looks like a redundancy design. I wonder if we have another alternative solution. cc @Garrybest for help.

Another optional idea is that we need to distinguish the situation of scheduler filling in the scheduling result.

Garrybest:
I guess [this fucntion](https://github.com/karmada-io/karmada/blob/2ef01942f17d51216decd97c72749c05dcfa7b3e/pkg/scheduler/scheduler.go#L515) could return the latest rb here but we ignore the return value.

Poor12:
> I guess [this fucntion](https://github.com/karmada-io/karmada/blob/2ef01942f17d51216decd97c72749c05dcfa7b3e/pkg/scheduler/scheduler.go#L515) could return the latest rb here but we ignore the return value.

I agree with @Garrybest. Ensuring that the binding is the latest binding before patching status is a good solution.

pkg/scheduler/scheduler.go
```diff
@@ -512,11 +512,13 @@ func (s *Scheduler) patchScheduleResultForResourceBinding(oldBinding *workv1alph
 		return nil
 	}
 
-	_, err = s.KarmadaClient.WorkV1alpha2().ResourceBindings(newBinding.Namespace).Patch(context.TODO(), newBinding.Name, types.MergePatchType, patchBytes, metav1.PatchOptions{})
+	result, err := s.KarmadaClient.WorkV1alpha2().ResourceBindings(newBinding.Namespace).Patch(context.TODO(), newBinding.Name, types.MergePatchType, patchBytes, metav1.PatchOptions{})
 	if err != nil {
```
chaunceyjiang, line 516: image

pkg/scheduler/scheduler.go
```diff
@@ -512,11 +512,13 @@ func (s *Scheduler) patchScheduleResultForResourceBinding(oldBinding *workv1alph
...
 		klog.Errorf("Failed to patch schedule to ResourceBinding(%s/%s): %v", oldBinding.Namespace, oldBinding.Name, err)
+		return err
 	}
 
+	*oldBinding = *result
```
chaunceyjiang, line 521: https://github.com/karmada-io/karmada/blob/5bb8eae68f8c7d94181fecd651c7f54a04765211/pkg/scheduler/scheduler.go#L301

pkg/scheduler/scheduler.go
Poor12: Does this phenomenon still occur with this fix? Would you like to share your steps to reproduce?

chaunceyjiang: You can  reproduce it using the steps in [the issue](https://github.com/karmada-io/karmada/issues/3467) description.

I tried using your patch to fix [the issue](https://github.com/karmada-io/karmada/issues/3467) I encountered, but it doesn't seem to have solved it.

pkg/scheduler/scheduler.go
```diff
@@ -512,11 +512,13 @@ func (s *Scheduler) patchScheduleResultForResourceBinding(oldBinding *workv1alph
 		return nil
 	}
 
-	_, err = s.KarmadaClient.WorkV1alpha2().ResourceBindings(newBinding.Namespace).Patch(context.TODO(), newBinding.Name, types.MergePatchType, patchBytes, metav1.PatchOptions{})
```
chaunceyjiang, line 515: Note: restart `karmada-controller-manager` will also increment the `generation` of RB, but it will not enter the scheduling process at this time.

RainbowMango: > restart karmada-controller-manager will also increment the generation of RB

Why?

pkg/scheduler/scheduler.go
Poor12: I'm trying to reproduce it using the steps in the issue, but it seems the generation and the schedulerObservedGeneration is the same.

pkg/scheduler/scheduler.go
Poor12: @chaunceyjiang is right. Through my analyze, it's mainly because of detector. For resource templates, we add it to the queue directly for adding events. However, for update events, we only add it to queue if spec of resource templates changes for performance. 

https://github.com/karmada-io/karmada/blob/ec7b3b14c66de00b43b11b517cad89baa8bbc382/pkg/detector/detector.go#L280-L308

After the status is aggregated into the resource template, the resourceVersion in rb.spec.resource is inconsistent with that in the actual resource template. It's currently working without issue. However, when controller-manager restarts, detector will add all resource templates again. So we will update rb.spec.resource.resourceVersion to the latest and then generation +1. But for scheduler, it think it do not need to schedule and not enter the scheduling process. So generation and schedulerObservedGeneration is not the same.

Poor12: In this case, return the latest binding after patching scheduler results will not work but the former redundancy design works.

chaunceyjiang: > but the former redundancy design works.

Yes.

RainbowMango: > However, when controller-manager restarts, detector will add all resource templates again. So we will update rb.spec.resource.resourceVersion to the latest and then generation +1. But for scheduler, it think it do not need to schedule and not enter the scheduling process. So generation and schedulerObservedGeneration is not the same.

Got it. Thanks.

Poor12:
@chaunceyjiang, could you please help verify whether this fix can solve your problem? Great thanks.

pkg/scheduler/scheduler.go
chaunceyjiang: > After the status is aggregated into the resource template, the resourceVersion in rb.spec.resource is inconsistent with that in the actual resource template. It's currently working without issue. However, when controller-manager restarts, detector will add all resource templates again. So we will update rb.spec.resource.resourceVersion to the latest and then generation +1.


By the way, currently the graceful-eviction-controller relies too much on this mechanism. I think the graceful-eviction-controller should also handle create event. 

https://github.com/karmada-io/karmada/pull/3475


@RainbowMango @Poor12 @RainbowMango What do you think?

chaunceyjiang:
> @chaunceyjiang, could you please help verify whether this fix can solve your problem? Great thanks.

Yes.  I also used this patch in my local env.

pkg/scheduler/scheduler.go
RainbowMango: > I think the graceful-eviction-controller should also handle create event.

What do you mean `handle create event`?

RainbowMango: Do we need to do something else during this PR?

chaunceyjiang: > handle create event.

https://github.com/karmada-io/karmada/blob/master/pkg/controllers/gracefuleviction/rb_graceful_eviction_controller.go#L87

chaunceyjiang: > Do we need to do something else during this PR?

No, This PR can solve my problem .

RainbowMango (approved):
/lgtm
/approve

PR #3457, opened 2023-04-24
add e2e to test dependency distribution base on ClusterPropagationPolicy

### LINKED ISSUE #3265: Add E2E to test dependency distribution base on ClusterPropagationPolicy
As stated in #3239, in order to test the dependency distribution base on `ClusterPropagationPolicy`, we need to add some e2e case.

We can do it refer to the existing cases: 

https://github.com/karmada-io/karmada/blob/master/test/e2e/dependenciesdistributor_test.go#L16

lxtywypc:
**What type of PR is this?**
/kind cleanup

Add some e2e test case for dependency distribution of resources which are propagated based on clusterPropagationPolicies 

Fixes #3265

```release-note

```

### DIFF
--- a/test/e2e/dependenciesdistributor_test.go
+++ b/test/e2e/dependenciesdistributor_test.go
@@ -14,7 +14,7 @@ import (
 )
 
 var _ = ginkgo.Describe("[DependenciesDistributor] automatically propagate relevant resources testing", func() {
-	ginkgo.Context("dependencies propagation testing", func() {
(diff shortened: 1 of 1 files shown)


lxtywypc:
@XiShanYongYe-Chang I'm sorry for filling this late. I think this part is just to make sure that dependency distribution can work expectly as well when by clusterPropagationPolicies, so I only add the case of configMap. If you consider that we also need test for other kinds of resources, I will add more then.

XiShanYongYe-Chang:
Thanks a lot~
/assign

Issue #3169, opened 2023-02-17
[Umbrella] Implement application failover


Poor12:
Refer to #3133 

Task breakdown list:

Failover based on application state:
- [ ] Introduce a new `Failover` field for `PropagationSpec`.
- [ ] Binding API modification adaptation to implement application failover.
- [ ] `karmada-controller-manager`: introduce a new controller which will watch the change of the health field of ResourceBindingStatus, determine whether the failover condition is reached, and trigger the failover.
- [ ] `karmada-scheduler`: Add a new filter plugin, to filter the evicted cluster.
- [ ] `karmada-webhook`: Perform extra validation work to prevent misleading configuration.
- [ ] E2E
- [ ] Document update.

RainbowMango:
We haven't made the final decision yet. Please list the iteration tasks in the proposal.

And you can re-open it once we reached an agreement. 
/close

PR #3461, opened 2023-04-24
Add application failover controller

Poor12:
**What type of PR is this?**
/kind feature

Add application failover controller(overall architecture).
This PR may depend on #3283.

Part of #3169 

None


RainbowMango (commented):
Please assign it to me once it is ready for review.

Poor12:
/assign @RainbowMango 
Keep the codegen aside and I will fix it with latest comments together.

pkg/apis/work/v1alpha2/well_known_constants.go
```diff
@@ -82,6 +82,10 @@ const (
 	// EvictionReasonTaintUntolerated describes the eviction is triggered
 	// because can not tolerate taint or exceed toleration period of time.
 	EvictionReasonTaintUntolerated = "TaintUntolerated"
+
+	// EvictionReasonWorkloadUnhealthy describes the eviction is triggered
+	// because workload is unhealthy and reaches the condition of ApplicationFailoverBehavior.
+	EvictionReasonWorkloadUnhealthy = "WorkloadUnhealthy"
```
RainbowMango, line 88:
```suggestion
	EvictionReasonApplicationFailure = "ApplicationFailure"
```

pkg/controllers/applicationfailover/common.go
```diff
@@ -0,0 +1,79 @@
...
+	m.workloadUnhealthy[key] = unHealthyClusters
+}
+
+func (m *workloadUnhealthyMap) get(key keys.ClusterWideKey) map[string]metav1.Time {
+	m.RLock()
+	defer m.RUnlock()
+	return m.workloadUnhealthy[key]
+}
+
+func detectFailure(aggregatedStatusItems []workv1alpha2.AggregatedStatusItem, gracefulEvictionTasks []workv1alpha2.GracefulEvictionTask,
```
RainbowMango, line 46: This function couples too many responsibilities:
1. manages the unhealthy map
2. filter clusters that are already in the eviction tasks
3. select clusters that need to be evicted after grace period

We should consider splitting it into small functions.

pkg/controllers/applicationfailover/common.go
```diff
@@ -0,0 +1,79 @@
...
+	unhealthyClusters map[string]metav1.Time, tolerationSeconds *int32) (bool, []string, map[string]struct{}) {
+	var needEvictClusters []string
+	needSecondDetection := false
```
RainbowMango, line 49: What's the meaning of `second delection`?

pkg/util/helper/cluster.go
```diff
@@ -21,3 +22,14 @@ func IsAPIEnabled(APIEnablements []clusterv1alpha1.APIEnablement, groupVersion s
 
 	return false
 }
+
+// CheckIfClusterInGracefulEvictionTasks checks if the target cluster is in the GracefulEvictionTasks.
+func CheckIfClusterInGracefulEvictionTasks(gracefulEvictionTasks []workv1alpha2.GracefulEvictionTask, clusterName string) bool {
```
RainbowMango, line 27: ~~CheckIf~~ClusterInGracefulEvictionTasks would be better.

pkg/controllers/applicationfailover/crb_application_failover_controller.go
```diff
@@ -0,0 +1,212 @@
...
+	resource, err := helper.ConstructClusterWideKey(binding.Spec.Resource)
+	if err != nil {
+		klog.Errorf("failed to get key from binding(%s)'s resource", binding.Name)
+		return 0, err
+	}
+
+	unhealthyClusters := c.workloadUnhealthyMap.get(resource)
+	if unhealthyClusters == nil {
+		unhealthyClusters = make(map[string]metav1.Time)
+	}
```
RainbowMango, line 90: Why do this initialization in a rush, can we do it after the eviction decision?

pkg/controllers/applicationfailover/crb_application_failover_controller.go
```diff
@@ -0,0 +1,212 @@
...
+	if retryDuration > 0 {
+		klog.V(4).Infof("Retry to check health status of the workload after %v minutes.", retryDuration.Minutes())
+		return controllerruntime.Result{RequeueAfter: retryDuration}, nil
+	}
+	return controllerruntime.Result{}, nil
+}
+
+func (c *CRBApplicationFailoverController) syncBinding(binding *workv1alpha2.ClusterResourceBinding) (time.Duration, error) {
+	// cleanup expired cluster
```
RainbowMango, line 79: This comment might be in the wrong place.

pkg/controllers/applicationfailover/crb_application_failover_controller.go
Poor12: `UnhealthyClusters` is used to check whether the cluster need to be evicted. I think it need to be inited before detecting failure.

pkg/controllers/applicationfailover/common.go
Poor12: It means when controller first finds the application on one cluster is unhealthy, it will detect the application after the tolerationSeconds is reached for the second time.

pkg/controllers/applicationfailover/common.go
Poor12: Through my analyze, there is a dependence in it because `unhealthy` map is used to select clusters that need to be evicted. We need to maintain the map meanwhile when we check whether the cluster need to be evicted.

pkg/controllers/applicationfailover/rb_application_failover_controller.go
```diff
@@ -0,0 +1,249 @@
...
+
+	// Cleanup clusters that have been evicted or removed in the workloadUnhealthyMap
+	c.workloadUnhealthyMap.deleteIrrelevantClusters(resource, allClusters)
+
+	return time.Duration(duration) * time.Second, nil
+}
+
+func (c *RBApplicationFailoverController) evictBinding(binding *workv1alpha2.ResourceBinding, clusters []string) error {
+	for _, cluster := range clusters {
+		if binding.Spec.Failover.Application.PurgeMode == policyv1alpha1.Graciously {
```
chaunceyjiang, line 137: Can we use `switch` statement ?

pkg/controllers/applicationfailover/rb_application_failover_controller.go
```diff
@@ -0,0 +1,249 @@
...
+			if !c.ResourceInterpreter.HookEnabled(resourceKey.GroupVersionKind(), configv1alpha1.InterpreterOperationInterpretHealth) {
+				return false
+			}
+
+			if !newObj.Spec.PropagateDeps {
+				return false
+			}
+			return true
+		},
+		DeleteFunc:  func(deleteEvent event.DeleteEvent) bool { return false },
```
chaunceyjiang, line 241: Why use `return false` here?

Poor12: For this controller, we only deal with the `update` event which updates the `deleteTimeStamp` in the `ResourceBinding`.

pkg/controllers/applicationfailover/rb_application_failover_controller.go
Poor12: For the statement of three conditional branches, I think it's ok to use if-else here.

RainbowMango: Yes, switch-case would be more clear.

pkg/controllers/applicationfailover/rb_application_failover_controller.go
RainbowMango: I guess because @Poor12 thought there is nothing to do for the delete event. 

In my opinion, we should clean up the `workloadUnhealthyMap` at 
```
	if err := c.Client.Get(ctx, req.NamespacedName, binding); err != nil {
		}
	}
```
instead of `if !binding.DeletionTimestamp.IsZero()`.

pkg/controllers/applicationfailover/rb_application_failover_controller.go
```diff
@@ -0,0 +1,249 @@
...
+
+func (c *RBApplicationFailoverController) syncBinding(binding *workv1alpha2.ResourceBinding) (time.Duration, error) {
+	tolerationSeconds := binding.Spec.Failover.Application.DecisionConditions.TolerationSeconds
+	resource, err := helper.ConstructClusterWideKey(binding.Spec.Resource)
+	if err != nil {
+		klog.Errorf("failed to get key from binding(%s)'s resource", binding.Name)
+		return 0, err
+	}
+
+	allClusters := make(map[string]struct{})
```
chaunceyjiang, line 109: Can we use `sets.New[string]`?

RainbowMango: +1

pkg/controllers/applicationfailover/rb_application_failover_controller.go
RainbowMango: > I am a little confused about the benefit between these two solutions.

The state of `DeletionTimestamp` is transitory, the controller might miss the event. Take RB deletion process as an example:
```
time1: start to delete the RB, requeue the RB after setting the `DeletionTimestamp`.
time2: RB is deleted, requeue the RB.
time3: The controller starts to reconcile the RB, it can't get the RB anymore.
```

pkg/controllers/applicationfailover/rb_application_failover_controller.go
```diff
@@ -0,0 +1,249 @@
...
+			if (oldObj.Spec.Failover == nil || oldObj.Spec.Failover.Application == nil) &&
+				(newObj.Spec.Failover == nil || newObj.Spec.Failover.Application == nil) {
+				return false
+			}
+
+			if len(newObj.Status.AggregatedStatus) == 0 {
+				return false
+			}
+
+			resourceKey, err := helper.ConstructClusterWideKey(newObj.Spec.Resource)
```
chaunceyjiang, line 225: Can we put this `if` statement to `L187`?

pkg/controllers/applicationfailover/rb_application_failover_controller.go
```diff
@@ -0,0 +1,249 @@
...
+				klog.Errorf("Failed to construct clusterWideKey from binding(%s/%s)", newObj.Namespace, newObj.Name)
+				return false
+			}
+
+			if !c.ResourceInterpreter.HookEnabled(resourceKey.GroupVersionKind(), configv1alpha1.InterpreterOperationInterpretHealth) {
+				return false
+			}
+
+			if !newObj.Spec.PropagateDeps {
+				return false
```
chaunceyjiang, line 237: If `ApplicationFailover `is set, does PP's `PropagateDeps` have to be true?

Poor12: Right.

pkg/controllers/applicationfailover/rb_application_failover_controller.go
```diff
@@ -0,0 +1,249 @@
...
+			}
+
+			resourceKey, err := helper.ConstructClusterWideKey(newObj.Spec.Resource)
+			if err != nil {
+				// Never reach
+				klog.Errorf("Failed to construct clusterWideKey from binding(%s/%s)", newObj.Namespace, newObj.Name)
+				return false
+			}
+
+			if !c.ResourceInterpreter.HookEnabled(resourceKey.GroupVersionKind(), configv1alpha1.InterpreterOperationInterpretHealth) {
```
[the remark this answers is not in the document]
Poor12: Sorry, I did not get what you mean. Could you please show the details?

pkg/controllers/applicationfailover/crb_application_failover_controller.go
```diff
@@ -0,0 +1,240 @@
...
+
+func (c *CRBApplicationFailoverController) clusterResourceBindingFilter(crb *workv1alpha2.ClusterResourceBinding) bool {
+	if len(crb.Status.AggregatedStatus) == 0 {
+		return false
+	}
+
+	resourceKey, err := helper.ConstructClusterWideKey(crb.Spec.Resource)
+	if err != nil {
+		// Never reach
+		klog.Errorf("Failed to construct clusterWideKey from binding(%s/%s)", crb.Namespace, crb.Name)
```
Poor12, line 231: Failed to construct clusterWideKey from clusterResourceBinding(%s), crb.Name

cmd/controller-manager/app/controllermanager.go
```diff
@@ -527,6 +529,31 @@ func startGracefulEvictionController(ctx controllerscontext.Context) (enabled bo
 	return false, nil
 }
 
+func startApplicationFailoverController(ctx controllerscontext.Context) (enabled bool, err error) {
+	if features.FeatureGate.Enabled(features.Failover) {
```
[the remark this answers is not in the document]
Poor12: I'd better keep the code style the same as `GracefulEvictionController`. I feel like just the code style differs.

RainbowMango:
@chaunceyjiang @jwcesign Please take another look.

pkg/controllers/applicationfailover/rb_application_failover_controller.go
```diff
@@ -0,0 +1,247 @@
...
+	if retryDuration > 0 {
+		klog.V(4).Infof("Retry to check health status of the workload after %v minutes.", retryDuration.Minutes())
+		return controllerruntime.Result{RequeueAfter: retryDuration}, nil
+	}
+	return controllerruntime.Result{}, nil
+}
+
+func (c *RBApplicationFailoverController) detectFailure(clusters []string, tolerationSeconds *int32, resource keys.ClusterWideKey) (int32, []string) {
+	var needEvictClusters []string
+	var duration int32
```
chaunceyjiang, line 86: What is this variable used for?

chaunceyjiang: This variable is always equal to `tolerationSeconds`, and `tolerationSeconds` has been constant. I haven't seen the purpose of this variable.

pkg/controllers/applicationfailover/crb_application_failover_controller.go
```diff
@@ -0,0 +1,244 @@
...
+}
+
+func (c *CRBApplicationFailoverController) detectFailure(clusters []string, tolerationSeconds *int32, resource keys.ClusterWideKey) (int32, []string) {
+	var needEvictClusters []string
+	var duration int32
+
+	for _, cluster := range clusters {
+		if !c.workloadUnhealthyMap.hasWorkloadBeenUnhealthy(resource, cluster) {
+			c.workloadUnhealthyMap.setTimeStamp(resource, cluster)
+			duration = *tolerationSeconds
```
jwcesign, line 91: I prefer to add continue, and delete `else`, which will make `Cyclomatic complexity` smaller.

chaunceyjiang: +1

pkg/controllers/applicationfailover/rb_application_failover_controller.go
Poor12: This variable is not always equal to `tolerationSeconds`. When there is no need to detect after the tolerance time, duration is equal to 0.

pkg/controllers/applicationfailover/common.go
```diff
@@ -0,0 +1,95 @@
...
+	"sync"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
+	"github.com/karmada-io/karmada/pkg/util/fedinformer/keys"
+)
+
+type workloadUnhealthyMap struct {
```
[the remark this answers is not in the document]
Poor12: I think these functions are used for application failover controllers. It will not be used in other places currently.

pkg/controllers/applicationfailover/crb_application_failover_controller.go
Poor12: +1

pkg/controllers/applicationfailover/rb_application_failover_controller.go
chaunceyjiang: Do we need a logic like this if if `metav1.Now().After(unHealthyTimeStamp.Add(time.Duration(*tolerationSeconds) * time.Second))` equal to `false` ?


``` go
duration = *tolerationSeconds - int32(metav1.Now().Sub(unHealthyTimeStamp.Time).Seconds())
```

Poor12: When the fault is detected for the first time, it will rejoin the queue after tolerationSeconds. I haven't thought of a scenario where I need to rejoin the queue halfway.

pkg/controllers/applicationfailover/rb_application_failover_controller.go
```diff
@@ -0,0 +1,247 @@
...
+
+	return time.Duration(duration) * time.Second, nil
+}
+
+func (c *RBApplicationFailoverController) evictBinding(binding *workv1alpha2.ResourceBinding, clusters []string) error {
+	for _, cluster := range clusters {
+		switch binding.Spec.Failover.Application.PurgeMode {
+		case policyv1alpha1.Graciously:
+			if features.FeatureGate.Enabled(features.GracefulEviction) {
+				binding.Spec.GracefulEvictCluster(cluster, workv1alpha2.NewTaskOptions(workv1alpha2.WithProducer(RBApplicationFailoverControllerName), workv1alpha2.WithReason(workv1alpha2.EvictionReasonApplicationFailure), workv1alpha2.WithGracePeriodSeconds(binding.Spec.Failover.Application.GracePeriodSeconds)))
```
chaunceyjiang, line 145: Split into multiple lines.

pkg/controllers/applicationfailover/rb_application_failover_controller.go
chaunceyjiang: >  I haven't thought of a scenario where I need to rejoin the queue halfway.

 I think this is necessary.

pkg/controllers/applicationfailover/rb_application_failover_controller.go
```diff
@@ -0,0 +1,259 @@
...
+			if duration > *tolerationSeconds {
+				duration = *tolerationSeconds
+			}
+			continue
+		}
+		// When the workload in a cluster is in an unhealthy state for more than the tolerance time,
+		// and the cluster has not been in the GracefulEvictionTasks,
+		// the cluster will be added to the list that needs to be evicted.
+		unHealthyTimeStamp := c.workloadUnhealthyMap.getTimeStamp(resource, cluster)
+		if metav1.Now().After(unHealthyTimeStamp.Add(time.Duration(*tolerationSeconds) * time.Second)) {
```
chaunceyjiang, line 101: I think it's better to use a single `metav1.Now()` consistently instead of using the latest time every time.

chaunceyjiang: like:
``` go
now:= metav1.Now()
now.After()
now.Sub()

```

pkg/controllers/applicationfailover/rb_application_failover_controller.go
```diff
@@ -0,0 +1,247 @@
...
+				needEvictClusters = append(needEvictClusters, cluster)
+			}
+		}
+	}
+
+	return duration, needEvictClusters
+}
+
+func (c *RBApplicationFailoverController) syncBinding(binding *workv1alpha2.ResourceBinding) (time.Duration, error) {
+	tolerationSeconds := binding.Spec.Failover.Application.DecisionConditions.TolerationSeconds
```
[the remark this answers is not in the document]
Poor12: It will be a seperate PR for it.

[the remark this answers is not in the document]
chaunceyjiang: > It will be a seperate PR for it.



Is it a new controller?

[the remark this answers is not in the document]
RainbowMango: This is a good question. 
That is because the API change of `RB` for `PreConditions` hasn't been ready yet.

I would recommend postponing the `PreConditions` to the next release,

pkg/controllers/applicationfailover/rb_application_failover_controller.go
Poor12: > I guess because @Poor12 thought there is nothing to do for the delete event.
> 
> In my opinion, we should clean up the `workloadUnhealthyMap` at
> 
> ```
> 	if err := c.Client.Get(ctx, req.NamespacedName, binding); err != nil {
> 		if apierrors.IsNotFound(err) {
> 			return controllerruntime.Result{}, nil
> 		}
> 		return controllerruntime.Result{Requeue: true}, err
> 	}
> ```
> 
> instead of `if !binding.DeletionTimestamp.IsZero()`.

I find that we cannot deal with it in the above logic. If apierrors.IsNotFound(err) == true which means that binding has been deleted, actually we cannot Get the resource of binding so that we do not know how to cleanup `workloadUnhealthyMap`.

pkg/controllers/applicationfailover/rb_application_failover_controller.go
RainbowMango: Please @chaunceyjiang take another look.

chaunceyjiang: Generally LGTM. Only a question. https://github.com/karmada-io/karmada/pull/3461#discussion_r1185872544

RainbowMango: In order to implement the `PreConditions` we have to record the read state that probably need to extend the API, but the API is not ready yet. As I mentioned at https://github.com/karmada-io/karmada/pull/3461#discussion_r1186612642.

RainbowMango (approved):
/lgtm
/approve

PR #3462, opened 2023-04-25
rename advancedcronjob directory

yike21:
**What type of PR is this?**
/kind cleanup

rename the advencedcronjob directory, which makes it more tidy. 
and fix the retention operation for broadcastjob customizations.

Fixes #

NONE

```release-note

```


pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/apps.kruise.io/v1alpha1/BroadcastJob/customizations.yaml
```diff
@@ -132,8 +132,8 @@ spec:
         function Retain(desiredObj, observedObj)
           if observedObj.spec.template.metadata ~= nil and observedObj.spec.template.metadata.labels ~= nil then
             desiredObj.spec.template.metadata.labels = observedObj.spec.template.metadata.labels
-            return desiredObj
```
yike21, line 135: The error I stumbled upon while re-reading the code. `karmada interpret --check` is a static check and didn't check for this problem. Also because `object.spec.template.metadata.labels` in testdata is always true, I neglected to see what happens when that property is false.

PR #3466, opened 2023-04-25
add ConcurrentDependenciesDistributorBindingSyncs and ConcurrentDependenciesDistributorResourceSyncs

snowplayfire:
**What type of PR is this?**

/kind cleanup
/kind feature

make it configurable to speed up clearing queues.

Fixes #

```release-note

```

### DIFF
--- a/cmd/controller-manager/app/controllermanager.go
+++ b/cmd/controller-manager/app/controllermanager.go
@@ -583,6 +583,8 @@ func setupControllers(mgr controllerruntime.Manager, opts *options.Options, stop
 			ResourceInterpreter: resourceInterpreter,
 			RESTMapper:          mgr.GetRESTMapper(),
 			EventRecorder:       mgr.GetEventRecorderFor("dependencies-distributor"),
+			ConcurrentDependenciesDistributorBindingSyncs:  opts.ConcurrentDependenciesDistributorBindingSyncs,
+			ConcurrentDependenciesDistributorResourceSyncs: opts.ConcurrentDependenciesDistributorResourceSyncs,
 		}
 		if err := mgr.Add(dependenciesDistributor); err != nil {
 			klog.Fatalf("Failed to setup dependencies distributor: %v", err)
--- a/cmd/controller-manager/app/options/options.go
+++ b/cmd/controller-manager/app/options/options.go
@@ -126,6 +126,11 @@ type Options struct {
 	// in scenario of dynamic replica assignment based on cluster free resources.
 	// Disable if it does not fit your cases for better performance.
 	EnableClusterResourceModeling bool
+
+	// ConcurrentDependenciesDistributorBindingSyncs is the number of bindings that are allowed to sync concurrently.
+	ConcurrentDependenciesDistributorBindingSyncs int
+	// ConcurrentDependenciesDistributorResourceSyncs is the number of resources that are allowed to sync concurrently.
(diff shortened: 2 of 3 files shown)


cmd/controller-manager/app/options/options.go
```diff
@@ -126,6 +126,9 @@ type Options struct {
 	// in scenario of dynamic replica assignment based on cluster free resources.
 	// Disable if it does not fit your cases for better performance.
 	EnableClusterResourceModeling bool
+
+	// ConcurrentDistributorSyncs is the number of bindings that are allowed to sync concurrently.
```
XiShanYongYe-Chang, line 130: Can here add more comment like `ConcurrentResourceTemplateSyncs`:
```
Larger number means responsive binding syncing but more CPU(and network) load.
```

pkg/dependenciesdistributor/dependencies_distributor.go
```diff
@@ -70,6 +70,9 @@ type DependenciesDistributor struct {
 	RESTMapper          meta.RESTMapper
 	ResourceInterpreter resourceinterpreter.ResourceInterpreter
 
+	// ConcurrentDistributorSyncs is the number of bindings that are allowed to sync concurrently.
+	ConcurrentDistributorSyncs int
```
XiShanYongYe-Chang, line 74: Suggest renaming it to `ConcurrentDependenciesDistributorSyncs`, this shows tath it is set up as dependent distribution.

snowplayfire: updated, please have a look.

XiShanYongYe-Chang (commented):
/lgtm
Thanks~

RainbowMango (commented):
/assign

Poor12:
I think we may need a release-note to remind users about these user-facing changes.

cmd/controller-manager/app/controllermanager.go
```diff
@@ -583,6 +583,8 @@ func setupControllers(mgr controllerruntime.Manager, opts *options.Options, stop
 			ResourceInterpreter: resourceInterpreter,
 			RESTMapper:          mgr.GetRESTMapper(),
 			EventRecorder:       mgr.GetEventRecorderFor("dependencies-distributor"),
+			ConcurrentDependenciesDistributorBindingSyncs:  opts.ConcurrentDependenciesDistributorBindingSyncs,
+			ConcurrentDependenciesDistributorResourceSyncs: opts.ConcurrentDependenciesDistributorResourceSyncs,
```
RainbowMango, line 587: We can reuse the `--concurrent-resource-template-syncs` and `--concurrent-resourcebinding-syncs` because the required concurrency depends on the number of `RB` and `resource templates`.

```suggestion
			ConcurrentDependenciesDistributorBindingSyncs:  opts.ConcurrentResourceBindingSyncs,
			ConcurrentDependenciesDistributorResourceSyncs: opts.ConcurrentResourceTemplateSyncs,
```
I'm concerned that introducing too many flags would complicate the configuration. For example, if the system holds a huge amount of resource templates, they probably need to set the `--concurrent-dependencies-distributor-resource-syncs`, but they should set the `--concurrent-resource-template-syncs` as well.

RainbowMango: What do you think? @snowplayfire

PR #3478, opened 2023-04-27
karmada-search: ResourceRegistry controller can't handle DeletedFinalStateUnknown it gets from controller framework

xigang:
**What type of PR is this?**

/kind bug

Fix panic in SearchController caused by receiving DeletedFinalStateUnknown object from the cache.

Fixes #

```
Fix panic in ResourceRegistry controller caused by receiving DeletedFinalStateUnknown object from the cache.
```

### DIFF
--- a/pkg/search/controller.go
+++ b/pkg/search/controller.go
@@ -331,7 +331,22 @@ func (c *Controller) updateResourceRegistry(oldObj, newObj interface{}) {
 
 // deleteResourceRegistry parse the resourceRegistry object and add deleted Cluster to the queue
 func (c *Controller) deleteResourceRegistry(obj interface{}) {
-	rr := obj.(*searchv1alpha1.ResourceRegistry)
+	rr, isRR := obj.(*searchv1alpha1.ResourceRegistry)
+	// We can get DeletedFinalStateUnknown instead of *searchv1alpha1.ResourceRegistry here and
+	// we need to handle that correctly.
+	if !isRR {
+		deletedState, ok := obj.(cache.DeletedFinalStateUnknown)
+		if !ok {
+			klog.ErrorS(nil, "Received unpexpected object", "object", obj)
+			return
+		}
+		rr, ok = deletedState.Obj.(*searchv1alpha1.ResourceRegistry)
+		if !ok {
+			klog.ErrorS(nil, "DeletedFinalStateUnknown contained non-ResourceRegistry object", "object", deletedState.Obj)
+			return
+		}
+	}
+
 	for _, cluster := range c.getClusters(rr.Spec.TargetCluster) {
 		v, ok := c.clusterRegistry.Load(cluster)
 		if !ok {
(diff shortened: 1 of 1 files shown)


XiShanYongYe-Chang:
Hi @xigang, thanks for your feedback.

Can you explain a little bit when a panic occurs? I wonder if we need to do this in other places for the delete event.

BTW, can you help update the release note.

xigang:
> Hi @xigang, thanks for your feedback.
> 
> Can you explain a little bit when a panic occurs? I wonder if we need to do this in other places for the delete event.
> 
> BTW, can you help update the release note.

@XiShanYongYe-Chang `DeleteFunc` is called when an existing resource is deleted. It gets the final state of the resource (if it is known). Otherwise, it gets an object of type DeletedFinalStateUnknown. This can happen if the watch is closed and misses the delete event and the controller doesn't notice the deletion until the subsequent re-list.

https://github.com/kubernetes/client-go/blob/master/tools/cache/delta_fifo.go#L800

XiShanYongYe-Chang:
@xigang Thanks for your explanation.

Ask @RainbowMango to help rerun the test, the ci failed for `setup environment` failed, it's no related with the updated code.

pkg/search/controller.go
```diff
@@ -331,7 +331,22 @@ func (c *Controller) updateResourceRegistry(oldObj, newObj interface{}) {
 
 // deleteResourceRegistry parse the resourceRegistry object and add deleted Cluster to the queue
 func (c *Controller) deleteResourceRegistry(obj interface{}) {
-	rr := obj.(*searchv1alpha1.ResourceRegistry)
+	rr, isRR := obj.(*searchv1alpha1.ResourceRegistry)
+	// We can get DeletedFinalStateUnknown instead of *clusterV1alpha1.Cluster here and
```
XiShanYongYe-Chang, line 335: clusterV1alpha1.Cluster -> searchv1alpha1.ResourceRegistry

XiShanYongYe-Chang (commented):
Hi @xigang, do you think we need to cherry-pick this patch to the previous branch?

pkg/search/controller.go
xigang: @XiShanYongYe-Chang The code comment has been fixed.

I think this patch need to cherry-pick the previous branch, and need to check the other controller `cache.ResourceEventHandlerFuncs` `DeleteFunc` dealt with the problem.

XiShanYongYe-Chang: > I think this patch need to cherry-pick the previous branch and need to check whether other controller of `DeleteFunc` handled `DeletedFinalStateUnknown` types of object.

Thanks~
Can you help do it?

XiShanYongYe-Chang (commented):
/lgtm
/approve

XiShanYongYe-Chang:
Can you help add the component name to the release note, just like that:
```
karmada-search: xxx
```

pkg/search/controller.go
xigang: > > I think this patch need to cherry-pick the previous branch and need to check whether other controller of `DeleteFunc` handled `DeletedFinalStateUnknown` types of object.
> 
> Thanks~ Can you help do it?

Okay, I will deal with this issue later.

xigang:
> Can you help add the component name to the release note, just like that:
> 
> ```
> karmada-search: xxx
> ```

done

XiShanYongYe-Chang:

PR #3494, opened 2023-05-05
Allow user decide whether to automatically generate Propagation-policy and its name when promote

### LINKED ISSUE #3429: Letting the user decide whether to automatically generate Propagation-policy when Promote
Adding a flag like `--auto-generate-propagation-policy=true` or `propagation-policy-name=*` to command
```shell
karmadactl promote <RESOURCE_TYPE> <RESOURCE_NAME> -n <NAME_SPACE> -C <CLUSTER_NAME>
```

**Why is this needed**:
My story is: I want to do a migration of non-federal applications on the legacy cluster to Karmada Control Plane, then propagate the application to other member clusters. I used the `promote` command and noticed it creates `propagation-policy` with a suffix in its succession. So I need to get the `pp` and edit it manually(which requires a little more work in a scription way).

if the `--auto-generate-propagation-policy=true` flag is supported, I can manually create the `pp` with my other specific requests rather than the default one.

if the `propagation-policy-name=*` flag is supported I don't have to distinguish which is the automatically generated `pp` and then edit it. And it will be easier to get all processes scripted.

Any of these flags will meet my needs here.

Since I have to edit the `propagation-policy` anyway, I think why not let me decide whether to create a default one or how -- decide its name, even more.

Ps. Isn't it more reasonable to let the user create the `propagation-policy` on their thoughts instead of `promote`? Or the `"Promote"` **IS** designed to imply the propagation part? What's your opinion?

LronDC:
**What type of PR is this?**

/kind feature

It's reasonable for users to let them decide whether to automatically create a policy and the name of auto-created PropagationPolicy.
Fixes #3429 

My first time contributing to Karmada so any advice and guidance are expected.
Is a test case needed here for this change?
And I think another PR of the document's change should be opened.

```release-note
`karmadactl`: Introduced `--auto-create-policy` and `--policy-name` flags for `promote` command to customize the policy during the promotion.
```


pkg/karmadactl/promote/promote.go
```diff
@@ -114,6 +114,14 @@ type CommandPromoteOption struct {
 	// DryRun tells if run the command in dry-run mode, without making any server requests.
 	DryRun bool
 
+	// AutoGeneratePropagationPolicy tells if the promote command create a PropagationPolicy automatically, otherwise one can create their own PropagationPolicy
+	AutoGeneratePropagationPolicy bool
```
jwcesign, line 118: `AutoCreatePropagationPolicy` is better?

pkg/karmadactl/promote/promote.go
```diff
@@ -114,6 +114,14 @@ type CommandPromoteOption struct {
...
+
+	// PropagationPolicyName is the name of a PropagationPolicy that automatically created by promote command,
+	// if not specified, a random suffix will be added to the PropagationPolicy.
+	// It is only valid when AutoGeneratePropagationPolicy is true.
+	PropagationPolicyName string
```
jwcesign, line 123: If this policy already exists, what will happen?

pkg/karmadactl/promote/promote.go
LronDC: Sure, will do

pkg/karmadactl/promote/promote.go
LronDC: before building one policy it will [check] if the name exists (https://github.com/karmada-io/karmada/pull/3494/files/1be3cf421633636a171dc4129069a249fd0569e0#diff-6d515eefa1bc0f6f344292ee9f555c0005bc04d0905307bb67357cb234b43ca6R402)

jwcesign:
Hi, @LronDC , please squash the two commits together.
Other lgtm, cc @lonelyCZ for checking

LronDC:
> Hi, @LronDC , please squash the two commits together. Other lgtm, cc @lonelyCZ for checking

Okay, may I wait for others' opinions? I will squash the commits after all changes be done.

jwcesign:
> I will squash the commits after all changes be done.

Sure.

pkg/karmadactl/promote/promote.go
```diff
@@ -127,6 +135,9 @@ type CommandPromoteOption struct {
 
 // AddFlags adds flags to the specified FlagSet.
 func (o *CommandPromoteOption) AddFlags(flags *pflag.FlagSet) {
+	flags.BoolVar(&o.AutoCreatePropagationPolicy, "auto-create-propagation-policy", true, "Create a PropagationPolicy automatically")
+	flags.StringVar(&o.PropagationPolicyName, "propagation-policy-name", "",
```
lonelyCZ, line 139: Could we only specify `propagation-policy-name`？When the name is nil, auto-gen the name, otherwise using the specific name.

What do you think? @LronDC @jwcesign

LronDC: You mean leave the pp creation automatically, just let the user decide its name, do I get it correctly?

jwcesign: Do you mean that there is only one option for `propagation-policy-name`? If that's the case, then the policy will be automatically created, which doesn't solve the issue of manual creation.

lonelyCZ: Got it, it has two phase.

I think it is ok.

pkg/karmadactl/promote/promote.go
```diff
@@ -127,6 +135,9 @@ type CommandPromoteOption struct {
...
+		"The name of a PropagationPolicy that automatically created by command. If not specified, a random suffix will be added to the PropagationPolicy")
```
lonelyCZ, line 140:
```suggestion
		"The name of a PropagationPolicy that automatically created by command. If not specified, a suffix will be added to the PropagationPolicy")
```

The suffix is not generated random. It is a specific hash code.

pkg/karmadactl/promote/promote.go
```diff
@@ -136,6 +147,7 @@ func (o *CommandPromoteOption) AddFlags(flags *pflag.FlagSet) {
 	flags.StringVar(&o.ClusterKubeConfig, "cluster-kubeconfig", "",
 		"Path of the legacy cluster's kubeconfig.")
 	flags.BoolVar(&o.DryRun, "dry-run", false, "Run the command in dry-run mode, without making any server requests.")
+
```
lonelyCZ, line 150: Please delete this blank line.

pkg/karmadactl/promote/promote.go
LronDC: Yes, you are right, nice catch! Thank you.

lonelyCZ:
Hi, @RainbowMango , please help to trigger the CI.

RainbowMango:
Done.

lonelyCZ:
Hi, @LronDC , have you tested it in your env? Could you please share the test result?

LronDC:
> Hi, @LronDC , have you tested it in your env? Could you please share the test result?

Yes. I tested it and it worked as expected. As you can see I failed last CI workflow on `e2e test` and I was tring to make it right, but bumped into some local env issues. I think I could figure it out today or tomorrow. Sorry for the late.

LronDC:
@RainbowMango  Hi, I failed the last CI on the e2e test, I've made some changes and tested it in my local. It worked fine and passed the e2e test now. Could you please help to trigger the CI again?

RainbowMango:
Done.

lonelyCZ:
I just tested it in my env, it also works fine as expected!

Very thanks @LronDC !

/approve

XiShanYongYe-Chang (commented):
For e2e part, LGTM

Because cross-file directories are involved, approve of @RainbowMango may be required.
/cc @RainbowMango

pkg/karmadactl/promote/promote.go
RainbowMango, line 139: Hold on, I'm looking at the flags.

pkg/karmadactl/promote/promote.go
```diff
@@ -127,6 +135,9 @@ type CommandPromoteOption struct {
 
 // AddFlags adds flags to the specified FlagSet.
 func (o *CommandPromoteOption) AddFlags(flags *pflag.FlagSet) {
+	flags.BoolVar(&o.AutoCreatePropagationPolicy, "auto-create-propagation-policy", true, "Create a PropagationPolicy automatically")
+	flags.StringVar(&o.PropagationPolicyName, "propagation-policy-name", "",
+		"The name of a PropagationPolicy that automatically created by command. If not specified, a suffix will be added to the PropagationPolicy")
```
RainbowMango, line 140: The flag `auto-create-propagation-policy` might be misleading, it would be better to describe it more clearly.

My suggestion:
```suggestion
	flags.BoolVar(&o.AutoCreatePolicy, "auto-create-policy", true,
		"Automatically create a PropagationPolicy for namespace-scoped resources or create a ClusterPropagationPolicy for cluster-scoped resources.")
	flags.StringVar(&o.PolicyName, "policy-name", "",
		"The name of the PropagationPolicy(or ClusterPropagationPolicy) that is automatically created after promotion. If not specified, the name will be the resource name with a random hash suffix.")
```

pkg/karmadactl/promote/promote.go
```diff
@@ -114,6 +114,14 @@ type CommandPromoteOption struct {
 	// DryRun tells if run the command in dry-run mode, without making any server requests.
 	DryRun bool
 
+	// AutoCreatePropagationPolicy tells if the promote command create a PropagationPolicy automatically, otherwise one can create their own PropagationPolicy
+	AutoCreatePropagationPolicy bool
+
+	// PropagationPolicyName is the name of a PropagationPolicy that automatically created by promote command,
+	// if not specified, a suffix will be added to the PropagationPolicy.
+	// It is only valid when AutoCreatePropagationPolicy is true.
+	PropagationPolicyName string
```
RainbowMango, line 123:
```suggestion
	// AutoCreatePolicy tells if the promote command should create the
	// PropagationPolicy(or ClusterPropagationPolicy) by default.
	// It defaults to true.
	AutoCreatePolicy bool
	
	// PolicyName is the name of the PropagationPolicy(or ClusterPropagationPolicy),
	// It defaults to the promoting resource name with a random hash suffix.
	// It will be ignored if AutoCreatePolicy is false.
	PolicyName string
```

RainbowMango (commented):
Please @lonelyCZ take another look at the suggestions.

pkg/karmadactl/promote/promote.go
LronDC: I think it's really more appropriate that include a description of namespace-scoped resources and cluster-scoped resources.

One thing. @lonelyCZ had mentioned that the default name suffix is not "random" but a specific hash code actually. Do we consider to add "random" word to the description?

lonelyCZ: How about `a specific hash suffix`?

LronDC: > How about `a specific hash suffix`?
If we are on a agreement here I'm gonna make the change.:)

RainbowMango: `with a hash suffix that is generated by resource metadata`?

RainbowMango: no big deal. I mean random because from user's respective he/she can't know the hash value and doesn't need to know.

LronDC: Got it.

LronDC: @RainbowMango Change be done. Need one more trigging to CI, and review labels 😄

RainbowMango (approved):
/lgtm
/approve

RainbowMango (commented):
This PR was blocked by fake testing, after the retesting it will be get merged automatically.

I just updated the release note according to the newly implementation.

PR #3500, opened 2023-05-08
Fix application failover

Poor12:
**What type of PR is this?**
/kind feature

This PR fix two bugs in application failover:
1. When the workload is in an unhealthy state and the tolerationSeconds has not yet expired, and the user manually clears the workload, the `workloadUnhealthyMap` will not be cleared, which will cause the tolerationSeconds to not work during the second application failover. To solve this issue, we use the `binding.NamespacedName` as the key instead of resource.gvk.
2. When the workload is in an unhealthy state and the tolerationSeconds has not yet expired, the workload came back to healthy now, we need to cleanup the cluster in `workloadUnhealthyMap`.

Fixes #

None

### DIFF
--- a/pkg/controllers/applicationfailover/common.go
+++ b/pkg/controllers/applicationfailover/common.go
@@ -4,37 +4,37 @@ import (
 	"sync"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
 
 	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
-	"github.com/karmada-io/karmada/pkg/util/fedinformer/keys"
 )
 
 type workloadUnhealthyMap struct {
 	sync.RWMutex
-	// key is the resource type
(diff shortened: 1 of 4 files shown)


Poor12:
Test cases:
1. No abnormal restart after upgrading karmada-controller-manager.
2. Failover is not configured in the pp.
3. Failover is configured in the policy, and propagateDeps is not enabled, and failover does not occur.
4. Configure failover in the policy, do not configure `interpretHealth` rules, and failover does not occur.
5. Failover is configured in the policy, purgeMode is `Graciously`, `Never` and `Immediately`, failover runs normally.
6. If the `failover` or `graceEviction` featureGates is not enabled, failover cannot run normally.
7. If a fault occurs after failover is normally enabled and the tolerationSeconds is not reached, propagateDeps is disabled or the `interpretHealth` rule is deleted, and failover does not occur.
8. If a fault occurs after failover is normally enabled and the tolerationSeconds is not reached, app came back to health before a second failover, failover runs normally.
9. If a fault occurs after failover is normally enabled and the tolerationSeconds is not reached, the app was deleted first and then re-deployed, and the failure occurred again.

RainbowMango (approved):
/lgtm
/approve

PR #3506, opened 2023-05-09
karmada operator: fixed adding altName to certification

calvin0327:
**What type of PR is this?**
/kind bug

When the service type of karmada apiserver is `ClusterIP`, we will add a empty dns or IP to AltNames object. It will thorw below error:

```shell
cannot parse IP address of length 0
```

Fixes #

```release-note
NONE
```


calvin0327:
@lonelyCZ @Poor12 /cc

PR #3510, opened 2023-05-09
karmada operator: implement installing karmada descheduler

calvin0327:
**What type of PR is this?**
/kind feature

Desheduler is one of the more common components that we need to support.

Here is test results:

add configuration in CR resource:

```yaml
kind: Karmada
metadata:
  name: karmada-demo
  namespace: test
spec:
  components:
    KarmadaDescheduler: {}
```

these pod of karmada descheduler is running:

```shell
NAME                                                    READY   STATUS    RESTARTS   AGE
karmada-demo-aggregated-apiserver-587bc5c697-dt6wh      1/1     Running   0          65s
karmada-demo-apiserver-55968d9f8c-ffdx7                 1/1     Running   0          106s
karmada-demo-controller-manager-64455f7fd4-49q5v        1/1     Running   0          58s
karmada-demo-descheduler-6d6ffd4d55-l7vw7               1/1     Running   0          57s
karmada-demo-etcd-0                                     1/1     Running   0          3m5s
karmada-demo-kube-controller-manager-584f978bbd-5r27x   1/1     Running   0          58s
karmada-demo-scheduler-6d77b7547-4mpr5                  1/1     Running   0          58s
karmada-demo-webhook-6f5944f5d8-p8pws                   1/1     Running   0          58s
```

Fixes #

```release-note
NONE
```

### DIFF
--- a/operator/pkg/controlplane/controlplane.go
+++ b/operator/pkg/controlplane/controlplane.go
@@ -7,6 +7,7 @@ import (
 	kuberuntime "k8s.io/apimachinery/pkg/runtime"
 	clientset "k8s.io/client-go/kubernetes"
 	clientsetscheme "k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/klog/v2"
 
 	operatorv1alpha1 "github.com/karmada-io/karmada/operator/pkg/apis/operator/v1alpha1"
 	"github.com/karmada-io/karmada/operator/pkg/constants"
@@ -24,8 +25,10 @@ func EnsureControlPlaneComponent(component, name, namespace string, client clien
 
 	deployment, ok := deployments[component]
 	if !ok {
-		return fmt.Errorf("no exist manifest for %s", component)
+		klog.Infof("Skip installing component %s(%s/%s)", component, namespace, name)
+		return nil
 	}
+
 	if err := apiclient.CreateOrUpdateDeployment(client, deployment); err != nil {
 		return fmt.Errorf("failed to create deployment resource for component %s, err: %w", component, err)
 	}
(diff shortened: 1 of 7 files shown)


operator/pkg/controlplane/controlplane.go
```diff
@@ -24,33 +25,49 @@ func EnsureControlPlaneComponent(component, name, namespace string, client clien
...
 	return nil
 }
 
 func getComponentManifests(name, namespace string, cfg *operatorv1alpha1.KarmadaComponents) (map[string]*appsv1.Deployment, error) {
-	kubeControllerManager, err := getKubeControllerManagerManifest(name, namespace, cfg.KubeControllerManager)
-	if err != nil {
-		return nil, err
+	manifest := map[string]*appsv1.Deployment{}
+
+	if cfg.KubeControllerManager != nil {
```
Poor12, line 41: When is cfg.KubeControllerManager eq to nil?

operator/pkg/controlplane/controlplane.go
```diff
@@ -24,33 +25,49 @@ func EnsureControlPlaneComponent(component, name, namespace string, client clien
 
 	deployment, ok := deployments[component]
 	if !ok {
-		return fmt.Errorf("no exist manifest for %s", component)
+		klog.V(2).InfoS("skip applying karmada component", "component", component, "karmada", klog.KRef(namespace, name))
```
Poor12, line 28: This log is a little strange.

operator/pkg/controlplane/controlplane.go
calvin0327: Currently, kubeControllerManager  can not be nil, but the KarmadaDescheduler could be nil,  here we'd better check if them is nil.

operator/pkg/controlplane/controlplane.go
calvin0327: How about  `Skip installing karmada componnet %s", component, "karmada", klog.KRef(namespace, name)`

operator/pkg/controlplane/controlplane.go
calvin0327: https://github.com/karmada-io/karmada/blob/master/operator/pkg/init.go#L225
We set defaults to karmada cr here, the mandatory components can not be nil.  I thinks add a check in func `getComponentManifests` is more rigorous.

Poor12: I feel like it's a bit overprotective and it will make people who look at the code think that this may be equal to nil.

operator/pkg/controlplane/controlplane.go
Poor12: Can you give a example about this log?

operator/pkg/controlplane/controlplane.go
calvin0327: ok

operator/pkg/controlplane/controlplane.go
calvin0327: example for karmada descheduler, if it's cfg is nil, will log the message: `skip installing component karmada-descheduler`.

operator/pkg/util/naming.go
```diff
@@ -8,66 +8,71 @@ import (
...
 }
 
-// KarmadaAggregatedAPIServerName return secret name of karmada aggregated apiserver
+// KarmadaAggregatedAPIServerName returns secret name of karmada-aggregated-apiserver
 func KarmadaAggregatedAPIServerName(karmada string) string {
 	return generateResourceName(karmada, "aggregated-apiserver")
 }
 
-// KarmadaEtcdName return karmada etcd name
+// KarmadaEtcdName returns name of karmada etcd
```
Poor12, line 41:
```suggestion
// KarmadaEtcdName returns name of karmada-etcd
```

operator/pkg/util/naming.go
```diff
@@ -8,66 +8,71 @@ import (
...
 func KarmadaEtcdName(karmada string) string {
 	return generateResourceName(karmada, "etcd")
 }
 
-// KarmadaEtcdClientName return karmada etcd client name
+// KarmadaEtcdClientName returns name of karmada etcd  client
```
Poor12, line 46:
```suggestion
// KarmadaEtcdClientName returns name of karmada-etcd client
```

PR #3511, opened 2023-05-10
enable concurrency for pp and cpp

zach593:
**What type of PR is this?**

/kind feature

Recently, we encountered a slow restart issue where it took 15 minutes to clear the items in queues. We suspect that the number of workers in pp/cpp could be one of the reasons for this problem. Therefore, it would be reasonable to add a concurrency configuration for this.

Although the pp/cpp reconcile function appears to be capable of handling concurrency, the number of workers is hardcoded as `1`. Please let me know if there is any particular reason for this.

Fixes #

```release-note
`karmada-controller-manager`: Introduced `--concurrent-propagation-policy-syncs`/`--concurrent-cluster-propagation-policy-syncs` flags to specify concurrent syncs for PropagationPolicy and ClusterPropagationPolicy.
```

### DIFF
--- a/cmd/controller-manager/app/controllermanager.go
+++ b/cmd/controller-manager/app/controllermanager.go
@@ -583,17 +583,19 @@ func setupControllers(mgr controllerruntime.Manager, opts *options.Options, stop
 	objectWatcher := objectwatcher.NewObjectWatcher(mgr.GetClient(), mgr.GetRESTMapper(), util.NewClusterDynamicClientSet, resourceInterpreter)
 
 	resourceDetector := &detector.ResourceDetector{
-		DiscoveryClientSet:              discoverClientSet,
-		Client:                          mgr.GetClient(),
-		InformerManager:                 controlPlaneInformerManager,
-		RESTMapper:                      mgr.GetRESTMapper(),
-		DynamicClient:                   dynamicClientSet,
-		SkippedResourceConfig:           skippedResourceConfig,
-		SkippedPropagatingNamespaces:    opts.SkippedNamespacesRegexps(),
-		ResourceInterpreter:             resourceInterpreter,
-		EventRecorder:                   mgr.GetEventRecorderFor("resource-detector"),
-		ConcurrentResourceTemplateSyncs: opts.ConcurrentResourceTemplateSyncs,
-		RateLimiterOptions:              opts.RateLimiterOpts,
+		DiscoveryClientSet:                      discoverClientSet,
+		Client:                                  mgr.GetClient(),
+		InformerManager:                         controlPlaneInformerManager,
+		RESTMapper:                              mgr.GetRESTMapper(),
+		DynamicClient:                           dynamicClientSet,
+		SkippedResourceConfig:                   skippedResourceConfig,
+		SkippedPropagatingNamespaces:            opts.SkippedNamespacesRegexps(),
+		ResourceInterpreter:                     resourceInterpreter,
+		EventRecorder:                           mgr.GetEventRecorderFor("resource-detector"),
+		ConcurrentPropagationPolicySyncs:        opts.ConcurrentPropagationPolicySyncs,
+		ConcurrentClusterPropagationPolicySyncs: opts.ConcurrentClusterPropagationPolicySyncs,
+		ConcurrentResourceTemplateSyncs:         opts.ConcurrentResourceTemplateSyncs,
+		RateLimiterOptions:                      opts.RateLimiterOpts,
 	}
 
 	if err := mgr.Add(resourceDetector); err != nil {
--- a/cmd/controller-manager/app/options/options.go
+++ b/cmd/controller-manager/app/options/options.go
@@ -110,6 +110,10 @@ type Options struct {
 	// ConcurrentNamespaceSyncs is the number of Namespace objects that are
 	// allowed to sync concurrently.
 	ConcurrentNamespaceSyncs int
+	// ConcurrentPropagationPolicySyncs is the number of PropagationPolicy that are allowed to sync concurrently.
+	ConcurrentPropagationPolicySyncs int
+	// ConcurrentClusterPropagationPolicySyncs is the number of ClusterPropagationPolicy that are allowed to sync concurrently.
+	ConcurrentClusterPropagationPolicySyncs int
 	// ConcurrentResourceTemplateSyncs is the number of resource templates that are allowed to sync concurrently.
 	ConcurrentResourceTemplateSyncs int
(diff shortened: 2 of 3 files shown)


cmd/controller-manager/app/controllermanager.go
```diff
@@ -3,6 +3,7 @@ package app
 import (
 	"context"
 	"flag"
+	"k8s.io/klog/v2"
```
XiShanYongYe-Chang, line 6: This line maybe needs to be put into the second import block.

XiShanYongYe-Chang (commented):
@zach593, thanks for your feedback, I think it's a reasonable update.

cmd/controller-manager/app/controllermanager.go
zach593: oops, didn't notice why that line was changed, fixed

XiShanYongYe-Chang (commented):
/lgtm

XiShanYongYe-Chang:
Hi @zach593 can you help add release note to describe the update?

RainbowMango (commented):
/assign
I'll take a look tomorrow.

cmd/controller-manager/app/options/options.go
```diff
@@ -202,6 +206,8 @@ func (o *Options) AddFlags(flags *pflag.FlagSet, allControllers, disabledByDefau
 	flags.IntVar(&o.ConcurrentResourceBindingSyncs, "concurrent-resourcebinding-syncs", 5, "The number of ResourceBindings that are allowed to sync concurrently.")
 	flags.IntVar(&o.ConcurrentWorkSyncs, "concurrent-work-syncs", 5, "The number of Works that are allowed to sync concurrently.")
 	flags.IntVar(&o.ConcurrentNamespaceSyncs, "concurrent-namespace-syncs", 1, "The number of Namespaces that are allowed to sync concurrently.")
+	flags.IntVar(&o.ConcurrentPropagationPolicySyncs, "concurrent-propagation-policy-syncs", 1, "The number of PropagationPolicy that are allowed to sync concurrently.")
+	flags.IntVar(&o.ConcurrentClusterPropagationPolicySyncs, "concurrent-cluster-propagation-policy-syncs", 1, "The number of ClusterPropagationPolicy that are allowed to sync concurrently.")
```
jwcesign, line 210: Do CPP and PP need to be configured separately? @XiShanYongYe-Chang

cmd/controller-manager/app/options/options.go
```diff
@@ -202,6 +206,8 @@ func (o *Options) AddFlags(flags *pflag.FlagSet, allControllers, disabledByDefau
 	flags.IntVar(&o.ConcurrentResourceBindingSyncs, "concurrent-resourcebinding-syncs", 5, "The number of ResourceBindings that are allowed to sync concurrently.")
 	flags.IntVar(&o.ConcurrentWorkSyncs, "concurrent-work-syncs", 5, "The number of Works that are allowed to sync concurrently.")
 	flags.IntVar(&o.ConcurrentNamespaceSyncs, "concurrent-namespace-syncs", 1, "The number of Namespaces that are allowed to sync concurrently.")
+	flags.IntVar(&o.ConcurrentPropagationPolicySyncs, "concurrent-propagation-policy-syncs", 1, "The number of PropagationPolicy that are allowed to sync concurrently.")
```
jwcesign, line 209: If it's slow by default, I prefer to set the default value bigger, like 3?

cmd/controller-manager/app/options/options.go
XiShanYongYe-Chang: For users, the usage of cpp and pp may be different, so separate control may be a little more fine-grained.

RainbowMango (commented):
I'd say it makes sense to add a flag to specify the concurrency, before looking into the code, I might want to ask some questions to better understand this issue.

> Recently, we encountered a slow restart issue where it took 15 minutes to clear the items in queues. 

I wonder which queue we are talking about here?  And How do you know it is the queue that is blocked starting?  Are there some logs or metrics?
How many resource templates are in your system?

zishen (commented):
Good to add concurrency configuration for pp and cpp
.

Poor12:
Hi @zach593, you can get metrics from karmada-controller-manager to prove that work queue is heavy to deal with items:
eg:
```
workqueue_adds_total{name="propagationPolicy reconciler"} 
workqueue_depth{name="propagationPolicy reconciler"}
```

RainbowMango:
> Good to add concurrency configuration for pp and cpp

Yes, I totally agree that allows people the specify the concurrency.

But I'm curious about the root cause of the issue:
> Recently, we encountered a slow restart issue where it took 15 minutes to clear the items in queues. 

We need to identify the root cause of the problem, develop solutions to fix it, and ultimately be able to verify it.

cmd/controller-manager/app/options/options.go
zach593: Before this pull request, concurrency was hardcoded as "1". I believe it would be better to set the default value to be the same as the value before the changes, so that users who did not modify this flag do not need to consider this change. If you think it's necessary, we can increase this value.

zach593:
image

this is monitoring data on grafana, for last time we restart the karmada-controller-manager. you could see pp reconciler used about 3-4 minutes to clear the queue items.

pp/cpp reconciler is the upstream reconciler of the resource detector, and the output (workload template) of the upstream reconciler (pp/cpp) will be added to the downstream (resource detector) queue. Therefore, if the upstream reconciler works slowly, the downstream will need to reconcile some items twice. The same situation can also occur between detector -> rb/crb, rb/crb -> work, and the slow processing speed of the upstream will amplify the number of reconciles that the downstream needs to do.

therefore, adding concurrency for pp/cpp is reasonable. This will reduce the total number of items that karmada-controller-manager needs to reconcile when restarted.

@RainbowMango @Poor12

zach593:
> Hi @zach593 can you help add release note to describe the update?

```
User can control the concurrency of PropagationPolicyWorker and ClusterPolicyReconcileWorker by using the flags `concurrent-propagation-policy-syncs` and `concurrent-cluster-propagation-policy-syncs`.
```

done

zach593:
> How about update like this:
> 
> ```
> karmada-controller-manager: add `concurrent-propagation-policy-syncs`, `concurrent-cluster-propagation-policy-syncs` flags to adjust pp/cpp syncs speed.
> ```

cool, done

RainbowMango (approved):
/approve

Many thanks to @zach593, we can go with the default concurrency (`1`), I appreciate all your effort on it.  If you have any evidence that we need to increase the concurrency, please feel free to send a PR. Thanks in advance. :)

PR #3517, opened 2023-05-11
Add cluster failover behavior

Poor12:
**What type of PR is this?**
/kind api-change

Adding cluster failover behavior can let users decide how to deal with applications when cluster failure occurs.

Currently we always do failover when detecting a cluster failure. All the behaviors are controlled by the global config.
But if we introduce cluster failover behavior and we define that if users do not config cluster failover behavior, Karmada will not do failover, This may introduce some changes to the original default behavior.

One optional solution is that we define the default cluster failover behavior just the same as before.

Fixes #

None


pkg/apis/policy/v1alpha1/propagation_types.go
```diff
@@ -178,7 +178,27 @@ type FailoverBehavior struct {
 	// Cluster indicates failover behaviors in case of cluster failure.
 	// If this value is nil, failover is disabled.
 	// +optional
-	// Cluster *ClusterFailoverBehavior `json:"cluster,omitempty"`
+	Cluster *ClusterFailoverBehavior `json:"cluster,omitempty"`
```
chaunceyjiang, line 181: What is the difference between this and the Failover that we currently have?

Poor12: Currently we have the failover capability, and we may need to open this cap to users for fine-grained configuration in the next step.

PR #3518, opened 2023-05-11
Fixed occasional errors with clusteraffinities

XiShanYongYe-Chang:
**What type of PR is this?**

/kind flake

In the `karmada-scheduler`, the patch is used to update the rb status. Before the update, deep copy is performed on the rb. As a result, the subsequent rb objects are still the old values, and patch errors may occur. 

So I modified this to keep the rb object up-to-date after the update to avoid affecting subsequent patch operations.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -549,15 +549,19 @@ func (s *Scheduler) scheduleResourceBindingWithClusterAffinities(rb *workv1alpha
 	return scheduleErr
 }
 
-func (s *Scheduler) patchScheduleResultForResourceBinding(oldBinding *workv1alpha2.ResourceBinding, placement string, scheduleResult []workv1alpha2.TargetCluster) error {
-	newBinding := oldBinding.DeepCopy()
-	if newBinding.Annotations == nil {
-		newBinding.Annotations = make(map[string]string)
+func (s *Scheduler) patchScheduleResultForResourceBinding(
+	rb *workv1alpha2.ResourceBinding,
+	placement string,
+	scheduleResult []workv1alpha2.TargetCluster,
+) error {
+	modifiedRB := rb.DeepCopy()
+	if modifiedRB.Annotations == nil {
+		modifiedRB.Annotations = make(map[string]string)
 	}
-	newBinding.Annotations[util.PolicyPlacementAnnotation] = placement
-	newBinding.Spec.Clusters = scheduleResult
+	modifiedRB.Annotations[util.PolicyPlacementAnnotation] = placement
+	modifiedRB.Spec.Clusters = scheduleResult
 
-	patchBytes, err := helper.GenMergePatch(oldBinding, newBinding)
+	patchBytes, err := helper.GenMergePatch(rb, modifiedRB)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


RainbowMango:
Does this PR try to address the flaking test?

https://github.com/karmada-io/karmada/actions/runs/5449884625/jobs/9914700141

```
Summarizing 2 Failures:
  [INTERRUPTED] [BasicPropagation] propagation testing Job propagation testing [DeferCleanup (Each)] job propagation testing
  /home/runner/work/karmada/karmada/test/e2e/propagationpolicy_test.go:382
  [FAIL] [ClusterAffinities] propagation testing when [ClusterPropagationPolicy] replica scheduling type is Duplicated schedule with multi clusterAffinity [It] propagate clusterRole and then update the cluster label
  /home/runner/work/karmada/karmada/test/e2e/framework/rbac.go:158

Ran 79 of 139 Specs in 578.314 seconds
FAIL! - Interrupted by Other Ginkgo Process -- 77 Passed | 2 Failed | 0 Pending | 60 Skipped
```

chaunceyjiang:
> Does this PR try to address the flaking test?

> https://github.com/karmada-io/karmada/actions/runs/5449884625/jobs/9914700141


Yes.

chaunceyjiang:
Have you already found the root cause of this issue? /cc @XiShanYongYe-Chang

XiShanYongYe-Chang:
> Have you already found the root cause of this issue?

I caught an error log earlier, but I seem to have lost it. When checking logs, it is found that the scheduler updates the status of the rb to an incorrect old status during subsequent update. As a result, the eventhandler does not meet the conditions during check.

RainbowMango:
another one:
https://github.com/karmada-io/karmada/actions/runs/5507318884/jobs/10037376281

pkg/scheduler/scheduler.go
chaunceyjiang, line 1: Can you remind me where the core modifications of this PR are? I feel like there haven't been any substantial changes compared to before.

XiShanYongYe-Chang: After the modification, for each patch operation of the rb object in the scheduler, the updated rb object will be in its latest state. This ensures that there will be no old state interference during the next patch.

RainbowMango:
another failing case on master:
https://github.com/karmada-io/karmada/actions/runs/5610539390/job/15200537208

XiShanYongYe-Chang:
This PR is ready, can you help review it?
/cc @RainbowMango @chaunceyjiang

RainbowMango (commented):
/assign

pkg/scheduler/scheduler.go
RainbowMango: Indeed, it's hard to figure out the core modifications.

pkg/scheduler/scheduler.go
```diff
@@ -809,21 +811,22 @@ func patchClusterBindingStatusWithAffinityName(karmadaClient karmadaclientset.In
...
 	if err != nil {
 		return err
 	}
 	if len(patchBytes) == 0 {
 		return nil
 	}
 
-	_, err = karmadaClient.WorkV1alpha2().ClusterResourceBindings().Patch(context.TODO(), crb.Name, types.MergePatchType, patchBytes, metav1.PatchOptions{}, "status")
+	_, err = karmadaClient.WorkV1alpha2().ClusterResourceBindings().Patch(context.TODO(), crb.Name, types.MergePatchType,
+		patchBytes, metav1.PatchOptions{}, "status")
```
RainbowMango, line 829: I'm afraid it doesn't make readability better.

pkg/scheduler/scheduler.go
```diff
@@ -809,21 +811,22 @@ func patchClusterBindingStatusWithAffinityName(karmadaClient karmadaclientset.In
...
-	updateCRB.Status.SchedulerObservedAffinityName = affinityName
-	return patchClusterResourceBindingStatus(karmadaClient, crb, updateCRB)
+	originalCRB := crb.DeepCopy()
+	crb.Status.SchedulerObservedAffinityName = affinityName
+	return patchClusterResourceBindingStatus(karmadaClient, originalCRB, crb)
 }
 
-func patchClusterResourceBindingStatus(karmadaClient karmadaclientset.Interface, crb, updateCRB *workv1alpha2.ClusterResourceBinding) error {
-	patchBytes, err := helper.GenMergePatch(crb, updateCRB)
+func patchClusterResourceBindingStatus(karmadaClient karmadaclientset.Interface, originalCRB, crb *workv1alpha2.ClusterResourceBinding) error {
```
RainbowMango, line 819: How about naming the parameter with `originalxxx` and `modifiedxxx`？

pkg/scheduler/scheduler.go
```diff
@@ -809,21 +811,22 @@ func patchClusterBindingStatusWithAffinityName(karmadaClient karmadaclientset.In
 	}
 
 	klog.V(4).Infof("Begin to patch status with affinityName(%s) to ClusterResourceBinding(%s).", affinityName, crb.Name)
-	updateCRB := crb.DeepCopy()
-	updateCRB.Status.SchedulerObservedAffinityName = affinityName
-	return patchClusterResourceBindingStatus(karmadaClient, crb, updateCRB)
+	originalCRB := crb.DeepCopy()
+	crb.Status.SchedulerObservedAffinityName = affinityName
+	return patchClusterResourceBindingStatus(karmadaClient, originalCRB, crb)
```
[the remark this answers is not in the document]
XiShanYongYe-Chang: updated

XiShanYongYe-Chang:
/hold
Let me think about whether it is possible to consolidate the update operations as much as possible, in order to reduce conflicts and impacts caused by multiple updates in different places.

XiShanYongYe-Chang:
I think using the patch method in #4094 should be able to solve the current issue with patch errors. What do you think? @chaunceyjiang

XiShanYongYe-Chang:
It seems that the issue has not occurs for a long time.
/close

PR #3448, opened 2023-04-21
Support rerun github failed workflow

liangyuanpeng:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

Create a issue comment of include "/retest" to  rerun workflow when CI failed.

**Which issue(s) this PR fixes**:
Fixes #3421

**Special notes for your reviewer**:

This event will only trigger a workflow run if the workflow file is on the default branch.

So you can comment at [this PR](https://github.com/liangyuanpeng/karmada/pull/8) for check what the happen when we create a issue comment of "/retest" on the PR.(Yes, rerun the failed workflow.)

Also run `hack/ghaction.sh` on the local,  Just like:

```shell
lan@lan:~/repo/git/karmada/hack$ GH_TOKEN=xxx  ISSUE_COMMENT="hello\r\n/retest" PR_NUM=8 REPO=liangyuanpeng/karmada ./ghaction.sh
Matching /retest and rerun workflow...
Reruning workflow...
PR:test retest github action 
URL:https://github.com/liangyuanpeng/karmada/pull/8 
ID:4763218620 
WorkflowRunName:CLI 
===============================
✓ Requested rerun of run 4763218620
Reruning workflow...
PR:test retest github action 
URL:https://github.com/liangyuanpeng/karmada/pull/8 
ID:4763218619 
WorkflowRunName:CI Workflow 
===============================
✓ Requested rerun of run 4763218619
Reruning workflow...
PR:test retest github action 
URL:https://github.com/liangyuanpeng/karmada/pull/8 
ID:4763218327 
WorkflowRunName:CLI 
===============================
✓ Requested rerun of run 4763218327
Reruning workflow...
PR:test retest github action 
URL:https://github.com/liangyuanpeng/karmada/pull/8 
ID:4763218326 
WorkflowRunName:CI Workflow 
===============================
✓ Requested rerun of run 4763218326
```

/cc @helen-frank @RainbowMango  @whitewindmills 

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

Issue #3424, opened 2023-04-18
Fix: func retainPodFields not include initContainer'volumeMounts


neteric:
**What happened**:
When I propagation a Pod with initContainer, I found that the status of the pod was not reflected to rb.Status

**What you expected to happen**:
Propagate a Pod with initContainer, pod's status can reflect to rb and pod resourceTemplate's status

**How to reproduce it (as minimally and precisely as possible)**:
Run a pod with initContainer in control panel，and create a propagationPolicy to propagation it to member cluster

**Anything else we need to know?**:
**Environment**:
- Karmada version:
v1.5.0
- kubectl-karmada or karmadactl version (the result of `kubectl-karmada version` or `karmadactl version`):
- Others:

/assign

RainbowMango:
> Run a pod with initContainer in control panel，and create a propagationPolicy to propagation it to member cluster

Could you please share the yamls?

chaunceyjiang:
Yes, I also noticed this issue. The volumesMounts of the initContainer should be retained.

chaunceyjiang:
![image](https://user-images.githubusercontent.com/17962021/232717654-f02b28c9-cc22-4534-a000-6d90526d39fd.png)

chaunceyjiang:
<img width="771" alt="image" src="https://user-images.githubusercontent.com/17962021/232719432-9bc7a0c4-3a5e-4b08-81c8-7009e5405b22.png">

neteric:
> > Run a pod with initContainer in control panel，and create a propagationPolicy to propagation it to member cluster
> 
> Could you please share the yamls?
just like this: 
``` 
apiVersion: v1
kind: Pod
metadata:
  name: myapp-pod
  labels:
    app.kubernetes.io/name: MyApp
spec:
  containers:
  - name: myapp-container
    image: busybox:1.28
    command: ['sh', '-c', 'echo The app is running! && sleep 3600']
  initContainers:
  - name: init-myservice
    image: busybox:1.28
    command: ['sh', '-c', "until nslookup myservice.$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace).svc.cluster.local; do echo waiting for myservice; sleep 2; done"]
  - name: init-mydb
    image: busybox:1.28
    command: ['sh', '-c', "until nslookup mydb.$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace).svc.cluster.local; do echo waiting for mydb; sleep 2; done"]

```

PR #3526, opened 2023-05-11
Fields that do not support updates of pod should be retained

### LINKED ISSUE #3525: [Bug] Got an error when updating unsupported fields of pod
**What happened**:
After propagating a pod to some member clusters, I'll get an error if I try to update some unsupported fields of pod.
```
error: Pod "demo" is invalid: spec: Forbidden: pod updates may not change fields other than `containers[*].image` or `spec.activeDeadlineSeconds`
```

Unsupported fields of pod should be retained.

Of course, only a few fields support being updated, in place.
https://kubernetes.io/docs/concepts/workloads/pods/#pod-update-and-replacement

**Environment**:
- Karmada version: all version
- Others:

whitewindmills:
**What type of PR is this?**
/kind bug

Fixes #3525

```release-note
karmada-controller-manager: Retain fields that do not support updates of pod.
```


XiShanYongYe-Chang (commented):
It seems that we need to split this method into smaller methods.

pkg/resourceinterpreter/default/native/retain.go
```diff
@@ -40,29 +41,43 @@ func retainPodFields(desired, observed *unstructured.Unstructured) (*unstructure
...
-		for desiredIndex, desiredContainer := range desiredPod.Spec.Containers {
+	// Pod updates may not change fields other than `spec.containers[*].image`, `spec.initContainers[*].image`,
+	// `spec.activeDeadlineSeconds`, `spec.tolerations` (only additions to existing tolerations) or
+	// `spec.terminationGracePeriodSeconds` (allow it to be set to 1 if it was previously negative).
+	// More info: https://kubernetes.io/docs/concepts/workloads/pods/#pod-update-and-replacement
+	for clusterIndex, clusterContainer := range clusterPod.Spec.InitContainers {
+		for _, desiredContainer := range desiredPod.Spec.InitContainers {
 			if desiredContainer.Name == clusterContainer.Name {
-				desiredPod.Spec.Containers[desiredIndex].VolumeMounts = clusterContainer.VolumeMounts
+				clusterPod.Spec.InitContainers[clusterIndex].Image = desiredContainer.Image
```
chaunceyjiang, line 51: I have two questions.
1. Why delete the `VolumeMounts` field?   Refer to https://github.com/karmada-io/karmada/issues/3424
2.  Can this image field be modified in the control panel?

whitewindmills: > Why delete the VolumeMounts field? Refer to https://github.com/karmada-io/karmada/issues/3424

No, I did not give up retaining `VolumeMounts `. You can see that `clusterPod` is returned instead of `desiredPod`.

> Can this image field be modified in the control panel?

Sure.

chaunceyjiang: > No, I did not give up retaining VolumeMounts . You can see that clusterPod is returned instead of desiredPod.

Got it.

> Can this image field be modified in the control panel?

Sorry, I didn't explain my idea clearly. 

What I meant was whether we can modify fields other than the `image` field? 

For example, the `label` field?

whitewindmills: Of course, you can. But you should know that _Pod_ update operations have some limitations.
More info: https://kubernetes.io/docs/concepts/workloads/pods/#pod-update-and-replacement

chaunceyjiang: > clusterPod is returned instead of `desiredPod`.

 I think the newly added `label` should be in `desiredPod`. But you returned `clusterPod`, right?

whitewindmills: > I think the newly added label should be in desiredPod. But you returned clusterPod, right?

It has nothing to do with this PR. You'll figure it out with function [retainClusterFields](https://github.com/karmada-io/karmada/blob/master/pkg/util/objectwatcher/objectwatcher.go#L99).

whitewindmills:
@RainbowMango 
Kindly help to rerun test.

pkg/resourceinterpreter/default/native/retain.go
chaunceyjiang: Have you tested it?

chaunceyjiang: For example, add a new  label field.

chaunceyjiang: If add add a new label, I don't think it will propagate to the member cluster.

whitewindmills: Sorry, I did not get you. I know what you mean now. It does need to be fixed. Thanks.

liangyuanpeng:
> Kindly help to rerun test.
@whitewindmills 

Before PR https://github.com/karmada-io/karmada/pull/3448 is merge youcan commit some message for rerun github action CI, other way is let the PR exist on yourself repo, just like https://github.com/liangyuanpeng/karmada/pulls , you don't need to way someone to rerun github action.

whitewindmills:
@liangyuanpeng Thanks for your suggestions.

RainbowMango:
> @RainbowMango Kindly help to rerun test.

Done.

pkg/resourceinterpreter/default/native/retain.go
chaunceyjiang: I have another question.

Currently, if users use an OverridePolicy to modify these fields(other than `spec.containers[*].image, spec.initContainers[*].image, spec.activeDeadlineSeconds or spec.tolerations`.), users will get some error messages. but If we retain these fields, use can't get any error message. is this expected ?

whitewindmills: It makes no sense, I think. For example, I use an OverridePolicy to modify resource's `ownerReferences`, I also won't see any error message. People should know thoes restrictions of K8s and Karmada.

whitewindmills:
@RainbowMango @XiShanYongYe-Chang 
pls check for merging when you're free.

pkg/resourceinterpreter/default/native/retain.go
```diff
@@ -40,34 +41,62 @@ func retainPodFields(desired, observed *unstructured.Unstructured) (*unstructure
...
+	// `spec.activeDeadlineSeconds`, `spec.tolerations` (only additions to existing tolerations) or
+	// `spec.terminationGracePeriodSeconds` (allow it to be set to 1 if it was previously negative).
+	// More info: https://kubernetes.io/docs/concepts/workloads/pods/#pod-update-and-replacement
+	clusterPod.Labels = desiredPod.Labels
+	clusterPod.Annotations = desiredPod.Annotations
+	updatePodImages(desiredPod, clusterPod)
+	updatePodActiveDeadlineSeconds(desiredPod, clusterPod)
+	updatePodTolerations(desiredPod, clusterPod)
+
+	unstructuredObj, err := helper.ToUnstructured(clusterPod)
```
XiShanYongYe-Chang, line 54: The returned object is clusterPod, which is different from the processing logic in other places and may misunderstand other developers or code readers. Can we add some redundant processing to maintain the original logic habits?

pkg/resourceinterpreter/default/native/retain.go
```diff
@@ -40,34 +41,62 @@ func retainPodFields(desired, observed *unstructured.Unstructured) (*unstructure
...
+		clusterPod.Spec.ActiveDeadlineSeconds = desiredPod.Spec.ActiveDeadlineSeconds
 	}
+}
 
-	return unstructuredObj, nil
+func updatePodTolerations(desiredPod, clusterPod *corev1.Pod) {
+	for _, desiredToleration := range desiredPod.Spec.Tolerations {
+		for _, clusterToleration := range clusterPod.Spec.Tolerations {
+			if clusterToleration == desiredToleration {
+				goto skip
```
XiShanYongYe-Chang, line 94: First of all, I don't think it's wrong to use the goto statement. If accepted, this will be the first time the Karmada project has used the goto language, which means more goto languages are likely to follow. In view of the disadvantages of the goto statement, I'm not sure if this operation is reasonable. Ask @RainbowMango to help take a look.
/cc @RainbowMango

pkg/resourceinterpreter/default/native/retain.go
XiShanYongYe-Chang: Or we can add additional comments to explain, my first reaction after seeing the code is whether the object is wrong, and I think @chaunceyjiang  also has a first impression misconception.

whitewindmills: Yeah, it's a good idea. There are many fields in _Pod_, but only a few fields that can be modified. I think it is unnecessary waste to copy from `clusterPod` to `desiredPod`.

XiShanYongYe-Chang:
Thanks~ LGTM
Ask @chaunceyjiang to help a look again.
/cc @chaunceyjiang

pkg/resourceinterpreter/default/native/retain.go
```diff
@@ -40,34 +41,65 @@ func retainPodFields(desired, observed *unstructured.Unstructured) (*unstructure
...
+func updatePodTolerations(desiredPod, clusterPod *corev1.Pod) {
+outer:
+	for _, desiredToleration := range desiredPod.Spec.Tolerations {
+		for _, clusterToleration := range clusterPod.Spec.Tolerations {
+			if clusterToleration == desiredToleration {
+				continue outer
+			}
+		}
+		clusterPod.Spec.Tolerations = append(clusterPod.Spec.Tolerations, desiredToleration)
+	}
```
XiShanYongYe-Chang, line 102: I have a question, why not just assign clustedesiredPodrPod.Spec.Tolerations to clusterPod.Spec.Tolerations?

whitewindmills: There are some limitations for updating some fields of a running Pod, in place.
* For `spec.tolerations`, you can only add new entries.
* For `spec.activeDeadlineSeconds`, two types of updates are allowed:
  1) setting the unassigned field to a positive number.
  2) updating the field from a positive number to a smaller, non-negative number.
* For `spec.containers`, `spec.initContainers`, only image can be updated.

XiShanYongYe-Chang: Are you worried that when resources are created on karmada, these fields are inconsistent with those on the cluster? For example, for `spec.tolerations`, some elements are reduced.

pkg/resourceinterpreter/default/native/retain.go
```diff
@@ -40,34 +41,65 @@ func retainPodFields(desired, observed *unstructured.Unstructured) (*unstructure
...
 	}
+}
 
-	return unstructuredObj, nil
+func updatePodTolerations(desiredPod, clusterPod *corev1.Pod) {
+outer:
+	for _, desiredToleration := range desiredPod.Spec.Tolerations {
+		for _, clusterToleration := range clusterPod.Spec.Tolerations {
+			if clusterToleration == desiredToleration {
+				continue outer
```
XiShanYongYe-Chang, line 98: Can we achieve the same goal by adding a variable, like `xxFlag`?

pkg/resourceinterpreter/default/native/retain.go
```diff
@@ -40,34 +41,65 @@ func retainPodFields(desired, observed *unstructured.Unstructured) (*unstructure
...
+		clusterPod.Spec.ActiveDeadlineSeconds = desiredPod.Spec.ActiveDeadlineSeconds
 	}
+}
 
-	return unstructuredObj, nil
+func updatePodTolerations(desiredPod, clusterPod *corev1.Pod) {
+outer:
+	for _, desiredToleration := range desiredPod.Spec.Tolerations {
+		for _, clusterToleration := range clusterPod.Spec.Tolerations {
+			if clusterToleration == desiredToleration {
```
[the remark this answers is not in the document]
whitewindmills: Cool, nice finding.

pkg/resourceinterpreter/default/native/retain.go
whitewindmills: Could you explain to me why we need to do this, or what is the benefit compared to label loop?

XiShanYongYe-Chang: Personally, I haven't used the label loop before, and it feels like the code is hard to read.
Ask @RainbowMango to help take a look.
/cc @RainbowMango

RainbowMango: OK. I will check it as soon as possible.

Poor12:
Great job! I have also encountered this problem and record in #3572. When retain failed, status can not be synced to the karmada control plane.

whitewindmills:
> Great job! I have also encountered this problem and record in #3572. When retain failed, status can not be synced to the karmada control plane.

thanks, I wonder if this PR solved your problem.

chaunceyjiang (commented):
LGTM

RainbowMango:
@Poor12 Does this patch resolve #3572?

PR #3445, opened 2023-04-21
karmada-search should take namespace into account for less memory usage

hanzai:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:
Namespace of resourceSelectors in ResourceRegistry is ignored by karmada-search.  
For example, karmada-search caches pods of all namespaces even when namespace is specified in ResourceRegistry .
```
apiVersion: search.karmada.io/v1alpha1
kind:  ResourceRegistry
metadata:
  name: example
spec:
  targetCluster:
    clusterNames:
    - rancher
  resourceSelectors:
    - apiVersion: v1
      kind: Pod
      namespace: test
```
Taking namespace into account  can reduce memory usage significantly in the proxy feature.

image

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:
```release-note
Karmada-search only cache resources of the namespace specified by ResourceRegistry in the proxy feature.
```

PR #3527, opened 2023-05-11
Support namespace filter in RR for search proxy

ikaven1024:
**What type of PR is this?**
/kind feature

Fixes #3445

```release-note
`karmada-search`: support namespace filter in RR for search proxy.
```

### DIFF
--- a/pkg/search/proxy/controller.go
+++ b/pkg/search/proxy/controller.go
@@ -170,16 +170,22 @@ func (ctl *Controller) reconcile(util.QueueKey) error {
 		return err
 	}
 
-	resourcesByClusters := make(map[string]map[schema.GroupVersionResource]struct{})
+	resourcesByClusters := make(map[string]map[schema.GroupVersionResource]*store.MultiNamespace)
 	for _, registry := range registries {
-		matchedResources := make(map[schema.GroupVersionResource]struct{}, len(registry.Spec.ResourceSelectors))
+		matchedResources := make(map[schema.GroupVersionResource]*store.MultiNamespace, len(registry.Spec.ResourceSelectors))
 		for _, selector := range registry.Spec.ResourceSelectors {
 			gvr, err := restmapper.GetGroupVersionResource(ctl.restMapper, schema.FromAPIVersionAndKind(selector.APIVersion, selector.Kind))
 			if err != nil {
 				klog.Errorf("Failed to get gvr: %v", err)
 				continue
 			}
-			matchedResources[gvr] = struct{}{}
+
+			nsSelector := matchedResources[gvr]
+			if nsSelector == nil {
+				nsSelector = store.NewMultiNamespace()
+				matchedResources[gvr] = nsSelector
+			}
+			nsSelector.Add(selector.Namespace)
 		}
 
 		if len(matchedResources) == 0 {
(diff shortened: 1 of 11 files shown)


XiShanYongYe-Chang:
Thanks a lot, I will take a look ASAP.

XiShanYongYe-Chang (commented):
Thanks a lot

XiShanYongYe-Chang:
/lgtm

Hi @hanzai, I think this pr has achieved the function you want, can you help take a look?

hanzai:
image

In this case, should proxy get pods of default namespace from cache or karmada cluster?

The code indicates that this action hits the cache when pods of default namespace is not cached. @ikaven1024 @XiShanYongYe-Chang 
> https://github.com/ikaven1024/karmada/blob/proxy/pkg/search/proxy/framework/plugins/cache/cache.go#L55

XiShanYongYe-Chang:
Hi @hanzai I'm sorry, I didn't catch your point. Do you mean if it's not in the search cache, it's directly obtained from the member cluster?

hanzai:
> Hi @hanzai I'm sorry, I didn't catch your point. Do you mean if it's not in the search cache, it's directly obtained from the member cluster?

I suppose it should get from the karmada cluster, not member cluster

> Hi @hanzai I'm sorry, I didn't catch your point. Do you mean if it's not in the search cache, it's directly obtained from the member cluster?

image

image

The Command `get pod -n default`  can get different results even if the admin of default namespace makes no operations.

In product environment, different namespaces are used by different teams. Someone adds RR with a namespace should not make affects in other namespace.

XiShanYongYe-Chang:
Hi @hanzai, thanks for your wonderful reply, I understand your concern. This is indeed a question worth considering.

According to the current design, ResourceRegistry is a system configuration. When an administrator performs operations on ResourceRegistry, all users who use karmada search to query data are affected. Therefore, all users need to know the existence of the ResourceRegistry, that is, users need to know which resources are cached on the current control plane.

How to let common users know whether the queried information is cached from the member cluster to the control plane? One method is that the user proactively queries the ResourceRegistry, and the other method is based on the returned result.

Currently, resources obtained from the cache carry labels of cluster information, but resources obtained from karmada do not have labels.

Perhaps we can consider removing `karmada-plugin` so that we can judge based on the returned error information. Of course, it's just an assumption.

Ask @ikaven1024 for some advice.

hanzai:
> `karmada-plugin` is useful in kubectl query non-resource request:

Yeah，I agree with it. Karmada-plugin is necessary in my use case. 
I figure that the `SupportRequest` function can take namespace into account. When the namespace is not cached, pass through to the next plugin, as following code.
> https://github.com/hanzai/karmada/blob/c7453080a056bc79138dba290c1bdd1eef64e208/pkg/search/proxy/framework/plugins/cache/cache.go#L59

XiShanYongYe-Chang:
> IMO, we should still hold this logic atpresent. And do it when we have sufficient cases.

I agree with this point.

Hi @hanzai, if you want to discuss it at the meeting, please feel free to add agenda in the next Karmada Community Meeting: https://docs.google.com/document/d/1y6YLVC-v7cmVAdbjedoyR5WL0-q45DBRXTvz5_I7bkA/edit#heading=h.g61sgp7w0d0c

hanzai:
> > IMO, we should still hold this logic atpresent. And do it when we have sufficient cases.
> 
> I agree with this point.
> 
> Hi @hanzai, if you want to discuss it at the meeting, please feel free to add agenda in the next Karmada Community Meeting: https://docs.google.com/document/d/1y6YLVC-v7cmVAdbjedoyR5WL0-q45DBRXTvz5_I7bkA/edit#heading=h.g61sgp7w0d0c

I'm tied up this month and glad to attend the meeting next month.

PR #3533, opened 2023-05-14
Fix overridePolicy do not work when failed to process one of the matched binding

### LINKED ISSUE #3532: [flake] application failover testing
#### Which jobs are flaking:
Occasional errors in application failover testing:
eg. https://github.com/karmada-io/karmada/actions/runs/4957199002/jobs/8868779613?pr=3504

#### Reason for failure:
OverridePolicy does not work. When binding-controller watches a new op, it will traverse all related Bindings and requeue those objects hit by overrideRS. If one of the Bindings fails to be processed, it will return nil directly which may cause the op event to be lost.
```
        for _, binding := range bindingList.Items {
                // Skip resourceBinding with different namespace of current overridePolicy.
                if len(namespace) != 0 && namespace != binding.Namespace {
                        continue
                }

                // Nil resourceSelectors means matching all resources.
                if len(overrideRS) == 0 {
                        klog.V(2).Infof("Enqueue ResourceBinding(%s/%s) as override policy(%s/%s) changes.", binding.Namespace, binding.Name, a.GetNamespace(), a.GetName())
                        requests = append(requests, reconcile.Request{NamespacedName: types.NamespacedName{Namespace: binding.Namespace, Name: binding.Name}})
                        continue
                }

                workload, err := helper.FetchResourceTemplate(c.DynamicClient, c.InformerManager, c.RESTMapper, binding.Spec.Resource)
                if err != nil {
                        klog.Errorf("Failed to fetch workload for resourceBinding(%s/%s). Error: %v.", binding.Namespace, binding.Name, err)
                        return nil
                }

                for _, rs := range overrideRS {
                        if util.ResourceMatches(workload, rs) {
                                klog.V(2).Infof("Enqueue ResourceBinding(%s/%s) as override policy(%s/%s) changes.", binding.Namespace, binding.Name, a.GetNamespace(), a.GetName())
                                requests = append(requests, reconcile.Request{NamespacedName: types.NamespacedName{Namespace: binding.Namespace, Name: binding.Name}})
                                break
                        }
                }
        }
```

Poor12:
**What type of PR is this?**
/kind failing-test

OverridePolicy does not work. When binding-controller watches a new op, it will traverse all related Bindings and requeue those objects hit by overrideRS. If one of the Bindings fails to be processed, it will return nil directly which may cause the op event to be lost.

Fixes #3532 

```release-note

```

### DIFF
--- a/pkg/controllers/binding/binding_controller.go
+++ b/pkg/controllers/binding/binding_controller.go
@@ -193,8 +193,10 @@ func (c *ResourceBindingController) newOverridePolicyFunc() handler.MapFunc {
 
 			workload, err := helper.FetchResourceTemplate(c.DynamicClient, c.InformerManager, c.RESTMapper, binding.Spec.Resource)
 			if err != nil {
+				// If we cannot fetch resource template from binding, this may be due to the fact that the resource template has been deleted.
+				// Just skip it so that it will not affect other bindings.
 				klog.Errorf("Failed to fetch workload for resourceBinding(%s/%s). Error: %v.", binding.Namespace, binding.Name, err)
-				return nil
+				continue
 			}
 
 			for _, rs := range overrideRS {
--- a/pkg/controllers/binding/cluster_resource_binding_controller.go
+++ b/pkg/controllers/binding/cluster_resource_binding_controller.go
@@ -180,8 +180,10 @@ func (c *ClusterResourceBindingController) newOverridePolicyFunc() handler.MapFu
 
 			workload, err := helper.FetchResourceTemplate(c.DynamicClient, c.InformerManager, c.RESTMapper, binding.Spec.Resource)
 			if err != nil {
+				// If we cannot fetch resource template from binding, this may be due to the fact that the resource template has been deleted.
+				// Just skip it so that it will not affect other bindings.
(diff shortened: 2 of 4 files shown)


pkg/controllers/binding/binding_controller.go
```diff
@@ -194,7 +194,7 @@ func (c *ResourceBindingController) newOverridePolicyFunc() handler.MapFunc {
 			workload, err := helper.FetchResourceTemplate(c.DynamicClient, c.InformerManager, c.RESTMapper, binding.Spec.Resource)
 			if err != nil {
 				klog.Errorf("Failed to fetch workload for resourceBinding(%s/%s). Error: %v.", binding.Namespace, binding.Name, err)
-				return nil
+				continue
```
XiShanYongYe-Chang, line 197: Can here add some comment to explain why we can continue?

chaunceyjiang: Does `cluster_binding_controller.go` have the same problem?

Poor12:
/cc @RainbowMango @mathlsj 
This issue may affect the new added application failover testing.
BTW, we can't let the error of one of the Bindings affect the processing of other Bindings, which is why I didn't distinguish the api.isNotFound error separately.

RainbowMango (approved):
/approve

PR #3541, opened 2023-05-15
Fix if not found workload in overrider policy change, it would be skip.

mathlsj:
**What type of PR is this?**

What happend?

1. delete the workload and garbage-controller have not deal
2. other workload's override policy change
3. the binding_controller and cluster_resource_binding_controller will return nil

it lead to override not take effect

Fixes #

```release-note

```


Poor12:
Thanks @mathlsj. This issue has been tracked by #3533. Would you like to help review it?

XiShanYongYe-Chang:
Good catch!

mathlsj:
Oh, sorry. i don't see it.

PR #3549, opened 2023-05-16
Delete redundant deepcopy

Poor12:
**What type of PR is this?**
/kind cleanup

The resource object which get from controller is a brand new memory. There will be not any risk of mutating the cache.
See https://github.com/karmada-io/karmada/pull/3538#issuecomment-1549124143 and [controller-runtime examples](https://github.com/kubernetes-sigs/controller-runtime/blob/main/examples/builtins/controller.go). 

Fixes #

None

### DIFF
--- a/pkg/controllers/execution/execution_controller.go
+++ b/pkg/controllers/execution/execution_controller.go
@@ -273,8 +273,7 @@ func (c *Controller) updateAppliedCondition(work *workv1alpha1.Work, status meta
 		}
 		updated := &workv1alpha1.Work{}
 		if err = c.Get(context.TODO(), client.ObjectKey{Namespace: work.Namespace, Name: work.Name}, updated); err == nil {
-			// make a copy, so we don't mutate the shared cache
-			work = updated.DeepCopy()
+			work = updated
 		} else {
 			klog.Errorf("Failed to get updated work %s/%s: %v", work.Namespace, work.Name, err)
 		}
--- a/pkg/controllers/federatedresourcequota/federated_resource_quota_status_controller.go
+++ b/pkg/controllers/federatedresourcequota/federated_resource_quota_status_controller.go
@@ -153,8 +153,7 @@ func (c *StatusController) collectQuotaStatus(quota *policyv1alpha1.FederatedRes
 
 		updated := &policyv1alpha1.FederatedResourceQuota{}
 		if err = c.Get(context.TODO(), client.ObjectKey{Namespace: quota.Namespace, Name: quota.Name}, updated); err == nil {
-			// make a copy, so we don't mutate the shared cache
-			quota = updated.DeepCopy()
+			quota = updated
 		} else {
 			klog.Errorf("Failed to get updated  federatedResourceQuota(%s): %v", klog.KObj(quota).String(), err)
 		}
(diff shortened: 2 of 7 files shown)


pkg/controllers/gracefuleviction/crb_graceful_eviction_controller.go
```diff
@@ -68,9 +68,8 @@ func (c *CRBGracefulEvictionController) syncBinding(binding *workv1alpha2.Cluste
 	}
 
 	objPatch := client.MergeFrom(binding)
-	modifiedObj := binding.DeepCopy()
-	modifiedObj.Spec.GracefulEvictionTasks = keptTask
-	err := c.Client.Patch(context.TODO(), modifiedObj, objPatch)
+	binding.Spec.GracefulEvictionTasks = keptTask
```
Poor12, line 71: /cc @XiShanYongYe-Chang for verifying there is extra usage.

XiShanYongYe-Chang: The situation here requires Deepcopy, refer to here:
vendor/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil/controllerutil.go:250

chaunceyjiang:
I see your this pr already contains patch #3533 , but E2E testing still failed.

Poor12:
> I see your this pr already contains patch #3533 , but E2E testing still failed.

My fault, I introduced a new error in eviction_controller.

pkg/controllers/gracefuleviction/crb_graceful_eviction_controller.go
Poor12: Thanks for reminding.

XiShanYongYe-Chang (commented):
Thanks
/lgtm

pkg/controllers/execution/execution_controller.go
```diff
@@ -273,8 +273,7 @@ func (c *Controller) updateAppliedCondition(work *workv1alpha1.Work, status meta
 		}
 		updated := &workv1alpha1.Work{}
 		if err = c.Get(context.TODO(), client.ObjectKey{Namespace: work.Namespace, Name: work.Name}, updated); err == nil {
-			// make a copy, so we don't mutate the shared cache
-			work = updated.DeepCopy()
+			work = updated
```
RainbowMango, line 276: Yes, we don't need copy here. but I think we can optimize the code. I think we don't need `updated` variable here.

RainbowMango (commented):
/assign

pkg/controllers/execution/execution_controller.go
```diff
@@ -271,11 +271,7 @@ func (c *Controller) updateAppliedCondition(work *workv1alpha1.Work, status meta
 		if updateErr == nil {
 			return nil
 		}
-		updated := &workv1alpha1.Work{}
-		if err = c.Get(context.TODO(), client.ObjectKey{Namespace: work.Namespace, Name: work.Name}, updated); err == nil {
-			// make a copy, so we don't mutate the shared cache
-			work = updated.DeepCopy()
-		} else {
+		if err = c.Get(context.TODO(), client.ObjectKey{Namespace: work.Namespace, Name: work.Name}, work); err != nil {
```
RainbowMango, line 274: Here will mutate the memory pointed by the `work` pointer, which might be dangerous as we don't know how the `work` be used after `updateAppliedCondition`.
```suggestion
		work = &workv1alpha1.Work{}
		if err = c.Get(context.TODO(), client.ObjectKey{Namespace: work.Namespace, Name: work.Name}, work); err != nil {
			klog.Errorf("Failed to get updated work %s/%s: %v", work.Namespace, work.Name, err)
		}
```
How about initializing a block of memory specifically for this function to use?

RainbowMango (approved):
/lgtm
/approve

PR #3576, opened 2023-05-24
karmada operator: add conditions to karmada cr

calvin0327:
**What type of PR is this?**
/kind feature

In init and deinit, we need more informations to observe the karmada current status.

here is test result:

```shell
kubectl get karmada -n test
NAME           READY   AGE
karmada-demo   False   27m
```

the cr status message:

image

Fixes #

```release-note
NONE
```

### DIFF
--- a/operator/pkg/apis/operator/v1alpha1/type.go
+++ b/operator/pkg/apis/operator/v1alpha1/type.go
@@ -25,7 +25,7 @@ import (
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 // +kubebuilder:subresource:status
 // +kubebuilder:path=karmadas,scope=Namespaced,categories={karmada-io}
-// +kubebuilder:printcolumn:JSONPath=`.status.controlPlaneReady`,name="Status",type=string
+// +kubebuilder:printcolumn:JSONPath=`.status.conditions[?(@.type=="Ready")].status`,name="Ready",type=string
 // +kubebuilder:printcolumn:JSONPath=`.metadata.creationTimestamp`,name="Age",type=date
 
 // Karmada enables declarative installation of karmada.
@@ -557,10 +557,6 @@ type HostCluster struct {
 type ConditionType string
 
 const (
-	// Unknown represent a condition type the karmada not be reconciled by operator
-	// or unpredictable condition.
-	Unknown ConditionType = "Unknown"
-
 	// Ready represent a condition type the all installation process to karmada have completed.
 	Ready ConditionType = "Ready"
 )
--- a/operator/pkg/controller/karmada/planner.go
+++ b/operator/pkg/controller/karmada/planner.go
@@ -1,6 +1,7 @@
 package karmada
 
 import (
+	"context"
 	"fmt"
 
 	"k8s.io/client-go/rest"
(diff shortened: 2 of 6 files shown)


calvin0327:
@Poor12 @lonelyCZ /cc

RainbowMango (commented):
/assign
I'll take a look.
Please solve the conflicts.

operator/pkg/apis/operator/v1alpha1/type.go
```diff
@@ -557,12 +557,11 @@ type HostCluster struct {
 type ConditionType string
 
 const (
-	// Unknown represent a condition type the karmada not be reconciled by operator
-	// or unpredictable condition.
-	Unknown ConditionType = "Unknown"
-
 	// Ready represent a condition type the all installtion process to karmada have compaleted.
```
Poor12, line 560:
```suggestion
	// Ready represent a condition type the all installation process to karmada have completed.
```

operator/pkg/apis/operator/v1alpha1/type.go
```diff
@@ -557,12 +557,11 @@ type HostCluster struct {
...
 	Ready ConditionType = "Ready"
+
+	// Shutdown represent a condition type the uninstall process to karmada have complaleted.
```
Poor12, line 563:
```suggestion
	// Shutdown represent a condition type the uninstall process to karmada have completed.
```

operator/pkg/apis/operator/v1alpha1/type.go
```diff
@@ -557,12 +557,11 @@ type HostCluster struct {
...
+	Shutdown ConditionType = "Shutdown"
```
Poor12, line 564: Seems that `Ready` means in the progressing of installing, `Shutdown` means in the progressing of uninstalling.

I don't think these words are clear enough.

calvin0327: Yes, I've been thinking about it for a long time.  I have some confusion here：

*   Do both of installation and uninstallation use the same condition type `Ready`？
*   If use a different ccondition type, these words are not clear enough.

Do you have any good suggestions?

operator/pkg/apis/operator/v1alpha1/type.go
```diff
@@ -557,12 +557,11 @@ type HostCluster struct {
 type ConditionType string
 
 const (
-	// Unknown represent a condition type the karmada not be reconciled by operator
-	// or unpredictable condition.
-	Unknown ConditionType = "Unknown"
-
-	// Ready represent a condition type the all installation process to karmada have completed.
+	// Ready represent a condition type the all installation process to karmada have compaleted.
```
RainbowMango, line 560: typo `compaleted`.

operator/pkg/apis/operator/v1alpha1/type.go
RainbowMango: When the condition would be `Shutdown`? Does deleting the `Karmada` means shut down?

calvin0327: When removing Karmada CR and triggers the process of uninstalling karmada, which takes about 5 seconds. In this short time, I want to use condition tyepe `shutdown` to represent the current state of karmada.  or is there a better condition type define?

calvin0327: > Controllers should apply their conditions to a resource the first time they visit the resource, even if the status is Unknown. This allows other components in the system to know that the condition exists and that the controller is making progress on reconciling that resource.

Sorry for late reply, ok, I got that means and I'm agress with you.

I think we should only need one condition type (`Ready` or other) to define the whole karmada lifecycle. for different operation  `init`, `upgrade` or `deint` denoted by different reason. for example:

Deletion operation condition:

```yaml
type: Ready
status: False
reason: Deleting
message: "the karmada is being deleted"
```

RainbowMango: Then, we don't need the `ShutDown` condition, right? Please remove it.

calvin0327: yes, I think so.

calvin0327:
@RainbowMango @Poor12 Please review again? thanks.

RainbowMango (approved):
/approve
For the API part.

operator/pkg/controller/karmada/planner.go
```diff
@@ -83,10 +84,44 @@ func recognizeActionFor(karmada *operatorv1alpha1.Karmada) Action {
 func (p *Planner) Execute() error {
 	klog.InfoS("Start execute the workflow", "workflow", p.action, "karmada", klog.KObj(p.karmada))
 
+	if err := p.preRunJob(); err != nil {
+		return err
+	}
 	if err := p.job.Run(); err != nil {
+		if err := p.runJobErr(err); err != nil {
```
Poor12, line 91: return p.runJobErr(err)

operator/pkg/controller/karmada/planner.go
```diff
@@ -83,10 +84,44 @@ func recognizeActionFor(karmada *operatorv1alpha1.Karmada) Action {
...
+	return p.Client.Status().Update(context.TODO(), p.karmada)
+}
+
+func (p *Planner) runJobErr(err error) error {
+	operatorv1alpha1.KarmadaFailed(p.karmada, operatorv1alpha1.Ready, err.Error())
+	return p.Client.Status().Update(context.TODO(), p.karmada)
+}
+
+func (p *Planner) afterRunJob() error {
+	operatorv1alpha1.KarmadaCompleted(p.karmada, operatorv1alpha1.Ready, "karmada init job was completed")
```
Poor12, line 121:
```suggestion
	operatorv1alpha1.KarmadaCompleted(p.karmada, operatorv1alpha1.Ready, "karmada init job is completed")
```

operator/pkg/controller/karmada/planner.go
```diff
@@ -83,10 +84,44 @@ func recognizeActionFor(karmada *operatorv1alpha1.Karmada) Action {
...
+
+	// the cr will also be deleted with karmada is be deleted, so we need not to
```
Poor12, line 123: So do you need to update the karmada status?

calvin0327:
@calvin0327 thanks for your review.

operator/pkg/controller/karmada/planner.go
calvin0327: I think we need not to update karmada status, Because karmada cr will be deleted immediately.

Poor12:
```
return p.Client.Status().Update(context.TODO(), p.karmada)
```
So the above code is unnecessary？

calvin0327: It's necessary, this is init workflow. 
It's unnecessary in `deinit` workflow, I update the comment.

calvin0327: updated.

PR #3585, opened 2023-05-25
samples from documentation

bivas:
**What type of PR is this?**

/kind documentation

Adding the samples from documentation to repository

```release-note
NONE
```


samples/service_discovery/import.yaml
```diff
@@ -0,0 +1,25 @@
...
+  namespace: discovery
+spec:
+  resourceSelectors:
+    - apiVersion: multicluster.x-k8s.io/v1alpha1
+      kind: ServiceImport
+      name: serve
+  placement:
+    clusterAffinity:
+      clusterNames:
+        - member2
```
chaunceyjiang, line 25: Please add a new line.

XiShanYongYe-Chang:
Hi, @bivas Welcome! :)

RainbowMango (commented):
> Adding the samples from documentation to repository

Are these yamls come from https://github.com/karmada-io/website repo?
Can you guys remind me why we need to do that?

bivas:
@RainbowMango 
> Are these yamls come from https://github.com/karmada-io/website repo? Can you guys remind me why we need to do that?

Yes, these are the YAML from the website, with some minor alterations.
I added them to avoid copy pasting from the docs. I don't think it's really required, but more of a nice to have.

RainbowMango:
> I added them to avoid copy pasting from the docs.

Sorry, I still can't get the benefit from it. Can you remind me?

My concern is we have a lot of yamls in the website repo, we don't need to maintain these yamls in two places I think.

bivas:
@RainbowMango I agree. There shouldn't be multiple places for these YAMLs.
It was just an easier way for me to try out several scenarios when going over the docs and though others might benefit from it as well.

Maybe the docs should only point to these YAMLs instead of embedding them? 
Again, I don't think it's that important, it's just to ease the evaluation process.

RainbowMango:
> Maybe the docs should only point to these YAMLs instead of embedding them?

I even feel that there is no need to maintain the links. Because the yamls in website repo might be changed over time, an extra effort is needed to maintain the link valid.  
The example yamls in the website repo should support use in a `copy-and-paste` model. We can put our effort to this point.

RainbowMango:
I'm going to close this for now, please feel free to reopen it if you want further discussions.
/close

PR #3616, opened 2023-06-01
external cluster health probe

### LINKED ISSUE #3612: imporve cluster health check logic

Configurable cluster health check plugin, to check member cluster could run workloads or not.

**Why is this needed**:

Current impl only check `/healthz` of apiserver, which nearly always returns healthy:

https://github.com/kubernetes/apiserver/blob/6e796557fa8a46ec48109166f6db572bbda87267/pkg/server/config.go#L363-L364

https://github.com/karmada-io/karmada/blob/dff7a5b7881ea9f101803ffab04f6f7d0e93bd29/pkg/controllers/status/cluster_status_controller.go#L383-L388

But, apiserver is healthy dosen't represent the cluster is healthy, it's may be all nodes not ready, etc. , pods will never run.

So we'd better add the ability to configure how to check a member cluster, since the cluster may in different distribution (Azure, AWS, Huawei, etc.), at least we need to check node status of cluster, to avoid all nodes not ready (due to cni failed, etc.)

tedli:
**What type of PR is this?**

/kind feature

Current cluster health checking logic only check `/healthz` of apiserver of member cluster, which nearly always returns OK. However, at the mean time all nodes in the member cluster may be not ready, and karmada controller will continue dispatch works to it.

Fixes #3612

Not ready to merge, just for review this idea. If it's a common needs, I'm glad to finish this.

NONE
```release-note

```

By defining a annotation value, to tell karmada how to check health of a member cluster, the owner of the member cluster should provide a probe and impl checking logic, because clusters varies.

```yaml
apiVersion: cluster.karmada.io/v1alpha1
kind: Cluster
metadata:
  name: member-dev
  annotations:
     cluster.karmada.io/health-checker: |-
        {"service":{"namespace":"some-system","name":"member-dev-health","port":80,"path":"/healthz"}}
  finalizers:
  - karmada.io/cluster-controller
spec:
  apiEndpoint: https://10.224.144.185:31212
  id: bdcb8c60-b85c-407a-aeb3-6d7623b6b6c5
  impersonatorSecretRef:
    name: member-dev-impersonator
    namespace: karmada-cluster
```

P.S.
Dose karmada willing to create a `awesome-karmada`, `contrib` liked repo under top karmada project? So, someone may like to add their unofficial out of tree controller to it. These 3rd controlers, compoents, may solve some cases but yet not a commonly need feature, or just not stable enough to merge intree.


tedli:
I give a health check probe impl, feel free to give it a try, and welcom feedback.

https://github.com/tedli/generic-cluster-healthz

chaunceyjiang:
I think it's a good idea, but is setting the `health-checker` configuration in annotations a good choice?

RainbowMango:
> I give a health check probe impl, feel free to give it a try, and welcom feedback.

Just want to let you know, it's in my queue. I'll take a look next week.

> I think it's a good idea, but is setting the health-checker configuration in annotations a good choice?

Good point! I tend to go with the approach of introducing a field in `Cluster` API.

Poor12:
This is quite similar to `livenessProbe` in kubernetes. I think we can learn from their API design to add a new field in the cluster.
From the description of PR, it may like:
```
apiVersion: cluster.karmada.io/v1alpha1
kind: Cluster
metadata:
  name: member-dev
  finalizers:
  - karmada.io/cluster-controller
spec:
  apiEndpoint: https://10.224.144.185:31212
  id: bdcb8c60-b85c-407a-aeb3-6d7623b6b6c5
  impersonatorSecretRef:
    name: member-dev-impersonator
    namespace: karmada-cluster
  livenessProbe:
    httpGet:
       path: /healthz
       port: 80
       host: member-dev-health.some-system.svc.cluster.local
```

chaunceyjiang:
> This is quite similar to livenessProbe in kubernetes. 

Amazing!!!!!!!!

jwcesign:
Hi, @tedli 
Just curious,

1. Where did you deploy the component corresponding to this service?
2. Different clusters will have different health checking services?
3. Is it possible that all clusters use one service to check the status, but with different query parameters?

tedli:
Hi @jwcesign ,

In my case:
1. I did deploy the probe component in the host cluster of karmada control plane.
2. Each cluster has its own probe component.
3. Of course, it's possible.

Reason:
2. Different cluster use different kubeconfig.
3. It's need to config multi kubeconfig of different clusters to this probe, so that it could use corresponding kubeconfig of the cluster specified through query, to check the cluster.

What I got **WRONG**:

The kubeconfig of member cluster that provided to karmada control plane during `join` must be something like `admin` or `cluster-admin`,that nearly has all role.

So karmada controller could do some basic checks, like checking n nodes ready, n non host networked pods ready (n could be configed) in tree, no need to add an extra probe.

Only if the basic checks like I mentioned above was not enough, then an external probe can be used.

PR #3631, opened 2023-06-04
Implement proxy headers of cluster API

Poor12:
**What type of PR is this?**
/kind feature

Since we have add proxy headers in cluster API, I'd like to implement proxy headers of cluster API.

Fixes #

```
`karmada-controller-manager`: implement proxy header of cluster API
```

### DIFF
--- a/pkg/util/membercluster_client.go
+++ b/pkg/util/membercluster_client.go
@@ -201,6 +201,10 @@ func BuildClusterConfig(clusterName string,
 			return nil, err
 		}
 		clusterConfig.Proxy = http.ProxyURL(proxy)
+
+		if len(cluster.Spec.ProxyHeader) != 0 {
+			clusterConfig.Wrap(NewProxyHeaderRoundTripperWrapperConstructor(clusterConfig.WrapTransport, cluster.Spec.ProxyHeader))
+		}
 	}
 
 	return clusterConfig, nil
--- a/pkg/util/round_trippers.go
+++ b/pkg/util/round_trippers.go
@@ -1,0 +1,51 @@
+package util
+
+import (
+	"net/http"
+	"net/textproto"
+	"strings"
+
+	"k8s.io/client-go/transport"
+)
+
+type proxyHeaderRoundTripper struct {
+	proxyHeaders http.Header
+	roundTripper http.RoundTripper
+}
+
+var _ http.RoundTripper = &proxyHeaderRoundTripper{}
+
+// NewProxyHeaderRoundTripperWrapperConstructor returns a RoundTripper wrapper that's usable within restConfig.WrapTransport.
+func NewProxyHeaderRoundTripperWrapperConstructor(wt transport.WrapperFunc, headers map[string]string) transport.WrapperFunc {
+	return func(rt http.RoundTripper) http.RoundTripper {
+		if wt != nil {
+			rt = wt(rt)
+		}
+		return &proxyHeaderRoundTripper{
+			proxyHeaders: parseProxyHeaders(headers),
+			roundTripper: rt,
+		}
+	}
+}
+
+// RoundTrip implements the http.RoundTripper interface.
+func (r *proxyHeaderRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	if tr, ok := r.roundTripper.(*http.Transport); ok {
+		tr = tr.Clone()
+		tr.ProxyConnectHeader = r.proxyHeaders
+		return tr.RoundTrip(req)
+	}
+	return r.roundTripper.RoundTrip(req)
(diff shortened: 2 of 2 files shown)


jwcesign (commented):
/lgtm

RainbowMango (commented):
Assigning this to controller owners.
/assign @XiShanYongYe-Chang @chaunceyjiang

chaunceyjiang (commented):
/assign

pkg/util/membercluster_client.go
```diff
@@ -203,6 +203,10 @@ func BuildClusterConfig(clusterName string,
 		clusterConfig.Proxy = http.ProxyURL(proxy)
 	}
 
+	if len(cluster.Spec.ProxyHeader) != 0 {
+		clusterConfig.WrapTransport = NewProxyHeaderRoundTripperWrapperConstructor(clusterConfig.WrapTransport, cluster.Spec.ProxyHeader)
+	}
```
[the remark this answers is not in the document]
XiShanYongYe-Chang: +1

XiShanYongYe-Chang (commented):
LGTM

pkg/util/round_trippers.go
```diff
@@ -0,0 +1,50 @@
...
+			proxyHeaders: parseProxyHeaders(headers),
+			roundTripper: rt,
+		}
+	}
+}
+
+// RoundTrip implements the http.RoundTripper interface.
+func (r *proxyHeaderRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	if tr, ok := r.roundTripper.(*http.Transport); ok {
+		tr.ProxyConnectHeader = r.proxyHeaders
```
chaunceyjiang, line 34: I think a `clone Transport` should be used here because the default value of Transport is `http.DefaultTransport`.

chaunceyjiang: like :
``` go
tr = tr.Clone()
```

chaunceyjiang: > I think a clone Transport should be used here because the default value of Transport is http.DefaultTransport.



I'm still reviewing.

Poor12: I still don't understand why we need clone, could you explain more clearly?

chaunceyjiang: https://github.com/golang/go/blob/master/src/net/http/client.go#L129-L142

chaunceyjiang: >	// RoundTrip should not modify the request, except for
	// consuming and closing the Request's Body. RoundTrip may
	// read fields of the request in a separate goroutine. Callers
	// should not mutate or reuse the request until the Response's
	// Body has been closed.

Poor12: I'm afraid not. `tr` is a temporary variable and will not be used outside of the if judge.

chaunceyjiang: > I'm afraid not. tr is a temporary variable and will not be used outside of the if judge.



It may be that I didn't describe it clearly. What I'm concerned about is the following situation.


image

XiShanYongYe-Chang: I think I understand what @chaunceyjiang  means. You are afraid that the value of r.roundTripper will be modified.

Poor12: I got it. I agree with you.

chaunceyjiang (commented):
Thanks~

/lgtm

pkg/util/membercluster_client.go
```diff
@@ -201,6 +201,10 @@ func BuildClusterConfig(clusterName string,
 			return nil, err
 		}
 		clusterConfig.Proxy = http.ProxyURL(proxy)
+
+		if len(cluster.Spec.ProxyHeader) != 0 {
+			clusterConfig.WrapTransport = NewProxyHeaderRoundTripperWrapperConstructor(clusterConfig.WrapTransport, cluster.Spec.ProxyHeader)
```
RainbowMango, line 206: Just a question. there is [a comment](https://github.com/kubernetes/kubernetes/blame/16534deedf1e3f7301b20041fafe15ff7f178904/staging/src/k8s.io/client-go/rest/config.go#L110-L111) on the `WrapTransport`:
> A future release will change this field to an array. Use config.Wrap() instead of setting this value directly.

What do you think of it?  
Given the comment was made 5 years ago, I don't know if this change is still on the plan list of Kubernetes. So, use `config.Wrap()` or not, both are OK to me.

Poor12: I think the latter is better.

RainbowMango (approved):
/lgtm
/approve

PR #3636, opened 2023-06-05
add proposal for MultiClusterService 

XiShanYongYe-Chang:
**What type of PR is this?**

/kind design

Add MultiClusterService API proposal.

Fixes #

```release-note
NONE
```


RainbowMango (commented):
Generally looks good to me for this approach. 
But we might need more details design. Including:
- What's the behavior when updating the MultiClusterService?
- What's the behavior when deleting the MultiClusterService?
- What's the behavior when the referencing service does not exist?
- What's the behavior when some of the `import cluster` doesn't exist?
- What's the behavior when some of the `import cluster` disappears(like unregistered from Karmada)?

RainbowMango (commented):
Generally looks good to me for this approach. 
But we might need more details design. Including:
- What's the behavior when updating the MultiClusterService?
- What's the behavior when deleting the MultiClusterService?
- What's the behavior when the referencing service does not exist?
- What's the behavior when some of the `import cluster` doesn't exist?
- What's the behavior when some of the `import cluster` disappears(like unregistered from Karmada)?

docs/proposals/networking/multiclusterservice.md
```diff
@@ -0,0 +1,184 @@
...
+	// ImportClusters specifies the destination cluster set to which services need to be imported.
+	// Only valid only for ClusterSetIP ServiceType.
+	// If not set, all clusters are selected.
+	// +optional
+	ImportClusters *ImportClusters `json:"importClusters"`
+}
+
+type ServiceBackend struct {
+	// Port specifies the exposed service port.
+	Port int32 `json:"port"`
```
chaunceyjiang, line 100: What is the difference between `Port` here and the `Ports` in `ServiceImport`?

https://github.com/kubernetes/enhancements/blob/master/keps/sig-multicluster/1645-multi-cluster-services-api/README.md#importing-services

docs/proposals/networking/multiclusterservice.md
```diff
@@ -0,0 +1,184 @@
...
+}
```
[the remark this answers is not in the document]
XiShanYongYe-Chang: Maybe `Protocol` and `Name`.

docs/proposals/networking/multiclusterservice.md
```diff
@@ -0,0 +1,184 @@
...
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// Spec is the desired state of the MultiClusterService.
+	// +optional
+	Spec ServiceSpec `json:"spec,omitempty"`
+
+	// Status is the current state of the MultiClusterService.
+	// +optional
+	Status corev1.ServiceStatus `json:"status,omitempty"`
```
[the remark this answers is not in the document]
XiShanYongYe-Chang: The current idea is to display LB information for users to read. Provider will only set this value, not use it.

docs/proposals/networking/multiclusterservice.md
XiShanYongYe-Chang: Port in ServiceBackend are used to expose services.
Ports in ServiceImport will be filled based on the ports info of the target Service when ServiceImport is automatically created.

XiShanYongYe-Chang: I update the port to ports.

docs/proposals/networking/multiclusterservice.md
```diff
@@ -0,0 +1,249 @@
+---
+title: Propose the MultiClusterService API to expose and discover multi-cluster services
+authors:
+- "@XiShanYongYe-Chang"
+- "@GitHubxsy"
+reviewers:
+- TBD
+approvers:
+- TBD
```
RainbowMango, line 9: Please update the reviewers and approvers. Then we can move forward.

XiShanYongYe-Chang: Done. Thanks~

docs/proposals/networking/multiclusterservice.md
```diff
@@ -0,0 +1,252 @@
...
+// service and are associated with the same MultiClusterService.
+// MultiClusterService can control the exposure of services to outside multiple
+// clusters, and also enable service discovery between clusters.
+type MultiClusterService struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// Spec is the desired state of the MultiClusterService.
+	// +optional
+	Spec MultiClusterServiceSpec `json:"spec,omitempty"`
```
chaunceyjiang, line 80: Is the `Spec` field also optional?

docs/proposals/networking/multiclusterservice.md
```diff
@@ -0,0 +1,252 @@
...
+2. Create the multiClusterService resource.
+3. The controller list/watch to the multiClusterService resource, generates ServiceExport and propagate it to the source clusters, and generates ServiceImport and propagate it to the destination clusters.
+4. Services in the destination clusters access multi-cluster services.
+
+```yaml
+apiVersion: networking.karmada.io/v1alpha1
+kind: MultiClusterService
+metadata:
+  name: foo
+spec:
```
chaunceyjiang, line 194: This sample needs to be updated.

docs/proposals/networking/multiclusterservice.md
```diff
@@ -0,0 +1,252 @@
...
+- `multicluster-cloud-provider` repo:
+  - Add shared interface for MultiClusterService and add/update related controller.
+
+### MultiClusterService resource validation
+
+- `.name`: It need to be equal with the service name.
+- `.spec.backend.name`: It must be the same as that defined in the service.
+- `.spec.backend.port`: It should be a reasonable port value
+- `.spec.exposureTypes`: It can not be empty.
+- `.spec.importClusters.clusterNames`: The cluster name must be a qualified string.
```
chaunceyjiang, line 245: This line need to be updated.

docs/proposals/networking/multiclusterservice.md
XiShanYongYe-Chang: Thanks for reminding, I just remove it.

docs/proposals/networking/multiclusterservice.md
XiShanYongYe-Chang: Updated.

docs/proposals/networking/multiclusterservice.md
```diff
@@ -0,0 +1,251 @@
...
+```yaml
+apiVersion: networking.karmada.io/v1alpha1
+kind: MultiClusterService
+metadata:
+  name: foo
+spec:
+  types:
+  - CrossCluster
+  range:
+    ClusterNames:
```
chaunceyjiang, line 197:
```suggestion
    clusterNames:
```

docs/proposals/networking/multiclusterservice.md
```diff
@@ -0,0 +1,251 @@
...
+// Services with the same name in different clusters are regarded as the same
+// service and are associated with the same MultiClusterService.
+// MultiClusterService can control the exposure of services to outside multiple
+// clusters, and also enable service discovery between clusters.
+type MultiClusterService struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// Spec is the desired state of the MultiClusterService.
+	Spec MultiClusterServiceSpec `json:"spec,omitempty"`
```
chaunceyjiang, line 79:
```suggestion
        // Spec is the desired state of the MultiClusterService.
        // +required
	Spec MultiClusterServiceSpec `json:"spec"`
```

docs/proposals/networking/multiclusterservice.md
```diff
@@ -0,0 +1,251 @@
...
+   - LoadBalance
+```
+
+![image](statics/service-exposure.png)
+
+### Cross-clusters service discovery
+
+1. Create Deployments and Services and propagate them to member clusters.
+2. Create the multiClusterService resource.
+3. The controller list/watch to the multiClusterService resource, generates ServiceExport and propagate it to the source clusters, and generates ServiceImport and propagate it to the destination clusters.
```
chaunceyjiang, line 185: > , generates ServiceExport and propagate it to the source clusters, and generates ServiceImport and propagate it to the destination clusters.

How does this new controller work in conjunction with existing` MCS controllers`?

docs/proposals/networking/multiclusterservice.md
XiShanYongYe-Chang: Thanks~

XiShanYongYe-Chang:
Hi @SAMZONG, thanks for your questions.

> If we don't create services in member clusters, how can we do service governance? 

Sorry, I didn't quite understand this question. Is there a specific scenario for this problem?

> In Karmada, all services are managed by the control plane. What is the backup plan if the control plane goes down? example: how we do service export or input

Currently, we are handling this logic in the controller of the Karmada control plane. The solution for MCS features indeed did not consider how multi-cluster service discovery would continue to work if Karmada goes down. Perhaps we can open a new issue to discuss this further.

> In this solution, if cross-cluster Pods cannot communicate properly, how do we discover and update the correct Service-Pod mapping? (In standard kubernetes, k8s will do Node Health Checks)

Good question. Currently, the MCS feature does not detect the connectivity of remote pod IDs on the target cluster. This requires users to provide assurance. However, if we can prompt users or directly remove disconnected pod IPs, it may have a better effect. We can continue discussing this in more detail and perhaps open a new issue to document and track it.

> Maybe we can discuss more depp?, with Chinese. :)

Sure, glad to make a deep discussion. :)

SAMZONG:
@XiShanYongYe-Chang @RainbowMango Thanks for your make Karmada more Power!

> Federation-service is also like `MCS`

* I believe it is necessary to consider and solve the local priority issue when supporting federation-service, which should become a default property.
* Using federation-service does not mean that we no longer need native services. We can simply provide federation-service based on the existing Service, making it easier to map to Services across multiple clusters.
* Karmada does not need to pay attention to changes in Pod IP within each cluster, as this would increase the burden of design and control plane.
* Federation-service can display Services across clusters, similar to routing handling, for example, forwarding `svc.ns.cluster.local` to `svc.ns.cluster.karmada.local`.
  * Similar to Ingress, but not Ingress; it is necessary to solve cross-cluster Ingress, as well as the ability for Ingress to route to Services in other clusters.
* Each cluster should have its own federation-service `svc.ns.cluster.karmada.local`, and distribute it to multiple clusters through pre-planned policies (PP), controlling forwarding within local cluster services or other cluster services.
* Essentially, the function of federation-service is to resolve requests to the corresponding cluster service, ensuring that when the native Service-Pod IP changes, the federation-service is unaffected.
* Load balancing (LB) is not a necessary option because it leads to too many traffic endpoints, expanding the scope of failures.
* If you only need to solve the network interconnection between multiple clusters, and do not need more advanced functions of Istio, providing a lightweight federation-service might be a better design choice.

This is my idea, hope your reply.

XiShanYongYe-Chang:
Thank you for your detailed and quick response. 

The points you mentioned take into consideration many aspects of cross-cluster service discovery and routing, and I feel that each point can be further discussed in depth. 

The current PR proposes MultiClusterService to expose multi-cluster services and provide users with a unified configuration for cross-cluster service discovery. I think there are no conflicts with the discussion points you raised, so we can proceed with it first.

And next, we can have a deep discussion about cross-cluster service discovery and routing issues. How about scheduling a meeting to discuss these ideas further?

Furthermore, I would like to invite you to take a look at this [proposal](#3694), which may potentially address one of the issues.

docs/proposals/networking/multiclusterservice.md
XiShanYongYe-Chang: Add more info, can you help review it again?

RainbowMango (approved):
/lgtm
/approve

PR #3648, opened 2023-06-08
Karmada metrics adapter deployment script

### LINKED ISSUE #3625: feat: provide the script to deploy karmada-metrics-adapter
* Provide a script to deploy `karmada-metrics-adapter`, similar to the one for `karmada-estimator`. You can find an example here: https://github.com/karmada-io/karmada/blob/master/hack/deploy-scheduler-estimator.sh
* Make sure to install `karmada-metrics-adapter` when running `hack/local-up-karmada.sh`.

**Why is this needed**:
Facilitate the deployment of `karmada-metrics-adapter` for users to utilize FederatedHPA and scale workloads across multiple clusters.

sbdtu5498:
**What type of PR is this?**
/kind feature

**What this PR does**:
Provides deployment scripts for karmada-metrics-adapter.

Fixes #3625 

```release-note
NONE
```

### DIFF
--- a/hack/deploy-metrics-adapter.sh
+++ b/hack/deploy-metrics-adapter.sh
@@ -1,0 +1,67 @@
+#!/bin/bash
+
+set -o errexit
+set -o nounset
+
+REPO_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
+source "${REPO_ROOT}"/hack/util.sh
+function usage() {
+  echo "This script will deploy karmada-metrics-adapter on host cluster"
+  echo "Usage: hack/deploy-metrics-adapter.sh  <HOST_CLUSTER_KUBECONFIG> <HOST_CLUSTER_NAME>>"
+  echo "Example: hack/deploy-metrics-adapter.sh ~/.kube/config karmada-host"
+}
+
+if [[ $# -ne 2 ]]; then
+  usage
+  exit 1
+fi
+
+# check kube config file existence
+if [[ ! -f "${1}" ]]; then
+  echo -e "ERROR: failed to get kubernetes config file: '${1}', not existed.\n"
+  usage
+  exit 1
+fi
+HOST_CLUSTER_KUBECONFIG=$1
+
+# check context existence
+if ! kubectl config get-contexts "${2}" --kubeconfig="${HOST_CLUSTER_KUBECONFIG}" > /dev/null 2>&1;
+then
+  echo -e "ERROR: failed to get context: '${2}' not in ${HOST_CLUSTER_KUBECONFIG}. \n"
+  usage
+  exit 1
+fi
+HOST_CLUSTER_NAME=$2
+
+# install metrics adapter to host cluster
+if [ -n "${KUBECONFIG+x}" ];then
+  CURR_KUBECONFIG=$KUBECONFIG # backup current kubeconfig
+fi
+
+export KUBECONFIG=$HOST_CLUSTER_KUBECONFIG
+echo "using kubeconfig: "$KUBECONFIG
+
+# deploy karmada-metrics-adapter
+kubectl --context="${HOST_CLUSTER_NAME}" apply -f "${REPO_ROOT}/artifacts/deploy/karmada-metrics-adapter.yaml"
+
+# make sure that karmada-metrics-adapter is ready
+util::wait_pod_ready "${HOST_CLUSTER_NAME}" "${KARMADA_METRICS_ADAPTER_LABEL}" "${KARMADA_SYSTEM_NAMESPACE}"
+
+# deploy karmada-metrics-adapter-apiservice
+kubectl --context="karmada-apiserver" apply -f "${REPO_ROOT}/artifacts/deploy/karmada-metrics-adapter-apiservice.yaml"
+
+# make sure that karmada-metrics-adapter-apiservice is ready
+util::wait_apiservice_ready "karmada-apiserver" "${KARMADA_METRICS_ADAPTER_LABEL}"
+
+# recover the kubeconfig before installing metrics adapter if necessary
+if [ -n "${CURR_KUBECONFIG+x}" ];then
+  export KUBECONFIG="${CURR_KUBECONFIG}"
+else
+  unset KUBECONFIG
+fi
+
+function print_success() {
+  echo "Karmada metrics adapter is deployed successfully."
+}
+
+print_success


hack/deploy-metrics-adapter.sh
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
+
+set -o errexit
+set -o nounset
+
+REPO_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
+source "${REPO_ROOT}"/hack/util.sh
+function usage() {
+  echo "This script will deploy karmada-metrics-adapter on host cluster"
+  echo "Usage: hack/deploy-metrics-adapter.sh  <HOST_CLUSTER_KUBECONFIG> <HOST_CLUSTER_NAME>>"
```
jwcesign, line 10: I think we need to specify the karmada-apiserver kubeconfig and context name

hack/deploy-metrics-adapter.sh
```diff
@@ -0,0 +1,67 @@
...
+echo "using kubeconfig: "$KUBECONFIG
+
+# deploy karmada-metrics-adapter
+kubectl --context="${HOST_CLUSTER_NAME}" apply -f "${REPO_ROOT}/artifacts/deploy/karmada-metrics-adapter.yaml"
+
+# make sure that karmada-metrics-adapter is ready
+util::wait_pod_ready "${HOST_CLUSTER_NAME}" "${KARMADA_METRICS_ADAPTER_LABEL}" "${KARMADA_SYSTEM_NAMESPACE}"
+
+# deploy karmada-metrics-adapter-apiservice
+kubectl --context="karmada-apiservice" apply -f "${REPO_ROOT}/artifacts/deploy/karmada-metrics-adaptor-apiservice.yaml"
```
jwcesign, line 51: We are uncertain whether the context name should be `karmada-apiserver`. This parameter should be specified as an input flag.

jwcesign (commented):
Thanks for your contribution, @sbdtu5498 . There are a few issues to address, but other than that, it looks good to me (LGTM).

jwcesign:
By the way, please run `chmod 755 hack/xxx` to grant the necessary permissions for the script.

hack/local-up-karmada.sh
```diff
@@ -126,13 +126,17 @@ kind load docker-image "${REGISTRY}/karmada-scheduler:${VERSION}" --name="${HOST
...
+kind load docker-image "${REGISTRY}/karmada-metrics-adapter:${VERSION}" --name="${HOST_CLUSTER_NAME}"
 kind load docker-image "${REGISTRY}/karmada-aggregated-apiserver:${VERSION}" --name="${HOST_CLUSTER_NAME}"
 kind load docker-image "${REGISTRY}/karmada-search:${VERSION}" --name="${HOST_CLUSTER_NAME}"
 
 #step5. install karmada control plane components
 "${REPO_ROOT}"/hack/deploy-karmada.sh "${MAIN_KUBECONFIG}" "${HOST_CLUSTER_NAME}"
 
-#step6. wait until the member cluster ready and join member clusters
+#step6. deploy karmada-metrics-adapter
+"${REPO_ROOT}"/hack/deploy-metrics-adapter.sh "${MAIN_KUBECONFIG}" "${HOST_CLUSTER_NAME}"
```
jwcesign, line 137: Do we need to install metrics-adapter by default when running `hack/local-up-karmada.sh`? cc @XiShanYongYe-Chang for checking

XiShanYongYe-Chang: Can we just provide the script and let the user decide whether to install it?

jwcesign: +1

sbdtu5498: Should we remove it then?

jwcesign: I prefer to delete it

hack/deploy-metrics-adapter.sh
sbdtu5498: The karmada-apiserver context is causing too much confusion. I think the kubeconfig would be the same as the host-cluster-kubeconfig and the context karmada-apiserver is being used as sort of control parameter for permissions and security purpose.

hack/deploy-metrics-adapter.sh
sbdtu5498: The logic behind taking karmada-apiserver as the context was that Karmada is using two different contexts i.e. karmada-apiserver and karmada-host that seem to be referring to the same entity. All the Kubectl request needs to be passed to the cluster where the Karmada control plane is hosted and aggregation should be with the API server of that cluster. In this case, to me, it seems both the karmada-apiserver and karmada-host are referring to the same entity. I took context as karmada-apiserver as it looked more of a convention to me as it was being used by both the API service for aggregation purposes. I wasn't sure either whether it is being used for security purposes or something else, so went with the convention.

sbdtu5498: There seems to be a small typo, "karmada-apiservice" -> "karmada-apiserver". I will fix it.

hack/local-up-karmada.sh
sbdtu5498: Sure!

hack/deploy-metrics-adapter.sh
jwcesign: Karmada is installed is a specific cluster(host cluster), it contains `kube-apiserver` component, named `karmada-apiserver`. So karmada-host is used to connect to the host cluster's kube-apiserver, karmada-apiserver is used to connect to `karmada-apiserver`. They are different entity.

jwcesign:
Hello, @sbdtu5498! You can pass the DCO checking by running `git commit --amend -s`.

hack/deploy-metrics-adapter.sh
sbdtu5498: I am still a bit confused, as it is a part of the same cluster kube-apiserver itself is capable of handling all such requests easily. Then why we are using two contexts here? I am not sure whether any permission boundaries exist or need to exist in the case of aggregated API.

hack/deploy-metrics-adapter.sh
sbdtu5498: The point is, karmada-apiserver is just an extension of kube-apiserver and in that case, even if we had a single context only for the HOST_CLUSTER, wouldn't it still work in the same way?

jwcesign: The architecture looks like K8s on K8s:
image

karmada-apiserver context is used to connect to the kube-apiserver in karmada-system.
karmada-host is used to connect to the kube-apiserver in the host cluster.

sbdtu5498:
> Hello, @sbdtu5498! You can pass the DCO checking by running `git commit --amend -s`.

Oh, shoot! did I miss it somewhere?

hack/deploy-metrics-adapter.sh
jwcesign: Karmada is responsible for workload propagation, if we use the clusters' apiserver, the workload will be created in the cluster, but what we want is: the deployment should be created in member clusters. So we have to create ns, deploy another K8s there, and cut some functions for not creating real workloads there, just sync it to member clusters with policy.

hack/deploy-metrics-adapter.sh
sbdtu5498: Ah, this makes more sense. It is namespaced in the same cluster for security and flexibility purposes.

jwcesign:
Hi @sbdtu5498 
There are only tow nits:
1. fix DOC checking with `git commit --amend -s`
2. Define the flags to specify the kubeconfig file path and context name for karmada-apiserver

sbdtu5498:
@jwcesign I just checked, DCO problem is persisting because a commit that came from reverting doesn't include the standard DCO comment that is probably being checked by the DCO bot, though it is signed. I will fix it. And regarding the kubeconfig file, I think we don't need a new kubeconfig file. It is totally possible for the same kubeconfig file to work differently with different contexts and I just checked hack/deploy_karmada.sh, they are also only passing HOSTC_CLUSTER_CONFIG as a parameter. So I don't think it should be a requirement. Please confirm.

jwcesign:
> So I don't think it should be a requirement. Please confirm.

Hi @sbdtu5498 
I think it's ok if we only specify only one Kubeconfig.

jwcesign:
By the way, you need to squash all the commits into one.

RainbowMango:
> By the way, you need to squash all the commits into one.

Here is the guide for squash commits:
https://karmada.io/docs/contributor/github-workflow#squash-commits

jwcesign (commented):
Great job! Thank you for your contribution! @sbdtu5498 

/lgtm

cc @RainbowMango  for final checking

sbdtu5498:
> Great job! Thank you for your contribution! @sbdtu5498

Thanks a lot, @jwcesign!

RainbowMango (commented):
/assign
I will add it to my queue and will try to review it tomorrow.

RainbowMango (commented):
We need to grant permission to this script, like what we did for other scripts:
```
-rwxr-xr-x  1 root root  2361 Jan 10 11:45 deploy-agent-and-estimator.sh
-rwxr-xr-x  1 root root  3840 Jan 10 11:45 deploy-karmada-agent.sh
-rwxr-xr-x  1 root root  1991 Jan 10 11:45 deploy-karmada-opensearch.sh
-rwxr-xr-x  1 root root 16546 Jun  5 19:28 deploy-karmada.sh
-rw-r--r--  1 root root  1999 Jun 15 17:36 deploy-metrics-adapter.sh   // we need 755
```

You can do that by following commands:
1. # chmod 755 hack/deploy-metrics-adapter.sh
2. # git add hack/deploy-metrics-adapter.sh
3. # git commit --amend
4. # git push origin -f

sbdtu5498:
@RainbowMango done.

hack/deploy-metrics-adapter.sh
```diff
@@ -0,0 +1,67 @@
...
+echo "using kubeconfig: "$KUBECONFIG
+
+# deploy karmada-metrics-adapter
+kubectl --context="${HOST_CLUSTER_NAME}" apply -f "${REPO_ROOT}/artifacts/deploy/karmada-metrics-adapter.yaml"
+
+# make sure that karmada-metrics-adapter is ready
+util::wait_pod_ready "${HOST_CLUSTER_NAME}" "${KARMADA_METRICS_ADAPTER_LABEL}" "${KARMADA_SYSTEM_NAMESPACE}"
+
+# deploy karmada-metrics-adapter-apiservice
+kubectl --context="karmada-apiserver" apply -f "${REPO_ROOT}/artifacts/deploy/karmada-metrics-adaptor-apiservice.yaml"
```
jwcesign, line 51:
```diff
 # deploy karmada-metrics-adapter-apiservice
-kubectl --context="karmada-apiserver" apply -f "${REPO_ROOT}/artifacts/deploy/karmada-metrics-adaptor-apiservice.yaml"
+kubectl --context="karmada-apiserver" apply -f "${REPO_ROOT}/artifacts/deploy/karmada-metrics-adapter-apiservice.yaml"
```

RainbowMango: What's the difference?

RainbowMango: What do you mean `after this patch`?

jwcesign: adaptor -> adapter.
After making this change, it worked perfectly in my testing environment.

sbdtu5498: I will fix the typo.

sbdtu5498:
@jwcesign it has been fixed. I think it'd be good to merge now. Kindly review. Thanks!

jwcesign:
I have tested it and it works perfectly fine. cc @RainbowMango for approving it.

RainbowMango (commented):
/lgtm

I already approved it, by the way.

Issue #3633, opened 2023-06-05
feat: version selection for karmadactl


jwcesign:
**What would you like to be added**:

**Why is this needed**:
Currently, we are importing the alpha release but encountering issues with version parsing. Specifically, when attempting to parse the git version `v1.7.0-alpha.1-2-gcd9ca28f`, it is incorrectly identifying the version as v1.7.0 which has not yet been released.

So how karmada should select the version? based on the logic of kubeadm, we can do the same as K8s.
![img](https://user-images.githubusercontent.com/19745536/243335781-d33a0968-7995-459e-96b4-f1ec78b37247.png)

To fetch the version from github, we can use following code:
```go
package main

import (
	"context"
	"fmt"
	"github.com/google/go-github/github"
)

func getLatestTag(owner, repo string) (string, error) {
	ctx := context.Background()
	client := github.NewClient(nil)

	release, _, err := client.Repositories.GetLatestRelease(ctx, owner, repo)
	if err != nil {
		return "", err
	}
	return *release.TagName, nil
}

func main() {
	fmt.Println(getLatestTag("karmada-io", "karmada"))
}
```

jwcesign:
cc @XiShanYongYe-Chang @RainbowMango @lonelyCZ for checking

RainbowMango:
What's the definition of `latest release` ?

jwcesign:
the latest release is the latest stable release(not contain pre-release), like https://github.com/karmada-io/karmada/releases/latest

RainbowMango:
I'm asking because each time i draft a new release, GitHub always let me to check if it is a `latest release`. 
Thanks, i get it. Will look into it tomorrow.

PR #3652, opened 2023-06-08
feat: Fix version parse

jwcesign:
**What type of PR is this?**

/kind cleanup

We are now introducing a pre-release that follows the K8s release process. However, in the previous implementation, there was an issue with the image tag used during initialization. Here's an example:
```
karmadactl git version: v1.7.0-alpha.1-3-gf20c721a

Once parsed, the card will utilize version 1.7.0 and Karmada's image tag will also be set to version 1.7.0, which has not yet been released.
```

To fix this problem, we will parse `v1.7.0-alpha.1-3-gf20c721a` as `v1.7.0-alpha.1`, which is what this patch does.

But with this patch, each release should have the CRD assets, which should be done in another PR.

Fixes https://github.com/karmada-io/karmada/issues/3633

```release-note
NONE
```

### DIFF
--- a/pkg/version/release.go
+++ b/pkg/version/release.go
@@ -2,8 +2,14 @@ package version
 
 import (
 	"fmt"
+	"regexp"
 
 	utilversion "k8s.io/apimachinery/pkg/util/version"
+)
+
+var (
+	// gitVersionSplitRE is a regexp to split a git version string.
+	gitVersionSplitRE = regexp.MustCompile("-[0-9]+-g[0-9a-z]{7}")
 )
 
 // ReleaseVersion represents a released version.
@@ -14,8 +20,9 @@ type ReleaseVersion struct {
 // ParseGitVersion parses a git version string, such as:
 // - v1.1.0-73-g7e6d4f69
 // - v1.1.0
+// - v1.1.0-alpha.1-3-gf20c721a
 func ParseGitVersion(gitVersion string) (*ReleaseVersion, error) {
-	v, err := utilversion.ParseGeneric(gitVersion)
+	v, err := utilversion.ParseSemantic(gitVersion)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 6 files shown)


jwcesign:
cc @lonelyCZ @RainbowMango

RainbowMango (commented):
Please fix the failing tests.

RainbowMango:
> cc @lonelyCZ @RainbowMango

I don't think the current information can help others understand why this patch is needed.

jwcesign:
cc @RainbowMango ready for checking

shilinlee (approved):

RainbowMango (commented):
/assign

pkg/version/release.go
```diff
@@ -25,22 +27,25 @@ func ParseGitVersion(gitVersion string) (*ReleaseVersion, error) {
...
-	return fmt.Sprintf("v%d.%d.%d", r.Version.Major(), r.Version.Minor(), r.Version.Patch())
+// removeGitVersionCommits removes the git commit info from the version
+// The git version looks like: v1.0.4-14-g2414721
+// the current head of my "parent" branch is based on v1.0.4,
+// but since it has a few commits on top of that, describe has added the number of additional commits ("14")
+// and an abbreviated object name for the commit itself ("2414721") at the end.
+func removeGitVersionCommits(gitVersion string) string {
+	// This match the commit info part of the git version
+	splitRE := regexp.MustCompile("-[0-9]+-g[0-9a-z]{7}")
+	match := splitRE.Split(gitVersion, 2)
```
RainbowMango, line 48:
```suggestion
	matches := splitRE.Split(gitVersion, 2)
```

pkg/version/release.go
```diff
@@ -25,22 +27,25 @@ func ParseGitVersion(gitVersion string) (*ReleaseVersion, error) {
...
+
+	return match[0]
```
RainbowMango, line 50: Can we ensure the `len(match)` is always not `nil`? 
I‘m concerned that there is a panic here for abnormal inputs.

pkg/version/release.go
```diff
@@ -25,22 +27,25 @@ func ParseGitVersion(gitVersion string) (*ReleaseVersion, error) {
 	}, nil
 }
 
-// FirstMinorRelease returns the minor release but the patch releases always be 0(vx.y.0). e.g:
-// - v1.2.1-12-g2eb92858 --> v1.2.0
-// - v1.2.3-12-g2e860210 --> v1.2.0
-func (r *ReleaseVersion) FirstMinorRelease() string {
+// ReleaseVersion returns the current version with format "vx.y.z".
```
RainbowMango, line 30: This function not only returns the `vx.y.z` anymore, but including a pre-releases format.
Please update the comments as well.

pkg/version/release.go
```diff
@@ -15,7 +16,8 @@ type ReleaseVersion struct {
 // - v1.1.0-73-g7e6d4f69
 // - v1.1.0
 func ParseGitVersion(gitVersion string) (*ReleaseVersion, error) {
-	v, err := utilversion.ParseGeneric(gitVersion)
+	filteredVersion := removeGitVersionCommits(gitVersion)
```
RainbowMango, line 19: The `ParseGitVersion` is responsible for parsing a version with `string format` to a structured version type, it doesn't know how this version will be consumed. 

I think the `remove git version` should be delayed to `ReleaseVersion()`.

RainbowMango (commented):
Generally looks good, just with some nits.

pkg/version/release.go
jwcesign: I tried with the following code, there is no panic, and work as expected:
```go
package main

import (
	"fmt"
	"regexp"
)

func removeGitVersionCommits(v string) string {
	splitRE := regexp.MustCompile("-[0-9]+-g[0-9a-z]{7}")
	matches := splitRE.Split(v, 2)

	fmt.Println(matches)
	return matches[0]
}

func main() {
	fmt.Println(removeGitVersionCommits(""))
}
```
The output:
```sh
[]
```

RainbowMango: OK.

pkg/version/release.go
```diff
@@ -25,22 +32,29 @@ func ParseGitVersion(gitVersion string) (*ReleaseVersion, error) {
 	}, nil
 }
 
-// FirstMinorRelease returns the minor release but the patch releases always be 0(vx.y.0). e.g:
+// ReleaseVersion returns the parsed version in the following format::
 // - v1.2.1-12-g2eb92858 --> v1.2.0
 // - v1.2.3-12-g2e860210 --> v1.2.0
```
RainbowMango, line 37: Input `v1.2.1-12-g2eb92858` would get `v1.2.1` rather than `v1.2.0`.

pkg/version/release_test.go
```diff
@@ -4,38 +4,29 @@ import "testing"
...
+			gitVersion:           "v1.1.1-6-gf20c721a",
+			expectReleaseVersion: "v1.1.1",
+			expectError:          false,
 		},
 		{
 			name:        "abnormal version",
 			gitVersion:  "vx.y.z-6-gf20c721a",
 			expectError: true,
 		},
+
```
RainbowMango, line 23: redundant new line.

RainbowMango (approved):
/lgtm
/approve

PR #3656, opened 2023-06-08
Support namespace to use regexp for resourceSelector of cpp and cop

Poor12:
**What type of PR is this?**
/kind feature

For cpp and cop, we expect resourceSelectors can hit multiple namespaces with the similar format, otherwise we may need to edit cpp and cop when we add a new namespace.

Assume that there is a use case that all business departments are registered to Karmada uniformly and isolated through different namespaces. It is agreed to use the format of "test-xxx" uniformly. Now they have registered two departments and have a global cpp to propagate their applications. Once they add a new department, they do not want to edit the cpp because they may cause rescheduling of applications.

In that case, the following config may help.
```
spec:
    resourceSelectors:
    - apiVersion: apps/v1
      kind: Deployment
      Namespace: test-.*
      Name: test
```

Fixes #

```release-note
ClusterPropagationPolicy/ClusterOverridePolicy: Support namespace to use regexp for resourceSelector.
karmada-webhook: validate that namespace of resourceSelector must be the same as that of pp/op.
```

### DIFF
--- a/pkg/util/validation/validation.go
+++ b/pkg/util/validation/validation.go
@@ -2,6 +2,7 @@ package validation
 
 import (
 	"fmt"
+	"regexp"
 	"strings"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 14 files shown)


RainbowMango:
This issue is tracked by 

https://github.com/karmada-io/karmada/actions/runs/5211054867/jobs/9403105260?pr=3656
```
Waiting for kubeconfig file /home/runner/.kube/members.config and clusters member3 to be ready...
Waiting for running............................................................................................................................................................................................................................................................................................................
Error:  Timeout waiting for condition running
```

pkg/util/validation/validation.go
```diff
@@ -23,6 +24,21 @@ func ValidatePropagationSpec(spec policyv1alpha1.PropagationSpec) field.ErrorLis
...
+	allErrs = append(allErrs, ValidateResourceSelectors(spec.ResourceSelectors, field.NewPath("spec").Child("resourceSelectors"))...)
+	return allErrs
+}
+
+// ValidateResourceSelectors validates ResourceSelectors before creation or update.
+func ValidateResourceSelectors(resourceSelectors []policyv1alpha1.ResourceSelector, fldPath *field.Path) field.ErrorList {
+	var allErrs field.ErrorList
+	for _, rs := range resourceSelectors {
+		if len(rs.Namespace) > 0 {
+			_, err := regexp.Compile(rs.Namespace)
```
chaunceyjiang, line 36:
```suggestion
			_, err := regexp.Compile("^" + rs.Namespace + "$")
```

Poor12: Nice finding！

chaunceyjiang (commented):
LGTM

GitHubxsy:
Very nice feature

RainbowMango (commented):
Please fix the failing CI. I'll take a look later.

RainbowMango:
Just realized, that this approach might conflict with the [PropagationPolicy Preemption](https://github.com/karmada-io/karmada/tree/master/docs/proposals/scheduling/policy-preemption) feature:
> we narrow down the scope of preemption by introducing a restriction: for pp, the name in ResourceSelector must be specified;  for cpp, the namespace/name for namespace-scoped resources and the name for cluster-scoped resources must be specific.

RainbowMango:

From an API perspective, I think each field should have clear meanings, otherwise, we might face challenges for both users and implementations.
So, I guess we can postpone this until there is a clear use case and an improved solution.
/close

PR #3453, opened 2023-04-23
karmadactl init add timeout flag

tanberBro:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:
Fixes https://github.com/karmada-io/karmada/issues/3406
Fixes #3611

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #3665, opened 2023-06-13
karmadactl init add timeout flag

### LINKED ISSUE #3406: karmadactl init add timeout flag
karmadactl init an optional timeout  flag to wait for the component to be ready.
**Why is this needed**:
The init command currently uses the default timeout to wait for the component to be ready.
But when I actually used it, due to my poor network environment, I found that the default timeout was too short, causing each init command to time out and exit.
So I want to add a timeout flag to configure the timeout time.

### LINKED ISSUE #3611: karmadactl init timedout
`karmadactl init ` has failed many times during [kurator](https://github.com/kurator-dev/kurator) test.

The logs shows like below error, i believe it is because of timeout deploying components.

```
error: wait for Deployment(default/karmada-apiserver) rollout: timedout wairting for......
```

After digging into it, karmada `WaitComponentReadyTimeout` is 30s when running init, this is a constant. 

Would suggest make it configurable, otherwise it is very fragile for bad network env

helen-frank:
**What type of PR is this?**
/kind feature

see https://github.com/karmada-io/karmada/pull/3453
Fixes #3406
Fixes #3611

/cc @lonelyCZ @XiShanYongYe-Chang 

```release-note
`karmadactl`: The `--wait-component-ready-timeout` flag has been introduced in the `init` command to specify the component installation timeout.
```

### DIFF
--- a/pkg/karmadactl/cmdinit/cmdinit.go
+++ b/pkg/karmadactl/cmdinit/cmdinit.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/karmada-io/karmada/pkg/karmadactl/cmdinit/cert"
 	"github.com/karmada-io/karmada/pkg/karmadactl/cmdinit/kubernetes"
+	cmdinitoptions "github.com/karmada-io/karmada/pkg/karmadactl/cmdinit/options"
 	"github.com/karmada-io/karmada/pkg/karmadactl/options"
 	"github.com/karmada-io/karmada/pkg/karmadactl/util"
 	"github.com/karmada-io/karmada/pkg/version"
@@ -139,7 +140,7 @@ func NewCmdInit(parentCommand string) *cobra.Command {
 	flags.Int32VarP(&opts.KarmadaWebhookReplicas, "karmada-webhook-replicas", "", 1, "Karmada webhook replica set")
 	flags.StringVarP(&opts.KarmadaAggregatedAPIServerImage, "karmada-aggregated-apiserver-image", "", kubernetes.DefaultKarmadaAggregatedAPIServerImage, "Karmada aggregated apiserver image")
 	flags.Int32VarP(&opts.KarmadaAggregatedAPIServerReplicas, "karmada-aggregated-apiserver-replicas", "", 1, "Karmada aggregated apiserver replica set")
-
+	flags.IntVarP(&opts.WaitComponentReadyTimeout, "wait-component-ready-timeout", "", cmdinitoptions.WaitComponentReadyTimeout, "Wait for karmada component ready timeout. 0 means wait forever")
 	return cmd
 }
 
--- a/pkg/karmadactl/cmdinit/kubernetes/deploy.go
+++ b/pkg/karmadactl/cmdinit/kubernetes/deploy.go
@@ -128,9 +128,12 @@ type CommandInitOption struct {
 	RestConfig                         *rest.Config
 	KarmadaAPIServerIP                 []net.IP
 	HostClusterDomain                  string
+	WaitComponentReadyTimeout          int
 }
 
 // Validate Check that there are enough flags to run the command.
+//
+//nolint:gocyclo
 func (i *CommandInitOption) Validate(parentCommand string) error {
 	if i.KarmadaAPIServerAdvertiseAddress != "" {
 		if netutils.ParseIPSloppy(i.KarmadaAPIServerAdvertiseAddress) == nil {
@@ -152,6 +155,10 @@ func (i *CommandInitOption) Validate(parentCommand string) error {
 
 	if i.EtcdStorageMode == etcdStorageModePVC && i.StorageClassesName == "" {
 		return fmt.Errorf("when etcd storage mode is PVC, storageClassesName is not empty. See '%s init --help'", parentCommand)
+	}
+
+	if i.WaitComponentReadyTimeout < 0 {
+		return fmt.Errorf("wait-component-ready-timeout must be greater than or equal to 0")
 	}
 
 	supportedStorageMode := SupportedStorageMode()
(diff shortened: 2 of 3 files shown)


pkg/karmadactl/cmdinit/cmdinit.go
```diff
@@ -139,7 +139,7 @@ func NewCmdInit(parentCommand string) *cobra.Command {
 	flags.Int32VarP(&opts.KarmadaWebhookReplicas, "karmada-webhook-replicas", "", 1, "Karmada webhook replica set")
 	flags.StringVarP(&opts.KarmadaAggregatedAPIServerImage, "karmada-aggregated-apiserver-image", "", kubernetes.DefaultKarmadaAggregatedAPIServerImage, "Karmada aggregated apiserver image")
 	flags.Int32VarP(&opts.KarmadaAggregatedAPIServerReplicas, "karmada-aggregated-apiserver-replicas", "", 1, "Karmada aggregated apiserver replica set")
-
+	flags.IntVarP(&opts.WaitComponentReadyTimeout, "wait-component-ready-timeout", "", 30, "Wait for karmada component ready timeout")
```
RainbowMango, line 142: We shouldn't change the default timeout during introducing the flag.

pkg/karmadactl/cmdinit/kubernetes/deploy.go
```diff
@@ -154,6 +157,10 @@ func (i *CommandInitOption) Validate(parentCommand string) error {
 		return fmt.Errorf("when etcd storage mode is PVC, storageClassesName is not empty. See '%s init --help'", parentCommand)
 	}
 
+	if i.WaitComponentReadyTimeout < 0 {
+		return fmt.Errorf("wait-component-ready-timeout must be greater than or equal to 0")
```
RainbowMango, line 161: Can the timeout equal `0`?

pkg/karmadactl/cmdinit/cmdinit.go
helen-frank: I've taken @lonelyCZ's advice
https://github.com/karmada-io/karmada/pull/3453#discussion_r1191947499

pkg/karmadactl/cmdinit/kubernetes/deploy.go
helen-frank: 0 is a viable option, representing never exit

RainbowMango: Good! If `0` means wait for ever, then we'd better explain it in flag usage, because it might be useful.

pkg/karmadactl/cmdinit/cmdinit.go
RainbowMango: I guess that's just an example of using a literal value rather than a variable.

helen-frank: In fact, I think both can be used, and it is good to use the default value uniformly

pkg/karmadactl/cmdinit/cmdinit.go
```diff
@@ -139,7 +140,7 @@ func NewCmdInit(parentCommand string) *cobra.Command {
 	flags.Int32VarP(&opts.KarmadaWebhookReplicas, "karmada-webhook-replicas", "", 1, "Karmada webhook replica set")
 	flags.StringVarP(&opts.KarmadaAggregatedAPIServerImage, "karmada-aggregated-apiserver-image", "", kubernetes.DefaultKarmadaAggregatedAPIServerImage, "Karmada aggregated apiserver image")
 	flags.Int32VarP(&opts.KarmadaAggregatedAPIServerReplicas, "karmada-aggregated-apiserver-replicas", "", 1, "Karmada aggregated apiserver replica set")
-
+	flags.IntVarP(&opts.WaitComponentReadyTimeout, "wait-component-ready-timeout", "", cmdinitoptions.WaitComponentReadyTimeout, "Wait for karmada component ready timeout. 0 means wait forever")
```
RainbowMango, line 143: How about increasing the default timeout to `120s`? I mean update const here:
https://github.com/karmada-io/karmada/blob/6d2832ff9a6a8263f2320c5f8099dfe34ee7466f/pkg/karmadactl/cmdinit/options/global.go#LL27C2-L27C28

pkg/karmadactl/cmdinit/kubernetes/deploy.go
```diff
@@ -361,7 +368,7 @@ func (i *CommandInitOption) initKarmadaAPIServer() error {
 	if _, err := i.KubeClientSet.AppsV1().StatefulSets(i.Namespace).Create(context.TODO(), etcdStatefulSet, metav1.CreateOptions{}); err != nil {
 		klog.Warning(err)
 	}
-	if err := util.WaitForStatefulSetRollout(i.KubeClientSet, etcdStatefulSet, options.WaitComponentReadyTimeout); err != nil {
+	if err := util.WaitForStatefulSetRollout(i.KubeClientSet, etcdStatefulSet, i.WaitComponentReadyTimeout); err != nil {
```
RainbowMango, line 371: We should follow the same timeout, right?

pkg/karmadactl/cmdinit/kubernetes/deploy.go
```diff
@@ -387,7 +394,7 @@ func (i *CommandInitOption) initKarmadaAPIServer() error {
 	if _, err := i.KubeClientSet.AppsV1().Deployments(i.Namespace).Create(context.TODO(), i.makeKarmadaAggregatedAPIServerDeployment(), metav1.CreateOptions{}); err != nil {
 		klog.Warning(err)
 	}
-	if err := util.WaitForDeploymentRollout(i.KubeClientSet, karmadaAggregatedAPIServerDeployment, options.WaitComponentReadyTimeout); err != nil {
+	if err := util.WaitForDeploymentRollout(i.KubeClientSet, karmadaAggregatedAPIServerDeployment, i.WaitComponentReadyTimeout); err != nil {
```
RainbowMango, line 397: ditto.

RainbowMango (commented):
Generally looks good to me.
1. I'd like to increase the timeout to 120s by default. 
2. Make sure all components share the same timeout.

RainbowMango (approved):
/lgtm
/approve

Thanks for your quick response, impressive!!

PR #3668, opened 2023-06-13
Fix cached RESTMapper when pass no versions

Poor12:
**What type of PR is this?**
/kind cleanup

Now cachedRESTMapper will do nothing but return an error when users do not pass versions parameter. But for the underlying RESTMapper, it can get the corresponding `RESTMapping` without versions. The reference code is [here](https://github.com/kubernetes/apimachinery/blob/6b1428efc73348cc1c33935f3a39ab0f2f01d23d/pkg/api/meta/restmapper.go#LL466C1-L466C1). So when no versions pass, we'd better try to consult the underlying mappers.

Fixes #

None

### DIFF
--- a/pkg/util/restmapper/restmapper.go
+++ b/pkg/util/restmapper/restmapper.go
@@ -1,7 +1,6 @@
 package restmapper
 
 import (
-	"fmt"
 	"sync"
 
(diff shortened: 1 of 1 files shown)


RainbowMango (approved):
/lgtm
/approve

/remove-kind bug
/kind cleanup

PR #3679, opened 2023-06-19
Clean PP association field as it is not used anymore.

yanggangtony:
**What type of PR is this?**
/kind cleanup

Clean PP association field as it is not used anymore.
Its now be replaced by propagateDeps.

Yes

```release-note
Delete PP's association field as it is not used anymore.
```

### DIFF
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -15854,10 +15854,6 @@
         "resourceSelectors"
       ],
       "properties": {
-        "association": {
-          "description": "Association tells if relevant resources should be selected automatically. e.g. a ConfigMap referred by a Deployment. default false. Deprecated: in favor of PropagateDeps.",
-          "type": "boolean"
-        },
         "dependentOverrides": {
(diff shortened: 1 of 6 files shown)


RainbowMango (commented):
I'm afraid we can't do that on v1alpha1 as it is kind of a breaking change.
The field will be removed on v1alpha2 or beta.

yanggangtony:
@RainbowMango 
Thanks for quick review.
So let the pr wait a while to be merged.
😀😀

RainbowMango:
We don't have a plan to promote the API yet. :) 
Let's close it for now because even if we are going to introduce `v1alpha2`, the field in `v1alpha1` shouldn't be removed. That is the deprecated fields will not represent in v1alpha2.

/close

yanggangtony:
Ok.
Got it.

liangyuanpeng:
@RainbowMango  Actually, this field is not working, so it should be not show it on the `kubectl explain`. 

Deprecated != not working.  IMO, PropagationPolicy v1alpha1 is  the wrong behavior for the field of association. It should be working before this field is deleted.

RainbowMango:
Usually, we don't deprecate a `field` without promoting API to the next stage(like v1alpha2), but the `association` was never been implemented. 
I checked #3830 which tried to add a validation for that, looks good to me.  
But still don't know how to handle the `kubectl explain` concern.

Issue #3674, opened 2023-06-16
Ignore Codecov errors


RainbowMango:
**Task description**:
When uploading code coverage data to [Codecov](https://about.codecov.io/), errors occasionally occur that prevent PR merging.
This error is caused by the unstable service of Codecov, we can set the [fail_ci_if_error](https://github.com/karmada-io/karmada/blob/master/.github/workflows/ci.yml#L104) with `false` to ignore this error.

The following error is echoed from https://github.com/karmada-io/karmada/actions/runs/5285457338/jobs/9564066853?pr=3652:
```
[2023-06-16T02:25:55.203Z] ['verbose'] https://codecov.io/upload/v4?package=github-action-3.1.4-uploader-0.5.0&branch=fix-version-parse-v2&build=5285457338&build_url=https%3A%2F%2Fgithub.com%2Fkarmada-io%2Fkarmada%2Factions%2Fruns%2F5285457338&commit=06efa55b46f2ca707a912d86e83044ef0f03e9c5&job=CI+Workflow&pr=3652&service=github-actions&slug=karmada-io%2Fkarmada&name=&tag=&flags=unittests&parent=
        Content-Type: 'text/plain'
        Content-Encoding: 'gzip'
        X-Reduced-Redundancy: 'false'
[2023-06-16T02:26:00.213Z] ['error'] There was an error running the uploader: Error uploading to [https://codecov.io:](https://codecov.io/) Error: getaddrinfo EAI_AGAIN codecov.io
[2023-06-16T02:26:00.213Z] ['verbose'] The error stack is: Error: Error uploading to [https://codecov.io:](https://codecov.io/) Error: getaddrinfo EAI_AGAIN codecov.io
    at main (/snapshot/repo/dist/src/index.js)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
[2023-06-16T02:26:00.213Z] ['verbose'] End of uploader: 5813 milliseconds
Error: Codecov: Failed to properly upload: The process '/home/runner/work/_actions/codecov/codecov-action/v3/dist/codecov' failed with exit code 255
```
**Solution**:
Just set the `fail_ci_if_error` with `false` as per [codecov-action arguments](https://github.com/codecov/codecov-action#arguments).
```diff
diff --git a/.github/workflows/ci.yml b/.github/workflows/ci.yml
index 79b44c00..b87ed8e2 100644
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
[...]

PR #3684, opened 2023-06-20
Add policy preemption proposal

Poor12:
**What type of PR is this?**
/kind design

This proposal propose a strategy which allows policies to support preemption by priority at runtime.
Even if workloads have been propagated by a policy, they can be preempted by a high priority of policy.

Fixes #

None


RainbowMango:
Retrying the CI jobs due to the issue tracked by #3674

docs/proposals/scheduling/policy-preemption/README.md
```diff
@@ -0,0 +1,214 @@
...
+Cluster administrators usually cannot foresee future expansion scenarios when configuring policies.
+They will usually start with a broad policy to set the base strategy. When an application requires special configuration,
+the administrator wants to provide a persionalized policy to take over the application.
+At this time, it hopes that the high-priority policy can preempt the low-priority policy.
+
+### Goals
+
+- Extend the API of PropagationPolicy/ClusterPropagationPolicy to allow high-priority policies to preempt low-priority policies.
+- Propose the implementation ideas for involved components, including `karmada-controller-manager` and `karmada-webhook`.
+- Prevent preemption caused by misconfiguration.
```
RainbowMango, line 50: > Prevent preemption caused by misconfiguration.

It would be great to give a more detail/clear description of this goal. 

I guess you want to say is that:
```
Propose a solution that is compatible with previous versions. The preemption capability will be disabled by default, upgrading the system from the previous version to the new one will not result in any changes in behavior and will not require any additional adaptation work.
```

docs/proposals/scheduling/policy-preemption/README.md
```diff
@@ -0,0 +1,214 @@
...
+```
+
+`nginx-propagation-preemption` will preempt `nginx-propagation` and the deployment will be scheduled to `member3`.
+
+### Components change
+
+#### karmada-controller-manager
+
+Currently, when a new policy creates/updates, `detector` will first check whether existing objects are no longer matched by the current policy to handle the changes of `resourceSelectors`. If yes, clean the labels on the unmatched object. Then it will find the corresponding ResourceBinding and make sure that PropagationPolicy's updates
+can be synchronized to ResourceBinding. Finally, it will try to find the matched resource templates which have not been scheduled.
```
RainbowMango, line 192: Are you trying to explain the `current` behavior? Just feel some misleading.

docs/proposals/scheduling/policy-preemption/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+Karmada needs to record the mapping relationship between policy and resource templates. For each preemption operation, the component needs to have a corresponding metrics record.
```
RainbowMango, line 194: > For each preemption operation, the component needs to have a corresponding metrics record.

Can a metric record each preemption operation?

RainbowMango: Can you propose the relevant metrics as well in this proposal?

docs/proposals/scheduling/policy-preemption/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+creation-date: 2023-06-19
+
+---
+
+# Policy supports preemption by priority at runtime
+
+## Summary
+
+Currently, `PropagationPolicy` and `ClusterPropagationPolicy` already have a notion of priority. If multiple PropagationPolicies match the workload, Karmada will select the one with the highest priority. This is the case when the workload has not been selected by a policy.
```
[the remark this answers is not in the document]
Poor12: Updated.

docs/proposals/scheduling/policy-preemption/README.md
Poor12: Updated.

docs/proposals/scheduling/policy-preemption/README.md
Poor12: I have thought about it. It may be redundant to record the details of each preemption, such as which policy is preempted by the record. This is more appropriate to use event to express. For metrics, we can at least record the total number of times a resource template is preempted.

docs/proposals/scheduling/policy-preemption/README.md
Poor12: Yes, sorry, maybe the description is not accurate.

docs/proposals/scheduling/policy-preemption/README.md
```diff
@@ -0,0 +1,229 @@
...
+Workloads are responsible for each business team. For the current business, I deploy a generic policy to match all.
+At some point, I want to change the scheduling strategy of one of the businesses, so I create a higher-priority policy.
+I hope that the high-priority policy can preempt the low-priority policy.
+
+#### As a cluster administrator, I hope that the propagationPolicy can preempt the clusterPropagationPolicy.
+
+As a cluster administrator, I am responsible for unified management of multi-cluster scheduling policies.
+Workloads are responsible for each business team. For the current business, I deploy a cpp to match all namespace-scoped resources.
+At some point, I want to change the scheduling strategy of one of the businesses, so I create a pp to match it only.
+I hope that the propagationPolicy can preempt the clusterPropagationPolicy.
```
RainbowMango, line 71: Is this a real use case? @GitHubxsy

RainbowMango: My concern is that not all users are trusted, a malicious user could create PropagationPolicy at the highest possible priorities,  causing other ClusterPropagationPolicy to be preempted.

Poor12: /cc @GitHubxsy to verify.

RainbowMango (commented):
/assign

docs/proposals/scheduling/policy-preemption/README.md
```diff
@@ -0,0 +1,256 @@
...
+      name: nginx
+  priority: 3
+  preemption: Always
+  placement:
+    clusterAffinity:
+      clusterNames:
+        - member3
+```
+
+`nginx-propagation-preemption` will preempt `nginx-propagation` and the deployment will be scheduled to `member3`.
```
chaunceyjiang, line 229: Do you mean that when a `PropagationPolicy` is set to `preemption: Always`, it will preempt another lower priority `PropagationPolicy` instead of **being preempted** by a higher priority `PropagationPolicy`?

chaunceyjiang: If I am an application developer, I distribute my application to a member cluster using a PropagationPolicy(`preemption: Never`). If another user preemptively uses a PropagationPolicy(`preemption: Always`) for my application, I may not even be aware of this behavior.

chaunceyjiang: I thought that when the `PropagationPolicy `is set to `preemption: Never`, it is not allowed to **be preempted**, while when the `PropagationPolicy` is set to `preemption: Always`, it is always allowed to **be preempted** by other users.

RainbowMango: > If I am an application developer, I distribute my application to a member cluster using a PropagationPolicy(preemption: Never). If another user preemptively uses a PropagationPolicy(preemption: Always) for my application, I may not even be aware of this behavior.

This is exactly what I'm concerned about.  I have been thinking do we need something like `I can preempt others, but not allows to be preempted`? 

Now, I think the concern could be solved by managing `Priority` properly, such as by managing who has permission to create which priority policies.

Poor12: > If I am an application developer, I distribute my application to a member cluster using a PropagationPolicy(`preemption: Never`). If another user preemptively uses a PropagationPolicy(`preemption: Always`) for my application, I may not even be aware of this behavior.

Regardless of whether this field is placed in the policy of preempting others or the policy of being preempted, there are essentially similar problems. If the above scenario is followed, then every time I want to deploy a higher priority policy and preempt, I need to perceive the old bound policy and update it(preemption: never -> always). **As an application developer, it's not convenient. It's not safe either specially when these two policies are owned by different users. You tampered with other people's policy.**

I agree with @RainbowMango. The concern could be solved by managing `Priority` properly. All preemptive behavior is defined by a set of published rules. Users need to understand the impact of the policy priority it sets on the system when it enables preemption.

docs/proposals/scheduling/policy-preemption/README.md
```diff
@@ -0,0 +1,256 @@
...
+const (
+    // PreemptAlways means that preemption is allowed.
+    //
+    // If it is applied to a PropagationPolicy, it can preempt any resource as
+    // per Priority, regardless of whether it has been claimed by a PropagationPolicy
+    // or a ClusterPropagationPolicy, as long as it can match the rules defined
+    // in ResourceSelector. In addition, if a resource has already been claimed by a 
+    // ClusterPropagationPolicy, the PropagationPolicy can still preempt it
+    // without considering Priority.
+    //
```
chaunceyjiang, line 167: > // if a resource has already been claimed by a 
// ClusterPropagationPolicy, the PropagationPolicy can still preempt it
    // without considering Priority.


Do you mean if a low-priority  PropagationPolicy can preempt a high-priority ClusterPropagationPolicy?

Poor12: Yeah, this is consistent with the behavior of the first match(always look for pp first and then cpp).

chaunceyjiang: > Yeah, this is consistent with the behavior of the first match(always look for pp first and then cpp).


Yes, but there is a condition: this resource has not been claimed by any pp or cpp. If a resource has already been claimed by a cpp, then the pp will also not take effect.

According to the new proposal(`preemption: Always`), the cpp will become invalid regardless of whether this resource has been declared by a cpp or not, then the pp will take effect. Right?

Poor12: If this resource has been claimed by a cpp, a pp will preempt the resource if it enables preemption and also match it.
What I want to emphasize is that whether it's an initial match or a preemption, we always follow the order:
high-priority pp > low-priority pp > high-priority cpp > low-priority cpp.

whitewindmills (commented):
I think we need to add feature gate `PolicyPreemption` to protect this feature.

docs/proposals/scheduling/policy-preemption/README.md
```diff
@@ -0,0 +1,256 @@
...
+This proposal maintains the backward compatibility, the system built with previous
+versions of Karmada can be seamlessly migrated to the new version.
+The previous configurations(yamls) could be applied to the new version of Karmada and
+without any behavior change.
+
+### Notes/Constraints/Caveats (Optional)
+
+The priority based on preemption do not include implicit priority when `.spec.priority` of policies are the same.
+
+For example, the following two policies do not preempt.
```
whitewindmills, line 89: Do you mean that `propagation-namematch` won't preempt `propagation-labelselectormatch` even if preemption is enabled?

Poor12: Yeah.

Poor12:
> I think we need to add feature gate `PolicyPreemption` to protect this feature.

I think it does not need. We already provide this field to enable the feature and it's `Never` by default.

RainbowMango:
Hi @chaunceyjiang @whitewindmills Do you have any further comments or questions?

docs/proposals/scheduling/policy-preemption/preemption_process.PNG
chaunceyjiang, line 1: image


What does this mean?

whitewindmills:
> I think it does not need. We already provide this field to enable the feature and it's Never by default.

Other than that I have no problem. The new feature may have unexpected impact on the original environment if the administrator enabled this new feature but did not know enough about it. And feature gate as global switch can decide whether the new feature is enabled. For exmaple as `PropagateDeps` feature which field is also `false` by default to disable it, but its feature gate is also provided.
/cc @RainbowMango

RainbowMango:
@Poor12 asked me for suggestions about the feature gate, I said the introduced `preemption` field is a kind of feature gate now, so we might not need to introduce a dedicated feature gate.

But as I mentioned on https://github.com/karmada-io/karmada/pull/3805#issuecomment-1641270486, we might need a more details design for it, if we are not fully confident, it would be better to provide an additional feature gate, let's take this as an experimental feature.

whitewindmills:
thanks, let's continue the previous discussion in this [PR](https://github.com/karmada-io/karmada/pull/3805).
LGTM

Poor12:
I update the proposal and add more details.

docs/proposals/scheduling/policy-preemption/resourcetemplate_policy.PNG
whitewindmills, line 1: another question, there are 3 policy: A > B > C, and resource template was bound to C. If B is allowed to preempt now, resource template will be clean up its policy labels and try to find matched policy. so should it be bound to A or B in the end? if A is also allowed to preempt, should it be bound to A or B in the end?

RainbowMango:
To be honest, I'm concerned about this feature, mainly because preemption is a dangerous operation that easily lead to unexpected consequences.  I can feel the concerns from the comments at https://github.com/karmada-io/karmada/pull/3684#discussion_r1263446901, https://github.com/karmada-io/karmada/pull/3684#discussion_r1265000327, and we do facing the scalability challenges during the implementation, such as https://github.com/karmada-io/karmada/pull/3805#issuecomment-1639876325 that might need to list all resource templates during response to preemption. I'm afraid that will have a significant impact on system performance. 

In addition, we also need to handle some corner cases like https://github.com/karmada-io/karmada/pull/3684#discussion_r1267702349 which seems is missing from the proposal.

So, after revisiting the user case, when users want to use a PP to preempt a certain resource, `the resource is often specific and can be specifically declared in the PP`(by resource selector), e.g.
```yaml
kind: PropagationPolicy
metadata:
  name: foo
spec:
  resourceSelectors:
    - apiVersion: apps/v1
      kind: Deployment
      name: nginx             # with a concrete name
```
Or with multiple resources:
```yaml
kind: PropagationPolicy
metadata:
  name: nginx-propagation
spec:
  resourceSelectors:
    - apiVersion: apps/v1
      kind: Deployment
      name: nginx            # with a concrete name
    - apiVersion: v1
      kind: ConfigMap
      name: nginx            # with a concrete name
```

So, I'm thinking if we can narrow down the scope of preemption, that is by introducing a restriction that:
`the resource name in ResourceSelecto must be specified`. If it is acceptable then it will be a relief for users to adopt the feature and for the burden of implementation.

docs/proposals/scheduling/policy-preemption/resourcetemplate_policy.PNG
RainbowMango: Nice finding.

RainbowMango:
@chaunceyjiang @GitHubxsy @whitewindmills What do you think of https://github.com/karmada-io/karmada/pull/3684#issuecomment-1641751052?

whitewindmills:
If possible, we can have a brief meeting to discuss, I have some questions about this proposal.

RainbowMango:
Sounds great, I'll contact you to schedule the meeting. And will post the meeting here.

RainbowMango:
Talked to @whitewindmills. 
We are going to call the meeting at 10:00 this morning. 
Meeting link: https://zoom.com/my/karmada

RainbowMango:
Meeting minutes:

Members Present:
- @chaunceyjiang @Garrybest @GitHubxsy @RainbowMango @wangyuan249 @whitewindmills 

- Reached an agreement: narrow down the scope of preemption as per https://github.com/karmada-io/karmada/pull/3684#issuecomment-1641751052. That is following restrictions should be added to resource selectors:
  - For `PropagationPolicy` , the `ApiVersion`/`Kind`/`Name` must be provided when enabling preemption. 
  - For `PropagationPolcy`, the `ApiVersion`/`Kind`/`Name`/`Namespace` must be provided when enabling preemption. 
- Introducing a feature gate for this feature(Considering the risk of this feature)
- Recognized a corner case that `How to transfer control to another PP/CPP during de-prioritize a PP/CPP?`, no ideal solution available yet, talked alternatives:
   - 1. Hack the PP update handler, such as [HandlePropagationPolicyCreationOrUpdate](https://github.com/karmada-io/karmada/blob/48e9a5b55163844b11b6ead2bec32076ba63449a/pkg/detector/detector.go#L983-L1027).
   - 2. Adding an annotation by `karmada-webhook` to record the priority change, and then reconcile in detector.
   - 3. maintain a mapping table between `PP(with .spec.preemption:always)` and `resource template` in memory.

RainbowMango (approved):
/lgtm
/approve

We can iterate the proposal during or after the implementation.

PR #3688, opened 2023-06-21
cleanup: the optimized update cachedResources code for multi cluster cache

xigang:
**What type of PR is this?**

/kind cleanup

optimization update `cachedResources` logic.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/search/proxy/store/multi_cluster_cache.go
+++ b/pkg/search/proxy/store/multi_cluster_cache.go
@@ -98,16 +98,8 @@ func (c *MultiClusterCache) UpdateCache(resourcesByCluster map[string]map[schema
 			newCachedResources[resource] = struct{}{}
 		}
 	}
-	for resource := range c.cachedResources {
-		if _, exist := newCachedResources[resource]; !exist {
-			delete(c.cachedResources, resource)
-		}
-	}
-	for resource := range newCachedResources {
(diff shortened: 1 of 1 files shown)


pkg/search/proxy/store/multi_cluster_cache.go
```diff
@@ -98,16 +98,8 @@ func (c *MultiClusterCache) UpdateCache(resourcesByCluster map[string]map[schema
...
-		if _, exist := newCachedResources[resource]; !exist {
-			delete(c.cachedResources, resource)
-		}
-	}
-	for resource := range newCachedResources {
-		if _, exist := c.cachedResources[resource]; !exist {
-			c.cachedResources[resource] = struct{}{}
-		}
-	}
+	c.cachedResources = newCachedResources
```
whitewindmills, line 101: Seems `c.cachedResources` value never be used. Can you replace `map` with `sets.Set` to modify `c.cachedResources` definition?

xigang: @whitewindmills This part of the code is used.

```
// HasResource return whether resource is cached.
func (c *MultiClusterCache) HasResource(resource schema.GroupVersionResource) bool {
	c.lock.RLock()
	defer c.lock.RUnlock()
	_, ok := c.cachedResources[resource]
	return ok
}
```

PR #3689, opened 2023-06-21
Add option way toget rest config of estimator schedule

ReOneK:
Get rest config from aggregate apiserver if member cluster kubeconfig not provided.

**What type of PR is this?**

Fixes #

```release-note

```


whitewindmills (commented):
Could you show your use case?

ReOneK:
> Could you show your use case?

In my use case,users may access the karmada control plane in different machine, some may not have member kubeconf exactly, and can also enable estimator across the control plane kubeconf, it's hard for user to manage all member cluster's kubeconf.
On the other hand, mounting kubeconfig explicitly is not secure in my case.It also doesn't make sense to provide kubeconfig after the cluster has been joined.

PR #3694, opened 2023-06-22
proposal for native service discovery

### LINKED ISSUE #2384: [MCS] Remove "derived-" prefix of MCS
Remove "derived-" prefix of MCS， so user can use the same k8s service name in other k8s clusters.

**Why is this needed**:
Karmada is supposed to offer one-cluster view. However user need to change there application settings when using MCS.

Maybe there could be an validation to refuse the ServiceImport object when there has been a same-name service in the target cluster.

bivas:
**What type of PR is this?**

/kind design

Fixes #2384

### DIFF
--- a/docs/proposals/service-discovery/README.md
+++ b/docs/proposals/service-discovery/README.md
@@ -1,0 +1,194 @@
+---
+title: Service discovery with native Kubernetes naming and resolution
+authors:
+- "@bivas"
+reviewers:
+- "@XiShanYongYe-Chang"
+- TBD
+approvers:
+- "@robot"
+- TBD
+
+creation-date: 2023-06-22
+
+---
+
+# Service discovery with native Kubernetes naming and resolution
+
+## Summary
+
+With the current `ServiceImportController` when a `ServiceImport` object is reconciled, the derived service is prefixed with `derived-` prefix.
+
+## Motivation
+
+<!--
+This section is for explicitly listing the motivation, goals, and non-goals of
+this KEP.  Describe why the change is important and the benefits to users.
+-->
+Having a `derived-` prefix for `Service` resources seems counterintuitive when thinking about service discovery:
+- Assuming the pod is exported as the service `foo`
+- Another pod that wishes to access it on the same cluster will simply call `foo` and Kubernetes will bind to the correct one
+- If that pod is scheduled to another cluster, the original service discovery will fail as there's no service by the name `foo`
+- To find the original pod, the other pod is required to know it is in another cluster and use `derived-foo` to work properly
+
+### Goals
+
+- Remove the "derived-" prefix from the service
+- User-friendly and native service discovery
+
+### Non-Goals
+
+- Multi cluster connectivity
+
+## Proposal
+
+Following are flows to support the service import proposal:
+
+1. `Deployment` and `Service` are created on cluster member1 and the `Service` imported to cluster member2 using `ServiceImport` (described below as [user story 1](#story-1))
+2. `Deployment` and `Service` are created on cluster member1 and both propagated to cluster member2. `Service` from cluster member1 is imported to cluster member2 using `ServiceImport` (described below as [user story 2](#story-2))
+
+The proposal for this flow is what can be referred to as local-and-remote service discovery. The options could be:
+
+1. **Local** only - In case there's a local service by the name `foo` Karmada never attempts to import the remote service and doesn't create an `EndPointSlice`
+2. **Local** and **Remote** - Users accessing the `foo` service will reach either member1 or member2
+3. **Remote** only - in case there's a local service by the name `foo` Karmada will remove the local `EndPointSlice` and will create an `EndPointSlice` pointing to the other cluster (e.g. instead of resolving member2 cluster is will reach member1)
+
+This proposal suggests to distinguish the services type by the extending the API `multicluster.x-k8s.io/v1alpha1` of `ServiceImport` with Karmada specific annotations:
+```yaml
+apiVersion: multicluster.x-k8s.io/v1alpha1
+kind: ServiceImport
+metadata:
+  name: serve
+  namespace: demo
+  annotations:
+    discovery.karmada.io/strategy: LocalAndRemote
+spec:
+  type: ClusterSetIP
+  ports:
+  - port: 80
+    protocol: TCP
+```
+
+The default could be `LocalAndRemote`.
+
+
+### User Stories (Optional)
+
+<!--
+Detail the things that people will be able to do if this KEP is implemented.
+Include as much detail as possible so that people can understand the "how" of
+the system. The goal here is to make this feel real for users without getting
+bogged down.
+-->
+
+#### Story 1
+
+As a Kubernetes cluster member,
+I want to access a service from another cluster member,
+So that I can communicate with the service using its original name.
+
+**Background**: The Service named `foo` is created on cluster member1 and imported to cluster member2 using `ServiceImport`.
+
+**Scenario**:
+
+1. Given that the `Service` named `foo` exists on cluster member1
+2. And the `ServiceImport` resource is created on cluster member2, specifying the import of `foo`
+3. When I try to access the service inside member2
+4. Then I can access the service using the name `foo.myspace.svc.cluster.local`
+
+#### Story 2
+
+As a Kubernetes cluster member,
+I want to handle conflicts when importing a service from another cluster member,
+So that I can access the service without collisions and maintain high availability.
+
+**Background**: The Service named `foo` is created on cluster member1 and has a conflict when attempting to import to cluster member2.
+Conflict refers to the situation where there is already a `Service` `foo` existing on the cluster (e.g. propagated with `PropagationPolicy`), but we still need to import `Service` `foo` from other clusters onto this cluster (using `ServiceImport`)
+
+**Scenario**:
+
+1. Given that the `Service` named `foo` exists on cluster member1
+2. And there is already a conflicting `Service` named foo on cluster member2
+3. When I attempt to access the service in cluster member2 using `foo.myspace.svc.cluster.local`
+4. Then the requests round-robin between the local `foo` service and the imported `foo` service (member1 and member2)
+
+### Notes/Constraints/Caveats (Optional)
+
+<!--
+What are the caveats to the proposal?
+What are some important details that didn't come across above?
+Go in to as much detail as necessary here.
+This might be a good place to talk about core concepts and how they relate.
+-->
+
+### Risks and Mitigations
+
+<!--
+What are the risks of this proposal, and how do we mitigate? 
+
+How will security be reviewed, and by whom?
+
+How will UX be reviewed, and by whom?
+
+Consider including folks who also work outside the SIG or subproject.
+-->
+Adding a `Service` that resolve to a remote cluster will add a network latency of communication between clusters. 
+
+## Design Details
+
+<!--
+This section should contain enough information that the specifics of your
+change are understandable. This may include API specs (though not always
+required) or even code snippets. If there's any ambiguity about HOW your
+proposal will be implemented, this is the place to discuss them.
+-->
+
+To achieve service discovery with native Kubernetes naming and resolution, the following design details should be considered:
+
+1. Service Import Controller Changes: The Service Import Controller needs to be updated to remove the `"derived-"` prefix from the derived service. This ensures that the imported service retains its original name when accessed within the importing cluster.
+2. Local-Only Service Discovery: To support local-only service discovery, Karmada should be enhanced to check if there is a local service with the same name as the imported service. If a local service exists, the remote service import should be skipped, and no `EndPointSlice` should be created for the remote service. This allows local services to be accessed directly without going through the remote service.
+3. Local and Remote Service Discovery: When a service is imported from another cluster, both local and remote services should be accessible. Users accessing the service should be able to reach either the local or remote member based on their location and cluster routing. This enables high availability and load balancing between the local and remote instances of the service.
+4. Remote-Only Service Discovery: In cases where there is local service with the same name as the imported service, Karmada should remove the local `EndPointSlice` and create an `EndPointSlice` pointing to the remote cluster. This allows users to access the service through the imported `EndPointSlice`, effectively reaching the remote cluster.
+
+Implementing the service registration flow, the following should be considered:
+1. Given that the `Service` named `foo` exists on cluster member1
+2. The `ServiceImport` resource is created on cluster member2, specifying the import of `foo`
+3. A `Service` resource is created in cluster member2, but does not prefix the name with `derived-`
+   1. If there is already an existing `Service` named `foo` on cluster member2
+   2. The `EndPointSlice` associated with the local `foo` service points to the local `Deployment` running on member2
+   3. The controller will ignore the conflict of existing `Service` and will not create a new `Service` object
+4. And the imported `EndPointSlice` is created, pointing to the remote `Deployment` running on member1
+5. And the `EndPointSlice` associated with the imported service is prefixed with `import-`
+6. Then the requests round-robin between the local `foo` service and the imported `foo` service (member1 and member2)
+
+### Test Plan
+
+<!--
+**Note:** *Not required until targeted at a release.*
+
+Consider the following in developing a test plan for this enhancement:
+- Will there be e2e and integration tests, in addition to unit tests?
+- How will it be tested in isolation vs with other components?
+
+No need to outline all test cases, just the general strategy. Anything
+that would count as tricky in the implementation, and anything particularly
+challenging to test, should be called out.
+
+-->
+
+## Alternatives
+
+<!--
+What other approaches did you consider, and why did you rule them out? These do
+not need to be as detailed as the proposal, but should include enough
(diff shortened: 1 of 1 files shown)


docs/proposals/service-discovery/README.md
```diff
@@ -0,0 +1,139 @@
...
+<!--
+Detail the things that people will be able to do if this KEP is implemented.
+Include as much detail as possible so that people can understand the "how" of
+the system. The goal here is to make this feel real for users without getting
+bogged down.
+-->
+
+#### Story 1
+
+`Service` from member1 is available in member2.
```
XiShanYongYe-Chang, line 67: Can you help provide a more detailed description of this story? Perhaps we can learn from the writing style used here: https://www.techtarget.com/searchsoftwarequality/definition/user-story#:~:text=A%20user%20story%20is%20a,simplified%20description%20of%20a%20requirement.

docs/proposals/service-discovery/README.md
```diff
@@ -0,0 +1,139 @@
...
+Include as much detail as possible so that people can understand the "how" of
+the system. The goal here is to make this feel real for users without getting
+bogged down.
+-->
+
+#### Story 1
+
+`Service` from member1 is available in member2. 
+
+The service can be accessed inside member2 as an ordinary `Service` with the name `foo.myspace.svc.cluster.local`. The import process doesn't prefix the service with `derived-` prefix but still uses `import-` for the `EndPointSlice`
```
XiShanYongYe-Chang, line 69: This description looks more like a specific implementation plan. We can place it in `Design Details`.

docs/proposals/service-discovery/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+How will security be reviewed, and by whom?
+
+How will UX be reviewed, and by whom?
+
+Consider including folks who also work outside the SIG or subproject.
+-->
+
+## Design Details
+
```
XiShanYongYe-Chang, line 105: We can provide a detailed description of the overall design scheme here.

docs/proposals/service-discovery/README.md
```diff
@@ -0,0 +1,139 @@
...
+Following are flows to support the service import proposal:
+
+1. `Deployment` and `Service` are created on cluster member1 and the `Service` imported to cluster member2 using `ServiceImport` (described below as [user story 1](#story-1))
+2. `Deployment` and `Service` are created on cluster member1 and both propagated to cluster member2. `Service` from cluster member1 is imported to cluster member2 using `ServiceImport` (described below as [user story 2](#story-2))
+
+The proposal for this flow is what can be referred to as local-and-remote service discovery. The options could be:
+
+1. **Local** only - In case there's a local service by the name `foo` Karmada never attempts to import the remote service and doesn't create an `EndPointSlice`
+2. **Local** and **Remote** - Users accessing the `foo` service will reach either member1 or member2
+3. **Remote** only - in case there's a local service by the name `foo` Karmada will remove the local `EndPointSlice` and will create an `EndPointSlice` pointing to the other cluster (e.g. instead of resolving member2 cluster is will reach member1)
```
XiShanYongYe-Chang, line 54: `Remote` under what circumstances would this be used? I feel like we could consider temporarily not supporting this method.

XiShanYongYe-Chang (commented):
Thanks a lot, @bivas I Just passed through the holiday season, and I am delighted to see your proposal. Let's work together to promote this proposal.

docs/proposals/service-discovery/README.md
bivas: `Remote` can be used for cases like privacy or business consideration that doesn't allow processing of customer data in specific location and we must pass the traffic to privacy controlled cluster.

docs/proposals/service-discovery/README.md
bivas: Will do!

docs/proposals/service-discovery/README.md
XiShanYongYe-Chang: For users, if they want to perform remote processing, why do they need to deploy a deployment in the current cluster? They can simply deploy it in another cluster.

Furthermore, from an implementation perspective, how can we ensure the deletion of EndpointSlice? This is controlled by controllers in the cluster, so even if we delete it, a new one will be created immediately.

bivas: I agree it is a rare case. I can think of a monolith that has multiple entry points and it can read messages of Kafka broker but isn't allowed to get REST API calls. So the user will deploy the code using a `Deployment` but will not want it exposed as REST endpoint. Faulty user architecture, but can work with Karmada. 

I don't have a good idea how to delete it, maybe we can push that feature for later.

XiShanYongYe-Chang: Thanks for your response!

XiShanYongYe-Chang:
Hi @zishen , can you help take a look? cc @zishen

docs/proposals/service-discovery/README.md
```diff
@@ -0,0 +1,171 @@
...
+**Background**: The Service named `foo` is created on cluster member1 and imported to cluster member2 using `ServiceImport`.
+
+**Scenario**:
+
+1. Given that the `Service` named `foo` exists on cluster member1
+2. And the `ServiceImport` resource is created on cluster member2, specifying the import of `foo`
+3. When I try to access the service inside member2
+4. Then I can access the service using the name `foo.myspace.svc.cluster.local`
+5. And the import process does not prefix the service name with `derived-`
+6. And the `EndPointSlice` associated with the imported service is prefixed with `import-`
```
XiShanYongYe-Chang, line 80: I think these two points are related to specific implementation and can be explained in the details of the implementation.

docs/proposals/service-discovery/README.md
```diff
@@ -0,0 +1,171 @@
...
+
+#### Story 2
+
+As a Kubernetes cluster member,
+I want to handle conflicts when importing a service from another cluster member,
+So that I can access the service without collisions and maintain high availability.
+
+**Background**: The Service named `foo` is created on cluster member1 and has a conflict when attempting to import to cluster member2.
```
XiShanYongYe-Chang, line 88: Can you provide more specific details about the conflict mentioned here?

docs/proposals/service-discovery/README.md
```diff
@@ -0,0 +1,171 @@
...
+**Scenario**:
+
+1. Given that the `Service` named `foo` exists on cluster member1
+2. And there is already a conflicting `Service` named foo on cluster member2
+3. And the `EndPointSlice` associated with the local `foo` service points to the local `Deployment` running on member2
+4. And the `ServiceImport` resource is created on cluster member2, specifying the import of `foo`
+5. When I attempt to access the service in cluster member2 using `foo.myspace.svc.cluster.local`
+6. Then the requests round-robin between the local `foo` service and the imported `foo` service (member1 and member2)
+7. And the imported `EndPointSlice` is created, pointing to the remote `Deployment` running on member1
+8. And the `EndPointSlice` associated with the imported service is prefixed with `import-`
```
XiShanYongYe-Chang, line 99: The 3rd, 4th, 7th, and 8th points are related to specific implementation details and do not need to be specified in the user story.

docs/proposals/service-discovery/README.md
```diff
@@ -0,0 +1,171 @@
...
+<!--
+This section should contain enough information that the specifics of your
+change are understandable. This may include API specs (though not always
+required) or even code snippets. If there's any ambiguity about HOW your
+proposal will be implemented, this is the place to discuss them.
+-->
+
+To achieve service discovery with native Kubernetes naming and resolution, the following design details should be considered:
+
+1. Service Import Controller Changes: The Service Import Controller needs to be updated to remove the `"derived-"` prefix from the derived service. This ensures that the imported service retains its original name when accessed within the importing cluster.
```
XiShanYongYe-Chang, line 133: Do we need to create another Service with the same name when there should already be a Service on the control panel?

XiShanYongYe-Chang (commented):
Thanks a lot~

zishen:
> 



> Hi @zishen , can you help take a look? cc @zishen

OK

docs/proposals/service-discovery/README.md
```diff
@@ -0,0 +1,171 @@
...
+
+2. Local-Only Service Discovery: To support local-only service discovery, Karmada should be enhanced to check if there is a local service with the same name as the imported service. If a local service exists, the remote service import should be skipped, and no `EndPointSlice` should be created for the remote service. This allows local services to be accessed directly without going through the remote service.
```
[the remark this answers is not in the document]
bivas: I think the most suitable location for that is the multi cluster services API with a Karmda annotation to control the behavior. Something like:
```
apiVersion: multicluster.x-k8s.io/v1alpha1
kind: ServiceImport
metadata:
  name: serve
  namespace: demo
  annotations:
    discovery.karmada.io/strategy: LocalAndRemote
spec:
  type: ClusterSetIP
  ports:
  - port: 80
    protocol: TCP
```

The default could be `LocalAndRemote`.

I don't think that the propagation policy spec can fit as it a more generic spec.

docs/proposals/service-discovery/README.md
```diff
@@ -0,0 +1,171 @@
...
+
+3. Local and Remote Service Discovery: When a service is imported from another cluster, both local and remote services should be accessible. Users accessing the service should be able to reach either the local or remote member based on their location and cluster routing. This enables high availability and load balancing between the local and remote instances of the service.
+
+4. Remote-Only Service Discovery: In cases where there is no local service with the same name as the imported service, Karmada should remove the local `EndPointSlice` and create an `EndPointSlice` pointing to the remote cluster. This allows users to access the service through the imported `EndPointSlice`, effectively reaching the remote cluster.
```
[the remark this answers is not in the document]
bivas: I think there's a typo here. Should read:
> Remote-Only Service Discovery: In cases where there is a local service with the same name as the imported service, Karmada should remove the local `EndPointSlice` and create an `EndPointSlice` pointing to the remote cluster.

The difference between RemoteOnly and LocalAndRemote is that the existence of a localized service and deployment  in both clusters. In RemoteOnly the cluster's local deployment should not be accessed and the service leads to a different cluster deployment.

docs/proposals/service-discovery/README.md
```diff
@@ -0,0 +1,171 @@
...
+<!--
+What are the caveats to the proposal?
+What are some important details that didn't come across above?
+Go in to as much detail as necessary here.
+This might be a good place to talk about core concepts and how they relate.
+-->
+
+### Risks and Mitigations
+
+<!--
```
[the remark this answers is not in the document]
bivas: Will update. The risk that I see with this proposal is the external cluster hop that could increase latency.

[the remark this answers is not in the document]
XiShanYongYe-Chang: +1, Cross-cluster access can indeed be slow.

docs/proposals/service-discovery/README.md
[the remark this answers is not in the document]
XiShanYongYe-Chang: > In RemoteOnly the cluster's local deployment should not be accessed and the service leads to a different cluster deployment.

It feels a bit difficult to achieve this. Based on the results of our previous discussion, can we temporarily set aside this approach?

docs/proposals/service-discovery/README.md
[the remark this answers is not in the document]
XiShanYongYe-Chang: It feels like a feasible approach. Can it be included in the proposal? 

Once the proposal is refined, would you be invited to share this proposal at the community meeting?

docs/proposals/service-discovery/README.md
```diff
@@ -0,0 +1,171 @@
...
+# Service discovery with native Kubernetes naming and resolution
+
+## Summary
+
+With the current `ServiceImportController` when a `ServiceImport` object is reconciled, the derived service is prefixed with `derived-` prefix.
+This seems counterintuitive when thinking about service discovery:
+- Assuming the pod is exported as the service `foo`
+- Another pod that wishes to access it on the same cluster will simply call `foo` and Kubernetes will bind to the correct one
+- If that pod is scheduled to another cluster, the original service discovery will fail as there's no service by the name `foo`
+- To find the original pod, the other pod is required to know it is in another cluster and use `derived-foo` to work properly
```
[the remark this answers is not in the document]
chaunceyjiang: By the way, personally I think it's best to use the following approach for multi-cluster services. 


The image is sourced from GKE.

image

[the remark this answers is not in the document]
chaunceyjiang: youtube video:


https://www.youtube.com/watch?v=MJ0LSvovKqk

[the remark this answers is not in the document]
bivas: @chaunceyjiang I actually wish to avoid using `clusterset` so the users will not have to know about `Service` placement and call it locally as with any other `Service` deployed locally

docs/proposals/service-discovery/README.md
bivas: When we import a `Service` from one cluster to another, the current solution will create `derived-` prefix for the imported `Service`. The issue in-hand is that both clusters have the same `Service` name (e.g. `foo`) - but we wish to use `ServiceImport` and that causes a conflict. 

Does the above better explain the situation?

docs/proposals/service-discovery/README.md
bivas: I believe we must create another `Service` by that name so that local DNS discovery will work. The PoC I tested locally simply ignored the conflict if a `Service` by that name already exist.

docs/proposals/service-discovery/README.md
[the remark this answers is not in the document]
bivas: We discussed this in 2 different threads 🤓 ([the other thread](https://github.com/karmada-io/karmada/pull/3694#discussion_r1241116184))
I agree it's an edge case. We should mention it, and if there's no consensus of it necessity - we should drop it.

docs/proposals/service-discovery/README.md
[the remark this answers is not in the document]
bivas: Yes, I would be happy to share it

docs/proposals/service-discovery/README.md
XiShanYongYe-Chang: Conflict refers to the situation where there is already a Service foo existing on the cluster, but we still need to import Service foo from other clusters onto this cluster. Am I understanding it correctly?

If the handling method mentioned in the proposal is used, will the conflict be resolved?

docs/proposals/service-discovery/README.md
XiShanYongYe-Chang: Ok, thx.
Can you help add more implementation details in the proposal so that it can help other guys confirm whether this solution can be ultimately implemented?

docs/proposals/service-discovery/README.md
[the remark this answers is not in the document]
XiShanYongYe-Chang: Ok, Thanks.

docs/proposals/service-discovery/README.md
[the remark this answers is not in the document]
XiShanYongYe-Chang: Here is the [Meeting Notes and Agenda](https://docs.google.com/document/d/1y6YLVC-v7cmVAdbjedoyR5WL0-q45DBRXTvz5_I7bkA/edit), please feel free to add the agenda with your time.

docs/proposals/service-discovery/README.md
bivas: Yes, you understood it perfectly. 
It will be resolved as we need to identify if that `Service` exists and ignore the conflict when importing a service from another cluster

docs/proposals/service-discovery/README.md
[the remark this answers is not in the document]
bivas: It seems that with my summer vacation coming up, I will be able to attend it only mid-September 😕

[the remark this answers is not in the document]
XiShanYongYe-Chang: It will be OK 😊

GitHubxsy:
We also need this feature in our products. How's this feature going now?

docs/proposals/service-discovery/README.md
RainbowMango: I basically agree with @XiShanYongYe-Chang that we might not touch the `local EndpointSlice` which is managed by the member cluster. 

I understand the case for `Remote Only` should be that:
There are no Pods on that cluster, but the cluster needs to consume the service. 
For example, a user might deploy DB services on a cluster, and allow other clusters to access the service.

RainbowMango:
Hi @bivas Thanks for sending the proposal, it's excellent!

Given that you are [on vocation,](https://github.com/karmada-io/karmada/pull/3694#discussion_r1265793370) I think we can move this PR forward, and @XiShanYongYe-Chang could continue to iterate it based on this version. Hope you don't mind.

RainbowMango (approved):
/lgtm
/approve

cc @XiShanYongYe-Chang

XiShanYongYe-Chang:
Ok,  I will continue to iterate it based on this version.
Hi @bivas Looking forward to discussing this proposal with you after you return from your vacation.

PR #3719, opened 2023-06-27
Cleanup redundant files in charts

Poor12:
**What type of PR is this?**
/kind cleanup

Cleanup redundant files in charts

Fixes #

None

### DIFF
--- a/charts/karmada/templates/post-install-job.yaml
+++ b/charts/karmada/templates/post-install-job.yaml
@@ -1,60 +1,6 @@
 {{- $name := include "karmada.name" . -}}
 {{- $namespace := include "karmada.namespace" . -}}
 {{- if eq .Values.installMode "host" }}
-{{- if eq .Values.certs.mode "custom" }}
-apiVersion: v1
-kind: ConfigMap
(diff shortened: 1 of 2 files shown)


RainbowMango:
Do you mean these files are never used and they are redundant with files in `pre-install-job.yaml`?
https://github.com/karmada-io/karmada/blob/master/charts/karmada/templates/pre-install-job.yaml#L91-L125

XiShanYongYe-Chang (commented):
The above cm `{{ $name }}-static-resources` should also be redundant.

charts/karmada/templates/post-install-job.yaml
```diff
@@ -2,59 +2,6 @@
 {{- $namespace := include "karmada.namespace" . -}}
 {{- if eq .Values.installMode "host" }}
 {{- if eq .Values.certs.mode "custom" }}
```
XiShanYongYe-Chang, line 4: This line also needs to be deleted.

Poor12:
/close 
In favor of #3721

PR #3725, opened 2023-06-27
add more samples to karmada operator

calvin0327:
**What type of PR is this?**
/kind feature

Add more samples for users to use karmada operator easily.
`simple` : the simplest karmada cr
`default`: the cr with defaults
`custom`: how to set labels, annotations, arguments and featureGates to karmada components
`addon`: how to enable karmada addon

Fixes #

```release-note
NONE
```


calvin0327:
@Poor12 @lonelyCZ /cc

Poor12:
I feel like such content is more suitable to be placed in the document.

calvin0327:
> I feel like such content is more suitable to be placed in the document.

should we add these examples to both the `Readme` and samples to make it easier for users to use?

calvin0327:
> I think `Readme` is enough. These are essentially different custom configurations.

Ok, I'm with you.

PR #3741, opened 2023-06-30
feat: optimize scheduling condition semantics

### LINKED ISSUE #3586: Confusing schedule result condition
I'm confused why the _ResourceBinding_'s condition is `BindingScheduled` when all clusters are filterd by filter plugins.
```
status:
  conditions:
  - lastTransitionTime: "2023-05-25T09:41:11Z"
    message: Failed to apply all works, see status.aggregatedStatus for details
    reason: FullyAppliedFailed
    status: "False"
    type: FullyApplied
  - lastTransitionTime: "2023-05-25T09:50:37Z"
    message: Binding has been scheduled
    reason: BindingScheduled
    status: "True"
    type: Scheduled
  schedulerObservedGeneration: 2
```
karmada-scheduler log:
```
I0525 09:50:37.081526       1 scheduler.go:336] Start to schedule ResourceBinding(default/test-pod) as scheduling type is duplicated
I0525 09:50:37.081536       1 scheduler.go:426] "Begin scheduling resource binding with ClusterAffinity" resourceBinding="default/test-pod"
I0525 09:50:37.081572       1 generic_scheduler.go:150] Cluster "member5715" is not fit, reason: cluster(s) didn't match the placement cluster affinity constraint
I0525 09:50:37.081582       1 scheduler.go:444] ResourceBinding(default/test-pod) scheduled to clusters []
I0525 09:50:37.081694       1 scheduler.go:450] "End scheduling resource binding with ClusterAffinity" resourceBinding="default/test-pod"
I0525 09:50:37.081741       1 scheduler.go:739] Begin to patch status condition to ResourceBinding(default/test-pod)
I0525 09:50:37.081839       1 event.go:294] "Event occurred" object="default/test-pod" fieldPath="" kind="ResourceBinding" apiVersion="work.karmada.io/v1alpha2" type="Warning" reason="ScheduleBindingFailed" message="0/1 clusters are available: 1 cluster(s) didn't match the placement cluster affinity constraint."
I0525 09:50:37.081859       1 event.go:294] "Event occurred" object="default/test" fieldPath="" kind="Pod" apiVersion="v1" type="Warning" reason="ScheduleBindingFailed" message="0/1 clusters are available: 1 cluster(s) didn't match the placement cluster affinity constraint."
```
Could someone explain to me the reason for this code design? /cc @RainbowMango @XiShanYongYe-Chang @Poor12 @chaunceyjiang 
image

I think the condition should be `BindingFailedScheduling` when all clusters are filterd by filter plugins. [The code](https://github.com/karmada-io/karmada/blob/master/pkg/scheduler/scheduler.go#LL452C19-L452C19) judgment here seems to be redundant.
**Environment**:
- Karmada version:
- Kubernetes version:
- Others:

whitewindmills:
**What type of PR is this?**
/kind feature

Fixes #3586

```release-note
karmada-scheduler: Introduce new schedule condition reasons, NoClusterFit, SchedulerError, Unschedulable, Success.
```

### DIFF
--- a/pkg/apis/work/v1alpha2/binding_types.go
+++ b/pkg/apis/work/v1alpha2/binding_types.go
@@ -307,6 +307,24 @@ const (
 	FullyApplied string = "FullyApplied"
 )
 
+// These are reasons for a binding's transition to a Scheduled condition.
+const (
+	// BindingReasonSuccess reason in Scheduled condition means that binding has been scheduled successfully.
+	BindingReasonSuccess = "Success"
+
+	// BindingReasonSchedulerError reason in Scheduled condition means that some internal error happens
+	// during scheduling, for example due to api-server connection error.
+	BindingReasonSchedulerError = "SchedulerError"
+
+	// BindingReasonNoClusterFit reason in Scheduled condition means that scheduling has finished
+	// due to no fit cluster.
+	BindingReasonNoClusterFit = "NoClusterFit"
+
+	// BindingReasonUnschedulable reason in Scheduled condition means that the scheduler can't schedule
+	// the binding right now, for example due to insufficient resources in the clusters.
+	BindingReasonUnschedulable = "Unschedulable"
+)
+
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
 // ResourceBindingList contains a list of ResourceBinding.
--- a/pkg/scheduler/helper.go
+++ b/pkg/scheduler/helper.go
@@ -2,11 +2,17 @@ package scheduler
 
 import (
 	"encoding/json"
+	"errors"
 	"reflect"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/klog/v2"
 
 	policyv1alpha1 "github.com/karmada-io/karmada/pkg/apis/policy/v1alpha1"
+	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
+	"github.com/karmada-io/karmada/pkg/scheduler/framework"
+	"github.com/karmada-io/karmada/pkg/util"
 )
 
 func placementChanged(
@@ -83,4 +89,34 @@ func getAffinityIndex(affinities []policyv1alpha1.ClusterAffinityTerm, observedN
 		}
 	}
 	return 0
+}
+
+// getConditionByError returns condition by error type, bool to indicate if ignore this error.
+func getConditionByError(err error) (metav1.Condition, bool) {
+	if err == nil {
+		return util.NewCondition(workv1alpha2.Scheduled, workv1alpha2.BindingReasonSuccess, successfulSchedulingMessage, metav1.ConditionTrue), true
+	}
+
+	var unschedulableErr *framework.UnschedulableError
+	if errors.As(err, &unschedulableErr) {
+		return util.NewCondition(workv1alpha2.Scheduled, workv1alpha2.BindingReasonUnschedulable, err.Error(), metav1.ConditionFalse), false
+	}
+
+	fitErrMatcher := func(e error) bool {
+		var fitErr *framework.FitError
+		return errors.As(e, &fitErr)
+	}
+	if fitErrMatcher(err) {
+		return util.NewCondition(workv1alpha2.Scheduled, workv1alpha2.BindingReasonNoClusterFit, err.Error(), metav1.ConditionFalse), true
+	}
+	var aggregatedErr utilerrors.Aggregate
+	if errors.As(err, &aggregatedErr) {
+		for _, ae := range aggregatedErr.Errors() {
+			if fitErrMatcher(ae) {
+				// if aggregated NoClusterFit error got, we do not ignore error but retry scheduling.
+				return util.NewCondition(workv1alpha2.Scheduled, workv1alpha2.BindingReasonNoClusterFit, err.Error(), metav1.ConditionFalse), false
+			}
+		}
+	}
+	return util.NewCondition(workv1alpha2.Scheduled, workv1alpha2.BindingReasonSchedulerError, err.Error(), metav1.ConditionFalse), false
 }
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -54,14 +54,13 @@ const (
 )
 
 const (
-	scheduleSuccessReason  = "BindingScheduled"
-	scheduleFailedReason   = "BindingFailedScheduling"
-	scheduleSuccessMessage = "Binding has been scheduled"
+	// DefaultScheduler defines the name of default scheduler.
+	DefaultScheduler = "default-scheduler"
 )
 
 const (
-	// DefaultScheduler defines the name of default scheduler.
-	DefaultScheduler = "default-scheduler"
+	// successfulSchedulingMessage defines the successful binding event message.
+	successfulSchedulingMessage = "Binding has been scheduled successfully."
 )
 
 // Scheduler is the scheduler schema, which is used to schedule a specific resource to specific clusters
@@ -415,18 +414,14 @@ func (s *Scheduler) doScheduleClusterBinding(name string) (err error) {
 
 func (s *Scheduler) scheduleResourceBinding(rb *workv1alpha2.ResourceBinding) (err error) {
 	defer func() {
-		var condition metav1.Condition
-		if err == nil {
-			condition = util.NewCondition(workv1alpha2.Scheduled, scheduleSuccessReason, scheduleSuccessMessage, metav1.ConditionTrue)
-		} else {
-			condition = util.NewCondition(workv1alpha2.Scheduled, scheduleFailedReason, err.Error(), metav1.ConditionFalse)
-		}
+		condition, ignoreErr := getConditionByError(err)
 		if updateErr := patchBindingStatusCondition(s.KarmadaClient, rb, condition); updateErr != nil {
+			// if patch error occurs, just return patch error to reconcile again.
+			err = updateErr
 			klog.Errorf("Failed to patch schedule status to ResourceBinding(%s/%s): %v", rb.Namespace, rb.Name, err)
-			if err == nil {
-				// schedule succeed but update status failed, return err in order to retry in next loop.
-				err = updateErr
-			}
+		} else if ignoreErr && err != nil {
+			// for finished schedule, we won't retry.
+			err = nil
 		}
 	}()
 
@@ -447,9 +442,9 @@ func (s *Scheduler) scheduleResourceBindingWithClusterAffinity(rb *workv1alpha2.
 	}
 
 	scheduleResult, err := s.Algorithm.Schedule(context.TODO(), &rb.Spec, &rb.Status, &core.ScheduleAlgorithmOption{EnableEmptyWorkloadPropagation: s.enableEmptyWorkloadPropagation})
-	var noClusterFit *framework.FitError
-	// in case of no cluster fit, can not return but continue to patch(cleanup) the result.
-	if err != nil && !errors.As(err, &noClusterFit) {
+	var fitErr *framework.FitError
+	// in case of no cluster error, can not return but continue to patch(cleanup) the result.
+	if err != nil && !errors.As(err, &fitErr) {
 		s.recordScheduleResultEventForResourceBinding(rb, err)
 		klog.Errorf("Failed scheduling ResourceBinding(%s/%s): %v", rb.Namespace, rb.Name, err)
 		return err
@@ -457,11 +452,11 @@ func (s *Scheduler) scheduleResourceBindingWithClusterAffinity(rb *workv1alpha2.
 
 	klog.V(4).Infof("ResourceBinding(%s/%s) scheduled to clusters %v", rb.Namespace, rb.Name, scheduleResult.SuggestedClusters)
 	patchErr := s.patchScheduleResultForResourceBinding(rb, string(placementBytes), scheduleResult.SuggestedClusters)
-	s.recordScheduleResultEventForResourceBinding(rb, utilerrors.NewAggregate([]error{err, patchErr}))
-
-	// only care about the patch result,
-	// for FitError already recorded by event
-	return patchErr
+	if patchErr != nil {
+		err = utilerrors.NewAggregate([]error{err, patchErr})
+	}
+	s.recordScheduleResultEventForResourceBinding(rb, err)
+	return err
 }
 
 func (s *Scheduler) scheduleResourceBindingWithClusterAffinities(rb *workv1alpha2.ResourceBinding) error {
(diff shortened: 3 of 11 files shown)


pkg/scheduler/helper.go
```diff
@@ -84,3 +89,16 @@ func getAffinityIndex(affinities []policyv1alpha1.ClusterAffinityTerm, observedN
 	}
 	return 0
 }
+
+func createConditionByError(err error) metav1.Condition {
+	if err == nil {
+		return util.NewCondition(workv1alpha2.Scheduled, scheduleSuccessReason, scheduleSuccessMessage, metav1.ConditionTrue)
+	}
+
+	var noClusterErr *core.NoClusterError
```
chaunceyjiang, line 98:
```suggestion
	var noClusterErr core.NoClusterError
```

RainbowMango (commented):
/assign

pkg/scheduler/core/util.go
```diff
@@ -15,8 +15,29 @@ import (
 	"github.com/karmada-io/karmada/pkg/util"
 )
 
+const (
+	NoClusterAvailableReason = "NoClusterAvailable"
+	NoClusterFitReason       = "NoClusterFit"
```
jwcesign, line 20: Why not put these in file `pkg/scheduler/scheduler.go` L59?

whitewindmills: If placed in file `pkg/scheduler/scheduler.go`, package `scheduler` will be imported in package `core`, which is a circular dependency.

jwcesign: I prefer to keep this information in the same place. So how about moving the related variable in `pkg/scheduler/scheduler.go` here?

whitewindmills: I will find a solution.

whitewindmills:
PTAL @XiShanYongYe-Chang @RainbowMango @jwcesign

jwcesign:
LGTM

XiShanYongYe-Chang (commented):
Thanks a lot~
/lgtm

Ask @Garrybest to help take a look.
/cc @Garrybest

XiShanYongYe-Chang:
Hi @whitewindmills, can you help update the release note more clearly, for example, add the new Reason name? It will be more intuitive when the version is released.

whitewindmills:
> Hi @whitewindmills, can you help update the release note more clearly, for example, add the new Reason name? It will be more intuitive when the version is released.

sure

pkg/scheduler/scheduler.go
```diff
@@ -54,9 +53,11 @@ const (
 )
 
 const (
-	scheduleSuccessReason  = "BindingScheduled"
-	scheduleFailedReason   = "BindingFailedScheduling"
-	scheduleSuccessMessage = "Binding has been scheduled"
+	scheduleSuccessReason    = "BindingScheduled"
+	scheduleFailedReason     = "BindingFailedScheduling"
+	noClusterAvailableReason = "NoClusterAvailable"
```
Garrybest, line 58: This two types is confusing, what's the difference?

I'd like to refer to what kubernetes does.

1. We can't find any clusters that pass the filter plugins.
```
    message: 0/1 clusters are available: 1 cluster(s) didn't match the placement cluster affinity constraint.
    reason: Unschedulable
    status: "False"
    type: BindingScheduled
```

2. we can't assign replicas due to not enough resources.

```
    message: failed to assignReplicas: xxxxxx
    reason: Unschedulable
    status: "False"
    type: BindingScheduled
```

This two types will cover the unschedulable condition.

whitewindmills: `NoCluster` is for the scenarios without clusters. As discussed before, we need to distinguish between scenarios with and without clusters. For scenarios without clusters, retry scheduling actually does not make any sense, but from our requirements, we need to regard it as a successful scheduling. If we introduce `Unschedulable`, how to define it?

Garrybest: Can you enlighten me about what is `without clusters`? Do you mean no clusters pass the filter plugin?

whitewindmills: yeah, and no member cluster

Garrybest: >  For scenarios without clusters, retry scheduling actually does not make any sense

What if a new cluster is added, or a taint is removed, should we retry?

I don't think this is a successful scheduling. For kubernetes, a pod is not scheduled successfully unless it is truely assigned to a node. We can easily distinguish no-cluster availablity and scheduler internal error by `reason`. `Unschedulable` is a typical reason for no more clusters fit or no more enough resources.

Garrybest: cc @XiShanYongYe-Chang @chaunceyjiang for opinions.

whitewindmills: > What if a new cluster is added, or a taint is removed, should we retry?

yeah, will retry scheduling driven by cluster events.

> Unschedulable is a typical reason for no more clusters fit or no more enough resources.

Sure, no more enough resources is failed scheduling, it is essentially different from the reason for no cluster. Just like insufficient CPU is also no more enough resources, however CPU resources are dynamically changing. When CPU resources return to normal, the scheduling can be successfully by retrying the scheduling. But in a scenario without any clusters, it will never be possible to heal itself without human intervention.

Garrybest: I only see two kind of errors for NoClusterError.

1. No member clusters.
2. No fit clusters.

Actually `No member clusters` is a special kind of `No fit clusters`. So basically they belong together. I guess you are concerned about the retrying because we don't need to retry when there are no member clusters. This issue should not confuse with conditions and I think this is the duty for **scheduling queue**.

Now let's figure out what **successful scheduling** means. In analogy to this definition for Kubernetes, I could give a definition like this:

The binding.Spec.Clusters are updated by scheduler when a binding is passed all steps of scheduler. Otherwise it fails and we do have different kind of reasons or conditions. **Correct me if I'm wrong here.**

1. `BindingReasonUnschedulable = "SchedulerError"`, means that some internal error happens during scheduling, for example due to nodeAffinity parsing errors.
2. `PodReasonUnschedulable = "Unschedulable"`, means that the scheduler can't schedule the binding right now, for example due to no candidate cluters or insufficient resources in candidate clusters.

This two binding scheduling failure reasons are the same as kube-scheduler. So I don't know if there are more reasons here or do you want to seperate this `Unschedulable` reason with more precise ones?

XiShanYongYe-Chang: I remember the conclusion of our previous community meeting was to add new reason to distinguish between errors during scheduling and no suitable cluster after scheduling. Hi @Garrybest, do you mean to use `BindingFailedSchedulingreason` to indicate all scheduling failures, and then use meassage to distinguish the above situations?

Garrybest: Emmm, not really. `BindingScheduled` is the type. And `SchedulerError`, `Unschedulable` is the reasons. Note that we distinguish failure and success by `status`. 

1. Success.
```
    status: "True"
    type: BindingScheduled
```
2. Internal Error
```
    message: internal error: xxxx
    reason: SchedulerError
    status: "False"
    type: BindingScheduled
```
3. Unschedulable
```
    message: 0/1 clusters are available: 1 cluster(s) didn't match the placement cluster affinity constraint.
    reason: Unschedulable
    status: "False"
    type: BindingScheduled
```

whitewindmills: @Garrybest thanks,  it's a good suggestion, but I really want to distinguish between the internal error and the error for no suitable cluster. If it is designed like this, I can only distinguish it by message, which will be very bad.

Garrybest: Got it. But I guess `no suitable cluster` is a Unschedulable one and `internal error` is a `SchedulerError` one, isn't it?

whitewindmills: of course you are right. can we seperate the `Unschedulable` reason into more precise ones? maybe we don't need `Unschedulable` but seperate it into specific reasons, like `NoClusterAvailable`, `NoClusterFit` and `InsufficientResources` etc. And the internal error is still `SchedulerError`. how do you feel about this?

Garrybest: I'm ok with it. But I'm not sure the difference between NoClusterAvailable and NoClusterFit. Can we merge them?

whitewindmills: sure, we can name it `NoClusterFit`. but for `NoClusterFit`, I think we do not need to return error to retry scheduling, cause it makes no sense and resources will be scheduled when new member cluster is joined or scheduling requirements are met.
Let's summarize, it looks like this:
1.  Unschedulable: no cluster fit
```
message: no clusters available to schedule
reason: NoClusterFit
status: "False"
type: Scheduled
```

2.  Unschedulable: insufficient resources 
```
message: cluster xxx has no sufficient memory
reason: InsufficientResources
status: "False"
type: Scheduled
```

3.  Internal error: eg. can not access api-server...
```
message: xxx timeout
reason: SchedulerError
status: "False"
type: Scheduled
```

4.  schedule successfully
```
message: Binding has been scheduled
reason: BindingScheduled
status: "True"
type: Scheduled
```

Garrybest: Great! Now we have reached an agreement.

XiShanYongYe-Chang:
Ask an again review from @Garrybest

pkg/apis/work/v1alpha2/binding_types.go
```diff
@@ -307,6 +307,24 @@ const (
 	FullyApplied string = "FullyApplied"
 )
 
+// These are reasons for a binding's transition to a Scheduled condition.
+const (
+	// BindingReasonSuccess reason in Scheduled condition means that binding has been scheduled successfully.
+	BindingReasonSuccess = "Success"
```
Garrybest, line 313: Actually, I don't think we need a reason for successful condition. If a pod is scheduled to a node successfully, reason will not be set as well.

whitewindmills:
I slightly adjusted the results of our previous discussions. I think `NoClusterFit` should be separated from `Unschedulable` due to its particularity. For `NoClusterFit`, we do not need to return error to retry scheduling. Now it looks like the following.
```
    message: Binding has been scheduled successfully.
    reason: Success
    status: "True"
    type: Scheduled
```

```
    message: 0/3 clusters are available: 3 cluster(s) didn't match the placement
      cluster affinity constraint.
    reason: NoClusterFit
    status: "False"
    type: Scheduled
```

```
    message: Clusters available replicas 3 are not enough to schedule.
    reason: Unschedulable
    status: "False"
    type: Scheduled
```

```
    message: failed to select clusters.
    reason: SchedulerError
    status: "False"
    type: Scheduled
```

/cc @Garrybest @XiShanYongYe-Chang

pkg/scheduler/scheduler.go
```diff
@@ -54,14 +54,13 @@ const (
...
-	scheduleSuccessMessage = "Binding has been scheduled"
+	// DefaultScheduler defines the name of default scheduler.
+	DefaultScheduler = "default-scheduler"
 )
 
 const (
-	// DefaultScheduler defines the name of default scheduler.
-	DefaultScheduler = "default-scheduler"
+	// successfulSchedulingMessage defines the successful binding event message.
+	successfulSchedulingMessage = "Binding has been scheduled successfully."
```
Garrybest, line 63: This msg is not required as well.

pkg/scheduler/core/util.go
```diff
@@ -17,6 +17,14 @@ import (
 
 type calculator func([]*clusterv1alpha1.Cluster, *workv1alpha2.ResourceBindingSpec) []workv1alpha2.TargetCluster
 
+type UnschedulableError struct {
```
Garrybest, line 20: It's better to put this struct to types.go

pkg/scheduler/helper.go
```diff
@@ -84,3 +90,20 @@ func getAffinityIndex(affinities []policyv1alpha1.ClusterAffinityTerm, observedN
...
+
+// getConditionByError returns condition by error type, bool to indicate if ignore this error.
+func getConditionByError(err error) (metav1.Condition, bool) {
+	if err == nil {
+		return util.NewCondition(workv1alpha2.Scheduled, workv1alpha2.BindingReasonSuccess, successfulSchedulingMessage, metav1.ConditionTrue), true
+	}
+
+	var fitErr *framework.FitError
+	if errors.As(err, &fitErr) {
+		return util.NewCondition(workv1alpha2.Scheduled, workv1alpha2.BindingReasonNoClusterFit, err.Error(), metav1.ConditionFalse), true
```
Garrybest, line 102: nit: if return true, scheduler will not retry by exponential backoff algorithm which could be unexpected from before.

pkg/scheduler/scheduler.go
```diff
@@ -458,10 +453,11 @@ func (s *Scheduler) scheduleResourceBindingWithClusterAffinity(rb *workv1alpha2.
 	klog.V(4).Infof("ResourceBinding(%s/%s) scheduled to clusters %v", rb.Namespace, rb.Name, scheduleResult.SuggestedClusters)
 	patchErr := s.patchScheduleResultForResourceBinding(rb, string(placementBytes), scheduleResult.SuggestedClusters)
```
Garrybest, line 454: What if aggregate patchErr and error first, and return this aggregation one?

pkg/apis/work/v1alpha2/binding_types.go
whitewindmills: > Actually, I don't think we need a reason for successful condition. If a pod is scheduled to a node successfully, reason will not be set as well.

I agree with you but it's required. BTW PodCondition overrides this field and marks it as optional.
image

pkg/scheduler/scheduler.go
whitewindmills: ditto

pkg/scheduler/helper.go
whitewindmills: no, I just keep it same as before.
https://github.com/karmada-io/karmada/blob/8388314fbe80ff3bcbf72c2223dd3ba8f167cf50/pkg/scheduler/scheduler.go#L452-L464

pkg/scheduler/scheduler.go
whitewindmills: we do not care about what type of error to retry scheduling, but for `NoClusterFit`, we cannot judge `FitError` by `errors.As` function if the aggregated error for `patchErr` and `fitErr` is returnd.

pkg/scheduler/core/util.go
whitewindmills: which `types.go`?

pkg/scheduler/scheduler.go
Garrybest:
```go
patchErr := s.patchScheduleResultForResourceBinding(rb, string(placementBytes), scheduleResult.SuggestedClusters)
err = utilerrors.NewAggregate([]error{err, patchErr})
s.recordScheduleResultEventForResourceBinding(rb, err)
return err
```

whitewindmills:
PTAL~ @Garrybest

Garrybest:
Good job.

/lgtm
/approve

XiShanYongYe-Chang:
Good job!
Thanks~

PR #3750, opened 2023-07-03
karmada operator: change karmada system namespace

calvin0327:
**What type of PR is this?**
/kind feature

Change the karmada controlplane system namespace to `karmada-system`.  keep the same behavior with chart and karmadactl. 

Fixes #

```release-note
NONE
```


Poor12:
Does this also mean that the namespace is not configurable? Must be `karmada-system`?

calvin0327:
> Does this also mean that the namespace is not configurable? Must be `karmada-system`?

@Poor12  Indeed, the `karmadaSystemNamespace` should be a configurable variable. I dont's know where is the best place to configure it, karmada operator flag?  or  karmada CR?

Poor12:
From my perspective, I think the namespace of components should be the same with the namespace of Karmada CR.

calvin0327:
> From my perspective, I think the namespace of components should be the same with the namespace of Karmada CR.

@Poor12  If so, the system namespace  of karmada instance seems to be different.

Poor12:
Yeah, in that way, the namespace of default Karmada CR should be `karmada-system`. And I don‘t think it's possible for Karmada's components to exist in multiple namespaces at the same time.

calvin0327:
In my scenario, we create a karmada named `e2etest-instancedefaulto6z` and in namespace `e2etest-instancedefaulto6z`.

```shell
[root@master ~]# kubectl get karmada -A
NAMESPACE                    NAME                         READY   AGE
e2etest-instancedefaulto6z   e2etest-instancedefaulto6z   True    29m
```

And we access the karmada controlplane:

image

It is same with karmada cr of namespace, so, the system namespace of every karmada instance is different.
So, I think it's better to keep all karmada instance system namespace the same.

@Poor12 @lonelyCZ

calvin0327:
@RainbowMango @jwcesign Do you have any comments?

operator/pkg/tasks/init/karmadaresource.go
```diff
@@ -68,7 +68,7 @@ func runSystemNamespace(r workflow.RunData) error {
 
 	err := apiclient.CreateNamespace(data.KarmadaClient(), &corev1.Namespace{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: data.GetNamespace(),
+			Name: constants.KarmadaSystemNamespace,
```
RainbowMango, line 71: The following two places still referencing the names with `data.GetNamespace`, I guess should be updated accordingly, right?
https://github.com/karmada-io/karmada/blob/8d1a8007ebdbd97a2fe12cf8ee7d4694cfb2ed82/operator/pkg/tasks/init/karmadaresource.go#L75

https://github.com/karmada-io/karmada/blob/8d1a8007ebdbd97a2fe12cf8ee7d4694cfb2ed82/operator/pkg/tasks/init/karmadaresource.go#L78

calvin0327: Yes, that's right. the namespace message of these logs should be consistent. 
updated, thanks.

RainbowMango (approved):
/lgtm
/approve

Issue #3752, opened 2023-07-04
GitHub Actions: Deprecating save-state and set-output commands


RainbowMango:
**What would you like to be added**:

Update [set-output](https://github.com/karmada-io/karmada/blob/3909fccf7d369160daba2188af11b44a4c8d390d/.github/workflows/lint-chart.yaml#L52) to the new `GITHUB_STATE` environment.

**Why is this needed**:
Got the warning that [--set-output](https://github.com/karmada-io/karmada/blob/3909fccf7d369160daba2188af11b44a4c8d390d/.github/workflows/lint-chart.yaml#L52) command is deprecated and will be disabled soon.

For more information see:
1. The warning: https://github.com/karmada-io/karmada/actions/runs/5449659220
2. The deprecation: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

cc @calvin0327

calvin0327:
/assign

I'll check out the latest usage

PR #3757, opened 2023-07-04
Update set-output command in lint-chart ci

calvin0327:
**What type of PR is this?**
/kind feature

The `set-output` is be deprecated，we need set the variable in step to `$GITHUB_OUTPUT` env.

Ref: 
https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#example-of-setting-an-output-parameter
https://github.com/helm/chart-testing-action#example-workflow

Fixes https://github.com/karmada-io/karmada/issues/3752

```release-note
NONE
```

### DIFF
--- a/.github/workflows/lint-chart.yaml
+++ b/.github/workflows/lint-chart.yaml
@@ -2,10 +2,10 @@
 name: Chart Lint
 
 env:
-  HELM_VERSION: v3.8.1
+  HELM_VERSION: v3.11.2
   KIND_VERSION: v0.14.0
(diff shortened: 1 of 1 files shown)


calvin0327:
> Could you test the `chart-lint` in your own repo since it will not run in the master branch?

Ok, I will test this change.

calvin0327:
https://github.com/calvin0327/karmada/actions/runs/5453894666/jobs/9923303876

image

Because the chart is not any change, so, the  job will skip the three steps.

RainbowMango (approved):
/approve

Issue #3774, opened 2023-07-06
The karmada-serach helm chart  code without the volumes of etcd certs causes pod failed


watermeion:
**What happened**:

The deployment of karmada-search generated by helm chart  is without  volumes and volumeMounts of /etc/etcd. This causes pod start error in v1.6.0. 
It is definitely missing In helm chart codes.

RainbowMango:
Hi @watermeion Thanks for spotting.
Can you explain how to reproduce it? 
cc @Poor12 @calvin0327 to take a look.

watermeion:
Easy to reproduce by exec helm install using etcd external mode.
Or You can dig into `charts/karmada/templates/karmada-search.yaml` file,  there are no volumeMounts with path `/etc/etcd/pki ....` , but using in container args . This will cause the container exit with no such cert files in order to connect etcd

Poor12:
I think it's definitely a bug. cc contributor @my-git9 to check.

calvin0327:
https://github.com/karmada-io/karmada/blob/master/charts/karmada/templates/karmada-search.yaml#L89

Indeed, this is a bug in external mode, the search not to mount the external certs of etcd, I think this a easy to fix bug reference the karmada apiserver yaml:  
https://github.com/karmada-io/karmada/blob/master/charts/karmada/templates/karmada-apiserver.yaml#L140

my-git9:
Sorry for the omission, I will try fix it in https://github.com/karmada-io/karmada/pull/3777/files

RainbowMango:
/assign @my-git9 
Thanks for help.

PR #3777, opened 2023-07-06
chart: fix karmada search external etcd no volume mount

my-git9:
**What type of PR is this?**
/kind bug

Fixes https://github.com/karmada-io/karmada/issues/3774

```release-note
Chart: Fixed the issue that `karmada-search` no ETCD secret volume mount when using external ETCD.
```

### DIFF
--- a/charts/karmada/templates/_helpers.tpl
+++ b/charts/karmada/templates/_helpers.tpl
@@ -273,6 +273,14 @@ app: {{- include "karmada.name" .}}-search
   secret:
     secretName: {{ .Values.search.kubeconfig }}
 {{- end -}}
+- name: etcd-certs
+  secret:
+  {{- if eq .Values.etcd.mode "internal" }}
+    secretName: {{ $name }}-cert
+  {{- end }}
+  {{- if eq .Values.etcd.mode "external" }}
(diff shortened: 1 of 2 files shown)


calvin0327:
https://github.com/karmada-io/karmada/blob/master/charts/karmada/templates/karmada-apiserver.yaml#L140

Whether the karmada apiserver can use the same temlate `karmada.search.kubeconfig.volume` ?

my-git9:
> karmada.search.kubeconfig.volume

It seems to be ok, but there are more parameter volumes included in the search, maybe it would be better to customize a parameter like `karmada.apiserver.volume` ?

calvin0327:
> > karmada.search.kubeconfig.volume
> 
> It seems to be ok, but there are more parameter volumes included in the search, maybe it would be better to customize a parameter like `karmada.apiserver.volume` ?

Ok, I got it.

RainbowMango:
@watermeion can you help to verify this patch?

You can fetch this patch like:
```
git fetch upstream pull/3777/head:pr3777
git checkout pr3777
```

RainbowMango (approved):
/approve

Issue #3746, opened 2023-07-03
feat: Support CronFederatedHPA to scale workload at a specific time


jwcesign:
**What would you like to be added**:
Support CronFederatedHPA to scale workload across clusters at a specific time.

**Why is this needed**:
For some certain predictable load change scenarios, for example, the Black Friday sale and Singles' Day sale, there will be sharp load peaks. With standard K8s HPA, it needs time to scale up the replicas, which may not be fast enough to handle sharp load peaks, resulting in service unavailability.

To solve this problem, there are already some approaches in a single cluster, called `CronHPA`, which can scale up/down the workloads at a specific time. Cluster administrators can scale up the workloads in advance with `CronHPA`, to meet the requirements of the sharp load peaks. So this proposal proposes `CronFederatedHPA` to meet the same requirements in the multi-cluster scenario.

There are following tasks:
- [x] Give a proposal about the design details of CronFederatedHPA. (https://github.com/karmada-io/karmada/pull/3664)
- [x] Defined the API of CronFederatedHPA. (https://github.com/karmada-io/karmada/pull/3692)
- [x] Implement CronFederatedHPA controller to scale the workloads. (https://github.com/karmada-io/karmada/pull/3729)
- [x] Add e2e cases to ensure the feature is always ready. (@jwcesign )
- [x] Support init validation configuration of CronFederatedHPA via karmadactl (@chaunceyjiang).
- [x] Add tutorials for CronFederatedHPA. (@jwcesign)

chaunceyjiang:
> Support init validation configuration of CronFederatedHPA via karmadactl.

Please assign this task to me.

jwcesign:
> Please assign this task to me.

Done

RainbowMango:
/close
as all tasks have been resolved.

PR #3791, opened 2023-07-13
e2e: Add e2e for CronFederatedHPA

jwcesign:
**What type of PR is this?**
/kind cleanup

Add e2e for CronFederatedHPA.
case 1:Scale FederatedHPA.
case 2:Scale deployment.
case 3:Test suspend rule with CronFederatedHPA

Part of https://github.com/karmada-io/karmada/issues/3746

none

```release-note

```

### DIFF
--- a/test/e2e/cronfederatedhpa_test.go
+++ b/test/e2e/cronfederatedhpa_test.go
@@ -1,0 +1,199 @@
+/*
+Copyright 2023 The Karmada Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package e2e
+
+import (
+	"time"
+
+	"github.com/onsi/ginkgo/v2"
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/util/rand"
+	"k8s.io/utils/pointer"
+
+	autoscalingv1alpha1 "github.com/karmada-io/karmada/pkg/apis/autoscaling/v1alpha1"
+	policyv1alpha1 "github.com/karmada-io/karmada/pkg/apis/policy/v1alpha1"
+	"github.com/karmada-io/karmada/test/e2e/framework"
+	"github.com/karmada-io/karmada/test/helper"
+)
+
+/*
+CronFederatedHPA focus on scaling FederatedHPA or other resource with scale subresource (e.g. Deployment, StatefulSet).
+Test Case Overview:
+
+	case 1:
+		Scale FederatedHPA.
+	case 2:
+		Scale deployment.
+	case 3:
+		Test suspend rule in CronFederatedHPA
+	case 4:
+		Test unsuspend rule then suspend it in CronFederatedHPA
+*/
+var _ = ginkgo.Describe("[CronFederatedHPA] CronFederatedHPA testing", func() {
+	var cronFHPAName, fhpaName, policyName, deploymentName string
+	var cronFHPA *autoscalingv1alpha1.CronFederatedHPA
+	var fhpa *autoscalingv1alpha1.FederatedHPA
+	var deployment *appsv1.Deployment
+	var policy *policyv1alpha1.PropagationPolicy
+
+	ginkgo.BeforeEach(func() {
+		cronFHPAName = cronFedratedHPANamePrefix + rand.String(RandomStrLength)
+		fhpaName = federatedHPANamePrefix + rand.String(RandomStrLength)
+		policyName = deploymentNamePrefix + rand.String(RandomStrLength)
+		deploymentName = policyName
+
+		deployment = helper.NewDeployment(testNamespace, deploymentName)
+		policy = helper.NewPropagationPolicy(testNamespace, policyName, []policyv1alpha1.ResourceSelector{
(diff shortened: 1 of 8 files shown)


chaunceyjiang (commented):
/assign

test/e2e/cronfederatedhpa_test.go
```diff
@@ -0,0 +1,154 @@
...
+			framework.CreateFederatedHPA(karmadaClient, fhpa)
+		})
+
+		ginkgo.AfterEach(func() {
+			framework.RemoveFederatedHPA(karmadaClient, testNamespace, fhpaName)
+			framework.RemoveCronFederatedHPA(karmadaClient, testNamespace, cronFHPAName)
+		})
+
+		ginkgo.It("Scale FederatedHPA testing", func() {
+			// Wait FederatedHPA to scale deployment's replicas to 1
```
chaunceyjiang, line 97: How do you ensure that  the replicas of the deployment can scale-down to 1 when I see that you have set MinReplicas to 1,  MaxReplicas to 10 and AverageUtilization to 80 for FederatedHPA? 


Is there such a situation where `nginx` may consume some CPU, causing the replicas to only scale-down to 2?

jwcesign: We do not request nginx pods, I think scaling down the replicas of nginx is certain.

chaunceyjiang: How about setting  MinReplicas to 1 and MaxReplicas to 1 ?

jwcesign: It's better idea.

test/e2e/cronfederatedhpa_test.go
```diff
@@ -0,0 +1,154 @@
...
+			cronFHPA = helper.NewCronFederatedHPAWithScalingDeployment(testNamespace, cronFHPAName, deploymentName, rule)
+		})
+
+		ginkgo.AfterEach(func() {
+			framework.RemoveCronFederatedHPA(karmadaClient, testNamespace, cronFHPAName)
+		})
+
+		ginkgo.It("Scale Deployment testing", func() {
+			// The init replicas of deployment should be 3
+			framework.WaitDeploymentReplicasFitWith(framework.ClusterNames(), testNamespace, deploymentName, 3)
```
chaunceyjiang, line 121:
```suggestion
			framework.WaitDeploymentReplicasFitWith(framework.ClusterNames(), testNamespace, deploymentName, int(*deployment.Spec.Replicas))
```

test/e2e/cronfederatedhpa_test.go
```diff
@@ -0,0 +1,154 @@
...
+
+		ginkgo.It("Test suspend rule with CronFederatedHPA", func() {
+			// The init replicas of deployment should be 3
+			framework.WaitDeploymentReplicasFitWith(framework.ClusterNames(), testNamespace, deploymentName, 3)
+
+			// Create CronFederatedHPA to scale Deployment
+			framework.CreateCronFederatedHPA(karmadaClient, cronFHPA)
+
+			// Wait 3 minutes and check whether the replicas changed
+			time.Sleep(time.Minute*1 + time.Second*30)
```
chaunceyjiang, line 150:
```suggestion
			time.Sleep(pollTimeout / 2)
```

jwcesign: Why? The CronFederatedHPA is supposed to execute every minute. I will wait for 1:30 and check if the replicas have changed.

chaunceyjiang: I understand, but `(time.Minute*1 + time.Second*30)` is a magic number, so I suggest using pollTimeout instead.

RainbowMango: By the way,
wait `3 minutes`?  Waiting time seems too long which will seriously slow down the test progress.

jwcesign: We should wait for this CronFederatredHPA rule, and check whether it's suspended, I didn't find other way to solve this.

jwcesign: `pollTimeout = 420 * time.Second` is too long, I add some comments to make it clearly:
```
// */1 * * * * means the rule will be triggered every 1 minute
// So wait for 1m30s and check whether the replicas changed and whether the suspend field works
```

test/e2e/cronfederatedhpa_test.go
```diff
@@ -0,0 +1,159 @@
...
+/*
+CronFederatedHPA focus on scaling FederatedHPA or other resource with scale subresource (e.g. Deployment, StatefulSet).
+Test Case Overview:
+
+	case 1:
+		Scale FederatedHPA.
+	case 2:
+		Scale deployment.
+	case 3:
+		Test suspend rule with CronFederatedHPA
```
chaunceyjiang, line 39: I think we need to add another case:  **modify** the existing rule("suspend=false") to "suspend=true".

jwcesign:
/cc @chaunceyjiang @XiShanYongYe-Chang 
PTAL

chaunceyjiang (commented):
LGTM

jwcesign:
/cc @ikaven1024
Please take a look at the script updates.

test/e2e/cronfederatedhpa_test.go
RainbowMango: @chaunceyjiang What do you think?
I didn't look into it.

chaunceyjiang: I also didn't find other way to solve this.

chaunceyjiang (commented):
/lgtm

test/e2e/cronfederatedhpa_test.go
XiShanYongYe-Chang: Wait until can be a slightly more elegant way to modify it. Although fundamentally still need to wait for a little over a minute.

PR #3792, opened 2023-07-13
update some lifted file logical

zishen:
What type of PR is this?
/kind cleanup

the lifted file pkg\util\lifted\objectwatcher.go and requestinfo.go changed some logical when synchronization them with community.

Which issue(s) this PR fixes:
some logical change discuss in #613 

update lifted files objectwatcher.go and requestinfo.go

NONE


zishen:
@XiShanYongYe-Chang PTAL

pkg/util/lifted/objectwatcher.go
```diff
@@ -84,5 +84,10 @@ func objectMetaObjEquivalent(a, b metav1.Object) bool {
 	if !reflect.DeepEqual(aLabels, bLabels) && (len(aLabels) != 0 || len(bLabels) != 0) {
 		return false
 	}
+	aAnnotations := a.GetAnnotations()
+	bAnnotations := b.GetAnnotations()
+	if !reflect.DeepEqual(aAnnotations, bAnnotations) && (len(aAnnotations) != 0 || len(bAnnotations) != 0) {
+		return false
+	}
```
XiShanYongYe-Chang, line 91: I think we don't need to add the judgment of Annotation, and I even think we can remove the judgment of label. Because this method is used to determine whether we need to synchronize control plane resources back to member clusters during the state collection process. It is very likely that controllers in member clusters will handle annotation and labels: 

https://github.com/karmada-io/karmada/blob/2d22b74dfce67e4f467fcfb02742cf4bfe77669b/pkg/util/objectwatcher/objectwatcher.go#L111-L117

Therefore, in most cases, this method will return true but actually it doesn't require synchronization, which wastes some computation.

XiShanYongYe-Chang (commented):
Thanks, can you help update the title more specific?

zishen:
> Thanks, can you help update the title more specific?

OK, How about "update lifted file equestinfo.go logical "?
I really don't know how to name changes in lifted file.

pkg/util/lifted/objectwatcher.go
zishen: > cher/objectwatcher.go

Yea, thank you for your careful review. Add I need to delete the compare of labeles in this pr?

XiShanYongYe-Chang: No need, it can be handled by another PR.

XiShanYongYe-Chang:
What dose equestinfo.go mean? I see you don't update this file?

zishen:
> What dose equestinfo.go mean? I see you don't update this file?

so, remove logical? or just use "update lifted files to come with version 1.26 ?"

RainbowMango (approved):
/approve

Issue #3728, opened 2023-06-28
[summer OSPP 2023] Karmada namespace deletion protection mechanism


XiShanYongYe-Chang:
**Description**

On the Karmada control plane, the same namespace manages the namespaces of all member clusters in a unified manner. If a user performs a misoperation, the namespaces are deleted from the Karmada control plane, which means that the namespaces with the same names in all member clusters, including resources in the namespace, are deleted. For example, Deployment, Job, Configmap, and Secret. The impact will be catastrophic.

Therefore, we plan to provide a namespace deletion protection mechanism in Karmada. Users can add tags to the target namespace to prevent incorrect deletion. When deleting a namespace, users need to remove the tags before deleting the namespace.

Project extension: The same processing logic can be used to process Cluster objects and resource template objects.

**Project link**

https://summer-ospp.ac.cn/org/prodetail/235c40372?list=org&navpage=org

XiShanYongYe-Chang:
All PRs have been merged in, thanks @Vacant2333 
Now we can close it.
/close

PR #3802, opened 2023-07-17
KEP: Resource deletion protection proposal

Vacant2333:
**What type of PR is this?**
Add resource deletion protection proposal
Issue: https://github.com/karmada-io/karmada/issues/3728
/kind feature
/kind documentation

```release-note
Add resource deletion protection proposal
```


docs/proposals/namespace-deletion-protection/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Motivation
+
+It is very necessary to avoid such mistakes. After implementing such a protection mechanism, we can extend this label to more types of resources, such as Deployment and Secret, etc., not only limited to protecting Namespace. In this way, it can be applicable in more scenarios.
+
+### Goals
+
+- Add or remove protection labels by karmadactl.
+- Prevent the user's deletion operation.
+- Extend to other resource types (Deployment, Secret, etc.)
```
XiShanYongYe-Chang, line 30: I think we can temporarily omit the explanation of handling other resources in our objectives.

docs/proposals/namespace-deletion-protection/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Proposal
+
+We can add a label "deletion-protected=true" to the resource(Namespace, Deployment, etc.) that needs protection. When a resource with this label is being deleted, we need to interrupt the user's operation in the controller.
```
XiShanYongYe-Chang: Can the key of this label include `karmada.io` information? Additionally, should this label be propagated during namespace propagation?

docs/proposals/namespace-deletion-protection/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+### Add protect label
+
+There are two ways to add or remove protection labels here.
+- To add a label using kubectl: `kubectl label <resource-type> <resource-name> deletion-protected=true`
```
XiShanYongYe-Chang, line 45: Can you please write down the commands for adding and removing that label separately?

docs/proposals/namespace-deletion-protection/README.md
```diff
@@ -0,0 +1,86 @@
...
+- To add a label using karmadactl: `karmadactl protect <resource-type> <resource-name>`
```
XiShanYongYe-Chang, line 46: Do we need to add a new "protect" subcommand to karmadactl? If kubectl can already handle it, maybe we don't need karmadactl for this purpose.

We can discuss this question in the community meeting.

docs/proposals/namespace-deletion-protection/README.md
```diff
@@ -0,0 +1,86 @@
...
+<!--
+This section should contain enough information that the specifics of your
+change are understandable. This may include API specs (though not always
+required) or even code snippets. If there's any ambiguity about HOW your
+proposal will be implemented, this is the place to discuss them.
+-->
+
+### Test Plan
+
+<!--
```
XiShanYongYe-Chang, line 55: What is our testing plan? UT or E2E.

XiShanYongYe-Chang (commented):
Thanks~

docs/proposals/namespace-deletion-protection/README.md
Vacant2333: We should be able to plan the implementation of other resources in advance here.

docs/proposals/namespace-deletion-protection/README.md
Vacant2333: We may can discuss this question in the community meeting.

Vacant2333:
Regarding the issue of users force deleting resources, I believe we shouldn't block this operation. Our goal is to prevent users from accidentally deleting resources, but if a user uses the `--force` parameter, we should assume that they know what they are doing. If anyone has any thoughts on this, please feel free to share them below.
/cc @XiShanYongYe-Chang

Vacant2333:
> Regarding the issue of users force deleting resources, I believe we shouldn't block this operation. Our goal is to prevent users from accidentally deleting resources, but if a user uses the `--force` parameter, we should assume that they know what they are doing. If anyone has any thoughts on this, please feel free to share them below. /cc @XiShanYongYe-Chang

/cc @zishen  @RainbowMango

XiShanYongYe-Chang:
> Regarding the issue of users force deleting resources, I believe we shouldn't block this operation. Our goal is to prevent users from accidentally deleting resources, but if a user uses the `--force` parameter, we should assume that they know what they are doing. If anyone has any thoughts on this, please feel free to share them below. /cc @XiShanYongYe-Chang

We can add this QA to the proposal.

XiShanYongYe-Chang:
Please sign off refer to https://github.com/karmada-io/karmada/pull/3802/checks?check_run_id=15416905578

Ask an review from @zishen @RainbowMango

docs/proposals/namespace-deletion-protection/README.md
```diff
@@ -0,0 +1,86 @@
...
+### Goals
+
+- Add or remove protection labels by karmadactl.
+- Prevent the user's deletion operation.
+
+## Proposal
+
+Protection Label Name: `karmada.io/deletion-protected`
+
+We can add a label `karmada.io/deletion-protected=true` to the namespace resource that needs protection. When a resource with this label is being deleted, we need to interrupt the user's operation request.
```
XiShanYongYe-Chang, line 37: Is this label added by the user or automatically added by the program?

docs/proposals/namespace-deletion-protection/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+#### Add the label
+`kubectl label <resource-type> <resource-name> karmada.io/deletion-protected=true`
+
+`karmadactl protect <resource-type> <resource-name>`
+
+#### Delete the label
+`kubectl label <resource-type> <resource-name> karmada.io/deletion-protected=false`
+
+`karmadactl protect remove <resource-type> <resource-name>`
```
XiShanYongYe-Chang, line 57: Is `remove` a subcommand of `protect`? How about `unprotect` or else? Or we can discuss this during tomorrow's meeting.

docs/proposals/namespace-deletion-protection/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Alternatives
+
+### Why choose Label over Annotations to tag resources?
+
+Annotations are typically used to store metadata that is not used for identification and selection, such as descriptive information or the status information of management tools. In this case, labels like "cannot be deleted" or "protected status" are not suitable as annotations.
+
+### What will happen If user want to delete namespace by `--force`?
+
+I believe we shouldn't block this operation. Our goal is to prevent users from accidentally deleting resources, but if a user uses the `--force` parameter, we should assume that they know what they are doing.
```
XiShanYongYe-Chang, line 86: If a user uses `--force` to perform a force delete, can the webhook detect this request and allow it?

docs/proposals/namespace-deletion-protection/README.md
zishen, line 57: When this protection is in effect, what happens if the deletion fails? Synchronous or asynchronous?

docs/proposals/namespace-deletion-protection/README.md
Vacant2333: in my opnion, add the label is by user， i will update here

Vacant2333:
Our label name may need to be changed. Does anyone have any suggestions?

docs/proposals/namespace-deletion-protection/README.md
Vacant2333: This is a good question. We might not be able to distinguish between force delete and normal delete.

docs/proposals/namespace-deletion-protection/README.md
zishen: > I didn't understand what this synchronization refers to. Could you please explain it more clearly?

It was mentioned at community meeting. When delete a source like namespace， and the resource belonging to this namespace has the `deletion-protected` label, so the namespace while delete failed and give the reasion. But if you delete by `--grace-period=0 --force`, there can see nothing but ok. So how do you show the failed reasion?

RainbowMango:
After a quick glance at this proposal, a question occurred to me: Will it be convenient enough for users to have `only one label`? What if users have a large number of namespaces to protect? They might need to add the labels to each of them.
@XiShanYongYe-Chang

XiShanYongYe-Chang:
If there are numerous resources that need to be protected, it would be tedious for users to add labels one by one. It would be more convenient if a configurable resource could be provided to inform Karmada which resources need protection.
@RainbowMango

Vacant2333:
> If there are numerous resources that need to be protected, it would be tedious for users to add labels one by one. It would be more convenient if a configurable resource could be provided to inform Karmada which resources need protection. @RainbowMango     Wondering whether such implementation would be overly complex, we only need to protect the resources of the control plane, and in most cases, there may not be too many resources that need protection.

Vacant2333:
> If there are numerous resources that need to be protected, it would be tedious for users to add labels one by one. It would be more convenient if a configurable resource could be provided to inform Karmada which resources need protection. @RainbowMango

Wondering whether such implementation would be overly complex, we only need to protect the resources of the control plane, and in most cases, there may not be too many resources that need protection.

XiShanYongYe-Chang:
> Wondering whether such implementation would be overly complex, we only need to protect the resources of the control plane, and in most cases, there may not be too many resources that need protection.

The deletion protection for individual resources is an atomic capability that we provide. If there are a large number of resources that need to be protected, we may need to consider how to optimize the user experience.

docs/proposals/namespace-deletion-protection/README.md
Vacant2333: It's a good question. In terms of the current proposal, we only protect the namespace. It indeed becomes an issue when considering the expansion to more resources later on.

Vacant2333:
> > Wondering whether such implementation would be overly complex, we only need to protect the resources of the control plane, and in most cases, there may not be too many resources that need protection.
> 
> The deletion protection for individual resources is an atomic capability that we provide. If there are a large number of resources that need to be protected, we may need to consider how to optimize the user experience.

Perhaps we can consider this issue after implementing the proposal?

XiShanYongYe-Chang:
> Perhaps we can consider this issue after implementing the proposal?
> At the moment, I don't have a good idea to solve this problem.

Yes, I agree with your point. 
In addition, we need to consider whether the default protection strategy for resources should require users to manually add labels to enable deletion protection, or if it should be automatically enabled once the resource template is created.

XiShanYongYe-Chang:
I think we can add a new featureGate that, when turned on, enables deletion protection for all namespace resources (and extends to other resources as well).
@RainbowMango @Vacant2333

Vacant2333:
This is a good idea, but perhaps users don't want to protect all namespaces by default? Maybe this featureGate could be solely responsible for enabling or disabling deletion protection.

docs/proposals/resource-deletion-protection/README.md
```diff
@@ -0,0 +1,162 @@
...
+- Namespace
+- Deployment/StatefulSet/DaemonSet
+- Service
+- ConfigMap/Secret
+
+## User Story
+
+As a user, I want to protect my resources, so that they won't be accidentally deleted.
+
+## Proposals
```
XiShanYongYe-Chang, line 41: We can explain our main proposal in the current chapter, such as the second proposal. Then we can provide explanations for the other proposals as optional choices.

docs/proposals/resource-deletion-protection/README.md
```diff
@@ -0,0 +1,162 @@
...
+
+- The user needs to add/remove tags to resources one by one.
+- The permission management is scattered, unable to follow the resources for permission control and unified handling.
+- A label is not a runtime state, and in certain situations, labels might be removed (such as when a controller is deleted and then the resource is re-created).
+
+### Proposal-2(add a CRD to specify the resources which should be protected)
+
+We can add a CRD `ResourceDeletionProtection` to specify the resources which should be protected.
+
+#### Design Details
```
XiShanYongYe-Chang, line 92: We need to provide the specific API definition for ResourceDeletionProtection.

docs/proposals/resource-deletion-protection/README.md
```diff
@@ -0,0 +1,162 @@
...
+        - deployment-1
+        - deployment-2
+    - resourceType: Service
+      resourceNames:
+        - service-1
+```
+
+##### Webhook implement detail
+
+During startup, the webhook reads all 'ResourceDeletionProtection' entries into memory and consolidates them into a map. It then watches all changes to this resource, continuously updating the map. When a user deletes a resource, this map is checked to implement caching.
```
XiShanYongYe-Chang, line 115: Is it appropriate to add this kind of processing logic in a webhook? It seems like something that should be done by a controller. I think we could directly use an informer, like this: https://github.com/karmada-io/karmada/blob/7c96e0db541989bd95772bbc1cc0109d8479371c/pkg/webhook/resourceinterpretercustomization/validating.go#L34

docs/proposals/resource-deletion-protection/README.md
```diff
@@ -0,0 +1,162 @@
...
+    },
+    "Service": {
+        "service-1": struct{}{},
+    },
+}
+```
+
+#### Alternatives
+
+##### Would this potentially lead to abuse of the webhook?
```
XiShanYongYe-Chang, line 131: This directory structure doesn't seem quite right. "Alternatives" should be placed at the end of this chapter to describe other solutions.

XiShanYongYe-Chang (commented):
Good job! Let's continue to improve this proposal.

RainbowMango:
Hi, Please [tidy the commits](https://karmada.io/docs/contributor/github-workflow#squash-commits), let's get it ready to move forward.

Vacant2333:
@kevin-wangzefeng 

> Hi, Please [tidy the commits](https://karmada.io/docs/contributor/github-workflow#squash-commits), let's get it ready to move forward.

~~ok

XiShanYongYe-Chang:
/lgtm
I think we can merge it first and then iteratively optimize it.

RainbowMango (commented):
/approve

PR #3803, opened 2023-07-17
fix:cannot reflect status to work when retain failed

### LINKED ISSUE #3572: Cannot reflect status to work when retain failed
**What happened**:
Situation 1:  When I deploy a Pod to a Kubernetes cluster which may automatically inject some fields into the Pod (e.g a certain volume is auto mounted like `k8s-api-access`, or add some default resource requests and limit), now Karmada will do the `retain` operation because of the difference of the control plane and member clusters but this may be blocked by admission controller. For Pod, we can only update some limited fields, like image and so on. At this time, the retain failure will cause the control plane to fail to collect the status of the member clusters.

Situation 2:  When I configure the error retain rules, the situation of the above situation will happen, too.

Reference codebase:

    if needUpdate {
            if err := c.ObjectWatcher.Update(clusterName, desiredObj, observedObj); err != nil {
                    klog.Errorf("Update %s failed: %v", fedKey.String(), err)
                    return err
            }
            // We can't return even after a success updates, because that might lose the chance to collect status.
            // Not all updates are real, they might be no change, in that case there will be no more event for this update,
            // this usually happens with those resources not enables 'metadata.generation', like 'Service'.
            // When a Service's status changes, it's 'metadata.resourceVersion' will be increased, but 'metadata.generation'
            // not increased(defaults to 0), the ObjectWatcher can't easily tell what happened to the object, so ObjectWatcher
            // also needs to update again. The update operation will be a non-operation if the event triggered by Service's
            // status changes.
    }

    klog.Infof("reflecting %s(%s/%s) status to Work(%s/%s)", observedObj.GetKind(), observedObj.GetNamespace(), observedObj.GetName(), workNamespace, workName)
    return c.reflectStatus(workObject, observedObj)

Even if the `retain` operation failed, Karmada can still collect the status of member clusters and keep them in sync.

**Environment**:
- Karmada version:
- Others:

### LINKED ISSUE #3587: After a resource is deleted, the resource fails to be delivered again. The work corresponding to the resource is not in the synchronization failure state.
**What happened**:
I created a service on the federation and successfully delivered a cluster. I deleted the service from the member cluster. Failed to deliver the federated data.
work status:
image

Indicates the synchronization failure status of the work corresponding to the resource.

Create a service on the federation and successfully delivered a cluster.  
Delete the service from the cluster.
Occupy the port configured for the service
 Failed to deliver the federated data again.

**Environment**:
- Karmada version: 1.5
- Others:

CuiDengdeng:
**What type of PR is this?**
/kind bug

Fixes #3587
Fixes #3572

```release-note

```

### DIFF
--- a/pkg/controllers/status/work_status_controller.go
+++ b/pkg/controllers/status/work_status_controller.go
@@ -11,6 +11,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/retry"
@@ -202,6 +203,10 @@ func (c *WorkStatusController) syncWorkStatus(key util.QueueKey) error {
 		return nil
 	}
 
+	defer func() {
+		c.reflectStatus(workObject, observedObj)
+	}()
+
 	desiredObj, err := c.getRawManifest(workObject.Spec.Workload.Manifests, observedObj)
 	if err != nil {
 		return err
@@ -235,7 +240,7 @@ func (c *WorkStatusController) syncWorkStatus(key util.QueueKey) error {
 	}
 
 	klog.Infof("reflecting %s(%s/%s) status to Work(%s/%s)", observedObj.GetKind(), observedObj.GetNamespace(), observedObj.GetName(), workNamespace, workName)
-	return c.reflectStatus(workObject, observedObj)
+	return err
 }
 
 func (c *WorkStatusController) handleDeleteEvent(key keys.FederatedKey) error {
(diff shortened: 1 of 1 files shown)


pkg/controllers/status/work_status_controller.go
```diff
@@ -271,11 +276,24 @@ func (c *WorkStatusController) recreateResourceIfNeeded(work *workv1alpha1.Work,
 		}
 
 		desiredGVK := schema.FromAPIVersionAndKind(manifest.GetAPIVersion(), manifest.GetKind())
-		if reflect.DeepEqual(desiredGVK, workloadKey.GroupVersionKind()) &&
+		if !(reflect.DeepEqual(desiredGVK, workloadKey.GroupVersionKind()) &&
```
whitewindmills, line 279: I don't understand what this expression means? Shouldn't it be like the following?
```
if !reflect.DeepEqual(desiredGVK,workloadKey.GroupVersionKind()) || manifest.GetNamespace() != workloadKey.Namespace || manifest.GetName() != workloadKey.Name)
```

whitewindmills (commented):
pls sign-off your commit.

XiShanYongYe-Chang:
We now have relevant plans in place to discuss the issue, so let's close it for now.
/close

PR #3804, opened 2023-07-18
reduce unnecessary update when processing in the workstatus controller

### LINKED ISSUE #3204: Why only labels are checked, but annotations are not checked.
https://github.com/karmada-io/karmada/blob/master/pkg/util/lifted/objectwatcher.go#L84

Why only labels are checked, but annotations are not checked.

**Environment**:
- Karmada version: master
- Kubernetes version:
- Others:

XiShanYongYe-Chang:
**What type of PR is this?**

/kind cleanup

In the `workstatus-controller`, when collecting the work status, it will check if there are any changes in the resources propagated by Karmada on the member clusters. At this time, it is necessary to check for changes in annotations and labels on the resources because we allow users to add new annotations and labels to the resources. Therefore, we only need to check those annotations and labels controlled by Karmada.

Fixes #3204 

```release-note
NONE
```

### DIFF
--- a/pkg/util/lifted/objectwatcher.go
+++ b/pkg/util/lifted/objectwatcher.go
@@ -24,11 +24,12 @@ package lifted
 
 import (
 	"fmt"
-	"reflect"
 	"strings"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


pkg/util/lifted/objectwatcher.go
```diff
@@ -72,17 +73,56 @@ func ObjectNeedsUpdate(desiredObj, clusterObj *unstructured.Unstructured, record
...
+	if desiredObj.GetNamespace() != clusterObj.GetNamespace() {
 		return false
 	}
-	aLabels := a.GetLabels()
-	bLabels := b.GetLabels()
-	if !reflect.DeepEqual(aLabels, bLabels) && (len(aLabels) != 0 || len(bLabels) != 0) {
+
+	return !(propagateAnnotationsChanged(desiredObj.GetAnnotations(), clusterObj.GetAnnotations()) ||
+		propagateLabelsChange(desiredObj.GetLabels(), clusterObj.GetLabels(),
+			strings.Split(clusterObj.GetAnnotations()[workv1alpha2.ManagedLabels], ",")))
```
zach593, line 86: use a new line for `strings.Split()` would be better

pkg/util/lifted/objectwatcher.go
```diff
@@ -72,17 +73,56 @@ func ObjectNeedsUpdate(desiredObj, clusterObj *unstructured.Unstructured, record
...
+}
+
+func propagateAnnotationsChanged(desireMaps, clusterMaps map[string]string) bool {
+	if desireMaps == nil {
+		return false
+	}
+
+	keysStr, exist := clusterMaps[workv1alpha2.ManagedAnnotation]
+	if !exist {
+		_, exist = desireMaps[workv1alpha2.ManagedAnnotation]
```
zach593, line 96: desireMaps -> desiredAnnotations

XiShanYongYe-Chang:
@zach593, I will do it later than #3808

Issue #3787, opened 2023-07-12
[umbrella] Add Implementation for PropagationPolicy priority and preemption


Poor12:
**What would you like to be added**:
In #3684, we propose a mechanism for PropagationPolicy priority and preemption.

The specific work items are as follows:

- [x] add `preemption` field for pp/cpp
- [x] add a new feature gate for preemption @Poor12 
- [x] add validation for resourceSelector if the policy enables preemption @whitewindmills 
- [x] extend the detector to implement preemption of high-priority pp/cpp -> low-priority pp/cpp @whitewindmills 
- [x] extend the detector to implement preemption of pp -> cpp @whitewindmills 
- [x] metrics and events about preemption operation
- [ ] add E2E coverage for the scenario for preemption
- [ ] document update
- [x] [Corner case recognized for PropagationPolicy](https://github.com/karmada-io/karmada/pull/3684#issuecomment-1645075275): How to transfer control to another PP/CPP during de-prioritize a PP/CPP? (#3965, @RainbowMango)
- [x] Corner case solution for ClusterPropagationPolicy.  (#3990, @chaosi-zju)

**Why is this needed**:

Ref proposal #3684

whitewindmills:
the first two are fine for me.
/assign

RainbowMango:
@whitewindmills The first task [is here](https://github.com/karmada-io/karmada/pull/3788) and you can help to review now.

Poor12:
Thanks @whitewindmills. Before you started the task, I'd like to ensure that you have enough time and be able to merge it before 7.25. If you are ok, I'd like to help you complete it. :)

RainbowMango:
> extend the detector to implement preemption of pp/cpp

I'm thinking if we can split the implementation into some small tasks. That will be easier to iterate and manage the progress.

whitewindmills:
> Thanks @whitewindmills. Before you start the task, I'd like to ensure that you have enough time and be able to merge it before 7.25. If you are ok, I'd like to help you complete it. :)

sure, glad to see you are willing to help me with this task. I have put it at the top of my work queue, I will submit a PR as soon as possible.

whitewindmills:
> I'm thinking if we can split the implementation into some small tasks. That will be easier to iterate and manage the progress.

if split, I'll pick one.

Poor12:
I have split it into some pieces of tasks. You can go ahead and implement the data structure first. @whitewindmills

whitewindmills:
[...]

PR #3805, opened 2023-07-18
feat: add policy priority annotation on resource template

whitewindmills:
**What type of PR is this?**
/kind feature

* add policy priority annotation on resource template to explain preemptive priority.
* it's part of [feature](https://github.com/karmada-io/karmada/issues/3787), see `implement the data structure to express the relationship between resource templates and type/priority of their matching policies`.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/detector/detector.go
+++ b/pkg/detector/detector.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"regexp"
+	"strconv"
 	"sync"
 	"time"
 
@@ -386,7 +387,7 @@ func (d *ResourceDetector) ApplyPolicy(object *unstructured.Unstructured, object
 		}
 	}()
 
-	if err := d.ClaimPolicyForObject(object, policy.Namespace, policy.Name); err != nil {
+	if err := d.ClaimPolicyForObject(object, policy); err != nil {
 		klog.Errorf("Failed to claim policy(%s) for object: %s", policy.Name, object)
 		return err
 	}
@@ -459,7 +460,7 @@ func (d *ResourceDetector) ApplyClusterPolicy(object *unstructured.Unstructured,
 		}
 	}()
 
-	if err := d.ClaimClusterPolicyForObject(object, policy.Name); err != nil {
+	if err := d.ClaimClusterPolicyForObject(object, policy); err != nil {
 		klog.Errorf("Failed to claim cluster policy(%s) for object: %s", policy.Name, object)
 		return err
 	}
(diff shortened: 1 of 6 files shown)


Poor12:
I envision this as a inner data structure inside the controller, like `waitingObjects map[keys.ClusterWideKey]struct{}`, it records the priority and type of a policy bound to a resource template.
Do we need to add an annotation on resource template? What do you think of it?

whitewindmills:
@Poor12
I don't tend to do that, maybe you can consider my thoughts.
* it is different from `waitingObjects`, this data structure will be very large if there are a lot of resources in the environment.
* we should introduce new lock for it, that is we cannot increase its throughput by increasing the number of workers.
* there are also some concurrent reading problems. when `detector` restarts, we can't get the mapping relationship between resource template and policy priority before it is added to this data structure.
* if we add new annotation on resource template, it can be synchronized along with policy labels without additional overhead, and easier to maintain.

whitewindmills:
> If there is an annotation, list bindings -> get object key -> get object -> check if match -> get priority from object -> update object if preemption.

path also can be: list objects by policy selector and resource selector -> get object -> check if match -> get priority from object -> update object if preemption.

RainbowMango (commented):
> > if we add new annotation on resource template, it can be synchronized along with policy labels without additional overhead, and easier to maintain.

I feel that this might be convenient when evaluating a PP should preempt a resource template or not, we can get the priority from the annotation instead of fetching the PP again.

A common question here is: what if the priority of a PP changes, can we(or should we) update the annotation? For example, if a PP with priority `1`, and then changes to `2`, should we update the annotation on the resource template?  How to do that?

RainbowMango:
I also feel that the implementation might be pretty challenging, Can we have a more details design about it? Feel free to call a meeting if necessary.

whitewindmills:
> A common question here is: what if the priority of a PP changes, can we(or should we) update the annotation? For example, if a PP with priority 1, and then changes to 2, should we update the annotation on the resource template? How to do that?

Of course it should be updated. Its update is triggered by this code.
https://github.com/karmada-io/karmada/blob/2a314eb46ec3cc1d18d1fd326559442515afbc19/pkg/detector/detector.go#L994-L1007

And I’ve added relevant code to support it's update.
```
// ClaimPolicyForObject set policy identifier which the object associated with.
func (d *ResourceDetector) ClaimPolicyForObject(object *unstructured.Unstructured, policy *policyv1alpha1.PropagationPolicy) error {
	claimedNS := util.GetLabelValue(object.GetLabels(), policyv1alpha1.PropagationPolicyNamespaceLabel)
	claimedName := util.GetLabelValue(object.GetLabels(), policyv1alpha1.PropagationPolicyNameLabel)
	claimedPriority := util.GetAnnotationValue(object.GetAnnotations(), util.PolicyPriorityAnnotation)
	policyPriority := strconv.Itoa(int(policy.ExplicitPriority()))

	// policy has no change, don't need to claim again
	if claimedNS == policy.Namespace && claimedName == policy.Name && claimedPriority == policyPriority {
		return nil
	}

	util.MergeLabel(object, policyv1alpha1.PropagationPolicyNamespaceLabel, policy.Namespace)
	util.MergeLabel(object, policyv1alpha1.PropagationPolicyNameLabel, policy.Name)
	util.MergeAnnotation(object, util.PolicyPriorityAnnotation, policyPriority)

	return d.Client.Update(context.TODO(), object)
}
```

whitewindmills:
based on the results of previous discussions, close this PR.
/close

PR #3808, opened 2023-07-18
only update object in member cluster by execution controller

### LINKED ISSUE #1362: Why use both executioncontroller and workstatuscontroller to create or update k8s object in member cluster
I'm confused that the `executioncontroller` and the `wotkstatuscontroller` both can create or update k8s object in member cluster for each work. 

I think all the operation about work can be implemented in one controller. k8s object in member cluster is just subresource of the work as pod to deploy. 

**Environment**:
- Karmada version: v1.0.1
- Kubernetes version: 
- Others:

### LINKED ISSUE #3572: Cannot reflect status to work when retain failed

### LINKED ISSUE #3587: After a resource is deleted, the resource fails to be delivered again. The work corresponding to the resource is not in the synchronization failure state.

### LINKED ISSUE #3807: Workstatus Controller Should Only Focus on Status and Only Update Object in Member Cluster by Execution Controller
Make `workstatus-controller` only focus on reflecting object status, and all operations on objects in member clusters should be finished by `execution-controller`.  

**Why is this needed**:  
Like in #1362   
1. When two controllers are attempting to update the same resource, the confict would occur easily, and it is a little hard to do some retries in one reconcile because we are not sure which one would be a later one clearly.  
2. Even both `execution-controller` and `workstatus-controller` are able to sync objects in member clusters, only `execution-controller` have a chance to update the `Applied` `condition` in `work status`. That seems a bit unreasonable. How would it be if the `workstatus-controller` failed to sync the object in member cluster and this condition last a time? The object in member cluster is not as expected, and status is wrong(old) in fed control plane, but what the `condition` in `work status` tells us is the resource is applied.

lxtywypc:
**What type of PR is this?**  
/kind feature  

As #3807   

Fixes #3807   
Fixes #1362  
Fixes #3572   
Fixes #3587 

```release-note
NONE  
```

### DIFF
--- a/cmd/agent/app/agent.go
+++ b/cmd/agent/app/agent.go
@@ -44,6 +44,7 @@ import (
 	"github.com/karmada-io/karmada/pkg/util/fedinformer/typedmanager"
 	"github.com/karmada-io/karmada/pkg/util/gclient"
 	"github.com/karmada-io/karmada/pkg/util/helper"
+	"github.com/karmada-io/karmada/pkg/util/memberclusterinformer"
 	"github.com/karmada-io/karmada/pkg/util/names"
 	"github.com/karmada-io/karmada/pkg/util/objectwatcher"
 	"github.com/karmada-io/karmada/pkg/util/restmapper"
@@ -217,7 +218,7 @@ func run(ctx context.Context, opts *options.Options) error {
 	crtlmetrics.Registry.MustRegister(metrics.ResourceCollectorsForAgent()...)
 	crtlmetrics.Registry.MustRegister(metrics.PoolCollectors()...)
 
-	if err = setupControllers(controllerManager, opts, ctx.Done()); err != nil {
+	if err = setupControllers(ctx, controllerManager, opts); err != nil {
 		return err
 	}
 
@@ -229,18 +230,18 @@ func run(ctx context.Context, opts *options.Options) error {
 	return nil
 }
 
-func setupControllers(mgr controllerruntime.Manager, opts *options.Options, stopChan <-chan struct{}) error {
+func setupControllers(ctx context.Context, mgr controllerruntime.Manager, opts *options.Options) error {
 	restConfig := mgr.GetConfig()
 	dynamicClientSet := dynamic.NewForConfigOrDie(restConfig)
-	controlPlaneInformerManager := genericmanager.NewSingleClusterInformerManager(dynamicClientSet, 0, stopChan)
+	controlPlaneInformerManager := genericmanager.NewSingleClusterInformerManager(dynamicClientSet, 0, ctx.Done())
 	controlPlaneKubeClientSet := kubeclientset.NewForConfigOrDie(restConfig)
 
 	// We need a service lister to build a resource interpreter with `ClusterIPServiceResolver`
 	// witch allows connection to the customized interpreter webhook without a cluster DNS service.
 	sharedFactory := informers.NewSharedInformerFactory(controlPlaneKubeClientSet, 0)
 	serviceLister := sharedFactory.Core().V1().Services().Lister()
-	sharedFactory.Start(stopChan)
-	sharedFactory.WaitForCacheSync(stopChan)
+	sharedFactory.Start(ctx.Done())
+	sharedFactory.WaitForCacheSync(ctx.Done())
 
 	resourceInterpreter := resourceinterpreter.NewResourceInterpreter(controlPlaneInformerManager, serviceLister)
 	if err := mgr.Add(resourceInterpreter); err != nil {
@@ -248,8 +249,11 @@ func setupControllers(mgr controllerruntime.Manager, opts *options.Options, stop
 	}
 
 	objectWatcher := objectwatcher.NewObjectWatcher(mgr.GetClient(), mgr.GetRESTMapper(), util.NewClusterDynamicClientSetForAgent, resourceInterpreter)
+	memberClusterInformer := memberclusterinformer.NewMemberClusterInformer(mgr.GetClient(), mgr.GetRESTMapper(), genericmanager.GetInstance(), opts.ClusterCacheSyncTimeout, util.NewClusterDynamicClientSetForAgent)
+
 	controllerContext := controllerscontext.Context{
 		Mgr:           mgr,
+		Ctx:           ctx,
 		ObjectWatcher: objectWatcher,
 		Opts: controllerscontext.Options{
 			Controllers:                        opts.Controllers,
@@ -269,8 +273,9 @@ func setupControllers(mgr controllerruntime.Manager, opts *options.Options, stop
 			CertRotationRemainingTimeThreshold: opts.CertRotationRemainingTimeThreshold,
 			KarmadaKubeconfigNamespace:         opts.KarmadaKubeconfigNamespace,
 		},
-		StopChan:            stopChan,
-		ResourceInterpreter: resourceInterpreter,
+		StopChan:              ctx.Done(),
+		ResourceInterpreter:   resourceInterpreter,
+		MemberClusterInformer: memberClusterInformer,
 	}
 
 	if err := controllers.StartControllers(controllerContext, controllersDisabledByDefault); err != nil {
@@ -279,7 +284,7 @@ func setupControllers(mgr controllerruntime.Manager, opts *options.Options, stop
 
 	// Ensure the InformerManager stops when the stop channel closes
 	go func() {
-		<-stopChan
+		<-ctx.Done()
 		genericmanager.StopInstance()
 	}()
 
@@ -315,14 +320,17 @@ func startClusterStatusController(ctx controllerscontext.Context) (bool, error) 
 
 func startExecutionController(ctx controllerscontext.Context) (bool, error) {
 	executionController := &execution.Controller{
-		Client:             ctx.Mgr.GetClient(),
-		EventRecorder:      ctx.Mgr.GetEventRecorderFor(execution.ControllerName),
-		RESTMapper:         ctx.Mgr.GetRESTMapper(),
-		ObjectWatcher:      ctx.ObjectWatcher,
-		PredicateFunc:      helper.NewExecutionPredicateOnAgent(),
-		InformerManager:    genericmanager.GetInstance(),
-		RatelimiterOptions: ctx.Opts.RateLimiterOptions,
+		Ctx:                   ctx.Ctx,
+		Client:                ctx.Mgr.GetClient(),
+		EventRecorder:         ctx.Mgr.GetEventRecorderFor(execution.ControllerName),
+		ObjectWatcher:         ctx.ObjectWatcher,
+		PredicateFunc:         helper.NewExecutionPredicateOnAgent(),
+		RatelimiterOptions:    ctx.Opts.RateLimiterOptions,
+		ConcurrentWorkSyncs:   ctx.Opts.ConcurrentWorkSyncs,
+		StopChan:              ctx.StopChan,
+		MemberClusterInformer: ctx.MemberClusterInformer,
 	}
+	executionController.RunWorkQueue()
 	if err := executionController.SetupWithManager(ctx.Mgr); err != nil {
 		return false, err
 	}
@@ -331,18 +339,14 @@ func startExecutionController(ctx controllerscontext.Context) (bool, error) {
 
 func startWorkStatusController(ctx controllerscontext.Context) (bool, error) {
 	workStatusController := &status.WorkStatusController{
-		Client:                      ctx.Mgr.GetClient(),
-		EventRecorder:               ctx.Mgr.GetEventRecorderFor(status.WorkStatusControllerName),
-		RESTMapper:                  ctx.Mgr.GetRESTMapper(),
-		InformerManager:             genericmanager.GetInstance(),
-		StopChan:                    ctx.StopChan,
-		ObjectWatcher:               ctx.ObjectWatcher,
-		PredicateFunc:               helper.NewExecutionPredicateOnAgent(),
-		ClusterDynamicClientSetFunc: util.NewClusterDynamicClientSetForAgent,
-		ClusterCacheSyncTimeout:     ctx.Opts.ClusterCacheSyncTimeout,
-		ConcurrentWorkStatusSyncs:   ctx.Opts.ConcurrentWorkSyncs,
-		RateLimiterOptions:          ctx.Opts.RateLimiterOptions,
-		ResourceInterpreter:         ctx.ResourceInterpreter,
+		Client:                    ctx.Mgr.GetClient(),
+		EventRecorder:             ctx.Mgr.GetEventRecorderFor(status.WorkStatusControllerName),
+		StopChan:                  ctx.StopChan,
+		PredicateFunc:             helper.NewExecutionPredicateOnAgent(),
+		ConcurrentWorkStatusSyncs: ctx.Opts.ConcurrentWorkSyncs,
+		RateLimiterOptions:        ctx.Opts.RateLimiterOptions,
+		ResourceInterpreter:       ctx.ResourceInterpreter,
+		MemberClusterInformer:     ctx.MemberClusterInformer,
 	}
 	workStatusController.RunWorkQueue()
 	if err := workStatusController.SetupWithManager(ctx.Mgr); err != nil {
--- a/pkg/controllers/execution/execution_controller.go
+++ b/pkg/controllers/execution/execution_controller.go
@@ -3,6 +3,7 @@ package execution
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
@@ -10,7 +11,9 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
@@ -26,9 +29,10 @@ import (
 	"github.com/karmada-io/karmada/pkg/metrics"
 	"github.com/karmada-io/karmada/pkg/sharedcli/ratelimiterflag"
 	"github.com/karmada-io/karmada/pkg/util"
-	"github.com/karmada-io/karmada/pkg/util/fedinformer/genericmanager"
+	"github.com/karmada-io/karmada/pkg/util/fedinformer"
 	"github.com/karmada-io/karmada/pkg/util/fedinformer/keys"
 	"github.com/karmada-io/karmada/pkg/util/helper"
+	"github.com/karmada-io/karmada/pkg/util/memberclusterinformer"
 	"github.com/karmada-io/karmada/pkg/util/names"
 	"github.com/karmada-io/karmada/pkg/util/objectwatcher"
 )
@@ -41,12 +45,19 @@ const (
 // Controller is to sync Work.
 type Controller struct {
 	client.Client      // used to operate Work resources.
+	Ctx                context.Context
 	EventRecorder      record.EventRecorder
-	RESTMapper         meta.RESTMapper
 	ObjectWatcher      objectwatcher.ObjectWatcher
 	PredicateFunc      predicate.Predicate
-	InformerManager    genericmanager.MultiClusterInformerManager
 	RatelimiterOptions ratelimiterflag.Options
+
+	// Extend execution controller like work status controller to handle event from member cluster.
+	worker util.AsyncWorker // worker process resources periodic from rateLimitingQueue.
+	// ConcurrentExecutionSyncs is the number of object that are allowed to sync concurrently.
+	ConcurrentWorkSyncs   int
+	eventHandler          cache.ResourceEventHandler // eventHandler knows how to handle events from the member cluster.
+	StopChan              <-chan struct{}
+	MemberClusterInformer memberclusterinformer.MemberClusterInformer
 }
 
 // Reconcile performs a full reconciliation for the object referred to by the Request.
@@ -65,6 +76,12 @@ func (c *Controller) Reconcile(ctx context.Context, req controllerruntime.Reques
 		return controllerruntime.Result{Requeue: true}, err
 	}
 
+	// Enqueue to try removing object in member cluster if work deleted.
+	if !work.DeletionTimestamp.IsZero() {
+		c.worker.Add(req)
+		return controllerruntime.Result{}, nil
+	}
+
 	clusterName, err := names.GetClusterName(work.Namespace)
 	if err != nil {
 		klog.Errorf("Failed to get member cluster name for work %s/%s", work.Namespace, work.Name)
@@ -77,31 +94,76 @@ func (c *Controller) Reconcile(ctx context.Context, req controllerruntime.Reques
 		return controllerruntime.Result{Requeue: true}, err
 	}
 
-	if !work.DeletionTimestamp.IsZero() {
-		// Abort deleting workload if cluster is unready when unjoining cluster, otherwise the unjoin process will be failed.
-		if util.IsClusterReady(&cluster.Status) {
-			err := c.tryDeleteWorkload(clusterName, work)
-			if err != nil {
-				klog.Errorf("Failed to delete work %v, namespace is %v, err is %v", work.Name, work.Namespace, err)
-				return controllerruntime.Result{Requeue: true}, err
-			}
-		} else if cluster.DeletionTimestamp.IsZero() { // cluster is unready, but not terminating
-			return controllerruntime.Result{Requeue: true}, fmt.Errorf("cluster(%s) not ready", cluster.Name)
-		}
-
-		return c.removeFinalizer(work)
-	}
-
 	if !util.IsClusterReady(&cluster.Status) {
 		klog.Errorf("Stop sync work(%s/%s) for cluster(%s) as cluster not ready.", work.Namespace, work.Name, cluster.Name)
 		return controllerruntime.Result{Requeue: true}, fmt.Errorf("cluster(%s) not ready", cluster.Name)
 	}
 
-	return c.syncWork(clusterName, work)
+	err = c.MemberClusterInformer.BuildResourceInformers(cluster, work, c.eventHandler)
+	if err != nil {
+		return controllerruntime.Result{Requeue: true}, err
+	}
+
+	c.worker.Add(req)
+
+	return controllerruntime.Result{}, nil
+}
+
+func (c *Controller) onUpdate(old, cur interface{}) {
+	oldObj := old.(*unstructured.Unstructured)
+	curObj := cur.(*unstructured.Unstructured)
+
+	oldObjCopy := oldObj.DeepCopy()
+	curObjCopy := curObj.DeepCopy()
+
+	clusterName, _ := names.GetClusterName(curObjCopy.GetLabels()[workv1alpha1.WorkNamespaceLabel])
+	if clusterName == "" {
+		return
+	}
+
+	if c.ObjectWatcher.NeedsUpdate(clusterName, oldObjCopy, curObjCopy) {
+		c.worker.Enqueue(curObj)
+	}
+}
+
+func (c *Controller) onDelete(obj interface{}) {
+	curObj := obj.(*unstructured.Unstructured)
+
+	c.worker.Enqueue(curObj)
+}
+
+// RunWorkQueue initializes worker and run it, worker will process resource asynchronously.
+func (c *Controller) RunWorkQueue() {
+	workerOptions := util.Options{
+		Name:          "work-execution",
+		KeyFunc:       generateKey,
+		ReconcileFunc: c.syncWork,
+	}
+	c.worker = util.NewAsyncWorker(workerOptions)
+	c.worker.Run(c.ConcurrentWorkSyncs, c.StopChan)
+}
+
+// generateKey generates a key from obj, the key contains cluster, GVK, namespace and name.
+func generateKey(obj interface{}) (util.QueueKey, error) {
+	resource, ok := obj.(*unstructured.Unstructured)
+	if !ok {
+		return nil, fmt.Errorf("object is not unstructured")
+	}
+
+	workName := util.GetLabelValue(resource.GetLabels(), workv1alpha1.WorkNameLabel)
+	workNamespace := util.GetLabelValue(resource.GetLabels(), workv1alpha1.WorkNamespaceLabel)
+
+	if workName == "" || workNamespace == "" {
+		return nil, nil
+	}
+
+	return controllerruntime.Request{NamespacedName: types.NamespacedName{Namespace: workNamespace, Name: workName}}, nil
 }
 
 // SetupWithManager creates a controller and register to controller manager.
 func (c *Controller) SetupWithManager(mgr controllerruntime.Manager) error {
+	c.eventHandler = fedinformer.NewHandlerOnEvents(nil, c.onUpdate, c.onDelete)
+
 	return controllerruntime.NewControllerManagedBy(mgr).
 		For(&workv1alpha1.Work{}, builder.WithPredicates(c.PredicateFunc)).
 		WithEventFilter(predicate.GenerationChangedPredicate{}).
@@ -111,20 +173,71 @@ func (c *Controller) SetupWithManager(mgr controllerruntime.Manager) error {
 		Complete(c)
 }
 
-func (c *Controller) syncWork(clusterName string, work *workv1alpha1.Work) (controllerruntime.Result, error) {
+func (c *Controller) syncWork(key util.QueueKey) error {
+	req, ok := key.(controllerruntime.Request)
+	if !ok {
+		klog.Warningf("Skip sync work for key(%+v) is not controllerruntime.Request type", key)
+		return nil
+	}
+
+	klog.Infof("Begin to sync work %s/%s", req.Namespace, req.Name)
+
 	start := time.Now()
-	err := c.syncToClusters(clusterName, work)
+
+	work := &workv1alpha1.Work{}
+	if err := c.Client.Get(c.Ctx, req.NamespacedName, work); err != nil {
+		// The resource may no longer exist, in which case we stop processing.
+		if apierrors.IsNotFound(err) {
+			return nil
+		}
+
+		return err
+	}
+
+	clusterName, err := names.GetClusterName(work.Namespace)
+	if err != nil {
+		klog.Errorf("Failed to get member cluster name for work %s/%s", work.Namespace, work.Name)
+		return err
+	}
+
+	cluster, err := util.GetCluster(c.Client, clusterName)
+	if err != nil {
+		klog.Errorf("Failed to get the given member cluster %s", clusterName)
+		return err
+	}
+
+	if !work.DeletionTimestamp.IsZero() {
+		// Abort deleting workload if cluster is unready when unjoining cluster, otherwise the unjoin process will be failed.
+		if util.IsClusterReady(&cluster.Status) {
+			err := c.tryDeleteWorkload(clusterName, work)
+			if err != nil {
+				klog.Errorf("Failed to delete work %v, namespace is %v, err is %v", work.Name, work.Namespace, err)
+				return err
+			}
+		} else if cluster.DeletionTimestamp.IsZero() { // cluster is unready, but not terminating
+			return fmt.Errorf("cluster(%s) not ready", cluster.Name)
+		}
+
+		return c.removeFinalizer(work)
+	}
+
+	if !util.IsClusterReady(&cluster.Status) {
+		klog.Errorf("Stop sync work(%s/%s) for cluster(%s) as cluster not ready.", work.Namespace, work.Name, cluster.Name)
+		return fmt.Errorf("cluster(%s) not ready", cluster.Name)
+	}
+
+	err = c.syncToClusters(clusterName, work)
 	metrics.ObserveSyncWorkloadLatency(err, start)
 	if err != nil {
 		msg := fmt.Sprintf("Failed to sync work(%s) to cluster(%s): %v", work.Name, clusterName, err)
 		klog.Errorf(msg)
 		c.EventRecorder.Event(work, corev1.EventTypeWarning, events.EventReasonSyncWorkloadFailed, msg)
-		return controllerruntime.Result{Requeue: true}, err
+		return err
 	}
-	msg := fmt.Sprintf("Sync work (%s) to cluster(%s) successful.", work.Name, clusterName)
+	msg := fmt.Sprintf("Sync work(%s) to cluster(%s) successfully.", work.Name, clusterName)
 	klog.V(4).Infof(msg)
 	c.EventRecorder.Event(work, corev1.EventTypeNormal, events.EventReasonSyncWorkloadSucceed, msg)
-	return controllerruntime.Result{}, nil
+	return nil
 }
 
 // tryDeleteWorkload tries to delete resource in the given member cluster.
@@ -143,7 +256,7 @@ func (c *Controller) tryDeleteWorkload(clusterName string, work *workv1alpha1.Wo
 			return err
 		}
 
-		clusterObj, err := helper.GetObjectFromCache(c.RESTMapper, c.InformerManager, fedKey)
+		clusterObj, err := c.MemberClusterInformer.GetObjectFromCache(fedKey)
 		if err != nil {
 			if apierrors.IsNotFound(err) {
 				return nil
@@ -153,7 +266,8 @@ func (c *Controller) tryDeleteWorkload(clusterName string, work *workv1alpha1.Wo
 		}
 
 		// Avoid deleting resources that not managed by karmada.
-		if util.GetLabelValue(clusterObj.GetLabels(), workv1alpha1.WorkNameLabel) != util.GetLabelValue(workload.GetLabels(), workv1alpha1.WorkNameLabel) {
+		if util.GetLabelValue(clusterObj.GetLabels(), workv1alpha1.WorkNameLabel) != util.GetLabelValue(workload.GetLabels(), workv1alpha1.WorkNameLabel) ||
+			util.GetLabelValue(clusterObj.GetLabels(), workv1alpha1.WorkNamespaceLabel) != util.GetLabelValue(workload.GetLabels(), workv1alpha1.WorkNamespaceLabel) {
 			klog.Infof("Abort deleting the resource(kind=%s, %s/%s) exists in cluster %v but not managed by karmada", clusterObj.GetKind(), clusterObj.GetNamespace(), clusterObj.GetName(), clusterName)
 			return nil
 		}
@@ -169,17 +283,17 @@ func (c *Controller) tryDeleteWorkload(clusterName string, work *workv1alpha1.Wo
 }
 
 // removeFinalizer remove finalizer from the given Work
-func (c *Controller) removeFinalizer(work *workv1alpha1.Work) (controllerruntime.Result, error) {
+func (c *Controller) removeFinalizer(work *workv1alpha1.Work) error {
 	if !controllerutil.ContainsFinalizer(work, util.ExecutionControllerFinalizer) {
-		return controllerruntime.Result{}, nil
+		return nil
 	}
 
 	controllerutil.RemoveFinalizer(work, util.ExecutionControllerFinalizer)
 	err := c.Client.Update(context.TODO(), work)
 	if err != nil {
-		return controllerruntime.Result{Requeue: true}, err
+		return err
 	}
-	return controllerruntime.Result{}, nil
+	return nil
 }
 
 // syncToClusters ensures that the state of the given object is synchronized to member clusters.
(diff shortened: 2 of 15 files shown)


pkg/controllers/execution/execution_controller.go
```diff
@@ -77,31 +95,74 @@ func (c *Controller) Reconcile(ctx context.Context, req controllerruntime.Reques
...
+func generateKey(obj interface{}) (util.QueueKey, error) {
+	resource, ok := obj.(*unstructured.Unstructured)
+	if !ok {
+		return nil, fmt.Errorf("object is not unstructured")
+	}
+
+	workName := util.GetLabelValue(resource.GetLabels(), workv1alpha1.WorkNameLabel)
+	workNamespace := util.GetLabelValue(resource.GetLabels(), workv1alpha1.WorkNamespaceLabel)
+
+	if len(workName) == 0 || len(workNamespace) == 0 {
```
jwcesign, line 155: `if workName == "" || workNamespace == ""` ?

pkg/controllers/status/work_status_controller.go
```diff
@@ -202,85 +194,18 @@ func (c *WorkStatusController) syncWorkStatus(key util.QueueKey) error {
...
-	needUpdate, err := c.ObjectWatcher.NeedsUpdate(clusterName, desiredObj, observedObj)
-	if err != nil {
+	if !helper.IsResourceApplied(&workObject.Status) {
+		err = fmt.Errorf("work hadn't been applied yet")
+		klog.Errorf("Failed to reflect status of %s(%s/%s) to Work(%s/%s): %v.",
+			observedObj.GetKind(), observedObj.GetNamespace(), observedObj.GetName(), workNamespace, workName, err)
 		return err
 	}
 
-	if needUpdate {
```
jwcesign, line 223: How is this part implemented in execution controller?

pkg/controllers/execution/execution_controller.go
lxtywypc: It was referenced from here:  
https://github.com/karmada-io/karmada/blob/d27faeac953043c2274bb8787bbbaaafa7d0dcd6/pkg/controllers/status/work_status_controller.go#L182-L187  

And I have found one more part like this, I will try to change them together.

pkg/controllers/status/work_status_controller_test.go
```diff
@@ -757,156 +703,151 @@ func newWork(workName, workNs string, raw []byte, deletionTimestampIsZero ...boo
...
-		assert.Empty(t, err)
-	})
-
-	t.Run("failed to UnmarshalJSON", func(t *testing.T) {
-		work.Spec.Workload.Manifests[0].RawExtension.Raw = []byte(`{"apiVersion":"v1","kind":"Pod","metadata":{"name":"pod","namespace":"default"}},`)
-		err := c.recreateResourceIfNeeded(work, fedKey)
-		assert.NotEmpty(t, err)
-	})
-}
+// TODO: Move to resourceinformer_test
```
XiShanYongYe-Chang, line 706: Can the move be completed in the current PR?

XiShanYongYe-Chang (commented):
Good job, thanks a lot~

pkg/controllers/execution/execution_controller.go
```diff
@@ -65,6 +77,12 @@ func (c *Controller) Reconcile(ctx context.Context, req controllerruntime.Reques
 		return controllerruntime.Result{Requeue: true}, err
 	}
 
+	// Enqueue to try removing object in member cluster if work deleted.
+	if work.DeletionTimestamp != nil {
```
whitewindmills, line 81: usually use `!work.DeletionTimestamp.IsZero()`.

XiShanYongYe-Chang:
Invite people who may be interested in this issue to take a look.
/cc @chaunceyjiang @whitewindmills @Poor12 @RainbowMango @CuiDengdeng @likakuli

pkg/controllers/status/work_status_controller.go
jwcesign: needUpdate is decided by `c.ObjectWatcher.NeedsUpdate(clusterName, desiredObj, observedObj)`, and it checks the labels when the generation is the same, but `GenerationChangedPredicate` only checks the generation, is this expected?

pkg/controllers/status/work_status_controller.go
```diff
@@ -374,109 +299,10 @@ func (c *WorkStatusController) mergeStatus(_ []workv1alpha1.ManifestStatus, newS
...
-			return nil, err
-		}
-		singleClusterInformerManager = c.InformerManager.ForCluster(dynamicClusterClient.ClusterName, dynamicClusterClient.DynamicClientSet, 0)
-	}
-	return singleClusterInformerManager, nil
-}
-
 // SetupWithManager creates a controller and register to controller manager.
 func (c *WorkStatusController) SetupWithManager(mgr controllerruntime.Manager) error {
+	c.eventHandler = fedinformer.NewHandlerOnEvents(c.onAdd, c.onUpdate, nil)
```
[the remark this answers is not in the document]
lxtywypc: In my opinion, when an object was deleted, the `execution controller` would take effect on the event to give a try to recreate the object. The `Applied` condition would be set to `False` if the synchronization failed. 

But maybe we can clear the `manifestStatuses` field when a synchronization failed, that could be completed in `execution controller`.

> BTW, can you help remove the unused NewHandlerOnAllEvents method?  

I'm glad to do this after we make sure it quite be no longer used.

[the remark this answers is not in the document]
XiShanYongYe-Chang: Ok, it's reasonable, thanks!

pkg/controllers/execution/execution_controller.go
```diff
@@ -77,31 +95,74 @@ func (c *Controller) Reconcile(ctx context.Context, req controllerruntime.Reques
...
+
+func (c *Controller) onUpdate(old, cur interface{}) {
+	unstructuredOld := old.(*unstructured.Unstructured)
+	unstructuredCur := cur.(*unstructured.Unstructured)
+
+	unstructuredOldCopy := unstructuredOld.DeepCopy()
+	unstructuredCurCopy := unstructuredCur.DeepCopy()
+
+	_ = prune.RemoveIrrelevantField(unstructuredOldCopy, prune.RemoveJobTTLSeconds)
+	_ = prune.RemoveIrrelevantField(unstructuredCurCopy, prune.RemoveJobTTLSeconds)
```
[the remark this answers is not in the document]
lxtywypc: Sounds great, I will take a try.

pkg/controllers/status/work_status_controller.go
```diff
@@ -171,10 +157,16 @@ func (c *WorkStatusController) syncWorkStatus(key util.QueueKey) error {
...
+
+	observedObj, err := c.ResourceInformer.GetObjectFromCache(fedKey)
 	if err != nil {
 		if apierrors.IsNotFound(err) {
-			return c.handleDeleteEvent(fedKey)
+			executionSpace := names.GenerateExecutionSpaceName(fedKey.Cluster)
+			workName := names.GenerateWorkName(fedKey.Kind, fedKey.Name, fedKey.Namespace)
+			klog.Warningf("Skip reflecting %s(%s/%s) status to Work(%s/%s) for not applied successfully yet.",
+				fedKey.Kind, fedKey.Namespace, fedKey.Name, executionSpace, workName)
+			return nil
```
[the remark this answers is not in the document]
lxtywypc: Like this:  
https://github.com/karmada-io/karmada/pull/3808/files/cdbc943859f432dd125ca24c4fe2e3d7e468ae2e#r1267468944  

If the object should be managed by karmada, `execution controller` would take an effect and update the status of `work`, and this event could be received by `bindingstatus controller` to update the status in `binding` and resource on fed-control-plane

pkg/controllers/status/work_status_controller.go
```diff
@@ -202,85 +194,18 @@ func (c *WorkStatusController) syncWorkStatus(key util.QueueKey) error {
...
-	}
-
-	// we should check if the observed status is consistent with the declaration to prevent accidental changes made
-	// in member clusters.
-	needUpdate, err := c.ObjectWatcher.NeedsUpdate(clusterName, desiredObj, observedObj)
-	if err != nil {
+	if !helper.IsResourceApplied(&workObject.Status) {
+		err = fmt.Errorf("work hadn't been applied yet")
+		klog.Errorf("Failed to reflect status of %s(%s/%s) to Work(%s/%s): %v.",
+			observedObj.GetKind(), observedObj.GetNamespace(), observedObj.GetName(), workNamespace, workName, err)
```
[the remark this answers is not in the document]
lxtywypc: It is necessary to return an error to trigger another reconcile, because this part may run after `execution controller` applied object successfully but before `execution controller` updated `Applied` condition in `work` successfully. That would cause we miss the chance to reflect the status until next update event.  

Or we have to watch the update event of the conditions of `work` and enqueue the object manually.  

How do you think about?

pkg/controllers/status/work_status_controller.go
lxtywypc: Like this:  
https://github.com/karmada-io/karmada/pull/3808/files/cdbc943859f432dd125ca24c4fe2e3d7e468ae2e#r1267840563  

I will think more about this. 

Nice remind, thanks.

pkg/controllers/status/work_status_controller.go
```diff
@@ -202,85 +194,18 @@ func (c *WorkStatusController) syncWorkStatus(key util.QueueKey) error {
...
-		// We can't return even after a success updates, because that might lose the chance to collect status.
-		// Not all updates are real, they might be no change, in that case there will be no more event for this update,
-		// this usually happens with those resources not enables 'metadata.generation', like 'Service'.
-		// When a Service's status changes, it's 'metadata.resourceVersion' will be increased, but 'metadata.generation'
-		// not increased(defaults to 0), the ObjectWatcher can't easily tell what happened to the object, so ObjectWatcher
-		// also needs to update again. The update operation will be a non-operation if the event triggered by Service's
-		// status changes.
-	}
-
+	// Now we only update object in member cluster by execution_controller.
```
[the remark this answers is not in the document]
lxtywypc: I just supposed to remind readers what I have done. Ofcourse it could be removed.

pkg/controllers/execution/execution_controller.go
lxtywypc: Nice remind, thanks.

pkg/controllers/status/work_status_controller.go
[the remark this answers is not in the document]
XiShanYongYe-Chang: Shouldn't the work status condition be handled by the execution-controller and the work status manifestStatuses be handled by the work-status-controller, so that they don't conflict with each other?

pkg/controllers/status/work_status_controller.go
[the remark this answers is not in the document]
XiShanYongYe-Chang: Thanks for explaining, I get this point. 
I have a new question, do we need to make this judgment?
```
if !helper.IsResourceApplied(&workObject.Status) {
```

cmd/agent/app/agent.go
```diff
@@ -226,7 +227,7 @@ func run(ctx context.Context, opts *options.Options) error {
 	return nil
 }
 
-func setupControllers(mgr controllerruntime.Manager, opts *options.Options, stopChan <-chan struct{}) error {
+func setupControllers(ctx context.Context, mgr controllerruntime.Manager, opts *options.Options, stopChan <-chan struct{}) error {
```
chaunceyjiang, line 230: `stopChan <-chan ` seems to be a redundant parameter.

pkg/util/resourceinformer/resourceinformer.go
```diff
@@ -0,0 +1,151 @@
...
+	"github.com/karmada-io/karmada/pkg/util/helper"
+	"github.com/karmada-io/karmada/pkg/util/restmapper"
+)
+
+var _ ResourceInformer = &resourceInformerImpl{}
+
+type ResourceInformer interface {
+	// BuildResourceInformers builds informer dynamically for managed resources in member cluster.
+	// The created informer watches resource change and then sync to the relevant Work object.
+	BuildResourceInformers(cluster *clusterv1alpha1.Cluster, work *workv1alpha1.Work, handler cache.ResourceEventHandler) (controllerruntime.Result, error)
```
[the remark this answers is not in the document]
chaunceyjiang: +1 

`controllerruntime.Result` doesn't seem to be a generic return type. I believe it would be beneficial to declare a new type for easier future expansion.

pkg/util/resourceinformer/resourceinformer.go
```diff
@@ -0,0 +1,151 @@
...
+	ClusterCacheSyncTimeout     metav1.Duration
+}
+
+// BuildResourceInformers builds informer dynamically for managed resources in member cluster.
+// The created informer watches resource change and then sync to the relevant Work object.
+func (r *resourceInformerImpl) BuildResourceInformers(cluster *clusterv1alpha1.Cluster, work *workv1alpha1.Work, handler cache.ResourceEventHandler) (controllerruntime.Result, error) {
+	err := r.registerInformersAndStart(cluster, work, handler)
+	if err != nil {
+		klog.Errorf("Failed to register informer for Work %s/%s. Error: %v.", work.GetNamespace(), work.GetName(), err)
+		return controllerruntime.Result{Requeue: true}, err
```
chaunceyjiang, line 60: Whether to "Requeue" or not, should be determined by the caller.

pkg/controllers/status/work_status_controller.go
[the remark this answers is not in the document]
lxtywypc: Yes, it seems to be better. That's quite a nice advice.

pkg/controllers/status/work_status_controller.go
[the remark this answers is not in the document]
lxtywypc: In my opinion, if an object hadn't been synced successfully yet, it is unnecessary to reflect a real status in member cluster based on an unexpected object template. It is enough that the `execution controller` would set the `Applied` condition as `False` to reflect this unexpected situation on the fed-control-plane. After the object been synced successfully, the status is grate probable to change. Return in time would reduce unnecessary updating a lot.  

And about your advice on https://github.com/karmada-io/karmada/pull/3808/files/cdbc943859f432dd125ca24c4fe2e3d7e468ae2e#r1267481686, which do you think is better when `Applied` condition changed to `False`? To keep the old status, or clean up it?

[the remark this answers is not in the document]
XiShanYongYe-Chang: Ok thanks.

> To keep the old status,

pkg/controllers/status/work_status_controller.go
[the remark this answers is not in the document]
lxtywypc: Sorry, after re-read about this part, I find that it seems not necessary to clean up the status in `work`. 

When we try to sync status of an object but the object is missed in cache:  
- If the object is not managed by `karmada`, we don't need to process it.  
- Else, we could wait for the object re-created by `execution-controller`. It could be considered as what we should do when `Applied` condition is `False`. Then the `bindingstatus controller` would take an effect on `binding` and resource on fed-control-plane.  

Or maybe I mis-understood your intention. If so, please explain to me in detail. :)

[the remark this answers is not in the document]
XiShanYongYe-Chang: Ok, I agree with you.

pkg/controllers/execution/execution_controller.go
```diff
@@ -77,31 +94,76 @@ func (c *Controller) Reconcile(ctx context.Context, req controllerruntime.Reques
...
-
 	if !util.IsClusterReady(&cluster.Status) {
 		klog.Errorf("Stop sync work(%s/%s) for cluster(%s) as cluster not ready.", work.Namespace, work.Name, cluster.Name)
 		return controllerruntime.Result{Requeue: true}, fmt.Errorf("cluster(%s) not ready", cluster.Name)
 	}
 
-	return c.syncWork(clusterName, work)
+	err = c.ResourceInformer.BuildResourceInformers(cluster, work, c.eventHandler)
+	if err != nil {
+		return controllerruntime.Result{Requeue: true}, err
```
whitewindmills, line 104: The codes `Requeue: true` is unnecessary.

pkg/controllers/status/work_status_controller.go
```diff
@@ -99,26 +91,25 @@ func (c *WorkStatusController) Reconcile(ctx context.Context, req controllerrunt
...
-func (c *WorkStatusController) getEventHandler() cache.ResourceEventHandler {
-	if c.eventHandler == nil {
-		c.eventHandler = fedinformer.NewHandlerOnAllEvents(c.worker.Enqueue)
+func (c *WorkStatusController) onAdd(obj interface{}) {
+	curObj := obj.(runtime.Object)
+	c.worker.Enqueue(curObj)
+}
+
+func (c *WorkStatusController) onUpdate(old, cur interface{}) {
+	// Still need compare the whole object because the interpreter is able to take any part of object to reflect status.
```
whitewindmills, line 108: // Still need to compare...

pkg/util/objectwatcher/objectwatcher.go
```diff
@@ -70,29 +66,21 @@ func (o *objectWatcherImpl) Create(clusterName string, desireObj *unstructured.U
...
-			klog.Errorf("Failed to get resource %v from member cluster, err is %v ", desireObj.GetName(), err)
-			return err
-		}
-
-		clusterObj, err := dynamicClusterClient.DynamicClientSet.Resource(gvr).Namespace(desireObj.GetNamespace()).Create(context.TODO(), desireObj, metav1.CreateOptions{})
-		if err != nil {
-			klog.Errorf("Failed to create resource(kind=%s, %s/%s) in cluster %s, err is %v ", desireObj.GetKind(), desireObj.GetNamespace(), desireObj.GetName(), clusterName, err)
-			return err
+		if apierrors.IsAlreadyExists(err) {
+			klog.Infof("Skip creating resource(kind=%s, %s/%s) in cluster %s as resource is already existing", desireObj.GetKind(), desireObj.GetNamespace(), desireObj.GetName(), clusterName)
```
whitewindmills, line 72: `Skip creating resource(kind=%s, %s/%s) in cluster %s because it already exists`

test/helper/resource.go
```diff
@@ -765,3 +786,29 @@ func NewPodDisruptionBudget(namespace, name string, maxUnAvailable intstr.IntOrS
 		},
 	}
 }
+
+// NewWork will build a new Work object.
+func NewWork(workName, workNs string, raw []byte, deletionTimestampIsZero ...bool) *workv1alpha1.Work {
```
whitewindmills, line 791: `deletionTimestampIsZero` seems unused, even if reserved, `deletionTimestampIsZero bool` is enough.

pkg/controllers/execution/execution_controller.go
lxtywypc: Yes, I just follow the original style. We could fix all of parts in this style in a new pull-request if need. :)

pkg/controllers/execution/execution_controller.go
lxtywypc: Done.

pkg/util/resourceinformer/resourceinformer.go
```diff
@@ -0,0 +1,151 @@
...
+type ResourceInformer interface {
+	// BuildResourceInformers builds informer dynamically for managed resources in member cluster.
+	// The created informer watches resource change and then sync to the relevant Work object.
+	BuildResourceInformers(cluster *clusterv1alpha1.Cluster, work *workv1alpha1.Work, handler cache.ResourceEventHandler) (controllerruntime.Result, error)
+
+	// GetObjectFromCache returns object in cache by federated key.
+	GetObjectFromCache(fedKey keys.FederatedKey) (*unstructured.Unstructured, error)
+}
+
+func NewResourceInformer(c client.Client, restMapper meta.RESTMapper, informerManager genericmanager.MultiClusterInformerManager, clusterCacheSyncTimeout metav1.Duration,
```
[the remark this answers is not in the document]
lxtywypc: Done.

pkg/util/resourceinformer/resourceinformer.go
```diff
@@ -0,0 +1,151 @@
...
+		ClusterCacheSyncTimeout:     clusterCacheSyncTimeout,
+	}
+}
+
+type resourceInformerImpl struct {
+	client.Client               // used to get Cluster and Secret resources.
+	RESTMapper                  meta.RESTMapper
+	InformerManager             genericmanager.MultiClusterInformerManager
+	ClusterDynamicClientSetFunc func(clusterName string, client client.Client) (*util.DynamicClusterClient, error)
+	ClusterCacheSyncTimeout     metav1.Duration
```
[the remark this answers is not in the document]
lxtywypc: Done.

pkg/util/resourceinformer/resourceinformer.go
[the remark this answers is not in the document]
lxtywypc: Done.

pkg/controllers/status/work_status_controller_test.go
lxtywypc: Done. And I also added some unit test for `execution controller`.

pkg/controllers/execution/execution_controller.go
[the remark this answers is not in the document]
lxtywypc: I decided to use `ObjectWatcher.NeedsUpdate` here to judge if an update event should trigger the synchronization. And https://github.com/karmada-io/karmada/pull/3804 is still a nice improvement.

pkg/controllers/execution/execution_controller.go
lxtywypc: Done.

cmd/agent/app/agent.go
lxtywypc: I have removed this parameter and use ctx.Done instead.  

There are still some controllers own the element `stopChan`, maybe we could change them to `ctx` later.

pkg/util/resourceinformer/resourceinformer.go
lxtywypc: Done.

pkg/controllers/execution/execution_controller.go
```diff
@@ -111,20 +173,71 @@ func (c *Controller) SetupWithManager(mgr controllerruntime.Manager) error {
...
+	err = c.syncToClusters(clusterName, work)
 	metrics.ObserveSyncWorkloadLatency(err, start)
 	if err != nil {
 		msg := fmt.Sprintf("Failed to sync work(%s) to cluster(%s): %v", work.Name, clusterName, err)
 		klog.Errorf(msg)
 		c.EventRecorder.Event(work, corev1.EventTypeWarning, events.EventReasonSyncWorkloadFailed, msg)
-		return controllerruntime.Result{Requeue: true}, err
+		return err
 	}
 	msg := fmt.Sprintf("Sync work (%s) to cluster(%s) successful.", work.Name, clusterName)
```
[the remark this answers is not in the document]
lxtywypc: Done.

pkg/controllers/status/work_status_controller.go
lxtywypc: Done.

pkg/util/objectwatcher/objectwatcher.go
lxtywypc: Done.

test/helper/resource.go
lxtywypc: Done.

lxtywypc:
I have made some modification following the advices in a new commit, I will squash them into one commit and remove useless parts after the review is almost finished.

Kindly ping @XiShanYongYe-Chang @chaunceyjiang @jwcesign @whitewindmills to take a review again. :)

XiShanYongYe-Chang:
@lxtywypc Thanks~ can you help to close the resolved questions?

lxtywypc:
> @lxtywypc Thanks~ can you help to close the resolved questions?

I closed some questions that with no dispute, and I left some that would be done after the review is almost finished, or the reviewers may have further thought about. :)

pkg/controllers/execution/execution_controller.go
```diff
@@ -77,31 +94,76 @@ func (c *Controller) Reconcile(ctx context.Context, req controllerruntime.Reques
...
+
+	oldObjCopy := oldObj.DeepCopy()
+	curObjCopy := curObj.DeepCopy()
+
+	clusterName, _ := names.GetClusterName(curObjCopy.GetLabels()[workv1alpha1.WorkNamespaceLabel])
+	if clusterName == "" {
+		return
+	}
+
+	if c.ObjectWatcher.NeedsUpdate(clusterName, oldObjCopy, curObjCopy) {
```
XiShanYongYe-Chang, line 124: NeedsUpdate is used to determine the differences between resources on the control plane and resources in the member cluster. It may not be appropriate to use it directly here. Can we directly compare the spec, annotation, and label of the resource instead?

pkg/controllers/status/work_status_controller.go
```diff
@@ -150,7 +141,7 @@ func generateKey(obj interface{}) (util.QueueKey, error) {
 // getClusterNameFromLabel gets cluster name from ownerLabel, if label not exist, means resource is not created by karmada.
 func getClusterNameFromLabel(resource *unstructured.Unstructured) (string, error) {
 	workNamespace := util.GetLabelValue(resource.GetLabels(), workv1alpha1.WorkNamespaceLabel)
-	if len(workNamespace) == 0 {
+	if workNamespace == "" {
```
XiShanYongYe-Chang, line 144: This modification doesn't seem to have any changes.

pkg/util/objectwatcher/objectwatcher.go
```diff
@@ -254,20 +262,20 @@ func (o *objectWatcherImpl) deleteVersionRecord(clusterName, resourceName string
 	delete(o.VersionRecord[clusterName], resourceName)
 }
 
-func (o *objectWatcherImpl) NeedsUpdate(clusterName string, desiredObj, clusterObj *unstructured.Unstructured) (bool, error) {
-	// get resource version
+func (o *objectWatcherImpl) NeedsUpdate(clusterName string, desiredObj, clusterObj *unstructured.Unstructured) bool {
```
XiShanYongYe-Chang, line 265: Perhaps we don't need this method.

pkg/controllers/execution/execution_controller.go
lxtywypc: The `ObjectWather.NeedsUpdate` is used in `workstatus controller` to determine if the object in member cluster needs to sync by comparing the `generation` in current object with that last updated recorded in RAM, and comparing the labels, (would be labels and annotations managed by `karmada` after #3804), between the object declared in `work` and that in member cluster. 

In fact, only labels(and annotations later) in desired object is useful for comparation. In a normal workflow, we consider the object in member cluster always the same with our desired. So if we treat the old object in an update event as the desired object, I think it matches the description you gave above.

XiShanYongYe-Chang: Thanks for explaining. Now that we have the updated resources before and after, can we directly compare their generation? So can we consider deleting the version stored in memory?

pkg/util/objectwatcher/objectwatcher.go
lxtywypc: As https://github.com/karmada-io/karmada/pull/3808#discussion_r1274514729, I still think comparing the `generation` between current object and recorded since last updating is a good and explicit way to tell us if the object has been modified unexpectly.

pkg/controllers/status/work_status_controller.go
lxtywypc: It is just a cleanup, for https://github.com/karmada-io/karmada/pull/3808#discussion_r1267455597.

pkg/util/objectwatcher/objectwatcher.go
XiShanYongYe-Chang: Yes, I agree with you. Sorry, I didn't make it clear. My point is that we can remove the version recorded in memory and directly compare the generations of the two objects.

pkg/controllers/execution/execution_controller.go
```diff
@@ -153,7 +266,8 @@ func (c *Controller) tryDeleteWorkload(clusterName string, work *workv1alpha1.Wo
 		}
 
 		// Avoid deleting resources that not managed by karmada.
-		if util.GetLabelValue(clusterObj.GetLabels(), workv1alpha1.WorkNameLabel) != util.GetLabelValue(workload.GetLabels(), workv1alpha1.WorkNameLabel) {
+		if util.GetLabelValue(clusterObj.GetLabels(), workv1alpha1.WorkNameLabel) != util.GetLabelValue(workload.GetLabels(), workv1alpha1.WorkNameLabel) ||
+			util.GetLabelValue(clusterObj.GetLabels(), workv1alpha1.WorkNamespaceLabel) != util.GetLabelValue(workload.GetLabels(), workv1alpha1.WorkNamespaceLabel) {
```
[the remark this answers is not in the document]
lxtywypc: It prevents some special cases. For example, if someone removed the `WorkNamespaceLabel` of object in member cluster, attempting to leave it alone in member cluster, and removed the object in fed-control-plane. Though this might not be a recommended usage, I think we had better do nothing.

XiShanYongYe-Chang:
Ask @whitewindmills @chaunceyjiang @jwcesign to help have a look again.

pkg/util/objectwatcher/objectwatcher.go
lxtywypc: Yes, you are right. I always imagine too much about some special cases not in a normal workflow. Maybe we could talk about these cases in a new issue. Here I will do some simple comparation. I still expect for #3804! :)

lxtywypc: Sorry, I just came up with that an important reason why I attempted to reuse this part:   

Comparing the `generation` in memory could prevent requeueing from the update event triggered by `karmada` self.

If we overlook this situation, we may have two attempts to update the object in a short time after once modification.   

What do you think about this situation?

XiShanYongYe-Chang: I'm sorry, I didn't understand that.

lxtywypc: For example, we have an object managed by `karmada` with `generation = 1` now in member cluster.  

Then I unexpectly modified the `spec` field of the object in member cluster. Now it is with `generation = 2`. This would trigger the updating from `execution controller` because of the changing of `generation`.  

The `execution controller` would sync the object in member cluster. Now the object is with `generation = 3`. This would immediately trigger another updating from `execution controller` again, and it is still because of the changing of `generation`.  

Besides, even the first updating is from the modification of `work`, the `execution controller` would attempt to update the  object twice because it would watch the changing of `generation` in object in member cluster caused by the modification by itself.

XiShanYongYe-Chang: Thanks, I got it.

lxtywypc: Do you have any advices about this? Or we had better keep use `ObjectWatcher.NeedsUpdate`?

XiShanYongYe-Chang: I think it can still be used, but the parameter names need to be modified in order to make them more consistent with their current meaning.

pkg/controllers/status/work_status_controller.go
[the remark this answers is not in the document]
lxtywypc: Done.

pkg/controllers/status/work_status_controller.go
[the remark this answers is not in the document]
lxtywypc: > BTW, can you help remove the unused NewHandlerOnAllEvents method?  

Done.

pkg/util/objectwatcher/objectwatcher.go
lxtywypc: Done.  

And I also made some modifications on the way to record version, to prevent enqueueing object on update event too early when the updated version hasn't been recorded yet.

pkg/util/objectwatcher/objectwatcher.go
```diff
@@ -203,71 +222,93 @@ func (o *objectWatcherImpl) genObjectKey(obj *unstructured.Unstructured) string
...
+	versionLock, exist := o.getVersionWithLockRecord(clusterName, resourceName)
+
+	if !exist {
+		versionLock = o.newVersionWithLockRecord(clusterName, resourceName)
+	}
+
+	versionLock.lock.Lock()
+	defer versionLock.lock.Unlock()
+
+	versionLock.version, err = fn()
```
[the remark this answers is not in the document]
lxtywypc: Done.  

That's quite a nice remind. Thanks~ :)

XiShanYongYe-Chang (commented):
LGTM
Thanks a lot.
Ask @RainbowMango to help take a look.

RainbowMango:
By the way, can we split this into some small PRs?

lxtywypc:
> By the way, can we split this into some small PRs?

A little hard to.  

The core of this pr is moving the part of syncing object in member cluster from `workstatus controller` to `execution controller`, this part cannot be split into several parts.   

We might only be able to split this pr into several parts as following:  
1. Abstract the `resourceinformer` for later using.  
2. Move the part of syncing object in member cluster from `workstatus controller` to `execution controller`.  
3. Fill the unit tests.  

The second part might be still large. And it might be better to put the testing part and logic part together.

RainbowMango:
Never mind, I'll take a look at it as soon as possible, perhaps this week.

RainbowMango (commented):
/assign

pkg/util/resourceinformer/resourceinformer.go
```diff
@@ -0,0 +1,155 @@
...
+	"github.com/karmada-io/karmada/pkg/util"
+	"github.com/karmada-io/karmada/pkg/util/fedinformer/genericmanager"
+	"github.com/karmada-io/karmada/pkg/util/fedinformer/keys"
+	"github.com/karmada-io/karmada/pkg/util/helper"
+	"github.com/karmada-io/karmada/pkg/util/restmapper"
+)
+
+var _ ResourceInformer = &resourceInformerImpl{}
+
+type ResourceInformer interface {
```
RainbowMango, line 25: I like the idea of abstracting the informer-related functionalities.
It is another informer manager based on [genericmanager.MultiClusterInformerManager](https://github.com/karmada-io/karmada/tree/master/pkg/util/fedinformer/genericmanager) with more context of Karmada resource types(such as it recognizes `Cluster`, `Work`).

 I just have a feeling that the name of it (`ResourceInformer`) seems `a little bit generic` and doesn't show its responsibility.
I'm not sure if `MemberClusterInformer` would be better. Looking forward to your opinions.

pkg/util/objectwatcher/objectwatcher.go
```diff
@@ -29,31 +27,36 @@ type ObjectWatcher interface {
 	Create(clusterName string, desireObj *unstructured.Unstructured) error
 	Update(clusterName string, desireObj, clusterObj *unstructured.Unstructured) error
 	Delete(clusterName string, desireObj *unstructured.Unstructured) error
-	NeedsUpdate(clusterName string, desiredObj, clusterObj *unstructured.Unstructured) (bool, error)
+	NeedsUpdate(clusterName string, oldObj, currentObj *unstructured.Unstructured) bool
```
RainbowMango, line 30: If there is any specific reason for changing the signature here? The `desired` represents the expectation state, it's might not an `old` one.

XiShanYongYe-Chang: The modified logic compares the resources before and after the update on the member cluster, so I suggest changing the naming here. "NeedUpdate" is actually not suitable to be placed here anymore, it needs to be optimized in the future.

pkg/util/resourceinformer/resourceinformer.go
lxtywypc: I just keep the name as what it was in `workstatus controller`.  
I think it is ok to use `MemberClusterInformer` to instead. I will fix it soon.

lxtywypc: Done.

lxtywypc:
Kindly ping @RainbowMango   
Is there still any problem?

RainbowMango:
Sorry for the delay, It's still on my TODO list. It's hard for me to have a big chunk of time to review it(that's exactly why I always suggested splitting PRs :)). But I'll try to take a look at it today if no comments from me controller owners (@XiShanYongYe-Chang @chaunceyjiang ) can help to move forward.

RainbowMango (approved):
/approve

Generally looks good to me, I believe this is a workable solution but didn't review it thoroughly, Please @chaunceyjiang and @XiShanYongYe-Chang make final decisions.

Sorry again for the delay.

XiShanYongYe-Chang:
Thanks @lxtywypc! This is a significant improvement. It's LGTM. 
Ask @chaunceyjiang to help to have another review.
/cc @chaunceyjiang

pkg/util/memberclusterinformer/memberclusterinformer.go
```diff
@@ -0,0 +1,155 @@
...
+			return nil, err
+		}
+		singleClusterInformerManager = m.informerManager.ForCluster(dynamicClusterClient.ClusterName, dynamicClusterClient.DynamicClientSet, 0)
+	}
+	return singleClusterInformerManager, nil
+}
+
+// getGVRsFromWork traverses the manifests in work to find groupVersionResource list.
+func (m *memberClusterInformerImpl) getGVRsFromWork(work *workv1alpha1.Work) (map[schema.GroupVersionResource]bool, error) {
+	gvrTargets := map[schema.GroupVersionResource]bool{}
```
chaunceyjiang, line 134:
```suggestion
	gvrTargets := sets.New[schema.GroupVersionResource]()
```

pkg/util/objectwatcher/objectwatcher.go
```diff
@@ -70,30 +73,21 @@ func (o *objectWatcherImpl) Create(clusterName string, desireObj *unstructured.U
...
-	}
-	_, err = helper.GetObjectFromCache(o.RESTMapper, o.InformerManager, fedKey)
+	clusterObj, err := dynamicClusterClient.DynamicClientSet.Resource(gvr).Namespace(desireObj.GetNamespace()).Create(context.TODO(), desireObj, metav1.CreateOptions{})
 	if err != nil {
-		if !apierrors.IsNotFound(err) {
-			klog.Errorf("Failed to get resource %v from member cluster, err is %v ", desireObj.GetName(), err)
-			return err
+		if apierrors.IsAlreadyExists(err) {
+			klog.Infof("Skip creating resource(kind=%s, %s/%s) in cluster %s because it already exists", desireObj.GetKind(), desireObj.GetNamespace(), desireObj.GetName(), clusterName)
+			return nil
```
chaunceyjiang, line 80: I believe that an error should still be returned here. 
The situation where the resource already exists should not be ignored.

chaunceyjiang (commented):
Stilling review.

pkg/util/memberclusterinformer/memberclusterinformer.go
lxtywypc: Done.

pkg/util/objectwatcher/objectwatcher.go
lxtywypc: Done.

pkg/controllers/execution/execution_controller.go
```diff
@@ -111,20 +173,71 @@ func (c *Controller) SetupWithManager(mgr controllerruntime.Manager) error {
...
+	if !work.DeletionTimestamp.IsZero() {
+		// Abort deleting workload if cluster is unready when unjoining cluster, otherwise the unjoin process will be failed.
+		if util.IsClusterReady(&cluster.Status) {
+			err := c.tryDeleteWorkload(clusterName, work)
+			if err != nil {
+				klog.Errorf("Failed to delete work %v, namespace is %v, err is %v", work.Name, work.Namespace, err)
+				return err
+			}
+		} else if cluster.DeletionTimestamp.IsZero() { // cluster is unready, but not terminating
+			return fmt.Errorf("cluster(%s) not ready", cluster.Name)
```
chaunceyjiang, line 218: https://github.com/karmada-io/karmada/blob/1d2028b10f9b9d9ca30fe8e963b4557d347c8f51/pkg/util/worker.go#L20-L21

https://github.com/karmada-io/karmada/blob/1d2028b10f9b9d9ca30fe8e963b4557d347c8f51/pkg/util/worker.go#L122-L128


Using the `AsyncWorker` pattern seems to have a problem.
 If a cluster is in the NotReady state for up to 2 hours, and during this period, a resource is deleted from Karmada. When the cluster recovers, it seems that the member resource will not be deleted.

XiShanYongYe-Chang: Yes, there was already such a problem before this modification. I think it is necessary for the administrator to intervene manually. Otherwise, we will have to retry indefinitely because the time it takes for users to fix the cluster can vary and cannot be determined.

lxtywypc: In fact, we have made the parameters about backoff or retry customized in our own environment.  
Maybe we could open another issue and invite @zach593 to talk about it.

chaunceyjiang: >  there was already such a problem before this modification.

`Controller Runtime` uses `workqueue.DefaultControllerRateLimiter`, which does not have a limit on the number of times. It will perform exponential backoff.

XiShanYongYe-Chang: Can we consider modifying the implementation of `asyncWorker`?

https://github.com/karmada-io/karmada/blob/bd0c511753a9f20a474af1777695483bf3e80d27/pkg/util/worker.go#L56

XiShanYongYe-Chang: Maybe we can open a new issue to trace this problem.

chaunceyjiang: > Can we consider modifying the implementation of asyncWorker?

For me, considering whether to use `asyncWorker` is the right thing to do. It's a significant change.

/cc @RainbowMango I can't make up my mind about this issue. Can you help me take another look at it? https://github.com/karmada-io/karmada/pull/3808#discussion_r1295679640

RainbowMango: > Yes, there was already such a problem before this modification.

@XiShanYongYe-Chang Can you show me the code where the problem exists? 

> For me, considering whether to use asyncWorker is the right thing to do. It's a significant change.

Yeah, I agree with you, I'm still thinking if there is a better way, such as using the reconciliation mechanism provided by controller-runtime.

XiShanYongYe-Chang: When using asyncWorker, if it fails to process a certain number of times due to its maximum retry limit, the event will be discarded.

https://github.com/karmada-io/karmada/blob/bd0c511753a9f20a474af1777695483bf3e80d27/pkg/controllers/status/work_status_controller.go#L131

https://github.com/karmada-io/karmada/blob/bd0c511753a9f20a474af1777695483bf3e80d27/pkg/controllers/status/work_status_controller.go#L224

The possible reason for the update error could be that the member cluster does not allow updates within a certain period of time, which exceeds the total retry time of the worker.

XiShanYongYe-Chang: Therefore, I think that this issue can be considered separately and not addressed in this PR.

RainbowMango: > Therefore, I think that this issue can be considered separately and not addressed in this PR.

I'm okay to address this problem with another PR, but shall we fix it before this PR?

My concern is that this problem is being magnified, before this PR it only affected the status collection(and the impact might automatically disappear by further events), but now it also affects the `Work` synchronization.

XiShanYongYe-Chang: IMOP, the scope of impact has not expanded, but rather the problem has shifted (the synchronization of work from resourceBinding is not applicable to error scenarios such as cluster failures), and the conditions under which the problem itself occurs are relatively strict. It requires a situation where cluster failures occur during the synchronization of work triggered by changes in member cluster resources. Therefore, the scope of impact for this issue is limited.

Therefore, I think that the issue being discussed can be advanced separately. In addition, the `asyncWorker` tool is currently widely used by us and its improvement measures need to be considered as a whole.

chaunceyjiang (commented):
Others LGTM, only this one. https://github.com/karmada-io/karmada/pull/3808/files#r1295679640.

XiShanYongYe-Chang:
> Others LGTM, only this one. https://github.com/karmada-io/karmada/pull/3808/files#r1295679640.

Thanks, @chaunceyjiang Regarding the issue you mentioned, I have already raised issue #3959 to track it. Let's discuss the solution there.

Let us merge it first.
/lgtm

PR #3812, opened 2023-07-19
make work name label value shorter

### LINKED ISSUE #3809: work may failed to create k8s resource due to label value over length limit
**What happened**:

The `work` created k8s resource, comes with a `work.karmada.io/name` label, which value is generated by this method, during `mergeLabel`, for back referencing work by k8s resource.

However, a name of a k8s object has a limit of length `253`, it's not likely someone will create such a long named object, since it's really hard to remember, and hard to type in. But, limit of label values is far shorter than limit of name length, which is `63`, and labels sometimes is generated by program, for referencing, not for typing or naming, it's easy to reach the limit.

And this happend in karmada.

https://github.com/karmada-io/karmada/blob/d27faeac953043c2274bb8787bbbaaafa7d0dcd6/pkg/util/names/names.go#L77-L99

image

Since name of k8s object didn't reach the limit of name (253), karmada should succefully propagate it to member cluster.

Propatate a k8s resource to member cluster with a name longer than 52 (plus `-0123456789` suffix will reach the `63` limit)

**Environment**:
- Karmada version:
- Others:

tedli:
**What type of PR is this?**

/kind bug

Long label value may reach the length limit `63`, blocks `work` applying kubernetes resource to member cluster.
We need to short the label value, yet makes the value distinct and readable.

Fixes #3809

short-name-xxxxxxxxxx, there a consistence hash suffix, so sort name together with suffix is unique, won't cause name conflict.

```release-note

```

### DIFF
--- a/pkg/util/names/names.go
+++ b/pkg/util/names/names.go
@@ -3,6 +3,7 @@ package names
 import (
 	"fmt"
 	"hash/fnv"
+	"regexp"
 	"strings"
 
 	"k8s.io/apimachinery/pkg/util/rand"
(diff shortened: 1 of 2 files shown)


pkg/util/names/names.go
```diff
@@ -95,7 +164,7 @@ func GenerateWorkName(kind, name, namespace string) string {
 	}
 	hash := fnv.New32a()
 	hashutil.DeepHashObject(hash, workName)
-	return fmt.Sprintf("%s-%s", name, rand.SafeEncodeString(fmt.Sprint(hash.Sum32())))
+	return fmt.Sprintf("%s-%s", ShortNameLabelValue(name), rand.SafeEncodeString(fmt.Sprint(hash.Sum32())))
```
jwcesign, line 167: Does the compatibility problem exist?  After upgrading Karmada, everything works right? like status syncing, and work syncing.

Also, Even if we truncate the name, the problem still exists, like the name is len(abcabacacb...abc)=63, so I prefer to move it to annotations.

tedli: `compatibility problem` do exist. `work` with name reached limit is very few, I use a script to delete such one or two work cr.
Then regenerated `work` works right.

Annotations is also a good place. I don't object.

**BUT**, **However** it happend in my **PRODUCTION**, I can't wait. With this pr, my production fixed.

Feel free to close this pr, if finally deside to use annotations.

RainbowMango: I will add it to my queue and will try to review it this week(or next week, too much on my plate recently.).

RainbowMango:
Hi @tedli , @jwcesign is analyzing this issue now, let's wait for a while to see if we have other solutions.

tedli:
Hi all,
I noticed that some native k8s controller, like the `StatefulSet` have the same issue.
The stateful set itself can had a name as long as 253, however, if the stateful set controller trying to create pod, it will fail.
It's because the stateful set controller will insert a label with key `controller-revision-hash` and value `<name-of-stateful-set>-<hash>` to pod's metadata, it reached the 63 length limit.

PR #3730, opened 2023-06-28
Bump Kubernetes dependencies to v1.27.3

RainbowMango:
**What type of PR is this?**

/kind cleanup

**What this PR does / why we need it**:

This PR updates Kubernetes dependencies to v1.27.3 as well as transitive dependencies and adoption:
- Update Kubernetes dependencies to v1.27.3
- Update controller-runtime to v0.15.0  (this is required during the Kubernetes dependencies update)
- Update cluster-api to v1.5.0-rc.0 (this is required during the controller-runtime update)
- Adopted Kubernetes changes https://github.com/kubernetes/kubernetes/pull/113542
- Adopted controller-runtime breaking change: https://github.com/kubernetes-sigs/controller-runtime/pull/2122
- Adopted controller-runtime breaking change: https://github.com/kubernetes-sigs/controller-runtime/pull/2296
- Adopted controller-runtime breaking change: https://github.com/kubernetes-sigs/controller-runtime/pull/2139
- Adopted controller-runtime breaking change: https://github.com/kubernetes-sigs/controller-runtime/pull/2135
- Adopted Kubernetes breaking change: https://github.com/kubernetes/kubernetes/pull/116218
- Adopted controller-runtime breaking change: https://github.com/kubernetes-sigs/controller-runtime/pull/2149
- Adopted controller-runtime breaking change: https://github.com/kubernetes-sigs/controller-runtime/pull/2150
- Adopted controller-runtime breaking change: https://github.com/kubernetes-sigs/controller-runtime/pull/2134
- Adopted controller-runtime breaking change: https://github.com/kubernetes-sigs/controller-runtime/pull/2293
- Adopted controller-runtime breaking change: https://github.com/kubernetes-sigs/controller-runtime/pull/2144
- Adopted Kubernetes breaking change: https://github.com/kubernetes/kubernetes/pull/115402
- Adopted Kubernetes breaking change: https://github.com/kubernetes/kubernetes/pull/117102
- Adopted controller-runtime breaking change: https://github.com/kubernetes-sigs/controller-runtime/pull/2316
- Adopted controller-runtime breaking change: https://github.com/kubernetes-sigs/controller-runtime/pull/2259

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
Karmada is now built with Kubernetes v1.27.3 dependencies.
```

PR #3819, opened 2023-07-20
build cached restmapper based on Kubernetes restmapper

RainbowMango:
**What type of PR is this?**

/kind cleanup

The reason why we introduced the [cachedRESTMapper](https://github.com/karmada-io/karmada/blob/d27faeac953043c2274bb8787bbbaaafa7d0dcd6/pkg/util/restmapper/restmapper.go#L25) is to improve performance, especially when retrieve the GVR as per a GVR, such as [here](https://github.com/karmada-io/karmada/blob/d27faeac953043c2274bb8787bbbaaafa7d0dcd6/pkg/detector/detector.go#L565).

We built the `cachedRESTMapper` [based on controller-runtime DynamicRESTMapper](https://github.com/karmada-io/karmada/blob/d27faeac953043c2274bb8787bbbaaafa7d0dcd6/pkg/util/restmapper/restmapper.go#L92-L106), but there is a breaking change happened in [v0.15.0](https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.15.0), and we can't build the `cachedRESTMapper` with it. This issue blocks the [updation of Kubernetes dependencies](https://github.com/karmada-io/karmada/pull/3730).

This PR tries to build the `cachedRESTMapper` with Kubernetes RESTMapper. This PR is a kind of predecessor task of #3730.

Fixes #

Before this change, the benchmark result:
```
# go test -bench=.
goos: linux
goarch: amd64
pkg: github.com/karmada-io/karmada/pkg/util/restmapper
cpu: Intel(R) Xeon(R) Gold 6278C CPU @ 2.60GHz
BenchmarkGetGroupVersionResource-4            	   62730	     17698 ns/op
BenchmarkGetGroupVersionResourceWithCache-4   	  924314	      1332 ns/op
PASS
ok  	github.com/karmada-io/karmada/pkg/util/restmapper	2.569s
-bash-5.0# pwd
/root/go/src/github.com/karmada-io/karmada/pkg/util/restmapper
```
The result shows that cachedRESTMapper has a significant performance advantage.

With this change:
```
# go test -bench=.
goos: linux
goarch: amd64
pkg: github.com/karmada-io/karmada/pkg/util/restmapper
cpu: Intel(R) Xeon(R) Gold 6278C CPU @ 2.60GHz
BenchmarkGetGroupVersionResource-4               	   59592	     20855 ns/op
BenchmarkGetGroupVersionResourceWithoutCache-4   	   67836	     18764 ns/op
BenchmarkGetGroupVersionResourceWithCache-4      	   41851	     30970 ns/op           // Seems weak, but it's not
```
The result seems to show that the performance after refactoring is weaker, but in fact, it's not. The performance of `cachedRESTMapper` appears lower because there is a [non-existent resource in the test suites](https://github.com/karmada-io/karmada/blob/d27faeac953043c2274bb8787bbbaaafa7d0dcd6/pkg/util/restmapper/restmapper_test.go#L98-L102), that cause `cachedRESTMapper` rebuild the underlying mapper. As a comparison, controller-runtime's restmapper doesn't rebuild so many times because it is [limited by the rate limiter](https://github.com/kubernetes-sigs/controller-runtime/blob/942d53bd982e54367ab6bfbf645ce7a1dab53f2f/pkg/client/apiutil/dynamicrestmapper.go#L193-L198). 

In Karmada, it is rare to query non-exist resources(unless a new CRD is created). After removing the non-exist resource test from the test suite, the result as follows:
```
# go test -bench=.
goos: linux
goarch: amd64
pkg: github.com/karmada-io/karmada/pkg/util/restmapper
cpu: Intel(R) Xeon(R) Gold 6278C CPU @ 2.60GHz
BenchmarkGetGroupVersionResource-4               	   72998	     16917 ns/op
BenchmarkGetGroupVersionResourceWithoutCache-4   	   64602	     16650 ns/op
BenchmarkGetGroupVersionResourceWithCache-4      	  823752	      1424 ns/op   
```
The performance advantage is still significant.

```release-note
NONE
```

### DIFF
--- a/pkg/util/restmapper/restmapper.go
+++ b/pkg/util/restmapper/restmapper.go
@@ -8,7 +8,6 @@ import (
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/restmapper"
-	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 )
 
 // GetGroupVersionResource is a helper to map GVK(schema.GroupVersionKind) to GVR(schema.GroupVersionResource).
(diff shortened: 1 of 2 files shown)


pkg/util/restmapper/restmapper.go
```diff
@@ -65,16 +67,49 @@ func (g *cachedRESTMapper) RESTMapping(gk schema.GroupKind, versions ...string)
...
+
+		newMapper := restmapper.NewDiscoveryRESTMapper(groupResources)
+		restMapping, err = newMapper.RESTMapping(gk, versions...)
+		if err == nil {
+			// hit here means after reloading, the new rest mapper can recognize
+			// the resource, we have to replace the mapper and clear cache.
+			g.mu.Lock()
+			g.restMapper = newMapper
+			g.mu.Unlock()
+			g.gvkToGVR.Range(func(key, value any) bool {
```
chaunceyjiang, line 90: Just a question.Why not directly use `g.gvkToGVR.Delete(key)`?

RainbowMango: The logic here is going to `clear` the whole cache, not just remove a specific key. 
For example, after removing a CRD, the key may remain in the cache(gvkToGVR), so the idea is to clear the cache and then rebuild the cache with the latest loaded resources.

RainbowMango:
cc @Poor12 @likakuli for comments.

PR #3822, opened 2023-07-20
feat: add proposal for cluster proxy enhancement

jwcesign:
**What type of PR is this?**
/kind design
/kind documentation

The single cluster users usually use `kubectl` to query resources from member clusters separately. So after migrating to Karmada with multiple clusters, it will be great if they still can use `kubectl` to query resources from member clusters and unaware the member clusters, maintain the same experience as when in a cluster.

Also, Karmada is usually used in multi-tenant environment, user A cloud not access to the resource of user B. This is always achieved by using RBAC. Which means the query operation should follow the RBAC privilege limitations in member clusters.

So this proposal proposes an enhancement to query resources from member clusters with native `kubectl`, unaware the member clusters, and follow RBAC privilege limitations in member clusters.

With this enhancement, users can feel the lease changes after migrating to Karmada, but enjoy the benefits of multiple clusters.

Fixes #none

```release-note

```


docs/proposals/cluster-proxy/clusterproxy.md
```diff
@@ -0,0 +1,263 @@
...
+### Notes/Constraints/Caveats (Optional)
+
+<!--
+What are the caveats to the proposal?
+What are some important details that didn't come across above?
+Go in to as much detail as necessary here.
+This might be a good place to talk about core concepts and how they relate.
+-->
+
+- If there are resources with the same name in different member clusters, operations such as `get`, `describe`, `exec`, and `logs` will fail.
```
zishen, line 131: How about add a external parameter like "--c clustername' to specify the cluster? If not point the multiple resource objects  confusion .

PR #3823, opened 2023-07-20
add wildcard support to `ResourceSelectors`

Laevos:
**What type of PR is this?**

/kind feature

Currently a `ClusterPropagationPolicy` is able to propagate both namespaced and cluster-wide resources to member clusters. However, there is not currently a way to say "I would like for all resources, both namespaced and cluster-wide, to be propagated according to this `ClusterPropagationPolicy`, regardless of their `APIVersion` or `Kind`, and in all `Namespaces` (except for the system-reserved namespaces)."

With this change, a wildcard feature is added to allow a `ClusterPropagationPolicy` to do just that. By Providing a "*" for any combination of the `APIVersion`, `Kind` and `Namespace`, a `ClusterPropagationPolicy` can be made to match on a wider array of 

Fixes #

```release-note
Adds a wildcard parsing feature to the ResourceSelector, allowing more flexible matching behavior.
```

### DIFF
--- a/pkg/util/selector.go
+++ b/pkg/util/selector.go
@@ -33,9 +33,9 @@ func ResourceMatches(resource *unstructured.Unstructured, rs policyv1alpha1.Reso
 
 // ResourceSelectorPriority tells the priority between the specific resource and the selector.
 func ResourceSelectorPriority(resource *unstructured.Unstructured, rs policyv1alpha1.ResourceSelector) ImplicitPriority {
-	if resource.GetAPIVersion() != rs.APIVersion ||
-		resource.GetKind() != rs.Kind ||
-		(len(rs.Namespace) > 0 && resource.GetNamespace() != rs.Namespace) {
+	if (resource.GetAPIVersion() != rs.APIVersion && rs.APIVersion != "*") ||
+		(resource.GetKind() != rs.Kind && rs.Kind != "*") ||
+		(len(rs.Namespace) > 0 && resource.GetNamespace() != rs.Namespace && rs.Namespace != "*") {
 		return PriorityMisMatch
 	}
 
--- a/pkg/util/selector_test.go
+++ b/pkg/util/selector_test.go
@@ -170,6 +170,17 @@ func TestResourceSelectorPriority(t *testing.T) {
 			want: PriorityMatchName,
 		},
 		{
+			name: "wildcard namespace matched",
+			args: args{
+				rs: policyv1alpha1.ResourceSelector{
+					APIVersion: "*",
+					Kind:       "*",
+					Namespace:  "*",
+				},
+			},
+			want: PriorityMatchAll,
+		},
+		{
 			name: "[case 1] name not matched",
 			args: args{
 				rs: policyv1alpha1.ResourceSelector{


XiShanYongYe-Chang:
Hi @Laevos, thanks for your feedback. If we use `*` for exact matching, we may not be able to select which resources will not be propagated, which could pose security issues as described in the comments:

https://github.com/karmada-io/karmada/blob/48e9a5b55163844b11b6ead2bec32076ba63449a/pkg/apis/policy/v1alpha1/propagation_types.go#L44-L50

How do you think?

MingZhang-YBPS:
Currently, ResourceSelectors supports selecting one instance or all instances of one kind. That seems good enough

RainbowMango:
There is a similar request that is for `namespace` part. #3656 
I'm looking at them now.

RainbowMango:
> However, there is not currently a way to say "I would like for all resources, both namespaced and cluster-wide, to be propagated according to this ClusterPropagationPolicy, regardless of their APIVersion or Kind, and in all Namespaces (except for the system-reserved namespaces)."

I'd like to hear your use case in more detail. We need to figure out the security risk before moving forward. 
Currently, Karmada's API server is essentially a `kube-apiserver` which will create some resources like Secrets that might not allowed to propagate.

Laevos:
> > However, there is not currently a way to say "I would like for all resources, both namespaced and cluster-wide, to be propagated according to this ClusterPropagationPolicy, regardless of their APIVersion or Kind, and in all Namespaces (except for the system-reserved namespaces)."
> 
> I'd like to hear your use case in more detail. We need to figure out the security risk before moving forward. Currently, Karmada's API server is essentially a `kube-apiserver` which will create some resources like Secrets that might not allowed to propagate.

Sure thing @RainbowMango: Our use case is that we are a platform team who provides managed Kubernetes clusters to our internal customers. However, one business requirement we have is that we cannot request our end uses to change the way they're currently writing manifests in any significant way. Namely, we can't ask that teams add PropagationPolicies to their clusters, as those should be managed by us. At the same time, we would like to avoid having to manage every individual `PropagationPolicy` and `ClusterPropagationPolicy` by hand. 

With our current architecture, we have only one member cluster, to which we want to propagate all manifests deployed against the Karmada cluster. The thought here was that by having a wildcard `ClusterPropagationPolicy`, we could do that as simply as possible. My understanding was that the `PropagationPolicy` controller already stopped some namespaces from propagating (namely, Karmada's system-reserved namespaces), is that incorrect?

An alternative solution is to create a `ResourceSelector` for each individual resource that needs to be propagated (`Deployment`, `Service`, etc.), but we'd prefer to avoid having to be aware of and manage a ResourceSelector for every CRD that our customers use.

XiShanYongYe-Chang:
Is it possible to consider using multiple PropagationPolicies, each selecting a specific namespace of a certain Kind? This way, compared to using a single ClusterPropagationPolicy, the control would be more precise. Additionally, the number of PropagationPolicies that administrators need to create is fixed and at a constant level.

Laevos:
@XiShanYongYe-Chang This is looking more like what we'll have to do, or possibly implement a controller in the future; the idea was to try and scope to as few PPs/CPPs as possible to avoid toilsome work of manually updating when a new Kind/CRD is introduced, but I appreciate the direction; it looks like this can be closed if the consensus is that this would be an anti-feature. Thanks for your time!

PR #3829, opened 2023-07-21
feat: validate resourceSelectors if Preemption is enabled

whitewindmills:
**What type of PR is this?**
/kind feature

validate resourceSelectors if Preemption is enabled

part of [feature](https://github.com/karmada-io/karmada/issues/3787)

```release-note
NONE
```


pkg/apis/policy/v1alpha1/propagation_types.go
```diff
@@ -155,6 +155,8 @@ type ResourceSelector struct {
 
 	// Name of the target resource.
 	// Default is empty, which means selecting all resources.
+	// For PropagationPolicy(ClusterPropagationPolicy), it cannot be empty
+	// if Preemption is "Always".
```
RainbowMango, line 159: I'd prefer to postpone adding this description(restriction) to ResourceSelector. 
This restriction is temporary and might be removed in the future.

pkg/util/validation/validation.go
```diff
@@ -21,6 +21,9 @@ const LabelValueMaxLength int = 63
 // ValidatePropagationSpec validates a PropagationSpec before creation or update.
 func ValidatePropagationSpec(spec policyv1alpha1.PropagationSpec) field.ErrorList {
 	var allErrs field.ErrorList
+	if spec.Preemption == policyv1alpha1.PreemptAlways {
+		allErrs = append(allErrs, ValidateResourceSelectors(spec.ResourceSelectors, field.NewPath("spec").Child("resourceSelectors"))...)
+	}
```
RainbowMango, line 26: I'd like to validate the most common case first, I mean put this logic to the end of this function.

pkg/util/validation/validation.go
```diff
@@ -29,6 +32,18 @@ func ValidatePropagationSpec(spec policyv1alpha1.PropagationSpec) field.ErrorLis
 	return allErrs
 }
 
+// ValidateResourceSelectors validates all resourceSelectors before creation or update.
+func ValidateResourceSelectors(resourceSelectors []policyv1alpha1.ResourceSelector, fldPath *field.Path) field.ErrorList {
```
RainbowMango, line 36: The function name doesn't match its responsibility. This function is actually focused on a limited scenario and is not general-purpose.

Poor12, line 36: For cpp, we need to validate the namespace together with name if it matches a namespace-scoped resource.

whitewindmills: It has a certain impact on performance because we need to query `APIResourceList` every time we verify, do we have to do this?

whitewindmills: when CRD is not installed, how to check if it is a namespaced resource? people can choose to create propagation policies first, then install CRDs and propagate resources when needed.
cc @RainbowMango for comments.

pkg/util/validation/validation.go
```diff
@@ -26,6 +26,15 @@ func ValidatePropagationSpec(spec policyv1alpha1.PropagationSpec) field.ErrorLis
 		allErrs = append(allErrs, field.Invalid(field.NewPath("spec").Child("propagateDeps"), spec.PropagateDeps, "application failover is set, propagateDeps must be true"))
 	}
 	allErrs = append(allErrs, ValidateFailover(spec.Failover, field.NewPath("spec").Child("failover"))...)
+	// validate ResourceSelectors if Preemption is Always.
+	if spec.Preemption == policyv1alpha1.PreemptAlways {
+		rsField := field.NewPath("spec").Child("resourceSelectors")
```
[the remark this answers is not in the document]
whitewindmills: thanks

pkg/util/validation/validation.go
```diff
@@ -26,6 +26,15 @@ func ValidatePropagationSpec(spec policyv1alpha1.PropagationSpec) field.ErrorLis
 		allErrs = append(allErrs, field.Invalid(field.NewPath("spec").Child("propagateDeps"), spec.PropagateDeps, "application failover is set, propagateDeps must be true"))
 	}
 	allErrs = append(allErrs, ValidateFailover(spec.Failover, field.NewPath("spec").Child("failover"))...)
+	// validate ResourceSelectors if Preemption is Always.
+	if spec.Preemption == policyv1alpha1.PreemptAlways {
```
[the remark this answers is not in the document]
whitewindmills: +1

pkg/util/validation/validation.go
```diff
@@ -26,6 +26,15 @@ func ValidatePropagationSpec(spec policyv1alpha1.PropagationSpec) field.ErrorLis
...
+		rsField := field.NewPath("spec").Child("resourceSelectors")
+		for index, resourceSelector := range spec.ResourceSelectors {
+			if len(resourceSelector.Name) == 0 {
+				allErrs = append(allErrs, field.Invalid(rsField.Index(index).Child("name"), resourceSelector.Name, "name can not be empty if preemption is Always"))
```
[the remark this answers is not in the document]
whitewindmills: seems reasonable.

jwcesign (commented):
/lgtm

pkg/util/validation/validation.go
RainbowMango: > I think there is not much diff in performance because if it's a namespace-scope resource, you need to traverse a namespace in the end if you don't fill in.

I didn't get it. Maybe @Poor12 can explain more.

pkg/util/validation/validation.go
```diff
@@ -26,6 +26,22 @@ func ValidatePropagationSpec(spec policyv1alpha1.PropagationSpec) field.ErrorLis
...
+
+// validateResourceSelectorsIfPreemptionEnabled validates ResourceSelectors if Preemption is Always.
+func validateResourceSelectorsIfPreemptionEnabled(spec policyv1alpha1.PropagationSpec, fldPath *field.Path) field.ErrorList {
+	if spec.Preemption != policyv1alpha1.PreemptAlways {
+		return nil
+	}
+
+	var allErrs field.ErrorList
+	for index, resourceSelector := range spec.ResourceSelectors {
+		if len(resourceSelector.Name) == 0 {
```
RainbowMango, line 41: Maybe we can validate the `namespace` as well. 
Given all `PropagationPolicy` will have the namespace.
https://github.com/karmada-io/karmada/blob/6ef427a98e4c669c6834f0e62a400876c4b7a1c3/pkg/webhook/propagationpolicy/mutating.go#L47-L56

RainbowMango: Oh, no, please ignore me. 
We can't do that because for `ClusterPropagationPolicy`, the namespace might be empty even in the case of `preemption: Always`.

pkg/util/validation/validation.go
whitewindmills: @Poor12 means that if the namespace of `ResourceSelector` unspecified for `CPP`, namespaced resource with the same name under different namespaces may be preempted during policy preemption. If I am wrong, please @Poor12 correct me.

> when CRD is not installed, how to check if it is a namespaced resource? people can choose to create propagation policies first, then install CRDs and propagate resources when needed.

But I think there are still some problems with doing this, I don't know if it‘s necessary to verify.

RainbowMango (approved):
/approve

PR #3830, opened 2023-07-22
Fix: Denied pp create request for association

liangyuanpeng:
**What type of PR is this?**
/kind bug

association  is not used anymore, and it's not working now. Denied create request of pp to avoid user wrong usage  this field.

/cc @RainbowMango 

Related:
- #3679

Fixes #

```release-note
Denied Clusterpropagationpolicy& Propagationpolicy creqte quest for association.
```

### DIFF
--- a/pkg/util/validation/validation.go
+++ b/pkg/util/validation/validation.go
@@ -25,6 +25,9 @@ func ValidatePropagationSpec(spec policyv1alpha1.PropagationSpec) field.ErrorLis
 	if spec.Failover != nil && spec.Failover.Application != nil && !spec.PropagateDeps {
 		allErrs = append(allErrs, field.Invalid(field.NewPath("spec").Child("propagateDeps"), spec.PropagateDeps, "application failover is set, propagateDeps must be true"))
 	}
+	if spec.Association {
+		allErrs = append(allErrs, field.Invalid(field.NewPath("spec").Child("association"), spec.PropagateDeps, "spec.association has been deprecated, use PropagateDeps instead."))
+	}
 	allErrs = append(allErrs, ValidateFailover(spec.Failover, field.NewPath("spec").Child("failover"))...)
 	return allErrs
 }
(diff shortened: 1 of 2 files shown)


pkg/webhook/propagationpolicy/validating.go
```diff
@@ -46,6 +46,13 @@ func (v *ValidatingAdmission) Handle(_ context.Context, req admission.Request) a
 			return admission.Denied(err.Error())
 		}
 	}
+	if req.Operation == admissionv1.Create {
+ 		if policy.Spec.Association {
+			err = fmt.Errorf("spec.association is deprecated, use PropagateDeps instead.")
```
RainbowMango, line 51:
```suggestion
			err = fmt.Errorf("spec.association has been deprecated, use PropagateDeps instead.")
```
By the way, we might need to apply this restriction to `ClusterPropagationPolicy` as well.

pkg/webhook/clusterpropagationpolicy/validating.go
```diff
@@ -46,6 +46,9 @@ func (v *ValidatingAdmission) Handle(_ context.Context, req admission.Request) a
 			return admission.Denied(err.Error())
 		}
 	}
+	if req.Operation == admissionv1.Create && policy.Spec.Association {
```
RainbowMango, line 49: What if the operation is `Update`? 

I suppose we can add the validation into `ValidatePropagationSpec()`. Sorry, I didn't point that out during the last review.

RainbowMango: > It will bother users who have already used Association, if you use 100 pp that use Association, then you need to modify these pp.

Good point! Just realized that this kind of a breaking change.

cc @chaunceyjiang for comments, do we really need to validate the never-used/deprecated field?

liangyuanpeng: I prefer to only verify the unused field of association in the create request, so that the user knows what he should use. At the same time, it reduces the interruption to existing users.

I used to cost some time on this field, and finally found the answer in github PR. “ohh,  it's not used field, not just deprecated”

During my K8S journey, the prompt is deprecated but the field should continue to work, unless the user is explicitly told that the field cannot be used.

RainbowMango: Can we have more clear comments on the field? Given the field was never been used, I guess nobody would expect it will work.

liangyuanpeng: Make sense, just clean the comment of field. I will create the pr to work for it. close this pr first.

PR #3836, opened 2023-07-25
color unit-test output

lxtywypc:
**What type of PR is this?**  
/kind feature  

Like https://github.com/golang/go/issues/46921  

It is a little hard to find out which failed in a failed unit-test job.  

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,12 +102,25 @@ push-chart:
 	helm push _output/charts/karmada-chart-${VERSION}.tgz oci://docker.io/karmada
 	helm push _output/charts/karmada-operator-chart-${VERSION}.tgz oci://docker.io/karmada
 
+COLOR_GOTEST_REGISTRY:=github.com/rakyll/gotest
+COLOR_GOTEST_VERSION:=aeb9f1f4739020c60963f21eec2e65672307a9ac
+COLOR_GOTEST_ENABLED?=
+GOTEST_PALETTE?=hired,higreen
+GOTEST=go test
+
+.PHONY: install_gotest
+install_gotest:
+ifdef COLOR_GOTEST_ENABLED 
+	go install ${COLOR_GOTEST_REGISTRY}@${COLOR_GOTEST_VERSION}
+GOTEST=gotest
+endif
+
 .PHONY: test
-test:
+test: install_gotest
 	mkdir -p ./_output/coverage/
-	go test --race --v ./pkg/... -coverprofile=./_output/coverage/coverage_pkg.txt -covermode=atomic
-	go test --race --v ./cmd/... -coverprofile=./_output/coverage/coverage_cmd.txt -covermode=atomic
-	go test --race --v ./examples/... -coverprofile=./_output/coverage/coverage_examples.txt -covermode=atomic
+	$(GOTEST) --race --v ./pkg/... -coverprofile=./_output/coverage/coverage_pkg.txt -covermode=atomic
+	$(GOTEST) --race --v ./cmd/... -coverprofile=./_output/coverage/coverage_cmd.txt -covermode=atomic
+	$(GOTEST) --race --v ./examples/... -coverprofile=./_output/coverage/coverage_examples.txt -covermode=atomic
 
 upload-images: images
 	@echo "push images to $(REGISTRY)"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,6 +80,8 @@ jobs:
     name: unit test
     needs: build
     runs-on: ubuntu-22.04
+    env:
+      COLOR_GOTEST_ENABLED: enabled
     steps:
       - name: checkout code
         uses: actions/checkout@v3


RainbowMango:
Interesting!
https://github.com/karmada-io/karmada/actions/runs/5655883775/job/15322147424?pr=3836
Will look at it later.

RainbowMango (commented):
I've looked into it and tried it on my side as well. It looks great, but concerned about the compatibility of the https://github.com/rakyll/gotest:

As per the [gotest Usage](https://github.com/rakyll/gotest#usage):
```
Accepts all the arguments and flags go test works with.
```
Given `go test` flags might change over different releases, I'm not sure can `rakyll/gotest` handle that.
In addition, `rakyll/gotest` seems lack of activities now, the last commit happened a year ago.

Overall, I'm hesitant to move forward with it.

RainbowMango:
> It is a little hard to find out which failed in a failed unit-test job.

+1

I agree with you. 

We usually focus on which test is failing, my idea is to introduce `-failfast` flag. 
```
# go help testflag
	-failfast
	    Do not start new tests after the first test failure.
```

lxtywypc:
But it seems not work for tests with `./...`.  

I made some explicit mistake in a test file, and got the result as following:  

```
$ go test --failfast ./pkg/util
--- FAIL: TestMergeAnnotation (0.00s)
    --- FAIL: TestMergeAnnotation/nil_annotations (0.00s)
        annotation_test.go:118: MergeAnnotation() = &{map[apiVersion:apps/v1 kind:Deployment metadata:map[annotations:map[foo:bar] name:demo-deployment] spec:map[replicas:2]]}, want &{map[apiVersion:apps/v1 kind:Deployment metadata:map[annotations:map[foo:bar] name:demo-deployment] spec:map[replicas:1]]}
FAIL
FAIL    github.com/karmada-io/karmada/pkg/util  0.024s
FAIL
```
```
$ go test --failfast ./pkg/util/...
?       github.com/karmada-io/karmada/pkg/util/fedinformer/genericmanager       [no test files]
?       github.com/karmada-io/karmada/pkg/util/fedinformer/typedmanager [no test files]
?       github.com/karmada-io/karmada/pkg/util/gclient  [no test files]
?       github.com/karmada-io/karmada/pkg/util/hash     [no test files]
--- FAIL: TestMergeAnnotation (0.00s)
    --- FAIL: TestMergeAnnotation/nil_annotations (0.00s)
        annotation_test.go:118: MergeAnnotation() = &{map[apiVersion:apps/v1 kind:Deployment metadata:map[annotations:map[foo:bar] name:demo-deployment] spec:map[replicas:2]]}, want &{map[apiVersion:apps/v1 kind:Deployment metadata:map[annotations:map[foo:bar] name:demo-deployment] spec:map[replicas:1]]}
FAIL
FAIL    github.com/karmada-io/karmada/pkg/util  0.082s
ok      github.com/karmada-io/karmada/pkg/util/fedinformer      0.035s
ok      github.com/karmada-io/karmada/pkg/util/fedinformer/keys 0.025s
ok      github.com/karmada-io/karmada/pkg/util/fixedpool        0.073s
?       github.com/karmada-io/karmada/pkg/util/lifted/scheduler/cache   [no test files]
?       github.com/karmada-io/karmada/pkg/util/lifted/scheduler/framework       [no test files]
?       github.com/karmada-io/karmada/pkg/util/lifted/scheduler/util    [no test files]
?       github.com/karmada-io/karmada/pkg/util/objectwatcher    [no test files]
ok      github.com/karmada-io/karmada/pkg/util/helper   0.836s
ok      github.com/karmada-io/karmada/pkg/util/imageparser      0.020s
ok      github.com/karmada-io/karmada/pkg/util/interpreter      0.065s
ok      github.com/karmada-io/karmada/pkg/util/lifted   0.047s
ok      github.com/karmada-io/karmada/pkg/util/lifted/pubkeypin 0.029s
ok      github.com/karmada-io/karmada/pkg/util/lifted/scheduler/framework/parallelize   0.029s
ok      github.com/karmada-io/karmada/pkg/util/lifted/selectors 0.035s
ok      github.com/karmada-io/karmada/pkg/util/metrics  0.012s
ok      github.com/karmada-io/karmada/pkg/util/names    0.013s
ok      github.com/karmada-io/karmada/pkg/util/overridemanager  0.058s
?       github.com/karmada-io/karmada/pkg/util/testing  [no test files]
ok      github.com/karmada-io/karmada/pkg/util/proxy    0.040s
ok      github.com/karmada-io/karmada/pkg/util/restmapper       0.021s
ok      github.com/karmada-io/karmada/pkg/util/validation       0.034s
FAIL
```  

And another problem is that there may be several failings in one test, if we set `-failfast`, we may have to run test for several times to find all of them out.

RainbowMango:
> But it seems not work for tests with ./...

Yes, that's indeed an annoying thing.

> And another problem is that there may be several failings in one test, if we set -failfast, we may have to run test for several times to find all of them out.

That good point, but usually there won't be a large number of cases failing at the same time, so it might be acceptable I guess.

lxtywypc:
https://github.com/rakyll/gotest/blob/master/main.go#L53-L65  

It seems that it just call original `go test` and modified the result given out. Maybe we do not need to consider about compatibility. 

@RainbowMango   
What do you think about?

RainbowMango:
> https://github.com/rakyll/gotest/blob/master/main.go#L53-L65

Good to know. Thanks.

Can we do this not by default? Such as having an environment to indicate if enable the gotest, we can start to use it in CI.

lxtywypc:
> Can we do this not by default? Such as having an environment to indicate if enable the gotest, we can start to use it in CI.  

Of course. I have made some modifications. Now we could set CI env `COLOR_GOTEST_ENABLED` `=` `[any value not empty]` to enable the colorful go test. :)

Makefile
```diff
@@ -102,12 +102,26 @@ push-chart:
...
+install_gotest:
+ifdef COLOR_GOTEST_ENABLED 
+	go install ${COLOR_GOTEST_REGISTRY}@${COLOR_GOTEST_VERSION}
+GOTEST=gotest
+endif
+
 .PHONY: test
-test:
+test:export GOTEST_PALETTE=${COLOR_GOTEST_PALETTE}
+test: install_gotest
```
RainbowMango, line 120: I'm not familiar with this usage, why use two targets here?

lxtywypc: The first one is to export the environment only in `test`.  

It seems not necessary to do this. I have removed it, and now it could be set straightly as an environment.  

Now the performance would be like:  
- If `COLOR_GOTEST_ENABLED` environment is not set, `make test` would use `go test` straightly.  
- If `COLOR_GOTEST_ENABLED` is set but `GOTEST_PALETTE` is not set, `make test` would color the output using `hired` for `FAIL` and `higreen` for `PASS` in default.  
- If both `COLOR_GOTEST_ENABLED` and `GOTEST_PALETTE` are set, `make test` would color the output by the declared color by `GOTEST_PALETTE` value, the value is `<color for FAIL>,<color for PASS>`

RainbowMango: I can't wait to enable it in CI, I guess what we need to do is:
```diff
diff --git a/.github/workflows/ci.yml b/.github/workflows/ci.yml
index 45a4982fb..2c321e4fb 100644
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,7 +88,9 @@ jobs:
         with:
           go-version: 1.20.4
       - name: make test
-        run: make test
+        run: |
+          export COLOR_GOTEST_ENABLED=true
+          make test
       - name: Upload coverage to Codecov
         # Prevent running from the forked repository that doesn't need to upload coverage.
         # In addition, running on the forked repository would fail as missing the necessary secret.
```

Can you help to add this in this PR? so that other guys can see the magic :)

lxtywypc: Done.  

```diff
diff --git a/.github/workflows/ci.yml b/.github/workflows/ci.yml
index 45a4982f..4daf2c55 100644
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,6 +80,8 @@ jobs:
     name: unit test
     needs: build
     runs-on: ubuntu-22.04
+    env:
+      COLOR_GOTEST_ENABLED: enabled
     steps:
       - name: checkout code
         uses: actions/checkout@v3
```

We just need to add an environment for `unit test` job. Hope you enjoy the colorful UT! :)  
https://github.com/karmada-io/karmada/actions/runs/5700866002/job/15451158398

lxtywypc:
Kindly ping @RainbowMango @XiShanYongYe-Chang @jwcesign to take a review again. :)

RainbowMango (approved):
/lgtm
/approve

Thanks.

Issue #3413, opened 2023-04-17
How does Karmada handle the multi-AZ cluster


MingZhang-YBPS:
**Please provide an in-depth description of the question you have**:
There is a case that Kubernetes may cross 3 AZ in order to the cluster high availble in cloud environment, https://kubernetes.io/docs/setup/best-practices/multiple-zones/
However, Karmada now only support mark a cluster with one AZ, how can we make the scheduling works in such 3AZ setup?

**What do you think about this question?**:

**Environment**:
- Karmada version:
- Kubernetes version:
- Others:

XiShanYongYe-Chang:
Hi @MingZhang-YBPS, thanks for your feedback, If a cluster is highly available in multiple AZs, can we directly set the region of the cluster without setting the AZ?

RainbowMango:
> However, Karmada now only support mark a cluster with one AZ, how can we make the scheduling works in such 3AZ setup?

Thanks for spotting! 
https://github.com/karmada-io/karmada/blob/6687479543631e781f2e58371cb09c42aab70753/pkg/apis/cluster/v1alpha1/types.go#L131-L133

The `Zone` now takes a `String`, it is inconvenient for the `multiple zones` cases.
Actually the relevant functionality, I mean spread by zones, is not ready for the time being. We were looking forward to implementing it based on real scenarios.

I think it's a good time to talk about it. Could you share your use case? 
cc people who might be interested in this. @whitewindmills 

By the way, we might introduce another field like `Zones []string` to deprecate the previous one, or let `Zone` accept comma-separated zones, like `zone1,zone2,zone3`.

whitewindmills:
> or let Zone accept comma-separated zones, like zone1,zone2,zone3.

It's a good idea to maintain backward compatibility.

MingZhang-YBPS:
@RainbowMango @XiShanYongYe-Chang @whitewindmills 
let me explain this use case in detail.
![image](https://user-images.githubusercontent.com/74715700/232650731-240198dc-e7a9-43b6-937f-10bc6e1893a7.png)


We try to build a multicluster setup in single region environment. Currently, there is a hub cluster in AZ1, and it manages members in multiple AZ. Some central services developed by ourselves and Karmada control plane are deployed in hub cluster. Karmada functionality supports this setup very well.

However, customer think the hub cluster in single zone is not robust enough. So we try to leverage kubernetes multizone architecture, and split nodes in hub cluster into more than 3 AZ. 
[...]

PR #3844, opened 2023-07-27
cluster registried in pull mode doesn't contains zone field

MingZhang-YBPS:
**What type of PR is this?**
/kind bug

User want to registry cluster in pull mode with region as well as zone field

Fixes #

NONE

```release-note

```

### DIFF
--- a/cmd/agent/app/agent.go
+++ b/cmd/agent/app/agent.go
@@ -139,6 +139,7 @@ func run(ctx context.Context, opts *options.Options) error {
 		ProxyServerAddress: opts.ProxyServerAddress,
 		ClusterProvider:    opts.ClusterProvider,
 		ClusterRegion:      opts.ClusterRegion,
+		ClusterZone:        opts.ClusterZone,
 		DryRun:             false,
 		ControlPlaneConfig: controlPlaneRestConfig,
 		ClusterConfig:      clusterConfig,
--- a/cmd/agent/app/options/options.go
+++ b/cmd/agent/app/options/options.go
@@ -109,6 +109,9 @@ type Options struct {
 	// ClusterRegion represents the region of the cluster locate in.
 	ClusterRegion string
 
+	// ClusterZone represents the zone of the cluster locate in.
+	ClusterZone string
+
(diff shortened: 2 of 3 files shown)


RainbowMango:
Hi @MingZhang-YBPS Does this PR related to https://github.com/karmada-io/karmada/issues/3413?

MingZhang-YBPS:
> Hi @MingZhang-YBPS Does this PR related to #3413?

No, it is not strongly related to multi-az scheduling. 
I just want to make push and pull behavior to be consistent since we are enabling pull mode cluster for project recently

MingZhang-YBPS:
> This PR reminds me of #2152.
> 
> Can you share your use case with the `zone`? Do you still need this once we introduce `zones` as per #3413?

The usage of `zone ` used in pull mode is the same as what we use `zone ` in push mode, this field is persisted into cluster object for two purposes afterwards
1. To to used in PP/CPP fieldselector
2. To be used for zone based spreadconstraint once Karmada supports it in future

And back to #2152  #3413 , `zones` field in cluster object is certainly a more flexiable choice compared with `zone`. It will be quite nice if karmada can provide some data migration tool to move values for old version cluster CR to the new version during components rollingupdate.

RainbowMango:
Given we are going to introduce `zones`, and after that, the previous [Zone](https://github.com/karmada-io/karmada/blob/f160ea785418fbfb28a1a5db95e55fd4121f30ca/pkg/apis/cluster/v1alpha1/types.go#L131-L133) might be deprecated, so my idea is to hold this PR for a while and get back after #3413. Maybe the flags we need would be `--cluster-zones` instead of `--cluster-zone`. What do you think?

MingZhang-YBPS:
> Given we are going to introduce `zones`, and after that, the previous [Zone](https://github.com/karmada-io/karmada/blob/f160ea785418fbfb28a1a5db95e55fd4121f30ca/pkg/apis/cluster/v1alpha1/types.go#L131-L133) might be deprecated, so my idea is to hold this PR for a while and get back after #3413. Maybe the flags we need would be `--cluster-zones` instead of `--cluster-zone`. What do you think?

It is OK

PR #3845, opened 2023-07-27
feat: implement preemption between propagation policies

whitewindmills:
**What type of PR is this?**
/kind feature

implement preemption between propagation policies

part of [feature](https://github.com/karmada-io/karmada/issues/3787)

```release-note
NONE
```

### DIFF
--- a/pkg/detector/detector.go
+++ b/pkg/detector/detector.go
@@ -422,6 +422,7 @@ func (d *ResourceDetector) ApplyPolicy(object *unstructured.Unstructured, object
 			bindingCopy.Spec.Placement = binding.Spec.Placement
 			bindingCopy.Spec.Failover = binding.Spec.Failover
 			bindingCopy.Spec.ConflictResolution = binding.Spec.ConflictResolution
+			excludeClusterPolicy(bindingCopy.Labels)
 			return nil
 		})
 		if err != nil {
@@ -598,18 +599,23 @@ func (d *ResourceDetector) GetUnstructuredObject(objectKey keys.ClusterWideKey) 
 
 // ClaimPolicyForObject set policy identifier which the object associated with.
 func (d *ResourceDetector) ClaimPolicyForObject(object *unstructured.Unstructured, policyNamespace string, policyName string) error {
-	claimedNS := util.GetLabelValue(object.GetLabels(), policyv1alpha1.PropagationPolicyNamespaceLabel)
-	claimedName := util.GetLabelValue(object.GetLabels(), policyv1alpha1.PropagationPolicyNameLabel)
-
-	// object has been claimed, don't need to claim again
-	if claimedNS == policyNamespace && claimedName == policyName {
-		return nil
+	objLabels := object.GetLabels()
+	if objLabels == nil {
+		objLabels = make(map[string]string)
+	} else if len(objLabels) > 0 {
+		// object has been claimed, don't need to claim again
+		if !excludeClusterPolicy(objLabels) &&
+			objLabels[policyv1alpha1.PropagationPolicyNamespaceLabel] == policyNamespace &&
+			objLabels[policyv1alpha1.PropagationPolicyNameLabel] == policyName {
+			return nil
+		}
 	}
 
-	util.MergeLabel(object, policyv1alpha1.PropagationPolicyNamespaceLabel, policyNamespace)
-	util.MergeLabel(object, policyv1alpha1.PropagationPolicyNameLabel, policyName)
-
-	return d.Client.Update(context.TODO(), object)
+	objLabels[policyv1alpha1.PropagationPolicyNamespaceLabel] = policyNamespace
+	objLabels[policyv1alpha1.PropagationPolicyNameLabel] = policyName
+	objectCopy := object.DeepCopy()
+	objectCopy.SetLabels(objLabels)
+	return d.Client.Update(context.TODO(), objectCopy)
 }
 
 // ClaimClusterPolicyForObject set cluster identifier which the object associated with.
@@ -621,8 +627,9 @@ func (d *ResourceDetector) ClaimClusterPolicyForObject(object *unstructured.Unst
 		return nil
 	}
 
-	util.MergeLabel(object, policyv1alpha1.ClusterPropagationPolicyLabel, policyName)
-	return d.Client.Update(context.TODO(), object)
+	objectCopy := object.DeepCopy()
+	util.MergeLabel(objectCopy, policyv1alpha1.ClusterPropagationPolicyLabel, policyName)
+	return d.Client.Update(context.TODO(), objectCopy)
 }
 
 // BuildResourceBinding builds a desired ResourceBinding for object.
--- a/pkg/detector/preemption.go
+++ b/pkg/detector/preemption.go
@@ -9,7 +9,6 @@ import (
 	policyv1alpha1 "github.com/karmada-io/karmada/pkg/apis/policy/v1alpha1"
 	"github.com/karmada-io/karmada/pkg/features"
 	"github.com/karmada-io/karmada/pkg/util"
-	"github.com/karmada-io/karmada/pkg/util/fedinformer/keys"
 	"github.com/karmada-io/karmada/pkg/util/helper"
 	"github.com/karmada-io/karmada/pkg/util/names"
 )
@@ -40,8 +39,13 @@ func (d *ResourceDetector) handlePropagationPolicyPreemption(policy *policyv1alp
 			continue
 		}
 
-		errs = append(errs, d.preemptPropagationPolicy(resourceTemplate, policy))
-		// TODO(whitewindmills): PP preempts CPP.
+		if err := d.preemptPropagationPolicy(resourceTemplate, policy); err != nil {
+			errs = append(errs, err)
+			continue
+		}
+		if err := d.preemptClusterPropagationPolicyDirectly(resourceTemplate, policy); err != nil {
+			errs = append(errs, err)
+		}
 	}
 
 	return utilerrors.NewAggregate(errs)
@@ -61,7 +65,9 @@ func (d *ResourceDetector) handleClusterPropagationPolicyPreemption(policy *poli
 			continue
 		}
 
-		errs = append(errs, d.preemptClusterPropagationPolicy(resourceTemplate, policy))
+		if err := d.preemptClusterPropagationPolicy(resourceTemplate, policy); err != nil {
+			errs = append(errs, err)
+		}
 	}
 
 	return utilerrors.NewAggregate(errs)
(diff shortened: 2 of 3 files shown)


whitewindmills:
### test result
1. create configmap `test-cm`.
```
apiVersion: v1
data:
  a: b
kind: ConfigMap
metadata:
  creationTimestamp: "2023-07-27T03:36:57Z"
  labels:
    propagationpolicy.karmada.io/name: test-pp
  name: test-cm
  namespace: default
  resourceVersion: "4121865"
  uid: 962cf2d0-55f1-4ab7-9286-50499fefe3b1
```
2. create cpp `test-cpp`.
```
kind: ClusterPropagationPolicy
metadata:
  creationTimestamp: "2023-07-27T03:39:34Z"
  generation: 2
  name: test-cpp
  resourceVersion: "4116922"
  uid: fc1f2c80-8ba0-4f65-bb91-7240d26a1c5a
spec:
  conflictResolution: Abort
  placement:
    clusterTolerations:
    - effect: NoExecute
      key: cluster.karmada.io/not-ready
      operator: Exists
      tolerationSeconds: 300
    - effect: NoExecute
      key: cluster.karmada.io/unreachable
      operator: Exists
      tolerationSeconds: 300
    replicaScheduling:
      replicaSchedulingType: Duplicated
  preemption: Never
  priority: 10
  resourceSelectors:
  - apiVersion: v1
    kind: ConfigMap
    name: test-cm
    namespace: default
```
3. configmap `test-cm` was bound to cpp `test-cpp`.
0880c4c80710bee1fd12
4. create pp `test-pp` which priority is lower than cpp `test-cpp`.
```
kind: PropagationPolicy
metadata:
  creationTimestamp: "2023-07-27T03:46:48Z"
  generation: 1
  name: test-pp
  namespace: default
  resourceVersion: "4121864"
  uid: 4b8cbbe2-ba23-40f1-8a17-d23b0f29711a
spec:
  conflictResolution: Abort
  placement:
    clusterTolerations:
    - effect: NoExecute
      key: cluster.karmada.io/not-ready
      operator: Exists
      tolerationSeconds: 300
    - effect: NoExecute
      key: cluster.karmada.io/unreachable
      operator: Exists
      tolerationSeconds: 300
    replicaScheduling:
      replicaDivisionPreference: Aggregated
  preemption: Always
  priority: 0
  resourceSelectors:
  - apiVersion: v1
    kind: ConfigMap
    name: test-cm
    namespace: default
```
5. pp `test-pp` preempts cpp `test-cpp` regardless of priority.
image


### related logs
0880c4c80710bea1af17

pkg/detector/preemption.go
```diff
@@ -113,6 +120,28 @@ func (d *ResourceDetector) preemptPropagationPolicy(resourceTemplate *unstructur
...
+		// should not happen.
+		return err
+	}
+	if err := d.ApplyPolicy(resourceTemplate, clusterWideKey, policy, policyv1alpha1.ClusterPropagationPolicyLabel); err != nil {
+		klog.Errorf("Failed to apply new propagation policy(%s/%s) on resource template(%s, kind=%s, %s) directly: %v.", policy.Namespace, policy.Name,
+			resourceTemplate.GetAPIVersion(), resourceTemplate.GetKind(), names.NamespacedKey(resourceTemplate.GetNamespace(), resourceTemplate.GetName()), err)
+		return err
+	}
+	klog.V(4).Infof("Propagation policy(%s/%s) has preempted another cluster propagation policy(%s).",
+		policy.Namespace, policy.Name, claimedPolicyName)
```
chaunceyjiang, line 141: I believe that it is necessary to add some events, but just printing logs may not be user-friendly for end users.

whitewindmills: see task details: https://github.com/karmada-io/karmada/issues/3787.
complete in the next pr.

RainbowMango (commented):
/assign

pkg/detector/detector.go
```diff
@@ -373,7 +373,7 @@ func (d *ResourceDetector) LookForMatchedClusterPolicy(object *unstructured.Unst
 }
 
 // ApplyPolicy starts propagate the object referenced by object key according to PropagationPolicy.
-func (d *ResourceDetector) ApplyPolicy(object *unstructured.Unstructured, objectKey keys.ClusterWideKey, policy *policyv1alpha1.PropagationPolicy) (err error) {
+func (d *ResourceDetector) ApplyPolicy(object *unstructured.Unstructured, objectKey keys.ClusterWideKey, policy *policyv1alpha1.PropagationPolicy, removedLabelKeys ...string) (err error) {
```
RainbowMango, line 376: Just feel weird about adding `removedLabelKeys` to `ApplyPolicy`. `ApplyPolicy` should not take the responsibility of `preemption`.

A generally idea is that can we handle preemption by replace the `claimed info`?
For example, when a PP preempts a CPP:
- 1. Remove the claimed label(`clusterpropagationpolicy.karmada.io/name`) from the resource template
- 2. Claim the resource template by PP's label (`clusterpropagationpolicy.karmada.io/name`)

whitewindmills: > For example, when a PP preempts a CPP:

> *  i.Remove the claimed label(clusterpropagationpolicy.karmada.io/name) from the resource template
> *  ii. Claim the resource template by PP's label (clusterpropagationpolicy.karmada.io/name)

1. the claimed label `clusterpropagationpolicy.karmada.io/name` on rb cannot be removed if we follow the above steps.
2. if we remove the claimed label `clusterpropagationpolicy.karmada.io/name` from the resource template and rb, and then claim new pp label on them. there will be two update operations, seems uncessary.

whitewindmills: BTW, the responsibility of `ApplyPolicy` is to apply the new policy, and I think it also includes the semantics of removing the old policy.

RainbowMango: > the claimed label clusterpropagationpolicy.karmada.io/name on rb cannot be removed if we follow the above steps.

That's a thing we have to handle.

>  there will be two update operations, seems necessary.

We can do that by updating once.

whitewindmills: so you just don't think it's appropriate to remove the old policy label and apply the new policy through `ApplyPolicy`, right?
We should create a new function to solve this problem, giving it a new meaning.

RainbowMango: Yes,  It'd be better not to couple the original logic. 

We can make a simple preemption function that focuses on preempting by changing the claim info on the resource template.
After that, the resource template would get a chance to reconcile, that is the previous apply process.

whitewindmills: > After that, the resource template would get a chance to reconcile, that is the previous apply process.

I'm afraid it won't work as expected if just update the resource template and have it reconcile again.
https://github.com/karmada-io/karmada/blob/cf90ffc64fbfa16083c8371905b4789fb4f98a6f/pkg/detector/detector.go#L394-L414

finally, binding will be like:
```
apiVersion: work.karmada.io/v1alpha2
kind: ResourceBinding
metadata:
  creationTimestamp: "2023-07-28T06:36:01Z"
  finalizers:
  - karmada.io/binding-controller
  generation: 2
  labels:
    clusterpropagationpolicy.karmada.io/name: xxxx
    propagationpolicy.karmada.io/name: yyyy
```
so, the possible way is to provide a function similar to `ApplyPolicy` but only for preemption. how do you think of this?

RainbowMango: > the possible way is to provide a function similar to ApplyPolicy but only for preemption. 

We'd better not. 

When building(updating) a RB, we know which claim label should exist at :
https://github.com/karmada-io/karmada/blob/cf90ffc64fbfa16083c8371905b4789fb4f98a6f/pkg/detector/detector.go#L394-L414

whitewindmills: function `util.DedupeAndMergeLabels` only merge new labels into old binding, not remove label `clusterpropagationpolicy.karmada.io/name`, which is the root of the problem. so is it necessary for us to change the behavior here?
```
```

RainbowMango: Here is an ugly solution:
```diff
diff --git a/pkg/detector/detector.go b/pkg/detector/detector.go                                     
index be715e631..86695f277 100644                                                                    
--- a/pkg/detector/detector.go                                                                       
+++ b/pkg/detector/detector.go                                                                       
@@ -422,6 +422,11 @@ func (d *ResourceDetector) ApplyPolicy(object *unstructured.Unstructured, object
                        bindingCopy.Spec.Placement = binding.Spec.Placement                          
                        bindingCopy.Spec.Failover = binding.Spec.Failover                            
                        bindingCopy.Spec.ConflictResolution = binding.Spec.ConflictResolution        
+                                                                                                    
+                       // cleanup in case of current ResourceBinding is preempted from
+                       // a ClusterPropagationPolicy
+                       delete(binding.Labels, policyv1alpha1.ClusterPropagationPolicyLabel)
+
                        return nil
                })
                if err != nil {

```

pkg/detector/preemption.go
```diff
@@ -113,6 +120,28 @@ func (d *ResourceDetector) preemptPropagationPolicy(resourceTemplate *unstructur
...
+	clusterWideKey, err := keys.ClusterWideKeyFunc(resourceTemplate)
+	if err != nil {
+		// should not happen.
+		return err
+	}
+	if err := d.ApplyPolicy(resourceTemplate, clusterWideKey, policy); err != nil {
+		klog.Errorf("Failed to apply new propagation policy(%s/%s) on resource template(%s, kind=%s, %s) directly: %v.", policy.Namespace, policy.Name,
+			resourceTemplate.GetAPIVersion(), resourceTemplate.GetKind(), names.NamespacedKey(resourceTemplate.GetNamespace(), resourceTemplate.GetName()), err)
+		return err
+	}
```
RainbowMango, line 139: Can we just claim the resource template here? like:
```golang
	err = d.ClaimPolicyForObject(resourceTemplate, policy.Namespace, policy.Name)
	if err != nil {
		klog.Errorf("xxxx")
		return err
	}
```
This might be cheaper than `ApplyPolicy`, and just need to sync/update RB once. What do you think?

RainbowMango:
Can you help to [run the test](https://github.com/karmada-io/karmada/pull/3845#issuecomment-1652906438) again? 
I just tested it on my side, seems it does not work as expected, I'll run again.

whitewindmills:
> I just tested it on my side, seems it does not work as expected, I'll run again.

could you share your karmada-controller-manager logs? I only need logs of the `detector`, like this screenshot.
image

RainbowMango:
Sorry, I forgot to set the feature gate... I usually test PR by `hack/local-up-karmada.sh`. 
```
W0731 11:48:38.771315       1 preemption.go:22] Cannot handle the preemption process because feature gate "PropagationPolicyPreemption" is not enabled.

```
It works fine~ Thanks.

RainbowMango (approved):
/lgtm
/approve

whitewindmills:
thanks~

PR #3853, opened 2023-07-28
feat: avoid calculating available replicas for non-workload

### LINKED ISSUE #3743: Avoid calculating available replicas for non-workload
We should avoid calculating available replicas for non-workload, like `ServiceAccount`, `ConfigMap` and `Secret` and etc.

**Why is this needed**:
For now, all resource are be selected to calculate its available replicas in every member cluster, that's unnecessary expensive cost.
https://github.com/karmada-io/karmada/blob/1d030d4f8c01f8370c80bb0ecdd52630d87a5bf0/pkg/scheduler/core/util.go#L37-L76

whitewindmills:
**What type of PR is this?**
/kind feature

avoid calculating available replicas for non-workload.

Fixes #3743

```release-note
NONE
```


Garrybest (approved):
/lgtm
/approve

PR #1112, opened 2021-12-15
set default for clusterCacheSync timeout and make it configurable

snowplayfire:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:
```
I1215 12:51:03.337598       1 cluster_status_controller.go:87] Syncing cluster status: ntgxh-d
E1215 12:51:33.343648       1 cluster_status_controller.go:252] Failed to sync cache for cluster: ntgxh-d, error: informer for /v1, Resource=pods hasn't synced
E1215 12:51:33.343680       1 cluster_status_controller.go:149] Failed to get or create informer for Cluster ntgxh-d. Error: informer for /v1, Resource=pods hasn't synced.
```

```
synced := c.InformerManager.WaitForCacheSyncWithTimeout(cluster.Name, util.CacheSyncTimeout)
if synced == nil {
	return fmt.Errorf("no informerFactory for cluster %s exist", cluster.Name)
}
for _, gvr := range gvrs {
	if !synced[gvr] {
		return fmt.Errorf("informer for %s hasn't synced", gvr)
	}
}
```
here `util.CacheSyncTimeout = 30 * time.Second`

The  `CacheSyncTimeout`  should be configurable  according to cluster size.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #3874, opened 2023-08-01
use ClusterCacheSyncTimeout for resources on fed control plane as well

lxtywypc:
**What type of PR is this?**  
/kind feature  

Now, the default timeout for waiting cache syncing in `controller-runtime` is 2 minutes. If we have a lot resources on the fed control plane, the default timeout might be reached and the process would return with an error like:  
```
E0801 06:15:22.538099       1 controllermanager.go:154] controller manager exits unexpectedly: [failed to wait for build-resource-informers for work caches to sync: timed out waiting for cache to be synced, failed waiting for all runnables to end within grace period of 30s: context deadline exceeded
```  
And the `karmada-controller-manager` would be hard to restart successfully.  

Fixes #

The log printed above is from `karmada-controller-manager v1.3.1`, but it still exists in the latest version.

```release-note
`karmada-controller-manager`: The `--cluster-cache-sync-timeout` flag is now used to specify the sync timeout of the control plane cache in addition to the member cluster's cache. The default value has been increased to 2 minutes.
```

### DIFF
--- a/pkg/controllers/federatedhpa/federatedhpa_controller.go
+++ b/pkg/controllers/federatedhpa/federatedhpa_controller.go
@@ -68,6 +68,7 @@ type FederatedHPAController struct {
 	RESTMapper                meta.RESTMapper
 	EventRecorder             record.EventRecorder
 	TypedInformerManager      typedmanager.MultiClusterInformerManager
+	ClusterCacheSyncTimeout   metav1.Duration
 
 	monitor monitor.Monitor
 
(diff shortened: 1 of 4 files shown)


pkg/controllers/federatedhpa/federatedhpa_controller.go
```diff
@@ -68,6 +68,7 @@ type FederatedHPAController struct {
 	RESTMapper                meta.RESTMapper
 	EventRecorder             record.EventRecorder
 	TypedInformerManager      typedmanager.MultiClusterInformerManager
+	ClusterCacheSyncTimeout   metav1.Duration
```
RainbowMango, line 71: Nice, 100% agree that we need to use the same configuration for informers inside of a controller.

RainbowMango (approved):
/lgtm
/approve

Thanks @lxtywypc .

I updated the release notes a little bit by the way.

I also revisited the https://github.com/karmada-io/karmada/pull/1112, thanks @snowplayfire  for introducing the flag. It means a lot to Karmada's scalability.

liangyuanpeng:
It's useful , let me cherry pick it to 1.6

RainbowMango:
We don't usually cherry-pick features, but considering the severity of this issue(and the fact that you have done it #3880  :)), I'm ok with it.

PR #3875, opened 2023-08-01
record binding-reference-key and work-name info in binding annotations

lxtywypc:
**What type of PR is this?**  
/kind feature  

Now, we could list `work` by `binding reference key`, but the key could only be generated in code. It would make the daily operations easier if we could record the key and the work name in the annotations of `binding`.

Fixes #

```release-note

```


jwcesign:
Hi @lxtywypc, in the future, we will replace `binding reference key` with the uid of the binding. This change should solve your problem.

lxtywypc:
> Hi @lxtywypc, in the future, we will replace `binding reference key` with the uid of the binding. This change should solve your problem.

Sounds great. Thanks~

PR #3879, opened 2023-08-02
fix bug: avoid updating directly cached resource template

### LINKED ISSUE #3878: Resource template cannot be reconciled again due to cache modification
**What happened**:
I deleted propagation policy `default/test-pp` to unbind from configmap `default/test-cm`, but find that configmap `default/test-cm` cannot be reconciled again. 
image

that means configmap `default/test-cm` cannot be matched by new propagation policy `default/test-pp-1`.
image

configmap `default/test-cm` can be matched by new propagation policy `default/test-pp-1`.

1. create such configmap(keep labels non-nil):
```
apiVersion: v1
data:
  a: b
kind: ConfigMap
metadata:
  labels:
    a: b
  name: test-cm
  namespace: default
```
2. create propagation policy.
3. delete propagation policy after configmap is propagated.
4. check if configmap can be reconcile again.

`detector` deletes directly policy labels from cached resource template.
https://github.com/karmada-io/karmada/blob/d916d877a4942d9abf0467d65562d471acf8138d/pkg/detector/detector.go#L1117-L1131

**Environment**:
- Karmada version:
- Others:

whitewindmills:
**What type of PR is this?**
/kind bug

avoid updating directly cached resource template.

Fixes #3878

```release-note
karmada-controller-manager: Avoid updating directly cached resource template.
```

### DIFF
--- a/pkg/detector/policy.go
+++ b/pkg/detector/policy.go
@@ -272,9 +272,9 @@ func (d *ResourceDetector) removeResourceLabelsIfNotMatch(objectReference workv1
 		return false, nil
 	}
 
-	for _, labelKey := range labelKeys {
-		util.RemoveLabel(object, labelKey)
-	}
+	object = object.DeepCopy()
+	util.RemoveLabels(object, labelKeys...)
+
 	err = d.Client.Update(context.TODO(), object)
 	if err != nil {
 		return false, err
--- a/pkg/detector/detector.go
+++ b/pkg/detector/detector.go
@@ -565,6 +565,9 @@ func (d *ResourceDetector) ApplyClusterPolicy(object *unstructured.Unstructured,
 }
 
 // GetUnstructuredObject retrieves object by key and returned its unstructured.
+// Any updates to this resource template are not recommended as it may come from the informer cache.
+// We should abide by the principle of making a deep copy first and then modifying it.
+// See issue: https://github.com/karmada-io/karmada/issues/3878.
 func (d *ResourceDetector) GetUnstructuredObject(objectKey keys.ClusterWideKey) (*unstructured.Unstructured, error) {
 	objectGVR, err := restmapper.GetGroupVersionResource(d.RESTMapper, objectKey.GroupVersionKind())
 	if err != nil {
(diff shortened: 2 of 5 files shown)


XiShanYongYe-Chang (commented):
Thanks a lot.

In order to prevent other places where the utility methods are called from modifying the objects in the cache, I think it would be better to directly make modifications within the utility methods.

Furthermore, this issue indicates that we do not have an end-to-end (E2E) coverage for this scenario. If possible, could you please provide an additional E2E use case or create an issue to track this task?

whitewindmills:
> I think it would be better to directly make modifications within the utility methods.

The number of resources may be huge, and the calling frequency of this utility function is very high. It is obviously not necessary to perform a deep copy every time we read a resource through this utility function. We only make a deep copy in a few places that need to be modified.

> Furthermore, this issue indicates that we do not have an end-to-end (E2E) coverage for this scenario. If possible, could you please provide an additional E2E use case or create an issue to track this task?

good suggestion.
tracked issue: https://github.com/karmada-io/karmada/issues/3884

XiShanYongYe-Chang:
> The number of resources may be huge, and the calling frequency of this utility function is very high. It is obviously not necessary to perform a deep copy every time we read a resource through this utility function. We only make a deep copy in a few places that need to be modified.

In this case, should we consider adding some comments in the utility func to indicate the need for attention when modifying cache issues, so as to prevent overlooking this factor when using it in new places?

whitewindmills:
> In this case, should we consider adding some comments in the utility func to indicate the need for attention when modifying cache issues, so as to prevent overlooking this factor when using it in new places?

:+1:

whitewindmills:
ptal @XiShanYongYe-Chang @RainbowMango @chaunceyjiang

XiShanYongYe-Chang (commented):
Thanks
/lgtm

pkg/detector/policy.go
```diff
@@ -272,9 +272,13 @@ func (d *ResourceDetector) removeResourceLabelsIfNotMatch(objectReference workv1
 		return false, nil
 	}
 
+	object = object.DeepCopy()
+	objLabels := object.GetLabels()
 	for _, labelKey := range labelKeys {
-		util.RemoveLabel(object, labelKey)
```
chaunceyjiang, line 276: Just a question, why not use the `util.RemoveLabel` function?

whitewindmills: improve execution efficiency. every call of `util.RemoveLabel` will get and set labels.

chaunceyjiang: Can we extend this function? Like this: `util.RemoveLabels(object, labelKeys...)`

whitewindmills: sure

chaunceyjiang (commented):
/lgtm

RainbowMango (commented):
/assign

whitewindmills:
done

RainbowMango (approved):
/approve

whitewindmills:
@XiShanYongYe-Chang @RainbowMango  should we cherry-pick it to the previous branch?

RainbowMango:
+1

PR #3891, opened 2023-08-03
add resources to webhook chart

yizhang-zen:
**What type of PR is this?**

- add resources to webhook chart
- add descheduler as part of container's name

Does webhook need livenessProbe?

Fixes #

```release-note

```

### DIFF
--- a/charts/karmada/templates/karmada-descheduler.yaml
+++ b/charts/karmada/templates/karmada-descheduler.yaml
@@ -42,7 +42,7 @@ spec:
       {{- end }}
       automountServiceAccountToken: false
       containers:
-        - name: {{ $name }}
+        - name: {{ $name }}-descheduler
           image: {{ template "karmada.descheduler.image" . }}
           imagePullPolicy: {{ .Values.descheduler.image.pullPolicy }}
           command:
--- a/charts/karmada/templates/karmada-scheduler-estimator.yaml
+++ b/charts/karmada/templates/karmada-scheduler-estimator.yaml
@@ -58,6 +58,8 @@ spec:
             - name: member-kubeconfig
               subPath: {{ $clusterName }}-kubeconfig
               mountPath: /etc/{{ $clusterName }}-kubeconfig
+          resources:
+          {{- toYaml .Values.schedulerEstimator.resources | nindent 12 }}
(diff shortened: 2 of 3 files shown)


calvin0327:
@yizhang-zen Do you test this change?

yizhang-zen:
> @yizhang-zen Do you test this change?

No, I didn't. What are the steps to test out?

chaosi-zju:
Hello, thank you for your pull request~

1. I wanna to known why you want to add descheduler as part of container's name? Are there any scenes that must be like this?

2. If you want to test this change, I think you can refer to [Installing the chart](https://github.com/karmada-io/karmada/tree/master/charts/karmada#installing-the-chart) and test it by `local installation`, then you shall observe if the installation was successful and whether the `rename` and `add resource` op done as you expected.

3. I really appreciate it that you add the `resources` to the chart, further more, it seems like the same bug exist in `karmada-scheduler-estimator`, could you please help us to fix it too?

yizhang-zen:
Thank you, @chaosi-zju 

> I wanna to known why you want to add descheduler as part of container's name? Are there any scenes that must be like this?

No, just keep the same name output as other components' chart ([aggregated apiserver](https://github.com/karmada-io/karmada/blob/master/charts/karmada/templates/karmada-aggregated-apiserver.yaml#L7) and [karmada apiserver ](https://github.com/karmada-io/karmada/blob/master/charts/karmada/templates/karmada-apiserver.yaml#L6)for example). 

> it seems like the same bug exist in karmada-scheduler-estimator, could you please help us to fix it too?

Added:)

RainbowMango:
@chaosi-zju can you help to take a look at this PR. This reminds me we should enable chart installation in CI.

@yizhang-zen I'm sorry for letting this PR wait so long. Can you help to rebase your PR? Because the test suites now might be different from the master branch.

yizhang-zen:
>@yizhang-zen I'm sorry for letting this PR wait so long. Can you help to rebase your PR? Because the test suites now might be different from the master branch.

You are good. I updated my pr just recently and this is not urgent at all. 
I have rebased my PR

RainbowMango (approved):
/approve

Issue #3835, opened 2023-07-25
Cleanup deprecated methods in wait package


RainbowMango:
**What would you like to be added**:

- [x] Update `wait.Poll` to `wait.PollUntilContextTimeout` (with `immediately = false`)  (@Affan-7, #3906 )
For example:
```diff
--- a/pkg/controllers/certificate/cert_rotation_controller.go
+++ b/pkg/controllers/certificate/cert_rotation_controller.go
@@ -165,7 +165,7 @@ func (c *CertRotationController) syncCertRotation(secret *corev1.Secret) error {

        var newCertData []byte
        klog.V(1).Infof("Waiting for the client certificate to be issued")
-       err = wait.Poll(1*time.Second, 5*time.Minute, func() (done bool, err error) {
+       err = wait.PollUntilContextTimeout(context.TODO(), 1*time.Second, 5*time.Minute, false, func(ctx context.Context) (done bool, err error) {
                csr, err := c.KubeClient.CertificatesV1().CertificateSigningRequests().Get(context.TODO(), csr, metav1.GetOptions{})
                if err != nil {
                        return false, fmt.Errorf("failed to get the cluster csr %s. err: %v", clusterName, err)
```

- [x] Update `wait.PollImmediate` to `wait.PollUntilContextTimeout` (with `immediately = true`)  (@liangyuanpeng, #3921)
For example:
```diff
diff --git a/pkg/controllers/cluster/cluster_controller.go b/pkg/controllers/cluster/cluster_controller.go
index ae123a86e..678bf7188 100644
--- a/pkg/controllers/cluster/cluster_controller.go
+++ b/pkg/controllers/cluster/cluster_controller.go
@@ -394,7 +394,7 @@ func (c *Controller) monitorClusterHealth(ctx context.Context) (err error) {
        for i := range clusters {
                cluster := &clusters[i]
                var observedReadyCondition, currentReadyCondition *metav1.Condition
-               if err = wait.PollImmediate(MonitorRetrySleepTime, MonitorRetrySleepTime*HealthUpdateRetry, func() (bool, error) {
+               if err = wait.PollUntilContextTimeout(ctx, MonitorRetrySleepTime, MonitorRetrySleepTime*HealthUpdateRetry, true, func(ctx context.Context) (bool, error) {
                        // Cluster object may be changed in this function.
                        observedReadyCondition, currentReadyCondition, err = c.tryUpdateClusterHealth(ctx, cluster)
                        if err == nil {

```

[...]

PR #3906, opened 2023-08-07
Replace deprecated wait.poll

Affan-7:
**What type of PR is this?**
/kind cleanup

This PR replaces the deprecated `wait.Poll()` function with the `wait.PollUntilContextTimeout()` function.

Related to #3835
1 out of 4 tasks

```release-note
NONE
```


carlory:
/lgtm

/assign @rainbowmango 

for workflow and approval.

RainbowMango (approved):
/approve
Thanks~

Affan-7:
The tests were failing, here are the logs: https://github.com/karmada-io/karmada/actions/runs/5783869909/job/15676511801

Should we increase the default value of `j.Wait`? Or there is some other problem?

RainbowMango:
It might be flak, let's give it another try.

RainbowMango:
> Should we increase the default value of j.Wait? Or there is some other problem?

I'm not sure yet. let's give it another try and try to find the root cause.

I'm curious that the test failed in `42` seconds, but the default waiting time(`j.Wait`) is `60` seconds.

Just echo the log here: 
https://github.com/karmada-io/karmada/actions/runs/5787607401/job/15697196471?pr=3906#step:5:2723
```
[FAILED] [42.440 seconds]
[cluster joined] reschedule testing Deployment propagation testing [AfterEach] testing clusterAffinity of the policy when the ReplicaScheduling of the policy is nil, reschedule testing [needCreateCluster]
  [AfterEach] /home/runner/work/karmada/karmada/test/e2e/rescheduling_test.go:176
  [It] /home/runner/work/karmada/karmada/test/e2e/rescheduling_test.go:271

  Captured StdOut/StdErr Output >>
  I0808 02:23:49.855080   46074 deployment.go:199] The ResourceBinding(karmadatest-2dj87/deploy-tjnp2-deployment) schedule result is: member1,member2
  cluster(member-e2e-lhr) is joined successfully
  I0808 02:23:59.098817   46074 deployment.go:199] The ResourceBinding(karmadatest-2dj87/deploy-tjnp2-deployment) schedule result is: member-e2e-lhr,member1,member2
  I0808 02:23:59.109167   46074 deployment.go:199] The ResourceBinding(karmadatest-2dj87/deploy-tjnp2-deployment) schedule result is: member-e2e-lhr,member1,member2
  E0808 02:23:59.135302   46074 unjoin.go:239] Failed to delete cluster object. cluster name: member-e2e-lhr, error: context deadline exceeded
  E0808 02:23:59.136891   46074 unjoin.go:174] Failed to delete cluster object. cluster name: member-e2e-lhr, error: context deadline exceeded
```

pkg/controllers/certificate/cert_rotation_controller.go
```diff
@@ -165,7 +165,7 @@ func (c *CertRotationController) syncCertRotation(secret *corev1.Secret) error {
 
 	var newCertData []byte
 	klog.V(1).Infof("Waiting for the client certificate to be issued")
-	err = wait.Poll(1*time.Second, 5*time.Minute, func() (done bool, err error) {
+	err = wait.PollUntilContextTimeout(context.TODO(), 1*time.Second, 5*time.Minute, false, func(ctx context.Context) (done bool, err error) {
```
Jorropo, line 168: Theses are not equivalent, I think you meant to use `PollWithContextTimeout` as the docs said to do, however this does not exists.

Instead you can do:
```suggestion
  ctx, cancel := context.WithTimeout(context.TODO(), 5*time.Minute)
  defer cancel()
	err = wait.PollWithContext(ctx, 1*time.Second, 5*time.Minute, func(ctx context.Context) (done bool, err error) {
```

pkg/controllers/certificate/cert_rotation_controller.go
```diff
@@ -165,7 +165,7 @@ func (c *CertRotationController) syncCertRotation(secret *corev1.Secret) error {
...
 		csr, err := c.KubeClient.CertificatesV1().CertificateSigningRequests().Get(context.TODO(), csr, metav1.GetOptions{})
```
Jorropo, line 169:
```suggestion
		csr, err := c.KubeClient.CertificatesV1().CertificateSigningRequests().Get(ctx, csr, metav1.GetOptions{})
```
Even if it's not passed up, this can be passed down.

Jorropo (changes requested):
If you change all of the `PollUntilContextTimeout` like I showed I belive it would work. :slightly_smiling_face:

pkg/controllers/certificate/cert_rotation_controller.go
Affan-7: Thanks for your collaboration @Jorropo, but unfortunately the `PollWithContext` method is also deprecated.

Reference: https://pkg.go.dev/k8s.io/apimachinery@v0.27.3/pkg/util/wait#PollWithContext

Jorropo: Ah right mb.

RainbowMango:
/remove-approve
We need to figure out the root cause, probably affected by https://github.com/kubernetes/kubernetes/issues/119533

Affan-7:
## Here is what I did:

**TL:DR;**

Value of `j.Wait` was zero in the both cases `wait.Poll` and `wait.PollUntilContextTimeout()`. Since the `wait.Poll()` is deprecated it no longer returns any errors. That's why `wait.Poll()` was passing the test and `wait.PollUntilContextTimeout()` was failing. I specified the value of `j.Wait` explicitly under the test file and everything works fine as expected. 

---

I added `fmt.Printf("+++++ value of j.Wait: %v\n", j.Wait)` before `wait.PollUntilContextTimeout(context.TODO(), 1*time.Second, j.Wait, .....)`. And I ran the test on the my fork of Karmada, using github actions. I was surprised that the value of `j.Wait` was `zero`.

Reference: https://github.com/Affan-7/karmada/actions/runs/5841035785/job/15840850551#step:5:2798

So I run the test again with the deprecated `wait.Poll()` method. But the test passed! Then I ran it again with `fmt.Printf("+++++ value of j.Wait: %v\n", j.Wait)` and the value of `j.Wait` was also `zero`.

Reference: https://github.com/Affan-7/karmada/actions/runs/5841648595/job/15842010633#step:5:2999

The value of `j.Wait` was still `zero` but the test case weren't failing. Because the `wait.Poll()` method is deprecated and doesn't return errors anymore. I found this in the documentation of `wait.Poll`.

``` 
Deprecated: This method does not return errors from context, use PollWithContextTimeout. Note that the new method will no longer return ErrWaitTimeout and instead return errors defined by the context package. Will be removed in a future release.
```

Reference: https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#Poll

I was assuming that the default value of `j.Wait` should be `60` because of `flags.DurationVar(&j.Wait, "wait", 60*time.Second, ....)` under the `unjoin.go`. But this block of code doesn't get executed during the test (because it's for CLI). Therefore the value of 'j.Wait' remains `zero`.

So I added the value of `j.Wait` explicitly under `rescheduling_test.go` by using `Wait: 60 * time.Second,`. It's very similar to the `Wait: 5 * options.DefaultKarmadactlCommandDuration,` on the [line 441 of the `karmadactl_test.go`](https://github.com/karmada-io/karmada/blob/6dea12190e9752eef937c2461e0223a5fe88ee6a/test/e2e/karmadactl_test.go#L441C15-L441C15).

After specifying the value of `j.Wait` under the `rescheduling_test.go` all the tests were passing successfully.
Reference: https://github.com/Affan-7/karmada/actions/runs/5841642875/job/15841998395#step:5:3461

Affan-7:
Can you please merge this @RainbowMango :blush:. Or is there  something that should I improve?

RainbowMango:
Thanks @Affan-7 for the excellent analysis, I think it is acceptable to set a timeout [here](https://github.com/karmada-io/karmada/pull/3906/files#diff-b6331f265a97596e2848f6e505111c716f47659dd9057127740cbf1ee1d0d484R124-R184).

I wonder if this is affected by https://github.com/kubernetes/kubernetes/issues/119533? 

I'm hesitant to do it before https://github.com/kubernetes/kubernetes/issues/119533.

pkg/karmadactl/register/register.go
```diff
@@ -509,7 +509,7 @@ func (o *CommandRegisterOption) constructKarmadaAgentConfig(bootstrapClient *kub
 	}
 
 	klog.V(1).Infof("Waiting for the client certificate to be issued")
-	err = wait.Poll(1*time.Second, o.Timeout, func() (done bool, err error) {
+	err = wait.PollUntilContextTimeout(context.TODO(), 1*time.Second, o.Timeout, false, func(ctx context.Context) (done bool, err error) {
```
zhzhuang-zju, line 512: Why `immediate` here is set to `false`? `condition` can be invoked immediately.

pkg/karmadactl/unjoin/unjoin.go
```diff
@@ -223,7 +223,7 @@ func (j *CommandUnjoinOption) deleteClusterObject(controlPlaneKarmadaClient *kar
 	}
 
 	// make sure the given cluster object has been deleted
-	err = wait.Poll(1*time.Second, j.Wait, func() (done bool, err error) {
+	err = wait.PollUntilContextTimeout(context.TODO(), 1*time.Second, j.Wait, false, func(ctx context.Context) (done bool, err error) {
```
zhzhuang-zju, line 226: ditto~

pkg/karmadactl/register/register.go
RainbowMango: That is because this behavior is equipment with previous.
`wait.Poll` always waits the interval before the run of `condition`.

pkg/karmadactl/unjoin/unjoin.go
RainbowMango: ditto~
answered at the last comment.

RainbowMango:
Just rebased this and fixed the golint issues.
/assign @XiShanYongYe-Chang

pkg/karmadactl/register/register.go
RainbowMango: That's the idea not run the condition so rush.

XiShanYongYe-Chang (commented):
Thanks~
/lgtm

PR #3908, opened 2023-08-07
fix unable to sync list

WulixuanS:
**What type of PR is this?**
/kind bug

When client disconnected from apiserver, it will trigger `replace` in delta_fifo.go.
https://github.com/kubernetes/kubernetes/blob/2c6c4566eff972d6c1320b5f8ad795f88c822d09/staging/src/k8s.io/client-go/tools/cache/delta_fifo.go#L690

If  obj was deleted at this time, `replace` will generate a DeletedFinalStateUnknown delta, and `transformer` will return a error. Finally, the `list` will always fail and report `unable to sync list result: xxx`.
https://github.com/kubernetes/kubernetes/blob/2c6c4566eff972d6c1320b5f8ad795f88c822d09/staging/src/k8s.io/client-go/tools/cache/delta_fifo.go#L456
https://github.com/kubernetes/kubernetes/blob/2c6c4566eff972d6c1320b5f8ad795f88c822d09/staging/src/k8s.io/client-go/tools/cache/reflector.go#L566

Fixes #

```
Fixed unable to sync list issue in case of client disconnects from member cluster. The fixes apply to following components:
- karmada-controller-manager
- karmada-agent
- karmada-descheduler
- karmada-scheduler
- karmada-metrics-adapter
- karmada-search
```

### DIFF
--- a/pkg/util/fedinformer/transform.go
+++ b/pkg/util/fedinformer/transform.go
@@ -7,11 +7,16 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/cache"
 )
 
 // StripUnusedFields is the transform function for shared informers,
 // it removes unused fields from objects before they are stored in the cache to save memory.
 func StripUnusedFields(obj interface{}) (interface{}, error) {
+	if tombstone, ok := obj.(cache.DeletedFinalStateUnknown); ok {
+		obj = tombstone.Obj
+	}
+
 	accessor, err := meta.Accessor(obj)
 	if err != nil {
 		// shouldn't happen
@@ -29,8 +34,17 @@ func StripUnusedFields(obj interface{}) (interface{}, error) {
 // Note: this function removes most of the fields, please make sure your controller
 // doesn't care for the removed fields, especially when use in shared informers.
 func NodeTransformFunc(obj interface{}) (interface{}, error) {
-	node, ok := obj.(*corev1.Node)
-	if !ok {
+	var node *corev1.Node
+	switch t := obj.(type) {
+	case *corev1.Node:
+		node = t
+	case cache.DeletedFinalStateUnknown:
+		var ok bool
+		node, ok = t.Obj.(*corev1.Node)
+		if !ok {
+			return obj, fmt.Errorf("expect resource Node but got %v", reflect.TypeOf(t.Obj))
+		}
+	default:
 		return obj, fmt.Errorf("expect resource Node but got %v", reflect.TypeOf(obj))
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/util/fedinformer/transform.go
```diff
@@ -2,6 +2,7 @@ package fedinformer
 
 import (
 	"fmt"
+	"k8s.io/client-go/tools/cache"
```
jwcesign, line 5: import order

pkg/util/fedinformer/transform.go
```diff
@@ -12,6 +13,10 @@ import (
 // StripUnusedFields is the transform function for shared informers,
 // it removes unused fields from objects before they are stored in the cache to save memory.
 func StripUnusedFields(obj interface{}) (interface{}, error) {
+	if tombstone, ok := obj.(cache.DeletedFinalStateUnknown); ok {
+		obj = tombstone.Obj
+	}
+
```
jwcesign, line 19: What does this code mean? Should we proceed or ignore it?

pkg/util/fedinformer/transform.go
WulixuanS: fixed

pkg/util/fedinformer/transform.go
WulixuanS: If the object is a DeletedFinalStateUnknown, we need to extract the actual object and proceed it

RainbowMango (commented):
@jwcesign 
Do you think we need a release note for this patch? If yes, could you please propose one?

/assign

jwcesign:
How about this one?
```
`fedinformer`: Fix resource listing from Informer when client-go disconnects from API server.
```

/cc @WulixuanS @RainbowMango

RainbowMango:
I think we should point out which component was affected in the release note.

jwcesign:
I check which component use this informer,  and:
```
`fedinformer`: Fix resource listing from Informer when client-go disconnects from API server, the affected components include：karmada-agent/karmada-controller-manager/karmada-descheduler/karmada-scheduler/karmada-metrics-adapter/kamrada-resource-interpreter/karmada-search
```

pkg/util/fedinformer/transform.go
```diff
@@ -7,11 +7,16 @@ import (
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/cache"
 )
 
 // StripUnusedFields is the transform function for shared informers,
 // it removes unused fields from objects before they are stored in the cache to save memory.
 func StripUnusedFields(obj interface{}) (interface{}, error) {
+	if tombstone, ok := obj.(cache.DeletedFinalStateUnknown); ok {
+		obj = tombstone.Obj
+	}
```
RainbowMango, line 18: Could you please add a comment on this to explain the reason why we need to check the state? I'm afraid not many people know the details.

RainbowMango (commented):
Great thanks to @WulixuanS !
I only asking for more comments on this.

@ikaven1024 Would you like to take a look at this? I believe you are the expert in this area.

RainbowMango:
I updated the release notes. `kamrada-resource-interpreter` is not a component by the way.

pkg/util/fedinformer/transform.go
WulixuanS: I will add a comment as follow, is that ok? @RainbowMango 
// This could happen if watch deletion event was missed while disconnected from apiserver

RainbowMango: We have lessons learned on this, such as #3478 fixing a similar issue.
I'll go over the code base and check if we missed it somewhere and add the comments accordingly. (It's not fair asking you to do this in this PR.).

RainbowMango (approved):
/approve

PR #3910, opened 2023-08-07
[karmadactl get] judge with label on the resources determine whether resource is adopted by Karmada

XiShanYongYe-Chang:
**What type of PR is this?**

/kind cleanup

We have added labels to the resources in the member cluster to indicate whether the current resource is propagated by Karmada:

https://github.com/karmada-io/karmada/blob/1953a0ee50d9d452b9ca5fffdb2f60338eed0b63/pkg/util/constants.go#L26-L30

Therefore, we can directly process this label in the `karmadactl get` command, without needing to retrieve RB/CRB resources.

Fixes #

One reason for this change is that we cannot assume that users using karmadactl by default have the authority to get rb/crb resources.

```release-note
NONE
```

### DIFF
--- a/pkg/karmadactl/get/get.go
+++ b/pkg/karmadactl/get/get.go
@@ -17,7 +17,6 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	metav1beta1 "k8s.io/apimachinery/pkg/apis/meta/v1beta1"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/runtime/schema"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/watch"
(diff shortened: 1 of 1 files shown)


lonelyCZ:
Looks nice, thanks for your work, I will look it ASAP.

/assign

RainbowMango:
> One reason for this change is that we cannot assume that users using karmadactl by default have the authority to get rb/crb resources.

+1

pkg/karmadactl/get/get.go
```diff
@@ -578,20 +571,25 @@ func (g *CommandGetOptions) reconstructObj(obj runtime.Object, mapping *meta.RES
...
+		unObj := &unstructured.Unstructured{}
+		err := unObj.UnmarshalJSON(table.Rows[rowIdx].Object.Raw)
+		if err != nil {
+			klog.Errorf("Failed to unmarshal unObj, error is: %v", err)
+			continue
+		}
+		v, exist := unObj.GetLabels()[karmadautil.ManagedByKarmadaLabel]
+		if exist && v == karmadautil.ManagedByKarmadaLabelValue {
+			table.Rows[rowIdx].Cells = append(table.Rows[rowIdx].Cells, "Y")
 		}
```
zishen, line 591: Here need a `continue`?  or `Y` and `N` can appear at the same time?

XiShanYongYe-Chang: Thanks for reminding.

pkg/karmadactl/get/get.go
```diff
@@ -548,16 +533,24 @@ func (g *CommandGetOptions) reconstructionRow(objs []Obj, table *metav1.Table) (
...
-			if _, ok := RBInfo[rbKey]; ok {
-				tempRow.Cells = append(tempRow.Cells, "Y")
-			} else {
-				tempRow.Cells = append(tempRow.Cells, "N")
+			v, exist := unObj.GetLabels()[karmadautil.ManagedByKarmadaLabel]
+			if exist && v == karmadautil.ManagedByKarmadaLabelValue {
+				table.Rows[rowIdx].Cells = append(table.Rows[rowIdx].Cells, "Y")
 			}
-			table.Rows[rowIdx].Cells = tempRow.Cells
+			table.Rows[rowIdx].Cells = append(table.Rows[rowIdx].Cells, "N")
```
lonelyCZ, line 553: It also need `else` branch.

lonelyCZ (commented):
I just tested it in my env, it worked fine.

pkg/karmadactl/get/get.go
XiShanYongYe-Chang: Thanks~, updated.

RainbowMango (approved):
/approve

PR #3915, opened 2023-08-08
grant clusterrole admin with karamda resource permission

XiShanYongYe-Chang:
**What type of PR is this?**

/kind feature

`admin` is a built-in ClusterRole in Kubernetes. It collects rules from other clusterRoles through aggregation method. This logic is managed by the `clusterrole-aggregation-controller` in kube-controller-manager, so we need to enable this controller.

In addition, we also need to grant `admin` the permissions to handle Karmada resources, including read and write access.

Part of #3916 

```release-note
karmada-kube-controller-manager: grant admin clusterrole with karamda resource permission
```

### DIFF
--- a/artifacts/deploy/admin-clusterrole-aggregation.yaml
+++ b/artifacts/deploy/admin-clusterrole-aggregation.yaml
@@ -1,0 +1,149 @@
+# This configuration is used to grant the admin clusterrole read
+# and write permissions for Karmada resources.
+
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  annotations:
+    # refer to https://kubernetes.io/docs/reference/access-authn-authz/rbac/#auto-reconciliation
+    # and https://kubernetes.io/docs/reference/access-authn-authz/rbac/#kubectl-auth-reconcile
+    rbac.authorization.kubernetes.io/autoupdate: "true"
+  labels:
+    # refer to https://kubernetes.io/docs/reference/access-authn-authz/rbac/#default-roles-and-role-bindings
+    kubernetes.io/bootstrapping: rbac-defaults
+    # used to aggregate rules to view clusterrole
+    rbac.authorization.k8s.io/aggregate-to-view: "true"
+  name: karmada-view
+rules:
+  - apiGroups:
+      - "autoscaling.karmada.io"
+    resources:
+      - cronfederatedhpas
+      - cronfederatedhpas/status
+      - federatedhpas
+      - federatedhpas/status
+    verbs:
+      - get
+      - list
+      - watch
+  - apiGroups:
+      - "multicluster.x-k8s.io"
+    resources:
+      - serviceexports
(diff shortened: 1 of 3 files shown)


RainbowMango (commented):
/assign

artifacts/deploy/admin-clusterrole-aggregation.yaml
```diff
@@ -0,0 +1,141 @@
...
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  annotations:
+    rbac.authorization.kubernetes.io/autoupdate: "true"
+  labels:
+    kubernetes.io/bootstrapping: rbac-defaults
+    rbac.authorization.k8s.io/aggregate-to-edit: "true"
+  name: system:karmada-edit
```
RainbowMango, line 77: I wonder why name the role with `system:` prefix.
> Caution: The prefix system: is reserved for Kubernetes system use, so you should ensure that you don't have users or groups with names that start with system: by accident.

XiShanYongYe-Chang: remove the `system:` prefix.

artifacts/deploy/admin-clusterrole-aggregation.yaml
```diff
@@ -0,0 +1,141 @@
...
+      - get
+      - list
+      - watch
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  annotations:
+    rbac.authorization.kubernetes.io/autoupdate: "true"
```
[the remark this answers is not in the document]
XiShanYongYe-Chang: [At each start-up, the API server updates default cluster roles with any missing permissions, and updates default cluster role bindings with any missing subjects.](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#auto-reconciliation) This explanation comes from the official Kubernetes documentation, but after testing it on my local kind cluster, I found that it does not take effect (It cannot rule out the possibility of issues with my testing environment and methods.).

I reviewed the relevant code and found that it is related to the `[kubectl auth reconcile](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#kubectl-auth-reconcile)` command. If the value corresponding to this annotation is set to false, the reconcile process will be skipped: 

https://github.com/kubernetes/kubernetes/blob/f0dcf0614036d8c3cd1c9f3b3cf8df4bb1d8e44e/staging/src/k8s.io/component-helpers/auth/rbac/reconciliation/reconcile_role.go#L181

https://github.com/kubernetes/kubernetes/blob/f0dcf0614036d8c3cd1c9f3b3cf8df4bb1d8e44e/staging/src/k8s.io/component-helpers/auth/rbac/reconciliation/reconcile_role.go#L134-L137

Therefore, my conclusion is that it is necessary to set this annotation.

RainbowMango (approved):
/lgtm
/approve

PR #3933, opened 2023-08-11
Extend Cluster API to support multiple zones

### LINKED ISSUE #3413: How does Karmada handle the multi-AZ cluster

RainbowMango:
**What type of PR is this?**

/kind api-change

This PR extends the `Cluster` API by adding a new field `Zones` to represent zones of a member cluster.

Fixes #3413

```release-note
API change: Introduced a new field `Zones` to represent multiple zones of a member cluster.
```

### DIFF
--- a/pkg/apis/cluster/types.go
+++ b/pkg/apis/cluster/types.go
@@ -117,8 +117,18 @@ type ClusterSpec struct {
 	Region string
 
 	// Zone represents the zone of the member cluster locate in.
+	// Deprecated: This filed was never been used by Karmada, and it will not be
+	// removed from v1alpha1 for backward compatibility, use Zones instead.
 	// +optional
 	Zone string
+
+	// Zones represents the failure zones(also called availability zones) of the
+	// member cluster. The zones are presented as a slice to support the case
+	// that cluster runs across multiple failure zones.
+	// Refer https://kubernetes.io/docs/setup/best-practices/multiple-zones/ for
+	// more details about running Kubernetes in multiple zones.
+	// +optional
+	Zones []string `json:"zones,omitempty"`
 
 	// Taints attached to the member cluster.
 	// Taints on the cluster have the "effect" on
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -16357,8 +16357,16 @@
           }
         },
         "zone": {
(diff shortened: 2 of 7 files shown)


pkg/apis/cluster/types.go
```diff
@@ -117,9 +117,18 @@ type ClusterSpec struct {
 	Region string
 
 	// Zone represents the zone of the member cluster locate in.
+	// Deprecated: This filed was never been used, use Zones instead.
 	// +optional
 	Zone string
```
whitewindmills, line 122: In our environment, this field is already in use, when we upgrade Karmada, the value of the `zone` field will not be populated to zones, so the `zones` field is still empty. I wonder if it is necessary to autofill the value of `zone` to `zones` in next pr. If so, we'll only care about `zones` and it's ok even if the `zone` field is removed in the future.

whitewindmills: > Disable the deprecated zone in v1.8 by introducing validation

except for this point, I think others make sense. can we disable the deprecated `zone` field in v1.9 or later to allow more time for users to adapt? we won't upgrade Karmada to v1.7 because our upgrade cycle is usually half a year. so we will choose to upgrade directly to v1.8.

RainbowMango: Sure. Then, we won't introduce the validation.
I will add more comments on the deprecated field to note that the field will never be removed from `v1alpha1`.

RainbowMango: Hi @whitewindmills Please take another look, and I just opened an issue (#3952) to track subsequent tasks. Some of them might be a little bit challenge.

whitewindmills: > and I just opened an issue (https://github.com/karmada-io/karmada/issues/3952) to track subsequent tasks. Some of them might be a little bit challenge.

thanks, I will try some tasks.

whitewindmills (commented):
/lgtm

Issue #3935, opened 2023-08-14
operator init operations may need bigger timeout


Vacant2333:
**What happened**:
When deploying Karmada using the operator, it's important to note that for the three components - karmada-etcd, karmada-apiserver, and karmada-aggregated-apiserver - their timeout is set to 30 seconds. In my environment (and even in most environments), if the images are not pre-pulled beforehand, the allotted time might not be sufficient. As a result, deployment could fail due to timeouts, necessitating the deletion of the Karmada resource before reattempting the deployment.
**What you expected to happen**:
Install success at once
**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:
![83d4c1419dcaa875d2e74015433b0c5d](https://github.com/karmada-io/karmada/assets/19872346/efbb1ef3-6d0c-44b8-8505-60d8b0479914)
![0a71d6f83ce6c6e7d54441112e12d838](https://github.com/karmada-io/karmada/assets/19872346/a7f5cf1d-43fa-4ede-8fc8-1ed9ee49de43)
![7bf89d25c6f9cd31f726b821206899e9](https://github.com/karmada-io/karmada/assets/19872346/65d27f88-f33d-4017-8dee-a01cc28faa4e)
<img width="991" alt="image" src="https://github.com/karmada-io/karmada/assets/19872346/abeecb4c-73cf-4a06-86e7-6b36f70c5689">

XiShanYongYe-Chang:
/cc @calvin0327 @Poor12 @lonelyCZ 
Hi, can you help take a look? Can we modify this time to be configurable?

Vacant2333:
> Which version of karmadactl we are talking about? I remember we have introduced a flag(`--wait-component-ready-timeout`) for this:
> 
> ```shell
> -bash-5.0# karmadactl init --help 
> Install the Karmada control plane in a Kubernetes cluster.
> 
> Options:
>     --wait-component-ready-timeout=120:
> 	Wait for karmada component ready timeout. 0 means wait forever
> 
> Usage:
>   karmadactl init [options]
> 
> Use "karmadactl options" for a list of global command-line options (applies to all commands).
> ```
> 
> Please try with the karmadactl on master.

its install by operator, not karmadactl
https://github.com/karmada-io/karmada/blob/master/operator/README.md

RainbowMango:
oh, sorry my bad.

Vacant2333:
may we can update to 120s, i think its enough at most situations

[...]

PR #3944, opened 2023-08-15
update operator default timeout

Vacant2333:
**What type of PR is this?**

Fixes https://github.com/karmada-io/karmada/issues/3935

```release-note
NONE
```

### DIFF
--- a/operator/pkg/karmadaresource/webhookconfiguration/webhookconfiguration.go
+++ b/operator/pkg/karmadaresource/webhookconfiguration/webhookconfiguration.go
@@ -12,8 +12,8 @@ import (
 	"github.com/karmada-io/karmada/operator/pkg/util/apiclient"
 )
 
-// EnsureWebhookconfiguration creates karmada webhook mutatingWebhookConfiguration and validatingWebhookConfiguration
-func EnsureWebhookconfiguration(client clientset.Interface, namespace, name, caBundle string) error {
+// EnsureWebhookConfiguration creates karmada webhook mutatingWebhookConfiguration and validatingWebhookConfiguration
+func EnsureWebhookConfiguration(client clientset.Interface, namespace, name, caBundle string) error {
 	if err := mutatingWebhookConfiguration(client, namespace, name, caBundle); err != nil {
 		return err
 	}
--- a/operator/pkg/tasks/init/wait.go
+++ b/operator/pkg/tasks/init/wait.go
@@ -14,12 +14,16 @@ import (
 )
 
 var (
+	// The timeout for wait each component be ready
+	// It includes the time for pulling the component image.
+	componentBeReadyTimeout = 120 * time.Second
+
 	etcdLabels                       = labels.Set{"karmada-app": constants.Etcd}
 	karmadaApiserverLabels           = labels.Set{"karmada-app": constants.KarmadaAPIServer}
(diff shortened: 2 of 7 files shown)


operator/pkg/karmadaresource/webhookconfiguration/webhookconfiguration.go
```diff
@@ -12,8 +12,8 @@ import (
 	"github.com/karmada-io/karmada/operator/pkg/util/apiclient"
 )
 
-// EnsureWebhookconfiguration creates karmada webhook mutatingWebhookConfiguration and validatingWebhookConfiguration
-func EnsureWebhookconfiguration(client clientset.Interface, namespace, name, caBundle string) error {
+// EnsureWebhookConfiguration creates karmada webhook mutatingWebhookConfiguration and validatingWebhookConfiguration
```
RainbowMango, line 15: Nice finding!
If you find an issue like this, please feel free to send a PR.

Vacant2333: thanks,  In most cases, I tend to make these changes while submitting a pr

RainbowMango (commented):
/lgtm

Leave approval to @calvin0327 for the final decision.

I see this PR updates the default timeout for APIserver and ETCD, what about the other components?

Vacant2333:
> /lgtm
> 
> Leave approval to @calvin0327 for the final decision.
> 
> I see this PR updates the default timeout for APIserver and ETCD, what about the other components?

During my testing, I found that there was insufficient time to make the changes in these specific areas, while the others were fine.

RainbowMango:
> During my testing, I found that there was insufficient time to make the changes in these specific areas, while the others were fine

It'd be great if we could keep all components with the same timeout.

Vacant2333:
image

I updated the timeout duration on lines 74, 75, and 77 from 30 seconds to 120 seconds. As for the other components, they are all covered on lines 80 and 81, and their default timeout is already set to 120 seconds, so there's no need for updates there. 
Alright, I'll make the modifications accordingly.

operator/pkg/tasks/init/wait.go
```diff
@@ -14,6 +14,8 @@ import (
 )
 
 var (
+	componentBeReadyTimeout = 120 * time.Second
```
calvin0327, line 17: Could you explain why it's 120 seconds?

Vacant2333: By default, they are both set to 30 seconds. Many times, if an image needs to be pulled, it's not sufficient. In my environment, it probably needs to be set to 60 seconds. Of course, I cannot provide an exact timeout, but I believe 120 seconds should be sufficient for use in the vast majority of scenarios.

calvin0327: Could you palease add some comments for the variable `componentBeReadyTimeout` ?

Vacant2333: ok

XiShanYongYe-Chang:
Ask an again review for @calvin0327 @RainbowMango

RainbowMango (commented):
/lgtm

Leave approval to @carlory for final decisions.

carlory:
It makes senses to me. 

/approve

A better way is to retry the reconcile func if an error occur. but it's a bit complicated to do now.

Issue #3946, opened 2023-08-16
feat: added performance metrics for CronFederatedHPA, FederatedHPA, and karmada-metrics-adapter


jwcesign:
**What would you like to be added**:
Added performance metrics for CronFederatedHPA, FederatedHPA, and karmada-metrics-adapter

**Why is this needed**:
In order to better identify relevant performance issues, we need relevant metrics for viewing and locating related performance, including:
CronFederatedHPA:
* the time cost of handling rules.
* the number of rules.
* the time cost of handling CronFederatedHPA.

FederatedHPA:
* the time of handling FederatedHPA.
* the time of pulling metrics.

karmada-metrics-adapter:
* the time of pulling metrics.

jwcesign:
Hi @whitewindmills , done

RainbowMango:
If you want to call for more developers, you might need to describe each task in more detail.

zhy76:
I want to try with this.

whitewindmills:
@zhy76 Hi
I've finished the first two tasks, you can pick the rest. Also hope you can help review my [PR](https://github.com/karmada-io/karmada/pull/3949).

zhy76:
I will checkout the FederatedHPA part.

whitewindmills:
> I will checkout the FederatedHPA part.
👍 
/assign @zhy76

chl178:
@whitewindmills Can I try the last one?  I'm new to the community and may need more time.

jwcesign:
> Can I try the last one? I'm new to the community and may need more time.

Sure, thanks a lot @chl178, and welcome to Karmada!

PR #3949, opened 2023-08-17
feat: add cron federated hpa metrics

whitewindmills:
**What type of PR is this?**
/kind feature

add cron federated hpa metrics

part of https://github.com/karmada-io/karmada/issues/3946

CronFederatedHPA:
* the time cost of handling rules.
* the time cost of handling CronFederatedHPA.

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/cronfederatedhpa/cronfederatedhpa_controller.go
+++ b/pkg/controllers/cronfederatedhpa/cronfederatedhpa_controller.go
@@ -15,6 +15,7 @@ package cronfederatedhpa
 
 import (
 	"context"
+	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
@@ -28,6 +29,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 
 	autoscalingv1alpha1 "github.com/karmada-io/karmada/pkg/apis/autoscaling/v1alpha1"
+	"github.com/karmada-io/karmada/pkg/metrics"
 	"github.com/karmada-io/karmada/pkg/sharedcli/ratelimiterflag"
 	"github.com/karmada-io/karmada/pkg/util/helper"
 )
@@ -70,6 +72,10 @@ func (c *CronFHPAController) Reconcile(ctx context.Context, req controllerruntim
 		return controllerruntime.Result{}, nil
 	}
 
+	var err error
+	startTime := time.Now()
+	defer metrics.ObserveProcessCronFederatedHPALatency(err, startTime)
+
 	origRuleSets := sets.New[string]()
 	for _, history := range cronFHPA.Status.ExecutionHistories {
 		origRuleSets.Insert(history.RuleName)
@@ -84,7 +90,7 @@ func (c *CronFHPAController) Reconcile(ctx context.Context, req controllerruntim
 
 	newRuleSets := sets.New[string]()
 	for _, rule := range cronFHPA.Spec.Rules {
-		if err := c.processCronRule(cronFHPA, rule); err != nil {
+		if err = c.processCronRule(cronFHPA, rule); err != nil {
 			return controllerruntime.Result{Requeue: true}, err
 		}
 		newRuleSets.Insert(rule.Name)
@@ -96,7 +102,7 @@ func (c *CronFHPAController) Reconcile(ctx context.Context, req controllerruntim
 			continue
 		}
 		c.CronHandler.StopRuleExecutor(req.NamespacedName.String(), name)
-		if err := c.removeCronFHPAHistory(cronFHPA, name); err != nil {
+		if err = c.removeCronFHPAHistory(cronFHPA, name); err != nil {
 			return controllerruntime.Result{Requeue: true}, err
 		}
 	}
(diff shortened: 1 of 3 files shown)


whitewindmills:
invite @jwcesign to help review.
/cc @jwcesign

pkg/controllers/cronfederatedhpa/cronfederatedhpa_controller.go
```diff
@@ -70,6 +72,9 @@ func (c *CronFHPAController) Reconcile(ctx context.Context, req controllerruntim
 		return controllerruntime.Result{}, nil
 	}
 
+	var err error
+	defer metrics.ObserveProcessCronFederatedHPALatency(err, time.Now())
```
jwcesign, line 76: This approach is correct, but it is very confusing， can we do like this:
```go
startTime := time.Now()
defer metrics.ObserveProcessCronFederatedHPALatency(err, startTime)
```

jwcesign (commented):
other lgtm

whitewindmills:
> other lgtm

done

jwcesign:
/lgtm 
Can  you help update the related documents: https://karmada.io/docs/next/reference/instrumentation/metrics/#metrics

RainbowMango:
@whitewindmills @jwcesign Have you tested it? What do these metrics look like?

jwcesign:
I tested, it works fine:
image

pkg/metrics/resource.go
```diff
@@ -9,12 +9,14 @@ import (
...
-	syncWorkloadDurationMetricsName        = "work_sync_workload_duration_seconds"
-	policyPreemptionMetricsName            = "policy_preemption_total"
+	resourceMatchPolicyDurationMetricsName  = "resource_match_policy_duration_seconds"
+	resourceApplyPolicyDurationMetricsName  = "resource_apply_policy_duration_seconds"
+	policyApplyAttemptsMetricsName          = "policy_apply_attempts_total"
+	syncWorkDurationMetricsName             = "binding_sync_work_duration_seconds"
+	syncWorkloadDurationMetricsName         = "work_sync_workload_duration_seconds"
+	policyPreemptionMetricsName             = "policy_preemption_total"
+	cronFederatedHPADurationMetricsName     = "cron_federated_hpa_process_duration_seconds"
+	cronFederatedHPARuleDurationMetricsName = "cron_federated_hpa_rule_process_duration_seconds"
```
RainbowMango, line 19:
```suggestion
	cronFederatedHPADurationMetricsName     = "cronfederatedhpa_process_duration_seconds"
	cronFederatedHPARuleDurationMetricsName = "cronfederatedhpa_rule_process_duration_seconds"
```
Do you think this makes it better?

pkg/metrics/resource.go
```diff
@@ -51,6 +53,18 @@ var (
 		Name: policyPreemptionMetricsName,
 		Help: "Number of preemption for the resource template. By the result, 'error' means a resource template failed to be preempted by other propagation policies. Otherwise 'success'.",
 	}, []string{"result"})
+
+	cronFederatedHPADurationHistogram = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Name:    cronFederatedHPADurationMetricsName,
+		Help:    "Duration in seconds to process a cron federated HPA. By the result, 'error' means a cron federated HPA failed to be processed. Otherwise 'success'.",
```
RainbowMango, line 59:
```suggestion
		Help:    "Duration in seconds to process a CronFederatedHPA. By the result, 'error' means a CronFederatedHPA failed to be processed. Otherwise 'success'.",
```
ditto.

pkg/metrics/resource.go
```diff
@@ -51,6 +53,18 @@ var (
...
+		Buckets: prometheus.ExponentialBuckets(0.001, 2, 12),
+	}, []string{"result"})
+
+	cronFederatedHPARuleDurationHistogram = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Name:    cronFederatedHPARuleDurationMetricsName,
+		Help:    "Duration in seconds to process a cron federated HPA rule. By the result, 'error' means a cron federated HPA rule failed to be processed. Otherwise 'success'.",
```
RainbowMango, line 65:
```suggestion
		Help:    "Duration in seconds to process a CronFederatedHPA rule. By the result, 'error' means a CronFederatedHPA rule failed to be processed. Otherwise 'success'.",
```

RainbowMango (commented):
Generally looks good, just a suggestion for naming.

pkg/metrics/resource.go
RainbowMango: I know previous metrics don't follow this rule, but I think they should. 
What do you think?

whitewindmills: do we need to change previous metrics to follow this rule?

RainbowMango: I think so, but it's up to you to do it or not by this PR.

whitewindmills: I accept this rule but may change other metrics to follow this rule in another PR.

RainbowMango: OK, thanks.

RainbowMango: Then, let's move this forward, and update them by another PR.

RainbowMango (approved):
/approve

PR #3955, opened 2023-08-18
helm: support PodDisruptionBudget for resources

a7i:
**What type of PR is this?**

/kind feature

Currently there's not a way to render PodDisruptionBudget for critical resources such as kube-apiserver

Fixes #

```release-note
helm: support PodDisruptionBudget for resources
```

### DIFF
--- a/charts/karmada/values.yaml
+++ b/charts/karmada/values.yaml
@@ -19,6 +19,9 @@ global:
 ## for more details about yaml anchors and aliases.
 karmadaImageVersion: &karmadaImageVersion latest
 
+podDisruptionBudget: &podDisruptionBudget {}
+  # maxUnavailable: 20%
+
 ## @param installMode "host" and "agent" are provided
 ## "host" means install karmada in the control-cluster
 ## "agent" means install agent client in the member cluster
@@ -192,6 +195,8 @@ scheduler:
     rollingUpdate:
       maxUnavailable: 0
       maxSurge: 50%
+  ## @param apiServer.podDisruptionBudget
+  podDisruptionBudget: *podDisruptionBudget
 
 ## webhook config
 webhook:
@@ -249,6 +254,8 @@ webhook:
     rollingUpdate:
       maxUnavailable: 0
       maxSurge: 50%
+  ## @param apiServer.podDisruptionBudget
+  podDisruptionBudget: *podDisruptionBudget
 
 ## controller manager config
 controllerManager:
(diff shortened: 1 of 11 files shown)


RainbowMango (commented):
/assign 

Thanks @a7i I'll take a look ASAP.

charts/karmada/values.yaml
```diff
@@ -19,6 +19,9 @@ global:
 ## for more details about yaml anchors and aliases.
 karmadaImageVersion: &karmadaImageVersion latest
 
+podDisruptionBudget: &podDisruptionBudget {}
+  # maxUnavailable: 20%
```
RainbowMango, line 23: Does this mean all components will share the same budget?

a7i: Each resource has its own `podDisruptionBudget` configuration which can get an override. This was following the existing pattern of `karmadaImageVersion` on line 20 but happy to remove the anchor tag if you suggest though.

RainbowMango: Looks good to me.

RainbowMango (approved):
/lgtm
/approve

chaosi-zju:
hello @a7i, thank you for your important commit! Though the PR is merged, I wanna to bother you some questions:

1. you seems like added a empty `PodDisruptionBudget` config to chart template, why not add a config like `minAvailable: 1`? On the one hand it seems more meaningful, on the other hand it can provide a example to users, for users may not familiar with the format and rule of `anchor`.

2. didn't you test it by local helm install, and in your testing how did you configured the value.yaml?

Thank you again~

RainbowMango:
> why not add a config like minAvailable: 1? On the one hand it seems more meaningful, 

I guess not all users need PDB, that's why take empty as default.

a7i:
1. @RainbowMango is exactly right. If you have a single replica for example, `minAvailable: 1` will block node rollout / cluster upgrades. As a result, PDB should only be defined for those who require high-availability and fault-tolerance.

2. It can be tested inline
```
  helm template host . --set controllerManager.podDisruptionBudget.maxUnavailable="20%" --set webhook.podDisruptionBudget.maxUnavailable="20%" --set apiServer.podDisruptionBudget.maxUnavailable="20%" --set kubeControllerManager.podDisruptionBudget.maxUnavailable="20%"
```
or by supplying your own values.yaml

chaosi-zju:
ok, got it~

PR #3965, opened 2023-08-21
Response to priority change of PropagationPolicy

RainbowMango:
**What type of PR is this?**

/kind feature

This PR provides a temporary solution for corner case: 
```
After the priority(.spec.priority) of PropagationPolicy changed from high priority (e.g. 5) to 
low priority(e.g. 3), we should try to check if there is a PropagationPolicy(e.g. with priority 4) 
that could preempt the targeted resources.
```

Fixes #

See https://github.com/karmada-io/karmada/pull/3684#issuecomment-1645075275 for the background discussion.

In addition, this PR focuses on dealing with `PropagationPolicy`, if it is acceptable, the `ClusterPropagationPolicy` should do a similar thing.

Test steps:
1. Apply two policies (`foo` with priority 5 and `bar` with priority 4) as follows
```yaml
kind: PropagationPolicy
metadata:
  name: foo
spec:
  priority: 5
  preemption: Always
  resourceSelectors:
    - apiVersion: apps/v1
      kind: Deployment
      name: nginx
  placement:
    clusterAffinity:
      clusterNames:
        - member1
---
kind: PropagationPolicy
metadata:
  name: bar
spec:
  priority: 4
  preemption: Always
  resourceSelectors:
    - apiVersion: apps/v1
      kind: Deployment
      name: nginx
  placement:
    clusterAffinity:
      clusterNames:
        - member2
```
2. Apply a Deployment as [this sample](https://github.com/karmada-io/karmada/blob/master/samples/nginx/deployment.yaml).
```
kubectl apply -f samples/nginx/deployment.yaml
```

3. The Deployment should be propagated by `foo` as it has higher priority
```
apiVersion: apps/v1
kind: Deployment
metadata:
  generation: 1
  labels:
    app: nginx
    propagationpolicy.karmada.io/name: foo
  name: nginx
```

4. Update the priority of `foo` from `5` to `3`, then the Deployment should be preempted by `bar`.
```
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: nginx
    propagationpolicy.karmada.io/name: bar
```
meanwhile, I can see the relevant logs:
```
I0821 13:28:39.213404       1 preemption.go:263] Enqueuing PropagationPolicy(default/bar) in case of PropagationPolicy(default/foo) priority changes
...
I0821 13:28:39.226380       1 preemption.go:127] Propagation policy(default/bar) has preempted another propagation policy(default/foo).
```

```release-note
NONE(Part of feature #3787)
```

### DIFF
--- a/pkg/detector/detector.go
+++ b/pkg/detector/detector.go
@@ -31,6 +31,7 @@ import (
 	policyv1alpha1 "github.com/karmada-io/karmada/pkg/apis/policy/v1alpha1"
 	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
 	"github.com/karmada-io/karmada/pkg/events"
+	"github.com/karmada-io/karmada/pkg/features"
 	"github.com/karmada-io/karmada/pkg/metrics"
 	"github.com/karmada-io/karmada/pkg/resourceinterpreter"
 	"github.com/karmada-io/karmada/pkg/sharedcli/ratelimiterflag"
@@ -783,7 +784,7 @@ func (d *ResourceDetector) OnPropagationPolicyAdd(obj interface{}) {
 }
 
 // OnPropagationPolicyUpdate handles object update event and push the object to queue.
-func (d *ResourceDetector) OnPropagationPolicyUpdate(_, newObj interface{}) {
+func (d *ResourceDetector) OnPropagationPolicyUpdate(oldObj, newObj interface{}) {
 	key, err := ClusterWideKeyFunc(newObj)
 	if err != nil {
 		return
@@ -791,6 +792,52 @@ func (d *ResourceDetector) OnPropagationPolicyUpdate(_, newObj interface{}) {
 
 	klog.V(2).Infof("Update PropagationPolicy(%s)", key)
 	d.policyReconcileWorker.Add(key)
+
+	// Temporary solution of corner case: After the priority(.spec.priority) of
+	// PropagationPolicy changed from high priority (e.g. 5) to low priority(e.g. 3),
+	// we should try to check if there is a PropagationPolicy(e.g. with priority 4)
+	// could preempt the targeted resources.
+	//
+	// Recognized limitations of the temporary solution are:
+	// - Too much logical processed in an event handler function will slow down
+	//   the overall reconcile speed.
+	// - If there is an error raised during the process, the event will be lost
+	//   and no second chance to retry.
+	//
+	// The idea of the long-term solution, perhaps PropagationPolicy could have
+	// a status, in that case we can record the observed priority(.status.observedPriority)
+	// which can be used to detect priority changes during reconcile logic.
+	if features.FeatureGate.Enabled(features.PolicyPreemption) {
+		var unstructuredOldObj *unstructured.Unstructured
+		var unstructuredNewObj *unstructured.Unstructured
+
+		unstructuredOldObj, err = helper.ToUnstructured(oldObj)
+		if err != nil {
+			klog.Errorf("Failed to transform oldObj, error: %v", err)
+			return
+		}
+		unstructuredNewObj, err = helper.ToUnstructured(newObj)
+		if err != nil {
+			klog.Errorf("Failed to transform newObj, error: %v", err)
+			return
+		}
+
+		var oldPolicy policyv1alpha1.PropagationPolicy
+		var newPolicy policyv1alpha1.PropagationPolicy
+
+		if err = helper.ConvertToTypedObject(unstructuredOldObj, &oldPolicy); err != nil {
+			klog.Errorf("Failed to convert typed PropagationPolicy(%s/%s): %v", unstructuredOldObj.GetNamespace(), unstructuredOldObj.GetName(), err)
+			return
+		}
+		if err = helper.ConvertToTypedObject(unstructuredNewObj, &newPolicy); err != nil {
+			klog.Errorf("Failed to convert typed PropagationPolicy(%s/%s): %v", newPolicy.GetNamespace(), newPolicy.GetName(), err)
+			return
+		}
+
+		if newPolicy.ExplicitPriority() < oldPolicy.ExplicitPriority() {
+			d.HandleDeprioritizedPropagationPolicy(oldPolicy, newPolicy)
+		}
+	}
 }
 
 // OnPropagationPolicyDelete handles object delete event and push the object to queue.
--- a/pkg/detector/preemption.go
+++ b/pkg/detector/preemption.go
@@ -4,6 +4,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/labels"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/klog/v2"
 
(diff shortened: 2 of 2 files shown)


pkg/detector/preemption.go
```diff
@@ -227,3 +228,41 @@ func (d *ResourceDetector) fetchResourceTemplate(rs policyv1alpha1.ResourceSelec
...
+// HandleDeprioritizedPropagationPolicy responses to priority change of a PropagationPolicy,
+// if the change is from high priority (e.g. 5) to low priority(e.g. 3), it will
+// check if there is another PropagationPolicy could preempt the targeted resource,
+// and put the PropagationPolicy in the queue to trigger preemption.
+func (d *ResourceDetector) HandleDeprioritizedPropagationPolicy(oldPolicy policyv1alpha1.PropagationPolicy, newPolicy policyv1alpha1.PropagationPolicy) {
+	if (newPolicy.Spec.Priority != nil && oldPolicy.Spec.Priority != nil) && *newPolicy.Spec.Priority < *oldPolicy.Spec.Priority {
+		klog.Infof("PropagationPolicy(%s/%s) priority changed from %d to %d", newPolicy.GetNamespace(), newPolicy.GetName(), *oldPolicy.Spec.Priority, *newPolicy.Spec.Priority)
+		policies, err := d.propagationPolicyLister.ByNamespace(newPolicy.GetNamespace()).List(labels.Everything())
+		if err != nil {
+			klog.Errorf("Failed list PropagationPolicy from namespace: %s, error: %v", newPolicy.GetNamespace(), err)
```
Poor12, line 241:
```suggestion
			klog.Errorf("Failed to list PropagationPolicy from namespace: %s, error: %v", newPolicy.GetNamespace(), err)
```

pkg/detector/preemption.go
```diff
@@ -227,3 +228,41 @@ func (d *ResourceDetector) fetchResourceTemplate(rs policyv1alpha1.ResourceSelec
...
+
+		for i := range policies {
+			var potentialPolicy policyv1alpha1.PropagationPolicy
+			if err = helper.ConvertToTypedObject(policies[i], &potentialPolicy); err != nil {
+				klog.Errorf("Failed to convert typed PropagationPolicy: %v", err)
+				continue
+			}
+			if potentialPolicy.Spec.Priority != nil &&
+				potentialPolicy.Spec.Preemption == policyv1alpha1.PreemptAlways &&
+				*potentialPolicy.Spec.Priority > *newPolicy.Spec.Priority { // potential policy must have higher priority.
```
Poor12, line 257: I think potentialPolicy's priority must be <= oldPolicy.Spec.Priority. Otherwise, there have been occured the preemption.

pkg/detector/preemption.go
RainbowMango: nice finding

pkg/detector/preemption.go
RainbowMango: Updated.
We can save the effort for this case.

pkg/detector/preemption.go
```diff
@@ -227,3 +228,48 @@ func (d *ResourceDetector) fetchResourceTemplate(rs policyv1alpha1.ResourceSelec
 	}
 	return resourceTemplate, nil
 }
+
+// HandleDeprioritizedPropagationPolicy responses to priority change of a PropagationPolicy,
+// if the change is from high priority (e.g. 5) to low priority(e.g. 3), it will
+// check if there is another PropagationPolicy could preempt the targeted resource,
+// and put the PropagationPolicy in the queue to trigger preemption.
+func (d *ResourceDetector) HandleDeprioritizedPropagationPolicy(oldPolicy policyv1alpha1.PropagationPolicy, newPolicy policyv1alpha1.PropagationPolicy) {
+	if (newPolicy.Spec.Priority != nil && oldPolicy.Spec.Priority != nil) && *newPolicy.Spec.Priority < *oldPolicy.Spec.Priority {
```
whitewindmills, line 237: use `newPolicy.ExplicitPriority() <= oldPolicy.ExplicitPriority()` instead.

pkg/detector/preemption.go
```diff
@@ -227,3 +228,48 @@ func (d *ResourceDetector) fetchResourceTemplate(rs policyv1alpha1.ResourceSelec
...
+		klog.Infof("PropagationPolicy(%s/%s) priority changed from %d to %d", newPolicy.GetNamespace(), newPolicy.GetName(), *oldPolicy.Spec.Priority, *newPolicy.Spec.Priority)
+		policies, err := d.propagationPolicyLister.ByNamespace(newPolicy.GetNamespace()).List(labels.Everything())
+		if err != nil {
+			klog.Errorf("Failed to list PropagationPolicy from namespace: %s, error: %v", newPolicy.GetNamespace(), err)
+			return
+		}
+
+		// TODO(@RainbowMango): Should sort the listed policies to ensure the
+		// higher priority PropagationPolicy be process first to avoid possible
+		// multiple preemption.
```
whitewindmills, line 247: I think it's necessary. in the end, only the policy with the highest priority and preemption-enabled can preempt it, and we don’t need to re-enqueue the rest of the policies.

pkg/detector/preemption.go
RainbowMango: +1, totally agree, thanks for the reminder.

pkg/detector/preemption.go
RainbowMango: Not exactly, since we don't know if the potential policy matches the target resource, so we have to en-queue all of them.

These comments want to address the case that there is more than one policy that could perform the preemption, but with different priorities.  We'd better ensure en-queue the high-priority policy first, then the low-priority policy won't perform the preemption. Otherwise, the high-priority policy may preempt the low-priority policy again.(That's what I mean multiple preemption).

Note that, en-queue high-priority policy can't guarantee high-priority policy could perform preemption ahead of the low-priority one. It depends on the speed of the reconciliation routine. That's just a relief.

RainbowMango: I leave the TODO here because we don't have convenient helper functions to sort the policy yet, i think we can get back on the TODO task after that.

whitewindmills: > I leave the TODO here because we don't have convenient helper functions to sort the policy yet, i think we can get back on the TODO task after that.

+1
> Not exactly, since we don't know if the potential policy matches the target resource, so we have to en-queue all of them.

we can resolve this problem by checking if the intersection of their `ResourceSelector` exists.

RainbowMango: Yes we can, but I'm hesitant to do it for this `temporary` solution. I hope we can keep it simple.

pkg/detector/preemption.go
RainbowMango: `<=` or `<`?
I guess it should be `<`.

pkg/detector/preemption.go
whitewindmills: Given that it's a temporary solution, I totally agree.

RainbowMango:
@wuyingjun-lucky Do you have any comments? I see you are looking at it :)

pkg/detector/preemption.go
whitewindmills: > I guess it should be <.

right, sorry, it's my mistake.

pkg/detector/detector.go
```diff
@@ -783,14 +784,60 @@ func (d *ResourceDetector) OnPropagationPolicyAdd(obj interface{}) {
...
+	//
+	// Recognized limitations of the temporary solution are:
+	// - Too much logical processed in an event handler function will slow down
+	//   the overall reconcile speed.
+	// - If there is an error raised during the process, the event will be lost
+	//   and no second chance to retry.
+	//
+	// The idea of the long-term solution, perhaps PropagationPolicy could have
+	// a status, in that case we can record the observed priority(.status.observedPriority)
+	// which can be used to detect priority changes during reconcile logic.
```
whitewindmills, line 809: it's a good idea.

wuyingjun-lucky:
LGTM and I want have a try on “multiple preemption”

RainbowMango:
> LGTM and I want have a try on “multiple preemption”

No, you don't, haha, that's a chaotic state(but will stabilize eventually.)

PR #3966, opened 2023-08-21
fix addAnnotationWithClusterName crash

### LINKED ISSUE #3960: karmada-search may crash due to `concurrent map read and map write`
**What happened**:

`search` pod crashed, `fatal error: concurrent map read and map write`.

If errors occurred, log detail, or if caused by concurrent map read and map write, add lock to avoid it.

No idea. :rofl:

[search.log](https://github.com/karmada-io/karmada/files/12392927/search.log)

The log didn't point out exactly where caused this, it makes sense, because when we talk about concurrent map read and map write, it must be at least 2 place, reading or writing, so it's not caused by a particular line of code, it should be caused by some code happened at a same time.

https://github.com/karmada-io/karmada/blob/5f1664060f271de3a86a2416ea7eedf3990d919f/pkg/registry/search/storage/cache.go#L142-L159

My point of view is that, it may caused by line `152`, adding entry to annotations map, meanwhile other logic reading it, because `runtime.Object` is a reference, not a value.

**Environment**:
- Karmada version: v1.6.1
- Others:

ikaven1024:
**What type of PR is this?**
/kind bug

Fixes #3960

When listing from an  Infromer, it returns reference of objects in `cache`.
https://github.com/karmada-io/karmada/blob/48a37ab5b1f535eaa2c70e3b6ca52909fa78372e/vendor/k8s.io/client-go/tools/cache/thread_safe_store.go#L245-L253

When two or more search requests come at the same time, and all requests list objects from informer and add cluster annotation into the same objecs, it raises crash.
https://github.com/karmada-io/karmada/blob/5763248181ed3a90684b36b4d5d2ffea2c68d987/pkg/registry/search/storage/cache.go#L141

```release-note
`karmada-search`: Fixed a panic due to concurrent mutating objects in the informer cache.
```

### DIFF
--- a/pkg/registry/search/storage/cache.go
+++ b/pkg/registry/search/storage/cache.go
@@ -119,7 +119,7 @@ func (r *SearchREST) getObjectItemsFromClusters(
 					objGVR, namespace, name, cluster.Name, err)
 				continue
 			}
-			items = append(items, addAnnotationWithClusterName([]runtime.Object{resourceObject}, cluster.Name)...)
+			items = append(items, addAnnotationWithClusterName([]runtime.Object{resourceObject.DeepCopyObject()}, cluster.Name)...)
 		} else {
 			var resourceObjects []runtime.Object
 			if len(namespace) > 0 {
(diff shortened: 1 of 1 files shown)


RainbowMango (commented):
/assign

@ikaven1024 Thank you so much for the quick response and hard work!

RainbowMango (approved):
/lgtm
/approve

RainbowMango:
Just opened #3968 to track the backport tasks.

PR #3974, opened 2023-08-22
fix: return 400(StatusBadRequest) http code when the validation failed

jwcesign:
**What type of PR is this?**
/kind bug

Currently, the webhook returns a 403 (forbidden) status code when validation fails. However, this is incorrect as it should only occur when there is insufficient privilege.

I have tested this with K8s and found that if the field is not correct, the expected status codes are either 400(StatusBadRequest) or 422(StatusUnprocessableEntity). Therefore, we should adhere to this rule.

There is no clear distinction between 400 and 422. In both cases, a parameter error occurs. Therefore, I have modified the relevant code to always return a response of 400.

Fixes #none

```release-note
`karmada-webhook`: If the field validation fails, return a 400 (StatusBadRequest) HTTP code.
```

### DIFF
--- a/pkg/webhook/clusteroverridepolicy/validating.go
+++ b/pkg/webhook/clusteroverridepolicy/validating.go
@@ -32,7 +32,7 @@ func (v *ValidatingAdmission) Handle(_ context.Context, req admission.Request) a
 
 	if errs := validation.ValidateOverrideSpec(&policy.Spec); len(errs) != 0 {
 		klog.Error(errs)
-		return admission.Denied(errs.ToAggregate().Error())
+		return admission.Errored(http.StatusBadRequest, errs.ToAggregate())
 	}
 
 	return admission.Allowed("")
--- a/pkg/webhook/clusterpropagationpolicy/validating.go
+++ b/pkg/webhook/clusterpropagationpolicy/validating.go
@@ -42,14 +42,14 @@ func (v *ValidatingAdmission) Handle(_ context.Context, req admission.Request) a
 		if policy.Spec.SchedulerName != oldPolicy.Spec.SchedulerName {
 			err = fmt.Errorf("the schedulerName should not be updated")
 			klog.Error(err)
-			return admission.Denied(err.Error())
+			return admission.Errored(http.StatusBadRequest, err)
 		}
 	}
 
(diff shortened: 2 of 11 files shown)


XiShanYongYe-Chang (commented):
Thanks~
/lgtm

RainbowMango:
Hi @jwcesign Could you double confirm it. 
I just found the definition of HTPP code 403 from [rfc9110](https://www.rfc-editor.org/rfc/rfc9110.html#name-403-forbidden):
```
The 403 (Forbidden) status code indicates that the server understood the request but refuses to fulfill it. A server that wishes to make public why the request has been forbidden can describe that reason in the response content (if any).

If authentication credentials were provided in the request, the server considers them insufficient to grant access. The client SHOULD NOT automatically repeat the request with the same credentials. The client MAY repeat the request with new or different credentials. However, a request might be forbidden for reasons unrelated to the credentials.

An origin server that wishes to "hide" the current existence of a forbidden [target resource](https://www.rfc-editor.org/rfc/rfc9110.html#target.resource) MAY instead respond with a status code of [404 (Not Found)](https://www.rfc-editor.org/rfc/rfc9110.html#status.404).
```
Note that `However, a request might be forbidden for reasons unrelated to the credentials.`

I guess that's the reason why [controller-runtime take `403` as the default code](https://github.com/kubernetes-sigs/controller-runtime/pull/400).

jwcesign:
While reviewing the K8s code, I noticed that when there are invalid fields, it triggers the `errors.NewInvalid` function to return errors:

https://github.com/kubernetes/apiserver/blob/2538d7ca48d0ea82219d2fd963d728f7aecbacdf/pkg/registry/rest/create.go#L123

The `NewInvalid` function is quite old, as it originates from the first version of K8s (release-0.4).

https://github.com/kubernetes/kubernetes/commit/34c40e4e482e104bc78605622621978fecc77bf1

jwcesign:
Based on the information provided in this link (https://github.com/karmada-io/karmada/pull/3974#issuecomment-1688082666), it is recommended to close this issue.

RainbowMango:
Yes, I think so. We are lack of evidence to accept this change. Thanks @jwcesign all the same~

Issue #3952, opened 2023-08-18
umbrella issue: deprecating zones in Cluster API


RainbowMango:
**What would you like to be added**:
This issue tracks subsequent tasks of [deprecating zone with zones](https://github.com/karmada-io/karmada/pull/3933).

- [x] Add validation for `zones`, possibly at [here](https://github.com/karmada-io/karmada/blob/276d6a8ad1f156546a3208aefb26ceb3252c3a7f/pkg/apis/cluster/validation/validation.go#L152-L164). (@zishen, #3975)
In addition, we should ensure that `zone` and `zones` will not co-exist.
- [x] Transform potential `zone` to `zones`, possibly at [here](https://github.com/karmada-io/karmada/blob/276d6a8ad1f156546a3208aefb26ceb3252c3a7f/pkg/registry/cluster/strategy.go#L129-L132). (@zishen, #3983)
- [x] Deprecate the [--cluster-zone flag](https://github.com/karmada-io/karmada/blob/276d6a8ad1f156546a3208aefb26ceb3252c3a7f/pkg/karmadactl/join/join.go#L131) of `join` and introduce a new flag `--cluster-zones`. (@whitewindmills, #3995)
- [x] Adopt `zones` in scheduler spread by constraints, possibly at [here](https://github.com/karmada-io/karmada/blob/276d6a8ad1f156546a3208aefb26ceb3252c3a7f/pkg/scheduler/core/spreadconstraint/group_clusters.go#L133-L161). (@zishen, #4002)
- [x] Adopt `zones` in scheduler filter, possibly at [here](https://github.com/karmada-io/karmada/blob/276d6a8ad1f156546a3208aefb26ceb3252c3a7f/pkg/scheduler/framework/plugins/spreadconstraint/spread_constraint.go#L44-L46). (@whitewindmills, #3996)
- [x] Adopt `zones` in resource selector, possibly at [here](https://github.com/karmada-io/karmada/blob/276d6a8ad1f156546a3208aefb26ceb3252c3a7f/pkg/util/selector.go#L113-L116)  (@whitewindmills, #4010)

**Why is this needed**:

whitewindmills:
> Adopt zones in scheduler spread by constraints

maybe we don't need this task because zone spread constraint has not been supported in the master branch. I'll consider `zones` fields in the new spread-constraint [pr](https://github.com/karmada-io/karmada/pull/3662).

RainbowMango:
> Adopt zones in resource selector,

[...]

PR #3983, opened 2023-08-23
transform zone to zones

zishen:
What type of PR is this?

/kind cleanup

transform zone to zones

Which issue(s) this PR fixes:
part of https://github.com/karmada-io/karmada/issues/3952

@XiShanYongYe-Chang @whitewindmills

none

### DIFF
--- a/pkg/apis/cluster/mutation/mutation.go
+++ b/pkg/apis/cluster/mutation/mutation.go
@@ -18,11 +18,20 @@ const (
 
 // MutateCluster mutates required fields of the Cluster.
 func MutateCluster(cluster *clusterapis.Cluster) {
-	MutateClusterTaints(cluster.Spec.Taints)
+	mutateClusterTaints(cluster.Spec.Taints)
+	migrateZoneToZones(cluster)
 }
 
-// MutateClusterTaints add TimeAdded field for cluster NoExecute taints only if TimeAdded not set.
-func MutateClusterTaints(taints []corev1.Taint) {
+// migrateZoneToZones add zones field for cluster if Zones not set but Zone set only.
+func migrateZoneToZones(cluster *clusterapis.Cluster) {
+	if cluster.Spec.Zone != "" && len(cluster.Spec.Zones) == 0 {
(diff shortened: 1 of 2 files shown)


zishen:
@whitewindmills @RainbowMango 

I cleared the zones when I set them because they can't exist at the same time.

I am not sure whether this change is reasonable. Could you please confirm it according to the business?

pkg/apis/cluster/mutation/mutation.go
```diff
@@ -18,11 +18,20 @@ const (
...
 func MutateCluster(cluster *clusterapis.Cluster) {
-	MutateClusterTaints(cluster.Spec.Taints)
+	mutateClusterTaints(cluster.Spec.Taints)
+	mutateClusterZone(cluster)
 }
 
-// MutateClusterTaints add TimeAdded field for cluster NoExecute taints only if TimeAdded not set.
-func MutateClusterTaints(taints []corev1.Taint) {
+// mutateClusterZone add zones field for cluster if Zones not set but Zone set only.
+func mutateClusterZone(cluster *clusterapis.Cluster) {
```
whitewindmills, line 26: could you use a more appropriate function name? such as `setDefaultClusterZones`.

whitewindmills:
> I am not sure whether this change is reasonable. Could you please confirm it according to the business?

I think it's okay.

pkg/apis/cluster/mutation/mutation.go
RainbowMango: How about the `MigrateZoneToZones()`, no big deal since it will be removed eventually.

whitewindmills: it's okay 😃

RainbowMango:
> I cleared the zones when I set them because they can't exist at the same time.

I guess you mean `cleared the .spec.zone` during migration. 
Yes, I think we should do that.

whitewindmills (commented):
/lgtm

RainbowMango (approved):
/approve

Issue #3959, opened 2023-08-21
After the failure cluster is recovered, the residual resources are not deleted


XiShanYongYe-Chang:
**What happened**:

As described in comment: https://github.com/karmada-io/karmada/pull/3808#discussion_r1295679640

**What you expected to happen**:

After the failure cluster is recovered, these residual resources can be deleted normally.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Karmada version:
- kubectl-karmada or karmadactl version (the result of `kubectl-karmada version` or `karmadactl version`):
- Others:

XiShanYongYe-Chang:
/cc @chaunceyjiang @lxtywypc @zach593 
Can you help take a look?

zach593:
how about let execution-controller watch cluster object's unready -> ready? if we are considering enqueue too often, `lastTransitionTime` could help determine how long the cluster was unavailable.

XiShanYongYe-Chang:
After the cluster is recovered, will all resources on the cluster be resynchronized? If so, will the pressure of the controller suddenly surge?

What does this strategy specifically refer to?
https://github.com/karmada-io/karmada/pull/3808#discussion_r1296588633

zach593:
> After the cluster is recovered, will all resources on the cluster be resynchronized? If so, will the pressure of the controller suddenly surge?

So I got a thought for a while, that make objectWatcher use `DeepEqual()` checks if there is content changes or not(mutating webhook and API compatibility between karmada and member clusters might be the problem), then it can reduce the frequency of updating member clusters when execution-controller reconciling.

> What does this strategy specifically refer to?
> [#3808 (comment)](https://github.com/karmada-io/karmada/pull/3808#discussion_r1296588633)

This makes every affected items ratelimited, clearing items from the ratelimiter may take a lot of time. And if one day we might let user control the options of ratelimiter and number of async worker retries (or just alter to controller-runtime). In that case, recovery mechanism that rely on dynamic parameters is unreliable.

On the other hand, there's no much difference between watch cluster objects with async worker/controller-runtime's retry mechanism, they both trigger the execution-controller reconciling.

chaunceyjiang:
/cc @XiShanYongYe-Chang @lxtywypc @RainbowMango

A new version is about to be released. If this issue is not resolved, I think we may need to revert #3808.

RainbowMango:
[...]

PR #3994, opened 2023-08-24
remove unnecessary needUpdate to avoid unexpected skipping updating before cache synced

lxtywypc:
**What type of PR is this?**  
/kind bug  

Consider if some error occured in `execution-controller`, it would update the condition of `work` from `true` to `false`, and requeue in a short time(5ms for first time as default now), but during this reconciling, the condition of got `work` might be still `true` because the cache of `work` might not be synced yet. If no error occurs this time, when calling `needUpdateCondition`, we would find the condition in both "old" and "new" `work status` is `true`, that would cause unexpected skipping updating `work status` with the condition is `false` in cluster in fact.  

Fixes #

I'm sorry for bringing this in #3808 , I intended to reduce the attempts to update `work` because now there are much more chances to update `work status` than before. But it seems quite a potential risk here. Maybe we could consider about a better way later to reduce the attempts to do unnecessary updating.

```release-note

```

### DIFF
--- a/pkg/controllers/execution/execution_controller.go
+++ b/pkg/controllers/execution/execution_controller.go
@@ -3,7 +3,6 @@ package execution
 import (
 	"context"
 	"fmt"
-	"reflect"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


RainbowMango:
> I'm sorry for bringing this in https://github.com/karmada-io/karmada/pull/3808 , I intended to reduce the attempts to update work because now there are much more chances to update work status than before. But it seems quite a potential risk here. Maybe we could consider about a better way later to reduce the attempts to do unnecessary updating.

I can't tell how bad it is right now. Please @XiShanYongYe-Chang take a look.

@lxtywypc Should we consider revert #3808?

XiShanYongYe-Chang:
Ok, I will take a look ASAP.

XiShanYongYe-Chang:
Hi @lxtywypc, thank you for your update. I indeed did not consider this point when reviewing it earlier. How did you discover it during testing?

XiShanYongYe-Chang (commented):
> I can't tell how bad it is right now. Please @XiShanYongYe-Chang take a look.
> 
> @lxtywypc Should we consider revert #3808?

We don't need to revert the previously merged pull request. The purpose of this modification was originally to reduce the number of work updates, but due to possible delayed cache updates, it could result in an incorrect work status. After removing this logic, the work status can be correctly updated, although there may be some unnecessary update operations.

/lgtm

Ask an again review from @chaunceyjiang 
/cc @chaunceyjiang

lxtywypc:
> How did you discover it during testing?  

Due to that our testing environment sometimes need a little long time(for example 1s) for cache syncing. Once we noticed a work failed applied for a long time and tried to figure out what happened, we found that it had been applied successfully after once retry through the log of `karmada-controller-manager`. Then we found the potential risk of this part.

lxtywypc:
Draft for reverting #3808 as comment https://github.com/karmada-io/karmada/issues/3959#issuecomment-1700339009

RainbowMango:
/close
I guess we don't need it anymore.

PR #3995, opened 2023-08-24
feat: deprecate cluster-zone flag and introduce cluster-zones flag

whitewindmills:
**What type of PR is this?**
/kind feature
/kind cleanup

deprecate cluster-zone flag and introduce cluster-zones flag

part of https://github.com/karmada-io/karmada/issues/3952

```release-note
`karmadactl`:  Deprecated `--cluster-zone` flag which will be removed in future releases and introduced `--cluster-zones` flag.
```

### DIFF
--- a/cmd/agent/app/agent.go
+++ b/cmd/agent/app/agent.go
@@ -405,8 +405,8 @@ func generateClusterInControllerPlane(opts util.ClusterRegisterOption) (*cluster
 			cluster.Spec.Provider = opts.ClusterProvider
 		}
 
-		if opts.ClusterZone != "" {
-			cluster.Spec.Zone = opts.ClusterZone
+		if len(opts.ClusterZones) > 0 {
+			cluster.Spec.Zones = opts.ClusterZones
 		}
 
 		if opts.ClusterRegion != "" {
--- a/pkg/karmadactl/join/join.go
+++ b/pkg/karmadactl/join/join.go
@@ -85,7 +85,15 @@ type CommandJoinOption struct {
 	ClusterRegion string
 
 	// ClusterZone represents the zone of the cluster locate in.
+	// Deprecated: Use ClusterZones instead.
 	ClusterZone string
+
+	// ClusterZones represents the failure zones(also called availability zones) of the
+	// member cluster. The zones are presented as a slice to support the case
+	// that cluster runs across multiple failure zones.
+	// Refer https://kubernetes.io/docs/setup/best-practices/multiple-zones/ for
+	// more details about running Kubernetes in multiple zones.
+	ClusterZones []string
 
 	// DryRun tells if run the command in dry-run mode, without making any server requests.
 	DryRun bool
@@ -116,6 +124,10 @@ func (j *CommandJoinOption) Validate(args []string) error {
 	if j.ClusterNamespace == util.NamespaceKarmadaSystem {
 		klog.Warningf("karmada-system is always reserved for Karmada control plane. We do not recommend using karmada-system to store secrets of member clusters. It may cause mistaken cleanup of resources.")
 	}
+
+	if j.ClusterZone != "" && len(j.ClusterZones) > 0 {
+		return fmt.Errorf("--cluster-zone and --cluster-zones can not co-exist")
+	}
 	return nil
 }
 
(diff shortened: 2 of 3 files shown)


RainbowMango (commented):
/assign

cmd/agent/app/agent.go
```diff
@@ -405,8 +405,8 @@ func generateClusterInControllerPlane(opts util.ClusterRegisterOption) (*cluster
 			cluster.Spec.Provider = opts.ClusterProvider
 		}
 
-		if opts.ClusterZone != "" {
-			cluster.Spec.Zone = opts.ClusterZone
+		if len(opts.ClusterZones) > 0 {
+			cluster.Spec.Zones = opts.ClusterZones
```
RainbowMango, line 409: Seems we can't just ignore the `--zone` flag, otherwise it acts like the flags have been removed.
We should migrate the flag value to `--zones`, and make sure `--zone` and `--zones` can't co-exist.

pkg/karmadactl/join/join.go
```diff
@@ -128,7 +136,10 @@ func (j *CommandJoinOption) AddFlags(flags *pflag.FlagSet) {
 		"Path of the cluster's kubeconfig.")
 	flags.StringVar(&j.ClusterProvider, "cluster-provider", "", "Provider of the joining cluster. The Karmada scheduler can use this information to spread workloads across providers for higher availability.")
 	flags.StringVar(&j.ClusterRegion, "cluster-region", "", "The region of the joining cluster. The Karmada scheduler can use this information to spread workloads across regions for higher availability.")
-	flags.StringVar(&j.ClusterZone, "cluster-zone", "", "The zone of the joining cluster")
+	flags.StringVar(&j.ClusterZone, "cluster-zone", "", "The zone of the joining cluster.")
+	// nolint: errcheck
```
RainbowMango, line 140: What kind of error here?

cmd/agent/app/agent.go
whitewindmills: it's okay because karmada-agent has not such flag `--zone`. it should be completed in pr https://github.com/karmada-io/karmada/pull/3844.

RainbowMango: Thanks for the reminder. That's a dead code.

pkg/karmadactl/join/join.go
whitewindmills: error that never occurs.
```go
// MarkDeprecated indicated that a flag is deprecated in your program. It will
// continue to function but will not show up in help or usage messages. Using
// this flag will also print the given usageMessage.
func (f *FlagSet) MarkDeprecated(name string, usageMessage string) error {
	flag := f.Lookup(name)
	if flag == nil {
		return fmt.Errorf("flag %q does not exist", name)
	}
	if usageMessage == "" {
		return fmt.Errorf("deprecated message for flag %q must be set", name)
	}
	flag.Deprecated = usageMessage
	flag.Hidden = true
	return nil
}
```

pkg/karmadactl/join/join.go
```diff
@@ -168,6 +179,7 @@ func (j *CommandJoinOption) RunJoinCluster(controlPlaneRestConfig, clusterConfig
 		ClusterProvider:    j.ClusterProvider,
 		ClusterRegion:      j.ClusterRegion,
 		ClusterZone:        j.ClusterZone,
+		ClusterZones:       j.ClusterZones,
```
RainbowMango, line 182: Here we don't migrate the `zone` to `zones`, and expect the cluster validation would deny the case that they co-exist. Am I right?

whitewindmills: yes, but I think it would be a little better to check ahead of time, I will improve it.

pkg/karmadactl/join/join.go
RainbowMango: Got it. Looks good to me.

```
golangci-lint has version 1.52.2 built with go1.20.2 from da04413a on 2023-03-25T18:11:28Z
pkg/karmadactl/join/join.go:140:22: Error return value of `flags.MarkDeprecated` is not checked (errcheck)
	flags.MarkDeprecated("cluster-zone", "This flag is deprecated and will be removed in future releases. Use --cluster-zones instead.")

```

pkg/karmadactl/join/join.go
RainbowMango: Yes. since the joining process is probably not atomic.
I think just a validation would be fine.

RainbowMango (approved):
/lgtm
/approve

Issue #3997, opened 2023-08-24
Note: Code Freeze by the end of this month


RainbowMango:
We are going to cut a new release(v1.7.0) by the end of this month(8.31).

Only limited PRs are accepted into the release codebase(`master` branch) during this time:
- PR which fixes critical bugs.
- PR which `very` `very` important for the release.
- PR which updates/adds documents

*Please let me know the PRs which you really want to be included in this release.*

Feature development and review are still ongoing during this period, thanks in advance for your cooperation.

RainbowMango:
Karmada v1.7.0 is released. Please refer to the [release page](https://github.com/karmada-io/karmada/releases/tag/v1.7.0) for more information.

/close

Issue #3866, opened 2023-07-31
Is it  appropriate to change `--skipped-propagating-apis` into `--managed-propagating-apis`?


snowplayfire:
**What would you like to be added**:

**Why is this needed**:
It's more clear to know how many resources are managed by karmada, which is helpful to be aware of how much it has endured.

snowplayfire:
@RainbowMango @XiShanYongYe-Chang  please have a look.

RainbowMango:
`--skipped-propagating-apis` is a kind of `ignore list`, but the `--managed-propagating-apis` is a kind of `allowed list`, 
Do you mean people have to edit the configuration every time they add a new CRD?

snowplayfire:
How about Hot reload `allowed list`?  The implementation method is open to further discussion.

RainbowMango:
> which is helpful to be aware of how much it has endured

I'd like to hear more details about this requirement. We might have other solutions.

RainbowMango:
How about talking about it at tomorrow's community meeting?

snowplayfire:
> How about talking about it at tomorrow's community meeting?

ok

RainbowMango:
I just added [an agenda](https://docs.google.com/document/d/1y6YLVC-v7cmVAdbjedoyR5WL0-q45DBRXTvz5_I7bkA/edit?pli=1#heading=h.bdr7uobkel4t) for you.

chaosi-zju:
It seems like there are four ways to achieve it:

1. add `--skipped-propagating-apis` as params of `karmada-controller-manager`: API is not intuitive to users、lack of management、lack of visualization
2. managed as a `allowed list` by RBAC: not intuitive to users too, somewhat tricky
3. add a new API: the workload involved may be a bit larger
4. managed as a `allowed list` in ConfigMap: similar to "add a new API" in basic principle, but not as clear as it is

So, I prefer to add a new API, which names like `PropagateAPIList`, then we can define the propagating-apis managed/skipped as a allow/ignore list in `spec` field, and we can observe how many resources are managed or how much it has endured in `status` field by `kubectl`. I favour it for its intuitive and visualization.

majoyz:
Is anything updated for the issue? It seems the point is how to show objects that the karmada controlled. Can we just create a default propagationpolicy and make it show all resources the karmada controlled? The pp needn't to be in effect and it doesn't need to add new API. 
By the way, it seems detector won't process resource objects without matched propagationpolicy, is it right?

[...]

PR #4011, opened 2023-08-29
EventFilter add filter for ConfigMap Endpoints

yanfeng1992:
**What type of PR is this?**
/kind cleanup

Some components on the old k8s version still use configmap and endpoints for leader election, I think these resources should be ignored

Fixes #

```release-note
`karmada-controller-manager`:  EventFilter add filter for ConfigMap Endpoints
```

### DIFF
--- a/pkg/detector/detector.go
+++ b/pkg/detector/detector.go
@@ -20,6 +20,7 @@ import (
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/leaderelection/resourcelock"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
@@ -274,6 +275,15 @@ func (d *ResourceDetector) EventFilter(obj interface{}) bool {
 		case corev1.SchemeGroupVersion.WithKind("Secret"):
 			secretType, found, _ := unstructured.NestedString(unstructObj.Object, "type")
 			if found && secretType == string(corev1.SecretTypeServiceAccountToken) {
+				return false
+			}
+		// The configmap and endpoints for leader election should be prevented from propagating.
+		case corev1.SchemeGroupVersion.WithKind("ConfigMap"):
+			if _, ok := unstructObj.GetAnnotations()[resourcelock.LeaderElectionRecordAnnotationKey]; ok {
+				return false
+			}
+		case corev1.SchemeGroupVersion.WithKind("Endpoints"):
+			if _, ok := unstructObj.GetAnnotations()[resourcelock.LeaderElectionRecordAnnotationKey]; ok {
 				return false
 			}
 		}


XiShanYongYe-Chang:
Hi @yanfeng1992 thanks a lot~
We will be releasing a new version (v1.7.0) by the end of this month (8.31), ref to #3997 
Let's continue this PR when the next version starts.

pkg/detector/detector.go
```diff
@@ -276,6 +277,14 @@ func (d *ResourceDetector) EventFilter(obj interface{}) bool {
 			if found && secretType == string(corev1.SecretTypeServiceAccountToken) {
 				return false
 			}
```
chaunceyjiang, line 279: Can you add some comments to here?

yanfeng1992: // The configmap and endpoints for leader election should be prevented from propagating. 

@chaunceyjiang

chaunceyjiang (commented):
LGTM

pkg/detector/detector.go
```diff
@@ -271,11 +272,20 @@ func (d *ResourceDetector) EventFilter(obj interface{}) bool {
...
 			secretType, found, _ := unstructured.NestedString(unstructObj.Object, "type")
 			if found && secretType == string(corev1.SecretTypeServiceAccountToken) {
 				return false
 			}
+		case corev1.SchemeGroupVersion.WithKind("ConfigMap"):
+			if _, ok := unstructObj.GetAnnotations()[resourcelock.LeaderElectionRecordAnnotationKey]; ok {
+				return false
+			}
+		case corev1.SchemeGroupVersion.WithKind("Endpoints"):
+			if _, ok := unstructObj.GetAnnotations()[resourcelock.LeaderElectionRecordAnnotationKey]; ok {
```
jwcesign, line 286: How about `Lease` resource?

jwcesign (commented):
I find there are many other builtin resources, for example:
```
-> # ka get secret -A
NAMESPACE         NAME                   TYPE     DATA   AGE
karmada-cluster   member1                Opaque   2      27h
karmada-cluster   member1-impersonator   Opaque   1      27h
karmada-cluster   member2                Opaque   2      27h
karmada-cluster   member2-impersonator   Opaque   1      27h
karmada-cluster   member3                Opaque   2      27h
karmada-cluster   member3-impersonator   Opaque   1      27h
```

But there isn't a good way to do it.
/cc @XiShanYongYe-Chang @whitewindmills @chaunceyjiang 

What do you think?

pkg/detector/detector.go
yanfeng1992: https://github.com/karmada-io/karmada/blob/7c96e0db541989bd95772bbc1cc0109d8479371c/pkg/util/apigroup.go#L54C3-L54C3

NewSkippedResourceConfig  disable Lease by default

jwcesign: Thanks for reminding me.

whitewindmills:
@jwcesign 
it's okay, they will be ignored.
https://github.com/karmada-io/karmada/blob/7c96e0db541989bd95772bbc1cc0109d8479371c/pkg/detector/detector.go#L236-L246

whitewindmills (commented):
good work. 👍 
/lgtm

pkg/detector/detector.go
```diff
@@ -271,11 +272,20 @@ func (d *ResourceDetector) EventFilter(obj interface{}) bool {
 		// The secret, with type 'kubernetes.io/service-account-token', is created along with `ServiceAccount` should be
 		// prevented from propagating.
 		// Refer to https://github.com/karmada-io/karmada/pull/1525#issuecomment-1091030659 for more details.
+		// The configmap and endpoints for leader election should be prevented from propagating.
```
XiShanYongYe-Chang, line 275: Shouldn't this comment be placed below? Other LGTM

yanfeng1992: ok，thanks for reminding

RainbowMango (commented):
/assign

XiShanYongYe-Chang:
Thanks
/lgtm

RainbowMango:
> Some components on the old k8s version still use configmap and endpoints for leader election,

Do you know from which version Kubernetes stopped using ConfigMap and Endpoints and started using Lease?

My concern is if the version prior to 1.16, we might not need to ignore them as Karmada apiserver requires kube-apiserver >= 1.16 (Karmada's CRD is generated with CRD/v1 (Kubernetes 1.16)).

yanfeng1992:
leaderelection in client-go supports 5 types of resource locks:

const (
    LeaderElectionRecordAnnotationKey = "[control-plane.alpha.kubernetes.io/leader](http://control-plane.alpha.kubernetes.io/leader)"
    EndpointsResourceLock             = "endpoints"
    ConfigMapsResourceLock            = "configmaps"
    LeasesResourceLock                = "leases"
    EndpointsLeasesResourceLock       = "endpointsleases"
    ConfigMapsLeasesResourceLock      = "configmapsleases"
)

In version 1.24, the community completely removed support for Endpoints/ConfigMaps as the primary resource lock.

https://github.com/kubernetes/kubernetes/pull/106852

https://github.com/kubernetes/kubernetes/issues/80289

These two types(endpointsleases, configmapsleases) are in a transition period, and it is necessary to obtain both Endpoints/Configmap and Lease resource locks (MultiLock) at the same time. The loss of any resource lock will cause the component to restart. When updating the lock, update Endpoints/Configmap first, and then update Lease. When judging loss, if both resources have holders but are inconsistent, an exception will be returned and the owner will be re-elected.

These two types of MultiLock still exist in our actual environment(v1.25.8). Of course, the overall trend in the community is to use the Leases type instead of the original Endpoints/Configmaps as resource locks. But we need to consider supporting MultiLock in transition phase

image

image

@RainbowMango

RainbowMango:
> These two types of MultiLock still exist in our actual environment(v1.25.8). 

If I remember right, all Karmada components now use `Lease` as locks, so, my question is why the other lock resources would exist in your environment?

Here is the lease I just get from `karmada-apiserver`:
```
# kubectl get leases.coordination.k8s.io -A 
NAMESPACE         NAME                         HOLDER                                                                                 AGE
karmada-cluster   member3                      member3                                                                                20h
karmada-system    karmada-agent-member3        karmada-agent-76cbbbc75-6djbm_c52b1d77-e152-4e56-8669-f62786a96116                     20h
karmada-system    karmada-controller-manager   karmada-controller-manager-5d8cc94745-67j85_88e0201c-a9f7-4834-bee6-5fab52e3e633       20h
karmada-system    karmada-descheduler          karmada-descheduler-7d654cbff6-sztzd_cfae23f7-ea84-429c-b81e-b4e6ac719d2c              20h
karmada-system    karmada-scheduler            karmada-scheduler-58b454779b-gj6gs_1e2a71e9-0e32-41f7-b73e-cbe183f990d2                20h
kube-system       kube-controller-manager      karmada-kube-controller-manager-66866b6d4-t6hgd_85a868e2-7eac-4a82-b954-1c3433947113   20h

```

yanfeng1992:
In our multi-cluster model, there are some other business team controllers that need resources in watch karmada-apiserver

image

RainbowMango:
> In our multi-cluster model, there are some other business team controllers that need resources to watch karmada-apiserver

Now I get it.  The business controllers use the `configmaps` or `endpoints` as the lock. 

One possible approach is just like this PR, let Karmada ignore these potential resources, but before that, I'm thinking a question: what's the problem here if Karmada does not do it?
If I understand correctly, without the event filter, these resources will be put on the `waiting list` and be there forever. No functionality issues, but they will be processed (non-opt) periodicity because the business controller would update them.

Another possible approach is to let the business controller start to use `Lease`, is it possible for you?

In addition, this case is quite similar to #3866, which requests to define an allow-list to control what kind of resource should be watched by the detector.

yanfeng1992:
> Another possible approach is to let the business controller start to use `Lease`, is it possible for you?

yes i agree with that，use `Lease` is a better way

>  No functionality issues, but they will be processed (non-opt) periodicity because the business controller would update them.

If the update frequency is relatively high, can it also be considered a problem? What I see now is that it is updated every two seconds, which is also related to the configuration of the main election

RainbowMango:
> If the update frequency is relatively high, can it also be considered a problem? What I see now is that it is updated every two seconds, which is also related to the configuration of the main election

I don't think you need to renew the lock too often, for Karmada, take `karmada-controller-manager` as an example,  the renew duration is `10s`.
```
      --leader-elect-renew-deadline duration                                                                                                                                                        
                The interval between attempts by the acting master to renew a leadership slot before it stops leading. This must be less than or equal to the lease duration. This is only applicable
                if leader election is enabled. (default 10s)
```

In addition to the `configmaps` and `endpoints`, there must be some other resources we need to filter, what I'm thinking is how to provide an extensible or configurable mechanism to handle these things instead of hard-code them one by one.

yanfeng1992:
> In addition to the `configmaps` and `endpoints`, there must be some other resources we need to filter, what I'm thinking is how to provide an extensible or configurable mechanism to handle these things instead of hard-code them one by one.

Okay, I agree with this too

PR #4027, opened 2023-09-05
fix depndencies lua scrpit bug

XiShanYongYe-Chang:
**What type of PR is this?**

/kind cleanup

When a field of an object is empty, retrieving its value will yield nil:

image

An error will occur when using this nil value to index a table.

image

Therefore, we need to correct these mistakes in writing

Fixes #

```release-note
NONE
```


RainbowMango (approved):
Echo from [Lua.org/manual/5.1](https://www.lua.org/manual/5.1/manual.html#2.2):
> Nil is the type of the value nil, whose main property is to be different from any other value; `it usually represents the absence of a useful value`.

/lgtm
/approve

PR #4052, opened 2023-09-12
Don't send an event when the object not matched any propagation policy

### LINKED ISSUE #4038: A confusing event "No policy match for resource"
 The karmada controller `ResourceDetector` watch all resource and try to find a policy that matches this resource. If resources do not match any policy then it sends a "ApplyPolicyFailed" and "No policy match for resource" event for this resource.
That event made me confused, Is there any problem with this resource?

https://github.com/karmada-io/karmada/blob/5a583c2c9f53ad726a941c620e1031b4b3f8af9c/pkg/detector/policy.go#L77-L82

If no policy matches this resource there is no need to send an event.

**Environment**:
- Karmada version: 1.6.2
- Kubernetes version: v1.23
- Others:

wu0407:
**What type of PR is this?**

/kind flake

remove the `ApplyPolicyFailed` event for objects not match any policy and add a log for debugging purposes.

Fixes #4038

```release-note
`karmada-controller-manager`: Skip emitting the `ApplyPolicyFailed` event for the resource template which does not match any PropagationPolicy.
```

### DIFF
--- a/pkg/detector/policy.go
+++ b/pkg/detector/policy.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"time"
 
-	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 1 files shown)


RainbowMango (commented):
/assign

RainbowMango (approved):
/lgtm
/approve

PR #4064, opened 2023-09-14
aggregate the HPA status from member clusters into control plane

chaosi-zju:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

HPA status aggregation. Aggregate the HPA status from different clusters onto the control plane HPA resource template using a resource interpreter, with the involved state field being .status.currentReplicas.

**Which issue(s) this PR fixes**:

part of #4057

**Special notes for your reviewer**:

You can test it by：

```bash
kind delete clusters --all; rm -r ~/.karmada/; rm ~/.kube/*config; rm -r /etc/karmada
hack/local-up-karmada.sh
export KUBECONFIG=~/.kube/karmada.config:~/.kube/members.config

cat >/tmp/hpa.yaml<<EOF
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx
  labels:
    app: nginx
    horizontalpodautoscaler.karmada.io/name: nginx
spec:
  replicas: 1
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - image: nginx
        name: nginx
        resources:
          requests:
            cpu: 25m
            memory: 64Mi
          limits:
            cpu: 25m
            memory: 64Mi
---
apiVersion: v1
kind: Service
metadata:
  name: nginx-service
spec:
  ports:
  - port: 80
    targetPort: 80
    nodePort: 30080
  selector:
    app: nginx
  type: NodePort
---
apiVersion: autoscaling/v2
kind: HorizontalPodAutoscaler
metadata:
  name: nginx
spec:
  scaleTargetRef:
    apiVersion: apps/v1
    kind: Deployment
    name: nginx
  minReplicas: 1
  maxReplicas: 10
  behavior:
    scaleDown:
      stabilizationWindowSeconds: 10
    scaleUp:
      stabilizationWindowSeconds: 10
  metrics:
    - type: Resource
      resource:
        name: cpu
        target:
          type: Utilization
          averageUtilization: 10
---
apiVersion: policy.karmada.io/v1alpha1
kind: PropagationPolicy
metadata:
  name: nginx-propagation
spec:
  resourceSelectors:
    - apiVersion: apps/v1
      kind: Deployment
      name: nginx
    - apiVersion: v1
      kind: Service
      name: nginx-service
    - apiVersion: autoscaling/v2
      kind: HorizontalPodAutoscaler
      name: nginx
  placement:
    clusterAffinity:
      clusterNames:
[...]

Issue #4058, opened 2023-09-13
[umbrella] Auto detect whether workload needs to retain replicas in the member cluster through Retain operation


XiShanYongYe-Chang:
**What would you like to be added**:

Add the `Retain` operation for workload resources, when there are relevant HPAs in effect, and keep the number of replicas in the member cluster.

**Why is this needed**:

When users propagate workload resources (such as Deployment) and HPA to member clusters through Kamada, the HPA controller in the member cluster will work and dynamically adjust the scale subresource of the workload. At the same time, Karmada control plane will also attempt to modify the replicas on the workload in the member cluster. These two operations will conflict. We need to retain the modification of replicas on workload by the HPA controller in the member cluster.

**It can be divided into the following tasks.**

- [x] Add a specific label to the workload resource to indicate its association with working HPAs. #4078
- [x] Add retain operation in native resourceinterpreter for Deployment resource. #4078
- [x] Set the retain operation more common and not limited to the deployment resource. 
       Ref task discussion: https://github.com/karmada-io/karmada/pull/4078#discussion_r1332455732
- [x] Add E2E to cover this case. #4764

chaunceyjiang:
Hi @XiShanYongYe-Chang, has anyone picked up this issue yet? If not, i can pick up this one.

XiShanYongYe-Chang:
> Hi @XiShanYongYe-Chang, has anyone picked up this issue yet? If not, i can pick up this one.

Thanks a lot~, I will assign it to you.
/assign @chaunceyjiang

chaosi-zju:
@chaunceyjiang hello, have you start your task? I am doing another task (#4057), however, I cannot test my code without the feature `Add retain operation in native resourceinterpreter for Deployment resource`, then my code may be imcomplete to merge, so, I wanna if you didn't start this subtask, may I do this subtask?

RainbowMango:
ping @XiShanYongYe-Chang to sync the status.

XiShanYongYe-Chang:
All sub-tasks have been finished. Let me close it.
/close

PR #4075, opened 2023-09-18
retain for hpa controlled Deployment resource.

chaosi-zju:
**What type of PR is this?**

/kind feature

Add the Retain operation for Deployment resources, when there are relevant HPAs in effect, keep the number of replicas in the member cluster.

part of #4058

Test Method：same as #4064

Test Report：

image
image

```release-note

```

### DIFF
--- a/pkg/resourceinterpreter/default/native/retain.go
+++ b/pkg/resourceinterpreter/default/native/retain.go
@@ -1,12 +1,16 @@
 package native
 
 import (
+	"context"
 	"fmt"
 
+	appsv1 "k8s.io/api/apps/v1"
+	autoscalingv2 "k8s.io/api/autoscaling/v2"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/karmada-io/karmada/pkg/util"
 	"github.com/karmada-io/karmada/pkg/util/helper"
@@ -16,8 +20,9 @@ import (
 // retentionInterpreter is the function that retains values from "observed" object.
 type retentionInterpreter func(desired *unstructured.Unstructured, observed *unstructured.Unstructured) (retained *unstructured.Unstructured, err error)
 
-func getAllDefaultRetentionInterpreter() map[schema.GroupVersionKind]retentionInterpreter {
+func getAllDefaultRetentionInterpreter(client client.Client) map[schema.GroupVersionKind]retentionInterpreter {
 	s := make(map[schema.GroupVersionKind]retentionInterpreter)
+	s[appsv1.SchemeGroupVersion.WithKind(util.DeploymentKind)] = retainDeploymentFields(client)
 	s[corev1.SchemeGroupVersion.WithKind(util.PodKind)] = retainPodFields
 	s[corev1.SchemeGroupVersion.WithKind(util.ServiceKind)] = lifted.RetainServiceFields
 	s[corev1.SchemeGroupVersion.WithKind(util.ServiceAccountKind)] = lifted.RetainServiceAccountFields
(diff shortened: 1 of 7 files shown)


pkg/resourceinterpreter/default/native/retain.go
```diff
@@ -122,3 +127,36 @@ func retainJobSelectorFields(desired, observed *unstructured.Unstructured) (*uns
 	}
 	return desired, nil
 }
+
+const IndexKeyHPAScaleTargetRef = "spec.scaleTargetRef"
+
+func retainDeploymentFields(cli client.Client) retentionInterpreter {
```
chaunceyjiang, line 133: I have another idea. We should write a `controller` to watch the creation of HPA. Once there is a change in HPA, I will add a `label` to the target resource pointed by HPA. 


Then we only need to check if the resource has this specific label here to determine whether we need to perform the `retain` operation.

pkg/resourceinterpreter/default/native/retain.go
```diff
@@ -122,3 +127,36 @@ func retainJobSelectorFields(desired, observed *unstructured.Unstructured) (*uns
...
+
+	return func(desired, observed *unstructured.Unstructured) (*unstructured.Unstructured, error) {
+		hpaList := &autoscalingv2.HorizontalPodAutoscalerList{}
+		deployIdx := autoscalingv2.CrossVersionObjectReference{Kind: desired.GetKind(), Name: desired.GetName(), APIVersion: desired.GetAPIVersion()}
+
+		if err := cli.List(context.Background(), hpaList, client.InNamespace(desired.GetNamespace()),
```
chaunceyjiang, line 139: Have a question:

If we have another custom resources, How to implementation such similar logic?

pkg/resourceinterpreter/default/native/retain.go
chaosi-zju: yes, we have considered this method, it may not better than this method

pkg/resourceinterpreter/default/native/retain.go
chaosi-zju: write a similar retain function？(Is custom resources somewhere more specifial in retain operator?)

besides, if you use `labels` method, where would you resolve custom resources retain operatoin? I think there's may be no major difference.

chaunceyjiang: >  if you use labels method, where would you resolve custom resources retain operatoin?

we can write a `lua` scripts, and we only use label


We can determine whether to execute the "retain" operation by writing a `Lua` script that `only focuses on` whether specific label exist in the resource's label.

If use your current approach, there is no `k8s client` in Lua.

chaosi-zju: `k8s client` is not required.

> execute the "retain" operation by writing a Lua script

can this method use custom resources informer? As long as we can list the resources, we can implement similar logic.

pkg/resourceinterpreter/default/native/retain.go
chaosi-zju: labels method may unable to resolve all scenes, follow my steps, e.g:

create `HPA` in karmada -> propagate `HPA` to `member1` -> [`hpa ctrl` find `deployement` failed] -> create `deployment` in karmada -> propagate `deployment` to member1 -> `HPA` in member1 updated (currentReplicas set to 1) -> `HPA` in karmada updated (currentReplicas set to 1) -> [`hpa ctrl` called, add labels to `deployment`] -> manually delete the labels in `deployment` -> add stress to `deployment` in member1 -> `HPA` in member1 add `deployment` replicas -> `deployment` in karmada control `deployment` in member1, replicas back to 1 -> the `currentReplicas` filed in `HPA` keep unchange -> [`hpa ctrl` not called, since `HPA` unchanged]

***

updated：it seems my e.g is wrong. Since we have collect `desiredReplicas` other than `currentReplicas`, when "add stress to `deployment` in member1", though `currentReplicas` may unchanged, but `desiredReplicas` would change.

pkg/resourceinterpreter/default/native/retain.go
chaunceyjiang: > can this method use custom resources informer?

To be honest, I don't know either.

pkg/resourceinterpreter/default/native/retain.go
chaunceyjiang: > though currentReplicas may unchanged, but desiredReplicas would change.


Yes.

chaunceyjiang: > manually delete the labels in deploymen

Even if you manually delete the labels in deployment, it can be added again by `hpa ctrl`.

pkg/resourceinterpreter/default/native/retain.go
chaosi-zju: at this rate, labels method is better

XiShanYongYe-Chang: If we finally use the label method, can the current pr be closed first?

chaosi-zju: yes

PR #4072, opened 2023-09-18
add replicas syncer for resources with HPA

lxtywypc:
**What type of PR is this?**  
/kind feature  

**What this PR does / why we need it**:  
As https://github.com/karmada-io/karmada/issues/4057

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #4078, opened 2023-09-19
retain for hpa controlled Deployment resource (labels method)

chaosi-zju:
**What type of PR is this?**

/kind feature

Write a controller to watch the creation of HPA. Once there is a change in HPA, It will add a label to the target resource pointed by HPA. Then we only need to check if the resource has this specific label here to determine whether we need to perform the retain operation.

part of #4058

~another mthod：#4075~

we should choose just one between two.

Test Method：same as #4064

Test Report：

image
image
image
image

```release-note
Introduced retain operation to deployments if it is controlled by a HPA and the HPA has been propagated, which means member cluster deployments replicas will be controlled by member cluster HPA only.
```

### DIFF
--- a/pkg/controllers/hpareplicassyncer/hpa_replicas_syncer_controller.go
+++ b/pkg/controllers/hpareplicassyncer/hpa_replicas_syncer_controller.go
@@ -9,48 +9,45 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/scale"
 	"k8s.io/klog/v2"
 	controllerruntime "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/event"
-	"sigs.k8s.io/controller-runtime/pkg/predicate"
+
+	"github.com/karmada-io/karmada/pkg/util"
 )
 
-var hpaPredicate = predicate.Funcs{
-	CreateFunc: func(e event.CreateEvent) bool {
-		return false
-	},
-	UpdateFunc: func(e event.UpdateEvent) bool {
-		oldHPA, ok := e.ObjectOld.(*autoscalingv2.HorizontalPodAutoscaler)
-		if !ok {
-			return false
-		}
-
-		newHPA, ok := e.ObjectNew.(*autoscalingv2.HorizontalPodAutoscaler)
-		if !ok {
-			return false
-		}
-
-		return oldHPA.Status.CurrentReplicas != newHPA.Status.CurrentReplicas
-	},
-	DeleteFunc: func(e event.DeleteEvent) bool {
-		return false
-	},
-}
+const (
+	// ControllerName is the controller name that will be used when reporting events.
+	ControllerName = "hpa-replicas-syncer"
+	// scaleRefWorkerNum is the async Worker number
+	scaleRefWorkerNum = 1
+)
 
 // HPAReplicasSyncer is to sync replicas from status of HPA to resource template.
 type HPAReplicasSyncer struct {
-	Client      client.Client
-	RESTMapper  meta.RESTMapper
-	ScaleClient scale.ScalesGetter
+	Client        client.Client
+	DynamicClient dynamic.Interface
+	RESTMapper    meta.RESTMapper
+
+	ScaleClient    scale.ScalesGetter
+	scaleRefWorker util.AsyncWorker
 }
 
 // SetupWithManager creates a controller and register to controller manager.
 func (r *HPAReplicasSyncer) SetupWithManager(mgr controllerruntime.Manager) error {
-	return controllerruntime.NewControllerManagedBy(mgr).Named("replicas-syncer").
-		For(&autoscalingv2.HorizontalPodAutoscaler{}, builder.WithPredicates(hpaPredicate)).
+	scaleRefWorkerOptions := util.Options{
+		Name:          "scale ref worker",
+		ReconcileFunc: r.reconcileScaleRef,
+	}
+	r.scaleRefWorker = util.NewAsyncWorker(scaleRefWorkerOptions)
+	r.scaleRefWorker.Run(scaleRefWorkerNum, context.Background().Done())
+
+	return controllerruntime.NewControllerManagedBy(mgr).
+		Named(ControllerName).
+		For(&autoscalingv2.HorizontalPodAutoscaler{}, builder.WithPredicates(r)).
 		Complete(r)
 }
 
--- a/pkg/controllers/hpareplicassyncer/hpa_replicas_syncer_predicate.go
+++ b/pkg/controllers/hpareplicassyncer/hpa_replicas_syncer_predicate.go
@@ -1,0 +1,77 @@
+package hpareplicassyncer
+
+import (
+	autoscalingv2 "k8s.io/api/autoscaling/v2"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+
+	policyv1alpha1 "github.com/karmada-io/karmada/pkg/apis/policy/v1alpha1"
+)
+
+var _ predicate.Predicate = &HPAReplicasSyncer{}
+
+func (r *HPAReplicasSyncer) Create(e event.CreateEvent) bool {
+	hpa, ok := e.Object.(*autoscalingv2.HorizontalPodAutoscaler)
+	if !ok {
+		klog.Errorf("create predicates in hpa controller called, but obj is not hpa type")
+		return false
+	}
+
+	// if hpa exist and has been propagated, add label to its scale ref resource
+	if hasBeenPropagated(hpa) {
+		r.scaleRefWorker.Add(labelEvent{addLabelEvent, hpa})
+	}
+
+	return false
+}
+
+func (r *HPAReplicasSyncer) Update(e event.UpdateEvent) bool {
+	oldHPA, ok := e.ObjectOld.(*autoscalingv2.HorizontalPodAutoscaler)
+	if !ok {
+		klog.Errorf("update predicates in hpa controller called, but old obj is not hpa type")
+		return false
+	}
+
+	newHPA, ok := e.ObjectNew.(*autoscalingv2.HorizontalPodAutoscaler)
+	if !ok {
+		klog.Errorf("update predicates in hpa controller called, but new obj is not hpa type")
+		return false
+	}
+
+	// hpa scale ref changed, remove old hpa label and add to new hpa
+	if oldHPA.Spec.ScaleTargetRef.String() != newHPA.Spec.ScaleTargetRef.String() {
+		// if scale ref has label, remove label, otherwise skip
+		r.scaleRefWorker.Add(labelEvent{deleteLabelEvent, oldHPA})
+	}
+
+	// if new hpa exist and has been propagated, add label to its scale ref resource
+	if hasBeenPropagated(newHPA) {
+		r.scaleRefWorker.Add(labelEvent{addLabelEvent, newHPA})
+	}
+
+	return oldHPA.Status.CurrentReplicas != newHPA.Status.CurrentReplicas
+}
+
+func (r *HPAReplicasSyncer) Delete(e event.DeleteEvent) bool {
+	hpa, ok := e.Object.(*autoscalingv2.HorizontalPodAutoscaler)
+	if !ok {
+		klog.Errorf("delete predicates in hpa controller called, but obj is not hpa type")
+		return false
+	}
+
+	// if scale ref has label, remove label, otherwise skip
+	r.scaleRefWorker.Add(labelEvent{deleteLabelEvent, hpa})
+
+	return false
+}
+
+func (r *HPAReplicasSyncer) Generic(e event.GenericEvent) bool {
+	return false
+}
+
+func hasBeenPropagated(hpa *autoscalingv2.HorizontalPodAutoscaler) bool {
+	_, ppExist := hpa.GetLabels()[policyv1alpha1.PropagationPolicyUIDLabel]
+	_, cppExist := hpa.GetLabels()[policyv1alpha1.ClusterPropagationPolicyUIDLabel]
+	return ppExist || cppExist
+}
--- a/pkg/controllers/hpareplicassyncer/hpa_scale_ref_worker.go
+++ b/pkg/controllers/hpareplicassyncer/hpa_scale_ref_worker.go
@@ -1,0 +1,137 @@
+package hpareplicassyncer
+
+import (
+	"context"
+	"fmt"
+
+	autoscalingv2 "k8s.io/api/autoscaling/v2"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/klog/v2"
+
+	"github.com/karmada-io/karmada/pkg/util"
+	"github.com/karmada-io/karmada/pkg/util/helper"
+)
+
+type labelEventKind int
+
+const (
+	// addLabelEvent refer to addding util.RetainReplicasLabel to resource scaled by HPA
+	addLabelEvent labelEventKind = iota
+	// deleteLabelEvent refer to deleting util.RetainReplicasLabel from resource scaled by HPA
+	deleteLabelEvent
+)
+
+type labelEvent struct {
+	kind labelEventKind
+	hpa  *autoscalingv2.HorizontalPodAutoscaler
+}
+
+func (r *HPAReplicasSyncer) reconcileScaleRef(key util.QueueKey) (err error) {
+	event, ok := key.(labelEvent)
+	if !ok {
+		klog.Errorf("Found invalid key when reconciling hpa scale ref: %+v", key)
+		return nil
+	}
+
+	switch event.kind {
+	case addLabelEvent:
+		err = r.addHPALabelToScaleRef(context.TODO(), event.hpa)
+	case deleteLabelEvent:
+		err = r.deleteHPALabelFromScaleRef(context.TODO(), event.hpa)
+	default:
+		klog.Errorf("Found invalid key when reconciling hpa scale ref: %+v", key)
+		return nil
+	}
+
+	if err != nil {
+		klog.Errorf("reconcile scale ref failed: %+v", err)
+	}
+	return err
+}
+
+func (r *HPAReplicasSyncer) addHPALabelToScaleRef(ctx context.Context, hpa *autoscalingv2.HorizontalPodAutoscaler) error {
+	targetGVK := schema.FromAPIVersionAndKind(hpa.Spec.ScaleTargetRef.APIVersion, hpa.Spec.ScaleTargetRef.Kind)
+	mapping, err := r.RESTMapper.RESTMapping(targetGVK.GroupKind(), targetGVK.Version)
+	if err != nil {
+		return fmt.Errorf("unable to recognize scale ref resource, %s/%v, err: %+v", hpa.Namespace, hpa.Spec.ScaleTargetRef, err)
+	}
+
+	scaleRef, err := r.DynamicClient.Resource(mapping.Resource).Namespace(hpa.Namespace).Get(ctx, hpa.Spec.ScaleTargetRef.Name, metav1.GetOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			klog.Infof("scale ref resource is not found (%s/%v), skip processing", hpa.Namespace, hpa.Spec.ScaleTargetRef)
+			return nil
+		}
+		return fmt.Errorf("failed to find scale ref resource (%s/%v), err: %+v", hpa.Namespace, hpa.Spec.ScaleTargetRef, err)
+	}
+
+	// use patch is better than update, when modification occur after get, patch can still success while update can not
+	newScaleRef := scaleRef.DeepCopy()
+	util.MergeLabel(newScaleRef, util.RetainReplicasLabel, util.RetainReplicasValue)
+	patchBytes, err := helper.GenMergePatch(scaleRef, newScaleRef)
+	if err != nil {
+		return fmt.Errorf("failed to gen merge patch (%s/%v), err: %+v", hpa.Namespace, hpa.Spec.ScaleTargetRef, err)
+	}
+	if len(patchBytes) == 0 {
+		klog.Infof("hpa labels already exist, skip adding (%s/%v)", hpa.Namespace, hpa.Spec.ScaleTargetRef)
+		return nil
+	}
+
+	_, err = r.DynamicClient.Resource(mapping.Resource).Namespace(newScaleRef.GetNamespace()).
+		Patch(ctx, newScaleRef.GetName(), types.MergePatchType, patchBytes, metav1.PatchOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			klog.Infof("scale ref resource is not found (%s/%v), skip processing", hpa.Namespace, hpa.Spec.ScaleTargetRef)
+			return nil
+		}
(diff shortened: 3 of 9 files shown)


chaosi-zju:
@chaunceyjiang hello, for yesterday's hpa retain problem, I also achieve `label method` version, do you have time review this?

chaosi-zju:
@lxtywypc hello, my this PR may have some conflict with your #4072, do you have time helping review my this PR? we can talk about what is the best practice to merge our two PR~

chaosi-zju:
@XiShanYongYe-Chang it is ready to review

chaosi-zju:
I have refactored my code to adapt to #4072, avoiding potential conflict. @XiShanYongYe-Chang

pkg/resourceinterpreter/default/native/retain.go
```diff
@@ -16,8 +17,11 @@ import (
 // retentionInterpreter is the function that retains values from "observed" object.
 type retentionInterpreter func(desired *unstructured.Unstructured, observed *unstructured.Unstructured) (retained *unstructured.Unstructured, err error)
 
+const deploymentControlByHPALabel = "horizontalpodautoscaler.karmada.io/name"
```
chaunceyjiang, line 20: I prefer to move this const to  the `pkg/util/constants.go` file.

XiShanYongYe-Chang:
Hi @chaosi-zju, #4072 has been merged, can you help rebase it?

pkg/resourceinterpreter/default/native/retain.go
```diff
@@ -122,3 +126,23 @@ func retainJobSelectorFields(desired, observed *unstructured.Unstructured) (*uns
 	}
 	return desired, nil
 }
+
+func retainDeploymentFields(desired, observed *unstructured.Unstructured) (*unstructured.Unstructured, error) {
+	labels, exist, err := unstructured.NestedStringMap(desired.Object, "metadata", "labels")
```
chaunceyjiang, line 131: I think this is a general logic, not just for `Deployment`, so I prefer to set it as a default function, similar to `reflectWholeStatus`.

chaunceyjiang: So we don't need to write the same logic again for `pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/apps.kruise.io/v1alpha1/CloneSet`.

XiShanYongYe-Chang: Operations on the replicas of Kubernetes Workload resources can be universal. Therefore, operations on replicas can be extracted as an independent function, such as `retainK8sWorkloadReplicas`.

XiShanYongYe-Chang: > So we don't need to write the same logic again for `pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/apps.kruise.io/v1alpha1/CloneSet`.

Are you currently using HPA to control CloneSets?

chaosi-zju:
> Hi @chaosi-zju, #4072 has been merged, can you help rebase it?

done

pkg/resourceinterpreter/default/native/retain.go
chaunceyjiang: I haven't used it in prod. 

However, I know that in the `OpenKruise` community, there are already users who use HPA to control CloneSet.

lxtywypc:
> @lxtywypc hello, my this PR may have some conflict with your https://github.com/karmada-io/karmada/pull/4072, do you have time helping review my this PR? we can talk about what is the best practice to merge our two PR~  

Hi @chaosi-zju I'm sorry for noticing this late.  

Thanks for your refactoring, I will help review it now.

pkg/resourceinterpreter/default/native/retain.go
chaosi-zju: done

pkg/controllers/hpareplicassyncer/hpa_replicas_syncer_predicate.go
```diff
@@ -0,0 +1,79 @@
...
+import (
+	autoscalingv2 "k8s.io/api/autoscaling/v2"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+
+	policyv1alpha1 "github.com/karmada-io/karmada/pkg/apis/policy/v1alpha1"
+)
+
+const KindHPA = "HorizontalPodAutoscaler"
```
lxtywypc, line 12: We could use this to replace:  

https://github.com/karmada-io/karmada/blob/8d1a8007ebdbd97a2fe12cf8ee7d4694cfb2ed82/pkg/util/constants.go#L108

pkg/controllers/hpareplicassyncer/hpa_scale_ref_worker.go
```diff
@@ -0,0 +1,128 @@
...
+type labelEvent struct {
+	kind labelEventKind
+	hpa  *autoscalingv2.HorizontalPodAutoscaler
+}
+
+func (r *HPAReplicasSyncer) ReconcileScaleRef(key util.QueueKey) error {
+	event, ok := key.(labelEvent)
+	if !ok {
+		klog.Errorf("Found invalid key when reconciling hpa scale ref: %+v", key)
+		return fmt.Errorf("invalid key")
```
lxtywypc, line 36: If the key is invalid, it might be better not to return error, because there is no chances to change the key, and the key would be left in queue forever.

pkg/controllers/hpareplicassyncer/hpa_scale_ref_worker.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	switch event.kind {
+	case addLabelEvent:
+		return r.addHPALabelToScaleRef(event.hpa)
+	case deleteLabelEvent:
+		return r.deleteHPALabelFromScaleRef(event.hpa)
+	default:
+		return fmt.Errorf("invalid key")
```
lxtywypc, line 45: Same as above.

pkg/controllers/hpareplicassyncer/hpa_scale_ref_worker.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+}
+
+func (r *HPAReplicasSyncer) addHPALabelToScaleRef(hpa *autoscalingv2.HorizontalPodAutoscaler) error {
+	scaleRef, mapping, err := r.getScaleRef(hpa)
+	if err != nil {
+		return err
```
lxtywypc, line 52: Better to add error log here.

pkg/controllers/hpareplicassyncer/hpa_scale_ref_worker.go
chaosi-zju: done

pkg/controllers/hpareplicassyncer/hpa_scale_ref_worker.go
chaosi-zju: done

pkg/controllers/hpareplicassyncer/hpa_scale_ref_worker.go
```diff
@@ -0,0 +1,134 @@
...
+		klog.Infof("hpa labels not exist, skip deleting (%+v)", resource)
+		return nil
+	}
+
+	_, err = r.DynamicClient.Resource(resource).Namespace(newScaleRef.GetNamespace()).
+		Patch(context.Background(), newScaleRef.GetName(), types.MergePatchType, patchBytes, metav1.PatchOptions{})
+
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			klog.Infof("scale ref resource is not found (%+v), skip processing", resource)
```
chaunceyjiang, line 127:
```suggestion
			klog.V(4).Infof("scale ref resource is not found (%+v), skip processing", resource)
```

pkg/controllers/hpareplicassyncer/hpa_scale_ref_worker.go
```diff
@@ -0,0 +1,134 @@
...
+	if err != nil {
+		return nil, nil, fmt.Errorf("unable to recognize scale ref resource: %+v", err)
+	}
+
+	scaleRef, err := r.DynamicClient.Resource(mapping.Resource).Namespace(hpa.Namespace).
+		Get(context.Background(), hpa.Spec.ScaleTargetRef.Name, metav1.GetOptions{})
+
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			klog.Infof("scale ref resource is not found (%+v), skip processing", hpa.Spec.ScaleTargetRef)
```
chaunceyjiang, line 103:
```suggestion
			klog.V(4).Infof("scale ref resource is not found (%+v), skip processing", hpa.Spec.ScaleTargetRef)
```

pkg/resourceinterpreter/default/native/retain.go
```diff
@@ -18,6 +19,7 @@ type retentionInterpreter func(desired *unstructured.Unstructured, observed *uns
 
 func getAllDefaultRetentionInterpreter() map[schema.GroupVersionKind]retentionInterpreter {
 	s := make(map[schema.GroupVersionKind]retentionInterpreter)
+	s[appsv1.SchemeGroupVersion.WithKind(util.DeploymentKind)] = retainK8sWorkloadReplicas
```
chaunceyjiang, line 22: Generally  LGTM.

However, I still think that this is a general solution, not just for `Deployment`.

pkg/resourceinterpreter/default/native/retain.go
XiShanYongYe-Chang: If needed, we can submit a single pr to set the Retain operation for the `CloneSet` and other resources.

pkg/resourceinterpreter/default/native/retain.go
XiShanYongYe-Chang: I may not have understood what you meant before, but now I understand that you mean adding a default operation to process all resources and retain the replicas as long as it contain the target label. Am I right?

pkg/controllers/hpareplicassyncer/hpa_scale_ref_worker.go
chaosi-zju: done, by the way, in which scene should we add `.V(4)`?

pkg/util/constants.go
```diff
@@ -28,6 +28,9 @@ const (
 
 	// ManagedByKarmadaLabelValue indicates that resources are managed by karmada controllers.
 	ManagedByKarmadaLabelValue = "true"
+
+	// ResourceScaledByHPALabel is added to mark resource template scaled by HPA, then its replicas will be retained when propagating.
+	ResourceScaledByHPALabel = "horizontalpodautoscaler.karmada.io/name"
```
XiShanYongYe-Chang, line 33: Can we set the key more specific so that users can understand the meaning of the key as soon as they see it?

chaosi-zju: so, how? I may not got your mean

pkg/controllers/hpareplicassyncer/hpa_scale_ref_worker.go
```diff
@@ -0,0 +1,134 @@
...
+	return scaleRef, mapping, nil
+}
+
+func (r *HPAReplicasSyncer) patchScaleRef(resource schema.GroupVersionResource, scaleRef, newScaleRef *unstructured.Unstructured) error {
+	patchBytes, err := helper.GenMergePatch(scaleRef, newScaleRef)
+	if err != nil {
+		return fmt.Errorf("failed to gen merge patch: %+v", err)
+	}
+	if len(patchBytes) == 0 {
+		klog.Infof("hpa labels not exist, skip deleting (%+v)", resource)
```
lxtywypc, line 118: I'm a little confused about this message. Both adding and deleting actions call this method, why empty patchBytes only matches deleting action?

pkg/controllers/hpareplicassyncer/hpa_scale_ref_worker.go
```diff
@@ -0,0 +1,134 @@
...
+		return nil
+	}
+
+	_, err = r.DynamicClient.Resource(resource).Namespace(newScaleRef.GetNamespace()).
+		Patch(context.Background(), newScaleRef.GetName(), types.MergePatchType, patchBytes, metav1.PatchOptions{})
+
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			klog.V(4).Infof("scale ref resource is not found (%+v), skip processing", resource)
```
lxtywypc, line 127: It might be better to add resource name and namespace. As well as in other similar info messages.

pkg/controllers/hpareplicassyncer/hpa_scale_ref_worker.go
```diff
@@ -0,0 +1,134 @@
...
+
+	newScaleRef := scaleRef.DeepCopy()
+	util.MergeLabel(newScaleRef, util.ResourceScaledByHPALabel, hpa.GetName())
+
+	// use patch is better than update, when modification occur after get, patch can still success while update can not
+	if err = r.patchScaleRef(mapping.Resource, scaleRef, newScaleRef); err != nil {
+		return err
+	}
+
+	klog.Infof("add hpa labels to %+v success", hpa.Spec.ScaleTargetRef)
```
lxtywypc, line 69: Would it be better to add namespace info?

pkg/controllers/hpareplicassyncer/hpa_replicas_syncer_predicate.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	newHPA, ok := e.ObjectNew.(*autoscalingv2.HorizontalPodAutoscaler)
+	if !ok {
+		klog.Errorf("update predicates in hpa controller called, but new obj is not hpa type")
+		return false
+	}
+
+	// hpa scale ref changed, remove old hpa label and add to new hpa
+	if oldHPA.Spec.ScaleTargetRef.String() != newHPA.Spec.ScaleTargetRef.String() {
```
XiShanYongYe-Chang, line 45: Is it necessary to judge `hasBeenPropagated` here?

pkg/controllers/hpareplicassyncer/hpa_replicas_syncer_predicate.go
```diff
@@ -0,0 +1,79 @@
...
+
+func (r *HPAReplicasSyncer) Delete(e event.DeleteEvent) bool {
+	hpa, ok := e.Object.(*autoscalingv2.HorizontalPodAutoscaler)
+	if !ok {
+		klog.Errorf("delete predicates in hpa controller called, but obj is not hpa type")
+		return false
+	}
+
+	// if scale ref has label, remove label, otherwise skip
+	r.ScaleRefWorker.Add(labelEvent{deleteLabelEvent, hpa})
```
XiShanYongYe-Chang, line 66: Is it necessary to judge `hasBeenPropagated` here?

pkg/controllers/hpareplicassyncer/hpa_replicas_syncer_predicate.go
```diff
@@ -0,0 +1,79 @@
...
+
+	return false
+}
+
+func (r *HPAReplicasSyncer) Generic(e event.GenericEvent) bool {
+	return e.Object.GetObjectKind().GroupVersionKind().Kind == HorizontalPodAutoscalerKind
```
XiShanYongYe-Chang, line 72: The Generic Event is not used yet. This part can be ignored.

pkg/controllers/hpareplicassyncer/hpa_scale_ref_worker.go
```diff
@@ -0,0 +1,134 @@
...
+	addLabelEvent labelEventKind = iota
+	deleteLabelEvent
+)
+
+type labelEvent struct {
+	kind labelEventKind
+	hpa  *autoscalingv2.HorizontalPodAutoscaler
+}
+
+func (r *HPAReplicasSyncer) ReconcileScaleRef(key util.QueueKey) (err error) {
```
XiShanYongYe-Chang, line 32: This method seems do not need to be exported.

pkg/resourceinterpreter/default/native/retain.go
```diff
@@ -122,3 +124,23 @@ func retainJobSelectorFields(desired, observed *unstructured.Unstructured) (*uns
...
+func retainK8sWorkloadReplicas(desired, observed *unstructured.Unstructured) (*unstructured.Unstructured, error) {
+	labels, exist, err := unstructured.NestedStringMap(desired.Object, "metadata", "labels")
+	if err != nil {
+		return nil, err
+	}
+
+	if _, hpaLabelExist := labels[util.ResourceScaledByHPALabel]; exist && hpaLabelExist {
+		replicas, exist, err := unstructured.NestedInt64(observed.Object, "spec", "replicas")
+		if err != nil || !exist {
+			return nil, fmt.Errorf("failed to get spec.replicas from deployment %s/%s", observed.GetName(), observed.GetNamespace())
```
lxtywypc, line 137: If this function could be used for other workloads, it might be better to use `observed.GetKind()` instead of `deployment`.

Or it might be better to change the name of function to make it specific.

XiShanYongYe-Chang (commented):
Thanks a lot~

pkg/resourceinterpreter/default/native/retain.go
chaunceyjiang: >  you mean adding a default operation to process all resources and retain the replicas as long as it contain the target label. Am I right?

Yes.

pkg/resourceinterpreter/default/native/retain.go
chaunceyjiang: +1

pkg/resourceinterpreter/default/native/retain.go
XiShanYongYe-Chang: Agree!

pkg/util/constants.go
chaosi-zju: how about: 

```go
// ResourceScaledByHPALabel is added to mark resource template scaled by HPA, possible conflict-resolution of replicas: retain-membercluster
ResourceScaledByHPALabel = "resolve-replicas-conflict-when-scaled-by-hpa"
```

i mean `resolve-replicas-conflict-when-scaled-by-hpa: retain-membercluster`

XiShanYongYe-Chang: Is it not limited to hpa? If the user adds this label, the function will also take effect. Therefore, we can focus on the function.

pkg/controllers/hpareplicassyncer/hpa_scale_ref_worker.go
chaosi-zju: nice finding, at the begining there are two same code in each function, I refactored the code to just one function but mistakenly overlooked this log content

pkg/util/constants.go
XiShanYongYe-Chang: How about `resolve-replicas-conflict: retain`

XiShanYongYe-Chang: cc @RainbowMango

pkg/controllers/hpareplicassyncer/hpa_scale_ref_worker.go
chaosi-zju: my refactor didn't really reduce code lines, I plan to change it back

pkg/controllers/hpareplicassyncer/hpa_replicas_syncer_predicate.go
chaosi-zju: can not do that, for `delete label` operation：
1. if hpa didn't propagated -> deployment has no labels -> delete label will be skipped. (still success)
2. if hpa propagated -> deployment has labels -> if pp of hpa is deleted earlier -> deployements' label should be deleted (judge `hasBeenPropagated` may result in skip delete)

pkg/controllers/hpareplicassyncer/hpa_replicas_syncer_predicate.go
chaosi-zju: ditto, not

pkg/controllers/hpareplicassyncer/hpa_replicas_syncer_predicate.go
chaosi-zju: fucntion must keep, for ensure complete implementation of interface. you mean we can just `return fasle`?

pkg/controllers/hpareplicassyncer/hpa_scale_ref_worker.go
chaosi-zju: done

pkg/controllers/hpareplicassyncer/hpa_replicas_syncer_predicate.go
XiShanYongYe-Chang: Yes, it is.

pkg/controllers/hpareplicassyncer/hpa_replicas_syncer_controller.go
```diff
@@ -9,51 +9,49 @@ import (
...
+	r.ScaleRefWorker = util.NewAsyncWorker(scaleRefWorkerOptions)
+	r.ScaleRefWorker.Run(scaleRefWorkerNum, context.Background().Done())
+
+	return controllerruntime.NewControllerManagedBy(mgr).
+		Named(ControllerName).
+		For(&autoscalingv2.HorizontalPodAutoscaler{}, builder.WithPredicates(r)).
 		Complete(r)
 }
 
+var _ reconcile.Reconciler = &HPAReplicasSyncer{}
```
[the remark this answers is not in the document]
chaosi-zju: deleted

pkg/controllers/hpareplicassyncer/hpa_replicas_syncer_predicate.go
XiShanYongYe-Chang: thx, I get it.

pkg/controllers/hpareplicassyncer/hpa_replicas_syncer_predicate.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+func hasBeenPropagated(hpa *autoscalingv2.HorizontalPodAutoscaler) bool {
+	ppLabel, ppExist := hpa.GetLabels()[policyv1alpha1.PropagationPolicyUIDLabel]
+	cppLabel, cppExist := hpa.GetLabels()[policyv1alpha1.ClusterPropagationPolicyUIDLabel]
+	return (ppExist && ppLabel != "") || (cppExist && cppLabel != "")
```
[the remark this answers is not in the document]
chaosi-zju: if not exist, the `ppLabel` string actually is empty. but what i considered is label exist but string value is empty, shall I take it into account?

pkg/resourceinterpreter/default/native/retain.go
chaosi-zju: done

pkg/util/constants.go
RainbowMango: I think a label should contain the Karmada domain name so that people can figure out if it is introduced/added by Karmada.

Propose another one:
```goland
	// RetainReplicas is a reserved label to indicate whether the replicas should be retained.
	RetainReplicas = "resourcetemplate.karmada.io/retain-replicas"
```

Given this label is oriented at `resource template`, so it could be organized to `resourcetemplate.karmada.io`.

E.g.

```
resourcetemplate.karmada.io/retain-replicas: true   // with value `true` indicates retain
resourcetemplate.karmada.io/retain-replicas: false  // with value `false` and others, indicates not retain
```

@chaunceyjiang What do you think?

pkg/resourceinterpreter/default/native/retain.go
XiShanYongYe-Chang: Hi @chaunceyjiang, We just tried it, and setting it to the default path may not work, because one reason is that if we set it to the default path, `HookEnable` is not registered and the existing objectwatcher needs to be adjusted. (If the modification is required, the modification is related to the reconstruction of the `HookEnable` interface.)
Cause 2: For the CRD resource, the replicas of the CRD resource is uncertain. If the CRD resource is set to the default path, some user scalability is lost.

XiShanYongYe-Chang: So, my suggestion is to keep the original writing first, but I think the `default path thing` is still meaningful, it still needs to be done, and I will add a new task item to the issue to describe the task. If you are interested, you can also assign this task.
How do you think @chaunceyjiang

XiShanYongYe-Chang:
Ask an again review from @chaunceyjiang @lxtywypc

XiShanYongYe-Chang:
Thanks~
/lgtm

RainbowMango (approved):
/approve

@chaunceyjiang @lxtywypc 
Please feel free to leave your comments here or https://github.com/karmada-io/karmada/issues/4058 if we missed something :) We still have time to iterate it.

Issue #3421, opened 2023-04-18
Fix karmada-bot /retest


helen-frank:
**What would you like to be added**:
Fix karmada-bot /retest
**Why is this needed**:
Accelerated PR test

helen-frank:
@RainbowMango Do you have time to repair the karmada-bot `/retest` function? It has been broken for a long time.😂

RainbowMango:
It's not broken, it's never working~

The robot command `/retest` is used to trigger the tests managed by `prombot`, but Karmada's tests are managed by `GitHub Action`, I don't think a robot can do that.

helen-frank:
> It's not broken, it's never working~
> 
> The robot command `/retest` is used to trigger the tests managed by `prombot`, but Karmada's tests are managed by `GitHub Action`, I don't think a robot can do that.

emmm, I haven't actually used `github action`, so I'm not quite sure what's going on here, except to see that other communities are available

whitewindmills:
+1
It's cool if the command `/retest` is available.

RainbowMango:
I'm going to close this issue for now as we don't have an ideal solution yet.

Now we should focus on building a stable testing environment.
/close

liangyuanpeng:
After my testing on prow, the current prow already supports the retest operation on github. 

here is my config for it:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  namespace: prow
  name: plugins
data:
  plugins.yaml: |
    triggers:
    - repos:
      - liangyuanpeng/liangyuanpeng.github.io
      - liangyuanpeng/jxline 
      only_org_members: true
      trigger_github_workflows: true
...
```

Just add configuration of `trigger_github_workflows: true` and it's working.

Check the PR of prow: https://github.com/kubernetes/test-infra/pull/25736

@RainbowMango

RainbowMango:
Fantastic! Which image version you are using? By the way, do you know where I can find all available prow images?

liangyuanpeng:
@RainbowMango 

This feature was introduced by this PR https://github.com/kubernetes/test-infra/pull/24726 , and the corresponding prow image version is: `v20220105-8e97316bf9`, This is the minimum prow version that you want to use `/retest`  for github action. The format is `{date}-{shortcommit}`

There is a known issue with `/retest` which seems to have some issues with PRs submitted by the depentbot bot

> do you know where I can find all available prow images?

[...]

PR #4094, opened 2023-09-30
Use Patch() when binding-status controller update workload's status

### LINKED ISSUE #4093: Binding-status controller update workload's status error rate too high

In our environment, we watched binding-status controller have 10% - 35% chance got 409(conflict) HTTP error when update workload resource's status, and it causes binding-status controller reconciliation error rate is also 10% - 35%.

Just fix the surface phenomenon is easy, there are many ways could fix this like use json patch instead of update(#4094 if you like), or get the newest object before update. But I'm still confused about what is the root cause. 

Usually after we update workload spec once, binding-status controller will be triggered to update the binding and workload status 8 - 15 times, but the HTTP PUT 409 error rate of updating ResourceBinding is only 3% - 5%. I highly suspect the slow work of the informer of untyped objects, but it's hard to statistic the duration from successful update to watch update event. BTW this both happen in dev and prod environment, dev env only have very few objects.

And here's the question: Did anyone ever meet this? Does anyone have any research on this?

**Environment**:
- Karmada version: branch forked from v1.5.1, cherry-picked #3202 and some other commits, and added some other customized changes which is not relevant to this issue.
- Kubernetes version: 1.19.3
- Others:

zach593:
Change-Id: Ia0b5fde0b9d69f73aa6c2e3135e2590a6a503075

**What type of PR is this?**

/kind cleanup

Discribed in  #4093

Fixes #4093

```release-note
`karmada-controller-manager: use Patch() instead of Update() when updating workload status because error rate is too high.`
```

### DIFF
--- a/pkg/controllers/status/common.go
+++ b/pkg/controllers/status/common.go
@@ -8,6 +8,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
@@ -18,6 +19,7 @@ import (
 	workv1alpha1 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha1"
 	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
 	"github.com/karmada-io/karmada/pkg/resourceinterpreter"
+	"github.com/karmada-io/karmada/pkg/util/helper"
 	"github.com/karmada-io/karmada/pkg/util/restmapper"
 )
 
@@ -69,13 +71,24 @@ func updateResourceStatus(
 		klog.Errorf("Failed to aggregate status for resource(%s/%s/%s, Error: %v", gvr, resource.GetNamespace(), resource.GetName(), err)
 		return err
 	}
-	if reflect.DeepEqual(resource, newObj) {
+
+	oldStatus, _, _ := unstructured.NestedFieldNoCopy(resource.Object, "status")
+	newStatus, _, _ := unstructured.NestedFieldNoCopy(newObj.Object, "status")
+	if reflect.DeepEqual(oldStatus, newStatus) {
 		klog.V(3).Infof("Ignore update resource(%s/%s/%s) status as up to date.", gvr, resource.GetNamespace(), resource.GetName())
 		return nil
 	}
 
-	if _, err = dynamicClient.Resource(gvr).Namespace(resource.GetNamespace()).UpdateStatus(context.TODO(), newObj, metav1.UpdateOptions{}); err != nil {
-		klog.Errorf("Failed to update resource(%s/%s/%s), Error: %v", gvr, resource.GetNamespace(), resource.GetName(), err)
+	patchBytes, err := helper.GenReplaceFieldJSONPatch("/status", oldStatus, newStatus)
+	if err != nil {
+		klog.Errorf("Failed to gen patch bytes for resource(%s/%s/%s, Error: %v", gvr, resource.GetNamespace(), resource.GetName(), err)
+		return err
+	}
+
+	_, err = dynamicClient.Resource(gvr).Namespace(resource.GetNamespace()).
+		Patch(context.TODO(), resource.GetName(), types.JSONPatchType, patchBytes, metav1.PatchOptions{}, "status")
+	if err != nil {
+		klog.Error("Failed to update resource(%s/%s/%s), Error: %v", gvr, resource.GetNamespace(), resource.GetName(), err)
 		return err
 	}
 	klog.V(3).Infof("Update resource(%s/%s/%s) status successfully.", gvr, resource.GetNamespace(), resource.GetName())
--- a/pkg/util/helper/patch.go
+++ b/pkg/util/helper/patch.go
@@ -3,9 +3,27 @@ package helper
 import (
 	"encoding/json"
 	"fmt"
+	"reflect"
 
 	jsonpatch "github.com/evanphx/json-patch/v5"
 )
+
+// RFC6902 JSONPatch operations
+const (
+	JSONPatchOPAdd     = "add"
+	JSONPatchOPReplace = "replace"
+	JSONPatchOPRemove  = "remove"
+	JSONPatchOPMove    = "move"
+	JSONPatchOPCopy    = "copy"
+	JSONPatchOPTest    = "test"
+)
+
+type jsonPatch struct {
+	OP    string      `json:"op"`
+	From  string      `json:"from,omitempty"`
+	Path  string      `json:"path"`
+	Value interface{} `json:"value,omitempty"`
+}
 
 // GenMergePatch will return a merge patch document capable of converting the
 // original object to the modified object.
@@ -31,4 +49,41 @@ func GenMergePatch(originalObj interface{}, modifiedObj interface{}) ([]byte, er
 	}
 
 	return patchBytes, nil
+}
+
+// GenReplaceFieldJSONPatch returns the RFC6902 JSONPatch array as []byte, which is used to simply
+// add/replace/delete certain JSON **Object** field.
+func GenReplaceFieldJSONPatch(path string, originalFieldValue, newFieldValue interface{}) ([]byte, error) {
+	if reflect.DeepEqual(originalFieldValue, newFieldValue) {
+		return []byte(`[]`), nil
+	}
+	if newFieldValue == nil {
+		return GenJSONPatch(JSONPatchOPRemove, "", path, nil)
+	}
+	// The implementation of “add” and “replace” for JSON objects is actually the same
+	// in “github.com/evanphx/json-patch/v5”, which is used by us and k8s.
+	// We implemented it here just to follow the RFC6902.
+	if originalFieldValue == nil {
+		return GenJSONPatch(JSONPatchOPAdd, "", path, newFieldValue)
+	}
+	return GenJSONPatch(JSONPatchOPReplace, "", path, newFieldValue)
+}
+
+// GenJSONPatch return JSONPatch array as []byte according to RFC6902
+func GenJSONPatch(op, from, path string, value interface{}) ([]byte, error) {
+	jp := jsonPatch{
+		OP:   op,
+		Path: path,
+	}
+	switch op {
+	case JSONPatchOPAdd, JSONPatchOPReplace, JSONPatchOPTest:
+		jp.Value = value
+	case JSONPatchOPMove, JSONPatchOPCopy:
+		jp.From = from
+	case JSONPatchOPRemove:
+	default:
+		return nil, fmt.Errorf("unrecognized JSONPatch OP: %s", op)
+	}
+
+	return json.Marshal([]jsonPatch{jp})
 }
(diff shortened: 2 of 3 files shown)


RainbowMango:
Hi @zach593 Karmada hasn't enable `/retest` now, and it is tracked by #3421 

The failing CI is due to a CI issue and it was resolved by #4107 yesterday. You might need to rebase this PR and push again.

zach593:
Cool, thanks for following this.

RainbowMango:
cc @chaunceyjiang @XiShanYongYe-Chang Can you help with this?

XiShanYongYe-Chang:
ok
/assign

pkg/controllers/status/common.go
```diff
@@ -69,13 +72,28 @@ func updateResourceStatus(
...
+	if newStatus != nil {
+		j, err := json.Marshal(newStatus)
+		if err != nil {
+			klog.Errorf("Failed to marshal status for resource(%s/%s/%s, Error: %v", gvr, resource.GetNamespace(), resource.GetName(), err)
+			return err
+		}
+		statusPatch = []byte(fmt.Sprintf(`[{"op": "replace", "path": "/status", "value": %s}]`, j))
+	} else {
+		statusPatch = []byte(`[{"op": "remove", "path": "/status"}]`)
+	}
```
XiShanYongYe-Chang, line 93: Is it possible to consider using `helper.GenMergePatch`?

zach593: @XiShanYongYe-Chang The reason I used JSONPatchType instead of MergePatchType is that I am worried that some fields may not be deleted when patch status, because we do not understand and cannot modify the definition of resource templates and the logic of related controllers. We even provide a webhook so that users can customize the status.

If you think jsonpatch is acceptable, I can move this part of the construction logic to helper/patch.go (I didn't notice this file existed before); or if you think merge is still a better choice, Then I will update this part.


Obviously I should reply here...😆

XiShanYongYe-Chang: :) haha

zach593: I have done some research and conducted some tests, and it seems that MergePatch can completely replace JSONPatch. Therefore, I choose to use MergePatch here. We can monitor if anyone reports any issues in the future.

zach593: done.

XiShanYongYe-Chang: Never mind. Full discussion and testing gave us a much better result 👍

zach593: Updated. kindly ping @XiShanYongYe-Chang to review again.

pkg/util/helper/patch_test.go
```diff
@@ -106,3 +107,157 @@ func TestGenMergePatch(t *testing.T) {
 		})
 	}
 }
+
+func TestGenJSONPatch(t *testing.T) {
+	//doc := `{"field1":1,"field2":"2","field3":[3],"field4":{"a":"a"}}`
```
XiShanYongYe-Chang, line 112: What does this comment mean?

pkg/util/helper/patch.go
```diff
@@ -32,3 +50,40 @@ func GenMergePatch(originalObj interface{}, modifiedObj interface{}) ([]byte, er
 
 	return patchBytes, nil
 }
+
+// GenReplaceFieldJSONPatch returns the RFC6902 JSONPatch array as []byte, which is used to simply
+// add/replace/delete certain JSON **Object** field.
+func GenReplaceFieldJSONPatch(path string, originalFieldValue, newFieldValue interface{}) ([]byte, error) {
```
XiShanYongYe-Chang, line 56: Can you help add some ut for this function?

XiShanYongYe-Chang (commented):
Thanks @zach593.
I'm sorry for replying late~

pkg/util/helper/patch.go
```diff
@@ -32,3 +50,40 @@ func GenMergePatch(originalObj interface{}, modifiedObj interface{}) ([]byte, er
...
+	if reflect.DeepEqual(originalFieldValue, newFieldValue) {
+		return []byte(`[]`), nil
+	}
+	if newFieldValue == nil {
+		return GenJSONPatch(JSONPatchOPRemove, "", path, nil)
+	}
+	// The implementation of “add” and “replace” for JSON objects is actually the same
+	// in “github.com/evanphx/json-patch/v5”, which is used by us and k8s.
```
XiShanYongYe-Chang, line 64: The presence of `us` word could be misleading.

`k8s` -> `K8s`

pkg/controllers/status/common.go
XiShanYongYe-Chang: Thank you for your explanation, I probably understand what you mean, can you help write an example to describe it, so that it is easier to follow?

pkg/controllers/status/common.go
```diff
@@ -69,13 +71,24 @@ func updateResourceStatus(
 		klog.Errorf("Failed to aggregate status for resource(%s/%s/%s, Error: %v", gvr, resource.GetNamespace(), resource.GetName(), err)
 		return err
 	}
-	if reflect.DeepEqual(resource, newObj) {
+
+	oldStatus, _, _ := unstructured.NestedFieldNoCopy(resource.Object, "status")
+	newStatus, _, _ := unstructured.NestedFieldNoCopy(newObj.Object, "status")
+	if reflect.DeepEqual(oldStatus, newStatus) {
```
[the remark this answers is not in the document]
zach593: Which part/scenario you mean?
`unstructured.NestedFieldNoCopy()` or `reflect.DeepEqual()`?
The resource never got the status field, or will the status be added/removed during this operation?

pkg/util/helper/patch.go
XiShanYongYe-Chang, line 56: Is there an implementation of this method in Kubernetes that we can call directly?

pkg/controllers/status/common.go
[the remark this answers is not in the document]
XiShanYongYe-Chang: Sorry for not being descriptive. I meant this `The resource never got the status field`.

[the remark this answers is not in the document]
zach593: image
image

I think it's easier to explain it this way :P

pkg/util/helper/patch_test.go
zach593: should be deleted, I'll update later.

pkg/controllers/status/common.go
```diff
@@ -69,13 +71,24 @@ func updateResourceStatus(
...
 		klog.V(3).Infof("Ignore update resource(%s/%s/%s) status as up to date.", gvr, resource.GetNamespace(), resource.GetName())
 		return nil
 	}
 
-	if _, err = dynamicClient.Resource(gvr).Namespace(resource.GetNamespace()).UpdateStatus(context.TODO(), newObj, metav1.UpdateOptions{}); err != nil {
-		klog.Errorf("Failed to update resource(%s/%s/%s), Error: %v", gvr, resource.GetNamespace(), resource.GetName(), err)
+	patchBytes, err := helper.GenReplaceFieldJSONPatch("/status", oldStatus, newStatus)
```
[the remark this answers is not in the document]
zach593: It will return ```[]byte(`[]`), nil```, ```[]byte(`[]`)``` representing an empty jsonpatch, which will be safely passed to apiserver when patching, the objects we patch will always `no change`. You can test it with kubectl.

pkg/controllers/status/common.go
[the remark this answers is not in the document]
XiShanYongYe-Chang: Thanks for your example.

pkg/util/helper/patch.go
zach593: Didn't see any jsonpatch helpers, only strategypatch and mergepatch helpers in `/staging/src/k8s.io/apimachinery/pkg/util/strategicpatch` and `/staging/src/k8s.io/apimachinery/pkg/util/jsonmergepatch`.

XiShanYongYe-Chang: Ok, thanks for your reply~

pkg/controllers/status/common.go
zach593: A simple example could be like this: [playground](https://go.dev/play/p/w0GjAogpevo).
If everything works on time but the informer holding an outdated resource template (currentA), using mergePatch may create an inappropriate mergePatchBytes for actual template object (currentB).

pkg/controllers/status/common.go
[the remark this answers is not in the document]
zach593: Saw that `GenMergePatch()` returns `nil` if `patchBytes == "{}"`, so I followed this, returned `nil` when no patching is required.

pkg/util/helper/patch.go
zach593: Done.

pkg/util/helper/patch.go
zach593: Done.

XiShanYongYe-Chang (commented):
Thanks a lot~
Can we then use this logic for ResourceBinding and Work status updates?

/lgtm
/cc @RainbowMango

zach593:
Updated some UT test name.. @XiShanYongYe-Chang 

While I feel that a controller that takes ownership of an object should not cause conflict errors when updating an object, and there seems to be no obvious problem with using `Patch()` to replace `Update()`, extending the use of `Patch()` may still require more discussion.

XiShanYongYe-Chang:
I got it, thanks @zach593 
/lgtm

RainbowMango (commented):
/assign

pkg/controllers/status/common.go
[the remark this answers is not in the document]
chaunceyjiang: Just a question, can we use `retry.RetryOnConflict()` ?

XiShanYongYe-Chang:
Kindly ping @RainbowMango

pkg/controllers/status/common.go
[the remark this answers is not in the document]
whitewindmills: I don't see how to prevent empty patches from being sent to apiserver.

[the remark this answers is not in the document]
zach593: there's a `DeepEqual()` judge in line 77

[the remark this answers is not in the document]
whitewindmills: okay, thanks for reminding me. can you make the patch more precise?
```
status:
  availableReplicas: 1
  conditions:
  - lastTransitionTime: "2023-12-05T09:25:15Z"
    lastUpdateTime: "2023-12-05T09:25:15Z"
    message: Deployment has minimum availability.
    reason: MinimumReplicasAvailable
    status: "True"
    type: Available
  - lastTransitionTime: "2023-11-21T11:43:12Z"
    lastUpdateTime: "2023-12-05T09:25:15Z"
    message: ReplicaSet "demo-57755f7c74" has successfully
      progressed.
    reason: NewReplicaSetAvailable
    status: "True"
    type: Progressing
  observedGeneration: 6
  readyReplicas: 0
  replicas: 1
  updatedReplicas: 1
```
if only the field `readyReplicas` changes from 0 to 1, we'd better use such a patch rather than the entire status:
```
[{
"op":"replace",
"path":"/status/readyReplicas",
"value":1
}]
```

chaunceyjiang:
Nice!!!

/lgtm

pkg/controllers/status/common.go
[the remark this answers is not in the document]
zach593: I believe implementing such functionality can indeed make the final patch appear very concise in certain situations (as you illustrated). However, it may introduce other issues: if there's a need to update multiple fields, should a separate patch operation be written for each field? If it's necessary to consolidate certain patch operations into one, what is the appropriate timing for the consolidation?

Implementing it this way seems like using JSONPatch to emulate MergePatch. If the goal is simplicity in patch operations, it might prove to be more trouble than it's worth in many scenarios, and the implementation is likely to be less straightforward.

If the concern is that other controllers might modify the status, as mentioned in https://github.com/karmada-io/karmada/issues/4093#issuecomment-1838828261, I believe this possibility is low. In the current implementation, `.status` is returned by the interpreterWebhook, and users have the opportunity to introduce other mechanisms for custom handling.

RainbowMango:
Hi @whitewindmills Do you have any comments?

RainbowMango (commented):
/approve

Issue #4111, opened 2023-10-09
CI APIServer compatibility is breaking


liangyuanpeng:
https://github.com/karmada-io/karmada/actions/workflows/ci-schedule-compatibility.yaml

![image](https://github.com/karmada-io/karmada/assets/28711504/5066923b-984c-49a0-8852-bb060ed35fd9)

setup e2e env timeout 

```log
...
Note: To enable scheduler estimator, please deploy other scheduler estimators of all clusters.
      After that, specify the option '--enable-scheduler-estimator=true' of karmada-scheduler.
Waiting for kubeconfig file /home/runner/.kube/members.config and clusters member3 to be ready...
Waiting for running............................................................................................................................................................................................................................................................................................................
Error:  Timeout waiting for condition running
Error: Process completed with exit code 1.
```

liangyuanpeng:
I'm trying to fix this by increasing the timeout, one strange thing is why PR CI does not time out.

jwcesign:
Hi @liangyuanpeng , I tried to limit the concurrency: https://github.com/jwcesign/karmada/blob/720c0db1a94bd863b08066785f1d9dff93315521/.github/workflows/ci-schedule-compatibility.yaml#L11C3-L11C3

It looks help and not time out happened

liangyuanpeng:
Do something for visibility of this workflow.

![image](https://github.com/karmada-io/karmada/assets/28711504/0ffec532-0645-49de-8c8d-f0e669533fdb)

PR #4115, opened 2023-10-10
[CI] schedule workflow support workflow dispatch

liangyuanpeng:
**What type of PR is this?**

/kind cleanup
/kind failing-test

Let schedule workflow support call normal and do not fail-fast for `APIServer compatibility`, Run all of workflow job for all of version.

It's all failed in my fork repo of karmada.

image

Releated: #4111 

```release-note

```

### DIFF
--- a/.github/workflows/ci-schedule-compatibility.yaml
+++ b/.github/workflows/ci-schedule-compatibility.yaml
@@ -1,5 +1,6 @@
 name: APIServer compatibility
 on:
+  workflow_dispatch:
   schedule:
     # Run this workflow "At 20:00 UTC on Sunday and Saturday"
     - cron: '0 20 * * 0,6'
@@ -11,6 +12,7 @@ jobs:
     if: ${{ github.repository == 'karmada-io/karmada' }}
     runs-on: ubuntu-22.04
     strategy:
+      fail-fast: false
       matrix:
         kubeapiserver-version: [ v1.21.10, v1.22.7, v1.23.4, v1.24.2, v1.25.0, v1.26.0, v1.27.3 ]
         karmada-version: [ release-1.4, release-1.5, release-1.6 ]
--- a/.github/workflows/ci-schedule.yml
+++ b/.github/workflows/ci-schedule.yml
@@ -1,5 +1,6 @@
 name: CI Schedule Workflow
 on:
+  workflow_dispatch:
   schedule:
     # Run this workflow "At 18:00 UTC on Sunday and Saturday"
     - cron: '0 18 * * 0,6'


liangyuanpeng:
Run workflow once immediately  would be great after this PR is merged.

```shell
lan@lan:~/repo/git$ export GITHUB_TOKEN=123
lan@lan:~/repo/git$ cd karmada
lan@lan:~/repo/git/karmada$ gh workflow run -R karmada-io/karmada
? Select a workflow  [Use arrows to move, type to filter]
> APIServer compatibility (ci-schedule-compatibility.yaml)
```

Choose the `APIServer compatibility` and run it.

liangyuanpeng:
@RainbowMango 

Do you have any concerns about this PR? If you don't want to be able to manually call these two workflows currently, then I think at least `APIServer compatibility` should not fail quickly, and we need to know all the results, Remind me if i missed something,Thanks,

jwcesign:
> at least APIServer compatibility should not fail quickly

+1

.github/workflows/ci-schedule-compatibility.yaml
```diff
@@ -11,6 +12,7 @@ jobs:
     if: ${{ github.repository == 'karmada-io/karmada' }}
     runs-on: ubuntu-22.04
     strategy:
+      fail-fast: false
```
RainbowMango, line 15: LGTM

.github/workflows/ci-schedule-compatibility.yaml
```diff
@@ -1,5 +1,6 @@
 name: APIServer compatibility
 on:
+  workflow_dispatch:
```
RainbowMango, line 3: The [workflow_dispatch](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch) can trigger the workflow manually. Does that mean all developers can trigger the workflow at anytime?

RainbowMango: > Do you have any concerns about this PR? If you don't want to be able to manually call these two workflows currently, then I think at least APIServer compatibility should not fail quickly,

Yes, it makes sense to me that let the scheduled workflow continue to run and avoid canceling due to other job failures.
But manually triggering the workflow is concerning.

RainbowMango: kindly ping @liangyuanpeng

liangyuanpeng: Hi @RainbowMango  I have move the change of `fail-fast: false` to https://github.com/karmada-io/karmada/pull/4238 and there are fix a problem of timeout. 

I want to keep this PR is open and i think let the workflow support workflow dispatch still relevant. Because I have done some exploration on combining this scheduled task with prow, I hope I will have an update on this aspect soon.

PR #4116, opened 2023-10-10
ci: fix schedule ci workflow failure

jwcesign:
**What type of PR is this?**

/kind failing-test

Now `https://github.com/karmada-io/karmada/actions/workflows/ci-schedule.yml` and `https://github.com/karmada-io/karmada/actions/workflows/ci-schedule-compatibility.yaml` both failed almost, there are different errors(cluster creation timeout and e2e case failed).

The problem could be github resource limitations, I limit the concurrency to 2 and test it: https://github.com/jwcesign/karmada/actions/runs/6468715625/job/17561283243

Part of #4111 

none

```release-note

```

### DIFF
--- a/.github/workflows/ci-schedule-compatibility.yaml
+++ b/.github/workflows/ci-schedule-compatibility.yaml
@@ -11,6 +11,9 @@ jobs:
     if: ${{ github.repository == 'karmada-io/karmada' }}
     runs-on: ubuntu-22.04
     strategy:
+      # max-parallel limits the max number of jobs running at the same time.
+      # We set it to 5 to avoid too many jobs running at the same time, causing the CI to fail because of resource limitations.
+      max-parallel: 5
       fail-fast: false
       matrix:
         kubeapiserver-version: [ v1.23.4, v1.24.2, v1.25.0, v1.26.0, v1.27.3, v1.28.0, v1.29.0 ]
@@ -51,8 +54,12 @@ jobs:
         with:
           go-version: ${{ matrix.go-version }}
       - name: setup e2e test environment
-        run: |
-          hack/local-up-karmada.sh
+        uses: nick-fields/retry@v2.9.0
+        with:
+          max_attempts: 3
+          timeout_minutes: 20
+          command: |
+            hack/local-up-karmada.sh
       - name: change kube-apiserver and kube-controller-manager version
         run: |
           # Update images
--- a/.github/workflows/ci-schedule.yml
+++ b/.github/workflows/ci-schedule.yml
@@ -10,6 +10,9 @@ jobs:
     if: ${{ github.repository == 'karmada-io/karmada' }}
     runs-on: ubuntu-22.04
     strategy:
+      # max-parallel limits the max number of jobs running at the same time.
+      # We set it to 5 to avoid too many jobs running at the same time, causing the CI to fail because of resource limitations.
+      max-parallel: 5
       fail-fast: false
       matrix:
         k8s: [ v1.23.4, v1.24.2, v1.25.0, v1.26.0, v1.27.3, v1.28.0, v1.29.0 ]
@@ -38,7 +41,10 @@ jobs:
         with:
           go-version: 1.20.11
       - name: setup e2e test environment
-        run: |
+        uses: nick-fields/retry@v2.9.0
+        max_attempts: 3
+        timeout_minutes: 20
+        command: |
           export CLUSTER_VERSION=kindest/node:${{ matrix.k8s }}
           hack/local-up-karmada.sh
       - name: run e2e


liangyuanpeng (approved):
That's great.

RainbowMango:
@jwcesign Does this PR fixes #4111?

jwcesign:
> wcesign Does this PR fixes #4111?

No, I still haven't identified the root cause.

liangyuanpeng (commented):
I did three sets of comparative tests:

| testName  |  test Content |  CI result |
| ------------- | ------------- | ------------- |
| testA | only setup env and not run e2e test, and not set max-parallel | 5 ✔️  |
| testB  | it is test of today, no change. (setup env , run e2e test and not  set max-parallel) | 2✔️ 3  ❌  |
| testC  |  setup env, run e2e test and set max-parallel to 6 | 5  ✔️ |


image


TestA at https://github.com/liangyuanpeng/karmada/pull/31
TestB and TestC at https://github.com/liangyuanpeng/karmada/pull/32



I think this PR is going in the right direction, If this is not the root cause, then we can also set max-parallel first to make CI turn green.

@RainbowMango @jwcesign

.github/workflows/ci-schedule-compatibility.yaml
```diff
@@ -11,6 +11,7 @@ jobs:
     if: ${{ github.repository == 'karmada-io/karmada' }}
     runs-on: ubuntu-22.04
     strategy:
+      max-parallel: 2
```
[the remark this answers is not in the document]
jwcesign: Hi, @liangyuanpeng, I am not sure, I remember it also failed when I set 2(sometime).

Is there any way to retry it once failed(with 5/6 max-parallel), the threshold could be 3.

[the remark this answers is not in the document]
liangyuanpeng: > , I remember it also failed when I set 2(sometimes).

It used to contain other errors and now only env setup timeout error, I might think the problem you are having is a test case error.

> Is there any way to retry it once fails (with 5/6 max-parallel), the threshold could be 3.


```yaml

jobs:
  e2e:
  ...

  retry:
    name: retry test
    needs: [ e2e ]
    // waiting for all of e2e job ends and then uses the gh command to try again if necessary.
    
```

I once did something similar but I lost the working record, see https://github.com/karmada-io/karmada/issues/4111#issuecomment-1803544337

> https://github.com/marketplace/actions/retry-action

This would leave our github action job in a nested state, which I would probably avoid.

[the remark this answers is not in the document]
jwcesign: I modified like this: 
https://github.com/jwcesign/karmada/blob/f77fdb98d119cec1238bc0887fc36e4b89922ec3/.github/workflows/ci-schedule-compatibility.yaml#L49C1-L55C37

Waitting for the testing result:
https://github.com/jwcesign/karmada/actions/runs/7525890008/job/20483011518

This format looks fine(not really nested state), what do you think?

chaosi-zju, line 14: Is `2` concurrency actually only 2 instances running Karmada? I feel like the numbers are a bit conservative.

jwcesign: Yes, only two, as @liangyuanpeng tested, we  can set it as 5 or 6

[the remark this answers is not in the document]
liangyuanpeng: > I modified it like this:
https://github.com/jwcesign/karmada/blob/f77fdb98d119cec1238bc0887fc36e4b89922ec3/.github/workflows/ci-schedule-compatibility.yaml#L49C1-L55C37

Cool, It's seems easy to understand And the changes are very small.

```yaml
...
      - name: setup e2e test environment
        uses: nick-fields/retry@v2.9.0
        with:
          max_attempts: 3
          timeout_minutes: 20
          command: |
            hack/local-up-karmada.sh
...
```

jwcesign:
/cc @liangyuanpeng @XiShanYongYe-Chang @RainbowMango 

Please take a look

.github/workflows/ci-schedule-compatibility.yaml
```diff
@@ -51,8 +52,12 @@ jobs:
         with:
           go-version: ${{ matrix.go-version }}
       - name: setup e2e test environment
-        run: |
-          hack/local-up-karmada.sh
+        uses: nick-fields/retry@v2.9.0
+        with:
+          max_attempts: 3
```
XiShanYongYe-Chang, line 57: If the first two attempts failed and only the third attempt succeeded, can we retrieve the logs from the first two attempts?

.github/workflows/ci-schedule-compatibility.yaml
```diff
@@ -11,6 +11,7 @@ jobs:
     if: ${{ github.repository == 'karmada-io/karmada' }}
     runs-on: ubuntu-22.04
     strategy:
+      max-parallel: 5
```
XiShanYongYe-Chang, line 14: Here can add some comments for this iteration to facilitate subsequent development and understanding.

.github/workflows/ci-schedule-compatibility.yaml
jwcesign: It will have the full log:
image

https://github.com/jwcesign/karmada/actions/runs/7536016090/job/20512835547#step:5:796

.github/workflows/ci-schedule-compatibility.yaml
jwcesign: Sure

RainbowMango (approved):
/lgtm
/approve

PR #4158, opened 2023-10-20
Introduced leaderElection options including: `--leader-elect-lease-duration`, `--leader-elect-renew-deadline`, `--leader-elect-retry-period`

yanfeng1992:
**What type of PR is this?**
/kind feature

scheduler add LeaderElection config

Fixes #

```release-note
`karmada-scheduler`: Introduced leaderElection options including: `--leader-elect-lease-duration`, `--leader-elect-renew-deadline`, `--leader-elect-retry-period`, the default value not changed compared to previous version.
```

### DIFF
--- a/cmd/scheduler/app/options/options.go
+++ b/cmd/scheduler/app/options/options.go
@@ -98,6 +98,19 @@ func (o *Options) AddFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&o.LeaderElection.LeaderElect, "leader-elect", true, "Enable leader election, which must be true when running multi instances.")
 	fs.StringVar(&o.LeaderElection.ResourceName, "leader-elect-resource-name", "karmada-scheduler", "The name of resource object that is used for locking during leader election.")
 	fs.StringVar(&o.LeaderElection.ResourceNamespace, "leader-elect-resource-namespace", util.NamespaceKarmadaSystem, "The namespace of resource object that is used for locking during leader election.")
+	fs.DurationVar(&o.LeaderElection.LeaseDuration.Duration, "leader-elect-lease-duration", defaultElectionLeaseDuration.Duration, ""+
+		"The duration that non-leader candidates will wait after observing a leadership "+
+		"renewal until attempting to acquire leadership of a led but unrenewed leader "+
+		"slot. This is effectively the maximum duration that a leader can be stopped "+
+		"before it is replaced by another candidate. This is only applicable if leader "+
+		"election is enabled.")
+	fs.DurationVar(&o.LeaderElection.RenewDeadline.Duration, "leader-elect-renew-deadline", defaultElectionRenewDeadline.Duration, ""+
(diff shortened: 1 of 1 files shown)


RainbowMango:
If I remember correctly, they all have default values, could you please explain why the default values do not fit?

yanfeng1992:
Because in our environment, karmada-scheduler has restarted many times due to lease reasons.

image

image

RainbowMango:
Have you figured out why the instance can't renew the lease? Normally the default value is long enough, it might not work just raise the timeout.

(I'm saying that doesn't mean I don't like this patch, but trying to figure out the root cause.)

yanfeng1992:
kube-apiserver disruption can happen for multiple reasons, including
1.kube-apiserver rollout on a non-HA cluster
2.networking disruption on the host running the client
3.networking disruption on the host running the server

We have seen all of these cases, and more, disrupt connections. Many controllers, and operators rely on the kube-apiserver for making leader election. 

the kube-apiserver downtime tolerance is floor(renewDeadline/retryPeriod)*retryPeriod-retryPeriod. When using the default configuration, tolerance is floor(10/2)*2-2 =8s. In actual production, Leader election needs to be able to tolerate 60s of interruptions. Recommended defaults are LeaseDuration=137s, RenewDealine=107s, RetryPeriod=26s.

In addition, let me explain the original source of this PR and why we are so concerned about component restarts. In our large-scale environment, the number of some CRs is tens of thousands. Every time karmada-controller-manager is restarted, all objects need to be reconciled. If the synchronization of the informer cache is not completed within 30s at this time, the apiserver of the member cluster will be frequently accessed, causing the client to limit the current flow, resulting in the execution_controller's overall syncWork speed being very slow.

@RainbowMango

RainbowMango (approved):
/lgtm
/approve

Thanks for the detailed clarification and feedback. 

It looks good:
```
      --leader-elect                                                                                                                                                                               
                Enable leader election, which must be true when running multi instances. (default true)
      --leader-elect-lease-duration duration                                                                                                                                                       
                The duration that non-leader candidates will wait after observing a leadership renewal until attempting to acquire leadership of a led but unrenewed leader slot. This is effectively
                the maximum duration that a leader can be stopped before it is replaced by another candidate. This is only applicable if leader election is enabled. (default 15s)
      --leader-elect-renew-deadline duration                                                                                                                                                       
                The interval between attempts by the acting master to renew a leadership slot before it stops leading. This must be less than or equal to the lease duration. This is only applicable
                if leader election is enabled. (default 10s)
      --leader-elect-resource-name string                                                                                                                                                          
                The name of resource object that is used for locking during leader election. (default "karmada-scheduler")
      --leader-elect-resource-namespace string                                                                                                                                                     
                The namespace of resource object that is used for locking during leader election. (default "karmada-system")
      --leader-elect-retry-period duration                                                                                                                                                         
                The duration the clients should wait between attempting acquisition and renewal of a leadership. This is only applicable if leader election is enabled. (default 2s)
```

Issue #3854, opened 2023-07-28
karmada-controller-manager: panic: runtime error: index out of range


tedli:
**What happened**:

`karmada-controller-manager` crashed, caused by `panic: runtime error: index out of range`

**What you expected to happen**:

index checked, not to panic, not to crash.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

The `index` may be `-1`.
https://github.com/karmada-io/karmada/blob/2526de856ebe8b4e57cdf5ad9e544604aa9f0e43/pkg/modeling/modeling.go#L162-L168

https://github.com/karmada-io/karmada/blob/2526de856ebe8b4e57cdf5ad9e544604aa9f0e43/pkg/modeling/modeling.go#L105-L114

**Environment**:
- Karmada version: v1.6.1
- kubectl-karmada or karmadactl version (the result of `kubectl-karmada version` or `karmadactl version`):
- Others:

liangyuanpeng:
Any message for reproduce it? and Environment of kubernetes cluster and karmada cluster .

RainbowMango:
The searchLastLessElement might return -1:
https://github.com/karmada-io/karmada/blob/cf90ffc64fbfa16083c8371905b4789fb4f98a6f/pkg/modeling/modeling.go#L116-L138

@tedli Do you know how to reproduce it?

tedli:
Hi @RainbowMango  @liangyuanpeng ，

My case as I mententioned, was caused by line `164`, `165`
https://github.com/karmada-io/karmada/blob/2526de856ebe8b4e57cdf5ad9e544604aa9f0e43/pkg/modeling/modeling.go#L162-L168

And I didn't dive deep to find how to reproduce. I just havd a quick fix, that adding index check, to avoid panic.

RainbowMango:
/help

Let's call for help to figure out the root cause.

I'm not sure if the author(@halfrost) has time to take a look :)

halfrost:
I can fix this issue. index returns negative one, which means that no matching resource model can be found. Let me write some test code to test it.

halfrost:
@tedli Hi, could you provide some parameters to initialize resource modeling? I want to reproduce this issue on my local machine.

tedli:
Hi @halfrost ,

Sorry for late reply, here is my `Cluster` cr.

<details><summary>Click to view</summary>
<p>

```yaml
apiVersion: cluster.karmada.io/v1alpha1
[...]

PR #4176, opened 2023-10-25
fix: change defaultModelSorting as local variable

jwcesign:
**What type of PR is this?**
/kind bug

Please check: https://github.com/karmada-io/karmada/issues/3854

Part of https://github.com/karmada-io/karmada/issues/3854

none

```release-note
`karmada-controller-manager`: Fix panic caused by concurrent global variable reads and writes.
```

### DIFF
--- a/pkg/modeling/modeling.go
+++ b/pkg/modeling/modeling.go
@@ -4,7 +4,6 @@ import (
 	"container/list"
 	"errors"
 	"math"
-	"sync"
 
 	rbt "github.com/emirpasic/gods/trees/redblacktree"
 	corev1 "k8s.io/api/core/v1"
@@ -14,15 +13,11 @@ import (
 	clusterapis "github.com/karmada-io/karmada/pkg/apis/cluster/v1alpha1"
 )
 
-var (
-	mu                  sync.Mutex
-	defaultModelSorting []clusterapis.ResourceName
-)
-
 // ResourceSummary records the list of resourceModels
 type ResourceSummary struct {
-	RMs           []resourceModels
-	modelSortings [][]resource.Quantity
+	RMs                       []resourceModels
+	modelSortings             [][]resource.Quantity
+	modelSortingResourceNames []clusterapis.ResourceName
 }
 
 // resourceModels records the number of each allocatable resource models.
@@ -82,9 +77,6 @@ func InitSummary(resourceModel []clusterapis.ResourceModel) (ResourceSummary, er
 		return ResourceSummary{}, errors.New("the number of resourceName is not equal the number of resourceList")
 	}
 
-	if len(rsName) != 0 {
-		defaultModelSorting = rsName
-	}
 	rms := make([]resourceModels, len(rsList))
 	// generate a sorted array by first priority of ResourceName
 	modelSortings := make([][]resource.Quantity, len(rsName))
@@ -93,7 +85,7 @@ func InitSummary(resourceModel []clusterapis.ResourceModel) (ResourceSummary, er
 			modelSortings[i] = append(modelSortings[i], rsList[index][name])
 		}
 	}
-	return ResourceSummary{RMs: rms, modelSortings: modelSortings}, nil
+	return ResourceSummary{RMs: rms, modelSortings: modelSortings, modelSortingResourceNames: rsName}, nil
 }
 
 // NewClusterResourceNode create new cluster resource node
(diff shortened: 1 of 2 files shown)


jwcesign:
@halfrost

XiShanYongYe-Chang (commented):
LGTM

chaunceyjiang:
Have you found the root cause?

jwcesign:
@chaunceyjiang  The root case is the same as this(https://github.com/karmada-io/karmada/issues/3854#issuecomment-1719319187).

pkg/modeling/modeling.go
```diff
@@ -14,15 +13,11 @@ import (
 	clusterapis "github.com/karmada-io/karmada/pkg/apis/cluster/v1alpha1"
 )
 
-var (
-	mu                  sync.Mutex
-	defaultModelSorting []clusterapis.ResourceName
```
RainbowMango, line 19: The resource models might be different from each other, for example, cluster 1's model could be:
```yaml
resourceModels:
  - grade: 0
    ranges:
    - max: "1"
      min: "0"
      name: cpu        # Note that only declared cpu
```

and cluster 2's model could be
```yaml
resourceModels:
  - grade: 0
    ranges:
    - max: 4Gi
      min: "0"
      name: memory   # Note that only declared memory
```

The global variable `defaultModelSorting` actually shared by different clusters and might be overwritten during init the model at https://github.com/karmada-io/karmada/blob/3267155766f69083d211c4bc804e8bda4b18fd4d/pkg/controllers/status/cluster_status_controller.go#L595C35-L595C46.

@halfrost Please correct me if this is not the case.

pkg/modeling/modeling.go
```diff
@@ -177,11 +162,11 @@ func (rs *ResourceSummary) AddToResourceSummary(crn ClusterResourceNode) {
 		found := false
 		// traverse linkedlist to add quantity of recourse modeling
 		for element := root.Front(); element != nil; element = element.Next() {
-			switch clusterResourceNodeComparator(element.Value, crn) {
+			switch rs.clusterResourceNodeComparator(element.Value, crn) {
 			case 0:
 				{
 					tmpCrn := element.Value.(ClusterResourceNode)
-					safeChangeNum(&tmpCrn.quantity, crn.quantity)
```
RainbowMango, line 184: @halfrost Can you remind me why we need the lock when adding the resource quantity to the `element`?

jwcesign: +1

RainbowMango (approved):
/lgtm
/approve

I think this PR resolved a functionality issue and a potential panic.

pkg/modeling/modeling.go
halfrost: Yes, the resource model should be different for each cluster. But the original requirement seems to be that this modeling is to build a large resource model. Karmada manages the resource models of all clusters in a unified way. For this reason, it is taken out as a global variable.

pkg/modeling/modeling.go
halfrost: This function `AddToResourceSummary` will be called concurrently. If no locking is performed, the eventual consistency of num will be incorrect.

halfrost: @jwcesign  I think the point where you're confused comes from whether it should be made a local variable. It is a global variable at first, so it needs to be locked here.

jwcesign: Hi, @halfrost Thanks for your clarification!

So, this PR is reasonable and correct?

halfrost: Hi @jwcesign ,

Your changes are correct; there are no issues. With your modifications, each cluster now has its own modeling information.

There are no global variables left. In my previous PR, I didn't turn `defaultModelSorting` into a member variable because I was under the impression that we needed a broader global modeling context.

So this PR is prefect now.

halfrost:
@RainbowMango These changes are both correct and reasonable. I've reviewed it carefully. 👍🏻

pkg/modeling/modeling.go
RainbowMango: Thank you @halfrost for the clarification. 
I believe it reasonable for each cluster to have its resource models.

halfrost: OK, this pr makes sense. Thanks for @jwcesign work. 👍🏻

PR #4180, opened 2023-10-26
[webhook] cleanup deprecated field of TLSMinVersion, Update to TLSOpts 

liangyuanpeng:
**What type of PR is this?**

/kind cleanup

Cleanup deprecated field of TLSMinVersion for webhook

Fixes #

```release-note

```


cmd/webhook/app/options/options.go
```diff
@@ -70,7 +70,7 @@ func (o *Options) AddFlags(flags *pflag.FlagSet) {
 		"The directory that contains the server key and certificate.")
 	flags.StringVar(&o.CertName, "tls-cert-file-name", "tls.crt", "The name of server certificate.")
 	flags.StringVar(&o.KeyName, "tls-private-key-file-name", "tls.key", "The name of server key.")
-	flags.StringVar(&o.TLSMinVersion, "tls-min-version", defaultTLSMinVersion, "Minimum TLS version supported. Possible values: 1.0, 1.1, 1.2, 1.3.")
+	flags.Float32Var(&o.TLSMinVersion, "tls-min-version", defaultTLSMinVersion, "Minimum TLS version supported. Possible values: 1.0, 1.1, 1.2, 1.3.")
```
RainbowMango, line 73: I'm not sure if changing the flag type would break the backward compatibility, what do you think?

cmd/webhook/app/webhook.go
```diff
@@ -101,12 +102,16 @@ func Run(ctx context.Context, opts *options.Options) error {
...
-			KeyName:       opts.KeyName,
-			TLSMinVersion: opts.TLSMinVersion,
+			Host:     opts.BindAddress,
+			Port:     opts.SecurePort,
+			CertDir:  opts.CertDir,
+			CertName: opts.CertName,
+			KeyName:  opts.KeyName,
+			TLSOpts: []func(config *tls.Config){
+				func(config *tls.Config) {
+					config.MinVersion = uint16(opts.TLSMinVersion)
```
RainbowMango, line 112: It probably not work, as the acceptable value for the tls.MinVersion are:
```
	VersionTLS10 = 0x0301
	VersionTLS11 = 0x0302
	VersionTLS12 = 0x0303
	VersionTLS13 = 0x0304
```
See https://github.com/golang/go/blob/555af99bcc6b7515ebad5a3c947f3e4ab147fd29/src/crypto/tls/common.go#L29-L32.

RainbowMango:
The deprecated file has been removed from controller-runtime v0.16. so we have to migrate `TLSMinVersion` to TLSOpts before bumping the dependency.

@yanfeng1992 @zhzhuang-zju Could you please take a look?

zhzhuang-zju:
> The deprecated file has been removed from controller-runtime v0.16. so we have to migrate `TLSMinVersion` to TLSOpts before bumping the dependency.
> 
> @yanfeng1992 @zhzhuang-zju Could you please take a look?



> The deprecated file has been removed from controller-runtime v0.16. so we have to migrate `TLSMinVersion` to TLSOpts before bumping the dependency.
> 
> @yanfeng1992 @zhzhuang-zju Could you please take a look?

/assign

RainbowMango:
Just send another PR(#4486) to bump controller-runtime, which provide a solution, see:
https://github.com/karmada-io/karmada/pull/4486/files#diff-f2c165604578f33903ee121fdc435684d682ccc4b0801efaad7243d5b3a44aa0R127-R143

cmd/webhook/app/webhook.go
yanfeng1992: image

It is recommended to add a parameter conversion method  @RainbowMango @liangyuanpeng

RainbowMango: +1 
on this approach.

liangyuanpeng:
I'm not working on this so I'll close this PR. Please feel free to pick it up.

closing..

cmd/webhook/app/webhook.go
zhzhuang-zju: agree

RainbowMango:
My idea is to move this forward first, then the #4486.

liangyuanpeng:
After being reminded by @RainbowMango that I can continue to take care of him. Thanks to the caring maintainers.

PR updated with no breaking changes to parameters,

@RainbowMango @zhzhuang-zju @yanfeng1992  PTAL,Thanks.

cmd/webhook/app/webhook.go
```diff
@@ -117,12 +118,30 @@ func Run(ctx context.Context, opts *options.Options) error {
...
+					case "1.0":
+						config.MinVersion = tls.VersionTLS10
+					case "1.1":
+						config.MinVersion = tls.VersionTLS11
+					case "1.2":
+						config.MinVersion = tls.VersionTLS12
+					case "1.3":
+						config.MinVersion = tls.VersionTLS13
+					default:
+						klog.Warningf("Have not match version for %s, using default version of %d", opts.TLSMinVersion, config.MinVersion)
```
liangyuanpeng, line 141: This may not achieve the desired effect. VersionTLS13 will be printed as 772 instead of 1.3. No more elegant method was found, so it was deleted.

cmd/webhook/app/webhook.go
```diff
@@ -117,12 +118,28 @@ func Run(ctx context.Context, opts *options.Options) error {
...
-			TLSMinVersion: opts.TLSMinVersion,
+			Host:     opts.BindAddress,
+			Port:     opts.SecurePort,
+			CertDir:  opts.CertDir,
+			CertName: opts.CertName,
+			KeyName:  opts.KeyName,
+			TLSOpts: []func(*tls.Config){
+				func(config *tls.Config) {
+					// Just transform the valid options as opts.TLSMinVersion
+					// can only accept "1.0", "1.1", "1.2", "1.3" and has default
```
yanfeng1992, line 129: Should a version check be added to validation.go? Other parts lgtm @liangyuanpeng @RainbowMango @zhzhuang-zju

RainbowMango: Good point. Yes, we can. 
But out of the scope of this PR. Please send another PR if you like.

Issue #4182, opened 2023-10-26
feat: support cluster certificate validation instead of configuring InsecureSkipVerify


jwcesign:
**What would you like to be added**:
feat: support cluster certificate validation instead of configuring InsecureSkipVerify

**Why is this needed**:
Now, when karmada-aggregated-apiserver access member clusters' apiserver, it doesn't check the server certificates, which could cause cluster server Counterfeiting:
https://github.com/karmada-io/karmada/blob/3267155766f69083d211c4bc804e8bda4b18fd4d/pkg/util/proxy/proxy.go#L79

So it's better to support cluster certificate validation instead of configuring InsecureSkipVerify

PR #4183, opened 2023-10-26
feat: Add ca to check the validation of the member clusters' server certificate

jwcesign:
**What type of PR is this?**
/kind feature

Fixes https://github.com/karmada-io/karmada/issues/4182

If I change the following line to false:
https://github.com/karmada-io/karmada/blob/3267155766f69083d211c4bc804e8bda4b18fd4d/pkg/util/proxy/proxy.go#L79

It will have the following error in `karmada-aggregated-apiserver`:
01d9281bfb18dfd30d78ab197884c43

And karmadactl couldn't list the pods in member clusters:
d2a7ad79b3a939f7263191265be2176

With this PR, everything works fine

```release-note
`karmada-aggregated-apiserver`: Add ca to check the validation of the member clusters' server certificate.
```

### DIFF
--- a/pkg/util/proxy/proxy.go
+++ b/pkg/util/proxy/proxy.go
@@ -3,6 +3,7 @@ package proxy
 import (
 	"context"
 	"crypto/tls"
+	"crypto/x509"
 	"errors"
 	"fmt"
 	"net/http"
@@ -25,9 +26,16 @@ import (
 	clusterapis "github.com/karmada-io/karmada/pkg/apis/cluster"
 )
 
+type SecretGetterFunc func(context.Context, string, string) (*corev1.Secret, error)
+
 // ConnectCluster returns a handler for proxy cluster.
-func ConnectCluster(ctx context.Context, cluster *clusterapis.Cluster, proxyPath string, secretGetter func(context.Context, string, string) (*corev1.Secret, error), responder registryrest.Responder) (http.Handler, error) {
-	location, proxyTransport, err := Location(cluster)
+func ConnectCluster(ctx context.Context, cluster *clusterapis.Cluster, proxyPath string, secretGetter SecretGetterFunc, responder registryrest.Responder) (http.Handler, error) {
+	tlsConfig, err := GetTlsConfigForCluster(ctx, cluster, secretGetter)
+	if err != nil {
+		return nil, err
+	}
+
+	location, proxyTransport, err := Location(cluster, tlsConfig)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


zishen (commented):
Good for validating  the member clusters' server certificate

RainbowMango:
The unit test is failing, @jwcesign :
https://github.com/karmada-io/karmada/actions/runs/6651924401/job/18075272929?pr=4183#step:4:5247

pkg/util/proxy/proxy.go
```diff
@@ -94,14 +102,39 @@ func NewThrottledUpgradeAwareProxyHandler(location *url.URL, transport http.Roun
 	return proxy.NewUpgradeAwareHandler(location, transport, wrapTransport, upgradeRequired, proxy.NewErrorResponder(responder))
 }
 
+func GetTlsConfigForCluster(ctx context.Context, cluster *clusterapis.Cluster, secretGetter SecretGetterFunc) (*tls.Config, error) {
+	if cluster.Spec.SecretRef == nil {
```
RainbowMango, line 106: Can you remind me when the secret would not exist?

pkg/util/proxy/proxy.go
```diff
@@ -94,14 +102,39 @@ func NewThrottledUpgradeAwareProxyHandler(location *url.URL, transport http.Roun
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get CA bundle for cluster %s: %v", cluster.Name, err)
+	}
+
+	caCertPool := x509.NewCertPool()
+	caCertPool.AppendCertsFromPEM([]byte(caBundle))
+	return &tls.Config{
+		RootCAs:            caCertPool,
+		MinVersion:         tls.VersionTLS13,
+		InsecureSkipVerify: cluster.Spec.InsecureSkipTLSVerification, //nolint:gosec
```
RainbowMango, line 126: Can we remove the `nolint:gosec` now?

Same as above.

pkg/util/proxy/proxy.go
jwcesign: From my point of view, it couldn't be empty. But I saw it's an optional field in the struct comment.

pkg/util/proxy/proxy.go
jwcesign: If delete it, there is an error: `cluster.Spec.InsecureSkipTLSVerification may be true`.

But it's allowed to be true(if users want to do it).

pkg/util/proxy/proxy.go
RainbowMango: For `Pull` mode cluster, the `karmada-agent` may not set the secret, see the usage from `karmada-agent`:
```bash
      --report-secrets strings                                                                                                                                                                     
                The secrets that are allowed to be reported to the Karmada control plane during registering. Valid values are 'KubeCredentials', 'KubeImpersonator' and 'None'. e.g
                'KubeCredentials,KubeImpersonator' or 'None'. (default [KubeCredentials,KubeImpersonator])

```

pkg/util/proxy/proxy.go
RainbowMango: Before:
```
pkg/util/proxy/proxy.go:109:24: G402: TLS InsecureSkipVerify set true. (gosec)
			InsecureSkipVerify: true,
```
After:
```
pkg/util/proxy/proxy.go:109:24: G402: TLS InsecureSkipVerify may be true. (gosec)
			InsecureSkipVerify: cluster.Spec.InsecureSkipTLSVerification,
```
we might need a comment to explain it.

pkg/util/proxy/proxy.go
```diff
@@ -94,14 +102,42 @@ func NewThrottledUpgradeAwareProxyHandler(location *url.URL, transport http.Roun
 	return proxy.NewUpgradeAwareHandler(location, transport, wrapTransport, upgradeRequired, proxy.NewErrorResponder(responder))
 }
 
+func GetTlsConfigForCluster(ctx context.Context, cluster *clusterapis.Cluster, secretGetter SecretGetterFunc) (*tls.Config, error) {
+	if cluster.Spec.SecretRef == nil {
+		return &tls.Config{
+			MinVersion: tls.VersionTLS13,
+			// 1.If no comment 'nolint:gosec', there will be golint error 'TLS InsecureSkipVerify may be true'.
+			// 2.When the cluster CustomResource is not configured with CA(For Pull mode cluster, the karmada-agent may not set the CA),
+			//   InsecureSkipTLSVerification of cluster CustomResource must be true to ensure karmada-aggregated-apiserver working fine.
```
RainbowMango, line 111:
```suggestion

			// Ignore false positive warning: "TLS InsecureSkipVerify may be true. (gosec)"
			// Whether to skip server certificate verification depends on the
			// configuration(.spec.insecureSkipTLSVerification, defaults to false) in a Cluster object.
```

pkg/util/proxy/proxy.go
RainbowMango:
```suggestion
	// The secret is optional for a pull-mode cluster, if no secret just returns a config with root CA unset.
	if cluster.Spec.SecretRef == nil {
```
Maybe we can have a comment here.

RainbowMango (approved):
/lgtm
/approve

PR #4187, opened 2023-10-27
Add proposal for Cost Optimized Strategy

### LINKED ISSUE #3714: Cost aware scheduling support

Cost aware scheduling with the respective scheduling-estimator, as per [#3447](https://github.com/karmada-io/karmada/issues/3447#issuecomment-1607741534).

**Why is this needed**:

Karmada could improve community adoption greatly if it would be possible to perform cost optimizations for certain types of workloads ([DRA integration](https://gitlab.com/nvidia/cloud-native/k8s-dra-driver) for GPU / SRIOV devices cost awareness might be required). So, it could be possible to switch from one cluster to another, when the specific ResourceClass became cheaper there, for cost optimization.

luckyrayy:
**What type of PR is this?**

/kind design

I want to introduce a new feature to achieve cost optimized during scheduling.

Fixes #3714 

```release-note
This is a design doc, so currently there is no change in user-facing.
```

### DIFF
--- a/docs/proposals/scheduling/cost-optimized-strategy/README.md
+++ b/docs/proposals/scheduling/cost-optimized-strategy/README.md
@@ -19,7 +19,9 @@ At present, there's no way to optimize costs by having the Karmada cluster sched
 - Enable the Karmada scheduler to be aware of the cost of member clusters.
 - Introduce a cost-optimized scheduling mode, where the scheduling takes cluster cost factors into consideration in this mode.
 ### Non-Goals
-When the workload has already been deployed, if the cluster prices change, the scheduler will reschedule based on the new prices.
+- Use public cloud pricing API to calculate cluster's price by Karmada itself.
+- When the workload has already been deployed, if the cluster prices change, the scheduler will reschedule based on the new prices.
+
 ## Proposal
 
 ### User Stories (Optional)
@@ -52,9 +54,8 @@ type AccurateSchedulerEstimatorServer struct {
 ```
 ### Get Member Cluster Price
 We will introduce a method for the scheduler to retrieve the cluster pricing:
-- The Estimator Server will poll the cluster pricing API every hour and update the `AccurateSchedulerEstimatorServer.price`.
-    - Implement common public cloud service price query API calls.
-    - For other private clusters, users are required to provide an HTTP interface following specific guidelines.
+- The Estimator Server will pull the cluster pricing API every hour and update the `AccurateSchedulerEstimatorServer.price`.
+    - Users are required to provide an HTTP interface following specific guidelines.
 ### New Strategy
 We will introduce a new scheduling strategy named `CostOptimizedStrategy`. This strategy will have a corresponding function called `assignByCostOptimizedStrategy`, with the following steps:
 - Calculate the available resource amount for each cluster.
--- a/docs/proposals/scheduling/cost-optimized-strategy/README.md
+++ b/docs/proposals/scheduling/cost-optimized-strategy/README.md
@@ -1,0 +1,78 @@
+---
+title: Cost Optimized Strategy Support
+authors:
+- "@realnumber666"
+reviewers:
+  - TBD
+approvers:
+  - TBD
+creation-date: 2023-10-26
+---
+## Summary
+
+Based on the current cross-cluster scheduling, which takes into account factors like remaining resource capacity, users are increasingly concerned about costs. If workloads can be preferentially scheduled to clusters with lower prices, it can significantly reduce user expenses. However, current schedulers cannot make decisions based on the price of a cluster.
+
+This KEP aims to add an interface to retrieve cluster prices, and during scheduling, support prioritizing workloads to be placed on clusters with lower costs.
+## Motivation
+At present, there's no way to optimize costs by having the Karmada cluster scheduler make decisions based on cluster prices. Therefore, we hope to enable the Karmada scheduler to be aware of cluster prices, and when users choose a cost-optimized scheduling strategy, prioritize placing workloads on the cluster with the lowest price.
+### Goals
+- Enable the Karmada scheduler to be aware of the cost of member clusters.
+- Introduce a cost-optimized scheduling mode, where the scheduling takes cluster cost factors into consideration in this mode.
+### Non-Goals
+When the workload has already been deployed, if the cluster prices change, the scheduler will reschedule based on the new prices.
+## Proposal
+
+### User Stories (Optional)
+
+As a cost-sensitive user, I currently have 2 clusters with different prices. I hope my workload can be scheduled in the most cost-effective way.
+- Workload: requests 12 replicas, each replica requests 8 cpu.
+- Cluster A: 20 nodes and each has 8 cpu remaining. The price is ¥0.5/h.
+- Cluster B: 10 nodes and each has 8 cpu remaining. The price is ¥0.3/h.
+
+In the end, 10 replicas of the workload will be created in Cluster B, and 2 replicas will be created in Cluster A. Because the price of Cluster B is lower than that of Cluster A, scheduling the workload to Cluster B as much as possible can make it more cost-effective.
+
+## Design Details
+### API Change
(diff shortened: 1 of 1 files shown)


docs/proposals/scheduling/cost-optimized-strategy/README.md
```diff
@@ -0,0 +1,78 @@
...
+    informerManager genericmanager.SingleClusterInformerManager  
+    parallelizer    parallelize.Parallelizer  
+    price           float32
+  
+    Cache schedcache.Cache  
+}
+```
+### Get Member Cluster Price
+We will introduce a method for the scheduler to retrieve the cluster pricing:
+- The Estimator Server will poll the cluster pricing API every hour and update the `AccurateSchedulerEstimatorServer.price`.
```
jwcesign, line 55: How can we calculate the price of the clusters? I am sure there is no such kind of pricing API for the public cloud.
It only has the price of node($x/hour and reserved instance, like $x for one year)

jwcesign:
Hi @realnumber666 , thanks for your great proposal, can u give more details about the following question?
1. What's the scenario, like ML which needs to care about the GPU price when there are multi-cloud(like https://github.com/skypilot-org/skypilot)?
1. The core concept is to schedule the workload to cheaper clusters, but there are different price nodes, how can we make sure it's cheaper? Crane is needed to query the price in details.
1. This scenario exists only for the public cloud, right? Does it work for on-premise clusters or mixed-type clusters?

docs/proposals/scheduling/cost-optimized-strategy/README.md
RainbowMango: Maybe this proposal just focuses on how to consume the price and take the price into account to make schedule decisions.  I guess we can assume users can define the price for their cluster.

realnumber666:
> Hi @realnumber666 , thanks for your great proposal, can u give more details about the following question.
> 
> 1. What's the scenario, like ML which needs to care about the GPU price when there are multi-cloud(for example https://github.com/skypilot-org/skypilot)?
> 2. The core concept is to schedule the workload to cheaper clusters, but there are different price nodes, how can we make sure it's cheaper? Crane is needed to query the price in details.
> 3. This scenario exists only for the public cloud, right? Does it work for on-premise clusters or mixed-type clusters?

Thanks for your review and these good questions, and sorry for my late reply. Following is my answer, please ask anything if you have other concern!

1. Based on the project I used to involve in, I think that large-scale ML training is one of the scenario. For instance, one of our large-scale experiments needed many CPU cores, which were scheduled across multiple clusters. These clusters varied in resource type. Some were idle resources spared by other internal business in on-premise clusters, thus offered to us for free. Others were Cloud Virtual Machines (CVM) from public cloud services, which were relatively cheaper, and then there were highly stable resources, which were more expensive. In this scenario, if we could prioritize scheduling to lower-cost clusters in addition to considering remaining resource capacity, it would save a significant amount of cost considering our vast resource needs and prolonged usage. Therefore, I believe in large-scale training, such as LLM, Reinforcement Learning experiments, this scenario is quite common.
2. That's a good question, and perhaps we can, as @RainbowMango suggested, start by assuming that users can set prices for their clusters. In my past experience, nodes with similar prices were often grouped in the same cluster, meaning a cluster had a unified price known to our developers. This aligns with RainbowMango's suggestion and can be applied in this scheme.
3. No. As I mentioned in previous content, we used both on-premise clusters and public cloud, making it a mixed-type clusters environment. These clusters had different prices, therefore making them suitable for this Cost Optimized Strategy.

docs/proposals/scheduling/cost-optimized-strategy/README.md
realnumber666: Thanks for your discussion and I think @RainbowMango's suggestion is a good idea for this proposal, it will make this proposal's target more clear. And I have updated my proposal according to this suggestion.

docs/proposals/scheduling/cost-optimized-strategy/README.md
```diff
@@ -0,0 +1,79 @@
...
+    informerFactory informers.SharedInformerFactory  
+    nodeLister      listv1.NodeLister  
+    replicaLister   *replica.ListerWrapper  
+    informerManager genericmanager.SingleClusterInformerManager  
+    parallelizer    parallelize.Parallelizer  
+    price           float32
+  
+    Cache schedcache.Cache  
+}
+```
```
jwcesign, line 54: How about just implementing a ScorePlugin
https://github.com/karmada-io/karmada/blob/e6a75afdc388b629e53c46a2a27402017aebfb63/pkg/scheduler/core/generic_scheduler.go#L130

The low-price clusters will have higher scores, and they will be selected first.

docs/proposals/scheduling/cost-optimized-strategy/README.md
```diff
@@ -0,0 +1,79 @@
...
+### Get Member Cluster Price
+We will introduce a method for the scheduler to retrieve the cluster pricing:
+- The Estimator Server will pull the cluster pricing API every hour and update the `AccurateSchedulerEstimatorServer.price`.
+    - Users are required to provide an HTTP interface following specific guidelines.
+### New Strategy
+We will introduce a new scheduling strategy named `CostOptimizedStrategy`. This strategy will have a corresponding function called `assignByCostOptimizedStrategy`, with the following steps:
+- Calculate the available resource amount for each cluster.
```
jwcesign, line 61: This strategy may have conflicts with the existing strategy(Divided and Duplicated)

realnumber666: Currently, we have 
- `ReplicaSchedulingType`
    - `Duplicated`
    - `Divided`
- `ReplicaDivisionPreference`
    - `Aggregated`
    - `Weighted`
- `Strategy`
    - `DuplicatedStrategy`
    - `AggregatedStrategy`
    - `DynamicWeightStrategy`
    - `StaticWeightStrategy`

Their mapping relationships are like here https://github.com/karmada-io/karmada/blob/dd9f54cbe206da415adb9ecc34f7b217436e352c/pkg/scheduler/core/assignment.go#L74-L87

My idea is to add a new `ReplicaDivisionPreference` named `CostOptimized` and a new strategy named `CostOptimizedStrategy`. When the `ReplicaSchedulingType` is `Divided` and the `ReplicaDivisionPreference` is `CostOptimized`, the strategy will be `CostOptimizedStrategy`. Under this design, this new strategy will not conflict with the existing `Divided` and `Duplicated` strategies. Could you please confirm if my understanding is correct?

realnumber666:
Hi @jwcesign, may I inquire, if we complete a new feature and need to conduct tests, should we set up a multi-cluster environment for testing by ourselves, or does Karmada provide a testing environment that we can use?

realnumber666:
Hi @jwcesign, 
I have addressed the concerns mentioned in your previous comments and updated my proposal accordingly. Please review the revised proposal at your convenience.

jwcesign:
> Hi @jwcesign, I have addressed the concerns mentioned in your previous comments and updated my proposal accordingly. Please review the revised proposal at your convenience.

Hi, sorry for delayed response, I will be back soon.
Thanks for your contribution!!!

docs/proposals/scheduling/cost-optimized-strategy/README.md
jwcesign: Correct. So this strategy only exists for Divided? Right?

jwcesign:
> sign, may I inquire, if we complete a new feature and need to conduct tests, should we set up a multi-cluster environment for testing by ourselves, or does Karmada provide a testing environment that we can use?

You can launch the testing environment with `hack/local-up-karmada.sh`

jwcesign:
Hi, @realnumber666 

I still believe that our cost estimation based on the cluster dimension is too rough. Making decisions based on this could potentially lead to higher costs in the end. 

 Could we analyze factors affecting costs, and then determine our fetch cost API? Factors such as CPU/memory requests, scheduling node costs, number of CPUs in a scheduling node, and amount of memory in a scheduling node.

PR #4188, opened 2023-10-27
fix: helm podDisruptionBudget for controller-manager

a7i:
**What type of PR is this?**
/kind bug

before
```
---
# Source: karmada/templates/karmada-controller-manager.yaml
apiVersion: policy/v1
kind: PodDisruptionBudget
metadata:
  name: karmada-controller-manager
  namespace: karmada-system
  labels:

    app: karmada-controller-manager
spec:
  selector:
    matchLabels:

      app: karmada-controller-manager

  maxUnavailable: 20%---  # <--- this is the issue
---
```

after:
```yaml
---
# Source: karmada/templates/karmada-controller-manager.yaml
apiVersion: policy/v1
kind: PodDisruptionBudget
metadata:
  name: karmada-controller-manager
  namespace: karmada-system
  labels:

    app: karmada-controller-manager
spec:
  selector:
    matchLabels:

      app: karmada-controller-manager

  maxUnavailable: 20%
---
```

N/A

```release-note
fixed helm render of podDisruptionBudget in controller-manager
```

### DIFF
--- a/charts/karmada/templates/karmada-controller-manager.yaml
+++ b/charts/karmada/templates/karmada-controller-manager.yaml
@@ -92,5 +92,4 @@ spec:
   {{ toYaml .Values.controllerManager.podDisruptionBudget | nindent 2 }}
 {{- end -}}
 
----
 {{- end }}


charts/karmada/templates/karmada-controller-manager.yaml
```diff
@@ -92,5 +92,4 @@ spec:
   {{ toYaml .Values.controllerManager.podDisruptionBudget | nindent 2 }}
 {{- end -}}
 
----
 {{- end }}
```
chaosi-zju, line 95: hello, do you known what is the difference between `{{- end -}}` and `{{- end }}`, when shall we use `{{- end -}}` rather than `{{- end }}`?

chaosi-zju: actually, if I just change `#line 93` from `{{- end -}}` to `{{- end }}`, it can also fix this problem.

I am not very familiar with helm syntax and fell it's marvelous.

However, to keep the same with other files, I prefer your means.

a7i: hi @chaosi-zju 
>https://pkg.go.dev/text/template#hdr-Text_and_spaces
`{{-`  "all trailing white space is trimmed from the immediately preceding text."
`-}}` "all leading white space is trimmed from the immediately following text."

Agreed that we should be consistent on trailing/leading whitespace trimming. Generally I prefer to have each block control the YAML separator `---`. Because it knows whether it's being rendered or not.

i.e.
```yaml
{{ if .Values.controllerManager.podDisruptionBudget }}
---
apiVersion: policy/v1
kind: PodDisruptionBudget
{{ end }}
```

above is preferred over below:
```yaml
{{ if .Values.controllerManager.podDisruptionBudget }}
---
apiVersion: policy/v1
kind: PodDisruptionBudget
---
{{ end }}
```

That's because if we add a new resource that is also based on a condition then the extra `---` could cause issues

```yaml
apiVersion: apps/v1
kind: Deployment

{{ if .Values.controllerManager.podDisruptionBudget }}
---
apiVersion: policy/v1
kind: PodDisruptionBudget
---
{{ end }}

{{ if .Values.controllerManager.newResourceForExample }}
apiVersion: example/v1
kind: SomeResource
{{ end }}
```

That's  because if `newResourceForExample` is enabled but `podDisruptionBudget` is not, then you end up with:
```yaml
apiVersion: apps/v1
kind: Deployment
apiVersion: example/v1
kind: SomeResource
```
Where there is no separator. Hence, starting YAML separator should belong to the if block

chaosi-zju: > https://pkg.go.dev/text/template#hdr-Text_and_spaces
> {{- "all trailing white space is trimmed from the immediately preceding text."
-}} "all leading white space is trimmed from the immediately following text."

concise and easy to understand, very thanks

> Generally I prefer to have each block control the YAML separator ---

You are absolutely right, I can't agree more！👍

RainbowMango (approved):
/approve

PR #4201, opened 2023-11-01
Update chart-testing-action to v2.6.0 and cosigin-installer to v3.1.1

liangyuanpeng:
**What type of PR is this?**

/kind cleanup

helm chart testing action is broken, https://github.com/karmada-io/karmada/actions/runs/6706548301/job/18223260799

xref: 
- https://github.com/helm/chart-testing-action/issues/132
- https://github.com/helm/chart-testing-action/issues/133

I have try to update to v2.5.0 and it's not working. https://github.com/liangyuanpeng/karmada/actions/runs/6714930957 

And v2.6.0 can fix it.

Since we are using cosign separately, it is better to follow the notification to upgrade the installer version.

So this PR have two change:
- Update chart-testing-action to v2.6.0
- Update cosigin-installer to v3.1.1   see https://blog.sigstore.dev/cosign-releases-bucket-deprecation/

Fixes #

```release-note
NONE
```

### DIFF
--- a/.github/workflows/dockerhub-latest-image.yml
+++ b/.github/workflows/dockerhub-latest-image.yml
@@ -40,7 +40,7 @@ jobs:
         with:
           go-version: 1.20.10
       - name: Install Cosign
-        uses: sigstore/cosign-installer@v3.0.3
+        uses: sigstore/cosign-installer@v3.1.1
         with:
           cosign-release: 'v1.13.1'
       - name: install QEMU
--- a/.github/workflows/dockerhub-released-image.yml
+++ b/.github/workflows/dockerhub-released-image.yml
@@ -36,7 +36,7 @@ jobs:
         with:
           go-version: 1.20.10
       - name: Install Cosign
-        uses: sigstore/cosign-installer@v3.0.3
+        uses: sigstore/cosign-installer@v3.1.1
         with:
           cosign-release: 'v1.13.1'
       - name: install QEMU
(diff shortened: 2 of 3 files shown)


RainbowMango (commented):
/assign
Seems this might be a blocker for the other's PR.
/priority important-soon

RainbowMango:
Seems this is due to [Sigstore no longer publishing cosign releases to GCS bucket](https://blog.sigstore.dev/cosign-releases-bucket-deprecation/).

There is an instruction for updating GitHub Actions if people using `cosign-installer`.
> If you are using cosign-installer, our GitHub Action that installs Cosign, you may need to update the action only if you are pinned by hash to a version earlier than v3.1.0. The following example shows how to update the action:

Do we need to update them?

liangyuanpeng:
> Do we need to update them?

Thanks for remind and i have tested it with `cosigin-installer@v3.0.3` and it's working for now,But I think it's better to follow the notification and upgrade the install version.

Also tested it with `cosigin-installer@v3.1.1` and it's working.

So we can update the installer version and keep the same cosign version.

RainbowMango (approved):
/lgtm
/approve

Actually we don't need to bump `sigstore/cosign-installer` as the action downloads the `sigstore` binary directly from GitHub(not the deprecated Google Bucket):
See the code for more details:
https://github.com/sigstore/cosign-installer/blob/main/action.yml#L184-L185

But it's not harm to bump it to a higher version.

liangyuanpeng:
Luckly, it is since 3.0.3 the cosign-installer download cosign from GitHub instead of GCS and karmada is using this version.

RainbowMango:
I don't get why it is the v3.1.0. I think any hash version earlier than v3.0.3 also downloads from GCS.

PR #4205, opened 2023-11-02
publish karmada operator helm chart archive

wrhight:
**What type of PR is this?**
/kind feature

This PR adds the operator helm chart artifacts to the release, so in a future PR the charts/index.yaml file can be updated to point to these artifacts. Users will then be able to pull the operator chart. This follows the same flow as the existing karmada helm chart.

Currently, the release workflow only pushes the chart to the docker registry, but it utilizes the same name and tag as the image artifact, which results in [this error](https://github.com/helm/helm/issues/10968)

no existing issue to reference

```release-note
NONE
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -76,6 +76,8 @@ jobs:
         files: |
           _output/charts/karmada-chart-${{ github.ref_name }}.tgz
           _output/charts/karmada-chart-${{ github.ref_name }}.tgz.sha256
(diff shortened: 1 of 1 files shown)


RainbowMango:
/assign @chaosi-zju 

Can you help with it?

RainbowMango (approved):
/lgtm
/approve

Thanks. It makes sense to upload the `karmada operator` charts to releases.

Issue #4208, opened 2023-11-04
Random reaction time for Application Failover 


lorenzogio97:
Hi,  
**What happened**:
i was trying to test the Application failover feature of Karmada, using a simple nginx deployment and a ClusterPropagationPolicy. I'm testing the exact same scenario reported here (https://karmada.io/docs/userguide/failover/application-failover/), but when I cordon the cluster where pods are currenty scheduled and delete them, karmada continue to try to re-schedule them in the same cluster, for more than the  tolerationSeconds that are expected. 
I made some tests with different tolerationSeconds value and I experience the following time to see the pods scheduled on the other cluster:
- 5s: 2 independent tests, 19 min and 16 min
- 8s: 12 min
- 15s: 29 min
From what I experienced, the time is not deterministic and i didn't see any relationship with tolerantSeconds value. 

**What you expected to happen**:
Pods scheduled (and then deleted, like it's done in doc) on the cordoned cluster should be re-scheduled on the other cluster after tolerationSeconds time value. 

**How to reproduce it (as minimally and precisely as possible)**:
- Deploy on karmada control plane this Deployment and ClusterPropagationPolicy (of course, change cluster names): 
[deployment.txt](https://github.com/karmada-io/karmada/files/13257280/deployment.txt)
[propagationpolicy.txt](https://github.com/karmada-io/karmada/files/13257281/propagationpolicy.txt)
- Cordon the primary cluster
- Delete pod on that cluster
- Wait that pods are re-scheduled on the backup cluster

**Anything else we need to know?**:
-

**Environment**:
- Karmada version: Karmada v1.7.0
- kubectl-karmada or karmadactl version (the result of `kubectl-karmada version` or `karmadactl version`): - 
- Others: The application has been deployed on Karmada control plane using ArgoCD

jwcesign:
Hi @lorenzogio97 
I tried with following yaml with `hack/local-up-karmada.sh`:
```yaml
apiVersion: policy.karmada.io/v1alpha1
kind: ClusterPropagationPolicy
metadata:
  name: nginx-propagation
spec:
  propagateDeps: true
  failover:
    application:
      decisionConditions:
        tolerationSeconds: 5
      purgeMode: Immediately
  resourceSelectors:
    - apiVersion: apps/v1
      kind: Deployment
      name: nginx
  placement:
    clusterAffinities:
    - affinityName: primary-clusters
      clusterNames:
        - member1
    - affinityName: backup-clusters
      clusterNames:
[...]

PR #4215, opened 2023-11-09
fix: do not reschedules the workload to the same unhealthy cluster when application failover enabled

jwcesign:
**What type of PR is this?**
/kind bug
/kind api-change

1. If `purgeMode: Immediately` is used in CPP/PP, the application-failover controller will not add the EvictionTask to (RB/CRB).spec.evictionTasks. This will result in the scheduler rescheduling to the eviction clusters.
2. If `purgeMode: Gracefully` is used in CPP/PP, but the `gracePeriodSeconds` is too short and the scheduler has a large number of items to handle, after the grace period passes, the graceful eviction controller will remove it from `(RB/CRB).spec.evictionTasks`, and then the scheduler will reschedule it to the eviction clusters.

Fixes https://github.com/karmada-io/karmada/issues/4208

Now, I refactor the logic as follows:
1. application-failover controller adds the eviction tasks to the `(RB/CRB).spec.evictionTasks`, no matter what `purgeMode`.
2. The scheduler will ignore the clusters in the  `(RB/CRB).spec.evictionTasks`, and reschedule the workload.
3. Scheduler will update `(RB/CRB).spec.evictionTasks[x].evicated` to true.
4. graceful eviction controller detects `(RB/CRB).spec.evictionTasks[x].evicated` with true, it will delete it from `(RB/CRB).spec.evictionTasks`.
5. The execution controller will delete the work once the `(RB/CRB).spec.evictionTasks[x]` exceeds the gracePeriodSeconds.

```release-note
`karmada-controller-manager`: Do not reschedule the workload to the same unhealthy cluster when application failover enabled.
```


pkg/apis/work/v1alpha2/binding_types.go
```diff
@@ -78,21 +78,20 @@ type ResourceBindingSpec struct {
...
 	// 2. The scheduler may perform a re-scheduler and probably select a substitute cluster
 	//    to take over the evicting workload(resource).
-	// 3. The graceful eviction controller takes care of the graceful eviction tasks and
+	// 3. The eviction controller(gracefully or ungracefully) takes care of the graceful eviction tasks and
 	//    performs the final removal after the workload(resource) is available on the substitute
 	//    cluster or exceed the grace termination period(defaults to 10 minutes).
 	//
 	// +optional
-	GracefulEvictionTasks []GracefulEvictionTask `json:"gracefulEvictionTasks,omitempty"`
+	EvictionTasks []EvictionTask `json:"evictionTasks,omitempty"`
```
XiShanYongYe-Chang, line 94: Will modifying this field directly affect the upgrade scenario?

pkg/apis/work/v1alpha2/binding_types.go
```diff
@@ -219,6 +218,10 @@ type GracefulEvictionTask struct {
 	// +required
 	Producer string `json:"producer"`
 
+	// Evicted indicates if the eviction is performed.
+	// +required
+	Evicted bool `json:"evicated"`
```
XiShanYongYe-Chang, line 223: Does this mean that the evicted elements will also continue to exist in the queue, and when will they be removed from the current queue?

pkg/apis/work/v1alpha2/binding_types.go
jwcesign: Yeah, it might impact the upgrade process. Should we add a new field instead of modifying the existing one?

pkg/apis/work/v1alpha2/binding_types.go
jwcesign: > Does this mean that the evicted elements will also continue to exist in the queue

Yes

> when will they be removed from the current queue?

graceful eviction controller detects (RB/CRB).spec.evictionTasks[x].evicated with true, it will delete it from (RB/CRB).spec.evictionTasks.

jwcesign:
/cc @RainbowMango 

I don't see the purpose of `Immediately`. Should we consider deprecating it?

RainbowMango:
> I don't see the purpose of `Immediately`. Should we consider deprecating it?

I added a comment on the PurgeMode Immediately with the real-world use case:
https://github.com/karmada-io/karmada/blob/6f4d69cf8b2847dbeeb2f299652920dc89b049d7/pkg/apis/policy/v1alpha1/propagation_types.go#L267-L275

RainbowMango:
/close
In favor of #5881

PR #4221, opened 2023-11-11
fix goroutine leak in watchMux

ikaven1024:
**What type of PR is this?**
/kind bug

Fixes #4212

```release-note
`karmada-search`: Fix lock race affects watch RestChan not close, causing client watch api to hang.
```

### DIFF
--- a/pkg/search/proxy/store/util.go
+++ b/pkg/search/proxy/store/util.go
@@ -195,9 +195,21 @@ func (w *watchMux) AddSource(watcher watch.Interface, decorator func(watch.Event
 
 // Start run the watcher
 func (w *watchMux) Start() {
+	wg := sync.WaitGroup{}
 	for _, source := range w.sources {
-		go w.startWatchSource(source.watcher, source.decorator)
+		source := source
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			w.startWatchSource(source.watcher, source.decorator)
+		}()
 	}
+
+	go func() {
+		// close result chan after all goroutines exit, avoiding data race.
+		defer close(w.result)
+		wg.Wait()
+	}()
 }
 
 // ResultChan implements watch.Interface
@@ -220,7 +232,6 @@ func (w *watchMux) Stop() {
 	case <-w.done:
 	default:
 		close(w.done)
-		close(w.result)
 	}
 }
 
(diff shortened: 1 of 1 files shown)


pkg/search/proxy/store/util.go
```diff
@@ -195,9 +197,67 @@ func (w *watchMux) AddSource(watcher watch.Interface, decorator func(watch.Event
...
+		for {
+			chosen, val, ok := reflect.Select(cases)
+			if !ok {
+				return
+			}
+
+			if chosen == doneCaseIndex {
+				// Received from done chan.
+				// In fact, this will never happen
+				panic(fmt.Sprintf("unexpectedly receive from done chan: %v", val.Interface()))
```
xigang, line 243: If panic is triggered, does the exception need to be caught?

pkg/search/proxy/store/util.go
```diff
@@ -195,9 +197,67 @@ func (w *watchMux) AddSource(watcher watch.Interface, decorator func(watch.Event
...
+
+	go func() {
+		defer func() {
+			for _, source := range w.sources {
+				source.watcher.Stop()
+			}
+		}()
+		defer close(w.result)
+		for {
+			chosen, val, ok := reflect.Select(cases)
```
xigang, line 235: The performance of `reflect.Select` is worse than that if `select`. if the scale of the federate cluster is too large, there may be performance issue.

pkg/search/proxy/store/util.go
```diff
@@ -246,19 +257,8 @@ func (w *watchMux) startWatchSource(source watch.Interface, decorator func(watch
 		select {
 		case <-w.done:
 			return
-		default:
+		case w.result <- copyEvent:
```
xigang, line 260: Is there a data race when multiple goroutines write w.result channel?

pkg/search/proxy/store/util.go
```diff
@@ -195,9 +195,21 @@ func (w *watchMux) AddSource(watcher watch.Interface, decorator func(watch.Event
...
+		source := source
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			w.startWatchSource(source.watcher, source.decorator)
+		}()
 	}
+
+	go func() {
+		// close result chan after all goroutines exit, avoiding data race.
```
XiShanYongYe-Chang, line 209: Just a question, why would there be data race in the previous approach?

RainbowMango: @ikaven1024 explained it at https://github.com/karmada-io/karmada/pull/4212#issuecomment-1805660860.

PR #4226, opened 2023-11-13
only update necessary part of binding status

lxtywypc:
**What type of PR is this?**  
/kind bug  

The cache may be outdated before updating status of bindings, and we should only update the aggregated status and conditions, otherwise the other parts may be overridden.  

Fixes #

```release-note

```

### DIFF
--- a/pkg/util/helper/workstatus.go
+++ b/pkg/util/helper/workstatus.go
@@ -45,8 +45,6 @@ func AggregateResourceBindingWorkStatus(
 	resourceTemplate *unstructured.Unstructured,
 	eventRecorder record.EventRecorder,
 ) error {
-	binding.GetName()
-
 	workList, err := GetWorksByBindingNamespaceName(c, binding.Namespace, binding.Name)
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


Garrybest (approved):
/lgtm

XiShanYongYe-Chang (commented):
Hi @lxtywypc, thanks

Do you think we need to cherry-pick this patch to the previous branch?

By the way, can you help fill with release note?

lxtywypc:
Hi @XiShanYongYe-Chang 

> Do you think we need to cherry-pick this patch to the previous branch?  

I think it is necessary since `gracefuleviction-controller` had been introduced.  

> By the way, can you help fill with release note?  

I have finished it.

XiShanYongYe-Chang:
> I think it is necessary since gracefuleviction-controller had been introduced.

Thanks @lxtywypc Can you help do it?

XiShanYongYe-Chang:
Hi @lxtywypc, @jwcesign has already helped to submit the cherry-pick, thanks~

Issue #4230, opened 2023-11-14
Simply code generation script


jwcesign:
Now, the script is highly duplicated:
https://github.com/karmada-io/karmada/blob/954f807aabd3e7d86e8d23c14c06aea3a652fd99/hack/update-codegen.sh#L39C1-L39C1

we can simplify it as follows:
```sh
echo "Generating with deepcopy-gen"
deepcopy-gen \
  --go-header-file hack/boilerplate/boilerplate.go.txt \
  --input-dirs=github.com/karmada-io/karmada/pkg/apis/cluster/v1alpha1 \
  --input-dirs=github.com/karmada-io/karmada/pkg/apis/cluster \
  --input-dirs=github.com/karmada-io/karmada/pkg/apis/policy/v1alpha1 \
  --input-dirs=github.com/karmada-io/karmada/pkg/apis/work/v1alpha1 \
  --input-dirs=github.com/karmada-io/karmada/pkg/apis/work/v1alpha2 \
  --input-dirs=github.com/karmada-io/karmada/pkg/apis/config/v1alpha1 \
  --output-package=github.com/karmada-io/karmada/pkg/apis \
  --input-dirs=github.com/karmada-io/karmada/pkg/apis/networking/v1alpha1 \
  --input-dirs=github.com/karmada-io/karmada/examples/customresourceinterpreter/apis/workload/v1alpha1 \
  --input-dirs=github.com/karmada-io/karmada/pkg/apis/search/v1alpha1 \
  --input-dirs=github.com/karmada-io/karmada/pkg/apis/search \
  --input-dirs=github.com/karmada-io/karmada/pkg/apis/autoscaling/v1alpha1 \
  --input-dirs=github.com/karmada-io/karmada/operator/pkg/apis/operator/v1alpha1 \
  --output-file-base=zz_generated.deepcopy
```

- [ ] Simplify deepcopy-gen
- [ ] Simplify register-gen
- [ ] Simplify client-gen
- [ ] Simplify lister-gen
- [ ] Simplify informer-gen

/assign

PR #4231, opened 2023-11-14
cleanup: Simplify deepcopy-gen code generation

jwcesign:
**What type of PR is this?**
/kind cleanup

Simplify deepcopy-gen code generation

Part of https://github.com/karmada-io/karmada/issues/4230

I removed the `--output-package` parameter as it has no effect, the generation works fine.

The CI verification now checks `pkg` and `examples/customresourceinterpreter/apis`, but does not include `operation` parts. I will add them in another PR.

```release-note

```


hack/update-codegen.sh
```diff
@@ -34,62 +34,18 @@ echo "Generating with deepcopy-gen"
...
-  --output-file-base=zz_generated.deepcopy
-deepcopy-gen \
-  --go-header-file hack/boilerplate/boilerplate.go.txt \
   --input-dirs=github.com/karmada-io/karmada/pkg/apis/networking/v1alpha1 \
-  --output-package=github.com/karmada-io/karmada/pkg/apis/networking/v1alpha1 \
-  --output-file-base=zz_generated.deepcopy
-deepcopy-gen \
-  --go-header-file hack/boilerplate/boilerplate.go.txt \
   --input-dirs=github.com/karmada-io/karmada/examples/customresourceinterpreter/apis/workload/v1alpha1 \
-  --output-package=github.com/karmada-io/karmada/examples/customresourceinterpreter/apis/workload/v1alpha1 \
```
XiShanYongYe-Chang, line 72: This item has a different output path from the other items, and may require separate handling. The operator package is also in the same situation.

XiShanYongYe-Chang: The CI doesn't echo error, maybe we need to add the verify script for `examples` and `operator`.

jwcesign: I removed the `--output-package` parameter as it has no effect.
The CI verification now checks pkg and examples/customresourceinterpreter/apis, but does not include operation parts. I will add them in another PR.

XiShanYongYe-Chang (commented):
LGTM

hack/update-codegen.sh
RainbowMango: > I removed the --output-package parameter as it has no effect.

I'm surprised, why no effect?

> The CI verification now checks pkg and examples/customresourceinterpreter/apis, but does not include operation parts.

What do you mean?

jwcesign: > I'm surprised, why no effect?

I reviewed the code of deepcopy-gen and found that it only uses `--output-base` to specify the output path.
Passing parameters.
image

Write to file.
image


related code: https://github.com/kubernetes/code-generator

The testing result:

image

jwcesign: > What do you mean?

We only verify directory `/pkg` and `/examples/customresourceinterpreter/apis`
https://github.com/karmada-io/karmada/blob/bb1d2e8f6feaf8881931129ef4abf8e573a9ccd4/hack/verify-codegen.sh#L9


But there is https://github.com/karmada-io/karmada/blob/master/operator/pkg/apis/operator/v1alpha1/zz_generated.deepcopy.go

We forgot to verify this file.

RainbowMango: Can we first fix this issue? 

I need some time to investigate why `code-generater` does not support the `--output-base`, and if we have a chance to use it in the future.

jwcesign: Knative also does not utilize this parameter: https://github.com/knative/serving/blob/936ecbdb547d896d1ef260c17edb4131cfd1d2ae/hack/update-codegen.sh#L83

jwcesign:
/cc @RainbowMango 
Any additional comments?

RainbowMango:
Not yet. Will do.

RainbowMango:
Starting from v1.30, the code generator removed the `--output-package` and required the `--output-file`.
So, we don't need this PR anymore. The adoption code is included in #5146. 
But still thanks for your effort. 
/close

PR #4237, opened 2023-11-14
Restrict “configmap/extension-apiserver-authentication” object from being distributed to member clusters

### LINKED ISSUE #4228: Restrict “configmap/extension-apiserver-authentication” object from being distributed to member clusters.
While implementing Karmada to support the distribution of resources under the "kube-system" namespace, the "configmap/extension-apiserver-authentication" object created by kube-apiserver will not be distributed to member clusters.

**Why is this needed**:
Since Karmada uses the "skipped-propagating-namespaces" configuration item to control the distribution of resources, this configuration item prevents users from distributing resources to namespaces starting with "kube-*". Users can modify this configuration item to distribute resources under the "kube-system" namespace. However, when this configuration item is modified, the "configmap/extension-apiserver-authentication" object previously created by kube-apiserver will be distributed to the member clusters. Therefore, one method should be provided to allow users to distribute resources under “kube-system“ namespace, while the "configmap/extension-apiserver-authentication" object remains under the kube-system namespace.

zhzhuang-zju:
**What type of PR is this?**
/kind feature

While implementing Karmada to support the distribution of resources under the "kube-system" namespace, the "configmap/extension-apiserver-authentication" object created by kube-apiserver should not be distributed to member clusters.
Fixes #4228

NONE

```release-note
`karmada-controller-manager`: Ignored `configmap/extension-apiserver-authentication` from propagation.
```

### DIFF
--- a/pkg/detector/detector.go
+++ b/pkg/detector/detector.go
@@ -233,16 +233,15 @@ func (d *ResourceDetector) Reconcile(key util.QueueKey) error {
 	return d.propagateResource(object, clusterWideKey)
 }
 
-// EventFilter tells if an object should be take care of.
+// EventFilter tells if an object should be taken care of.
 //
-// All objects under Kubernetes reserved namespace should be ignored:
-// - kube-*
 // All objects under Karmada reserved namespace should be ignored:
 // - karmada-system
 // - karmada-cluster
 // - karmada-es-*
 //
-// If '--skipped-propagating-namespaces' is specified, all APIs in the skipped-propagating-namespaces will be ignored.
+// If '--skipped-propagating-namespaces'(defaults to kube-.*) is specified,
+// all resources in the skipped-propagating-namespaces will be ignored.
 func (d *ResourceDetector) EventFilter(obj interface{}) bool {
 	key, err := ClusterWideKeyFunc(obj)
 	if err != nil {
@@ -277,6 +276,14 @@ func (d *ResourceDetector) EventFilter(obj interface{}) bool {
 				return false
 			}
 		}
+	}
+
+	// Prevent configmap/extension-apiserver-authentication from propagating as it is generated
+	// and managed by kube-apiserver.
+	// Refer to https://github.com/karmada-io/karmada/issues/4228 for more details.
+	if clusterWideKey.Namespace == "kube-system" && clusterWideKey.Kind == "ConfigMap" &&
+		clusterWideKey.Name == "extension-apiserver-authentication" {
+		return false
 	}
 
 	return true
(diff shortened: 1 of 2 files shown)


artifacts/deploy/karmada-controller-manager.yaml
```diff
@@ -32,6 +32,7 @@ spec:
             - --failover-eviction-timeout=30s
             - --controllers=*,hpaReplicasSyncer
             - --feature-gates=PropagationPolicyPreemption=true
+            - --skipped-propagating-namespaces=kube-.*
```
RainbowMango, line 35: Why bring this flag here?

zhzhuang-zju: The default value of flag ```--skipped-propagating-namespaces``` is ```kube-.*```, and I found one another installation method also add this flag, like 
https://github.com/karmada-io/karmada/blob/bb1d2e8f6feaf8881931129ef4abf8e573a9ccd4/cmd/controller-manager/app/options/options.go#L194-L196

zhzhuang-zju: Thanks for correcting me, I understood it wrong before. This file actually adds a default value to the controller-manager's flag ```--skipped-propagating-namespaces```, so it's no need to bring this flag here again.

zhzhuang-zju:
Verification report.
1. set controller-manager's flag ```--skipped-propagating-namespaces``` to ```test-.*```
2. install karmada by ```hack/local-up-karmada.sh```
3. deploy a configmap for comparison
```yaml                                       
apiVersion: v1
kind: ConfigMap
metadata:
  name: my-config
  namespace: kube-system
data:
  nginx.properties: |
    proxy-connect-timeout: "10s"
```
4. deploy PropagationPolicy
```yaml
kind: ClusterPropagationPolicy
metadata:
  name: example-policy
spec:
  resourceSelectors:
    - apiVersion: v1
      kind: ConfigMap
  conflictResolution: Overwrite
  placement:
    clusterAffinity:
      clusterNames:
        - member1
        - member2
```
5.  Configmap ```my-config``` has been scheduled to member1 and member2, but configmap ```extension-apiserver-authentication``` has not.
```bash
➜  karmada git:(top) ✗ ka get work -A |grep my-config
karmada-es-member1   my-config-64bc957c                True      14m
karmada-es-member2   my-config-64bc957c                True      14m
➜  karmada git:(top) ✗ ka get work -A |grep extension-apiserver-authentication
```

RainbowMango (commented):
/lgtm

Wait for a while to see if others have any concern.

PR #4245, opened 2023-11-15
ensure resource template be un-claimed when deleting policy

### LINKED ISSUE #4233: cannot unbind from propagation policy when cleanup error on resource occurs
**What happened**:
resource template cannot be unbound from propagation policy when deleting propagation policy.
image

it should be unbound.

**Environment**:
- Karmada version: all versions
- Others:

whitewindmills:
**What type of PR is this?**
/kind bug
/kind cleanup

Fixes #4233

```release-note
`karmada-controller-manager`: Fix the bug that losing the chance to un-claim resource template in case of PropagationPolicy/ClusterPropagationPolicy is removed.
```

### DIFF
--- a/pkg/detector/detector.go
+++ b/pkg/detector/detector.go
@@ -1039,18 +1039,19 @@ func (d *ResourceDetector) HandlePropagationPolicyDeletion(policyNS string, poli
 	}
 
 	for index, binding := range rbs.Items {
-		// Cleanup the labels from the reference binding so that the karmada scheduler won't reschedule the binding.
-		if err := d.CleanupResourceBindingLabels(&rbs.Items[index], policyv1alpha1.PropagationPolicyNamespaceLabel, policyv1alpha1.PropagationPolicyNameLabel); err != nil {
-			klog.Errorf("Failed to cleanup label from resource binding(%s/%s) when propagation policy(%s/%s) removing, error: %v",
-				binding.Namespace, binding.Name, policyNS, policyName, err)
+		// Must remove the labels from the resource template ahead of ResourceBinding, otherwise might lose the chance
+		// to do that in a retry loop(in particular, the label was successfully removed from ResourceBinding, but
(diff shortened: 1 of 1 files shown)


jwcesign (commented):
/lgtm

jwcesign:
I think we need release notes

RainbowMango (commented):
/assign

pkg/detector/detector.go
```diff
@@ -1039,20 +1039,20 @@ func (d *ResourceDetector) HandlePropagationPolicyDeletion(policyNS string, poli
...
 	for index, binding := range rbs.Items {
-		// Cleanup the labels from the reference binding so that the karmada scheduler won't reschedule the binding.
-		if err := d.CleanupResourceBindingLabels(&rbs.Items[index], policyv1alpha1.PropagationPolicyNamespaceLabel, policyv1alpha1.PropagationPolicyNameLabel); err != nil {
-			klog.Errorf("Failed to cleanup label from resource binding(%s/%s) when propagation policy(%s/%s) removing, error: %v",
-				binding.Namespace, binding.Name, policyNS, policyName, err)
-			return err
-		}
-
-		// Cleanup the labels from the object referencing by binding.
+		// Clean up the labels from the object referencing by binding first to ensure idempotence.
```
RainbowMango, line 1042: We may not need to emphasize idempotence here, as it is not easy to understand. We only need to explain why we need to do that. Something like:
```
		// Must remove the labels from the resource template ahead of ResourceBinding, otherwise might lose the chance
		// to do that in a retry loop(in particular, the label was successfully removed from ResourceBinding, but
		// resource template not), since the ResourceBinding will not be listed again.
```

whitewindmills: done

RainbowMango (approved):
/approve

Thanks @whitewindmills for the root cause analysis and quick fix.
I updated the release notes and PR title by the way.

leave LGTM to @jwcesign to take another look.

PR #4277, opened 2023-11-18
karmada-search: Fix List resource paging issue.

xigang:
**What type of PR is this?**

/kind bug

`Cacher` should be accessed when List request parameter `resourceVersion="0"` and `Limit>0`.

This is the default request parameter of client-go `ListWatch` `resourceVersion="0"` and `Limit=500`.

ref: https://github.com/kubernetes/kubernetes/blob/1f07da7575452353354a544d22264923bc4236b8/staging/src/k8s.io/apiserver/pkg/storage/cacher/cacher.go#L720

Fixes #

```release-note
NONE

```


RainbowMango:
Why closed?

PR #4296, opened 2023-11-21
sh: make cluster version configuration unified

jwcesign:
**What type of PR is this?**
/kind cleanup

sh: make cluster version configuration unified

Fixes #none

```release-note

```


XiShanYongYe-Chang (commented):
LGTM

chaosi-zju:
when can we use v1.28 ?

jwcesign:
> when can we use v1.28 ?

I am not sure.  Let's first consolidate all version modifications.

hack/util.sh
```diff
@@ -25,6 +25,8 @@ KARMADA_GO_PACKAGE="github.com/karmada-io/karmada"
 
 MIN_Go_VERSION=go1.20.0
 
+DEFAULT_CLUSTER_VERSION="kindest/node:v1.27.3"
```
RainbowMango, line 28: It makes sense to take this as the single source of truth, so when next time we bump kind, we only need to make an update here. Is it the idea?

jwcesign: Yes, that's right.

RainbowMango (approved):
/approve

chaunceyjiang:
image

My MAC m1 is not working. @jwcesign

Issue #4220, opened 2023-11-11
enhancement of dividing replicas by static weight evenly


chaosi-zju:
**What would you like to be added**:

enhancement the assignment of static weight scheduling. 

**Why is this needed**:

In static weight assignment, if the "total replicas" is not divisible by the "sum weights", then a remainder is generated.

The remainder is preferred allocating to clusters with higher dictionary order of cluster name when they have equal weights,
which caused unevenness of replicas distribution.

Therefore, Karmada needs to do:

* The remainder should be assigned to clusters having equal weights with equal probability 
* Scheduler rescheduling should ensure inertia of the scheduling result

**Related activities**

- [x] Proposal @chaosi-zju #4218
- [x] Code implementation @chaosi-zju #4219
- [ ] UT test
* - [x] #4298
* - [x] #4389

jwcesign:
Please assign e2e to me, thanks

chaosi-zju:
we at least need 5 e2e case, described as my proposal：

* Expanded replicas scenario、Reduced replicas scenario /assign @jwcesign 
* Modifying Weight Scenarios /assign @zhzhuang-zju 
* Expanded clusters scenario、Reduced clusters scenario /assign @chaosi-zju

PR #4298, opened 2023-11-22
add test for static weight random remainder

chaosi-zju:
**What type of PR is this?**

/kind failing-test

add test for static weight random remainder

Fixes part of #4220

```release-note
none
```


pkg/scheduler/core/generic_scheduler_test.go
```diff
@@ -0,0 +1,92 @@
...
+	clusterv1alpha1 "github.com/karmada-io/karmada/pkg/apis/cluster/v1alpha1"
+	policyv1alpha1 "github.com/karmada-io/karmada/pkg/apis/policy/v1alpha1"
+	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
+	"github.com/karmada-io/karmada/test/helper"
+)
+
+func Test_genericScheduler_AssignReplicas(t *testing.T) {
+	tests := []struct {
+		name       string
+		retryTimes int
```
XiShanYongYe-Chang, line 15: What is the reason for adding a retry here?

XiShanYongYe-Chang (commented):
Can more test cases be added to increase code coverage, such as when the cluster quantity is 0 or the replica number is 0?

chaosi-zju:
> Can more test cases be added to increase code coverage, such as when the cluster quantity is 0 or the replica number is 0?

of course, but the code should follow small step commits, I will add more when I next have time~

pkg/scheduler/core/generic_scheduler_test.go
chaosi-zju: it is the acceptance cases given by testers for this requirement.

she want to check when it meets random, In other word, retry multi times, the result should have both 2:1 and 1:2

chaosi-zju:
@XiShanYongYe-Chang @RainbowMango can this test be merged？

pkg/scheduler/core/generic_scheduler_test.go
XiShanYongYe-Chang: Thanks, I get it. 

This test case is a bit complex. Perhaps we can add some simple comments to explain the purpose of the test.

RainbowMango: How many times did you run the tests? 
My concern is that testing random by specific times might not stable, which means the test might fail, even though the probability is very slow.

chaosi-zju: > This test case is a bit complex. Perhaps we can add some simple comments to explain the purpose of the test.

okey

chaosi-zju: > How many times did you run the tests? My concern is that testing random by specific times might not stable, which means the test might fail, even though the probability is very slow.

10 times, in order to keep the same as `alpha test case` running by product tester.

the probability of fail is (1/2)^10= 0.1%

PR #4307, opened 2023-11-22
resourceModels supports extended resources

### LINKED ISSUE #4050: resourceModels supports extended resources
I am able to register clusters with push mode, and use default resourceModel which only supports cpu, memory, ephemeral-storage and storage.
When I add extended resources like gpu in the resourceModels in the cluster like
```
  resourceModels:
  - grade: 0
    ranges:
    - max: "72"
      min: "0"
      name: cpu
    - max: 560Gi
      min: "0"
      name: memory
    - max: "0"
      min: "0"
      name: nvidia.com/gpu
    - max: "0"
      min: "0"
      name: myexample.com/gpu-v100
  - grade: 1
    ranges:
    - max: "96"
      min: "72"
      name: cpu
    - max: 1.6Ti
      min: 560Gi
      name: memory
    - max: "4"
      min: "0"
      name: nvidia.com/gpu
    - max: "4"
      min: "0"
      name: myexample.com/gpu-v100
```
I get
```
Unsupported value: "nvidia.com/gpu": supported values: "cpu", "ephemeral-storage", "memory", "storage"
```

My understanding is that [General Cluster Modeling](https://karmada.io/docs/userguide/scheduling/cluster-resources#general-cluster-modeling) use the resourceSummary to check the allocatable/allocated to schedule the pod.
But we also want to have GPU in [Customized Cluster Modeling](https://karmada.io/docs/userguide/scheduling/cluster-resources#customized-cluster-modeling), I don't think gpu will have the fragmentation issues like gpu/memory when using general cluster modeling, as ppl cannot claim partial gpu. But it would be still preferred to have extended resources in the customized cluster modeling just make it consistent with cluster  resources ?
I also run Karmada dashboard, it shows cpu, memory and storage. But it would be nice to show extended resources ?

**Environment**:
- Karmada version: latest
- Kubernetes version: 1.24
- Others:

chaosi-zju:
**What type of PR is this?**

/kind feature

resourceModels supports extended resources

Fixes #4050

Demo：https://h3ld32xlpo.feishu.cn/wiki/Rj5mwS3TCigHY4kMKzhcqTggnlg

A perfect way to edit extension resource model in cluster object is like:

```shell
for cluster in "member1" "member2"; do
  kubectl --context karmada-apiserver patch cluster ${cluster} -n karmada-system --type='json' -p '[
      {"op": "add", "path": "/spec/resourceModels/0/ranges/-", "value": {"name": "karmada.io/gpu", "min": "0", "max": "4"}},
      {"op": "add", "path": "/spec/resourceModels/1/ranges/-", "value": {"name": "karmada.io/gpu", "min": "4", "max": "8"}},
      {"op": "add", "path": "/spec/resourceModels/2/ranges/-", "value": {"name": "karmada.io/gpu", "min": "8", "max": "12"}},
      {"op": "add", "path": "/spec/resourceModels/3/ranges/-", "value": {"name": "karmada.io/gpu", "min": "12", "max": "16"}},
      {"op": "add", "path": "/spec/resourceModels/4/ranges/-", "value": {"name": "karmada.io/gpu", "min": "16", "max": "20"}},
      {"op": "add", "path": "/spec/resourceModels/5/ranges/-", "value": {"name": "karmada.io/gpu", "min": "20", "max": "24"}},
      {"op": "add", "path": "/spec/resourceModels/6/ranges/-", "value": {"name": "karmada.io/gpu", "min": "24", "max": "28"}},
      {"op": "add", "path": "/spec/resourceModels/7/ranges/-", "value": {"name": "karmada.io/gpu", "min": "28", "max": "32"}},
      {"op": "add", "path": "/spec/resourceModels/8/ranges/-", "value": {"name": "karmada.io/gpu", "min": "32", "max": "9223372036854775807"}},
  ]'
done
```

```release-note
`karmada-controller-manager`: Resource models now support any arbitrary resource type and are no longer limited to `cpu`, `memory`, `storage`, and `ephemeral-storage`.
```

### DIFF
--- a/pkg/apis/cluster/types.go
+++ b/pkg/apis/cluster/types.go
@@ -22,26 +22,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// ResourceName is the name identifying various resources in a ResourceList.
-type ResourceName string
-
-// Resource names must be not more than 63 characters, consisting of upper- or lower-case alphanumeric characters,
-// with the -, _, and . characters allowed anywhere, except the first or last character.
-// The default convention, matching that for annotations, is to use lower-case names, with dashes, rather than
-// camel case, separating compound words.
-// Fully-qualified resource typenames are constructed from a DNS-style subdomain, followed by a slash `/` and a name.
-const (
-	// ResourceCPU in cores. (e,g. 500m = .5 cores)
-	ResourceCPU ResourceName = "cpu"
-	// ResourceMemory in bytes. (e,g. 500Gi = 500GiB = 500 * 1024 * 1024 * 1024)
-	ResourceMemory ResourceName = "memory"
-	// ResourceStorage is volume size, in bytes (e,g. 5Gi = 5GiB = 5 * 1024 * 1024 * 1024)
-	ResourceStorage ResourceName = "storage"
-	// ResourceEphemeralStorage is local ephemeral storage, in bytes. (e,g. 500Gi = 500GiB = 500 * 1024 * 1024 * 1024)
-	// The resource name for ResourceEphemeralStorage is alpha and it can change across releases.
-	ResourceEphemeralStorage ResourceName = "ephemeral-storage"
-)
-
 //revive:disable:exported
 
 // +genclient
@@ -233,7 +213,7 @@ type ResourceModel struct {
 type ResourceModelRange struct {
 	// Name is the name for the resource that you want to categorize.
 	// +required
-	Name ResourceName
+	Name corev1.ResourceName
 
 	// Min is the minimum amount of this resource represented by resource name.
 	// Note: The Min value of first grade(usually 0) always acts as zero.
--- a/pkg/apis/cluster/validation/validation.go
+++ b/pkg/apis/cluster/validation/validation.go
@@ -54,8 +54,7 @@ func ValidateClusterName(name string) []string {
 }
 
 var (
-	supportedSyncModes     = sets.NewString(string(api.Pull), string(api.Push))
-	supportedResourceNames = sets.NewString(string(api.ResourceCPU), string(api.ResourceMemory), string(api.ResourceStorage), string(api.ResourceEphemeralStorage))
+	supportedSyncModes = sets.NewString(string(api.Pull), string(api.Push))
 )
 
 // ValidateCluster tests if required fields in the Cluster are set.
@@ -208,9 +207,6 @@ func ValidateClusterResourceModels(fldPath *field.Path, models []api.ResourceMod
 		}
 
 		for j, resourceModelRange := range resourceModel.Ranges {
-			if !supportedResourceNames.Has(string(resourceModelRange.Name)) {
-				return field.NotSupported(fldPath.Child("ranges").Child("name"), resourceModelRange.Name, supportedResourceNames.List())
-			}
 			if resourceModelRange.Max.Cmp(resourceModelRange.Min) <= 0 {
 				return field.Invalid(fldPath, models, "The max value of each resource must be greater than the min value")
 			}
--- a/pkg/apis/cluster/mutation/mutation.go
+++ b/pkg/apis/cluster/mutation/mutation.go
@@ -84,12 +84,12 @@ func SetDefaultClusterResourceModels(cluster *clusterapis.Cluster) {
 			Grade: 0,
 			Ranges: []clusterapis.ResourceModelRange{
 				{
-					Name: clusterapis.ResourceCPU,
+					Name: corev1.ResourceCPU,
 					Min:  *resource.NewQuantity(0, resource.DecimalSI),
 					Max:  *resource.NewQuantity(1, resource.DecimalSI),
 				},
 				{
-					Name: clusterapis.ResourceMemory,
+					Name: corev1.ResourceMemory,
 					Min:  *resource.NewQuantity(0, resource.BinarySI),
 					Max:  *resource.NewQuantity(4*GB, resource.BinarySI),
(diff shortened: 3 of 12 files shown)


pkg/apis/cluster/types.go
```diff
@@ -6,26 +6,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// ResourceName is the name identifying various resources in a ResourceList.
```
chaunceyjiang, line 9: I remember why we didn't directly reference `corev1.ResourceName` at first, it's because we considered that there might be some extensions here, different from the `corev1`.

chaosi-zju: Hi @chaunceyjiang, `corev1.ResourceName` is just a `typedef`, not a `enum`, extensions can also assign to `corev1.ResourceName`

There is a demo: https://h3ld32xlpo.feishu.cn/wiki/Rj5mwS3TCigHY4kMKzhcqTggnlg

pkg/apis/cluster/types.go
```diff
@@ -217,7 +197,7 @@ type ResourceModel struct {
 type ResourceModelRange struct {
 	// Name is the name for the resource that you want to categorize.
 	// +required
-	Name ResourceName
+	Name corev1.ResourceName
```
XiShanYongYe-Chang, line 200: I want to confirm one thing, this modification should not introduce any compatibility issues, right?

pkg/apis/cluster/validation/validation.go
```diff
@@ -38,8 +38,7 @@ func ValidateClusterName(name string) []string {
 }
 
 var (
-	supportedSyncModes     = sets.NewString(string(api.Pull), string(api.Push))
-	supportedResourceNames = sets.NewString(string(api.ResourceCPU), string(api.ResourceMemory), string(api.ResourceStorage), string(api.ResourceEphemeralStorage))
```
XiShanYongYe-Chang, line 42: I understand that the value of user resources should be planned in advance and not too arbitrary. Can we set the key values of supported resources for users from a configurable perspective, while still retaining the validation behavior?

pkg/apis/cluster/types.go
chaosi-zju: good viewpoint

pkg/apis/cluster/validation/validation.go
chaosi-zju: how about a env of karmada-apiserver？I don't want to add any new api for this, a launch param or env is acceptable

XiShanYongYe-Chang: I personally prefer to start with parameter or configuration files or resources.

pkg/apis/cluster/types.go
chaosi-zju: if I deploy karmada by master branch, and then replace the image of aa、controller-manager、scheduler by this commit, if it support extensions, can it verify is compatible？

XiShanYongYe-Chang: I understand that it is possible. The Cluster object should already exist before the upgrade, and after the upgrade, the Cluster resources can still be used normally, which should meet the requirements for upgrading.

pkg/apis/cluster/types.go
RainbowMango: @chaunceyjiang Can you remind me what kind of extensions are here? I totally forgot it.

cc @Poor12

pkg/apis/cluster/validation/validation.go
```diff
@@ -192,9 +191,6 @@ func ValidateClusterResourceModels(fldPath *field.Path, models []api.ResourceMod
 		}
 
 		for j, resourceModelRange := range resourceModel.Ranges {
-			if !supportedResourceNames.Has(string(resourceModelRange.Name)) {
-				return field.NotSupported(fldPath.Child("ranges").Child("name"), resourceModelRange.Name, supportedResourceNames.List())
-			}
```
RainbowMango, line 197: The key point of this PR is removing this validation, right?

pkg/apis/cluster/types.go
chaunceyjiang: https://github.com/karmada-io/karmada/pull/2367#discussion_r946298844

pkg/apis/cluster/validation/validation.go
chaosi-zju: yes

pkg/apis/cluster/validation/validation.go
chaosi-zju: I think it is no big influence without validation, if user mistake in writing in any one grade, we have other ways to report error (the resource's range should be defined in every grade, others report error)

so, I advice to keep unchanged for the moment.

pkg/apis/cluster/types.go
chaosi-zju: I did a carefully test as above described, I is actually compatible~

chaosi-zju:
I have carefully tested this feature, it is ok. 

Is there any further comments? If you have any other suggestions, please let me know~

XiShanYongYe-Chang:
Ok, thanks~
LGTM

RainbowMango (commented):
/lgtm

@chaunceyjiang Do you have any further comment?

chaosi-zju:
Test report updated

Poor12:
LGTM.
Thank you for enhancing this feature.

chaunceyjiang (approved):
/lgtm

RainbowMango (approved):
/approve

I updated the release note by the way.

PR #4289, opened 2023-11-21
proposal about Statization of Propagation Policy

chaosi-zju:
**What type of PR is this?**

/kind design
/kind documentation

**What this PR does / why we need it**:

In this paper, We propose an improvement to the Policy enforcement mechanism, called Policy staticization.
Before the introduction of this feature, modification of a Policy would immediately affect the resource templates it manages, potentially causing a huge shock to the system, which is too sensitive.

We want Policy to be just some API-based static system strategies. Modification of the Policy should not actively cause changes to the system behavior, but should be triggered by ResourceTemplate. 

Just like if a ResourceTemplate has already been propagated by a Policy, don't change its states for any Policy changes, but you can re-propagate it by latest Policy when ResourceTemplate itself updated.

More details refer to #4288 and proposal content.

**Which issue(s) this PR fixes**:

Fixes #4288

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
introduces a modification to the behavior of Propagation Policy, making the Propagation Policy to be static.
```

Issue #4288, opened 2023-11-21
Statization of Propogation Policy


chaosi-zju:
**What would you like to be added**:

it is proposed to modify the default behavior of Propogation Policy to achieve the statization:

* If a ResourceTemplate is not managed by a Policy, when there is a matching Policy in the system, 
  the ResourceTemplate will be distributed according to the Policy with the highest priority.
* If the ResourceTemplate is already managed by a Policy, the distribution status of the Resource Template remains unchanged, 
  regardless of whether the Policy has been modified or a higher-priority Policy applied.
* When a user submits a change to a ResourceTemplate, if there is a matchable higher priority Policy in the system, 
  the ResourceTemplate will follow the Policy to adjust its distribution states; 
  otherwise, if the previous matching Policy is modified, the ResourceTemplate will also follow the modified Policy to adjust the distribution states.

**Why is this needed**:

The current Propogation Policy mechanism has limitations in some scenarios, If we can staticize the Policy, will make the Policy more reasonable and flexible.

* PropagationPolicy / ClusterPropagationPolicy should not have any API fields that can actively cause system changes.
* PropagationPolicy / ClusterPropagationPolicy are just a bunch of API-enabled system functions, changes should be triggered from the Resource Template.
* If workloads are running fine, don't roughly change them!

**Expected Examples**

#### Case 1

1. create ResourceTemplate DeployA
2. create PropagationPolicy Policy1 (match for DeployA)

```mermaid
sequenceDiagram
  participant DeployA
  participant Policy1
  participant Karmada

  
  DeployA ->> Karmada: create
  activate Karmada
  Policy1 ->> Karmada: create
  Karmada -->> DeployA: propagate it by Policy1
  deactivate Karmada
```

> Note: DeployA is constantly waiting for a matchable Policy, so when Policy1 applied, DeployA would be directly propagated
> by Policy1 without an additional updates.

#### Case 2

1. create PropagationPolicy Policy1 (match for DeployA)
2. create ResourceTemplate DeployA
3. update Policy1
4. update DeployA

```mermaid
sequenceDiagram
  participant DeployA
  participant Policy1
  participant Karmada

  Policy1 ->> Karmada: create
  activate Karmada
  DeployA ->> Karmada: create
  Karmada -->> DeployA: propagate it by Policy1
  deactivate Karmada
  
  Policy1 ->> Karmada: update
  activate Karmada
  loop
    Policy1 --> Karmada: no action util DeployA updated
  end
[...]

PR #4321, opened 2023-11-25
implementation of Statization of Propagation Policy

chaosi-zju:
**What type of PR is this?**

/kind feature

In #4289, we propose an improvement to the Policy enforcement mechanism, called Policy staticization.
Before the introduction of this feature, modification of a Policy would immediately affect the resource templates it manages, potentially causing a huge shock to the system, which is too sensitive.

We want Policy to be just some API-based static system strategies. Modification of the Policy should not actively cause changes to the system behavior, but should be triggered by ResourceTemplate.

Just like if a ResourceTemplate has already been propagated by a Policy, don't change its states for any Policy changes, but you can re-propagate it by latest Policy when ResourceTemplate itself updated.

More details refer to https://github.com/karmada-io/karmada/issues/4288 and proposal content.

Fixes part of #4288

```release-note
introduces a modification to the behavior of Propagation Policy, making the Propagation Policy to be static.
```


chaosi-zju:
/close

This implementation is no longer used as the default behavior of Policy, and the design is rejected.

PR #4338, opened 2023-11-30
fix incorrect forType in clusterResourceBinding status controller

lxtywypc:
**What type of PR is this?**  
/kind bug  

**What this PR does / why we need it**:  
Fix the incorrect `forType` declared in `clusterresourcebinding-status controller`  

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:  

**Does this PR introduce a user-facing change?**:

```release-note
`karmada-controller-manager`: fix incorrect `forType` in `cluster-resource-binding-status controller` from `ResouceBinding` to `ClusterResourceBinding` 
```

PR #4340, opened 2023-11-30
add forType in binding status controller to inherit concurrency declared in global options

### LINKED ISSUE #4339: Unexpected Setting Concurrency  as 1 in Binding Status Controller
**What happened**:  
Because of missing `forType` in `binding-status controller`, the concurrency of the controller would be always set as 1.  

The concurrency of `binding-status controller` could inherit from the concurrency of `binding` declared in global options.  

If a controller is built WITHOUT concurrency options, AND it HAS a `forType`, the concurrency of the controller would inherit from the concurrency for `forType` declared in global options.  

https://github.com/karmada-io/karmada/blob/01b086d2e1a231d470cbf2742b2667d87d656e2d/vendor/sigs.k8s.io/controller-runtime/pkg/builder/controller.go#L354-L370   

Up to now, all the concurrency of controllers in `Karmada` inherit from the global options. And all controllers except `binding-status controller` has a `forType`. So only `binding-status controller` would has a concurrency as `1`.  

There are two ways to solve the problem:  
1. Add a `forType` with a fake predicate to inherit concurrency from global options and keep the same performance at the same time.  
2. Add concurrency options when building the `binding-status controller`.  

This problem occurs after we seprate `binding-status controller` from `binding controller`. On the latest `master` branch, this problem is fixed after the #4138 coincidentally, but in the earlier version it still exists.  

It might be hard to cherry-pick #4138 straightly because it seems a fix for feature `hpaReplicasSyncer`, that might not necessary in earlier version.  

I would try to raise a PR to solve it in first way soon. But it would bring differences on commit history between `master` branch and other branches of eralier version, I wonder if it is acceptable, or what the recommend way is.

**Environment**:
- Karmada version:
- Others:

lxtywypc:
**What type of PR is this?**  
/kind bug  

Fixes #4339 

```release-note
`karmada-controllr-manager`: concurrency of `binding-status controller` inherit from the concurrency of `binding` declared in global options
```

### DIFF
--- a/pkg/controllers/status/common.go
+++ b/pkg/controllers/status/common.go
@@ -21,6 +21,15 @@ import (
 	"github.com/karmada-io/karmada/pkg/util/restmapper"
 )
 
+// Fake predicate for inheriting concurrency of `forType` from global options.
+// Matching the feature in later version.
+var bindingPredicateFn = builder.WithPredicates(predicate.Funcs{
+	CreateFunc:  func(event.CreateEvent) bool { return false },
+	UpdateFunc:  func(event.UpdateEvent) bool { return false },
+	DeleteFunc:  func(event.DeleteEvent) bool { return false },
+	GenericFunc: func(event.GenericEvent) bool { return false },
+})
+
 var workPredicateFn = builder.WithPredicates(predicate.Funcs{
 	CreateFunc: func(e event.CreateEvent) bool { return false },
 	UpdateFunc: func(e event.UpdateEvent) bool {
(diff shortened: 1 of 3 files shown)


XiShanYongYe-Chang:
Ask @RainbowMango to help rerun the CI.
/cc @RainbowMango

pkg/controllers/status/common.go
```diff
@@ -21,6 +21,15 @@ import (
 	"github.com/karmada-io/karmada/pkg/util/restmapper"
 )
 
+// Fake predicate for inheriting concurrency of `forType` from global options.
+// Matching the feature in later version.
+var bindingPredicateFn = builder.WithPredicates(predicate.Funcs{
```
XiShanYongYe-Chang, line 26: I think this name can be set more commonly, how about `interceptAllPredicateFn` or something else?

XiShanYongYe-Chang (commented):
Thanks~

XiShanYongYe-Chang:
/cc @RainbowMango 
Ask @RainbowMango to help review it.

RainbowMango (approved):
/approve

Seems the #4338 fixed the issue, this PR more like a cherry-pick on release-1.7.

PR #4373, opened 2023-12-05
fix estimator license

whitewindmills:
**What type of PR is this?**
/kind cleanup

Fixes #

```release-note
NONE
```


hack/verify-license.sh
```diff
@@ -35,12 +35,10 @@ missing_license_header_files="$($ADDLICENSE_BIN \
   -ignore "docs/**" \
   -ignore ".github/**" \
   -ignore "third_party/**" \
-  -ignore "**/pb/**" \
   -ignore "**/*.md" \
   -ignore "**/*.yaml" \
   -ignore "**/*.yml" \
   -ignore "**/*.json" \
-  -ignore "**/*.proto" \
```
RainbowMango, line 43: Just a question, seems the `verify-license.sh` would ignore the `generated` files, but here doesn't have a flag like `-ignore *generated*.go*`, do you know why?

whitewindmills: The script `verify-license.sh` uses [`addlicense`](https://github.com/google/addlicense/tree/master) to check licenses. It is the hard-coded default behavior.
image

RainbowMango: Thanks for the clarification. I see it the file with `Code Generated .* DO NOT EDIT` will be treated as a generated file.
https://github.com/google/addlicense/blob/4caba19b7ed7818bb86bc4cd20411a246aa4a524/main.go#L360-L370

RainbowMango (approved):
/lgtm
/approve
Thanks.

PR #4184, opened 2023-10-27
ci: add schedule image-scanning

zhzhuang-zju:
**What type of PR is this?**

**What this PR does / why we need it**:
Add the workflow to scan release version images(release-1.5/1.6/1.7) on schedule.
**Which issue(s) this PR fixes**:
Fixes #none

**Special notes for your reviewer**:
Refer to https://github.com/zhzhuang-zju/karmada/actions/runs/6662001229
**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #4384, opened 2023-12-06
add image scanning on pull_request

zhzhuang-zju:
**What type of PR is this?**
/kind cleanup

add image scanning on pull_request to protect against security vulnerabilities
Fixes #

NONE

```release-note
NONE
```


zhzhuang-zju:
> @liangyuanpeng Can you help to take a look?

Hi~If interested, you can also review pr #4184 together, which also integrates trivy to ci, but for security scanning of older images.

liangyuanpeng:
Interesting,I will add this to my list ASAP :)

liangyuanpeng (commented):
I can understand what this PR hopes to do, but I think scheduled execution may be enough, and I think the value added by scanning container images on the PR may be very small, even if it does not hinder the PR but does take up some github action CI time.

Could you please add that scheduled scan image is not sufficient? Remind me if i missed something,Thanks.
 @zhzhuang-zju

.github/workflows/ci-image-scanning.yaml
```diff
@@ -1,6 +1,11 @@
 name: image-scanning
 on:
-  push: 
+  push:
+    branches:
+      - master
```
liangyuanpeng, line 5: Maybe we can add `release-*` to here for push event.

liangyuanpeng: I probably don't think we need to scan container images in PRs. we can refer to something similar done on https://github.com/kubernetes-sigs/descheduler/pull/898/files

```yaml
name: "Security"

on:
  push:
    branches:
      - main
      - master
      - release-*
  schedule:
    - cron: '30 1 * * 0'
```

zhzhuang-zju: > Maybe we can add `release-*` to here for push event.

The current strategy is that the master branch image scan is triggered when pushing, and then because there are fewer PRs for release, it is triggered by timing, referring to #4184. Of course, if you think it is necessary, we can also add `release-*` to here for push event.

zhzhuang-zju: > I probably don't think we need to scan container images in PRs. we can refer to something similar done on https://github.com/kubernetes-sigs/descheduler/pull/898/files

Adding image scanning on pull_request is because we encounter a scenario. If a pr is proposed to solve a certain vulnerability, the scanning results need to be seen in PRs to determine whether the vulnerability is solved correctly.

liangyuanpeng: Thanks for the context, make sense for me.Should we only run this CI when we need to check? For example the PR has the label `area/fix-cve` or `area/security`

I tried to see how the kubernetes community does this:

1. The PR title is created in the format of [CVE-ID]
2. There is a security-related scheduled task that scans related PRs every few hours and updates the data to the dashboard, https://testgrid.k8s.io/sig-security-snyk-scan#ci-kubernetes-snyk-master

But this may not apply to karmada.

zhzhuang-zju: > Thanks for the context, make sense for me.Should we only run this CI when we need to check? For example the PR has the label `area/fix-cve` or `area/security`
> 
Good idea~ Do you know how to achieve this? I didn't find the corresponding documentation.
> I tried to see how the kubernetes community does this:
> 
> 1. The PR title is created in the format of [CVE-ID]
> 2. There is a security-related scheduled task that scans related PRs every few hours and updates the data to the dashboard, https://testgrid.k8s.io/sig-security-snyk-scan#ci-kubernetes-snyk-master
> 
> But this may not apply to karmada.

Thanks for your research. I also think it makes sense to create a diff between vulnerability scan report run on master branch and the one run on HEAD (current) branch. This can be used as an optimization point for the next iteration.

liangyuanpeng: Here is a quickly written shell script.

```shell
  checklabels:
    name: check-pr-labels
    runs-on: ubuntu-22.04
    if: ${{ github.event_name == 'pull_request' }}
    outputs:
      areaSecurity: ${{ steps.checklabels.outputs.security }}
    steps:
      - name: checklabels 
        id: checklabels
        env: 
          GHEVENT: ${{ toJSON(github.event.pull_request) }}
        run: |
          if echo  $GHEVENT | jq .labels | grep -q "area/security"; then
            echo "security=true" >> "$GITHUB_OUTPUT"
          else
            echo "Label does not exist area/security"
          fi

  use-trivy-to-scan-image:
    needs: checklabels
    if: needs.checklabels.outputs.areaSecurity == 'true'
    name: image-scanning
```

Ask other people if they agree with it before start doing it.


cc  @RainbowMango @XiShanYongYe-Chang

liangyuanpeng: > and then because there are fewer PRs for release, it is triggered by timing,

Actually, CVE events will occur even if there is no PR. I will suggess to create two github action workflow,

1. schedule and push event to  run for master & release-* (today, this is release-1.6, release-1.7, release-1.8 ),
2. on pull_request to run for PR

@zhzhuang-zju  WDYH

zhzhuang-zju: > Actually, CVE events will occur even if there is no PR. I will suggess to create two github action workflow,
> 
> 1. schedule and push event to  run for master & release-* (today, this is release-1.6, release-1.7, release-1.8 ),

Currently, only release 1.8 and above branches have workflow files for image scanning, so if you want to achieve this 
goal, release 1.6 and release 1.7 must also add this workflow file.

> 2. on pull_request to run for PR


Combined with the work in #4184, my current thinking is that the master branch is triggered during push and PR because it is relatively active, and the release-* branch are triggered on schedule because there are relatively few changes. We may need to find other people to decide which method is better.
cc @RainbowMango @XiShanYongYe-Chang

liangyuanpeng: The advantage of placing schedule and push with `master/release-*` in the same workflow is that you don't need to do things like `if: ${{ github.event_name == 'push' }}`  when on pull_request for PR, and the structure will be clearer.

Essentially master and release are the same thing. That's Just My Two Cents. :)

zhzhuang-zju: > The advantage of placing schedule and push with `master/release-*` in the same workflow is that you don't need to do things like `if: ${{ github.event_name == 'push' }}` when on pull_request for PR, and the structure will be clearer.
> 
That's make sense for me
> Essentially master and release are the same thing, master also need to schedule scan. That's Just My Two Cents. :)

zhzhuang-zju: > > and then because there are fewer PRs for release, it is triggered by timing,
> 
> Actually, CVE events will occur even if there is no PR. I will suggess to create two github action workflow,
> 
> 1. schedule and push event to  run for master & release-* (today, this is release-1.6, release-1.7, release-1.8 ),
> 2. on pull_request to run for PR
> 

Hi @liangyuanpeng, I decided to use your suggestion. Modification is ongoing, looking forward to your code review~

zhzhuang-zju:
@liangyuanpeng @RainbowMango 
Hi~ On the basis of the original image scanning, I tried to add some vulnerability analysis. By comparing the vulnerabilities scanned in the PR with those scanned in the Baseline branch, the following content can be analyzed:
- the vulnerabilities fixed by PR
refer to https://github.com/zhzhuang-zju/karmada/actions/runs/7582306257/job/20651617732?pr=3
image
- the vulnerabilities introduced by this PR
refer to https://github.com/zhzhuang-zju/karmada/actions/runs/7582111908/job/20651024425
image

The CI failure is because a merge of #4567  is required first.

XiShanYongYe-Chang:
This effect looks great.

PR #4386, opened 2023-12-07
Proposal for cross cluster stateful app

liangyuanpeng:
**What type of PR is this?**

/kind design
/kind documentation

Fixes #

```release-note

```

### DIFF
--- a/docs/proposals/corss-cluster-statefulset/README.md
+++ b/docs/proposals/corss-cluster-statefulset/README.md
@@ -1,0 +1,118 @@
+---
+title: Your short, descriptive title
+authors:
+- "@robot" # Authors' github accounts here.
+reviewers:
+- "@robot"
+- TBD
+approvers:
+- "@robot"
+- TBD
+
+creation-date: yyyy-mm-dd
+
+---
+
+# Your short, descriptive title
+
+<!--
+This is the title of your KEP. Keep it short, simple, and descriptive. A good
+title can help communicate what the KEP is and should be considered as part of
+any review.
+-->
+
+## Summary
+
+With the development of multi-cluster technology, some users begin to want to migrate stateful applications to multi-cluster scenarios, which cannot be covered well by the current version of karmada. Each user needs to implement a `karmada-operator` that does something similar, so we can propose an API to handle the common coordination related work logic, while setting aside a mechanism for the user's `karmada-operator` to complete the specific logic, such as rolling upgrades, scaling, etc.
+
+## Motivation
+
+Provides a new API for stateful applications across clusters, enabling karmada to adapt to stateful applications across clusters.
+
+### Goals
+
+- Defining an API(TBD,karmada/StatefulSet?) enables users to implement specific logic for stateful operator applications across clusters.
+-  Propose the implementation ideas for involved components, including the new controller of `cross_cluster_statefulset_controller` in karma -controller-manager.
+
+### Non-Goals
+
+## Proposal
+
+<!--
+This is where we get down to the specifics of what the proposal actually is.
+This should have enough detail that reviewers can understand exactly what
+you're proposing, but should not include things like API designs or
+implementation. What is the desired outcome and how do we measure success?.
+The "Design Details" section below is for the real
+nitty-gritty.
+-->
+
+### User Stories (Optional)
+
+#### As a user, I want to deploy a cross cluster stateful app.
+
+I want to let my kubernetes operator can working for cross cluster stateful with karmada and not need to implement other operator for karmada.
+
+#### Story 1
+
+#### Story 2
+
+### Notes/Constraints/Caveats (Optional)
+
+<!--
+What are the caveats to the proposal?
+What are some important details that didn't come across above?
+Go in to as much detail as necessary here.
+This might be a good place to talk about core concepts and how they relate.
+-->
+
+### Risks and Mitigations
+
+<!--
+What are the risks of this proposal, and how do we mitigate? 
+
+How will security be reviewed, and by whom?
+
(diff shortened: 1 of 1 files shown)


liangyuanpeng:
This is still in draft status, but existing content can be discussed if possible. Any discussion would be help.

/cc @RainbowMango 
cc the end user  @tpiperatgod @Phoenix500526

RainbowMango (commented):
Great thanks to @liangyuanpeng for leading the effort on this. I'm interested in this approach.

docs/proposals/corss-cluster-statefulset/README.md
```diff
@@ -0,0 +1,213 @@
...
+```yaml
+# member1
+Name:         xline
+Namespace:    xline
+API Version:  apps.my.io/v1alpha1
+Kind:         XlineCluster
+Annotations:
+    corssclusterstatefulset.karmada.io/replicas: [
+      {"member1":"1","self":true},
+      {"member1":"2","self":false}
```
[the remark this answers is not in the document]
liangyuanpeng: > I might not get what the self means here,

The only purpose of this comment is for the network topology of the current cluster operator component. Since the CR obtained by each member cluster actually already has the corresponding  replicas, I will confirm whether self is necessary later. If If it is not necessary, it may be: 

```json
  {"member1":"1","member2":"2"}
```


> but why not put them to the API instead of an annotation?

An important reason is to reduce intrusion. Operators of member clusters do not have to connect to karmada, only  monitor the CR changes of the current cluster.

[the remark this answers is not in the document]
liangyuanpeng: cc @Phoenix500526 @tpiperatgod

docs/proposals/corss-cluster-statefulset/README.md
```diff
@@ -0,0 +1,213 @@
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// Spec is the desired state of the MultiClusterIngress.
+	// +optional
+	Spec CrossClusterStatefulsetSetSpec `json:"spec,omitempty"`
+}
+
+// CrossClusterStatefulsetSetSpec is the desired state of the CrossClusterStatefulsetSet.
+type CrossClusterStatefulsetSetSpec struct {
+	ResourceSelector policyv1alpha1.ResourceSelector `json:"resourceSelector,omitempty"`
```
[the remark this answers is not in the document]
liangyuanpeng: Apparently I forgot to add `UpdateStrategy` here. `type=rollupdate` is the only value, nothing else.

Invite users to add more expectations.

@Phoenix500526 @tpiperatgod

docs/proposals/corss-cluster-statefulset/README.md
```diff
@@ -0,0 +1,243 @@
...
+implementation. What is the desired outcome and how do we measure success?.
+The "Design Details" section below is for the real
+nitty-gritty.
+-->
+
+### User Stories (Optional)
+
+#### As a user, I want to deploy a cross cluster stateful app.
+
+In order to enable my stateful application to achieve disaster recovery in multiple clusters and obtain better regional services,I want to let my kubernetes operator from member cluster can working for cross cluster stateful with karmada and not need to implement other operator for karmada.
```
jwcesign, line 58: I think we should give some specific scenarios, then we can see what we expect.

Like MySQL statefulset in multiple clusters, the start index should be different(mysql-0, mysql-1), then we expect only one master MySQL instance. And these instances will detect others' through mysql-{x}.default.xxx.

docs/proposals/corss-cluster-statefulset/README.md
```diff
@@ -0,0 +1,243 @@
...
+propagate schedule result with CRD annotation,just like:
+
+```yaml
+# member1
+Name:         xline
+Namespace:    xline
+API Version:  apps.my.io/v1alpha1
+Kind:         XlineCluster
+Annotations:
+    corssclusterstatefulset.karmada.io/replicas: [
```
jwcesign, line 145: What's the default behavior for the K8s default statefulset, should we set the start index for different clusters?

docs/proposals/corss-cluster-statefulset/README.md
```diff
@@ -0,0 +1,243 @@
...
+Annotations:
+    corssclusterstatefulset.karmada.io/processer: true #true/false
+---
+# member2
+Name:         xline
+Namespace:    xline
+API Version:  apps.my.io/v1alpha1
+Kind:         XlineCluster
+Annotations:
+    corssclusterstatefulset.karmada.io/processer: true #true/false
```
jwcesign, line 190: I don't understand what this field means. Could you explain it further?

docs/proposals/corss-cluster-statefulset/README.md
```diff
@@ -0,0 +1,243 @@
...
+// all possible update strategies for the CrossClusterStatefulSet controller.
+type CrossClusterStatefulUpdateStrategyType string
+
+const (
+	// RollingUpdateCrossClusterStatefulSetStrategyType indicates that update will be
+	// applied to all CRs in the StatefulSet with respect to the StatefulSet
+	// ordering constraints. When a scale operation is performed with this
+	// strategy, new CR will be created from the specification version indicated
+	// by the CrossClusterStatefulSet's updateRevision.
+	RollingUpdateCrossClusterStatefulSetStrategyType CrossClusterStatefulSetUpdateStrategyType = "RollingUpdate"
```
jwcesign, line 127: What is the purpose of this field? Could you provide an example?

PR #4398, opened 2023-12-11
adjust the Priority of General and Accurate Estimator of Karmada Scheduler

### LINKED ISSUE #3333: Adjust the Priority of General and Accurate Estimator of Karmada Scheduler

Currently, there are three estimating methods in scheduler estimator including `ClusterResourceSummary`, `ResourceModeling` and `Accurate estimator`. The `ResourceModeling` is prior to `ClusterResourceSummary` and parallel with `Accurate estimator`. 

https://github.com/karmada-io/karmada/blob/1d030d4f8c01f8370c80bb0ecdd52630d87a5bf0/pkg/scheduler/core/util.go#L46-L64

When all those are enabled, `karmada-scheduler` will use the minumal value that they estimate. However, the estimating result of `accurate estimator` is more accurate than `ResourceModeling` and `ResourceModeling`. So if the `accurate estimator` is enabled and `karmada-scheduler-estimator` components also are enabled, the `accurate estimator` should be prior to `ResourceModeling` rather than using the minumal value of them.

**Why is this needed**:

The estimating result of `accurate estimator` is more accurate to `ResourceModeling` and `ClusterResourceSummary`.

chaosi-zju:
**What type of PR is this?**

/kind feature

 If the accurate estimator is enabled and `karmada-scheduler-estimator` components also are enabled, the accurate estimator should be prior to `ResourceModeling` rather than using the minumal value of them.

Fixes #3333

CC @lonelyCZ @whitewindmills 

```release-note
introduced priority to schedule estimator so that higher-priority estimator can override the result given by lower-priority.
```

### DIFF
--- a/pkg/descheduler/core/helper.go
+++ b/pkg/descheduler/core/helper.go
@@ -62,26 +62,60 @@ func NewSchedulingResultHelper(binding *workv1alpha2.ResourceBinding) *Schedulin
 func (h *SchedulingResultHelper) FillUnschedulableReplicas(unschedulableThreshold time.Duration) {
 	reference := &h.Spec.Resource
 	undesiredClusters, undesiredClusterNames := h.GetUndesiredClusters()
+	// clusterIndexToEstimatorPriority key refers to index of cluster slice,
+	// value refers to the EstimatorPriority of who gave its estimated result.
+	clusterIndexToEstimatorPriority := make(map[int]estimatorclient.EstimatorPriority)
+
 	// Set the boundary.
 	for i := range undesiredClusters {
 		undesiredClusters[i].Unschedulable = math.MaxInt32
 	}
-	// Get the minimum value of MaxAvailableReplicas in terms of all estimators.
+
+	// Get all replicaEstimators, which are stored in TreeMap.
 	estimators := estimatorclient.GetUnschedulableReplicaEstimators()
 	ctx := context.WithValue(context.TODO(), util.ContextKeyObject,
 		fmt.Sprintf("kind=%s, name=%s/%s", reference.Kind, reference.Namespace, reference.Name))
-	for _, estimator := range estimators {
-		res, err := estimator.GetUnschedulableReplicas(ctx, undesiredClusterNames, reference, unschedulableThreshold)
-		if err != nil {
-			klog.Errorf("Max cluster unschedulable replicas error: %v", err)
-			continue
+
+	// List all unschedulableReplicaEstimators in order of descending priority. The estimators are grouped with different
+	// priorities, e.g: [priority:20, {estimators:[es1, es3]}, {priority:10, estimators:[es2, es4]}, ...]
+	estimatorGroups := estimators.Values()
+
+	// Iterate the estimator groups in order of descending priority
+	for _, estimatorGroup := range estimatorGroups {
+		// if higher-priority estimators have formed a full result of member clusters, no longer to call lower-priority estimator.
+		if len(clusterIndexToEstimatorPriority) == len(undesiredClusterNames) {
+			break
 		}
-		for i := range res {
-			if res[i].Replicas == estimatorclient.UnauthenticReplica {
+		estimatorsWithSamePriority := estimatorGroup.(map[string]estimatorclient.UnschedulableReplicaEstimator)
+		// iterate through these estimators with the same priority.
+		for _, estimator := range estimatorsWithSamePriority {
+			res, err := estimator.GetUnschedulableReplicas(ctx, undesiredClusterNames, reference, unschedulableThreshold)
+			if err != nil {
+				klog.Errorf("Max cluster unschedulable replicas error: %v", err)
 				continue
 			}
-			if undesiredClusters[i].ClusterName == res[i].Name && undesiredClusters[i].Unschedulable > res[i].Replicas {
-				undesiredClusters[i].Unschedulable = res[i].Replicas
+			for i := range res {
+				// the result of this cluster estimated failed, ignore the corresponding result
+				if res[i].Replicas == estimatorclient.UnauthenticReplica {
+					continue
+				}
+				// the cluster name not match, ignore, which hardly ever happens
+				if res[i].Name != undesiredClusters[i].ClusterName {
+					klog.Errorf("unexpected cluster name in the result of estimator with %d priority, "+
+						"expected: %s, got: %s", estimator.Priority(), undesiredClusters[i].ClusterName, res[i].Name)
+					continue
+				}
+				// the result of this cluster has already been estimated by higher-priority estimator,
+				// ignore the corresponding result by this estimator
+				if priority, ok := clusterIndexToEstimatorPriority[i]; ok && estimator.Priority() < priority {
+					continue
+				}
+				// if multiple estimators are called, choose the minimum value of each estimated result,
+				// record the priority of result provider.
+				if res[i].Replicas < undesiredClusters[i].Unschedulable {
+					undesiredClusters[i].Unschedulable = res[i].Replicas
+					clusterIndexToEstimatorPriority[i] = estimator.Priority()
+				}
 			}
 		}
 	}
--- a/pkg/estimator/client/accurate.go
+++ b/pkg/estimator/client/accurate.go
@@ -32,23 +32,25 @@ import (
 
 // RegisterSchedulerEstimator will register a SchedulerEstimator.
 func RegisterSchedulerEstimator(se *SchedulerEstimator) {
-	replicaEstimators["scheduler-estimator"] = se
-	unschedulableReplicaEstimators["scheduler-estimator"] = se
+	registerReplicaEstimator("scheduler-estimator", se)
+	registerUnschedulableReplicaEstimator("scheduler-estimator", se)
 }
 
 type getClusterReplicasFunc func(ctx context.Context, cluster string) (int32, error)
 
 // SchedulerEstimator is an estimator that calls karmada-scheduler-estimator for estimation.
 type SchedulerEstimator struct {
-	cache   *SchedulerEstimatorCache
-	timeout time.Duration
+	cache    *SchedulerEstimatorCache
+	timeout  time.Duration
+	priority EstimatorPriority
 }
 
 // NewSchedulerEstimator builds a new SchedulerEstimator.
-func NewSchedulerEstimator(cache *SchedulerEstimatorCache, timeout time.Duration) *SchedulerEstimator {
+func NewSchedulerEstimator(cache *SchedulerEstimatorCache, timeout time.Duration, priority EstimatorPriority) *SchedulerEstimator {
 	return &SchedulerEstimator{
-		cache:   cache,
-		timeout: timeout,
+		cache:    cache,
+		timeout:  timeout,
+		priority: priority,
 	}
 }
 
@@ -65,6 +67,11 @@ func (se *SchedulerEstimator) MaxAvailableReplicas(
 	return getClusterReplicasConcurrently(parentCtx, clusterNames, se.timeout, func(ctx context.Context, cluster string) (int32, error) {
 		return se.maxAvailableReplicas(ctx, cluster, replicaRequirements.DeepCopy())
 	})
+}
+
+// Priority provides the priority of this estimator.
+func (se *SchedulerEstimator) Priority() EstimatorPriority {
+	return se.priority
 }
 
 // GetUnschedulableReplicas gets the unschedulable replicas which belong to a specified workload by calling karmada-scheduler-estimator.
--- a/pkg/estimator/client/general.go
+++ b/pkg/estimator/client/general.go
@@ -32,15 +32,17 @@ import (
 
 // GeneralEstimator is the default replica estimator.
 func init() {
-	replicaEstimators["general-estimator"] = NewGeneralEstimator()
+	registerReplicaEstimator("general-estimator", NewGeneralEstimator(General))
 }
 
 // GeneralEstimator is a normal estimator in terms of cluster ResourceSummary.
-type GeneralEstimator struct{}
+type GeneralEstimator struct {
+	priority EstimatorPriority
+}
 
 // NewGeneralEstimator builds a new GeneralEstimator.
-func NewGeneralEstimator() *GeneralEstimator {
-	return &GeneralEstimator{}
+func NewGeneralEstimator(priority EstimatorPriority) *GeneralEstimator {
+	return &GeneralEstimator{priority: priority}
 }
 
 // MaxAvailableReplicas estimates the maximum replicas that can be applied to the target cluster by cluster ResourceSummary.
@@ -51,6 +53,11 @@ func (ge *GeneralEstimator) MaxAvailableReplicas(_ context.Context, clusters []*
 		availableTargetClusters[i] = workv1alpha2.TargetCluster{Name: cluster.Name, Replicas: maxReplicas}
 	}
 	return availableTargetClusters, nil
+}
+
+// Priority provides the priority of this estimator.
+func (ge *GeneralEstimator) Priority() EstimatorPriority {
+	return ge.priority
 }
 
 func (ge *GeneralEstimator) maxAvailableReplicas(cluster *clusterv1alpha1.Cluster, replicaRequirements *workv1alpha2.ReplicaRequirements) int32 {
--- a/pkg/estimator/client/interface.go
+++ b/pkg/estimator/client/interface.go
@@ -20,6 +20,8 @@ import (
 	"context"
 	"time"
 
+	"github.com/emirpasic/gods/maps/treemap"
+
 	clusterv1alpha1 "github.com/karmada-io/karmada/pkg/apis/cluster/v1alpha1"
 	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
 )
@@ -30,26 +32,83 @@ import (
 const UnauthenticReplica = -1
 
 var (
-	replicaEstimators              = map[string]ReplicaEstimator{}
-	unschedulableReplicaEstimators = map[string]UnschedulableReplicaEstimator{}
+	// replicaEstimators are organized in a TreeMap, sorted by descending priority.
+	// The key is of type EstimatorPriority, indicating the estimator's priority level.
+	// The value is a map with string keys and ReplicaEstimator values, grouping estimators by their respective priorities.
+	replicaEstimators = treemap.NewWith(estimatorPriorityComparator)
+
+	// unschedulableReplicaEstimators are organized in a TreeMap, sorted by descending priority.
+	// The key is of type EstimatorPriority, indicating the estimator's priority level.
+	// The value is a map with string keys and UnschedulableReplicaEstimator values, grouping estimators by their respective priorities.
+	unschedulableReplicaEstimators = treemap.NewWith(estimatorPriorityComparator)
 )
+
+// registerReplicaEstimator add a estimator to replicaEstimators
+func registerReplicaEstimator(estimatorName string, estimator ReplicaEstimator) {
+	if val, ok := replicaEstimators.Get(estimator.Priority()); !ok {
+		replicaEstimators.Put(estimator.Priority(), map[string]ReplicaEstimator{estimatorName: estimator})
+	} else {
+		estimatorsWithSamePriority := val.(map[string]ReplicaEstimator)
+		estimatorsWithSamePriority[estimatorName] = estimator
+	}
+}
+
+// registerUnschedulableReplicaEstimator add a estimator to unschedulableReplicaEstimators
+func registerUnschedulableReplicaEstimator(estimatorName string, estimator UnschedulableReplicaEstimator) {
+	if val, ok := unschedulableReplicaEstimators.Get(estimator.Priority()); !ok {
+		unschedulableReplicaEstimators.Put(estimator.Priority(), map[string]UnschedulableReplicaEstimator{estimatorName: estimator})
+	} else {
+		estimatorsWithSamePriority := val.(map[string]UnschedulableReplicaEstimator)
+		estimatorsWithSamePriority[estimatorName] = estimator
+	}
+}
 
 // ReplicaEstimator is an estimator which estimates the maximum replicas that can be applied to the target cluster.
 type ReplicaEstimator interface {
 	MaxAvailableReplicas(ctx context.Context, clusters []*clusterv1alpha1.Cluster, replicaRequirements *workv1alpha2.ReplicaRequirements) ([]workv1alpha2.TargetCluster, error)
+	Priority() EstimatorPriority
 }
 
 // UnschedulableReplicaEstimator is an estimator which estimates the unschedulable replicas which belong to a specified workload.
 type UnschedulableReplicaEstimator interface {
 	GetUnschedulableReplicas(ctx context.Context, clusters []string, reference *workv1alpha2.ObjectReference, unschedulableThreshold time.Duration) ([]workv1alpha2.TargetCluster, error)
+	Priority() EstimatorPriority
 }
 
 // GetReplicaEstimators returns all replica estimators.
-func GetReplicaEstimators() map[string]ReplicaEstimator {
+func GetReplicaEstimators() *treemap.Map {
 	return replicaEstimators
 }
 
 // GetUnschedulableReplicaEstimators returns all unschedulable replica estimators.
-func GetUnschedulableReplicaEstimators() map[string]UnschedulableReplicaEstimator {
+func GetUnschedulableReplicaEstimators() *treemap.Map {
 	return unschedulableReplicaEstimators
+}
+
+// EstimatorPriority the priority of estimator
+//  1. If two estimators are of the same priority, call both and choose the minimum value of each estimated result.
+//  2. If higher-priority estimators have formed a full result of member clusters, no longer to call lower-priority estimator.
+//  3. If higher-priority estimators haven't given the result for certain member clusters, lower-priority estimator will
+//     continue to estimate for such clusters haven't got a result.
+type EstimatorPriority int32
+
+const (
+	// General general priority, e.g: ResourceModel
+	General EstimatorPriority = 10
+	// Accurate accurate priority, e.g: SchedulerEstimator
+	Accurate EstimatorPriority = 20
+)
+
+// estimatorPriorityComparator provides a basic comparison on EstimatorPriority.
+func estimatorPriorityComparator(a, b interface{}) int {
+	aAsserted := a.(EstimatorPriority)
+	bAsserted := b.(EstimatorPriority)
+	switch {
+	case aAsserted > bAsserted:
+		return -1
+	case aAsserted < bAsserted:
+		return 1
+	default:
+		return 0
+	}
 }
--- a/pkg/scheduler/core/util.go
+++ b/pkg/scheduler/core/util.go
@@ -52,7 +52,11 @@ func getDefaultWeightPreference(clusters []*clusterv1alpha1.Cluster) *policyv1al
 }
 
 func calAvailableReplicas(clusters []*clusterv1alpha1.Cluster, spec *workv1alpha2.ResourceBindingSpec) []workv1alpha2.TargetCluster {
+	// availableTargetClusters stores the result of estimated replicas for each clusters
 	availableTargetClusters := make([]workv1alpha2.TargetCluster, len(clusters))
+	// clusterIndexToEstimatorPriority key refers to index of cluster slice,
+	// value refers to the EstimatorPriority of who gave its estimated result.
+	clusterIndexToEstimatorPriority := make(map[int]estimatorclient.EstimatorPriority)
 
 	// Set the boundary.
 	for i := range availableTargetClusters {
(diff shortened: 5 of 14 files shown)


chaunceyjiang:
Just a question, will this modification require compatibility? Will it force users to upgrade the Estimator component?

chaosi-zju:
> will this modification require compatibility

I don't think it matters. What do you think?

> Will it force users to upgrade the Estimator component

I think it's only about the scheduler, and doesn't require upgrading estimator

chaosi-zju:
CC @RainbowMango may be ready to merge, any further comments?

pkg/scheduler/core/util.go
```diff
@@ -79,11 +88,21 @@ func calAvailableReplicas(clusters []*clusterv1alpha1.Cluster, spec *workv1alpha
...
 		for i := range res {
+			if res[i].Name != availableTargetClusters[i].Name {
+				continue
+			}
 			if res[i].Replicas == estimatorclient.UnauthenticReplica {
 				continue
 			}
-			if availableTargetClusters[i].Name == res[i].Name && availableTargetClusters[i].Replicas > res[i].Replicas {
+			// means there are already replicas calculated by higher-priority estimator
+			if estimator.GetPriority() < availableResultPriority[i] {
```
whitewindmills, line 98: if `SchedulerEstimator` give the result first, can we prevent calling `GeneralEstimator`?

whitewindmills: So I think a better solution is to sort all estimators in descending order of priority, call each estimator in turn, and only when the current estimator call fails we call the lower priority estimator.

pkg/estimator/client/interface.go
```diff
@@ -53,3 +54,16 @@ func GetReplicaEstimators() map[string]ReplicaEstimator {
...
 }
+
+// EstimatorPriority
+// 1. Higher-priority estimator can directly override the result given by lower-priority estimator
+// 2. Lower replicas estimated will be chosen if the priority of estimator are same
+// 3. Lower-priority estimator would be ignored if the result was already given by higher-priority estimator
+type EstimatorPriority int
+
+const (
+	Undefined EstimatorPriority = iota
```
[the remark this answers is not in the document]
chaosi-zju: mainly because:

https://github.com/karmada-io/karmada/blob/1cd8b0b1be4b29f0a0c7b8a02c750803776d8537/pkg/scheduler/core/util.go#L68-L70

this slice has default value 0, I don't want it equal to any valid priority.

may be I can remove `Undefined` priority and  change above line as:

```go
for i := range availableResultPriority { 
 	availableResultPriority[i] = -1
 } 
```

do you perfer I deleting the `Undefined` priority?

pkg/scheduler/core/util.go
chaosi-zju: sorry, I didn't got your mean, what does ` give the result first` mean?

chaosi-zju: I mean we can save time on several estimator calls？

pkg/estimator/client/interface.go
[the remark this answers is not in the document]
whitewindmills: deleting the `Undefined` priority is ok. BTW, could you consider designing priorities to be more scalable? if I want to add a new estimator which priority is between `SchedulerEstimator` and `GeneralEstimator`, how should I expand it?

pkg/scheduler/core/util.go
whitewindmills: right, when `SchedulerEstimator` has calculated the result, do you think it is still necessary to call `GeneralEstimator`?

pkg/estimator/client/interface.go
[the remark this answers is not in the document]
chaosi-zju: > if I want to add a new estimator which priority is between SchedulerEstimator and GeneralEstimator, how should I expand it?

You are absolutely right, thank you very much for your advice~

pkg/scheduler/core/util.go
chaosi-zju: you are right~

whitewindmills: I think the lower priority estimator is just a fallback option rather than a must-have.

chaosi-zju:
@whitewindmills sorry for delayed a while, I have updated the code, do you have time checking once more?

pkg/estimator/client/interface.go
```diff
@@ -30,26 +30,40 @@ import (
...
+// EstimatorPriority
+// 1. If higher-priority estimator has given the result, no longer to call lower-priority estimator.
+// 2. If two estimators are of the same priority, call both and choose the minimum value of each estimated result.
+// 3. If an estimator can only give a part of result of member clusters, same priority estimator will be called, but low-priority estimator not.
+type EstimatorPriority int32
+
+const (
+	Rough    EstimatorPriority = 100
+	General  EstimatorPriority = 200
+	Accurate EstimatorPriority = 300
```
whitewindmills, line 68: Do these numbers mean anything? If not, you can define a default priority in the range 0 to 100.

pkg/estimator/client/interface.go
```diff
@@ -30,26 +30,40 @@ import (
...
 }
+
+// EstimatorPriority
+// 1. If higher-priority estimator has given the result, no longer to call lower-priority estimator.
+// 2. If two estimators are of the same priority, call both and choose the minimum value of each estimated result.
+// 3. If an estimator can only give a part of result of member clusters, same priority estimator will be called, but low-priority estimator not.
+type EstimatorPriority int32
+
+const (
+	Rough    EstimatorPriority = 100
```
whitewindmills, line 66: We don't need `Rough` yet.

pkg/estimator/client/interface.go
chaosi-zju: No special meaning, but intentionally enumerated from the hundreds, so that if you want to add a priority between the General and Accurate levels, just give him a value between 200 and 300.

pkg/scheduler/core/util.go
```diff
@@ -68,20 +69,33 @@ func calAvailableReplicas(clusters []*clusterv1alpha1.Cluster, spec *workv1alpha
...
+			break
+		}
 		res, err := estimator.MaxAvailableReplicas(ctx, clusters, spec.ReplicaRequirements)
 		if err != nil {
 			klog.Errorf("Max cluster available replicas error: %v", err)
 			continue
 		}
+		priorityOfAvailableEstimator = estimator.GetPriority()
 		for i := range res {
+			// if an estimator can only give a part of result of member clusters, same priority estimator will be called, but low-priority estimator not.
```
whitewindmills, line 94: I think low-priority estimator may be called. We would rather choose lower precision results than not return results at all. And I noticed that there are some corner cases here.
### Eg:
The priority of estimators(`A, B, C`) are `3, 3, 2`.
* Estimator `A` only gave a part of result of member clusters, `C` should be called when`B` failed or gave the partial result.
* Estimator `A` gave the full result of member clusters, `C` shouldn't be called when `B` failed or gave the partial result.
* Estimator `A` and `B` both gave the partial result, but when they can form a complete result, `C` shouldn't be called.

chaosi-zju: @whitewindmills If you want to cover all corner cases, things can be more complex, like:

The priority of estimators(A, B, C, D) are 3, 3, 2, 2.

* Estimator A and B both gave the partial result and can't form a complete result, but after call C, they form a complete result, however D has the same priority with C, now should D be called?

whitewindmills: Of course `D` should be called. We cannot guarantee that all corner cases will be covered. That's just what I can think of right now. From another perspective, this PR may not need to be considered so completely, just follow the simple fallback principle.

pkg/estimator/client/interface.go
whitewindmills: Only when the corresponding estimator exists, we add its priority. And The definition of this field seems very vague at present. Maybe we will never use this kind of imprecise estimator, so please remove it.

pkg/scheduler/core/util.go
chaosi-zju: done, thanks

pkg/estimator/client/interface.go
chaosi-zju: done

pkg/descheduler/core/helper.go
```diff
@@ -62,24 +63,39 @@ func NewSchedulingResultHelper(binding *workv1alpha2.ResourceBinding) *Schedulin
...
+			break
+		}
 		res, err := estimator.GetUnschedulableReplicas(ctx, undesiredClusterNames, reference, unschedulableThreshold)
 		if err != nil {
 			klog.Errorf("Max cluster unschedulable replicas error: %v", err)
 			continue
 		}
+		priorityOfAvailableEstimator = estimator.GetPriority()
 		for i := range res {
+			// if an estimator can only give a part of result of member clusters, same priority estimator will be called.
```
whitewindmills, line 94:
```go
// if an estimator only gave a part of result of member clusters, the next estimator will be called.
```

pkg/descheduler/core/helper.go
```diff
@@ -95,6 +111,16 @@ func (h *SchedulingResultHelper) FillUnschedulableReplicas(unschedulableThreshol
 	klog.V(4).Infof("Target undesired cluster of unschedulable replica result: %s", pretty.Sprint(undesiredClusters))
 }
 
+// clustersFullyEstimated whether estimators has gaven a full result of member clusters
```
whitewindmills, line 114:
```go
// clustersFullyEstimated checks whether estimators has gaven a full result of member clusters.
```

pkg/scheduler/core/util.go
```diff
@@ -68,20 +69,33 @@ func calAvailableReplicas(clusters []*clusterv1alpha1.Cluster, spec *workv1alpha
...
+			break
+		}
 		res, err := estimator.MaxAvailableReplicas(ctx, clusters, spec.ReplicaRequirements)
 		if err != nil {
 			klog.Errorf("Max cluster available replicas error: %v", err)
 			continue
 		}
+		priorityOfAvailableEstimator = estimator.GetPriority()
 		for i := range res {
+			// if an estimator can only give a part of result of member clusters, same priority estimator will be called.
```
whitewindmills, line 94:
```go
// if an estimator only gave a part of result of member clusters, the next estimator will be called.
```

pkg/scheduler/core/util.go
```diff
@@ -100,6 +114,16 @@ func calAvailableReplicas(clusters []*clusterv1alpha1.Cluster, spec *workv1alpha
 	return availableTargetClusters
 }
 
+// ClustersFullyEstimated whether estimators has gaven a full result of member clusters
```
whitewindmills, line 117:
```go
// ClustersFullyEstimated checks whether estimators have gaven a full result of member clusters.
```

chaosi-zju: done, thanks

whitewindmills:
thanks for your hard work.
/lgtm

chaosi-zju:
CC @RainbowMango

pkg/scheduler/core/util.go
```diff
@@ -68,20 +69,33 @@ func calAvailableReplicas(clusters []*clusterv1alpha1.Cluster, spec *workv1alpha
...
+		return estimators[i].GetPriority() > estimators[j].GetPriority()
+	})
+
+	priorityOfAvailableEstimator := estimatorclient.EstimatorPriority(math.MinInt32)
 	ctx := context.WithValue(context.TODO(), util.ContextKeyObject,
 		fmt.Sprintf("kind=%s, name=%s/%s", spec.Resource.Kind, spec.Resource.Namespace, spec.Resource.Name))
+
 	for _, estimator := range estimators {
+		// if higher-priority estimators have formed a full result of member clusters, no longer to call lower-priority estimator.
+		if estimator.GetPriority() < priorityOfAvailableEstimator && clustersFullyEstimated(availableTargetClusters) {
```
RainbowMango, line 84: Does that mean we have estimators with the same priority?
Otherwise, we don't need to compare the priority in range loops as the priority in different loops must have different priorities.

pkg/estimator/client/interface.go
```diff
@@ -30,26 +30,38 @@ import (
...
-	replicaEstimators              = map[string]ReplicaEstimator{}
-	unschedulableReplicaEstimators = map[string]UnschedulableReplicaEstimator{}
+	replicaEstimators              []ReplicaEstimator
+	unschedulableReplicaEstimators []UnschedulableReplicaEstimator
 )
 
 // ReplicaEstimator is an estimator which estimates the maximum replicas that can be applied to the target cluster.
 type ReplicaEstimator interface {
 	MaxAvailableReplicas(ctx context.Context, clusters []*clusterv1alpha1.Cluster, replicaRequirements *workv1alpha2.ReplicaRequirements) ([]workv1alpha2.TargetCluster, error)
+	GetPriority() EstimatorPriority
```
RainbowMango, line 40: In my opinion, a estimator should not declare priority for `itself`, the priority is reflected by consumers how to use it.

RainbowMango (commented):
It makes sense to me that `accurate estimator` should bypass `generall estimator` as in some cases, like mis-config of resource module, the general estimator might get lower result, we should trust the precise result( from accurate estimator).

But the current implementation is concerning to me.
cc owner @Garrybest to help take a look.

pkg/estimator/client/interface.go
Garrybest: Maybe we could decide the default priority? I think accurate scheduler estimator should not be enabled along with other estimators.

RainbowMango: Yeah, I agree. the accurate scheduler estimator can bypass the general estimator by default.

But there is a corner case we need to consider:
What if some of the clusters do not enable the `karmada-estimator`?  Or, what if some of the `karmada-estimator`s can not estimate the replicas at that time(such as timeout)? Should the scheduler back off to `general estimator`?

Garrybest: Got it, this is possible. Fall back off to general estimator could be a nice choice.

chaosi-zju: I have a new perspective here, but I haven't had time to update and implement it:

if the higher-priority estimators have gaven results for all members clusters, we don't need to call lower-priority estimators.

if the higher-priority estimators only gave part of result for members clusters, which means some of the member clusters didn't get result, we then call lower-priority estimators, but its result only apply to member clusters which didn't get result.

I mean those member clusters which already been successfully calculated by `karmada-estimator`, no need back off to `general`, back off only aims at those not be successfully calculated.

RainbowMango: Yes, for the cluster which already estimated by its `karmada-estimator` no need to back off to `general estimator`.

chaosi-zju: done

chaosi-zju:
@whitewindmills @RainbowMango @Garrybest 

I have improved the logic as:

* priority is not declared by estimator itself, but by the function caller, just like:

```go
// NewGeneralEstimator builds a new GeneralEstimator.
func NewGeneralEstimator(priority EstimatorPriority) *GeneralEstimator {
	return &GeneralEstimator{priority: priority}
}

// GetPriority return the priority of this estimator
func (ge *GeneralEstimator) GetPriority() EstimatorPriority {
	return ge.priority
}
```

* If higher-priority estimators haven't given the result for certain member clusters, lower-priority estimator **only continue to estimate for such clusters haven't got a result**. Just like for the cluster which already estimated by its `karmada-estimator` no need to back off to `general estimator`.

Sorry for delayed, and thanks for review again~

pkg/descheduler/core/helper.go
```diff
@@ -62,26 +63,52 @@ func NewSchedulingResultHelper(binding *workv1alpha2.ResourceBinding) *Schedulin
...
+		priorityOfLastAvailableEstimator = estimator.GetPriority()
 		for i := range res {
+			// the result of this cluster estimated failed, ignore the corresponding result
 			if res[i].Replicas == estimatorclient.UnauthenticReplica {
 				continue
 			}
-			if undesiredClusters[i].ClusterName == res[i].Name && undesiredClusters[i].Unschedulable > res[i].Replicas {
+			// the cluster name not match, ignore, which hardly ever happens
+			if res[i].Name != undesiredClusters[i].ClusterName {
+				continue
```
whitewindmills, line 102: If not matched, we need to add logs to locate the problem because something goes wrong in the estimator.

chaosi-zju: got it

pkg/estimator/client/accurate.go
```diff
@@ -67,6 +85,11 @@ func (se *SchedulerEstimator) MaxAvailableReplicas(
 	})
 }
 
+// GetPriority return the priority of this estimator
```
whitewindmills, line 88: // GetPriority provides the priority of this estimator.

chaosi-zju: done

pkg/scheduler/core/util.go
```diff
@@ -68,22 +72,46 @@ func calAvailableReplicas(clusters []*clusterv1alpha1.Cluster, spec *workv1alpha
...
 	}
 
-	// Get the minimum value of MaxAvailableReplicas in terms of all estimators.
+	// get estimators and sort it by priority.
 	estimators := estimatorclient.GetReplicaEstimators()
+	sort.Slice(estimators, func(i, j int) bool {
+		return estimators[i].GetPriority() > estimators[j].GetPriority()
+	})
+
+	priorityOfLastAvailableEstimator := estimatorclient.EstimatorPriority(math.MinInt32)
```
Garrybest, line 81: I think the solution is a little bit complicated. Here is a simple solution:

1. Group the estimators with different priorities, e.g., 
[{estimators:[es1, es3], priority:9}, {estimators:[es2, es4], priority:8}, ...]
2. Loop these estimators group, do the same previous caculation. if the group is returned, skip the rest ones.

pkg/estimator/client/general.go
```diff
@@ -32,15 +32,17 @@ import (
 
 // GeneralEstimator is the default replica estimator.
 func init() {
-	replicaEstimators["general-estimator"] = NewGeneralEstimator()
+	replicaEstimators = append(replicaEstimators, NewGeneralEstimator(General))
```
Garrybest, line 35: It's upset that golang does not have data structure like rbtree, otherwise we don't need to change `replicaEstimators` to slice.

pkg/estimator/client/accurate.go
```diff
@@ -32,23 +32,41 @@ import (
 
 // RegisterSchedulerEstimator will register a SchedulerEstimator.
 func RegisterSchedulerEstimator(se *SchedulerEstimator) {
-	replicaEstimators["scheduler-estimator"] = se
-	unschedulableReplicaEstimators["scheduler-estimator"] = se
+	replicaEstimators = append(replicaEstimators, se)
```
Garrybest, line 35: I suggest check if "scheduler-estimator" exists in slice.

As a result, we don't need to add `UnRegisterSchedulerEstimator` any more because I see it is only used in test files which could be redundant.

Garrybest (changes requested):

pkg/estimator/client/general.go
chaosi-zju: feel upset too

pkg/scheduler/core/util.go
chaosi-zju: Thanks for the suggestion, but I don't think it's getting any better, it leads to one more structure definition and one more layer of loop iteration.

chaosi-zju: Complex structures or more loops can also complicate the problem.

pkg/estimator/client/accurate.go
chaosi-zju: I actually think there's a flaw here too

chaosi-zju: > we don't need to add UnRegisterSchedulerEstimator any more because I see it is only used in test files which could be redundant

I also think there's a flaw here

pkg/estimator/client/general.go
whitewindmills: good idea, kindly refer [PR](https://github.com/karmada-io/karmada/pull/4555) about rbTree.

pkg/estimator/client/accurate.go
chaosi-zju: > I suggest check if "scheduler-estimator" exists in slice.

But, how can I judge "scheduler-estimator" exists in slice, this may requires that `type ReplicaEstimator interface` add a `GetName()` function.

chaosi-zju: Or, we do not judge "scheduler-estimator" exists in slice, we just:

```go
var once sync.Once

// RegisterSchedulerEstimator will register a SchedulerEstimator.
func RegisterSchedulerEstimator(se *SchedulerEstimator) {
	once.Do(func() {
		replicaEstimators = append(replicaEstimators, se)
		unschedulableReplicaEstimators = append(unschedulableReplicaEstimators, se)
	})
}
```

pkg/estimator/client/general.go
chaosi-zju: @whitewindmills If we change `replicaEstimators []ReplicaEstimator` to `replicaEstimators *rbt.Tree`,

how can I prevent `func RegisterSchedulerEstimator(se *SchedulerEstimator)` function been called multiple times in `descheduler_test.go`

chaosi-zju: Beside, it can not store object in same priority if the key is priority:

```go
tree := rbt.NewWithIntComparator() // empty (keys are of type int)

tree.Put(1, "x") // 1->x
tree.Put(2, "b") // 1->x, 2->b (in order)
tree.Put(1, "a") // 1->a, 2->b (in order, replacement)
```

pkg/scheduler/core/util.go
```diff
@@ -68,22 +72,46 @@ func calAvailableReplicas(clusters []*clusterv1alpha1.Cluster, spec *workv1alpha
...
 	ctx := context.WithValue(context.TODO(), util.ContextKeyObject,
 		fmt.Sprintf("kind=%s, name=%s/%s", spec.Resource.Kind, spec.Resource.Namespace, spec.Resource.Name))
+
 	for _, estimator := range estimators {
+		// if higher-priority estimators have formed a full result of member clusters, no longer to call lower-priority estimator.
+		if estimator.GetPriority() < priorityOfLastAvailableEstimator && len(clusterIndexToEstimatorPriority) == len(clusters) {
```
RainbowMango, line 87: This is the part where I feel complicated.
If we can make sure the estimator is looped by priority, then each estimator can focus on clusters that haven't been estimated.

pkg/estimator/client/accurate.go
```diff
@@ -32,23 +32,41 @@ import (
...
+	unschedulableReplicaEstimators = append(unschedulableReplicaEstimators, se)
+}
+
+// UnRegisterSchedulerEstimator will unregister a SchedulerEstimator.
+func UnRegisterSchedulerEstimator(se *SchedulerEstimator) {
```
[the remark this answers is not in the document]
chaosi-zju: the test is in another package, `replicaEstimators` can not been accessed by it.

pkg/scheduler/core/util.go
chaosi-zju: @Garrybest suggest grouping the estimators with different priorities, e.g.,[{estimators:[es1, es3], priority:9}, {estimators:[es2, es4], priority:8}, ...]

@whitewindmills suggest using `github.com/emirpasic/gods/trees/redblacktree`

@RainbowMango suggest looping by priority

So, combine your three suggestions, how about using `github.com/emirpasic/gods/trees/redblacktree`, the key is `priority`, the value is `map[string]estimators{}` with same priority, and loop by `priority` key?

chaosi-zju: actually, `github.com/emirpasic/gods/maps/treemap` is more suitable.

Garrybest: Hey guys, I think we have a misunderstanding here. It's not about rbtree or slice, both are OK for me. Let me give a example:

```
	// sort by priority
	sort.Slice(estimators, func(i, j int) bool {
		return estimators[i].GetPriority() > estimators[j].GetPriority()
	})

	ctx := context.WithValue(context.TODO(), util.ContextKeyObject,
		fmt.Sprintf("kind=%s, name=%s/%s", spec.Resource.Kind, spec.Resource.Namespace, spec.Resource.Name))

	// group by priority
	var estimatorsByPriority [][]estimatorclient.ReplicaEstimator
	for i, j := 0, 0; i < len(estimators); {
		for ; j < len(estimators); j++ {
			if estimator[j].GetPriority() != estimators[i].GetPriority() {
				break
			}
		}
		estimatorsByPriority = append(estimatorsByPriority, estimators[i:j])
		i = j
	}

loop:
	for _, estimatorGroup := range estimatorsByPriority {
		for _, estimator := range estimatorGroup {
			res, err := estimator.MaxAvailableReplicas(ctx, clusters, spec.ReplicaRequirements)
			if err != nil {
				klog.Errorf("Max cluster available replicas error: %v", err)
				continue
			}
			for i := range res {
				if res[i].Replicas == estimatorclient.UnauthenticReplica {
					continue
				}
				if availableTargetClusters[i].Name == res[i].Name && availableTargetClusters[i].Replicas > res[i].Replicas {
					availableTargetClusters[i].Replicas = res[i].Replicas
				}
			}
			// estimator with higher priority finished, get the result and skip other ones
			break loop
		}
	}
```

Just group them and add one more loop. It is better?

pkg/scheduler/core/util.go
chaosi-zju: > Group the estimators with different priorities, e.g., [{estimators:[es1, es3], priority:9}, {estimators:[es2, es4], priority:8}, ...]. Loop these estimators group, do the same previous caculation. if the group is returned, skip the rest ones.

done, I have applied this strategy.

the nits is there are triple nested for loop now.

pkg/estimator/client/accurate.go
chaosi-zju: done, I midified the type of `replicaEstimators`, and the redundant `UnRegisterSchedulerEstimator` function is removed.

pkg/estimator/client/accurate.go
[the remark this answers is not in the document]
chaosi-zju: done, removed, now it is no longer needed.

pkg/scheduler/core/util.go
chaosi-zju: I actually understand it that way, but I've implemented it using a treemap to make code more clean.

The reason I'm torn between using a treemap instead of a redblackmap is that the redblackmap is not convenient to get values by key.

pkg/scheduler/core/util.go
```diff
@@ -68,22 +72,51 @@ func calAvailableReplicas(clusters []*clusterv1alpha1.Cluster, spec *workv1alpha
...
-		res, err := estimator.MaxAvailableReplicas(ctx, clusters, spec.ReplicaRequirements)
-		if err != nil {
-			klog.Errorf("Max cluster available replicas error: %v", err)
-			continue
+
+	// List all replicaEstimators in order of descending priority. The estimators are grouped with different priorities,
+	// e.g: [priority:20, {estimators:[es1, es3]}, {priority:10, estimators:[es2, es4]}, ...]
+	estimatorGroups := replicaEstimators.Values()
+
+	// Iterate the estimator groups in order of descending priority
```
whitewindmills, line 84: From the perspective of code implementation, this is just a simple multi-level loop, and I don't think grouping has any advantages but increases code complexity (two-level loops become three-level loops, although the total number of loops remains unchanged). So the main reason is that I don't see how the advantages of grouping are used.

pkg/estimator/client/interface.go
```diff
@@ -30,26 +32,83 @@ import (
...
+	// unschedulableReplicaEstimators are organized in a TreeMap, sorted by descending priority.
+	// The key is of type EstimatorPriority, indicating the estimator's priority level.
+	// The value is a map with string keys and UnschedulableReplicaEstimator values, grouping estimators by their respective priorities.
+	unschedulableReplicaEstimators = treemap.NewWith(estimatorPriorityComparator)
 )
 
+// registerReplicaEstimator add a estimator to replicaEstimators
+func registerReplicaEstimator(estimatorName string, estimator ReplicaEstimator) {
+	if val, ok := replicaEstimators.Get(estimator.Priority()); !ok {
+		replicaEstimators.Put(estimator.Priority(), map[string]ReplicaEstimator{estimatorName: estimator})
```
whitewindmills, line 49: Is it necessary to use `map[string]ReplicaEstimator` here? I see that the key of map is ignored when using it. Maybe the slice  `[]ReplicaEstimator` is enough.

pkg/scheduler/core/util.go
chaosi-zju: @Garrybest @RainbowMango This is just what I memtioned above that `Complex structures or more loops can also complicate the problem.`

This is consistent with my initial view, but since it has been changed, I don't want to change it back. What do you all think?

pkg/estimator/client/interface.go
chaosi-zju: No, this can prevent the `RegisterSchedulerEstimator` from being called repeatedly. 

One obvious issue is, the test function `TestDescheduler_worker` will fail since it call `RegisterSchedulerEstimator` for many times, if slice used, there will be redundant and dirty estimators, while if map used, it can help with deduplication. 

This is also why `UnRegisterSchedulerEstimator ` be added in my previous commit, since the map is introduced, the `UnRegisterSchedulerEstimator` is needless.

whitewindmills: We don’t have to compromise code implementation because of the test function `TestDescheduler_worker` but modify the test code.
```go
        // register estimator outside the loop.
        estimatorCache := estimatorclient.NewSchedulerEstimatorCache()
	schedulerEstimator := estimatorclient.NewSchedulerEstimator(estimatorCache, 5*time.Second)
	estimatorclient.RegisterSchedulerEstimator(schedulerEstimator)

	for _, tt := range tests {
			ctx, cancel := context.WithCancel(context.TODO())
			defer cancel()

			binding, err := buildBinding(tt.args.name, tt.args.namespace, tt.args.target, tt.args.status)
			if err != nil {
				t.Fatalf("build binding error: %v", err)
			}

			karmadaClient := fakekarmadaclient.NewSimpleClientset(binding)
			factory := informerfactory.NewSharedInformerFactory(karmadaClient, 0)

			desched := &Descheduler{
				KarmadaClient:           karmadaClient,
				informerFactory:         factory,
				bindingInformer:         factory.Work().V1alpha2().ResourceBindings().Informer(),
				bindingLister:           factory.Work().V1alpha2().ResourceBindings().Lister(),
				clusterInformer:         factory.Cluster().V1alpha1().Clusters().Informer(),
				clusterLister:           factory.Cluster().V1alpha1().Clusters().Lister(),
				schedulerEstimatorCache: estimatorCache,
				unschedulableThreshold:  5 * time.Minute,
				eventRecorder:           record.NewFakeRecorder(1024),
			}
```

chaosi-zju: I haven't closely examined whether the test code you provided works, but the essence of the problem isn't about compromising with the test code. The test code merely exposes this issue, what I expect is that even in normal code, issues of duplicate registration will also be avoided.
Besides, the advantages of slice over map are not that significant.

PR #4400, opened 2023-12-11
add expand replicas test about even distribution of replicas

### LINKED ISSUE #4220: enhancement of dividing replicas by static weight evenly

chaosi-zju:
**What type of PR is this?**

/kind failing-test

add expand replicas test about even distribution of replicas

Fixes #4220

```release-note

```

### DIFF
--- a/pkg/scheduler/core/generic_scheduler_test.go
+++ b/pkg/scheduler/core/generic_scheduler_test.go
@@ -28,49 +28,51 @@ import (
 )
 
 type testcase struct {
-	name      string
-	clusters  []*clusterv1alpha1.Cluster
-	object    *workv1alpha2.ResourceBindingSpec
-	placement *policyv1alpha1.Placement
-	// changeCondForNextSchedule following cases will schedule twice, this function aims to change the schedule condition for second schedule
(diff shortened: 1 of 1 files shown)


chaosi-zju:
CC @XiShanYongYe-Chang

chaosi-zju:
how about now? CC @XiShanYongYe-Chang

XiShanYongYe-Chang (commented):
Thanks a lot~
/lgtm
/approve

PR #4422, opened 2023-12-14
skip recreating resources that is not supposed to be propagated

jwcesign:
**What type of PR is this?**
/kind bug

1.K8s create multiple endpointslice for one service.
2.Karmada collects these endpointslice to control plane as work.
3.K8s delete some endpointslice.
4.Karmad work_status_controller.go will detect this and recreate it.

We expect the endpointslice will be deleted correctly.

Fixes #none

none

```release-note
`karmada-controller-manager`: Fixed the issue that always trying to re-create resources that should not be propagated.
```

### DIFF
--- a/pkg/controllers/status/work_status_controller.go
+++ b/pkg/controllers/status/work_status_controller.go
@@ -276,6 +276,10 @@ func (c *WorkStatusController) handleDeleteEvent(key keys.FederatedKey) error {
 		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


jwcesign:
BTW, this problem exists for ServiceImport/Export

RainbowMango:
/retitle skip recreating resources that is not supposed to be propagated

RainbowMango:
> BTW, this problem exists for ServiceImport/Export

I guess this patch also fixed this issue. Please correct me if I'm wrong.

RainbowMango (approved):
/lgtm
/approve

PR #4423, opened 2023-12-14
fix: Update RB once new clusters are added to trigger the propagation of service to these new clusters

jwcesign:
**What type of PR is this?**
/kind bug

When new clusters are joined, we should propagate the service to the new clusters.

Before, if the providerClusters/consumerClusters are empty, there was no modification even new clusters are joined.
In this scenario, the scheduler will not schedule the service to the newly joined clusters.

So we should update the RB to trigger it.

Fixes #none

none

```release-note

```


RainbowMango (approved):
/lgtm
/approve

PR #4427, opened 2023-12-15
prevent CI from fast fail when just one k8s version e2e failed

chaosi-zju:
**What type of PR is this?**

/kind flake

We now run e2e test on multi releases version of Kubernetes, if e2e test failed at just one releases version, the other e2e test will be interrupted by fast fail strategy.

It is not conducive to check the root cause of failure, so it is advised to disable fast fail strategy.

Fixes #

```release-note

```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -112,6 +112,7 @@ jobs:
     needs: build
     runs-on: ubuntu-22.04
     strategy:
+      fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


XiShanYongYe-Chang:
I don't quite understand why disabling `fail-fast` can make it easier to quickly locate problems. Can you explain the reason a little bit?

RainbowMango (approved):
/lgtm
/approve

Issue #4036, opened 2023-09-06
[umbrella] support promote dependent resources automatically


zhy76:
**What would you like to be added**:
Provide an automatic promotion mechanism for dependent resources in karmadactl. When promoting a resource, all the resources that it depends on will be automatically promoted as well. For example, promoting the Secret that is dependent by a Deployment.

**Why is this needed**:
Currently, karmadactl promote only support promote resources one by one, which is inconvenient.

**Related issue**:
#3842 #1862 

**Task Decomposition**:

- [ ] support promote native dependent resources automatically.
- [ ] support promote thirdparty dependent resources automatically.
- [ ] support promote CRD dependent resources automatically.
    - [ ] implement webhook interpreter adapter to get dependencies.
    - [ ] use configuration interpreter to get dependencies.
- [ ] E2E tests.
- [ ] Doc update. 

> Tips: Task splitting and details may change as features are developed.

PR #4445, opened 2023-12-18
feat: support port-forward for access webhook interpreter pod in cluster

zhy76:
**What type of PR is this?**
/kind feature

feat: support port-forward for access webhook interpreter pod in cluster

The portforward implementation here will still temporarily open a random port locally to transmit webhook access requests. Transferring data directly to the established portforward connection is a bit complicated and I haven't been able to successfully implement it yet, maybe it can be optimized in the future.

Part of #4036 

cc @jwcesign 

```release-note
karmadactl: support port-forward for access webhook interperter pod in cluster
```

### DIFF
--- a/pkg/karmadactl/promote/portforward.go
+++ b/pkg/karmadactl/promote/portforward.go
@@ -1,0 +1,560 @@
+/*
+Copyright 2023 The Karmada Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package promote
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"net/url"
+	"os"
+	"path"
+	"sort"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/httpstream"
+	"k8s.io/apimachinery/pkg/util/runtime"
+	webhookutil "k8s.io/apiserver/pkg/util/webhook"
+	v1 "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/transport/spdy"
+	"k8s.io/klog/v2"
+	"k8s.io/kube-aggregator/pkg/apiserver"
+	utiltrace "k8s.io/utils/trace"
+
+	configv1alpha1 "github.com/karmada-io/karmada/pkg/apis/config/v1alpha1"
+	"github.com/karmada-io/karmada/pkg/karmadactl/util/apiclient"
+	"github.com/karmada-io/karmada/pkg/resourceinterpreter/customized/webhook/configmanager"
+	"github.com/karmada-io/karmada/pkg/resourceinterpreter/customized/webhook/request"
+	"github.com/karmada-io/karmada/pkg/util/fedinformer/genericmanager"
+	interpreterutil "github.com/karmada-io/karmada/pkg/util/interpreter"
+)
+
+// PortForwardProtocolV1Name is the subprotocol used for port forwarding.
+const PortForwardProtocolV1Name = "portforward.k8s.io"
+const LocalHost = "127.0.0.1"
+
+// customizedInterpreter interpret custom resource with webhook configuration.
+type customizedInterpreter struct {
+	// hookManager caches all webhook configurations.
+	hookManager configmanager.ConfigManager
+	// clientManager builds REST clients to talk to webhooks.
+	clientManager *webhookutil.ClientManager
+}
+
+// newCustomizedInterpreter return a new CustomizedInterpreter.
+func newCustomizedInterpreter(informer genericmanager.SingleClusterInformerManager, serviceLister v1.ServiceLister) (*customizedInterpreter, error) {
+	cm, err := webhookutil.NewClientManager(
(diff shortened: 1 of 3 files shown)


pkg/karmadactl/promote/promote.go
```diff
@@ -19,6 +19,7 @@ package promote
 import (
 	"context"
 	"fmt"
+	"github.com/karmada-io/karmada/pkg/resourceinterpreter/customized/webhook/request"
```
zhzhuang-zju, line 22:
```shell
diff --git a/pkg/karmadactl/promote/promote.go b/pkg/karmadactl/promote/promote.go
index 1f8b8fdf..5471e9c3 100644
--- a/pkg/karmadactl/promote/promote.go
+++ b/pkg/karmadactl/promote/promote.go
@@ -19,7 +19,6 @@ package promote
 import (
        "context"
        "fmt"
-       "github.com/karmada-io/karmada/pkg/resourceinterpreter/customized/webhook/request"
        "os"
 
        "github.com/spf13/cobra"
@@ -50,6 +49,7 @@ import (
        "github.com/karmada-io/karmada/pkg/karmadactl/util"
        "github.com/karmada-io/karmada/pkg/resourceinterpreter/customized/declarative"
        "github.com/karmada-io/karmada/pkg/resourceinterpreter/customized/webhook"
+       "github.com/karmada-io/karmada/pkg/resourceinterpreter/customized/webhook/request"
        "github.com/karmada-io/karmada/pkg/resourceinterpreter/default/native"
        "github.com/karmada-io/karmada/pkg/resourceinterpreter/default/native/prune"
        "github.com/karmada-io/karmada/pkg/resourceinterpreter/default/thirdparty"
```
 Hi @zhy76 The goimport is not formatted and therefore also fails ```ci```

zhy76: Fixed, Thanks~

pkg/karmadactl/promote/portforward.go
```diff
@@ -0,0 +1,560 @@
...
+	defer pf.streamConn.Close()
+
+	rep, err := pf.forward(ctx, wh)
+	return rep, err
+}
+
+// Close stops the listener of PortForwarder.
+func (pf *PortForwarder) Close() {
+	// stop the listener
+	if err := pf.listener.Close(); err != nil {
```
zhzhuang-zju, line 302: @zhy76 Panic may occur here! If an error is returned on line 291, the close of the listener here is triggered. But because line 295 haven't been reached, the listener has not been initialized yet.

zhzhuang-zju:
> 8. Promote the workload cr in member1 to the control plane using karmadactl and portforward excute:
> 
> ```shell
> ./karmadactl promote workload nginx -C member1 --kubeconfig=/root/.kube/karmada.config --portforward-context=karmada-host --portforward-kubeconfig=/root/.kube/karmada.config --portforward-namespace=karmada-system --portforward-pod=karmada-interpreter-webhook-example-69b895cbc6-h7njb --portforward-port=8445 -d -p
> ```
 Hi @zhy76 Can you help explain how `--portforward-pod=karmada-interpreter-webhook-example-69b895cbc6-h7njb` came about? I didn't find it in the steps you reproduced, thank you.

zhzhuang-zju:
> Hi @zhy76 Can you help explain how `--portforward-pod=karmada-interpreter-webhook-example-69b895cbc6-h7njb` came about? I didn't find it in the steps you reproduced, thank you.

Don't worry about it, I figured it out

zhzhuang-zju:
Hello, thanks for your important commit! Except for the previous review, I didn't find other problems. But I have some questions about the implementation.
1.Currently, the kubeconfig of the karmada host cluster is used, which may be difficult to meet in a production environment. I tried port-forwarding the service under the `karmada apiserver` to replace the direct port-forward pod, but an error was reported `Service is defined without a selector`. Referring to https://github.com/kubernetes/website/pull/8028/files. The path I tried may not work.
2.Is it feasible to put the service under the `karmada host` to replace the direct port-forward pod? The pod name may change and is not suitable as an input parameter.

zhy76:
> 2.Is it feasible to put the service under the karmada host to replace the direct port-forward pod? The pod name may change and is not suitable as an input parameter.

Thanks for your suggestion and review. I think it's doable, I will update the code.

jwcesign:
```
./karmadactl promote workload nginx -C member1 --kubeconfig=/root/.kube/karmada.config --portforward-context=karmada-host --portforward-kubeconfig=/root/.kube/karmada.config --portforward-namespace=karmada-system --portforward-pod=karmada-interpreter-webhook-example-69b895cbc6-h7njb --portforward-port=8445 -d -p
```

Some users may not have the privilege to access the Karmada host cluster, so this may not be a feasible approach.

There are two ways:
1. There is another component, and it's the backend of one apiservice, and karmadactl will call this apiservice to parse the dependencies. (Need another component)

2. Support the command `karmadactl promote xxx --webhookurl=http://xxx/interpret` to enable external access to the webhook, such as when the webhook service is of LB type.

3. Do not support promoting the dependencies if the resource has a webhook interpreter(karmadactl should give the warning information). 

I prefer method 3.  

/cc @zhzhuang-zju @XiShanYongYe-Chang @zhy76 @RainbowMango

zhy76:
> 3. Do not support promoting the dependencies if the resource has a webhook interpreter(karmadactl should give the warning information).

Does it means if the user does not have the privilege to access the Karmada host cluster, and if the resource has a webhook interpreter, do not support promoting the dependencies and give a warning info by karmadactl? I prefer this one too.

jwcesign:
> Does it means if the user does not have the privilege to access the Karmada host cluster, and if the resource has a webhook interpreter, do not support promoting the dependencies and give a warning info by karmadactl? I prefer this one too.

When installing Karmada, the namespace may not be karmada-system, also, the users may not have the privilege to access the host clusters.

So I think it's better just not to support parsing the dependencies with webhook.

zhzhuang-zju:
> So I think it's better just not to support parsing the dependencies with webhook.

I agree

zhy76:
So can we just use karmadactl's default restconfig instead of letting the user specify portforward-context and portforward-config, and let the user use the command: `./karmadactl promote workload nginx -C member1 --kubeconfig=/root/.kube/karmada. config  --portforward-namespace=karmada-system(If the user has permission to the cluster, he should know the namespace that corresponds to interpret pod/svc) —portforward-obj=svc(or pod)/karmada-interpreter-webhook-example-69b895cbc6-h7njb --portforward-port=8445 -d -p`. If the user does not have the corresponding permission, the service cannot be found, and a warning is displayed. This allows users with permissions to automatically promote the resource dependency, while users without permissions report a warning.

jwcesign:
/cc @RainbowMango 
For suggestions

PR #4446, opened 2023-12-18
cleanup(mcs): do not overwrite the original service if the service exists

jwcesign:
**What type of PR is this?**
/kind bug

When one service exists in the member cluster.
If users use MCS to propagate the service, Karmada may not overwrite the service.

In the future, we will introduce an API field to let users configure the behavior.

Fixes none

none

```release-note
`karmada-controller-manager`: do not overwrite the original service if the service exists
```

### DIFF
--- a/pkg/controllers/multiclusterservice/mcs_controller.go
+++ b/pkg/controllers/multiclusterservice/mcs_controller.go
@@ -444,7 +444,7 @@ func (c *MCSController) buildResourceBinding(svc *corev1.Service, mcs *networkin
 		},
 		Spec: workv1alpha2.ResourceBindingSpec{
 			Placement:          placement,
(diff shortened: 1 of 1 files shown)


pkg/controllers/multiclusterservice/mcs_controller.go
```diff
@@ -444,7 +444,7 @@ func (c *MCSController) buildResourceBinding(svc *corev1.Service, mcs *networkin
 		},
 		Spec: workv1alpha2.ResourceBindingSpec{
 			Placement:          placement,
-			ConflictResolution: policyv1alpha1.ConflictOverwrite,
+			ConflictResolution: policyv1alpha1.ConflictAbort,
```
XiShanYongYe-Chang, line 447: How about adding a TODO to describe the todo things?

PR #4453, opened 2023-12-19
execute InterpretHealth even if statusRaw is nil

### LINKED ISSUE #4430: The service can't be propagated to the newly joined cluster  if the cluster is unjoined 1 minute agao
**What happened**:
1.Launch Karmada with hack/local-up-karmada.sh
2.Deploying multiclusterservice with the following yaml
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx
spec:
  replicas: 2
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: hello-app
        image: gcr.io/google-samples/hello-app:1.0
        ports:
        - containerPort: 8080
          protocol: TCP
---
apiVersion: v1
kind: Service
metadata:
  name: nginx
spec:
  ports:
  - port: 81
    targetPort: 8080
  selector:
    app: nginx
---
kind: PropagationPolicy
metadata:
  name: mci-workload
spec:
  resourceSelectors:
    - apiVersion: apps/v1
      kind: Deployment
      name: nginx
  placement:
    clusterAffinity:
      clusterNames:
        - member1
---
apiVersion: networking.karmada.io/v1alpha1
kind: MultiClusterService
metadata:
   name: nginx
spec:
  types:
    - CrossCluster
```
3.Check service
image

They are propagated to member clusters.

4.Unjoin cluster member2

cluster_controller.go will add following taints to cluster object.
```yaml
      min: 1Ti
      name: memory
  secretRef:
    name: member2
    namespace: karmada-cluster
  syncMode: Push
  taints:
  - effect: NoExecute
    key: cluster.karmada.io/terminating
    timeAdded: "2023-12-15T07:09:41Z"
status:
  apiEnablements:
  - groupVersion: admissionregistration.k8s.io/v1
```

5.The taint manager will add EvicationTask to RB.
image

6.The gracefulevication controller could not parse the service's health status, the EvicationTask will be kept for 10min.
https://github.com/karmada-io/karmada/blob/31b83d14e4067352a99d225c0b27989378f5cb6e/pkg/controllers/gracefuleviction/evictiontask.go#L112

7.When the member2 cluster joins again, the scheduler will filter it out because of EvicationTask.

The service will be propagated to member2 cluster

As described before

None

**Environment**:
- Karmada version:
- Others:

chaosi-zju:
**What type of PR is this?**

/kind bug

continue to execute InterpretHealth even if statusRaw is nil, otherwise, the graceful eviction controller could not parse the service's health status, the EvictionTask will be kept for 10min, details described in https://github.com/karmada-io/karmada/issues/4430

Fixes #4430

Test Report：

```yaml
status:
  aggregatedStatus:
  - applied: true
    clusterName: member1
    health: Healthy
  - applied: true
    clusterName: member2
    health: Healthy
  - applied: true
    clusterName: member3
    health: Healthy
  conditions:
  - lastTransitionTime: "2023-12-20T07:52:02Z"
    message: Binding has been scheduled successfully.
    reason: Success
    status: "True"
    type: Scheduled
  - lastTransitionTime: "2023-12-20T07:52:05Z"
    message: All works have been successfully applied
    reason: FullyAppliedSuccess
    status: "True"
    type: FullyApplied
```

image

```release-note
continue to execute InterpretHealth even if statusRaw is nil when reflectStatus 
```

### DIFF
--- a/pkg/controllers/status/work_status_controller.go
+++ b/pkg/controllers/status/work_status_controller.go
@@ -312,10 +312,6 @@ func (c *WorkStatusController) reflectStatus(work *workv1alpha1.Work, clusterObj
 	}
 	c.EventRecorder.Eventf(work, corev1.EventTypeNormal, events.EventReasonReflectStatusSucceed, "Reflect status for object(%s/%s/%s) succeed.", clusterObj.GetKind(), clusterObj.GetNamespace(), clusterObj.GetName())
 
-	if statusRaw == nil {
-		return nil
-	}
-
 	var resourceHealth workv1alpha1.ResourceHealth
 	// When an unregistered resource kind is requested with the ResourceInterpreter,
 	// the interpreter will return an error, we treat its health status as Unknown.


chaosi-zju:
CC @XiShanYongYe-Chang @jwcesign

jwcesign (commented):
/lgtm

XiShanYongYe-Chang (commented):
Thanks~
/lgtm
/cc @RainbowMango

pkg/controllers/status/work_status_controller.go
```diff
@@ -312,23 +312,28 @@ func (c *WorkStatusController) reflectStatus(work *workv1alpha1.Work, clusterObj
...
 	var resourceHealth workv1alpha1.ResourceHealth
-	// When an unregistered resource kind is requested with the ResourceInterpreter,
-	// the interpreter will return an error, we treat its health status as Unknown.
-	healthy, err := c.ResourceInterpreter.InterpretHealth(clusterObj)
-	if err != nil {
-		resourceHealth = workv1alpha1.ResourceUnknown
-		c.EventRecorder.Eventf(work, corev1.EventTypeWarning, events.EventReasonInterpretHealthFailed, "Interpret health of object(%s/%s/%s) failed, err: %s.", clusterObj.GetKind(), clusterObj.GetNamespace(), clusterObj.GetName(), err.Error())
-	} else if healthy {
+	if statusRaw == nil {
+		// When a resource didn't implement ReflectStatus and its status field didn't collect from member cluster,
```
chaunceyjiang, line 317: > When a resource didn't implement ReflectStatus and its status field didn't collect from member cluster,

We have corresponding implementations for all resources. `reflectWholeStatus`

RainbowMango: I don't quite get it.
Do you mean you also take `empty` status as healthy? But implementing it in resource interpreter?

RainbowMango: I get it, you mean that all resource status will be collected by `reflectWholeStatus`:
https://github.com/karmada-io/karmada/blob/88bed31b92e1de6297be845c2284d96482f14e0e/pkg/resourceinterpreter/default/native/default.go#L151

chaosi-zju: hi @chaunceyjiang, two questions:

1. how you handle resource like `clusterip` type `Service`, it exist in `e.reflectStatusHandlers[object.GroupVersionKind()]` but will return nil status in `reflectServiceStatus` now.
2. you mean you expect not return nil status when reflect no status filed resources, then may be should also have a default `InterpretHealth` function for this resources?

RainbowMango:
After revisiting the implementation of `InterpretHealth`:
https://github.com/karmada-io/karmada/blob/88bed31b92e1de6297be845c2284d96482f14e0e/pkg/resourceinterpreter/default/native/default.go#L154-L163

We only provided default implementations for a few `workload` resource types, which is reasonable as most of `configuration` resource types like `Secret, ConfigMap, RBAC, and so on` might not need to be evaluated.

So, let's assume that resources that do not implement the `InterpretHealth` interface are considered `healthy` by default. 
In addition, the `InterpretHealth` was designed for fail-over purposes, we shouldn't require all resource types to implement the interface.  What do you think?  @XiShanYongYe-Chang @chaunceyjiang  Does that make sense to you?

chaunceyjiang:
> So, let's assume that resources that do not implement the InterpretHealth interface are considered healthy by default.
In addition, the InterpretHealth was designed for fail-over purposes, we shouldn't require all resource types to implement the interface


+1

pkg/controllers/status/work_status_controller.go
chaosi-zju: changed to `continue to execute InterpretHealth even if statusRaw is nil`

chaosi-zju:
update: change implemetion to `continue to execute InterpretHealth even if statusRaw is nil`

XiShanYongYe-Chang:
> update: change implemetion to `continue to execute InterpretHealth even if statusRaw is nil`

/lgtm

Please help update the title.

chaosi-zju:
update: re-tested okay.

jwcesign:
I think we need release-notes

chaosi-zju:
> I think we need release-notes

done

RainbowMango:
> update: change implemetion to continue to execute InterpretHealth even if statusRaw is nil

That makes sense to me. No matter whether the status is empty or not, just let `InterpretHealth` to evaluate the healthy status.

I understand this PR focuses on fixing the issue tracked by #4430, and the improvement discussed above(https://github.com/karmada-io/karmada/pull/4453#issuecomment-1863933988) should be tracked by another issue/PR.

RainbowMango (approved):
/lgtm
/approve

Issue #4292, opened 2023-11-21
[umbrella] Add Implementation for Service discovery with native Kubernetes naming and resolution


jwcesign:
**What would you like to be added**:
In https://github.com/karmada-io/karmada/pull/4287, we propose a way to implement service discovery with native Kubernetes naming and resolution.

The specific work items are as follows:

release-1.8(1130)
- [ ] Add proposal for ` service discovery with native Kubernetes naming and resolution` - https://github.com/karmada-io/karmada/pull/4287
- [ ] Add corresponding API - https://github.com/karmada-io/karmada/pull/4290 
- [ ] For mcs-controller, List&watch mcs and service, reconcile the work in execution namespace @Rains6 
- [ ] For mcs-eps-controller, List&watch mcs, collect the corresponding EndpointSlice from `ServerLocations` and create/update the corresponding Work. @jwcesign 

release-1.9(1215)
- [ ] For mcs-controller, List&watch cluster creation/deletion, reconcile the work in corresponding cluster execution namespace.
- [ ] For mcs-eps-controller, List&watch cluster creation/deletion, reconcile the EndpointSlice's work in corresponding cluster execution namespace.
- [ ] If cluster gets unhealth, mcs-eps-controller should delete the EndpointSlice from all the cluster execution namespace. 

**Why is this needed**:

Ref proposal https://github.com/karmada-io/karmada/pull/4287

Help

Anyone who wants to participate in one of the tasks can come and assign the tasks. Feel free to let me know in the comments and I will mark it after the corresponding task item, thanks~

/help

jwcesign:
Now, the EPS synchronization logic is as follows:
![image](https://github.com/karmada-io/karmada/assets/19745536/3b071db4-6ada-44e1-9d41-48c1dcd8624e)

1.List&watch mcs, build informer to list&watch the eps in `ServiceProvisionClusters`
2.List&watch eps in ServiceProvisionClusters
3.create the corresponding work, the name rule is: `mcs-{endpointslice name}-{hash}`
4.Sync the endpointslice's work in step 3 to `ServiceConsumptionClusters`, the endpoint slice name will be changed to `{source-cluster-name}-{endpointslcie-name}`
5.The work will trigger to create the endpointslice int `ServiceConsumptionClusters`.

Affan-7:
Hi @jwcesign, can you please assign me something to work on?

PR #4454, opened 2023-12-19
feat: prevent updates to mcs.types or when multiple types are involved.

jwcesign:
**What type of PR is this?**
/kind feature

Currently, the north-south traffic routing is implemented through ServiceImport/ServiceExport by collecting EndpointSlice from member clusters to Karmada control plane for Work resources. 

The east-west MCS will also collect EndpointSlice for Work, and both sides will jointly manage the overlay of Work operations.

If both types are configured at the same time, when the north-south Work is overlaid (mcs label is deleted), the east-west MCS cannot synchronize the provider cluster's endpointslice to consumer clusters.

This PR temporarily prohibits mixed configuration and will resolve related conflict issues in future updates.

Part of https://github.com/karmada-io/karmada/issues/4292

none

```release-note
`karmada-webhook`: prevent updates to mcs.types or when multiple types are involved.
```

### DIFF
--- a/pkg/webhook/multiclusterservice/validating.go
+++ b/pkg/webhook/multiclusterservice/validating.go
@@ -19,6 +19,7 @@ package multiclusterservice
 import (
 	"context"
 	"net/http"
+	"reflect"
 	"strings"
 
 	admissionv1 "k8s.io/api/admission/v1"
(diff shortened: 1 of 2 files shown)


RainbowMango (commented):
/assign

pkg/webhook/multiclusterservice/validating.go
```diff
@@ -97,12 +108,19 @@ func (v *ValidatingAdmission) validateMultiClusterServiceSpec(mcs *networkingv1a
...
 	typesPath := specPath.Child("types")
 	for i := range mcs.Spec.Types {
 		typePath := typesPath.Index(i)
 		exposureType := mcs.Spec.Types[i]
+		typesSet.Insert(string(exposureType))
 		allErrs = append(allErrs, v.validateExposureType(&exposureType, typePath)...)
 	}
+	if len(typesSet) > 1 {
+		allErrs = append(allErrs, field.Invalid(typesPath, mcs.Spec.Types, "MultiClusterService types should not contain more than one type"))
+	}
```
RainbowMango, line 122: Maybe we don't need to introduce another set, how about this:
```suggestion
	// temporarily disable two expose types
	if len(mcs.Spec.Types) > 1 {
		allErrs = append(allErrs, field.Invalid(typesPath, mcs.Spec.Types, "MultiClusterService types should not contain more than one type"))
	}
```

RainbowMango (commented):
I understand this is a temporary protection measure to avoid conflicts between `CrossCluster`(MCS) and `Loadbalancer`(MCI). However, this conflict issue should be resolved in the release-1.9 cycle.

cc @GitHubxsy @chaunceyjiang Please look at this, need your comments here.

pkg/webhook/multiclusterservice/validating.go
jwcesign: mcs.Spec.Types can be configured as {LoadBalancer, LoadBalancer}. In this situation, it's the same type.

jwcesign: BTW, we should deduplicate, but in another PR.

GitHubxsy:
lgtm

pkg/webhook/multiclusterservice/validating.go
jwcesign: /cc @RainbowMango

XiShanYongYe-Chang: We need to consider whether to prohibit users from filling in the same type of element. My idea is that it should be prohibited.

RainbowMango (approved):
/lgtm
/approve

PR #3762, opened 2023-07-05
Change "delete all clusters" to "delete related clusters" in local_up_karmada.sh

LronDC:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:
#3707 
**Which issue(s) this PR fixes**:
Fixes #3707 

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #4462, opened 2023-12-21
only delete necessary clusters when installed by local-up

zhzhuang-zju:
**What type of PR is this?**
/kind cleanup

Sometimes, we create our own k8s clusters through kind in the local environment, but they are deleted when karmada is installed through local-up.
Fixes #

```release-note
NONE
```


RainbowMango:
I remember we did something similar before, but I forgot which PR now. I'll try to find it out to revist.

liangyuanpeng:
Since this is a user-visible change, I think it's worth writing a release note.

> I remember we did something similar before, but I forgot which PR now. I'll try to find it out to revist.

@RainbowMango @zhzhuang-zju 
I think it is https://github.com/karmada-io/karmada/pull/3762

zhzhuang-zju:
cc @liangyuanpeng @village-way The resource need to be deleted are used as input to the function, so that the function is a bit more generic

hack/util.sh
```diff
@@ -464,6 +464,36 @@ function util::delete_all_clusters() {
...
+  local config_file_arr=$(echo $config_files| tr ',' ' ')
+  local cluster_arr=$(echo $clusters| tr ',' ' ')
+  local log_file="${log_path}"/delete-necessary-resources.log
+  rm -rf ${log_file}
+  mkdir -p ${log_path}
+
+  for config_file in $config_file_arr
+  do
+    rm -f $config_file
+  done
```
chaosi-zju, line 487: there doesn't need `for config_file in $config_file_arr`, just `rm ${config_file_arr}`, which equals to `rm ~/.kube/karmada.config ~/.kube/members.config`

hack/util.sh
```diff
@@ -464,6 +464,36 @@ function util::delete_all_clusters() {
...
+
+  for cluster in $cluster_arr
```
chaosi-zju, line 489: ditto, no need `for cluster in $cluster_arr`, just `kind delete clusters member1 member2 member3`

chaosi-zju (commented):
thank you for your important commit! Review is ongoing, and my viewpoint just for reference only.

hack/util.sh
zhzhuang-zju: done~thanks for your review

hack/util.sh
```diff
@@ -464,6 +464,26 @@ function util::delete_all_clusters() {
...
+  local config_files=${1}
+  local clusters=${2}
+  local log_path=${3}
+
+  local log_file="${log_path}"/delete-necessary-resources.log
+  rm -rf ${log_file}
+  mkdir -p ${log_path}
+
+  rm -f config_file
+  kind delete clusters $clusters >> "${log_file}" 2>&1
```
chaosi-zju, line 483: This modification introduces new problem:

* `rm -f config_file`, what does `config_file` refer to? you mean `rm -rf ${config_files}` ?

* the previous tr ',' ' ' needs to be retained, `rm -rf ~/.kube/karmada.config,~/.kube/members.config` is invalid, while `rm -rf ~/.kube/karmada.config ~/.kube/members.config` is valid, you known what I mean?

chaosi-zju, line 483: ditto, `kind delete clusters karmada-host,member1,member2,member3` is invalid, while `kind delete clusters karmada-host member1 member2 member3` is valid.

chaosi-zju, line 483: For code style consistency, I suggest `$clusters` transform to `${clusters}`, to keep the same variable style as others.

liangyuanpeng: +1

hack/util.sh
```diff
@@ -464,6 +464,26 @@ function util::delete_all_clusters() {
   echo "Deleted all clusters and the log file is in ${log_file}"
 }
 
+# util::delete_necessary_resources deletes clusters(karmada-host, member1, member2 and member3) and related resources directly
+# util::delete_necessary_resourcesactually do three things: delete cluster、remove kubeconfig、record delete log
```
liangyuanpeng, line 468:
```suggestion
# util::delete_necessary_resources actually do three things: delete cluster、remove kubeconfig、record delete log
```

zhzhuang-zju:
cc @liangyuanpeng @chaosi-zju

hack/util.sh
```diff
@@ -464,6 +464,28 @@ function util::delete_all_clusters() {
...
+  local clusters=${2}
+  local log_path=${3}
+
+  local log_file="${log_path}"/delete-necessary-resources.log
+  rm -rf ${log_file}
+  mkdir -p ${log_path}
+  local config_file_arr=$(echo ${config_files}| tr ',' ' ')
+  local cluster_arr=$(echo ${clusters}| tr ',' ' ')
+
+  rm -f ${config_file_arr}
```
chaosi-zju, line 484: why above `rm -rf ${log_file}` has a `-r` option, while this line `rm -f ${config_file_arr}` not.

Is there any purpose?

zhzhuang-zju: There is no special intention, just to be consistent with the previous function.
https://github.com/karmada-io/karmada/blob/e7192a9ecc7bd01f1ef103411bdb78a5cd8fcddc/hack/util.sh#L458-L469

chaosi-zju: I think it is not a good habit, if it which you removed is a dirtory, append `-r`; if it is just a file, no need.

hack/local-up-karmada.sh
```diff
@@ -112,7 +112,7 @@ echo -e "Preparing kindClusterConfig in path: ${TEMP_PATH}"
 cp -rf "${REPO_ROOT}"/artifacts/kindClusterConfig/member1.yaml "${TEMP_PATH}"/member1.yaml
 cp -rf "${REPO_ROOT}"/artifacts/kindClusterConfig/member2.yaml "${TEMP_PATH}"/member2.yaml
 
-util::delete_all_clusters "${MAIN_KUBECONFIG}" "${MEMBER_CLUSTER_KUBECONFIG}" "${KIND_LOG_FILE}"
```
RainbowMango, line 115: Should we remove the unused function `util::delete_all_clusters`? @chaosi-zju

chaosi-zju: yes

RainbowMango (approved):
/lgtm
/approve

PR #4463, opened 2023-12-21
Bump Kubernetes dependencies to v1.28.5

RainbowMango:
**What type of PR is this?**

/kind cleanup

This PR updates Kubernetes dependencies to v1.28.5.

Fixes #

This PR adopted the following changes:
- https://github.com/kubernetes/kubernetes/pull/116648
- https://github.com/kubernetes/kubernetes/pull/119207

```release-note
Kubernetes dependencies now have been bumped to v1.28.5.
```


pkg/search/proxy/store/store.go
```diff
@@ -191,6 +191,11 @@ func (s *store) Count(string) (int64, error) {
 	return 0, fmt.Errorf("count is not supported in proxy store")
 }
 
+// RequestWatchProgress implements storage.Interface.
+func (s *store) RequestWatchProgress(context.Context) error {
+	return fmt.Errorf("not implemented")
+}
+
```
RainbowMango, line 198: @XiShanYongYe-Chang @ikaven1024 @xigang 
Can you help to confirm if it OK to leave this interface unimplemented?
This interface was introduced by https://github.com/kubernetes/kubernetes/pull/119207.

hack/tools/swagger/lib/render.go
```diff
@@ -80,6 +83,8 @@ func RenderOpenAPISpec(cfg Config) (string, error) {
 	}
 
 	serverConfig := genericapiserver.NewRecommendedConfig(cfg.Codecs)
+	serverConfig.ClientConfig = &clientgorest.Config{}
+	_ = utilfeature.DefaultMutableFeatureGate.Set(fmt.Sprintf("%s=%v", features.APIPriorityAndFairness, false))
```
RainbowMango, line 87: @chaosi-zju 
Can you help to confirm if it is OK? and help to confirm the generated `swagger.json`?

pkg/search/proxy/store/store.go
XiShanYongYe-Chang: By examining the calling path of the code, confirm that the current return of error has no impact. 

Test in the environment, and verify that the search function is working properly.

RainbowMango: Thanks.

PR #4465, opened 2023-12-21
Add kubernetes v1.29.0 to CI

liangyuanpeng:
**What type of PR is this?**

/kind cleanup

/hold

kubernetes 1.29 has been released, but there may be no need to rush to add 1.29 to CI. This is currently a placeholder.

Fixes #

```release-note
NONE
```

### DIFF
--- a/.github/workflows/lint-chart.yaml
+++ b/.github/workflows/lint-chart.yaml
@@ -3,9 +3,9 @@ name: Chart Lint
 
 env:
   HELM_VERSION: v3.11.2
-  KIND_VERSION: v0.14.0
-  KIND_NODE_IMAGE: kindest/node:v1.27.3
-  K8S_VERSION: v1.26.4
+  KIND_VERSION: v0.20.0
+  KIND_NODE_IMAGE: kindest/node:v1.29.0
+  K8S_VERSION: v1.29.0
 
 on:
   push:
--- a/.github/workflows/ci-schedule-compatibility.yaml
+++ b/.github/workflows/ci-schedule-compatibility.yaml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        kubeapiserver-version: [ v1.21.10, v1.22.7, v1.23.4, v1.24.2, v1.25.0, v1.26.0, v1.27.3, v1.28.0 ]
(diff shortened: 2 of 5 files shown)


RainbowMango:
Should we enable v1.29 in the following tests?

https://github.com/karmada-io/karmada/blob/bff8aa90fb5ed19f155388227e69065beb5bffcd/.github/workflows/cli.yaml#L21
https://github.com/karmada-io/karmada/blob/bff8aa90fb5ed19f155388227e69065beb5bffcd/.github/workflows/ci-schedule-compatibility.yaml#L16
https://github.com/karmada-io/karmada/blob/bff8aa90fb5ed19f155388227e69065beb5bffcd/.github/workflows/ci-schedule.yml#L15

liangyuanpeng:
> Should we enable v1.29 in the following tests?

Absolutely, But I think we may not have to rush to replace v1.26, so I just simply modify the CI workflow and hold on this PR.

We can add v1.29 to the scheduled task first, and then add v1.29 to `workflows/ci.yaml` and  `workflows/cli.yaml`  after a period of time.

What do you think?  Another option is to add v1.29 to CI immediately and delete v1.26.
@RainbowMango

RainbowMango:
> Absolutely, But I think we may not have to rush to replace v1.26, so I just simply modify the CI workflow and hold on this PR.

What's your concern?

liangyuanpeng:
I'm worried that the iteration speed may be too fast, but obviously every version faces such a problem, because the iteration speed of the Kubernetes project is like this.

So i update all of place to v1.29.0

.github/workflows/lint-chart.yaml
```diff
@@ -3,9 +3,9 @@ name: Chart Lint
 
 env:
   HELM_VERSION: v3.11.2
-  KIND_VERSION: v0.14.0
-  KIND_NODE_IMAGE: kindest/node:v1.27.3
-  K8S_VERSION: v1.26.4
+  KIND_VERSION: v0.20.0
+  KIND_NODE_IMAGE: kindest/node:v1.29.0
+  K8S_VERSION: v1.29.0
```
liangyuanpeng, line 8: This PR would not trigger this CI, so i have standalone trigger CI for this change and CI is green, see: https://github.com/karmada-io/karmada/actions/runs/7308593108/job/19915359313

RainbowMango (approved):
/lgtm
/approve
Thanks.

PR #4473, opened 2023-12-25
bump golang.org/x/tools to v0.16.1

zhzhuang-zju:
**What type of PR is this?**
/kind cleanup

Bumping tools to include the fix for a nil pointer deref error in go/types. See https://github.com/golang/go/issues/64812 for more details.
This fix may be needed for when we bump to go1.22. See https://github.com/kubernetes/kubernetes/pull/122412 for more details.

Fixes #

```release-note
none
```


liangyuanpeng (approved):
This is a fix for golang 1.22 and karmada currently uses 1.20, but a version upgrade will not hurt.

RainbowMango:
> This fix may be needed for when we bump to go1.22. See https://github.com/kubernetes/kubernetes/pull/122412 for more details.

Does that mean the `golang.org/x/tools` before v0.16.1 is not compatible with go1.22?

zhzhuang-zju:
> Does that mean the `golang.org/x/tools` before v0.16.1 is not compatible with go1.22?

They will cause panic in certain scenarios, but whether Karmada will encounter these scenarios remains to be known through experiments.

RainbowMango:
The root cause is the implementation of `SizeFor` has been adjusted in Go1.22, which is not compatible with `golang.org/x/tools` before v0.16.1.

See [Go1.22 release notes](https://tip.golang.org/doc/go1.22):
> The implementation of [SizesFor](https://tip.golang.org/pkg/go/types#SizesFor) has been adjusted to compute the same type sizes as the compiler when the compiler argument for SizesFor is "gc". The default [Sizes](https://tip.golang.org/pkg/go/types#Sizes) implementation used by the type checker is now types.SizesFor("gc", "amd64").

RainbowMango (approved):
/lgtm
/approve

PR #4437, opened 2023-12-15
Improve cache with structed objects

ikaven1024:
**What type of PR is this?**
/kind flake

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:
Fixes https://github.com/karmada-io/karmada/issues/4405#issuecomment-1853367471

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:
```release-note
`karmada-search`: improve with structed objects in cache.
```

PR #4479, opened 2023-12-27
if watch search cache request accept first is protobuf,watch.go will …

niuyueyang1996:
…message does not implement the protobuf marshalling interface and cannot be encoded to a protobuf message,watcher will close. my client watch will quickly close. so provide some resource Unstructured to pbMarshal.

like pod/node

**What type of PR is this?**

Fixes #

```release-note

```

### DIFF
--- a/pkg/search/proxy/store/multi_cluster_cache.go
+++ b/pkg/search/proxy/store/multi_cluster_cache.go
@@ -67,10 +67,10 @@ type cachedResourcesInfo struct {
 }
 
 var _ Store = &MultiClusterCache{}
-var scheme = runtime.NewScheme()
+var encodeScheme = runtime.NewScheme()
 
 func init() {
-	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
+	utilruntime.Must(clientgoscheme.AddToScheme(encodeScheme))
 }
 
 // NewMultiClusterCache return a cache for resources from member clusters
@@ -141,7 +141,7 @@ func (c *MultiClusterCache) UpdateCache(resourcesByCluster map[string]map[schema
 		if err != nil {
 			continue
 		}
-		_, err = scheme.New(kind)
+		_, err = encodeScheme.New(kind)
 		if err != nil {
 			continue
 		}
@@ -417,7 +417,7 @@ func encodeObjectWithPb(encodePb bool, gvr schema.GroupVersionResource, e watch.
 		if ok {
 			obj = cacheObj.GetObject()
 		}
-		newObj, err := scheme.UnsafeConvertToVersion(obj, &schema.GroupVersion{
+		newObj, err := encodeScheme.UnsafeConvertToVersion(obj, &schema.GroupVersion{
 			Group:   gvr.Group,
 			Version: gvr.Version,
 		})
--- a/pkg/search/proxy/store/multi_cluster_cache.go
+++ b/pkg/search/proxy/store/multi_cluster_cache.go
@@ -25,14 +25,13 @@ import (
 	"sync"
 	"time"
 
-	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
-
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metainternalversion "k8s.io/apimachinery/pkg/apis/meta/internalversion"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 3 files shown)


niuyueyang1996:
cache list() also have this question.
Unstructured List does not implement the protobuf marshalling interface and cannot be encoded to a protobuf message.

XiShanYongYe-Chang:
Hi @niuyueyang1996 , Thanks for your contribution!
/assign

May I ask if you have already been using Karmada in the production environment?

niuyueyang1996:
> Hi @niuyueyang1996 , Thanks for your contribution! /assign
> 
> May I ask if you have already been using Karmada in the production environment?

not yet, this is the plan for the next stage.

pkg/search/proxy/store/multi_cluster_cache.go
```diff
@@ -121,11 +132,31 @@ func (c *MultiClusterCache) UpdateCache(resourcesByCluster map[string]map[schema
...
 		}
 	}
+
+	// update cachedResource encode
+	c.updateCachedResourcesEncode()
 	return nil
 }
+func (c *MultiClusterCache) updateCachedResourcesEncode() {
+	for gvr, v := range c.cachedResources {
+		kind, err := c.restMapper.KindFor(gvr)
```
XiShanYongYe-Chang, line 145: How about renaming `kind` to `gvk`?

pkg/search/proxy/store/multi_cluster_cache.go
```diff
@@ -121,11 +132,31 @@ func (c *MultiClusterCache) UpdateCache(resourcesByCluster map[string]map[schema
...
+		}
+		_, err = encodeScheme.New(kind)
+		if err != nil {
+			continue
+		}
+		if !v.encode {
+			klog.V(1).Infof("gvr %v will encode with kind %+v", gvr, kind)
+			v.encode = true
+			c.cachedResources[gvr] = v
+		}
```
XiShanYongYe-Chang, line 157: Can you add some comments to explain why it is handled this way?

pkg/search/proxy/store/multi_cluster_cache.go
```diff
@@ -322,6 +353,20 @@ func (c *MultiClusterCache) List(ctx context.Context, gvr schema.GroupVersionRes
...
 }
 
+func watchUseProtobuf(ctx context.Context) bool {
+	acceptValues := ctx.Value("Accept")
+	if acceptValues != nil {
+		avs := acceptValues.([]string)
+		for _, v := range avs {
+			vs := strings.Split(v, ",")
+			if vs[0] == "application/vnd.kubernetes.protobuf" {
+				return true
```
XiShanYongYe-Chang, line 363: There seem to be some default conventions here. Can you explain why only the element at index 0 is being checked?

pkg/search/proxy/store/multi_cluster_cache.go
```diff
@@ -355,16 +400,39 @@ func (c *MultiClusterCache) Watch(ctx context.Context, gvr schema.GroupVersionRe
 		if err != nil {
 			return nil, err
 		}
-
-		mux.AddSource(w, func(e watch.Event) {
+		encodePb := c.cachedResources[gvr].encode && usePb
+		mux.AddSource(w, func(e watch.Event) watch.Event {
```
XiShanYongYe-Chang, line 404: I don't quite understand why the return value needs to be modified. The handling of resources should directly affect `e.Object`, am I right?

XiShanYongYe-Chang (commented):
Hi @niuyueyang1996, sorry for replying so late. I have some questions.

And can you help make the title more intuitive so that it is easier to understand? You can provide a detailed description of the problem in the PR description. In addition, I saw multiple commits with the same description, can you help squash your commits?

XiShanYongYe-Chang:
Hi @ikaven1024, can you help take a review?
/cc @ikaven1024

pkg/search/proxy/store/multi_cluster_cache.go
niuyueyang1996: becase this is first accept header.

pkg/search/proxy/store/multi_cluster_cache.go
niuyueyang1996: watch event.object if is CacheableObject ,use pb will message does not implement the protobuf marshalling interface and cannot be encoded to a protobuf message.
so if use watch, should modify the object.

niuyueyang1996:
> @niuyueyang1996, does #4437 solve it?

i think #4437 only fix list() func protobuf bug.

niuyueyang1996:
> @niuyueyang1996, does #4437 solve it?

I'm not sure whether it can be solved watch pb protobuf.
because if use watch_cache.go watch() func,all event object will setCachingObjects.

Issue #4490, opened 2024-01-02
[umbrella] Enable golangci-lint(revive) to improve code quality


jwcesign:
**What would you like to be added**:

Enable the appropriate revive check rules to improve PR quality.

**Why is this needed**:

For golangci-lint (revive), in the absence of any specific configuration, default checking rules will be applied as reference:
https://github.com/golangci/golangci-lint/blob/0264eaa5c7851a35883e2f462e66d648ab8598df/pkg/golinters/revive.go#L251

In PR https://github.com/karmada-io/karmada/pull/3567, a single rule is configured as shown in the image:

 ![image](https://github.com/karmada-io/karmada/assets/19745536/1e0edccd-2207-4c2a-a2af-f9817ca7fdba). 

This configuration results in only this rule being used and disables checking for other rules.

Thus, it's advisable to activate the default rules and rectify the associated warnings(You can pick one or multiple):

- [x] VarDeclarations @jwcesign, https://github.com/karmada-io/karmada/pull/4492
- [x] PackageComments @zhzhuang-zju, https://github.com/karmada-io/karmada/pull/4503
- [x] DotImports @zhzhuang-zju, https://github.com/karmada-io/karmada/pull/4504
- [x] BlankImports @jwcesign, https://github.com/karmada-io/karmada/pull/4495
- [x] Exported @jwcesign, https://github.com/karmada-io/karmada/pull/4491
- [x] VarNaming @jwcesign, https://github.com/karmada-io/karmada/pull/4496
- [x] IndentErrorFlow @jwcesign, https://github.com/karmada-io/karmada/pull/4495
- [x] Range @Larry-shuo, https://github.com/karmada-io/karmada/pull/4511
- [x] Errorf @zhzhuang-zju, https://github.com/karmada-io/karmada/pull/4504
- [x] ErrorNaming @Larry-shuo, https://github.com/karmada-io/karmada/pull/4511
- [x] ErrorStrings @chaosi-zju, https://github.com/karmada-io/karmada/pull/4498
- [x] ReceiverNaming @chaosi-zju, https://github.com/karmada-io/karmada/pull/4499
- [x] IncrementDecrement @chaosi-zju, https://github.com/karmada-io/karmada/pull/4501
- [x] ErrorReturn @chaosi-zju, https://github.com/karmada-io/karmada/pull/4502
- [x] UnexportedReturn @XiShanYongYe-Chang https://github.com/karmada-io/karmada/pull/4552
[...]

PR #4491, opened 2024-01-02
cleanup: enable revive(exported rule) checking

jwcesign:
**What type of PR is this?**
/kind cleanup

cleanup: enable revive(exported rule) checking

Part of https://github.com/karmada-io/karmada/issues/4490

none

```release-note

```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -78,6 +78,7 @@ linters-settings:
       - name: error-naming
       - name: dot-imports
       - name: errorf
+      - name: exported
   staticcheck:
     checks:
       - all
--- a/pkg/util/proxy/proxy.go
+++ b/pkg/util/proxy/proxy.go
@@ -42,11 +42,12 @@ import (
 	clusterapis "github.com/karmada-io/karmada/pkg/apis/cluster"
 )
 
+// SecretGetterFunc is a function to get secret.
 type SecretGetterFunc func(context.Context, string, string) (*corev1.Secret, error)
 
 // ConnectCluster returns a handler for proxy cluster.
 func ConnectCluster(ctx context.Context, cluster *clusterapis.Cluster, proxyPath string, secretGetter SecretGetterFunc, responder registryrest.Responder) (http.Handler, error) {
-	tlsConfig, err := GetTlsConfigForCluster(ctx, cluster, secretGetter)
+	tlsConfig, err := GetTLSConfigForCluster(ctx, cluster, secretGetter)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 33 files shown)


.golangci.yml
```diff
@@ -62,10 +62,9 @@ linters-settings:
       - Prefix(github.com/karmada-io/karmada)
   revive:
     rules:
-      # Disable if-return as it is too strict and not always useful.
-      # https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#if-return
-      - name: if-return
-        disabled: true
```
RainbowMango, line 68: This `if-return` rule is on or off now?

jwcesign: In the newest version, it's disabled by default: https://github.com/golangci/golangci-lint/pull/4087/files#diff-a133b3199860b0c64f3827116b9a35c9154cd171062ec70c2c94591788f015a0

Vacant2333 (approved):

.golangci.yml
```diff
@@ -62,13 +62,12 @@ linters-settings:
...
       - name: superfluous-else
         arguments:
           - preserveScope
+      - name: exported
+        arguments:
+          - disableStutteringCheck
```
jwcesign, line 70: I am not sure whether we need this, if this is enabled, it will have following error:
```
pkg/karmadactl/top/top_pods.go:50:6: exported: type name will be used as top.TopPodOptions by other packages, and that stutters; consider calling this PodOptions (revive)
type TopPodOptions struct {
     ^
```

I prefer to disable it.

jwcesign: Less than 10.

pkg/util/proxy/proxy.go
```diff
@@ -122,6 +123,7 @@ func NewThrottledUpgradeAwareProxyHandler(location *url.URL, transport http.Roun
 	return proxy.NewUpgradeAwareHandler(location, transport, wrapTransport, upgradeRequired, proxy.NewErrorResponder(responder))
 }
 
+// GetTlsConfigForCluster returns a tls config for the cluster.
```
zhzhuang-zju, line 126:
```suggestion
// GetTLSConfigForCluster returns a tls config for the cluster.
func GetTLSConfigForCluster(ctx context.Context, cluster *clusterapis.Cluster, secretGetter SecretGetterFunc) (*tls.Config, error) {
```
Refer to 4496, iit may be better to change it this way to avoid repeated modifications in the same place.

jwcesign: fixed

PR #4501, opened 2024-01-03
cleanup: enable revive(increment-decrement) checking

chaosi-zju:
**What type of PR is this?**

/kind cleanup

cleanup: enable revive(increment-decrement) checking

Fixes part of #4490

```release-note

```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -73,6 +73,7 @@ linters-settings:
       - name: error-strings
       - name: error-return
       - name: receiver-naming
+      - name: increment-decrement
   staticcheck:
     checks:
       - all


jwcesign (commented):
/lgtm

.golangci.yml
```diff
@@ -73,6 +73,7 @@ linters-settings:
       - name: error-strings
       - name: error-return
       - name: receiver-naming
+      - name: increment-decrement
```
RainbowMango, line 76: @jwcesign have you analyzed how many rules we don't need to fix?
Can we enable them with a single PR?

jwcesign: Around 8.

> Can we enable them with a single PR?

Sure.

RainbowMango: Go for it. It makes no sense for contributors to just add a rule without the fix.
/lgtm
/approve

PR #4510, opened 2024-01-04
Add FilterResourceSelectors API to PP and CPP

chaosi-zju:
**What type of PR is this?**

/kind api-change

Add FilterResourceSelectors API to PP and CPP

Fixes #

```release-note
NONE（Part of feature）
```


pkg/apis/policy/v1alpha1/propagation_types.go
```diff
@@ -180,6 +185,32 @@ type ResourceSelector struct {
 	LabelSelector *metav1.LabelSelector `json:"labelSelector,omitempty"`
 }
 
+// FilterResourceSelector the resources will be filtered.
+type FilterResourceSelector struct {
```
chaunceyjiang, line 189: Should there also be a mandatory requirement for resources in FilterResourceSelector to also be in ResourceSelectors?

RainbowMango: Why do you think that? I was thinking about it too. But don't have a clue yet.

chaosi-zju: > Should there also be a mandatory requirement for resources in FilterResourceSelector to also be in ResourceSelectors?

Sorry, I didn't get it, can you give a simple example?

RainbowMango:
The draft idea of the API:
```go
type PropagationSpec struct {
	// ExcludedResources used to exclude resources.
	// It consists of a slice of exclusion rules and a resource will be excluded
	// if it matches any of the rules.
	//
	// Note that, if a resource has been propagated and then excluded, its behavior
	// is equivalent to the deletion of PropagationPolicy or ClusterPropagationPolicy.
	// That is the propagated resource will continue to exist in member cluster(s)
	// until it is removed or took over by another policy.
	ExcludedResources []ExcludedResource `json:"excludedResources,omitempty"`
}

// ExcludedResource specifies which resources need to be excluded.
type ExcludedResource struct {
	// APIVersion represents the API version of the target resources.
	// If not empty, resources under this API group will be excluded.
	// +optional
	APIVersion string `json:"apiVersion,omitempty"`

	// Kind represents the Kind of the target resources.
	// If not empty, resources with this Kind will be excluded.
	// +optional
	Kind string `json:"kind,omitempty"`

	// Namespace of the target resource.
	// If not empty, resources under this namespace will be excluded.
	// +optional

	// Name of the target resource.
	// If not empty, resources with this name will be excluded.
	// +optional
	Name string `json:"name,omitempty"`

	// A label query over a set of resources.
	// If not empty, resources match the label query will be excluded.
	// +optional
	LabelSelector *metav1.LabelSelector `json:"labelSelector,omitempty"`
}
```

chaosi-zju:
> The draft idea of the API:

done

pkg/apis/policy/v1alpha1/propagation_types.go
chaosi-zju: @chaunceyjiang Can you explain your opinion a little more? 
Besides, I've updated the API, could you please see if there are any problems with the latest version?

pkg/apis/policy/v1alpha1/propagation_types.go
```diff
@@ -65,6 +65,16 @@ type PropagationSpec struct {
...
 	ResourceSelectors []ResourceSelector `json:"resourceSelectors"`
 
+	// ExcludedResources used to exclude resources.
+	// It consists of a slice of exclusion rules and a resource will be excluded
+	// if it matches any of the rules.
+	//
+	// Note that, if a resource has been propagated and then excluded, its behavior
+	// is equivalent to the deletion of PropagationPolicy or ClusterPropagationPolicy.
+	// That is the propagated resource will continue to exist in member cluster(s)
+	// until it is removed or took over by another policy.
```
RainbowMango, line 75:
```suggestion
	// until it is removed or took over by another policy.
	// +optional
```
Add `+optional`.

pkg/apis/policy/v1alpha1/propagation_types.go
chaunceyjiang: 例如：

```
resourceSelectors:

- apiVersion: aaa/v1
   kind: cluster
- apiVersion: bbb/v1
   kind: cluster             # Newly added kind

excludedResources:
- kind: cluster
```

Will this newly added 'kind' resource propagate to the member clusters?

RainbowMango: I think the answer is `no`, as the newly added resource already been `excluded`.

chaosi-zju: @chaunceyjiang 
> Will this newly added 'kind' resource propagate to the member clusters?

No, because `aaa/v1/cluster` and `bbb/v1/cluster` are all hit by `excludedResources` condition `kind: cluster`, so neither `aaa/v1/cluster` nor `bbb/v1/cluster` be propagate to the member clusters.

chaunceyjiang:
```
resourceSelectors:

- apiVersion: aaa/v1
   kind: cluster
- apiVersion: aaa/v1
   kind: cluster
- apiVersion: ccc/v1
   kind: cluster             # Newly added kind

excludedResources:
- kind: cluster
  name: bar
```

> so neither aaa/v1/cluster nor bbb/v1/cluster be propagate to the member clusters.


Yes, this is exactly what I was worried about. I added a new GVK, but it surprisingly couldn't be propagated to the member clusters. Apparently, when filling out excludedResources, there was no intention to filter this resource.


So, I believe that the benefit of setting `apiVersion` and `kind` optional to reduce user costs is really minimal.

chaosi-zju: > So, I believe that the benefit of setting apiVersion and kind optional to reduce user costs is really minimal.

Thank you for your suggest, but I think this has nothing to do with it setting `apiVersion` and `kind` optional, this is because you did not follow the rules when writing the policy.

Even if we set `apiVersion` and `kind` required, consider the following example:

```yaml
resourceSelectors:
- apiVersion: aaa/v1
   kind: cluster
   namespace: a
- apiVersion: aaa/v1
   kind: cluster
   namespace: b       # newly added namespace
excludedResources:
- apiVersion: aaa/v1
   kind: cluster
```

now `apiversion` and `kind` is required, then you add a `nameapce: b` resourceSelectors, but it couldn't be propagated to the member clusters, this is essentially the same principle as the example you provided above.

I think it essentially not surprising, but you did not follow the rules when updating the policy.

chaosi-zju: In your case, if you truly want to exclude `cluster` resources under the specified `apiversion`, you should write it fully in the `excludedResources` item. Correspondingly, if you don't write it fully, we can assume that you are trying to exclude `cluster` resources under any `apiversion`, and that the modifier himself should pay attention to the `excludedResources` field when updating `resourceSelectors`.

chaosi-zju:
Is there any further comments?

chaunceyjiang (approved):
/lgtm

chaosi-zju:
@RainbowMango Can this PR to be merged? I will take the next step forward.

PR #4513, opened 2024-01-04
Add linting for redefined builtin ids

Affan-7:
**What type of PR is this?**

/kind cleanup

This PR enables linting for the redefined built-in ids.

Constant names like `false`, `true`, `nil`, function names like `append`, `make`, and basic type names like `bool`, and `byte` are not reserved words of the language; therefore the can be redefined. Even if possible, redefining these built in names can lead to bugs very difficult to detect.

In this PR the `new` keyword was redefined.

Part of #4490

```release-note
NONE
```


pkg/util/binding.go
```diff
@@ -72,28 +72,28 @@ func ConvertToClusterNames(clusters []workv1alpha2.TargetCluster) sets.Set[strin
 }
 
 // MergeTargetClusters will merge the replicas in two TargetCluster
-func MergeTargetClusters(old, new []workv1alpha2.TargetCluster) []workv1alpha2.TargetCluster {
+func MergeTargetClusters(old_cluster, new_cluster []workv1alpha2.TargetCluster) []workv1alpha2.TargetCluster {
```
zhzhuang-zju, line 75: Hi~ @Affan-7 Is it better to use hump case for parameters? Like `oldCluster`.

Affan-7: Thanks for the suggestion, I have edited the code.

RainbowMango:
cc @jwcesign

.golangci.yml
```diff
@@ -83,6 +83,7 @@ linters-settings:
       - name: blank-imports
       - name: indent-error-flow
       - name: unreachable-code
+      - name: redefines-builtin-id
```
jwcesign, line 86: Hi, @Affan-7 , To avoid multiple PR conflicts, you can add this and fix it locally. The modifications here will be submitted together in the end.

RainbowMango: I just rebased this PR.

RainbowMango (approved):
/lgtm
/approve

PR #4520, opened 2024-01-06
chore: verify whether docker is available by executing docker ps command

zll600:
**What type of PR is this?**
/kind feature

The `./hack/local-up-karmada.sh` did not execute successfully but no error was reported when I installed docker but docker was not enabled.
```shell
~/Code/github/karmada #v1.7.2 ❯ ./hack/local-up-karmada.sh                                                                                                                                                                                  
Preparing: 'kind' existence check - passed
Preparing: 'kubectl' existence check - passed
Preparing kindClusterConfig in path: /tmp/tmp.GcYYjKiQ1p
~/Code/github/karmada #v1.7.2 ❯ docker ps
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
```

So I want to make sure docker is available by running `docker ...` command instead of `command -v docker`.

Fixes #

```release-note
None
```

### DIFF
--- a/hack/local-up-karmada.sh
+++ b/hack/local-up-karmada.sh
@@ -75,8 +75,8 @@ fi
 util::cmd_must_exist "go"
 util::verify_go_version
 
-# Make sure docker exists
-util::cmd_must_exist "docker"
+# Make sure docker is available
+util::verify_docker
 
 # install kind and kubectl
 kind_version=v0.20.0
--- a/hack/util.sh
+++ b/hack/util.sh
@@ -112,6 +112,13 @@ function util::cmd_must_exist {
     fi
 }
 
+function util::verify_docker {
+  if ! docker ps -q >/dev/null 2>&1; then
+      echo "Docker is not available, Please verify docker is installed and available"
+      exit 1
+  fi
+}
+
 function util::verify_go_version {
     local go_version
     IFS=" " read -ra go_version <<< "$(GOFLAGS='' go version)"


hack/local-up-karmada.sh
```diff
@@ -75,8 +75,8 @@ fi
 util::cmd_must_exist "go"
 util::verify_go_version
 
-# Make sure docker exists
-util::cmd_must_exist "docker"
+# Make sure docker is available
+util::verify_docker
```
zhzhuang-zju, line 79: Nice finding! Just a small question, does the `usage` function https://github.com/karmada-io/karmada/blob/65376b28d5037c27ff7ec0e56542c2a345d1a120/hack/local-up-karmada.sh#L26 need to add a description of prerequisites for docker is available?

zll600: During my trip, I set up the environment according to the prerequisites in the README. refer:  https://github.com/karmada-io/karmada/blob/65376b28d5037c27ff7ec0e56542c2a345d1a120/README.md?plain=1#L100
So would it be better to put it in the README? I think the `usage` function should describe functions of scripts or usage options of scripts.

And so far there is no description of prerequisites in usage. If we add docker-related descriptions to usage, do we also need to add related descriptions to other prerequisites?

zhzhuang-zju: > So would it be better to put it in the README? I think the `usage` function should describe functions of scripts or usage options of scripts.

That makes sense to me

zhzhuang-zju: > So would it be better to put it in the README? I think the `usage` function should describe functions of scripts or usage options of scripts.

Are you interested in submitting a PR for this?

RainbowMango (approved):
/approve
Thanks.

hack/local-up-karmada.sh
zll600: > > So would it be better to put it in the README? I think the `usage` function should describe functions of scripts or usage options of scripts.
> 
> Are you interested in submitting a PR for this?

Sorry for the late reply.
I think it's a bit strange to add a "make sure docker is available" in prerequisites.
It feels incompatible with existing items. https://github.com/karmada-io/karmada/blob/master/README.md?plain=1#L100,L103
What do you think?

zhzhuang-zju: >  What do you think?

How about 
```
- [docker](https://www.docker.com/) version vx.x.x+ 
note: execute command `docker ps` to make sure docker is available
```

zll600: > > What do you think?
> 
> How about
> 
> ```
> - [docker](https://www.docker.com/) version vx.x.x+ 
> note: execute command `docker ps` to make sure docker is available
> ```

This is OK.
But since I'm new to the project, I'm currently not sure about the docker version. Can we determine the minimum docker version? Alternatively, is there a way to help me to determine the lowest docker version?

RainbowMango: I don't think it's a good idea to list a long list of prerequisites in README.
It's common sense that have docker installed.

zll600: ok, Let's do less first.
If it's really confusing here, we can add more details.
OK?

RainbowMango: Sure. Thanks @zll600. 
Now everything looks fine. Let's wait and see if others need more details.

PR #4524, opened 2024-01-09
Fix misjudgment when updating attached binding

whitewindmills:
**What type of PR is this?**
/kind bug

Fix misjudgment when updating attached binding.
image

We should return `nil` after successfully updating attached binding.

Fixes #

```release-note
NONE
```


chaunceyjiang (commented):
Thanks~

/lgtm

RainbowMango (approved):
/approve
Thanks.

Seems it is kind of cleanup. Please correct me if I'm wrong.
/remove-kind bug
/kind cleanup

PR #4532, opened 2024-01-10
reduce useless logic to improve dependencies distributor

whitewindmills:
**What type of PR is this?**
/kind feature

The dependent distributor does not pay attention to deleted or deleting attached resource templates, so optimize this part of the logic.

Fixes #

```release-note
NONE
```


XiShanYongYe-Chang:
I don't quite understand why dependencies distributor doesn't need to be concerned about deletion events. When the attached resource template is deleted, the corresponding rb should also need to be cleaned up.

whitewindmills:
> I don't quite understand why dependencies distributor doesn't need to be concerned about deletion events. When the attached resource template is deleted, the corresponding rb should also need to be cleaned up.

Since the attached resource template is deleted, so it's resourceBinding is deleted, cleaning up makes no sense.

XiShanYongYe-Chang:
> Since the attached resource template is deleted, so it's resourceBinding is deleted, cleaning up makes no sense.

Are you referring to the gc controller responsible for cleaning up the resource bindings of the resource template?

whitewindmills:
> Are you referring to the gc controller responsible for cleaning up the resource bindings of the resource template?

sure

XiShanYongYe-Chang:
Thanks, I get it. Can you help add some comments on the event handler to explain why we don't need to care about this?

whitewindmills:
> Can you help add some comments on the event handler to explain why we don't need to care about this?

okay

pkg/dependenciesdistributor/dependencies_distributor.go
```diff
@@ -457,19 +469,27 @@ func (d *DependenciesDistributor) isOrphanAttachedBindings(
...
+				if err != nil {
+					// return nil err if resource template not exist.
+					if apierrors.IsNotFound(err) {
+						return false, nil
+					}
+					return false, err
+				}
+				// return nil err if resource template is being deleted.
+				if !rawObject.GetDeletionTimestamp().IsZero() {
+					return false, nil
```
XiShanYongYe-Chang, line 489: I understand that placing this part of the code outside the for loop (before the loop starts) would be easier to understand.

XiShanYongYe-Chang (commented):
Other parts LGTM

XiShanYongYe-Chang:
Kind ping @chaunceyjiang @RainbowMango 
/cc @chaunceyjiang @RainbowMango

pkg/dependenciesdistributor/dependencies_distributor.go
```diff
@@ -572,7 +592,9 @@ func (d *DependenciesDistributor) Start(ctx context.Context) error {
 		},
 		ReconcileFunc: d.reconcile,
 	}
-	d.eventHandler = fedinformer.NewHandlerOnEvents(d.OnAdd, d.OnUpdate, d.OnDelete)
+	// For attached resource templates, its ResourceBinding will be deleted when it's deleted, any update on its ResourceBinding makes no sense.
+	// So we don't pay attention to deleted attached resource templates.
```
RainbowMango, line 596: I am confused here. 
1. We don't usually say `attached resource template`
2. The `DependenciesDistributor` actually watches `ResourceBinding`, so why here say don't pay attention to `attached resource templates`?

RainbowMango: By the way, I am always confused when I read the code of the dependencies feature.

I often can't tell `dependency vs dependent`, and attached ResourceBinding.
Is anyone like me?

XiShanYongYe-Chang: The `attached resource template` resource means resources like `ConfigMap`, `Secret`, etc.

We should unify these terms in the code and then record them somewhere in the code, such as the readme file in the current directory, to facilitate understanding when reading the code.

whitewindmills: > The attached resource template resource means resources like ConfigMap, Secret, etc.

Yes, deleting a ConfigMap of Deployment has no effect on the `DependenciesDistributor`. Now, The `DependenciesDistributor` actually watches `ResourceBinding` and their dependencies. When any of them generates a K8s event, reconciliation will be triggered.

> We should unify these terms in the code and then record them somewhere in the code, such as the readme file in the current directory, to facilitate understanding when reading the code.

+1

whitewindmills:
Could you help move it forward?
/assign @RainbowMango

RainbowMango:
@chaunceyjiang Do you want to take another look? As you are the related code author.

whitewindmills:
@chaunceyjiang Could you help take a look?

XiShanYongYe-Chang:
Hi @whitewindmills, We can move forward with this PR. Can you rebase it and confirm if there are any required changes?

whitewindmills:
ping @RainbowMango @XiShanYongYe-Chang @chaunceyjiang

pkg/dependenciesdistributor/dependencies_distributor.go
```diff
@@ -132,20 +133,26 @@ func (d *DependenciesDistributor) OnAdd(obj interface{}) {
...
-	unstructuredOldObj, err := helper.ToUnstructured(oldObj)
+	unstructuredNewObj, err := helper.ToUnstructured(newObj)
 	if err != nil {
-		klog.Errorf("Failed to transform oldObj, error: %v", err)
+		klog.Errorf("Failed to transform newObj, error: %v", err)
+		return
+	}
+	if !unstructuredNewObj.GetDeletionTimestamp().IsZero() {
+		klog.V(4).Infof("Ignore update event of object(%s, kind=%s, %s) as being deleted.",
+			unstructuredNewObj.GetAPIVersion(), unstructuredNewObj.GetKind(), names.NamespacedKey(unstructuredNewObj.GetNamespace(), unstructuredNewObj.GetName()))
```
RainbowMango, line 143: It's better to check it in the reconciling loop in case there are some clean-up tasks.

RainbowMango (commented):
/assign @XiShanYongYe-Chang @chaunceyjiang

pkg/dependenciesdistributor/dependencies_distributor.go
whitewindmills: but it's for resource template, we won't have any cleanup.

XiShanYongYe-Chang: +1

Our current model cannot discard this event, otherwise it would not be possible to do the cleanup.

Other parts LGTM.

XiShanYongYe-Chang: For this part:

https://github.com/karmada-io/karmada/blob/d676996b240f43ca0430d492b8c54eb0b4986c2d/pkg/dependenciesdistributor/dependencies_distributor.go#L256-L263

whitewindmills: I don't think there is any negative effect since a independent binding's delete event will handle it.

XiShanYongYe-Chang: We have added a finalizer. The delete event does not handle this.

whitewindmills: > I don't have a strong opinion just feel not good to do such things in event filter.
I'm not sure that's a good pattern, my thinking is the event filter should be simple.

I don't totally agree. if unnecessary events can be filtered out before they are queued, this can significantly improve the efficiency of the work queue. Of course, keeping it simple is also a consideration, but compared to this type of processing in the detector, this obviously does not add much complexity.
https://github.com/karmada-io/karmada/blob/6fbdd9e7dae2547e9255e65890abcb56b2e8a34b/pkg/detector/detector.go#L907-L953

whitewindmills:
ping @chaunceyjiang

pkg/dependenciesdistributor/dependencies_distributor.go
```diff
@@ -132,20 +133,26 @@ func (d *DependenciesDistributor) OnAdd(obj interface{}) {
...
 // OnUpdate handles object update event and push the object to queue.
 func (d *DependenciesDistributor) OnUpdate(oldObj, newObj interface{}) {
-	unstructuredOldObj, err := helper.ToUnstructured(oldObj)
+	unstructuredNewObj, err := helper.ToUnstructured(newObj)
 	if err != nil {
-		klog.Errorf("Failed to transform oldObj, error: %v", err)
+		klog.Errorf("Failed to transform newObj, error: %v", err)
+		return
+	}
+	if !unstructuredNewObj.GetDeletionTimestamp().IsZero() {
```
chaunceyjiang, line 141: I remember, if a resource doesn't have `Finalizers`, then when it is deleted, there won't be any update events.

whitewindmills: what this PR wants to do is to ignore the resources that need to be deleted.

whitewindmills:
Hi @XiShanYongYe-Chang, I've rebased this PR onto the latest master and added a comment explaining why we don't need to check DeletionTimestamp in the update handler. The comment clarifies:

1. For independent bindings, deletion events trigger cleanup via handleIndependentBindingDeletion.
2. For attached resource templates, when deleted, their ResourceBindings are also cleaned up by GC.

Please take another look when you have time. Thanks!

XiShanYongYe-Chang:
Thanks, @whitewindmills 
the lint ci is failed, can you help fix it?

Issue #4369, opened 2023-12-04
Could we consider ResourceQuota in scheduling ?


wengyao04:
**What would you like to be added**:
In kubernetes, ResourceQuota is a default admission plugin. In Karmada federation system, could we include ResourceQuota evaluation during scheduling ? Use cases are explained in `Why is this needed`

- could we support it as a filter with a feature gate ?
  - If we support it as a filter then we might need to include the ResourceQuota in the scheduler cache. It would be too heavy if manage thousands of nodes
- or support it in the estimator with a feature gate ?
  - it might be easier to support it in the estimator  

**Why is this needed**:
In our multi-tenant k8s clusters, we have namespace/ResourceQuota for each tenant duplicated on all clusters. 
For example
- cluster 1: namespace s-ns-1, resource quota hard: limits.cpu 40, limits.memory: 512Gi, requests.cpu 40, requests.memory: 512Gi
- cluster 2: namespace s-ns-1, resource quota hard: limits.cpu 40, limits.memory: 512Gi, requests.cpu 40, requests.memory: 512Gi

We want to use Karmada to manage cluster 1 and cluster 2. When users submit jobs from karmada, it should take ResourceQuota into account during the scheduling process.  In a scenario where clients uses the entire quota from "s-ns-1" in Cluster 1 but still have sufficient quota to execute jobs in Cluster 2, karmada should filter out cluster 1 and pick cluster 2 as the target cluster.

wengyao04:
If we support it as a filter then we might need to include the ResourceQuota in the scheduler cache

wengyao04:
> hi @wengyao04, besides `ResourceQuota`, I have another question to ask you. In your scenario, if a kubeflow job `TFJob` is Master (replicas=2) + Worker (replicas=4), which are your expected schedule results:
> 
> * all 6 replicas must be scheduled to same member cluster
> * 1 master replicas and 2 worker replicas can schedule to `member1` cluster, while another 1 master replicas and 2 worker replicas be scheduled to `member2` cluster
> * 2 master replicas can schedule to `member1` cluster, while 4 worker replicas be scheduled to `member2` cluster.
> 
> what I asked is mainly to make sure that whether a group of pods in a `TFJob` can be split to different clusters, thanks

Hi @chaosi-zju In our cluster setup we don't allow pods communication cross multiple clusters, all underlying pods of kubeflow.TFJob should be assigned to the same cluster

PR #4534, opened 2024-01-11
[MVP] add resourcequota plugin in scheduler-estimator: create framework for scheulder-estimator

wengyao04:
**What type of PR is this?**
feature

***why we need it***:
This PR is to resolve https://github.com/karmada-io/karmada/issues/4369

In our multi-tenant k8s clusters, we have namespace/ResourceQuota for each tenant duplicated on all clusters. 
For example
- cluster 1: namespace s-ns-1, resource quota hard: limits.cpu 40, limits.memory: 512Gi, requests.cpu 40, requests.memory: 512Gi
- cluster 2: namespace s-ns-1, resource quota hard: limits.cpu 40, limits.memory: 512Gi, requests.cpu 40, requests.memory: 512Gi

We want to use Karmada to manage cluster 1 and cluster 2. When users submit jobs from karmada, it should take ResourceQuota into account during the scheduling process.  In a scenario where clients uses the entire quota from "s-ns-1" in Cluster 1 but still have sufficient quota to execute jobs in Cluster 2, karmada should filter out cluster 1 and pick cluster 2 as the target cluster.

*** what this PR does***:
In this PR we
- Introduce the plugin framework for scheduler-estimator
  - The plugin framework follows kubernetes scheduling framework design pattern
  - The interface is defined in `pkg/estimator/server/framework/interface.go`
  - The frameworkImpl is in `pkg/estimator/server/framework/runtime/framework.go`
  - The individual plugin is controlled by k8s featureGates

I will send another PR to add resourceQuotaEstimator plugin once the framework PR is merged. The resourceQuotaEstimator plugin will include
- Add resourceQuotaEstimator plugin
  - It is an alpha feature, default to false in the featureGates
  - The resourceQuotaEstimator only supports
    - compute resources (cpu/memory), extended resources (like gpu) and storage resource (ephemeral-storage) 
    - priorityClass scope selector
  - add two optional fields `Namespace` and `PriorityClassName` in `pkg/estimator/pb/types.go`
    ```
        // ReplicaRequirements represents the requirements required by each replica.
    type ReplicaRequirements struct {
	    // NodeClaim represents the NodeAffinity, NodeSelector and Tolerations required by each replica.
	    // +optional
	    NodeClaim *NodeClaim `json:"nodeClaim,omitempty" protobuf:"bytes,1,opt,name=nodeClaim"`
	    // ResourceRequest represents the resources required by each replica.
	    // +optional
	    ResourceRequest corev1.ResourceList `json:"resourceRequest,omitempty" protobuf:"bytes,2,rep,name=resourceRequest,casttype=k8s.io/api/core/v1.ResourceList,castkey=k8s.io/api/core/v1.ResourceName"`
	    // +optional
	    Namespace string `json:"namespace,omitempty" protobuf:"bytes,3,opt,name=namespace"`
	    // +optional
	    PriorityClassName string `json:"priorityClassName,omitempty" protobuf:"bytes,4,opt,name=priorityClassName"`
    }
    ```
  - add  two optional fields `Namespace` and `PriorityClassName` in `pkg/apis/work/v1alpha2/binding_types.go`
    ```
        // ReplicaRequirements represents the requirements required by each replica.
    type ReplicaRequirements struct {
	    // NodeClaim represents the node claim HardNodeAffinity, NodeSelector and Tolerations required by each replica.
	    // +optional
	    NodeClaim *NodeClaim `json:"nodeClaim,omitempty"`
    
	    // ResourceRequest represents the resources required by each replica.
	    // +optional
	    ResourceRequest corev1.ResourceList `json:"resourceRequest,omitempty"`
    
	    // Namespace represents the resources namespaces
	    // +optional
    
	    // PriorityClassName represents the resources priorityClassName
	    // +optional
    }
    ```

Fixes #https://github.com/karmada-io/karmada/issues/4369

@Garrybest, @RainbowMango and @chaosi-zju, could you help review the plugin framework structure ?

I will add unit-tests this week and then turn it into a PR.

```release-note

```


pkg/estimator/pb/types.go
```diff
@@ -57,6 +57,10 @@ type ReplicaRequirements struct {
 	// ResourceRequest represents the resources required by each replica.
 	// +optional
 	ResourceRequest corev1.ResourceList `json:"resourceRequest,omitempty" protobuf:"bytes,2,rep,name=resourceRequest,casttype=k8s.io/api/core/v1.ResourceList,castkey=k8s.io/api/core/v1.ResourceName"`
+	// +optional
+	Namespace string `json:"namespace,omitempty" protobuf:"bytes,3,opt,name=namespace"`
+	// +optional
+	PriorityClassName string `json:"priorityClassName,omitempty" protobuf:"bytes,4,opt,name=priorityClassName"`
```
RainbowMango, line 63: Could you please add a comment on both `Namespace` and `PriorityClassName`? It would be great to explain a little bit how they will be used when a quota measures usage with them.

pkg/apis/work/v1alpha2/binding_types.go
```diff
@@ -175,6 +175,14 @@ type ReplicaRequirements struct {
 	// ResourceRequest represents the resources required by each replica.
 	// +optional
 	ResourceRequest corev1.ResourceList `json:"resourceRequest,omitempty"`
+
+	// Namespace represents the resources namespaces
+	// +optional
+	Namespace string `json:"namespace,omitempty"`
```
RainbowMango, line 181: The `namespace` is already in the `.spec.resource` of ResourceBinding:
https://github.com/karmada-io/karmada/blob/181af49291b25f3f710c645afec9683456dead3b/pkg/apis/work/v1alpha2/binding_types.go#L73

It's worth discussing whether a duplicate field is necessary.

pkg/estimator/server/framework/plugins/feature/feature_gate.go
RainbowMango, line 1: We do not need to build feature gates for a specific component. 
A feature gate is usually shared by all components, as a feature might consist of multiple components' efforts.

See https://github.com/karmada-io/karmada/blob/master/pkg/features/features.go for reference.

pkg/estimator/server/framework/runtime/framework.go
```diff
@@ -0,0 +1,127 @@
+/*
+Copyright 2021 The Karmada Authors.
```
RainbowMango, line 2:
```suggestion
Copyright 2024 The Karmada Authors.
```

Garrybest (commented):
BTW, the title of this PR has a misspelling.

pkg/estimator/server/estimate.go
```diff
@@ -72,9 +72,12 @@ func (es *AccurateSchedulerEstimatorServer) estimateReplicas(
 	if requirements.NodeClaim != nil {
 		tolerations = requirements.NodeClaim.Tolerations
 	}
-
 	// TODO(Garrybest): design a framework and make filter and score plugins
 	var res int32
+	replicas, err := es.estimateFramework.RunEstimateReplicasPlugins(ctx, &requirements)
```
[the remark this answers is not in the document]
wengyao04: Thank you ! I will split this PR to two
- PR with framework only
- once we agree on the framework, I will send another PR for resoucequota plugin

pkg/estimator/pb/types.go
```diff
@@ -57,6 +57,10 @@ type ReplicaRequirements struct {
 	// ResourceRequest represents the resources required by each replica.
 	// +optional
 	ResourceRequest corev1.ResourceList `json:"resourceRequest,omitempty" protobuf:"bytes,2,rep,name=resourceRequest,casttype=k8s.io/api/core/v1.ResourceList,castkey=k8s.io/api/core/v1.ResourceName"`
+	// +optional
+	Namespace string `json:"namespace,omitempty" protobuf:"bytes,3,opt,name=namespace"`
```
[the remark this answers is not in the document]
wengyao04: I make it optional because we want to backward compatibility ?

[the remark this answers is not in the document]
RainbowMango: Does the backward compatibility mean a lower version(like v1.8) of `karmada-scheduler-estimator` which can not handle namespace`, but it can work with a higher version (like the coming v1.9) of `karmada-scheduler`?
When the `karmada-scheduler-estimator` receives a request, it will drop the fields that it doesn't recognized. am I right?

pkg/estimator/server/estimate.go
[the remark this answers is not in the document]
RainbowMango: Good idea!
I see @Garrybest leave a `TODO` that hopes to introduce a framework.

pkg/estimator/pb/types.go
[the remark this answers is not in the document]
Garrybest: > I make it optional because we want to backward compatibility ?

Required field could be backward compatible as well, all the non-omitempty fields that are not filled will be empty strings. BTW, we have already transfered name and namespace between scheduler and estimator, you can pay attention to the input string `object`.

``` go
func (es *AccurateSchedulerEstimatorServer) EstimateReplicas(ctx context.Context, object string, request *pb.MaxAvailableReplicasRequest) (int32, error) 
```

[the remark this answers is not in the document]
RainbowMango: > all the non-omitempty fields that are not filled will be empty strings. 

I'm not familiar with this. I'm not sure if I understand correctly.  Both client and server must handle the `required` field, right? So, client and server must upgrade together, isn't?

Seems the `required` field has been deprecated from protoc3.

Echo from https://protobuf.dev/programming-guides/proto2/#field-labels:
> required: Do not use. Required fields are so problematic they were removed from proto3. Semantics for required field should be implemented at the application layer. When it is used, a well-formed message must have exactly one of this field.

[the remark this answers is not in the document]
Garrybest: > Both client and server must handle the required field, right? So, client and server must upgrade together, isn't?

Well, this `required` that we are talking about is not about required fields in pb. I mean something like this:

```go
        // +required
	Namespace string `json:"namespace" protobuf:"bytes,3,opt,name=namespace"`
```
So it only affects the json format. We use go-to-protobuf to generate proto files, all fields in that file is `optional` or `repeated`, so it does not matter.

cmd/scheduler-estimator/app/scheduler-estimator.go
```diff
@@ -128,7 +128,12 @@ func run(ctx context.Context, opts *options.Options) error {
 	dynamicClient := dynamic.NewForConfigOrDie(restConfig)
 	discoveryClient := discovery.NewDiscoveryClientForConfigOrDie(restConfig)
 
-	e := server.NewEstimatorServer(kubeClient, dynamicClient, discoveryClient, opts, ctx.Done())
+	e, err := server.NewEstimatorServer(kubeClient, dynamicClient, discoveryClient, opts, ctx.Done())
+	if err != nil {
+		klog.Errorf("fail to create estimator server: %v", err)
```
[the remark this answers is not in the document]
chaosi-zju: hi, I am curious abount where did this coding guideline come from?

pkg/estimator/pb/types.go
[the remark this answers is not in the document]
chaosi-zju: I'm trying to understand what you mean:

Presiously, the generated `.proto` file is

```
message ReplicaRequirements {
  // +optional
  optional string namespace = 3;
}
```

after introduced your `// +required` and removed `omitempty `, the generated `.proto` file will become like this

```
message ReplicaRequirements {
  // +required
  optional string namespace = 3;
}
```

This comment is just a way of annotating the fields, they do not actually affect the runtime behaviour of gRPC. 

At runtime of gRPC, they are all optional fields and are backward compatible.

Or put it more directly, the two are simply different annotations for fields. It won't actually work, it's just for more readable (If you need to ensure that a field is required or optional, you ultimately rely on validating and processing it in the other custom program logic).

Am I right?

wengyao04:
Hi, @Garrybest, @RainbowMango and @chaosi-zju, I remove the resourceQuota plugin. This PR only introduces the framework for scheduler-estimator.

pkg/estimator/server/framework/runtime/framework_test.go
```diff
@@ -0,0 +1,150 @@
...
+type TestPlugin struct {
+	name string
+	inj  injectedResult
+}
+
+func (pl *TestPlugin) Name() string {
+	return pl.name
+}
+
+func (pl *TestPlugin) Estimate(ctx context.Context, replicaRequirements *pb.ReplicaRequirements) (int32, error) {
```
chaosi-zju, line 52: there is a lint error which advices changing this line to:

```suggestion
func (pl *TestPlugin) Estimate(_ context.Context, _ *pb.ReplicaRequirements) (int32, error) {
```

pkg/estimator/server/framework/interface.go
```diff
@@ -0,0 +1,51 @@
...
+
+	"k8s.io/client-go/informers"
+
+	"github.com/karmada-io/karmada/pkg/estimator/pb"
+)
+
+// Framework manages the set of plugins in use by the estimator.
+type Framework interface {
+	Handle
+	RunEstimateReplicasPlugins(ctx context.Context, replicaRequirements *pb.ReplicaRequirements) (int32, error)
```
chaosi-zju, line 30: this inferface returns a `int32` value, I suggest add a simple comment to describe what does the numeric return value mean， though I known it represent the calculated replicas.

just for more readable, like another interface in `Karmada`:

https://github.com/chaosi-zju/karmada/blob/d017467fed08d24b1c5622604edccd23cd68e1aa/pkg/scheduler/framework/interface.go#L175-L182

pkg/estimator/server/framework/interface.go
```diff
@@ -0,0 +1,51 @@
...
+// Plugin is the parent type for all the scheduling framework plugins.
+type Plugin interface {
+	Name() string
+}
+
+// EstimateReplicasPlugin is an interface for replica estimation plugins.
+// These filters are used to estimate the replicas for a given pb.ReplicaRequirements
+type EstimateReplicasPlugin interface {
+	Plugin
+	Estimate(ctx context.Context, replicaRequirements *pb.ReplicaRequirements) (int32, error)
```
chaosi-zju, line 42: ditto

cmd/scheduler-estimator/app/scheduler-estimator.go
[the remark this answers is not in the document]
Garrybest: https://google.github.io/styleguide/go/decisions.html#error-strings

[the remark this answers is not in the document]
chaosi-zju: Thanks, got it.

By the way, may be @wengyao04 you can also fix the log in the following lines to keep the log style consistently.

```go
if err = e.Start(ctx); err != nil {
        klog.Errorf("estimator server exits unexpectedly: %v", err)
	return err
}
```

charts/karmada/values.yaml
```diff
@@ -828,6 +828,8 @@ schedulerEstimator:
       maxSurge: 50%
   ## @param apiServer.podDisruptionBudget
   podDisruptionBudget: *podDisruptionBudget
+  ## @param featureGate to schedulerEstimator
+  featureGates: {}
```
chaosi-zju, line 832: If I want to use this filed, what content should I write into it?

Can you give a example as comments above `featureGates: {}`

RainbowMango (commented):
Generally looks good to me. Thanks~

chaosi-zju:
Thank you for your important commit ! Besides, you can mark above multiple comments as `resolved`~

/LGTM

pkg/estimator/server/framework/interface.go
```diff
@@ -0,0 +1,59 @@
...
+
+// Framework manages the set of plugins in use by the estimator.
+type Framework interface {
+	Handle
+	// RunEstimateReplicasPlugins runs the set of configured EstimateReplicasPlugins
+	// for estimating replicas based on the given replicaRequirements.
+	// It returns an integer and an error.
+	// The integer represents the minimum calculated value of estimated replicas from each EstimateReplicasPlugin.
+	RunEstimateReplicasPlugins(ctx context.Context, replicaRequirements *pb.ReplicaRequirements) (int32, error)
+	// TODO(wengyao04): we can add filter and score plugin extension points if needed in the future
```
RainbowMango, line 35: +1, That's exactly what I'm gonna say.

RainbowMango:
/lgtm

/assign @Garrybest 
help to take another look.

@chaosi-zju Can you help to confirm if we have E2E tests for `karmada-scheduler-estimator`?

pkg/estimator/pb/types.go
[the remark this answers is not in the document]
Garrybest: Exactly.

pkg/estimator/server/framework/runtime/framework.go
```diff
@@ -0,0 +1,130 @@
...
+// RunEstimateReplicasPlugins runs the set of configured EstimateReplicasPlugins
+// for estimating replicas based on the given replicaRequirements.
+// It returns an integer and an error.
+// The integer represents the minimum calculated value of estimated replicas from each EstimateReplicasPlugin.
+func (frw *frameworkImpl) RunEstimateReplicasPlugins(ctx context.Context, replicaRequirements *pb.ReplicaRequirements) (int32, error) {
+	startTime := time.Now()
+	defer func() {
+		metrics.FrameworkExtensionPointDuration.WithLabelValues(estimator).Observe(utilmetrics.DurationInSeconds(startTime))
+	}()
+	var result int32 = math.MaxInt32
```
Garrybest, line 108: If no plugins exist, we should return zero?

pkg/estimator/server/estimate.go
```diff
@@ -73,8 +73,11 @@ func (es *AccurateSchedulerEstimatorServer) estimateReplicas(
 		tolerations = requirements.NodeClaim.Tolerations
 	}
 
-	// TODO(Garrybest): design a framework and make filter and score plugins
 	var res int32
+	replicas, err := es.estimateFramework.RunEstimateReplicasPlugins(ctx, &requirements)
+	if err != nil {
+		return replicas, err
+	}
 	processNode := func(i int) {
```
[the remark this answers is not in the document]
Garrybest: Or set the snapshot as the input, maybe it's better?

pkg/estimator/server/framework/interface.go
Garrybest: Snapshot could a parameter here.

pkg/estimator/server/server.go
```diff
@@ -120,9 +124,18 @@ func NewEstimatorServer(
 		es.informerManager.Lister(gvr)
 	}
 
+	registry := frameworkplugins.NewInTreeRegistry()
+	estimateFramework, err := frameworkruntime.NewFramework(registry,
+		frameworkruntime.WithInformerFactory(informerFactory),
```
Garrybest, line 129: Add kubeClient here, cause you may need to retreive some objects like resource quota or something else. If possile, make 'get kubeClient' as a interface func could be reasonable.

Garrybest (changes requested):

pkg/estimator/server/framework/runtime/framework.go
wengyao04: If no plugins, the replica is calculated using node resources. If set to zero when not plugins, then the the final estimation will be zero even if there are enough resources to run the job (we compare the results and return the smaller one ?).

If no plugins, return math.MaxInt32 means no operation.

pkg/estimator/server/server.go
wengyao04: In the inteface.go https://github.com/karmada-io/karmada/pull/4534/files#diff-517e6d53713eb677525f7bc765326759bd4d46cf6fa1940e4fabfc814b3ae561

I have Handle interface
```
// Handle provides data and some tools that plugins can use. It is
// passed to the plugin factories at the time of plugin initialization. Plugins
// must store and use this handle to call framework functions.
// We follow the design pattern as kubernetes scheduler framework
type Handle interface {
	SharedInformerFactory() informers.SharedInformerFactory
}
```

If we need KubeClient in the future we can add Kubeclientand provide a withKubeClientOption func like how we pass sharedInformerFactory

wengyao04: In our ResourceQuota Plugin, I need the `SharedInformerFactory` and will use the Lister to get ResourceQuota from local indexer. The estimator will cache all ResourceQuota.

pkg/estimator/server/estimate.go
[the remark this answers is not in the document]
Garrybest: This snapshot is different from kube-scheduler. `kube-scheduler` only schedules one pod simultaneously, so snapshot could be a part of handler. However, estimator could handle several requests simultaneously, they do have their own snapshots. As a result, take snapshot as a member field of handler may not be resonable.

pkg/estimator/server/framework/runtime/framework.go
Garrybest: > If we set to zero, then the the final estimation will be zero even if there are enough resources to run the job (we compare the results and return the smaller one ?

That's what I have mentioned above. We'd better transform estimation by node idle resources to a new plugin. Along with resource plugin and quota plugin, we decide the final estimation result. Does it make sense?

pkg/estimator/server/estimate.go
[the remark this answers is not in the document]
wengyao04: Got it. Then we can add `cache *schedcache.Cache` in the Handler if needed ?

pkg/estimator/server/server.go
wengyao04: I add The KubeClient

pkg/estimator/server/framework/runtime/framework.go
wengyao04: I prefer to think estimation by node idle resources as the "core" and plugin is for add-on features and more light-weight.

If we really want to convert that to NodeResourceEstimator as default feature, I prefer to add `*scheduler.Cache` in the Handle and keep the `RunEstimateReplicasPlugins` as
```
RunEstimateReplicasPlugins(ctx context.Context, replicaRequirements *pb.ReplicaRequirements) (int32, error)
```

wengyao04: I think more about it. It might be useful to change estimation by node idle resource as a plugin because I hope to introduce eviction by priority class. 

My native think about that is to write a plugin to estimate the replica based on node idle resources and resources taken by pods' resources with lower priority. I will turn off the default estimation by node idle resource. 
The pod eviction is handled by the assigned clusters' k8s scheduler.

wengyao04: @Garrybest, @RainbowMango and @chaosi-zju if we agree to change the current estimation by idle resource to the default plugin. Could we keep RunEstimateReplicasPlugins as it is now

`RunEstimateReplicasPlugins(ctx context.Context, replicaRequirements *pb.ReplicaRequirements) (int32, error)`

and add `*scheduler.Cache` in the Handle as I explained in ?
https://github.com/karmada-io/karmada/pull/4534#discussion_r1456891393

Garrybest: Snapshot from a cache is a little bit expensive. Different plugins could not share a snapshot if make cache as a part of handler cause every plugin should do the snapshot work if needed, right?

pkg/estimator/server/estimate.go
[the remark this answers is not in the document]
RainbowMango: This approach is feasible I think.

But we should know that if each plugin builds the snapshot itself, there is a possibility that the snapshot built by two plugins may be different because the cache is subject to change all the time. I don't see any issues here though.

pkg/estimator/server/estimate.go
```diff
@@ -84,6 +87,10 @@ func (es *AccurateSchedulerEstimatorServer) estimateReplicas(
 		atomic.AddInt32(&res, maxReplica)
 	}
 	es.parallelizer.Until(ctx, len(allNodes), processNode)
+
+	if replicas < res {
```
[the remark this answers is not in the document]
RainbowMango: Yes, that's the idea. Did I miss anything?

pkg/estimator/server/server.go
RainbowMango: It's not an objection, but we don't need the `KubeClient` for this PR, we can add it anytime when we need it. It's ok to keep it if you think we will use it. :)

pkg/estimator/server/estimate.go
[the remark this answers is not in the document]
Garrybest: I think we can assume all plugins share one snapshot at the same time. Cluster changes all the time. Event if plugin has their own snapshot based on that time, there still could be some changes during caculation. Normally, estimation will finish in 100ms so it may not bring too many differences.

pkg/estimator/server/framework/runtime/framework.go
wengyao04: It is so true, we need to put snapshot as an input arg
How about 
```
Estimate(ctx context.Context, snapshot *schedcache.Snapshot, replicaRequirements *pb.ReplicaRequirements) (int32, error)
```
make it similar
 [estimateReplicas](https://github.com/karmada-io/karmada/blob/master/pkg/estimator/server/estimate.go#L58-L61 )?

@Garrybest and @RainbowMango and @chaosi-zju do you think if I should convert the current [estimateReplicas](https://github.com/karmada-io/karmada/blob/master/pkg/estimator/server/estimate.go#L58-L61 ) to a default plugin or make that change in another PR ?

RainbowMango: Need @Garrybest 's point here. I'm not sure if it's the original idea or just makes filter and score plugins.
https://github.com/karmada-io/karmada/blob/697170b90463176a519b19f9e4eab22acd07e8a0/pkg/estimator/server/estimate.go#L76

Both are OK to me for now. 
Keeping the current estimation as a `core` makes sense because we can go back to consider this after the resource quota plugin gets onboard.
Moving the current estimation to the plugin also makes sense because it `seems` can support more use cases (just as mentioned https://github.com/karmada-io/karmada/pull/4534#discussion_r1456914573 above). But, we need more effort for it ,such as make plugin configurable.

But, please don't do it in this PR. (Keep each small and we can move faster)

RainbowMango:
@wengyao04 the lint check is failing, you can follow [the instructions](https://github.com/karmada-io/karmada/actions/runs/7565622541/job/20603590529?pr=4534#step:5:828) to fix it:
> hack/../vendor, 'go.mod' or 'go.sum' is out of date. Please run hack/update-vendor.sh

pkg/estimator/server/framework/runtime/framework.go
Garrybest: Exactly, I think we just add snapshot as an input and optimize resource plugin in next PR. This PR generally looks good.

pkg/estimator/server/server.go
Garrybest: Sure.

wengyao04:
Hi @Garrybest, @RainbowMango and @chaosi-zju I submit another commit to address Garry's comment
- put snapshot as an input as `RunEstimateReplicasPlugins`, assuming all plugins share the same cluster view.
  - I understand Garry's concern because in the estimator, we start from an empty snapshot and call the  `UpdateSnapshot`. We don't really utilize the optimization of using snapshotGeneration check (as k8s scheduler snapshot). It would be expensive if we load empty snapshot and update snapshot for plugins that need snapshot. But please correct me if I misunderstand the cache in lifted package.
  - My initial concern is that clusters resources might be change for each plugin. But I think more :D. Loading snapshot for every plugin does not really make the estimator more accurate, because the final results is from the smallest estimated replica. Let's think about the case, we need two replica for our job. Plugin-1 returns 1 replica and Plugin-2 returns 4 ( cluster free some resources during estimation), but our estimator sever will return 1 and the job is not able to be assigned to its managed cluster.
  - There are always race conditions, but karmada scheduler will put that resourcebinding back to the queue with a rate-limiter. So I agree to let all plugins share the same snapshot to make every estimator fast (You also mentions that it usually with 100ms, we should be good)
- introduce Result, similar to karmada scheduler plugin [Result](https://github.com/karmada-io/karmada/blob/697170b90463176a519b19f9e4eab22acd07e8a0/pkg/scheduler/framework/interface.go#L63-L70)
  - I add the Nooperation code which means empty plugin list or all plugins are disabled
  - I think we can give user clearer message if estimation is not succeeded
  - It also addresses Garry's comment about return value if plugins are empty, https://github.com/karmada-io/karmada/pull/4534#discussion_r1455798078. We check the return code instead of error.
- add KubeClient Handle in case that we need it in the future
- rerun `make update` to fix lint check   

Could we go through the review iteration again as Result is the new code. Thx !!

pkg/estimator/server/framework/runtime/framework.go
RainbowMango: Got it.

RainbowMango: > Snapshot from a cache is a little bit expensive. 

Good point! I'm sold.

RainbowMango (commented):
/lgtm

All comments resolved.
Apprecated your effort @wengyao04 @Garrybest @chaosi-zju 

Leave approval to owner @Garrybest for another look.

RainbowMango:
By the way,
It is recommended to keep each commit present a unit of work, that's why we sometimes ask contributors to [squash commits](https://karmada.io/docs/contributor/github-workflow#squash-commits). This PR contains `address comments` and `fix lint error` which could be squashed.

No big deal, as I labeled this PR with `tide/merge-method-squash`, the robot will squash it when merging. 
I'm saying that because some reviewers might ask you to do so.

RainbowMango:
/lgtm
new commits fixed a typo found in unit test.

Garrybest (approved):
Thanks! It's a nice job! @wengyao04 

/lgtm
/approve

Issue #4516, opened 2024-01-05
labels not cleaned when updating ResourceTemplate which resulting in it no longer match a Policy


chaosi-zju:
**What happened**:

When updating ResourceTemplate and which result in it no longer match a Policy, the corresponding behavior should be consistent with the behavior of deleting the policy, like deleting the label `propagationpolicy.karmada.io/name: xxx` from ResourceTemplate. But the actual corresponding label has not been deleted.

**What you expected to happen**:

the label `propagationpolicy.karmada.io/name: xxx` should be deleted.

**How to reproduce it (as minimally and precisely as possible)**:

1. write a deployment (pay attention to it has two labels):

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx
  labels:
    app: nginx
    region: cn
spec:
  replicas: 2
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
        - image: nginx
          name: nginx
          resources:
            limits:
              cpu: 25m
              memory: 64Mi
```

2. write two policy:

```yaml
apiVersion: policy.karmada.io/v1alpha1
kind: PropagationPolicy
metadata:
  name: nginx-pp
spec:
  priority: 2
  placement:
    clusterAffinity:
      clusterNames:
        - member1
  resourceSelectors:
    - apiVersion: apps/v1
      kind: Deployment
      labelSelector:
        matchLabels:
          app: nginx
---
apiVersion: policy.karmada.io/v1alpha1
kind: PropagationPolicy
metadata:
  name: nginx-pp2
spec:
  priority: 1
  placement:
    clusterAffinity:
      clusterNames:
        - member2
  resourceSelectors:
    - apiVersion: apps/v1
      kind: Deployment
      labelSelector:
        matchLabels:
          region: cn
```

3. remove the label `app: nginx` in `Deployment/nginx`, it will no longer match `nginx-pp`:
    * we should remove the label `propagationpolicy.karmada.io/name: nginx-pp` from `Deployment/nginx`, but it doesn't.
    * the `Deployment/nginx` should be propagated by `nginx-pp2`, but it doesn't.

**Anything else we need to know?**:

**Environment**:
- Karmada version:
- kubectl-karmada or karmadactl version (the result of `kubectl-karmada version` or `karmadactl version`):
- Others:

PR #4537, opened 2024-01-11
filter deprecated resources which once altered groups in detector discovering

chaosi-zju:
**What type of PR is this?**

/kind cleanup
/kind deprecation

filter deprecated resources which once altered groups in detector discovering.

Some resources, during their evolutionary iterations, undergo modifications that involve not only changing the version but also altering the group. e.g: Ingress(extensions/v1beta1) is deprecated in v1.14+, unavailable in v1.22+,
while Ingress(networking.k8s.io/v1beta1) is the preferred substitute in v1.14 ~ v1.18, and Ingress(networking.k8s.io/v1) is the preferred substitute in v1.19+.

As for v1.19 ~ v1.21 kubernetes cluster, two group Ingress are actually the same resources, when you create Ingress(networking.k8s.io/v1), you can get both Ingress(networking.k8s.io/v1) and Ingress(extensions/v1beta1) from apiserver, vice versa.
Thereby, when user creates an Ingress(networking.k8s.io/v1) and specifies a PropagationPolicy to propagate it to the member clusters, the detector will listen two resource creation events: Ingress(networking.k8s.io/v1) and Ingress(extensions/v1beta1).

However, the PropagationPolicy only support the ServerPreferredResources since detector only listen resource filtered by `discoveryClient.ServerPreferredResources()`. Just like HPA once had v1/v2beta2/v2 three different versions in k8s v1.25, originally events of three versions would all be generated,
but two of them are filtered not to discover in detector, only keeping ServerPreferredResources version discovered.

Those resources altered groups are just special cases of ServerPreferredResources, essentially should follow the same principle. That means the deprecated resource like Ingress(extensions/v1beta1) should be filtered not to discover, and the preferred resource like Ingress(networking.k8s.io/v1) should be kept discovering.

Fixes part of #4516

```release-note
Ingress(extensions/v1beta1) will no longer be supported in PropagationPolicy when Ingress(networking.k8s.io/v1) exist.
```

### DIFF
--- a/pkg/detector/detector.go
+++ b/pkg/detector/detector.go
@@ -175,9 +175,45 @@ var (
 	_ manager.LeaderElectionRunnable = &ResourceDetector{}
 )
 
+// Some resources have changed group for historical reasons, thus available in more than one group in the same kubernetes version.
+// e.g: Ingress(extensions/v1beta1) and Ingress(networking.k8s.io/v1)
+// Resource in such different groups are actually the same resource, however, one group are preferred while others are deprecated.
+// sameResourceInDifferentGroup is a map enumerating such resources which once changed group.
+// the key refers to the preferred GroupVersionResource in a given k8s version,
+// while the value refers to corresponding deprecated GroupVersionResource existing in another group.
+var sameResourceInDifferentGroup = map[schema.GroupVersionResource]schema.GroupVersionResource{
+	{Group: "networking.k8s.io", Version: "v1", Resource: "ingresses"}:      {Group: "extensions", Version: "v1beta1", Resource: "ingresses"},
+	{Group: "networking.k8s.io", Version: "v1beta1", Resource: "ingresses"}: {Group: "extensions", Version: "v1beta1", Resource: "ingresses"},
+}
+
 func (d *ResourceDetector) discoverResources(period time.Duration) {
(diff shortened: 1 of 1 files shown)


chaosi-zju:
CC @XiShanYongYe-Chang help for a review

XiShanYongYe-Chang:
Thanks~
/assign

chaosi-zju:
CC @RainbowMango Looking forward to your review

XiShanYongYe-Chang (commented):
/lgtm

I'm not sure if it's appropriate to describe this PR as a "feature". Additionally, it may be necessary to state in the release notes that the extension Ingress is no longer supported.

chaosi-zju:
> I'm not sure if it's appropriate to describe this PR as a "feature". Additionally, it may be necessary to state in the release notes that the extension Ingress is no longer supported.

got it, done

chaosi-zju:
As described in [Document](https://karmada.io/docs/next/troubleshooting/trouble-shooting#schedule-failed-due-to-clusters-did-not-have-the-api-resource), Karmada detector focus only on resources with karmada-apiserver preferred version.

This PR is about same resource in different group, can this PR continue to be pushed ? 

ping @XiShanYongYe-Chang @RainbowMango

PR #4554, opened 2024-01-16
search add check member cluster api enable

yanfeng1992:
**What type of PR is this?**
/kind bug

When crd exists in the karmada control plane but does not exist in the member cluster, search will report an error.
Checks for member crd should be added

karmada version: 1.5

image

Fixes #

```release-note
`karmada-search`: Add the logic of checking whether the resource API to be retrieved is installed in the cluster.
```

### DIFF
--- a/pkg/search/controller.go
+++ b/pkg/search/controller.go
@@ -46,6 +46,7 @@ import (
 	"github.com/karmada-io/karmada/pkg/util/fedinformer"
 	"github.com/karmada-io/karmada/pkg/util/fedinformer/genericmanager"
 	"github.com/karmada-io/karmada/pkg/util/gclient"
+	"github.com/karmada-io/karmada/pkg/util/helper"
 	"github.com/karmada-io/karmada/pkg/util/restmapper"
 )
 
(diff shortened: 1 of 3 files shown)


yanfeng1992:
Please take a look, thanks @XiShanYongYe-Chang

XiShanYongYe-Chang (commented):
/lgtm
Thanks~

Ask @ikaven1024 to help take a review.
/cc @ikaven1024

XiShanYongYe-Chang:
How about this: `karmada-search: Add the logic of checking whether the resource API to be retrieved is installed in the cluster.`?

yanfeng1992:
> How about this: `karmada-search: Add the logic of checking whether the resource API to be retrieved is installed in the cluster.`?
> 
> Describe in more detail.

ok

XiShanYongYe-Chang:
Thanks~
/approve

XiShanYongYe-Chang:
We need to synchronize it with the previous three versions: release 1.6, 1.7 and 1.8. I see that the version you found is 1.5, are you still using version 1.5?

yanfeng1992:
> are you still using version 1.5?

we are still at version  1.5

XiShanYongYe-Chang:
> we are still at version 1.5

Sorry for replying late. Our current strategy is to only maintain the latest three versions, version 1.5 may not be within the scope of synchronization.

PR #4558, opened 2024-01-17
refactor: cleanup resource labels when delelte policy in detector

XiShanYongYe-Chang:
**What type of PR is this?**

/kind cleanup

When the Policy is deleted, the labels that need to be cleaned up will change with development iterations. This modification point has a bit of code smell and requires modifications in multiple places, so some refactoring has been done.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/detector/detector.go
+++ b/pkg/detector/detector.go
@@ -1129,6 +1129,7 @@ func (d *ResourceDetector) ReconcileClusterPropagationPolicy(key util.QueueKey) 
 //
 // Note: The relevant ResourceBinding will continue to exist until the resource template is gone.
 func (d *ResourceDetector) HandlePropagationPolicyDeletion(policyNS string, policyName string) error {
+	var errs []error
 	labelSet := labels.Set{
 		policyv1alpha1.PropagationPolicyNamespaceLabel: policyNS,
 		policyv1alpha1.PropagationPolicyNameLabel:      policyName,
@@ -1140,24 +1141,20 @@ func (d *ResourceDetector) HandlePropagationPolicyDeletion(policyNS string, poli
 		return err
 	}
 
-	for index, binding := range rbs.Items {
-		// Must remove the labels from the resource template ahead of ResourceBinding, otherwise might lose the chance
-		// to do that in a retry loop(in particular, the label was successfully removed from ResourceBinding, but
-		// resource template not), since the ResourceBinding will not be listed again.
-		if err := d.CleanupLabels(binding.Spec.Resource, policyv1alpha1.PropagationPolicyNamespaceLabel, policyv1alpha1.PropagationPolicyNameLabel, policyv1alpha1.PropagationPolicyPermanentIDLabel); err != nil {
-			klog.Errorf("Failed to clean up label from resource(%s-%s/%s) when propagation policy(%s/%s) removing, error: %v",
-				binding.Spec.Resource.Kind, binding.Spec.Resource.Namespace, binding.Spec.Resource.Name, policyNS, policyName, err)
-			return err
+	removeLabels := func(labels map[string]string) map[string]string {
+		for _, l := range propagationPolicyRefLabels {
+			delete(labels, l)
 		}
+		return labels
+	}
 
-		// Clean up the labels from the reference binding so that the karmada scheduler won't reschedule the binding.
-		if err := d.CleanupResourceBindingLabels(&rbs.Items[index], policyv1alpha1.PropagationPolicyNamespaceLabel, policyv1alpha1.PropagationPolicyNameLabel, policyv1alpha1.PropagationPolicyPermanentIDLabel); err != nil {
-			klog.Errorf("Failed to clean up label from resource binding(%s/%s) when propagation policy(%s/%s) removing, error: %v",
-				binding.Namespace, binding.Name, policyNS, policyName, err)
-			return err
+	for index := range rbs.Items {
+		err = d.cleanupPolicyRefNamespaceScopeResources(policyNS, policyName, &rbs.Items[index], removeLabels)
+		if err != nil {
+			errs = append(errs, err)
 		}
 	}
-	return nil
+	return errors.NewAggregate(errs)
 }
 
 // HandleClusterPropagationPolicyDeletion handles ClusterPropagationPolicy delete event.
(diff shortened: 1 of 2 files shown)


zhzhuang-zju:
> /cc @zhzhuang-zju

/assign

pkg/detector/detector.go
```diff
@@ -1171,28 +1168,22 @@ func (d *ResourceDetector) HandleClusterPropagationPolicyDeletion(policyName str
 		policyv1alpha1.ClusterPropagationPolicyLabel: policyName,
 	}
 
+	removeLabels := func(labels map[string]string) map[string]string {
```
chaosi-zju, line 1171: minor flaw, I think this variable name `removeLabels` is a little simple and violent, it could be improved

pkg/detector/helper.go
```diff
@@ -0,0 +1,14 @@
+package detector
```
chaosi-zju, line 1: this file doesn't have license header.

pkg/detector/detector.go
XiShanYongYe-Chang: Any suggestion?

zhzhuang-zju (commented):
Thank you for your work, the suggestions I made are just suggestions.

pkg/detector/detector.go
```diff
@@ -1401,15 +1412,10 @@ func (d *ResourceDetector) CleanupResourceBindingLabels(rb *workv1alpha2.Resourc
...
-	for _, l := range labels {
-		delete(bindingLabels, l)
-	}
-
+// cleanupCRBLabels removes labels from cluster resource binding.
+func (d *ResourceDetector) cleanupCRBLabels(crb *workv1alpha2.ClusterResourceBinding, removeLabels RemoveMapKeysFunc) error {
 	return retry.RetryOnConflict(retry.DefaultRetry, func() (err error) {
-		crb.SetLabels(bindingLabels)
+		crb.SetLabels(removeLabels(crb.GetLabels()))
 		updateErr := d.Client.Update(context.TODO(), crb)
```
zhzhuang-zju, line 1419: How about:
```suggestions
	bindingLabels := removeLabels(crb.GetLabels())
	return retry.RetryOnConflict(retry.DefaultRetry, func() (err error) {
		crb.SetLabels(bindingLabels)
```
Avoid repeatedly calling `removeLabels` when retrying to prevent failure.

pkg/detector/detector.go
chaosi-zju: As for me, I would prefer declare a common function like:

```go
func TrimMapByKeys(inputMap map[string]interface{}, keys []string) map[string]interface{}
```

This name clearly conveys the purpose and action of the function. It indicates that the function trims/removes a slice of specific key-value pair from the input map and returns the updated map as the output. 

By including "Trim" and "ByKeys" in the function name, it provides a clear indication of the operation being performed on the map.

Then, we invoke the function as `TrimMapByKeys(workload.GetLabels(), propagationPolicyRefLabels)`

In this case, may be this function will not be required as a parameter and been passed on for many times.

However, just for reference only.

pkg/detector/detector.go
```diff
@@ -1347,8 +1325,46 @@ func (d *ResourceDetector) HandleClusterPropagationPolicyCreationOrUpdate(policy
...
-// CleanupLabels removes labels from object referencing by objRef.
-func (d *ResourceDetector) CleanupLabels(objRef workv1alpha2.ObjectReference, labelKeys ...string) error {
+func (d *ResourceDetector) cleanupPolicyRefNamespaceScopeResources(policyNS, policyName string, rb *workv1alpha2.ResourceBinding, removeLabels RemoveMapKeysFunc) error {
+	// Must remove the labels from the resource template ahead of ResourceBinding, otherwise might lose the chance
+	// to do that in a retry loop(in particular, the label was successfully removed from ResourceBinding, but
+	// resource template not), since the ResourceBinding will not be listed again.
+	if err := d.cleanupResourceTemplateLabels(rb.Spec.Resource, removeLabels); err != nil {
+		klog.Errorf("Failed to clean up label from resource(%s-%s/%s) when propagation policy(%s/%s) removing, error: %v",
+			rb.Spec.Resource.Kind, rb.Spec.Resource.Namespace, rb.Spec.Resource.Name, policyNS, policyName, err)
+		return err
```
zhzhuang-zju, line 1335: The function `cleanupPolicyRefNamespaceScopeResources` is inconsistent with the original implementation. When `cleanupResourceTemplateLabels` fails, it will directly throw an error and return. However, the original implementation will continue to execute `cleanupRBLabels`.

pkg/detector/detector.go
```diff
@@ -1347,8 +1325,46 @@ func (d *ResourceDetector) HandleClusterPropagationPolicyCreationOrUpdate(policy
...
+}
+
+func (d *ResourceDetector) cleanupPolicyRefClusterScopeResources(policyName string, crb *workv1alpha2.ClusterResourceBinding, removeLabels RemoveMapKeysFunc) error {
+	// Must remove the labels from the resource template ahead of ClusterResourceBinding, otherwise might lose the chance
+	// to do that in a retry loop(in particular, the label was successfully removed from ClusterResourceBinding, but
+	// resource template not), since the ClusterResourceBinding will not be listed again.
+	if err := d.cleanupResourceTemplateLabels(crb.Spec.Resource, removeLabels); err != nil {
+		klog.Errorf("Failed to clean up label from resource(%s-%s) when cluster propagation policy(%s) removing, error: %v",
+			crb.Spec.Resource.Kind, crb.Spec.Resource.Name, policyName, err)
+		return err
```
zhzhuang-zju, line 1354: ditto

pkg/detector/detector.go
XiShanYongYe-Chang: The idea here is not to bring the specific label key to the bottom layer for cleaning. For the bottom layer, it only needs to return the processed label. Therefore, an anonymous function is used here for processing.

pkg/detector/helper.go
XiShanYongYe-Chang: Thanks~, done

pkg/detector/detector.go
```diff
@@ -1401,15 +1412,10 @@ func (d *ResourceDetector) CleanupResourceBindingLabels(rb *workv1alpha2.Resourc
...
 
-// CleanupClusterResourceBindingLabels removes labels from cluster resource binding.
-func (d *ResourceDetector) CleanupClusterResourceBindingLabels(crb *workv1alpha2.ClusterResourceBinding, labels ...string) error {
-	bindingLabels := crb.GetLabels()
-	for _, l := range labels {
-		delete(bindingLabels, l)
-	}
-
+// cleanupCRBLabels removes labels from cluster resource binding.
+func (d *ResourceDetector) cleanupCRBLabels(crb *workv1alpha2.ClusterResourceBinding, removeLabels RemoveMapKeysFunc) error {
```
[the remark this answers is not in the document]
XiShanYongYe-Chang: In the context of the entire call chain, the function of this anonymous function is to delete a specified label. Therefore, I think it is easier to understand if we keep the "remove" field. I have modified it as "removeLabelHandler". What do you think?

For RemoveMapKeysFunc, we need to modify the map, so let's rename it as MutateMapFunc.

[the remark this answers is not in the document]
zhzhuang-zju: I'm fine with that.

pkg/detector/helper.go
```diff
@@ -0,0 +1,14 @@
...
+
+import (
+	policyv1alpha1 "github.com/karmada-io/karmada/pkg/apis/policy/v1alpha1"
+)
+
+var (
+	propagationPolicyRefLabels = []string{policyv1alpha1.PropagationPolicyNamespaceLabel, policyv1alpha1.PropagationPolicyNameLabel, policyv1alpha1.PropagationPolicyPermanentIDLabel}
```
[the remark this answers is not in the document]
XiShanYongYe-Chang: Done.

XiShanYongYe-Chang:
After some consideration, I feel that the benefits of this refactoring are not significant, and it also reduces readability. Let's close it for now.

PR #4561, opened 2024-01-19
e2e: Remove annotaions when pp/cpp is deleted

zhzhuang-zju:
**What type of PR is this?**
/kind cleanup

Remove annotaions when pp/cpp is deleted
Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/detector/detector.go
+++ b/pkg/detector/detector.go
@@ -112,7 +112,7 @@ type ResourceDetector struct {
 
 	RESTMapper meta.RESTMapper
 
-	// waitingObjects tracks of objects which haven't be propagated yet as lack of appropriate policies.
+	// waitingObjects tracks of objects which haven't been propagated yet as lack of appropriate policies.
 	waitingObjects map[keys.ClusterWideKey]struct{}
 	// waitingLock is the lock for waitingObjects operation.
 	waitingLock sync.RWMutex
@@ -1242,7 +1242,7 @@ func (d *ResourceDetector) HandleClusterPropagationPolicyDeletion(policyID strin
 			}
 
 			// Clean up the marks from the reference binding so that the Karmada scheduler won't reschedule the binding.
-			if err := d.CleanupClusterResourceBindingLabels(&crbs.Items[index], cleanupMarksFun); err != nil {
+			if err := d.CleanupClusterResourceBindingMarks(&crbs.Items[index], cleanupMarksFun); err != nil {
 				klog.Errorf("Failed to clean up marks from clusterResourceBinding(%s) when clusterPropagationPolicy removed, error: %v",
 					binding.Name, err)
 				errs = append(errs, err)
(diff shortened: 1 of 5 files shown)


zhzhuang-zju:
cc @Vacant2333  @XiShanYongYe-Chang

Vacant2333:
@zhzhuang-zju i will help u to review after the ci pass~~~

pkg/detector/detector.go
```diff
@@ -91,7 +91,7 @@ type ResourceDetector struct {
 
 	RESTMapper meta.RESTMapper
 
-	// waitingObjects tracks of objects which haven't be propagated yet as lack of appropriate policies.
+	// waitingObjects tracks of objects which haven't been propagated yet as lack of appropriate policies.
```
Vacant2333, line 94: nice catch

Vacant2333 (commented):
looks good to me, can u help to take a look @XiShanYongYe-Chang ~

XiShanYongYe-Chang:
Thanks a lot
/assign

pkg/detector/cleanup.go
```diff
@@ -0,0 +1,196 @@
...
+	workload.SetAnnotations(cleanupMapKeysFunc(workload.GetAnnotations(), annotations))
+
+	gvr, err := restmapper.GetGroupVersionResource(d.RESTMapper, workload.GroupVersionKind())
+	if err != nil {
+		klog.Errorf("Failed to delete resource(%s/%s) labels as mapping GVK to GVR failed: %v", workload.GetNamespace(), workload.GetName(), err)
+		return err
+	}
+
+	newWorkload, err := d.DynamicClient.Resource(gvr).Namespace(workload.GetNamespace()).Update(context.TODO(), workload, metav1.UpdateOptions{})
+	if err != nil {
```
chaunceyjiang, line 132: Just a question: Should we consider the situation where the workload cannot be found here?

zhzhuang-zju: Yes, we should do nothing if resource template not exist, since it might have been removed. This step has been implemented in 
```go
// removes labels from object referencing by objRef.
func (d *ResourceDetector) cleanupObjectRefLabelsAndAnnotations(objRef workv1alpha2.ObjectReference, labels, annotations []string) error {
	workload, err := helper.FetchResourceTemplate(d.DynamicClient, d.InformerManager, d.RESTMapper, objRef)
	if err != nil {
		// do nothing if resource template not exist, it might have been removed.
			return nil
		}
		klog.Errorf("Failed to fetch resource(kind=%s, %s/%s): %v", objRef.Kind, objRef.Namespace, objRef.Name, err)
		return err
	}

	workload = workload.DeepCopy()

	return d.cleanupWorkLoadLabelsAndAnnotations(workload, labels, annotations)
}
```

chaunceyjiang: Clearly, just because you can find this resource here doesn't necessarily mean it will still exist when you update it.

zhzhuang-zju: Strictly speaking, there is no guarantee that the resource will still exist at the time of update. If the resource no longer exists, an error will be reported during the update and the retry loop will be entered. Then it will be detected that it no longer exists and the loop will exit. There will be one more retry.

zhzhuang-zju:
since https://github.com/karmada-io/karmada/pull/4743 has been merged, so this pr changes to add some related e2e scenarios
@XiShanYongYe-Chang @Vacant2333

XiShanYongYe-Chang (commented):
Thanks a lot~
/lgtm
/approve

XiShanYongYe-Chang (commented):
/assign @RainbowMango

RainbowMango (approved):
/approve

Obviously, we missed to rename it at #4743.

Issue #4563, opened 2024-01-19
Propagation Policy Support Lazy Reconcile feature


chaosi-zju:
**What would you like to be added**:

Introduced a Lazy Reconcile feature to PropagationPolicy. When it is enabled, the resource will not immediately respond to policy modifications if it has already been distributed according to the previous policy. Instead, it will only respond to the policy changes during the next update of the resource.

**Why is this needed**:

**Backgound**：

In some case, the user uses a global policy to manage multiple resource templates. One day he wants to modify the distribution strategy in that global policy. Since the current policy modifications take effect immediately, the user are concerned that the changes may cause failures in business applications outside the designated change window. 

Therefore, the user wish to delay the implementation of policy modifications until the scheduled change window for the business applications.

**Proposal Detail**

1. Provide Lazy Reconcile feature which is controlled by global configuration (feature gate).

2. When feature enabled, the effective time of policy modification will be delayed until the resource template is modified.

3. Karmada's own components do not trigger the Policy to take effect for resource template modifications.

4. If user needs the ability of Policy to take effect immediately when Lazy Reconcile feature enabled, 
    he can trigger the Policy to take effect via the `kubectl label` command. 
    e.g: `kubectl label deploy,svc,cm -l app=nginx --overwrite refresh-time=$(date +%s)` 

**Notes**：

This feature has some system constraints and risks, which will be described in detail proposal document.

***

**Related Actions**：

1、submit the proposal about this feature @chaosi-zju 

- [ ] proposal

2、Small-step commit related code implementation @chaosi-zju 

- [ ] add a feature-gate for Lazy Reconcile feature

On Going ...

chaosi-zju:
CC @RainbowMango 

#4564 and #4570 are all PRs about code modification, and ready to review, thanks.

XiShanYongYe-Chang:
Can this issue be closed first? The E2E task can be tracked by #4607. @chaosi-zju

PR #4564, opened 2024-01-19
add a feature-gate for Lazy Reconcile feature

chaosi-zju:
**What type of PR is this?**

/kind feature

Small-step commit related code implementation for Policy Lazy Reconcile feature.

This PR only handle: add a feature-gate for Lazy Reconcile feature

More info refer to #4563

add a feature-gate for Lazy Reconcile feature

Fixes part of #4563

```release-note

```


zhzhuang-zju:
Hi~ Do this pr need to add a release note?
others LGTM

pkg/features/features.go
```diff
@@ -54,6 +58,7 @@ var (
 		CustomizedClusterResourceModeling: {Default: true, PreRelease: featuregate.Beta},
 		PolicyPreemption:                  {Default: false, PreRelease: featuregate.Alpha},
 		MultiClusterService:               {Default: false, PreRelease: featuregate.Alpha},
+		PolicyLazyReconcile:               {Default: false, PreRelease: featuregate.Alpha},
```
chaunceyjiang, line 61: Why is there only one global switch? Will you also introduce related fields in PP next?

What I mean is, can the system have both an immediately effective PP and a `Lazy Reconcile` PP at the same time?

chaunceyjiang: Because if there's only one of two options, then assuming all future defaults are `Lazy Reconcile`, most users would refuse to upgrade due to their lack of experience with `Lazy Reconcile`.

RainbowMango: +1
`lazy` should not be the default behavior.

chaosi-zju: Lazy is not the default behavior as `PolicyLazyReconcile{Default: false}`

What chaunceyjiang mean is that it is better to add a filed to Policy to control the the effect strategy.

chaunceyjiang: What I mean is: It would be best if these two behaviors could coexist in the karmada.

chaunceyjiang: I still have one question I don't fully understand, which is: if a resource has not changed but the PP has changed, then this resource will continue to use the old PP. If a rescheduling happens at this time, should the scheduler use the old PP or the new PP? If it uses the new PP, then that behavior would be somewhat inconsistent with our expected behavior. 
Our expectation is that the new PP would only be applied if the resource itself has changed.

chaosi-zju:
/close

this feature-gate is depracated, replaced by Policy field, refer to #4577.

Issue #4191, opened 2023-10-28
Set MinVersion to VersionTLS13 for tlsconfig to fix CVE-2016-2183


zhzhuang-zju:
**What would you like to be added**:
set MinVersion to VersionTLS13 for tlsconfig to fix [CVE-2016-2183](https://github.com/advisories/GHSA-w2rw-pv8p-h9c8). Through nmap scanning, the following components were found to have this vulnerability. 
- karmada-apiserver
- karmada-aggregated-apiserver
- karmada-search
- karmada-metrics-adapter
- etcd

Referring to [Installation overview](https://karmada.io/docs/installation/), there are four installation methods. So, this issue will also be divided into four parts to solve.
- [x] Installation from Source  (@zhzhuang-zju)
   - [x] karmada-apiserver (@zhzhuang-zju, #4198)
   - [x] karmada-aggregated-apiserver (@zhzhuang-zju, #4198)
   - [x] karmada-search (@zhzhuang-zju, #4198)
   - [x] karmada-metrics-adapter (@zhzhuang-zju, #4198)
   - [x] etcd (@zhzhuang-zju, #4253)
- [x] Installation of CLI Tools
   - [x] karmada-apiserver  (@yanfeng1992, #4181)
   - [x] karmada-aggregated-apiserver (@yanfeng1992, #4181)
   - [x] karmada-search (@yanfeng1992, #4193)
   - [x] karmada-metrics-adapter (@yanfeng1992, #4193)
   - [x] etcd (@zhzhuang-zju, #4253)
- [x] Install Karmada by Helm Chart Deployment
   - [x] karmada-apiserver (@zhzhuang-zju, #4198)
   - [x] karmada-aggregated-apiserver (@zhzhuang-zju, #4198)
   - [x] karmada-search (@zhzhuang-zju, #4198)
   - [x] karmada-metrics-adapter (not involving)
   - [x] etcd (@zhzhuang-zju, #4253)
- [x] Install Karmada by Karmada Operator
   - [x] karmada-apiserver (@zhzhuang-zju, #4198)
   - [x] karmada-aggregated-apiserver (@zhzhuang-zju, #4198)
   - [x] karmada-search (not involving)
   - [x] karmada-metrics-adapter (@zhzhuang-zju, #4198)
   - [x] etcd (@zhzhuang-zju, #4253)
- [x] File a [security advisory](https://github.com/karmada-io/karmada/security/advisories) or a document for this CVE
 
**Why is this needed**:
[...]

PR #4565, opened 2024-01-19
specify `appProtocol` for services that use https

a7i:
**What type of PR is this?**

/kind feature

For clusters that use istio, you must specify port name following format $name-$protocol or use appProtocol.

Istio Docs: https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/#explicit-protocol-selection

Fixes #

```release-note
add `appProtocol` to manifests for kind service
```

### DIFF
--- a/artifacts/deploy/karmada-search.yaml
+++ b/artifacts/deploy/karmada-search.yaml
@@ -77,7 +77,8 @@ metadata:
     apiserver: "true"
 spec:
   ports:
-    - port: 443
+    - appProtocol: https
+      port: 443
       protocol: TCP
       targetPort: 443
   selector:
--- a/artifacts/deploy/karmada-webhook.yaml
+++ b/artifacts/deploy/karmada-webhook.yaml
@@ -63,5 +63,7 @@ spec:
   selector:
     app: karmada-webhook
   ports:
-    - port: 443
+    - appProtocol: https
+      port: 443
+      protocol: TCP
       targetPort: 8443
--- a/charts/karmada/templates/karmada-aggregated-apiserver.yaml
+++ b/charts/karmada/templates/karmada-aggregated-apiserver.yaml
@@ -119,7 +119,8 @@ metadata:
     {{- include "karmada.aggregatedApiServer.labels" . | nindent 4 }}
 spec:
   ports:
-    - port: 443
+    - appProtocol: https
+      port: 443
       protocol: TCP
       targetPort: 443
   selector:
--- a/charts/karmada/templates/karmada-search.yaml
+++ b/charts/karmada/templates/karmada-search.yaml
@@ -103,7 +103,8 @@ metadata:
     {{- include "karmada.search.labels" . | nindent 4 }}
 spec:
   ports:
-    - port: 443
+    - appProtocol: https
+      port: 443
       protocol: TCP
       targetPort: 443
   selector:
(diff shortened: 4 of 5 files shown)


artifacts/deploy/karmada-search.yaml
```diff
@@ -77,7 +77,8 @@ metadata:
     apiserver: "true"
 spec:
   ports:
-    - port: 443
+    - appProtocol: https
+      port: 443
```
zhzhuang-zju, line 81: Karmada has different installation methods, and the directories `artifact/deploy` and `charts/karmada/templates` correspond to two different installation methods. Should the configurations of different installation methods be the same? If so, you can refer to https://github.com/karmada-io/karmada/issues/4191 to find the configuration files for different installation methods.

zhzhuang-zju (commented):
Hi~ @a7i  thanks for your work! Review is ongoing, and my viewpoint just for reference only.

zhzhuang-zju:
Hi~ Ask out of curiosity, have you combined Karmada with Istio? If so, could you share your use cases and experiences? Thank you.

artifacts/deploy/karmada-search.yaml
```diff
@@ -77,7 +77,8 @@ metadata:
     apiserver: "true"
 spec:
   ports:
-    - port: 443
+    - appProtocol: https
```
RainbowMango, line 80: Is this also useful when working with other system other than Istio? In other words, if another system requires another appProtocol?
(I'm not an expert in this area, maybe stupid question)

artifacts/deploy/karmada-search.yaml
a7i: `appProtocol` is well supported field regardless of the install method. shouldn't these two be in sync as much as possible?

a7i:
> Hi~ Ask out of curiosity, have you combined Karmada with Istio? If so, could you share your use cases and experiences? Thank you.

@zhzhuang-zju we do not have karmada in the mesh to prevent dependencies on cluster critical resources. With that being said, istio requires `appProtocol` or port `name` to follow this [Protocol Selection](https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/). We have opa-gatekeeper policies in place to ensure that if a Service exposes port 80/443 that it specifies `appProtocol` or follow correct naming convention. Otherwise, this breaks routing for ALL services listening on 80/443, causing a cluster-level outage. See related issue: https://github.com/istio/istio/issues/47088

artifacts/deploy/karmada-search.yaml
a7i: Also not an expert as we only use Istio :)

artifacts/deploy/karmada-search.yaml
zhzhuang-zju: Yes, they should be in sync as much as possible. So you may need to make the same changes in different installation methods.

artifacts/deploy/karmada-search.yaml
zhzhuang-zju: > In other words, if another system requires another appProtocol? (I'm not an expert in this area, maybe stupid question)

I think the configuration files we provide here should be universal and independent of specific usage. Can we use them as a practice to adapt Karmada to Istio and showcase it in tutorials or other documents?

a7i:
We ended up adding a kustomize patch to do this. I think it's a good change to add upstream since `appProtocol` is HTTPS

PR #4566, opened 2024-01-19
[MVP] add resourcequota plugin in scheduler-estimator: add resourcequota plugin

wengyao04:
**What type of PR is this?**
This is the followup PR of https://github.com/karmada-io/karmada/pull/4534

The resourceQuotaEstimator plugin will include
- Add resourceQuotaEstimator plugin
  - It is an alpha feature, default to false in the featureGates
  - The resourceQuotaEstimator only supports
    - compute resources (cpu/memory), extended resources (like gpu) and storage resource (ephemeral-storage) 
    - priorityClass scope selector
  - add two optional fields `Namespace` and `PriorityClassName` in `pkg/estimator/pb/types.go`
    ```
        // ReplicaRequirements represents the requirements required by each replica.
    type ReplicaRequirements struct {
	    // NodeClaim represents the NodeAffinity, NodeSelector and Tolerations required by each replica.
	    // +optional
	    NodeClaim *NodeClaim `json:"nodeClaim,omitempty" protobuf:"bytes,1,opt,name=nodeClaim"`
	    // ResourceRequest represents the resources required by each replica.
	    // +optional
	    ResourceRequest corev1.ResourceList `json:"resourceRequest,omitempty" protobuf:"bytes,2,rep,name=resourceRequest,casttype=k8s.io/api/core/v1.ResourceList,castkey=k8s.io/api/core/v1.ResourceName"`
	    // +optional
	    Namespace string `json:"namespace,omitempty" protobuf:"bytes,3,opt,name=namespace"`
	    // +optional
	    PriorityClassName string `json:"priorityClassName,omitempty" protobuf:"bytes,4,opt,name=priorityClassName"`
    }
    ```
  - add  two optional fields `Namespace` and `PriorityClassName` in `pkg/apis/work/v1alpha2/binding_types.go`
    ```
        // ReplicaRequirements represents the requirements required by each replica.
    type ReplicaRequirements struct {
	    // NodeClaim represents the node claim HardNodeAffinity, NodeSelector and Tolerations required by each replica.
	    // +optional
	    NodeClaim *NodeClaim `json:"nodeClaim,omitempty"`
    
	    // ResourceRequest represents the resources required by each replica.
	    // +optional
	    ResourceRequest corev1.ResourceList `json:"resourceRequest,omitempty"`
    
	    // Namespace represents the resources namespaces
	    // +optional
    
	    // PriorityClassName represents the resources priorityClassName
	    // +optional
    }
    ```

We discuss if we need to add duplicate `namespace` filed `workv1alpha2.ReplicaRequirements`. I add it because we also need that in the resource interpreter https://github.com/karmada-io/karmada/blob/master/pkg/resourceinterpreter/interpreter.go#L46-L47
if we don't include the namespace in `workv1alpha2.ReplicaRequirements`, I have to make this interface change to return namespace. 
```
// GetReplicas returns the desired replicas of the object as well as the requirements of each replica.
GetReplicas(object *unstructured.Unstructured) (replica int32, replicaRequires *workv1alpha2.ReplicaRequirements, err error)
```

Fixes # 
Fixes #https://github.com/karmada-io/karmada/issues/4369

@RainbowMango, @Garrybest, @chaosi-zju could you help check ? Thx

```release-note

```


pkg/util/helper/binding.go
```diff
@@ -397,8 +397,10 @@ func GenerateReplicaRequirements(podTemplate *corev1.PodTemplateSpec) *workv1alp
 
 	if nodeClaim != nil || resourceRequest != nil {
 		return &workv1alpha2.ReplicaRequirements{
-			NodeClaim:       nodeClaim,
-			ResourceRequest: resourceRequest,
+			NodeClaim:         nodeClaim,
+			ResourceRequest:   resourceRequest,
+			Namespace:         podTemplate.Namespace,
+			PriorityClassName: podTemplate.Spec.PriorityClassName,
```
Garrybest, line 403: I remember priority class name can be left empty in pod template and K8s pod admission hook will set the default priority class name for pod. So this name could be empty.

Do we handle this default value in the plugin?

Garrybest (commented):
BTW, I am not so familiar with resource quota, kindly invite @RainbowMango and @XiShanYongYe-Chang to take a look as well.

pkg/apis/work/v1alpha2/binding_types.go
```diff
@@ -175,6 +175,14 @@ type ReplicaRequirements struct {
 	// ResourceRequest represents the resources required by each replica.
 	// +optional
 	ResourceRequest corev1.ResourceList `json:"resourceRequest,omitempty"`
+
+	// Namespace represents the resources namespaces
+	// +optional
+	Namespace string `json:"namespace,omitempty"`
```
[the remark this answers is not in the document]
RainbowMango: That is because the ReplicaEstimator interface shares the same `ReplicaRequirements` struct with ResourceBinding:
https://github.com/karmada-io/karmada/blob/f0543135210e3b692a1d453e92687d9002615c44/pkg/estimator/client/interface.go#L39
I also noticed this issue at https://github.com/karmada-io/karmada/pull/4534#discussion_r1448859091.


Another reason just as mentioned in PR description:
> We discuss if we need to add duplicate namespace filed workv1alpha2.ReplicaRequirements. I add it because we also need that in the resource interpreter https://github.com/karmada-io/karmada/blob/master/pkg/resourceinterpreter/interpreter.go#L46-L47
if we don't include the namespace in workv1alpha2.ReplicaRequirements, I have to make this interface change to return namespace.

pkg/util/helper/binding.go
wengyao04: In this MVP, we probably not, otherwise I need to listwatch PriorityClass in the karmada controller. I can add a comment that if user don't specify in the pod spec, priorityClassName will be empty.

pkg/estimator/server/framework/plugins/registry.go
```diff
@@ -17,11 +17,14 @@ limitations under the License.
...
+	"github.com/karmada-io/karmada/pkg/estimator/server/framework/plugins/resourcequota"
 	"github.com/karmada-io/karmada/pkg/estimator/server/framework/runtime"
 )
 
 // NewInTreeRegistry builds the registry with all the in-tree plugins.
 func NewInTreeRegistry() runtime.Registry {
-	registry := runtime.Registry{}
+	registry := runtime.Registry{
+		resourcequota.Name: resourcequota.New,
+	}
```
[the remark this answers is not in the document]
wengyao04: Instead of control in the registry, do you think if we can have feature-gate check in each plugin.New function like https://github.com/karmada-io/karmada/pull/4566/files#diff-e489afa02e13c3ea92d007b4a8575045b8695f34cd31539769f2e28c3c0cb322R72-R77 ?

pkg/util/helper/binding.go
RainbowMango: > priorityClassName will be empty

Then the plugin will be skipped for those without priorityClassName, right?
That's kind of a limitation, we can do that for this MVP.

pkg/estimator/server/framework/plugins/registry.go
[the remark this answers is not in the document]
RainbowMango: I'm saying that just because I see there are two places we need to check the feature-gate in the plugin. :)
I'm not insisting on it, both will work. No big deal.

pkg/util/helper/binding.go
wengyao04: It will handle by the `matchScope` function https://github.com/karmada-io/karmada/pull/4566/files#diff-e489afa02e13c3ea92d007b4a8575045b8695f34cd31539769f2e28c3c0cb322R240-R261

Garrybest: > Besides that, I think client should explicitly specify the priorityClass because the default global priorityClass might be different among managed clusters

But we can't determine the action of users. This field is filled by users.

You could try to watch PriorityClass in estimator instead. If we have a request of empty PriorityClass field, the estimator tries to get default value of its cluster and run this plugin.

pkg/estimator/server/framework/plugins/resourcequota/resourcequota.go
```diff
@@ -0,0 +1,360 @@
...
+// 1) Terminating
+// 2) NotTerminating
+// 3) BestEffort
+// 4) NotBestEffort
+// 5) PriorityClass
+// 6) CrossNamespacePodAffinity
+// In the MVP, we only support ComputeResource (cpu, memory and extended resource) and PriorityClass scope, the reasons are
+// 1) pb.ReplicaRequirements only contains requested resources currently, we cannot determine if its QoS (BestEffort or not)
+// 2) storage and object count doest not support scope selector
+// ToDo (wengyao04): we can extend the resourceQuotaEstimator support if further requirement are needed
```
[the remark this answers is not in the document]
zhzhuang-zju:
```suggestion
// TODO (@wengyao04): we can extend the resourceQuotaEstimator support if further requirements are needed
```

wengyao04:
I create an umbrella issue https://github.com/karmada-io/karmada/issues/4578

pkg/util/helper/binding.go
wengyao04: @Garrybest I think client should explicitly specify the priorityClass because the default global priorityClass might be different among managed clusters.

RainbowMango: I guess you mean that the administrator might set different default priority classes for each cluster, such as by [PriorityClass](https://github.com/kubernetes/kubernetes/blob/1ad143177c365212e53793aff204c6f1c9727d45/staging/src/k8s.io/api/scheduling/v1/types.go#L37-L47). 

So, the default priority in Karmada might not be accurate.

pkg/estimator/server/framework/plugins/resourcequota/resourcequota.go
```diff
@@ -0,0 +1,361 @@
...
+		if selector.Operator == corev1.ScopeSelectorOpExists {
+			// This is just checking for existence of a priorityClass on the pod,
+			// no need to take the overhead of selector parsing/evaluation.
+			return len(priorityClassName) != 0, nil
+		}
+		return matchesSelector(priorityClassName, selector)
+	case corev1.ResourceQuotaScopeCrossNamespacePodAffinity:
+		return false, nil
+	default:
+		klog.Info("selector is not supported", "scope name", selector.ScopeName)
```
RainbowMango, line 259:
```suggestion
		// hit here means Kubernetes introduced a new resource quota scope
		klog.Warning("Unrecognized scope name of resource quota", "scope name", selector.ScopeName)
```
Here deserves a warning log, right?

pkg/estimator/server/framework/plugins/resourcequota/resourcequota.go
```diff
@@ -0,0 +1,361 @@
...
+			}
+		}
+		resource := util.NewResource(resourceList)
+		resource.AllowedPodNumber = math.MaxInt64
+		allowed := resource.MaxDivided(filteredRequiredResourceList)
+		// continue the loop to avoid integer overflow
+		if allowed == math.MaxInt64 {
+			continue
+		}
+		replica := int32(allowed)
```
RainbowMango, line 173: Is the `allowed` possible to exceed `math.MaxInt32`, the converted value might not be expected?

pkg/apis/work/v1alpha2/binding_types.go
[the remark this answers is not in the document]
RainbowMango: This issue is now tracked by https://github.com/karmada-io/karmada/issues/4578.

pkg/util/helper/binding.go
```diff
@@ -399,6 +399,10 @@ func GenerateReplicaRequirements(podTemplate *corev1.PodTemplateSpec) *workv1alp
 		return &workv1alpha2.ReplicaRequirements{
 			NodeClaim:       nodeClaim,
 			ResourceRequest: resourceRequest,
+			Namespace:       podTemplate.Namespace,
```
RainbowMango, line 402: Shall we add a feature gate check here? The two fields are unnecessary for users that do not enable this feature?

RainbowMango: By the way, we are sure that we have the opportunity to set the `PriorityClassName` in the `if nodeClaim != nil || resourceRequest != nil`,  because the `resourceRequest` can not be nil, right?

RainbowMango (commented):
Generally looks good to me.

pkg/util/helper/binding.go
wengyao04: Hi, I am not quite sure if I understand this question correct, I think we handle the nil replicaRequirement in estimator [here](https://github.com/karmada-io/karmada/blob/master/pkg/estimator/client/accurate.go#L88-L100)

The default `pb.ReplicaRequirements` is empty, it will be populated only if ReplicaRequirements is not nil
```
	req := &pb.MaxAvailableReplicasRequest{
		Cluster:             cluster,
		ReplicaRequirements: pb.ReplicaRequirements{},
	}
	if replicaRequirements != nil {
		req.ReplicaRequirements.ResourceRequest = replicaRequirements.ResourceRequest
		if replicaRequirements.NodeClaim != nil {
			req.ReplicaRequirements.NodeClaim = &pb.NodeClaim{
				NodeAffinity: replicaRequirements.NodeClaim.HardNodeAffinity,
				NodeSelector: replicaRequirements.NodeClaim.NodeSelector,
				Tolerations:  replicaRequirements.NodeClaim.Tolerations,
			}
		}
	}
```
If the pb.ReplicaRequirements is empty, which means empty ResourceList, the current [logic](https://github.com/karmada-io/karmada/blob/master/pkg/estimator/server/estimate.go#L98-L106) still calculate the replica
```
func (es *AccurateSchedulerEstimatorServer) nodeMaxAvailableReplica(node *framework.NodeInfo, rl corev1.ResourceList) int32 {
	rest := node.Allocatable.Clone().SubResource(node.Requested)
	// The number of pods in a node is a kind of resource in node allocatable resources.
	// However, total requested resources of all pods on this node, i.e. `node.Requested`,
	// do not contain pod resources. So after subtraction, we should cope with allowed pod
	// number manually which is the upper bound of this node available replicas.
	rest.AllowedPodNumber = util.MaxInt64(rest.AllowedPodNumber-int64(len(node.Pods)), 0)
	return int32(rest.MaxDivided(rl))
}
```

pkg/estimator/server/framework/plugins/resourcequota/resourcequota.go
wengyao04: fix it according to the comment.

RainbowMango:
Just helped rebase the code.
Looks good to me now.

RainbowMango (approved):
/lgtm
/approve

PR #4577, opened 2024-01-24
Introduced a lazy activation preference to PropagationPolicy/ClusterPropagationPolicy

chaosi-zju:
**What type of PR is this?**

/kind feature

Introduced a lazy activation preference to PropagationPolicy/ClusterPropagationPolicy.

ActivationPreference indicates how the referencing resource template will be propagated, in case of policy changes.

If empty, the resource template will respond to policy changes immediately, in other words, any policy changes will drive the resource template to be propagated immediately as per the current propagation rules.

If the value is 'Lazy' means the policy changes will not take effect for now but defer to the resource template changes, in other words, the resource template will not be propagated as per the current propagation rules until there is an update on it.

This is an experimental feature that might help in a scenario where a policy manages huge amount of resource templates, changes to a policy typically affect numerous applications simultaneously. A minor misconfiguration could lead to widespread failures. With this feature, the change can be gradually rolled out through iterative modifications of resource templates.

More info refer to #4563

Fixes part of #4563

```release-note
Introduced a lazy activation preference to PropagationPolicy/ClusterPropagationPolicy.
```

### DIFF
--- a/pkg/apis/policy/v1alpha1/propagation_types.go
+++ b/pkg/apis/policy/v1alpha1/propagation_types.go
@@ -152,6 +152,27 @@ type PropagationSpec struct {
 	// +kubebuilder:validation:Enum=Abort;Overwrite
 	// +optional
 	ConflictResolution ConflictResolution `json:"conflictResolution,omitempty"`
+
+	// ActivationPreference indicates how the referencing resource template will
+	// be propagated, in case of policy changes.
+	//
+	// If empty, the resource template will respond to policy changes
+	// immediately, in other words, any policy changes will drive the resource
+	// template to be propagated immediately as per the current propagation rules.
+	//
+	// If the value is 'Lazy' means the policy changes will not take effect for now
+	// but defer to the resource template changes, in other words, the resource
+	// template will not be propagated as per the current propagation rules until
+	// there is an update on it.
+	// This is an experimental feature that might help in a scenario where a policy
+	// manages huge amount of resource templates, changes to a policy typically
+	// affect numerous applications simultaneously. A minor misconfiguration
+	// could lead to widespread failures. With this feature, the change can be
+	// gradually rolled out through iterative modifications of resource templates.
+	//
+	// +kubebuilder:validation:Enum=Lazy
+	// +optional
+	ActivationPreference ActivationPreference `json:"activationPreference,omitempty"`
 }
 
 // ResourceSelector the resources will be selected.
@@ -516,6 +537,16 @@ const (
 
 	// ConflictAbort means that do not resolve the conflict and stop propagating.
 	ConflictAbort ConflictResolution = "Abort"
+)
+
+// ActivationPreference indicates how the referencing resource template will be propagated, in case of policy changes.
+type ActivationPreference string
+
+const (
+	// LazyActivation means the policy changes will not take effect for now but defer to the resource template changes,
+	// in other words, the resource template will not be propagated as per the current propagation rules until
+	// there is an update on it.
+	LazyActivation ActivationPreference = "Lazy"
 )
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
--- a/pkg/detector/detector.go
+++ b/pkg/detector/detector.go
@@ -154,7 +154,7 @@ func (d *ResourceDetector) Start(ctx context.Context) error {
 
 	detectorWorkerOptions := util.Options{
 		Name:               "resource detector",
-		KeyFunc:            ClusterWideKeyFunc,
+		KeyFunc:            ResourceItemKeyFunc,
 		ReconcileFunc:      d.Reconcile,
 		RateLimiterOptions: d.RateLimiterOptions,
 	}
@@ -226,11 +226,14 @@ func (d *ResourceDetector) NeedLeaderElection() bool {
 // Reconcile performs a full reconciliation for the object referred to by the key.
 // The key will be re-queued if an error is non-nil.
 func (d *ResourceDetector) Reconcile(key util.QueueKey) error {
-	clusterWideKey, ok := key.(keys.ClusterWideKey)
+	clusterWideKeyWithConfig, ok := key.(keys.ClusterWideKeyWithConfig)
 	if !ok {
 		klog.Error("Invalid key")
 		return fmt.Errorf("invalid key")
 	}
+
+	clusterWideKey := clusterWideKeyWithConfig.ClusterWideKey
+	resourceChangeByKarmada := clusterWideKeyWithConfig.ResourceChangeByKarmada
 	klog.Infof("Reconciling object: %s", clusterWideKey)
 
 	object, err := d.GetUnstructuredObject(clusterWideKey)
@@ -243,7 +246,7 @@ func (d *ResourceDetector) Reconcile(key util.QueueKey) error {
 			// currently we do that by setting owner reference to derived objects.
 			return nil
 		}
-		klog.Errorf("Failed to get unstructured object(%s), error: %v", clusterWideKey, err)
+		klog.Errorf("Failed to get unstructured object(%s), error: %v", clusterWideKeyWithConfig, err)
 		return err
 	}
 
@@ -255,7 +258,7 @@ func (d *ResourceDetector) Reconcile(key util.QueueKey) error {
 		return nil
 	}
 
-	return d.propagateResource(object, clusterWideKey)
+	return d.propagateResource(object, clusterWideKey, resourceChangeByKarmada)
 }
 
 // EventFilter tells if an object should be taken care of.
@@ -320,7 +323,7 @@ func (d *ResourceDetector) OnAdd(obj interface{}) {
 	if !ok {
 		return
 	}
-	d.Processor.Enqueue(runtimeObj)
+	d.Processor.Enqueue(ResourceItem{Obj: runtimeObj})
 }
 
 // OnUpdate handles object update event and push the object to queue.
@@ -337,12 +340,25 @@ func (d *ResourceDetector) OnUpdate(oldObj, newObj interface{}) {
 		return
 	}
 
+	newRuntimeObj, ok := newObj.(runtime.Object)
+	if !ok {
+		klog.Errorf("Failed to assert newObj as runtime.Object")
+		return
+	}
+
 	if !eventfilter.SpecificationChanged(unstructuredOldObj, unstructuredNewObj) {
 		klog.V(4).Infof("Ignore update event of object (kind=%s, %s/%s) as specification no change", unstructuredOldObj.GetKind(), unstructuredOldObj.GetNamespace(), unstructuredOldObj.GetName())
 		return
 	}
 
-	d.OnAdd(newObj)
+	resourceChangeByKarmada := eventfilter.ResourceChangeByKarmada(unstructuredOldObj, unstructuredNewObj)
+
+	resourceItem := ResourceItem{
+		Obj:                     newRuntimeObj,
+		ResourceChangeByKarmada: resourceChangeByKarmada,
+	}
+
+	d.Processor.Enqueue(resourceItem)
 }
 
 // OnDelete handles object delete event and push the object to queue.
@@ -407,7 +423,8 @@ func (d *ResourceDetector) LookForMatchedClusterPolicy(object *unstructured.Unst
 }
 
 // ApplyPolicy starts propagate the object referenced by object key according to PropagationPolicy.
-func (d *ResourceDetector) ApplyPolicy(object *unstructured.Unstructured, objectKey keys.ClusterWideKey, policy *policyv1alpha1.PropagationPolicy) (err error) {
+func (d *ResourceDetector) ApplyPolicy(object *unstructured.Unstructured, objectKey keys.ClusterWideKey,
+	resourceChangeByKarmada bool, policy *policyv1alpha1.PropagationPolicy) (err error) {
 	start := time.Now()
 	klog.Infof("Applying policy(%s/%s) for object: %s", policy.Namespace, policy.Name, objectKey)
 	var operationResult controllerutil.OperationResult
@@ -424,6 +441,15 @@ func (d *ResourceDetector) ApplyPolicy(object *unstructured.Unstructured, object
 	if err != nil {
 		klog.Errorf("Failed to claim policy(%s) for object: %s", policy.Name, object)
 		return err
+	}
+
+	// If this Reconcile action is triggered by Karmada itself and the current bound Policy is lazy activation preference,
+	// resource will delay to sync the placement from Policy to Binding util resource is updated by User.
+	if resourceChangeByKarmada && util.IsLazyActivationEnabled(policy.Spec.ActivationPreference) {
+		operationResult = controllerutil.OperationResultNone
+		klog.Infof("Skip refresh Binding for the change of resource (%s/%s) is from Karmada and activation "+
+			"preference of current bound policy (%s) is enabled.", object.GetNamespace(), object.GetName(), policy.Name)
+		return nil
 	}
 
 	policyLabels := map[string]string{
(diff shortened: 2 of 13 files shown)


chaosi-zju:
CC @RainbowMango @XiShanYongYe-Chang

pkg/detector/detector.go
```diff
@@ -243,7 +246,7 @@ func (d *ResourceDetector) Reconcile(key util.QueueKey) error {
 			// currently we do that by setting owner reference to derived objects.
 			return nil
 		}
-		klog.Errorf("Failed to get unstructured object(%s), error: %v", clusterWideKey, err)
+		klog.Errorf("Failed to get unstructured object(%s), error: %v", clusterWideKeyWithConfig, err)
```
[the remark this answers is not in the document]
chaosi-zju: it has implemented the type `Stringer` interface.

pkg/apis/policy/v1alpha1/propagation_types.go
```diff
@@ -518,6 +533,18 @@ const (
 	ConflictAbort ConflictResolution = "Abort"
 )
 
+// ActivationStrategy declares how creation/modification/deletion to the Policy affect the resources it matches.
+type ActivationStrategy string
+
+const (
+	// DelayActivation means the matched resource will delay to respond to the creation/modification/deletion until
+	// the resource itself is modified
+	DelayActivation ActivationStrategy = "Delay"
```
RainbowMango, line 542:
```suggestion
	DelayActivation ActivationStrategy = "Lazy"
```

pkg/apis/policy/v1alpha1/propagation_types.go
```diff
@@ -518,6 +533,18 @@ const (
...
+
+	// ImmediateActivation means the matched resource will respond to the creation/modification/deletion immediately.
+	ImmediateActivation ActivationStrategy = "Immediate"
```
RainbowMango, line 545: I think we can remove it for now. And add it when needed.

pkg/apis/policy/v1alpha1/propagation_types.go
```diff
@@ -152,6 +152,21 @@ type PropagationSpec struct {
...
+	// The value "Delay" means the matched resource will delay to respond to the creation/modification/deletion until
+	// the resource itself is modified (In 1:N scenario where one Policy manages many resource templates,
+	// changes to a Policy typically affect numerous applications simultaneously. If there are errors in the Policy
+	// configuration, it could lead to widespread failures. By using this setting to delay the implementation of changes,
+	// the changed Policy can be gradually rolled out through iterative modifications to application configurations).
+	//
+	// If the field is unset, the default strategy is "Immediate".
+	//
+	// +optional
+	ActivationStrategy ActivationStrategy `json:"activationStrategy,omitempty"`
```
RainbowMango, line 169:
```suggestion
	ActivationPreference ActivationPreference `json:"activationPreference,omitempty"`
```

pkg/apis/policy/v1alpha1/propagation_types.go
```diff
@@ -152,6 +152,21 @@ type PropagationSpec struct {
 	// +kubebuilder:validation:Enum=Abort;Overwrite
 	// +optional
 	ConflictResolution ConflictResolution `json:"conflictResolution,omitempty"`
+
+	// ActivationStrategy declares how creation/modification/deletion to the Policy affect the resources it matches.
+	//
+	// The value "Immediate" means the matched resource will respond to the creation/modification/deletion immediately.
+	//
+	// The value "Delay" means the matched resource will delay to respond to the creation/modification/deletion until
+	// the resource itself is modified (In 1:N scenario where one Policy manages many resource templates,
```
chaunceyjiang, line 161: 1. After the cluster label is modified, it triggers a rescheduling. Will the scheduler apply the new PP or old PP?
2. When introducing new features for PP in the future, do we need to consider whether this feature applies to `Delay` or `Immediate`?
3. Is this `activationStrategy` field allowed to be modified?
4. When the `activationStrategy` set to `Delay`, different resources matched by the same PP may apply different versions of PP. How can users distinguish which version of PP is applied to current resources?  -  When scheduling results do not meet expectations, how can users determine whether the scheduling result of current resources is due to a mistake in writing PP or a bug in scheduler?

chaosi-zju: 1. what does `cluster label` mean?
2. Actually, the behavior `Immediate` and `Delay` should be exactly the same, the only diffience should be before judging `CreateOrUpdateResourceBinding` in `Reconcile` of ResourceTemplate, we add a step to judge whether `Reconcile` is triggled by Karmada itself and the current bound Policy is `Delay`, then the resource skip to `CreateOrUpdateResourceBinding`. (It's just that there are some small problems in our code implementation, for example, Karmada add a label named `propagationpolicy.karmada.io/name` on the resource, which is also considered as a resource change, but this is actually Karmada's behaviour, and it should not considered as a resource change.)
3. yes
4. The real policy is in ResourceBinding，PropagationPolicy can be interpreted as a preconfigured Policy. If it is `Immediate`, it will Immediately refresh to ResourceBinding; if it is `Delay`, it will delay to refresh to ResourceBinding util user modified the Resource.

chaunceyjiang: > what does cluster label mean?

https://github.com/karmada-io/karmada/blob/master/pkg/scheduler/event_handler.go#L206


> The real policy is in ResourceBinding，PropagationPolicy can be interpreted as a preconfigured Policy. If it is Immediate, it will Immediately refresh to ResourceBinding; if it is Delay, it will delay to refresh to ResourceBinding util user modified the Resource.


It might be different, not all PP information is in RB.

My core question is: PP has multiple versions, how do users know which version they are currently using?

chaosi-zju: **The `placement` in Binding is real Policy, while `placement` in PropagationPolicy is actually just preconfigured Policy.**

When the matched resource sync the `placement` in preconfigured Policy (PropagationPolicy) to real Policy (Binding), it means Policy take efftect.

When does the Policy take efftect, or saying when the matched resourece do this sync, it makes sense that it can be configured, just by this flied.

Even though PP has multiple versions, resource will always see just one version. Besides, without any doubt, resoure will response to each version Policy's changement, the only difference is before syncing the `placement` from current Policy to Binding (`CreateOrUpdateResourceBinding`), we add a step to check whether the syncing should be done immediately.

If you want to see the real `placement`, just see it in ResourceBinding. 
* The `placement` in PropagationPolicy must be euqal to the `placement` in ResourceBinding in case of the PropagationPolicy's `ActivationPreference` is `Immediate `.
* The `placement` in PropagationPolicy may be different with the `placement` in ResourceBinding in case of the PropagationPolicy's `ActivationPreference` is `Lazy `.

chaosi-zju: Notes: my PR has been updated, the added field has been renamed to `ActivationPreference`

chaunceyjiang: When `ActivationPreference` is `Lazy`, does it mean that the `placement` in PP is Lazy, `Failover`, `ConflictResolution` etc. are `Immediate`, right?

chaosi-zju: > When `ActivationPreference` is `Lazy`, does it mean that the `placement` in PP is Lazy, `Failover`, `ConflictResolution` etc. are `Immediate`, right?

Wrong. But you this comment is very important！！！I should update my code explanation !

`Failover` and `ConflictResolution` is the same as `placement`, this flied both exist in Policy and Binding, should be judge by `ActivationPreference` whether sync it to binding immediately.

A critical sentence is `The impact of Policy changes on the bondage relationship between resources and Policy is always immediate`, so some filed like `ResourceSelector`、`priority`、`premetion` are only works for "the bondage relationship between resources and Policy", since it has been solved, it no need sync to bindding.

So, if a filed in Policy is not works for "the bondage relationship between resources and Policy" and it not exist in binding, we should add it to binding.

chaunceyjiang: >. So, if a filed in Policy is not works for "the bondage relationship between resources and Policy" and it not exist in binding, we should add it to binding.

+1 We can proceed this way first, try it out for a while.

In the end, I suggest adding a featureGate to control this feature because I believe we haven't fully mastered it yet.

chaosi-zju: The existing plan is to add a global default value. If you configured the `ActivationPreference`, use the configured value; if you didn't configure it, use global default value.

Shall we add another featureGate in addition to this global default value?

chaunceyjiang: I can accept this way.

chaosi-zju: any other comments?

chaunceyjiang: >  So, if a filed in Policy is not works for "the bondage relationship between resources and Policy" and it not exist in binding, we should add it to binding.

I suggest that you submit the API modifications in a separate PR, which would make it more convenient for everyone to review.

pkg/apis/policy/v1alpha1/propagation_types.go
```diff
@@ -152,6 +152,23 @@ type PropagationSpec struct {
...
+	// configuration, it could lead to widespread failures. By using this setting to make modified content delay
+	// synchronization from Policy to Binding, the changed Policy can be gradually rolled out through
+	// iterative modifications to resource templates.)
+	//
+	// In a word, The impact of Policy changes on the bondage relationship between resources and Policy is always immediate,
+	// but whether the resource needs to immediately sync the propagation related content to Binding is determined by this field.
+	//
+	// If the field is unset, the default preference is do the sync immediately.
+	//
+	// +optional
```
RainbowMango, line 170:
```suggestion
	// ActivationPreference indicates how the referencing resource template will
	// be propagated, in case of policy changes.
	//
	// If empty, the resource template will respond to policy changes
	// immediately, in other words, any policy changes will drive the resource
	// template to be propagated immediately as per the current propagation rules.
	//
	// If the value is 'Lazy' means the policy changes will not take effect for now
	// but defer to the resource template changes, in other words, the resource
	// template will not be propagated as per the current propagation rules until
	// there is an update on it.
	// This is an experimental feature that might help in a scenario where a policy
	// manages huge amount of resource templates, changes to a policy typically
	// affect numerous applications simultaneously. A minor misconfiguration
	// could lead to widespread failures. With this feature, the change can be
	// gradually rolled out through iterative modifications of resource templates.
	//
	// +kubebuilder:validation:Enum=Lazy
	// +optional
```
How this look like? @chaunceyjiang @chaosi-zju

chaosi-zju: lgtm

pkg/apis/policy/v1alpha1/propagation_types.go
chaosi-zju: This PR is being discussed by some other partners, and it is not convenient for me to move it right away.
I will strictly abide by your this standard next time (ಥ_ಥ) .

pkg/apis/policy/v1alpha1/propagation_types.go
```diff
@@ -152,6 +152,27 @@ type PropagationSpec struct {
...
+	// template will not be propagated as per the current propagation rules until
+	// there is an update on it.
+	// This is an experimental feature that might help in a scenario where a policy
+	// manages huge amount of resource templates, changes to a policy typically
+	// affect numerous applications simultaneously. A minor misconfiguration
+	// could lead to widespread failures. With this feature, the change can be
+	// gradually rolled out through iterative modifications of resource templates.
+	//
+	// +kubebuilder:validation:Enum=Lazy
+	// +optional
```
chaunceyjiang, line 174: This is just a question, have you considered putting this field in `Placement`?

When the user fills in `Placement`, the format they fill in is as follows:
```
spec:
  placement:
    activationPreference: lazy
    clusterAffinity: 
      clusterNames:
      - member1
```

chaosi-zju: I have considered it.

However, since the target user of this scenario expects that any modification he makes to the Policy should not trigger a change to the `work`, the semantics of this `lazy` does not only involve `placement`, but other fields such as `conflictResolution` should also follow the `lazy`.

chaunceyjiang: Got it.

RainbowMango (approved):
/lgtm
/approve

chaunceyjiang:
I found that deployment behaves differently from configMap. When I use the following PP, the deployment can propagate to member clusters, but the configMap cannot.

Note: Only apply this file.

``` yaml

kind: PropagationPolicy
metadata:
  name: nginx-propagation
spec:
  activationPreference: Lazy
  resourceSelectors:
    - apiVersion: apps/v1
      kind: Deployment
      name: nginx
    - apiVersion: v1
      kind: ConfigMap
      name: game-demo
  placement:
    clusterAffinity:
      clusterNames:
        - member1
        - member2
    replicaScheduling:
      weightPreference:
        staticWeightList:
          - targetCluster:
              clusterNames:
                - member1
            weight: 1
          - targetCluster:
              clusterNames:
                - member2
            weight: 1
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: game-demo
data:
  # property-like keys; each key maps to a simple value
  player_initial_lives: "3"
  ui_properties_file_name: "user-interface.properties"

  # file-like keys
  game.properties: |
    enemy.types=aliens,monsters
    player.maximum-lives=5
  user-interface.properties: |
    color.good=purple
    color.bad=yellow
    allow.textmode=true
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx
  labels:
    app: nginx
spec:
  replicas: 2
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - image: nginx
        name: nginx

```

chaosi-zju:
> I found that deployment behaves differently from configMap. When I use the following PP, the deployment can propagate to member clusters, but the configMap cannot.

I think this may because the detail resource applied in order, just like: configmap --> pp --> deploy

our principle is `creation/updation/deletion of pp shouldn't actively triggle the propagation change of the resource`. So:

If the resource create before pp, after pp created, the resource keep unchanged.

If the pp create before resource, after resource created, the resource propagated by the pp.

chaosi-zju:
> I found that deployment behaves differently from configMap. When I use the following PP, the deployment can propagate to member clusters, but the configMap cannot.

If you use `Lazy` policy, you should understand PropagationPolicy is a preconfigured policy.

If the pp and resource are all not exist, you should apply pp first, then to apply resource.

chaunceyjiang:
> you should apply pp first, then to apply resource.


So, I think this is the issue, it requires users to perceive the apply order forcibly. Obviously, in a real-world environment, it's hard for us to determine the create order of users.

chaosi-zju:
> > you should apply pp first, then to apply resource.
> 
> So, I think this is the issue, it requires users to perceive the apply order forcibly. Obviously, in a real-world environment, it's hard for us to determine the create order of users.

yes, policy should create before resource.

chaosi-zju:
> > you should apply pp first, then to apply resource.
> 
> So, I think this is the issue, it requires users to perceive the apply order forcibly. Obviously, in a real-world environment, it's hard for us to determine the create order of users.

yes, lazy policy should apply before resource.

when lazy policy created, it will not triggle resource in waiting list to refresh binding (it only triggle `ClaimPolicyForObject` to establish boundge relationship, but not triggle `CreateOrUpdateResourceBinding`)

resource in waiting list can also has `ResourceBiding` and `Work`, the related `pods` running just fine, creation of lazy policy shouldn't change its propagation states.

chaunceyjiang:
> yes, lazy policy should apply before resource.

It's really difficult for users.

chaosi-zju:
> It's really difficult for users.

Is there a compromise?

PR #4586, opened 2024-01-26
fix: reschedule the replicas of the disappear clusters in PP

### LINKED ISSUE #4560: Can't re-schedule the replicas to new clusters within updating PP
**What happened**:
1. Create following PP and Deployment
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx
  labels:
    app: nginx
spec:
  replicas: 2
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - image: nginx
        name: nginx
---
kind: PropagationPolicy
metadata:
  name: nginx-propagation
spec:
  resourceSelectors:
    - apiVersion: apps/v1
      kind: Deployment
      name: nginx
  placement:
    clusterAffinity:
      clusterNames:
        - member1
    replicaScheduling:
      weightPreference:
```
2. The 2 replicas will be launched in member1
3. Update PP, and change member1 to member2
```yaml
...
  placement:
    clusterAffinity:
      clusterNames:
      - member2
...
```
4. The replicas are not launch in member2 cluster
```sh
root@karmada [02:14:19 PM] [~/workspace/git/karmada/samples/nginx] [master *]
-> # karmadactl get po
NAME                     CLUSTER   READY   STATUS    RESTARTS   AGE
nginx-77b4fdf86c-t9z5h   member1   1/1     Running   0          143m
nginx-77b4fdf86c-vtztd   member1   1/1     Running   0          143m
```

The replicas in member1 will be deleted and relaunched in member2

as described before

Related code: https://github.com/karmada-io/karmada/blame/0e501b607a95bccf909efcaef95a694244ef4feb/pkg/scheduler/core/assignment.go#L166

**Environment**:
- Karmada version: master branch
- kubectl-karmada or karmadactl version (the result of `kubectl-karmada version` or `karmadactl version`): master branch
- Others:none

jwcesign:
**What type of PR is this?**
/kind bug

Please see https://github.com/karmada-io/karmada/issues/4560

Fixes #4560

Please note：
As described in: https://github.com/karmada-io/karmada/issues/4560#issuecomment-1911303202
Due to the changes in the score plugin, the scheduling results of instance counts may oscillate among multiple clusters.

We should find a better way to fix this in the future.

none

```release-note
`karmada-scheduler`: reschedule the replicas of the disappear clusters in PP/CPP.
```

### DIFF
--- a/pkg/scheduler/core/assignment.go
+++ b/pkg/scheduler/core/assignment.go
@@ -90,7 +90,21 @@ func newAssignState(candidates []*clusterv1alpha1.Cluster, placement *policyv1al
 }
 
 func (as *assignState) buildScheduledClusters() {
-	as.scheduledClusters = as.spec.Clusters
+	candidateClusterSet := sets.Set[string]{}
+	for _, c := range as.candidates {
+		candidateClusterSet.Insert(c.Name)
+	}
+	as.scheduledClusters = []workv1alpha2.TargetCluster{}
+	for _, c := range as.spec.Clusters {
+		// Ignore clusters that are no longer candidates, to ensure we can get real
+		// 'assigned' replicas from the previous schedule result. The ignored replicas
+		// will be treated as scaled-up replicas that will be assigned to other
+		// candidate clusters.
+		if !candidateClusterSet.Has(c.Name) {
+			continue
+		}
+		as.scheduledClusters = append(as.scheduledClusters, c)
+	}
 	as.assignedReplicas = util.GetSumOfReplicas(as.scheduledClusters)
 }
 
--- a/pkg/scheduler/core/assignment_test.go
+++ b/pkg/scheduler/core/assignment_test.go
@@ -580,8 +580,8 @@ func Test_dynamicScale(t *testing.T) {
 			},
 			want: []workv1alpha2.TargetCluster{
 				{Name: ClusterMember1, Replicas: 7},
-				{Name: ClusterMember2, Replicas: 8},
-				{Name: ClusterMember4, Replicas: 9},
+				{Name: ClusterMember3, Replicas: 6},
+				{Name: ClusterMember4, Replicas: 11},
 			},
 			wantErr: false,
 		},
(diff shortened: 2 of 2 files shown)


pkg/scheduler/core/generic_scheduler.go
```diff
@@ -94,20 +94,20 @@ func (g *genericScheduler) Schedule(
...
-	clustersWithReplicas, err := g.assignReplicas(clusters, spec.Placement, spec)
+	clustersWithReplicas, err := g.assignReplicas(candidateClusters, feasibleClusters, spec.Placement, spec)
 	if err != nil {
 		return result, fmt.Errorf("failed to assign replicas: %w", err)
 	}
 	klog.V(4).Infof("Assigned Replicas: %v", clustersWithReplicas)
 
 	if scheduleAlgorithmOption.EnableEmptyWorkloadPropagation {
-		clustersWithReplicas = attachZeroReplicasCluster(clusters, clustersWithReplicas)
+		clustersWithReplicas = attachZeroReplicasCluster(candidateClusters, clustersWithReplicas)
```
zhzhuang-zju, line 110: thanks for your work~I have some questions
- Why is the input parameter of func `attachZeroReplicasCluster` is `candidateClusters` instead of `feasibleClusters`?
- In the scenario of using spreadConstraints with maxGroup set to 2,
 if the previous scheduling result is [member1 replicas 2, member2 replicas 1], 
 feasibleClusters are [member1, member2, member3],
 candidateClusters are [member1, member3], 
 the final assign result is [member1 replicas 2, member2 replicas 1]. 
 In this case, the return result of this function is [member1 replicas 2, member2 replicas 1, member3 replicas 0]. The length exceeds maxGroup. Will it cause misunderstanding?

zhzhuang-zju (commented):
others LGTM

zhzhuang-zju:
Does this PR also affect Cpp? If so, we could include it in the release note.

RainbowMango (commented):
/assign

pkg/scheduler/core/assignment.go
```diff
@@ -90,7 +90,17 @@ func newAssignState(candidates []*clusterv1alpha1.Cluster, placement *policyv1al
...
 
 func (as *assignState) buildScheduledClusters() {
-	as.scheduledClusters = as.spec.Clusters
+	validClusterSet := sets.Set[string]{}
+	for _, c := range as.candidates {
+		validClusterSet.Insert(c.Name)
+	}
+	as.scheduledClusters = []workv1alpha2.TargetCluster{}
+	for _, c := range as.spec.Clusters {
+		if !validClusterSet.Has(c.Name) {
```
RainbowMango, line 99: It would be great to have a comment here, like:
```
		// Ignore clusters that are no longer candidates, to ensure we can get real
		// 'assigned' replicas from the previous schedule result. The ignored replicas
		// will be treated as scaled-up replicas that will be assigned to other
		// candidate clusters.
```

pkg/scheduler/core/assignment.go
```diff
@@ -90,7 +90,17 @@ func newAssignState(candidates []*clusterv1alpha1.Cluster, placement *policyv1al
 }
 
 func (as *assignState) buildScheduledClusters() {
-	as.scheduledClusters = as.spec.Clusters
+	validClusterSet := sets.Set[string]{}
```
RainbowMango, line 93:
```suggestion
	candidateClusterSet := sets.Set[string]{}
```

pkg/scheduler/core/assignment_test.go
```diff
@@ -580,8 +580,8 @@ func Test_dynamicScale(t *testing.T) {
 			},
 			want: []workv1alpha2.TargetCluster{
 				{Name: ClusterMember1, Replicas: 7},
-				{Name: ClusterMember2, Replicas: 8},
```
RainbowMango, line 583: This case is not right, as cluster 2 already disappeared, it should not be in the schedule result.

RainbowMango: If this is a legacy issue, please track it by another issue.

jwcesign: The calculation logic is as follows:
1. Calculate the number of instances that need to be reallocated: 24-4 = 20.
2. Sort the member clusters by weight: m4:m3:m1=12:8:4
3. Allocate the number of instances: m4：20/24*12=10, m3:20/24*8=6, m1:4+20/24*4=7
4. Distribute the remainder 1, resulting in m4：11, m3:6, m1:7

RainbowMango (approved):
/lgtm
/approve

PR #4602, opened 2024-02-01
Proposal of the LazyActivation preference for Policy

chaosi-zju:
**What type of PR is this?**

/kind design
/kind documentation

In the scenario where `Policy` responsibilities are separated from `Resource` responsibilities,  one role is dedicated to `Policy` management, which is referred to as cluster administrator, and the other role is dedicated to `Resource` management, which is referred to as user. The cluster administrator would preconfigure some `Policies` for users, then the users apply their own `Resources` and propagated them according to the preconfigured `Policy`.

However, the administrator has requests to modify the `Policy` in the future, such as migrating propagation clusters. As the current `Policy` modification is effective immediately, the modification will affect the propagation status of a large number of resource templates and have a great impact on the system.

The administrator is worried that the modification will lead to the failure of the business application in the non-change window, so the administrator hopes that the modification of `Policy` can be delayed until the business application change window takes effect.

This proposal introduced a `ActivationPreference` filed to Policy, indicating how the referencing resource template will be propagated, in case of policy changes. 

If empty, the resource template will respond to policy changes immediately, in other words, any policy changes will drive the resource template to be propagated immediately as per the current propagation rules.

If the value is `Lazy` means the policy changes will not take effect for now but defer to the resource template changes, in other words, the resource template will not be propagated as per the current propagation rules until there is an update on it.

Fixes part of #4563

The most fundamental criterion is:

* **Policy changes cannot actively cause changes to the propagation status of resource templates.**

The system behavior can be concluded as one-sentence summary: 

* **When a user creates/deletes/modifies a Policy, if the currently active Policy is `Lazy`, it will not take effect immediately.**

Another key guideline is:

* **When Policy changes, modifications involving the binding relationship between Policy and resources are always processed immediately. Only changes involving the propagation status of resources can be `Lazy`.**

```release-note
none
```

### DIFF
--- a/docs/proposals/scheduling/activation-preference/lazy-activation-preference.md
+++ b/docs/proposals/scheduling/activation-preference/lazy-activation-preference.md
@@ -1,0 +1,462 @@
+---
+title: Introduce a lazy activation preference to Policy
+authors:
+  - "@chaosi-zju"
+reviewers:
+  - "@RainbowMango"
+  - "@chaunceyjiang"
+  - "TBD"
+approvers:
+  - "@RainbowMango"
+  - "TBD"
+
+creation-date: 2024-01-30
+---
+
+# Introduce a lazy activation preference to Policy
+
+## Background
+
+In the scenario where `Policy` responsibilities are separated from `Resource` responsibilities, 
+one role is dedicated to `Policy` management, which is referred to as cluster administrator, 
+and the other role is dedicated to `Resource` management, which is referred to as user. 
+The cluster administrator would preconfigure some `Policies` for users, then the users apply their own `Resources` and 
+propagated them according to the preconfigured `Policy`.
+
+However, the administrator has requests to modify the `Policy` in the future, such as migrating propagation clusters.
+As the current `Policy` modification is effective immediately, the modification will affect the propagation status of
+a large number of resource templates and have a great impact on the system.
+
+The administrator is worried that the modification will lead to the failure of the business application in the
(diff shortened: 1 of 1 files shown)


chaosi-zju:
CC @chaunceyjiang

docs/proposals/scheduling/activation-preference/lazy-activation-preference.md
```diff
@@ -0,0 +1,459 @@
...
+the business application change window takes effect.
+
+### Goals
+
+The most fundamental criterion is:
+
+* **Policy changes cannot actively cause changes to the propagation status of resource templates.**
+
+Under this criterion, a new delayed effective mechanism is introduced into `Policy`, 
+by which the referenced resource template will delay responding to the change of `Policy` until the resource template itself changes.
```
chaunceyjiang, line 41: > until the resource template itself changes.


Perhaps further clarification is needed on what exactly `changes` refers to here. Is it `resourceVersion`, or `generation`, or some other situation?

chaosi-zju: done, I have added the description to the proposal.

> changes of resource template refer to any filed is modified except `status`, `metadata.managedFields`, `metadata.resourceVersion`, `metadata.generation`, or the label / annotation whose keys are prefixed with `.karmada.io`.

RainbowMango (approved):
/lgtm
/approve

PR #4620, opened 2024-02-08
`karmadactl`: Add the reserved label `karmada.io/system` to resources created by the `join` command

a7i:
**What type of PR is this?**

/kind feature

Follow-up to https://github.com/karmada-io/karmada/pull/3262

Adds the well-known "karmada.io/managed" label to resources created by karmada controllers

We need a way to bypass creation of some ClusterRole/Rolebindings by our OPA Gatekeeper and we don't have a consistent label to select on. When we join a cluster via push command, it creates ClusterRole/ClusterRoleBinding that are too open (`*` and `*`) which is by default blocked by our opa-gatekeeper policies. This allows us to bypass it by the karmada managed label.

Fixes #

```release-note
`karmadactl`: Add the reserved label `karmada.io/system` to resources created by the `join` command.
```

### DIFF
--- a/pkg/karmadactl/register/register.go
+++ b/pkg/karmadactl/register/register.go
@@ -378,7 +378,7 @@ func (o *CommandRegisterOption) Run(parentCommand string) error {
 
 	// It's necessary to set the label of namespace to make sure that the namespace is created by Karmada.
 	labels := map[string]string{
-		karmadautil.ManagedByKarmadaLabel: karmadautil.ManagedByKarmadaLabelValue,
+		karmadautil.KarmadaSystemLabel: karmadautil.KarmadaSystemLabelValue,
 	}
 	// ensure namespace where the karmada-agent resources be deployed exists in the member cluster
 	if _, err := karmadautil.EnsureNamespaceExistWithLabels(o.memberClusterClient, o.Namespace, o.DryRun, labels); err != nil {
@@ -523,6 +523,9 @@ func (o *CommandRegisterOption) constructKarmadaAgentConfig(bootstrapClient *kub
 	certificateSigningRequest := &certificatesv1.CertificateSigningRequest{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: csrName,
+			Labels: map[string]string{
+				karmadautil.KarmadaSystemLabel: karmadautil.KarmadaSystemLabelValue,
+			},
 		},
 		Spec: certificatesv1.CertificateSigningRequestSpec{
 			Request: pem.EncodeToMemory(&pem.Block{
@@ -591,6 +594,11 @@ func (o *CommandRegisterOption) createSecretAndRBACInMemberCluster(karmadaAgentC
 		return fmt.Errorf("failure while serializing karmada-agent kubeConfig. %w", err)
 	}
 
+	// It's necessary to set the label of namespace to make sure that the namespace is created by Karmada.
+	labels := map[string]string{
+		karmadautil.KarmadaSystemLabel: karmadautil.KarmadaSystemLabelValue,
+	}
+
 	kubeConfigSecret := &corev1.Secret{
 		TypeMeta: metav1.TypeMeta{
 			APIVersion: "v1",
@@ -599,6 +607,7 @@ func (o *CommandRegisterOption) createSecretAndRBACInMemberCluster(karmadaAgentC
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      KarmadaKubeconfigName,
 			Namespace: o.Namespace,
+			Labels:    labels,
 		},
 		Type:       corev1.SecretTypeOpaque,
 		StringData: map[string]string{KarmadaKubeconfigName: string(configBytes)},
@@ -611,7 +620,8 @@ func (o *CommandRegisterOption) createSecretAndRBACInMemberCluster(karmadaAgentC
 
 	clusterRole := &rbacv1.ClusterRole{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: KarmadaAgentName,
+			Name:   KarmadaAgentName,
+			Labels: labels,
 		},
 		Rules: []rbacv1.PolicyRule{
 			{
@@ -635,6 +645,7 @@ func (o *CommandRegisterOption) createSecretAndRBACInMemberCluster(karmadaAgentC
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      KarmadaAgentServiceAccountName,
 			Namespace: o.Namespace,
+			Labels:    labels,
 		},
 	}
 
@@ -646,7 +657,8 @@ func (o *CommandRegisterOption) createSecretAndRBACInMemberCluster(karmadaAgentC
 
 	clusterRoleBinding := &rbacv1.ClusterRoleBinding{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: KarmadaAgentName,
+			Name:   KarmadaAgentName,
+			Labels: labels,
 		},
 		RoleRef: rbacv1.RoleRef{
 			APIGroup: "rbac.authorization.k8s.io",
--- a/pkg/util/constants.go
+++ b/pkg/util/constants.go
@@ -52,8 +52,13 @@ const (
 	// FederatedResourceQuotaNameLabel is added to Work to specify associated FederatedResourceQuota's name.
 	FederatedResourceQuotaNameLabel = "federatedresourcequota.karmada.io/name"
 
-	// ManagedByKarmadaLabel is a reserved karmada label to indicate whether resources are managed by karmada controllers.
+	// ManagedByKarmadaLabel is a reserved karmada label to indicate whether resources are member cluster resources
+	// synchronized by karmada controllers.
 	ManagedByKarmadaLabel = "karmada.io/managed"
+
+	// KarmadaSystemLabel is a reserved karmada label to indicate whether resources are system level resources
+	// managed by karmada controllers.
+	KarmadaSystemLabel = "karmada.io/system"
 
 	// EndpointSliceDispatchControllerLabelValue indicates the endpointSlice are controlled by Karmada
 	EndpointSliceDispatchControllerLabelValue = "endpointslice-dispatch-controller.karmada.io"
@@ -71,8 +76,11 @@ const (
 )
 
 const (
-	// ManagedByKarmadaLabelValue indicates that resources are managed by karmada controllers.
+	// ManagedByKarmadaLabelValue indicates that these are workloads in member cluster synchronized by karmada controllers.
 	ManagedByKarmadaLabelValue = "true"
+
+	// KarmadaSystemLabelValue indicates that resources are system level resources managed by karmada controllers.
+	KarmadaSystemLabelValue = "true"
 
 	// RetainReplicasValue is an optional value of RetainReplicasLabel, indicating retain
 	RetainReplicasValue = "true"
--- a/pkg/util/credential.go
+++ b/pkg/util/credential.go
@@ -59,7 +59,7 @@ func ObtainCredentialsFromMemberCluster(clusterKubeClient kubeclient.Interface, 
 	var err error
 	// It's necessary to set the label of namespace to make sure that the namespace is created by Karmada.
 	labels := map[string]string{
-		ManagedByKarmadaLabel: ManagedByKarmadaLabelValue,
+		KarmadaSystemLabel: KarmadaSystemLabelValue,
 	}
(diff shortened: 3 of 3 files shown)


grosser (approved):
✨

pkg/util/credential.go
```diff
@@ -48,6 +48,11 @@ var (
 			Verbs:           []string{"get"},
 		},
 	}
+
+	// managedByKarmadaLabels: set the label to indicate that the resource is created by Karmada.
+	managedByKarmadaLabels = map[string]string{
```
zhzhuang-zju, line 53: The variable `managedByKarmadaLabels` is extracted here. Would it be better if it was exported so that other files can also use it? Like, https://github.com/karmada-io/karmada/blob/f03b5e9803e4738706bb6fcf807350a55a4ad703/pkg/controllers/certificate/cert_rotation_controller.go#L239-L245 could be changed to 
```go
	certificateSigningRequest := &certificatesv1.CertificateSigningRequest{
			Name: csrName,
			Labels: util.ManagedByKarmadaLabels,
			},
		},
```

pkg/util/credential.go
```diff
@@ -57,20 +62,20 @@ func ObtainCredentialsFromMemberCluster(clusterKubeClient kubeclient.Interface,
...
+	if _, err = EnsureNamespaceExistWithLabels(clusterKubeClient, opts.ClusterNamespace, opts.DryRun, managedByKarmadaLabels); err != nil {
 		return nil, nil, err
 	}
 
 	if opts.IsKubeImpersonatorEnabled() {
 		// create a ServiceAccount for impersonation in cluster.
-		impersonationSA := &corev1.ServiceAccount{}
-		impersonationSA.Namespace = opts.ClusterNamespace
-		impersonationSA.Name = names.GenerateServiceAccountName("impersonator")
+		impersonationSA := &corev1.ServiceAccount{
```
chaunceyjiang, line 72: https://github.com/karmada-io/karmada/blob/adef1e59748e1e1d31cb75fffe406b5dd69a66d7/pkg/util/serviceaccount.go#L104-L114
Does the `secret` here also need to be handled in the same way?

RainbowMango: Nice finding! I think Yes.

chaunceyjiang: /cc @a7i

XiShanYongYe-Chang: Kindly ping @a7i

a7i: nice catch, let me fix that and rebase

pkg/util/credential.go
a7i: Good suggestion, implemented ✅

pkg/util/credential.go
a7i: Implemented ✅

pkg/karmadactl/register/register.go
```diff
@@ -378,9 +378,7 @@ func (o *CommandRegisterOption) Run(parentCommand string) error {
 	}
 
 	// It's necessary to set the label of namespace to make sure that the namespace is created by Karmada.
-	labels := map[string]string{
-		util.ManagedByKarmadaLabel: util.ManagedByKarmadaLabelValue,
-	}
+	labels := util.ManagedByKarmadaLabels
```
a7i, line 381: left this since it goes nicely with the comment above it

XiShanYongYe-Chang:
Hi @chaunceyjiang, can you help take a review again?

chaunceyjiang (commented):
Thanks~


/LGTM

XiShanYongYe-Chang (commented):
/lgtm
/cc @RainbowMango

RainbowMango (commented):
/assign
Another PR I missed :( :(   I'll look at it ASAP.

RainbowMango (approved):
/lgtm
/approve

Forced pushed to solve conflicts and updated release-notes.

pkg/controllers/certificate/cert_rotation_controller.go
```diff
@@ -238,7 +238,8 @@ func (c *CertRotationController) createCSRInControlPlane(clusterName string, pri
 
 	certificateSigningRequest := &certificatesv1.CertificateSigningRequest{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: csrName,
+			Name:   csrName,
+			Labels: util.ManagedByKarmadaLabels,
```
whitewindmills, line 242: /hold
since it's ceated in the Karmada control-plane, there is no need to add this label here.

pkg/controllers/cluster/cluster_controller.go
```diff
@@ -422,7 +422,8 @@ func (c *Controller) createExecutionSpace(cluster *clusterv1alpha1.Cluster) erro
 		// create only when not exist
 		executionSpace := &corev1.Namespace{
 			ObjectMeta: metav1.ObjectMeta{
-				Name: executionSpaceName,
+				Name:   executionSpaceName,
+				Labels: util.ManagedByKarmadaLabels,
```
whitewindmills, line 426: ditto

whitewindmills (commented):
isn’t this label `karmada.io/managed` used to add to resources in the member cluster? this indicates that the resources on the member cluster are managed by Karmada, so Karmada can update or delete them.
https://github.com/karmada-io/karmada/blob/b217a9ce2d6cc042725d370952696c31d03778ca/pkg/util/objectwatcher/objectwatcher.go#L303-L305

pkg/controllers/certificate/cert_rotation_controller.go
RainbowMango: Good point!
@a7i do you agree with that?

a7i: when I added this label, my intention was to exclude it from opa-gatekeepr policies. for our use-case, we still run opa at the api level (otherwise, the rollouts would pass, but propagation would fail).

XiShanYongYe-Chang:
Kindly ping @a7i

a7i:
Hi @XiShanYongYe-Chang I commented [here](https://github.com/karmada-io/karmada/pull/4620#discussion_r1603280104) on the rationale / use-case

grosser:
@XiShanYongYe-Chang and @whitewindmills 
would love to get this moving again
... we need a label on these resources, it does not really matter which
afaik ManagedByKarmadaLabelValue would make sense, but if we can unblock with a new "ManagedByKarmadaControllerLabel" then that's also fine, just need a clear decision

a7i:
happy to adjust the PR based on feedback from karmada team.

- do you suggest adding a new label? if so, what should key/value be?

XiShanYongYe-Chang:
We need to express two semantics, one to indicate that resources in the member clusters are managed by the Karmada system, and the other to indicate that resources on the Karmada control plane are managed by the Karmada system. I propose to distinguish between these two cases, so I came up with two ways:
1. Use two labels to represent this, for example: `karmada.io/system: true` and `karmada.io/managed: true`
2. Use different label values to distinguish it, for example: `karmada.io/manager: member-cluster` and `karmada.io/manager: control-plane`

Ask more ideas and suggestions from @chaunceyjiang @RainbowMango @whitewindmills @zhzhuang-zju

whitewindmills:
solution 1 looks good to me.

RainbowMango:
I'll revisit this tomorrow and get back to you then.

> In addition, if we have another choice for OPA to bypass?

What I'm thinking is if we want a label we should make sure every resource that was created during the joining process should have it, not just those credentials resources.

RainbowMango:
> 1 Use two labels to represent this, for example: karmada.io/system: true and karmada.io/managed: true
2 Use different label values to distinguish it, for example: karmada.io/manager: member-cluster and karmada.io/manager: control-plane

I vote for option 1, too.

Please @a7i update this PR if you also agree with it.

XiShanYongYe-Chang:
Kindly ping @a7i 

I will also update the labels on the Endpointslice collected on the control plane in sync.

a7i:
@XiShanYongYe-Chang no rush, but should be ready for your review

XiShanYongYe-Chang:
OK, thanks~
/retest

XiShanYongYe-Chang (commented):
Thanks~

pkg/util/constants.go
```diff
@@ -238,6 +249,21 @@ const (
...
+	}
+
+	// KarmadaSystemLabels is map of reserved karmada label to indicate whether resources are system level
+	// resources managed by karmada controllers.
+	KarmadaSystemLabels = map[string]string{
+		KarmadaSystemLabel: KarmadaSystemLabelValue,
+	}
+
+	// KarmadaAgentLabels is map of reserved karmada label for karmada-agent deployment.
+	KarmadaAgentLabels = map[string]string{"app": KarmadaAgentName}
```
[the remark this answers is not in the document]
a7i: Implemented per this comment from five months ago: https://github.com/karmada-io/karmada/pull/4620#discussion_r1493130385

Happy to revert to the way it was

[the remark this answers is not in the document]
XiShanYongYe-Chang: I suggest using the way it was before the PR was submitted.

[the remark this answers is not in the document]
a7i: so to be clear:

re: `karmadaAgentLabels` was a map before the PR was submitted so leave as-is: https://github.com/karmada-io/karmada/blob/71de3dc02247e22da249c285fa0c0c032ae01445/pkg/karmadactl/register/register.go#L115

re: `KarmadaManagedLabels` and disregard @zhzhuang-zju suggestion and create maps inline as opposed to reusing one?

re: `KarmadaSystemLabels` follow managed labels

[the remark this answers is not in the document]
zhzhuang-zju: @a7i @XiShanYongYe-Chang It's great to have a discussion about how to make the code better~ In fact, this question can be divided into two parts: 1. whether it is necessary to wrap the label as a map. 2. if it is wrapped as a map, whether it is necessary to export it. refer to https://github.com/karmada-io/karmada/pull/4620#discussion, I meant that if it is wrapped as a map, it is possible to export it for better scalability. I think what @XiShanYongYe-Chang means is that step 1 is not necessary because there is only one label, so there is no need to wrap it into a map. That's something I agree with, and I think the final opinion is that we can leave merged code alone, and the new code does not need to wrap the label into a map again. What do you think?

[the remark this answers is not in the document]
a7i: Can you please be more specific with the three labels to avoid back and forth?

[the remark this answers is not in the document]
XiShanYongYe-Chang: Just make it keep with before will be ok, just like this:
```go
var karmadaAgentLabels = map[string]string{"app": KarmadaAgentName}

func XXX(){
       ...
	        Name:      KarmadaAgentName,
	        Namespace: o.Namespace,
        	Labels:    karmadaAgentLabels,
        },
       ...
}
```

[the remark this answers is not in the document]
zhzhuang-zju: combine with https://github.com/karmada-io/karmada/pull/4620#discussion_r1673916240 and https://github.com/karmada-io/karmada/pull/4620#discussion_r1673921429 is ok. @a7i ,If we didn't make it clear, please let us know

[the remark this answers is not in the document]
a7i: >just replaced with labelName: labelvalue wherever used.

It's not very DRY, but at this point it's been 5 months and I'd like to get the PR moving. Commit coming up soon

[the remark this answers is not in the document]
a7i: I hope I got it right this time 🤞🏼 

ready for your review

XiShanYongYe-Chang:
Thanks~
/lgtm

zhzhuang-zju:
/lgtm
and failed e2e cases also rerun successfully~
cc @whitewindmills I observed that you added the `/hold` label earlier, please take a look at it again for any other comments? Note that once `hold` is canceled, pr will get merged since there is `approve` label

pkg/util/constants.go
```diff
@@ -52,9 +52,14 @@ const (
 	// FederatedResourceQuotaNameLabel is added to Work to specify associated FederatedResourceQuota's name.
 	FederatedResourceQuotaNameLabel = "federatedresourcequota.karmada.io/name"
 
-	// ManagedByKarmadaLabel is a reserved karmada label to indicate whether resources are managed by karmada controllers.
+	// ManagedByKarmadaLabel is a reserved karmada label to indicate whether resources are member cluster resources
```
whitewindmills, line 55: we'd better keep the first letter capitalized for `Karmada` in comments. others look good to me.
@RainbowMango do you wanna review it again?

RainbowMango: no big deal.
I'm looking at it.

PR #822, opened 2021-10-16
replace deprecated package io/ioutil with packages io and os

gy95:
Signed-off-by: guoyao <1015105054@qq.com>

**What type of PR is this?**

**What this PR does / why we need it**:
package `io/ioutil` is deprecated in Go1.16, and according to the official doc(https://golang.org/doc/go1.16#ioutil), we could use definitions in `io` and `os`.
**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #4640, opened 2024-02-20
chores(deps): support depguard rules in golang-lint

dongjiang1989:
**What type of PR is this?**
/kind deprecation

ref: 
#2717  
#822 

Add static check to report the deprecated warnings

Fixes #

```release-note
NONE(user not facing this change)
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -48,8 +48,15 @@ linters:
   - misspell
   - whitespace
   - revive
+  - depguard
 
 linters-settings:
+  depguard:
+    rules:
+      main:
+        deny:
+        - pkg: "io/ioutil"
+          desc: "Use corresponding 'os' or 'io' functions instead."
   goimports:
     local-prefixes: github.com/karmada-io/karmada
   gocyclo:
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -56,7 +56,7 @@ linters-settings:
       main:
         deny:
         - pkg: "io/ioutil"
-          desc: "Use corresponding 'os' or 'io' functions instead."
+          desc: "replaced by io and os packages since Go 1.16: https://tip.golang.org/doc/go1.16#ioutil"
   goimports:
     local-prefixes: github.com/karmada-io/karmada
   gocyclo:


.golangci.yml
```diff
@@ -48,8 +48,15 @@ linters:
...
   - revive
+  - depguard
 
 linters-settings:
+  depguard:
+    rules:
+      main:
+        deny:
+        - pkg: "io/ioutil"
+          desc: "Use corresponding 'os' or 'io' functions instead."
```
RainbowMango, line 59:
```suggestion
          desc: "replaced by io and os packages since Go 1.16: https://tip.golang.org/doc/go1.16#ioutil"
```
How about this description? This would be more helpful for developers to fix lint warnings.
Copied from https://github.com/OpenPeeDeeP/depguard?tab=readme-ov-file#golangci-lint.

RainbowMango (commented):
Thanks @dongjiang1989 for doing this. 

My concern is should we maintain a list of deprecated packages? As usually any deprecated packages will be denied by `staticcheck`, (but the check has been disabled due to a widely warnings, see https://github.com/karmada-io/karmada/pull/4640/files#diff-6179837f7df53a6f05c522b6b7bb566d484d5465d9894fb04910dd08bb40dcc9R101-R104).

.golangci.yml
dongjiang1989: Great.

dongjiang1989:
Thanks, @RainbowMango Please re-check.

RainbowMango:
How do you think the concern I mentioned above https://github.com/karmada-io/karmada/pull/4640#pullrequestreview-1889411264?

It seems not right to maintain the list of deprecated packages, as all deprecated packages are not allowed by default.
In the future, we will encounter more deprecated packages, do we need to add them to this list every time?

dongjiang1989:
> Thanks for your quick response. How do you think the concern I mentioned above [#4640 (review)](https://github.com/karmada-io/karmada/pull/4640#pullrequestreview-1889411264)?
> 
> It seems not right to maintain the list of deprecated packages, as all deprecated packages are not allowed by default. In the future, we will encounter more deprecated packages, do we need to add them to this list every time?

Hmmm.... It is necessary to maintain a certain range of deprecated packages or recommended packages. This has a lot to do with the golang version in `go.mod`, and it also standardizes the selection of basic libraries for developers to use.

like:
```yaml
linters-settings:
  depguard:
    rules:
      main:
        deny:
        - pkg: "sync/atomic"
          desc: "Use go.uber.org/atomic instead of sync/atomic"
        - pkg: "github.com/stretchr/testify/assert"
          desc: "Use github.com/stretchr/testify/require instead of github.com/stretchr/testify/assert"
        - pkg: "regexp"
          desc: "Use github.com/grafana/regexp instead of regexp"
        - pkg: "github.com/pkg/errors"
          desc: "Use 'errors' or 'fmt' instead of github.com/pkg/errors"
        ....
```

RainbowMango:
> it also standardizes the selection of basic libraries for developers to use.

That makes more sense than the deprecated thing. 

/lgtm
/approve
Thanks.

PR #4647, opened 2024-02-21
feat: support wildcard `Kind` in `resourceSelectors` of `PropgationPolicy`

a7i:
**What type of PR is this?**

/kind feature

It would be ideal for karmada to allow all `Kind` from a certain `apiGroup`. One use-case is opa-gatekeeper that treats each policy as a constraint `Kind`.

Our cluster currently has 68 constraint kinds and counting:
```
kubectl api-resources | grep constraints.gatekeeper.sh/v1beta1 | wc -l
      68
```

Adding each kind is not easily maintainable and ability to provide a wildcard kind support would reduce our overhead significantly.

Fixes #

```release-note
feat: support wildcard `Kind` in `resourceSelectors` of `PropgationPolicy`
```

### DIFF
--- a/pkg/util/selector.go
+++ b/pkg/util/selector.go
@@ -43,6 +43,8 @@ const (
 	PriorityMatchLabelSelector
 	// PriorityMatchName means the Name of ResourceSelector matches the resource.
 	PriorityMatchName
+	// wildcardKind indicates that any kinds within an apiGroup can be matched for a policy
+	wildcardKind = "*"
 )
 
 // ResourceMatches tells if the specific resource matches the selector.
@@ -52,9 +54,9 @@ func ResourceMatches(resource *unstructured.Unstructured, rs policyv1alpha1.Reso
 
 // ResourceSelectorPriority tells the priority between the specific resource and the selector.
 func ResourceSelectorPriority(resource *unstructured.Unstructured, rs policyv1alpha1.ResourceSelector) ImplicitPriority {
-	if resource.GetAPIVersion() != rs.APIVersion ||
-		resource.GetKind() != rs.Kind ||
-		(len(rs.Namespace) > 0 && resource.GetNamespace() != rs.Namespace) {
+	if rs.APIVersion != resource.GetAPIVersion() ||
+		(rs.Kind != wildcardKind && rs.Kind != resource.GetKind()) ||
+		(len(rs.Namespace) > 0 && rs.Namespace != resource.GetNamespace()) {
 		return PriorityMisMatch
 	}
 
--- a/pkg/util/selector_test.go
+++ b/pkg/util/selector_test.go
@@ -294,6 +294,30 @@ func TestResourceSelectorPriority(t *testing.T) {
 			},
 			want: PriorityMisMatch,
 		},
+		{
+			name: "wildcard Kind, same apiVersion, matched",
+			args: args{
+				rs: policyv1alpha1.ResourceSelector{
+					APIVersion: resource.GetAPIVersion(),
+					Kind:       wildcardKind,
+					Name:       resource.GetName(),
+					Namespace:  resource.GetNamespace(),
+				},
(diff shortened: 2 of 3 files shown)


grosser (approved):

RainbowMango:
I guess you want to propagate gatekeeper resources by PropagationPolicy, I can understand the pain when managing so many resource kinds in ResourceSelector of PropagationPolicy, like:
```yaml
spec:
  resourceSelectors:
    - apiVersion: constraints.gatekeeper.sh/v1beta1
      kind: k1
    - apiVersion: constraints.gatekeeper.sh/v1beta1
      kind: k2
    - ...
    - apiVersion: constraints.gatekeeper.sh/v1beta1
      kind: k68
```
Is this your case? In addition, I wonder how many PropagationPolicy you are gona have?

It's a reasonable quest, but we also need to facing the concerns like discussed at #3823, and #3656.

a7i:
> Is this your case? In addition, I wonder how many PropagationPolicy you are gona have?

@RainbowMango that's exactly the case. At the moment, we plan on having a single ClusterPropagationPolicy for cluster-critical components (e.g. CNI, opa-gatekeeper, CSI drivers). Do you foresee any issues with this approach?

> It's a reasonable quest, but we also need to facing the concerns like discussed at #3823, and #3656.

Regarding #3823, using wildcard for `APIVersion` and `Kind` can pose risks, but I believe we can trust all kinds within an api group (at least in the case of opa-gatekeeper constraints). As far as the comment [here](https://github.com/karmada-io/karmada/blob/48e9a5b55163844b11b6ead2bec32076ba63449a/pkg/apis/policy/v1alpha1/propagation_types.go#L44-L50), this still poses a risk for core group but perhaps a risk accepted by the cluster/karmada admin. if there are any validations that you suggest, let me know and I'm happy to implement.

Regarding #3656, it seems that this will break preemption logic as it queries by API, Kind, Name, Namespace. We currently don't use preemption. Do you have any recommendations on a way around this?

RainbowMango:
> As far as the comment [here](https://github.com/karmada-io/karmada/blob/48e9a5b55163844b11b6ead2bec32076ba63449a/pkg/apis/policy/v1alpha1/propagation_types.go#L44-L50), this still poses a risk for core group but perhaps a risk accepted by the cluster/karmada admin. if there are any validations that you suggest, let me know and I'm happy to implement.

Yes, this might be an acceptable risk if the administrator knows the behavior. Indeed the core group poses a risk for people who don't know the behavior.  A validation dedicated to the core group seems a little bit weird.

What I'm thinking now is if the declaration always meets the user's expectations in the case of:
- CRD upgrading, might introduce new resource types within the current apigroup.
- Karmada-apiserver upgrading, same as CRD upgrading, new resource types might be introduced.
Do they care what new kind of resources will be propagated by the PropagationPolicy?
On the other hand, from an observability perspective, do they need to get a full list of propagated resource types from PropagationPolicy?

> Regarding https://github.com/karmada-io/karmada/pull/3656, it seems that this will break preemption [logic](https://github.com/karmada-io/karmada/blob/f04c17f72749a8db253da7b2790b65d8e912ce24/pkg/util/helper/binding.go#L246C25-L246C48) as it queries by API, Kind, Name, Namespace. We currently don't use preemption. Do you have any recommendations on a way around this?

Yes, this might be the challenge for now. Even though the preemption feature is still in the alpha stage, and we don't have so much feedback for it, breaking this functionality still concerning.

RainbowMango:
I'm still thinking about the idea mentioned on https://github.com/karmada-io/karmada/pull/3823#issuecomment-1658753667, how do you think this approach? Something like injector automation which can manage the resource selector automatically? Which could be implemented by means like a controller or mutating webhook?

a7i:
We started using discovery api to build the propagation policy and update frequently. A controller seemed overkill for us

PR #4682, opened 2024-03-05
fix: replacement of `::` to be dns RFC 1123 compliant

### LINKED ISSUE #4681: Failed to create ResourceBinding for resources with `::` in the name
**What happened**:

If a Role has two colons in the name `::`, then a ResourceBinding nor Work is created. Karmada replaced colon with dot (not sure why), but this causes issues as the name ends up having two dots in it which is not RFC 1123 compliant

```
Apply cluster policy(duplicate) failed: ResourceBinding.work.karmada.io "system..leader-locking-kube-controller-manager-role" is invalid: metadata.name: Invalid value: "system..leader-locking-kube-controller-manager-role": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
```

The name to be sanitized by karmada-controller-manager

**Environment**:
- Karmada version: v1.9.0
- Others:

a7i:
**What type of PR is this?**

/kind bug

Fixes #4681

This may cause issues if we have two roles:
`foo::bar`
and
`foo:bar`

```release-note
fix creation of ResourceBinding for resources with :: in the name
```

### DIFF
--- a/pkg/util/names/names.go
+++ b/pkg/util/names/names.go
@@ -70,6 +70,7 @@ func GenerateBindingName(kind, name string) string {
 	// For more information about the DNS subdomain name, please refer to
 	// https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names.
 	if strings.Contains(name, ":") {
+		name = strings.ReplaceAll(name, "::", ".") // `foo..bar` is not valid
 		name = strings.ReplaceAll(name, ":", ".")
 	}
 
--- a/pkg/util/names/names_test.go
+++ b/pkg/util/names/names_test.go
@@ -166,13 +166,13 @@ func TestGenerateBindingName(t *testing.T) {
 			testCase: "clusterRole kind resource",
 			kind:     "ClusterRole",
 			name:     "system::tom",
-			expect:   "system..tom-clusterrole",
+			expect:   "system.tom-clusterrole",
 		},
 		{
 			testCase: "roleBinding kind resource",
 			kind:     "RoleBinding",
 			name:     "system::tt:tom",
-			expect:   "system..tt.tom-rolebinding",
+			expect:   "system.tt.tom-rolebinding",
 		},
 		{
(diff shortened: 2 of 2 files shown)


chaunceyjiang:
> This may cause issues if we have two roles:

Can you provide me with a complete example? I want to test it locally. @a7i

pkg/util/names/names.go
```diff
@@ -70,6 +70,7 @@ func GenerateBindingName(kind, name string) string {
 	// For more information about the DNS subdomain name, please refer to
 	// https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names.
 	if strings.Contains(name, ":") {
+		name = strings.ReplaceAll(name, "::", ".") // `foo..bar` is not valid
```
chaunceyjiang, line 73: Can you add a unit test for this?

pkg/util/names/names_test.go
```diff
@@ -166,13 +166,13 @@ func TestGenerateBindingName(t *testing.T) {
...
 			kind:     "ClusterRole",
 			name:     "system::tom",
-			expect:   "system..tom-clusterrole",
+			expect:   "system.tom-clusterrole",
 		},
 		{
 			testCase: "roleBinding kind resource",
 			kind:     "RoleBinding",
 			name:     "system::tt:tom",
-			expect:   "system..tt.tom-rolebinding",
```
RainbowMango, line 175: Here is the UT. :) @chaunceyjiang

a7i:
Hi @chaunceyjiang , does the example mentioned in the linked issue help?
https://github.com/karmada-io/karmada/issues/4681#issuecomment-1980025646

PR #4702, opened 2024-03-12
Make cleanup in detector

XiShanYongYe-Chang:
**What type of PR is this?**

/kind cleanup

- Modify more readable method naming.
- Modify the typo in the comment.
- Adjustment of variable names.

Fixes #

```release-note
NONE
```


RainbowMango:
There is a panic, I'm not sure if this is related.
https://github.com/karmada-io/karmada/actions/runs/8249313355/job/22561748903?pr=4702#step:6:910

pkg/detector/detector.go
```diff
@@ -732,7 +726,8 @@ func (d *ResourceDetector) ClaimPolicyForObject(object *unstructured.Unstructure
 	objLabels := object.GetLabels()
 	if objLabels == nil {
 		objLabels = make(map[string]string)
-	} else if len(objLabels) > 0 {
+	}
+	if len(objLabels) > 0 {
```
RainbowMango, line 730: Why?

pkg/detector/detector.go
```diff
@@ -1272,15 +1267,15 @@ func (d *ResourceDetector) HandleClusterPropagationPolicyDeletion(policyName str
...
-	err := d.cleanPPUnmatchedResourceBindings(policy.Namespace, policy.Name, policy.Spec.ResourceSelectors)
+	err := d.cleanupUnmatchedRBsWithPP(policy.Namespace, policy.Name, policy.Spec.ResourceSelectors)
 	if err != nil {
 		return err
 	}
 
 	// When updating fields other than ResourceSelector, should first find the corresponding ResourceBinding
 	// and add the bound object to the processor's queue for reconciliation to make sure that
 	// PropagationPolicy's updates can be synchronized to ResourceBinding.
-	resourceBindings, err := d.listPPDerivedRB(policy.Namespace, policy.Name)
```
RainbowMango, line 1283: I more tend to the original name.

pkg/detector/detector.go
XiShanYongYe-Chang: Reduce one layer of cyclomatic complexity without changing the logic.

pkg/detector/detector.go
XiShanYongYe-Chang: The relationship between RB and PP can't be described as a derived relationship, right?

pkg/detector/detector.go
RainbowMango: Probably not, if-else is different from `if-if`, and the cyclomatic is not reduced.

pkg/detector/detector.go
RainbowMango: hard to say.

pkg/detector/detector.go
XiShanYongYe-Chang: You're right. The cyclomatic complexity of using `if` and `else if` is the same.

Issue #4109, opened 2023-10-09
The replicas of deployment is incorrectly when the related HPA is abnormal,


Rains6:
**What happened**:
The hpaReplicasSyncer controller is enabled. When the hpa delivered to the member cluster is abnormal, the desiredReplicas of the hpa is 0. In this case, the replicas synchronized to the control plane deployment are incorrect. Expected to use currentReplicas instead of desiredReplicas as the calculated value when hpa is abnormal

**What you expected to happen**:
Expected to use currentReplicas instead of desiredReplicas as the calculated value when hpa is abnormal.

**How to reproduce it (as minimally and precisely as possible)**:
1.The hpaReplicasSyncer controller is enabled. Delivering Deployment and HPA to member cluster A.
2.The HPA in cluster A is abnormal. In this case, the value of desiredReplicas of the HPA is 0.
![image](https://github.com/karmada-io/karmada/assets/49423798/8bdd575a-85ea-4c35-8b5d-4abfee3f41b2)
3.On the control plane, the replicas of the deployment is 0, which is expected to be 1.

**Anything else we need to know?**:

**Environment**:
- Karmada version: v1.7.0.alpha3
- kubectl-karmada or karmadactl version (the result of `kubectl-karmada version` or `karmadactl version`):
- Others:

chaunceyjiang:
The root cause of this issue is the instability of HPA.

The current implementation of #4072 heavily relies on HPA. If there is an exception with HPA, the number of replicas synchronized from the member cluster to the control plane becomes meaningless. And since HPA itself also heavily relies on the stability of metrics-server, HPA itself becomes even more unstable.

There are two failures that can occur with HPA:
1. Exception with metrics-server.
2. Accidental deletion of HPA.

Therefore, we are trying to introduce a new mechanism to avoid strong dependency on HPA:
Solution 1:
Directly query the scale sub-resource of workloads in the member cluster. This can accurately obtain the number of replicas for workloads. However, it cannot be used in Karmada's PULL mode.

Solution 2:
Aggregate status resources for workloads in the control plane, but for some custom resources, there may not be replica-related resources in their status.

chaunceyjiang:
@XiShanYongYe-Chang @jwcesign @lxtywypc @RainbowMango Do you have any other solutions?

XiShanYongYe-Chang:
Look for more people's ideas.
/cc @GitHubxsy

lxtywypc:
[...]

PR #4707, opened 2024-03-13
sync deployment replicas when it is controlled by hpa

chaosi-zju:
**What type of PR is this?**

/kind feature

As for issue #4058, the previous implemention can not cover all cases, such as:

* member cluster hpa abnormal, we would sync a wrong replicas to resource template, just as described in #4109
* when user manually update `spec.replicas` in template, the modifition would be retained, and thus causing ready replias not equal to expect replicas when getting deployment (e.g: the user get deployment and see replicas ready/expect=2/2; then user manually update `spec.replicas` to 3; then user get deployment agagin and see replicas ready/expect=2/3; this displayment is not suitable for user to understand)
* In `Duplicated` schedule type, the `spec.replicas` in resource template has different meaning, it is not suitable to do this kind of replicas sync.
* ...

So, I introduced a `deploymentReplicasSyncer` to replace previous `hpaReplicasSyncer`:
* we intend to support replicas syncer only for deployment resources (`hpa` is not the way we recommend, while `FederatedHPA` is what we recommend. However, a small number of existing users have already used deployment in this way, and we have to support it; while other resources have not been used by users, we hope that they will directly use the recommended way.)
* replicas syncer indirectly by watching `HPA` resource takes a detour and hides a lot of unpredictable results. It is more suitable to directly watching `Deployment` resource. That is, replace `spec.replicas` with `status.replicas` if the two is not equal. 

As for **expected system behavior**, it can be **concluded as follows**:
* If member cluster has deployment resource, and it is now controlled by hpa, change to template replicas will not sync to this member cluster
* If member cluster doesn't even have deployment resource, not to mention being controlled by the hpa, new added replicas can ben synced from template change.

Fixes #

much appreciate everyone involved in this issue to participate in the review @XiShanYongYe-Chang @lxtywypc @chaunceyjiang 

feel free to provide better resolution, however, **your resolution should generally meet follow cases**:

**case 1)** template (ready/expect=2/2), propagate to two cluster -> hpa scale all member cluster replicas to 4 -> template (ready/expect=4/4), even in the scenario of hpa abnormal、member cluster power down...

**case 2)** template (ready/expect=2/2), propagate to two cluster -> user update `spec.replicas` in template from 2 to 3 -> no matter what intermediate states is but should ensure that the final state is 2/2, and cannot be 2/3.

**case 3)** template (ready/expect=1/1), propagate to two cluster -> user update `spec.replicas` in template from 1 to 2 -> final state should be 2/2, that means one member cluster previously doesn't have deployment, now it has one replica.

```release-note
Introduced `deploymentReplicasSyncer` controller which syncs Deployment's replicas from the member cluster to the control plane, while previous `hpaReplicasSyncer` been replaced.
```


pkg/controllers/hpaautolabelretain/hpa_auto_label_retain_predicate.go
```diff
@@ -44,7 +44,7 @@ func (r *HPAReplicasSyncer) Create(e event.CreateEvent) bool {
 }
 
 // Update implements UpdateEvent filter
-func (r *HPAReplicasSyncer) Update(e event.UpdateEvent) bool {
+func (r *HPAAutoLabelRetain) Update(e event.UpdateEvent) bool {
```
XiShanYongYe-Chang, line 47: How to directly filter out ScaleTargetRef as non-Deployment?

pkg/controllers/hpaautolabelretain/hpa_auto_label_retain_controller.go
```diff
@@ -0,0 +1,66 @@
...
+const (
+	// ControllerName is the controller name that will be used when reporting events.
+	ControllerName = "hpa-auto-label-retain"
+	// scaleRefWorkerNum is the async Worker number
+	scaleRefWorkerNum = 1
+)
+
+// HPAAutoLabelRetain is to automatically add `retain-replicas` label to resource template mananged by HPA.
+type HPAAutoLabelRetain struct {
+	DynamicClient dynamic.Interface
```
XiShanYongYe-Chang, line 40: We can use `clientset.Interface` instead.

pkg/controllers/deploymentreplicassyncer/deployment_replicas_syncer_controller.go
```diff
@@ -0,0 +1,165 @@
...
+			len(binding.Status.AggregatedStatus), len(binding.Spec.Clusters))
+		return controllerruntime.Result{}, fmt.Errorf("retry to wait status in binding collected")
+	}
+	for _, status := range binding.Status.AggregatedStatus {
+		if status.Status == nil {
+			klog.Errorf("wait until aggregated status of cluster %s collected", status.ClusterName)
+			return controllerruntime.Result{}, fmt.Errorf("retry to wait status in binding collected")
+		}
+	}
+	klog.Infof("all %d clusters status collected success", len(binding.Status.AggregatedStatus))
```
XiShanYongYe-Chang, line 151: The information described in this log may be incorrect. It is valid only when some new clusters are propagated.

pkg/controllers/deploymentreplicassyncer/deployment_replicas_syncer_controller.go
```diff
@@ -0,0 +1,165 @@
...
+
+	// update replicas
+	oldReplicas := *deployment.Spec.Replicas
+	deployment.Spec.Replicas = &deployment.Status.Replicas
+	if err := r.Client.Update(ctx, deployment); err != nil {
+		klog.Errorf("failed to update deployment (%s/%s) replicas: %+v", deployment.Namespace, deployment.Name, err)
+		return controllerruntime.Result{}, err
+	}
+
+	klog.Infof("succeffully udpate deployment (%s/%s) replicas from %d to %d", deployment.Namespace,
```
XiShanYongYe-Chang, line 161: typo: `successfully update`

pkg/controllers/hpaautolabelretain/hpa_auto_label_retain_controller.go
```diff
@@ -0,0 +1,66 @@
...
+	"k8s.io/client-go/dynamic"
+	controllerruntime "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+
+	"github.com/karmada-io/karmada/pkg/util"
+)
+
+const (
+	// ControllerName is the controller name that will be used when reporting events.
+	ControllerName = "hpa-auto-label-retain"
```
RainbowMango, line 33: Do we have a better name?
I'm still thinking of it.

pkg/controllers/hpaautolabelretain/hpa_auto_label_retain_predicate.go
chaosi-zju: Due to we have document for users [Use the Retain Interpreter to Resolve Control Conflict between the Control Plane and Member Clusters](https://karmada.io/docs/userguide/globalview/customizing-resource-interpreter/#use-the-retain-interpreter-to-resolve-control-conflict-between-the-control-plane-and-member-clusters) to teach how to configure retain, may be we shall keep this logic.

pkg/controllers/hpaautolabelretain/hpa_auto_label_retain_controller.go
chaosi-zju: ditto

cmd/controller-manager/app/controllermanager.go
```diff
@@ -206,7 +206,7 @@ func Run(ctx context.Context, opts *options.Options) error {
 var controllers = make(controllerscontext.Initializers)
 
 // controllersDisabledByDefault is the set of controllers which is disabled by default
-var controllersDisabledByDefault = sets.New("hpaReplicasSyncer")
+var controllersDisabledByDefault = sets.New("hpaAutoLabelRetain", "deploymentReplicasSyncer")
```
RainbowMango, line 209:
```suggestion
var controllersDisabledByDefault = sets.New("hpaScaleTargetMarker", "deploymentReplicasSyncer")
```
How about taking the `hpaScaleTargetMarker` name?
In my opinion, the controller is responsible for marking objects that scaled by a HPA. 
The controller uses a `label` to mark these objects, how this label will be applied, it's irrelevant to this controller.

@XiShanYongYe-Chang What do you think?

chaosi-zju: 👍

RainbowMango:
This feature needs a release notes.

/assign @XiShanYongYe-Chang

XiShanYongYe-Chang (commented):
/lgtm

PR #4721, opened 2024-03-18
fix deployment replicas syncer in case deployment status changed before label added

chaosi-zju:
**What type of PR is this?**

/kind bug

fix deployment replicas syncer in case deployment status changed before label added

how to produce:
  1. propagate deployment with 1 replicas to two clusters by static weight 1:1 strategy policy.
  2. propagate hpa which `minReplicas=2` to two clusters.
  
then, we may observed the following results in order:
  1. hpa scale deployment in member1 cluster from 1 to 2.
  2. deployment status change collected, its status.replicas change from 1 to 2.
  3. `hpaScaleTargetMarker` add label `retain-replicas` to deployment.
  4. get deployment and see `ready/expect=2/1`

in this case, `deploymentReplicasSyncer` didn't do anything because after `retain-replicas` labeled, the deployment's `spec` and `status` had no changes.

Fixes #

bugfix for #4707

```release-note

```

### DIFF
--- a/pkg/controllers/deploymentreplicassyncer/deployment_replicas_syncer_controller.go
+++ b/pkg/controllers/deploymentreplicassyncer/deployment_replicas_syncer_controller.go
@@ -57,6 +57,13 @@ var predicateFunc = predicate.Funcs{
 			return false
 		}
 
+		// if old deployment is not labeled `retain-replicas`, but new is labeled, reconcile is needed.
+		// in case of deployment status changed before `retain-replicas` labeled.
+		oldRetainReplicasLabel := util.GetLabelValue(oldObj.GetLabels(), util.RetainReplicasLabel)
+		if oldRetainReplicasLabel != util.RetainReplicasValue {
+			return true
+		}
+
 		if oldObj.Spec.Replicas == nil || newObj.Spec.Replicas == nil {
 			klog.Errorf("spec.replicas field unexpectedly become nil: %+v, %+v", oldObj.Spec.Replicas, newObj.Spec.Replicas)
 			return false


pkg/controllers/deploymentreplicassyncer/deployment_replicas_syncer_controller.go
```diff
@@ -57,6 +57,13 @@ var predicateFunc = predicate.Funcs{
 			return false
 		}
 
+		// if old deployment is not labeled `retain-replicas`, but new is labeled, reconcile is needed.
+		// in case of deployment status changed before `retain-replicas` labeled.
+		oldRetainReplicasLabel := util.GetLabelValue(oldObj.GetLabels(), util.RetainReplicasLabel)
+		if oldRetainReplicasLabel != util.RetainReplicasValue {
+			return true
+		}
+
```
chaunceyjiang, line 66: It seems that as long as `util.RetainReplicasLabel` changes, then it should be re-`reconcile`?

chaosi-zju: yes, then get latest deployment, only if deployment's `spec.replicas` not equal to `status.replicas` would do the update.

chaosi-zju: > It seems that as long as `util.RetainReplicasLabel` changes, then it should be re-`reconcile`?

To be precise, there was no such label before, but now there is, it will be reconciled.

chaunceyjiang: Got it.

chaunceyjiang:
Please fill in the release notes in the PR description.

chaosi-zju:
> Please fill in the release notes in the PR description.

Hi, actually, this is a bugfix for PR #4707, which merged last week, and the previous PR already has release note.

In this case, do we need to add extra release note?

RainbowMango (approved):
/approve

PR #4729, opened 2024-03-20
fix deployment replicas syncer in case that `status.replicas` haven't been collected from member cluster to template

### LINKED ISSUE #4758: deployment propagated failed after deployment replicas syncer feature enabled
**What happened**:

ci e2e test failed: https://github.com/karmada-io/karmada/actions/runs/8336573479/job/22814077291

just propagate 2 replicas into two clusters with static weight 1:1, it is expected to see `2/2` when getting deployment.

however, after #4707 merged, it actually may got `1/1` result with a high probability.

it is expected to see `2/2` when getting deployment finally.

just propagate 2 replicas into two clusters with static weight 1:1.

**Environment**:
- Karmada version:
- Others:

chaosi-zju:
**What type of PR is this?**

/kind bug

This PR mainly handles **two problem**:

**1、Problem 1**

we should make sure the latest collected status.replicas is synced from binding to template.

e.g: user set deployment.spec.replicas = 2, then sum replicas in binding.status becomes 2, however, if now deployment.status.replicas is still 1, we shouldn't update spec.replicas to 1 until status.replicas becomes 2.

**2、Problem 2**

as described in #4758, fix deployment replicas syncer in case that the `status.replicas` in `aggregatedStatus` of binding is 0.

actually, if member cluster deployment is controlled by hpa, its status.replicas must > 0 (since hpa.minReplicas > 0),

so, if its `status.replicas` is 0, means the status haven't been collected from member cluster, and needs waiting before do the following update.

Fixes #4758

bugfix for #4707

```release-note
fix deployment replicas syncer in case that `status.replicas` haven't been collected from member cluster to template.
```

### DIFF
--- a/pkg/controllers/deploymentreplicassyncer/deployment_replicas_syncer_controller.go
+++ b/pkg/controllers/deploymentreplicassyncer/deployment_replicas_syncer_controller.go
@@ -18,7 +18,8 @@ package deploymentreplicassyncer
 
 import (
 	"context"
-	"fmt"
+	"encoding/json"
+	"time"
 
 	appsv1 "k8s.io/api/apps/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -38,6 +39,8 @@ import (
 const (
 	// ControllerName is the controller name that will be used when reporting events.
 	ControllerName = "deployment-replicas-syncer"
+
+	waitDeploymentStatusInterval = 1 * time.Second
 )
 
 // DeploymentReplicasSyncer is to sync deployment replicas from status field to spec field.
@@ -98,6 +101,7 @@ func (r *DeploymentReplicasSyncer) Reconcile(ctx context.Context, req controller
 	binding := &workv1alpha2.ResourceBinding{}
 	bindingName := names.GenerateBindingName(util.DeploymentKind, req.Name)
 
+	// 1. get latest binding
 	if err := r.Client.Get(ctx, client.ObjectKey{Namespace: req.Namespace, Name: bindingName}, binding); err != nil {
 		if apierrors.IsNotFound(err) {
 			klog.Infof("no need to update deployment replicas for binding not found")
@@ -106,11 +110,12 @@ func (r *DeploymentReplicasSyncer) Reconcile(ctx context.Context, req controller
 		return controllerruntime.Result{}, err
 	}
 
-	// if it is not divided schedule type, no need to update replicas
+	// 2. if it is not divided schedule type, no need to update replicas
 	if binding.Spec.Placement.ReplicaSchedulingType() != policyv1alpha1.ReplicaSchedulingTypeDivided {
 		return controllerruntime.Result{}, nil
 	}
 
+	// 3. get latest deployment
 	if err := r.Client.Get(ctx, req.NamespacedName, deployment); err != nil {
 		if apierrors.IsNotFound(err) {
 			klog.Infof("no need to update deployment replicas for deployment not found")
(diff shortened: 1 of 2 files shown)


chaosi-zju:
@XiShanYongYe-Chang @chaunceyjiang can you review on this ?

pkg/controllers/deploymentreplicassyncer/deployment_replicas_syncer_controller.go
```diff
@@ -98,69 +106,112 @@ func (r *DeploymentReplicasSyncer) Reconcile(ctx context.Context, req controller
...
-	// update replicas
-	oldReplicas := *deployment.Spec.Replicas
-	deployment.Spec.Replicas = &deployment.Status.Replicas
-	if err := r.Client.Update(ctx, deployment); err != nil {
-		klog.Errorf("failed to update deployment (%s/%s) replicas: %+v", deployment.Namespace, deployment.Name, err)
-		return controllerruntime.Result{}, err
+	if deployment.Status.Replicas != bindingStatusReplicas {
+		klog.V(4).Infof("wait until deployment status replicas (%d) equal to binding status replicas (%d)",
+			deployment.Status.Replicas, bindingStatusReplicas)
+		return false
```
chaosi-zju, line 213: Notes: Actually, only these lines has been changed, I added two wait condition.

Other line's change is just because I wrapped a inner retry function  `wait.PollUntilContextTimeout`, for the previous retry interval of `Reconcile` would getting larger and larger, it better to keep same interval in this case.

XiShanYongYe-Chang (commented):
/lgtm

pkg/controllers/deploymentreplicassyncer/deployment_replicas_syncer_controller.go
```diff
@@ -98,69 +106,112 @@ func (r *DeploymentReplicasSyncer) Reconcile(ctx context.Context, req controller
 	binding := &workv1alpha2.ResourceBinding{}
 	bindingName := names.GenerateBindingName(util.DeploymentKind, req.Name)
 
-	if err := r.Client.Get(ctx, client.ObjectKey{Namespace: req.Namespace, Name: bindingName}, binding); err != nil {
-		if apierrors.IsNotFound(err) {
-			klog.Infof("no need to update deployment replicas for binding not found")
+	err := wait.PollUntilContextTimeout(ctx, waitDeploymentStatusInterval, waitDeploymentStatusTimeout, true,
```
chaunceyjiang, line 109: I think using `PollUntilContextTimeout` in the `controller` is not a very good solution. The controller should `watch` for these `changes `to do the corresponding `Reconcile`, instead of constantly `querying` for these changes.

RainbowMango: +1,

The controller might need to wait for an event that it doesn't watch, in that case, it can re-queue the resource key.

chaosi-zju:
the max backoff time of workqueue is 1000s @RainbowMango 

https://github.com/karmada-io/karmada/blob/13dafc62a2c60d876ae76c139005646a3a173c1f/pkg/sharedcli/ratelimiterflag/ratelimiterflag.go#L42-L48

https://github.com/karmada-io/karmada/blob/13dafc62a2c60d876ae76c139005646a3a173c1f/vendor/k8s.io/client-go/util/workqueue/default_rate_limiters.go#L37-L45

chaunceyjiang:
> the max backoff time of workqueue is 1000s

The resource key should be able to re-queue infinitely many times, it's just that the maximum allowed backoff time is 1000s.

pkg/resourceinterpreter/default/native/reflectstatus.go
```diff
@@ -74,7 +75,20 @@ func reflectDeploymentStatus(object *unstructured.Unstructured) (*runtime.RawExt
...
+
+	grabStatusRaw, err := helper.BuildStatusRawExtension(grabStatus)
+	if err != nil {
+		return nil, err
+	}
+
+	// if status is empty struct, it actually means status haven't been collected from member cluster.
+	if bytes.Equal(grabStatusRaw.Raw, []byte("{}")) {
+		klog.Errorf("Empty status grabed rom %s(%s/%s) which should been non-empty field.",
+			object.GetKind(), object.GetNamespace(), object.GetName())
```
RainbowMango, line 87: Is it worth logging an `error`?
In addition, there are 3 typos there, `grabed`, `rom`, `been`.

chaosi-zju: beacuse above not exist log is `errorf`, it should keep the same, so should we update the two into `info`?

https://github.com/karmada-io/karmada/blob/db15f17b8ecfa0e606579a175fc9371b45c0fd9b/pkg/resourceinterpreter/default/native/reflectstatus.go#L59-L63

chaosi-zju: sorry for typo error, it will be fix as:

```suggestion
		klog.Errorf("Empty status grabbed from %s(%s/%s) which should be non-empty field.",
			object.GetKind(), object.GetNamespace(), object.GetName())
```

RainbowMango (approved):
/approve

RainbowMango (commented):
/lgtm

PR #4737, opened 2024-03-21
Fix eps remain when karmada-controller-manager restart

XiShanYongYe-Chang:
**What type of PR is this?**

/kind bug

After the karmada control plane resource is deleted and the propagated resource in the member cluster is also deleted, but the deletion event of the `EndpointSlice` resource has not been processed by the controller in the control plane, the `karmada-controller-manager` is restarted due to an OOM reason. `EndpointSlice` resources are not deleted from the Karmada control plane.

Fixes #

```release-note
`karmada-controller-manager`: Fixed the issue of EndpointSlice residual in case of the karmada-controller-manager restart.
```

### DIFF
--- a/pkg/controllers/mcs/endpointslice_controller.go
+++ b/pkg/controllers/mcs/endpointslice_controller.go
@@ -29,6 +29,7 @@ import (
 	controllerruntime "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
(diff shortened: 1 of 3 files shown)


pkg/controllers/mcs/endpointslice_controller.go
```diff
@@ -54,21 +55,20 @@ func (c *EndpointSliceController) Reconcile(ctx context.Context, req controllerr
 	work := &workv1alpha1.Work{}
 	if err := c.Client.Get(ctx, req.NamespacedName, work); err != nil {
 		if apierrors.IsNotFound(err) {
-			// Cleanup derived EndpointSlices after work has been removed.
-			err = helper.DeleteEndpointSlice(c.Client, labels.Set{
```
chaunceyjiang, line 58: I think this code can be kept without deleting. 

When upgrading from a lower version to this version, this code can clean up any leftover EndpointSlices.

XiShanYongYe-Chang: Thanks, it makes sense.

XiShanYongYe-Chang: Updated, please help cc again.

RainbowMango (commented):
/assign

pkg/controllers/mcs/service_export_controller.go
```diff
@@ -139,6 +143,52 @@ func (c *ServiceExportController) RunWorkQueue() {
 	}
 	c.worker = util.NewAsyncWorker(workerOptions)
 	c.worker.Run(c.WorkerNumber, c.StopChan)
+
+	go c.enqueueReportedEpsServiceExport()
+}
+
+func (c *ServiceExportController) enqueueReportedEpsServiceExport() {
+	workList := &workv1alpha1.WorkList{}
+	err := wait.PollUntilContextCancel(context.TODO(), time.Second, true, func(ctx context.Context) (done bool, err error) {
```
RainbowMango, line 152: There is [an issue](https://github.com/kubernetes/kubernetes/pull/119762) with the `wait.PollUntilContextCancel`, that makes the condition will be executed twice immediately.

This issue has been fixed in Kubernetes v1.29 but we haven't adopted it. We can continue to use the deprecated functions for now. This deprecation work is tracked by #3835.

XiShanYongYe-Chang: Ok, thanks, let me update it to wait.Poll first.

XiShanYongYe-Chang: Update it to the `wait.PollUntil`, please help take a review again.

RainbowMango:
@chaunceyjiang Do you have any further comments?

chaunceyjiang (commented):
/lgtm

PR #4751, opened 2024-03-26
fix wrong annotation marks on resource

XiShanYongYe-Chang:
**What type of PR is this?**

/kind bug

Using the `MergeAnnotation` function will filter out existing keys. When the key value that needs to be marked changes, such as policy preemption, resource recovery, we need to use the `ReplaceAnnotation` function to replace the value of the key in the annotation.

Fixes #

```release-note
`karmada-controller-manager`: Fixed incorrect annotation markup when policy preemption occurs.
```


whitewindmills (commented):
Good finding!
/lgtm

RainbowMango (approved):
/approve

PR #4646, opened 2024-02-21
Add service-name-resolution-detector

wm775825:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

Add cluster problem detector to detect key component failure and sync condition to cluster object,  co-working with remedy APIs #4635 to ClusterState-based policy orchestration solutions.

**Which issue(s) this PR fixes**:

Like [node-problem-detector](https://github.com/kubernetes/node-problem-detector) , cluster-problem-detector is deployed as DaemonSet on member clusters, aiming to detect failure periodically and export it to node condition. 

The leader pod of the daemonset will evaluate the specific condition on all nodes and report condition to cluster object.

The remedy contollers will watch these conditions of clusters and take action based on cluster state and remedy policies.

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
Add cluster problem detector to detect key component failure, such as domain name resolution, and sync condition to cluster object.
```

PR #4763, opened 2024-03-28
bugfix: generate event if reporting condition to karmada failed

wm775825:
**What type of PR is this?**

/kind bug

service-name-resolution-detector-example, which is introduced in #4646 , requires a kubeconfig to control plane to report conditions to karmada. If kubeconfig is invalid or expired, reporting will fail and events are needed to inform user.

Fixes #

```release-note

```

### DIFF
--- a/pkg/servicenameresolutiondetector/coredns/detector.go
+++ b/pkg/servicenameresolutiondetector/coredns/detector.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"sync"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


pkg/servicenameresolutiondetector/coredns/detector.go
```diff
@@ -156,13 +167,13 @@ func (d *Detector) Run(ctx context.Context) {
 			observed := lookupOnce(logger)
 			curr, err := d.conditionStore.Load(d.nodeName)
 			if err != nil {
-				d.eventRecorder.Eventf(localReference, corev1.EventTypeWarning, "LoadCorednsConditionFailed", "failed to load condition: %v", err)
+				d.reportErrorEvent(corev1.EventTypeWarning, loadCorednsConditionFailed, "failed to load condition: %v", err)
```
XiShanYongYe-Chang, line 170: Does it also need to wait an hour to report?

pkg/servicenameresolutiondetector/coredns/detector.go
```diff
@@ -199,13 +210,34 @@ func (d *Detector) processNextWorkItem(ctx context.Context) bool {
...
 	return true
 }
 
+func (d *Detector) reportErrorEvent(eventType string, reason string, messageFmt string, args ...interface{}) {
+	key := fmt.Sprintf("%s-%s-"+messageFmt, eventType, reason, args)
+	now := time.Now()
+
+	d.errorEventLock.Lock()
+	lastReportTime, ok := d.errorEventMap[key]
+	if ok && !lastReportTime.IsZero() && now.Sub(lastReportTime) < time.Hour {
```
XiShanYongYe-Chang, line 231: Whether the time is necessary as a configurable parameter of the component?

PR #4774, opened 2024-03-28
fix bug of wrong delete endpointslice collect from member cluster

XiShanYongYe-Chang:
**What type of PR is this?**

/kind bug

The last few CI runs have been failing frequently, for example: https://github.com/karmada-io/karmada/actions/runs/8463439534/job/23186473124

Through analysis, it is found that a bug was introduced in #4737, which will cause when creating another new ServiceEport, it will mistakenly delete the endpointslice that collected by other ServiceExports.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/mcs/service_export_controller.go
+++ b/pkg/controllers/mcs/service_export_controller.go
@@ -401,7 +401,7 @@ func (c *ServiceExportController) reportEndpointSliceWithServiceExportCreate(ser
 		return err
 	}
 
-	err = c.removeOrphanWork(endpointSliceObjects, serviceExportKey.Cluster)
+	err = c.removeOrphanWork(endpointSliceObjects, serviceExportKey)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


pkg/controllers/mcs/service_export_controller.go
```diff
@@ -424,12 +424,15 @@ func (c *ServiceExportController) removeOrphanWork(endpointSliceObjects []runtim
 
 	collectedEpsWorkList := &workv1alpha1.WorkList{}
 	if err := c.List(context.TODO(), collectedEpsWorkList, &client.ListOptions{
-		Namespace: names.GenerateExecutionSpaceName(targetCluster),
+		Namespace: names.GenerateExecutionSpaceName(serviceExportKey.Cluster),
 		LabelSelector: labels.SelectorFromSet(labels.Set{
 			util.PropagationInstruction: util.PropagationInstructionSuppressed,
+			util.ServiceNamespaceLabel:  serviceExportKey.Namespace,
+			util.ServiceNameLabel:       serviceExportKey.Name,
```
RainbowMango, line 431: This is the key point that fetched Works do not belong to the current service, and these Works will be accidentally deleted.

PR #4776, opened 2024-03-28
remove old MergeAnnotation func and rename ReplaceAnnotation to MergeAnnotation

XiShanYongYe-Chang:
**What type of PR is this?**

/kind cleanup

Learn from #4751, to avoid ambiguity and to be consistent with the Annotation naming convention:

https://github.com/karmada-io/karmada/blob/57c1989667d826864aa09105daa04434e2c75dd9/pkg/util/annotation.go#L29-L51

remove old MergeAnnotation func and rename ReplaceAnnotation to MergeAnnotation.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/util/annotation_test.go
+++ b/pkg/util/annotation_test.go
@@ -25,118 +25,6 @@ import (
 	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
 )
 
-func TestMergeAnnotation(t *testing.T) {
-	tests := []struct {
-		name            string
-		obj             *unstructured.Unstructured
-		annotationKey   string
-		annotationValue string
-		expected        *unstructured.Unstructured
-	}{
-		{
-			name: "nil annotations",
-			obj: &unstructured.Unstructured{
-				Object: map[string]interface{}{
-					"apiVersion": "apps/v1",
-					"kind":       "Deployment",
-					"metadata": map[string]interface{}{
-						"name": "demo-deployment",
-					},
-					"spec": map[string]interface{}{
-						"replicas": 2,
-					},
-				},
-			},
-			expected: &unstructured.Unstructured{
-				Object: map[string]interface{}{
-					"apiVersion": "apps/v1",
-					"kind":       "Deployment",
-					"metadata": map[string]interface{}{
-						"name":        "demo-deployment",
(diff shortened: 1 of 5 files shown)


whitewindmills (commented):
/lgtm

pkg/controllers/binding/common.go
```diff
@@ -154,18 +154,18 @@ func mergeTargetClusters(targetClusters []workv1alpha2.TargetCluster, requiredBy
 
 func mergeLabel(workload *unstructured.Unstructured, binding metav1.Object, scope apiextensionsv1.ResourceScope) map[string]string {
 	var workLabel = make(map[string]string)
-	util.MergeLabel(workload, util.ManagedByKarmadaLabel, util.ManagedByKarmadaLabelValue)
+	util.ReplaceLabel(workload, util.ManagedByKarmadaLabel, util.ManagedByKarmadaLabelValue)
```
RainbowMango, line 157: fells not right, as the workload might not have the label, how does `replace`?

XiShanYongYe-Chang: Any suggestion?

XiShanYongYe-Chang: I'm thinking that MergeLabel is the  appropriate name. The original MergeAnnotation function may not be necessary:

https://github.com/karmada-io/karmada/blob/4e2548250b19a8f8148c004defa02b0e99f5a3e7/pkg/util/annotation.go#L29-L40

pkg/util/annotation_test.go
```diff
@@ -25,118 +25,6 @@ import (
 	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
 )
 
-func TestMergeAnnotation(t *testing.T) {
```
RainbowMango, line 28: Why remove this test?

XiShanYongYe-Chang: When ReplaceAnnotation is replaced with MergeAnnotation, the test method name is not changed. Just update.

XiShanYongYe-Chang:
Can you help take a look again @whitewindmills @RainbowMango ?

RainbowMango (approved):
/lgtm
/approve

PR #4778, opened 2024-03-29
Pin plural name for all APIs to avoid unexpected change

RainbowMango:
**What type of PR is this?**

/kind cleanup

When the controller-tools generate the CRDs, it [guesses the plural name](https://github.com/kubernetes-sigs/controller-tools/blob/80fc62f695b13677135377f47b61c61908661f10/pkg/crd/spec.go#L66) by [github.com/gobuffalo/flect](https://github.com/gobuffalo/flect) if the plural name is not explicitly specified.

But the rule of generating plural names might change over different [github.com/gobuffalo/flect](https://github.com/gobuffalo/flect) versions. See https://github.com/kubernetes-sigs/controller-tools/pull/804#discussion_r1178098768 for an example.

So, this PR pin plural name for all APIs to avoid this unexpected change.
(We don't depend on [github.com/gobuffalo/flect](https://github.com/gobuffalo/flect), but the controller-tools reply on it. We might update controller-tools.)
Fixes #

I'm trying to update the [controller-tools](https://github.com/karmada-io/karmada/blob/70afc1ad6d170bd4773b5e7a9735a4116816f814/hack/update-crdgen.sh#L22) but found the plural name of Karmada operator changed unexpectedly(see the fix at #4777), that reminds me we'd better pin the plural name.

```release-note
NONE
```

### DIFF
--- a/charts/karmada/_crds/bases/config/config.karmada.io_resourceinterpretercustomizations.yaml
+++ b/charts/karmada/_crds/bases/config/config.karmada.io_resourceinterpretercustomizations.yaml
@@ -9,6 +9,8 @@ metadata:
 spec:
   group: config.karmada.io
   names:
+    categories:
+    - karmada-io
     kind: ResourceInterpreterCustomization
     listKind: ResourceInterpreterCustomizationList
     plural: resourceinterpretercustomizations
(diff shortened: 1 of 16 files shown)


XiShanYongYe-Chang (commented):
Thanks a lot~
/lgtm

Issue #4709, opened 2024-03-13
Does  karmada can  prevent removal of these managed resources


CharlesQQ:
**What would you like to be added**:
Under normal circumstances,  delete resource  in  karmada  control plane， resource in mamber cluster will be deleted;
Is there any way to prevent removal of these managed resources?

**Why is this needed**:
Like  kubefed Annotation `kubefed.io/orphan: true`; this feature is useful to  rollback  when  resource  migrate to karmada; 
refer to (https://github.com/kubernetes-retired/kubefed/blob/master/docs/userguide.md)

RainbowMango:
That sounds like a reasonable feature. I'd like to know when and how this feature would be used.

The Kubefed annotation `kubefed.io/orphan: true` is added to the `federated resources`, corresponding to Karmada `PropagationPolicy`. I tend to extend the API instead of introducing an annotation.

XiShanYongYe-Chang:
I propose a simple idea. Add a new API, call it `RemovePolicy`, and define the target resource and the corresponding deletion policy in its resources, such as whether to keep the resources in the member cluster, or delete the resources in the member cluster synchronously, how much the deletion delay is, etc.

The idea is very rough, throw bricks lead jade.

CharlesQQ:
|  That sounds like a reasonable feature. I'd like to know when and how this feature would be used.

Used to migrate some services from karmda. If the services migrated to karmada do not meet expectations, provide a rollback capability.

CharlesQQ:
> I propose a simple idea. Add a new API, call it RemovePolicy, and define the target resource and the corresponding deletion policy in its resources, such as whether to keep the resources in the member cluster, or delete the resources in the member cluster synchronously, how much the deletion delay is, etc.

I agreed @RainbowMango idea, Add a new API might be too complicated

XiShanYongYe-Chang:
Hi everyone, let's use umbrella issue #5577 to track this requirement.
/close

Issue #3004, opened 2022-12-29
How to prevent deletion of resources for member clusters?


chaunceyjiang:
**Please provide an in-depth description of the question you have**:

I used the following command by mistake. Now I want to delete coredns on the control plane, but I don’t want to delete the coredns of the member cluster. How should I do it?

```
karmadactl promote deployment coredns -n kube-system -C cluster1
```

**What do you think about this question?**:

**Environment**:
- Karmada version: v.1.4.0
- Kubernetes version: 
- Others:

RainbowMango:
We have a draft proposal for this:https://github.com/karmada-io/karmada/tree/master/docs/proposals/cleanup-propagated-resources

The proposal is not to solve the `mistake` but to provide a mechanism to control if the resource should be removed from member clusters.

chaunceyjiang:
> provide a mechanism to control if the resource should be removed from member clusters.

> This document proposes a mechanism to specify that a member cluster resource should be removed from a managed cluster when **leaving**.

I don't want to remove the member cluster, I just want to delete the `coredns` of the control plane.

chaunceyjiang:
https://github.com/karmada-io/karmada/blob/d1617303a41c9fb472e29550acd16dc4cbd1320d/pkg/util/constants.go#L12-L18

I think we can extend this instruction.

sunbinnnnn:
I got the same case, my workaround is stopping the controller, and force delete RB, workload and WORK, then start the controller...

chaunceyjiang:
>  and almost cause terrible disaster. 

Yes

> my workaround is stopping the controller, and force delete RB, workload and WORK, then start the controller...

This operation has caused me a lot of trouble.

RainbowMango:
I'm trying to figure out which PR/Issue should be included in the coming v1.7 release which is planned at the end of this month.
I guess we don't have enough time for this feature, so I'm moving this to v1.8.

XiShanYongYe-Chang:
Hi @chaunceyjiang Do you need to keep work at the same time in your use case?

Ref https://github.com/karmada-io/karmada/pull/4788#discussion_r1575971218

RainbowMango:
This issue could be resolved by https://github.com/karmada-io/karmada/issues/5577#issuecomment-2372856091.

/close

PR #4788, opened 2024-04-01
Add annotation to the  the resource template to prevent removal managed resources

CharlesQQ:
**What type of PR is this?**

Extend the API field orphaningDeletion to prevent removal managed resources

Fixes https://github.com/karmada-io/karmada/issues/4709

```release-note

```

### DIFF
--- a/pkg/apis/work/v1alpha1/work_types.go
+++ b/pkg/apis/work/v1alpha1/work_types.go
@@ -30,6 +30,13 @@ const (
 	ResourcePluralWork = "works"
 	// ResourceNamespaceScopedWork indicates if Work is NamespaceScoped.
 	ResourceNamespaceScopedWork = true
+
+	// DeletionPolicyAnnotation is added to resource template annotation to indicates how removal action of managed resources.
+	DeletionPolicyAnnotation = "deletionpolicy.karmada.io"
+
+	// DeletionCompleteAnnotation means that the resource in member clusters completely independent of karmada
+	// Not only cleanup the labels/annotations from the resource on member clusters, but also delete work object in control plane
+	DeletionCompleteAnnotation = "complete"
 )
 
 // +genclient
--- a/pkg/controllers/execution/execution_controller.go
+++ b/pkg/controllers/execution/execution_controller.go
@@ -152,11 +152,21 @@ func (c *Controller) tryDeleteWorkload(clusterName string, work *workv1alpha1.Wo
 			klog.Errorf("Failed to unmarshal workload, error is: %v", err)
 			return err
 		}
-
-		err = c.ObjectWatcher.Delete(clusterName, workload)
-		if err != nil {
-			klog.Errorf("Failed to delete resource in the given member cluster %v, err is %v", clusterName, err)
-			return err
+		deletionPolicy := workload.GetAnnotations()[workv1alpha1.DeletionPolicyAnnotation]
+		switch deletionPolicy {
+		case workv1alpha1.DeletionCompleteAnnotation:
+			klog.V(4).Infof("Found deletionPolicy (%s) on %s. cluster is %s, Removing label/annotation.", deletionPolicy, work.Name, clusterName)
+			err = c.cleanUpLabelAnnotation(clusterName, work)
+			if err != nil {
+				klog.Errorf("failed to remove the label/annotation from all resources previously managed by work %s/%q", work.Namespace, work.Name)
+				return err
+			}
+		default:
+			err = c.ObjectWatcher.Delete(clusterName, workload)
+			if err != nil {
+				klog.Errorf("Failed to delete resource in the given member cluster %v, err is %v", clusterName, err)
+				return err
+			}
 		}
 	}
 
@@ -273,4 +283,21 @@ func (c *Controller) eventf(object *unstructured.Unstructured, eventType, reason
 		return
 	}
 	c.EventRecorder.Eventf(ref, eventType, reason, messageFmt, args...)
+}
+
+func (c *Controller) cleanUpLabelAnnotation(clusterName string, work *workv1alpha1.Work) error {
+	for _, manifest := range work.Spec.Workload.Manifests {
+		workload := &unstructured.Unstructured{}
+		err := workload.UnmarshalJSON(manifest.Raw)
+		if err != nil {
+			klog.Errorf("Failed to unmarshal workload, error is: %v", err)
+			return err
+		}
+		util.CleanUpLabelAnnotation(workload)
+		if err = c.tryCreateOrUpdateWorkload(clusterName, workload); err != nil {
+			klog.Errorf("Failed to create or update resource(%v/%v) in the given member cluster %s, err is %v", workload.GetNamespace(), workload.GetName(), clusterName, err)
+			return err
+		}
+	}
+	return nil
 }
--- a/pkg/controllers/execution/execution_controller_test.go
+++ b/pkg/controllers/execution/execution_controller_test.go
@@ -1,0 +1,138 @@
+package execution
+
+import (
+	"context"
+	workv1alpha1 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha1"
+	"github.com/karmada-io/karmada/pkg/sharedcli/ratelimiterflag"
+	"github.com/karmada-io/karmada/pkg/util"
+	"github.com/karmada-io/karmada/pkg/util/fedinformer/genericmanager"
+	"github.com/karmada-io/karmada/pkg/util/objectwatcher"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+	dynamicfake "k8s.io/client-go/dynamic/fake"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/tools/record"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	fakecontrollerruntimeclient "sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"testing"
+	"time"
+)
+
+func TestController_cleanUpLabelAnnotation(t *testing.T) {
+	type fields struct {
+		Client             client.Client
+		EventRecorder      record.EventRecorder
+		RESTMapper         meta.RESTMapper
+		ObjectWatcher      objectwatcher.ObjectWatcher
+		PredicateFunc      predicate.Predicate
+		InformerManager    genericmanager.MultiClusterInformerManager
+		RatelimiterOptions ratelimiterflag.Options
+	}
+	type args struct {
+		clusterName string
+		work        *workv1alpha1.Work
+		client      dynamic.Interface
+	}
+	restMapper := func() meta.RESTMapper {
+		m := meta.NewDefaultRESTMapper([]schema.GroupVersion{corev1.SchemeGroupVersion})
+		m.Add(schema.GroupVersionKind{Version: "v1", Kind: "ConfigMap"}, meta.RESTScopeNamespace)
+		m.Add(appsv1.SchemeGroupVersion.WithKind("Deployment"), meta.RESTScopeNamespace)
+		return m
+	}()
+	tests := []struct {
+		name                string
+		fields              fields
+		args                args
+		informerManagerFunc func(clusterName string, client dynamic.Interface) genericmanager.MultiClusterInformerManager
+		aop                 func(client dynamic.Interface) func()
+		wantErr             bool
+	}{
+		{
+			name: "workload has karamda's labels/annotations",
+			fields: fields{
+				Client:        fakecontrollerruntimeclient.NewFakeClient(),
+				EventRecorder: record.NewFakeRecorder(2000),
+				RESTMapper:    restMapper,
+			},
+			args: args{
+				clusterName: "cluster1",
+				work: &workv1alpha1.Work{
+					Spec: workv1alpha1.WorkSpec{
+						Workload: workv1alpha1.WorkloadTemplate{
+							Manifests: []workv1alpha1.Manifest{
+								{
+									RawExtension: runtime.RawExtension{
+										Raw: []byte(`{"apiVersion":"apps/v1","kind":"Deployment","metadata":{"name":"test-deployment","namespace":"test-namespace","annotations":{"propagationpolicy.karmada.io/name":"demo-pp","resourcetemplate.karmada.io/managed-annotations":"appkey,kubectl.kubernetes.io/last-applied-configuration,propagationpolicy.karmada.io/name,propagationpolicy.karmada.io/namespace,resourcebinding.karmada.io/name,resourcebinding.karmada.io/namespace,resourcetemplate.karmada.io/managed-annotations,resourcetemplate.karmada.io/managed-labels,resourcetemplate.karmada.io/uid,work.karmada.io/conflict-resolution,work.karmada.io/name,work.karmada.io/namespace"},"labels":{"karmada.io/managed":"true"}}}`),
+									},
+								},
+							},
+						},
+					},
+				},
+				client: dynamicfake.NewSimpleDynamicClient(scheme.Scheme),
+			},
+			informerManagerFunc: func(clusterName string, client dynamic.Interface) genericmanager.MultiClusterInformerManager {
+				mgr := genericmanager.GetInstance()
+				mgr.ForCluster(clusterName, client, time.Second*10)
+				mgr.Start(clusterName)
+				mgr.WaitForCacheSync(clusterName)
+				return mgr
+			},
+			aop: func(client dynamic.Interface) func() {
+				utObj := &unstructured.Unstructured{
+					Object: map[string]interface{}{
+						"apiVersion": "apps/v1",
+						"kind":       "Deployment",
+						"metadata": map[string]interface{}{
+							"name":      "test-deployment",
+							"namespace": "test-namespace",
+							"labels": map[string]interface{}{
+								"karmada.io/managed": "true",
+							},
+							"annotations": map[string]interface{}{
+								"propagationpolicy.karmada.io/name":               "demo-pp",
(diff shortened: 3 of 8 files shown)


pkg/apis/policy/v1alpha1/propagation_types.go
```diff
@@ -173,6 +173,12 @@ type PropagationSpec struct {
 	// +kubebuilder:validation:Enum=Lazy
 	// +optional
 	ActivationPreference ActivationPreference `json:"activationPreference,omitempty"`
+
+	// OrphaningDeletion  represents whether to prevent removal of these managed resources
+	// +kubebuilder:default="Never"
+	// +kubebuilder:validation:Enum=Never;Complete;Incomplete
+	// +optional
+	OrphaningDeletion OrphaningDeletion `json:"orphaningDeletion,omitempty"`
```
XiShanYongYe-Chang, line 181: The first thing that strikes users about this name is that its value is a bool. Is there a better name for it?

CharlesQQ: > The first thing that strikes users about this name is that its value is a bool. Is there a better name for it?

how about  OrphaningDeletionPolicy

pkg/apis/policy/v1alpha1/propagation_types.go
```diff
@@ -549,6 +555,22 @@ const (
...
+	// OrphaningDeletionNever means that the resource in member clusters will be deleted when removal of managed resources
+	OrphaningDeletionNever OrphaningDeletion = "Never"
+
+	// OrphaningDeletionComplete means that the resource in member clusters completely independent of karmada
+	// Not only cleanup the labels/annotations from the resource on member clusters, but also delete work object in control plane
+	OrphaningDeletionComplete OrphaningDeletion = "Complete"
+
+	// OrphaningDeletionIncomplete means that work object and the resource willed be kept.
+	// But if work object deleted manually, cleanup the labels/annotations from the resource on member clusters
+	OrphaningDeletionIncomplete OrphaningDeletion = "Incomplete"
```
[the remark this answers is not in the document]
CharlesQQ: > Is there a usage scenario for this type?

Refer to https://github.com/karmada-io/karmada/issues/3004

pkg/apis/policy/v1alpha1/propagation_types.go
XiShanYongYe-Chang: I think it's okay. Let me give one for reference: `PropogateDeletionPolicy`.

/cc @chaunceyjiang @zhy76 @RainbowMango

pkg/apis/policy/v1alpha1/propagation_types.go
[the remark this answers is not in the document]
XiShanYongYe-Chang: After reading the description and discussion in this issue, it seems that the work is not required to be retained. In terms of use, what can be done to retain work?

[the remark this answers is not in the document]
chaunceyjiang: We can first just implement `Complete` and `Never`, and discuss the other options later once we have more specific requirements.

pkg/controllers/binding/binding_controller.go
```diff
@@ -82,9 +82,11 @@ func (c *ResourceBindingController) Reconcile(ctx context.Context, req controlle
 
 	if !binding.DeletionTimestamp.IsZero() {
 		klog.V(4).Infof("Begin to delete works owned by binding(%s).", req.NamespacedName.String())
-		if err := helper.DeleteWorks(c.Client, req.Namespace, req.Name, binding.Labels[workv1alpha2.ResourceBindingPermanentIDLabel]); err != nil {
-			klog.Errorf("Failed to delete works related to %s/%s: %v", binding.GetNamespace(), binding.GetName(), err)
-			return controllerruntime.Result{}, err
+		if binding.Spec.OrphaningDeletion == policyv1alpha1.OrphaningDeletionNever || binding.Spec.OrphaningDeletion == policyv1alpha1.OrphaningDeletionComplete {
```
chaunceyjiang, line 85: I'd suggest using a `switch` here, as it will make the logic clearer. It will also make it easier for us to add comments for each `case`.

pkg/apis/policy/v1alpha1/propagation_types.go
```diff
@@ -173,6 +173,12 @@ type PropagationSpec struct {
 	// +kubebuilder:validation:Enum=Lazy
 	// +optional
 	ActivationPreference ActivationPreference `json:"activationPreference,omitempty"`
+
+	// OrphaningDeletion  represents whether to prevent removal of these managed resources
```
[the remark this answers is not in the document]
chaunceyjiang: I think we also need to explain each of the options - what "Never" means, what "Complete" means.

pkg/apis/policy/v1alpha1/propagation_types.go
chaunceyjiang: > how about OrphaningDeletionPolicy

+1

pkg/apis/policy/v1alpha1/propagation_types.go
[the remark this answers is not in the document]
XiShanYongYe-Chang: +1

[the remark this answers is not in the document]
CharlesQQ: > We can first just implement Complete and Never, and discuss the other options later once we have more specific requirements.

All right!

pkg/controllers/binding/binding_controller.go
CharlesQQ: > I'd suggest using a switch here, as it will make the logic clearer. It will also make it easier for us to add comments for each case.

Only one case now, it's not necessary

pkg/apis/policy/v1alpha1/propagation_types.go
```diff
@@ -173,6 +173,12 @@ type PropagationSpec struct {
...
+	// +kubebuilder:default="Never"
+	// +kubebuilder:validation:Enum=Never;Complete
+	// +optional
+	OrphaningDeletionPolicy OrphaningDeletionPolicyType `json:"orphaningDeletionPolicy,omitempty"`
```
chaunceyjiang, line 181: I believe we only need to add this field in `PP`. The `work` does not require this field.

In addition, I don't think the current implementation can fully handle all situations.

Here's a new idea:

When users set `orphaningDeletionPolicy` of PP as `Complete`.

The specific implementation should be adding an `annotation` on the `workloadTemplat`e of the control plane, similar to `orphaningDeletionPolicy.karmada.io: "Complete"`

Then when deleting `workloadTemplate` in `execution_controller`, check if `workloadTemplate` has annotation `orphaningDeletionPolicy.karmada.io`. If it is `Complete`, do not delete it.

The benefits are:
1. The personnel creating the `workloadTemplate` can also control whether this application needs to be retained in member clusters, rather than only those who create PP.
2. Resources like namespace that do not have `PP` can also be deleted on the control plane without being deleted from member clusters.

CharlesQQ: > Then when deleting workloadTemplate in execution_controller, check if workloadTemplate has annotation orphaningDeletionPolicy.karmada.io. If it is Complete, do not delete it.

Execution controller must watch all origin resource such as  resourceTemlate 、namespace and so on, which make too complicated, I don't think that's a good idea.
I think for namespace, can add annotation like `orphaningDeletionPolicy.karmada.io: "Complete"`; FederatedResourceQuota those create work object direct,  by adding field `orphaningDeletionPolicy: Complete` if necessary

chaunceyjiang: > Execution controller must watch all origin resource such as resourceTemlate 、namespace and so on

There's no need, this is existing logic.

XiShanYongYe-Chang: This is no different from directly adding annotations to the resource template. In addition, users can directly skip pp to control the annotation value on the resource template.

chaunceyjiang: > This is no different from directly adding annotations to the resource template. In addition, users can directly skip pp to control the annotation value on the resource template.



Yes, I think that's reasonable.

XiShanYongYe-Chang:
Hi @CharlesQQ @chaunceyjiang Is there any blockage in this PR now? Can we move forward?

pkg/apis/policy/v1alpha1/propagation_types.go
```diff
@@ -549,6 +555,18 @@ const (
...
+// OrphaningDeletionPolicyType indicates how removal action of managed resources
+type OrphaningDeletionPolicyType string
+
+const (
+	// OrphaningDeletionNever means that the resource in member clusters will be deleted when removal of managed resources
+	OrphaningDeletionNever OrphaningDeletionPolicyType = "Never"
+
+	// OrphaningDeletionComplete means that the resource in member clusters completely independent of karmada
+	// Not only cleanup the labels/annotations from the resource on member clusters, but also delete work object in control plane
+	OrphaningDeletionComplete OrphaningDeletionPolicyType = "Complete"
```
XiShanYongYe-Chang, line 567: Complete sounds a bit confusing. Is it possible to call it `CompleteDelete`?

chaunceyjiang:
> Hi @CharlesQQ @chaunceyjiang Is there any blockage in this PR now? Can we move forward?

https://github.com/karmada-io/karmada/pull/4788#discussion_r1582008275  Regarding this plan, it seems that there is no conclusion yet?

XiShanYongYe-Chang:
Let's invite @RainbowMango and @whitewindmills give some opinion.
/cc @RainbowMango @whitewindmills

pkg/controllers/binding/binding_controller.go
```diff
@@ -82,9 +82,11 @@ func (c *ResourceBindingController) Reconcile(ctx context.Context, req controlle
 
 	if !binding.DeletionTimestamp.IsZero() {
 		klog.V(4).Infof("Begin to delete works owned by binding(%s).", req.NamespacedName.String())
-		if err := helper.DeleteWorks(c.Client, req.Namespace, req.Name, binding.Labels[workv1alpha2.ResourceBindingPermanentIDLabel]); err != nil {
-			klog.Errorf("Failed to delete works related to %s/%s: %v", binding.GetNamespace(), binding.GetName(), err)
-			return controllerruntime.Result{}, err
+		if binding.Spec.OrphaningDeletionPolicy == policyv1alpha1.OrphaningDeletionNever || binding.Spec.OrphaningDeletionPolicy == policyv1alpha1.OrphaningDeletionComplete {
```
whitewindmills, line 85: feel unnecessary to add such a statement.

pkg/apis/policy/v1alpha1/propagation_types.go
whitewindmills: > I believe we only need to add this field in PP. The work does not require this field.

+1, but I prefer to decouple from PP/CPP, cause I'm afraid this is very unfriendly to administrators if the system also has the `PropagationPolicyPreemption` feature turned on. in addition, I feel that propagation policies are becoming more and more complex, and almost all new features must consider adding APIs to it.

XiShanYongYe-Chang: > but I prefer to decouple from PP/CPP, cause I'm afraid this is very unfriendly to administrators if the system also has the PropagationPolicyPreemption feature turned on.

Can you give an example of this?

whitewindmills: orphaning-delete

if decoupled from PP/CPP, the administrator don't care about it.

pkg/apis/policy/v1alpha1/propagation_types.go
buji-code: ArgoCD uses the term ["cascade"](https://argo-cd.readthedocs.io/en/stable/user-guide/app_deletion/)

pkg/apis/policy/v1alpha1/propagation_types.go
XiShanYongYe-Chang: Thanks, I get your meaning.  How do you think @CharlesQQ @chaunceyjiang @RainbowMango

pkg/apis/policy/v1alpha1/propagation_types.go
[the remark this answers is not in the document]
RainbowMango: The field is ambiguous: delete `what`?  `resource template` or `PropagationPolicy`?

I know you mean `resource template`, but it should be clearly noted(try with field name). Otherwise, user might think it works for `PropagationPolicy` since this is a specification from PropagationPolicy.

pkg/apis/policy/v1alpha1/propagation_types.go
RainbowMango: > but I prefer to decouple from PP/CPP, cause I'm afraid this is very unfriendly to administrators if the system also has the PropagationPolicyPreemption feature turned on.

I do not quite understand how `unfriendly` here, I think the preemption is clear that after a PropagationPolicy(with higher priority) successfully preempted a resource, then the resource should be propagated as per the PropagationPolicy.

pkg/apis/policy/v1alpha1/propagation_types.go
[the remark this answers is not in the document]
buji-code: What do you think about:
`ResourceDeletionPolicy: Background/Foreground/Orphan`, like appears on [Kubernetes Docs](https://kubernetes.io/docs/tasks/administer-cluster/use-cascading-deletion/)

When:
**Background** (default) - marks the resource from Karmada for deletion, as well as its propagated resources, and lets the GC collect the resources asynchronically.
**Foreground** - marks the resource from Karmada for deletion, as well as its propagated resources, and waits for removal of owned/propagated resources before collecting the main resource.
**Orphan** - marks the resource from Karmada for deletion, but leaves its propagated ones untouched.

CharlesQQ:
>  Regarding this plan, it seems that there is no conclusion yet?

Yes, To sum up, it seems inappropriate to put fields in pp/cpp

pkg/apis/policy/v1alpha1/propagation_types.go
chaunceyjiang: > It seems that using ["cascade"](https://argo-cd.readthedocs.io/en/stable/user-guide/app_deletion/) would be better, as it can reduce the learning cost for users.

+1

It seems that using ["cascade"](https://argo-cd.readthedocs.io/en/stable/user-guide/app_deletion/) would be better, as it can reduce the learning cost for users.

chaunceyjiang:
@XiShanYongYe-Chang @RainbowMango @CharlesQQ @whitewindmills 
Do we have a consensus on the current proposal?

XiShanYongYe-Chang:
We may not reached a unanimous conclusion yet. Work suspension might also can learn from this discussion. They are both processing policies for resource templates and can be decoupled from PP.

There should be three options at this point,

- Solution 1: Define a resource deletion policy in the PP. 
- Solution 2: Add an annotation to the resource template to describe the deletion policy of the target resource.
- Solution 3: A new CRD is defined to describe a deletion policy of one or more resources.

You can publish it, or we can have a meeting to discuss and decide.

XiShanYongYe-Chang:
> @XiShanYongYe-Chang @RainbowMango @CharlesQQ @whitewindmills Do we have a consensus on the current proposal?

Can we set up a time to discuss this plan? When would be a good time for you, or maybe next Tuesday at the community meeting?

buji-code:
Any update on this one?

XiShanYongYe-Chang:
@buji-code Yeah, after discussion at the community meeting, we decided to try out solution 2 and 3. This work will start in the next release, early next month.

XiShanYongYe-Chang:
Hi guys @CharlesQQ @whitewindmills @chaunceyjiang @RainbowMango @buji-code 

Can we proceed with `solution 2: Add an annotation to the resource template to describe the deletion policy of the target resource` first, and then consider designing and adding a CRD to describe the deletion policy as the scenario develops? (Of course, the use of annotations is not excluded as the ideal end use.)

pkg/controllers/execution/execution_controller.go
```diff
@@ -152,11 +152,21 @@ func (c *Controller) tryDeleteWorkload(clusterName string, work *workv1alpha1.Wo
 			klog.Errorf("Failed to unmarshal workload, error is: %v", err)
 			return err
 		}
-
```
chaunceyjiang, line 155: I think we can write an e2e to test this PR.

1. Check whether the resources in the member cluster have cleared the unique label of karmada.
2. Detect whether the resource still exists in the member cluster after the control plane deletes it.
3. The resources tested should include `namespace` resources.

pkg/apis/work/v1alpha1/work_types.go
```diff
@@ -30,6 +30,13 @@ const (
 	ResourcePluralWork = "works"
 	// ResourceNamespaceScopedWork indicates if Work is NamespaceScoped.
 	ResourceNamespaceScopedWork = true
+
+	// OrphaningDeletionPolicyAnnotation is added to resource template annotation to indicates how removal action of managed resources.
+	OrphaningDeletionPolicyAnnotation = "orphaningdeletionpolicy.karmada.io"
```
XiShanYongYe-Chang, line 35: I have a naming suggestion: `resourcetemplate.karmada.io/deletionpolicy`

pkg/apis/work/v1alpha1/work_types.go
```diff
@@ -30,6 +30,13 @@ const (
...
+
+	// OrphaningDeletionCompleteAnnotation means that the resource in member clusters completely independent of karmada
+	// Not only cleanup the labels/annotations from the resource on member clusters, but also delete work object in control plane
+	OrphaningDeletionCompleteAnnotation = "complete"
```
XiShanYongYe-Chang, line 39: I raise one question, for the resources that follow the distribution, do we also need to consider, for example, when the deployment is deleted, what is the deletion policy of the following resources such as comfigmap and secret? How do you resolve conflicts when a resource that follows a distribution has its own deletion policy?

XiShanYongYe-Chang (commented):
Can we submit a proposal to analyze and design the current feature? In this way, after the feature is completed, the feature design process can be recorded in a document, which facilitates subsequent iteration.

pkg/apis/work/v1alpha1/work_types.go
lcw2: Another question, when deleting namespace in control plane without OrphaningDeletionPolicyAnnotation , but there  is still resources in namespace in member cluster,  will the namespace in the member clusters be deleted or not?
I think it's better to retain the namespace in member cluster.

CharlesQQ: Your comments will be considered and a new proposal will be submitted for discussion

a7i:
👋🏼 happy to contribute if I can push this forward (design or implementation)

RainbowMango:
> 👋🏼 happy to contribute if I can push this forward (design or implementation)

Sure thing! Thank you Amir in advance. 
I'll look at the proposal and see how I can help with it.

XiShanYongYe-Chang:
Hi everyone, let's use umbrella issue https://github.com/karmada-io/karmada/issues/5577 to track this requirement.
/close

PR #4743, opened 2024-03-25
Deprecate name/namespace/key labels

whitewindmills:
**What type of PR is this?**
/kind feature
/kind cleanup

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:
part of [isse](https://github.com/karmada-io/karmada/issues/4711)

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #4793, opened 2024-04-02
Use merge strategy to avoid work.karmada.io/permanent-id changes

whitewindmills:
**What type of PR is this?**
/kind bug

Since labels will be overwritten every time they are updated and created, this causes `work.karmada.io/permanent-id` to constantly change.
I add some debug logs in karmada-webhook to capture the bug:
image

Fixes #

```release-note
karmada-controller-manager: Fix the problem that work.karmada.io/permanent-id constantly changes with every update.
```

### DIFF
--- a/pkg/util/helper/work.go
+++ b/pkg/util/helper/work.go
@@ -78,8 +78,8 @@ func CreateOrUpdateWork(client client.Client, workMeta metav1.ObjectMeta, resour
 			}
 
 			runtimeObject.Spec = work.Spec
-			runtimeObject.Labels = work.Labels
-			runtimeObject.Annotations = work.Annotations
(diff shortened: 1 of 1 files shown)


XiShanYongYe-Chang (commented):
Nice finding! Did you discover the problem during development?

/lgtm
/kind bug

whitewindmills:
> Did you discover the problem during development?

Yes, this causes CI to always fail.

whitewindmills:
Another PR https://github.com/karmada-io/karmada/pull/4743 still needs this, could you take look?
/cc @RainbowMango

XiShanYongYe-Chang:
Hi @whitewindmills, how about adding the release note, I think we should cherry-pick this patch to the previous branch, how do you think?

RainbowMango (approved):
/approve

Issue #4637, opened 2024-02-18
Bump Prow to v20240517-ea10bd814


RainbowMango:
**Why is this needed**:
We want the feature that triggers failed workflow which was introduced since v20220105-8e97316bf9.
See the discussion https://github.com/karmada-io/karmada/issues/3421#issuecomment-1751961909 for more details.

**Migration progress**:
- [x] [karmada-io/karmada](https://github.com/karmada-io/karmada)
- [x] [karmada-io/website](https://github.com/karmada-io/website)
- [x] [karmada-io/community](https://github.com/karmada-io/community)
- [x] [karmada-io/kubernetes](https://github.com/karmada-io/kubernetes)
- [x] [karmada-io/multicluster-cloud-provider](https://github.com/karmada-io/multicluster-cloud-provider)
- [x] [karmada-io/api](https://github.com/karmada-io/api)
- [x] [karmada-io/dashboard](https://github.com/karmada-io/dashboard)
- [x] [karmada-io/multi-cluster-ingress-nginx](https://github.com/karmada-io/multi-cluster-ingress-nginx)
- [x] [karmada-io/playground](https://github.com/karmada-io/playground)

All images can be found here: https://console.cloud.google.com/gcr/images/k8s-prow

RainbowMango:
I tested the basic functionality against the https://github.com/karmada-io/website/pull/563 and it works well.
The next step is to test the `retest` function.

Here is the main configuration update(add the `triggers`):
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  namespace: karmada-prowbot
  name: config
data:
  config.yaml: |
    triggers:
    - repos:
      - karmada-io/karmada
      - karmada-io/website
      # JoinOrgURL is a link that redirects users to a location where they
      # should be able to read more about joining the organization in order
      # to become trusted members.
      join_org_url: "https://github.com/karmada-io/community/blob/main/community-membership.md#member"
      only_org_members: true
      # TriggerGitHubWorkflows enables workflows run by github to be triggered by prow.
      trigger_github_workflows: true
```

jwcesign:
I tried it in my env and it works fine: https://github.com/jwcesign/kubectl-proxy-demo/pull/3

My config:
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  namespace: prow
  name: plugins
data:
  plugins.yaml: |
    triggers:
    - repos:
      - jwcesign/kubectl-proxy-demo
[...]

PR #4794, opened 2024-04-02
use single source for go version

grosser:
**What type of PR is this?**

/kind cleanup

- unify version that ci uses
- make life easier for anyone using a go version manager

```release-note
NONE
```


XiShanYongYe-Chang:
Hi @grosser, do you have any documentation on how to use the `.go-version` file?

RainbowMango:
I'm also curious about how it works.
I couldn't find more information, except the asdf thing:
https://github.com/asdf-community/asdf-golang?tab=readme-ov-file#version-selection

liangyuanpeng (commented):
You still need to using it on the workflow, otherwise it will not working.

https://github.com/etcd-io/etcd/blob/547aa643e69d55328b203f40ccec50256b44158e/.github/workflows/build.yaml#L24C1-L28C63

grosser:
trying with `go-version-file` that hopefully works 🤞 

tools like go-env / asdf / mise can read this file to set the local go version
some of them can also read `.go.mod`, but asdf and mise refuse to read from it because of some rabbithole

.github/workflows/ci.yml
```diff
@@ -41,9 +41,9 @@ jobs:
         with:
           path: ${{ env.GOPATH }}/src/github.com/karmada-io/karmada
       - name: install Go
-        uses: actions/setup-go@v3
+        uses: actions/setup-go@v5
         with:
-          go-version: 1.21.8
+          go-version-file: .go-version
```
liangyuanpeng, line 46:
```suggestion
          go-version-file: ${{ env.GOPATH }}/src/github.com/karmada-io/karmada/.go-version
```

grosser:
👋 need a CI approval

liangyuanpeng:
The karmada's maintainers will return from vacation tomorrow.

/lgtm

liangyuanpeng:
I think it's need to squash the commit

grosser:
squashed into single commit

if single commit is a requirement, making the PR template say it would speed up PRs

grosser:
and need another approval @liangyuanpeng

RainbowMango:
> and need another approval @liangyuanpeng

Done.

.github/workflows/ci-schedule.yml
```diff
@@ -38,9 +38,9 @@ jobs:
           # We need to guess version via git tags.
           fetch-depth: 0
       - name: install Go
-        uses: actions/setup-go@v3
+        uses: actions/setup-go@v5
         with:
-          go-version: 1.21.8
+          go-version-file: .go-version
```
RainbowMango, line 43: It seems the `go-version-file` accepts the path of `go.mod` or `go.work`.
The `actions/setup-go` first tries to parse the go version from `go.mod` or `go.work` by a regular expression `/^go (\d+(\.\d+)*)/m`, and fall back to `trim` the specified file, echo the [relevant code](https://github.com/actions/setup-go/blob/be1aa1186e83689f3e45adaf3697af86f816fb2e/dist/setup/index.js#L88328-L88336) here:
```js
function parseGoVersionFile(versionFilePath) {
    const contents = fs_1.default.readFileSync(versionFilePath).toString();
    if (path.basename(versionFilePath) === 'go.mod' ||
        path.basename(versionFilePath) === 'go.work') {
        const match = contents.match(/^go (\d+(\.\d+)*)/m);
        return match ? match[1] : '';
    }
    return contents.trim();
}
```
That would work I think.

.github/workflows/ci.yml
```diff
@@ -41,9 +41,9 @@ jobs:
         with:
           path: ${{ env.GOPATH }}/src/github.com/karmada-io/karmada
       - name: install Go
-        uses: actions/setup-go@v3
+        uses: actions/setup-go@v5
         with:
-          go-version: 1.21.8
+          go-version-file: ${{ env.GOPATH }}/src/github.com/karmada-io/karmada/.go-version
```
RainbowMango, line 46: Can you remind me why we have to specify the whole path of `.go-version` here?

RainbowMango (commented):
I believe the ideal approach is to have only one place in the repository where the Go version is specified, and that is the `go.mod` file.
And the `setup-go` is working towards that direction, see https://github.com/actions/setup-go/pull/458.

@grosser @liangyuanpeng 
I would like to hear your opinions. (Note that I'm not opposing the current approach, jus for discussion and try to work out an ideal solution.)

RainbowMango:
Hey guys, I just made a test on my forked repo, and seems `actions/setup-go@v5` can read Go version from `go.mod` file.
Here is [the change](https://github.com/RainbowMango/karmada/commit/55f7a1fdcf604a325630839241ec2fbf0ce7ae9e):
```diff
diff --git a/.github/workflows/ci.yml b/.github/workflows/ci.yml
index 6cec21009..3c91b0dbf 100644
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,9 +16,9 @@ jobs:
       - name: checkout code
         uses: actions/checkout@v3
       - name: install Go
-        uses: actions/setup-go@v3
+        uses: actions/setup-go@v5
         with:
-          go-version: 1.21.8
+          go-version-file: go.mod
       - name: verify license
         run: hack/verify-license.sh
       - name: vendor
diff --git a/go.mod b/go.mod
index 13d6c12c9..3901dc429 100644
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/karmada-io/karmada

-go 1.21
+go 1.21.1
 
 require (
        github.com/adhocore/gronx v1.6.3
```
I intentionally set the Go version to v1.21.1 instead of the current using version(v1.21.8), to test if it works. The test result shows it indeed works as expected. See the workflow logs here: https://github.com/RainbowMango/karmada/actions/runs/8585586406/job/23527142126#step:3:18

RainbowMango:
> unify version that ci uses

I guess we can solve it by the test approach mentioned on above.

> make life easier for anyone using a go version manager

Which go version manager are you using? can it read go version from the go.mod file?

grosser:
I'm using `mise`, which refuses to read the go.mod sadly (already opened issue, but they say "go.mod says what the minimum version is")
(same for `asdf`)
go-env does read it though
using go.mod for this and having a .go-version file would also work fine

.github/workflows/ci.yml
grosser: because the checkout is in a different directory

.github/workflows/ci-schedule.yml
grosser: yeah that could work ... never saw go.work being used

RainbowMango:
> I'm using mise, which refuses to read the go.mod sadly (already opened issue, but they say "go.mod says what the minimum version is")

Which `mise`? Can you share the issue link?

grosser:
https://github.com/jdx/mise/issues/1689#issuecomment-1955357190

.github/workflows/ci-schedule.yml
RainbowMango: Yes, Karmada-io/karmada only has one module inside, so it doesn't need the go.work.

RainbowMango:
I haven't used `mise` before, but I just tried it out.  I still don't know how the [auto-switching](https://mise.jdx.dev/lang/go.html#go-version-file-support) works, and what its benefit is. Can you help to explain it a little bit?

grosser:
auto-switching switches the go version when cd-ing into a directory according to the directories .go-version

grosser:
I'm ok with either ... just want someone with merge permission to tell me what they would be willing to merge and will make adjustments accordingly.
Would also love to merge as is and create followup for adding to go.mod to move the discussion 🤷

RainbowMango:
It seems both `mise` and `asdf` provided the auto-switching feature:
- For asdf: https://asdf-vm.com/manage/configuration.html#tool-versions
- mise: https://mise.jdx.dev/configuration.html#tool-versions

They both support `.tool-versions`, can we use this file?

RainbowMango:
> auto-switching switches the go version when cd-ing into a directory according to the directories .go-version

This feature is awesome by the way. Just tried it out.

grosser:
why would we need .tool-versions, this repo only uses go right ?

RainbowMango:
yeah, since `.tool-versions` can be consumed by both asdf and mise users. 

I mean take `go.mod` as a single source of truth, and the `.tool-versions` dedicated to those go-version managers.

liangyuanpeng:
I think @RainbowMango  is mean we want to use `go.mod` for the project and CI, then `.tool-versions` to working for go-version managers.

There are two topic, the Part1 is go version for build & CI, Part2 is the file for people who using the go-version managers.

I can see common practice ( kubernetes/kind/etcd ) is that `go mod` is set to the language version, and `.go-version` uses the higher version for compilation and building.

But as @RainbowMango  say, using `go.mod` to unify go versions in CI may be a direction, so maybe we can try to using `go.mod` for CI and `.tool-versions` to working go version manager.

> auto-switching switches the go version when cd-ing into a directory according to the directories .go-version

would you like to try with `.tool-versions` ?

grosser:
asdf also [supports .go-version](https://github.com/asdf-community/asdf-golang?tab=readme-ov-file#version-selection), so can we avoid 1 level of indirection by using that instead of .tool-versions ?

RainbowMango:
> asdf also [supports .go-version](https://github.com/asdf-community/asdf-golang?tab=readme-ov-file#version-selection), so can we avoid 1 level of indirection by using that instead of .tool-versions ?

Thanks for the reminder. I didn't notice it. then I belive `.go-version` is better than `.tool-versions`.

RainbowMango:
By the way, the `actions/setup-go` has been updated to `v5` by #4756, you might need to rebase when making adjustments.

grosser:
rebased + updated go.mod to have full version

... is this mergeable now or what still needs to change ?

RainbowMango (commented):
Looks good to me!
/lgtm

@liangyuanpeng Do you have further comments?

grosser:
plz restart the 1 failed ci job, seems unrelated ...

RainbowMango:
Done. 
This reminds me to hurry up the #4637, after that we can restart the test by /retest command.

grosser:
still 2 checks pending ... any idea how to get them to pass ?

RainbowMango:
> still 2 checks pending ... any idea how to get them to pass ?

All good now. Ready to move forward now.
/approve
@liangyuanpeng If you have any further comments, please continue...

PR #4795, opened 2024-04-03
when controller return error is non-nil not need to set Result.Requeue 

yanfeng1992:

**What type of PR is this?**
/kind cleanup

The Controller will requeue the Request to be processed again if an error is non-nil or Result.Requeue is true, an error is non-nil has higher priority. Therefore, there is no need to set Result.Requeue to true when the error is non-nil.

Fixes #

```release-note
NONE
```


yanfeng1992:
> Thanks @yanfeng1992, I've actually had a partial cleanup before. /assign
> 
> A ut has been failed, can you help fix it?

fixed it @XiShanYongYe-Chang

XiShanYongYe-Chang (commented):
/lgtm
/approve

PR #4801, opened 2024-04-04
[WIP] feat/karmadactl: extra etcd support without cert.

liangyuanpeng:
**What type of PR is this?**

/kind feature
**What this PR does / why we need it**:

I'm using karmadactl to deploy karmada and i will get the error when i only set the `external-etcd-servers` without certificate.

So this PR is working for support without certificate for external etcd when using karmadactl to init karmada.

I can without the certificate of etcd when I use `kubeadm` to configure external etcd,  so I want to get the same experience here.

> And what is the effect when configue external etcd whitout providing the certificates?

It will skip to deploy etcd (static pod)

doc ref:
(doc about caFile/certFile/keyFile of ExternalEtcd  is not correct, it is not required)
- https://kubernetes.io/docs/reference/config-api/kubeadm-config.v1beta3/#kubeadm-k8s-io-v1beta3-ExternalEtcd
- https://kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm-init/#init-workflow   (step 4)

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
kramadactl: extra etcd support without cert now.
```

PR #4824, opened 2024-04-10
chore: remove unuse ghaction step and collect logs of karmadactl test.

liangyuanpeng:
**What type of PR is this?**

/kind cleanup

this CI is using the kubernetes from created by `cli-testing-environment` and have not used `helm/kind-action`, so juse remove it..

https://github.com/karmada-io/karmada/blob/4898c0fc55a13033fcf21fcc46a41318548c5713/hack/cli-testing-environment.sh#L53-L55

the second change is collect logs for kind kubernetes.

/assign @RainbowMango 

Fixes #

```release-note

```

### DIFF
--- a/.github/workflows/cli.yaml
+++ b/.github/workflows/cli.yaml
@@ -36,9 +36,7 @@ jobs:
         uses: actions/setup-go@v5
         with:
           go-version-file: go.mod
-      - uses: helm/kind-action@v1.10.0
-        with:
-          version: "v0.22.0"
+
       - name: run karmadactl init test
         run: |
           export CLUSTER_VERSION=kindest/node:${{ matrix.k8s }}
@@ -50,11 +48,18 @@ jobs:
           export KUBECONFIG=${HOME}/karmada/karmada-apiserver.config
           GO111MODULE=on go install github.com/onsi/ginkgo/v2/ginkgo
           ginkgo -v --race --trace -p  --focus="[BasicPropagation] propagation testing deployment propagation testing"  ./test/e2e/
-      - uses: chainguard-dev/actions/kind-diag@main
-        # Only upload logs on failure.
-        if: ${{ failure() }}
+      - name: export logs 
+        if: always()
+        run: |
+          export ARTIFACTS_PATH=${{ github.workspace }}/karmadactl-test-logs/${{ matrix.k8s }}/
+          mkdir -p $ARTIFACTS_PATH
+
+          mkdir -p $ARTIFACTS_PATH/karmada-host
+          kind export logs --name=karmada-host $ARTIFACTS_PATH/karmada-host
+      - name: upload logs
+        if: always()
+        uses: actions/upload-artifact@v4
         with:
-          cluster-resources: nodes,namespaces,
-          namespace-resources: configmaps,pods,svc
-          artifact-name: logs-${{ matrix.k8s}}
+          name: karmadactl_test_logs_${{ matrix.k8s }}
+          path: ${{ github.workspace }}/karmadactl-test-logs/${{ matrix.k8s }}/
 
--- a/hack/cli-testing-environment.sh
+++ b/hack/cli-testing-environment.sh
@@ -33,6 +33,29 @@ MEMBER_CLUSTER_2_NAME=${MEMBER_CLUSTER_2_NAME:-"member2"}
 CLUSTER_VERSION=${CLUSTER_VERSION:-"${DEFAULT_CLUSTER_VERSION}"}
 BUILD_PATH=${BUILD_PATH:-"_output/bin/linux/amd64"}
 
+
+# install kind and kubectl
+kind_version=v0.22.0
+echo -n "Preparing: 'kind' existence check - "
+if util::cmd_exist kind; then
+  echo "passed"
+else
+  echo "not pass"
+  util::install_tools "sigs.k8s.io/kind" $kind_version
+fi
+# get arch name and os name in bootstrap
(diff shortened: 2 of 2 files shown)


RainbowMango:
@zhzhuang-zju Please take a look

.github/workflows/cli.yaml
```diff
@@ -48,11 +45,20 @@ jobs:
           export KUBECONFIG=${HOME}/karmada/karmada-apiserver.config
           GO111MODULE=on go install github.com/onsi/ginkgo/v2/ginkgo
           ginkgo -v --race --trace -p  --focus="[BasicPropagation] propagation testing deployment propagation testing"  ./test/e2e/
-      - uses: chainguard-dev/actions/kind-diag@main
-        # Only upload logs on failure.
-        if: ${{ failure() }}
+      - name: export logs 
+        if: always()
```
zhzhuang-zju, line 49: Is it necessary to collect logs when workflow succeeds or is canceled as well?

.github/workflows/cli.yaml
```diff
@@ -48,11 +45,20 @@ jobs:
...
+        run: |
+          export ARTIFACTS_PATH=${{ github.workspace }}/karmadactl-test-logs/${{ matrix.k8s }}/
+          mkdir -p $ARTIFACTS_PATH
+
+          mkdir -p $ARTIFACTS_PATH/{karmada-host,member1,member2}
+          kind export logs --name=karmada-host $ARTIFACTS_PATH/karmada-host
+          kind export logs --name=member1 $ARTIFACTS_PATH/member1
```
zhzhuang-zju, line 56: This workflow is used to test the functionality of `karmadactl init`, is it necessary to collect logs of member clusters?

zhzhuang-zju:
@liangyuanpeng  This PR consists of two parts actually as you mentioned in the PR description:
- remove step `helm/kind-action`
   Just as you said, this CI is using the kubernetes from created by cli-testing-environment, so step `kind-action` can be removed. But I have two questions. 
  1. When was `kind` installed on the environment, I don't find anywhere in `cli-testing-environment` where `kind` is installed, am I missing it?
  2. How did you find this? I mean, "kind-action" isn't used, but it shouldn't have an effect on workflow and is hard to detect.
- collect logs for kind kubernetes
  I‘m okay with this part.

BTW, thanks for this commit~

liangyuanpeng:
> When was kind installed on the environment, I don't find anywhere in cli-testing-environment where kind is installed, am I missing it?

Good catch! seems like it's provided by github action or some github action cache. i will update the `cli-testing-environment.sh` to check and install the command of `kind`.

> How did you find this? I mean, "kind-action" isn't used, but it shouldn't have an effect on workflow and is hard to detect.

I am wroting the code for https://github.com/karmada-io/karmada/pull/4801 and I was not satisfied with the workflow. He increased the number of parallel runs of the workflow, so i more research this workflow and found it.

@zhzhuang-zju

.github/workflows/cli.yaml
liangyuanpeng: Also running the test of `BasicPropagation`, this is the reason what i want to collect them.

.github/workflows/cli.yaml
liangyuanpeng: I hope it can be treated the same as CI, https://github.com/karmada-io/karmada/blob/4898c0fc55a13033fcf21fcc46a41318548c5713/.github/workflows/ci.yml#L158

liangyuanpeng:
In the next PR, I will continue to update this workflow, use one action to build the image and karmadactl, and then test it in other actions, CI workflow is the same thing, and let's talk about it in other space.

.github/workflows/cli.yaml
zhzhuang-zju: If it's because the test of `BasicPropagation`, would it be enough to just collect the logs of `karmada control plane`?
Take `run-e2e.sh` for example, it only collects the logs of `karmada host plane` and `pull-mode-cluster`.  And through investigations, they collect the logs of `pull-mode-cluster` just for `karmada-agent`. 
  
https://github.com/karmada-io/karmada/blob/4898c0fc55a13033fcf21fcc46a41318548c5713/hack/run-e2e.sh#L59-L71

zhzhuang-zju:
> use one action to build the image and karmadactl

There seems to have been a discussion about it before，https://github.com/karmada-io/karmada/pull/3168#discussion_r1129052862 @RainbowMango 

> In the next PR, I will continue to update this workflow

Very welcome. If you make any progress, please ping me~

liangyuanpeng:
rebased main

/assign @zhzhuang-zju

.github/workflows/cli.yaml
```diff
@@ -50,11 +48,18 @@ jobs:
...
+        if: always()
+        run: |
+          export ARTIFACTS_PATH=${{ github.workspace }}/karmadactl-test-logs/${{ matrix.k8s }}/
+          mkdir -p $ARTIFACTS_PATH
+
+          mkdir -p $ARTIFACTS_PATH/karmada-host
+          kind export logs --name=karmada-host $ARTIFACTS_PATH/karmada-host
+      - name: upload logs
+        if: always()
+        uses: actions/upload-artifact@v3
```
zhzhuang-zju, line 61:
```suggestion
        uses: actions/upload-artifact@v4
```

zhzhuang-zju (commented):
others LGTM

.github/workflows/cli.yaml
liangyuanpeng: done

PR #4836, opened 2024-04-16
Add finalizer for propagation policy

whitewindmills:
**What type of PR is this?**
/kind feature

See https://github.com/karmada-io/karmada/pull/4743#discussion_r1560637648

```release-note
NONE
```

### DIFF
--- a/pkg/detector/detector.go
+++ b/pkg/detector/detector.go
@@ -138,7 +138,7 @@ func (d *ResourceDetector) Start(ctx context.Context) error {
 		Version:  policyv1alpha1.GroupVersion.Version,
 		Resource: "propagationpolicies",
 	}
-	policyHandler := fedinformer.NewHandlerOnEvents(d.OnPropagationPolicyAdd, d.OnPropagationPolicyUpdate, d.OnPropagationPolicyDelete)
+	policyHandler := fedinformer.NewHandlerOnEvents(d.OnPropagationPolicyAdd, d.OnPropagationPolicyUpdate, nil)
 	d.InformerManager.ForResource(propagationPolicyGVR, policyHandler)
 	d.propagationPolicyLister = d.InformerManager.Lister(propagationPolicyGVR)
 
@@ -148,7 +148,7 @@ func (d *ResourceDetector) Start(ctx context.Context) error {
 		Version:  policyv1alpha1.GroupVersion.Version,
 		Resource: "clusterpropagationpolicies",
 	}
-	clusterPolicyHandler := fedinformer.NewHandlerOnEvents(d.OnClusterPropagationPolicyAdd, d.OnClusterPropagationPolicyUpdate, d.OnClusterPropagationPolicyDelete)
+	clusterPolicyHandler := fedinformer.NewHandlerOnEvents(d.OnClusterPropagationPolicyAdd, d.OnClusterPropagationPolicyUpdate, nil)
 	d.InformerManager.ForResource(clusterPropagationPolicyGVR, clusterPolicyHandler)
 	d.clusterPropagationPolicyLister = d.InformerManager.Lister(clusterPropagationPolicyGVR)
 
@@ -377,6 +377,11 @@ func (d *ResourceDetector) LookForMatchedPolicy(object *unstructured.Unstructure
 			klog.Errorf("Failed to convert PropagationPolicy from unstructured object: %v", err)
 			return nil, err
 		}
+
+		if !policy.DeletionTimestamp.IsZero() {
+			klog.V(4).Infof("Propagation policy(%s/%s) cannot match any resource template because it's being deleted.", policy.Namespace, policy.Name)
+			continue
+		}
 		policyList = append(policyList, policy)
 	}
 
@@ -402,6 +407,11 @@ func (d *ResourceDetector) LookForMatchedClusterPolicy(object *unstructured.Unst
 		if err = helper.ConvertToTypedObject(policyObjects[index], policy); err != nil {
 			klog.Errorf("Failed to convert ClusterPropagationPolicy from unstructured object: %v", err)
 			return nil, err
+		}
+
+		if !policy.DeletionTimestamp.IsZero() {
+			klog.V(4).Infof("Cluster propagation policy(%s) cannot match any resource template because it's being deleted.", policy.Name)
+			continue
 		}
 		policyList = append(policyList, policy)
 	}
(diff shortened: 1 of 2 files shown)


pkg/detector/detector.go
chaunceyjiang, line 1: The webhook can only `mutate` newly created PPs, and cannot `mutate` PPs that have already been created. If the user upgrades from an old version to a new version, there will definitely be existing PPs. So I think you should also be compatible with this upgrade behavior.

whitewindmills: Nice find! Thanks~

RainbowMango: Good point!

pkg/detector/detector.go
```diff
@@ -1007,19 +1006,39 @@ func (d *ResourceDetector) ReconcilePropagationPolicy(key util.QueueKey) error {
...
+		if controllerutil.RemoveFinalizer(propagationObject, util.PropagationPolicyControllerFinalizer) {
+			if err = d.Client.Update(context.TODO(), propagationObject); err != nil {
+				klog.Errorf("Failed to remove finalizer for PropagationPolicy(%s), err: %v", ckey.NamespaceKey(), err)
+				return err
+			}
+		}
+		return nil
+	}
+
+	klog.Infof("PropagationPolicy(%s) has been added.", ckey.NamespaceKey())
```
RainbowMango, line 1035: This log is not so precise because policy updates also arise in this log, but obviously not an `added` event.

pkg/detector/detector.go
```diff
@@ -1007,19 +1006,39 @@ func (d *ResourceDetector) ReconcilePropagationPolicy(key util.QueueKey) error {
...
 
-	klog.Infof("PropagationPolicy(%s) has been added.", ckey.NamespaceKey())
 	propagationObject := &policyv1alpha1.PropagationPolicy{}
 	if err = helper.ConvertToTypedObject(unstructuredObj, propagationObject); err != nil {
 		klog.Errorf("Failed to convert PropagationPolicy(%s) from unstructured object: %v", ckey.NamespaceKey(), err)
 		return err
 	}
+
+	if !propagationObject.DeletionTimestamp.IsZero() {
+		klog.Infof("PropagationPolicy(%s) has been deleted.", ckey.NamespaceKey())
```
RainbowMango, line 1022:
```suggestion
		klog.Infof("PropagationPolicy(%s) is being delted.", ckey.NamespaceKey())
```

pkg/detector/detector.go
```diff
@@ -377,6 +377,11 @@ func (d *ResourceDetector) LookForMatchedPolicy(object *unstructured.Unstructure
 			klog.Errorf("Failed to convert PropagationPolicy from unstructured object: %v", err)
 			return nil, err
 		}
+
+		if !policy.DeletionTimestamp.IsZero() {
+			klog.V(2).Infof("Propagation policy(%s/%s) cannot match any resource template because it's being deleted.", policy.Namespace, policy.Name)
+			continue
+		}
```
RainbowMango, line 384: Impressive!  I didn't realize this part, but it makes sense. Thanks.

RainbowMango: But ,I think klog.V(`4`) would be enough.

pkg/detector/detector.go
RainbowMango: Update this log or just remove it are both OK to me.

RainbowMango (commented):
Good Job!

pkg/detector/detector.go
```diff
@@ -1117,20 +1125,40 @@ func (d *ResourceDetector) ReconcileClusterPropagationPolicy(key util.QueueKey)
...
+		return nil
+	}
+
+	klog.Infof("ClusterPropagationPolicy(%s) has been added.", ckey.NamespaceKey())
+	if controllerutil.AddFinalizer(propagationObject, util.ClusterPropagationPolicyControllerFinalizer) {
+		if err = d.Client.Update(context.TODO(), propagationObject); err != nil {
+			klog.Errorf("Failed to add finalizer for ClusterPropagationPolicy(%s), err: %v", ckey.NamespaceKey(), err)
+			return err
+		}
```
chaunceyjiang, line 1160: The update will trigger the PP update again, does this mean that `HandleClusterPropagationPolicyCreationOrUpdate` will be executed twice .

RainbowMango: nice catch!
How about just return after the update, that means in the first reconcile loop, we focus on adding the finalizer and sync by the second reconcile loop? 
like:
```diff
diff --git a/pkg/detector/detector.go b/pkg/detector/detector.go
index 893f9f611..5bbc2b2fc 100644
--- a/pkg/detector/detector.go
+++ b/pkg/detector/detector.go
@@ -1158,6 +1158,7 @@ func (d *ResourceDetector) ReconcileClusterPropagationPolicy(key util.QueueKey)
                        klog.Errorf("Failed to add finalizer for ClusterPropagationPolicy(%s), err: %v", ckey.NamespaceKey(), err)
                        return err
                }
+               return nil
        }
        return d.HandleClusterPropagationPolicyCreationOrUpdate(propagationObject)
 }
```

pkg/detector/detector.go
```diff
@@ -1007,19 +1006,39 @@ func (d *ResourceDetector) ReconcilePropagationPolicy(key util.QueueKey) error {
...
+			if err = d.Client.Update(context.TODO(), propagationObject); err != nil {
+				klog.Errorf("Failed to remove finalizer for PropagationPolicy(%s), err: %v", ckey.NamespaceKey(), err)
+				return err
+			}
+		}
+		return nil
+	}
+
+	klog.Infof("PropagationPolicy(%s) has been added.", ckey.NamespaceKey())
+	if controllerutil.AddFinalizer(propagationObject, util.PropagationPolicyControllerFinalizer) {
```
XiShanYongYe-Chang, line 1036: Can the logic of the new finalizer also be added to `karmada-webhook`? I understand that the code here is for compatibility and can be removed in the next version. 

So according to the [comments](https://github.com/karmada-io/karmada/pull/4836#discussion_r1568177480), the policy is always enqueued twice when it is first created, which is not very elegant. When there are too many policies waiting in the queue, the interval between the two times will be large.

XiShanYongYe-Chang: @RainbowMango @chaunceyjiang

RainbowMango: > Can the logic of the new finalizer also be added to karmada-webhook?

Yes, we should, please do it at another PR. (split another task?)
But for now, I don't think reconciling twice is a big problem, because the first reconcile is cheap.

pkg/detector/detector.go
whitewindmills: +1

pkg/detector/detector.go
whitewindmills: > Yes, we should, please do it at another PR. (split another task?)

yeah, maybe we don't need add it to karmada-webhook for now, otherwise it would appear redundant. We can completely transplant it to karmada-webhook when we remove this part of the logic in the future. I'll add some comments for it!

RainbowMango:
LGTM
@XiShanYongYe-Chang @chaunceyjiang 
Do you guys have any further comments?

XiShanYongYe-Chang (commented):
Thanks
/lgtm

chaunceyjiang (commented):
Thanks~

/lgtm

RainbowMango (approved):
/approve

Issue #4688, opened 2024-03-08
Ability to suspend work


a7i:
**What would you like to be added**:
Ability to suspend work to ensure that changes are not being reconciled.

**Why is this needed**:

User Story 1: As a cluster admin, we may get stuck in a reconciliation loop where karmada controller-manager will update a resource but due to some unknown reason (perhaps a controller on the member cluster) the resource is reverted. The operator can decide to suspend work while debugging the issue.

User Story 2: We're using Karmada to migrate workload from one blue cluster to a green cluster. Once we move a workload, we want to suspend any updates from blue to green until we cut over to green.

### Persona
Cluster Admin who is oncall and has permission to modify karmada resources

### Implementation Details
There are three ways that we can go about this:

#### 1.  Expose `suspend` on `Work` CRD
This is the simplest approach but requires the Cluster Admin to identify `Work` in the `karmada-es-${cluster}` namespace before patching them. Once the field is set, the controller will get out early. In the code here https://github.com/karmada-io/karmada/blob/master/pkg/controllers/status/work_status_controller.go#L93-L96, we would add the following:

```go
	if work.Susped {
		return controllerruntime.Result{}, nil
	}
```

Effort: Low

#### 2.  Expose `suspend` on `ResourceBinding` CRD
This requires the `suspend` field on `ResourceBinding` which will also get updated to `Work`. The Cluster Admin will have to identify the `ResourceBinding` in the workload namespace before patching them. The name of resource binding is more predictable than `Work`.

Effort: Medium

#### 3.  Expose `suspend` on `PropagationPolicy` and `ClusterPropagationPolicy` CRD
The Cluster Admin will have to know the PP or CPP with the highest priority and decide to suspend them. Changes to`suspend` will have to get updated to `ResourceBinding` and `Work`.

Effort: X-Large

a7i:
Created a PR to get feedback: https://github.com/karmada-io/karmada/pull/4689

chaunceyjiang:
Related to it https://github.com/karmada-io/karmada/pull/1859

RainbowMango:
> User Story 1: As a cluster admin, we may get stuck in a reconciliation loop where karmada controller-manager will update a resource but due to some unknown reason (perhaps a controller on the member cluster) the resource is reverted. The operator can decide to suspend work while debugging the issue.

+1 on this user story.
This feature would be helpful for debugging. 

[...]

PR #4838, opened 2024-04-17
feat: cluster-level resource propagation pause and resume capabilities

a7i:
**What type of PR is this?**

/kind feature

Ability to suspend work to ensure that changes are not being reconciled.

Fixes https://github.com/karmada-io/karmada/issues/4688

```release-note
cluster-level resource propagation pause and resume capabilities
```


pkg/apis/policy/v1alpha1/propagation_types.go
```diff
@@ -173,6 +173,13 @@ type PropagationSpec struct {
 	// +kubebuilder:validation:Enum=Lazy
 	// +optional
 	ActivationPreference ActivationPreference `json:"activationPreference,omitempty"`
+
+	// Suspend tells the controller to suspend subsequent executions.
```
XiShanYongYe-Chang, line 177: Whether the suspension operation is transparently transferred to the `work` object or the policy execution is suspended.

a7i: The suspension operation is transparently transferred to the `work` object.

Is this better?
```suggestion
	// Suspend will pause all work objects to pause propagation to member clusters.
```

XiShanYongYe-Chang: I wonder if what really needs to be paused is the propagation of `work`. Pauses on pp and rb are to facilitate the transfer of user intent to `work`.

XiShanYongYe-Chang: Please ignore my question above. I see that's how you've implemented it now.

pkg/apis/policy/v1alpha1/propagation_types.go
```diff
@@ -173,6 +173,13 @@ type PropagationSpec struct {
 	// +kubebuilder:validation:Enum=Lazy
 	// +optional
 	ActivationPreference ActivationPreference `json:"activationPreference,omitempty"`
+
+	// Suspend will instruct all work objects to pause propagation to member clusters.
+	// Defaults to false.
+	//
+	// +kubebuilder:default=false
+	// +optional
+	Suspend bool `json:"suspend,omitempty"`
```
XiShanYongYe-Chang, line 182: Do we need to consider setting pause fields for different clusters so that different pause policies can be set for different work of the same resource?

/cc @CharlesQQ @RainbowMango

a7i:
@XiShanYongYe-Chang  What's your feeling on this PR? Should I rebase and keep it up to date or do you not see this as a feature?

XiShanYongYe-Chang:
I agree with this feature. I'm currently working on this area to see if karmada can provide users with faster capabilities based on this. I'll sync when I make progress. Thanks.

XiShanYongYe-Chang:
Hi @a7i, we discussed how to design the API to provide users with the resources deletion policy in https://github.com/karmada-io/karmada/pull/4788#discussion_r1582008275. This is similar to Work suspension. They both describe the synchronization behavior of resource templates in member clusters, and can be resolved in [three ways](https://github.com/karmada-io/karmada/pull/4788#issuecomment-2111522122). Therefore, I'd like to know your opinions on how to design the API. Thanks a lot.

a7i:
@XiShanYongYe-Chang any updates? it sounds like annotations are preferred over fields on Spec ([ref](https://github.com/karmada-io/karmada/pull/4788#issuecomment-2127211394))?

XiShanYongYe-Chang:
That's it for now, but there's no discussion about the suspension of the work propagation. I'll continue to analyze this feature starting with the next release (i.e. next month).

RainbowMango (commented):
/assign

RainbowMango:
Hi @a7i Just an update on this PR.
This feature starts with a simple request that wants the Work can be paused, and then a few more use cases are added that this feature might be used in rollout scenarios, I think #5118 addresses all these scenarios. 
I will focus on the proposal first and then get back on this after we make the decision there.

I think this feature will be included in the coming v1.11 release(by the end of Aug), does that meet your schedule?

a7i:
That sounds good @RainbowMango 

I would be interested in contributing to the feature, if it's easy to split up.

a7i:
@XiShanYongYe-Chang and @RainbowMango would appreciate some early feedback to make sure I'm on the right path. Currently working on remaining items:
- add the condition to status in work status controller
- add more e2e from the proposal
- add webhook validation (happy to carve this out to a separate PR if you suggest)

anything else missing?

XiShanYongYe-Chang:
Hi @a7i thanks for your hard work. I think we can take small steps and submit PRs for each of the items you listed.

a7i:
Yes it is ready, excluding the items listed above

pkg/controllers/status/work_status_controller.go
```diff
@@ -95,6 +96,11 @@ func (c *WorkStatusController) Reconcile(ctx context.Context, req controllerrunt
 		return controllerruntime.Result{}, nil
 	}
 
+	if ptr.Deref(work.Spec.SuspendDispatching, false) {
```
XiShanYongYe-Chang, line 99: If we skip here, status collection will also be suspended. The expected effect should be to not affect state collection.

a7i: I see. So for now, we don't do anything in this controller? Obviously in a follow-up PR, I would need to add the status condition.

pkg/controllers/execution/execution_controller.go
```diff
@@ -108,6 +109,11 @@ func (c *Controller) Reconcile(ctx context.Context, req controllerruntime.Reques
 		return c.removeFinalizer(work)
 	}
 
+	if ptr.Deref(work.Spec.SuspendDispatching, false) {
```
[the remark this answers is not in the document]
XiShanYongYe-Chang: What I'm thinking is that all downstream actions will be paused, including resource creation and deletion. When the pause is resumed, the deletion will be executed.

pkg/controllers/status/work_status_controller.go
XiShanYongYe-Chang: Maybe we need to skip the update operation:

https://github.com/karmada-io/karmada/blob/c38f1699166b6d6664cfaa989ca546ccd4a6183e/pkg/controllers/status/work_status_controller.go#L243

RainbowMango (commented):
LGTM
for the API part.
Please assign it to me for approval if no other comments.

XiShanYongYe-Chang (commented):
Thanks~
/lgtm
/cc @whitewindmills @CharlesQQ @RainbowMango

a7i:
Created a PR for webhook validation that is rebased on top of this

Going to create another PR for work status condition and rebase on top of this

whitewindmills:
I'll take a look ASAP.

RainbowMango (approved):
/approve
It looks good to me. 
@whitewindmills Let's move this forward, we still have time for any further comments.

PR #4852, opened 2024-04-20
Remove print in test

khanhtc1202:
**What type of PR is this?**

/kind failing-test

This `fmt.Print` causes an unexpected line in stdout for test run output JSON, which causes an error for future possible go test tools. And we don't use it anyway, so I think it's better to remove it.

```console
{"Time":"2024-04-19T23:11:03.422137+07:00","Action":"output","Package":"command-line-arguments","Test":"TestGetMinTolerationTime/no_noExecuteTaints","Output":"-1ns"}
```

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/util/helper/taint_test.go
+++ b/pkg/util/helper/taint_test.go
@@ -17,7 +17,6 @@ limitations under the License.
 package helper
 
 import (
-	"fmt"
 	"reflect"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


RainbowMango (approved):
/lgtm
/approve

I also realized this print code might be just for debuging, agree that this should be removed.

Issue #4840, opened 2024-04-17
Introduce a mechanism to actively trigger rescheduling


chaosi-zju:
**What would you like to be added**:

Introducing a mechanism of actively triggering rescheduling. 

**Why is this needed**:

Proposal of introducing a rebalance mechanism to actively trigger rescheduling of resource.

Assuming the user has propagated the workloads to member clusters, in some scenarios the current replicas distribution
is not the most expected, such as:

* replicas migrated due to cluster failover, while now cluster recovered.
* replicas migrated due to application-level failover, while now each cluster has sufficient resources to run the replicas.
* as for `Aggregated` schedule strategy, replicas were initially distributed across multiple clusters due to resource
  constraints, but now one cluster is enough to accommodate all replicas.

Therefore, the user desires for an approach to trigger rescheduling so that the replicas distribution can do a rebalance.

***

### The issue is created to track all related activities：

**Proposal：** 
 * - [x] #4698

**Implementation：** 
 * - [x] Scheduler: #4848
 * - [x] WorkloadRebalancer：#4841
 * - [x] ~Webhook: #4859~
 * - [x] bugfix: #4875
 * - [x] Support update: #4860
 * - [x] Support auto delete: #4894

**UT**
 * - [x] #4876

**E2E Test：**
 * - [x] #4869
 * - [x] #4961 

**Doc of Best Practice：**
 * - [x] **en** tutorial: https://github.com/karmada-io/website/pull/608
 * - [x] **zh** tutorial: https://github.com/karmada-io/website/pull/627
 * - [x] **en** user guide: https://github.com/karmada-io/website/pull/612
 * - [x] **zh** user guide: https://github.com/karmada-io/website/pull/628

RainbowMango:
We need a document for this feature that is missing from this umbrella issue.

RainbowMango:
FYI
Add two tasks for the zh documents.

RainbowMango:
/close
since all tasks have been addressed.

PR #4860, opened 2024-04-23
feat: support update event in WorkloadRebalancer

chaosi-zju:
**DO NOT MERGE** until the following PR have been merged: 

* #4875

**What type of PR is this?**

/kind feature

feat: support update event in WorkloadRebalancer

When spec filed of WorkloadRebalancer updated, we shall refresh the workload list in status.observedWorkloads:

  1. a new workload added to spec list, just add it into status list too and do the rebalance.
  2. a workload deleted from previous spec list, keep it in status list if already success, and remove it if not.
  3. a workload is modified, just regard it as deleted an old one and inserted a new one.
  4. just list order is disrupted, no additional action.

Fixes  part of #4840

```release-note
feat: support update event in WorkloadRebalancer
```

### DIFF
--- a/pkg/controllers/workloadrebalancer/workloadrebalancer_controller.go
+++ b/pkg/controllers/workloadrebalancer/workloadrebalancer_controller.go
@@ -19,9 +19,12 @@ package workloadrebalancer
 import (
 	"context"
 	"fmt"
+	"reflect"
+	"sort"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
 	controllerruntime "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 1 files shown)


pkg/controllers/workloadrebalancer/workloadrebalancer_controller.go
```diff
@@ -93,15 +99,46 @@ func (c *RebalancerController) Reconcile(ctx context.Context, req controllerrunt
...
+			Workload: workload,
 		})
 	}
-	return &appsv1alpha1.WorkloadRebalancerStatus{
-		ObservedWorkloads: resourceList,
+	return &appsv1alpha1.WorkloadRebalancerStatus{ObservedWorkloads: observedWorkloads}
+}
+
+// When spec filed of WorkloadRebalancer updated, we shall refresh the workload list in status.observedWorkloads:
+//  1. a new workload added to spec list, just add it into status list too and do the rebalance.
```
RainbowMango, line 112:
```suggestion
//  1. a new workload added to spec list, just add it into status list and do the rebalance.
```

pkg/controllers/workloadrebalancer/workloadrebalancer_controller.go
```diff
@@ -110,6 +147,8 @@ func (c *RebalancerController) doWorkloadRebalance(ctx context.Context, rebalanc
 	newStatus = rebalancer.Status.DeepCopy()
```
RainbowMango, line 147: Seems this deep copy is redundant as it will be overwritten by the following `buildWorkloadRebalancerStatus` or `syncWorkloadsFromSpecToStatus`.

pkg/controllers/workloadrebalancer/workloadrebalancer_controller.go
```diff
@@ -110,6 +147,8 @@ func (c *RebalancerController) doWorkloadRebalance(ctx context.Context, rebalanc
...
 	if len(newStatus.ObservedWorkloads) == 0 {
 		newStatus = c.buildWorkloadRebalancerStatus(rebalancer)
+	} else {
+		newStatus = c.syncWorkloadsFromSpecToStatus(rebalancer)
```
RainbowMango, line 151: Maybe we don't need to check if it is first reconcile, just use `newStatus = c.syncWorkloadsFromSpecToStatus(rebalancer)` to gather observed items?

RainbowMango:
Please rebase this PR if there is another push.

pkg/controllers/workloadrebalancer/workloadrebalancer_controller.go
chaosi-zju: you are right, just use `syncWorkloadsFromSpecToStatus` is enough.

however, in `syncWorkloadsFromSpecToStatus`, the resource list in `status` is disordered (previously is mainly following the input order). I prefer to sort them in dict order when just using `syncWorkloadsFromSpecToStatus`, as you can see in my latest push.

RainbowMango: sort +1.

RainbowMango (approved):
/lgtm
/approve

PR #4882, opened 2024-04-28
Use webhook to set associated labels and annotations of work

whitewindmills:
**What type of PR is this?**
/kind cleanup

This PR mainly migrates the adding action of work-related labels and annotations to webhook. Let's focus on these labels and annotations in the work mutating, which make it easier to maintain.

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/certificate/cert_rotation_controller.go
+++ b/pkg/controllers/certificate/cert_rotation_controller.go
@@ -239,9 +239,6 @@ func (c *CertRotationController) createCSRInControlPlane(clusterName string, pri
 	certificateSigningRequest := &certificatesv1.CertificateSigningRequest{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: csrName,
-			Labels: map[string]string{
-				util.ManagedByKarmadaLabel: util.ManagedByKarmadaLabelValue,
-			},
 		},
 		Spec: certificatesv1.CertificateSigningRequestSpec{
 			Request:           csrData,
--- a/pkg/controllers/status/work_status_controller.go
+++ b/pkg/controllers/status/work_status_controller.go
@@ -231,7 +231,6 @@ func (c *WorkStatusController) syncWorkStatus(key util.QueueKey) error {
 		return err
 	}
 
-	util.MergeLabel(desiredObj, workv1alpha2.WorkPermanentIDLabel, util.GetLabelValue(workObject.Labels, workv1alpha2.WorkPermanentIDLabel))
 	// we should check if the observed status is consistent with the declaration to prevent accidental changes made
 	// in member clusters.
 	needUpdate, err := c.ObjectWatcher.NeedsUpdate(clusterName, desiredObj, observedObj)
--- a/pkg/webhook/work/mutating.go
+++ b/pkg/webhook/work/mutating.go
@@ -67,6 +67,11 @@ func (a *MutatingAdmission) Handle(_ context.Context, req admission.Request) adm
 			return admission.Errored(http.StatusInternalServerError, err)
 		}
 
+		// Skip label/annotate the workload of Work that is not intended to be propagated.
+		if work.Labels[util.PropagationInstruction] != util.PropagationInstructionSuppressed {
+			setLabelsAndAnnotationsForWorkload(workloadObj, work)
+		}
+
 		workloadJSON, err := workloadObj.MarshalJSON()
 		if err != nil {
 			klog.Errorf("Failed to marshal workload of the work(%s/%s), err: %s", work.Namespace, work.Name, err)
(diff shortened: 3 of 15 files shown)


whitewindmills:
@XiShanYongYe-Chang @RainbowMango please help review it.

XiShanYongYe-Chang:
Thanks~
/assign

pkg/webhook/work/mutating.go
```diff
@@ -64,6 +66,8 @@ func (a *MutatingAdmission) Handle(_ context.Context, req admission.Request) adm
 			return admission.Errored(http.StatusInternalServerError, err)
 		}
 
+		setAssociatedLabelsAndAnnotations(workloadObj, work)
```
XiShanYongYe-Chang, line 69: Should we skip this operation for resources collected from member clusters, such as EndpointSlice?

whitewindmills: why it should be skipped?

XiShanYongYe-Chang: Do they use these labels and annotations?

whitewindmills: maybe not. but the behavior now is almost the same as before, and we can treat it as standard behavior, I think it's ok.

XiShanYongYe-Chang: I think we can distinguish between downward synchronization and upward collection. These are two modes. It's better to wait until the need arises in the future.

XiShanYongYe-Chang: Currently, we use the label `propagation.karmada.io/instruction` to identify the EndpointSlice collected from the cluster.

whitewindmills: thanks

pkg/webhook/work/mutating.go
```diff
@@ -87,3 +92,22 @@ func (a *MutatingAdmission) Handle(_ context.Context, req admission.Request) adm
...
+		workv1alpha2.WorkNameAnnotation:            work.GetName(),
+	}))
+	if conflictResolution, ok := work.Annotations[workv1alpha2.ResourceConflictResolutionAnnotation]; ok {
+		util.MergeAnnotation(workload, workv1alpha2.ResourceConflictResolutionAnnotation, conflictResolution)
+	}
+	util.RecordManagedAnnotations(workload)
+
+	workload.SetLabels(util.DedupeAndMergeLabels(workload.GetLabels(), map[string]string{
+		util.ManagedByKarmadaLabel:        util.ManagedByKarmadaLabelValue,
+		workv1alpha2.WorkPermanentIDLabel: work.Labels[workv1alpha2.WorkPermanentIDLabel],
```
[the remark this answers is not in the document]
whitewindmills: can we adjust it in another PR?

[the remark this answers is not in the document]
XiShanYongYe-Chang: Sure!

pkg/controllers/certificate/cert_rotation_controller.go
```diff
@@ -239,9 +239,6 @@ func (c *CertRotationController) createCSRInControlPlane(clusterName string, pri
 	certificateSigningRequest := &certificatesv1.CertificateSigningRequest{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: csrName,
-			Labels: map[string]string{
-				util.ManagedByKarmadaLabel: util.ManagedByKarmadaLabelValue,
```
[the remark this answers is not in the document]
whitewindmills: cause it should not be here, the label `karmada.io/managed` indicates these resources of member clusters are managed by Karmada. am I right?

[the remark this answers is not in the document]
XiShanYongYe-Chang: It looks like this, the resource is created in the control plane

pkg/webhook/work/mutating.go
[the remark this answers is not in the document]
chaunceyjiang: I have a question: Is there a situation where 'work' does not have a PermanentID? That is, only after you assign an empty PermanentID to the workload, will 'work' have a PermanentID.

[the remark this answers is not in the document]
whitewindmills: > Is there a situation where 'work' does not have a PermanentID?

never happen. all current work is created through this function. this part of the logic will be completely migrated to webhook later.
https://github.com/karmada-io/karmada/blob/b93d71ec5aa56b709664a53a4c42a7373b9e6d5d/pkg/util/helper/work.go#L82-L85

pkg/controllers/status/work_status_controller.go
```diff
@@ -231,7 +231,6 @@ func (c *WorkStatusController) syncWorkStatus(key util.QueueKey) error {
 		return err
 	}
 
-	util.MergeLabel(desiredObj, workv1alpha2.WorkPermanentIDLabel, util.GetLabelValue(workObject.Labels, workv1alpha2.WorkPermanentIDLabel))
```
chaunceyjiang, line 234: Here can it be removed?

When karmada has an `agent`, it should be the `agent` that adds these labels, right?

whitewindmills: we add it in work mutating-webhook now, so it can be safely removed.

pkg/webhook/work/mutating.go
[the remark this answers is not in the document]
chaunceyjiang: If a user upgrades from a lower version to the latest version, there is no `PermanentID` in the `work` of the lower version. A `PermanentID` will only be available after updating the `work`. Will this update operation trigger the webhook again?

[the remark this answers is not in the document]
whitewindmills: > Will this update operation trigger the webhook again?

of course

pkg/controllers/status/work_status_controller.go
chaunceyjiang: What I mean is, this was originally handled by the karmada-agent, but it has been moved to be managed by the control panel. I'm not sure if this is appropriate.

chaunceyjiang (commented):
Nice work!

Thanks~

/lgtm

pkg/webhook/work/mutating.go
```diff
@@ -87,3 +92,22 @@ func (a *MutatingAdmission) Handle(_ context.Context, req admission.Request) adm
 
 	return admission.PatchResponseFromRaw(req.Object.Raw, marshaledBytes)
 }
+
+// setAssociatedLabelsAndAnnotations sets the associated work object labels and annotations.
+func setAssociatedLabelsAndAnnotations(workload *unstructured.Unstructured, work *workv1alpha1.Work) {
```
RainbowMango, line 97:
```suggestion
func setLabelsAndAnnotationsForWorkload(workload *unstructured.Unstructured, work *workv1alpha1.Work) {
```
How about renaming it? As this function is dedicated to `workload` instead of `work`.
Please update the comments accordingly if it makes sense.

pkg/webhook/work/mutating.go
```diff
@@ -67,6 +67,11 @@ func (a *MutatingAdmission) Handle(_ context.Context, req admission.Request) adm
 			return admission.Errored(http.StatusInternalServerError, err)
 		}
 
+		// We only set these labels and annotations on resources propagated down.
```
RainbowMango, line 70: We don't usually say `propagate down` or `propagate up`, so proposing the comments like:
```
// Skip label/annotate the workload of Work that is not intended to be propagated.
```

RainbowMango (commented):
Good Job!
Just some nits, otherwise LGTM.

pkg/webhook/work/mutating.go
whitewindmills: make sense, thanks.

RainbowMango (approved):
/lgtm
/approve
Thanks for the quick response.

PR #4875, opened 2024-04-28
fix patching WorkloadRebalancer failed due to misuse of shared slices

chaosi-zju:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:

fix patching WorkloadRebalancer failed due to misuse of shared slices.

in previous code:

```go
// get previous status and update basing on it
newStatus = rebalancer.Status
if len(newStatus.ObservedWorkloads) == 0 {
	newStatus = c.buildWorkloadRebalancerStatus(rebalancer)
}
```

actually `newStatus` and `rebalancer.Status` shared the slice of `rebalancer.Status.ObservedWorkloads`.

if this slice didn't change from nil to non-nil or occurs expansion, modification to `newStatus` will also be applied to the `rebalancer.Status`, which will lead to `patch status` failed. 

**Which issue(s) this PR fixes**:

Fixes part of #4840

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #4894, opened 2024-04-30
support auto delete WorkloadRebalancer when time up

chaosi-zju:
**What type of PR is this?**

/kind feature

Support auto delete WorkloadRebalancer when time up:

referring to [Automatic Cleanup for Finished Jobs](https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/).

Introduces field `ttlSecondsAfterFinished` which limits the lifetime of a WorkloadRebalancer that has finished execution 
(finished execution means each target workload is finished with result of `Successful` or `Failed`).

* If this field is set, `ttlSecondsAfterFinished` after the WorkloadRebalancer finishes, it is eligible to be automatically deleted.
* If this field is unset, the WorkloadRebalancer won't be automatically deleted.
* If this field is set to zero, the WorkloadRebalancer becomes eligible to be deleted immediately after it finishes.

Considering several corner cases:

* case 1: if a new target workload was added into `WorkloadRebalancer` before `ttlSecondsAfterFinished` expired, 
  which means the finish time of the `WorkloadRebalancer` is refreshed, so the `delete` action is deferred since expire time is refreshed too.
* case 2: if `ttlSecondsAfterFinished` is modified before `ttlSecondsAfterFinished` expired, 
  `delete` action should be performed according to latest `ttlSecondsAfterFinished`.
* case 3: when we have got and checked latest `WorkloadRebalancer` object and try to delete it, 
  if a modification to `WorkloadRebalancer` occurred just right between the two time point, the previous `delete` action should be Interrupted.

Several key implementation:
* A `WorkloadRebalancer` is judged as finished should meet two requirements:
  *  all expected workloads are finished with result of `Successful` or `Failed`.
  *  introduce a new field named `ObservedGeneration` to `Status` of WorkloadRebalancer, and it should be equal to `.metadata.Generation`,
     to prevent that the WorkloadRebalancer is updated but controller hasn't in time refreshed its `Status`.
* When `WorkloadRebalancer` is `Created` or `Updated`, add it to the workqueue and calculate its expiring time, and 
  call `workqueue.AddAfter()` function to re-enqueue it once more if it hasn't expired.
* Before deleting the `WorkloadRebalancer`, do a final sanity check. Use the latest `WorkloadRebalancer` directly 
  fetched from api server to see if the TTL truly expires, rather than object from lister cache.
* When deleting the `WorkloadRebalancer`, it is needed to confirm that the `resourceVersion` of the deleted object is as expected,
  to prevent from above corner case 3.

Fixes part of #4840

```release-note
support auto delete WorkloadRebalancer when time up
```

### DIFF
--- a/pkg/apis/apps/v1alpha1/workloadrebalancer_types.go
+++ b/pkg/apis/apps/v1alpha1/workloadrebalancer_types.go
@@ -57,6 +57,14 @@ type WorkloadRebalancerSpec struct {
 	// +kubebuilder:validation:MinItems=1
 	// +required
 	Workloads []ObjectReference `json:"workloads"`
+
+	// TTLSecondsAfterFinished limits the lifetime of a WorkloadRebalancer that has finished execution (means each
+	// target workload is finished with result of Successful or Failed).
+	// If this field is set, ttlSecondsAfterFinished after the WorkloadRebalancer finishes, it is eligible to be automatically deleted.
+	// If this field is unset, the WorkloadRebalancer won't be automatically deleted.
+	// If this field is set to zero, the WorkloadRebalancer becomes eligible to be deleted immediately after it finishes.
+	// +optional
+	TTLSecondsAfterFinished *int32 `json:"ttlSecondsAfterFinished,omitempty"`
 }
 
 // ObjectReference the expected resource.
@@ -85,6 +93,16 @@ type WorkloadRebalancerStatus struct {
 	// ObservedWorkloads contains information about the execution states and messages of target resources.
 	// +optional
 	ObservedWorkloads []ObservedWorkload `json:"observedWorkloads,omitempty"`
+
+	// ObservedGeneration is the generation(.metadata.generation) observed by the controller.
+	// If ObservedGeneration is less than the generation in metadata means the controller hasn't confirmed
+	// the rebalance result or hasn't done the rebalance yet.
+	// +optional
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+
+	// FinishTime represents the finish time of rebalancer.
+	// +optional
+	FinishTime *metav1.Time `json:"finishTime,omitempty"`
 }
 
 // ObservedWorkload the observed resource.
@@ -117,7 +135,7 @@ type RebalanceFailedReason string
 
 const (
 	// RebalanceObjectNotFound the resource referenced binding not found.
-	RebalanceObjectNotFound RebalanceFailedReason = "NotFound"
+	RebalanceObjectNotFound RebalanceFailedReason = "ReferencedBindingNotFound"
 )
 
 // +kubebuilder:resource:scope="Cluster"
--- a/pkg/controllers/workloadrebalancer/workloadrebalancer_controller.go
+++ b/pkg/controllers/workloadrebalancer/workloadrebalancer_controller.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"reflect"
 	"sort"
+	"time"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -51,7 +52,7 @@ type RebalancerController struct {
 // SetupWithManager creates a controller and register to controller manager.
 func (c *RebalancerController) SetupWithManager(mgr controllerruntime.Manager) error {
 	var predicateFunc = predicate.Funcs{
-		CreateFunc: func(event.CreateEvent) bool { return true },
+		CreateFunc: func(_ event.CreateEvent) bool { return true },
 		UpdateFunc: func(e event.UpdateEvent) bool {
 			oldObj := e.ObjectOld.(*appsv1alpha1.WorkloadRebalancer)
 			newObj := e.ObjectNew.(*appsv1alpha1.WorkloadRebalancer)
@@ -83,19 +84,30 @@ func (c *RebalancerController) Reconcile(ctx context.Context, req controllerrunt
 		return controllerruntime.Result{}, err
 	}
 
-	// 2. get and update referenced binding to trigger a rescheduling
-	newStatus, successNum, retryNum := c.doWorkloadRebalance(ctx, rebalancer)
-
-	// 3. update status of WorkloadRebalancer
-	if err := c.updateWorkloadRebalancerStatus(ctx, rebalancer, newStatus); err != nil {
+	// 2. list latest target workloads and trigger its rescheduling by updating its referenced binding.
+	newStatus, err := c.handleWorkloadRebalance(ctx, rebalancer)
+	if err != nil {
 		return controllerruntime.Result{}, err
 	}
-	klog.Infof("Finish handling WorkloadRebalancer (%s), %d/%d resource success in all, while %d resource need retry",
-		rebalancer.Name, successNum, len(newStatus.ObservedWorkloads), retryNum)
 
-	if retryNum > 0 {
-		return controllerruntime.Result{}, fmt.Errorf("%d resource reschedule triggered failed and need retry", retryNum)
+	if rebalancer.Status.FinishTime == nil {
+		// should never reach here.
+		klog.Errorf("finishTime shouldn't be nil, current status: %+v", rebalancer.Status)
+		return controllerruntime.Result{}, nil
 	}
+
+	// 3. when all workloads finished, judge whether the rebalancer needs cleanup.
+	rebalancer.Status = *newStatus
+	if rebalancer.Spec.TTLSecondsAfterFinished != nil {
+		remainingTTL := timeLeft(rebalancer)
+		if remainingTTL > 0 {
+			return controllerruntime.Result{RequeueAfter: remainingTTL}, nil
+		}
+		if err := c.deleteWorkloadRebalancer(ctx, rebalancer); err != nil {
+			return controllerruntime.Result{}, err
+		}
+	}
+
 	return controllerruntime.Result{}, nil
 }
 
@@ -132,27 +144,56 @@ func (c *RebalancerController) syncWorkloadsFromSpecToStatus(rebalancer *appsv1a
 	sort.Slice(observedWorkloads, func(i, j int) bool {
 		wi := observedWorkloads[i].Workload
 		wj := observedWorkloads[j].Workload
-		stri := fmt.Sprintf("%s#%s#%s#%s", wi.APIVersion, wi.Kind, wi.Namespace, wi.Name)
-		strj := fmt.Sprintf("%s#%s#%s#%s", wj.APIVersion, wj.Kind, wj.Namespace, wj.Name)
+		stri := fmt.Sprintf("%s %s %s %s", wi.APIVersion, wi.Kind, wi.Namespace, wi.Name)
+		strj := fmt.Sprintf("%s %s %s %s", wj.APIVersion, wj.Kind, wj.Namespace, wj.Name)
 		return stri < strj
 	})
 
-	return &appsv1alpha1.WorkloadRebalancerStatus{ObservedWorkloads: observedWorkloads}
+	return &appsv1alpha1.WorkloadRebalancerStatus{ObservedWorkloads: observedWorkloads, ObservedGeneration: rebalancer.Generation}
 }
 
-func (c *RebalancerController) doWorkloadRebalance(ctx context.Context, rebalancer *appsv1alpha1.WorkloadRebalancer) (
-	newStatus *appsv1alpha1.WorkloadRebalancerStatus, successNum int64, retryNum int64) {
-	// get previous status and update basing on it
+func (c *RebalancerController) handleWorkloadRebalance(ctx context.Context, rebalancer *appsv1alpha1.WorkloadRebalancer) (*appsv1alpha1.WorkloadRebalancerStatus, error) {
+	// 1. get previous status and update basing on it
+	newStatus := c.syncWorkloadsFromSpecToStatus(rebalancer)
 
-	newStatus = c.syncWorkloadsFromSpecToStatus(rebalancer)
+	// 2. trigger the rescheduling of target workloads listed in newStatus
+	newStatus, retryNum := c.triggerReschedule(ctx, rebalancer.ObjectMeta, newStatus)
 
-	successNum, retryNum = int64(0), int64(0)
+	// 3. judge whether status changed and update finishTime field if so
+	statusChanged := false
+	newStatus.FinishTime = rebalancer.Status.FinishTime
+	// compare whether newStatus equals to old status except finishTime field.
+	if !reflect.DeepEqual(rebalancer.Status, *newStatus) {
+		statusChanged = true
+		// retryNum is 0 means the latest rebalancer has finished, which should update finishTime.
+		if retryNum == 0 {
+			finishTime := metav1.Now()
+			newStatus.FinishTime = &finishTime
+		}
+	}
+
+	// 4. update status of WorkloadRebalancer if statusChanged
+	if statusChanged {
+		if err := c.updateWorkloadRebalancerStatus(ctx, rebalancer, newStatus); err != nil {
+			return newStatus, fmt.Errorf("update status failed: %w", err)
+		}
+	}
+
+	if retryNum > 0 {
+		return newStatus, fmt.Errorf("%d resource reschedule triggered failed and need retry", retryNum)
+	}
+	return newStatus, nil
+}
+
+func (c *RebalancerController) triggerReschedule(ctx context.Context, metadata metav1.ObjectMeta, newStatus *appsv1alpha1.WorkloadRebalancerStatus) (
+	*appsv1alpha1.WorkloadRebalancerStatus, int64) {
+	successNum, retryNum := int64(0), int64(0)
 	for i, resource := range newStatus.ObservedWorkloads {
 		if resource.Result == appsv1alpha1.RebalanceSuccessful {
 			successNum++
 			continue
 		}
-		if resource.Result == appsv1alpha1.RebalanceFailed && resource.Reason == appsv1alpha1.RebalanceObjectNotFound {
+		if resource.Result == appsv1alpha1.RebalanceFailed {
 			continue
 		}
 
@@ -166,8 +207,8 @@ func (c *RebalancerController) doWorkloadRebalance(ctx context.Context, rebalanc
 				continue
 			}
 			// update spec.rescheduleTriggeredAt of referenced fetchTargetRefBindings to trigger a rescheduling
-			if c.needTriggerReschedule(rebalancer.CreationTimestamp, binding.Spec.RescheduleTriggeredAt) {
-				binding.Spec.RescheduleTriggeredAt = &rebalancer.CreationTimestamp
+			if c.needTriggerReschedule(metadata.CreationTimestamp, binding.Spec.RescheduleTriggeredAt) {
+				binding.Spec.RescheduleTriggeredAt = &metadata.CreationTimestamp
 
 				if err := c.Client.Update(ctx, binding); err != nil {
 					klog.Errorf("update binding for resource %+v failed: %+v", resource.Workload, err)
@@ -184,8 +225,8 @@ func (c *RebalancerController) doWorkloadRebalance(ctx context.Context, rebalanc
 				continue
 			}
 			// update spec.rescheduleTriggeredAt of referenced clusterbinding to trigger a rescheduling
-			if c.needTriggerReschedule(rebalancer.CreationTimestamp, clusterbinding.Spec.RescheduleTriggeredAt) {
-				clusterbinding.Spec.RescheduleTriggeredAt = &rebalancer.CreationTimestamp
+			if c.needTriggerReschedule(metadata.CreationTimestamp, clusterbinding.Spec.RescheduleTriggeredAt) {
+				clusterbinding.Spec.RescheduleTriggeredAt = &metadata.CreationTimestamp
 
 				if err := c.Client.Update(ctx, clusterbinding); err != nil {
 					klog.Errorf("update cluster binding for resource %+v failed: %+v", resource.Workload, err)
@@ -196,7 +237,10 @@ func (c *RebalancerController) doWorkloadRebalance(ctx context.Context, rebalanc
 			c.recordAndCountRebalancerSuccess(&newStatus.ObservedWorkloads[i], &successNum)
 		}
 	}
-	return
+
+	klog.V(4).Infof("Finish handling WorkloadRebalancer (%s), %d/%d resource success in all, while %d resource need retry",
+		metadata.Name, successNum, len(newStatus.ObservedWorkloads), retryNum)
+	return newStatus, retryNum
 }
 
 func (c *RebalancerController) needTriggerReschedule(creationTimestamp metav1.Time, rescheduleTriggeredAt *metav1.Time) bool {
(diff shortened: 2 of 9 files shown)


chaosi-zju:
/hold

for #4875 #4860

chaosi-zju:
CC @RainbowMango can this begin to review?

pkg/apis/apps/v1alpha1/workloadrebalancer_types.go
```diff
@@ -57,6 +57,14 @@ type WorkloadRebalancerSpec struct {
...
 	// +required
 	Workloads []ObjectReference `json:"workloads"`
+
+	// ttlMinutesAfterFinished limits the lifetime of a WorkloadRebalancer that has finished execution (means each
+	// target workload is finished with result of Successful or Failed).
+	// If this field is set, ttlSecondsAfterFinished after the WorkloadRebalancer finishes, it is eligible to be automatically deleted.
+	// If this field is unset, the WorkloadRebalancer won't be automatically deleted.
+	// If this field is set to zero, the WorkloadRebalancer becomes eligible to be deleted immediately after it finishes.
+	// +optional
+	TTLMinutesAfterFinished *int32 `json:"ttlMinutesAfterFinished,omitempty"`
```
RainbowMango, line 67: Just a question, why not use `seconds` but `minutes` here?

pkg/controllers/workloadrebalancer/ttlafterfinished_worker.go
```diff
@@ -0,0 +1,141 @@
...
+		return expireAt, nil
+	}
+
+	c.ttlAfterFinishedWorker.AddAfter(client.ObjectKey{Name: r.Name}, *remainingTTL)
+	return nil, nil
+}
+
+func timeLeft(r *appsv1alpha1.WorkloadRebalancer) (*time.Duration, *time.Time, error) {
+	now := time.Now()
+	finishAt := r.Status.LastUpdateTime.Time
```
RainbowMango, line 108: Can you show me where to set the `LastUpdateTime`?

pkg/controllers/workloadrebalancer/workloadrebalancer_controller.go
```diff
@@ -35,32 +35,49 @@ import (
...
+	ttlAfterFinishedWorkerNum = 1
 )
 
 // RebalancerController is to handle a rebalance to workloads selected by WorkloadRebalancer object.
 type RebalancerController struct {
-	Client client.Client
+	Client             client.Client // used to operate WorkloadRebalancer resources from cache.
+	ControlPlaneClient client.Client // used to fetch arbitrary resources from api server.
+
+	ttlAfterFinishedWorker util.AsyncWorker
```
RainbowMango, line 54: Smells not right, can't we check and remove the object in the reconcile loop?

pkg/apis/apps/v1alpha1/workloadrebalancer_types.go
```diff
@@ -85,6 +93,16 @@ type WorkloadRebalancerStatus struct {
...
+
+	// ObservedGeneration is the generation(.metadata.generation) observed by the controller.
+	// If ObservedGeneration is less than the generation in metadata means the controller hasn't confirmed
+	// the rebalance result or hasn't done the rebalance yet.
+	// optional
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+
+	// LastUpdateTime represents the last update time of any field in WorkloadRebalancerStatus other than itself.
+	// optional
+	LastUpdateTime *metav1.Time `json:"lastUpdateTime,omitempty"`
```
RainbowMango, line 105: Do we need to update this `LastUpdateTime` during each reconcile loop?

pkg/controllers/workloadrebalancer/ttlafterfinished_worker.go
chaosi-zju: Sorry, it is my fault when my last `rebase` process, I encountered many conflict.

When I resolving the conflict, I missed the key line to update `LastUpdateTime`.

https://github.com/karmada-io/karmada/compare/7b02e4e882620de6b4f41fff6d8262c43755b669..32819f2972c056ee53a2a51ebbf0d1f6bf7e7cc9

image

I will do a quick fix.

pkg/apis/apps/v1alpha1/workloadrebalancer_types.go
chaosi-zju: when any field in `WorkloadRebalancerStatus` other than `LastUpdateTime` itself changed, we should update this field.

pkg/apis/apps/v1alpha1/workloadrebalancer_types.go
chaosi-zju: In my expected usage scenarios of `WorkloadRebalancer`, users do not need to be accurate to seconds, only to minutes.

I think `rebalance` is a key operation. In actual usge, it may take more than ten minutes to observe whether the status of the instance meets expectations after the operation. If there is an exception, it may take another ten minutes to use this resource object to locate the problem. So the deletion time must be in minutes.

chaosi-zju: besides, since the usage scenario is `minites` level, `minutes` may be more readable than `seconds`.

pkg/controllers/workloadrebalancer/workloadrebalancer_controller.go
chaosi-zju: I didn't understand, can you give an example?

pkg/apis/apps/v1alpha1/workloadrebalancer_types.go
chaosi-zju: Updated to `finishTime`.

pkg/controllers/workloadrebalancer/workloadrebalancer_controller.go
chaosi-zju: ready to continue to review~ @RainbowMango

pkg/apis/apps/v1alpha1/workloadrebalancer_types.go
```diff
@@ -57,6 +57,14 @@ type WorkloadRebalancerSpec struct {
 	// +kubebuilder:validation:MinItems=1
 	// +required
 	Workloads []ObjectReference `json:"workloads"`
+
+	// ttlSecondsAfterFinished limits the lifetime of a WorkloadRebalancer that has finished execution (means each
```
RainbowMango, line 61:
```suggestion
	// TTLSecondsAfterFinished limits the lifetime of a WorkloadRebalancer that has finished execution (means each
```

pkg/apis/apps/v1alpha1/workloadrebalancer_types.go
```diff
@@ -85,6 +93,16 @@ type WorkloadRebalancerStatus struct {
 	// ObservedWorkloads contains information about the execution states and messages of target resources.
 	// +optional
 	ObservedWorkloads []ObservedWorkload `json:"observedWorkloads,omitempty"`
+
+	// ObservedGeneration is the generation(.metadata.generation) observed by the controller.
+	// If ObservedGeneration is less than the generation in metadata means the controller hasn't confirmed
+	// the rebalance result or hasn't done the rebalance yet.
+	// optional
```
RainbowMango, line 100:
```suggestion
	// +optional
```

pkg/apis/apps/v1alpha1/workloadrebalancer_types.go
```diff
@@ -85,6 +93,16 @@ type WorkloadRebalancerStatus struct {
...
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+
+	// FinishTime represents the finish time of rebalancer.
+	// optional
```
RainbowMango, line 104:
```suggestion
	// +optional
```

RainbowMango (commented):
Don't have enough time to finish the review, just covered the API part this time.
Will get back on this next week.

pkg/controllers/workloadrebalancer/workloadrebalancer_controller.go
```diff
@@ -35,32 +35,49 @@ import (
...
 	ControllerName = "workload-rebalancer"
+
+	ttlAfterFinishedWorkerNum = 1
 )
 
 // RebalancerController is to handle a rebalance to workloads selected by WorkloadRebalancer object.
 type RebalancerController struct {
-	Client client.Client
+	Client             client.Client // used to operate WorkloadRebalancer resources from cache.
+	ControlPlaneClient client.Client // used to fetch arbitrary resources from api server.
```
RainbowMango, line 52: Here looks weird to have two control plane clients, one is used to fetch resources from the cache, and the other bypasses the cache.
Do we need to do that in such a way?

By the way, the comments are not precise enough, as the client(comes from controller-manager framework) not only operates resources against the cache.

pkg/controllers/workloadrebalancer/ttlafterfinished_worker.go
```diff
@@ -0,0 +1,125 @@
...
+	// 3. The WorkloadRebalancer's TTL is assumed to have expired, but the WorkloadRebalancer TTL might be stale.
+	// Before deleting the WorkloadRebalancer, do a final sanity check.
+	// If TTL is modified before we do this check, we cannot be sure if the TTL truly expires.
+	fresh := &appsv1alpha1.WorkloadRebalancer{}
+	if err := c.ControlPlaneClient.Get(context.TODO(), objKey, fresh); err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil
+		}
+		return err
+	}
```
RainbowMango, line 65: I can understand that you don't 100% trust the data from the cache, so you fetch the object directly from the api-server.

But the object can be any time, consider the situation that the updation happened just after you fetch the object, and before deleting, the updation will be dropped, right?

pkg/controllers/workloadrebalancer/workloadrebalancer_controller.go
```diff
@@ -35,32 +35,49 @@ import (
...
 		DeleteFunc:  func(event.DeleteEvent) bool { return false },
 		GenericFunc: func(event.GenericEvent) bool { return false },
 	}
 
+	ttlAfterFinishedWorkerOptions := util.Options{
+		Name:          "ttl-after-finished-worker",
+		ReconcileFunc: c.deleteExpiredRebalancer,
+	}
+	c.ttlAfterFinishedWorker = util.NewAsyncWorker(ttlAfterFinishedWorkerOptions)
+	c.ttlAfterFinishedWorker.Run(ttlAfterFinishedWorkerNum, context.Background().Done())
```
RainbowMango, line 79: I am still not sure if we should introduce another queue to check the TTL yet, just feel a little bit heavy. I will look at the Job controller from Kubernetes, and then get back on this.

pkg/controllers/workloadrebalancer/ttlafterfinished_worker.go
chaosi-zju: > consider the situation that the updation happened just after you fetch the object, and before deleting, the updation will be dropped, right?

**No, it will not be droped, the delete action will wait for updation reconciled.**

 a corner, supposing the updation is modify `ttlSecondsAfterFinished` from 10s to 100s, the real delete action will defer to 100s after the latest updation.

by the way, this double check is also refer to job controller https://github.com/kubernetes/kubernetes/blob/a31030543c47aac36cf323b885cfb6d8b0a2435f/pkg/controller/ttlafterfinished/ttlafterfinished_controller.go#L206-L241

pkg/controllers/workloadrebalancer/workloadrebalancer_controller.go
chaosi-zju: >  the comments are not precise enough, as the client(comes from controller-manager framework) not only operates resources against the cache.

got it, thanks.

> Here looks weird to have two control plane clients, ... Do we need to do that in such a way?

this is the same problem as next comment below.

RainbowMango:
```suggestion
	Client             client.Client
	ClientWithOutCache versioned.Interface
```
How about using the raw client instead of the one from the controller-runtime framework?

cmd/controller-manager/app/controllermanager.go
```diff
@@ -711,7 +711,8 @@ func startRemedyController(ctx controllerscontext.Context) (enabled bool, err er
 
 func startWorkloadRebalancerController(ctx controllerscontext.Context) (enabled bool, err error) {
 	workloadRebalancer := workloadrebalancer.RebalancerController{
-		Client: ctx.Mgr.GetClient(),
+		Client:             ctx.Mgr.GetClient(),
+		ControlPlaneClient: ctx.ControlPlaneClient,
```
RainbowMango, line 715:
```suggestion
		ClientWithOutCache: versioned.NewForConfigOrDie(ctx.Mgr.GetConfig()),
```

pkg/controllers/context/context.go
```diff
@@ -109,6 +110,7 @@ type Context struct {
 	StopChan                    <-chan struct{}
 	DynamicClientSet            dynamic.Interface
 	KubeClientSet               clientset.Interface
+	ControlPlaneClient          client.Client
```
RainbowMango, line 113:
```suggestion
```
Then, we don't need to introduce the unusual client to the context.

chaosi-zju:
CC @RainbowMango ready to continue

besides, I modified a little code in e2e test, please help me do a check @XiShanYongYe-Chang

pkg/controllers/workloadrebalancer/ttlafterfinished_worker.go
```diff
@@ -0,0 +1,125 @@
...
+
+	// 4. Use the latest WorkloadRebalancer directly from api server to see if the TTL truly expires.
+	if expiredAt, err := c.processTTL(fresh); err != nil {
+		return err
+	} else if expiredAt == nil {
+		return nil
+	}
+
+	// 5. deletes the WorkloadRebalancer if TTL truly expires.
+	options := metav1.DeleteOptions{Preconditions: &metav1.Preconditions{ResourceVersion: &fresh.ResourceVersion}}
```
RainbowMango, line 75: The delete option(`ResourceVersion`) reminds me that we might not need to fetch the object from the APIServer.
Since in case of the object is rotten, the delete option would prevent this object from deleting.

/hold
Please help to run a test for it to prove it is necessary.

cmd/controller-manager/app/controllermanager.go
```diff
@@ -711,7 +711,7 @@ func startRemedyController(ctx controllerscontext.Context) (enabled bool, err er
 
 func startWorkloadRebalancerController(ctx controllerscontext.Context) (enabled bool, err error) {
 	workloadRebalancer := workloadrebalancer.RebalancerController{
-		Client: ctx.Mgr.GetClient(),
+		Client:             ctx.Mgr.GetClient(),
```
RainbowMango, line 714: might be a lint issue.

pkg/controllers/workloadrebalancer/workloadrebalancer_controller.go
```diff
@@ -220,16 +230,93 @@ func (c *RebalancerController) recordAndCountRebalancerFailed(resource *appsv1al
 
 func (c *RebalancerController) updateWorkloadRebalancerStatus(ctx context.Context, rebalancer *appsv1alpha1.WorkloadRebalancer,
 	newStatus *appsv1alpha1.WorkloadRebalancerStatus) error {
-	rebalancerPatch := client.MergeFrom(rebalancer)
+	if reflect.DeepEqual(rebalancer.Status, newStatus) {
+		return nil
+	}
+
```
chaosi-zju, line 236: Here is a bug, `rebalancer.Status` and `newStatus` will always not equal, which is not expected.

1. `rebalancer.Status.FinishTime` may be not nil, but `newStatus` will always be nil, we have not assigned a value to `newStatus.FinishTime`.
2. `rebalancer.Status` is a struct, while `newStatus` is a pointer.

This will lead to: even though `rebalancer.Status` unchanged, we will update `rebalancer.Status.FinishTime` when the rebalancer is reconciled.

pkg/controllers/workloadrebalancer/ttlafterfinished_worker.go
chaosi-zju: done, updated

RainbowMango (approved):
/lgtm
/approve

RainbowMango:
/hold
Please @XiShanYongYe-Chang take another look, particularly the e2e part.

chaosi-zju:
@XiShanYongYe-Chang not only e2e, but also ut

chaosi-zju:
@RainbowMango sorry, I did a updation after you send lgtm: https://github.com/karmada-io/karmada/compare/e405822231187e59ef62505a349aea75b8411ed2..f5076cdbeeb796d7df6426beb7ad98a1e790ceac

XiShanYongYe-Chang:
Ok~
/assign

pkg/controllers/workloadrebalancer/workloadrebalancer_controller.go
```diff
@@ -84,18 +92,34 @@ func (c *RebalancerController) Reconcile(ctx context.Context, req controllerrunt
...
-	if err := c.updateWorkloadRebalancerStatus(ctx, rebalancer, newStatus); err != nil {
-		return controllerruntime.Result{}, err
+	// 3. update status of WorkloadRebalancer if statusChanged
+	if result.statusChanged {
+		if err := c.updateWorkloadRebalancerStatus(ctx, rebalancer, result.newStatus); err != nil {
+			return controllerruntime.Result{}, err
+		}
 	}
+
 	klog.Infof("Finish handling WorkloadRebalancer (%s), %d/%d resource success in all, while %d resource need retry",
```
XiShanYongYe-Chang, line 104: Is this line of log really effective in locating the problem or is the log level too high? Logs about the processing failure of each workload are generated.

pkg/controllers/workloadrebalancer/workloadrebalancer_controller.go
```diff
@@ -84,18 +92,34 @@ func (c *RebalancerController) Reconcile(ctx context.Context, req controllerrunt
 	}
 
 	// 2. get and update referenced binding to trigger a rescheduling
-	newStatus, successNum, retryNum := c.doWorkloadRebalance(ctx, rebalancer)
+	result := c.doWorkloadRebalance(ctx, rebalancer)
 
-	// 3. update status of WorkloadRebalancer
-	if err := c.updateWorkloadRebalancerStatus(ctx, rebalancer, newStatus); err != nil {
-		return controllerruntime.Result{}, err
+	// 3. update status of WorkloadRebalancer if statusChanged
```
XiShanYongYe-Chang, line 97: Is step 3 part of step 2, or can the status update be put into method `doWorkloadRebalance`?

chaosi-zju: I don't think step 3 is part of step 2, these 3 steps are parallel: read rebalancer -> work according to rebalancer -> write rebalancer

pkg/controllers/workloadrebalancer/workloadrebalancer_controller.go
```diff
@@ -84,18 +92,34 @@ func (c *RebalancerController) Reconcile(ctx context.Context, req controllerrunt
...
+		return controllerruntime.Result{}, fmt.Errorf("%d resource reschedule triggered failed and need retry", result.retryNum)
 	}
-	klog.Infof("Finish handling WorkloadRebalancer (%s), %d/%d resource success in all, while %d resource need retry",
-		rebalancer.Name, successNum, len(newStatus.ObservedWorkloads), retryNum)
 
-	if retryNum > 0 {
-		return controllerruntime.Result{}, fmt.Errorf("%d resource reschedule triggered failed and need retry", retryNum)
+	// 4. when no more resource needs retry, judge whether it needs cleanup.
+	rebalancer.Status = *result.newStatus
+	if rebalancer.Spec.TTLSecondsAfterFinished != nil && rebalancer.Status.FinishTime != nil {
```
XiShanYongYe-Chang, line 113: This judgement `rebalancer.Status.FinishTime != nil` seems will always be true.

pkg/controllers/workloadrebalancer/workloadrebalancer_controller.go
```diff
@@ -220,16 +255,37 @@ func (c *RebalancerController) recordAndCountRebalancerFailed(resource *appsv1al
...
 		}
+		klog.V(4).Infof("Patch WorkloadRebalancer(%s) successful", rebalancer.Name)
 		return nil
 	})
 }
+
+func (c *RebalancerController) deleteWorkloadRebalancer(ctx context.Context, rebalancer *appsv1alpha1.WorkloadRebalancer) error {
+	klog.V(4).Infof("Start to clean up WorkloadRebalancer(%s)", rebalancer.Name)
+
+	options := &client.DeleteOptions{Preconditions: &metav1.Preconditions{ResourceVersion: &rebalancer.ResourceVersion}}
```
XiShanYongYe-Chang, line 275: Learned a new skill, thanks.

pkg/apis/apps/v1alpha1/workloadrebalancer_types.go
```diff
@@ -57,6 +57,14 @@ type WorkloadRebalancerSpec struct {
...
 	// +required
 	Workloads []ObjectReference `json:"workloads"`
+
+	// TTLSecondsAfterFinished limits the lifetime of a WorkloadRebalancer that has finished execution (means each
+	// target workload is finished with result of Successful or Failed).
+	// If this field is set, ttlSecondsAfterFinished after the WorkloadRebalancer finishes, it is eligible to be automatically deleted.
+	// If this field is unset, the WorkloadRebalancer won't be automatically deleted.
+	// If this field is set to zero, the WorkloadRebalancer becomes eligible to be deleted immediately after it finishes.
+	// +optional
+	TTLSecondsAfterFinished *int32 `json:"ttlSecondsAfterFinished,omitempty"`
```
XiShanYongYe-Chang, line 67: Can we point out that the unit of time here is seconds?

pkg/controllers/workloadrebalancer/workloadrebalancer_controller.go
chaosi-zju: done

test/e2e/framework/workloadrebalancer.go
```diff
@@ -76,3 +83,19 @@ func WaitRebalancerObservedWorkloads(client karmada.Interface, name string, expe
...
 	})
 }
+
+// WaitRebalancerDisappear wait WorkloadRebalancer disappear until timeout.
+func WaitRebalancerDisappear(client karmada.Interface, name string) {
+	ginkgo.By(fmt.Sprintf("Waiting for WorkloadRebalancer(%s) disappears", name), func() {
+		gomega.Eventually(func() error {
+			rebalancer, err := client.AppsV1alpha1().WorkloadRebalancers().Get(context.TODO(), name, metav1.GetOptions{})
+			if err == nil {
+				return err
```
XiShanYongYe-Chang, line 93: Use `return nil` maybe more readable.

XiShanYongYe-Chang (commented):
Thanks~
Generally OK.

pkg/controllers/workloadrebalancer/workloadrebalancer_controller.go
chaosi-zju: updated

RainbowMango:
> [FAILED] in [DeferCleanup (Each)] - /home/runner/work/karmada/karmada/test/e2e/framework/workloadrebalancer.go:46 @ 05/25/24 10:13:54.839

It seems related, see [the logs here](https://github.com/karmada-io/karmada/actions/runs/9234550079/job/25408399107?pr=4894#step:6:3815).

```
[FAILED] Unexpected error:
      <*errors.StatusError | 0xc0005c8d20>: 
      workloadrebalancers.apps.karmada.io "rebalancer-lhnmp" not found
      {
          ErrStatus: {
              TypeMeta: {Kind: "", APIVersion: ""},
              ListMeta: {
                  SelfLink: "",
                  ResourceVersion: "",
                  Continue: "",
                  RemainingItemCount: nil,
              },
              Status: "Failure",
              Message: "workloadrebalancers.apps.karmada.io \"rebalancer-lhnmp\" not found",
              Reason: "NotFound",
              Details: {
                  Name: "rebalancer-lhnmp",
                  Group: "apps.karmada.io",
                  Kind: "workloadrebalancers",
                  UID: "",
                  Causes: nil,
                  RetryAfterSeconds: 0,
              },
              Code: 404,
          },
      }
  occurred
  In [DeferCleanup (Each)] at: /home/runner/work/karmada/karmada/test/e2e/framework/workloadrebalancer.go:46 @ 05/25/24 10:13:54.839
```

chaosi-zju:
CI problem  fixed, is there any further comments?

XiShanYongYe-Chang:
/lgtm
Thanks

PR #4896, opened 2024-05-02
proxy list with rv=0 no longer access member cluster etcd

ikaven1024:
**What type of PR is this?**
/kind bug

When client request list with rv=0, proxy fetch objects from cache. It works as expected. 
While when pagination occurs, a `continue` will be responsed. This continue will filled with cluster's versions.
https://github.com/karmada-io/karmada/blob/02dad486f9628410de07eb4f3156972c8170073d/pkg/search/proxy/store/multi_cluster_cache.go#L295
And the next paging request will access etcd with version in `continue`. We should make it accessing cache also to improve performance.

```release-note
`karmada-search`: proxy list with rv=0 no longer access member cluster etcd
```


RainbowMango:
Thanks @ikaven1024 , please assign this to @XiShanYongYe-Chang or me once it's ready for review.

PR #4916, opened 2024-05-08
Call shutdown after being stopped

whitewindmills:
**What type of PR is this?**
/kind cleanup
/kind bug

we forget to call `shutdown` to close workqueue in these two components.
* karmada-scheduler
* karmada-metrics-adapter

Fixes #

```release-note
NONE
```

### DIFF
--- a/cmd/metrics-adapter/app/options/options.go
+++ b/cmd/metrics-adapter/app/options/options.go
@@ -80,7 +80,7 @@ func (o *Options) AddFlags(fs *pflag.FlagSet) {
 }
 
 // Config returns config for the metrics-adapter server given Options
-func (o *Options) Config() (*metricsadapter.MetricsServer, error) {
+func (o *Options) Config(stopCh <-chan struct{}) (*metricsadapter.MetricsServer, error) {
 	restConfig, err := clientcmd.BuildConfigFromFlags("", o.KubeConfig)
 	if err != nil {
 		klog.Errorf("Unable to build restConfig: %v", err)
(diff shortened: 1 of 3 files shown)


XiShanYongYe-Chang:
Thanks
/assign

XiShanYongYe-Chang (commented):
I'm wondering, if the queue is not shut down, what is the impact? Memory leak?

whitewindmills:
> Just a little question, if the queue is not shut down, what is the impact? Memory leak?

`ShutDown` will cause workqueue to ignore all new items added to it and immediately instruct the worker goroutines(include metrics goroutine) to exit. it is used to exit the currently running Goroutines gracefully. this is essential for managing resource cleanup and preventing memory leaks.

XiShanYongYe-Chang:
Thanks a lot~
/lgtm
/cc @chaunceyjiang @RainbowMango 

Do we need the release note and cherry pick this patch to the previous release?

whitewindmills:
ping @RainbowMango @chaunceyjiang

whitewindmills:
> Do we need the release note and cherry pick this patch to the previous release?

+1

whitewindmills:
> But does it cause memory leaks?

not absolutely, it depends on what is done in this goroutine. but we'd better follow good practices.

> do we need to stop the others, like InformerFactory, TypedInformerManager?

* for `InformerFactory`, judging from its implementation, it seems that calling the `Shutdown` function does not make sense.
https://github.com/karmada-io/karmada/blob/6f29134f1e3de90029e1e020dfab01e5df6d49cf/pkg/generated/informers/externalversions/factory.go#L146-L152

* for `TypedInformerManager`, even if the `StopInstance` function is called, 
https://github.com/karmada-io/karmada/blob/6f29134f1e3de90029e1e020dfab01e5df6d49cf/pkg/util/fedinformer/typedmanager/multi-cluster-manager.go#L30-L58
it will not take effect because the controller does not use the internal `stopCh` channel at all during initialization. this seems to be illegal. maybe we need other PR to modify this part of the logic.
https://github.com/karmada-io/karmada/blob/6f29134f1e3de90029e1e020dfab01e5df6d49cf/pkg/metricsadapter/controller.go#L71
https://github.com/karmada-io/karmada/blob/6f29134f1e3de90029e1e020dfab01e5df6d49cf/pkg/metricsadapter/controller.go#L134-L140
`context.TODO().Done()` is used as a stop semaphore.

RainbowMango:
> not absolutely, it depends on what is done in this goroutine. but we'd better follow good practices.

Yeah, Gracefully shutting down is good. I just want to clarify the impact as well as the benefit, the decision regarding to backport or not is based on the analysis. 
In my opinion, there won't be any memory leak issue here, so no need to backport this. 
One of the benefits I can get is that, for the informer, a graceful shutdown could close the connection between the target Kubernetes API server in time, there isn't a need to wait time out for the apiserver side to close the connection.

RainbowMango:
> it will not take effect because the controller does not use the internal stopCh channel at all during initialization. this seems to be illegal. maybe we need other PR to modify this part of the logic.

Just quickly looked at the code, yes, the usage of Stopch and context seems casual, welcome to optimize it a little bit.

whitewindmills:
ping @RainbowMango

RainbowMango (approved):
/lgtm
/approve

PR #4932, opened 2024-05-12
chore: support auto bump container images with dependabot.

### LINKED ISSUE #4908: [CVE-2024-2511] openssl: Unbounded memory growth with session handling in TLSv1.3
Bump base image alpine([here](https://github.com/karmada-io/karmada/blob/c401a5d9f1434bffa8442d5dd270dee311b332e4/cluster/images/buildx.Dockerfile#L15), and [here](https://github.com/karmada-io/karmada/blob/c401a5d9f1434bffa8442d5dd270dee311b332e4/cluster/images/Dockerfile#L15)) on all supported branches.
- [x] master (https://github.com/karmada-io/karmada/pull/4973)
- [x] release-1.9 (https://github.com/karmada-io/karmada/pull/4971)
- [x] release-1.8 (https://github.com/karmada-io/karmada/pull/4975)
- [x] release-1.7 (https://github.com/karmada-io/karmada/pull/4972)

**Why is this needed**:
There is a vulnerability alert reported by code scanning, that is the [CVE-2024-2511](https://avd.aquasec.com/nvd/cve-2024-2511), no evidence shows Karmada is affected by this issue, but we can bump the base image to silence this alert.

liangyuanpeng:
**What type of PR is this?**

/kind cleanup

support auto bump container images with dependabot, but I'm not sure whether we should treat release branch like this, it's means we need to update release branch from this file when a new release is out.

option2 is only auto bump container image for master and cherry pick the PR of dependabot to release branch if need.

Fixes #4908

```release-note

```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,3 +11,26 @@ updates:
   schedule:
     # Check for updates to GitHub Actions every week
     interval: "weekly"
+    
+- package-ecosystem: docker
+  directory: /cluster/images/
+  schedule:
+    interval: weekly
+
+- package-ecosystem: docker
+  directory: /cluster/images/
+  target-branch: "release-1.9"
+  schedule:
+    interval: weekly
+
+- package-ecosystem: docker
+  directory: /cluster/images/
+  target-branch: "release-1.8"
+  schedule:
+    interval: weekly
+
+- package-ecosystem: docker
+  directory: /cluster/images/
+  target-branch: "release-1.7"
+  schedule:
+    interval: weekly


RainbowMango:
Amazing!
> I'm not sure whether we should treat release branch like this, it's means we need to update release branch from this file when a new release is out.

I think letting Dependabot handle these things for both master and release branches would be better. 

Two more things we should consider about:
1. we need to update the configuration(`.github/dependabot.yml`) after a new release and remove the one that is out of maintenance. 
2. The PR sent by Dependabot doesn't use the PR template, how to set the release notes?

liangyuanpeng:
1) We can update it when needed together with https://github.com/karmada-io/karmada/blob/c7ca958987520d9699a5032b2b9d61366a5966c5/.github/workflows/ci-schedule-compatibility.yaml#L20, so it won't be forgotten

2) apparently I forgot about this, dependabot can't cover this. We may need other things to do this. like: edit the PR description if PR author is dependabot[bot] and PR have the label `docker` 
 2.1)  write the javascript with github action `actions/github-script`
 2.2) wirte the code with go and run it with github action.
 2.3) write a prow plugin.
 
 For simplicity I would choose option 2.2, other idea is welcome .

.github/dependabot.yml
```diff
@@ -11,3 +11,26 @@ updates:
...
+- package-ecosystem: docker
+  directory: /cluster/images/
+  schedule:
+    interval: weekly
+
+- package-ecosystem: docker
+  directory: /cluster/images/
+  target-branch: "release-1.9"
+  schedule:
+    interval: monthly
```
RainbowMango, line 24:
```suggestion
    interval: weekly
```
How about also doing it weekly? 

I'm still thinking if we should run it by daily. Is there some side-effect if we shorten the interval?

RainbowMango:
> We can update it when needed together with

OK.

> apparently I forgot about this, dependabot can't cover this. We may need other things to do this. like: edit the PR description if PR author is dependabot[bot] and PR have the label docker
2.1) write the javascript with github action actions/github-script
2.2) wirte the code in karmada with go repo and run it with github action.
2.3) write a prow plugin.

> For simplicity I would choose option 2.2, other idea is welcome .

Will see, those solutions sound a little bit heavy.

.github/dependabot.yml
liangyuanpeng: just want to reduce the instability, we can run it with daily if we believe that the latest is always stable.

RainbowMango: The reason why we bump the base image is usually regarding security, so i want to do it ASAP.
Given that we are using the latest `official` release, so I think we can trust it.

RainbowMango: Weekly would be fine for me.

RainbowMango (commented):
/assign

.github/dependabot.yml
liangyuanpeng: updated to Weekly

RainbowMango (approved):
/lgtm
/approve

PR #4943, opened 2024-05-15
feat: remove `PersistentVolumeClaim` annotation `volume.kubernetes.io/selected-node`

a7i:
**What type of PR is this?**

/kind bug

As my colleague @jklaw90  discovered that selected-node is for the node that volume is mounted on. If this node doesn't exist, then it is never removed. See upstream issue: https://github.com/kubernetes/kubernetes/issues/100485

we should also remove this field since it's irrelevant, given that node names are likely not going to be the same.

Fixes #

```release-note
karmada-controller-manager: remove cluster specific `PersistentVolume` annotation `volume.kubernetes.io/selected-node`
```


RainbowMango (commented):
/approve

Leave LGTM to @XiShanYongYe-Chang to take another look.

XiShanYongYe-Chang (commented):
Thanks~
/lgtm

Issue #4945, opened 2024-05-15
`ResourceInterpreterCustomization` for `Prune`


a7i:
**What would you like to be added**:
Currently, `Prune` is only supported for built-in, and there are use-cases where we could use Prune for custom resources.

**Why is this needed**:
[this blog](https://xinzhao.me/posts/guide-to-karmada-resource-interpreter-webhook/) does a good job of covering the interpreters and based on the image, we may want to exclude labels/annotations before Work is created
![image](https://github.com/karmada-io/karmada/assets/9448877/578a53c6-ad09-4e1c-882f-82cb903f7399)

XiShanYongYe-Chang:
@a7i Thanks for your feedback, I think this is an interesting feature point.

Before we start discussing how to do this, let's invite some ideas from people who might be interested in this.

/cc @chaunceyjiang @whitewindmills @RainbowMango @yike21 @iawia002

XiShanYongYe-Chang:
> we may want to exclude labels/annotations before Work is created

Hi @a7i I'd like to know under what circumstances this would be necessary. Because after communicating with @RainbowMango, I realized that `retain` can be used to handle the conflict between the Karmada control plane and member clusters.

a7i:
Prune would be to strip certain fields (e.g. label) from being propagated. With Retain, the label still goes to the member cluster but the value is essentially ignored.

One use-case is that we use a label to determine if a workload should be duplicated in all member clusters or not (the label is matched by the policy labelSelector). There is no need for this label to get propagated to the member clusters.

XiShanYongYe-Chang:
Thanks for the use case. I think it makes sense.
cc @RainbowMango

RainbowMango:
It seems to make sense, I'll get back to it later.

whitewindmills:
> One use-case is that we use a label to determine if a workload should be duplicated in all member clusters or not (the label is matched by the policy labelSelector). There is no need for this label to get propagated to the member clusters.

make sense.

RainbowMango:
> One use-case is that we use a label to determine if a workload should be duplicated in all member clusters or not (the label is matched by the policy labelSelector). There is no need for this label to get propagated to the member clusters.

It seems a reasonable use case that you label resources and let PropagationPolicy match them. 
But I have several questions:
1. I know there is no need for this label to get propagated, but that doesn't do any harm, right?
2. Can we prune the unnecessary label by OverridePolicy? I'm wondering if this is an option.

[...]

PR #4946, opened 2024-05-16
fix: prune deployment revision annotation

a7i:
**What type of PR is this?**

/kind bug

We noticed that karmada and member cluster controller-manager are fighting against each other where we get over 6k updates for a single deployment per hour.

the culprit seems to be the revision annotation.

Screenshot 2024-05-15 at 8 49 36 PM

Fixes #

while generally the karmdaa setup doesn't run the Deployment controller on the karmada-apiserver, it does support promoting resources which can lead to having these annotations cause a diff with the member cluster(s).

```release-note
fix: prune deployment revision annotations
```

### DIFF
--- a/pkg/resourceinterpreter/default/native/prune/prune.go
+++ b/pkg/resourceinterpreter/default/native/prune/prune.go
@@ -24,6 +24,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	storagevolume "k8s.io/component-helpers/storage/volume"
+	utildeployment "k8s.io/kubectl/pkg/util/deployment"
 
 	"github.com/karmada-io/karmada/pkg/util"
 	"github.com/karmada-io/karmada/pkg/util/helper"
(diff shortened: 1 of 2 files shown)


RainbowMango:
> Hi @a7i I think that the requirement proposed by #4945 can solve the issue raised by the current PR.

Yes, but before that, what's your opinion on this PR?
It makes sense to me to remove those annotations from Deployment as it is the Kubernetes native resources.

XiShanYongYe-Chang:
Previous view: Can we put the conflicts caused by controllers that are not enabled into the new hook point?

> Previous view: It makes sense to me to remove those annotations from Deployment as it is the Kubernetes native resources.

Based on this, new view: For Kubernetes native resources, we can put it into the default behavior.

a7i:
>the conflict comes when Karmada tries to update them back?

That's correct!

>I think that the requirement proposed by https://github.com/karmada-io/karmada/issues/4945 can solve the issue raised by the current PR.

Retain or Prune will work here.  With Retain, we'll just get an extra Update after first creation but ignored after that.

Given it's kubernetes native annotation, I think it makes sense to make it a default behavior but will defer to you all.

RainbowMango:
> Given it's kubernetes native annotation, I think it makes sense to make it a default behavior but will defer to you all.

+1
LGTM
But seems you might need to rebase and resolve the conflicts.

RainbowMango:
@a7i Can you help to rebase this PR and resolve the conflicts?

RainbowMango (approved):
/lgtm
/approve

PR #4928, opened 2024-05-10
cleanup dead code of generating Work permanent id

whitewindmills:
**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**:
Since the PR(https://github.com/karmada-io/karmada/pull/4474) has been merged, we're going to remove those codes which generate work permanent ID outside of webhook.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #4474, opened 2023-12-25
feat: add permanent id label to pp/cpp/rb/crb/work resources

jwcesign:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
Please referring to: https://github.com/karmada-io/karmada/issues/4000 and https://github.com/karmada-io/karmada/pull/4199#discussion_r1393842038

**Which issue(s) this PR fixes**:
Part of https://github.com/karmada-io/karmada/issues/4000

**Special notes for your reviewer**:
The corresponding yaml configuration will be done in another PR.

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #4949, opened 2024-05-16
cleanup dead code of generating PropagationPolicy permanent id

whitewindmills:
**What type of PR is this?**
/kind cleanup

Since the PR(https://github.com/karmada-io/karmada/pull/4474) has been merged, we're going to remove those codes which generate policy permanent ID outside of webhook.
Same as https://github.com/karmada-io/karmada/pull/4928

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/detector/detector.go
+++ b/pkg/detector/detector.go
@@ -720,26 +720,9 @@ func (d *ResourceDetector) GetUnstructuredObject(objectKey keys.ClusterWideKey) 
 	return unstructuredObj, nil
 }
 
-func (d *ResourceDetector) getPropagationPolicyID(policy *policyv1alpha1.PropagationPolicy) (string, error) {
-	id := util.GetLabelValue(policy.GetLabels(), policyv1alpha1.PropagationPolicyPermanentIDLabel)
-	if id == "" {
-		id = uuid.New().String()
-		policy.Labels = util.DedupeAndMergeLabels(policy.Labels, map[string]string{policyv1alpha1.PropagationPolicyPermanentIDLabel: id})
-		if err := d.Client.Update(context.TODO(), policy); err != nil {
-			return id, err
-		}
-	}
-
-	return id, nil
-}
-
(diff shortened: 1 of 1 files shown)


RainbowMango (commented):
/assign

XiShanYongYe-Chang (commented):
Thanks a lot
/lgtm

pkg/detector/detector.go
```diff
@@ -720,26 +720,9 @@ func (d *ResourceDetector) GetUnstructuredObject(objectKey keys.ClusterWideKey)
...
-}
-
 // ClaimPolicyForObject set policy identifier which the object associated with.
 func (d *ResourceDetector) ClaimPolicyForObject(object *unstructured.Unstructured, policy *policyv1alpha1.PropagationPolicy) (string, error) {
-	policyID, err := d.getPropagationPolicyID(policy)
-	if err != nil {
-		klog.Errorf("Get PropagationPolicy(%s/%s) ID error:%v", policy.Namespace, policy.Name, err)
-		return "", err
-	}
+	policyID := policy.Labels[policyv1alpha1.PropagationPolicyPermanentIDLabel]
```
RainbowMango, line 725: Have you ever thought about what's gonna happen if there is no `permanent` ID?
I know it is not a normal case, but just guess, is there something bad that would happen? And is it easy to identify that the issue is caused by the lack of permanent ID?

RainbowMango (approved):
/approve

pkg/detector/detector.go
whitewindmills: good question. 

> is there something bad that would happen?

this is very likely to cause the resource template to be unable to be properly unbound from the policy.

> And is it easy to identify that the issue is caused by the lack of permanent ID?

it's very hard.

since the policy is a user-oriented resource, can we introduce a protection mechanism for the permanent ID? For example, add verification to the webhook to protect this label from being modified or deleted.

RainbowMango: +1
on the webhook validation.

PR #4955, opened 2024-05-17
upgrade rsa key size from 2048 to 3072

chaosi-zju:
**What type of PR is this?**

/kind cleanup

[According to NSA](https://apps.nsa.gov/iaarchive/programs/iad-initiatives/cnsa-suite.cfm) and [ANSSI](http://www.ssi.gouv.fr/uploads/2015/01/RGS_v-2-0_B1.pdf), RSA with 3072 bit-modulus is the minimum to protect up to TOP SECRET.

We should not be in the red line of cryptography. This security by default will bring the needed protection for users.

Fixes #

```release-note

```


RainbowMango:
@zhzhuang-zju @yanfeng1992 Would you like to take a look?

pkg/util/lifted/pubkeypin/pubkeypin_test.go
```diff
@@ -24,7 +24,7 @@ import (
 )
 
 // testCertPEM is a simple self-signed test certificate issued with the openssl CLI:
-// openssl req -new -newkey rsa:2048 -days 36500 -nodes -x509 -keyout /dev/null -out test.crt
+// openssl req -new -newkey rsa:3072 -days 36500 -nodes -x509 -keyout /dev/null -out test.crt
```
zhzhuang-zju, line 27: Do the test certificate also need to be regenerated? The rsa key size of testCertPEM is 2048
```bash
(base) ➜  karmada git:(main) ✗  openssl x509 -noout -text -in test.crt
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number:
            98:15:68:25:ef:0b:a1:e4
        Signature Algorithm: sha1WithRSAEncryption
        Issuer: CN = kubeadm-keypins-test
        Validity
            Not Before: Jul  5 17:43:10 2017 GMT
            Not After : Jun 11 17:43:10 2117 GMT
        Subject: CN = kubeadm-keypins-test
        Subject Public Key Info:
            Public Key Algorithm: rsaEncryption
                RSA Public-Key: (2048 bit)
```

charts/karmada/templates/pre-install-job.yaml
```diff
@@ -388,12 +388,12 @@ spec:
           mkdir -p /opt/configs
           mkdir -p /opt/certs
           cp -r -L /opt/mount/* /opt/configs/
-          openssl req -x509 -sha256 -new -nodes -days 365 -newkey rsa:2048 -keyout "/opt/certs/server-ca.key" -out "/opt/certs/server-ca.crt" -subj "/C=xx/ST=x/L=x/O=x/OU=x/CN=ca/emailAddress=x/"
-          openssl req -x509 -sha256 -new -nodes -days 365 -newkey rsa:2048 -keyout "/opt/certs/front-proxy-ca.key" -out "/opt/certs/front-proxy-ca.crt" -subj "/C=xx/ST=x/L=x/O=x/OU=x/CN=ca/emailAddress=x/"
+          openssl req -x509 -sha256 -new -nodes -days 365 -newkey rsa:3072 -keyout "/opt/certs/server-ca.key" -out "/opt/certs/server-ca.crt" -subj "/C=xx/ST=x/L=x/O=x/OU=x/CN=ca/emailAddress=x/"
```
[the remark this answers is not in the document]
chaosi-zju: good point, done

hack/util.sh
```diff
@@ -213,7 +213,7 @@ function util::create_signing_certkey {
     # Create ca
     ${sudo} /usr/bin/env bash -e <<EOF
     rm -f "${dest_dir}/${id}.crt" "${dest_dir}/${id}.key"
-    ${OPENSSL_BIN} req -x509 -sha256 -new -nodes -days 3650 -newkey rsa:2048 -keyout "${dest_dir}/${id}.key" -out "${dest_dir}/${id}.crt" -subj "/CN=${cn}/"
+    ${OPENSSL_BIN} req -x509 -sha256 -new -nodes -days 3650 -newkey rsa:3072 -keyout "${dest_dir}/${id}.key" -out "${dest_dir}/${id}.crt" -subj "/CN=${cn}/"
```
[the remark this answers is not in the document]
chaosi-zju: I don't think it's necessary to define this place as a variable.

pkg/util/lifted/pubkeypin/pubkeypin_test.go
chaosi-zju: nice, done

operator/pkg/certs/certs.go
```diff
@@ -46,7 +46,7 @@ import (
 const (
 	// CertificateBlockType is a possible value for pem.Block.Type.
 	CertificateBlockType = "CERTIFICATE"
-	rsaKeySize           = 2048
+	rsaKeySize           = 3072
```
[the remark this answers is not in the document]
chaosi-zju: I prefer these two components just have their own global variables.

Currently, these two components have no common dependencies except for the API. I don't want them to have common dependencies coupled together just because of this small tip.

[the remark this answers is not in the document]
zhzhuang-zju: `operator` and `karmadactl` are more like two installation methods than two components. And karmada should behave the same in different installations, so I think there could be a common set of base configurations directly for different installations. 
This issue can be put aside for the time being, and if necessary, we can submit a new pr and discuss it later.

charts/karmada/templates/pre-install-job.yaml
```diff
@@ -388,12 +388,12 @@ spec:
           mkdir -p /opt/configs
           mkdir -p /opt/certs
           cp -r -L /opt/mount/* /opt/configs/
-          openssl req -x509 -sha256 -new -nodes -days 365 -newkey rsa:2048 -keyout "/opt/certs/server-ca.key" -out "/opt/certs/server-ca.crt" -subj "/C=xx/ST=x/L=x/O=x/OU=x/CN=ca/emailAddress=x/"
-          openssl req -x509 -sha256 -new -nodes -days 365 -newkey rsa:2048 -keyout "/opt/certs/front-proxy-ca.key" -out "/opt/certs/front-proxy-ca.crt" -subj "/C=xx/ST=x/L=x/O=x/OU=x/CN=ca/emailAddress=x/"
+          openssl req -x509 -sha256 -new -nodes -days 365 -newkey rsa:{{ printf `"%s"` .Values.certs.auto.rsaSize }} -keyout "/opt/certs/server-ca.key" -out "/opt/certs/server-ca.crt" -subj "/C=xx/ST=x/L=x/O=x/OU=x/CN=ca/emailAddress=x/"
```
yanfeng1992, line 391: {{ .Values.certs.auto.rsaSize | quote }}

chaosi-zju: hi, what is this ` | quote` intention?

yanfeng1992: https://helm.sh/zh/docs/chart_template_guide/function_list/#quote-and-squote 

When the value read from .Values is turned into a string, it can be achieved by calling the quote template function.


to fix 

image

chaosi-zju: So it is more normalize to replace `printf  "%s" ...` with `... | quote` ?

There are still a few lines below that also use `printf  "%s" ...` functions, do we need to unify them all into `... | quote` ?

yanfeng1992: sorry，my bad

{{ .Values.certs.auto.rsaSize }} should be used directly here. Using {{ .Values.certs.auto.rsaSize | quote }} will result in extra double quotes.

image


You can use helm template command to test locally

helm template ./charts/karmada/

image

chaosi-zju: Now I totally get it. Thanks!

charts/karmada/templates/pre-install-job.yaml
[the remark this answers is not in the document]
RainbowMango: +1

RainbowMango (approved):
/approve

PR #4959, opened 2024-05-17
[Cleanup] Add empty judgment for placement in taint-manager

XiShanYongYe-Chang:
**What type of PR is this?**

/kind cleanup

1. Add empty judgment for placement in taint-manager, avoid panic when the obtained placement is nil:
https://github.com/karmada-io/karmada/blob/90b461603c8025edc9696e3a0c54181e36b4b607/pkg/controllers/cluster/taint_manager.go#L267
2. Add more logs to facilitate fault locating.

Fixes #

In order to make the program more robust, nil judgment is added. 

This is because when users define their own scheduler, the placement recorded in the annotation in the binding may be empty.

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/cluster/taint_manager.go
+++ b/pkg/controllers/cluster/taint_manager.go
@@ -144,6 +144,8 @@ func (tc *NoExecuteTaintManager) syncBindingEviction(key util.QueueKey) error {
 		return fmt.Errorf("invalid key")
 	}
 	cluster := fedKey.Cluster
+	klog.V(4).Infof("Begin to sync ResourceBinding(%s) with taintManager for Cluster(%s)",
+		fedKey.ClusterWideKey.NamespaceKey(), cluster)
 
 	binding := &workv1alpha2.ResourceBinding{}
 	if err := tc.Client.Get(context.TODO(), types.NamespacedName{Namespace: fedKey.Namespace, Name: fedKey.Name}, binding); err != nil {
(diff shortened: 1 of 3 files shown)


pkg/controllers/cluster/taint_manager.go
```diff
@@ -245,11 +253,14 @@ func (tc *NoExecuteTaintManager) syncClusterBindingEviction(key util.QueueKey) e
...
-func (tc *NoExecuteTaintManager) needEviction(clusterName string, appliedPlacement map[string]string) (bool, time.Duration, error) {
-	placement, err := helper.GetAppliedPlacement(appliedPlacement)
+func (tc *NoExecuteTaintManager) needEviction(clusterName string, annotations map[string]string) (bool, time.Duration, error) {
+	placement, err := helper.GetAppliedPlacement(annotations)
 	if err != nil {
 		return false, -1, err
 	}
+	if placement == nil {
+		return false, -1, fmt.Errorf("the applied placement for ResourceBining can not be empty")
+	}
```
RainbowMango, line 263: Can you remind me in what situations the placement could be nil?

XiShanYongYe-Chang: This problem does not occur in the current implementation of Karmada, but it depends on the scheduler. If the user does not use the default scheduler, this problem may occur.

RainbowMango: Got, then please add a comment on it, otherwise, it brings misunderstanding.

XiShanYongYe-Chang: Updated.

RainbowMango:
You have to wait #4966

RainbowMango (approved):
/lgtm
/approve

PR #4964, opened 2024-05-20
Add validation on policy permanent ID

whitewindmills:
**What type of PR is this?**
/kind feature

Add protection mechanism for the permanent ID, see https://github.com/karmada-io/karmada/pull/4949#discussion_r1606293748

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/webhook/clusterpropagationpolicy/mutating.go
+++ b/pkg/webhook/clusterpropagationpolicy/mutating.go
@@ -23,6 +23,7 @@ import (
 	"net/http"
 
 	"github.com/google/uuid"
+	admissionv1 "k8s.io/api/admission/v1"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
 	policyv1alpha1 "github.com/karmada-io/karmada/pkg/apis/policy/v1alpha1"
@@ -82,7 +83,7 @@ func (a *MutatingAdmission) Handle(_ context.Context, req admission.Request) adm
 		}
 	}
 
-	if util.GetLabelValue(policy.Labels, policyv1alpha1.ClusterPropagationPolicyPermanentIDLabel) == "" {
+	if req.Operation == admissionv1.Create {
 		util.MergeLabel(policy, policyv1alpha1.ClusterPropagationPolicyPermanentIDLabel, uuid.New().String())
 	}
 
--- a/pkg/webhook/clusterpropagationpolicy/validating.go
+++ b/pkg/webhook/clusterpropagationpolicy/validating.go
@@ -60,6 +60,16 @@ func (v *ValidatingAdmission) Handle(_ context.Context, req admission.Request) a
 			klog.Error(err)
 			return admission.Denied(err.Error())
 		}
+
+		if policy.Labels[policyv1alpha1.ClusterPropagationPolicyPermanentIDLabel] !=
+			oldPolicy.Labels[policyv1alpha1.ClusterPropagationPolicyPermanentIDLabel] {
+			return admission.Denied(fmt.Sprintf("label %s is immutable, it can only be set by the system during creation",
+				policyv1alpha1.ClusterPropagationPolicyPermanentIDLabel))
+		}
+	}
+	if _, exist := policy.Labels[policyv1alpha1.ClusterPropagationPolicyPermanentIDLabel]; !exist {
+		return admission.Denied(fmt.Sprintf("label %s is required, it should be set by the mutating admission webhook during creation",
+			policyv1alpha1.ClusterPropagationPolicyPermanentIDLabel))
 	}
 
 	errs := validation.ValidatePropagationSpec(policy.Spec)
--- a/pkg/webhook/propagationpolicy/validating.go
+++ b/pkg/webhook/propagationpolicy/validating.go
@@ -60,6 +60,16 @@ func (v *ValidatingAdmission) Handle(_ context.Context, req admission.Request) a
 			klog.Error(err)
 			return admission.Denied(err.Error())
 		}
+
+		if policy.Labels[policyv1alpha1.PropagationPolicyPermanentIDLabel] !=
+			oldPolicy.Labels[policyv1alpha1.PropagationPolicyPermanentIDLabel] {
+			return admission.Denied(fmt.Sprintf("label %s is immutable, it can only be set by the system during creation",
+				policyv1alpha1.PropagationPolicyPermanentIDLabel))
+		}
+	}
+	if _, exist := policy.Labels[policyv1alpha1.PropagationPolicyPermanentIDLabel]; !exist {
+		return admission.Denied(fmt.Sprintf("label %s is required, it should be set by the mutating admission webhook during creation",
+			policyv1alpha1.PropagationPolicyPermanentIDLabel))
 	}
 
 	errs := validation.ValidatePropagationSpec(policy.Spec)
--- a/pkg/webhook/propagationpolicy/mutating.go
+++ b/pkg/webhook/propagationpolicy/mutating.go
@@ -23,6 +23,7 @@ import (
 	"net/http"
 
 	"github.com/google/uuid"
+	admissionv1 "k8s.io/api/admission/v1"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
@@ -94,7 +95,7 @@ func (a *MutatingAdmission) Handle(_ context.Context, req admission.Request) adm
 		}
 	}
 
-	if util.GetLabelValue(policy.Labels, policyv1alpha1.PropagationPolicyPermanentIDLabel) == "" {
+	if req.Operation == admissionv1.Create {
 		util.MergeLabel(policy, policyv1alpha1.PropagationPolicyPermanentIDLabel, uuid.New().String())
 	}
 


whitewindmills:
test result:
image

whitewindmills:
unrelated E2E case: https://github.com/karmada-io/karmada/actions/runs/9155349434/job/25167705895?pr=4964#step:6:3402

XiShanYongYe-Chang:
> unrelated E2E case: https://github.com/karmada-io/karmada/actions/runs/9155349434/job/25167705895?pr=4964#step:6:3402

Yes, this is currently being fixed.

pkg/webhook/clusterpropagationpolicy/mutating.go
```diff
@@ -82,7 +83,7 @@ func (a *MutatingAdmission) Handle(_ context.Context, req admission.Request) adm
 		}
 	}
 
-	if util.GetLabelValue(policy.Labels, policyv1alpha1.ClusterPropagationPolicyPermanentIDLabel) == "" {
+	if req.Operation == admissionv1.Create {
```
XiShanYongYe-Chang, line 86: Will we be able to do this in the next version? For upgraded resources, it may no longer have the opportunity to experience creation events.

RainbowMango: Actually, all propagation policies are already labeled, at least in v1.9, so no problem with upgrading.

See the code here, which was introduced in v1.9-:
https://github.com/karmada-io/karmada/pull/4949/files#diff-fdb6430a46e82ace9f449b971619671eac711f7a82d6b862c130236a65cd5101L723-L734

whitewindmills: also applies to the release-1.8.

pkg/webhook/clusterpropagationpolicy/validating.go
```diff
@@ -60,6 +60,12 @@ func (v *ValidatingAdmission) Handle(_ context.Context, req admission.Request) a
 			klog.Error(err)
 			return admission.Denied(err.Error())
 		}
+
+		if policy.Labels[policyv1alpha1.ClusterPropagationPolicyPermanentIDLabel] !=
+			oldPolicy.Labels[policyv1alpha1.ClusterPropagationPolicyPermanentIDLabel] {
+			return admission.Denied(fmt.Sprintf("the label %q is protected, any updates are prohibited",
+				policyv1alpha1.ClusterPropagationPolicyPermanentIDLabel))
```
RainbowMango, line 67:
```suggestion
			return admission.Denied(fmt.Sprintf("label %s is immutable, it can only be set by the system during creation",
				policyv1alpha1.ClusterPropagationPolicyPermanentIDLabel))
```

pkg/webhook/clusterpropagationpolicy/validating.go
```diff
@@ -60,6 +60,12 @@ func (v *ValidatingAdmission) Handle(_ context.Context, req admission.Request) a
 			klog.Error(err)
 			return admission.Denied(err.Error())
 		}
+
+		if policy.Labels[policyv1alpha1.ClusterPropagationPolicyPermanentIDLabel] !=
+			oldPolicy.Labels[policyv1alpha1.ClusterPropagationPolicyPermanentIDLabel] {
+			return admission.Denied(fmt.Sprintf("the label %q is protected, any updates are prohibited",
```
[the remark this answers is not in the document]
whitewindmills: if I try to delete this label, I feel like such description `the value of ...` is not very accurate. just echo `the label %q can not be updated or deleted`?

pkg/webhook/clusterpropagationpolicy/validating.go
RainbowMango: I tend not to use `%q` as it is used for printing integers. Feels tricky just for quotes.

RainbowMango: In addition, we still need to ensure the label's existence, even in the case of creation.

whitewindmills: I see the suggestion of @RainbowMango, I feel like it's ok, what do you think? @XiShanYongYe-Chang

pkg/webhook/propagationpolicy/validating.go
```diff
@@ -60,6 +60,12 @@ func (v *ValidatingAdmission) Handle(_ context.Context, req admission.Request) a
 			klog.Error(err)
 			return admission.Denied(err.Error())
 		}
+
+		if policy.Labels[policyv1alpha1.PropagationPolicyPermanentIDLabel] !=
+			oldPolicy.Labels[policyv1alpha1.PropagationPolicyPermanentIDLabel] {
+			return admission.Denied(fmt.Sprintf("the label %q is protected, any updates are prohibited",
```
RainbowMango, line 66: ditto.

pkg/webhook/clusterpropagationpolicy/validating.go
XiShanYongYe-Chang: ok

whitewindmills: > In addition, we still need to ensure the label's existence, even in the case of creation.

sorry, I don't fully understand it. do you mean that we still need to validate it when `Request.Operation` is `CREATE`?

XiShanYongYe-Chang: I think maybe we don't need it.

RainbowMango: We have to ensure the label `clusterpropagationpolicy.karmada.io/permanent-id` exists, that's why I ask to introduce validation, we can not skip it even though it is already added by mutating. 
Note that, validation runs after mutation.

whitewindmills: The only scenario I can think of is that the user privately accesses the policy's mutating webhook, which may result in the policy being not labeled.

RainbowMango: In general, the `validation admission webhook` is used to guarantee all fields are set as expected, it usually doesn't assume the `mutating admission webhook` is already well processed.

whitewindmills: make sense, thanks

whitewindmills: so, how about the error message `label %s is required, it should not happen` for the case of creation?

RainbowMango: `label %s is required, it should be set by the mutating admission webhook during creation`.
How about this, we can give some useful hints for people to figure out the root cause.

RainbowMango: > I think maybe we don't need it.

@XiShanYongYe-Chang Why do you think we don't need it?

XiShanYongYe-Chang: I think it is good.

pkg/webhook/propagationpolicy/validating.go
```diff
@@ -60,6 +60,17 @@ func (v *ValidatingAdmission) Handle(_ context.Context, req admission.Request) a
 			klog.Error(err)
 			return admission.Denied(err.Error())
 		}
+
+		if policy.Labels[policyv1alpha1.PropagationPolicyPermanentIDLabel] !=
+			oldPolicy.Labels[policyv1alpha1.PropagationPolicyPermanentIDLabel] {
+			return admission.Denied(fmt.Sprintf("label %s is immutable, it can only be set by the system during creation",
+				policyv1alpha1.PropagationPolicyPermanentIDLabel))
+		} else if req.Operation == admissionv1.Create {
```
XiShanYongYe-Chang, line 68: Isn't that the wrong level of judgment?

RainbowMango: Yeah, just check the label would be fine, don't have to care what the operation is. (don't need that `else if`)
```
	if _, exist := policy.Labels[policyv1alpha1.ClusterPropagationPolicyPermanentIDLabel]; !exist {
		return admission.Denied(fmt.Sprintf("label %s is required, it should be set by the mutating admission webhook during creation",
			policyv1alpha1.ClusterPropagationPolicyPermanentIDLabel))
	}
```

RainbowMango: Just like the following `validation.ValidatePropagationSpec(policy.Spec)`.

whitewindmills: feel unnecessary to check twice when the operation is `UPDATE`.

RainbowMango: ?
Why do you think it checks twice?
In the condition `req.Operation == admissionv1.Update`, we only foucs on the label can not be updated, consider the situation that mutating admission webhook might not work for whatever reason, both `old` and `new` might not have the label, so it can pass the check.

But we do require the label to be set, so we need to check it, but this time we focus on the existence.

whitewindmills: if the mutating admission webhook doesn't work, the policy cannot be created cause we check it when creating. 
> both old and new might not have the label, so it can pass the check.

so it will never happen.

when `req.Operation == admissionv1.Update`, it will only pass the check if there is no problem with the policy label. so we don't need check it again.

XiShanYongYe-Chang: Update like this:
```diff
diff --git a/pkg/webhook/propagationpolicy/validating.go b/pkg/webhook/propagationpolicy/validating.go                      
index 2e7c02c8e..98a73cecd 100644                                                                                           
--- a/pkg/webhook/propagationpolicy/validating.go                                                                           
+++ b/pkg/webhook/propagationpolicy/validating.go                                                                           
@@ -48,6 +48,10 @@ func (v *ValidatingAdmission) Handle(_ context.Context, req admission.Request) a                         
        }                                                                                                                   
        klog.V(2).Infof("Validating PropagationPolicy(%s/%s) for request: %s", policy.Namespace, policy.Name, req.Operation)
                                                                                                                            
+       if _, exist := policy.Labels[policyv1alpha1.PropagationPolicyPermanentIDLabel]; !exist {
+               return admission.Denied(fmt.Sprintf("label %s is required, it should be set by the mutating admission webhook during creation",
+                       policyv1alpha1.PropagationPolicyPermanentIDLabel))
+       }
        if req.Operation == admissionv1.Update {
                oldPolicy := &policyv1alpha1.PropagationPolicy{}
                err = v.Decoder.DecodeRaw(req.OldObject, oldPolicy)
@@ -65,11 +69,6 @@ func (v *ValidatingAdmission) Handle(_ context.Context, req admission.Request) a
                        oldPolicy.Labels[policyv1alpha1.PropagationPolicyPermanentIDLabel] {
                        return admission.Denied(fmt.Sprintf("label %s is immutable, it can only be set by the system during creation",
                                policyv1alpha1.PropagationPolicyPermanentIDLabel))
-               } else if req.Operation == admissionv1.Create {
-                       if _, exist := policy.Labels[policyv1alpha1.PropagationPolicyPermanentIDLabel]; !exist {
-                               return admission.Denied(fmt.Sprintf("label %s is required, it should be set by the mutating admission webhook during creation",
-                                       policyv1alpha1.PropagationPolicyPermanentIDLabel))
-                       }
                }
        }

```

RainbowMango: I get what you mean.  And you are right, that case i mentioned will not happen.

But, I tend to a simple pattern that lets validating admission webhook not care what exactly the operation is, and just focus on the validation.
For those checks really need to compare the `old` and `new`, we can put all these checks together. 

The pattern is simple and readable.

XiShanYongYe-Chang (commented):
Thanks a lot~
/lgtm

RainbowMango (approved):
/approve

XiShanYongYe-Chang:
Do we also need to add the permanentID validation for the RB/CRB and Work? @RainbowMango @whitewindmills

whitewindmills:
I think it’s reasonable regardless of whether to add it or not cause they are usually not visible to users. In addition, I think it is more necessary to add the validation for resource templates.

XiShanYongYe-Chang:
> I think it is more necessary to add the validation for resource templates.

What behavior are you referring to?

whitewindmills:
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  creationTimestamp: "2024-05-23T03:00:31Z"
  annotations:
    clusterpropagationpolicy.karmada.io/name: foo-cpp
  labels:
    clusterpropagationpolicy.karmada.io/name: foo-cpp
    clusterpropagationpolicy.karmada.io/permanent-id: c105115c-1243-4a4c-189c-990936097gs8
  name: foo
  namespace: default
  resourceVersion: "63274530"
  uid: b205114b-3443-4b4d-a89c-410936097ed5
```
protect these labels and annotations from modification by users, but I haven't figured out how to do it yet.

XiShanYongYe-Chang:
cc @RainbowMango

XiShanYongYe-Chang:
> I think it’s reasonable regardless of whether to add it or not cause they are usually not visible to users. 

Do we need to consider whether it is a user-aware API when validating?

PR #4965, opened 2024-05-20
fix e2e unstable failure caused by taint cluster

chaosi-zju:
**What type of PR is this?**

/kind flake

fix e2e unstable failure caused by taint cluster.

In this new merged e2e, we add a taint `{Key: "workload-rebalancer-test", Effect: corev1.TaintEffectNoExecute}` to member1 cluster.

Though it has specified key, we mistakenly think that it has no effect on other e2e cases. Actually, as long as the cluster has `NoExecute` taint, no matter what the key it is, could affect other e2e cases.

Fixes #

```release-note

```

### DIFF
--- a/test/e2e/workloadrebalancer_test.go
+++ b/test/e2e/workloadrebalancer_test.go
@@ -43,7 +43,7 @@ import (
 //	schedule strategy: static weight, dynamic weight, aggregated
 //	resource type: workload type like deployment, non-workload type like clusterrole
 //	expected result: successful, not found failure
-var _ = ginkgo.Describe("workload rebalancer testing", func() {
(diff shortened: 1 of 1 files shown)


XiShanYongYe-Chang:
/assign

We can give it a few more runs today.

chaosi-zju:
> /assign
> 
> We can give it a few more runs today.

I have run it for 10 times, 3 in here, and 7 in https://github.com/chaosi-zju/karmada/actions/runs/9156098150/job/25169774759

RainbowMango (approved):
/lgtm
/approve

Given this would block PRs, it should be fixed as soon as possible.

PR #4980, opened 2024-05-23
cleanup: update relate binding name and comments for ease of understanding

XiShanYongYe-Chang:
**What type of PR is this?**

/kind cleanup

**What this PR does / why we need it**:

In the dependencies distributor module, we have two kinds of binding, one is the binding that is followed, and the other is the binding that follows.

We have a clear explanation of these names:

https://github.com/karmada-io/karmada/blob/0084c8a97eba3b9ff3a0a355cbc95531bcb5c634/pkg/dependenciesdistributor/dependencies_distributor.go#L83-L84

But in the code, some places are not named according to such names, which makes us a little confused when reading the code and we need to repeatedly confirm the intent of the code.

Therefore, this PR unified the confusing naming and modified some comments.

In addition, the code uses two words, `depended` and `dependencies`, which correspond to the binding that is followed and binding that follows, respectively.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

/cc @RainbowMango @chaunceyjiang @whitewindmills

Issue #1741, opened 2022-05-08
The label value of the dependent resoucebinding over 63 characters


mathlsj:
The dependent resourcebinding label value consists of three components namspace+"_"+name+"-"+kind. it is easy to over 63 characters. we need to split it into three labels:
resourcebinding.karmada.io/depended-by-hash.namespace: namespace
resourcebinding.karmada.io/depended-by-hash.name: name
resourcebinding.karmada.io/depended-by-hash.kind: kind

mathlsj:
generate resourcebinding depended label in  dependencies_distributor (https://github.com/karmada-io/karmada/blob/master/pkg/dependenciesdistributor/dependencies_distributor.go)

```
func generateBindingDependedByLabel(bindingNamespace, bindingName string) map[string]string {
	labelKey := generateBindingDependedByLabelKey(bindingNamespace, bindingName)
        # labelValue = ns +"_" + bindingName (name + "-" + kind)
	labelValue := fmt.Sprintf(bindingNamespace + "_" + bindingName)
	return map[string]string{labelKey: labelValue}
}
```

RainbowMango:
Right, here probably exceed 64 characters. May I know what the `name` and `namespace` look like in your system?

RainbowMango:
Hi @mathlsj I see you send a PR for this. 
I'd like to invite the author to talk about the solution. @mrlihanbo

mathlsj:
OK。

mathlsj:
@RainbowMango What is the progress of this?

RainbowMango:
I'm sorry I missed this from my radar. I'll be back on this next week.

RainbowMango:
> Right, here probably exceed 64 characters. May I know what the `name` and `namespace` look like in your system?

@mathlsj Can you give an example of your `name` and `namespace`?

RainbowMango:
> we need to split it into three labels:
resourcebinding.karmada.io/depended-by-hash.namespace: namespace
resourcebinding.karmada.io/depended-by-hash.name: name
resourcebinding.karmada.io/depended-by-hash.kind: kind

Split the label could relieve the issue, but I'm afraid it can't solve the problem fundamentally, because the `name` itself may exceed 63 characters.

Investigation still on-going.

RainbowMango:
I've got two solutions here:
1. Instead of putting the key `resourcebinding.karmada.io/depended-by-xxx` to a label, we can put it in the `annotations`.

[...]

PR #4982, opened 2024-05-23
add verify for customized dependencies response

XiShanYongYe-Chang:
**What type of PR is this?**

/kind feature

After the user customizes interpretation behavior, the Dependencies returned by the `InterpretDependency` interpreter are verified to prevent incorrect return values ​​from penetrating into subsequent use logic.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/resourceinterpreter/customized/webhook/request/resourceinterpretercontext.go
+++ b/pkg/resourceinterpreter/customized/webhook/request/resourceinterpretercontext.go
@@ -25,6 +25,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/uuid"
 
 	configv1alpha1 "github.com/karmada-io/karmada/pkg/apis/config/v1alpha1"
+	"github.com/karmada-io/karmada/pkg/util/interpreter/validation"
 )
 
 // CreateResourceInterpreterContext returns the unique request uid, the ResourceInterpreterContext object to send the webhook,
@@ -114,6 +115,10 @@ func verifyResourceInterpreterContext(operation configv1alpha1.InterpreterOperat
 		res.ReplicaRequirements = response.ReplicaRequirements
 		return res, nil
 	case configv1alpha1.InterpreterOperationInterpretDependency:
+		err := validation.VerifyDependencies(response.Dependencies)
+		if err != nil {
+			return nil, err
+		}
 		res.Dependencies = response.Dependencies
 		return res, nil
 	case configv1alpha1.InterpreterOperationPrune, configv1alpha1.InterpreterOperationReviseReplica,
--- a/pkg/resourceinterpreter/default/native/default.go
+++ b/pkg/resourceinterpreter/default/native/default.go
@@ -26,6 +26,7 @@ import (
 
 	configv1alpha1 "github.com/karmada-io/karmada/pkg/apis/config/v1alpha1"
 	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
+	"github.com/karmada-io/karmada/pkg/util/interpreter/validation"
 )
 
 // DefaultInterpreter contains all default operation interpreter factory
@@ -136,7 +137,12 @@ func (e *DefaultInterpreter) GetDependencies(object *unstructured.Unstructured) 
 	if !exist {
 		return dependencies, fmt.Errorf("default interpreter for operation %s not found", configv1alpha1.InterpreterOperationInterpretDependency)
 	}
-	return handler(object)
+
+	dependencies, err = handler(object)
+	if err != nil {
+		return
+	}
+	return dependencies, validation.VerifyDependencies(dependencies)
 }
 
 // ReflectStatus returns the status of the object.
--- a/pkg/util/interpreter/validation/validation.go
+++ b/pkg/util/interpreter/validation/validation.go
@@ -1,0 +1,40 @@
+/*
+Copyright 2024 The Karmada Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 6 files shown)


chaunceyjiang:
It's just a question, how did you discover that optimization could be done here?

XiShanYongYe-Chang:
I've been working on the #1741 issue recently. I read the code of the dependencies distributor module and found some confusing naming. So I submitted PR #4980. 

In the #4980, there is an error info output:

https://github.com/karmada-io/karmada/blob/f414fa04b8e87ccd3fe11f8766a79339dfc8f8a6/pkg/dependenciesdistributor/dependencies_distributor.go#L326

I think this error message should be discovered much earlier, so I submitted this PR

pkg/resourceinterpreter/customized/webhook/request/resourceinterpretercontext.go
```diff
@@ -114,6 +115,10 @@ func verifyResourceInterpreterContext(operation configv1alpha1.InterpreterOperat
 		res.ReplicaRequirements = response.ReplicaRequirements
 		return res, nil
 	case configv1alpha1.InterpreterOperationInterpretDependency:
+		err := validate.VerifyResponseDependencies(response.Dependencies)
```
whitewindmills, line 118: why not validate the result at the entrance?
https://github.com/karmada-io/karmada/blob/2220124b4d273f8a8415765d9992ab0670871e9d/pkg/resourceinterpreter/interpreter.go#L250-L278

for example
```go
 func (i *customResourceInterpreterImpl) GetDependencies(object *unstructured.Unstructured) (dependencies []configv1alpha1.DependentObjectReference, err error) { 
 	defer func() {
 	 	if err == nil && hookEnabled {
 	 	 	err = validate.VerifyResponseDependencies(dependencies)
 	 	}
 	}

 	... ...
```

pkg/resourceinterpreter/customized/validate/validate.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"fmt"
+
+	configv1alpha1 "github.com/karmada-io/karmada/pkg/apis/config/v1alpha1"
+)
+
+// VerifyResponseDependencies verifies dependencies in the Response.
+func VerifyResponseDependencies(dependencies []configv1alpha1.DependentObjectReference) error {
+	for index, dependency := range dependencies {
+		if len(dependency.Name) == 0 && dependency.LabelSelector == nil {
```
whitewindmills, line 28: why not validate other fields(`APIVersion`, `Kind`)?

whitewindmills: I see they are required but no component is responsible for validating them.
https://github.com/karmada-io/karmada/blob/2220124b4d273f8a8415765d9992ab0670871e9d/pkg/apis/config/v1alpha1/interpretercontext_types.go#L156-L162

whitewindmills: I don't think the json deserialization can do that. maybe you can test it.

XiShanYongYe-Chang: Let me make a test.

XiShanYongYe-Chang: @whitewindmills you are right:

image

XiShanYongYe-Chang: Updated. Can you help review again?

whitewindmills:
thanks for your quick response.
/lgtm

chaunceyjiang (commented):
/lgtm

RainbowMango (commented):
Some nits, and a question:
Why not verify the dependencies returned from the default interpreters?

XiShanYongYe-Chang:
> Why not verify the dependencies returned from the default interpreters?

I think that the return body of external requests should be verified. The default interpreter does not involve external requests. It should be determined that the code is normal when it is written.

RainbowMango:
@chaunceyjiang what's your opinion on this?

chaunceyjiang:
> Why not verify the dependencies returned from the default interpreters?


Good question! I believe third-party interpreters also need this.

RainbowMango:
Yes, that's exactly the reason why I'm asking.

XiShanYongYe-Chang:
Validation for the InterpretDependency operation has been added for all interpreters.

pkg/resourceinterpreter/default/native/default.go
```diff
@@ -131,6 +132,12 @@ func (e *DefaultInterpreter) AggregateStatus(object *unstructured.Unstructured,
 
 // GetDependencies returns the dependent resources of the given object.
 func (e *DefaultInterpreter) GetDependencies(object *unstructured.Unstructured) (dependencies []configv1alpha1.DependentObjectReference, err error) {
+	defer func() {
+		if err != nil {
+			return
+		}
+		err = validation.VerifyDependencies(dependencies)
+	}()
```
RainbowMango, line 140: How about doing the check like this:
```golang
	dependencies, err = handler(object)
	if err != nil {
		return nil, err
	}

	if validation.VerifyDependencies(dependencies) != nil {
		return nil, err
	}

	return dependencies, nil
```

pkg/util/interpreter/validation/validation.go
RainbowMango, line 1: Shall we add some unit tests on this?

pkg/resourceinterpreter/default/native/default.go
XiShanYongYe-Chang: Updated.

RainbowMango (approved):
/lgtm
/approve

PR #4985, opened 2024-05-24
Skip schedule deleting bindings

whitewindmills:
**What type of PR is this?**
/kind feature

we should skip schedule deleting bindings cause it makes no sense.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -334,6 +334,12 @@ func (s *Scheduler) doScheduleBinding(namespace, name string) (err error) {
 		}
 		return err
 	}
+	if !rb.DeletionTimestamp.IsZero() {
+		s.recordScheduleResultEventForResourceBinding(rb, nil, fmt.Errorf("skip schedule deleting resourceBinding: %s/%s", rb.Namespace, rb.Name))
+		klog.V(4).InfoS("Skip schedule deleting ResourceBinding", "ResourceBinding", klog.KObj(rb))
+		return nil
+	}
+
 	rb = rb.DeepCopy()
 
 	if rb.Spec.Placement == nil {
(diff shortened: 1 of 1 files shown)


pkg/scheduler/scheduler.go
```diff
@@ -334,6 +334,12 @@ func (s *Scheduler) doScheduleBinding(namespace, name string) (err error) {
 		}
 		return err
 	}
+	if !rb.DeletionTimestamp.IsZero() {
+		s.recordScheduleResultEventForResourceBinding(rb, nil, fmt.Errorf("skip schedule deleting resourceBinding: %s/%s", rb.Namespace, rb.Name))
+		klog.V(4).Info("Skip schedule deleting resourceBinding", "resourceBinding", klog.KObj(rb))
```
RainbowMango, line 339:
```suggestion
		klog.V(4).InfoS("Skip schedule deleting ResourceBinding", "ResourceBinding", klog.KObj(rb))
```

pkg/scheduler/scheduler.go
```diff
@@ -398,6 +404,12 @@ func (s *Scheduler) doScheduleClusterBinding(name string) (err error) {
 		}
 		return err
 	}
+	if !crb.DeletionTimestamp.IsZero() {
+		s.recordScheduleResultEventForClusterResourceBinding(crb, nil, fmt.Errorf("skip schedule deleting clusterResourceBinding: %s", crb.Name))
+		klog.V(4).Info("Skip schedule deleting clusterResourceBinding", "clusterResourceBinding", klog.KObj(crb))
```
RainbowMango, line 409:
```suggestion
		klog.V(4).InfoS("Skip schedule deleting ClusterResourceBinding", "ClusterResourceBinding", klog.KObj(crb))
```

whitewindmills:
@RainbowMango PTAL

pkg/scheduler/scheduler.go
```diff
@@ -334,6 +334,12 @@ func (s *Scheduler) doScheduleBinding(namespace, name string) (err error) {
 		}
 		return err
 	}
+	if !rb.DeletionTimestamp.IsZero() {
+		s.recordScheduleResultEventForResourceBinding(rb, nil, fmt.Errorf("skip schedule deleting resourceBinding: %s/%s", rb.Namespace, rb.Name))
```
RainbowMango, line 338: Should we emit an event for this scenario?
I suppose, no one would expect the ResourceBing would be scheduled during deletion phase, right?
So, why do they need a warning event?

Sorry, I didn't realize it during last review.

whitewindmills: maybe the administrator didn't expect that his binding was accidentally deleted by others, so I think it is more reasonable to emit an event.
in addition, referring to the design of k8s in this scenario.
https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/schedule_one.go#L370

RainbowMango: Thanks, I'm sold.

RainbowMango (approved):
/lgtm
/approve

PR #4993, opened 2024-05-28
Propose binding priority and preemption

### LINKED ISSUE #4938: [Scheduler] about priority scheduling
The current scheduler only schedules resources in order, that is, the resources triggered first can be scheduled first. They are stored in a first-in-first-out queue. This seems too simplistic at the moment, maybe it's time to introduce priority-based scheduling behavior.

**Why is this needed**:
In large-scale clusters, scheduling responses can become slow, An example use case is for data science workloads. A user may submit a job that they want to be prioritized above other workloads, but do not wish to affect existing work. The high priority resourceBinding will be scheduled ahead of other queued resourceBindings, as soon as sufficient cluster resources "naturally" become free.
priority-scheduling

whitewindmills:
**What type of PR is this?**
/kind design

Propose binding priority and preemption

Fixes #4938

```release-note
NONE
```


docs/proposals/scheduling/binding-priority-preemption/README.md
```diff
@@ -0,0 +1,340 @@
...
+Preemption should happen only when a member cluster does not have enough resources for a binding.
+In such cases, preemption happens only when the priority of the pending binding (preemptor) is higher than the preempted bindings.
+Preemption must not happen when there is no pending binding, or when the pending bindings have equal or lower priority than the preempted bindings.
+If preemption happens in such scenarios, please fell free to file an issue.
+
+## Design Details
+
+### Effect of priority on scheduling
+
+One could generally expect a binding with higher priority has a higher chance of getting scheduled than the same binding with lower priority.
```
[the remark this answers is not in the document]
whitewindmills: > the same binding Is it a description error?

no, this means that the binding has the same input conditions except for the priority.

> How to process bindings with the same priority?

queue up according to the time of joining the queue.

docs/proposals/scheduling/binding-priority-preemption/README.md
```diff
@@ -0,0 +1,340 @@
...
+However, there are other parameters that affect scheduling decisions. So, a high priority binding may or may not be scheduled before lower priority bindings.
```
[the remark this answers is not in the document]
whitewindmills: > What does other parmaters mean and what is the impact on the current description?

as the example says: `For example, a high priority binding is marked a....`

> Will a binding already in progress be interrupted by a higher priority binding?

never, scheduling is a serial job

docs/proposals/scheduling/binding-priority-preemption/README.md
```diff
@@ -0,0 +1,340 @@
...
+For example, a high priority binding is marked as unschedulable due to insufficient resources of member clusters, but the preemption behavior is not enabled.
+Karmada will try to schedule others with lower priority.
+
+### Effect of priority on preemption
+
+Generally, lower priority bindings are more likely to get preempted by higher priority bindings when member clusters have reached a threshold.
```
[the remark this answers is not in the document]
whitewindmills: when a binding is unschedulable and thas a higher priority than other bindings.

RainbowMango:
Hi @whitewindmills I guess it's a good chance to introduce this proposal at tomorrow's community meeting, what do you think?

docs/proposals/scheduling/binding-priority-preemption/README.md
```diff
@@ -0,0 +1,340 @@
...
+By default, preemption is disabled.
+
+```go
+// ResourceBindingSpec represents the expectation of ResourceBinding.
+type ResourceBindingSpec struct {
+    // The priority value. The karmada-scheduler component use this field to find the
+    // priority of the binding.
+    // The higher the value, the higher the priority.
+    // +optional
+    Priority *int32 `json:"priority,omitempty"`
```
chaunceyjiang, line 212: If the `PriorityClass` is deleted during the `ResourceBinding` scheduling process, will this field be cleared?

whitewindmills: no, just keep it.

docs/proposals/scheduling/binding-priority-preemption/README.md
```diff
@@ -0,0 +1,340 @@
...
+
+### Preemption order
+
+When scheduling a pending binding, scheduler tries to place the binding on a member cluster that does not require preemption.
+If there is no such a member cluster, scheduler may favor a member cluster where the number and/or priority of victims (preempted bindings) is smallest.
+After choosing the member cluster, scheduler considers the lowest priority bindings for preemption first.
+Scheduler starts from the lowest priority and considers enough bindings that should be preempted to allow the pending binding to schedule.
+Scheduler only considers bindings that have lower priority than the pending binding.
+
+Scheduler will try to minimize the number of preempted bindings.
```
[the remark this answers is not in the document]
whitewindmills: I have to say this is a point worth noting. but everything has its pros and cons, and we can only seek the way with the least negative impact. overall, reducing the number of preempted bindings has the least impact on the entire system.

cc @RainbowMango @XiShanYongYe-Chang @chaunceyjiang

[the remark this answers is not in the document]
wu0407: find the first  resource bindings requests that meet the needs requirements.
Maybe we can refer to this implementation.

> // getPodsToPreemptByDistance finds the pods that have pod requests >= admission requirements.
> // Chooses pods that minimize "distance" to the requirements.
> // If more than one pod exists that fulfills the remaining requirements,
> // it chooses the pod that has the "smaller resource request"
> // This method, by repeatedly choosing the pod that fulfills as much of the requirements as possible,
> // attempts to minimize the number of pods returned.

https://github.com/kubernetes/kubernetes/blob/85fad5b874c7723fd1811c0ecc2bf56059c538f1/pkg/kubelet/preemption/preemption.go#L134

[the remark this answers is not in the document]
whitewindmills: thanks, I will research it.

whitewindmills:
hi all, can we continue this proposal?

Monokaix:
> hi all, can we continue this proposal?

+1

docs/proposals/scheduling/binding-priority-preemption/README.md
```diff
@@ -0,0 +1,340 @@
...
+When a new binding has certain scheduling requirements that makes it infeasible on any member cluster,
+scheduler may choose to clear the scheduling result of lower priority bindings to satisfy the scheduling requirements of the new binding.
+We call this operation "binding preemption". Binding preemption is distinguished from "[policy preemption](../policy-preemption/README.md)" where high-priority policies preempt low-priority policies.
+
+### User Stories (Optional)
+
+#### As a user, I want to the task I submitted to be scheduled as soon as possible.
+
+I submit a workload that is prioritized above other workloads, but do not wish to discard existing work by preempting scheduled bindings.
+I hope that the high priority workload will be scheduled ahead of other queued bindings, as soon as sufficient cluster resources "naturally" become free.
```
RainbowMango, line 66: In case of a workload with higher priority can not be scheduled due to insufficient resources, will it block the others with lower priority?

1. If yes, it means no more workload could be scheduled until the one with the highest priority gets scheduled.
2. If not, the highest-priority workload might not be scheduled forever if low-priority workloads keep coming in.

Personally, I tend to option 1, which is exactly the user expectation I guess.

docs/proposals/scheduling/binding-priority-preemption/README.md
```diff
@@ -0,0 +1,340 @@
...
+### Risks and Mitigations
+
+Binding priority and preemption can have unwanted side effects. Here is an example of potential problems and ways to deal with them.
+
+#### Bindings are preempted unnecessarily
+
+Preemption removes existing bindings from some member clusters under resource pressure to make room for higher priority pending bindings.
+If you give high priorities to certain bindings by mistake, these unintentionally high priority bindings may cause preemption in your multi-cluster system.
+
+To address the problem, you can disable the preemption behavior for your bindings.
```
RainbowMango, line 87: Do you mean `disable preemption feature` across the system? Or disable preemption for particular ResourceBindings?

docs/proposals/scheduling/binding-priority-preemption/README.md
```diff
@@ -0,0 +1,340 @@
...
+
+When a binding is preempted, there will be events recorded for the preempted binding.
+Preemption should happen only when a member cluster does not have enough resources for a binding.
+In such cases, preemption happens only when the priority of the pending binding (preemptor) is higher than the preempted bindings.
+Preemption must not happen when there is no pending binding, or when the pending bindings have equal or lower priority than the preempted bindings.
+If preemption happens in such scenarios, please fell free to file an issue.
```
RainbowMango, line 93:
```suggestion
If preemption happens in such scenarios, please feel free to file an issue.
```
A typo :)

docs/proposals/scheduling/binding-priority-preemption/README.md
RainbowMango: I see @whitewindmills [answers below](https://github.com/karmada-io/karmada/blob/f1b5c033a3d06986de7b8c6ad88675fed62c2df4/docs/proposals/scheduling/binding-priority-preemption/README.md#effect-of-priority-on-scheduling):
> For example, a high priority binding is marked as unschedulable due to insufficient resources of member clusters, but the preemption behavior is not enabled. Karmada will try to schedule others with lower priority.

@LeonZh0u @wengyao04 What's your opinion on this?

docs/proposals/scheduling/binding-priority-preemption/README.md
```diff
@@ -0,0 +1,340 @@
...
+Generally, lower priority bindings are more likely to get preempted by higher priority bindings when member clusters have reached a threshold.
+In such a case, scheduler may decide to preempt lower priority bindings to release enough resources for higher priority pending bindings.
+As mentioned before, there are other parameters that affect scheduling decisions, such as cluster affinity and spread constraints.
+If scheduler determines that a high priority binding cannot be scheduled even if lower priority bindings are preempted, it will not preempt lower priority bindings.
+
+### New API
+
+#### Priority Classes
+
+This proposal are going to add a new API object called `PriorityClass`. The priority class defines the mapping between the priority name and its value.
```
RainbowMango, line 115: Do you mean to use the `scheduling.k8s.io/v1 PriorityClass` or introduce a brand new PriorityClass to Karmada API?

In my opinion, we can share the one from Kubernetes and don't have to introduce a new one.

@LeonZh0u [mentioned the concept of `federated priority class`](https://docs.google.com/document/d/1MixmgLwnmiRrukyFP25JvE2hqiqB3f_JKT12tXiOsbc/edit?disco=AAABUpFnU68). You both have similar ideas, so we can clarify them here.

whitewindmills: In order to facilitate the expansion of these APIs in the future, we'd better introduce a brand new PriorityClass to Karmada API.

docs/proposals/scheduling/binding-priority-preemption/README.md
```diff
@@ -0,0 +1,340 @@
...
+Binding priority and preemption can have unwanted side effects. Here is an example of potential problems and ways to deal with them.
+
+#### Bindings are preempted unnecessarily
+
+Preemption removes existing bindings from some member clusters under resource pressure to make room for higher priority pending bindings.
+If you give high priorities to certain bindings by mistake, these unintentionally high priority bindings may cause preemption in your multi-cluster system.
+
+To address the problem, you can disable the preemption behavior for your bindings.
+
+When a binding is preempted, there will be events recorded for the preempted binding.
```
seanlaii, line 89: Hi, I have a quick question regarding the behavior of preempted Bindings, specifically for Deployment templates.

From my understanding of the current implementation in this proposal, if a ResourceBinding with a Deployment template is preempted, it won't be automatically recreated. Is this correct?

docs/proposals/scheduling/binding-priority-preemption/README.md
LeonZh0u: Hi, I think this makes sense. should we make it a feature gate (e.g. strict preemption mode)?
or we can have another grpc request type for estimate maxreplica that accounts for resources tied up in preemption at the moment.

docs/proposals/scheduling/binding-priority-preemption/README.md
LeonZh0u: we are following @whitewindmills proposal's federated priority class. and will be ready for a demo on the priority workflow in our next community meeting.

docs/proposals/scheduling/binding-priority-preemption/README.md
whitewindmills: if there are enough resources, it will be recreated.

docs/proposals/scheduling/binding-priority-preemption/README.md
RainbowMango: Yeah, we do need a feature gate. But here we are talking about the behavior.
I agree with @whitewindmills that without preemption if scheduling fails due to insufficient resources(or can not find feasible clusters), we should continue attempting to schedule other pending ResourcesBindings. This behavior aligns with he current implementation.

docs/proposals/scheduling/binding-priority-preemption/README.md
RainbowMango: That makes sense.

docs/proposals/scheduling/binding-priority-preemption/README.md
RainbowMango: @LeonZh0u Please help to add an agenda the [Meeting Notes](https://docs.google.com/document/d/1y6YLVC-v7cmVAdbjedoyR5WL0-q45DBRXTvz5_I7bkA/edit).

@whitewindmills will you attend the next meeting? Probably next Tuesday 23:00 PM UTC+8.

docs/proposals/scheduling/binding-priority-preemption/README.md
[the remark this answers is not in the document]
RainbowMango: > Scheduler will try to minimize the number of preempted bindings.

I agree. But I think we'd better consider the priority first. That means ResourceBining with lower priority will be preempted first.

docs/proposals/scheduling/binding-priority-preemption/README.md
```diff
@@ -0,0 +1,340 @@
...
+```
+, scheduler will preempt the priority 2 binding only and leaves priority 1 and priority 0 running.
+
+### Preemption in multi-scheduler
+
+Karmada allows multiple schedulers to exist in its control plane, This introduces a race condition where multiple schedulers may perform round-robin preemption.
+That is, scheduler A may schedule binding A, Scheduler B preempts binding A to schedule binding B which is then preempted by scheduler A to schedule binding A and we go in a loop.
+
+Therefore, we strongly recommend not to enable preemption behavior in multiple schedulers.
+If you must enable it, unless you can ensure that they will not have any intersection during preemption.
```
RainbowMango, line 290: Good point!
I think each scheduler can only preempt ResourceBinding managed by itself.

docs/proposals/scheduling/binding-priority-preemption/README.md
whitewindmills: sorry, I may not be able to attend.

docs/proposals/scheduling/binding-priority-preemption/README.md
whitewindmills: yes, that’s the point.

seanlaii:
Hi all, 

Thank you all for the amazing feedback today! In summary, our design currently only targets the resource that is scheduled in a single cluster, and the preemption only happens in one cluster.

Here are the points we agree with in this proposal:
1. Need to redesign the **scheduler queue** so that it is aware of the priority and it will pop out the `resource binding` with highest priority first.
2. Reuse the native `priorityClass` and create a new API if we need to extend the functionality in the future.
3. Resolving the `priority` value and `preemptionPolicy/preemptionBehavior` to the resource binding so that the scheduler won’t need to query the `priorityClass` to find the values. 
4. If scheduling fails due to insufficient resources (or can not find feasible clusters), we should continue attempting to schedule other pending `ResourcesBindings` instead of blocking the whole scheduling process.
5. Reschedule the preempted `bindings`. (Need to carefully consider the backoff time for the preempted bindings to avoid reschedule the preempted bindings before the preemptor binding.) 

These are the points where we have different views or additional questions and thoughts:
1. Could you please clarify if there are any use cases that require binding the `priorityClassName` with the `PropagationPolicy/ClusterPropagationPolicy`? Maybe because you don't want to enforce that all the resources must have a `priorityClassName` filed?  We plan to put the `priority` and `preemptionBehavior/preemptionPolicy` in the `replicaRequirements`. So the user can just specify the `priorityClassName` in the resource and create the resource the same as what they do in the single cluster. We can use `custom resource interpreter` to propagate the `priority` and `preemptionPolicy` to the `replicaRequirements`.
3. We propose using `scheduler-estimators` to find the victim bindings since they have the accurate information of the clusters, and the scheduler will use the victim bindings to decide which bindings to preempt and perform preemption.

Thanks for your time again! Please feel free to provide any comment.
Reference: https://docs.google.com/document/d/1MixmgLwnmiRrukyFP25JvE2hqiqB3f_JKT12tXiOsbc/edit?tab=t.0
cc @RainbowMango @kevin-wangzefeng @wengyao04 @LeonZh0u

docs/proposals/scheduling/binding-priority-preemption/README.md
```diff
@@ -0,0 +1,286 @@
...
+```
+
+**Important notes**: Only one PriorityClass can be marked as `globalDefault`. However, if more than one PriorityClasses exists with their `globalDefault` field set to true, the
+smallest value of such global default PriorityClasses will be used as the default priority.
+
+### API change
+
+#### ResourceBinding/ClusterResourceBinding API change
+
+This proposal are going to add two new field: `Priority` and ` ` for determining the behavior for preempting.
```
seanlaii, line 136:
```suggestion
This proposal are going to add two new field: `Priority` and `PreemptionPolicy` for determining the behavior for preempting.
```

whitewindmills: @seanlaii  thanks

whitewindmills:
@RainbowMango can this proposal be merged?

RainbowMango:
I want to give it another look and see if any comments from the others.

docs/proposals/scheduling/binding-priority-preemption/README.md
```diff
@@ -0,0 +1,286 @@
...
+When a new binding has certain scheduling requirements that makes it infeasible on any member cluster,
+scheduler may choose to clear the scheduling result of lower priority bindings to satisfy the scheduling requirements of the new binding.
+We call this operation "binding preemption". Binding preemption is distinguished from "[policy preemption](../policy-preemption/README.md)" where high-priority policies preempt low-priority policies.
+
+### User Stories (Optional)
+
+#### As a user, I want to the task I submitted to be scheduled as soon as possible.
+
+I submit a workload that is prioritized above other workloads, but do not wish to discard existing work by preempting scheduled bindings.
+I hope that the high priority workload will be scheduled ahead of other queued bindings, as soon as sufficient cluster resources "naturally" become free.
```
RainbowMango, line 70:
```suggestion
I usually classify my workloads based on priority and want high-priority workloads to be scheduled ahead of others, rather than waiting to be scheduled in the order they were queued. This will allow me to maintain the quality of service for important workloads even during resource contention.
```

docs/proposals/scheduling/binding-priority-preemption/README.md
```diff
@@ -0,0 +1,290 @@
...
+
+```go
+// Placement represents the rule for select clusters.
+type Placement struct {
+    // PriorityClassName indicates bindings will use the PriorityClass to resolve it's priority and preemptionPolicy if specified.
+    // Any other name must be defined by creating a PriorityClass object with that name.
+    // If not specified, the binding will use the global default PriorityClass first.
+    // If the global default API does not exist, the binding priority will be zero, binding preemptionPolicy will be "Never".
+    // +optional
+    PriorityClassName string `json:"priorityClassName,omitempty"`
```
RainbowMango, line 175: 1. I guess we can present a complete design in this proposal, and notes, currently we only implement xxx.
I mean we can have a field `PriorityClassSource`, the valid options could be `FederatedPriorityClass`,`KubePriorityClass`,`PodPriorityClass`, defaults to `KubePriorityClass`.
Note that, we might don't add this filed when implementing this feature at alpha maturity.

2. Another thing open for discussion is where the filed should be placed, under `spec.placement` or `spec`.

whitewindmills: sounds good, I know that `KubePriorityClass` should be `priorityclasses.scheduling.k8s.io`, but what's `PodPriorityClass`?

RainbowMango: The `PriorityClassSource` represents where to find the priority class, there will be 3 ways:
- From the FederatedPriorityClass, which Karmada might introduce in the future.
- From the Kubernetes PriorityClass
- From the [Pod template](https://github.com/kubernetes/kubernetes/blob/16e8911fdccbdb348a0b3291acf6102494494f9b/staging/src/k8s.io/api/core/v1/types.go#L3887-L3894)

So, the `PodPriorityClass` means parsing the priority from the pod template.

whitewindmills: so, you mean, for CRDs, we still need a [ResourceInterpreter interface](https://github.com/karmada-io/karmada/blob/master/pkg/resourceinterpreter/interpreter.go#L39) to determine priority?
> From the [Pod template](https://github.com/kubernetes/kubernetes/blob/16e8911fdccbdb348a0b3291acf6102494494f9b/staging/src/k8s.io/api/core/v1/types.go#L3887-L3894)

and this rule requires the priority class must exist in Karmada control plane. so in my opinion, it is not much different from `KubePriorityClass`, and the API becomes more difficult to understand.
I don't know if I understand correctly. if not,  please correct me.

RainbowMango: > and this rule requires the priority class must exist in Karmada control plane.

I'm not sure where validates the existence of PriorityClass in Kubernetes, but karmada-apiserver doesn't have such a restriction. People can apply Deployment and specify a non-existent PriorityClass.  

Note: The idea of `FederatedPriorityClass` and `PodPriorityClass` is still not mature, just like a placeholder in this proposal, for extension purpose.

whitewindmills: > I'm not sure where validates the existence of PriorityClass in Kubernetes

not K8s, I mean Karmada should do this. that's weird - Karmada can find the priority class name from the Pod template, but this priority class doesn't exist.

RainbowMango: The PriorityClass is a kind of cluster-lever configuration, it is usually managed by the infra team, and it might synced from the legacy metadata system. In other words, that configuration might not propagated by Karmada.

By the way, Kubernetes doesn't require the PriorityClass to exist. I just ran a test with it, it shows that everything works fine:
```yaml
-bash-5.0# karmadactl get deployments.apps --operation-scope=members --clusters=member1 nginx -o yaml 
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx
  namespace: default
  resourceVersion: "1428701"
  uid: 0215af9b-d8e8-44e1-bda4-92f574be2cc4
spec:
    # ...
    spec:
      containers:
      - image: nginx
      priorityClassName: horen  // not-exist PriorityClass
```

whitewindmills: @RainbowMango first, sorry for the slow response.
I mean, Karmada should should understand the priority class to apply it to ResourceBinding.
as your example shows, the deployment `nginx` did work fine, but it cannot co-work with ResourceBinding priority & preemption because the priority class `horen` doesn't exists in Karmada.
in plain words, when people set the field `PriorityClassSource` to `PodPriorityClass`, Karmada will parse the priority class name from the pod template, then find the priority class by name to apply it to ResourceBinding.

docs/proposals/scheduling/binding-priority-preemption/README.md
```diff
@@ -0,0 +1,290 @@
...
+kind: PriorityClass
+metadata:
+  name: foo
+globalDefault: true
+preemptionPolicy: PreemptLowerPriority
+value: 2000000000
+```
+
+**Important notes**: Only one PriorityClass can be marked as `globalDefault`. However, if more than one PriorityClasses exists with their `globalDefault` field set to true, the
+smallest value of such global default PriorityClasses will be used as the default priority.
```
zhzhuang-zju, line 134: > However, if more than one PriorityClasses exists with their `globalDefault` field set to true, the
smallest value of such global default PriorityClasses will be used as the default priority.

[Only one native k8s PriorityClass with `globalDefault` set to true can exist in the system](https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/#priorityclass). Can Karmada directly inherit this limitation?

```bash
Error from server (Forbidden): error when creating "/tmp/template/priority/globalDefault_bak.yaml": priorityclasses.scheduling.k8s.io "high-priority-bak" is forbidden: PriorityClass high-priority is already marked as default. Only one default can exist
```

docs/proposals/scheduling/binding-priority-preemption/README.md
RainbowMango: OK, then please ignore the PodPriorityClass.

zhzhuang-zju: > Kubernetes doesn't require the PriorityClass to exist. I just ran a test with it, it shows that everything works fine:

When the Deployment `nginx` is propagated to the member clusters, it will fail to create Pods due to not finding the specified PriorityClass.

```yaml
status:
  conditions:
  - lastTransitionTime: "2025-01-06T07:50:17Z"
    lastUpdateTime: "2025-01-06T07:50:17Z"
    message: 'pods "nginx-7bcd4f75b7-" is forbidden: no PriorityClass with name horen
      was found'
    reason: FailedCreate
    status: "True"
    type: ReplicaFailure
```

Furthermore, if directly deploy a Pod that specifies a non-existent PriorityClass, it will result in an error immediately.
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: nginx
  labels:
    env: test
spec:
  containers:
  - name: nginx
    image: nginx
    imagePullPolicy: IfNotPresent
  priorityClassName: test-priority
```

```bash
$ karmadactl apply -f test-pod.yaml
Error from server (Forbidden): error when creating "test-pod.yaml": pods "nginx" is forbidden: no PriorityClass with name test-priority was found
```

docs/proposals/scheduling/binding-priority-preemption/README.md
RainbowMango: Where is this validation performed? @zhzhuang-zju

zhzhuang-zju: https://github.com/kubernetes/kubernetes/blob/8f8c94a04d00e59d286fe4387197bc62c6a4f374/plugin/pkg/admission/priority/admission.go#L206-L225

zhzhuang-zju: It's a admission plugins, and can be disabled by setting `- --disable-admission-plugins=Priority`

whitewindmills: yes, I've tested it, apiserver did validate it. and the same in Karmada. since this parameter is configured in the apiserver, so we cannot control whether users turn on this plugin. BTW, it's enabled by default.

zhzhuang-zju: make sense

docs/proposals/scheduling/binding-priority-preemption/README.md
RainbowMango: > Another thing open for discussion is where the filed should be placed, under spec. placement or spec.

Maybe we can introduce a field named `SchedulePriority` to PropagationPolicy.spec as I think `speck.placement` indicates where to put the workloads. The definition of `SchedulePriority` like(based on the design in this proposal):

```golang
// SchedulePriority represents the scheduling priority of workloads.
type SchedulePriority struct {
	// PriorityClassSource indicates where to find the PriorityClass definition.
	// Valid options are:
	// - KubePriorityClass: refers to Kubernetes PriorityClass with API version scheduling.k8s.io/v1.
	// - PodPriorityClass: refers to PodTemplate(PodSpec.PriorityClassName)
	// - FederatedPriorityClass: refers to Karmada FederatedPriorityClass with the latest API version.
	//
	// +kubebuilder:default="KubePriorityClass"
	// +kubebuilder:validation:Enum=FederatedPriorityClass;KubePriorityClass;PodPriorityClass
	// +required
	PriorityClassSource PriorityClassSource `json:"priorityClassSource,omitempty"`

	// PriorityClassName is the name of the PriorityClass that will be used to resolve the workload's
	// priority and preemption policy. The behavior of this field depends on the value of PriorityClassSource:
	//
	// - If PriorityClassSource is KubePriorityClass:
	//   This field should specify the name of a Kubernetes PriorityClass, if not specified, the workload will
	//   use the global default PriorityClass, if the global default PriorityClass does not exist, the workload's
	//   priority will be set to 0, and its preemption policy will be "Never".
	// - If PriorityClassSource is PodPriorityClass:
	//   This field should be omitted, and the workload will use the name declared in PodTemplate, if the referenced
	//   PriorityClass does not exist, will use the global default, if the global default PriorityClass does not
	//   exist, the workload's priority will be set to 0, and its preemption policy will be "Never".
	// - If PriorityClassSource is FederatedPriorityClass:
	//   This field must specify the name of a FederatedPriorityClass (currently undefined but may be implemented
	//   in future releases).
	// +optional
}

type PriorityClassSource string

const (
	// FederatedPriorityClassSource indicates that the workload uses a custom FederatedPriorityClass
	// as the source of PriorityClass.
	// Note: FederatedPriorityClass is currently undefined and may be implemented in future releases.
	FederatedPriorityClassSource PriorityClassSource = "FederatedPriorityClass"

	// KubePriorityClassSource indicates that the workload uses the Kubernetes PriorityClass
	// as the source of PriorityClass.
	KubePriorityClassSource PriorityClassSource = "KubePriorityClass"

	// PodPriorityClassSource indicates that the workload uses the PriorityClass declared in PodTemplate.
	PodPriorityClassSource PriorityClassSource = "PodPriorityClass"
)
```

seanlaii: Sounds good to me. Thanks!
I agree that we should put `SchedulePriority` outside of the `placement`, and have `PriorityClassSource` for future enhancement.

docs/proposals/scheduling/binding-priority-preemption/README.md
```diff
@@ -0,0 +1,312 @@
...
+With this proposal, a priority scheduling queue will be implemented to replace the current work queue.
+It should retain all the functionality of the current work queue and additionally implement sorting by priority.
+
+In addition, if preemption behavior is enabled, the scheduler should identify the circumstances under which the pending binding should preempt other bindings,
+calculate which bindings need to be preempted to meet current resource requirements.
+
+#### karmada-webhook
+
+Since we have strictly defined the binding priority and preemption, so the webhook should perform extra
+validation work to prevent misleading configuration.
```
seanlaii, line 302: Could you please clarify the intended responsibilities of the webhook?
Is it designed to restrict users from updating the federated priority class (a potential future enhancement) or to prevent modifications to the priority of the `resourceBinding`?
For the Kubernetes priorityClass, there is already a restriction.

whitewindmills: yes, I think so.

docs/proposals/scheduling/binding-priority-preemption/README.md
whitewindmills: I'll update it today.

seanlaii: Thanks, I will also modify the API accordingly.

RainbowMango: No rush. We are currently in the process of reviewing, and the content is subject to change.

docs/proposals/scheduling/binding-priority-preemption/README.md
```diff
@@ -0,0 +1,325 @@
...
+    // +optional
+    Priority *int32 `json:"priority,omitempty"`
+
+    // PreemptionBehavior is the Policy for preempting bindings with lower priority.
+    // One of Never, PreemptLowerPriority.
+    // Defaults to Never if unset.
+    // +optional
+    PreemptionBehavior PreemptionBehavior `json:"preemptionBehavior,omitempty"`
+
+    ... ...
```
RainbowMango, line 157: Regarding the API change on ResourceBinding, I propose to introduce a composited field to hold all priority and preemption-related configurations(The ResourceBinding gets bigger and bigger, I'd like to classify fields a little bit.). Like:
```golang
	// SchedulePriority represents the scheduling priority of workloads.
	// +optional
	SchedulePriority *SchedulePriority `json:"schedulePriority,omitempty"`
```
Note that this filed is optional.

The definition would be:
```golang
// SchedulePriority represents the scheduling priority of workloads.
type SchedulePriority struct {
	// Priority is the priority value that will be used by the scheduler to sort bindings.
	// The higher the value, the higher the priority.
	// +required
	Priority int32 `json:"priority,omitempty"`

	// PreemptionBehavior describes the behavior of if and when to preempt bindings.
	// One of Never, PreemptLowerPriority.
	// +kubebuilder:validation:Enum=Never;PreemptLowerPriority
	// +required
	PreemptionBehavior PreemptionBehavior `json:"preemptionBehavior,omitempty"`
}
```
Let me know what you think of it.

RainbowMango: By the way, can you move the `ResourceBinding/ClusterResourceBinding API change` behind of `PropagationPolicy/ClusterPropagationPolicy API change`, that would help people to understand this change.

whitewindmills: +1

RainbowMango (commented):
I don't have any other comments regarding the priority scheduling part so far except the feature gate name.

Next, I would like to ask @kevin-wangzefeng to review this design again. If no further comments, I think we can start coding. Additionally, the preemption part is relatively complex and needs further discussion.

docs/proposals/scheduling/binding-priority-preemption/README.md
```diff
@@ -0,0 +1,333 @@
...
+    // Priority is the priority value that will be used by the scheduler to sort bindings.
+    // The higher the value, the higher the priority.
+    // +required
+    Priority int32 `json:"priority,omitempty"`
+
+    // PreemptionBehavior describes the behavior of if and when to preempt bindings.
+    // One of Never, PreemptLowerPriority.
+    // +kubebuilder:validation:Enum=Never;PreemptLowerPriority
+    // +required
+    PreemptionBehavior PreemptionBehavior `json:"preemptionBehavior,omitempty"`
```
seanlaii, line 219: Should we name it `preemptionPolicy` according to line 194?

RainbowMango: Good question!
@whitewindmills Can you explain why not name it with `PreemptionPolicy` in the first place?

whitewindmills: thanks, I just forgot to update it.

docs/proposals/scheduling/binding-priority-preemption/README.md
```diff
@@ -0,0 +1,333 @@
...
+2. If a binding cannot meet the resource requirement of the pending binding,
+we don't stop accumulating its resource requirements until the resource requirements of the pending binding are met.
+3. If there is no binding can meet the resource requirements of the pending binding, no preemption will be executed when the preemption loop ends.
+So we'll try to preempt the accumulated bindings to meet the resource requirements of the pending binding, if can't, just fail.
+
+### Feature gate
+
+This binding preemption feature is an experimental feature, so we will introduce the following feature gates, which are disabled by default.
+- ResourceBindingPreemption: indicates if a high-priority binding could preempt a low-priority binding, it's a global enablement.
+- CrossNamespaceResourceBindingPreemption: indicates if a high-priority binding could preempt a low-priority binding across namespaces.
```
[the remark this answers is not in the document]
whitewindmills: do you mean we need a feature gate for priority scheduling?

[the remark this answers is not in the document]
RainbowMango: I mean we just need one feature gate which controls:
- Priority-based scheduling (include both RB/CRB)
- Priority-based preemption (include both RB/CRB)

Given that scheduling and preemption are highly relevant features, they should be controlled by one feature gate.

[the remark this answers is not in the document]
RainbowMango: So, how many feature gates are in your mind? 
The `ResourceBindingPreemption` and `CrossNamespaceResourceBindingPreemption` are both used for preemption.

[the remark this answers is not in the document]
whitewindmills: I think it is not necessary to add feature gate for priority scheduling. for preemption, there are three scenarios:
- S1: ResourceBinding Preemption within the same namespace
- S2: ResourceBinding Preemption across namespaces
- S3: ClusterResourceBinding Preemption

so,  we do need a global feature gate to control all preemption scenarios, maybe name it `ResourceBindingPreemption`? I'm not sure now. if you have any better name, just tell me.
first, S1 is the basic function of preemption, so we don't need an additional feature gate. 
then for S2 & S3, I think they are more risky and we may need additional feature gates, like `CrossNamespaceResourceBindingPreemption`. 
_Note: I don't create a feature gate for S3 because I've claimed we don't support it now._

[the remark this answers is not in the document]
RainbowMango: > I think it is not necessary to add feature gate for priority scheduling.

Given that priority scheduling would introduce significant changes to the scheduler, we can't 100% ensure the compatibility, I think it's a must to have a feature gate for it.

In addition, the feature gate name should be user-oriented, so I'd prefer to avoid the word `ResourceBinding ' from it.

I can understand your concern about this being a huge feature that you want to iterate faster, I can accept two feature gates for priority scheduling and preemption respectively.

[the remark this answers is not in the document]
whitewindmills: okay, but I don't have suitable names for them yet. I think you may have more professional suggestions.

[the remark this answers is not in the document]
RainbowMango: How about `PriorityBasedScheduling` and `PriorityBasedPreemptiveScheduling`?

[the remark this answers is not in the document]
whitewindmills: thanks, good names, I'll update this proposal.

RainbowMango (commented):
/lgtm

As I mentioned on https://github.com/karmada-io/karmada/pull/4993#pullrequestreview-2560374644, looks great for the priority scheduling part.

kevin-wangzefeng (commented):
The preemption part looks not yet complete. I discussed with @RainbowMango offline, and have the following suggestions:

1. find a place in the proposal to mark that the preemption part currently is for readers to easier understand prioirity based scheduling. We will need to improve the design when we start implementing preemption part.
2. add a section `implementation plan` to track the overall design and implementation plan, better align with target-feature-maturity in each releases as well.

whitewindmills:
@kevin-wangzefeng @RainbowMango PTAL

docs/proposals/scheduling/binding-priority-preemption/README.md
```diff
@@ -0,0 +1,355 @@
...
+
+In the current proposal, the preemptive scheduling feature is proposed to better understand the priority scheduling feature.
+Our design of preemptive scheduling is not complete and needs more iterations.
+
+### Implementation plan
+
+release 1.13: release priority scheduling feature, provide alpha version preview, feature gate is disable by default.
+release 1.14: complete the preemptive scheduling design and iterate the priority scheduling feature.
+release 1.15: release preemptive scheduling feature, provide alpha version preview, feature gate is disable by default.
+release 1.16+: we will continue to iterate on the above features based on user feedback. When a feature reaches maturity, the corresponding feature gate will be enabled by default.
```
RainbowMango, line 355:
```suggestion
- release 1.13: release priority scheduling feature, provide alpha version preview, feature gate is disable by default.
- release 1.14: complete the preemptive scheduling design and iterate the priority scheduling feature.
- release 1.15: release preemptive scheduling feature, provide alpha version preview, feature gate is disable by default.
- release 1.16+: we will continue to iterate on the above features based on user feedback. When a feature reaches maturity, the corresponding feature gate will be enabled by default.
```

RainbowMango (commented):
Generally looks good to me for the implementation plan.

docs/proposals/scheduling/binding-priority-preemption/README.md
whitewindmills: thanks, I forgot to set the format.

RainbowMango (approved):
/lgtm
/approve

PR #5008, opened 2024-05-31
Workflows such as e2e are not triggered when only the files under docs are changed.

zhzhuang-zju:
**What type of PR is this?**
/kind cleanup

Workflows such as e2e are not triggered when only the files under docs are changed.
Fixes #

When all the path names match patterns in `paths-ignore`, the workflow will not run. If any path names do not match patterns in `paths-ignore`, even if some path names match the patterns, the workflow will run.

```release-note
NONE
```

### DIFF
--- a/.github/workflows/ci-image-scanning.yaml
+++ b/.github/workflows/ci-image-scanning.yaml
@@ -5,6 +5,8 @@ on:
     # for PRs initiated by Dependabot.
     branches-ignore:
       - 'dependabot/**'
+    paths-ignore:
+      - 'docs/**'
 jobs:
   use-trivy-to-scan-image:
     name: image-scanning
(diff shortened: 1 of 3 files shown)


RainbowMango:
I'm hesitant to do that because the only benefit is that we can get the CI result faster.
But I always hoped the test would be run more frequently so, that we could find potential bugs.

liangyuanpeng:
This is what I have always wanted to do, the reason I didn't modify it is because I think the CI system is not under too much load at the moment.

XiShanYongYe-Chang:
> I'm hesitant to do that because the only benefit is that we can get the CI result faster.
> But I always hoped the test would be run more frequently so, that we could find potential bugs.

I agree with this point.

zhzhuang-zju:
Combining the recommendations of https://github.com/karmada-io/karmada/pull/5008#issuecomment-2141676867 and https://github.com/karmada-io/karmada/pull/5008#issuecomment-2141691180, when the CI system is not under heavy pressure, tests should be run as frequently as possible to detect occasional problems.
I agree with this opinion, especially since CI has occasionally failed recently. So I'll close this issue, maybe in the future when the CI system is overstressed, then reopen it.

/close

PR #5010, opened 2024-06-01
helm install karmada components in order

### LINKED ISSUE #4917: [Flaking Test] chart-lint-test not stable recently
#### Which jobs are flaking:
chart-lint-test

See the logs below:

- https://github.com/karmada-io/karmada/actions/runs/8982272797/job/24669506880 (found on master after [#4860](https://github.com/karmada-io/karmada/pull/4860))
- https://github.com/karmada-io/karmada/actions/runs/8997316919/job/24715232276 (found on master after [#4883](https://github.com/karmada-io/karmada/pull/4883))
- https://github.com/karmada-io/karmada/actions/runs/8948612862/job/24581937332 (found on master after [#4887](https://github.com/karmada-io/karmada/pull/4887))
- https://github.com/karmada-io/karmada/actions/runs/8997316919/job/24715232276 (found on master after [#4883](https://github.com/karmada-io/karmada/pull/4883))
- https://github.com/karmada-io/karmada/actions/runs/9153421301/job/25162342796 (found on master after [#4949](https://github.com/karmada-io/karmada/pull/4949), at 20/05/2024.)

#### Reason for failure:
TBD

chaosi-zju:
**What type of PR is this?**

/kind feature
/kind flake

helm install karmada components in order, for detailed discussion, see issue #4917

Fixes #4917

```release-note

```

### DIFF
--- a/charts/karmada/templates/_helpers.tpl
+++ b/charts/karmada/templates/_helpers.tpl
@@ -306,6 +306,10 @@ app: {{- include "karmada.name" .}}-search
 {{- include "karmada.commonLabels" . -}}
 {{- end -}}
 
+{{- define "karmada.staticResourceJob.labels" -}}
+{{- include "karmada.commonLabels" . -}}
+{{- end -}}
+
 {{- define "karmada.postInstallJob.labels" -}}
 {{- include "karmada.commonLabels" . -}}
 {{- end -}}
@@ -573,4 +577,56 @@ Return the proper Docker Image Registry Secret Names
 - --{{ $key }}={{ $value }}
 {{- end }}
 {{- end }}
+{{- end -}}
+
+{{- define "karmada.init-sa-secret.volume" -}}
+{{- $name := include "karmada.name" . -}}
+- name: init-sa-secret
+  secret:
+    secretName: {{ $name }}-hook-job
+{{- end -}}
+
+{{- define "karmada.init-sa-secret.volumeMount" -}}
+- name: init-sa-secret
+  mountPath: /opt/mount
+{{- end -}}
+
+{{- define "karmada.initContainer.build-kubeconfig" -}}
+TOKEN=$(cat /opt/mount/token)
+kubectl config set-cluster karmada-host --server=https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT} --certificate-authority=/opt/mount/ca.crt
+kubectl config set-credentials default --token=$TOKEN
+kubectl config set-context karmada-host-context --cluster=karmada-host --user=default --namespace=default
+kubectl config use-context karmada-host-context
+{{- end -}}
+
+{{- define "karmada.initContainer.waitEtcd" -}}
+- name: wait
+  image: {{ include "karmada.kubectl.image" . }}
+  imagePullPolicy: {{ .Values.kubectl.image.pullPolicy }}
+  command:
+    - /bin/sh
+    - -c
+    - |
+      bash <<'EOF'
+      {{- include "karmada.initContainer.build-kubeconfig" . | nindent 6 }}
+      kubectl rollout status statefulset etcd -n {{ include "karmada.namespace" . }}
+      EOF
+  volumeMounts:
+    {{- include "karmada.init-sa-secret.volumeMount" .| nindent 4 }}
+{{- end -}}
+
+{{- define "karmada.initContainer.waitStaticResource" -}}
+- name: wait
+  image: {{ include "karmada.kubectl.image" . }}
+  imagePullPolicy: {{ .Values.kubectl.image.pullPolicy }}
+  command:
+    - /bin/sh
+    - -c
+    - |
+      bash <<'EOF'
+      {{- include "karmada.initContainer.build-kubeconfig" . | nindent 6 }}
+      kubectl wait --for=condition=complete job {{ include "karmada.name" . }}-static-resource -n {{ include "karmada.namespace" . }}
+      EOF
+  volumeMounts:
+    {{- include "karmada.init-sa-secret.volumeMount" .| nindent 4 }}
 {{- end -}}
--- a/charts/karmada/templates/karmada-aggregated-apiserver.yaml
+++ b/charts/karmada/templates/karmada-aggregated-apiserver.yaml
@@ -29,6 +29,8 @@ spec:
     spec:
       {{- include "karmada.aggregatedApiServer.imagePullSecrets" . | nindent 6 }}
       automountServiceAccountToken: false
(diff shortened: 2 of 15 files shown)


charts/karmada/templates/_helpers.tpl
```diff
@@ -574,3 +578,59 @@ Return the proper Docker Image Registry Secret Names
...
+
+{{- define "karmada.initContainer.build-kubeconfig" -}}
+TOKEN=$(cat /opt/mount/token)
+kubectl config set-cluster karmada-host --server=https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT} --certificate-authority=/opt/mount/ca.crt
+kubectl config set-credentials default --token=$TOKEN
+kubectl config set-context karmada-host-context --cluster=karmada-host --user=default --namespace=default
+kubectl config use-context karmada-host-context
+{{- end -}}
+
+{{- define "karmada.initContainer.waitETCD" -}}
```
calvin0327, line 602:
```suggestion
{{- define "karmada.initContainer.waitEtcd" -}}
```

The `ETCD` for etcd are a bit strange.

charts/karmada/values.yaml
```diff
@@ -128,6 +138,7 @@ certs:
       "*.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}",
       "*.{{ .Release.Namespace }}.svc",
       "localhost",
+      "172.18.0.2",
```
calvin0327, line 141: why need to add this `Alt`?

charts/karmada/values.yaml
```diff
@@ -38,6 +38,11 @@ podDisruptionBudget: &podDisruptionBudget {}
 ## "component" means install selected components in the control-cluster
 installMode: "host"
 
+## @param installInOrder
+## ture means install karmada components in order based on their dependencies
+## false means the installation would ignore the components order and just rely on its reconciliation
+installInOrder: true
```
calvin0327, line 44: At any time, we need to ensure the order of installation, this parameter seems like it should always be true, so it feels like we don't need it.

charts/karmada/templates/post-install-job.yaml
```diff
@@ -38,6 +38,7 @@ spec:
       nodeSelector:
       {{- toYaml . | nindent 8 }}
       {{- end }}
+      serviceAccountName: {{ $name }}-pre-job
```
calvin0327, line 41: Using the serviceAccount  `{{ $name }}-pre-job` in the post hook job doesn't seem very reasonable. can we modify `{{ $name }}-pre-job` to `{{ $name }}-hook-job`?

charts/karmada/templates/karmada-aggregated-apiserver.yaml
```diff
@@ -29,6 +29,8 @@ spec:
     spec:
       {{- include "karmada.aggregatedApiServer.imagePullSecrets" . | nindent 6 }}
       automountServiceAccountToken: false
+      initContainers:
```
calvin0327, line 32: `karmada-aggregated-apiserver` should wait for `karmada-apiserver`?

Or we can skip adding the initContainer, because the issues it causes are not significant.

charts/karmada/values.yaml
chaosi-zju: sorry, my fault

charts/karmada/templates/post-install-job.yaml
chaosi-zju: I also want to change a name

charts/karmada/templates/karmada-aggregated-apiserver.yaml
chaosi-zju: `agg-apiserver` really does need to wait for `apiserver`, otherwise it will be failed with following errors:

```console
Defaulted container "karmada-53dujqro77-aggregated-apiserver" out of: karmada-53dujqro77-aggregated-apiserver, wait (init)
I0603 03:38:38.500024       1 options.go:97] karmada-aggregated-apiserver version: version.Info{GitVersion:"v1.10.0-preview4-179-g9c85fbddf", GitCommit:"9c85fbddf914f4af7a05ad7611e12e2e73ffdea1", GitTreeState:"clean", BuildDate:"2024-05-30T03:23:07Z", GoVersion:"go1.21.10", Compiler:"gc", Platform:"linux/amd64"}
E0603 03:38:39.916483       1 run.go:74] "command failed" err="unable to load configmap based request-header-client-ca-file: Get \"https://karmada-53dujqro77-apiserver.karmada-53dujqro77.svc.cluster.local:5443/api/v1/namespaces/kube-system/configmaps/extension-apiserver-authentication\": dial tcp 100.12.20.194:5443: connect: connection refused"
```

charts/karmada/values.yaml
chaosi-zju: my concern is that our method is using a `init-container`, is it possible that users can accept the restart instead of our customized init container?

chaosi-zju:
Now, our `Chart Lint test` CI just only runs 3 minutes, while previously need almost 17 minute because compnonets restart for many times.

previous:
image

now:
image

liangyuanpeng:
> Now, our Chart Lint test CI just only runs 3 minutes, while previously need almost 17 minute because compnonets restart for many times.

it's a big promote!

charts/karmada/values.yaml
calvin0327: I haven't thought of scenarios where users wouldn't accept it, because it's a more user-friendly approach compared to the constant restarting method. we can look into other people's thoughts on this?

chaosi-zju: OK, we can hang on this comment as unresolved for some time. If nobady think it is useful, we can remove this `installInOrder`. For future, if somebody truly has this need, we then add it back.

chaosi-zju:
CC @RainbowMango

RainbowMango (commented):
@a7i Can you help to take a look?
The issue we are trying to address is that the components launched by Helm don't have a solid order, which will lead to components crashing during the installation, such as karmada-controller-manager crashes if it starts before the karmada-apiserver.

The approach of this PR is trying to add an init container to the components if it reply on others. I'm not sure if this is the best practice for Helm chart dealing with many components.

RainbowMango (approved):
/approve

a7i:
@RainbowMango apologies for the delay in my response as I was traveling for work.

I reviewed the linked Issue and the approach here makes sense. Additionally, in my organization, we don't use Helm so the solution proposed addresses it in a generic fashion that works for all.

Thank you @chaosi-zju for this!

PR #5015, opened 2024-06-02
update to go 1.22

grosser:
**What type of PR is this?**

/kind cleanup

Just to keep recent

```release-note
Karmada is now built with Go 1.22.4.
```


RainbowMango (commented):
Thanks for doing this.
We are going to bump Kubernetes v1.30 at this release(1.11), which requires Go 1.22. 

But, could you please also update the Go version at [README Prerequisites](https://github.com/karmada-io/karmada/blob/d8e67b7360dc888e22087f4c64b1ddc95a215965/README.md#prerequisites), as with this PR, we require go >= 1.22.3 now.

RainbowMango:
By the way, we also check the Go version at [hack/util.sh](https://github.com/karmada-io/karmada/blob/dee30c821b042984a40234f3291d2e034a6f7f57/hack/util.sh#L40), we probably don't need it anymore, since from Go1.21, people cannot build with lower Golang version than specified in go.mod.
It's up to you to clean it or not by this PR.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/karmada-io/karmada
 
-go 1.21.10
+go 1.22.3 // keep in sync with .go-version
```
liangyuanpeng, line 3: 1.22.4 is live now,how about  update version to 1.22.4, it's resolve a security issue (CVE-2024-24790 ).

Releated:
- https://github.com/golang/go/issues/67680
- https://go.dev/doc/devel/release#go1.22.4

RainbowMango: +1

grosser:
updated to 1.22.4 and made a list of places to keep in sync in go.mod
I'm not sure if removing from util.sh is safe when someone is using <1.21 so leaving that alone for now

RainbowMango (approved):
/lgtm
/approve

Issue #4695, opened 2024-03-11
`kubectl karmada join` does not support `--context --as --as-group`


grosser:
**What would you like to be added**:
support for these 3 flags

**Why is this needed**:
otherwise I need to setup a special context and `use-context` to be able to use `join`

RainbowMango:
For the `--context`, karmadactl(same as kubectl karmada) provides the two flags to specify the context for both karmada and member cluster:
```
    --cluster-context='':
	Name of cluster context in kubeconfig. The current context is used by default.
    --karmada-context='':
	The name of the kubeconfig context to use
```
Given `karmadatl join` might need two kubeconfigs, just one `--context` might be misleading as it can't represent which kubeconfig it applies to. 

Please let us know if the two flags meet your needs. Maybe I didn't understand your use case yet.

RainbowMango:
For the `--as` and `--as-group`, I guess you mean the two flags in kubectl:
```bash
    --as='':
	Username to impersonate for the operation. User could be a regular user or a service account in a namespace.

    --as-group=[]:
	Group to impersonate for the operation, this flag can be repeated to specify multiple groups.
```

I'd like to hear more details about the use case if you don't mind. We also need to consider which kubeconfig the two flags will be applied.

grosser:
I'd need to impersonate admin on the host cluster, so `--cluster-as`

grosser:
hmmm no the host cluster would be `--karmada-as` :)

RainbowMango:
Yes, `--karmada-as` and `--karmada-as-group` are sounds good to me.
cc @XiShanYongYe-Chang What do you think?

XiShanYongYe-Chang:
These two flags look good, but I don't understand one thing. Why do you need to use impersonate to join a cluster in Karmada?

grosser:
I need to be cluster admin to join the cluster, our default context is
read-only.

XiShanYongYe-Chang:
> I need to be cluster admin to join the cluster, our default context is read-only.

Is this safe? Should the admin user be the one to join the cluster?

grosser:
It needs to be the admin user since the read-only user can't do it. We do
the same ATM by adding a new context first and then use use-context.

XiShanYongYe-Chang:
> It needs to be the admin user since the read-only user can't do it. 

So you need to have the read-only user act as the admin user to join the cluster, is that correct?

> We do the same ATM by adding a new context first and then use use-context.

[...]

PR #5016, opened 2024-06-02
add --as-* options to join command

grosser:
**What type of PR is this?**

/kind feature

otherwise I need to setup a special context and use-context to be able to use join

Partially deals with #4695

```release-note
karamdactl join: add --karmada-context --karmada-as --karmada-as-group --karmada-as-uid
```

### DIFF
--- a/pkg/karmadactl/join/join.go
+++ b/pkg/karmadactl/join/join.go
@@ -107,6 +107,15 @@ type CommandJoinOption struct {
 	// more details about running Kubernetes in multiple zones.
 	ClusterZones []string
 
+	// HostAs represents the host cluster username to impersonate for the operation. User could be a regular user or a service account in a namespace
+	HostAs string
+
+	// HostAsGroups represents the host cluster group to impersonate for the operation, this flag can be repeated to specify multiple groups
+	HostAsGroups []string
+
+	// HostAsUID represents the host cluster UID to impersonate for the operation.
+	HostAsUID string
+
 	// DryRun tells if run the command in dry-run mode, without making any server requests.
 	DryRun bool
 }
@@ -150,6 +159,12 @@ func (j *CommandJoinOption) AddFlags(flags *pflag.FlagSet) {
 	flags.StringVar(&j.ClusterProvider, "cluster-provider", "", "Provider of the joining cluster. The Karmada scheduler can use this information to spread workloads across providers for higher availability.")
 	flags.StringVar(&j.ClusterRegion, "cluster-region", "", "The region of the joining cluster. The Karmada scheduler can use this information to spread workloads across regions for higher availability.")
 	flags.StringSliceVar(&j.ClusterZones, "cluster-zones", nil, "The zones of the joining cluster. The Karmada scheduler can use this information to spread workloads across zones for higher availability.")
+	flags.StringVar(&j.HostAs, "host-as", "",
+		"Host cluster username to impersonate for the operation. User could be a regular user or a service account in a namespace.")
+	flags.StringArrayVar(&j.HostAsGroups, "host-as-group", []string{},
+		"Host cluster group to impersonate for the operation, this flag can be repeated to specify multiple groups.")
+	flags.StringVar(&j.HostAsUID, "host-as-uid", "",
+		"Host cluster UID to impersonate for the operation.")
 	flags.BoolVar(&j.DryRun, "dry-run", false, "Run the command in dry-run mode, without making any server requests.")
 }
 
@@ -164,6 +179,11 @@ func (j *CommandJoinOption) Run(f cmdutil.Factory) error {
 		return fmt.Errorf("failed to get control plane rest config. context: %s, kube-config: %s, error: %v",
 			*options.DefaultConfigFlags.Context, *options.DefaultConfigFlags.KubeConfig, err)
 	}
+
+	// Configure impersonation
+	controlPlaneRestConfig.Impersonate.UserName = j.HostAs
+	controlPlaneRestConfig.Impersonate.Groups = j.HostAsGroups
+	controlPlaneRestConfig.Impersonate.UID = j.HostAsUID
 
 	// Get cluster config
 	clusterConfig, err := apiclient.RestConfig(j.ClusterContext, j.ClusterKubeConfig)


XiShanYongYe-Chang:
Thanks @grosser 
/assign

pkg/karmadactl/join/join.go
```diff
@@ -150,6 +159,12 @@ func (j *CommandJoinOption) AddFlags(flags *pflag.FlagSet) {
 	flags.StringVar(&j.ClusterProvider, "cluster-provider", "", "Provider of the joining cluster. The Karmada scheduler can use this information to spread workloads across providers for higher availability.")
 	flags.StringVar(&j.ClusterRegion, "cluster-region", "", "The region of the joining cluster. The Karmada scheduler can use this information to spread workloads across regions for higher availability.")
 	flags.StringSliceVar(&j.ClusterZones, "cluster-zones", nil, "The zones of the joining cluster. The Karmada scheduler can use this information to spread workloads across zones for higher availability.")
+	flags.StringVar(&j.KarmadaAs, "karmada-as", "",
+		"Host cluster username to impersonate for the operation. User could be a regular user or a service account in a namespace.")
```
XiShanYongYe-Chang, line 163: What does `Host cluster` refer to?

pkg/karmadactl/join/join.go
```diff
@@ -165,6 +180,11 @@ func (j *CommandJoinOption) Run(f cmdutil.Factory) error {
 			*options.DefaultConfigFlags.Context, *options.DefaultConfigFlags.KubeConfig, err)
 	}
 
+	// Configure impersonation
```
XiShanYongYe-Chang, line 183: Is there a scenario where field `.Impersonate.Extra` needs to be set?

XiShanYongYe-Chang (commented):
Sorry for the late reply.

pkg/karmadactl/join/join.go
grosser: the cluster that the given cluster will join to, "host" is supposed to match the "ClusterConfig.Host" etc naming

pkg/karmadactl/join/join.go
grosser: yeah might be useful, I did not have any use for it (neither for UID)
can be a simple later PR if anyone needs it
... also willing to add if anyone wants it

pkg/karmadactl/join/join.go
XiShanYongYe-Chang: thanks for your explain.

pkg/karmadactl/join/join.go
XiShanYongYe-Chang: We can wait to add it if we need to.

pkg/karmadactl/join/join.go
```diff
@@ -150,6 +159,12 @@ func (j *CommandJoinOption) AddFlags(flags *pflag.FlagSet) {
 	flags.StringVar(&j.ClusterProvider, "cluster-provider", "", "Provider of the joining cluster. The Karmada scheduler can use this information to spread workloads across providers for higher availability.")
 	flags.StringVar(&j.ClusterRegion, "cluster-region", "", "The region of the joining cluster. The Karmada scheduler can use this information to spread workloads across regions for higher availability.")
 	flags.StringSliceVar(&j.ClusterZones, "cluster-zones", nil, "The zones of the joining cluster. The Karmada scheduler can use this information to spread workloads across zones for higher availability.")
+	flags.StringVar(&j.KarmadaAs, "karmada-as", "",
+		"Host cluster username to impersonate for the operation. User could be a regular user or a service account in a namespace.")
+	flags.StringArrayVar(&j.KarmadaAsGroups, "karmada-as-group", []string{},
+		"Host cluster group to impersonate for the operation, this flag can be repeated to specify multiple groups.")
+	flags.StringVar(&j.KarmadaAsUID, "karmada-as-uid", "",
+		"Host cluster UID to impersonate for the operation.")
```
XiShanYongYe-Chang, line 167: The karmada naming seems a bit strange. Is it possible to reconsider the following three naming methods?
```
as-username-on-host
as-group-on-host
as-uid-on-host

host-as
host-as-group
host-as-uid

as
as-group
as-uid
```
What would be a better name for it?

RainbowMango: I would prefer to the first group as it more clear and straightforward.
@grosser what do you think?
Sorry for giving the wrong directions at https://github.com/karmada-io/karmada/issues/4695#issuecomment-1989757658.

grosser: - 1 and 2 seem fine, 3 is confusing
- `as-username-on-host` don't like this since it breaks the symmetry to kubectl
- prefer 2 in general since it means all options for the host stay neatly grouped together

thoughts ?

RainbowMango: I prefer 1, and also OK with 2.

> as-username-on-host don't like this since it breaks the symmetry to kubectl

Yes, we usually follow the symmetry of Kubernetes, which would make it easier for users to understand.
We all agree that 3 is confusing, so we can't 100% follow kubectl, so I tend to the one more straightforward.

Group 1 is more clear than Group 2. But no strong opinion.  You can make the call.

> prefer 2 in general since it means all options for the host stay neatly grouped together

+1

RainbowMango: I prefer 1, and also OK with 2.

> as-username-on-host don't like this since it breaks the symmetry to kubectl

Yes, we usually follow the symmetry of Kubernetes, which would make it easier for users to understand.
We all agree that 3 is confusing, so we can't 100% follow kubectl, so I tend to the one more straightforward.

Group 1 is more clear than Group 2. But no strong opinion.  You can make the call.

> prefer 2 in general since it means all options for the host stay neatly grouped together

+1

RainbowMango:
/approve

Leave LGTM to @XiShanYongYe-Chang for another look.

grosser:
... oh ,man just found `karmada-context`
so afaik either this needs to change to `host-context` or we need to stay with `karmada-` prefix
to not make things inconsistent

XiShanYongYe-Chang:
> ... oh ,man just found `karmada-context`
> so afaik either this needs to change to `host-context` or we need to stay with `karmada-` prefix
> to not make things inconsistent

I think you're referring to the `cluster-context` parameter.

Modifying existing parameters affects users who are using the parameters. Therefore, I do not recommend changing the parameter name directly.

RainbowMango:
That's the legacy option, if we want to change it we should deprecate it. This PR focus on add the right options.

grosser:
@XiShanYongYe-Chang  I'm saying there is an option already called `karmada-context`, so adding things starting with `host` (which means the same thing) seems confusing.
Unless we agree that `karmada-context` should be deprecated in favor of `host-context` (either now or in a later PR)

XiShanYongYe-Chang:
Oh, I get it.

If we decide to do it, we can do it in another pr.

XiShanYongYe-Chang:
/lgtm

Hi @grosser do you have any plans to continue to modify the naming of karmada-context?

XiShanYongYe-Chang:
Hi @grosser, can you help squash the commits into one?

grosser:
I can make a follow-up to make host-context an alias for deprecated karmada-context (idk how deprecation works, but can at least mark as deprecated in the description)

XiShanYongYe-Chang:
@grosser thanks
/hold cancel
/lgtm

grosser:
https://github.com/karmada-io/karmada/pull/5066 for karmada-context and karmada-kubeconfig deprecation

PR #5023, opened 2024-06-05
Adding FlinkDeployment CRD to supported third party resource customizations

### LINKED ISSUE #4953: Add default ResourceInterpreterCustomization for FlinkDeployments

We would like to add a default ResourceInterpreterCustomization for the `FlinkDeployment CRD` which is [described here](https://nightlies.apache.org/flink/flink-kubernetes-operator-docs-release-1.8/docs/custom-resource/reference/.). 

**Why is this needed**:

We are investigating if we will be using Karmada for our native Flink application use-case. As part of the investigation we've already implemented a custom interpreter for this resource. After speaking with the community, it seems that this integration would be beneficial. 

Since this code is already implemented on our side, I'm happy to assign this to myself.

mszacillo:
**What type of PR is this?**
/kind feature

This PR adds a default resource customization for FlinkDeployments. Below is a summary of the supported default interpretations:

**Supported**
1. InterpretHealth
    - We monitor the internal flink state that is added to the FlinkDeployment status via Flink Operator: `spec.status.jobStatus.state`. FlinkDeployments can enter errored states due to user related code issues, rather than due to bad cluster health or lack of cluster resources. As a result we want to differentiate from these cases so that Karmada only reschedules these apps when needed. 
    - We chose to mark FlinkDeployments as `Unhealthy` only if the internal flink state is trapped in a non-terminal state - these states include:
        1. CREATED (job was created but not started)
        2. RECONCILING (state unknown, normally due to JM unable to communicate with operator)
    - More info on flink state transitions here: https://nightlies.apache.org/flink/flink-docs-master/docs/internals/job_scheduling/ 
    
2. InterpretReplica
    - FlinkDeployments do not currently get subdivided amongst clusters. The reasoning here is that each Flink Cluster (JM + TMs) should be colocated to reduce request latency and boost performance. 
    - Since Karmada interprets scheduling using replicas, we set the replica of the FlinkDeployment to 1, and Karmada will simply schedule the FlinkDeployment CRD to a member cluster, at which point the FlinkOperator will take over. 
    - Replica resource requirements for CPU and Memory are set to equal (n * JM) + (k * TM), where n is the number of JobManagers and k is the number of Task Managers. 
    
3. InterpretStatus
    - We currently pull the full FlinkDeployment status from the resource. 

4. AggregateStatus
    - Since the replica count is 1, the aggregation isn't technically necessary. But in order for the status to show up on the FlinkDeployment resource in Karmada, this definition is needed. I've defined the interpreter to set the Aggregate status equal to the latest status pulled by Karmada.

**Not Supported**
1. Retention
    - At the moment not necessary. We do not plan on changes being made to FlinkDeployment spec on the cluster that need to be retained by Karmada.

2. ReplicaRevision
    - As mentioned, we currently don't plan on increasing the replica size for the FlinkDeployment. Even though JM and TM values can be scaled, the total number of FlinkDeployments will stay 1. 

3. InterpretDependency
    - There are no general dependencies needed for FlinkDeployments. Secrets or other configmaps can be handled on a case-by-case basis. 

Fixes #4953

**Testing**
Tested on cluster and confirmed default interpreters work with FlinkDeployment. 

```release-note
Adding FlinkDeployment v1beta1 CRD to supported third party resource customizatons.
```

### DIFF
--- a/pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/flink.apache.org/v1beta1/FlinkDeployment/customizations.yaml
+++ b/pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/flink.apache.org/v1beta1/FlinkDeployment/customizations.yaml
@@ -38,8 +38,18 @@ spec:
             jm_replicas = 1
           end
 
-          parallelism = observedObj.spec.job.parallelism
-          tm_replicas = math.ceil(parallelism / observedObj.spec.flinkConfiguration['taskmanager.numberOfTaskSlots'])
+          -- TaskManager replica setting takes precedence over parallelism setting
+
+          tm_replicas = observedObj.spec.taskManager.replicas
+          if isempty(tm_replicas) then
+            parallelism = observedObj.spec.job.parallelism
+            task_slots = observedObj.spec.flinkConfiguration['taskmanager.numberOfTaskSlots']
+            if isempty(parallelism) or isempty(task_slots) then
+              tm_replicas = 1
+            else
+              tm_replicas = math.ceil(parallelism / observedObj.spec.flinkConfiguration['taskmanager.numberOfTaskSlots'])
+            end
+          end
 
           replica = jm_replicas + tm_replicas
 
--- a/pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/flink.apache.org/v1beta1/FlinkDeployment/customizations.yaml
+++ b/pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/flink.apache.org/v1beta1/FlinkDeployment/customizations.yaml
@@ -11,7 +11,11 @@ spec:
       luaScript: >
         function InterpretHealth(observedObj)
           if observedObj.status ~= nil and observedObj.status.jobStatus ~= nil then
-            return observedObj.status.jobStatus.state ~= 'CREATED' and observedObj.status.jobStatus.state ~= 'RECONCILING'
+            if observedObj.status.jobStatus.state ~= 'CREATED' and observedObj.status.jobStatus.state ~= 'RECONCILING' then
+              return true
+            else
+              return observedObj.status.jobManagerDeploymentStatus == 'ERROR'
+            end
           end
           return false
         end
@@ -24,32 +28,36 @@ spec:
         end
 
         function GetReplicas(observedObj)
-          -- FlinkDeployments presently will not be subdivided among clusters, replica should be 1
-          replica = 1
           requires = {
             resourceRequest = {},
+            nodeClaim = {},
           }
-          -- Add jobmanager resources into replica requirement
 
           jm_replicas = observedObj.spec.jobManager.replicas
           if isempty(jm_replicas) then
             jm_replicas = 1
           end
 
-          for i = 1, jm_replicas do
-            requires.resourceRequest.cpu = kube.resourceAdd(requires.resourceRequest.cpu, tostring(observedObj.spec.jobManager.resource.cpu))
-            requires.resourceRequest.memory = kube.resourceAdd(requires.resourceRequest.memory, observedObj.spec.jobManager.resource.memory)
-          end
-
-          -- Add task manager resources into replica requirement
-
           parallelism = observedObj.spec.job.parallelism
-          tms = math.ceil(parallelism / observedObj.spec.flinkConfiguration['taskmanager.numberOfTaskSlots'])
+          tm_replicas = math.ceil(parallelism / observedObj.spec.flinkConfiguration['taskmanager.numberOfTaskSlots'])
 
-          for i = 1, tms do
-            requires.resourceRequest.cpu = kube.resourceAdd(requires.resourceRequest.cpu, tostring(observedObj.spec.taskManager.resource.cpu))
-            requires.resourceRequest.memory = kube.resourceAdd(requires.resourceRequest.memory, observedObj.spec.taskManager.resource.memory)
+          replica = jm_replicas + tm_replicas
+
+          -- Until multiple podTemplates are supported in replicaRequirements, take max of cpu + memory values as requirement
+
+          requires.resourceRequest.cpu = math.max(observedObj.spec.taskManager.resource.cpu, observedObj.spec.jobManager.resource.cpu)
+          jm_memory_value = kube.getResourceQuantity(observedObj.spec.jobManager.resource.memory)
+          tm_memory_value = kube.getResourceQuantity(observedObj.spec.taskManager.resource.memory)
+          if jm_memory_value > tm_memory_value then
+            requires.resourceRequest.memory = observedObj.spec.jobManager.resource.memory
+          else
+            requires.resourceRequest.memory = observedObj.spec.taskManager.resource.memory
           end
+
+          -- Until multiple podTemplates are supported, interpreter will only take affinity and toleration input to common podTemplate
+
+          requires.nodeClaim.nodeSelector = observedObj.spec.podTemplate.spec.nodeSelector
+          requires.nodeClaim.tolerations = observedObj.spec.podTemplate.spec.tolerations
 
           return replica, requires
         end
--- a/pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/flink.apache.org/v1beta1/FlinkDeployment/customizations.yaml
+++ b/pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/flink.apache.org/v1beta1/FlinkDeployment/customizations.yaml
@@ -56,8 +56,10 @@ spec:
 
           -- Until multiple podTemplates are supported, interpreter will only take affinity and toleration input to common podTemplate
 
-          requires.nodeClaim.nodeSelector = observedObj.spec.podTemplate.spec.nodeSelector
-          requires.nodeClaim.tolerations = observedObj.spec.podTemplate.spec.tolerations
+          if observedObj.spec.podTemplate ~= nil and observedObj.spec.podTemplate.spec ~= nil then
+            requires.nodeClaim.nodeSelector = observedObj.spec.podTemplate.spec.nodeSelector
+            requires.nodeClaim.tolerations = observedObj.spec.podTemplate.spec.tolerations
+          end
 
           return replica, requires
         end
--- a/pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/flink.apache.org/v1beta1/FlinkDeployment/customizations.yaml
+++ b/pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/flink.apache.org/v1beta1/FlinkDeployment/customizations.yaml
@@ -1,0 +1,110 @@
+apiVersion: config.karmada.io/v1alpha1
+kind: ResourceInterpreterCustomization
+metadata:
+  name: declarative-configuration-flinkdeployment
+spec:
+  target:
+    apiVersion: flink.apache.org/v1beta1
+    kind: FlinkDeployment
+  customizations:
+    healthInterpretation:
+      luaScript: >
+        function InterpretHealth(observedObj)
+          if observedObj.status ~= nil and observedObj.status.jobStatus ~= nil then
+            return observedObj.status.jobStatus.state ~= 'CREATED' and observedObj.status.jobStatus.state ~= 'RECONCILING'
+          end
+          return false
+        end
+    replicaResource:
+      luaScript: >
+        local kube = require("kube")
+
+        local function isempty(s)
+          return s == nil or s == ''
+        end
+
+        function GetReplicas(observedObj)
+          -- FlinkDeployments presently will not be subdivided among clusters, replica should be 1
+          replica = 1
+          requires = {
+            resourceRequest = {},
+          }
+          -- Add jobmanager resources into replica requirement
+
+          jm_replicas = observedObj.spec.jobManager.replicas
+          if isempty(jm_replicas) then
+            jm_replicas = 1
+          end
+
+          for i = 1, jm_replicas do
+            requires.resourceRequest.cpu = kube.resourceAdd(requires.resourceRequest.cpu, tostring(observedObj.spec.jobManager.resource.cpu))
+            requires.resourceRequest.memory = kube.resourceAdd(requires.resourceRequest.memory, observedObj.spec.jobManager.resource.memory)
+          end
+
+          -- Add task manager resources into replica requirement
+
+          parallelism = observedObj.spec.job.parallelism
+          tms = math.ceil(parallelism / observedObj.spec.flinkConfiguration['taskmanager.numberOfTaskSlots'])
+
+          for i = 1, tms do
+            requires.resourceRequest.cpu = kube.resourceAdd(requires.resourceRequest.cpu, tostring(observedObj.spec.taskManager.resource.cpu))
+            requires.resourceRequest.memory = kube.resourceAdd(requires.resourceRequest.memory, observedObj.spec.taskManager.resource.memory)
+          end
+
+          return replica, requires
+        end
+    statusAggregation:
+      luaScript: >
+        function AggregateStatus(desiredObj, statusItems)
+          if statusItems == nil then
+            return desiredObj
+          end
+          if desiredObj.status == nil then
+            desiredObj.status = {}
+          end
+          clusterInfo = {}
+          jobManagerDeploymentStatus = ''
+          jobStatus = {}
+          lifecycleState = ''
+          observedGeneration = 0
+          reconciliationStatus = {}
+          taskManager = {}
+
+          for i = 1, #statusItems do
+            currentStatus = statusItems[i].status
+            if currentStatus ~= nil then
+              clusterInfo = currentStatus.clusterInfo
+              jobManagerDeploymentStatus = currentStatus.jobManagerDeploymentStatus
+              jobStatus = currentStatus.jobStatus
+              observedGeneration = currentStatus.observedGeneration
+              lifecycleState = currentStatus.lifecycleState
+              reconciliationStatus = currentStatus.reconciliationStatus
+              taskManager = currentStatus.taskManager
+            end
+          end
+
+          desiredObj.status.clusterInfo = clusterInfo
+          desiredObj.status.jobManagerDeploymentStatus = jobManagerDeploymentStatus
+          desiredObj.status.jobStatus = jobStatus
+          desiredObj.status.lifecycleState = lifecycleState
+          desiredObj.status.observedGeneration = observedGeneration
+          desiredObj.status.reconciliationStatus = reconciliationStatus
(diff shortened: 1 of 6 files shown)


yike21:
@mszacillo Thanks for your contribution!

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/flink.apache.org/v1beta1/FlinkDeployment/customizations.yaml
```diff
@@ -0,0 +1,112 @@
...
+          if desiredObj.status == nil then
+            desiredObj.status = {}
+          end
+          clusterInfo = {}
+          jobManagerDeploymentStatus = ''
+          jobStatus = {}
+          lifecycleState = ''
+          observedGeneration = 0
+          savepointInfo = {}
+          lastSavepoint = {}
```
yike21, line 71: How about deleting these two fields as they are not used?

chaunceyjiang (commented):
Thanks~

/lgtm

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/flink.apache.org/v1beta1/FlinkDeployment/customizations.yaml
```diff
@@ -0,0 +1,110 @@
...
+          parallelism = observedObj.spec.job.parallelism
+          tms = math.ceil(parallelism / observedObj.spec.flinkConfiguration['taskmanager.numberOfTaskSlots'])
+
+          for i = 1, tms do
+            requires.resourceRequest.cpu = kube.resourceAdd(requires.resourceRequest.cpu, tostring(observedObj.spec.taskManager.resource.cpu))
+            requires.resourceRequest.memory = kube.resourceAdd(requires.resourceRequest.memory, observedObj.spec.taskManager.resource.memory)
+          end
+
+          return replica, requires
+        end
```
RainbowMango, line 55: > FlinkDeployments presently will not be subdivided among clusters, replica should be 1

We probably can not just add both `job manager` and `task manager`'s requested resources and treat it just has `one` `super replica`. Otherwise, the karmada-scheduler will try to filter the cluster which has a super node that can hold the the super replica. But that isn't the case, actually, the replicas of `job manager` and `task manager` can be scheduled among different nodes, they do not require a super node to hold all of them.

RainbowMango: This is a limitation of Karmada for now(there is a plan for this). If a workload(Like FlinkDeployment) has more than one pod template(`JobManager` and `TaskManager`), the resource interpreter can not present each pod template's requirement.

The workaround is that:
1. Sum replicas of `JobManager` and `TaskManager` as the total number of replicas of `FlinkDeployment`
2. When determining the resources for replicas, take the higher resource requirement between JobManager and TaskManager.

bharathguvvala, line 55: How are the tolerations and nodeAffinity constraints interpreted from the flinkdeployment podTemplate?

mszacillo: Ah I see! I'll try this out.

mszacillo: FlinkDeployments allow you to define podTemplates in a layered manner - meaning you can have a podTemplate with shared attributes between JobManagers and Taskmanagers, as well as individual podTemplates for both JM and TMs. 

The tolerations and nodeAffinity are something that will be handled by the Flink Operator. 

More info here: https://nightlies.apache.org/flink/flink-kubernetes-operator-docs-release-1.8/docs/custom-resource/pod-template/

mszacillo: This somewhat works, but I would still need to set the replica = 1. Karmada schedules replicas, so if the replica count is greater than 1, Karmada might decide to divide the replicas among clusters and result in duplicate FlinkDeployments on multiple clusters. This is not desired. 

We could keep the replica = 1, and instead set the requirement to be equal to the higher resource requirement between JM and TM as an estimation.

RainbowMango: Yes, as I said, it is still a workaround before your proposal #5085. But it'll work.

The annoying thing about this workaround approach is it's hard to calculate the required resources.

case 1: If both the JobManager and TaskManger require the same set of resources like they both require CPU and memory, that's easy to figure out which component requires more resources, and then we just need to take higher resource requirements. (lost some accurateness) (but if JobManger requires more CPU and less memory than TaskManager, drop to case 2)

case 2: But what if JobManger and TaskManager require different set of resources, for example, JobManager requires CPU but TaskManager requires memory(or even GPU resources). for this case, draft idea in my mind is that we can take the higher resource requirement of each resource type across both components, something like:
```
          resourceRequest = {}
        
          for resourceType := range taskManager.resource {
              if resourceRequest[resourceType] < taskManager.resource[resourceType] { 
                  resourceRequest[resourceType] = taskManager.resource[resourceType]
              }
          }
        
          for resourceType := range jobManger.resource {
              if resourceRequest[resourceType] < jobManger.resource[resourceType] { 
                  resourceRequest[resourceType] = jobManger.resource[resourceType]
              }
          }
```

bharathguvvala: While scheduling, the resource estimator takes the tolerations and nodeAffinity constraints of the workload into account to determine the number of replicas that can be accommodated within a member cluster (based on the taints etc). I was wondering how that information is interpreted from the FlinkDeployment.

bharathguvvala: I do see that there's a way to provide that eg: https://github.com/karmada-io/karmada/blob/84b971a501ba82c53a5ad455c2fe84d842cd7d4e/pkg/apis/config/v1alpha1/resourceinterpretercustomization_types.go#L138

bharathguvvala: This is a valid suggestion. It eliminates the false negatives where the cluster has adequate capacity but the scheduler disqualifies it due to the lack of accommodating a super replica.

RainbowMango: I guess what @bharathguvvala means current PR should parse the toleration and nodeAffinity constraints from the pod template and set to [NodeClaim](https://github.com/karmada-io/karmada/blob/0372cabba6f2e26346319aad8b8600de5acff832/pkg/apis/work/v1alpha2/binding_types.go#L183).

bharathguvvala: Yes @RainbowMango . That's what I meant. Something like this would work.

```
requires.nodeClaim = {}
requires.nodeClaim.nodeSelector = observedObj.spec.podTemplate.spec.nodeSelector
requires.nodeClaim.tolerations = observedObj.spec.podTemplate.spec.tolerations
```

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/flink.apache.org/v1beta1/FlinkDeployment/customizations.yaml
```diff
@@ -0,0 +1,110 @@
...
+spec:
+  target:
+    apiVersion: flink.apache.org/v1beta1
+    kind: FlinkDeployment
+  customizations:
+    healthInterpretation:
+      luaScript: >
+        function InterpretHealth(observedObj)
+          if observedObj.status ~= nil and observedObj.status.jobStatus ~= nil then
+            return observedObj.status.jobStatus.state ~= 'CREATED' and observedObj.status.jobStatus.state ~= 'RECONCILING'
```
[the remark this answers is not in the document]
bharathguvvala: Also I see there are other states such as 'RUNNING' which are possible, why is the health interpreted as false for other states ?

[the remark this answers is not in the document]
bharathguvvala: I see 4 different states that are present in the status which represent the different aspects of the deployment health which should be factored in this healthInterpretation logic.

1.  jobManagerDeploymentStatus
2. lifecycleState
3. jobStatus.state
4. state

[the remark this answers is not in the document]
mszacillo: I really should put WIP on this PR, as the code here is outdated with what we are using. To answer your questions from top-down:

1. No it is not possible for the state to be CREATED and RECONCILING at the same time. Reconciling represents an unknown state, meaning the operator is unable to get the state from the JobManager for a certain period of time. Flink state follows a state machine - there should never be a time where two states are shared. 

2. The intended operator is "and". We set the health of the flink application to be unhealthy, only if the application has been in a CREATED or RECONCILING state for more than the allowed toleration time, meaning the application has not been scheduled yet. If we used an "or", the health interpreter would return true always. 

3. In our updated health interpreter takes jobStatus.state and jobManagerDeploymentStatus into account - we felt that these two states were sufficient for determining whether the application has been scheduled or not. If you find an edge case that is not covered by these two states, please let us know!

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/flink.apache.org/v1beta1/FlinkDeployment/customizations.yaml
mszacillo: Got it, thanks for this information!

mszacillo: I think again since we don't support multiple pod templates yet, we can just take the nodeAffinity and tolerations from the common podTemplate. Otherwise we can't set differing nodeAffinities / tolerations for JM and TMs.

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/flink.apache.org/v1beta1/FlinkDeployment/customizations.yaml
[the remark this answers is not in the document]
mszacillo: Oh just realized I forgot to answer one of your questions - why the health is interpreted as true for states that aren't CREATED or RECONCILING. 

We want to differentiate between scheduling errors and application errors. If you take a look at the jobstatus state transition diagram: https://nightlies.apache.org/flink/flink-docs-master/docs/internals/job_scheduling/, you'll see that each application will start in a CREATED state. Then to a RUNNING state, and ultimately to one of the terminal states. Since we plan on using Karmada for descheduling, we need to only interpret the health if we suspect the Flink application is unable to start due to a scheduling issue (ignoring cases in which the application itself is failing, potentially due to a code-error for instance).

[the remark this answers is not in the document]
bharathguvvala: My bad. Since I am new to Lua scripting I took '~=' to be an equality operator hence my previous question about the jobStatus being both CREATED and RECONCILING.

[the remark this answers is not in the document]
bharathguvvala: > Oh just realized I forgot to answer one of your questions - why the health is interpreted as true for states that aren't CREATED or RECONCILING.
> 
> We want to differentiate between scheduling errors and application errors. If you take a look at the jobstatus state transition diagram: https://nightlies.apache.org/flink/flink-docs-master/docs/internals/job_scheduling/, you'll see that each application will start in a CREATED state. Then to a RUNNING state, and ultimately to one of the terminal states. Since we plan on using Karmada for descheduling, we need to only interpret the health if we suspect the Flink application is unable to start due to a scheduling issue (ignoring cases in which the application itself is failing, potentially due to a code-error for instance).

Thank you for the detailed response. I had gone through the state transition diagram earlier which helped me understand the justification for that predicate.

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/flink.apache.org/v1beta1/FlinkDeployment/customizations.yaml
```diff
@@ -0,0 +1,120 @@
...
+          if observedObj.status ~= nil and observedObj.status.jobStatus ~= nil then
+            if observedObj.status.jobStatus.state ~= 'CREATED' and observedObj.status.jobStatus.state ~= 'RECONCILING' then
+              return true
+            else
+              return observedObj.status.jobManagerDeploymentStatus == 'ERROR'
+            end
+          end
+          return false
+        end
+    replicaResource:
```
bharathguvvala, line 22: @mszacillo we should support replicaRevision interpretation operation given that the replica derivation logic has changed from replica = 1 to replica = jobmanager replica + task manager replica

bharathguvvala: we might have to implement retention as well if the flink deployment is enabled with [Flink Autoscaler](https://nightlies.apache.org/flink/flink-kubernetes-operator-docs-release-1.9/docs/custom-resource/autoscaler/) .

mszacillo: I was curious about this, but because Karmada does not currently support handling of multiple podTemplates, I'm not sure this can currently be implemented. From local testing I see that the ResourceBinding always updates with the correct number of replicas (by editing the existing FlinkDeployment). I believe this is because the InterpretReplica rule will get triggered on creates and edits. Perhaps @RainbowMango help can clarify?

mszacillo: For autoscaler, we don't personally use it but I would be open to adding this to the default interpreter if there are other use-cases for it. What would you need to retain in the job?

bharathguvvala: > For autoscaler, we don't personally use it but I would be open to adding this to the default interpreter if there are other use-cases for it. What would you need to retain in the job?

We'd want to retain the fields such as job.parallelism which would get changed in the member cluster.

RainbowMango: > Perhaps @RainbowMango can help clarify?

For the time being, the replicas of FlinkDeloyment are not supposed to be divided across multiple member clusters, which means the scheduling type of FlinkDepoyment is `Duplicated`, Karmada won't `revise` the replicas in this case, so for now, we can skip the `replicaRevision`.

RainbowMango: > we might have to implement retention as well if the flink deployment is enabled with [Flink Autoscaler](https://nightlies.apache.org/flink/flink-kubernetes-operator-docs-release-1.9/docs/custom-resource/autoscaler/) .

Yeah, the retention is needed in case working with HPA or Autoscaler. 
But we need to be very cautious because it will affect users who are not utilizing HPA (lost the capability to make changes from Karmada side).

This case is similar with Deployment and HPA, we introduced `resourcetemplate.karmada.io/retain-replicas` label which will be used to decide if specific workload need to retain replicas. See the example here:
https://github.com/karmada-io/karmada/blob/581f392525c1bac8fffdbcfd7e3ab9fcb29d752e/pkg/resourceinterpreter/default/native/retain.go#L151-L160

bharathguvvala: @RainbowMango  Exactly. And we'd have to build something similar to hpaReplicasSyncer for FlinkDeployments. In our company, where the flink deployments are autoscaled we are planning to build something similar to introduce retention of parallelism and a sync controller. If required we can take this as a separate requirement and open a new PR with these changes.

mszacillo: @bharathguvvala Since you're more familiar with this specific use-case, would you be open to creating a ticket for this and potentially opening a new PR after this one is completed?

bharathguvvala: @mszacillo Sure, that's what I was proposing. I'll take this up as a separate PR.

RainbowMango: > In our company, where the flink deployments are autoscaled we are planning to build something similar to introduce retention of parallelism and a sync controller. If required we can take this as a separate requirement outside this PR and open a new PR with these changes.

+1 on the separate requirement outside of this PR

The retention part is simple all we need to do is take the label into account when implementing the interpreter.

But the sync controller part might not be that simple, as this will force us to import relevant API, that's the concern. We probably need a common mechanism(not limited to specified resource type) to handle the syncing work.

RainbowMango (commented):
/lgtm

@bharathguvvala Do you have any further comments?

bharathguvvala:
> /lgtm
> 
> @bharathguvvala Do you have any further comments?

LGTM for me as well.

RainbowMango (approved):
/approve

PR #5066, opened 2024-06-20
deprecate karmada-* options in favor of host-* options

grosser:
as followup to https://github.com/karmada-io/karmada/pull/5016 which added host-as/group/uid

cli will be stable, not sure if anyone uses the karmada code directly to set these options

**What type of PR is this?**

/kind deprecation

makes cli options consistent

```release-note
- deprecates --karmada-context in favor of --host-context
- deprecates --karmada-kubeconfig in favor of --host-kubeconfig
```

### DIFF
--- a/cmd/agent/app/options/options.go
+++ b/cmd/agent/app/options/options.go
@@ -58,8 +58,8 @@ type Options struct {
 	KarmadaKubeConfig string
 	// ClusterContext is the name of the cluster context in control plane KUBECONFIG file.
 	// Default value is the current-context.
-	KarmadaContext string
-	ClusterName    string
+	HostContext string
+	ClusterName string
 	// ClusterNamespace holds the namespace name where the member cluster secrets are stored.
 	ClusterNamespace string
 	// ClusterStatusUpdateFrequency is the frequency that controller computes and report cluster status.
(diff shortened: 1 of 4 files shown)


XiShanYongYe-Chang:
Thanks @grosser 

Hi @RainbowMango Can we do this in the current version?

cmd/agent/app/options/options.go
```diff
@@ -184,7 +184,8 @@ func (o *Options) AddFlags(fs *pflag.FlagSet, allControllers []string) {
 		"The duration the clients should wait between attempting acquisition and renewal "+
 		"of a leadership. This is only applicable if leader election is enabled.")
 	fs.StringVar(&o.KarmadaKubeConfig, "karmada-kubeconfig", o.KarmadaKubeConfig, "Path to karmada control plane kubeconfig file.")
-	fs.StringVar(&o.KarmadaContext, "karmada-context", "", "Name of the cluster context in karmada control plane kubeconfig file.")
```
RainbowMango, line 187: I guess the `--karmada-context` is exactly the context of the Karmada control plane, not the host cluster.
So, why rename it?

grosser: oh I though host and karmada would be the same
if they are not then I need to change https://github.com/karmada-io/karmada/pull/5016 to be karmada-as/karmada-group/karmada-uid

RainbowMango: Yeah, I apologize that I misunderstood your use case, I thought you wanted to impersonate users on host clusters, but after revisiting the original issue (#4695), I figured out the user that you need to impersonate is on the Karmada control plane.

> if they are not then I need to change https://github.com/karmada-io/karmada/pull/5016 to be karmada-as/karmada-group/karmada-uid

+1 Yes, that's the one we talked about at https://github.com/karmada-io/karmada/issues/4695#issuecomment-1989757658.

RainbowMango: By the way, we usually call a cluster `host` because the cluster will host Karmada's components.
So, the `--host` might only be present in `karmadactl init` or `karmadactl addon` command, which is used to install components on a host cluster.

grosser: kk, so I'll send a rename PR for the -as stuff :)

grosser: @RainbowMango https://github.com/karmada-io/karmada/pull/5099

PR #5067, opened 2024-06-20
In full query mode, the serial query of each cluster takes a long tim…

tongxin13:
/kind feature

When there are a large number of clusters, the serial query logic takes a long time to execute.  Executing queries in parallel and then aggregating results can improve query speed


tongxin13:
add Signed

PR #5077, opened 2024-06-21
additional print columns for CRDs: `ResourceInterpreterCustomization`…

a7i:
… `ClusterPropagationPolicy` and `PropagationPolicy`

**What type of PR is this?**

/kind feature

currently:
```
NAME                                    AGE
retain-configmap                        43d
retain-crd                              63d
retain-cronjob                          46d
retain-daemonset                        46d
retain-deployment                       46d
retain-mutatingwebhookconfiguration     63d
retain-pvc-annotations                  52d
retain-secret                           63d
retain-statefulset                      46d
retain-validatingwebhookconfiguration   63d
```

vs.
```
NAME                                    TARGET-API-VERSION                TARGET-KIND
retain-configmap                        v1                                ConfigMap
retain-crd                              apiextensions.k8s.io/v1           CustomResourceDefinition
retain-cronjob                          batch/v1                          CronJob
retain-daemonset                        apps/v1                           DaemonSet
retain-deployment                       apps/v1                           Deployment
retain-mutatingwebhookconfiguration     admissionregistration.k8s.io/v1   MutatingWebhookConfiguration
retain-pv-annotations                   v1                                PersistentVolume
retain-pvc-annotations                  v1                                PersistentVolumeClaim
retain-secret                           v1                                Secret
retain-statefulset                      apps/v1                           StatefulSet
retain-validatingwebhookconfiguration   admissionregistration.k8s.io/v1   ValidatingWebhookConfiguration
```

Fixes #

```release-note
`ResourceInterpreterCustomization`: Added two additional printer columns, TARGET-API-VERSION and TARGET-KIND, to represent the target resource type, these columns will be displayed in the output of kubectl get.

`PropagationPolicy`/`ClusterPropagationPolicy`: Added two additional printer columns, `Conflict-Resolution` and `Priority`, to represent the conflict resolution strategy and priority, these columns will be displayed in the output of kubectl get.
```

### DIFF
--- a/pkg/apis/config/v1alpha1/resourceinterpretercustomization_types.go
+++ b/pkg/apis/config/v1alpha1/resourceinterpretercustomization_types.go
@@ -36,6 +36,8 @@ const (
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 // +kubebuilder:resource:path=resourceinterpretercustomizations,scope="Cluster",shortName=ric,categories={karmada-io}
 // +kubebuilder:storageversion
+// +kubebuilder:printcolumn:JSONPath=`.spec.target.apiVersion`,name="TARGET-API-VERSION",type=string,priority=1
+// +kubebuilder:printcolumn:JSONPath=`.spec.target.kind`,name="TARGET-KIND",type=string,priority=1
 
 // ResourceInterpreterCustomization describes the configuration of a specific
 // resource for Karmada to get the structure.
--- a/pkg/apis/policy/v1alpha1/propagation_types.go
+++ b/pkg/apis/policy/v1alpha1/propagation_types.go
@@ -44,6 +44,8 @@ const (
 // +genclient
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 // +kubebuilder:resource:path=propagationpolicies,scope=Namespaced,shortName=pp,categories={karmada-io}
+// +kubebuilder:printcolumn:JSONPath=`.spec.conflictResolution`,name="CONFLICT-RESOLUTION",type=string,priority=1
+// +kubebuilder:printcolumn:JSONPath=`.spec.priority`,name="PRIORITY",type=string,priority=1
 
 // PropagationPolicy represents the policy that propagates a group of resources to one or more clusters.
 type PropagationPolicy struct {
@@ -562,6 +564,8 @@ type PropagationPolicyList struct {
 // +genclient:nonNamespaced
 // +kubebuilder:resource:path=clusterpropagationpolicies,scope="Cluster",shortName=cpp,categories={karmada-io}
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:printcolumn:JSONPath=`.spec.conflictResolution`,name="CONFLICT-RESOLUTION",type=string,priority=1
+// +kubebuilder:printcolumn:JSONPath=`.spec.priority`,name="PRIORITY",type=string,priority=1
 
 // ClusterPropagationPolicy represents the cluster-wide policy that propagates a group of resources to one or more clusters.
 // Different with PropagationPolicy that could only propagate resources in its own namespace, ClusterPropagationPolicy
--- a/charts/karmada/_crds/bases/config/config.karmada.io_resourceinterpretercustomizations.yaml
+++ b/charts/karmada/_crds/bases/config/config.karmada.io_resourceinterpretercustomizations.yaml
@@ -18,7 +18,16 @@ spec:
     singular: resourceinterpretercustomization
   scope: Cluster
   versions:
-  - name: v1alpha1
+  - additionalPrinterColumns:
+    - jsonPath: .spec.target.apiVersion
+      name: TARGET-API-VERSION
+      priority: 1
+      type: string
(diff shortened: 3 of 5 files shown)


pkg/apis/policy/v1alpha1/propagation_types.go
```diff
@@ -44,6 +44,8 @@ const (
 // +genclient
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 // +kubebuilder:resource:path=propagationpolicies,scope=Namespaced,shortName=pp,categories={karmada-io}
+// +kubebuilder:printcolumn:JSONPath=`.spec.conflictResolution`,name="Conflict-Resolution",type=string
+// +kubebuilder:printcolumn:JSONPath=`.spec.priority`,name="Priority",type=string
```
XiShanYongYe-Chang, line 48: Is it possible to set `priority=1` for these two values, so that they are visible when `-o wide` is specified?

pkg/apis/config/v1alpha1/resourceinterpretercustomization_types.go
```diff
@@ -36,6 +36,8 @@ const (
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 // +kubebuilder:resource:path=resourceinterpretercustomizations,scope="Cluster",shortName=ric,categories={karmada-io}
 // +kubebuilder:storageversion
+// +kubebuilder:printcolumn:JSONPath=`.spec.target.apiVersion`,name="Target-API-Version",type=string
+// +kubebuilder:printcolumn:JSONPath=`.spec.target.kind`,name="Target-Kind",type=string
```
XiShanYongYe-Chang, line 40: It is better to use all uppercase for the name. It is also all uppercase when displayed.

XiShanYongYe-Chang (commented):
Thanks~

pkg/apis/config/v1alpha1/resourceinterpretercustomization_types.go
chaunceyjiang: +1

RainbowMango, line 40:
```suggestion
// +kubebuilder:printcolumn:JSONPath=`.spec.target.apiVersion`,name="TARGET-API-VERSION",type=string
// +kubebuilder:printcolumn:JSONPath=`.spec.target.kind`,name="TARGET-KIND",type=string
```
I guess that's what @XiShanYongYe-Chang and @chaunceyjiang mean.
LGTM. 
But the previous APIs don't follow this rule ;) :) :)

pkg/apis/policy/v1alpha1/propagation_types.go
RainbowMango: Echo from [kubebuilder document](https://book.kubebuilder.io/reference/markers/crd), the priority indicates
> Lower priority (higher numbered) columns will be hidden if the terminal width is too small.

So, they will be visible as long as the terminal width is sufficient. 

We can set priority after we have more columns.

pkg/apis/config/v1alpha1/resourceinterpretercustomization_types.go
RainbowMango: So, I don't have strong opinions on this.
@XiShanYongYe-Chang Are there any documents about best practices or something?

XiShanYongYe-Chang: I don't have this best practice.

pkg/apis/policy/v1alpha1/propagation_types.go
a7i: that's a good suggestion, will apply

pkg/apis/config/v1alpha1/resourceinterpretercustomization_types.go
a7i: I was keep it consistent with the rest of karmada repo, but I can certainly do this

RainbowMango (approved):
/lgtm
/approve

pkg/apis/policy/v1alpha1/propagation_types.go
RainbowMango: Just run a test with this:
```
[rainbowmango]# kubectl get propagationpolicies.policy.karmada.io nginx-propagation 
NAME
nginx-propagation
[rainbowmango]# 
[rainbowmango]# kubectl get propagationpolicies.policy.karmada.io nginx-propagation -o wide
NAME                CONFLICT-RESOLUTION   PRIORITY
nginx-propagation   Abort                 0
```
The new columns will not be shown even the terminal width is large enough. They can be shown with `-o wide`.

@XiShanYongYe-Chang @a7i Is this behavior you want?

I think the columens is helpful, it shouldn't be hidden by default, given they are all we got by now.

XiShanYongYe-Chang: It seems that the default AGE column is missing.

My understanding is that the two new columns are not necessarily necessary for users viewing Policy. If enough users want to view the column directly (without adding -o wide), we can adjust its priority.

RainbowMango: > It seems that the default AGE column is missing.

That's bad, let me fix this.

RainbowMango: Sent the PR #5138.

Issue #4969, opened 2024-05-22
Add a label/annotation to the resource being rescheduled during failover


Dyex719:
**What would you like to be added**:
We propose adding a label/annotation during failover so that webhooks like Kyverno can perform the necessary checks/changes before the job is rescheduled. We are also open to discussing other ideas and contributing back to the community.

**Why is this needed**:
Stateful applications may need to read the last saved state to resume processing after failover. This may involve a change in the spec so that the path to read from can be specified.
It would be useful to know when a failover happened so that stateful applications can perform the necessary checks/changes before restarting.

In our particular use-case where we are migrating Flink applications using Karmada. The step by step process would be:
1. Karmada deploys the FlinkDeployment CR to a member cluster
2. The CR fails over due to a cluster failover / application failover
3. The label/annotation would get added during this process (For example: "failover" : true)
4.  When Karmada reschedules the application to a different member cluster, a webhook like Kyverno could mutate the spec by checking for this label ("failover" : true) and restart the application from the last state only if this label exists
5. Resume processing of the stateful application from the last state

chaunceyjiang:
It seems this issue is very similar to the `pause` we have been discussing before. /cc @XiShanYongYe-Chang 

https://github.com/karmada-io/karmada/issues/4688

https://github.com/karmada-io/karmada/issues/4421

XiShanYongYe-Chang:
Hi @Dyex719, thanks for your feedback.

I have a few questions. Are you referring to the labels being automatically added in by the system? When Karmada reschedules, how does the system trigger the webhook or Kverno to run?

Dyex719:
Hi @XiShanYongYe-Chang,
Thanks for getting back to us!

> Are you referring to the labels being automatically added in by the system?

Yes, Karmada would add a label after failover so that the next time it is rescheduled it would have the label.

> how does the system trigger the webhook or Kverno to run?

Kyverno would be deployed on the member clusters and is therefore called in the scheduling process. Kyverno would check if the label exists on the spec and if it does, read the last state accordingly.

XiShanYongYe-Chang:
I'sorry, I still don't understand the whole process. 

Hi @chaunceyjiang do you understand the requirement?

chaunceyjiang:
My understanding is that @Dyex719 wants a label to indicate that the current resource is undergoing failover. This is because he expects this transition state to be recognized by other third-party software.

[...]

PR #5085, opened 2024-06-22
Proposal for accurate estimator improvement for CRD scheduling

mszacillo:
**What type of PR is this?**

/kind design

Described in document.

Fixes #

Proposal doc for CRD scheduling improvements. Posting proposal following discussion in community meeting.

```release-note
NONE
```


mszacillo:
@RainbowMango - I've added this proposal to the discussion section of the community meeting tomorrow. By chance, would it be possible to move the meeting 30 minutes earlier? I've got a conflict at the moment.

RainbowMango:
> @RainbowMango - I've added this proposal to the discussion section of the community meeting tomorrow. By chance, would it be possible to move the meeting 30 minutes earlier? I've got a conflict at the moment.

I'm ok with it since this is the only topic for this meeting. I'll send a notice to the mailing group and slack channel, and gather feedback.

docs/proposals/scheduling/crd-scheduling-improvements/crd-scheduling-improvements.md
```diff
@@ -0,0 +1,138 @@
...
+# CRD Component Scheduler Estimation
+
+## Summary
+
+Users may want to use Karmada for resource-aware scheduling of Custom Resources (CRDs). This can be done
+if the CRD is comprised of a single podTemplate, which Karmada can already parse if the user defines
+the ReplicaRequirements with this in mind. Resource-aware scheduling becomes more difficult however,
+if the CRD is comprised of multiple podTemplates or pods of differing resource requirements.
+
+In the case of FlinkDeployments, there is only one podTemplate per CRD. However this podTemplate contains information
```
RainbowMango, line 24:
```suggestion
In the case of [FlinkDeployments](https://raw.githubusercontent.com/apache/flink-kubernetes-operator/release-1.9/examples/basic.yaml), there is only one podTemplate per CRD. However this podTemplate contains information
```
Maybe we can leave a link to FlinkDeployments for people who aren't familiar with it.

mszacillo:
@RainbowMango Thanks so much for reviewing this with me during the community meeting!

Just to add more context here, as I heard that is also work being done to support CRDs with multiple pod templates (like FlinkDeployment, or TensorFlow jobs for instance). For the FlinkDeployment, we cannot have replicas for the same job scheduled on different clusters - meaning we either schedule all pods on one cluster, or do not schedule at all. Once we schedule the CRD to a member cluster, all pod scheduling will be taken care of by the Flink operator.

That is why having some way to define components or distinct podTemplate resource requirements would be helpful for estimation.

docs/proposals/scheduling/multi-podtemplate-support/multiple-pod-template-support.md
```diff
@@ -0,0 +1,195 @@
...
+podTemplates to find requirements.
+
+```go
+
+type ResourceBindingSpec struct {
+
+    . . .
+
+    // The total number of replicas scheduled by this resource. Each replica will represented by exactly one component of the resource.
+    TotalReplicas int32 `json:"totalReplicas,omitempty"`
```
mszacillo, line 87: I've included this field as a replacement for the existing `Replicas` field, which is used very frequently within the Karmada codebase. Even though we are introducing the concept of components, Karmada will still ultimately be scheduling replicas - so I believe this slight refactor will make the implementation of this change less complex. This is again making an assumption that resources with more than 1 component will be scheduled on the same cluster.

docs/proposals/scheduling/multi-podtemplate-support/multiple-pod-template-support.md
```diff
@@ -0,0 +1,195 @@
...
+Currently the accurate estimator will calculate the maxReplica count by:
+1. Running the maxReplica calculation (`maxReplicas`) for each plugin enabled by the accurate estimator.
+2. The accurate estimator will then loop through all nodes and sum up the amount of replicas (`sumReplicas`) that can fit in each node. This is to account for the resource fragmentation issue.
+3. The result returned will be: `Math.min(maxReplicas, sumReplicas)`.
+
+![Accurate-Scheduler-Steps](Accurate-Scheduler-Steps.png)
+
+For the proposed implementation, please refer to the next section.
+
+### Accurate Estimator Changes
```
mszacillo, line 152: Apologies if the explanation of the implementation proposal is a little bit dense. I can format this to be a little more clear, but otherwise please let me know if you have any questions that I can clarify.

mszacillo:
Hi @RainbowMango @whitewindmills, I've gone ahead and made an update to the proposal doc with a more precise explanation of the implementation details. Please let me know if you have any comments or questions - and apologies in advance if this is a little dense, perhaps I can type this in LateX and attach an image of the algorithm specific sections. :)

docs/proposals/scheduling/multi-podtemplate-support/multiple-pod-template-support.md
```diff
@@ -0,0 +1,195 @@
...
+	- `If components > 1`: maxReplicas = (totalReplicas) * (Math.min(Available CPU Resource / sum(replica_cpu), Available Memory Resource / sum(replica_memory))).
+		- `Note`: We multiply the calculation by totalReplicas to bring the unit back to replicas. The calculation is done from the perspective of the entire CRD, but Karmada interprets replicas. That means the maxReplica calculation will always be a multiple of the number of totalReplicas.
+
+Here is an example to illustrate the case in which components > 1. Let's assume we have a CRD with `totalReplicas` = 3, with two components = {component_1: {replicas: 1, cpu: 1, memory: 2GB}, component_2: {replicas: 2, cpu: 1, memory: 1GB}}. We are estimating maxReplica count for a target cluster with a ResourceQuota that has 6CPU available and 8GB of memory available.
+
+During maxReplica estimation, we will take the sum of all resource requirement for the CRD.
+
+Total_CPU = component_1.replicas * (component_1.cpu) + component_2.replicas * (component_2.cpu) = (1 * 1) + (2 * 1) = 3 CPU.
+Total_Memory = component_1.replicas * (component_1.memory) + component_2.replicas * (component_2.memory) = (1 * 2GB) + (2 * 1GB) = 8GB.
```
Monokaix, line 173: Isn't 4G?

mszacillo: Oops, yes :)

mszacillo:
> So this design doesn't consider replica division and also is not an completely accurate calculation cause fragment issue is unavoidable, am I right?

Yes this approach would not consider divided replicas based off the use-cases we've compiled (#5115). At the moment, most CRDs get scheduled to a single cluster rather than spread across multiple. 

In terms of the precision of the calculation, you're right that it's not completely accurate. It's an estimate of how many CRDs could be fully packed on the destination cluster. However, we do guarantee that at least 1 CRD can be scheduled on a member. So at least there should never be a scenario where we schedule a CRD to a member cluster that does not have sufficient resources to hold it.

RainbowMango:
Hi @mszacillo I'm going to take two weeks off and might be slow to respond during this time. 

I believe this feature is extremely important, and I'll be focusing on this once I get back. Given this feature would get controllers, and schedulers involved, it's not that easy to come up with an ideal solution in a short time. 

By the way, I guess, with the help of [default resource interpreter of FlinkDeployment](https://github.com/karmada-io/karmada/tree/master/pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/flink.apache.org/v1beta1/FlinkDeployment)(thanks for your contribution, by the way), this probably not a blocker for you, am I right? Do you think the [Application Failover Feature](#5116) has a higher priority than this?

mszacillo:
Hi @RainbowMango, thanks for the heads up and enjoy your time off!

> this probably not a blocker for you, am I right? Do you think the [Application Failover Feature](https://github.com/karmada-io/karmada/pull/5085#5116) has a higher priority than this?

Yes, this is currently not a blocker for us. We can get around with the existing maxReplica estimation while we determine a solution for multiple podTemplate support, and we are instead focusing on the failover feature enhancements. For our MVP using Karmada we need two things:

1. Karmada should attach a label / annotation to the workload if it fails over to another cluster: #4969. This is currently implemented and we're simply doing some testing and some internal code review before opening up a PR upstream. This change will also address an existing bug that was detailed here: #4215, but guaranteeing Karmada will not reschedule workloads to the cluster it has failed-over from.

2. Address the limitation in the FederatedResourceQuota discussed in #5179, to which I am currently implementing a workaround. 

After we've completed the above tickets our order of priority will be publishing the implementation for the later steps of the failover history proposal, and then working on the multiple pod template support.

docs/proposals/scheduling/multi-podtemplate-support/multiple-pod-template-support.md
```diff
@@ -0,0 +1,195 @@
...
+type ResourceBindingSpec struct {
+
+    . . .
+
+    // The total number of replicas scheduled by this resource. Each replica will represented by exactly one component of the resource.
+    TotalReplicas int32 `json:"totalReplicas,omitempty"`
+
+    // Defines the requirements of an individual component of the resource.
+	// +optional
+    Components []Components `json:"components,omitempty"`
```
KunWuLuan, line 91: `Components []ComponentRequirements`
Right?

KunWuLuan:
Hi, did you make any progress or is there any available prototype to try?

KunWuLuan:
Will you support new replicaSchedulingType like Gang ?

docs/proposals/scheduling/multi-podtemplate-support/multiple-pod-template-support.md
```diff
@@ -0,0 +1,195 @@
...
+![Accurate-Scheduler-Steps](Accurate-Scheduler-Steps.png)
+
+For the proposed implementation, please refer to the next section.
+
+### Accurate Estimator Changes
+
+`Assumption 1`: Resources with more than one replica will always be scheduled to the same cluster.
+	- This simplifies the scope of the problem, and accounts for the fact that it is  non-trivial to schedule components of the same CRD across multiple clusters.
+
+`Assumption 2`: MaxReplica estimation will use a sum of all the resource requirement for every component's replica.
```
Monokaix, line 157: So please correct me if I was wrong，you mean that we consider all components(podTemplate) together to compute how many CRDs a cluster can deploy?The assign result is a replica number of a whole crd，not one pod，right？What I am concerned about is what the targetCluster field of `RB` returns, whether it remains the same or returns different replicas for different podtemplates.
And if that was true, is it not considered that multiple pods under one crd are deployed on different nodes?

docs/proposals/scheduling/multi-podtemplate-support/multiple-pod-template-support.md
RainbowMango: +1

RainbowMango:
> Hi, did you make any progress or is there any available prototype to try?

Not yet. I like this proposal very much, and glad to help move this forward.

RainbowMango:
> Will you support new replicaSchedulingType like Gang ?

Maybe not as Gang scheduling is already supported I guess.  Please let us know your use case if you are interested in this proposal.

KunWuLuan:
> > Will you support new replicaSchedulingType like Gang ?
> 
> Maybe not as Gang scheduling is already supported I guess. Please let us know your use case if you are interested in this proposal.

I found the proposal in #5218. I think this is what I need. We need to ensure the minimum replicas of job can be run in one cluster.

docs/proposals/scheduling/multi-podtemplate-support/multiple-pod-template-support.md
mszacillo: Will try to answer these questions, let me know if anything else needs clarification:

> we consider all components(podTemplate) together to compute how many CRDs a cluster can deploy

Correct, but just for the estimation portion. If we sum all requirements together we can estimate how many multiples of the entire CRD can be scheduled on the cluster's available resources (even if that ends up being an overestimate due to resource-partitioning). 

> assign result is a replica number of a whole crd，not one pod，right

The assign result for the resource will still be in terms of pods, since that's the atomic scheduling unit in Karmada. In terms of the targetCluster field I believe it would be the same for both podTemplates, since this design assumes that CRDs with multiple pod templates want to be scheduled to the same cluster. 

Providing an example for FlinkDeployment (for simplicity lets just estimate CPU capacity): 

```
JobManagerPodTemplate
   replicas: 1
   replicaRequirements:
       cpu: 1
TaskManagedPodTemplate
   replicas: 2
   replicaRequirements:
       cpu: 1
```

Let's assume Cluster A has 7 available CPU. For estimation purposes we sum up the total CPU = (1 + 2) = 3. Total # of CRDs that can fit will be 7 / 3 = 2. 

Once we verify that all individual components can be scheduled on available nodes (to account for resource fragmentation), we return the estimation in terms of replicas = 2 * (3) = 6 replicas. 

The Karmada scheduler will then assign the 3 total replicas of the CRD to the 6 available replicas in Cluster A. 

> And if that was true, is it not considered that multiple pods under one crd are deployed on different nodes?

Well it is possible that different pods of the same CRD are deployed on different nodes. Unless you require something like gang-scheduling, where all pods must be scheduled on the same node. Is that something you require?

mszacillo:
> I found the proposal in #5218. I think this is what I need. We need to ensure the minimum replicas of job can be run in one cluster.

Out of curiosity, which CRD are you working with? Until this pod template proposal is implemented, you could customize the way your replica's are interpreted so that the resource requirement takes the max(cpu, memory) of all your templates. Replica estimation would then be able to determine if you could schedule all your replicas on a single cluster.

Monokaix:
Hi, we also have some user case of multi pod template of volcano job(see https://github.com/volcano-sh/volcano), and we have designed a complete API of multi pod template here https://github.com/karmada-io/karmada/blob/79e374dd3840254e40d8abb33479fc81175a356b/docs/proposals/multi-template-partitionting-and-scheduler-extensibility/README.md#update-resourcebinding-api-definition, which we think is more robust, could you take a look?

Monokaix:
Here is an volcano job CRD example: https://github.com/volcano-sh/volcano/blob/master/example/MindSpore-example/mindspore_gpu/mindspore-gpu.yaml

seanlaii:
Hi @RainbowMango @Monokaix @mszacillo , 

We also have a use-case involving [RayJob](https://github.com/ray-project/kuberay/blob/master/ray-operator/apis/ray/v1alpha1/rayjob_types.go) and [RayCluster](https://github.com/ray-project/kuberay/blob/master/ray-operator/apis/ray/v1alpha1/raycluster_types.go) that requires support for multiple replica templates.
Could you please let me know whether there is a roadmap or timeline for advancing this feature? I would be delighted to help move it forward.

RainbowMango:
I'm glad to see more and more use cases require this feature. You can see this feature is planned in release-1.15 from [the backlog](https://github.com/orgs/karmada-io/projects/13). I believe it can be accelerated with your help. I will try to add some of my draft ideas to this proposal and then maybe you can iterate on this further.

RainbowMango (approved):
/lgtm
/approve
This proposal clearly explains why this feature is needed and the workable API changes. 
Regarding the estimator part, I will add an alternative to this proposal by the following PR.
In addition, the Resource Interpreter interface might be extended as well.

PR #5101, opened 2024-06-26
docs(proposal): Migration Rollback Protection

CharlesQQ:
**What type of PR is this?**

part of  https://github.com/karmada-io/karmada/issues/4709

```
NONE

```


docs/proposals/prevent-removal-managed-resources/README.md
```diff
@@ -0,0 +1,61 @@
...
+#### Story 1
+Usually, member clusters are built and pods created already, then join to karmada; For now, use create resource and pp in control plane, and resource in member cluster 
+is updated by execution controller; But find pods replicas topology or pods are not expected, The usual approach is roll back to previous state. So a way is needed where 
+member cluster resources and karmada control planes are completely independent.
+
+
+## Design Details
+
+### API changes
+- Add Annotation `karmada.io/deletionpolicy=complete` for resourcetemplate or other resource which create work object directly.
```
XiShanYongYe-Chang, line 52: I recommend using this key: `resourcetemplate.karmada.io/deletionpolicy`.

docs/proposals/prevent-removal-managed-resources/README.md
```diff
@@ -0,0 +1,61 @@
...
+- The value of `karmada.io/deletionpolicy` should be enum type for scalability
+
```
XiShanYongYe-Chang, line 54: Can you explain the namespace processing policy in the proposal? Like this comment: https://github.com/karmada-io/karmada/pull/4788#discussion_r1630573867

docs/proposals/prevent-removal-managed-resources/README.md
```diff
@@ -0,0 +1,61 @@
...
+### Dependencies resource deletion policy follow the main resource
+![Dependencies resource deletion policy](statics/dependencies-resource-deletion-policy.png)
```
XiShanYongYe-Chang, line 56: Can you expand a little on this diagram?

XiShanYongYe-Chang (commented):
Thanks a lot @CharlesQQ 

Hi, guys @RainbowMango @chaunceyjiang @whitewindmills @zhy76 @buji-code @lcw2
We can discuss here to improve the resource deletion policy.

docs/proposals/prevent-removal-managed-resources/README.md
CharlesQQ: > I recommend using this key: resourcetemplate.karmada.io/deletionpolicy.

But this Annotation not only used for resourcetemplate, such as namespace.

docs/proposals/prevent-removal-managed-resources/README.md
CharlesQQ: >  Can you explain the namespace processing policy in the proposal?

IF retain the namespace when deleted in control plane,  namespace will be keep all the time or deleted in member cluster manually by user

whitewindmills (commented):
does this annotation have a knock-on effect on dependent resources?―that is,  one adds such an annotation to his Deployment, will the annotation affect its dependent resources(like `ConfigMap`)?

docs/proposals/prevent-removal-managed-resources/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+A deletion mechanism and its API is provided to prevent cascade resource in member cluster deleted by control plane.
+
+## Motivation
+
+All resources reconciled by the execution controller have a finalizer `karmada.io/execution-controller` added to work object's metedata.
+This finalizer will prevent deletion of work object until the execution controller has a chance to perform pre-deletion cleanup.
+
+However, in some cases, user want to delete resource in control plane but resource in member cluster kept, like `kubectl delete sts xxx --cascade=orphan`, sts is deleted but pods not.
+karmada have different types of resources, different resource might have different prevent removal process.
```
whitewindmills, line 25: it is recommended to keep the first letter capitalized for `Karmada`.

docs/proposals/prevent-removal-managed-resources/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+1. resource template: cleanup the labels/annotations from the resource on member clusters and delete work object.
```
whitewindmills, line 27: clean up ...

docs/proposals/prevent-removal-managed-resources/README.md
```diff
@@ -0,0 +1,63 @@
...
+2. dependencies resource: cleanup the labels/annotations from the dependencies resource on member clusters,such as configmap/secret, delete work object.
+3. directly generate work object such as namespace/cronfederatedhpa/federaredhpa/federatedresourcequota execute the same strategy as resource template.
```
whitewindmills, line 29: I see there are some grammatical errors here, could you fix them first?

docs/proposals/prevent-removal-managed-resources/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+### Goals
+- Provides API to prevent delete cascade resource.
+- Provides command line to use feature like `karmadactl orphaningdeletion complete/disable <resource type> <name>`
```
whitewindmills, line 33: do you mean that the command `karmadactl orphaningdeletion complete` will only patch such an annotaion `resourcetemplate.karmada.io/deletionpolicy=complete` to resource templates?
if so, does this command seem too useless? can you provide such a command `karmadactl delete --cascade=orphan`?

docs/proposals/prevent-removal-managed-resources/README.md
```diff
@@ -0,0 +1,63 @@
...
+is updated by execution controller; But find pods replicas topology or pods are not expected, The usual approach is roll back to previous state. So a way is needed where 
+member cluster resources and Karmada control planes are completely independent.
+
+
+## Design Details
+
+### API changes
+- Add Annotation `resourcetemplate.karmada.io/deletionpolicy=complete` for resourcetemplate or other resource which create work object directly.
+- The value of `karmada.io/deletionpolicy` should be enum type for scalability
+    - complete: cleanup the labels/annotation from the resource on member clusters and delete work object.
```
chaunceyjiang, line 52: I believe that when clearing labels or annotations in the resource template, only clear the keys recorded in `resourcetemplate.karmada.io/managed-labels` or `resourcetemplate.karmada.io/managed-annotations`.

docs/proposals/prevent-removal-managed-resources/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+All resources reconciled by the execution controller have a finalizer `karmada.io/execution-controller` added to work object's metedata.
+This finalizer will prevent deletion of work object until the execution controller has a chance to perform pre-deletion cleanup.
+
+However, in some cases, user want to delete resource in control plane but resource in member cluster kept, like `kubectl delete sts xxx --cascade=orphan`, sts is deleted but pods not.
+Karmada have different types of resources, different resource might have different prevent removal process.
+
+1. resource template: clean up the labels/annotations from the resource in member clusters and delete work object.
+2. dependent resource: clean up the labels/annotations from the dependent resource in member clusters,such as configmap/secret,  and then delete work object.
+3. resource which directly generate work object such as namespace/cronfederatedhpa/federaredhpa/federatedresourcequota execute the same strategy as resource template.
```
chaunceyjiang, line 29: What does `same strategy` mean?

I think namespace is different from the other resources, I believe we need to explain namespace separately.

Not all namespaces directly generate work. They can also be matched by some PP. (For example, I have already turned off the ns-controller.)

docs/proposals/prevent-removal-managed-resources/README.md
```diff
@@ -0,0 +1,63 @@
...
+## Motivation
+
+All resources reconciled by the execution controller have a finalizer `karmada.io/execution-controller` added to work object's metedata.
+This finalizer will prevent deletion of work object until the execution controller has a chance to perform pre-deletion cleanup.
+
+However, in some cases, user want to delete resource in control plane but resource in member cluster kept, like `kubectl delete sts xxx --cascade=orphan`, sts is deleted but pods not.
+Karmada have different types of resources, different resource might have different prevent removal process.
+
+1. resource template: clean up the labels/annotations from the resource in member clusters and delete work object.
+2. dependent resource: clean up the labels/annotations from the dependent resource in member clusters,such as configmap/secret,  and then delete work object.
```
chaunceyjiang, line 28: image
When we delete the `deploymentA` of the control plane, do we support retaining the `cm` in `member1` cluster, but deleting the `cm` in `member2` cluster?

RainbowMango: I don't think so, at least for now based on the use cases we collected.

By the way, a little confused by the `deploymentB` in the picture, are you trying to say the case that a configmap shared with multiple deployments?

docs/proposals/prevent-removal-managed-resources/README.md
```diff
@@ -0,0 +1,63 @@
...
+- Provides API to prevent delete cascade resource.
+- Provides command line to use feature like `karmadactl orphaningdeletion complete/disable <resource type> <name>`
+
+### Non-Goals/Future Work
+- Other delete policy, like retain work object.
+
+## Proposal
+### User stories
+
+#### Story 1
```
RainbowMango, line 41:
```suggestion
#### Story 1: As an administrator, I want a rollback mechanism during workload migration to Karmada, so that I can I can revert to the pre-migration state in case of any unexpected situation. 
```

docs/proposals/prevent-removal-managed-resources/README.md
chaunceyjiang: > a configmap shared with multiple deployments?

Yes.

This image mainly wants to express two things:
1. Can we control whether dependent resources are kept separately in the member clusters?
2. Can we control whether resources are kept in specific member clusters?

docs/proposals/use-cascading-deletion/README.md
```diff
@@ -0,0 +1,137 @@
...
+### Goals
+<!--
+- 提供只删除控制面资源模板,而成员集群资源不删除,同时清理Karmada附加给成员集群资源的labels/annotations的能力
+- 提供控制面资源模版删除之后,不阻塞等待成员集群资源删除的能力
+- 其他直接产生work对象的资源如namespace(ns-controller开启)/cronfederatedhpa/federaredhpa/federatedresourcequota,需要有和资源模板相同的能力
+- 提供karmadactl命令行,可以执行级联删除策略, 如 `karmadactl delete deployment --cascade=orphan`
+-->
+
+- Provides the ability to delete only the control plane resource templates without deleting the member cluster resources, and at the same time clean up the labels/annotations attached to the member cluster resources by Karmada.
+- Provides the ability to not block waiting for member cluster resource deletion after the control plane resource template is deleted.
```
XiShanYongYe-Chang, line 41: I am sorry, but what does it mean to not block waiting for member cluster resource deletion?

docs/proposals/use-cascading-deletion/README.md
```diff
@@ -0,0 +1,137 @@
...
+<!--
+作为管理员, 我希望可以有一种手段可以将某个测试集群从Karmada控制面移除,同时测试集群上面的workload不受影响
+-->
+As an administrator, I hope there is a way to remove a test cluster from the Karmada control plane without affecting the workload on the test cluster.
+
+#### Story 3
+<!--
+作为管理员,为了降低接口耗时,我希望在删除控制面资源模板之后可以立即返回,不阻塞等待直到成员集群资源删除。
+-->
+As an administrator, in order to reduce the interface time-consuming, I hope that I can return immediately after deleting the control plane resource template without blocking and waiting until the member cluster resources are deleted.
```
XiShanYongYe-Chang, line 73: I have some confusion about this point, what exactly is meant by "interface time-consuming"? How can the reduction in consumption of the current plan be measured?

docs/proposals/use-cascading-deletion/README.md
```diff
@@ -0,0 +1,137 @@
...
+  - orphan: 删除work对象,清理Karmada附加给成员集群的workload的annotation/labels,不删除成员集群workload对象.
+  - foreground: 默认情况前台删除,等同于不存在上述注解或者注解的值为空;删除动作会阻塞等待直到成员集群workload被删除掉.
+  - background: 后台删除表示删除资源模板之后立即返回,由控制器在后台逐步清理成员集群workload.
+-->
+
+- Since the object of the deletion operation is the resource template, add Annotation to the resource template, named: `resourcetemplate.karmada.io/cascadedeletion`
+- For scalability, value should be an enum type
+    - orphan: Delete the work object, clean up the annotation/labels of the workload attached by Karmada to the member cluster, and do not delete the member cluster workload object.
+    - foreground: By default, the foreground is deleted, which means that the above annotation does not exist or the value of the annotation is empty; the deletion action will block and wait until the member cluster workload is deleted.
+    - background: Background deletion means returning immediately after deleting the resource template, and the controller gradually cleans up the member cluster workload in the background.
```
XiShanYongYe-Chang, line 94: The current behavior of the system is like this, right?

docs/proposals/use-cascading-deletion/README.md
```diff
@@ -0,0 +1,137 @@
...
+- User triggers deletion action
+- execution-controller performs corresponding actions according to the cascade deletion policy
+  ![use-cascading-deletion](statics/use-cascading-deletion.png)
+
+### The cascading deletion policy of dependent resources and main resources does not force binding
+<!--
+依赖资源和主资源的级联删除策略不强制绑定
+由于依赖资源可能被多个资源模版共享,在这种情况下很难决策依赖资源的删除策略以哪个删除策略为准; 不强制和主资源绑定,由用户自己决策,灵活性和扩展性更好
+-->
+Since dependent resources may be shared by multiple resource templates, in this case it is difficult to decide which deletion strategy should be used for the dependent resources; it is not forced to be bound to the main resource, and is left to the user to decide, with greater flexibility and scalability. good
```
XiShanYongYe-Chang, line 120: This could also easily lead to such a situation where the deletion policy for the following resources (such as ConfigMap) is set to cascade deletion, but the deletion policy for the leading resources (such as Deployment) is set to retention. When users delete the following resources, it causes the applications in the member clusters to become abnormal.

docs/proposals/use-cascading-deletion/README.md
```diff
@@ -0,0 +1,316 @@
...
+      kind: Deployment
+      name: foo
+      namespace: default
+```
+
+### Solution Four: Extended by Annotation & Extend the fields of PropagationPolicy/ClusterPropagationPolicy
+
+Equivalent to supporting both solution one and solution two
+
+### Solution comparison
```
XiShanYongYe-Chang, line 278: Add some pros and cons analysis.
```
Solution One:
Disadvantages:
- When the execution-controller determines whether to cascade delete resources in the member clusters, it needs to parse the resources in the Work template from the manifest, which adds an extra parsing overhead.

Solution Two:
Disadvantages:
- For resources that are not distributed through PropagationPolicy, such as namespace, federatedresourcequota, it is not possible to specify a deletion policy.
```

XiShanYongYe-Chang (commented):
Also ask @RainbowMango to help take a look.
/cc @RainbowMango

docs/proposals/use-cascading-deletion/README.md
```diff
@@ -0,0 +1,320 @@
...
+| Name           | Supported control plane resources | Extend API resources | User learning cost |
```
chaosi-zju, line 275: I didn't got what is the different below the title "Supported control plane resources".

docs/proposals/use-cascading-deletion/README.md
```diff
@@ -0,0 +1,320 @@
...
+|----------------|-----------------------------------|----------------------|--------------------|
+| Solution One   | all resources                     | None                 | lowest             |
+| Solution Two   | resource template                 | PP/CPP/RB/CRB/WORK   | lowest             |
+| Solution Three | all resources                     | new CRD/WORK         | Highest            |
+| Solution Four  | all resources                     | PP/CPP/RB/CRB/WORK   | lower              |
```
chaosi-zju, line 280: I think we can highlight another `comparison` like solution one can set the `orphan` attribute at the granularity of each resource template, while solution two can only set the orphan` attribute at the granularity of policy. In 1 policy vs multi resource scene, we cann't execute orphan delete just by per resource.

docs/proposals/use-cascading-deletion/README.md
```diff
@@ -0,0 +1,321 @@
...
+  cascadeDeletion: orphan
+```
+### Solution Three: Extended by adding new CRD
+<!--
+通过CRD定义控制面资源级联删除的策略, 好处是无需对PP和RB的字段进行扩展
+在控制面资源生成work对象的时候, 根据CascadeDeletionPolicy的级联删除策略设置work的spec.cascadeDeletion字段内容
+-->
+
+Define the policy for cascading deletion of control plane resources through CRD. The advantage is that there is no need to expand the fields of PP and RB.
+When the control plane resource generates a work object, set the spec.cascadeDeletion field content of the work according to the cascade deletion policy of CascadeDeletionPolicy.
```
RainbowMango, line 180: What's the disadvantage of this approach? Why we didn't choose it?

docs/proposals/use-cascading-deletion/README.md
```diff
@@ -0,0 +1,321 @@
...
+  resourceSelectors:
+    - apiVersion: apps/v1
+      kind: Deployment
+      name: foo
+      namespace: default
+```
+
+### Solution Four: Extended by Annotation & Extend the fields of PropagationPolicy/ClusterPropagationPolicy
+
+Equivalent to supporting both solution one and solution two
```
RainbowMango, line 272: ditto. Why we didn't proceed with this approach?

docs/proposals/use-cascading-deletion/README.md
```diff
@@ -0,0 +1,321 @@
...
+
+## Proposal
+
+### User Stories (Optional)
+
+#### Story 1
+<!--
+作为管理员，我希望在工作负载迁移到Karmada期间有一个回滚机制，以便在出现任何意外情况时可以恢复到迁移前的状态。
+-->
+As an administrator, I would like to have a rollback mechanism during workload migration to Karmada so that in case of any unexpected situation, I can revert to the pre-migration state.
```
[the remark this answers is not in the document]
grosser: For example the migration to karmada creates bug or update loops when 2 controllers are fighting, then we want to "just roll it back", meaning we want to un-manage all newly managed resources.

docs/proposals/use-cascading-deletion/README.md
```diff
@@ -0,0 +1,321 @@
...
+metadata:
+  name: nginx-propagation
+spec:
+  resourceSelectors:
+    - apiVersion: apps/v1
+      kind: Deployment
+      name: nginx
+  cascadeDeletion: orphan
+```
+### Solution Three: Extended by adding new CRD
```
grosser, line 173:
```suggestion
### Solution Three: Extended by adding new CRD field
```

grosser (commented):
We could start with "here is a magic annotations",
just to get the most important code landed.
(or go straight to "here is a new work spec field")
but not include pp/cpp/rb/crb changes in the initial PR to keep things simple and let ppl play with the actual "does the cascading work as expected"
it can be an alpha or undocumented feature until it's "complete"

docs/proposals/use-cascading-deletion/README.md
```diff
@@ -0,0 +1,321 @@
...
+#### API changes
+<!--
+- 为资源模板添加Annotation, 名称为: `resourcetemplate.karmada.io/cascadedeletion`
+- 为了可扩展性,value应该是一个枚举类型
+  - orphan: 删除work对象,清理Karmada附加给成员集群的workload的annotation/labels,不删除成员集群workload对象.
+-->
+
+- Since the object of the deletion operation is the resource template, add Annotation to the resource template, named: `resourcetemplate.karmada.io/cascadedeletion`
+- For scalability, value should be an enum type
+    - orphan: Delete the work object, clean up the annotation/labels of the workload attached by Karmada to the member cluster, and do not delete the member cluster workload object.
```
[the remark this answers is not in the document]
XiShanYongYe-Chang: The reason why this enumeration value was not added before is that when the annotation does not exist on the resource, it behaves.

docs/proposals/use-cascading-deletion/README.md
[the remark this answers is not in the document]
RainbowMango: It makes sense. In that case, you might just want to stop Karmada syncing to member users.

But, the new feature [cluster-level resource propagation pause](https://github.com/karmada-io/karmada/tree/master/docs/proposals/dispatch-suspension) seems can do it as well. During the migration, there must be a PropagationPolicy(or ClusterPropagationPolicy), if you want to stop syncing, you can use the suspension capability, like:
```yaml
kind: PropagationPolicy
metadata:
  name: nginx-propagation
spec:
  ...
  suspension:
    suspendDispatching: true
```
What do you think? 
cc @XiShanYongYe-Chang here as well.

docs/proposals/use-cascading-deletion/README.md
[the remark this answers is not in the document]
RainbowMango: `delete` is the default behavior. Do we need to declare it particularly?

docs/proposals/use-cascading-deletion/README.md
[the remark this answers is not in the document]
XiShanYongYe-Chang: If user want to analyze the cause of a fight between two controllers, we can use the pause feature to cover this scenario. 

After the cause analysis is complete, user want to unfederate resources (as opposed to adding or federation resources), this scenario can be be covered by the resource deletion policy.

Therefore, the latter may look more like a user story.

docs/proposals/use-cascading-deletion/README.md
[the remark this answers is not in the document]
grosser: not important, but would be nice it it was an allowed value, but mostly for readability when managing lots of resources and being able to say "each one needs to have the annotation set" so nothing is forgotten by accident

XiShanYongYe-Chang:
Made some modifications and additions based on the comments. 
cc @RainbowMango @chaosi-zju

XiShanYongYe-Chang:
> I'm now more tend to `Extended by Annotation & Extend the fields of Work`

You can think through the expansion of solution 1.

Solution 4 not only extends work. For the current proposal, it is a combination of two solutions, which I think is a bit redundant.

CharlesQQ:
> I see, what do the author @CharlesQQ thinik of https://github.com/karmada-io/karmada/pull/5101#discussion_r1739589723?

I'm now more tend to solution 1.

docs/proposals/use-cascading-deletion/README.md
```diff
@@ -0,0 +1,397 @@
...
+| Name           | Supported control plane resources | Extend API resources | User learning cost | Execute orphan delete just by per resource |
+|----------------|-----------------------------------|----------------------|--------------------|--------------------------------------------|
+| Solution One   | all resources                     | None                 | Lowest             | YES                                        |
+| Solution Two   | resource template                 | PP/CPP/RB/CRB/WORK   | Lowest             | No                                         |
+| Solution Three | all resources                     | new CRD/WORK         | Highest            | YES                                        |
+| Solution Four  | all resources                     | PP/CPP/RB/CRB/WORK   | Lower              | YES                                        |
+
+Solution One:
+Disadvantages:
+- When the `execution-controller` determines whether to cascade delete resources in the member clusters, it needs to parse the resources in the Work template from the manifest, which adds an extra parsing overhead.
```
RainbowMango, line 352: For now, during the process of deleting the workload, the `execution-controller` unmarshals the manifest already, does this approach adds extra overhead?
https://github.com/karmada-io/karmada/blob/b27e669f481a683fa4eb6dec6066e77ccf67a5d2/pkg/controllers/execution/execution_controller.go#L165-L182

XiShanYongYe-Chang: Good point. For this approach, no extra parsing overhead is required.

docs/proposals/use-cascading-deletion/README.md
```diff
@@ -0,0 +1,381 @@
...
+
+reviewers:
+- "@robot"
+- TBD
+
+approvers:
+- "@robot"
+- TBD
+
+creation-date: 2024-07-01
```
RainbowMango, line 15:
```suggestion
---
title: Use Cascading Deletion in Karmada

authors:
- "@CharlesQQ"

reviewers:
- "@robot"
- TBD

approvers:
- "@robot"
- TBD

creation-date: 2024-07-01

---
```
Please update the reviewers by adding the guys who reviewed this proposal.

docs/proposals/use-cascading-deletion/README.md
```diff
@@ -0,0 +1,388 @@
+---
+title: Use Cascading Deletion in Karmada
```
RainbowMango, line 2: How about naming this feature as `Migration Rollback Protection`? 

Forget `cascading` to avoid misleading.

XiShanYongYe-Chang: +1

XiShanYongYe-Chang:
cc @RainbowMango 
Help take a review again.

docs/proposals/migration-rollback-protection/README.md
```diff
@@ -0,0 +1,393 @@
...
+---
+
+# Migration Rollback Protection
+
+## Summary
+
+<!--
+提供一种联邦资源的删除策略，可供用户选择当删除联邦层工作负载时，是否同步删除成员集群中的工作负载。
+此设置在工作负载迁移场景中特别有用，可以确保在不影响成员集群上运行的工作负载的情况下快速执行回滚。
+-->
```
RainbowMango, line 25: Please remove all Chinese characters.

docs/proposals/migration-rollback-protection/README.md
```diff
@@ -0,0 +1,393 @@
...
+
+This setting is particularly useful during workload migration scenarios to ensure that rollback can occur quickly without affecting the workloads running on the member clusters.
+
+## Motivation
+
+<!--
+Karmada 系统的当前行为是这样的，当用户删除 Karmada 控制面的资源时，会同步删除成员集群中被分发的资源。当在某些场景下，例如工作负载迁移回滚场景，用户希望能够保留成员集群中的工作负载。
+-->
+
+The current behavior of the Karmada system is that when a user deletes resources from the Karmada control plane, the distributed resources in the member clusters are also deleted synchronously. However, in certain scenarios, such as workload migration rollout scenarios, users may wish to retain the workloads in the member clusters.
```
RainbowMango, line 37:
```suggestion
The current behavior of the Karmada system is that when a user deletes resources from the Karmada control plane, the distributed resources in the member clusters are also deleted synchronously. However, in certain scenarios, such as workload migration scenarios, users may wish to retain the workloads in the member clusters.
```

RainbowMango (approved):
/lgtm
/approve

PR #5102, opened 2024-06-26
multi cluster cache list and get skip problem clusters

yanfeng1992:

**What type of PR is this?**
/kind bug

When some member clusters are in the ready state but the member cluster apiserver has problems, search cannot obtain search results through /apis/search.karmada.io/v1alpha1/proxying/karmada/proxy

log in karmada-search

```
E0625 09:15:01.629270       1 cacher.go:450] cacher (xxxx): unexpected ListAndWatch error: failed to list xxx, Kind=xxxx: Get "https://10.253.48.22:6443/apis/xxx.io/v1alpha1/xxxx?limit=10000": http: server gave HTTP response to HTTPS client; reinitializing...

W0625 09:15:02.635162       1 reflector.go:424] storage/cacher.go:xxx: failed to list xxx, Kind=xxx Get "https://10.253.48.22:6443/apis/xxx.io/v1alpha1/xxx?limit=10000": http: server gave HTTP response to HTTPS client
```

Fixes #

```release-note
`karmada-search`: multi cluster cache list and get skip problem clusters
```

### DIFF
--- a/pkg/search/proxy/store/multi_cluster_cache.go
+++ b/pkg/search/proxy/store/multi_cluster_cache.go
@@ -264,14 +264,16 @@ func (c *MultiClusterCache) List(ctx context.Context, gvr schema.GroupVersionRes
 		for _, cluster := range clusters {
 			_, _, err := listFunc(cluster)
 			if err != nil {
-				return nil, err
+				klog.Errorf("Fail to list %s from %s: %v", gvr, cluster, err)
+				continue
 			}
 		}
 	} else {
 		for clusterIdx, cluster := range clusters {
 			n, cont, err := listFunc(cluster)
 			if err != nil {
-				return nil, err
+				klog.Errorf("Fail to list %s from %s: %v", gvr, cluster, err)
+				continue
 			}
 
 			options.Limit -= int64(n)
@@ -407,7 +409,8 @@ func (c *MultiClusterCache) getResourceFromCache(ctx context.Context, gvr schema
 			continue
 		}
 		if err != nil {
-			return nil, "", nil, fmt.Errorf("fail to get %v %v from %v: %v", namespace, name, clusterName, err)
+			klog.Errorf("Fail to get %v %v from %v: %v", namespace, name, clusterName, err)
+			continue
 		}
 		findObjects = append(findObjects, obj)
 		findCaches = append(findCaches, cache)


XiShanYongYe-Chang:
> When some member clusters are in the ready state but the member cluster apiserver has problems,

Hi @yanfeng1992, do you mean that the health check of the member cluster kube-apiserver is normal, but other APIs cannot be accessed? When will that happen?

yanfeng1992:
> > When some member clusters are in the ready state but the member cluster apiserver has problems,
> 
> Hi @yanfeng1992, do you mean that the health check of the member cluster kube-apiserver is normal, but other APIs cannot be accessed? When will that happen?

In the case of pull+anp, anp has a problem
host->member apiserver not ok
member -> host karmada apiserver is ok
At this time, the cluster is healthy
but search will have problems

@XiShanYongYe-Chang

XiShanYongYe-Chang:
What does `host` mean?

yanfeng1992:
host cluster，karmada control plane cluster

pkg/search/proxy/store/multi_cluster_cache.go
```diff
@@ -264,14 +264,16 @@ func (c *MultiClusterCache) List(ctx context.Context, gvr schema.GroupVersionRes
 		for _, cluster := range clusters {
 			_, _, err := listFunc(cluster)
 			if err != nil {
-				return nil, err
+				klog.Errorf("Fail to list %s from %s: %v", gvr, cluster, err)
```
XiShanYongYe-Chang, line 267: The error returned may not be caused by network issues; would it be inappropriate to handle it this way?

XiShanYongYe-Chang (commented):
Sorry for reply late. Ask @ikaven1024 to help take a look.
/cc @ikaven1024

pkg/search/proxy/store/multi_cluster_cache.go
yanfeng1992: > The error returned may not be caused by network issues; would it be inappropriate to handle it this way?

What I am worried about is that if there is a problem with one cluster, an error will be returned here, which will cause all clusters to fail to find results.

In addition, I found that the cache is also processed by continue

https://github.com/karmada-io/karmada/blob/992802084f06a8ddf195577ebe988c26965a7d46/pkg/registry/search/storage/cache.go#L133-L138

XiShanYongYe-Chang: > What I am worried about is that if there is a problem with one cluster, an error will be returned here, which will cause all clusters to fail to find results.

This is actually a unification of the behavior of the list and watch APIs. When an internal error occurs, the entire request fails. Therefore, for errors such as cluster network unavailability, should we prioritize handling this error first, and not modify the behavior of other errors for now?

> In addition, I found that the cache is also processed by continue

This does not necessarily indicate that the cache's handling method is better; these two features are currently independent of each other. I feel that we can consider these two features separately for now.

xigang:
I understand that we should not simply handle problematic clusters by using continue. if a cluster fails, we should inform the users.

yanfeng1992:
> > I understand that we should not simply handle problematic clusters by using continue. if a cluster fails, we should inform the users.
> > cc @ikaven1024
> 
> I am agreed to this point. Mostly listing gets objets from cache, occasional disconnect to member clusters will not return fail. And if the disconnect is always, it should be maintained by operators.

Is there a better way that does not affect list query and can prompt operators? @xigang @ikaven1024

xigang:
When a member cluster fails, should the pods in that cluster still respond to user requests? If they should respond, how can the client be informed that these pods belong to a failed cluster?

PR #5113, opened 2024-07-01
cleanup helm chart

chaosi-zju:
**What type of PR is this?**

/kind cleanup
/kind bug

cleanup helm chart:
* set `karmada-metrics-adapter` image pull policy to `karmadaImagePullPolicy`, in order to keep the same as other components
* fix a occasional bug in `post-install-job`, detail refers to code comments

Fixes #

```release-note

```

### DIFF
--- a/charts/karmada/templates/post-install-job.yaml
+++ b/charts/karmada/templates/post-install-job.yaml
@@ -50,6 +50,13 @@ spec:
         - |
           bash <<'EOF'
           set -ex
+
+          # The `post-install hook job` may be applied before all karmada components are ready that rely on `static-resource job` and `hook-job secret`.
+          # So, we have to postpone the deletion of the `static-resource job` and `hook-job secret` until all karmada components are up and running.
+          while [[ $(kubectl get pods -n {{ $namespace }} --field-selector=status.phase!=Running,status.phase!=Succeeded -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}{end}' | wc -l) > 0 ]]; do
+            echo "waiting for all pods of karmada control plane ready..."; sleep 1;
+          done
+
           kubectl delete job {{ $name }}-static-resource -n {{ $namespace }}
           kubectl delete secret {{ $name }}-hook-job -n {{ $namespace }}
           EOF
--- a/charts/karmada/values.yaml
+++ b/charts/karmada/values.yaml
@@ -511,10 +511,7 @@ metricsAdapter:
     registry: docker.io
     repository: karmada/karmada-metrics-adapter
     tag: *karmadaImageVersion
-    ## Specify a imagePullPolicy
-    ## Defaults to 'Always' if image tag is 'latest', else set to 'IfNotPresent'
-    ##
-    pullPolicy: Always
+    pullPolicy: *karmadaImagePullPolicy
     ## Optionally specify an array of imagePullSecrets.
     ## Secrets must be manually created in the namespace.
     ## Example:


chaosi-zju:
@calvin0327 can you please help review this?

charts/karmada/templates/post-install-job.yaml
```diff
@@ -50,6 +50,15 @@ spec:
...
           bash <<'EOF'
           set -ex
+
+          # notes: the post-install hook job may be applied before all pods ready who relied on job/secret to be deleted,
+          # actually, thoses pods may be running init-container while main-container haven't be created, so here wait for main-container created.
+          # the sub-command `sed '1d'` is to remove the title line
+          # the sub-command `grep -vE 'Running|Completed'` is to exclude the Running or Completed pods
+          while [[ $(kubectl get pod -n {{ $namespace }} | sed '1d' | grep -vE 'Running|Completed' | wc -l) > 0 ]]; do
+            echo "waiting for all pods in karmada control plane ready..."; sleep 5;
+          done
```
RainbowMango, line 60: Is this all about guaranteeing the static-resource job to be executed before it gets cleaned up?

charts/karmada/templates/post-install-job.yaml
```diff
@@ -50,6 +50,15 @@ spec:
         - |
           bash <<'EOF'
           set -ex
+
+          # notes: the post-install hook job may be applied before all pods ready who relied on job/secret to be deleted,
+          # actually, thoses pods may be running init-container while main-container haven't be created, so here wait for main-container created.
+          # the sub-command `sed '1d'` is to remove the title line
+          # the sub-command `grep -vE 'Running|Completed'` is to exclude the Running or Completed pods
+          while [[ $(kubectl get pod -n {{ $namespace }} | sed '1d' | grep -vE 'Running|Completed' | wc -l) > 0 ]]; do
```
RainbowMango, line 58:
```suggestion
          while [[ $(kubectl get pods -n {{ $namespace }} --field-selector=status.phase!=Running,status.phase!=Succeeded -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}{end}' | wc -l) > 0 ]]; do
```
Probably more straightforward than sed and grep.

charts/karmada/templates/post-install-job.yaml
```diff
@@ -50,6 +50,15 @@ spec:
...
+            echo "waiting for all pods in karmada control plane ready..."; sleep 5;
```
RainbowMango, line 59:
```suggestion
            echo "waiting for all pods of karmada control plane ready..."; sleep 5;
```
Another question, can we shorten the sleep time? Given the `kubectl` command is cheap enough, how about shortening to 1s?

charts/karmada/templates/post-install-job.yaml
chaosi-zju: Not very accurate, main purpose is to prevent the following timing events from occurring:
1. karmada components, i.e: karmada-controller-manager, be appiled, attention please, just appiled, not running.
2. post-install-job running, and delete the `static-resource job` and `hook-job secret` quickly.
3. karmada components, i.e: karmada-controller-manager, failed with lack of `hook-job secret` or its `kubectl wait --for=condition=complete job xx-static-resource` never meets.

So, we expect the deletion of  `static-resource job` and `hook-job secret` should be delayed to when we never need them, that is, all karmada components running.

charts/karmada/templates/post-install-job.yaml
chaosi-zju: nice

charts/karmada/templates/post-install-job.yaml
chaosi-zju: > Given the kubectl command is cheap enough, how about shortening to 1s?
ok

charts/karmada/templates/post-install-job.yaml
chaosi-zju: done and tested ok

charts/karmada/templates/post-install-job.yaml
chaosi-zju: done and tested ok

charts/karmada/templates/post-install-job.yaml
```diff
@@ -50,6 +50,14 @@ spec:
         - |
           bash <<'EOF'
           set -ex
+
+          # The `post-install hook job` may be applied before all karmada components ready who relied on `static-resource job` and `hook-job secret`,
+          # So, we expect the deletion of `static-resource job` and `hook-job secret` should be delayed to when we never need them,
+          # that is, all karmada components running.
```
RainbowMango, line 56:
```suggestion
          # The `post-install hook job` may be applied before all karmada components are ready that rely on `static-resource job` and `hook-job secret`.
          # So, we have to postpone the deletion of the `static-resource job` and `hook-job secret` until all karmada components are up and running.
```

RainbowMango (approved):
/lgtm
/approve

PR #5114, opened 2024-07-01
Improve code reusability

rxy0210:
**What type of PR is this?**

Improve code reusability
Fixes #

```release-note
NONE
```

### DIFF
--- a/operator/pkg/certs/certs.go
+++ b/operator/pkg/certs/certs.go
@@ -347,10 +347,6 @@ func NewSignedCert(cc *CertConfig, key crypto.Signer, caCert *x509.Certificate, 
 	}
 
 	RemoveDuplicateAltNames(&cc.Config.AltNames)
-	notAfter := time.Now().Add(constants.CertificateValidity).UTC()
-	if cc.NotAfter != nil {
-		notAfter = *cc.NotAfter
-	}
 
 	certTmpl := x509.Certificate{
 		Subject: pkix.Name{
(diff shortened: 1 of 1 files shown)


operator/pkg/certs/certs.go
```diff
@@ -347,10 +347,8 @@ func NewSignedCert(cc *CertConfig, key crypto.Signer, caCert *x509.Certificate,
 	}
 
 	RemoveDuplicateAltNames(&cc.Config.AltNames)
-	notAfter := time.Now().Add(constants.CertificateValidity).UTC()
-	if cc.NotAfter != nil {
-		notAfter = *cc.NotAfter
-	}
+
+	cc.defaultNotAfter()
```
RainbowMango, line 351: Seems the `cc.NotAfter` already be initilized at https://github.com/karmada-io/karmada/blob/dee30c821b042984a40234f3291d2e034a6f7f57/operator/pkg/certs/certs.go#L294

So, I guess this piece of code is redundant, right?

rxy0210: Yes, it seems redundant. But I think the writers of `func NewSignedCert()` probably added this initialization code because they thought this method would be called elsewhere.

RainbowMango: Do you mean we should keep the code(initialize it again) here?

rxy0210: yes. I think it's possible that this func could be called somewhere else.

calvin0327: @rxy0210 Thanks for pointing out this. your current changes are completely correct for the existing logic.

However, for the `NewSignedCert` method, it's more like a general method. in this, `certConfig` should already be fully initalized, and its values should not be changed futher. I thinks the `defaultNotAfter` method should be called earlier， rather than at this method. 

Therefore, I think your changes are not meaningful.

rxy0210: ok. I remove redundant code.

RainbowMango (approved):
/lgtm
/approve

Thanks @rxy0210 for doing this.

PR #5116, opened 2024-07-01
Stateful Failover Proposal

### LINKED ISSUE #5006: Advice on how to restore state after failover for stateful applications
We are using Karmada's failover and scheduling features to migrate our Flink jobs from one cluster to another during failover. Since Flink like other stateful applications need to resume from the last processed state, we need to make a call to a persistent storage system like S3/HDFS to restore the latest state from there.

Since the process of restoring state from a persistent store will be a common theme for all stateful applications, we want to know if the community is open to adding this as a feature in Karmada itself. This will prevent the user having to implement their own custom solution using other third party tools.

**Environment**:
- Karmada version:
v1.9.0
- Kubernetes version:
v1.29

Dyex719:
**What type of PR is this?**
Proposal for stateful failover

Explained in the doc

Fixes #5006, #4969

N/A

NONE

### DIFF
--- a/docs/proposals/stateful-failover/stateful-failover-proposal.md
+++ b/docs/proposals/stateful-failover/stateful-failover-proposal.md
@@ -1,0 +1,197 @@
+---
+title: Stateful Failover Support in Karmada
+authors:
+  - "@Dyex719"
+  - "@mszacillo"
+reviewers:
+- "@RainbowMango"
+- "@XiShanYongYe-Chang"
+- "@zhzhuang-zju"
+approvers:
+- "@RainbowMango"
+
+create-date: 2024-06-28
+
+---
+
+# Stateful Failover Support in Karmada
+
+## Summary
+
+Karmada can be currently used to intelligently schedule all types of resources (both generic Kubernetes objects as well as user-applied CRDs). It is particularly useful for ensuring stateful application resilience in a multi-cluster environment in which applications may be rescheduled if a cluster becomes healthy.
+
+However, Karmada’s scheduling logic runs on the assumption that resources that are scheduled and rescheduled are stateless. In some cases, users may desire to conserve a certain state so that applications can resume from where they left off in the previous cluster. 
+
+For CRDs dealing with data-processing (such as Flink or Spark), it can be particularly useful to restart applications from a previous checkpoint. That way applications can seamlessly resume processing data while avoiding double processing. 
+
+This proposal would like to introduce a more generalized way for users to define failover actions, which can be used in the case of stateful CRD failover.
+
+
+## Motivation
+
+This proposal aims to provide a framework for stateful CRDs to keep track of the state required during failover so that processing can be resumed from that state after failover has completed.
+
+### Goals
+
+To add support for stateful application failover:
+- Extend ResourceBinding API to include FailoverHistory field
+- Extend PropagationPolicy API to include status items that should be persisted during failover
+
+## Proposal
+
+Stateful applications need a way to read the last saved state to resume processing from that state after failover. 
+
+To enable this we split this proposal into two sections: 
+- A mechanism for when a failover happened so that stateful applications can load the previous state first before resuming processing.
+- At the very minimum, a way of storing and reading state/metadata related to the job being failed over.
+
+Since stateful applications have different implementations of how to retrieve the last state given this the job metadata we would then rely on those individual implementations to fetch all the details related to the last state. 
+
+**NOTE: To be discussed with Karmada community**
+
+One important detail is that if all the replicas of the stateful application are not migrated together, it is not clear when the state needs to be restored. In this proposal we focus on the use case where all the replicas of a stateful application are migrated together. One way to ensure this is to make all the replicas scheduled together using spreadConstraints.
+
+```
+spreadConstraints:
+  - spreadByField: cluster
+    maxGroups: 1
+    minGroups: 1
+```
+
+
+### Use-case with Flink:
+
+In Flink for example, checkpoints are snapshots that store the state of the application till a particular moment in time. As a result they contain information about all the records processed till that moment. This information is collected and persisted continuously to some persistent storage at specific intervals.
+
+In Flink, to retrieve the last state from the persistent store, we would require metadata about the job being restored (Job ID) along with the path where the state is being persisted. 
+
+The Flink Operator can then use this information to retrieve the last state by providing a few modifications to the Flink CRD spec.
+
+```
+Spec:
+Flink Configuration:
+	initialSavepointPath: “desired checkpoint path to be resumed from (s3p://)”
+upgradeMode: savepoint 
+state: running
+```
+
+We currently use a mutating webhook (Kyverno) along with a custom microservice to retrieve the last state from the persistent store and mutate the CRD spec to to include the above details.
+
+We believe that Karmada would benefit from having a generic way to store the job metadata/information required for failover and a label/annotation to indicate that failover has happened so that this can be extended to any stateful application.
+
+To summarize, we currently resume from the last state with some custom changes and third party services which:
+ 1. Append a flag to the ResourceBinding of the CRD to indicate a failover. When the application is rescheduled, a label is appended to the CRD to mark it as failed-over.
+ 2. A custom Kyverno policy then reads the failover label and grabs the latest checkpoint for the application based on its jobID. 
+ 3. The application can then resume from the fetched checkpoint and recommence processing.  
+
+## Design Details
+
+### Support Stateful Failover Options
+
+#### ResourceBinding API Change
+
+We can extend the ResourceBindingStatus with a new field "FailoverHistory", which would be added by the cluster + application failover controller to keep track when the replica has been failed-over. 
+
+```
+// ResourceBindingStatus represents the overall status of the strategy as well as the referenced resources.
+type ResourceBindingStatus struct {
+	// SchedulerObservedGeneration is the generation(.metadata.generation) observed by the scheduler.
+	// If SchedulerObservedGeneration is less than the generation in metadata means the scheduler hasn't confirmed
+	// the scheduling result or hasn't done the schedule yet.
+	// +optional
+	SchedulerObservedGeneration int64 `json:"schedulerObservedGeneration,omitempty"`
+
+	// SchedulerObservedAffinityName is the name of affinity term that is
+	// the basis of current scheduling.
+	// +optional
+	SchedulerObservedAffinityName string `json:"schedulerObservingAffinityName,omitempty"`
+
+	// Conditions contain the different condition statuses.
+	// +optional
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+
+	// AggregatedStatus represents status list of the resource running in each member cluster.
+	// +optional
+	AggregatedStatus []AggregatedStatusItem `json:"aggregatedStatus,omitempty"`
+
+    // FailoverHistory represents history of the previous failovers of this resource
+	FailoverHistory []FailoverHistoryItem `json:"failoverHistory,omitempty"`
+
+}
+```
+
+The FailoverHistory is a list of FailoverHistoryItem objects and is updated every time a failover happens until a certain limit. The limit is set by another field persistedFields.maxHistory which is defined in the propogation policy.
+
+```
+type FailoverHistoryItem struct {
+
+    // FailoverTime represents the timestamp when the workload failed over
+    // It is represented in RFC3339 form(like '2021-04-25T10:02:10Z') and is in UTC
+	FailoverTime *metav1.Time `json:"failoverTime,omitempty"`
+
+    // OriginCluster represents the cluster name from which the workload failed over from
+	OriginCluster string `json:"originCluster,omitempty"`
+
+    // DestinationCluster represents the cluster name from which the workload failed over to
+	DestinationCluster string `json:"destinationCluster,omitempty"`
+
+    // PersistedDuringFailover contains the fields required by the stateful application to resume from that state after failover
+	PersistedDuringFailover []PersistedFailoverItem `json:"persistedFailoverItem,omitempty"` 
+}
+```
+
+The FailoverHistoryItem object contains information relevant to a failover and an additional object called "PersistedDuringFailover" which keeps track of the metadata (both the fields that need to be persisted and how to access those fields) that is required by the stateful operation to resume processing from that state.
+
+```
+type PersistedDuringFailover struct {
+	
+    // LabelName represents the name of the line that will be persisted for the replica
+    // in case there is a failover to a new cluster.
+	LabelName string `json:"labelName,omitempty"`
+
+    // PersistedItem is a pointer to the status item that should be persisted to the rescheduled 
+    // replica during a failover. This should be input in the form: obj.status.<path-to-item> 
+	PersistedStatusItem string `json:"persistedStatusItem,omitempty"`
+}
+```
+
+The PersistedDuringFailover object keeps track of the value of a field during failover so that this information can be used to resume processing from this point onward. This object consists of two fields, LabelName and PersistedStatusItem which are both defined in the propogation policy.
+
+#### PropagationPolicy API Change
+
+We propose to add two fields to a propogation policy to enable stateful failover.
+1. persistedFields.maxHistory: This sets the max limit on the amount of stateful failover history that is persisted before the older entries are overwritten. If this is set to 5, the resourcebinding will store a maximum of 5 failover entries in FailoverHistory before it overwrites the older history.
+2. persistedFields.fields: This is a list of the fields that are required by the stateful application to be persisted during failover to resume processing from a particular case. This takes in a list of field names as well as how to access them from the spec. 
+
+Example propagation policy for Flink jobs that uses the persistedFields.maxHistory and persistedFields.fields:
+
+```
+apiVersion: policy.karmada.io/v1alpha1
+kind: PropagationPolicy
+metadata:
+  name: flinkdep-policy
+spec:
+  failover:
+    application:
+      decisionConditions:
+        tolerationSeconds: 90
+      purgeMode: Graciously
+      gracePeriodSeconds: 10
+    persistedFields.maxHistory: 5
+    persistedFields.fields:
+      - LabelName: jobID
+        PersistedStatusItem: obj.status.jobStatus.jobID
+  resourceSelectors:
+    - apiVersion: flink.apache.org/v1beta1
+      kind: FlinkDeployment
+      namespace: <>
+  propagateDeps: true
+  placement:
+    replicaScheduling:
+      replicaDivisionPreference: Aggregated
+      replicaSchedulingType: Divided
+    spreadConstraints:
+      - spreadByField: cluster
+        maxGroups: 1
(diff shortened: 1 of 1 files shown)


RainbowMango (commented):
/assign 
Thanks @Dyex719 Just quickly went through the propoasl, I believe this is meaningful enhancement and very inspiring.

Dyex719:
Thanks @RainbowMango! One assumption that we wanted to discuss about (I wrote this in the proposal too):
```
All the replicas of the stateful application are not migrated together, it is not clear when the state needs to be restored. In this proposal we focus on the use case where all the replicas of a stateful application are migrated together.
```

Do you think this is a valid assumption? This narrows our scope a little, but defines the problem more clearly.

RainbowMango:
> Do you think this is a valid assumption? This narrows our scope a little, but defines the problem more clearly.

I 100% agree. 
1. Partially replica migration is not technically failover, but more like elastic scaling of replicas. 
2. The conditions under which failover triggers are fully configurable, if people tolerate the failure of part of replicas, then migration should not be triggered, if they don't, the application should be rebuilt, by leveraging failover.

docs/proposals/stateful-failover/stateful-failover-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+}
+```
+
+The PersistedDuringFailover object keeps track of the value of a field during failover so that this information can be used to resume processing from this point onward. This object consists of two fields, LabelName and PersistedStatusItem which are both defined in the propogation policy.
+
+#### PropagationPolicy API Change
+
+We propose to add two fields to a propogation policy to enable stateful failover.
+1. persistedFields.maxHistory: This sets the max limit on the amount of stateful failover history that is persisted before the older entries are overwritten. If this is set to 5, the resourcebinding will store a maximum of 5 failover entries in FailoverHistory before it overwrites the older history.
+2. persistedFields.fields: This is a list of the fields that are required by the stateful application to be persisted during failover to resume processing from a particular case. This takes in a list of field names as well as how to access them from the spec.
```
RainbowMango, line 164: Is there any restriction for the field type here? Can we or should we support types other than `String`?

docs/proposals/stateful-failover/stateful-failover-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+```
+type PersistedDuringFailover struct {
+	
+    // LabelName represents the name of the line that will be persisted for the replica
+    // in case there is a failover to a new cluster.
+	LabelName string `json:"labelName,omitempty"`
+
+    // PersistedItem is a pointer to the status item that should be persisted to the rescheduled 
+    // replica during a failover. This should be input in the form: obj.status.<path-to-item> 
+	PersistedStatusItem string `json:"persistedStatusItem,omitempty"`
```
RainbowMango, line 154: Given this `persistedStatusItem` will be applied as a label value, which brings a limitation that the length can not exceed 63 characters. Does Flink's checkpoint path exceed this?

docs/proposals/stateful-failover/stateful-failover-proposal.md
Dyex719: persistedFields.fields contains the name of the field we want to persist and the path to obtain the value of that label. Since both the name and value is tied to a label which supports only strings, I think having a string is sufficient.

```
persistedFields.fields:
      - LabelName: jobID
         PersistedStatusItem: obj.status.jobStatus.jobID
```

docs/proposals/stateful-failover/stateful-failover-proposal.md
Dyex719: I should make a change here:

> // persistedStatusItem is ~~a pointer to the status item~~ the value of the status item that should be persisted to the rescheduled replica during a failover. This should be input in the form: ~~obj.status.path-to-item~~ value at the location obj.status.path-to-item

In the case of flink, the flink process will create a jobID at obj.status.jobStatus.jobID which would look like `fd72014d4c864993a2e5a9287b4a9c5d` so the persistedStatusItem would store this value in the PersistedStatusItem field.

Flink Deployment spec would have:
```
status:
  jobStatus:
    jobID: fd72014d4c864993a2e5a9287b4a9c5d
```

PersistedDuringFailover would store jobID as the labelName and fd72014d4c864993a2e5a9287b4a9c5d as the PersistedStatusItem (value)

docs/proposals/stateful-failover/stateful-failover-proposal.md
Dyex719: Since we are storing the metadata related to the state here, I think this limitation is okay. The actual state of the flink job will be stored in some durable storage and may be very big. In other stateful applications as well I think we can make this assumption that metadata related to the state should be persisted with the resourceBindingStatus and the actual data can be stored elsewhere.

docs/proposals/stateful-failover/stateful-failover-proposal.md
Dyex719: Since we are storing the metadata related to the state here, I think the limitation of the value being a string and less than 63 characters is okay. The actual state of the flink job will be stored in some durable storage and may be very big. In other stateful applications as well I think we can make this assumption that metadata related to the state should be persisted with the resourceBindingStatus and the actual data can be stored elsewhere.

I will include this in the proposal. Let me know what you think

docs/proposals/stateful-failover/stateful-failover-proposal.md
RainbowMango: It works I believe. But what I'm thinking here is maybe we can make room for extensibility. The info we can take can be more generic, like a string, integer, and so on. I'm still thinking about it. will get back to you once I have an idea.

RainbowMango: Tried to draft the API, please take a look:
```golang
// ApplicationFailoverBehavior indicates application failover behaviors.
type ApplicationFailoverBehavior struct {
	// other fields.

	// StatePreservation defines the policy for preserving and restoring state data
	// during failover events for stateful applications.
	//
	// When an application fails over from one cluster to another, this policy enables
	// the extraction of critical data from the original resource configuration.
	// Upon successful migration, the extracted data is then re-injected into the new
	// resource, ensuring that the application can resume operation with its previous
	// state intact.
	// This is particularly useful for stateful applications where maintaining data
	// consistency across failover events is crucial.
	// If not specified, means no state data will be preserved.
	// +optional
	StatePreservation *StatePreservation `json:"statePreservation,omitempty"`
}

// StatePreservation defines the policy for preserving state during failover events.
type StatePreservation struct {
	// Rules contains a list of StatePreservationRule configurations.
	// Each rule specifies a JSONPath expression targeting specific pieces of
	// state data to be preserved during failover events. An AliasName is associated
	// with each rule, serving as a label key when the preserved data is passed
	// to the new cluster.
	// +required
	Rules []StatePreservationRule `json:"rules"`

	// Note: We probably need more policies to control how to feed the new cluster with the
	// preserved state data in the future. Such as:
	// - Is it always acceptable to feed data as the label? Is there a need for annotation?
	// - If each label name should be started with a prefix, like `karmada.io/failover-preserving-<fieldname>: <state>`
	// Sure, we can default with the label, this structure just makes room for future extensions.
	//
	// For instance, we can introduce a policy if someone wants to control how the preserving state
	// feed to new clusters. This is probably not included this time.
	// RestorePolicy determines when and how the preserved state should be restored.
	// RestorePolicy RestorePolicy `json:"restorePolicy"`
}

// StatePreservationRule defines a single rule for state preservation.
// It includes a JSONPath expression and an alias name that will be used
// as a label key when passing state information to the new cluster.
type StatePreservationRule struct {
	// AliasName is the name that will be used as a label key when the preserved
	// data is passed to the new cluster. This facilitates the injection of the
	// preserved state back into the application resources during recovery.
	// +required
	AliasName string `json:"aliasName"`

	// JSONPath is the JSONPath query used to identify the state data
	// to be preserved from the original resource configuration.
	// Example: ".spec.template.spec.containers[?(@.name=='my-container')].image"
	// This example selects the image of a container named 'my-container'.
	// +required
	JSONPath string `json:"jsonPath"`
}
```

RainbowMango:
Hi @Dyex719, @mszacillo, I've been thinking this feature recently and came up with some ideas, this feature consists of 3 parts:

The first part is how to declare which state data(fields) should be preserved during failover, and I post my idea at https://github.com/karmada-io/karmada/pull/5116#discussion_r1685365859. Please take a look.

The second part is how to store the preserved state data, one approach is to store them in the history items(this is our first idea), the challenge things is that it's hard to maintain the history item, especially figuring out what is the destination cluster. 

I think it is worth thinking about storing them in the [GracefulEvictionTask](https://github.com/karmada-io/karmada/blob/f2b37a43a3e7d85277b6481eea876b13b4cb8a88/pkg/apis/work/v1alpha2/binding_types.go#L225-L277), during the failover process, just before creating the eviction task, it makes sense to grab some state(fields) or snapshot of scheduled cluster list. With this grabbed data, the controller would get to know which cluster is the destination by comparing the snapshot and the newly scheduled cluster. 

The third part is how to feed(or inject) the preserved state data to the destination cluster. This is going to be not that complex as long as the controller can figure out the destination cluster and only feed the state data when creating the application.

Dyex719:
We aim to implement this feature in two parts
1. Failover Flag implementation
2. Changes to Propogation policy

I have been working on Part 1 for now. To address your comment:

> The second part is how to store the preserved state data, one approach is to store them in the history items(this is our first idea), the challenge things is that it's hard to maintain the history item, especially figuring out what is the destination cluster.

One thing we were thinking about it is to only store the cluster from which the workload failed over from. This would help keep the logic simple without involving multiple components. The current cluster the workload is scheduled on can always be inferred from the resourcebinding.

With this we would achieve: 
```
spec:
  clusters:
  - name: member3
    replicas: 2
...
failoverHistory:
    - failoverTime: "2024-07-18T19:03:06Z"
      originCluster: member1
      reason: "applicationFailover"
    - failoverTime: "2024-07-18T19:08:45Z"
      originCluster: member2
      reason: "clusterFailover"
```

The entire trace of the workload could still be inferred from the current state + failoverHistory, the workload was originally on member1, then it migrated to member2, then it moved to member3 which can be inferred from spec.clusters in the resource binding where it is currently scheduled.

Work in progress implementation is available here: https://github.com/karmada-io/karmada/compare/master...Dyex719:karmada:stateful-failover-flag.

Dyex719:
> The third part is how to feed(or inject) the preserved state data to the destination cluster. This is going to be not that complex as long as the controller can figure out the destination cluster and only feed the state data when creating the application.

Our idea was to preserve only the metadata of the state rather than the actual state which may be large and be of different formats in the resourcebinding.

As you know, we use Kyverno for fetching the actual state using the persisted metadata in the resourcebinding. My understanding is that Kyverno is well supported by Karmada so we could use Kyverno to do this injection. It also allows a lot of customization that the user can perform to suit their needs.

Is your idea to have another component (maybe the scheduler) do this injection? My only concern is that customization may become difficult.

docs/proposals/stateful-failover/stateful-failover-proposal.md
RainbowMango: @Dyex719 @mszacillo What do you think of this API? 

I haven't put history thing here, but seems it should not in StatePreservation or `persistedFields` as referenced by #5116.

Dyex719: Forgot to reply to this! I think this looks good!
We can probably move this API to a separate file so that it can be shared by both the cluster and application failover controllers.

docs/proposals/stateful-failover/stateful-failover-proposal.md
Dyex719: Looks like as you said annotations values can support 253 characters whereas label values can only be 64 characters long. Annotation keys can only be 64 characters whereas label keys can be 253 characters.

docs/proposals/stateful-failover/stateful-failover-proposal.md
RainbowMango: Good point, let me think about it.

RainbowMango: It makes 100% sense to move the `StatePreservation` to `FailoverBehavior`, so that both Cluster failover and application failover can share the same configurations.

But, given the cluster failover configuration is still not provided, also not sure when it will come, I'm afraid it might slow down the process if we consider cluster failover in this proposal. So, I tend to focus on application failover right now.

In addition, if cluster failover also needs this configuration later, we can deprecate the one in the application in a compatible way, it's not a concern for now.

RainbowMango: By the way, now I'm going to look at the second part(as I mentioned on https://github.com/karmada-io/karmada/pull/5116#issuecomment-2241085462), which is how to grab and store the state data in ResourceBining.

RainbowMango: > Are there plans to include a separate cluster failover implementation? We would prefer to have cluster failover be in scope of this proposal as it is important for flink applications to recover from these failures as well.

I don't mind having the cluster failover in this proposal.  You might need to refresh the proposal. 
I'm sure this gonna be a large proposal.

RainbowMango:
@mszacillo @Dyex719 
I tried to design the API and asked @XiShanYongYe-Chang for a demo, the demo shows it is feasible. I hope we can demonstrate the demo at the next meeting. And I also hope you can help to take a look it.

API Design: https://github.com/karmada-io/karmada/compare/master...RainbowMango:karmada:api_draft_application_failover
Demo: https://github.com/karmada-io/karmada/compare/master...XiShanYongYe-Chang:karmada:api_draft_application_failover

RainbowMango:
@XiShanYongYe-Chang I see @mszacillo and @Dyex719 added an agenda to [this week's community meeting](https://docs.google.com/document/d/1y6YLVC-v7cmVAdbjedoyR5WL0-q45DBRXTvz5_I7bkA/edit?pli=1&tab=t.0#heading=h.7wgntwllhbdy), I wonder if you can show a demo during the meeting, or share the test report here.

XiShanYongYe-Chang:
> @XiShanYongYe-Chang I see @mszacillo and @Dyex719 added an agenda to [this week's community meeting](https://docs.google.com/document/d/1y6YLVC-v7cmVAdbjedoyR5WL0-q45DBRXTvz5_I7bkA/edit?pli=1&tab=t.0#heading=h.7wgntwllhbdy), I wonder if you can show a demo during the meeting, or share the test report here.

Okay, let me show you the demo during the meeting tonight.

XiShanYongYe-Chang:
Let me describe the demo in words.

Create each resource according to the following configuration file:

<details>

<summary>unfold me to see the yaml</summary>

```yaml
# deploy.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx
  labels:
    app: nginx
spec:
  replicas: 2
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - image: nginx
        name: nginx
===================================
# pp.yaml 
# propagate the target Deployment to member1, member2 member clusters, only to one cluster by setting spreadConstraints
kind: PropagationPolicy
metadata:
  name: nginx-propagation
spec:
  resourceSelectors:
    - apiVersion: apps/v1
      kind: Deployment
      name: nginx
  placement:
    clusterAffinity:
      clusterNames:
        - member1
        - member2
    spreadConstraints:
      - spreadByField: cluster
        maxGroups: 1
        minGroups: 1
  propagateDeps: true
  failover:
    application:
      decisionConditions:
        tolerationSeconds: 60
      purgeMode: Graciously
      statePreservation:
        rules:
          - aliasLabelName: cztest-replicas
            jsonPath: ".updatedReplicas"
===================================
# op.yaml 
# Modify the Deployment image distributed to the member1 cluster via OverridePolicy to simulate an application failure on member1
kind: OverridePolicy
metadata:
  name: nginx-op
spec:
  resourceSelectors:
    - apiVersion: apps/v1
      kind: Deployment
      name: nginx
  overrideRules:
    - targetCluster:
        clusterNames:
          - member1
      overriders:
        imageOverrider:
          - component: "Registry"
            operator: replace
            value: "fake"
```
</details>

Execute the following commands to deploy the above resources to the karmada control plane:

```
 kubectl --context karmada-apiserver apply -f op.yaml
 kubectl --context karmada-apiserver apply -f pp.yaml
 kubectl --context karmada-apiserver apply -f deployment.yaml
```

The deployment will be propagated to the member1 cluster first, but due to a image error, after waiting for 1 minute, it will be propagated to the member2 cluster after application failover.

Then execute the following command to watch the deployment resource on the member2 cluster:

```
 kubectl --kubeconfig /root/.kube/members.config --context member2 get deployments.apps -oyaml -w
```

You will observe that the target `cztest-replicas: "2"` appears in the label of the deployment nginx resource on the member2 cluster.

image

mszacillo:
Hi Hongcai and Chang,

Thanks for the heads up! I'm a little surprised that there has been a
different branch created for the change - as this is something we had
proposed and put up for review. If there were comments on implementation
changes, I had hoped those would be brought up during the review process
(such as changes that should be done to the apis). We currently use the
failover feature in our own DEV setup, and although I appreciate all the
work you've done I was hoping we could use most of what has already been
completed on our end. Perhaps we can discuss more during the meeting?

Cheers.

RainbowMango:
> I'm a little surprised that there has been a
different branch created for the change - as this is something we had
proposed and put up for review (for reference we currently use the failover
feature in our own DEV setup), but thank you for putting so much thought into this feature.

Thanks for letting me know! 
We discussed this proposal a lot. I want to help move forward with the feature, so I tried to refine the API design based on your idea. The new branch was just used to explain my thoughts. I'm looking forward to hearing your thoughts, and if we reach a consensus, they should be put in this proposal.

RainbowMango:
Hi @Dyex719, @mszacillo I'm reviewing this feature to identify any remaining tasks that need to be completed, and then determine where to start from.
I wonder if you still want to refresh this proposal according to what we have done in the last release.

My two cents: 
1. This proposal should focus on application failover instead of `cluster failover` which we can start another feature/proposal.
2. For the failover history item part, I think it can also included in the cluster failover feature.

By the way, I and @XiShanYongYe-Chang currently working on the cluster failover feature enhancement, will let you updated once we work out some ideas.

mszacillo:
Hi @RainbowMango,

I'll go ahead and make edits to the proposal to match up with the API changes that have been made as part of #5788. I'm happy to keep this proposal application failover specific - in the meantime we'll maintain a [small internal commit ](https://github.com/karmada-io/karmada/compare/master...mszacillo:karmada:cluster-failover-with-state) that allows us to re-use the graceful eviction task logic for cluster failover. The commit is primarily changes to the related test files, the actual code change is gated behind the failover feature flag and is ~20 lines. 

And sounds good! Please keep me updated about cluster failover design, as we are planning on using the feature.

RainbowMango:
The application failover should be part of cluster failover, we will revisit the entire feature in the near future, at which time we will re-describe the feature in the design doc or website documentation. So, I think we can close this PR for now, however if it can be updated to align with the current implementation, it can be reopened and merged at any time.

/close

Issue #4937, opened 2024-05-14
[Feature] Support suspend ResourceBinding when create


Vacant2333:
**What would you like to be added**:
I hope to add the ability to suspend `ResourceBinding`. If set to `True`, `karmada-scheduler` will ignore this `ResourceBinding` until `suspend=false` before starting to schedule it.

**Why is this needed**:
- As a cluster admin, I hope `Karmada` can have a queueing capability similar to the combination of `Kueue` and `kube-scheduler`. This would enable distributing workloads/jobs to Worker Clusters based on priority when there are a large number of loads/jobs.
- As a cluster admin, I would like the option to temporarily withhold the scheduling of certain `PropagationPolicy` instances that I deploy, until I set `PropagationPolicy.spec.suspend` to `false` (and propagate it to `ResourceBinding`) in specific circumstances.

**Question And Answers**:
Q: How does `[Kueue](https://kubernetes.io/blog/2022/10/04/introducing-kueue/)` provide queue capabilities to `kube-scheduler`?
A: Taking `Job` as an example, `Job` contains a `suspend` field. If this field is set to `True`, `kube-scheduler` will ignore the `Job` until it becomes False. `Kueue` on the other hand, provides an additional controller. It includes queue capabilities along with sorting algorithms. `Kueue` considers multiple conditions (such as task priority, resource availability) before updating `suspend` to false and handing it over to `kube-scheduler` for normal scheduling thereafter.

Q: Does `PropagationPolicy` also need to have a `suspend` field?
A: Yes, as I understand it, `PropagationPolicy` is user-facing resource. Users only need to set `suspend` here, and it will automatically propagate to associated resources (`ResourceBinding`, `Work`).

Q: Whats the difference with issue `Ability to suspend work` #4688 ?
A: The focus of this issue is not to react to new changes. My focus is on pausing the distribution of the entire `PropagationPolicy` resource. Another point is that the `Work` resource is the result of `Karmada` scheduling. We need to ensure that the desired queue capability guarantees that resources have not yet reached the scheduling stage (considering resource quota usage).

**Additional Considerations**:
[...]

Issue #4421, opened 2023-12-14
support  canary release


CharlesQQ:
**What would you like to be added**:
1. karmada allow pause work object propagate to member cluster until  user want to 
2. during propagation is paused,  user changed the overriepolicy By API, such as  `annotationsOverrider` field
3. user set and allow new work object be to propagated to member cluster

**Why is this needed**:
Our canary release strategy is declare by  deployment.metedata.annotation,  process is  controller by our owner custom controller;
So before the latest deployment is synced to member cluster, deployment.metedata.annotation must be set

chaosi-zju:
Hi @CharlesQQ, I am very interested in your feature

Actually, rather than just implement this feature, I am more interested in how you use canary release strategy by in karmada.

Can you share more info with me on how using canary release strategy in karmada? Thanks very much~

chaunceyjiang:
I believe it's time to push forward with this feature. This is a feature that has been discussed for a long time. 
@RainbowMango @XiShanYongYe-Chang @CharlesQQ

chaosi-zju:
> I believe it's time to push forward with this feature. This is a feature that has been discussed for a long time.

is there any related issue or meeting notes?

XiShanYongYe-Chang:
> I believe it's time to push forward with this feature. This is a feature that has been discussed for a long time.

I agree with this.

CharlesQQ:
> Hi @CharlesQQ, I am very interested in your feature
> 
> Actually, rather than just implement this feature, I am more interested in how you use canary release strategy by in karmada.
> 
> Can you share more info with me on how using canary release strategy in karmada? Thanks very much~

This has nothing to do with karmada, it's related to deployment Annotation,  as Annotation‘s value changed, customer controller calculate newRS's replicas and oldRS's Replicas;  before deployment synced to member cluster, Annotation's value must be changed by our owner controller, so needed paused sync work feature

<img width="340" alt="image" src="https://github.com/karmada-io/karmada/assets/18431718/4d5700a1-0173-4b93-8370-d99c211e3179">

zach593:
After reading the proposal https://github.com/karmada-io/karmada/pull/5118/files, I have some questions.
Of course, the proposal focuses on cluster-level pauses, so how to implement application-level pauses mentioned in this issue? To be determined?

Using pp/cpp for application-level pauses is OK, but not good enough.
[...]

Issue #4919, opened 2024-05-08
[Feature] Karmada-scheduler support custom-plugin when ReplicaScheduling


Vacant2333:
**What would you like to be added**:
Currently, the `karmada-scheduler` only supports adding custom plugins in the first two scheduling stages (`FilterClusters`, `ScoreCluster`). I hope that custom plugins can be supported in the last stage (`ReplicaScheduling`) to achieve some functionalities.

You can check this document about the 4 scheduling stages

https://karmada.io/docs/next/developers/customize-karmada-scheduler/

**Why is this needed**:
Currently, I am working hard to enable [`Volcano`](https://volcano.sh/en/) to support [multi-cluster AI workload scheduling](https://mentorship.lfx.linuxfoundation.org/project/132a4971-6969-4ca6-a695-783ece3ac768). `Volcano` currently only has the capability for single-cluster operations. We believe that `Karmada` is very suitable for combining with `Volcano` to achieve multi-cluster workload scheduling capabilities. 

In `Volcano`, the most basic and important scheduling strategy is `gang` scheduling, which essentially means that the cluster's resources must be sufficient to launch all the pods required by the target job. In Kubernetes' default scheduler, if a job requires 100 pods, each with 100GB of memory, but the cluster only has 50GB available, the job will still start. However, due to the insufficient number of pods started, the job will occupy resources while blocking resources for the entire cluster.

If custom-plugin support is enabled in the `ReplicaScheduling` stage, then we can determine during this stage whether the job can be fully scheduled to a single target cluster. If it cannot, we can halt the scheduling process. Currently, our scheduling goal is that once a user creates a `vcjob` and `PropagationPolicy`, the `vcjob` can be scheduled to a single cluster (splitting the job across multiple clusters is more complex and will be considered later). This ensures that the `vcjob` is not divided and that the target cluster can fully start the `vcjob`.

<img width="784" alt="image" src="https://github.com/karmada-io/karmada/assets/19872346/74f014aa-179f-4f43-966d-6a878777bc0d">

Karmada-scheduler has four scheduling phases, but currently only the first two phases (FilterCluster, ScoreCluster) support custom plugins. I believe it is a reasonable requirement for all four phases to support custom plugins, which would enhance the extensibility of the scheduler, enabling support for a wider range of scenarios. The current two extension points are insufficient.

Reference Material:
[...]

PR #5118, opened 2024-07-02
[Proposal] Support for cluster-level resource propagation pause and resume capabilities

### LINKED ISSUE #1567: How to Support dynamically adding overriders
We have a custom controller for Annotation-based gray code release; The different code release batch for cluster has the different metedata.annotation value, the value  needs to be dynamically added in the overridepolicy and it's changed for each code release batch;

**Why is this needed**:
Support to dynamically add and complete this part of overriders before the work object is generated. I think of two ways to achieve this:
1. Add a new type of interpreter webhook
2. Allow user-defined controllers to watch overridepolicy's create/update events, and dynamically patch overriders, but it needs to be done before the work object is generated

XiShanYongYe-Chang:
**What type of PR is this?**

/kind design
/kind documentation

Fixes #1567, #4421, #4688

Other related issues: #4937 

```release-note
NONE
```

### DIFF
--- a/docs/proposals/dispatch-suspension/README.md
+++ b/docs/proposals/dispatch-suspension/README.md
@@ -1,0 +1,444 @@
+---
+title: Support for cluster-level resource propagation pause and resume capabilities
+
+authors:
+- "@XiShanYongYe-Chang"
+
+reviewers:
+- "@a7i"
+- "@CharlesQQ"
+- "@chaunceyjiang"
+- "@Monokaix"
+- "@RainbowMango"
+- "@Vacant2333"
+- "@whitewindmills"
+
+approvers:
+- "@kevin-wangzefeng"
+- "@RainbowMango"
+
+creation-date: 2024-07-01
+
+---
+
+# Support for cluster-level resource propagation pause and resume capabilities
+
+## Summary
+
+The process of resource propagation in Karmada has gone through stages from the Resource Template to `ResourceBinding/ClusterResourceBinding`, to `Work`, and then to synchronization with member clusters. These three stages can be said to be seamless at present. The only point where users can intervene in the middle process is when `ResourceBinding/ClusterResourceBinding` generates `Work`, which applies the `OverridePolicy/ClusterOverridePolicy` resource. However, this point also requires users to have created the `OverridePolicy/ClusterOverridePolicy` resource in advance, and there is no way to handle the situation where users need to customize the `OverridePolicy/ClusterOverridePolicy` based on the scheduling results. Therefore, if the resource propagation process can be paused, it will provide users with greater control and flexibility.
+
+This article provides a capability that supports the pause and resume of resource propagation at the cluster granularity.
+
+## Motivation
+
+The current Karmada system is quite sensitive to changes in resources. When there are changes in the Resource Template, `PropagationPolicy/ClusterPropagationPolicy`, or `OverridePolicy/ClusterOverridePolicy`, they will be immediately synchronized to the member clusters. In most cases, this is the expected behavior, but there are still some situations where users do not want such changes to be immediately synchronized to the member clusters. To go further, users would like to control the timing of the synchronization of the above resource changes themselves, bringing more possibilities to the development and operation and maintenance of the federated system.
+
+### Goals
+
+- Provide the capability to pause the propagation of resources;
+- Provide the capability to resume the propagation of resources;
+- Provide orchestration capabilities for pausing and resuming the propagation of resources at the cluster granularity.
+
+### Non-Goals
+
+- Provide out-of-the-box capabilities suitable for multi-cluster scenarios such as canary releases, rolling releases, and blue-green deployments;
+- Provide automated workload releasing according to a specified cluster sequence.
+
+## Proposal
+
+### User Stories (Optional)
+
+#### Story 1
+
+As an operator, when the Karmada control plane competes with member clusters for control of resources, there is a situation where resources are repeatedly updated. I hope to pause the process of synchronizing the resource to the member clusters, which would help me quickly locate the problem.
+
+#### Story 2
+
+As a manager, I wish to specify the release order of clusters during the application version rollout. I want to proceed with the release to the next cluster only after the release has successfully been deployed to the current one.
+
+#### Story 3
+
+During the business deployment, by setting the partition field in some workload-type resources, such as StatefulSet, you can specify the update of a subset of instances, achieving a certain amount of canary deployment.
+
+As an operator, when I specify to update a portion of instances for a Deployment resource in the federated control plane, it is necessary to split the number of replicas to be updated when distributing to different clusters. I would like to pause the propagation process of the Deployment resource when calculating the split number for different clusters. After the calculation is complete and the results are applied to the system, I would then resume the propagation process of the Deployment resource to achieve the goal of canary deployment.
+
+Related issue: https://github.com/karmada-io/karmada/issues/1567, https://github.com/karmada-io/karmada/issues/4421.
+
+#### Story 4
+
+As an operator, I hope that when making business changes in the federated control plane, the changes are not directly synchronized to the member clusters, but are instead retained within the control plane. This allows me to check the changes in the control plane's resources, such as the contents of ResourceBinding, Work, and other resources, to determine if the system is executing as expected. Once I am certain that everything is in order, I can then resume the business changes and synchronize the results to the member clusters.
+
+### Notes/Constraints/Caveats (Optional)
+
+### Risks and Mitigations
+
+1. When the workload is in a paused state, even if the user has enabled the Failover feature gate, the fault migration will be paused until the user cancels the pause.
+
+## Design Details
+
+By extending the `PropagationPolicy/ClusterPropagationPolicy` API, we introduce new suspension structures and pass critical suspension information through to the `ResourceBinding/ClusterResourceBinding` resources, which in turn are propagated to the `Work` resources. The controller within the `Karmada-controller-manager` component will then determine whether to synchronize resources to target member clusters based on the suspension information found in the `Work` resources.
+
+### API Change
+
+#### PropagationPolicy/ClusterPropagationPolicy
+
+```go
+// PropagationSpec represents the desired behavior of PropagationPolicy.
+type PropagationSpec struct {
+	...
+
+	// Suspension declares the policy for suspending different aspects of propagation.
+	// nil means no suspension. no default values.
+	// +optional
+	Suspension *Suspension `json:"suspension,omitempty"`
+}
+
+// Suspension defines the policy for suspending different aspects of propagation.
+type Suspension struct {
+	// SuspendScheduling controls whether scheduling should be suspended.
+	// +optional
+	// Note: Postpone this until there is a solid use case.
+	// SuspendScheduling *bool `json:"suspendScheduling,omitempty"`
+
+	// SuspendDispatching controls whether dispatching should be suspended.
+	// nil means not suspend.
+	// Note: true means stop propagating to all clusters. Can not co-exist
+	// with DispatchingOnClusters which is used to suspend particular clusters.
+	// +optional
+	SuspendDispatching *bool `json:"suspendDispatching,omitempty"`
+
+	// SuspendDispatchingOnClusters declares a list of clusters to which
+	// the dispatching should be suspended.
+	// Note: Can not co-exist with Dispatching which is used to suspend all.
+	// +optional
+	SuspendDispatchingOnClusters *SuspendClusters `json:"suspendDispatchingOnClusters,omitempty"`
+}
+
+// SuspendClusters represents a group of clusters that should be suspended from propagating.
+// Note: No plan to introduce the label selector or field selector to select clusters yet, as it
+// would make the system unpredictable.
+type SuspendClusters struct {
+	// ClusterNames is the list of clusters to be selected.
+	// +optional
+	ClusterNames []string `json:"clusterNames,omitempty"`
+}
+```
+
+#### ResourceBinding/ClusterResourceBinding
+
+```go
+// ResourceBindingSpec represents the expectation of ResourceBinding.
+type ResourceBindingSpec struct {
+    ...
+
+	// Suspension declares the policy for suspending different aspects of propagation.
+	// nil means no suspension. no default values.
+	// +optional
+	Suspension *Suspension `json:"suspension,omitempty"`
+}
+```
+
+#### Work
+
+```go
+// WorkSpec defines the desired state of Work.
+type WorkSpec struct {
+	...
+
+	// SuspendDispatching controls whether dispatching should
+	// be suspended, nil means not suspend.
+	// Note: true means stop propagating to all clusters.
+	// +optional
+	SuspendDispatching *bool `json:"suspendDispatching,omitempty"`
+}
+```
+
+A new ConditionType called `SuspendDispatching` is introduced to describe the distribution suspension status of the Work resource. Meanwhile, the system will also report events to record the suspension of resource distribution.
+
+### User usage example
+
+Users set the Deployment(default/nginx) resource distribution to be paused in the `PropagationPolicy`. When users modify this deployment, Karmada will not synchronize the changes to all target clusters.
+
+```yaml
+apiVersion: policy.karmada.io/v1alpha1
+kind: PropagationPolicy
+metadata:
+  name: nginx-propagation
+spec:
+  resourceSelectors:
+    - apiVersion: apps/v1
+      kind: Deployment
+      name: nginx
+  placement:
+    clusterAffinity:
+      clusterNames:
+        - member1
+        - member2
+        - member3
+  suspension:
+    suspendDispatching: true
+```
+
+At this point, all `Work` resources related to this `Deployment` on the Karmada control plane will be in a paused state, taking the `Work` resource corresponding to the member1 cluster as an example:
+
+```yaml
+apiVersion: work.karmada.io/v1alpha1
+kind: Work
+metadata:
+  name: nginx-xxx
+  namespace: karmada-es-member1
+spec:
+  workload:
+    manifests:
+      - xxx
+  suspendDispatching: true
+status:
+  conditions:
+    - lastTransitionTime: "2024-07-01T08:33:28Z"
+      message: resource distribution is in a pause state.
+      reason: SuspendDispatching
+      status: "True"
+      type: Dispatching
+```
+
+When users cancel the pause in the PropagationPolicy, it will resume the resource synchronization operation of Deployment(default/nginx) in all clusters.
+
+For instance, if the Deployment(default/nginx) has been propagated to the clusters member1, member2, and member3, and when a user configures according to the following file and updates the image of the Deployment(default/nginx) resource to release a new version, the new version of the resource will only be synchronized to the member1 cluster.
+
+```yaml
+apiVersion: policy.karmada.io/v1alpha1
+kind: PropagationPolicy
+metadata:
+  name: nginx-propagation
+spec:
+  resourceSelectors:
+    - apiVersion: apps/v1
+      kind: Deployment
+      name: nginx
+  placement:
+    clusterAffinity:
+      clusterNames:
+        - member1
+        - member2
+        - member3
+  suspension:
+    suspendDispatchingOnClusters:
+      clusterNames:
+        - member2
+        - member3
+```
+
+When the user confirms that the deployment resource in member1 has been successfully published, and modifies the synchronization configuration as follows, the new version of the resource can then be synchronized to the member2 cluster:
+
+```yaml
+apiVersion: policy.karmada.io/v1alpha1
+kind: PropagationPolicy
+metadata:
+  name: nginx-propagation
+spec:
+  resourceSelectors:
+    - apiVersion: apps/v1
+      kind: Deployment
+      name: nginx
+  placement:
+    clusterAffinity:
+      clusterNames:
+        - member1
+        - member2
+        - member3
+  suspension:
+    suspendDispatchingOnClusters:
+      clusterNames:
+        - member3
+```
+
+If, after synchronizing the new version of the resource to the member1 cluster, issues are found with the application that cause the release to fail, the user can roll back the version of the Deployment resource in the Karmada control plane, thereby protecting the business in the member2 and member3 clusters from being affected.
+
+Q&A: How to remove the annotation `propagation.karmada.io/instruction` from Work? It can be carried by the `Dispatching` field in the Work.
+
+### Test Plan
+
+#### UT
+
+Add unit tests to cover the new functions.
+
+#### E2E
+
+- Test the resource dispatch suspension capability.
+- Test the resource dispatch resume capability.
+- Test the cluster-level resource dispatch suspension capability.
+- Test at the namespace level and the cluster level respectively.
+
+## Alternatives
+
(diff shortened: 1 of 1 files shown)


XiShanYongYe-Chang:
/cc @CharlesQQ @a7i @chaunceyjiang @whitewindmills 
Hi guys, can you help take a look?

whitewindmills:
@XiShanYongYe-Chang 
it's also releated [issue](https://github.com/karmada-io/karmada/issues/4937)

docs/proposals/work-pause/README.md
```diff
@@ -0,0 +1,500 @@
...
+	// Note: The extension design is not a focal point of the current proposal.
+	ScheduleSuspension *bool `json:"scheduleSuspension,omitempty"`
+
+	// DistributionSuspendClusters is used to specify a list of clusters where resource
+	// synchronization needs to be paused. Clusters in this list indicate that resource
+	// synchronization for targeted resources will be suspended to these clusters.
+	// Clusters removed from this list indicate that resource synchronization should be resumed.
+	// If there is a cluster element with an asterisk "*" in the list, it indicates that
+	// the synchronization of resources to all clusters is paused for the matched resources.
+	DistributionSuspendClusters []string `json:"distributionSuspendClusters,omitempty"`
```
chaunceyjiang, line 197: Just one question: Why doesn't it support `labelSelector` like ClusterAffinity?

docs/proposals/work-pause/README.md
```diff
@@ -0,0 +1,500 @@
...
```
chaunceyjiang, line 172: Just one question: Why doesn't it support labelSelector like ClusterAffinity?

docs/proposals/work-pause/README.md
```diff
@@ -0,0 +1,500 @@
...
+spec:
+  resourceSelectors:
+    - apiVersion: apps/v1
+      kind: Deployment
+      name: nginx
+  placement:
+    clusterAffinity:
+      clusterNames:
+        - member1
+        - member2
```
chaunceyjiang, line 287:
```suggestion
        - member2
        - member3
```

docs/proposals/work-pause/README.md
XiShanYongYe-Chang: Update it with a new structure. Please take a look again.

docs/proposals/work-pause/README.md
```diff
@@ -0,0 +1,621 @@
...
+	// set to false indicates the pause is canceled.
+	// Default value is false.
+	// Note: The extension design is not a focal point of the current proposal.
+	ResourceMatchPause *bool `json:"ResourceMatchPause,omitempty"`
+
+	// SchedulePause specifies the pause scheduling of resources matched by the policy,
+	// set to false indicates the pause is canceled.
+	// Default value is false.
+	// Note: The extension design is not a focal point of the current proposal.
+	SchedulePause *bool `json:"schedulePause,omitempty"`
```
XiShanYongYe-Chang, line 166: cc @chaunceyjiang @RainbowMango 
After further consideration, I think that using a boolean type should be sufficient. If an enumeration type is used, it would add a lot of enumeration types, and most of the values would also only have two states: pause and resume. Based on the experience with [Deployment pause](https://github.com/kubernetes/kubernetes/blob/95debfb5b6dcfa2467f2700d81df6767a2c91bac/staging/src/k8s.io/api/apps/v1/types.go#L411-L413), a boolean value is already adequate. If more granular pause effects are added in the future, I think it would be a good choice to expand horizontally in the PauseStrategy.

XiShanYongYe-Chang:
Hi @CharlesQQ @chaunceyjiang @whitewindmills @RainbowMango , it's ready to review again.

docs/proposals/work-pause/README.md
```diff
@@ -0,0 +1,621 @@
...
+}
+
+type DistributionPauseStrategy struct {
+	// PauseAllClusters specifies the overall suspension of resource distribution across all clusters.
+	// If PauseAllClusters field is set, PauseClusters field will be ignored.
+	// Default value is false.
+	PauseAllClusters bool `json:"pauseAllClusters,omitempty"`
+
+	// PauseClusters specifies the cluster list where resource distribution needs to be paused.
+	PauseClusters []string `json:"pauseClusters,omitempty"`
```
a7i, line 179: thoughts on reusing [clusterAffinity concept](https://github.com/karmada-io/karmada/blob/c8acebc295e5ae220760abce91b075a4073a94e8/pkg/apis/policy/v1alpha1/propagation_types.go#L382)?

XiShanYongYe-Chang: Thanks, good suggestion!

Vacant2333 (commented):
Hi!
Thanks for your proposal!

In Solution One, the `SchedulePause` of `ResourceBinding` is propagated through `PropagationPolicy`. If I create a `PropagationPolicy` with a `PauseStrategy` set to `nil` by default, is it feasible to only update the `SchedulePause` of `ResourceBinding` to true via a webhook when the `karmada detector` creates the `ResourceBinding`? If this is feasible, does "propagation" mean that the associated `ResourceBinding/Work` is only updated when the `PropagationPolicy` updates the `PauseStrategy` field?

I also have a suggestion: can this issue be considered as a user story? In brief, it involves pausing the `ResourceBinding` at the time of creation through a webhook, and then using a custom controller to determine when certain `ResourceBindings` can be scheduled.

https://github.com/karmada-io/karmada/issues/4919

docs/proposals/work-pause/README.md
```diff
@@ -0,0 +1,625 @@
...
+```go
+// WorkSpec defines the desired state of Work.
+type WorkSpec struct {
+	...
+
+	// DistributionPause indicates a pause in synchronizing the resource
+	// templates in the current Work to the target cluster.
+	// It sets to false indicates the suspension is canceled.
+	// Default value is false.
+	DistributionPause *bool `json:"distributionPause,omitempty"`
```
Monokaix, line 228: Does `work` also need spec `DistributionPauseStrategy`?

docs/proposals/work-pause/README.md
```diff
@@ -0,0 +1,625 @@
...
+    kind: Deployment
+    name: nginx
+    namespace: default
+status:
+  conditions:
+  - lastTransitionTime: "2024-07-01T08:33:28Z"
+    message: Scheduling is in a paused state.
+    reason: SchedulePause
+    status: "True"
+    type: SchedulePause
```
Monokaix, line 400: Who will update this and when wil this condition be updated? Will this be remove when rb.PauseStrategy.SchedulePause is set to false?

docs/proposals/work-pause/README.md
```diff
@@ -0,0 +1,625 @@
...
+
+type PauseStrategy struct {
+	// SchedulePause specifies the pause scheduling of resources matched by the policy,
+	// set to false indicates the pause is canceled.
+	// Default value is false.
+	// Note: The extension design is not a focal point of the current proposal.
+	SchedulePause *bool `json:"schedulePause,omitempty"`
+
+	// DistributionPauseStrategy specifies the strategy for pausing resource distribution.
+	DistributionPauseStrategy *DistributionPauseStrategy `json:"distributionPauseStrategy,omitempty"`
```
Monokaix, line 211: This filed is only transparently transmitted to the `work` resource and will not be used at the `rb` level, right?

Monokaix:
> Hi! Thanks for your proposal!
> 
> In Solution One, the `SchedulePause` of `ResourceBinding` is propagated through `PropagationPolicy`. If I create a `PropagationPolicy` with a `PauseStrategy` set to `nil` by default, is it feasible to only update the `SchedulePause` of `ResourceBinding` to true via a webhook when the `karmada detector` creates the `ResourceBinding`? If this is feasible, does "propagation" mean that the associated `ResourceBinding/Work` is only updated when the `PropagationPolicy` updates the `PauseStrategy` field?
> 
> I also have a suggestion: can this issue be considered as a user story? In brief, it involves pausing the `ResourceBinding` at the time of creation through a webhook, and then using a custom controller to determine when certain `ResourceBindings` can be scheduled.
> 
> #4919

+1

XiShanYongYe-Chang:
> is it feasible to only update the SchedulePause of ResourceBinding to true via a webhook when the karmada detector creates the ResourceBinding?

For solution 1, the source of user control is in the PropagationPolicy. If it is modified through webhook, it will be inconsistent with the declaration of PropagationPolicy.

I think solution 2 might be better for your scenario.

XiShanYongYe-Chang:
> In brief, it involves pausing the ResourceBinding at the time of creation through a webhook, and then using a custom controller to determine when certain ResourceBindings can be scheduled.

I understand that this is a specific implementation, not a user case. In fact, the two actions you point out can be done by controlling the pause of RB. Karmada is not aware of when you pause and resume.

That's why I pointed out that solution 2 is more suitable for your use case.

docs/proposals/work-pause/README.md
XiShanYongYe-Chang: Good question!
I think the binding-status-controller will take care of this task.

> Will this be remove when rb.PauseStrategy.SchedulePause is set to false?

The status will be set to false when rb.PauseStrategy.SchedulePause is set to false.

docs/proposals/work-pause/README.md
XiShanYongYe-Chang: It does not. `Work` only needs to know whether it is paused. It does not need to know the pause status of other clusters.

docs/proposals/work-pause/README.md
XiShanYongYe-Chang: Yes, it is.

Monokaix:
When the user sets pause at the pp or cr level, other components such as wenhook may also set pause/resume at the rb or work level. There will be a conflict here. We should explain this limitation: )

XiShanYongYe-Chang:
Yes, if you need to set it up via a webhook, the user needs to be aware of the restrictions.

docs/proposals/work-pause/README.md
```diff
@@ -0,0 +1,625 @@
...
+
+这个用例可以参考 https://github.com/karmada-io/karmada/issues/4688 中描述的用例2，当执行完这些步骤之后，https://github.com/karmada-io/karmada/issues/4688#issuecomment-2065332647，用户应该会需要执行多个操作，如：修改 Deployment 副本数为3，修改 PropagationPolicy 将 cluster-blue 移除 clusteraffinify 中，删除 OverridePolicy 等，当所有这些操作完成后，再将 work 从暂停中恢复，这样不会因中间的某个操作导致资源被同步下去，从而对用户业务产生影响。
+-->
+
+As an operator, I am currently facing a scenario where there is a Deployment with a replica of 0 that has been distributed to member1 and member2 clusters. Through the OverridePolicy configuration, the replica for distribution to member2 is set to 3. Now I need to change the replica count of the Deployment to 3, and modify the OverridePolicy to set the replica for distribution to member1 to 0. These two steps cannot be executed simultaneously, so I hope to pause the synchronization of the Deployment distribution. Once both of the above steps have been fully executed, I can then resume the distribution process, ensuring that the intermediate execution results do not affect the changes in the Deployment on the member clusters.
```
RainbowMango, line 103: The scenario does not seem like a common use case, it's hard to understand. Why does `the operator` distribute `0` replica deployment to clusters?

XiShanYongYe-Chang: I'm not sure if this scenario is common, this scenario occurs when workloads are migrated from cluster A to cluster B.

docs/proposals/work-pause/README.md
```diff
@@ -0,0 +1,625 @@
...
+```yaml
+apiVersion: work.karmada.io/v1alpha1
+kind: Work
+metadata:
+  name: nginx-xxx
+  namespace: karmada-es-member1
+spec:
+  workload:
+    manifests:
+      - xxx
```
RainbowMango, line 274: Missing the specification field.
```suggestion
spec:
  distributionPause: true
  workload:
    manifests:
      - xxx
```

CharlesQQ:
In fact, what I am concerned about is whether the new work object has been modified by the latest overridepolicy after resumption of propagation. This must be ensured.

XiShanYongYe-Chang:
> In fact, what I am concerned about is whether the new work object has been modified by the latest overridepolicy after resumption of propagation. This must be ensured.

Good question! What you say should be the definitive expected behavior, because when you change the op, work is updated, and when work is resumed, it is updated with the latest spec.

docs/proposals/work-pause/README.md
```diff
@@ -0,0 +1,625 @@
...
+	// Default value is false.
+	// Note: The extension design is not a focal point of the current proposal.
+	SchedulePause *bool `json:"schedulePause,omitempty"`
+
+	// DistributionPauseStrategy specifies the strategy for pausing resource distribution.
+	DistributionPauseStrategy *DistributionPauseStrategy `json:"distributionPauseStrategy,omitempty"`
+}
+```
+
+A new ConditionType called `SchedulePause` is added to describe the scheduling pause status of the ResourceBinding/ClusterResourceBinding resources.
```
chaunceyjiang, line 215: I think we should also generate some events similar to `SchedulePause`, right?

RainbowMango:
I would prefer to introduce this functionality in PropagationPolicy as it is more convenient for users and straightforward than introducing another API.  

Also proposing the API:
```golang
type PropagationSpec struct {
	// Suspension declares the policy for suspending different aspects of propagation.
	// nil means no suspension. no default values.
	// +optional
	Suspension *Suspension `json:"suspension,omitempty"`
}

// Suspension defines the policy for suspending different aspects of propagation.
type Suspension struct {
	// Scheduling controls whether scheduling should be suspended.
	// +optional
	// Note: Postpone this until there is a solid use case.
	// Scheduling *bool `json:"scheduling,omitempty"`

	// Dispatching controls whether dispatching should be suspended.
	// nil means not suspend, no default value, only accepts 'true'.
	// Note: true means stop propagating to all clusters.
	// +optional
	Dispatching *bool `json:"dispatching,omitempty"`

	// DispatchingOnClusters declares a list of clusters to which the dispatching
	// should be suspended.
	// +optional
	DispatchingOnClusters *SuspendClusters `json:"dispatchingOnClusters,omitempty"`
}

// SuspendClusters represents a group of clusters that should be suspended from propagating to.
// Note: No plan to introduce the label selector or field selector to select clusters yet, as it
// would make the system unpredictable.
type SuspendClusters struct {
	// ClusterNames is the list of clusters to be selected.
	// +optional
	ClusterNames []string `json:"clusterNames,omitempty"`
}
```

Monokaix:
Another question here, after scheduling or distribution paused, will there be any problems with resource occupation calculation? For example, will resources be pre-occupied after the work is paused? If resources are occupied and multiple workare resumed at the same time, will resource calculation errors be caused?

docs/proposals/work-pause/README.md
XiShanYongYe-Chang: Thanks, I think yes!

XiShanYongYe-Chang:
> Another question here, after scheduling or distribution paused, will there be any problems with resource occupation calculation? For example, will resources be pre-occupied after the work is paused? If resources are occupied and multiple workare resumed at the same time, will resource calculation errors be caused?

For scheduling paused, resources will not be pre-occupied. For distribution pause, I think it will.

RainbowMango:
I vote for solution one and sent the draft API on https://github.com/karmada-io/karmada/pull/5118#issuecomment-2226864919.

a7i:
That sounds good!

Once approved/finalized, would you be open to contributions or did you want to lead this change?

XiShanYongYe-Chang:
> Once approved/finalized, would you be open to contributions or did you want to lead this change?

Of course, I will create a new issue to track the tasks that need to be completed, and I look forward to your participation in some of these tasks.

RainbowMango:
Hi Amir, After we make the decision here, we can get started from #4838(the one you sent 3 months ago, sorry again for the delay). 
 As @XiShanYongYe-Chang said above, he will help to open an umbrella issue to track all the tasks there, including code, testing, etc, feel free to take any tasks you are interested in, even all of them, just leave a comment on that umbrella issue would be fine. Thank you in advance!

CharlesQQ:
> I vote for solution one and sent the draft API on https://github.com/karmada-io/karmada/pull/5118#issuecomment-2226864919.

+1

docs/proposals/work-pause/README.md
```diff
@@ -0,0 +1,541 @@
...
+## Proposal
+
+### User Stories (Optional)
+
+#### Story 1
+
+<!--
+作为一名运维者，当 Karmada 控制面与成员集群争夺资源的控制权时，会出现资源反复被更新的情况，我希望能够暂停该资源向成员集群进行同步的流程，从而能够帮助我快速进行问题定位。
+
+PS：为什么不是直接暂停 ResourceBinding 和 Work 的生成呢？这是因为生成 Work 表示 Karmada 控制面中的工作已经全部完成，只剩资源同步了，这样可以更好的帮忙用户进行紧急维护或故障排除。
```
Vacant2333, line 73: this line have not add the translate content.

docs/proposals/work-pause/README.md
```diff
@@ -0,0 +1,541 @@
...
+<!--
+在业务发布时，通过在一些 workload 类型的资源中设置 partition 字段，例如 StatefulSet 资源，可以指定更新部分实例，达成一定数量的灰度发布。
+
+作为一名运维者，我在联邦控制面为 Deployment 资源指定更新部分实例，在分发到不同集群时，需要将待更新的副本数量进行拆分。在计算不同集群的拆分数量时，希望能够先暂停 Deployment 资源的分发流程，待计算完成之后并将计算结果应用于系统中，然后再恢复 Deployment 资源的分发流程，从而达到灰度发布的目的。相关issue：https://github.com/karmada-io/karmada/issues/1567 和 https://github.com/karmada-io/karmada/issues/4421。
+
+这个用例其实与小红书公司的用例 https://github.com/karmada-io/karmada/pull/5056 有点像，区别是这个用例中需要处理的 partition 字段是位于 annotation 中的。
+
+https://github.com/karmada-io/karmada/issues/4421 与 https://github.com/karmada-io/karmada/issues/1567 相关，前者其实是后者的一个具体解，但前者描述的实际上也是一个 Y 问题，X 问题其实是“在联邦层面，如果保证Deployment应用发布与单集群保持一致”。
+
+Kubernetes中单个Deployment资源实际上不支持灰度发布/分批次发布，issue中描述的金丝雀发布是陌陌公司选择的灰度发布实践中的一种，陌陌针对该实践进行了额外的开发工作，使其能够满足单个Deployment资源即可进行灰度发布。
```
Vacant2333, line 97: ditto

docs/proposals/work-pause/README.md
```diff
@@ -0,0 +1,541 @@
...
+As an operator, when I specify to update a portion of instances for a Deployment resource in the federated control plane, it is necessary to split the number of replicas to be updated when distributing to different clusters. I would like to pause the propagation process of the Deployment resource when calculating the split number for different clusters. After the calculation is complete and the results are applied to the system, I would then resume the propagation process of the Deployment resource to achieve the goal of canary deployment.
+
+Related issue: https://github.com/karmada-io/karmada/issues/1567, https://github.com/karmada-io/karmada/issues/4421.
+
+#### Story 4
+
+<!--
+作为一名运维者，我希望当在联邦控制面进行业务变更时，不直接将变更内容同步至成员集群，而是停留在控制面中，这也我可以通过检查控制面中的资源变化，例如 ResourceBinding、Work 等资源中的内容，来确定系统是否按照预期执行，待确定完成之后，再恢复业务的变更，将变更结果同步至成员集群中。
+
+这个用例可以参考 https://github.com/karmada-io/karmada/issues/4688 中描述的用例2，当执行完这些步骤之后，https://github.com/karmada-io/karmada/issues/4688#issuecomment-2065332647，用户应该会需要执行多个操作，如：修改 Deployment 副本数为3，修改 PropagationPolicy 将 cluster-blue 移除 clusteraffinify 中，删除 OverridePolicy 等，当所有这些操作完成后，再将 work 从暂停中恢复，这样不会因中间的某个操作导致资源被同步下去，从而对用户业务产生影响。
```
Vacant2333, line 111: ditto

docs/proposals/work-pause/README.md
```diff
@@ -0,0 +1,541 @@
...
+<!--
+当前，我提出了两种方案来实现当前特性，请大家给出一些具体的意见，也欢迎大家提出新的方案。让我们一同来选择一种方案，推进该特性向前发展。
+-->
+
+Currently, I've proposed two approaches to implement the current feature, and I'd appreciate concrete feedback from all of you. I also welcome suggestions for alternative solutions. Let's collaborate to select an approach and move forward with developing this feature.
+
+### Solution one
+
+<!--
+在 PP 上增加字段来描述各种暂停需求，包括：PP功能暂停、分发功能暂停、调度功能暂停、指定集群暂停等。
```
Vacant2333, line 133: ditto

docs/proposals/work-pause/README.md
```diff
@@ -0,0 +1,541 @@
+---
+title: Support for cluster-level resource propagation pause and resume capabilities
+
+authors:
+- "@XiShanYongYe-Chang"
+
+reviewers:
+- TBD
```
Vacant2333, line 8: need update

docs/proposals/work-pause/README.md
XiShanYongYe-Chang: I'm going to delete that part of the proposal so that it's not part of the final proposal.

docs/proposals/dispatch-suspension/README.md
```diff
@@ -0,0 +1,489 @@
...
+
+// Suspension defines the policy for suspending different aspects of propagation.
+type Suspension struct {
+	// Scheduling controls whether scheduling should be suspended.
+	// +optional
+	// Note: Postpone this until there is a solid use case.
+	// Scheduling *bool `json:"scheduling,omitempty"`
+
+	// Dispatching controls whether dispatching should be suspended.
+	// nil means not suspend, no default value, only accepts 'true'.
```
RainbowMango, line 104:
```suggestion
	// nil means not suspend.
```
Here we don't have to restrict only accepting `true`.

docs/proposals/dispatch-suspension/README.md
```diff
@@ -0,0 +1,489 @@
...
+```go
+// WorkSpec defines the desired state of Work.
+type WorkSpec struct {
+	...
+
+	// Dispatching controls whether dispatching should be suspended.
+	// nil means not suspend, no default value, only accepts 'true'.
+	// Note: true means stop propagating to all clusters.
+	// +optional
+	Dispatching *bool `json:"dispatching,omitempty"`
```
RainbowMango, line 152:
```suggestion
	SuspendDispatching *bool `json:"suspendDispatching,omitempty"`
```
nil means do not suspend, and it's the default behavior.

docs/proposals/dispatch-suspension/README.md
```diff
@@ -0,0 +1,489 @@
...
+```yaml
+apiVersion: work.karmada.io/v1alpha1
+kind: Work
+metadata:
+  name: nginx-xxx
+  namespace: karmada-es-member1
+spec:
+  workload:
+    manifests:
+      - xxx
```
RainbowMango, line 197:
```suggestion
  workload:
    manifests:
      - xxx
  suspendDispatching: true
```

RainbowMango (approved):
/lgtm
/approve

Issue #5048, opened 2024-06-13
[Umbrella] Enhancement of Karmada maturity based on Clomonitor check sets


zhzhuang-zju:
**What would you like to be added**:
[CLOMonitor](https://clomonitor.io/) is a tool that periodically checks open source projects repositories to verify they meet certain project health best practices, and will provide a score card for each project based on the check results. Here is the score card for karmada: https://clomonitor.io/projects/cncf/karmada. As you can see, there's still some work to be done.
Here list the check set that karmada did not pass.
- [x] [license scanning](https://clomonitor.io/docs/topics/checks/#license-scanning) (@B1F030 #5050)
   Karamda completes the license scanning software scans in [fossa](https://github.com/karmada-io/karmada/blob/master/.github/workflows/fossa.yml) workflow, so we only need add the `FOSSA` link in the `README` file. Regexps used: `"(https://app.fossa.(?:io|com)/projects/[^"'\)]+)"`

- [x] [Artifact Hub badge](https://clomonitor.io/docs/topics/checks/#artifact-hub-badge) (@B1F030 #5051)

- [x] [OpenSSF Scorecard badge](https://clomonitor.io/docs/topics/checks/#openssf-scorecard-badge) (@B1F030 #5055)
We have added the `OpenSSF` Scorecard badge in the `README` file in #5022, but the url `https://api.scorecard.dev/projects/github.com/karmada-io/karmada/badge` does not match the specified regexps `"(https://api.securityscorecards.dev/projects/github.com/[^/]+/[^/]+)/badge"` and needs to be modified.

- [x] [Dependencies policy](https://clomonitor.io/docs/topics/checks/#dependencies-policy) （@zhzhuang-zju）

- [x] [Software bill of materials (SBOM)](https://clomonitor.io/docs/topics/checks/#software-bill-of-materials-sbom)(@zhzhuang-zju #5110)

- [x] [Security insights](https://clomonitor.io/docs/topics/checks/#security-insights) (@zhzhuang-zju, #5140)

- [x] [Signed releases](https://clomonitor.io/docs/topics/checks/#signed-releases-from-openssf-scorecard)
Karmada images have been signed with `cosign` since `release1.7`, but we need add the [SLSA provenance file](https://slsa.dev/spec/v0.1/index) in the assets for release (*.intoto.jsonl) (@zhzhuang-zju #5178)

[...]

PR #5119, opened 2024-07-02
set the permissions required for job in ci-image-scanning.yaml

zhzhuang-zju:
**What type of PR is this?**
/kind cleanup

refer to https://github.com/karmada-io/karmada/issues/5048#issuecomment-2201857283, set the permissions required for the job in ci-image-scanning.yaml

Parts of #5048

```release-note

```

### DIFF
--- a/.github/workflows/ci-image-scanning.yaml
+++ b/.github/workflows/ci-image-scanning.yaml
@@ -5,13 +5,12 @@ on:
     # for PRs initiated by Dependabot.
     branches-ignore:
       - 'dependabot/**'
-
 permissions:
-  contents: read # for actions/checkout to fetch code
-  security-events: write # for github/codeql-action/upload-sarif to upload SARIF results
-  
+  contents: read
 jobs:
   use-trivy-to-scan-image:
+    permissions:
+      security-events: write  # for github/codeql-action/upload-sarif to upload SARIF results
     name: image-scanning
     if: ${{ github.repository == 'karmada-io/karmada' }}
     runs-on: ubuntu-22.04


.github/workflows/ci-image-scanning.yaml
```diff
@@ -5,13 +5,12 @@ on:
...
-
 permissions:
-  contents: read # for actions/checkout to fetch code
-  security-events: write # for github/codeql-action/upload-sarif to upload SARIF results
-  
+  contents: read
 jobs:
   use-trivy-to-scan-image:
+    permissions:
+      security-events: write  # for github/codeql-action/upload-sarif to upload SARIF results
```
liangyuanpeng, line 13: I'm a little hesitant about this, it will only distract we, especially since this only have one job.

zhzhuang-zju: The change here is actually due to two considerations:
- Avoid top-level permissions inheriting to job-level, making job-level permissions too large. This is because the permissions needed for different jobs can vary greatly. A good practice is: setting as read-only at the [top level](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#permissions) and the required write permissions are declared at the [run-level](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idpermissions)
- refer to https://scorecard.dev/viewer/?uri=github.com/karmada-io/karmada, it's not recommended that topLevel 'security-events' permission set to 'write'.

liangyuanpeng: thanks for explain, it's make sense for me. 
/lgtm 
/approve

PR #5124, opened 2024-07-03
Supports controlling the starting order number of statefulset in mult…

wangxf1987:
…iple cluster

**What type of PR is this?**
Supports controlling the starting order number of statefulset in multiple cluster

In multiple cluster, which controlled by karmada. it's expect to order number for statfulset, that like to single cluster.
we  improve the scheduler and controller-manager, and controller to order number in the multiple cluster.
image

Fixes #

```release-note

```

### DIFF
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -20220,6 +20220,11 @@
           "description": "Replicas in target cluster",
           "type": "integer",
           "format": "int32"
+        },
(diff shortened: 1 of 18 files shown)


liangyuanpeng:
It seems that what you really need is a cross-cluster stateful application,There is a draft proposal for this.

Have you applied your changes to a prod environment? This is well worth sharing at a conference if you want!

PR #5141, opened 2024-07-05
 Fix test for corev1.LoadBalancerIngress,add IP model to VIP when ip is setting.

liangyuanpeng:
**What type of PR is this?**

/kind failing-test

since kubernetes 1.30 begin set IPMode of LoadBalancer service to VIP by default if IP is setting, So we need to set the IPMode to VIP for test.
check failing tests: https://github.com/karmada-io/karmada/actions/runs/9734577204/job/26862785337

check the kubernetes side: https://github.com/kubernetes/kubernetes/blob/d27e3462baa419532209a7515bd24de36b2a5020/pkg/apis/core/v1/defaults.go#L147-L151

also can check the reproduce workflow: https://github.com/karmada-io/karmada/actions/runs/9804835683/job/27073502218?pr=5141

just change the karmada-apiserver to v1.30.0 for reproduce it.

and i open a PR https://github.com/karmada-io/karmada/pull/5142 to adding the master branch to run karmada-apiserver compatibility workflow, So that we can find karmada-apiserver compatibility issues on the master branch.

Fixes #

```release-note

```

### DIFF
--- a/test/e2e/resource_test.go
+++ b/test/e2e/resource_test.go
@@ -172,12 +172,12 @@ var _ = ginkgo.Describe("[resource-status collection] resource status collection
 				for index, clusterName := range framework.ClusterNames() {
 					clusterClient := framework.GetClusterClient(clusterName)
 					gomega.Expect(clusterClient).ShouldNot(gomega.BeNil())
-
 					ingresses := []corev1.LoadBalancerIngress{{IP: fmt.Sprintf("172.19.1.%d", index+6)}}
 					for _, ingress := range ingresses {
 						svcLoadBalancer.Ingress = append(svcLoadBalancer.Ingress, corev1.LoadBalancerIngress{
 							IP:       ingress.IP,
 							Hostname: clusterName,
+							IPMode:   ingress.IPMode,
 						})
 					}
 
@@ -198,8 +198,14 @@ var _ = ginkgo.Describe("[resource-status collection] resource status collection
 					latestSvc, err := kubeClient.CoreV1().Services(serviceNamespace).Get(context.TODO(), serviceName, metav1.GetOptions{})
 					g.Expect(err).NotTo(gomega.HaveOccurred())
 
+					// TODO:  Once karmada-apiserver v1.30 deploy by default,delete the following five lines, see https://github.com/karmada-io/karmada/pull/5141
+					for i := range latestSvc.Status.LoadBalancer.Ingress {
+						if latestSvc.Status.LoadBalancer.Ingress[i].IPMode != nil {
+							latestSvc.Status.LoadBalancer.Ingress[i].IPMode = nil
+						}
+					}
 					klog.Infof("the latest serviceStatus loadBalancer: %v", latestSvc.Status.LoadBalancer)
-					return reflect.DeepEqual(latestSvc.Status.LoadBalancer, svcLoadBalancer), nil
+					return reflect.DeepEqual(latestSvc.Status.LoadBalancer.Ingress, svcLoadBalancer.Ingress), nil
 				}, pollTimeout, pollInterval).Should(gomega.Equal(true))
 			})
 		})


zhzhuang-zju:
> and i open a PR https://github.com/karmada-io/karmada/pull/5142 to adding the master branch to run karmada-apiserver compatibility workflow, So that we can find karmada-apiserver compatibility issues on the master branch.

Support, so that incompatibilities can be detected as early as possible

zhzhuang-zju:
That's weird, workflow `e2e test(v1.30.0)` fails too.

liangyuanpeng:
Update the test for it, the feature gate LoadBalancerIPMode is beta at v1.30,  see https://github.com/kubernetes/kubernetes/blob/95debfb5b6dcfa2467f2700d81df6767a2c91bac/pkg/features/kube_features.go#L903-L908

zhzhuang-zju:
@liangyuanpeng This change may need to be synchronized to release-1.10. Can you remind me why release-1.8 and release-1.9 are seems not affected? Refer to https://github.com/karmada-io/karmada/actions/runs/9734577204/job/26862785337, `e2e test(release-1.8)` and `e2e test(release-1.9)` all succeeded.

liangyuanpeng:
> This change may need to be synchronized to release-1.10. 

Yes

> Can you remind me why release-1.8 and release-1.9 are seems not affected?

@zhzhuang-zju 
This field is added in Kubernetes 1.29, it's mean client-go requires 0.29 k8s deps to see this field. the karmada 1.10 k8s deps version is v0.29.4  , so the service object have field of IPMode, and then test failed.

zhzhuang-zju:
> @zhzhuang-zju This field (IPMode) is added in Kubernetes 1.29, it's mean client-go requires 0.29 k8s deps to see this field. the karmada 1.10 k8s deps version is v0.29.4 , so the service object have field of IPMode, and then test failed.

thanks, get it~

zhzhuang-zju:
kindly ping @liangyuanpeng 
How's it going?

liangyuanpeng:
@zhzhuang-zju 
CI is green, https://github.com/karmada-io/karmada/actions/runs/9832782430

so i  squash the commits and looking for `lgtm` and `approve`

zhzhuang-zju:
thanks. The conclusion so far is that if the karmada-apiserver version is `v1.30.0`, and the k8s deps version is larger than `v0.29.0`, it will cause the current e2e to fail. Is that right?

Keep digging, will the following configuration also have the same phenomenon?
```yaml
karmada:
     k8s deps: v1.29.4

member1:
     k8s deps: v1.29.4
     kube-apiserver: v1.30.0
```

liangyuanpeng:
> karmada:
>     k8s deps: v1.29.4
>
>member1:
>      k8s deps: v1.29.4
>     kube-apiserver: v1.30.0

@zhzhuang-zju 
Seems like you're talking about  the CI of `CI Workflow / e2e test (v1.30.0) (pull_request) ` , it's green now.

zhzhuang-zju:
> Seems like you're talking about the CI of `CI Workflow / e2e test (v1.30.0) (pull_request) ` 

e2e test(v1.30.0) actually refers to the version of `karmada-apiserver` being `v1.30.0`, not the `kube-apiserver` of the member cluster.
But as you said, just upgrading the `kube-apiserver` of the member clusters to `v1.30.0` is not going to be a problem. Since `karmada-apiserver` does not yet support filed `ipmode` when updating the status of a service, the final aggregateStatus of the service will not have `ipmode`.

zhzhuang-zju:
@liangyuanpeng thanks for your work, and I think your modification would solve the problem we are currently experiencing. My concern, though, is whether our e2e test is required to deal with versioning issues like this, and we may encounter more of these in the future. To address this issue, can we ignore the `ipmode` field, and only determine whether the part of status that does not contain `ipmode` is consistent before and after? So that we can avoid the inconsistency caused by version changes.  WDYT. And cc @XiShanYongYe-Chang for another look.

liangyuanpeng:
> e2e test(v1.30.0) actually refers to the version of karmada-apiserver being v1.30.0, not the kube-apiserver of the member cluster.

No, it's kube-apiserver version for of member cluster and karmada host. check: 

- https://github.com/karmada-io/karmada/blob/2ec689dff287826fe51401d7814358a2e850a01a/.github/workflows/ci.yml#L144
-   https://github.com/karmada-io/karmada/blob/2ec689dff287826fe51401d7814358a2e850a01a/hack/create-cluster.sh#L101

>  can we ignore the ipmode field, and only determine whether the part of status that does not contain ipmode is consistent before and after? So that we can avoid the inconsistency caused by version changes

Sure, let me update for it

zhzhuang-zju:
> No, it's kube-apiserver version of member cluster and karmada host. 

Oh, I see. You mean e2e test(v1.30.0) in ci `CI Workflow`, not ci `APIServer compatibility`.

> Sure, let me update for it

Thanks for the quick response, but I'm not quite sure if this is better yet. cc @XiShanYongYe-Chang

XiShanYongYe-Chang:
I think it's okay. Maybe the comparison logic could be simpler and just compare the slice lengths of loadbalancer.

zhzhuang-zju:
> I think it's okay. Maybe the comparison logic could be simpler and just compare the slice lengths of loadbalancer.

Do you mean it is enough to just compare the value of `len(latestSvc.Status.LoadBalancer.Ingress` and `len(svcLoadBalancer.Ingress)`?

liangyuanpeng:
CI APIServer compatibility is green: https://github.com/karmada-io/karmada/actions/runs/10233972881

liangyuanpeng:
@zhzhuang-zju @XiShanYongYe-Chang   I think this is ready to go :)

zhzhuang-zju:
> @zhzhuang-zju @XiShanYongYe-Chang I think this is ready to go :)

/lgtm
yeah, I hope ci `APIServer compatibility` can run successfully on this weekend. cc @XiShanYongYe-Chang

test/e2e/resource_test.go
```diff
@@ -198,8 +198,13 @@ var _ = ginkgo.Describe("[resource-status collection] resource status collection
 					latestSvc, err := kubeClient.CoreV1().Services(serviceNamespace).Get(context.TODO(), serviceName, metav1.GetOptions{})
 					g.Expect(err).NotTo(gomega.HaveOccurred())
 
+					for i := range latestSvc.Status.LoadBalancer.Ingress {
+						if latestSvc.Status.LoadBalancer.Ingress[i].IPMode != nil {
+							latestSvc.Status.LoadBalancer.Ingress[i].IPMode = nil
```
XiShanYongYe-Chang, line 203: Will the code need to be adjusted if all our running versions support this field in the future? Maybe add some notes here.

XiShanYongYe-Chang (commented):
Thanks @liangyuanpeng 
/approve

XiShanYongYe-Chang:
/cc @zhzhuang-zju 
Do we need to cherry-pick it?

XiShanYongYe-Chang (commented):
/lgtm

PR #5144, opened 2024-07-05
GroupClusters should sort by score and availableReplica count

### LINKED ISSUE #5129: GenerateClustersInfo should take AvailableReplicas into account when sorting
**What happened**:
When scheduling, Karmada will select the clusters out of the list of candidates (ClusterDetailInfo) sorted by cluster.Score descending. This sorting is done by [spreadconstraint#sortClusters](https://github.com/karmada-io/karmada/blob/4a32670cfb91b3e4a4d8ad40a434352c7c1027b6/pkg/scheduler/core/spreadconstraint/util.go#L43), and called when running [generateClustersInfo](https://github.com/karmada-io/karmada/blob/4a32670cfb91b3e4a4d8ad40a434352c7c1027b6/pkg/scheduler/core/spreadconstraint/group_clusters.go#L146). Currently the sort only accounts for cluster.Score, but in edge cases where the score between clusters is the same, we should prefer scheduling on clusters with more availableReplicas. Instead, we just pick the first cluster our of the sorted list:

```
I0703 18:28:55.882315       1 util.go:78] MaxAvailableReplica scores calculated by estimator general-estimator for workload(kind=FlinkDeployment, s-spaaseng/mszacillo-karmada): [{cluster-A 60} {cluster-B 64}]
I0703 18:28:55.886486       1 util.go:78] MaxAvailableReplica scores calculated by estimator scheduler-estimator for workload(kind=FlinkDeployment, s-spaaseng/mszacillo-karmada): [{cluster-A 9} {cluster-B 20}]
I0703 18:28:55.886575       1 util.go:102] Target cluster calculated by estimators (available cluster && maxAvailableReplicas): [{cluster-A 9} {cluster-B 20}]
I0703 18:28:55.886594       1 select_clusters_by_cluster.go:34] Selecting best clusters using scores: [{cluster-A 0 9 cluster-A} {cluster-B 0 20 cluster-B}].
I0703 18:28:55.886668       1 generic_scheduler.go:101] Selected clusters: [cluster-A]
```

Ideally, when creating the default list of ClusterInfo, we should additionally sort by availableReplicas. So in the above example, Karmada should select `Cluster-B` since the score is the same, but Cluster-B has more availableReplicas (20).

This is easily reproducible when clusters have the same ClusterAffinity / ClusterLocality scores. In that case, the clusters are only sorted by score rather than availableReplica.  

**Potential fixes**:
I was able to address this by adding an additional compareFunction to the generateClusterInfo method:

```
sortClusters(info.Clusters, func(i *ClusterDetailInfo, j *ClusterDetailInfo) *bool {
	if i.AvailableReplicas != j.AvailableReplicas {
		return pointer.Bool(i.AvailableReplicas > j.AvailableReplicas)
	}
	return nil
})
```

Out of curiosity, do we see future cases for custom compareFunctions in sortClusters definition? This could also be fixed by sorting clusters by default using these rules:
- Sort by cluster Score
- If Score is equal, sort by availableReplicas

**Environment**:
- Karmada version: v1.9.0
- Kubernetes version: v1.29.0

mszacillo:
**What type of PR is this?**

/kind feature

With the PR, Karmada will sort clusterInfo by score, and fallback to sorting by availableReplicas. This will help cases in which some clusters have the same score, but we should prefer scheduling on clusters with higher availableReplicas.

Fixes #5129

```release-note
`karmada-scheduler`: GroupClusters will sort clusters by score and availableReplica count.
```

### DIFF
--- a/pkg/scheduler/core/spreadconstraint/group_clusters.go
+++ b/pkg/scheduler/core/spreadconstraint/group_clusters.go
@@ -17,6 +17,8 @@ limitations under the License.
 package spreadconstraint
 
 import (
+	"k8s.io/utils/ptr"
+
 	clusterv1alpha1 "github.com/karmada-io/karmada/pkg/apis/cluster/v1alpha1"
 	policyv1alpha1 "github.com/karmada-io/karmada/pkg/apis/policy/v1alpha1"
 	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
@@ -143,7 +145,12 @@ func (info *GroupClustersInfo) generateClustersInfo(clustersScore framework.Clus
 		info.Clusters[i].AvailableReplicas += int64(rbSpec.AssignedReplicasForCluster(clustersReplica.Name))
 	}
 
-	sortClusters(info.Clusters)
+	sortClusters(info.Clusters, func(i *ClusterDetailInfo, j *ClusterDetailInfo) *bool {
+		if i.AvailableReplicas != j.AvailableReplicas {
+			return ptr.To(i.AvailableReplicas > j.AvailableReplicas)
+		}
+		return nil
+	})
 }
 
 func (info *GroupClustersInfo) generateZoneInfo(spreadConstraints []policyv1alpha1.SpreadConstraint) {
--- a/pkg/scheduler/core/spreadconstraint/util_test.go
+++ b/pkg/scheduler/core/spreadconstraint/util_test.go
@@ -20,6 +20,8 @@ import (
 	"reflect"
 	"testing"
 
+	"k8s.io/utils/ptr"
+
 	policyv1alpha1 "github.com/karmada-io/karmada/pkg/apis/policy/v1alpha1"
 )
 
@@ -75,9 +77,10 @@ func TestIsSpreadConstraintExisted(t *testing.T) {
 
 func Test_sortClusters(t *testing.T) {
 	tests := []struct {
-		name  string
-		infos []ClusterDetailInfo
-		want  []ClusterDetailInfo
+		name            string
+		infos           []ClusterDetailInfo
+		want            []ClusterDetailInfo
+		compareFunction func(*ClusterDetailInfo, *ClusterDetailInfo) *bool
 	}{
 		{
 			name: "different scores",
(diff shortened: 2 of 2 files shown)


pkg/scheduler/core/spreadconstraint/group_clusters.go
```diff
@@ -17,6 +17,8 @@ limitations under the License.
 package spreadconstraint
 
 import (
+	"k8s.io/utils/ptr"
```
mszacillo, line 20: Worth noting - the linter was failing because the original import (`k8s.io/utils/pointer`) is deprecated. This new import will work for newer karmada versions, but is not currently compatible with the version we use (v1.9.0) because it uses an older k8s-util version.

RainbowMango: Thanks for the information. 
I wonder if you maintain some private commits based on v1.9.0? 
I'm asking because I want to know if this going to be a problem for you. I mean if this patch not present on release 1.9.

mszacillo: We have a couple private commits on our v1.9.0 branch, but most of them we are planning to contribute via our published proposals. I think we can try updating to the v1.10.0 release branch which will have the updated k8s-util version and can pick up this commit. I'll test v1.10 with out setup and confirm.

whitewindmills (commented):
/lgtm
note that you need to squash your commits

mszacillo:
Thanks for the review - squashed my commits.

pkg/scheduler/core/spreadconstraint/group_clusters.go
mszacillo: Confirmed changes work with v1.10 - I think we can move forward with upgrading our local Karmada version.

mszacillo:
@whitewindmills Will this commit be cherrypicked into the v1.10 release branch?

RainbowMango:
Normally we don't backport this kind of PR, but this seems no harm/risk in doing that, so I'm happy to backport it to meet users' expectations. 

If no objection from @whitewindmills, we can do that. (PS: there is a doc explains [how to cherry-pick PRs](https://karmada.io/docs/next/contributor/cherry-picks/))

whitewindmills:
> but this seems no harm/risk in doing that, so I'm happy to backport it to meet users' expectations.

I'm ok with this.

PR #5145, opened 2024-07-05
Proposal: Custom CRD Download Strategy Support for Karmada Operator

jabellard:
**What type of PR is this?**

/kind design

Details are in the design document.

Fixes #

```release-note
NONE
```

### DIFF
--- a/docs/proposals/operator-custom-crd-download-strategy/README.md
+++ b/docs/proposals/operator-custom-crd-download-strategy/README.md
@@ -1,0 +1,110 @@
+---
+title: Custom CRD Download Strategy Support for Karmada Operator
+authors:
+- "@jabellard"
+reviewers:
+- "@RainbowMango"
+approvers:
+- "@RainbowMango"
+
+creation-date: 2024-07-05
+
+---
+
+# Custom CRD Download Strategy Support for Karmada Operator
+
+## Summary
+
+As a prerequisite for provisioning a new Karmada instance, the Karmada operator needs to ensure required CRDs are downloaded.
+By default, it will attempt to download the CRDs tarball from a release artifact on GitHub. The URL for that release artifact is a well-known location
+that is constructed based on the release version of the operator. This strategy makes the following assumptions, which may not always be correct:
+- The constructed URL is accessible from the network where the operator is running. In an internal corporate environment with network policies that,
+by default, prevent access to resources on the internet, this may not be true.
+- The version of each Karmada instance managed by the operator is identical to that of the operator's release version. That may not always be true. When creating a new Karmada instance,
+it's possible to explicitly specify the version of that instance's control plane components. Furthermore, the operator may have to manage multiple Karmada instances of heterogeneous versions.
+
+
+## Motivation
+
+The motive of this document is to propose a custom CRD downloading strategy to address the assumptions made by the Karmada operator which may not always be correct.
+
+
+### Goals
+
+The goal of this proposal is to provide the ability to specify a custom CRD downloading strategy for how the operator will download the CRDs tarball for a Karmada instance.
+Furthermore, this strategy will also ensure the version of the CRDs used to provision a new Karmada instance is based on the intended version of that instance, not the release version of the operator.
+
+## Proposal
+
+Currently, when provisioning a new Karmada instance, as a prerequisite, the Karmada operator will ensure that required CRDs are downloaded. By default, the CRDs tarball is
+downloaded from a well-known location on GitHub based on the release version of the operator. As aforementioned, this is based on the assumption that the version of the Karmada instance 
+is identical to that of the operator's release version, which may not always be true. When the version of the Karmada control plane components is not specified, the default behavior is to assume
+those components should have the same version as the operator's release version. In that scenario, the assumption is justified and holds true. However, when creating a new Karmada instance, it's possible to explicitly
+specify the version of the control plane components. In that scenario, this assumption breaks, and we encounter the risk of the operator installing the wrong CRDs for that Karmada instance.
+
+The default behavior is sensible and will be retained. However, it should not be the behavior when the intent is to create a new Karmada instance with control plane components of a different version than that of the operator's release version. 
+To support that scenario, we require the ability to explicitly specify, as part of the Karmada resource, how the operator should download the CRDs tarball for that Karmada instance.
+
+In addition to making sure required CRDs are downloaded, the Karmada operator also employs a caching strategy to avoid re-downloading CRDs it has already downloaded. It's a cache on the local file system.
+Downloaded CRDs are stored at a well-known location on the local file system based the release version of the Karmada operator. When provisioning a new Karmada instance, before attempting to download the required CRDs, the operator will fist check if
+they already exist in the cache. To support custom downloading strategies, we require this caching strategy be extended to be more generic.
+
+## Design Details
+
+To support custom CRD download strategies, we need to add the new `CRDTarball` field to the Karmada resource spec:
+```go
+// CRDDownloadPolicy specifies a policy for how the operator will download the Karmada CRD tarball
+type CRDDownloadPolicy string
+
+const (
+	// DownloadAlways instructs the Karmada operator to always download the CRD tarball from a remote location.
+	DownloadAlways CRDDownloadPolicy = "Always"
+
+	// DownloadIfNotPresent instructs the Karmada operator to download the CRDs tarball from a remote location only if it is not yet present in the local cache.
+	DownloadIfNotPresent CRDDownloadPolicy = "IfNotPresent"
+)
+
+// HTTPSource specifies how to download the CRD tarball via either HTTP or HTTPS protocol.
+type HTTPSource struct {
+	// URL specifies the URL of the CRD tarball resource.
+	URL string `json:"url,omitempty"`
+}
+
+// CRDTarball specifies the source from which the Karmada CRD tarball should be downloaded, along with the download policy to use.
+type CRDTarball struct {
+	// HTTPSource specifies how to download the CRD tarball via either HTTP or HTTPS protocol.
+	// +optional
+	HTTPSource *HTTPSource `json:"httpSource,omitempty"`
+
+	// CRDDownloadPolicy specifies a policy that should be used to download the CRD tarball.
+	// Valid values are "Always" and "IfNotPresent".
+	// Defaults to "IfNotPresent".
+	// +kubebuilder:validation:Enum=Always;IfNotPresent
+	// +kubebuilder:default=IfNotPresent
+	// +optional
+	CRDDownloadPolicy *CRDDownloadPolicy `json:"crdDownloadPolicy,omitempty"`
+}
+
+// KarmadaSpec is the specification of the desired behavior of the Karmada.
+type KarmadaSpec struct {
+	// Other fields
+
+	// CRDTarball specifies the source from which the Karmada CRD tarball should be downloaded, along with the download policy to use.
+	// If not set, the operator will download the tarball from a GitHub release.
+	// By default, it will download the tarball of the same version as the operator itself.
+	// For instance, if the operator's version is v1.10.0, the tarball will be downloaded from the following location:
+	// https://github.com/karmada-io/karmada/releases/download/v1.10.0/crds.tar.gz
+	// By default, the operator will only attempt to download the tarball if it's not yet present in the local cache.
+	// +optional
+	CRDTarball *CRDTarball `json:"crdTarball,omitempty"`
+}
+```
+
+The default downloading strategy is sensible and will be retained. As such, this new field is optional. If the intent is to provision a new Karmada instance with control plane components of the same version as the operator's
+release version, it should be left unspecified as part of the Karmada resource. Otherwise, it should be explicitly set to instruct the operator how to download the CRD tarball.
+
+Today, an `HTTPSource` can be specified to download the CRD tarball via either HTTP or HTTPS protocols. However, this design is extensible, so in the future, if needed, it's possible to extend it to support other sources.
+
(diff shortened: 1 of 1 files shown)


docs/proposals/operator-custom-crd-download-url/README.md
```diff
@@ -0,0 +1,79 @@
...
+	
+	Components *KarmadaComponents `json:"components,omitempty"`
+	
+	FeatureGates map[string]bool `json:"featureGates,omitempty"`
+
+	// Custom URL that should be used to download required CRDs for this Karmada instance.
+	// Omit if the intent is to provision a new Karmada instance with control plane components of the same version as the Karmada operator's release version.
+	// Otherwise, it should be explicitly set to ensure the right version of the CRDs is installed for this instance.
+	// +optional
+	CRDRemoteURL string `json:"crdRemoteURL,omitempty"`
```
RainbowMango, line 70: 1. We can add more comments describing where the tarball will be downloaded by default.
2. And, shall we consider adding the default value to this during the operator starts to reconcile the object?

RainbowMango: I wonder if we should leave room for future extensions.
For now, this field just requires a URL for the operator to download the tarball through `http`, shall we reserve the room for other protocols, like FTP,SFTP, etc.

RainbowMango: In case of the tarball exists in the cache, shall we always take it from the cache or download it from the URL?
Shall we introduce the download policy or something, like `Always` means always download from URL and `IfNotPresent` means will check the cache first, and then download it only when the cache missing.

RainbowMango:
@jabellard Thank you very much for the excellent proposal.
I just realized that currently, the operator can only install the Karmada instance with the version of its release version.
It's a great chance to improve it!

RainbowMango:
@calvin0327 Is there a chance for you to present [this week's community meeting](https://docs.google.com/document/d/1y6YLVC-v7cmVAdbjedoyR5WL0-q45DBRXTvz5_I7bkA/edit?pli=1#heading=h.brem0il0iadd) and talk with @jabellard and this proposal?

docs/proposals/operator-custom-crd-download-url/README.md
chaunceyjiang, line 70: In fact, within our company, we have copied the crd file into the image of karmada-operator.

So what I want to ask is, if my crd file is stored in the image, how should I fill in the crdRemoteURL?

jabellard:
@RainbowMango , thanks so so much for the feedback. I'll push an update to the proposal based on what we discussed during the meeting.

docs/proposals/operator-custom-crd-download-url/README.md
RainbowMango: > we have copied the crd file into the image of karmada-operator.

How to do that? What I can imagine there are two ways to do that:
1. Using a private karmada-operator image, the tarball can be put into the image
2. Mount the tarball into the pod, by a ConfigMap when deploying the operator

The tarball is located in the cache, so karmada-opeator doesn't need to download the tarball from GitHub.

chaunceyjiang: So what I mean is, should `crdRemoteURL` support a format like `file:///var/loca/crds`?

jabellard: This is possible and can be supported (e.g., the cache entry can just be a symbolic link the referenced location on the local filesystem), but can you please elaborate more on why that strategy would be preferred to downloading the CRDs from a remote location? How would you manually manage storage for multiple versions of the CRDs? Would that not require more work and maintenance on your part?

jabellard:
@RainbowMango , I just submitted a [pr](https://github.com/karmada-io/karmada/pull/5185) for the implementation of this proposal.

RainbowMango (commented):
/assign

docs/proposals/operator-custom-crd-download-url/README.md
```diff
@@ -0,0 +1,103 @@
...
+// Omit if the intent is to provision a new Karmada instance with control plane components of the same version as the operator's release version.
+// In that scenario, the CRDs tarball will be downloaded from a release artifact on GitHub that's of the same version as the operator's release version.
+// When that is not the intent, this field should be explicitly set to ensure the right version of the CRDs is installed for this Karmada instance.
+// +optional
+CRDRemoteURL string `json:"crdRemoteURL,omitempty"`
+
+// Policy that should be used by the Karmada operator to retrieve the Karmada CRDs when provisioning this Karmada instance.
+// Defaults to IfNotPresent.
+// +optional
+CRDDownloadPolicy CRDDownloadPolicy `json:"crdDownloadPolicy,omitempty"`
```
RainbowMango, line 89:
```suggestion
CRDDownloadPolicy *CRDDownloadPolicy `json:"crdDownloadPolicy,omitempty"`
```
ditto.

Using a pointer type might not be that convenient sometimes, like, we probably need to check whether it is nil. But I'd prefer to follow the API convention.

docs/proposals/operator-custom-crd-download-url/README.md
```diff
@@ -0,0 +1,103 @@
...
+Components *KarmadaComponents `json:"components,omitempty"`
+
+FeatureGates map[string]bool `json:"featureGates,omitempty"`
+
+// Custom URL that should be used to download required CRDs for this Karmada instance.
+// Omit if the intent is to provision a new Karmada instance with control plane components of the same version as the operator's release version.
+// In that scenario, the CRDs tarball will be downloaded from a release artifact on GitHub that's of the same version as the operator's release version.
+// When that is not the intent, this field should be explicitly set to ensure the right version of the CRDs is installed for this Karmada instance.
+// +optional
+CRDRemoteURL string `json:"crdRemoteURL,omitempty"`
```
RainbowMango, line 84: What if another user stores the tarball on a FTP server, is it a reasonable case? can we support that case?

Can we get some inspiration from Kubernetes [ProbeHandler](https://github.com/kubernetes/kubernetes/blob/4a0a7e809237e74afd4ea1499520063f96158e1e/staging/src/k8s.io/api/core/v1/types.go#L2768-L2784) which can supports HTTP, TCP, GRPC, and so on. The `ProbeHandler` is highly extendable, it can add more actions without effect the previous users.

What do you think?

jabellard: Given the ubiquitous use of HTTP/HTTPS for accessing artifacts, I find it very hard to think of scenarios where using FTP/SFTP would be preferred. However, that could be because I'm missing some context. I do agree with you that we should leave the door open to support that use case down the line if the need for it arises. I think implementing it now though is out of scope for this feature, but I'm happy you brought it up. This proposed design is flexible, so if we do decide to support other protocols like FTP/SFTP down the line, it's possible without changes to the API.
For instance, `CRDRemoteURL` can just be specified like so: `sftp://my-server.com/karmada/artifacts/v1.10.2/crds.tar.gz`. Then at runtime, based the the URL's scheme, we can decide which strategy should be used to download the CRDs, and leverage a pattern like the [ProbeHandler](https://github.com/kubernetes/kubernetes/blob/4a0a7e809237e74afd4ea1499520063f96158e1e/staging/src/k8s.io/api/core/v1/types.go#L2768-L2784) as you suggested to implement the logic.

RainbowMango: > I think implementing it now though is out of scope for this feature

Yes, exactly. I don't expect to support other protocols in this proposal. Just make room for future extensions, without breaking changes and being ambitious.  Let me see if I can help with it.

By the way, can you help to share more details about your HTTPServer where servers the CRD tarball? Does it support both HTTP and HTTPS? If it verifies the client's certificate?

RainbowMango: Hi Joe, Please help to look at how this look like:
```golang
// KarmadaSpec is the specification of the desired behavior of the Karmada.
type KarmadaSpec struct {
	// otherfields

	// CRDTarball defines the source location from where the CRD tarball should be downloaded,
	// along with the download policy.
	// If not set, the Karmada Operator will download the CRDs from the GitHub release.
	// By default, it will fetch the tarball corresponding to the version of the operator itself.
	// For instance, if the operator's version is v1.10.0, it will download from:
	// https://github.com/karmada-io/karmada/releases/download/v1.10.0/crds.tar.gz
	// Note that the download will only occur if the tarball is not already present in the local cache.
	// +optional
	CRDTarball *CRDTarball `json:"crdTarball,omitempty"`
}


// CRDDownloadPolicy specifies a policy for how the Karmada operator will retrieve
// the Karmada CRDs when provisioning a Karmada instance.
type CRDDownloadPolicy string

const (
	// DownloadAlways instructs the Karmada operator to always download the CRDs tarball from a remote location.
	DownloadAlways CRDDownloadPolicy = "Always"

	// DownloadIfNotPresent instructs the Karmada operator to download the CRDs tarball from a remote location
	// only if it is not yet present in the local cache.
	DownloadIfNotPresent CRDDownloadPolicy = "IfNotPresent"
)

// CRDTarball defines the source from which the CRD tarball can be downloaded and the download policy.
type CRDTarball struct {

	// HTTPSource is the full URL for downloading the CRD tarball using either HTTP or HTTPS protocol.
	// +optional
	HTTPSource *string `json:"httpSource,omitempty"`

	// NOTE: We still can add more fields like TLS certificate if there is a need in the future.

	// NOTE: We can add more protocol support here. But this will not be included this time.
	// FTPSource is the full URL for downloading the CRD tarball using FTP protocol.
	// If specified, HTTPSource must not be set.
	// +optional
	// FTPSource *string `json:"ftpServer,omitempty"`

	// DownloadPolicy specifies a policy that should be used by the Karmada operator to retrieve
	// the Karmada CRDs when provisioning this Karmada instance.
	// Valid values are "Always" and "IfNotPresent".
	// Defaults to "IfNotPresent".
	// +kubebuilder:validation:Enum=Always;IfNotPresent
	// +kubebuilder:default=IfNotPresent
	// +optional
	DownloadPolicy *CRDDownloadPolicy `json:"downloadPolicy,omitempty"`
}
```

jabellard: Thanks for clarifying. This structure does provide a great point of extensibility.

jabellard: Can use either HTTP OR HTTPS. No specific cert requirements.

jabellard: Will push update shortly to address comments.

docs/proposals/operator-custom-crd-download-strategy/README.md
```diff
@@ -0,0 +1,110 @@
...
+	// URL specifies the URL of the CRD tarball resource.
+	// +optional
+	URL *string `json:"url,omitempty"`
+}
+
+// CRDTarball specifies the source from which the Karmada CRD tarball should be downloaded, along with the download policy to use.
+type CRDTarball struct {
+	// HTTPSource specifies how to download the CRD tarball via either HTTP or HTTPS protocol.
+	// +optional
+	HTTPSource *HTTPSource `json:"httpSource,omitempty"`
```
RainbowMango, line 78: Looks great!

Would you mind [Squash commits](https://karmada.io/docs/next/contributor/github-workflow/#squash-commits) ? and we are ready to move forward if no further comments from other guys, like @chaunceyjiang .

jabellard: Perfect! Just squashed. Thanks again for the constructive feedback.

RainbowMango (commented):
/lgtm

@chaunceyjiang Do you have any further comments?

docs/proposals/operator-custom-crd-download-url/README.md
chaunceyjiang: > but can you please elaborate more on why that strategy would be preferred to downloading the CRDs from a remote location?


Simply put, we don't want our end users to download from remote http/https servers. Because that requires our end users to do some extra work, such as preparing additional http/https servers, etc. We would rather have the karmada-operator read the crd file from the karmada-operator image.

In this way, users only need to use our provided `Web-UI` function to deploy a brand new karmada instance with `one click`. There's no need for extra attention on where the crd is stored.

docs/proposals/operator-custom-crd-download-strategy/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+	// DownloadIfNotPresent instructs the Karmada operator to download the CRDs tarball from a remote location only if it is not yet present in the local cache.
+	DownloadIfNotPresent CRDDownloadPolicy = "IfNotPresent"
+)
+
+// HTTPSource specifies how to download the CRD tarball via either HTTP or HTTPS protocol.
+type HTTPSource struct {
+	// URL specifies the URL of the CRD tarball resource.
+	// +optional
+	URL *string `json:"url,omitempty"`
```
chaunceyjiang, line 71: Final question. Why is it `*string` here instead of `string`?

The `URL` here should be very similar to `RepoURL`.

https://github.com/argoproj/argo-cd/blob/master/pkg/apis/application/v1alpha1/applicationset_types.go#L395

RainbowMango: Yeah, the URL could be required, as if HTTPSource is not nil, the URL must be provided.

jabellard: Will address and push an update

jabellard: Addressed. Can we sign off and merge if there are no other comments? I left the implementation PR in draft state, and plan on finishing it once we're aligned on the design.

chaunceyjiang (commented):
/lgtm

Thanks~

jabellard:
Perfect! I opened the implementation [pr](https://github.com/karmada-io/karmada/pull/5185).

RainbowMango (approved):
/approve
Thanks.

PR #5151, opened 2024-07-07
fix: update scheduler estimator cache when cluster changed

### LINKED ISSUE #5152: update scheduler estimator cache when cluster changed
**What happened**:
Currently, when a cluster update occurs, if the cache already contains a clientWrapper for the corresponding cluster, the cache data will not be updated. 

https://github.com/karmada-io/karmada/blob/c8acebc295e5ae220760abce91b075a4073a94e8/pkg/estimator/client/cache.go#L66
https://github.com/karmada-io/karmada/blob/c8acebc295e5ae220760abce91b075a4073a94e8/pkg/estimator/client/cache.go#L101
https://github.com/karmada-io/karmada/blob/c8acebc295e5ae220760abce91b075a4073a94e8/pkg/scheduler/event_handler.go#L201

**Environment**:
- Karmada version:
- Others:

zhy76:
**What type of PR is this?**
/kind bug

fix: update scheduler estimator cache when cluster changed

Currently, when a cluster update occurs, if the cache already contains a clientWrapper for the corresponding cluster, the cache data will not be updated. This issue needs to be fixed.
Fixes #5152 

```release-note

```

### DIFF
--- a/pkg/estimator/client/cache.go
+++ b/pkg/estimator/client/cache.go
@@ -57,15 +57,13 @@ func (c *SchedulerEstimatorCache) IsEstimatorExist(name string) bool {
 	return exist
 }
 
-// AddCluster adds a grpc connection and associated client into the cache.
-func (c *SchedulerEstimatorCache) AddCluster(name string, connection *grpc.ClientConn, client estimatorservice.EstimatorClient) {
+// AddOrUpdateCluster adds or updates a grpc connection and associated client into the cache.
+func (c *SchedulerEstimatorCache) AddOrUpdateCluster(name string, connection *grpc.ClientConn, client estimatorservice.EstimatorClient) {
 	c.lock.Lock()
 	defer c.lock.Unlock()
-	// If more than one worker have established connections at the same time,
-	// only the first one will be used. The ladder ones would be abandoned.
+
 	if _, exist := c.estimator[name]; exist {
-		_ = connection.Close()
-		return
+		_ = c.estimator[name].connection.Close()
 	}
 	c.estimator[name] = &clientWrapper{
 		connection: connection,
@@ -98,10 +96,6 @@ func (c *SchedulerEstimatorCache) GetClient(name string) (estimatorservice.Estim
 
 // EstablishConnection establishes a new gRPC connection with the specified cluster scheduler estimator.
 func EstablishConnection(kubeClient kubernetes.Interface, name string, estimatorCache *SchedulerEstimatorCache, estimatorServicePrefix string, grpcConfig *grpcconnection.ClientConfig) error {
-	if estimatorCache.IsEstimatorExist(name) {
-		return nil
-	}
-
 	serverAddr, err := resolveCluster(kubeClient, util.NamespaceKarmadaSystem,
 		names.GenerateEstimatorServiceName(estimatorServicePrefix, name), int32(grpcConfig.TargetPort))
 	if err != nil {
@@ -115,7 +109,7 @@ func EstablishConnection(kubeClient kubernetes.Interface, name string, estimator
 		return err
 	}
 	c := estimatorservice.NewEstimatorClient(cc)
-	estimatorCache.AddCluster(name, cc, c)
+	estimatorCache.AddOrUpdateCluster(name, cc, c)
 	klog.Infof("Connection with estimator server(%s) of cluster(%s) has been established.", serverAddr, name)
 	return nil
 }
--- a/pkg/descheduler/descheduler_test.go
+++ b/pkg/descheduler/descheduler_test.go
@@ -505,7 +505,7 @@ func TestDescheduler_worker(t *testing.T) {
 					mock.MatchedBy(func(context.Context) bool { return true }),
 					mock.MatchedBy(func(in *pb.UnschedulableReplicasRequest) bool { return in.Cluster == cluster.Name }),
 				).Return(mockResultFn, nil)
-				desched.schedulerEstimatorCache.AddCluster(cluster.Name, nil, mockClient)
+				desched.schedulerEstimatorCache.AddOrUpdateCluster(cluster.Name, nil, mockClient)
 			}
 
 			desched.informerFactory.Start(ctx.Done())


pkg/estimator/client/cache.go
```diff
@@ -57,16 +57,10 @@ func (c *SchedulerEstimatorCache) IsEstimatorExist(name string) bool {
 	return exist
 }
 
-// AddCluster adds a grpc connection and associated client into the cache.
-func (c *SchedulerEstimatorCache) AddCluster(name string, connection *grpc.ClientConn, client estimatorservice.EstimatorClient) {
+// AddUpdateCluster adds or updates a grpc connection and associated client into the cache.
+func (c *SchedulerEstimatorCache) AddUpdateCluster(name string, connection *grpc.ClientConn, client estimatorservice.EstimatorClient) {
```
zhzhuang-zju, line 61:
```suggestion
func (c *SchedulerEstimatorCache) AddOrUpdateCluster(name string, connection *grpc.ClientConn, client estimatorservice.EstimatorClient) {
```

pkg/estimator/client/cache.go
```diff
@@ -98,8 +92,14 @@ func (c *SchedulerEstimatorCache) GetClient(name string) (estimatorservice.Estim
...
 func EstablishConnection(kubeClient kubernetes.Interface, name string, estimatorCache *SchedulerEstimatorCache, estimatorServicePrefix string, grpcConfig *grpcconnection.ClientConfig) error {
+	// check if the connection already exists, if so, close it and establish a new connection
 	if estimatorCache.IsEstimatorExist(name) {
-		return nil
+		// close it and establish a new connection
+		err := estimatorCache.estimator[name].connection.Close()
+		if err != nil {
+			klog.Errorf("Failed to close connection of cluster(%s): %v.", name, err)
+			return err
+		}
```
zhzhuang-zju, line 102: There may be concurrency issues here. If more than one worker have established connections at the same time, a connection may be closed multiple times.

pkg/estimator/client/cache.go
zhy76: fixed,Thanks

pkg/estimator/client/cache.go
zhy76: fixed

pkg/estimator/client/cache.go
```diff
@@ -47,26 +47,13 @@ func NewSchedulerEstimatorCache() *SchedulerEstimatorCache {
 type clientWrapper struct {
 	connection *grpc.ClientConn
 	client     estimatorservice.EstimatorClient
+	closeLock  sync.Mutex
```
whitewindmills, line 50: why add this lock? it seems that all operations are protected by this write lock `c.lock`.

pkg/estimator/client/cache.go
```diff
@@ -98,9 +89,22 @@ func (c *SchedulerEstimatorCache) GetClient(name string) (estimatorservice.Estim
...
 // EstablishConnection establishes a new gRPC connection with the specified cluster scheduler estimator.
 func EstablishConnection(kubeClient kubernetes.Interface, name string, estimatorCache *SchedulerEstimatorCache, estimatorServicePrefix string, grpcConfig *grpcconnection.ClientConfig) error {
-	if estimatorCache.IsEstimatorExist(name) {
-		return nil
+	// check if the connection already exists, if so, close it and establish a new connection
+	estimatorCache.lock.Lock()
+	es, exist := estimatorCache.estimator[name]
+	if exist {
+		// Use a lock to prevent if more than one worker close connection at the same time,
+		// a connection may be closed multiple times.
```
whitewindmills, line 97: is it possible that multiple workers can acquire this lock `estimatorCache.lock` at the same time?

whitewindmills: BTW, I think this close function `es.connection.Close()` is thread safe.

pkg/estimator/client/cache.go
zhzhuang-zju: agree, Continuing to use lock in estimatorCache is sufficient.
Would it be better to modify it like this?
```go
// EstablishConnection establishes a new gRPC connection with the specified cluster scheduler estimator.
func EstablishConnection(kubeClient kubernetes.Interface, name string, estimatorCache *SchedulerEstimatorCache, estimatorServicePrefix string, grpcConfig *grpcconnection.ClientConfig) error {
	serverAddr, err := resolveCluster(kubeClient, util.NamespaceKarmadaSystem,
		names.GenerateEstimatorServiceName(estimatorServicePrefix, name), int32(grpcConfig.TargetPort))
	if err != nil {
		return err
	}

	klog.Infof("Start dialing estimator server(%s) of cluster(%s).", serverAddr, name)
	cc, err := grpcConfig.DialWithTimeOut(serverAddr, 5*time.Second)
	if err != nil {
		klog.Errorf("Failed to dial cluster(%s): %v.", name, err)
		return err
	}
	c := estimatorservice.NewEstimatorClient(cc)
	estimatorCache.AddOrUpdateCluster(name, cc, c)
	klog.Infof("Connection with estimator server(%s) of cluster(%s) has been established.", serverAddr, name)
	return nil
}

// AddOrUpdateCluster adds or updates a grpc connection and associated client into the cache.
func (c *SchedulerEstimatorCache) AddOrUpdateCluster(name string, connection *grpc.ClientConn, client estimatorservice.EstimatorClient) {
	c.lock.Lock()
	defer c.lock.Unlock()

	if _, exist := c.estimator[name]; exist {
		_ = c.estimator[name].connection.Close()
	}
	c.estimator[name] = &clientWrapper{
		connection: connection,
		client:     client,
	}
}
```

zhy76: make sense，Thanks

pkg/estimator/client/cache.go
zhzhuang-zju: > I think this close function es.connection.Close() is thread safe.

It is indeed thread-safe, but it may cause some unused connections to not be closed properly.

pkg/estimator/client/cache.go
```diff
@@ -49,23 +49,13 @@ type clientWrapper struct {
 	client     estimatorservice.EstimatorClient
 }
 
-// IsEstimatorExist checks whether the cluster estimator exists in the cache.
-func (c *SchedulerEstimatorCache) IsEstimatorExist(name string) bool {
-	c.lock.RLock()
-	defer c.lock.RUnlock()
-	_, exist := c.estimator[name]
-	return exist
-}
```
zhzhuang-zju, line 58: There is no need to remove this method. It is an exported function and may be used elsewhere.

zhy76: done,Thanks~

pkg/estimator/client/cache.go
```diff
@@ -98,10 +96,6 @@ func (c *SchedulerEstimatorCache) GetClient(name string) (estimatorservice.Estim
 
 // EstablishConnection establishes a new gRPC connection with the specified cluster scheduler estimator.
 func EstablishConnection(kubeClient kubernetes.Interface, name string, estimatorCache *SchedulerEstimatorCache, estimatorServicePrefix string, grpcConfig *grpcconnection.ClientConfig) error {
-	if estimatorCache.IsEstimatorExist(name) {
-		return nil
-	}
-
```
whitewindmills, line 104: if the cluster state changes frequently, I am worried that this modification will cause the connection to be updated too frequently. that would be bad.

pkg/estimator/client/cache.go
```diff
@@ -57,15 +57,13 @@ func (c *SchedulerEstimatorCache) IsEstimatorExist(name string) bool {
...
-// AddCluster adds a grpc connection and associated client into the cache.
-func (c *SchedulerEstimatorCache) AddCluster(name string, connection *grpc.ClientConn, client estimatorservice.EstimatorClient) {
+// AddOrUpdateCluster adds or updates a grpc connection and associated client into the cache.
+func (c *SchedulerEstimatorCache) AddOrUpdateCluster(name string, connection *grpc.ClientConn, client estimatorservice.EstimatorClient) {
 	c.lock.Lock()
 	defer c.lock.Unlock()
-	// If more than one worker have established connections at the same time,
-	// only the first one will be used. The ladder ones would be abandoned.
+
 	if _, exist := c.estimator[name]; exist {
```
whitewindmills, line 65: suggest:
```go
if oldConn, exist := c.estimator[name]; exist {
    oldConn.Close()
}
```

whitewindmills:
invite @Garrybest to take a look.

pkg/estimator/client/cache.go
zhzhuang-zju: > is it possible that multiple workers can acquire this lock `estimatorCache.lock` at the same time?

`EstablishConnection` is uesd as a part of `ReconcileFunc` of scheduler/descheduler.schedulerEstimatorWorker, and has the ability to run concurrently.

pkg/estimator/client/cache.go
zhzhuang-zju: agree, updating too often is also a problem, unless it has to be updated.
@zhy76 Can you remind me what the problem is if the scheduler estimator cache is not updated when the cluster changes? Does it have to be updated every time? Or is it possible to update only part of it? For example, does the connection need to be updated?

zhy76: I've reconsidered, and it seems there is no impact. It looks like the author watches for updates only to ensure that all clusters are added to the cache. Only uses the cluster name to find the corresponding estimator server's service and has no other usage. This PR may no longer be necessary. Thank you all for your help. Ask @Garrybest for check and final decision.

Garrybest: Exactly, the cluster name is used to connect the corresponding estimator service in host cluster. I think there is no such circumstance to update this service, right?

Meanwhile, a cluster status could change frequently, so too much unnecessary connection establishment could be expensive.

zhy76: Thanks for your reply~

PR #5158, opened 2024-07-09
docs(proposal): specify replicas strategy

CharlesQQ:
**What type of PR is this?**

/kind design
/kind documentation

Add a new scheduling strategy

Fixes #

```
NONE

```

### DIFF
--- a/docs/proposals/scheduling/specify-replicas/specify-replicas.md
+++ b/docs/proposals/scheduling/specify-replicas/specify-replicas.md
@@ -1,0 +1,602 @@
+---
+title: Specify replicas to schedule
+
+authors:
+- "@CharlesQQ"
+
+reviewers:
+- "@robot"
+- TBD
+approvers:
+- "@robot"
+- TBD
+
+creation-date: 2024-07-08
+
+# Specify replicas to schedule
+
+## Summary
+<!--
+一种新的调度策略及其扩展API,允许调度器按照用户指定的副本数进行调度
+-->
+
+A new scheduling strategy and its extension API allow the scheduler to schedule according to the number of replicas specified by the user
+## Motivation
+<!--
+在某些场景下,用户希望服务的副本数分布是自己直接指定的结果,而不是根据权重间接计算得出的,这样更加直观
+-->
+In some scenarios, users hope that the distribution of the number of replicas in a service can be directly specified by themselves, rather than calculated indirectly based on weights, which is more intuitive.
+### Goals
+<!--
+- 扩展API,允许按照用户使用新的调度策略并指定副本数
+- 调度器支持指定副本数调度的机制
+-->
+- Extended API to allow using new scheduling strategies and specifying the number of replicas by user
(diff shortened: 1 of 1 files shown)


docs/proposals/scheduling/specify-replicas/specify-replicas.md
```diff
@@ -0,0 +1,602 @@
...
+指定每个集群的副本数场景
+当deployment第一次接入Karmada的时候, 用户希望服务的副本数分布不变, 例如:
+迁移之前的服务实数分布是: `sum=7, cluster1=2,cluster2=5`
+迁移之后的服务实数分布依然是: `sum=7, cluster1=2,cluster2=5`
+-->
+
+Specify the number of replicas per cluster scenario
+When the deployment connects to Karmada for the first time, the user hopes that the distribution of the number of replicas in the service will remain unchanged, for example:
+The service real number distribution before migration is: `sum=7, cluster1=2,cluster2=5`
+The service real number distribution after migration is still: `sum=7, cluster1=2,cluster2=5`
```
XiShanYongYe-Chang, line 54: At the beginning of integrating the application with Karmada, can we use OverridePolicy to make differentiated configurations?

docs/proposals/scheduling/specify-replicas/specify-replicas.md
```diff
@@ -0,0 +1,602 @@
...
+### Notes/Constraints/Caveats (Optional)
+
+### Risks and Mitigations
+
+#### Risk
+<!--
+在Divided场景下,指定的总副本数之和可能和deployment.spec.replicas不相等
+-->
+
+In the Divided scenario, the sum of the specified total number of replicas may not be equal to deployment.spec.replicas
```
XiShanYongYe-Chang, line 111: When would such a situation occur?

docs/proposals/scheduling/specify-replicas/specify-replicas.md
CharlesQQ: > At the beginning of integrating the application with Karmada, can we use OverridePolicy to make differentiated configurations?

Use overridepolicy to modify member cluster replicas. After the next release, remove the modification of the number of replicas. The cluster topology will still change, which will cause trouble to users.

docs/proposals/scheduling/specify-replicas/specify-replicas.md
CharlesQQ: > When would such a situation occur?

The problem may lie in the upper-layer API, so that the total number of replicas specified in PP is different from deployment.spec.replicas

Monokaix:
How about adding a new extension point at `assignReplica` and implenment it in a plugin, which is more clear and avoid a heavy framework.

CharlesQQ:
> How about adding a new extension point at assignReplica and implenment it in a plugin, which is more clear and avoid a heavy framework.

How about adding a new proposal for this design?

Issue #4870, opened 2024-04-25
[Umbrella] Align generation and observedGeneration fields by reflectstatus and aggregatestatus operation


yike21:
**What would you like to be added**:
There are some third-party resources and native resources that need to align `.metadata.generation`  and `.status.observedGeneration ` fields. Please ref to #4866 .

**Why is this needed**:

Continue the work of #4867  and break down the tasks to be done.
- Native resource
  - [x] GroupVersion: apps/v1
    Kind: 
      - [x] Deployment @veophi #4867 
      - [x] DaemonSet @whitewindmills #5165 
      - [x] StatefulSet @liangyuanpeng #5204
- Thirdparty resource
  - [x] GroupVersion: apps.kruise.io/v1alpha1
    Kind: 
      - [x] CloneSet @veophi #5057
      - [x] DaemonSet @whitewindmills #5167 
  - [x] GroupVersion: apps.kruise.io/v1beta1
    Kind: 
      - [x] StatefulSet @zhy76 #5094
  - [x] GroupVersion: helm.toolkit.fluxcd.io/v2beta1
    Kind: 
      - [x] HelmRelease @yike21 #5311 
  - [x] GroupVersion: kustomize.toolkit.fluxcd.io/v1
    Kind: 
      - [x] Kustomization @yike21 #5084 
  - [x] GroupVersion: source.toolkit.fluxcd.io/v1
    Kind: 
      - [x] GitRepository @yike21 #5086 
  - [x] GroupVersion: source.toolkit.fluxcd.io/v1beta2
    Kind: 
     - [x] Bucket @yike21  #5193 
     - [x] HelmChart @yike21 #5194
     - [x] HelmRepository @yike21 #5196
     - [x] OCIRepository @yike21 #5197

We need to examine the meaning of `.metadata.generation` and `.status.observedGeneration` fields for each resource above, and then modify the `ReflectStatus` and `AggregateStatus` operations to align them. The modifications are likely to be similar, so please wait for `CloneSet` and `Deployment` modifications to be completed.

/help

liangyuanpeng:
I want to try it for StatefulSet

yike21:
> I want to try it for StatefulSet

Thanks a lot~

XiShanYongYe-Chang:
@yike21 Good Job!

whitewindmills:
I wanna pick up `DaemonSet`, thanks.

yike21:
[...]

PR #5165, opened 2024-07-10
Align federated DaemonSet's observedGeneration semantics with its native

whitewindmills:
**What type of PR is this?**
/kind feature

part of #4870

```release-note
`karmada-controller-manager`: Mark `.status.observedGeneration` of DaemonSet with `.metadata.generation` only when all members' statuses are algined with its resource template generation.
```

### DIFF
--- a/pkg/resourceinterpreter/default/native/aggregatestatus.go
+++ b/pkg/resourceinterpreter/default/native/aggregatestatus.go
@@ -26,6 +26,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	networkingv1 "k8s.io/api/networking/v1"
 	policyv1 "k8s.io/api/policy/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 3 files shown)


pkg/resourceinterpreter/default/native/aggregatestatus.go
```diff
@@ -279,39 +280,44 @@ func aggregateDaemonSetStatus(object *unstructured.Unstructured, aggregatedStatu
...
 
-	if oldStatus.ObservedGeneration == newStatus.ObservedGeneration &&
-		oldStatus.CurrentNumberScheduled == newStatus.CurrentNumberScheduled &&
-		oldStatus.DesiredNumberScheduled == newStatus.DesiredNumberScheduled &&
-		oldStatus.NumberAvailable == newStatus.NumberAvailable &&
-		oldStatus.NumberMisscheduled == newStatus.NumberMisscheduled &&
-		oldStatus.NumberReady == newStatus.NumberReady &&
-		oldStatus.UpdatedNumberScheduled == newStatus.UpdatedNumberScheduled &&
-		oldStatus.NumberUnavailable == newStatus.NumberUnavailable {
+	if equality.Semantic.DeepEqual(oldStatus, newStatus) {
```
XiShanYongYe-Chang, line 320: Other fields in the status may affect the current judgment.

XiShanYongYe-Chang (commented):
Thanks~

pkg/resourceinterpreter/default/native/aggregatestatus.go
whitewindmills: other fields should be empty.

pkg/resourceinterpreter/default/native/status_type.go
```diff
@@ -18,13 +18,24 @@ package native
 
 import appsv1 "k8s.io/api/apps/v1"
 
-// WrappedDeploymentStatus is a wrapper for appsv1.DeploymentStatus.
-type WrappedDeploymentStatus struct {
-	appsv1.DeploymentStatus `json:",inline"`
-
+// FederatedGeneration holds the generation of the same resource in the member cluster and the Karmada control plane.
+type FederatedGeneration struct {
```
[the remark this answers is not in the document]
whitewindmills: I think it does have a federal meaning, which includes member's generation and host's generation.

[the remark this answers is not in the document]
XiShanYongYe-Chang: okay

pkg/resourceinterpreter/default/native/aggregatestatus.go
XiShanYongYe-Chang: If other fields are added in the future, such as condition, the judgment here will be invalid.

whitewindmills: I don't think there will be this problem. even if a new field is added, DaemonSet of Karmada control plane will have this field at the end, and this judgment is still legal.

XiShanYongYe-Chang: Does that mean that other types of resources can be treated the same way?

whitewindmills: yes, but I only modify this because of a lint error: cyclomatic complexity too high :joy:

XiShanYongYe-Chang: okay
/lgtm
/cc @yike21 @veophi

yike21: thanks!
/lgtm

XiShanYongYe-Chang:
Thanks @yike21 
/approve

Issue #5166, opened 2024-07-10
Define metrices port name in container field of related karmada components


chaosi-zju:
**What would you like to be added**:

Define metrices port name in container field of related karmada components, so that the port name can be specified in [PodMonitor](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/user-guides/getting-started.md#using-podmonitors) to form prothumes metrics dashboard.

**Why is this needed**:

Detail info refers to #4877

**TODO list:**

* Helm installation:

- [x] Expose the default port  for the karmada-controller-manager, scheduler and agent when creating a PodMonitor. #5139
- [x] Expose the metrics port  for the karmada-webhook、karmada-operator、descheduler and scheduler-estimator in helm #5168

* karmadactl installation:
- [x] Expose the metrics port  for the karmada-controller-manager, scheduler、agent、karmada-webhook、descheduler and scheduler-estimator in karmadactl init #5169

* karmada-operator installation:
- [x] Expose the metrics port  for the karmada-controller-manager, scheduler、karmada-webhook、descheduler in karmada-operator #5465

* local-up installation:
- [x] Expose the metrics port  for the karmada-controller-manager, scheduler、agent、karmada-webhook、descheduler and scheduler-estimator in local-up-karmada. #5428

whitewindmills:
[Helm]
> Expose the metrics port for the karmada-webhook、karmada-operator、descheduler and scheduler-estimator in helm

[karmadactl]
>  Expose the metrics port for the karmada-controller-manager, scheduler、agent、karmada-webhook、karmada-operator、descheduler and scheduler-estimator in karmadactl init

/assign

RainbowMango:
/close
As all tasks are done. Thank you guys who contributed to this effort. @chaosi-zju @dzcvxe @whitewindmills @wangxf1987

PR #5169, opened 2024-07-10
Expose metrics port for karmadactl installation

whitewindmills:
**What type of PR is this?**
/kind cleanup

part of https://github.com/karmada-io/karmada/issues/5166

```release-note
`karmadactl`: Expose the metrics port for PodMonitor.
```

### DIFF
--- a/pkg/karmadactl/addons/descheduler/manifests.go
+++ b/pkg/karmadactl/addons/descheduler/manifests.go
@@ -46,6 +46,7 @@ spec:
             - /bin/karmada-descheduler
             - --kubeconfig=/etc/kubeconfig
             - --bind-address=0.0.0.0
+            - --secure-port=10358
             - --leader-elect-resource-namespace={{ .Namespace }}
             - --scheduler-estimator-ca-file=/etc/karmada/pki/ca.crt
             - --scheduler-estimator-cert-file=/etc/karmada/pki/karmada.crt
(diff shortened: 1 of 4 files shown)


liangyuanpeng (commented):
suggest add a release-note for it,maybe:

```
karmadactl: exporting the metrics port for PodMonitor.
```

RainbowMango:
Either have a release note or not, both OK to me.

pkg/karmadactl/addons/descheduler/manifests.go
```diff
@@ -60,6 +60,10 @@ spec:
             initialDelaySeconds: 15
             periodSeconds: 15
             timeoutSeconds: 5
+          ports:
+            - containerPort: 10358
+              name: metrics
+              protocol: TCP
```
chaosi-zju, line 66: Since the value `10358` is hard-coded here, I suggest adding `--secure-port` param to the `command` field as well, just like:

```yaml
- --bind-address=0.0.0.0
- --secure-port=10358
```

otherwise, supposing I defined `--secure-port` to other port or someone updated `defaultPort` in code, easily missed changes lead to inconsistency

maybe  other components ditto.

chaosi-zju: because now other guys may don't known here 10358 is just the value of `--secure-port`

RainbowMango (approved):
/approve
Thanks.

PR #5174, opened 2024-07-11
Add missing command line arguments for karmada-operator

whitewindmills:
**What type of PR is this?**
/kind cleanup

Fixes #

```release-note
`karmada-operator`: Introduced `--metrics-bind-address` and `--health-probe-bind-address` flags, it's a compatible change as the default value does not change from previous versions. 
```

### DIFF
--- a/operator/cmd/operator/app/options/options.go
+++ b/operator/cmd/operator/app/options/options.go
@@ -39,11 +39,6 @@ type Options struct {
 	Controllers []string
 	// LeaderElection defines the configuration of leader election client.
 	LeaderElection componentbaseconfig.LeaderElectionConfiguration
-	// BindAddress is the IP address on which to listen for the --secure-port port.
-	BindAddress string
-	// SecurePort is the port that the the server serves at.
-	// Note: We hope support https in the future once controller-runtime provides the functionality.
-	SecurePort int
 	// KubeAPIQPS is the QPS to use while talking with karmada-apiserver.
 	KubeAPIQPS float32
 	// KubeAPIBurst is the burst to allow while talking with karmada-apiserver.
@@ -56,6 +51,10 @@ type Options struct {
 	// It can be set to "0" to disable the metrics serving.
 	// Defaults to ":8080".
 	MetricsBindAddress string
+	// HealthProbeBindAddress is the TCP address that the controller should bind to
+	// for serving health probes
+	// It can be set to "0" or "" to disable serving the health probe.
+	HealthProbeBindAddress string
 	// ConcurrentKarmadaSyncs is the number of karmada objects that are allowed to sync concurrently.
 	ConcurrentKarmadaSyncs int
 }
@@ -73,8 +72,6 @@ func NewOptions() *Options {
 			ResourceNamespace: "karmada-system",
 			ResourceName:      "karmada-operator",
 		},
-		BindAddress:            "0.0.0.0",
-		SecurePort:             8443,
 		KubeAPIQPS:             50,
 		KubeAPIBurst:           100,
 		ConcurrentKarmadaSyncs: 5,
@@ -84,6 +81,8 @@ func NewOptions() *Options {
 
 // AddFlags adds flags to the specified FlagSet.
 func (o *Options) AddFlags(fs *pflag.FlagSet, allControllers []string, disabledByDefaultControllers []string) {
+	fs.StringVar(&o.MetricsBindAddress, "metrics-bind-address", ":8080", "The TCP address that the controller should bind to for serving prometheus metrics(e.g. 127.0.0.1:8080, :8080). It can be set to \"0\" to disable the metrics serving.")
+	fs.StringVar(&o.HealthProbeBindAddress, "health-probe-bind-address", ":8443", "The TCP address that the controller should bind to for serving health probes(e.g. 127.0.0.1:8443, :8443). It can be set to \"0\" or \"\" to disable serving the health probe.")
 	fs.DurationVar(&o.ResyncPeriod.Duration, "resync-period", o.ResyncPeriod.Duration, "ResyncPeriod determines the minimum frequency at which watched resources are reconciled.")
 	fs.Float32Var(&o.KubeAPIQPS, "kube-api-qps", o.KubeAPIQPS, "QPS to use while talking with kubernetes apiserver.")
 	fs.Int32Var(&o.KubeAPIBurst, "kube-api-burst", o.KubeAPIBurst, "Burst to use while talking with kubernetes apiserver.")
--- a/operator/cmd/operator/app/operator.go
+++ b/operator/cmd/operator/app/operator.go
@@ -20,9 +20,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
-	"net"
 	"os"
-	"strconv"
 
 	"github.com/spf13/cobra"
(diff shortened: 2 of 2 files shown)


whitewindmills:
PTAL
/assign @chaosi-zju @RainbowMango

chaosi-zju:
looks good

operator/cmd/operator/app/options/options.go
```diff
@@ -84,6 +87,9 @@ func NewOptions() *Options {
 
 // AddFlags adds flags to the specified FlagSet.
 func (o *Options) AddFlags(fs *pflag.FlagSet, allControllers []string, disabledByDefaultControllers []string) {
+	fs.StringVar(&o.BindAddress, "bind-address", defaultBindAddress, "The IP address on which to listen for the --secure-port port.")
+	fs.IntVar(&o.SecurePort, "secure-port", defaultPort, "The port that the the health server serves at.")
```
RainbowMango, line 91: We don't support TLS verification yet when serving health probe. So this is not a `secure port`.

Do you think we need to introduce port for health probe and metrics separately?

By the way, the `--secure-port` has the same issue.

whitewindmills: > We don't support TLS verification yet when serving health probe. So this is not a secure port.

how about `admin port`?

> Do you think we need to introduce port for health probe and metrics separately?

whether or not to separate them, I'm ok. but if we name this port `admin port`, I think it's unnecessary to separate them.
if we reach a consensus, there are other components that also have this problem, and we will also need other PRs to solve it.

RainbowMango: @zhzhuang-zju Please look at this issue. My concern is that the `--secure-port` isn't a real `secure` one.

> if we reach a consensus, there are other components that also have this problem, and we will also need other PRs to solve it.

Yes, exactly.

zhzhuang-zju: > My concern is that the --secure-port isn't a real secure one.

`--secure-port` isn't a real secure one. refer to https://github.com/karmada-io/karmada/blob/a2e782836d51189242bc1336fa3393adcc0adaa7/operator/cmd/operator/app/options/options.go#L45, there are plans to upgrade to https.

> Do you think we need to introduce port for health probe and metrics separately?

`MetricsBindAddress` contains the port, so there's no need to introduce another port.

> how about admin port?

`admin port` or `sever port` are ok to me

> there are other components that also have this problem, and we will also need other PRs to solve it.

If we want to reach a consensus between different components, we need to sort out the port naming of each component before determining the appropriate portname to avoid conflicts.

whitewindmills: here are the ports used by all the components I organized

- karmada-operator:
  * `--secure-port` & `--bind-address`: only for http health probe
  * `--metrics-bind-address`: only for metrics
- karmada-agent:
  * `--secure-port` & `--bind-address`: only for http health probe
  * `--metrics-bind-address`: only for metrics
- karmada-controller-manager:
  * `--secure-port` & `--bind-address`: only for http health probe
  * `--metrics-bind-address`: only for metrics
- karmada-descheduler:
  * `--secure-port` & `--bind-address`: for http health probe and metrics
- karmada-scheduler:
  * `--secure-port` & `--bind-address`: for http health probe and metrics
- karmada-scheduler-estimator:
  * `--secure-port` & `--bind-address`: for http health probe and metrics
  * `--server-port`: grpc server port
- karmada-webhook:
  * `--secure-port` & `--bind-address`: only for webhook server
  * `--metrics-bind-address`: only for metrics
  * `--health-probe-bind-address`: only for http health probe

zhzhuang-zju: From the perspective of consensus among all components, it might be better to separate `health probes` and `metrics`.
- Facilitates subsequent implementation of tls certificate authentication 
  From a data perspective, data from `health probe` is less valuable to do encryption than `metrics`; From the perspective of call frequency, `health probe` is called more frequently and the certificate authentication performance overhead will be higher; Also, the karmada components are mostly `liveness-http` services, so if configure mutual TLS authentication and the `kubelet` doesn't have a certificate issued by karmada, the health check might fail. Refer to https://github.com/kubernetes-sigs/controller-runtime/pull/2407,  controller-runtim is now be possible to serve metrics via tls. So the two have different requirements for tls certificate authentication, in order to facilitate the subsequent implementation of tls certificate authentication, separating the two makes it easier to utilize the extension.
- Consistency with `controller-runtime`
`helath probe` and `metrics` are carried in two separate fields in the controller-runtime's Options, and are not necessarily related to each other.

What do you think? @whitewindmills @RainbowMango

RainbowMango: I agree with this. 
But another thing I'm not sure about is how what the separated flags look like.  
1. Should we separate the IP address and port?
  * Yes, it might look like:  `--metrics-bind-address` and `--metrics-port`.
  * No, it might look like `--metrics-bind-address`, such as `x.x.x.x:8080`
2. Can the health probes and metrics share the same address and IP?

zhzhuang-zju: >1. Should we separate the IP address and port?
 
I prefer not to separate ip and port to reduce the number of parameter options to be configured.

> Can the health probes and metrics share the same address and IP?

When choosing to separate `health probes` and `metrics`, we are providing the ability to configure them with different address and ip, which is fine with me if the user configures both as the same.

RainbowMango: > which is fine with me if the user configures both as the same.

I mean it might be impossible to bind the same port for two HTTPServers...

zhzhuang-zju: What you concerns are correct. I tried to change the metrics port of `controller-manage` to match the health probe, and then `controller-manager` reported an error
```bash
E0717 07:02:07.663335       1 controllermanager.go:199] controller manager exits unexpectedly: failed to start metrics server: failed to create listener: listen tcp :10357: bind: address already in use
```

whitewindmills: > What you concerns are correct. I tried to change the metrics port of controller-manage to match the health probe, and then controller-manager reported an error

yes, controller-manager will create two http server for them.

> I mean it might be impossible to bind the same port for two HTTPServers...

usually we don't need to bind the same port for the two http server. and in the production environment, we choose to turn off the metrics .

> No, it might look like --metrics-bind-address, such as x.x.x.x:8080

I vote for this. we just need to be consistent with controller-manager.

zhzhuang-zju: > I vote for this. we just need to be consistent with controller-manager.

agree

RainbowMango: +1

zhzhuang-zju: Looks like we're in agreement :sun_with_face:
@whitewindmills Would you be interested in combing through this task and submitting an issue？

whitewindmills: SURE, will do that.

RainbowMango (commented):
/approve

Luckily, we haven't exposed the two flags yet, so we can skip the deprecation period.

Please @zhzhuang-zju help to take another look.

zhzhuang-zju:
thanks~
/lgtm

PR #5180, opened 2024-07-11
Automated cherry pick of #5144: GroupClusters should sort by score and availableReplica count

mszacillo:
Cherry pick of #5144 into release-1.10. 

With this small change, we now sort clusters by taking their availableReplicas into account, rather than defaulting to the cluster's name. This helps spread resource usage across all clusters and would be helpful to use as part of the current v1.10 release.


RainbowMango (approved):
/approve
Thanks.

PR #5185, opened 2024-07-12
Custom CRD Download Strategy Support for Karmada Operator

### LINKED ISSUE #5122: Add the Ability to Provide a Custom CRD download URL for Karmada Operator

As a prerequisite for provisioning new Karmada instances, the Karmada operator needs to ensure required CRDs are downloaded.
By default, it will attempt to download them from a [release artifact](https://github.com/karmada-io/karmada/releases) on GitHub.
This feature request is for having the ability to specify a custom URL that should be used to download the CRDs tarball.

**Why is this needed**:
At Bloomberg, we're currently building a managed Karmada platform and want to use the Karmada operator to manage the entire lifecycle of managed Karmada instances. By default,  services running in our internal network can't access resources on the internet. Whenever possible, we have a strong preference for  hydrating external resources to locations that are accessible within our internal network, and would like to follow that approach for downloading the CRDs.

jabellard:
**What type of PR is this?**

/kind api-change

This is an implementation of the design outlined in this [proposal](https://github.com/karmada-io/karmada/pull/5145).

Fixes #5122

```release-note
`karmada-operator`: Introduced CRD download strategy that allows downloading CRD from a private source.
```


RainbowMango:
@chaunceyjiang would you like to take a look? to evaluate if there is any effect on your deployment.

operator/pkg/tasks/init/crd.go
```diff
@@ -156,3 +180,17 @@ func existCrdsTar(crdsDir string) bool {
...
 	return false
 }
+
+func getCrdsDir(data InitData) (string, error) {
+	crdTarball := data.CrdTarball()
+	key := ""
+	if crdTarball.HTTPSource != nil {
+		key = crdTarball.HTTPSource.URL
+	} else {
+		return "", errors.New("[getCrdsDir] No CRD download strategy has been specified")
```
RainbowMango, line 190: The error message might be misleading with the download policy. How about that:

```suggestion
		return "", errors.New("[getCrdsDir] No CRD download source has been specified")
```
But this shouldn't happen, right? as we have default HTTP source.

operator/pkg/tasks/init/crd.go
```diff
@@ -92,31 +108,37 @@ func runCrdsDownload(r workflow.RunData) error {
...
+	}
+
+	crdTarball := data.CrdTarball()
+	if crdTarball.HTTPSource != nil {
+		if err := util.DownloadFile(crdTarball.HTTPSource.URL, crdsTarPath); err != nil {
+			return fmt.Errorf("failed to download CRD tar, err: %w", err)
 		}
 	} else {
-		klog.V(2).InfoS("[download-crds] The crd tar exists on disk", "path", crdsDir, "karmada", klog.KObj(data))
+		return errors.New("[download-crds] No CRD download strategy has been specified")
```
RainbowMango, line 138: ditto.

RainbowMango (commented):
Generally looks good to me, thanks to @jabellard.

operator/pkg/tasks/init/crd.go
jabellard: Good point. Pushed change to address.

RainbowMango (commented):
/lgtm

Let's wait @chaunceyjiang for a while.

operator/pkg/tasks/init/crd.go
```diff
@@ -73,7 +80,17 @@ func skipCrdsDownload(r workflow.RunData) (bool, error) {
 		return false, errors.New("prepare-crds task invoked with an invalid data struct")
 	}
 
-	crdsDir := path.Join(data.DataDir(), data.KarmadaVersion())
+	crdTarball := data.CrdTarball()
+	if crdTarball.CRDDownloadPolicy != nil && *crdTarball.CRDDownloadPolicy == operatorv1alpha1.DownloadAlways {
+		klog.V(2).InfoS("[skipCrdsDownload] CrdDownloadPolicy is 'Always', skipping cache check")
+		return false, nil
+	}
+
```
chaunceyjiang, line 88: 1. This code is not compatible with the previous behavior.

2. I believe it is necessary to be compatible with the previous behavior, as this will facilitate user upgrades.

Therefore, I suggest that this code needs to check whether the path `path.Join(data.DataDir(), data.KarmadaVersion())` exists. If it does, it should also skip the CRD download."

chaunceyjiang: As I mentioned in your proposal earlier, we have already stored the CRD files in this path in advance. We don't want the karmada-operator to download any files from the internet.

RainbowMango: I understand your concern is about changing the cache directory, before this feature, the tarball is cached at `data.DataDir()/<KarmadaVersion>`, and after this feature, the path becomes `data.DataDir()/cache/<hashkey>`. 

Since you embed the tarball into the image, after this change you have to adopt the new path, which probably needs to make some changes to your Dockerfile. Am I right?

chaunceyjiang: >  which probably needs to make some changes to your Dockerfile

Yes, I need to modify the Dockerfile. But I am more concerned about: I am not clear about the value of `hashkey` in the path `data.DataDir()/cache/<hashkey>`. It seems that every Karmada version needs to guess what this value is.

jabellard: The approach you're using for pre-loading the CRD tarball can still work. For instance,  the CRD tarball for version `v1.10.3`, by default, would be downloaded via this URL:  `https://github.com/karmada-io/karmada/releases/download/v1.10.3/crds.tar.gz`. To preload, just store it at location `data.DataDir()/cache/<cacheKey>`,  where `cacheKey` is simply the SHA-256 hash of that URL. It's deterministic.

RainbowMango: On the one hand, the cache maintained by the karmada-operator doesn't guarantee compatibility over versions, as during the upgrading, the cache will be dropped along with the new container. That means we can change the cache directory freely without concern about compatibility. 
@chaunceyjiang In your case, you don't want the karmada-operator to download anything from a remote server, I think we can consider enhancing it by providing a solution to support that case.

On the other hand, we can revisit the reason why introduce the hash-key to the cache directory.

Echo from the proposal [Custom CRD Download Strategy Support for Karmada Operator](https://github.com/karmada-io/karmada/tree/master/docs/proposals/operator-custom-crd-download-strategy):
> Currently, the cache for previously downloaded CRDs lives on the local file system at /var/lib/karmada/<operator-release-version>. To support custom CRD download strategies, we need to a new directory structure for the cache. The Karmada operator can potentially be managing the lifecycle of multiple Karmada instances of heterogeneous versions, and as such, may have to download CRD tarballs from multiple sources. We need to cache the CRDs for each source. Given some source, it's cache entry will live on the local file system at path /var/lib/karmada/cache/<cache-key>. The cache key will be the computed SHA-256 hash of some property derived from that source. In the case of HTTPSource, it will be the hash of the URL.

The reason why introducing the hash-key is to identify different source, my question is do we really need to do that? As the tarball should be consistent even it comes from different sources. Is there some thing I don't get here?

jabellard: I think this caching strategy works just fine and follows an approach similar to how OCI images are pulled from registries. 

With this feature, although the directory structure will change to accommodate the use of custom download strategies, the use case for pre-populating the CRD in the filesystem can still easily be supported.

However, given that this design is extensible, I do think we can provide a more natural solution for this use case. Today, an `HTTPSource` is supported, but, for instance, it's possible to also support a `LocalFileSystemSource` that may help solve this problem more naturally, but I think that's out of scope for this PR.

chaunceyjiang: https://github.com/calvin0327/karmada/blob/master/cluster/images/operator-offline.Dockerfile#L22-L27

This is the Dockerfile we are using. How do I calculate the `cacheKey`? Can I use `sha256sum`?

jabellard: Yes.

jabellard: @chaunceyjiang , any other comments/concerns?

chaunceyjiang: Thank you for your reply.

RainbowMango (approved):
It seems we have reached an agreement!
Great thanks to @jabellard for the excellent job!

/approve

PR #5188, opened 2024-07-13
Fix the issue of residual work in the MultiClusterService feature.

XiShanYongYe-Chang:
**What type of PR is this?**

/kind bug

The finalizer `MCSEndpointSliceDispatchControllerFinalizer` was added too late, and when the dispatched work is still in the process of creating, deleting the `MultiClusterService` object may result in residual dispatched work resources.

Fixes #

```release-note
karmada-controller-manager: fix the issue of residual work in the MultiClusterService feature.
```


chaosi-zju:
So, before this fix introduced, if the deletion of `eps` happens to occur within the short limit time after the creation of `Work` and before the creation of `Finalizer`, it will lead to the problem of residual work, which is guaranteed after this modification. 👍

RainbowMango (approved):
/approve
It makes sense to append the finalizer during born.

PR #5193, opened 2024-07-13
Align federated buckets.source.toolkit.fluxcd.io observedGeneration semantics with its native

yike21:
**What type of PR is this?**
/kind feature

part of #4870

```release-note
`karmada-controller-manager`: Mark `.status.observedGeneration` of buckets.source.toolkit.fluxcd.io with `.metadata.generation` only when all members' statuses are algined with its resource template generation.
```


XiShanYongYe-Chang (commented):
Thanks a lot~
/lgtm
/approve

PR #5215, opened 2024-07-17
Adding generic handling of priorityclass + namespace for flinkdeployment interpreter

mszacillo:
**What type of PR is this?**
/kind feature

The default interpreter for FlinkDeployment does not currently account for namespace or priorityclass. This was done because in our use-case, the priorityclass is shared amongst tenants and can be set statically. But after thinking again, it makes more sense to have this interpretation be generic.

This is a quick PR that adds that capability. 

```release-note
Added generic handling of priorityclass and namespace for default flinkdeployment interpreter
```


RainbowMango (commented):
/lgtm
/assign @yike21 
to take another look and get approval.

yike21 (commented):
Thanks for your contribution!
/approve

Issue #5219, opened 2024-07-18
[Umbrella] Specification of merics and health probe port parameters


whitewindmills:
**What would you like to be added**:

Currently, the port parameter definitions of the metrics and health probes are different for each component, which will cause some ambiguity to users, so we plan to standardize the port definitions of each component.

- karmada-operator:
  * `--secure-port` & `--bind-address`: only for http health probe
  * `--metrics-bind-address`: only for metrics
- karmada-agent:
  * `--secure-port` & `--bind-address`: only for http health probe
  * `--metrics-bind-address`: only for metrics
- karmada-controller-manager:
  * `--secure-port` & `--bind-address`: only for http health probe
  * `--metrics-bind-address`: only for metrics
- karmada-descheduler:
  * `--secure-port` & `--bind-address`: for http health probe and metrics
- karmada-scheduler:
  * `--secure-port` & `--bind-address`: for http health probe and metrics
- karmada-scheduler-estimator:
  * `--secure-port` & `--bind-address`: for http health probe and metrics
  * `--server-port`: grpc server port
- karmada-webhook:
  * `--secure-port` & `--bind-address`: only for webhook server
  * `--metrics-bind-address`: only for metrics
  * `--health-probe-bind-address`: only for http health probe
 
This will be basically the same as controller-manager:
```go
// MetricsBindAddress is the TCP address that the controller should bind to
// for serving prometheus metrics.
// It can be set to "0" to disable the metrics serving.
MetricsBindAddress string

// HealthProbeBindAddress is the TCP address that the controller should bind to
// for serving health probes
// It can be set to "0" to disable serving the health probe.
HealthProbeBindAddress string
```

**Note that you need to mark the old parameters as deprecated and we will remove them in a later version.**

- [x] karmada-operator: https://github.com/karmada-io/karmada/pull/5174
- [x] karmada-agent: https://github.com/karmada-io/karmada/pull/5223
- [x] karmada-controller-manager: https://github.com/karmada-io/karmada/pull/5290
- [x] karmada-descheduler: https://github.com/karmada-io/karmada/pull/5435
- [x] karmada-scheduler: @liangyuanpeng https://github.com/karmada-io/karmada/pull/5437
- [x] karmada-scheduler-estimator: https://github.com/karmada-io/karmada/pull/5273

RainbowMango:
Great thanks to @whitewindmills for picking this. 
[...]

PR #5223, opened 2024-07-18
Standardize the health probe arguments

whitewindmills:
**What type of PR is this?**
/kind deprecation
/kind feature
/kind cleanup

part of https://github.com/karmada-io/karmada/issues/5219

```release-note
`karmada-agent`:  Add health probe argument `health-probe-bind-address`, deprecate `--bind-address` and `--secure-port` flag.
```

### DIFF
--- a/cmd/agent/app/agent.go
+++ b/cmd/agent/app/agent.go
@@ -20,8 +20,6 @@ import (
 	"context"
 	"flag"
 	"fmt"
-	"net"
-	"strconv"
 
 	"github.com/spf13/cobra"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -80,6 +78,10 @@ func NewAgentCommand(ctx context.Context) *cobra.Command {
 plane and sync manifests from the Karmada control plane to the member cluster. In addition, it also syncs the status of member
 cluster and manifests to the Karmada control plane.`,
 		RunE: func(_ *cobra.Command, _ []string) error {
+			// complete options
+			if err := opts.Complete(); err != nil {
+				return err
+			}
 			// validate options
 			if errs := opts.Validate(); len(errs) != 0 {
 				return errs.ToAggregate()
@@ -205,7 +207,7 @@ func run(ctx context.Context, opts *options.Options) error {
 		LeaseDuration:              &opts.LeaderElection.LeaseDuration.Duration,
 		RenewDeadline:              &opts.LeaderElection.RenewDeadline.Duration,
 		RetryPeriod:                &opts.LeaderElection.RetryPeriod.Duration,
-		HealthProbeBindAddress:     net.JoinHostPort(opts.BindAddress, strconv.Itoa(opts.SecurePort)),
+		HealthProbeBindAddress:     opts.HealthProbeBindAddress,
 		LivenessEndpointName:       "/healthz",
 		Metrics:                    metricsserver.Options{BindAddress: opts.MetricsBindAddress},
 		MapperProvider:             restmapper.MapperProvider,
--- a/cmd/agent/app/options/options.go
+++ b/cmd/agent/app/options/options.go
@@ -18,6 +18,8 @@ package options
 
 import (
 	"fmt"
+	"net"
+	"strconv"
 	"strings"
 	"time"
 
@@ -51,9 +53,11 @@ type Options struct {
 	Controllers    []string
 	LeaderElection componentbaseconfig.LeaderElectionConfiguration
 	// BindAddress is the IP address on which to listen for the --secure-port port.
+	// Deprecated: Use HealthProbeBindAddress instead. And will be removed in release 1.12+.
 	BindAddress string
 	// SecurePort is the port that the the server serves at.
 	// Note: We hope support https in the future once controller-runtime provides the functionality.
+	// Deprecated: Use HealthProbeBindAddress instead. And will be removed in release 1.12+.
 	SecurePort        int
 	KarmadaKubeConfig string
 	// ClusterContext is the name of the cluster context in control plane KUBECONFIG file.
@@ -104,6 +108,11 @@ type Options struct {
 	// It can be set to "0" to disable the metrics serving.
 	// Defaults to ":8080".
 	MetricsBindAddress string
+	// HealthProbeBindAddress is the TCP address that the controller should bind to
+	// for serving health probes
+	// It can be set to "0" to disable serving the health probe.
+	// Defaults to ":10357".
+	HealthProbeBindAddress string
 
 	RateLimiterOpts ratelimiterflag.Options
 
@@ -168,6 +177,10 @@ func (o *Options) AddFlags(fs *pflag.FlagSet, allControllers []string) {
 		"The IP address on which to listen for the --secure-port port.")
 	fs.IntVar(&o.SecurePort, "secure-port", defaultPort,
 		"The secure port on which to serve HTTPS.")
+	// nolint: errcheck
+	fs.MarkDeprecated("bind-address", "This flag is deprecated and will be removed in release 1.12+. Use --health-probe-bind-address instead.")
+	// nolint: errcheck
+	fs.MarkDeprecated("secure-port", "This flag is deprecated and will be removed in release 1.12+. Use --health-probe-bind-address instead.")
 	fs.BoolVar(&o.LeaderElection.LeaderElect, "leader-elect", true, "Start a leader election client and gain leadership before executing the main loop. Enable this when running replicated components for high availability.")
 	fs.StringVar(&o.LeaderElection.ResourceNamespace, "leader-elect-resource-namespace", util.NamespaceKarmadaSystem, "The namespace of resource object that is used for locking during leader election.")
 	fs.DurationVar(&o.LeaderElection.LeaseDuration.Duration, "leader-elect-lease-duration", defaultElectionLeaseDuration.Duration, ""+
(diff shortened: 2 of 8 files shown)


cmd/agent/app/options/options.go
```diff
@@ -206,6 +219,7 @@ func (o *Options) AddFlags(fs *pflag.FlagSet, allControllers []string) {
 	fs.IntVar(&o.ConcurrentWorkSyncs, "concurrent-work-syncs", 5, "The number of Works that are allowed to sync concurrently.")
 	fs.StringSliceVar(&o.ReportSecrets, "report-secrets", []string{"KubeCredentials", "KubeImpersonator"}, "The secrets that are allowed to be reported to the Karmada control plane during registering. Valid values are 'KubeCredentials', 'KubeImpersonator' and 'None'. e.g 'KubeCredentials,KubeImpersonator' or 'None'.")
 	fs.StringVar(&o.MetricsBindAddress, "metrics-bind-address", ":8080", "The TCP address that the controller should bind to for serving prometheus metrics(e.g. 127.0.0.1:8080, :8080). It can be set to \"0\" to disable the metrics serving.")
+	fs.StringVar(&o.HealthProbeBindAddress, "health-probe-bind-address", ":0", "The TCP address that the controller should bind to for serving health probes(e.g. 127.0.0.1:10357, :10357). It can be set to \"0\" or \"\" to disable serving the health probe.")
```
zhzhuang-zju, line 222:
```suggestion
	fs.StringVar(&o.HealthProbeBindAddress, "health-probe-bind-address", "0.0.0.0:10357", "The TCP address that the controller should bind to for serving health probes(e.g. 127.0.0.1:10357, :10357). It can be set to \"0\" or \"\" to disable serving the health probe.")
```

Default values should remain the same as before

cmd/agent/app/options/options.go
```diff
@@ -219,3 +233,23 @@ func (o *Options) AddFlags(fs *pflag.FlagSet, allControllers []string) {
 	features.FeatureGate.AddFlag(fs)
 	o.ProfileOpts.AddFlags(fs)
 }
+
+// Complete ensures that options are valid and marshals them if necessary.
+func (o *Options) Complete() error {
+	if o.BindAddress == "" && o.SecurePort == 0 {
+		return nil
+	}
+	if o.HealthProbeBindAddress != ":0" {
```
zhzhuang-zju, line 242:
```suggestion
	if o.HealthProbeBindAddress != "0.0.0.0:10357" {
```

cmd/agent/app/options/options.go
whitewindmills: we cannot set it to `0.0.0.0:10357`, cause `--bind-address` and `--secure-port` can not co-exist with `--health-probe-bind-address`. if one set flags like:
```shell
/bin/karmada-agent ----bind-address=0.0.0.0 --secure-port=8443 --health-probe-bind-address=0.0.0.0:10357
```
it will be validate.

zhzhuang-zju: If users does not set these arguments before or after the upgrade, then the `health-probe-bind-address` used before the upgrade is `0.0.0.0:10357` and after the upgrade becomes `:0`, is this inconsistency acceptable?

whitewindmills: fixing it :heart:

whitewindmills:
@RainbowMango @zhzhuang-zju  PTAL

cmd/agent/app/options/options.go
RainbowMango, line 222:
```suggestion
	fs.StringVar(&o.HealthProbeBindAddress, "health-probe-bind-address", "", "The TCP address that the controller should bind to for serving health probes(e.g. 127.0.0.1:10357, :10357). It can be set to \"0\" or \"\" to disable serving the health probe. Defaults to 0.0.0.0:10357.")
```
We can put the default to the usage in this release. so that we can figure out if the user starts to use this new flag in `complete` phase.

RainbowMango: I'm afraid a dynamic port is not acceptable, otherwise, we can't configure the [liveness probe](https://github.com/karmada-io/karmada/blob/bf2a6044a0de3a6219933dbd029e005cfaf7569c/artifacts/agent/karmada-agent.yaml#L40).

cmd/agent/app/options/options.go
```diff
@@ -219,3 +233,23 @@ func (o *Options) AddFlags(fs *pflag.FlagSet, allControllers []string) {
 	features.FeatureGate.AddFlag(fs)
 	o.ProfileOpts.AddFlags(fs)
 }
+
+// Complete ensures that options are valid and marshals them if necessary.
+func (o *Options) Complete() error {
```
RainbowMango, line 238:
```golang
func (o *Options) Complete() error {
	if len(o.HealthProbeBindAddress) == 0 {
		o.HealthProbeBindAddress = net.JoinHostPort(o.BindAddress, strconv.Itoa(o.SecurePort))
	}
	return nil
}
```
All we need to do is migrate the deprecated flag values to the new one, and only if the new one is not in use.

cmd/agent/app/options/options.go
```diff
@@ -51,9 +54,11 @@ type Options struct {
 	Controllers    []string
 	LeaderElection componentbaseconfig.LeaderElectionConfiguration
 	// BindAddress is the IP address on which to listen for the --secure-port port.
+	// Deprecated
```
RainbowMango, line 57:
```suggestion
	// Deprecated: Use HealthProbeBindAddress instead. And will be removed in release 1.12+.
```

cmd/agent/app/options/options.go
```diff
@@ -51,9 +54,11 @@ type Options struct {
...
 	BindAddress string
 	// SecurePort is the port that the the server serves at.
 	// Note: We hope support https in the future once controller-runtime provides the functionality.
+	// Deprecated
```
RainbowMango, line 61: Ditto.

cmd/agent/app/options/options.go
```diff
@@ -164,10 +173,14 @@ func (o *Options) AddFlags(fs *pflag.FlagSet, allControllers []string) {
 		"A list of controllers to enable. '*' enables all on-by-default controllers, 'foo' enables the controller named 'foo', '-foo' disables the controller named 'foo'. All controllers: %s.",
 		strings.Join(allControllers, ", "),
 	))
-	fs.StringVar(&o.BindAddress, "bind-address", defaultBindAddress,
+	fs.StringVar(&o.BindAddress, "bind-address", o.BindAddress,
 		"The IP address on which to listen for the --secure-port port.")
-	fs.IntVar(&o.SecurePort, "secure-port", defaultPort,
+	fs.IntVar(&o.SecurePort, "secure-port", o.SecurePort,
```
RainbowMango, line 178: We don't need to change the default value here. As the old default will be transmit to new default.

zhzhuang-zju: Would it be better to put the default value in usage as well as `o.HealthProbeBindAddress`?  So that we can figure out if the user still use the old flag in `complete` phase.

cmd/agent/app/agent.go
```diff
@@ -205,7 +207,7 @@ func run(ctx context.Context, opts *options.Options) error {
 		LeaseDuration:              &opts.LeaderElection.LeaseDuration.Duration,
 		RenewDeadline:              &opts.LeaderElection.RenewDeadline.Duration,
 		RetryPeriod:                &opts.LeaderElection.RetryPeriod.Duration,
-		HealthProbeBindAddress:     net.JoinHostPort(opts.BindAddress, strconv.Itoa(opts.SecurePort)),
```
chaunceyjiang, line 208: https://github.com/karmada-io/karmada/blob/master/artifacts/agent/karmada-agent.yaml#L34

Should we also make some changes here?

zhzhuang-zju: > Should we also make some changes here?

Yes, considering that the old flags will be removed when upgrading from `1.11` to `1.12`, replacing the old flags with the new ones at `1.11` is a good choice

cmd/agent/app/options/options.go
liangyuanpeng: So this may be a breaking change, if people specify `BindAddress` and `SecurePort` different from the defaults in 1.10, when upgrading to 1.11, unexpected things will happen.
This is uncommon but could happen.

whitewindmills: what will happen?

liangyuanpeng: set SecurePort to 10086 in 1.10, and then upgrade to 1.11, karmada agent is using `0.0.0.0:10357` from `HealthProbeBindAddress`, so it will change the user’s intent

whitewindmills: BindAddress and SecurePort still wok if HealthProbeBindAddress is unspecified.

liangyuanpeng: > fs.StringVar(&o.HealthProbeBindAddress, "health-probe-bind-address", "", "The TCP address that the controller should bind to for serving health probes(e.g. 127.0.0.1:10357, :10357). It can be set to \"0\" to disable serving the health probe. Defaults to 0.0.0.0:10357.")

Yes, you're right, I was misled by  `Defaults to 0.0.0.0:10357.`.

cmd/agent/app/options/options.go
```diff
@@ -219,3 +233,11 @@ func (o *Options) AddFlags(fs *pflag.FlagSet, allControllers []string) {
 	features.FeatureGate.AddFlag(fs)
 	o.ProfileOpts.AddFlags(fs)
 }
+
+// Complete ensures that options are valid and marshals them if necessary.
+func (o *Options) Complete() error {
+	if len(o.HealthProbeBindAddress) == 0 {
+		o.HealthProbeBindAddress = net.JoinHostPort(o.BindAddress, strconv.Itoa(o.SecurePort))
```
zhzhuang-zju, line 240: Just one queston, can 'HealthProbeBindAddress' be configured at the same time as 'BindAddress' and 'SecurePort'?

whitewindmills: yes, but only `HealthProbeBindAddress` works.

zhzhuang-zju: The advantage of this approach is that it facilitates code implementation, as there is no need to check for the co-existence of deprecated flags 'BindAddress' and 'SecurePort' with the replacement flag 'HealthProbeBindAddress'. 
However, if a user sets 'BindAddress' and 'SecurePort' that conflict with the 'HealthProbeBindAddress', it may lead to confusion for the user.

I currently cannot determine which solution is better, cc @RainbowMango @chaunceyjiang @liangyuanpeng for further advice~

liangyuanpeng: Can we at least add a warning log?

whitewindmills: I've deprecated `BindAddress` and `SecurePort`, you'll see such warning log: `This flag is deprecated and will be removed in release 1.12+. Use --health-probe-bind-address instead.` I think it's enough.

zhzhuang-zju: Adding warning log is a good idea. Can we add a description, for example "If both `BindAddress` or `SecurePort` and `health-probe-bind-address` are set, the system will prioritize and utilize the value of `health-probe-bind-address`."

whitewindmills: yeah, we can do that.

RainbowMango: > I've deprecated BindAddress and SecurePort, you'll see such warning log: This flag is deprecated and will be removed in release 1.12+. Use --health-probe-bind-address instead. I think it's enough.

+1 on this. The deprecation warning log is enough.

RainbowMango (approved):
/lgtm
/approve

PR #5227, opened 2024-07-19
fix the error of cluster status old condition update will overwrite the newest condition

XiShanYongYe-Chang:
**What type of PR is this?**

/kind bug
/kind failing-test

During the cluster status update, due to the presence of an old Condition in the previous Cluster status, this Condition is incorrectly updated to the Cluster object during the status refresh. The expected Cluster status conditions should no longer include the old condition.

Fixes #

```release-note
karmada-controller-manager: fix the error of cluster status old condition update will overwrite the newest condition
```

### DIFF
--- a/pkg/controllers/status/cluster_status_controller.go
+++ b/pkg/controllers/status/cluster_status_controller.go
@@ -153,7 +153,11 @@ func (c *ClusterStatusController) Reconcile(ctx context.Context, req controllerr
 		return controllerruntime.Result{Requeue: true}, nil
 	}
 
-	return c.syncClusterStatus(ctx, cluster)
+	err := c.syncClusterStatus(ctx, cluster)
+	if err != nil {
+		return controllerruntime.Result{}, err
+	}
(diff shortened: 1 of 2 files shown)


pkg/controllers/status/cluster_status_controller.go
```diff
@@ -153,7 +153,11 @@ func (c *ClusterStatusController) Reconcile(ctx context.Context, req controllerr
 		return controllerruntime.Result{Requeue: true}, nil
 	}
 
-	return c.syncClusterStatus(ctx, cluster)
+	err := c.syncClusterStatus(ctx, cluster)
+	if err != nil {
+		return controllerruntime.Result{}, err
+	}
+	return controllerruntime.Result{RequeueAfter: c.ClusterStatusUpdateFrequency.Duration}, nil
```
RainbowMango, line 160: Is this a bug that the status controller didn't report the status as per `ClusterStatusUpdateFrequency.Duration`?

RainbowMango: No,  i see. ignore me.

RainbowMango (approved):
/lgtm
/approve

Issue #5231, opened 2024-07-22
Mutating+Validating Webhook names are too generic


grosser:
**What would you like to be added**:
rename MutatingWebhookConfiguration mutating-config -> mutations.karmada.io

**Why is this needed**:
the current name is too generic for a cluster-level resource and is unclear what project it belongs to at first look + could easily overwritten if another project also uses this generic name

whitewindmills:
feel free to fix it.

XiShanYongYe-Chang:
+1

grosser:
https://github.com/karmada-io/karmada/pull/5240

PR #5240, opened 2024-07-23
rename webhooks to be clearly identifiable as from karmada

grosser:
**What type of PR is this?**

/kind cleanup
/kind deprecation

Fixes https://github.com/karmada-io/karmada/issues/5231

I used `app: karmada-webhook` since that is where the requests are sent to, made more sense to me then declaring 2 apps that only have the hooks

```release-note
- after deploying delete the mutating-config MutatingWebhookConfiguration and validating-config ValidatingWebhookConfiguration (they were renamed)
```

TODO:
- [ ] merge label change and rebase
- [ ] add helm clenup job

### DIFF
--- a/artifacts/deploy/webhook-configuration.yaml
+++ b/artifacts/deploy/webhook-configuration.yaml
@@ -1,9 +1,9 @@
 apiVersion: admissionregistration.k8s.io/v1
 kind: MutatingWebhookConfiguration
 metadata:
-  name: mutating-config
+  name: karmada-mutations
   labels:
-    app: mutating-config
+    app: karmada-webhook
 webhooks:
   - name: propagationpolicy.karmada.io
     rules:
@@ -121,9 +121,9 @@ webhooks:
 apiVersion: admissionregistration.k8s.io/v1
 kind: ValidatingWebhookConfiguration
 metadata:
-  name: validating-config
+  name: karmada-validations
   labels:
-    app: validating-config
+    app: karmada-webhook
 webhooks:
   - name: propagationpolicy.karmada.io
     rules:
--- a/charts/karmada/templates/_karmada_webhook_configuration.tpl
+++ b/charts/karmada/templates/_karmada_webhook_configuration.tpl
@@ -5,9 +5,9 @@
 apiVersion: admissionregistration.k8s.io/v1
 kind: MutatingWebhookConfiguration
 metadata:
-  name: mutating-config
+  name: karmada-mutations
   labels:
-    app: mutating-config
+    app: karmada-webhook
     {{- include "karmada.commonLabels" . | nindent 4 }}
 webhooks:
   - name: propagationpolicy.karmada.io
@@ -126,9 +126,9 @@ webhooks:
 apiVersion: admissionregistration.k8s.io/v1
 kind: ValidatingWebhookConfiguration
 metadata:
-  name: validating-config
+  name: karmada-validations
   labels:
-    app: validating-config
+    app: karmada-webhook
     {{- include "karmada.commonLabels" . | nindent 4 }}
 webhooks:
   - name: propagationpolicy.karmada.io
--- a/operator/pkg/karmadaresource/webhookconfiguration/manifests.go
+++ b/operator/pkg/karmadaresource/webhookconfiguration/manifests.go
@@ -22,9 +22,9 @@ const (
 apiVersion: admissionregistration.k8s.io/v1
 kind: MutatingWebhookConfiguration
 metadata:
-  name: mutating-config
+  name: karmada-mutations
   labels:
-    app: mutating-config
+    app: karmada-webhook
 webhooks:
   - name: propagationpolicy.karmada.io
     rules:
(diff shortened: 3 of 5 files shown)


XiShanYongYe-Chang (commented):
Thanks @grosser 

For the upgrade scenario, such as upgrading from version 1.10 to 1.11, how should users operate? Or can it be done seamlessly without user awareness?

The release note seems not correct.

grosser:
afaik helm does not delete/rename so users have to delete the old hook to avoid the hook getting called twice

XiShanYongYe-Chang:
Is it possible to make the user insensible?

And are other installations affected?

grosser:
- we could leave webhooks with the old name that do nothing, so if a user does not see the update instructions there would be no issue
- we could add a helm pre-upgrade hook, idk if that would work for all users + would be lots of effort to validate + need to figure out rbac

XiShanYongYe-Chang:
Ask more guys to help take a look.
cc @chaunceyjiang @whitewindmills @chaosi-zju

whitewindmills:
we need to delete MutatingWebhookConfiguration/ValidatingWebhookConfiguration with the old name after the upgrade. and this cleanup is temporary and will be removed in the release v1.12+.

XiShanYongYe-Chang:
> we need to delete MutatingWebhookConfiguration/ValidatingWebhookConfiguration with the old name after the upgrade.

Is the delete operation performed by the user?

whitewindmills:
> Is the delete operation performed by the user?

no, it should be executed by karmada-operator/helm/karmadactl.

XiShanYongYe-Chang:
Hi @grosser Can you help us evaluate the feasibility and follow-up plan?

grosser:
I'm not deep enough in this projects deploy workflow to know how this would work.
Sounds like there is a desire to fix this via helm, but I think this would not work for us since we just render helm via kustomize and then apply. My initial idea was that users would have to `kubectl delete` after upgrading, but I'm not sure what happens if 2 webhooks are running at the same time, I assume they would not step over each other and just work.

whitewindmills:
> but I'm not sure what happens if 2 webhooks are running at the same time, I assume they would not step over each other and just work.

we cannot guarantee that all webhooks are idempotent. this may cause some unexpected errors. so we need to first figure out how to delete MutatingWebhookConfiguration/ValidatingWebhookConfiguration with the old name through each installation method(karmada-operator/helm/karmadactl).

chaunceyjiang:
> we cannot guarantee that all webhooks are idempotent. this may cause some unexpected errors. so we need to first figure out how to delete MutatingWebhookConfiguration/ValidatingWebhookConfiguration with the old name through each installation method(karmada-operator/helm/karmadactl).

+1

grosser:
So maybe best to add noop webhooks with the old names to make this safe, then the helm upgrade process is not that important and even if it goes wrong or someone does not use helm they can safely upgrade.

artifacts/deploy/webhook-configuration.yaml
```diff
@@ -1,9 +1,9 @@
 apiVersion: admissionregistration.k8s.io/v1
 kind: MutatingWebhookConfiguration
 metadata:
-  name: mutating-config
+  name: karmada-mutations
   labels:
-    app: mutating-config
```
chaunceyjiang, line 6: Can we modify the labels first?

Based on the description in the issue, the user is unable to distinguish which project the webhook belongs to. If we already have a keyword like `karmada` in the labels, I think the user would be able to distinguish them.

chaunceyjiang: image

I checked the `Validatingwebhookconfigurations` in one of my clusters. They can indeed be distinguished by their names which project they belong to. 😂

whitewindmills:
> So maybe best to add noop webhooks with the old names to make this safe

this may not be a good approach, we will eventually delete them anyway.

artifacts/deploy/webhook-configuration.yaml
grosser: https://github.com/karmada-io/karmada/pull/5246

grosser:
yeah it's not great, but it makes the transition easier,
otherwise you have a a time when there are either no webhooks or 2 webhooks
if we leave a noop copy then users have time to verify manually after upgrading
(we can still do the whole try to delete automatically thing)

whitewindmills:
> otherwise you have a a time when there are either no webhooks or 2 webhooks

during the upgrade process, this is no big deal.

* for helm installation, we can add a job to complete this cleanup work.
* for karmada-operator/karmadactl, we can delete MutatingWebhookConfiguration/ValidatingWebhookConfiguration with the old name in the code.

this cleanup is temporary and will be removed in the release v1.12+. @grosser can you break it down into multiple tasks to track?

grosser:
added todo list to PR if that is what you meant

XiShanYongYe-Chang:
For users who use binary or image upgrade, is it only possible to handle it manually? Is there any specific operation guide?

Aslo invite @RainbowMango to evaluate the upgrade impact.
/cc @RainbowMango

RainbowMango:
First, let's revisit the origin issue, according to #5231, what we want to solve is the configuration file naming problem, not a functional issue. We have already mitigated its impact in #5246. The next step is to handle the upgrade concerns properly.

For that, we need a clear plan to move forward, including the following:
- upgrade guidance
- consider the difference between all kind of installation tools(karmadactl init/operator/helm) in the guidance

Any thoughts? And anyone willing to take on this task?

grosser:
I think it's not worth the effort 😞

PR #5251, opened 2024-07-25
Add failover history information

Dyex719:
**What type of PR is this?**
/kind feature

Adding failover history information so that applications can keep a record of what failovers happened in the past.
Stateful applications can use this information to detect failures and continue processing from a particular state

Fixes #5116 #4969  

```release-note
NONE
```


pkg/controllers/utils/common_test.go
```diff
@@ -0,0 +1,90 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
+	"github.com/karmada-io/karmada/pkg/util/gclient"
+)
+
```
Dyex719, line 29: This test fails with:
```
E0722 13:25:50.730922   76848 common.go:27] Invalid failover type passed into updateFailoverStatus
--- FAIL: TestUpdateFailoverStatus (0.00s)
    --- FAIL: TestUpdateFailoverStatus/application_failover (0.00s)
        common_test.go:69: 
                Error Trace:    /Users/adityaaddepalli/Documents/karmada/pkg/controllers/utils/common_test.go:69
                Error:          Received unexpected error:
                                resourcebindings.work.karmada.io "binding" not found
                Test:           TestUpdateFailoverStatus/application_failover
    --- FAIL: TestUpdateFailoverStatus/cluster_failover (0.00s)
        common_test.go:69: 
                Error Trace:    /Users/adityaaddepalli/Documents/karmada/pkg/controllers/utils/common_test.go:69
                Error:          Received unexpected error:
                                resourcebindings.work.karmada.io "binding" not found
                Test:           TestUpdateFailoverStatus/cluster_failover
FAIL
FAIL    github.com/karmada-io/karmada/pkg/controllers/utils     3.257s
```
I think this is due to the mocked client unable to perform actions on the resource binding.
 
UpdateFailoverStatus calls [UpdateStatus](https://github.com/karmada-io/karmada/blob/bf2a6044a0de3a6219933dbd029e005cfaf7569c/pkg/util/helper/status.go#L39) which performs a get and mutate on the resouce. I think this error is coming from the get call in that function.

Is there is a way to get around this, I don't see test functions for other methods like [AggregateResourceBindingWorkStatus](https://github.com/karmada-io/karmada/blob/bf2a6044a0de3a6219933dbd029e005cfaf7569c/pkg/util/helper/workstatus.go#L58) in the respective [test file ](https://github.com/karmada-io/karmada/blob/bf2a6044a0de3a6219933dbd029e005cfaf7569c/pkg/util/helper/workstatus_test.go)either 

Would I need to write my own on mocked UpdateStatus for this?

Dyex719:
Hi All,
I have implemented and tested the failover information feature and wanted to request a review from the community on what changes might be required.
I notice that there are some minor linter issues that I will fix.
I have also addressed the failing test case in the comment above.

Thank you!

pkg/controllers/binding/common.go
```diff
@@ -154,9 +154,43 @@ func mergeTargetClusters(targetClusters []workv1alpha2.TargetCluster, requiredBy
 	return targetClusters
 }
 
+func checkFailoverCondition(resourceBinding *workv1alpha2.ResourceBinding) string {
```
Dyex719, line 157: Rename - we don't use condition for checking

pkg/controllers/utils/common.go
```diff
@@ -0,0 +1,117 @@
...
+			// set binding status with the newest condition
+			currentTime := metav1.Now()
+			failoverHistoryItem := workv1alpha2.FailoverHistoryItem{
+				FailoverTime:  &currentTime,
+				OriginCluster: cluster,
+				Reason:        reason,
+			}
+			binding.Status.FailoverHistory = append(binding.Status.FailoverHistory, failoverHistoryItem)
+			klog.V(4).Infof("Failover history is %+v", binding.Status.FailoverHistory)
+			existingCondition := meta.FindStatusCondition(binding.Status.Conditions, failoverType)
```
Dyex719, line 98: Conditions are redundant, we can infer everything from the failover history. Will not populate conditions with failover information.

mszacillo:
Hi @RainbowMango ! Could we get a review of this PR when you get a chance? Failures are due to the test case describes above in the comments. We can decide how to tackle this moving forward. Greatly appreciate it!

RainbowMango:
Sure. And sorry for letting this sit again!

pkg/apis/work/v1alpha2/binding_types.go
```diff
@@ -319,6 +319,23 @@ type ResourceBindingStatus struct {
 	// AggregatedStatus represents status list of the resource running in each member cluster.
 	// +optional
 	AggregatedStatus []AggregatedStatusItem `json:"aggregatedStatus,omitempty"`
+
+	// FailoverHistory represents history of the previous failovers of this resource
+	// + optional
```
RainbowMango, line 324:
```suggestion
	// +optional
```

pkg/apis/work/v1alpha2/binding_types.go
```diff
@@ -319,6 +319,23 @@ type ResourceBindingStatus struct {
...
+type FailoverHistoryItem struct {
+	// FailoverTime represents the timestamp when the workload failed over.
+	// It is represented in RFC3339 form(like '2021-04-25T10:02:10Z') and is in UTC.
+	FailoverTime *metav1.Time `json:"failoverTime,omitempty"`
+
+	// OriginCluster denotes the name of the cluster from which the workload was failed over.
+	OriginCluster string `json:"originCluster,omitempty"`
+
+	// Reason denotes the reason why the workload failed over.
+	Reason string `json:"reason,omitempty"`
```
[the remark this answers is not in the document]
RainbowMango: So that each history item can provide enough context for us to figure out where the application was migrated.

In addition, we can also have the preserved state in this history item.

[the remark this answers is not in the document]
RainbowMango: Yeah, you are right. 
In fact, the demo at https://github.com/karmada-io/karmada/pull/5116#issuecomment-2379037424 does not use the history at all.

mszacillo:
Hi @RainbowMango,

I've taken a closer look at your demo branch (https://github.com/karmada-io/karmada/compare/master...XiShanYongYe-Chang:karmada:api_draft_application_failover). I don't have any major complaints from the API side. It seems that logically this effort can be divided into two parts:
1. Adding FailoverHistoryItem to the resource binding API (which is handled in this PR).
2. Adding StatePreservation to the propagationpolicy API (which has been done in your branch). 

Would you be open to dividing up the work as such? The StatePreservation work can be merged after the FailoverHistoryItem is added. I've already updated the FailoverHistoryItem API to align better with the proposed changes in your branch, but I've left the StatePreservation out as Chang seems to have implemented that already and I wouldn't consider it fair to copy his work under our PR. :)

`Note:` I haven't updated the codegen for this PR yet so there are some failing checks. Will get that sorted.

mszacillo:
Follow-up question related to the proposed changes for [state preservation](https://github.com/karmada-io/karmada/compare/master...XiShanYongYe-Chang:karmada:api_draft_application_failover#diff-4a4c5e7654f51fc1c0d813dba54a1ded9145e6860ff989336309c5f537224855R168), are we planning on supporting this only for applications that failover gracefully?

In our use-case we use `Immediately`, since some of our users have strict only-once processing requirements. So when a failover occurs, we want to remove the application as fast as possible to reduce the likelihood of duplicate applications. Perhaps the state preservation could be kept as part of the status of the resourcebinding and referenced during the scheduling step, similar to how we handle the FailoverHistoryItem in our PR.

RainbowMango:
> It seems that logically this effort can be divided into two parts:
Adding FailoverHistoryItem to the resource binding API (which is handled in this PR).
Adding StatePreservation to the propagationpolicy API (which has been done in your branch).

It makes sense to me. 

>  I've already updated the FailoverHistoryItem API to align better with the proposed changes in your branch, but I've left the StatePreservation out as Chang seems to have implemented that already and I wouldn't consider it fair to copy his work under our PR. :)

As I explained at https://github.com/karmada-io/karmada/pull/5116#issuecomment-2418694001, that code in my branch is just a demo for demonstrating the general idea, and that's all based on your idea. So we would be happy if it could be accepted in this proposal.

RainbowMango:
> Follow-up question related to the proposed changes for [state preservation](https://github.com/karmada-io/karmada/compare/master...XiShanYongYe-Chang:karmada:api_draft_application_failover#diff-4a4c5e7654f51fc1c0d813dba54a1ded9145e6860ff989336309c5f537224855R168), are we planning on supporting this only for applications that failover gracefully?

I'm still thinking about this, I believe cluster failover also requires this configuration. 

> In our use-case we use Immediately, since some of our users have strict only-once processing requirements. So when a failover occurs, we want to remove the application as fast as possible to reduce the likelihood of duplicate applications.

Glad to know the user case, when designing this `PurgeMode`, there was a discussion about whether any user would want to remove the old application immediately, now I think this is a clear use case for it.

pkg/apis/work/v1alpha2/binding_types.go
```diff
@@ -362,6 +366,41 @@ type AggregatedStatusItem struct {
...
+	// +required
+	StartTime metav1.Time `json:"failoverTime"`
+
+	// ClustersBeforeFailover records the cluster where the application was running prior to failover.
+	// +required
+	ClusterBeforeFailover string `json:"originalCluster"`
+
+	// ClustersAfterFailover records the cluster where the application is running after failover.
+	// +optional
+	ClusterAfterFailover string `json:"targetCluster,omitempty"`
```
RainbowMango, line 385: I know in your case no matter before or after the failover, there is only one cluster. But I think we can do it commonly, that is record the clusters in a slice.  What do you think?

pkg/apis/work/v1alpha2/well_known_constants.go
```diff
@@ -124,6 +124,12 @@ const (
 	// Additional options will be added here in the future.
 	DeletionProtectionLabelKey = "resourcetemplate.karmada.io/deletion-protected"
 	DeletionProtectionAlways   = "Always"
+
+	// ResourceBindingFailoverLabel If a resource is failed over by karmada, this label will be attached
+	// to the rescheduled workload. The value will denote the type of failover that occurred, either cluster or application.
+	// This can be useful if applications are stateful and need to know when they have been failed over by Karmada,
+	// as opposed to being scheduled fresh.
+	ResourceBindingFailoverLabel = "resourcebinding.karmada.io/failover-type"
```
RainbowMango, line 132: I'm not sure if we need this label yet. 
But given this PR focuses on the failover history, I suggest moving this out of this PR, and letting this PR focus on:
- failover history API design
- the logic of generating the history
- use the history to prevent scheduling failure clusters again

pkg/apis/work/v1alpha2/binding_types.go
mszacillo: I had noticed that the original definition was an array - does that mean we want to support failover for applications that are spread across multiple clusters? Or is the array simply going to contain a running history of the clusters on which is was scheduled before?

For example, let's say application X fails over from Cluster B to Cluster C -> the ClustersBeforeFailover will be: ["B"]. 

If application X later fails over from Cluster C to Cluster A -> the ClustersBeforeFailover will be ["B", "A"]?

pkg/apis/work/v1alpha2/well_known_constants.go
mszacillo: That sounds fair to me. We currently use the label for our e2e failover feature, but I see no problem separating this out.

pkg/apis/work/v1alpha2/binding_types.go
mszacillo: Oops! Sorry

pkg/apis/work/v1alpha2/binding_types.go
RainbowMango: No. 
I just hope that the history information faithfully records the status, it doesn't care about how many clusters was scheduled before or after the failover. 

For your example, application X fails over from Cluster B to Cluster C, and then to Cluster A, then there will be two items on the history:
```yaml
- startTime: "Monday"
  clusterBeforeFailover: [B]
  clusterAfterFailover: [C]
- startTime: "Tuesday"
  clusterBeforeFailover: [C]
  clusterAfterFailover: [A]
```

pkg/apis/work/v1alpha2/well_known_constants.go
RainbowMango: Great!
The reason why I prefer small PRs because they usually focus on a small area and make it easier to make final decisions.

pkg/apis/work/v1alpha2/binding_types.go
mszacillo: Ah understood,  thanks for clarification!

RainbowMango:
/retest
Seems the failing test is unrelated.
https://github.com/karmada-io/karmada/actions/runs/11465282263/job/31903724746?pr=5251
```
STEP: Creating Remedy(remedy-8rs8j) @ 10/22/24 17:50:43.218
  STEP: update Cluster(member1) ServiceDomainNameResolutionReady condition to false @ 10/22/24 17:50:43.225
  STEP: wait Cluster(member1) status has TrafficControl RemedyAction @ 10/22/24 17:50:43.246
  STEP: remove Remedy resource @ 10/22/24 17:50:43.26
  STEP: wait Cluster(member1) status doesn't has TrafficControl RemedyAction @ 10/22/24 17:50:43.266
  [FAILED] in [It] - /home/runner/work/karmada/karmada/test/e2e/framework/cluster.go:308 @ 10/22/24 17:57:43.267
  << Timeline

  [FAILED] Timed out after 420.001s.
  Expected
      <bool>: false
  to equal
      <bool>: true
  In [It] at: /home/runner/work/karmada/karmada/test/e2e/framework/cluster.go:308 @ 10/22/24 17:57:43.267

  Full Stack Trace
    github.com/karmada-io/karmada/test/e2e/framework.WaitClusterFitWith({0x583bfc0, 0xc000c29c20}, {0xc000726fc9, 0x7}, 0x5358650)
    	/home/runner/work/karmada/karmada/test/e2e/framework/cluster.go:308 +0x23e
    github.com/karmada-io/karmada/test/e2e.init.func10.1.3.4()
    	/home/runner/work/karmada/karmada/test/e2e/remedy_test.go:121 +0x78
    github.com/karmada-io/karmada/test/e2e.init.func10.1.3()
    	/home/runner/work/karmada/karmada/test/e2e/remedy_test.go:120 +0x3f1
```

pkg/apis/work/v1alpha2/binding_types.go
RainbowMango, line 385:
```suggestion
	// ClustersBeforeFailover records the cluster where the application was running prior to failover.
	// +required
	ClusterBeforeFailover []string `json:"clusterBeforeFailover"`

	// ClustersAfterFailover records the cluster where the application is running after failover.
	// +required
	ClusterAfterFailover []string `json:"clusterAfterFailover"`
```
That's my thought.

mszacillo: I've updated the type to be `[]TargetCluster` so that it's aligned with the types used in ResourceBinding, just haven't pushed the changes yet. Let me know if that is okay as well -

RainbowMango: We usually use the field name as the json tag.  I mean `originalCluster` and `targetCluster`.

In addition, do you think `ClusterBeforeFailover` and `ClusterAfterFailover` should be `required` filed?
I'm not sure yet, but I think we can go with `required` for now.

mszacillo: Oh, yes this is updated as well! Will reflect once I push my recent changes. :)

pkg/apis/work/v1alpha2/well_known_constants.go
RainbowMango: Are we going to remove this label from this PR?

pkg/apis/work/v1alpha2/binding_types.go
RainbowMango: Thanks, i see.

pkg/apis/work/v1alpha2/well_known_constants.go
mszacillo: Yes, I'll be pulling out the failover flag logic from this PR. :)

pkg/controllers/utils/common.go
```diff
@@ -0,0 +1,94 @@
...
+	if placementPtr == nil {
+		return false
+	}
+	return verifySchedulingTypeSupported(placementPtr.ReplicaScheduling, placementPtr.SpreadConstraints)
+}
+
+// FailoverHistoryInfo is currently not supported for the following scheduling types
+//  1. Duplicated (as these resources do not failover)
+//  2. Divided resources that can be scheduled across multiple clusters. In this case, state is harder to conserve since
+//     the application's replicas will not be migrating together.
```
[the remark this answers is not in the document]
RainbowMango: > But I think the current API definition should be generalizable. 

Yes, exactly.

mszacillo:
Hi @RainbowMango, I updated the FailoverHistory API definition to remove the fromCluster definition as it's a bit redundant. We can use the clustersBeforeFailover field as is. I've also removed the failover label from this PR and will open a follow-up PR with that implementation after this one. 

The failover history should now cover various use-cases - did some e2e tests:

1. With replicas failing over together, from cluster B to A and then from cluster A to B.

```
  failoverHistory:
  - clustersAfterFailover:
    - name: cluster-A
      replicas: 3
    clustersBeforeFailover:
    - name: cluster-B
      replicas: 3
    reason: ApplicationFailure
    startTime: "2024-10-25T22:25:39Z"
  - clustersAfterFailover:
    - name: cluster-B
      replicas: 3
    clustersBeforeFailover:
    - name: cluster-A
      replicas: 3
    reason: ApplicationFailure
    startTime: "2024-10-25T22:26:10Z"
```

2. Divided replicas failing over (reschedule fails since we filter out clusters from before failover, and I was only using 2 member clusters for tests).

```
failoverHistory:
  - clustersAfterFailover: []
    clustersBeforeFailover:
    - name: spaas-ob-qa02
      replicas: 2
    - name: spaas-pw-qa02
      replicas: 1
    reason: ApplicationFailure
    startTime: "2024-10-25T23:19:16Z"
```

mszacillo:
Will need to check the failing e2e test `reschedule when policy is dynamic weight schedule type`, seems the filtering step I added might have broken this.

RainbowMango:
> I updated the FailoverHistory API definition to remove the fromCluster definition as it's a bit redundant. We can use the clustersBeforeFailover field as is.

Fair enough. From `ClustersBeforeFailover` and `ClustersAfterFailover` we can easily figure out the failover direction.

pkg/apis/work/v1alpha2/well_known_constants.go
RainbowMango: I guess we can remove the definition as well from this PR.

RainbowMango:
> Divided replicas failing over (reschedule fails since we filter out clusters from before failover, and I was only using 2 member clusters for tests).

My two cents. I think history should only hold the successful failover. For those failover fails which should not have a substantial impact on the system and also should not be included in the history.  What do you think?

If this is the case, we may need to consider recording the history when cleaning the eviction task. More works to do, such as:
- Ensure all types of failover would start with an eviction task. (No matter what the PurgeMode it is.)
- Record the history only when cleaning the eviction task, and only record the succeed one.

pkg/controllers/cluster/taint_manager.go
```diff
@@ -170,6 +171,10 @@ func (tc *NoExecuteTaintManager) syncBindingEviction(key util.QueueKey) error {
 	// Case 2: Need eviction after toleration time. If time is up, do eviction right now.
 	// Case 3: Tolerate forever, we do nothing.
 	if needEviction || tolerationTime == 0 {
+		err := controllerUtils.UpdateFailoverStatus(tc.Client, binding, workv1alpha2.EvictionReasonTaintUntolerated)
+		if err != nil {
+			klog.Errorf("Failed to update status with failover information. Error: %v", err)
+		}
```
RainbowMango, line 177: Here adds a duplicated history item, doesn't it? We recorded the history at the application controller when triggering the failover.

pkg/scheduler/scheduler.go
RainbowMango, line 1: TBD.
Maybe letting the scheduler take care of the history is not a good idea. Because the whole failover process is handled by controllers.
Anyway, need to further explore.

mszacillo:
> My two cents. I think history should only hold the successful failover. For those failover fails which should not have a substantial impact on the system and also should not be included in the history. What do you think?

I think that's fair, we can apply the full failover history once scheduling has been decided. 

> If this is the case, we may need to consider recording the history when cleaning the eviction task. More works to do, such as:
> * Ensure all types of failover would start with an eviction task. (No matter what the PurgeMode it is.)
> * Record the history only when cleaning the eviction task, and only record the succeed one.

I could use the eviction task, but it would need to become more general to encompass `Immediately` in addition to the already supported `Gracefully` PurgeMode. Given that the eviction tasks themselves are called `GracefulEvictionTasks`, this might need more thought and refactoring which I'd argue is starting to move out of scope of this PR. The other issue is if the gracePeriod is too low, the eviction task will be cleared before scheduling take place, meaning the we won't be able to construct the failoverHIstory. 

What I would suggest is we could use a read-write safe failover cache which stores the ResourceBinding's Name/Namespace` as a key, with the value being `[][]TargetCluster`. How it would work:
- When a failover is triggered, add the ResourceBinding to the cache and append []TargetCluster representing the current evicted clusters. 
- During scheduling, check the cache for the ResourceBinding and take the most recent []TargetCluster to represent the `clustersBeforeFailover`. Clear the entry of the cache if that is the last item in the slice. 

I still have to think over edge cases, but let me know your ideas.

pkg/scheduler/framework/plugins/clustereviction/cluster_eviction.go
```diff
@@ -47,11 +48,30 @@ func (p *ClusterEviction) Name() string {
...
 // Filter checks if the target cluster is in the GracefulEvictionTasks which means it is in the process of eviction.
-func (p *ClusterEviction) Filter(_ context.Context, bindingSpec *workv1alpha2.ResourceBindingSpec, _ *workv1alpha2.ResourceBindingStatus, cluster *clusterv1alpha1.Cluster) *framework.Result {
+func (p *ClusterEviction) Filter(_ context.Context, bindingSpec *workv1alpha2.ResourceBindingSpec, bindingStatus *workv1alpha2.ResourceBindingStatus, cluster *clusterv1alpha1.Cluster) *framework.Result {
+	failoverHistory := bindingStatus.FailoverHistory
+	if len(failoverHistory) != 0 {
+		lastFailover := failoverHistory[len(failoverHistory)-1]
+		if containsCluster(lastFailover.ClustersBeforeFailover, cluster.Name) {
+			klog.V(2).Infof("Workload has been failed over from this cluster %s.", cluster.Name)
+			return framework.NewResult(framework.Unschedulable, fmt.Sprintf("workload has been failed over from this cluster %s", cluster.Name))
+		}
```
RainbowMango, line 58: I know in your scenario an application can only failover from one cluster to another. 
Even though I'm not sure if there is a case that an application is divided into multiple clusters and failover together yet, I think it would be great to support that case here.
In other words, here `failover from` might be not just one cluster. 

See https://github.com/karmada-io/karmada/pull/5251#issuecomment-2439277014.

RainbowMango:
> I could use the eviction task, but it would need to become more general to encompass Immediately in addition to the already supported Gracefully PurgeMode. Given that the eviction tasks themselves are called GracefulEvictionTasks, this might need more thought and refactoring which I'd argue is starting to move out of scope of this PR.

Yes, it is beyond the scope of this PR. I'm still investigating it. 
Invite @XiShanYongYe-Chang here to confirm if there are any other concerns regarding to `Ensure all types of failover would start with an eviction task. (No matter what the PurgeMode it is.)`.

pkg/scheduler/framework/plugins/clustereviction/cluster_eviction.go
mszacillo: Perhaps the log line just needs updating. The clustersBeforeFailover already covers scenarios in which replicas are divided across different clusters. Since clustersBeforeFailure only include clusters that have been marked for eviction, we shouldn’t reschedule replicas to those clusters. 

That’s why I added this filter step

pkg/scheduler/scheduler.go
mszacillo: Hmm but in that case where would we append clustersAfterFailover? The controller could maybe be triggered to update the attribute after determining scheduling, but might be trickier.

pkg/controllers/cluster/taint_manager.go
mszacillo: I can double check, but I don’t believe so. My understanding was that this is used for cluster failover, which is separate from the application failover feature.

RainbowMango:
> What I would suggest is we could use a read-write safe failover cache which stores the ResourceBinding's Name/Namespaceas a key, with the value being[][]TargetCluster`. 

1. Compared to the cache, I prefer using API to record the state, because it's a simple pattern, and don't have to worry about cases like component restarts and reloading. Let's see if we can address these changes before turning the direction. 
2. Firstly, since the cache is built in `karmada-controller-manager`'s memory, no way to share it with the `karmada-scheduler`. Secondly, I'm hesitant to let the `karmada-scheduler` involved too much.

mszacillo:
> Firstly, since the cache is built in `karmada-controller-manager`'s memory, no way to share it with the `karmada-scheduler`. 

Oof…good point. :(

pkg/controllers/cluster/taint_manager.go
RainbowMango: > My understanding was that this is used for cluster failover, which is separate from the application failover feature.

Oh, yes, you are right.  My bad.

pkg/scheduler/framework/plugins/clustereviction/cluster_eviction.go
RainbowMango: > Since clustersBeforeFailure only include clusters that have been marked for eviction, we shouldn’t reschedule replicas to those clusters.

Maybe I have a different view on this. The `clusterBeforeFailover` is designed to hold all clusters before the failover, including the normal clusters and those targeted to be evicted.

This piece of code will filter all clusters out, not just the one marked for eviction, right?
```golang
if containsCluster(lastFailover.ClustersBeforeFailover, cluster.Name) {
	klog.V(2).Infof("Workload has been failed over from this cluster %s.", cluster.Name)
	return framework.NewResult(framework.Unschedulable, fmt.Sprintf("workload has been failed over from this cluster %s", cluster.Name))
}
```

mszacillo: We added the UpdateFailoverStatus method within the `evictBinding` method of the rb application failover controller. Based off https://github.com/karmada-io/karmada/blob/master/pkg/controllers/applicationfailover/rb_application_failover_controller.go#L135, the only clusters passed into the eviction method are clusters for which replicas are unhealthy, and where the toleration time has elapsed. 

So we'll only be appending clusters that are meant for eviction to the `clustersBeforeFailover`, which is why we want to filter them out before scheduling.

mszacillo: I guess my understanding of the `clustersBeforeFailover` attribute was that we're tracking only the clusters that are undergoing eviction - that way we can track which replicas have been failed over (from their originating clusters to their destinations).

mszacillo:
@RainbowMango Would you be opposed to making the `clustersAfterFailover` optional for now, while we determine the best way to update that field? As the other fields (`clustersBeforeFailover` included) are all pretty clear and easily set during the initial failover step. 

Otherwise I've been thinking about workarounds to accurately capturing the `clustersAfterFailover` without relying on the scheduler, and without having to keep track of the order of failoverHistoryItems. Perhaps instead of combining `clustersBeforeFailover` and `clustersAfterFailover`, we could introduce a concept of a `FailoverEvent` which can take two forms (names WIP):
1. FailoverTriggered event -> (`clustersBeforeFailover`, `startTime`, `reason`)
2. FailoverCompleted event -> (`clustersAfterFailover`, `endTime`).

FailoverTriggered could be controlled by the existing failover controller, whereas FailoverCompleted could be controlled by a new failover controller which tracks changes in RB placement.

The benefits include the scheduler does not need to update the failover history status. It would also decouple the `clustersBeforeFailover` and `clustersAfterFailover` fields, which are difficult to track in order in cases of divided replicas. The drawback is that having different events could make readability more difficult. It would also require altering one of the existing failover controllers or adding a new controller to track failover completion. 

Just throwing out an idea ~

RainbowMango:
> Would you be opposed to making the clustersAfterFailover optional for now, while we determine the best way to update that field? As the other fields (clustersBeforeFailover included) are all pretty clear and easily set during the initial failover step.

I've been investigating this. It seems we can generate the failover history in [the resource-binding-graceful-eviction-controller](https://github.com/karmada-io/karmada/blob/38ee277de7aaaf4f93fec42d49bfc71cc4b0be76/pkg/controllers/gracefuleviction/rb_graceful_eviction_controller.go#L80-L100) when clearing the eviction tasks.

Regarding to the concern on above:
> The other issue is if the gracePeriod is too low, the eviction task will be cleared before scheduling take place, meaning the we won't be able to construct the failoverHIstory.

The [the resource-binding-graceful-eviction-controller](https://github.com/karmada-io/karmada/blob/38ee277de7aaaf4f93fec42d49bfc71cc4b0be76/pkg/controllers/gracefuleviction/rb_graceful_eviction_controller.go#L80-L100) guarantees the eviction task to be cleared after the scheduler react on that.

@XiShanYongYe-Chang  We are going to figure out when to generate the failover history. 
Both application failover and cluster failover ended in [the resource-binding-graceful-eviction-controller](https://github.com/karmada-io/karmada/blob/38ee277de7aaaf4f93fec42d49bfc71cc4b0be76/pkg/controllers/gracefuleviction/rb_graceful_eviction_controller.go#L80-L100), so this might be the best place to generate the history.  What do you think?

pkg/controllers/utils/common.go
```diff
@@ -0,0 +1,59 @@
...
+
+// UpdateFailoverStatus adds a failoverHistoryItem to the failoverHistory field in the ResourceBinding.
+func UpdateFailoverStatus(client client.Client, binding *workv1alpha2.ResourceBinding, failoverType workv1alpha2.FailoverReason) (err error) {
+	klog.V(4).Infof("Updating failover status for ResourceBinding(%s/%s)", binding.Name, binding.Namespace)
+	err = retry.RetryOnConflict(retry.DefaultRetry, func() (err error) {
+		_, err = helper.UpdateStatus(context.Background(), client, binding, func() error {
+			failoverHistoryItem := workv1alpha2.FailoverHistoryItem{
+				StartTime:              metav1.Time{Time: time.Now()},
+				Reason:                 failoverType,
+				ClustersBeforeFailover: binding.Spec.Clusters,
```
XiShanYongYe-Chang, line 40: The assignment here seems to be inconsistent with the [comment](https://github.com/karmada-io/karmada/pull/5251/files#r1818329655), where the ClustersBeforeFailover assignment should be all clusters that were last scheduled.

pkg/controllers/cluster/taint_manager.go
```diff
@@ -170,6 +171,10 @@ func (tc *NoExecuteTaintManager) syncBindingEviction(key util.QueueKey) error {
 	// Case 2: Need eviction after toleration time. If time is up, do eviction right now.
 	// Case 3: Tolerate forever, we do nothing.
 	if needEviction || tolerationTime == 0 {
+		err := controllerUtils.UpdateFailoverStatus(tc.Client, binding, workv1alpha2.EvictionReasonTaintUntolerated)
```
XiShanYongYe-Chang, line 174: The type of the third input parameter is incorrect. The correct type is workv1alpha2.FailoverReason.

pkg/controllers/utils/common.go
mszacillo: You're totally right, this should be the clusters that are used in `evictBinding`.

mszacillo:
> Both application failover and cluster failover ended in [the resource-binding-graceful-eviction-controller](https://github.com/karmada-io/karmada/blob/38ee277de7aaaf4f93fec42d49bfc71cc4b0be76/pkg/controllers/gracefuleviction/rb_graceful_eviction_controller.go#L80-L100), so this might be the best place to generate the history. What do you think?

If we're interested in generating the historyItem in one place then I think that's a good idea. Additionally we are reliant on a failover label to be appended to the workload in our use-case, so if we can guarantee that the eviction task is not cleaned up before scheduling, we can check the eviction reason and append the label if necessary.

mszacillo:
@RainbowMango The e2e test failure is due to interference with the existing cluster filtration step that checks `clustersBeforeFailover` and the workload rebalancer. Since the failoverHistory is not ephemeral, it will always filter out clusters in `clustersBeforeFailover` until there is a new failover. This doesn't work alongside the workload rebalancer.

Seems we should keep the existing filter logic that only checks eviction tasks (as long as we make sure these are cleaned up post scheduling).

RainbowMango:
> Seems we should keep the existing filter logic that only checks eviction tasks (as long as we make sure these are cleaned up post scheduling).

If that's the case, we can narrow down the scope of this PR and focus on generating the history.

Note that, currently, for purge modes `Graciously` and `Never`, we can ensure the eviction task exists before scheduling, as we will build an eviction task for them. However, we still need to take care of the `Immediately` mode. We can iterate it by separate PR.

mszacillo:
/retest

Seems agents were interrupted.

pkg/controllers/utils/common.go
```diff
@@ -0,0 +1,64 @@
...
+
+	policyv1alpha1 "github.com/karmada-io/karmada/pkg/apis/policy/v1alpha1"
+	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
+	"github.com/karmada-io/karmada/pkg/util/helper"
+)
+
+// UpdateFailoverStatus adds a failoverHistoryItem to the failoverHistory field in the ResourceBinding.
+func UpdateFailoverStatus(client client.Client, binding *workv1alpha2.ResourceBinding, clusters []string, failoverType workv1alpha2.FailoverReason) (err error) {
+	// If the resource is Duplicated, then it does not have a concept of failover. We skip attaching that status here.
+	if binding.Spec.Placement.ReplicaScheduling.ReplicaSchedulingType == policyv1alpha1.ReplicaSchedulingTypeDuplicated {
```
mszacillo, line 36: I've added this filter as I was noticing failoverHistory being added to duplicated resources in response to cluster failures. I really don't believe duplicated resources should have a failover history, since there is no concept of failover in those cases. 

Let me know if you have concerns.

mszacillo:
@RainbowMango Thanks for putting together the stateful application support checklist, its very comprehensive! 

I've just updated this PR to remove the last bit of the failover flag implementation, and have this PR solely focus on the API change. Running some tests for both divided and grouped replicas looked as we expect:

**Single replica**
```
failoverHistory:
  - clustersAfterFailover:
    - cluster-A
    clustersBeforeFailover:
    - cluster-B
    reason: ApplicationFailure
    startTime: "2024-11-05T19:04:14Z"
  - clustersAfterFailover:
    - cluster-B
    clustersBeforeFailover:
    - cluster-A
    reason: ApplicationFailure
    startTime: "2024-11-05T19:04:45Z"
```

**Divided replicas**
```
failoverHistory:
  - clustersAfterFailover:
    - cluster-A
    - cluster-B
    clustersBeforeFailover:
    - cluster-A
    - cluster-B
    reason: ApplicationFailure
    startTime: "2024-11-05T19:00:07Z"
  - clustersAfterFailover:
    - cluster-A
    - cluster-B
    clustersBeforeFailover:
    - cluster-B
    - cluster-A
    reason: ApplicationFailure
    startTime: "2024-11-05T19:00:38Z"
```

Note: The purgeMode for this resource was set to immediately, and since we've decided to go with the evictionTask strategy for filtering out clusters, the Karmada scheduler rescheduled the replicas to the same clusters. This should be resolved in the future. Please let me know if anything else needs to be addressed in this PR. 

Cheers!

mszacillo:
Additional comment before I forget - we still append the `clustersAfterFailover` within the scheduler for now. As discussed, this logic should be moved into the evictionTask controller before the eviction task gets cleared. We can tackle that in an additional PR. 

If you'd like to remove the setting of the failoverHistory and keep this as simply an API change, I could do that as well. Let me know.

XiShanYongYe-Chang:
Hi @Dyex719 @mszacillo @RainbowMango 

Regarding the specific implementation details of the plan, I have combined previous discussions and preliminary conclusions to conduct some analysis and organization, which should aid in the implementation of the feature. Can you help take a look?

https://docs.google.com/document/d/1Np67nVXokSCAfzrI8P8JwQgvqJF64068Yu6leDn2Kjs/edit?tab=t.0#heading=h.eyvom3e0lj8w

pkg/controllers/utils/common.go
RainbowMango: Nice finding!
But for resources with scheduling type duplicated, why not disable failover in the first place? So that, there won't be any eviction task for them.
@XiShanYongYe-Chang What do you think?

RainbowMango:
> Additional comment before I forget - we still append the `clustersAfterFailover` within the scheduler for now. As discussed, this logic should be moved into the evictionTask controller before the eviction task gets cleared. We can tackle that in an additional PR.
> 
> If you'd like to remove the setting of the failoverHistory and keep this as simply an API change, I could do that as well. Let me know.

I'm still concerned about letting the scheduler get involved in the failover process. But haven't gotten a chance to explore further. 

In addition, according to #5788, seems we don't have a strong dependency on the history to figure out which cluster the application would be migrated to. So, the history information is kind of optional for this feature. That's why I put this task to part 3. (But it still makes sense from an instrumentation point of view.)

So, I'm thinking should we focus on Part 1 and Part 2 first, and come back on the history thing later?

pkg/controllers/utils/common.go
XiShanYongYe-Chang: > But for resources with scheduling type duplicated, why not disable failover in the first place? 

Does that mean for duplicated resources, we don't put them in the eviction queue?

RainbowMango: Yes, I mean we should revisit which scheduling type is applicable for failover. 

But, wait, I think for scheduling type duplicated, there might be a case that also needs failover, especially if they use spread constraints. Anyway, this deserves to double confirm.

XiShanYongYe-Chang: This may be another separate point to consider for a failover, not subordinate to the current task.

XiShanYongYe-Chang:
> Regarding the specific implementation details of the plan, I have combined previous discussions and preliminary conclusions to conduct some analysis and organization, which should aid in the implementation of the feature. Can you help take a look?
> 
> https://docs.google.com/document/d/1Np67nVXokSCAfzrI8P8JwQgvqJF64068Yu6leDn2Kjs/edit?tab=t.0#heading=h.eyvom3e0lj8w

Hi @mszacillo do you have time to come to the community meeting tomorrow and exchange ideas about the program?

RainbowMango:
Hi @mszacillo you are not going to KubeCon this time, right?

mszacillo:
Hi @RainbowMango, sadly I won’t be there this time. :( 

I’m thinking about attending KubeCon EU next year, but that’s still far away.

mszacillo:
Hi @RainbowMango, we can go ahead and close this!

mszacillo:
> We just released v1.12 last week, have you ever tested it on your side? I want to know if it works as we expected.
We're working on rebasing our local branch onto v1.12. I'll be sure to give an update once we are finished and can thoroughly test. 

> In addition, v1.12 covered application failover, shall we go ahead with cluster failover?
And yes that would be great. Perhaps we can discuss design ideas in the next community meeting?

RainbowMango:
> And yes that would be great. Perhaps we can discuss design ideas in the next community meeting?

That would be great!

XiShanYongYe-Chang:
Hi @mszacillo I'm not sure if this is the right place to ask you a question. I'm currently designing the optimization of the cluster failover feature. I'd like to know how you migrate the workload when the cluster is faulty in the scenario where you use Flink Deployment?

When the cluster is faulty, the Karmada control-plane cannot delete resources from the cluster. Therefore, it seems that workload in the old cluster cannot be deleted and workloads in the new cluster cannot be created during the migration. This is the effect of the `Immediately` migration mode provided in Application Failover.

mszacillo:
> I'd like to know how you migrate the workload when the cluster is faulty in the scenario where you use Flink Deployment?
> 
> When the cluster is faulty, the Karmada control-plane cannot delete resources from the cluster. Therefore, it seems that workload in the old cluster cannot be deleted and workloads in the new cluster cannot be created during the migration. 

Yes this was a pretty big concern for us. Some of our users run jobs with only-once-processing guarantees. Since these guarantees depend on Flink's internal state, running two duplicate jobs in parallel without configuring the entire pipeline (sink included) to handle duplicate records can lead to problems.

In the case of a cluster failure, or if Karmada cannot communicate with the member cluster's control-plane, we still immediately failover the resource. But we've instrumented an initcontainer + sidecar solution, which handles assuming leadership and prevents duplicate applications from both processing at the same time. If Karmada eventually establishes connection to the previous cluster, the old FlinkDeployment will be deleted, and the new FlinkDeployment should start initializing. 

> This is the effect of the `Immediately` migration mode provided in Application Failover.

Is this something that you are intending to add as a standard bahavior for the `Immediately` purgeMode? Because currently Karmada does not wait for the previous resource to be fully deleted before rescheduling. But I am not opposed to this idea - perhaps this is something that could be further configured? 

For example, by default `Immediately` purgeMode could make Karmada delete the resource, and force Karmada to wait for confirmation that the resource is fully deleted before rescheduling. There could be another configuration like `safeRescheduling` which would default to true, but a user could toggle this off to allow Karmada to reschedule the workload without waiting for a delete confirmation. That could be helpful in cases that users already have their own instrumentation setup to handle duplicate deployments.

XiShanYongYe-Chang:
Thank you for replying!

> In the case of a cluster failure, or if Karmada cannot communicate with the member cluster's control-plane, we still immediately failover the resource. 

Did you do this by turning off `GracefulEviction` FeatureGate?

> But we've instrumented an initcontainer + sidecar solution, which handles assuming leadership and prevents duplicate applications from both processing at the same time. 

Although I don't know the details, thanks to your explanation, I have a general understanding of the intention of doing so.

>  If Karmada eventually establishes connection to the previous cluster, the old FlinkDeployment will be deleted, and the new FlinkDeployment should start initializing.

Does this mean that Flink Deployment on the new cluster will not start as long as the cluster is not healthy? It doesn't affect your application, does it?

> Is this something that you are intending to add as a standard bahavior for the Immediately purgeMode? Because currently Karmada does not wait for the previous resource to be fully deleted before rescheduling. But I am not opposed to this idea - perhaps this is something that could be further configured?

Yes, that's really one of our goals, to make the `Immediately` migration of application failures the name matches the reality . I want to confirm whether the `Immediately` migration behavior is consistent for both the application failover and the cluster failover, that is, waiting for the old workload to be completely deleted before starting a new workload.

If the two have the same understanding of `Immediately`, we can think together about how to achieve this. If they are not consistent, we can consider them separately.

My concern now is whether `Immediately` makes sense for cluster failover. If the faulty cluster is not recovered, new workloads cannot be started. Is this your expectation?

> For example, by default Immediately purgeMode could make Karmada delete the resource, and force Karmada to wait for confirmation that the resource is fully deleted before rescheduling. There could be another configuration like safeRescheduling which would default to true, but a user could toggle this off to allow Karmada to reschedule the workload without waiting for a delete confirmation. That could be helpful in cases that users already have their own instrumentation setup to handle duplicate deployments.

This solution looks interesting. I look forward to further discussing the specific solution with you. Maybe we can open a new issue or proposal. I happen to have some ideas on how to achieve it, but it may not be mature enough now. 

But before we start the discussion, we might also want to make sure that we do this for cluster failover as well, which is my question above. If we do not consider the processing of cluster failover for the time being, we can focus on only the processing of application failover.

mszacillo:
> Did you do this by turning off GracefulEviction FeatureGate?

No, we still use the GracefulEviction FeatureGate, but we do have a small internal commit that reuses the changes we made to support `Immediately` purgeMode within GracefulEvictionTask, which are currently used only for application failover. Ideally with your optimizations, the same state preservation feature we added to application failover should be made available to use for cluster failover. 

> Does this mean that Flink Deployment on the new cluster will not start as long as the cluster is not healthy? It doesn't affect your application, does it?

Yeah, the application will wait in an initializing state until the other application is successfully deleted by Karmada. (Or in the case that the entire cluster really went down, the application should eventually start as soon as the existing heartbeat expires). 

There are some implications here though - while the initContainer is waiting to assume leadership, the application will still be marked as unhealthy. Meaning Karmada will eventually try to reschedule the job after the tolerationSeconds elapses. There are two improvements here:

1. The stateful preservation label should be conserved on subsequent failovers if the failed-over application never became healthy. (I'm currently working on this - perhaps I can publish a PR later after I test more thoroughly). 

2. Make `tolerationSeconds` field more fine-grained in the `failover` API. For example, we could have an `initTolerationSeconds` and a regular `tolerationSeconds`. The first field describes how long Karmada will wait for the application to successfully transition to healthy for the first time. If the application becomes unhealthy later on, the regular tolerationSeconds will be used. 

This is helpful since some of our users have applications with very high data volumes, which can take some time for the application to start up. But they also want to be sensitive to application failures and failover when necessary - at the moment we need to set the `tolerationSeconds >= time to initialize`.
 
> My concern now is whether Immediately makes sense for cluster failover. If the faulty cluster is not recovered, new workloads cannot be started. Is this your expectation?

Yeah this is not ideal, although I'm not sure what the best approach is. Karmada won't be able to differentiate cases in which it temporarily does not have connectivity with a cluster versus the cluster being entirely down and unrecoverable. I think this is something that perhaps we can leave up to the end user? If they want to be careful, they can configure failover so that applications only reschedule if Karmada is able to delete the resource. In the case that the cluster is unrecoverable, the user will need to manually reapply their job. 

> This solution looks interesting. I look forward to further discussing the specific solution with you. Maybe we can open a new issue or proposal. I happen to have some ideas on how to achieve it, but it may not be mature enough now.

> But before we start the discussion, we might also want to make sure that we do this for cluster failover as well, which is my question above. If we do not consider the processing of cluster failover for the time being, we can focus on only the processing of application failover.

We are definitely interested in cluster failover as well (since we rely on it currently to monitor cluster outages, and allow for maintenance-windows where we trigger taint-based cluster evictions). 

Definitely looking forward to seeing more of your proposal and design! Apologies for the lengthy response - there is a lot to go over here. :)

XiShanYongYe-Chang:
> No, we still use the GracefulEviction FeatureGate, but we do have a small internal commit that reuses the changes we made to support Immediately purgeMode within GracefulEvictionTask, which are currently used only for application failover. Ideally with your optimizations, the same state preservation feature we added to application failover should be made available to use for cluster failover.

Thanks, I get it.

> Or in the case that the entire cluster really went down, the application should eventually start as soon as the existing heartbeat expires

Is that another idea of yours, or is that what you're already doing.

> There are some implications here though - while the initContainer is waiting to assume leadership, the application will still be marked as unhealthy. Meaning Karmada will eventually try to reschedule the job after the tolerationSeconds elapses. 

It's a great detail, and we've thought about it before, but we didn't deal with it, and we only supported the most basic functions, which is where we need to continue iteration.

> There are two improvements here:

I understand that the second point should be a specific solution idea for the first point. This is a new task, and we should track it with issues. 

> Yeah this is not ideal, although I'm not sure what the best approach is. Karmada won't be able to differentiate cases in which it temporarily does not have connectivity with a cluster versus the cluster being entirely down and unrecoverable. I think this is something that perhaps we can leave up to the end user? If they want to be careful, they can configure failover so that applications only reschedule if Karmada is able to delete the resource. In the case that the cluster is unrecoverable, the user will need to manually reapply their job.

+1, Let the user define it first. After more user feedback is accumulated, we can further provide more common capabilities for users.

> We are definitely interested in cluster failover as well (since we rely on it currently to monitor cluster outages, and allow for maintenance-windows where we trigger taint-based cluster evictions).
Definitely looking forward to seeing more of your proposal and design! Apologies for the lengthy response - there is a lot to go over here. :)

Thanks a lot for your lengthy response!

> Is this something that you are intending to add as a standard bahavior for the Immediately purgeMode? Because currently Karmada does not wait for the previous resource to be fully deleted before rescheduling. But I am not opposed to this idea - perhaps this is something that could be further configured?

Now I'm sure that the cluster failover optimization I'm currently working on will not affect this part (I'll submit a proposal later), and we can provide a solution for this goal.

mszacillo:
I have filed #6232 and #6233 as follow-ups!

XiShanYongYe-Chang:
> I have filed #6232 and #6233 as follow-ups!

Thanks~ I see it, we can have further discussions under these issues.

PR #5253, opened 2024-07-26
fix: kustomize syntax for webhook caBundle

a7i:
**What type of PR is this?**
/kind cleanup

The current syntax does not work for kustomize
```
❯ kustomize build charts/karmada/_crds/
# Warning: 'patchesStrategicMerge' is deprecated. Please use 'patches' instead. Run 'kustomize edit fix' to update your Kustomization automatically.
Error: wrong node kind: expected ScalarNode but got MappingNode: node contents:
{? {caBundle: ''} : ''}
```

Adding quotes around it makes it work for kustomize build while still working for helm charts. This is a no risk change for those that prefer using kustomize.

Fixes #

```release-note
NONE
```


XiShanYongYe-Chang (commented):
Thanks~
/lgtm
/approve

PR #5255, opened 2024-07-26
fix unexpected interference between different e2e use cases

### LINKED ISSUE #5256: flake test: the resource selector  scope of some Policies in E2E use case is too large to affect other use cases.
**What happened**:

The resource selector  scope of some Policies in E2E use case is too large to affect other use cases.

take below case as an example:

https://github.com/karmada-io/karmada/blob/1430ecbd2eb7e8321724ed9b6b08436b9b0b051f/test/e2e/propagationpolicy_test.go#L558-L568

this policy matches all deployment, however, e2e use cases are running in parallel, it may be result in other case failure.

e2e use cases do not affect each other

I will list relevant failures in the comments below

**Environment**:
- Karmada version:
- Others:

chaosi-zju:
**What type of PR is this?**

/kind failing-test

The scope of some Policy's resourceSelector in E2E use case is too large to affect other use cases.

detail refer to #5256

Fixes #5256

```release-note

```

### DIFF
--- a/test/e2e/clusterpropagationpolicy_test.go
+++ b/test/e2e/clusterpropagationpolicy_test.go
@@ -570,6 +570,7 @@ var _ = ginkgo.Describe("[ImplicitPriority] propagation testing", func() {
 					APIVersion: deployment.APIVersion,
 					Kind:       deployment.Kind,
 					Name:       deployment.Name,
+					Namespace:  deployment.Namespace,
 				},
 			}, policyv1alpha1.Placement{
 				ClusterAffinity: &policyv1alpha1.ClusterAffinity{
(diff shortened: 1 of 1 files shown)


XiShanYongYe-Chang:
A little more background on the problem?

chaosi-zju:
> A little more background on the problem?

yes, I will add a issue to describe it~

test/e2e/propagationpolicy_test.go
```diff
@@ -538,6 +538,7 @@ var _ = ginkgo.Describe("[ImplicitPriority] propagation testing", func() {
 					APIVersion: deployment.APIVersion,
 					Kind:       deployment.Kind,
 					Name:       deployment.Name,
+					Namespace:  deployment.Namespace,
```
zhzhuang-zju, line 541: In fact, there's no need to change it here, because the namespace of the deployment is by default the same as the one of the pp

zhzhuang-zju:
/lgtm
thanks~

XiShanYongYe-Chang (commented):
/approve

Issue #5136, opened 2024-07-04
deployment replicas syncer testing


RainbowMango:
#### Which test(s) are flaking:
https://github.com/karmada-io/karmada/actions/runs/9788476477/job/27026725265

#### Anything else we need to know:
```
Timeline >>
  STEP: Creating PropagationPolicy(karmadatest-m5dpb/deploy-g69xd) @ 07/04/24 04:42:38.3
  STEP: Creating Deployment(karmadatest-m5dpb/deploy-g69xd) @ 07/04/24 04:42:38.31
  STEP: Creating HPA(karmadatest-m5dpb/deploy-g69xd) @ 07/04/24 04:42:38.315
  STEP: step1: propagate 4 replicas to two clusters @ 07/04/24 04:42:38.323
  STEP: Waiting for deployment(karmadatest-m5dpb/deploy-g69xd) synced on member clusters @ 07/04/24 04:42:38.323
  STEP: Waiting for deployment(karmadatest-m5dpb/deploy-g69xd) synced on member clusters @ 07/04/24 04:42:43.332
  STEP: step2: hpa scale each member cluster replicas from 2 to 3 @ 07/04/24 04:42:43.349
  STEP: Updating HPA(karmadatest-m5dpb/deploy-g69xd) @ 07/04/24 04:42:43.349
  STEP: Waiting for deployment(karmadatest-m5dpb/deploy-g69xd) synced on member clusters @ 07/04/24 04:42:43.369
  STEP: Waiting for deployment(karmadatest-m5dpb/deploy-g69xd) synced on member clusters @ 07/04/24 04:42:58.399
  STEP: step3: manually add deployment template replicas from 6 to 10 @ 07/04/24 04:42:58.422
  STEP: Updating Deployment(karmadatest-m5dpb/deploy-g69xd)'s replicas to 10 @ 07/04/24 04:42:58.422
  STEP: Waiting for deployment(karmadatest-m5dpb/deploy-g69xd) synced on member clusters @ 07/04/24 04:42:58.447
  STEP: Waiting for deployment(karmadatest-m5dpb/deploy-g69xd) synced on member clusters @ 07/04/24 04:42:58.461
  [FAILED] in [It] - /home/runner/work/karmada/karmada/test/e2e/deploymentreplicassyncer_test.go:247 @ 07/04/24 04:43:58.472
  STEP: Removing PropagationPolicy(karmadatest-m5dpb/deploy-g69xd) @ 07/04/24 04:43:58.473
[...]

PR #5263, opened 2024-07-27
print all binding and related objects to log file after each e2e failure case

chaosi-zju:
**What type of PR is this?**

/kind flake

When e2e failed, the test environment is already destroyed, but we maybe we need to locate problem by refering ResourceTemplate/ResourceBinding/Work. So,we want to record all binding and related objects to log file after each e2e failure case.

Fixes part of #5136

If here is any e2e case failed, you can find the corresponding exported logs in `ci runner -> Summary -> Artifacts -> karmada_kind_log_v1.x.x`, such as:

image

```release-note

```

### DIFF
--- a/test/e2e/suite_log_test.go
+++ b/test/e2e/suite_log_test.go
@@ -1,0 +1,102 @@
+/*
+Copyright 2024 The Karmada Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package e2e
+
+import (
+	"context"
+	"encoding/json"
+	"os"
+	"path/filepath"
+
+	"github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
+	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
+	"github.com/karmada-io/karmada/pkg/util/restmapper"
+)
+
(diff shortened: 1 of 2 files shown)


test/e2e/suite_log_test.go
```diff
@@ -0,0 +1,99 @@
...
+	gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+	resource, err := dynamicClient.Resource(gvr).Namespace(spec.Resource.Namespace).Get(context.TODO(), spec.Resource.Name, metav1.GetOptions{})
+	gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+
+	// print resource template to log file
+	printObjectToFile(file, resource)
+
+	for _, clusterName := range framework.ClusterNames() {
+		workNamespace := names.GenerateExecutionSpaceName(clusterName)
+		workList, err := karmadaClient.WorkV1alpha1().Works(workNamespace).List(context.TODO(), metav1.ListOptions{
```
XiShanYongYe-Chang, line 76: If namespaces are not distinguished, can a list request query all target work objects?

test/e2e/suite_log_test.go
```diff
@@ -0,0 +1,99 @@
...
+const logDir = "/tmp/karmada/objects"
+
+func printAllBindingAndRelatedObjects() {
+	filePath := filepath.Join(logDir, testNamespace+".txt")
+	err := os.MkdirAll(logDir, 0700)
+	gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+	logFile, err := os.OpenFile(filePath, os.O_CREATE|os.O_APPEND|os.O_RDWR, 0600)
+	gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+	defer logFile.Close()
+
```
XiShanYongYe-Chang, line 45: Some cases also need to query the information about the cluter object. Can the information be output synchronously?

test/e2e/suite_log_test.go
chaosi-zju: it‘s surprising, refer to https://github.com/kubernetes/client-go/issues/159#issuecomment-288624475, it really can query all objects in just one list request.

test/e2e/suite_log_test.go
chaosi-zju: done

XiShanYongYe-Chang (commented):
Thanks~
/lgtm
/approve

In the future, we can consider using parameters to determine whether to output the information, because some local tests may not need to output the information.

PR #5266, opened 2024-07-28
make golangci-lint run locally with correct version

grosser:
**What type of PR is this?**

/kind cleanup
/kind failing-test

currently verify-staticcheck does not work for me:
```
./hack/verify-staticcheck.sh
Installing golangci-lint v1.59.0
golangci/golangci-lint info checking GitHub for tag 'v1.59.0'
golangci/golangci-lint info found version: 1.59.0 for v1.59.0/darwin/arm64
golangci/golangci-lint info installed /Users/mgrosser/go/bin/golangci-lint
./hack/verify-staticcheck.sh: line 36: golangci-lint: command not found

Please review the above warnings.
Tips: The golangci-lint might help you fix some issues, try with the command "golangci-lint run --fix".
If the above warnings do not make sense, feel free to file an issue.
```

also it uses whatever random local version a user might have installed which is unreliable

```release-note
NONE
```


hack/verify-staticcheck.sh
```diff
@@ -13,32 +13,31 @@
...
+# - run golangci-lint
 
 set -o errexit
 set -o nounset
 set -o pipefail
 
 REPO_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
-GOLANGCI_LINT_VER="v1.59.0"
+GOLANGCI_LINT_VER="1.59.0"
+GOLANGCI_LINT="./bin/golangci-lint"
```
liangyuanpeng, line 25: seems like you want to download the tools to the project folder and not `GOPATH`, I think this makes sense, but currently karmada downloads tools into `GOPATH`, which might be worth discussing.

grosser: yeah, other tools would need to follow that pattern, so better get consensus
idk of any reason not to do it 🤷

grosser:
I know why the error happened, but I don't want it to happen + I want it to use the correct version

hack/verify-staticcheck.sh
liangyuanpeng: I don't have much background here, I think it would be possible to add a parameter to have the tool download to the specifically directory, defaulting to gopath, what do you think?  

Maybe:
```shell
GO_PATH=$(go env GOPATH)
TOOLS_PATH=${TOOLS_PATH:-GO_PATH}
```

liangyuanpeng: cc @zhzhuang-zju @XiShanYongYe-Chang @RainbowMango

grosser: that would fix the error

if we also want the correct version it would need to switch and replace the binaries in gopath depending on what version the current project needs, should mostly be fine just weird when using different projects with different version requirements

chaosi-zju:
> also it uses whatever random local version a user might have installed which is unreliable

I think this makes sense.

Maybe the user himself has installed an unexpected version of binary under `/usr/local/bin`. When we execute the `golangci-lint run` command, we may use an unexpected version. 

No matter where we place the binary we download, it would be better to use absolute paths.

hack/verify-staticcheck.sh
zhzhuang-zju: I think we can provide the capability to install to a specific path, but considering that minimizing the impact on users who may have already installed their own `golang-lint`, the default path could be set to `./bin/`

.gitignore
```diff
@@ -10,8 +10,11 @@
 
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
+
+# Folders used by make
 _tmp/
 _output/
+bin/
```
zhzhuang-zju, line 17:
```suggestion
_tool
```
How about mimicking the `_output` directory and naming it `_tool` or `_bin`? This directory can be specifically used to place tools for maintaining Karmada.

hack/verify-staticcheck.sh
```diff
@@ -13,32 +13,31 @@
...
-  echo "Installing golangci-lint ${GOLANGCI_LINT_VER}"
-  # https://golangci-lint.run/usage/install/#other-ci
-  curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin ${GOLANGCI_LINT_VER}
+if ! (test -f "${GOLANGCI_LINT}" && ${GOLANGCI_LINT} --version | grep " ${GOLANGCI_LINT_VER} " >/dev/null); then
+  rm -f ${GOLANGCI_LINT} && echo "Installing ${GOLANGCI_LINT} ${GOLANGCI_LINT_VER}"
+  curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b bin -d "v${GOLANGCI_LINT_VER}"
 fi
 
-if golangci-lint run; then
+if ${GOLANGCI_LINT} run; then
```
chaosi-zju, line 35:
```suggestion
if $(go env GOPATH)/bin/golangci-lint run; then
```

maybe this is enough to solve your problem.

As for where to put the downloaded files, we can continue to discuss in the above comment.

.gitignore
grosser: the beauty of `bin/` is that if users have `PATH=./bin` then things "just work" :)

not against _tool either to be consistent with the other _ folders
never really saw a project use _ folders 🤷 
everything could be in `tmp/` (tmp/output tmp/bin/ etc)

but either way is fine for me

zhzhuang-zju: > the beauty of bin/ is that if users have PATH=./bin then things "just work" :)

agree

XiShanYongYe-Chang:
Hi @grosser, how is it going now?

grosser:
afaik this is ready to merge
... then can make more PRs to move the other local tools to ./bin (did not check how many more there are)

XiShanYongYe-Chang:
There seems to be some discussion going on.

grosser:
yeah, not sure where that discussion is going or what the exact proposal is
... still think bin/ is the best and would work for everyone

PR #5271, opened 2024-07-29
Enhancing default FederatedResourceQuota, extending pp + rb api, adding admission webhook

### LINKED ISSUE #5179: FederatedResourceQuota should be failover friendly
A way for the FederatedResourceQuota to monitoring existing ResourceQuotas (without managing those ResourceQuotas) and impose resource limits on the user based off the sum of all currently used Quota. 

**Why is this needed**:
The existing FederatedResourceQuota mirrors the behavior of a typical Kubernetes ResourceQuota by imposing total resource limits in a multi-cluster setup. This is done by creating statically distributed ResourceQuotas across the specified member clusters, who's limits will total the limits defined in the FederatedResourceQuota. This works if the user **does not** need to worry about DR events which require back-up resources dedicated for failover in the event of a disaster. 

In our case, since we are using Karmada for it's failover feature, we would like clusters to have additionally available quota for each namespace so that in the event of a DR event, all applicatons are able to be rescheduled: 

Screenshot 2024-07-11 at 2 49 14 PM

In the diagram above, we can see that the total limits of the FederatedResourceQuota is 40/40 CPU and 50Gb / 50 Gb Memory. Individual clusters will have the same limit, so that in the case of a DR event, all workloads can be scheduled on one cluster. 

Screenshot 2024-07-11 at 2 54 42 PM

Above, we see that during a failover all workloads from Cluster A will be migrated to Cluster B, where there will be enough available resources to schedule all required pods. With the existing statically defined ResourceQuotas, we cannot support this type of failover.

We've created this ticket to start a discussion on how best to address this limitation, and if this use-case is valid.

mszacillo:
**What type of PR is this?**

/kind feature

**What this PR does**:

This PR introduces a new default behavior to the FederatedResourceQuota which allows users to set overall limits for resources that are applied directly to the Karmada control plane. Any resource that is applied and goes above limit defined in the quota will be denied by the relevant admission webhook:

Screenshot 2024-07-29 at 11 00 25 AM

**Why is this needed?** The use-case is defined in detail in #5179, but in short, the staticAssignment configuration does not currently account for resources needed during failover. Users might want to define resource limits on member clusters individually, which will accommodate large applications, but also set an overall limit from the Karmada side. This feature enables that behavior.

**Change List and Limitations**
- API change to PP to add `FederatedResourceQuota` field. This links resources with a FRQ.
- New FRQ status controller which updates the used resources field by summing up the replicaRequirements of each linked resource in the namespace.
- Admission webhook prevents resources from being applied to the Karmada control plane if they go beyond quota limits.
- Added a safeguard that prevents enabling or disabling staticAssignment via updates. Since we have two status controllers, we want to only toggle them if the FRQ resource is deleted and then reapplied. 
- This change calculates resource requests by using the resourceInterpreter. Currently, support is only added for default and thirdparty interpreters.

Fixes #5179

I have currently opened this as a single PR with two commits (one for the api change and one for the admission webhook). If it is easier to review, I can open two separate PRs instead.

```release-note
Extends pp and rb API to include FederatedResourceQuota field, adding new controller, adding admission webhook
```

### DIFF
--- a/artifacts/deploy/webhook-configuration.yaml
+++ b/artifacts/deploy/webhook-configuration.yaml
@@ -223,6 +223,20 @@ webhooks:
     sideEffects: None
     admissionReviewVersions: [ "v1" ]
     timeoutSeconds: 3
+  - name: federatedresourcequotalimit.karmada.io
+    rules:
+      - operations: ["CREATE", "UPDATE"]
+        apiGroups: ["flink.apache.org"]
+        apiVersions: ["v1beta1"]
+        resources: ["*"]
+        scope: "Namespaced"
+    clientConfig:
+      url: https://karmada-webhook.karmada-system.svc:443/validate-federatedresourcequotalimit
+      caBundle: {{caBundle}}
+    failurePolicy: Fail
+    sideEffects: None
+    admissionReviewVersions: [ "v1" ]
+    timeoutSeconds: 10
   - name: multiclusteringress.karmada.io
     rules:
       - operations: ["CREATE", "UPDATE"]
--- a/charts/karmada/_crds/bases/policy/policy.karmada.io_clusterpropagationpolicies.yaml
+++ b/charts/karmada/_crds/bases/policy/policy.karmada.io_clusterpropagationpolicies.yaml
@@ -169,6 +169,11 @@ spec:
                     - decisionConditions
                     type: object
                 type: object
+              federatedResourceQuota:
+                description: |-
+                  FederatedResourceQuota will link resources with an existing quota.
+                  Quotas can be used to limit the amount of resources users have access to.
+                type: string
               placement:
                 description: Placement represents the rule for select clusters to
                   propagate resources.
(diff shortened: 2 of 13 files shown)


artifacts/deploy/webhook-configuration.yaml
```diff
@@ -223,6 +223,20 @@ webhooks:
     sideEffects: None
     admissionReviewVersions: [ "v1" ]
     timeoutSeconds: 3
+  - name: federatedresourcequotalimit.karmada.io
+    rules:
+      - operations: ["CREATE", "UPDATE"]
+        apiGroups: ["flink.apache.org"]
```
mszacillo, line 229: Note: This is a placeholder that I wanted some feedback from the community for. Currently if we set the admission webhook to apply to all resources, the webhook becomes pretty chatty... 

Wondering if there is any recommendation on how we could limit the webhook to only resources linked to a FederatedResourceQuota.

PR #5273, opened 2024-07-29
Standardize the health probe and metrics arguments of scheduler-estimator

seanlaii:
**What type of PR is this?**
/kind deprecation
/kind feature
/kind cleanup

part of https://github.com/karmada-io/karmada/issues/5219

```release-note
`karmada-scheduler-estimator`:  Add health probe argument `health-probe-bind-address` and metrics argument `metrics-bind-address`. Deprecate `--bind-address` and `--secure-port` flags.
```

### DIFF
--- a/cmd/scheduler-estimator/app/options/options.go
+++ b/cmd/scheduler-estimator/app/options/options.go
@@ -17,6 +17,9 @@ limitations under the License.
 package options
 
 import (
+	"net"
+	"strconv"
+
 	"github.com/spf13/pflag"
 
 	"github.com/karmada-io/karmada/pkg/features"
@@ -35,8 +38,10 @@ type Options struct {
 	Master      string
 	ClusterName string
 	// BindAddress is the IP address on which to listen for the --secure-port port.
+	// Deprecated: To specify the TCP addresse for serving health probes, use HealthProbeBindAddress instead. To specify the TCP addresse for serving prometheus metrics, use MetricsBindAddress instead. This will be removed in release 1.12+.
 	BindAddress string
 	// SecurePort is the port that the server serves at.
+	// Deprecated: To specify the TCP addresse for serving health probes, use HealthProbeBindAddress instead. To specify the TCP addresse for serving prometheus metrics, use MetricsBindAddress instead. This will be removed in release 1.12+.
 	SecurePort int
 	// ServerPort is the port that the server gRPC serves at.
 	ServerPort int
@@ -55,6 +60,16 @@ type Options struct {
 	// Parallelism defines the amount of parallelism in algorithms for estimating. Must be greater than 0. Defaults to 16.
 	Parallelism int
 	ProfileOpts profileflag.Options
+	// MetricsBindAddress is the TCP address that the server should bind to
+	// for serving prometheus metrics.
+	// It can be set to "0" to disable the metrics serving.
+	// Defaults to ":10351".
+	MetricsBindAddress string
+	// HealthProbeBindAddress is the TCP address that the server should bind to
+	// for serving health probes
+	// It can be set to "0" to disable serving the health probe.
+	// Defaults to ":10351".
+	HealthProbeBindAddress string
 }
 
 // NewOptions builds an empty options.
@@ -80,7 +95,24 @@ func (o *Options) AddFlags(fs *pflag.FlagSet) {
 	fs.Float32Var(&o.ClusterAPIQPS, "kube-api-qps", 20.0, "QPS to use while talking with apiserver.")
 	fs.IntVar(&o.ClusterAPIBurst, "kube-api-burst", 30, "Burst to use while talking with apiserver.")
 	fs.IntVar(&o.Parallelism, "parallelism", o.Parallelism, "Parallelism defines the amount of parallelism in algorithms for estimating. Must be greater than 0. Defaults to 16.")
+	// nolint: errcheck
+	fs.MarkDeprecated("bind-address", "This flag is deprecated and will be removed in release 1.12+. Use --health-probe-bind-address and --metrics-bind-address instead. Note: In release 1.12+, these two addresses cannot be the same.")
+	// nolint: errcheck
+	fs.MarkDeprecated("secure-port", "This flag is deprecated and will be removed in release 1.12+. Use --health-probe-bind-address and --metrics-bind-address instead. Note: In release 1.12+, these two addresses cannot be the same.")
+	fs.StringVar(&o.MetricsBindAddress, "metrics-bind-address", "", "The TCP address that the server should bind to for serving prometheus metrics(e.g. 127.0.0.1:10351, :10351). It can be set to \"0\" to disable the metrics serving. Defaults to 0.0.0.0:10351.")
+	fs.StringVar(&o.HealthProbeBindAddress, "health-probe-bind-address", "", "The TCP address that the server should bind to for serving health probes(e.g. 127.0.0.1:10351, :10351). It can be set to \"0\" to disable serving the health probe. Defaults to 0.0.0.0:10351.")
 	features.FeatureGate.AddFlag(fs)
 
 	o.ProfileOpts.AddFlags(fs)
+}
+
+// Complete ensures that options are valid and marshals them if necessary.
+func (o *Options) Complete() error {
+	if len(o.HealthProbeBindAddress) == 0 {
+		o.HealthProbeBindAddress = net.JoinHostPort(o.BindAddress, strconv.Itoa(o.SecurePort))
+	}
+	if len(o.MetricsBindAddress) == 0 {
+		o.MetricsBindAddress = net.JoinHostPort(o.BindAddress, strconv.Itoa(o.SecurePort))
+	}
+	return nil
 }
--- a/cmd/scheduler-estimator/app/scheduler-estimator.go
+++ b/cmd/scheduler-estimator/app/scheduler-estimator.go
@@ -19,10 +19,8 @@ package app
 import (
 	"context"
 	"fmt"
-	"net"
 	"net/http"
 	"os"
-	"strconv"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus/promhttp"
@@ -83,6 +81,11 @@ func NewSchedulerEstimatorCommand(ctx context.Context) *cobra.Command {
 		Long: `The karmada-scheduler-estimator runs an accurate scheduler estimator of a cluster. It 
 provides the scheduler with more accurate cluster resource information.`,
 		RunE: func(_ *cobra.Command, _ []string) error {
+			// complete options
+			if err := opts.Complete(); err != nil {
+				return err
+			}
+
 			// validate options
 			if errs := opts.Validate(); len(errs) != 0 {
 				return errs.ToAggregate()
@@ -114,7 +117,7 @@ provides the scheduler with more accurate cluster resource information.`,
 
 func run(ctx context.Context, opts *options.Options) error {
 	klog.Infof("karmada-scheduler-estimator version: %s", version.Get())
-	go serveHealthzAndMetrics(net.JoinHostPort(opts.BindAddress, strconv.Itoa(opts.SecurePort)))
+	serveHealthzAndMetrics(opts.HealthProbeBindAddress, opts.MetricsBindAddress)
 
 	profileflag.ListenAndServe(opts.ProfileOpts)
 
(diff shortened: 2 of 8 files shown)


whitewindmills (commented):
/assign @RainbowMango @zhzhuang-zju

cmd/scheduler-estimator/app/options/options.go
```diff
@@ -35,8 +38,10 @@ type Options struct {
 	Master      string
 	ClusterName string
 	// BindAddress is the IP address on which to listen for the --secure-port port.
+	// Deprecated: Use HealthProbeBindAddress ＆ MetricsBindAddress instead. And will be removed in release 1.12+.
```
zhzhuang-zju, line 41: how about
```suggestion
	// Deprecated: To specify the TCP addresses for serving health probes, use HealthProbeBindAddress instead. To specify the TCP addresses for serving prometheus metrics, use MetricsBindAddress instead. And will be removed in release 1.12+.
```
or
```suggestion
	// Deprecated: Use HealthProbeBindAddress for health probes and MetricsBindAddress for Prometheus metrics. And will be removed in release 1.12+.
```

cmd/scheduler-estimator/app/options/options.go
```diff
@@ -80,7 +95,24 @@ func (o *Options) AddFlags(fs *pflag.FlagSet) {
 	fs.Float32Var(&o.ClusterAPIQPS, "kube-api-qps", 20.0, "QPS to use while talking with apiserver.")
 	fs.IntVar(&o.ClusterAPIBurst, "kube-api-burst", 30, "Burst to use while talking with apiserver.")
 	fs.IntVar(&o.Parallelism, "parallelism", o.Parallelism, "Parallelism defines the amount of parallelism in algorithms for estimating. Must be greater than 0. Defaults to 16.")
+	// nolint: errcheck
+	fs.MarkDeprecated("bind-address", "This flag is deprecated and will be removed in release 1.12+. Use --health-probe-bind-address instead.")
```
zhzhuang-zju, line 99: > Use --health-probe-bind-address instead

The deprecation message should also mention MetricsBindAddress

cmd/scheduler-estimator/app/options/options.go
```diff
@@ -80,7 +95,24 @@ func (o *Options) AddFlags(fs *pflag.FlagSet) {
...
+	// nolint: errcheck
+	fs.MarkDeprecated("secure-port", "This flag is deprecated and will be removed in release 1.12+. Use --health-probe-bind-address instead.")
+	fs.StringVar(&o.MetricsBindAddress, "metrics-bind-address", "", "The TCP address that the server should bind to for serving prometheus metrics(e.g. 127.0.0.1:10351, :10351). It can be set to \"0\" to disable the metrics serving. Defaults to 0.0.0.0:10357.")
+	fs.StringVar(&o.HealthProbeBindAddress, "health-probe-bind-address", "", "The TCP address that the server should bind to for serving health probes(e.g. 127.0.0.1:10351, :10351). It can be set to \"0\" to disable serving the health probe. Defaults to 0.0.0.0:10357.")
```
zhzhuang-zju, line 103:
```suggestion
	fs.StringVar(&o.MetricsBindAddress, "metrics-bind-address", "", "The TCP address that the server should bind to for serving prometheus metrics(e.g. 127.0.0.1:10351, :10351). It can be set to \"0\" to disable the metrics serving. Defaults to 0.0.0.0:10351.")
	fs.StringVar(&o.HealthProbeBindAddress, "health-probe-bind-address", "", "The TCP address that the server should bind to for serving health probes(e.g. 127.0.0.1:10351, :10351). It can be set to \"0\" to disable serving the health probe. Defaults to 0.0.0.0:10351.")
```

cmd/scheduler-estimator/app/scheduler-estimator.go
```diff
@@ -143,26 +146,62 @@ func run(ctx context.Context, opts *options.Options) error {
...
+func serveCombined(address string) {
 	mux := http.NewServeMux()
-	mux.HandleFunc("/healthz", func(w http.ResponseWriter, _ *http.Request) {
-		w.WriteHeader(http.StatusOK)
-		_, _ = w.Write([]byte("ok"))
-	})
+	mux.HandleFunc("/healthz", healthzHandler)
+	mux.Handle("/metrics", promhttp.HandlerFor(ctrlmetrics.Registry, promhttp.HandlerOpts{
+		ErrorHandling: promhttp.HTTPErrorOnError,
+	}))
```
zhzhuang-zju, line 169:
```suggestion
	mux.Handle("/metrics", metricsHandler)
```

cmd/scheduler-estimator/app/scheduler-estimator.go
```diff
@@ -143,26 +146,62 @@ func run(ctx context.Context, opts *options.Options) error {
 	return nil
 }
 
-func serveHealthzAndMetrics(address string) {
+func serveHealthzAndMetrics(healthProbeBindAddress, metricsBindAddress string) {
+	if healthProbeBindAddress == metricsBindAddress {
+		if healthProbeBindAddress != "0" {
```
zhzhuang-zju, line 151: From the implementation perspective, you have retained the capability to serve Prometheus metrics and health probes using the same TCP address. I agree, but it can also lead to inconsistent behavior compared to components that use `controller-runtime`.

zhzhuang-zju (commented):
thanks @seanlaii

cmd/scheduler-estimator/app/options/options.go
seanlaii: Sounds good! Thanks for the comment. I will update them with the first description.

cmd/scheduler-estimator/app/options/options.go
seanlaii: Sorry, forgot to check this one. Thanks for the catch!

cmd/scheduler-estimator/app/scheduler-estimator.go
seanlaii: Yes, but I was thinking if we change it now, users will have to use the new flag right after the update. Otherwise, they might get `address already in use` error. If we want to make it consistent with `controller-runtime`, maybe I can add this description in the deprecated flag:

> This flag is deprecated and will be removed in release 1.12+. Use --health-probe-bind-address and metrics-bind-address instead. Note: In the next release, these two addresses cannot be the same.

and remove this functionality in the next release?

zhzhuang-zju: This is a viable solution that can be implemented through a smooth upgrade process. cc @RainbowMango  @whitewindmills

whitewindmills: +1

cmd/scheduler-estimator/app/options/options.go
```diff
@@ -55,6 +60,16 @@ type Options struct {
 	// Parallelism defines the amount of parallelism in algorithms for estimating. Must be greater than 0. Defaults to 16.
 	Parallelism int
 	ProfileOpts profileflag.Options
+	// MetricsBindAddress is the TCP address that the server should bind to
+	// for serving prometheus metrics.
+	// It can be set to "0" to disable the metrics serving.
+	// Defaults to ":10351".
+	MetricsBindAddress string
```
seanlaii, line 67: @zhzhuang-zju should I change the default of metricsBindAddress to 8080 to make it the same as https://github.com/karmada-io/karmada/blob/master/cmd/agent/app/options/options.go#L106-L110?

whitewindmills: we can do it in the release-1.12.

pkg/karmadactl/addons/estimator/manifests.go
```diff
@@ -48,12 +48,12 @@ spec:
...
             - --kubeconfig=/etc/{{ .MemberClusterName}}-kubeconfig
             - --cluster-name={{ .MemberClusterName}}
-            - --bind-address=0.0.0.0
-            - --secure-port=10351
             - --grpc-auth-cert-file=/etc/karmada/pki/karmada.crt
             - --grpc-auth-key-file=/etc/karmada/pki/karmada.key
             - --client-cert-auth=true
             - --grpc-client-ca-file=/etc/karmada/pki/ca.crt
+            - --metrics-bind-address=0.0.0.0:10351
+            - --health-probe-bind-address=0.0.0.0:10351
```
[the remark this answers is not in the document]
zhzhuang-zju: If we want to change the default value of `metrics-bind-address` to `:8080` in release 1.12, would it be better to explicitly set it to `:8080` in release 1.11?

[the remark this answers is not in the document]
whitewindmills: it will break backward compatibility.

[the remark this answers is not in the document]
zhzhuang-zju: Why would it break backward compatibility? The default values remain unchanged.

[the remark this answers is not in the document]
whitewindmills: users still collect metrics from port 10351.

cmd/scheduler-estimator/app/scheduler-estimator.go
XiShanYongYe-Chang: > From the implementation perspective, you have retained the capability to serve Prometheus metrics and health probes using the same TCP address. I aggre

+1

> but it can also lead to inconsistent behavior compared to components that use controller-runtime.

This may be a break change for users, and the services related to users must be adapted. I think it is unnecessary. On the back end, these components do support address reuse, and we don't necessarily have to ask the user to differentiate.

seanlaii: Thanks for the feedback and insight! Could you please clarify if you are suggesting that we should remove `go serveCombined(healthProbeBindAddress)` and change the default of `metrics-bind-address` to `8080` in this PR? I want to ensure I fully understand your recommendation. Thanks!

XiShanYongYe-Chang: I am sorry, I didn't explain it clearly. What I meant was that they can use the same address to avoid a breaking change.

whitewindmills:
do you have any comment?
/cc @XiShanYongYe-Chang @zhzhuang-zju

cmd/scheduler-estimator/app/scheduler-estimator.go
zhzhuang-zju: @whitewindmills Do you think we should enforce that the values of `healthProbeBindAddress` and `metricsBindAddress` be different in release 1.12+?

whitewindmills: since we gonna make some break changes in release 1.12+, it should be consistent with other components, that is, these two addresses cannot be the same.

whitewindmills: only this one component can allow these two addresses to be the same, which would be weird and difficult to maintain.

zhzhuang-zju: cc @RainbowMango WDYT?

whitewindmills: ping @RainbowMango

RainbowMango: Will get back on this, but now busying with KubeCon things....

RainbowMango: Generally I like this approach that allows the two endpoints to share the same address, for backward compatibility.
From v1.12, we can give different default values to the `--metrics-bind-address` and `--health-probe-bind-address`. And then, from v1.13 remove the combined style and force separate the two addresses just for ease of maintenance effort.

What do you think @whitewindmills @zhzhuang-zju ?
Sorry for the delay by the way.

pkg/karmadactl/addons/estimator/manifests.go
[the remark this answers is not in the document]
RainbowMango: > If we want to change the default value of metrics-bind-address to :8080 in release 1.12, would it be better to explicitly set it to :8080 in release 1.11?

I don't think so. Just keeping the default port unchanged would be fine.

RainbowMango:
Sorry for letting this sit. I like the idea. There is a conflict that needs to be resolved, I'll try to help deal with it, and then we can move forward if no further comments.

RainbowMango (commented):
/approve

Leaves LGTM to @whitewindmills or any further comments.

PR #5276, opened 2024-07-29
Proposal: Support to Specify Extra Volumes and Volume Mounts for Karmada Control Plane Components

jabellard:
**What type of PR is this?**

/kind design

- Details are contained in the proposal.

Fixes #

```release-note
NONE
```

### DIFF
--- a/docs/proposals/karmada-operator/api-server-extra-volumes-volume-mounts/README.md
+++ b/docs/proposals/karmada-operator/api-server-extra-volumes-volume-mounts/README.md
@@ -1,0 +1,73 @@
+---
+title: Support to Specify Extra Volumes and Volume Mounts for Karmada API Server Component
+authors:
+- "@jabellard"
+reviewers:
+- "@RainbowMango"
+approvers:
+- "@RainbowMango"
+
+creation-date: 2024-07-29
+
+---
+
+# Support to Specify Extra Volumes and Volume Mounts for Karmada API Server Component
+
+
+## Summary
+
+This proposal aims to introduce the ability to specify extra volumes and volume mounts for the Karmada API server component when creating a Karmada instance managed by the Karmada operator. 
+This enhancement will extend the current configuration capabilities, allowing users to meet additional requirements such as having the ability to configure encryption at rest or to configure
+custom authentication webhooks.
+
+## Motivation
+Currently, the Karmada operator allows specifying extra arguments for the Karmada API server component, but lacks support for specifying extra volumes and volume mounts for it. 
+This limitation hinders certain use cases that require additional configurations for security and customization. For instance, as of today, it's not possible to configure a custom authentication 
+webhook nor encryption at rest. 
+
+### Goals
+- Enable users to specify extra volumes and volume mounts for Karmada API server component to unlock the following use cases:
+    - Ability to configure a custom authentication webhook
+    - Ability to configure encryption at rest
+
+It's important to note that although this support will unlock the aforementioned use cases, given that there is a lot that can be configured for the API server component by having the ability to specify not only extra args, but 
+also extra volumes and volume mounts, other potential use cases requiring similar configurations are also unlocked.
+
+
+### Non-Goals
+- Introducing changes to the core functionality of the Karmada API server component.
+- Overhauling the existing configuration of the Karmada API server component beyond the scope of adding extra volumes and volume mounts.
+
+
+## Proposal
+Introduce the new optional fields `ExtraVolumes` and `ExtraVolumeMounts` within the Karmada API server component configuration.
+
+## Design Details
+
+Introduce the new optional fields `ExtraVolumes` and `ExtraVolumeMounts` within Karmada API server component configuration.
+With these new fields, the configuration would look as follows:
+```go
+// KarmadaAPIServer holds settings to kube-apiserver component of the kubernetes.
+// Karmada uses it as its own apiserver in order to provide Kubernetes-native APIs.
+type KarmadaAPIServer struct {
+	// Other existing fields
+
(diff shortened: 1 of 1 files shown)


docs/proposals/karmada-components-extra-volumes/README.md
```diff
@@ -0,0 +1,58 @@
...
+- Enable users to specify extra volumes and volume mounts for Karmada control plane components.
+
+
+### Non-Goals
+- Introducing changes to the core functionality of the Karmada control plane components.
+- Overhauling the existing configuration of Karmada control plane components beyond the scope of adding extra volumes and volume mounts.
+
+
+## Proposal
+Introduce the new optional fields `ExtraVolumes` and `ExtraVolumeMounts` within Karmada control plane component configurations.
```
RainbowMango, line 38: Could you please make a list of components that would introduce the fields? That would help to make clear the scope of this proposal.

docs/proposals/karmada-components-extra-volumes/README.md
```diff
@@ -0,0 +1,58 @@
...
+Let's take the API server component as an example. With these new fields, the configuration would look as follows:
+```go
+// KarmadaAPIServer holds settings to kube-apiserver component of the kubernetes.
+// Karmada uses it as its own apiserver in order to provide Kubernetes-native APIs.
+type KarmadaAPIServer struct {
+	// Other existing fields
+
+	// ExtraVolumes specifies a list of extra volumes for the API server's pod
+	// +optional
+	ExtraVolumes []corev1.Volume `json:"extraVolumes,omitempty"`
```
RainbowMango, line 52: There are a lot of volume source types. We might need to declare which are supported.

docs/proposals/karmada-components-extra-volumes/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Summary
+
+This proposal aims to introduce the ability to specify extra volumes and volume mounts for Karmada control plane components when creating a Karmada instance managed by the Karmada operator. 
+This enhancement will extend the current configuration capabilities, allowing users to meet additional requirements such as having the ability to configure encryption at rest or to configure
+custom authentication webhooks.
+
+## Motivation
+Currently, the Karmada operator allows specifying extra arguments for the Karmada control plane components, but lacks support for specifying extra volumes and volume mounts for those components. 
+This limitation hinders certain use cases that require additional configurations for security and customization. For instance, as of today, it's not possible to configure a custom authentication
```
RainbowMango, line 25: Could you please describe the use case in more detail? How this volumes will be used?

jabellard:
@RainbowMango , thanks for the constructive feedback. I just pushed some changes to address your comments, so I think this should be ready to go. Currently working on the implementation PR and will push that soon.

RainbowMango (commented):
/assign

docs/proposals/karmada-operator/api-server-extra-volumes-volume-mounts/README.md
```diff
@@ -0,0 +1,63 @@
...
+type KarmadaAPIServer struct {
+	// Other existing fields
+
+	// ExtraVolumes specifies a list of extra volumes for the API server's pod
+	// +optional
+	ExtraVolumes []corev1.Volume `json:"extraVolumes,omitempty"`
+
+	// ExtraVolumeMounts specifies a list of extra volume mounts to be mounted into the API server's container
+	// +optional
+	ExtraVolumeMounts []corev1.VolumeMount `json:"extraVolumeMounts,omitempty"`
```
RainbowMango, line 61: Hi @jabellard I'm looking at the field names, both fields are prefixed with `Extra`.

I can understand why you named it with `extra`, that is we already have some hardcoded volumes and mounts, like:
```yaml
  volumes:
  - name: apiserver-cert
    secret:
      defaultMode: 420
      secretName: karmada-demo-cert
  - name: etcd-cert
    secret:
      defaultMode: 420
      secretName: karmada-demo-etcd-cert
```
and 
```yaml
    volumeMounts:
    - mountPath: /etc/karmada/pki
      name: apiserver-cert
      readOnly: true
    - mountPath: /etc/etcd/pki
      name: etcd-cert
      readOnly: true
```

I have a feeling that a little bit of context is missing here.

Maybe we should go without `extra` or add more comments on that. What do you think?

RainbowMango: By the way, we have the same feeling on [ExtraArgs](https://github.com/karmada-io/karmada/blob/f2f57c19da9db728a72b5ef50107baf676a69fdf/operator/pkg/apis/operator/v1alpha1/type.go#L291) just on above of the proposed fields. From the users's perspective, they might not know what kind of args are already set.

jabellard: Thanks for sharing that perspective. I do agree with you that a user creating the CR will lack context into what will be set by default if it's not documented. For volumes and volumes mounts, by default, the code will add volumes and volume mounts for certs as that's a base requirement for a functioning control plane. As you mentioned above, we can remediate the lack of context from a user's perspective by adding more documentation to clearly communicate what is automatically handled by the operator for a base setup, and how those new fields can be used the further customize that setup.  Do you see that as sensible way forward?

RainbowMango: Yes, I agree, I guess you mean adding more documentation is more comments on that.

jabellard: Yes. Just pushed some changes for that.

RainbowMango (approved):
/lgtm
/approve
Thanks.

Issue #5235, opened 2024-07-23
[lfx-mentorship-2024-September-November] Enhance Karmada controller-manager and schedule testing coverage


XiShanYongYe-Chang:
**CNCF LFX mentorship**
https://github.com/cncf/mentoring/blob/main/programs/lfx-mentorship/2024/03-Sep-Nov/README.md

Mentor: @XiShanYongYe-Chang 

**What would you like to be added:**

Karmada would like to improve the UT coverage of the code to better maintain the quality of the code and reduce the introduction of defects.

Increase the UT coverage rate to 50% to 60% (currently, the UT coverage rate is [28.26%](https://app.codecov.io/gh/karmada-io/karmada) ).

The entire Karmada repository is a bit large for one project, so we will split it into two projects. The current project needs to improve UT coverage in the following directory:

```
pkg/controllers
pkg/dependenciesdistributor
pkg/descheduler
pkg/detector
pkg/estimator
pkg/scheduler
pkg/resourceinterpreter
pkg/util
```

The above directories mainly contain the `karmada-controller-manager` and `karmada-scheduler` components of Karmada.

I hope that the students who apply for this project can give the UT coverage rate of each directory plan to achieve, we will move towards these specific goals.

**Why is this needed:**

yashpandey06:
Hey @XiShanYongYe-Chang ,hey looks like a great learning ground for me ,I would like to contribute to this issue under LFX Mentorship program.

Harsh4902:
@XiShanYongYe-Chang I want to work on this. Can you provide some pretests?

XiShanYongYe-Chang:
Just for reference: We can get the code coverage of different files under the current master branch from [here](https://app.codecov.io/gh/karmada-io/karmada/tree/master/).

mani1911:
Hey @XiShanYongYe-Chang this seems apt for me to learn and I would love to contribute to this

XiShanYongYe-Chang:
[...]

PR #5302, opened 2024-08-05
Add 100% unit test coverage for annotation and label in pkg/util

NishantBansal2003:
**What type of PR is this?**
/kind failing-test

The test case coverage for annotation.go has been increased from 73.9% to 100% and for label.go has been increased from 97.8% to 100% . 

Ref #5235 

To verify the changes in the pkg/util directory run the following commands:
```
go test ./... -coverprofile=coverage.out
go tool cover -html=coverage.out -o coverage.html
open coverage.html
```

```release-note
NONE
```

### DIFF
--- a/pkg/util/annotation_test.go
+++ b/pkg/util/annotation_test.go
@@ -288,7 +288,7 @@ func TestRecordManagedAnnotations(t *testing.T) {
 			},
 		},
 		{
-			name: "object has has annotations",
+			name: "object has annotations",
 			object: &unstructured.Unstructured{
 				Object: map[string]interface{}{
 					"apiVersion": "apps/v1",
@@ -322,7 +322,7 @@ func TestRecordManagedAnnotations(t *testing.T) {
 			},
 		},
 		{
-			name: "object has has annotations and labels",
+			name: "object has annotations and labels",
 			object: &unstructured.Unstructured{
 				Object: map[string]interface{}{
 					"apiVersion": "apps/v1",
@@ -331,6 +331,47 @@ func TestRecordManagedAnnotations(t *testing.T) {
 						"name": "demo-deployment-1",
 						"annotations": map[string]interface{}{
 							"foo": "foo",
+						},
+						"labels": map[string]interface{}{
+							"bar": "bar",
+						},
+					},
+					"spec": map[string]interface{}{
+						"replicas": 2,
+					},
+				},
+			},
+			expected: &unstructured.Unstructured{
+				Object: map[string]interface{}{
+					"apiVersion": "apps/v1",
+					"kind":       "Deployment",
+					"metadata": map[string]interface{}{
+						"name": "demo-deployment-1",
+						"annotations": map[string]interface{}{
+							workv1alpha2.ManagedAnnotation: "foo,resourcetemplate.karmada.io/managed-annotations,resourcetemplate.karmada.io/managed-labels",
+							"foo":                          "foo",
+						},
+						"labels": map[string]interface{}{
+							"bar": "bar",
+						},
+					},
+					"spec": map[string]interface{}{
+						"replicas": 2,
+					},
+				},
+			},
+		},
+		{
+			name: "object has recorded annotations and labels",
+			object: &unstructured.Unstructured{
+				Object: map[string]interface{}{
+					"apiVersion": "apps/v1",
+					"kind":       "Deployment",
+					"metadata": map[string]interface{}{
+						"name": "demo-deployment-1",
+						"annotations": map[string]interface{}{
+							workv1alpha2.ManagedAnnotation: "foo,resourcetemplate.karmada.io/managed-annotations,resourcetemplate.karmada.io/managed-labels",
+							"foo":                          "foo",
 						},
 						"labels": map[string]interface{}{
 							"bar": "bar",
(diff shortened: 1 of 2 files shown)


NishantBansal2003:
Dear @XiShanYongYe-Chang,
I have added test coverage to `pkg/utils/annotation.go` and `pkg/utils/label.go` to better familiarize myself with Karmada's codebase. I would appreciate it if you could review the changes. I am open to any suggestions or modifications you may have.
Thanks

XiShanYongYe-Chang:
Thanks @NishantBansal2003 
/assign

pkg/util/label_test.go
```diff
@@ -249,7 +281,9 @@ func TestRemoveLabel(t *testing.T) {
 						},
 						"spec": map[string]interface{}{
 							"replicas": 2,
-						}}},
```
XiShanYongYe-Chang, line 252: The original writing is a style and, I feel it do not need be modified.

pkg/util/annotation_test.go
```diff
@@ -361,6 +361,47 @@ func TestRecordManagedAnnotations(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "object has has annotations and labels",
```
XiShanYongYe-Chang, line 365: It has the same case as the above case, how can we distinguish them?

pkg/util/annotation_test.go
```diff
@@ -425,3 +466,223 @@ func TestMergeAnnotation(t *testing.T) {
...
+			existAnnotation: nil,
+			newAnnotation: map[string]string{
+				"newAnnotationKey": "newAnnotationValues",
+			},
+			expectedAnnotation: map[string]string{
+				"newAnnotationKey": "newAnnotationValues",
+			},
+		},
+		{
+			name: "Nil  New Annotation",
```
XiShanYongYe-Chang, line 494: An extra space was found

pkg/util/annotation_test.go
```diff
@@ -425,3 +466,223 @@ func TestMergeAnnotation(t *testing.T) {
...
+
+func TestDedupeAndMergeAnnotations(t *testing.T) {
+	tests := []struct {
+		name               string
+		existAnnotation    map[string]string
+		newAnnotation      map[string]string
+		expectedAnnotation map[string]string
+	}{
+		{
+			name:               "Nil both Annotation",
```
XiShanYongYe-Chang, line 478: The case name can be named in the same style as the existing case. For example: `nil both annotation.`

XiShanYongYe-Chang (commented):
Thanks

pkg/util/annotation_test.go
NishantBansal2003: Sure I will update with suitable case, also in the other name "has" is 2 times I will rectify that also...

pkg/util/label_test.go
NishantBansal2003: But in some positions, it was different -
https://github.com/karmada-io/karmada/blob/b4b6d692ca963c3d53de4a61077f4380613be0a5/pkg/util/label_test.go#L161-L166
which is why I changed it. Should I change this back to the original?

XiShanYongYe-Chang: I think it's fine to keep the status, it's not something that needs to be corrected.

pkg/util/annotation_test.go
```diff
@@ -288,7 +288,7 @@ func TestRecordManagedAnnotations(t *testing.T) {
 			},
 		},
 		{
-			name: "object has has annotations",
```
XiShanYongYe-Chang, line 291: Good catch.

pkg/util/label_test.go
```diff
@@ -317,12 +359,18 @@ func TestRemoveLabel(t *testing.T) {
...
 		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			RemoveLabels(tt.args.obj, tt.args.labelKey)
+			if tt.args.labelKey == "" {
+				RemoveLabels(tt.args.obj)
+			} else {
+				RemoveLabels(tt.args.obj, tt.args.labelKey)
+			}
```
XiShanYongYe-Chang, line 373: You've given me a good reminder. I feel that the previous test method is not good. Maybe we can change `labelKey` to `labelKeys`, so that we can test `labelKeys` of different lengths.

Like this:

```go
RemoveLabels(tt.args.obj, tt,args.labelKeys...)
```

pkg/util/annotation_test.go
```diff
@@ -425,3 +466,223 @@ func TestMergeAnnotation(t *testing.T) {
...
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if tt.args.key == "" {
+				RemoveAnnotations(tt.args.obj)
+			} else {
+				RemoveAnnotations(tt.args.obj, tt.args.key)
+			}
```
XiShanYongYe-Chang, line 682: Same comment with https://github.com/karmada-io/karmada/pull/5302#discussion_r1706682168

NishantBansal2003:
@XiShanYongYe-Chang  PTAL...

XiShanYongYe-Chang (commented):
Good job @NishantBansal2003 
/lgtm
/approve

Issue #5150, opened 2024-07-06
Improve the user experience of Failover


XiShanYongYe-Chang:
**What would you like to be added**:

Improve the user experience of Failover

**Why is this needed**:

The `Failover` and `GracefulEviction` features are currently in the Beta phase, which means they are enabled by default. 

There is a scenario where users propagate configuration resources by directly specifying the cluster names. When a cluster is disconnected from the Karmada control plane for several hours, it is identified as `NotReady`. Once the cluster recovers, the configuration resources on that cluster are deleted unexpectedly. If this occurs in a production environment, it could lead to serious consequences. 

Therefore, we need to optimize the Failover feature for this scenario to provide users with a more stable and reliable experience.

whitewindmills:
are you saying that resources on unhealthy member clusters will be deleted cause they're migrated to other member clusters? if so, how to improve it?

XiShanYongYe-Chang:
> are you saying that resources on unhealthy member clusters will be deleted cause they're migrated to other member clusters? 

That's it. One thing to note is that it's not migrated to another cluster, it's just moved out of the failed cluster.

>  if so, how to improve it?

I hope to hear everyone's opinion.

whitewindmills:
> One thing to note is that it's not migrated to another cluster, it's just moved out of the failed cluster.

let me guest how it happened, no fit cluster?

XiShanYongYe-Chang:
In other words, the clusters that need to be distributed have been listed, and these clusters will be distributed with configuration resources.

whitewindmills:
does it look like this? if the cluster `foo` becomes unhealthy, the configuration will be stuck in the cluster until the cluster becomes healthy and then it is deleted. am I right?
```yaml
apiVersion: policy.karmada.io/v1alpha1
kind: PropagationPolicy
metadata:
  name: test-pp
  namespace: default
spec:
  resourceSelectors:
    - apiVersion: v1
      kind: ConfigMap
      name: conf
  placement:
    clusterAffinity:
      clusterNames:
      - foo
      - bar
  ...

```

XiShanYongYe-Chang:
Yes, you are right.

whitewindmills:
yes, this is a noteworthy case where we would prefer not to delete resources when there is no new cluster to migrate to.

NickYadance:
we just had an issue that when a cluster becomes not ready for only about 2minutes, karmada still deleting all the resources in the unhealthy cluster. is this expected in the current version ?

[...]

PR #5305, opened 2024-08-06
fix controller can't restart in helm for dependent secret not found

### LINKED ISSUE #5233: karmada-controller-manager can't restart in helm installation for dependent secret not found
**What happened**:

In helm installation method, when installing `karmada-controller-manager`, we use a `initContainer` to wait for the ready status of `karmada-apiserver`, which prevents the `karmada-controller-manager` from `CrashLoopBack`. This feature is introduced in #5010.

In order to access host cluster kube-apiserver in `initContainer`, we mounted a `service-account-token` type `Secret`, because the deployment of `karmada-controller-manager` is defined `automountServiceAccountToken: false`. Unset `automountServiceAccountToken` is introduced in #2523. 

However, in #5010, we deleted the `Secret` mentioned above when we finished installation. Actually, we still need this secret after installation finished, otherwise `karmada-controller-manager` can't restart for lack of the mounted secret.

Now, we have two ways to resolve it:

* **Method one**: after the installation is complete, the secret remains.

> But I am not clear about is this reserved secret sensitive information? Is there any security issue?

* **Method two**:  no longer manually mount the `Secret` but directly set `automountServiceAccountToken`.

In #2523, we unset `automountServiceAccountToken` because we think `karmada-controller-manager` has no need to interact with host cluster, but now it seems that it needs to interact with the host cluster. So, maybe we can reset `automountServiceAccountToken`.

1. install karmada by `helm` method (the latest version)
2. delete the pod of `karmada-controller-manager`
3. check whether the pod of `karmada-controller-manager` ready

**Environment**:
- Karmada version: latest
- kubectl-karmada or karmadactl version (the result of `kubectl-karmada version` or `karmadactl version`): latest
- Others:

chaosi-zju:
**What type of PR is this?**

/kind bug

fix controller can't restart in helm for dependent secret not found.

In helm installation method, when installing `karmada-controller-manager`, we use a `initContainer` to wait for the ready status of `karmada-apiserver`, which prevents the `karmada-controller-manager` from `CrashLoopBack`. This feature is introduced in #5010.

In order to access host cluster kube-apiserver in `initContainer`, we mounted a `service-account-token` type `Secret`, because the deployment of `karmada-controller-manager` is defined `automountServiceAccountToken: false`. Unset `automountServiceAccountToken` is introduced in #2523. 

However, in #5010, we deleted the `Secret` mentioned above when we finished installation. Actually, we still need this secret after installation finished, otherwise `karmada-controller-manager` can't restart for lack of the mounted secret.

Fixes #5233

target installation order in helm after the PR:

1. deploy `etcd`
2. deploy `karmada-apiserver` (it has a init-container, it keeps checking etcd connectivity with the `curl` command, waiting for etcd to be ready)
3. deploy `Job/karmada-static-resource`, which is used to deploy static resources such as crd (it use `kubectl rollout status` command to wait for karmada-apiserver readay), when finished, it writes a configmap to karmada-apiserver.
4. deploy other components (they each have a int-container, using `kubectl get` command to wait for above configmap exist, which means `job/karmada-static-resource` has finished applying current version crds to apiserver)

```release-note
fix controller can't restart in helm for dependent secret not found
```

### DIFF
--- a/charts/karmada/templates/_helpers.tpl
+++ b/charts/karmada/templates/_helpers.tpl
@@ -586,40 +586,31 @@ Return the proper Docker Image Registry Secret Names
 {{- end }}
 {{- end -}}
 
-{{- define "karmada.init-sa-secret.volume" -}}
-{{- $name := include "karmada.name" . -}}
-- name: init-sa-secret
-  secret:
-    secretName: {{ $name }}-hook-job
-{{- end -}}
-
-{{- define "karmada.init-sa-secret.volumeMount" -}}
-- name: init-sa-secret
-  mountPath: /opt/mount
-{{- end -}}
-
-{{- define "karmada.initContainer.build-kubeconfig" -}}
-TOKEN=$(cat /opt/mount/token)
-kubectl config set-cluster karmada-host --server=https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT} --certificate-authority=/opt/mount/ca.crt
-kubectl config set-credentials default --token=$TOKEN
-kubectl config set-context karmada-host-context --cluster=karmada-host --user=default --namespace=default
-kubectl config use-context karmada-host-context
-{{- end -}}
-
 {{- define "karmada.initContainer.waitEtcd" -}}
 - name: wait
-  image: {{ include "karmada.kubectl.image" . }}
+  image: {{ include "karmada.cfssl.image" . }}
   imagePullPolicy: {{ .Values.kubectl.image.pullPolicy }}
   command:
     - /bin/sh
     - -c
     - |
       bash <<'EOF'
-      {{- include "karmada.initContainer.build-kubeconfig" . | nindent 6 }}
-      kubectl rollout status statefulset etcd -n {{ include "karmada.namespace" . }}
+      set -ex
+      while true; do
+        ETCD_ENDPOINT=${ETCD_CLIENT_SERVICE_HOST}":"${ETCD_CLIENT_SERVICE_PORT}
+
+        # check etcd connectivity by executing curl.
+        # if etcd is ready, the response of curl would be `curl: (52) Empty reply from server`, with return code 52.
+        # if not, the response of curl would be like `curl: (7) Failed to connect to .....`, with other return code.
+        if curl --connect-timeout 2 ${ETCD_ENDPOINT} || [ $? -eq 52 ]; then
+          break
+        fi
+
+        echo "failed to connect to "${ETCD_ENDPOINT}
+        sleep 2
+      done
+      echo "successfully connect to "${ETCD_ENDPOINT}
       EOF
-  volumeMounts:
-    {{- include "karmada.init-sa-secret.volumeMount" .| nindent 4 }}
 {{- end -}}
 
 {{- define "karmada.initContainer.waitStaticResource" -}}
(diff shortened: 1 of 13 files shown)


XiShanYongYe-Chang:
Hi @chaosi-zju, this char lint has failed.

XiShanYongYe-Chang:
Ask @calvin0327 @zhzhuang-zju to help take a review~
/cc @calvin0327 @zhzhuang-zju

iawia002 (commented):
Basically LGTM

RainbowMango (commented):
/assign

charts/karmada/templates/_helpers.tpl
```diff
@@ -631,9 +622,18 @@ kubectl config use-context karmada-host-context
...
     - |
       bash <<'EOF'
-      {{- include "karmada.initContainer.build-kubeconfig" . | nindent 6 }}
-      kubectl wait --for=condition=complete job {{ include "karmada.name" . }}-static-resource -n {{ include "karmada.namespace" . }}
+      set -ex
+
+      # here are three cases:
+      # case first installation: no `cm/karmada-versions` at first, so when you get it, it means `karmada-static-resource-job` finished.
+      # case restart: already has `cm/karmada-versions`, which means `karmada-static-resource-job` already finished.
+      # case upgrading: already has `cm/karmada-versions`, but it may be old version, we should wait until `.data.karmadaImageVersion` euqal to current `.Values.karmadaImageVersion`.
```
RainbowMango, line 630:
```suggestion
      # case upgrading: already has `cm/karmada-versions`, but it may be old version, we should wait until `.data.karmadaImageVersion` equal to current `.Values.karmadaImageVersion`.
```

charts/karmada/templates/_helpers.tpl
```diff
@@ -631,9 +622,18 @@ kubectl config use-context karmada-host-context
...
+      while [[ $(kubectl --kubeconfig /etc/kubeconfig get configmap karmada-versions -n {{ .Values.systemNamespace }} -o jsonpath='{.data.karmadaImageVersion}') != {{ .Values.karmadaImageVersion }} ]]; do
+        echo "wait for karmada-static-resource-job finshed"; sleep 2
```
RainbowMango, line 632:
```suggestion
        echo "wait for karmada-static-resource-job finished"; sleep 2
```

charts/karmada/templates/karmada-static-resource-job.yaml
```diff
@@ -42,6 +42,17 @@ spec:
           kubectl apply -k /crds --kubeconfig /etc/kubeconfig
           kubectl apply -f /static-resources/system-namespace.yaml --kubeconfig /etc/kubeconfig
           kubectl apply -f /static-resources/ --kubeconfig /etc/kubeconfig
+
+          kubectl --kubeconfig /etc/kubeconfig apply -f - <<InnerEOF
+          apiVersion: v1
+          kind: ConfigMap
+          metadata:
+            name: karmada-versions
```
RainbowMango, line 50:
```suggestion
            name: karmada-version
```
Here just stores one version, right?

charts/karmada/templates/karmada-static-resource-job.yaml
```diff
@@ -42,6 +42,17 @@ spec:
...
+            namespace: {{ .Values.systemNamespace }}
+          data:
+            karmadaImageVersion: {{ .Values.karmadaImageVersion }}
```
RainbowMango, line 53:
```suggestion
            karmadaVersion: {{ .Values.karmadaImageVersion }}
```

RainbowMango (commented):
1. I just realized it's a mistake for initcontainer holds the secret of the host cluster.
2. I feels a little bit tricky to have a configmap to represent Karmada version, but I don't have a better idea.

Given this patch is an improvement, we can move forward even it is not the idea solution. 

This is much better than the solution in #5150. Thanks.

RainbowMango:
> However, I am worried that after deleting the file post-install-job.yaml, it will be troublesome to get this file back if we want to add some logic to this post-install hook point in the future.

Why can't get this file back?

RainbowMango (approved):
/lgtm
/approve

Anillc:
Where the `ETCD_CLIENT_SERVICE_HOST` environment variable is being set? I searched the entire repository but couldn't find it. I'm using etcd in external mode, but the `wait` container in karmada-apiserver is throwing the following error:

```
 curl: (3) URL using bad/illegal format or missing URL
 failed to connect to :
 + '[' 3 -eq 52 ']'
 + echo 'failed to connect to :'
 + sleep 2
 + true
 + ETCD_ENDPOINT=:
 + curl --connect-timeout 2 :
```

It looks like these two environment variables haven't been set.

PR #5317, opened 2024-08-07
work suspension: work status conditions

a7i:
**What type of PR is this?**
/kind feature

Fixes #5118

```release-note
`API change`: Introduced `Dispatching` condition to represent the dispatching status.
```

### DIFF
--- a/pkg/controllers/execution/execution_controller.go
+++ b/pkg/controllers/execution/execution_controller.go
@@ -30,7 +30,6 @@ import (
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
-	"k8s.io/utils/ptr"
 	controllerruntime "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -53,6 +52,10 @@ import (
 const (
 	// ControllerName is the controller name that will be used when reporting events.
 	ControllerName = "execution-controller"
+	// workSuspendDispatchingConditionReason is the reason for the WorkDispatching condition when dispatching is suspended.
+	workSuspendDispatchingConditionReason = "SuspendDispatching"
+	// workDispatchingConditionReason is the reason for the WorkDispatching condition when dispatching is not suspended.
+	workDispatchingConditionReason = "Dispatching"
 )
 
 // Controller is to sync Work.
@@ -94,7 +97,12 @@ func (c *Controller) Reconcile(ctx context.Context, req controllerruntime.Reques
 		return controllerruntime.Result{}, err
 	}
 
-	if ptr.Deref(work.Spec.SuspendDispatching, false) {
+	if err := c.updateWorkDispatchingConditionIfNeeded(ctx, work); err != nil {
+		klog.Errorf("Failed to update work condition type %s. err is %v", workv1alpha1.WorkDispatching, err)
+		return controllerruntime.Result{}, err
+	}
+
+	if helper.IsWorkSuspendDispatching(work) {
 		klog.V(4).Infof("Skip syncing work(%s/%s) for cluster(%s) as work dispatch is suspended.", work.Namespace, work.Name, cluster.Name)
 		return controllerruntime.Result{}, nil
 	}
@@ -209,7 +217,7 @@ func (c *Controller) syncToClusters(ctx context.Context, clusterName string, wor
 	if len(errs) > 0 {
 		total := len(work.Spec.Workload.Manifests)
 		message := fmt.Sprintf("Failed to apply all manifests (%d/%d): %s", syncSucceedNum, total, errors.NewAggregate(errs).Error())
-		err := c.updateAppliedCondition(work, metav1.ConditionFalse, "AppliedFailed", message)
+		err := c.updateAppliedCondition(ctx, work, metav1.ConditionFalse, "AppliedFailed", message)
 		if err != nil {
 			klog.Errorf("Failed to update applied status for given work %v, namespace is %v, err is %v", work.Name, work.Namespace, err)
 			errs = append(errs, err)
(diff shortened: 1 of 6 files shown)


pkg/controllers/status/work_status_controller.go
```diff
@@ -92,6 +99,11 @@ func (c *WorkStatusController) Reconcile(ctx context.Context, req controllerrunt
 		return controllerruntime.Result{}, err
 	}
 
+	c.updateWorkDispatchingCondition(ctx, work)
+	if helper.IsWorkSuspendDispatching(work) {
+		return controllerruntime.Result{}, nil
+	}
```
RainbowMango, line 105: The `SuspendDispatching` indicates whether the controller should dispatch workload to the member cluster, it does not block status collection.  The reason why naming this field instead of a simple `Suspend` is to leave room for future extensions, like some users might not want to collect status probably due to performance concerns, then we can introduce another field like `suspendStatus`.

So, in my opinion, we should remove this check.

pkg/controllers/status/work_status_controller.go
```diff
@@ -217,6 +229,10 @@ func (c *WorkStatusController) syncWorkStatus(key util.QueueKey) error {
 		return err
 	}
 
+	if helper.IsWorkSuspendDispatching(workObject) {
+		return nil
+	}
```
RainbowMango, line 234: ditto.

pkg/controllers/status/work_status_controller.go
a7i: I see. So we might as well add the condition in execution controller?

a7i: Pushed up a commit, let me know what you think!

pkg/controllers/execution/execution_controller.go
```diff
@@ -94,7 +97,12 @@ func (c *Controller) Reconcile(ctx context.Context, req controllerruntime.Reques
 		return controllerruntime.Result{}, err
 	}
 
-	if ptr.Deref(work.Spec.SuspendDispatching, false) {
+	if err := c.updateWorkDispatchingConditionIfNeeded(ctx, work); err != nil {
+		klog.Errorf("Failed to update work condition type %s. err is %v", workv1alpha1.WorkDispatching, err)
+		return controllerruntime.Result{}, err
+
```
XiShanYongYe-Chang, line 103: The lint failed for this empty line~

pkg/controllers/execution/execution_controller.go
```diff
@@ -253,8 +261,31 @@ func (c *Controller) tryCreateOrUpdateWorkload(ctx context.Context, clusterName
 	return nil
 }
 
+func (c *Controller) updateWorkDispatchingConditionIfNeeded(ctx context.Context, work *workv1alpha1.Work) error {
```
chaunceyjiang, line 264: Just a question: In addition to setting `Condition`, should we also send some `event`s?

pkg/controllers/status/work_status_controller.go
RainbowMango: > I see. So we might as well add the condition in execution controller?

Oh, yes. Thanks for the reminder, I didn't notice it yet.

pkg/controllers/execution/execution_controller.go
RainbowMango: Even if there is an event, that's the `Normal` type(Not a warning) I don't know if this event is desired.
I guess @XiShanYongYe-Chang can help track this on the umbrella issue, we can add it by another PR if there is a need.

RainbowMango (commented):
/lgtm

@XiShanYongYe-Chang @CharlesQQ Do you want to take another look?

pkg/controllers/execution/execution_controller.go
XiShanYongYe-Chang: @chaunceyjiang has mentioned adding an event at the community meeting, it's something, we can do it,  

let me track it in the umbrella issue, and do it in another pr.

XiShanYongYe-Chang (commented):
Thanks
/lgtm

pkg/controllers/execution/execution_controller.go
a7i: https://github.com/karmada-io/karmada/pull/5332

chaunceyjiang (commented):
/lgtm

pkg/controllers/execution/execution_controller.go
```diff
@@ -253,8 +261,31 @@ func (c *Controller) tryCreateOrUpdateWorkload(ctx context.Context, clusterName
...
+	newWorkDispatchingCondition := metav1.Condition{
+		Type:               workv1alpha1.WorkDispatching,
+		LastTransitionTime: metav1.Now(),
+	}
+
+	if helper.IsWorkSuspendDispatching(work) {
+		newWorkDispatchingCondition.Status = metav1.ConditionFalse
+		newWorkDispatchingCondition.Reason = workSuspendDispatchingConditionReason
+		newWorkDispatchingCondition.Message = "Work dispatching is in a suspended state."
```
chaunceyjiang, line 273: Should we record the names of the member clusters here?

XiShanYongYe-Chang: The cluster information may be obtained from the namespace of work.

chaunceyjiang: > The cluster information may be obtained from the namespace of work.

I know what you mean - if the cluster name is also present in the condition, I think it would be more intuitive.

XiShanYongYe-Chang: You have a point.
Hi @a7i, would you be willing to change it, or shall we do it separately in the next pr.

RainbowMango: Let's see if we need the cluster info in the message.

This kind of engineering fly trap, `this could be useful`. 
 In my opinion, the cluster name might be redundant. I think we can postpone it until there is more clear evidence.

RainbowMango (approved):
/approve

PR #5318, opened 2024-08-07
fix expected patch operations may be missed when AggregateStatus

### LINKED ISSUE #5285: Expected patch operations may be missed when AggregateStatus
**Background**:

Previously, we use `Patch` method to update resource template status, it is introduced in [use Patch() instead of Update() when updating workload status because error rate is too high](https://github.com/karmada-io/karmada/pull/4094).

Yet, `Patch` has two forms, that is `JSONPatchType` and `MergePatchType`, we once choose `JSONPatchType` for the reason of https://github.com/karmada-io/karmada/issues/4093#issuecomment-1768004382.

**What happened**:

When updating resource template status in `AggregateStatus` process, we get `oldStatus` from **cache** and send patch request directly to apiserver, so, here may be such extreme cases:

1. in first `Reconcile`, we got new status `{xx-field: 4}`, and finished the patch request, **supposing the cache is also updated to `{xx-field: 4}`**.
2. in second `Reconcile`, we got old status `{xx-field: 4}` and new status `{xx-field: 2}`, and finished the patch request to apiserver, however, **supposing now haven't had time to update to cache yet**.
3. in third `Reconcile`, we got old status from cache, that is still `{xx-field: 4}`, **supposing the new status is also `{xx-field: 4}`**, here is no diff filed between the old and new, so we got empty `JSONPatchType` patchBytes and keep `xx-field` unchanged.

Finally, the status is `{xx-field: 2}`, but we expect it to be `{xx-field: 4}`.

we expect patch operations not be missed when `AggregateStatus`, especially if the cache update is not timely.

Appear occasionally, but it is difficult to reproduce.

* this problem appeared twice in latest week.
* related failed CI log: https://github.com/karmada-io/karmada/issues/5136#issuecomment-2255391438

**Environment**:
- Karmada version: latest
- kubectl-karmada or karmadactl version (the result of `kubectl-karmada version` or `karmadactl version`): latest
- Others:

chaosi-zju:
**What type of PR is this?**

/kind bug

fix expected patch operations may be missed when AggregateStatus, this PR includes:

* reduce the probability of update conflicts by avoiding read data from the client cache.
* make the operations to `Get` and `Update` resources template as close as possible.
* replace `Patch` with `Update` when update resource status.

Fixes #5285

```release-note
`karmada-controller-manager`: Fixed the issue that status aggregation against the resource template might be missed due to slow cache sync.
```


chaosi-zju:
I will fix the UT failure later as soon.

pkg/controllers/status/common.go
```diff
@@ -105,43 +106,53 @@ func updateResourceStatus(
...
-	patchBytes, err := helper.GenReplaceFieldJSONPatch("/status", oldStatus, newStatus)
-	if err != nil {
-		klog.Errorf("Failed to gen patch bytes for resource(%s/%s/%s, Error: %v", gvr, resource.GetNamespace(), resource.GetName(), err)
-		return err
-	}
-
-	_, err = dynamicClient.Resource(gvr).Namespace(resource.GetNamespace()).
-		Patch(ctx, resource.GetName(), types.JSONPatchType, patchBytes, metav1.PatchOptions{}, "status")
-	if err != nil {
+	if _, err = dynamicClient.Resource(gvr).Namespace(resource.GetNamespace()).UpdateStatus(ctx, newObj, metav1.UpdateOptions{}); err != nil {
```
chaunceyjiang, line 149: I'm not very clear, why would using update be less prone to conflicts than using patch?

chaosi-zju:
> I will fix the UT failure later as soon.

the ut failure rely on the PR #5324

pkg/controllers/status/common.go
chaosi-zju: Previously, the root cause of conflicts is that we get RT (resource template) from cache, and that cache may not be updated in time (Now we get RT directly from apiserver to reduce the probability conflicts).

The number of conflicts has nothing to do with using `update` or `patch`, the difference is only that `patch` ignore conflict, while `update` will retry on conflict.

I replace `patch` with `update` is just I think `update` is more elegant, for `patch` will let apiserver always write etcd, while `update` will avoid write when `resourceVersion` is equal.

whitewindmills, line 149: use `retry.RetryOnConflict`?

chaosi-zju: hi @whitewindmills. do you mean you prefer following code?

```go
var resource *unstructured.Unstructured
if err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
	// get resource template directly from apiserver, rather than informer cache, detail refers to:
	// https://github.com/karmada-io/karmada/issues/5285
	resource, err = dynamicClient.Resource(gvr).Namespace(objRef.Namespace).Get(ctx, objRef.Name, metav1.GetOptions{})
	if err != nil {
			// It might happen when the resource template has been removed but the garbage collector hasn't removed
			// the ResourceBinding which dependent on resource template.
			// So, just return without retry(requeue) would save unnecessary loop.
			return nil
		}
		klog.Errorf("Failed to fetch resource(%s/%s) for resourceBinding. Error: %v.", resource.GetNamespace(), resource.GetName(), err)
		return err
	}

	newObj, err := interpreter.AggregateStatus(resource, bindingStatus.AggregatedStatus)
	if err != nil {
		klog.Errorf("Failed to aggregate status for resource(%s/%s/%s), Error: %v", gvr, resource.GetNamespace(), resource.GetName(), err)
		return err
	}

	if reflect.DeepEqual(resource, newObj) {
		klog.V(3).Infof("Ignore update resource(%s/%s/%s) status as up to date.", gvr, resource.GetNamespace(), resource.GetName())
		return nil
	}

	if _, err = dynamicClient.Resource(gvr).Namespace(resource.GetNamespace()).UpdateStatus(ctx, newObj, metav1.UpdateOptions{}); err != nil {
		klog.Errorf("Failed to update resource(%s/%s/%s), Error: %v", gvr, resource.GetNamespace(), resource.GetName(), err)
		return err
	}
	return nil
}); err != nil {
	eventRecorder.Event(resource, corev1.EventTypeWarning, events.EventReasonAggregateStatusFailed, err.Error())
	return err
}

eventRecorder.Event(resource, corev1.EventTypeNormal, events.EventReasonAggregateStatusSucceed, "Update Resource with AggregatedStatus successfully.")
klog.V(3).Infof("Update resource(%s/%s/%s) status successfully.", gvr, resource.GetNamespace(), resource.GetName())
```

whitewindmills: not exactly. `retry.RetryOnConflict` should start after `interpreter.AggregateStatus` because `interpreter.AggregateStatus` may be very time consuming, which causes `resource` to expire.

pkg/controllers/status/common.go
```diff
@@ -105,43 +106,53 @@ func updateResourceStatus(
...
 	if err != nil {
-		klog.Errorf("Failed to aggregate status for resource(%s/%s/%s, Error: %v", gvr, resource.GetNamespace(), resource.GetName(), err)
+		klog.Errorf("Failed to aggregate status for resource(%s/%s/%s), Error: %v", gvr, resource.GetNamespace(), resource.GetName(), err)
 		return err
 	}
 
-	oldStatus, _, _ := unstructured.NestedFieldNoCopy(resource.Object, "status")
-	newStatus, _, _ := unstructured.NestedFieldNoCopy(newObj.Object, "status")
-	if reflect.DeepEqual(oldStatus, newStatus) {
+	if reflect.DeepEqual(resource, newObj) {
```
whitewindmills, line 144: why do you make this change? it is indeed cheaper to just compare the status, and this can eliminate the interference of irrelevant fields.

pkg/util/helper/workstatus.go
```diff
@@ -246,18 +240,37 @@ func GetManifestIndex(manifests []workv1alpha1.Manifest, clusterObj *unstructure
 	return -1, fmt.Errorf("no such manifest exist")
 }
 
-func equalIdentifier(targetIdentifier *workv1alpha1.ResourceIdentifier, ordinal int, workload *unstructured.Unstructured) (bool, error) {
-	groupVersion, err := schema.ParseGroupVersion(workload.GetAPIVersion())
+// getManifestIndex gets the index of clusterObj in manifest list, if not exist return -1.
+func getManifestIndex(manifests []workv1alpha1.Manifest, objRef workv1alpha2.ObjectReference) (int, error) {
```
whitewindmills, line 244: I don't recommend adding such a tool function, it seems very redundant.

pkg/controllers/status/common.go
whitewindmills: it is recommended to use `helper.UpdateStatus` to update the status.
https://github.com/karmada-io/karmada/blob/5a10d758280c51a8ffe7a279816b08cbd0df14cb/pkg/util/helper/workstatus.go#L78-L85

pkg/controllers/status/common.go
chaosi-zju: These lines are a rollback to the previous code, but what you're saying makes more sense

pkg/util/helper/workstatus.go
chaosi-zju: but `GetManifestIndex` is also used by `work_status_controller` and has different `function parameter`

pkg/controllers/status/common.go
chaosi-zju: Supposing I have a CRD `XxxDeployment`, its status like: {"a": 1, "b": 1}

filed `b` is controller by resource interpreter `AggregatedStatus`, and is changed from 1 to 2: {"a": 1, "b": 2}

when I run to `UpdateStatus` function, here happens:

field `a` is controlled by my customized controller, and is change from 1 to 2, {"a": 2, "b": 1}

here conflicts occurs, the `XxxDeployment` is out-dated, I should get latest `XxxDeployment` and re-run `AggregatedStatus`.

What do you think of this thing, since `AggregateStatus(oldResource, aggregatedStatusItems)` is deifned by users, we cannot assume that all types of resources are completely covered by latest binding.

pkg/util/helper/workstatus.go
whitewindmills: maybe such arg `objRef workv1alpha2.ObjectReference` is not wanted.

pkg/controllers/status/common.go
whitewindmills: > field a is controlled by my customized controller, and is change from 1 to 2, {"a": 2, "b": 1}

I don't understand why this scenario exists, and I have never seen such usage. since the user chooses to use the `AggregatedStatus`, the control of the status should be fully handed over to `AggregatedStatus`. so I think your worries are unnecessary.

chaosi-zju: What I want to express is, resource interpreter `AggregateStatus(oldResource, aggregatedStatusItems)` is controlled by two args `oldResource` and `aggregatedStatusItems`, when conflicts, it means `oldResource` is updated, then the output of `AggregateStatus` may be different.

According to your statement, it seems to be assuming that output depends entirely on the `aggregatedStatusItems`, then this resource interpreter shouldn't have two args.

Supposing I have a CRD `XxxDeployment`, its status like: {"a": 1, "b": 1}, however, the field `status.a`  is rely on `spec.c` field, while `spec.c` field is updated by users...

whitewindmills: lol, so vexing. anyway, I'm sold.

> interpreter.AggregateStatus may be very time consuming, which causes resource to expire.

do you have any idea to resolve it?

chaosi-zju: /cc hi @XiShanYongYe-Chang, JunHua has the same point, what do you think now?

XiShanYongYe-Chang: I have been persuaded that it is better not to use retry.

chaosi-zju:
new failure found: https://github.com/karmada-io/karmada/pull/5338#issuecomment-2283008790

pkg/util/helper/workstatus.go
chaosi-zju: hi, can you check again? I I choose to reuse a new parameter.

pkg/controllers/status/common.go
chaosi-zju: hi, now, we get resource directly from apiserver and normally only this controller would change its status, so almost impossible conflict. even if there is a conflict, we can perform the reconcile again, no need to use `RetryOnConflict`.

whitewindmills: > and normally only this controller would change its status

I agree.

> so almost impossible conflict

you cannot guarantee that user controllers won't frequently change their resources, maybe `spec` or `metadata` things. anyway, we can make the update as efficient as possible with `retry.RetryOnConflict` .

> do we still need to stick on this comment?

no, it's up to you. even if you don't take my advice, it's acceptable to me.

chaosi-zju: I got another point, that is, we can't re-use `operationResult, err = UpdateStatus(ctx, c, binding, func() error` such code, since the inner client is `client.Client`, which re-fetch resource from cache, we should use `dynamicClient`.

using `retry.RetryOnConflict` is ok, but since there are no reusable `UpdateStatus` functions, the final code style may still be like this: https://github.com/karmada-io/karmada/pull/5318#discussion_r1709326892, then so that we don't have to trapped in whether need call `interpreter.AggregateStatus` once again.

pkg/util/helper/workstatus.go
```diff
@@ -227,37 +222,44 @@ func assembleWorkStatus(works []workv1alpha1.Work, workload *unstructured.Unstru
 	return statuses, nil
 }
 
+// ManifestObject identifies an object in manifest list
+type ManifestObject struct {
```
RainbowMango, line 226:
```suggestion
type ManifestReference struct {
```

pkg/controllers/status/common.go
```diff
@@ -105,43 +108,63 @@ func updateResourceStatus(
...
-	oldStatus, _, _ := unstructured.NestedFieldNoCopy(resource.Object, "status")
-	newStatus, _, _ := unstructured.NestedFieldNoCopy(newObj.Object, "status")
-	if reflect.DeepEqual(oldStatus, newStatus) {
-		klog.V(3).Infof("Ignore update resource(%s/%s/%s) status as up to date.", gvr, resource.GetNamespace(), resource.GetName())
-		return nil
-	}
+	resource := &unstructured.Unstructured{}
+	if err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		// get resource template directly from apiserver, rather than informer cache, detail refers to:
+		// https://github.com/karmada-io/karmada/issues/5285
```
RainbowMango, line 129:
```suggestion
		// Fetch resource template from karmada-apiserver instead of informer cache, to avoid retry due to
		// resource conflict which often happens, especially with a huge amount of resource templates and
		// the informer cache doesn't sync quickly enough.
		// For more details refer to https://github.com/karmada-io/karmada/issues/5285.
```

pkg/controllers/status/common.go
```diff
@@ -105,43 +108,63 @@ func updateResourceStatus(
...
+		resource, err = dynamicClient.Resource(gvr).Namespace(objRef.Namespace).Get(ctx, objRef.Name, metav1.GetOptions{})
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				// It might happen when the resource template has been removed but the garbage collector hasn't removed
+				// the ResourceBinding which dependent on resource template.
+				// So, just return without retry(requeue) would save unnecessary loop.
+				return nil
+			}
+			klog.Errorf("Failed to fetch resource(%s/%s) for resourceBinding. Error: %v.", resource.GetNamespace(), resource.GetName(), err)
```
RainbowMango, line 138:
```suggestion
			klog.Errorf("Failed to fetch resource template(%s/%s). Error: %v.", resource.GetNamespace(), resource.GetName(), err)
```

pkg/controllers/status/common.go
```diff
@@ -105,43 +108,63 @@ func updateResourceStatus(
...
+		}
 
-	patchBytes, err := helper.GenReplaceFieldJSONPatch("/status", oldStatus, newStatus)
-	if err != nil {
-		klog.Errorf("Failed to gen patch bytes for resource(%s/%s/%s, Error: %v", gvr, resource.GetNamespace(), resource.GetName(), err)
-		return err
-	}
+		newObj, err := interpreter.AggregateStatus(resource, bindingStatus.AggregatedStatus)
+		if err != nil {
+			klog.Errorf("Failed to aggregate status for resource(%s/%s/%s), Error: %v", gvr, resource.GetNamespace(), resource.GetName(), err)
```
RainbowMango, line 144:
```suggestion
			klog.Errorf("Failed to aggregate status for resource template(%s/%s/%s), Error: %v", gvr, resource.GetNamespace(), resource.GetName(), err)
```

pkg/controllers/status/common.go
RainbowMango: In addition, add gvr to the log.

pkg/controllers/status/common.go
```diff
@@ -105,43 +108,63 @@ func updateResourceStatus(
...
+				// It might happen when the resource template has been removed but the garbage collector hasn't removed
+				// the ResourceBinding which dependent on resource template.
+				// So, just return without retry(requeue) would save unnecessary loop.
+				return nil
+			}
+			klog.Errorf("Failed to fetch resource(%s/%s) for resourceBinding. Error: %v.", resource.GetNamespace(), resource.GetName(), err)
+			return err
+		}
 
-	patchBytes, err := helper.GenReplaceFieldJSONPatch("/status", oldStatus, newStatus)
```
RainbowMango, line 133: Why not remove the unused `GenReplaceFieldJSONPatch` function?

pkg/controllers/status/common.go
chaosi-zju: ok, this line I finally changed to: 

```go
klog.Errorf("Failed to fetch resource template(%s/%s/%s), Error: %v.", gvr, objRef.Namespace, objRef.Name, err)
```

I found that if err is not nil, that `resource` may be nil, so I use `objRef.Name` to print log.

pkg/util/helper/workstatus_test.go
```diff
@@ -192,7 +192,9 @@ func TestGetManifestIndex(t *testing.T) {
 	}
 
 	t.Run("Service", func(t *testing.T) {
-		index, err := GetManifestIndex(manifests, service)
+		manifestObj := ManifestReference{APIVersion: service.GetAPIVersion(), Kind: service.GetKind(),
```
RainbowMango, line 195: ditto.

RainbowMango (commented):
Just some nits, otherwise LGTM

pkg/controllers/status/work_status_controller.go
```diff
@@ -390,7 +390,9 @@ func (c *WorkStatusController) reflectStatus(ctx context.Context, work *workv1al
 }
 
 func (c *WorkStatusController) buildStatusIdentifier(work *workv1alpha1.Work, clusterObj *unstructured.Unstructured) (*workv1alpha1.ResourceIdentifier, error) {
-	ordinal, err := helper.GetManifestIndex(work.Spec.Workload.Manifests, clusterObj)
+	manifestObject := helper.ManifestReference{APIVersion: clusterObj.GetAPIVersion(), Kind: clusterObj.GetKind(),
```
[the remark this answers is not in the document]
chaosi-zju: you are right, pretty thanks

chaosi-zju:
hi, thanks everybody, do you have any further comments? If you have any other suggestions, please let me know~

RainbowMango:
cc @lxtywypc @whitewindmills @zach593 Any comments?

whitewindmills (commented):
/lgtm

zach593:
lgtm 🎉

RainbowMango (approved):
/approve
Glad that we reached an agreement. Thanks @whitewindmills @zach593 for the review, you input gives us confidence.

PR #5324, opened 2024-08-08
repair the test for rb status controller

chaosi-zju:
**What type of PR is this?**

/kind failing-test

repair the test for rb status controller:

In the UT of `RBStatusController`, here is a fake client, as:

```go
c.Client = fake.NewClientBuilder().WithScheme(gclient.NewSchema()).Build(),
```

the ut function wants to preset some resource by the fake client, such as:

```go
if err := c.Client.Create(context.Background(), binding); err != nil {
    t.Fatalf("Failed to create binding: %v", err)
}
```

then it wants to update the status of that resource, such as:

```go
if err := c.Client.Status().Update(context.Background(), binding); err != nil {
    t.Fatalf("Failed to update status of binding: %v", err)
}
```

Actually, the update function will failed with `Error: resourcebindings.work.karmada.io "binding" not found`.

the preset resource doesn't have subresource `status`, so update status will failed, a feasible way is like:

```go
c.Client = fake.NewClientBuilder().WithScheme(gclient.NewSchema()).WithObjects(binding).WithStatusSubresource(binding).Build()
```

Fixes #

```release-note

```

### DIFF
--- a/pkg/controllers/status/crb_status_controller_test.go
+++ b/pkg/controllers/status/crb_status_controller_test.go
@@ -34,6 +34,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 
 	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
+	"github.com/karmada-io/karmada/pkg/resourceinterpreter/default/native"
 	"github.com/karmada-io/karmada/pkg/util/fedinformer/genericmanager"
 	"github.com/karmada-io/karmada/pkg/util/gclient"
 )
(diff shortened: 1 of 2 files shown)


pkg/controllers/status/crb_status_controller_test.go
```diff
@@ -100,6 +101,7 @@ func TestCRBStatusController_Reconcile(t *testing.T) {
 				ObjectMeta: metav1.ObjectMeta{
 					Name:              "binding",
 					Namespace:         "default",
+					Finalizers:        []string{"test"},
```
chaosi-zju, line 104: the `Finalizers` field is required, otherwise the function `.WithObjects(binding).WithStatusSubresource(binding)` will failed with the error `refusing to create obj binding with metadata.deletionTimestamp but no finalizers`

XiShanYongYe-Chang: We can set it in the comment, which facilitates understanding during subsequent simulations.

pkg/controllers/status/rb_status_controller_test.go
```diff
@@ -219,3 +222,13 @@ func TestRBStatusController_syncBindingStatus(t *testing.T) {
...
+
+var _ resourceinterpreter.ResourceInterpreter = &FakeResourceInterpreter{}
+
+type FakeResourceInterpreter struct {
+	*native.DefaultInterpreter
+}
+
+func (f FakeResourceInterpreter) Start(_ context.Context) (err error) {
+	return nil
+}
```
XiShanYongYe-Chang, line 234: In the future, `FakeResourceInterpreter` can be placed in `pkg/resourceinterpreter`.

XiShanYongYe-Chang (commented):
Thanks~
/lgtm
/approve

PR #5216, opened 2024-07-18
skip the filter if the cluster is already in the list of scheduling result even if the API is missed

yanfeng1992:
Signed-off-by: huangyanfeng <huangyanfeng1992@gmail.com>

**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:

Because when obtaining APIEnablements, it may be empty or partial.
This will cause the scheduler to delete the scheduling results, resulting in incorrect deletion

So if the cluster is already in the list of scheduling results, even if the API is not hit, skip the filter

https://github.com/karmada-io/karmada/blob/a87ec2a61abadf9f102440b1f64cf338b2ee5518/pkg/controllers/status/cluster_status_controller.go#L233-L239

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #5325, opened 2024-08-08
avoid setting partial api enablements in cluster status

NickYadance:
**What type of PR is this?**
bug

avoid setting partial api enablements in cluster status

Fixes #5309

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/status/cluster_status_controller.go
+++ b/pkg/controllers/status/cluster_status_controller.go
@@ -18,6 +18,7 @@ package status
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"sort"
@@ -231,12 +232,16 @@ func (c *ClusterStatusController) setCurrentClusterStatus(clusterClient *util.Cl
 
 	// get the list of APIs installed in the member cluster
 	apiEnables, err := getAPIEnablements(clusterClient)
-	if len(apiEnables) == 0 {
-		klog.Errorf("Failed to get any APIs installed in Cluster %s. Error: %v.", cluster.GetName(), err)
-	} else if err != nil {
-		klog.Warningf("Maybe get partial(%d) APIs installed in Cluster %s. Error: %v.", len(apiEnables), cluster.GetName(), err)
+	if err != nil {
+		klog.Errorf("Failed to get APIs installed in Cluster %s. Error: %v.", cluster.GetName(), err)
 		return err
 	}
+
+	if len(apiEnables) == 0 && len(currentClusterStatus.APIEnablements) > 0 {
+		klog.Warningf("Unable to get any APIs installed in the existing Cluster %s, will retry", cluster.GetName())
+		return errors.New("empty APIs in cluster")
+	}
+
 	currentClusterStatus.APIEnablements = apiEnables
 
 	if c.EnableClusterResourceModeling {
--- a/pkg/controllers/status/cluster_status_controller.go
+++ b/pkg/controllers/status/cluster_status_controller.go
@@ -235,6 +235,7 @@ func (c *ClusterStatusController) setCurrentClusterStatus(clusterClient *util.Cl
 		klog.Errorf("Failed to get any APIs installed in Cluster %s. Error: %v.", cluster.GetName(), err)
 	} else if err != nil {
 		klog.Warningf("Maybe get partial(%d) APIs installed in Cluster %s. Error: %v.", len(apiEnables), cluster.GetName(), err)
+		return err
 	}
 	currentClusterStatus.APIEnablements = apiEnables
 
--- a/pkg/controllers/status/cluster_status_controller_test.go
+++ b/pkg/controllers/status/cluster_status_controller_test.go
@@ -1090,7 +1090,7 @@ func TestClusterStatusController_initLeaseController(_ *testing.T) {
 }
 
 func mockServer(statusCode int, existError bool) *httptest.Server {
-	respBody := "test"
+	respBody := "{}"
 	resp := &http.Response{
 		StatusCode: statusCode,
 		Body:       io.NopCloser(bytes.NewBufferString(respBody)),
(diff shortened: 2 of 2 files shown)


XiShanYongYe-Chang:
Thanks @NickYadance 
/assign

XiShanYongYe-Chang:
Hi @NickYadance, thanks~

Here, we do indeed need to continue updating the Cluster object instead of exiting directly, because there is a situation where, after a cluster has just been successfully created and joined Karmada, a certain API has not yet been successfully installed, but other types of APIs are functioning normally. If we skip it directly at this point, it will affect the distribution of normally functioning API types.


Currently, the processing logic here may not be very reasonable, and there are [issues](https://github.com/karmada-io/karmada/pull/5216) that reflect problems related to this. One of my ideas is to analyze the errors returned by the request to decide whether to update the Cluster status. I also have a strange thought: if the apienablements field in the cluster status is empty, then update it; otherwise, return an error.

whitewindmills:
indeed, updating APIENABLEMENTS is a very dangerous operation, which may cause the scheduler to delete the scheduled results. we're struggling with this lately. it caused us huge losses in the production environment.

NickYadance:
> I also have a strange thought: if the apienablements field in the cluster status is empty, then update it; otherwise, return an error
sounds indeed strange to me, to update aipenablements for only once.

it should be fine to exit on error here since it will requeue and succeed at sometime later. the thing is to not bring in-contact apienablements into the cluster status.

NickYadance:
> indeed, updating APIENABLEMENTS is a very dangerous operation, which may cause the scheduler to delete the scheduled results. we're struggling with this lately. it caused us huge losses in the production environment.

yes it's just too dangerous for karmada to draw massive deletions to the cluster. Extra handling in scheduler https://github.com/karmada-io/karmada/pull/5216 is very helpful, still, keeping apiEnablements contact in cluster helps avoid potential issues too.

pkg/controllers/status/cluster_status_controller.go
```diff
@@ -231,11 +232,16 @@ func (c *ClusterStatusController) setCurrentClusterStatus(clusterClient *util.Cl
 
 	// get the list of APIs installed in the member cluster
 	apiEnables, err := getAPIEnablements(clusterClient)
-	if len(apiEnables) == 0 {
-		klog.Errorf("Failed to get any APIs installed in Cluster %s. Error: %v.", cluster.GetName(), err)
-	} else if err != nil {
-		klog.Warningf("Maybe get partial(%d) APIs installed in Cluster %s. Error: %v.", len(apiEnables), cluster.GetName(), err)
+	if err != nil {
+		klog.Errorf("Failed to get APIs installed in Cluster %s. Error: %v.", cluster.GetName(), err)
+		return err
```
yanfeng1992, line 237: I am a little worried that this may cause other problems. If an error is returned here, the cluster will be unhealthy. If some APIs cannot be obtained, the cluster will be unhealthy.

NickYadance: would it be unhealthy ? seems the cluster needs to be healthy&online to reach getAPIEnablements, if err returned the reconcile will be requeue.

yanfeng1992: When changing from an unhealthy state to a healthy state, if some APIs have problems, an error will be returned here and the queue will be requeued. Unable to update the cluster to a healthy state

XiShanYongYe-Chang:
> indeed, updating APIENABLEMENTS is a very dangerous operation, which may cause the scheduler to delete the scheduled results. we're struggling with this lately. it caused us huge losses in the production environment.

Do you have any better ideas?

pkg/controllers/status/cluster_status_controller.go
NickYadance: sounds reasonable, maybe it's better keep apiEnablements unchanged when err happens.

NickYadance: no for above, if we do that the incorrect status will be kept until next reconcile which is not good at all

whitewindmills:
that's what https://github.com/karmada-io/karmada/pull/5325 and https://github.com/karmada-io/karmada/pull/5216 work on. lets move them forward.

whitewindmills:
okay

XiShanYongYe-Chang:
> > that's what #5325 and #5216 work on. lets move them forward.
> 
> How about we discuss this at a community meeting, get more people's opinions?

cc @NickYadance @yanfeng1992

whitewindmills:
next week's meeting agenda is already quite busy, I'm afraid we won't have a chance to discuss it.

RainbowMango:
> next week's meeting agenda is already quite busy, I'm afraid we won't have a chance to discuss it.

Don't worry let's try it. At least this brings attention.

whitewindmills:
hi @NickYadance, after the community meeting discussion, we still keep it as it is but add a new condition(named `InvalidAPIEnablements`? whatever you name it, it's up to you) when `err != nil` or `apiEnablements is empty`.
the scheduler will consider this condition.

NickYadance:
> hi @NickYadance, after the community meeting discussion, we still keep it as it is but add a new condition(named `InvalidAPIEnablements`? whatever you name it, it's up to you) when `err != nil` or `apiEnablements is empty`. the scheduler will consider this condition.

understand your concern, i'll keep the pr open, feel free to close it.

whitewindmills:
@NickYadance you can continue this PR.

XiShanYongYe-Chang:
Hi @NickYadance, just as @whitewindmills said, would you like to continue to contribute?

NickYadance:
> Hi @NickYadance, just as @whitewindmills said, would you like to continue to contribute?

no maybe, i would personally prefer to return on err here and stay simple, if the cluster cannot be healthy due to not being able to retrieve apienablements, so be unhealthy.

PR #5331, opened 2024-08-08
Add unit tests for common.go in pkg/util

NishantBansal2003:
**What type of PR is this?**
/kind failing-test

The test case coverage for common.go has been increased to 100%.
Ref #5235 

To verify the changes in the pkg/util directory run the following commands:
```
go test ./... -coverprofile=coverage.out
go tool cover -html=coverage.out -o coverage.html
open coverage.html
```

```release-note
NONE
```

### DIFF
--- a/pkg/util/common_test.go
+++ b/pkg/util/common_test.go
@@ -54,51 +54,74 @@ var (
 )
 
 func TestDiffKey(t *testing.T) {
-	t.Run("all old remove, all new added", func(t *testing.T) {
-		previous := map[string]sigMultiCluster{
-			"kubefed": kubefed,
-		}
-		current := map[string]string{
-			"karmada": "v1.6.0",
-		}
-		added, removed := DiffKey(previous, current)
-		if len(added) != 1 || added[0] != "karmada" {
-			t.Errorf("added = %v, want []string{\"karmada\"}", added)
-		}
-		if len(removed) != 1 || removed[0] != "kubefed" {
-			t.Errorf("removed = %v, want []string{\"kubefed\"}", removed)
-		}
-	})
-	t.Run("no old, all new added", func(t *testing.T) {
-		previous := map[string]sigMultiCluster{}
-		current := map[string]string{
-			"karmada": "v1.6.0",
-		}
-		added, removed := DiffKey(previous, current)
-		if len(added) != 1 || added[0] != "karmada" {
-			t.Errorf("added = %v, want []string{\"karmada\"}", added)
-		}
-		if len(removed) != 0 {
-			t.Errorf("removed = %v, want nil", removed)
-		}
-	})
-	t.Run("removed, added and remained", func(t *testing.T) {
-		previous := map[string]sigMultiCluster{
-			"kubefed":         kubefed,
-			"virtual-kubelet": virtualKubelet,
-		}
-		current := map[string]string{
-			"karmada":         "v1.6.0",
-			"virtual-kubelet": virtualKubelet.Comment,
-		}
-		added, removed := DiffKey(previous, current)
-		if len(added) != 1 || added[0] != "karmada" {
-			t.Errorf("added = %v, want []string{\"karmada\"}", added)
-		}
-		if len(removed) != 1 || removed[0] != "kubefed" {
-			t.Errorf("removed = %v, want []string{\"kubefed\"}", removed)
-		}
-	})
+	tests := []struct {
+		name            string
+		previous        map[string]sigMultiCluster
+		current         map[string]string
+		expectedAdded   []string
+		expectedRemoved []string
+	}{
+		{
+			name: "all old remove, all new added",
+			previous: map[string]sigMultiCluster{
+				"kubefed": kubefed,
+			},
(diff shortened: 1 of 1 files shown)


NishantBansal2003:
Hey @XiShanYongYe-Chang added 100% test coverage in `pkg/util/common.go`. PTAL...

XiShanYongYe-Chang:
Thanks @NishantBansal2003 
/assign

NishantBansal2003:
Hey @XiShanYongYe-Chang, I have some doubts I hope you can resolve:
1. Suppose there is a helper function used to test the functionality of another function, but not all aspects of the helper function are utilized (and therefore remain untested). Should we create separate tests specifically for this helper function, or is it unnecessary to test it?
2. I've read some of your reviews on PRs where you mentioned, `Although the current test content improves the file coverage, the test cases have no real meaning. Additionally, the maintenance cost is increased.` Could you explain this further so I can avoid making similar mistakes?

XiShanYongYe-Chang:
> Suppose there is a helper function used to test the functionality of another function, but not all aspects of the helper function are utilized (and therefore remain untested). Should we create separate tests specifically for this helper function, or is it unnecessary to test it?

I'm afraid I answered wrong. do you have any specific examples?

> I've read some of your reviews on PRs where you mentioned, Although the current test content improves the file coverage, the test cases have no real meaning. Additionally, the maintenance cost is increased. Could you explain this further so I can avoid making similar mistakes?

The functions to be tested should have clear purposes and the logic can be tested. For example, different test paths can be used to form different test cases. If the test case requires the same maintenance cost as the function under test, then I don't think the test is needed.

pkg/util/common_test.go
```diff
@@ -54,51 +54,79 @@ var (
...
+			isRemovedExpected := reflect.DeepEqual(removed, tt.expectedRemoved)
+			if !isAddedExpected && !isRemovedExpected {
+				t.Errorf("added = %v want %v, removed = %v want %v", added, tt.expectedAdded, removed, tt.expectedRemoved)
+			}
+			if !isAddedExpected {
+				t.Errorf("added = %v, want %v", added, tt.expectedAdded)
+			}
+			if !isRemovedExpected {
+				t.Errorf("removed = %v want %v", removed, tt.expectedRemoved)
+			}
```
XiShanYongYe-Chang, line 127: How about update like this:
```diff
                        added, removed := DiffKey(tt.previous, tt.current)
-                       isAddedExpected := reflect.DeepEqual(added, tt.expectedAdded)
-                       isRemovedExpected := reflect.DeepEqual(removed, tt.expectedRemoved)
-                       if !isAddedExpected && !isRemovedExpected {
-                               t.Errorf("added = %v want %v, removed = %v want %v", added, tt.expectedAdded, removed, tt.expectedRemoved)
-                       }
-                       if !isAddedExpected {
+                       if !reflect.DeepEqual(added, tt.expectedAdded) {
                                t.Errorf("added = %v, want %v", added, tt.expectedAdded)
                        }
-                       if !isRemovedExpected {
+                       if !reflect.DeepEqual(removed, tt.expectedRemoved) {
                                t.Errorf("removed = %v want %v", removed, tt.expectedRemoved)
                        }
```

pkg/util/common_test.go
```diff
@@ -111,15 +139,26 @@ func TestStringerJoin(t *testing.T) {
...
+	})
+	t.Run("empty keys", func(t *testing.T) {
+		var mcs map[string]sigMultiCluster
+		got := Keys(mcs)
+		var expect []string
+		sort.Strings(got)
+		if !reflect.DeepEqual(got, expect) {
+			t.Errorf("got = %v, want %v", got, expect)
+		}
+	})
```
XiShanYongYe-Chang, line 163: How about rewriting this test as tabel drive as well? Just like your rewrite above.

NishantBansal2003: I thought the same. Now I'm doing it...

NishantBansal2003:
> I'm afraid I answered wrong. do you have any specific examples?

Yeah like this -
https://github.com/karmada-io/karmada/blob/6e9136d5e99dcd7bcd7ae620bfb51ab668a12210/pkg/util/testing/mock_responder.go#L32-L36
Is used to test `TestConnectCluster` in https://github.com/karmada-io/karmada/blob/master/pkg/util/proxy/proxy_test.go
But other functions like -
https://github.com/karmada-io/karmada/blob/6e9136d5e99dcd7bcd7ae620bfb51ab668a12210/pkg/util/testing/mock_responder.go#L39-L48
Has not been used, Hence remain untested. So should I create separate tests specifically for this(untested) function, or is it unnecessary to test it?

NishantBansal2003:
PTAL...

XiShanYongYe-Chang (commented):
Thanks~
/lgtm
/approve

XiShanYongYe-Chang:
> Has not been used, Hence remain untested. So should I create separate tests specifically for this(untested) function, or is it unnecessary to test it?

I understand that this is part of the test framework, no need to test them.

Issue #4544, opened 2024-01-13
e2e: Publish e2e test completion tasks


zhzhuang-zju:
**What would you like to be added**:
As functions evolve, current e2e tests may not cover key function points well. Therefore, the e2e test is organized and published here. While writing e2e test cases, understanding of Karmada's functions can also be enhanced, which is a good way to contribute.
Thanks to the summary of @Vacant2333 , combined with the unfinished items I found, the e2e test case tasks are as follows:
- [ ] [clusterpropagation_test.go](https://github.com/karmada-io/karmada/blob/master/test/e2e/clusterpropagationpolicy_test.go)
   - [ ] add the test case when the [deployment updates](https://karmada.io/docs/next/userguide/scheduling/resource-propagating#update-deployment)
   - [ ] add the test case for the **same implicit priority** propagation
   - [ ] add the test case for **delete** the clusterPropagationPolicy
      - [ ] delete clusterPropagationPolicy and check if labels are deleted correctly (@zhzhuang-zju, #4539)
      - [ ] delete clusterPropagationPolicy and check if reference resource template and binding(clusterBinding) are deleted
      - [ ] delete a clusterPropagationPolicy when some clusterPropagationPolicies exist
   - [ ] update [clusterpropagation_test.md](https://github.com/karmada-io/karmada/blob/master/test/e2e/coverage_docs/clusterpropagation_test.md)  
- [ ] [propagation_test.go](https://github.com/karmada-io/karmada/blob/master/test/e2e/propagationpolicy_test.go)
   - [ ] add the test case for **delete** the propagationPolicy
      - [ ] delete propagationPolicy and check if labels are deleted correctly  (@zhzhuang-zju, #4539)
      - [ ] delete propagationPolicy and check if reference resource template and binding are deleted
   - [ ] update [propagation_test.md](https://github.com/karmada-io/karmada/blob/master/test/e2e/coverage_docs/propagation_test.md)
- [ ] [failover_test.go](https://github.com/karmada-io/karmada/blob/master/test/e2e/failover_test.go)
   - [ ] add a test case [when the GracePeriodSeconds is reach out](https://karmada.io/docs/next/userguide/failover/application-failover/#:~:text=after%20a%20timeout%20is%20reached%20before%20evicting%20the%20application)
   - [ ] update [failover_test.md](https://github.com/karmada-io/karmada/blob/master/test/e2e/coverage_docs/failover_test.md)
[...]

Issue #4262, opened 2023-11-17
Failed to join cluster in Pull-Mode


maaft:
**What happened**:

1. karmada installed in host-cluster and api-server is publicly exposed under <domain> and is reachable and functional (can do kubectl apply  etc)
2. in on-premise cluster (private networking; but access to internet) I try to join the host-cluster:

```shell
karmadactl register <domain> --token bsbmb3.<secret> --discovery-token-ca-cert-hash sha256:bb5593ab60547232a812109d87d48d3f16834bb68fa99035e64c8beeea442dd3
[preflight] Running pre-flight checks
[prefligt] All pre-flight checks were passed
[karmada-agent-start] Waiting to perform the TLS Bootstrap
[karmada-agent-start] Waiting to construct karmada-agent kubeconfig
Unable to connect to the server: dial tcp: lookup karmada-apiserver.karmada-system.svc.cluster.local on 127.0.0.53:53: server misbehaving
```
The join fails.

**What you expected to happen**:

I expect the member cluster to be added successfully.

RainbowMango:
> Unable to connect to the server: dial tcp: lookup karmada-apiserver.karmada-system.svc.cluster.local on 127.0.0.53:53: server misbehaving

Can you access karmada-apiserver on the machine where running the `karmadactl register`?

maaft:
yes, I can access it (via `kubectl`) from the machine where I'm running `karmadactl register`.

I'm not sure, why karmadactl even tries to resolve `karmada-apiserver.karmada-system.svc.cluster.local` on my local DNS, since I specified that karmada-api server must be found at `<domain>`?

maaft:
I did some further debugging:

`karmadaClusterInfo` in `pgk/karmadactl/register/register.go:340` contains this
```json
{
  "LocationOfOrigin": "",
  "server": "https://karmada-apiserver.karmada-system.svc.cluster.local:5443",
  "certificate-authority-data": "LS0tLS1C..."
}
```

Shouldn't there be `karmada-apiserver.example.com` instead?

maaft:
Okay, I think I found the issue:

1. karmada retrieves control plane info using bootstrap token
2. this info contains `karmada-apiserver.karmada-system.svc.cluster.local` and not `karmada-apiserver.example.com` 
3. agent kubeconfig is build with that info
4. agent tries to connect with that config and fails.

[...]

PR #5334, opened 2024-08-09
test/e2e/karmadactl_test.go: test register command

mohamedawnallah:
**What type of PR is this?**

/kind failing-test

Part of #4544 

**Context**

Dependent on #5429

```release-note
NONE
```

### DIFF
--- a/test/e2e/suites/base/karmadactl_test.go
+++ b/test/e2e/suites/base/karmadactl_test.go
@@ -40,7 +40,7 @@ import (
 	bootstrapapi "k8s.io/cluster-bootstrap/token/api"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	k8syaml "sigs.k8s.io/yaml"
+	"sigs.k8s.io/yaml"
 
 	clusterv1alpha1 "github.com/karmada-io/karmada/pkg/apis/cluster/v1alpha1"
 	policyv1alpha1 "github.com/karmada-io/karmada/pkg/apis/policy/v1alpha1"
@@ -1521,7 +1521,7 @@ var _ = ginkgo.Describe("Karmadactl apply testing", func() {
 // - error: An error if there was an issue during the process, otherwise nil.
 func WriteYamlToFile(obj interface{}, filePath string) error {
 	// Marshal the object to YAML.
-	yamlData, err := k8syaml.Marshal(obj)
+	yamlData, err := yaml.Marshal(obj)
 	if err != nil {
 		return err
 	}
@@ -1581,9 +1581,7 @@ var _ = framework.SerialDescribe("Karmadactl register testing", ginkgo.Ordered, 
 
 			// Extract the endpoint for Karmada APIServer.
 			endpointRegex := regexp.MustCompile(`(\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}):(\d{1,5})`)
-			endpointMatches := endpointRegex.FindStringSubmatch(output)
-			gomega.Expect(len(endpointMatches)).Should(gomega.BeNumerically(">=", 2))
-			karmadaAPIEndpoint = fmt.Sprintf("%s:%s", endpointMatches[1], endpointMatches[2])
+			karmadaAPIEndpoint = endpointRegex.FindString(output)
 			gomega.Expect(karmadaAPIEndpoint).Should(gomega.Equal(karmadaAPIEndpointExpected))
 
 			// Extract token.
--- a/test/e2e/suites/base/karmadactl_test.go
+++ b/test/e2e/suites/base/karmadactl_test.go
@@ -40,7 +40,7 @@ import (
 	bootstrapapi "k8s.io/cluster-bootstrap/token/api"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/yaml"
+	k8syaml "sigs.k8s.io/yaml"
 
 	clusterv1alpha1 "github.com/karmada-io/karmada/pkg/apis/cluster/v1alpha1"
 	policyv1alpha1 "github.com/karmada-io/karmada/pkg/apis/policy/v1alpha1"
@@ -1521,7 +1521,7 @@ var _ = ginkgo.Describe("Karmadactl apply testing", func() {
 // - error: An error if there was an issue during the process, otherwise nil.
 func WriteYamlToFile(obj interface{}, filePath string) error {
 	// Marshal the object to YAML.
-	yamlData, err := yaml.Marshal(obj)
+	yamlData, err := k8syaml.Marshal(obj)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


mohamedawnallah:
Currently this test case `should register a cluster with CA verification` fails with the following error:
```
  [FAILED] Unexpected error:
      <exec.CodeExitError>:
      error running /home/karmada/go/bin/karmadactl --kubeconfig=/home/karmada/.kube/member-e2e-gfj.config register 172.18.0.8:6443 --token bxfkm7.0xx94rzy1mda3iun --discovery-token-ca-cert-hash sha256:49514bb5d379ef5940a6e22709bfba9085950a0e54d2811c642fb75e4bdfd76c --cluster-name member-e2e-gfj --karmada-agent-image docker.io/karmada/karmada-agent:latest --ca-cert-path /tmp/karmada-e2e-301700272/ca.crt:
      Command stdout:
      [preflight] Running pre-flight checks
      [preflight] All pre-flight checks were passed
      [karmada-agent-start] Waiting to perform the TLS Bootstrap
      [karmada-agent-start] Waiting to construct karmada-agent kubeconfig

      stderr:
      Unable to connect to the server: dial tcp: lookup member-e2e-gfj-control-plane on 127.0.0.53:53: server misbehaving

      error:
      exit status 1
      {
          Err: <*errors.errorString | 0xc000062060>{
              s: "error running /home/karmada/go/bin/karmadactl --kubeconfig=/home/karmada/.kube/member-e2e-gfj.config register 172.18.0.8:6443 --token bxfkm7.0xx94rzy1mda3iun --discovery-token-ca-cert-hash sha256:49514bb5d379ef5940a6e22709bfba9085950a0e54d2811c642fb75e4bdfd76c --cluster-name member-e2e-gfj --karmada-agent-image docker.io/karmada/karmada-agent:latest --ca-cert-path /tmp/karmada-e2e-301700272/ca.crt:\nCommand stdout:\n[preflight] Running pre-flight checks\n[preflight] All pre-flight checks were passed\n[karmada-agent-start] Waiting to perform the TLS Bootstrap\n[karmada-agent-start] Waiting to construct karmada-agent kubeconfig\n\nstderr:\nUnable to connect to the server: dial tcp: lookup member-e2e-gfj-control-plane on 127.0.0.53:53: server misbehaving\n\nerror:\nexit status 1",
          },
          Code: 1,
      }
  occurred
```

Not sure if this is dependent on this issue #4262 ?

XiShanYongYe-Chang:
Thanks @mohamedawnallah 
The E2E seems to have failed for all versions.
/assign @zhzhuang-zju

mohamedawnallah:
I've checked the logs, and there is an issue with creating the [KarmadaDir](https://github.com/karmada-io/karmada/blob/37488bd1093a2c5cf808b6cf87d601caab07a05f/pkg/karmadactl/register/register.go#L80) at `/etc/karmada` because this directory is owned by `root` and requires `sudo` privileges. I've bypassed this locally by running:

```bash
sudo -E env "PATH=$PATH" ginkgo -focus "Karmadactl register testing"
```

This allows the test cases to bypass the issue, but I think there are a couple of problems with this approach. Currently, running Ginkgo with `sudo` in the CI environment could lead to issues later on?

A potential improvement could be to add a flag, something like `--karmada-directory`, to the `karmadactl register` command, similar to the `--ca-cert-path` flag. This would provide more flexibility for the end-user by allowing the use of a temporary directory during testing, thereby solving the issue out of the box. WDYT?

XiShanYongYe-Chang:
Hi @mohamedawnallah based on your previous E2E and UT, I think your test organization ideas are clear and your code is clean and easy to read. I look forward to your continued contributions. ^-^

zhzhuang-zju:
> A potential improvement could be to add a flag, something like `--karmada-directory`, to the `karmadactl register` command, similar to the `--ca-cert-path` flag. This would provide more flexibility for the end-user by allowing the use of a temporary directory during testing, thereby solving the issue out of the box. WDYT?

that make sense. Users might have the same issue under a non-root user account.

zhzhuang-zju:
> Hi @mohamedawnallah, I understand that the associated blockage has been resolved by #5862 . Do you have time to push this pr forward?

Thanks for reminding this, I almost forgot about this.

mohamedawnallah:
> Hi @mohamedawnallah, I understand that the associated blockage has been resolved by https://github.com/karmada-io/karmada/pull/5862 . Do you have time to push this pr forward?

Thanks, @XiShanYongYe-Chang, for bringing this to my attention. I've updated the test cases following the resolution of PR #5862. The remaining issue for this PR seems to be the error: `Unable to connect to the server: dial tcp: lookup member-e2e-jnpl9-control-plane on 127.0.0.53:53: server misbehaving`, which occurs when registering a cluster using a bootstrap token. I'm not sure if this is a known issue with the `karmada register` operation or something similar to what was discussed in #4262? I would love to hear your thoughts all.

XiShanYongYe-Chang:
Ask @zhzhuang-zju to help take a look~

zhzhuang-zju:
I‘ll take a look when the ci is completed

mohamedawnallah:
> The remaining issue for this PR seems to be the error: Unable to connect to the server: dial tcp: lookup member-e2e-jnpl9-control-plane on 127.0.0.53:53: server misbehaving, which occurs when registering a cluster using a bootstrap token. I'm not sure if this is a known issue with the karmada register operation or something similar to what was discussed in https://github.com/karmada-io/karmada/issues/4262? I would love to hear your thoughts all.

I've solved the issue. I found that I was using the member cluster config to create the bootstrap token instead of the Karmada APIServer config file. Changing this resolved the problem.

> error validating "/tmp/deploy-4jdtb.yaml": error validating data: [apiVersion not set, kind not set]; if you choose to ignore these errors, turn validation off with --validate=false

The e2e tests for this registering operation pass locally, but the error occurred in the CI tests and seems unrelated to the changes proposed in this PR. I've rebased onto the master branch since this PR has been open for a while, but the issue still persists. Any thoughts, @zhzhuang-zju?

mohamedawnallah:
> The remaining issue for this PR seems to be the error: Unable to connect to the server: dial tcp: lookup member-e2e-jnpl9-control-plane on 127.0.0.53:53: server misbehaving, which occurs when registering a cluster using a bootstrap token. I'm not sure if this is a known issue with the karmada register operation or something similar to what was discussed in https://github.com/karmada-io/karmada/issues/4262? I would love to hear your thoughts all.

I've solved that issue. I found that I was using the member cluster config to create the bootstrap token instead of the Karmada APIServer config file. Changing this resolved the problem.

> error validating "/tmp/deploy-4jdtb.yaml": error validating data: [apiVersion not set, kind not set]; if you choose to ignore these errors, turn validation off with --validate=false

The e2e tests for this registering operation pass locally, but the error occurred in the CI tests and seems unrelated to the changes proposed in this PR. I've rebased onto the master branch since this PR has been open for a while, but the issue still persists. Any thoughts, @zhzhuang-zju?

test/e2e/suites/base/karmadactl_test.go
```diff
@@ -26,6 +27,7 @@ import (
 
 	"github.com/onsi/ginkgo/v2"
 	"github.com/onsi/gomega"
+	"gopkg.in/yaml.v3"
```
zhzhuang-zju, line 30: @mohamedawnallah  Is it caused by the change of this YAML library, from "sigs.k8s.io/yaml" to "gopkg.in/yaml.v3"?

mohamedawnallah: Great observation, @zhzhuang-zju! 👌 That's right—Visual Studio Code's Ctrl+S failed me this time! 😂

mohamedawnallah:
>  s: "error running /home/runner/go/bin/karmadactl --kubeconfig=/home/runner/.kube/karmada.config register 172.18.0.4:6443 --token 2mnk2l.4segshmnxwcyshc9 --discovery-token-ca-cert-hash sha256:3477db6b4299125dc026573a04ebba3ebad161593c11498a7b66e3a150685ef2 --kubeconfig=/home/runner/.kube/member-e2e-9kc55.config --cluster-name member-e2e-9kc55 --karmada-agent-image docker.io/karmada/karmada-agent:latest:\nCommand stdout:\n[preflight] Running pre-flight checks\n[preflight] All pre-flight checks were passed\n[karmada-agent-start] Waiting to perform the TLS Bootstrap\n\nstderr:\nUnable to connect to the server: dial tcp: lookup karmada-host-control-plane on 127.0.0.53:53: server misbehaving\n\nerror:\nexit status 1",

It looks like the end-to-end test cases fail due to interference with other members. Running the register test case sequentially instead of in parallel may resolve the issue.

mohamedawnallah:
Hey @zhzhuang-zju, How am I supposed to move this PR forward? I've already addressed the feedback and all the test cases pass :)

XiShanYongYe-Chang:
Good job!  Isn't it okay for us to merge it?

mohamedawnallah:
> Hey @zhzhuang-zju, How am I supposed to move this PR forward? I've already addressed the feedback and all the test cases pass :)

> Sorry for the delay, I'll take a look ASAP

Never mind! I understand you've been busy. Just ping me if you have any additional feedback, and I'd be happy to address it :)

test/e2e/suites/base/karmadactl_test.go
```diff
@@ -1542,6 +1544,232 @@ func WriteYamlToFile(obj interface{}, filePath string) error {
...
+			tokenRegex := regexp.MustCompile(`--token\s+(\S+)`)
+			tokenMatches := tokenRegex.FindStringSubmatch(output)
+			gomega.Expect(len(tokenMatches)).Should(gomega.BeNumerically(">", 1))
+			token = tokenMatches[1]
+
+			// Extract discovery token CA cert hash.
+			hashRegex := regexp.MustCompile(`--discovery-token-ca-cert-hash\s+(\S+)`)
+			hashMatches := hashRegex.FindStringSubmatch(output)
+			gomega.Expect(len(hashMatches)).Should(gomega.BeNumerically(">", 1))
+			discoveryTokenCACertHash = hashMatches[1]
```
zhzhuang-zju, line 1688: Since the default validity period of the token is 24 hours, we can use the same token for different tests of the register command. I suggest moving this code into the beforeEach section for use by the subsequent two test suites. In this way, the function `extractAPIServerEndpoint` can also be omitted.

test/e2e/suites/base/karmadactl_test.go
```diff
@@ -1542,6 +1544,232 @@ func WriteYamlToFile(obj interface{}, filePath string) error {
...
+			gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+			fallbackContext = config.CurrentContext
+		})
+
+		ginkgo.By(fmt.Sprintf("Creating cluster: %s", newClusterName), func() {
+			err := createCluster(newClusterName, kubeConfigPath, controlPlane, clusterContext)
+			gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+		})
+
+		ginkgo.By(fmt.Sprintf("Set host-control-plane kubeconfig current context to %s", karmadaContext), func() {
```
zhzhuang-zju, line 1573: We can specify the kubeconfig and context through flags, so is it necessary to specifically set the current context?

test/e2e/suites/base/karmadactl_test.go
mohamedawnallah: > Since the default validity period of the token is 24 hours, we can use the same token for different tests of the register command. I suggest moving this code into the beforeEach section for use by the subsequent two test suites

I agree with reusing the token creation logic since tokens expire after 24 hours. However, as only one test case for CA Verification uses this token, is reusing the code necessary?

> In this way, the function extractAPIServerEndpoint can also be omitted.

It looks like `extractAPIServerEndpoint` function is still needed  in `should register a cluster without CA verification` test case?

test/e2e/suites/base/karmadactl_test.go
mohamedawnallah: > so is it necessary to specifically set the current context?

I don't recall the exact context back then but will try it and let you know.

test/e2e/suites/base/karmadactl_test.go
mohamedawnallah: > Since the default validity period of the token is 24 hours, we can use the same token for different tests of the register command. I suggest moving this code into the beforeEach section for use by the subsequent two test suites

I agree with reusing the same token for different tests if any since tokens expire after 24 hours. However, as only one test case for CA Verification uses this token, is putting the token creation logic in `beforeEach` necessary?

> In this way, the function extractAPIServerEndpoint can also be omitted.

It looks like `extractAPIServerEndpoint` function is still needed  in `should register a cluster without CA verification` test case?

test/e2e/suites/base/karmadactl_test.go
mohamedawnallah: > We can specify the kubeconfig and context through flags,

That's right. I've removed the code for setting the context, and the register e2e tests passed. I don't recall the rationale for manually setting the context back then.

test/e2e/suites/base/karmadactl_test.go
mohamedawnallah: > In this way, the function extractAPIServerEndpoint can also be omitted.

Upon reviewing the testing code more closely, I now recall that I used `extractAPIServerEndpoint` to test the actual APIServerEndpoint we retrieve from `--print-register-command` against the one we already know is true inside the `kubeconfig` global variable

zhzhuang-zju:
Great job~ thanks @mohamedawnallah 
/lgtm
leave Approve to @XiShanYongYe-Chang to take a look

test/e2e/suites/base/karmadactl_test.go
```diff
@@ -1542,6 +1542,150 @@ func WriteYamlToFile(obj interface{}, filePath string) error {
...
+				"token", "create", "--print-register-command="+"true",
+			)
+			output, err := cmd.ExecOrDie()
+			gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+
+			// Extract the endpoint for Karmada APIServer.
+			endpointRegex := regexp.MustCompile(`(\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}):(\d{1,5})`)
+			endpointMatches := endpointRegex.FindStringSubmatch(output)
+			gomega.Expect(len(endpointMatches)).Should(gomega.BeNumerically(">=", 2))
+			karmadaAPIEndpoint = fmt.Sprintf("%s:%s", endpointMatches[1], endpointMatches[2])
```
zhzhuang-zju, line 1586: There is an issue here. If the length of `endpointMatches` is 2, then accessing `endpointMatches[2]` will cause a panic.

We don't have the requirement to extract the IP and port separately. So, can we directly use `FindString` instead of `FindStringSubmatch`?

zhzhuang-zju: @mohamedawnallah If you are going to fix this issue, you can also remove the "k8syaml" alias for the "sigs.k8s.io/yaml" package. Currently, there is no need to add this alias.

zhzhuang-zju:
/lgtm
cc @XiShanYongYe-Chang

XiShanYongYe-Chang (commented):
Thanks a lot
/approve

PR #5362, opened 2024-08-13
charts/karmada: remove duplicate app label config in scheduler-estimator

iawia002:
**What type of PR is this?**

/kind cleanup

The scheduler-estimator Pod already has an app label configured:

https://github.com/karmada-io/karmada/blob/e5b12986081bf878bf4b70f3db2f8500e0bceb02/charts/karmada/templates/karmada-scheduler-estimator.yaml#L33-L35

And we also have a default app label value in the `values.yaml`:

https://github.com/karmada-io/karmada/blob/e5b12986081bf878bf4b70f3db2f8500e0bceb02/charts/karmada/values.yaml#L801-L802

This results in duplicate app labels in the scheduler-estimator Pod:

```yaml
labels:
  app: karmada-scheduler-estimator-xx
  app: karmada-scheduler-estimator
```

 This doesn't affect the helm installation, but it prevents kustomize from building it:

```
mapping key "app" already defined
```

This patch removes the default scheduler-estimator label config in `values.yaml` to make kustomize happy.

Fixes #

```release-note
none
```

### DIFF
--- a/.github/workflows/lint-chart.yaml
+++ b/.github/workflows/lint-chart.yaml
@@ -3,6 +3,7 @@ name: Chart Lint
 
 env:
   HELM_VERSION: v3.11.2
+  KUSTOMIZE_VERSION: 5.4.3
   KIND_VERSION: v0.22.0
   KIND_NODE_IMAGE: kindest/node:v1.29.0
   K8S_VERSION: v1.29.0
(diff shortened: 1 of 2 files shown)


liangyuanpeng:
> This patch removes the default scheduler-estimator label config in values.yaml to make kustomize happy.

It is possible to add a CI for it?

chaosi-zju:
> This doesn't affect the helm installation, but it prevents kustomize from building it:

hi, could you please help explaining what does `prevents kustomize from building it` mean? how could I reproduce your following error?

iawia002:
It means if you use `kustomize build` and it depends on the Karmada chart, it will directly report an error:

```
mapping key "app" already defined
```

The root cause of the error is that there are two keys with the same name, and Kustomize's validation is very strict, not allowing such a situation. This issue can be reproduced as long as you use Kustomize and depend on the Karmada chart, even without any customization. I will add CI to validate it.
 
 > This results in duplicate app labels in the scheduler-estimator Pod:
>
> ```yaml
> labels:
>   app: karmada-scheduler-estimator-xx
>   app: karmada-scheduler-estimator
> ```

iawia002:
Please see [this action](https://github.com/karmada-io/karmada/actions/runs/10380473227/job/28740361520?pr=5362). I simulate the situation of duplicate map keys. CI will report an error directly.

RainbowMango:
Wow Xinzhao, It's been quite a while, nice to meet you again!

charts/karmada/values.yaml
```diff
@@ -798,8 +798,7 @@ schedulerEstimator:
           XXXXXXXXXXXXXXXXXXXXXXXXXXX
           -----END RSA PRIVATE KEY-----
   ## @param schedulerEstimator.labels labels of the scheduler-estimator deployment
-  labels:
-    app: karmada-scheduler-estimator
+  labels: {}
```
chaosi-zju, line 801: I may prefer keeping this label (`app: karmada-scheduler-estimator`) in `value.yaml`.

But change the key name of the label in `karmada-scheduler-estimator.yaml`, such as `targetCluster: {{ $clusterName}}`

.github/workflows/lint-chart.yaml
```diff
@@ -34,13 +35,36 @@ jobs:
...
       - name: Run chart-testing (template)
         run: |
-          helm template --set components={"search,descheduler,schedulerEstimator"} --dependency-update ./charts/karmada --debug > /dev/null
-          helm template --set components={"search,descheduler,schedulerEstimator"},certs.mode=custom --dependency-update ./charts/karmada --debug > /dev/null
-          helm template --set components={"search,descheduler,schedulerEstimator"},installMode=component --dependency-update ./charts/karmada --debug > /dev/null
-          helm template --set installMode=agent --dependency-update ./charts/karmada --debug > /dev/null
-          helm template --dependency-update ./charts/karmada-operator --debug > /dev/null   
+          cat <<EOF > post-render.sh
+          #!/bin/sh
+          cat > all.yaml
```
chaosi-zju, line 47: so you build a empty file named `all.yaml`? and then when executing `kustomize` it will build all yaml from `template` dirtory?

charts/karmada/values.yaml
iawia002: I used a more concise name: `cluster`, and this label already existed.

.github/workflows/lint-chart.yaml
iawia002: Nope, this is used in `helm --post-render`, so it's a pipe, `all.yaml` contains helm stdout contents, this makes `kustomize` able to build all template files.

charts/karmada/values.yaml
chaosi-zju: okey, yours is better~ 👍

.github/workflows/lint-chart.yaml
chaosi-zju: Thank you for your reply

I see you added a line of explanation in the comment, this makes it easier to understand. 👍

chaosi-zju:
/lgtm

cc @RainbowMango

RainbowMango (commented):
/assign

RainbowMango (approved):
/approve

PR #5369, opened 2024-08-14
charts/karmada: ignore the static-resource Pod in the post-install check

iawia002:
**What type of PR is this?**

/kind cleanup

I encountered the following situation: the `static-resource` Job failed during execution, so it retried once. The retry was successful, and although the Job eventually completed successfully, the initial failure resulted in a failed Pod. Because of this failed Pod, the post-install check script kept running, even though all Karmada components were already running. This patch makes the post-install script ignore Pods related to `static-resource`.

Fixes #

```release-note
none
```

### DIFF
--- a/charts/karmada/templates/post-install-job.yaml
+++ b/charts/karmada/templates/post-install-job.yaml
@@ -53,7 +53,8 @@ spec:
 
           # The `post-install hook job` may be applied before all karmada components are ready that rely on `static-resource job` and `hook-job secret`.
           # So, we have to postpone the deletion of the `static-resource job` and `hook-job secret` until all karmada components are up and running.
-          while [[ $(kubectl get pods -n {{ $namespace }} --field-selector=status.phase!=Running,status.phase!=Succeeded -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}{end}' | wc -l) > 0 ]]; do
+          while [[ $(kubectl get pods -n {{ $namespace }} --field-selector=status.phase!=Running,status.phase!=Succeeded -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}{end}' | grep -v static-resource | wc -l) > 0 ]];
+          do
             echo "waiting for all pods of karmada control plane ready..."; sleep 1;
           done
 


chaosi-zju:
Hi, this problem realted to issue #5233

if you continue to try helm installation, you may also encounter this problem.

I have a batter resolution in PR #5305, but lack of people who good at helm reviews.

I think you are very proficient in helm. Can you help me review it?

when that pr merged, this problem will gone.

chaosi-zju:
> even if we merge #5305, the `static-resource` Job will still exist, right?

yes

>  causing the post-install check to fail.

no, after my #5305, it will no longer check `static-resource` Job pod.

> I didn't see how this relates to #5233
> I remember it had nothing to do with the secret

the relation is, you can see that wait condition in `static-resource` job and some init-container are newly introduced, which aims to install each component in order. However, these wait conditions are unreasonably implemented.

unreasonable wait condition caused #5233 and yours failure, so we hope to completely rectify these wait conditions, eventually we want to change it to #5305 

Of cource, if your issue is more urgent, we can also merge yours in first and then I'll think about moving forward #5305.

> but I'm new to the Karmada chart, I don't know the big picture about it

actually, I mainly want to ask for your opinion on the issue of sequentially installing deployments. I wonder if you know some best practices.

iawia002:
We may not on the same page. The `post-install` Job is still checking if all the Pods under the `karmada-system` namespace are `Ready`, isn't it? I didn't see any updates to the post-install check script in #5305. It still checks all the Pods under the entire namespace, including the static-resource, of course.

My issue is that if the `static-resource` Job has ever failed, the `post-install` Job will not exit. It will keep waiting because there is a failed static-resource Pod in the `karmada-system` namespace.

> actually, I mainly want to ask for your opinion on the issue of sequentially installing deployments. I wonder if you know some best practices.

Hardly at all, Helm's ability to handle dependency installation is very weak.

chaosi-zju:
> The post-install Job is still checking if all the Pods under the karmada-system namespace are Ready, isn't it?

Oh, sorry ! My fault, this pr has been around for nearly 10 days, and my memory is a little confused (((;꒪ꈊ꒪;)))

I originally intended to remove this waiting check in post-install Job directly (let job auto cleanup after finished), and then for some reason did not do so, I mistakenly thought I did.

So, you this PR is not conflict with that PR, I will continue to review it~

charts/karmada/templates/post-install-job.yaml
```diff
@@ -53,7 +53,8 @@ spec:
 
           # The `post-install hook job` may be applied before all karmada components are ready that rely on `static-resource job` and `hook-job secret`.
           # So, we have to postpone the deletion of the `static-resource job` and `hook-job secret` until all karmada components are up and running.
-          while [[ $(kubectl get pods -n {{ $namespace }} --field-selector=status.phase!=Running,status.phase!=Succeeded -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}{end}' | wc -l) > 0 ]]; do
+          while [[ $(kubectl get pods -n {{ $namespace }} --field-selector=status.phase!=Running,status.phase!=Succeeded -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}{end}' | grep -v static-resource | wc -l) > 0 ]];
+          do
```
chaosi-zju, line 57: I still don't think it's an efficient solution ( ･´ω`･ ) .

There is a misunderstanding here: the `status.phase==Running` of the karmada component means that the component has been installed successfully, actually not.

Example: `karmada-controller-manager` will soon into `running` state, but about 10 seconds lates, it finds out it miss CRD, then it will turn into `CrashLoopBackOff` state. 

I have tested a lot before, and there is a common case:

  1. `static-resource` always failed, which means it didn't successfully applied CRDs.
  2. `karmada-controller-manager` turn `running`, and several seconds later it turn into `CrashLoopBackOff` due to misssing CRD, it repeats back and forth.
  3. just during that `Running` period, the waiting condition passed and the `static-resource` job was deleted, and the CRD will no longer have chance to apply.

This is also the reason for why it used to be waiting `static-resource` job `status.phase==Succeeded`.

iawia002: I wait for the `static-resource` Job to complete before waiting for the karmada component to be ready. I can't exclude only the failed static-resource Pods in the script below. How about this?

iawia002: > I wait for the `static-resource` Job to complete before waiting for the karmada component to be ready. I can't exclude only the failed static-resource Pods in the script below.

Hold on, I just saw we have an init container for our components to wait for `static-resource` Job to be completed, So as long as the component is in the `Running` state, the `static-resource` Job must be completed, because the init container is finished, so we no longer need this, the original way is right, we can safely exclude `static-resource` Pods in post-install check script, right?

chaosi-zju: > So as long as the component is in the Running state, the static-resource Job must be completed,

You seem to have caught the key point. I was convinced, you were right!

>  the original way is right,

right

iawia002: I have reset the commit, PTAL

chaosi-zju:
/lgtm

cc @RainbowMango

RainbowMango (approved):
/approve
Thanks.

PR #5378, opened 2024-08-15
charts/karmada: fix the creation condition of metrics-adapter related APIService

iawia002:
**What type of PR is this?**

/kind bug

I did not enable the `metricsAdapter` component, so the metrics adapter-related deployment was not created:

https://github.com/karmada-io/karmada/blob/cd99897506a42b9398f5f958f1ffe6821414f419/charts/karmada/templates/karmada-metrics-adapter.yaml#L1

However, its `APIService` was still created, causing the controller-manager to continuously output the following error:

```
Discovery failed for some groups, 3 failing: unable to retrieve the complete list of server APIs: custom.metrics.k8s.io/v1beta1: stale GroupVersion discovery: custom.metrics.k8s.io/v1beta1, custom.metrics.k8s.io/v1beta2: stale GroupVersion discovery: custom.metrics.k8s.io/v1beta2, metrics.k8s.io/v1beta1: stale GroupVersion discovery: metrics.k8s.io/v1beta1
```

Screenshot 2024-08-15 at 17 10 45

This patch updates the creation condition of the related `APIService`, ensuring that they will be created together with the deployment.

Fixes #

```release-note
Fix the creation condition of metrics-adapter related APIService.
```

### DIFF
--- a/charts/karmada/templates/_karmada_apiservice.tpl
+++ b/charts/karmada/templates/_karmada_apiservice.tpl
@@ -32,6 +32,8 @@ metadata:
 spec:
   type: ExternalName
   externalName: {{ $name }}-aggregated-apiserver.{{ include "karmada.namespace" . }}.svc.{{ .Values.clusterDomain }}
+{{- end }}
+{{- if has "metricsAdapter" .Values.components }}
 ---
 apiVersion: apiregistration.k8s.io/v1
 kind: APIService


chaosi-zju:
great finding, I 've seen this error before, but I haven't gone into the cause.

but, why not use the same condition as metrics-adapter?

charts/karmada/templates/_karmada_apiservice.tpl
```diff
@@ -32,6 +32,7 @@ metadata:
 spec:
   type: ExternalName
   externalName: {{ $name }}-aggregated-apiserver.{{ include "karmada.namespace" . }}.svc.{{ .Values.clusterDomain }}
+{{- if has "metricsAdapter" .Values.components }}
```
chaosi-zju, line 35: I think it might be in the wrong position.

The outer condition is `{{- if eq .Values.installMode "host" }}`, then if my installMode is component, it will not run into this line?

chaosi-zju: I think this block should be on the same level as the below block `{{- if has "search" .Values.components }}`

iawia002: Nice catch! Fixed.

chaosi-zju:
e2e failure rely on #5382

/retest

chaosi-zju:
Hi @a7i @XiShanYongYe-Chang  this ci failure have occurred twice, which is introduced recently~

https://github.com/karmada-io/karmada/actions/runs/10401933004/job/28839752385?pr=5378

a7i:
> Hi @a7i @XiShanYongYe-Chang this ci failure have occurred twice, which is introduced recently~
> 
> https://github.com/karmada-io/karmada/actions/runs/10401933004/job/28839752385?pr=5378

taking a look!

chaosi-zju:
/lgtm

cc @RainbowMango

iawia002:
ping @RainbowMango @XiShanYongYe-Chang

RainbowMango (approved):
/approve

That makes 100% sense. Sorry for the delay.

PR #1804, opened 2022-05-17
Add metadataSufficient status to indicate wether cluster have get all api-resource 

duanmengkk:
Signed-off-by: duanmeng <duanmeng_yewu@cmss.chinamobile.com>

**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:
Fixes #1673

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #5400, opened 2024-08-19
add new cluster condition: CompleteAPIEnablements

whitewindmills:
**What type of PR is this?**
/kind feature

since https://github.com/karmada-io/karmada/pull/5325 is closed, I open this PR to do it.

Fixes #1804

```release-note
API change: Introduced a new condition `CompleteAPIEnablements` to represent api collection status of clusters.
```

### DIFF
--- a/pkg/apis/cluster/v1alpha1/types.go
+++ b/pkg/apis/cluster/v1alpha1/types.go
@@ -278,6 +278,9 @@ type LocalSecretReference struct {
 const (
 	// ClusterConditionReady means the cluster is healthy and ready to accept workloads.
 	ClusterConditionReady = "Ready"
+
+	// ClusterConditionCompleteAPIEnablements indicates whether the cluster's API enablements(.status.apiEnablements) is complete.
+	ClusterConditionCompleteAPIEnablements = "CompleteAPIEnablements"
 )
 
 // ClusterStatus contains information about the current status of a
--- a/pkg/controllers/status/cluster_status_controller.go
+++ b/pkg/controllers/status/cluster_status_controller.go
@@ -67,6 +67,10 @@ const (
 	clusterNotReachableReason = "ClusterNotReachable"
 	clusterNotReachableMsg    = "cluster is not reachable"
 	statusCollectionFailed    = "StatusCollectionFailed"
+
+	apiEnablementsComplete             = "Complete"
+	apiEnablementPartialAPIEnablements = "Partial"
+	apiEnablementEmptyAPIEnablements   = "Empty"
 )
 
 var (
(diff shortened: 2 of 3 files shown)


XiShanYongYe-Chang:
Thanks @whitewindmills 
/assign

pkg/apis/cluster/v1alpha1/types.go
```diff
@@ -278,6 +278,10 @@ type LocalSecretReference struct {
 const (
 	// ClusterConditionReady means the cluster is healthy and ready to accept workloads.
 	ClusterConditionReady = "Ready"
+
+	// ClusterConditionInvalidAPIEnablements means the cluster's APIEnablements are invalid.
+	// This condition usually occurs when the APIEnablements are not successfully collected.
+	ClusterConditionInvalidAPIEnablements = "InvalidAPIEnablements"
```
XiShanYongYe-Chang, line 284: Wouldn't it be better to use a more neutral descriptive word for this, such as APIEnablement completeness or something else?

pkg/controllers/status/cluster_status_controller.go
```diff
@@ -229,13 +233,22 @@ func (c *ClusterStatusController) setCurrentClusterStatus(clusterClient *util.Cl
...
 		klog.Errorf("Failed to get any APIs installed in Cluster %s. Error: %v.", cluster.GetName(), err)
 	} else if err != nil {
+		apiEnablementCondition = util.NewCondition(clusterv1alpha1.ClusterConditionInvalidAPIEnablements,
+			apiEnablementPartialAPIEnablements, fmt.Sprintf("might collect partial APIEnablements(%d) from the cluster", len(apiEnables)), metav1.ConditionTrue)
 		klog.Warningf("Maybe get partial(%d) APIs installed in Cluster %s. Error: %v.", len(apiEnables), cluster.GetName(), err)
+	} else {
+		apiEnablementCondition = util.NewCondition(clusterv1alpha1.ClusterConditionInvalidAPIEnablements,
+			apiEnablementSyncSucceed, "collected full APIEnablements from the cluster", metav1.ConditionFalse)
 	}
+	meta.SetStatusCondition(&currentClusterStatus.Conditions, apiEnablementCondition)
```
XiShanYongYe-Chang, line 251: It may not be possible to set this directly. I just recently fixed a bug related to this, ref #5227

whitewindmills:
@XiShanYongYe-Chang PTAL

XiShanYongYe-Chang (commented):
/lgtm
/cc @RainbowMango

pkg/apis/cluster/v1alpha1/types.go
```diff
@@ -278,6 +278,10 @@ type LocalSecretReference struct {
 const (
 	// ClusterConditionReady means the cluster is healthy and ready to accept workloads.
 	ClusterConditionReady = "Ready"
+
+	// ClusterConditionIncompleteAPIEnablements means the cluster's APIEnablements are incomplete.
+	// This condition usually occurs when the APIEnablements are not successfully collected.
+	ClusterConditionIncompleteAPIEnablements = "IncompleteAPIEnablements"
```
RainbowMango, line 284: We should try to avoid `double-negative` confusion. Let `true` represent normal, and `false` represent abnormal.

RainbowMango (commented):
I'm still thinking another approach:

In some cases, the controller can't collect a full list of API list continuously, such as there is a broken APIService in the member cluster, and thus will block the controller update the status of the Cluster object. And that's exactly the reason that we accept the partial list.

I'm thinking what if we just skip the .status.apiEnablements updating in case of we can't get full list of APIs? Is it doable?

whitewindmills:
> I'm thinking what if we just skip the .status.apiEnablements updating in case of we can't get full list of APIs? Is it doable?

yes, this pr won't block the controller update.

XiShanYongYe-Chang:
> I'm thinking what if we just skip the .status.apiEnablements updating in case of we can't get full list of APIs? Is it doable?

I had the same idea. Analyze the length of err or apienablements. Only update the status if err is not empty and apienablements is not empty.

whitewindmills:
> I'm thinking what if we just skip the .status.apiEnablements updating in case of we can't get full list of APIs? Is it doable?

assuming this is the case, if a user installs a new CRD, this means that the new CRD will not be updated to the cluster state, resulting in the failure to propagate such CRs.

whitewindmills:
at least the current situation is that if an API exists in the cluster state, we can consider it reliable, but if an API does not exist, we cannot consider it reliable.

chaunceyjiang:
> I'm thinking what if we just skip the .status.apiEnablements updating in case of we can't get full list of APIs? Is it doable?

Our scheduling plugin `APIEnablement` checks if the resource `kind `is present in `cluster.status.apiEnablements`. If not, it won't propagate to the member clusters.

whitewindmills:
ping @RainbowMango

whitewindmills:
@RainbowMango @XiShanYongYe-Chang @chaunceyjiang 
could you help move it forward, it's very important for us.

RainbowMango:
Sure! Sorry, too many on my plate recently, it is also very important to the project.

pkg/apis/cluster/v1alpha1/types.go
```diff
@@ -278,6 +278,9 @@ type LocalSecretReference struct {
 const (
 	// ClusterConditionReady means the cluster is healthy and ready to accept workloads.
 	ClusterConditionReady = "Ready"
+
+	// ClusterConditionCompleteAPIEnablements means the cluster's APIEnablements are complete.
+	ClusterConditionCompleteAPIEnablements = "CompleteAPIEnablements"
```
RainbowMango, line 283: Do we need to add the condition to file https://github.com/karmada-io/karmada/blob/70afc1ad6d170bd4773b5e7a9735a4116816f814/pkg/apis/cluster/types.go?

@XiShanYongYe-Chang

pkg/controllers/status/cluster_status_controller.go
```diff
@@ -67,6 +67,10 @@ const (
 	clusterNotReachableReason = "ClusterNotReachable"
 	clusterNotReachableMsg    = "cluster is not reachable"
 	statusCollectionFailed    = "StatusCollectionFailed"
+
+	apiEnablementSyncSucceed           = "SyncSucceed"
+	apiEnablementPartialAPIEnablements = "PartialAPIEnablements"
+	apiEnablementEmptyAPIEnablements   = "EmptyAPIEnablements"
```
RainbowMango, line 73:
```suggestion
	apiEnablementsComplete = "CompleteAPIEnablements"
	apiEnablementsPartial  = "PartialAPIEnablements"
	apiEnablementsEmpty    = "EmptyAPIEnablements"
```
How about this? Just feel `SyncSucceed` not so straightforward.

pkg/apis/cluster/v1alpha1/types.go
```diff
@@ -278,6 +278,9 @@ type LocalSecretReference struct {
 const (
 	// ClusterConditionReady means the cluster is healthy and ready to accept workloads.
 	ClusterConditionReady = "Ready"
+
+	// ClusterConditionCompleteAPIEnablements means the cluster's APIEnablements are complete.
```
RainbowMango, line 282:
```suggestion
	// ClusterConditionCompleteAPIEnablements indicates whether the cluster's API enablements(.status.apiEnablements) is complete.
```

RainbowMango:
Please help to rebase this as the test matrix changed after release-1.11.

RainbowMango:
Also, we need a release notes for this, this kind of API change.

pkg/apis/cluster/v1alpha1/types.go
XiShanYongYe-Chang: Although there is no logic to call this constant, I recommend adding it to this file, consistent with the previous ClusterConditionReady.

RainbowMango (commented):
/lgtm

Since no tests cover this so far, have you tested it locally?

RainbowMango:
I happen to see the conditions on my local environment:
```yaml
  - lastTransitionTime: "2024-09-13T09:04:15Z"
    message: might collect partial APIEnablements(22) from the cluster
    reason: PartialAPIEnablements
    status: "False"
    type: CompleteAPIEnablements
```
And 
```yaml
  - lastTransitionTime: "2024-09-13T09:05:35Z"
    message: collected complete APIEnablements from the cluster
    reason: CompleteAPIEnablements
    status: "True"
    type: CompleteAPIEnablements
```

Do you think we should have a shorter reason? Like `Complete`/`Partial`/`Empty`?

whitewindmills:
> Do you think we should have a shorter reason? Like `Complete`/`Partial`/`Empty`?

looks reasonable.

whitewindmills:
> have you tested it locally?

yes, I did test it but forgot to paste the result here.

RainbowMango (approved):
/lgtm
/approve

RainbowMango:
@yanfeng1992 @whitewindmills @chaunceyjiang 
Although this PR includes API changes and seems unsuitable for backport, given the significant potential impact, I think it is necessary to backport it to the release branches. What do you think?

By the way, introducing the condition doesn't break backward compatibility, and @XiShanYongYe-Chang has run some tests with it.

whitewindmills:
sounds good!

Issue #5363, opened 2024-08-13
[Umbrella] standardize the naming of karmada secrets across different installation methods


chaosi-zju:
**What would you like to be added**:

standardize the naming of karmada secrets across different installation methods.

**Why is this needed**:

In karmada, here are two important secrets, which is mount by most karmada components. One is `karmada-cert`, which contains a series of cert files like `ca.crt`, `apiserver.crt` and so on; another is `karmada-kubeconfig`, which contains a kubeconfig of karmada-apiserver.

However, in different installation methods, we used inconsistent secret naming or file path naming, which can potentially cause some unnecessary problems, such as:

* As for `secert/karmada-cert`, is named `karmada-cert-secret` in `local up` method, while named `karmada-cert` in `karmadactl` method, which potentially caused:
  * #5345
  * #5207
* In `secert/karmada-cert`, ca cert file is named `ca.crt` in `local up` method, while named `server-ca.crt` in `helm` method, which potentially caused:
  * #5322
* As for `secert/karmada-kubeconfig`, is named `kubeconfig` in `karmadactl init` and `local up` method, while named `karmada-kubeconfig` in `helm` method.
* In `helm` method, if I install karmada by `helm install karmada-xxx ...`, it will create a secret naming `karmada-xxx-cert`. Then if I install the single `karmada-scheduler-estimator` component by `helm install karmada-scheduler-estimator-xxx --set installMode=component ...`, the component will look for secret naming `karmada-scheduler-estimator-xxx-cert`, the name is inconsistent. In this case, since we cannot reuse the same `.Release.Name` when executing helm install, the secret name is not advised to be prefixed with `.Release.Name`.

**How to do**

I advise to standardize the naming of these karmada secrets like below in all installation methods:

* secretName: client-cert
mountPath:  /etc/karmada/pki/client (tls.crt、tls.key、ca.crt)

* secretName: server-cert
mountPath:  /etc/karmada/pki/server (tls.crt、tls.key、ca.crt)

* secretName: front-proxy-client-cert
mountPath:  /etc/karmada/pki/front-proxy-client (tls.crt、tls.key、ca.crt)

* secretName: etcd-client-cert
mountPath:  /etc/karmada/pki/etcd-client (tls.crt、tls.key、ca.crt)

* secretName: etcd-server-cert
mountPath:  /etc/karmada/pki/etcd-server (tls.crt、tls.key、ca.crt)

* secretName: webhook-server-cert
[...]

Issue #5353, opened 2024-08-12
通过helm部署，apiserver如何使用nodeport方式暴露出来


ayewwww:
**What happened**:

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Karmada version:
- kubectl-karmada or karmadactl version (the result of `kubectl-karmada version` or `karmadactl version`):
- Others:

chaosi-zju:
You can try `kubectl --context karmada-host -n karmada-system port-forward karmada-apiserver-xxx-xxx 5443 --address 0.0.0.0`

that `karmada-apiserver-xxx-xxx` should be the pod name of your karmada-apiserver

By the way, are you having any problems? 

Besides, as for me, I usually set `hostNetwork=true` in `values.yaml` when install karmada-apiserver by helm, actually other installation method set `hostNetwork=true` by default.

PR #5426, opened 2024-08-26
standardize the naming of karmada secrets in helm method

chaosi-zju:
**What type of PR is this?**

/kind cleanup

In karmada, here are two important secrets, which is mount by most karmada components. One is karmada-cert, which contains a series of cert files like ca.crt, apiserver.crt and so on; another is karmada-kubeconfig, which contains a kubeconfig of karmada-apiserver.

However, in different installation methods, we used inconsistent secret naming or file path naming, which can potentially cause some unnecessary problems, detail refer to #5363.

This PR aims to standardize the naming of karmada secrets in helm installation method.

Fixes part of #5363

In `helm` method, if I install karmada by `helm install karmada-xxx ...`, it will create a secret naming `karmada-xxx-cert`. Then if I install the single `karmada-scheduler-estimator` component by `helm install karmada-scheduler-estimator-xxx --set installMode=component ...`, the component will look for secret naming `karmada-scheduler-estimator-xxx-cert`, the name is inconsistent. In this case, since we cannot reuse the same `.Release.Name` when executing helm install, so the secret name is not advised to be prefixed with `.Release.Name`.

```release-note
helm chart: standardize the naming of karmada secrets in helm installation method
```

### DIFF
--- a/charts/karmada/templates/_helpers.tpl
+++ b/charts/karmada/templates/_helpers.tpl
@@ -110,17 +110,40 @@ app: {{- include "karmada.name" .}}-kube-controller-manager
 {{- end }}
 {{- end -}}
 
-{{- define "karmada.kubeconfig.volume" -}}
-{{- $name := include "karmada.name" . -}}
-- name: kubeconfig-secret
+{{- define "karmada.karmada-certs.volume" -}}
+- name: karmada-certs
   secret:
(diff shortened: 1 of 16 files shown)


zhzhuang-zju:
@chaosi-zju Can you remind me of the relationship between this PR and #5353？

PR #5442, opened 2024-08-28
charts/karmada: automatically clean up the static-resource Job after it completes

iawia002:
**What type of PR is this?**

/kind cleanup

After #5305, we can now delete the static-resource job directly after its completion.

The new [`ttlSecondsAfterFinished` feature](https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/) can handle this task elegantly. However, this feature only reached GA in version 1.23. Before this version, the feature requires manually enabling the feature gate, and we cannot assume that users' clusters have this feature gate enabled. Therefore, unlike the https://github.com/karmada-io/karmada/pull/5305#issuecomment-2313929232, considering that Karmada needs to run on various versions of Kubernetes, we cannot directly remove the post-install Job. The current strategy is:

1. For Kubernetes versions 1.23 or higher, use `ttlSecondsAfterFinished` for cleanup.
2. For Kubernetes versions lower than 1.23, continue using the post-install Job for cleanup.

Fixes #

```release-note
none
```

/cc @chaosi-zju @RainbowMango

### DIFF
--- a/charts/karmada/templates/karmada-static-resource-job.yaml
+++ b/charts/karmada/templates/karmada-static-resource-job.yaml
@@ -11,6 +11,9 @@ metadata:
 spec:
   parallelism: 1
   completions: 1
+  {{- if semverCompare ">=1.23.0-0" .Capabilities.KubeVersion.GitVersion }}
+  ttlSecondsAfterFinished: {{ .Values.staticResourceJob.ttlSecondsAfterFinished }}
+  {{- end }}
   template:
     metadata:
       name: {{ $name }}
--- a/charts/karmada/templates/post-install-job.yaml
+++ b/charts/karmada/templates/post-install-job.yaml
@@ -1,3 +1,4 @@
+{{- if semverCompare "<1.23.0-0" .Capabilities.KubeVersion.GitVersion }}
 {{- $name := include "karmada.name" . -}}
 {{- $namespace := include "karmada.namespace" . -}}
 {{- if eq .Values.installMode "host" }}
@@ -51,8 +52,8 @@ spec:
           bash <<'EOF'
           set -ex
 
-          # The `post-install hook job` may be applied before all karmada components are ready that rely on `static-resource job` and `hook-job secret`.
-          # So, we have to postpone the deletion of the `static-resource job` and `hook-job secret` until all karmada components are up and running.
+          # The `post-install hook job` may be applied before all karmada components are ready that rely on `static-resource job`.
+          # So, we have to postpone the deletion of the `static-resource job` until all karmada components are up and running.
           while [[ $(kubectl get pods -n {{ $namespace }} --field-selector=status.phase!=Running,status.phase!=Succeeded -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}{end}' | grep -v static-resource | wc -l) > 0 ]];
           do
             echo "waiting for all pods of karmada control plane ready..."; sleep 1;
@@ -60,4 +61,5 @@ spec:
 
           kubectl delete job {{ $name }}-static-resource -n {{ $namespace }}
           EOF
+{{- end }}
 {{- end }}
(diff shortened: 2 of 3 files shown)


charts/karmada/values.yaml
```diff
@@ -102,6 +102,9 @@ preInstallJob:
 staticResourceJob:
   tolerations: []
   nodeSelector: {}
+  ## Set a TTL for the static-resource Job, the Job will be automatically cleaned up after this time.
+  ## This only works on Kubernetes version 1.23 or higher.
+  ttlSecondsAfterFinished: 10
```
RainbowMango, line 107: What's the impact if we don't clean the job?

charts/karmada/templates/karmada-static-resource-job.yaml
```diff
@@ -11,6 +11,9 @@ metadata:
 spec:
   parallelism: 1
   completions: 1
+  {{- if semverCompare ">=1.23.0-0" .Capabilities.KubeVersion.GitVersion }}
```
RainbowMango, line 14: Where does the `.Capabilities.KubeVersion.GitVersion` come from?

Is this version always comparable? I mean I can't assume all Kubernetes version format is `x.x.y-0xxx`.

charts/karmada/values.yaml
iawia002: The finished Pod holds extra resources, e.g., IP. Since this Job is useless, we need to release those resources.

charts/karmada/templates/karmada-static-resource-job.yaml
iawia002: > Where does the `.Capabilities.KubeVersion.GitVersion` come from?

It's helm's [built-in objects](https://helm.sh/docs/chart_template_guide/builtin_objects/).

> Is this version always comparable? I mean I can't assume all Kubernetes version format is `x.x.y-0xxx`.

Yes, don't worry about the Kubernetes version format, it is semantic versioning. So the `-0` suffix here is used to find pre-release versions, e.g., 1.24-beta.1 will still be considered greater than 1.23, see https://masterminds.github.io/sprig/semver.html for more information.

charts/karmada/values.yaml
RainbowMango: Do you mean the Pod IP, right? Yeah, I agree that we should try to clean up temporary resources after the installation.
I'm asking because I want to evaluate the benefit we can get from this. Thanks for the clarification.

RainbowMango (approved):
/approve
Thanks.

charts/karmada/templates/post-install-job.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if semverCompare "<1.23.0-0" .Capabilities.KubeVersion.GitVersion }}
```
a7i, line 1: any particular reason why this was added here?

iawia002: We don't need this Job for K8s version higher than 1.23, the Job will be cleaned automatically by `ttlSecondsAfterFinished`. Please read the description of this PR.

RainbowMango: > This Job is no longer rendering for me

Do you mean it can't be installed on your side?

a7i: It is no longer rendered.

The explanation makes sense. I didn't realize post-install was split into static-resources and post-install, essentially making the latter useless if using ttlSecondsAfterFinished.

RainbowMango: Is there anything we need to do regarding this?

By the way, I pined you on the slack @a7i about the chart ownership thing :)

a7i: Amazing ❤️

PR #5329, opened 2024-08-08
add plaintext object overrider

Patrick0308:
**What type of PR is this?**

**What this PR does / why we need it**:
Add plaintextObjectOverrider to support modify a json or yaml field value.

For example:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: myconfigmap
  #...
data:
  test.json:  |
   {"key": {"test": "value"}, {"key1": "value"}}
```
The plaintextObjectOverrider to update the value of key's test field to "value1"
```
apiVersion: policy.karmada.io/v1alpha1
kind: OverridePolicy
metadata:
  name: example
spec:
  #...
  overrideRules:
    - overriders:
        plaintextObjectOverrider:
          - path: /data/test.json 
            plaintext: 
            - path: /key/test
               operator: replace
               value: value1
```

**Which issue(s) this PR fixes**:
Fixes #5330 

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
Introduced `PlaintextObjectOverrider`  for overriding values in YAML or JSON plaintext objects.
```

PR #5449, opened 2024-08-28
Overriding values inside JSON and YAML

### LINKED ISSUE #5330: Support json or yaml plaintext field overriding
Support overriding a json or yaml  field value.
ConfigMap data's always has yaml or json value. For example:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: myconfigmap
  #...
data:
  test.json:  |
   {"key": {"test": "value"}, {"key1": "value"}}
```
We need a overrider to modify test's value.
```
kind: OverridePolicy
metadata:
  name: example
spec:
  #...
  overrideRules:
    - overriders:
        plaintextObjectOverrider:
          - path: /data/test.json 
            plaintext: 
            - path: /key/test
               operator: replace
               value: value1
```

**Why is this needed**:

When configmap data is very big, this feature is powerful.

sophiefeifeifeiya:
**What type of PR is this?**

/kind design
/kind documentation

Overriding values inside JSON and YAML to simplify modifications.

Fixes #5330

Related PR: #5329 

```release-note
Introduced `FieldOverrider ` for overriding values in YAML or JSON.
```

### DIFF
--- a/docs/proposals/structured-configuration/README.md
+++ b/docs/proposals/structured-configuration/README.md
@@ -1,0 +1,543 @@
+---
+title: Structured configuration overrider
+authors:
+- "@Patrick0308"
+- "@sophiefeifeifeiya"
+reviewers:
+- "@chaunceyjiang"
+approvers:
+- "@chaunceyjiang"
+creation-date: 2024-08-12
+---
+
+# Structured configuration overrider
+## Summary
+The proposal introduces a new feature that allows users to partially override values inside  JSON and YAML fields. This is achieved using JSON patch operation. This design enables users to override the values within JSON/YAML fields partially, rather than replacing a whole JSON/YAML fields with `PlaintextOverrider`. Currently, `PlaintextOverrider`  applies JSON patch operations to whole fields, rather than specific values within fields, making it unsuitable for cases where users need to override individual values within those fields.
+
+## Motivation
+### Goals
++ Allow users to override specific values inside JSON and YAML in resources (e.g, configmap).
++ Support JSON patch (“add”, “remove”, “replace”) for both JSON and YAML.
+### Non-Goals
++ Support all data formats, like XML.
++ Support every operation of YAML and JSON.
+## Proposal
+### User Stories (Optional)
+#### Story 1
+As an administrator and developer, I want to update specific values within JSON/YAML in resources to without replacing the entire configuration, ensuring that my changes are minimal and targeted.
+### Notes/Constraints/Caveats (Optional)
+Illustrated in YAML Implementation.
+### Risks and Mitigations
+## Design Details
+### API Change
+```go
+type Overriders struct {
+    ...
+    // FieldOverrider represents the rules dedicated to modifying a specific field in any Kubernetes resource.
+    // This allows changing a single field within the resource with multiple operations.
+    // It is designed to handle structured field values such as those found in ConfigMaps or Secrets.
+    // The current implementation supports JSON and YAML formats, but can easily be extended to support XML in the future.
+    // +optional
+    FieldOverrider []FieldOverrider `json:"fieldOverrider,omitempty"`
+}
+
+type FieldOverrider struct {
+    // FieldPath specifies the initial location in the instance document where the operation should take place.
+    // The path uses RFC 6901 for navigating into nested structures. For example, the path "/data/db-config.yaml"
+    // specifies the configuration data key named "db-config.yaml" in a ConfigMap: "/data/db-config.yaml".
+    // +required
+    FieldPath string `json:"fieldPath"`
+
+    // JSON represents the operations performed on the JSON document specified by the FieldPath.
+    // +optional
+    JSON []JSONPatchOperation `json:"json,omitempty"`
+
+    // YAML represents the operations performed on the YAML document specified by the FieldPath.
+    // +optional
+    YAML []YAMLPatchOperation `json:"yaml,omitempty"`
+}
+
+// JSONPatchOperation represents a single field modification operation for JSON format.
+type JSONPatchOperation struct {
+    // SubPath specifies the relative location within the initial FieldPath where the operation should take place.
+    // The path uses RFC 6901 for navigating into nested structures.
+    // +required
+    SubPath string `json:"subPath"`
+
+    // Operator indicates the operation on target field.
+    // Available operators are: "add", "remove", and "replace".
+    // +kubebuilder:validation:Enum=add;remove;replace
+    // +required
+    Operator OverriderOperator `json:"operator"`
+
+    // Value is the new value to set for the specified field if the operation is "add" or "replace".
+    // For "remove" operation, this field is ignored.
+    // +optional
+    Value apiextensionsv1.JSON `json:"value,omitempty"`
+}
+
+// YAMLPatchOperation represents a single field modification operation for YAML format.
+type YAMLPatchOperation struct {
+    // SubPath specifies the relative location within the initial FieldPath where the operation should take place.
+    // The path uses RFC 6901 for navigating into nested structures.
+    // +required
+    SubPath string `json:"subPath"`
+
+    // Operator indicates the operation on target field.
+    // Available operators are: "add", "remove", and "replace".
+    // +kubebuilder:validation:Enum=add;remove;replace
+    // +required
+    Operator OverriderOperator `json:"operator"`
+
+    // Value is the new value to set for the specified field if the operation is "add" or "replace".
+    // For "remove" operation, this field is ignored.
+    // +optional
+    Value apiextensionsv1.JSON `json:"value,omitempty"`
+}
+
+// OverriderOperator is the set of operators that can be used in an overrider.
+type OverriderOperator string
+
+// These are valid overrider operators.
+const (
+    OverriderOpAdd     OverriderOperator = "add"
+    OverriderOpRemove  OverriderOperator = "remove"
+    OverriderOpReplace OverriderOperator = "replace"
+)
+```
+
+### User usage example
+For example, consider a ConfigMap with the following data in member cluster:
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: example-config
+data:
+  db-config.yaml: |
+    database:
+      host: localhost
+      port: 3306
+```
+The following is OverridePolicy which uses FieldOverrider:
+```yaml
+apiVersion: karmada.io/v1alpha1
+kind: OverridePolicy
+metadata:
+  name: example-configmap-override
+spec:
+  resourceSelectors:
+    - apiVersion: v1
+      kind: ConfigMap
+      name: example-config
+  overrideRules:
+    - overriders:
+        fieldOverrider:
+          - fieldPath: /data/db-config.yaml
+            yaml:
+              - subPath: /database/host
+                operator: replace
+                value: "remote-db.example.com"
+              - subPath: /database/port
+                operator: replace
+                value: "3307"
+```
+After we apply this policy, we can modify the db-config.yaml
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: example-config
+data:
+  db-config.yaml: |
+    database:
+      host: remote-db.example.com
+      port: 3307
+```
+
+### YAML Implementation
+
+We choose Plan1.
+(1) Plan1 is easy to implement by converting YAML directly to JSON, while Plan2 has to encapsulate ytt's grammar as if it is JSON patch.
+(2) Plan1 and Plan2 both have the same issues illustrated in the following.
+
+#### Plan 1: directly convert to and back from JSON (chosen)
+If YAML is directly converted to JSON and then converted back using (‘sigs.k8s.io/yaml’), some data type information might be lost.
+1. Dates and Times
+   ```yaml
+   dob: 1979-05-27T07:32:00Z
+   date_of_birth: 1979-05-27
+   
+   dob: time.Time
+   date_of_birth: time.Time
+   
+   # after transformation
+   
+   dob: string
+   date_of_birth: string
+   ```
+2. Supports anchors (&) and aliases (*) to reference and reuse values
+   ```yaml
+   default: &default
+     name: Alice
+     age: 30
+   employee1:
+     <<: *default
+     role: Developer
+   employee2:
+     <<: *default
+     role: Designer
+   
+   # after transformation
+   
+   default:
+     age: 30
+     name: Alice
+   employee1:
+     age: 30
+     name: Alice
+     role: Developer
+   employee2:
+     age: 30
+     name: Alice
+     role: Designer
+   ```
+   **Can be applied if:**
+   (1) Do not consider the situation described above
+   (2) Write cases to deal with different types **(large maintenance costs)**
+
+#### Plan 2: ytt (Aborted)
+Use the third party libraries **ytt**, supporting the overlays https://carvel.dev/ytt/docs/v0.50.x/ytt-overlays/ to implement and json operation similar operations. Implementation has its own specific syntax for ytt.
+Documents: https://carvel.dev/ytt/docs/v0.50.x/ytt-overlays/
+Joining methods: https://github.com/carvel-dev/ytt/blob/develop/examples/integrating-with-ytt/apis.md#as-a-go-module
+After testing, it has the same problem as the **Plan 1**.
+1. Dates and Times
+```yaml
+dob: 1979-05-27T07:32:00Z
+date_of_birth: 1979-05-27
+
+# after transformation
+
+dob: "1979-05-27T07:32:00Z"
+date_of_birth: "1979-05-27"
+```
+
+2. Supports anchors (&) and aliases (*) to reference and reuse values
+```yaml
+default: &default
+  name: Alice
+  age: 30
+employee1:
+  <<: *default
+  role: Developer
+employee2:
+  <<: *default
+  role: Designer
+
+# after transformation
+
+default:
+  name: Alice
+  age: 30
+employee1:
+  name: Alice
+  age: 30
+  role: Developer
+employee2:
+  name: Alice
+  age: 30
+  role: Designer
+number: 12345
+string: "6789"
+```
+**Can be applied if:**
+(1) Do not consider the situation described above
+(2) Ability to maintain functionality despite incompatibilities arising from encapsulating ytt as JSON operations later on.
+
+### Test Plan
+#### UT
+- Add unit tests to cover the new functions.
+#### E2E
++ Write proposal in `coverage_docs/overridepolicy_test.md` : deployment `FieldOverrider` testing;
++ Use ginkgo to complete the code `overridepolicy_test.go`.
+
+## Alternatives
+There are three API designs to achieve this:
+### (1) Each data format has different name with same struct
+```go
+type Overriders struct {
+    ...
+    // JSONPlaintextOverrider represents the rules dedicated to handling json object overrides
+    // +optional
+    JSONPlaintextOverrider []JSONPlaintextOverrider `json:"jsonPlaintextOverrider,omitempty"`
+    // YAMLPlaintextOverrider represents the rules dedicated to handling yaml object overrides
+    // +optional
+    YAMLPlaintextOverrider []YAMLPlaintextOverrider `json:"yamlPlaintextOverrider,omitempty"`
+}
+
+type JSONPlaintextOverrider struct {
+    // Path indicates the path of target field
+    Path string `json:"path"`
+    // JSONPatch represents json patch rules defined with plaintext overriders.
+    Patch []Patch `json:"patch"`
+    // MergeValue t represents the object value to be merged into the object.
+    MergeValue apiextensionsv1.JSON `json:"mergeValue"`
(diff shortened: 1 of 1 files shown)


docs/proposals/override-values-inside-structured-data-files/README.md
```diff
@@ -0,0 +1,227 @@
...
+
+```go
+type Overriders struct {
+	...
+	// PlaintextObjectOverrider represents the rules dedicated to handling yaml or json object overrides
+	// +optional
+	PlaintextObjectOverrider []PlaintextObjectOverrider `json:"plaintextObjectOverrider,omitempty"`
+}
+
+type PlaintextObjectOverrider struct {
```
chaunceyjiang, line 150: The `ConfigMap` often stores some configuration files of the application, such as xml, application.properties, toml, etc. commonly used in Java applications.

So in your proposal, would it be better to use `YAMLPlaintextOverrider` instead of `PlaintextObjectOverrider`? 

This way, we can later expand to other `Overrider` like `XMLPlaintextOverrider`, `TOMLPlaintextOverrider`, etc.

Patrick0308: If there are all these types overrider configs such as YAML, JSON, TOML，XML and etc and all of them are modified by json-merge or json-patch, maybe the API should be :
```
type Overriders struct {
	...
	// YAMLPlaintextOverrider represents the rules dedicated to handling yaml  object overrides
	// +optional
	YAMLPlaintextOverrider []PlaintextObjectOverrider `json:"yamlPlaintextOverrider,omitempty"`

	// JSONPlaintextOverrider represents the rules dedicated to handling json object overrides
	// +optional
	JSONPlaintextOverrider []PlaintextObjectOverrider `json:"jsonPlaintextOverrider,omitempty"`
}

type PlaintextObjectOverrider struct {
        // Path indicates the path of target field
        Path string `json:"path"`
        // JSONPatch represents json patch rules defined with plaintext overriders.
        JSONPatch []JSONPatch `json:"jsonPatch"`
        // MergePatch represents merge patch string.
        MergePatch string `json:"mergePatch"`
}

type JSONPatch  struct {
	// Path indicates the path of target field
	Path string `json:"path"`
	// From indicates the path of original field when operator is move and copy
        From string `json:"from"`
	// Operator indicates the operation on target field.
	// Available operators are: add, replace and remove.
	// +kubebuilder:validation:Enum=add;remove;replace
	Operator JSONPatchOperator `json:"operator"`
	// Value to be applied to target field.
	// Must be empty when operator is Remove.
	// +optional
	Value apiextensionsv1.JSON `json:"value,omitempty"`
}

// PatchOperator is the set of operators that can be used in an overrider.
type JSONPatchOperator string

// These are valid patch operators.
const (
	PatchOpAdd     JSONPatchOperator = "add"
	PatchOpRemove  JSONPatchOperator = "remove"
	PatchOpReplace  JSONPatchOperator = "replace"
	PatchOpMove  JSONPatchOperator = "move"
	PatchOpTest  JSONPatchOperator = "test"
	PatchOpCopy  JSONPatchOperator = "copy"
)

```

chaunceyjiang: +1

XiShanYongYe-Chang: Is it possible to introduce an enumerated value to indicate different data formats? In the future, only the enumerated value needs to be extended.

chaunceyjiang: >. Is it possible to introduce an enumerated value to indicate different data formats?

It might not work too well. If we introduce a field to represent the data format, that would mean different data formats would use `JSON Patch` to implement overrides. For data formats like XML, they have their own patching methods.
https://www.rfc-editor.org/rfc/rfc5261#appendix-A.6

chaunceyjiang: So if we want XML to use its own xml-patch instead of json-patch, that wouldn't scale very well. - Though that's probably a long shot. 😂

docs/proposals/override-values-inside-structured-data-files/README.md
```diff
@@ -0,0 +1,227 @@
...
+- "@chaunceyjiang"
+- TBD
+
+creation-date: 2024-08-12
+
+---
+
+
+
+# Overriding Values inside Structured Data Files
```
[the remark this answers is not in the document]
sophiefeifeifeiya: Changed to `Configurations of Different Data Formats`

docs/proposals/override-values-inside-structured-data-files/README.md
```diff
@@ -0,0 +1,227 @@
+---
+title: Support Variable in OverridePolicy
```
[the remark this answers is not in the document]
sophiefeifeifeiya: Yes you are right. I have made a mistake in there. This is the title of another proposal.

docs/proposals/override-values-inside-structured-data-files/README.md
```diff
@@ -0,0 +1,227 @@
...
+
+<!--
+What is out of scope for this KEP? Listing non-goals helps to focus discussion
+and make progress.
+-->
+
+
+
+
+
```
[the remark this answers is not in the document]
sophiefeifeifeiya: I will delete this.

docs/proposals/override-values-inside-structured-data-files/README.md
sophiefeifeifeiya: I will write two kinds of api in my proposal.

Two solutions give different implementation.

(1) each data format has different different struct
json -> json * op -> json ->  JSONPlaintextOverrider []JSONPlaintextOverrider
yaml -> yaml * op -> yaml -> YAMLPlaintextOverrider []YAMLPlaintextOverrider
xml -> xml * op -> xml -> XMLPlaintextOverrider []XMLPlaintextOverrider
...

For example json has 5 json operations, yaml has 3 yaml operations, and xml has 4 xml operations, ...
It should be 5+3+4+.. operations in total, which is a huge number.

So, each data format has different different struct would be a good solution if we choose to use native operation for each data type.

(2)  enum
json -> json * op -> json -> enum
yaml -> json * op -> yaml -> enum
xml -> json * op -> xml -> enum
...

Because we just need the number of operations of json, so we can use enum to contain all.


For me, the operations should be tailored to each specific data format. If I’m working with XML, I believe the operations should adhere to the XML format itself, rather than converting it to JSON and applying JSON-based operations.

docs/proposals/override-values-inside-structured-data-files/README.md
```diff
@@ -0,0 +1,227 @@
...
+What are the risks of this proposal, and how do we mitigate? 
+
+How will security be reviewed, and by whom?
+
+How will UX be reviewed, and by whom?
+
+Consider including folks who also work outside the SIG or subproject.
+-->
+
++ Introducing PlaintextObjectOverrider increases the complexity of OverridePolicy. To mitigate this, thorough documentation and examples will be provided.
```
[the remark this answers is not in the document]
sophiefeifeifeiya: Deleted.

docs/proposals/override-values-inside-structured-data-files/README.md
```diff
@@ -0,0 +1,227 @@
...
+
+
+<!--
+This section is incredibly important for producing high-quality, user-focused
+documentation such as release notes or a development roadmap. 
+
+A good summary is probably at least a paragraph in length.
+-->
+
+The proposal introduces a new feature, `PlaintextObjectOverrider` within the OverridePolicy in Karmada, allowing users to override values inside JSON or YAML plaintext objects. This feature expands the current capabilities of `PlaintextOverrider`, which only supports whole-field replacements, by enabling granular modifications within complex JSON or YAML structures.
```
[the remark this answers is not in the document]
sophiefeifeifeiya: Removed this description. Given a more clear description: The proposal introduces a new feature that allows users to partially override configurations of different data formats (e.g., JSON, YAML, XML) by data-format specific operations.

docs/proposals/override-values-inside-structured-data-files/README.md
sophiefeifeifeiya: I modified the proposal, and gives three api designs.

Patrick0308: @sophiefeifeifeiya Now yaml，toml，properties doesn't has own patch proposal. So we usually convert these data to JSON for [JSON-PATCH](https://datatracker.ietf.org/doc/html/rfc6902) in code.
If some structure data (such as XML) has their own patch rules, we can extend the API by defining their own overrider in the future.

sophiefeifeifeiya: Okay, given that we just support data formats with JSON patch, we can continue with "Each data format has different name with same struct" api design plan https://github.com/karmada-io/karmada/pull/5449#discussion_r1740129736, as enum is hard to extend in the future. What do you guys think about this? @XiShanYongYe-Chang @chaunceyjiang 

If this one is good, then I will modify the proposal details based on this and try to implement it.

XiShanYongYe-Chang: I think we can continue to push forward as we thought.

I understand that JSON and YAML data should be the main concern of this proposal. From the perspective of solution design, it is best to reserve design for other types of data. You can explain the design idea in the current proposal.

sophiefeifeifeiya:
## Updates
Some updates after testing:
1. The issue with “Explicit Type Tags” in Plan 1 of the YAML implementation (yaml -> json -> yaml) was identified as follows:

   ```
   number: !!int 12345
   string: !!str 6789
   # after transformation
   number: 12345
   string: "6789"
   ```
In this process, the explicit type tags are removed, and the values are transformed into the data types indicated by the tags. However, this previously identified issue has now been resolved.

2. After testing Plan2 ytt, it still has the same problem as Plan 1.
image
As we have to use encapsulate it, and teh grammar of ytt is quite different from JSON, Plan 1 is preferable.

3. Final API design is in the proposal. Maybe some tidy changes can be made. Feel free to give suggestions.

The proposal is modified again, and you can refer to new modified proposal.

## Conclusion
Thus, for the YAML implementation, Plan 1 is preferable. However, it’s important to inform users that:
​1. Dates and times may result in data type loss.
​2. Anchors (&) and aliases (*) used for referencing and reusing values will be hardcoded after transformation.

What do you guys think about?
@Patrick0308 @chaunceyjiang @XiShanYongYe-Chang

XiShanYongYe-Chang:
At the same time cc @RainbowMango

docs/proposals/override-values-inside-configurations-of-different-data-formats/README.md
```diff
@@ -0,0 +1,415 @@
...
+Given that other most of data formats can be converted to JSON. So, they can be converted to JSON first, do JSON operations, and converts back.
+There are two API designs to achieve this:
+### (1) Each data format has different name with same struct
+```go
+type Overriders struct {
+	...
+  // JSONPlaintextOverrider represents the rules dedicated to handling json object overrides
+	// +optional
+	JSONPlaintextOverrider []PlaintextObjectOverrider `json:"jsonPlaintextOverrider,omitempty"`
+	// YAMLPlaintextOverrider represents the rules dedicated to handling yaml object overrides
```
chaunceyjiang, line 314: image


The format here seems to have some issues.

docs/proposals/override-values-inside-configurations-of-different-data-formats/README.md
```diff
@@ -0,0 +1,415 @@
...
+	PatchOpMove  JSONPatchOperator = "move"
+	PatchOpTest  JSONPatchOperator = "test"
+	PatchOpCopy  JSONPatchOperator = "copy"
+)
+
+type YAMLPlaintextOverrider struct {
+  // Path indicates the path of target field
+  Path string `json:"path"`
+  // JSONPatch represents json patch rules defined with plaintext overriders.
+  JSONPatch []JSONPatch `json:"jsonPatch"`
```
chaunceyjiang, line 92: Here, using `jsonPatch` doesn't seem very reasonable.

docs/proposals/override-values-inside-configurations-of-different-data-formats/README.md
```diff
@@ -0,0 +1,415 @@
...
+
+creation-date: 2024-08-12
+
+
+---
+
+# Overriding values inside Configurations of Different Data Formats
+## Summary
+The proposal introduces a new feature that allows users to partially override configurations of different data formats (e.g., JSON, YAML, XML) by data-format specific operations. This feature expands the current capabilities of `PlaintextOverrider` which only supports whole-field replacements. Take JSON as an example, we can use JSON patch and JSON merge to override the configurations, rather than give a whole JSON file to replace.
+## Motivation
```
RainbowMango, line 22: Can you give an example here to explain why we need this proposal in detail, such as 
- what is the problem this proposal hopes to solve
- why this problem can not be handled by the current `PlaintextOverrider`

docs/proposals/override-values-inside-configurations-of-different-data-formats/README.md
RainbowMango: Why? Hard to write the Patch in yaml?

docs/proposals/override-values-inside-json-and-yaml/README.md
```diff
@@ -0,0 +1,545 @@
+---
+title: Overriding values inside JSON and YAML
```
RainbowMango, line 2:
```suggestion
title: Structured configuration overrider
```

docs/proposals/override-values-inside-json-and-yaml/README.md
RainbowMango, line 1: Please help move this proposal to `docs/proposals/override/structured-configuration/README.md`.

docs/proposals/override-values-inside-json-and-yaml/README.md
```diff
@@ -0,0 +1,545 @@
...
+authors:
+- "@Patrick0308"
+- "@sophiefeifeifeiya"
+reviewers:
+- "@chaunceyjiang"
+- TBD
```
RainbowMango, line 8:
```suggestion
- "@chaunceyjiang"
```

docs/proposals/override-values-inside-json-and-yaml/README.md
```diff
@@ -0,0 +1,545 @@
...
+approvers:
+- "@chaunceyjiang"
+- TBD
```
RainbowMango, line 11:
```suggestion
```

docs/proposals/override-values-inside-json-and-yaml/README.md
```diff
@@ -0,0 +1,545 @@
...
+const (
+    OverriderOpAdd     OverriderOperator = "add"
+    OverriderOpRemove  OverriderOperator = "remove"
+    OverriderOpReplace OverriderOperator = "replace"
+)
+```
+
+### User usage example
+For example, consider a ConfigMap with the following data in member cluster:
+```json
```
RainbowMango, line 113:
```suggestion
```yaml
```

docs/proposals/override-values-inside-json-and-yaml/README.md
```diff
@@ -0,0 +1,545 @@
...
+metadata:
+  name: example-config
+data:
+  db-config.yaml: |
+    database:
+      host: localhost
+      port: 3306
+```
+The following is OverridePolicy which uses FieldOverrider:
+```json
```
RainbowMango, line 125:
```suggestion
```yaml
```

docs/proposals/override-values-inside-json-and-yaml/README.md
```diff
@@ -0,0 +1,545 @@
...
+  	- overriders:
+   			fieldOverrider:
+        -	fieldPath: /data/db-config.yaml
+       		yaml:
+            - subPath: /database/host
+              operator: replace
+              value: "remote-db.example.com"
+            - subPath: /database/port
+              operator: replace
+              value: "3307"
```
RainbowMango, line 145: Format issue.

docs/proposals/override-values-inside-json-and-yaml/README.md
```diff
@@ -0,0 +1,545 @@
...
+```
+After we apply this policy, we can modify the db-config.yaml
+```json
```
RainbowMango, line 148:
```suggestion
```yaml
```

docs/proposals/override-values-inside-json-and-yaml/README.md
```diff
@@ -0,0 +1,545 @@
...
+json -> json * op -> json -> enum
+yaml -> json * op -> yaml -> enum
+xml -> json * op -> xml -> enum
+...
+This one is designed for **JSON operations** for all data formats.
+
+### User usage example for (1)
+For example, consider a ConfigMap with the following data in member cluster:
+
+```json
```
RainbowMango, line 449:
```suggestion
```yaml
```

docs/proposals/override-values-inside-json-and-yaml/README.md
```diff
@@ -0,0 +1,545 @@
...
+      "keyC": "valueC",
+      "keyD": "valueD",
+      "keyE": "valueE",
+      "keyF": "valueF"
+    }
+```
+
+The following is OverridePolicy which uses JSONPlaintextOverrider:
+
+```json
```
RainbowMango, line 469:
```suggestion
```yaml
```

docs/proposals/override-values-inside-json-and-yaml/README.md
```diff
@@ -0,0 +1,545 @@
...
+2. Add: Adds an empty string as the value for keyD.
+3. Remove: Removes keyB and its value.
+4. Replace: Replaces the value of keyC with `newly added value`.
+5. Move: Moves the value of keyD to keyF, which effectively deletes keyD and sets the value of keyF to `valueD`.
+6. Copy: Copies the value of keyE to keyG.
+7. Merge: Adds new keys keyH and keyI with values `valueH` and `valueI`.
+8. Merge Raw Value: Adds keys keyJ and keyK with values `valueJ` and `valueK`.
+   Finally, we get a new config.json with JSON operation.
+
+```json
```
RainbowMango, line 526:
```suggestion
```yaml
```

RainbowMango (commented):
Leaving some nits, otherwise LGTM.

@chaunceyjiang @Patrick0308 Please take another look, if no further comments I guess we can move it forward then.

chaunceyjiang (commented):
LGTM.

/cc @Patrick0308 @RainbowMango

docs/proposals/structured-configuration/README.md
```diff
@@ -0,0 +1,543 @@
...
+reviewers:
+- "@chaunceyjiang"
+approvers:
+- "@chaunceyjiang"
+creation-date: 2024-08-12
+---
+
+# Structured configuration overrider
+## Summary
+The proposal introduces a new feature that allows users to partially override values inside  JSON and YAML fields. This is achieved using JSON patch operation. This design enables users to to override the values within JSON/YAML fields partially, rather than replacing a whole JSON/YAML fields with `PlaintextOverrider`. Currently, `PlaintextOverrider`  applies JSON patch operations to whole fields, rather than specific values within fields, making it unsuitable for cases where users need to override individual values within those fields.
```
Patrick0308, line 15: nit hit: About `to to override`, word `to` is duplicated.

docs/proposals/structured-configuration/README.md
```diff
@@ -0,0 +1,543 @@
...
+  db-config.yaml: |
+    database:
+      host: remote-db.example.com
+      port: 3307
+```
+
+### YAML Implementation
+
+We choose Plan1.
+(1) Plan1 is easy to implmenet by converting YAML directly to JSON, while Plan2 has to encapsulate ytt's grammar as if it is JSON patch.
```
Patrick0308, line 161: nit hit: Spell mistake `implmenet`

docs/proposals/structured-configuration/README.md
```diff
@@ -0,0 +1,543 @@
...
+(2) Plan1 and Plan2 both have the same issues illustrated in the following.
+
+#### Plan 1: directly convert to and back from JSON (chosen)
+If YAML directly converts to JSON and convert back with ("sigs.k8s.io/yaml"), it will lose some data type information.
```
Patrick0308, line 165:
```suggestion
If YAML is directly converted to JSON and then converted back using the 'sigs.k8s.io/yaml' package, it will lose some data type information.
```

Patrick0308 (commented):
Some nit hits. LGTM

docs/proposals/structured-configuration/README.md
RainbowMango, line 15:
```suggestion
The proposal introduces a new feature that allows users to partially override values inside  JSON and YAML fields. This is achieved using JSON patch operation. This design enables users to override the values within JSON/YAML fields partially, rather than replacing a whole JSON/YAML fields with `PlaintextOverrider`. Currently, `PlaintextOverrider`  applies JSON patch operations to whole fields, rather than specific values within fields, making it unsuitable for cases where users need to override individual values within those fields.
```

RainbowMango (commented):
/lgtm

@chaunceyjiang @Patrick0308 Do you have any further comments?

chaunceyjiang (approved):
/approve

PR #5457, opened 2024-08-29
Provide detailed error message when the  e2e test fails

zhzhuang-zju:
**What type of PR is this?**
/kind cleanup

The e2e test `Karmadactl top testing Karmadactl top pod which does not exist [It] Karmadactl top pod which does not exist` has been intermittently failing recently, for example:
- https://github.com/karmada-io/karmada/actions/runs/10611034056/job/29409966885?pr=5454

Due to the lack of effective positioning information, it's difficult to identify the root cause. Therefore, we plan to add error messages to facilitate subsequent troubleshooting.

Fixes #

None

```release-note
None
```

### DIFF
--- a/test/e2e/karmadactl_test.go
+++ b/test/e2e/karmadactl_test.go
@@ -546,7 +546,8 @@ var _ = ginkgo.Describe("Karmadactl top testing", func() {
 			for _, clusterName := range framework.ClusterNames() {
 				cmd := framework.NewKarmadactlCommand(kubeconfig, karmadaContext, karmadactlPath, "", karmadactlTimeout, "top", "pod", podName, "-n", testNamespace, "-C", clusterName)
 				_, err := cmd.ExecOrDie()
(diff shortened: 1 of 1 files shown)


zhzhuang-zju:
cc @XiShanYongYe-Chang

XiShanYongYe-Chang (commented):
/lgtm
/approve

zhzhuang-zju:
The added detailed error message works!，see https://github.com/karmada-io/karmada/actions/runs/12197101877/job/34026335469?pr=5772#step:6:1770

e2e fails because the member cluster's server is not available at this time.

> error: cluster(member1): the server is currently unable to handle the request (get pods.metrics.k8s.io pod-z682t)

But how to fix it? WDYT? @XiShanYongYe-Chang

zhzhuang-zju:
We are unable to determine why `metrics-server` in member1 cannot handle requests due to the absence of logs from the member cluster and uncertainty regarding whether this is a transient service outage. Should we enhance the end-to-end (e2e) stability by implementing retry logic? @XiShanYongYe-Chang

XiShanYongYe-Chang:
I think we can.

PR #5472, opened 2024-09-02
karmada-scheduler-estimator add the support for custom DNS suffix

zhzhuang-zju:
**What type of PR is this?**
/kind feature

when dial scheduler-estimator, use ServiceName/serviceNamspace instead of the full domain name. Because the DNS suffix `.svc.cluster.local` is configurable

Fixes #

```release-note

```

### DIFF
--- a/pkg/estimator/client/service.go
+++ b/pkg/estimator/client/service.go
@@ -31,7 +31,7 @@ import (
 // ResolveCluster parses Service resource content by itself.
 // Fixes Issue https://github.com/karmada-io/karmada/issues/2487
 // Modified from "k8s.io/apiserver/pkg/util/proxy/proxy.go:92 => func ResolveCluster"
-func resolveCluster(kubeClient kubernetes.Interface, namespace, id string, port int32) (string, error) {
+func resolveCluster(kubeClient kubernetes.Interface, namespace, id string, port int32) ([]string, error) {
 	svc, err := kubeClient.CoreV1().Services(namespace).Get(context.TODO(), id, metav1.GetOptions{})
 	if err != nil {
 		if apierrors.IsNotFound(err) {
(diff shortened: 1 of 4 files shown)


pkg/estimator/client/service.go
```diff
@@ -41,7 +41,7 @@ func resolveCluster(kubeClient kubernetes.Interface, namespace, id string, port
 			 * But the Service resource is defined in Host Kubernetes Cluster. So we cannot get its content here.
 			 * The best thing we can do is just glue host:port together, and try to connect to it.
 			 */
-			return net.JoinHostPort(fmt.Sprintf("%s.%s.svc.cluster.local", id, namespace), fmt.Sprintf("%d", port)), nil
```
chaunceyjiang, line 44: Is this compatible with the old version?

zhzhuang-zju: you mean change `%s.%s.svc.cluster.local` to `%s.%s`?

whitewindmills: can this provide any improvement?

zhzhuang-zju: The purpose of this modification is that some environments have a default DNS suffix that is not `svc.cluster.local`. By completely removing that segment, the estimator can run properly in these environments.

zhzhuang-zju: However, there is now a certificate issue. The old certificate does not include `*.karmada-system`, which can cause upgrade problems. I am currently trying to resolve this.

zhzhuang-zju: refer to [Service DNS](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#services),
>  with a name of the form `my-svc.my-namespace.svc.cluster-domain.example`

looks like the `svc` part of DNS is not configurable. If so, we can use `ServiceName.serviceNamspace.svc` as the address of the gRPCConnection, without worrying about the incompatibility issues of certificates, because the current karmada certificate already has the host `*.karmada-system.svc`. 

WDYT @chaunceyjiang @whitewindmills

whitewindmills: seems that such domain `ServiceName.serviceNamspace.svc` is not be supported by windows nodes.
https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#dns-windows

chaunceyjiang: I believe we can first use `*.karmada-system.svc.`. Then in the certificate, add `*.karmada-system.`. In the next version, use `*.karmada-system.`.

zhzhuang-zju: > seems that such domain ServiceName.serviceNamspace.svc is not be supported by windows nodes.
https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#dns-windows

@whitewindmills Yeah, looks like windows nodes only support domain `ServiceName`. So, changing it to `*.karmada-system.svc` will result in the loss of support for Windows nodes while adding support for Linux custom DNS suffixes. I am not certain if this trade-off is entirely acceptable.

I currently have two proposals:
- Change `ServiceName.serviceNamespace.svc.cluster.local` to `ServiceName.serviceNamespace.svc`.
- Implement a mechanism in gRPC for polling multiple addresses, using the first successfully dialed address from either `ServiceName.serviceNamespace.svc.cluster.local` or `ServiceName.serviceNamespace.svc` to establish the gRPC connection. 

The benefit of the second proposal is that it maintains compatibility with Windows nodes.

WDYT? @chaunceyjiang @whitewindmills

zhzhuang-zju: kindly ping @chaunceyjiang @whitewindmills

whitewindmills: I see a lot of `*.svc.cluster.local` usage in karmada-operator and hack/*.sh, I'm just curious why there is no negative impact in such an environment.

zhzhuang-zju: > I see a lot of *.svc.cluster.local usage in karmada-operator and hack/*.sh

For `hack/*.sh`, `*.svc.cluster.local` and `*.svc` appear together, which means that `*.svc` can support custom DNS suffixes.

zhzhuang-zju: However, for `karmada-operator`, I haven't studied it in detail yet. From my understanding, `karmada-operator` should not directly use `*.svc.cluster.local`, because `karmada-operator` supports custom DNS suffixes.

https://github.com/karmada-io/karmada/blob/bf83dfe87c9bb222047197a2d2e91e7fa8464f5e/operator/pkg/apis/operator/v1alpha1/type.go#L173-L178

whitewindmills: however, this is just one of them. I have seen such usage mainly concentrated on the deployment of etcd.
https://github.com/karmada-io/karmada/blob/6bbbf9d8aeca49ae1f412da117ca17f995c42077/operator/pkg/controlplane/etcd/etcd.go#L54-L58

pkg/util/grpcconnection/config.go
```diff
@@ -138,6 +136,25 @@ func (c *ClientConfig) DialWithTimeOut(path string, timeout time.Duration) (*grp
 	}
 
 	opts = append(opts, grpc.WithTransportCredentials(cred))
+
+	var cc *grpc.ClientConn
+	var err error
+	var allErrs []error
+	for _, path := range paths {
+		cc, err = createGRPCConnection(path, timeout, opts...)
```
chaunceyjiang, line 144: LGTM

pkg/estimator/client/service.go
zhzhuang-zju: Yes, for `karmada-operator`, there needs to be better support for the API `DNSDomain`.

zhzhuang-zju: > however, this is just one of them. I have seen such usage mainly concentrated on the deployment of etcd.

@whitewindmills  The `karmadas.operator.karmada.io` API provides a field for custom DNSDomain, but it is not actually used. I will track this through an issue. Besides this, do you have any review comments on this pull request?

RainbowMango (commented):
/assign

RainbowMango (approved):
/approve

PR #5475, opened 2024-09-03
MultiClusterService: use patch instead of update for change in servic…

a7i:
…e label/annotations

**What type of PR is this?**

/kind feature

Using patch instead of update for labels/annotations will likely reduce conflicts / race conditions

Fixes #

```release-note

```

### DIFF
--- a/pkg/controllers/multiclusterservice/mcs_controller.go
+++ b/pkg/controllers/multiclusterservice/mcs_controller.go
@@ -19,6 +19,7 @@ package multiclusterservice
 import (
 	"context"
 	"fmt"
+	"reflect"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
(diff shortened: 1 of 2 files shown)


pkg/controllers/multiclusterservice/mcs_controller_test.go
```diff
@@ -0,0 +1,147 @@
+/*
+Copyright 2023 The Karmada Authors.
```
whitewindmills, line 2:
```go
Copyright 2024 The Karmada Authors.
```
I prefer this to be a cleanup.
/remove-kind feature
/kind cleanup

pkg/controllers/multiclusterservice/mcs_controller.go
```diff
@@ -337,8 +338,8 @@ func (c *MCSController) retrieveService(ctx context.Context, mcs *networkingv1al
 		delete(svcCopy.Labels, networkingv1alpha1.MultiClusterServicePermanentIDLabel)
 	}
 
-	if err = c.Client.Update(ctx, svcCopy); err != nil {
```
RainbowMango, line 340: Is there a sign that here conflicts offen?
I suppose the Service won't be updated very frequently, right?

Issue #5470, opened 2024-09-02
Project Plan: Enhance Karmada controller-manager and  schedule testing coverage (LFX 2024 TERM 3)


anujagrawal699:
# Project Plan

**Objective:**
To improve the unit test coverage for the specified directories in the Karmada repository.

**Plan:**

1. **Project Understanding**
   - Understand the functionality and interdependencies within the karmada-controller-manager and karmada-scheduler components.
   - Document the critical code paths that are essential for the core functionality of each module.
   - Analyze the current test coverage reports and existing test cases.

2. **Test Coverage Enhancement Strategy**
   - **Target Directories:**
     - pkg/controllers (Currently 22.1%) (Estimated completion for the package 25 September)
     - pkg/detector (Currently 19.58%)
     - pkg/scheduler (Currently 52.36%)
     - pkg/descheduler (Currently 65.37%)
     - pkg/util (Currently 59.35%)
     - pkg/estimator (Currently 42.46%)
     - pkg/dependenciesdistributor (Currently 51.36%)
     - pkg/resourceinterpreter (Currently 46.22%)
   
   - **Critical Logic Testing:**
     - Focus on core business logic and test the functions needed without increasing the maintenance cost.
   
   - **Testing Methodologies:**
     - Unit Testing: For individual functions, methods, and classes focusing on isolated code units.
     - Table-Driven Tests: Implement table-driven tests where possible to efficiently test multiple scenarios using test cases.

3. **Test Implementation**
   - **Directory-by-Directory Implementation:**
     - Begin with the most critical directories, i.e., pkg/controllers and the one with the least coverage, i.e., pkg/detector.
     - Move to the next directory in the suggested order after completing testing for each directory.

   - **Continuous Improvement:**
     - Regularly review new tests to ensure they are meaningful, efficient, and cover the intended scenarios.
     - Refactor old code and tests to remove redundancy and improve maintainability.

4. **Evaluation**
   - Perform a mid-point review of the coverage reports to ensure that the UT coverage is on track to increase by more than 15%.
   - Optimize any tests that might be redundant or overly complex to maintain test efficiency.

5. **Expected Outcomes**
   - **Increased Coverage:**
     - Achieve a significant increase in UT coverage, around 60% coverage overall.
   
   - **Improved Code Quality:**
     - Enhance the quality and maintainability of the Karmada codebase, reducing the likelihood of defects.

**Mentors:** @XiShanYongYe-Chang , @zhzhuang-zju 
[...]

PR #5481, opened 2024-09-04
Added tests for cronfederatedhpa controller

anujagrawal699:
**Description:**
This PR introduces tests for cronfederatedhpa controller. This PR tests the cronfederatedhpa_handler and some functions of cronfederatedhpa_job module. The cronfederatedhpa_controller module is harder and complex to test and we could approach it after completion of tests which are necessary, maintainable and easy.

**Test Coverage:**
The test coverage of the directory _pkg/controllers/cronfederatedhpa_ has been increased from 0.00% to 27.8%. 

**What type of PR is this?**
/kind documentation
/kind failing-test
/kind feature

This PR adds comprehensive tests for testing functionality of cronfederatedhpa controller.

Fixes a part of #5470 

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/cronfederatedhpa/cronfederatedhpa_handler_test.go
+++ b/pkg/controllers/cronfederatedhpa/cronfederatedhpa_handler_test.go
@@ -1,0 +1,298 @@
+/*
+Copyright 2024 The Karmada Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cronfederatedhpa
+
+import (
+	"testing"
+	"time"
+
+	"github.com/go-co-op/gocron"
+	"github.com/stretchr/testify/assert"
+	autoscalingv2 "k8s.io/api/autoscaling/v2"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/record"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	autoscalingv1alpha1 "github.com/karmada-io/karmada/pkg/apis/autoscaling/v1alpha1"
+	"github.com/karmada-io/karmada/pkg/util/helper"
+)
+
+func TestCronFHPAScaleTargetRefUpdates(t *testing.T) {
+	tests := []struct {
+		name           string
+		cronFHPAKey    string
+		initialTarget  autoscalingv2.CrossVersionObjectReference
+		updatedTarget  autoscalingv2.CrossVersionObjectReference
+		expectedUpdate bool
+	}{
+		{
+			name:          "New scale target",
+			cronFHPAKey:   "default/new-cronhpa",
+			initialTarget: autoscalingv2.CrossVersionObjectReference{}, // Empty for new target
+			updatedTarget: autoscalingv2.CrossVersionObjectReference{
+				Kind:       "Deployment",
+				Name:       "test-deployment",
+				APIVersion: "apps/v1",
+			},
+			expectedUpdate: false,
+		},
+		{
+			name:        "Same scale target",
+			cronFHPAKey: "default/test-cronhpa",
+			initialTarget: autoscalingv2.CrossVersionObjectReference{
+				Kind:       "Deployment",
+				Name:       "test-deployment",
+				APIVersion: "apps/v1",
+			},
+			updatedTarget: autoscalingv2.CrossVersionObjectReference{
+				Kind:       "Deployment",
+				Name:       "test-deployment",
+				APIVersion: "apps/v1",
+			},
+			expectedUpdate: false,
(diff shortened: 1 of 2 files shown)


anujagrawal699:
@zhzhuang-zju

pkg/controllers/cronfederatedhpa/cronfederatedhpa_handler_test.go
```diff
@@ -0,0 +1,300 @@
...
+				Kind:       "Deployment",
+				Name:       "test-deployment",
+				APIVersion: "apps/v1",
+			},
+			updatedTarget: autoscalingv2.CrossVersionObjectReference{
+				Kind:       "Deployment",
+				Name:       "test-deployment",
+				APIVersion: "apps/v1",
+			},
+			expectedUpdate: false,
```
zhzhuang-zju, line 72: What's the difference between test case `New scale target` and `Same scale target`?

anujagrawal699: Yeah they are redundant. I thought to test both return statements with it 
1. If there's no existing target for the given key, it stores the new target and returns false
2. If there is an existing target, it compares it with the new target and returns whether they are different.
I think we should update the _New Scale Target_ test to start with no initial target in the handler's map.

zhzhuang-zju: maybe we can use different cronFHPAKey to achieve the effect of a new scale target

pkg/controllers/cronfederatedhpa/cronfederatedhpa_handler_test.go
```diff
@@ -0,0 +1,300 @@
...
+				Name:       "test-statefulset",
+				APIVersion: "apps/v1",
+			},
+			expectedUpdate: true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			handler.CronFHPAScaleTargetRefUpdates(tt.cronFHPAKey, tt.initialTarget)
```
zhzhuang-zju, line 93: Is there a possibility of conflict when using the same handler for different test cases?

anujagrawal699: Yeah, what you said makes sense. I'll update the tests to ensure that each test case uses a fresh handler instance.

anujagrawal699: but the handler map has cronFHPAKey as the keys

pkg/controllers/cronfederatedhpa/cronfederatedhpa_handler_test.go
anujagrawal699: Okay, good idea. I think different keys also won't cause conflicts.

pkg/controllers/cronfederatedhpa/cronfederatedhpa_handler_test.go
anujagrawal699: what do u suggest @zhzhuang-zju ?

zhzhuang-zju: > but the handler map has cronFHPAKey as the keys

sorry, I didn't get it. Could you explain what issue this might cause?

anujagrawal699: I meant using different cronFHPAKey values for each test case, even with the same handler instance, should not lead to conflicts. What do u say?

zhzhuang-zju: This does indeed avoid conflicts, but using different cronFHPAKeys also increases maintenance costs in the long run.

anujagrawal699: Okay, i'll use a sepearate cronFHPAKey for new scale target and we'll create a new handler for each test case. Is is good?

zhzhuang-zju: yeah~make sense to me

anujagrawal699: Thanks!

anujagrawal699:
@zhzhuang-zju

anujagrawal699:
@zhzhuang-zju Please take a look.

zhzhuang-zju (commented):
others LGTM
cc @XiShanYongYe-Chang for another look

pkg/controllers/cronfederatedhpa/cronfederatedhpa_handler_test.go
```diff
@@ -0,0 +1,301 @@
...
+			timeZone: stringPtr("America/New_York"),
+			wantErr:  false,
+		},
+		{
+			name: "Valid rule with invalid time zone",
+			rule: autoscalingv1alpha1.CronFederatedHPARule{
+				Name:     "test-rule-invalid-tz",
+				Schedule: "*/5 * * * *",
+			},
+			timeZone: stringPtr("Invalid/TimeZone"),
```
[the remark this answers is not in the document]
anujagrawal699: Okay, i'll update it.

anujagrawal699:
@XiShanYongYe-Chang CI failed.

anujagrawal699:
@XiShanYongYe-Chang PTAL, Thanks.

zhzhuang-zju:
/lgtm
cc @XiShanYongYe-Chang

XiShanYongYe-Chang (commented):
Thanks~
/approve

PR #5490, opened 2024-09-07
feat: skip collection of resource status to Karmada control plane

### LINKED ISSUE #5489: service propagation removes status field in host cluster
**What happened**:

- create svc in host (load balancer that stores the lb name in the status)
- add propagation to 1 member cluster
- karmada updates status to `{loadBalancer: {}}`

it not to touch the status

I can see 2 requests by karmada:
- 1 from ClaimPolicyForObject adding annotation and label
- 1 from ??? removing the status (sending the full service object but not the status) to status subresource
image

tried disabling service import/export/mcs but still happens

**Environment**:
- Karmada version: v1.11.0

a7i:
**What type of PR is this?**

/kind feature

Fixes https://github.com/karmada-io/karmada/issues/5489

While this does not represent a regular karmada setup (i.e. federated control-plane and member clusters), it doesn't introduce any regression or complexity of retaining that. It also enables using karmada in different setup mechanisms where it can be installed on a "source" cluster with existing resources/controllers.

```release-note

```

### DIFF
--- a/cmd/controller-manager/app/options/options.go
+++ b/cmd/controller-manager/app/options/options.go
@@ -157,6 +157,10 @@ type Options struct {
 	// in scenario of dynamic replica assignment based on cluster free resources.
 	// Disable if it does not fit your cases for better performance.
 	EnableClusterResourceModeling bool
+	// SkipStatusCollection specifies whether to bypass the aggregation and collection of resource statuses.
+	// By default, it is set to false, meaning resource statuses will reflect in the federated control-plane.
+	// When set to true, the aggregated resource statuses will not reflect.
+	SkipStatusCollection bool
 }
 
 // NewOptions builds an empty options.
@@ -246,6 +250,7 @@ func (o *Options) AddFlags(flags *pflag.FlagSet, allControllers, disabledByDefau
 	flags.BoolVar(&o.EnableClusterResourceModeling, "enable-cluster-resource-modeling", true, "Enable means controller would build resource modeling for each cluster by syncing Nodes and Pods resources.\n"+
 		"The resource modeling might be used by the scheduler to make scheduling decisions in scenario of dynamic replica assignment based on cluster free resources.\n"+
 		"Disable if it does not fit your cases for better performance.")
+	flags.BoolVar(&o.SkipStatusCollection, "skip-status-collection", false, "Skip status collection. By default, it is false which means that resource status will be reflected in federated control-plane.")
 
 	o.RateLimiterOpts.AddFlags(flags)
 	o.ProfileOpts.AddFlags(flags)
--- a/cmd/controller-manager/app/controllermanager.go
+++ b/cmd/controller-manager/app/controllermanager.go
@@ -370,26 +370,28 @@ func startBindingController(ctx controllerscontext.Context) (enabled bool, err e
 
 func startBindingStatusController(ctx controllerscontext.Context) (enabled bool, err error) {
 	rbStatusController := &status.RBStatusController{
-		Client:              ctx.Mgr.GetClient(),
-		DynamicClient:       ctx.DynamicClientSet,
-		InformerManager:     ctx.ControlPlaneInformerManager,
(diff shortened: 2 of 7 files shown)


a7i:
Alternatively, investigating usage of AggregateStatus interpreter
image

RainbowMango (commented):
/assign
Thanks for bringing this up.
I have been thinking about this feature for a long time. Status collection brings concerns of performance, and that is not necessary in some cases.

I will focus on whether we need more fine-grained control over status collection.

cmd/controller-manager/app/options/options.go
```diff
@@ -246,6 +250,7 @@ func (o *Options) AddFlags(flags *pflag.FlagSet, allControllers, disabledByDefau
 	flags.BoolVar(&o.EnableClusterResourceModeling, "enable-cluster-resource-modeling", true, "Enable means controller would build resource modeling for each cluster by syncing Nodes and Pods resources.\n"+
 		"The resource modeling might be used by the scheduler to make scheduling decisions in scenario of dynamic replica assignment based on cluster free resources.\n"+
 		"Disable if it does not fit your cases for better performance.")
+	flags.BoolVar(&o.SkipStatusCollection, "skip-status-collection", false, "Skip status collection. By default, it is false which means that resource status will be reflected in federated control-plane.")
```
RainbowMango, line 253: This flag just suspends status aggregation, that is `AggregateStatus` from the flow of the interpreter, but the status is still collected to `Work` and `ResourceBinding`, do you think that is expected? Or would this flag be misleading to people?

a7i: Makes sense. `--skip-status-aggregation`?

RainbowMango: Obviously, `--skip-status-aggregation` is much better.

The component flag is a kind of  `global` configuration, once this `--skip-status-aggregation` is set to `true`, that means skip status aggregation to all types of resources. 

So, I'm thinking if we need to provide finer control, for example, introduce some fields to the [PropagationPolicy Suspension](https://github.com/karmada-io/karmada/blob/07e297681c40e9dddce8455a2103684a261ffc64/pkg/apis/policy/v1alpha1/propagation_types.go#L218-L232).

I want to hear your thoughts.

a7i: In our unique setup, a global flag makes sense because we never want to aggregate status back to the "source"/blue cluster.

But I could see the use-case of it being at the policy level for karmada regular (as a federated-api) install mode.

```
type Suspension struct {
	// Dispatching controls whether dispatching should be suspended.
	// nil means not suspend, no default value, only accepts 'true'.
	// Note: true means stop propagating to all clusters. Can not co-exist
	// with DispatchingOnClusters which is used to suspend particular clusters.
	// +optional
	Dispatching *bool `json:"dispatching,omitempty"`
	
+	// Skip Status Aggregation controls whether collecting status from
+     // member clusters should be skipped.
+	// nil means to not skip, no default value, only accepts 'true'.
+	// Note: true means stop collecting status from all member clusters.
+	// +optional
+	SkipStatusAggregation *bool `json:"skipStatusAggregation,omitempty"`
```

I don't know if there is a use-case to say `skipStatusCollectionFromClusters`

RainbowMango: That's the idea. 
Let's say, here are two approaches:
- The first one is the global flags `--skip-status-aggregation` 
- The second one is the API way.

> In our unique setup, a global flag makes sense because we never want to aggregate status back to the "source"/blue cluster.

If we go with the second approach, you have to configure it to your each PropagationPolicy, is it a burden for you?

> I don't know if there is a use-case to say skipStatusCollectionFromClusters

I don't think so, so far.

RainbowMango: OK, then, I would prefer to go with the second approach, as it is more flexible. 
Do you think we can have a proposal to address the possible use case, the effect on performance, the features that rely on the status collection, and so on?

PR #5493, opened 2024-09-07
Added tests for controllers/deploymentreplicassyncer

anujagrawal699:
**Description:**
This PR introduces tests for deploymentreplicassyncer controller. The new addition tests significant logic of deploymentreplicassyncer controller.

**Additions**:
1. TestPredicateFunc:
   - Added tests for various scenarios including label changes, replica changes, and different event types (Create, Update, Delete, Generic).
   - Ensures the predicate function correctly filters relevant events.

2. TestSetupWithManager:
   - Verifies that the controller can be properly set up with a manager.

3. TestReconcile:
   - Added test cases for different reconciliation scenarios:
     - Successful replicas sync
     - Binding not found
     - Non-divided scheduling type
     - Replicas already synced
   - Checks for correct result, error handling, and replica updates.

4. TestIsDeploymentStatusCollected:
   - Covers various scenarios:
     - Fully collected status
     - Mismatched replicas
     - Generation mismatches
     - Incomplete aggregated status
     - Invalid JSON in status
     - Zero replicas in status
     - Mismatched status replicas

**Test Coverage**:
The test coverage of the deploymentreplicassyncer controller has been increased to 80.60% .

**What type of PR is this?**
/kind feature

Fixes a part of #5470 

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/deploymentreplicassyncer/deployment_replicas_syncer_controller_test.go
+++ b/pkg/controllers/deploymentreplicassyncer/deployment_replicas_syncer_controller_test.go
@@ -1,0 +1,596 @@
+/*
+Copyright 2024 The Karmada Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package deploymentreplicassyncer
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	appsv1 "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	policyv1alpha1 "github.com/karmada-io/karmada/pkg/apis/policy/v1alpha1"
+	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
+	"github.com/karmada-io/karmada/pkg/util"
+	"github.com/karmada-io/karmada/pkg/util/names"
+)
+
+// predicateTestCase defines the structure for testing predicate functions
+type predicateTestCase struct {
+	name      string
+	oldObj    *appsv1.Deployment
+	newObj    *appsv1.Deployment
+	expected  bool
+	eventType string
+}
+
+func TestPredicateFunc(t *testing.T) {
+	testCases := []predicateTestCase{
+		{
+			name: "retain-replicas label added",
+			oldObj: &appsv1.Deployment{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{},
+				},
+			},
+			newObj: &appsv1.Deployment{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{util.RetainReplicasLabel: util.RetainReplicasValue},
+				},
+			},
+			expected:  true,
+			eventType: "update",
+		},
+		{
+			name: "replicas changed",
+			oldObj: &appsv1.Deployment{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{util.RetainReplicasLabel: util.RetainReplicasValue},
+				},
+				Spec: appsv1.DeploymentSpec{
+					Replicas: ptr.To[int32](3),
+				},
+				Status: appsv1.DeploymentStatus{
+					Replicas: 3,
+				},
+			},
+			newObj: &appsv1.Deployment{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{util.RetainReplicasLabel: util.RetainReplicasValue},
+				},
+				Spec: appsv1.DeploymentSpec{
+					Replicas: ptr.To[int32](3),
+				},
+				Status: appsv1.DeploymentStatus{
+					Replicas: 4,
+				},
+			},
+			expected:  true,
+			eventType: "update",
+		},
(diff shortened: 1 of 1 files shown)


anujagrawal699:
The TestSetupWithManager is removed in the latest commit because it was causing CI failures and provided limited value in terms of testing the controller's core functionality.

anujagrawal699:
@XiShanYongYe-Chang Please take a look.

pkg/controllers/deploymentreplicassyncer/deployment_replicas_syncer_controller_test.go
```diff
@@ -0,0 +1,496 @@
...
+					ObjectOld: tc.oldObj,
+					ObjectNew: tc.newObj,
+				})
+			}
+			assert.Equal(t, tc.expected, result)
+		})
+	}
+}
+
+func TestReconcile(t *testing.T) {
```
zhzhuang-zju, line 152:
```bash
=== RUN   TestReconcile/binding_not_found
I1111 14:56:55.825772 4025449 deployment_replicas_syncer_controller.go:107] no need to update deployment replicas for binding not found
=== RUN   TestReconcile/non-divided_scheduling_type
I1111 14:56:55.826253 4025449 deployment_replicas_syncer_controller.go:107] no need to update deployment replicas for binding not found
=== RUN   TestReconcile/replicas_already_synced
I1111 14:56:55.826684 4025449 deployment_replicas_syncer_controller.go:107] no need to update deployment replicas for binding not found
```

The three test cases for this scenario actually only test the `binding not found` situation. Additionally, they lack a test case for a successful `Reconcile`.

anujagrawal699: Hi @zhzhuang-zju , after multiple attempts, i think that the successful reconcile case is very complex to test. Instead of unit tests, this functionality would be better verified through integration tests. I think we should remove the whole TestReconcile.

zhzhuang-zju: If it's difficult to construct a successful scenario, we can temporarily skip it. What about other scenarios?

anujagrawal699: I think we should remove the redundant test cases since they're not actually testing different scenarios, and keep only the "binding not found" case for now. The test coverage is also not improved by other test cases.

anujagrawal699: I'll simplify the test.

zhzhuang-zju: >  keep only the "binding not found" case for now. The test coverage is also not improved by other test cases.

Can we keep the previous two scenarios? Why won't the coverage improve?

anujagrawal699: The other test cases are already not containing the binding and thus the coverage is not improved. If you say i could add the previous two scenarios but a single test case is cleaner than having multiple test cases that effectively test the same condition. What do you think?

zhzhuang-zju: In fact, the `binding was not found` because the binding name in the test case was set incorrectly. It should be `test-deployment-deployment`, not `deployment-test-deployment`.

```bash
=== RUN   TestReconcile/binding_not_found
I1112 16:55:59.382569  110110 deployment_replicas_syncer_controller.go:107] no need to update deployment replicas for binding not found
=== RUN   TestReconcile/non-divided_scheduling_type
=== RUN   TestReconcile/replicas_already_synced
I1112 16:55:59.387745  110110 deployment_replicas_syncer_controller.go:129] replicas in spec field (3) already equal to in status field (3)
```

anujagrawal699: ohh, sorry. Thanks!

anujagrawal699: That was the reason why successful reconcile test was failing. I'll also add test case for successful reconcile now. Thanks!

zhzhuang-zju: It would be better to use `func names.GenerateBindingName` to generate the name of binding

zhzhuang-zju:
/lgtm
thanks cc @XiShanYongYe-Chang

pkg/controllers/deploymentreplicassyncer/deployment_replicas_syncer_controller_test.go
```diff
@@ -0,0 +1,496 @@
...
+					AggregatedStatus: []workv1alpha2.AggregatedStatusItem{
+						{
+							ClusterName: "cluster1",
+							Status:      &runtime.RawExtension{Raw: []byte(`{"replicas": 2}`)},
+						},
+					},
+				},
+			},
+			expected: false,
+		},
```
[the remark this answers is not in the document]
zhzhuang-zju: @anujagrawal699 Is this completed, or did I miss?

XiShanYongYe-Chang:
Thanks~
/assign

pkg/controllers/deploymentreplicassyncer/deployment_replicas_syncer_controller_test.go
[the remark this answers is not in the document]
anujagrawal699: I forgot it, Thanks!

pkg/controllers/deploymentreplicassyncer/deployment_replicas_syncer_controller_test.go
```diff
@@ -0,0 +1,585 @@
...
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			var result bool
+			switch {
+			case tc.name == "create event":
+				result = predicateFunc.Create(event.CreateEvent{Object: tc.newObj})
+			case tc.name == "delete event":
+				result = predicateFunc.Delete(event.DeleteEvent{Object: tc.oldObj})
+			case tc.name == "generic event":
```
XiShanYongYe-Chang, line 140: Generally, the name is read-only and may be changed frequently in the future. If the switch case statement is used, it is recommended that a type field be added to indicate the name.

XiShanYongYe-Chang (commented):
Other parts lgtm, thanks~

pkg/controllers/deploymentreplicassyncer/deployment_replicas_syncer_controller_test.go
anujagrawal699: I'll add an event type structure.

XiShanYongYe-Chang (commented):
/lgtm
/approve

Issue #5491, opened 2024-09-07
[LFX-Mentorship-Term3-2024-Tracking-Issue] Enhance the test coverage for the Karmada search, operator, and webhook components


mohamedawnallah:
## Upstream Issue

#5236

## Legend

- ⬜: No PR issued yet
- ⏳: In Review
- ✅: Merged

## Milestone 1 - Karmada Webhooks Test Coverage

| Status | Period             | Directory                              | Old Coverage | New Coverage | PR | Code Quality | Docs Quality | Bug Discovery |
|----------------|--------------------|----------------------------------------|------------------|------------------|------------------|------------------|------------------|------------------|
| <div align="center">✅</div>      | Sep 7-13           | pkg/webhook/propagationpolicy          | <div align="center">0%</div>               |   <div align="center">N/A</div> | <div align="center">#5492</div> | <div align="center">#5537</div> | <div align="center">N/A</div> | <div align="center">N/A</div> |
| <div align="center">✅</div>     | Sep 7-13           | pkg/webhook/clusterpropagationpolicy   | <div align="center">0%</div>               | <div align="center">N/A</div> | <div align="center">#5494</div> | <div align="center">#5521</div> | <div align="center">N/A</div> | <div align="center">N/A</div> |
| <div align="center">✅</div>      | Sep 7-13           | pkg/webhook/overridepolicy             | <div align="center">0%</div>               | <div align="center">N/A</div> | <div align="center">#5495</div> | <div align="center">#5508 #5521</div> | <div align="center">N/A</div> | <div align="center">N/A</div> |
| <div align="center">✅</div>      | Sep 7–13           | pkg/webhook/clusteroverridepolicy      | <div align="center">0%</div>               | <div align="center">N/A</div> | <div align="center">#5496</div> | <div align="center">#5508</div> | <div align="center">N/A</div> | <div align="center">N/A</div> |
| <div align="center">✅</div>      | Sep 7-13           | pkg/webhook/resourcebinding            | <div align="center">0%</div>               | <div align="center">N/A</div> | <div align="center">#5497</div> | <div align="center">N/A</div> | <div align="center">N/A</div> | <div align="center">N/A</div> |
| <div align="center">✅</div>      | Sep 7-13           | pkg/webhook/clusterresourcebinding     | <div align="center">0%</div>               | <div align="center">N/A</div> | <div align="center">#5498</div> | <div align="center">N/A</div> | <div align="center">N/A</div> | <div align="center">N/A</div> |
[...]

PR #5494, opened 2024-09-09
pkg/webhook: unit test cluster propagation policy

mohamedawnallah:
**Description**

In this commit, we introduce unit tests for both the `ValidatingAdmission` and `MutatingAdmission` webhooks for the `ClusterPropagationPolicy` resource.

The tests include:

- Validation webhook tests:
  - Tests how the webhook responds when decoding the request object fails.
  - Ensures updates to the `SchedulerName` field are denied, as it should remain immutable.
  - Validates that changes to the `ClusterPropagationPolicyPermanentIDLabel` label and its absence in creation requests are correctly handled.
  - Confirms that valid `ClusterPropagationPolicy` objects are allowed through without errors.

- Mutation webhook tests:
  - Provides full coverage of cluster policy mutation with default values and patch application.

**What type of PR is this?**

/kind failing-test
/kind feature

Part of #5491

```release-note
NONE
```

### DIFF
--- a/pkg/webhook/clusterpropagationpolicy/mutating_test.go
+++ b/pkg/webhook/clusterpropagationpolicy/mutating_test.go
@@ -1,0 +1,222 @@
+/*
+Copyright 2024 The Karmada Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package clusterpropagationpolicy
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"net/http"
+	"reflect"
+	"testing"
+
+	admissionv1 "k8s.io/api/admission/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


mohamedawnallah:
Closing and reopening some flaky e2e failures...

pkg/webhook/clusterpropagationpolicy/mutating_test.go
```diff
@@ -0,0 +1,218 @@
...
+	// Instantiate the mutating handler.
+	mutatingHandler := NewMutatingHandler(
+		notReadyTolerationSeconds, unreachableTolerationSeconds, decoder,
+	)
+
+	// Call the Handle function.
+	got := mutatingHandler.Handle(context.Background(), req)
+
+	// Verify that the only patch applied is for the UUID label. If any other patches are present, it indicates that the policy was not handled as expected.
+	if len(got.Patches) > 0 {
```
XiShanYongYe-Chang, line 207: Is the length of got.Patches only equal to 1?

XiShanYongYe-Chang: The logic of the previously incorporated code is the same. I have read it now and may need to be adjusted.

mohamedawnallah: Yes makes sense. I was thinking the same when working on unit testing the Work (#5532) PR to integrate it in Milestone 6 for refinement but let's do it now :)

mohamedawnallah: All is done :)

mohamedawnallah: > The logic of the previously incorporated code is the same. I have read it now and may need to be adjusted.

Everything is completed on both the current unmerged and merged PRs, with follow-up PRs :)

XiShanYongYe-Chang (commented):
Thanks
/lgtm
/approve

PR #5495, opened 2024-09-09
pkg/webhook: unit test override policy

mohamedawnallah:
**Description**

In this commit, we introduce unit tests for both the `ValidationAdmission` and `MutatingAdmission` webhooks specifically for the `OverridePolicy` resource.

The tests include:

- Validation webhook tests:
  - Tests the behavior when decoding the request object fails, verifying that admission is denied with an appropriate error message.
  - Validates that the webhook denies admission when encountering invalid label values in the `OverridePolicy` spec, ensuring proper error messaging.
  - Confirms that valid `OverridePolicy` objects are admitted without errors.

- Mutation webhook tests:
  - Provides full coverage of policy mutation, including setting default namespaces and validating no unnecessary patches.

**What type of PR is this?**

/kind failing-test
/kind feature

Part of #5491

```release-note
NONE
```

### DIFF
--- a/pkg/webhook/overridepolicy/mutating_test.go
+++ b/pkg/webhook/overridepolicy/mutating_test.go
@@ -1,0 +1,160 @@
+/*
+Copyright 2024 The Karmada Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


XiShanYongYe-Chang (commented):
Thanks~
/lgtm
/approve

PR #5499, opened 2024-09-09
pkg/webhook: unit test resource deletion protection

mohamedawnallah:
**Description**

In this commit, we add unit tests for the `ValidatingAdmission` validation webhook specifically for the resource deletion protection feature. These tests ensure that the webhook correctly handles various admission scenarios:

- Tests the behavior when decoding the request object fails, verifying that admission is denied with an appropriate error message.
- Ensures that resources labeled with `DeletionProtectionAlways` are denied deletion, validating that the error message is returned.
- Confirms that resources without the deletion protection label are allowed to be deleted without errors.
- Validates that resources with other labels are not impacted and are allowed to be deleted.
- Confirms that non-delete operations (e.g., create) are allowed by default.

**What type of PR is this?**

/kind failing-test
/kind feature

Part of #5491

```release-note
NONE
```

### DIFF
--- a/pkg/webhook/resourcedeletionprotection/validating_test.go
+++ b/pkg/webhook/resourcedeletionprotection/validating_test.go
@@ -1,0 +1,167 @@
+/*
+Copyright 2024 The Karmada Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


XiShanYongYe-Chang (commented):
Thanks~
/lgtm
/approve

PR #5507, opened 2024-09-09
pkg/webhook: unit test Federated HPA

mohamedawnallah:
**Description**

In this commit, we introduce unit tests for both the `MutatingAdmission` and `ValidatingAdmission` webhooks specifically for the `FederatedHPA` resource.

The tests include:

- Validation webhook tests:
  - Verifies that admission is denied with an appropriate error message when decoding the request object fails.
  - Ensures admission is denied when the `FederatedHPA` spec contains invalid values, such as a `minReplicas` set to zero, and the correct error message is returned.
  - Confirms that a valid `FederatedHPA` object is allowed through without errors.

- Mutation webhook tests:
  - Verifies that admission is denied when decoding the request object fails, returning the appropriate error message.
  - Ensures that the `FederatedHPA` object is correctly mutated when provided with default values, including scale policies and CPU utilization.
  - Confirms that no patches are applied when the `FederatedHPA` object is handled correctly and that admission is allowed for valid `FederatedHPA` objects.

**What type of PR is this?**

Part of #5491.

```release-note
NONE
```

### DIFF
--- a/pkg/webhook/federatedhpa/mutating_test.go
+++ b/pkg/webhook/federatedhpa/mutating_test.go
@@ -1,0 +1,214 @@
+/*
+Copyright 2024 The Karmada Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


mohamedawnallah:
Closing and reopening some flaky e2e test failures...

mohamedawnallah:
All tests passed now—awesome! :)

XiShanYongYe-Chang (commented):
Thanks~
/lgtm
/approve

PR #5512, opened 2024-09-10
proposal: introducing a scheduling strategy to customize cluster propagation priorities

chaosi-zju:
**What type of PR is this?**

/kind feature
/kind design

This proposal introduces a new user scenario for scheduling, where users have preferences for clusters with 
varying priorities and want replicas assigned to their preferred clusters.

To support this scenario, we propose two alternative solutions: one is adding `aggregatePreference` to the existing 
`Weighted` replica scheduling strategy, and the other is enhancing the capability of multiple scheduling groups.

Ultimately, we compare these two options and decide on the best implementation to provide a scheduling strategy 
that allows users to customize cluster propagation priorities.

Fixes #

```release-note

```

### DIFF
--- a/docs/proposals/scheduling/aggregate-by-priority/aggregate-by-priority.md
+++ b/docs/proposals/scheduling/aggregate-by-priority/aggregate-by-priority.md
@@ -1,0 +1,294 @@
+---
+title: Introducing a scheduling strategy to customize cluster propagation priorities
+authors:
+- "@chaosi-zju"
+reviewers:
+- "@TBD"
+approvers:
+- "@RainbowMango"
+
+creation-date: 2024-09-07
+---
+
+# Introducing a scheduling strategy to customize cluster propagation priorities
+
+## Summary
+
+This proposal introduces a new user scenario for scheduling, where users have preferences for clusters with
+varying priorities and want replicas assigned to their preferred clusters.
+
+To support this scenario, we propose a new scheduling strategy that allows users to customize the priority of clusters,
+ensuring that replicas are preferentially allocated to higher-priority clusters.
+
+## Motivation
+
+Karmada currently supports three `Divided` type scheduling strategies, 
+these are `Dynamic Weighted`, `Static Weight` and `Aggregated`.
+However, the current strategies are still not satisfied in certain user scenarios.
+
+<!-- 例如，假设用户有多个集群，用户希望副本优先分配到他偏好的集群，只有当他偏好的集群资源不足时，再分配到其他的集群 -->
+
+For example, the user has multiple clusters, and he wants to use some preferred clusters first.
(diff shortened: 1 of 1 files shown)


docs/proposals/scheduling/aggregate-preference/aggregate-preference.md
```diff
@@ -0,0 +1,256 @@
...
+```go
+type Placement struct {
+    // ...
+    // +optional
+    ClusterAffinities []ClusterAffinityTerm `json:"clusterAffinities,omitempty"`
+
++   // ...
++   // +kubebuilder:validation:Enum=sequential
++   // +optional
++   ClusterAffinitiesOrder ClusterAffinitiesOrder `json:"clusterAffinitiesOrder,omitempty"`
```
zhzhuang-zju, line 205: Could explicit priority declaration for clusters be implemented, considering the following points:
- We can define the priority of clusters as an inherent attribute of the clusters themselves. Various scheduling methods would then act merely as consumers of this data, avoiding binding them directly to any specific scheduling approach.
- The current Solution 2 has coupling between different clusterAffinities, which requires determining the priority of clusters based on the order in which clusters appear within the clusterAffinity. Additionally, it does not allow for distinguishing the priority relationship between clusters that appear for the first time within the same clusterAffinity.

docs/proposals/scheduling/aggregate-by-priority/aggregate-by-priority.md
```diff
@@ -0,0 +1,283 @@
...
+```go
+// ReplicaDivisionPreference describes options of how replicas can be scheduled.
+type ReplicaDivisionPreference string
+
+const (
+    ReplicaDivisionPreferenceAggregated ReplicaDivisionPreference = "Aggregated"
+    ReplicaDivisionPreferenceWeighted ReplicaDivisionPreference = "Weighted"
++	// ReplicaDivisionPreferencePriorityAggregated assigns replicas to higher-priority clusters first, 
++	// then to lower-priority clusters if resources are insufficient.
++	ReplicaDivisionPreferencePriorityAggregated ReplicaDivisionPreference = "PriorityAggregated"
```
RainbowMango, line 100:
```suggestion
	// ReplicaDivisionPreferencePriorityAggregated assigns replicas to higher-priority clusters first, 
	// then to lower-priority clusters if resources are insufficient.
	ReplicaDivisionPreferencePriorityAggregated ReplicaDivisionPreference = "PriorityAggregated"
```

RainbowMango: Does the `PriorityAggregated` still honor the score plugins in the scheduler? In another world, what's the relationship of them?

PR #5525, opened 2024-09-11
Fix occasional suspension case failed

XiShanYongYe-Chang:
**What type of PR is this?**

/kind failing-test

Recently, there have been many e2e failures related to suspension, like this: https://github.com/karmada-io/karmada/actions/runs/10791713495/job/29930012414

It is found that the logic of creating and modifying PP/CPP is too close to each other, which may cause the suspension effect to not be synchronized to the RB/CRB object.

Therefore, I added logic to wait for resources to be distributed to member clusters to avoid them being too close together.

In addition, some adjustments have been made to the E2E structure.

Fixes #

/cc @a7i @chaosi-zju @RainbowMango 

```release-note
NONE
```

### DIFF
--- a/test/e2e/clusterpropagationpolicy_test.go
+++ b/test/e2e/clusterpropagationpolicy_test.go
@@ -1075,8 +1075,7 @@ var _ = ginkgo.Describe("[DeleteCase] ClusterPropagationPolicy testing", func() 
 	})
 })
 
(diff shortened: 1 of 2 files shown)


a7i (approved):
/lgtm

nice one, thank you! 🥇

RainbowMango (approved):
/approve

PR #5530, opened 2024-09-11
skip interpret health of resources without a hook

a7i:
**What type of PR is this?**
/kind feature

In one of our small staging clusters, we get a lot of InterpretHealthFailed events (over 100k per week). After investigations, it seems to be resources such as ClusterRole and ClusterRoleBinding that never have a status, so there is nothing to inspect.

Screenshot 2024-09-11 at 4 21 34 PM

This is also repeated over and over again in karmada-controller-manager logs:

```
"Interpret health of object(ClusterRole//oncall) failed, err: default InterpretHealth interpreter for \"rbac.authorization.k8s.io/v1, Kind=ClusterRole\" not found." logger="events" type="Warning" object={"kind":"Work","namespace":"karmada-es-REDACTED","name":"oncall-76bd899f7f","uid":"7a93a59e-3196-4192-975c-ccdc3dd5e6f7","apiVersion":"work.karmada.io/v1alpha1","resourceVersion":"6603066766"} reason="InterpretHealthFailed"
```

**Which issue(s)

is PR fixes**:
Fixes #

```release-note

```

### DIFF
--- a/pkg/controllers/status/work_status_controller.go
+++ b/pkg/controllers/status/work_status_controller.go
@@ -38,6 +38,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	clusterv1alpha1 "github.com/karmada-io/karmada/pkg/apis/cluster/v1alpha1"
+	configv1alpha1 "github.com/karmada-io/karmada/pkg/apis/config/v1alpha1"
 	workv1alpha1 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha1"
 	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
 	"github.com/karmada-io/karmada/pkg/events"
@@ -365,20 +366,7 @@ func (c *WorkStatusController) reflectStatus(ctx context.Context, work *workv1al
 	}
 	c.EventRecorder.Eventf(work, corev1.EventTypeNormal, events.EventReasonReflectStatusSucceed, "Reflect status for object(%s/%s/%s) succeed.", clusterObj.GetKind(), clusterObj.GetNamespace(), clusterObj.GetName())
 
-	var resourceHealth workv1alpha1.ResourceHealth
-	// When an unregistered resource kind is requested with the ResourceInterpreter,
-	// the interpreter will return an error, we treat its health status as Unknown.
-	healthy, err := c.ResourceInterpreter.InterpretHealth(clusterObj)
-	if err != nil {
-		resourceHealth = workv1alpha1.ResourceUnknown
-		c.EventRecorder.Eventf(work, corev1.EventTypeWarning, events.EventReasonInterpretHealthFailed, "Interpret health of object(%s/%s/%s) failed, err: %s.", clusterObj.GetKind(), clusterObj.GetNamespace(), clusterObj.GetName(), err.Error())
-	} else if healthy {
-		resourceHealth = workv1alpha1.ResourceHealthy
-		c.EventRecorder.Eventf(work, corev1.EventTypeNormal, events.EventReasonInterpretHealthSucceed, "Interpret health of object(%s/%s/%s) as healthy.", clusterObj.GetKind(), clusterObj.GetNamespace(), clusterObj.GetName())
-	} else {
-		resourceHealth = workv1alpha1.ResourceUnhealthy
-		c.EventRecorder.Eventf(work, corev1.EventTypeNormal, events.EventReasonInterpretHealthSucceed, "Interpret health of object(%s/%s/%s) as unhealthy.", clusterObj.GetKind(), clusterObj.GetNamespace(), clusterObj.GetName())
-	}
+	resourceHealth := c.interpretHealth(clusterObj, work)
 
 	identifier, err := c.buildStatusIdentifier(work, clusterObj)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/resourceinterpreter/default/native/default.go
```diff
@@ -159,6 +160,11 @@ func (e *DefaultInterpreter) ReflectStatus(object *unstructured.Unstructured) (s
 
 // InterpretHealth returns the health state of the object.
 func (e *DefaultInterpreter) InterpretHealth(object *unstructured.Unstructured) (bool, error) {
+	if !utilhelper.HasStatus(object) {
+		klog.V(4).Infof("skipping health status of object without status: %v %s/%s.", object.GroupVersionKind(), object.GetNamespace(), object.GetName())
```
a7i, line 164: perhaps should be higher than v4

whitewindmills:
@a7i 
we cannot stop interpreting health for resources without status. see https://github.com/karmada-io/karmada/pull/4453.
anyway, this is definitely worth optimizing, you might consider skipping it if `InterpretHealth` is not enabled.
```go
if !c.ResourceInterpreter.HookEnabled(resourceKey.GroupVersionKind(), configv1alpha1.InterpreterOperationInterpretHealth) {
	// skip interpreting health ....
}
```

pkg/resourceinterpreter/default/native/default.go
XiShanYongYe-Chang: v5?

XiShanYongYe-Chang (commented):
I think it's reasonable.
LGTM

XiShanYongYe-Chang:
> we cannot stop interpreting health for resources without status. see #4453.

Sorry, I don't understand. The current modification does not skip the health parsing step, but directly considers health as true when the object does not have status.

whitewindmills:
> The current modification does not skip the health parsing step, but directly considers health as true when the object does not have status.

it's fake health. we should avoid modify the default interpreter if there are no new features or bug fixes, and we should have called `HookEnabled` before calling `InterpretHealth`, which would have solved the problem as well, wouldn't it?

whitewindmills:
in addition, it seems that we have no regulations that do not interpret the health of resources without status. maybe there will be such resources in the future.

XiShanYongYe-Chang:
> in addition, it seems that we have no regulations that do not interpret the health of resources without status.

That's the point. So logically, this pr modification is correct, but there is an alternative, which is to call the HookEnabled method, which I think is OK. How do you think @a7i ?

RainbowMango:
Thanks @a7i for bringing this up, and thanks to @whitewindmills for the reminder that we have revisited the `IntepretHealth` design at https://github.com/karmada-io/karmada/pull/4453#issuecomment-1863933988.

The agreement we made at that time was: `The resources which do not implement InterpretHealth should be treated as healthy by default`. I was hoping to [make some improvement](https://github.com/karmada-io/karmada/pull/4453#issuecomment-1864077193) after #4453, but unfortunately I forgot it!  (Maybe next time I will require the track PR before approving :))

Speaking of the improvements, I agree with @whitewindmills that we should skip evaluating those resources that don't implement `InterpretHealth`, treat them as healthy by default, and should avoid unexpected, surprising warnings.

a7i:
Thanks for the feedback. Pushed up a commit, let me know what you think!

pkg/controllers/status/work_status_controller.go
```diff
@@ -365,20 +366,7 @@ func (c *WorkStatusController) reflectStatus(ctx context.Context, work *workv1al
...
-	healthy, err := c.ResourceInterpreter.InterpretHealth(clusterObj)
-	if err != nil {
-		resourceHealth = workv1alpha1.ResourceUnknown
-		c.EventRecorder.Eventf(work, corev1.EventTypeWarning, events.EventReasonInterpretHealthFailed, "Interpret health of object(%s/%s/%s) failed, err: %s.", clusterObj.GetKind(), clusterObj.GetNamespace(), clusterObj.GetName(), err.Error())
-	} else if healthy {
-		resourceHealth = workv1alpha1.ResourceHealthy
-		c.EventRecorder.Eventf(work, corev1.EventTypeNormal, events.EventReasonInterpretHealthSucceed, "Interpret health of object(%s/%s/%s) as healthy.", clusterObj.GetKind(), clusterObj.GetNamespace(), clusterObj.GetName())
-	} else {
-		resourceHealth = workv1alpha1.ResourceUnhealthy
-		c.EventRecorder.Eventf(work, corev1.EventTypeNormal, events.EventReasonInterpretHealthSucceed, "Interpret health of object(%s/%s/%s) as unhealthy.", clusterObj.GetKind(), clusterObj.GetNamespace(), clusterObj.GetName())
```
a7i, line 380: this is odd. resource is considered `workv1alpha1.ResourceUnhealthy` but the event is `events.EventReasonInterpretHealthSucceed`

a7i: if it's intended, perhaps a comment is needed explaining why

XiShanYongYe-Chang: EventReasonInterpretHealthSucceed: This event is used to describe whether the parsing is successful. The parsing result is reflected in the message.

Do you have any better suggestions for the type of event?

pkg/controllers/status/work_status_controller.go
```diff
@@ -400,6 +388,30 @@ func (c *WorkStatusController) reflectStatus(ctx context.Context, work *workv1al
...
+		klog.V(5).Infof("skipping health status of objects: %v %s/%s.", clusterObj.GroupVersionKind(), clusterObj.GetNamespace(), clusterObj.GetName())
+	}
+
+	var resourceHealth workv1alpha1.ResourceHealth
+	if err != nil {
+		resourceHealth = workv1alpha1.ResourceUnknown
+		c.EventRecorder.Eventf(work, corev1.EventTypeWarning, events.EventReasonInterpretHealthFailed, "Interpret health of object(%s/%s/%s) failed, err: %s.", clusterObj.GetKind(), clusterObj.GetNamespace(), clusterObj.GetName(), err.Error())
+	} else if healthy {
+		resourceHealth = workv1alpha1.ResourceHealthy
+		c.EventRecorder.Eventf(work, corev1.EventTypeNormal, events.EventReasonInterpretHealthSucceed, "Interpret health of object(%s/%s/%s) as healthy.", clusterObj.GetKind(), clusterObj.GetNamespace(), clusterObj.GetName())
```
XiShanYongYe-Chang, line 407: One question, do we still need to report the event in the skip situation?

pkg/controllers/status/work_status_controller.go
a7i: Oh I see! Now that you put it like it, it makes sense

pkg/controllers/status/work_status_controller.go
a7i: Good question, open to feedback. I personally would prefer if we didn't record the event

whitewindmills:
lgtm
/retest

pkg/controllers/status/work_status_controller.go
XiShanYongYe-Chang: cc @RainbowMango

RainbowMango (commented):
/assign

pkg/controllers/status/work_status_controller.go
```diff
@@ -400,6 +388,30 @@ func (c *WorkStatusController) reflectStatus(ctx context.Context, work *workv1al
...
+	} else {
+		resourceHealth = workv1alpha1.ResourceUnhealthy
+		c.EventRecorder.Eventf(work, corev1.EventTypeNormal, events.EventReasonInterpretHealthSucceed, "Interpret health of object(%s/%s/%s) as unhealthy.", clusterObj.GetKind(), clusterObj.GetNamespace(), clusterObj.GetName())
+	}
+	return resourceHealth
```
RainbowMango, line 412: Is this more clear?
```golang
	// For kind that doesn't have health check, we treat it as healthy.
	if !c.ResourceInterpreter.HookEnabled(clusterObj.GroupVersionKind(), configv1alpha1.InterpreterOperationInterpretHealth) {
		klog.V(5).Infof("skipping health assessment for object: %v %s/%s as missing customization and will treat it as healthy.", clusterObj.GroupVersionKind(), clusterObj.GetNamespace(), clusterObj.GetName())
		return workv1alpha1.ResourceHealthy
	}

	var resourceHealth workv1alpha1.ResourceHealth
	healthy, err := c.ResourceInterpreter.InterpretHealth(clusterObj)
	if err != nil {
		resourceHealth = workv1alpha1.ResourceUnknown
		c.EventRecorder.Eventf(work, corev1.EventTypeWarning, events.EventReasonInterpretHealthFailed, "Interpret health of object(%s/%s/%s) failed, err: %s.", clusterObj.GetKind(), clusterObj.GetNamespace(), clusterObj.GetName(), err.Error())
	} else if healthy {
		resourceHealth = workv1alpha1.ResourceHealthy
		c.EventRecorder.Eventf(work, corev1.EventTypeNormal, events.EventReasonInterpretHealthSucceed, "Interpret health of object(%s/%s/%s) as healthy.", clusterObj.GetKind(), clusterObj.GetNamespace(), clusterObj.GetName())
	} else {
		resourceHealth = workv1alpha1.ResourceUnhealthy
		c.EventRecorder.Eventf(work, corev1.EventTypeNormal, events.EventReasonInterpretHealthSucceed, "Interpret health of object(%s/%s/%s) as unhealthy.", clusterObj.GetKind(), clusterObj.GetNamespace(), clusterObj.GetName())
	}
	return resourceHealth
```

RainbowMango (commented):
Generally looks good to me!
I like the idea that have a dedicated methods for healthy assessment very much! Thank you!

pkg/controllers/status/work_status_controller.go
RainbowMango: @a7i What do you say?

a7i: Much cleaner! thanks for the feedback

RainbowMango (approved):
/lgtm
/approve

PR #5532, opened 2024-09-12
pkg/webhook: unit test Work

mohamedawnallah:
**Description**

In this commit, we introduce unit tests for the `MutatingAdmission` webhook specifically for the `Work` resource. The mutation webhook tests include the following:

- Provides full coverage of policy mutation, including setting default and validating no unnecessary patches.

**What type of PR is this?**

Part of #5491.

```release-note
NONE
```


XiShanYongYe-Chang (commented):
Thanks
/lgtm
/approve

PR #5559, opened 2024-09-16
operator/pkg/certs: unit test Certificate Manager and Karmada Store

mohamedawnallah:
**Description**

In this commit, we introduce unit tests for operator `certs` package. The test cases include but not limited to:

- Test private key generation.
- Test self-signed certificate authority (CA) generation.
- Verify creation of self-signed certificate and key files within the CA.
- Include tests for utility functions, such as appending to Subject Alternative Names in SSL/TLS certificates.
- Add tests for Karmada Store operations related to certificates stored in Kubernetes Secrets:
  - Create
  - Read
  - Update
  - Load

**What type of PR is this?**

**Which issue(s) this PR fixes**:
Part of #5491.

**Dependency**
This PR depends on #5560 and #5561.

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #5561, opened 2024-09-16
operator/pkg/certs: validate X509 Cert key types and improve error handling

mohamedawnallah:
**Description**
In this commit, we enhance the `GeneratePrivateKey` and `ParsePrivateKeyPEM` functions:

- Added validation for unsupported key types and provided a more descriptive error message. This function supports ECDSA (using P-256) and RSA (with a key size of 3072 bits) algorithms. It returns an error for unsupported key types.
- Improved error handling to include the type of the unsupported key format in the error message. This function now provides more informative feedback when the private key format is neither RSA nor ECDSA.

**Motivation and Context**
During the testing of the Certificate Manager and Karmada Store (#5559), it was observed that when unsupported key types such as `UnknownPublicKeyAlgorithm`, `DSA`, or `Ed25519` (as documented in the [Go standard library crypto package](https://pkg.go.dev/crypto/x509#PublicKeyAlgorithm)) are passed to the `GeneratePrivateKey` function, the function incorrectly returns an `RSA` key. Instead, it should return an error when encountering unsupported key types.

**What type of PR is this?**
/kind bug

```release-note
NONE
```

### DIFF
--- a/operator/pkg/certs/certs.go
+++ b/operator/pkg/certs/certs.go
@@ -25,7 +25,6 @@ import (
 	"crypto/x509"
 	"crypto/x509/pkix"
 	"encoding/pem"
-	"errors"
 	"fmt"
 	"math"
 	"math/big"
@@ -247,14 +246,20 @@ func (cert *KarmadaCert) KeyName() string {
 	return pair + keyExtension
 }
 
-// GeneratePrivateKey generates cert key with default size if 1024. it supports
-// ECDSA and RAS algorithm.
+// GeneratePrivateKey generates a certificate key. It supports both
+// ECDSA (using the P-256 elliptic curve) and RSA algorithms. For RSA,
+// the key is generated with a size of 3072 bits. If the keyType is
+// x509.UnknownPublicKeyAlgorithm, the function defaults to generating
+// an RSA key.
 func GeneratePrivateKey(keyType x509.PublicKeyAlgorithm) (crypto.Signer, error) {
-	if keyType == x509.ECDSA {
+	switch keyType {
+	case x509.ECDSA:
 		return ecdsa.GenerateKey(elliptic.P256(), cryptorand.Reader)
+	case x509.RSA, x509.UnknownPublicKeyAlgorithm:
+		return rsa.GenerateKey(cryptorand.Reader, rsaKeySize)
+	default:
+		return nil, fmt.Errorf("unsupported key type: %T, supported key types are RSA and ECDSA", keyType)
 	}
-
-	return rsa.GenerateKey(cryptorand.Reader, rsaKeySize)
 }
 
 // NewCertificateAuthority creates new certificate and private key for the certificate authority
@@ -428,7 +433,7 @@ func ParsePrivateKeyPEM(keyData []byte) (crypto.Signer, error) {
 	case *ecdsa.PrivateKey:
 		key = k
 	default:
-		return nil, errors.New("the private key is neither in RSA nor ECDSA format")
+		return nil, fmt.Errorf("the private key is in an unsupported format: %s, supported formats are RSA and ECDSA", caPrivateKey)
 	}
 
 	return key, nil
--- a/pkg/karmadactl/cmdinit/cert/cert.go
+++ b/pkg/karmadactl/cmdinit/cert/cert.go
@@ -54,13 +54,20 @@ const (
 // NewPrivateKey returns a new private key.
 var NewPrivateKey = GeneratePrivateKey
 
-// GeneratePrivateKey Generate CA Private Key
+// GeneratePrivateKey generates a certificate key. It supports both
+// ECDSA (using the P-256 elliptic curve) and RSA algorithms. For RSA,
+// the key is generated with a size of 3072 bits. If the keyType is
+// x509.UnknownPublicKeyAlgorithm, the function defaults to generating
+// an RSA key.
 func GeneratePrivateKey(keyType x509.PublicKeyAlgorithm) (crypto.Signer, error) {
-	if keyType == x509.ECDSA {
+	switch keyType {
+	case x509.ECDSA:
 		return ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+	case x509.RSA, x509.UnknownPublicKeyAlgorithm:
+		return rsa.GenerateKey(rand.Reader, rsaKeySize)
+	default:
+		return nil, fmt.Errorf("unsupported key type: %T, supported key types are RSA and ECDSA", keyType)
 	}
-
-	return rsa.GenerateKey(rand.Reader, rsaKeySize)
 }
 
 // CertsConfig is a wrapper around certutil.Config extending it with PublicKeyAlgorithm.


zhzhuang-zju:
The method `GeneratePrivateKey` currently uses the `CertConfig.PublicKeyAlgorithm` that is not yet exposed to users. In practice, the operator only uses the algorithm `x509.RSA`. The purpose of this PR is to strictly limit the values of `CertConfig.PublicKeyAlgorithm` to `ECDSA` or `RSA`. Therefore, rather than categorizing this as `/kind bug`, I lean more towards `/kind cleanup`

/remove-kind bug
/kind cleanup

operator/pkg/certs/certs.go
```diff
@@ -428,7 +429,8 @@ func ParsePrivateKeyPEM(keyData []byte) (crypto.Signer, error) {
 	case *ecdsa.PrivateKey:
 		key = k
 	default:
-		return nil, errors.New("the private key is neither in RSA nor ECDSA format")
+		keyType := fmt.Sprintf("%T", caPrivateKey)
```
zhzhuang-zju, line 432: Can you provide an example of what the value of `KeyType` would be when the algorithm is not `RSA` or `ECDSA`?

mohamedawnallah: The `vendor/k8s.io/client-go/util/keyutil` package currently supports returning either `RSA` or `ECDSA` keys, or an error if the key type is not recognized. It does not include support for other key types such as `Ed25519`, though new types may be added in the future. I’ve incorporated this as a defensive measure to handle any potential new key types that might be introduced in the upstream package. What do you think?

https://github.com/karmada-io/karmada/blob/ec0521a754b19043c7f5ef946fbdc575e8c2afc5/vendor/k8s.io/client-go/util/keyutil/key.go#L168-L184

mohamedawnallah: Also for more info about the algorithms could be used it could be `UnknownPublicKeyAlgorithm`, `RSA`, `DSA`, `ECDSA`, or `Ed25519` as appear in the [Go standard library crypto package](https://pkg.go.dev/crypto/x509#PublicKeyAlgorithm).

zhzhuang-zju: >  I’ve incorporated this as a defensive measure to handle any potential new key types that might be introduced in the upstream package. What do you think?

LGTM, and as the type of  `caPrivateKey` is `interface{}`, `%T` should be the best way to represent the type of caPrivateKey.

mohamedawnallah:
@XiShanYongYe-Chang all test cases passed :)

XiShanYongYe-Chang:
cc @zhzhuang-zju

mohamedawnallah:
Hi @XiShanYongYe-Chang and @zhzhuang-zju,

Could we please get this PR merged? The unit test certificates PR (#5625) has been stale for a while. If anything is needed to move it forward, please let me know.

Thanks! 🙏

XiShanYongYe-Chang:
Hi @zhzhuang-zju, can you help take a review?

zhzhuang-zju:
> Hi @zhzhuang-zju, can you help take a review?

sure, review is in progress

operator/pkg/certs/certs.go
```diff
@@ -428,7 +429,8 @@ func ParsePrivateKeyPEM(keyData []byte) (crypto.Signer, error) {
...
+		return nil, fmt.Errorf("the private key is in an unsupported format: %s, supported formats are RSA and ECDSA", keyType)
```
zhzhuang-zju, line 433: Could these two lines merged? like:

```suggestion
		return nil, fmt.Errorf("the private key is in an unsupported format: %T, supported formats are RSA and ECDSA", caPrivateKey)
```

operator/pkg/certs/certs.go
```diff
@@ -247,14 +246,16 @@ func (cert *KarmadaCert) KeyName() string {
...
 
-// GeneratePrivateKey generates cert key with default size if 1024. it supports
-// ECDSA and RAS algorithm.
+// GeneratePrivateKey generates a certificate key. It supports both
+// ECDSA (using the P-256 elliptic curve) and RSA algorithms. For RSA,
+// the key is generated with a size of 3072 bits.
 func GeneratePrivateKey(keyType x509.PublicKeyAlgorithm) (crypto.Signer, error) {
 	if keyType == x509.ECDSA {
 		return ecdsa.GenerateKey(elliptic.P256(), cryptorand.Reader)
+	} else if keyType == x509.RSA {
```
zhzhuang-zju, line 255: https://github.com/karmada-io/karmada/blob/c230b14827d01c09db98b9531e4224d93429ee67/pkg/karmadactl/cmdinit/cert/cert.go#L58-L64

`karmadactl init` has a similar implementation; can this be changed together?

zhzhuang-zju (commented):
Thanks, others LGTM

mohamedawnallah:
Thanks @zhzhuang-zju for the feedback! I've integrated it! Could you please take a look if everything looks good? 🙏

operator/pkg/certs/certs.go
```diff
@@ -247,14 +246,18 @@ func (cert *KarmadaCert) KeyName() string {
...
-// ECDSA and RAS algorithm.
+// GeneratePrivateKey generates a certificate key. It supports both
+// ECDSA (using the P-256 elliptic curve) and RSA algorithms. For RSA,
+// the key is generated with a size of 3072 bits. If the keyType is
+// x509.UnknownPublicKeyAlgorithm, the function defaults to generating
+// an RSA key.
 func GeneratePrivateKey(keyType x509.PublicKeyAlgorithm) (crypto.Signer, error) {
 	if keyType == x509.ECDSA {
 		return ecdsa.GenerateKey(elliptic.P256(), cryptorand.Reader)
+	} else if keyType == x509.RSA || keyType == x509.UnknownPublicKeyAlgorithm {
```
zhzhuang-zju, line 257: Why is an RSA key generated when `keyType` is `x509.UnknownPublicKeyAlgorithm`?

mohamedawnallah: Before adding these checks, the default implementation behaves as follows: when no algorithm is provided, the default value for `x509.PublicKeyAlgorithm` is `x509.UnknownPublicKeyAlgorithm`. In this case, the system falls back to using `RSA` for generating the private key. However, I’m not 100% sure if this is the expected behavior for the current case?

zhzhuang-zju: Currently, neither `karmadactl init` nor `karmada-operator` support customizing the `keyType`, and `karmada init` does not set a default `keyType`. As a result, the `keyType` argument passed to the `GeneratePrivateKey` function is actually `x509.UnknownPublicKeyAlgorithm`.

zhzhuang-zju: However, it is odd to categorize `x509.UnknownPublicKeyAlgorithm` and `x509.RSA` in the same group.

mohamedawnallah: > However, it is odd to categorize x509.UnknownPublicKeyAlgorithm and x509.RSA in the same group.

I don't think I see it odd since if no public key algorithm provided it fall back to `x509.RSA` as the previous implementation already categorized `x509.UnknownPublicKeyAlgorithm` and `x509.RSA` in the same group.

What do you think we should do here? 🤔

PS:
For better syntax update, we could use a `switch` statement to handle this:

```go
switch keyType {
case x509.ECDSA:
	return ecdsa.GenerateKey(elliptic.P256(), cryptorand.Reader)
case x509.RSA, x509.UnknownPublicKeyAlgorithm:
	return rsa.GenerateKey(cryptorand.Reader, rsaKeySize)
default:
	return nil, fmt.Errorf("unsupported key type: %T, supported key types are RSA and ECDSA", keyType)
}
```

zhzhuang-zju: What you said makes sense. This modification will not change the existing functionality, I'm ok with this.

> What you said makes sense. This modification will not change the existing functionality, I'm ok with this.

agree

XiShanYongYe-Chang (commented):
Thanks~
/approve

PR #5562, opened 2024-09-16
pkg/webhook: unit test ResourceInterpreterCustomization

mohamedawnallah:
**Description**

In this commit, we introduce unit tests for `ValidationAdmission` webhook for the `ResourceInterpreterCustomization` resource. The tests include:

- Tests how the webhook responds when decoding the request object fails.
- Tests how the webhook responds when it fails to retrieve list of objects from a namespace.
- Ensures that invalid `ResourceInterpreterCustomization` spec denies the admission.
- Confirms that valid `ResourceInterpreterCustomization` spec are allowed through without errors.

**What type of PR is this?**

Part of #5491.

```release-note
NONE
```

### DIFF
--- a/pkg/webhook/resourceinterpretercustomization/validation_test.go
+++ b/pkg/webhook/resourceinterpretercustomization/validation_test.go
@@ -1,0 +1,195 @@
+/*
+Copyright 2024 The Karmada Authors.
+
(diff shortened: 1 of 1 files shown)


XiShanYongYe-Chang (commented):
Thanks
/lgtm
/approve

PR #5568, opened 2024-09-18
Avoid treating PVC managed by VolumeClaimTemplate as dependencies

jklaw90:
**What type of PR is this?**
/kind bug

It was incorrectly handling PVCs when it belongs to the StatefulSet VolumeClaimTemplates.
If you set the claim name in the to match the StatefulSet VolumeClaimTemplate the StatefulSet controller would replace it for the pod, so the name isn't a backed by a real pvc.

create the following statefulset, and note the pod created has a claim with name `www-web-0`
```yaml
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: web
spec:
  selector:
    matchLabels:
      app: nginx # has to match .spec.template.metadata.labels
  serviceName: "nginx"
  replicas: 3 # by default is 1
  minReadySeconds: 10 # by default is 0
  template:
    metadata:
      labels:
        app: nginx # has to match .spec.selector.matchLabels
    spec:
      volumes:
      - name: www
        persistentVolumeClaim:
          claimName: www
      terminationGracePeriodSeconds: 10
      containers:
      - name: nginx
        image: registry.k8s.io/nginx-slim:0.24
        ports:
        - containerPort: 80
          name: web
        volumeMounts:
        - name: www
          mountPath: /usr/share/nginx/html
  volumeClaimTemplates:
  - metadata:
      name: www
    spec:
      accessModes: [ "ReadWriteOnce" ]
      storageClassName: "my-storage-class"
      resources:
        requests:
          storage: 1Gi

``` 

Fixes # n/a

```release-note
fixed StatefulSet Dependencies with PVCs created via the VolumeClaimTemplates.
```

### DIFF
--- a/pkg/resourceinterpreter/default/native/dependencies.go
+++ b/pkg/resourceinterpreter/default/native/dependencies.go
@@ -27,6 +27,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/sets"
 	mcsv1alpha1 "sigs.k8s.io/mcs-api/pkg/apis/v1alpha1"
 
 	configv1alpha1 "github.com/karmada-io/karmada/pkg/apis/config/v1alpha1"
@@ -132,7 +133,36 @@ func getStatefulSetDependencies(object *unstructured.Unstructured) ([]configv1al
 		return nil, err
 	}
 
-	return helper.GetDependenciesFromPodTemplate(podObj)
+	deps, err := helper.GetDependenciesFromPodTemplate(podObj)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(statefulSetObj.Spec.VolumeClaimTemplates) == 0 {
+		return deps, nil
+	}
+
+	// ignore the PersistentVolumeClaim dependency if it was created by the StatefulSet VolumeClaimTemplates
+	// the PVC dependency is not needed because the StatefulSet will manage the pvcs in the member cluster,
+	// if it exists here it was just a placeholder not a real PVC
+	var validDeps []configv1alpha1.DependentObjectReference
+	volumeClaimTemplateNames := sets.Set[string]{}
+	for i := range statefulSetObj.Spec.VolumeClaimTemplates {
+		volumeClaimTemplateNames.Insert(statefulSetObj.Spec.VolumeClaimTemplates[i].Name)
+	}
+
+	for i := range deps {
+		if deps[i].Kind != util.PersistentVolumeClaimKind {
+			validDeps = append(validDeps, deps[i])
+			continue
+		}
+		if volumeClaimTemplateNames.Has(deps[i].Name) {
+			continue
+		}
+		validDeps = append(validDeps, deps[i])
+	}
+
+	return validDeps, nil
 }
 
 func getIngressDependencies(object *unstructured.Unstructured) ([]configv1alpha1.DependentObjectReference, error) {
--- a/pkg/resourceinterpreter/default/native/dependencies_test.go
+++ b/pkg/resourceinterpreter/default/native/dependencies_test.go
@@ -861,6 +861,39 @@ func Test_getStatefulSetDependencies(t *testing.T) {
 			want:    testPairs[2].dependentObjectReference,
 			wantErr: false,
 		},
+		{
+			name: "statefulset with partial dependencies 4",
+			object: &unstructured.Unstructured{
+				Object: map[string]interface{}{
+					"apiVersion": "apps/v1",
+					"kind":       "StatefulSet",
+					"metadata": map[string]interface{}{
+						"name":      "fake-statefulset",
+						"namespace": namespace,
+					},
+					"spec": map[string]interface{}{
+						"serviceName": "fake-service",
+						"selector": map[string]interface{}{
+							"matchLabels": map[string]interface{}{
+								"app": "fake",
+							},
+						},
+						"template": map[string]interface{}{
+							"spec": testPairs[0].podSpecsWithDependencies.Object,
+						},
+						"volumeClaimTemplates": []interface{}{
+							map[string]interface{}{
+								"metadata": map[string]interface{}{
+									"name": "test-pvc",
+								},
+							},
+						},
+					},
+				},
+			},
+			want:    testPairs[0].dependentObjectReference[:3], // remove the pvc dependency because it was found in the volumeClaimTemplates
+			wantErr: false,
+		},
 	}
 
 	for i := range tests {


RainbowMango:
cc @XiShanYongYe-Chang for help

XiShanYongYe-Chang:
Thanks @jklaw90 
/assign

XiShanYongYe-Chang:
Hi @jklaw90 Do you know how to get the correct pvc name from podTemplate?

jklaw90:
> Hi @jklaw90 Do you know how to get the correct pvc name from podTemplate?

Hi @XiShanYongYe-Chang 
I don't think we can derive it from the podTemplate alone, we need the volumeClaimTemplates to determine the real name.  The pvc name would be `<template-claim-name>-<sts-name>-<index>`. 
That being said should the StatefulSet have dependencies or the pod in this case?

XiShanYongYe-Chang:
You're right, we cannot obtain the PVC name directly from the StatefulSet, as it also needs to be combined with the pod's index. Therefore, the current method of obtaining the dependent PVC name is incorrect.

Provide the Pod's YAML information here:
```yaml
apiVersion: v1
kind: Pod
metadata:
  generateName: web-
  labels:
    app: nginx
    apps.kubernetes.io/pod-index: "0"
    controller-revision-hash: web-59cfccf6c
    statefulset.kubernetes.io/pod-name: web-0
  name: web-0
  namespace: default
  ownerReferences:
  - apiVersion: apps/v1
    blockOwnerDeletion: true
    controller: true
    kind: StatefulSet
    name: web
    uid: 13e01c3d-4c7e-4d3a-8611-a0b41dd30be9
spec:
  containers:
  - image: registry.k8s.io/nginx-slim:0.24
    imagePullPolicy: IfNotPresent
    name: nginx
    ports:
    - containerPort: 80
      name: web
      protocol: TCP
    resources: {}
    terminationMessagePath: /dev/termination-log
    terminationMessagePolicy: File
    volumeMounts:
    - mountPath: /usr/share/nginx/html
      name: www
  dnsPolicy: ClusterFirst
  enableServiceLinks: true
  hostname: web-0
  preemptionPolicy: PreemptLowerPriority
  priority: 0
  restartPolicy: Always
  securityContext: {}
  serviceAccount: default
  serviceAccountName: default
  subdomain: nginx
  terminationGracePeriodSeconds: 10
  volumes:
  - name: www
    persistentVolumeClaim:
      claimName: www-web-0
```

cc @RainbowMango

XiShanYongYe-Chang:
Hi @jklaw90 how are you currently obtaining the dependent PVC, or is it also not being obtained?

jklaw90:
@XiShanYongYe-Chang 

It's not being obtained currently. I assumed [getPodDependencies](https://github.com/karmada-io/karmada/blob/106eb02b992b305ba7d3c8b47e4f343098efee37/pkg/resourceinterpreter/default/native/dependencies.go#L98C6-L98C24)  would handle this for pods?

XiShanYongYe-Chang:
The `getPodDependencies` function can serve to obtain the pod's dependent resources, but we will not be able to retrieve the associated PVC resources when distributing StatefulSets from the Karmada control plane.

jklaw90:
We could use replicas and StatefulSet name with the claim template to derive pvc names. 
But we have the possibility of missing some of the replicas were scaled down at all.

jklaw90:
if you remove the template then that PVC should be a dependency. 
so i think the current pr handles the cases you want.

XiShanYongYe-Chang:
Should we execute different logic depending on whether .spec.volumeClaimTemplates exists or not?

jklaw90:
> Should we execute different logic depending on whether .spec.volumeClaimTemplates exists or not?

I might not understand the question. but i think the current code handles this, the [first](https://github.com/karmada-io/karmada/pull/5568/files#diff-73e03e6e3b62d9df8f49eab79b28704fb103dabaeb2ab1daf2ef2ba93de77b0bR787) and [new](https://github.com/karmada-io/karmada/pull/5568/files#diff-73e03e6e3b62d9df8f49eab79b28704fb103dabaeb2ab1daf2ef2ba93de77b0bR865) test should cover both cases.

XiShanYongYe-Chang:
Oh @jklaw90 thanks, I get you, you are right.

pkg/resourceinterpreter/default/native/dependencies.go
```diff
@@ -132,7 +132,28 @@ func getStatefulSetDependencies(object *unstructured.Unstructured) ([]configv1al
...
-	return helper.GetDependenciesFromPodTemplate(podObj)
+	var validDeps []configv1alpha1.DependentObjectReference
+	deps, err := helper.GetDependenciesFromPodTemplate(podObj)
+	if err != nil {
+		return nil, err
+	}
+
+	// ignore the PersistentVolumeClaim dependency if it was created by the StatefulSet VolumeClaimTemplates
+	for _, dep := range deps {
+		if dep.Kind != "PersistentVolumeClaim" ||
```
XiShanYongYe-Chang, line 143: How about using the constant `PersistentVolumeClaimKind` in the `constants.go`.

pkg/resourceinterpreter/default/native/dependencies.go
```diff
@@ -132,7 +132,28 @@ func getStatefulSetDependencies(object *unstructured.Unstructured) ([]configv1al
...
 	}
 
-	return helper.GetDependenciesFromPodTemplate(podObj)
+	var validDeps []configv1alpha1.DependentObjectReference
+	deps, err := helper.GetDependenciesFromPodTemplate(podObj)
+	if err != nil {
+		return nil, err
+	}
+
+	// ignore the PersistentVolumeClaim dependency if it was created by the StatefulSet VolumeClaimTemplates
```
XiShanYongYe-Chang, line 141: Can  you add more comment to explain why VolumeClaimTemplates can be ignored when the sts template contains VolumeClaimTemplates.

XiShanYongYe-Chang (commented):
Thanks~
/lgtm
/cc @RainbowMango

RainbowMango (commented):
/assign

pkg/resourceinterpreter/default/native/dependencies.go
```diff
@@ -132,7 +132,29 @@ func getStatefulSetDependencies(object *unstructured.Unstructured) ([]configv1al
...
+		}
+		for _, volumeClaimTemplate := range statefulSetObj.Spec.VolumeClaimTemplates {
+			if dep.Name == volumeClaimTemplate.Name {
+				continue
+			}
+			validDeps = append(validDeps, dep)
+		}
+	}
+
+	return validDeps, nil
```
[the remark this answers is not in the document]
jklaw90: done, thank you

RainbowMango:
@jklaw90 Could you please [squash your commits](https://karmada.io/docs/next/contributor/github-workflow/#squash-commits)? Then things will be all set.

jklaw90:
@RainbowMango i've been having some issues with `chart-lint-test`
It's not reporting the status, but i've also had it fail on some of the installation steps. seems unrelated to this pr.

RainbowMango (approved):
/lgtm
/approve

XiShanYongYe-Chang:
Hi @jklaw90 @RainbowMango do you think we need to rebase this commit to the previous branch?

RainbowMango:
I guess you mean if we need to backport this PR to release branches.  I think we can do it although it's not a critical bug fix, as I don't see any risk with it.

PR #5569, opened 2024-09-19
feat: Implement group score calculation

ipsum-0320:
**What type of PR is this?**

/kind api-change

Optimize the scoring mechanism used for distribution constraints. Currently, the score of a region is determined by the highest score of the clusters within the region, which is not reasonable. We need a scoring method that reflects the overall situation of the region.

@whitewindmills 

```release-note
```


ipsum-0320:
@whitewindmills it fixed #3314, can you take a look?

ipsum-0320:
@whitewindmills i have solved the lint problem.

PR #5615, opened 2024-09-27
one cluster should have only one transport to reduce the number of TCP connections

### LINKED ISSUE #5574: Too many duplicated tcp connections from aggregated apiserver to one member cluster
**What happened**:
image
As the requests to access the aggregated API server increase, the number of connections will grow more, sometimes, the number of connection is 3000

a hostIP should have a tcp connection

Execute tosfedctl get pods requests concurrently, then monitor the number of connections.

I think the root case is,  

image

every reqeust will create a httptransport, so can't reuse a TCP connection.

**Environment**:
- Karmada version: 0.9 
- Others:

lcw2:
**What type of PR is this?**

It can solve the problem about too many tcp connection between aggregated apiserver and apiserver of  member's cluster
Fixes #5574

截屏2024-09-27 10 46 40

test.sh
````
KUBECONFIG="/root/.kube/karmada.config"
SLEEP_INTERVAL=0.1
MAX_JOBS=100
function run_karmadactl() {
  for ((i = 1; i <= 50; i++)); do
    karmadactl  --kubeconfig "$KUBECONFIG" get node --operation-scope=members
    sleep "$SLEEP_INTERVAL"
  done
}
for ((i = 1; i <= MAX_JOBS; i++)); do
  run_karmadactl &
done
wait
````

result.sh
````
#!/bin/bash

while true
do
  tcp_count=$(netstat -anp | grep 6443| wc -l)
  sleep 1
  echo "$(date '+%Y-%m-%d %H:%M:%S') - Current total TCP connections: $((tcp_count))"
done
````

fix before
企业微信截图_f9386612-1d3f-402c-8aee-243149a67406

fix after
企业微信截图_ace06070-69d8-4391-853f-ba3a488df8f2

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -27,6 +27,7 @@ require (
 	github.com/yuin/gopher-lua v0.0.0-20220504180219-658193537a64
 	go.uber.org/mock v0.4.0
 	golang.org/x/net v0.24.0
+	golang.org/x/sync v0.7.0
 	golang.org/x/term v0.19.0
 	golang.org/x/text v0.14.0
 	golang.org/x/time v0.5.0
@@ -174,7 +175,6 @@ require (
 	golang.org/x/exp v0.0.0-20231226003508-02704c960a9b // indirect
 	golang.org/x/mod v0.17.0 // indirect
 	golang.org/x/oauth2 v0.18.0 // indirect
-	golang.org/x/sync v0.7.0 // indirect
 	golang.org/x/sys v0.19.0 // indirect
 	google.golang.org/appengine v1.6.8 // indirect
 	google.golang.org/genproto v0.0.0-20231212172506-995d672761c0 // indirect
--- a/pkg/util/proxy/proxy.go
+++ b/pkg/util/proxy/proxy.go
@@ -18,16 +18,21 @@ package proxy
 
 import (
 	"context"
+	"crypto/md5" //nolint:gosec
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
 	"path"
+	"sort"
 	"strings"
+	"sync"
 	"time"
 
+	"golang.org/x/sync/singleflight"
 	authenticationv1 "k8s.io/api/authentication/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/httpstream"
@@ -43,6 +48,13 @@ import (
 
 // SecretGetterFunc is a function to get secret.
 type SecretGetterFunc func(context.Context, string, string) (*corev1.Secret, error)
+
+type clusterEndpointInfo struct {
+	Transport *http.Transport
+}
+
+var clusterEndpointInfoStore sync.Map
+var singleExecution singleflight.Group
 
 // ConnectCluster returns a handler for proxy cluster.
 func ConnectCluster(ctx context.Context, cluster *clusterapis.Cluster, proxyPath string, secretGetter SecretGetterFunc, responder registryrest.Responder) (http.Handler, error) {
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: pkg/util/proxy/proxy.go [hunk 1]
```diff
@@ -159,12 +168,32 @@ func Location(cluster *clusterapis.Cluster, tlsConfig *tls.Config) (*url.URL, ht
...
+			return value, nil
+		}
+		proxyTransport, err := createProxyTransport(cluster, tlsConfig)
+		if err != nil {
+			return nil, err
+		}
+		clusterEndpoint := clusterEndpointInfo{
+			Transport: proxyTransport,
+		}
+		clusterEndpointInfoStore.Store(cluster.UID, clusterEndpointInfo{
```
chaunceyjiang, line 187: Just a question: When will the `clusterEndpointInfo` be removed from the `clusterEndpointInfoStore`?

go.mod
```diff
@@ -174,7 +175,6 @@ require (
 	golang.org/x/exp v0.0.0-20231226003508-02704c960a9b // indirect
 	golang.org/x/mod v0.17.0 // indirect
 	golang.org/x/oauth2 v0.18.0 // indirect
-	golang.org/x/sync v0.7.0 // indirect
```
whitewindmills, line 177: it seems that the go.mod file has not changed substantially. you need to restore it.

pkg/util/proxy/proxy.go
whitewindmills: +1

lcw2:
two questions:
1) when  ProxyURL and ProxyHeader of cluster changed， how to handle transport ?  
store ProxyURL and ProxyHeader in  clusterEndpointInfo, when a request comes in, compare the current values with the stored ones. If those fields have changed, create a new transport ?  or other idea? 

2) when cluster is deleted， should the  transport of the cluster also be deleted？
If deletion is required,  can this be archieved by periodically list cluster?  30s?
any suggestion? @whitewindmills @chaunceyjiang @chaunceyjiang

whitewindmills:
I just feel like this approach seems too complicated and easy to go wrong. can we consider using all key fields(such as `APIEndpoint`, `ProxyURL` ...) to generate a hash value as the `clusterEndpointInfoStore` key? and we might do not need to remove it when cluster is deleted.

go.mod
lcw2: difference in indirect

XiShanYongYe-Chang:
/cc @whitewindmills @chaunceyjiang @yanfeng1992 
Hi, guys, can you help take a review again?

go.mod
whitewindmills: I don't think it's necessary.

pkg/util/proxy/proxy.go
```diff
@@ -238,3 +271,18 @@ func SkipGroup(group string) bool {
...
+	usedFields := make([]string, 0)
+	proxyHeaderKeys := make([]string, 0, len(cluster.Spec.ProxyHeader))
+	for key := range cluster.Spec.ProxyHeader {
+		proxyHeaderKeys = append(proxyHeaderKeys, key)
+	}
+	sort.Strings(proxyHeaderKeys)
+	for _, key := range proxyHeaderKeys {
+		usedFields = append(usedFields, key, cluster.Spec.ProxyHeader[key])
+	}
+	usedFields = append(usedFields, string(cluster.UID), cluster.Spec.ProxyURL, cluster.Spec.APIEndpoint)
```
whitewindmills, line 285: is UID used by `createProxyTransport`?

whitewindmills (commented):
have you tested it yet? I thought of a scenario: when the cluster goes offline, the connection will be automatically disconnected after a period of time. when the cluster comes back online, can it work fine?

pkg/util/proxy/proxy.go
lcw2: UID changed means the cluster rejoined,   should create a new transport.

pkg/util/proxy/proxy.go
```diff
@@ -159,12 +171,33 @@ func Location(cluster *clusterapis.Cluster, tlsConfig *tls.Config) (*url.URL, ht
...
+	if v, ok := clusterEndpointInfoStore.Load(clusterHash); ok {
+		clusterEndpointsInfo := v.(clusterEndpointInfo)
+		return location, clusterEndpointsInfo.Transport, nil
+	}
+
+	endpointInfo, err, _ := singleExecution.Do(clusterHash, func() (interface{}, error) {
+		if value, ok := clusterEndpointInfoStore.Load(clusterHash); ok {
+			return value, nil
+		}
+		proxyTransport, err := createProxyTransport(cluster, tlsConfig)
```
yanfeng1992, line 184: What happens if tlsConfig changes?

lcw2:
> have you tested it yet? I thought of a scenario: when the cluster goes offline, the connection will be automatically disconnected after a period of time. when the cluster comes back online, can it work fine?

I stop the apiserver  the minutes ,and  restart the apiserver, karmadactl get nodes work fine.

pkg/util/proxy/proxy.go
lcw2: It will not  work fine,  cannot request successfully. But I think no need to consider the scenario abourt secrets  changes.

whitewindmills:
I'm not sure if you built the cache before stoping apiserver. maybe we haven't considered it comprehensively enough. can we talk about it in the next community meeting?

lcw2:
> I'm not sure if you built the cache before stoping apiserver. maybe we haven't considered it comprehensively enough. can we talk about it in the next community meeting?

OK

RainbowMango:
As discussed at the [community meeting 2024-10-22](https://docs.google.com/document/d/1y6YLVC-v7cmVAdbjedoyR5WL0-q45DBRXTvz5_I7bkA/edit?pli=1&tab=t.0#heading=h.ahk2q34vbskt), we are going to disable keep alive as this is a simpler way to avoid connection exhaust issue.

We are still open to discussion on the approach that share transport, as long as it can handle the certificate rotation concerns.

/close

PR #5617, opened 2024-09-27
operator/pkg: mitigate null pointer dereference for `opt.Karmada.Spec.HostCluster` in `newRunData()` function

moawnallah:
**Description**

In this commit, we mitigate the null pointer dereference issue in the deinit operator pkg by adding defensive checks on 
`opt.Karmada.Spec.HostCluster` in `newRunData` func.

**Motivation and Context**

While unit testing the Karmada Init and DeInit in the operator package (#5613), deliberately passing empty components on those fields triggers a null pointer exception error. This PR resolves this issue by adding defensive checks. 

While testing 

**What type of PR is this?**

/kind cleanup

```release-note
NONE
```

### DIFF
--- a/operator/pkg/init.go
+++ b/operator/pkg/init.go
@@ -166,6 +166,11 @@ func newRunData(opt *InitOptions) (*initData, error) {
 		}
 	}
 
+	var hostClusterDNSDomain string
+	if opt.Karmada.Spec.HostCluster != nil && opt.Karmada.Spec.HostCluster.Networking != nil && opt.Karmada.Spec.HostCluster.Networking.DNSDomain != nil {
+		hostClusterDNSDomain = *opt.Karmada.Spec.HostCluster.Networking.DNSDomain
+	}
+
 	return &initData{
 		name:                    opt.Name,
 		namespace:               opt.Namespace,
@@ -178,7 +183,7 @@ func newRunData(opt *InitOptions) (*initData, error) {
 		privateRegistry:         privateRegistry,
 		components:              opt.Karmada.Spec.Components,
 		featureGates:            opt.Karmada.Spec.FeatureGates,
-		dnsDomain:               *opt.Karmada.Spec.HostCluster.Networking.DNSDomain,
+		dnsDomain:               hostClusterDNSDomain,
 		CertStore:               certs.NewCertStore(),
 	}, nil
 }
--- a/operator/pkg/controller/karmada/validating.go
+++ b/operator/pkg/controller/karmada/validating.go
@@ -18,6 +18,7 @@ package karmada
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/url"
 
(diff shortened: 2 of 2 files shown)


operator/pkg/init.go
```diff
@@ -65,6 +65,9 @@ func (opt *InitOptions) Validate() error {
 			return fmt.Errorf("unexpected invalid crds remote url %s", opt.CRDTarball.HTTPSource.URL)
 		}
 	}
+	if opt.Karmada == nil || opt.Karmada.Spec.Components == nil || opt.Karmada.Spec.Components.KarmadaAPIServer == nil {
+		return fmt.Errorf("invalid Karmada configuration: Karmada, Karmada components, and Karmada API server must be defined")
+	}
```
RainbowMango, line 70: > opt.Karmada.Spec.Components == nil

I'm curious why components are not allowed to be nil. From the API, it seems all fields are optional. 
Do you know some background? @chaosi-zju

mohamedawnallah: I'm eager to move this PR forward. Since I'm new to this part of the codebase, I want to make sure I've understood everything correctly. Any additional hints would be greatly appreciated. Thanks! 🙏 

cc @XiShanYongYe-Chang, @chaosi-zju

mohamedawnallah:
Hey @chaosi-zju,

Just checking in on this PR. Any feedback you might have would be greatly appreciated!

It looks like PR #5613 is waiting on this one. Thanks!

mohamedawnallah:
Hey @chaosi-zju,

Just checking in on this PR. Any feedback you might have would be greatly appreciated!

It looks like PR #5613 is waiting on this one. Thanks!

RainbowMango:
@mohamedawnallah Could you please rebase this PR? There is a conflict needs to be resolved.
Also, what's the benefit of this? Only for ease of the unit tests?

cc @zhzhuang-zju @jabellard

zhzhuang-zju:
> While unit testing the Karmada Init and DeInit in the operator package (https://github.com/karmada-io/karmada/pull/5613), deliberately passing empty components on those fields triggers a null pointer exception error. This PR resolves this issue by adding defensive checks.

https://github.com/karmada-io/karmada/blob/c31c8e4a58430bc8bdf4b28fa7232dffbc3fce49/operator/pkg/apis/operator/v1alpha1/register.go#L49-L54

We registered a method in the init function to set default values for Karmada,  which will be invoked when Default() is called. So in practice, there won’t be any null pointer issues because Default() is called before this code segment. When writing unit tests (UT), calling operatorscheme.Scheme.Default(karmada) can achieve the same effect.

Of course, adding some defensive checks can also prevent panics in case operatorscheme.Scheme.Default(karmada) is not called correctly in the code.

mohamedawnallah:
It looks like some of those defensive checks already added in that PR #5814. Updated this PR to handle when `karmada *operatorv1alpha1.Karmada` is `nil` and when `opt.Karmada.Spec.HostCluster` is `nil` part of `newRunData`.

PR #5621, opened 2024-09-28
Implement calc group score for divided to assign replicas evenly

### LINKED ISSUE #4805: Scheduler: support the ability to automatically assign replicas evenly
### Background
We want to introduce a new replica assignment strategy in the scheduler, which supports an even assignment of the target replicas across the currently selected clusters.

### Explanation
After going through the filtering, prioritization, and selection phases, three clusters(`member1`, `member2`, `member3`) were selected. We will automatically assign 9 replicas equally among these three clusters, the result we expect is `[{member1: 3}, {member2: 3}, {member3: 3}]`.

**Why is this needed**:
### User Story
As a developer, we have a deployment with 2 replicas that needs to be deployed with high availability across AZs. We hope Karmada can schedule it to two AZs and ensure that there is a replica on each AZ.
2AZ

Our PropagationPolicy might look like this:
```yaml
kind: PropagationPolicy
metadata:
  name: foo
  namespace: bar
spec:
  resourceSelectors:
    - apiVersion: apps/v1
      kind: Deployment
      name: foo
  placement:
    replicaScheduling:
      weightPreference:
    spreadConstraints:
    - spreadByField: zone
      maxGroups: 2
      minGroups: 2
    - spreadByField: cluster
      maxGroups: 2
      minGroups: 2
```
But unfortunately, the strategy `AvailableReplicas` does not guarantee that our replicas are evenly assigned.

### Any ideas?
We can introduce a new replica assignment strategy like `AvailableReplicas`, maybe we can name it `AverageReplicas`. It is essentially different from static weight assignment, because it does not support spread constraints and is mandatory. When assigning replicas, it does not consider whether the cluster can place so many replicas.

ipsum-0320:
**What type of PR is this?**
/kind feature

Optimize the score calculation method to better measure the pros and cons of the group. This PR implements calc group score for divided to assign replicas evenly.

Fixes #4805

@whitewindmills 

```release-note
```

### DIFF
--- a/pkg/scheduler/core/spreadconstraint/group_clusters.go
+++ b/pkg/scheduler/core/spreadconstraint/group_clusters.go
@@ -17,6 +17,8 @@ limitations under the License.
 package spreadconstraint
 
 import (
+	"math"
+
 	"k8s.io/utils/ptr"
 
 	clusterv1alpha1 "github.com/karmada-io/karmada/pkg/apis/cluster/v1alpha1"
@@ -40,7 +42,7 @@ type GroupClustersInfo struct {
 // ProviderInfo indicate the provider information
 type ProviderInfo struct {
 	Name              string
-	Score             int64 // the highest score in all clusters of the provider
+	Score             int64 // the comprehensive score in all clusters of the provider
 	AvailableReplicas int64
 
 	// Regions under this provider
@@ -54,7 +56,7 @@ type ProviderInfo struct {
 // RegionInfo indicate the region information
 type RegionInfo struct {
 	Name              string
-	Score             int64 // the highest score in all clusters of the region
+	Score             int64 // the comprehensive score in all clusters of the region
 	AvailableReplicas int64
 
 	// Zones under this provider
@@ -66,7 +68,7 @@ type RegionInfo struct {
 // ZoneInfo indicate the zone information
 type ZoneInfo struct {
 	Name              string
-	Score             int64 // the highest score in all clusters of the zone
+	Score             int64 // the comprehensive score in all clusters of the zone
 	AvailableReplicas int64
 
 	// Clusters under this zone, sorted by cluster.Score descending.
@@ -109,9 +111,9 @@ func groupClustersBasedTopology(
 	}
 	groupClustersInfo.calAvailableReplicasFunc = calAvailableReplicasFunc
 	groupClustersInfo.generateClustersInfo(clustersScore, rbSpec)
-	groupClustersInfo.generateZoneInfo(spreadConstraints)
-	groupClustersInfo.generateRegionInfo(spreadConstraints)
-	groupClustersInfo.generateProviderInfo(spreadConstraints)
+	groupClustersInfo.generateZoneInfo(spreadConstraints, rbSpec)
+	groupClustersInfo.generateRegionInfo(spreadConstraints, rbSpec)
+	groupClustersInfo.generateProviderInfo(spreadConstraints, rbSpec)
 
 	return groupClustersInfo
 }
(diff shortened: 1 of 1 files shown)


pkg/scheduler/core/spreadconstraint/group_clusters.go
```diff
@@ -128,6 +128,46 @@ func groupClustersIgnoringTopology(
 	return groupClustersInfo
 }
 
+const weightUnit int64 = 1000
```
whitewindmills, line 131: to explain this `weightUnit`, pls add some comments here.

pkg/scheduler/core/spreadconstraint/group_clusters.go
```diff
@@ -128,6 +128,46 @@ func groupClustersIgnoringTopology(
...
+
+func (info *GroupClustersInfo) calcGroupScore(
+	clusters []ClusterDetailInfo,
+	rbSpec *workv1alpha2.ResourceBindingSpec,
+	minGroups int) int64 {
+	// if the replica scheduling type is divided, the score is calculated by followed.
+	int32MinGroups := int32(minGroups)
+	restReplica := rbSpec.Replicas % int32MinGroups
```
whitewindmills, line 139: unnecessary to calculate `restReplica`

pkg/scheduler/core/spreadconstraint/group_clusters.go
```diff
@@ -128,6 +128,46 @@ func groupClustersIgnoringTopology(
...
+		if sumAvailableReplica >= targetReplica {
+			break
+		}
+	}
+
+	// cluster.Score is 0 or 100. To minimize the impact of Score,
+	// set the atomic value of targetReplica to 1000. This way,
+	// when sorting by Group Score, targetReplica will be considered first,
+	// and if the Weights are the same, then Score will be considered.
+
```
whitewindmills, line 161: remove this blank line.

pkg/scheduler/core/spreadconstraint/group_clusters.go
```diff
@@ -128,6 +128,46 @@ func groupClustersIgnoringTopology(
...
+	// clusters have been sorted by cluster.Score descending,
+	// and if the cluster.Score is the same, the cluster.availableReplica is ascending.
+	var sumAvailableReplica int64
+	var sumScore int64
+	var validClusters int64
+	for _, cluster := range clusters {
+		sumAvailableReplica += cluster.AvailableReplicas
+		sumScore += cluster.Score
+		validClusters++
+		if sumAvailableReplica >= targetReplica {
```
whitewindmills, line 152: a corner case here: 
targetReplica is `10`
group is `[{"availableReplicas": 10000, "score": 100}, {"availableReplicas": 10000, "score": 100}, {"availableReplicas": 10000, "score": 100}]`
sumAvailableReplica only adds the `AvailableReplicas` of a cluster, that is, this will cause the algorithm to degenerate into a situation where only one cluster score is calculated.
### Improvement measures
always take `minGroups` number of clusters.

pkg/scheduler/core/spreadconstraint/group_clusters.go
```diff
@@ -178,13 +218,26 @@ func (info *GroupClustersInfo) generateZoneInfo(spreadConstraints []policyv1alph
 		}
 	}
 
+	isGroupByZone := false
```
whitewindmills, line 221: remove it, `isGroupByZone` is always `true`.

pkg/scheduler/core/spreadconstraint/group_clusters.go
```diff
@@ -212,13 +265,26 @@ func (info *GroupClustersInfo) generateRegionInfo(spreadConstraints []policyv1al
 		info.Regions[region] = regionInfo
 	}
 
+	isGroupByRegion := false
```
whitewindmills, line 268: ditto

pkg/scheduler/core/spreadconstraint/group_clusters.go
ipsum-0320: ok, i'll do it.

pkg/scheduler/core/spreadconstraint/group_clusters.go
ipsum-0320: ok, now i use `math.ceil()`.

pkg/scheduler/core/spreadconstraint/group_clusters.go
ipsum-0320: ok

ipsum-0320:
I have fixed all the problem. @whitewindmills

whitewindmills (commented):
thanks for your hard work, let @RainbowMango make the final decision.
/approve
/assgin @RainbowMango

RainbowMango (commented):
/assign

RainbowMango:
@ipsum-0320 Are you going to squash the commits?
https://karmada.io/docs/contributor/github-workflow/#squash-commits

ipsum-0320:
> @ipsum-0320 Are you going to squash the commits?你要压缩这些提交吗？ https://karmada.io/docs/contributor/github-workflow/#squash-commits

@RainbowMango yeah, after your review, I will compress these commits.

ipsum-0320:
> @ipsum-0320 Are you going to squash the commits? https://karmada.io/docs/contributor/github-workflow/#squash-commits

now, i have squashed the commits. @RainbowMango

pkg/scheduler/core/spreadconstraint/group_clusters.go
```diff
@@ -128,6 +130,62 @@ func groupClustersIgnoringTopology(
...
+// Thus, when sorting by Group Score, targetReplica will be considered first, and if the Weights are the same, then Score will be considered.
+const weightUnit int64 = 1000
+
+func (info *GroupClustersInfo) calcGroupScore(
+	clusters []ClusterDetailInfo,
+	rbSpec *workv1alpha2.ResourceBindingSpec,
+	minGroups int) int64 {
+	// if the replica scheduling type is divided, the score is calculated by followed.
+	float64MinGroups := float64(minGroups)
+	targetReplica := int64(math.Ceil(float64(rbSpec.Replicas) / float64MinGroups))
```
RainbowMango, line 144: I understand that here estimates the number of replicas possibly assigned to the current group. 

Why do you think it is reasonable to evenly distribute the total number of replicas among each group? Will this have a significant impact on the subsequent algorithm?

pkg/scheduler/core/spreadconstraint/group_clusters.go
RainbowMango: This const is used in the `calcGroupScore` method, why not put it in as a variable?

pkg/scheduler/core/spreadconstraint/group_clusters.go
```diff
@@ -128,6 +130,62 @@ func groupClustersIgnoringTopology(
...
+	// set the atomic value of targetReplica to 1000. This way,
+	// when sorting by Group Score, targetReplica will be considered first,
+	// and if the Weights are the same, then Score will be considered.
+	if sumAvailableReplica < targetReplica {
+		sumAvailableReplica = sumAvailableReplica * weightUnit
+		return sumAvailableReplica + sumScore/int64(len(clusters))
+	}
+
+	targetReplica = targetReplica * weightUnit
+	return targetReplica + sumScore/validClusters
```
RainbowMango, line 186: This piece of code is difficult to understand. Can we add some examples?  like https://github.com/karmada-io/karmada/blob/adef1e59748e1e1d31cb75fffe406b5dd69a66d7/pkg/scheduler/core/spreadconstraint/select_clusters_by_cluster.go#L57-L65

RainbowMango (commented):
Generally looks good, just some questions and suggestions.

pkg/scheduler/core/spreadconstraint/group_clusters.go
```diff
@@ -252,8 +328,16 @@ func (info *GroupClustersInfo) generateProviderInfo(spreadConstraints []policyv1
 		info.Providers[provider] = providerInfo
 	}
 
+	var minGroups int
+	if rbSpec != nil && rbSpec.Placement != nil && rbSpec.Placement.SpreadConstraints != nil {
+		for _, sc := range rbSpec.Placement.SpreadConstraints {
```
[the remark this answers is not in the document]
ipsum-0320: Yes, I took a closer look at the source of the parameters and the call stack, and I found that rbSpec.Placement.SpreadConstraints is indeed spreadConstraints. I’ll streamline the parameters. Thank you for your thorough review. There is the code: https://github.com/karmada-io/karmada/blob/master/pkg/scheduler/core/generic_scheduler.go#L97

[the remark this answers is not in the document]
ipsum-0320: @RainbowMango  Yes, I took a closer look at the source of the parameters and the call stack, and I found that rbSpec.Placement.SpreadConstraints is indeed spreadConstraints. I’ll streamline the parameters. Thank you for your thorough review. There is the code: https://github.com/karmada-io/karmada/blob/master/pkg/scheduler/core/generic_scheduler.go#L97

pkg/scheduler/core/spreadconstraint/group_clusters.go
ipsum-0320: @RainbowMango  This approach is based on statistical principles. Due to the MinGroups constraint of each Group, the final allocation outcome will inevitably assign the replicas across MinGroups groups. For example, if we need to allocate 100 replicas across 5 groups, I’ve simplified this model by abstracting it as a random allocation (though there might be stricter allocation rules, which would require detailed statistical data analysis). Consequently, this allocation will follow a normal distribution, meaning that each Group will receive a number of replicas around the average. In this case, each Group will likely receive around 20 replicas (even if it’s not exactly 20, the probability is highest for receiving around that number). Therefore, I believe it is reasonable to distribute the replicas evenly across the groups, as it aligns with statistical principles.

This allocation method only impacts the scoring calculation of the algorithm itself. The score affects the quality of subsequent selection algorithm results but does not lead to cases where there are sufficient replicas available but scheduling fails. This aspect is ensured by the specific selection algorithm, which guarantees successful scheduling. Once scheduling is successful, the selection of Groups is based on their Group Score.

pkg/scheduler/core/spreadconstraint/group_clusters.go
ipsum-0320: Great, I’ll add the relevant comments. @RainbowMango

ipsum-0320:
I have resolved the issues you raised; please review it once again. @RainbowMango

pkg/scheduler/core/spreadconstraint/group_clusters.go
ipsum-0320: Since the `weightUnit` variable represents a unit, I have defined it as a constant. Additionally, because `weightUnit` has a significant impact on the algorithm, I defined it outside the function to emphasize its importance. This also makes it easier to adjust in the future. @RainbowMango

RainbowMango: OK, then. no big deal.

RainbowMango (commented):
/lgtm

PR #5630, opened 2024-09-30
operator/pkg: add test helpers

mohamedawnallah:
**Description**

Add test helpers that would be reused across tasks in the operator package.

**What type of PR is this?**

/kind cleanup

```release-note
NONE
```

### DIFF
--- a/operator/pkg/tasks/init/test_helpers.go
+++ b/operator/pkg/tasks/init/test_helpers.go
@@ -1,0 +1,157 @@
+/*
+Copyright 2024 The Karmada Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tasks
+
+import (
+	"fmt"
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+
+	operatorv1alpha1 "github.com/karmada-io/karmada/operator/pkg/apis/operator/v1alpha1"
+	"github.com/karmada-io/karmada/operator/pkg/certs"
+)
+
+// TestInterface defines the interface for retrieving test data.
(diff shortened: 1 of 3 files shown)


operator/pkg/tasks/init/test_helpers.go
```diff
@@ -0,0 +1,210 @@
...
+//
+// Parameters:
+// - client: Kubernetes client interface.
+// - namespace: Namespace for pod creation.
+// - componentName: Base name for pods and containers.
+// - replicaCount: Number of pods to create.
+// - labels: Labels for pods creation.
+//
+// Returns an error if pod creation fails.
+func createPods(client clientset.Interface, namespace string, componentName string, replicaCount int32, labels map[string]string) error {
```
XiShanYongYe-Chang, line 172: func `createPods` is unused (unused)

mohamedawnallah: @XiShanYongYe-Chang It is utilized in the PRs that depend on this PR.

mohamedawnallah: More specifically, those are the PRs that utilize `createPods` later: #5632 #5634

XiShanYongYe-Chang: I seem to have found a loop between these dependencies, so can you help point out which pr to look at first?

XiShanYongYe-Chang: > Could we ignore this linting issue in this PR?

The lint error of the current pr cannot be ignored. Therefore, if the CI fails, the pr cannot be integrated.

> More specifically, those are the PRs that utilize createPods later: https://github.com/karmada-io/karmada/pull/5632 https://github.com/karmada-io/karmada/pull/5634

Perhaps you can add this method later when you use the createPods method.

mohamedawnallah: > Perhaps you can add this method later when you use the createPods method.

Sounds good. I will write the createPods twice in those PRs and we could handle them later by rebasing. 👍

mohamedawnallah: @XiShanYongYe-Chang Or What do you think adding this lint directive `// nolint:unused` above createPods and this could be removed later in the dependencies instead of repeating `createPods` twice function ?

XiShanYongYe-Chang: I tend to wait for the function to be used before filling in, as to whether to write the same function twice, can be refactored to reduce code redundancy.

XiShanYongYe-Chang: Let's modify this pr first so that subsequent prs can not be blocked.

XiShanYongYe-Chang (commented):
Thanks
/lgtm
/approve

PR #5637, opened 2024-10-02
Added tests for scheduler/cache

anujagrawal699:
**Description:**
This PR enhances the unit tests for the Scheduler Cache in the cache package. The improvements aim to increase code reliability, maintainability, and test coverage for the Scheduler Cache.

**Additions:**
1. pkg/scheduler/cache/cache_test.go

**Test Coverage:**
pkg/cache : 7.50% to 100%

**What type of PR is this?**
/kind feature

Fixes a part of #5470 

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/cache/cache_test.go
+++ b/pkg/scheduler/cache/cache_test.go
@@ -1,0 +1,256 @@
+/*
+Copyright 2024 The Karmada Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cache
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	clusterv1alpha1 "github.com/karmada-io/karmada/pkg/apis/cluster/v1alpha1"
+	clusterlister "github.com/karmada-io/karmada/pkg/generated/listers/cluster/v1alpha1"
+)
+
+func TestNewCache(t *testing.T) {
+	tests := []struct {
+		name          string
+		clusterLister clusterlister.ClusterLister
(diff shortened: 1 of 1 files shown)


pkg/scheduler/cache/cache_test.go
```diff
@@ -0,0 +1,268 @@
...
+
+	clusterv1alpha1 "github.com/karmada-io/karmada/pkg/apis/cluster/v1alpha1"
+	clusterlister "github.com/karmada-io/karmada/pkg/generated/listers/cluster/v1alpha1"
+)
+
+func TestNewCache(t *testing.T) {
+	tests := []struct {
+		name          string
+		clusterLister clusterlister.ClusterLister
+		wantNil       bool
```
XiShanYongYe-Chang, line 36: Is this parameter necessary? Is there a time when true?

pkg/scheduler/cache/cache_test.go
```diff
@@ -0,0 +1,268 @@
...
+			}
+
+			// Test GetCluster for non-existent cluster
+			assert.Nil(t, snapshot.GetCluster("non-existent-cluster"), "GetCluster(non-existent-cluster) should return nil")
+
+			// Verify that the snapshot is a deep copy
+			for i, cluster := range tt.clusters {
+				snapshotCluster := snapshot.GetClusters()[i].Cluster()
+				assert.Equal(t, cluster, snapshotCluster, "Snapshot cluster should be equal to original")
+				assert.NotSame(t, cluster, snapshotCluster, "Snapshot cluster should be a deep copy")
```
XiShanYongYe-Chang, line 172: What's the difference between these two judgments?

pkg/scheduler/cache/cache_test.go
anujagrawal699: I think it makes test more maintainable as if in future a change is made in NewCache where it returns nil. However, if you say i could remove it.

pkg/scheduler/cache/cache_test.go
anujagrawal699: The first one for checking that the content of the snapshot cluster is identical to the original cluster and the second one for ensuring that the snapshot is indeed a deep copy and not just a reference to the original object.

pkg/scheduler/cache/cache_test.go
XiShanYongYe-Chang: NewCache returns a nil, which is meaningless.

pkg/scheduler/cache/cache_test.go
XiShanYongYe-Chang: Yes, I think so.

pkg/scheduler/cache/cache_test.go
anujagrawal699: Okay, i'll remove it.

XiShanYongYe-Chang (commented):
Thanks~
/lgtm
/approve

PR #5660, opened 2024-10-10
refactor ensure work

XiShanYongYe-Chang:
**What type of PR is this?**

/kind cleanup

The `ensureWork` function has some load, and its complexity will continue to increase during ongoing iterations. Therefore, some refactoring has been done on this function.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/binding/common.go
+++ b/pkg/controllers/binding/common.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"strconv"
 
-	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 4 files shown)


RainbowMango:
OK. will do.

pkg/controllers/binding/common.go
```diff
@@ -37,112 +36,98 @@ import (
...
-	var conflictResolutionInBinding policyv1alpha1.ConflictResolution
-	var suspension *policyv1alpha1.Suspension
-	switch scope {
-	case apiextensionsv1.NamespaceScoped:
+// workSyncer is design for syncing work objects as expected.
+type workSyncer struct {
+	client      client.Client
+	interpreter resourceinterpreter.ResourceInterpreter
+	overrider   overridemanager.OverrideManager
+}
```
RainbowMango, line 44: The `workSyncer` holds some immutable parameters, mainly used to reduce the number of parameters passed to ensureWork(). Is this the idea?

pkg/controllers/binding/common.go
```diff
@@ -37,112 +36,98 @@ import (
...
-			if resourceInterpreter.HookEnabled(clonedWorkload.GroupVersionKind(), configv1alpha1.InterpreterOperationReviseReplica) {
-				clonedWorkload, err = resourceInterpreter.ReviseReplica(clonedWorkload, int64(targetCluster.Replicas))
-				if err != nil {
-					klog.Errorf("Failed to revise replica for %s/%s/%s in cluster %s, err is: %v",
-						workload.GetKind(), workload.GetNamespace(), workload.GetName(), targetCluster.Name, err)
-					return err
-				}
-			}
+// ensureWork ensure Work to be created or updated in the target cluster.
+func (s *workSyncer) ensureWork(ctx context.Context, workload *unstructured.Unstructured, binding metav1.Object, targetCluster workv1alpha2.TargetCluster, config extraConfig) error {
```
RainbowMango, line 82: The `ensureWork` is dedicated to syncing a single work, which is split from the original `ensureWork`.

RainbowMango (commented):
I tried to understand the ideal, please correct me if not the case.
The benefits include:
- Reduced the number of parameters passed in `ensureWork`.
- Split the `ensureWork` into two functions.

pkg/controllers/binding/common.go
XiShanYongYe-Chang: It's like this.

pkg/controllers/binding/common.go
XiShanYongYe-Chang: Yes, more dependencies on single work generation will be extended in extraConfig in the future.

pkg/controllers/binding/common.go
```diff
@@ -37,112 +36,98 @@ import (
...
-	var targetClusters []workv1alpha2.TargetCluster
-	var placement *policyv1alpha1.Placement
-	var requiredByBindingSnapshot []workv1alpha2.BindingSnapshot
-	var replicas int32
-	var conflictResolutionInBinding policyv1alpha1.ConflictResolution
-	var suspension *policyv1alpha1.Suspension
-	switch scope {
-	case apiextensionsv1.NamespaceScoped:
+// workSyncer is design for syncing work objects as expected.
+type workSyncer struct {
```
chaunceyjiang, line 40: I believe it would be more appropriate to have this `workSyncer` placed in a separate new Go file.

XiShanYongYe-Chang: I plan to change the `common.go` file name to `worksyncer.go` when the review is almost complete. This makes it easier to see which parts have been modified during review.

PR #5662, opened 2024-10-10
hasResource should return whether resource is matched not cached

yanfeng1992:
**What type of PR is this?**

/kind bug

For detailed information of in tree plugins' execution order, please see:
// https://github.com/karmada-io/karmada/tree/master/docs/proposals/resource-aggregation-proxy#request-routing

image

But in fact, I found that if the requested resource is defined in ResourceRegistry but there is no cluster, the query is the resource of the karmada apiserver

image

```
apiVersion: search.karmada.io/v1alpha1
kind: ResourceRegistry
metadata:
  name: test
spec:
  resourceSelectors:
  - apiVersion: v1
    kind: Node
  - apiVersion: v1
    kind: Pod
  - apiVersion: v1
    kind: Namespace
```

This will cause two problems

case1.When there is no cluster registered, the query resource comes from karmada-apiserver

case2.When using watch, after deleting all clusters, watch will fail

image

Fixes #

```release-note
`karmada-search`: Modify the logic of checking whether the resource is registered when selecting the plugin .
```

### DIFF
--- a/pkg/search/proxy/store/multi_cluster_cache.go
+++ b/pkg/search/proxy/store/multi_cluster_cache.go
@@ -40,7 +40,7 @@ import (
 
 // Store is the cache for resources from multiple member clusters
 type Store interface {
-	UpdateCache(resourcesByCluster map[string]map[schema.GroupVersionResource]*MultiNamespace) error
+	UpdateCache(resourcesByCluster map[string]map[schema.GroupVersionResource]*MultiNamespace, registeredResources map[schema.GroupVersionResource]struct{}) error
 	HasResource(resource schema.GroupVersionResource) bool
 	GetResourceFromCache(ctx context.Context, gvr schema.GroupVersionResource, namespace, name string) (runtime.Object, string, error)
 	Stop()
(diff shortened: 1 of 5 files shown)


pkg/search/proxy/store/multi_cluster_cache.go
```diff
@@ -52,10 +52,10 @@ type Store interface {
 
 // MultiClusterCache caches resource from multi member clusters
 type MultiClusterCache struct {
-	lock            sync.RWMutex
-	cache           map[string]*clusterCache
-	cachedResources map[schema.GroupVersionResource]struct{}
-	restMapper      meta.RESTMapper
+	lock             sync.RWMutex
+	cache            map[string]*clusterCache
+	matchedResources map[schema.GroupVersionResource]struct{}
```
XiShanYongYe-Chang, line 57: How about rename `matched` to `registried`?

pkg/search/proxy/store/multi_cluster_cache.go
```diff
@@ -137,11 +120,11 @@ func (c *MultiClusterCache) Stop() {
 	}
 }
 
-// HasResource return whether resource is cached.
+// HasResource return whether resource is matched.
```
XiShanYongYe-Chang, line 123: Would it be better to use `registried` for the full text?

pkg/search/proxy/store/multi_cluster_cache.go
yanfeng1992: OK，registered

XiShanYongYe-Chang (commented):
Thanks~
/lgtm
/assign @ikaven1024

XiShanYongYe-Chang:
Hi @yanfeng1992, can you help add the release note? Maybe we need to cherry-pick this patch to the previous branch.

XiShanYongYe-Chang:
Let's merge it first.
/approve

Issue #5577, opened 2024-09-21
[Umbrella] Defining resource deleteion policy to support migration rollback


XiShanYongYe-Chang:
**What would you like to be added**:

Provides the ability to roll back migration safely. 

**Why is this needed**:

For specific details, please refer to #5101.

The specific work items can be planned as follows:

- [x] New proposal @CharlesQQ @XiShanYongYe-Chang #5101
- [x] Proposing API to support migration rollback @RainbowMango #5575 
- [x] Function Implementation @a7i #5597 
- [x] Refactor: optimize the processing of labels and annotations during resource depletion @a7i #5747
- [x] Rollback the migration of dependency distributed resources @XiShanYongYe-Chang #5717 
- [x] karmadactl promote can help rollout safely @wulemao #5601
- [x] Test Coverage 
  - [x] UT @a7i #5597 
  - [x] E2E @wulemao #5609
- [x] Document Description @wulemao https://github.com/karmada-io/website/pull/698 https://github.com/karmada-io/website/pull/697

XiShanYongYe-Chang:
Hi @a7i would you like to contribute this feature?
cc @a7i

a7i:
> Hi @a7i would you like to contribute this feature? cc @a7i

Yes absolutely! 🎉

wulemao:
Hi, everybody, I am working on #5380 tasks, which is highly relevant to this issue.

I am interested in this issue too, could you please assign some related test job to me?

XiShanYongYe-Chang:
Hi @a7i, how about handing over E2E tasks to @wulemao?

RainbowMango:
I added one more task, which we should let `karmadactl promote` leverage this feature to guarantee the promotion can be rolled back safely.  To address the issue #3004 sent by @chaunceyjiang .

wulemao:
> I added one more task, which we should let `karmadactl promote` leverage this feature to guarantee the promotion can be rolled back safely. To address the issue #3004 sent by @chaunceyjiang .

It is also a very interesting optimization. could I assign for it?

XiShanYongYe-Chang:
Thanks all~
/close

PR #5717, opened 2024-10-22
Keep preserveResourcesOnDeletion of the dependent resource consistent with that of the primary resource

XiShanYongYe-Chang:
**What type of PR is this?**

/kind feature

When users enable dependency distribution for a resource template, it will inherit the `PreserveResourcesOnDeletion` policy from the resource they are following. As a result, we can rollback the migration of dependency distributed resources.

Part of #5577 

```release-note
`karmada-controller-manager`: keep preserveResourcesOnDeletion of the dependent resource consistent with that of the primary resource
```

### DIFF
--- a/pkg/dependenciesdistributor/dependencies_distributor.go
+++ b/pkg/dependenciesdistributor/dependencies_distributor.go
@@ -546,6 +546,7 @@ func (d *DependenciesDistributor) removeScheduleResultFromAttachedBindings(bindi
 		delete(attachedBindings[index].Labels, bindingLabelKey)
 		updatedSnapshot := deleteBindingFromSnapshot(bindingNamespace, bindingName, attachedBindings[index].Spec.RequiredBy)
 		attachedBindings[index].Spec.RequiredBy = updatedSnapshot
+		attachedBindings[index].Spec.PreserveResourcesOnDeletion = nil
 		if err := d.Client.Update(context.TODO(), attachedBindings[index]); err != nil {
 			klog.Errorf("Failed to update binding(%s/%s): %v", binding.Namespace, binding.Name, err)
 			errs = append(errs, err)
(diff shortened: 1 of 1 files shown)


pkg/dependenciesdistributor/dependencies_distributor.go
```diff
@@ -545,7 +551,14 @@ func (d *DependenciesDistributor) removeScheduleResultFromAttachedBindings(bindi
 	for index, binding := range attachedBindings {
 		delete(attachedBindings[index].Labels, bindingLabelKey)
 		updatedSnapshot := deleteBindingFromSnapshot(bindingNamespace, bindingName, attachedBindings[index].Spec.RequiredBy)
+		preserveResourcesOnDeletion, err := obtainPreserveStrategyFromBinding(ctx, d.Client, binding)
```
RainbowMango, line 554: The attached bindings should have the `PreserveResourcesOnDeletion` configuration. Why do we have to fetch it from PP?

XiShanYongYe-Chang: It may have been overridden by `PreserveResourcesOnDeletion` from dependent resources. Therefore, it needs to be obtained from the PP of the current resource.

pkg/dependenciesdistributor/dependencies_distributor.go
```diff
@@ -546,6 +546,7 @@ func (d *DependenciesDistributor) removeScheduleResultFromAttachedBindings(bindi
 		delete(attachedBindings[index].Labels, bindingLabelKey)
 		updatedSnapshot := deleteBindingFromSnapshot(bindingNamespace, bindingName, attachedBindings[index].Spec.RequiredBy)
 		attachedBindings[index].Spec.RequiredBy = updatedSnapshot
+		attachedBindings[index].Spec.PreserveResourcesOnDeletion = nil
```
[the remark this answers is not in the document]
XiShanYongYe-Chang: A resource will be handled by only one distribution policy, either directly or through indirect follow-up distribution.

So, in your example, the system will deny secretA to be dependent on deployA.

RainbowMango (approved):
/approve

PR #5731, opened 2024-10-23
set conflictResolution attribute for resources that follow dependency distribution

chaosi-zju:
**What type of PR is this?**

/kind feature

When users enable dependency distribution for a resource template, it will inherit the `ConflictResolution` attribute from the resource they are following. As a result, we can migrate resources that follow dependency distribution.

Fixes 

```release-note
`karmada-controller-manager`: set conflictResolution attribute for resources that follow dependency distribution
```


RainbowMango:
> Fixes https://github.com/karmada-io/karmada/issues/5577

??

RainbowMango:
Duplicate of #4418.

RainbowMango:
Duplicate of #4418

Issue #5672, opened 2024-10-11
Watch Request Blocked When Member Cluster Offline


xigang:
**What happened**:

When the member cluster goes offline, there is a scenario where the client's Watch request gets blocked and does not receive pod events.

**What you expected to happen**:

Should we set a timeout: Set a reasonable timeout for cache.Watch() calls using context.WithTimeout or context.WithDeadline to control the operation time?

https://github.com/karmada-io/karmada/blob/master/pkg/search/proxy/store/multi_cluster_cache.go#L354

example code:
```
        // 使用带超时的上下文
        watchCtx, cancel := context.WithTimeout(ctx, watchTimeout)
        defer cancel()

        w, err := cache.Watch(watchCtx, options)
        if err != nil {
            if err == context.DeadlineExceeded {
                klog.Warningf("Watch timeout for cluster %s: %v", cluster, err)
                continue // 跳过这个集群，继续处理下一个
            }
            mux.Stop()
            return nil, fmt.Errorf("failed to watch cluster %s: %v", cluster, err)
        }
```

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Karmada version:
- kubectl-karmada or karmadactl version (the result of `kubectl-karmada version` or `karmadactl version`):
- Others:

XiShanYongYe-Chang:
> When the member cluster goes offline, there is a scenario where the client's Watch request gets blocked and does not receive pod events.

Are events in other normal clusters affected?

xigang:
@XiShanYongYe-Chang Unable to receive events from all member clusters through the aggregated apiserver, suspecting that the watch is blocked in `cache.watch()`.

XiShanYongYe-Chang:
Thank you for your reply.

According to your method, the watch connection will be disconnected after a certain period of time, and then the client needs to initiate a watch request again. Do I understand it correctly?

xigang:
> Thank you for your reply.
> 
> According to your method, the watch connection will be disconnected after a certain period of time, and then the client needs to initiate a watch request again. Do I understand it correctly?

Yes, but my scenario is quite special. The member cluster has gone offline, but since it hasn't been removed from the  `ResourceRegistry`, the rewatch requests still target the offline cluster, causing the watch requests to get stuck.

XiShanYongYe-Chang:
Hi @xigang, why close this issue?

xigang:
> Hi @xigang, why close this issue?

[...]

PR #5732, opened 2024-10-23
Add timeout for watch requests to member clusters to prevent request …

xigang:
**What type of PR is this?**
/kind bug

When the federate-apiserver's watch request to the member cluster gets stuck, it will cause the watch request from the federated client to get stuck as well.

Fixes # https://github.com/karmada-io/karmada/issues/5672

```release-note
NONE
```

### DIFF
--- a/pkg/search/proxy/store/multi_cluster_cache.go
+++ b/pkg/search/proxy/store/multi_cluster_cache.go
@@ -334,15 +334,34 @@ func (c *MultiClusterCache) Watch(ctx context.Context, gvr schema.GroupVersionRe
 		if cache == nil {
 			continue
 		}
-		w, err := cache.Watch(ctx, options)
-		if err != nil {
-			return nil, err
-		}
 
-		mux.AddSource(w, func(e watch.Event) {
-			setObjectResourceVersionFunc(cluster, e.Object)
-			addCacheSourceAnnotation(e.Object, cluster)
-		})
+		// The following logic adds a 30-second timeout to prevent watch requests to member clusters from hanging,
+		// which could cause the client watch to hang.
+		watchChan := make(chan watch.Interface, 1)
+		errChan := make(chan error, 1)
+
+		go func() {
+			w, err := cache.Watch(ctx, options)
+			if err != nil {
+				errChan <- fmt.Errorf("failed to start watch for resource %v in cluster %q: %v", gvr.String(), cluster, err)
+				return
+			}
+			watchChan <- w
+		}()
+
+		select {
+		case w := <-watchChan:
+			mux.AddSource(w, func(e watch.Event) {
+				setObjectResourceVersionFunc(cluster, e.Object)
+				addCacheSourceAnnotation(e.Object, cluster)
+			})
+		case err := <-errChan:
+			// If the watch request fails, return the error, and the client will retry.
+			return nil, err
+		case <-time.After(30 * time.Second):
+			// If the watch request times out, return an error, and the client will retry.
+			return nil, fmt.Errorf("timeout waiting for watch for resource %v in cluster %q", gvr.String(), cluster)
+		}
 	}
 	mux.Start()
 	return mux, nil


xigang:
cc @XiShanYongYe-Chang @RainbowMango @ikaven1024 PTAL.

pkg/search/proxy/store/multi_cluster_cache.go
```diff
@@ -334,15 +334,34 @@ func (c *MultiClusterCache) Watch(ctx context.Context, gvr schema.GroupVersionRe
...
+			mux.AddSource(w, func(e watch.Event) {
+				setObjectResourceVersionFunc(cluster, e.Object)
+				addCacheSourceAnnotation(e.Object, cluster)
+			})
+		case err := <-errChan:
+			// If the watch request fails, return the error, and the client will retry.
 			return nil, err
+		case <-time.After(30 * time.Second):
+			// If the watch request times out, return an error, and the client will retry.
+			return nil, fmt.Errorf("timeout waiting for watch for resource %v in cluster %q", gvr.String(), cluster)
```
zhzhuang-zju, line 363: @xigang Hi, if a watch request is hanging and causes a timeout, will the hanging watch request continue to exist in the subprocess?

xigang: @zhzhuang-zju Yes, there is this issue. When a watch request times out, the goroutine needs to be terminated.

RainbowMango: Good point! Then that case we have to cancel the context passed to cache.Watch().

RainbowMango: Another question:
Before starting the `Watch`, we tried to get the cache of that cluster, I'm curious why this cache still exists even after the cluster is gone. Do we have a chance to clean the cache?

https://github.com/karmada-io/karmada/blob/e7b6513db0e44090a9d11fcff15b1fec097b3b2a/pkg/search/proxy/store/multi_cluster_cache.go#L333-L336

xigang: > Another question: Before starting the `Watch`, we tried to get the cache of that cluster, I'm curious why this cache still exists even after the cluster is gone. Do we have a chance to clean the cache?
> 
> https://github.com/karmada-io/karmada/blob/e7b6513db0e44090a9d11fcff15b1fec097b3b2a/pkg/search/proxy/store/multi_cluster_cache.go#L333-L336

@RainbowMango When the member cluster goes offline but the Cluster resources in the control plane are not deleted, it can prevent the offline clusters in the ResourceRegistry from being removed, resulting in the resource cache being retained for a short time.

xigang: > @xigang Hi, if a watch request is hanging and causes a timeout, will the hanging watch request continue to exist in the subprocess?

@zhzhuang-zju Fixed, please take a look.

pkg/search/proxy/store/multi_cluster_cache.go
```diff
@@ -334,15 +334,44 @@ func (c *MultiClusterCache) Watch(ctx context.Context, gvr schema.GroupVersionRe
...
-		w, err := cache.Watch(ctx, options)
-		if err != nil {
+
+		// The following logic adds a 30-second timeout to prevent watch requests to member clusters from hanging,
+		// which could cause the client watch to hang.
+		watchChan := make(chan watch.Interface, 1)
+		errChan := make(chan error, 1)
+
+		go func(cluster string) {
+			w, err := cache.Watch(ctx, options)
```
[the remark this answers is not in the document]
xigang: If the watcher times out after 30 seconds during creation, it will trigger a time.After timeout, return an error, and call cancel to stop the watcher goroutine.

```
https://github.com/karmada-io/karmada/blob/e65e993d2d0cf920b889daa7f5c36e94d6829787/vendor/k8s.io/apiserver/pkg/endpoints/handlers/get.go#L263
```

pkg/search/proxy/store/multi_cluster_cache.go
```diff
@@ -334,15 +334,44 @@ func (c *MultiClusterCache) Watch(ctx context.Context, gvr schema.GroupVersionRe
...
+		select {
+		case w := <-watchChan:
+			mux.AddSource(w, func(e watch.Event) {
+				setObjectResourceVersionFunc(cluster, e.Object)
+				addCacheSourceAnnotation(e.Object, cluster)
+			})
+		case err := <-errChan:
+			// If the watch request fails, return the error, and the client will retry.
 			return nil, err
+		case <-time.After(30 * time.Second):
```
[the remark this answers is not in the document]
xigang: Parallel processing would work here as well; I'll adjust the code.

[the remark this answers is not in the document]
xigang: > It seems wait 30s for each cluster. Should we wait all clusters paralleled?
@ikaven1024 There’s no issue here; as long as a single `cache.Watch` times out, the Watch request will return with an error and end. There’s no problem with that here.😄

xigang:
@RainbowMango @ikaven1024 @zhzhuang-zju Our production environment is running without issues; could you take a look to see if the code can be merged?

pkg/search/proxy/store/multi_cluster_cache.go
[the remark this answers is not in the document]
xigang: > > > It seems wait 30s for each cluster. Should we wait all clusters paralleled?
> > 
> > 
> > @ikaven1024 There’s no issue here; as long as a single `cache.Watch` times out, the Watch request will return with an error and end. There’s no problem with that here.😄
> 
> While if every cluster create watching takes 20s, not timeout, the total time spends 20s * N.

@ikaven1024 I understand, but the code in the current release also traverses all the clusters to create watches, which takes N*M time.

zhzhuang-zju:
https://github.com/karmada-io/karmada/blob/0cc294f99898e4933b7874e9c5a3503d382f42de/vendor/k8s.io/apiserver/pkg/endpoints/handlers/get.go#L262-L268

@xigang The context used when establishing a watch request on the cache also has a timeout. Can we directly utilize this context to prevent the watch request from hanging?

https://github.com/karmada-io/karmada/blob/0cc294f99898e4933b7874e9c5a3503d382f42de/vendor/k8s.io/apiserver/pkg/endpoints/handlers/get.go#L254-L260
https://github.com/karmada-io/karmada/blob/0cc294f99898e4933b7874e9c5a3503d382f42de/cmd/karmada-search/app/karmada-search.go#L210-L215
The timeout for the context is actually initialized when setting up `search.Config`, with a value of 1800 seconds. It might be worth considering setting this to 30 seconds instead.

pkg/search/proxy/store/multi_cluster_cache.go
```diff
@@ -334,15 +334,44 @@ func (c *MultiClusterCache) Watch(ctx context.Context, gvr schema.GroupVersionRe
...
+		watchChan := make(chan watch.Interface, 1)
+		errChan := make(chan error, 1)
+
+		go func(cluster string) {
+			w, err := cache.Watch(ctx, options)
+			if err != nil {
+				select {
+				case errChan <- fmt.Errorf("failed to start watch for resource %v in cluster %q: %v", gvr.String(), cluster, err):
+				case <-ctx.Done():
+				}
```
zhzhuang-zju, line 349: Can `case <-ctx.Done()` be reached? And is it necessary to respond to the `ctx.Done` event?

xigang:
> https://github.com/karmada-io/karmada/blob/0cc294f99898e4933b7874e9c5a3503d382f42de/vendor/k8s.io/apiserver/pkg/endpoints/handlers/get.go#L262-L268
> 
> ~@xigang The context used when establishing a watch request on the cache also has a timeout. Can we directly utilize this context to prevent the watch request from hanging?~
> 
> https://github.com/karmada-io/karmada/blob/0cc294f99898e4933b7874e9c5a3503d382f42de/vendor/k8s.io/apiserver/pkg/endpoints/handlers/get.go#L254-L260
> 
> 
> https://github.com/karmada-io/karmada/blob/0cc294f99898e4933b7874e9c5a3503d382f42de/cmd/karmada-search/app/karmada-search.go#L210-L215
> 
> 
> ~The timeout for the context is actually initialized when setting up `search.Config`, with a value of 1800 seconds. It might be worth considering setting this to 30 seconds instead.~
> This context represents the sum of execution times of multiple functions, not only `watch`

@zhzhuang-zju The timeout for client-go watch requests is around 10 minutes. The default value of minRequestTimeout is 1800 seconds, and if the client specifies a timeout, minRequestTimeout will not be used. What issues might arise if you set this Context for federation watch and all member clusters' Watch?

pkg/search/proxy/store/multi_cluster_cache.go
xigang: The main purpose here is to handle edge cases. There won’t be any issues even if there is no case <-ctx.Done().

RainbowMango:
Can someone tell me how to reproduce this issue? I'd like to reproduce and verify the patch on my side.

zhzhuang-zju:
> @zhzhuang-zju What issues might arise if you set this Context for federation watch and all member clusters' Watch?

Sorry, please ignore the [comment](https://github.com/karmada-io/karmada/pull/5732#issuecomment-2461198865). Initially, I thought about setting the timeout for the context to 30 seconds to solve the issue of the watch request hanging. And then, I realized that this approach was not appropriate, so I commented it out.

pkg/search/proxy/store/multi_cluster_cache.go
zhzhuang-zju: If removing the `case <-ctx.Done()` statement does not affect the actual functionality, it is recommended to remove it to improve code readability.

xigang:
> Can someone tell me how to reproduce this issue? I'd like to reproduce and verify the patch on my side.

In our scenario, a member cluster is taken offline, and all nodes within the cluster are shut down. At this point, the cluster name still exists in the ResourceRegistry, so the Pod Cacher also remains. Additionally, during the cluster decommissioning, client Watch long connections continue to access the federate-apiserver.

As a result, the client’s Watch requests end up hanging.

PR #5319, opened 2024-08-07
skip cleanup when the poliy claim metadata changed

zhzhuang-zju:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
Refer to #5307, there is a chance that poliy marks be lost when deleting the old pp/cpp and binding a new one. When performing a clean up of marks on rb/crb, we can compare the value of the label `policyv1alpha1.ClusterPropagationPolicyPermanentIDLabel/policyv1alpha1.PolicyPermanentIDLabel` on the cb/crb with the policyId value of the deleted CPP/PP. If they are not the same, it indicates that the rb/crb has already claimed to a new CPP/PP, and there is no need to clean up the label/annotation.

**Which issue(s) this PR fixes**:
Fixes #5307

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
`karmada-controller-manager`: Fixed an issue that policy claim metadata might be lost during the rapid deletion and creation of PropagationPolicy(s)/ClusterPropagationPolicy(s)
```

PR #5741, opened 2024-10-24
cherry pick of #5319: skip cleanup when the policy claim metadata changed

zhzhuang-zju:
Cherry pick of https://github.com/karmada-io/karmada/pull/5319 on release-1.11.
https://github.com/karmada-io/karmada/pull/5319: skip cleanup when the policy claim metadata changed

```release-note
`karmada-controller-manager`: Fixed an issue that policy claim metadata might be lost during the rapid deletion and creation of PropagationPolicy(s)/ClusterPropagationPolicy(s)
```

### DIFF
--- a/pkg/detector/detector.go
+++ b/pkg/detector/detector.go
@@ -78,6 +78,20 @@ var (
 	}
 )
 
+// NeedCleanupPolicyMarks determines whether the object's policy marks needs to be cleaned up.
+// We need to ensure that the policy marks being deleted belong to the current PropagationPolicy/ClusterPropagationPolicy,
+// otherwise, there is a risk of mistakenly deleting the ones belonging to another PropagationPolicy/ClusterPropagationPolicy.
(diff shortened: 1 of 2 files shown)


RainbowMango:
This is a manual cherry-pick, right?

Then the risk would be beyond the issue itself. 

I'm kind of hesitant to cherry-pick this PR.  @whitewindmills Do you think it is worth backporting this fix to the previous release?

zhzhuang-zju:
> This is a manual cherry-pick, right?

Yes, the bugfix on the master branch was done based on the refactoring, so I chose to manually cherry-pick it.

whitewindmills:
this bug is unlikely to occur in a real environment, right? at least no one has reported this issue in the past few releases. so I think there is no value in backporting this fix.

zhzhuang-zju:
> this bug is unlikely to occur in a real environment, right? at least no one has reported this issue in the past few releases. so I think there is no value in backporting this fix.

Yes, this was also discovered due to CI failures. Based on the probability of occurrence, we can wait for the issue to appear in real usage scenarios before making a fix.

PR #5743, opened 2024-10-25
Disable a single resource multi dependency distribution

XiShanYongYe-Chang:
**What type of PR is this?**

/kind feature

From the majority of user feedback received so far, there are no scenarios where a single resource is depended on by multiple resources, or where a resource is both depended on by other resources and also distributed through PropagationPolicy/ClusterPropagationPolicy. Therefore, we have decided to restrict the above usage in the Karmada system, with the ultimate goal of ensuring that a resource can only be subject to distribution decisions by one PropagationPolicy/ClusterPropagationPolicy. 

By imposing such a restriction, we can make the behavior of the Karmada system more explicit and also facilitate the iteration of the follow-on distribution feature.

Fixes #

The current PR is far from the end. In order to complete the entire restriction activity, I plan to carry out the following tasks:

1. Add events for the target resources to describe the restriction operations performed by the Karmada system.
2. In the next version (v1.13), deprecate the DisableMultiDependencyDistribution switch and clean up the existing code, no longer considering the situation where a resource is depended on by multiple resources. For example, labels with the prefix resourcebinding.karmada.io/depended-by- will be deprecated.
3. Optimize the restriction logic so that when a resource is depended on by another resource and an attachedBinding is generated, an annotation or label is added to the resource to record its depender, thus avoiding the need to query the corresponding resourceBinding object during restriction judgment. Additionally, this information may also be processed through karmadactl for further extensions.

```release-note
karmada-controller-manager: disable a single resource multi dependency distribution
```

### DIFF
--- a/cmd/controller-manager/app/options/options.go
+++ b/cmd/controller-manager/app/options/options.go
@@ -143,6 +143,17 @@ type Options struct {
 	// in scenario of dynamic replica assignment based on cluster free resources.
 	// Disable if it does not fit your cases for better performance.
 	EnableClusterResourceModeling bool
+
+	// DisableMultiDependencyDistribution indicates disable the ability to a resource from being depended on by multiple
+	// resources or being distributed by PropagationPolicy/ClusterPropagationPolicy while being depended on.
+	//
+	// Before v1.12, this capability is allowed by default. If you still wish to enable this capability, you can set
+	// this flag to false. However, you will need to bear some side effects that come with it.
+	// For example, you can refer to https://github.com/karmada-io/karmada/pull/5717. When the primary resource is deleted,
+	// it does not consider other resources that currently depend on the resource or any PropagationPolicy associated with it.
+	//
+	// It is recommended that you adapt your business accordingly to avoid continued use.
+	DisableMultiDependencyDistribution bool
 }
 
 // NewOptions builds an empty options.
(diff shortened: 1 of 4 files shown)


pkg/detector/detector.go
```diff
@@ -245,10 +250,22 @@ func (d *ResourceDetector) Reconcile(key util.QueueKey) error {
 			// currently we do that by setting owner reference to derived objects.
 			return nil
 		}
-		klog.Errorf("Failed to get unstructured object(%s), error: %v", clusterWideKeyWithConfig, err)
+		klog.Errorf("Failed to get unstructured object(%s), error: %v", clusterWideKeyWithConfig.ClusterWideKey, err)
 		return err
 	}
 
+	if d.DisableMultiDependencyDistribution {
```
chaunceyjiang, line 257: Can you remind me what situation this logic is handling?

XiShanYongYe-Chang: A parameter is added to determine whether the Karmada system allows the same resource to be dependent on different resources. By default, the value is true, indicating that multiple dependencies will be prohibited in future evolution, including that a resource is processed by the PropagationPolicy and distributed along with the resource.

chaunceyjiang: I understand the purpose of your `DisableMultiDependencyDistribution` parameter. But I haven't grasped what the code you wrote in the `detector` is meant to handle.

Looking at the e2e scenario, this code seems to be breaking compatibility.

XiShanYongYe-Chang: >  But I haven't grasped what the code you wrote in the detector is meant to handle.

The logic of this code is incorrect. I will correct it later.

> Looking at the e2e scenario, this code seems to be breaking compatibility.

When `DisableMultiDependencyDistribution` is false, it adds some restrictions, and it really can be understood as break change.
When `DisableMultiDependencyDistribution` is true, it keeps the original behavior.

RainbowMango (commented):
/assign

cmd/controller-manager/app/options/options.go
```diff
@@ -143,6 +143,17 @@ type Options struct {
...
+	// DisableMultiDependencyDistribution indicates disable the ability to a resource from being depended on by multiple
+	// resources or being distributed by PropagationPolicy/ClusterPropagationPolicy while being depended on.
+	//
+	// Before v1.12, this capability is allowed by default. If you still wish to enable this capability, you can set
+	// this flag to false. However, you will need to bear some side effects that come with it.
+	// For example, you can refer to https://github.com/karmada-io/karmada/pull/5717. When the primary resource is deleted,
+	// it does not consider other resources that currently depend on the resource or any PropagationPolicy associated with it.
+	//
+	// It is recommended that you adapt your business accordingly to avoid continued use.
+	DisableMultiDependencyDistribution bool
```
RainbowMango, line 156: Would it be better to have a feature gate than a command flag? Something like:  `DisableDependencySharing`.
For example:
- release-1.12 starts from alpha, which is disabled by default, for backward compatibility concerns.
- release-1.23 move to beta, which is enabled by default, and allows people to disable it as an escape hatch
- release-1.24 move to stable, enabled by default, and don't allow to disable it.

@XiShanYongYe-Chang @chaunceyjiang What do you think?

PR #5754, opened 2024-10-29
pkg/karmadactl/addons: unit test install addons

mohamedawnallah:
**Description**

In this commit, we unit test install addons to make sure the following addons `karmada-descheduler`, `karmada-metrics-adapter`, `karmada-scheduler-estimator`, and `karmada-search` are already configured.

**What type of PR is this?**

Part of #5491.

```release-note
NONE
```


pkg/karmadactl/addons/install/install_test.go
```diff
@@ -0,0 +1,71 @@
...
+	Install()
+
+	tests := []struct {
+		name          string
+		key           string
+		expectedAddon interface{}
+	}{
+		{
+			name:          "Install_WithKarmadaDeschedulerAddon_Installed",
+			key:           "karmada-descheduler",
```
XiShanYongYe-Chang, line 41: It is recommended that you change the component name to use constants so that the entire code base can be unified. If you think it is feasible, you can mention a separate pr.

XiShanYongYe-Chang (commented):
/lgtm
/approve

pkg/karmadactl/addons/install/install_test.go
mohamedawnallah: Good idea! I've mentioned this in a separate PR #5908. I would love to receive any feedback! 🙏

Issue #5767, opened 2024-11-01
[Feature] Expose APIServer Service provisioned by karmada-operator 


RainbowMango:
**Background**
Currently, there is no reliable method to reference the Service for the Karmada APIServer. Users have to infer the Service based on internal implementation details, which can change over time, making this approach fragile and unreliable. This feature aims to provide a robust and stable way to expose Karmada services, particularly the APIServer, to external clusters or networks.

**Proposal**
https://github.com/karmada-io/karmada/pull/5599

The goal of this feature is to enhance the karmada-operator to include the Karmada API Server Service information in the status of Karmada resource objects. This will provide a standardized and reliable way for third-party applications to discover and interact with the Karmada API Server Service. The feature will ensure that the service information is consistently available and up-to-date, regardless of internal implementation changes.

**Iteration Tasks**
- [x] Feature proposal. (@jabellard, #5599)
- [x] Restrict the supported [ServiceType](https://github.com/karmada-io/karmada/blob/057cf860f41afc60f7505196c6df5979a2642068/operator/pkg/apis/operator/v1alpha1/type.go#L280-L283) by adding some validations. (@RainbowMango, #5769)
- [x] Enable Service type : `LoadBalancer`. (@chaosi-zju, #5773)
  * The [runUploadAdminKubeconfig()](https://github.com/karmada-io/karmada/blob/adef1e59748e1e1d31cb75fffe406b5dd69a66d7/operator/pkg/tasks/init/upload.go#L62-L114) didn't handle service with type `LoadBalancer`.
- [x] Feature implementation
  - [x] Code (@jabellard, https://github.com/karmada-io/karmada/pull/5775)
  - [x] Test (@jabellard, https://github.com/karmada-io/karmada/pull/5775)
  - [x] Documentation (@RainbowMango, #6015)
    * We should have a doc to introduce how to access the karmada-apiserver, like [Access The Argo CD API Server](https://argo-cd.readthedocs.io/en/stable/getting_started/#3-access-the-argo-cd-api-server)

RainbowMango:
@jabellard @chaosi-zju What do you think of having a doc for this feature?
Let me know if you are going to help with it.

jabellard:
> @jabellard @chaosi-zju What do you think of having a doc for this feature? Let me know if you are going to help with it.

[...]

Issue #4634, opened 2024-02-16
When installing karmada using the operator, provide a way to annotate karmadaAPIServer services


endrec:
**What would you like to be added**:

The current schema provides a `spec.components.karmadaAPIServer.annotations` field, but this is only used for the created pod, and not the service.
It would be useful to specify annotations on the service as well, either by populating service annotations using this same field, or by introducing a new field `serviceAnnotations`.

**Why is this needed**:

Annotations on `service` resources are used by different add-ons. 
In my current use case I'd like to expose karmada-api to argoCD, but our argoCD runs in a different cluster, even in a different cloud. I can see two ways to achieve this:
- use `serviceType: LoadBalancer` and annotate the service with `external-dns.alpha.kubernetes.io/hostname`, or
- create an `Ingress`. I'm using Google's built in ingress controller, which needs an annotation on the service to enable HTTPS traffic to the backend. (Nginx uses an annotation on the ingress to achieve the same.)

RainbowMango:
It's a reasonable use case, thanks for the spotting.
cc operator authors here . @calvin0327 and also @chaosi-zju

calvin0327:
This case does make sense,  introducing a field `serviceAnnotations` is easy to implement the feature, but i don't know whether it's the best way? @chaosi-zju what do you think?

RainbowMango:
Hi @endrec the feature that enables annotate karmadaAPIServer services has been included in v1.10. Don't know if you noticed that~ And looking forward to any feedback.

PR #5769, opened 2024-11-01
Restrict supported service types of Karamda APIServer supported by karmada-operator

RainbowMango:
**What type of PR is this?**

/kind cleanup
/kind api-change

This PR restrics the service types supported by karmada-operator.

Part of #5767 

This PR introduces validation rules which narrow down the range of acceptable configuration, but this has no impact on users who are currently using `ClusterIP` and `NodePort`.  And for `Loadbalancer` and `ExternalName`, currently not supported. (We are considering introduce the support of `Loadbalancer` to address the use case reported at #4634).

```release-note
API change: The ServiceType of Karmada API server now has been restrict to `ClusterIP`, `NodePort` and `LoadBalancer`.
```


RainbowMango:
cc @chaosi-zju @jabellard @calvin0327

PR #5599, opened 2024-09-24
Proposal: Add API Server Service Information to `KarmadaStatus`

jabellard:
**What type of PR is this?**

/kind design

**What this PR does / why we need it**:
- Details are contained in the proposal.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #5775, opened 2024-11-02
Add API Server Service Information to `KarmadaStatus`

### LINKED ISSUE #5567: Add API Server Service Info to Karmada Status
Once a Karmada instance has been provisioned by the operator,  it's status will contain [a ref to the secret](https://github.com/karmada-io/karmada/blob/release-1.11/operator/pkg/apis/operator/v1alpha1/type.go#L649) that contains the admin kubeconfig for the control plane. As such, there's no need to guess where that secret is stored. Similarly, I think it'd be useful to also include the name of  the service for the control plane's API server in the status.

**Why is this needed**:
At Bloomberg, we're currently building a managed Karmada platform and want to use the Karmada operator to manage the entire lifecycle of managed Karmada instances. Tenant control planes will run in our management clusters. As such, the Karmada operator will be running in those clusters. Additionally, we'll also have a higher level operator running in those clusters that, at a very high level, will do the following:
- Delegate all base control plane functionality to the Karmada operator
-  Perform any additional tasks required for provisioning a tenancy such as creating an ingress resource to allow external traffic to the tenancy's API server running in the management cluster

As of now, when creating an ingress for an API server, we have to guess the name of the service for that server. Although it seems like  there's some convention used for generating the service names, relying on that is a brittle approach as it depends on knowledge of an implementation detail that may likely change in the future.

jabellard:
**What type of PR is this?**

/kind feature

This is an implementation of this [proposal](https://github.com/karmada-io/karmada/pull/5599) for adding information about the API Server service of a Karmada instance to its status.

Fixes #5567
Part of #5767

```release-note
The new `ApiServerService` field added as part of the Karmada instance status can be used to reference the API Server service for that instance. This is useful for scenarios where higher level operators need to discover the API Server service of a Karmada instance  for tasks like setting up ingress traffic.
```


jabellard:
@RainbowMango

RainbowMango (commented):
/assign

RainbowMango (approved):
/lgtm
/approve

PR #5779, opened 2024-11-04
Add index for work

zach593:
**What type of PR is this?**

not sure which kind this is

Add index for `Work` resource could speed up the processing of resource binding related controllers.

In one of our enviorment with 300k+ `works`, this commit helps increase the progressing speed about 30X times on `resourcebinding controller` and `resourcebinding status controller,` and lower the cpu usage about 60%.

Here's a monitoring data comparision with/without this commit, in one of our simulated enviorment.

enviorment configuration: 

- 48 cpus for karmada-controller-manager
- 2 member cluters
- 180k workloads
- 180k resourcebindings
- 780k works (there's some works are not created by karmada)

There's some difference between our version and community version of karmada-controller-manager, the one that will affect the result is, in `resroucebinding status controller`, the create predicate of `resourcebinding` and `work` both return true.

Without this commit:

image
image
image

With this commit:

image
image
image

Fixes #

```release-note
NONE
```


RainbowMango (commented):
/assign
Interesting, thanks.

RainbowMango:
Failing test is unrelated.
/retest
```
make: Leaving directory '/home/runner/work/karmada/karmada'
Image: "docker.io/karmada/karmada-controller-manager:latest" with ID "sha256:17b6c6b5f4675595f1db1c8d05a9724f65170e449aacfb2cf2a2c6eca71accc7" found to be already present on all nodes.
namespace/karmada-system created
customresourcedefinition.apiextensions.k8s.io/karmadas.operator.karmada.io created
clusterrole.rbac.authorization.k8s.io/karmada-operator created
clusterrolebinding.rbac.authorization.k8s.io/karmada-operator created
serviceaccount/karmada-operator created
deployment.apps/karmada-operator created
error: no matching resources found
```

RainbowMango:
HI @zach593 , there are some conflicts you might need to resolve after #5780.

pkg/util/helper/work.go
```diff
@@ -117,15 +118,19 @@ func GetWorksByLabelsSet(ctx context.Context, c client.Client, ls labels.Set) (*
...
-		ls = labels.Set{workv1alpha2.ClusterResourceBindingPermanentIDLabel: bindingID}
+		key = workv1alpha2.ClusterResourceBindingPermanentIDLabel
 	}
-
-	return GetWorksByLabelsSet(ctx, c, ls)
+	workList := &workv1alpha1.WorkList{}
+	listOpt := &client.ListOptions{
+		FieldSelector: fields.OneTermEqualSelector(key, bindingID),
+	}
+	return workList, c.List(ctx, workList, listOpt)
```
RainbowMango, line 133: I believe this can significantly accelerate the query efficiency, as it doesn't need to query all Works.

I'm thinking about how to make a benchmark test for this, I'm curious how much performance it enhanced based on the upstream codebase. Any thoughts?

test/e2e/suite_test.go
```diff
@@ -173,6 +176,11 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
 
 	framework.InitClusterInformation(karmadaClient, controlPlaneClient)
 
+	controllerManager = framework.InitControllerManagerAndStartCache(restConfig)
+
+	err = utilhelper.IndexWork(context.TODO(), controllerManager)
+	gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+
 	testNamespace = fmt.Sprintf("karmadatest-%s", rand.String(RandomStrLength))
 	err = setupTestNamespace(testNamespace, kubeClient)
```
zach593, line 185: Hi, @RainbowMango 

I just noticed this part, is it all e2e tests are using a shared cluster?

pkg/util/helper/work.go
zach593: One thing for sure is, if the test result is triggered by CI, then we must add the test before this commit.

test/e2e/suite_test.go
zach593: Just checked ci and scripts and it seems the cluster is built via local-up-karmada.sh, so I have no questions now.

pkg/util/helper/index.go
```diff
@@ -0,0 +1,40 @@
...
+		GenOneLabelEqualIndexerFunc(workv1alpha2.ClusterResourceBindingPermanentIDLabel))
+	if err != nil {
+		klog.Errorf("failed to create index for work, err: %v", err)
+		return err
+	}
+	return nil
+}
+
+// GenOneLabelEqualIndexerFunc returns an IndexerFunc used to index resource with the given key as label key.
+func GenOneLabelEqualIndexerFunc(key string) client.IndexerFunc {
```
RainbowMango, line 32: Can we rename the function here? like
```golang
// IndexerFuncBasedOnLabel returns an IndexerFunc used to index resources with the given label key.
func IndexerFuncBasedOnLabel(key string) client.IndexerFunc {
```

cmd/controller-manager/app/controllermanager.go
```diff
@@ -741,6 +741,10 @@ func setupControllers(mgr controllerruntime.Manager, opts *options.Options, stop
 	sharedFactory.Start(stopChan)
 	sharedFactory.WaitForCacheSync(stopChan)
 
+	if err := helper.IndexWork(context.TODO(), mgr); err != nil {
+		klog.Fatalf("Failed to index Work: %v", err)
+	}
```
RainbowMango, line 746: I think we can build the indexer at the place where configure the controller manager because it's kind of an overall configuration.
`setupControllers` is mostly for configure controllers.
 https://github.com/karmada-io/karmada/blob/20af47da0b968bd83cae482c0017f1346071a036/cmd/controller-manager/app/controllermanager.go#L193

We can set the index just before setupControllers.

test/e2e/suite_test.go
```diff
@@ -173,6 +176,11 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
 
 	framework.InitClusterInformation(karmadaClient, controlPlaneClient)
 
+	controllerManager = framework.InitControllerManagerAndStartCache(restConfig)
+
+	err = utilhelper.IndexWork(context.TODO(), controllerManager)
+	gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
```
RainbowMango, line 182: Could you please remind me where referencing the controllerManager?

In my opinion, the E2E tests don't need to build controllers themselves.

RainbowMango (commented):
Generally looks good to me, just some suggestions and questions.

In addition, wait for a while to explore how to build a benchmark for this patch.

test/e2e/suite_test.go
zach593: https://github.com/karmada-io/karmada/blob/20af47da0b968bd83cae482c0017f1346071a036/test/e2e/clusterresourcebinding_test.go#L83-L86

any suggestions for that case?

cmd/controller-manager/app/controllermanager.go
zach593: updated

pkg/util/helper/index.go
zach593: updated, and slightly changed the implementation

pkg/util/helper/work.go
zach593: I think I found a way to benchmark this part without using fakeClient or network requests.

The `cache.Cache` from controleller-runtime could be either `client.Reader` and `client.FieldIndexer`, so we could use this cache as a resource provider to pass into the function. But we may need to change the signature of function `GetWorksByBindingID()`, replace `c client.Client` with `c client.Reader`.  For the benchmark without index, we can only write another function.

As a scenario with different performances when the amount of resources is different, I think the amount of resources can be used as the input parameter of the benchmark. This is different from the common benchmark, but I think it will be more effective in showing performance gaps.

test/e2e/suite_test.go
RainbowMango: cc @XiShanYongYe-Chang to look at it.

XiShanYongYe-Chang: Let me take a look.

> In my opinion, the E2E tests don't need to build controllers themselves.

I agree with this point.

I have looked at the E2E test content. Perhaps our case check is not end-to-end enough. After all, Work is a product in the middle of the system.

In addition, if the case needs to be retained, the GetWorksByBindingID method invoked by the case can be maintained separately in the E2E without introducing a controllerManager in the E2E because the original invoked function changes.

For the current case, my suggestion is that we delete it first. What do you think?

XiShanYongYe-Chang: Hi @zach593 do you need me to bring up a pr to clean up?

zach593: If the direction is decoupled from this helper function, I think we can use `helper.GetWorksByLabelsSet()` here. This way is also very simple, and it maintains the original effect.

pkg/util/helper/work.go
zach593: Updated. I finally chose to use the envtest package in controller-runtime to start the local apiserver, to implement the benchmark.

kindly ping @RainbowMango to review again.

image

pkg/util/helper/work_test.go
```diff
@@ -107,3 +130,165 @@ func TestGenEventRef(t *testing.T) {
...
+			LabelSelector: labels.SelectorFromSet(labels.Set{key: bindingID}),
+		}
+		return workList, c.List(ctx, workList, listOpt)
+	}
+	benchmarkGetWorksByBindingID(b, false, GetWorksByBindingID)
+}
+
+func BenchmarkGetWorksByBindingID(b *testing.B) {
+	benchmarkGetWorksByBindingID(b, true, GetWorksByBindingID)
+}
```
RainbowMango, line 153: Can you help to share the benchmark result here?

I tried to run it on my side but failed as I didn't prepare the CRDs and enviornment.

RainbowMango (commented):
Hi @zach593
Great thanks for your hard work!
I don't have any further comments except the benchmark. 

I feel the benchmark is a bit too complex, so my suggestion is to remove the benchmark from this PR(but I would appreciate it if you could provide some test reports) for now.

I still believe the benchmark is important. However, I hope there will be a more generic way, so that it can be used for the other index as well.

Let me know what you think.

pkg/util/helper/work_test.go
zach593: I post the result in https://github.com/karmada-io/karmada/pull/5779#discussion_r1836746062 
With 50k works, the time taken by the function using index is reduced by 3000 times. Of course, this will have different results at other orders of magnitude, but it is a bit troublesome to test them one by one.

zach593:
I think there is a conflict here. If there is a ready environment for each function that needs to benchmark cache and index, then these functions are likely to affect each other during testing. Otherwise, as you can see in this PR, there must be a lot of logic to initialize the environment, data and cache before testing.

Unless we use fakeClient when benchmarking, but its structure is too different from the real client, the benchmark result should not be convincing.

RainbowMango:
> I think there is a conflict here. If there is a ready environment for each function that needs to benchmark cache and index, then these functions are likely to affect each other during testing.

I agree. It's complicated. It needs further exploration, so I suggest removing the benchmark part from this PR.

zach593:
> I agree. It's complicated. It needs further exploration, so I suggest removing the benchmark part from this PR.

Of course. Before I remove it, would you like to see the benchmark results at other `Work` amounts?

RainbowMango:
> Of course. Before I remove it, would you like to see the benchmark results at other Work amounts?

I believe this approach is effective, but it would be even better if you had more test reports to share. 

By the way, please squash the commits when next push.

RainbowMango:
Thanks @zach593 .
The failing tests are not related and will be fixed by #5834.
I will take care of it after #5834.

RainbowMango:
> If anyone wants to run this test on their enviorment, I think they can use [405da48](https://github.com/karmada-io/karmada/commit/405da483b576165f59c98e0334d5ae2c6636ee5f).

Great! I also backed up that commit on my side. :):)

zach593:
@RainbowMango done

RainbowMango (approved):
/lgtm
/approve

zach593:
Add some test reports.

image
image
image

RainbowMango:
Great thanks @zach593.

The test result confirms the effect of the performance improvement.

By the way, at the last community meeting, we are going to build a performance team that will focus on performance improvement, let me know if you are interested in being part of it.

zach593:
Ah, yes, of course I am very interested, thank you very much for inviting me. @RainbowMango

pkg/util/helper/index.go
```diff
@@ -0,0 +1,40 @@
...
+// IndexerFuncBasedOnLabel returns an IndexerFunc used to index resource with the given key as label key.
+func IndexerFuncBasedOnLabel(key string) client.IndexerFunc {
+	return func(obj client.Object) []string {
+		val, ok := obj.GetLabels()[key]
+		if !ok {
+			return nil
+		}
+		return []string{val}
+	}
+}
```
chaosi-zju, line 40: sorry to add this, but is there any length restriction on the index value?

zach593: I've thought about how to handle this part. I think if a work object really has a label like `"xxx/permanent-id": ""`, allowing it to be indexed should be appropriate. So, there is no length restriction on the index value.
However, in `GetWorksByBindingID()`, I wrote it simply and didn't check the label value length. I don't think this will have any actual impact. If it does, please let me know.

PR #5780, opened 2024-11-04
fix crb_status_controller ut is referencing namespaced objects

zach593:
**What type of PR is this?**

/kind cleanup

It is clearly that crb should only handle root scope resources, so it is inappropriate for the tests in the crb status controller to reference pod.

Fixes #

```release-note
NONE
```


RainbowMango (commented):
/lgtm
/approve
Thanks.

PR #5801, opened 2024-11-11
karmada-search: support field selector for corev1 resources

SataQiu:
**What type of PR is this?**
/kind feature

karmada-search: support field selector for corev1 resources.

This supports most scenarios that use fieldSelector, such as finding a list of pods by node name.

Fixes #

without this patch:
```sh
 # kubectl get pod --kubeconfig karmada-search.config --field-selector="spec.nodeName=member1-worker"
Error from server (BadRequest): Unable to find "/v1, Resource=pods" that match label selector "", field selector "spec.nodeName=member1-worker": "spec.nodeName" is not a known field selector: only "metadata.name", "metadata.namespace"
```

with this patch:
```sh
 # kubectl get pod --kubeconfig karmada-search.config --field-selector="spec.nodeName=member1-worker"
example-app-6c847589df-xmjm5                2024-10-29T03:15:25Z
...
```

```release-note
karmada-search: support field selector for corev1 resources.
```

### DIFF
--- a/pkg/search/proxy/framework/plugins/cache/scheme.go
+++ b/pkg/search/proxy/framework/plugins/cache/scheme.go
@@ -1,0 +1,35 @@
+/*
+Copyright 2024 The Karmada Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cache
+
+import (
+	"k8s.io/apimachinery/pkg/runtime"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+
+	corev1 "github.com/karmada-io/karmada/pkg/search/proxy/framework/plugins/cache/apis/core/v1"
+)
+
+var cacheScheme = runtime.NewScheme()
+
+func init() {
+	AddToScheme(cacheScheme)
+}
+
+// AddToScheme applies all the stored functions to the scheme.
+func AddToScheme(scheme *runtime.Scheme) {
+	utilruntime.Must(corev1.AddToScheme(scheme))
+}
--- a/pkg/search/proxy/framework/plugins/cache/apis/core/v1/conversion.go
+++ b/pkg/search/proxy/framework/plugins/cache/apis/core/v1/conversion.go
@@ -1,0 +1,167 @@
+/*
+Copyright 2024 The Karmada Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1
+
+import (
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// addConversionFuncs ensures that the cache plugin can handle the field selectors for corev1 resources.
+// It is copied from "k8s.io/kubernetes/pkg/apis/core/v1/conversion.go".
+func addConversionFuncs(scheme *runtime.Scheme) error {
+	// Add field conversion funcs.
+	err := scheme.AddFieldLabelConversionFunc(SchemeGroupVersion.WithKind("Pod"),
+		func(label, value string) (string, string, error) {
+			switch label {
+			case "metadata.name",
+				"metadata.namespace",
+				"spec.nodeName",
+				"spec.restartPolicy",
+				"spec.schedulerName",
+				"spec.serviceAccountName",
+				"spec.hostNetwork",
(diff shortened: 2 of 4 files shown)


XiShanYongYe-Chang:
Cool! Thanks~
/assign

pkg/search/proxy/framework/plugins/cache/scheme.go
```diff
@@ -0,0 +1,136 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// cacheScheme is the runtime.Scheme for the cache plugin.
+var cacheScheme = runtime.NewScheme()
+
+func init() {
+	// This is to ensure that the cache plugin can handle the field selectors for these resources.
```
XiShanYongYe-Chang, line 30: Resources that are not listed here can't be processed, can they? So how do you determine which resources can be listed here? In other words, are there any plans to add resources that are not listed?

SataQiu: Thanks @XiShanYongYe-Chang I checked the resource kinds of Kubernetes and the corev1 group is the most important and the most common used. IMO, we should support the corev1 resource kinds first, and the other resource types can be gradually added based on community feedback.

XiShanYongYe-Chang: @SataQiu, thanks for your quick reply!

Where does the Kubernetes community maintain Corev1 resources?

> Waiting for the Kubernetes runtime scheme package to support automatic ConversionFuncs registration, the community is trying to remove the legacyscheme

Is the Kuberenetes community working on this plan? Do we have to rely on this plan to add resources in the future?

XiShanYongYe-Chang: Let me understand method two again, which means that when the upstream community implements this, other resources will automatically support it, and we don't need to register additional conversion functions. Am I right?

SataQiu: Yes, you understand correctly.

pkg/search/proxy/framework/plugins/cache/scheme.go
```diff
@@ -0,0 +1,136 @@
...
+			switch label {
+			case "type",
+				"metadata.namespace",
+				"metadata.name":
+				return label, value, nil
+			default:
+				return "", "", fmt.Errorf("field label not supported: %s", label)
+			}
+		},
+	)
```
XiShanYongYe-Chang, line 122: For these methods, can we directly import Kubernetes open functions?

pkg/search/proxy/framework/plugins/cache/scheme.go
```diff
@@ -0,0 +1,136 @@
...
+			case "metadata.namespace",
+				"metadata.name",
+				"spec.clusterIP",
+				"spec.type":
+				return label, value, nil
+			default:
+				return "", "", fmt.Errorf("field label not supported: %s", label)
+			}
+		},
+	)
```
XiShanYongYe-Chang, line 135: I don't seem to find any service code in the Kubernetes repository. Can you point it out?

XiShanYongYe-Chang (commented):
Thanks a lot~

pkg/search/proxy/framework/plugins/cache/scheme.go
XiShanYongYe-Chang: Thanks~

> According to users feedback, support more resource types, such as apps, these can be added in subsequent PRs

I think we can do what you say. In addition, can you help create an issue to track this task, and if we don't get feedback on other resources for a long time, we may be able to proactively plan in the future.

pkg/search/proxy/framework/plugins/cache/scheme.go
SataQiu: Emm... I guess we can't, because these functions are defined in the k/k repository, but we can't import it directly, which is not recommended.

pkg/search/proxy/framework/plugins/cache/scheme.go
SataQiu: https://github.com/kubernetes/kubernetes/blob/6b031e50b2abfe15434beef6c7e01056091a26ea/pkg/apis/core/v1/conversion.go#L494-L507

pkg/search/proxy/framework/plugins/cache/scheme.go
XiShanYongYe-Chang: Okay, I get it. It doesn't exit in the https://github.com/kubernetes/api repo.

pkg/search/proxy/framework/plugins/cache/scheme.go
XiShanYongYe-Chang: Oh, I can see why I can't find it, it hasn't existed since version 1.31 and my code is still in v1.28.

pkg/search/proxy/framework/plugins/cache/scheme.go
XiShanYongYe-Chang: Would it be better if we added a comment to the file stating that the implementation is referenced from xxx?

SataQiu: Good idea, updated.

XiShanYongYe-Chang:
Thanks~
/lgtm
/cc @yanfeng1992 @RainbowMango

yanfeng1992:
lgtm

yanfeng1992 (approved):

SataQiu:
@XiShanYongYe-Chang need approve, PTAL!

XiShanYongYe-Chang:
> @XiShanYongYe-Chang need approve, PTAL!

Let's wait for a review from @RainbowMango 

Hi @SataQiu, are you using Karmada now?

RainbowMango:
/assign
But hoping the author @ikaven1024 could take a look.

SataQiu:
ping @ikaven1024

Issue #5790, opened 2024-11-06
resource_match_policy  takes too long time


CharlesQQ:
**What would you like to be added**:
https://github.com/karmada-io/karmada/blob/b3d3bcde6622654a1ed048fbce29e3b2d2deaaf7/pkg/detector/detector.go#L356-L387

In order to find the matching propagationPolicy, all propagationPolicies will be listed, and the for loop will be used to find the matching propagationPolicy in turn. The for loop here may take too long.
Is there a better way to speed up the lookup?

**Why is this needed**:
As shown below, The resource_match_policy_duration_seconds_bucket metric indicates that the execution took more than 0.5 seconds or even 0.9 seconds; This may cause the execution time of the resource detector controller reconcile to be too long, and the workqueue queue to create a long backlog.
![image](https://github.com/user-attachments/assets/6f3f0559-94b4-474b-9597-efaeff6e0be4)

CharlesQQ:
By removing the outermost for loop and the call to the function ConvertToTypedObject, after testing, it can be seen that the performance of the resource detector controller is significantly optimized.

<img width="1235" alt="image" src="https://github.com/user-attachments/assets/4d759ff3-191b-448d-bb1d-b3b67f688583">

RainbowMango:
/assign @CharlesQQ 
In favor of #5802

CharlesQQ:
Adjusting the parameter --concurrent-resourcebinding-syncs from 5 to 30, the resourcebinding controller's queue backlog optimization effect is not obvious, and it takes almost 5 minutes.

<img width="1160" alt="image" src="https://github.com/user-attachments/assets/604e26e0-9652-4746-a82e-d52a08a1bc3c">

CharlesQQ:
For the resource detector controller, by adjusting the parameter concurrent-resource-template-syncs=60, the queue backlog is reduced from 16 minutes to less than 1 minute.

CharlesQQ:
By setting the parameters --kube-api-qps=200 --kube-api-burst=300, bind-controller's queue backlog is reduced to 1 minute and 30 seconds.

RainbowMango:
@CharlesQQ Could you please help to confirm if all tasks planned have been done?

/close
Plese reopen it if anything left.

PR #5802, opened 2024-11-11
feat(detector):  resource detector matched policy potimization

CharlesQQ:
**What type of PR is this?**
/kind feature

Fixes https://github.com/karmada-io/karmada/issues/5790

```release-note

```


CharlesQQ:
> In that case, if the creation time of deployment and policy are very close, maybe d.client could not find the policy.

Both obtain resource objects from the cache. I think If d.Client has this problem, d.InformerManager may also have it.

pkg/detector/compare.go
```diff
@@ -64,31 +70,36 @@ func getHighestPriorityPropagationPolicy(policies []*policyv1alpha1.PropagationP
...
 }
 
-func getHighestPriorityClusterPropagationPolicy(policies []*policyv1alpha1.ClusterPropagationPolicy, resource *unstructured.Unstructured, objectKey keys.ClusterWideKey) *policyv1alpha1.ClusterPropagationPolicy {
+func getHighestPriorityClusterPropagationPolicy(policies []policyv1alpha1.ClusterPropagationPolicy, resource *unstructured.Unstructured, objectKey keys.ClusterWideKey) *policyv1alpha1.ClusterPropagationPolicy {
 	matchedClusterPolicyImplicitPriority := util.PriorityMisMatch
 	matchedClusterPolicyExplicitPriority := int32(math.MinInt32)
 	var matchedClusterPolicy *policyv1alpha1.ClusterPropagationPolicy
 
 	for _, policy := range policies {
+		policyPointer := ptr.To(policy)
```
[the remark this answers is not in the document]
CharlesQQ: For range slicing, the elements is struct type, and there may be problems if you directly obtain the element content through &

XiShanYongYe-Chang:
You are advised to obtain the policy list from karmada-apiserver and check the performance consumption caused by karmada-apiserver to see if the benefits are greatly improved.

zach593:
Is the purpose of this PR to remove the dependency on unstructured?
Why do we need this generated typed informer manager instead of using the controller-runtime informer?

CharlesQQ:
> Why do we need this generated typed informer manager instead of using the controller-runtime informer?

If controller-runtime is used, the logic of resource detector resources entering the workqueue also needs to use controller-runtime.

zach593:
> If controller-runtime is used, the logic of resource detector resources entering the workqueue also needs to use controller-runtime.

This is not the case... I tried replacing the generic InformerManger with the controller-runtime cache and it passes all the tests and works perfectly. This did not require replacing the asyncWorker, just replacing the lister and informer was enough. 

Here is the code: https://github.com/ctripcloud/karmada/commit/e358fd1ab0f64d09769ad75f22eb2311022a576b

CharlesQQ:
> I tried replacing the generic InformerManger with the controller-runtime cache and it passes all the tests and works perfectly.

you replaced the informer with  controller-runtime, so you think use controller-runtime is better than genericInformer?

zach593:
> so you think use controller-runtime is better than genericInformer?

Yes, because controller-runtime informer is also a typed informer.

CharlesQQ:
+ @XiShanYongYe-Chang

pkg/detector/detector.go
```diff
@@ -132,24 +133,30 @@ func (d *ResourceDetector) Start(ctx context.Context) error {
...
+	cppInformer, err := d.ControllerRuntimeCache.GetInformer(ctx, &policyv1alpha1.ClusterPropagationPolicy{})
+	if err != nil {
+		klog.Errorf("Failed to get informer for ClusterPropagationPolicy: %v", err)
+		return err
+	}
+	_, err = cppInformer.AddEventHandler(clusterPolicyHandler)
+	if err != nil {
+		klog.Errorf("Failed to add event handler for ClusterPropagationPolicy: %v", err)
+		return err
+	}
```
zach593, line 159: you need to put those two `AddEventHandler()` after the `d.Processor.Run(d.ConcurrentResourceTemplateSyncs, d.stopCh)`, because `AddEventHandler()` will immediately start the workers, and in the end of the workers logic, they will call `d.Processor.Add()`

CharlesQQ: you mean d.Processor must run before d.ReconcilePropagationPolicy or d.ReconcileClusterPropagationPolicy trigger logic? If d.Processor has not been created, d.Processor.Add() might have problems?

zach593: yes, occasional panic on startup

CharlesQQ: I got it, tks

pkg/detector/detector.go
```diff
@@ -131,36 +132,41 @@ func (d *ResourceDetector) Start(ctx context.Context) error {
...
+	d.EventHandler = fedinformer.NewFilteringHandlerOnAllEvents(d.EventFilter, d.OnAdd, d.OnUpdate, d.OnDelete)
 	go d.discoverResources(30 * time.Second)
```
zach593, line 170: I would suggest adding a empty line before line 169, or moving this part after `d.Processor.Run()`

pkg/detector/detector.go
```diff
@@ -359,63 +365,59 @@ func (d *ResourceDetector) LookForMatchedPolicy(object *unstructured.Unstructure
...
+	return getHighestPriorityPropagationPolicy(policies, object, objectKey), nil
 }
 
 // LookForMatchedClusterPolicy tries to find a ClusterPropagationPolicy for object referenced by object key.
 func (d *ResourceDetector) LookForMatchedClusterPolicy(object *unstructured.Unstructured, objectKey keys.ClusterWideKey) (*policyv1alpha1.ClusterPropagationPolicy, error) {
 	klog.V(2).Infof("Attempts to match cluster policy for resource(%s)", objectKey)
-	policyObjects, err := d.clusterPropagationPolicyLister.List(labels.Everything())
+	policyList := &policyv1alpha1.ClusterPropagationPolicyList{}
+	err := d.Client.List(context.TODO(), policyList, &client.ListOptions{
+		UnsafeDisableDeepCopy: ptr.To(true),
```
zach593, line 399: would you consider move two of this to https://github.com/karmada-io/karmada/pull/5813? I think this feature needs to be considered as a whole.

CharlesQQ: > I think this feature needs to be considered as a whole.

This is new code, you mean remove UnsafeDisableDeepCopy in this pr, and add in pr 5813?

zach593: ah, yes.

pkg/detector/detector.go
```diff
@@ -359,63 +365,59 @@ func (d *ResourceDetector) LookForMatchedPolicy(object *unstructured.Unstructure
...
-		}
-
+	policies := make([]*policyv1alpha1.ClusterPropagationPolicy, 0, len(policyList.Items))
+	for _, policy := range policyList.Items {
 		if !policy.DeletionTimestamp.IsZero() {
 			klog.V(4).Infof("Cluster propagation policy(%s) cannot match any resource template because it's being deleted.", policy.Name)
 			continue
 		}
-		policyList = append(policyList, policy)
+		policyPtr := &policy
```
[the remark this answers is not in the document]
CharlesQQ: > we still need this? this repo is already using go 1.22.

In go 1.22.,  not need this, i'll update it

pkg/detector/detector.go
CharlesQQ: d.EventHandler  is used in function `d. discoverResources `, I mean create d. EventHandler before  `d.discoverResources` is enough maybe

pkg/detector/detector.go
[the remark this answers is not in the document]
zach593: I thought about it again, this behavior has nothing to do with the golang version or loopvar, it's no different than using `&policy` directly in `append()`, it's just a bit wordy.

pkg/detector/detector.go
zach593: What I mean is that `Processor`, `ppHandler`, `cppHandler`, each of them has a blank line as a separator. The two lines I quoted (169, 170) have little connection with cpp, but there is no separation between them and cppHandler, and they look like a whole. So I suggest either adding a blank line to separate them, or putting these two lines together with the Processor part above because they are more closely related.

pkg/detector/detector.go
CharlesQQ: so, This pr must br merged before 5813, but it's not guaranteed

pkg/detector/detector.go
CharlesQQ: I got it

RainbowMango:
> For pp and cpp resources, use controller-runtime informer to replace InformerManager to avoid CPU consumption caused by type conversion in reconcile

I'm interested in this approach. Thanks.

RainbowMango (commented):
/assign

pkg/detector/detector.go
RainbowMango: > because AddEventHandler() will immediately start the workers,

Which workers we are talking about here?

pkg/detector/detector.go
```diff
@@ -899,12 +900,24 @@ func (d *ResourceDetector) GetMatching(resourceSelectors []policyv1alpha1.Resour
 
 // OnPropagationPolicyAdd handles object add event and push the object to queue.
 func (d *ResourceDetector) OnPropagationPolicyAdd(obj interface{}) {
-	d.policyReconcileWorker.Enqueue(obj)
+	policyObj := obj.(*policyv1alpha1.PropagationPolicy)
+	policyObj.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   policyv1alpha1.GroupVersion.Group,
+		Version: policyv1alpha1.GroupVersion.Version,
+		Kind:    policyv1alpha1.ResourceKindPropagationPolicy,
+	})
```
RainbowMango, line 908: I'm not sure but remember the controller-runtime fills the missing GVK.  Can you confirm it?

By the way, if it indeed needs to set the GVK, please add a comment to explain the reason.

RainbowMango (commented):
Generally looks good to me. but hope to rebase this PR because the base image is a bit old. I will take another look after that. 

In addition, have you ever thought about introducing a dedicated `propagationpolicy-controller`? So that we don't have to use low-lever functionalities of controller-runtime?

pkg/detector/detector.go
zach593: > Which workers we are talking about here?

`policyReconcileWorker` and `clusterPolicyReconcileWorker`. Before this commit, these workers would be started after the `InformerManager` was started, so there would be no problem.

pkg/detector/detector.go
zach593: > I'm not sure but remember the controller-runtime fills the missing GVK. Can you confirm it?

It does not fill the GVK, I have also encountered it in my environment.

CharlesQQ: > I'm not sure but remember the controller-runtime fills the missing GVK. Can you confirm it?

I used controller-runtime, kubernetes go-client, karmada go-client and kubernetes dynamic go-client to create informers for PropagationPolicy and Deployment to observe whether the objects triggered by the handler contain GVK. The results are as follows:
 | client type      | PropagationPolicy |Deployment|Unstructured|
| ----------- | ----------- |----------- |----------- |
| controller-runtime      | YES       |NO|-|
| kubernetes go-client   |-        |NO|-|
| karmada go-client  | YES        |NO|-|
|kubernetes dynamic go-client | YES|-|YES|

CharlesQQ:
> In addition, have you ever thought about introducing a dedicated propagationpolicy-controller? So that we don't have to use low-lever functionalities of controller-runtime?

Has this been discussed before?

zach593:
image

Using typed clients for pp/cpp helped us reduce karmada-controller-manager memory usage by about 10% in one of our production environments.

CharlesQQ:
> It seems that K8S native objects are stored in etcd using Protobuf format, while CRD objects are stored in json format.

CRD use apiextension-apiserver to store into etcd, It use json format

But kube-apiserver use protobuf default.

https://github.com/kubernetes/kubernetes/blob/83bb5d570580a3f477737fec5c24ba8fc3554264/staging/src/k8s.io/apiextensions-apiserver/pkg/cmd/server/server.go#L29-L30

https://github.com/kubernetes/kubernetes/blob/83bb5d570580a3f477737fec5c24ba8fc3554264/staging/src/k8s.io/apiserver/pkg/server/options/etcd.go#L76-L84

image

pkg/detector/detector.go
```diff
@@ -901,12 +902,24 @@ func (d *ResourceDetector) GetMatching(resourceSelectors []policyv1alpha1.Resour
 
 // OnPropagationPolicyAdd handles object add event and push the object to queue.
 func (d *ResourceDetector) OnPropagationPolicyAdd(obj interface{}) {
-	d.policyReconcileWorker.Enqueue(obj)
+	policyObj := obj.(*policyv1alpha1.PropagationPolicy)
+	policyObj.SetGroupVersionKind(schema.GroupVersionKind{
```
XiShanYongYe-Chang, line 906: For pp/cpp, do we not need to set its gvk, because we do not need this information when consuming queues, so we can use a separate keyFunc.

pkg/detector/detector.go
```diff
@@ -361,63 +368,57 @@ func (d *ResourceDetector) LookForMatchedPolicy(object *unstructured.Unstructure
 	}
 
 	klog.V(2).Infof("Attempts to match policy for resource(%s)", objectKey)
-	policyObjects, err := d.propagationPolicyLister.ByNamespace(objectKey.Namespace).List(labels.Everything())
+	policyList := &policyv1alpha1.PropagationPolicyList{}
+	err := d.Client.List(context.TODO(), policyList, &client.ListOptions{
+		Namespace:             objectKey.Namespace,
+		UnsafeDisableDeepCopy: ptr.To(true),
```
XiShanYongYe-Chang, line 374: It would be nice to add some comments here to prevent changes to the `policyList` during subsequent iterations.

pkg/detector/preemption.go
```diff
@@ -263,25 +253,21 @@ func (d *ResourceDetector) fetchResourceTemplate(rs policyv1alpha1.ResourceSelec
 // and put the PropagationPolicy in the queue to trigger preemption.
 func (d *ResourceDetector) HandleDeprioritizedPropagationPolicy(oldPolicy policyv1alpha1.PropagationPolicy, newPolicy policyv1alpha1.PropagationPolicy) {
 	klog.Infof("PropagationPolicy(%s/%s) priority changed from %d to %d", newPolicy.GetNamespace(), newPolicy.GetName(), *oldPolicy.Spec.Priority, *newPolicy.Spec.Priority)
-	policies, err := d.propagationPolicyLister.ByNamespace(newPolicy.GetNamespace()).List(labels.Everything())
+	policyList := &policyv1alpha1.PropagationPolicyList{}
+	err := d.Client.List(context.TODO(), policyList, client.InNamespace(newPolicy.GetNamespace()))
```
XiShanYongYe-Chang, line 257: Why not we set `UnsafeDisableDeepCopy` here?

pkg/detector/detector.go
XiShanYongYe-Chang: After reading your discussion under #5813, I agree that it is better to move this setting to #5813 for unified discussion. 

On the one hand, we can consider the overall performance improvement brought by this setting and on the other hand, we can think about how to better avoid potential program threats brought by this setting, if there is no good protection measure. It seems a little difficult to block threats with reviewers alone.

pkg/detector/preemption.go
CharlesQQ: > Why not we set UnsafeDisableDeepCopy here?

It should be set, it should be missed

pkg/detector/detector.go
CharlesQQ: Controller-runtime has a detailed description of `UnsafeDisableDeepCopy`, which is very clear. Is there no need to add it?

https://github.com/karmada-io/karmada/blob/65e7f74c991631a5888f5176342af118f4d5712e/vendor/sigs.k8s.io/controller-runtime/pkg/client/options.go#L495-L499

pkg/detector/detector.go
CharlesQQ: > so we can use a separate keyFunc.

In this case, do we need a new structure to replace ClusterWideKey, excluding the gvk field?

CharlesQQ:
4000 workload, 12C cpu

before:
image

pkg/detector/detector.go
XiShanYongYe-Chang: The structure can be universal. When consuming the policy worker queue, we do not need to obtain the gvk.

CharlesQQ: maybe newFunc just need remove gvk check logic?

https://github.com/karmada-io/karmada/blob/9ea0382293c80fdf8cf13da914136dfc4df3ec51/pkg/util/fedinformer/keys/keys.go#L96-L103

XiShanYongYe-Chang: Yes, I think so.

zach593:
> In addition, have you ever thought about introducing a dedicated `propagationpolicy-controller`? So that we don't have to use low-lever functionalities of controller-runtime?

I am interested in this idea (I have been looking forward to it for a long time). This will unify the implementation of each worker, allowing other asyncWorkers to also enjoy the enhanced capabilities of controller-runtime, eliminating the need to maintain the implementation of asyncWorkers, and unify the prometheus monitoring.

CharlesQQ:
>  I am interested in this idea (I have been looking forward to it for a long time). This will unify the implementation of each worker, allowing other asyncWorkers to also enjoy the enhanced capabilities of controller-runtime, eliminating the need to maintain the implementation of asyncWorkers, and unify the prometheus monitoring.

Maybe this case should create a new issue to discuss?

XiShanYongYe-Chang:
> Maybe this case should create a new issue to discuss?

+1

Can you help create an issue to track?

pkg/detector/detector.go
XiShanYongYe-Chang: How about not setting this value in the current pr?

XiShanYongYe-Chang (commented):
Other parts lgtm

zach593:
> Can you help create an issue to track?

I just created https://github.com/karmada-io/karmada/issues/6251 to track it.

CharlesQQ:
> How about not setting this value in the current pr?

I remove that, and plan to move it into pr https://github.com/karmada-io/karmada/pull/5813

XiShanYongYe-Chang:
Thanks a lot~
The title of this pr does not seem to be the same as what we actually modified. Can you help me to modify it?

/lgtm

For now, the changes that the pr focuses on are specific enough that I feel can continue to move forward.
cc @zach593 @RainbowMango

pkg/detector/detector.go
```diff
@@ -922,35 +919,12 @@ func (d *ResourceDetector) OnPropagationPolicyUpdate(oldObj, newObj interface{})
 	// The idea of the long-term solution, perhaps PropagationPolicy could have
 	// a status, in that case we can record the observed priority(.status.observedPriority)
 	// which can be used to detect priority changes during reconcile logic.
+	policyObj := newObj.(*policyv1alpha1.PropagationPolicy)
```
zach593, line 922: This line should be moved into the `if features.FeatureGate.Enabled(features.PolicyPreemption)` if block, I don't see it used somewhere outside of the if block.

pkg/detector/detector.go
```diff
@@ -1023,35 +991,11 @@ func (d *ResourceDetector) OnClusterPropagationPolicyUpdate(oldObj, newObj inter
 	// The idea of the long-term solution, perhaps ClusterPropagationPolicy could have
 	// a status, in that case we can record the observed priority(.status.observedPriority)
 	// which can be used to detect priority changes during reconcile logic.
+	policyObj := newObj.(*policyv1alpha1.ClusterPropagationPolicy)
```
zach593, line 994: ditto.

CharlesQQ: done

pkg/detector/detector.go
CharlesQQ: done

pkg/util/fedinformer/keys/keys.go
```diff
@@ -111,6 +111,19 @@ func ClusterWideKeyFunc(obj interface{}) (ClusterWideKey, error) {
...
+	key := ClusterWideKey{}
+
+	metaInfo, err := meta.Accessor(obj)
+	if err != nil { // should not happen
+		return key, fmt.Errorf("object has no meta: %v", err)
+	}
+	key.Namespace = metaInfo.GetNamespace()
+	key.Name = metaInfo.GetName()
+	return key, nil
+}
```
RainbowMango, line 125: This function does not return a standard `ClusterWideKey`, actually a NamespacedKey.

So, shall we introduce a `NamespacedKey` in addition to `FederatedKey` and `ClusterWideKey`? Like:
```golang
type NamespacedKey struct {
	Namespace string
	Name      string
}
```

RainbowMango:
I added a release not based on the test report in PR description:
> Optimization effect: 12000+ workload, running karmada-controller on a 96C cpu machine, when restarting, the cpu peak utilization dropped from 70% to 45%

and https://github.com/karmada-io/karmada/pull/5802#issuecomment-2753304290.

RainbowMango (approved):
/lgtm
/approve

PR #5813, opened 2024-11-13
disable deepcopy for list action

CharlesQQ:
**What type of PR is this?**
/kind feature

When the node CPU usage rate is almost over 90%, it seems that deepcopy cost the most time:

```
I1113 14:36:11.161745      18 overridemanager.go:181] applyNamespacedOverrides, list configmap-beta-moa-demo-prod-681,  cluster member-cluster1, cost: 0.980144234
I1113 14:36:12.445315      18 overridemanager.go:181] applyNamespacedOverrides, list configmap-beta-moa-demo-prod-681,  cluster member-cluster2, cost: 0.811084427
```
image

After turning off list deepcopy, the list time is reduced to 0.1s or even lower.
```
I1113 15:43:39.212238      17 overridemanager.go:182] applyNamespacedOverrides, list beta-moa-demo-prod-283,  cluster member-cluster1, cost: 0.040210384
I1113 15:43:39.251189      17 overridemanager.go:182] applyNamespacedOverrides, list beta-moa-demo-prod-97,  cluster member-cluster2, cost: 0.018000362
I1113 15:43:39.314731      17 overridemanager.go:182] applyNamespacedOverrides, list beta-moa-demo-prod-283,  cluster member-cluster2, cost: 0.023638913
```
image

part of  https://github.com/karmada-io/karmada/issues/5790

```release-note
`Performance`: `karmada-controller-manager`: Significant performance improvements have been achieved by reducing deepcopy operations during list processes:
- ​​Binding Controller​​: Response time reduced by 50%
​​- Dependencies Controller​​: Execution time improved 5× faster 
​​- Detector Controller​​: Processing time cut by 50% 
```

### DIFF
--- a/pkg/controllers/binding/binding_controller.go
+++ b/pkg/controllers/binding/binding_controller.go
@@ -29,6 +29,7 @@ import (
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/klog/v2"
+	"k8s.io/utils/ptr"
 	controllerruntime "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
@@ -189,19 +190,23 @@ func (c *ResourceBindingController) newOverridePolicyFunc() handler.MapFunc {
 			return nil
 		}
 
-		bindingList := &workv1alpha2.ResourceBindingList{}
-		if err := c.Client.List(ctx, bindingList); err != nil {
+		readonlyBindingList := &workv1alpha2.ResourceBindingList{}
+		listOption := &client.ListOptions{
+			UnsafeDisableDeepCopy: ptr.To(true),
+		}
+		if len(namespace) > 0 {
+			listOption = &client.ListOptions{
+				Namespace:             namespace,
+				UnsafeDisableDeepCopy: ptr.To(true),
+			}
+		}
+		if err := c.Client.List(ctx, readonlyBindingList, listOption); err != nil {
 			klog.Errorf("Failed to list resourceBindings, error: %v", err)
 			return nil
 		}
 
 		var requests []reconcile.Request
-		for _, binding := range bindingList.Items {
-			// Skip resourceBinding with different namespace of current overridePolicy.
-			if len(namespace) != 0 && namespace != binding.Namespace {
-				continue
-			}
-
+		for _, binding := range readonlyBindingList.Items {
 			// Nil resourceSelectors means matching all resources.
 			if len(overrideRS) == 0 {
 				klog.V(2).Infof("Enqueue ResourceBinding(%s/%s) as override policy(%s/%s) changes.", binding.Namespace, binding.Name, a.GetNamespace(), a.GetName())
--- a/pkg/dependenciesdistributor/dependencies_distributor.go
+++ b/pkg/dependenciesdistributor/dependencies_distributor.go
@@ -36,6 +36,7 @@ import (
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
+	"k8s.io/utils/ptr"
 	controllerruntime "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
@@ -172,16 +173,18 @@ func (d *DependenciesDistributor) reconcileResourceTemplate(key util.QueueKey) e
 		return fmt.Errorf("invalid key")
 	}
 	klog.V(4).Infof("DependenciesDistributor start to reconcile object: %s", resourceTemplateKey)
-	bindingList := &workv1alpha2.ResourceBindingList{}
-	err := d.Client.List(context.TODO(), bindingList, &client.ListOptions{
-		Namespace:     resourceTemplateKey.Namespace,
-		LabelSelector: labels.Everything()})
+	readonlyBindingList := &workv1alpha2.ResourceBindingList{}
+	err := d.Client.List(context.TODO(), readonlyBindingList, &client.ListOptions{
+		Namespace:             resourceTemplateKey.Namespace,
+		LabelSelector:         labels.Everything(),
+		UnsafeDisableDeepCopy: ptr.To(true),
+	})
 	if err != nil {
 		return err
 	}
 
-	for i := range bindingList.Items {
-		binding := &bindingList.Items[i]
+	for i := range readonlyBindingList.Items {
+		binding := &readonlyBindingList.Items[i]
 		if !binding.DeletionTimestamp.IsZero() {
 			continue
 		}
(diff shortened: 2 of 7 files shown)


zach593:
Regarding this performance optimization, which clearly introduces side effects, I have two suggestions:

1. How much of a performance improvement will it bring? Could you add benchmarks for these calls along with e2e performance comparison?

2. Is it possible to add some mechanisms to mitigate these side effects? For example, adding unit tests or e2e tests so that if someone later introduces code that triggers this side effect (i.e., directly modifying the informer cache), our test suite can detect such behavior.

CharlesQQ:
> How much of a performance improvement will it bring? Could you add benchmarks for these calls along with e2e performance comparison?

Through golang pprof analysis, it is found that during the restart period, deep copy will have obvious CPU time loss. After removing it, this part of the loss disappeared.
Maybe this change is difficult to compare through e2e testing. Do you have any good testing suggestions?

CharlesQQ:
if seems that deepcopy cost the most time:
image

image
image

image

After turning off list deepcopy, the list time is reduced to 0.1s or even lower.
```
I1113 15:43:39.212238      17 overridemanager.go:182] applyNamespacedOverrides, list beta-moa-demo-prod-283,  cluster member-cluster1, cost: 0.040210384
I1113 15:43:39.251189      17 overridemanager.go:182] applyNamespacedOverrides, list beta-moa-demo-prod-97,  cluster member-cluster2, cost: 0.018000362
I1113 15:43:39.314731      17 overridemanager.go:182] applyNamespacedOverrides, list beta-moa-demo-prod-283,  cluster member-cluster2, cost: 0.023638913
```
image

CharlesQQ:
pls take a look at the e2e performance comparison  @zach593

zach593:
> 2\. Is it possible to add some mechanisms to mitigate these side effects? For example, adding unit tests or e2e tests so that if someone later introduces code that triggers this side effect (i.e., directly modifying the informer cache), our test suite can detect such behavior.

There is an env `KUBE_CACHE_MUTATION_DETECTOR`, which may cause panic when the informer cache is illegally modified, I think if we can enable this environment when running e2e tests, I guess it could eliminate the side effects of disable deepcopy.

CharlesQQ:
> Reducing the time taken for an operation by an order of magnitude is great, but I think it would be more helpful to see how much of an overall improvement the change made, especially for an optimization that has potential side effects.


I will test in a real test environment and give optimization data

CharlesQQ:
> here is an env KUBE_CACHE_MUTATION_DETECTOR, which may cause panic when the informer cache is illegally modified, I think if we can enable this environment when running e2e tests, I guess it could eliminate the side effects of disable deepcopy.

KUBE_CACHE_MUTATION_DETECTOR=true take effect in go-client sharedinformer, does it also work in controller-runtime? If so, this environment variable should be added in e2e tests

zach593:
> `KUBE_CACHE_MUTATION_DETECTOR=true` take effect in go-client sharedinformer, does it also work in controller-runtime? 

https://github.com/karmada-io/karmada/blob/0713e4b9be36e4b079f1dc30ac6570d17c9e6451/vendor/sigs.k8s.io/controller-runtime/pkg/cache/internal/informers.go#L59-L61


The answer is yes, it also applies to controller-runtime, because controller-runtime also uses client-go informer.

CharlesQQ:
> A new concern just popped into my mind: does disabling deepCopy consistently lead to data being in a state of race conditions?

I don't understand what this sentence means, can you give me an example?

CharlesQQ:
4000 workload, 8000 rb, 12C cpu:

before optimization, it tooks 9 minutes; after optimization, it takes 4 minutes and 30 seconds

image

image

zach593:
I'm wondering if there will be a data race between the reading thread and the informer writing thread. This is something I suddenly realized, and I didn't think about it very seriously. Since I raised this question, I did some research and the result is **it should not cause a data race**.

https://github.com/karmada-io/karmada/blob/9ea0382293c80fdf8cf13da914136dfc4df3ec51/vendor/k8s.io/client-go/tools/cache/controller.go#L548-L557

From the code above, you can see that when writing to the cache, the entire variable is replaced with the new object from the `deltas` .If the read thread holds an object, when the write thread updates the cache, the object held by the read thread and the cache will no longer have any connection.

In other words, if the reading thread modifies the held object in the `*a.xxx = yyy` way, then this will cause a data race, and will also pollute the cache and objects already held by other reading threads. The object read from the cache later will also be polluted. Pollution will continue until the informer updates the object.

CharlesQQ:
> In other words, if the reading thread modifies the held object in the *a.xxx = yyy way, then this will cause a data race, and will also pollute the cache and objects already held by other reading threads. The object read from the cache later will also be polluted. Pollution will continue until the informer updates the object.

list action disabling deepCopy, must make sure the list item will not be modified, or make a deepcopy of it before modifying it.

CharlesQQ:
detector controller:   takes 2 minutes and 30 seconds
image

after disable deepcopy for list action, it takes 1 minute and 15 seconds
image

RainbowMango (commented):
Disable deepcopy indeed can improve performance, but should be used very cautiously.

/assign

pkg/controllers/binding/binding_controller.go
```diff
@@ -190,7 +191,7 @@ func (c *ResourceBindingController) newOverridePolicyFunc() handler.MapFunc {
 		}
 
 		bindingList := &workv1alpha2.ResourceBindingList{}
-		if err := c.Client.List(ctx, bindingList); err != nil {
+		if err := c.Client.List(ctx, bindingList, &client.ListOptions{UnsafeDisableDeepCopy: ptr.To(true)}); err != nil {
```
RainbowMango, line 194: Does it make sense to add a `namespace` filter? Like:
```diff
                bindingList := &workv1alpha2.ResourceBindingList{}
-               if err := c.Client.List(ctx, bindingList); err != nil {
+               if err := c.Client.List(ctx, bindingList, &client.ListOptions{UnsafeDisableDeepCopy: ptr.To(true), Namespace: namespace}); err != nil {
```

RainbowMango: `bindingList`  --> `readonlyBindingList`?

pkg/dependenciesdistributor/dependencies_distributor.go
```diff
@@ -175,8 +176,10 @@ func (d *DependenciesDistributor) reconcileResourceTemplate(key util.QueueKey) e
 	klog.V(4).Infof("DependenciesDistributor start to reconcile object: %s", resourceTemplateKey)
 	bindingList := &workv1alpha2.ResourceBindingList{}
```
RainbowMango, line 177:
```suggestion
	readonlyBindingList := &workv1alpha2.ResourceBindingList{}
```

RainbowMango (commented):
Generally looks good to me. 
But I think we should rename variable name or add more comments to notice the object can not be mutated.

pkg/controllers/binding/binding_controller.go
CharlesQQ: > Does it make sense to add a namespace filter

Yes

CharlesQQ:
>  But I think we should rename variable name or add more comments to notice the object can not be mutated.

Controller-runtime has a detailed description of `UnsafeDisableDeepCopy`, which is very clear. Is there no need to add it?

https://github.com/karmada-io/karmada/blob/65e7f74c991631a5888f5176342af118f4d5712e/vendor/sigs.k8s.io/controller-runtime/pkg/client/options.go#L495-L499

RainbowMango (approved):
/lgtm
/approve

Issue #5788, opened 2024-11-05
[Feature] Stateful Application Failover Support


RainbowMango:
**Summary**
Karmada’s scheduling logic runs on the assumption that resources that are scheduled and rescheduled are stateless. In some cases, users may desire to conserve a certain state so that applications can resume from where they left off in the previous cluster.

For CRDs dealing with data-processing (such as Flink or Spark), it can be particularly useful to restart applications from a previous checkpoint. That way applications can seamlessly resume processing data while avoiding double processing.

This feature aims to introduce a generalized way for users to define application state preservation in the context of cluster-to-cluster failovers.

**Proposal**
- [ ] state preservation (@Dyex719, #5116)

**Iteration Tasks -- Part-1: Ensure scheduler skips clusters where triggers the failover**
- [x] API change:  Comment use case for [PurgeMode Immediately](https://github.com/karmada-io/karmada/blob/aa61791ccb40fec260d228dc4bfa9c3e34a0fce8/pkg/apis/policy/v1alpha1/propagation_types.go#L267-L269) base on discussion on https://github.com/karmada-io/karmada/pull/5251#issuecomment-2425931916. (@RainbowMango)
- [x] API change: Introduce [PurgeMode ](https://github.com/karmada-io/karmada/blob/aa61791ccb40fec260d228dc4bfa9c3e34a0fce8/pkg/apis/policy/v1alpha1/propagation_types.go#L304-L311) to [GracefulEvictionTask in ResourceBinding](https://github.com/karmada-io/karmada/blob/aa61791ccb40fec260d228dc4bfa9c3e34a0fce8/pkg/apis/work/v1alpha2/binding_types.go#L237-L291). (@mszacillo, [#5816](https://github.com/karmada-io/karmada/pull/5816))
- [x] Make changes to [GracefulEvictCluster()](https://github.com/karmada-io/karmada/blob/aa61791ccb40fec260d228dc4bfa9c3e34a0fce8/pkg/apis/work/v1alpha2/binding_types_helper.go#L130C31-L130C51) to set PurgeMode during eviction process. (@XiShanYongYe-Chang, [#5821](https://github.com/karmada-io/karmada/pull/5821))
[...]

PR #5816, opened 2024-11-13
feat(stateful failover support) Introduce PurgeMode to GracefulEvictionTask in ResourceBinding

mszacillo:
**What type of PR is this?**
/kind api-change

Part of #5788. This small PR adds PurgeMode to the GracefulEvictionTask as a prerequisite for ensuring scheduler skips clusters which triggered the failover.

```release-note
Introduce PurgeMode to GracefulEvictionTask in ResourceBinding
```

### DIFF
--- a/pkg/apis/work/v1alpha2/binding_types.go
+++ b/pkg/apis/work/v1alpha2/binding_types.go
@@ -240,6 +240,13 @@ type GracefulEvictionTask struct {
 	// +required
 	FromCluster string `json:"fromCluster"`
 
+	// PurgeMode represents how to deal with the legacy applications on the
+	// cluster from which the application is migrated.
+	// Valid options are "Immediately", "Graciously" and "Never".
+	// +kubebuilder:validation:Enum=Immediately;Graciously;Never
+	// +optional
+	PurgeMode policyv1alpha1.PurgeMode `json:"purgeMode,omitempty"`
+
 	// Replicas indicates the number of replicas should be evicted.
 	// Should be ignored for resource type that doesn't have replica.
 	// +optional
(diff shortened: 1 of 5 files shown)


RainbowMango (commented):
/assign

pkg/apis/work/v1alpha2/binding_types.go
```diff
@@ -240,6 +240,15 @@ type GracefulEvictionTask struct {
 	// +required
 	FromCluster string `json:"fromCluster"`
 
+	// PurgeMode represents how to deal with the legacy applications on the
+	// cluster from which the application is migrated.
+	// Valid options are "Immediately", "Graciously" and "Never".
+	// Defaults to "Graciously".
+	// +kubebuilder:validation:Enum=Immediately;Graciously;Never
+	// +kubebuilder:default=Graciously
```
RainbowMango, line 248: For application failover, the PurgeMode should be inherited from PropagationPolicy which manages the current workload, but for cluster failover, as I mentioned at #5788, we can set `Graciously` as the default.

So, from the API perspective, it might not have the default preference. 

```suggestion
	// +kubebuilder:validation:Enum=Immediately;Graciously;Never
```

XiShanYongYe-Chang: Does it need to be a required field?

mszacillo: Since we inherit PurgeMode from the PropagationPolicy (which has a default) and will have a default value for cluster failover, I don't believe this field should ever be empty. I can change this to required if we'd like.

XiShanYongYe-Chang: It's not necessarily necessary, it's just a question of mine.

XiShanYongYe-Chang:
LGTM

RainbowMango (approved):
/lgtm
/approve

PR #5818, opened 2024-11-14
fix: failed to sync work when binding's target clusters are empty

vie-serendipity:
**What type of PR is this?**

/kind bug
For now, when binding is not scheduled, which means clusters field is empty, work will still be synced. And this log will be printed.
[Sync work of resourceBinding(%s/%s) successful](https://github.com/karmada-io/karmada/blob/e08cea1e602d0114d1e82e0088e7261a154625d6/pkg/controllers/binding/binding_controller.go#L138)
But actually no work is created. I think it's necessary to check whether clusters field of binding is empty. I'not sure this should be added to predicates of controller or just put it into syncbinding function, like my current pr code. I think the latter is better, in this way, users can see scheduling encounter some problems.

Fixes #

```release-note
NONE
```


XiShanYongYe-Chang:
Hi @vie-serendipity thanks for your feedback. We can really decide whether to reconcile the ResourceBinding based on whether the scheduling has been completed (the ClusterResourceBinding has the same processing logic), which can save some unnecessary tuning. Therefore, I understand that this should be an optimization, not a bug fix.

In addition, we can judge whether the scheduling is completed in this way: `obj.Status.SchedulerObservedGeneration == obj.Generation`, this can cover the situation where there is no scheduling result.

vie-serendipity:
> We can really decide whether to reconcile the ResourceBinding based on whether the scheduling has been completed (the ClusterResourceBinding has the same processing logic), which can save some unnecessary tuning.

For now, I find the logic to reconcile depends on `WithEventFilter(predicate.GenerationChangedPredicate{})`. I didn't find any other logic to prevent reconciliation (if any, please remind me). But in my case, my scheduler crashed, and after the binding was updated, I directly saw [Sync work of resourceBinding(%s/%s) successful](https://github.com/karmada-io/karmada/blob/e08cea1e602d0114d1e82e0088e7261a154625d6/pkg/controllers/binding/binding_controller.go#L138) in controller manager's log. So I think this is unexpected.

We can add `obj.Status.SchedulerObservedGeneration == obj.Generation` to predicates, not only generationchanged?

XiShanYongYe-Chang:
> For now, I find the logic to reconcile depends on WithEventFilter(predicate.GenerationChangedPredicate{}). I didn't find any other logic to prevent reconciliation (if any, please remind me). 

You are right.

> But in my case, my scheduler crashed

Does this mean the component crashed and restarted?

> I directly saw [Sync work of resourceBinding(%s/%s) successful](https://github.com/karmada-io/karmada/blob/e08cea1e602d0114d1e82e0088e7261a154625d6/pkg/controllers/binding/binding_controller.go#L138) in controller manager's log.

I think this operation has no negative impact, am I right?

> We can add obj.Status.SchedulerObservedGeneration == obj.Generation to predicates, not only generationchanged?

According to the information you describe, I think so, we can save unnecessary reconcile.

vie-serendipity:
> Does this mean the component crashed and restarted?

The scheduler failed to pull the image, so it doesn't do scheduling.

> I think this operation has no negative impact, am I right?

Yes, I think so. I will update code in my pr.

XiShanYongYe-Chang:
> The scheduler failed to pull the image, so it doesn't do scheduling.

Now, I want to make sure that if the image is correctly pulled, can you find the log you're referring to?

vie-serendipity:
Actually, I see this log in controller manager's log. Because my scheduler is not ready, I think this log is unexpected. And I can't find log because I am testing and the environment changes significantly.
[Sync work of resourceBinding(%s/%s) successful](https://github.com/karmada-io/karmada/blob/e08cea1e602d0114d1e82e0088e7261a154625d6/pkg/controllers/binding/binding_controller.go#L138)

XiShanYongYe-Chang:
> Actually, I see this log in controller manager's log. Because my scheduler is not ready, I think this log is unexpected. And I can't find log because I am testing and the environment changes significantly. [Sync work of resourceBinding(%s/%s) successful](https://github.com/karmada-io/karmada/blob/e08cea1e602d0114d1e82e0088e7261a154625d6/pkg/controllers/binding/binding_controller.go#L138)

You can check whether the log exists after adjusting the environment.

vie-serendipity:
The controller manager pod restarts, so the log doesn't exist.

PR #5824, opened 2024-11-16
pkg/search/backendstore: unit test opensearch

mohamedawnallah:
**Description**

In this commit, we unit test opensearch backend store for karmada search aggregated apiserver as cache registry on creating new instance and getting/creating an index.

**What type of PR is this?**

Part of #5491.

```release-note
NONE
```


mohamedawnallah:
It looks like the in-memory and OpenSearch backend stores haven't been tested yet. It might be beneficial to add some tests for these.

I've also rebased onto the latest master and resolved all conflicts. Any feedback on this PR would be greatly appreciated!

cc: @XiShanYongYe-Chang, @zhzhuang-zju

mohamedawnallah:
It looks like the in-memory and OpenSearch backend stores haven't been tested yet. It might be beneficial to add some tests for these in the codebase.

I've also rebased onto the latest master and resolved all conflicts. Any feedback on this PR would be greatly appreciated!

cc: @XiShanYongYe-Chang, @zhzhuang-zju

XiShanYongYe-Chang:
Thanks @mohamedawnallah 
/assign

pkg/search/backendstore/opensearch.go
```diff
@@ -257,29 +260,33 @@ func (os *OpenSearch) indexName(us *unstructured.Unstructured) (string, error) {
...
+		return os.handleIndexAlreadyExists(resp.String(), name)
 	}
 
-	klog.Infof("create index response: %s", resp.String())
-	os.indices[name] = struct{}{}
+	if createIndexIfNotExist {
+		klog.Infof("create index response: %s", resp.String())
+		os.indices[name] = struct{}{}
+		return name, nil
+	}
```
XiShanYongYe-Chang, line 280: When createIndexIfNotExist is set to false, the original logic has been modified. Is it inappropriate to modify the original logic in the current PR?

pkg/search/backendstore/opensearch_test.go
```diff
@@ -0,0 +1,457 @@
+/*
+Copyright 2024 The Karmada Authors.
```
zhzhuang-zju, line 2:
```suggestion
Copyright 2025 The Karmada Authors.
```

pkg/search/backendstore/opensearch.go
zhzhuang-zju: The original use of `openseach.delete` with `os.indexName` might not be quite appropriate either. If the index doesn't exist, it could result in the index being created first and then deleted.

zhzhuang-zju: The responsibility of the `indexName` function should be to create the index by default if it doesn't exist, so adding a parameter `createIndexIfNotExist` might not be reasonable.

mohamedawnallah: Thanks for the feedback. Let me take a closer look at that code change later today since that PR has been for a while :)

mohamedawnallah: > Thanks for the feedback. Let me take a closer look at that code change later today since that PR has been for a while :)

This comment might've been unclear. I meant that I was missing some context about the purpose of that code change, since it's been a while.

mohamedawnallah: > The original use of openseach.delete with os.indexName might not be quite appropriate either. If the index doesn't exist, it could result in the index being created first and then deleted.

Yes I believe this was my rational back then.

> The responsibility of the indexName function should be to create the index by default if it doesn't exist, so adding a parameter createIndexIfNotExist might not be reasonable.

That's right.

> When createIndexIfNotExist is set to false, the original logic has been modified. Is it inappropriate to modify the original logic in the current PR?

I've created that PR #6320 that solves that issue of `delete` of creating index if not index before deleting it and also some refactoring. Let's move our discussion regards that proposed code change there?

This PR now depends on that PR #6320.

mohamedawnallah: > The original use of openseach.delete with os.indexName might not be quite appropriate either. If the index doesn't exist, it could result in the index being created first and then deleted.

Yes I believe this was my rational back then behind that proposed code change.

> The responsibility of the indexName function should be to create the index by default if it doesn't exist, so adding a parameter createIndexIfNotExist might not be reasonable.

That's right.

> When createIndexIfNotExist is set to false, the original logic has been modified. Is it inappropriate to modify the original logic in the current PR?

I've created that PR #6320 that solves that issue of `delete` of creating index if not index before deleting it and also some refactoring. Let's move our discussion regards that proposed code change there?

This PR now depends on that PR #6320.

mohamedawnallah: > The original use of openseach.delete with os.indexName might not be quite appropriate either. If the index doesn't exist, it could result in the index being created first and then deleted.

Yes I believe this was my rational back then behind that proposed code change.

> The responsibility of the indexName function should be to create the index by default if it doesn't exist, so adding a parameter createIndexIfNotExist might not be reasonable.

That's right.

> When createIndexIfNotExist is set to false, the original logic has been modified. Is it inappropriate to modify the original logic in the current PR?

That's right as well. I've created that PR #6320 that solves that issue of `delete` of creating index if not index before deleting it and also some refactoring. Let's move our discussion regards that proposed code change there?

This PR now depends on that PR #6320.

pkg/search/controllers_test.go
```diff
@@ -359,27 +366,69 @@ func TestAddResourceRegistryEventHandler(t *testing.T) {
...
+
+				if err := upsertResourceRegistry(clientConnector, resourceSelectors, backendStoreCfg, registryName, resourceVersion, []string{clusterName}); err != nil {
 					return err
 				}
 				if err := cacheNextWrapper(controller); err != nil {
 					return err
 				}
 
+				// TODO: Why creating that test-deployment object in the control plane client for the resource registry
+				// doesn't trigger the upsert resource registry event handler in OpenSearch?
```
mohamedawnallah, line 430: Still can’t figure out how to trigger the `AddFunc` event handler, even after creating `test-deployment`?https://github.com/karmada-io/karmada/blob/c160708fa151787fe545dec012e0695bfd44f2c6/pkg/search/backendstore/opensearch.go#L141-L145

PR #5829, opened 2024-11-17
pkg/karmadactl: unit test cordon

mohamedawnallah:
**Description**

In this commit, we unit test cordon on karmadactl client making sure the cordon/uncordon working as expected on a given cluster.

**What type of PR is this?**

Part of #5491.

```release-note
NONE
```

### DIFF
--- a/pkg/karmadactl/cordon/cordon_test.go
+++ b/pkg/karmadactl/cordon/cordon_test.go
@@ -1,0 +1,268 @@
+/*
+Copyright 2024 The Karmada Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cordon
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"os"
+	"strings"
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+
+	clusterv1alpha1 "github.com/karmada-io/karmada/pkg/apis/cluster/v1alpha1"
+	karmadaclientset "github.com/karmada-io/karmada/pkg/generated/clientset/versioned"
+	fakekarmadaclient "github.com/karmada-io/karmada/pkg/generated/clientset/versioned/fake"
+	"github.com/karmada-io/karmada/pkg/karmadactl/util"
+)
+
+type testFactory struct {
+	cmdutil.Factory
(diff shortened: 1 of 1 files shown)


pkg/karmadactl/cordon/cordon_test.go
```diff
@@ -0,0 +1,263 @@
...
+var fTCordonCondition = func(cluster *clusterv1alpha1.Cluster) error {
+	if len(cluster.Spec.Taints) == 0 {
+		return fmt.Errorf("expected one noschedule taint for cluster %s, but got empty taints", cluster.GetName())
+	}
+	taintExpected := corev1.Taint{
+		Key:    clusterv1alpha1.TaintClusterUnscheduler,
+		Effect: corev1.TaintEffectNoSchedule,
+	}
+	if !reflect.DeepEqual(cluster.Spec.Taints[0], taintExpected) {
+		return fmt.Errorf("expected taint %v, but got %v", taintExpected, cluster.Spec.Taints[0])
```
XiShanYongYe-Chang, line 62: It is better to determine whether there are such taint points in the cluster status. There may be multiple taint points in the cluster. The first one may not be the target taint point.

pkg/karmadactl/cordon/cordon_test.go
```diff
@@ -0,0 +1,263 @@
...
+		opts                *CommandCordonOption
+		prep                func(f util.Factory, opts *CommandCordonOption, desiredCordonStatus int) error
+		verify              func(util.Factory) error
+		wantErr             bool
+		logMsg              string
+	}{
+		{
+			name:                "RunCordonOrUncordon_CordonUncordonedCluster_ClusterCordoned",
+			desiredCordonStatus: DesiredCordon,
+			f:                   testFactory{client: fakekarmadaclient.NewSimpleClientset()},
```
XiShanYongYe-Chang, line 89: How about using `NewClientset` to replace `NewSimpleClientset`?

pkg/karmadactl/cordon/cordon_test.go
```diff
@@ -0,0 +1,263 @@
...
+	return t.client, nil
+}
+
+func (t testFactory) FactoryForMemberCluster(string) (cmdutil.Factory, error) {
+	panic("not implemented")
+}
+
+type fTaintCondition func(cluster *clusterv1alpha1.Cluster) error
+
+var fTCordonCondition = func(cluster *clusterv1alpha1.Cluster) error {
```
XiShanYongYe-Chang, line 53: `fTCordonCondition`  The name of this variable is a bit difficult to understand. Same as `fTaintCondition`.

pkg/karmadactl/cordon/cordon_test.go
mohamedawnallah: Good point! Updated it. Thanks! 🙏

pkg/karmadactl/cordon/cordon_test.go
XiShanYongYe-Chang: You're right. I've just noticed that in this sense, direct revisions are not necessarily effective. You can try them.

mohamedawnallah: I found using `NewClientset` is not feasible since it is part of `kubernetes/fake` package and `util.Factory` accepts client of type `karmadaclientset.Interface` which part of `karmada/pkg/generated/clientset/versioned` package.

XiShanYongYe-Chang (commented):
Thanks~
/lgtm
/approve

PR #5794, opened 2024-11-08
Proposal: Support Custom CA Certificate for Karmada Instance in Operator

jabellard:
**What type of PR is this?**

/kind design

**What this PR does / why we need it**:
- Details are contained in the proposal.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #5842, opened 2024-11-20
Support Custom API Server CA Certificate for Karmada Instance in Operator

### LINKED ISSUE #5784: Add Support for Custom Certificates in Karmada Operator
Today, when creating a `Karmada` instance to be managed by the operator, it's not possible to use custom certificates for the instance. This feature request is for providing the ability to use custom certificates when creating a `Karmada` instance.
For reference, the helm chart already has [support](https://github.com/karmada-io/karmada/blob/release-1.11/charts/karmada/values.yaml#L127) for this feature. 

**Why is this needed**:
At Bloomberg, we're currently building a managed Karmada platform and want to use the Karmada operator to manage the entire lifecycle of managed `Karmada` instances. To fulfill some of our use cases, we would like to have the ability of specifying custom certificates when creating a `Karmada` instance.

jabellard:
**What type of PR is this?**

/kind feature

This is an implementation of this [proposal](https://github.com/karmada-io/karmada/pull/5794) to add support for custom API server CA certificates in the Karmada operator.

Fixes #5784

```release-note
The new `CustomCertificate.ApiServerCACert` field added as part of the Karmada spec can be used to specify the reference to a secret that contains a custom CA certificate for the Karmada API Server.
```


RainbowMango:
/retest
the failing unit test is not related but an known issue.

operator/pkg/certs/certs.go
```diff
@@ -212,25 +212,25 @@ func KarmadaCertEtcdClient() *CertConfig {
 // KarmadaCert is karmada certificate, it includes certificate basic message.
 // we can directly get the byte array of certificate key and cert from the object.
 type KarmadaCert struct {
-	pairName string
-	caName   string
-	cert     []byte
-	key      []byte
+	PairName string
```
zhzhuang-zju, line 215: Can we avoid exposing internal fields by adding a new function to generate `KarmadaCert`, thereby minimizing the scope of changes?

operator/pkg/init.go
```diff
@@ -42,13 +42,14 @@ var (
...
-	KarmadaVersion string
-	CRDTarball     operatorv1alpha1.CRDTarball
-	KarmadaDataDir string
-	Karmada        *operatorv1alpha1.Karmada
+	Name                    string
+	Namespace               string
+	Kubeconfig              *rest.Config
+	KarmadaVersion          string
+	CRDTarball              operatorv1alpha1.CRDTarball
+	CustomCertificateConfig operatorv1alpha1.CustomCertificate
```
zhzhuang-zju, line 50: We can consider not adding a new field `CustomCertificateConfig` since `Karmada` already contains the configuration for `customcertificates`.

operator/pkg/tasks/init/cert.go
```diff
@@ -101,6 +104,31 @@ func runCATask(kc *certs.CertConfig) func(d workflow.RunData) error {
...
+				PairName: kc.Name,
+				CAName:   kc.CAName,
+				Cert:     certData,
+				Key:      keyData,
+			}
+
+			data.AddCert(customKarmadaCert)
+			klog.V(2).InfoS("[certs] Successfully added custom CA certificate to cert store", "certName", kc.Name)
+			return nil
+		}
```
zhzhuang-zju, line 130: How about extracting two functions: one to determine whether it is a custom CA, and another to load the `karmadaCert` using `secretRef`. This may enhance the extensibility of the code.

operator/pkg/certs/certs.go
RainbowMango: @zhzhuang-zju So, what's your suggestion?
Do you mean to introduce some `set` methods?

zhzhuang-zju: How about 
```bash
func NewKarmadaCert(pairName,caName string , cert, key []byte) *KarmadaCert{
	return &KarmadaCert{
		pairName: pairName,
		caName: caName,
		cert: cert,
		key: key,
	}
}
```

jabellard: @zhzhuang-zju , thanks for the suggestion.

operator/pkg/init.go
jabellard: That's a good point. Thanks for bringing it up. I considered that, but if you take a closer look, the root `Karmada` object is not part of this type. It's possible to add it and get rid of a lot of fields here, but that would take a refactor with very wide scope. I think it's fine for now, as this object is meant to carry the only the info needed in downstream functions.

operator/pkg/tasks/init/cert.go
jabellard: I think this is fine. It's  a one liner that's very specific to the API Server CA. It's not general.

jabellard:
@zhzhuang-zju , thanks for comments and suggestions. Left some notes and just pushed some changes.

operator/pkg/init.go
zhzhuang-zju: > the root Karmada object is not part of this type

Sorry, I didn't get it. Could you please explain it more?

operator/pkg/tasks/init/cert.go
zhzhuang-zju: > I think this is fine. It's a one liner that's very specific to the API Server CA. It's not general.

Get it~ 
Could we abstract the process of loading the `karmadaCert` by `secretRef` into a function? This way, if there's a need for custom other certificates in the future, this function could be used universally.

operator/pkg/init.go
jabellard: We can add the entire `Karmada` object to this field and get rid of some of the other fields, but that will introduce a refactor of a very wide scope with a lot of changes to the code. Can we avoid doing that for now?

operator/pkg/tasks/init/cert.go
jabellard: Ok. Will make the change.

jabellard: Not sure what you mean.  That's actually already done in the `loadCACertFromSecret` function.

zhzhuang-zju: I mean L111-L125 can be extracted into a function. However, this is not important either; we can consider it when there is a real need.

operator/pkg/init.go
jabellard: Ok. Now I see where the confusion is coming from. I could have it explained it a bit better, so it's my fault. This is the [struct](https://github.com/karmada-io/karmada/blob/master/operator/pkg/init.go#L75) I meant to refer to it. It's what's used to carry information to the downstream functions, not `InitOptions`. Notice how the root `Karmada` type is not in there.

operator/pkg/tasks/init/cert.go
jabellard: Yeah. I agree with you on not abstracting the logic prematurely.

operator/pkg/init.go
jabellard: That struct is then abstracted behind this [interface](https://github.com/karmada-io/karmada/blob/master/operator/pkg/tasks/init/data.go#L28) which is what is used directly in those functions.

jabellard: @zhzhuang-zju , just let me know if have any other comments/questions.

zhzhuang-zju: hah, Thank you for the clarification. I understand that `CustomCertificateConfig` is required for `initdata`, but is it possible to obtain it through `initoptions.Karmada.Spec.CustomCertificate` instead of adding it directly to `initoptions`? I actually don't want `initoptions` to become too large.

zhzhuang-zju:
thanks, others LGTM

operator/pkg/init.go
RainbowMango: > We can add the entire Karmada object to this field and get rid of some of the other fields, but that will introduce a refactor of a very wide scope with a lot of changes to the code. Can we avoid doing that for now?

Yeah, I agree that.

RainbowMango (approved):
/lgtm
/approve

RainbowMango:
Oh... I meant to squash the commits...

PR #5847, opened 2024-11-20
pkg/karmadactl: unit test describe

mohamedawnallah:
**Description**

In this commit, we unit test describe command on karmadactl by creating a fake rest client builder and make sure the described k8s object returned successfully when the command submitted.

**What type of PR is this?**

Part of #5491.

```release-note
NONE
```

### DIFF
--- a/pkg/karmadactl/describe/describe_test.go
+++ b/pkg/karmadactl/describe/describe_test.go
@@ -1,0 +1,110 @@
+/*
+Copyright 2024 The Karmada Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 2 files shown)


pkg/karmadactl/describe/describe_test.go
```diff
@@ -0,0 +1,243 @@
...
+				t.Errorf("unexpected error, got: %v", err)
+			}
+			if err != nil && test.wantErr && !strings.Contains(err.Error(), test.errMsg) {
+				t.Errorf("expected error message %s to be in %s", test.errMsg, err.Error())
+			}
+		})
+	}
+}
+
+func TestRunDescribe(t *testing.T) {
```
XiShanYongYe-Chang, line 123: This test is too complex and the method being tested is not an easy function to test. Can we delete this test case?

XiShanYongYe-Chang (commented):
Thanks~
/lgtm
/approve

PR #5861, opened 2024-11-22
Enable the deprecation check which was disabled for version upgrade

seanlaii:
**What type of PR is this?**
/kind deprecation

Re-enable the deprecation check which was disabled for the version upgrade.

Part of #5796 

```release-note
Replace `PostStartHookContext.StopCh` with `PostStartHookContext.Done` since `StopCh` is deprecated.
Replace `Run(ctx.Done())` with `RunWithContext`.
```

### DIFF
--- a/pkg/util/grpcconnection/config.go
+++ b/pkg/util/grpcconnection/config.go
@@ -103,6 +103,8 @@ func (s *ServerConfig) NewServer() (*grpc.Server, error) {
 // DialWithTimeOut will attempt to create a client connection based on the given targets, one at a time, until a client connection is successfully established.
 func (c *ClientConfig) DialWithTimeOut(paths []string, timeout time.Duration) (*grpc.ClientConn, error) {
 	opts := []grpc.DialOption{
+		// grpc.WithBlock is deprecated. TODO: Perhaps need to reconsider the approach in a future PR
+		//nolint:staticcheck
 		grpc.WithBlock(),
 	}
 
@@ -155,6 +157,8 @@ func createGRPCConnection(path string, timeout time.Duration, opts ...grpc.DialO
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
 	defer cancel()
 
+	// grpc.DialContext is deprecated. TODO: Perhaps need to reconsider the approach in a future PR
+	//nolint:staticcheck
 	cc, err := grpc.DialContext(ctx, path, opts...)
 	if err != nil {
 		return nil, fmt.Errorf("dial %s error: %v", path, err)
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -82,10 +82,6 @@ linters-settings:
       - name: unexported-return
       - name: time-naming
       - name: empty-block
-  staticcheck:
-    # TODO(@RainbowMango): Disable deprecation check temporary for Kubernetes dependency update.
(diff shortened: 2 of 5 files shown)


pkg/util/grpcconnection/config.go
```diff
@@ -151,14 +147,10 @@ func (c *ClientConfig) DialWithTimeOut(paths []string, timeout time.Duration) (*
 	return nil, utilerrors.NewAggregate(allErrs)
 }
 
-func createGRPCConnection(path string, timeout time.Duration, opts ...grpc.DialOption) (conn *grpc.ClientConn, err error) {
-	ctx, cancel := context.WithTimeout(context.Background(), timeout)
-	defer cancel()
-
-	cc, err := grpc.DialContext(ctx, path, opts...)
```
seanlaii, line 158: Hi @RainbowMango , 

According to the documentation from [go-grpc](https://github.com/grpc/grpc-go/blob/master/Documentation/anti-patterns.md), the `grpc.WithBloack()` and `grpc.DialContext` are deprecated, and it seems that using the timeout functionality in the manner we have might not be recommended. Instead, they suggest creating clients using `NewClient` rather than the deprecated Dial method, which aligns with best practices for managing client connections.

Additionally, I saw a relevant [PR](https://github.com/open-telemetry/opentelemetry-go/pull/5151) in the OpenTelemetry project, which also moved away from the timeout feature in favor of the new client creation method. Based on these findings, I am considering removing the timeout functionality from our implementation.

What do you think about this? 

Thank you for your guidance and support!

RainbowMango: cc the author @zhzhuang-zju to take a look

zhzhuang-zju: @seanlaii Thank you for pointing out this issue. It can be confirmed that using `grpc.Dial` to create `ClientConn` is not the best practice. I will review the documentation you provided in detail and use it as a basis to review the current modifications.

zhzhuang-zju: Compared to other modifications, the changes to the gRPC module will involve functionality. It might be better to submit them as a separate PR, which would allow for a more focused review. WDYT?

seanlaii: Sounds good. I disable the syntax check for `grpc.DialContext` and `grpc.WithBlock` with some comments. We can discuss it in a separate PR. Thank you!

RainbowMango (approved):
/lgtm
/approve

RainbowMango:
> Sounds good. I disable the syntax check for grpc.DialContext and grpc.WithBlock with some comments. We can discuss it in a separate PR. Thank you!

Cool, I added a follow-up to the #5796

pkg/util/grpcconnection/config.go
RainbowMango: @zhzhuang-zju 
Any update on this issue?

zhzhuang-zju: > @zhzhuang-zju
Any update on this issue?

Not yet. @seanlaii Are you going to work on this issue?

seanlaii: Sorry for the late reply. Yes, I will work on this.

zhzhuang-zju: Thanks for your reply, feel free to ask if you have any question

PR #5865, opened 2024-11-23
let rb status controller onCreate predicate return true

zach593:
**What type of PR is this?**

/kind bug

Currently, the rb/crb status controller watchs rb/crb and work, but the predicates for creating events are all false, which means that if the work status is up to date, but the queue of the rb/crb status controller is not cleared, and it crashes, these unprocessed items will not be requeued after the next restart, and the status of rb/crb and workload will not be updated, until the relevant member cluster objects update the status again, triggering the work status controller and rb/crb status controller to process again.

Fixes #

```release-note
`karmada-controller-manager`: Fixed the corner case where the reconciliation of aggregating status might be missed in case of component restart.
```

### DIFF
--- a/pkg/controllers/status/common.go
+++ b/pkg/controllers/status/common.go
@@ -43,7 +43,13 @@ import (
 )
 
 var bindingPredicateFn = builder.WithPredicates(predicate.Funcs{
-	CreateFunc: func(event.CreateEvent) bool { return false },
+	CreateFunc: func(event.CreateEvent) bool {
+		// Although we don't need to process the ResourceBinding immediately upon its creation,
+		// but it's necessary to ensure that all existing ResourceBindings are processed
+		// uniformly once when the component restarts.
+		// This guarantees that no ResourceBinding is missed after a controller restart.
+		return true
+	},
 	UpdateFunc: func(e event.UpdateEvent) bool {
 		var oldResourceVersion, newResourceVersion string
 


RainbowMango:
> Currently, the rb/crb status controller watchs rb/crb and work, but the predicates for creating events are all false, which means that if the work status is up to date, but the queue of the rb/crb status controller is not cleared, and it crashes, these unprocessed items will not be requeued after the next restart, and the status of rb/crb and workload will not be updated

Another approach is letting the `resource-binding-status-controller` respond to the Work creation event. See the code here:
https://github.com/karmada-io/karmada/blob/478b8c6456856d85401b332e2d56e1ccde325bd1/pkg/controllers/status/common.go#L80

Which approach do you prefer? @zach593 @chaunceyjiang @XiShanYongYe-Chang

XiShanYongYe-Chang:
> letting the resource-binding-status-controller respond to the Work creation event.

I agree with this approach.

XiShanYongYe-Chang:
>  from this point of view, introducing more interdependencies between objects is unnecessary.

Hi @zach593, Are you referring to the concern that duplicate rb/crbs are being processed?

zach593:
> Are you referring to the concern that duplicate rb/crbs are being processed?

No, unually extra reconciliation should not be harmful, unless it runs into some performance issues.

> I understand that we have increased attention to the create event, the only reason is to worry about controller restart, am I right?

Yes, this PR is mainly focused on controller restarts

@XiShanYongYe-Chang

zach593:
> Do you think this patch would have an impact on performance?

After https://github.com/karmada-io/karmada/pull/5779, I think it should not.

pkg/controllers/status/common.go
```diff
@@ -43,7 +43,7 @@ import (
 )
 
 var bindingPredicateFn = builder.WithPredicates(predicate.Funcs{
-	CreateFunc: func(event.CreateEvent) bool { return false },
+	CreateFunc: func(event.CreateEvent) bool { return true },
```
RainbowMango, line 46:
```suggestion
		CreateFunc: func(event.CreateEvent) bool {
		// Although we don't need to process the ResourceBinding immediately upon its creation,
		// but it's necessary to ensure that all existing ResourceBindings are processed
		// uniformly once when the component restarts.
		// This guarantees that no ResourceBinding is missed after a controller restart.
		return true
	},
```

No matter which approach we select, it would be better to have some comments on it.

zach593:
Our goal is to add a self-check mechanism to the status of RB. From a design perspective, when faced with multiple options, if one option can reduce dependencies on other resources, it helps clarify the boundaries between subsystems, achieving what is known as "low in coupling and high in cohesion." 
From this perspective, the first option is preferable. Additionally, from a practical impact perspective, this approach also introduces fewer reconciliation operations, which is even better. @XiShanYongYe-Chang

XiShanYongYe-Chang (commented):
Thanks, I agree.
/lgtm

XiShanYongYe-Chang:
Do we need a release note and cherry-pick it to the previous branch?

RainbowMango:
> Do we need a release note and cherry-pick it to the previous branch?

No strong opinion from me. @zach593 what do you say?

I added a release notes, by the way.

zach593:
cool, lgtm

RainbowMango (approved):
/approve

@XiShanYongYe-Chang Do you think cherry-pick is needed?

XiShanYongYe-Chang:
> Do you think cherry-pick is needed?

According to cherry-pick, the current pr meets the standard, let me build an issue to track it.

Issue #5843, opened 2024-11-20
After deleting the cluster, also need to delete related metrics label.


CharlesQQ:
**What happened**:
After member cluster is deleted from karmada,  the metrics  label  `cluster_name` for this cluster  not deleted  at the same

**What you expected to happen**:
when cluster deleted,  remove deleted this cluster_name label value

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Karmada version:
- kubectl-karmada or karmadactl version (the result of `kubectl-karmada version` or `karmadactl version`):
- Others:

RainbowMango:
Which metric you are talking about?
Similar to #5836?

CharlesQQ:
> Which metric you are talking about?
I find some metrics:
- cluster_ready_state
- cluster_node_number
- cluster_pod_allocated_number

LavredisG:
Same issue as https://github.com/karmada-io/karmada/issues/5839.

XiShanYongYe-Chang:
In favor of #5866 
/assign @CharlesQQ

PR #5866, opened 2024-11-23
fix(cluster):  remove related metrics label when cluster is deleted

CharlesQQ:
**What type of PR is this?**
/kind bug

Fixes https://github.com/karmada-io/karmada/issues/5843

```release-note
`Instrumentation`: The cluster status-related metrics, emitting from `karmada-controller-manager`, will be cleaned up after the cluster is removed.
```

### DIFF
--- a/pkg/metrics/cluster.go
+++ b/pkg/metrics/cluster.go
@@ -135,6 +135,20 @@ func RecordClusterSyncStatusDuration(cluster *v1alpha1.Cluster, startTime time.T
 	clusterSyncStatusDuration.WithLabelValues(cluster.Name).Observe(utilmetrics.DurationInSeconds(startTime))
 }
 
+// CleanupMetricsForCluster removes the cluster status metrics after the cluster is deleted.
+func CleanupMetricsForCluster(clusterName string) {
+	clusterReadyGauge.DeleteLabelValues(clusterName)
+	clusterTotalNodeNumberGauge.DeleteLabelValues(clusterName)
+	clusterReadyNodeNumberGauge.DeleteLabelValues(clusterName)
+	clusterMemoryAllocatableGauge.DeleteLabelValues(clusterName)
+	clusterCPUAllocatableGauge.DeleteLabelValues(clusterName)
+	clusterPodAllocatableGauge.DeleteLabelValues(clusterName)
+	clusterMemoryAllocatedGauge.DeleteLabelValues(clusterName)
+	clusterCPUAllocatedGauge.DeleteLabelValues(clusterName)
+	clusterPodAllocatedGauge.DeleteLabelValues(clusterName)
+	clusterSyncStatusDuration.DeleteLabelValues(clusterName)
+}
+
 // ClusterCollectors returns the collectors about clusters.
 func ClusterCollectors() []prometheus.Collector {
 	return []prometheus.Collector{
--- a/pkg/controllers/status/cluster_status_controller.go
+++ b/pkg/controllers/status/cluster_status_controller.go
@@ -135,6 +135,7 @@ func (c *ClusterStatusController) Reconcile(ctx context.Context, req controllerr
 			c.GenericInformerManager.Stop(req.NamespacedName.Name)
 			c.TypedInformerManager.Stop(req.NamespacedName.Name)
 			c.clusterConditionCache.delete(req.NamespacedName.Name)
+			metrics.CleanupMetricsForCluster(req.NamespacedName.Name)
 
 			// stop lease controller after the cluster is gone.
 			// only used for clusters in Pull mode because no need to set up lease syncing for Push clusters.


pkg/controllers/cluster/cluster_controller.go
```diff
@@ -240,6 +241,9 @@ func (c *Controller) syncCluster(ctx context.Context, cluster *clusterv1alpha1.C
 
 func (c *Controller) removeCluster(ctx context.Context, cluster *clusterv1alpha1.Cluster) (controllerruntime.Result, error) {
 	// add terminating taint before cluster is deleted
+	defer func() {
+		metrics.RemoveClusterStatus(cluster)
+	}()
```
RainbowMango, line 246: What if the removeCluster raises an error? Do we still clean up the metrics?

RainbowMango (commented):
Have you tested it? Can you share the metrics before and after cluster is deleted?

pkg/controllers/cluster/cluster_controller.go
CharlesQQ: > What if the removeCluster raises an error? Do we still clean up the metrics?

shouldn't clean up the metrics, I'll fix it!

CharlesQQ: > Have you tested it? Can you share the metrics before and after cluster is deleted?
I join a cluster named `host`
Before metrics:  
```
cluster_ready_node_number{cluster_name="host"} 0
cluster_ready_state{cluster_name="host"} 1
cluster_sync_status_duration_seconds_bucket{cluster_name="host",le="0.005"} 0
cluster_sync_status_duration_seconds_bucket{cluster_name="host",le="0.01"} 0
cluster_sync_status_duration_seconds_bucket{cluster_name="host",le="0.025"} 0
cluster_sync_status_duration_seconds_bucket{cluster_name="host",le="0.05"} 1
cluster_sync_status_duration_seconds_bucket{cluster_name="host",le="0.1"} 3
cluster_sync_status_duration_seconds_bucket{cluster_name="host",le="0.25"} 3
cluster_sync_status_duration_seconds_bucket{cluster_name="host",le="0.5"} 4
cluster_sync_status_duration_seconds_bucket{cluster_name="host",le="1"} 4
cluster_sync_status_duration_seconds_bucket{cluster_name="host",le="2.5"} 4
cluster_sync_status_duration_seconds_bucket{cluster_name="host",le="5"} 4
cluster_sync_status_duration_seconds_bucket{cluster_name="host",le="10"} 4
cluster_sync_status_duration_seconds_bucket{cluster_name="host",le="+Inf"} 4
cluster_sync_status_duration_seconds_sum{cluster_name="host"} 0.667916613
cluster_sync_status_duration_seconds_count{cluster_name="host"} 4
```

XiShanYongYe-Chang (commented):
Thanks~
It tests with expection. And when I rejoin the cluster, the related cluster metrics occur as while.

XiShanYongYe-Chang (commented):
/lgtm
Can you help add the release note?

RainbowMango (commented):
/assign

pkg/controllers/cluster/cluster_controller.go
```diff
@@ -238,8 +239,13 @@ func (c *Controller) syncCluster(ctx context.Context, cluster *clusterv1alpha1.C
 	return c.ensureFinalizer(ctx, cluster)
 }
 
-func (c *Controller) removeCluster(ctx context.Context, cluster *clusterv1alpha1.Cluster) (controllerruntime.Result, error) {
+func (c *Controller) removeCluster(ctx context.Context, cluster *clusterv1alpha1.Cluster) (ret controllerruntime.Result, retErr error) {
 	// add terminating taint before cluster is deleted
+	defer func() {
+		if retErr == nil {
```
RainbowMango, line 245: The `retErr == nil` can not ensure the `Cluster` has been successfully deleted.
It's possible that the cluster is still waiting for the next retry attempt, see https://github.com/karmada-io/karmada/blob/b605d9d6371885cbf0d850eb734ae3bedb98d576/pkg/controllers/cluster/cluster_controller.go#L260-L261

RainbowMango (changes requested):
I think the metrics can only be cleaned up `after` we get the `deleted` signal, thus, we may consider perform the cleanup at 
https://github.com/karmada-io/karmada/blob/b605d9d6371885cbf0d850eb734ae3bedb98d576/pkg/controllers/cluster/cluster_controller.go#L184-L186

In addition, please rebase this PR against master.

CharlesQQ:
> I think the metrics can only be cleaned up after we get the deleted signal, thus, we may consider perform the cleanup at

Is it true that after the cluster is physically deleted, it will be reconcile again and trigger the following logic? Let me test it.

https://github.com/karmada-io/karmada/blob/b605d9d6371885cbf0d850eb734ae3bedb98d576/pkg/controllers/cluster/cluster_controller.go#L184-L186

RainbowMango:
OK

CharlesQQ:
> > I think the metrics can only be cleaned up after we get the deleted signal, thus, we may consider perform the cleanup at
> 
> Is it true that after the cluster is physically deleted, it will be reconcile again and trigger the following logic? Let me test it.
> 
> https://github.com/karmada-io/karmada/blob/b605d9d6371885cbf0d850eb734ae3bedb98d576/pkg/controllers/cluster/cluster_controller.go#L184-L186


After testing, the following logic will not be triggered when the cluster is deleted

CharlesQQ:
After `removeFinalizer` sucessfully, cluster will be deleted immediately, may be can perform the cleanup  at
https://github.com/karmada-io/karmada/blob/622a5ef1e51ecffc3b4b8ab31f01c49f5b2bd4b3/pkg/controllers/cluster/cluster_controller.go#L381-L393

RainbowMango:
> I don't think cluster controller can get deleted signal, may be the metrics can be cleaned up after cluster's deletetimestamp is not nil or executing removeCluster action

Do you mean the following condition would never be reached?
https://github.com/karmada-io/karmada/blob/b605d9d6371885cbf0d850eb734ae3bedb98d576/pkg/controllers/cluster/cluster_controller.go#L184-L186

CharlesQQ:
> Do you mean the following condition would never be reached?
Yes

RainbowMango:
That's weird, @XiShanYongYe-Chang Can you help to take a look?

XiShanYongYe-Chang:
> That's weird, @XiShanYongYe-Chang Can you help to take a look?

Okay, let me take a look.

XiShanYongYe-Chang:
I add a log here:
```go
	klog.V(4).Infof("Reconciling cluster %s", req.NamespacedName.Name)

	cluster := &clusterv1alpha1.Cluster{}
	if err := c.Client.Get(ctx, req.NamespacedName, cluster); err != nil {
		// The resource may no longer exist, in which case we stop processing.
			klog.Infof("The cluster(%s) is no longer exist.", req.NamespacedName)
		}

	}

        ...
```

Then I install a karmada instance by running `hack/local-up-karmada.sh` script, and unjoin `member1` cluster with command `karmadactl unjoin member1`.

I can find the target log in the `karmada-controller-manager` pod:

```log
I0213 01:58:46.789249       1 cluster_controller.go:179] Reconciling cluster member1
I0213 01:58:46.789521       1 cluster_controller.go:185] The cluster(/member1) is no longer exist.
```

CharlesQQ:
Maybe it's accidental? I didn't find the log `cluster is not found, clusterName` after karmadactl unjoin


```
	klog.V(4).Infof("Reconciling cluster %s", req.NamespacedName.Name)

	cluster := &clusterv1alpha1.Cluster{}
	if err := c.Client.Get(ctx, req.NamespacedName, cluster); err != nil {
		// The resource may no longer exist, in which case we stop processing.
			klog.Infof("cluster is not found, clusterName: %s", cluster.Name)
			metrics.RemoveClusterStatus(cluster)
		}

	}

```

CharlesQQ:
my mistake, I didn't build controller binary;  After unjoin cluster, find  log `cluster is not found, clusterName: xxx`, and cleanup the  metrics successfully.  I'll fix pr, pls take  a look again?

```
	klog.V(4).Infof("Reconciling cluster %s", req.NamespacedName.Name)

	cluster := &clusterv1alpha1.Cluster{}
	if err := c.Client.Get(ctx, req.NamespacedName, cluster); err != nil {
		// The resource may no longer exist, in which case we stop processing.
			klog.Infof("cluster is not found, clusterName: %s", req.NamespacedName.Name)
			metrics.RemoveClusterStatus(req.NamespacedName.Name)
		}

	}
```

pkg/metrics/cluster.go
```diff
@@ -135,6 +135,20 @@ func RecordClusterSyncStatusDuration(cluster *v1alpha1.Cluster, startTime time.T
 	clusterSyncStatusDuration.WithLabelValues(cluster.Name).Observe(utilmetrics.DurationInSeconds(startTime))
 }
 
+// RemoveClusterStatus removes the cluster status metrics when the cluster is deleted.
+func RemoveClusterStatus(clusterName string) {
```
RainbowMango, line 139:
```suggestion
// CleanupMetricsForCluster removes the cluster status metrics after the cluster is deleted.
func CleanupMetricsForCluster(clusterName string) {
```

pkg/controllers/cluster/cluster_controller.go
```diff
@@ -182,6 +183,7 @@ func (c *Controller) Reconcile(ctx context.Context, req controllerruntime.Reques
 	if err := c.Client.Get(ctx, req.NamespacedName, cluster); err != nil {
 		// The resource may no longer exist, in which case we stop processing.
 		if apierrors.IsNotFound(err) {
+			metrics.RemoveClusterStatus(req.NamespacedName.Name)
```
RainbowMango, line 186: The cluster metrics are created by `cluster-status-controller`, It would be better to clean them by the same controller.
To be specifically, I think we can do it at 
https://github.com/karmada-io/karmada/blob/fc1df776dcd3a8ca6a8f5507d90743471ccd066f/pkg/controllers/status/cluster_status_controller.go#L134-L148

CharlesQQ: I update pr and test it!

RainbowMango (approved):
/lgtm
/approve

PR #5871, opened 2024-11-24
pkg/karmadactl: unit test join

mohamedawnallah:
**Description**

In this commit, we unit test the process of joining a cluster to the Karmada control plane. The tests validate the arguments passed to the join cluster operation and ensure the creation of the new cluster runtime object. Additionally, we verify that the necessary access secrets for the new cluster in push mode are correctly generated in the Karmada control plane.

**What type of PR is this?**

Part of #5491.

```release-note
NONE
```

### DIFF
--- a/pkg/karmadactl/join/join.go
+++ b/pkg/karmadactl/join/join.go
@@ -17,6 +17,7 @@ limitations under the License.
 package join
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 
@@ -135,10 +136,10 @@ func (j *CommandJoinOption) Complete(args []string) error {
 // Validate checks option and return a slice of found errs.
 func (j *CommandJoinOption) Validate(args []string) error {
 	if len(args) > 1 {
-		return fmt.Errorf("only the cluster name is allowed as an argument")
+		return errors.New("only the cluster name is allowed as an argument")
 	}
 	if len(j.ClusterName) == 0 {
-		return fmt.Errorf("cluster name is required")
+		return errors.New("cluster name is required")
 	}
 	if errMsgs := validation.ValidateClusterName(j.ClusterName); len(errMsgs) != 0 {
 		return fmt.Errorf("invalid cluster name(%s): %s", j.ClusterName, strings.Join(errMsgs, ";"))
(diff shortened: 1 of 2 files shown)


pkg/karmadactl/join/join.go
```diff
@@ -196,11 +197,21 @@ func (j *CommandJoinOption) Run(f cmdutil.Factory) error {
...
 
+var controlPlaneKubeClientBuilder = func(controlPlaneRestConfig *rest.Config) kubeclient.Interface {
+	return kubeclient.NewForConfigOrDie(controlPlaneRestConfig)
+}
+var karmadaClientBuilder = func(controlPlaneRestConfig *rest.Config) karmadaclientset.Interface {
+	return karmadaclientset.NewForConfigOrDie(controlPlaneRestConfig)
+}
+var clusterKubeClientBuilder = func(clusterConfig *rest.Config) kubeclient.Interface {
+	return kubeclient.NewForConfigOrDie(clusterConfig)
+}
```
XiShanYongYe-Chang, line 208: The implementation of controlPlaneKubeClientBuilder is the same as that of clusterKubeClientBuilder.

pkg/karmadactl/join/join_test.go
```diff
@@ -0,0 +1,358 @@
...
+		joinOpts                             *CommandJoinOption
+		controlPlaneRestCfg, clusterCfg      *rest.Config
+		controlKubeClient, clusterKubeClient kubeclient.Interface
+		karmadaClient                        karmadaclientset.Interface
+		prep                                 func(karmadaclientset.Interface, kubeclient.Interface, kubeclient.Interface, *CommandJoinOption, types.UID, string) error
+		verify                               func(karmadaclientset.Interface, kubeclient.Interface, kubeclient.Interface, *CommandJoinOption, types.UID) error
+		wantErr                              bool
+		errMsg                               string
+	}{
+		{
```
XiShanYongYe-Chang, line 111: I run this case failed on my local side:
```
=== RUN   TestRunJoinCluster/RunJoinCluster_ObtainClusterID_FailedToObtainClusterID
    join_test.go:212: expected error message connection refused to be in the server could not find the requested resource (get namespaces kube-system)
    --- FAIL: TestRunJoinCluster/RunJoinCluster_ObtainClusterID_FailedToObtainClusterID (0.02s)
```

pkg/karmadactl/join/join_test.go
```diff
@@ -0,0 +1,358 @@
...
+				t.Errorf("unexpected error, got: %v", err)
+			}
+			if err != nil && test.wantErr && !strings.Contains(err.Error(), test.errMsg) {
+				t.Errorf("expected error message %s to be in %s", test.errMsg, err.Error())
+			}
+		})
+	}
+}
+
+func TestRunJoinCluster(t *testing.T) {
```
XiShanYongYe-Chang, line 97: After what's happened, it's probably not our intention to introduce such complex tests, and maybe we're not dealing with such difficult-to-test methods. What do you think?

XiShanYongYe-Chang (commented):
Thanks~

pkg/karmadactl/join/join.go
XiShanYongYe-Chang: I think we can share a more common method variable naming.

mohamedawnallah: @XiShanYongYe-Chang We could unify them under a single method-naming variable, but doing so would prevent us from distinguishing between the `controlPlaneKubeClient` and `clusterKubeClient` in the test cases. Since we don't use the `*rest.Config` in the test cases, we instead mock the functions that return those types of clients. By using a single method-naming variable, we would lose the flexibility to mock and test the clients separately, reducing control in our test cases.

pkg/karmadactl/join/join_test.go
mohamedawnallah: I agree that the test cases for the joining cluster operation is somehow complex. What do you think about assuming no issues occur while communicating with Kubernetes? For example, we could remove the test cases where network issues happen, such as `RunJoinCluster_ObtainClusterID_FailedToObtainClusterID` and `RunJoinCluster_CheckIfClusterIDIsUnique_FailedToCheckIfClusterIDIsUnique`. Doing so would reduce the test cases for `TestRunJoinCluster` by 50%.

XiShanYongYe-Chang: > What do you think about assuming no issues occur while communicating with Kubernetes? For example, we could remove the test cases where network issues happen, such as RunJoinCluster_ObtainClusterID_FailedToObtainClusterID and RunJoinCluster_CheckIfClusterIDIsUnique_FailedToCheckIfClusterIDIsUnique

I think we can do this first.

pkg/karmadactl/join/join_test.go
```diff
@@ -0,0 +1,321 @@
...
+		})
+	}
+}
+
+// createNamespace creates a Kubernetes namespace with the specified name and clusterID.
+func createNamespace(name string, clusterID types.UID, client kubeclient.Interface) error {
+	ns := &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: name,
+			UID:  clusterID,
```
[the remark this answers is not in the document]
mohamedawnallah: I just debugged the `createNamespace` function and found that the UID remains the same before and after creating namespace resource.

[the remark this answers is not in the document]
XiShanYongYe-Chang: Okay, looks like fakeClient allows this. In the actual environment, the specified mode cannot be used. A new UID is generated during creation.

XiShanYongYe-Chang (commented):
Thanks
/lgtm
/approve

PR #5881, opened 2024-11-26
Failover controllers now build eviction tasks for purgemode immediately

mszacillo:
**What type of PR is this?**
/kind feature

Part of #5788 

```release-note
Build eviction task for application failover when using purgeMode Immediately
```


pkg/controllers/applicationfailover/rb_application_failover_controller.go
```diff
@@ -174,7 +177,11 @@ func (c *RBApplicationFailoverController) evictBinding(binding *workv1alpha2.Res
 				return err
 			}
 		case policyv1alpha1.Immediately:
-			binding.Spec.RemoveCluster(cluster)
+			binding.Spec.GracefulEvictCluster(cluster, workv1alpha2.NewTaskOptions(
+				workv1alpha2.WithPurgeMode(policyv1alpha1.Immediately),
+				workv1alpha2.WithProducer(RBApplicationFailoverControllerName),
+				workv1alpha2.WithReason(workv1alpha2.EvictionReasonApplicationFailure),
+				workv1alpha2.WithGracePeriodSeconds(binding.Spec.Failover.Application.GracePeriodSeconds)))
```
mszacillo, line 184: I included a grace period window of 0s for the Immediately purgeMode as otherwise if the grace period is not set (which it will not be for Immediately), the default grace period will be set to 10 minutes: https://github.com/karmada-io/karmada/blob/master/cmd/controller-manager/app/options/options.go#L226

pkg/controllers/applicationfailover/rb_application_failover_controller.go
```diff
@@ -157,8 +157,11 @@ func (c *RBApplicationFailoverController) evictBinding(binding *workv1alpha2.Res
 		switch binding.Spec.Failover.Application.PurgeMode {
 		case policyv1alpha1.Graciously:
 			if features.FeatureGate.Enabled(features.GracefulEviction) {
-				binding.Spec.GracefulEvictCluster(cluster, workv1alpha2.NewTaskOptions(workv1alpha2.WithProducer(RBApplicationFailoverControllerName),
-					workv1alpha2.WithReason(workv1alpha2.EvictionReasonApplicationFailure), workv1alpha2.WithGracePeriodSeconds(binding.Spec.Failover.Application.GracePeriodSeconds)))
+				binding.Spec.GracefulEvictCluster(cluster, workv1alpha2.NewTaskOptions(
```
mszacillo, line 160: Eventually this can be combined into a single case statement with Graciously + Immediately, but while the FeatureGate exists, we'll need to separate them.

RainbowMango (commented):
/assign

pkg/controllers/applicationfailover/rb_application_failover_controller.go
XiShanYongYe-Chang: If we set GracePeriodSeconds to 0, the following effect will occur, as soon as the cluster is added to the eviction task, it will be cleaned up immediately, which may not be what we expected.

Therefore, I understand that it is OK to use the maximum wait time of 10 minutes here. For clusters with Immediately purgeMode, the logic for clearing the eviction queue can be the same as that for Graceful purgeMode. The difference lies in whether the old `work` is deleted immediately.

How do you think?

pkg/controllers/applicationfailover/rb_application_failover_controller.go
XiShanYongYe-Chang: Agree.

Besides, Do we also need to deal with Never's situation?

XiShanYongYe-Chang (commented):
Thanks a lot~

pkg/controllers/applicationfailover/crb_application_failover_controller.go
```diff
@@ -172,7 +176,11 @@ func (c *CRBApplicationFailoverController) evictBinding(binding *workv1alpha2.Cl
 				return err
 			}
 		case policyv1alpha1.Immediately:
-			binding.Spec.RemoveCluster(cluster)
+			binding.Spec.GracefulEvictCluster(cluster, workv1alpha2.NewTaskOptions(
+				workv1alpha2.WithPurgeMode(binding.Spec.Failover.Application.PurgeMode),
```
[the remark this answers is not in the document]
RainbowMango: @XiShanYongYe-Chang 
What's the difference?

pkg/controllers/applicationfailover/rb_application_failover_controller.go
RainbowMango: > Besides, Do we also need to deal with Never's situation?

+1

pkg/controllers/applicationfailover/rb_application_failover_controller.go
mszacillo: Makes sense. As long as we can insure that the old work is cleared immediately before we decide to reschedule, that should be okay.

pkg/controllers/applicationfailover/rb_application_failover_controller.go
mszacillo: I see, so in the case of Never we can generate a GracefulEvictionTask (which won't actually evict the workload), but will be used to filter out the cluster that the application is failing over from?

What are the use cases for an application never being cleaned up? Just out of curiosity.

XiShanYongYe-Chang: > I see, so in the case of Never we can generate a GracefulEvictionTask (which won't actually evict the workload), but will be used to filter out the cluster that the application is failing over from?

Yes, it's like this.

XiShanYongYe-Chang (commented):
Thanks~
/lgtm
cc @RainbowMango

PR #5885, opened 2024-11-26
Introduce StatePreservation to PropagationPolicy API

RainbowMango:
**What type of PR is this?**

/kind api-change
/kind feature

This PR introduces `StatePreservation` to PropagationPolicy which will be used to preserve status in case of application failover.

Part of #5788

This PR focuses on application failover behavior, for the cluster failover, will be included in a separate PR.

```release-note
`API Change`: Introduced `StatePreservation` to `PropagationPolicy`, which will be used to preserve status in case of application failover.
```


RainbowMango:
cc @mszacillo 
This PR does not include cluster failover. I plan to consider it in a subsequent PR(likely in the next release), as it's more complex and involves changing the default behavior of cluster failover. 
Let me know what do you think.

a7i (approved):
/lgtm

curious, could this not be achieved via a customized interpreter?

mszacillo:
> This PR does not include cluster failover. I plan to consider it in a subsequent PR(likely in the next release), as it's more complex and involves changing the default behavior of cluster failover. Let me know what do you think.

Should be alright. Until then we can probably get around with our own local cluster failover change that would give us the failover flag like we want. Let me know if you need help with the cluster failover part!

RainbowMango:
> /lgtm
> 
> curious, could this not be achieved via a customized interpreter?

Probably not, as the data could be grabbed from the status that has been collected to ResourceBinding, and I also [made a demo for grabbing the data](https://github.com/karmada-io/karmada/commit/7ab66e341abbf65806d868f670e5fbcf9d8fa69b).

XiShanYongYe-Chang (commented):
LGTM

RainbowMango:
> Until then we can probably get around with our own local cluster failover change that would give us the failover flag like we want. Let me know if you need help with the cluster failover part!

Absolutely, I'm glad to help. But I need to focus on the application failover part as I hope this could be included in this release.

PR #5889, opened 2024-11-28
Cleanup works from cluster if purgemode is immediately

mszacillo:
**What type of PR is this?**
/kind feature

Within the [binding controller](https://github.com/karmada-io/karmada/blob/38ee277de7aaaf4f93fec42d49bfc71cc4b0be76/pkg/controllers/binding/binding_controller.go#L145-L164) and [cluster binding controller](https://github.com/karmada-io/karmada/blob/38ee277de7aaaf4f93fec42d49bfc71cc4b0be76/pkg/controllers/binding/cluster_resource_binding_controller.go#L144-L160) we find and remove orphan works by checking the works against the existing clusters.

Existing clusters are populated by joining the resource binding's `spec.Clusters` field with any clusters that are found in the GracefulEvictionTasks. With the addition of PurgeMode, we should now check that the PurgeMode is not Immediately when populating the existing clusters. 

If the cluster has an eviction task with Immediately purgeMode, we should treat the work on that cluster as an orphan and evict it.

Part of #5788 

```release-note
Cleanup works from clusters with eviction task when purge mode is immediately
```

### DIFF
--- a/pkg/util/helper/binding.go
+++ b/pkg/util/helper/binding.go
@@ -186,7 +186,11 @@ func ObtainBindingSpecExistingClusters(bindingSpec workv1alpha2.ResourceBindingS
 	}
 
 	for _, task := range bindingSpec.GracefulEvictionTasks {
-		clusterNames.Insert(task.FromCluster)
+		// EvictionTasks with Immediately PurgeMode should not be treated as existing clusters
+		// Work on those clusters should be removed immediately and treated as orphans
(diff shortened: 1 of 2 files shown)


RainbowMango (commented):
/lgtm
/assign @XiShanYongYe-Chang

XiShanYongYe-Chang (commented):
Thanks a lot~
/approve

PR #5897, opened 2024-11-29
Add stateful application failover status injection feature gate

RainbowMango:
**What type of PR is this?**

/kind feature

This PR adds a feature gate `StatefulFailoverInjection` which controls whether Karmada collects state information from the source cluster during a failover event for stateful applications and injects this information into the application configuration when it is moved to the target cluster.

Part of #5788

```release-note
Introduced feature gate `StatefulFailoverInjection` to control whether Karmada collects and injects state information during a failover event for stateful application.
```

### DIFF
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -18291,7 +18291,7 @@
           "type": "string"
         },
         "statePreservation": {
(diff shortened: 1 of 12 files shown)


RainbowMango:
cc @mszacillo @XiShanYongYe-Chang 
We better have a feature gate for this feature, as we will continue to iterate it in the following releases.

RainbowMango:
Forced pushed to fix the unit test which requires enable feature gate.

PR #5899, opened 2024-11-29
Disable cluster failover by default which should be explicitly enabled by administrators

RainbowMango:
**What type of PR is this?**

/kind feature

This PR updates the `Failover` feature gate to be disabled by default. 
The changes are based on user feedback,

In the future, the cluster failover feature will focus on making it more user-friendly and visible, for example, providing the support of explicitly specifying applications that are eligible for cluster failover when configuring application distribution policies.

Fixes #

```release-note
The feature `Failover` now has been disabled by default, which should be explicitly enabled to avoid unexpected incidents. 
```

### DIFF
--- a/pkg/controllers/applicationfailover/crb_application_failover_controller.go
+++ b/pkg/controllers/applicationfailover/crb_application_failover_controller.go
@@ -36,7 +36,6 @@ import (
 
 	configv1alpha1 "github.com/karmada-io/karmada/pkg/apis/config/v1alpha1"
 	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
-	"github.com/karmada-io/karmada/pkg/features"
 	"github.com/karmada-io/karmada/pkg/resourceinterpreter"
 	"github.com/karmada-io/karmada/pkg/sharedcli/ratelimiterflag"
 	"github.com/karmada-io/karmada/pkg/util/helper"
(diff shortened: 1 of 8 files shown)


pkg/controllers/applicationfailover/crb_application_failover_controller.go
```diff
@@ -173,11 +172,6 @@ func (c *CRBApplicationFailoverController) updateBinding(ctx context.Context, bi
 	for _, cluster := range needEvictClusters {
 		allClusters.Delete(cluster)
 	}
-	if !features.FeatureGate.Enabled(features.GracefulEviction) {
-		for _, cluster := range needEvictClusters {
-			helper.EmitClusterEvictionEventForClusterResourceBinding(binding, cluster, c.EventRecorder, nil)
-		}
-	}
```
XiShanYongYe-Chang, line 180: In the same way, we might need to clean up these places:

https://github.com/karmada-io/karmada/blob/54be414d9f24eea643b74f4b7c1f5e294778dfb9/pkg/controllers/cluster/taint_manager.go#L193-L195

https://github.com/karmada-io/karmada/blob/54be414d9f24eea643b74f4b7c1f5e294778dfb9/pkg/controllers/cluster/taint_manager.go#L255-L257

RainbowMango: Updated as suggested.

RainbowMango:
@LivingCcj 
As we discussed in WeChat, this PR aims to disable cluster failover which will be included in v1.12.0.

In addition, after the failover feature is disabled, the controller will no longer taint clusters upon cluster failure. As a result, the default taint tolerance policies specified in the PropagationPolicy( by karmada-webhook) will have no practical effect.

Let me know if you have further concerns.

RainbowMango:
@wangxf1987 This is we discussed at KubeCon China, from the v1.12.0, the failover will be disabled by default.

RainbowMango:
/retest
The failing test is unrelated.

XiShanYongYe-Chang (commented):
LGTM

PR #5933, opened 2024-12-10
fix flake test of KarmadactlTopPod

zhzhuang-zju:
**What type of PR is this?**
/kind flake
/kind failing-test

The e2e test `Karmadactl top testing Karmadactl top pod which does not exist [It] Karmadactl top pod which does not exist` has been intermittently failing. 
See,  https://github.com/karmada-io/karmada/actions/runs/12197101877/job/34026335469?pr=5772#step:6:1770

the e2e fails because the member cluster's server is not available at this time.
> error: cluster(member1): the server is currently unable to handle the request (get pods.metrics.k8s.io pod-z682t)

We are unable to determine why the metrics-server in member1 cannot handle requests due to the absence of logs from the member cluster and uncertainty regarding whether this is a transient service outage. However, we can enhance the end-to-end (e2e) stability by implementing retry logic.

Fixes #

```release-note
NONE.
```


XiShanYongYe-Chang (commented):
Thanks
/lgtm
/approve

PR #5936, opened 2024-12-11
pkg/search/controllers_test: reduce the probability of flaky test cases by waiting for cache to sync

### LINKED ISSUE #5926: pkg/search/controller: flaky tests
#### Which jobs are flaking:
pkg/search/controller

https://github.com/karmada-io/karmada/actions/runs/12228313905/job/34106544113

#### Reason for failure:
Work in progress ⏳

mohamedawnallah:
**What type of PR is this?**

/kind flake

This PR reduce the probability of flaky test cases by waiting for cache to sync. It seems fake client has some flaky issues with watch implementation that is different from real implementation as seen in https://github.com/kubernetes/kubernetes/issues/87472#issuecomment-577472496?

Fixes #5926

```release-note
NONE
```

### DIFF
--- a/pkg/search/controllers_test.go
+++ b/pkg/search/controllers_test.go
@@ -41,6 +41,7 @@ import (
 	informerfactory "github.com/karmada-io/karmada/pkg/generated/informers/externalversions"
 	"github.com/karmada-io/karmada/pkg/search/backendstore"
 	"github.com/karmada-io/karmada/pkg/util"
+	"github.com/karmada-io/karmada/pkg/util/fedinformer/genericmanager"
 )
 
 var apiGroupResources = []*restmapper.APIGroupResources{
@@ -125,7 +126,7 @@ func TestAddClusterEventHandler(t *testing.T) {
 		client     *fakekarmadaclient.Clientset
 		restMapper meta.RESTMapper
 		stopCh     chan struct{}
-		prep       func(*fakekarmadaclient.Clientset, *rest.Config, meta.RESTMapper) (*Controller, informerfactory.SharedInformerFactory, error)
+		prep       func(*fakekarmadaclient.Clientset, *rest.Config, meta.RESTMapper, chan struct{}) (*Controller, informerfactory.SharedInformerFactory, error)
 		verify     func(*fakekarmadaclient.Clientset, *Controller) error
 	}{
 		{
@@ -134,9 +135,9 @@ func TestAddClusterEventHandler(t *testing.T) {
 			client:     fakekarmadaclient.NewSimpleClientset(),
 			restMapper: meta.NewDefaultRESTMapper(nil),
 			stopCh:     make(chan struct{}),
-			prep: func(clientConnector *fakekarmadaclient.Clientset, restConfig *rest.Config, restMapper meta.RESTMapper) (*Controller, informerfactory.SharedInformerFactory, error) {
+			prep: func(clientConnector *fakekarmadaclient.Clientset, restConfig *rest.Config, restMapper meta.RESTMapper, stopCh chan struct{}) (*Controller, informerfactory.SharedInformerFactory, error) {
 				factory := informerfactory.NewSharedInformerFactory(clientConnector, time.Second)
-				controller, err := createController(restConfig, factory, restMapper)
+				controller, err := createController(restConfig, factory, restMapper, stopCh)
 				return controller, factory, err
 			},
 			verify: func(clientConnector *fakekarmadaclient.Clientset, controller *Controller) error {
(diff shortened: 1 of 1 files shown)


pkg/search/controllers_test.go
```diff
@@ -577,11 +584,12 @@ func TestDeleteResourceRegistryEventHandler(t *testing.T) {
 // createController initializes a new Controller instance using the provided
 // Kubernetes REST configuration, shared informer factory, and REST mapper.
 // It returns the created Controller or an error if initialization fails.
-func createController(restConfig *rest.Config, factory informerfactory.SharedInformerFactory, restMapper meta.RESTMapper) (*Controller, error) {
+func createController(restConfig *rest.Config, factory informerfactory.SharedInformerFactory, restMapper meta.RESTMapper, stopCh chan struct{}) (*Controller, error) {
 	newController, err := NewController(restConfig, factory, restMapper)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create new controller, got: %v", err)
 	}
+	newController.InformerManager = genericmanager.NewMultiClusterInformerManager(stopCh)
```
XiShanYongYe-Chang, line 592: Is this modification unnecessary? newController.InformerManager has been initialized.

XiShanYongYe-Chang:
There is indeed such a setting in the controller startup, and I think this should work:

https://github.com/karmada-io/karmada/blob/471d850a1985acf6ee79f7d21b3dee5b33466204/pkg/search/controller.go#L136

zhzhuang-zju:
I executed `go test -v pkg/search/controller.go pkg/search/controllers_test.go --count=2000` twice in my environment, and it finally failed.

XiShanYongYe-Chang:
In terms of implementation, the WaitForCacheSync method should be called. There may be other problems with the ut failure. I feel that I can continue to push this pr and explore other problems.

Hi @mohamedawnallah can you revise it according to the comments?

pkg/search/controllers_test.go
mohamedawnallah: The new multicluster informer manager global instance is created using the singleton pattern (`once`), which ensures it is instantiated only once. As a result, test cases depend on this single instance, leading to unintended dependencies between the tests. What do you think?

https://github.com/karmada-io/karmada/blob/8d89d77110cf378d09b5f43dcb2e09bb393d61f1/pkg/util/fedinformer/genericmanager/multi-cluster-manager.go#L39-L44

XiShanYongYe-Chang: Yes, you're right, it's right not to share data between each test case.

mohamedawnallah:
> I executed go test -v pkg/search/controller.go pkg/search/controllers_test.go --count=2000 twice in my environment, and it finally failed.

Thanks, @zhzhuang-zju, for testing this 2000 times, twice! You're right. I’ve also tried this, and as mentioned in the PR description, the proposed changes reduce the likelihood of flaky test cases. Before, the tests were failing more frequently. It would be interesting to gather more empirical observations from the GitHub CI test failures as @XiShanYongYe-Chang  pointed, if any. What are your thoughts?

mohamedawnallah:
> In terms of implementation, the WaitForCacheSync method should be called. There may be other problems with the ut failure. I feel that I can continue to push this pr and explore other problems.

Currently, the `Start` method on the controller is busy-waiting for a signal from `stopCh`, and during this process, it caches the next element every second. After testing this, I encountered some timing issues. We might need to introduce static sleep in the verification code to allow time for the busy-waiting goroutine to execute the caching. This could reduce control over the verification code in the test cases. For instance, when a cluster is upserted, we would want to cache it immediately after.

What do you think, @XiShanYongYe-Chang?

https://github.com/karmada-io/karmada/blob/471d850a1985acf6ee79f7d21b3dee5b33466204/pkg/search/controller.go#L131-L145

mohamedawnallah:
Apologies for the delayed response to the PR feedback. I've been quite busy recently, but I’ve now added comments addressing all the feedback. I’d love to hear your thoughts on them. 🙏

XiShanYongYe-Chang:
> We might need to introduce static sleep in the verification code to allow time for the busy-waiting goroutine to execute the caching.

I don't quite understand why we need to add static time to wait. Events are blocked waiting, and as long as events can be fetched from the queue, processing is executed. So I understand that the changes now are sufficient.

mohamedawnallah:
> I don't quite understand why we need to add static time to wait. Events are blocked waiting, and as long as events can be fetched from the queue, processing is executed. So I understand that the changes now are sufficient.

Sorry for not providing enough context! 🙏 I was referring to what happens if we use `controller.Start(test.stopCh)`.

XiShanYongYe-Chang:
Oh! Thanks @mohamedawnallah 

> Apologies for the delayed response to the PR feedback.

It's okay. As long as you have time, you can always contribute.

zhzhuang-zju:
thanks
/lgtm

PR #5690, opened 2024-10-13
[Proposal] Support for resource scheduling suspend and resume capabilities

Vacant2333:
**What type of PR is this?**

/kind design
/kind documentation

**Which issue(s) this PR fixes**:
Fixes https://github.com/karmada-io/karmada/issues/4937

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

Issue #5217, opened 2024-07-18
[Umbrella] Support for cluster-level resource propagation pause and resume capabilities


XiShanYongYe-Chang:
**What would you like to be added:**

Support for cluster-level resource propagation pause and resume capabilities.

**Why is this needed:**

Received feedback from multiple issues, and there is a hope that Karmada can provide the capability to pause distribution, allowing users to expand more capabilities with it.

Related issue: https://github.com/karmada-io/karmada/issues/1567, https://github.com/karmada-io/karmada/issues/4421, https://github.com/karmada-io/karmada/issues/4688, https://github.com/karmada-io/karmada/issues/4559

- [x] Add proposal for this feature @XiShanYongYe-Chang #5118
- [x] API Modification @a7i #4838 
  - [x] PropagationPolicy/ClusterPropagationPolicy
  - [x] ResourceBinding/ClusterResourceBinding
  - [x] Work
- [x] Implement the capability to pause resource distribution in the controller.
  - [x] Suspension field transfer and distribution judge @a7i #4838
  - [x] Handle with suspension status condition @a7i #5317
  - [x] Webhook validate @a7i #5282 
  - [x] Add event to describe work dispatching condition @a7i #5332 
- [x] Test for the feature
  - [x] UT
  - [x] E2E @a7i #4838 #5317 #5354
- [x] Add user documentation for this feature. @a7i https://github.com/karmada-io/website/pull/668

XiShanYongYe-Chang:
Hi @a7i, I created a umbrella issue for resource propagation pause.
/cc @a7i

a7i:
Awesome! I will start on https://github.com/karmada-io/karmada/pull/4838 tomorrow to align with the new proposal.

XiShanYongYe-Chang:
Thanks a lot @a7i we've completed all the subtasks.
Looking forward to the feedback from different partners in the community. Let's keep iterating this feature.
/close

PR #5937, opened 2024-12-11
feat: cluster-level resource scheduling suspend and resume capabilities

Monokaix:
**What type of PR is this?**
/kind api-change
/kind feature

Add rb suspension capability, background: https://github.com/karmada-io/karmada/pull/5690

Fixes https://github.com/karmada-io/karmada/pull/5690

It extends the Suspension fileds and support rb susupend and resume, mostly like the previous issue https://github.com/karmada-io/karmada/issues/5217 and PRs implemented in it. 
 This pr just implemented  binding suspension in `scheduler `, setting status condition in `controller` will be implemented in another pr after https://github.com/karmada-io/karmada/pull/6002 merged because it may lead to status update conflict between scheduler and controller.

```release-note
`API Change`: Introduced `Scheduling` suspension in both `ResourceBinding` and `ClusterResourceBinding` which will be used for third-party systems to suspend application scheduling.
```

### DIFF
--- a/pkg/apis/work/v1alpha2/binding_types.go
+++ b/pkg/apis/work/v1alpha2/binding_types.go
@@ -322,9 +322,19 @@ type BindingSnapshot struct {
 	Clusters []TargetCluster `json:"clusters,omitempty"`
 }
 
-// Suspension defines the policy for suspending of propagation.
+// Suspension defines the policy for suspending dispatching and scheduling.
 type Suspension struct {
 	policyv1alpha1.Suspension `json:",inline"`
+
+	// Scheduling controls whether scheduling should be suspended, the scheduler will pause scheduling and not
+	// process resource binding when the value is true and resume scheduling when it's false or nil.
+	// This is designed for third-party systems to temporarily pause the scheduling of applications, which enabling
+	// manage resource allocation, prioritize critical workloads, etc.
+	// It is expected that third-party systems use an admission webhook to suspend scheduling at the time of
+	// ResourceBinding creation. Once a ResourceBinding has been scheduled, it cannot be paused afterward, as it may
+	// lead to ineffective suspension.
+	// +optional
+	Scheduling *bool `json:"scheduling,omitempty"`
 }
 
 // ResourceBindingStatus represents the overall status of the strategy as well as the referenced resources.
--- a/pkg/util/binding.go
+++ b/pkg/util/binding.go
@@ -110,3 +110,19 @@ func RescheduleRequired(rescheduleTriggeredAt, lastScheduledTime *metav1.Time) b
 	}
 	return rescheduleTriggeredAt.After(lastScheduledTime.Time)
 }
+
+// IsBindingSuspendScheduling tells whether resource binding is scheduling suspended.
+func IsBindingSuspendScheduling(rb *workv1alpha2.ResourceBinding) bool {
+	if rb == nil || rb.Spec.Suspension == nil || rb.Spec.Suspension.Scheduling == nil {
+		return false
+	}
+	return *rb.Spec.Suspension.Scheduling
+}
+
+// IsClusterBindingSuspendScheduling tells whether cluster resource binding is scheduling suspended.
+func IsClusterBindingSuspendScheduling(crb *workv1alpha2.ClusterResourceBinding) bool {
+	if crb == nil || crb.Spec.Suspension == nil || crb.Spec.Suspension.Scheduling == nil {
+		return false
+	}
+	return *crb.Spec.Suspension.Scheduling
+}
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -20435,7 +20435,7 @@
       }
     },
     "com.github.karmada-io.karmada.pkg.apis.work.v1alpha2.Suspension": {
-      "description": "Suspension defines the policy for suspending of propagation.",
+      "description": "Suspension defines the policy for suspending dispatching and scheduling.",
       "type": "object",
       "properties": {
         "dispatching": {
@@ -20445,6 +20445,10 @@
         "dispatchingOnClusters": {
           "description": "DispatchingOnClusters declares a list of clusters to which the dispatching should be suspended. Note: Can not co-exist with Dispatching which is used to suspend all.",
           "$ref": "#/definitions/com.github.karmada-io.karmada.pkg.apis.policy.v1alpha1.SuspendClusters"
+        },
+        "scheduling": {
+          "description": "Scheduling controls whether scheduling should be suspended, the scheduler will pause scheduling and not process resource binding when the value is true and resume scheduling when it's false or nil. This is designed for third-party systems to temporarily pause the scheduling of applications, which enabling manage resource allocation, prioritize critical workloads, etc. It is expected that third-party systems use an admission webhook to suspend scheduling at the time of ResourceBinding creation. Once a ResourceBinding has been scheduled, it cannot be paused afterward, as it may lead to ineffective suspension.",
+          "type": "boolean"
         }
       }
     },
--- a/charts/karmada/_crds/bases/work/work.karmada.io_clusterresourcebindings.yaml
+++ b/charts/karmada/_crds/bases/work/work.karmada.io_clusterresourcebindings.yaml
@@ -1281,6 +1281,16 @@ spec:
                           type: string
                         type: array
                     type: object
+                  scheduling:
+                    description: |-
+                      Scheduling controls whether scheduling should be suspended, the scheduler will pause scheduling and not
+                      process resource binding when the value is true and resume scheduling when it's false or nil.
+                      This is designed for third-party systems to temporarily pause the scheduling of applications, which enabling
+                      manage resource allocation, prioritize critical workloads, etc.
+                      It is expected that third-party systems use an admission webhook to suspend scheduling at the time of
+                      ResourceBinding creation. Once a ResourceBinding has been scheduled, it cannot be paused afterward, as it may
+                      lead to ineffective suspension.
+                    type: boolean
                 type: object
             required:
             - resource
--- a/charts/karmada/_crds/bases/work/work.karmada.io_resourcebindings.yaml
+++ b/charts/karmada/_crds/bases/work/work.karmada.io_resourcebindings.yaml
@@ -1281,6 +1281,16 @@ spec:
                           type: string
                         type: array
                     type: object
+                  scheduling:
+                    description: |-
+                      Scheduling controls whether scheduling should be suspended, the scheduler will pause scheduling and not
+                      process resource binding when the value is true and resume scheduling when it's false or nil.
+                      This is designed for third-party systems to temporarily pause the scheduling of applications, which enabling
+                      manage resource allocation, prioritize critical workloads, etc.
+                      It is expected that third-party systems use an admission webhook to suspend scheduling at the time of
+                      ResourceBinding creation. Once a ResourceBinding has been scheduled, it cannot be paused afterward, as it may
+                      lead to ineffective suspension.
+                    type: boolean
                 type: object
             required:
             - resource
--- a/pkg/apis/work/v1alpha2/zz_generated.deepcopy.go
+++ b/pkg/apis/work/v1alpha2/zz_generated.deepcopy.go
@@ -421,6 +421,11 @@ func (in *ResourceBindingStatus) DeepCopy() *ResourceBindingStatus {
 func (in *Suspension) DeepCopyInto(out *Suspension) {
 	*out = *in
 	in.Suspension.DeepCopyInto(&out.Suspension)
+	if in.Scheduling != nil {
+		in, out := &in.Scheduling, &out.Scheduling
+		*out = new(bool)
+		**out = **in
+	}
 	return
 }
 
--- a/pkg/generated/openapi/zz_generated.openapi.go
+++ b/pkg/generated/openapi/zz_generated.openapi.go
@@ -7461,7 +7461,7 @@ func schema_pkg_apis_work_v1alpha2_Suspension(ref common.ReferenceCallback) comm
 	return common.OpenAPIDefinition{
 		Schema: spec.Schema{
 			SchemaProps: spec.SchemaProps{
-				Description: "Suspension defines the policy for suspending of propagation.",
+				Description: "Suspension defines the policy for suspending dispatching and scheduling.",
 				Type:        []string{"object"},
 				Properties: map[string]spec.Schema{
 					"dispatching": {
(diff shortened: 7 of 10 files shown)


Monokaix:
cc @RainbowMango @XiShanYongYe-Chang @chaosi-zju

pkg/scheduler/scheduler_test.go
```diff
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"k8s.io/utils/ptr"
```
XiShanYongYe-Chang, line 23: This import line needs to be put into the next import block.

Monokaix: done.

pkg/apis/policy/v1alpha1/propagation_types.go
```diff
@@ -248,6 +248,11 @@ type Suspension struct {
 	// Note: Can not co-exist with Dispatching which is used to suspend all.
 	// +optional
 	DispatchingOnClusters *SuspendClusters `json:"dispatchingOnClusters,omitempty"`
+
+	// Scheduling controls whether scheduling should be suspended.
+	// nil means not suspend, no default value, only accepts 'true'.
+	// Karmada scheduler will pause scheduling when value is true and resume scheduling when it's nil.
+	Scheduling *bool `json:"scheduling,omitempty"`
```
RainbowMango, line 255: Can you explain a little bit why `PropagationPolicy` needs this?

pkg/apis/work/v1alpha2/binding_types.go
```diff
@@ -385,6 +385,9 @@ const (
 	// FullyApplied represents the condition that the resource referencing by ResourceBinding or ClusterResourceBinding
 	// has been applied to all scheduled clusters.
 	FullyApplied string = "FullyApplied"
+
+	// Suspended represents the condition that the ResourceBinding or ClusterResourceBinding is suspended to schedule.
+	Suspended string = "Suspended"
```
RainbowMango, line 390: This is ambiguous as people can not tell what's suspended.

pkg/controllers/binding/common.go
```diff
@@ -319,3 +335,72 @@ func shouldSuspendDispatching(suspension *policyv1alpha1.Suspension, targetClust
...
+
+	if bindingSuspendScheduling {
+		newBindingSchedulingCondition.Status = metav1.ConditionTrue
+		newBindingSchedulingCondition.Message = SuspendedSchedulingConditionMessage
+		newBindingSchedulingCondition.Reason = SuspendedSchedulingConditionReason
+	} else {
+		// There is no need to set an extra Suspended condition when scheduling not suspended and Suspended condition not exists.
+		if meta.FindStatusCondition(*conditions, workv1alpha2.Suspended) == nil {
+			return nil
+		}
```
RainbowMango, line 369: Looks like a bug here, losing the chance to update the condition from `Suspended-true` to `Suspended-false`.

pkg/events/events.go
```diff
@@ -56,6 +56,8 @@ const (
 	EventReasonSyncWorkloadSucceed = "SyncSucceed"
 	// EventReasonWorkDispatching indicates that work is dispatching or not.
 	EventReasonWorkDispatching = "WorkDispatching"
+	// EventReasonBindingScheduling indicates that binding is scheduling or not.
+	EventReasonBindingScheduling = "BindingScheduling"
```
RainbowMango, line 60: Can you remind me the benefit of having this event?

pkg/scheduler/scheduler.go
```diff
@@ -358,6 +358,11 @@ func (s *Scheduler) doScheduleBinding(namespace, name string) (err error) {
 		return nil
 	}
 
+	if util.IsBindingSuspendScheduling(rb) {
+		klog.V(4).InfoS("Skip schedule suspended ResourceBinding", "ResourceBinding", klog.KObj(rb))
+		return nil
+	}
```
RainbowMango, line 364: I think we should skip reconciling the suspended ResourceBinding earlier. 
Probably at https://github.com/karmada-io/karmada/blob/59b8835bda0710c42e2ae30f6c13c13f2a513480/pkg/scheduler/event_handler.go#L93-L113

RainbowMango (commented):
Don't we need to introduce a field to ResourceBinding?

pkg/apis/policy/v1alpha1/propagation_types.go
Monokaix: The `Suspension` is common struct which indicates a suspension semantics, so based on the existing `Suspension`, I added scheduling suspension other than dispatcher. If you think it is inappropriate, I can add a new suspension structure just for scheduling, but this seems a bit redundant.

pkg/apis/work/v1alpha2/binding_types.go
Monokaix: As far as I know, the Rb resource is an abstract of scheduling, so maybe it's enough? And other conditions are `Scheduled string = "Scheduled"`, which are also a semantic related to scheduling, or could you give some advice?

pkg/controllers/binding/common.go
Monokaix: This check only retrun when `Suspended` condition is not existing, and if `Suspended` condition is true, it will be updated to false in the code below: )

pkg/events/events.go
Monokaix: The user can have an intuitive message to know what is going on about the scheduling process.

pkg/apis/policy/v1alpha1/propagation_types.go
RainbowMango: Sorry, I didn't make it clear. 
In my understanding, what you need is introducing scheduling suspension to ResourceBinding, my question is do you need to do that to PropgationPolicy as well?

PS:
I know both ResourceBinding and PropgationPolicy share the same typed structure (`type Suspension struct`). They may no longer be able to share the existing structure if we don't want to introduce scheduling suspension to PropgationPolicy.

pkg/controllers/binding/common.go
RainbowMango: OK, got it.

pkg/events/events.go
RainbowMango: OK.

Seems we should put it on the `const block` below where defines the events for ResourceBinding.

pkg/scheduler/scheduler.go
Monokaix: seems OK.

pkg/apis/policy/v1alpha1/propagation_types.go
Monokaix: Yeah you are right, currenly I don't need the filed in PropgationPolicy, unless we just want to debug this feature easily.

pkg/events/events.go
Monokaix: It makes sense: )

pkg/apis/policy/v1alpha1/propagation_types.go
Monokaix: Yeah you are 
right, has modified: )

pkg/apis/policy/v1alpha1/propagation_types.go
```diff
@@ -248,6 +248,11 @@ type Suspension struct {
 	// Note: Can not co-exist with Dispatching which is used to suspend all.
 	// +optional
 	DispatchingOnClusters *SuspendClusters `json:"dispatchingOnClusters,omitempty"`
+
+	// Scheduling controls whether scheduling should be suspended.
+	// nil means not suspend, no default value, only accepts 'true'.
+	// scheduler will pause scheduling when value is true and resume scheduling when it's nil.
+	Scheduling *bool `json:"scheduling,omitempty"`
```
kevin-wangzefeng, line 255: It's not clear why we need flag field here, also the field "scheduling" looks confusing what does the admin really want to stop system from doing.
Whether the resourcebinding shall be created?

pkg/apis/work/v1alpha2/binding_types.go
```diff
@@ -385,6 +385,9 @@ const (
 	// FullyApplied represents the condition that the resource referencing by ResourceBinding or ClusterResourceBinding
 	// has been applied to all scheduled clusters.
 	FullyApplied string = "FullyApplied"
+
+	// Suspended represents the condition that the ResourceBinding or ClusterResourceBinding is suspended to schedule.
+	Suspended string = "SchedulingSuspended"
```
kevin-wangzefeng, line 390:
```suggestion
	// SchedulingSuspended represents the condition that the ResourceBinding or ClusterResourceBinding is suspended to schedule.
	SchedulingSuspended string = "SchedulingSuspended"
```

pkg/apis/policy/v1alpha1/propagation_types.go
Monokaix: The resourcebinding will be created but scheduler will not scheduler, because we need some admission check such as quota check, multi-tenancy prioritization, etc before schduler process resourcebinding.
Will add more comments: )

pkg/apis/work/v1alpha2/binding_types.go
Monokaix: done

pkg/apis/policy/v1alpha1/propagation_types.go
Monokaix: Have decoupled in https://github.com/karmada-io/karmada/pull/5974.

pkg/apis/work/v1alpha2/binding_types.go
Monokaix: modified.

pkg/apis/work/v1alpha2/binding_types.go
```diff
@@ -322,9 +322,17 @@ type BindingSnapshot struct {
 	Clusters []TargetCluster `json:"clusters,omitempty"`
 }
 
-// Suspension defines the policy for suspending of propagation.
+// Suspension defines the policy for suspending of propagation, dispatching and scheduling suspension are independent
+// of each other. The propagation suspension inherited from PropagationPolicy can be transparently transmitted here,
+// and ResourceBinding can also set scheduling suspension separately.
```
RainbowMango, line 327:
```suggestion
// Suspension defines the policy for suspending dispatching and scheduling.
```

pkg/apis/work/v1alpha2/binding_types.go
```diff
@@ -322,9 +322,17 @@ type BindingSnapshot struct {
...
 type Suspension struct {
 	policyv1alpha1.Suspension `json:",inline"`
+
+	// Scheduling controls whether scheduling should be suspended, the scheduler will pause scheduling and not
+	// process resource binding after it's created when value is true and resume scheduling when it's false or nil.
+	// This is usually used for admission checks by third-party such as quota check, multi-tenancy prioritization
+	// and maybe more checks in the future before scheduler really does scheduling.
```
RainbowMango, line 334:
```suggestion
	// Scheduling controls whether scheduling should be suspended, the scheduler will pause scheduling and not
	// process resource binding when the value is true and resume scheduling when it's false or nil.
	// This is designed for third-party systems to temporarily pause the scheduling of applications, which enabling
	// manage resource allocation, prioritize critical workloads, etc.
	//
	// It is expected that third-party systems use an admission webhook to suspend scheduling at the time of
	// ResourceBinding creation. Once a ResourceBinding has been scheduled, it cannot be paused afterward, as it may
	// lead to ineffective suspension.
	// +optional
```

RainbowMango: 1. note that scheduling can not be paused after it has been scheduled.
2. +optional

pkg/util/binding.go
```diff
@@ -110,3 +110,19 @@ func RescheduleRequired(rescheduleTriggeredAt, lastScheduledTime *metav1.Time) b
 	}
 	return rescheduleTriggeredAt.After(lastScheduledTime.Time)
 }
+
+// IsBindingSuspendScheduling return whether resource binding is scheduling suspended.
```
RainbowMango, line 114:
```suggestion
// IsBindingSuspendScheduling tells if the scheduling of ResourceBinding is suspended.
```

pkg/util/binding.go
```diff
@@ -110,3 +110,19 @@ func RescheduleRequired(rescheduleTriggeredAt, lastScheduledTime *metav1.Time) b
...
+func IsBindingSuspendScheduling(rb *workv1alpha2.ResourceBinding) bool {
+	if rb == nil || rb.Spec.Suspension == nil || rb.Spec.Suspension.Scheduling == nil {
+		return false
+	}
+	return *rb.Spec.Suspension.Scheduling
+}
+
+// IsClusterBindingSuspendScheduling return whether cluster resource binding is scheduling suspended.
```
RainbowMango, line 122: ditto.

returns or tells.

RainbowMango (commented):
Generally looks good to me except the condition part.
I suggest moving the condition out of this PR and starting another PR to handle it.

Monokaix:
> Generally looks good to me except the condition part. I suggest moving the condition out of this PR and starting another PR to handle it.

You mean seperate the webhook validation to another pr?

RainbowMango:
I mean not introduce the condition by this PR:
```golang
	// SchedulingSuspended represents the condition that the ResourceBinding or ClusterResourceBinding is suspended to schedule.
	SchedulingSuspended string = "SchedulingSuspended"
```

pkg/util/binding.go
Monokaix: done.

Monokaix:
> I mean not introduce the condition by this PR:
> 
> ```go
> 	// SchedulingSuspended represents the condition that the ResourceBinding or ClusterResourceBinding is suspended to schedule.
> 	SchedulingSuspended string = "SchedulingSuspended"
> ```

Upated.

pkg/util/binding.go
Monokaix: done.

pkg/apis/work/v1alpha2/binding_types.go
```diff
@@ -347,6 +357,8 @@ type ResourceBindingStatus struct {
 
 	// Conditions contain the different condition statuses.
 	// +optional
+	// +listType=map
+	// +listMapKey=type
```
RainbowMango, line 361: Why add these? What does it mean?

RainbowMango: It will be useful to manage conditions as a map. but current PR doesn't require it.

Monokaix: Yeah we can track it at https://github.com/karmada-io/karmada/pull/6002, which is needed by `ApplyStatus` method.

Monokaix: removed.

RainbowMango (approved):
/lgtm
/approve

PR #5939, opened 2024-12-11
fix deep equal check failure in CreateOrUpdateWork(), by replace the marshaler

### LINKED ISSUE #5938: deepequal check in helper.CreateOrUpdateWork() always return false, because resource.MarshalJSON() will add \n at the end of JSON
**What happened**:

deepequal check in `helper.CreateOrUpdateWork()` always return false, because `resource.MarshalJSON()` will add `\n` at the end of JSON, but the workloads read from work.Spec.Workload.Manifests never have this `\n`. So everytime this function is called, an update operation will occur.

Fortunately, the same thing doesn't happen to work-status-controller, rb/crb-status-controller. Although they are both using `DeepEqual()` on `runtime.RawExtension`, but the byte slices are obtained from resource interpret webhook.

remove unecessary update operations.

```go
func Test(t *testing.T) {
	mgr, err := ctrl.NewManager(restConf, ctrl.Options{
		Scheme: gclient.NewSchema(),
	})
	if err != nil {
		panic(err)
	}
	go mgr.GetCache().Start(context.TODO())
	time.Sleep(time.Second)

	work := &workv1alpha1.Work{}
	err = mgr.GetClient().Get(context.TODO(), client.ObjectKey{Namespace: "example-ns", Name: "example-name"}, work)
	if err != nil {
		panic(err)
	}
	fmt.Println(work.Spec.Workload.Manifests[0].Raw)

	_ = json.Unmarshal(work.Spec.Workload.Manifests[0].Raw, workload)

	j0, _ := json.Marshal(workload)
	fmt.Println(j0)

	j1, _ := workload.MarshalJSON()
	fmt.Println(j1)
}
```
image

@snowplayfire found this, I'm just providing some possible fixes here. Many thanks to her, I sincerely hope she has a happy holiday. 

This is the root cause why binding controller can’t use `DeepEqual()` to skip the update operation when calling the `ensureWork()`. But for deepequal to work, we need another PR to copy part of the karmada webhook logic to the update operation, which I will complete as soon as possible.

For fixing this issue, one way is just need one line: replace `resource.MarshalJSON()` with `json.Marshal(resource)`, it could work. #5939

But deepequal checks should be used in semantic scenarios, not to judge byte streams. Therefore, if we want to use semantic deepequal checking to fix it, we need to directly judge the unmarshaled workload read from work. #5940

**Environment**:
- Karmada version: any
- Others:

zach593:
**What type of PR is this?**

/kind bug

see #5938

Fixes #5938

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/ctrlutil/work.go
+++ b/pkg/controllers/ctrlutil/work.go
@@ -18,6 +18,7 @@ package ctrlutil
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -48,7 +49,7 @@ func CreateOrUpdateWork(ctx context.Context, client client.Client, workMeta meta
 		}
 	}
 
-	workloadJSON, err := resource.MarshalJSON()
+	workloadJSON, err := json.Marshal(resource)
 	if err != nil {
 		klog.Errorf("Failed to marshal workload(%s/%s), error: %v", resource.GetNamespace(), resource.GetName(), err)
 		return err


XiShanYongYe-Chang (commented):
LGTM, the current pr resolves the deepcopy failure issue and dose not introduce new performance risks.

About the `kind/bug` label. do you expect to sync the current pr to the previous release versions? If we don't need to, maybe we can modify the label.

zach593:
> Do you think there are better optimization opportunities for my case?

@chaosi-zju As we discussed in zoom, for the performance optimization, this PR is not enough, I may need another 2 PRs to complete this goal. Since I need more than one PR, I created #6017 to track it, and we can have further discussion in that issue.

zach593:
BTW, I think we could consider use `json.Marshal()` instead of all of `x.MarshalJSON()`. Because `json.Marshal()` is more general, it will also call `x.MarshalJSON()` internally, and will also perform some other optimizations.

XiShanYongYe-Chang:
> Sure, can you help with that? I'm not familiar with commands with `/`

/remove-kind bug
/kind cleanup

RainbowMango (commented):
/assign
I will take a look this week.

pkg/util/helper/work.go
```diff
@@ -54,7 +55,7 @@ func CreateOrUpdateWork(ctx context.Context, client client.Client, workMeta meta
 		}
 	}
 
-	workloadJSON, err := resource.MarshalJSON()
+	workloadJSON, err := json.Marshal(resource)
```
RainbowMango, line 58: Is there any difference between `encoding/json.Marshal` and `unstructured.MarshalJSON` other than the `/n`?
I am mostly curious about functionality and performance.

RainbowMango: In addition, I think we need to check all of the places where using the `unstructured.MarshalJSON`. 

PS: I don't mean to check and fix them all in this PR. Just remind me.

RainbowMango: Tracked this on #6031, please correct me if anyone feels this is unnecessary.

RainbowMango: I'm starting to worry whether it is safe to replace `unstructured.MarshalJson`, considering that it includes these optimizations.  @zach593 @chaosi-zju What do you think?

Is there any option we can solve this?

RainbowMango: Doubt it. Can you help test it?
I'm afraid the `json.Compact` also removes the space characters from the body, which probably makes the deep copy always return false.

chaosi-zju: > I'm afraid the `json.Compact` also removes the space characters from the body, which probably makes the deep copy always return false.

I did the test as follows:

```go
workloadJSON, err := resource.MarshalJSON()

workloadJSON2, err := json.Marshal(resource)

buffer := new(bytes.Buffer)
if err = json.Compact(buffer, workloadJSON); err != nil {}
workloadJSON3 := buffer.Bytes()

klog.Infof("diff workloadJSON VS workloadJSON2: %+v", cmp.Diff(workloadJSON, workloadJSON2))
klog.Infof("diff workloadJSON VS workloadJSON3: %+v", cmp.Diff(workloadJSON, workloadJSON3))
klog.Infof("diff workloadJSON2 VS workloadJSON3: %+v", cmp.Diff(workloadJSON2, workloadJSON3))
```

the result is: `workloadJSON2` and `workloadJSON3` is equal, and `workloadJSON` is only one more `\n` that the other two.

zach593: https://github.com/golang/go/blob/go1.22.9/src/encoding/json/encode.go#L430
```go
func marshalerEncoder(e *encodeState, v reflect.Value, opts encOpts) {
	if v.Kind() == reflect.Pointer && v.IsNil() {
		e.WriteString("null")
		return
	}
	m, ok := v.Interface().(Marshaler)
	if !ok {
		e.WriteString("null")
		return
	}
	b, err := m.MarshalJSON()
	if err == nil {
		e.Grow(len(b))
		out := e.AvailableBuffer()
		out, err = appendCompact(out, b, opts.escapeHTML)
		e.Buffer.Write(out)
	}
	if err != nil {
		e.error(&MarshalerError{v.Type(), err, "MarshalJSON"})
	}
}
```

https://github.com/golang/go/blob/go1.22.9/src/encoding/json/indent.go#L39
```go
// Compact appends to dst the JSON-encoded src with
// insignificant space characters elided.
func Compact(dst *bytes.Buffer, src []byte) error {
	dst.Grow(len(src))
	b := dst.AvailableBuffer()
	b, err := appendCompact(b, src, false)
	dst.Write(b)
	return err
}
```

`json.Marshal()` also does the same thing as `json.Compact()`. As I understand it, this is the only real difference between `json.Marshal()` and `unstructured.MarshalJSON()`, aside from some behaviors that might affect performance.

chaosi-zju: > `json.Marshal()` also does the same thing as `json.Compact()`. As I understand it, this is the only real difference between `json.Marshal()` and `unstructured.MarshalJSON()`, aside from some behaviors that might affect performance.

You're great 👍 

So, simply understand, `m.MarshalJSON()` is already overwrited, and `json.Marshal()` = `m.MarshalJSON()` + `json.Compact`

RainbowMango (approved):
/lgtm
/approve

chaosi-zju:
I have checked other place where used `(*unstructured.Unstructured).MarshalJSON()`:

- [ ] https://github.com/karmada-io/karmada/blob/80d3d9e007b59e7d1db4cfb139f73d48fc7458d2/pkg/util/overridemanager/overridemanager.go#L273-L278
      **not affect, `obj.MarshalJSON()` generates a extra blank line, the following `patch.Apply(objectJSONBytes)` removes the blank line.**
- [ ] https://github.com/karmada-io/karmada/blob/80d3d9e007b59e7d1db4cfb139f73d48fc7458d2/pkg/resourceinterpreter/customized/webhook/customized.go#L298-L302
      **not affect, `object.MarshalJSON()` generates a extra blank line, the following `patchObj.Apply(objectJSONBytes)` removes the blank line.**
- [ ] https://github.com/karmada-io/karmada/blob/80d3d9e007b59e7d1db4cfb139f73d48fc7458d2/pkg/webhook/work/mutating.go#L79-L88
      **not affect, `workloadObj.MarshalJSON()` generates a extra blank line, the following `json.Marshal(work)` removes the blank line.**
- [ ] https://github.com/karmada-io/karmada/blob/80d3d9e007b59e7d1db4cfb139f73d48fc7458d2/pkg/registry/cluster/storage/aggregate.go#L300-L316
      **a little affect, the result returned by the REST API has an extra blank line, and its impact depends on how the caller uses it, I think it should be improved.**

RainbowMango:
> a little affect, the result returned by the REST API has an extra blank line, and its impact depends on how the caller uses it, I think it should be improved.

I don't think so. There is a reason why `MarshalJSON()` adds the new line at the end, see [the comments here](https://github.com/golang/go/blob/80a2982a801eaedc416d59801ac8fefcf1e4acf5/src/encoding/json/stream.go#L214-L221). This time we focus on checking whether this issue could lead to deepcheck to fail.

chaosi-zju:
> I don't think so. There is a reason why `MarshalJSON()` adds the new line at the end, see [the comments here](https://github.com/golang/go/blob/80a2982a801eaedc416d59801ac8fefcf1e4acf5/src/encoding/json/stream.go#L214-L221). This time we focus on checking whether this issue could lead to deepcheck to fail.

I don't see why that comment is the reason for newline in this REST API response. 

Server-returned JSON data generally lacks a trailing newline. HTTP doesn't require it to mark the end, and data is sent compactly for efficiency.

But doesn't matter, if only focus on deepcheck failure, none such issue left.

PR #5941, opened 2024-12-12
Disable cluster failover by default which should be explicitly enabled by administrators after a fully evaluation

XiShanYongYe-Chang:
Cherry pick of #5899 on release-1.11.
```release-note
The feature `Failover` now has been disabled by default, which should be explicitly enabled to avoid unexpected incidents. 
```

### DIFF
--- a/cmd/controller-manager/app/controllermanager.go
+++ b/cmd/controller-manager/app/controllermanager.go
@@ -592,9 +592,6 @@ func startGracefulEvictionController(ctx controllerscontext.Context) (enabled bo
 }
 
 func startApplicationFailoverController(ctx controllerscontext.Context) (enabled bool, err error) {
-	if !features.FeatureGate.Enabled(features.Failover) {
-		return false, nil
-	}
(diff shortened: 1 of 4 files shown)


cmd/controller-manager/app/controllermanager.go
```diff
@@ -592,9 +592,6 @@ func startGracefulEvictionController(ctx controllerscontext.Context) (enabled bo
 }
 
 func startApplicationFailoverController(ctx controllerscontext.Context) (enabled bool, err error) {
-	if !features.FeatureGate.Enabled(features.Failover) {
-		return false, nil
-	}
```
RainbowMango, line 597: It makes sense to me that this patch won't affect users who are using application failover.

RainbowMango (commented):
/lgtm

cc @chaunceyjiang @whitewindmills

cmd/controller-manager/app/controllermanager.go
whitewindmills: yes, it makes sense.
/lgtm

RainbowMango (approved):
/approve

PR #5959, opened 2024-12-17
Add test case for parsing static value with parseJSONValue func

XiShanYongYe-Chang:
**What type of PR is this?**

/kind cleanup

Add a new test case for parsing static value with parseJSONValue func, refer to https://github.com/karmada-io/karmada/issues/5788#issuecomment-2545862772

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/applicationfailover/common_test.go
+++ b/pkg/controllers/applicationfailover/common_test.go
@@ -226,6 +226,15 @@ func Test_parseJSONValue(t *testing.T) {
 			wantErr: assert.NoError,
 			want:    "2",
 		},
+		{
+			name: "get a static value just as jsonPath write",
+			args: args{
+				rawStatus: []byte(`{"replicas": 2}`),
+				jsonPath:  "true",
+			},
+			wantErr: assert.NoError,
+			want:    "true",
+		},
 		// Build the following test cases in terms of what the function supports (which we don't use now).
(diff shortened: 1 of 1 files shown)


pkg/controllers/applicationfailover/common_test.go
```diff
@@ -226,6 +226,15 @@ func Test_parseJSONValue(t *testing.T) {
 			wantErr: assert.NoError,
 			want:    "2",
 		},
+		{
```
zhzhuang-zju, line 229: It would be nice to add a comment describing what the test case does.

zhzhuang-zju (commented):
thanks, others LGTM

pkg/controllers/applicationfailover/common_test.go
XiShanYongYe-Chang: We can get it from the test case `get a static value`.

XiShanYongYe-Chang: Update with `get a static value just as jsonPath write`

RainbowMango (commented):
/hold
This behavior looks weird and still under discussion. Let's hold for a while.
See https://github.com/karmada-io/karmada/issues/5788#issuecomment-2548420154.

XiShanYongYe-Chang:
From a usage perspective, this does not seem to be our recommended usage. Close it first.
/close

Issue #5961, opened 2024-12-17
[Feature] [Alpha] Application Priority and Preemption


RainbowMango:
**Summary**
Currently, the `karmada-scheduler` operates based on a First-In-First-Out scheduling order. In some real-world scenarios, like AI training, some jobs are more critical than others and require preferential treatment in terms of scheduling and resource allocation.

We need the feature to enable users to specify the importance of an application. Applications with higher priority are scheduled ahead of other applications with lower priority. When member clusters do not have enough capacity for running a high priority application, the scheduler preempts and removes lower priority applications in order to make room for the high priority application.

**Proposal**:

- [x] [Application Priority and Preemption](https://github.com/karmada-io/karmada/pull/4993)

Author:
- @LeonZh0u
- @seanlaii 
- @wengyao04
- @whitewindmills
- @zclyne (Yifan Zhang)

References:
- https://github.com/karmada-io/karmada/issues/4938
- Proposal in Google Doc: https://docs.google.com/document/d/1MixmgLwnmiRrukyFP25JvE2hqiqB3f_JKT12tXiOsbc/edit#
- https://github.com/karmada-io/karmada/issues/802
- https://github.com/karmada-io/karmada/issues/935

**Iteration Tasks  -- Part-1: Scheduling based on application priority**:
- [x] API change: Introduce `priority class` to `PropagationPolicy` (@seanlaii, https://github.com/karmada-io/karmada/pull/5962)
- [x] API change: Introduce `priority` to `ResourceBinding` (@seanlaii, https://github.com/karmada-io/karmada/pull/5963)
- [x] Feature gate: Introduce a feature gate for this feature. (@RainbowMango, https://github.com/karmada-io/karmada/pull/6151)
- [x] Controller: Build application priority when creating `ResourceBinding` (@LeonZh0u )
- [ ] ~~webhook: Introduce validation to restrict PriorityClass modification~~
- [x] scheduler: Introduce PriorityQueue to replace the current RateLimitedQueue. (@whitewindmills, #6149 )
- [x] E2E tests (@seanlaii, #6306)
- [x] Documentation (@seanlaii, https://github.com/karmada-io/website/pull/822)

**Iteration Tasks  -- Part-2: Preemption in case of member clusters lack of capacity**:
- [ ] API change: Introduce preemptionPolicy to `ResourceBinding`
- [ ] TBD

**Welcome Join us**
[...]

PR #5962, opened 2024-12-19
Add priorityClassName in PropagationPolicy & ClusterPropagationPolicy

seanlaii:
**What type of PR is this?**
/kind api-change
/kind feature

Add SchedulePriority to PropagationPolicy & ClusterPropagationPolicy to support flexible priority class resolution strategies:
- KubePriorityClass: Use Kubernetes native PriorityClass
- PodPriorityClass: Use priority from workload's PodTemplate
- FederatedPriorityClass: Reserved for future Karmada-specific priority classes

This enhancement allows users to:
1. Specify which priority class source to use
2. Control resource scheduling priority
3. Maintain consistent priority behavior with Kubernetes

Part of #5961 

- Default source is KubePriorityClass
- Falls back to global default if unspecified
- FederatedPriorityClass is reserved for future implementation
- All new fields are optional to maintain backward compatibility

```release-note
Add SchedulePriority field to PropagationPolicy and ClusterPropagationPolicy:
- Supports multiple priority class sources (Kubernetes, PodTemplate, Federated[future])
- Configurable priority class resolution strategy
- Maintains compatibility with Kubernetes priority behavior
- Optional fields preserve backward compatibility
```


RainbowMango (commented):
/assign
Start working on it!

pkg/apis/policy/v1alpha1/propagation_types.go
```diff
@@ -200,6 +200,33 @@ type PropagationSpec struct {
...
+	// - Uses PriorityClassName from PodTemplate
+	// - Falls back to global default if not found
+	// - If no valid PriorityClass found: Sets priority=0 and preemptionPolicy=Never
+	//
+	// For FederatedPriorityClass:
+	// - Reserved for future use
+	// - Current behavior undefined
+	//
+	// +optional
+	SchedulePriority *SchedulePriority `json:"schedulePriority,omitempty"`
```
RainbowMango, line 229: Maybe we don't have to explain the valid options outside, because they are described inside of `SchedulePriority`. But we can give some overall descriptions, like:
```suggestion
	// SchedulePriority defines how Karmada should resolve the priority and preemption policy
	// for workload scheduling.
	//
	// This setting is particularly useful for controlling the scheduling behavior of offline workloads.
	// By setting a higher or lower priority, users can control which workloads are scheduled first.
	// Additionally, it also allows to specify the preempt policy that higher-priority workloads can
	// preempt lower-priority ones in scenarios where resource contention occurs.
	//
	// Note: This feature is currently in the alpha stage. The priority-based scheduling functionality is
	// controlled by the PriorityBasedScheduling feature gate, and preemption is controlled by the
	// PriorityBasedPreemptiveScheduling feature gate. At this time, only priority-based scheduling is
	// supported, while preemption functionality is not yet available and will be introduced in the future
	// releases as the feature matures.
	//
	// +optional
	SchedulePriority *SchedulePriority `json:"schedulePriority,omitempty"`
```

pkg/apis/policy/v1alpha1/propagation_types.go
```diff
@@ -660,6 +687,59 @@ const (
...
+type SchedulePriority struct {
+	// PriorityClassSource specifies where Karmada should look for the PriorityClass definition.
+	// Available options:
+	// - KubePriorityClass (default): Uses Kubernetes PriorityClass (scheduling.k8s.io/v1)
+	// - PodPriorityClass: Uses PriorityClassName from PodTemplate (PodSpec.PriorityClassName)
+	// - FederatedPriorityClass: Uses Karmada FederatedPriorityClass (not yet implemented)
+	//
+	// +kubebuilder:default="KubePriorityClass"
+	// +kubebuilder:validation:Enum=FederatedPriorityClass;KubePriorityClass;PodPriorityClass
+	// +optional
```
RainbowMango, line 701: Currently, we provide support for `KubePriorityClass` only, right?

Shall we add a comment to `PodPriorityClass` to explain a little bit, something like `not yet implemented`?

In addition, please shorten the `+kubebuilder:validation:Enum` with supported options.

pkg/apis/policy/v1alpha1/propagation_types.go
```diff
@@ -660,6 +687,59 @@ const (
...
+	// For PodPriorityClass:
+	// - Uses PriorityClassName from PodTemplate
+	// - Falls back to global default if not found
+	// - If no valid PriorityClass found: Sets priority=0 and preemptionPolicy=Never
+	//
+	// For FederatedPriorityClass:
+	// - Reserved for future use
+	// - Current behavior undefined
+	//
+	// +optional
```
RainbowMango, line 722:
```suggestion
	// +required
```
Can we restrict this as a required in the current version, just hope to work out a simple version quickly.

@seanlaii In your use cases, will you set a specific PriorityClassName for each workload, or will you rely on the default one?

RainbowMango (commented):
@whitewindmills Would you like to take another look?

pkg/apis/policy/v1alpha1/propagation_types.go
```diff
@@ -200,6 +200,33 @@ type PropagationSpec struct {
...
+	// PriorityClassName behavior:
+	//
+	// For KubePriorityClass:
+	// - If specified: Uses the named Kubernetes PriorityClass
+	// - If empty: Uses cluster's global default PriorityClass
+	// - If neither exists: Sets priority=0 and preemptionPolicy=Never
+	//
+	// For PodPriorityClass:
+	// - Uses PriorityClassName from PodTemplate
+	// - Falls back to global default if not found
```
whitewindmills, line 221: I'm afraid people don't know what the global default is.

pkg/apis/policy/v1alpha1/propagation_types.go
seanlaii: Sure, we currently set a default priority class for every workload if no priority class is set.

pkg/apis/policy/v1alpha1/propagation_types.go
seanlaii: Thanks, updated.

pkg/apis/policy/v1alpha1/propagation_types.go
seanlaii: Changed it to required.

pkg/apis/policy/v1alpha1/propagation_types.go
seanlaii: I changed the description to
```
// If the specified PriorityClass is not found, falls back to the cluster's default PriorityClass
//   (i.e., the PriorityClass marked as the global default in the cluster).
```
WDYT?

pkg/apis/policy/v1alpha1/propagation_types.go
RainbowMango: > Sure, we currently set a default priority class for every workload if no priority class is set.

+1 
I think that making the configuration clearer, and also a relief for the implementation.

RainbowMango (approved):
/lgtm
/approve

I'm going to move this forward to avoid blocking the other tasks.
And I'll double check the comments before releasing, ensuring the comments consist with the minimal version.

PR #5963, opened 2024-12-19
Add Priority and PreemptionPolicy fields in ResourceBinding

seanlaii:
**What type of PR is this?**
/kind api-change
/kind feature

Add SchedulePriority, Priority, and PreemptionPolicy fields in ResourceBinding to support priority-based scheduling. This enables:
- Priority-based scheduling for ResourceBindings
- Preemption control through PreemptionPolicy
- Better control over scheduling order of workloads across clusters

Part of #5961

- Priority field is added as int32 to align with Kubernetes conventions
- PreemptionPolicy supports two modes: Never (default) and PreemptLowerPriority

```release-note
Add SchedulePriority, Priority, and PreemptionPolicy fields to ResourceBinding API to support priority-based scheduling and preemption control.
- Priority: Defines scheduling priority (higher values = higher priority)
- PreemptionPolicy: Controls whether binding can preempt lower priority bindings
```


pkg/apis/work/v1alpha2/binding_types.go
```diff
@@ -159,6 +169,19 @@ type ResourceBindingSpec struct {
 	// This setting applies to all Work objects created under this binding object.
 	// +optional
 	PreserveResourcesOnDeletion *bool `json:"preserveResourcesOnDeletion,omitempty"`
+
+	// Priority represents the scheduling priority of the binding.
+	// Bindings with higher priority values are scheduled ahead of bindings with lower priority values.
+	// If not specified, the priority value is set to 0.
+	// +optional
+	Priority int32 `json:"priority,omitempty"`
```
seanlaii, line 177: Since we don't allow the user to override the priority in the binding, and the priority value can only be derived by the priority class, I use `int32` instead of `*int32` for now. Open to discussion.

RainbowMango (commented):
/assign

pkg/apis/work/v1alpha2/binding_types.go
```diff
@@ -337,6 +341,35 @@ type Suspension struct {
...
+
+	// PreemptionPolicy controls the preemption policy of this binding.
+	// Valid values are:
+	// - Never: (default) prevents the binding from preempting other bindings
+	// - PreemptLowerPriority: allows the binding to preempt lower-priority bindings
+	// +kubebuilder:default=Never
+	// +kubebuilder:validation:Enum=Never;PreemptLowerPriority
+	// +optional
+	PreemptionPolicy PreemptionPolicy `json:"preemptionPolicy,omitempty"`
```
RainbowMango, line 360: How about removing the `PreemptionPolicy` this time? As we are not going to implement this in this release.

seanlaii: Make sense, I removed the preemption part.

PR #5970, opened 2024-12-24
introduce flag --feature-gates to karmadactl register

zhzhuang-zju:
**What type of PR is this?**
/kind feature

introduce flag `--feature-gates` to `karmadactl register` to describe feature gates for alpha/experimental features of `karmada-agent`.

Fixes #

```bash
$ karmadactl register -h
Options:

    --feature-gates='':
        Comma-separated list of key=value pairs that describe feature gates for alpha/experimental features
        of karmada-agent. More info: https://github.com/karmada-io/karmada/blob/master/pkg/features/features.go
```

```release-note
`karmadactl`: introduce flag `--feature-gates` to command `register` to describe feature gates for alpha/experimental features of `karmada-agent`.
```

### DIFF
--- a/pkg/karmadactl/register/register.go
+++ b/pkg/karmadactl/register/register.go
@@ -36,6 +36,7 @@ import (
 	rbacv1 "k8s.io/api/rbac/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	k8srand "k8s.io/apimachinery/pkg/util/rand"
 	"k8s.io/apimachinery/pkg/util/wait"
 	kubeclient "k8s.io/client-go/kubernetes"
@@ -49,6 +50,7 @@ import (
 	"k8s.io/kubectl/pkg/util/templates"
 
 	"github.com/karmada-io/karmada/pkg/apis/cluster/validation"
+	"github.com/karmada-io/karmada/pkg/features"
 	karmadaclientset "github.com/karmada-io/karmada/pkg/generated/clientset/versioned"
 	"github.com/karmada-io/karmada/pkg/karmadactl/options"
 	cmdutil "github.com/karmada-io/karmada/pkg/karmadactl/util"
@@ -196,7 +198,7 @@ func NewCmdRegister(parentCommand string) *cobra.Command {
 	flags.Int32Var(&opts.CertExpirationSeconds, "cert-expiration-seconds", DefaultCertExpirationSeconds, "The expiration time of certificate.")
 	flags.BoolVar(&opts.DryRun, "dry-run", false, "Run the command in dry-run mode, without making any server requests.")
 	flags.StringVar(&opts.ProxyServerAddress, "proxy-server-address", "", "Address of the proxy server that is used to proxy to the cluster.")
-
+	flags.StringVar(&opts.FeatureGates, "feature-gates", "", "Comma-separated list of key=value pairs that describe feature gates for alpha/experimental features of karmada-agent. More info: https://github.com/karmada-io/karmada/blob/master/pkg/features/features.go")
 	return cmd
 }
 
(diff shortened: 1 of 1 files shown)


zhzhuang-zju:
> Timeout waiting for file exist /home/runner/.kube/member-tmp-member2.config
https://github.com/karmada-io/karmada/actions/runs/12478154855/job/34825314108?pr=5970#step:5:710

/retest

RainbowMango (commented):
/assign

pkg/karmadactl/register/register.go
```diff
@@ -196,6 +198,9 @@ func NewCmdRegister(parentCommand string) *cobra.Command {
 	flags.Int32Var(&opts.CertExpirationSeconds, "cert-expiration-seconds", DefaultCertExpirationSeconds, "The expiration time of certificate.")
 	flags.BoolVar(&opts.DryRun, "dry-run", false, "Run the command in dry-run mode, without making any server requests.")
 	flags.StringVar(&opts.ProxyServerAddress, "proxy-server-address", "", "Address of the proxy server that is used to proxy to the cluster.")
+	flags.StringVar(&opts.FeatureGates, "feature-gates", "", ""+
+		"A set of key=value pairs that describe feature gates for alpha/experimental features. "+
+		"Options are:\n"+strings.Join(features.FeatureGate.KnownFeatures(), "\n"))
```
RainbowMango, line 203:
```suggestion
	features.FeatureGate.AddFlag(flags)
```
Can we do that this way, which is the way we did at all the other components, like https://github.com/karmada-io/karmada/blob/6144ea5a9b2fbd8bfd33e651d0901842a9f759de/cmd/controller-manager/app/options/options.go#L235

zhzhuang-zju: I've tried this way, but ran into an issue where I couldn't get the string value of the feature gates when injecting them into the agent's manifest.

RainbowMango: why?

RainbowMango: It is worth noting that, you don't need to introduce an additional field into `CommandRegisterOption` anymore.
```
	FeatureGates string
```

zhzhuang-zju: > why?

https://github.com/karmada-io/karmada/blob/7112723f3de7ba640336202f19dd818e2ca393ca/vendor/k8s.io/component-base/featuregate/feature_gate.go#L122-L153

`features.FeatureGate.AddFlag(flags)` has no method to get the string value of the feature gate. Hard to complete `fmt.Sprintf(‘--feature-gates=%s’, o.FeatureGates)`.

zhzhuang-zju:
I have a question: does it need to verify whether the user-defined featuregate is valid? The method `features.FeatureGate.Validate()` is used to determine if the user-defined feature gate has been added in [pkg/features/features.go]. This leads to the validation result being highly related to the code at the time karmadactl was compiled. However, `karmadactl register` should support different versions of karmada-agent.
For example:
- version v1.12 the supported featuregates are: featuregateA and featuregateB,
- version v1.13 the supported featuregates are: featuregateA,

So, using the v1.13 version of karmadactl to execute the `register` command, setting the flags `--karmada-agent-image docker.io/karmada/karmada-agent:v1.12.0 --feature-gates=featuregateB=true` will result in an error.

zhzhuang-zju:
I researched the mechanism of `kubeadm` for setting component feature gates. In fact, kubeadm does not support directly customizing a component's feature gates. Instead, it allows customizing the flags of components through a [configuration file](https://kubernetes.io/zh-cn/docs/setup/production-environment/tools/kubeadm/control-plane-flags/#apiserver-flags) by specifying the key and corresponding value. Moreover, by examining the [source code](https://github.com/kubernetes/kubernetes/blob/3c229949f992d9f798bcfd9f79dab88c21500c36/cmd/kubeadm/app/apis/kubeadm/validation/validation.go#L727-L737), there seems no special handling when the key is 'feature-gates'. 

The advantage of this approach is that it not only addresses the need to customize feature gates but also allows for the customization of other flags. Furthermore, this mechanism can be reused for the `init` process.

PR #5989, opened 2024-12-28
fix rebalancer auto deleted failed

### LINKED ISSUE #5985: WorkloadRebalancer doesn't get deleted after TTL.
**What happened**:
A WorkloadRebalancer created with a TTL(Time To Live) field in the spec is not getting deleted after TTL has reached.

WorkloadRebalancer should get deleted after TTL.

Create a workloadRebalancer with .spec.ttlSecondsAfterFinished . Verify that it is not deleted after TTL has been passed.

NA

**Possible Cause**:
I debugged this and found few issues with the code of the [workloadrebalancer_controller.go](https://github.com/karmada-io/karmada/blob/master/pkg/controllers/workloadrebalancer/workloadrebalancer_controller.go) file.

- The controller reconciles when the workloadrebalancer is created or the spec changes. [See here](https://github.com/karmada-io/karmada/blob/8fa7c2cc7df1a8d4061f6cfea035cb0625694605/pkg/controllers/workloadrebalancer/workloadrebalancer_controller.go#L59). 
- We are checking for the **rebalance.Status.FinishTime == nil** before setting the rebalancer.Status = *newStatus [here](https://github.com/karmada-io/karmada/blob/8fa7c2cc7df1a8d4061f6cfea035cb0625694605/pkg/controllers/workloadrebalancer/workloadrebalancer_controller.go#L93). So, line 93 always evaluates to true and it never reconciles again and the workload rebalancer doesn't get deleted.

**Environment**:
- Karmada version: 1.11
- Others:

chaosi-zju:
**What type of PR is this?**

/kind bug

fix rebalancer auto deleted failed

Fixes #5985

```release-note
`karmada-controller-manager`: Fixed the bug of WorkloadRebalancer doesn't get deleted after TTL.
```

### DIFF
--- a/pkg/controllers/workloadrebalancer/workloadrebalancer_controller.go
+++ b/pkg/controllers/workloadrebalancer/workloadrebalancer_controller.go
@@ -90,15 +90,12 @@ func (c *RebalancerController) Reconcile(ctx context.Context, req controllerrunt
 		return controllerruntime.Result{}, err
 	}
 
-	if rebalancer.Status.FinishTime == nil {
-		// should never reach here.
-		klog.Errorf("finishTime shouldn't be nil, current status: %+v", rebalancer.Status)
-		return controllerruntime.Result{}, nil
-	}
-
 	// 3. when all workloads finished, judge whether the rebalancer needs cleanup.
 	rebalancer.Status = *newStatus
 	if rebalancer.Spec.TTLSecondsAfterFinished != nil {
+		if rebalancer.Status.FinishTime == nil {
+			return controllerruntime.Result{}, fmt.Errorf("finish time should not be nil")
+		}
 		remainingTTL := timeLeft(rebalancer)
 		if remainingTTL > 0 {
 			return controllerruntime.Result{RequeueAfter: remainingTTL}, nil
--- a/test/e2e/workloadrebalancer_test.go
+++ b/test/e2e/workloadrebalancer_test.go
@@ -193,6 +193,12 @@ var _ = framework.SerialDescribe("workload rebalancer testing", func() {
 				framework.WaitRebalancerDisappear(karmadaClient, rebalancerName)
 			})
 		})
+
+		ginkgo.It("create rebalancer with ttl and verify it can auto clean", func() {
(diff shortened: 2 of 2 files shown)


pkg/controllers/workloadrebalancer/workloadrebalancer_controller.go
```diff
@@ -90,7 +90,7 @@ func (c *RebalancerController) Reconcile(ctx context.Context, req controllerrunt
 		return controllerruntime.Result{}, err
 	}
 
-	if rebalancer.Status.FinishTime == nil {
+	if newStatus.FinishTime == nil {
 		// should never reach here.
```
RainbowMango, line 94: My question is why we need to keep this check since it `should never reach here`?

RainbowMango (commented):
/assign

pkg/controllers/workloadrebalancer/workloadrebalancer_controller.go
RainbowMango: It would be better to move this where referencing the `FinishTime`.

pkg/controllers/workloadrebalancer/workloadrebalancer_controller.go
```diff
@@ -90,15 +90,12 @@ func (c *RebalancerController) Reconcile(ctx context.Context, req controllerrunt
...
-		// should never reach here.
-		klog.Errorf("finishTime shouldn't be nil, current status: %+v", rebalancer.Status)
-		return controllerruntime.Result{}, nil
-	}
-
 	// 3. when all workloads finished, judge whether the rebalancer needs cleanup.
 	rebalancer.Status = *newStatus
 	if rebalancer.Spec.TTLSecondsAfterFinished != nil {
+		if rebalancer.Status.FinishTime == nil {
+			return controllerruntime.Result{}, fmt.Errorf("finishTime shouldn't be nil, current status: %+v", rebalancer.Status)
```
RainbowMango, line 97: Looks better now.
But we should not dump so many status.
```suggestion
			return controllerruntime.Result{}, fmt.Errorf("finish time should not be nil")
```

RainbowMango (approved):
/lgtm
/approve

PR #6016, opened 2025-01-04
support namespaceselectors for cluster propagation policies

grosser:
**What type of PR is this?**

/kind feature

we want to match a clusterpropagationpolicy to resources in only some namespaces

POC, need comments on if this makes sense and ideas on how to get the namespace object

```release-note

```

### DIFF
--- a/pkg/apis/policy/v1alpha1/propagation_types.go
+++ b/pkg/apis/policy/v1alpha1/propagation_types.go
@@ -68,6 +68,10 @@ type PropagationSpec struct {
 	// +kubebuilder:validation:MinItems=1
 	ResourceSelectors []ResourceSelector `json:"resourceSelectors"`
 
+	// NamespaceSelectors used to select resources.
+	// +optional
+	NamespaceSelectors []NamespaceSelector `json:"namespaceSelectors"`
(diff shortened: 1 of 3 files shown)


XiShanYongYe-Chang (commented):
For users, this increases the flexibility of pp use, but increases the complexity of pp and provides the complexity of the system. Can we consider extending a layer above pp and leaving clear resource descriptions to pp?

grosser:
afaik this only makes sense for cpps, but since they share the resource selector with pp I opted to not duplicate everything

XiShanYongYe-Chang:
Hi @grosser, I still recommend influencing cpp at a more abstract level, rather than directly modifying cpp's API to provide similar capabilities.

The more changes that cpp or pp perceives, the more complex the system will be. For example, the current solution requires cpp to pay more attention to namespace resource changes. Can we abstract out these more flexible user settings?

grosser:
no idea how that could work though :D
... closing this since I'm not working on it for a while

XiShanYongYe-Chang:
Thank you for replying to inform you of your progress. If necessary, we will discuss further.

Issue #6017, opened 2025-01-05
Make binding controller don't need to update work when there's no content changes.


zach593:
**What would you like to be added**:

Make binding controller don't need to update work when there's no content changes.

It contains two part:

1. Fixed an issue where deepequal checks would directly compare byte slices, which caused the check to always return false, tracked in #5938
3. Copy karmada-webhook's mutating work logic to `helper.CreateOrUpdateWork()` so deepequal can return true when no actual content has changed.

**Why is this needed**:

When binding controllers, this can help reduce the update requests to the kube-apiserver, thus speeding up controller processing.

It might be possible to add similar logic to the execution controller and the job status controller, but the member cluster webhooks would be a very difficult hurdle to overcome, and I think we can discuss how to deal with it later.

I needed a place to discuss and track related PRs, so I created this issue. Please notify me if this is a misuse.

zach593:
@RainbowMango referenced wrong issue? #6013 is a PR of "expose metrics of karmada-metrics-adapter"

PR #6018, opened 2025-01-05
move CreateOrUpdateWork() and related functions to controllers/ctrlutl

zach593:
**What type of PR is this?**
/kind cleanup

When I followed up on #6017, I ran into a circular reference issue, move this package to pkg/util could help solve it.

Fixes #

```release-note
NONE
```


RainbowMango (commented):
/assign

RainbowMango:
I guess the circular dependency issue is due to the package `pkg/util/helper` importing its upper-level package `pkg/util`, which is not right.
Another option would be moving [these consts](https://github.com/karmada-io/karmada/blob/6144ea5a9b2fbd8bfd33e651d0901842a9f759de/pkg/util/constants.go#L216-L228) to `pkg/util/names`, can you help to confirm if it works?

zach593:
Thank you for your attention @RainbowMango , I'm sorry that I didn't provide the detail about the import chain. 

What I met is I want to call `prune.RemoveIrrelevantFields()` in `helper.CreateOrUpdateWork()` , but `prune.removeJobIrrelevantField()` calls `helper.ConvertToTypedObject()`. so that causes the import cycle.

OTOH, the `pkg/helper` package contains a lot of code that is very close to the business logic layer. Therefore, even if a's direct dependency on  `pkg/helper/util` is removed, other packages referenced by  `pkg/helper` will still depend on `pkg/helper/util` . So, removing the direct dependency of the `pkg/helper` package on `pkg/helper/util` should have little significance.

RainbowMango:
> What I met is I want to call prune.RemoveIrrelevantFields() in helper.CreateOrUpdateWork() ,
> OTOH, the pkg/helper package contains a lot of code that is very close to the business logic layer.

That's the key point. Generally speaking, both `pkg/util` and `pkg/util/helper` provide fundamental functions which should be consumed by the business logic layer, and can not refer to business layer.

The function [CreateOrUpdateWork](https://github.com/karmada-io/karmada/blob/e2827b1b2ef0ed49d8956b2be39a3b03536e759b/pkg/util/helper/work.go#L43) you mentioned it's kind of business logic, also part of controllers. Do you think it should be moved out from `pkg/util/helpers`?

zach593:
> The function [CreateOrUpdateWork](https://github.com/karmada-io/karmada/blob/e2827b1b2ef0ed49d8956b2be39a3b03536e759b/pkg/util/helper/work.go#L43) you mentioned it's kind of business logic, also part of controllers. Do you think it should be moved out from pkg/util/helpers?

I aggre with you, I'll try it and update this PR later.

zach593:
replace title "move pkg/util/helper/unstructured.go to pkg/util" with "move CreateOrUpdateWork() and related functions to controllers/common"

Looks like it could work, although there is still a lot of business logic in `util/helper`, but I just moved the necessary parts according to my needs. @RainbowMango

RainbowMango (commented):
Looks great!

How about naming this new package with the following options?
- ctrlutil 
- ctrlhelper
- ctrlcommon

It's no big deal, just feel the `common` seems to be used everywhere, and I don't like the import aliases. 

I prefer the first one, and then the second.

RainbowMango:
> Looks like it could work, although there is still a lot of business logic in util/helper, but I just moved the necessary parts according to my needs. @RainbowMango

Thanks for the reminder, I will be more careful when people try to put more things to `util/helper` and move things out when they become a blocker.

zach593:
sure, updated the pkg name.

RainbowMango (approved):
/lgtm
/approve

/hold
until the ci gets green
Please feel free to unhold this.

PR #6026, opened 2025-01-08
Replace grpc.DialContext and grpc.WithBlock with grpc.NewClient

seanlaii:
**What type of PR is this?**
/kind deprecation

- `grpc.DialContext` and `grpc.WithBlock` are deprecated
- Updates client connection creation to use modern gRPC patterns
- Leverages DNS resolver for better service discovery
- Maintains backward compatibility while enabling future features

Part of #5796 

Replace `grpc.DialContext` and `grpc.WithBlock` with `grpc.NewClient` to follow the recommended pattern in https://github.com/grpc/grpc-go/blob/master/Documentation/anti-patterns.md.
Since we only allow users to configure the authentication setting of the grpc client without any custom [dial options](https://pkg.go.dev/google.golang.org/grpc#DialOption), I think we should be safe to perform the replacement.

```release-note
Replace `grpc.DialContext` and `grpc.WithBlock` with `grpc.NewClient` since DialContext and WithBlock are deprecated while maintaining the original functionality.
```

### DIFF
--- a/pkg/util/grpcconnection/config.go
+++ b/pkg/util/grpcconnection/config.go
@@ -25,6 +25,7 @@ import (
 	"time"
 
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/connectivity"
 	grpccredentials "google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/credentials/insecure"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
@@ -102,11 +103,7 @@ func (s *ServerConfig) NewServer() (*grpc.Server, error) {
 
 // DialWithTimeOut will attempt to create a client connection based on the given targets, one at a time, until a client connection is successfully established.
 func (c *ClientConfig) DialWithTimeOut(paths []string, timeout time.Duration) (*grpc.ClientConn, error) {
-	opts := []grpc.DialOption{
-		// grpc.WithBlock is deprecated. TODO: Perhaps need to reconsider the approach in a future PR
-		//nolint:staticcheck
-		grpc.WithBlock(),
-	}
+	var opts []grpc.DialOption
 
 	var cred grpccredentials.TransportCredentials
 	if c.ServerAuthCAFile == "" && !c.InsecureSkipServerVerify {
@@ -136,9 +133,7 @@ func (c *ClientConfig) DialWithTimeOut(paths []string, timeout time.Duration) (*
 		}
 		cred = grpccredentials.NewTLS(config)
 	}
-
 	opts = append(opts, grpc.WithTransportCredentials(cred))
-
 	var cc *grpc.ClientConn
 	var err error
 	var allErrs []error
@@ -147,7 +142,7 @@ func (c *ClientConfig) DialWithTimeOut(paths []string, timeout time.Duration) (*
 		if err == nil {
 			return cc, nil
 		}
-		allErrs = append(allErrs, err)
+		allErrs = append(allErrs, fmt.Errorf("dial %s error: %v", path, err))
 	}
 
 	return nil, utilerrors.NewAggregate(allErrs)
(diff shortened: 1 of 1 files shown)


seanlaii:
@zhzhuang-zju please help take a look. Thank you so much!

pkg/util/grpcconnection/config.go
```diff
@@ -143,26 +137,13 @@ func (c *ClientConfig) DialWithTimeOut(paths []string, timeout time.Duration) (*
 	var err error
 	var allErrs []error
 	for _, path := range paths {
-		cc, err = createGRPCConnection(path, timeout, opts...)
-		if err == nil {
-			return cc, nil
+		cc, err = grpc.NewClient(path, opts...)
```
zhzhuang-zju, line 140: @seanlaii I found that `grpc.NewClient` also returns a client when path doesn't exist, with no errors reported. That's not what we expected. 

Can we proceed to refine according to the method of validating grpc configuration mentioned in the [document](https://github.com/grpc/grpc-go/blob/724f450f77a09bade8174e5052625977069aaf81/Documentation/anti-patterns.md?plain=1#L63-L69)?
```md
Some users of Dial use it as a way to validate the configuration of their system. If you wish to maintain this behavior but migrate to NewClient, you can call GetState, then Connect if the state is Idle and WaitForStateChange until the channel is connected. However, if this fails, it does not mean that your configuration was bad - it could also mean the service is not reachable by the client due to connectivity reasons.
```

seanlaii: Sure! Thanks for the advice!

pkg/util/grpcconnection/config.go
```diff
@@ -147,22 +142,37 @@ func (c *ClientConfig) DialWithTimeOut(paths []string, timeout time.Duration) (*
...
-	cc, err := grpc.DialContext(ctx, path, opts...)
+	cc, err := grpc.NewClient(path, opts...)
 	if err != nil {
-		return nil, fmt.Errorf("dial %s error: %v", path, err)
+		return nil, err
 	}
+	defer func() {
+		if err != nil && cc != nil {
+			cc.Close()
+		}
```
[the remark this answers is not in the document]
seanlaii: Yes, you are correct! Thanks for pointing out this! Updated.

pkg/util/grpcconnection/config.go
```diff
@@ -157,12 +152,27 @@ func createGRPCConnection(path string, timeout time.Duration, opts ...grpc.DialO
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
 	defer cancel()
 
-	// grpc.DialContext is deprecated. TODO: Perhaps need to reconsider the approach in a future PR
-	//nolint:staticcheck
-	cc, err := grpc.DialContext(ctx, path, opts...)
+	cc, err := grpc.NewClient(path, opts...)
```
RainbowMango, line 155:
```suggestion
	conn, err = grpc.NewClient(path, opts...)
```
We can use the variable defined in return types, which makes the defer function clear about which error it referring to.

seanlaii: I tried it and encountered an issue: if there is an error, conn will be set to nil, resulting in `nil dereference` when trying to use `conn.Close`. Therefore, I changed it back, which also follows the implementation in [here](https://github.com/grpc/grpc-go/blob/f35fb347b7a7b57b92d25ca2e012dd8359b2a305/clientconn.go#L225-L241). 
Please take a look. Thank you!

RainbowMango: > I tried it and encountered an issue: if there is an error, conn will be set to nil, resulting in nil dereference when trying to use conn.Close in defer function.

Nice finding! Seems you tested the abnormal case, thanks.

RainbowMango (commented):
/lgtm

@chaosi-zju @zhzhuang-zju 
Do you have any further comments?

chaosi-zju:
> Do you have any further comments?

/lgtm, thanks @seanlaii

pkg/util/grpcconnection/config.go
```diff
@@ -102,11 +103,7 @@ func (s *ServerConfig) NewServer() (*grpc.Server, error) {
 
 // DialWithTimeOut will attempt to create a client connection based on the given targets, one at a time, until a client connection is successfully established.
 func (c *ClientConfig) DialWithTimeOut(paths []string, timeout time.Duration) (*grpc.ClientConn, error) {
-	opts := []grpc.DialOption{
-		// grpc.WithBlock is deprecated. TODO: Perhaps need to reconsider the approach in a future PR
-		//nolint:staticcheck
-		grpc.WithBlock(),
-	}
+	var opts []grpc.DialOption
```
zhzhuang-zju, line 106: @seanlaii Thanks for your work! Just a suggestion, function name `DialWithTimeOut` can be considered to be changed to `NewClient`

zhzhuang-zju: Get it, please ignore this comment

RainbowMango (approved):
/approve

RainbowMango:
/lgtm
/approve

PS: re-trigger prow bot

PR #6028, opened 2025-01-08
[WIP]add APIVersion and Kind checks for helper.ToUnstructured()

zach593:
**What type of PR is this?**

/kind bug

This issue blocks me to continue https://github.com/karmada-io/karmada/issues/6017.

In function `helper.ToUnstructured()`, karmada will convert other types of object to `unstructured.Unstructured`, but if object is missed `TypeMeta`, the generated `unstructured.Unstructured` will also missed `APIVersion` and `Kind`, of course in most of the time these are harmful.

Missing `TypeMeta` objects usually come from typed clients or objects that manually created in code.
In kubernetes issue https://github.com/kubernetes/client-go/issues/541 talked about typed client , looks like these guys won't fix this.

This issue affected many tests, and caused some tests to falsely report success. I fixed all the affected code I could see, but there may be still some I haven't noticed. **This may also have an impact in non-test code**, which I need to confirm further.

After fixing all the affected codes, I think adding an APIVersion and Kind checks to helper.ToUnstructured() can prevent the impact from expanding further. However, this check obviously needs to be added after ensuring that all the affected codes have been modified to avoid causing a stronger impact. 

Fixes #

```release-note
NONE
```


RainbowMango (commented):
/assign

pkg/scheduler/event_handler.go
```diff
@@ -130,20 +129,26 @@ func (s *Scheduler) onResourceBindingAdd(obj interface{}) {
...
-	unstructuredNewObj, err := helper.ToUnstructured(cur)
+	oldMeta, err := meta.Accessor(old)
 	if err != nil {
-		klog.Errorf("Failed to transform newObj, error: %v", err)
+		klog.Errorf("Failed to transform oldObj as metav1.Object, error: %v", err)
 		return
 	}
 
-	if unstructuredOldObj.GetGeneration() == unstructuredNewObj.GetGeneration() {
-		klog.V(4).Infof("Ignore update event of object (kind=%s, %s/%s) as specification no change", unstructuredOldObj.GetKind(), unstructuredOldObj.GetNamespace(), unstructuredOldObj.GetName())
```
RainbowMango, line 146: Does this `unstructuredOldObj.GetKind()` always return ""? If yes, is there any bad effect in addition to the log?

RainbowMango:
I'm trying to understand what bug this PR fixes.

> This issue affected many tests, and caused some tests to falsely report success. I fixed all the affected code I could see, but there may be still some I haven't noticed. This may also have an impact in non-test code, which I need to confirm further.

Do you mean tests are invalid? always returning success due to this issue? Can you help to elaborate on it?

pkg/scheduler/event_handler.go
zach593: > Does this unstructuredOldObj.GetKind() always return ""? 

Yes.

Bad effects are minimal because this function only deal with rb and crb, and with or without namespace is enough to explain what is happening here. The reason I needed to change it because I update the `ToUnstructured()` which was obviously affected it.

RainbowMango: OK. I will leave my question to ToUnstructured() part.

pkg/util/helper/unstructured.go
```diff
@@ -62,5 +63,12 @@ func ToUnstructured(obj interface{}) (*unstructured.Unstructured, error) {
...
 		return nil, err
 	}
-	return &unstructured.Unstructured{Object: uncastObj}, nil
+	unstructuredObj := &unstructured.Unstructured{Object: uncastObj}
+	if unstructuredObj.GetKind() == "" {
+		return nil, runtime.NewMissingKindErr(klog.KObj(unstructuredObj).String())
+	}
+	if unstructuredObj.GetAPIVersion() == "" {
+		return nil, runtime.NewMissingVersionErr(klog.KObj(unstructuredObj).String())
+	}
```
RainbowMango, line 72: I still can't understand why this `ToUnstructured` needs to be so picky, requiring apiVersion/kind to be set. In other words, what problem would arise without this? 

Basically, I'm asking `why we need it`, note that missing apiVersion/kind is not `ToUnstructured`'s fault, it just do the conversion job.

zach593: > note that missing apiVersion/kind is not ToUnstructured's fault, it just do the conversion job.

I thought so too. But the problem is that the root cause of this kind of error is too obscure, and even if I fix that test, there is no way to prevent the next person from writing the next invalid test. 
From the perspective of preventing it from happening, if there is a way to completely prevent such errors from happening, I would welcome it no matter what it is, so I chose to modify this function, which seems to be the lowest cost solution.

RainbowMango:
> In kubernetes issue https://github.com/kubernetes/client-go/issues/541 talked about typed client , looks like these guys won't fix this.

By the way, the issue tracks this as well https://github.com/kubernetes/kubernetes/issues/80609

pkg/util/helper/unstructured.go
RainbowMango: > I thought so too. But the problem is that the root cause of this kind of error is too obscure, and even if I fix that test, there is no way to prevent the next person from writing the next invalid test.

I agree. 
Another question, is doing all this just to prevent writing ineffective unit tests? Could business code potentially run into this issue due to oversight?

zach593: > Another question, is doing all this just to prevent writing ineffective unit tests? Could business code potentially run into this issue due to oversight?

According to my statistics, currently only the test code is actually affected. 

Generally, I think new features are fully tested before being merged, so I don't think it is easy for production code to mix in problematic implementations. However, when writing https://github.com/karmada-io/karmada/pull/6076, this problem has puzzled me for a long time. In this PR, many `TypeMeta` additions are to ensure that https://github.com/karmada-io/karmada/pull/6076 can pass the test.

RainbowMango: > This issue blocks me to continue https://github.com/karmada-io/karmada/issues/6017.

So, this issue actually not a blocker of #6017, right?

Anyway, I will look at #6076 and get back to this, in order to understand the affection.

zach593: Specifically, it‘s the tests involved in this PR that blocked #6076.

zach593:
After zoom meeting, we decided to put this PR aside for a while, as we have not yet decided whether to add checks in `helper.ToUnstructured()`.

The rest of the content in this PR will be submitted from other PRs to avoid continuing to block #6076.

PR #6034, opened 2025-01-10
fix the attached binding deletion problem

whitewindmills:
**What type of PR is this?**
/kind bug

I created a deployment that used a PVC as it's dependent resource, and the PVC was able to follow the deployment propagation correctly. When I deleted the PVC and recreated it, I found that the PVC did not follow the deployment propagation because it's ResourceBinding was not created.

dependency_pr

none

```release-note
NONE
```


XiShanYongYe-Chang (commented):
This pr is similar to this logic processing scenario. Can we refer to this scenario: https://github.com/karmada-io/karmada/blob/8313813f767aa6ebd81f7a30c6450f7ea21199d7/pkg/detector/detector.go#L459-L464

whitewindmills:
@XiShanYongYe-Chang PTAL

pkg/dependenciesdistributor/dependencies_distributor.go
```diff
@@ -568,6 +568,13 @@ func (d *DependenciesDistributor) createOrUpdateAttachedBinding(attachedBinding
 	bindingKey := client.ObjectKeyFromObject(attachedBinding)
 	err := d.Client.Get(context.TODO(), bindingKey, existBinding)
 	if err == nil {
+		// If this binding exists and its owner is not the input object, return error and let garbage collector
+		// delete this binding and try again later. See https://github.com/karmada-io/karmada/issues/2090.
+		if ownerRef := metav1.GetControllerOfNoCopy(existBinding); ownerRef != nil && ownerRef.UID != attachedBinding.OwnerReferences[0].UID {
```
XiShanYongYe-Chang, line 573: From the current function, can the 0 subscript be checked to prevent panic?

pkg/dependenciesdistributor/dependencies_distributor.go
```diff
@@ -568,6 +568,13 @@ func (d *DependenciesDistributor) createOrUpdateAttachedBinding(attachedBinding
 	bindingKey := client.ObjectKeyFromObject(attachedBinding)
 	err := d.Client.Get(context.TODO(), bindingKey, existBinding)
 	if err == nil {
+		// If this binding exists and its owner is not the input object, return error and let garbage collector
+		// delete this binding and try again later. See https://github.com/karmada-io/karmada/issues/2090.
```
XiShanYongYe-Chang, line 572: Do we need to change the issue info to the current pr?

pkg/dependenciesdistributor/dependencies_distributor_test.go
```diff
@@ -2617,6 +2660,12 @@ func Test_createOrUpdateAttachedBinding(t *testing.T) {
 						Namespace:       "test",
 						ResourceVersion: "1000",
 						Labels:          map[string]string{"foo": "bar"},
+						OwnerReferences: []metav1.OwnerReference{
+							{
+								UID:        "foo-bar",
+								Controller: pointer.Bool(true),
+							},
+						},
```
XiShanYongYe-Chang, line 2668: Maybe we can add a related case.

XiShanYongYe-Chang (commented):
Thanks~

pkg/dependenciesdistributor/dependencies_distributor.go
whitewindmills: seems a bit redundant. `attachedBinding` is created by the func `buildAttachedBinding`, which guarantees this. and this func `createOrUpdateAttachedBinding` is not exported.

pkg/dependenciesdistributor/dependencies_distributor_test.go
whitewindmills: yes

pkg/dependenciesdistributor/dependencies_distributor.go
XiShanYongYe-Chang: Okay, but it looks worried at first, lest it lead to panic.

whitewindmills:
@XiShanYongYe-Chang PTAL

XiShanYongYe-Chang (commented):
Thanks~
/lgtm
Do you think we need to cherry-pick this patch to the previous branch?

/cc @RainbowMango

whitewindmills:
> Do you think we need to cherry-pick this patch to the previous branch?

yes, it's necessary.

whitewindmills:
kindly ping @RainbowMango

XiShanYongYe-Chang:
> yes, it's necessary.

Can you help add a release note?

RainbowMango (commented):
/assign

RainbowMango (approved):
/approve

RainbowMango:
I added a release note to this PR.

whitewindmills:
@RainbowMango @XiShanYongYe-Chang 
Which branch do you think I should start cherry-picking from?

RainbowMango:
All maintains branches, release-1.12, release-1.11, and release-1.10.

PR #6044, opened 2025-01-14
feat: add operator build_info metrics and go runtime metrics

dongjiang1989:
**What type of PR is this?**

/kind feature
More metrics info for prometheus and pod hpa

Fixes #

```release-note
add operator build_info metrics and go runtime metrics
```


RainbowMango:
Thanks @dongjiang1989, can you help to explain why these metrics are necessary for you?
Also, can you give a list of metrics ?

dongjiang1989:
> Thanks @dongjiang1989, can you help to explain why these metrics are necessary for you? Also, can you give a list of metrics ?

Thanks @RainbowMango
When inappropriate GOGC, GOMAXPROCS and GOMEMLIMIT are set in the container, it will cause varying degrees of `abnormalities' in go's scheduling, memory usage, GC and distributed lock time. 
Therefore, multiple `golang runtime` metrics are needed for further monitoring and alarming.

`build_info` for image upgrade monitoring.

List Metrics:
```
# HELP karmada_operator_build_info A metric with a constant '1' value labeled by version, revision, branch, goversion from which karmada_operator was built, and the goos and goarch for the build.
# TYPE karmada_operator_build_info gauge
karmada_operator_build_info{branch="",goarch="amd64",goos="darwin",goversion="go1.23.3",revision="851c785640e5d66b653f31d5a5ece7f7288116c8-modified",tags="unknown",version=""} 1
# HELP go_sync_mutex_wait_total_seconds_total Approximate cumulative time goroutines have spent blocked on a sync.Mutex, sync.RWMutex, or runtime-internal lock. This metric is useful for identifying global changes in lock contention. Collect a mutex or block profile using the runtime/pprof package for more detailed contention data.
# TYPE go_sync_mutex_wait_total_seconds_total counter
go_sync_mutex_wait_total_seconds_total 0.000339344
# HELP go_sched_gomaxprocs_threads The current runtime.GOMAXPROCS setting, or the number of operating system threads that can execute user-level Go code simultaneously.
# TYPE go_sched_gomaxprocs_threads gauge
go_sched_gomaxprocs_threads 8
# HELP go_sched_goroutines_goroutines Count of live goroutines.
# TYPE go_sched_goroutines_goroutines gauge
go_sched_goroutines_goroutines 19
# HELP go_sched_latencies_seconds Distribution of the time goroutines have spent in the scheduler in a runnable state before actually running. Bucket counts increase monotonically.
# TYPE go_sched_latencies_seconds histogram
go_sched_latencies_seconds_bucket{le="6.399999999999999e-08"} 25
go_sched_latencies_seconds_bucket{le="6.399999999999999e-07"} 27
go_sched_latencies_seconds_bucket{le="7.167999999999999e-06"} 41
go_sched_latencies_seconds_bucket{le="8.191999999999999e-05"} 51
go_sched_latencies_seconds_bucket{le="0.0009175039999999999"} 52
go_sched_latencies_seconds_bucket{le="0.010485759999999998"} 52
go_sched_latencies_seconds_bucket{le="0.11744051199999998"} 52
go_sched_latencies_seconds_bucket{le="+Inf"} 52
go_sched_latencies_seconds_sum 0.000162688
go_sched_latencies_seconds_count 52
```

RainbowMango:
Thanks for the clarification. I will take a look and get back to you.

RainbowMango:
By the way, are you building [some services](https://stack.kubeservice.cn/) based on Karmada?

RainbowMango:
Is it possible to introduce your platform on the Karmada [community meeting](https://github.com/karmada-io/karmada?tab=readme-ov-file#meeting)?

RainbowMango (commented):
/assign

dongjiang1989:
> Is it possible to introduce your platform on the Karmada [community meeting](https://github.com/karmada-io/karmada?tab=readme-ov-file#meeting)?

Very willing to share it.

RainbowMango:
That would be great!
Note that, the latest English Meeting will be on 2025-01-21, Chinese Meeting will at 2025-02-11. 
Feel free to add an agenda to [the Meeting Notes](http://bit.ly/karmada-meeting-notes).

operator/cmd/operator/app/operator.go
```diff
@@ -110,6 +114,28 @@ func Run(ctx context.Context, o *options.Options) error {
 		return err
 	}
 
+	if err := manager.AddReadyzCheck("readyz", healthz.Ping); err != nil {
+		klog.Errorf("Unable to set up ready check: %v", err)
+		return err
+	}
```
RainbowMango, line 120: Can you explain why you need the readyz in addition to the healthy check?

operator/cmd/operator/app/operator.go
```diff
@@ -110,6 +114,28 @@ func Run(ctx context.Context, o *options.Options) error {
...
+	ctrlmetrics.Registry.MustRegister(
+		collectors.NewGoCollector(
+			collectors.WithGoCollectorRuntimeMetrics(
+				collectors.MetricsGC,
+				collectors.MetricsScheduler,
+				collectors.MetricsMemory,
+				collectors.GoRuntimeMetricsRule{Matcher: regexp.MustCompile(`^/sync/.*`)},
+			),
+		),
+	)
```
RainbowMango, line 134: Is the intention here to select only useful metrics? Unregister the default ones and then add the needed one ?

operator/cmd/operator/app/operator.go
```diff
@@ -110,6 +114,28 @@ func Run(ctx context.Context, o *options.Options) error {
...
+	ctrlmetrics.Registry.MustRegister(
+		prometheusversion.NewCollector("karmada_operator"),
+	)
```
RainbowMango, line 137: I bet this meric would be useful, but I doubt it works as expected, because it tries to get the version info from the [prometheus version package](https://github.com/karmada-io/karmada/blob/3f919e8c479b5aeb65f5852bd031a8d9ccaee25b/vendor/github.com/prometheus/common/version/info.go#L27) and we don't set version to that. Have you tested it?

operator/cmd/operator/app/operator.go
dongjiang1989: Thanks @RainbowMango 
`readyz` and `healthz` correspond to `readyzHandler` and `healthzHandler` in controller-runtime. 
Hmmm.... 🤔️ It may be necessary to refine it? 

ref: https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/manager/internal.go#L97-L101

operator/cmd/operator/app/operator.go
dongjiang1989: Thanks @RainbowMango 
In `collectors.NewGoCollector()` just basic metrics(e.g. "go_goroutines" and "go_info",  etc).  Add `MetricsGC`, `MetricsScheduler`,  `MetricsMemory` and `GoRuntimeMetricsRule` more info, and it includes basic metrics.

operator/cmd/operator/app/operator.go
dongjiang1989: Thanks @RainbowMango 
Got it.

I miss `gflags` to add it.

operator/cmd/operator/app/operator.go
RainbowMango: Get it. Thanks. I never looked into these metrics before, I will test it on my side.

operator/cmd/operator/app/operator.go
dongjiang1989: > Should we introduce the two checks on all the other karmada components? I would suggest we can start another PR or issue to discuss that topic, and let this PR focus on metrics thing.

Got it.
Remove it frist.

dongjiang1989:
> That would be great! Note that, the latest English Meeting will be on 2025-01-21, Chinese Meeting will at 2025-02-11. Feel free to add an agenda to [the Meeting Notes](http://bit.ly/karmada-meeting-notes).

@RainbowMango I don't have editing permissions. Maybe can share it  at 2025-02-11.

RainbowMango:
Echo from the notes from meeting notes:
> Please know:
By joining the google groups you will be able to edit the meeting notes.
Join google group mailing list: https://groups.google.com/forum/#!forum/karmada

Please join the mail group and all members from the mail group can edit this doc.

RainbowMango (commented):
Given we already have some metrics related to gc, goroutine, like:
```
# HELP go_gc_duration_seconds A summary of the pause duration of garbage collection cycles.
# TYPE go_gc_duration_seconds summary
go_gc_duration_seconds{quantile="0"} 3.0146e-05
go_gc_duration_seconds{quantile="0.25"} 4.6354e-05
go_gc_duration_seconds{quantile="0.5"} 5.3922e-05
go_gc_duration_seconds{quantile="0.75"} 0.000103302
go_gc_duration_seconds{quantile="1"} 0.000227208
go_gc_duration_seconds_sum 0.001537905
go_gc_duration_seconds_count 18
# HELP go_goroutines Number of goroutines that currently exist.
# TYPE go_goroutines gauge
go_goroutines 43
# HELP go_info Information about the Go environment.
# TYPE go_info gauge
go_info{version="go1.22.9"} 1
# HELP go_memstats_alloc_bytes Number of bytes allocated and still in use.
# TYPE go_memstats_alloc_bytes gauge
go_memstats_alloc_bytes 6.101576e+06
```

I wound like to understand the list of metrics introduced or removed by this PR and the reason behind these changes.
Additionally, If this change is accepted, my intension is to extend these metrics to all other components, not just the `karmada-operator`.

operator/cmd/operator/app/operator.go
```diff
@@ -110,6 +113,23 @@ func Run(ctx context.Context, o *options.Options) error {
...
+	// Unregister default NewGoCollector
+	ctrlmetrics.Registry.Unregister(collectors.NewGoCollector())
+
+	ctrlmetrics.Registry.MustRegister(
+		collectors.NewGoCollector(
+			collectors.WithGoCollectorRuntimeMetrics(
+				collectors.MetricsGC,
+				collectors.MetricsScheduler,
+				collectors.MetricsMemory,
+				collectors.GoRuntimeMetricsRule{Matcher: regexp.MustCompile(`^/sync/.*`)},
```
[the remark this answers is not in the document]
dongjiang1989: Git it.
Add comments for each metric

BTW,  I will extend these metrics to all other components.

pkg/version/version.go
```diff
@@ -19,13 +19,17 @@ package version
...
 	"runtime"
+
+	"github.com/prometheus/client_golang/prometheus"
 )
 
 // Info contains versioning information.
 type Info struct {
 	GitVersion   string `json:"gitVersion"`
 	GitCommit    string `json:"gitCommit"`
+	GitRevision  string `json:"gitRevision"`
```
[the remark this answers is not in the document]
dongjiang1989: `GitCommit` for git commit ID from `git rev-parse HEAD` and `GitRevision` for commit short ID from `git rev-parse --short HEAD`.  Short ID is more friendly to CTL version help info and prometheus tag value.

RainbowMango:
> BTW, I will extend these metrics to all other components.

Hold on, we can postpone this until we make the final decision.

pkg/version/version.go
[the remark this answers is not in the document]
RainbowMango: I didn't fully understand what you meant by `more friendly`, besides being displayed in metrics, where else would it be used? 

I think there is almost no difference between short and long commit IDs, for instance:
```
karmada-operator_build_info{revision="696daab07e78ede7e04616be3d57282b736fdcc3"...} 1
```
VS
```
karmada-operator_build_info{revision="696daab07"...} 1
```

I'm asking because the length of short commit ID might be changed over time, which is computed based on the approximate number of packed objects in your repository, it might increase in the future. I'm not sure if this would break anything. For example, the length of short commit ID in Karmada and Kubernetes is different:
- In karmada: the length is `9`, like `696daab07`
- In Kubernetes: the length is `11`, like `bae2c62678d`.

Echo relevant docs from git:
https://git-scm.com/docs/git-rev-parse
https://git-scm.com/docs/git-config

> --short[=<length>]
Same as --verify but shortens the object name to a unique prefix with at least length characters. The minimum length is 4, the default is the effective value of the core.abbrev configuration variable (see [git-config[1]](https://git-scm.com/docs/git-config)).

> core.abbrev
Set the length object names are abbreviated to. If unspecified or set to "auto", an appropriate value is computed based on the approximate number of packed objects in your repository, which hopefully is enough for abbreviated object names to stay unique for some time. If set to "no", no abbreviation is made and the object names are shown in their full length. The minimum length is 4.

RainbowMango:
I left my comments on https://github.com/karmada-io/karmada/pull/6044#discussion_r1953734588 after exploring some docs about the short commit id.

pkg/version/version.go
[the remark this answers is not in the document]
dongjiang1989: > I think there is almost no difference between short and long commit IDs, for instance:
> 
> ```
> karmada-operator_build_info{revision="696daab07e78ede7e04616be3d57282b736fdcc3"...} 1
> ```
> 
> VS
> 
> ```
> karmada-operator_build_info{revision="696daab07"...} 1
> ```
> 

Hmmm...  Short commit ID makes it easier to display and manually check in Grafana monitoring list.

And this Short commit ID is released with the version and is determined as a static value. No more conflicts and disruptions.

Such examples include:
`Container ID 3d8b34cd6e99b815b3b4ddf87c32a608a2611b860e9967dfe5ddc287b8f96636` and `Container Short ID 3d8b34cd6e99`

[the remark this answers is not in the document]
RainbowMango: OK. Since it is primarily used for human manual inspection, even if the length changes, it doesn't make any side effects.

pkg/version/base.go
```diff
@@ -26,6 +26,7 @@ var (
 	gitVersion   = "v0.0.0-master"
 	gitCommit    = "unknown" // sha1 from git, output of $(git rev-parse HEAD)
 	gitTreeState = "unknown" // state of git tree, either "clean" or "dirty"
+	gitRevision  = "unknown" // short sha1 for git, output of $(git rev-parse --short HEAD)
```
RainbowMango, line 29:
```suggestion
	gitAbbreviativeCommit = "unknown" // short sha1 from git, output of $(git rev-parse --short HEAD)
```
Follow the concept `abbreviative` from git. Which also removes the ambiguity between `gitVersion` and `gitRevision`.
What do you think?

hack/util.sh
```diff
@@ -707,9 +707,15 @@ function util::get_version() {
...
 
+function util::get_revision() {
+  git rev-parse --short HEAD
+}
+
 function util::version_ldflags() {
   # Git information
   GIT_VERSION=$(util::get_version)
+  # Git revision
+  GIT_REVISION=$(util::get_revision)
```
RainbowMango, line 718:
```suggestion
  GIT_ABBREVIATIVE_COMMIT=$(git rev-parse --short HEAD)
```

pkg/version/base.go
dongjiang1989: Thanks @RainbowMango 
Got it
Change variable name done.

hack/util.sh
```diff
@@ -707,9 +707,15 @@ function util::get_version() {
...
 
+function util::get_revision() {
+  git rev-parse --short HEAD
+}
+
 function util::version_ldflags() {
   # Git information
   GIT_VERSION=$(util::get_version)
+  # Git revision
+  GIT_ABBREVIATIVE_COMMIT=$(util::get_revision)
```
RainbowMango, line 718:
```suggestion
function util::version_ldflags() {
  # Git information
  GIT_VERSION=$(util::get_version)
  GIT_ABBREVIATIVE_COMMIT=$(git rev-parse --short HEAD)
```
How about inlining it?
PS: I can't tell much difference between `revision` and `version`.

pkg/version/version.go
```diff
@@ -41,12 +44,38 @@ func (info Info) String() string {
...
+		Platform:              fmt.Sprintf("%s/%s", runtime.GOOS, runtime.GOARCH),
 	}
 }
+
+// NewCollector returns a collector that exports metrics about current version
+// information.
+func NewCollector(program string) prometheus.Collector {
+	return prometheus.NewGaugeFunc(
+		prometheus.GaugeOpts{
+			Namespace: program,
```
RainbowMango, line 63: Maybe we can omit the component name?
Because the metric emitted from a component indicates it's build info.

pkg/version/version.go
```diff
@@ -41,12 +44,38 @@ func (info Info) String() string {
...
+			Name:      "build_info",
+			Help: fmt.Sprintf(
+				"A metric with a constant '1' value labeled by version, revision, branch, goversion from which %s was built, and the goos and goarch for the build.",
+				program,
+			),
+			ConstLabels: prometheus.Labels{
+				"version":   Get().GitVersion,
+				"revision":  Get().GitAbbreviativeCommit,
```
RainbowMango, line 71:
```suggestion
				"commit":  Get().GitCommit,
				"abbr-commit":  Get().GitAbbreviativeCommit,
```

RainbowMango (commented):
Thanks for the quick response, generally looks good to me for the short commit part except a few more nits.

I will run another look for the metrcis.

pkg/version/version.go
dongjiang1989: Hmmmm....🤔  @RainbowMango  Probably best to stay consistent with prometheus's specification.

Ref: https://github.com/prometheus/client_golang/blob/main/prometheus/collectors/version/version.go#L35-L42

pkg/version/version.go
dongjiang1989: The metrcs specification defines different metrics names, and is not distinguished by different tags of the same metrics name.
ref: https://github.com/prometheus/client_golang/blob/main/prometheus/collectors/version/version.go#L28-L30

pkg/version/version.go
RainbowMango: I apologize for not explaining why I suggested those changes.
I think the most important aspect of naming is clarity, and I feel the terms `version` and `reversion` are ambiguous, which could lead to misunderstandings. For example, I can't tell their difference in the context of Prometheus. (Can you give an example of the build info metric from Prometheus?)

Regarding your point about maintaining consistency with Prometheus, I'd love to, but Karmada might emit both commit and abbreviate commit, we should differentiate them in terms of naming to avoid confusion.

operator/cmd/operator/app/operator.go
```diff
@@ -110,6 +113,34 @@ func Run(ctx context.Context, o *options.Options) error {
...
+				// go runtime sync lock metrics. (e.g. `go_sync_mutex_wait_total_seconds_total`
+				// means Approximate cumulative time goroutines have spent blocked on a sync.Mutex, sync.RWMutex, or runtime-internal lock)
+				collectors.GoRuntimeMetricsRule{Matcher: regexp.MustCompile(`^/sync/.*`)},
+			),
+		),
+	)
+	// `karmada_operator_build_info` metrics for operator version upgrade
+	ctrlmetrics.Registry.MustRegister(
+		version.NewCollector("karmada_operator"),
+	)
```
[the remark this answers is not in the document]
dongjiang1989: Yes. update controller-runtime to v0.19.6

pkg/version/version.go
dongjiang1989: example:
```bash
prometheus_operator_build_info{branch="refs/tags/v0.67.1", goarch="amd64", goos="linux", goversion="go1.20.6", instance="xxx.xxx.xxx.xxx:10250", job="kube-prometheus-stack-operator", namespace="monitoring", pod="kube-prometheus-stack-operator-84fbdc5578-wxqll", revision="ce782eafa", tags="unknown", version="0.67.1"}
````
Keep `branch`, `revision`, `version` consistent with prometheus's specification.

File: pkg/version/base.go [hunk 12]
```diff
@@ -23,9 +23,10 @@ package version
 // version for ad-hoc builds (e.g. `go build`) that cannot get the version
 // information from git.
 var (
-	gitVersion   = "v0.0.0-master"
-	gitCommit    = "unknown" // sha1 from git, output of $(git rev-parse HEAD)
-	gitTreeState = "unknown" // state of git tree, either "clean" or "dirty"
+	gitVersion            = "v0.0.0-master"
+	gitCommit             = "unknown" // sha1 from git, output of $(git rev-parse HEAD)
+	gitTreeState          = "unknown" // state of git tree, either "clean" or "dirty"
+	gitAbbreviativeCommit = "unknown" // short sha1 from git, output of $(git rev-parse --short HEAD)
```
RainbowMango, line 29: The Copilot gives one comment regarding the field name `gitAbbreviativeCommit` as follows:

> The field name 'GitAbbreviativeCommit' might be less clear than alternatives such as 'GitShortCommit'. Consider renaming for clarity and to align with common versioning terminology.

I'm sold by Copilot, and feel `GitShortCommit` might be more clear, and apologize for my back-and-forth.
Let me know how you think?

pkg/version/version.go
RainbowMango: An ideal metric in my mind would look something like this:
```golang
// NewBuildInfoCollector returns a collector that exports metrics about the current version
// information.
func NewBuildInfoCollector() prometheus.Collector {
	return prometheus.NewGaugeFunc(
		prometheus.GaugeOpts{
			Name: "component_build_info",
			Help: "Component build metadata exposed as labels with a constant value of 1.",
			ConstLabels: prometheus.Labels{
				"git_version":      Get().GitVersion,
				"git_commit":       Get().GitCommit,
				"git_short_commit": Get().GitShortCommit,
				"git_tree_state":   Get().GitTreeState,
				"build_date":       Get().BuildDate,
				"go_version":       Get().GoVersion,
				"compiler":         Get().Compiler,
				"platform":         Get().Platform,
			},
		},
		func() float64 { return 1 },
	)
}
```

1. I don't think the component name needed to be as `namespace`, because the component emitting the metric inherently `owns` it.
2. All build metadata information is clearly expressed through standardized labels to avoid ambiguity.
3. I'd love to follow the mature practice like you mentioned above, but I don't think that fit us, mostly because we want more clear label names.

RainbowMango:
Two more comments are:
1. I'd love to put the new build info metric in package `pkg/metrics` (maybe a new file `version.go`) where we manage metrics across the project.
2. Squash the commits once we are ready to go.

dongjiang1989:
> Two more comments are:
> 
> 1. I'd love to put the new build info metric in package `pkg/metrics` (maybe a new file `version.go`) where we manage metrics across the project.
> 2. Squash the commits once we are ready to go.

Thanks @RainbowMango  Please re-check it

pkg/metrics/version.go
```diff
@@ -0,0 +1,51 @@
...
+
+	"github.com/karmada-io/karmada/pkg/version"
+)
+
+// NewCollector returns a collector that exports metrics about current version
+// information.
+func NewCollector(program string) prometheus.Collector {
+	return prometheus.NewGaugeFunc(
+		prometheus.GaugeOpts{
+			Namespace: program,
```
RainbowMango, line 32: @CharlesQQ  Do you think using the component name as a prefix for the metrics is necessary?

pkg/metrics/version.go
```diff
@@ -0,0 +1,51 @@
...
+	"fmt"
+
+	"github.com/prometheus/client_golang/prometheus"
+
+	"github.com/karmada-io/karmada/pkg/version"
+)
+
+// NewCollector returns a collector that exports metrics about current version
+// information.
+func NewCollector(program string) prometheus.Collector {
```
RainbowMango, line 29:
```suggestion
func NewBuildInfoCollector(program string) prometheus.Collector {
```
At least we can use a more straightforward name which I suggested at https://github.com/karmada-io/karmada/pull/6044#discussion_r1976809533.

pkg/metrics/version.go
CharlesQQ: No, it is usually used in combination with metrics and prometheus; when prometheus aggregates data, users usually add specific labels related to components from different sources.

pkg/metrics/version.go
dongjiang1989: Fixed

RainbowMango: Seems not :)

dongjiang1989: > Seems not :)

Done

RainbowMango:
Generally looks good to me. (though, I still can not understand why it is necessary to have the component name `karmada_operator` as the prefix of the metric name.)
@jabellard , I wonder if you use metrics for monitoring? Do you want to take another look?

dongjiang1989:
> Generally looks good to me. (though, I still can not understand why it is necessary to have the component name `karmada_operator` as the prefix of the metric name.) @jabellard , I wonder if you use metrics for monitoring? Do you want to take another look?

Thanks @RainbowMango 

Some message: 
```
foobar_build_info (for a pseudo-metric that provides `metadata` about the running binary)
```
from  [Prometheus Metric and label naming](https://prometheus.io/docs/practices/naming/)

RainbowMango:
Thanks for the info. 
I'm not concerned with the naming practice. In my opinion, we are not violating the Prometheus naming practice you mentioned above, because the build info is kind of a generic metric and can be shared by all of the components, so, another approach is we can name the metric with `karamda_build_info`.

Compared to the naming rule, I'm more curious about how you use the metric, especially what issues might raise if component names are not used as prefixes for the metric.

jabellard:
> Generally looks good to me. (though, I still can not understand why it is necessary to have the component name `karmada_operator` as the prefix of the metric name.)
> @jabellard , I wonder if you use metrics for monitoring? Do you want to take another look?

Observabity is very important for us as we want to be informed of events such as a control plane failing to reach a registered member cluster, amongst other things, via alerts/alarms. Currently,  we're not using the metrics from the operator for anything, but maybe will make use of them at some point.

RainbowMango (commented):
/lgtm

Now waiting for CI signals.

RainbowMango (approved):
/lgtm
/approve

PR #6052, opened 2025-01-16
config security context

zhzhuang-zju:
**What type of PR is this?**
/kind cleanup

A [security context](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) defines privilege and access control settings for a Pod or Container. More details can refer to https://www.dynatrace.com/engineering/blog/kubernetes-security-part-3-security-context/

This PR restricted the component's security context, including,
- set `allowPrivilegeEscalation` to false
- set `privileged` to false,
- configured the `seccompProfile` to `RuntimeDefault`

Regarding `runAsUser` and `runAsNonRoot`, unfortunately, imposing these restrictions resulted in abnormal functionality of the component, so they could not be applied. As for `readOnlyRootFilesystem`, implementing this would reduce convenience during troubleshooting. Given that the manifests files maintained under the `artifacts` directory are primarily used by developers, this PR decides not to set this option for now.

Fixes #

For other installation methods, this configuration can be adapted later, and it should provide users with the ability to customize security context.

```release-note

```

### DIFF
--- a/artifacts/deploy/karmada-aggregated-apiserver.yaml
+++ b/artifacts/deploy/karmada-aggregated-apiserver.yaml
@@ -24,6 +24,9 @@ spec:
         - name: karmada-aggregated-apiserver
           image: docker.io/karmada/karmada-aggregated-apiserver:latest
           imagePullPolicy: IfNotPresent
+          securityContext:
+            allowPrivilegeEscalation: false
+            privileged: false
           command:
             - /bin/karmada-aggregated-apiserver
             - --kubeconfig=/etc/karmada/config/karmada.config
@@ -77,6 +80,9 @@ spec:
         - name: etcd-client-cert
           secret:
             secretName: karmada-aggregated-apiserver-etcd-client-cert
+      securityContext:
+        seccompProfile:
+          type: RuntimeDefault
 ---
 apiVersion: v1
 kind: Service
--- a/artifacts/deploy/karmada-controller-manager.yaml
+++ b/artifacts/deploy/karmada-controller-manager.yaml
@@ -21,6 +21,9 @@ spec:
         operator: Exists
       containers:
         - name: karmada-controller-manager
+          securityContext:
+            allowPrivilegeEscalation: false
+            privileged: false
           image: docker.io/karmada/karmada-controller-manager:latest
           imagePullPolicy: IfNotPresent
           command:
@@ -53,3 +56,6 @@ spec:
         - name: karmada-config
           secret:
             secretName: karmada-controller-manager-config
+      securityContext:
+        seccompProfile:
+          type: RuntimeDefault
--- a/artifacts/deploy/karmada-descheduler.yaml
+++ b/artifacts/deploy/karmada-descheduler.yaml
@@ -21,6 +21,9 @@ spec:
           operator: Exists
       containers:
         - name: karmada-descheduler
+          securityContext:
+            allowPrivilegeEscalation: false
+            privileged: false
           image: docker.io/karmada/karmada-descheduler:latest
           imagePullPolicy: IfNotPresent
           command:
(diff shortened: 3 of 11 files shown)


artifacts/deploy/karmada-controller-manager.yaml
```diff
@@ -53,3 +56,6 @@ spec:
         - name: karmada-config
           secret:
             secretName: karmada-controller-manager-config
+      securityContext:
+        seccompProfile:
+          type: RuntimeDefault
```
RainbowMango, line 61: end of line

artifacts/deploy/karmada-descheduler.yaml
```diff
@@ -58,3 +61,6 @@ spec:
         - name: scheduler-estimator-client-cert
           secret:
             secretName: karmada-descheduler-scheduler-estimator-client-cert
+      securityContext:
+        seccompProfile:
+          type: RuntimeDefault
```
RainbowMango, line 66: ditto

artifacts/deploy/karmada-controller-manager.yaml
zhzhuang-zju: done

artifacts/deploy/karmada-aggregated-apiserver.yaml
```diff
@@ -24,6 +24,9 @@ spec:
         - name: karmada-aggregated-apiserver
           image: docker.io/karmada/karmada-aggregated-apiserver:latest
           imagePullPolicy: IfNotPresent
+          securityContext:
+            allowPrivilegeEscalation: false
+            privileged: false
```
RainbowMango, line 29: The `privileged` [defaults to](https://github.com/kubernetes/kubernetes/blob/16e8911fdccbdb348a0b3291acf6102494494f9b/staging/src/k8s.io/api/core/v1/types.go#L7521-L7526) `false`, how about the `allowPrivilegeEscalation`?

zhzhuang-zju: it defaults to `true`

code: https://github.com/kubernetes/kubernetes/blob/488fac490463d9499ff22dc7500f2c93497417dc/pkg/securitycontext/util.go#L174-L187

When `allowPrivilegeEscalation` is nil, it will not add the `no_new_privs` option, which means a process can not gain more privileges than its parent process.

- container with setting `allowPrivilegeEscalation=false`
```bash
/ # cat /proc/1/status |grep NoNewPrivs
NoNewPrivs:     1
```
- container without setting `allowPrivilegeEscalation`
```bash
/ # cat /proc/1/status |grep NoNewPrivs
NoNewPrivs:     0
```

zhzhuang-zju:
@RainbowMango Should other installation methods also synchronize this configuration? Additionally, there is a conflict between this configuration and the `spec.os.windows` field; should we consider this scenario?

https://github.com/karmada-io/karmada/blob/5bfdf615c7129dedd12d441d585636554932d5fd/vendor/k8s.io/api/core/v1/types.go#L3957-L3989

RainbowMango:
> Should other installation methods also synchronize this configuration?

I think yes.

> Additionally, there is a conflict between this configuration and the spec.os.windows field; should we consider this scenario?

Do you have any idea how to support windows?

zhzhuang-zju:
> Do you have any idea how to support windows?

The Pod Security Context for Windows OS can be specifically configured using `securityContext.windowsOptions`, which includes the following fields:
https://github.com/karmada-io/karmada/blob/5bfdf615c7129dedd12d441d585636554932d5fd/vendor/k8s.io/api/core/v1/types.go#L7627-L7652

If there is a requirement for Karmada control plane components to run on Windows OS, providing a configurable Security Context would address this need. If there is no requirement to run on Windows OS, default configurations would be sufficient.

PR #6059, opened 2025-01-16
feat: Add automatic memory limit handling

dongjiang1989:
**What type of PR is this?**

Enable automatic detection of memory limits and configure GOMEMLIMIT to match.Also includes a flag to allow controlling the reserved ratio.

Fixes #

```release-note
Add automatic memory limit handling
```


zhzhuang-zju:
@dongjiang1989  Using `GOMEMLIMIT` for GC tuning can help reduce OOM risks and, combined with `GOGC=off`, decrease GC trigger frequency. Since `GOGC` is enabled by default, does this PR aim to reduce OOM risks?

PR #6061, opened 2025-01-17
enhance ensureWork to continue processing clusters after individual failures

chaosi-zju:
**What type of PR is this?**

/kind feature

When `ensureWork`, one cluster fails would prevent other clusters from syncing, this PR aims to enhance `ensureWork` to continue processing clusters after individual failures.

real user case:

1. `member1` cluster is discontected, while `member2`/`member3` is normal
2. delete deployment from control-plane, it will be successfully deleted on all clusters except member1, besides, the `work` of `member1` cluster will remain due to the finalizer
4. then, assuming `member1` cluster recovered
3. recreate the deployment with the same name, deployments to all clusters may be blocked for up to 1000s.

cause:

when `ensureWork` failed on member1 cluster due to the old `work` is exist and being deleted, we fast failed without continue to try rest clusters. the problem will persist until the old `work` in `member1` is successfully deleted in reconciliation process.

benefit:

when recreating the deployment, it can be quickly rolled out to all clusters except member1. However, deployments to the member1 cluster may still be blocked for up to 1000s.

Fixes #

reproduce testing method:
* create any simple deployment and pp
* add a `finalizer` to the `work` in `karmada-es-member1` namespace to prevent that `work` from being deleted
* delete the deployment (residual `work` in `karmada-es-member1` will remain)
* recreate deployment, then no `work` will be created (a more reasonable result is member1's work creation fails, while other clusters succeed)

just like:

```shell
kubectl create deploy nginx --image=nginx
kubectl get deploy nginx -o yaml | karmadactl apply --all-clusters=true -f -

kubectl patch work nginx-687f7fb96f --type='json' -p '[{"op": "add", "path": "/metadata/finalizers/1", "value": "test"}]' -n karmada-es-member1

kubectl delete deploy nginx
kubectl create deploy nginx --image=nginx
```

```release-note

```

### DIFF
--- a/pkg/controllers/binding/common.go
+++ b/pkg/controllers/binding/common.go
@@ -23,6 +23,7 @@ import (
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog/v2"
 	"k8s.io/utils/ptr"
@@ -45,19 +46,8 @@ func ensureWork(
 	ctx context.Context, c client.Client, resourceInterpreter resourceinterpreter.ResourceInterpreter, workload *unstructured.Unstructured,
 	overrideManager overridemanager.OverrideManager, binding metav1.Object, scope apiextensionsv1.ResourceScope,
 ) error {
-	var targetClusters []workv1alpha2.TargetCluster
-	var bindingSpec workv1alpha2.ResourceBindingSpec
-	switch scope {
-	case apiextensionsv1.NamespaceScoped:
-		bindingObj := binding.(*workv1alpha2.ResourceBinding)
-		bindingSpec = bindingObj.Spec
-	case apiextensionsv1.ClusterScoped:
-		bindingObj := binding.(*workv1alpha2.ClusterResourceBinding)
-		bindingSpec = bindingObj.Spec
-	}
-
-	targetClusters = bindingSpec.Clusters
-	targetClusters = mergeTargetClusters(targetClusters, bindingSpec.RequiredBy)
+	bindingSpec := getBindingSpec(binding, scope)
+	targetClusters := mergeTargetClusters(bindingSpec.Clusters, bindingSpec.RequiredBy)
 
 	var jobCompletions []workv1alpha2.TargetCluster
 	var err error
@@ -68,6 +58,7 @@ func ensureWork(
 		}
 	}
 
+	var errs []error
 	for i := range targetClusters {
 		targetCluster := targetClusters[i]
 		clonedWorkload := workload.DeepCopy()
@@ -82,7 +73,8 @@ func ensureWork(
 				if err != nil {
 					klog.Errorf("Failed to revise replica for %s/%s/%s in cluster %s, err is: %v",
 						workload.GetKind(), workload.GetNamespace(), workload.GetName(), targetCluster.Name, err)
-					return err
+					errs = append(errs, err)
+					continue
 				}
 			}
 
@@ -94,7 +86,8 @@ func ensureWork(
 				if err = helper.ApplyReplica(clonedWorkload, int64(jobCompletions[i].Replicas), util.CompletionsField); err != nil {
 					klog.Errorf("Failed to apply Completions for %s/%s/%s in cluster %s, err is: %v",
 						clonedWorkload.GetKind(), clonedWorkload.GetNamespace(), clonedWorkload.GetName(), targetCluster.Name, err)
-					return err
+					errs = append(errs, err)
+					continue
 				}
 			}
 		}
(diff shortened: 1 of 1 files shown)


RainbowMango:
> add a finalizer to the work in karmada-es-member1 namespace to prevent that work from being deleted

Not the real word use case.
Please tell us why this patch is necessary, and what's the effect if not have it.

chaosi-zju:
> Not the real word use case.
> Please tell us why this patch is necessary, and what's the effect if not have it.

That is not an attempt to describe a real use case; rather, it is a quick and convenient method for reproduction, intended for testing and validation.

**the corresponding real use case is:**
1. member1 cluster is discontected, while member2/member3 is normal
2. delete deployment from control-plane, then due to the inability to delete resources in the member1 cluster, the corresponding `work` will remain due to the finalizer
3. then, member1 cluster recovered
4. recreate the deployment with the same name, then it will failed to create `work` in `ensureWork`, due to the old `work` is exist and being deleted, see: https://github.com/karmada-io/karmada/blob/820fd06409c46e8ea279ac7ec90de7461338cb5c/pkg/controllers/ctrlutil/work.go#L79-L81

so, the patch is merely an attempt to simulate a residual work in that cluster.

RainbowMango (commented):
/assign

RainbowMango:
> delete deployment from control-plane, then due to the inability to delete resources in the member1 cluster, the corresponding work will remain due to the finalizer

Has the deployment been successfully deleted on all clusters except member1? If I understand correctly, the member1 cluster won't block the deletion on the other clusters.

chaosi-zju:
> > delete deployment from control-plane, then due to the inability to delete resources in the member1 cluster, the corresponding work will remain due to the finalizer
> 
> Has the deployment been successfully deleted on all clusters except member1?

yes, the member1 cluster won't block the deletion on the other clusters, so the deployment been successfully deleted on all clusters except member1.

RainbowMango:
OK.

> recreate the deployment with the same name, then it will failed to create work in ensureWork, due to the old work is exist and being deleted, see:

Before recreate the deployment, what will happen to the residual work after the member1 is recovered? Will it remain permanently, or the system clean it up eventually?

chaosi-zju:
> Before recreate the deployment, what will happen to the residual work after the member1 is recovered? Will it remain permanently, or the system clean it up eventually?

the system clean it up eventually (about 1000s)

RainbowMango:
So, the value of this PR is :
When recreating the deployment, it can be quickly rolled out to all clusters except member1. However, deployments to the member1 cluster may still be blocked for up to 1000s. 
Right?

chaosi-zju:
> When recreating the deployment, it can be quickly rolled out to all clusters except member1. However, deployments to the member1 cluster may still be blocked for up to 1000s.
> Right?

yes

pkg/controllers/binding/common.go
```diff
@@ -137,12 +132,29 @@ func ensureWork(
 			ctrlutil.WithSuspendDispatching(shouldSuspendDispatching(bindingSpec.Suspension, targetCluster)),
 			ctrlutil.WithPreserveResourcesOnDeletion(ptr.Deref(bindingSpec.PreserveResourcesOnDeletion, false)),
 		); err != nil {
-			return err
+			errs = append(errs, err)
+			continue
 		}
 	}
+	if len(errs) > 0 {
+		return fmt.Errorf("errors occurred while processing target clusters: %+v", errs)
```
RainbowMango, line 140: why use "%`+`v" here? Have you tested it? What does the error look like?

> '+'	always print a sign for numeric values;

I remember the "+" is used for print numeric values.

chaosi-zju: In this case, `%+v` is the same as `%v`, try here: https://go.dev/play/p/6V52lFSbAZP

my personal habit prefer `%+v`, I can change back to %v for style consistency.

RainbowMango: OK. Then we can move forward once we see the test report.

chaosi-zju: for `struct` object, `%v` shows only the field values. while `%+v` shows field names and field values. other scenes their output is the same.

as for logging scenario, we may mostly want to known which field corresponds to the value, so I always prefer `%+v`

chaosi-zju:
latest test result ok

```shell
$ karmadactl --karmada-context karmada-apiserver get deploy --operation-scope=members
NAME    CLUSTER   READY   UP-TO-DATE   AVAILABLE   AGE   ADOPTION
nginx   member2   1/1     1            1           9s    Y
nginx   member3   1/1     1            1           9s    Y
```

by running above test method, after recreate deployment, we can see deployment can be quickly rolled out to all clusters except member1

RainbowMango:
I want to see the error message.

chaosi-zju:
> I want to see the error message.

one cluster fail:

```shell
E0124 01:30:20.001513       1 binding_controller.go:131] Failed to transform resourceBinding(default/nginx-deployment) to works. Error: work karmada-es-member1/nginx-687f7fb96f is being deleted.
```

two cluster fail:
```shell
E0124 01:58:13.499553       1 binding_controller.go:131] Failed to transform resourceBinding(default/nginx-deployment) to works. Error: [work karmada-es-member1/nginx-687f7fb96f is being deleted, work karmada-es-member2/nginx-687f7fb96f is being deleted].
```

RainbowMango (approved):
/lgtm
/approve

PR #6065, opened 2025-01-18
fix(search/proxy/controller): resourceregistry merge ns

### LINKED ISSUE #6064: karmada-search ResourceRegistry resourceSelectors can not merge the same GVK's namespace
**What happened**:
I have two yaml config of resourceregistry,they just namespace different
```yaml
apiVersion: search.karmada.io/v1alpha1
kind: ResourceRegistry
metadata:
  name: secret-search-ns1
spec:
  targetCluster:
    clusterNames:
      - cluster-1
  resourceSelectors:
    - apiVersion: v1
      kind: Secret
      namespace: ns1
```
```yaml
apiVersion: search.karmada.io/v1alpha1
kind: ResourceRegistry
metadata:
  name: secret-search-ns2
spec:
  targetCluster:
    clusterNames:
      - cluster-1
  resourceSelectors:
    - apiVersion: v1
      kind: Secret
      namespace: ns2
```
when I apply them ,karmada-search just cached one namespace resource,the latter one override the previous one namespace
get ResourceRegistry:
Image
get secret,juet get ns1's secret:
Image

I think karmada-search should merge these namespace,
it should cached ns1 and ns2,this should be a bug.

**Environment**:
- Karmada version: 1.12.2
- Others:

JimDevil:
**What type of PR is this?**

karmada-search resourceregistry can merge the same cluster gvk's namespaces
get ResourceRegistry:
image
get secret:
image

Fixes #6064

```release-note
`karmada-search`: Fixed the issue that namespaces in different ResourceRegistry might be overwritten.
```

### DIFF
--- a/pkg/search/proxy/controller.go
+++ b/pkg/search/proxy/controller.go
@@ -228,27 +228,34 @@ func (ctl *Controller) reconcile(util.QueueKey) error {
 				klog.Warningf("cluster %s is notReady", cluster.Name)
 				continue
 			}
-
-			if _, exist := resourcesByClusters[cluster.Name]; !exist {
-				resourcesByClusters[cluster.Name] = make(map[schema.GroupVersionResource]*store.MultiNamespace)
-			}
-
-			for resource, multiNS := range matchedResources {
-				gvk, err := ctl.restMapper.KindFor(resource)
-				if err != nil {
-					klog.Errorf("Failed to get gvk: %v", err)
-					continue
-				}
-				if !helper.IsAPIEnabled(cluster.Status.APIEnablements, gvk.GroupVersion().String(), gvk.Kind) {
-					klog.Warningf("Resource %s is not enabled for cluster %s", resource.String(), cluster)
-					continue
-				}
-				resourcesByClusters[cluster.Name][resource] = multiNS
-			}
+			ctl.mergeResourcesByClusters(resourcesByClusters, cluster, matchedResources)
 		}
 	}
 
 	return ctl.store.UpdateCache(resourcesByClusters, registeredResources)
+}
+
+func (ctl *Controller) mergeResourcesByClusters(resourcesByClusters map[string]map[schema.GroupVersionResource]*store.MultiNamespace, cluster *clusterv1alpha1.Cluster, matchedResources map[schema.GroupVersionResource]*store.MultiNamespace) {
+	if _, exist := resourcesByClusters[cluster.Name]; !exist {
+		resourcesByClusters[cluster.Name] = make(map[schema.GroupVersionResource]*store.MultiNamespace)
+	}
+
+	for resource, multiNS := range matchedResources {
+		gvk, err := ctl.restMapper.KindFor(resource)
+		if err != nil {
+			klog.Errorf("Failed to get gvk: %v", err)
+			continue
+		}
+		if !helper.IsAPIEnabled(cluster.Status.APIEnablements, gvk.GroupVersion().String(), gvk.Kind) {
+			klog.Warningf("Resource %s is not enabled for cluster %s", resource.String(), cluster)
+			continue
+		}
+		if ns, exist := resourcesByClusters[cluster.Name][resource]; !exist {
+			resourcesByClusters[cluster.Name][resource] = multiNS
+		} else {
+			resourcesByClusters[cluster.Name][resource] = ns.Merge(multiNS)
+		}
+	}
 }
 
 type errorHTTPHandler struct {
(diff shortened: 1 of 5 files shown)


pkg/search/proxy/controller.go
```diff
@@ -228,29 +228,36 @@ func (ctl *Controller) reconcile(util.QueueKey) error {
...
-				resourcesByClusters[cluster.Name][resource] = multiNS
-			}
+			ctl.reconcileResourcesByClusters(resourcesByClusters, cluster, matchedResources)
 		}
 	}
 
 	return ctl.store.UpdateCache(resourcesByClusters, registeredResources)
 }
 
+func (ctl *Controller) reconcileResourcesByClusters(resourcesByClusters map[string]map[schema.GroupVersionResource]*store.MultiNamespace, cluster *clusterv1alpha1.Cluster, matchedResources map[schema.GroupVersionResource]*store.MultiNamespace) {
```
RainbowMango, line 238:
```suggestion
func (ctl *Controller) reconcileResourcesByClusters(resourcesByClusters map[string]map[schema.GroupVersionResource]*store.MultiNamespace, cluster *clusterv1alpha1.Cluster, matchedResources map[schema.GroupVersionResource]*store.MultiNamespace) {
```
Does the name `MergeResourcesByCluster` look better?

RainbowMango: I'm glad to see that you added a new method which makes it easier to have a unit test.
Can we try to add some unit tests on it this time?

RainbowMango (commented):
Generally looks good to me.
But also invite guys who might be interested in this
cc @yanfeng1992

RainbowMango:
This is a bug I think, and added a release note to the PR description.
/kind bug

pkg/search/proxy/controller.go
RainbowMango: > ok!by the way,your suggestion no changes

My bad, updated! Thanks for the reminder.

RainbowMango: > Do you mean adding unit tests for MergeResourcesByCluster? I saw there are already unit test for reconcile

Oh, right! I didn't notice it. But the unit test missed this bug, right? Shall we add a test to it?

RainbowMango: That's another story. Now focus on fixing the bug, even if the resourceSelector.namespace uses a slice, this issue also needs to be fixed.

RainbowMango: Thanks in advance. We probably need to explain the reason and the real-world use case. Welcome to propose a new feature or enhancement if you like.

By the way, are you using karmada-search now?

pkg/search/proxy/controller_test.go
```diff
@@ -104,19 +102,26 @@ func TestController(t *testing.T) {
 }
 
 func TestController_reconcile(t *testing.T) {
-	echoStrings := func(ss ...string) string {
-		sort.Strings(ss)
-		return strings.Join(ss, ",")
+	newMultiNs := func(namespaces ...string) *store.MultiNamespace {
+		multiNs := store.NewMultiNamespace()
+		if len(namespaces) == 0 {
+			multiNs.Add("")
```
RainbowMango, line 108:
```suggestion
			multiNs.Add(metav1.NamespaceAll)
```
for readability :)

RainbowMango (commented):
Just a nits, otherwise LGTM.

RainbowMango (commented):
/lgtm

pkg/search/proxy/controller.go
RainbowMango: Sounds good!

RainbowMango:
Let's wait for a while and see if others have any comments.

yanfeng1992:
lgtm

RainbowMango:
Thank you @yanfeng1992 .

@JimDevil I wonder have you tested it on your side? Even though we tested by unit test, it would be great to have another test.
And do you want this fix to be backported to release branches? Like release 1.12, so this patch would be included in v1.12.3.

yanfeng1992:
> @JimDevil I wonder have you tested it on your side? Even though we tested by unit test, it would be great to have another test. 

How about adding a related test case in e2e testing? @JimDevil @RainbowMango

RainbowMango:
> How about adding a related test case in e2e testing? @JimDevil @RainbowMango

+1

RainbowMango:
Opened #6075 to track the cherry-pick and E2E.
@JimDevil Would you like to help?

RainbowMango (approved):
/lgtm
/approve

Issue #6071, opened 2025-01-21
karmada controller print too many "No need to sync binding" log


CharlesQQ:
**What would you like to be added**:
Set the level of the following log to level 5
```
I0121 10:49:43.942873      16 dependencies_distributor.go:189] No need to sync binding(default/xxxxxxx)
```

**Why is this needed**:

The deployments have dependent resources in control Panel, like configmap; when karmada controller restarts, those dependent resources trigger DependenciesDistributor reconcile and try to find matched rb resources, and print too many log;  The following log will always be printed:

![Image](https://github.com/user-attachments/assets/1488f379-022e-4f6d-a8b9-5496b3cc0bc5)

XiShanYongYe-Chang:
I think it's more like a cleanup action.
/cleanup
/remove-kind feature

In favor of #6072 
/assign @CharlesQQ

RainbowMango:
> The deployments have dependent resources in control Panel, like configmap; when karmada controller restarts, those dependent resources trigger DependenciesDistributor reconcile and try to find matched rb resources, and print too many log; 

This means even if you have only one Deployment enabled with the dependent feature, it will iterate all ResourceBindings, and each unrelated ResourceBinding will print a log like this, correct?

CharlesQQ:
> This means even if you have only one Deployment enabled with the dependent feature, it will iterate all ResourceBindings, and each unrelated ResourceBinding will print a log like this, correct?

Yes!

PR #6072, opened 2025-01-21
feat(dependenciesdistributor):  remove log

CharlesQQ:
**What type of PR is this?**

fix controller print too many "No need to sync binding" log when restarts

Fixes https://github.com/karmada-io/karmada/issues/6071

```release-note

```

### DIFF
--- a/pkg/dependenciesdistributor/dependencies_distributor.go
+++ b/pkg/dependenciesdistributor/dependencies_distributor.go
@@ -186,7 +186,6 @@ func (d *DependenciesDistributor) reconcileResourceTemplate(key util.QueueKey) e
 
 		matched := matchesWithBindingDependencies(resourceTemplateKey, binding)
 		if !matched {
-			klog.V(4).Infof("No need to sync binding(%s/%s)", binding.Namespace, binding.Name)
 			continue
 		}
 


XiShanYongYe-Chang (commented):
I agree with this log level adjustment.
/lgtm

From the pr title description, the word fix is a bit inappropriate, can you help modify it?

pkg/dependenciesdistributor/dependencies_distributor.go
```diff
@@ -186,7 +186,7 @@ func (d *DependenciesDistributor) reconcileResourceTemplate(key util.QueueKey) e
 
 		matched := matchesWithBindingDependencies(resourceTemplateKey, binding)
 		if !matched {
-			klog.V(4).Infof("No need to sync binding(%s/%s)", binding.Namespace, binding.Name)
+			klog.V(5).Infof("No need to sync binding(%s/%s)", binding.Namespace, binding.Name)
```
RainbowMango, line 189: The logic here is more like selecting a small number of items from a large set, logging for items that do not match isn't really helpful, right?

In other words, shall we remove this log?

XiShanYongYe-Chang: Sounds good. We already have the matching logs, so the rest is negligible unmatched resources.

CharlesQQ: I agree to remove this log

RainbowMango:
Hi @CharlesQQ CI is failing

RainbowMango (approved):
/lgtm
/approve

PR #6073, opened 2025-01-21
Updating FlinkDeployment interpreter to display error status, improving health interpreter

### LINKED ISSUE #6023: FlinkDeployment health interpreter does not account for ImagePullBackOff Error
**What happened**: 

Tests done by @deefreak and @bharathguvvala show that the current FlinkDeployment health interpreter determines that the FlinkDeployment is unhealthy when it has an ImagePullBackOff Error. ~I have confirmed this behavior and have identified that the root cause is due to the interpreter not accounting for this specific error case.~

The FlinkDeployment health interpreter should only mark FlinkDeployments unhealthy if there are unable to be scheduled (meaning they are stuck in a RECONCILING or CREATED state, unrelated to a jobManagerDeploymentStatus of ERROR).

Karmada Version: v1.12.2

mszacillo:
**What type of PR is this?**
/kind feature

After doing some load testing of the FlinkDeployment failover (which overall has been looking quite good, but may need to address a couple edge cases), I found that the interpreter is missing one of the ephemeral states that FlinkDeployments can transition through. 

Occasionally on failover, the FlinkDeployment will transition from RECONCILING -> INITIALIZING -> CREATED, before finally ending on RUNNING. Additionally, we can make use of the status.error field to further improve the health interpretation. 

In this PR I've added:
1. INITIALIZING state as an ephemeral state which we should check during health interpretation.
2. Checking if status.error != nil. If the deployment has a published error, then we treat it as healthy, as this indicates that the job failed due to user error. In the future, we may consider adding an ignore list of errors which we would like to failover. 

Fixes #6023

```release-note
`karmada-controller-manager`: FlinkDeployment health interpreter improvements, adding status.error to reflected status
```

### DIFF
--- a/pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/flink.apache.org/v1beta1/FlinkDeployment/customizations.yaml
+++ b/pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/flink.apache.org/v1beta1/FlinkDeployment/customizations.yaml
@@ -7,14 +7,27 @@ spec:
     apiVersion: flink.apache.org/v1beta1
     kind: FlinkDeployment
   customizations:
+# FlinkDeployment health is interpreted based on the application's state.
+#
+# Health Rules:
+# 1. If the job is in a terminal state [Failed, Finished, Canceled, Suspended] or in the Running state, it is considered healthy.
+# 2. If the job is in an ephemeral state [Reconciling, Initializing, Created]:
+#    - It is treated as healthy ONLY if there is a published error (e.g., user-related issues like an incorrect image path).
+#    - Otherwise, it is treated as unhealthy and may be rescheduled.
+# 3. Short-lived states [Cancelling, Failing, Restarting] are treated as healthy because they will directly transition to their respective terminal states:
+#    - Cancelling -> Canceled / Suspended
+#    - Failing -> Failed
+#    - Restarting triggers a restart, bringing the job back to the Created state.
+#
+# For more information on the Flink state diagram, refer to the official documentation: https://nightlies.apache.org/flink/flink-docs-release-1.20/docs/internals/job_scheduling/
     healthInterpretation:
       luaScript: >
         function InterpretHealth(observedObj)
-          if observedObj.status ~= nil and observedObj.status.jobStatus ~= nil then
-            if observedObj.status.jobStatus.state ~= 'CREATED' and observedObj.status.jobStatus.state ~= 'RECONCILING' then
+          if observedObj.status ~= nil and observedObj.status.jobStatus ~= nil and observedObj.status.jobStatus.state ~= nil then
+            if observedObj.status.jobStatus.state ~= 'CREATED' and observedObj.status.jobStatus.state ~= 'INITIALIZING' and observedObj.status.jobStatus.state ~= 'RECONCILING' then
               return true
             else
-              return observedObj.status.jobManagerDeploymentStatus == 'ERROR'
+              return observedObj.status.error ~= nil or observedObj.status.jobManagerDeploymentStatus == 'ERROR'
             end
           end
           return false
@@ -91,6 +104,7 @@ spec:
             desiredObj.status = {}
           end
           clusterInfo = {}
+          error = ''
           jobManagerDeploymentStatus = ''
           jobStatus = {}
           lifecycleState = ''
@@ -102,6 +116,7 @@ spec:
             currentStatus = statusItems[i].status
             if currentStatus ~= nil then
               clusterInfo = currentStatus.clusterInfo
+              error = currentStatus.error
               jobManagerDeploymentStatus = currentStatus.jobManagerDeploymentStatus
               jobStatus = currentStatus.jobStatus
               observedGeneration = currentStatus.observedGeneration
@@ -112,6 +127,7 @@ spec:
           end
 
           desiredObj.status.clusterInfo = clusterInfo
+          desiredObj.status.error = error
           desiredObj.status.jobManagerDeploymentStatus = jobManagerDeploymentStatus
           desiredObj.status.jobStatus = jobStatus
           desiredObj.status.lifecycleState = lifecycleState
@@ -128,6 +144,7 @@ spec:
             return status
           end
           status.clusterInfo = observedObj.status.clusterInfo
+          status.error = observedObj.status.error
           status.jobManagerDeploymentStatus = observedObj.status.jobManagerDeploymentStatus
           status.jobStatus = observedObj.status.jobStatus
           status.observedGeneration = observedObj.status.observedGeneration


RainbowMango:
@yike21 @chaunceyjiang can you take a look?

yike21:
> @yike21 @chaunceyjiang can you take a look?

O, I'll take a look at it ASAP :-)

RainbowMango:
Here is the FlinkDeployment reference where you can find the definition of the status:
https://nightlies.apache.org/flink/flink-kubernetes-operator-docs-main/docs/custom-resource/reference/

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/flink.apache.org/v1beta1/FlinkDeployment/customizations.yaml
```diff
@@ -10,11 +10,11 @@ spec:
...
       luaScript: >
         function InterpretHealth(observedObj)
-          if observedObj.status ~= nil and observedObj.status.jobStatus ~= nil then
-            if observedObj.status.jobStatus.state ~= 'CREATED' and observedObj.status.jobStatus.state ~= 'RECONCILING' then
+          if observedObj.status ~= nil and observedObj.status.jobStatus ~= nil and observedObj.status.jobStatus.state ~= nil then
+            if observedObj.status.jobStatus.state ~= 'CREATED' and observedObj.status.jobStatus.state ~= 'INITIALIZING' and observedObj.status.jobStatus.state ~= 'RECONCILING' then
               return true
             else
-              return observedObj.status.jobManagerDeploymentStatus == 'ERROR'
+              return observedObj.status.error ~= nil or observedObj.status.jobManagerDeploymentStatus == 'ERROR'
```
RainbowMango, line 17: > return observedObj.status.error ~= nil or observedObj.status.jobManagerDeploymentStatus == 'ERROR'

Does that mean
1. if `observedObj.status.error ~= nil`( there is an error) ==> return true ==> in healthy state
2. Or, if `observedObj.status.jobManagerDeploymentStatus == 'ERROR'` (there is an error) ==> return true ==> in healthy state

I'm not quite sure about that. Seems it should be unhealthy.

By the way, it is recommended to implement the `InterpretHealth` function like:
1. find the status that indicates it's in a healthy state, then return true
2. all the other statuses should return false.

I'm still not familiar with the FlinkDeployment state, I'm not sure if we can implement it that way.

mszacillo: That's correct. There are a variety of user related errors that should not result in a failover. An example of this would be an incorrect image pointer - if we didn't check the error then the application would continue to failover from cluster to cluster causing a lot of churn. 

An example status in this case:
```
status:
    clusterInfo: {}
    error: '{"type":"org.apache.flink.kubernetes.operator.exception.DeploymentFailedException","message":"rpc
      error: code = NotFound desc = failed to pull and unpack image \"dummy-image\":
      failed to resolve reference \"dummy-image\":
      dummy-image: not found","additionalMetadata":{"reason":"ErrImagePull"},"throwableList":[]}'
    jobManagerDeploymentStatus: ERROR
    jobStatus:
      jobId: 86cdf08f1e2ae3a6e193aa04112dc261
      savepointInfo:
        lastPeriodicSavepointTimestamp: 0
        savepointHistory: []
      state: RECONCILING
    lifecycleState: DEPLOYED
    reconciliationStatus:
```

That said, I believe just keeping the `status.error` may be enough. I had kept the `jobManagerDeploymentStatus` in the health interpreter for continuity. 

> By the way, it is recommended to implement the InterpretHealth function like - 

Please correct me if I'm misunderstanding, but the existing interpreter behaves this way. We immediately return true if we are in a terminal state. Otherwise, if we are in an ephemeral state (RECONCILING, INITIALIZING, CREATED), we verify that there are no user errors, which would suggest a cluster issue impacting the deployment.

Maybe we could add some sort of error allowlist which we can check to verify the error is user-related? If it is, return healthy. That would give us some more control over the health interpretation.

mszacillo: Sounds good, I'll update the PR shortly.

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/flink.apache.org/v1beta1/FlinkDeployment/customizations.yaml
```diff
@@ -7,14 +7,26 @@ spec:
...
+    # If the job is in an ephemeral state during deployment [Reconciling, Initializing, Created],
+    #    then it is treated as healthy only if there is a published error, which suggests a user-related issue such as
+    #    an incorrect image pointer path.
+    # Else, the job is treated as unhealthy and will be eventually rescheduled.
+    #
+    # Note: the states [Cancelling, Failing, Restarting] are treated as healthy as these are short-lived states which will directly
+    #       transition to their respective terminal states. Cancelling -> Canceled, Failing -> Failed.
+    #       Restarting will trigger a restart of the application, taking the job back to the starting Created state.
+    #
+    # For more information on the Flink state diagram, please see https://nightlies.apache.org/flink/flink-docs-release-1.20/docs/internals/job_scheduling/
```
yike21, line 21:
```suggestion
# FlinkDeployment health is interpreted based on the application's state.
#
# Health Rules:
# 1. If the job is in a terminal state [Failed, Finished, Canceled, Suspended] or in the Running state, it is considered healthy.
# 2. If the job is in an ephemeral state [Reconciling, Initializing, Created]:
#    - It is treated as healthy ONLY if there is a published error (e.g., user-related issues like an incorrect image path).
#    - Otherwise, it is treated as unhealthy and may be rescheduled.
# 3. Short-lived states [Cancelling, Failing, Restarting] are treated as healthy because they will directly transition to their respective terminal states:
#    - Cancelling -> Canceled
#    - Failing -> Failed
#    - Restarting triggers a restart, bringing the job back to the Created state.
#
# For more information on the Flink state diagram, refer to the official documentation:
# https://nightlies.apache.org/flink/flink-docs-release-1.20/docs/internals/job_scheduling/
```

Just tidying up the formatting. Thank you very much for your contribution!

mszacillo: Thank you for the suggestions!

yike21:
/lgtm
Thanks a lot! @mszacillo

RainbowMango (approved):
Looks great!
/approve

Issue #6031, opened 2025-01-09
[Performance] Enhancements in release-1.13


RainbowMango:
**Summary**:

This issue tracks all planned performance optimization efforts targeted for this release cycle.

Additionally, we are building a team dedicated to addressing performance-related issues, welcome to join us!
Following is our bi-weekly meeting: 
https://docs.google.com/document/d/1BUZBDvtbtfU9MmOz7KVdvGiiCny-zz_34FuLlW4AGmM/edit

**Iteration Tasks**
- [x] Fix ineffective DeepEqual check in CreateOrUpdateWork. (@zach593, #5939)
- [x] DoubleCheck all places where using `unstructured.MarshalJSON`. (See https://github.com/karmada-io/karmada/pull/5939#issuecomment-2658615599)
- [x] Re-organize controller utils to make it easier to extend. (@zach593, #6018)
- [x] https://github.com/karmada-io/karmada/pull/6076

PR #6076, opened 2025-01-22
add work mutating in ctrlutil.CreateOrUpdateWork()

zach593:
**What type of PR is this?**

/kind feature

It could remove unnecessary update operations in `ctrlutil.CreateOrUpdateWork()`, by mutating the `Work` in advance, and let `DeepEqual()` return true. when karmada-controller-manager restart, this may help reduce the time it takes to clear the work queue.

Should be part of #6031, if it's not true please correct me.

Fixes #

Tests will remain failing until #6028 is merged and this PR is rebased on it.

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/ctrlutil/work.go
+++ b/pkg/controllers/ctrlutil/work.go
@@ -31,13 +31,15 @@ import (
 
 	workv1alpha1 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha1"
 	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
+	"github.com/karmada-io/karmada/pkg/resourceinterpreter/default/native/prune"
 	"github.com/karmada-io/karmada/pkg/util"
+	"github.com/karmada-io/karmada/pkg/util/helper"
 )
 
 // CreateOrUpdateWork creates a Work object if not exist, or updates if it already exists.
-func CreateOrUpdateWork(ctx context.Context, client client.Client, workMeta metav1.ObjectMeta, resource *unstructured.Unstructured, options ...WorkOption) error {
+func CreateOrUpdateWork(ctx context.Context, c client.Client, workMeta metav1.ObjectMeta, resource *unstructured.Unstructured, options ...WorkOption) error {
+	resource = resource.DeepCopy()
 	if workMeta.Labels[util.PropagationInstruction] != util.PropagationInstructionSuppressed {
-		resource = resource.DeepCopy()
 		// set labels
 		util.MergeLabel(resource, util.ManagedByKarmadaLabel, util.ManagedByKarmadaLabelValue)
 		// set annotations
@@ -48,26 +50,16 @@ func CreateOrUpdateWork(ctx context.Context, client client.Client, workMeta meta
 			util.MergeAnnotation(resource, workv1alpha2.ResourceConflictResolutionAnnotation, conflictResolution)
 		}
 	}
-
-	workloadJSON, err := json.Marshal(resource)
+	// Do the same thing as the mutating webhook does, remove the irrelevant fields for the resource.
+	// This is to avoid unnecessary updates to the Work object, especially when controller starts.
+	err := prune.RemoveIrrelevantFields(resource, prune.RemoveJobTTLSeconds)
 	if err != nil {
-		klog.Errorf("Failed to marshal workload(%s/%s), error: %v", resource.GetNamespace(), resource.GetName(), err)
+		klog.Errorf("Failed to prune irrelevant fields for resource %s/%s. Error: %v", resource.GetNamespace(), resource.GetName(), err)
 		return err
 	}
 
 	work := &workv1alpha1.Work{
 		ObjectMeta: workMeta,
-		Spec: workv1alpha1.WorkSpec{
-			Workload: workv1alpha1.WorkloadTemplate{
-				Manifests: []workv1alpha1.Manifest{
-					{
-						RawExtension: runtime.RawExtension{
-							Raw: workloadJSON,
-						},
-					},
-				},
-			},
-		},
 	}
 
 	applyWorkOptions(work, options)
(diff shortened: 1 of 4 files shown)


RainbowMango (commented):
/assign

pkg/webhook/work/mutating.go
```diff
@@ -55,49 +53,15 @@ func (a *MutatingAdmission) Handle(_ context.Context, req admission.Request) adm
...
-		err = prune.RemoveIrrelevantFields(workloadObj, prune.RemoveJobTTLSeconds)
-		if err != nil {
-			klog.Errorf("Failed to remove irrelevant fields for the work(%s/%s), err: %v", work.Namespace, work.Name, err)
-			return admission.Errored(http.StatusInternalServerError, err)
-		}
-
-		// Skip label/annotate the workload of Work that is not intended to be propagated.
-		if work.Labels[util.PropagationInstruction] != util.PropagationInstructionSuppressed {
-			setLabelsAndAnnotationsForWorkload(workloadObj, work)
-		}
```
RainbowMango, line 77: @XiShanYongYe-Chang @whitewindmills Do you remember why we skipped the step of setting the label?

pkg/controllers/ctrlutil/work.go
```diff
@@ -72,10 +74,35 @@ func CreateOrUpdateWork(ctx context.Context, client client.Client, workMeta meta
...
+	}
+	if existingWork.Labels[workv1alpha2.WorkPermanentIDLabel] != "" {
+		if work.Labels == nil {
+			work.Labels = make(map[string]string)
+		}
+		work.Labels[workv1alpha2.WorkPermanentIDLabel] = existingWork.Labels[workv1alpha2.WorkPermanentIDLabel]
+	}
+	// mutate work here to let the deepEqual() have chance to return true, to reduce the HTTP UPDATE requests.
+	// otherwise it will always return false, because the informer cache is already mutated.
+	err = mutating.MutateWork(work)
```
RainbowMango, line 96: Would cleaning up irrelevant fields at the beginning of the function also achieve the same effect?
Something like:
```diff
diff --git a/pkg/controllers/ctrlutil/work.go b/pkg/controllers/ctrlutil/work.go
index b35d69598..e0c5dc8ef 100644
--- a/pkg/controllers/ctrlutil/work.go
+++ b/pkg/controllers/ctrlutil/work.go
@@ -31,6 +31,7 @@ import (

        workv1alpha1 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha1"
        workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
+       "github.com/karmada-io/karmada/pkg/resourceinterpreter/default/native/prune"
        "github.com/karmada-io/karmada/pkg/util"
 )

@@ -49,6 +50,7 @@ func CreateOrUpdateWork(ctx context.Context, client client.Client, workMeta meta
                }
        }

+       prune.RemoveIrrelevantFields(resource, prune.RemoveJobTTLSeconds)
        workloadJSON, err := json.Marshal(resource)
        if err != nil {
                klog.Errorf("Failed to marshal workload(%s/%s), error: %v", resource.GetNamespace(), resource.GetName(), err)

```

RainbowMango (commented):
I believe this approach works functionality, but it smells more complex, let's see if we can find a cheap and clean approach.

cc @XiShanYongYe-Chang for help

pkg/webhook/work/mutating.go
XiShanYongYe-Chang: For resources collected from member clusters, such as `EndpointSlice` resources in the MCS feature, the work is not propagated. Therefore, we do not need to set labels and annotations.

pkg/controllers/ctrlutil/work.go
XiShanYongYe-Chang: A little bit of the same feeling. Also, do we not need to execute `MutateWork` in webhook anymore?

zach593: > Would cleaning up irrelevant fields at the beginning of the function also achieve the same effect?

I made both the webhook and `CreateOrUpdateWork()` call the same function `mutating.MutateWork()`, because the `Work` object in the informer cache has already been modified by the webhook. As a result, the success of the deepequal check in `createOrUpdateWork` implicitly depends on the webhook’s implementation.

Of course, merely calling `prune.RemoveIrrelevantFields()` might currently achieve our goal. However, if someone later updates some logic in the webhook, the deepequal check could fail—and we might not notice this through unit tests or e2e tests, potentially taking months to discover the issue. Therefore, I believe it is necessary to explicitly call the same function to ensure that this contract remains unbroken.

zach593: > do we not need to execute `MutateWork` in webhook anymore?

I believe it should be retained, because in some deep usage scenarios of Karmada, users develop their own controllers to create `Work`, and then leverage Karmada’s capabilities to dispatch these `Work` objects to member clusters. I think it is still important for these scenarios to keep the functionality of RemoveIrrelevantFields and the addition of a permanentID within the webhook.

XiShanYongYe-Chang: There's a point in your example.

RainbowMango: > Of course, merely calling prune.RemoveIrrelevantFields() might currently achieve our goal.

Sorry, my bad, it can not achieve the goal as the [newly built Work object](https://github.com/karmada-io/karmada/blob/a2f92ed6bacaa24d62540dd68574977036812cc0/pkg/controllers/ctrlutil/work.go#L60-L73) doesn't have the `work.karmada.io/permanent-id` label.

zach593: After offline discussion, we decided not to directly reference the webhook code in the controller-manager for the time being, partly because this would lead to repeated marshal/unmarshal for the workload.

RainbowMango (commented):
/lgtm

Leave approval to @XiShanYongYe-Chang for another look.

XiShanYongYe-Chang (commented):
Thanks
/approve

chaosi-zju:
> @chaosi-zju Hi, chaosi, I remember you wanted to test the performance improvement of removing redundant updates to Work, and this commit would actually make a difference, you could test based on it.

thank you for reminding me and sorry for late!

I'm busy with other things right now, so that needs to be postponed.

PR #6066, opened 2025-01-19
Bump Go version to 1.22.11

y1hao:
**What type of PR is this?**

/kind cleanup

**What this PR does / why we need it**:

Bump Go version to latest 1.22 release.

**Which issue(s) this PR fixes**:
Part of #6063

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
Karmada now built with Golang v1.22.11.
```

PR #6086, opened 2025-01-25
Avoid coupling tests with hardcoded test CA certs

y1hao:
**What type of PR is this?**

/kind flake

As we discovered in https://github.com/karmada-io/karmada/pull/6066, 2 test cases in `membercluster_client_test.go` depend directly on a hardcoded test CA certificate, which was duplicated from Go's `net/http/internal` package. This can cause these tests to fail when a new version of Go has regenerated that internal certificate. The test server constructed with Go's `httptest` package already exposes a method to retrieve its trusted CA. This PR updates the tests to use that method to retrieve the CA cert directly from the test server.

Other test cases in this file don't go through a test server, so they can use any cert. They don't need to be updated.

```release-note
NONE
```


RainbowMango:
Thank you @y1hao ! Excellent!

> Other test cases in this file don't go through a test server, so they can use any cert. They don't need to be updated.

Yes! You are right, I tested it on my side and all tests passed after reverting the [copied certificate](https://github.com/karmada-io/karmada/blob/83dae7f11b977c71945bdadd065bca72598ff012/pkg/util/membercluster_client_test.go#L41-L61).

Shall we deprecate (remove) the copied certificate? I think the right practice is like what you did in this PR, which generates a fresh certificate from the test server. Even the rest tests don't go through the test server, but they need a cert, why not give them a fresh one?

My concern about keeping the copied certificate is that sometimes the developers imitate the legacy tests when they want to introduce new test cases. They may meet the same issue again.

y1hao:
@RainbowMango Yeah, I agree that the best practice would be to use a fresh cert for each test case. But the concern here is that generating  a new certificate is a heavy process, both in terms of compute resource consumed and the amount of code needed. For this reason, I think people usually don't create a new cert on the fly for testing, instead, they just use a pre-generated one.

When we retrieve the cert from the test server, it looks like that we are getting a fresh one for each test case, but in fact, we are just doing some decoding to retrieve the exact same cert every time, which is the previously copied one. If you compare the cert retrieved from the test server and the cert we copied from Go's `net/http/internal`, you'll find they are exactly the same. Go's test server doesn't bother to regenerate a new cert every time.

Similarly, k8s also has a pre-generated test certificate to use in tests: https://github.com/kubernetes/kubernetes/blob/f015fd66ce95d02cd66efc263eb9e5441b42a17d/hack/testdata/ca/ca.crt

So, I've updated this PR to use a new pre-generated certificate to replace the copied one, and documented the command used to generate it. This is the same practice as used in Go and k/k.

> My concern about keeping the copied certificate is that sometimes the developers imitate the legacy tests when they want to introduce new test cases. They may meet the same issue again.

With the new cert, if someone incorrectly used the pre-generated test cert when they should retrieve the cert from the test server, their test will fail so they can fix this before merging. I've also moved the code to retrieve the cert from the test server to a helper method, to help people understand its purpose.

RainbowMango (approved):
/lgtm
/approve

Looks great! Thanks.

Issue #6051, opened 2025-01-15
【Karmada config && certificates】secret and path naming convention


chaosi-zju:
## Task one. [Karmada Config] secret and path naming convention

**Naming convention:**

* secretName: `${karmada_instance_name}-${component}-config`
* volumeName: `karmada-config`
* volumeMountPath: `/etc/karmada/config`
> `${karmada_instance_name}` defaults to value `karmada`, but theoretically it also could be `karmada-xxx` or `xxx-karmada`.

**Needed PRs:**

* local up: https://github.com/karmada-io/karmada/pull/5679 @chaosi-zju
* karmadactl: https://github.com/karmada-io/karmada/pull/5797 @chaosi-zju
* helm: @help-wanted
* karmada-operator: @help-wanted

**Examples:**

<details>
<summary>Karmada config</summary>

### 1. karmada-aggregated-apiserver
	
```yaml
secretName: karmada-aggregated-apiserver-config
volumeName: karmada-config
volumeMountPath: /etc/karmada/config
command:
  - --kubeconfig=/etc/karmada/config/karmada.config
  - --authentication-kubeconfig=/etc/karmada/config/karmada.config
  - --authorization-kubeconfig=/etc/karmada/config/karmada.config
```

### 2. karmada-controller-manager

```yaml
secretName: karmada-controller-manager-config
volumeName: karmada-config
volumeMountPath: /etc/karmada/config
command:
  - --kubeconfig=/etc/karmada/config/karmada.config
```

### 3. karmada-scheduler

```yaml
secretName: karmada-scheduler-config
volumeName: karmada-config
volumeMountPath: /etc/karmada/config
command:
  - --kubeconfig=/etc/karmada/config/karmada.config
```

### 4. karmada-descheduler

```yaml
secretName: karmada-descheduler-config
volumeName: karmada-config
volumeMountPath: /etc/karmada/config
command:
  - --kubeconfig=/etc/karmada/config/karmada.config
```

### 5. karmada-metrics-adapter

```yaml
secretName: karmada-metrics-adapter-config
volumeName: karmada-config
volumeMountPath: /etc/karmada/config
command:
  - --kubeconfig=/etc/karmada/config/karmada.config
  - --authentication-kubeconfig=/etc/karmada/config/karmada.config
  - --authorization-kubeconfig=/etc/karmada/config/karmada.config
```

### 6. karmada-search

```yaml
secretName: karmada-search-config
volumeName: karmada-config
[...]

PR #5797, opened 2024-11-09
standardize the naming of karmada config in karmadactl method

chaosi-zju:
**What type of PR is this?**

/kind feature
/kind cleanup

**What this PR does / why we need it**:

In karmada, here are two important secrets, which is mount by most karmada components. One is karmada-cert, which contains a series of cert files like ca.crt, apiserver.crt and so on; another is karmada-kubeconfig, which contains a kubeconfig of karmada-apiserver.

However, in different installation methods, we used inconsistent secret naming or file path naming, which can potentially cause some unnecessary problems, detail refer to https://github.com/karmada-io/karmada/issues/5363.

This PR aims to standardize the naming of karmada config in `karmadactl` installation method.

**Which issue(s) this PR fixes**:

Fixes part of #6051

**Special notes for your reviewer**:

karmada-agent is a bit different, this PR doesn't include the rename of karmada-agent

**Does this PR introduce a user-facing change?**:

```release-note
`karmadactl`: standardize the naming of karmada config in karmadactl installation method
```

PR #6135, opened 2025-02-18
Standardize the naming of karmada config in Helm chart

seanlaii:
**What type of PR is this?**
/kind feature
/kind cleanup

In karmada, here are two important secrets, which is mount by most karmada components. One is karmada-cert, which contains a series of cert files like ca.crt, apiserver.crt and so on; another is karmada-kubeconfig, which contains a kubeconfig of karmada-apiserver.

However, in different installation methods, we used inconsistent secret naming or file path naming, which can potentially cause some unnecessary problems, detail refer to https://github.com/karmada-io/karmada/issues/5363.

This PR aims to standardize the naming of karmada config in Helm installation method.

Part of #6051 

For the users who use the kubeconfig-secret in their custom webhook, they might need to change the secret name and the kubeconfig path in their command.

```release-note
`helm`: standardize the naming of karmada config in helm installation method
```

### DIFF
--- a/charts/karmada/templates/_helpers.tpl
+++ b/charts/karmada/templates/_helpers.tpl
@@ -112,15 +112,14 @@ app: {{- include "karmada.name" .}}-kube-controller-manager
 
 {{- define "karmada.kubeconfig.volume" -}}
 {{- $name := include "karmada.name" . -}}
-- name: kubeconfig-secret
+- name: karmada-config
   secret:
-    secretName: {{ $name }}-kubeconfig
+    secretName: {{ $name }}-config
 {{- end -}}
 
 {{- define "karmada.kubeconfig.volumeMount" -}}
-- name: kubeconfig-secret
-  subPath: kubeconfig
-  mountPath: /etc/kubeconfig
+- name: karmada-config
+  mountPath: /etc/karmada/config
 {{- end -}}
 
 {{- define "karmada.kubeconfig.caData" -}}
@@ -197,11 +196,11 @@ app: {{$name}}
 {{- define "karmada.descheduler.kubeconfig.volume" -}}
 {{ $name :=  include "karmada.name" . }}
 {{- if eq .Values.installMode "host" -}}
-- name: kubeconfig-secret
+- name: karmada-config
   secret:
-    secretName: {{ $name }}-kubeconfig
+    secretName: {{ $name }}-descheduler-config
 {{- else -}}
-- name: kubeconfig-secret
+- name: karmada-config
   secret:
     secretName: {{ .Values.descheduler.kubeconfig }}
 {{- end -}}
@@ -324,14 +323,14 @@ app: {{- include "karmada.name" .}}-search
 - name: k8s-certs
   secret:
     secretName: {{ $name }}-cert
-- name: kubeconfig-secret
+- name: karmada-config
   secret:
-    secretName: {{ $name }}-kubeconfig
+    secretName: {{ $name }}-search-config
 {{- else -}}
 - name: k8s-certs
   secret:
     secretName: {{ .Values.search.certs }}
-- name: kubeconfig-secret
+- name: karmada-config
   secret:
     secretName: {{ .Values.search.kubeconfig }}
 {{- end -}}
@@ -628,7 +627,7 @@ Return the proper Docker Image Registry Secret Names
       # case first installation: no `cm/karmada-version` at first, so when you get it, it means `karmada-static-resource-job` finished.
       # case restart: already has `cm/karmada-version`, which means `karmada-static-resource-job` already finished.
       # case upgrading: already has `cm/karmada-version`, but it may be old version, we should wait until `.data.karmadaVersion` equal to current `.Values.karmadaImageVersion`.
-      while [[ $(kubectl --kubeconfig /etc/kubeconfig get configmap karmada-version -n {{ .Values.systemNamespace }} -o jsonpath='{.data.karmadaVersion}') != {{ .Values.karmadaImageVersion }} ]]; do
+      while [[ $(kubectl --kubeconfig /etc/karmada/config/karmada.config get configmap karmada-version -n {{ .Values.systemNamespace }} -o jsonpath='{.data.karmadaVersion}') != {{ .Values.karmadaImageVersion }} ]]; do
         echo "wait for karmada-static-resource-job finished"; sleep 2
       done
 
@@ -636,4 +635,28 @@ Return the proper Docker Image Registry Secret Names
       EOF
   volumeMounts:
     {{- include "karmada.kubeconfig.volumeMount" .| nindent 4 }}
+{{- end -}}
+
+{{- define "karmada.kubeconfigComponents" -}}
+- aggregated-apiserver
+- controller-manager
+- kube-controller-manager
+- scheduler
+- webhook
+- metrics-adapter
+- descheduler
+- search
+{{- end }}
+
+{{- define "karmada.shouldCreateComponentKubeconfig" -}}
+{{- $component := .component }}
+{{- $components := .components }}
+{{- $specialComponentMap := dict "metrics-adapter" "metricsAdapter" "descheduler" "descheduler" "search" "search" }}
+{{- $shouldProcess := true }}
+{{- if hasKey $specialComponentMap $component }}
+  {{- if not (has (get $specialComponentMap $component) $components) }}
+    {{- $shouldProcess = false }}
+  {{- end }}
+{{- end }}
+{{- $shouldProcess }}
 {{- end -}}
--- a/charts/karmada/templates/karmada-kubeconfig.yaml
+++ b/charts/karmada/templates/karmada-kubeconfig.yaml
@@ -3,10 +3,10 @@
 apiVersion: v1
 kind: Secret
 metadata:
-  name: {{ $name }}-kubeconfig
+  name: {{ $name }}-config
   namespace: {{ include "karmada.namespace" . }}
 stringData:
-  kubeconfig: |-
+  karmada.config: |-
     apiVersion: v1
     kind: Config
     clusters:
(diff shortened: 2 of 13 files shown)


charts/karmada/templates/_helpers.tpl
```diff
@@ -637,3 +636,16 @@ Return the proper Docker Image Registry Secret Names
...
+{{- $components := .components }}
+{{- $specialComponentMap := dict "metrics-adapter" "metricsAdapter" "descheduler" "descheduler" "search" "search" }}
+{{- $shouldProcess := true }}
+{{- if hasKey $specialComponentMap $component }}
+  {{- if not (has (get $specialComponentMap $component) $components) }}
+    {{- $shouldProcess = false }}
+  {{- end }}
+{{- end }}
+{{- $shouldProcess }}
+{{- end -}}
```
seanlaii, line 651: Create a template to check if we need to create the component kubeconfig by referencing the components in values.yaml to avoid breaking change.

charts/karmada/templates/_helpers.tpl
```diff
@@ -637,3 +636,27 @@ Return the proper Docker Image Registry Secret Names
   volumeMounts:
     {{- include "karmada.kubeconfig.volumeMount" .| nindent 4 }}
 {{- end -}}
+
+{{- define "karmada.kubeconfigComponents" -}}
```
seanlaii, line 640: Hardcoded the components list to reference the component name.

charts/karmada/templates/pre-install-job.yaml
```diff
@@ -279,14 +279,45 @@ data:
         {{ print "{{ crt }}" }}
       tls.key: |-
         {{ print "{{ key }}" }}
+  {{- range (include "karmada.kubeconfigComponents" . | fromYamlArray) }}
```
seanlaii, line 282: Create `kubeconfig.yaml` for each component.

seanlaii:
Hi @chaosi-zju , please help review this PR. Thank you!
Also, considering your comment in https://github.com/karmada-io/karmada/pull/5797, do we have to change the config for karmada-agent?

This is the result in my local:
image
Screenshot 2025-02-25 at 3 30 06 PM

chaosi-zju:
> Hi @chaosi-zju , please help review this PR. Thank you!

thank you for you quick effort

BTW, have you checked why some component like controller had restarted?

seanlaii:
> > Hi @chaosi-zju , please help review this PR. Thank you!
> 
> thank you for you quick effort
> 
> BTW, have you checked why some component like controller had restarted?

Thanks for the good catch! I appreciate you pointing this out.

I've investigated the issue further and found the following.  After reviewing the controller logs, I observed this error message in the `controller-manager`:

```log 
E0301 20:12:13.682692       1 kind.go:71] "if kind is a CRD, it should be installed before calling Start" err="no matches for kind \"Cluster\" in version \"cluster.karmada.io/v1alpha1\"" logger="controller-runtime.source.EventHandler" kind="Cluster.cluster.karmada.io"
```

Additionally, when running kubectl get cluster against the karmada-apiserver, I received this error which might indicate there is a connection error in `aggregated-apiserver service`:

```log 
Error from server (ServiceUnavailable): the server is currently unable to handle the request (get clusters.cluster.karmada.io)
```

Interestingly, I encountered the same behavior when installing both the **old** and **new** Helm charts. This led me to believe the problem might lie within my installation process. Here are the steps I initially followed:

- Run `./hack/setup-dev-base.sh` to set up the Kind clusters.
- Run `helm install karmada-vetwnpm8z3 charts/karmada --namespace karmada-vetwnpm8z3 --wait --timeout 800s --set 'components={search,descheduler,metricsAdapter},apiServer.hostNetwork=true' --debug` to install the Helm chart.

However, after removing the `hostNetwork` setting and installing the chart using only `helm install karmada charts/karmada --namespace karmada-vetwnpm8z3 --wait --timeout 800s --set 'components={search,descheduler,metricsAdapter}'`, I was able to retrieve the expected results:
Screenshot 2025-03-01 at 4 01 48 PM

Based on these observations, it seems setting `apiServer.hostNetwork=true` is causing the issue. Do you have any insights into why this might be happening, or perhaps have any suggestions for further debugging steps I could take?

Thanks again for your time and assistance!Thanks for your time!

RainbowMango (commented):
/assign

Sorry for letting this sit. I will put it in my queue.

RainbowMango:
I tried to install it as per the [Local installation](https://github.com/karmada-io/karmada/tree/master/charts/karmada#local-installation), it raises a timeout error:
```
-bash-5.0# helm install karmada -n karmada-system --create-namespace --dependency-update ./charts/karmada
Error: INSTALLATION FAILED: failed pre-install: 1 error occurred:
	* timed out waiting for the condition
```

Note that I tried against this patch and master, both results in the timeout. 

Do we need to change the default time out?  @chaosi-zju

seanlaii:
Hi @RainbowMango , thanks for testing the installation!
Could you please provide more details about which specific step timed out?
I just ran a local test using the following commands:
```bash
wei-chenglai@Wei-Chengs-MacBook-Pro karmada % ./hack/setup-dev-base.sh
wei-chenglai@Wei-Chengs-MacBook-Pro karmada % helm install karmada -n karmada-system --create-namespace --dependency-update ./charts/karmada

NAME: karmada
LAST DEPLOYED: Mon Mar 10 21:39:49 2025
NAMESPACE: karmada-system
STATUS: deployed
REVISION: 1
TEST SUITE: None
```
The download itself was successful. However, I did notice that pulling the cfssl image took approximately 4 minutes, which is a bit longer.
If the timeout occurred during image pulling, increasing the Helm timeout might be a solution.

seanlaii:
Sorry, I clicked the wrong button... Please ignore the close of the repo.

charts/karmada/templates/karmada-kubeconfig.yaml
```diff
@@ -3,10 +3,10 @@
 apiVersion: v1
 kind: Secret
 metadata:
-  name: {{ $name }}-kubeconfig
+  name: {{ $name }}-config
```
chaosi-zju, line 6: so, as I get, this old `{{ $name }}-config` is still needed and kept for `kubectl` in `karmada-static-resource-job` and `initContainer` during installation.

charts/karmada/templates/karmada-kubeconfig.yaml
```diff
@@ -26,4 +26,36 @@ stringData:
           user: {{ $name }}-apiserver
         name: {{ $name }}-apiserver
     current-context: {{ $name }}-apiserver
+---
+{{- range $component := (include "karmada.kubeconfigComponents" . | fromYamlArray) }}
+{{- if eq (include "karmada.shouldCreateComponentKubeconfig" (dict "component" $component "components" $.Values.components)) "true" }}
```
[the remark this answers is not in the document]
chaosi-zju: But I still think there is half a duplicate code here. Can't this single `{{$name }}-config` be a member of the kubeconfigComponents?

chaosi-zju (commented):
thanks for you work!

**This is indeed a difficult PR. I have carefully considered the modification of each line. I think the overall logic is rigorous and complete, depending on whether some places can be simplified more.**

charts/karmada/templates/_helpers.tpl
chaosi-zju: I can understand your intention here, but these few lines of high-level syntax seem to be unfriendly to people who are not familiar with helm. I reserve opinions and see if other viewers think it needs to be simplified here.

charts/karmada/templates/karmada-kubeconfig.yaml
chaosi-zju: I can understand your intention here, but these few lines of high-level syntax seem to be unfriendly to people who are not familiar with helm. I reserve opinions and see if other viewers think it needs to be simplified here.

chaosi-zju:
@RainbowMango 

> -bash-5.0# helm install karmada -n karmada-system --create-namespace --dependency-update ./charts/karmada
Error: INSTALLATION FAILED: failed pre-install: 1 error occurred:
	* timed out waiting for the condition

as @seanlaii said, the reason is the time to pull `cfssl` image is so long.

this image is `1.43GB`, and your network bandwidth is so poor, you need `8min 30s` to pull it, while the default timeout for `pre-install` hook job is just `5min`.

> Do we need to change the default time out?

you bandwidth seems as slow as `1Mb/s`, to some extent, a special case.

no other users reported this, maybe no need to change for now.

charts/karmada/templates/_helpers.tpl
seanlaii: What if I create all the `karmada.config` for all the components regardless of the values in `component`, so I can remove this logic?
WDYT? Thank you!

chaosi-zju: > create all the karmada.config for all the components regardless of the values in component

I didn't get it, can you show me a example?

ps: I'm very sorry for the slow reply because I've been busy with other things recently.

seanlaii: No worries!
Apologies for any confusion. The logic in this implementation conditionally generates component `kubeconfig` secrets, {component}-config; a `kubeconfig` secret is only created if its corresponding component is enabled via the component's Helm chart [values](https://github.com/karmada-io/karmada/blob/master/charts/karmada/values.yaml#L47-L52). 
We could choose to generate `kubeconfig` secrets for all components unconditionally. Removing this conditional logic would simplify the code and potentially improve maintainability.

charts/karmada/templates/karmada-search.yaml
```diff
@@ -53,14 +53,13 @@ spec:
             - name: etcd-certs
               mountPath: /etc/etcd/pki
               readOnly: true
-            - name: kubeconfig-secret
-              subPath: kubeconfig
-              mountPath: /etc/kubeconfig
+            - name: karmada-config
+              mountPath: /etc/karmada/config
```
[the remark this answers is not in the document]
seanlaii: Updated!

PR #6137, opened 2025-02-19
fix karmadactl path

sivchari:
**What type of PR is this?**

/kind bug

GOBIN and GOPATH is configurable, so use `which command` to fetch PATH of karmadactl.

Fixes #

```release-note

```

### DIFF
--- a/hack/local-up-karmada.sh
+++ b/hack/local-up-karmada.sh
@@ -66,8 +66,7 @@ export KUBECONFIG="${MAIN_KUBECONFIG}"
 "${REPO_ROOT}"/hack/deploy-karmada.sh "${MAIN_KUBECONFIG}" "${HOST_CLUSTER_NAME}"
 
 # step3. join push mode member clusters and install scheduler-estimator
-GOPATH=$(go env GOPATH | awk -F ':' '{print $1}')
-KARMADACTL_BIN="${GOPATH}/bin/karmadactl"
+KARMADACTL_BIN="$(which karmadactl)"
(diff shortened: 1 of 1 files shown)


RainbowMango:
Hi @sivchari Thanks for spotting, but I'm not sure what problem this PR is supposed to solve, Could you please explain it with more details?

sivchari:
Hi @RainbowMango 

As I said, GOPATH and GOBIN is configurable, then the priority is following when running go install.
1. If the GOBIN is set, the destination is GOBIN
2. if the GOBIN isn't set and GOPATH is set, the destination is GOPATH/bin

Thus if GOBIN is set to another location, karmadactl might not be installed to the following location.

```
GOPATH=$(go env GOPATH | awk -F ':' '{print $1}')
KARMADACTL_BIN="${GOPATH}/bin/karmadactl"
```

It would be good to use `which command` to ensure success in any environment.
I feel free that you close this PR if it's unnecessary change.

Thanks!

RainbowMango:
Thank you for the explanation, I understand that the current implementation requires `GOBIN` to remain as its default value ($GOPATH/bin), otherwise the `KARMADACTL_BIN` will point to a non-existent file. 

However, changing to `which` brings another concern that it requires the `GOBIN` or `GOPATH/bin` already included in the `PATH` environment variable.  What do you think of this situation?

sivchari:
At now, local-up-karmada.sh needs us to locate the karmadactl in $GOPATH/bin in advance. But, as far as I know, there are some [way](https://karmada.io/docs/installation/install-cli-tools) to install karmadactl and it seems to some way not to install to $GOPATH/bin. So I thought it prefers not to specify $GOPATH/bin.

RainbowMango:
@liangyuanpeng Would you like to take a look?

PR #6142, opened 2025-02-20
add some metrics when testing performance

LivingCcj:
**What type of PR is this?**
/kind feature

there is append some import metrics when test karmada performance

**binding_sync_work_total** : means karmada-controller-manage sync work count, to calculate its qps
**work_sync_workload_total**: means karmada-controller-manage or karmada -agent sync workload count, to calculate its qps

### DIFF
--- a/pkg/metrics/resource.go
+++ b/pkg/metrics/resource.go
@@ -29,7 +29,9 @@ const (
 	resourceApplyPolicyDurationMetricsName     = "resource_apply_policy_duration_seconds"
 	policyApplyAttemptsMetricsName             = "policy_apply_attempts_total"
 	syncWorkDurationMetricsName                = "binding_sync_work_duration_seconds"
+	syncWorkTotalMetricsName                   = "binding_sync_work_total"
 	syncWorkloadDurationMetricsName            = "work_sync_workload_duration_seconds"
+	syncWorkloadTotalMetricsName               = "work_sync_workload_total"
 	recreateResourceToCluster                  = "recreate_resource_to_cluster"
 	updateResourceToCluster                    = "update_resource_to_cluster"
 	policyPreemptionMetricsName                = "policy_preemption_total"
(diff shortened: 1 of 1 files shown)


pkg/metrics/resource.go
```diff
@@ -63,12 +65,22 @@ var (
 		Buckets: prometheus.ExponentialBuckets(0.001, 2, 12),
 	}, []string{"result"})
 
+	syncWorkTotalCounter = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: syncWorkTotalMetricsName,
+		Help: "Total number of sync works for ResourceBinding.",
+	}, []string{"result"})
```
RainbowMango, line 71: Did you notice there is already a similar metric? for example:
```
controller_runtime_reconcile_total{controller="execution-controller",result="error"} 10
controller_runtime_reconcile_total{controller="execution-controller",result="requeue"} 0
controller_runtime_reconcile_total{controller="execution-controller",result="requeue_after"} 0
controller_runtime_reconcile_total{controller="execution-controller",result="success"} 6
```
Which shows the total number of reconciliation. Is that you want?

LivingCcj: `controller_runtime_reconcile_total` is a metric which reported by controller-runtime sdk，it means some informations  controller-runtime queue. The new metrics are mainly concerned with the result of `ensureWork` or `syncToClusters`，which can be better for observing the performance with different result.

RainbowMango: Emm, the `work_sync_workload_total` indicates the number of requests against the member clusters. right?

pkg/metrics/resource.go
```diff
@@ -29,7 +29,9 @@ const (
 	resourceApplyPolicyDurationMetricsName     = "resource_apply_policy_duration_seconds"
 	policyApplyAttemptsMetricsName             = "policy_apply_attempts_total"
 	syncWorkDurationMetricsName                = "binding_sync_work_duration_seconds"
+	syncWorkTotalMetricsName                   = "binding_sync_work_total"
```
RainbowMango, line 32: The `binding_sync_work_total` is essentially records how many times the [ensureWork](https://github.com/karmada-io/karmada/blob/af610cc58e03e2ca959143c8bed5887c66450023/pkg/controllers/binding/common.go#L45) is invoked.
Can you remind me what clue we can get from this to observe the performance?

pkg/metrics/resource.go
LivingCcj: Yes, that is a wonderful advice, we ignore this scenario because we use karmada-agent to `pull` the workload，so `work_sync_workload_total ` is reported from sub-cluster.
For `pull` and `push` mode, appending cluster_name label to work_sync_workload_total is more comprehensive.

pkg/metrics/resource.go
LivingCcj: that considers to report the result qps of sync work for binding controler，and there is no distinction between rb and crb.

PR #6148, opened 2025-02-23
update object watcher metrics and simplify cluster object version recording

zach593:
**What type of PR is this?**

/kind cleanup

**What this PR does / why we need it**:

This PR brings more specific operation counters for create/update/delete member cluster objects.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
karmada-controller-manager: metric `recreate_resource_to_cluster` has been merged into `create_resource_to_cluster`.
```

PR #6150, opened 2025-02-23
Add deepequal check for objectWatcher.Update()

zach593:
**What type of PR is this?**

/kind feature

Add deepequal check for objectWatcher.Update() in karmada-controller-manager to reduce redundant HTTP update operations on member clusters, which helps reduce the time consumption of karmada-controller-manager startup.

Fixes #

If #6148 can be merged and this PR can be rebased on it, the changes brought by this PR can be more clearly shown through these metrics.

```release-note
`Performance`/`karmada-controller-manager`: Now when karmada-controller-manager tries to update a resource to a member cluster, it will attempt to compare the contents to skip redundant update operations. The optimization significantly reduces the execution-controller by 80% during the controller start-up.
```

### DIFF
--- a/pkg/util/objectwatcher/objectwatcher.go
+++ b/pkg/util/objectwatcher/objectwatcher.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"sync"
 
+	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -31,6 +32,7 @@ import (
 	configv1alpha1 "github.com/karmada-io/karmada/pkg/apis/config/v1alpha1"
 	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
 	"github.com/karmada-io/karmada/pkg/resourceinterpreter"
+	"github.com/karmada-io/karmada/pkg/resourceinterpreter/default/native/prune"
 	"github.com/karmada-io/karmada/pkg/util"
 	"github.com/karmada-io/karmada/pkg/util/fedinformer/genericmanager"
 	"github.com/karmada-io/karmada/pkg/util/fedinformer/keys"
@@ -56,7 +58,7 @@ type ObjectWatcher interface {
 	Create(ctx context.Context, clusterName string, desireObj *unstructured.Unstructured) error
 	Update(ctx context.Context, clusterName string, desireObj, clusterObj *unstructured.Unstructured) (operationResult OperationResult, err error)
 	Delete(ctx context.Context, clusterName string, desireObj *unstructured.Unstructured) error
-	NeedsUpdate(clusterName string, desiredObj, clusterObj *unstructured.Unstructured) (bool, error)
+	NeedsUpdate(clusterName string, desiredObj, clusterObj *unstructured.Unstructured) bool
 }
 
 // ClientSetFunc is used to generate client set of member cluster
@@ -123,7 +125,7 @@ func (o *objectWatcherImpl) Create(ctx context.Context, clusterName string, desi
 	return nil
 }
 
-func (o *objectWatcherImpl) retainClusterFields(desired, observed *unstructured.Unstructured) (*unstructured.Unstructured, error) {
+func (o *objectWatcherImpl) retainClusterFields(desired, observed *unstructured.Unstructured) *unstructured.Unstructured {
 	// Pass the same ResourceVersion as in the cluster object for update operation, otherwise operation will fail.
 	desired.SetResourceVersion(observed.GetResourceVersion())
 
@@ -143,11 +145,7 @@ func (o *objectWatcherImpl) retainClusterFields(desired, observed *unstructured.
 	// and be set by user in karmada-controller-plane.
 	util.RetainLabels(desired, observed)
 
-	if o.resourceInterpreter.HookEnabled(desired.GroupVersionKind(), configv1alpha1.InterpreterOperationRetain) {
-		return o.resourceInterpreter.Retain(desired, observed)
-	}
-
-	return desired, nil
+	return desired
 }
 
 func (o *objectWatcherImpl) Update(ctx context.Context, clusterName string, desireObj, clusterObj *unstructured.Unstructured) (OperationResult, error) {
@@ -170,10 +168,23 @@ func (o *objectWatcherImpl) Update(ctx context.Context, clusterName string, desi
 		return OperationResultNone, err
 	}
 
-	desireObj, err = o.retainClusterFields(desireObj, clusterObj)
-	if err != nil {
-		klog.Errorf("Failed to retain fields for resource(kind=%s, %s/%s) in cluster %s: %v", clusterObj.GetKind(), clusterObj.GetNamespace(), clusterObj.GetName(), clusterName, err)
-		return OperationResultNone, err
+	desireObj = o.retainClusterFields(desireObj, clusterObj)
+	if o.resourceInterpreter.HookEnabled(desireObj.GroupVersionKind(), configv1alpha1.InterpreterOperationRetain) {
+		desireObj, err = o.resourceInterpreter.Retain(desireObj, clusterObj)
+		if err != nil {
+			klog.Errorf("Failed to retain fields for resource(kind=%s, %s/%s) in cluster %s: %v", clusterObj.GetKind(), clusterObj.GetNamespace(), clusterObj.GetName(), clusterName, err)
+			return OperationResultNone, err
+		}
+	}
+
+	// If there's no actual content changes, skip the update and record the current version.
+	// Whether the DeepEqual check returns true will depend on the implementation of the member cluster's mutating webhook.
+	// In theory, if the member cluster's mutating webhook modifies the content on every update,
+	// then this DeepEqual check may always return false. So the objectWatcher's VersionRecord is still needed.
+	if o.resourceDeepEqual(desireObj, clusterObj, clusterName) {
+		klog.Infof("No need to update resource(kind=%s, %s/%s) in cluster %s because the content has not changed", clusterObj.GetKind(), clusterObj.GetNamespace(), clusterObj.GetName(), clusterName)
+		o.recordVersion(clusterObj, clusterName)
+		return OperationResultNone, nil
 	}
 
 	resource, err := dynamicClusterClient.DynamicClientSet.Resource(gvr).Namespace(desireObj.GetNamespace()).Update(ctx, desireObj, metav1.UpdateOptions{})
@@ -284,15 +295,10 @@ func (o *objectWatcherImpl) deleteVersionRecord(clusterName, resourceName string
 	delete(o.VersionRecord[clusterName], resourceName)
 }
 
-func (o *objectWatcherImpl) NeedsUpdate(clusterName string, desiredObj, clusterObj *unstructured.Unstructured) (bool, error) {
+func (o *objectWatcherImpl) NeedsUpdate(clusterName string, desiredObj, clusterObj *unstructured.Unstructured) bool {
 	// get resource version
-	version, exist := o.getVersionRecord(clusterName, desiredObj.GroupVersionKind().String()+"/"+desiredObj.GetNamespace()+"/"+desiredObj.GetName())
-	if !exist {
-		klog.Errorf("Failed to update the resource(kind=%s, %s/%s) in the cluster %s because the version record does not exist.", desiredObj.GetKind(), desiredObj.GetNamespace(), desiredObj.GetName(), clusterName)
-		return false, fmt.Errorf("failed to update resource(kind=%s, %s/%s) in cluster %s for the version record does not exist", desiredObj.GetKind(), desiredObj.GetNamespace(), desiredObj.GetName(), clusterName)
-	}
-
-	return lifted.ObjectNeedsUpdate(desiredObj, clusterObj, version), nil
+	version, _ := o.getVersionRecord(clusterName, desiredObj.GroupVersionKind().String()+"/"+desiredObj.GetNamespace()+"/"+desiredObj.GetName())
+	return lifted.ObjectNeedsUpdate(desiredObj, clusterObj, version)
 }
 
 func (o *objectWatcherImpl) isManagedResource(clusterObj *unstructured.Unstructured) bool {
@@ -316,4 +322,22 @@ func (o *objectWatcherImpl) allowUpdate(clusterName string, desiredObj, clusterO
 	}
 
 	return false
+}
+
+// resourceDeepEqual checks if the desired object and the cluster object are equal.
+func (o *objectWatcherImpl) resourceDeepEqual(desiredObj, clusterObj *unstructured.Unstructured, clusterName string) bool {
+	clusterObjCopy := clusterObj.DeepCopy()
+	// Remove fields that should not be propagated to member clusters,
+	// applying the same pruning strategy as used for the desired object.
+	err := prune.RemoveIrrelevantFields(clusterObjCopy, prune.RemoveJobTTLSeconds)
+	if err != nil {
+		klog.Warningf("Failed to remove irrelevant fields for cluster resource(kind=%s, %s/%s) in cluster %s: %v", clusterObj.GetKind(), clusterObj.GetNamespace(), clusterObj.GetName(), clusterName, err)
+		return false
+	}
+
+	// Retain fields that are mandatory for update operations,
+	// applying the same retain strategy as used ofr the desired object.
+	clusterObjCopy = o.retainClusterFields(clusterObjCopy, clusterObj)
+
+	return equality.Semantic.DeepEqual(desiredObj, clusterObjCopy)
 }
--- a/pkg/controllers/status/work_status_controller.go
+++ b/pkg/controllers/status/work_status_controller.go
@@ -249,11 +249,7 @@ func (c *WorkStatusController) updateResource(ctx context.Context, observedObj *
 
 	// we should check if the observed status is consistent with the declaration to prevent accidental changes made
 	// in member clusters.
-	needUpdate, err := c.ObjectWatcher.NeedsUpdate(clusterName, desiredObj, observedObj)
-	if err != nil {
-		return err
-	}
-
+	needUpdate := c.ObjectWatcher.NeedsUpdate(clusterName, desiredObj, observedObj)
 	if needUpdate {
 		operationResult, updateErr := c.ObjectWatcher.Update(ctx, clusterName, desiredObj, observedObj)
 		metrics.CountUpdateResourceToCluster(updateErr, desiredObj.GetAPIVersion(), desiredObj.GetKind(), clusterName, string(operationResult))


RainbowMango:
> If https://github.com/karmada-io/karmada/pull/6148 can be merged and this PR can be rebased on it, the changes brought by this PR can be more clearly shown through these metrics.

FYI. The #6148 has been merged.

zach593:
> FYI. The #6148 has been merged.

Got it, I just rebased the commit. It may need in-depth testing, I'll post the test results ASAP.

RainbowMango (commented):
/assign

pkg/util/objectwatcher/objectwatcher.go
```diff
@@ -172,10 +174,33 @@ func (o *objectWatcherImpl) Update(ctx context.Context, clusterName string, desi
 
 	desireObj, err = o.retainClusterFields(desireObj, clusterObj)
 	if err != nil {
-		klog.Errorf("Failed to retain fields for resource(kind=%s, %s/%s) in cluster %s: %v", clusterObj.GetKind(), clusterObj.GetNamespace(), clusterObj.GetName(), clusterName, err)
+		klog.Errorf("Failed to retain fields for desired resource(kind=%s, %s/%s) in cluster %s: %v", clusterObj.GetKind(), clusterObj.GetNamespace(), clusterObj.GetName(), clusterName, err)
```
RainbowMango, line 177: The `desired` might not be correct. It's hard to say the retain operation is against the desired or the observed object.

pkg/util/objectwatcher/objectwatcher.go
```diff
@@ -172,10 +174,33 @@ func (o *objectWatcherImpl) Update(ctx context.Context, clusterName string, desi
...
+		klog.Errorf("Failed to remove irrelevant fields for cluster resource(kind=%s, %s/%s) in cluster %s: %v", clusterObj.GetKind(), clusterObj.GetNamespace(), clusterObj.GetName(), clusterName, err)
+		return OperationResultNone, err
+	}
+
+	// call retain interpret hook again.
+	clusterObjCopy, err = o.retainClusterFields(clusterObjCopy, clusterObj)
+	if err != nil {
+		klog.Errorf("Failed to retain fields for cluster resource(kind=%s, %s/%s) in cluster %s: %v", clusterObj.GetKind(), clusterObj.GetNamespace(), clusterObj.GetName(), clusterName, err)
+		return OperationResultNone, err
+	}
```
RainbowMango, line 193: This (call retain interpret hook) is concerning, as it is a remote call if the interpreter is implemented as a webhook, and in such as a case, it's an expensive call.

pkg/util/objectwatcher/objectwatcher.go
```diff
@@ -172,10 +174,33 @@ func (o *objectWatcherImpl) Update(ctx context.Context, clusterName string, desi
...
+
+	// If there's no actual content changes, skip the update and record the current version.
+	// if there's some strange mutating webhook in member cluster, theoretically this DeepEqual check will fail,
+	// so objectWatcher's VersionRecord is still needed.
```
RainbowMango, line 197: Can you elaborate on what kind of mutating logic?

RainbowMango (commented):
I might need to wait for the test reports to compare the results before and after optimization.

pkg/util/objectwatcher/objectwatcher.go
zach593: I guess since it returns `desireObj`, it shouldn't be a big problem?

pkg/util/objectwatcher/objectwatcher.go
zach593: This is equivalent to replacing the request to the member cluster apiserver with a request to interpretwebhook, since the request to interpretwebhook does not occupy the rate limit, there should still be some benefits.

pkg/util/objectwatcher/objectwatcher.go
zach593: Let's take a most common example:

For native resources, objects will be mutated when entering karmada-apiserver, so when they are updated into member clusters by objectWatcher, they will perform the mutating logic again, which may be exactly the same with the one in karmada-apiserver, so deepequal could return true.

But for custom resources, when cluster administrators integrate karmada, they may only deploy mutating webhooks in member clusters, and desiredObject will not be mutated, while clusterObject is always mutated, so deepequal may always return false.

zach593: In a very extreme case, adding deepequal may be harmful.

Suppose there is such a webhook in the member cluster, which will mutate the value of a field to 1 when the input is 0, and mutate it to 2 when the input is 1.

Then if the value of this field of this object in the member cluster is 1, the value of this field in desiredObject is also 1. Without deepequal, the update will succeed and change the value of the member cluster to 2. Then after adding deepequal, because deepequal will return true, the update is skipped, and this field of this object in the member cluster cannot be modified to 2.

Of course, if a mutatingwebhook is implemented in this way, I think it is very, very strange.

zach593:
test comparison

setup:

- run `hack/local-up-karmada.sh` in current master branch
- apply yaml files in `samples/nginx/`
- restart karmada-controller-manager

without this commit:

image
image

with this commit:

image
image

pkg/util/objectwatcher/objectwatcher.go
RainbowMango: It's no big deal, but we'd better revert this change, because retain is not for `desired` object.

zach593: sure

zach593:
after updated to [a472610](https://github.com/karmada-io/karmada/commit/a4726100e874b7e30c5167608d28c073988e3726) , the equality check still could work.

image
image

pkg/util/objectwatcher/objectwatcher.go
```diff
@@ -170,11 +168,25 @@ func (o *objectWatcherImpl) Update(ctx context.Context, clusterName string, desi
...
+	desireObj = o.retainClusterFields(desireObj, clusterObj)
+	if o.resourceInterpreter.HookEnabled(desireObj.GroupVersionKind(), configv1alpha1.InterpreterOperationRetain) {
+		desireObj, err = o.resourceInterpreter.Retain(desireObj, clusterObj)
+		if err != nil {
+			klog.Errorf("Failed to retain fields for resource(kind=%s, %s/%s) in cluster %s: %v", clusterObj.GetKind(), clusterObj.GetNamespace(), clusterObj.GetName(), clusterName, err)
+			return OperationResultNone, err
+		}
+	}
+
+	deepEqual, err := o.resourceDeepEqual(desireObj, clusterObj, clusterName)
```
RainbowMango, line 180: As talked at yesterday's meeting, I want to know the difference between `desiredObj` and `clusterObj` at this time.

In addition, have you ever thought of comparing `desiredObj` and `prune.RemoveIrrelevantFields(clusterObj)`?

zach593: The test result with deployment is that before this call, desireObj will lack `creationTimestamp`, `generation`, `managedFields`, `resourceVersion`, `uid` and `status` compared to clusterObj.

zach593: > In addition, have you ever thought of comparing `desiredObj` and `clusterObj)`?

If only call  `prune.RemoveIrrelevantFields()` without calling `retainClusterFields()`, clusterObj will definitely be missing `resourceVersion`, and may also be missing `finalizers` and `ownerReferences` (if the resource uses them).

After moving the retain hook out of `retainClusterFields()`, calling this function again on clusterObj should have little impact.

XiShanYongYe-Chang: Taking deployment resources as an example, when I modify the labels of deployment resources in the control plane, the difference between the two obtained is:
```diff
5a6
>                       "deployment.kubernetes.io/revision": "1",
13c14
<                       "resourcetemplate.karmada.io/managed-labels": "app,cz,karmada.io/managed,propagationpolicy.karmada.io/permanent-id,resourcebinding.karmada.io/permanent-id,work.karmada.io/permanent-id",
---
>                       "resourcetemplate.karmada.io/managed-labels": "app,karmada.io/managed,propagationpolicy.karmada.io/permanent-id,resourcebinding.karmada.io/permanent-id,work.karmada.io/permanent-id",
18a20,21
>               "creationTimestamp": "2025-04-09T08:07:32Z",
>               "generation": 1,
21d23
<                       "cz": "test",
26a29,144
>               "managedFields": [{
>                       "apiVersion": "apps/v1",
>                       "fieldsType": "FieldsV1",
>                       "fieldsV1": {
>                               "f:metadata": {
>                                       "f:annotations": {
>                                               ".": {},
>                                               "f:kubectl.kubernetes.io/last-applied-configuration": {},
>                                               "f:propagationpolicy.karmada.io/name": {},
>                                               "f:propagationpolicy.karmada.io/namespace": {},
>                                               "f:resourcebinding.karmada.io/name": {},
>                                               "f:resourcebinding.karmada.io/namespace": {},
>                                               "f:resourcetemplate.karmada.io/generation": {},
>                                               "f:resourcetemplate.karmada.io/managed-annotations": {},
>                                               "f:resourcetemplate.karmada.io/managed-labels": {},
>                                               "f:resourcetemplate.karmada.io/uid": {},
>                                               "f:work.karmada.io/conflict-resolution": {},
>                                               "f:work.karmada.io/name": {},
>                                               "f:work.karmada.io/namespace": {}
>                                       },
>                                       "f:labels": {
>                                               ".": {},
>                                               "f:app": {},
>                                               "f:karmada.io/managed": {},
>                                               "f:propagationpolicy.karmada.io/permanent-id": {},
>                                               "f:resourcebinding.karmada.io/permanent-id": {},
>                                               "f:work.karmada.io/permanent-id": {}
>                                       }
>                               },
>                               "f:spec": {
>                                       "f:progressDeadlineSeconds": {},
>                                       "f:replicas": {},
>                                       "f:revisionHistoryLimit": {},
>                                       "f:selector": {},
>                                       "f:strategy": {
>                                               "f:rollingUpdate": {
>                                                       ".": {},
>                                                       "f:maxSurge": {},
>                                                       "f:maxUnavailable": {}
>                                               },
>                                               "f:type": {}
>                                       },
>                                       "f:template": {
>                                               "f:metadata": {
>                                                       "f:labels": {
>                                                               ".": {},
>                                                               "f:app": {}
>                                                       }
>                                               },
>                                               "f:spec": {
>                                                       "f:containers": {
>                                                               "k:{\"name\":\"nginx\"}": {
>                                                                       ".": {},
>                                                                       "f:image": {},
>                                                                       "f:imagePullPolicy": {},
>                                                                       "f:name": {},
>                                                                       "f:resources": {},
>                                                                       "f:terminationMessagePath": {},
>                                                                       "f:terminationMessagePolicy": {}
>                                                               }
>                                                       },
>                                                       "f:dnsPolicy": {},
>                                                       "f:restartPolicy": {},
>                                                       "f:schedulerName": {},
>                                                       "f:securityContext": {},
>                                                       "f:terminationGracePeriodSeconds": {}
>                                               }
>                                       }
>                               }
>                       },
>                       "manager": "karmada-controller-manager",
>                       "operation": "Update",
>                       "time": "2025-04-09T08:07:32Z"
>               }, {
>                       "apiVersion": "apps/v1",
>                       "fieldsType": "FieldsV1",
>                       "fieldsV1": {
>                               "f:metadata": {
>                                       "f:annotations": {
>                                               "f:deployment.kubernetes.io/revision": {}
>                                       }
>                               },
>                               "f:status": {
>                                       "f:availableReplicas": {},
>                                       "f:conditions": {
>                                               ".": {},
>                                               "k:{\"type\":\"Available\"}": {
>                                                       ".": {},
>                                                       "f:lastTransitionTime": {},
>                                                       "f:lastUpdateTime": {},
>                                                       "f:message": {},
>                                                       "f:reason": {},
>                                                       "f:status": {},
>                                                       "f:type": {}
>                                               },
>                                               "k:{\"type\":\"Progressing\"}": {
>                                                       ".": {},
>                                                       "f:lastTransitionTime": {},
>                                                       "f:lastUpdateTime": {},
>                                                       "f:message": {},
>                                                       "f:reason": {},
>                                                       "f:status": {},
>                                                       "f:type": {}
>                                               }
>                                       },
>                                       "f:observedGeneration": {},
>                                       "f:readyReplicas": {},
>                                       "f:replicas": {},
>                                       "f:updatedReplicas": {}
>                               }
>                       },
>                       "manager": "kube-controller-manager",
>                       "operation": "Update",
>                       "subresource": "status",
>                       "time": "2025-04-09T08:07:36Z"
>               }],
28c146,148
<               "namespace": "default"
---
>               "namespace": "default",
>               "resourceVersion": "8209",
>               "uid": "d1c6a331-312e-4d5a-a416-b22ae602199e"
68a189,210
>       },
>       "status": {
>               "availableReplicas": 1,
>               "conditions": [{
>                       "lastTransitionTime": "2025-04-09T08:07:36Z",
>                       "lastUpdateTime": "2025-04-09T08:07:36Z",
>                       "message": "Deployment has minimum availability.",
>                       "reason": "MinimumReplicasAvailable",
>                       "status": "True",
>                       "type": "Available"
>               }, {
>                       "lastTransitionTime": "2025-04-09T08:07:32Z",
>                       "lastUpdateTime": "2025-04-09T08:07:36Z",
>                       "message": "ReplicaSet \"nginx-676b6c5bbc\" has successfully progressed.",
>                       "reason": "NewReplicaSetAvailable",
>                       "status": "True",
>                       "type": "Progressing"
>               }],
>               "observedGeneration": 1,
>               "readyReplicas": 1,
>               "replicas": 1,
>               "updatedReplicas": 1
```

XiShanYongYe-Chang: I add the log like this:

```diff
diff --git a/pkg/util/objectwatcher/objectwatcher.go b/pkg/util/objectwatcher/objectwatcher.go
index 02df1e2cc..6bf2514fa 100644
--- a/pkg/util/objectwatcher/objectwatcher.go
+++ b/pkg/util/objectwatcher/objectwatcher.go
@@ -149,6 +149,10 @@ func (o *objectWatcherImpl) retainClusterFields(desired, observed *unstructured.
 }

 func (o *objectWatcherImpl) Update(ctx context.Context, clusterName string, desireObj, clusterObj *unstructured.Unstructured) (OperationResult, error) {
+       dObj, _ := desireObj.DeepCopy().MarshalJSON()
+       cObj, _ := clusterObj.DeepCopy().MarshalJSON()
+       klog.Infof("desireObj: %s", string(dObj))
+       klog.Infof("clusterObj: %s", string(cObj))
        updateAllowed := o.allowUpdate(clusterName, desireObj, clusterObj)
        if !updateAllowed {
                // The existing resource is not managed by Karmada, and no conflict resolution found, avoid updating the existing resource by default.
```

pkg/util/objectwatcher/objectwatcher.go
```diff
@@ -317,3 +324,20 @@ func (o *objectWatcherImpl) allowUpdate(clusterName string, desiredObj, clusterO
...
+	if err != nil {
+		klog.Errorf("Failed to remove irrelevant fields for cluster resource(kind=%s, %s/%s) in cluster %s: %v", clusterObj.GetKind(), clusterObj.GetNamespace(), clusterObj.GetName(), clusterName, err)
+		return false, err
+	}
+
+	clusterObjCopy = o.retainClusterFields(clusterObjCopy, clusterObj)
+
+	// If there's no actual content changes, skip the update and record the current version.
+	// if there's some strange mutating webhook in member cluster, theoretically this DeepEqual check will fail,
+	// so objectWatcher's VersionRecord is still needed.
```
RainbowMango, line 341: 1. This method focuses on checking if the two objects are equal; it doesn't care if the version should be recorded or not. So, we can move the comment outside of this method to where the version is recorded.
2. I still don't get what the `strage mutating webhook` mean, and why deep equal check would fail.

pkg/util/objectwatcher/objectwatcher.go
```diff
@@ -317,3 +324,20 @@ func (o *objectWatcherImpl) allowUpdate(clusterName string, desiredObj, clusterO
 
 	return false
 }
+
+// resourceDeepEqual checks if the desired object and the cluster object are equal.
+func (o *objectWatcherImpl) resourceDeepEqual(desiredObj, clusterObj *unstructured.Unstructured, clusterName string) (bool, error) {
```
RainbowMango, line 329:
```suggestion
func (o *objectWatcherImpl) resourceDeepEqual(desiredObj, clusterObj *unstructured.Unstructured, clusterName string) (bool) {
```

I think this method should be non-disruptive, in other words, it defaults to returning false on any unexpected errors, let the program continue running.

pkg/util/objectwatcher/objectwatcher.go
```diff
@@ -317,3 +324,20 @@ func (o *objectWatcherImpl) allowUpdate(clusterName string, desiredObj, clusterO
...
+	clusterObjCopy := clusterObj.DeepCopy()
+	err := prune.RemoveIrrelevantFields(clusterObjCopy, prune.RemoveJobTTLSeconds)
```
RainbowMango, line 331:
```suggestion
	// Remove fields that should not be propagated to member clusters,
	// applying the same pruning strategy as used for the desired object.
	err := prune.RemoveIrrelevantFields(clusterObjCopy, prune.RemoveJobTTLSeconds)
```

pkg/util/objectwatcher/objectwatcher.go
```diff
@@ -317,3 +324,20 @@ func (o *objectWatcherImpl) allowUpdate(clusterName string, desiredObj, clusterO
...
+	if err != nil {
+		klog.Errorf("Failed to remove irrelevant fields for cluster resource(kind=%s, %s/%s) in cluster %s: %v", clusterObj.GetKind(), clusterObj.GetNamespace(), clusterObj.GetName(), clusterName, err)
+		return false, err
+	}
+
+	clusterObjCopy = o.retainClusterFields(clusterObjCopy, clusterObj)
```
RainbowMango, line 337:
```suggestion
	// Retain fields that are mandatory for update operations,
	// applying the same retain strategy as used ofr the desired object.
	clusterObjCopy = o.retainClusterFields(clusterObjCopy, clusterObj)
```

pkg/util/objectwatcher/objectwatcher.go
RainbowMango: We should add the log just before the line for updating, and I've run a test on my side, it shows the followings fields are in `clusterObject`, but not in `desiredObject`:
```
.metadata.generation
.metadata.creationTimestamp
.metadata.managedFields
.metadata.uid
.status
```

pkg/util/objectwatcher/objectwatcher.go
zach593: > and why deep equal check would fail.

"Fail" means the DeepEqual check will return false.

> I still don't get what the strage mutating webhook mean

There’s some explanation in https://github.com/karmada-io/karmada/pull/6150#discussion_r1989292491, though it might indeed be not entirely clear.
The term “strange webhook” refers to webhooks that always mutate the resource during update operations.

One possible scenario could be a webhook that injects the current cluster name into the resource's labels, and is deployed in both the control plane and member clusters.
In such an environment, the deepEqual check will always return false, because the desiredObject would carry the cluster name of the control plane, while the clusterObj carries the member cluster’s name.

BTW, I also updated the comment to make it clearer.

RainbowMango (commented):
/lgtm

Do we have a test report? I believe this patch would improve performance significantly, it should be high-lighted in release notes.

zach593:
> Do we have a test report? I believe this patch would improve performance significantly, it should be high-lighted in release notes.

I'll make one this week.

This PR can theoretically reduce the performance pressure of the member cluster apiserver and reduce the average time cost of the `execution controller` and `work status controller`.
Based on #6192, if the objectWatcher is limited by the rate limiter, it can also reduce this limitation. However, because #6192 has not been merged yet, that part is not suitable to be regarded as an optimization brought by this PR. :smile:

zach593:
### Performance Test Report

To reduce the influence of other controllers on work-related controllers and to extend the processing time for observation purposes, a rather extreme configuration was used:

image

The test data consisted of 20,000 Deployments distributed across 2 member clusters.

The impact of this commit:

- During startup, no HTTP Update requests will be sent to member clusters anymore.
- The CPU usage of the member clusters’ kube-apiserver can be reduced from 2 cores to 1.5 cores.
- The reconciliation time cost of execution controller can be reduced from 3 ms to 556 µs, a decrease of about 80%.
- It ensures that the work-status controller works properly during startup, instead of failing and placing items into the rate limit queue (seems we accidentally fix a bug).
- The CPU usage of the karmada controller-manager has increased slightly, which is reasonable due to the introduction of `DeepEqual`.
- The CPU usage of the karmada-apiserver has also increased. It is suspected to be related to the fact that the work-status controller is now functioning properly. Further verification is needed.

<details>
<summary> Click to view detailed report </summary>

### Without this commit
It took 2 minutes to clear the queue.
For the reconciliation time per execution controller, the P90 and average were: 9.19ms, 2.98ms.

image

However, it can be observed that the work-status controller cleared its queue within 30 seconds, but there were still a large number of items in the rate limit queue.
The reason is that the work-status processing failed at the beginning when the controller just started.

image

In `objectWatcher.Update()`, the update operations on the member clusters did not cause any actual modifications.

image

During startup:
- Member clusters’ kube-apiserver CPU peaked at 0.5 cores, memory usage peaked at 2GB.
- In the karmada cluster, the controller-manager's CPU peaked at 0.9 cores, memory usage peaked at 4GB.
- The karmada-apiserver's CPU peaked at 0.5 cores, and memory usage peaked at 9.5GB.

image

### With this commit

It took 3 minutes to clear the queue.  
For the reconciliation time per execution controller, the P90 and average were: 915 µs, 556 µs.

It is obvious that the work-status controller was functioning properly this time, leading to an increase in processing time. However, it appears that the execution controller's queue was cleared within just 20 seconds.

image

In `objectWatcher.Update()`, no HTTP update requests to the member clusters were triggered.
image

During startup:
- Member clusters’ kube-apiserver CPU peaked at 0.15 cores, memory usage peaked at 2GB.
- In the karmada cluster, the controller-manager's CPU peaked at 1.1 cores, memory usage peaked at 4.2GB.
- The karmada-apiserver's CPU peaked at 1.3 cores, and memory usage peaked at 9.2GB.

image

</details>

RainbowMango (approved):
/approve
GOOD JOB!

Issue #6154, opened 2025-02-24
Fix the invalid behavior in the detector to determine whether the binding object needs to be updated.


CharlesQQ:
**What would you like to be added**:
After restart karmada controller, detector will always update some rb resource, those resourcebindings have dependent resources
```
I0224 17:12:12.236354   17981 detector.go:] Update ResourceBinding(default/xxxx-deployment) successfully.
```

I add new log to compare the changes in resources before updating; The reason is that the finalizers field of rb has changed from `karmada.io/binding-controller,karmada.io/binding-dependencies-distributor` to `karmada.io/binding-controller`
```
	existing := obj.DeepCopyObject()
	if err := mutate(f, key, obj); err != nil {
		return controllerutil.OperationResultNone, err
	}
	patchBytes, err := generateJSONPatch(existing, obj)
	if err != nil {
		panic(err)
	}
	diff := cmp.Diff(existing, obj)
	klog.Infof("after mutate,   %s , equal: %v \n", string(patchBytes), equality.Semantic.DeepEqual(existing, obj))
	if equality.Semantic.DeepEqual(existing, obj) {
		return controllerutil.OperationResultNone, nil
	}

	if err := c.Update(ctx, obj); err != nil {
		return controllerutil.OperationResultNone, err
	}
	return controllerutil.OperationResultUpdated, nil
```

![Image](https://github.com/user-attachments/assets/1960ead5-7d6c-4a54-a108-6ca16a9f28e8)

**Why is this needed**:
Performance optimization, reducing invalid updates to existing resources during restart and reducing work queue resource pressure

XiShanYongYe-Chang:
In favor of #6157 
/assign @CharlesQQ

RainbowMango:
/remove-kind feature
/kind bug

This is more like a bug.

PR #4983, opened 2024-05-24
add binding dependencies distributor finalizer to the independent binding

XiShanYongYe-Chang:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

In dependencies distributor module, we need to deal with independent binding cleanup with the right pattern -- control with finalizer, so I add a new finalizer `karmada.io/binding-dependencies-distributor`.

In addition, in the issue #1741, we can see that the label value of `resourcebinding.karmada.io/depended-by-xxx` will exceed 63 characters, we need to use the binding permanent ID to replace it, this also prompts us to use finalizer to clean up resources.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #6157, opened 2025-02-24
fix(detector): Fix the invalid behavior in the detector to determine whether the binding object needs to be updated.

CharlesQQ:
**What type of PR is this?**
/kind feature

Fixes https://github.com/karmada-io/karmada/issues/6154

```release-note

`karmada-controller-manager`: Fixed the issue where the `detector` unnecessary updates for RB issue.

```

### DIFF
--- a/pkg/detector/detector.go
+++ b/pkg/detector/detector.go
@@ -466,6 +466,7 @@ func (d *ResourceDetector) ApplyPolicy(object *unstructured.Unstructured, object
 			// Just update necessary fields, especially avoid modifying Spec.Clusters which is scheduling result, if already exists.
 			bindingCopy.Annotations = util.DedupeAndMergeAnnotations(bindingCopy.Annotations, binding.Annotations)
 			bindingCopy.Labels = util.DedupeAndMergeLabels(bindingCopy.Labels, binding.Labels)
+			bindingCopy.Finalizers = util.DedupeAndMergeFinalizers(bindingCopy.Finalizers, binding.Finalizers)
 			bindingCopy.OwnerReferences = binding.OwnerReferences
 			bindingCopy.Finalizers = binding.Finalizers
 			bindingCopy.Spec.Resource = binding.Spec.Resource
@@ -560,6 +561,7 @@ func (d *ResourceDetector) ApplyClusterPolicy(object *unstructured.Unstructured,
 				// Just update necessary fields, especially avoid modifying Spec.Clusters which is scheduling result, if already exists.
 				bindingCopy.Annotations = util.DedupeAndMergeAnnotations(bindingCopy.Annotations, binding.Annotations)
 				bindingCopy.Labels = util.DedupeAndMergeLabels(bindingCopy.Labels, binding.Labels)
+				bindingCopy.Finalizers = util.DedupeAndMergeFinalizers(bindingCopy.Finalizers, binding.Finalizers)
 				bindingCopy.OwnerReferences = binding.OwnerReferences
 				bindingCopy.Finalizers = binding.Finalizers
 				bindingCopy.Spec.Resource = binding.Spec.Resource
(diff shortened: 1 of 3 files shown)


XiShanYongYe-Chang:
Nice catching @CharlesQQ 

We add a new finalizer `karmada.io/binding-dependencies-distributor` to the binding object corresponding to the resource to be followed in the dependencies distributor in #4983, this results in inconsistency of finalizers when the binding object is updated in the detector.

Based on this change, when the controller restarts, does the check on the binding object deepequal take effect? Do you have the test result?

CharlesQQ:
> does the check on the binding object deepequal take effect? Do you have the test result?

After testing, the detector will not update rb that has not changed during the restart, and the results are in line with expectations.

image

XiShanYongYe-Chang (commented):
Thanks~
Do we need a release note?
/lgtm

/cc @RainbowMango

RainbowMango (commented):
/assign

pkg/util/label.go
```diff
@@ -124,3 +125,20 @@ func RecordManagedLabels(object *unstructured.Unstructured) {
 	annotations[workv1alpha2.ManagedLabels] = strings.Join(managedKeys, ",")
 	object.SetAnnotations(annotations)
 }
+
+// DedupeAndMergeFinalizers merges the new finalizers into exist finalizers.
+func DedupeAndMergeFinalizers(existFinalizers, newFinalizers []string) []string {
+	if existFinalizers == nil {
+		return existFinalizers
+	}
```
RainbowMango, line 133: I don't understand why it needs a return here, but I don't think it should be.
Consider one situation where len(existFinalizer) != 0 and len(newFinalizers) > 0.

pkg/util/label.go
```diff
@@ -124,3 +125,20 @@ func RecordManagedLabels(object *unstructured.Unstructured) {
...
+	existFinalizerSets := sets.Set[string]{}
+	existFinalizerSets.Insert(existFinalizers...)
+
+	newFinalizerSets := sets.Set[string]{}
+	newFinalizerSets.Insert(existFinalizers...)
+	newFinalizerSets.Insert(newFinalizers...)
+	if equality.Semantic.DeepEqual(existFinalizerSets, newFinalizerSets) {
+		return existFinalizers
+	}
+	return newFinalizerSets.UnsortedList()
```
RainbowMango, line 143: Here should return a sorted list, right?

RainbowMango (changes requested):

pkg/util/label.go
RainbowMango: The order should be 

items in `exist`  --> items in `new`(but not in `exist`)

CharlesQQ: I'll fix it, pls review again  @RainbowMango

XiShanYongYe-Chang:
I update the labels.
/kind bug
/remove-kind feature

XiShanYongYe-Chang:
@CharlesQQ can you help add a release note? We need it when we cherry-pick it.

pkg/detector/detector.go
```diff
@@ -560,6 +560,7 @@ func (d *ResourceDetector) ApplyClusterPolicy(object *unstructured.Unstructured,
 				// Just update necessary fields, especially avoid modifying Spec.Clusters which is scheduling result, if already exists.
 				bindingCopy.Annotations = util.DedupeAndMergeAnnotations(bindingCopy.Annotations, binding.Annotations)
 				bindingCopy.Labels = util.DedupeAndMergeLabels(bindingCopy.Labels, binding.Labels)
+				bindingCopy.Finalizers = util.DedupeAndMergeFinalizers(bindingCopy.Finalizers, binding.Finalizers)
```
RainbowMango, line 563: Is this the only place that needs modification?
My understanding is that after the detector creates ResourceBindng/ClusterResourceBinding, downstream controllers could potentially append finalizers. Therefore, I think all places where updating RB/CRB should be fixed.

@XiShanYongYe-Chang What's your opinion?

CharlesQQ: >  I think all places where updating RB/CRB should be fixed

Oh

RainbowMango: One more place:
https://github.com/karmada-io/karmada/blob/3ae9f5e99f3153ff52f90a63af004929a51ae63f/pkg/detector/detector.go#L470-L469

CharlesQQ: done

CharlesQQ:
> can you help add a release note? We need it when we cherry-pick it.

Which release should I cherry-pick to?  @XiShanYongYe-Chang

pkg/detector/detector.go
XiShanYongYe-Chang: For the CRB, it is OK not to modify the CRB, because the dependency distribution does not add the related finalizer to the CRB resource. 

Of course, the modification is better, which can avoid the same problem in the future.

XiShanYongYe-Chang:
> Which release should I cherry-pick to? @XiShanYongYe-Chang

We introduced this problem in v1.10, and we need to maintain the last three versions, so we need to cherry-pick it to the release-1.12, release-1.11 and release-1.10 branch.

RainbowMango (approved):
/lgtm
/approve

PR #6158, opened 2025-02-24
operator: fix the issue that the default values of Karmada are not correctly applied

zhzhuang-zju:
**What type of PR is this?**
/kind bug

When the operator reconciles the Karmada CR, it will first validate the Karmada CR and then set the default values. This causes the validation to fail when users omit the configuration of certain fields.

It is necessary to set the default values first and then perform the validation. 

Fixes #

How to reproduce it:
1. deploy karmada-operator
```bash
$ hack/deploy-karmada-operator.sh {path/to/kubeconfig} {context}
```

2. deploy a karmada CR
```bash
$ kubectl apply -v=4 -f - <<EOF
kind: Karmada
metadata:
  name: karmada-demo
  namespace: test
spec:
  components:
    karmadaDescheduler: {}
    etcd: {}
EOF
```

3. view operator log
```bash
E0224 11:56:27.802480       1 controller.go:96] Validation failed for karmada: validation errors: [spec.components.etcd: Invalid value: v1alpha1.Etcd{Local:(*v1alpha1.LocalEtcd)(nil), External:(*v1alpha1.ExternalEtcd)(nil)}: unexpected empty etcd configuration]
```

```release-note
`karmada-operator`: fix the error of validation failure when some fields of the Karmada CR are not configured. 
```

### DIFF
--- a/operator/pkg/controller/karmada/controller.go
+++ b/operator/pkg/controller/karmada/controller.go
@@ -92,13 +92,13 @@ func (ctrl *Controller) Reconcile(ctx context.Context, req controllerruntime.Req
 		return ctrl.removeFinalizer(ctx, karmada)
 	}
 
+	if err := ctrl.updateKarmada(ctx, karmada); err != nil {
+		return controllerruntime.Result{}, err
+	}
+
 	if err := ctrl.validateKarmada(ctx, karmada); err != nil {
 		klog.Errorf("Validation failed for karmada: %+v", err)
 		return controllerruntime.Result{}, nil
-	}
-
-	if err := ctrl.ensureKarmada(ctx, karmada); err != nil {
-		return controllerruntime.Result{}, err
 	}
 
 	return controllerruntime.Result{}, ctrl.syncKarmada(karmada)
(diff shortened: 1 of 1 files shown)


operator/pkg/controller/karmada/controller.go
```diff
@@ -140,12 +143,7 @@ func (ctrl *Controller) ensureKarmada(ctx context.Context, karmada *operatorv1al
 		updated = true
 	}
 
-	older := karmada.DeepCopy()
-
-	// Set the defaults for karmada
-	operatorscheme.Scheme.Default(karmada)
-
-	if updated || !reflect.DeepEqual(karmada.Spec, older.Spec) {
+	if updated || isSpecUpdated {
```
XiShanYongYe-Chang, line 146: Can the update operation also be placed before validateKarmada?

XiShanYongYe-Chang: I'm not sure if it's appropriate. Can this update be placed before validate?

https://github.com/karmada-io/karmada/blob/006cf70132c953a522f376cc26418802b74e792c/operator/pkg/controller/karmada/controller.go#L136-L141

operator/pkg/controller/karmada/controller.go
```diff
@@ -92,12 +92,15 @@ func (ctrl *Controller) Reconcile(ctx context.Context, req controllerruntime.Req
 		return ctrl.removeFinalizer(ctx, karmada)
 	}
 
+	older := karmada.DeepCopy()
+	// Set the defaults for karmada
+	operatorscheme.Scheme.Default(karmada)
```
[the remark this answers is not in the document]
RainbowMango: I have the same feeling~

operator/pkg/controller/karmada/controller.go
zhzhuang-zju: > I'm not sure if it's appropriate. Can this update be placed before validate?
@XiShanYongYe-Chang updated,  does it meet your expectations?

XiShanYongYe-Chang: Thanks~
/lgtm

RainbowMango (approved):
/lgtm
/approve

PR #6163, opened 2025-02-26
Make priorityClassSource in the SchedulePriority of PropagationPolicy a required field

seanlaii:
**What type of PR is this?**
/kind api-change
/kind feature

Making it a required field allows for more flexibility in future changes. For example, if we need to modify the default value later, we can make it optional and explicitly define a default value.

Part of #5961

```release-note
Make priorityClassSource in the SchedulePriority of PropagationPolicy a required field.
```


RainbowMango (approved):
/lgtm
/approve
Thanks.

PR #6181, opened 2025-03-03
Proposal to enhance FederatedResourceQuota to enforce resource limits directly on Karmada level

### LINKED ISSUE #5179: FederatedResourceQuota should be failover friendly

mszacillo:
**What type of PR is this?**
/kind design

This proposal enhances the FederatedResourceQuota so that it can impose namespaced resource limits directly on the Karmada control-plane level.

Fixes #5179 

Following our discussion during the community meeting, there were some questions regarding API server request latency as a result of adding an admission webhook to enforce resource limits. I ran some tests by measuring the time taken to apply some FlinkDeployments to the Karmada control-plane:

**Without webhook**: Average e2e request latency over 100 retries: 370 ms
**With webhook**: Average e2e request latency over 100 retries: 390 ms

The request latency increases slightly, as expected. But can run some more comprehensive performance tests for this feature. 

```release-note
NONE
```


RainbowMango (commented):
/assign
Start working on this.

docs/proposals/federatedresourcequota/federated-resource-quota-enhancement.md
```diff
@@ -0,0 +1,83 @@
...
+ - Enhance the FederatedResourceQuota to support enforcing namespaces resource limits on the Karmada control-plane level
+ - Make this feature configurable and toggleable
+
+### Non-Goals
+
+## Proposal
+
+1. FederatedResourceQuota should enforce Overall resource limits if Static Assignment is not defined.
+    - FederatedResourceQuota will be updated whenever resources are applied against the relevant namespace
+    - We can either updated FRQ by default, or consider including a scope for the quota
```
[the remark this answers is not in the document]
mszacillo: I should have been more clear here. The FederatedResourceQuota **status** will be updated whenever resources are applied against the relevant namespace.

That said, I think I can reframe this as a separate API which enforces resource limits on the Karmada control-plane, and keep the existing federated resource quota as is. That way we have separation between the APIs and can introduce the feature without stepping over existing use-cases.

docs/proposals/federatedresourcequota/federated-resource-quota-enhancement.md
```diff
@@ -0,0 +1,83 @@
...
+}
+```
+
+## Design Details
+
+### Controller Change
+
+**Reconciliation**: Controller will reconcile whenever a resource binding is created, updated, or deleted. Controller will only reconcile if the resource in question has a pointer to a FederatedResourceQuota.
+
+**Reconcile Logic**:  When reconciling, the controller will fetch the list of ResourceBindings by namespace and add up their resource requirements. The existing implementation grabs all RBs by namespace, however this could probably be improved by only calculating the delta of the applied resource, rather than calculating the entire resource footprint of the namespace.
```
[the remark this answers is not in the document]
mszacillo: Good point, this was a concern of mine as well. I'm thinking we could perhaps maintain a ResourceBinding cache kept by the controller. In the event of an update to a ResourceBinding, the controller would:

1. Check the cache to see if the ResourceBinding exists.
2. If yes, take the previous spec and calculate a resource delta to update the FRQ status
3. If no, update the FRQ status using the existing resource delta
4. Update the cache with teh new ResourceBinding spec. 

In the case that the controller pod is restarted or crashes, the cache will be lost and will need to be repopulated. But this would be an improvement over fetching the whole list of ResourceBindings each reconcile loop.

docs/proposals/federatedresourcequota/federated-resource-quota-enhancement.md
```diff
@@ -0,0 +1,83 @@
...
+### Scheduler Change
+
+Note: Since the controller is listening to RBs, the FederatedResourceQuota will be calculated after the resource binding has been created or updated.
+If a user bulk-applies a bunch of resources at once, it could be possible for the user to go above the quota’s limits. In this case, we should also check that the quota is honored before deciding to schedule the resource to a member cluster.
+
+### Admission Webhook
+
+As part of this change, we will do two things:
+
+1. Edit the existing FederatedResourceQuota validating webhook to prevent users from toggling StaticAssignments on/off when the resource quota is already in use. Since the overall limits and static assignments have different controllers, we don’t want them both reconciling the resource at once.
```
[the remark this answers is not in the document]
mszacillo: The issue is that both the existing static assignment controller and the new proposed controller will touch the same parts of the FederatedResourceQuota status. Since they calculate the resource usage differently, there would be conflicting changes in the status. 

The more I think about this, the more I'm tempted to separate out this resource enforcement into its own API.

docs/proposals/federatedresourcequota/federated-resource-quota-enhancement.md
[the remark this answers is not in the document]
RainbowMango: I will look at the separated API later, but it seems concerning because this proposal is exactly under the scope of FederatedResourceQuota. The new API still needs to clarify its relationship with FederatedResourceQuota, unless its intention is to replace FederatedResourceQuota.

docs/proposals/federatedresourcequota/federated-resource-quota-enhancement.md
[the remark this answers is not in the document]
RainbowMango: This an option I guess. For now, perhaps we don't have to dive into the implementation details. I belive there are many ways to handle that. I haven't looked at how Kubernetes calculates the used quota yet, maybe we can share some from it.

In addition, even without the cache you mentioned, we still can build an index with the field like `.spec.resourceQuotaName` against ResourceBinding, we can get all of the ResourceBindings enabled quota.(Don't have to list all of them).  By the way, there is an effort to manage all of the indexers across the system. If you want to know how the index works, maybe you can refer to #6204.

docs/proposals/resource-quota-enforcer/resource-quota-enforcer.md
```diff
@@ -0,0 +1,159 @@
...
+
+Eventually, all clusters are full and no more resources can be scheduled.
+
+![failover-example-1](resources/failover-example-1.png)
+
+However, let's now assume there was a cluster failure which triggers a cluster failover. In this case, since the ResourceQuotas are statically assigned, the fallback cluster will not be able to schedule any additional workloads. Jobs will be unable to failover, and will have to wait until the original cluster comes back up. This is not acceptable.
+
+![failover-example-2](resources/failover-example-2.png)
+
+Could we support dynamic assignment of FederatedResourceQuota? Potentially yes, but there are some drawbacks with that approach:
```
RainbowMango, line 46: Definitely yes! :) 
I wanted this about two years ago!

docs/proposals/resource-quota-enforcer/resource-quota-enforcer.md
```diff
@@ -0,0 +1,159 @@
...
+1. Each time an application failovers, the FederatedResourceQuota will need to check that the feasible clusters have enough quota, and if not, rebalance the resource quotas before scheduling work. This adds complexity to the scheduling step, and would increase E2E failover latency.
```
RainbowMango, line 47: I may depend on the way how to balance the quota dynamically.

How about just maintaining the overall quota and used data at the Karmada control plane? no ResourceQuota will be split into member cluster.

RainbowMango:
@whitewindmills Would you like to take a look at this proposal as well? It is highly likely that the scheduler should be involved to enforce the FederatedResourceQuota.

whitewindmills:
get, I'll take a look ASAP.
/assign

mszacillo:
Hi @RainbowMango, I've updated the proposal following the discussions we had during Kubecon earlier last week!

I've also included another user story for the Apache Spark Platform use-case, which was written by @CaesarTY. Hopefully we can begin working on this feature soon, before the next Karmada release.

docs/proposals/federated-resource-quota-enhancement/federated-resource-quota-enhancement.md
```diff
@@ -0,0 +1,164 @@
...
+ - Make this feature configurable and toggleable
+
+### Non-Goals
+ - Support of dynamic resource quota allocation is outside the scope of this proposal.
+
+## Proposal
+
+1. FederatedResourceQuota API should enforce namespaced overall resource limits.
+    - FederatedResourceQuota Status will be updated whenever resources are applied against the relevant namespace
+    - We consider including a resource selector scope for the quota
```
zhzhuang-zju, line 93: What are the approximate functions and APIs of the resource selector scope?

docs/proposals/federated-resource-quota-enhancement/federated-resource-quota-enhancement.md
```diff
@@ -0,0 +1,164 @@
...
+3. If the FederatedResourceQuota has a scope, determine if the resource matches the scope, and then add the pointer in the binding if needed.
+
+```go
+// ResourceBindingSpec represents the expectation of ResourceBinding.
+type ResourceBindingSpec struct {
+    ***
+
+    //  FederatedResourceQuota represents the name of the quota that will be used for this resource
+	// +optional
+	 FederatedResourceQuota string `json:"federatedResourceQuota,omitempty"`
```
zhzhuang-zju, line 128: Does this mean that only one FederatedResourceQuota can be applied within the same namespace?

docs/proposals/federated-resource-quota-enhancement/federated-resource-quota-enhancement.md
```diff
@@ -0,0 +1,164 @@
...
+1. If the reconciled ResourceBinding is not present in the cache.
+2. The reconciled ResourceBinding has a spec change and should be updated in the cache.
+
+Resource usage delta would
+
+### Scheduler Change
+
+Note: Since the controller is listening to RBs, the FederatedResourceQuota will be calculated after the resource binding has been created or updated.
+
+If a user bulk-applies a bunch of resources at once, it could be possible for the user to go above the quota’s limits. In this case, we should also check that the quota is honored before deciding to schedule the resource to a member cluster.
```
zhzhuang-zju, line 158: We can draw on the implementation of the native resourcequota to avoid the quota restrictions becoming ineffective due to user bulk applications. I have an idea here that we can discuss. Maybe during the next meeting?

zhzhuang-zju (commented):
Hey @mszacillo, Thank you for your proposal. It is very interesting and meaningful. I will keep an eye on this proposal and hope that I can be of some help with it.

docs/proposals/federated-resource-quota-enhancement/federated-resource-quota-enhancement.md
mszacillo: I can flesh this out and update the proposal. But the general idea was to support a PriorityClass scope in the same way the Kubernetes ResourceQuota does: https://kubernetes.io/docs/concepts/policy/resource-quotas/#quota-scopes

However, for a first pass of this feature this may not be necessary. If a FederatedResourceQuota is defined in the namespace, then it can monitor all resources in that namespace rather than filtering by scope. Let me know what you think.

docs/proposals/federated-resource-quota-enhancement/federated-resource-quota-enhancement.md
mszacillo: Sounds great!! Yes this is definitely a concern, which is why we were discussing a change to the scheduler may be necessary to additionally enforce quota limits there.

docs/proposals/federated-resource-quota-enhancement/federated-resource-quota-enhancement.md
```diff
@@ -0,0 +1,164 @@
...
+
+Instead, we would like to enhance the existing FederatedResourceQuota API so that it can enforce resource limits directly on the Karmada control-plane and allow users to configure their cluster federation to be ready for failovers.
+
+### User Story 2: Apache Spark on Karmada
+
+**User**: @CaesarTY
+
+Certain types of applications, especially big data and analytics jobs (e.g., Apache Spark on Kubernetes), do not always request all required resources up front. Instead, they may dynamically request additional resources (executors) at runtime, depending on current workload size, task parallelism, and job progress.
+
+In a multi-cluster environment managed by Karmada, we currently lack a way to enforce resource usage limits across clusters for such elastic workloads. Traditional Kubernetes ResourceQuota objects operate at the namespace level within a single cluster and are not designed to account for cross-cluster resource consumption or dynamic scaling behavior.
```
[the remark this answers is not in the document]
CaesarTY: Sure. A typical example is Spark application running in [dynamic allocation mode](https://spark.apache.org/docs/latest/configuration.html#dynamic-allocation), which adjusts the number of executor pods at runtime based on the job's needs. Here is a simple spark submit command:
```
spark-submit \
  --class org.apache.spark.examples.JavaWordCount \
  --master k8s://https://<api-server> \
  --deploy-mode cluster \
  --conf spark.dynamicAllocation.enabled=true \
  --conf spark.dynamicAllocation.minExecutors=1 \
  --conf spark.dynamicAllocation.initialExecutors=5 \
  --conf spark.dynamicAllocation.maxExecutors=50 \
  --conf spark.kubernetes.namespace=analytics \
  --conf spark.kubernetes.container.image=myrepo/spark:latest \
  local:///opt/spark/examples/jars/spark-examples_2.12-3.3.0.jar \
  s3a://my-bucket/text-data/
  ```
In this setup:
1. The job starts with 5 executors.
2. Executors are dynamically scaled up to 50 as needed.
3. Each executor is a pod that may land on any eligible node. 

It will be great if we can have a way to enforce resource usage limits across clusters for such elastic workloads.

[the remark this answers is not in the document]
Comment (mszacillo) [reply docs/proposals/federated-resource-quota-enhancement/federated-resource-quota-enhancement.md:73]: As long as the max # of executors is something that can be found in the CRD spec, you can probably configure your custom interpreter to account for this in your replica requirements, which in turn would be reflected in the FederatedResourceQuota.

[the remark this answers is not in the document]
Comment (zhzhuang-zju) [reply docs/proposals/federated-resource-quota-enhancement/federated-resource-quota-enhancement.md:73]: > The job starts with 5 executors.
Executors are dynamically scaled up to 50 as needed.

@CaesarTY Thanks to your detailed description. For this elastic scenario, do you want to calculate the usage based on 5, 50, or the number of replicas in the member cluster?

File: docs/proposals/federated-resource-quota-enhancement/federated-resource-quota-enhancement.md (hunk 6 above)
Comment (zhzhuang-zju) [reply docs/proposals/federated-resource-quota-enhancement/federated-resource-quota-enhancement.md:93]: > support a PriorityClass scope in the same way the Kubernetes ResourceQuota does

Get it. This is also one of the features I expect. 

> However, for a first pass of this feature this may not be necessary. If a FederatedResourceQuota is defined in the namespace, then it can monitor all resources in that namespace rather than filtering by scope. Let me know what you think.

In terms of API design, we can try to make it as complete as possible or ensure sufficient scalability. In terms of implementation, we can achieve it step by step according to a certain rhythm.

File: docs/proposals/federated-resource-quota-enhancement/federated-resource-quota-enhancement.md (hunk 7 above)
Comment (zhzhuang-zju) [reply docs/proposals/federated-resource-quota-enhancement/federated-resource-quota-enhancement.md:128]: > This just means that the ResourceBinding will be tied to at most 1 FederatedResourceQuota.

Could you explain the reason for this constraint? It seems that the native ResourceQuota doesn't limit a pod to using only one ResourceQuota.

File: docs/proposals/federated-resource-quota-enhancement/federated-resource-quota-enhancement.md [hunk 10]
```diff
@@ -0,0 +1,164 @@
...
+
+However, let's now assume there was a cluster failure which triggers a cluster failover. In this case, since the ResourceQuotas are statically assigned, the fallback cluster will not be able to schedule these applications because the ResourceQuota is already full. Jobs will be unable to be rescheduled, and will have to wait until the original cluster comes back up. This is not acceptable.
+
+![failover-example-1](resources/failover-example-1.png)
+
+In the following image, we show what we would ideally like to support. The Karmada control plane limits the global resource usage for the user's namespace, but does not statically assign ResourceQuotas. Both member clusters have identical ResourceQuotas of 40 CPU and 50 Gb Memory, so when Member Cluster 1 fails, Member Cluster 2 will have enough space to host FlinkDeployments A and B:
+
+![failover-example-2](resources/failover-example-2.png)
+
+Could we support dynamic assignment of FederatedResourceQuota? Potentially yes, but there are some drawbacks with that approach:
```
zhzhuang-zju, line 61: @mszacillo For the scenario of failure migration, the work in the failed cluster may not be cleared immediately. Do you want the resource usage of the failed cluster to be included in the total resource usage, or do you prefer to ignore this part?

docs/proposals/federated-resource-quota-enhancement/federated-resource-quota-enhancement.md
```diff
@@ -0,0 +1,164 @@
...
+
+Note: Since the controller is listening to RBs, the FederatedResourceQuota will be calculated after the resource binding has been created or updated.
+
+If a user bulk-applies a bunch of resources at once, it could be possible for the user to go above the quota’s limits. In this case, we should also check that the quota is honored before deciding to schedule the resource to a member cluster.
+
+### Admission Webhook
+
+As part of this change, we will introduce a new validating webhook:
+
+1. The new validating webhook will watch all types of resources, at least all native workloads (Deployments) and supported CRDs (FlinkDeployments). The webhook will use Karmada's `ResourceInterpreter#GetReplicas` to calculate the predicted delta resource usage for the quota. If the applied resource goes above the limit, then the webhook will deny the request.
```
zhzhuang-zju, line 164: If we implement admission control on the control plane, is it still necessary to propagate the resource quota to the member clusters?

docs/proposals/federated-resource-quota-enhancement/federated-resource-quota-enhancement.md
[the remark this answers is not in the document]
CaesarTY: Ideally, the usage should be calculated by the real-time number of replicas (driver + executors pods resources) and enforced by FederatedResourceQuota across the tier

docs/proposals/federated-resource-quota-enhancement/federated-resource-quota-enhancement.md
mszacillo: No, we won't need to propagate resource quotas.

docs/proposals/federated-resource-quota-enhancement/federated-resource-quota-enhancement.md
mszacillo: I think this can be ignored, because it should be expected for Karmada to clean up the resource. From the perspective of the user, they have applied their resource once to the Karmada control-plane, so it should be reflected in the resource quota accordingly.

docs/proposals/federated-resource-quota-enhancement/federated-resource-quota-enhancement.md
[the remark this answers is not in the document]
RainbowMango: Note that once the Spark application is deployed to a member cluster, there is no way for Karmada to suspend the executor scale-up. (Sure, Karmada can calculate the real-time resource consumption, but it can only be used to check the following Spark applications.)

I think @zhzhuang-zju's question is whether we should enforce quota interception based on the initial or maximum usage.

In my opinion, we should always use the maximum usage that can ensure the total amount of resource usage does not exceed the quota limits.

docs/proposals/federated-resource-quota-enhancement/federated-resource-quota-enhancement.md
RainbowMango: Echo from the proposal:
> Reconciliation: Controller will reconcile whenever a resource binding is created, updated, or deleted. Controller will only reconcile if the resource in question has a pointer to the FederatedResourceQuota.

Is this the reason of introducing the reference? For the controller to figure out which ResourceBinding should be cared about.

mszacillo: > Could you explain the reason for this constraint? It seems that the native ResourceQuota doesn't limit a pod to using only one ResourceQuota.

This isn't a hard constraint, but this decision was meant to decrease the scope of the proposal. I would also want to know more use-cases for this, since both of the user stories in the proposal only require one FederatedResourceQuota per ResourceBinding.

mszacillo: > The question is why ResourceBinding needs to keep the reference to the FederatedResourceQuota?

Your reasoning below was correct, this was to make the controller logic simpler. Since the controller is updating the FederatedResourceQuota status, it needs to know which ResourceBindings are tied to the specific FederatedResourceQuota - hence the reference.

zhzhuang-zju: https://github.com/kubernetes/kubernetes/blob/44c230bf5c321056e8bc89300b37c497f464f113/staging/src/k8s.io/apiserver/pkg/quota/v1/interfaces.go#L45-L65
Hi @mszacillo, I guess the role of `FederatedResourceQuota` is similar to the method `Evaluator.Matches` of the native resourceQuota. Both are used to determine whether the resource can be applied to this quota. It's just that one maintains a method, while the other maintains a field in the API.

zhzhuang-zju: In terms of extensibility, I prefer the latter. Moreover, maintaining an API field can also be quite burdensome. Of course, if I'm wrong in my understanding, please correct me.

docs/proposals/federated-resource-quota-enhancement/federated-resource-quota-enhancement.md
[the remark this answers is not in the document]
RainbowMango: If the resource name is `limits.cpu`, then we should use maximum usage, take `maxExecutors` into account.
```yaml
kind: FederatedResourceQuota
metadata:
  name: foo
spec:
  overall:
    limits.cpu: 100
```
If the resource name is `requests.cpu`, then should use the requested usage, take `initialExecutors` into account.
```yaml
kind: FederatedResourceQuota
metadata:
  name: foo
spec:
  overall:
    requests.cpu: 100
```

@CaesarTY I don't know if that meets your case. It would be great to share a sample of your ResourceQuota configuration.

docs/proposals/federated-resource-quota-enhancement/federated-resource-quota-enhancement.md
mszacillo: Huge apologies for not being able to make the meeting this week. Perhaps we can discuss your thoughts here? @zhzhuang-zju

docs/proposals/federated-resource-quota-enhancement/federated-resource-quota-enhancement.md
mszacillo: @zhzhuang-zju I'll take a look at the `Evaluator.Matches` implementation a bit more, as I'm not familiar with the method. But I agree that maintaining API fields can be burdensome. Let me catch up on how `Evaluator.Matches` works and I can update the proposal with this information.

docs/proposals/federated-resource-quota-enhancement/federated-resource-quota-enhancement.md
zhzhuang-zju: Of course. When dealing with bulk applications, the admission control of the native resource quota will aggregate the validation requests according to the namespace and process all the requests under one namespace serially at one time. 
fccf5cd49e01ca3cee6b66080b6aa13
For the convenience of understanding, I will use this diagram to explain. As you can see from this diagram, the verification request will first enter the namespace queue. At the same time, a corresponding map will be maintained, where the key is the namespace and the values are the verification requests that have already entered into the queue. Then the work routine will retrieve the namespace from the namespace queue and obtain all the verification requests under that namespace from the map. In this way, the verification requests under one namespace will always be processed within one work routine, and they are processed serially, which ensures that in the scenario of bulk applications, it will not exceed the quota limit.

zhzhuang-zju: Based on this, if we need to achieve the same effect, the webhook may not be able to meet the requirements, and we need to add a new controller.

zhzhuang-zju: However, karmada-webhook supports multiple instances, which may make it difficult to directly draw on.

zhzhuang-zju:
Hi @mszacillo, can you add me to the reviewers of your proposal？Thanks~

mszacillo:
> Hi @mszacillo, can you add me to the reviewers of your proposal？Thanks~

Yes, absolutely :)

docs/proposals/federated-resource-quota-enhancement/federated-resource-quota-enhancement.md
mszacillo: Got it! This makes a lot of sense, thank you for the thorough explanation. Thinking about this more, it think its clear that the webhook approach will not cover our edge cases (such as bulk applies for instance) - I'll update the proposal with the admission controller approach instead, borrowing from the ResourceQuota implemention in Kubernetes.

mszacillo:
Hello - apologies for delay, had Easter holidays last weekend and I was quite sick. I've gone ahead and updated the proposal to address latest feedback. @zhzhuang-zju I've also included your diagram in the proposal for the admission controller section.

zhzhuang-zju:
> Hello - apologies for delay, had Easter holidays last weekend and I was quite sick.

It’s all good! Hope you’re fully recovered now.


> I've also included your diagram in the proposal for the admission controller section.

Thanks

RainbowMango:
> Hello - apologies for delay, had Easter holidays last weekend and I was quite sick. I've gone ahead and updated the proposal to address latest feedback. @zhzhuang-zju I've also included your diagram in the proposal for the admission controller section.

Sorry to hear that, take care. Hope you doing fine now. No worries!
@zhzhuang-zju talked a lot with me about this proposal, he worked out several potential ways to implement this feature, and hope to talk with you at the next meeting.  Hopefully, we could make an agreement there.

mszacillo:
@RainbowMango @zhzhuang-zju Sounds great! See you both at the community meeting.

RainbowMango:
@mszacillo For this proposal, I think we already have a clear user case and definition of the feature. Could we merge this version first and then fill in the implementation details later by following PRs? This would allow us to focus on the updated parts incrementally, which might help move this feature faster.

RainbowMango:
/lgtm
/approve
/hold
Just add all necessary tags for merging this PR. 

If you agree to merge this first, please unhold this by leaving a comment `/hold cancel` in a separate line. After that, the PR will be merged, and then we can start a new PR for further updates.

mszacillo:
@RainbowMango Sure, that sounds good to me! Happy to add updates to this proposal as we implement.

RainbowMango:
Great thanks to @mszacillo. Could you please create a new issue for the umbrella tasks? So we can track the progress there.

In addition, for the MVP versions, I prefer to have both the webhook(for validation) and controller(for calculation), because they are mandatory parts for this feature. So, shall we organize all tasks that should be in MVP into part-1, and leave others to part-2?

PR #6192, opened 2025-03-09
use singleton rate limiter for dynamically created member cluster clients

### LINKED ISSUE #6094: http rate limit parameters not working properly
**What happened**:

Currently, there are 4 cmdline parameters for http rate limiting for all karmada components: `cluster-api-qps`, `cluster-api-burst`, `kube-api-qps`, and `kube-api-burst`. However, the limits of these parameters are not strict and sometimes do not work at all.

The request qps of the Karmada component should strictly follow the qps/burst parameter.

Using karmada-controller-manager as an example, it could also work on karmada-agent

1. a karmada cluster with lots of workloads and visiable monitoring, clusters generated by `hack/local-up-karmada/sh` could also work.
2. disable leader-elect, to avoid interference caused by failure of election.
3. set `concurrent-work-syncs`, `concurrent-resourcebinding-syncs`, `concurrent-resource-template-syncs` to 100 or larger, set `kube-api-qps`, `kube-api-burst` to 20, set `cluster-api-qps`, `cluster-api-burst` to 5, and start karmada-controller-manager.

Now we can check the monitoring through promql: `sum by (pod, host)(increase(rest_client_requests_total{namespace="karmada-system",pod=~"karmada-controller-manager.+"}[1m]))/60`, we can see that the actual qps accessing the karmada cluster and the member cluster far exceeds the limit we set.

Image

There are two main reasons.

1. QPS/burst parameters are not passed to the related clients.

This affects controller-manager and agent.
The `cluster-api-qps` and `cluster-api-burst` parameters are only passed to cluster-status-controller, while objectwatcher, which has more requirements for qps/burst, is not limited by these two parameters. The actual limit of its qps are `concurrent-work-syncs` and `kube-api-qps`/`kube-api-burst`, which limit the speed of updating work objects.

2. When creating multiple clients in a component, even if the same `rest.Config` is used, their ratelimits are still independent, not shared.

This affects most of the karmada components. I won't list them one by one.

https://github.com/karmada-io/karmada/blob/ae2d1c10b7a2d836388ec3f81156fb5182b5cb8c/cmd/scheduler/app/scheduler.go#L149-L153

In the code above, there are 3 clients with the same qps/limit, and the total qps/burst is 3 times the parameters we set. 

The actual impact of this issue is more severe in the controller manager because it needs to update a large number of karmada objects and workloads, which also use different clients. Not only clients accessing the karmada cluster are affected, but also clients accessing the member clusters because clients are generated dynamically. In the objectwatcher, it even generates a client for each update operation.

**Environment**:
- Karmada version: current master
- Others:

zach593:
**What type of PR is this?**

/kind bug

see #6094, this PR add the a ratelimiterGetter singleton for dynamically created cluster clients, to let the rate limiter to more accurately limit the total qps of multiple dynamically created clients .

Fixes #6094

I found it difficult to test using mock client, because the existing fake client implementations do not accept ratelimiter parameters. However, I added some tests in `pkg/util/ratelimiter_test.go` to simulate and illustrate the final effect.

If we want to test the rate limit more completely, we may need to introduce envTest (a sub package of controller-runtime), but this requires more general considerations, because the impact is large, so it will not be expanded in this PR.

```release-note
karmada-controller-manager, karmada-agent, karmada-metrics-adapter: rate limit parameters of  'cluster-api-qps' and 'cluster-api-burst', these parameters will be stricter and administrators may need to adjust them to get the same effect as before the update.
```


pkg/controllers/status/cluster_status_controller.go
```diff
@@ -343,7 +343,7 @@ func (c *ClusterStatusController) initializeGenericInformerManagerForCluster(clu
 		return
 	}
 
-	dynamicClient, err := c.ClusterDynamicClientSetFunc(clusterClient.ClusterName, c.Client)
+	dynamicClient, err := c.ClusterDynamicClientSetFunc(clusterClient.ClusterName, c.Client, c.ClusterClientOption)
```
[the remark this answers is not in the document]
zach593: The code compiles, so that's not a problem.

pkg/controllers/status/cluster_status_controller_test.go
```diff
@@ -113,11 +113,8 @@ func TestClusterStatusController_Reconcile(t *testing.T) {
 				Client:                 fake.NewClientBuilder().WithScheme(gclient.NewSchema()).Build(),
 				GenericInformerManager: genericmanager.GetInstance(),
 				TypedInformerManager:   typedmanager.GetInstance(),
-				ClusterClientOption: &util.ClientOption{
-					QPS:   5,
-					Burst: 10,
-				},
-				ClusterClientSetFunc: util.NewClusterClientSet,
+				ClusterClientOption:    &util.ClientOption{},
```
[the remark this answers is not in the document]
zach593: The code implementation allows option to be empty, and the test code does not require very fine control of rate limit, so I don't think this is a problem.

zach593:
/remove-kind feature
/kind bug

I think the issue with the rate limiter not working is more like a bug, please feel free to correct me if anyone thinks it should not be.

pkg/util/membercluster_client.go
```diff
@@ -132,37 +133,51 @@ func NewClusterClientSetForAgent(clusterName string, _ client.Client, clientOpti
...
+				clusterConfig.RateLimiter = clientOption.RateLimiterGetter(clusterName)
+			}
 		}
 		clusterClientSet.KubeClient = kubeclientset.NewForConfigOrDie(clusterConfig)
 	}
 	return &clusterClientSet, nil
 }
 
+// NewClusterDynamicClientSetFunc is a function that returns a dynamic client for the given member cluster.
+type NewClusterDynamicClientSetFunc = func(clusterName string, client client.Client, clientOption *ClientOption) (*DynamicClusterClient, error)
```
XiShanYongYe-Chang, line 146: What is the difference between removing the equal sign and defining the function variable directly and the current definition method?
`type NewClusterDynamicClientSetFunc func(clusterName string, client client.Client, clientOption *ClientOption) (*DynamicClusterClient, error)`

XiShanYongYe-Chang (commented):
Genericly lgtm, ask @RainbowMango to help a look
/cc @RainbowMango

RainbowMango:
OK. Trying my best... I'm not sure if I will have enough time for this soon.

RainbowMango (commented):
/assign

pkg/util/ratelimiter.go
```diff
@@ -0,0 +1,70 @@
...
+	return defaultRateLimiterGetter
+}
+
+// RateLimiterGetter is a struct to get rate limiter.
+type RateLimiterGetter struct {
+	limiters map[string]flowcontrol.RateLimiter
+	mu       sync.Mutex
+	qps      float32
+	burst    int
+}
```
RainbowMango, line 38: This structure looks weird, as `RateLimiterGetter` actually manages and holds multiple rate-limiter instances; it should not hold the `qps` and `burst`. Otherwise lead to unexpected behavior, think the following flow:
1. SetLimits(10, 20)
2. SetLimits(15, 30)
3. GetRateLimiter("member-cluster")  // the rate-limiter will be initilized with qps=15, burst=30

zach593: Perhaps changing the function and field names would avoid this misuse.

```golang
// RateLimiterGetter is a struct to get rate limiter.
type RateLimiterGetter struct {
	limiters             map[string]flowcontrol.RateLimiter
	mu                    sync.Mutex
	defaultQPS      float32
	defaultBurst    int
}

// SetDefaultLimits sets the QPS and Burst used by the getter to create the ratelimiter.
// If qps and limit are not set when using Getter, it will return 5 and 10 (the default qps and burst of client-go dynamicClient).
// After setting qps and limit, the ratelimiter created by the getter will use the new qps and burst, but the existing ratelimiter will not be affected. 
// So it is best to call this function before the first use.
func (r *RateLimiterGetter) SetDefaultLimits(defaultQPS float32, defaultBurst int) *RateLimiterGetter {
	r.mu.Lock()
	defer r.mu.Unlock()
	r.defaultQPS = defaultQPS
	r.defaultBurst = defaultBurst
	return r
}
```

zach593: > Maybe not, as the `RateLimiterGetter` holds the rate limiter for both karmada-apiserver and member clusters. They might have different default QPS and burst.

Is this a hypothetical scenario? Currently, the getter doesn't hold the rate limiter for karmada-apiserver, that part is implemented in #6095. 

OTOH, if the getter implemented `AddRateLimiter(target string, qps float32, burst int)`, we will need a wrapper func as the `RateLimiterGetter` in the ClientOption, maybe looks like:

```go
	rateLimiterFunc := func(target string) flowcontrol.RateLimiter {
		getter := util.GetRateLimiterGetter()
		rateLimiter := getter.GetRateLimiter(target)
		if rateLimiter != nil {
			return rateLimiter
		}
		getter.AddRateLimiter(target, qps, burst)
		return getter.GetRateLimiter(target)
	}
	clusterClientOption := &util.ClientOption{RateLimiterGetter: rateLimiterFunc}
```

It's just like moving the "dirty" logic to another place, hiding the qps and burst into the closure function, I'm not sure this is what you want.

Another way is still passing the qps and burst with the ClientOption, but it will be more weird when getting the limiter:

```go
	if clusterConfig != nil {
		if clientOption != nil {
			qps := clientOption.QPS
			burst := clientOption.Burst
			getter := GetRateLimiterGetter()
			// what if the existing limiter's qps and burst are different from values in the option??
			clusterConfig.RateLimiter = getter.GetRateLimiter(clusterName)
			if clusterConfig.RateLimiter == nil {
				getter.AddRateLimiter(clusterName, qps, burst)
				clusterConfig.RateLimiter = getter.GetRateLimiter(clusterName)
			}
		}
		clusterClientSet.DynamicClientSet = dynamic.NewForConfigOrDie(clusterConfig)
	}
```

zach593:
I just updated, please let me know if there's anything else you would like me to adjust or improve. @RainbowMango

RainbowMango:
OK. Will do, thanks.

pkg/util/membercluster_ratelimiter.go
```diff
@@ -0,0 +1,76 @@
...
+type ClusterRateLimiterGetter struct {
+	limiters map[string]flowcontrol.RateLimiter
+	mu       sync.Mutex
+	qps      float32
+	burst    int
+}
+
+// SetDefaultLimits sets the default qps and burst for new rate limiters.
+// Do not call this method after calling GetRateLimiter, otherwise the call will not take effect.
+// If qps and limit are not set when using Getter, it will use 5 and 10 by default (the default qps and burst of client-go rest.Config).
```
RainbowMango, line 44:
```suggestion
// If qps and limit are not set when using Getter, it will use 40 and 60 by default (the default qps and burst of '--cluster-api-qps' and '--cluster-api-burst' in karmada-controller-manager).
```
Shall we keep the default consistent with the flags?

pkg/util/membercluster_ratelimiter.go
```diff
@@ -0,0 +1,76 @@
...
+	limiter, ok := r.limiters[key]
+	if !ok {
+		qps := r.qps
+		if qps <= 0 {
+			qps = 5
+		}
+		burst := r.burst
+		if burst <= 0 {
+			burst = 10
+		}
```
RainbowMango, line 71:
```suggestion
		qps := r.qps
		if qps <= 0 {
			qps = 40
		}
		burst := r.burst
		if burst <= 0 {
			burst = 60
		}
```
ditto.

pkg/search/controller.go
```diff
@@ -352,7 +352,7 @@ func (c *Controller) getRegistryBackendHandler(cluster string, matchedRegistries
 }
 
 var clusterDynamicClientBuilder = func(cluster string, controlPlaneClient client.Client) (*util.DynamicClusterClient, error) {
-	return util.NewClusterDynamicClientSet(cluster, controlPlaneClient)
+	return util.NewClusterDynamicClientSet(cluster, controlPlaneClient, nil)
```
RainbowMango, line 355: Why pass nil here?

RainbowMango (commented):
Generally looks good to me, except the default values and a question.

pkg/search/controller.go
zach593: karmada-search didn't provide `--cluster-api-qps' and '--cluster-api-burst` flags, currently its qps and burst are always 5 and 10, i.e. default value from `rest.Config`.
maybe later we will consider add these flags to it, but in this PR I think we could keep it untouched.

RainbowMango: Makes sense. But doesn't harm to add a comment to explain why pass nil here :)

RainbowMango (approved):
/lgtm
/approve

Issue #5971, opened 2024-12-24
dependencies distributor controller performance optimization


CharlesQQ:
**What would you like to be added**:
The following code:
https://github.com/karmada-io/karmada/blob/c9ca6aca6aa701879a01be5304ca226ce39519d8/pkg/dependenciesdistributor/dependencies_distributor.go#L173-L197

For example, in our production environment, there are 3372 dependent resources and 13572 resourcebinding resources; each dependent resource will be for looped 13572 times to match the resourcebinding object;
According to the log, a `matchesWithBindingDependencies` action takes approximately 0.000612s; calculated according to --concurrent-dependent-resource-syncs=30, it will take time to complete all executions
0.000612 * 3372 * 13572 /30 /60 =15 minutes

**Why is this needed**:

As part of performance optimization

RainbowMango:
/assign @CharlesQQ 
In favor of #6194

PR #6194, opened 2025-03-09
Dependency index

CharlesQQ:
**What type of PR is this?**
/kind feature

Fixes https://github.com/karmada-io/karmada/issues/5971

```release-note

```


CharlesQQ:
test result:
deployment beta-moa-demo-prod-976 and beta-moa-demo-prod-977 have dependency resource configmap-beta-moa-demo-prod-977; During configmap-beta-moa-demo-prod-977 updated, depencency index find all rb resources
I add new log
```
	if len(bindingList.Items) > 0 {
		klog.V(4).Infof("resource(%s/%s) matched %d  bindings", resourceTemplateKey.Namespace, resourceTemplateKey.Name, len(bindingList.Items))
	}
```
image

CharlesQQ:
The reason for the failure is that the dependent resource only has endpointslice labelSelector but no name, which causes the index matching to fail.

image

RainbowMango (commented):
/assign

pkg/util/helper/index.go
```diff
@@ -54,3 +80,39 @@ func IndexerFuncBasedOnLabel(key string) client.IndexerFunc {
...
+	err := mgr.GetFieldIndexer().IndexField(ctx, &workv1alpha2.ResourceBinding{}, IndexNameForResourceBindingDependency, IndexFuncResourceBindingDependencies())
+	if err != nil {
+		klog.Errorf("failed to create index for resource.dependencies, err: %v", err)
+		return err
+	}
+	return nil
+}
+
+// IndexFuncResourceBindingDependencies returns an IndexerFunc used to index resource with the given dependency resource.
+func IndexFuncResourceBindingDependencies() client.IndexerFunc {
```
XiShanYongYe-Chang, line 95: It seems that we don't need this index yet. Should the current pr focus on registering the index in the controller only once?

pkg/util/helper/index.go
```diff
@@ -18,15 +18,41 @@ package helper
...
 	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
 )
 
+const (
+	// IndexNameForResourceBindingDependency index name for resourcebinding dependency used for fieldSelector
+	IndexNameForResourceBindingDependency = "resourcebinding.dependency"
+)
+
+var (
+	registerOnce sync.Once
```
zach593, line 39: you don't need this `sync.Once`, add same index twice will return an error.

https://github.com/karmada-io/karmada/blob/6dfe381395c78b935aceaf73a0a873f92b66543c/vendor/k8s.io/client-go/tools/cache/thread_safe_store.go#L123-L135

pkg/util/helper/index.go
CharlesQQ: > Should the current pr focus on registering the index in the controller only once?

Maybe can use another pr, focus on registering the index in the controller only once

pkg/util/helper/index.go
CharlesQQ: sync.Once is to avoid registering the same index multiple times, we cannot rely on errors

zach593: But you're only calling it once now, you can add `sync.Once` later when you really start calling it multiple times in the subsequent PR. At least in this PR it's useless. For example, code like `setupControllers()`, `controllerManager.Start()` also only called once, but no one will consider add `sync.Once` for them.

CharlesQQ:
Indexer management call refer to this pr: https://github.com/karmada-io/karmada/pull/6204

pkg/util/helper/index.go
CharlesQQ: refer to this pr: https://github.com/karmada-io/karmada/pull/6204;

CharlesQQ: this pr 6194 in order to add a new index, not for management index

RainbowMango:
@CharlesQQ do we still need this?

RainbowMango:
Kindly ping @CharlesQQ

CharlesQQ:
>  do we still need this?


I cannot use fields that depend on resources to create index keys because of labels. I'll close this request for now.

PR #6199, opened 2025-03-11
fix(dependenciesdistributor): reduce update trigger

CharlesQQ:
**What type of PR is this?**
/kind feature

Each time a dependency resource is updated, dependenciesdistributor reconcile will be triggered and executed twice, if  `labels` field not updated, the second time is unnecessary.

When dependent resources are updated, reconcile is triggered twice. If labels are not updated, the second reconcile is meaningless.

The queue provided by go-client will automatically deduplicate if the same resource object key exists, provided that the resource object key is comparable, that is, implements the Comparable interface. The following types can be compared:

image

For the type `LabelsKey`, because there is an element labels, which is a map type, and  `LabelsKey` is pointer type, it becomes an incomparable type.

https://github.com/karmada-io/karmada/blob/65e7f74c991631a5888f5176342af118f4d5712e/pkg/dependenciesdistributor/dependencies_distributor.go#L603-L624

https://github.com/karmada-io/karmada/blob/56b72f1a80634d18b12bbfa87131a7a367b7e2ff/pkg/dependenciesdistributor/dependencies_distributor.go#L87-L90

https://github.com/karmada-io/karmada/blob/56b72f1a80634d18b12bbfa87131a7a367b7e2ff/vendor/k8s.io/client-go/util/workqueue/rate_limiting_queue.go#L93-L110

https://github.com/karmada-io/karmada/blob/56b72f1a80634d18b12bbfa87131a7a367b7e2ff/vendor/k8s.io/client-go/util/workqueue/queue.go#L240-L263

https://github.com/karmada-io/karmada/blob/56b72f1a80634d18b12bbfa87131a7a367b7e2ff/vendor/k8s.io/client-go/util/workqueue/queue.go#L222-L224

```release-note

```

### DIFF
--- a/pkg/dependenciesdistributor/dependencies_distributor.go
+++ b/pkg/dependenciesdistributor/dependencies_distributor.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -151,7 +152,9 @@ func (d *DependenciesDistributor) OnUpdate(oldObj, newObj interface{}) {
 		klog.V(4).Infof("Ignore update event of object (%s, kind=%s, %s) as specification no change", unstructuredOldObj.GetAPIVersion(), unstructuredOldObj.GetKind(), names.NamespacedKey(unstructuredOldObj.GetNamespace(), unstructuredOldObj.GetName()))
 		return
 	}
-	d.OnAdd(oldObj)
+	if !equality.Semantic.DeepEqual(unstructuredOldObj.GetLabels(), unstructuredNewObj.GetLabels()) {
+		d.OnAdd(oldObj)
+	}
 	d.OnAdd(newObj)
 }
 
@@ -214,7 +217,6 @@ func matchesWithBindingDependencies(resourceTemplateKey *LabelsKey, independentB
 			independentBinding.Namespace, independentBinding.Name, dependencies, err)
 		return false
 	}
-
 	if len(dependenciesSlice) == 0 {
 		return false
 	}
--- a/pkg/dependenciesdistributor/dependencies_distributor_test.go
+++ b/pkg/dependenciesdistributor/dependencies_distributor_test.go
@@ -114,7 +114,7 @@ func Test_OnUpdate(t *testing.T) {
 					},
 				},
 			},
-			wantQueueSize: 2,
+			wantQueueSize: 1,
 		},
 		{
 			name: "do not update the object, no specification changed",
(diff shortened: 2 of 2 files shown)


pkg/dependenciesdistributor/dependencies_distributor.go
```diff
@@ -151,7 +151,6 @@ func (d *DependenciesDistributor) OnUpdate(oldObj, newObj interface{}) {
 		klog.V(4).Infof("Ignore update event of object (%s, kind=%s, %s) as specification no change", unstructuredOldObj.GetAPIVersion(), unstructuredOldObj.GetKind(), names.NamespacedKey(unstructuredOldObj.GetNamespace(), unstructuredOldObj.GetName()))
 		return
 	}
-	d.OnAdd(oldObj)
```
XiShanYongYe-Chang, line 154: Currently, we support the use of labelselector to follow resource selection. When a resource is updated, the label of the resource may change. Therefore, the resource may meet the requirements before the update but not after the update. If we delete this line, we might miss this sceneary.

CharlesQQ: So under the cleanup of labelsSelector unchanged, it can be deleted

XiShanYongYe-Chang: Yes, it is.

pkg/dependenciesdistributor/dependencies_distributor_test.go
```diff
@@ -3308,3 +3342,206 @@ func Test_deleteBindingFromSnapshot(t *testing.T) {
 		})
 	}
 }
+
+// LabelsKeyString is the object key which is a unique identifier under a cluster, across all resources.
+// labels field is the serialized labels string; the reason for this is that map is an incomparable type in golang
+// thus, the queue's add method cannot deduplicate
+type LabelsKeyString struct {
```
XiShanYongYe-Chang, line 3349: I think this is the test code.

pkg/dependenciesdistributor/dependencies_distributor.go
```diff
@@ -151,7 +152,9 @@ func (d *DependenciesDistributor) OnUpdate(oldObj, newObj interface{}) {
...
+	if !equality.Semantic.DeepEqual(unstructuredOldObj.GetLabels(), unstructuredNewObj.GetLabels()) {
```
XiShanYongYe-Chang, line 155: Can `DeepEqual` be used to determine whether two maps are the same? If so, can you help add the corresponding ut?

pkg/dependenciesdistributor/dependencies_distributor_test.go
CharlesQQ: yes

pkg/dependenciesdistributor/dependencies_distributor.go
CharlesQQ: Add  test case in  Test_OnUpdate

pkg/dependenciesdistributor/dependencies_distributor_test.go
XiShanYongYe-Chang: I'm sorry, I didn't make it clear. I mean, we don't need this test.

pkg/dependenciesdistributor/dependencies_distributor.go
XiShanYongYe-Chang: What if there are more than two elements in the map?

CharlesQQ: > What if there are more than two elements in the map?

add new test case

XiShanYongYe-Chang (commented):
Thanks a lot~
/lgtm
/approve

PR #6204, opened 2025-03-14
feat(helper): A way to manage indexer

CharlesQQ:
**What type of PR is this?**

A way to centrally manage indexer
Provide a registration function for each indexer so that it can be registered wherever needed

Fixes #

```release-note

```

### DIFF
--- a/pkg/util/helper/index.go
+++ b/pkg/util/helper/index.go
@@ -28,21 +28,32 @@ import (
 	"github.com/karmada-io/karmada/pkg/util/indexregistry"
 )
 
-// IndexWork creates index for Work.
-func IndexWork(ctx context.Context, mgr ctrl.Manager) error {
-	err := mgr.GetFieldIndexer().IndexField(ctx, &workv1alpha1.Work{}, indexregistry.WorkIndexByResourceBindingID,
-		IndexerFuncBasedOnLabel(workv1alpha2.ResourceBindingPermanentIDLabel))
-	if err != nil {
-		klog.Errorf("failed to create index for work, err: %v", err)
-		return err
-	}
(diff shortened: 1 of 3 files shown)


RainbowMango (commented):
/assign

pkg/util/helper/index.go
```diff
@@ -18,15 +18,53 @@ package helper
...
+)
+
+// Options Optional when defining the registration indexer
+type Options struct {
+	// EnableTaintManager If set to true will register resourcebinding.spec.clusters and clusterresourcebinding.spec.clusters indexer
+	EnableTaintManager bool
+}
+
+// RegisterFieldIndexes register all field indexes
+func RegisterFieldIndexes(ctx context.Context, mgr ctrl.Manager, opt *Options) error {
```
RainbowMango, line 50: I don't think it's a good idea to couple controller options.
I think each index function should own one sync.Once, so that it can be involved everywhere need it.

CharlesQQ: This is an option. I refer to openkruise, which chooses to register all indexers into one function.

https://github.com/openkruise/kruise/blob/14d9ebdbfc248a86b459702fae1a228987a6674a/pkg/util/fieldindex/register.go#L43-L99

RainbowMango: Thanks for the information. 
What if the other component other than `karmada-controller-manager` also needs to register indexers? Note that not all components share the same set of indexers.

CharlesQQ: > Note that not all components share the same set of indexers.

May be a  good suggestion is to create an index registration function for each component that needs to create indexers.

RainbowMango:
I was hoping this could be part of #6217, but it has been included in #6372

PR #6227, opened 2025-03-24
fix the issue where the binding gracefulEvictionTask can never be clear

### LINKED ISSUE #3747: graceful eviction tasks in resource binding never cleaned
**What happened**:

the `gracefulEvictionTasks` field in spec of resource binding, grows, never cleaned.

if the task already finished, task should removed from `gracefulEvictionTasks`

patch `gracefulEvictionTasks` add item to trigger eviction, ensure the task finished, check the graceful eviction tasks field, the task still remain.

it may because of line 73, the `Patch` acts a merge behaviour, which won't remove tasks not kept :

https://github.com/karmada-io/karmada/blob/b01cf50caee8c895c808c2ba7d7dbb75eff2a5b8/pkg/controllers/gracefuleviction/rb_graceful_eviction_controller.go#L64-L76

**Environment**:
- Karmada version:
- Others:

XiShanYongYe-Chang:
**What type of PR is this?**

/kind bug

When the binding object is scheduled but there is no result, the SchedulerObservedGeneration in the binding object status will not be set to the generation value, which results in graceful-eviction-controller will not be able to handle gracefulEvictionTask and will always remain.

Fixes #3747

```release-note
`karmada-controller-manager`: fix the issue where the binding gracefulEvictionTask can never be clear.
```

### DIFF
--- a/pkg/controllers/gracefuleviction/crb_graceful_eviction_controller.go
+++ b/pkg/controllers/gracefuleviction/crb_graceful_eviction_controller.go
@@ -78,7 +78,12 @@ func (c *CRBGracefulEvictionController) Reconcile(ctx context.Context, req contr
 }
 
 func (c *CRBGracefulEvictionController) syncBinding(ctx context.Context, binding *workv1alpha2.ClusterResourceBinding) (time.Duration, error) {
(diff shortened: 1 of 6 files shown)


RainbowMango (commented):
/assign

RainbowMango (approved):
/lgtm
/approve

RainbowMango:
Please help to cherry-pick this to release branches.

Issue #6091, opened 2025-01-27
[lfx-mentorship-2025-March-May] Karmada Self-Signed Certificate Content Standardization


chaosi-zju:
**CNCF LFX mentorship:**  https://github.com/cncf/mentoring/tree/main/programs/lfx-mentorship/2025/01-Mar-May

**Mentor:** @chaosi-zju @XiShanYongYe-Chang 

**Description:**

In the existing Karmada architecture, each component should have its own unique certificates to ensure clear identity and security. Best practices dictate that each component's name be used as the Common Name (`CN`) in its certificate to facilitate identity differentiation. However, currently, all Karmada components share same identical certificate content, leading to confusion and potential security risks.

The objective of this project is to enhance the compliance of the Karmada certificate system by ensuring that each component possesses distinct certificates that reflect its identity. This will improve system security, reduce management complexity, and align with industry standards. This project aims to achieve the following standards:

* Utilize a single CA certificate for the entire Karmada system.
* Issue individual server certificates for each server component, using the component name as the `CN`.
* Issue individual client certificates for each client component, using the component name as the `CN`, same client can use consistent certificate for different servers.

**Requirements:**

* Familiarity with Golang, Kubernetes, and Karmada.
* Basic understanding of certificate management.

**Expected outcomes:**

- [ ] Complete the issuance of different certificates for 8 server components and import the certificate content into the corresponding certificate Secrets.
- [ ] Complete the issuance of different certificates for 11 client components and import the certificate content into the corresponding certificate Secrets or Config Secrets.

DeshDeepakKant:
@chaosi-zju 
I am very interested in working on this issue as it aligns well with my skills in Golang, Kubernetes, and certificate management. Ensuring proper certificate standardization for Karmada components is a crucial security improvement, and I would love to contribute by implementing distinct certificates for each component, improving security and compliance. I am eager to collaborate with the team, learn from the process, and help make Karmada’s certificate system more robust. Please let me know how I can get started and where my contributions would be most valuable!.

SkySingh04:
@chaosi-zju Looking forward to working on this for this term of LFX!

XiShanYongYe-Chang:
lfx project site: https://mentorship.lfx.linuxfoundation.org/project/8d2d522f-8838-4baa-9be4-d13dab30289b

tiansuo114:
[...]

PR #6246, opened 2025-03-31
design:Add component certificate identification for client components

tiansuo114:
**What type of PR is this?**
/kind design

This pr is about [issue#6091](https://github.com/karmada-io/karmada/issues/6091)

Previously, all Karmada components shared a set of similar certificates. In this case, we cannot identify the request source by the certificate used in the request, which may lead to some ambiguities. This PR aims to propose a design for testing purposes so that once we confirm there are no issues with the implementation, we can synchronize it to the deployment methods used in production environments, such as in `karmadactl` and `karmada-operator`.

Fixes [#6091](https://github.com/karmada-io/karmada/issues/6091)

@XiShanYongYe-Chang 
@zhzhuang-zju 

```release-note

```

### DIFF
--- a/artifacts/deploy/karmada-aggregated-apiserver.yaml
+++ b/artifacts/deploy/karmada-aggregated-apiserver.yaml
@@ -73,13 +73,30 @@ spec:
       volumes:
         - name: karmada-config
           secret:
-            secretName: karmada-aggregated-apiserver-config
+            secretName: karmada-aggregated-apiserver-secret
+            items:
+              - key: karmada.config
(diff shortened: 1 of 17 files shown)


tiansuo114:
Due to the community discussion that some of the current modification proposal is not reasonable, the content of this pr is temporarily abandoned, and the part of script changes to the [issue](https://github.com/karmada-io/karmada/issues/6264)are merged into the [pr](https://github.com/karmada-io/karmada/pull/6246)

PR #6249, opened 2025-03-31
don't add cluster name to hostname field when aggregate service status

### LINKED ISSUE #6248: The default resource interpreter should not fill clusterName to Hostname field when aggregating status
**What happened**:

When we were using karmada to distribute `Service` resources recently, we found that when the lb entry of a loadBalancer type Service is of ip type, karmada's default resourceinterpreter would write the name of the member cluster in the `Hostname` field. The original intention was obviously to make it easier to identify when the member cluster returns two different lb ips, but the `Hostname` field clearly indicates that it should be a legal domain name that can be queried by dns. So we think that filling in the member cluster name is inappropriate.

In actual situations, it does have an impact. Since our karmada will connect to multiple provider of multiple AZs, some provider can provide domain names as lb entries , while others provide ip. When these two services are in the same list, it causes interference in UI display and program identification.

When aggregate status for Service and Ingress, default resourceInterpreter should not fill `Hostname` field with cluster name.

- run a local up karmada
- propagate a service or ingress resource to member clusters
- use `kubectl edit --subresource=status` to update the resource status in the member cluster, manually fill in the ingress ip
- print the service/ingress in karmada cluster.

At present, we have found 3 places with this behavior: the default interpreter of service, ingress and pdb. Since there may be pods with the same name in multiple clusters, I guess we can only temporarily not modify pdb. So I proposed a pr #6249 to remove the behavior of adding clusterName for service and ingress. If users still want to identify the lb domain name and ip for each member cluster, I think they should read the `ResourceBinding`.

But removing clusterName will expose another problem. Most cloud providers can provide the same lb entry for services with the same name in multiple clusters. It probably doesn't make sense to repeat the same information multiple times in the status, so I wrote another PR #6252 to deduplicate it, but the code implementation is very verbose, so I would very much welcome reviews that can help optimize the implementation.

**Environment**:
- Karmada version:
- Others:

zach593:
**What type of PR is this?**

/kind bug

see #6248

Fixes #6248

```release-note
karmada-controller-manager: The default resource interpreter will no longer populate the `.status.LoadBalancer.Ingress[].Hostname` field with the member cluster name for Service and Ingress resources.
```

### DIFF
--- a/pkg/resourceinterpreter/default/native/aggregatestatus.go
+++ b/pkg/resourceinterpreter/default/native/aggregatestatus.go
@@ -145,13 +145,6 @@ func aggregateServiceStatus(object *unstructured.Unstructured, aggregatedStatusI
 		klog.V(3).Infof("Grab service(%s/%s) status from cluster(%s), loadBalancer status: %v",
 			service.Namespace, service.Name, item.ClusterName, temp.LoadBalancer)
 
-		// Set cluster name as Hostname by default to indicate the status is collected from which member cluster.
-		for i := range temp.LoadBalancer.Ingress {
-			if temp.LoadBalancer.Ingress[i].Hostname == "" {
-				temp.LoadBalancer.Ingress[i].Hostname = item.ClusterName
-			}
-		}
-
 		newStatus.LoadBalancer.Ingress = append(newStatus.LoadBalancer.Ingress, temp.LoadBalancer.Ingress...)
 	}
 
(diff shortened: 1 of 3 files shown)


XiShanYongYe-Chang:
Hi @zach593, Do you think we need to cherry-pick this patch to the previous release branch?

XiShanYongYe-Chang (commented):
Thanks~
/lgtm
/approve

zach593:
> Hi @zach593, Do you think we need to cherry-pick this patch to the previous release branch?
 
This is also a problem that has existed for a very long time. I think cherry-pick is not necessary.

XiShanYongYe-Chang:
> This is also a problem that has existed for a very long time. I think cherry-pick is not necessary.

Okay, thanks.

PR #6252, opened 2025-04-01
Deduplicate and sort status for Service and Ingress when aggregating status

zach593:
**What type of PR is this?**

/kind feature

Like described in https://github.com/karmada-io/karmada/issues/6248, If we no longer add the cluster name when aggregating the status of Services and Ingress, there will be redundant items with the same content in multiple ingresses in the status. like:

```
status:
  loadBalancer:
    ingress:
    - hostname: xxx.xxx.com
    - hostname: xxx.xxx.com
```

This pr adds the cluster name as an identifier when the member cluster status is not nil, and deduplicates and sorts the identical items in the list.

Fixes #

This PR includes a commit included in #6249. It is better to rebase and review this PR after merging #6249.

```release-note
karmada-controller-manager: The default resource interpreter will deduplicate and sort `status.Loadbalancer.Ingress` field for Service and Ingress resource when aggregating status.
```

### DIFF
--- a/pkg/resourceinterpreter/default/native/aggregatestatus.go
+++ b/pkg/resourceinterpreter/default/native/aggregatestatus.go
@@ -147,6 +147,7 @@ func aggregateServiceStatus(object *unstructured.Unstructured, aggregatedStatusI
 
 		newStatus.LoadBalancer.Ingress = append(newStatus.LoadBalancer.Ingress, temp.LoadBalancer.Ingress...)
 	}
+	newStatus.LoadBalancer.Ingress = helper.DedupeAndSortServiceLoadBalancerIngress(newStatus.LoadBalancer.Ingress)
 
 	if reflect.DeepEqual(service.Status, *newStatus) {
 		klog.V(3).Infof("Ignore update service(%s/%s) status as up to date", service.Namespace, service.Name)
@@ -179,6 +180,7 @@ func aggregateIngressStatus(object *unstructured.Unstructured, aggregatedStatusI
 
 		newStatus.LoadBalancer.Ingress = append(newStatus.LoadBalancer.Ingress, temp.LoadBalancer.Ingress...)
 	}
+	newStatus.LoadBalancer.Ingress = helper.DedupeAndSortIngressLoadBalancerIngress(newStatus.LoadBalancer.Ingress)
 
 	if reflect.DeepEqual(ingress.Status, *newStatus) {
 		klog.V(3).Infof("Ignore update ingress(%s/%s) status as up to date", ingress.Namespace, ingress.Name)
--- a/pkg/util/helper/service.go
+++ b/pkg/util/helper/service.go
@@ -1,0 +1,123 @@
+/*
+Copyright 2025 The Karmada Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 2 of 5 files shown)


pkg/resourceinterpreter/default/native/aggregatestatus.go
```diff
@@ -145,8 +146,18 @@ func aggregateServiceStatus(object *unstructured.Unstructured, aggregatedStatusI
 		klog.V(3).Infof("Grab service(%s/%s) status from cluster(%s), loadBalancer status: %v",
 			service.Namespace, service.Name, item.ClusterName, temp.LoadBalancer)
 
+		// If Error is not nil, add cluster name to the error message to indicate the error is collected from which member cluster.
+		for _, v := range temp.LoadBalancer.Ingress {
+			for _, port := range v.Ports {
+				if port.Error != nil {
+					*port.Error = fmt.Sprintf("(%v) %v", item.ClusterName, *port.Error)
```
XiShanYongYe-Chang, line 153: Does Adding Prefixes Affect Field [Format Requirements](https://github.com/karmada-io/karmada/blob/4c7c15ae5bd62b2cbb46070ebdf80b3f6c7f6179/vendor/k8s.io/api/core/v1/types.go#L7746-L7750)?

pkg/util/helper/service.go
```diff
@@ -0,0 +1,123 @@
...
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+// DedupeAndSortServiceLoadBalancerIngress dedupes and sorts the ServiceLoadBalancerIngress slice.
+func DedupeAndSortServiceLoadBalancerIngress(ingresses []corev1.LoadBalancerIngress) []corev1.LoadBalancerIngress {
+	ipModeMap := make(map[corev1.LoadBalancerIPMode]*corev1.LoadBalancerIPMode)
+	errMap := make(map[string]*string)
+	for i, ingress := range ingresses {
+		if ingress.IPMode != nil {
+			// let same IPMode use same ptr, to let *IPMode be comparable
```
XiShanYongYe-Chang, line 33: Just a question: for IPMode, its enumeration value has been declared, so for the same value, its address is the same?

XiShanYongYe-Chang (commented):
Thanks~
Other parts lgtm

pkg/util/helper/service.go
zach593: I haven't tested this in a real environment, but the status is unmarshaled from the rb/crb status by the informer, so they should be at different addresses.
If the variables are declared in the code, they could be at the same address.

As a helper function, I think its implementation should not be affected by these scenarios.

zhzhuang-zju:
> karmada-controller-manager: For Service and Ingress resources, the default resource interpreter will add member cluster name to the `.status.LoadBalancer.Ingress[].Ports[].Error` field if Error is not nil

@zach593 Can you remind me of the implementation corresponding to this release note?

zach593:
> > karmada-controller-manager: For Service and Ingress resources, the default resource interpreter will add member cluster name to the `.status.LoadBalancer.Ingress[].Ports[].Error` field if Error is not nil
> 
> @zach593 Can you remind me of the implementation corresponding to this release note?

I neglected to update the release note after the implementation was removed in response to the review comments.

zhzhuang-zju:
Thanks~
/lgtm

XiShanYongYe-Chang:
Let's merge it first.
/approve

PR #6258, opened 2025-04-02
when cluster is not-ready doesn't cleanup MultiClusterService and EndpointSlice work

### LINKED ISSUE #6255: Service access will be interrupted when CrossCluster MCS is used and the cluster's status is abnormal
**What happened**:
When the CrossCluster MCS is used for cross-cluster service access, when the provider cluster's status is abnormal (For example upgrade the cluster) . but the service is normal, the MCS-controller clears the work of the MCS. As a result, the service access will be interrupted.

When the CrossCluster MCS is used and the cluster status is abnormal, the mcs not automatically be cleared. 
https://github.com/karmada-io/karmada/blob/138537bdb201239b949408d8be65250b4e5afc4b/pkg/controllers/multiclusterservice/mcs_controller.go#L152，In this line, check whether the provider cluster status is abnormal. If the provider cluster status is abnormal, the work of the MCS and EPS will be deleted. If the cluster status is abnormal but the service is normal, service access will be interrupted. Therefore, do not delete the work of the MCS and EPS when the provider cluster status is abnormal. Access failures may occur when the cluster and services are both abnormal. However, the status of the cluster and services should be determined separately. It is not assumed that the services are also abnormal when the cluster status is abnormal.

Create an CrossCluster MCS. If the provider cluster's status is abnormal, service access from the consumer cluster will be interrupted.

**Environment**:
- Karmada version: 1.10.7
- kubectl-karmada or karmadactl version (the result of `kubectl-karmada version` or `karmadactl version`): 1.10.7

XiShanYongYe-Chang:
**What type of PR is this?**

/kind feature

For details about the scenarios, see #6255.

According to the confirmation with the feature author @Rains6 and @jwcesign, when the cluster is not ready, whether the pod in the cluster is healthy cannot be determined. Therefore, the endpointSliceresources collected from the cluster are deleted in a unified manner to prevent requests from accessing the pod when the pod is unhealthy.

The cluster is not ready, indicating that the Karmada control plane cannot access the kube-apiserver component of the member cluster or the component is faulty. In most cases, the pods on the nodes in the cluster are still running and providing services. If the Karmada control plane deletes the endpointSlice, services may be interrupted.

According to the preceding description, the current cluster health detection method cannot meet the following requirements: When the cluster is not ready, pods in the cluster are running properly or abnormally.

Therefore, it is best to support a case with less impact, that is, when the cluster is not ready, we do not delete the backend record EndpointSlice resource of the service. 

**This modification introduces new restrictions on the MultiClusterService feature. When the cluster is not ready and the backend pod of a service in the cluster cannot run properly, multi-cluster service discovery may fail. In this case, users need to delete the EndpointSlice resources that cannot provide services.**

Fixes #6255 

```release-note
`karmada-controller-manager`: when cluster is not-ready doesn't clean MultiClusterSerivce and EndpointSlice work.
```


XiShanYongYe-Chang:
After the PR is merged, we need to describe the feature restrictions in the website:

https://karmada.io/docs/userguide/service/multi-cluster-service-with-native-svc-access

Rains6:
LGTM

RainbowMango (approved):
/lgtm
/approve

PR #2388, opened 2022-08-16
Implement `karmdactl register` for joining clusters with PULL mode

lonelyCZ:
Signed-off-by: lonelyCZ <531187475@qq.com>

**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
Register a new member cluster.

```
[root@master68 lonelyCZ]# ./karmadactl register 10.10.103.67:32443 --cluster-name member68 --token hlvq6f.szq0580jalq1v9o2 --discovery-token-ca-cert-hash sha256:5d9fc2b87c186aaac219f883b4267704a94a0b2e2cc61b2c03bed2699dba9a55  --karmada-agent-image='swr.ap-southeast-1.myhuaweicloud.com/karmada/karmada-agent:latest'
[preflight] Running pre-flight checks
[prefligt] All pre-flight checks were passed
[karmada-agent-start] Waiting to perform the TLS Bootstrap
[karmada-agent-start] Waiting to construct karmada-agent kubeconfig
[karmada-agent-start] Waiting the necessary secret and RBAC
[karmada-agent-start] Waiting karmada-agent Deployment
W0823 15:34:47.857253     739 check.go:52] pod: karmada-agent-5d659b4746-fk8gl not ready. status: ContainerCreating
I0823 15:34:48.863019     739 check.go:49] pod: karmada-agent-5d659b4746-fk8gl is ready. status: Running

cluster(member68) is joined successfully
```

On karmada control plane.
```
[root@master67 karmada]# kapi get clusters
NAME       VERSION   MODE   READY   AGE
member68   v1.22.3   Pull   True    36s
```

If `karmada-agent` already installed in the member clusters.

```
[root@master68 lonelyCZ]# ./karmadactl register 10.10.103.67:32443 --cluster-name member68 --token hlvq6f.szq0580jalq1v9o2 --discovery-token-ca-cert-hash sha256:5d9fc2b87c186aaac219f883b4267704a94a0b2e2cc61b2c03bed2699dba9a55  --karmada-agent-image='swr.ap-southeast-1.myhuaweicloud.com/karmada/karmada-agent:latest'
[preflight] Running pre-flight checks
error execution phase preflight: [preflight] Some fatal errors occurred:
        [ERROR]: /etc/karmada/karmada-agent.conf already exists
        [ERROR]: /etc/karmada/pki/ca.crt already exists
        [ERROR]: karmada-system/karmada-kubeconfig Secret already exists
[...]

PR #6270, opened 2025-04-07
fix the issue where discovery-timeout fails to work properly

zhzhuang-zju:
**What type of PR is this?**
/kind bug

When the command "register" processes the "discovery-timeout", it directly uses the value of "int(o.Timeout)" as the number of seconds for the "timeout", which causes the "timeout" to become ineffective.  

For example:
- set flag `discovery-timeout` to 5m, the `int(o.Timeout)` will be 300000000000, leading to the timeout duration being greatly magnified. 
- set flag `discovery-timeout` to 10m, the command `register` will fail immediately

Fixes #

```release-note
`karmadactl`: Fixed the issue where option `discovery-timeout` fails to work properly
```

### DIFF
--- a/pkg/karmadactl/util/check.go
+++ b/pkg/karmadactl/util/check.go
@@ -59,10 +59,10 @@ func WaitForStatefulSetRollout(c kubernetes.Interface, sts *appsv1.StatefulSet, 
 	return nil
 }
 
-// WaitForDeploymentRollout  wait for Deployment reaches the ready state or timeout.
-func WaitForDeploymentRollout(c kubernetes.Interface, dep *appsv1.Deployment, timeoutSeconds int) error {
+// WaitForDeploymentRollout wait for Deployment reaches the ready state or timeout.
+func WaitForDeploymentRollout(c kubernetes.Interface, dep *appsv1.Deployment, timeout time.Duration) error {
 	var lastErr error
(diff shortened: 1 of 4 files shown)


pkg/karmadactl/util/check.go
```diff
@@ -60,9 +60,9 @@ func WaitForStatefulSetRollout(c kubernetes.Interface, sts *appsv1.StatefulSet,
 }
 
 // WaitForDeploymentRollout  wait for Deployment reaches the ready state or timeout.
-func WaitForDeploymentRollout(c kubernetes.Interface, dep *appsv1.Deployment, timeoutSeconds int) error {
+func WaitForDeploymentRollout(c kubernetes.Interface, dep *appsv1.Deployment, duration time.Duration) error {
```
RainbowMango, line 63:
```suggestion
func WaitForDeploymentRollout(c kubernetes.Interface, dep *appsv1.Deployment, timeout time.Duration) error {
```
timeout or timeoutDuration

RainbowMango (approved):
/lgtm
/approve

zhzhuang-zju:
@RainbowMango This issue was first introduced by #2388, so if a backport is needed, I'll cherry-pick this pr to release-1.11, release-1.12, and release-1.13

PR #6271, opened 2025-04-07
replace stopCh with context

zach593:
**What type of PR is this?**

/kind cleanup

In earlier versions of the Kubernetes library, `chan struct{}` was widely used to asynchronously signal cancellation events. However, as newer versions of Golang popularized the use of `context`, and Kubernetes itself began offering context-based library functions, its usage has become the standard. Due to historical reasons, Karmada’s current code still relies heavily on `chan struct{}`.

This mix has led to inconsistent propagation of `context` throughout the code. Developers often find themselves converting between `chan struct{}` and `context`, or even ignoring `context` altogether, which results in frequent use of `context.TODO()` and undermines the benefits that `context` is intended to provide.

Since using `context` allows for cancellation through `context.Done()`—which is compatible with the functionality of `chan struct{}`—this PR aims to replace all instances of `chan struct{}` with `context` wherever possible. This change will address the aforementioned issues and align the Karmada codebase with modern, idiomatic Golang practices.

Fixes #

```release-note
Many functions that used `stopCh chan struct{}` will be converted to use `context.Context`.
```


pkg/controllers/hpascaletargetmarker/hpa_scale_target_marker_controller.go
```diff
@@ -53,7 +53,7 @@ func (r *HpaScaleTargetMarker) SetupWithManager(mgr controllerruntime.Manager) e
 		ReconcileFunc: r.reconcileScaleRef,
 	}
 	r.scaleTargetWorker = util.NewAsyncWorker(scaleTargetWorkerOptions)
-	r.scaleTargetWorker.Run(scaleTargetWorkerNum, context.Background().Done())
+	r.scaleTargetWorker.Run(context.Background(), scaleTargetWorkerNum)
```
XiShanYongYe-Chang, line 56: In this case, is it better to pass ctx through the controller parameter?

pkg/scheduler/framework/runtime/metrics_recorder.go
```diff
@@ -42,20 +43,23 @@ type metricsRecorder struct {
 	// how often the recorder runs to flush the metrics.
 	interval time.Duration
 
-	// stopCh is used to stop the goroutine which periodically flushes metrics. It's currently only
+	// ctx and cancel are used to stop the goroutine which periodically flushes metrics. They're currently only
 	// used in tests.
-	stopCh chan struct{}
+	ctx    context.Context
+	cancel context.CancelFunc
```
XiShanYongYe-Chang, line 49: `cancel` seems to be completely unused, including testing.

zach593: `stopCh` is not used either, I just replaced it with `context` and `cancel` in place and skipped considering its use. After all, if there is no `cancel`, it is also meaningless to just hold this `context`

pkg/controllers/hpascaletargetmarker/hpa_scale_target_marker_controller.go
zach593: For the entire repo that's true, but I don't really want to make this change in this PR. Since this PR already contains too many modifications, it should be a pure refactor that doesn't affect any specific functionality. Introducing changes like linking the upstream and downstream `context`, which could potentially impact the code logic, would make it harder to review.

XiShanYongYe-Chang: Thanks for clarifying.

XiShanYongYe-Chang:
Thanks
/lgtm

RainbowMango (commented):
/assign
@zach593 Here is a conflict needs to be resolved.

cmd/descheduler/app/descheduler.go
```diff
@@ -144,12 +144,6 @@ func run(opts *options.Options, stopChan <-chan struct{}) error {
 	karmadaClient := karmadaclientset.NewForConfigOrDie(restConfig)
 	kubeClient := kubernetes.NewForConfigOrDie(restConfig)
 
-	ctx, cancel := context.WithCancel(context.Background())
-	go func() {
-		<-stopChan
-		cancel()
-	}()
```
RainbowMango, line 151: +1
neat

pkg/resourceinterpreter/interpreter.go
```diff
@@ -101,7 +101,7 @@ func (i *customResourceInterpreterImpl) Start(ctx context.Context) (err error) {
 	i.informer.Start()
 	i.informer.WaitForCacheSync()
 	<-ctx.Done()
-	klog.Infof("Stopped as stopCh closed.")
+	klog.Infof("Stopped as context canceled.")
```
RainbowMango, line 104: nice catch!

pkg/scheduler/internal/queue/scheduling_queue.go
```diff
@@ -153,7 +153,7 @@ func NewSchedulingQueue(opts ...Option) SchedulingQueue {
 //   - unschedulableBindings holds bindings that were already attempted for scheduling and
 //     are currently determined to be unschedulable.
 type prioritySchedulingQueue struct {
-	// stopCh is used to stop the goroutine which periodically flushes pending bindings.
+	// stop is used to stop the goroutine which periodically flushes pending bindings.
```
RainbowMango, line 156: nice catch +1

RainbowMango (approved):
/lgtm
/approve

Good Job! Thanks for the hard work.

Sorry for letting this sit so long.  My suggestion is splitting this kind of work to several small pieces, it helps to move PR faster.

Issue #6266, opened 2025-04-03
[Umbrella] Bind the listening address of karmada component  to the POD IP


XiShanYongYe-Chang:
**What would you like to be added**:

I submitted a document on the website: https://github.com/karmada-io/website/pull/813

In the configuration guide, user can set the listening address to the POD IP address. We can change the listening addresses of Karmada components in different installation modes based on the suggestions.

Currently, Karmada supports the following installation modes:
 
- [x] kind @seanlaii #6285 
- [x] helm @seanlaii #6324 
- [x] operator @seanlaii #6321
- [x] karmadactl init @seanlaii #6323 

The Karmada components that can be configured include:

- karmada-controller-manager
- karmada-scheduler
- karmada-aggregated-apiserver
- karmada-search
- karmada-scheduler-estimator
- karmada-agent
- karmada-webhook
- karmada-metrics-adapter
- karmada-descheduler

**Why is this needed**:

When a user deploys a Karmada component, the Karmada component will use the default listening address and port if the listening-related parameters are not configured. The default listening address of the Karmada component is `0.0.0.0`, which means that the component will listen to all network interfaces on the server (for a container, this is a virtual network interface that is managed by the container runtime, such as Docker).

All-zero listening will increase the risk of being attacked. So, we can change the listening address to the POD IP address.

seanlaii:
Hi @XiShanYongYe-Chang, thanks for the info.

Just to confirm my understanding: are you suggesting changing the default address settings (like the --metrics-bind-address flag, default to 0.0.0.0) to use the specific Pod IP address?

For example, would this mean modifying the [--bind-address in karmada-scheduler.yaml](https://github.com/karmada-io/karmada/blob/master/artifacts/deploy/karmada-scheduler.yaml#L45-L46) from 0.0.0.0 to the Pod's IP?

seanlaii:
Got it. Thanks! I could help work on this.

XiShanYongYe-Chang:
Thanks a lot, @seanlaii 

So let's deal with the kind installation, which is through the hack script?

seanlaii:
Yes, sounds good to me.

XiShanYongYe-Chang:
Hi @seanlaii, Would you like to continue contributing to the rest?

seanlaii:
Sure, I could take the operator first.

seanlaii:
[...]

PR #6285, opened 2025-04-11
Update bind addresses to use POD_IP for local-up script

seanlaii:
**What type of PR is this?**
/kind cleanup

This PR updates the listening addresses for multiple components (agent, metrics adapter, webhook, etc.) to use the Pod's IP address via the `POD_IP` environment variable. By binding to the specific Pod IP rather than a generic address like `0.0.0.0`, the components are restricted to listening only on their assigned network interface. This improves security by reducing exposure, and ensures that metrics and health probes are correctly bound to the right interface, aiding in more predictable network behavior.

Part of #6266 

- The changes affect all relevant deployments.
- Ensure that after these changes, all inter-component communications (e.g., metrics scrape, liveness, and readiness probes) continue to work as expected.

```release-note
NONE
```

### DIFF
--- a/artifacts/agent/karmada-agent.yaml
+++ b/artifacts/agent/karmada-agent.yaml
@@ -24,6 +24,11 @@ spec:
         - name: karmada-agent
           image: docker.io/karmada/karmada-agent:latest
           imagePullPolicy: {{image_pull_policy}}
+          env:
+            - name: POD_IP
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.podIP
           command:
             - /bin/karmada-agent
             - --karmada-kubeconfig=/etc/karmada/config/karmada.config
@@ -31,8 +36,8 @@ spec:
             - --cluster-name={{member_cluster_name}}
             - --cluster-api-endpoint={{member_cluster_api_endpoint}}
             - --cluster-status-update-frequency=10s
-            - --health-probe-bind-address=0.0.0.0:10357
-            - --metrics-bind-address=:8080
+            - --metrics-bind-address=$(POD_IP):8080
+            - --health-probe-bind-address=$(POD_IP):10357
             - --feature-gates=CustomizedClusterResourceModeling=true,MultiClusterService=true
             - --v=4
           livenessProbe:
--- a/artifacts/deploy/karmada-webhook.yaml
+++ b/artifacts/deploy/karmada-webhook.yaml
@@ -26,11 +26,17 @@ spec:
             privileged: false
           image: docker.io/karmada/karmada-webhook:latest
           imagePullPolicy: IfNotPresent
+          env:
+            - name: POD_IP
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.podIP
           command:
             - /bin/karmada-webhook
             - --kubeconfig=/etc/karmada/config/karmada.config
-            - --bind-address=0.0.0.0
-            - --metrics-bind-address=:8080
+            - --bind-address=$(POD_IP)
+            - --metrics-bind-address=$(POD_IP):8080
+            - --health-probe-bind-address=$(POD_IP):8000
             - --default-not-ready-toleration-seconds=30
             - --default-unreachable-toleration-seconds=30
             - --secure-port=8443
--- a/artifacts/deploy/karmada-aggregated-apiserver.yaml
+++ b/artifacts/deploy/karmada-aggregated-apiserver.yaml
@@ -27,6 +27,11 @@ spec:
           securityContext:
             allowPrivilegeEscalation: false
             privileged: false
+          env:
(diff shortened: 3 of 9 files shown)


XiShanYongYe-Chang:
Thanks @seanlaii 
/assign

artifacts/agent/karmada-agent.yaml
```diff
@@ -24,17 +24,25 @@ spec:
...
-            - --cluster-api-endpoint={{member_cluster_api_endpoint}}
-            - --cluster-status-update-frequency=10s
-            - --health-probe-bind-address=0.0.0.0:10357
-            - --metrics-bind-address=:8080
-            - --feature-gates=CustomizedClusterResourceModeling=true,MultiClusterService=true
-            - --v=4
+            - /bin/sh
+            - -c
+            - |
+              exec /bin/karmada-agent \
```
XiShanYongYe-Chang, line 36: Do we need to modify the startup mode of the component?

artifacts/deploy/karmada-webhook.yaml
```diff
@@ -26,16 +26,25 @@ spec:
...
-            - --default-unreachable-toleration-seconds=30
-            - --secure-port=8443
-            - --cert-dir=/etc/karmada/pki/server
-            - --v=4
+            - /bin/sh
+            - -c
+            - |
+              exec /bin/karmada-webhook \
+                --kubeconfig=/etc/karmada/config/karmada.config \
+                --bind-address=0.0.0.0 \
```
XiShanYongYe-Chang, line 40: We can also change the binding-address address to the pod IP address.

artifacts/agent/karmada-agent.yaml
seanlaii: I found that we might need to use `$(POD_IP)` instead of `${POD_IP}`. With this, we can reuse the original command.

XiShanYongYe-Chang: (I understand you may write it backward.) Yes, I also found out, is that because the system parses differently?

XiShanYongYe-Chang: I suggest that we do not change the startup mode for this time. Currently, try ${POD_IP} first. When the startup command mode needs to be changed in the future, we can consider changing the startup mode.

seanlaii: Yes, I agree that we should minimize the change to the startup method. Therefore, I change them back and use $(POD_IP) instead. Please check the latest change. Thank you!

> (I understand you may write it backward.) Yes, I also found out, is that because the system parses differently?

It seems that it is the official method for referencing `env` with `command` in Kubernetes: https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#use-environment-variables-to-define-arguments.

XiShanYongYe-Chang: Let's discuss it in a separate pr if we need to modify it.

artifacts/agent/karmada-agent.yaml
```diff
@@ -24,15 +24,20 @@ spec:
...
           command:
             - /bin/karmada-agent
             - --karmada-kubeconfig=/etc/karmada/config/karmada.config
             - --karmada-context={{karmada_context}}
             - --cluster-name={{member_cluster_name}}
             - --cluster-api-endpoint={{member_cluster_api_endpoint}}
             - --cluster-status-update-frequency=10s
-            - --health-probe-bind-address=0.0.0.0:10357
-            - --metrics-bind-address=:8080
+            - --metrics-bind-address=$(POD_IP):8080
```
XiShanYongYe-Chang, line 39: It doesn't seem to match what I understand. Shouldn't we use the way environment variables are replaced? `${POD_IP}`

seanlaii: Thanks for the observation. In Kubernetes `command`/`args`, the specific `$(VAR_NAME)` syntax is required for environment variable substitution handled by Kubernetes itself *before* the command runs.

The `${VAR_NAME}` syntax is for `shell` expansion, which doesn't automatically happen here unless we explicitly invoke a shell (e.g., `sh -c "..."`). Therefore, `$(POD_IP)` is the correct way to reference the variable in this context according to Kubernetes conventions.

If we use `${POD_IP}` with the current start up method, we would encounter:
```bash
E0412 17:40:11.649275       1 controllermanager.go:184] Failed to build controller manager: error listening on ${POD_IP}:10357: listen tcp: lookup ${POD_IP}: no such host
E0412 17:40:11.649306       1 run.go:72] "command failed" err="error listening on ${POD_IP}:10357: listen tcp: lookup ${POD_IP}: no such host"
```

XiShanYongYe-Chang (commented):
Thanks a lot~
/lgtm
/approve

Issue #6273, opened 2025-04-07
[Umbrella] Address golangci-lint v2 formatting issues


seanlaii:
**What would you like to be added**:
Applies code formatting changes to resolve linting violations reported by the [upgraded golangci-lint v2](https://github.com/karmada-io/karmada/pull/6237).

Ref: https://github.com/karmada-io/karmada/actions/runs/14120914184/job/39560873286?pr=6237

**Why is this needed**:
The following issues stem from new checks [introduced in v2](https://github.com/karmada-io/karmada/pull/6237), which were temporarily disabled for incremental resolution.
Fix #6237 

**_staticcheck_**:

- [x] QF1008: could remove embedded field "xxx" from selector (staticcheck). (Decided not to fix, but add comment) #6295  
- [x] ST1019: package "xxx" is being imported more than once (staticcheck). #6289 
- [x] ST1005: error strings should not be capitalized (staticcheck). #6290 
- [x] QF1004: could use strings.ReplaceAll instead (staticcheck). #6291 
- [x] ST1011: var periodSeconds is of type time.Duration; don't use unit-specific suffix "Seconds" (staticcheck). #6296 
- [x] QF1003: could use tagged switch on operationResult (staticcheck) #6292
- [x] QF1001: could apply De Morgan's law (staticcheck) #6297 
- [x] ST1000: at least one file in a package should have a package comment (staticcheck) #6299 

**_revive_**
- [x] package-comments: should have a package comment (revive) #6299

seanlaii:
Hi @RainbowMango , @XiShanYongYe-Chang , and @zhzhuang-zju , 

Looking at the `staticcheck ST1000` and `revive package-comments` errors, it seems fixing them involves adding comments to the packages across the entire project. This represents a substantial amount of work. 
I wanted to raise the question of whether we should prioritize fixing these right now, or perhaps exclude them. 
Happy to hear your thoughts on this.

PR #6288, opened 2025-04-12
Fix lint issue QF1008

seanlaii:
**What type of PR is this?**
/kind cleanup

Fix lint issue - `QF1008: could remove embedded field "xxx" from selector (staticcheck)`.

Part of #6273 

```release-note
NONE
```


XiShanYongYe-Chang (commented):
The code modification is fine, but it seems to me that the code before the modification is easier to read.

WDYT?

seanlaii:
> The code modification is fine, but it seems to me that the code before the modification is easier to read.
> 
> WDYT?

Yes, I prefer the original format as well. I can close this PR if there is no other opinion. Thanks for your time!

XiShanYongYe-Chang:
It's ok for me. How about other guys, @RainbowMango @zhzhuang-zju @dongjiang1989

zhzhuang-zju:
> Yes, I prefer the original format as well. I can close this PR if there is no other opinion. Thanks for your time!

+1. I tried to find more information about this static check, such as the benefits it can bring, but I couldn't find any. If it reduces the readability of the code without any obvious benefits, I think we can put off adopting it.

RainbowMango:
https://staticcheck.dev/docs/checks#QF1008

I don't see any benefits from it. I think we can disable it. And add a comment to explain it.

seanlaii:
Sounds good. Will close this and create another PR to add the comment.

PR #6320, opened 2025-04-24
pkg/search/backendstore: fix OpenSearch create index if not exist before deleting it in `delete`

mohamedawnallah:
## Description

In this commit, we fix the open search deleting behavior. When delete is invoked and the index not exist it create the index and then delete it which is not expected. 

**What type of PR is this?**

/kind bug

## Motivation and Context

Follow-up on discussion https://github.com/karmada-io/karmada/pull/5824#discussion_r2055370955.

```release-note
NONE
```

### DIFF
--- a/pkg/search/backendstore/opensearch.go
+++ b/pkg/search/backendstore/opensearch.go
@@ -36,7 +36,10 @@ import (
 	searchv1alpha1 "github.com/karmada-io/karmada/pkg/apis/search/v1alpha1"
 )
 
-var defaultPrefix = "kubernetes"
+var (
+	defaultPrefix       = "kubernetes"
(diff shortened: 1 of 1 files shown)


mohamedawnallah:
cc: @zhzhuang-zju, @XiShanYongYe-Chang

RainbowMango:
Currently we have Copilot, but don't know why it can not start reviewing automatically. :)

moawnallah:
> Currently we have Copilot, but don't know why it can not start reviewing automatically. :)

So currently it seems there is choice between using the existing Copilot reviewer or the gemini one proposed in this PR #6515, right?

Gemini seems has large context window which seems helpful in most PRs:
https://developers.google.com/gemini-code-assist

What do you think, @RainbowMango?

PR #6358, opened 2025-05-08
Proposal: Support Suspension of Reconciliation for Karmada Control Planes

jabellard:
**What type of PR is this?**
/kind design

Details are in the design doc.

Fixes #

```release-note
The new optional `spec.suspend` field can be used to temporarily pause/suspend the reconciliation of a `Karmada` object.
```

### DIFF
--- a/docs/proposals/karmada-operator/reconciliation-suspension/README.md
+++ b/docs/proposals/karmada-operator/reconciliation-suspension/README.md
@@ -1,0 +1,83 @@
+---
+title: Support Suspension of Reconciliation for Karmada Control Planes
+authors:
+- "@jabellard"
+reviewers:
+- "@RainbowMango"
+approvers:
+- "@RainbowMango"
+creation-date: 2025-05-08
+---
+
+# Support Suspension of Reconciliation for Karmada Control Planes
+
+## Summary
+
+This proposal adds support to pause/suspend the reconciliation of a Karmada resource by the operator. When paused/suspended, the operator should refrain from enqueuing reconciliation requests for changes to the primary `Karmada` resource, or any of its related objects.
+
+
+## Motivation
+
+1. **Emergency Troubleshooting**  
+   Platform engineers can hot‑patch components without having the Karmada operator overwrite their critical emergency patches.
+
+2. **Coordinated Maintenance**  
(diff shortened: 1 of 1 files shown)


docs/proposals/karmada-operator/reconciliation-suspension/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+Add the optional `paused` **boolean** field to the `Karmada` spec:
+
+```go
+// KarmadaSpec defines the desired state of Karmada.
+type KarmadaSpec struct {
+    // ...existing fields...
+    
+    // Paused indicates that the operator should suspend reconciliation
+    // for this Karmada control plane and all its managed resources.
```
RainbowMango, line 56: Add additional comments like `it does not apply to already started instance`? And also the default value?

docs/proposals/karmada-operator/reconciliation-suspension/README.md
```diff
@@ -0,0 +1,83 @@
...
+    // +optional
+    Paused bool `json:"paused,omitempty"`
```
RainbowMango, line 58:
```suggestion
    Paused *bool `json:"paused,omitempty"`
```

RainbowMango: In addition, comparing `suspend` and `pause`, which one do you prefer?

jabellard: I think either one is fine, but `suspend` actually aligns more closely with the terminology used in other projects like like ArgoCD and FluxCD, so I can make the change to use that term instead.

docs/proposals/karmada-operator/reconciliation-suspension/README.md
jabellard: Will update to make the intent more clear.

jabellard:
@RainbowMango , thanks for the constructive feedback. I just pushed a change to address your comments.

RainbowMango (commented):
/assign

RainbowMango (approved):
/approve

Leave LGTM to @zhzhuang-zju for another look.

PS: I was thinking if we need to introduce another condition to show the suspended status, we can postpone it until there is a real use case.

PR #6359, opened 2025-05-09
Support Suspension of Reconciliation for Karmada Control Planes

jabellard:
**What type of PR is this?**

/kind feature

Implementation of this [proposal](https://github.com/karmada-io/karmada/pull/6358).

Fixes #

```release-note
`karmada-operator`: The new optional `spec.suspend` field can be used to temporarily pause/suspend the reconciliation of a `Karmada` object.
```

### DIFF
--- a/operator/pkg/controller/karmada/controller.go
+++ b/operator/pkg/controller/karmada/controller.go
@@ -159,9 +159,16 @@ func (ctrl *Controller) SetupWithManager(mgr controllerruntime.Manager) error {
 		For(&operatorv1alpha1.Karmada{},
 			builder.WithPredicates(predicate.Funcs{
 				UpdateFunc: ctrl.onKarmadaUpdate,
-			})).
-		WithEventFilter(skipIfSuspended()).
+			}, skipIfSuspended())).
 		Complete(ctrl)
+}
+
+// skipIfSuspended filters reconciliation of Karmada instances that have been suspended.
+func skipIfSuspended() predicate.Predicate {
+	return predicate.NewPredicateFuncs(func(obj client.Object) bool {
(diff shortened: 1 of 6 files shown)


RainbowMango (commented):
/assign

operator/pkg/controller/karmada/controller.go
```diff
@@ -160,6 +160,7 @@ func (ctrl *Controller) SetupWithManager(mgr controllerruntime.Manager) error {
 			builder.WithPredicates(predicate.Funcs{
 				UpdateFunc: ctrl.onKarmadaUpdate,
 			})).
+		WithEventFilter(skipIfSuspended()).
```
RainbowMango, line 163: Put the event filter into `For`?  `builder.WithPredicates(skipIfSuspended())`

No big deal though. 
Note that the EventFileter will be the global filter that works for `Owns` and `Watches`, in addition to `For`. We might don't need the `Owns` and `Watches` now, but once they are involved in the future, the event filter probably not work, because the object type passing in the function might not be `Karmada`.

For the operator controller, we might don't need to use `Owns` and `Watches`, but once they are used, the event  filter w

RainbowMango (commented):
Generally looks good to me, just a nit.

By the way, anything else need to do along with this?

jabellard:
> Generally looks good to me, just a nit.
> 
> By the way, anything else need to do along with this?

Just pushed update to address. And no, don't think so. This is a simple change.

RainbowMango (approved):
/lgtm
/approve

RainbowMango:
/retest
The failing test is unrelated

Issue #6350, opened 2025-05-06
[Feature] FederatedResourceQuota Enhancement


mszacillo:
**Summary**
This feature aims to enhance the existing FederatedResourceQuota so it can impose namespaced resource limits directly on the Karmada control-plane level. This feature enhancement has several benefits, including:

1. Track resource consumption and limits from a unified place
2. Prevent users from over-consuming aggregate resources across clusters
3. Easier for users of application and cluster failover to ensure that fallback clusters have enough resources to house all the user's workflows

**Proposal:**

- [X] FederatedResourceQuota Enhancement: (https://github.com/karmada-io/karmada/pull/6181)

**Iteration Tasks -- Part-1: FederatedResourceQuota Enhancement MVP**

 - [ ] Update federated-resource-quota-status-controller to skip reconcile if staticAssignment is not set
 - [ ] FederatedResourceQuota validating webhook should block updates to enable staticAssignment on Update requests
 - [ ] Create new overall federated resource quota status controller:
         - Watch create/update events for FRQ, and update federatedresourcequota.status.overall
         - Watch deletion of RB -> find related FRQ -> recalculate quota usage
         - Periodically recalculate FRQ status
- [ ] Add ResourceBinding validation webhook to monitor updates to resourcebinding.spec and resourcebinding.spec.replicaRequirements after scheduling result
         - Check delta in quota usage and deny request if above limits
         - Otherwise update federatedresourcequota.status.overallUsed and then approve update request
- [ ] Determine if changes necessary to scheduler when reacting to denied RB updates
- [ ] FederatedResourceQuota controller should requeue unschedulable RBs to be scheduled in the case of a change to overall quota usage

**Iteration Tasks -- Part-2:  Further Enhancements beyond MVP**

- [ ] [API Change] Add Scope to the FederatedResourceQuotaSpec
- [ ] Consider improvements to reconcile strategy for overall FRQ status controller (use cache instead of listing all RB each reconcile)

mszacillo:
@RainbowMango Let me know if this looks alright! 

Also, @zhzhuang-zju, do you have any preference on how to subdivide this work between us?

RainbowMango:
Generally looks great! Thanks.
I will leave comments if I want to twist or add something. 
[...]

PR #6363, opened 2025-05-10
FederatedResourceQuota: skip creating redundant ResourceQuotas to member clusters

zhzhuang-zju:
**What type of PR is this?**
/kind feature

Adjust the behavior of the legacy staticAssignments to skip creating redundant ResourceQuotas to member clusters

Parts of #6350

```release-note

```

### DIFF
--- a/pkg/controllers/federatedresourcequota/federated_resource_quota_sync_controller.go
+++ b/pkg/controllers/federatedresourcequota/federated_resource_quota_sync_controller.go
@@ -18,6 +18,7 @@ package federatedresourcequota
 
 import (
 	"context"
+	"reflect"
 
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -77,6 +78,11 @@ func (c *SyncController) Reconcile(ctx context.Context, req controllerruntime.Re
 		return controllerruntime.Result{}, err
 	}
 
+	if err := c.cleanUpOrphanWorks(ctx, quota); err != nil {
+		klog.Errorf("Failed to cleanup orphan works for federatedResourceQuota(%s), error: %v", req.NamespacedName.String(), err)
+		return controllerruntime.Result{}, err
+	}
+
 	clusterList := &clusterv1alpha1.ClusterList{}
 	if err := c.Client.List(ctx, clusterList); err != nil {
 		klog.Errorf("Failed to list clusters, error: %v", err)
@@ -96,8 +102,15 @@ func (c *SyncController) Reconcile(ctx context.Context, req controllerruntime.Re
 // SetupWithManager creates a controller and register to controller manager.
 func (c *SyncController) SetupWithManager(mgr controllerruntime.Manager) error {
 	fn := handler.MapFunc(
-		func(ctx context.Context, _ client.Object) []reconcile.Request {
+		func(ctx context.Context, obj client.Object) []reconcile.Request {
 			var requests []reconcile.Request
+			var clusterName string
+			switch obj.(type) {
+			case *clusterv1alpha1.Cluster:
+				clusterName = obj.GetName()
+			default:
+				return nil
+			}
 
 			FederatedResourceQuotaList := &policyv1alpha1.FederatedResourceQuotaList{}
 			if err := c.Client.List(ctx, FederatedResourceQuotaList); err != nil {
@@ -105,6 +118,10 @@ func (c *SyncController) SetupWithManager(mgr controllerruntime.Manager) error {
 			}
 
 			for _, federatedResourceQuota := range FederatedResourceQuotaList.Items {
+				hard := extractClusterHardResourceList(federatedResourceQuota.Spec, clusterName)
+				if hard == nil {
+					continue
+				}
 				requests = append(requests, reconcile.Request{
 					NamespacedName: types.NamespacedName{
 						Namespace: federatedResourceQuota.Namespace,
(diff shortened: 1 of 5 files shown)


zhzhuang-zju:
@RainbowMango @mszacillo Do we need to verify whether the clusters in staticAssignments exist? For example, Karmada manages three member clusters: member1, member2, and member3, the staticAssignments of a FRQ is:
```yaml
  staticAssignments:
  - clusterName: member4
    hard:
      cpu: 40m
```
Should we report an error, or ignore this configuration? If we ignore it, but member4 registers to Karmada later, would we need to respond to the cluster create event and create a ResourceQuota for the member?

RainbowMango (commented):
/assign

fivemtebex:
Excellent contribution to open source! The community will benefit greatly from this.

fivemserverhostings:
Very clean code with good test coverage. This is a great example of best practices.

fivemeyestore:
This is a well-structured codebase. I appreciate the clean separation of concerns.

fivemhostingserver:
Excellent contribution to open source! The community will benefit greatly from this.

scriptforfivem:
I'm impressed by the modularity of this codebase. Makes it very maintainable.

pkg/controllers/federatedresourcequota/federated_resource_quota_sync_controller.go
```diff
@@ -164,15 +179,40 @@ func (c *SyncController) cleanUpWorks(ctx context.Context, namespace, name strin
...
+func (c *SyncController) cleanUpOrphanWorks(ctx context.Context, quota *policyv1alpha1.FederatedResourceQuota) error {
+	var errs []error
+	workList := &workv1alpha1.WorkList{}
+	if err := c.List(ctx, workList, client.MatchingLabels{
+		util.FederatedResourceQuotaNamespaceLabel: quota.GetNamespace(),
+		util.FederatedResourceQuotaNameLabel:      quota.GetName(),
+	}); err != nil {
+		klog.Errorf("Failed to list works, err: %v", err)
+		return err
+	}
```
RainbowMango, line 191: Given that listing all works is an extremely expensive operation, we should minimize the reconcile frequency by introducing an event handler that only responds to 
- `create event`: only when `.spec.staticAssignments` not nil
- `update event`: only when `.spec.staticAssignments` changed
- `delete event`: only when `.spec.staticAssignments` not nil

pkg/controllers/federatedresourcequota/federated_resource_quota_sync_controller.go
```diff
@@ -83,7 +83,11 @@ func (c *SyncController) Reconcile(ctx context.Context, req controllerruntime.Re
 		return controllerruntime.Result{}, err
 	}
 
-	if err := c.buildWorks(ctx, quota, clusterList.Items); err != nil {
+	if err := c.cleanUpOrphanWorks(ctx, quota); err != nil {
```
RainbowMango, line 86: Move `cleanUpOrphanWorks` ahead of `Client.List(ctx, clusterList)`, so that if something wrong happened when cleaning the works could save the effort of listing clusters.

pkg/controllers/federatedresourcequota/federated_resource_quota_sync_controller.go
```diff
@@ -164,15 +179,40 @@ func (c *SyncController) cleanUpWorks(ctx context.Context, namespace, name strin
...
+
+func (c *SyncController) buildWorks(ctx context.Context, quota *policyv1alpha1.FederatedResourceQuota, clusters []string) error {
 	var errs []error
 	for _, cluster := range clusters {
 		resourceQuota := &corev1.ResourceQuota{}
 		resourceQuota.APIVersion = "v1"
 		resourceQuota.Kind = "ResourceQuota"
 		resourceQuota.Namespace = quota.Namespace
 		resourceQuota.Name = quota.Name
-		resourceQuota.Spec.Hard = extractClusterHardResourceList(quota.Spec, cluster.Name)
```
RainbowMango, line 175: It's quite easy to ignore the non-existent cluster in staticAssignments by checking the return value of `extractClusterHardResourceList(quota.Spec, cluster.Name)`.

I mean we don't need `extractExistedClusters()`.

pkg/controllers/federatedresourcequota/federated_resource_quota_sync_controller.go
zhzhuang-zju: Good concern. However, when upgrading from a lower version to v1.14.0, there may be orphan works that were not cleaned up

pkg/controllers/federatedresourcequota/federated_resource_quota_sync_controller.go
```diff
@@ -132,9 +149,26 @@ func (c *SyncController) SetupWithManager(mgr controllerruntime.Manager) error {
 		},
 	})
 
+	frqPredicate := builder.WithPredicates(predicate.Funcs{
+		// TODO(zhzhuang-zju) To maintain compatibility during upgrades, filtering for work creation events will not be added for now. Once version v1.15 is released, we will be able to exclude works where obj.Spec.StaticAssignments is nil.
```
RainbowMango, line 153:
```suggestion
		// TODO(zhzhuang-zju): Should ignore the create event in case of the FRQ's .spec.staticAssignments is nil, as
		// no longer sync empty ResourceQuotas by default starting from v1.14. Here temporarily omitting this filter to
		// ensure that during upgrades from older versions, there is an opportunity to clean up legacy ResourceQuotas synced
		// from previous versions.
```
We are not filter `work` here.

pkg/controllers/federatedresourcequota/federated_resource_quota_sync_controller.go
```diff
@@ -132,9 +149,26 @@ func (c *SyncController) SetupWithManager(mgr controllerruntime.Manager) error {
...
+		CreateFunc: func(event.CreateEvent) bool {
+			return true
+		},
+		UpdateFunc: func(updateEvent event.UpdateEvent) bool {
+			oldObj := updateEvent.ObjectOld.(*policyv1alpha1.FederatedResourceQuota)
+			newObj := updateEvent.ObjectNew.(*policyv1alpha1.FederatedResourceQuota)
+			return !reflect.DeepEqual(oldObj.Spec.StaticAssignments, newObj.Spec.StaticAssignments)
+		},
+		// TODO(zhzhuang-zju) To maintain compatibility during upgrades, filtering for work deletion events will not be added for now. Once version v1.15 is released, we will be able to exclude works where obj.Spec.StaticAssignments is nil.
```
RainbowMango, line 162:
```suggestion
		// TODO(zhzhuang-zju): Should ignore the delete event in case of the FRQ's .spec.staticAssignments is nil, as
		// no longer sync empty ResourceQuotas by default starting from v1.14. Here temporarily omitting this filter to
		// ensure that during upgrades from older versions, there is an opportunity to clean up legacy ResourceQuotas synced
		// from previous versions.
```

RainbowMango (commented):
Generally looks good to me except some nits.

RainbowMango (approved):
/lgtm
/approve

PR #6365, opened 2025-05-12
e2e: cleanup all namespaces created both in control plane and member clusters

zhzhuang-zju:
**What type of PR is this?**
/kind cleanup

In the `ginkgo.SynchronizedBeforeSuite`, we create two namespaces: testNamespace and secondaryTestNamespace. However, during the cleanup phase, only testNamespace will be cleaned up, resulting in a leftover namespace.

Fixes #

```release-note

```

### DIFF
--- a/test/e2e/suites/base/suite_test.go
+++ b/test/e2e/suites/base/suite_test.go
@@ -118,13 +118,10 @@ var (
 	controlPlaneClient client.Client
 	// testNamespace is the main namespace for testing.
 	// It is the default namespace where most test resources are created and validated.
-	testNamespace string
-	// secondaryTestNamespace is an additional namespace used for testing.
-	// It is only required in specific scenarios where resources need to be created and tested across multiple namespaces.
-	secondaryTestNamespace string
-	clusterProvider        *cluster.Provider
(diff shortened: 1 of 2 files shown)


XiShanYongYe-Chang:
Thanks
/assign

zhzhuang-zju:
Since `secondaryTestNamespace` is only used in search tests, we could consider creating it only there, thus eliminating the need for its creation and deletion in other contexts.

test/e2e/suites/base/suite_test.go
```diff
@@ -211,6 +211,9 @@ var _ = ginkgo.SynchronizedAfterSuite(func() {
 	// It will not return error even if there is no such namespace in there that may happen in case setup failed.
 	err := cleanupTestNamespace(testNamespace, kubeClient)
 	gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+
```
XiShanYongYe-Chang, line 214: Because of the comment, this and the above content should belong to the same code block, we can delete the blank line.

XiShanYongYe-Chang (commented):
/lgtm
/approve

PR #6367, opened 2025-05-13
Create federated-resource-quota-enforcement-controller

mszacillo:
**What type of PR is this?**

Adds the federated-resource-quota-enforcement controller. 

Reconciles in response to one of the following events:
1. Creation or update of a FederatedResourceQuota
2. Deletion of a ResourceBinding

At the moment I did not include the third planned reconcile: `Periodically recalculate FRQ status`, since I was worried about using `RequeueAfter` and hammering the controller with too many reconcile events. Perhaps we can add this as a follow-up iteration task. 

/kind feature

Part of #6350.

```release-note
`karmada-controller-manager`: Added `federated-resource-quota-enforcement-controller` as part of ResourceQuotaEnforcement feature.
```

### DIFF
--- a/pkg/controllers/federatedresourcequota/federated_resource_quota_enforcement_controller.go
+++ b/pkg/controllers/federatedresourcequota/federated_resource_quota_enforcement_controller.go
@@ -1,0 +1,213 @@
+/*
+Copyright 2025 The Karmada Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package federatedresourcequota
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/record"
+	"k8s.io/client-go/util/retry"
+	"k8s.io/klog/v2"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"github.com/karmada-io/karmada/pkg/apis/policy/v1alpha1"
+	policyv1alpha1 "github.com/karmada-io/karmada/pkg/apis/policy/v1alpha1"
+	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
+	"github.com/karmada-io/karmada/pkg/events"
+	"github.com/karmada-io/karmada/pkg/util/helper"
+)
+
+const (
+	// QuotaEnforcementControllerName is the controller name that will be used when reporting events.
+	QuotaEnforcementControllerName = "federated-resource-quota-enforcement-controller"
+)
+
+// QuotaEnforcementController reflects the overall resource usage in namespaced FederatedResourceQuotas.
+// The controller will determine resource usage based off the replicaRequirements defined in the relevant
+// ResourceBindings found in the namespace where a FederatedResourceQuota is applied.
+//
+// The total sum of replicaRequirements cannot exceed the limits set by the quota.
+type QuotaEnforcementController struct {
+	client.Client // used to operate Work resources.
+	EventRecorder record.EventRecorder
+}
+
+// Reconcile performs a full reconciliation for the object referred to by the Request.
+// The SyncController will requeue the Request to be processed again if an error is non-nil or
+// Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
+func (c *QuotaEnforcementController) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	klog.V(4).Infof("QuotaEnforcementController reconciling %s", req.NamespacedName.String())
+
+	// First try fetching FederatedResourceQuota
+	quota := &policyv1alpha1.FederatedResourceQuota{}
+	if err := c.Get(ctx, req.NamespacedName, quota); err == nil {
+		if !quota.DeletionTimestamp.IsZero() {
+			return ctrl.Result{}, nil
+		}
+
+		if err := c.collectQuotaStatus(quota); err != nil {
+			klog.Errorf("Failed to collect status for FederatedResourceQuota(%s), error: %v", req.NamespacedName.String(), err)
+			c.EventRecorder.Eventf(quota, corev1.EventTypeWarning, events.EventReasonCollectFederatedResourceQuotaStatusFailed, err.Error())
+			return ctrl.Result{}, err
+		}
+	} else if apierrors.IsNotFound(err) {
+		// The resource may no longer exist, in which case we stop processing.
+		return ctrl.Result{}, nil
+	}
+
+	// If not FederatedResourceQuota, check if its a deleted ResourceBinding
+	binding := &workv1alpha2.ResourceBinding{}
+	if err := c.Get(ctx, req.NamespacedName, binding); err != nil {
+		if apierrors.IsNotFound(err) {
+			c.collectQuotaStatuses(ctx, binding.Namespace)
+		} else {
+			return ctrl.Result{}, err
+		}
+	}
+
+	c.EventRecorder.Eventf(quota, corev1.EventTypeNormal, events.EventReasonCollectFederatedResourceQuotaStatusSucceed, "Collect status of FederatedResourceQuota(%s) succeed.", req.NamespacedName.String())
+	return ctrl.Result{}, nil
+}
+
+// SetupWithManager creates a controller and register to controller manager.
+func (c *QuotaEnforcementController) SetupWithManager(mgr ctrl.Manager) error {
+	fn := handler.MapFunc(
+		func(ctx context.Context, obj client.Object) []reconcile.Request {
+			var quotaList policyv1alpha1.FederatedResourceQuotaList
+			if err := c.Client.List(ctx, &quotaList, client.InNamespace(obj.GetNamespace())); err != nil {
+				return nil
+			}
+
+			if len(quotaList.Items) == 0 {
+				return nil // no quotas in namespace, skip reconcile
+			}
+
+			return []reconcile.Request{{
+				NamespacedName: types.NamespacedName{
+					Namespace: obj.GetNamespace(),
+					Name:      obj.GetName(),
+				},
+			}}
+		},
+	)
+
+	resourceBindingPredicate := builder.WithPredicates(predicate.Funcs{
+		CreateFunc: func(createEvent event.CreateEvent) bool {
+			return false // ignore creates
+		},
+		UpdateFunc: func(updateEvent event.UpdateEvent) bool {
+			return false // ignore updates
+		},
+		DeleteFunc: func(deleteEvent event.DeleteEvent) bool {
+			return true // only care about RB deletions
+		},
+		GenericFunc: func(event.GenericEvent) bool {
+			return false
+		},
+	})
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&policyv1alpha1.FederatedResourceQuota{}).
+		Watches(&workv1alpha2.ResourceBinding{}, handler.EnqueueRequestsFromMapFunc(fn), resourceBindingPredicate).
+		Complete(c)
+}
+
+func (c *QuotaEnforcementController) collectQuotaStatuses(ctx context.Context, namespace string) error {
+	var quotaList policyv1alpha1.FederatedResourceQuotaList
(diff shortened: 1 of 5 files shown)


pkg/controllers/federatedresourcequota/federated_resource_quota_enforcement_controller.go
```diff
@@ -0,0 +1,215 @@
...
+
+// Reconcile performs a full reconciliation for the object referred to by the Request.
+// The SyncController will requeue the Request to be processed again if an error is non-nil or
+// Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
+func (c *QuotaEnforcementController) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	klog.V(4).Infof("QuotaEnforcementController reconciling %s", req.NamespacedName.String())
+
+	// First try fetching FederatedResourceQuota
+	quota := &policyv1alpha1.FederatedResourceQuota{}
+	if err := c.Get(ctx, req.NamespacedName, quota); err == nil {
```
zhzhuang-zju, line 68:
```suggestion
	if err := c.Get(ctx, req.NamespacedName, quota); err != nil {
		if apierrors.IsNotFound(err) {
			return controllerruntime.Result{}, nil
		}

		return controllerruntime.Result{}, err
	}
```

Just a suggestion for better readability

pkg/controllers/federatedresourcequota/federated_resource_quota_enforcement_controller.go
```diff
@@ -0,0 +1,215 @@
...
+
+			if len(quotaList.Items) == 0 {
+				return nil // no quotas in namespace, skip reconcile
+			}
+
+			return []reconcile.Request{{
+				NamespacedName: types.NamespacedName{
+					Namespace: obj.GetNamespace(),
+					Name:      obj.GetName(),
+				},
```
zhzhuang-zju, line 116: When a deletion event of ResourceBinding is detected, it's possible to return a QuotaList instead of the ResourceBinding itself, allowing the reconcile logic to focus only on FederatedResourceQuota.

Here is an example:
https://github.com/karmada-io/karmada/blob/e2f0e6e9f701f430dcb47811ff26279ec9fa9817/pkg/controllers/federatedresourcequota/federated_resource_quota_sync_controller.go#L104-L131

mszacillo: Oh nice! I'll add this in. Thanks

mszacillo: Thought about this more - I think we do want to make a distinction between ResourceBinding and FederatedResourceQuota reconciles. 

When reconciling the quota, we will only update the `Overall` field of the quota. When reconciling a ResourceBinding deletion, we will update both `Overall` and `OverallUsed`. That way the validation webhook is entirely responsible for updating the used resources, and there is a separation of concerns.

RainbowMango (commented):
/assign

pkg/controllers/federatedresourcequota/federated_resource_quota_enforcement_controller.go
zhzhuang-zju: > When reconciling the quota, we will only update the `Overall` field of the quota. When reconciling a ResourceBinding deletion, we will update both Overall and OverallUsed.

It's not entirely the case; when the quota's hard limits change, it will also be necessary to update the OverallUsed.

pkg/controllers/federatedresourcequota/federated_resource_quota_enforcement_controller.go
```diff
@@ -0,0 +1,219 @@
...
+		if err := c.Get(ctx, req.NamespacedName, binding); err != nil {
+			if apierrors.IsNotFound(err) {
+				err := c.collectQuotaStatuses(ctx, binding.Namespace)
+				if err != nil {
+					return controllerruntime.Result{}, err
+				}
+			} else {
+				return controllerruntime.Result{}, err
+			}
+		}
```
RainbowMango, line 86: Maybe you misunderstand the way how the controller works.

The object key passed in `Reconcile()` function can only reference to `FederatedResourceQuota`, that is decided by `For(&policyv1alpha1.FederatedResourceQuota{})` when building this controller.

pkg/controllers/federatedresourcequota/federated_resource_quota_enforcement_controller.go
```diff
@@ -0,0 +1,219 @@
...
+		},
+		DeleteFunc: func(_ event.DeleteEvent) bool {
+			return true // only care about RB deletions
+		},
+		GenericFunc: func(event.GenericEvent) bool {
+			return false
+		},
+	})
+	return controllerruntime.NewControllerManagedBy(mgr).
+		For(&policyv1alpha1.FederatedResourceQuota{}).
```
zhzhuang-zju, line 139: we can minimize the reconcile frequency by introducing an event handler that only responds to:
- `create event`
- `update event`: only when `quota.spec.overall` changed

The above event handler is provided for reference only

pkg/controllers/federatedresourcequota/federated_resource_quota_enforcement_controller.go
```diff
@@ -0,0 +1,219 @@
...
+		}
+		quotaStatus.OverallUsed = calculateUsedWithResourceBinding(bindingList.Items)
+	}
+
+	if reflect.DeepEqual(quota.Status, *quotaStatus) {
+		klog.V(4).Infof("New quotaStatus is equal with old federatedResourceQuota(%s) status, no update required.", klog.KObj(quota).String())
+		return nil
+	}
+
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
zhzhuang-zju, line 180: When an update fails due to a conflict, it should not be retried; instead, a new reconciliation should be triggered, to avoid overwriting the changes made by the validation webhook.

pkg/controllers/federatedresourcequota/federated_resource_quota_enforcement_controller.go
```diff
@@ -0,0 +1,219 @@
...
+		err := c.collectQuotaStatus(&quota, true)
+		if err != nil {
+			fmt.Printf("failed to collect status for FederatedResourceQuota(%s), error: %v", klog.KObj(&quota).String(), err)
+		}
+	}
+
+	return nil
+}
+
+func (c *QuotaEnforcementController) collectQuotaStatus(quota *policyv1alpha1.FederatedResourceQuota, fullUpdate bool) error {
```
zhzhuang-zju, line 160: Refer to https://github.com/karmada-io/karmada/pull/6367#discussion_r2088331745, no need to introduce parameter `fullUpdate` to distinguish between quota and resourceBinding.
```suggestion
func (c *QuotaEnforcementController) collectQuotaStatus(quota *policyv1alpha1.FederatedResourceQuota) error {
```

pkg/controllers/federatedresourcequota/federated_resource_quota_enforcement_controller.go
```diff
@@ -0,0 +1,219 @@
...
+		})
+		return err
+	})
+}
+
+func calculateUsedWithResourceBinding(resourceBindings []workv1alpha2.ResourceBinding) corev1.ResourceList {
+	overallUsed := corev1.ResourceList{}
+	for _, binding := range resourceBindings {
+		// TODO: Consider skipping resourcebinding if it is not applied
+		if binding.Spec.ReplicaRequirements == nil {
```
zhzhuang-zju, line 193:
```suggestion
		if binding.Spec.ReplicaRequirements == nil || binding.Spec.Clusters == nil {
```

pkg/controllers/federatedresourcequota/federated_resource_quota_enforcement_controller.go
```diff
@@ -0,0 +1,219 @@
...
+		replicas := binding.Spec.Replicas
+
+		for name, quantity := range used {
+			// Update quantity to reflect number of replicas in resource
+			q := quantity.DeepCopy()
+			q.Mul(int64(replicas))
+
+			// Update quantity to reflect the number of clusters resource is duplicated to
+			if binding.Spec.Placement != nil && binding.Spec.Placement.ReplicaSchedulingType() == policyv1alpha1.ReplicaSchedulingTypeDuplicated {
+				q.Mul(int64(len(binding.Spec.Clusters)))
```
zhzhuang-zju, line 206: Is it possible to calculate `overallUsed` using only `spec.cluster` and `spec.ReplicaRequirements`? One advantage of doing so would be reducing the number of fields that need to be considered.

zhzhuang-zju (commented):
Thanks for your great work! @mszacillo 
My opinion is for reference only; open to discussion 😄

pkg/controllers/federatedresourcequota/federated_resource_quota_enforcement_controller.go
mszacillo: Over complicated things here, I can fix to make it more idiomatic to the way controllers work. Thanks :)

pkg/controllers/federatedresourcequota/federated_resource_quota_enforcement_controller.go
mszacillo: I could do that, but I'd need to loop through the clusters and count the number of total replicas across clusters. I think I prefer just multiplying by the replica count and cluster count.

pkg/controllers/federatedresourcequota/federated_resource_quota_enforcement_controller.go
```diff
@@ -0,0 +1,209 @@
...
+	// TODO: Consider adding filtering step to ResourceBinding list once scope is added to the quota
+	bindingList, err := helper.GetResourceBindingsByNamespace(c.Client, quota.Namespace)
+	if err != nil {
+		klog.Errorf("Failed to list resourcebindings tracked by FederatedResourceQuota(%s), error: %v", klog.KObj(quota).String(), err)
+		return err
+	}
+
+	quotaStatus := quota.Status.DeepCopy()
+	quotaStatus.Overall = quota.Spec.Overall
+	quotaStatus.OverallUsed = calculateUsedWithResourceBinding(bindingList.Items)
```
zhzhuang-zju, line 165: The result returned by `calculateUsedWithResourceBinding` needs to be filtered before it can be used for `overallUsed`. For example, if the 'overall' only declares CPU, then 'overallUsed' should contain only CPU.

pkg/controllers/federatedresourcequota/federated_resource_quota_enforcement_controller.go
```diff
@@ -0,0 +1,209 @@
...
+
+// QuotaEnforcementController reflects the status of the namespaced FederatedResourceQuotas.
+// The controller will reconcile in two cases:
+//  1. When a FederatedResourceQuota is CREATED or UPDATED
+//  2. When a ResourceBinding is DELETED
+//
+// When reconciling the controller will update Overall and OverallUsed, the later of which is only
+// updated in response to ResourceBinding deletions. OverallUsed is primarily populated by the RB validation webhook
+type QuotaEnforcementController struct {
+	client.Client // used to operate Work resources.
```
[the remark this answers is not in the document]
RainbowMango: -1
This is the client of Karmada APIServer, actually, it is used to operate both FederatedResourceQuota and ResourceBinding in this controller.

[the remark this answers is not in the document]
zhzhuang-zju: > it is used to operate both FederatedResourceQuota and ResourceBinding in this controller.

OK, this controller will use the client to list RB. 
```suggestion
	client.Client // used to operate FederatedResourceQuota and ResourceBinding resources.
```

pkg/controllers/federatedresourcequota/federated_resource_quota_enforcement_controller.go
```diff
@@ -0,0 +1,209 @@
...
+	if err := c.collectQuotaStatus(quota); err != nil {
+		klog.Errorf("Failed to collect status for FederatedResourceQuota(%s), error: %v", req.NamespacedName.String(), err)
+		c.EventRecorder.Eventf(quota, corev1.EventTypeWarning, events.EventReasonCollectFederatedResourceQuotaStatusFailed, err.Error())
+		if apierrors.IsConflict(err) {
+			return controllerruntime.Result{Requeue: true}, nil
+		}
+		return controllerruntime.Result{}, err
+	}
+
+	c.EventRecorder.Eventf(quota, corev1.EventTypeNormal, events.EventReasonCollectFederatedResourceQuotaStatusSucceed, "Collect status of FederatedResourceQuota(%s) succeed.", req.NamespacedName.String())
```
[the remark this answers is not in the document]
RainbowMango: Yes, introducing another pair of events would be better.

RainbowMango (commented):
Looks great!

pkg/controllers/federatedresourcequota/federated_resource_quota_enforcement_controller.go
zhzhuang-zju: Also, Karmada only supports resource requests and not limits. Could we consider enhancing the `ResourceInterpreter GetReplicas` by adding a `ResourceLimit` return value?

https://github.com/karmada-io/karmada/blob/94a9f8bdbb1e24f6533305920c4fedc131a958b9/pkg/apis/work/v1alpha2/binding_types.go#L196-L213

WDYT? @mszacillo @RainbowMango

pkg/controllers/federatedresourcequota/federated_resource_quota_enforcement_controller.go
zhzhuang-zju: Thanks for the explanation, make sense to me~

pkg/controllers/federatedresourcequota/federated_resource_quota_enforcement_controller.go
mszacillo: I can record this as an iteration task of part II - or we can decide to split this off into a separate effort. Either way I'll list it there for now.

RainbowMango (commented):
/lgtm

@zhzhuang-zju Do you have any further comments?

zhzhuang-zju:
> /lgtm
> 
> @zhzhuang-zju Do you have any further comments?

/lgtm

RainbowMango (approved):
/approve

Issue #6317, opened 2025-04-23
[Umbrella] Optimize Cluster Failover


XiShanYongYe-Chang:
**Background**

I put forward a proposal to optimize Cluster Failover in #6274. Now, according to the content designed in the proposal, the content that needs to be modified is divided as follows:

- [ ] 0. Proposal #6274
- [ ] 1. Judgment of Cluster Failure Conditions
  - [ ] Add the ClusterTaintPolicy API (1.14)
  - [ ] Add a new controller to monitor changes in the cluster status conditions and decide how to modify the taints of the Cluster object according to the user-defined ClusterTaintPolicy (1.14)
  - [ ] UT & E2E (1.14)
  - [ ] Introduction in the functional documentation (1.14)
- [ ] 2. Design of Taint Types and Behaviors
  - [ ]  Modify the controller logic so that Karmada no longer actively adds taints to the Cluster (1.14)
  - [ ]  Modify the comment descriptions of the NoSchedule and NoExecute taint effects to explain their effective situations (1.14)
  - [ ]  Introduce the new taint type: taint{effect: PreferNoExecute} (1.14)
- [ ] 3. Application-level Cluster Failure Migration Strategy
  - [ ]  Introduce the API of the application-level active cluster failure migration strategy in the PropagationPolicy / Cluster Propagation Policy (1.14)
  - [ ]  Modify the controller to pass this strategy through to the ResourceBinding  / ClusterResourceBinding objects (1.14)
  - [ ]  Modify the API definition of.spec.placement.clusterTolerations in the PP/CPP to explain its effective scope (1.14)
  - [ ]  Introduction in the functional documentation (1.14)
- [ ] 4. Eviction Queue
  - [ ]  Determine the resources affected by the cluster failure according to the taint type and add them to the eviction queue (1.14)
  - [ ]  Configure multiple rate levels for the eviction queue according to the overall health status of the clusters in Karmada (1.15)
  - [ ]  Exclude as many situations where scheduling fails as possible, and the dequeued elements will be abandoned for processing (1.14)
  - [ ]  Add metrics to the eviction queue (1.15)
  - [ ]  Handle the scenario of the failed cluster recovery (1.14)
- [ ] 5. Feature Switches
  - [ ]  Adjust the control scope of the Failover FeatureGate to control the start and stop of the entire Cluster Failover function (1.14)
  - [ ]  Deprecate the GracefulEviction FeatureGate (1.14)
[...]

PR #6371, opened 2025-05-14
Avoid reconcile conflict in cluster controller

zhzhuang-zju:
**What type of PR is this?**
/kind cleanup

The cluster controller updates the Cluster object twice during a single reconciliation.
- taint cluster by condition
- add/remove finalizer

The second update will fail due to a conflict.
```bash
Controller.Reconcile() error = Operation cannot be fulfilled on clusters.cluster.karmada.io "test-cluster": object was modified
```

Given that additional operations such as creating or deleting ExecutionSpace take place between the two updates, and considering the significant impact of taints on scheduling and related features, timely updates are crucial. As a result, this fix chooses to replace UPDATE operations with PATCH to prevent conflicts, instead of merging the two updates into a single operation at the end of the reconcile loop.

Fixes #

```release-note

```

### DIFF
--- a/pkg/controllers/cluster/cluster_controller.go
+++ b/pkg/controllers/cluster/cluster_controller.go
@@ -244,7 +244,7 @@ func (c *Controller) syncCluster(ctx context.Context, cluster *clusterv1alpha1.C
 
 func (c *Controller) removeCluster(ctx context.Context, cluster *clusterv1alpha1.Cluster) (controllerruntime.Result, error) {
 	// add terminating taint before cluster is deleted
-	if err := c.updateClusterTaints(ctx, []*corev1.Taint{TerminatingTaintTemplate}, nil, cluster); err != nil {
+	if err := c.patchClusterTaints(ctx, []*corev1.Taint{TerminatingTaintTemplate}, nil, cluster); err != nil {
 		klog.ErrorS(err, "Failed to update terminating taint", "cluster", cluster.Name)
 		return controllerruntime.Result{}, err
 	}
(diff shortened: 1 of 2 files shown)


RainbowMango:
> The cluster controller updates the Cluster object twice during a single reconciliation.
taint cluster by condition
add/remove finalizer

The cluster controller will not taint the cluster after #6317.
Please @XiShanYongYe-Chang confirm if this is still needed after that.

XiShanYongYe-Chang:
Ok, let me take a look.
/assign

pkg/controllers/cluster/cluster_controller.go
```diff
@@ -244,7 +244,7 @@ func (c *Controller) syncCluster(ctx context.Context, cluster *clusterv1alpha1.C
 
 func (c *Controller) removeCluster(ctx context.Context, cluster *clusterv1alpha1.Cluster) (controllerruntime.Result, error) {
 	// add terminating taint before cluster is deleted
-	if err := c.updateClusterTaints(ctx, []*corev1.Taint{TerminatingTaintTemplate}, nil, cluster); err != nil {
+	if err := c.patchClusterTaints(ctx, []*corev1.Taint{TerminatingTaintTemplate}, nil, cluster); err != nil {
```
XiShanYongYe-Chang, line 247: Hi @zhzhuang-zju, thank you for doing this. 

I plan to delete the taint in #6317 [The system will no longer automatically add taints to clusters].  Therefore, the taint can be modified and the finalizer processing can be retained.

zhzhuang-zju: > Therefore, the taint can be modified and the finalizer processing can be retained.

Thank you for the explanation. It seems that this issue will no longer exist on the master branch. I will close this PR, and if there is a need to address this issue on a maintenance branch, I would be happy to contribute.

zhzhuang-zju:
ref to https://github.com/karmada-io/karmada/pull/6371#discussion_r2094749026

/closed

PR #6396, opened 2025-05-22
Update cluster-taint-policy-controller to reconcile ClusterTaintPolicy

XiShanYongYe-Chang:
**What type of PR is this?**

/kind feature

In order to make the logic of `cluster-taint-policy-controller` more smoothly, update cluster-taint-policy-controller to reconcile ClusterTaintPolicy

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/taint/clustertaintpolicy_controller.go
+++ b/pkg/controllers/taint/clustertaintpolicy_controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"sort"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


RainbowMango (commented):
/assign

pkg/controllers/taint/clustertaintpolicy_controller.go
```diff
@@ -217,30 +232,35 @@ func (c *ClusterTaintPolicyController) SetupWithManager(mgr controllerruntime.Ma
...
+			if err := c.Client.List(ctx, clusterTaintPolicyList, listOption); err != nil {
+				klog.Errorf("Failed to list ClusterTaintPolicy, error: %v", err)
 				return nil
 			}
 
-			policy := policyObj.(*policyv1alpha1.ClusterTaintPolicy)
+			policies := clusterTaintPolicyList.Items
+			sort.Slice(policies, func(i, j int) bool {
+				return policies[i].CreationTimestamp.Before(&policies[j].CreationTimestamp)
+			})
```
RainbowMango, line 247: Sorting the policy here won't help as these policies are not proceed as this order, especially in concurrent cases.

pkg/controllers/taint/clustertaintpolicy_controller.go
```diff
@@ -57,28 +58,30 @@ type ClusterTaintPolicyController struct {
...
 func (c *ClusterTaintPolicyController) Reconcile(ctx context.Context, req controllerruntime.Request) (controllerruntime.Result, error) {
-	klog.V(4).Infof("Reconciling Cluster %s.", req.Name)
-
-	clusterObj := &clusterv1alpha1.Cluster{}
-	if err := c.Client.Get(ctx, req.NamespacedName, clusterObj); err != nil {
-		klog.Errorf("Failed to get Cluster(%s), error: %v", req.Name, err)
+	klog.V(4).Infof("Reconciling ClusterTaintPolicy %s.", req.Name)
+	policy := &policyv1alpha1.ClusterTaintPolicy{}
+	if err := c.Client.Get(ctx, req.NamespacedName, policy); err != nil {
+		klog.Errorf("Failed to get ClusterTaintPolicy(%s), error: %v", req.Name, err)
```
RainbowMango, line 64: This pattern will log a useless log if the policy `NotFound`.

pkg/controllers/taint/clustertaintpolicy_controller.go
```diff
@@ -57,28 +58,30 @@ type ClusterTaintPolicyController struct {
...
+		klog.Errorf("Failed to list Cluster, error: %v", err)
 		return controllerruntime.Result{}, err
 	}
 
-	clusterCopyObj := clusterObj.DeepCopy()
-	for _, policy := range clusterTaintPolicyList.Items {
+	var errs []error
+	for index := range clusterList.Items {
+		clusterObj := &clusterList.Items[index]
+		clusterCopyObj := clusterObj.DeepCopy()
```
RainbowMango, line 83: Maybe we can skip the deep copy here and at the same time not use the `UnsafeDisableDeepCopy` option when listing the Clusters above.

pkg/controllers/taint/clustertaintpolicy_controller.go
XiShanYongYe-Chang: In this way, when we use ClusterTaintPolicy as the reconcile object, maybe we cannot guarantee the idempotent of the execution result.

XiShanYongYe-Chang:
It's not appropriate to continue pushing the pr now.
/close

PR #6403, opened 2025-05-24
No longer add terminting NoExecute taint for the Deleting Cluster

XiShanYongYe-Chang:
**What type of PR is this?**

/kind feature

No longer add the NoExecute taint for the Deleting Cluster.

Part of #6317 

```release-note
`karmada-controller`: The `cluster-controller` no longer relies on taints to clean up resources on clusters scheduled for deletion, but instead depends on the scheduler to reschedule workloads to other available clusters.
```

### DIFF
--- a/pkg/controllers/cluster/cluster_controller.go
+++ b/pkg/controllers/cluster/cluster_controller.go
@@ -75,20 +75,12 @@ var (
 		Key:    clusterv1alpha1.TaintClusterNotReady,
 		Effect: corev1.TaintEffectNoSchedule,
 	}
-
-	// TerminatingTaintTemplate is the taint for when a cluster is terminating executing resources.
-	// Used for taint based eviction.
-	TerminatingTaintTemplate = &corev1.Taint{
-		Key:    clusterv1alpha1.TaintClusterTerminating,
-		Effect: corev1.TaintEffectNoExecute,
-	}
 )
 
 // Controller is to sync Cluster.
 type Controller struct {
-	client.Client      // used to operate Cluster resources.
-	EventRecorder      record.EventRecorder
-	EnableTaintManager bool
+	client.Client // used to operate Cluster resources.
+	EventRecorder record.EventRecorder
 
 	// ClusterMonitorPeriod represents cluster-controller monitoring period, i.e. how often does
 	// cluster-controller check cluster health signal posted from cluster-status-controller.
@@ -100,9 +92,10 @@ type Controller struct {
 	ClusterMonitorGracePeriod time.Duration
 	// When cluster is just created, e.g. agent bootstrap or cluster join, we give a longer grace period.
 	ClusterStartupGracePeriod time.Duration
-
-	ClusterTaintEvictionRetryFrequency time.Duration
-	ExecutionSpaceRetryFrequency       time.Duration
+	// CleanupCheckInterval defines the fixed interval for polling resource deletion status during cluster removal.
+	// The fixed interval bypasses exponential backoff mechanism to ensure the check frequency remains balanced
+	// - neither too frequent to risk system overload nor too sparse to cause delays.
+	CleanupCheckInterval time.Duration
 
 	// Per Cluster map stores last observed health together with a local time when it was observed.
 	clusterHealthMap   *clusterHealthMap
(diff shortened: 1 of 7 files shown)


pkg/controllers/cluster/cluster_controller.go
```diff
@@ -101,8 +93,8 @@ type Controller struct {
 	// When cluster is just created, e.g. agent bootstrap or cluster join, we give a longer grace period.
 	ClusterStartupGracePeriod time.Duration
 
-	ClusterTaintEvictionRetryFrequency time.Duration
-	ExecutionSpaceRetryFrequency       time.Duration
+	ResourceCleanupWaitingRetryFrequency      time.Duration
+	ExecutionSpaceRemoveWaitingRetryFrequency time.Duration
```
RainbowMango, line 97: Given the two are both for defining the interval to cleanup during cluster termination.
How about merging them into a single config?
Like
```golang
	// CleanupCheckInterval defines the fixed interval for polling resource deletion status
	// during cluster removal.
	// The fixed interval bypasses exponential backoff mechanisms to ensure the check frequency remains balanced
	// - neither too frequent to risk system overload nor too sparse to cause delays.
	CleanupCheckInterval time.Duration
```

pkg/controllers/cluster/cluster_controller.go
```diff
@@ -244,21 +230,19 @@ func (c *Controller) removeCluster(ctx context.Context, cluster *clusterv1alpha1
...
-			return controllerruntime.Result{}, err
-		} else if !done {
-			klog.InfoS("Terminating taint eviction process has not finished yet, will try again later", "cluster", cluster.Name)
-			return controllerruntime.Result{RequeueAfter: c.ClusterTaintEvictionRetryFrequency}, nil
-		}
+	if done, err := c.isTargetClusterRemoved(ctx, cluster); err != nil {
+		klog.ErrorS(err, "Failed to check whether target cluster is removed from bindings", "cluster", cluster.Name)
+		return controllerruntime.Result{}, err
+	} else if !done {
+		klog.InfoS("Terminating taint eviction process has not finished yet, will try again later", "cluster", cluster.Name)
```
RainbowMango, line 244: Should twist the logs as here does not rely on the eviction process anymore.

pkg/controllers/cluster/cluster_controller.go
XiShanYongYe-Chang: It sounds good!

pkg/controllers/cluster/cluster_controller.go
XiShanYongYe-Chang: Updated.

pkg/scheduler/scheduler.go
```diff
@@ -512,6 +525,25 @@ func (s *Scheduler) doScheduleClusterBinding(name string) (err error) {
 	return nil
 }
 
+func (s *Scheduler) someScheduledClusterDeleted(bindingSpec *workv1alpha2.ResourceBindingSpec) bool {
```
RainbowMango, line 528:
```suggestion
// HasTerminatingTargetClusters checks whether any cluster in the ResourceBinding's target list
// is marked for deletion (i.e., has a non-zero DeletionTimestamp).
//
// This is used to trigger rescheduling when bound clusters are being terminated, ensuring
// workloads get migrated before cluster resources become unavailable.
func (s *Scheduler) HasTerminatingTargetClusters(bindingSpec *workv1alpha2.ResourceBindingSpec) bool {
```

pkg/scheduler/scheduler.go
```diff
@@ -429,6 +430,12 @@ func (s *Scheduler) doScheduleBinding(namespace, name string) (err error) {
 		return err
 	}
 	// TODO(dddddai): reschedule bindings on cluster change
```
RainbowMango, line 432: We can remove the TODO now I guess.

pkg/scheduler/scheduler.go
```diff
@@ -499,6 +506,12 @@ func (s *Scheduler) doScheduleClusterBinding(name string) (err error) {
...
```
RainbowMango, line 508: todo
ditto

RainbowMango (commented):
Generally looks good except some nits.

1. Please remember to add a new E2E test to cover ClusterResourceBinding be take care during cluster termination
2. Please consider another situation `what if the cluster be removed forcelly, I mean without terminating?`

RainbowMango:
/title No longer relying on eviction to cleanup resources for clusters in terminating

PR #6404, opened 2025-05-24
Add failover flags to control the behavior of NoExecute

XiShanYongYe-Chang:
**What type of PR is this?**

/kind feature

NoExecute is a passive migration mode, which is dangerous. We add a controller and disable it by default.

Fixes #

```release-note
karmada-controller-manager: add failover flags to control the behavior of NoExecute.
```

### DIFF
--- a/cmd/controller-manager/app/options/failover.go
+++ b/cmd/controller-manager/app/options/failover.go
@@ -1,0 +1,56 @@
+/*
+Copyright 2025 The Karmada Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package options
+
+import (
+	"github.com/spf13/pflag"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+// FailoverOptions holds the Failover configurations.
+type FailoverOptions struct {
+	// EnablePassiveMigration determines if the controller should respect NoExecute taints
+	// and enable passive pod migration.
+	EnablePassiveMigration bool
+	// NoExecutePurgeMode defines the behavior of pod eviction during passive migration.
+	// Valid modes:
+	// - "gracefully": the system will wait for workload become healthy on the new cluster.
+	// - "immediately": the system will immediately evict the legacy workload.
+	// Default: "gracefully".
+	PassiveMigrationMode string
+}
+
+// NewFailoverOptions creates a new FailoverOptions object with default parameters.
+func NewFailoverOptions() *FailoverOptions {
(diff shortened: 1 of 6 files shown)


pkg/controllers/cluster/taint_manager.go
```diff
@@ -171,19 +174,29 @@ func (tc *NoExecuteTaintManager) syncBindingEviction(key util.QueueKey) error {
 		return err
 	}
 
+	var purgeMode policyv1alpha1.PurgeMode
```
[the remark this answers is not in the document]
XiShanYongYe-Chang: The controller parameter value has been verified during failoverOptions verification. No empty value will be displayed.

cmd/controller-manager/app/options/failover.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// AddFlags adds flags related to FailoverOptions for controller manager to the specified FlagSet.
+func (o *FailoverOptions) AddFlags(flags *pflag.FlagSet) {
+	if o == nil {
+		return
+	}
+
+	flags.BoolVar(&o.SupportNoExecute, "support-no-execute", false, "It determines if the controller should respect NoExecute taints.")
+	flags.StringVar(&o.NoExecutePurgeMode, "no-execute-purge-mode", "gracefully", "It defines the behavior of workload eviction during migration triggered by NoExecute taint. Valid modes: \"gracefully\" and \"immediately\".")
```
RainbowMango, line 45:
```
--enable-noexecute-taint-eviction bool
Enables controller response to NoExecute taints on clusters, which triggers immediate eviction of workloads without explicit tolerations. Given the high risk of eviction, this parameter is designed to remain disabled by default and requires careful evaluation by administrators before being enabled.

--noexecute-taint-eviction-purge-mode String
Controls resource cleanup behavior for NoExecute-triggered evictions (only active when --enable-noexecute-taint-eviction=true). Supported values are "Directly", and "Gracefully". "Directly" mode directly evicts workloads first (risking temporary service interruption) and then triggers rescheduling to other clusters, while "Gracefully" mode first schedules workloads to new clusters and then cleans up original workloads after successful startup elsewhere to ensure service continuity.

```

cmd/controller-manager/app/options/failover.go
```diff
@@ -0,0 +1,61 @@
...
+	NoExecuteTaintEvictionPurgeMode string
+}
+
+// AddFlags adds flags related to FailoverOptions for controller manager to the specified FlagSet.
+func (o *FailoverOptions) AddFlags(flags *pflag.FlagSet) {
+	if o == nil {
+		return
+	}
+
+	flags.BoolVar(&o.EnableNoExecuteTaintEviction, "enable-no-execute-taint-eviction", false, "Enables controller response to NoExecute taints on clusters, which triggers immediate eviction of workloads without explicit tolerations. Given the high risk of eviction, this parameter is designed to remain disabled by default and requires careful evaluation by administrators before being enabled.\n")
```
kevin-wangzefeng, line 47:
```suggestion
	flags.BoolVar(&o.EnableNoExecuteTaintEviction, "enable-no-execute-taint-eviction", false, "Enables controller response to NoExecute taints on clusters, which triggers eviction of workloads without explicit tolerations. Given the impact of eviction caused by NoExecute Taint, this parameter is designed to remain disabled by default and requires careful evaluation by administrators before being enabled.\n")
```

PR #6407, opened 2025-05-26
WIP periodically reconcile FederatedResourceQuota to update the usage state

zhzhuang-zju:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #6411, opened 2025-05-27
Skip reconcile ResourceBindings if it has not been scheduled

RainbowMango:
**What type of PR is this?**

/kind cleanup

It is unnecessary to sync ResourceBinding/ClusterResourceBinding status if it has not been scheduled. 
This also reduces the conflict risk as mentioned at https://github.com/karmada-io/karmada/pull/6407#issuecomment-2911513234

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/status/common.go
+++ b/pkg/controllers/status/common.go
@@ -43,14 +43,50 @@ import (
 )
 
 var bindingPredicateFn = builder.WithPredicates(predicate.Funcs{
-	CreateFunc: func(event.CreateEvent) bool {
+	CreateFunc: func(e event.CreateEvent) bool {
 		// Although we don't need to process the ResourceBinding immediately upon its creation,
 		// but it's necessary to ensure that all existing ResourceBindings are processed
 		// uniformly once when the component restarts.
 		// This guarantees that no ResourceBinding is missed after a controller restart.
-		return true
+
+		// Ignore the ResourceBinding/ClusterResourceBinding if it has not been scheduled.
+		// Relies on status.lastScheduledTime as the reliable indicator of successful scheduling because
+		// it is only set by the scheduler after successful scheduling.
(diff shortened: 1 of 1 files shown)


zhzhuang-zju:
> Reproduced the issue 5 times out of 10 test runs(https://github.com/karmada-io/karmada/pull/6407#issuecomment-2911513234).

In addition to the `rb_status_controller`, the detector also has areas for optimization. After optimization, 20 local executions were performed with 0 conflicts.

pkg/controllers/status/common.go
```diff
@@ -39,18 +39,48 @@ import (
...
+		return scheduled
 	},
 	UpdateFunc: func(e event.UpdateEvent) bool {
+		// Ignore the ResourceBinding/ClusterResourceBinding if it has not been scheduled.
+		// TODO(@RainbowMango): The Scheduled condition is not a reliable indicator of final scheduling status
+		// because it may change after the initial successful scheduling if subsequent scheduling attempts fail.
+		var scheduled bool
+		switch bindingObj := e.ObjectNew.(type) {
+		case *workv1alpha2.ResourceBinding:
+			scheduled = helper.IsBindingScheduled(&bindingObj.Status) && (len(bindingObj.Spec.Clusters) > 0)
```
zhzhuang-zju, line 74: @RainbowMango Is it feasible to determine whether the binding has been scheduled by only using the length of `bindingObj.Spec.Clusters`?

zhzhuang-zju: As your comment said, the Scheduled condition is not a reliable indicator. If the `spec.clusters` is not empty, it indicates that the resource has an expected cluster distribution status, and then the aggregation status of the resource can start to be collected.

zhzhuang-zju: Make sense~

zhzhuang-zju:
Thanks~
/lgtm

PR #6412, opened 2025-05-27
Add E2E for ClusterTaintPolicy controller

XiShanYongYe-Chang:
**What type of PR is this?**

/kind cleanup

part of #6317 

```release-note
NONE
```

### DIFF
--- a/test/e2e/suites/base/clustertaintpolicy_test.go
+++ b/test/e2e/suites/base/clustertaintpolicy_test.go
@@ -1,0 +1,270 @@
+/*
+Copyright 2025 The Karmada Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package base
+
+import (
+	"fmt"
+
+	"github.com/onsi/ginkgo/v2"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/rand"
+
+	clusterv1alpha1 "github.com/karmada-io/karmada/pkg/apis/cluster/v1alpha1"
+	policyv1alpha1 "github.com/karmada-io/karmada/pkg/apis/policy/v1alpha1"
+	"github.com/karmada-io/karmada/pkg/util/helper"
+	"github.com/karmada-io/karmada/test/e2e/framework"
+	karmadaresource "github.com/karmada-io/karmada/test/e2e/framework/resource/karmada"
+)
+
+// The following tests involve adding NoSchedule and NoExecute effect taints to clusters
+// in the test environment, which will impact the resource scheduling of other concurrently
+// executed test cases. Therefore, these tests need to be configured to run serially.
+var _ = framework.SerialDescribe("Taint cluster with ClusterTaintPolicy", func() {
+	var policyName string
+	var clusterTaintPolicy *policyv1alpha1.ClusterTaintPolicy
+	var targetClusterName string
+
+	ginkgo.JustBeforeEach(func() {
+		karmadaresource.CreateClusterTaintPolicy(karmadaClient, clusterTaintPolicy)
+		ginkgo.DeferCleanup(func() {
+			karmadaresource.RemoveClusterTaintPolicy(karmadaClient, policyName)
+		})
+	})
+
+	ginkgo.BeforeEach(func() {
+		policyName = clusterTaintPolicyNamePrefix + rand.String(RandomStrLength)
+		clusterTaintPolicy = &policyv1alpha1.ClusterTaintPolicy{
+			TypeMeta: metav1.TypeMeta{
+				APIVersion: policyv1alpha1.SchemeGroupVersion.String(),
+				Kind:       "ClusterTaintPolicy",
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Name: policyName,
+			},
+		}
+	})
+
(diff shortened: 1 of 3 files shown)


RainbowMango (commented):
/assign

test/e2e/suites/base/clustertaintpolicy_test.go
```diff
@@ -0,0 +1,356 @@
...
+						{
+							Key:    "testing/not-ready",
+							Effect: "NoSchedule",
+						},
+					},
+				},
+			}
+		})
+
+		ginkgo.It("Trigger tainting cluster by making cluster not ready", func() {
```
RainbowMango, line 90: Do we alreay have the test check the condition in case the cluster becomes unreachable?
This case is testing another feature, for the cluster-taint-policy-controller it don't care how the conditions be added.

test/e2e/suites/base/clustertaintpolicy_test.go
```diff
@@ -0,0 +1,267 @@
...
+	"k8s.io/apimachinery/pkg/util/rand"
+
+	clusterv1alpha1 "github.com/karmada-io/karmada/pkg/apis/cluster/v1alpha1"
+	policyv1alpha1 "github.com/karmada-io/karmada/pkg/apis/policy/v1alpha1"
+	"github.com/karmada-io/karmada/pkg/util/helper"
+	"github.com/karmada-io/karmada/test/e2e/framework"
+	karmadaresource "github.com/karmada-io/karmada/test/e2e/framework/resource/karmada"
+)
+
+var _ = framework.SerialDescribe("Taint cluster with ClusterTaintPolicy", func() {
```
RainbowMango, line 37: How about adding some comments here to explain why this test suite needs to be run in serial?

test/e2e/suites/base/clustertaintpolicy_test.go
```diff
@@ -0,0 +1,267 @@
...
+		clusterTaintPolicy = &policyv1alpha1.ClusterTaintPolicy{
+			TypeMeta: metav1.TypeMeta{
+				APIVersion: policyv1alpha1.SchemeGroupVersion.String(),
+				Kind:       "ClusterTaintPolicy",
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Name: policyName,
+			},
+		}
+	})
```
RainbowMango, line 60: Just a question, both `BeforeEach` and `JustBeforeEach` are used to initialize the ClusterTaintPolicy, why separate them?

test/e2e/suites/base/clustertaintpolicy_test.go
```diff
@@ -0,0 +1,267 @@
...
+					{
+						ConditionType: "NetworkReady",
+						Operator:      policyv1alpha1.MatchConditionOpIn,
+						StatusValues:  []metav1.ConditionStatus{metav1.ConditionTrue},
+					},
+				},
+				Taints: []policyv1alpha1.Taint{
+					{
+						Key:    "testing/network-not-ready",
+						Effect: "NoSchedule",
```
RainbowMango, line 86: Is it mandatory to use a real `effect`? 
If we use a dummy effect here, this test can be run in parallel, right?

test/e2e/suites/base/clustertaintpolicy_test.go
```diff
@@ -0,0 +1,267 @@
...
+					},
+				},
+			}
+		})
+
+		ginkgo.It("one condition match", func() {
```
RainbowMango:
```suggestion
		ginkgo.It("taint should be added and removed upon single condition match", func() {
```

test/e2e/suites/base/clustertaintpolicy_test.go
```diff
@@ -0,0 +1,267 @@
...
+	ginkgo.Context("Control multi taints", func() {
+		ginkgo.BeforeEach(func() {
+			targetClusterName = framework.ClusterNames()[1]
+			clusterTaintPolicy.Spec = policyv1alpha1.ClusterTaintPolicySpec{
+				TargetClusters: &policyv1alpha1.ClusterAffinity{
+					ClusterNames: []string{targetClusterName},
+				},
+				AddOnConditions: []policyv1alpha1.MatchCondition{
+					{
+						ConditionType: "UnknownReady",
```
RainbowMango:
```suggestion
						ConditionType: "NetworkReady",
```
Use a meaningful condition name?

test/e2e/suites/base/clustertaintpolicy_test.go
```diff
@@ -0,0 +1,267 @@
...
+			ginkgo.By(fmt.Sprintf("wait for the cluster(%s) taint removed", targetClusterName), func() {
+				framework.WaitClusterFitWith(controlPlaneClient, targetClusterName, func(cluster *clusterv1alpha1.Cluster) bool {
+					return !helper.TaintExists(cluster.Spec.Taints, &corev1.Taint{Key: "testing/all-not-ready", Effect: "NoSchedule"})
+				})
+			})
+		})
+
+	})
+
+	ginkgo.Context("Control multi taints", func() {
```
RainbowMango:
```suggestion
	ginkgo.Context("multiple taints should be added and removed upon condition match", func() {
```

test/e2e/suites/base/clustertaintpolicy_test.go
XiShanYongYe-Chang: The execution sequence of JustBeforeEach is different from that of BeforeEach. BeforeEach is executed in the declared order. JustBeforeEach is executed before each It module is executed.

test/e2e/suites/base/clustertaintpolicy_test.go
XiShanYongYe-Chang: Yes, if we use a dummy effect taint, it can be executed in parallel. However, the value of effect will be checked later, and serialization is still required.

test/e2e/suites/base/clustertaintpolicy_test.go
XiShanYongYe-Chang: Of course, I added some comments.

test/e2e/suites/base/clustertaintpolicy_test.go
RainbowMango: got it.

RainbowMango (approved):
/lgtm
/approve

Issue #6414, opened 2025-05-28
APIServer compatibility  is failing with karmada-apiserver 1.30


liangyuanpeng:
**What happened**:

https://github.com/karmada-io/karmada/actions/runs/15241477746/job/42862086548

KEP-3998 brings JobSuccessPolicy, it means that the feature-gate is enabled, JobSuccessCriteriaMet status needs to be added when the job is completed .  

this feature-gate introduced as :
- alpha in 1.30 (false by default)
- beta in 1.31 (true by default)
- GA in 1.33 (lock to true )

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Karmada version:
- kubectl-karmada or karmadactl version (the result of `kubectl-karmada version` or `karmadactl version`):
- Others:

zhzhuang-zju:
> If set JobManagedBy to true, We will get the same error on karmada-apiserver 1.31:

@liangyuanpeng Sorry, I don't understand why setting `JobManagedBy` to true triggers Invalid value error. Can you explain it in more detail? Thanks

zhzhuang-zju:
https://github.com/kubernetes/kubernetes/blob/ce9ba818025ef28555bac9e25da917acec2ccb1d/pkg/registry/batch/job/strategy.go#L345-L380
```go
	if utilfeature.DefaultFeatureGate.Enabled(features.JobManagedBy) {
		// A strengthened validation of the Job status transitions is needed since the
		// Job managedBy field let's the Job object be controlled by external
		// controllers. We want to make sure the transitions done by the external
		// controllers meet the expectations of the clients of the Job API.
		// For example, we verify that a Job in terminal state (Failed or Complete)
		// does not flip to a non-terminal state.
```
The introduction of FeatureGate JobManagedBy has strengthened the validation of job status in Kubernetes. Since JobManagedBy defaults to true starting from v1.32.0, our APIServer compatibility testing began to fail at this version.

RainbowMango:
I'm going to move this from v1.15 to v1.16.

RainbowMango:
[...]

PR #6415, opened 2025-05-28
fix job policy

liangyuanpeng:
**What type of PR is this?**

/kind failing-test
releated:
- https://github.com/karmada-io/karmada/issues/6414

Fixes #

```release-note

```

### DIFF
--- a/.github/workflows/ci-schedule-compatibility.yaml
+++ b/.github/workflows/ci-schedule-compatibility.yaml
@@ -3,7 +3,7 @@ on:
   schedule:
     # Run this workflow "At 20:00 UTC on Sunday and Saturday"
     - cron: '0 20 * * 0,6'
-  pull:
+  pull_request:
 
 permissions:
   contents: read # Required by actions/checkout to fetch the repository contents.


pkg/resourceinterpreter/default/native/aggregatestatus.go
```diff
@@ -213,6 +213,25 @@ func aggregateJobStatus(object *unstructured.Unstructured, aggregatedStatusItems
...
+		if c.Type == batchv1.JobComplete {
+			if c.Status == corev1.ConditionTrue {
+				jobFinished = true
+				break
+			}
+		}
+	}
+	if jobFinished {
+		now := metav1.Now()
+		job.Status.Conditions = append(job.Status.Conditions, batchv1.JobCondition{
```
liangyuanpeng, line 227: This might not be the correct fix because when karmada-apiserver is 1.30.x, karmada-controller-manager will complain :  `status.conditions: Invalid value: cannot set SuccessCriteriaMet to NonIndexed Job 和 status.conditions: Invalid value: cannot set SuccessCriteriaMet=True for Job without SuccessPolicy，`  (the reason is the feature-gate JobSuccessPolicy is false )

RainbowMango:
/close
In favor of #6964

Issue #6230, opened 2025-03-26
Structured Logging for Karmada Components


jabellard:
What would you like to be added:

I would like the ability to configure structured logging for Karmada components (e.g., `karmada-controller-manager`) in the same manner as standard Kubernetes components. Specifically, this could be achieved via a command-line flag like `--logging-format=json`.

As an example, standard Kubernetes components such as `kube-apiserver` and `kube-controller-manager` can already emit JSON logs when passed `--logging-format=json`. For instance:

```json
{
  "ts":1742950806321.6917,
  "logger":"garbage-collector-controller",
  "caller":"garbagecollector/graph_builder.go:212",
  "msg":"using a shared informer",
  "v":4,
  "resource":"cluster.karmada.io/v1alpha1, Resource=clusters",
  "kind":"cluster.karmada.io/v1alpha1, Kind=Cluster"
}
{
  "ts":1742950806321.7192,
  "logger":"garbage-collector-controller",
  "caller":"garbagecollector/graph_builder.go:212",
  "msg":"using a shared informer",
  "v":4,
  "resource":"policy.karmada.io/v1alpha1, Resource=clusteroverridepolicies",
  "kind":"policy.karmada.io/v1alpha1, Kind=ClusterOverridePolicy"
}
```

However, for Karmada components such as `karmada-controller-manager`, logs are currently unstructured. For example:
```text
I0326 01:04:43.392467       1 cluster_status_controller.go:129] Syncing cluster status: member1
```

Having a structured logging option would greatly improve log ingestion and analysis by allowing key fields (e.g., cluster names, log levels) to be extracted more reliably. A possible structured version of this log might look like this:

```json
{
  "ts":"2023-03-26T01:04:43.392467Z",
  "logger":"cluster_status_controller",
  "level": "info",
  "msg":"Syncing cluster status",
  "clusterName":"member1"
}
```

Why is this needed:

- Better Log Ingestion: Structured logs (e.g., JSON) can be seamlessly integrated into various log pipelines and services (Elastic, Loki, Splunk, etc.) without relying on brittle and maintenance-heavy log parsers or regular expressions used to extra fields from unstructured log data.

- Efficient Querying: When logs are structured, fields can be indexed and searched efficiently. This enables quicker troubleshooting and data analysis.

- Consistency with Kubernetes: Since upstream Kubernetes components already provide a structured logging option, it’s natural to extend the same functionality to Karmada for consistency and ease of use.

[...]

PR #6430, opened 2025-06-06
JSON logging support for Karmada Webhook component

seanlaii:
**What type of PR is this?**
/kind feature

Part of #6230 

Tests with feature on:
```
{"ts":1749182199458.613,"caller":"app/webhook.go:126","msg":"karmada-webhook version: version.Info{GitVersion:\"v1.14.0-10-gbc4348d4a-dirty\", GitCommit:\"bc4348d4a0d15dc06b724d115ade2f5a01c10904\", GitShortCommit:\"bc4348d4a\", GitTreeState:\"dirty\", BuildDate:\"2025-06-06T03:53:39Z\", GoVersion:\"go1.24.2\", Compiler:\"gc\", Platform:\"linux/arm64\"}","v":0}
{"ts":1749182199465.4758,"caller":"features/envvar.go:172","msg":"Feature gate default state","v":1,"feature":"ClientsPreferCBOR","enabled":false}
{"ts":1749182199465.6233,"caller":"features/envvar.go:172","msg":"Feature gate default state","v":1,"feature":"InformerResourceVersion","enabled":false}
{"ts":1749182199465.6404,"caller":"features/envvar.go:172","msg":"Feature gate default state","v":1,"feature":"WatchListClient","enabled":false}
{"ts":1749182199465.6487,"caller":"features/envvar.go:172","msg":"Feature gate default state","v":1,"feature":"ClientsAllowCBOR","enabled":false}
{"ts":1749182199475.0796,"caller":"app/webhook.go:174","msg":"Registering webhooks to the webhook server","v":0}
{"ts":1749182199475.2085,"logger":"controller-runtime.webhook","caller":"webhook/server.go:183","msg":"Registering webhook","path":"/validate-cronfederatedhpa","v":0}
```

New Flags after change:
```
--log-flush-frequency duration                                                
            Maximum number of seconds between log flushes (default 5s)
--log-json-info-buffer-size quantity                                          
            [Alpha] In JSON format with split output streams, the info messages can
            be buffered for a while to increase performance. The default value of
            zero bytes disables buffering. The size can be specified as number of
            bytes (512), multiples of 1000 (1K), multiples of 1024 (2Ki), or powers
            of those (3M, 4G, 5Mi, 6Gi). Enable the LoggingAlphaOptions feature
            gate to use this.
--log-json-split-stream                                                       
            [Alpha] In JSON format, write error messages to stderr and info
            messages to stdout. The default is to write a single stream to stdout.
            Enable the LoggingAlphaOptions feature gate to use this.
--log-text-info-buffer-size quantity                                          
            [Alpha] In text format with split output streams, the info messages can
            be buffered for a while to increase performance. The default value of
            zero bytes disables buffering. The size can be specified as number of
            bytes (512), multiples of 1000 (1K), multiples of 1024 (2Ki), or powers
            of those (3M, 4G, 5Mi, 6Gi). Enable the LoggingAlphaOptions feature
            gate to use this.
--log-text-split-stream                                                       
            [Alpha] In text format, write error messages to stderr and info
            messages to stdout. The default is to write a single stream to stdout.
            Enable the LoggingAlphaOptions feature gate to use this.
--logging-format string                                                       
            Sets the log format. Permitted formats: "json" (gated by
            LoggingBetaOptions), "text". (default "text")
```

```release-note
`karmada-webhook`: Introduced `--logging-format` flag which can be set to `json` to enable JSON logging.
```

### DIFF
--- a/cmd/webhook/app/webhook.go
+++ b/cmd/webhook/app/webhook.go
@@ -26,6 +26,8 @@ import (
 	"github.com/spf13/cobra"
 	"k8s.io/client-go/util/flowcontrol"
 	cliflag "k8s.io/component-base/cli/flag"
+	"k8s.io/component-base/logs"
+	logsv1 "k8s.io/component-base/logs/api/v1"
 	"k8s.io/component-base/term"
 	"k8s.io/klog/v2"
 	controllerruntime "sigs.k8s.io/controller-runtime"
@@ -37,6 +39,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook/conversion"
 
 	"github.com/karmada-io/karmada/cmd/webhook/app/options"
+	"github.com/karmada-io/karmada/pkg/features"
 	versionmetrics "github.com/karmada-io/karmada/pkg/metrics"
 	"github.com/karmada-io/karmada/pkg/sharedcli"
 	"github.com/karmada-io/karmada/pkg/sharedcli/klogflag"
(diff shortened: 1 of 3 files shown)


cmd/webhook/app/webhook.go
```diff
@@ -65,12 +68,37 @@ import (
...
+				return err
+			}
+			logs.InitLogs()
+
+			// Starting from version 0.15.0, controller-runtime expects its consumers to set a logger through log.SetLogger.
+			// If SetLogger is not called within the first 30 seconds of a binaries lifetime, it will get
+			// set to a NullLogSink and report an error. Here's to silence the "log.SetLogger(...) was never called; logs will not be displayed" error
+			// by setting a logger through log.SetLogger.
+			// More info refer to: https://github.com/karmada-io/karmada/pull/4885.
+			controllerruntime.SetLogger(klog.Background())
```
seanlaii, line 99: Still need to set the logger with `controllerruntime.SetLogger` since the webhooks are registered by directly using `controller-runtime`'s `hookServer.Register`. These handlers have a fallback dependency on the global logger, and setting the logger globally is therefore required to prevent potential nil pointer errors during registering webhooks.

jabellard:
> /assign
> 
> @jabellard Do you have any comments?



> /assign
> 
> @jabellard Do you have any comments?

Generally looks good to me.

cmd/webhook/app/webhook.go
RainbowMango: Can you elaborate on it? (I'm not familiar with the log process :( )

And any specific reason to move the code here?

seanlaii: Sure! Thanks for the review!

The main reason for moving `controllerruntime.SetLogger` from `main()` into `PersistentPreRunE` is to ensure that the logger passed to `controller-runtime` respects the command-line flags provided by the user (e.g., `--v`, `--log-file`).
Although it is fine to put it in the original place (`main()` function) since `ValidateAndApply` will affect globally, I think this change makes the application's logging behavior more predictable.

Reference: `SetLogger` is also called after `ValidateAndApply` in [cluster-api](https://github.com/kubernetes-sigs/cluster-api/blob/375fc00183f5c032707429b384fb828da1198a30/main.go#L308-L314).

RainbowMango: Yeah, that makes sense to me.

PR #6431, opened 2025-06-06
fix zones match

whitewindmills:
**What type of PR is this?**
/kind bug

For clusters across multiple AZs, should we relax the restrictions so that as long as there is a matching zone attribute, the cluster can be used as a scheduling candidate cluster? Otherwise, the scheduling results cannot be calculated in many scenarios.

Fixes #

```release-note
`karmada-scheduler`: Relaxed zone affinity matching for multi-zone clusters, so a cluster is eligible when any configured zone overlaps the selector.
```

### DIFF
--- a/pkg/util/selector.go
+++ b/pkg/util/selector.go
@@ -204,11 +204,11 @@ func extractClusterFields(cluster *clusterv1alpha1.Cluster) labels.Set {
 	return clusterFieldsMap
 }
 
-// matchZones checks if zoneMatchExpression can match zones and returns true if it matches.
+// matchZones checks whether the zone selector matches the cluster zones.
 // For unknown operators, matchZones always returns false.
 // The matching rules are as follows:
-// 1. When the operator is "In", zoneMatchExpression must contain all zones, otherwise it doesn't match.
-// 2. When the operator is "NotIn", zoneMatchExpression mustn't contain any one of zones, otherwise it doesn't match.
+// 1. When the operator is "In", any overlapping zone is enough for a match.
+// 2. When the operator is "NotIn", none of the cluster zones can appear in the expression values.
 // 3. When the operator is "Exists", zones mustn't be empty, otherwise it doesn't match.
 // 4. When the operator is "DoesNotExist", zones must be empty, otherwise it doesn't match.
 func matchZones(zoneMatchExpression *corev1.NodeSelectorRequirement, zones []string) bool {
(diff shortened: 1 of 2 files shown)


XiShanYongYe-Chang:
Hi @whitewindmills, the UT has failed.

whitewindmills:
> Hi @whitewindmills, the UT has failed.

thanks, fixed.

whitewindmills:
seems to be unrelated test cases.
/retest

RainbowMango (commented):
/assign

RainbowMango:
https://github.com/karmada-io/karmada/blob/adef1e59748e1e1d31cb75fffe406b5dd69a66d7/pkg/util/selector.go#L210

This PR looks good to me. But this comment shows the previous behavior was in line with expectations and was not a bug.

Can you remind me why it was designed this way before?

whitewindmills:
> But this comment shows the previous behavior was in line with expectations and was not a bug.

Maybe I had not considered the actual usage scenario at that time.

XiShanYongYe-Chang:
This PR is more accurately a behavioral adjustment rather than a bug fix, so could we consider changing the label? Additionally, do you need to update the comment content for the `matchZones` function?

RainbowMango:
Kindly ping @whitewindmills

whitewindmills:
ok, back here next week.

whitewindmills:
@RainbowMango Thanks for calling this out. I revisited the previous behavior and the reason for changing it is tied to how `cluster.spec.zones` is modeled and how the field selector is used in scheduling.

A cluster can span multiple AZs, while a selector like `zone in [zone2]` is usually expressing "the workload is eligible for clusters that can serve zone2", not "every zone carried by that cluster must be inside the selector set". Under the previous implementation, `In` only matched when all cluster zones were contained in the selector values. That made a multi-zone cluster such as `[zone1, zone2]` fail `zone in [zone2]`, even though the cluster does satisfy the requested zone affinity through `zone2`.

That behavior is too strict for the practical scheduling case. What we really need here is overlap semantics:
- `In`: match if there is at least one overlapping zone between the selector and the cluster.
- `NotIn`: keep the existing strict behavior, meaning none of the cluster zones can appear in the selector values.

So this change is not trying to redefine arbitrary selector behavior. It is specifically correcting the zone matching logic for multi-AZ clusters, where requiring all zones to be listed in the selector can exclude valid candidate clusters and lead to scheduling results that are narrower than the user intended.

I also noticed the old comment on `matchZones` still described the previous "all zones must match" behavior, which made the change look inconsistent. I have updated the comment and test names as well so the code, tests, and intended semantics are aligned.

RainbowMango:
@whitewindmills do we need a release note?

whitewindmills:
@RainbowMango Thanks for flagging this. I think this one does need a release note because it changes user-visible scheduling behavior: a multi-zone cluster can now match a zone selector when there is any overlap, whereas before the selector effectively required all cluster zones to be covered.

I have updated the PR description and added a release note to make that behavior change explicit.

RainbowMango:
Thanks, I updated the format of the release notes.

PR #6434, opened 2025-06-09
Ensure EndpointSlice informer cache is synced before reporting EndpointSlice

### LINKED ISSUE #6433: Work has been deleted incorrectly when the karmada-controller restarts
**What happened**:
when the karmada-controller restarts，when processing the seerviceexport object create event, the serviceexport-controller does not check whether the cache is ready before obtaining the cluster eps. as a result, the work object is incorrectly deleted in the subsequent logic.

Do not delete  work incorrectly.

 restarts karmada-controller.

no

**Environment**:
- Karmada version: 1.10.7
- kubectl-karmada or karmadactl version (the result of `kubectl-karmada version` or `karmadactl version`): 1.10.7

XiShanYongYe-Chang:
**What type of PR is this?**

/kind bug

When the `karmada-controller-manager` is restarted, the `servcie-export-controller` will requeue the ServiceExport associated with the collected EndpointSlice resources for processing:

https://github.com/karmada-io/karmada/blob/bd5692f08a259c1a959dc3fd1944071be9fda729/pkg/controllers/mcs/service_export_controller.go#L149-L159

Then, during the processing, the EndpointSlice resource is obtained from the cluster cache:

https://github.com/karmada-io/karmada/blob/bd5692f08a259c1a959dc3fd1944071be9fda729/pkg/controllers/mcs/service_export_controller.go#L399-L421

The cache may not have been synchronized at this time, and an empty EndpointSlice list will be obtained, which will be deleted in the `removeOrphanWork` function.

Fixes #6433 

```release-note
karmada-controller-manager: ensure EndpointSlice informer cache is synced before reporting EndpointSlice
```

### DIFF
--- a/pkg/controllers/mcs/service_export_controller.go
+++ b/pkg/controllers/mcs/service_export_controller.go
@@ -408,6 +408,14 @@ func (c *ServiceExportController) reportEndpointSliceWithServiceExportCreate(ctx
 		return nil
 	}
 
+	// Before retrieving EndpointSlice objects from the informer, ensure the informer cache is synced.
+	// This is necessary because the informer for EndpointSlice is created dynamically in the Reconcile() routine
+	// when a Work resource containing an ServiceExport is detected for the cluster. If the informer is not yet synced,
+	// return an error and wait a retry at the next time.
+	if !singleClusterManager.IsInformerSynced(endpointSliceGVR) {
+		return fmt.Errorf("the informer for cluster %s has not been synced, wait a retry at the next time", serviceExportKey.Cluster)
+	}
+
 	endpointSliceLister := singleClusterManager.Lister(endpointSliceGVR)
 	if endpointSliceObjects, err = endpointSliceLister.ByNamespace(serviceExportKey.Namespace).List(labels.SelectorFromSet(labels.Set{
 		discoveryv1.LabelServiceName: serviceExportKey.Name,
@@ -481,6 +489,14 @@ func (c *ServiceExportController) reportEndpointSliceWithEndpointSliceCreateOrUp
 	singleClusterManager := c.InformerManager.GetSingleClusterManager(clusterName)
 	if singleClusterManager == nil {
 		return nil
+	}
+
+	// Before retrieving ServiceExport objects from the informer, ensure the informer cache is synced.
+	// This is necessary because the informer for ServiceExport is created dynamically in the Reconcile() routine
+	// when a Work resource containing an ServiceExport is detected for the cluster. If the informer is not yet synced,
+	// return an error and wait a retry at the next time.
+	if !singleClusterManager.IsInformerSynced(serviceExportGVR) {
+		return fmt.Errorf("the informer for cluster %s has not been synced, wait a retry at the next time", clusterName)
 	}
 
 	serviceExportLister := singleClusterManager.Lister(serviceExportGVR)
@@ -614,6 +630,7 @@ func cleanEndpointSliceWork(ctx context.Context, c client.Client, work *workv1al
 			klog.Errorf("Failed to update work(%s/%s): %v", work.Namespace, work.Name, err)
 			return err
 		}
+		klog.Infof("Successfully updated work(%s/%s)", work.Namespace, work.Name)
 		return nil
 	}
 
@@ -621,6 +638,7 @@ func cleanEndpointSliceWork(ctx context.Context, c client.Client, work *workv1al
 		klog.Errorf("Failed to delete work(%s/%s), Error: %v", work.Namespace, work.Name, err)
 		return err
 	}
+	klog.Infof("Successfully deleted work(%s/%s)", work.Namespace, work.Name)
 
 	return nil
 }


pkg/controllers/mcs/service_export_controller.go
```diff
@@ -190,6 +190,12 @@ func (c *ServiceExportController) enqueueReportedEpsServiceExport() {
 			continue
 		}
 
+		err = c.waitEpsSyncedOnCluster(context.TODO(), clusterName, 1*time.Second)
+		if err != nil {
+			// never reach here
+			continue
+		}
```
RainbowMango: It looks weird to wait for the cache synced here.

cc @zach593 
Can you help to take a look?

RainbowMango:
Please cc me again once you figure out a better solution.

XiShanYongYe-Chang:
> Please cc me again once you figure out a better solution.

Another modification is to perform a check when using the cluster Informer, which introduces the cost of an additional cache check.

Apart from this, no better solution has been thought of for now.

pkg/controllers/mcs/service_export_controller.go
zach593: In the current implementation of informerManager, checking whether the cache has synced is unavoidable. Personally, I feel it would be better to perform the check or wait before using the informer, as this would enhance the cohesion of the informerManager logic. Also, it seems this approach wouldn’t increase the number of checks or waits, since the current PR also performs a check for each Work resource.

In the long term, we could consider refactoring informerManager to behave more like controller-runtime, blocking until the cache is fully synced, or times out, or the informerManager for that cluster is closed.

pkg/controllers/mcs/service_export_controller.go
```diff
@@ -450,6 +467,11 @@ func (c *ServiceExportController) removeOrphanWork(ctx context.Context, endpoint
 		return err
 	}
 
+	dynamicClusterClient, err := c.ClusterDynamicClientSetFunc(serviceExportKey.Cluster, c.Client, c.ClusterClientOption)
+	if err != nil {
+		klog.ErrorS(err, "Failed to build dynamic cluster client", "cluster", serviceExportKey.Cluster)
```
GitHubxsy: Miss  `return err`?

XiShanYongYe-Chang: Yes, looking forward to trying again next time.

RainbowMango: Where to retry?

Note that the `dynamicClusterClient` will be used in the for-loop which will result in a panic.

pkg/controllers/mcs/service_export_controller.go
```diff
@@ -435,6 +450,8 @@ func (c *ServiceExportController) removeOrphanWork(ctx context.Context, endpoint
 		workName := names.GenerateWorkName(endpointSlice.GetKind(), endpointSlice.GetName(), endpointSlice.GetNamespace())
 		willReportWorks.Insert(workName)
 	}
+	klog.V(5).Infof("Will remove endpointslice(%s/%s) doesn't under namespace %s",
+		serviceExportKey.Namespace, serviceExportKey.Name, names.GenerateExecutionSpaceName(serviceExportKey.Cluster))
```
RainbowMango: I'm confused by this log.
1. Are you going to remove the EndpointSlice or the Work that holds the EndpointSlice?
2. Why are you going to remove the EndpointSlice anyway?

pkg/controllers/mcs/service_export_controller.go
```diff
@@ -450,7 +467,7 @@ func (c *ServiceExportController) removeOrphanWork(ctx context.Context, endpoint
 		return err
 	}
 
-	var errs []error
+	var prepareToDeleteWorks []*workv1alpha1.Work
 	for index := range collectedEpsWorkList.Items {
 		work := collectedEpsWorkList.Items[index]
 		if !util.IsWorkContains(work.Spec.Workload.Manifests, endpointSliceGVK) {
```
RainbowMango: Just a question, the `collectedEpsWorkList` seems to be a bunch of Work that holds the `EndpointSlice`. So, why would the `work.Spec.Workload.Manifests` not contain `endpointSliceGVK`?

I guess the `collectedEpsWorkList` consists of `EndpointSlice` and `ServiceExport` work. right. What you want to do is skip the `ServiceExport` work, right?

pkg/controllers/mcs/service_export_controller.go
```diff
@@ -450,7 +467,7 @@ func (c *ServiceExportController) removeOrphanWork(ctx context.Context, endpoint
 		return err
 	}
 
-	var errs []error
+	var prepareToDeleteWorks []*workv1alpha1.Work
```
RainbowMango:
```suggestion
	var OrphaEndpointSliceWorks []*workv1alpha1.Work
```
Use a self-explanatory variable name.

pkg/controllers/mcs/service_export_controller.go
```diff
@@ -435,6 +450,8 @@ func (c *ServiceExportController) removeOrphanWork(ctx context.Context, endpoint
 		workName := names.GenerateWorkName(endpointSlice.GetKind(), endpointSlice.GetName(), endpointSlice.GetNamespace())
 		willReportWorks.Insert(workName)
 	}
+	klog.V(5).Infof("Will remove endpointslice(%s/%s) doesn't under namespace %s",
```
GitHubxsy: "Removing orphan Work resources for EndpointSlices (%s/%s) in namespace %s"

pkg/controllers/mcs/service_export_controller.go
```diff
@@ -466,7 +483,44 @@ func (c *ServiceExportController) removeOrphanWork(ctx context.Context, endpoint
...
+	}
+
+	var errs []error
+	for index := range prepareToDeleteWorks {
+		work := prepareToDeleteWorks[index]
+
+		if len(work.Spec.Workload.Manifests) == 0 {
+			continue
+		}
+		workload := &unstructured.Unstructured{}
```
RainbowMango:
```suggestion
		orphanEndpointSlice := &unstructured.Unstructured{}
```

pkg/controllers/mcs/service_export_controller.go
XiShanYongYe-Chang: You're right; the log description here is wrong. What we need to delete is ` the Work that holds the EndpointSlice`.

pkg/controllers/mcs/service_export_controller.go
```diff
@@ -466,7 +483,44 @@ func (c *ServiceExportController) removeOrphanWork(ctx context.Context, endpoint
...
+		return err
+	}
+
+	var errs []error
+	for index := range prepareToDeleteWorks {
+		work := prepareToDeleteWorks[index]
+
+		if len(work.Spec.Workload.Manifests) == 0 {
+			continue
+		}
```
RainbowMango: It deserves a comment to explain the situation why a work holds nothing here.

pkg/controllers/mcs/service_export_controller.go
```diff
@@ -408,6 +419,10 @@ func (c *ServiceExportController) reportEndpointSliceWithServiceExportCreate(ctx
 		return nil
 	}
 
+	err = c.waitTargetResourceSyncedOnCluster(ctx, serviceExportKey.Cluster, time.Second, endpointSliceGVR)
+	if err != nil {
+		return err
+	}
```
RainbowMango: It feels extremely redundant and weird to check whether the cache is synced or not every time.

pkg/controllers/mcs/service_export_controller.go
XiShanYongYe-Chang: It might still be a bit inappropriate.

pkg/controllers/mcs/service_export_controller.go
XiShanYongYe-Chang: You're right, that's exactly what I was worried about. However, after you raised this issue, I feel that the inspection here might be redundant. I will come back separately to confirm whether it is indeed an unnecessary check.

pkg/controllers/mcs/service_export_controller.go
```diff
@@ -466,7 +483,48 @@ func (c *ServiceExportController) removeOrphanWork(ctx context.Context, endpoint
 			continue
 		}
 
-		err := cleanEndpointSliceWork(ctx, c.Client, &work)
+		orphaEndpointSliceWorks = append(orphaEndpointSliceWorks, &work)
```
Rains6: orpha -> orphan

pkg/controllers/mcs/service_export_controller.go
```diff
@@ -204,6 +204,17 @@ func (c *ServiceExportController) enqueueReportedEpsServiceExport() {
...
+func (c *ServiceExportController) waitTargetResourceSyncedOnCluster(ctx context.Context, cluster string, interval time.Duration, gvr schema.GroupVersionResource) error {
+	return wait.PollUntilContextCancel(ctx, interval, true,
+		func(_ context.Context) (done bool, err error) {
+			synced := c.InformerManager.WaitForCacheSync(cluster)
+			if synced == nil || !synced[gvr] {
+				return false, nil
+			}
+			return true, nil
+		})
+}
```
RainbowMango:
```suggestion
// waitForResourceSyncOnCluster waits until the specified resource's cache is synced on the given cluster.
// It polls at 1 second intervals until the resource is synced or the context is cancelled.
// Returns an error if the context is cancelled before the resource is synced.
// targetResource: the GroupVersionResource to wait for cache sync.
// targetCluster: the name of the cluster to check for cache sync.
// Returns nil if the resource cache is synced, or an error if the context is cancelled.
func (c *ServiceExportController) waitForResourceSyncOnCluster(ctx context.Context, targetResource schema.GroupVersionResource, targetCluster string) error {
	return wait.PollUntilContextCancel(ctx, time.Second, true, func(_ context.Context) (done bool, err error) {
		synced := c.InformerManager.WaitForCacheSync(targetCluster)
		if synced == nil || !synced[targetResource] {
			return false, nil
		}
		return true, nil
	})
}
```
I don't usually comment on what a code block does, but I expect an exception for this controller because of the complexity.

pkg/controllers/mcs/service_export_controller.go
RainbowMango:
```suggestion
	// Before retrieving EndpointSlice objects from the informer, ensure the informer cache is synced.
	// This is necessary because the informer for EndpointSlice is created dynamically in the Reconcile() routine
	// when a Work resource containing an EndpointSlice is detected for the cluster. If the informer is not yet synced,
	// we wait for the cache to be ready to avoid reading incomplete or missing data.
	if !singleClusterManager.IsInformerSynced(endpointSliceGVR) {
		err = c.waitForResourceSyncOnCluster(ctx, endpointSliceGVR, serviceExportKey.Cluster)
		if err != nil {
			return err
		}
	}
```

RainbowMango: There are 3 approaches we can do when detecting the informer has not yet synced:
1. Just return and retry with a backoff time. It won't block the other items in the queue.
2. Wait for cache sync; it will block the other items in the queue.
3. Turn back to the kube-apiserver of the target member cluster, and avoid retrieving data from the unsynced cache.

RainbowMango: I prefer the 1st approach I mentioned above. Please consider how much time it would hanging here waiting for cache sync.

RainbowMango (commented):
Review is still ongoing.

pkg/controllers/mcs/service_export_controller.go
```diff
@@ -466,7 +483,48 @@ func (c *ServiceExportController) removeOrphanWork(ctx context.Context, endpoint
...
+	}
+
+	dynamicClusterClient, err := c.ClusterDynamicClientSetFunc(serviceExportKey.Cluster, c.Client, c.ClusterClientOption)
+	if err != nil {
+		klog.ErrorS(err, "Failed to build dynamic cluster client", "cluster", serviceExportKey.Cluster)
+		return err
+	}
+
+	// Before deleting the EndpointSlice Work, perform a secondary confirmation by directly accessing the apiserver
+	// of member cluster to verify whether the target EndpointSlice resource exists.
```
RainbowMango:
```suggestion
	// Deleting an EndpointSlice Work is a sensitive operation, as it may cause temporary service unavailability
	// even if the resource will be recreated by a retry loop. To avoid accidental disruption, we perform a double-check
	// by directly querying the member cluster's apiserver to confirm that the EndpointSlice resource is truly absent
	// before proceeding with deletion.
```

pkg/controllers/mcs/service_export_controller.go
```diff
@@ -466,7 +483,48 @@ func (c *ServiceExportController) removeOrphanWork(ctx context.Context, endpoint
...
+	var errs []error
+	for index := range orphanEndpointSliceWorks {
+		orphanEpsWork := orphanEndpointSliceWorks[index]
+
+		// To prevent a panic caused by the absence of the index during the unmarshalling of the Work Manifests.
```
RainbowMango:
```suggestion
		// This check avoids a potential panic due to a nil reference when accessing the Work Manifests.
		// It can also avoid annoying false positive lint checks.
```

pkg/controllers/mcs/service_export_controller.go
```diff
@@ -435,6 +450,8 @@ func (c *ServiceExportController) removeOrphanWork(ctx context.Context, endpoint
 		workName := names.GenerateWorkName(endpointSlice.GetKind(), endpointSlice.GetName(), endpointSlice.GetNamespace())
 		willReportWorks.Insert(workName)
 	}
+	klog.V(5).InfoS("Will remove Works that holds Endpointslice reported by ServiceExport(%s/%s) under cluster",
+		"ServiceExport", serviceExportKey.Namespace+"/"+serviceExportKey.Name, "cluster", serviceExportKey.Cluster)
```
RainbowMango: Still confusing, Please remove this log.

pkg/controllers/mcs/service_export_controller.go
```diff
@@ -466,7 +483,48 @@ func (c *ServiceExportController) removeOrphanWork(ctx context.Context, endpoint
...
+			klog.ErrorS(err, "Failed to unmarshal workload from work", "namespace", orphanEpsWork.GetNamespace(), "name", orphanEpsWork.GetName())
+			errs = append(errs, err)
+			continue
+		}
+
+		_, err = dynamicClusterClient.DynamicClientSet.Resource(endpointSliceGVR).Namespace(orphanEps.GetNamespace()).Get(ctx, orphanEps.GetName(), metav1.GetOptions{})
+		// err equals nil means the target EndpointSlice exist
+		if err == nil {
+			continue
+		}
```
RainbowMango:
```suggestion
		if err == nil {
			// The target EndpointSlice exists, so we skip the deletion.
			continue
		}
```

RainbowMango:
/retitle Ensure EndpointSlice informer cache is synced before reporting EndpointSlice

RainbowMango (approved):
/lgtm
/approve

Let's move this forward. This PR is small, so it should be easy to backport into the release branches. 

Please help to open another issue to track the follow-ups:
- Add some necessary unit tests or integrate tests for this issue, to ensure this not gonna happy again
- Clarify the reason why it is necessary to list-watch the ServiceExport object
- Send a notice email to the Karmada Adopter Group to acknowledge the end users.
- Anything else you think necessary.

RainbowMango:
@zhzhuang-zju Just a reminder, this fix should be included in the next patch release.

PR #6435, opened 2025-06-09
JSON logging for cluster controllers

jabellard:
**What type of PR is this?**

JSON logging support. More details are highlighted in https://github.com/karmada-io/karmada/issues/6230.

Which issue(s) this PR fixes:
Fixes #
Partof https://github.com/karmada-io/karmada/issues/6230

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #6440, opened 2025-06-11
update logging for execution controller

SujoyDutta:
**What type of PR is this?**

Fixes #

```release-note
NONE
```

### DIFF
--- a/.github/workflows/dockerhub-latest-image.yml
+++ b/.github/workflows/dockerhub-latest-image.yml
@@ -42,7 +42,7 @@ jobs:
         with:
           go-version-file: go.mod
       - name: Install Cosign
-        uses: sigstore/cosign-installer@v3.8.2
+        uses: sigstore/cosign-installer@v3.9.0
         with:
           cosign-release: 'v2.2.3'
       - name: install QEMU
--- a/.github/workflows/dockerhub-released-image.yml
+++ b/.github/workflows/dockerhub-released-image.yml
@@ -42,7 +42,7 @@ jobs:
         with:
           go-version-file: go.mod
       - name: Install Cosign
-        uses: sigstore/cosign-installer@v3.8.2
+        uses: sigstore/cosign-installer@v3.9.0
         with:
           cosign-release: 'v2.2.3'
       - name: install QEMU


jabellard:
@SujoyDutta, this is looking great so far. For the logging keys, controller-runtime will automatically add the `name` and `namespace` log keys in the context we used to instantiate the logger, so we actually don't have to manually add them.

pkg/controllers/execution/execution_controller.go
```diff
@@ -80,27 +80,28 @@ type Controller struct {
 // The Controller will requeue the Request to be processed again if an error is non-nil or
 // Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
 func (c *Controller) Reconcile(ctx context.Context, req controllerruntime.Request) (controllerruntime.Result, error) {
-	klog.V(4).Infof("Reconciling Work %s", req.NamespacedName.String())
+	logger := log.FromContext(ctx)
+	logger.Info("Reconciling Work %s", req.NamespacedName.String())
```
XiShanYongYe-Chang: Can we still set the log level after using logger?

jabellard: I think it's fine to use the default log level for these. That will also allow us to set the verbosity of the program to a level lower than 4 to avoid a lot of noise from things such as a replica continously reporting that it has lost a leader election.

jabellard: We can still set the log level if we need to, but for info and error logs, I think it's fine to just use the default log level.

XiShanYongYe-Chang: Thank you for the explanation. I understand that by lowering the log level, users can obtain more useful information at the default log level, rather than having to set the log level to 4, which makes it difficult to distinguish from noisy logs.

RainbowMango:
cc @jabellard for help

jabellard:
@SujoyDutta  , thanks for this work. For now, we're going to refrain from using the controller runtime logging adaptor to keep the scope of the parent task more constrained. As such, we'll continue using the `klog` logging funcs directly in the controllers. Please take a look at this [PR](https://github.com/karmada-io/karmada/pull/6435), which is already merged, and make changes accordingly for this one. Thanks!

SujoyDutta:
> @SujoyDutta , thanks for this work. For now, we're going to refrain from using the controller runtime logging adaptor to keep the scope of the parent task more constrained. As such, we'll continue using the `klog` logging funcs directly in the controllers. Please take a look at this [PR](https://github.com/karmada-io/karmada/pull/6435), which is already merged, and make changes accordingly for this one. Thanks!

@jabellard  sounds good, i'll update it accordingly

SujoyDutta:
@jabellard opened another pr https://github.com/karmada-io/karmada/pull/6487, please take a look

PR #6445, opened 2025-06-12
JSON logging for namespace sync controller

Bhaumik10:
What type of PR is this?

Which issue(s) this PR fixes:
Fixes #
Partof https://github.com/karmada-io/karmada/issues/6230

NONE

### DIFF
--- a/pkg/controllers/namespace/namespace_sync_controller.go
+++ b/pkg/controllers/namespace/namespace_sync_controller.go
@@ -29,13 +29,13 @@ import (
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/record"
-	"k8s.io/klog/v2"
 	controllerruntime "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/log"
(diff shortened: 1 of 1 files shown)


pkg/controllers/namespace/namespace_sync_controller.go
```diff
@@ -68,7 +68,8 @@ type Controller struct {
 // The Controller will requeue the Request to be processed again if an error is non-nil or
 // Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
 func (c *Controller) Reconcile(ctx context.Context, req controllerruntime.Request) (controllerruntime.Result, error) {
-	klog.V(4).Infof("Namespaces sync controller reconciling %s", req.NamespacedName.String())
+	logger := log.FromContext(ctx)
+	logger.V(4).Info("Namespaces sync controller reconciling", "namespace", req.NamespacedName.String())
```
jabellard: I think we should just use the default log level (i.e., `logger.Info(...)`). Level 4 will be grouped with logs that's way too noisy. Also, controller runtime will automatically add log keys such as the name and the namespace of the object being reconciled, so we don't have to explicitly set those keys. Please make those changes.

jabellard (changes requested):
Small changes.

pkg/controllers/namespace/namespace_sync_controller.go
Bhaumik10: I will push suggested changes.

RainbowMango:
cc @jabellard for help

jabellard:
@Bhaumik10 , thanks for this work. For now, we're going to refrain from using the controller runtime logging adaptor to keep the scope of the parent task more constrained. As such, we'll continue using the `klog` logging funcs directly in the controllers. Please take a look at this [PR](https://github.com/karmada-io/karmada/pull/6435), which is already merged, and make changes accordingly for this one. Thanks!

Bhaumik10:
closing this PR as open another one https://github.com/karmada-io/karmada/pull/6482

PR #6454, opened 2025-06-13
make healthz/metrics server ports of the scheduler/descheduler unit tests to be detected automatically to avoid flaky test failures

### LINKED ISSUE #6441: cmd/scheduler unit tests are flaky
#### Which jobs are flaking:
`cmd/scheduler/app`

`TestServeHealthzAndMetrics`

#### Reason for failure:
```
E0611 19:08:35.345640   77644 scheduler.go:284] Failed to serve healthz on 127.0.0.1:8082: listen tcp 127.0.0.1:8082: bind: address already in use
E0611 19:08:35.345723   77644 scheduler.go:284] Failed to serve metrics on 127.0.0.1:8083: listen tcp 127.0.0.1:8083: bind: address already in use
```

#### Anything else we need to know:
As the unit tests of both the `cmd/scheduler` and `cmd/descheduler` are using port 8082 and 8083, they seem to be affecting each other in some scenarios.

Scheduler: https://github.com/karmada-io/karmada/blob/bd5692f08a259c1a959dc3fd1944071be9fda729/cmd/scheduler/app/scheduler_test.go#L69-L70
Descheduler: https://github.com/karmada-io/karmada/blob/bd5692f08a259c1a959dc3fd1944071be9fda729/cmd/descheduler/app/descheduler_test.go#L69-L70

In fact, after I changed either of them to be `8084` and `8085`, unit tests are passing in my environment. I don't have a certain way to reproduce this issue as the unit tests are currently passing on my laptop but failing on my desktop unless I change the ports used in those unit tests.

In order to make unit tests more robust, I suggest using different ports in scheduler and descheduler unit tests. I'm happy to make this simple change if it looks good to you.

Thank you!

zclyne:
**What type of PR is this?**

/kind failing-test

It makes the ports of the metrics and healthz server of the scheduler and the descheduler's unit tests to be detected automatically to avoid flaky test failures.

Fixes #6441 

```release-note
NONE
```

### DIFF
--- a/cmd/descheduler/app/descheduler_test.go
+++ b/cmd/descheduler/app/descheduler_test.go
@@ -18,6 +18,7 @@ package app
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"testing"
 	"time"
@@ -27,6 +28,7 @@ import (
 
 	"github.com/karmada-io/karmada/cmd/descheduler/app/options"
 	"github.com/karmada-io/karmada/pkg/util/names"
+	testingutil "github.com/karmada-io/karmada/pkg/util/testing"
 )
 
 func TestNewDeschedulerCommand(t *testing.T) {
@@ -66,8 +68,10 @@ func TestDeschedulerCommandFlagParsing(t *testing.T) {
 }
 
 func TestServeHealthzAndMetrics(t *testing.T) {
-	healthAddress := "127.0.0.1:8082"
-	metricsAddress := "127.0.0.1:8083"
+	ports, err := testingutil.GetFreePorts("127.0.0.1", 2)
+	require.NoError(t, err)
+	healthAddress := fmt.Sprintf("127.0.0.1:%d", ports[0])
+	metricsAddress := fmt.Sprintf("127.0.0.1:%d", ports[1])
 
 	go serveHealthzAndMetrics(healthAddress, metricsAddress)
 
--- a/cmd/scheduler/app/scheduler_test.go
+++ b/cmd/scheduler/app/scheduler_test.go
@@ -18,6 +18,7 @@ package app
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"testing"
 	"time"
@@ -27,6 +28,7 @@ import (
 
 	"github.com/karmada-io/karmada/cmd/scheduler/app/options"
 	"github.com/karmada-io/karmada/pkg/util/names"
+	testingutil "github.com/karmada-io/karmada/pkg/util/testing"
 )
 
 func TestNewSchedulerCommand(t *testing.T) {
@@ -66,8 +68,10 @@ func TestSchedulerCommandFlagParsing(t *testing.T) {
 }
 
 func TestServeHealthzAndMetrics(t *testing.T) {
-	healthAddress := "127.0.0.1:8082"
-	metricsAddress := "127.0.0.1:8083"
+	ports, err := testingutil.GetFreePorts("127.0.0.1", 2)
+	require.NoError(t, err)
+	healthAddress := fmt.Sprintf("127.0.0.1:%d", ports[0])
+	metricsAddress := fmt.Sprintf("127.0.0.1:%d", ports[1])
 
 	go serveHealthzAndMetrics(healthAddress, metricsAddress)
 
--- a/pkg/util/testing/helpers.go
+++ b/pkg/util/testing/helpers.go
@@ -21,7 +21,9 @@ import (
 	"crypto/rsa"
 	"crypto/x509"
 	"encoding/pem"
+	"fmt"
 	"math/big"
+	"net"
 	"time"
 )
 
@@ -68,4 +70,38 @@ func GenerateTestCACertificate() (string, string, error) {
 	privKeyPEMData := pem.EncodeToMemory(privKeyPEM)
 
 	return string(certPEMData), string(privKeyPEMData), nil
+}
+
+// GetFreePorts attempts to find n available TCP ports on the specified host. It
+// returns a slice of allocated port numbers or an error if it fails to acquire
+// them.
+func GetFreePorts(host string, n int) ([]int, error) {
+	ports := make([]int, 0, n)
+	listeners := make([]net.Listener, 0, n)
+
+	// Make sure we close all listeners if there's an error.
+	defer func() {
+		for _, l := range listeners {
+			l.Close()
+		}
+	}()
+
+	for i := 0; i < n; i++ {
+		listener, err := net.Listen("tcp", fmt.Sprintf("%s:0", host))
+		if err != nil {
+			return nil, err
+		}
+		listeners = append(listeners, listener)
+		tcpAddr, ok := listener.Addr().(*net.TCPAddr)
+		if !ok {
+			return nil, fmt.Errorf("listener address is not a *net.TCPAddr")
+		}
+		ports = append(ports, tcpAddr.Port)
+	}
+
+	// At this point we have all ports, so we can close the listeners.
+	for _, l := range listeners {
+		l.Close()
+	}
+	return ports, nil
 }


XiShanYongYe-Chang (commented):
It's strange. Why is port 8084 occupied?

zclyne:
> It's strange. Why is port 8084 occupied?

I'm also a bit confused why setting to 8084 and 8085 would fail but 8085 and 8086 succeeds, but I haven't got time to look into it yet. Will continue investigating it

cmd/descheduler/app/descheduler_test.go
```diff
@@ -66,8 +66,8 @@ func TestDeschedulerCommandFlagParsing(t *testing.T) {
 }
 
 func TestServeHealthzAndMetrics(t *testing.T) {
-	healthAddress := "127.0.0.1:8082"
-	metricsAddress := "127.0.0.1:8083"
+	healthAddress := "127.0.0.1:8085"
+	metricsAddress := "127.0.0.1:8086"
```
moawnallah: I think a more reliable approach is to delegate assigning free port to the OS. It could be having something along those lines and perhaps putting it as utility in karmada testing framework:
```go
func getFreePort(host string) (int, error) {
	listener, err := net.Listen("tcp", fmt.Sprintf("%s:0", host)
	if err != nil {
		return 0, err
	}
	defer listener.Close()
	return listener.Addr().(*net.TCPAddr).Port, nil
}
```

zclyne: This looks awesome! I will put it into my PR and thank you so much for the suggestion!

moawnallah: Awesome just ping me whenever you have this diff ready and I would be happy to review it 🙇‍♂️

RainbowMango:
cc @XiShanYongYe-Chang

cmd/descheduler/app/descheduler_test.go
zclyne: Hi @mohamedawnallah I've merged your suggested change into this PR and could you please check it out again when you get a chance? Thank you!

pkg/util/testing/helpers.go
```diff
@@ -69,3 +71,19 @@ func GenerateTestCACertificate() (string, string, error) {
...
+	if err != nil {
+		return 0, err
+	}
+	defer listener.Close()
+	tcpAddr, ok := listener.Addr().(*net.TCPAddr)
+	if !ok {
+		return 0, fmt.Errorf("listener address is not a *net.TCPAddr")
+	}
+	return tcpAddr.Port, nil
+}
```
moawnallah: If we endup using the freeport package mentioned here https://github.com/karmada-io/karmada/pull/6454#discussion_r2218219871 we no longer need this util

cmd/scheduler/app/scheduler_test.go
```diff
@@ -66,8 +68,12 @@ func TestSchedulerCommandFlagParsing(t *testing.T) {
...
 
 func TestServeHealthzAndMetrics(t *testing.T) {
-	healthAddress := "127.0.0.1:8082"
-	metricsAddress := "127.0.0.1:8083"
+	healthzEnpointPort, err := testingutils.GetFreePort("127.0.0.1")
+	require.NoError(t, err)
+	metricsEndpointPort, err := testingutils.GetFreePort("127.0.0.1")
+	require.NoError(t, err)
+	healthAddress := fmt.Sprintf("127.0.0.1:%d", healthzEnpointPort)
+	metricsAddress := fmt.Sprintf("127.0.0.1:%d", metricsEndpointPort)
```
moawnallah: This overall looks functionally good. But I was thinking that `healthzEnpointPort` and `metricsEndpointPort` may end up being equal and would cause the same/similar flaky issue. I found there is a freeport package by hashicorp that provides this guarantee safety. I think we can safely use it e.g `GetN` API:
https://pkg.go.dev/github.com/hashicorp/consul/sdk/freeport

cmd/descheduler/app/descheduler_test.go
moawnallah: > Hi @mohamedawnallah I've merged your suggested change into this PR and could you please check it out again when you get a chance? Thank you!

I have left one potential optimization https://github.com/karmada-io/karmada/pull/6454#discussion_r2218219871 and I think It would be good to go from there 🚀

cmd/scheduler/app/scheduler_test.go
zclyne: This package seems extremely useful. I will make the change and let you know once it's done. Thank you for the suggestion.

zclyne:
> @zclyne Thanks the changes looks really good. I have just two nits otherwise it LGTM 💪:
> 
> * Squashing git commits into one to be ready for potential merging
> * Moving the PR out of the draft mode to signal to reviewers it is ready for review

Thanks for reviewing. I have marked this PR ready for review and will squash the commits as suggested.

pkg/util/testing/helpers.go
XiShanYongYe-Chang: Hi @mohamedawnallah @zclyne, I noticed that we have introduced a new repo: github.com/hashicorp/consul, I'm not sure if this will have any impact on our repo.

I would like to know if we can achieve our goal by modifying the auxiliary function: obtaining n unused IP addresses.

moawnallah: > we have introduced a new repo: github.com/hashicorp/consul, I'm not sure if this will have any impact on our repo.

@XiShanYongYe-Chang That is a good point. Technically this package serves us reliably to determine freeports in the context of testing. I think if we wanna achieve the same reliability perhaps we needs to write it from scratch here. Non-technically there may be unnecessary future dependencies...

> I would like to know if we can achieve our goal by modifying the auxiliary function: obtaining n unused IP addresses.

We can generalize the above utility for `N` ports and I think It would give us sufficiently what we want given the test cases not parallelized if they are I think it would be problematic. But for now something along those lines could do the job:
```go
// GetFreePortN attempts to find n available TCP ports on the specified host.
// It returns a slice of allocated port numbers or an error if it fails to acquire them.
func GetFreePortN(host string, n int) ([]int, error) {
    ports := make([]int, 0, n)
    listeners := make([]net.Listener, 0, n)
    
    // Make sure we close all listeners if there's an error
    defer func() {
        for _, l := range listeners {
            l.Close()
        }
    }()
    
    for i := 0; i < n; i++ {
        listener, err := net.Listen("tcp", fmt.Sprintf("%s:0", host))
        if err != nil {
            return nil, err
        }
        listeners = append(listeners, listener)
        
        tcpAddr, ok := listener.Addr().(*net.TCPAddr)
        if !ok {
            return nil, fmt.Errorf("listener address is not a *net.TCPAddr")
        }
        
        ports = append(ports, tcpAddr.Port)
    }
    
    // At this point we have all ports, so we can close the listeners
    for _, l := range listeners {
        l.Close()
    }
    
    return ports, nil
}
```

moawnallah: @zclyne was so patient and committed admirably to this PR! We would love to hear your thoughts here as well 🙏

moawnallah: @zclyne – I understand the back and forth here and that is an appreciated effort. I opened a Pull request in your fork repo for this change  you can copy pasta this commit diff https://github.com/zclyne/karmada/pull/1/commits/8baa1d755008e5961d5e289670b2e86daf89acfb while removing the `freeport` package. I've already used some of your utility code you commited earlier so please feel free to preserve fully your commit authorship 🙏

Here is the PR on your fork:
https://github.com/zclyne/karmada/pull/1

zclyne: Thanks @mohamedawnallah for implementing the functionality. The code looks good to me! I'm just curious about why introducing a new dependency is not expected?

XiShanYongYe-Chang (commented):
Thanks a lot~
@zclyne @mohamedawnallah 
/lgtm
/approve

Issue #6467, opened 2025-06-17
When the federated resource quota is insufficient, the workload scale-up will still succeed


zhzhuang-zju:
**What happened**:
We introduced federatedresourcequota enforcement in v1.14 to achieve unified quota management across clusters. This is realized by intercepting modifications to workload requirements and the `spec.clusters` field of resource bindings. However, the federated resource quota mechanism fails when workloads scale up.

The root cause is that the binding controller synchronizes workload modifications to member clusters. When the replicas scheduling policy is set to "duplicated", the binding controller does not revise the workload's replicas according to the scheduling results in `bindingSpec.clusters`. As a result, the scale-up operations of workloads are directly synchronized to member clusters, rendering the federated resource quota ineffective.

The detailed process is as follows:
1. Edit the replicas of a workload.
2. The binding controller synchronizes the workload's replicas to member clusters.
3. Karmada-scheduler performs rescheduling, and the request to update `bindingSpec.clusters` is intercepted due to insufficient quota.

As described above, even though the update request for `bindingSpec.clusters` is blocked, the replicas of the workload in member clusters are still synchronized, leading to the failure of the federated resource quota.

**What you expected to happen**:
 When the federated resource quota is insufficient, the workload scale-up will be blocked.

**How to reproduce it (as minimally and precisely as possible)**:
step1: setup karmada env by `hack/local-up-karmada.sh`
step2: deploy a federatedresourcequota
```yaml
apiVersion: policy.karmada.io/v1alpha1
kind: FederatedResourceQuota
metadata:
  name: test
spec:
  overall:
    cpu: 50m
```
step3: deploy a deployment and propagate it to member1
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx
  labels:
    app: nginx
spec:
  replicas: 1
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx:1.14.2
        ports:
        - containerPort: 80
        resources:
          requests:
            cpu: 50m
---
apiVersion: policy.karmada.io/v1alpha1
kind: PropagationPolicy
metadata:
  name: nginx-6d7f8d5f5b
  namespace: default
spec:
  placement:
    clusterAffinity:
      clusterNames:
      - member1
  resourceSelectors:
  - apiVersion: apps/v1
    kind: Deployment
    name: nginx
    namespace: default
```
[...]

PR #6474, opened 2025-06-19
Fix workload scale up bypass FederatedResourceQuota check issue

zhzhuang-zju:
**What type of PR is this?**
/kind bug

We introduced federatedresourcequota enforcement in v1.14 to achieve unified quota management across clusters. This is realized by intercepting modifications to workload requirements and the `spec.clusters` field of resource bindings. However, the federated resource quota mechanism fails when workloads scale up.

The root cause is that the binding controller synchronizes workload modifications to member clusters. When the replicas scheduling policy is set to "duplicated", the binding controller does not revise the workload's replicas according to the scheduling results in `bindingSpec.clusters`. As a result, the scale-up operations of workloads are directly synchronized to member clusters, rendering the federated resource quota ineffective.

The detailed process is as follows:
1. Edit the replicas of a workload.
2. The binding controller synchronizes the workload's replicas to member clusters.
3. Karmada-scheduler performs rescheduling, and the request to update `bindingSpec.clusters` is intercepted due to insufficient quota.

As described above, even though the update request for `bindingSpec.clusters` is blocked, the replicas of the workload in member clusters are still synchronized, leading to the failure of the federated resource quota.

This PR modifies the logic for the binding controller to sync workloads to member clusters. It syncs the replicas in `bindingSpec.clusters` to member clusters instead of the replicas of the workload.

Parts of #6467

```release-note
`karmada-controller-manager`: Fixed the issue that a workload propagated with `duplicated mode` can bypass quota checks during scale up.
```

### DIFF
--- a/pkg/controllers/binding/common.go
+++ b/pkg/controllers/binding/common.go
@@ -48,26 +48,30 @@ func ensureWork(
 ) error {
 	bindingSpec := getBindingSpec(binding, scope)
 	targetClusters := mergeTargetClusters(bindingSpec.Clusters, bindingSpec.RequiredBy)
+	var err error
+	var errs []error
 
 	var jobCompletions []workv1alpha2.TargetCluster
-	var err error
-	if workload.GetKind() == util.JobKind {
+	if workload.GetKind() == util.JobKind && needReviseJobCompletions(bindingSpec.Replicas, bindingSpec.Placement) {
 		jobCompletions, err = divideReplicasByJobCompletions(workload, targetClusters)
 		if err != nil {
 			return err
 		}
 	}
 
-	var errs []error
 	for i := range targetClusters {
 		targetCluster := targetClusters[i]
 		clonedWorkload := workload.DeepCopy()
 
 		workNamespace := names.GenerateExecutionSpaceName(targetCluster.Name)
 
-		// If and only if the resource template has replicas, and the replica scheduling policy is divided,
-		// we need to revise replicas.
-		if needReviseReplicas(bindingSpec.Replicas, bindingSpec.Placement) {
+		// When syncing workloads to member clusters, the controller MUST strictly adhere to the scheduling results
+		// specified in bindingSpec.Clusters for replica allocation, rather than using the replicas declared in the
+		// workload's resource template.
+		// This rule applies regardless of whether the workload distribution mode is "Divided" or "Duplicated".
+		// Failing to do so could allow workloads to bypass the quota checks performed by the scheduler
+		// (especially during scale-up operations) or skip queue validation when scheduling is suspended.
+		if needReviseReplicas(bindingSpec.Replicas) {
 			if resourceInterpreter.HookEnabled(clonedWorkload.GroupVersionKind(), configv1alpha1.InterpreterOperationReviseReplica) {
 				clonedWorkload, err = resourceInterpreter.ReviseReplica(clonedWorkload, int64(targetCluster.Replicas))
 				if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/controllers/binding/common.go
```diff
@@ -48,26 +48,21 @@ func ensureWork(
...
-		// If and only if the resource template has replicas, and the replica scheduling policy is divided,
-		// we need to revise replicas.
-		if needReviseReplicas(bindingSpec.Replicas, bindingSpec.Placement) {
+		// If and only if the resource template has replicas, we need to revise replicas.
+		// The replicas value in "Work" should be consistent with the scheduling result bindingSpec.Clusters,
+		// or bindingSpec.RequiredBy controlled by the dependencies-distributor, rather than bindingSpec.Replicas.
+		// This remains true even when the replica scheduling policy is duplicated.
+		// In certain scenarios, such as scheduling suspension or insufficient federation quota,
+		// bindingSpec.Replicas should not be directly synchronized to the member clusters.
+		if needReviseReplicas(bindingSpec.Replicas) {
```
mszacillo, line 74: ~I might be missing something, but should we perhaps reorder `needReviseReplicas` and  `needReviseJobCompletions`? At the moment, since jobs have replicas, `needReviseReplicas` will return true and skip the `needReviseJobCompletions` segment altogether if ReviseReplica goes through without errors.~

mszacillo (commented):
Thanks for catching this!!

pkg/controllers/binding/common.go
zhzhuang-zju: > needReviseReplicas will return true and skip the needReviseJobCompletions segment altogether if ReviseReplica goes through without errors.

Hi @mszacillo 
Thank you for your quick review! I might not have grasped your meaning. When ReviseReplica proceeds without errors, needReviseJobCompletion will also be executed.

pkg/controllers/binding/common.go
```diff
@@ -77,6 +72,20 @@ func ensureWork(
...
+		// jobSpec.Completions specifies the desired number of successfully finished pods the job should be run with.
+		// When the replica scheduling policy is set to "divided", jobSpec.Completions should also be divided accordingly.
+		// The weight assigned to each cluster roughly equals that cluster's jobSpec.Parallelism value. This approach helps
+		// balance the execution time of the job across member clusters.
+		if needReviseJobCompletions(bindingSpec.Replicas, bindingSpec.Placement) {
+			var jobCompletions []workv1alpha2.TargetCluster
+			if workload.GetKind() == util.JobKind {
+				jobCompletions, err = divideReplicasByJobCompletions(workload, targetClusters)
+				if err != nil {
+					return err
```
zhzhuang-zju: When `err` is not `nil`, do not return directly. Instead, add `err` to `errs`, wait for the `for` loop to complete, and then return it, maintaining consistency with error handling inside the `for` loop.

zhzhuang-zju:
@mszacillo Do you think it's necessary to backport this PR to `release-1.14`?

pkg/controllers/binding/common.go
mszacillo: Ah! I misread the snippet, please disregard :)

mszacillo:
> @mszacillo Do you think it's necessary to backport this PR to `release-1.14`?

Yes I think so. I currently rebased our internal branch against the latest 1.14 release, so having this fix in that release would be helpful.

RainbowMango (commented):
/assign

pkg/controllers/binding/common.go
```diff
@@ -48,26 +48,30 @@ func ensureWork(
...
 
-		// If and only if the resource template has replicas, and the replica scheduling policy is divided,
-		// we need to revise replicas.
-		if needReviseReplicas(bindingSpec.Replicas, bindingSpec.Placement) {
+		// If and only if the resource template has replicas, we need to revise replicas.
+		// The replicas value in "Work" should be consistent with the scheduling result bindingSpec.Clusters,
+		// or bindingSpec.RequiredBy controlled by the dependencies-distributor, rather than bindingSpec.Replicas.
+		// This remains true even when the replica scheduling policy is duplicated.
+		// In certain scenarios, such as scheduling suspension or insufficient federation quota,
+		// bindingSpec.Replicas should not be directly synchronized to the member clusters.
```
RainbowMango:
```suggestion
		// When syncing workloads to member clusters, the controller MUST strictly adhere to the scheduling results
		// specified in bindingSpec.Clusters for replica allocation, rather than using the replicas declared in the
		// workload's resource template.
		// This rule applies regardless of whether the workload distribution mode is "Divided" or "Duplicated".
		// Failing to do so could allow workloads to bypass the quota checks performed by the scheduler
		// (especially during scale-up operations) or skip queue validation when scheduling is suspended.
```
How about this?

pkg/controllers/binding/common.go
```diff
@@ -77,18 +81,22 @@ func ensureWork(
 					continue
 				}
 			}
+		}
 
+		// jobSpec.Completions specifies the desired number of successfully finished pods the job should be run with.
+		// When the replica scheduling policy is set to "divided", jobSpec.Completions should also be divided accordingly.
+		// The weight assigned to each cluster roughly equals that cluster's jobSpec.Parallelism value. This approach helps
+		// balance the execution time of the job across member clusters.
+		if needReviseJobCompletions(bindingSpec.Replicas, bindingSpec.Placement) && len(jobCompletions) > 0 {
```
RainbowMango:
```suggestion
		if len(jobCompletions) > 0 {
```
redundant check.

RainbowMango:
/retitle Fix workload scale up bypass FederatedResourceQuota check issue

RainbowMango:
> > @mszacillo Do you think it's necessary to backport this PR to `release-1.14`?
> 
> Yes I think so. I currently rebased our internal branch against the latest 1.14 release, so having this fix in that release would be helpful.

I don't mind backporting this. 
But I don't think this issue will affect @mszacillo's scenario, as it only applies to workloads using `Duplicated` mode.

PR #6477, opened 2025-06-20
Use rbSpec.clusters and rbSpec.ReplicaRequirements to calculate rb' resource usage

zhzhuang-zju:
**What type of PR is this?**
/kind cleanup

As stated in #6467, the workload replicas of member clusters are consistent with `rb.Clusters` rather than `rb.replicas`. Therefore, when the `federated-resource-quota-enforcement-controller` periodically refreshes `overallUsage` based on `rb`, it should only use `rbSpec.clusters` and `rbSpec.ReplicaRequirements` for calculation.

For example, when a Deployment is scaled up and intercepted by the federated resource quota, its `rb` might be:

```yaml
spec:
  clusters:
  - name: member1
    replicas: 2
  - name: member2
    replicas: 2
  replicas: 3
```

- If calculated based on `rbSpec.replicas`, the total replicas = 3 * len(clusters) = 6  
- If calculated based on `rbSpec.clusters`, the total replicas = 2 + 2 = 4  

The two results are inconsistent.  

The purpose of this PR is to modify the algorithm for calculating `rb` resource usage to more accurately reflect the actual usage.

Parts of #6467
Parts of #6486

The method for calculating RB resource usage selected in this PR is consistent with that of the [validation webhook](https://github.com/karmada-io/karmada/blob/2af742b2d38745007c4c8e687f8799cf422b1e36/pkg/webhook/resourcebinding/validating.go#L325-L360). A common function could be extracted to implement this logic. However, considering that the purpose of this PR is to fix a bug rather than to refactor, I decided to fix the issue first and extract the common function in a follow-up change. This approach also makes the code review and code backport easier. Of course, if you believe it's better to extract the common function right away, please let me know your thoughts.

```release-note
`karmada-controller-manager`: Fixed the issue where the federated-resource-quota-enforcement-controller miscalculates quota usage.
```

### DIFF
--- a/pkg/controllers/federatedresourcequota/federated_resource_quota_enforcement_controller.go
+++ b/pkg/controllers/federatedresourcequota/federated_resource_quota_enforcement_controller.go
@@ -266,32 +266,50 @@ func (c *QuotaEnforcementController) collectQuotaStatus(quota *policyv1alpha1.Fe
 func calculateUsedWithResourceBinding(resourceBindings []workv1alpha2.ResourceBinding, overall corev1.ResourceList) corev1.ResourceList {
 	overallUsed := corev1.ResourceList{}
 	for _, binding := range resourceBindings {
-		if binding.Spec.ReplicaRequirements == nil || binding.Spec.Clusters == nil {
-			continue
-		}
-		used := binding.Spec.ReplicaRequirements.ResourceRequest
-		replicas := binding.Spec.Replicas
-
-		for name, quantity := range used {
-			// Update quantity to reflect number of replicas in resource
-			q := quantity.DeepCopy()
-			q.Mul(int64(replicas))
-
-			// Update quantity to reflect the number of clusters resource is duplicated to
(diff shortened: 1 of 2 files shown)


zhzhuang-zju:
cc @mszacillo @seanlaii

seanlaii (commented):
The PR looks good to me! Thank you! Just curious if we could also do this task, `Extract common resources usage calculation function to binding utils`, in this PR?

zhzhuang-zju:
> Just curious if we could also do this task, Extract common resources usage calculation function to binding utils, in this PR?

Since this PR will be backported to release-1.14, I don't recommend doing refactoring work in this PR.

seanlaii (approved):
Make sense. Thanks!

RainbowMango:
> > Just curious if we could also do this task, Extract common resources usage calculation function to binding utils, in this PR?
> 
> Since this PR will be backported to release-1.14, I don't recommend doing refactoring work in this PR.

So, shall we have a release note?

RainbowMango (commented):
/lgtm
/hold 
for waiting release note

zhzhuang-zju:
> for waiting release note

Done~ 
If the release note is not suitable, please let me know.

PR #6481, opened 2025-06-23
Add quota exceeded conditions in case scheduler fails to update resourcebinding

mszacillo:
**What type of PR is this?**

/kind feature

In case quota is exceeded, it would be helpful to notify Karmada users as to what is happening. This short PR improves the deny message verbosity of our validation webhook and adds this message as a condition to the ResourceBinding spec. 

Example:

```
status:
  conditions:
  - lastTransitionTime: "2025-06-22T14:43:55Z"
    message: 'admission webhook "resourcebinding.karmada.io" denied the request: FederatedResourceQuota(test-namespace/federated-quota)
      exceeded for resource cpu: requested sum 6, limit 4.'
    reason: QuotaExceeded
    status: "False"
    type: Scheduled
```

Part of #6350 

```release-note
`karmada-scheduler`: Added `QuotaExceeded` as the reason for the legacy `Scheduled` condition on ResourceBinding when scheduling fails due to FederatedResourceQuota limits.
```

### DIFF
--- a/pkg/scheduler/helper.go
+++ b/pkg/scheduler/helper.go
@@ -19,8 +19,10 @@ package scheduler
 import (
 	"encoding/json"
 	"errors"
+	"net/http"
 	"reflect"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/klog/v2"
@@ -125,12 +127,18 @@ func getConditionByError(err error) (metav1.Condition, bool) {
 	if fitErrMatcher(err) {
 		return util.NewCondition(workv1alpha2.Scheduled, workv1alpha2.BindingReasonNoClusterFit, err.Error(), metav1.ConditionFalse), true
 	}
+
 	var aggregatedErr utilerrors.Aggregate
 	if errors.As(err, &aggregatedErr) {
 		for _, ae := range aggregatedErr.Errors() {
 			if fitErrMatcher(ae) {
 				// if aggregated NoClusterFit error got, we do not ignore error but retry scheduling.
 				return util.NewCondition(workv1alpha2.Scheduled, workv1alpha2.BindingReasonNoClusterFit, err.Error(), metav1.ConditionFalse), false
+			}
+			// ResourceBinding validation webhook will return error with "FederatedResourceQuota" if quota exceeded
+			var statusErr *apierrors.StatusError
+			if errors.As(ae, &statusErr) && statusErr.Status().Code == http.StatusForbidden && statusErr.Status().Reason == util.QuotaExceededReason {
+				return util.NewCondition(workv1alpha2.Scheduled, workv1alpha2.BindingReasonQuotaExceeded, ae.Error(), metav1.ConditionFalse), false
 			}
 		}
 	}
--- a/pkg/webhook/resourcebinding/validating.go
+++ b/pkg/webhook/resourcebinding/validating.go
@@ -26,6 +26,7 @@ import (
 	admissionv1 "k8s.io/api/admission/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -309,17 +310,31 @@ func (v *ValidatingAdmission) processSingleFRQ(frqItem policyv1alpha1.FederatedR
 
 	potentialNewOverallUsedForThisFRQ := addResourceLists(frqItem.Status.OverallUsed, deltaForThisFRQ)
 
-	if !isAllowed(potentialNewOverallUsedForThisFRQ, frqItem.Spec.Overall) {
-		errMsg := fmt.Sprintf("Quota exceeded for FederatedResourceQuota %s/%s. ResourceBinding %s/%s will be denied.",
+	isAllowed, errMsg := isAllowed(potentialNewOverallUsedForThisFRQ, frqItem)
+	if !isAllowed {
+		klog.Warningf("Quota exceeded for FederatedResourceQuota %s/%s. ResourceBinding %s/%s will be denied.",
 			frqItem.Namespace, frqItem.Name, rbNamespace, rbName)
-		klog.Warning(errMsg)
-		resp := admission.Denied(errMsg)
-		return nil, "", &resp
+		resp := buildDenyResponse(errMsg)
+		return nil, "", resp
 	}
 
 	msg := fmt.Sprintf("Quota check passed for FRQ %s/%s.", frqItem.Namespace, frqItem.Name)
 	klog.V(3).Infof("FRQ %s/%s will be updated. New OverallUsed: %v", frqItem.Namespace, frqItem.Name, potentialNewOverallUsedForThisFRQ)
 	return potentialNewOverallUsedForThisFRQ, msg, nil
+}
+
+func buildDenyResponse(errMsg string) *admission.Response {
+	resp := admission.Response{
+		AdmissionResponse: admissionv1.AdmissionResponse{
+			Allowed: false,
+			Result: &metav1.Status{
+				Message: errMsg,
+				Reason:  util.QuotaExceededReason,
+				Code:    int32(http.StatusForbidden),
+			},
+		},
+	}
+	return &resp
 }
 
 func calculateResourceUsage(rb *workv1alpha2.ResourceBinding) (corev1.ResourceList, error) {
@@ -437,9 +452,10 @@ func addResourceLists(list1, list2 corev1.ResourceList) corev1.ResourceList {
 	return result
 }
 
-func isAllowed(requested, allowedLimits corev1.ResourceList) bool {
+func isAllowed(requested corev1.ResourceList, frqItem policyv1alpha1.FederatedResourceQuota) (bool, string) {
+	allowedLimits := frqItem.Spec.Overall
 	if allowedLimits == nil {
-		return true
+		return true, ""
 	}
 	for name, reqQty := range requested {
 		if reqQty.IsZero() {
@@ -451,11 +467,12 @@ func isAllowed(requested, allowedLimits corev1.ResourceList) bool {
 			continue
 		}
 		if reqQty.Cmp(limitQty) > 0 {
-			klog.Warningf("Quota exceeded for resource %s: requested sum %s, limit %s", name, reqQty.String(), limitQty.String())
-			return false
+			msg := fmt.Sprintf("FederatedResourceQuota(%s/%s) exceeded for resource %s: requested sum %s, limit %s.", frqItem.Namespace, frqItem.Name, name, reqQty.String(), limitQty.String())
+			klog.Warning(msg)
+			return false, msg
 		}
 	}
-	return true
+	return true, ""
 }
 
 func filterResourceListByKeys(original corev1.ResourceList, filterKeySource corev1.ResourceList) corev1.ResourceList {
--- a/pkg/apis/work/v1alpha2/binding_types.go
+++ b/pkg/apis/work/v1alpha2/binding_types.go
@@ -432,6 +432,11 @@ const (
 	// BindingReasonUnschedulable reason in Scheduled condition means that the scheduler can't schedule
 	// the binding right now, for example due to insufficient resources in the clusters.
 	BindingReasonUnschedulable = "Unschedulable"
+
+	// BindingReasonQuotaExceeded reason in Scheduled condition means that the scheduler can't schedule
+	// the binding because the resource requirement exceeds one or more of the FederatedResourceQuotas
+	// defined in the namespace.
+	BindingReasonQuotaExceeded = "QuotaExceeded"
 )
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
--- a/pkg/util/constants.go
+++ b/pkg/util/constants.go
@@ -20,6 +20,7 @@ import (
 	"time"
 
 	discoveryv1 "k8s.io/api/discovery/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
 )
@@ -229,6 +230,11 @@ const (
 	ParallelismField = "parallelism"
 	// CompletionsField indicates the 'completions' field of a job
 	CompletionsField = "completions"
+)
+
+const (
+	// QuotaExceededReason is a unique reason to describe QuotaExceeded events
+	QuotaExceededReason metav1.StatusReason = "QuotaExceeded"
 )
 
 // ContextKey is the key of context.
--- a/pkg/webhook/resourcebinding/validating_test.go
+++ b/pkg/webhook/resourcebinding/validating_test.go
@@ -405,10 +405,7 @@ func TestValidatingAdmission_Handle(t *testing.T) {
 			decoder:            &fakeDecoder{decodeObj: rbCreateExceeds},
 			clientObjects:      []client.Object{frqForCreateExceeds},
 			featureGateEnabled: true,
-			wantResponse: admission.Denied(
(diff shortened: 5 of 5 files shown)


mszacillo:
@zhzhuang-zju @seanlaii

zhzhuang-zju:
thanks~ 
/assign

pkg/scheduler/helper.go
```diff
@@ -134,5 +135,11 @@ func getConditionByError(err error) (metav1.Condition, bool) {
 			}
 		}
 	}
+
+	// ResourceBinding validation webhook will return error with "FederatedResourceQuota" if quota exceeded
+	if strings.Contains(err.Error(), "FederatedResourceQuota") {
```
zhzhuang-zju: When the update of the binding fails due to quota exceeded, the status code returned from the validating webhook is 403. Can this also be used as one of the criteria for determining quota exceedance?
```go
        var statusErr *apierrors.StatusError
	if errors.As(err, &statusErr) && statusErr.Status().Code == http.StatusForbidden && strings.Contains(statusErr.Error(), "FederatedResourceQuota"){
```

pkg/webhook/resourcebinding/validating.go
```diff
@@ -451,11 +451,12 @@ func isAllowed(requested, allowedLimits corev1.ResourceList) bool {
 			continue
 		}
 		if reqQty.Cmp(limitQty) > 0 {
-			klog.Warningf("Quota exceeded for resource %s: requested sum %s, limit %s", name, reqQty.String(), limitQty.String())
-			return false
+			msg := fmt.Sprintf("FederatedResourceQuota(%s/%s) exceeded for resource %s: requested sum %s, limit %s.", frqItem.Namespace, frqItem.Name, name, reqQty.String(), limitQty.String())
```
zhzhuang-zju, line 470: Is the modification intended to display more detailed information in the message for the quota-exceeded condition? If so, from the user's perspective, is the intercepted scheduling result valuable to be displayed?

pkg/webhook/resourcebinding/validating.go
```diff
@@ -437,9 +437,9 @@ func addResourceLists(list1, list2 corev1.ResourceList) corev1.ResourceList {
 	return result
 }
 
-func isAllowed(requested, allowedLimits corev1.ResourceList) bool {
+func isAllowed(requested, allowedLimits corev1.ResourceList, frqItem policyv1alpha1.FederatedResourceQuota) (bool, string) {
```
zhzhuang-zju:
```suggestion
func isAllowed(requested corev1.ResourceList, frqItem policyv1alpha1.FederatedResourceQuota) (bool, string) {
```

zhzhuang-zju (commented):
@mszacillo Good job~ The condition `QuotaExceeded` will significantly improve users' perception of insufficient quotas.
In addition, when the detector modifies rbSpec.requirement and is rejected by the webhook due to insufficient quota, is there also a condition?

mszacillo:
@zhzhuang-zju Thanks for the review! Apologies for the delay here. 

> when the detector modifies rbSpec.requirement and is rejected by the webhook due to insufficient quota, is there also a condition

I didn't cover this in the PR, but I can add this. Should we cover both the `ApplyPolicy` and `ApplyClusterPolicy` methods?

pkg/webhook/resourcebinding/validating.go
mszacillo: Correct — the goal here was to make the error message more specific by indicating which resource limit was exceeded, as the previous message was too generic.

Regarding the scheduling result, I don't think it should be included. From my perspective, QuotaExceeded events relate to the federated quota, meaning the user has exceeded their overall limits across all clusters. The specific scheduling outcome isn't relevant, since the request would be denied regardless of which cluster was selected.

zhzhuang-zju:
> @zhzhuang-zju Thanks for the review! Apologies for the delay here.  

It's okay. Just work at your own pace.   :-D

> I didn't cover this in the PR, but I can add this. Should we cover both the ApplyPolicy and ApplyClusterPolicy methods?  

No need to rush. To be honest, I'm not sure whether a condition or an event is sufficient for this scenario. However, if it's necessary, I think both the `ApplyPolicy` and `ApplyClusterPolicy` methods should be covered.

pkg/webhook/resourcebinding/validating.go
zhzhuang-zju: Make sense~ thanks
PS: If more information needs to be displayed, we can add it later.

pkg/webhook/resourcebinding/validating.go
```diff
@@ -451,11 +452,12 @@ func isAllowed(requested, allowedLimits corev1.ResourceList) bool {
 			continue
 		}
 		if reqQty.Cmp(limitQty) > 0 {
-			klog.Warningf("Quota exceeded for resource %s: requested sum %s, limit %s", name, reqQty.String(), limitQty.String())
-			return false
+			msg := fmt.Sprintf("FederatedResourceQuota(%s/%s) exceeded for resource %s: requested sum %s, limit %s.", frqItem.Namespace, frqItem.Name, name, reqQty.String(), limitQty.String())
+			klog.Warningf(msg)
```
zhzhuang-zju:
```suggestion
			klog.Warning(msg)
```

This is a newly introduced check: the first argument passed to a formatting function (such as `fmt.Errorf` or `klog.Infof`) must be a compile-time string literal. If not, just use the corresponding non-formatting function.

pkg/scheduler/helper.go
```diff
@@ -134,5 +137,12 @@ func getConditionByError(err error) (metav1.Condition, bool) {
 			}
 		}
 	}
+
+	// ResourceBinding validation webhook will return error with "FederatedResourceQuota" if quota exceeded
+	var statusErr *apierrors.StatusError
+	if errors.As(err, &statusErr) && statusErr.Status().Code == http.StatusForbidden && strings.Contains(err.Error(), "FederatedResourceQuota") {
+		return util.NewCondition(workv1alpha2.Scheduled, workv1alpha2.BindingReasonQuotaExceeded, err.Error(), metav1.ConditionFalse), false
+	}
+
```
zhzhuang-zju: It may not be enough to only judge the error type here and update the condition.  

The error type returned by the functions `scheduleResourceBindingWithClusterAffinity`/`scheduleResourceBindingWithClusterAffinities` is a `utilerrors.Aggregate`.

https://github.com/karmada-io/karmada/blob/d74b6da145cfa48e1b0998d55fa01c745975414c/pkg/scheduler/scheduler.go#L657-L661

https://github.com/karmada-io/karmada/blob/d74b6da145cfa48e1b0998d55fa01c745975414c/pkg/scheduler/scheduler.go#L591-L596

@mszacillo Can you help confirm whether the current `scheduleErr` can be recognized as an `apierrors.StatusError`? If not, we need to follow the handling approach of `fiterror`.

https://github.com/karmada-io/karmada/blob/d74b6da145cfa48e1b0998d55fa01c745975414c/pkg/scheduler/helper.go#L128-L136

RainbowMango:
I can tell the idea of this PR is introducing a new reason for the legacy `Scheduled` condition. It's a doable option. But before that, I'd like to revisit the design of the `Scheduled` condition.

The condition can sometimes be ambiguous. For example, in the example you mentioned in the PR description:
```yaml
  conditions:
  - lastTransitionTime: "2025-06-22T14:43:55Z"
    message: 'admission webhook "resourcebinding.karmada.io" denied the request: FederatedResourceQuota(test-namespace/federated-quota)
      exceeded for resource cpu: requested sum 6, limit 4.'
    reason: QuotaExceeded
    status: "False"
    type: Scheduled
```
The workload might be blocked by the webhook during `scaling up`. So, what does `Scheduled: False` mean in this situation? It feels like the workload is never scheduled once. 
I'd like to hear your thoughts on this condition. Is there any way to improve it?

also cc  @whitewindmills here, I'm not sure if it is you talked about this issue before.

whitewindmills:
Thanks for your invitation. I quickly looked through this PR, indeed, when we first designed this condition, it was to reflect whether the resource was successfully scheduled. If it is `False`, it usually means that there is no scheduling result. Therefore, I prefer to add a new type to express `QuotaExceeded`.

RainbowMango:
> Therefore, I prefer to add a new type to express QuotaExceeded.

Maybe, but not sure yet. If we have a condition which can represent the real schedule status, we still can do this with a new `reason`.

mszacillo:
> Therefore, I prefer to add a new type to express QuotaExceeded.

@RainbowMango @whitewindmills Would it make more sense to use the FullyApplied condition? With the condition being false, and the reason being QuotaExceeded and the status being the detailed quota exceeded message? I think this would make a bit more logical sense as the Quota enforcement prevents the works from being applied.

zhzhuang-zju:
> @RainbowMango @whitewindmills Would it make more sense to use the FullyApplied condition? With the condition being false, and the reason being QuotaExceeded and the status being the detailed quota exceeded message? I think this would make a bit more logical sense as the Quota enforcement prevents the works from being applied.

https://github.com/karmada-io/karmada/blob/9966a3fdb7cf2c1e4bde0579b8905cdc8faa8192/pkg/apis/work/v1alpha2/binding_types.go#L414-L416

I'm afraid not, as the condition `FullyApplied` represents that the resource referenced by ResourceBinding or ClusterResourceBinding has been applied to all scheduled clusters. This is more about the process from RB to the work. However, the quota takes effect in the process from the resource template to RB.

mszacillo:
Ah I see. In that case I think we can create a new condition, named something like `QuotaEnforcement`. If the quota is exceeded, we can set the condition to true, with the relevant error message returned by the validation webhook. Otherwise, we can skip the condition altogether, or add it as false and state that quota has not been exceeded.

zhzhuang-zju:
`QuotaEnforcement` is fine to me

RainbowMango:
I'm hesitant to introduce a brand new condition, which would add additional burden to the scheduler. The quota-exceeding situation is a scheduling issue. 

My original intent was to leverage this opportunity to optimize the historical legacy issue(trying to refine the condition `Scheduled`). I see two possible approaches:
1. Continue using `Scheduled`: The condition represents the latest scheduling status, but needs to enhance documentation/comments for clarification.
2. Replace `Scheduled` with two brand new conditions: one represents whechter the referencing workload was scheduled once, and another one reflecting the latest scheduling status.
We can have another dedicated issue to discuss this further.

But, I don't want this to block the current PR. We can proceed by adding a new reason under the existing `Scheduled` condition.

pkg/scheduler/helper.go
zhzhuang-zju: @mszacillo I tested the quota-exceed scenario locally, and the result is as follows:
```bash
status:
  conditions:
  - lastTransitionTime: "2025-07-15T02:29:00Z"
    message: 'admission webhook "resourcebinding.karmada.io" denied the request: FederatedResourceQuota(default/test)
      exceeded for resource cpu: requested sum 150m, limit 100m.'
    reason: SchedulerError
    status: "False"
    type: Scheduled
```
Here, the `condition.Reason` is `SchedulerError` instead of `QuotaExceeded`.

The reason is that the scheduleErr is of type `utilerrors.Aggregate` rather than `apierrors.StatusError`.

It can be modified to:
```go
	var aggregatedErr utilerrors.Aggregate
	if errors.As(err, &aggregatedErr) {
		for _, ae := range aggregatedErr.Errors() {
			if fitErrMatcher(ae) {
				// if aggregated NoClusterFit error got, we do not ignore error but retry scheduling.
				return util.NewCondition(workv1alpha2.Scheduled, workv1alpha2.BindingReasonNoClusterFit, err.Error(), metav1.ConditionFalse), false
			}
			// ResourceBinding validation webhook will return error with "FederatedResourceQuota" if quota exceeded
			var statusErr *apierrors.StatusError
			if errors.As(ae, &statusErr) && statusErr.Status().Code == http.StatusForbidden && strings.Contains(ae.Error(), "FederatedResourceQuota") {
				return util.NewCondition(workv1alpha2.Scheduled, workv1alpha2.BindingReasonQuotaExceeded, ae.Error(), metav1.ConditionFalse), false
			}
		}
	}
```

mszacillo: Oh! Apologies, I forgot to loop back and address this. Lemme do that now.

mszacillo:
Updated PR. Condition is being populated as follows:

```
conditions:
  - lastTransitionTime: "2025-07-15T10:25:16Z"
    message: 'admission webhook "resourcebinding.karmada.io" denied the request: FederatedResourceQuota(test-namespace/federated-quota)
      exceeded for resource cpu: requested sum 6, limit 4.'
    reason: QuotaExceeded
    status: "False"
    type: Scheduled
```

RainbowMango (commented):
/assign

pkg/scheduler/helper.go
```diff
@@ -125,13 +128,19 @@ func getConditionByError(err error) (metav1.Condition, bool) {
...
 	if errors.As(err, &aggregatedErr) {
 		for _, ae := range aggregatedErr.Errors() {
 			if fitErrMatcher(ae) {
 				// if aggregated NoClusterFit error got, we do not ignore error but retry scheduling.
 				return util.NewCondition(workv1alpha2.Scheduled, workv1alpha2.BindingReasonNoClusterFit, err.Error(), metav1.ConditionFalse), false
 			}
+			// ResourceBinding validation webhook will return error with "FederatedResourceQuota" if quota exceeded
+			var statusErr *apierrors.StatusError
+			if errors.As(ae, &statusErr) && statusErr.Status().Code == http.StatusForbidden && strings.Contains(ae.Error(), "FederatedResourceQuota") {
+				return util.NewCondition(workv1alpha2.Scheduled, workv1alpha2.BindingReasonQuotaExceeded, ae.Error(), metav1.ConditionFalse), false
```
RainbowMango: A little bit concerning here because it is using a hardcoded string as an interface, which is brittle if the error message changes.

Can we customize something when the webhook denies the request?

RainbowMango (commented):
Another question is how to pickup all ResourceBindings which is blocked due to the quota insufficent?

pkg/scheduler/helper.go
zhzhuang-zju: > Can we customize something when the webhook denies the request?

~~How about `result.code`? Now it is `http.StatusForbidden`, perhaps we can customize a specific code for indicating quota exceeded.~~

zhzhuang-zju: > How about result.code?

It seems that the `AdmissionResponse` does not allow customizing `result.code`; only HTTP status codes can be used. I tried setting `result.code` to 1, but what the receiver got was `http.StatusBadRequest` instead.

Upon reviewing the relevant fields of `AdmissionResponse`, I discovered a field within it that is suitable for customization - `result.reason`.

We can define a unique `reason` for "quota exceeded" to distinguish it from other reasons why the webhook denies the request. For example:
```go
const reason metav1.StatusReason = "QuotaExceeded"
```

mszacillo: This is great! I'll update the PR with this suggestion.

mszacillo:
> Another question is how to pickup all ResourceBindings which is blocked due to the quota insufficent?

From the user perspective? I typically just use jq and parse the status:

```
mszacillo% kubectl get resourcebindings -A -o json \
  | jq '.items[] | select(.status.conditions[]? | select(.reason == "QuotaExceeded")) | {namespace: .metadata.namespace, name: .metadata.name}'
  
{
  "namespace": "test-namespace",
  "name": "a-23-ticker-demo-flinkdeployment"
}
```

mszacillo:
Hi @RainbowMango, I've cleaned up this PR and fixed the lint issues. Is there anything else that needs addressing?

pkg/webhook/resourcebinding/validating.go
```diff
@@ -309,19 +310,33 @@ func (v *ValidatingAdmission) processSingleFRQ(frqItem policyv1alpha1.FederatedR
...
+		resp := buildDenyResponse(errMsg)
+		return nil, "", resp
 	}
 
 	msg := fmt.Sprintf("Quota check passed for FRQ %s/%s.", frqItem.Namespace, frqItem.Name)
 	klog.V(3).Infof("FRQ %s/%s will be updated. New OverallUsed: %v", frqItem.Namespace, frqItem.Name, potentialNewOverallUsedForThisFRQ)
 	return potentialNewOverallUsedForThisFRQ, msg, nil
 }
 
+func buildDenyResponse(errMsg string) *admission.Response {
```
zhzhuang-zju, line 326: Func `buildDenyResponse` could take `Reason` as a parameter to improve its generality. However, this is not a blocking issue.

zhzhuang-zju:
@mszacillo Great work! Thanks~
/lgtm

RainbowMango:
This is still on my queue, will get back to this ASAP. Sorry for the delay.

RainbowMango (approved):
Looks great!
/approve
Sorry for the delay.

I updated the release notes in the PR description, by the way.

PR #6490, opened 2025-06-25
Bump go version to 1.24.4

seanlaii:
**What type of PR is this?**
/kind cleanup

Bump go version to 1.24.4 for upgrading Kubernetes dependency.
Part of #6489

```release-note
Bump go version to 1.24.4.
```


seanlaii:
Hi @RainbowMango ,

I've encountered a recurring issue that causes both `golangci-lint` and the unit tests (via `make test`) to fail. The error is `non-constant format string`.
This error is triggered by `govet` when a variable is used as the format string in a function call that accepts variadic arguments (`...interface{}`). 
This error seems to affect many projects: https://github.com/kubernetes/kubernetes/issues/127191
Root cause: https://github.com/golang/go/issues/60529#issuecomment-2043669945

A common example in the codebase is:

```go// This fails the 'vet' check
setCondition(hpa, ..., reason, message) 
```
This happens because the `setCondition` function signature is (`..., message string, args ...interface{}`), and `vet` expects a constant string (like "`%s`") instead of the message variable as the format template. The vet-compliant fix would be `setCondition(hpa, ..., reason, "%s", message)`.

To address this cleanly, I propose the following plan:
- Fix the lint issue and unit tests issue in a separate PR.
- Once the fix PR is merged, we can come back to this PR to upgrade the version.

Please let me know if this plan sounds good to you. Thanks for your time and guidance!

-----
This is the fix PR: https://github.com/karmada-io/karmada/pull/6494

RainbowMango:
> Please let me know if this plan sounds good to you. Thanks for your time and guidance!

+1
I appreciate the clarification.

RainbowMango (approved):
/lgtm
/approve

PR #6494, opened 2025-06-26
fix: Address all `govet` printf issues across the codebase

### LINKED ISSUE #6489: [Umbrella] Bump Kubernetes dependency from v1.32.3 to v1.33.2
This issue is used to track Kubernetes dependency update tasks.

**Prerequisites**
- [ ] Bump Golang to v1.24.4
- [ ] TBD

**Bump**
- [ ] Bump Kubernetes with minimal changes. (@RainbowMango )
  * Bumps k8s.io/* to v0.33.2
    * Adopted breaking changes: 
      * TBD

**Follow Up**
Things should be cleaned up after the dependency bump.

- [ ] TBD

seanlaii:
**What type of PR is this?**
/kind cleanup

This PR resolves all `non-constant format string` errors reported by `govet` across the codebase when upgrading to **Go 1.24**.

These errors began causing [CI failures](https://github.com/karmada-io/karmada/pull/6490) in `golangci-lint` and `make test` when building with newer Go toolchains (e.g., **Go 1.24+**). This cleanup is a necessary prerequisite to unblock the main goal of upgrading the project's Go version. By fixing these issues now, we can ensure the subsequent version bump PR is clean and focused solely on the upgrade itself.

Fixes #6489

The changes in this PR are purely mechanical refactoring and do not alter any program logic. The modifications consistently follow two main patterns:
- `fmt.Errorf(variable)` has been replaced with `errors.New(variable)` for creating simple errors from a pre-existing string.
- `klog.V(X).Infof(variable)` has been replaced with `klog.V(X).Info(variable)` for logging a pre-existing message without formatting.

This approach ensures the codebase is compliant with stricter govet checks while using the most idiomatic Go constructs for each case.

```release-note
None
```

### DIFF
--- a/pkg/controllers/federatedhpa/federatedhpa_controller.go
+++ b/pkg/controllers/federatedhpa/federatedhpa_controller.go
@@ -517,7 +517,7 @@ func (c *FHPAController) scaleForTargetCluster(ctx context.Context, clusters []s
 		if err != nil {
 			errMsg := fmt.Sprintf("couldn't convert selector into a corresponding internal selector object: %v", err)
 			c.EventRecorder.Event(hpa, corev1.EventTypeWarning, "InvalidSelector", errMsg)
-			setCondition(hpa, autoscalingv2.ScalingActive, corev1.ConditionFalse, "InvalidSelector", errMsg)
+			setCondition(hpa, autoscalingv2.ScalingActive, corev1.ConditionFalse, "InvalidSelector", "%s", errMsg)
 			continue
 		}
 
@@ -633,7 +633,7 @@ func (c *FHPAController) computeReplicasForMetrics(ctx context.Context, hpa *aut
 	// return an error and set the condition of the hpa based on the first invalid metric.
 	// Otherwise set the condition as scaling active as we're going to scale
 	if invalidMetricsCount >= len(metricSpecs) || (invalidMetricsCount > 0 && replicas < specReplicas) {
-		setCondition(hpa, invalidMetricCondition.Type, invalidMetricCondition.Status, invalidMetricCondition.Reason, invalidMetricCondition.Message)
+		setCondition(hpa, invalidMetricCondition.Type, invalidMetricCondition.Status, invalidMetricCondition.Reason, "%s", invalidMetricCondition.Message)
 		return -1, "", statuses, time.Time{}, invalidMetricError
 	}
 	setCondition(hpa, autoscalingv2.ScalingActive, corev1.ConditionTrue, "ValidMetricFound", "the HPA was able to successfully calculate a replica count from %s", metric)
@@ -678,7 +678,7 @@ func (c *FHPAController) validateAndParseSelector(hpa *autoscalingv1alpha1.Feder
 	if err != nil {
 		errMsg := fmt.Sprintf("couldn't convert selector into a corresponding internal selector object: %v", err)
 		c.EventRecorder.Event(hpa, corev1.EventTypeWarning, "InvalidSelector", errMsg)
-		setCondition(hpa, autoscalingv2.ScalingActive, corev1.ConditionFalse, "InvalidSelector", errMsg)
+		setCondition(hpa, autoscalingv2.ScalingActive, corev1.ConditionFalse, "InvalidSelector", "%s", errMsg)
 		return nil, errors.New(errMsg)
 	}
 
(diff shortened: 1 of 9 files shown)


RainbowMango:
@zhzhuang-zju Could you please take a look?

pkg/controllers/federatedhpa/federatedhpa_controller.go
```diff
@@ -517,7 +517,7 @@ func (c *FHPAController) scaleForTargetCluster(ctx context.Context, clusters []s
 		if err != nil {
 			errMsg := fmt.Sprintf("couldn't convert selector into a corresponding internal selector object: %v", err)
 			c.EventRecorder.Event(hpa, corev1.EventTypeWarning, "InvalidSelector", errMsg)
-			setCondition(hpa, autoscalingv2.ScalingActive, corev1.ConditionFalse, "InvalidSelector", errMsg)
+			setCondition(hpa, autoscalingv2.ScalingActive, corev1.ConditionFalse, "InvalidSelector", "%s", errMsg)
```
zhzhuang-zju, line 520: I understand why such changes are made, but it looks a bit odd. Can we refer to the naming conventions of "event" and "eventf", and split "setCondition" into "setCondition" and "setConditionf" to handle whether Sprintf is needed respectively? 
For example:
```go
setCondition(hpa, autoscalingv2.ScalingActive, corev1.ConditionFalse, "InvalidSelector", errMsg)
setConditionf(hpa, autoscalingv2.ScalingActive, corev1.ConditionFalse, "InvalidSelector", "err: %s", errMsg)
```

pkg/util/lifted/lua/string_safe.go
```diff
@@ -231,7 +231,7 @@ func strGmatch(L *lua.LState) int {
 	validateStrParamsLen(L, pattern)
 	mds, err := pm.Find(pattern, []byte(str), 0, -1)
 	if err != nil {
-		L.RaiseError(err.Error())
+		L.RaiseError("%s", err.Error())
```
zhzhuang-zju, line 234: https://github.com/karmada-io/karmada/blob/ea9bcecb6a3b7fa93685f69d426aa5d0857cae61/pkg/util/lifted/lua/string_safe.go#L122

Just curious, doesn't this need to be modified?

pkg/controllers/federatedhpa/federatedhpa_controller.go
XiShanYongYe-Chang: The current way of modifying to be consistent with Kubernetes seems to be fine, ref this comment: https://github.com/kubernetes/kubernetes/pull/127422#discussion_r1766921935

pkg/util/lifted/lua/string_safe.go
XiShanYongYe-Chang: +1

Does such a check mean no modification is needed?
https://github.com/karmada-io/karmada/blob/ea9bcecb6a3b7fa93685f69d426aa5d0857cae61/vendor/github.com/yuin/gopher-lua/state.go#L685-L687

pkg/controllers/federatedhpa/federatedhpa_controller.go
zhzhuang-zju: > The current way of modifying to be consistent with Kubernetes seems to be fine, ref this comment: https://github.com/kubernetes/kubernetes/pull/127422#discussion_r1766921935

Thank you for the information. Although I still think "%s" is quite odd, it can be accepted to keep in line with Kubernetes.

XiShanYongYe-Chang: +1 It feels strange, if it's necessary, we can handle it in a separate PR.

pkg/util/lifted/lua/string_safe.go
seanlaii: Good catch! After investigation, I found that what `govet` actually checks is that the first argument passed to a formatting function (like `fmt.Errorf` or `klog.Infof`) must be a `compile-time string literal`, not a `variable` whose value is only determined at **runtime**.

For example:
`L.RaiseError("constant string")` will work, while
```
errStr := "constant string"
L.RaiseError(errStr)
```
will not work.
Sample error message: `./string_safe.go:156:16: non-constant format string in call to (*github.com/yuin/gopher-lua.LState).RaiseError`

zhzhuang-zju: Get it, thanks~

PR #6495, opened 2025-06-26
[Proposal] Support cluster failover for stateful workloads

XiShanYongYe-Chang:
**What type of PR is this?**

/kind feature
/kind documentation

Support cluster failover for stateful workloads.

Fixes #

```release-note
NONE
```

### DIFF
--- a/docs/proposals/failover/stateful-workload-cluster-failover.md
+++ b/docs/proposals/failover/stateful-workload-cluster-failover.md
@@ -1,0 +1,414 @@
+---
+title: Support cluster failover for stateful workloads
+authors:
+  - "@XiShanYongYe-Chang"
+reviewers:
+  - "@mszacillo"
+  - "@Dyex719"
+  - "@RainbowMango"
+approvers:
+  - "@kevin-wangzefeng"
+  - "@RainbowMango"
+creation-date: 2025-06-26
+---
+
+This proposal is modified base on the design of [#5116](https://github.com/karmada-io/karmada/pull/5116).
+
+## Summary
+
+The advantage of stateless workloads is their higher fault tolerance, as the loss of workloads does not impact application correctness or require data recovery. In contrast, stateful workloads (like Flink) depend on checkpointed or saved state to recover after failures. If that state is lost, the job may not be able to resume from where it left off.
+
+For workloads, failover scenarios include two cases: the first is when the workload itself fails and needs to be migrated to another cluster; the second is when the cluster encounter a failure, such as a network partition, requiring some workloads within the cluster to be migrated.
+
+Workload failover may be triggered either by the workload's own failure (requiring migration to another cluster) or by cluster failures (e.g., node crashes, network partitions, or control plane outages) that necessitate relocating affected workloads.
+
+The Karmada community has completed the design and development for the first scenario in version v1.12: [#5788](https://github.com/karmada-io/karmada/issues/5788). This proposal will build upon the previous design experience to support the second scenario, which involves stateful workload failover caused by cluster failures. The API design will unify the handling of common aspects in both scenarios to avoid redundant fields and improve the user experience.
+
+## Motivation
+
+Karmada currently supports propagating various types of resources, including Kubernetes objects and CRDs, this also includes stateful workload, it enabling multi-cluster resilience and elastic scaling for distributed applications. However, the Karmada system's processing logic is based on the assumption that the propagated resources are stateless. In failover scenarios, users may want to preserve a certain state before failure so that workloads can resume from the point where they stopped in the previous cluster.
+
+For CRDs that define resources involved in data processing (such as Flink or Spark), restarting the application from a previous checkpoint can be particularly useful. This allows the application to seamlessly resume processing from the point of interruption while avoiding duplicate processing. Karmada v1.12 supports stateful workload migration after application failures, and the next step is to extend support for stateful workload migration following cluster failures.
+
+### Goals
+
+Support stateful application cluster failover, allowing users to customize the state information that needs to be preserved for workloads in the failed cluster within Karmada. This information is then passed to the migrated workloads, enabling the newly started workloads to recover from the state where the previous cluster stopped.
+
+### No-Goals
+
+- No optimization will be made for cluster failover functionality.
+- No optimization will be made for application failover functionality.
+
+## Proposal
+
+This proposal introduces a way for users to persist state required for their application in the event of the cluster failover. That way, once rescheduled, users can configure their applications to resume using the labels saved by Karmada. Since stateful applications have different implementations of how to retrieve the last state, it is important that this feature is configurable.
+
+To support this feature, we will need at minimum:
+
+- A mechanism with which stateful applications can preserve state during cluster failover and reload before resuming.
+- A way users can configure which state will be preserved by the Karmada.
+- A way users can configure how the saved information is displayed on the restored stateful application labels.
+
+> Note: One important detail is that if all the replicas of the stateful application are not migrated together, it is not clear when the state needs to be restored. In this proposal we focus on the use case where all the replicas of a stateful application are migrated together. One way to ensure this is to make all the replicas scheduled together using spreadConstraints.
+
+```yaml
+spreadConstraints:
+  - spreadByField: cluster
+    maxGroups: 1
+    minGroups: 1
+```
+
+### Flink use case
+
+This section helps to provide a basic understanding of how Flink applications work.
+
+In Flink applications, checkpoints are snapshots of the application's state at a specific point in time. They contain information about all records processed up to that moment. This information is continuously persisted to durable storage at regular intervals:
+
+```yaml
+jobStatus:
+  checkpointInfo:
+    formatType: FULL
+    lastCheckpoint:
+      formatType: FULL
+      timeStamp: 1734462491693
+      triggerType: PERIODIC
+    lastPeriodicCheckpointTimestamp: 1734462491693
+    triggerId: 447b790bb9c88da9ae753f00f45acb0e
+    triggerTimestamp: 1734462506836
+    triggerType: PERIODIC
+  jobId: e6fdb5c0997c11b0c62d796b3df25e86
+```
+
+To retrieve the latest Flink application state from persistent storage, the `jobId` of the job being recovered and the path to the persisted state are required. The Flink Operator can then use this information to retrieve the last state by providing a few modifications to the Flink resource spec:
+
+```yaml
+spec:
+  jobSpec:
+    initialSavepointPath: “desired checkpoint path to be resumed from (s3p://)”
+    upgradeMode: savepoint
+    state: running
+```
+
+### User Stories
+
+#### Story 1
+
+As an administrator or operator, I can define the state preservation strategy for Flink applications deployed on Karmada after a cluster failure. This includes preserving the `jobID` of the Flink application, allowing the application to resume processing from the point of failure using the `jobID` after migration.
+
+The recovery process for Flink workloads can be summarized as follows:
+
+1. Karmada obtains the `{ .jobStatus.jobId }` from the FlinkDeployment in the failed cluster.
+2. Karmada synchronizes the `jobId` to the migrated FlinkDeployment using a label with a `<user-defined key>:<jobId>`.
+3. The user retrieves checkpoint data using the `jobId` and generates the initialSavepointPath: `/<shared-path>/<job-namespace>/<jobID>/checkpoints/<checkpoint>`.
+4. The user injects the initialSavepointPath information into the migrated FlinkDeployment.
+5. The Flink Operator restarts Flink from the savepoint.
+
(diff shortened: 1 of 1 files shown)


XiShanYongYe-Chang:
Hi @mszacillo, any comments?

docs/proposals/failover/stateful-workload-cluster-failover.md
```diff
@@ -0,0 +1,264 @@
...
+  - "@kevin-wangzefeng"
+  - "@RainbowMango"
+creation-date: 2025-06-26
+---
+
+This proposal is modified base on the design of [#5116](https://github.com/karmada-io/karmada/pull/5116).
+
+## Summary
+
+The advantage of stateless workloads is their higher fault tolerance, as the loss of workloads does not affect user sessions. However, for stateful workloads (such as Flink), the loss of workloads may result in the loss of session data, preventing the workload from recovering correctly.
```
RainbowMango: What's the `session data` thing here?

docs/proposals/failover/stateful-workload-cluster-failover.md
```diff
@@ -0,0 +1,264 @@
...
+
+For workloads, failover scenarios include two cases: the first is when the workload itself fails and needs to be migrated to another cluster; the second is when the cluster encounter a failure, such as a network partition, requiring some workloads within the cluster to be migrated.
```
RainbowMango, line 21: Does `network partition` mean the network of a specific cluster is blocked? 
I doubt it's the typical scenarios that should trigger the cluster failover.

docs/proposals/failover/stateful-workload-cluster-failover.md
```diff
@@ -0,0 +1,264 @@
...
+
+## Motivation
+
+Karmada currently supports propagating various types of resources, including Kubernetes objects and CRDs, which is particularly useful for ensuring the resilience of stateful workloads in multi-cluster environments. However, the Karmada system's processing logic is based on the assumption that the propagated resources are stateless. In failover scenarios, users may want to preserve a certain state before failure so that workloads can resume from the point where they stopped in the previous cluster.
+
+For CRDs that handle data processing (such as Flink or Spark), restarting the application from a previous checkpoint can be particularly useful. This allows the application to seamlessly resume processing from the point of interruption while avoiding duplicate processing. Karmada v1.12 supports stateful workload migration after application failures, and the next step is to extend support for stateful workload migration following cluster failures.
```
RainbowMango: CRD can't handle data processing...

docs/proposals/failover/stateful-workload-cluster-failover.md
```diff
@@ -0,0 +1,264 @@
...
+
+The Karmada community has completed the design and development for the first scenario in version v1.12: [#5788](https://github.com/karmada-io/karmada/issues/5788). This proposal will build upon the previous design experience to support the second scenario, which involves stateful workload failover caused by cluster failures. The API design will unify the handling of common aspects in both scenarios to avoid redundant fields and improve the user experience.
+
+## Motivation
+
+Karmada currently supports propagating various types of resources, including Kubernetes objects and CRDs, which is particularly useful for ensuring the resilience of stateful workloads in multi-cluster environments. However, the Karmada system's processing logic is based on the assumption that the propagated resources are stateless. In failover scenarios, users may want to preserve a certain state before failure so that workloads can resume from the point where they stopped in the previous cluster.
```
RainbowMango: > Karmada currently supports propagating various types of resources, including Kubernetes objects and CRDs, which is particularly useful for ensuring the resilience of stateful workloads in multi-cluster environments.

I don't get why `supports various resources` is useful for stateful workloads.

docs/proposals/failover/stateful-workload-cluster-failover.md
RainbowMango, line 1: I think it's worth the have a `Risks and Mitigations` to explain the risks, such as 
1. The cluster failover detection is not that accurate yet.
2. purge mode `Directly` might be blocked if the connection between Karmada and the target cluster is broken.

docs/proposals/failover/stateful-workload-cluster-failover.md
```diff
@@ -0,0 +1,264 @@
...
+
+In Karmada v1.12, the `StatefulFailoverInjection` FeatureGate was introduced to control whether state recovery is performed during the recovery process of stateful applications after application failure. In this proposal, we extend the control of this feature gate to cover stateful application recovery scenarios after cluster failure, thereby unifying the control of stateful application failover recovery.
+
+## System Failure Impact and Mitigation
+
+What situations may cause the feature to fail, and will it cause system crashes?
+
+As mentioned above, this proposal focuses on scenarios where all replicas of a stateful application are migrated together. If users split the replicas of a stateful application and propagate them across different clusters, and the system migrates only some replicas, the stateful application failover recovery feature may not work effectively.
+
+This feature will not cause system crashes.
```
RainbowMango: We won't accept any feature that would cause the system to crash; that's common sense, so it's not necessary to explain.

> As mentioned above, this proposal focuses on scenarios where all replicas of a stateful application are migrated together. If users split the replicas of a stateful application and propagate them across different clusters, and the system migrates only some replicas, the stateful application failover recovery feature may not work effectively.

1. This is the limitation or constraint of this feature, I would prefer to put it to `Notes/Constraints/Caveats (Optional)`.
2. Another question is, can the system identify whether a statefull workload can be split across clusters, considering the use case of propagation ETCD replicas to multiple clusters.

docs/proposals/failover/stateful-workload-cluster-failover.md
```diff
@@ -0,0 +1,264 @@
...
+
+## Graduation Criteria
+
+### Alpha to Beta
+
+- The `StatePreservation` definition in the `PropagationPolicy` API will transition to Beta;
+- The `PreservedLabelState` and `ClustersBeforeFailover` fields in the `ResourceBinding` API will transition to Beta;
```
RainbowMango: When speaking of graduation criteria, we are saying the maturity of a feature, not an API field...

docs/proposals/failover/stateful-workload-cluster-failover.md
```diff
@@ -0,0 +1,264 @@
...
+- End-to-end (E2E) testing has been completed.
```
RainbowMango: How many E2E tests are planned for this feature? I mean, how to know the E2E testing is completed?

docs/proposals/failover/stateful-workload-cluster-failover.md
```diff
@@ -0,0 +1,264 @@
...
+    // "{.status.availableReplicas}".
+    //
+    // +required
+    JSONPath string `json:"jsonPath"`
+}
+```
+
+In this proposal, we have uncommented the `Cluster` field in the `FailoverBehavior` struct to define behaviors related to cluster failover. We have also added the `StatePreservation` field to the `ClusterFailoverBehavior` struct, which is used to specify the policy for preserving and restoring state data of stateful applications during failover events. The type of this field is exactly the same as the `StatePreservation` field in the `ApplicationFailoverBehavior` struct.
+
+The `ResourceBinding` API already meets the design goals and can directly reuse the `PreservedLabelState` and `ClustersBeforeFailover` fields in the `gracefulEvictionTask` without introducing new modifications.
```
RainbowMango: I don't see where explains the `PurgeMode` thing.

docs/proposals/failover/stateful-workload-cluster-failover.md
XiShanYongYe-Chang: Lift it from #5116 

It may be:

- User Session Information: This includes data such as the user's login status, session identifiers, session start time, and other session-related metadata.
- Temporary Data: Data that the application generates while processing user requests, which needs to be consistent across multiple requests.
- State Information: The state that the application maintains during its operation, such as intermediate computation results, checkpoints, and other runtime data.

XiShanYongYe-Chang: Would it be easier to understand if we use `persistent data`?

docs/proposals/failover/stateful-workload-cluster-failover.md
XiShanYongYe-Chang: `network partition` is a term, and we can refer to the explanation provided by [Wikipedia](https://en.wikipedia.org/wiki/Network_partition).

Let me provie a few more typical scenarios:

`e.g., node crashes, network partitions, or control plane outages`

docs/proposals/failover/stateful-workload-cluster-failover.md
mszacillo: How about:

The advantage of stateless workloads is their higher fault tolerance, as the loss of workloads does not impact application correctness or require data recovery. In contrast, stateful workloads (like Flink) depend on checkpointed or saved state to recover after failures. If that state is lost, the job may not be able to resume from where it left off.

docs/proposals/failover/stateful-workload-cluster-failover.md
```diff
@@ -0,0 +1,264 @@
...
+    // Application indicates failover behaviors in case of application failure.
+    // If this value is nil, failover is disabled.
+    // If set, the PropagateDeps should be true so that the dependencies could
+    // be migrated along with the application.
+    // +optional
+    Application *ApplicationFailoverBehavior `json:"application,omitempty"`
+
+    // Cluster indicates failover behaviors in case of cluster failure.
+    // +optional
+    Cluster *ClusterFailoverBehavior `json:"cluster,omitempty"`
```
mszacillo, line 144: Apologies for the delay here, I have been traveling recently. 

I don’t have strong objections to introducing ClusterFailoverBehavior as a separate field. That said, from my perspective, do we expect users to define different state preservation between cluster and application failover?

Right now, we're running on an internal commit that simply reuses the same state preservation defined under FailoverBehavior during cluster failover events, and that’s worked well so far.

I think defining a separate ClusterFailoverBehavior makes more sense once the cluster failover logic is more fully developed — for example, if users can define custom cluster failover criteria. At the moment, including state preservation options in both places feels somewhat redundant. But I also understand if this is just a first step towards having a more formally defined ClusterFailover feature, so maybe we can address this later.

docs/proposals/failover/stateful-workload-cluster-failover.md
XiShanYongYe-Chang: Lift it from https://github.com/karmada-io/karmada/pull/5116

Let me revise the wording.

docs/proposals/failover/stateful-workload-cluster-failover.md
XiShanYongYe-Chang: > This is the limitation or constraint of this feature, I would prefer to put it to Notes/Constraints/Caveats (Optional).

+1

> Another question is, can the system identify whether a statefull workload can be split across clusters, considering the use case of propagation ETCD replicas to multiple clusters.

Sorry, I didn't quite understand the question.

docs/proposals/failover/stateful-workload-cluster-failover.md
XiShanYongYe-Chang: Thanks @mszacillo 

> Apologies for the delay here, I have been traveling recently.

It sounds wonderful. 👍 

> At the moment, including state preservation options in both places feels somewhat redundant. But I also understand if this is just a first step towards having a more formally defined ClusterFailover feature, so maybe we can address this later.

@RainbowMango and I have also discussed this issue. Currently, defining `StatePreservation` in both places does seem somewhat redundant. However, we are also considering another point: we may not yet have sufficient scenarios to prove that the methods of obtaining information during application migration caused by cluster failures and application failures are exactly the same.

Therefore, this redundancy might be an intermediate state that will be eliminated in the future; it could also be validated by new scenarios and become the final state.

I would like to understand whether the current redundancy design is convenient to modify for the solution you are currently running.

RainbowMango: > However, we are also considering another point: we may not yet have sufficient scenarios to prove that the methods of obtaining information during application migration caused by cluster failures and application failures are exactly the same.

Yes, it's exactly the reason. We don't know if the requirements are exactly the same between application failover and cluster failover. Maybe they require different labels to identify the reason.

@XiShanYongYe-Chang This is a good question, and I think we should explain it in this proposal.

docs/proposals/failover/stateful-workload-cluster-failover.md
XiShanYongYe-Chang: Supplemented in the `Test Planning` section.

XiShanYongYe-Chang:
Hi @RainbowMango @mszacillo An updated version has been made based on the comments. Can you help review it again?

docs/proposals/failover/stateful-workload-cluster-failover.md
XiShanYongYe-Chang: Thanks @mszacillo

XiShanYongYe-Chang:
Hi @kevin-wangzefeng, can you help take a review?
/cc @kevin-wangzefeng

Issue #6424, opened 2025-06-03
Design question: weight calculation logic


ryanwuer:
**Please provide an in-depth description of the question you have**:

**What do you think about this question?**:

**Environment**:
- Karmada version: v1.13.0
- Kubernetes version: v1.19.3
- Others:

I try to schedule a replicaset to two K8s, the replicas is 2, the PropagationPolicy is like below:

```
    replicaScheduling:
      replicaDivisionPreference: Weighted
      replicaSchedulingType: Divided
      weightPreference:
        staticWeightList:
          - targetCluster:
              clusterNames:
                - a
            weight: 2
          - targetCluster:
              clusterNames:
                - b
            weight: 1
```

As a result, there're a replicaset with replicas of 2 scheduled to guiyang-test-1, and none scheduled to guiyang-test-2, does this meet design pattern?

I guess the logic is:
```
a= 2/(2+1) * 2 = 1.333 -> 2
b= 2 - 2 = 0
```

What i actually expect is 1 for a and 1 for b. At least 1 for replicas is given to workloads scheduled by karmada. Is that reasonable?

zhzhuang-zju:
> As a result, there're a replicaset with replicas of 2 scheduled to guiyang-test-1, and none scheduled to guiyang-test-2, does this meet design pattern?

@ryanwuer Thanks for reporting this. In fact, the relevant scheduling logic is almost the same as what you guessed. The details are as follows:

Updated:
```text
remain = 2
a= 2/(2+1) * 2 = 1.333 -> 1
remain = 1
b = 1/(2+1) * 2 = 0
remain = 1

# begin assigning the rest replicas
a ++  -> 2
remain --  -> 0
# break as no rest replicas
```

When the replicas is small or when the weight of a certain cluster is too high, it is indeed possible that the replicas in a certain cluster is zero.

ryanwuer:
Is it necessary or reasonable to make sure each target cluster has at least one replica? What do you think?

zhzhuang-zju:
If we ensure that each target cluster has at least one replica, this constraint may conflict with the static weight. For example, suppose the total number of replicas is 50 and there are three target clusters with weights: a 99, b 1, c 1. If each target cluster is guaranteed at least one replica, the final number of replicas should be: a 48, b 1, c 1. In this case, the proportion of cluster a is 0.96, which differs significantly from the set ratio of 99/101.

ryanwuer:
[...]

PR #6520, opened 2025-07-08
docs: add kep about supporting minReplicas for replicas scheduling

### LINKED ISSUE #6511: Support for MinReplicas

It is expected to ensure a minimum replicas in each cluster when scheduling replicas. Currently, using the `divide` strategy, this issue occurs with both dynamic and static cluster preferences.
With dynamic, it might happen that one cluster has no replicas due to uneven resource distribution between clusters.
With static, it could be because unbalanced weights lead to no replicas in the cluster with a smaller weight.

I see some discussion on https://github.com/karmada-io/karmada/issues/4805, but no conclusion. I think it's necessary to extend MinReplicas field.

**Environment**:
- Karmada version:
- Kubernetes version:
- Others:

vie-serendipity:
**What type of PR is this?**

/kind feature

This PR is a proposal about supporting MinReplicas for replicas scheduling.

Fixes #6511 #6424

```release-note
1. `API Change`: Introduced `ClusterConstraint` to the PropagationPolicy API to support MinReplicas.
2. `karmada-webhook`: Add new validation logic.
3. `karmada-scheduler`: Adjust the replicas scheduling logic.
```


docs/proposals/scheduling/minreplicas/README.md
```diff
@@ -0,0 +1,201 @@
...
+
+# Support Per-Cluster MinReplicas in PropagationPolicy
+
+## Summary
+
+Currently, Karmada's PropagationPolicy does not support any `minReplicas` constraint for replica scheduling. This limits flexibility in multi-cluster scenarios, as users cannot enforce minimum replica requirements at either the global or per-cluster level. This proposal introduces a new, extensible struct `ClusterConstraint` under `ClusterPreferences`, allowing users to specify `minReplicas` globally or for specific clusters/groups via `ClusterConstraintTerms`. This enhancement is controlled by a feature gate and aims to provide fine-grained control over replica distribution.
+
+## Motivation
+
+In real-world multi-cluster deployments, users expect every candidate cluster participating in scheduling to have at least one or more replicas of the workload. This ensures that each cluster is always ready to handle business traffic, especially in failover scenarios when other clusters become unavailable. However, due to imbalanced resource capacities and current scheduling strategies, some candidate clusters may end up with zero replicas for both static and dynamic weight preferences.
```
RainbowMango, line 23: > In real-world multi-cluster deployments, users expect every candidate cluster participating in scheduling to have at least one or more replicas of the workload.

This might not be what all users expect. For instance, some users only want Karmada to schedule replicas into a pool consisting of multiple clusters without limiting the number of clusters. What's more, in the case [ReplicaDivisionPreference: Aggregated](https://github.com/karmada-io/karmada/blob/962d56ff086f963b99360289881ff2f33151f916/pkg/apis/policy/v1alpha1/propagation_types.go#L556-L558) where users would rather have as few clusters as possible.

RainbowMango: > This ensures that each cluster is always ready to handle business traffic, especially in failover scenarios when other clusters become unavailable.

I guess what you want to say `High Available` sceanrio instead of `failover`, in your use case you probably don'
t Karmada to migrate your applications when clusters become unavailable.

docs/proposals/scheduling/minreplicas/README.md
```diff
@@ -0,0 +1,201 @@
...
+
+```go
+// ClusterConstraint defines global and per-cluster/group scheduling constraints for replicas.
+type ClusterConstraint struct {
+    // MinReplicas specifies the minimum number of replicas that should be assigned to each selected cluster.
+    // This is a global field for all clusters, mainly used when ClusterConstraintTerms is not set.
+    // If a cluster/group is specified in ClusterConstraintTerms, the value in ClusterConstraintTerms takes precedence.
+    // +kubebuilder:validation:Minimum=0
+    // +optional
+    MinReplicas *int32 `json:"minReplicas,omitempty"`
```
RainbowMango, line 73: This is the constraint for the scheduler to make the decision. As I mentioned above, it might not apply to all usage scenarios, and it will make it difficult for the scheduler to make ideal decisions in certain usage scenarios. For example, in the scenario I mentioned earlier, when the replica division preference is "Aggregated", it may not be able to balance both aspects.

In addition, what if the total replica can't meet each selected cluster has a `minReplicas`? For example, the scheduler trying to assign `1` replica to two clusters.

docs/proposals/scheduling/minreplicas/README.md
vie-serendipity: > This might not be what all users expect. For instance, some users only want Karmada to schedule replicas into a pool consisting of multiple clusters without limiting the number of clusters.

Yes, I mean some user might expect every candidate cluster participating in scheduling. I will adjust this later.

docs/proposals/scheduling/minreplicas/README.md
vie-serendipity: > For example, in the scenario I mentioned earlier, when the replica division preference is "Aggregated", it may not be able to balance both aspects.

This constraint field will only make effect when `replicaSchedulingType` is `divided` and `replicaDivisionPreference` is `weighted`.

vie-serendipity: > In addition, what if the total replica can't meet each selected cluster has a minReplicas? For example, the scheduler trying to assign 1 replica to two clusters.

This work should be done in the propagationpolicy validating webhook.
```
Webhook
Add new validation logic in the webhook to ensure that:

The sum of minReplicas across all clusters does not exceed the total replicas.
```

docs/proposals/scheduling/minreplicas/README.md
vie-serendipity: > I guess what you want to say High Available sceanrio instead of failover, in your use case you probably don'
t Karmada to migrate your applications when clusters become unavailable.

As I understand, these two are inseparable. I mean, when there are clusters A and B, and users choose weighted and dynamic, but cluster B has fewer resources, resulting in no replicas being allocated to it. In this case, the application is actually single-cluster deployed and not highly available. (As a user, I want to use dynamic feature and wish my application is high Available. It's unreasonable to adjust my cluster resources to achieve this.)

If cluster A fails at this point, there would be no active replicas, and the business would be impacted. And if cluster B has a `MinReplicas` guarantee, then during this period (cluster A fails), it can handle the traffic.

vie-serendipity:
> So, I'd prefer to start by analyzing the current practices (static weight, dynamic weight), and then propose solutions for it.

Agree. We should confirm that current implementations lack some abilities and then determine the necessity of introducing MinReplicas.

docs/proposals/scheduling/minreplicas/README.md
RainbowMango: I don't think karmada-webhook can do that, the number of clusters is identified during the scheduling process, not the creation of PropagationPolicy.

Anyway, will see after we clarify the use case and the gap.

RainbowMango:
> > So, I'd prefer to start by analyzing the current practices (static weight, dynamic weight), and then propose solutions for it.
> 
> Agree. We should confirm that current implementations lack some abilities and then determine the necessity of introducing MinReplicas.

Yes, exactly! Usually, I don't try to find a solution before I fully understand the use case and the gap.

vie-serendipity:
First make sure fine-tuning the scheduler algorithm is available, specifically checking whether it would impact the existing replicas distribution of existing Karmada users when they upgrade to a new version. The design principle is to not affect the distribution of workload in the legacy clusters.

Later on, consider some other solutions (TBD):
- Use the `preferredWeightList` to refactor the current static and dynamic preferences. 
- Introduce new fields like this proposal.
- Use feature-gate to control the breaking change to existing users.

BTW, the current webhook of this proposal solution can't address the `MinReplicas` validation when pp is created first.

/cc @RainbowMango

RainbowMango:
/close
Postpone it as the discussion at https://github.com/karmada-io/karmada/issues/6511#issuecomment-3561744184.

Issue #5115, opened 2024-07-01
[Feature] Multi-components workload scheduling


RainbowMango:
**What would you like to be added**:
Provide fine-grained support for workloads which usually consist of two or more components, like:

- Kubeflow Training/Big-data workloads
  * [TFJob](https://github.com/kubeflow/training-operator/blob/2b39d3cbc1525dc29eb40aa58b149dbeef00aa0f/pkg/apis/kubeflow.org/v1/tensorflow_types.go#L50)(Tensorflow Job), which may consist of multiple components like `PS`, `Worker`, `Chief`, `Master`, and `Evaluator`.
  * [PyTorchJob](https://github.com/kubeflow/training-operator/blob/2b39d3cbc1525dc29eb40aa58b149dbeef00aa0f/pkg/apis/kubeflow.org/v1/pytorch_types.go#L56) which may consist of two components, like `Master` and `Worker`.
  * [MXJob](https://github.com/kubeflow/training-operator/blob/2b39d3cbc1525dc29eb40aa58b149dbeef00aa0f/pkg/apis/kubeflow.org/v1/mxnet_types.go#L112), which may consist of multiple components, like `Scheduler`, `Server`, `Worker`, `TunerTracker`, `TunerServer`, and `Tuner`.
  * [XGBoostJob](https://github.com/kubeflow/training-operator/blob/2b39d3cbc1525dc29eb40aa58b149dbeef00aa0f/pkg/apis/kubeflow.org/v1/xgboost_types.go#L66), which may consist of two components, like `Master` and `Worker`.
  * [MPIJob](https://github.com/kubeflow/training-operator/blob/2b39d3cbc1525dc29eb40aa58b149dbeef00aa0f/pkg/apis/kubeflow.org/v1/mpi_types.go#L51C6-L51C13), which may consist of two components, like `Launcher` and `Worker`.
  * [PaddleJob](https://github.com/kubeflow/training-operator/blob/2b39d3cbc1525dc29eb40aa58b149dbeef00aa0f/pkg/apis/kubeflow.org/v1/paddlepaddle_types.go#L56), which may consist of two components, like `Master` and `Worker`.
  * [SparkApplication](https://github.com/kubeflow/spark-operator/blob/16cd35a0a220da0304383517c73a5fdb36f4213c/pkg/apis/sparkoperator.k8s.io/v1beta2/types.go#L172), which may consist of two components, like `driver` and `executor`.
[...]

Issue #6518, opened 2025-07-08
Failed over workloads should be balanced amongst feasible clusters


mszacillo:
**What would you like to be added?**

We use Karmada's cluster failover mechanism to handle DR (disaster recovery) scenarios using taint-based evictions. However, we have observed that during failover rescheduling, all workloads tend to be placed on the same destination cluster, even if multiple clusters are available and eligible.

**Example scenario:**
Assume we have three clusters: A, B, and C.
- We have 15 applications (each with 1 replica) running on Cluster A.
- We simulate a DR event by adding taints to Cluster A.
- This triggers a taint-based eviction, and Karmada begins rescheduling the 15 apps to other available clusters.

With Karmada’s scheduler estimators, we get the following replica availability from the feasible clusters:

> Cluster B: 10 available replicas
> Cluster C: 5 available replicas

Ideally, we would want Karmada to distribute the 15 applications across B and C (e.g., 10 to B, 5 to C). However, in practice, Karmada schedules all 15 applications to Cluster B.

We believe this happens because rescheduling happens in rapid succession, and the scheduler estimators results are not updated between each decision. As a result, each scheduling decision sees the same stale replica estimation (10 on B, 5 on C), and all pods are assigned to Cluster B without realizing it's filling up.

**Why is this needed?**

This enhancement would:
- Improve load balancing across clusters during failovers
- Ensure more accurate and up-to-date resource usage estimation
- Prevent Karmada from overcommitting a single cluster beyond its capacity

Right now, Karmada may unintentionally place more workloads on a cluster than it can actually handle, even when other clusters are available. Improving how estimations are refreshed or distributed during bulk rescheduling would significantly improve resilience and fairness.

PR #6535, opened 2025-07-15
Refine Multiple Pod Templates Scheduling proposal

RainbowMango:
**What type of PR is this?**

/kind feature

Update Multiple Pod Templates Scheduling Proposal as per [the proposal template](https://github.com/karmada-io/karmada/tree/master/docs/proposals/proposal-template).

Part of #5115

I think this feature can roughly be divided into 3 parts:
1. Extension of the Resource Interpreter Framework, like introducing a new interface to parse resource requirements from CRD with multiple pod templates
2. Extension of the ResourceBinding API, to hold the resource requirements
3. Extension of the Scheduler(include estimator) to make accurate estimation.

~~This PR hasn't yet detailed these three parts specifically; it simply adds things like globals/user story content for now.~~

```release-note
NONE
```


RainbowMango:
cc @mszacillo Could you please take a look?

docs/proposals/scheduling/multi-podtemplate-support/multiple-pod-template-support.md
```diff
@@ -1,141 +1,111 @@
...
-multiple pods, and the JM and TM frequently do not have the same resource requirements.
+We propose to extend the scheduling logic and the Resource Interpreter Framework to support a list of `ComponentRequirements`, each specifying the name, replica count, and resource requirements for a component. The scheduler estimator will be enhanced to determine whether a target cluster can accommodate a workload with multiple pod templates (components), considering the feasibility of scheduling all components together. Additionally, the estimator will provide context and metrics that can be used for scoring clusters during scheduling, rather than simply summing resource requirements or calculating how many full CRDs fit. The estimator's logic will be improved to reflect the complexity of multi-component workloads and to support more informed scheduling decisions.
 
-## Motivation
+### User Stories (Optional)
+
+#### Story 1
+
+As a user, I want to schedule a FlinkDeployment CRD with both JobManager and TaskManager components, each with different resource requirements, and have Karmada accurately estimate if the deployment can fit on a target cluster.
```
RainbowMango, line 46: @seanlaii 
Does this story apply to RayJob and RayCluster as well? As you mentioned at https://github.com/karmada-io/karmada/pull/5085#issuecomment-3057685260.

seanlaii: Thank you so much! I will take a look tonight!

RainbowMango:
Force pushed to add the following design details:
- Resource Interpreter Framework Extension
- Feature Gate: MultiplePodTemplatesScheduling

docs/proposals/scheduling/multi-podtemplate-support/multiple-pod-template-support.md
```diff
@@ -1,141 +1,129 @@
...
 
-	// Namespace represents the resources namespaces
-	// +optional
-	Namespace string `json:"namespace,omitempty"`
+// A component is a unique representation of a resource's replica.
+type ComponentRequirements struct {
+    // Name of this component
+    Name string `json:"name,omitempty"`
 
-	// PriorityClassName represents the components priorityClassName
```
mszacillo, line 131: Just to clarify, with this change the ComponentRequirements will be added to ResourceBinding spec by default, even if the number of components is 1? I think this would make sense, as the interpreter would default to getReplicas and we can just populate the new ComponentRequirements fields using this single component.

docs/proposals/scheduling/multi-podtemplate-support/multiple-pod-template-support.md
```diff
@@ -1,141 +1,129 @@
...
+    Replicas int32 `json:"replicas,omitempty"`
 
+    // ReplicaRequirements represents the requirements required by each replica for this component.
+    // +optional
+    ReplicaRequirements *ReplicaRequirements `json:"replicaRequirements,omitempty"`
 }
 ```
 
-### Overview of Code Changes
+### 3. Scheduler Changes
```
mszacillo, line 235: More of an implementation detail, but did you have any opinions on the scheduler estimator changes?

docs/proposals/scheduling/multi-podtemplate-support/multiple-pod-template-support.md
```diff
@@ -1,141 +1,129 @@
 ---
-title: CRD Component Scheduler Estimation
+title: Multiple Pod Templates Scheduling
 authors:
   - "@mszacillo"
```
mszacillo, line 4: @RainbowMango You should definitely add yourself as an author here. :)

mszacillo:
Just read through the new suggested changes. Thanks for condensing this proposal and making it easier to follow! Left some comments but I don't have major concerns w.r.t the API changes.

docs/proposals/scheduling/multi-podtemplate-support/multiple-pod-template-support.md
RainbowMango: Good questions! Actually, I haven't gotten a chance to revisit this part yet; that's the next move. I mean the ResourceBinding changes as well as deal with the relationship between the new ComponentRequirements and the legacy fields.

A draft idea in my mind is that the legacy fields `Replicas` and `ReplicaRequirements` will co-exist with the new `ComponentRequirements` for a while. The legacy field will be populated as usual, but in case of the feature gate is enabled, the new and legacy fields will be both populated. 

The new ComponentRequirements field will replace the legacy fields eventually, once the new feature becomes mature.

Will think and explore more about this.

docs/proposals/scheduling/multi-podtemplate-support/multiple-pod-template-support.md
RainbowMango: Yeah, I'm hesitant to change this part yet as I'm not quite sure I understand it.  But indeed, I have another idea about how to implement the estimator, as I talked to you at Hong Kong during the KubeCon China 2025.

This approach follows the existing estimator interface `MaxAvailableReplicas`, The draft idea in my mind is we can do that by introducing another interface like `ComponentsAccommodationEvaluate`(placeholder name), which can tell if a specific cluster can hold the workload with multiple components. It will only be involved in a limited cases, when scheduling a workload with multiple components and only requires one cluster.

docs/proposals/scheduling/multi-podtemplate-support/multiple-pod-template-support.md
RainbowMango: :)
OK.

I'm surprised you get up so early, I remember you are in the timezone UTC-4.

mszacillo: I'm working from Poland this week, so my working hours are a bit earlier now haha

docs/proposals/scheduling/multi-podtemplate-support/multiple-pod-template-support.md
seanlaii: Thanks! This story applys to RayJob and RayCluster.

docs/proposals/scheduling/multi-podtemplate-support/multiple-pod-template-support.md
seanlaii: Just to clarify, the current design will not calculate the actual max replica of each node, is that correct? Thanks!

docs/proposals/scheduling/multi-podtemplate-support/multiple-pod-template-support.md
RainbowMango: Oh, I see. That explains it.

docs/proposals/scheduling/multi-podtemplate-support/multiple-pod-template-support.md
RainbowMango: After looking at the API definition of [RayJob](https://github.com/ray-project/kuberay/blob/b0ddad8fa5c7c46094ee8d8003075db42bd73c44/ray-operator/apis/ray/v1alpha1/rayjob_types.go#L45)and [RayCluster](https://github.com/ray-project/kuberay/blob/b0ddad8fa5c7c46094ee8d8003075db42bd73c44/ray-operator/apis/ray/v1alpha1/raycluster_types.go#L186), I just raised another question: what's the relation between them?
It seems a RayJob consists of a `SubmitterPod` and a `RayCluster`, and a `RayCluster` contains a `HeadGroup` and `WorkerGroup`

Does that mean RayJob consists of 3 pod templates?

docs/proposals/scheduling/multi-podtemplate-support/multiple-pod-template-support.md
RainbowMango: Yes, this proposal aims to resolve the issue.

docs/proposals/scheduling/multi-podtemplate-support/multiple-pod-template-support.md
RainbowMango: Update: Please see the latest changes. `API changes part`.

RainbowMango:
Force pushed for updating the `API changes` part.

I will then look into the 3 part `Scheduler changes`. Hopefuly I can finish it before next community meeting then we can discuss it about.

docs/proposals/scheduling/multi-podtemplate-support/multiple-pod-template-support.md
seanlaii: Yes, sometimes it might consist of more than 3 pod templates. To be more specific,
`RayCluster`: Defines a Ray computing environment, composed of a `HeadGroup` for orchestration and **multiple** `WorkerGroups`. Its key feature is enabling heterogeneous node configurations (e.g., CPU, GPU) within a single cluster to support diverse workloads.

`RayJob`: A higher-level abstraction for executing a batch job. It automates the process by provisioning a `RayCluster` and launching a `SubmitterPod` responsible for submitting the workload to that cluster.

RainbowMango: Oh, I see. Thanks for the clarification. 
Added a note to the proposal:

**Note:** This proposal applies to a wide range of multi-component workloads, such as [RayJob](https://github.com/ray-project/kuberay/blob/b0ddad8fa5c7c46094ee8d8003075db42bd73c44/ray-operator/apis/ray/v1alpha1/rayjob_types.go#L109), [TensorFlowJob](https://github.com/kubeflow/trainer/blob/2b39d3cbc1525dc29eb40aa58b149dbeef00aa0f/pkg/apis/kubeflow.org/v1/tensorflow_types.go#L50), and other workload types listed in [Karmada issue #5115](https://github.com/karmada-io/karmada/issues/5115), including PyTorchJob, MXJob, XGBoostJob, MPIJob, PaddleJob, SparkApplication, TrainJob, Volcano Job, and more. These workloads typically consist of multiple components, each with their own resource requirements, and benefit from fine-grained, multi-component scheduling support.

seanlaii: cc @zclyne for taking a look, as he is also interested in the proposal.

seanlaii: cc @CaesarTY for taking a look as well, as he might also need this for `SparkJob`.

RainbowMango: Yeah, SparkJob might be special, as far as I know, the replica number is sometimes not provided in the resource template.

RainbowMango:
Force pushed for making room for different alternatives, especially for `Scheuler Changes`.

@mszacillo I've made some changes to your content and added a summary, but I didn't intend to alter your original meaning. Please let me know if you find anything wrong.

zclyne:
Thanks @seanlaii for involving me. This proposal looks good and I would be glad to help with any implementation requests. Thanks

mszacillo:
@RainbowMango Just read through your changes. Looks great, thanks for summarizing.

RainbowMango:
The proposal is ready for review. cc guys who might be interested in this
@likakuli @Garrybest @kevin-wangzefeng @zclyne @seanlaii @likku123 @KunWuLuan @Monokaix

liaolecheng:
Hey, just wondering — does `ResourceBinding.spec.clusters` need to change for multi-podTemplate workloads?

docs/proposals/scheduling/multi-podtemplate-support/multiple-pod-template-support.md
```diff
@@ -128,37 +146,147 @@ type ReplicaRequirements struct {
...
-	- We could run a maxReplica estimation for each component as is - the difficulty is determining if both components can be scheduled on the same cluster. If we maintain a maxReplica estimate for each component, not only is the estimation more complex, but it is possible to run into edge cases where both components cannot fit on the same cluster even though individually they could be scheduled.
-	- Once the MaxReplica estimation is complete, we will return the unit to replicas by multiplying the result with `totalReplicas` field.
-
-Given the above assumptions, we will describe the maxReplica estimation in two parts:
+We will describe the maxReplica estimation in two parts:
 
 1. The accurate estimator will create an estimate for the maxReplicas that can be scheduled. This estimate will depend on the amount of components set for the resource. If the number of components is 1, then the estimation will be done in the same way that it's done today. If the number of components is greater than 1, then we will sum up all resources required by all replicas to see how many of the total CRD can fit in the available resources.
```
mszacillo: _Below I've included a more concise and clear explanation of this estimation logic:_ 

The proposed enhancement introduces a more accurate estimation method (maxReplicas) for determining how many replicas of a CRD can be scheduled on a target cluster, particularly when a CRD includes multiple components.

Currently, the estimation logic assumes each CRD has a single component. This enhancement adjusts the logic to handle multi-component CRDs more accurately by summing the resource requirements across all components.

Let:
- `AvailableCPU` = cluster’s available CPU (from ResourceQuota)
- `AvailableMemory` = cluster’s available memory (from ResourceQuota)
- `TotalReplicas` = sum of all component replicas defined in the CRD

**Case 1: Single-component CRDs (componentCount = 1)**

This follows existing behavior:

```
maxReplicas = min(AvailableCPU / replicaCPU, AvailableMemory / replicaMemory)
```

**Case 2: Multi-component CRDs (componentCount > 1)**

In this case, we compute the total resources required by the entire CRD and estimate how many full instances of the CRD can be scheduled:

```
maxCRDs = min(AvailableCPU / totalCRDCPU, AvailableMemory / totalCRDMemory)
maxReplicas = maxCRDs * TotalReplicas
```

Where:
- `totalCRDCPU` = sum of (replicas × cpu) for each component
- `totalCRDMemory` = sum of (replicas × memory) for each component

Note: We multiply by TotalReplicas to convert the number of schedulable CRDs into the equivalent replica count, since Karmada interprets results in units of replicas, not CRD instances or sets.

**Example**
Let's assume:
- TotalReplicas = 3
- Cluster ResourceQuota: 6 CPUs, 8Gi Memory
- CRD has two components:
```yaml
component_1:
  replicas: 1
  cpu: 1
  memory: 2Gi
component_2:
  replicas: 2
  cpu: 1
  memory: 1Gi
```

**Step 1: Calculate total resource requirements of one CRD**

```
totalCRDCPU = (1 * 1) + (2 * 1) = 3 CPUs
totalCRDMemory = (1 * 2Gi) + (2 * 1Gi) = 4Gi
```

**Step 2: Estimate how many full CRDs can be scheduled**

```
maxCRDs = min(6 CPUs / 3 CPUs, 8Gi / 4Gi) = min(2, 2) = 2
```

**Step 3: Convert to replica count**

```
maxReplicas = 2 CRDs * 3 replicas = 6 replicas
```

docs/proposals/scheduling/multi-podtemplate-support/multiple-pod-template-support.md
seanlaii: cc @tessapham as well.

docs/proposals/scheduling/multi-podtemplate-support/multiple-pod-template-support.md
zhzhuang-zju: @mszacillo Thank you for your detailed explanation. Part 1 is now very clear and I fully agree with this calculation process and algorithm. However, I have some questions about Part 2 `Bin Packing Verification Step`:

- The `first-fit decreasing algorithm` appears to require the replicas of components, as well as the CPU, memory and other resource information for each component's replicas, but the current Estimator Interface `MaxAvailableReplicas` lacks this information. So does Part 2 need to extend the Estimator Interface?

- Part 2 can verify whether Part 1's results are feasible, but Part 1's results might be overestimated. For example, a cluster might appear capable of hosting 3 full CRD sets based on available resources, but due to node resource fragmentation, it can actually only host 2 full CRD sets. Following this approach, the final result might be 0 because it fails the verification step. Please correct me if my understanding is wrong.

RainbowMango: @mszacillo Updated this to the proposal. Please take a look.

mszacillo: @RainbowMango Looks good, thank you!

@zhzhuang-zju We may need to include a new method in the EstimateReplicasPlugin interface to include the new ComponentRequirements field, as it currently only supports replicaRequirements:

```golang
type EstimateReplicasPlugin interface {
	Plugin
	// Estimate is called for each MaxAvailableReplicas request.
	Estimate(ctx context.Context, snapshot *schedcache.Snapshot, replicaRequirements *pb.ReplicaRequirements) (int32, *Result)
}
```

> Part 1's results might be overestimated. For example, a cluster might appear capable of hosting 3 full CRD sets based on available resources, but due to node resource fragmentation, it can actually only host 2 full CRD sets.

You're right that the resource-based estimation in Part 1 can lead to an overestimate. However, in practice, we’re only ever scheduling one instance of the CRD to the chosen cluster, so we’re primarily interested in whether at least one full CRD set can be placed. In that sense, even if the estimate is high, what really matters is whether the bin-packing verification confirms capacity for a single CRD. That said, if we're scheduling a bunch of CRDs one after another I think we'd run into this issue: https://github.com/karmada-io/karmada/issues/6518

But anyways, the aim here was to just treat the verification step as authoritative — and if that fails (even for one), the final estimate should be 0. I agree this isn’t ideal, but it works for our current needs and ensures we don’t incorrectly schedule workloads that can’t be placed.

zhzhuang-zju: > We may need to include a new method in the EstimateReplicasPlugin interface to include the new ComponentRequirements field, as it currently only supports replicaRequirements:

Yes, I think this should be necessary. However, we need to pay attention to whether this modification will affect the `MaxAvailableReplicas` interface, since the input parameters of the `EstimateReplicasPlugin` interface will likely need to come from the `MaxAvailableReplicas` interface.

> However, in practice, we're only ever scheduling one instance of the CRD to the chosen cluster, so we're primarily interested in whether at least one full CRD set can be placed.

Got it. So the `Bin Packing Verification Step` is not validating the results of `Part 1`, but rather verifying whether the cluster's nodes can accommodate the entire CRD.

Garrybest:
Hi @RainbowMango, it's glad to see this proposal. It's really interesting because I have thought about this idea and maken an internal version in my production environment two years ago. At that time, I focused on AI jobs such as Pytorch/TF Job, scheduling them to different GPU clusters.

I thinks the most important thing is the algorithm of estimator. This proposal have an idea based on **set**. A set is a sort of scheduling unit for estimator to **place** them as many as possible. And that's almost the same as what I thought before as well.

The only difference between **Alternative 2** and **Alternative 3** is the definition of interface. **Alternative 3** is definitely similar with my previous version. A brandly new interface could have a distinct meaning without ambiguity, and make it easier to handle and implement.

Another solution is that we don't need to calculate the max set, we only cares if a cluster could accomodate this Job. Generally, this kind of job is likely to run in a single cluster, typically GPU jobs. Given this, the estimator only returns true/false to notify scheduler whether this job could be accomodated in this cluster.

RainbowMango:
So glad to hear from you! Thanks for the time! 
I can get from above, you prefer the `Alternative 3` (introduce a brand new gRPC interface). 

> Another solution is that we don't need to calculate the max set, we only cares if a cluster could accomodate this Job. Generally, this kind of job is likely to run in a single cluster, typically GPU jobs. Given this, the estimator only returns true/false to notify scheduler whether this job could be accomodated in this cluster.

Yeah, the estimator only returns `true` or `false` can meet most of the requirements as of now.  The reason why calculate the maximum sets is that it makes room for future extension, like the schedule can sort clusters by their available capacity. Just as I mentioned at https://github.com/karmada-io/karmada/pull/6535/files#diff-5a2cc3ff58bf5394fffdf2c4be3c263a1fe4003da4c2a3df4cae9cbd9d071582R407-R408.  Also, echo here:
> - **Resource-Aware Scheduling Decisions:** By knowing the exact number of full CRD sets a cluster can accommodate, the scheduler can make more informed decisions. For example, it can prioritize clusters with higher available capacity, leading to better resource utilization and balancing across clusters.
> - **Cluster Sorting and Ranking:** With `maxSets`, the scheduler can sort clusters by their available capacity for the workload, enabling strategies such as bin-packing, spreading, or prioritizing clusters with the most headroom.

But thanks for the reminder, I guess you are concerned that calculating the maximum sets would slow down the estimation process.  So, given we don't have clear evidence that shows the `future extension` is necessary, we can let the estimator return immediately once it schedules the first `set`. That means the `maxSet` is either `0` or `1`.  Does that make sense?

Garrybest:
> But thanks for the reminder, I guess you are concerned that calculating the maximum sets would slow down the estimation process. So, given we don't have clear evidence that shows the `future extension` is necessary, we can let the estimator return immediately once it schedules the first `set`. That means the `maxSet` is either `0` or `1`. Does that make sense?

Exactly. This is a really good idea! Keep the `maxSet` and just return 0 or 1, enhance it if necessary in the future.

RainbowMango:
@mszacillo @seanlaii @likakuli any comments? which alternative do you prefer?

mszacillo:
> @mszacillo @seanlaii @likakuli any comments? which alternative do you prefer?

Ah! Apologies for delay. I think based on conversations we had in the last community meeting, I'm personally leaning towards approach 2, but I could be convinced of approach 3 as well. To me it feels like a trade-off between maintaining backwards compatibility and code simplicity vs. having a clear separation of concerns.  

Generally I think the core estimation logic itself will be quite similar between single and multi-component workloads, so if the main differential here is the request type, I'm not sure if this warrants supporting a completely separate interface (yet). Perhaps we could start with approach 2, and after deprecating the legacy path, separate out to a different interface if needed.

docs/proposals/scheduling/multi-podtemplate-support/multiple-pod-template-support.md
```diff
@@ -128,59 +146,210 @@ type ReplicaRequirements struct {
...
-	- This simplifies the scope of the problem, and accounts for the fact that it is  non-trivial to schedule components of the same CRD across multiple clusters.
+Currently, the estimation logic assumes each CRD has a single component. This enhancement adjusts the logic to handle multi-component CRDs more accurately by summing the resource requirements across all components.
 
-`Assumption 2`: MaxReplica estimation will use a sum of all the resource requirement for every component's replica.
-	- We could run a maxReplica estimation for each component as is - the difficulty is determining if both components can be scheduled on the same cluster. If we maintain a maxReplica estimate for each component, not only is the estimation more complex, but it is possible to run into edge cases where both components cannot fit on the same cluster even though individually they could be scheduled.
-	- Once the MaxReplica estimation is complete, we will return the unit to replicas by multiplying the result with `totalReplicas` field.
+Let:
+- `AvailableCPU` = cluster’s available CPU (from ResourceQuota)
+- `AvailableMemory` = cluster’s available memory (from ResourceQuota)
```
RainbowMango, line 291: 1. I'm not sure where the ResourceQuota comes from, or maybe the FederatedResourceQuota, anyway. It feels like a hard restriction that the user must use ResourceQuota or FederatedResourceQuota. In addition, it requires that every resource type be explicitly declared in the quota configurations. It might be challenging.

2. If we are using the total available resources, we are going to suffer from the resource fragmentation issue. Like each node left a small piece of resource, which probably can not accommodate even 1 pod, but it seems that the cluster can hold a whole CRD with multiple pods when looking at the total available resources, because it might have a bunch of nodes. 

I forgot to share my thoughts at the last community meeting about this approach. 
@mszacillo Please correct me if I misunderstood it.

RainbowMango:
> To me it feels like a trade-off between maintaining backwards compatibility and code simplicity vs. having a clear separation of concerns.

Yes, exactly. 

I lean toward alternative 3 since the new gRPC interface can fully replace the legacy one in the future. Another reason, like you mentioned, we can easily separate the original feature and the new one.

Like we talked at the community meeting, I will continue to evaluate the risk of alternative 2. If we lack confidence in the backward compatibility, we'll turn to alternative 3 instead.

RainbowMango:
@seanlaii Do you insist on alternative 2?

seanlaii:
Sorry for the delay. 
I can also go with approach 3. It makes sense to have a separate interface which leads to a clearer and more maintainable implementation.

RainbowMango:
Thank you all for the comments. 
I've opened an issue(#6641) for tracking the iteration tasks.

docs/proposals/scheduling/multi-podtemplate-support/multiple-pod-template-support.md
```diff
@@ -1,121 +1,139 @@
...
-These basic fields are necessary to allow the accurate estimator to determine whether all components of the CRD replica
-will be able to fit on the destination namespace.
+	// GetComponentReplicas extracts the resource requirements for multiple components from the given object.
+	// This interpreter hook is designed for CRDs with multiple components (e.g., FlinkDeployment), but can
+	// also be used for single-component resources like Deployment.
+	// If implemented, the controller will use this hook to obtain per-component replica and resource
+	// requirements, and will not call GetReplicas.
+	// If not implemented, the controller will fall back to GetReplicas for backward compatibility.
+	// This hook will only be called when the feature gate 'MultiplePodTemplatesScheduling' is enabled.
+	GetComponentReplicas(object *unstructured.Unstructured) (components []workv1alpha2.ComponentRequirements, err error)
```
likakuli, line 82: In the current design, adding `GetComponentReplicas` directly to ResourceInterpreter requires all implementers, including those that only support single-component resources (e.g., Deployment), to implement this method. This introduces two main concerns:

- Redundant code: implementers may have to provide placeholder implementations (e.g., returning nil).

- Increased maintenance overhead: any future changes to the method signature would unnecessarily affect all implementers.

  
Perhaps we could consider defining a dedicated MultiComponentInterpreter interface instead
```
type MultiComponentInterpreter interface {
    GetComponentReplicas(obj *unstructured.Unstructured) ([]workv1alpha2.ComponentRequirements, error)
}
```

This approach ensures that only interpreters requiring multi-component support implement the method, thereby reducing boilerplate code, improving maintainability, and isolating future changes.

RainbowMango: I don't think introducing a new method to the Resource Interpreter Interface would force every resource to implement it. 
The `GetComponentReplicas` is totally optional for all resource types. 

As of now, we have introduced the interface, but did not implement this method for any resource except `FlinkDeployment`, everything works fine so far. 

But thanks for the input, your concerns occur if we decide to deprecate `GetReplicas` method in the future. Currenly no such plan yet.

RainbowMango:
PS: 
I will send another PR for the scheduler part design and sync the changes during implementation.

Issue #6516, opened 2025-07-07
[Umbrella] Performance optimizations planned for v1.15


zach593:
### Features that can be added before the v1.15 release deadline

1. Since we have disabled `deepcopy` during `list` operations in [#5813](https://github.com/karmada-io/karmada/pull/5813) , we should enable the `MutationDetector` in our e2e tests to prevent any future code from mutating the cache.
    - [x] Enable mutation detection in E2E testing. (@XiShanYongYe-Chang, #6541)
    - [x] detector: eliminate the mutation of the informer cache. (@zhzhuang-zju, #6544)

2. Introduce the [controller-runtime priority queue](https://github.com/kubernetes-sigs/controller-runtime/issues/2374). This enhancement allows controllers built on controller-runtime to prioritize the latest changes after a restart or leader transition, deferring the processing of inevitable create events. This can significantly reduce downtime during service restarts and failovers.   * For this feature, we will likely need detailed test reports to ensure it does not introduce negative side effects.
    - [x] feat: introduce ControllerPriorityQueue feature gate. (@zach593, #6662)

3. Scheduler Performance enhancement
    - [x] Improve the performance of karmada-scheduler. (@LivingCcj, #6597)

### Features that may progress before the v1.15 release deadline but are unlikely to be completed

1. Where feasible, migrate `async-worker` to use `controller-runtime` controllers. #6251

### Features unlikely to start before the v1.15 release deadline but are likely to be pursued in the future

1. Add automated e2e benchmark tests in GitHub CI.
2. Introduce controller-runtime warmup. This feature is still under development https://github.com/kubernetes-sigs/controller-runtime/pull/3192. Combined with the priority queue, it could theoretically reduce downtime to zero during multi-replica leader transitions,  which is especially important in large-scale clusters.
3. Continue adding indices to improve `list` performance wherever applicable. 
    - [ ] [Umbrella] Use index to optimize list operations https://github.com/karmada-io/karmada/issues/6530
---

The above list includes what I am currently aware of. If there are any omissions, please feel free to add or suggest them.

zach593:
I will create separate issues for each task this week.

RainbowMango:
Great thanks @zach593 for your hard work~ Take care!

RainbowMango:
Speaking of the flags in `karmada-controller-manager`, like:
```
      --concurrent-cluster-propagation-policy-syncs int                                                                                                                                                                      
[...]

Issue #6513, opened 2025-07-04
ClaimMetadata not properly cleaned on deletion of Lazy activationPreference propagationpolicy


zhzhuang-zju:
**What happened**:
When I first deploy a Deployment and then deploy a PropagationPolicy with `Lazy` activationPreference, no corresponding ResourceBinding is generated. This is normal. 
However, if I then delete this PropagationPolicy, the Deployment still retains the PropagationPolicy-related labels and metadata. Subsequently, when I deploy a new PropagationPolicy with a non-`Lazy` activationPreference, the Deployment fails to match this newly created PropagationPolicy.

**What you expected to happen**:
When a PropagationPolicy with `Lazy` activationPreference is deleted, the ClaimMetadata should be properly cleaned up so that it does not prevent the Resource Template from binding to a new PropagationPolicy.

**How to reproduce it (as minimally and precisely as possible)**:
1. Set up a Karmada environment:
```bash
$ hack/local-up-karmada.sh
```

2. Deploy a Deployment:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx
  labels:
    app: nginx
spec:
  replicas: 2
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - image: nginx
        name: nginx
```

3. Deploy a PropagationPolicy with `Lazy` activationPreference:
```yaml
apiVersion: policy.karmada.io/v1alpha1
kind: PropagationPolicy
metadata:
  name: nginx-propagation
spec:
  activationPreference: Lazy
  resourceSelectors:
  - apiVersion: apps/v1
    kind: Deployment
  placement:
    clusterAffinity:
      clusterNames:
      - member1
      - member2
    replicaScheduling:
      replicaDivisionPreference: Weighted
      replicaSchedulingType: Divided
      weightPreference:
        staticWeightList:
        - targetCluster:
            clusterNames:
            - member1
          weight: 1
        - targetCluster:
            clusterNames:
            - member2
          weight: 1
```
At this point, no ResourceBinding is generated, and the Deployment already has PropagationPolicy-related ClaimMetadata:
```yaml
  annotations:
    propagationpolicy.karmada.io/name: nginx-propagation
    propagationpolicy.karmada.io/namespace: default
  labels:
    app: nginx
    propagationpolicy.karmada.io/permanent-id: 537816cb-e95b-45ca-96c8-e7d5cfadae6e
```

[...]

PR #6541, opened 2025-07-16
enable mutation detection in e2e

XiShanYongYe-Chang:
**What type of PR is this?**

/kind cleanup

Since we have disabled deepcopy during list operations in https://github.com/karmada-io/karmada/pull/5813 , we should enable the MutationDetector in our e2e tests to prevent any future code from mutating the cache.

Part of #6516

We need to fully consider the potential impact of introducing `KUBE_CACHE_MUTATION_DETECTOR=true` in the testing process.

Once `KUBE_CACHE_MUTATION_DETECTOR=true` is enabled, we will see the following log:

https://github.com/kubernetes/kubernetes/blob/ce9ba818025ef28555bac9e25da917acec2ccb1d/staging/src/k8s.io/client-go/tools/cache/mutation_detector.go#L54

This appears to be alarming. Therefore, we need to ensure that our normal usage does not encounter such issues, just like the problems faced in https://github.com/kubernetes/kubernetes/issues/60854.

For Karmada CI E2E testing, the test environment is reclaimed after execution, so we do not need to worry about memory leaks in the test environment.

In addition, Kubernetes has further iteratively optimized this feature, significantly reducing the issue of memory leaks:

https://github.com/kubernetes/kubernetes/pull/85350/commits/81d05e91b5b5e653e20cc6ab620de08abfd0cbc2 and https://github.com/kubernetes/kubernetes/pull/88131,  this can further alleviate our concerns about using `KUBE_CACHE_MUTATION_DETECTOR=true` in E2E.

**Why didn't I enable this check for the karmada-agent component?**

The `karmada-agent` and `karmada-controller-manager` share the same codebase. The `karmada-controller-manager` is sufficient to identify issues related to memory modifications in the code, thus eliminating the need for redundant checks on the `karmada-agent` component.

```release-note
NONE
```


RainbowMango:
Any documentation for the Mutation Detector feature?

XiShanYongYe-Chang:
> Any documentation for the Mutation Detector feature?

I have not yet found the relevant functional introduction document.

XiShanYongYe-Chang:
During the execution of this test, it was discovered that the controller-manager woul paninnc. Here is an example of an error log:

```log
CACHE *unstructured.Unstructured[68] ALTERED!
(*unstructured.Unstructured)({                                                                                                         (*unstructured.Unstructured)({
  Object: (map[string]interface {}) (len=4) {                                                                                            Object: (map[string]interface {}) (len=4) {
    (string) (len=8) "metadata": (map[string]interface {}) (len=6) {                                                                       (string) (len=8) "metadata": (map[string]interface {}) (len=4) {
      (string) (len=3) "uid": (string) (len=36) "2bfcde52-81a6-4a23-9915-57c8a8d355ed",                                                      (string) (len=15) "resourceVersion": (string) (len=4) "2855",
      (string) (len=15) "resourceVersion": (string) (len=4) "2855",                                                                          (string) (len=17) "creationTimestamp": (string) (len=20) "2025-07-16T01:45:03Z",
      (string) (len=17) "creationTimestamp": (string) (len=20) "2025-07-16T01:45:03Z",                                                       (string) (len=4) "name": (string) (len=32) "system:test-clusterrole-x525v-01",
      (string) (len=6) "labels": (map[string]interface {}) (len=1) {                                                                         (string) (len=3) "uid": (string) (len=36) "2bfcde52-81a6-4a23-9915-57c8a8d355ed"
        (string) (len=48) "clusterpropagationpolicy.karmada.io/permanent-id": (string) (len=36) "30033a80-2606-4bf5-9569-994581cc1973"     },
      },                                                                                                                                   (string) (len=5) "rules": (interface {}) <nil>,
      (string) (len=11) "annotations": (map[string]interface {}) (len=1) {                                                                 (string) (len=4) "kind": (string) (len=11) "ClusterRole",
        (string) (len=40) "clusterpropagationpolicy.karmada.io/name": (string) (len=17) "clusterrole-x525v"                                (string) (len=10) "apiVersion": (string) (len=28) "rbac.authorization.k8s.io/v1"
      },                                                                                                                                 }
      (string) (len=4) "name": (string) (len=32) "system:test-clusterrole-x525v-01"                                                    })
    },
    (string) (len=5) "rules": (interface {}) <nil>,
    (string) (len=4) "kind": (string) (len=11) "ClusterRole",
    (string) (len=10) "apiVersion": (string) (len=28) "rbac.authorization.k8s.io/v1"
  }
})


panic: cache *unstructured.Unstructured modified

goroutine 1638 [running]:
k8s.io/client-go/tools/cache.(*defaultCacheMutationDetector).CompareObjects(0xc000ee5ae0)
        /root/go/src/github.com/karmada-io/karmada/vendor/k8s.io/client-go/tools/cache/mutation_detector.go:165 +0x58e
k8s.io/client-go/tools/cache.(*defaultCacheMutationDetector).Run(0xc000ee5ae0, 0xc0012902a0)
        /root/go/src/github.com/karmada-io/karmada/vendor/k8s.io/client-go/tools/cache/mutation_detector.go:109 +0x149
k8s.io/client-go/tools/cache.(*sharedIndexInformer).RunWithContext.(*Group).StartWithChannel.func3()
        /root/go/src/github.com/karmada-io/karmada/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:55 +0x1b
k8s.io/apimachinery/pkg/util/wait.(*Group).Start.func1()
        /root/go/src/github.com/karmada-io/karmada/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:72 +0x4c
created by k8s.io/apimachinery/pkg/util/wait.(*Group).Start in goroutine 1364
        /root/go/src/github.com/karmada-io/karmada/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:70 +0x73
```

zhzhuang-zju:
> > Any documentation for the Mutation Detector feature?
> 
> I have not yet found the relevant functional introduction document.

Logs from the `karmada-controller-manager component` can indicate whether any cache modifications have occurred

XiShanYongYe-Chang:
> Logs from the `karmada-controller-manager component` can indicate whether any cache modifications have occurred

Yes, it is. Unfortunately, the panic log does not directly indicate which specific line of code caused the issue.

zhzhuang-zju:
@XiShanYongYe-Chang While resolving #6513, I discovered that the detector performs an operation that modifies the cache. I will submit a PR to address this issue. And check whether it can pass this mutation detection after the fix.

zhzhuang-zju:
@XiShanYongYe-Chang In #6544, I mitigated the behavior that mutates the informer cache, and it appears from the logs that this issue has been resolved. However, I have some doubts about this:
The effect of the environment variable `KUBE_CACHE_MUTATION_DETECTOR` seems to be triggering a pod panic when a cache mutation is detected. However, since many of our e2e tests include failure retries, it's possible for the test to pass after the component restarts, thereby bypassing the mutation detection. For example, in [this GitHub Actions run](https://github.com/karmada-io/karmada/actions/runs/16315287250), only one e2e test failed while the other two passed, even though the controller-manager clearly restarted based on the logs.

XiShanYongYe-Chang:
> The effect of the environment variable `KUBE_CACHE_MUTATION_DETECTOR` seems to be triggering a pod panic when a cache mutation is detected. However, since many of our e2e tests include failure retries, it's possible for the test to pass after the component restarts, thereby bypassing the mutation detection. For example, in [this GitHub Actions run](https://github.com/karmada-io/karmada/actions/runs/16315287250), only one e2e test failed while the other two passed, even though the controller-manager clearly restarted based on the logs.

Your question is excellent, and it's something I've been pondering as well. 

I have two thoughts: 

1. The fact that E2E runs without issues doesn't necessarily mean that the `karmada-controller-manager` component hasn't restarted; 
2. A restart of the karmada-controller-manager component doesn't necessarily indicate that it was caused by a panic.

For the first point, perhaps we could introduce a check for component restarts, but as for the second point, I don't have any ideas at the moment.

zhzhuang-zju:
> For the first point, perhaps we could introduce a check for component restarts, but as for the second point, I don't have any ideas at the moment.

When a pod restarts, there are some indicators that can be used to identify the reason for the restart. For example, the Last State of the container.

When a pod restarts due to panic, it has the following characteristics:
- The Restart Count of the container is greater than 0
- The Reason of the Last State is "Error"
- The Exit Code of the Last State is 2

Although these characteristics are not unique to go panic, they can distinguish between oom and other types of restarts, reducing the risk of false positives.

```bash
$ kubectl describe pods --namespace karmada-system karmada-controller-manager-7b74766c6f-qlw72
Containers:
  karmada-controller-manager:
    Last State:     Terminated
      Reason:       Error
      Exit Code:    2
      Started:      Thu, 17 Jul 2025 16:36:54 +0800
      Finished:     Thu, 17 Jul 2025 16:38:15 +0800
    Ready:          True
    Restart Count:  1
```

XiShanYongYe-Chang:
Wait for #6544 
/gemini-review

zhzhuang-zju:
@XiShanYongYe-Chang #6544 has been merged, we can continue with this work now.

XiShanYongYe-Chang:
> @XiShanYongYe-Chang #6544 has been merged, we can continue with this work now.

Thanks, let me rebase the master branch.

zhzhuang-zju:
Great! All the cache mutation detection tests have passed successfully.

hack/run-e2e.sh
```diff
@@ -52,10 +52,26 @@ REPO_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
...
 
+TEST_RESULT="PASS"
+
 set +e
 ginkgo -v --race --trace --fail-fast -p --randomize-all ./test/e2e/suites/base -- --karmada-context=karmada-apiserver
 TESTING_RESULT=$?
 
+if [ $TESTING_RESULT -eq 0 ]; then
+  echo "E2E run successfully."
+  echo "Checking if karmada-controller-manager has restarted..."
```
zhzhuang-zju: It would be better to explain why add the component restarted check

hack/run-e2e.sh
```diff
@@ -52,10 +52,26 @@ REPO_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
...
+  POD_NAME=$(kubectl --context="${KARMADA_HOST_CLUSTER_NAME}" get pod -n karmada-system -l app=karmada-controller-manager -o=jsonpath='{.items[0].metadata.name}')
+  RESTART_COUNT=$(kubectl --context="${KARMADA_HOST_CLUSTER_NAME}" get pod -n karmada-system -l app=karmada-controller-manager -o=jsonpath='{.items[0].status.containerStatuses[0].restartCount}')
```
zhzhuang-zju: The component Karmada-controller-manager has two replicas, and checking only one of them may result in omissions.

hack/run-e2e.sh
```diff
@@ -76,4 +92,8 @@ ls -al "$ARTIFACTS_PATH"
 # Post run e2e for delete extra components
 "${REPO_ROOT}"/hack/post-run-e2e.sh
 
-exit $TESTING_RESULT
+if [ "$TEST_RESULT" == "FAIL" ]; then
+    exit 1
+else
+    exit 0
```
zhzhuang-zju: This approach would mask different error codes when e2e tests fail, resulting in information loss.

XiShanYongYe-Chang: Sorry, I didn't quite understand what you meant. Could you please explain it again?

hack/run-e2e.sh
XiShanYongYe-Chang: update~

hack/run-e2e.sh
XiShanYongYe-Chang: update~

hack/run-e2e.sh
zhzhuang-zju: My bad, I didn't explain clearly. The e2e test result might not be limited to just 0 and 1, although our current implementation may only have 0 and 1.

XiShanYongYe-Chang: Thanks, I get it.

hack/deploy-scheduler-estimator.sh
```diff
@@ -85,6 +88,7 @@ rm -rf "${TEMP_PATH}"
 TEMP_PATH=$(mktemp -d)
 cp "${REPO_ROOT}"/artifacts/deploy/karmada-scheduler-estimator.yaml "${TEMP_PATH}"/karmada-scheduler-estimator.yaml
 sed -i'' -e "s/{{member_cluster_name}}/${MEMBER_CLUSTER_NAME}/g" "${TEMP_PATH}"/karmada-scheduler-estimator.yaml
+sed -i'' -e "s/{{KUBE_CACHE_MUTATION_DETECTOR}}/${KUBE_CACHE_MUTATION_DETECTOR}/g" "${TEMP_PATH}"/karmada-scheduler-estimator.yaml
```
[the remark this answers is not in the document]
XiShanYongYe-Chang: @gemini-code-assist
Why do I feel like the readability has decreased?

hack/run-e2e.sh
```diff
@@ -66,7 +65,7 @@ if [ "$KARMADA_RUNNING_ON_KIND" = true ]; then
   kind export logs --name="$KARMADA_HOST_CLUSTER_NAME" "$ARTIFACTS_PATH/$KARMADA_HOST_CLUSTER_NAME"
 
   echo "Collecting $KARMADA_PULL_CLUSTER_NAME logs..."
-  mkdir -p "$ARTIFACTS_PATH/KARMADA_PULL_CLUSTER_NAME"
+  mkdir -p "$ARTIFACTS_PATH/$KARMADA_PULL_CLUSTER_NAME"
```
zhzhuang-zju: Nice catch!

hack/run-e2e.sh
```diff
@@ -76,4 +75,78 @@ ls -al "$ARTIFACTS_PATH"
 # Post run e2e for delete extra components
 "${REPO_ROOT}"/hack/post-run-e2e.sh
 
+# If E2E test failed, exit directly with the test result
```
zhzhuang-zju, line 82: When e2e fails, can we also continue to check component restart? Earlier exposure to risks may be better.

hack/run-e2e.sh
```diff
@@ -76,4 +75,78 @@ ls -al "$ARTIFACTS_PATH"
...
+# List of components to check (label=component_name)
+components=(
+  "app=karmada-controller-manager:karmada-controller-manager"
+  "app=karmada-descheduler:karmada-descheduler"
+  "app=karmada-metrics-adapter:karmada-metrics-adapter"
+  "app=karmada-scheduler:karmada-scheduler"
+  "app=karmada-search:karmada-search"
+  "app=karmada-scheduler-estimator-member1:karmada-scheduler-estimator-member1"
+  "app=karmada-scheduler-estimator-member2:karmada-scheduler-estimator-member2"
+  "app=karmada-scheduler-estimator-member3:karmada-scheduler-estimator-member3"
```
zhzhuang-zju: It would be better to avoid hard-coding here since the cluster name can be changed using variables.

zhzhuang-zju (commented):
Thanks again for your work!  Others LGTM

hack/run-e2e.sh
XiShanYongYe-Chang: If the E2E has already failed, why is it still necessary to check for component restarts?

zhzhuang-zju: For example, my new code submission not only brings functional issues but also adds memory modification operations. If the e2e fails and skips component checks, I may need to fix the functional issues before detecting the memory modification problem, requiring two e2e runs.

XiShanYongYe-Chang: I understand what you're saying, but I think this approach might actually lead to some misdirection in problem identification. Take the example you mentioned: if the issue is caused by a new function, it could also potentially lead to component restarts, such as errors caused by panics. 

In such cases, the component restart check might mislead us into thinking there's still a memory modification error (which is one of the main purposes of this check), thereby diverting more effort into investigating this aspect. 

Additionally, for E2E failures, unless they are sporadic, we need to check the component logs, which can also reveal whether the component has restarted.

Therefore, if the new feature simultaneously has logical errors and memory modification issues, since the component restart check results cannot distinguish between these two types of errors, I prefer to address them one by one.

hack/run-e2e.sh
XiShanYongYe-Chang: Update, can you help take a review again?

hack/run-e2e.sh
zhzhuang-zju: > Therefore, if the new feature simultaneously has logical errors and memory modification issues, since the component restart check results cannot distinguish between these two types of errors, I prefer to address them one by one.

That makes sense, too. Since this is a brand new check, we can adjust the process later if it is needed.

XiShanYongYe-Chang: I think it's feasible. If there are more inputs later on, we can further iterate and optimize it.

RainbowMango (approved):
/lgtm
/approve

PR #6544, opened 2025-07-16
detector: eliminate the mutation of the informer cache

zhzhuang-zju:
**What type of PR is this?**
/kind cleanup

When an object matches a PropagationPolicy, the detector adds metadata related to the PropagationPolicy to the object. To ensure the object used in subsequent binding generation is up-to-date, the function `ClaimPolicyForObject` directly modifies the object itself.
https://github.com/karmada-io/karmada/blob/c60a2e05f6e1f90fdfd3580a902bccd51bfdbff4/pkg/detector/detector.go#L695-L711

However, since the object is retrieved from the cache and no deep copy is made before modification, `ClaimPolicyForObject` ends up altering the informer cache.
https://github.com/karmada-io/karmada/blob/c60a2e05f6e1f90fdfd3580a902bccd51bfdbff4/pkg/detector/detector.go#L663-L670

Considering that the current implementation requires the caller to manually pass in a deep copy of the object when invoking ClaimPolicyForObject, which is not user-friendly and prone to cause similar issues again, the function was modified to perform a deep copy internally. This avoids modifying the original object. In addition, the updated objectCopy is returned to meet the needs of scenarios that require the latest object.

Fixes #

```release-note
`karmada-controller-manager`: Fixed the issue that the informer cache gets unexpectedly modified during usage,
```

### DIFF
--- a/pkg/detector/detector.go
+++ b/pkg/detector/detector.go
@@ -657,9 +657,6 @@ func (d *ResourceDetector) ApplyClusterPolicy(object *unstructured.Unstructured,
 }
 
 // GetUnstructuredObject retrieves object by key and returned its unstructured.
-// Any updates to this resource template are not recommended as it may come from the informer cache.
-// We should abide by the principle of making a deep copy first and then modifying it.
-// See issue: https://github.com/karmada-io/karmada/issues/3878.
 func (d *ResourceDetector) GetUnstructuredObject(objectKey keys.ClusterWideKey) (*unstructured.Unstructured, error) {
 	objectGVR, err := restmapper.GetGroupVersionResource(d.RESTMapper, objectKey.GroupVersionKind())
 	if err != nil {
@@ -689,7 +686,8 @@ func (d *ResourceDetector) GetUnstructuredObject(objectKey keys.ClusterWideKey) 
 		return nil, err
 	}
 
-	return unstructuredObj, nil
+	// perform a deep copy to avoid modifying the cached object from informer
+	return unstructuredObj.DeepCopy(), nil
 }
 
 // ClaimPolicyForObject set policy identifier which the object associated with.
--- a/pkg/detector/policy.go
+++ b/pkg/detector/policy.go
@@ -270,7 +270,6 @@ func (d *ResourceDetector) removeResourceClaimMetadataIfNotMatched(objectReferen
 		return false, nil
 	}
 
-	object = object.DeepCopy()
 	util.RemoveLabels(object, labels...)
 	util.RemoveAnnotations(object, annotations...)
 
--- a/pkg/util/helper/binding.go
+++ b/pkg/util/helper/binding.go
@@ -238,9 +238,6 @@ func RemoveOrphanWorks(ctx context.Context, c client.Client, works []workv1alpha
 }
 
 // FetchResourceTemplate fetches the resource template to be propagated.
-// Any updates to this resource template are not recommended as it may come from the informer cache.
-// We should abide by the principle of making a deep copy first and then modifying it.
-// See issue: https://github.com/karmada-io/karmada/issues/3878.
 func FetchResourceTemplate(
 	ctx context.Context,
 	dynamicClient dynamic.Interface,
@@ -266,12 +263,13 @@ func FetchResourceTemplate(
 		// fall back to call api server in case the cache has not been synchronized yet
 		klog.Warningf("Failed to get resource template (%s/%s/%s) from cache, Error: %v. Fall back to call api server.",
 			resource.Kind, resource.Namespace, resource.Name, err)
-		object, err = dynamicClient.Resource(gvr).Namespace(resource.Namespace).Get(ctx, resource.Name, metav1.GetOptions{})
+		objectFromAPIServer, err := dynamicClient.Resource(gvr).Namespace(resource.Namespace).Get(ctx, resource.Name, metav1.GetOptions{})
 		if err != nil {
 			klog.Errorf("Failed to get resource template (%s/%s/%s) from api server, Error: %v",
 				resource.Kind, resource.Namespace, resource.Name, err)
 			return nil, err
 		}
+		return objectFromAPIServer, nil
 	}
 
 	unstructuredObj, err := ToUnstructured(object)
@@ -280,7 +278,7 @@ func FetchResourceTemplate(
 		return nil, err
 	}
 
-	return unstructuredObj, nil
+	return unstructuredObj.DeepCopy(), nil
 }
 
 // FetchResourceTemplatesByLabelSelector fetches the resource templates by label selector to be propagated.


XiShanYongYe-Chang:
Thanks a lot
/assign

pkg/detector/detector.go
```diff
@@ -693,36 +695,40 @@ func (d *ResourceDetector) GetUnstructuredObject(objectKey keys.ClusterWideKey)
...
 	claimedID := util.GetLabelValue(object.GetLabels(), policyv1alpha1.ClusterPropagationPolicyPermanentIDLabel)
 	// object has been claimed, don't need to claim again
 	if claimedID == policyID {
-		return policyID, nil
+		return object, policyID, nil
 	}
 
-	AddCPPClaimMetadata(object, policyID, policy.ObjectMeta)
-	return policyID, d.Client.Update(context.TODO(), object)
+	objectCopy := object.DeepCopy()
```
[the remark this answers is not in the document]
zhzhuang-zju: @gemini-code-assist
Can you further explain? And does the current implementation meet your suggestions?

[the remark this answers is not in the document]
zhzhuang-zju: > Your implementation is correct and efficient in this regard.

Thanks for your clarification. I'll close this comment as resolved.

zhzhuang-zju:
@gemini-code-assist
Any further comments on this pr?

XiShanYongYe-Chang (commented):
Thanks~
/lgtm

This patch involves the content of preemption, invite @whitewindmills to help review it.
/cc @whitewindmills

pkg/detector/detector.go
```diff
@@ -693,36 +695,41 @@ func (d *ResourceDetector) GetUnstructuredObject(objectKey keys.ClusterWideKey)
...
 
-	objLabels := object.GetLabels()
-	if len(objLabels) > 0 {
-		// object has been claimed, don't need to claim again
-		if !excludeClusterPolicy(object) &&
-			objLabels[policyv1alpha1.PropagationPolicyPermanentIDLabel] == policyID {
-			return policyID, nil
-		}
+	claimedID := util.GetLabelValue(object.GetLabels(), policyv1alpha1.PropagationPolicyPermanentIDLabel)
+	_, ok := object.GetLabels()[policyv1alpha1.ClusterPropagationPolicyPermanentIDLabel]
```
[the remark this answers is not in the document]
whitewindmills: +1, just some nits otherwise LGTM.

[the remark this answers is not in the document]
zhzhuang-zju: done~

zhzhuang-zju:
Thanks @whitewindmills @XiShanYongYe-Chang for your review~
All issues have been addressed. Please take a look.

whitewindmills:
LGTM
by the way, I just want to make sure whether there are similar problems, because I remember that we have fixed this kind of problem before, but I didn't expect it to still exist now.

zhzhuang-zju:
> LGTM by the way, I just want to make sure whether there are similar problems, because I remember that we have fixed this kind of problem before, but I didn't expect it to still exist now.

@whitewindmills, we are going to add a cache mutation check in pr https://github.com/karmada-io/karmada/pull/6541. Hoping this check can help avoid the same issue

pkg/detector/detector.go
```diff
@@ -693,36 +695,42 @@ func (d *ResourceDetector) GetUnstructuredObject(objectKey keys.ClusterWideKey)
...
-		// object has been claimed, don't need to claim again
-		if !excludeClusterPolicy(object) &&
-			objLabels[policyv1alpha1.PropagationPolicyPermanentIDLabel] == policyID {
-			return policyID, nil
-		}
+	claimedID := util.GetLabelValue(objLabels, policyv1alpha1.PropagationPolicyPermanentIDLabel)
+	_, ok := objLabels[policyv1alpha1.ClusterPropagationPolicyPermanentIDLabel]
+	// object has been claimed, don't need to claim again
+	if !ok && claimedID == policyID {
+		return object, policyID, nil
```
RainbowMango: This code is somewhat hard to understand.

zhzhuang-zju: How about extracting a function to represent whether the object's CPP metadata needs cleaning?
```go
func (d *ResourceDetector) ClaimPolicyForObject(object *unstructured.Unstructured, policy *policyv1alpha1.PropagationPolicy) (*unstructured.Unstructured, string, error) {
	policyID := policy.Labels[policyv1alpha1.PropagationPolicyPermanentIDLabel]

	objLabels := object.GetLabels()
	claimedID := util.GetLabelValue(objLabels, policyv1alpha1.PropagationPolicyPermanentIDLabel)
	// object has been claimed, don't need to claim again
	// In the case pp preempts cpp, it is necessary to first clear the cpp metadata.
	if !needCleanupCPPClaimMetadata(object) && claimedID == policyID {
		return object, policyID, nil
	}

	objectCopy := object.DeepCopy()
	CleanupCPPClaimMetadata(objectCopy)
	AddPPClaimMetadata(objectCopy, policyID, policy.ObjectMeta)
	err := d.Client.Update(context.TODO(), objectCopy)
	return objectCopy, policyID, err
}
```

RainbowMango: I guess we can have two helper functions for checking if a resource template has been claimed by PP/CPP or not.

RainbowMango:
@zhzhuang-zju @whitewindmills @CharlesQQ 
~~When I trying to find the root cause of this issue, I realized that this issue was introduced by #5813~~ 

[edit]: See the root cause analysis below: https://github.com/karmada-io/karmada/pull/6544#issuecomment-3163178853

So, my opinon on this PR are:
1. This RP currently addresses a symptom, not the root cause.  The real issue is that #5813 introduced `UnsafeDisableDeepCopy` without establishing clear rules about cache object usage. This PR fixes one instance, but similar issues likely exist elsewhere.
2. We should revist #5813 and establish a clear rule of when the `UnsafeDisableDeepCopy` is allowed, like the listed objects cannot leak outside the function scope.

RainbowMango:
I don't see any signs of this issue really break any functional, so we don't have to fix this in rush.

zhzhuang-zju:
@RainbowMango @whitewindmills @XiShanYongYe-Chang After further confirmation, this issue was not introduced by #5813. #5813 disables deepcopy in the scenario of listing resourcebinding/policy, while this issue involves modifying the resourcetemplate informer cache. The function used to retrieve resourcetemplate is as follows:
https://github.com/karmada-io/karmada/blob/64fbe7560d0b6ce1022f6bb94411107b8a0b04bf/pkg/detector/detector.go#L659-L693

As you can see from the function comments, the return value of this function may come from the informer cache, so it requires users to perform a deepcopy before use. Issue 3878 documented this problem. However, in actual operations, since the resource template returned by the `GetUnstructuredObject` function goes through a deep function call chain, this comment information eventually gets lost. Therefore, I suggest that the `GetUnstructuredObject` function should perform a deepcopy before returning the resource template, thus ensuring that subsequent usage won't directly modify the informer cache. 

PS: Although both this issue and #3878 are problems caused by using `GetUnstructuredObject`, the scenarios are different and won't lead to the problem recorded in #3878.

RainbowMango:
cc @CharlesQQ

CharlesQQ:
I understand that this PR wants to solve the hidden dangers rather than the problem. If you want to do this, are there other similar hidden dangers that need to be solved instead of just this one? Or make sure that all resource objects obtained from the informer need to return deep copy objects to avoid modifying the informer cache.

zhzhuang-zju:
> I understand that this PR wants to solve the hidden dangers rather than the problem. If you want to do this, are there other similar hidden dangers that need to be solved instead of just this one? Or make sure that all resource objects obtained from the informer need to return deep copy objects to avoid modifying the informer cache.

Yes, you're absolutely right. We should establish a constraint to prevent similar issues from occurring. However, investigating potential hidden risks shouldn't block the current PR, as this PR addresses a known issue that definitively modifies the informer cache, making it a higher priority fix. 
PS: After fixing this issue, it will pass the cache mutation check.

We can leverage the cache mutation check provided by #6541 to identify issues in new code and establish constraints to standardize new implementations while investigating and improving existing ones.

zhzhuang-zju:
Regarding how to handle the risk of informer cache modification, we need to establish a set of related coding standards. My ideas are:

- For getting resource from informer cache: Consider directly returning a deep copy of the resource
- For listing resources from informer cache: Consider adding a "readonly" prefix to function names and variable names to indicate to users that they should not modify the objects until performing a deep copy

pkg/detector/detector.go
RainbowMango, line 1: Are the changes in this file the minimum fix for this issue?
If so, please keep only these changes in this PR to facilitate cherry-pick.

zhzhuang-zju: > Are the changes in this file the minimum fix for this issue?

I will do further verification by adding cache mutation checks and running comprehensive e2e tests. And show you the test report.

> If so, please keep only these changes in this PR to facilitate cherry-pick.

Get it~ Related enhancements can be done in the next step. This PR focuses on fixing the informer cache modification issue.

RainbowMango: Yeah, let's move faster.

zhzhuang-zju:
Test Report: https://github.com/karmada-io/karmada/actions/runs/16873498290/job/47793071215?pr=6544

Based on the logs, cache modifications are no longer being detected by the cache mutation check

PS: If this PR is ready to be merged, I will remove the cache mutation check.

zhzhuang-zju:
cc @RainbowMango @CharlesQQ @XiShanYongYe-Chang 
This should be the minimum fix for this issue

zhzhuang-zju:
> Aug 13 07:19:55 karmada-host-control-plane kubelet[707]: E0813 07:19:55.387460     707 pod_workers.go:1301] "Error syncing pod, skipping" err="failed to \"StartContainer\" for \"kube-apiserver\" with ImagePullBackOff: \"Back-off pulling image \\\"registry.k8s.io/kube-apiserver:v1.31.3\\\": ErrImagePull: rpc error: code = DeadlineExceeded desc = failed to pull and unpack image \\\"registry.k8s.io/kube-apiserver:v1.31.3\\\": failed to copy: httpReadSeeker: failed open: failed to do request: Get \\\"https://us-east5-docker.pkg.dev/v2/k8s-artifacts-prod/images/kube-apiserver/manifests/sha256:c040161602d7b5c0f899ef41d0e53b28fe007120819e91440f89d6a4b15fbcee\\\": dial tcp 172.253.132.82:443: i/o timeout\"" pod="test/karmada-demo-apiserver-77dbbd8856-r8djq" podUID="f39aaa70-917e-402f-81d1-c25b312df316"

Image pulling error
/retest

zhzhuang-zju:
> Do you think this should be backported?

Yes, I think it's necessary. I'll submit an issue to track it.

RainbowMango:
Then, please add a release note. 
/kind bug

PR #6564, opened 2025-07-21
Use structured log for karmada operator

zhangsquared:
**What type of PR is this?**

/kind cleanup

most of the logs have already been use new structured format... I only found 2 exceptions...
I am also not sure whether even number of etcd should qualify as an error, or just `InfoS` it...

```release-note
none
```


operator/pkg/controller/karmada/validating.go
```diff
@@ -80,7 +80,7 @@ func validateETCD(etcd *operatorv1alpha1.Etcd, karmadaName string, fldPath *fiel
 		replicas := *etcd.Local.CommonSettings.Replicas
 
 		if (replicas % 2) == 0 {
-			klog.Warningf("invalid etcd replicas %d, expected an odd number", replicas)
+			klog.InfoS("Invalid etcd replicas, expected an odd number", "replicas", replicas)
```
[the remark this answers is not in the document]
zhangsquared: ??? I don't see any `WarningS`...
Should we create another `WarningS`?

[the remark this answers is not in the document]
zhangsquared: I am also not sure whether even number of etcd should qualify as an error, or just InfoS it...
please advise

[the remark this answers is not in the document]
jabellard: We can ignore this. That method does not exist. The reason why we run this check is because etcd clusters should usually have an odd number of members.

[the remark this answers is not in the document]
jabellard: Same as above.

jabellard:
/lgtm

Generally looks good to me. 

/cc @RainbowMango  for another look.

operator/pkg/controller/karmada/validating.go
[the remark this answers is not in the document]
RainbowMango: So, what's going to happen if the replica number is not odd? 
Shall we fail the validation to avoid potential unexpected configuration?

operator/pkg/controller/karmada/controller.go
```diff
@@ -97,7 +97,7 @@ func (ctrl *Controller) Reconcile(ctx context.Context, req controllerruntime.Req
 	}
 
 	if err := ctrl.validateKarmada(ctx, karmada); err != nil {
-		klog.Errorf("Validation failed for karmada: %+v", err)
+		klog.ErrorS(err, "Validation failed for karmada")
```
RainbowMango:
```suggestion
		klog.ErrorS(err, "Validation failed for karmada", "name", karmada.Name)
```
Does this make sense to log the instance name? I guess it might be useful for your case. @jabellard

jabellard: > Does this make sense to log the instance name? I guess it might be useful for your case. @jabellard

I think it's useful. Otherwise, when looking at the logs, it's not clear for which instance this error is for.

jabellard: Regarding number of members when clustering etcd:
It's recommended to always cluster an etcd instance with an odd number of members for the reasons outlined in the [operations guide](https://etcd.io/docs/v3.3/faq/#why-an-odd-number-of-cluster-members). However, have a cluster with an even number of members is not wrong.

jabellard: We can chose to enforce having an odd number of members, but I think this is out of scope for this PR. Also, that change won't be backwards compatible.

jabellard: @zhangsquared , maybe we should update the log message to indicate that having an odd number of members is not recommended instead of indicating that it's invalid? What do you think?

RainbowMango: +1

zhangsquared: thank you guys for the advise!

jabellard:
@zhangsquared thanks for making the changes.

/lgtm

/cc @RainbowMango for another look.

RainbowMango (approved):
/approve
Squashing commits by adding label.

PR #6570, opened 2025-07-22
JSON logging for deployment replicas syncer controller

NUMBKV:
**What type of PR is this?**

/kind cleanup

JSON structured logging support for deployment replicas syncer controller. Also added more object context to the logging. More details are highlighted in https://github.com/karmada-io/karmada/issues/6230.

Altered the logging to use structured logs. Also I addressed some grammar issues and added more object context (namespace, name) in the logs that didn't originally have this. Please let me know if this is needed or not.

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/deploymentreplicassyncer/deployment_replicas_syncer_controller.go
+++ b/pkg/controllers/deploymentreplicassyncer/deployment_replicas_syncer_controller.go
@@ -19,6 +19,7 @@ package deploymentreplicassyncer
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"time"
 
 	appsv1 "k8s.io/api/apps/v1"
@@ -71,7 +72,10 @@ var predicateFunc = predicate.Funcs{
 		}
 
 		if oldObj.Spec.Replicas == nil || newObj.Spec.Replicas == nil {
-			klog.Errorf("spec.replicas field unexpectedly become nil: %+v, %+v", oldObj.Spec.Replicas, newObj.Spec.Replicas)
+			klog.ErrorS(fmt.Errorf("spec.replicas is nil in either old or new deployment object"),
+				"spec.replicas field unexpectedly become nil",
+				"oldReplicas", oldObj.Spec.Replicas, "newReplicas", newObj.Spec.Replicas,
+				"namespace", oldObj.Namespace, "name", oldObj.Name)
 			return false
 		}
 
@@ -99,7 +103,7 @@ func (r *DeploymentReplicasSyncer) SetupWithManager(mgr controllerruntime.Manage
 // The Controller will requeue the Request to be processed again if an error is non-nil or
 // Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
 func (r *DeploymentReplicasSyncer) Reconcile(ctx context.Context, req controllerruntime.Request) (controllerruntime.Result, error) {
-	klog.V(4).Infof("Reconciling for Deployment %s/%s", req.Namespace, req.Name)
+	klog.V(4).InfoS("Reconciling for Deployment", "namespace", req.Namespace, "name", req.Name)
 
 	deployment := &appsv1.Deployment{}
 	binding := &workv1alpha2.ResourceBinding{}
@@ -108,7 +112,8 @@ func (r *DeploymentReplicasSyncer) Reconcile(ctx context.Context, req controller
 	// 1. get latest binding
 	if err := r.Client.Get(ctx, client.ObjectKey{Namespace: req.Namespace, Name: bindingName}, binding); err != nil {
 		if apierrors.IsNotFound(err) {
-			klog.Infof("no need to update deployment replicas for binding not found")
+			klog.InfoS("no need to update deployment replicas as binding was not found",
+				"namespace", req.Namespace, "name", bindingName)
 			return controllerruntime.Result{}, nil
 		}
 		return controllerruntime.Result{}, err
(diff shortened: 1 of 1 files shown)


RainbowMango:
/assign
Will do ASAP.

pkg/controllers/deploymentreplicassyncer/deployment_replicas_syncer_controller.go
```diff
@@ -108,7 +112,8 @@ func (r *DeploymentReplicasSyncer) Reconcile(ctx context.Context, req controller
 	// 1. get latest binding
 	if err := r.Client.Get(ctx, client.ObjectKey{Namespace: req.Namespace, Name: bindingName}, binding); err != nil {
 		if apierrors.IsNotFound(err) {
-			klog.Infof("no need to update deployment replicas for binding not found")
+			klog.InfoS("no need to update deployment replicas as binding was not found",
+				"namespace", req.Namespace, "name", bindingName)
```
RainbowMango:
```suggestion
				"namespace", req.Namespace, "name", req.Name)
```
Here should log the deployment name, as this log indicates the reconciling for this deployment is skipped.

pkg/controllers/deploymentreplicassyncer/deployment_replicas_syncer_controller.go
```diff
@@ -157,41 +166,44 @@ func (r *DeploymentReplicasSyncer) Reconcile(ctx context.Context, req controller
 func isDeploymentStatusCollected(deployment *appsv1.Deployment, binding *workv1alpha2.ResourceBinding) bool {
 	// make sure the replicas change in deployment.spec can sync to binding.spec, otherwise retry
 	if deployment.Spec.Replicas == nil || *deployment.Spec.Replicas != binding.Spec.Replicas {
-		klog.V(4).Infof("wait until replicas of binding (%d) equal to replicas of deployment (%d)",
-			binding.Spec.Replicas, *deployment.Spec.Replicas)
```
RainbowMango, line 161: I guess here is a panic risk. 

The condition checks `deployment.Spec.Replicas == nil`, but then in the log statement, it dereferences `*deployment.Spec.Replicas`.

If deployment.Spec.Replicas is nil, dereferencing it with * will cause a panic.

@chaosi-zju @zhzhuang-zju Can you help to take a look?

zhzhuang-zju: Excellent discovery! After investigation, we can rule out the risk of panic. The deployment resource has registered a default value setting method `setDefaults_deployment`, which sets the default value of `deployment.Spec.Replicas` to 1.

https://github.com/kubernetes/kubernetes/blob/dd4e4f1dd13f68896595120661af239e55a3ba67/pkg/apis/apps/v1/defaults.go#L38-L42 https://github.com/kubernetes/kubernetes/blob/dd4e4f1dd13f68896595120661af239e55a3ba67/pkg/apis/apps/v1/zz_generated.defaults.go#L373-L374

So theoretically, there should be no panic. However, the code implementation has a slight code smell.

jabellard: @RainbowMango, @zhzhuang-zju , should we improve this as part of this PR?

zhzhuang-zju: From a defensive programming perspective, this code segment can be slightly modified.

```go
	if deployment.Spec.Replicas == nil {
		// should never happen
		klog.ErrorS(nil, "deployment replicas is nil", "namespace", deployment.Namespace, deployment.Name)
		return false
	}
	if *deployment.Spec.Replicas != binding.Spec.Replicas {
		klog.V(4).InfoS("wait until binding replicas are equal to deployment replicas",
			"bindingReplicas", binding.Spec.Replicas, "deploymentReplicas", *deployment.Spec.Replicas,
			"namespace", deployment.Namespace, "deploymentName", deployment.Name, "bindingName", binding.Name)
		return false
	}
```

RainbowMango: Thanks @zhzhuang-zju for the clarification. Now we know it's a false-positive warning.

We can improve it either in this PR or start another PR for this. @NUMBKV What's your opinion?

But, anyway, I would suggest adding more comments on why the `deployment.Spec.Replicas == nil` never happen. like:
```
should never happen as karmada-apiserver defaults `deployment.spec.Replicas` to 1 if it is not set.
```

NUMBKV: Thanks all for the clarification! I'm ok with adding it to this PR since the scope of changes is relatively small👍

NUMBKV:
@jabellard @zhzhuang-zju @RainbowMango Thanks for the review! I have addressed the comments in the latest commit. Can you help take a look when you get a chance? tysm!

jabellard:
> @jabellard @zhzhuang-zju @RainbowMango Thanks for the review! I have addressed the comments in the latest commit. Can you help take a look when you get a chance? tysm!

Thanks! Generally looks good to me.

/cc @RainbowMango for another look.

RainbowMango:
@zhzhuang-zju Please help take a look

zhzhuang-zju:
Thank~ 
/lgtm

zhzhuang-zju:
Leave APPROVE to @RainbowMango

RainbowMango (approved):
/lgtm
/approve

RainbowMango:
Thanks all. I helped tidy the commits and pushed to your branch. @NUMBKV .

Everything should be fine now.

PR #6576, opened 2025-07-25
Use Structure Logging Enhancement for `Cluster Resource Binding Controller`

zhuyulicfc49:
**What type of PR is this?**  
This change is part of the initiative to adopt structured logging. More details can be found in [#6230](https://github.com/karmada-io/karmada/issues/6230), 

Part of [#6230](https://github.com/karmada-io/karmada/issues/6230)

None

None

### DIFF
--- a/pkg/controllers/status/crb_status_controller.go
+++ b/pkg/controllers/status/crb_status_controller.go
@@ -58,7 +58,7 @@ type CRBStatusController struct {
 // The Controller will requeue the Request to be processed again if an error is non-nil or
 // Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
 func (c *CRBStatusController) Reconcile(ctx context.Context, req controllerruntime.Request) (controllerruntime.Result, error) {
-	klog.V(4).Infof("Reconciling ClusterResourceBinding %s.", req.NamespacedName.String())
+	klog.V(4).InfoS("Reconciling ClusterResourceBinding", "name", req.NamespacedName.Name)
 
 	binding := &workv1alpha2.ClusterResourceBinding{}
 	if err := c.Client.Get(ctx, req.NamespacedName, binding); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/controllers/status/crb_status_controller.go
```diff
@@ -112,8 +112,7 @@ func (c *CRBStatusController) SetupWithManager(mgr controllerruntime.Manager) er
 func (c *CRBStatusController) syncBindingStatus(ctx context.Context, binding *workv1alpha2.ClusterResourceBinding) error {
 	err := helper.AggregateClusterResourceBindingWorkStatus(ctx, c.Client, binding, c.EventRecorder)
 	if err != nil {
-		klog.Errorf("Failed to aggregate workStatues to clusterResourceBinding(%s), Error: %v",
-			binding.Name, err)
+		klog.ErrorS(err, "Failed to aggregate workStatues to clusterResourceBinding", "clusterResourceBinding", klog.KObj(binding))
```
jabellard: Please use the `name` key here instead of `clusterResourceBinding` with the corresponding name of the object as the value for consistency with the other controllers.

jabellard: Also there's other logs calls in the file that I think we missed. Please update those as well

jabellard (changes requested):
Left some minor comments. Other than that, all looks great!

pkg/controllers/status/crb_status_controller.go
zhuyulicfc49: Thank you Joe, that makes sense! I’ve addressed your feedback by updating the log key to use "name" for consistency.

While reviewing other usage in the codebase, I noticed that some controllers (e.g., [cluster_resource_binding_controller.go](https://github.com/karmada-io/karmada/blob/master/pkg/controllers/binding/cluster_resource_binding_controller.go#L148)) use `ClusterResourceBinding` as the key instead of `name`, like:

```
klog.V(4).InfoS("Reconciling ClusterResourceBinding.", "ClusterResourceBinding", req.NamespacedName.Name)
klog.ErrorS(err, "Failed to delete works related to ClusterResourceBinding.", "ClusterResourceBinding", clusterResourceBinding.Name)
```

I think for my change I'll just keep key with `name` here to align with conventions and your suggestion, I can help improve the one I referenced if you're ok with it. Thanks!

jabellard:
Thanks!

/lgtm

/cc @RainbowMango  for another look.

RainbowMango (approved):
/approve

PR #6596, opened 2025-08-02
Add support for specifying env vars and extra args for the Karmada operator

jabellard:
**What type of PR is this?**

/kind feature

This PR adds support for injecting arbitrary environment variables (e.g., http proxy env vars)  into the Karmada Operator deployment.
It also adds supports for specifying extra arguments (e.g., `--log-format=json`) for the operator binary.

Fixes #

```release-note
`Helm chart`: Karmada Operator chart now possible to set environment variables and extra arguments for the Karmada operator deployment.
```

### DIFF
--- a/charts/karmada-operator/templates/karmada-operator-deployment.yaml
+++ b/charts/karmada-operator/templates/karmada-operator-deployment.yaml
@@ -44,6 +44,13 @@ spec:
         - /bin/karmada-operator
         - --leader-elect-resource-namespace={{ .Release.Namespace }}
         - --v=2
+        {{- range .Values.operator.extraArgs }}
+        - {{ . }}
+        {{- end }}
+        {{- with .Values.operator.env }}
+        env:
+        {{- toYaml . | nindent 10 }}
+        {{- end }}
         {{- if .Values.operator.resources }}
         resources: {{- toYaml .Values.operator.resources | nindent 12 }}
         {{- end }}
--- a/charts/karmada-operator/values.yaml
+++ b/charts/karmada-operator/values.yaml
@@ -58,7 +58,31 @@ operator:
     ##   - myRegistryKeySecretName
     ##
     pullSecrets: []
-  ## @param.resources
+
+  ## @param operator.env List of environment variables to inject
+  ##
+  ## - Each entry must be a valid Kubernetes EnvVar object.
+  ## - Supports both literal values and valueFrom references (ConfigMap, Secret, fieldRef, etc.).
+  ## - If omitted or set to an empty array (`[]`), no env stanza will be included.
+  ##
+  ## A sample stanza is shown below.
+  ##
+#  env:
+#    - name: http_proxy
+#      value: "http://best-awesome-proxy.com:8080"
+#    - name: https_proxy
+#      value: "http://best-awesome-proxy.com:8080"
+#    - name: no_proxy
+#      value: "localhost,127.0.0.1,*.svc,*.cluster.local"
+
+  ## @param operator.extraArgs List of extra arguments for the operator binary
+  ##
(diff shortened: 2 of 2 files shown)


charts/karmada-operator/templates/karmada-operator-deployment.yaml
```diff
@@ -44,6 +44,10 @@ spec:
         - /bin/karmada-operator
         - --leader-elect-resource-namespace={{ .Release.Namespace }}
         - --v=2
+        {{- if .Values.operator.env }}
+        env:
+        {{- toYaml .Values.operator.env | nindent 10 }}
+        {{- end }}
```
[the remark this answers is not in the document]
jabellard: Done.

jabellard:
Hey @RainbowMango . Can you please take a look?

RainbowMango:
Sure, by the way, are you using this chart to install karmada-operator?

charts/karmada-operator/values.yaml
```diff
@@ -58,7 +58,37 @@ operator:
...
+#    - name: http_proxy
+#      value: "http://best-awesome-proxy.com:8080"
+#    - name: HTTPS_PROXY
+#      value: "http://best-awesome-proxy.com:8080"
+#    - name: https_proxy
+#      value: "http://best-awesome-proxy.com:8080"
+#    - name: NO_PROXY
+#      value: "localhost,127.0.0.1,*.svc,*.cluster.local"
+#    - name: no_proxy
+#      value: "localhost,127.0.0.1,*.svc,*.cluster.local"
```
RainbowMango: If i remember correctly, it is not allowed to specify the same environment variable twice, one in uppercase and one in lowercase. 

My question is why we need to have them both in example. kind of misleading to me.

jabellard: Good question. It's possible to set both. There's some history as to why it may be useful to do that, but in this case that actually does not really matter, so I'll simplify this to only the lower case variants which should be honored by most *nix tools and  libraries.

jabellard: @RainbowMango , thanks for reviewing. I just pushed changes to address and then squashed.

jabellard:
> Sure, by the way, are you using this chart to install karmada-operator?

Yes. I use it to install the operator on our host/management clusters.

RainbowMango:
/retest
unrelated tests. https://github.com/karmada-io/karmada/actions/runs/16804289194/job/47593236644?pr=6596
```
  At one point, however, the function did return successfully.
  Yet, Eventually failed because the matcher was not satisfied:
  Expected
      <bool>: false
  to equal
      <bool>: true
  In [It] at: /home/runner/work/karmada/karmada/test/e2e/suites/base/resource_test.go:711 @ 08/07/25 13:09:23.418

  Full Stack Trace
    github.com/karmada-io/karmada/test/e2e/suites/base.init.func69.1.3.3()
    	/home/runner/work/karmada/karmada/test/e2e/suites/base/resource_test.go:711 +0xa2
    github.com/karmada-io/karmada/test/e2e/suites/base.init.func69.1.3()
    	/home/runner/work/karmada/karmada/test/e2e/suites/base/resource_test.go:698 +0x487
```

charts/karmada-operator/values.yaml
RainbowMango: Oh, you are right, thanks for correcting me. I ran a test, but still, feel unbelievable :):
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: env-test
spec:
  containers:
  - name: test-container
    image: busybox
    command: ["/bin/sh", "-c", "env"]
    env:
    - name: FOO
      value: "uppercase"
    - name: foo
      value: "lowercase"

```

Issue #6600, opened 2025-08-05
After the karmada-controller retarts, the dependency distributed resources are deleted unexpectedly


XiShanYongYe-Chang:
**What happened**:

We simulated the following test: 

We created a series of `Deployment` resources in a Karmada control plane and defined the dependency of `Deployment` on `Service` resources through the `ResourceInterpreterCustomization` resource. This way, the `Service` can be distributed to member clusters along with the `Deployment`. Then, at regular intervals, which are sufficient to ensure that the existing resources have been reconciled, we restarted the `karmada-controller-manager` component. After the component restarted, we observed whether the resources in the member clusters were deleted. The result occasionally showed that the `Service` resources in the member clusters were deleted (at this point, we are not concerned with whether the resources in the member clusters can be recreated, so the deletion of `Service` resources could potentially lead to service disruption).

**What you expected to happen**:

Regardless of whether the `karmada-controller-manager` component is running normally or has restarted, resources distributed to member clusters by Karmada will not be automatically deleted by the system, which could otherwise lead to service disruption.

**How to reproduce it (as minimally and precisely as possible)**:

Just as described in the **What happened** part.

**Anything else we need to know?**:

Through the analysis of the above issues, we can address the problem step by step:

- [x] Fix the issue that the cache is out of sync when configManager's `HasSynced` is queried in ConfigurableInterpreter. @XiShanYongYe-Chang #6602 
- [x] Fix the issue that the cache is out of sync when hookManager's `HasSynced` is queried in CustomizedInterpreter. @liaolecheng #6612 
- [x] Ensure resource interpreter cache sync before starting controllers. @XiShanYongYe-Chang #6626
- [ ] ~~Correctly handle the case where `HasSynced` returns false in a resource interpreter and throw the error instead of hiding it.~~(cancel this task, see https://github.com/karmada-io/karmada/pull/6628#issuecomment-3199619113) @liaolecheng #6628
- [x] Fix the issue that the cache is out of sync within the resource interpreter when using karmadactl promote @XiShanYongYe-Chang #6669 

**Environment**:
- Karmada version: v1.14
- kubectl-karmada or karmadactl version (the result of `kubectl-karmada version` or `karmadactl version`):
- Others:

XiShanYongYe-Chang:
[...]

PR #6602, opened 2025-08-06
Fix the issue that the cache is out of sync when configManager's HasSynced is called in ConfigurableInterpreter

XiShanYongYe-Chang:
**What type of PR is this?**

/kind bug

ref #6600

Part of https://github.com/karmada-io/karmada/issues/6600#issuecomment-3157191026

```release-note
`karmada-controller-manager`: Fix the issue that the cache is out of sync when configManager's HasSynced is called in ConfigurableInterpreter
```


RainbowMango (commented):
/assign

pkg/resourceinterpreter/customized/declarative/configmanager/manager.go
```diff
@@ -84,35 +80,42 @@ func NewInterpreterConfigManager(informer genericmanager.SingleClusterInformerMa
...
 		informer.ForResource(resourceInterpreterCustomizationsGVR, configHandlers)
 	}
 
 	return manager
 }
 
-func (configManager *interpreterConfigManager) updateConfiguration() {
+func (configManager *interpreterConfigManager) updateConfiguration() error {
+	if configManager.informer == nil {
+		return errors.New("ResourceInterpreterCustomization informer not initialized")
```
RainbowMango:
```suggestion
		return errors.New("informer manager is not configured")
```
I remember error messages usually do not start with a capital character.  (Do not know why lint missed it.)

pkg/resourceinterpreter/customized/declarative/configmanager/manager.go
```diff
@@ -84,35 +80,42 @@ func NewInterpreterConfigManager(informer genericmanager.SingleClusterInformerMa
...
+	}
+	if !configManager.informer.IsInformerSynced(resourceInterpreterCustomizationsGVR) {
+		return errors.New("ResourceInterpreterCustomization informer not synced")
```
RainbowMango:
```suggestion
		return errors.New("informer of ResourceInterpreterCustomization not synced")
```
ditto.

pkg/resourceinterpreter/customized/declarative/configmanager/manager.go
```diff
@@ -84,35 +80,42 @@ func NewInterpreterConfigManager(informer genericmanager.SingleClusterInformerMa
...
+			func(_, _ interface{}) { _ = manager.updateConfiguration() },
+			func(_ interface{}) { _ = manager.updateConfiguration() })
 		informer.ForResource(resourceInterpreterCustomizationsGVR, configHandlers)
 	}
 
 	return manager
 }
 
-func (configManager *interpreterConfigManager) updateConfiguration() {
+func (configManager *interpreterConfigManager) updateConfiguration() error {
```
RainbowMango, line 101:
```suggestion
// updateConfiguration is used as the event handler for the ResourceInterpreterCustomization resource.
// Any changes (add, update, delete) to these resources will trigger this method, which loads all
// ResourceInterpreterCustomization resources and refreshes the internal cache accordingly.
// Note: During startup, some events may be missed if the informer has not yet synced. If all events
// are missed during startup, updateConfiguration will be called when HasSynced() is invoked for the
// first time, ensuring the cache is updated on first use.
func (configManager *interpreterConfigManager) updateConfiguration() error {
```
Please add a comment to clarify the missed events during startup.

pkg/resourceinterpreter/customized/declarative/configmanager/manager.go
XiShanYongYe-Chang: Lint may not have this check item; Goland will provide reminders during programming.

RainbowMango:
/retitle Ensure configManager waits for informer cache sync before reporting as synced in ConfigurableInterpreter

RainbowMango (approved):
/approve

Leave LGTM to @GitHubxsy for another look.

GitHubxsy (commented):
/lgtm

PR #6603, opened 2025-08-06
Fix the issue of infinite loop caused by connection failure

whitewindmills:
**What type of PR is this?**
/kind bug

We encountered an OOM problem in the test environment due to network reasons. From the log before the pod exited, we found that the httpserver encountered a recursive function call that caused a stack overflow when processing an error.
image
image

Fixes #

```release-note
NONE
```


whitewindmills:
The community code has the same bug, so it is fixed synchronously.
/cc @RainbowMango @XiShanYongYe-Chang

pkg/webhook/interpreter/http.go
```diff
@@ -98,7 +98,8 @@ func (wh *Webhook) writeResponse(w io.Writer, response Response) {
 func (wh *Webhook) writeResourceInterpreterResponse(w io.Writer, interpreterContext configv1alpha1.ResourceInterpreterContext) {
 	if err := json.NewEncoder(w).Encode(interpreterContext); err != nil {
 		klog.Errorf("unable to encode the response: %v", err)
-		wh.writeResponse(w, Errored(http.StatusInternalServerError, err))
+		// If we fail to write a response, we can't write a failure response.
```
XiShanYongYe-Chang: > If we fail to write a response

Are you referring to an error occurring when calling the `writeResponse` function? Is my understanding correct?

pkg/webhook/interpreter/http.go
```diff
@@ -98,7 +98,8 @@ func (wh *Webhook) writeResponse(w io.Writer, response Response) {
...
+		// This prevents an infinite recursion if the connection is closed.
```
XiShanYongYe-Chang: What is the reason for this situation?

whitewindmills: Network failure.

pkg/webhook/interpreter/http.go
whitewindmills: yes

XiShanYongYe-Chang:
`controller-runtime` encounters the same error, we can learn from it: https://github.com/kubernetes-sigs/controller-runtime/pull/1930

pkg/webhook/interpreter/http.go
```diff
@@ -97,8 +97,16 @@ func (wh *Webhook) writeResponse(w io.Writer, response Response) {
...
 func (wh *Webhook) writeResourceInterpreterResponse(w io.Writer, interpreterContext configv1alpha1.ResourceInterpreterContext) {
 	if err := json.NewEncoder(w).Encode(interpreterContext); err != nil {
-		klog.Errorf("unable to encode the response: %v", err)
-		wh.writeResponse(w, Errored(http.StatusInternalServerError, err))
+		klog.Errorf("unable to encode and write the response: %v", err)
+		// Since the `ar configv1alpha1.ResourceInterpreterContext` is a clear and legal object,
+		// it should not have problem to be marshalled into bytes.
+		// The error here is probably caused by the abnormal HTTP connection,
+		// e.g., broken pipe, so we can only write the error response once,
+		// to avoid endless circular calling.
```
XiShanYongYe-Chang, line 105: Can we also post the PR from `controller-runtime` here for easier traceability?

pkg/webhook/interpreter/http_test.go
```diff
@@ -299,7 +299,7 @@ func TestWriteResourceInterpreterResponse(t *testing.T) {
 			res: configv1alpha1.ResourceInterpreterContext{
 				Response: &configv1alpha1.ResourceInterpreterResponse{},
 			},
-			rec: &limitedBadResponseWriter{maxFailures: 3},
+			rec: &limitedBadResponseWriter{maxFailures: 1},
```
XiShanYongYe-Chang: Do we still need to make this change?

And would it be convenient to also synchronize the newly added test cases from https://github.com/kubernetes-sigs/controller-runtime/pull/1930? If it's not easy to handle(I looked at the framework for the test, and it was used differently), I can separately record a task to address it later.

pkg/webhook/interpreter/http.go
whitewindmills: +1

pkg/webhook/interpreter/http_test.go
whitewindmills: This test case seems to detect the number of recursive calls, so the test case will fail because there is no longer a recursive call. I don't understand why this test case was introduced at the time.

XiShanYongYe-Chang: Let me take a look at the UT.

XiShanYongYe-Chang: Hi @whitewindmills, the test case provides reasonable coverage for testing recursive error handling, but since the current pr will no longer encounter recursive errors, this test case can be removed.

pkg/webhook/interpreter/http_test.go
```diff
@@ -299,7 +299,7 @@ func TestWriteResourceInterpreterResponse(t *testing.T) {
 			res: configv1alpha1.ResourceInterpreterContext{
```
XiShanYongYe-Chang, line 288: How about adding a new test case to cover the new adding code:

```go
type brokenWriter struct {
	http.ResponseWriter
}

func (bw *brokenWriter) Write(_ []byte) (int, error) {
	return 0, fmt.Errorf("mock: write: broken pipe")
}

...

        {
			name:        "should never run into circular calling if the writer has broken",
			mockHandler: &HTTPMockHandler{},
			res: configv1alpha1.ResourceInterpreterContext{
				Response: &configv1alpha1.ResourceInterpreterResponse{},
			},
			rec: &brokenWriter{},
			verify: func(writer io.Writer, response *configv1alpha1.ResourceInterpreterResponse) error {
				// reaching here means not running into circular calling
				return nil
			},
		},
...
```

whitewindmills:
@XiShanYongYe-Chang PTAL

XiShanYongYe-Chang (commented):
Thanks a lot
/lgtm
/cc @RainbowMango

XiShanYongYe-Chang:
Hi @whitewindmills, a lint error here:
```
Error: pkg/webhook/interpreter/http_test.go:292:17: unused-parameter: parameter 'writer' seems to be unused, consider removing or renaming it as _ (revive)
			verify: func(writer io.Writer, response *configv1alpha1.ResourceInterpreterResponse) error {
			             ^
1 issues:
* revive: 1
```

RainbowMango:
Please cc me again after you fixed the lint issue.

RainbowMango:
What a small world! @wengyao04  Is that you? :)
I found a Slack thread from https://kubernetes.slack.com/archives/CAR30FCJZ/p1654871872956339.

RainbowMango (commented):
/assign

RainbowMango (approved):
/approve

I added a release note by the way.

XiShanYongYe-Chang:
/lgtm
Let me create an issue to track the cherry-pick task.

RainbowMango:
@karmada-io/karmada-adopter-group

Please review this PR and evaluate how this issue may impact your envioment.
In short, if you are using the [webhook interpreter scaffolding](https://github.com/karmada-io/karmada/tree/master/pkg/webhook/interpreter) to build your webhook, then you are affected.

Note that, the webhook interpreter is maintained by yourself, you might need to cherry-pick the patch by yourself.

PR #6605, opened 2025-08-06
fix:Prevent premature task cleanup and correct related tests

whosefriendA:
What type of PR is this?

/kind bug

This PR addresses a subtle bug in the graceful-eviction-controller and corrects the corresponding integration tests that were masking this bug.

The core issue was found in the allScheduledResourceInHealthyState helper function. It would incorrectly return true when a binding had no target clusters in its spec.clusters field. This could lead to a scenario where a graceful eviction task was prematurely finalized simply because the scheduler had not yet assigned the workload to new clusters.

Furthermore, the original integration tests for the Reconcile loop had a flaw that accidentally triggered and masked this bug:

The test case for an "active eviction task" did not define any spec.clusters for the ResourceBinding.

This incomplete test data triggered the bug in allScheduledResourceInHealthyState, causing the controller to incorrectly decide the task was "finished".

This resulted in a RequeueAfter duration of 0.

The test case had a flawed assertion that incorrectly expected RequeueAfter to be 0, causing the test to pass for the wrong reasons.

This PR fixes both the production code and the test code:

It modifies allScheduledResourceInHealthyState to correctly return false when there are no target clusters to check.

It refactors the integration tests for both RBGracefulEvictionController and CRBGracefulEvictionController to:

Provide complete and realistic ResourceBinding objects for test cases.

Assert on the controller's behavior (e.g., expectRequeue: true) rather than asserting on brittle, exact return values.

With these changes, the graceful eviction logic is more robust, and the test suite accurately verifies the controller's intended behavior.

The key takeaway from this PR is the correction of a subtle bug where allScheduledResourceInHealthyState would return a "vacuously true" result. The accompanying test fixes were necessary because the original tests were not only flawed in their assertions but also accidentally triggered and masked this underlying bug. The new tests are now robust and correctly verify the intended controller behavior.

### DIFF
--- a/pkg/controllers/gracefuleviction/evictiontask.go
+++ b/pkg/controllers/gracefuleviction/evictiontask.go
@@ -88,6 +88,10 @@ func assessSingleTask(task workv1alpha2.GracefulEvictionTask, opt assessmentOpti
 }
 
 func allScheduledResourceInHealthyState(opt assessmentOption) bool {
+	if len(opt.scheduleResult) == 0 {
+		return false
+	}
+
 	for _, targetCluster := range opt.scheduleResult {
 		var statusItem *workv1alpha2.AggregatedStatusItem
 
--- a/pkg/controllers/gracefuleviction/crb_graceful_eviction_controller_test.go
+++ b/pkg/controllers/gracefuleviction/crb_graceful_eviction_controller_test.go
@@ -77,14 +77,20 @@ func TestCRBGracefulEvictionController_Reconcile(t *testing.T) {
 							CreationTimestamp: &now,
 						},
 					},
+					Clusters: []workv1alpha2.TargetCluster{
+						{Name: "new-cluster"},
+					},
 				},
 				Status: workv1alpha2.ResourceBindingStatus{
 					SchedulerObservedGeneration: 1,
+					AggregatedStatus: []workv1alpha2.AggregatedStatusItem{
+						{ClusterName: "new-cluster", Health: workv1alpha2.ResourceUnhealthy},
+					},
 				},
 			},
 			expectedResult:  controllerruntime.Result{},
 			expectedError:   false,
-			expectedRequeue: false,
+			expectedRequeue: true,
 		},
 		{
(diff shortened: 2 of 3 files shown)


RainbowMango:
@XiShanYongYe-Chang Please take a look.

pkg/controllers/gracefuleviction/evictiontask.go
```diff
@@ -88,6 +88,10 @@ func assessSingleTask(task workv1alpha2.GracefulEvictionTask, opt assessmentOpti
 }
 
 func allScheduledResourceInHealthyState(opt assessmentOption) bool {
+	if len(opt.scheduleResult) == 0 {
```
XiShanYongYe-Chang, line 91: Hi @whosefriendA, thanks for your contribution.

In fact, this situation has already been taken into account at the function's call site:

https://github.com/karmada-io/karmada/blob/64fbe7560d0b6ce1022f6bb94411107b8a0b04bf/pkg/controllers/gracefuleviction/evictiontask.go#L81-L83

Unlike directly judging the scheduling result as empty, judging the generation excludes a scenario where the scheduling result is cleared due to the inability to find a suitable cluster during rescheduling. Recently, there has been a discussion about whether this behavior is appropriate. 

Currently, I think that this situation does not need to be treated as a bug, but rather as a behavioral adjustment. We can also wait for further feedback from the community users to decide whether to improve it. 

Therefore, I suggest temporarily putting aside the modification of this PR and removing the bug label.

How do you think?

whosefriendA: Hi @XiShanYongYe-Chang ,

Thanks for the detailed feedback and the clarification!

I understand your point that the correct behavior in this edge case is a broader design discussion, making it a 'behavioral adjustment' rather than a simple bug. 

I agree with your suggestion to put this on hold. I'm closing this PR for now and will follow the community discussion on this topic.

XiShanYongYe-Chang: Thanks~ @whosefriendA

PR #6611, opened 2025-08-07
Optimize the ResourceBinding validation webhook for better extensibility

liaolecheng:
**What type of PR is this?**
/kind cleanup

This PR restructures the validating webhook for ResourceBinding to enhance its architecture and extensibility.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/webhook/resourcebinding/validating.go
+++ b/pkg/webhook/resourcebinding/validating.go
@@ -26,7 +26,6 @@ import (
 	admissionv1 "k8s.io/api/admission/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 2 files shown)


RainbowMango:
@seanlaii would you like to take a look?

seanlaii:
/assign
Yes, thanks for the info. Will take a look!

pkg/webhook/resourcebinding/validating.go
```diff
@@ -56,34 +56,85 @@ type frqProcessOutcome struct {
 
 // Handle implements admission.Handler interface.
 func (v *ValidatingAdmission) Handle(ctx context.Context, req admission.Request) admission.Response {
+	rb, oldRB, decodeResp := v.decodeRBs(req)
```
seanlaii:
```suggestion
	rb, oldRB, decodeErrResp := v.decodeRBs(req)
```

pkg/webhook/resourcebinding/validating.go
```diff
@@ -56,34 +56,85 @@ type frqProcessOutcome struct {
...
+	if decodeResp != nil {
+		return *decodeResp
+	}
+	klog.V(2).Infof("Processing ResourceBinding(%s/%s) for request: %s (%s)", rb.Namespace, rb.Name, req.Operation, req.UID)
+
+	var validationMessages []string
+
+	// The response will be nil if validation passes or is skipped.
```
seanlaii:
```suggestion
	// The response will be nil if validation is skipped.
```

pkg/webhook/resourcebinding/validating.go
```diff
@@ -56,34 +56,85 @@ type frqProcessOutcome struct {
...
+			return &resp
+		}
 	}
 
+	isDryRun := req.DryRun != nil && *req.DryRun
+
 	newRbTotalUsage, oldRbTotalUsage, respCalc := v.calculateRBUsages(rb, oldRB)
 	if respCalc != nil {
-		return *respCalc
+		return respCalc
```
seanlaii: Maybe we can also rename this to `calcErrResp`

pkg/webhook/resourcebinding/validating.go
```diff
@@ -56,34 +56,85 @@ type frqProcessOutcome struct {
...
 		klog.V(5).Infof("FederatedQuotaEnforcement feature gate is disabled, skipping validation for ResourceBinding %s/%s", req.Namespace, req.Name)
-		return admission.Allowed("")
+		return nil // Returning nil to allow other validators to run
 	}
 
-	isDryRun := req.DryRun != nil && *req.DryRun
+	if req.Operation == admissionv1.Create && len(rb.Spec.Clusters) == 0 {
+		klog.V(4).Infof("ResourceBinding %s/%s is being created but not yet scheduled, skipping quota validation.", rb.Namespace, rb.Name)
+		resp := admission.Allowed("ResourceBinding not yet scheduled for Create operation.")
+		return &resp
```
seanlaii: Perhaps we can wrap this into a function.

seanlaii (commented):
Hi @liaolecheng , 

Thanks for working on this optimization!

Would you be open to adjusting the function so that it returns an `error` instead of an `admission.Response`, and then wrapping it into `admission.Allowed/Denied/Errored `only in the top-level handler?
I think this could help keep the lower-level functions focused purely on business logic, make them easier to reuse in other places, and centralize the response handling in one place.

But, this would be a relatively larger change.
I realized I didn’t consider this approach when I originally wrote it — apologies for that.

Totally fine if you think this is out of scope for this PR — just wanted to check if you’d be interested in making the change now. 
Other comments are mainly for the original implementation.

I’m also open to discussing this further with the community to see if this aligns with our broader coding conventions.

cc @RainbowMango @zhzhuang-zju

RainbowMango:
> Would you be open to adjusting the function so that it returns an error instead of an admission.Response, and then wrapping it into admission.Allowed/Denied/Errored only in the top-level handler?
I think this could help keep the lower-level functions focused purely on business logic, make them easier to reuse in other places, and centralize the response handling in one place.

+1

liaolecheng:
@seanlaii Thank you for the suggestion. I completely agree and will revise it accordingly.

pkg/webhook/resourcebinding/validating.go
liaolecheng: Could you kindly share your thoughts on whether the information is necessary in cases where access is allowed? I’ve reviewed other verification processes, and it seems that most "allowed" scenarios don’t include such information. If this information is required, it might mean that many functions would need to return both a message and an error value, which would make the code structure feel less optimal. Thank you for your input!

liaolecheng:
> Would you be open to adjusting the function so that it returns an `error` instead of an `admission.Response`, and then wrapping it into `admission.Allowed/Denied/Errored `only in the top-level handler?
> I think this could help keep the lower-level functions focused purely on business logic, make them easier to reuse in other places, and centralize the response handling in one place.

 @seanlaii I've written a version following this approach. Would you have a moment to take a look? It's worth noting that there's no additional information for the `admission.Allowed` case.

RainbowMango:
@liaolecheng please help resolve the conflicts as the #6481 just merged.

pkg/webhook/resourcebinding/validating.go
seanlaii: The message is not necessary in my opinion. If keeping it is less optimal, I would suggest keeping your current implementation. Thanks!

pkg/webhook/resourcebinding/validating.go
```diff
@@ -314,29 +323,15 @@ func (v *ValidatingAdmission) processSingleFRQ(frqItem policyv1alpha1.FederatedR
 	if !isAllowed {
 		klog.Warningf("Quota exceeded for FederatedResourceQuota %s/%s. ResourceBinding %s/%s will be denied.",
 			frqItem.Namespace, frqItem.Name, rbNamespace, rbName)
-		resp := buildDenyResponse(errMsg)
-		return nil, "", resp
+		klog.Warning(errMsg)
```
seanlaii: Maybe we could merge this with the above log.

pkg/webhook/resourcebinding/validating.go
```diff
@@ -50,41 +49,70 @@ var _ admission.Handler = &ValidatingAdmission{}
 type frqProcessOutcome struct {
 	validationMessages []string
 	updatedFRQCount    int
-	noFRQsFound        bool                // True if this attempt found no FRQs
-	earlyExitResponse  *admission.Response // Response if this attempt decided to exit early (denial, permanent error)
-	ProcessError       error               // Error for RetryOnConflict (e.g., conflict error or nil for success/final decision)
+	noFRQsFound        bool  // True if this attempt found no FRQs
+	earlyExitResponse  error // Response if this attempt decided to exit early (denial, permanent error)
```
seanlaii: perhaps we could change this to `earlyExitError`?

pkg/webhook/resourcebinding/validating.go
```diff
@@ -121,7 +149,7 @@ func (v *ValidatingAdmission) executeFRQProcessingAttempt(ctx context.Context, r
 	}
 
 	for _, frqItem := range frqList.Items {
-		newStatus, msg, denialResp := v.processSingleFRQ(frqItem, rb.Namespace, rb.Name, totalRbDelta)
+		newStatus, msg, denialResp := v.processSingleFRQ(&frqItem, rb.Namespace, rb.Name, totalRbDelta)
```
seanlaii: maybe we could change it to denialError
listResp could be changed as well

seanlaii (commented):
Thanks! Looks great overall—thanks for addressing the earlier feedback.
I just have a couple of naming nits before we merge.

pkg/webhook/resourcebinding/validating.go
```diff
@@ -155,20 +182,20 @@ func (v *ValidatingAdmission) executeFRQProcessingAttempt(ctx context.Context, r
 	return outcome // attemptError is nil, successful attempt
 }
 
-func (v *ValidatingAdmission) finalizeResponse(rb *workv1alpha2.ResourceBinding, outcome frqProcessOutcome) admission.Response {
+func (v *ValidatingAdmission) finalizeResponse(rb *workv1alpha2.ResourceBinding, outcome frqProcessOutcome) error {
```
seanlaii:
```suggestion
func (v *ValidatingAdmission) handleFRQOutcome(rb *workv1alpha2.ResourceBinding, outcome frqProcessOutcome) error {
```

seanlaii (commented):
Looks good, thanks! I left one nitpick.
/lgtm
cc @RainbowMango for a final check.

RainbowMango (approved):
/lgtm
/approve

/retest
unrelated failures.

PR #6631, opened 2025-08-13
Add feature gates support to karmada-scheduler config (Helm Chart)

seanlaii:
**What type of PR is this?**
/kind feature

This PR adds feature gates support to the `karmada-scheduler` Helm Chart configuration. Users can now configure feature gates for the scheduler component via the `scheduler.featureGates` field in `values.yaml`, similar to other components like [controllerManager](https://github.com/karmada-io/karmada/blob/master/charts/karmada/values.yaml#L353) and [schedulerEstimator](https://github.com/karmada-io/karmada/blob/master/charts/karmada/values.yaml#L890). This provides a consistent and flexible way to enable or disable scheduler features across all installation modes.

Fixes #

```release-note
`Helm Chart`: Users can now configure feature gates for karmada-scheduler via the Helm Chart using the `scheduler.featureGates` field in values.yaml.
```

### DIFF
--- a/charts/karmada/templates/_helpers.tpl
+++ b/charts/karmada/templates/_helpers.tpl
@@ -556,18 +556,22 @@ Return the proper Docker Image Registry Secret Names
 {{ include "common.images.pullSecrets" (dict "images" (list .Values.cfssl.image .Values.kubectl.image .Values.etcd.internal.image .Values.agent.image .Values.apiServer.image .Values.controllerManager.image .Values.descheduler.image .Values.schedulerEstimator.image .Values.scheduler.image .Values.webhook.image .Values.aggregatedApiServer.image .Values.metricsAdapter.image .Values.search.image .Values.kubeControllerManager.image) "global" .Values.global) }}
 {{- end -}}
 
+{{- /* 
+Generate the --feature-gates command line argument for karmada-controllerManager.
+Iterates over .Values.controllerManager.featureGates and constructs a comma-separated key=value list.
+If any feature gates are set, outputs: --feature-gates=Foo=true,Bar=false
+If none are set, outputs nothing.
+*/ -}}
 {{- define "karmada.controllerManager.featureGates" -}}
-     {{- if (not (empty .Values.controllerManager.featureGates)) }}
-          {{- $featureGatesFlag := "" -}}
+     {{- if .Values.controllerManager.featureGates -}}
+          {{- $featureGates := list -}}
           {{- range $key, $value := .Values.controllerManager.featureGates -}}
-               {{- if not (empty (toString $value)) }}
-                    {{- $featureGatesFlag = cat $featureGatesFlag $key "=" $value ","  -}}
+               {{- if not (empty (toString $value)) -}}
+                    {{- $featureGates = append $featureGates (printf "%s=%s" $key (toString $value)) -}}
                {{- end -}}
           {{- end -}}
-
-          {{- if gt (len $featureGatesFlag) 0 }}
-               {{- $featureGatesFlag := trimSuffix "," $featureGatesFlag  | nospace -}}
-               {{- printf "%s=%s" "--feature-gates" $featureGatesFlag -}}
+          {{- if $featureGates -}}
+               {{- printf "--feature-gates=%s" (join "," $featureGates) | nospace -}}
           {{- end -}}
      {{- end -}}
 {{- end -}}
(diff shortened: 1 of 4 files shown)


zhzhuang-zju:
Thanks~
/assign

charts/karmada/values.yaml
```diff
@@ -299,11 +301,11 @@ webhook:
     rollingUpdate:
       maxUnavailable: 0
       maxSurge: 50%
-  ## @param apiServer.podDisruptionBudget
+  ## @param webhook.podDisruptionBudget
```
zhzhuang-zju, line 304: nice catch

charts/karmada/templates/_helpers.tpl
```diff
@@ -594,6 +594,28 @@ If none are set, outputs nothing.
...
+                    {{- $featureGatesFlag = cat $featureGatesFlag $key "=" $value ","  -}}
+               {{- end -}}
+          {{- end -}}
+
+          {{- if gt (len $featureGatesFlag) 0 }}
+               {{- $featureGatesFlag := trimSuffix "," $featureGatesFlag  | nospace -}}
+               {{- printf "%s=%s" "--feature-gates" $featureGatesFlag -}}
+          {{- end -}}
+     {{- end -}}
+{{- end -}}
```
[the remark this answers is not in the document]
zhzhuang-zju: Hi @seanlaii, can you confirm whether this suggestion is feasible?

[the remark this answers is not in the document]
seanlaii: Sure! It is feasible, but we need to add another `toString` in this line `{{- $featureGates = append $featureGates (printf "%s=%s" $key $value) -}}` to handle boolean values.
I have changed all the parsing logic with this, which uses the `helm` function to improve readability.

charts/karmada/templates/karmada-scheduler-estimator.yaml
```diff
@@ -55,9 +55,9 @@ spec:
             - --grpc-client-ca-file=/etc/karmada/pki/server-ca.crt
             - --metrics-bind-address=$(POD_IP):8080
             - --health-probe-bind-address=$(POD_IP):10351
-            {{- with (include "karmada.schedulerEstimator.featureGates" (dict "featureGatesArg" $.Values.schedulerEstimator.featureGates)) }}
+            {{- with (include "karmada.schedulerEstimator.featureGates" $) }}
```
seanlaii, line 63: Need to use `$` to pass in the root context so that the `_helpers.tql` can find the `featureGates` values.

seanlaii:
Have tested the following test cases on `webhook`, `scheduler`, `controllerManager`, and `schedulerEstimator`:

```
featureGates: {}  --> empty
```
---
```
featureGates:     --> empty
```
---
```
featureGates:     --> `--feature-gates=featureA=true`
  featureA: true
```
---
```
featureGates:     --> `--feature-gates=featureA=true,featureB=false`
  featureA: true
  featureB: "false"
```
---
```
featureGates:     --> `--feature-gates=featureA=false`
  featureA: true
  featureA: "false"
```
The last value will override the first one.

charts/karmada/templates/karmada-scheduler-estimator.yaml
zhzhuang-zju: Oh, I see. Here is within the `range` loop,  `.` refers to the current iteration's context (the key-value pair), while `$` refers to the root template context where `.Values` is accessible.

zhzhuang-zju: @seanlaii Could we add a comment here to explain why we use `$` instead of `.`, to prevent future incorrect changes from `$` to `.`? To be honest, before understanding your intention, I was inclined to use `.` for consistency with other similar implementations, but that would actually be incorrect. : )

PS: Our current CI probably can't test this scenario, so comments are an important means to ensure correctness at this point.

zhzhuang-zju:
Thanks @seanlaii for your great work! I have no further comment beside the https://github.com/karmada-io/karmada/pull/6631#discussion_r2278407335. Once it fix, we can move this feature forward~

charts/karmada/templates/karmada-scheduler-estimator.yaml
seanlaii: Sure! Added. Thanks for the feedback. 
Actually, the CI did help capture the error: https://github.com/karmada-io/karmada/actions/runs/16979605684/job/48136616572#step:6:53.
I also tried using `.` previously, and got this error:
```
nil pointer evaluating interface {}.schedulerEstimator 
```

zhzhuang-zju:
Thanks
/lgtm
/approve

zhzhuang-zju:
Hi @a7i, do you want to take a look?

PR #894, opened 2021-11-01
add Timeout in WaitForCacheSync

mrlihanbo:
Signed-off-by: lihanbo <lihanbo2@huawei.com>

**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
There exists a scenario that some of joined cluster are unhealth in karmada. At the moment, if karmada controller manager restart, it will be blocked in `WaitForCacheSync` process until the unhealth cluster recovers. 

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:
"NONE"

PR #6647, opened 2025-08-14
Fix unreasonable informer manager stop behavior

### LINKED ISSUE #6598: Revisit Current Informer Stop Behavior
Karmada, as a multi-cluster project, leverages the informer mechanism to handle resources from different member clusters as well as control plane resources. This approach significantly reduces the pressure on both karmada-apiserver and member cluster kube-apiservers.

Many informers are dynamically loaded, which greatly increases the complexity of informer management. Among these operations, `informer.Stop()` is particularly high-risk.

The current codebase contains various `informer.Stop()` calls that may have inconsistent or potentially problematic behavior patterns. This issue aims to comprehensively review and refactor the `informer.Stop()` behavior.

This presents an example of problematic informer shutdown behavior. Both the `execution controller` and `serviceExportController` share the same `MultiClusterInformerManager` instance obtained via `genericmanager.GetInstance()`. 
https://github.com/karmada-io/karmada/blob/b8f6874c5896b877b85b528f4e4b13ec0b59e615/pkg/util/fedinformer/genericmanager/multi-cluster-manager.go#L35-L40
When the `serviceExportController` encounters cache synchronization failures or timeouts for `serviceexports` and `endpointslices` resources, it terminates all informers for the affected cluster. 
https://github.com/karmada-io/karmada/blob/b8f6874c5896b877b85b528f4e4b13ec0b59e615/pkg/controllers/mcs/service_export_controller.go#L275-L290
This shutdown operation has a cascading effect on other controllers that are actively utilizing informers for the same cluster, thereby broadening the impact radius beyond the original failing component.

**Why is this needed**:

zhzhuang-zju:
**What type of PR is this?**
/kind cleanup

Ref to #6598, the unreasonable SingleClusterInformerManager stop behavior has a cascading effect on other controllers that are actively utilizing informers for the same cluster, thereby broadening the impact radius beyond the original failing component. So, in this pr, we try to remove the unreasonable SingleClusterInformerManager stop behaviors.

Fixed behavior:

1. pkg/controllers/mcs/service_export_controller.go#L275-L290
2. pkg/controllers/multiclusterservice/endpointslice_collect_controller.go#L212-L229
3. pkg/controllers/status/work_status_controller.go#L488-L502
4. pkg/controllers/status/cluster_status_controller.go#L380-L395

When `InformerManager.WaitForCacheSyncWithTimeout` returns a result that doesn't contain the target GVR, it will return an error and re-enter reconcile.

5. pkg/controllers/federatedhpa/federatedhpa_controller.go#L569-L582

When a single cluster's `InformerManager.WaitForCacheSyncWithTimeout` returns a result that doesn't contain the target GVR, this error will be ignored unless all target clusters encounter similar issues. The federated_hpa_controller will re-enter reconcile every HorizontalPodAutoscalerSyncPeriod, so it has the opportunity to re-execute `WaitForCacheSyncWithTimeout`.

In summary, all the above controllers have the opportunity to retry when the `WaitForCacheSyncWithTimeout` result doesn't meet expectations. And retries should be able to resolve most failures, except for the issues documented in https://github.com/karmada-io/karmada/issues/6598#issuecomment-3154216918. However, this issue was not introduced by this PR, so it's not considered in this PR.

Fixes #6598

```release-note
`karmada-controller-manager/karmada-agent`:  Fixed the issue that informer cache sync failures for specific resources incorrectly shut down all informers for that cluster, affecting resource distribution and work status synchronization.
```

### DIFF
--- a/pkg/controllers/status/cluster_status_controller.go
+++ b/pkg/controllers/status/cluster_status_controller.go
@@ -390,7 +390,6 @@ func (c *ClusterStatusController) buildInformerForCluster(clusterClient *util.Cl
 		return nil
 	}(); err != nil {
 		klog.ErrorS(err, "Failed to sync cache for cluster", "cluster", clusterClient.ClusterName)
-		c.TypedInformerManager.Stop(clusterClient.ClusterName)
 		return nil, err
 	}
 
--- a/pkg/controllers/federatedhpa/federatedhpa_controller.go
+++ b/pkg/controllers/federatedhpa/federatedhpa_controller.go
@@ -577,7 +577,6 @@ func (c *FHPAController) buildPodInformerForCluster(clusterScaleClient *util.Clu
 		return nil
 	}(); err != nil {
 		klog.ErrorS(err, "Failed to sync cache for cluster", "cluster", clusterScaleClient.ClusterName)
-		c.TypedInformerManager.Stop(clusterScaleClient.ClusterName)
 		return nil, err
 	}
 
--- a/pkg/controllers/mcs/service_export_controller.go
+++ b/pkg/controllers/mcs/service_export_controller.go
@@ -285,7 +285,6 @@ func (c *ServiceExportController) registerInformersAndStart(cluster *clusterv1al
 		return nil
 	}(); err != nil {
 		klog.ErrorS(err, "Failed to sync cache for cluster", "cluster", cluster.Name)
-		c.InformerManager.Stop(cluster.Name)
 		return err
 	}
 
--- a/pkg/controllers/multiclusterservice/endpointslice_collect_controller.go
+++ b/pkg/controllers/multiclusterservice/endpointslice_collect_controller.go
@@ -224,7 +224,6 @@ func (c *EndpointSliceCollectController) registerInformersAndStart(cluster *clus
 		return nil
 	}(); err != nil {
 		klog.ErrorS(err, "Failed to sync cache for cluster", "cluster", cluster.Name)
-		c.InformerManager.Stop(cluster.Name)
 		return err
 	}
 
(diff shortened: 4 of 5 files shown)


RainbowMango (commented):
/assign

pkg/controllers/status/cluster_status_controller.go
```diff
@@ -390,7 +390,6 @@ func (c *ClusterStatusController) buildInformerForCluster(clusterClient *util.Cl
 		return nil
 	}(); err != nil {
 		klog.ErrorS(err, "Failed to sync cache for cluster", "cluster", clusterClient.ClusterName)
-		c.TypedInformerManager.Stop(clusterClient.ClusterName)
```
RainbowMango, line 393: PS:
The background of why it stops the informer is for disabling annoying warning logs :( 
https://github.com/karmada-io/karmada/pull/894#discussion_r1045377367

Obviously, it's not the right approach.

RainbowMango (commented):
/approve

/hold
cc @lyuyun to take another look.

@snowplayfire We introduced `--cluster-cache-sync-timeout` flag at https://github.com/karmada-io/karmada/pull/1112, but it seems not enough, the fundamental flaw is due to close the entire informer factory abruptly . 

While adjusting timeout parameters may avoid this same issue described at #1112, but other users will encounter this problem again, because they might don't know this flag.

I don't know the configuration of `--cluster-cache-sync-timeout` in your environment, but I'm sure this issue is still concerning, for both you and other users. 

Any thoughs?

RainbowMango:
/kind bug
@zhzhuang-zju Please add a release note and this should be backported to release branches.

zhzhuang-zju:
> /kind bug @zhzhuang-zju Please add a release note and this should be backported to release branches.

Done~ I'll submit an issue to track the backport tasks once this pr is merged

PR #6656, opened 2025-08-16
fixed the issue where pp-related claimMetadata was not properly cleaned up

### LINKED ISSUE #6513: ClaimMetadata not properly cleaned on deletion of Lazy activationPreference propagationpolicy

zhzhuang-zju:
**What type of PR is this?**
/kind bug

Fixed the issue where pp-related claimMetadata was not properly cleaned up

Fixes #6513

```release-note
`karmada-controller-manager`: Fixed the issue where pp-related claimMetadata was not properly cleaned up when deleting propagationpolicy with Lazy activationPreference.
```

### DIFF
--- a/pkg/detector/detector.go
+++ b/pkg/detector/detector.go
@@ -690,6 +690,64 @@ func (d *ResourceDetector) GetUnstructuredObject(objectKey keys.ClusterWideKey) 
 	return unstructuredObj.DeepCopy(), nil
 }
 
+// fetchResourceBinding fetches a ResourceBinding from the client or dynamic client.
+func (d *ResourceDetector) fetchResourceBinding(ctx context.Context, rbNamespace, rbName string) (*workv1alpha2.ResourceBinding, error) {
+	// First try to get ResourceBinding using cached client
+	rb := &workv1alpha2.ResourceBinding{}
+	err := d.Client.Get(ctx, client.ObjectKey{Namespace: rbNamespace, Name: rbName}, rb)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			// If not found with client, try using dynamic client
+			gvr := workv1alpha2.SchemeGroupVersion.WithResource(workv1alpha2.ResourcePluralResourceBinding)
+			unstructuredRB, dynamicErr := d.DynamicClient.Resource(gvr).Namespace(rbNamespace).Get(ctx, rbName, metav1.GetOptions{})
+			if dynamicErr != nil {
+				return nil, dynamicErr
+			}
+
+			// Convert unstructured to ResourceBinding
+			if err = helper.ConvertToTypedObject(unstructuredRB, rb); err != nil {
+				klog.Errorf("Failed to convert unstructured to ResourceBinding(%s/%s): %v", rbNamespace, rbName, err)
+				return nil, err
+			}
+
+			return rb, nil
+		}
+		klog.Errorf("Failed to get ResourceBinding(%s/%s): %v", rbNamespace, rbName, err)
+		return nil, err
+	}
+
+	return rb, nil
+}
+
+// fetchClusterResourceBinding fetches a ClusterResourceBinding from the client or dynamic client.
+func (d *ResourceDetector) fetchClusterResourceBinding(ctx context.Context, crbName string) (*workv1alpha2.ClusterResourceBinding, error) {
+	// First try to get ClusterResourceBinding using cached client
+	crb := &workv1alpha2.ClusterResourceBinding{}
+	err := d.Client.Get(ctx, client.ObjectKey{Name: crbName}, crb)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			// If not found with client, try using dynamic client
+			gvr := workv1alpha2.SchemeGroupVersion.WithResource(workv1alpha2.ResourcePluralClusterResourceBinding)
+			unstructuredRB, dynamicErr := d.DynamicClient.Resource(gvr).Get(ctx, crbName, metav1.GetOptions{})
+			if dynamicErr != nil {
+				return nil, dynamicErr
+			}
+
+			// Convert unstructured to ClusterResourceBinding
+			if err = helper.ConvertToTypedObject(unstructuredRB, crb); err != nil {
+				klog.Errorf("Failed to convert unstructured to ClusterResourceBinding(%s): %v", crbName, err)
+				return nil, err
+			}
+
+			return crb, nil
+		}
+		klog.Errorf("Failed to get ClusterResourceBinding(%s): %v", crbName, err)
+		return nil, err
+	}
+
+	return crb, nil
+}
+
 // ClaimPolicyForObject set policy identifier which the object associated with.
 // It will add policy labels and annotations to the object, and update the obj with the content returned by the Server, ensuring that the obj used when generating the binding is the latest.
 func (d *ResourceDetector) ClaimPolicyForObject(object *unstructured.Unstructured, policy *policyv1alpha1.PropagationPolicy) (string, error) {
@@ -940,7 +998,7 @@ func (d *ResourceDetector) ReconcilePropagationPolicy(key util.QueueKey) error {
 
 	if !propagationObject.DeletionTimestamp.IsZero() {
 		klog.Infof("PropagationPolicy(%s) is being deleted.", nkey.NamespaceKey())
-		if err = d.HandlePropagationPolicyDeletion(propagationObject.Labels[policyv1alpha1.PropagationPolicyPermanentIDLabel]); err != nil {
+		if err = d.HandlePropagationPolicyDeletion(propagationObject.Labels[policyv1alpha1.PropagationPolicyPermanentIDLabel], propagationObject.Spec.ResourceSelectors); err != nil {
 			return err
 		}
 		if controllerutil.RemoveFinalizer(propagationObject, util.PropagationPolicyControllerFinalizer) {
@@ -1012,7 +1070,7 @@ func (d *ResourceDetector) ReconcileClusterPropagationPolicy(key util.QueueKey) 
 
 	if !propagationObject.DeletionTimestamp.IsZero() {
 		klog.Infof("ClusterPropagationPolicy(%s) is being deleted.", nkey.NamespaceKey())
-		if err = d.HandleClusterPropagationPolicyDeletion(propagationObject.Labels[policyv1alpha1.ClusterPropagationPolicyPermanentIDLabel]); err != nil {
+		if err = d.HandleClusterPropagationPolicyDeletion(propagationObject.Labels[policyv1alpha1.ClusterPropagationPolicyPermanentIDLabel], propagationObject.Spec.ResourceSelectors); err != nil {
 			return err
 		}
 		if controllerutil.RemoveFinalizer(propagationObject, util.ClusterPropagationPolicyControllerFinalizer) {
@@ -1033,32 +1091,26 @@ func (d *ResourceDetector) ReconcileClusterPropagationPolicy(key util.QueueKey) 
 // the resource template a change to match another policy).
 //
 // Note: The relevant ResourceBinding will continue to exist until the resource template is gone.
-func (d *ResourceDetector) HandlePropagationPolicyDeletion(policyID string) error {
+func (d *ResourceDetector) HandlePropagationPolicyDeletion(policyID string, resources []policyv1alpha1.ResourceSelector) error {
 	claimMetadata := labels.Set{policyv1alpha1.PropagationPolicyPermanentIDLabel: policyID}
-	rbs, err := helper.GetResourceBindings(d.Client, claimMetadata)
-	if err != nil {
-		klog.Errorf("Failed to list propagation bindings with policy permanentID(%s): %v", policyID, err)
-		return err
-	}
-
 	var errs []error
-	for index, binding := range rbs.Items {
-		// Must remove the claim metadata, such as labels and annotations, from the resource template ahead of ResourceBinding,
-		// otherwise might lose the chance to do that in a retry loop (in particular, the claim metadata was successfully removed
-		// from ResourceBinding, but resource template not), since the ResourceBinding will not be listed again.
-		if err := d.CleanupResourceTemplateClaimMetadata(binding.Spec.Resource, claimMetadata, CleanupPPClaimMetadata); err != nil {
-			klog.Errorf("Failed to clean up claim metadata from resource(%s-%s/%s) when propagationPolicy removed, error: %v",
-				binding.Spec.Resource.Kind, binding.Spec.Resource.Namespace, binding.Spec.Resource.Name, err)
-			errs = append(errs, err)
-			// Skip cleaning up policy labels and annotations from ResourceBinding, give a chance to do that in a retry loop.
-			continue
+	for _, resource := range util.ExtractUniqueNamespacedSelectors(resources) {
+		objRef := workv1alpha2.ObjectReference{
+			APIVersion: resource.APIVersion,
+			Kind:       resource.Kind,
+			Namespace:  resource.Namespace,
 		}
 
-		// Clean up the claim metadata from the reference binding so that the karmada scheduler won't reschedule the binding.
-		if err := d.CleanupResourceBindingClaimMetadata(&rbs.Items[index], claimMetadata, CleanupPPClaimMetadata); err != nil {
-			klog.Errorf("Failed to clean up claim metadata from resource binding(%s/%s) when propagationPolicy removed, error: %v",
-				binding.Namespace, binding.Name, err)
+		rawObjects, err := helper.FetchResourceTemplatesByLabelSelector(d.DynamicClient, d.InformerManager, d.RESTMapper, objRef, labels.SelectorFromSet(claimMetadata))
+		if err != nil {
 			errs = append(errs, err)
+			continue
+		}
+		for _, rawObject := range rawObjects {
+			err := d.handleResourceTemplateAndBindingCleanup(rawObject, objRef, claimMetadata, CleanupPPClaimMetadata)
+			if err != nil {
+				errs = append(errs, err)
+			}
 		}
 	}
 	return errors.NewAggregate(errs)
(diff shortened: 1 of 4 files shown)


pkg/detector/detector.go
```diff
@@ -955,6 +955,12 @@ func (d *ResourceDetector) ReconcilePropagationPolicy(key util.QueueKey) error {
 		if err = d.HandlePropagationPolicyDeletion(propagationObject.Labels[policyv1alpha1.PropagationPolicyPermanentIDLabel]); err != nil {
 			return err
 		}
+		if util.IsLazyActivationEnabled(propagationObject.Spec.ActivationPreference) {
+			claimMetadata := labels.Set{policyv1alpha1.PropagationPolicyPermanentIDLabel: propagationObject.Labels[policyv1alpha1.PropagationPolicyPermanentIDLabel]}
+			if err = d.cleanupResidualResourceTemplateClaimMetadata(claimMetadata, propagationObject.Spec.ResourceSelectors, CleanupPPClaimMetadata); err != nil {
+				return err
+			}
+		}
```
RainbowMango: Is there any particular reason not to put this logic into `HandlePropagationPolicyDeletion()`?

The current approach would miss the chance to clean up the claimed label when reconciling the resource template: 
https://github.com/karmada-io/karmada/blob/d0f74642463df113fb1cd207e8662074bdd39a78/pkg/detector/policy.go#L116.

pkg/detector/detector.go
```diff
@@ -1143,6 +1161,69 @@ func (d *ResourceDetector) HandleClusterPropagationPolicyDeletion(policyID strin
...
+					workload.GetKind(), workload.GetNamespace(), workload.GetName(), err)
+				errs = append(errs, err)
+			}
+		}
+	}
+
+	return errors.NewAggregate(errs)
+}
+
+func (d *ResourceDetector) cleanupResourceTemplateClaimMetadata(gvr schema.GroupVersionResource, obj *unstructured.Unstructured, targetClaimMetadata map[string]string, cleanupFunc func(obj metav1.Object)) error {
```
RainbowMango: The methods `cleanupResourceTemplateClaimMetadata` and `CleanupResourceTemplateClaimMetadata` are ambiguous; why not just use the previous one? Seems all you need to build a `workv1alpha2.ObjectReference`.

pkg/detector/detector.go
```diff
@@ -1143,6 +1161,69 @@ func (d *ResourceDetector) HandleClusterPropagationPolicyDeletion(policyID strin
 	return errors.NewAggregate(errs)
 }
 
+func (d *ResourceDetector) cleanupResidualResourceTemplateClaimMetadata(targetClaimMetadata labels.Set, resourceSelector []policyv1alpha1.ResourceSelector, cleanupFunc func(obj metav1.Object)) error {
```
RainbowMango: Need another name for clarity. 

`cleanupResidualResourceTemplateClaimMetadata` similar to `cleanupResourceTemplateClaimMetadata`

pkg/detector/detector.go
zhzhuang-zju: > Is there any particular reason not to put this logic into HandlePropagationPolicyDeletion()?

Yes, the input parameters of function `HandlePropagationPolicyDeletion` do not include `ActivationPreference` information. Given that the current approach is only an interim solution, I don't want to modify the existing `HandlePropagationPolicyDeletion`.

> The current approach would miss the chance to clean up the claimed label when reconciling the resource template:

Good point! I also noticed this, but I think this is not an issue with the new logic, but rather that when the resource template detects that the pp has been deleted, it should not call `HandlePropagationPolicyDeletion` in the first place - it's expensive and unnecessary. The resource template should focus on its own cleanup. When the resource template detects that the pp is no longer present, it just needs to clean up its own remaining labels. Actually, this approach is more reasonable. First: it reduces the overhead of calling `HandlePropagationPolicyDeletion` and avoids redundant cleanup operations for the same set of objects; Second: if resource binding generation fails, it won't block the resource template's own cleanup.

pkg/detector/detector.go
zhzhuang-zju: > why not just use the previous one? Seems all you need to build a workv1alpha2.ObjectReference.

Because the current `CleanupResourceTemplateClaimMetadata` first uses a `GET` request to retrieve the ResourceTemplate and then `Update` it. However, the new logic retrieves all ResourceTemplates through `List` and then cleans them up one by one. Therefore, `GET` requests are not needed.

pkg/detector/detector.go
```diff
@@ -1143,6 +1167,68 @@ func (d *ResourceDetector) HandleClusterPropagationPolicyDeletion(policyID strin
...
+							if apierrors.IsNotFound(getErr) {
+								return nil
+							}
+							return getErr
+						}
+						workload = newWorkload
+					}
+					return err
+				}
+				klog.V(2).Infof("Updated resource template(kind=%s, %s/%s) successfully", workload.GetKind(), workload.GetNamespace(), workload.GetName())
```
RainbowMango:
```suggestion
				klog.V(2).Infof("Clean claimed label for resource template(kind=%s, %s/%s) successfully", workload.GetKind(), workload.GetNamespace(), workload.GetName())
```

zhzhuang-zju:
cc @gemini-code-assist

RainbowMango (commented):
/assign

pkg/detector/policy.go
```diff
@@ -110,11 +110,19 @@ func (d *ResourceDetector) getAndApplyPolicy(object *unstructured.Unstructured,
 	resourceChangeByKarmada bool, policyNamespace, policyName, claimedID string) error {
 	matchedPropagationPolicy := &policyv1alpha1.PropagationPolicy{}
 	err := d.Client.Get(context.TODO(), client.ObjectKey{Namespace: policyNamespace, Name: policyName}, matchedPropagationPolicy)
-	if err != nil {
-		if apierrors.IsNotFound(err) {
-			klog.V(4).Infof("PropagationPolicy(%s/%s) has been removed.", policyNamespace, policyName)
-			return d.HandlePropagationPolicyDeletion(claimedID)
+	if apierrors.IsNotFound(err) || (err == nil && !matchedPropagationPolicy.DeletionTimestamp.IsZero()) {
+		// Policy not found or being deleted, clean up metadata
```
RainbowMango: > (err == nil && !matchedPropagationPolicy.DeletionTimestamp.IsZero()

It does not sound right to get involved in the lifecycle management of PropagationPolicy. 
When PropagationPolicy is deleted, there is a workflow to clean up labels on resource templates.

pkg/detector/detector.go
```diff
@@ -690,6 +690,64 @@ func (d *ResourceDetector) GetUnstructuredObject(objectKey keys.ClusterWideKey)
 	return unstructuredObj.DeepCopy(), nil
 }
 
+// fetchResourceBinding fetches a ResourceBinding from the client or dynamic client.
+func (d *ResourceDetector) fetchResourceBinding(rbNamespace, rbName string) (*workv1alpha2.ResourceBinding, error) {
+	// First try to get ResourceBinding using client
```
RainbowMango:
```suggestion
	// First try to get ResourceBinding using cached client
```

pkg/detector/detector.go
```diff
@@ -690,6 +690,64 @@ func (d *ResourceDetector) GetUnstructuredObject(objectKey keys.ClusterWideKey)
...
+			unstructuredRB, dynamicErr := d.DynamicClient.Resource(gvr).Namespace(rbNamespace).Get(context.TODO(), rbName, metav1.GetOptions{})
+			if dynamicErr != nil {
+				return nil, dynamicErr
+			}
+
+			// Convert unstructured to ResourceBinding
+			if err := helper.ConvertToTypedObject(unstructuredRB, rb); err != nil {
+				klog.Errorf("Failed to convert unstructured to ResourceBinding(%s/%s): %v", rbNamespace, rbName, err)
+				return nil, err
+			}
```
RainbowMango: We might need to introduce a typed client for retrieving RB/CRB resources.
We can take this as a follow up.

pkg/detector/detector.go
```diff
@@ -690,6 +690,64 @@ func (d *ResourceDetector) GetUnstructuredObject(objectKey keys.ClusterWideKey)
...
+
+			return rb, nil
+		}
+		klog.Errorf("Failed to get ResourceBinding(%s/%s): %v", rbNamespace, rbName, err)
+		return nil, err
+	}
+
+	return rb.DeepCopy(), nil
```
RainbowMango: Do we have to return a copy? 
The `Client` returns a copy by default.

pkg/detector/detector.go
```diff
@@ -690,6 +690,64 @@ func (d *ResourceDetector) GetUnstructuredObject(objectKey keys.ClusterWideKey)
...
+}
+
+// fetchClusterResourceBinding fetches a ClusterResourceBinding from the client or dynamic client.
+func (d *ResourceDetector) fetchClusterResourceBinding(crbName string) (*workv1alpha2.ClusterResourceBinding, error) {
```
RainbowMango: ditto.

pkg/detector/detector.go
zhzhuang-zju: Thank you for correcting me.

https://github.com/karmada-io/karmada/blob/17caa1bc7d8b27ad9ccd99d5ddbd20f6b95af5b0/vendor/sigs.k8s.io/controller-runtime/pkg/cache/internal/cache_reader.go#L84-L90

pkg/detector/policy.go
zhzhuang-zju: I'm afraid we need to pay attention to whether `matchedPropagationPolicy.DeletionTimestamp` is nil.
Let me give an example:
Suppose a policy matches 1000 deployments. When the policy is being deleted, only 999 deployments are listed, so only the claimMetadata of these 999 deployments are cleaned up. The last deployment enters reconciliation, but at this time, since the policy deletion takes a relatively long time, the policy still exists, and thus this deployment will not enter the cleanup process. As a result, the policy-related claimMetadata of this deployment will remain indefinitely until the next reconciliation occurs.

RainbowMango: We can check the state of the PropagationPolicy before applying it; if it is being deleted, we can return and wait   . Again, it is not the responsibility and shouldn't get involved in the lifecycle management of PropagationPolicy.

zhzhuang-zju: Thanks for the suggestion! Update as recommened

pkg/detector/policy.go
```diff
@@ -110,14 +110,28 @@ func (d *ResourceDetector) getAndApplyPolicy(object *unstructured.Unstructured,
 	resourceChangeByKarmada bool, policyNamespace, policyName, claimedID string) error {
 	matchedPropagationPolicy := &policyv1alpha1.PropagationPolicy{}
 	err := d.Client.Get(context.TODO(), client.ObjectKey{Namespace: policyNamespace, Name: policyName}, matchedPropagationPolicy)
-	if err != nil {
-		if apierrors.IsNotFound(err) {
-			klog.V(4).Infof("PropagationPolicy(%s/%s) has been removed.", policyNamespace, policyName)
-			return d.HandlePropagationPolicyDeletion(claimedID)
+	if apierrors.IsNotFound(err) {
+		// Policy not found or being deleted, clean up metadata
```
RainbowMango: The Policy now is not being deleted.

```suggestion
		// The expected PropagationPolicy has been removed, indicating the cleanup process has completed.
		// However, if the claimed data still exists in the resource template, it likely means the cleanup
		// for the resource template was missed. This can happen if the claimed data was added to the resource
		// template and then the PropagationPolicy was removed very shortly after.
		// When cleaning up resource templates based on PropagationPolicy deletion, the cache may not have received
		// the resource template update in time, resulting in a missed cleanup.
```
Please check if the reason on above that we need to do the cleanup.

pkg/detector/policy.go
```diff
@@ -142,15 +156,28 @@ func (d *ResourceDetector) getAndApplyClusterPolicy(object *unstructured.Unstruc
 	resourceChangeByKarmada bool, policyName, policyID string) error {
 	matchedClusterPropagationPolicy := &policyv1alpha1.ClusterPropagationPolicy{}
 	err := d.Client.Get(context.TODO(), client.ObjectKey{Name: policyName}, matchedClusterPropagationPolicy)
-	if err != nil {
-		if apierrors.IsNotFound(err) {
-			klog.V(4).Infof("ClusterPropagationPolicy(%s) has been removed.", policyName)
-			return d.HandleClusterPropagationPolicyDeletion(policyID)
+	if apierrors.IsNotFound(err) {
+		// cluster Policy not found or being deleted, clean up metadata
```
RainbowMango: ditto

pkg/detector/detector.go
```diff
@@ -1069,66 +1120,63 @@ func (d *ResourceDetector) HandlePropagationPolicyDeletion(policyID string) erro
...
-			}
+func (d *ResourceDetector) handleResourceTemplateAndBindingCleanup(template *unstructured.Unstructured, objRef workv1alpha2.ObjectReference, targetClaimMetadata map[string]string, cleanupFunc func(obj metav1.Object)) error {
+	bindingName := names.GenerateBindingName(template.GetKind(), template.GetName())
+	if template.GetNamespace() != "" {
+		// Clean up the claim metadata from the reference binding so that the karmada scheduler won't reschedule the binding.
+		// Must remove the claim metadata, such as labels and annotations, from the ResourceBinding ahead of resource template,
+		// otherwise might lose the chance to do that in a retry loop (in particular, the claim metadata was successfully removed
+		// from resource template, but ResourceBinding not), since the resource template will not be listed again.
+		if err := d.CleanupResourceBindingClaimMetadata(template.GetNamespace(), bindingName, targetClaimMetadata, cleanupFunc); err != nil {
+			klog.Errorf("Failed to clean up claim metadata from resource binding(%s/%s) when propagationPolicy removed, error: %v",
```
RainbowMango: The log is inaccurate; no such context indicates it is called due to the PropagationPolicy deletion.
```suggestion
			klog.Errorf("Failed to clean up claim metadata from ResourceBinding(%s/%s), error: %v",
```

Same as below.

pkg/detector/detector.go
```diff
@@ -1069,66 +1121,64 @@ func (d *ResourceDetector) HandlePropagationPolicyDeletion(policyID string) erro
...
 
-			// Clean up the claim metadata from the reference binding so that the Karmada scheduler won't reschedule the binding.
-			if err := d.CleanupClusterResourceBindingClaimMetadata(&crbs.Items[index], labelSet); err != nil {
-				klog.Errorf("Failed to clean up claim metadata from clusterResourceBinding(%s) when clusterPropagationPolicy removed, error: %v",
-					binding.Name, err)
+		rawObjects, err := helper.FetchResourceTemplatesByLabelSelector(d.DynamicClient, d.InformerManager, d.RESTMapper, objRef, labels.SelectorFromSet(claimMetadata))
+		if err != nil {
+			errs = append(errs, err)
+			continue
+		}
```
RainbowMango, line 1139: It is the place where we introduced issue #6913.

Issue #6582, opened 2025-07-29
[Umbrella] follow-up action plan based on 6434


XiShanYongYe-Chang:
Based on the commet https://github.com/karmada-io/karmada/pull/6434#pullrequestreview-3057857773, we record task items:

- [ ] Add necessary unit tests or integrate tests for this issue, to ensure #6434 not gonna happy again
- [ ] Clarify the reason why it is necessary to list-watch the ServiceExport object
- [ ] Send a notice email to the Karmada Adopter Group to acknowledge the end users.
- [ ] Backport this to release branches. release-1.14, release-1.13, release-1.12.

RainbowMango:
> Send a notice email to the Karmada Adopter Group to acknowledge the end users.

Done.

PR #6664, opened 2025-08-19
add comments for service-export-controller

XiShanYongYe-Chang:
**What type of PR is this?**

/kind cleanup

Clarify the reason why it is necessary to list-watch the ServiceExport object.

Part of #6582

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/mcs/service_export_controller.go
+++ b/pkg/controllers/mcs/service_export_controller.go
@@ -64,6 +64,9 @@ import (
 const ServiceExportControllerName = "service-export-controller"
 
 // ServiceExportController is to sync ServiceExport and report EndpointSlices of exported service to control-plane.
+// It creates Informers for clusters where ServiceExport resources have been propagated,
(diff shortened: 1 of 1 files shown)


pkg/controllers/mcs/service_export_controller.go
```diff
@@ -64,6 +64,9 @@ import (
 const ServiceExportControllerName = "service-export-controller"
 
 // ServiceExportController is to sync ServiceExport and report EndpointSlices of exported service to control-plane.
+// In order to avoid creating informer for each member cluster, the controller will watch Work resources containing
```
RainbowMango: misleading. The controller needs to create informer for member clusters, but only for clusters where have the ServiceExport resource propagated.

XiShanYongYe-Chang: Updated, can you help review it again?

RainbowMango: OK.

PR #6322, opened 2025-04-25
fix(detector): fix pp and rb inconsistent

CharlesQQ:
**What type of PR is this?**

**What this PR does / why we need it**:

When continuously updating the ·spec.suspension.dispatching· field of pp, the field of rb may be inconsistent with pp.

**Which issue(s) this PR fixes**:
fixes https://github.com/karmada-io/karmada/issues/6379

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

Issue #6660, opened 2025-08-18
[Suspension] .spec.suspension of ResourceBinding not removed after .spec.suspension of PropagationPolicy has been removed


ryanwuer:
**What happened**:
.spec.suspension of ResourceBinding not removed after .spec.suspension of PropagationPolicy has been removed

**What you expected to happen**:
.spec.suspension of ResourceBinding will be removed after .spec.suspension of PropagationPolicy has been removed

**How to reproduce it (as minimally and precisely as possible)**:
1. Add suspension into PropagationPolicy
2. Suspension is added automatically into ResourceBinding
3. Remove suspension of PropagationPolicy
4. Suspension of ResourceBinding  cannot be removed automatically, which makes Suspension still take effect.
5. After removed Suspension of ResourceBinding manually, the Suspension effect is gone.

**Anything else we need to know?**:

**Environment**: 
- Karmada version: v1.14.0
- kubectl-karmada or karmadactl version (the result of `kubectl-karmada version` or `karmadactl version`):
- Others:

XiShanYongYe-Chang:
~Hi @ryanwuer the current behavior is as expected; even if the PropagationPolicy is deleted, the ResourceBinding will still remain~, and its configuration will be consistent with that before the PropagationPolicy was deleted (unless another PropagationPolicy with the same name as the current resource exists, thereby overriding the configuration of the ResourceBinding).

Could you help explain why you need to delete the PropagationPolicy and the use cases where you do not set suspension to true before deleting it?

ryanwuer:
@XiShanYongYe-Chang I'm saying the behavior of deletion of 'Suspension', not the PropagationPolicy itself.

Here is the reference doc: https://karmada.io/zh/docs/userguide/scheduling/resource-propagating/#%E6%9A%82%E5%81%9C%E5%90%91%E7%89%B9%E5%AE%9A%E9%9B%86%E7%BE%A4%E5%88%86%E5%8F%91%E8%B5%84%E6%BA%90

XiShanYongYe-Chang:
~Sorry, I misread it. This issue may be consistent with the problem reported by #5999.~

This issue description addresses the same problem: https://github.com/karmada-io/karmada/issues/5996

ryanwuer:
[...]

PR #6674, opened 2025-08-21
fix the issue that the relevant fields in rb and pp are inconsistent

### LINKED ISSUE #5996: The resource spec contents of pp and rb are not synchronized in time.
**What happened**:
1. During controller restart, the .spec.suspension fields of rb and pp are different

https://www.processon.com/v/6772523638625f07c0ec6273
karmada 灰度发布流程

During the restart, the resource detector controller processes the currently released workload later than the binding controller, causing the pause setting to become invalid; our custom controller extension (for grayscale release) does not update the op resource, and the number of new pods more than partition.

2. After  pp sets .spec.preserveResourcesOnDeletion=true, execute the deletion operation immediately, and the resources of the member clusters may be deleted in cascade

3. When updating the Dispatching field of pp to false, and then updating deployment, the Dispatching field of rb is first updated to false, and then updated to true again.

Image

Image
Add the following log in ApplyPolicy, and the cached data will be correct and meet expectations, but when rb is updated at the end, the order will be disrupted

```
	klog.Infof("Applying policy(%s/%s) for object: %s", policy.Namespace, policy.Name, objectKey)
	if policy.Spec.Suspension != nil && policy.Spec.Suspension.Dispatching != nil {
		klog.Infof("Applying policy(%s/%s) with dispatching suspension: %v", policy.Namespace, policy.Name, *policy.Spec.Suspension.Dispatching)
	} else {
		klog.Infof("Applying policy(%s/%s) without dispatching suspension", policy.Namespace, policy.Name)
	}
```

```
	switch operationResult {
	case controllerutil.OperationResultCreated:
		klog.Infof("Create ResourceBinding(%s/%s) successfully.", binding.GetNamespace(), binding.GetName())
	case controllerutil.OperationResultUpdated:
		if bindingCopy.Spec.Suspension != nil && bindingCopy.Spec.Suspension.Dispatching != nil {
			klog.Infof("Update ResourceBinding(%s/%s) successfully. dispatching: %v", binding.GetNamespace(), binding.GetName(), *bindingCopy.Spec.Suspension.Dispatching)
		} else {
			klog.Infof("Update ResourceBinding(%s/%s) successfully.", binding.GetNamespace(), binding.GetName())
		}
	default:
		klog.V(2).Infof("ResourceBinding(%s/%s) is up to date.", binding.GetNamespace(), binding.GetName())
	}
```
Image

Image
Add the following log after the CreateOrUpdate rb function to show that the original execution order changes after this function.

```
	if policy.Spec.Suspension != nil && policy.Spec.Suspension.Dispatching != nil {
		klog.Infof("CreateOrUpdate (%s/%s) with dispatching suspension: %v, cost: %v", policy.Namespace, policy.Name, *policy.Spec.Suspension.Dispatching, time.Since(start3))
	} else {
		klog.Infof("CreateOrUpdate (%s/%s) without dispatching suspension", policy.Namespace, policy.Name)
	}
```

Image

Add log in function `CreateOrUpdate`, Indicates that an optimistic lock conflict occurs when updating the rb object, and a retry occurs, causing the update to be delayed.
```
	startTime := time.Now()
	if NewErr := c.Update(ctx, obj); NewErr != nil {
		klog.Infof("Failed to update (%s/%s) used time: %v, err: %v", obj.GetNamespace(), obj.GetName(), time.Since(startTime), NewErr)
		return controllerutil.OperationResultNone, NewErr
	}
```

Image

Q:  Why dispatching=true is triggered 3 times?

Deployment is released, it will be updated once, remove label `propagationpolicy.karmada.io/permanent-id` and Annonation 
`propagationpolicy.karmada.io/namespace`, `propagationpolicy.karmada.io/name`, after finding pp through `LookForMatchedPolicy` function, it will pass through `ClaimPolicyForObject` function, this will trigger twice, pp will be updated once more, a total of 3 times

Q: how the workers consume the worker queue?
https://github.com/karmada-io/karmada/blob/c5b2d3c65a7357f37ab6fc7d331b4ad53e887e43/pkg/util/fedinformer/keys/keys.go#L114-L121

Image

set struct `ClusterWideKeyWithConfig`  field `ResourceChangeByKarmada=false`
Image

Add log for Reconcile funciton:
```
rand.Seed(time.Now().UnixNano())
	randStr := randomString(20)
	startTime := time.Now()
	defer func() {
		klog.Infof("Finished Reconcile, (%s/%s), cost: %v, randStr: %s", clusterWideKeyWithConfig.ClusterWideKey.Namespace, clusterWideKeyWithConfig.ClusterWideKey.Name, time.Since(startTime), randStr)
	}()

	clusterWideKey := clusterWideKeyWithConfig.ClusterWideKey
	resourceChangeByKarmada := clusterWideKeyWithConfig.ResourceChangeByKarmada
	klog.Infof("Reconciling object: %s, resourceChangeByKarmada: %v, randStr: %s", clusterWideKey, resourceChangeByKarmada, randStr)
```

Image

Image

Because the Resource Change By Karmada field has changed, different workers are triggered to execute the Reconcile function at the same time.

The relevant fields in pp, rb, and work can be consistent. If the fields in rb and work are inconsistent with pp, then the subsequent operations related to rb and work do not need to be performed until the synchronization of the relevant fields is completed

1. have lost of resouce template And pp,  restart controller and set  .spec.suspension.dispatching=true, check rb field .spec.suspension.dispatching.
2. set pp field .spec.preserveResourcesOnDeletion=true, and deletion operation immediately, check whether the resource has been cascade deleted  in member cluster.

**Environment**:
- Karmada version:
v1.12
- Others:

### LINKED ISSUE #6379: The fields related to pp and rb may be different
**What happened**:
Move issue https://github.com/karmada-io/karmada/issues/5996  case 3 to this one, to better track this issue

When updating the pp object field `.spec.suspension.dispatching` to false, and then updating deployment, the  rb field `.spec.suspension.dispatching` updated to false, and then updated to true again.

The relevant fields of rb need to be the same as pp

1. update pp field `spec.suspension.dispatching` to true
2. update pp field `spec.suspension.dispatching` to false
3. update resourcetemplate
retry above steps some times,  may find field `spec.suspension.dispatching` is not same between pp with rb

Image

Image

Add the following log in ApplyPolicy, and the cached data will be correct and meet expectations, but when rb is updated at the end, the order will be disrupted

```
	klog.Infof("Applying policy(%s/%s) for object: %s", policy.Namespace, policy.Name, objectKey)
	if policy.Spec.Suspension != nil && policy.Spec.Suspension.Dispatching != nil {
		klog.Infof("Applying policy(%s/%s) with dispatching suspension: %v", policy.Namespace, policy.Name, *policy.Spec.Suspension.Dispatching)
	} else {
		klog.Infof("Applying policy(%s/%s) without dispatching suspension", policy.Namespace, policy.Name)
	}
```

```
	switch operationResult {
	case controllerutil.OperationResultCreated:
		klog.Infof("Create ResourceBinding(%s/%s) successfully.", binding.GetNamespace(), binding.GetName())
	case controllerutil.OperationResultUpdated:
		if bindingCopy.Spec.Suspension != nil && bindingCopy.Spec.Suspension.Dispatching != nil {
			klog.Infof("Update ResourceBinding(%s/%s) successfully. dispatching: %v", binding.GetNamespace(), binding.GetName(), *bindingCopy.Spec.Suspension.Dispatching)
		} else {
			klog.Infof("Update ResourceBinding(%s/%s) successfully.", binding.GetNamespace(), binding.GetName())
		}
	default:
		klog.V(2).Infof("ResourceBinding(%s/%s) is up to date.", binding.GetNamespace(), binding.GetName())
	}
```
Image

Image
Add the following log after the CreateOrUpdate rb function to show that the original execution order changes after this function.

```
	if policy.Spec.Suspension != nil && policy.Spec.Suspension.Dispatching != nil {
		klog.Infof("CreateOrUpdate (%s/%s) with dispatching suspension: %v, cost: %v", policy.Namespace, policy.Name, *policy.Spec.Suspension.Dispatching, time.Since(start3))
	} else {
		klog.Infof("CreateOrUpdate (%s/%s) without dispatching suspension", policy.Namespace, policy.Name)
	}
```

Image

Add log in function `CreateOrUpdate`, Indicates that an optimistic lock conflict occurs when updating the rb object, and a retry occurs, causing the update to be delayed.
```
	startTime := time.Now()
	if NewErr := c.Update(ctx, obj); NewErr != nil {
		klog.Infof("Failed to update (%s/%s) used time: %v, err: %v", obj.GetNamespace(), obj.GetName(), time.Since(startTime), NewErr)
		return controllerutil.OperationResultNone, NewErr
	}
```

Image

Q:  Why dispatching=true is triggered 3 times?

Deployment is released, it will be updated once, remove label `propagationpolicy.karmada.io/permanent-id` and Annonation 
`propagationpolicy.karmada.io/namespace`, `propagationpolicy.karmada.io/name`, after finding pp through `LookForMatchedPolicy` function, it will pass through `ClaimPolicyForObject` function, this will trigger twice, pp will be updated once more, a total of 3 times

Q: how the workers consume the worker queue?
https://github.com/karmada-io/karmada/blob/c5b2d3c65a7357f37ab6fc7d331b4ad53e887e43/pkg/util/fedinformer/keys/keys.go#L114-L121

Image

set struct `ClusterWideKeyWithConfig`  field `ResourceChangeByKarmada=false`
Image

Add log for Reconcile funciton:
```
rand.Seed(time.Now().UnixNano())
	randStr := randomString(20)
	startTime := time.Now()
	defer func() {
		klog.Infof("Finished Reconcile, (%s/%s), cost: %v, randStr: %s", clusterWideKeyWithConfig.ClusterWideKey.Namespace, clusterWideKeyWithConfig.ClusterWideKey.Name, time.Since(startTime), randStr)
	}()

	clusterWideKey := clusterWideKeyWithConfig.ClusterWideKey
	resourceChangeByKarmada := clusterWideKeyWithConfig.ResourceChangeByKarmada
	klog.Infof("Reconciling object: %s, resourceChangeByKarmada: %v, randStr: %s", clusterWideKey, resourceChangeByKarmada, randStr)
```

Image

Image

The root cause is `ResourceChangeByKarmada` field has changed, different workers are triggered to execute the Reconcile function at the same time.

**Environment**:
- Karmada version:
- Others:

zhzhuang-zju:
**What type of PR is this?**
/kind bug

In lazy mode, the processor needs to differentiate if the enqueue is caused by Karmada to take different handling paths. For non-lazy policies, this distinction is unnecessary.
Note: toggling ResourceChangeByKarmada changes the effective queue key. Mixing both true/false for the same resource may lead to two different QueueKeys enqueued and processed concurrently, causing race conditions. Therefore, set ResourceChangeByKarmada only in lazy mode.

Fixes #5996
Fixes #6379 
Fixes #6322

```release-note
`karmada-controller-manager`: Fixed the issue that the relevant fields in rb and pp are inconsistent.
```

### DIFF
--- a/pkg/detector/detector.go
+++ b/pkg/detector/detector.go
@@ -343,13 +343,26 @@ func (d *ResourceDetector) OnUpdate(oldObj, newObj interface{}) {
 		return
 	}
 
-	resourceChangeByKarmada := eventfilter.ResourceChangeByKarmada(unstructuredOldObj, unstructuredNewObj)
-
-	resourceItem := ResourceItem{
-		Obj:                     newRuntimeObj,
-		ResourceChangeByKarmada: resourceChangeByKarmada,
+	isLazyActivation, err := d.isClaimedByLazyPolicy(unstructuredNewObj)
+	if err != nil {
+		// should never come here
+		klog.Errorf("Failed to check if the object (kind=%s, %s/%s) is bound by lazy policy. err: %v", unstructuredNewObj.GetKind(), unstructuredNewObj.GetNamespace(), unstructuredNewObj.GetName(), err)
 	}
 
+	if isLazyActivation {
+		resourceItem := ResourceItem{
+			Obj:                     newRuntimeObj,
+			ResourceChangeByKarmada: eventfilter.ResourceChangeByKarmada(unstructuredOldObj, unstructuredNewObj),
+		}
+
+		d.Processor.Enqueue(resourceItem)
+		return
+	}
+
+	// For non-lazy policies, it is no need to distinguish whether the change is from Karmada or not.
+	resourceItem := ResourceItem{
+		Obj: newRuntimeObj,
+	}
 	d.Processor.Enqueue(resourceItem)
 }
 
@@ -1158,7 +1171,7 @@ func (d *ResourceDetector) HandlePropagationPolicyCreationOrUpdate(policy *polic
 		if err != nil {
 			return err
 		}
-		d.Processor.Add(keys.ClusterWideKeyWithConfig{ClusterWideKey: resourceKey, ResourceChangeByKarmada: true})
+		d.enqueueResourceTemplateForPolicyChange(resourceKey, policy.Spec.ActivationPreference)
 	}
 
 	// check whether there are matched RT in waiting list, is so, add it to processor
@@ -1176,7 +1189,7 @@ func (d *ResourceDetector) HandlePropagationPolicyCreationOrUpdate(policy *polic
 
 	for _, key := range matchedKeys {
 		d.RemoveWaiting(key)
-		d.Processor.Add(keys.ClusterWideKeyWithConfig{ClusterWideKey: key, ResourceChangeByKarmada: true})
+		d.enqueueResourceTemplateForPolicyChange(key, policy.Spec.ActivationPreference)
 	}
 
 	// If preemption is enabled, handle the preemption process.
@@ -1225,14 +1238,14 @@ func (d *ResourceDetector) HandleClusterPropagationPolicyCreationOrUpdate(policy
 		if err != nil {
 			return err
 		}
-		d.Processor.Add(keys.ClusterWideKeyWithConfig{ClusterWideKey: resourceKey, ResourceChangeByKarmada: true})
+		d.enqueueResourceTemplateForPolicyChange(resourceKey, policy.Spec.ActivationPreference)
 	}
 	for _, crb := range clusterResourceBindings.Items {
 		resourceKey, err := helper.ConstructClusterWideKey(crb.Spec.Resource)
 		if err != nil {
 			return err
 		}
-		d.Processor.Add(keys.ClusterWideKeyWithConfig{ClusterWideKey: resourceKey, ResourceChangeByKarmada: true})
+		d.enqueueResourceTemplateForPolicyChange(resourceKey, policy.Spec.ActivationPreference)
 	}
 
 	matchedKeys := d.GetMatching(policy.Spec.ResourceSelectors)
@@ -1249,7 +1262,7 @@ func (d *ResourceDetector) HandleClusterPropagationPolicyCreationOrUpdate(policy
 
 	for _, key := range matchedKeys {
 		d.RemoveWaiting(key)
-		d.Processor.Add(keys.ClusterWideKeyWithConfig{ClusterWideKey: key, ResourceChangeByKarmada: true})
+		d.enqueueResourceTemplateForPolicyChange(key, policy.Spec.ActivationPreference)
 	}
 
 	// If preemption is enabled, handle the preemption process.
@@ -1372,4 +1385,22 @@ func (d *ResourceDetector) applyReplicaInterpretation(object *unstructured.Unstr
 	}
 
 	return nil
+}
+
+// enqueueResourceTemplateForPolicyChange enqueues a resource template key for reconciliation in response to a
+// PropagationPolicy or ClusterPropagationPolicy change. If the policy's ActivationPreference is set to Lazy,
+// the ResourceChangeByKarmada flag is set to true, indicating that the resource template is being enqueued
+// due to a policy change and should not be propagated to member clusters. For non-lazy policies, this flag
+// is omitted as the distinction is unnecessary.
+//
+// Note: Setting ResourceChangeByKarmada changes the effective queue key. Mixing both true/false for the same
+// resource may result in two different queue keys being processed concurrently, which can cause race conditions.
+// Therefore, only set ResourceChangeByKarmada in lazy activation mode.
+// For more details, see: https://github.com/karmada-io/karmada/issues/5996.
+func (d *ResourceDetector) enqueueResourceTemplateForPolicyChange(key keys.ClusterWideKey, pref policyv1alpha1.ActivationPreference) {
+	if util.IsLazyActivationEnabled(pref) {
+		d.Processor.Add(keys.ClusterWideKeyWithConfig{ClusterWideKey: key, ResourceChangeByKarmada: true})
+		return
+	}
+	d.Processor.Add(keys.ClusterWideKeyWithConfig{ClusterWideKey: key})
 }
--- a/pkg/detector/detector_test.go
+++ b/pkg/detector/detector_test.go
@@ -429,7 +429,6 @@ func TestOnUpdate(t *testing.T) {
 		oldObj                    interface{}
 		newObj                    interface{}
 		expectedEnqueue           bool
-		expectedChangeByKarmada   bool
 		expectToUnstructuredError bool
 	}{
 		{
@@ -460,8 +459,7 @@ func TestOnUpdate(t *testing.T) {
 					},
 				},
 			},
-			expectedEnqueue:         true,
-			expectedChangeByKarmada: false,
+			expectedEnqueue: true,
 		},
 		{
 			name: "update without changes",
@@ -524,8 +522,7 @@ func TestOnUpdate(t *testing.T) {
 					},
 				},
 			},
-			expectedEnqueue:         true,
-			expectedChangeByKarmada: true,
+			expectedEnqueue: true,
 		},
 		{
 			name: "core v1 object",
@@ -573,7 +570,6 @@ func TestOnUpdate(t *testing.T) {
 				assert.IsType(t, ResourceItem{}, mockProcessor.lastEnqueued, "Enqueued item should be of type ResourceItem")
 				enqueued := mockProcessor.lastEnqueued.(ResourceItem)
 				assert.Equal(t, tt.newObj, enqueued.Obj, "Enqueued object should match the new object")
-				assert.Equal(t, tt.expectedChangeByKarmada, enqueued.ResourceChangeByKarmada, "ResourceChangeByKarmada flag should match expected value")
 			} else {
 				assert.Equal(t, 0, mockProcessor.enqueueCount, "Object should not be enqueued")
 			}
@@ -963,6 +959,71 @@ func TestApplyClusterPolicy(t *testing.T) {
 					assert.Equal(t, tt.object.GetName(), binding.Spec.Resource.Name)
 				}
 			}
+		})
+	}
+}
+
+func TestEnqueueResourceKeyWithActivationPref(t *testing.T) {
+	testClusterWideKey := keys.ClusterWideKey{
+		Group:     "foo",
+		Version:   "foo",
+		Kind:      "foo",
+		Namespace: "foo",
+		Name:      "foo",
+	}
+	tests := []struct {
+		name string
+		key  keys.ClusterWideKey
+		pref policyv1alpha1.ActivationPreference
+		want keys.ClusterWideKeyWithConfig
+	}{
+		{
+			name: "lazy pp and resourceChangeByKarmada is true",
+			key:  testClusterWideKey,
+			pref: policyv1alpha1.LazyActivation,
+			want: keys.ClusterWideKeyWithConfig{
+				ClusterWideKey:          testClusterWideKey,
+				ResourceChangeByKarmada: true,
+			},
+		},
+		{
+			name: "non-lazy ignores ResourceChangeByKarmada",
+			key:  testClusterWideKey,
+			pref: "",
+			want: keys.ClusterWideKeyWithConfig{
+				ClusterWideKey:          testClusterWideKey,
+				ResourceChangeByKarmada: false,
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ctx, cancel := context.WithCancel(context.Background())
+			detector := ResourceDetector{
+				Processor: util.NewAsyncWorker(util.Options{
+					Name:    "resource detector",
+					KeyFunc: ResourceItemKeyFunc,
+					ReconcileFunc: func(key util.QueueKey) (err error) {
+						defer cancel()
+						defer func() {
+							assert.NoError(t, err)
+						}()
+						clusterWideKeyWithConfig, ok := key.(keys.ClusterWideKeyWithConfig)
+						if !ok {
+							err = fmt.Errorf("invalid key")
+							return err
+						}
+						if clusterWideKeyWithConfig != tt.want {
+							err = fmt.Errorf("unexpected key. want:%+v, got:%+v", tt.want, clusterWideKeyWithConfig)
+							return err
+						}
+						return nil
+					},
+				}),
+			}
+			detector.Processor.Run(ctx, 1)
+			detector.enqueueResourceTemplateForPolicyChange(tt.key, tt.pref)
+			<-ctx.Done()
 		})
 	}
 }
--- a/pkg/detector/policy.go
+++ b/pkg/detector/policy.go
@@ -332,6 +332,45 @@ func (d *ResourceDetector) listCPPDerivedCRBs(policyID, policyName string) (*wor
 	return bindings, nil
 }
 
+func (d *ResourceDetector) isClaimedByLazyPolicy(obj *unstructured.Unstructured) (bool, error) {
+	policyAnnotations := obj.GetAnnotations()
+	policyLabels := obj.GetLabels()
+	policyNamespace := util.GetAnnotationValue(policyAnnotations, policyv1alpha1.PropagationPolicyNamespaceAnnotation)
+	policyName := util.GetAnnotationValue(policyAnnotations, policyv1alpha1.PropagationPolicyNameAnnotation)
+	claimedID := util.GetLabelValue(policyLabels, policyv1alpha1.PropagationPolicyPermanentIDLabel)
+	if policyNamespace != "" && policyName != "" && claimedID != "" {
+		matchedPropagationPolicy := &policyv1alpha1.PropagationPolicy{}
+		err := d.Client.Get(context.TODO(), client.ObjectKey{Namespace: policyNamespace, Name: policyName}, matchedPropagationPolicy)
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				return false, nil
+			}
+
+			return false, err
+		}
+
+		return util.IsLazyActivationEnabled(matchedPropagationPolicy.Spec.ActivationPreference), nil
+	}
+
+	policyName = util.GetAnnotationValue(policyAnnotations, policyv1alpha1.ClusterPropagationPolicyAnnotation)
+	claimedID = util.GetLabelValue(policyLabels, policyv1alpha1.ClusterPropagationPolicyPermanentIDLabel)
+	if policyName != "" && claimedID != "" {
+		matchedClusterPropagationPolicy := &policyv1alpha1.ClusterPropagationPolicy{}
+		err := d.Client.Get(context.TODO(), client.ObjectKey{Name: policyName}, matchedClusterPropagationPolicy)
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				return false, nil
+			}
+
+			return false, err
+		}
+
+		return util.IsLazyActivationEnabled(matchedClusterPropagationPolicy.Spec.ActivationPreference), nil
+	}
+
+	return false, nil
+}
+
 // excludeClusterPolicy excludes cluster propagation policy.
 // If propagation policy was claimed, cluster propagation policy should not exist.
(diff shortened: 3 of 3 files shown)


pkg/detector/detector.go
```diff
@@ -1355,3 +1355,19 @@ func (d *ResourceDetector) CleanupClusterResourceBindingClaimMetadata(crb *workv
...
+// Note: toggling ResourceChangeByKarmada changes the effective queue key. Mixing both true/false
+// for the same resource may lead to two different QueueKeys enqueued and processed concurrently,
+// causing race conditions. Therefore, set ResourceChangeByKarmada only in lazy mode.
+// More info can refer to https://github.com/karmada-io/karmada/issues/5996.
+func (d *ResourceDetector) enqueueResourceKeyWithActivationPref(key keys.ClusterWideKey, pref policyv1alpha1.ActivationPreference, resourceChangeByKarmada bool) {
+	if util.IsLazyActivationEnabled(pref) {
+		d.Processor.Add(keys.ClusterWideKeyWithConfig{ClusterWideKey: key, ResourceChangeByKarmada: resourceChangeByKarmada})
+		return
+	}
+	d.Processor.Add(keys.ClusterWideKeyWithConfig{ClusterWideKey: key})
```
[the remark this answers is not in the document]
zhzhuang-zju: Not very acceptable. The current implementation is already straightforward, and the contrast effect is also obvious.

RainbowMango (commented):
/assign 
/gemini review

pkg/detector/detector.go
```diff
@@ -1373,3 +1373,19 @@ func (d *ResourceDetector) applyReplicaInterpretation(object *unstructured.Unstr
...
+
+// enqueueResourceKeyWithActivationPref enqueues the given key with optional ResourceChangeByKarmada flag
+// and only honors the flag when ActivationPreference is lazy.
+// In lazy mode, the processor needs to differentiate if the enqueue is caused by Karmada
+// to take different handling paths. For non-lazy policies, this distinction is unnecessary.
+// Note: toggling ResourceChangeByKarmada changes the effective queue key. Mixing both true/false
+// for the same resource may lead to two different QueueKeys enqueued and processed concurrently,
+// causing race conditions. Therefore, set ResourceChangeByKarmada only in lazy mode.
+// More info can refer to https://github.com/karmada-io/karmada/issues/5996.
+func (d *ResourceDetector) enqueueResourceKeyWithActivationPref(key keys.ClusterWideKey, pref policyv1alpha1.ActivationPreference, resourceChangeByKarmada bool) {
```
RainbowMango: I see all of the 5 places set the `ResourceChangeByKarmada` with value `true`? 
I wonder if there a case that expects `false`? If not, why need that parmater?

zhzhuang-zju: This is to distinguish whether the reconciliation of a resource is triggered by its own modification. For example, if a user modifies the deployment and triggers the reconcile, `ResourceChangeByKarmada` will be false.

https://github.com/karmada-io/karmada/blob/708f3e453adea1172ea4c11930b0fc64f56ba1a8/pkg/detector/detector.go#L313-L359

I noticed a missing part where `ResourceChangeByKarmada` is also set in `onUpdate`. It needs to be taken care of.

RainbowMango: How about 
```golang
// enqueueResourceTemplateForPolicyChange enqueues a resource template key for reconciliation in response to a
// PropagationPolicy or ClusterPropagationPolicy change. If the policy's ActivationPreference is set to Lazy,
// the ResourceChangeByKarmada flag is set to true, indicating that the resource template is being enqueued
// due to a policy change and should not be propagated to member clusters. For non-lazy policies, this flag
// is omitted as the distinction is unnecessary.
//
// Note: Setting ResourceChangeByKarmada changes the effective queue key. Mixing both true/false for the same
// resource may result in two different queue keys being processed concurrently, which can cause race conditions.
// Therefore, only set ResourceChangeByKarmada in lazy activation mode.
// For more details, see: https://github.com/karmada-io/karmada/issues/5996.
func (d *ResourceDetector) enqueueResourceTemplateForPolicyChange(key keys.ClusterWideKey, activationPref policyv1alpha1.ActivationPreference) {
	if util.IsLazyActivationEnabled(activationPref) {
		d.Processor.Add(keys.ClusterWideKeyWithConfig{ClusterWideKey: key, ResourceChangeByKarmada: true})
		return
	}
	d.Processor.Add(keys.ClusterWideKeyWithConfig{ClusterWideKey: key})
}
```

pkg/detector/detector.go
```diff
@@ -1373,3 +1400,87 @@ func (d *ResourceDetector) applyReplicaInterpretation(object *unstructured.Unstr
...
+// PropagationPolicy or ClusterPropagationPolicy change. If the policy's ActivationPreference is set to Lazy,
+// the ResourceChangeByKarmada flag is set to true, indicating that the resource template is being enqueued
+// due to a policy change and should not be propagated to member clusters. For non-lazy policies, this flag
+// is omitted as the distinction is unnecessary.
+//
+// Note: Setting ResourceChangeByKarmada changes the effective queue key. Mixing both true/false for the same
+// resource may result in two different queue keys being processed concurrently, which can cause race conditions.
+// Therefore, only set ResourceChangeByKarmada in lazy activation mode.
+// For more details, see: https://github.com/karmada-io/karmada/issues/5996.
+func (d *ResourceDetector) enqueueResourceKeyWithActivationPref(key keys.ClusterWideKey, pref policyv1alpha1.ActivationPreference) {
```
RainbowMango: I still prefer the name `enqueueResourceTemplateForPolicyChange` as I proposed at https://github.com/karmada-io/karmada/pull/6674#discussion_r2292533325, which clearly tells/hints when this method would be invoked.

zhzhuang-zju:
> Creating cluster "member1" ...
 • Ensuring node image (kindest/node:v1.33.0) 🖼  ...

Stuck at image kindest/node:v1.33.0 pulling stage, causing karmada installation failure.

/retest

zhzhuang-zju:
same error as https://github.com/karmada-io/karmada/pull/6674#issuecomment-3222494837
/retest

pkg/detector/detector.go
```diff
@@ -343,13 +343,26 @@ func (d *ResourceDetector) OnUpdate(oldObj, newObj interface{}) {
 		return
 	}
 
-	resourceChangeByKarmada := eventfilter.ResourceChangeByKarmada(unstructuredOldObj, unstructuredNewObj)
+	isLazyActivation, err := d.isClaimedByLazyPolicy(unstructuredNewObj)
+	if err != nil {
+		// should never come here
+		klog.Errorf("Failed to check if the object (kind=%s, %s/%s) is bound by lazy policy. err: %v", unstructuredNewObj.GetKind(), unstructuredNewObj.GetNamespace(), unstructuredNewObj.GetName(), err)
+	}
```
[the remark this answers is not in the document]
zhzhuang-zju: @gemini-code-assist Thank you for pointing out this issue, but would it not cause event loss if we simply return? Also, isLazyActivation is obtained through caching get and should not result in network issues.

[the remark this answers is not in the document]
zhzhuang-zju: Regarding the Event Loss section, I'm very interested. I would like to confirm whether directly using 'return' would cause event loss. Thank you for the reminder regardless.

zhzhuang-zju:
> @zhzhuang-zju Do you still want to make some changes in the current state?

Yes, please wait a moment

zhzhuang-zju:
Ref to https://github.com/karmada-io/karmada/pull/6674#discussion_r2300363410, handling errors in `Onupdate` can be challenging. So I implemented another solution that uses a self-maintained cache to determine if it's a lazy policy.

Here is the link to the alternative options: https://github.com/karmada-io/karmada/compare/master...zhzhuang-zju:changebykarmada-new?expand=1 
@RainbowMango please review this and see if it is feasible, thanks

ryanwuer:
> /lgtm
> 
> @zhzhuang-zju Do you still want to make some changes in the current state?
> 
> @CharlesQQ @ryanwuer Can you help run a quick test for this?

How can i make the test, checkout the branch and build an image?

zhzhuang-zju:
> How can i make the test, checkout the branch and build an image?

@ryanwuer Yes, you're right

ryanwuer:
> > How can i make the test, checkout the branch and build an image?
> 
> 
> 
> @ryanwuer Yes, you're right

I am using v1.14.0, can i (or i need/must) bump all components to the temp version that i build locally?

zhzhuang-zju:
> I am using v1.14.0, can i (or i need/must) bump all components to the temp version that i build locally?

Is it in the validation environment? If you're only verifying this issue, upgrading the controller-manager alone should be sufficient.

CharlesQQ:
Can we add an error log here to facilitate troubleshooting and verification when the `CreateOrUpdate`  operation returns err that is not nil?  @zhzhuang-zju 

https://github.com/karmada-io/karmada/blob/78ffbd528015c277a0938b358813c4b9c3334b99/pkg/detector/detector.go#L584-L585

zhzhuang-zju:
> Can we add an error log here to facilitate troubleshooting and verification when the `CreateOrUpdate` operation returns err that is not nil? @zhzhuang-zju
> 
> https://github.com/karmada-io/karmada/blob/78ffbd528015c277a0938b358813c4b9c3334b99/pkg/detector/detector.go#L584-L585

https://github.com/karmada-io/karmada/blob/78ffbd528015c277a0938b358813c4b9c3334b99/pkg/detector/detector.go#L588-L590

Is this log sufficient?

ryanwuer:
> > I am using v1.14.0, can i (or i need/must) bump all components to the temp version that i build locally?
> 
> 
> 
> Is it in the validation environment? If you're only verifying this issue, upgrading the controller-manager alone should be sufficient.
> 
> 

Yes, the test environment, by the way, if this is merged, new release is published, how do i make the upgrade in the production environment, upgrade all components or the controller manager alone?

CharlesQQ:
> Is this log sufficient?

This is not enough. This log can only limit the final update results, and cannot show each failure to createOrUpdate. In this case https://github.com/karmada-io/karmada/issues/6379, it is necessary to know the failures that led to retries.

zhzhuang-zju:
> Yes, the test environment, by the way, if this is merged, new release is published, how do i make the upgrade in the production environment, upgrade all components or the controller manager alone?

https://karmada.io/docs/administrator/upgrading/
We recommend upgrading together to avoid potential unavailability of features.

zhzhuang-zju:
> This is not enough. This log can only limit the final update results, and cannot show each failure to createOrUpdate. In this case #6379, it is necessary to know the failures that led to retries.

If a resource can only be processed by one worker at a time, would retrying after a failure during the process also be fine?

CharlesQQ:
> If a resource can only be processed by one worker at a time, would retrying after a failure during the process also be fine?

Yes, but the problem at that time was that a resource was processed by more than one worker. I want to  observe the log to see if this problem still existed.

zhzhuang-zju:
> Yes, but the problem at that time was that a resource was processed by more than one worker. I want to observe the log to see if this problem still existed.

Get it. But this seems more like information needed during a validation phase; is it necessary to bring it into the repo?

ryanwuer:
> > How can i make the test, checkout the branch and build an image?
> 
> @ryanwuer Yes, you're right

I've just tested using karmada-controller-manager image built locally, but the behavior is not as expected. 
My original issue is here: https://github.com/karmada-io/karmada/issues/6660

After suspension deleted in PP, suspension in rb is not deleted automatically.

zhzhuang-zju:
> I've just tested using karmada-controller-manager image built locally, but the behavior is not as expected. My original issue is here: #6660
> 
> After suspension deleted in PP, suspension in rb is not deleted automatically.

Is this an intermittent issue?

ryanwuer:
> > I've just tested using karmada-controller-manager image built locally, but the behavior is not as expected. My original issue is here: #6660
> > After suspension deleted in PP, suspension in rb is not deleted automatically.
> 
> Is this an intermittent issue?

I've tested twice, it looks like it's not an intermittent issue

zhzhuang-zju:
> I've tested twice, it looks like it's not an intermittent issue

That might not be the same issue. The problem I fixed was the inconsistency between rb and pp, but it was a very occasional issue. I need to locate it according to your original issue.

zhzhuang-zju:
https://github.com/karmada-io/karmada/compare/master...zhzhuang-zju:upstream/rbsuspension?expand=1

@ryanwuer Can you make this change and verify it again?

ryanwuer:
> https://github.com/karmada-io/karmada/compare/master...zhzhuang-zju:upstream/rbsuspension?expand=1
> 
> @ryanwuer Can you make this change and verify it again?

I'm seeing that you make a comment here to clarify you will fix this issue ASAP, so the commit above is the fix method?

https://github.com/karmada-io/karmada/issues/6660#issuecomment-3232352171

zhzhuang-zju:
Yes, you can first verify using this commit.

ryanwuer:
> Yes, you can first verify using this commit.

The fix is useful. Can you put the fix into current PR?

zhzhuang-zju:
> The fix is useful. Can you put the fix into current PR?

Thank you for the verification. I've submitted a new PR to address this issue, see https://github.com/karmada-io/karmada/pull/6707

CharlesQQ:
> @CharlesQQ @ryanwuer   Can you help run a quick test for this?

I add new log
```
	rand.Seed(time.Now().UnixNano())
	randStr := RandomString(20)
	startTime := time.Now()
	defer func() {
		klog.Infof("Finished Reconcile, (%s/%s), cost: %v, randStr: %s", clusterWideKeyWithConfig.ClusterWideKey.Namespace, clusterWideKeyWithConfig.ClusterWideKey.Name, time.Since(startTime), randStr)
	}()

	clusterWideKey := clusterWideKeyWithConfig.ClusterWideKey
	resourceChangeByKarmada := clusterWideKeyWithConfig.ResourceChangeByKarmada
	klog.Infof("Reconciling object: %s, resourceChangeByKarmada: %v, randStr: %s", clusterWideKey, resourceChangeByKarmada, randStr)
```
Issues fixed: https://github.com/karmada-io/karmada/issues/6379
From the log, the same resource object will not be executed by multiple reconcile at the same time; the value of `resourceChangeByKarmada` is always false

image


add error log:
```
	err = retry.RetryOnConflict(retry.DefaultRetry, func() (err error) {
		operationResult, err = controllerutil.CreateOrUpdate(context.TODO(), d.Client, bindingCopy, func() error {
			// If this binding exists and its owner is not the input object, return error and let garbage collector
			// delete this binding and try again later. See https://github.com/karmada-io/karmada/issues/2090.
			if ownerRef := metav1.GetControllerOfNoCopy(bindingCopy); ownerRef != nil && ownerRef.UID != object.GetUID() {
				return fmt.Errorf("failed to update binding due to different owner reference UID, will " +
					"try again later after binding is garbage collected, see https://github.com/karmada-io/karmada/issues/2090")
			}

			bindingCopy.Annotations = util.DedupeAndMergeAnnotations(bindingCopy.Annotations, binding.Annotations)
			bindingCopy.Finalizers = util.DedupeAndMergeFinalizers(bindingCopy.Finalizers, binding.Finalizers)
			bindingCopy.OwnerReferences = binding.OwnerReferences
			bindingCopy.Spec.Resource = binding.Spec.Resource
			bindingCopy.Spec.ReplicaRequirements = binding.Spec.ReplicaRequirements
			bindingCopy.Spec.Replicas = binding.Spec.Replicas
			bindingCopy.Spec.PropagateDeps = binding.Spec.PropagateDeps
			bindingCopy.Spec.SchedulerName = binding.Spec.SchedulerName
			bindingCopy.Spec.Placement = binding.Spec.Placement
			bindingCopy.Spec.Failover = binding.Spec.Failover
			bindingCopy.Spec.ConflictResolution = binding.Spec.ConflictResolution
			bindingCopy.Spec.PreserveResourcesOnDeletion = binding.Spec.PreserveResourcesOnDeletion
			bindingCopy.Spec.SchedulePriority = binding.Spec.SchedulePriority
			if binding.Spec.Suspension != nil {
				if bindingCopy.Spec.Suspension == nil {
					bindingCopy.Spec.Suspension = &workv1alpha2.Suspension{}
				}
				bindingCopy.Spec.Suspension.Suspension = binding.Spec.Suspension.Suspension
			}
			excludeClusterPolicy(bindingCopy)
			return nil
		})
		if err != nil {
			klog.Errorf("Inner Failed to apply policy(%s) for object: %s. error: %v", policy.Name, objectKey, err)
			return err
		}
```
Even if you createOrUpdate rb  try again, it will not affect the result
image

RainbowMango (approved):
/approve

PR #6684, opened 2025-08-24
Fix unreasonable informer stop behavior

Mou887:
**What type of PR is this?**

/kind bug

This is an automated cherry-pick of #6647 into the `release-1.14` branch.
It fixes the unreasonable informer stop behavior so informer managers are not stopped on transient cache-sync failures, and switches logging to structured `klog.ErrorS`.


zhzhuang-zju:
Thanks
/lgtm

PR #6685, opened 2025-08-24
Adding proposal for workload affinity and anti-affinity support

mszacillo:
**What type of PR is this?**
/kind feature

This PR submits a proposal for adding workload affinity and workload anti-affinity as rules that the user can define on the PropagationPolicy. These rules would allow Karmada to support more complex scheduling challenges such as:
- Spreading duplicate FlinkDeployment pipelines across clusters to guarantee HA
- Scheduling many smaller training jobs to be co-located in same cluster


RainbowMango:
Thank you @mszacillo . We can try to start it in release-1.16.
/milstone v1.16

mszacillo:
Hi @kevin-wangzefeng, we went over this proposal (specifically going over the use-cases and the API design) during the community meeting today. I think the next step here is agreeing on the API changes, and once that is solidified, starting to determine an implementation strategy. Was hoping you could take a look, given your contributions to this feature in Kubernetes. Thank you!!

docs/proposals/scheduling/anti-affinity-scheduling-support/README.md
```diff
@@ -0,0 +1,256 @@
...
+
+In order to support workload affinity / anti-affinity, we will need to make changes to the PropagationPolicy API spec. We will be borrowing the concept of pod affinity and anti-affinity from Kubernetes to create an analogous behavior on a workload level.
+
+### PropagationPolicy
+
+```yaml
+// Placement represents the rule for select clusters.
+type Placement struct {
+
+  // WorkloadAffinity represents the inter-workload anti-affinity scheduling policies.
```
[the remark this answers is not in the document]
RainbowMango: +1
It seems a typo.

docs/proposals/scheduling/anti-affinity-scheduling-support/README.md
```diff
@@ -0,0 +1,256 @@
...
+- A FlinkDeployment with Kafka Consumers (for ultra-low latency requirements)
+- A RayService that relies on RedisCluster for caching (colocated prevents cross-cluster network hops)
+
+## API Changes
+
+In order to support workload affinity / anti-affinity, we will need to make changes to the PropagationPolicy API spec. We will be borrowing the concept of pod affinity and anti-affinity from Kubernetes to create an analogous behavior on a workload level.
+
+### PropagationPolicy
+
+```yaml
```
RainbowMango:
```suggestion
```golang
```
same as below.

RainbowMango (commented):
Looks pretty good to me.

mszacillo:
Thanks for the review @RainbowMango! I wasn't sure whether to be more thorough with the implementation details - perhaps we can discuss this more in the next community meeting. I can start to draft some iteration tasks in the meantime.

RainbowMango:
I was focusing on the API changes part during last review. I will leave my comments here about the `code changes` part.  Also, I hope to explore the way how Kubernetes implements the affinity/anti-affinity feature.

mszacillo:
Hi @XiShanYongYe-Chang @RainbowMango, are there any comments / thoughts on the proposal or implementation details?

RainbowMango:
My apologies for the delay. I'm currently focused on delivering the planned features for this release cycle, along with some mentor projects that are approaching their deadlines. I will get back on this as soon as possible.

docs/proposals/scheduling/anti-affinity-scheduling-support/README.md
```diff
@@ -0,0 +1,256 @@
...
+  // and the ones listed in the namespaces field.
+  // null selector and null or empty namespaces list means "this workload's namespace".
+  // An empty selector ({}) matches all namespaces.
+  // +optional
+  NamespaceSelector *metav1.LabelSelector `json:"namespaceSelector,omitempty"`
+
+  // Cluster label key defining the co-location domain (e.g., topology.karmada.io/cluster|zone|region).
+  // +required
+  TopologyKey string `json:"topologyKey,omitempty"`
+}
```
RainbowMango, line 296: An idea to simplify the API is to let the `WorkloadAffinityTerm` as a rule for identifying a group of workloads. And then, we can get a group of Clusters from the location they have been scheduled.
- for affinity case: the entire group cluster could be the candidate; 
- for anti-affinity case: the entire group cluster would be skipped from selection.

mszacillo: Perhaps I am not fully understanding the idea - would this be some sort of pre-filtering step when determining feasible clusters, so that the "feasible" clusters follow these descriptions:
- for affinity case: the entire group cluster could be the candidate;
- for anti-affinity case: the entire group cluster would be skipped from selection.

The way I've been thinking about this problem is during the scheduling cycle we'd do the following:
1. A new ResourceBinding is generated with AntiAffinityTerm 
2. Based on the matching selection criteria (labelSelector, namespace, etc), we can create some sort of affinity-key that can be used to fetch matching topologyKeys. For example, if we have a workload on namespace `test` with `karmada.io/group: my-indexer-job`, the anti-affinity key could be `test-my-indexer-job`. The value would be a set of clusters with a workload matching the antiAffinityTerm. 
3. Once the binding is added to the scheduling queue, we reference our ClusterSnapshot which would point to this antiAffinity map. 
4. We check if the map contains the antiAffinityKey. If so, we filter out clusters in the returned set. 
5. After a scheduling result is made, we update the map with the cluster for the matching affinityKey before the next scheduling cycle. 

I'm still working on typing this up and updating the proposal, but at least this would provide a more efficient lookup rather than looping through all clusters and all bindings with matching affinity-terms just to making a scheduling decision.

mszacillo: That said, this still isn't perfect since in order to main symmetric affinity rules we'd still need to check all other workloads on the cluster with affinity rules to make sure the workload about to be scheduled does not violate those terms.

mszacillo:
Hi @RainbowMango @kevin-wangzefeng, I've made several changes to the proposal based on our conversation during the last community meeting. 

1. Like Kevin had mentioned, I believe we should provide a new config that can be passed to the scheduler that will define the labels we must index over when validating affinity rules. Other labels will simply be ignored. 
2. I've decreased the complexity of the API and provided two options. The first gives the user more control to define a LabelSelector as well as MatchLabelKeys. The other option simplifies the API entirely and just turns affinity into something the user can enable. The default behavior would be to match workloads off the values of the label keys provided to the scheduler. 

Let me know what you think.

RainbowMango:
Thanks @mszacillo , I'm looking at it and will get back to you soon.

docs/proposals/scheduling/anti-affinity-scheduling-support/README.md
```diff
@@ -0,0 +1,356 @@
...
+- [ ] `ClusterInfo` and `ClusterSnapshot` enhancements to include affinity metadata
+- [ ] Introduce new Karmada Scheduler configurations to define labels for indexing affinity terms (`AntiAffinityLabelKeys`, `AffinityLabelKeys`)
+- [ ] Karmada Scheduler changes required to support affinity rules
+- [ ] Addition of affinity / anti-affinity filter plugins
+
+## API Changes
+
+In order to support workload affinity / anti-affinity, we will need to make changes to the PropagationPolicy API spec. We will be borrowing the concept of pod affinity and anti-affinity from Kubernetes to create an analogous behavior on a workload level.
+
+`Note:` Since indexing can become very expensive when making scheduling decisions based on anti-affinity / affinity terms, we are suggesting to expose a configuration to the Karmada Scheduler which would limit the set of labels that Karmada must keep track of. For example, if the user would like to enforce anti-affinity using the `karmada.io/group` label then they would define this argument on the scheduler:
```
RainbowMango: This is concerning from a user experience and flexibility perspective. (As you mentioned in the section below). 

ps: I'm still thinking about it...

RainbowMango: @mszacillo I'm thinking if we can define the group name in `PropagationPolicy.`, like:
```yaml
apiVersion: karmada.io/v1alpha1
kind: PropagationPolicy
metadata:
  name: flink-primary  # primary flink instance
spec:
  resourceSelectors:
  - apiVersion: flink.apache.org/v1beta1
    kind: FlinkDeployment
    name: flink
  placement:
    clusterAffinity:
      labelSelector:
        matchLabels:
          cluster-type: production
    workloadAffinity:
      affinityGroupNames: # define group names
      - flink.payment.ha
---

apiVersion: karmada.io/v1alpha1
kind: PropagationPolicy
metadata:
  name: flink-backup  # backup flink instance
spec:
  resourceSelectors:
  - apiVersion: flink.apache.org/v1beta1
    kind: FlinkDeployment
    name: payment-processor-b
  placement:
    clusterAffinity:
      labelSelector:
        matchLabels:
          cluster-type: production
    workloadAffinity:
      antiAffinity:
        affinityGroupNames:
        - flink.payment.ha
```
By the way, currently, this is still a draft idea, and if you want see it, maybe you can go to https://github.com/karmada-io/karmada/compare/master...RainbowMango:karmada:demo_affinity_feature.

mszacillo: Hmm I considered an approach similar to this, but how would this scale up if we are running hundreds of jobs on a single namespace? Would we create a propagationpolicy for every affinityGroupName? Or would the affinityGroupName refer to the label that we are indexing against? For example:

`karmada.group: indexer-job` or `karmada.group: aggregator`?

RainbowMango: That's an awesome input. I'll think about it. 

One more input I want from you is how your FlinkDeployments are managed by PropagationPolicy. Each FlinkDeployment has a PropagationPolicy; do all of them share a PropagationPolicy?

mszacillo: All FlinkDeployments on a namespace share a PropagationPolicy. At the moment to support duplicate pipelines we use two PropagationPolicies that select FlinkDeployments based on labels (example: primary, backup).  I haven't finished internally implementing workload anti-affinity, but the goal is to merge both these PropagationPolicies together into a single policy and have Karmada handle separating the grouped pipelines by cluster.

RainbowMango: Got it. Then, we still need to determine affinity relationships based on `labels`. My idea is to add a field under `.spec.placement.workloadAffinity` to specify which label should be used to determine affinity relationships. This would allow the scheduler to build indexes for ResourceBinding, for fast matching during scheduling, (Don't have to search all ResourceBindings).

mszacillo: Yes, that would be perfect

RainbowMango:
@mszacillo @kevin-wangzefeng Could you please take a look at this API design? The design idea is explained in the comments. 
```golang
// WorkloadAffinity defines inter-workload affinity and anti-affinity rules.
type WorkloadAffinity struct {
	// AffinityLabelKey declares the label key on the workload template that exposes
	// its affinity identity. The label value under this key represents the
	// workload's group identifier that other workloads can match against.
	//
	// Only a single key is supported initially. If multiple keys are needed in
	// the future, a new field (e.g., labelKeys) can be added without breaking
	// compatibility.
	//
	// The key must be a valid Kubernetes label key.
	//
	// +required
	AffinityLabelKey string `json:"affinityLabelKey"`

	// Affinity represents the inter-workload affinity scheduling rules.
	// These are hard requirements - workloads will only be scheduled to clusters that
	// satisfy the affinity term if specified.
	//
	// +optional
	Affinity *WorkloadAffinityTerm `json:"affinity,omitempty"`

	// AntiAffinity represents the inter-workload anti-affinity scheduling rules.
	// These are hard requirements - workloads will be scheduled to avoid clusters
	// where matching workloads are already scheduled.
	//
	// +optional
	AntiAffinity *WorkloadAntiAffinityTerm `json:"antiAffinity,omitempty"`

	// Note: The Affinity and AntiAffinity terms are both required to be met during scheduling.
	// If we need more flexible rules (e.g., preferred scheduling), we can consider adding
	// PreferredAffinity and PreferredAntiAffinity fields in the future.
}

// WorkloadAffinityTerm defines affinity rules for co-locating with specific workload groups.
type WorkloadAffinityTerm struct {
	// RequireSameGroup indicates the workload should co-locate with workloads
	// that share the same affinity group. When enabled, the scheduler will only
	// permit clusters where workloads with the same affinity group already exist.
	//
	// This field is required currently as more expressive selectors are not yet
	// supported. When additional selector fields are introduced in the future,
	// this field will be made optional to allow alternative affinity matching strategies.
	//
	// +kubebuilder:validation:Enum=true
	// +required
	RequireSameGroup *bool `json:"requireSameGroup"`

	// AllowBootstrap controls whether this workload can be scheduled when no workloads
	// with the same affinity group currently exist in the system.
	//
	// It defaults to true, which means the affinity requirement is relaxed for the first workload of
	// a group - if no existing workloads with the same affinity group are found, the scheduler will not
	// block scheduling. This allows bootstrapping new workload groups without
	// encountering scheduling deadlocks, providing the best user experience out of the box.
	//
	// When set to false, strict affinity enforcement is applied - the workload
	// will only be scheduled if workloads with the same affinity group already exist in the system.
	//
	// +kubebuilder:default=true
	// +optional
	AllowBootstrap *bool `json:"allowBootstrap,omitempty"`
}

// WorkloadAntiAffinityTerm defines anti-affinity rules for separating from specific workload groups.
type WorkloadAntiAffinityTerm struct {
	// RejectSameGroup indicates the workload should reject co-locating with workloads
	// that share the same affinity group. When enabled, the scheduler will reject
	// clusters where workloads with the same affinity group already exist.
	// This is a hard constraint - workloads will not be scheduled to such clusters.
	//
	// This field is required currently as more expressive selectors are not yet
	// supported. When additional selector fields are introduced in the future,
	// this field will be made optional to allow alternative anti-affinity matching strategies.
	//
	// +kubebuilder:validation:Enum=true
	// +required
	RejectSameGroup *bool `json:"rejectSameGroup"`
}
```

mszacillo:
> Could you please take a look at this API design? The design idea is explained in the comments.

@RainbowMango Thanks for the API update! Reading through it just now, it seems like it covers my requirements quite well and is simple enough for new users to be able to test the feature themselves without having to alter any of the scheduler configurations. 

The one potentially confusing thing for users might be the `RejectSameGroup ` and `RequireSameGroup` settings, since they essentially just serve the function of "toggling on" the Affinity type. I understand why this was designed this way though, since we expect more expressive selectors to potentially be supported in the future. Perhaps we could just have these default to true if the user has already defined `Affinity` or `AntiAffinity`, so that the user doesn't have to set this additional field themselves? That's more a nit than anything though. 

By the way, I mocked up an internal implementation of the feature using a simplified API that ended up quite similar to yours actually - my AntiAffinityWorkloadTerm is:

```golang
type WorkloadAntiAffinityTerm struct {
	// Determines if Karmada scheduler will take affinity terms into account for this workload.
	// Affinity labels are controlled via the --affinity-label-key configuration defined on the Karmada Scheduler.
	// Affinity and anti-affinity peers are determined based off matching values of the keys specified in the scheduler configuration.
	// +required
	AffinityEnabled bool `json:"affinityEnabled,omitempty"`
```

I used the original idea of a scheduler config and then parse the `--affinity-label-key` config (`karmada.io/group` in this case) and use that when generating the ResourceBinding, which will include the label in the ObjectReference. I reuse this label's value during scheduling using the updated cache + snapshot which I described a bit in the proposal. And then filter out clusters using a custom plugin. I'll share a link to the implementation tomorrow morning - I need to port it over to my github repo since its currently saved on an internal repo which I can't provide a link to.

mszacillo:
@RainbowMango Went ahead and updated my POC with your API suggestion. Did some E2E testing to confirm that this will work, which it did! :) 

My truncated PropagationPolicy with the new fields:

```yaml
spec:
  placement:
    workloadAffinity:
      affinityLabelKey: karmada.io/group
      antiAffinity:
        rejectSameGroup: true
```
When scheduling my initial FlinkDeployment which has been labeled with the affinityLabelKey:

```yaml
apiVersion: flink.apache.org/v1beta1
kind: FlinkDeployment
metadata:
  labels:
    karmada.io/group: aggregator-job   <---- affinity label
  name: a-22-ticker-demo
  namespace: s-spaaseng
...
```

This generates a ResourceBinding that has parsed out the relevant affinity label key:

```yaml
spec:
  clusters:
  - name: spaas-kaas-ob-dev
    replicas: 2
  placement:
    workloadAffinity:
      affinityLabelKey: karmada.io/group
      antiAffinity:
        rejectSameGroup: true
  resource:
    affinityGroupLabel:
      karmada.io/group: aggregator-job   <---- affinity label
    apiVersion: flink.apache.org/v1beta1
    kind: FlinkDeployment
    name: a-22-ticker-demo
    namespace: s-spaaseng
    resourceVersion: "10794181"
    uid: fc39af17-d5a2-4c9b-94ee-a30fdd8edf32
```

During scheduling phase, I updated the snapshot object + scheduler cache to index these resourcebindings depending on their affinity label keys. Once the snapshot is generated, we re-use it during our filter plugin which will filter out clusters that violate the affinity term for the resource being scheduled. I rescheduled two more applications with the same label key to test this out, and saw the scheduler filtering out clusters that had a conflicting RB!

Second application:

```
I1216 22:29:20.123054       1 scheduler.go:437] Start to schedule ResourceBinding(s-spaaseng/a-4-ticker-demo-flinkdeployment) as placement changed
I1216 22:29:20.123189       1 scheduler.go:605] "Begin scheduling resource binding with ClusterAffinity" resourceBinding="s-spaaseng/a-4-ticker-demo-flinkdeployment"
I1216 22:29:20.123555       1 generic_scheduler.go:142] Cluster "spaas-kaas-ob-dev" is not fit, reason: cluster violates this resource bindings anti-affinity term
I1216 22:29:20.123821       1 generic_scheduler.go:142] Cluster "spaas-kaas-tt-dev" is not fit, reason: cluster(s) had untolerated taint {cluster.karmada.io/maintenance:NoSchedule}
I1216 22:29:20.123863       1 generic_scheduler.go:89] Feasible clusters found: [spaas-kaas-pw-dev]
```

Third application (scheduling unsuccessful):

```
I1216 22:30:41.408245       1 scheduler.go:437] Start to schedule ResourceBinding(s-spaaseng/a-5-ticker-demo-flinkdeployment) as placement changed
I1216 22:30:41.408299       1 scheduler.go:605] "Begin scheduling resource binding with ClusterAffinity" resourceBinding="s-spaaseng/a-5-ticker-demo-flinkdeployment"
I1216 22:30:41.408667       1 generic_scheduler.go:142] Cluster "spaas-kaas-ob-dev" is not fit, reason: cluster violates this resource bindings anti-affinity term
I1216 22:30:41.408706       1 generic_scheduler.go:142] Cluster "spaas-kaas-pw-dev" is not fit, reason: cluster violates this resource bindings anti-affinity term
I1216 22:30:41.408785       1 generic_scheduler.go:142] Cluster "spaas-kaas-tt-dev" is not fit, reason: cluster(s) had untolerated taint {cluster.karmada.io/maintenance:NoSchedule}
I1216 22:30:41.408818       1 scheduler.go:623] ResourceBinding(s-spaaseng/a-5-ticker-demo-flinkdeployment) scheduled to clusters []
I1216 22:30:41.421980       1 scheduler.go:724] Patch schedule to ResourceBinding(s-spaaseng/a-5-ticker-demo-flinkdeployment) succeed
I1216 22:30:41.422066       1 event_handler.go:181] Ignore update event of resourceBinding s-spaaseng/a-5-ticker-demo-flinkdeployment as specification no change
I1216 22:30:41.422066       1 scheduler.go:629] "End scheduling resource binding with ClusterAffinity" resourceBinding="s-spaaseng/a-5-ticker-demo-flinkdeployment"
I1216 22:30:41.422227       1 event.go:389] "Event occurred" object="s-spaaseng/a-5-ticker-demo" fieldPath="" kind="FlinkDeployment" apiVersion="flink.apache.org/v1beta1" type="Warning" reason="ScheduleBindingFailed" message="0/3 clusters are available: 1 cluster(s) had untolerated taint {cluster.karmada.io/maintenance:NoSchedule}, 2 cluster violates this resource bindings anti-affinity term."
```

Changing the group value allows scheduling to succeed. 

Obviously some optimizations will be necessary, but I thought this would give us some confidence that this API seems to work quite well. Please take a look at the POC branch I've linked below, I've already started drafting some iteration tasks based off the POC implementation.

Here is the [branch](https://github.com/mszacillo/karmada/commit/752f5d9befd25f54e603489c537bf48b1c1e9cee) for reference.

RainbowMango:
> Perhaps we could just have these default to true if the user has already defined Affinity or AntiAffinity, so that the user doesn't have to set this additional field themselves?

Thanks for the feedback! I understand the concern about the seemingly redundant field. However, the current design prioritizes future extensibility and backward compatibility.

The key consideration is what happens when we introduce more expressive selectors in the future. If we default `RequireSameGroup` and `RejectSameGroup` to true now, we would need to change that default behavior later when alternative matching strategies become available. This could break existing user configurations in subtle ways.

Here's the scenario I'm concerned about:
1. `Today`: User creates a policy with `affinity: {}`, and `RequireSameGroup` defaults to true
1. `Future`: We add new selector fields like `targetGroups`, and probably we would change or remove the default setting.
1. `Problem`: The old policy still has the implicit `RequireSameGroup: true` default, which may conflict with or override the new selector semantics

By requiring users to explicitly set `RequireSameGroup: true` and `RejectSameGroup: true` now, we:
- Make the current behavior explicit and clear in the configuration
- Reserve the flexibility to make this field optional later without changing defaults
- Ensure that when we add new selectors, existing policies won't have hidden implicit behaviors that interfere with the new fields

When we introduce additional selector options, we can simply make `RequireSameGroup` and  `RejectSameGroup` optional, and users will be able to choose between different matching strategies without any migration concerns. This is a fully backward-compatible evolution path.

I agree it's a bit verbose for the initial use case, but I believe it's a worthwhile trade-off for long-term API stability and extensibility.

kevin-wangzefeng:
Thank you @mszacillo and @RainbowMango for the thoughtful discussion and valuable input on this proposal.

Designing APIs is always challenging. For inter-workload affinity, directly copying the upstream Pod affinity API is not an ideal choice for Karmada's use case. 

Over the past weeks, I have been continuously exploring simpler primitives to make this feature more straightforward and easier to use. One option I ever considered was introducing a concept similar to Volcano's PodGroup, but that would bring too much disruption to the Karmada Scheduler architecture.

Given the current state of the proposal discussion and the Karmada scheduler framework, I'd like to propose the following API design based on @RainbowMango 's version, please feel free to comment and share your feedback.

**Proposed API Design**
```go
// WorkloadAffinity defines inter-workload affinity and anti-affinity rules.
type WorkloadAffinity struct {
    // Affinity represents the inter-workload affinity scheduling rules.
    // These are hard requirements - workloads will only be scheduled to clusters that
    // satisfy the affinity term if specified.
    //
    // For the first workload of an affinity group (when no workloads with matching
    // label value exist in the system), the scheduler will not block scheduling.
    // This allows bootstrapping new workload groups without encountering scheduling
    // deadlocks, providing the best user experience.
    //
    // +optional
    Affinity *WorkloadAffinityTerm `json:"affinity,omitempty"`

    // AntiAffinity represents the inter-workload anti-affinity scheduling rules.
    // These are hard requirements - workloads will be scheduled to avoid clusters
    // where matching workloads are already scheduled.
    //
    // +optional
    AntiAffinity *WorkloadAntiAffinityTerm `json:"antiAffinity,omitempty"`

    // Note: The Affinity and AntiAffinity terms are both required to be met during scheduling.
    // If we need more flexible rules (e.g., preferred scheduling), we can consider adding
    // PreferredAffinity and PreferredAntiAffinity fields in the future.
}

// WorkloadAffinityTerm defines affinity rules for co-locating with specific workload groups.
type WorkloadAffinityTerm struct {
    // GroupByLabelKey declares the label key on the workload resource template that
    // determines the affinity group. Workloads with the same label value under this key
    // belong to the same affinity group.
    //
    // The scheduler maintains a global index of affinity groups in memory for efficient
    // lookup. Each affinity group is identified by a serialized key-value pair, and
    // contains all workload resource templates that belong to the group.
    //
    // The key must be a valid Kubernetes label key.
    //
    // Example: If GroupByLabelKey is "app.group", workloads with label "app.group=frontend"
    // will form one affinity group, while those with "app.group=backend" will form another.
    //
    // +required
    GroupByLabelKey string `json:"groupByLabelKey"`
}

// WorkloadAntiAffinityTerm defines anti-affinity rules for separating from specific workload groups.
type WorkloadAntiAffinityTerm struct {
    // GroupByLabelKey declares the label key on the workload resource template that
    // determines the affinity group. Workloads with the same label value under this key
    // belong to the same affinity group and should be separated.
    //
    // The scheduler maintains a global index of affinity groups in memory for efficient
    // lookup. Each affinity group is identified by a serialized key-value pair, and
    // contains all workload resource templates that belong to the group.
    //
    // The key must be a valid Kubernetes label key.
    //
    // Example: If GroupByLabelKey is "app.group", workloads with label "app.group=frontend"
    // will avoid clusters where other "app.group=frontend" workloads exist.
    //
    // +required
    GroupByLabelKey string `json:"groupByLabelKey"`
}
```

**Usage Examples**

**Example 1: Affinity Scheduling**

```yaml
kind: PropagationPolicy
metadata:
  name: frontend-affinity
spec:
  workloadAffinity:
    affinity:
      groupByLabelKey: "app.tier"
  # Workloads with label app.tier=frontend will be scheduled to clusters
  # where other app.tier=frontend workloads already exist.
  # The first frontend workload will be scheduled normally (bootstrap behavior).
```

**Example 2: Anti-Affinity Scheduling**

```yaml
kind: PropagationPolicy
metadata:
  name: app-anti-affinity
spec:
  workloadAffinity:
    antiAffinity:
      groupByLabelKey: "app.instance"
  # Workloads with label app.instance=primary will avoid clusters
  # where other app.instance=primary workloads exist.
```

**Example 3: Combined Affinity and Anti-Affinity**

```yaml
kind: PropagationPolicy
metadata:
  name: complex-affinity
spec:
Comment view  workloadAffinity:
    affinity:
      groupByLabelKey: "app.group"
    antiAffinity:
      groupByLabelKey: "app.instance"
  # Workloads need to co-locate with those having the same app.group value,
  # but avoid co-locating with those having the same app.instance value.
```

RainbowMango:
It looks awesome! This version is even better than the one I proposed above. 
- It eliminates the confusion fields `RejectSameGroup` and `RequireSameGroup`
- Allows the combination use case of both affinity and anti-affinity. 
- Make the `AllowBootstrap` the reasonable default behavior.

@mszacillo, Please take a look when you get back from holiday. No rush, I'm also on vacation for the next three days. :)

RainbowMango:
PS: I opened an [umbrella issue](https://github.com/karmada-io/karmada/issues/7064) for this feature and left some draft iteration tasks there, hoping to share and discuss at the next community meeting.

mszacillo:
Thank you @kevin-wangzefeng! Agreed with @RainbowMango, I think that API design is simple and eliminates some of the fields I had some reservations about (RejectSameGroup, RequireSameGroup). I think it could also be plugged into the demo branch I shared previously with minimal changes. I'll update the proposal with this API definition unless there are additional concerns.

RainbowMango:
The next and the last API design for this feature would be the `ResourceBinding`. 

Once we introduce the `WorkloadAffinity` field into the `.spec.placement` of `PropagationPolicy`, the `ResourceBinding` will inherit it from the shared `type Placement struct` structure. For the time being, the `ResourceBinding` doesn't need it, but the scheduler might need it when the `WorkloadAffinity` becomes more expressive(like holding more rules). So we don't have to change anything, just introduce the new `WorkloadAffinity` field to PropagationPolicy as planned. 

We still need a place/field to hold the `instantiated` affinity and anti-affinity group names. The field will be populated by controllers and consumed by schedulers for decisions. 

In conclusion, I'm proposing the following API to the `.spec` of `ResourceBinding`:
```golang
type ResourceBindingSpec struct {
	// ... the existing fields

	// WorkloadAffinityGroups represents instantiated grouping results from Placement.WorkloadAffinity, used to
	// keep workloads with the same affinity group co-located or those with the same anti-affinity group
	// separated across clusters. Populated by controllers, the scheduler consumes it for decisions.
	// +optional
	WorkloadAffinityGroups *WorkloadAffinityGroups `json:"workloadAffinityGroups,omitempty"`
}

// WorkloadAffinityGroups stores the instantiated affinity and anti-affinity group names derived from
// PropagationPolicy/ClusterPropagationPolicy WorkloadAffinity. Each group name is serialized as
// "<labelKey>=<labelValue>" (for example: "app.group=frontend"), and is used by the scheduler to
// co-locate workloads in the same affinity group and to separate workloads in the same anti-affinity
// group.
// Note: If multiple groups are needed later, prefer adding list-typed fields (e.g., WorkloadAffinityGroups, AntiAffinityGroups)
// to keep backward compatibility.
type WorkloadAffinityGroups struct {
	// AffinityGroup is the instantiated group name derived from affinity rules.
	// +optional
	AffinityGroup string `json:"affinityGroup,omitempty"`

	// AntiAffinityGroup is the instantiated group name derived from anti-affinity rules.
	// +optional
	AntiAffinityGroup string `json:"antiAffinityGroup,omitempty"`
}
```

Please let me know what you think of it or want to change anything. @kevin-wangzefeng @mszacillo

mszacillo:
> For the time being, the ResourceBinding doesn't need it, but the scheduler might need it when the WorkloadAffinity becomes more expressive(like holding more rules). So we don't have to change anything, just introduce the new WorkloadAffinity field to PropagationPolicy as planned.

Makes sense - sounds like something we can come back to once there are more complex use-cases for affinity rules. 

> Please let me know what you think of it or want to change anything

Generally I think it looks great. Out of curiosity, would these affinity / anti-affinity groups be namespace specific? Would it be an issue if two PropagationPolicies use the same GroupLabelKey on different namespaces? My thought is that they should be separate, but perhaps this can be handled during the actual indexing portion. 

In my demo I construct an anti-key which is just `namespace-labelkey-groupvalue`. With the new AffinityGroup + AntiAffinityGroup fields on the RB, this would just be `namespace-affinitygroup` or `namespace-antiaffinitygroup`. But I wonder what your thought is. 

Otherwise, I think this is functionally exactly what we need. I've updated the proposal to include the new agreed upon API definitions, and I've updated the code changes section to match the iteration tasks as well as the demo branch which I had linked previously.

RainbowMango:
> Generally I think it looks great. Out of curiosity, would these affinity / anti-affinity groups be namespace specific? Would it be an issue if two PropagationPolicies use the same GroupLabelKey on different namespaces? My thought is that they should be separate, but perhaps this can be handled during the actual indexing portion.

That's a good question!

According to the current PropagationPolicy API design, we only specify a label to identify the affinity/anti-affinity group for workloads -- but we don't declare whether this grouping should be scoped to a namespace. As a result, affinity and anti-affinity are currently applied at the `cluster-wide` level, not per namespace.

If we want to honor affinity/anti-affinity rules within a namespace, we'll need to extend the API, otherwise, workloads in different namespaces cannot share the same label.

Are you considering how to set labels for workloads running across multiple tenant namespaces?

mszacillo:
> Are you considering how to set labels for workloads running across multiple tenant namespaces?

Well, in our case we're going to set the group label for our users based on the name of their application. For duplicate pipelines our users typically use the same name, such as `indexer-job-a` and `indexer-job-b`. For our case, the group label would be `indexer-job`.  

The issue starts if we happen to have a user name their job exactly the same as a job in an another namespace - this would affect scheduling through workload anti-affinity rules. Perhaps this is low probability. 

> If we want to honor affinity/anti-affinity rules within a namespace, we'll need to extend the API, otherwise, workloads in different namespaces cannot share the same label.

Is this true? Could we not index RBs based on their namespace and group name, which wouldn't require an API change. Or do you prefer that this type of behavior is clearly defined and controlled by the API design?

RainbowMango:
> The issue starts if we happen to have a user name their job exactly the same as a job in an another namespace - this would affect scheduling through workload anti-affinity rules. Perhaps this is low probability.

Yeah, I get it. I think we shouldn't take those chances. :)

> Is this true? Could we not index RBs based on their namespace and group name, which wouldn't require an API change. Or do you prefer that this type of behavior is clearly defined and controlled by the API design?

Yes, I prefer that the behavior be clearly defined in the API. Just like the [Namespaces](https://github.com/kubernetes/kubernetes/blob/efc3126b76833fb60ff92de47bfeaf4e15a108f6/staging/src/k8s.io/api/core/v1/types.go#L3924-L3930) and [NamespaceSelector](https://github.com/kubernetes/kubernetes/blob/efc3126b76833fb60ff92de47bfeaf4e15a108f6/staging/src/k8s.io/api/core/v1/types.go#L3937-L3943) in `the PodAffinityTerm`.

> Could we ~~not~~ index RBs based on their namespace and group name, which wouldn't require an API change.

(I guess you mean index RBs based on both namespace and group name).
This kind of the default behavior. It is doable, but the behavior should be documented somewhere.

@kevin-wangzefeng Any thoughts?

RainbowMango:
The use case is now very clear—you actually need workloads to have anti-affinity scoped within their own namespace.  Now, I don't think we need to change the current API; it can be extended later if such a need arises. For now, we just need to document this behavior clearly in the API documentation. @mszacillo what do you say?

mszacillo:
> The use case is now very clear—you actually need workloads to have anti-affinity scoped within their own namespace. Now, I don't think we need to change the current API; it can be extended later if such a need arises. For now, we just need to document this behavior clearly in the API documentation. @mszacillo what do you say?

Sounds good, I'll make the update to the API documentation in the proposal shortly!

RainbowMango:
I will take a look on Monday.

docs/proposals/scheduling/anti-affinity-scheduling-support/README.md
```diff
@@ -0,0 +1,384 @@
...
+### Goals
+
+- Extend the Propagation Policy API to support workload affinity and anti-affinity based scheduling rules.
+- Provide a way for users who want HA for their data processing pipelines (or other workloads) with anti-affinity.
+- Provide a solution for users who need workloads to be colocated and scheduled to the same clusters.
+
+### Non-Goals
+
+- The support of `PreferredDuringScheduling` is deferred from this proposal, as the use-case is not totally defined and this scheduling strategy is more computationally expensive
+- Cross-namespace anti-affinity / affinity is not supported. Existing API design inherits namespace from the PropagationPolicy, so affinity terms are limited to within the namespace. This can be added in the future if there are sufficient use-cases.
```
RainbowMango, line 38: The `namespace` would be grabbed from the resource templates, right? As the API design is shared by both `ClusterPropagationPolicy` and `PropagationPolicy`.

docs/proposals/scheduling/anti-affinity-scheduling-support/README.md
```diff
@@ -0,0 +1,384 @@
...
+
+	// AntiAffinityGroup is the instantiated group name derived from anti-affinity rules.
+	// +optional
+	AntiAffinityGroup string `json:"antiAffinityGroup,omitempty"`
+}
+```
+
+### WorkloadAffinityTerm API Definition Options
+
+While constructing the WorkloadAffinity and WorkloadAntiAffinityTerm API definitions, we thought of a couple options. They are presented here in order, with option 3 being the one we moved forward with.
```
RainbowMango, line 221: Great to have the alternatives here! Thanks.

PR #6686, opened 2025-08-25
When controller calls IsAPIEnabled, it also considers whether the judgment result is trustworthy

XiShanYongYe-Chang:
**What type of PR is this?**

/kind cleanup

The `IsAPIEnabled` function, when determining whether a target API is enabled, only relies on the `APIEnablements` field in the current cluster status and does not take into account the `CompleteAPIEnablements` condition type. As a result, the outcome of the `IsAPIEnabled` function's judgment may be untrusted.

The `IsAPIEnabled` function is called by multiple controllers, which may affect business logic due to incorrect judgments made by IsAPIEnabled during processing. The specific impact analysis is as follows:

1. https://github.com/karmada-io/karmada/blob/17caa1bc7d8b27ad9ccd99d5ddbd20f6b95af5b0/pkg/controllers/multiclusterservice/endpointslice_dispatch_controller.go#L341-L344 When the `APIEnablement` information collection in the consumer cluster is incomplete and the `discovery.k8s.io/v1 EndpointSlice` is not collected, it will affect the propagation of EndpointSlice resources to that cluster. The east-west capability of the `MulticlusterService` feature will be impaired, and all `MulticlusterService` resources in the consumer cluster that include this cluster will be affected.
2. https://github.com/karmada-io/karmada/blob/17caa1bc7d8b27ad9ccd99d5ddbd20f6b95af5b0/pkg/controllers/multiclusterservice/mcs_controller.go#L291-L294 When the `APIEnablement` information collection for the provider cluster is incomplete and the `discovery.k8s.io/v1 EndpointSlice` is not collected, it will affect the creation of work objects in the namespace corresponding to that cluster in the control plane for the `MulticlusterService` resource. The functionality of the `MulticlusterService` will be impaired, and all `MulticlusterService` resources in the provider cluster that include this cluster will be affected.
3. https://github.com/karmada-io/karmada/blob/17caa1bc7d8b27ad9ccd99d5ddbd20f6b95af5b0/pkg/search/controller.go#L415-L418 When the cluster API enablement information is incomplete, it may affect the karmada-search component's caching of resources from the target cluster; however, resources that have already been cached remain unaffected.
4. https://github.com/karmada-io/karmada/blob/17caa1bc7d8b27ad9ccd99d5ddbd20f6b95af5b0/pkg/search/proxy/controller.go#L249-L252 When the cluster API Enablement information is incomplete, it will affect the functionality of the karmada-search component proxy. As a result, the corresponding resources will not be able to establish a cache, and any already cached resources will also be deleted.

Fixes #

```release-note
NONE
```


RainbowMango (commented):
/assign

pkg/controllers/multiclusterservice/endpointslice_dispatch_controller.go
```diff
@@ -338,11 +338,18 @@ func (c *EndpointsliceDispatchController) dispatchEndpointSlice(ctx context.Cont
 				"Consumer cluster %s is not ready, skip to propagate EndpointSlice", clusterName)
 			continue
 		}
-		if !helper.IsAPIEnabled(clusterObj.Status.APIEnablements, util.EndpointSliceGVK.GroupVersion().String(), util.EndpointSliceGVK.Kind) {
```
RainbowMango, line 341: The `EndpointSlice, discovery.k8s.io/v1` API is kind of a hard restriction for the MultiClusterService feature. 

I'm thinking if it is appropriate to check this here.   Any thoughts?

pkg/controllers/multiclusterservice/mcs_controller.go
```diff
@@ -288,11 +288,18 @@ func (c *MCSController) propagateMultiClusterService(ctx context.Context, mcs *n
 				"Provider cluster %s is not ready, skip to propagate MultiClusterService", clusterName)
 			continue
 		}
-		if !helper.IsAPIEnabled(clusterObj.Status.APIEnablements, util.EndpointSliceGVK.GroupVersion().String(), util.EndpointSliceGVK.Kind) {
```
RainbowMango, line 291: Ditto.

pkg/controllers/multiclusterservice/endpointslice_dispatch_controller.go
XiShanYongYe-Chang: For example, directly bypassing this check, assuming that the API is supported in the default cluster, and even if it is not supported, an error will occur during the resource allocation process, making it noticeable to the user?

pkg/apis/cluster/v1alpha1/cluster_helper.go
```diff
@@ -16,7 +16,59 @@ limitations under the License.
...
+//
+// If the API is enabled and the result is trusted, it is recommended to
+// consider that the API is enabled.
+// If the API is not enabled and the result is trusted, it is recommended to
+// consider that the API is not enabled.
+// If the API is enabled and the result is untrusted, this case should
+// never happen.
+// If the API is not enabled and the result is untrusted, it is recommended to
+// consider that the API is enabled.
+func (c *Cluster) IsAPIEnabledAndTrusted(gvk schema.GroupVersionKind) (enabled bool, trusted bool) {
```
RainbowMango: It is not convenient to be consumed with two return booleans. How about this:
```golang

type APIEnablementStatus string

const (
	// APIEnabled means a cluster supports the specified API.
	APIEnabled APIEnablementStatus = "APIEnabled"
	// APIDisabled means a cluster does not support the specified API.
	APIDisabled APIEnablementStatus = "APIDisabled"
	// APIUnknown means it is unknown whether the cluster supports the specified API.
	APIUnknown APIEnablementStatus = "APIUnknown"
)

// APIEnablement checks if the target API (or CRD) referenced by gvk has been installed in the cluster.
// The check takes the CompleteAPIEnablements condition into account. If the CompleteAPIEnablements condition indicates
// the current APIEnablements is Partial, it returns APIEnabled if the gvk is found in the list; otherwise, the status is considered APIUnknown.
// This means that when the APIEnablements is Partial and the gvk is not present, we cannot definitively say the API is disabled.
func (c *Cluster) APIEnablement(gvk schema.GroupVersionKind) APIEnablementStatus {
	targetGroupVersion := gvk.GroupVersion().String()
	for _, apiEnablement := range c.Status.APIEnablements {
		if apiEnablement.GroupVersion != targetGroupVersion {
			continue
		}
		for _, resource := range apiEnablement.Resources {
			if resource.Kind != gvk.Kind {
				continue
			}
			return APIEnabled
		}
	}

	// At this point, we have the complete APIEnablements list for the cluster,
	// so we can confidently determine that the API is disabled if it was not found above.
	if meta.IsStatusConditionPresentAndEqual(c.Status.Conditions, ClusterConditionCompleteAPIEnablements, metav1.ConditionTrue) {
		return APIDisabled
	}

	return APIUnknown
}
```

pkg/scheduler/framework/plugins/apienablement/api_enablement.go
```diff
@@ -68,7 +68,8 @@ func (p *APIEnablement) Filter(
 	// we enforce strict checks to maintain consistency. This may occasionally
 	// exclude clusters prematurely. Users requiring a specific number of target
 	// clusters should use SpreadConstraints(in PropagationPolicy) to meet their requirements.
-	if helper.IsAPIEnabled(cluster.Status.APIEnablements, bindingSpec.Resource.APIVersion, bindingSpec.Resource.Kind) {
+	apiEnablement := cluster.APIEnablement(schema.FromAPIVersionAndKind(bindingSpec.Resource.APIVersion, bindingSpec.Resource.Kind))
+	if apiEnablement == clusterv1alpha1.APIEnabled {
```
XiShanYongYe-Chang: Does the logic on the scheduler side need to be modified synchronously: when it is Unknown, schedule.
@RainbowMango

RainbowMango: Why?

XiShanYongYe-Chang: Similar to the processing logic of the controller, when it is uncertain whether the API supports it, directly propagate it

pkg/apis/cluster/v1alpha1/cluster_helper.go
```diff
@@ -16,7 +16,52 @@ limitations under the License.
...
+		for _, resource := range apiEnablement.Resources {
+			if resource.Kind != gvk.Kind {
+				continue
+			}
+			return APIEnabled
+		}
+	}
+
+	// At this point, we have the complete APIEnablements list for the cluster,
+	// so we can confidently determine that the API is disabled if it was not found above.
```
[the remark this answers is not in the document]
RainbowMango: +1

pkg/scheduler/framework/plugins/apienablement/api_enablement.go
RainbowMango: No. This plugin is used to select candidates from a cluster pool, it should enforce strict checks. This is quite differenct from other cases.

pkg/search/proxy/controller.go
```diff
@@ -246,8 +245,9 @@ func (ctl *Controller) mergeResourcesByClusters(resourcesByClusters map[string]m
 			klog.Errorf("Failed to get gvk: %v", err)
 			continue
 		}
-		if !helper.IsAPIEnabled(cluster.Status.APIEnablements, gvk.GroupVersion().String(), gvk.Kind) {
-			klog.Warningf("Resource %s is not enabled for cluster %s", resource.String(), cluster)
+		apiEnablement := cluster.APIEnablement(gvk)
+		if apiEnablement == clusterv1alpha1.APIDisabled {
```
RainbowMango:
```suggestion
		if cluster.APIEnablement(gvk) == clusterv1alpha1.APIDisabled {
```
Why need an extra variable here?

test/e2e/framework/customresourcedefine.go
```diff
@@ -73,7 +74,8 @@ func WaitCRDPresentOnClusters(client karmada.Interface, clusters []string, crdAP
 			gomega.Eventually(func(g gomega.Gomega) (bool, error) {
 				cluster, err := FetchCluster(client, clusterName)
 				g.Expect(err).NotTo(gomega.HaveOccurred())
-				return helper.IsAPIEnabled(cluster.Status.APIEnablements, crdAPIVersion, crdKind), nil
+				apiEnablement := cluster.APIEnablement(schema.FromAPIVersionAndKind(crdAPIVersion, crdKind))
+				return apiEnablement == clusterv1alpha1.APIEnabled, nil
```
RainbowMango: ditto.

pkg/scheduler/framework/plugins/apienablement/api_enablement.go
```diff
@@ -68,7 +68,8 @@ func (p *APIEnablement) Filter(
 	// we enforce strict checks to maintain consistency. This may occasionally
 	// exclude clusters prematurely. Users requiring a specific number of target
 	// clusters should use SpreadConstraints(in PropagationPolicy) to meet their requirements.
-	if helper.IsAPIEnabled(cluster.Status.APIEnablements, bindingSpec.Resource.APIVersion, bindingSpec.Resource.Kind) {
+	apiEnablement := cluster.APIEnablement(schema.FromAPIVersionAndKind(bindingSpec.Resource.APIVersion, bindingSpec.Resource.Kind))
```
RainbowMango: Why need an extra variable here?

RainbowMango:
cc @GitHubxsy

PR #6690, opened 2025-08-26
purgeMode not properly set when tolerationTime is 0 when cluster failover occurs

### LINKED ISSUE #6676: purgeMode not properly set when tolerationTime is 0 in taintManager's syncBindingEviction function
**What happened**:

There is a logical issue in the syncBindingEviction function of the NoExecuteTaintManager where the purgeMode variable is not properly initialized when tolerationTime is 0.

In the current implementation:

https://github.com/karmada-io/karmada/blob/a5103d07acc781d372a8eb17a1dcce28acc604fd/pkg/controllers/cluster/taint_manager.go#L176-L184

When tolerationTime == 0 but needEviction is false, the purgeMode variable remains uninitialized (zero value), but the eviction still proceeds. This means:
- The purgeMode will be set to the zero value of policyv1alpha1.PurgeMode
- The eviction task may not behave as expected since the purge mode is not explicitly configured
- This could lead to inconsistent eviction behavior

The purgeMode should be properly set based on tc.NoExecuteTaintEvictionPurgeMode configuration whenever an eviction is going to be performed.

**Environment**:
- Karmada version: v1.14
- Others:

XiShanYongYe-Chang:
**What type of PR is this?**

/kind bug

ref #6676 

Fixes #6676

```release-note
`karmada-controller-manager`: Fixed the issue that purgeMode not properly set when tolerationTime is 0 when cluster failover occurs
```


pkg/controllers/cluster/taint_manager.go
```diff
@@ -268,8 +264,8 @@ func (tc *NoExecuteTaintManager) syncClusterBindingEviction(key util.QueueKey) e
 }
 
 // needEviction returns whether the binding should be evicted from target cluster right now.
-// If a toleration time is found, we return false along with a minimum toleration time as the
-// second return value.
+// If a toleration time is found, we return false expect the toleration time is 0 in which case
```
RainbowMango: `expect`? or `except`? That's quite different.

XiShanYongYe-Chang: Sorry, I made a mistake; it should be "except"

RainbowMango (commented):
/assign
/gemini review

pkg/controllers/cluster/taint_manager.go
```diff
@@ -267,9 +263,9 @@ func (tc *NoExecuteTaintManager) syncClusterBindingEviction(key util.QueueKey) e
 	return nil
 }
 
-// needEviction returns whether the binding should be evicted from target cluster right now.
-// If a toleration time is found, we return false along with a minimum toleration time as the
-// second return value.
+// needEviction returns whether the binding should be evicted from the target cluster right now.
+// If it returns false, further check of the toleration time is needed. A toleration time less than 0
+// means tolerate forever; otherwise, eviction need happen after the toleration time.
```
RainbowMango: > If it returns false, further check of the toleration time is needed.

This confuses me as not clear who should `further check`, it can be this function itself or the caller outside.

```suggestion
// needEviction determines if a binding should be evicted from the specified cluster based on taints and tolerations.
// Returns:
//   - needEviction: true if the binding should be evicted (when to evict is indicated by the second return value).
//   - tolerationTime: if needEviction is true, this is the duration to wait before eviction (0 means evict immediately,
//     >0 means evict after this duration).
//     If needEviction is false, a value < 0 indicates the binding should be tolerated indefinitely (no eviction).
//   - error: any error encountered during evaluation.
```

RainbowMango: Actually, this function doesn't have to return `toleration time : -1`, as if the first return value shows if a binding needs to be evicted. Right?

XiShanYongYe-Chang: You are right.

Issue #6641, opened 2025-08-14
[Umbrella] Multi-components workload scheduling - phase I


RainbowMango:
**Summary**:

This issue is used to track the iteration tasks of feature `Multi-components workload scheduling` described at #5115.
See proposal(#6535) for more details.

**Iteration Tasks -- Part-1: accurate quota statistics**:

- [x] Introduce a new feature gate `MultiplePodTemplatesScheduling` for this feature, keep it disabled by default for backward compatibility. (@seanlaii, #6655)
- [x] Introduce a new method `GetComponentReplicas` to the `ResourceInterpreter` interface. (@RainbowMango, #6659)
- [x] Introduce a new interpreter operation. (@RainbowMango, #6668)
- [x] Implement the new method for `ConfigurableInterpreter`. (@seanlaii, #6678)
- [ ] ~~Implement the new method for `CustomizedInterpreter`.~~ (Postpone to release-1.16)
- [x] Implement the new method for `thirdpartyInterpreter`. (@seanlaii, #6687)
- [ ] ~~Implement the new method for `defaultInterpreter`.~~ (Postpone as we might not have to, as no such resource types?)
- [x] Introduce the `components` field to the `ResourceBinding` API. (@RainbowMango, #6649)
- [x] Parse the components via `GetComponentReplicas` interface and store the result in ResourceBinding. (@seanlaii, #6665)
- [x] Implements the customized interpreter for `FlinkDeployment`. (@mszacillo, #6697)
- [x] Enhance the quota statistics. (@liaolecheng, #6701)
- [x] Implement validation logic that enforces the name's presence and uniqueness for the component name. (@zclyne
, #6699)
- [x] enhancement to karmadactl interpret for easier debugging. (@seanlaii, #6696)

**Iteration Tasks -- Part-2: accurate estimate scheduling**:

- [ ] TBD

liaolecheng:
Hi @RainbowMango, I’m interested in this feature. You can assign me some tasks.

RainbowMango:
@liaolecheng Sure thank you in advance. Currently I'm working on the API and the interface change. Before that I believe most of the tasks above will be blocked.
But maybe you can try this one:
> Implements the customized interpreter for FlinkDeployment.

RainbowMango:
> Introduce a new feature gate MultiplePodTemplatesScheduling for this feature, keep it disabled by default for backward compatibility.

[...]

PR #6699, opened 2025-08-26
added validation check for the components section of resourcebinding

zclyne:
**What type of PR is this?**

/kind feature

This PR adds that enforces the name's presence and ensures the uniqueness of the component name when there are multiple components.

Part of #6641 

```release-note
NONE
```

### DIFF
--- a/pkg/apis/work/v1alpha2/binding_types.go
+++ b/pkg/apis/work/v1alpha2/binding_types.go
@@ -229,8 +229,8 @@ type Component struct {
 	// It is required when the resource contains multiple components to ensure proper identification,
 	// and must also be unique within the same resource.
 	// +kubebuilder:validation:MaxLength=32
-	// +optional
-	Name string `json:"name,omitempty"`
+	// +required
+	Name string `json:"name"`
 
 	// Replicas represents the replica number of the resource's component.
 	// +required
--- a/pkg/webhook/resourcebinding/validating.go
+++ b/pkg/webhook/resourcebinding/validating.go
@@ -26,6 +26,7 @@ import (
 	admissionv1 "k8s.io/api/admission/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/util/validation/field"
 	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -72,7 +73,12 @@ func (v *ValidatingAdmission) Handle(ctx context.Context, req admission.Request)
 		return admission.Denied(err.Error())
 	}
 
-	// further validation can be added here if needed
+	if features.FeatureGate.Enabled(features.MultiplePodTemplatesScheduling) {
+		if err := v.validateComponents(rb.Spec.Components, field.NewPath("spec").Child("components")); err != nil {
+			klog.Errorf("Admission denied for ResourceBinding %s/%s: %v", rb.Namespace, rb.Name, err)
+			return admission.Denied(err.Error())
+		}
+	}
 
 	return admission.Allowed("")
 }
@@ -322,6 +328,22 @@ func (v *ValidatingAdmission) processSingleFRQ(frqItem *policyv1alpha1.Federated
 	msg := fmt.Sprintf("Quota check passed for FRQ %s/%s.", frqItem.Namespace, frqItem.Name)
 	klog.V(3).Infof("FRQ %s/%s will be updated. New OverallUsed: %v", frqItem.Namespace, frqItem.Name, potentialNewOverallUsedForThisFRQ)
 	return potentialNewOverallUsedForThisFRQ, msg, nil
+}
+
+// validateComponents checks the validity of the Components field in the ResourceBinding.
+func (v *ValidatingAdmission) validateComponents(components []workv1alpha2.Component, fldPath *field.Path) error {
+	var allErrs field.ErrorList
+	componentNames := make(map[string]struct{})
+	for index, component := range components {
+		if len(component.Name) == 0 {
+			allErrs = append(allErrs, field.Invalid(fldPath.Index(index).Child("name"), component.Name, "component names must be non-empty"))
+		} else if _, exists := componentNames[component.Name]; exists {
+			allErrs = append(allErrs, field.Invalid(fldPath.Index(index).Child("name"), component.Name, "component names must be unique"))
+		} else {
+			componentNames[component.Name] = struct{}{}
+		}
+	}
+	return allErrs.ToAggregate()
 }
 
 func isQuotaRelevantFieldChanged(oldRB, newRB *workv1alpha2.ResourceBinding) bool {
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -21216,12 +21216,14 @@
       "description": "Component represents the requirements for a specific component.",
       "type": "object",
       "required": [
+        "name",
         "replicas"
       ],
       "properties": {
         "name": {
           "description": "Name of this component. It is required when the resource contains multiple components to ensure proper identification, and must also be unique within the same resource.",
-          "type": "string"
+          "type": "string",
+          "default": ""
         },
         "replicaRequirements": {
           "description": "ReplicaRequirements represents the requirements required by each replica for this component.",
--- a/charts/karmada/_crds/bases/work/work.karmada.io_clusterresourcebindings.yaml
+++ b/charts/karmada/_crds/bases/work/work.karmada.io_clusterresourcebindings.yaml
@@ -450,6 +450,7 @@ spec:
                       format: int32
                       type: integer
                   required:
+                  - name
                   - replicas
                   type: object
                 type: array
--- a/charts/karmada/_crds/bases/work/work.karmada.io_resourcebindings.yaml
+++ b/charts/karmada/_crds/bases/work/work.karmada.io_resourcebindings.yaml
@@ -450,6 +450,7 @@ spec:
                       format: int32
                       type: integer
                   required:
+                  - name
                   - replicas
                   type: object
                 type: array
--- a/pkg/generated/openapi/zz_generated.openapi.go
+++ b/pkg/generated/openapi/zz_generated.openapi.go
@@ -7260,6 +7260,7 @@ func schema_pkg_apis_work_v1alpha2_Component(ref common.ReferenceCallback) commo
 					"name": {
 						SchemaProps: spec.SchemaProps{
 							Description: "Name of this component. It is required when the resource contains multiple components to ensure proper identification, and must also be unique within the same resource.",
+							Default:     "",
 							Type:        []string{"string"},
 							Format:      "",
 						},
@@ -7279,7 +7280,7 @@ func schema_pkg_apis_work_v1alpha2_Component(ref common.ReferenceCallback) commo
 						},
 					},
 				},
-				Required: []string{"replicas"},
+				Required: []string{"name", "replicas"},
 			},
 		},
 		Dependencies: []string{
--- a/pkg/webhook/resourcebinding/validating_test.go
+++ b/pkg/webhook/resourcebinding/validating_test.go
@@ -398,40 +398,41 @@ func TestValidatingAdmission_Handle(t *testing.T) {
 	)
 
 	tests := []struct {
-		name               string
-		req                admission.Request
-		decoder            admission.Decoder
-		clientObjects      []client.Object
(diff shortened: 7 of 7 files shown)


zclyne:
A simpler approach I tried was to add the following two markers to the `Components` field:
```
// +kubebuilder:validation:XValidation:rule="size(self) <= 1 || self.all(i, has(i.name) && i.name != '')",message="name is required and must be non-empty when multiple components are specified"
// +kubebuilder:validation:XValidation:rule="size(self) <= 1 || self.all(i, self.exists_one(j, has(j.name) && j.name == i.name ))",message="component names must be unique"
```

However, Kubernetes considered the cost of evaluating the two rules is too high and rejected the request to apply the CRD:
```
Error from server (Invalid): error when creating "/var/folders/nd/gq6b_xss3_q9l3pr_qs7xp1w0000gn/T/tmp.YarUg9lcgd/_crds": CustomResourceDefinition.apiextensions.k8s.io "resourcebindings.work.karmada.io" is invalid: [spec.versions[1].schema.openAPIV3Schema.properties[spec].properties[components].x-kubernetes-validations[1].rule: Forbidden: estimated rule cost exceeds budget by factor of more than 100x (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared), spec.versions[1].schema.openAPIV3Schema.properties[spec].properties[components].x-kubernetes-validations[1].rule: Forbidden: contributed to estimated rule cost total exceeding cost limit for entire OpenAPIv3 schema, spec.versions[1].schema.openAPIV3Schema.properties[spec].properties[components].x-kubernetes-validations[0].rule: Forbidden: contributed to estimated rule cost total exceeding cost limit for entire OpenAPIv3 schema, spec.versions[1].schema.openAPIV3Schema: Forbidden: x-kubernetes-validations estimated rule cost total for entire OpenAPIv3 schema exceeds budget by factor of more than 100x (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared)]
```

Another downside of this approach is that the validation would no longer be controlled by the feature gate.

pkg/webhook/resourcebinding/validating.go
```diff
@@ -72,6 +72,12 @@ func (v *ValidatingAdmission) Handle(ctx context.Context, req admission.Request)
 	}
 
 	// further validation can be added here if needed
```
zhzhuang-zju:
```suggestion
```

pkg/webhook/resourcebinding/validating.go
```diff
@@ -331,6 +337,26 @@ func (v *ValidatingAdmission) processSingleFRQ(frqItem *policyv1alpha1.Federated
 	return potentialNewOverallUsedForThisFRQ, msg, nil
 }
 
+// validateComponents checks the validity of the Components field in the ResourceBinding.
+// It takes ctx, req and oldRB parameters for potential future use but does not use them currently.
+func (v *ValidatingAdmission) validateComponents(_ context.Context, _ admission.Request, rb, _ *workv1alpha2.ResourceBinding) error {
```
zhzhuang-zju: Keeping the parameters of `validateComponents` and `validateFederatedResourceQuota` consistent is for better extensibility in the future. However, I prefer to keep only the currently used parameters and add more later if needed. This approach improves code readability, and besides, the parameters of `validateFederatedResourceQuota` aren't necessarily the complete set. What do you think?

pkg/webhook/resourcebinding/validating.go
```diff
@@ -331,6 +337,26 @@ func (v *ValidatingAdmission) processSingleFRQ(frqItem *policyv1alpha1.Federated
 	return potentialNewOverallUsedForThisFRQ, msg, nil
 }
 
+// validateComponents checks the validity of the Components field in the ResourceBinding.
+// It takes ctx, req and oldRB parameters for potential future use but does not use them currently.
```
zhzhuang-zju:
```suggestion
// 1. checks that if multiple components are specified, each must have a non-empty unique name.
```

pkg/webhook/resourcebinding/validating.go
```diff
@@ -331,6 +337,26 @@ func (v *ValidatingAdmission) processSingleFRQ(frqItem *policyv1alpha1.Federated
...
+func (v *ValidatingAdmission) validateComponents(_ context.Context, _ admission.Request, rb, _ *workv1alpha2.ResourceBinding) error {
+	// TODO: for better maintainability, consider extracting a common interface for all sorts of validations on a certain resource
+	if len(rb.Spec.Components) <= 1 {
```
zhzhuang-zju: Just to confirm, if the `Components` length is 1, is it expected behavior that its name can be empty?

pkg/webhook/resourcebinding/validating.go
seanlaii: +1

pkg/webhook/resourcebinding/validating.go
seanlaii: Yes, looks like it is only required when there are multiple components: https://github.com/karmada-io/karmada/blob/master/pkg/apis/work/v1alpha2/binding_types.go#L228-L233.

zhzhuang-zju: Thanks for the confirmation

pkg/webhook/resourcebinding/validating.go
RainbowMango: +1. 
See example here: https://github.com/karmada-io/karmada/blob/9edc1b49403bab4be8ade955be472f1db925c85c/pkg/util/validation/validation.go#L206-L205

pkg/webhook/resourcebinding/validating.go
RainbowMango: Not sure yet. But for now each component requires a non-empty name. So, please remove this block:
```golang
	// TODO: for better maintainability, consider extracting a common interface for all sorts of validations on a certain resource
	if len(rb.Spec.Components) <= 1 {
		return nil
	}
```

pkg/webhook/resourcebinding/validating.go
RainbowMango: The signature should be `validateComponents(components []workv1alpha2.ComponentRequirements, fldPath *field.Path) error`

pkg/webhook/resourcebinding/validating.go
zclyne: Sure, I will remove this block. However in that case I believe it's better also change the comment that @seanlaii linked above, otherwise it is misleading

RainbowMango: Thanks for the reminder. Another idea is marking the name as `required` for now. 
```diff
--- a/pkg/apis/work/v1alpha2/binding_types.go
+++ b/pkg/apis/work/v1alpha2/binding_types.go
@@ -229,7 +229,7 @@ type ComponentRequirements struct {
        // It is required when the resource contains multiple components to ensure proper identification,
        // and must also be unique within the same resource.
        // +kubebuilder:validation:MaxLength=32
-       // +optional
+       // +required
        Name string `json:"name,omitempty"`

        // Replicas represents the replica number of the resource's component.

```

zclyne: Yes, I think we should mark it as required and probably remove the `omitempty` marker as well

RainbowMango: OK. Let me know if you want to help make that change. Both OK to me to do it in this PR or a separate PR.

zclyne: I can help make the change in this PR!

pkg/webhook/resourcebinding/validating.go
zclyne: Hi @RainbowMango , if the `name` field becomes required no matter how many components there are, the entire validation logic can be simplified to the following two markers:
```
// +listType=map
// +listMapKey=name
```
and we won't need to implement the `validateComponents` function any more. Do you think it looks good to you?

RainbowMango: Just updating the marker `+optional` to `+required` would be more straightforward. As I suggested at https://github.com/karmada-io/karmada/pull/6699#discussion_r2305937492.

This doesn't work?

zhzhuang-zju:
```yaml
// +listType=map
// +listMapKey=name
```

I guess these two markers are meant to ensure that the name of each component is unique. Ref: https://kubernetes.io/docs/reference/using-api/server-side-apply/#merge-strategy

```md
The merging strategy, implemented with Server-Side Apply, provides a generally more stable object lifecycle. Server-Side Apply tries to merge fields based on the actor who manages them instead of overruling based on values. This way multiple actors can update the same object without causing unexpected interference.
```
@zclyne According to the documentation, the marker `listMapKey` is implemented with Server-Side Apply. Is it also available for Client-Side Apply?

zclyne: > Just updating the marker `+optional` to `+required` would be more straightforward. As I suggested at [#6699 (comment)](https://github.com/karmada-io/karmada/pull/6699#discussion_r2305937492).
> 
> This doesn't work?

As @zhzhuang-zju pointed out, the `+required` marker would be directly added to the `name` field to indicate that the `name` field is mandatory. While
```
// +listType=map
// +listMapKey=name
```

will be added to the `components` field to indicate that the list will be treated as a map with the `name` field being the key, so that the uniqueness is gauranteed by kubernetes.

Follow up:
This might not perform validation check when it is not using server-side apply. I'll look into a CEL expression-based validation approach instead.

zclyne: @zhzhuang-zju What do you think about using a CEL expression? I think that would be easier than implementing a go function for it although the downside is still that it would not be controlled by the feature gate

RainbowMango: I'm afraid we don't have enough time to clarify this. So, can we go with the simple way?

zclyne: Do you mean the go code-based approach? Sure I can still go for it and we can revisit this part later!

zhzhuang-zju: Agree. If there are no usage restrictions on `listMapKey`, I believe it would be a good implementation. We can optimize it in the next version.

zhzhuang-zju: Agree. If there are no usage restrictions on `listMapKey`, I believe it would be a good implementation. We can optimize it in the next version.

RainbowMango: > Do you mean the go code-based approach? Sure I can still go for it and we can revisit this part later!

Yeah.

zclyne: Thanks! Just pushed a new commit to include the change. Please have a look when you get a chance and I'll continue addressing comments tomorrow if there's any.

pkg/webhook/resourcebinding/validating.go
```diff
@@ -324,6 +330,59 @@ func (v *ValidatingAdmission) processSingleFRQ(frqItem *policyv1alpha1.Federated
...
+		} else if _, exists := componentNames[component.Name]; exists {
+			allErrs = append(allErrs, field.Invalid(fldPath.Index(index).Child("name"), component.Name, "component names must be unique"))
+		} else {
+			componentNames[component.Name] = struct{}{}
+		}
+	}
+	return allErrs.ToAggregate()
+}
+
+func calculateResourceUsage(rb *workv1alpha2.ResourceBinding) (corev1.ResourceList, error) {
```
RainbowMango: I guess this is added when resolving the conflicts :) Actually, this function has been removed by #6701.

pkg/apis/work/v1alpha2/binding_types.go
```diff
@@ -229,8 +229,8 @@ type ComponentRequirements struct {
 	// It is required when the resource contains multiple components to ensure proper identification,
 	// and must also be unique within the same resource.
 	// +kubebuilder:validation:MaxLength=32
-	// +optional
-	Name string `json:"name,omitempty"`
+	// +required
```
RainbowMango, line 232: You might need to run `make update` to update the auto-generated files after touching the API.

RainbowMango (commented):
Just some nits block the tests. I can help fix it on your branch to move this faster.
I'm glad to see you added so many tests.  Thanks.

RainbowMango (approved):
/lgtm
/approve

pkg/webhook/resourcebinding/validating.go
zclyne: Ah yes. Thanks for pointing it out. It was brought back in when I rebased my branch

zclyne:
Just logged in to fix it and realized that this is merged. Thanks @RainbowMango for working on it!

PR #6702, opened 2025-08-27
Fix the issue that the cache is out of sync within the resource inter…

Mou887:
…preter when using karmadactl promote

**What type of PR is this?**

/kind bug

This is a cherry pick pr .
The release notes of PR #6669 are as follows-
```
`karmadactl`: Fixed the issue of `promote` command that the interpreter cache is out of sync before first use.

```


XiShanYongYe-Chang (commented):
Thanks~
/lgtm
/approve

PR #6743, opened 2025-09-10
Handle the DeletedFinalStateUnknown object in deletion events

XiShanYongYe-Chang:
**What type of PR is this?**

/kind cleanup

In the Kubernetes client-go library, actual deletion events may be missed when object deletion events occur during a disruption of the API server connection. To handle this situation, DeltaFIFO creates a `DeletedFinalStateUnknown` object to mark this deletion event. This object wraps the real deleted object, but we cannot determine if its final state is accurate.

This patch adds checks for `DeletedFinalStateUnknown` objects in all deletion event handling functions, ensuring that the wrapped real object can be correctly extracted for processing. This avoids event handling errors caused by type assertion failures. This is a standard practice in Kubernetes controller development to ensure reliable handling of deletion events.

Fixes #

```release-note
NONE
```


pkg/clusterdiscovery/clusterapi/clusterapi.go
```diff
@@ -112,6 +112,9 @@ func (d *ClusterDetector) OnUpdate(_, newObj interface{}) {
 
 // OnDelete handles object delete event and push the object to queue.
 func (d *ClusterDetector) OnDelete(obj interface{}) {
+	if tombstone, ok := obj.(cache.DeletedFinalStateUnknown); ok {
+		obj = tombstone.Obj
+	}
```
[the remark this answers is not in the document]
XiShanYongYe-Chang: update

RainbowMango:
That's all we missed before?

XiShanYongYe-Chang:
> That's all we missed before?

Yes, after inspecting the entire warehouse, only those areas remain untreated.

pkg/clusterdiscovery/clusterapi/clusterapi.go
```diff
@@ -112,6 +112,13 @@ func (d *ClusterDetector) OnUpdate(_, newObj interface{}) {
 
 // OnDelete handles object delete event and push the object to queue.
 func (d *ClusterDetector) OnDelete(obj interface{}) {
+	if tombstone, ok := obj.(cache.DeletedFinalStateUnknown); ok {
+		obj = tombstone.Obj
+		if obj == nil {
+			klog.V(4).Infof("Received DeletedFinalStateUnknown with nil object, key: %s", tombstone.Key)
```
RainbowMango: Here deserves a warning log, as we are going to miss the deletion event. 
```suggestion
			klog.Warningf("Failed to get object(%s) from tombstone", tombstone.Key)
```
Same as below.

XiShanYongYe-Chang: updated.

PR #6751, opened 2025-09-12
correct the controller-manager long message

zhzhuang-zju:
**What type of PR is this?**
/kind cleanup

Ref to https://github.com/karmada-io/website/pull/888#discussion_r2332606135, the indentation for these lines is incorrect.
Before:
```bash
$ _output/bin/linux/amd64/karmada-controller-manager --help
The karmada-controller-manager runs various controllers.
                The controllers watch Karmada objects and then talk to the underlying
                clusters' API servers to create regular Kubernetes resources.
```
 Now:
```bash
$ _output/bin/linux/amd64/karmada-controller-manager --help
The karmada-controller-manager runs various controllers.
The controllers watch Karmada objects and then talk to the underlying
clusters' API servers to create regular Kubernetes resources.
```

Fixes #

```release-note
None.
```

### DIFF
--- a/cmd/controller-manager/app/controllermanager.go
+++ b/cmd/controller-manager/app/controllermanager.go
@@ -119,9 +119,9 @@ func NewControllerManagerCommand(ctx context.Context) *cobra.Command {
 
 	cmd := &cobra.Command{
 		Use: names.KarmadaControllerManagerComponentName,
-		Long: `The karmada-controller-manager runs various controllers.
-		The controllers watch Karmada objects and then talk to the underlying
-		clusters' API servers to create regular Kubernetes resources.`,
+		Long: "The karmada-controller-manager runs various controllers.\n" +
+			"The controllers watch Karmada objects and then talk to the underlying\n" +
+			"clusters' API servers to create regular Kubernetes resources.",
 
 		PersistentPreRunE: func(_ *cobra.Command, _ []string) error {
 			if err := logsv1.ValidateAndApply(logConfig, features.FeatureGate); err != nil {


zhzhuang-zju:
cc @luyb177

cmd/controller-manager/app/controllermanager.go
```diff
@@ -120,8 +120,8 @@ func NewControllerManagerCommand(ctx context.Context) *cobra.Command {
 	cmd := &cobra.Command{
 		Use: names.KarmadaControllerManagerComponentName,
 		Long: `The karmada-controller-manager runs various controllers.
-		The controllers watch Karmada objects and then talk to the underlying
-		clusters' API servers to create regular Kubernetes resources.`,
+The controllers watch Karmada objects and then talk to the underlying
+clusters' API servers to create regular Kubernetes resources.`,
```
RainbowMango:
```golang
		Long: "The karmada-controller-manager runs various controllers.\n" +
		"The controllers watch Karmada objects and then talk to the underlying\n" +
		"clusters' API servers to create regular Kubernetes resources.",
```
Looks better?

zhzhuang-zju: even better!

RainbowMango:
@zhzhuang-zju Please check the failing tests, it's unrelated, but the test has not been stable recently.

zhzhuang-zju:
> @zhzhuang-zju Please check the failing tests, it's unrelated, but the test has not been stable recently.

Unfortunately, the root cause of the e2e failure can no longer be traced. Let me briefly summarize the failure steps in the e2e test:

1.  Used `member1Client` to create a secret on cluster `member1`.
2.  Waited for this secret to appear on cluster `member1`, which timed out and failed.

We lack the logs from cluster `member1`, so we cannot determine exactly why step 2 kept failing. However, it's somewhat reassuring that this issue appears to be largely unrelated to Karmada.

## Improvement for the future:
When implementing `WaitXXXXPresentOnClusterFitWith`, if the GET request fails, we should print the error log to help narrow down the scope of the investigation.
https://github.com/karmada-io/karmada/blob/4594ca0b4b6fcdf0eb3678a0d93896ddc96aaa9a/test/e2e/framework/secret.go#L62-L69

/retest

RainbowMango:
>When implementing WaitXXXXPresentOnClusterFitWith, if the GET request fails, we should print the error log to help narrow down the scope of the investigation.

+1

Issue #6734, opened 2025-09-05
[Umbrella] Multi-components workload scheduling - phase ​​II​​


RainbowMango:
**Summary**:

This issue is used to track the iteration tasks of feature `Multi-components workload scheduling` described at #5115.
See [proposal](https://github.com/karmada-io/karmada/blob/master/docs/proposals/scheduling/multi-podtemplate-support/multiple-pod-template-support.md) for more details.

**Iteration Tasks -- Part-1: Resource Interpreter Enhancement**:
- [x] Introduce components to [the ResourceInterpreterResponse](https://github.com/karmada-io/karmada/blob/master/pkg/apis/config/v1alpha1/interpretercontext_types.go) API. (@RainbowMango, #6740)
- [x] Implement the [GetComponents()](https://github.com/karmada-io/karmada/blob/b0b0bb482e6a963d10850b8b72723079c83588af/pkg/resourceinterpreter/interpreter.go#L63C2-L63C15) for [webhook interpreter](https://github.com/karmada-io/karmada/tree/master/pkg/resourceinterpreter/customized/webhook).  (@seanlaii, https://github.com/karmada-io/karmada/pull/6745)
- [x] Twist [the example interpreter](https://github.com/karmada-io/karmada/tree/master/examples/customresourceinterpreter) to add the components. (@seanlaii, #6752 )
- [x] [Resource Interpreter documentation](https://karmada.io/docs/next/userguide/globalview/customizing-resource-interpreter) update.  (@seanlaii, https://github.com/karmada-io/website/pull/935)

**Iteration Tasks -- Part-2: accurate estimate scheduling**:
- [x] Introduce a new method to the [ReplicaEstimator interface](https://github.com/karmada-io/karmada/blob/adef1e59748e1e1d31cb75fffe406b5dd69a66d7/pkg/estimator/client/interface.go#L38), focus on defining the input and output for estimator implementations. (@mszacillo, #6765)
- [x] Implements the new `ReplicaEstimator interface` for [GeneralEstimator](https://github.com/karmada-io/karmada/blob/d231a87aed73252a915a80edf48d96030036cb86/pkg/estimator/client/general.go#L39) with resource summary. (@mszacillo, #6812)
- [x] Implements the new `ReplicaEstimator interface` for `GeneralEstimator` with resource models. (@mszacillo, #6912 )
[...]

PR #6752, opened 2025-09-14
Include `InterpretComponent` operation in interpreter-webhook example

seanlaii:
**What type of PR is this?**
/kind documentation

Include `InterpretComponent` operation in [interpreter-webhook example](https://github.com/karmada-io/karmada/tree/master/examples/customresourceinterpreter).

Part of #6734

```release-note
NONE
```


examples/customresourceinterpreter/README.md
```diff
@@ -247,67 +261,14 @@ At this point, you have successfully installed the `karmada-interpreter-webhook-
...
-              clusterNames:
-                - member2
-            weight: 1
-          - targetCluster:
-              clusterNames:
-                - member3
-            weight: 1
+Create the Workload CR:
+```bash
+kubectl --kubeconfig $HOME/.kube/karmada.config --context karmada-apiserver apply -f examples/customresourceinterpreter/workload-sample.yaml
```
seanlaii, line 295: Can directly create the `workload` and `propagationpolicy` with the workload-sample.yaml and workload-propagationpolicy.yaml.

examples/customresourceinterpreter/README.md
```diff
@@ -235,7 +249,7 @@ Please set the correct certificate and add the domain address to the CN field of
 
 In the testing environment of Karmada, this is controlled in script `hack/deploy-karmada.sh`:
 
-https://github.com/karmada-io/karmada/blob/303f2cd24bf5d750c2391bb6699ac89d78b3c43f/hack/deploy-karmada.sh#L155
+https://github.com/karmada-io/karmada/blob/release-1.15/hack/deploy-karmada.sh#L213
```
seanlaii: Update the link with the current version.

examples/customresourceinterpreter/README.md
```diff
@@ -179,18 +150,61 @@ In the example of this article, you can directly run the following script to dep
...
-export interpreter_webhook_example_service_external_ip_address=$(kubectl config view --template='{{range $_, $value := .clusters }}{{if eq $value.name "karmada-apiserver"}}{{$value.cluster.server}}{{end}}{{end}}' | \
-  awk -F/ '{print $3}' | \
-  sed 's/:.*//' | \
-  awk -F. '{printf "%s.%s.%s.8",$1,$2,$3}')
-
-cp -rf "examples/customresourceinterpreter/webhook-configuration.yaml" "${temp_path}/temp.yaml"
-sed -i'' -e "s/{{caBundle}}/${ca_string}/g" -e "s/{{karmada-interpreter-webhook-example-svc-address}}/${interpreter_webhook_example_service_external_ip_address}/g" "${temp_path}/temp.yaml"
-kubectl --kubeconfig $HOME/.kube/karmada.config --context karmada-apiserver apply -f "${temp_path}/temp.yaml"
-rm -rf "${temp_path}"
+set -euo pipefail
```
seanlaii, line 182: Update the script to allow it to be run on macOS. Otherwise, the user on macOS will encounter the `sed` syntax error:
```
ed: -e: No such file or directory error: error parsing /var/folders/_b/j6r7yjl14tbf38cqtbsjclbh0000gn/T/tmp.jopsTKr7lm/temp.yaml: error converting YAML to JSON: yaml: invalid map key: map[interface {}]interface {}{"caBundle":interface {}(nil)}
```

examples/customresourceinterpreter/README.md
```diff
@@ -81,36 +81,7 @@
...
-      clusterNames:
-        - member1
-        - member2
-        - member3
-```
-</details>
-
-```bash
-kubectl --kubeconfig $HOME/.kube/karmada.config --context karmada-apiserver apply -f workload-crd-cpp.yaml
+kubectl --kubeconfig $HOME/.kube/karmada.config --context karmada-apiserver create -f examples/customresourceinterpreter/apis/workload.example.io_workloads.yaml
```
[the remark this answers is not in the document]
seanlaii: @gemini-code-assist But I will encounter `The CustomResourceDefinition "workloads.workload.example.io" is invalid: metadata.annotations: Too long: must have at most 262144 bytes`.

zhzhuang-zju:
Thanks
/assign

examples/customresourceinterpreter/README.md
```diff
@@ -179,18 +150,61 @@ In the example of this article, you can directly run the following script to dep
...
+if [[ ! -f "$TEMPLATE" ]]; then
+  echo "ERROR: Template not found: $TEMPLATE" >&2
+  exit 2
+fi
+if ! command -v kubectl >/dev/null 2>&1; then
+  echo "ERROR: kubectl not found in PATH" >&2
+  exit 2
+fi
+
+TMPFILE="$(mktemp /tmp/interpreter-webhook-config-sample.yaml)"
```
zhzhuang-zju:
```suggestion
TMPFILE="$(mktemp /tmp/interpreter-webhook-config-sample-XXX.yaml)"
```
 ```bash
$ mktemp /tmp/interpreter-webhook-config-sample.yaml
mktemp: too few X's in template ‘/tmp/interpreter-webhook-config-sample.yaml’
```

File: examples/customresourceinterpreter/README.md [hunk 6]
```diff
@@ -81,36 +81,7 @@ EOF
 Install `Workload` CRD in `karmada-apiserver` by running the following command:
 
 ```bash
-kubectl --kubeconfig $HOME/.kube/karmada.config --context karmada-apiserver apply -f examples/customresourceinterpreter/apis/workload.example.io_workloads.yaml
-```
-
-And then, create a `ClusterPropagationPolicy` resource object to propagate `Workload` CRD to all member clusters:
-
-<details>
-
```
zhzhuang-zju: I'm afraid we have to keep `workload-crd-cpp.yaml` to propagate the `Workload` CRD to all member clusters. Otherwise, the scheduler would exclude all member clusters during scheduling, resulting in a "NoClusterFit" error.
```yaml
status:
  conditions:
  - lastTransitionTime: "2025-09-15T06:14:50Z"
    message: '0/3 clusters are available: 3 cluster(s) did not have the API resource.'
    reason: NoClusterFit
    status: "False"
    type: Scheduled
```

examples/customresourceinterpreter/README.md
```diff
@@ -375,4 +336,25 @@ You can check if the `status` field of `Workload` object is aggregated correctly
...
-> Note: If you want to use `Retain`/`InterpretStatus`/`InterpretHealth` function in Pull mode cluster, you need to deploy karmada-interpreter-webhook-example in the Pull mode cluster.
\ No newline at end of file
+> Note: If you want to use `Retain`/`InterpretStatus`/`InterpretHealth` function in Pull mode cluster, you need to deploy karmada-interpreter-webhook-example in the Pull mode cluster.
+
+#### InterpretComponent
+
+To test the `InterpretComponent` operation:
+1. Enable the `MultiplePodTemplatesScheduling` feature gate
+2. Update the webhook configuration to include the `InterpretComponent` operation:
+```yaml
```
zhzhuang-zju: Could these two steps be moved forward, for example, into "Step 2: Deploy webhook configuration in `karmada-apiserver`"?

- First, this would avoid the need to use `kubectl edit` to modify the webhook configuration later.  
- Second, the `Usage` section first deploys the Workload and its corresponding Policy, and then explains the specific behavior of each operation. Therefore, if the webhook configuration is edited at the InterpretComponent operation, the bindings generated from the earlier deployed Workload might not apply the latest webhook configuration.

examples/customresourceinterpreter/README.md
zhzhuang-zju:
```suggestion
https://github.com/karmada-io/karmada/blob/ed1b8f86de6c4e901d0128d5609ded8599916fe5/hack/deploy-karmada.sh#L215
```
From the context, what needs to be conveyed to the user is the location where the certificate's CN field is set, so line 215 might be more direct than line 213. Additionally, maintaining `release-1.15` would increase maintenance overhead.

examples/customresourceinterpreter/webhook/app/workloadwebhook.go
```diff
@@ -81,6 +84,21 @@ func (e *workloadInterpreter) responseWithExploreReplica(workload *workloadv1alp
 	return res
 }
 
+func (e *workloadInterpreter) responseWithExploreComponent(workload *workloadv1alpha1.Workload) interpreter.Response {
+	res := interpreter.Succeeded("")
+	replicas := int32(1)
```
zhzhuang-zju, line 93: Since the `WorkLoad` API specifies that the default value of `workload.Spec.Replicas` is 1, it is perfectly correct for `responseWithExploreComponent` to return 1 when `workload.Spec.Replicas` is `nil`.
https://github.com/karmada-io/karmada/blob/ed1b8f86de6c4e901d0128d5609ded8599916fe5/examples/customresourceinterpreter/apis/workload/v1alpha1/workload_types.go#L43-L47

I think `responseWithExploreReplica` could handle this case in the same way, to ensure consistency between the results of `responseWithExploreReplica` and `responseWithExploreComponent`.

seanlaii: Great point! Updated.

examples/customresourceinterpreter/README.md
seanlaii: Thanks!

examples/customresourceinterpreter/README.md
seanlaii: You are absolutely correct. My bad, I thought it was for Workload CR.

examples/customresourceinterpreter/README.md
seanlaii: Hi @zhzhuang-zju , thank you so much for your detailed review!
For this case, the reason we need to update it here is that the `interpretReplica` will not be called if we specify `InterpretComponent` in the webhook configuration. The interpretComponent will take precedence over `interpretReplica` according to [here](https://github.com/karmada-io/karmada/blob/ed1b8f86de6c4e901d0128d5609ded8599916fe5/pkg/detector/detector.go#L1455-L1482).

zhzhuang-zju: Got it, you have a good point.

How about modifying it like this?
```md
Create the Workload resource and verify the components field is interpreted correctly:

kubectl --kubeconfig $HOME/.kube/karmada.config --context karmada-apiserver apply -f examples/customresourceinterpreter/workload-sample.yaml
```
to
```md
Create a new Workload resource or edit the existing Workload resource to apply the `InterpretComponent` interpretation, and verify the components field is interpreted correctly:

kubectl --kubeconfig $HOME/.kube/karmada.config --context karmada-apiserver edit -f examples/customresourceinterpreter/workload-sample.yaml # For example, modify `workloadSpec.replicas`.
```
> The interpretComponent will take precedence over interpretReplica according to [here](https://github.com/karmada-io/karmada/blob/ed1b8f86de6c4e901d0128d5609ded8599916fe5/pkg/detector/detector.go#L1455-L1482).

It seems like information worth noting, and would be better placed in a Note block.

seanlaii: Sounds good. Thanks for the suggestion!
Updated.

zhzhuang-zju:
/lgtm
@seanlaii Thanks again for your hard work!

zhzhuang-zju:
ask @RainbowMango to take a look~

RainbowMango (approved):
/approve

PR #6759, opened 2025-09-16
fix(hack): Update go install syntax for modern Go versions

fillol:
**What type of PR is this?**

/kind bug
/kind cleanup

The `hack/local-up-karmada.sh` script fails during the setup process on systems with modern Go versions (Go >= 1.16). This failure is caused by the `go install` command inside `hack/setup-dev-base.sh`, which uses a deprecated syntax without a version specifier.

The Go toolchain now requires an explicit version specifier (e.g., `@latest`) for `go install`. This PR updates the command to use the modern syntax, allowing the setup script to complete successfully and improving the developer experience.

Fixes #None

This is a straightforward fix to align the setup script with the project's minimum Go version requirement (currently Go 1.22+), which fully supports this syntax. No side effects are expected on older, unsupported Go versions.

```release-note
NONE
```


zhzhuang-zju:
> The hack/local-up-karmada.sh script fails during the setup process on systems with modern Go versions (Go >= 1.16). 

@fillol Thanks for reporting this issue. Just to confirm, will the current hack/local-up-karmada.sh script fail in your environment? Do you have any logs from when it failed?

fillol:
Hi @zhzhuang-zju, thanks for the quick reply.

Yes, that's correct. The current `hack/local-up-karmada.sh` script fails in my environment. Here is the exact error log I'm getting:

```
go: 'go install' requires a version when current directory is not in a module
        Try 'go install github.com/karmada-io/karmada/cmd/karmadactl@latest' to install the latest version
```

The issue seems to be the old `go install` syntax in the `hack/setup-dev-base.sh` script. My local Go version is 1.24.6, which now requires a version specifier like `@latest` for that command.

Let me know if you need anything else from my side.

zhzhuang-zju:
Could you please run the command `go list -m all | grep karmada` locally? I'd like to see what Go modules you have in your local environment.

fillol:
Hi @zhzhuang-zju,
the command failed at first because of the project's vendoring setup. I had to add the `-mod=readonly` flag to get it to work.

Here is the output you requested, run from the project root:

```bash
github.com/karmada-io/karmada
```

Hope this helps!

zhzhuang-zju:
@fillol Thanks, that helps~  
My environment is the same as yours, with Go version 1.24.6. However, the script `hack/local-up-karmada.sh` has been working well in my environment. Moreover, as you can see, Karmada's CI frequently uses `hack/local-up-karmada.sh` to set up test environments, and it has been working fine so far. So I doubt that this is not a bug, but rather an environment configuration issue.

```bash
$ go help install
If the arguments don't have version suffixes, "go install" may run in
module-aware mode or GOPATH mode, depending on the GO111MODULE environment
variable and the presence of a go.mod file. See 'go help modules' for details.
If module-aware mode is enabled, "go install" runs in the context of the main
module.
```

As shown in the output of `go help install`, it does support commands without version suffixes.

The command:
```bash
GO111MODULE=on go install "github.com/karmada-io/karmada/cmd/karmadactl"
```
is intended to build the `karmadactl` binary from the local source code and install it into the `$GOPATH/bin` directory.

In contrast, the command:
```bash
GO111MODULE=on go install "github.com/karmada-io/karmada/cmd/karmadactl@latest"
```
downloads the latest version (i.e., v1.15.0) of the code to build `karmadactl`, and installs it into `$GOPATH/bin`.

Although the latter works in your environment, I don't recommend making this change. The purpose of the script `hack/local-up-karmada.sh` is to set up a Karmada development environment using **local code**. If it pulls remote code instead, any local modifications won't be reflected or tested.

```bash
go: 'go install' requires a version when current directory is not in a module
        Try 'go install github.com/karmada-io/karmada/cmd/karmadactl@latest' to install the latest version
```

From your error log, it’s also clear that `go install` does support version-less arguments. The real issue is that the current directory is not being recognized as part of a module. Could you possibly check whether your local code is complete? In particular, does the `go.mod` file exist in the project root?

PR #6764, opened 2025-09-18
Change kubectl repo reference (Bitnami deprecation)

vsha96:
**What type of PR is this?**

/kind bug
/kind deprecation

Fixes error: `Error response from daemon: manifest for bitnami/kubectl:latest` when trying to load image for karmada

Old docker repo `bitnami/kubectl` was cleaned
New repo for legacy kubectl is `bitnamilegacy/kubectl`

Please, take a look: https://github.com/bitnami/containers/issues/83267

```release-note
NONE
```


vsha96:
Hello, @chaosi-zju, @zhzhuang-zju! Please take a look, we need to use new legacy repo, and, probably, update some current old released versions

zhzhuang-zju:
thanks
/assign

zhzhuang-zju:
@vsha96 First of all, thank you for reporting this image repository change (strangely, Karmada's CI doesn't seem to have caught this failure). I'm looking into whether there are alternative solutions, such as removing the dependency on `bitnami/kubectl`, since `bitnamilegacy/kubectl` was only intended for temporary migration purposes.

BTW, are you currently using Karmada? We'd love to hear about your use case if you're open to sharing!

RainbowMango:
The tests on master are failing, may be due to [the brownout](https://github.com/bitnami/containers/issues/83267#issuecomment-3296097823):
- https://github.com/karmada-io/karmada/actions/runs/17828172134/job/50686266982
- https://github.com/karmada-io/karmada/actions/runs/17828172134/job/50686266903
- https://github.com/karmada-io/karmada/actions/runs/17828172134/job/50686266923

>  helm install karmada-4o6bv2jsxa charts/karmada --namespace karmada-4o6bv2jsxa --wait --timeout 800s --set components={search,metricsAdapter,descheduler},apiServer.hostNetwork=true
Error: INSTALLATION FAILED: failed pre-install: 1 error occurred:
	* timed out waiting for the condition

RainbowMango:
@vsha96 Thank you for reporting this, given that we can not (for free) get a versioned kubectl image from Bitnami, we should try to replace it with other alternatives.

Just share some draft thoughts here:
1. We can try to replace bitnami/kubectl with the one from Kubernetes(registry.k8s.io/kubectl),or [alpine/kubectl](https://hub.docker.com/r/alpine/kubectl)
2. We should not latest tag, as suggested by gemini above. https://github.com/karmada-io/karmada/pull/6764#discussion_r2358527030

also cc @seanlaii

seanlaii:
Thanks! 
/assign
Kubernetes Official one ([registry.k8s.io/kubectl](https://explore.ggcr.dev/?repo=registry.k8s.io%2Fkubectl)) is good because it follows the upstream release & support policy and is governed by the Kubernetes community. 
However, it is a distroless-style image and doesn't contain `sh` which we use in the jobs. To use this, we might need to maintain our own images or refactor the job commands.

zhzhuang-zju:
> We can try to replace bitnami/kubectl with the one from Kubernetes (registry.k8s.io/kubectl), or [alpine/kubectl](https://hub.docker.com/r/alpine/kubectl).  

> The Kubernetes official image ([registry.k8s.io/kubectl](https://explore.ggcr.dev/?repo=registry.k8s.io%2Fkubectl)) is a good choice because it follows the upstream release and support policy and is governed by the Kubernetes community.  
> However, it is a distroless-style image and does not include `sh`, which we currently use in our jobs. To adopt this image, we might need to maintain our own custom images or refactor the job commands.

@vsha96 @RainbowMango @seanlaii The `registry.k8s.io/kubectl` image indeed won't work for our use case. However, `alpine/kubectl` appears to be a viable alternative, as it includes `/bin/sh`. Although it doesn't have `bash`, we can adapt our command accordingly:

```yaml
          command:
            - /bin/sh
            - -c
            - |
              bash <<'EOF'
              set -ex
              kubectl apply -f /search-apiservice --kubeconfig /etc/kubeconfig
              EOF
```

can be changed to:

```yaml
          command:
            - /bin/sh
            - -c
            - |
              set -ex
              kubectl apply -f /search-apiservice --kubeconfig /etc/kubeconfig
```

I did a quick local verification, and the result is as follows:

```bash
$ kubectl get pods --namespace karmada-system
NAME                                               READY   STATUS    RESTARTS   AGE
etcd-0                                             1/1     Running   0          6m52s
karmada-aggregated-apiserver-69b68dcbbd-9h47f      1/1     Running   0          6m52s
karmada-apiserver-6c74c6f6b4-ctjqt                 1/1     Running   0          6m52s
karmada-controller-manager-d65c74465-htf8b         1/1     Running   0          6m52s
karmada-kube-controller-manager-5c796cbd6c-4cmwh   1/1     Running   0          6m52s
karmada-scheduler-d65b88d9f-2r7g2                  1/1     Running   0          6m52s
karmada-webhook-5549f69d4c-xdrh4                   1/1     Running   0          6m52s
```

FYI: https://github.com/karmada-io/karmada/compare/master...zhzhuang-zju:bitnami?expand=1

RainbowMango:
> alpine/kubectl appears to be a viable alternative, as it includes /bin/sh. Although it doesn't have bash, we can adapt our command accordingly:

Great! I prefer this approach.

> However, it is a distroless-style image and doesn't contain sh which we use in the jobs. To use this, we might need to maintain our own images or refactor the job commands.

This is an option, but it might involve more maintenance effort. In this case, it seems not worth going this way. 

@seanlaii what do you think?

@zhzhuang-zju can you show git diff here? So that @vsha96 can update this PR accordingly. 

Or merge this PR and then migrate by a follow up PR? @vsha96 any thoughts?

zhzhuang-zju:
> @zhzhuang-zju can you show git diff here? So that @vsha96 can update this PR accordingly.

Sure~ here is the git diff 
https://github.com/karmada-io/karmada/compare/master...zhzhuang-zju:karmada:bitnami

seanlaii:
> @seanlaii what do you think?

Sounds good to me! This avoids the maintenance overhead of building our own image.
Just want to double check if [the license of the image](https://github.com/alpine-docker/multi-arch-docker-images/blob/master/LICENSE) (GNU GPLv3) will have any impact or concern? (I don't think it will have any concern since we only reference it as the default image. Just want to verify that. Thanks for your time!)

RainbowMango:
Good point!

GPL is not on [the allowlist of CNCF](https://github.com/cncf/foundation/blob/main/policies-guidance/allowed-third-party-license-policy.md#approved-licenses-for-allowlist).

vsha96:
@zhzhuang-zju @seanlaii @RainbowMango thank you for taking a look!

The solution with [alpine/kubectl](https://hub.docker.com/r/alpine/kubectl) sounds great. I will update the PR a bit later, please give me some time

vsha96:
> BTW, are you currently using Karmada? We'd love to hear about your use case if you're open to sharing!

@zhzhuang-zju I am using it in one of my projects. I will leave some comments later

zhzhuang-zju:
> The solution with [alpine/kubectl](https://hub.docker.com/r/alpine/kubectl) sounds great. I will update the PR a bit later — please give me some time.

@vsha96 Wait a moment — actually, [alpine/kubectl](https://hub.docker.com/r/alpine/kubectl) also doesn’t meet our requirements, because its license is GNU GPLv3, which is **not** on [the CNCF allowlist](https://github.com/cncf/foundation/blob/main/policies-guidance/allowed-third-party-license-policy.md#approved-licenses-for-allowlist).

> @zhzhuang-zju I am using it in one of my projects. I will leave some comments later.

Great~ Looking forward to your reply!

RainbowMango:
@vsha96 Can we use `docker.io/bitnamisecure` temporarily? As the [alpine/kubectl](https://hub.docker.com/r/alpine/kubectl) licensed with GPL which is not allowed by [CNCF license policy](https://github.com/cncf/foundation/blob/main/policies-guidance/allowed-third-party-license-policy.md#cncf-allowlist-license-policy).

RainbowMango:
For the short term, we are still using the `latest` kubectl, all we need to do is migrate the image registry from `docker.io/bitnami` to `docker.io/bitnamisecure`, and then we will not be affected by the removal of docker.io/bitnami on September 29th.

vsha96:
> Can we use docker.io/bitnamisecure temporarily? 

@RainbowMango, I think yes, we can use it as a long-term solution also, since they mention:
-  _"Free images are only available in the latest tag. See our Dockerhub for a list of what’s free"_ (see https://bitnami.com/ -> search for `are bitnami secure images free`)
- _"Only a subset of BSI applications are available for free."_ (see https://hub.docker.com/r/bitnamisecure/kubectl)

I will change to bitnamisecure

PR #6765, opened 2025-09-20
Adding maxAvailableComponentSets to estimator interface

mszacillo:
**What type of PR is this?**
/kind feature

Introduces a new method `MaxAvailableComponentSets` which will return a list of TargetClusters, where the replica # will correspond to the total number of component sets for multiple-component CRDs. 

I decided to add the new method as a part of a separate interface called `ComponentSetEstimator`. This was done because if you add the `MaxAvailableComponentSets`, all downstream implementers of `ReplicaEstimator` will expect an implementation of `MaxAvailableComponentSets`. 

By adding a separate extended interface, other estimators (such as GeneralEstimator) can keep their implementations of ReplicaEstimator, and only estimators that support multi-component workloads will implement ComponentSetEstimator.

Part of #6734

```release-note
Adding maxAvailableComponentSets to estimator interface
```

### DIFF
--- a/pkg/estimator/client/interface.go
+++ b/pkg/estimator/client/interface.go
@@ -34,9 +34,34 @@ var (
 	unschedulableReplicaEstimators = map[string]UnschedulableReplicaEstimator{}
 )
 
-// ReplicaEstimator is an estimator which estimates the maximum replicas that can be applied to the target cluster.
+// ReplicaEstimator estimates how many replicas a workload can run on each target cluster,
+// based on resource availability and node constraints.
 type ReplicaEstimator interface {
+	// MaxAvailableReplicas returns the maximum number of replicas of a single-component workload that each cluster can host.
 	MaxAvailableReplicas(ctx context.Context, clusters []*clusterv1alpha1.Cluster, replicaRequirements *workv1alpha2.ReplicaRequirements) ([]workv1alpha2.TargetCluster, error)
+	// MaxAvailableComponentSets returns the maximum number of complete multi-component sets (in terms of replicas) that each cluster can host.
+	MaxAvailableComponentSets(ctx context.Context, req *ComponentSetEstimationRequest) ([]ComponentSetEstimationResponse, error)
+}
+
+// ComponentSetEstimationRequest carries input parameters for estimating multi-component set availability per cluster.
+// Fields can be extended over time without changing the method signature.
+type ComponentSetEstimationRequest struct {
+	// Clusters represents a list of feasible clusters to estimate against.
+	Clusters []*clusterv1alpha1.Cluster
+	// Components are the components that form a multi-component workload.
+	Components []*workv1alpha2.Component
+	// Namespace is the namespace of the workload being estimated.
+	// It is used by the accurate estimator to check the quota configurations
+	// in the target member cluster. This field is required for quota-aware estimation.
+	Namespace string
+}
+
+// ComponentSetEstimationResponse represents how many complete component sets a cluster can accommodate.
+type ComponentSetEstimationResponse struct {
+	// Name is the cluster name.
+	Name string
+	// Sets is the maximum number of complete component sets that can be scheduled on the cluster.
+	Sets int32
 }
 
 // UnschedulableReplicaEstimator is an estimator which estimates the unschedulable replicas which belong to a specified workload.
--- a/pkg/estimator/client/accurate.go
+++ b/pkg/estimator/client/accurate.go
@@ -67,6 +67,15 @@ func (se *SchedulerEstimator) MaxAvailableReplicas(
 	})
 }
 
+// MaxAvailableComponentSets returns the maximum number of complete multi-component sets (in terms of replicas) that each cluster can host.
+func (se *SchedulerEstimator) MaxAvailableComponentSets(
+	_ context.Context,
+	_ *ComponentSetEstimationRequest) ([]ComponentSetEstimationResponse, error) {
+	// Dummy implementation: return nothing for now
+	// TODO: Implement as part of #6734
+	return nil, nil
+}
+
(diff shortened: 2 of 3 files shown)


mszacillo:
Hi @RainbowMango, I've currently listed the return type as `[]workv1alpha2.TargetCluster`. TargetCluster currently does not have a concept of ComponentSets, just replicas. Perhaps in the implementation of the method, we can change the unit from sets -> replicas? Or would we rather extend TargetCluster to have a ComponentSets number as well?

pkg/estimator/client/interface.go
```diff
@@ -34,11 +34,21 @@ var (
...
 	MaxAvailableReplicas(ctx context.Context, clusters []*clusterv1alpha1.Cluster, replicaRequirements *workv1alpha2.ReplicaRequirements) ([]workv1alpha2.TargetCluster, error)
 }
 
+// ComponentSetEstimator extends ReplicaEstimator to handle multi-component workloads,
+// where all components in a set must be scheduled together.
+type ComponentSetEstimator interface {
+	ReplicaEstimator
+	// MaxAvailableComponentSets returns the maximum number of complete multi-component sets (in terms of replicas) that each cluster can host.
+	MaxAvailableComponentSets(ctx context.Context, clusters []*clusterv1alpha1.Cluster, components []*workv1alpha2.Component) ([]workv1alpha2.TargetCluster, error)
+}
```
RainbowMango: 1. The interface can have its own structure, it doesn't have to rely on the structure defined in the ResourceBinding API. 
2. We do need to think about how to let the ResourceBinding hold the `set` concept more straightforwardly. 
3. I suggest having a struct to hold the request and response, respectively.  That makes room for extending more parameters, like we are likely to have a field like `reserved resources` to tell estimator implementations to reserve some resources before the estimation. 

```suggestion
// ComponentSetEstimator extends ReplicaEstimator to handle multi-component workloads,
// where all components in a set must be scheduled together.
type ComponentSetEstimator interface {
	// MaxAvailableComponentSets returns, for each cluster, the maximum number of complete multi-component sets it can host.
	// The input is provided via a request struct to allow future extensibility, and the result reflects number of sets per cluster.
	MaxAvailableComponentSets(ctx context.Context, req *ComponentSetEstimationRequest) ([]ComponentSetEstimationResponse, error)
}

// ComponentSetEstimationRequest carries input parameters for estimating multi-component set availability per cluster.
// Fields can be extended over time without changing the method signature.
type ComponentSetEstimationRequest struct {
	// Clusters are the candidate clusters to estimate against.
	Clusters []*clusterv1alpha1.Cluster
	// Components are the components that form a multi-component workload.
	Components []*workv1alpha2.Component
	// Namespace is the namespace of the workload being estimated.
	// It is used by the accurate estimator to check the quota configurations
	// in the target member cluster. This field is required for quota-aware estimation.
	Namespace string
}

// ComponentSetEstimationResponse represents how many complete component sets a cluster can accommodate.
type ComponentSetEstimationResponse struct {
	// Name is the cluster name.
	Name string
	// Sets is the maximum number of complete component sets that can be scheduled on the cluster.
	Sets int32
}
```

RainbowMango: In addition, can we put the `MaxAvailableComponentSets` method into the legacy `type ReplicaEstimator interface`?
All we need to do is let the current estimator implementations(`general-estimator` and `scheduler-estimator`) to implement the new method. (Sure, we can have a dummy implementation in this PR.)

mszacillo: > We do need to think about how to let the ResourceBinding hold the set concept more straightforwardly.

To me set can just be an internal concept to the implementers of MaxAvailableComponentSets. At the end of the day, a set is comprised of replicas - why not just convert the unit of sets into replicas after the calculation if complete? 

> I suggest having a struct to hold the request and response, respectively. That makes room for extending more parameters, like we are likely to have a field like reserved resources to tell estimator implementations to reserve some resources before the estimation.

Sounds good! I wasn't sure if this was within the scope of the ticket, but happy to include this. 

> In addition, can we put the MaxAvailableComponentSets method into the legacy type ReplicaEstimator interface?

I had separated the interfaces specifically so I wouldn't need to make dummy implementations haha, but yes I can also do this if it makes this logically easier to follow.

RainbowMango:
> At the end of the day, a set is comprised of replicas - why not just convert the unit of sets into replicas after the calculation if complete?

I think it's doable, at least for now, if you mean the number of replicas for each cluster (just one cluster for multiple components workload) is always `1`. Are we on the same page?

RainbowMango:
> Sounds good! I wasn't sure if this was within the scope of the ticket, but happy to include this.

No, the extended idea(should have another proposal before starting) is not in the scope. Currently we are still focusing on implementing [the proposal](https://github.com/karmada-io/karmada/blob/master/docs/proposals/scheduling/multi-podtemplate-support/multiple-pod-template-support.md).

pkg/estimator/client/interface.go
RainbowMango: Any update?

mszacillo:
> I think it's doable, at least for now, if you mean the number of replicas for each cluster (just one cluster for multiple components workload) is always 1. Are we on the same page?

Something like that would work yes - as long as the maxSet calculation unit is consistent, the scheduler would still be able to make stack-ranked decisions on which cluster has the most available resources.

pkg/estimator/client/interface.go
mszacillo: Updated - I've taken your suggestion to add the `ComponentSetEstimationRequest` to make the inputs more easily extensible. I also added back the `MaxAvailableComponentSets` into the original ReplicaEstimator interface and marked the implementations as TODO.

mszacillo: Let me know if you have further questions / comments.

Issue #6588, opened 2025-07-29
[lfx-mentorship-2025-Sept-Nov] Support RayCluster and RayJob in Karmada via Default Resource Interpreters


RainbowMango:
**CNCF LFX mentorship**
https://github.com/cncf/mentoring/tree/main/programs/lfx-mentorship/2025/03-Sep-Nov

**What would you like to be added**:
Implement default resource interpreters for [RayCluster](https://github.com/ray-project/kuberay) and [RayJob](https://github.com/ray-project/kuberay) workloads in Karmada. This includes interpreter development, validation, and comprehensive documentation on deploying and managing Ray workloads with Karmada. Present the integration and usage guide to the KubeRay community.

**Why is this needed**:
Refer to https://github.com/karmada-io/karmada/issues/5115, and https://github.com/karmada-io/karmada/pull/6535

Devashishbasu:
Hi @RainbowMango 

My name is Devashish basu. I have basic knowledge of Go and Kubernetes, and I’m eager to deepen those skills through active contribution. I was selected for Google Summer of Code 2024, where I worked on an open-source project and learned to collaborate effectively with a community.

I’ve already looked through the repository and would love to start working on “good first issues” right away. Could you please let me know which initial tasks you’d recommend and your preferred way to communicate?

debaa98:
hey @RainbowMango is there any  Coding challenge for mentees ? for this application ?

owenowenisme:
There's actually also another CR from Ray, `RayService`, we should also support that in Karmada.

RainbowMango:
I wasn't aware of this API, and I also prefer not to assign too many tasks to a single LFX program. But I'd be great if we could support it together. Note that it doesn't mean the scope of this LFX program should be extended; we can take the `RayService` as an optional task.
Thanks, Owen, for the reminder.

RainbowMango:
@Devashishbasu @debaa98 Thank you for your interest, but we don't have enough slots for all of you. So you are welcome to apply again in the next term.

Divyansh200102:
Hi @RainbowMango ,
I’m interested in karmada and participating in LFX as part of it in the upcoming term. I was curious if you have any suggestions on how I should get started, and how the selection process works?

PR #6766, opened 2025-09-21
Add custom interpreter for Raycluster

owenowenisme:
**What type of PR is this?**
/kind feature

Part of #6588

```release-note

```

### DIFF
--- a/pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/ray.io/v1/RayCluster/customizations.yaml
+++ b/pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/ray.io/v1/RayCluster/customizations.yaml
@@ -1,0 +1,17 @@
+apiVersion: config.karmada.io/v1alpha1
+kind: ResourceInterpreterCustomization
+metadata:
+  name: declarative-configuration-raycluster
+spec:
+  target:
+    apiVersion: ray.io/v1
+    kind: RayCluster
+  customizations:
+    healthInterpretation:
+      luaScript: >
+        function InterpretHealth(observedObj)
+          if observedObj.status == nil or observedObj.status.state == nil then
+            return false
+          end
+          return observedObj.status.state == 'ready'
+        end
--- a/pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/ray.io/v1/RayCluster/customizations_tests.yaml
+++ b/pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/ray.io/v1/RayCluster/customizations_tests.yaml
@@ -1,0 +1,3 @@
+tests:
+  - observedInputPath: testdata/observed-raycluster.yaml
+    operation: InterpretHealth
--- a/pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/ray.io/v1/RayCluster/testdata/observed-raycluster.yaml
+++ b/pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/ray.io/v1/RayCluster/testdata/observed-raycluster.yaml
@@ -1,0 +1,90 @@
+apiVersion: ray.io/v1
+kind: RayCluster
+metadata:
+  name: sample
+spec:
+  rayVersion: '2.46.0'
+  headGroupSpec:
+    rayStartParams: {}
+    template:
+      spec:
+        containers:
+        - name: ray-head
+          image: rayproject/ray:2.46.0
+          resources:
+            limits:
+              cpu: 1
+              memory: 2G
+            requests:
+              cpu: 1
+              memory: 2G
+          ports:
+          - containerPort: 6379
+            name: gcs-server
+          - containerPort: 8265
+            name: dashboard
+          - containerPort: 10001
+            name: client
+  workerGroupSpecs:
+  - replicas: 1
+    minReplicas: 1
+    maxReplicas: 5
+    groupName: workergroup
+    rayStartParams: {}
+    template:
+      spec:
+        containers:
+          image: rayproject/ray:2.46.0
+          resources:
+            limits:
+              cpu: 1
+              memory: 1G
+            requests:
+              cpu: 1
+              memory: 1G
+status:
+  availableWorkerReplicas: 1
+  conditions:
+  - lastTransitionTime: "2025-09-21T03:55:30Z"
+    message: ""
+    reason: HeadPodRunningAndReady
+    status: "True"
+    type: HeadPodReady
+  - lastTransitionTime: "2025-09-21T03:55:45Z"
+    message: All Ray Pods are ready for the first time
+    reason: AllPodRunningAndReadyFirstTime
+    status: "True"
+    type: RayClusterProvisioned
+  - lastTransitionTime: "2025-09-21T03:54:44Z"
+    message: ""
+    reason: RayClusterSuspended
+    status: "False"
+    type: RayClusterSuspended
+  - lastTransitionTime: "2025-09-21T03:54:44Z"
+    message: ""
+    reason: RayClusterSuspending
+    status: "False"
+    type: RayClusterSuspending
+  desiredCPU: "2"
+  desiredGPU: "0"
+  desiredMemory: 3G
+  desiredTPU: "0"
+  desiredWorkerReplicas: 1
+  endpoints:
+    client: "10001"
+    dashboard: "8265"
+    gcs-server: "6379"
+    metrics: "8080"
+  head:
+    podIP: 10.244.0.6
+    podName: raycluster-kuberay-head-9cvfc
+    serviceIP: 10.244.0.6
+    serviceName: raycluster-kuberay-head-svc
+  lastUpdateTime: "2025-09-21T03:55:45Z"
+  maxWorkerReplicas: 5
+  minWorkerReplicas: 1
+  observedGeneration: 1
+  readyWorkerReplicas: 1
+  state: ready
+  stateTransitionTimes:
+    ready: "2025-09-21T03:55:45Z"


owenowenisme:
@RainbowMango Would you mind reviewing this PR?

RainbowMango:
> @RainbowMango Would you mind reviewing this PR?

Sure, will do by this week. (Not sure if I have enough time before Wednesday, as I'm on a business trip.)

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/ray.io/v1/RayCluster/customizations.yaml
```diff
@@ -0,0 +1,170 @@
...
+          return totalReplicas, requires
+        end
+    healthInterpretation:
+      luaScript: >
+        function InterpretHealth(observedObj)
+          if observedObj.status == nil or observedObj.status.state == nil then
+            return false
+          end
+          return observedObj.status.state == 'ready'
+        end
```
RainbowMango: I see the `State` field has been deprecated and replaced by Conditions.
https://github.com/ray-project/kuberay/blob/e36183d0fc7b09abdf991b7d5e249cfac8ab2cdd/ray-operator/apis/ray/v1/raycluster_types.go#L222-L224

Shall we use the conditions to evaluate the healthy?

If yes, I guess the only condition that could lead to `unhealty` may be `ReplicaFailure == true`.

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/ray.io/v1/RayCluster/customizations.yaml
```diff
@@ -0,0 +1,170 @@
...
+              desiredObj.status.desiredTPU = currentStatus.desiredTPU
+              desiredObj.status.endpoints = currentStatus.endpoints
+              desiredObj.status.head = currentStatus.head
+              desiredObj.status.lastUpdateTime = currentStatus.lastUpdateTime
+              desiredObj.status.maxWorkerReplicas = currentStatus.maxWorkerReplicas
+              desiredObj.status.minWorkerReplicas = currentStatus.minWorkerReplicas
+              desiredObj.status.observedGeneration = currentStatus.observedGeneration
+              desiredObj.status.readyWorkerReplicas = currentStatus.readyWorkerReplicas
+            end
+          end
```
RainbowMango, line 178: It is OK when the length of `statusItems` is exactly `1`. But in case of len(statusItems) >1, the status field will be overwritten by the following item in `statusItems`.

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/ray.io/v1/RayCluster/customizations_tests.yaml
```diff
@@ -0,0 +1,12 @@
...
+    statusInputPath: testdata/status-file.yaml
+    operation: AggregateStatus
+  - desiredInputPath: testdata/desired-raycluster.yaml
+    operation: InterpretReplica
+  - observedInputPath: testdata/observed-raycluster.yaml
+    operation: InterpretHealth
+  - observedInputPath: testdata/observed-raycluster.yaml
+    operation: InterpretStatus
+  - observedInputPath: testdata/observed-raycluster.yaml
+    operation: InterpretComponent
```
RainbowMango: Please add an empty line at the end of this file.

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/ray.io/v1/RayCluster/testdata/desired-raycluster.yaml
```diff
@@ -0,0 +1,44 @@
...
+      spec:
+        containers:
+          image: rayproject/ray:2.46.0
+          resources:
+            limits:
+              cpu: 1
+              memory: 1G
+            requests:
+              cpu: 1
+              memory: 1G
```
RainbowMango: ditto.

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/ray.io/v1/RayCluster/testdata/observed-raycluster.yaml
```diff
@@ -0,0 +1,103 @@
...
+    serviceIP: 10.244.0.6
+    serviceName: sample-head-svc
+  lastUpdateTime: "2025-09-21T03:55:45Z"
+  maxWorkerReplicas: 5
+  minWorkerReplicas: 1
+  observedGeneration: 1
+  readyWorkerReplicas: 1
+  state: ready
+  stateTransitionTimes:
+    ready: "2025-09-21T03:55:45Z"
```
RainbowMango: ditto.

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/ray.io/v1/RayCluster/testdata/status-file.yaml
```diff
@@ -0,0 +1,48 @@
...
+    serviceIP: 10.244.0.6
+    serviceName: sample-head-svc
+  lastUpdateTime: "2025-09-21T03:55:45Z"
+  maxWorkerReplicas: 5
+  minWorkerReplicas: 1
+  observedGeneration: 1
+  readyWorkerReplicas: 1
+  state: ready
+  stateTransitionTimes:
+    ready: "2025-09-21T03:55:45Z"
```
RainbowMango: ditto.

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/ray.io/v1/RayCluster/customizations.yaml
```diff
@@ -0,0 +1,170 @@
...
+                replicas = replicas,
+                replicaRequirements = requires
+              }
+              table.insert(components, wgComponent)
+            end
+          end
+
+          return components
+        end
+    replicaResource:
```
XiShanYongYe-Chang: We currently support multiple template workloads; do we no longer need to implement this hook point?
cc @owenowenisme @RainbowMango

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/ray.io/v1/RayCluster/customizations.yaml
XiShanYongYe-Chang: +1

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/ray.io/v1/RayCluster/customizations.yaml
```diff
@@ -0,0 +1,170 @@
...
+              desiredObj.status.observedGeneration = currentStatus.observedGeneration
+              desiredObj.status.readyWorkerReplicas = currentStatus.readyWorkerReplicas
+            end
+          end
+
+          return desiredObj
+        end
+    statusReflection:
+      luaScript: >
+        function ReflectStatus(observedObj)
```
XiShanYongYe-Chang: It seems like we have copied all the fields from the status, right? If that's the case, does this hook point not need to be implemented, since the default logic is to collect the entire status?

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/ray.io/v1/RayCluster/testdata/status-file.yaml
```diff
@@ -0,0 +1,48 @@
+applied: true
```
XiShanYongYe-Chang, line 1: Can multiple statuses be defined in the current file to cover the status aggregation logic?

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/ray.io/v1/RayCluster/customizations.yaml
owenowenisme: Yeah make sense, updated!

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/ray.io/v1/RayCluster/customizations.yaml
owenowenisme: Updated.

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/ray.io/v1/RayCluster/customizations.yaml
owenowenisme: Removed

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/ray.io/v1/RayCluster/customizations.yaml
owenowenisme: Removed

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/ray.io/v1/RayCluster/testdata/status-file.yaml
owenowenisme: Made the file have 2 member cluster so we can observe result.

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/ray.io/v1/RayCluster/customizations.yaml
```diff
@@ -0,0 +1,214 @@
...
+          local headPodReady = false
+          local clusterProvisioned = false
+          local replicaFailure = false
+          
+          for _, condition in ipairs(observedObj.status.conditions) do
+            if condition.type == 'HeadPodReady' and condition.status == 'True' then
+              headPodReady = true
+            elseif condition.type == 'RayClusterProvisioned' and condition.status == 'True' then
+              clusterProvisioned = true
+            elseif condition.type == 'RayClusterReplicaFailure' and condition.status == 'True' then
```
XiShanYongYe-Chang, line 71: Is the status of the condition "RayClusterReplicaFailure" being true indicating that a failure has occurred?

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/ray.io/v1/RayCluster/customizations.yaml
```diff
@@ -0,0 +1,214 @@
...
+              replicaFailure = true
+            end
+          end
+          
+          return headPodReady and clusterProvisioned and not replicaFailure
```
XiShanYongYe-Chang, line 76: I see that there is a state field in the status. What does this field mean?

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/ray.io/v1/RayCluster/customizations.yaml
```diff
@@ -0,0 +1,214 @@
...
+          -- Aggregate status from all member clusters
+          for i = 1, #statusItems do
+            local currentStatus = statusItems[i].status
+            if currentStatus ~= nil then
+              -- Merge conditions from all clusters with cluster name context
+              if currentStatus.conditions ~= nil then
+                for _, condition in ipairs(currentStatus.conditions) do
+                  -- Add cluster name to identify the source cluster
+                  if statusItems[i].clusterName ~= nil then
+                    condition.clusterName = statusItems[i].clusterName
```
XiShanYongYe-Chang: Does the condition contain the field clusterName?

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/ray.io/v1/RayCluster/customizations.yaml
```diff
@@ -0,0 +1,214 @@
...
+              end
+
+              -- Take the most recent lastUpdateTime
+              if currentStatus.lastUpdateTime ~= nil then
+                if lastUpdateTime == nil or currentStatus.lastUpdateTime > lastUpdateTime then
+                  lastUpdateTime = currentStatus.lastUpdateTime
+                end
+              end
+
+              -- Take the minimum observedGeneration (most conservative)
```
XiShanYongYe-Chang: Is the observedGeneration field collected useless?

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/ray.io/v1/RayCluster/customizations.yaml
owenowenisme: Yes

https://github.com/ray-project/kuberay/blob/b4cd1e279c7fe8cc155f541ed6d315814d724d5c/ray-operator/apis/ray/v1/raycluster_types.go#L331-L332

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/ray.io/v1/RayCluster/customizations.yaml
owenowenisme: Oh that's a deprecate field and is replaced by condition

https://github.com/ray-project/kuberay/blob/b4cd1e279c7fe8cc155f541ed6d315814d724d5c/ray-operator/apis/ray/v1/raycluster_types.go#L245-L247

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/ray.io/v1/RayCluster/customizations.yaml
owenowenisme: No but I think we should indicate which cluster is the condition come from  when there's multiple items so I added this field to condition.

Output will look like this
```yaml
status:
    availableWorkerReplicas: 3
    conditions:
        - clusterName: member1
          lastTransitionTime: "2025-09-21T03:55:30Z"
          message: ""
          reason: HeadPodRunningAndReady
          status: "True"
          type: HeadPodReady
        - clusterName: member1
          lastTransitionTime: "2025-09-21T03:55:45Z"
          message: All Ray Pods are ready for the first time
          reason: AllPodRunningAndReadyFirstTime
          status: "True"
          type: RayClusterProvisioned
...
```

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/ray.io/v1/RayCluster/customizations.yaml
owenowenisme: Yes, removed

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/ray.io/v1/RayCluster/customizations.yaml
XiShanYongYe-Chang: In that case, shouldn't it return false?

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/ray.io/v1/RayCluster/customizations.yaml
XiShanYongYe-Chang: Thanks~

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/ray.io/v1/RayCluster/customizations.yaml
XiShanYongYe-Chang: Thanks, I understand, but this field actually cannot be reflected on the RayCluster resource.

owenowenisme:
> I would like to ask whether it is necessary to process the current resources regarding the dependencyInterpretation hook point.

Normally we don't need to depend on other resource in RayCluster so I think it's okay to not implement that?

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/ray.io/v1/RayCluster/customizations.yaml
owenowenisme: Okay I'll remove clusterName from conditions then, but might be difficult for users to get useful info from this

XiShanYongYe-Chang: Yes, this is also something that the current API cannot solve.

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/ray.io/v1/RayCluster/customizations.yaml
owenowenisme: We return the value with this 
`return headPodReady and clusterProvisioned and not replicaFailure`
So replicaFailure == true indicate this RayCluster isn't healthy

XiShanYongYe-Chang: Oh, thanks. My bad.

XiShanYongYe-Chang:
Hi @owenowenisme can you help squash the commits into one? Then we can merge it.

Issue #6739, opened 2025-09-08
[Umbrella] Enhance Static Weight Feature and Develop Best Practices for Application High Availability


RainbowMango:
**Summary**:
With more users adopting Karmada for multi-cluster application high availability, the static weight feature has become a commonly used mechanism. However, certain corner cases still need resolution (e.g., issue #6424, #6511, proposal #6520). This initiative aims to:

1. Review the current maturity of the static weight feature
2. Address existing gaps and limitations
3. Create comprehensive best practice documentation to help users implement this feature effectively

**Iteration Tasks**:
- [x] Review current static weight design: identify all known issues and limitations.
- [x] Fix replicas assignments not incrementally issue as tracked by #6735
  - [x] Setup the base tests before refactor the dispenser. (@RainbowMango, #6757)
  - [x] Introduce Webster method queue. (@RainbowMango, #6772)
  - [x] Introduce a new method to Dispenser(prepare for migrating static weight) (@zhzhuang-zju, #6793)
  - [x] Migrate static weight assignment (@zhzhuang-zju, #6793)
  - [x] Migrate dynamic weight assignment (@zhzhuang-zju, #6837)
  - [x] Migrate aggregated assignment (@zhzhuang-zju, #6837)
- [x] (Documentation) Create practical guides showing how to configure static weights for different high availability scenarios properly. (@RainbowMango, https://github.com/karmada-io/website/pull/951)
- Pending features and proposals: Ensure each cluster has at least one replica assigned
  - Issues: #6511, #6424 
  - Proposal: #6520
  - Reason why postponed: https://github.com/karmada-io/karmada/issues/6511#issuecomment-3561744184

zhzhuang-zju:
> Migrate aggregated assignment(consider if the webster method is the best choice for this case)

I think applying the Webster method to the aggregated assignment is feasible. Aggregated assignment—i.e., distributing replicas among as few clusters as possible—is ensured by selecting the minimal set of clusters based on their available resources during cluster selection. In practice, this has little to do with the subsequent replica assignment.

zhzhuang-zju:
@RainbowMango The task `Migrate aggregated assignment (consider if the Webster method is the best choice for this case)` has actually been completed.

RainbowMango:
updated

Issue #6735, opened 2025-09-06
StaticWeightList/TakeByWeight may reduce replicas in some clusters when total replicas increase


akakakakakaa:
### Description

When distributing replicas across clusters using `[TakeByWeight](https://github.com/karmada-io/karmada/blob/b0b0bb482e6a963d10850b8b72723079c83588af/pkg/util/helper/binding.go#L112)`, the rounding and remainder handling logic can cause temporary inconsistencies in the total number of active replicas.

Specifically, after rounding down fractional allocations (`int32` cast), the remaining replicas are redistributed sequentially by incrementing each cluster’s replica count. However, this can lead to scenarios where one cluster’s replicas are reduced compared to the previous iteration, causing the overall system to run with fewer replicas than intended until all clusters catch up.

This is particularly problematic for us because we deploy **LLM servers across multiple regions**, and **a new server can take up to \~10 minutes to become ready**. During that time, the system may run noticeably below the requested replica count, affecting availability.

### Weights

The following examples use our production weights:

* `southcentralus = 170`
* `westus = 110`
* `eastus2 = 60`
  (sum = 340)

### Example

* **Total replicas = 5**

  * Ideal (weight/sum × total):

    * `southcentralus` = 5 × 170/340 = **2.5** → 2
    * `westus`        = 5 × 110/340 = **1.6176…** → 1
    * `eastus2`       = 5 × 60/340  = **0.8823…** → 0
  * After remainder pass (current logic: +1 in order): **3, 2, 0**

* **Total replicas = 6**

  * Ideal:

    * `southcentralus` = 6 × 170/340 = **3.0** → 3
    * `westus`        = 6 × 110/340 = **1.9411…** → 1
    * `eastus2`       = 6 × 60/340  = **1.0588…** → 1
  * After remainder pass: **4, 1, 1**
  * **Issue:** Compared to the 5-replica case, `westus` drops from **2 → 1**. Given \~10 minutes startup for new LLM servers, we effectively run at only **4 active replicas** for a noticeable period.

### Expected Behavior

* When the global total increases, **no cluster’s replica count should decrease**

RainbowMango:
> When the global total increases, no cluster’s replica count should decrease

I totally agree. During scale-up, the replica on the existing cluster should never be reduced. 

I will try to reproduce it on my side.

@akakakakakaa by the way, do you have an idea how to fix it?

akakakakakaa:
[...]

PR #6772, opened 2025-09-22
Introduce webster method for refactor dispenser

RainbowMango:
**What type of PR is this?**

/kind cleanup

This PR introduces a WebsterPriorityQueue, which implements [the Webster method, also called the Sainte-Laguë method](https://en.wikipedia.org/wiki/Sainte-Lagu%C3%AB_method), the queue will be used by dividing replicas among clusters according to static/dynamic weight. 

PS: The rest of the tasks are tracked by #6739. 

Part of #6739 
Original issue #6735

This PR implements **Webster’s method** (Sainte-Laguë) for proportional replica scheduling. Unlike D’Hondt, which favors larger clusters, Webster’s approach minimizes seat-to-vote ratio deviations and ensures smaller clusters receive replicas sooner. This is critical for:  
1. **High Availability (HA)**: Faster distribution across clusters, even with low replica counts.  
2. **HPA Activation**: Guarantees ≥1 replica per cluster to trigger autoscaling immediately.  

D’Hondt was rejected because its large-cluster bias delays HA readiness for small clusters and risks leaving clusters replica-less (breaking HPA). Webster aligns better with Karmada’s goals of fairness and resilience.

```release-note
NONE
```

### DIFF
--- a/pkg/util/helper/webstermethod.go
+++ b/pkg/util/helper/webstermethod.go
@@ -1,0 +1,161 @@
+/*
+Copyright 2025 The Karmada Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package helper
+
+import (
+	"container/heap"
+	"sort"
+)
+
+// Party represents a party in the Webster apportionment method.
+type Party struct {
+	// Name is the name of the party.
+	Name string
+	// Votes is the number of votes the party received.
+	Votes int64
+	// Seats is the number of seats assigned to the party.
+	Seats int32
+}
+
+// WebsterPriorityQueue implements heap.Interface using the Webster (Sainte-Laguë) method.
+// The party with the highest Webster priority (Votes/(2*Seats+1)) is at the top.
+type WebsterPriorityQueue struct {
+	// Parties contains all the parties participating in the Webster apportionment method.
+	// Each party has a name, vote count, and current seat allocation.
+	Parties []Party
+
+	// TieBreaker is an optional function used to break ties when two parties have
+	// the same Webster priority. If nil, ties are broken by vote count, then seat count,
+	// then party name to ensure a deterministic result.
+	TieBreaker func(a Party, b Party) bool
+}
+
+// Check if our WebsterPriorityQueue implements necessary interfaces
+var _ heap.Interface = &WebsterPriorityQueue{}
+
+// Len returns the number of Parties in the queue.
+func (pq *WebsterPriorityQueue) Len() int {
+	return len(pq.Parties)
+}
+
+// Less compares two Parties by their Webster priority.
+// The party with the higher priority comes first.
+func (pq *WebsterPriorityQueue) Less(i, j int) bool {
+	// In the Webster method, compare the priority of two Parties:
+	// the one with the higher value of Votes/(2*Seats+1) gets the next seat.
+	iPriority := float64(pq.Parties[i].Votes) / (float64(2*pq.Parties[i].Seats + 1))
+	jPriority := float64(pq.Parties[j].Votes) / (float64(2*pq.Parties[j].Seats + 1))
+	if iPriority == jPriority {
+		if pq.TieBreaker != nil {
+			return pq.TieBreaker(pq.Parties[i], pq.Parties[j])
+		}
+
+		// The party with fewer seats wins the tie.
+		// This is because the Webster method inherently favors smaller parties (or, in Karmada's context, smaller clusters).
+		// In Karmada, this method is often used to assign replicas based on static or dynamic weights.
+		// The rationale is that assigning to smaller clusters earlier helps establish High Availability (HA) and enables
+		// Horizontal Pod Autoscaling (HPA) more effectively. By preferring parties (clusters) with fewer seats (replicas),
+		// we promote a more balanced and resilient distribution, which is desirable for HA scenarios.
+		if pq.Parties[i].Seats != pq.Parties[j].Seats {
+			return pq.Parties[i].Seats < pq.Parties[j].Seats
+		}
+
+		// If both parties have the same priority and the same number of seats,
+		// that means they also have the same number of votes. So we don't have to compare the votes now.
+		// The party with the lexicographically smaller name wins the tie.
+		// For example, party 'a' wins over party 'b' because 'a' < 'b' lexicographically.
+		return pq.Parties[i].Name < pq.Parties[j].Name
+	}
+
+	return iPriority > jPriority
+}
+
+// Swap swaps two Parties in the queue.
+func (pq *WebsterPriorityQueue) Swap(i, j int) {
+	pq.Parties[i], pq.Parties[j] = pq.Parties[j], pq.Parties[i]
+}
+
+// Push adds a new party to the queue.
+func (pq *WebsterPriorityQueue) Push(x interface{}) {
+	pq.Parties = append(pq.Parties, x.(Party))
+}
+
+// Pop removes and returns the party with the highest priority.
+func (pq *WebsterPriorityQueue) Pop() interface{} {
+	old := pq.Parties
+	n := len(old)
+	item := old[n-1]
+	pq.Parties = old[0 : n-1]
+	return item
+}
+
+// AllocateWebsterSeats allocates new seats using the Webster method.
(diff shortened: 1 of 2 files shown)


pkg/util/helper/webstermethod.go
```diff
@@ -0,0 +1,164 @@
...
+	return len(pq.Parties)
+}
+
+// Less compares two Parties by their Webster priority.
+// The party with the higher priority comes first.
+func (pq *WebsterPriorityQueue) Less(i, j int) bool {
+	// In the Webster method, compare the priority of two Parties:
+	// the one with the higher value of Votes/(2*Seats+1) gets the next seat.
+	iPriority := pq.Parties[i].Votes / (int64(2*pq.Parties[i].Seats + 1))
+	jPriority := pq.Parties[j].Votes / (int64(2*pq.Parties[j].Seats + 1))
```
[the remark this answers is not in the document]
RainbowMango: @Copilot, please tell me the details of edge cases?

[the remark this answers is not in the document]
RainbowMango: Fixed as suggested.

pkg/util/helper/webstermethod_test.go
```diff
@@ -0,0 +1,327 @@
...
+	}
+	totalSeats := int32(4)
+	result := AllocateWebsterSeats(totalSeats, partiesVotes, nil, nil)
+	for _, p := range result {
+		fmt.Printf("%s: %d seats\n", p.Name, p.Seats)
+	}
+	// Output:
+	// Alpha: 2 seats
+	// Beta: 1 seats
+	// Gamma: 1 seats
```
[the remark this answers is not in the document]
RainbowMango: Fixed.

pkg/util/helper/webstermethod.go
```diff
@@ -0,0 +1,164 @@
...
+			}
+		}
+
+		// If a party is present in partyVotes but not in initialAssignments,
+		// it will be initialized with 0 seats and its vote count, and will participate
+		// in the allocation for new seats.
+		if !found {
+			pq.Parties = append(pq.Parties, Party{Name: n, Votes: v, Seats: 0})
+		}
+	}
```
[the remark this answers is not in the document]
zhzhuang-zju: +1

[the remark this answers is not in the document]
RainbowMango: Updated.

zhzhuang-zju:
thanks
/lgtm

RainbowMango:
cc @whitewindmills take another look?

pkg/util/helper/webstermethod.go
```diff
@@ -0,0 +1,159 @@
...
+		if p, ok := partiesMap[n]; ok {
+			p.Votes = v
+			partiesMap[n] = p
+		} else {
+			partiesMap[n] = Party{Name: n, Votes: v, Seats: 0}
+		}
+	}
+
+	pq := WebsterPriorityQueue{
+		Parties:    make([]Party, 0, len(initialAssignments)+len(partyVotes)),
```
XiShanYongYe-Chang: At this point, the length of partiesMap has already been determined. Would it be better to use `len(partiesMap)`?

pkg/util/helper/webstermethod_test.go
```diff
@@ -0,0 +1,327 @@
...
+
+package helper
+
+import (
+	"fmt"
+	"reflect"
+	"testing"
+)
+
+func ExampleAllocateWebsterSeats() {
```
XiShanYongYe-Chang, line 25: What is the purpose of maintaining this test separately?

pkg/util/helper/webstermethod.go
RainbowMango: Good point!
Updated as suggested.

pkg/util/helper/webstermethod_test.go
RainbowMango: Demonstrate the usage of a queue. The `Example` case is very common in the Go std lib.

pkg/util/helper/webstermethod.go
```diff
@@ -0,0 +1,159 @@
...
+
+		// The party with more votes wins the tie.
+		if pq.Parties[i].Votes != pq.Parties[j].Votes {
+			return pq.Parties[i].Votes > pq.Parties[j].Votes
+		}
+
+		// The party with fewer seats wins the tie.
+		if pq.Parties[i].Seats != pq.Parties[j].Seats {
+			return pq.Parties[i].Seats < pq.Parties[j].Seats
+		}
```
XiShanYongYe-Chang: If we prefer small parties, is it better to compare Seats first and then Votes, so that small parties that do not get seats will get seats first?

pkg/util/helper/webstermethod_test.go
XiShanYongYe-Chang: Got it, thanks for the explanation~

pkg/util/helper/webstermethod.go
RainbowMango: I don't think so. The way to calculate the priority already tends to small parties. If both parties have the same priority, big party should win for fairness.

XiShanYongYe-Chang: Regardless of the choice, there should be no issue of unfairness. 

Personally, I just feel that from an algorithmic perspective, choosing to optimize for smaller parties, it is also appropriate to continue prioritizing smaller parties at this point. This way, when there is only one seat left, it can be awarded to a smaller party that has not yet secured a seat, thereby increasing the number of participating parties. In the context of multiple clusters, this can lead to more clusters participating in the splitting of replica numbers.

This is merely a personal reference opinion and does not oppose the current actions.

RainbowMango: Let's say we have two parties:
- PartyA: got votes 300
- PartyB:  got votes 100

they run for 2 seats:
- fist round, Party A would get the seat.
- second round
  * Priority of Party A: `300/(2*1 +1)` = `100`
  * Priority of Party B: `100/(0*1 +1)` = `100`

If we favor the smaller parties, then each of the two parties will obtain one seat.
It should be noted that the number of votes received by Party A is three times that of Party B.
That's an example I feel is not fair for Party A.

PS: If PartyA got 200 votes, I feel ok if I assign the second seat to a small party.

XiShanYongYe-Chang: > If we favor the smaller parties, then each of the two parties will obtain one seat.
It should be noted that the number of votes received by Party A is three times that of Party B.
That's an example I feel is not fair for Party A.

I think this reflects the prioritization of smaller parties. When the priority between large and small parties is equal, it allows smaller parties to participate as much as possible (since larger parties have already secured their seats).

> PS: If PartyA got 200 votes, I feel ok if I assign the second seat to a small party.

According to the current algorithm, this is undoubtedly the correct result.

zhzhuang-zju: This is a really interesting discussion! I asked several AI assistants, but their answers varied. Whether to prioritize comparing seats or votes when quotients are equal seems more like a matter of design preference than a strict correctness issue.

However, I came across a compelling perspective: the goal of seat allocation is to make the "votes per seat" as equal as possible across parties.  
For example, if there are 3 total seats and two parties have votes in a 2:1 ratio, assigning seats in a 2:1 split ensures both parties end up with exactly 1 vote per seat.

From this viewpoint, prioritizing fewer seats in tie-breaking better aligns with the principle of proportional fairness.

Consider two parties:

- Party A: votes = v₁, seats = s₁  
- Party B: votes = v₂, seats = s₂  
- with v₁ > v₂

If their Sainte-Laguë quotients are equal:  
v₁ · (2s₂ + 1) = v₂ · (2s₁ + 1)

Dividing both sides by v₁v₂ gives:  
2·(s₂/v₂) + 1/v₂ = 2·(s₁/v₁) + 1/v₁

Which implies:  
s₂/v₂ < s₁/v₁ → v₂/s₂ > v₁/s₁

This means Party B currently has a higher votes-per-seat ratio. To bring the two ratios closer together, we should allocate the next seat to Party B (i.e., increase s₂).

XiShanYongYe-Chang: > Whether to prioritize comparing seats or votes when quotients are equal seems more like a matter of design preference than a strict correctness issue.

Yes, it's not a discussion about correctness, but rather a discussion at the algorithmic level.

RainbowMango: So, both of you are suggesting breaking the tie with the seat number, right?

Note that, if two parties have the same priority, we only need to compare their votes or seats, because if their votes are equal, their seats must also be equal, and vice versa.

zhzhuang-zju: > Note that, if two parties have the same priority, we only need to compare their votes or seats, because if their votes are equal, their seats must also be equal, and vice versa.

Oh, right! So in practice, we only need to choose either `vote` or `seat` in the tie breaker, with `name` serving as the final tiebreaker.

RainbowMango: @whitewindmills What's your opinion?

RainbowMango: Force pushed for tending to small parties.

PR #6779, opened 2025-09-24
Fix: Ignore historical assigned replicas when clusters no longer meet updated scheduling constraints.

### LINKED ISSUE #6771: Inaccurate Cluster MaxAvailableReplicas Calculation When Scheduling Constraints Change Due to Inclusion of Historically Assigned Replicas
**What happened**:
When the scheduling constraints of a workload (e.g., nodeAffinity, resourceRequests) are modified, rendering previously assigned clusters unsuitable, the Karmada scheduler may still prioritize these clusters during scaling operations. This occurs because the current logic for calculating a cluster's "available replicas" incorporates historically assigned replica counts (rbSpec.AssignedReplicasForCluster). For example, if a Deployment initially had 10 replicas on Cluster A, and a constraint change now makes Cluster A unsuitable, the scheduler might still calculate Cluster A's total capacity as 10 (assigned) + 3 (newly estimated) = 13. This can lead to suboptimal scheduling decisions, such as assigning new replicas to Cluster A instead of more suitable clusters like Cluster B, which might have higher actual capacity under the new constraints. This behavior is particularly impactful when using the Aggregated replica scheduling strategy, which favors clusters with higher total perceived capacity .

When scheduling constraints are updated, the scheduler should primarily rely on the current estimated available replicas (from calAvailableReplicasFunc) under the new constraints, not historical assignment data. Clusters that no longer meet the new constraints should have their historically assigned replicas disregarded or heavily discounted in capacity calculations. The scheduler should prioritize clusters that best satisfy the updated constraints for new replica assignments, ensuring optimal placement and avoiding placement on incompatible clusters.

1. Deploy a Deployment with 10 replicas using a PropagationPolicy with replicaDivisionPreference: Aggregated. Ensure it schedules all replicas to Cluster A (e.g., using clusterAffinity or labels matching Cluster A).
2. Modify the Pod template's nodeAffinity (or other constraints like resource.requests) such that Cluster A no longer satisfies the new constraints (e.g., require a label only present on Cluster B).
3. Scale the Deployment to 15 replicas.
4. Observe scheduler behavior: instead of prioritizing Cluster B (which now satisfies the constraints), the scheduler may still assign additional replicas to Cluster A due to the inflated "available replica" count from historical assignments.

- The problem might be more pronounced when using estimators like the "accurate estimator" , if historical data influences their calculations.

**Environment**:
Karmada version: v1.10 (likely affects versions incorporating the historical replica assignment logic)
kubectl-karmada or karmadactl version: NA
Others: The issue might be more observable when using accurate estimation.

rohan-019:
… constraints

**What type of PR is this?**
/kind bug

This PR fixes an issue in the Karmada scheduler where historical replica assignments (AssignedReplicasForCluster) were unconditionally added to a cluster’s available replicas, even when the cluster no longer satisfies the workload’s updated scheduling constraints (e.g., nodeAffinity, resourceRequests).

With the previous logic, clusters that became unsuitable after constraint changes could appear more attractive due to inflated available replicas, leading to suboptimal scheduling decisions.

This PR ensures that historical assignments are only counted for clusters that still meet the current constraints, so new replicas are scheduled to clusters that can actually accommodate them.

Fixes #6771

Modified pkg/scheduler/core/spreadconstraint/group_clusters.go to conditionally include historical assignments.

Verified behavior with unit tests and scaling scenarios under constraint changes.

**Does this PR introduce a user-facing change?**: NO

```release-note
karmada-scheduler: Fixed a scheduling bug where historical assigned replicas were incorrectly counted for clusters that no longer meet updated scheduling constraints, preventing suboptimal replica placement.

```


rohan-019:
Hey, @GitHubxsy @zhzhuang-zju
please review!

GitHubxsy:
> Hey, @GitHubxsy @zhzhuang-zju please review!

@rohan-019   I apologize for the delay in seeing your PR. Your PR can only solve part of the scenario, such as when affinity rules are modified and the cluster has no nodes that meet the new affinity requirements. However, if the user modifies the resource requests (e.g., CPU/memory requests for the Pods), the current calculated maximum available replicas will still be incorrect.
My idea is to remove the existing logic from the karmada-scheduler and delegate the responsibility of providing an accurate maximum available replica count entirely to the karmada-scheduler-estimator.
​How the karmada-scheduler-estimator can accurately calculate the maximum available replicas:​​
The resource view snapshot maintained by the karmada-scheduler-estimator should subtract the resources already consumed by the Pods of this workload that are currently deployed. Perhaps we need to add a new selector parameter. This parameter would be used to match the Pods belonging to this specific workload, allowing us to accurately account for the resources they are already occupying within the resource view snapshot. Then, this updated resource view would be used to calculate the true maximum available replicas.

Issue #6780, opened 2025-09-24
[Umbrella] Reduce unnecessary resource watch events in the karmada controllers


XiShanYongYe-Chang:
The `karmada-controller-manager` is a composite of multiple controllers, each of which watches for changes in target resources and performs reconciliation to align the actual state with the desired state. 

Currently, the event watching and filtering operations of multiple controllers on target resources are somewhat redundant to a certain extent. If we can eliminate these redundant events or resource changes, it would reduce unnecessary reconciliations by the controllers, thereby enhancing the performance of the controllers to some degree.

The following lists the controllers that can be optimized:

- [ ] **binding-controller/cluster-resource-binding-controller**: There is no need to monitor the deletion events of the ResourceBinding object. Additionally, fine-grained filtering requires attention to the spec field:
     - `.spec.requiredBy`
     - `.spec.clusters`
     - `.spec.gracefulEvictionTasks`
     - `.spec.resource`
     - `.spec.suspension`
     - `.spec.preserveResourcesOnDeletion`
     - `.spec.conflictResolution`
- [ ] **cert-rotation-controller**: Only need to watch the creation event of the cluster object, as the cluster object will be processed periodically, so there is no need to watch the update or deletion events of the cluster.
- [ ] **cluster-controller**: Only need to watch the creation events of the cluster object, as well as the updates and deletion timestamp changes of the Ready Status Condition (currently watching all events and changes in spec, status, and metadata).
- [ ] **taint-manager**: Only need to watch the creation events of the cluster and the updates to taints (currently watching all events as well as changes in spec, status, and metadata).
- [ ] **execution-controller**: There is no need to watch the deletion event of the Work object, as the work has a finalizer, and the deletion process will handle it by checking whether the deletion timestamp is empty.
- [ ] **federated-resource-quota-sync-controller**: Some todo items have been identified by @zhzhuang-zju and can be optimized according to the plan.
- [ ] **resource-binding-graceful-eviction-controller/cluster-resource-binding-graceful-eviction-controller**: For update events, maintain the existing logic and only monitor changes to the gracefulEvictionTasks field.
- [ ] **service-export-controller**: It is necessary to watch the creation event of the work object, but not for the update event. The judgment of whether the work object contains the ServiceExport resource can be moved to the predicate logic.
- [ ] **endpointslice-controller**: There is no need to watch the deletion event of the work object.
[...]

PR #6784, opened 2025-09-25
Optimize binding controllers with fine-grained event filtering to reduce redundant reconciliations

rohan-019:
# Optimize binding controllers with custom event predicate

## Summary

Introduces a custom predicate for binding controllers to filter events and reduce unnecessary reconciliations:

- Ignores delete events (cleanup handled by finalizers)  
- Always processes create events  
- Processes updates only for meaningful changes (generation, DeletionTimestamp, key `.spec` fields)

**Result:** Better performance with no behavior changes.

## Motivation

Current controllers watch resources with broad predicates, causing redundant reconciliations. Filtering irrelevant events reduces CPU/memory usage, API calls, and cache churn while preserving finalizer cleanup.

## Changes

- **Added:** `ResourceBindingPredicate`  
- **Updated:** Binding controllers to use the new predicate  
- **Added:** Unit tests  

### Predicate Logic

- ✅ **Create:** Always processed  
- ❌ **Delete:** Ignored  
- ⚠️ **Update:** Only if generation changes, DeletionTimestamp is set, or critical `.spec` fields change (`requiredBy`, `clusters`, `resource`, `gracefulEvictionTasks`, `suspension`, `preserveResourcesOnDeletion`, `conflictResolution`)

## Event Behavior

| Event Type           | Before      | After       | Reason                       |
|---------------------|------------|------------|------------------------------|
| Create              | ✅ Process | ✅ Process | Always needed                |
| Delete              | ✅ Process | ❌ Ignore  | Finalizers handle cleanup    |
| Update (generation) | ✅ Process | ✅ Process | Meaningful change            |
| Update (status only)| ✅ Process | ❌ Ignore  | No reconciliation needed     |

## Files Changed

**Added:**  
- `pkg/controllers/binding/binding_predicate.go`  
- `pkg/controllers/binding/binding_predicate_test.go`  

**Modified:**  
- `pkg/controllers/binding/binding_controller.go`  
- `pkg/controllers/binding/cluster_resource_binding_controller.go`  

## Testing

- Unit tests cover all event types  
- Delete events ignored as expected  
- Critical `.spec` changes trigger reconciliation  
- Finalizer cleanup verified via DeletionTimestamp  

## Backward Compatibility

- No API changes  
- Finalizer cleanup preserved  
- Override policy watch behavior unchanged  

## Performance Impact

- Fewer reconciliations  
- Reduced API calls and cache churn  
- Lower CPU/memory usage  

## Linked Issue

Part of: #6780

## Checklist

- [x] Unit tests added/passing  
- [x] No breaking changes  
- [x] Predicate applied to both RB and CRB controllers  
- [x] Finalizer handling verified via DeletionTimestamp path

### DIFF
--- a/pkg/controllers/binding/binding_controller.go
+++ b/pkg/controllers/binding/binding_controller.go
@@ -31,6 +31,7 @@ import (
 	"k8s.io/klog/v2"
 	"k8s.io/utils/ptr"
 	controllerruntime "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
@@ -187,7 +188,7 @@ func (c *ResourceBindingController) checkDirectPurgeOrphanWorks(ctx context.Cont
 func (c *ResourceBindingController) SetupWithManager(mgr controllerruntime.Manager) error {
 	return controllerruntime.NewControllerManagedBy(mgr).
 		Named(ControllerName).
-		For(&workv1alpha2.ResourceBinding{}).
+		For(&workv1alpha2.ResourceBinding{}, builder.WithPredicates(newBindingPredicate())).
 		WithEventFilter(predicate.GenerationChangedPredicate{}).
 		Watches(&policyv1alpha1.OverridePolicy{}, handler.EnqueueRequestsFromMapFunc(c.newOverridePolicyFunc())).
 		Watches(&policyv1alpha1.ClusterOverridePolicy{}, handler.EnqueueRequestsFromMapFunc(c.newOverridePolicyFunc())).
--- a/pkg/controllers/binding/binding_predicate.go
+++ b/pkg/controllers/binding/binding_predicate.go
@@ -1,0 +1,88 @@
+/*
+Copyright 2025 The Karmada Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package binding
+
+import (
+	"reflect"
+
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+
+	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
+)
+
+// hasSignificantSpecFieldChanged checks if any significant spec fields have changed
+func hasSignificantSpecFieldChanged(oldObj, newObj interface{}) bool {
+	var oldSpec, newSpec workv1alpha2.ResourceBindingSpec
+
+	switch oldBinding := oldObj.(type) {
+	case *workv1alpha2.ResourceBinding:
+		oldSpec = oldBinding.Spec
+	case *workv1alpha2.ClusterResourceBinding:
+		oldSpec = oldBinding.Spec
+	default:
+		klog.V(4).InfoS("Unknown object type in predicate", "type", reflect.TypeOf(oldObj))
+		return false
+	}
+
+	switch newBinding := newObj.(type) {
+	case *workv1alpha2.ResourceBinding:
+		newSpec = newBinding.Spec
+	case *workv1alpha2.ClusterResourceBinding:
+		newSpec = newBinding.Spec
+	default:
+		klog.V(4).InfoS("Unknown object type in predicate", "type", reflect.TypeOf(newObj))
+		return false
+	}
+
+	// Check critical fields that require reconciliation
+	return !reflect.DeepEqual(oldSpec.RequiredBy, newSpec.RequiredBy) ||
+		!reflect.DeepEqual(oldSpec.Clusters, newSpec.Clusters) ||
+		!reflect.DeepEqual(oldSpec.GracefulEvictionTasks, newSpec.GracefulEvictionTasks) ||
+		!reflect.DeepEqual(oldSpec.Resource, newSpec.Resource) ||
+		!reflect.DeepEqual(oldSpec.Suspension, newSpec.Suspension) ||
+		!reflect.DeepEqual(oldSpec.PreserveResourcesOnDeletion, newSpec.PreserveResourcesOnDeletion) ||
+		oldSpec.ConflictResolution != newSpec.ConflictResolution
+}
+
+// newBindingPredicate returns an inline predicate for ResourceBinding and ClusterResourceBinding
+// to optimize event filtering and reduce unnecessary reconciliations
+func newBindingPredicate() predicate.Predicate {
+	return predicate.Funcs{
+		CreateFunc: func(event.CreateEvent) bool {
+			return true
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			// Trigger reconciliation when deletion starts to ensure timely finalizer/cleanup handling
+			if e.ObjectOld.GetDeletionTimestamp() != e.ObjectNew.GetDeletionTimestamp() {
+				return true
+			}
+			// Check if any critical spec fields have changed
+			return hasSignificantSpecFieldChanged(e.ObjectOld, e.ObjectNew)
+		},
+		DeleteFunc: func(event.DeleteEvent) bool {
+			// Ignore delete events; reconciliation is driven by generation bump and finalizers.
+			// Cleanup is handled by controller logic (e.g., work deletion).
+			return false
+		},
+		GenericFunc: func(event.GenericEvent) bool {
+			// Process generic events (rarely used)
+			return true
+		},
+	}
+}
--- a/pkg/controllers/binding/binding_predicate_test.go
+++ b/pkg/controllers/binding/binding_predicate_test.go
@@ -1,0 +1,288 @@
+/*
+Copyright 2025 The Karmada Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package binding
+
+import (
+	"testing"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+
+	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
+)
+
+func TestBindingPredicate_Create(t *testing.T) {
+	predicate := newBindingPredicate()
+
+	// Test ResourceBinding create event
+	rb := &workv1alpha2.ResourceBinding{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "test-rb",
+			Namespace: "default",
+		},
+	}
+
+	createEvent := event.CreateEvent{Object: rb}
+	if !predicate.Create(createEvent) {
+		t.Error("Expected Create event to return true")
+	}
+
+	// Test ClusterResourceBinding create event
+	crb := &workv1alpha2.ClusterResourceBinding{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "test-crb",
+		},
+	}
+
+	createEventCRB := event.CreateEvent{Object: crb}
+	if !predicate.Create(createEventCRB) {
+		t.Error("Expected Create event for ClusterResourceBinding to return true")
+	}
+}
+
+func TestBindingPredicate_Delete(t *testing.T) {
+	predicate := newBindingPredicate()
+
+	// Test ResourceBinding delete event
+	rb := &workv1alpha2.ResourceBinding{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "test-rb",
+			Namespace: "default",
+		},
+	}
+
+	deleteEvent := event.DeleteEvent{Object: rb}
+	if predicate.Delete(deleteEvent) {
+		t.Error("Expected Delete event to return false")
+	}
+
+	// Test ClusterResourceBinding delete event
+	crb := &workv1alpha2.ClusterResourceBinding{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "test-crb",
+		},
+	}
+
+	deleteEventCRB := event.DeleteEvent{Object: crb}
+	if predicate.Delete(deleteEventCRB) {
+		t.Error("Expected Delete event for ClusterResourceBinding to return false")
+	}
+}
+
+func TestBindingPredicate_Update(t *testing.T) {
+	predicate := newBindingPredicate()
+
+	tests := []struct {
+		name     string
+		oldObj   *workv1alpha2.ResourceBinding
+		newObj   *workv1alpha2.ResourceBinding
+		expected bool
+	}{
+		{
+			name: "deletion timestamp set",
+			oldObj: &workv1alpha2.ResourceBinding{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test-rb",
+					Namespace: "default",
+				},
+			},
+			newObj: &workv1alpha2.ResourceBinding{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:              "test-rb",
+					Namespace:         "default",
+					DeletionTimestamp: &metav1.Time{Time: time.Now()},
+				},
+			},
+			expected: true,
+		},
+		{
+			name: "clusters field change",
+			oldObj: &workv1alpha2.ResourceBinding{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test-rb",
+					Namespace: "default",
+				},
+				Spec: workv1alpha2.ResourceBindingSpec{
+					Clusters: []workv1alpha2.TargetCluster{
+						{Name: "cluster1"},
+					},
+				},
+			},
+			newObj: &workv1alpha2.ResourceBinding{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test-rb",
+					Namespace: "default",
+				},
+				Spec: workv1alpha2.ResourceBindingSpec{
+					Clusters: []workv1alpha2.TargetCluster{
+						{Name: "cluster1"},
+						{Name: "cluster2"},
+					},
+				},
+			},
+			expected: true,
+		},
+		{
+			name: "resource field change",
+			oldObj: &workv1alpha2.ResourceBinding{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test-rb",
+					Namespace: "default",
+				},
+				Spec: workv1alpha2.ResourceBindingSpec{
+					Resource: workv1alpha2.ObjectReference{
+						APIVersion: "apps/v1",
+						Kind:       "Deployment",
+						Name:       "test-deployment",
+						Namespace:  "default",
+					},
+				},
+			},
+			newObj: &workv1alpha2.ResourceBinding{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test-rb",
+					Namespace: "default",
+				},
+				Spec: workv1alpha2.ResourceBindingSpec{
+					Resource: workv1alpha2.ObjectReference{
+						APIVersion: "apps/v1",
+						Kind:       "Deployment",
+						Name:       "test-deployment-v2",
+						Namespace:  "default",
+					},
+				},
+			},
+			expected: true,
+		},
+		{
+			name: "conflict resolution change",
+			oldObj: &workv1alpha2.ResourceBinding{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test-rb",
+					Namespace: "default",
+				},
+				Spec: workv1alpha2.ResourceBindingSpec{
+					ConflictResolution: "Abort",
+				},
+			},
+			newObj: &workv1alpha2.ResourceBinding{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test-rb",
+					Namespace: "default",
+				},
+				Spec: workv1alpha2.ResourceBindingSpec{
+					ConflictResolution: "Overwrite",
+				},
+			},
+			expected: true,
+		},
+		{
+			name: "no significant change",
+			oldObj: &workv1alpha2.ResourceBinding{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test-rb",
+					Namespace: "default",
+				},
+				Spec: workv1alpha2.ResourceBindingSpec{
+					Resource: workv1alpha2.ObjectReference{
+						APIVersion: "apps/v1",
+						Kind:       "Deployment",
+						Name:       "test-deployment",
+						Namespace:  "default",
+					},
+				},
+			},
+			newObj: &workv1alpha2.ResourceBinding{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test-rb",
+					Namespace: "default",
+					Labels:    map[string]string{"test": "label"},
+				},
+				Spec: workv1alpha2.ResourceBindingSpec{
+					Resource: workv1alpha2.ObjectReference{
+						APIVersion: "apps/v1",
+						Kind:       "Deployment",
+						Name:       "test-deployment",
+						Namespace:  "default",
+					},
+				},
+			},
+			expected: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			updateEvent := event.UpdateEvent{
+				ObjectOld: tt.oldObj,
+				ObjectNew: tt.newObj,
+			}
+
+			result := predicate.Update(updateEvent)
+			if result != tt.expected {
+				t.Errorf("Expected %v, got %v", tt.expected, result)
+			}
+		})
+	}
+}
+
+func TestResourceBindingPredicate_Generic(t *testing.T) {
+	predicate := newBindingPredicate()
+
+	genericEvent := event.GenericEvent{Object: &workv1alpha2.ResourceBinding{}}
+	if !predicate.Generic(genericEvent) {
+		t.Error("Expected Generic event to return true")
+	}
+}
+
+func TestResourceBindingPredicate_ClusterResourceBinding(t *testing.T) {
+	predicate := newBindingPredicate()
+
+	// Test ClusterResourceBinding update with clusters change
+	oldCRB := &workv1alpha2.ClusterResourceBinding{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "test-crb",
+		},
+		Spec: workv1alpha2.ResourceBindingSpec{
+			Clusters: []workv1alpha2.TargetCluster{
+				{Name: "cluster1"},
+			},
+		},
+	}
+
+	newCRB := &workv1alpha2.ClusterResourceBinding{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "test-crb",
+		},
+		Spec: workv1alpha2.ResourceBindingSpec{
+			Clusters: []workv1alpha2.TargetCluster{
+				{Name: "cluster1"},
+				{Name: "cluster2"},
+			},
+		},
+	}
+
+	updateEvent := event.UpdateEvent{
+		ObjectOld: oldCRB,
+		ObjectNew: newCRB,
+	}
+
+	if !predicate.Update(updateEvent) {
+		t.Error("Expected ClusterResourceBinding update with clusters change to return true")
+	}
+}
--- a/pkg/controllers/binding/cluster_resource_binding_controller.go
+++ b/pkg/controllers/binding/cluster_resource_binding_controller.go
@@ -31,6 +31,7 @@ import (
 	"k8s.io/klog/v2"
 	"k8s.io/utils/ptr"
 	controllerruntime "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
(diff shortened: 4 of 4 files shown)


rohan-019:
Hey, @XiShanYongYe-Chang 
Please take a look at this!

pkg/controllers/binding/cluster_resource_binding_controller.go
```diff
@@ -187,7 +186,7 @@ func (c *ClusterResourceBindingController) SetupWithManager(mgr controllerruntim
 		Named(ClusterResourceBindingControllerName).
 		For(&workv1alpha2.ClusterResourceBinding{}).
 		Watches(&policyv1alpha1.ClusterOverridePolicy{}, handler.EnqueueRequestsFromMapFunc(c.newOverridePolicyFunc())).
-		WithEventFilter(predicate.GenerationChangedPredicate{}).
+		WithEventFilter(NewResourceBindingPredicate()).
```
XiShanYongYe-Chang: This event filter is global; if modified, it will also affect other resources being watched. We can add an event filter related to the binding object within the `For` block.

pkg/controllers/binding/binding_predicate.go
```diff
@@ -0,0 +1,115 @@
...
+// Optimization strategy:
+// 1. Ignore Delete events - garbage collection and owner references handle cleanup
+// 2. Always process Create events - new bindings must trigger reconciliation
+// 3. Process Update events only when:
+//    - Generation changes (normal behavior)
+//    - DeletionTimestamp is set (for finalizer cleanup)
+//    - Critical spec fields change: requiredBy, clusters, gracefulEvictionTasks, resource, suspension, preserveResourcesOnDeletion, conflictResolution
+type ResourceBindingPredicate struct{}
+
+var _ predicate.Predicate = &ResourceBindingPredicate{}
```
XiShanYongYe-Chang: This method feels slightly more complex compared to directly defining an anonymous structure. ref: https://github.com/karmada-io/karmada/blob/70411ab4d30d53b15d091591791be903e46e3ef6/pkg/controllers/namespace/namespace_sync_controller.go#L201-L214

How do you think?

pkg/controllers/binding/binding_predicate.go
```diff
@@ -0,0 +1,115 @@
...
+func (r *ResourceBindingPredicate) Update(e event.UpdateEvent) bool {
+	// Check if generation changed (normal behavior)
+	if e.ObjectOld.GetGeneration() != e.ObjectNew.GetGeneration() {
+		return true
+	}
+
+	// Check if deletion timestamp is set (for finalizer cleanup)
+	if e.ObjectOld.GetDeletionTimestamp() == nil && e.ObjectNew.GetDeletionTimestamp() != nil {
+		return true
+	}
```
XiShanYongYe-Chang: When a binding object is deleted, its generation number automatically increments by 1, so we do not need to additionally check the DeletionTimestamp.

Kubernetes ref code:

https://github.com/kubernetes/kubernetes/blob/0969bb173cbbd74ea1e87cd8646746f414fc58ca/staging/src/k8s.io/apiserver/pkg/registry/rest/delete.go#L166-L172

pkg/controllers/binding/binding_predicate.go
```diff
@@ -0,0 +1,115 @@
...
+
+	// Check if critical spec fields changed
+	return r.hasCriticalSpecFieldChanged(e.ObjectOld, e.ObjectNew)
+}
+
+// Delete implements DeleteEvent filter
+func (r *ResourceBindingPredicate) Delete(_ event.DeleteEvent) bool {
+	// Ignore delete events - garbage collection and owner references handle cleanup
+	// The controller still reconciles properly using create/update and finalizer logic
```
XiShanYongYe-Chang: This comment is not entirely accurate; we did not use ownerReference.

XiShanYongYe-Chang:
Thanks a lot 
@rohan-019

pkg/controllers/binding/binding_predicate.go
rohan-019: Hi @XiShanYongYe-Chang,

I tried simplifying our binding predicate to trigger only on generation changes for deletions (since Kubernetes bumps generation when deletion starts). But in tests, we found a gap: sometimes the update event with DeletionTimestamp arrives before we see the generation bump due to informer timing. If we rely only on generation, we can miss the first reconciliation needed for finalizer/cleanup.

To fix this, I’ve restored the explicit check for when DeletionTimestamp changes from nil → non-nil. This ensures we always catch the start of deletion, even if the generation bump hasn’t shown up yet. We still keep the generation check and spec-diff checks to avoid extra reconciles.

If we later confirm that the generation bump always comes before the delete signal in practice, we can simplify again. For now, this approach keeps deletion handling reliable without adding noise.

pkg/controllers/binding/binding_predicate.go
rohan-019: I simplified the predicate by replacing the custom struct with an inline predicate. Funcs returned from NewResourceBindingPredicate, keeping the same logic and making it closer to the referenced pattern.

pkg/controllers/binding/binding_predicate.go
rohan-019: I updated the misleading comment in the Delete predicate to remove mention of ownerReferences and clarify the actual cleanup mechanism.

pkg/controllers/binding/cluster_resource_binding_controller.go
rohan-019: I moved the predicate into the For block so it only filters events for ClusterResourceBinding, and removed the global filter.

pkg/controllers/binding/binding_predicate.go
rohan-019: Please correct me if I was wrong

pkg/controllers/binding/cluster_resource_binding_controller.go
XiShanYongYe-Chang: We may not need to remove the global event filter. For OverridePolicy/ClusterOverridePolicy, it remains effective.

pkg/controllers/binding/binding_predicate.go
XiShanYongYe-Chang: > But in tests, we found a gap: sometimes the update event with DeletionTimestamp arrives before we see the generation bump due to informer timing. 

From the Kubernetes source code, it appears that the update of DeletionTimestamp and the update of generation occur simultaneously, so they should not be two separate update events.

I'm quite curious about the test results. Would you mind telling me how you conducted the test?

pkg/controllers/binding/binding_predicate.go
```diff
@@ -0,0 +1,98 @@
...
+		!reflect.DeepEqual(oldSpec.Resource, newSpec.Resource) ||
+		!reflect.DeepEqual(oldSpec.Suspension, newSpec.Suspension) ||
+		!reflect.DeepEqual(oldSpec.PreserveResourcesOnDeletion, newSpec.PreserveResourcesOnDeletion) ||
+		oldSpec.ConflictResolution != newSpec.ConflictResolution
+}
+
+// NewResourceBindingPredicate returns an inline predicate for ResourceBinding and ClusterResourceBinding
+// to optimize event filtering and reduce unnecessary reconciliations.
+//
+// Optimization strategy:
```
XiShanYongYe-Chang: I think this may not need an explanation.

pkg/controllers/binding/binding_predicate.go
```diff
@@ -0,0 +1,98 @@
+/*
+Copyright 2024 The Karmada Authors.
```
XiShanYongYe-Chang: 2024 -> 2025

Additionally, would it be better to rename the file to "predicate"?

pkg/controllers/binding/binding_predicate_test.go
```diff
@@ -0,0 +1,306 @@
+/*
+Copyright 2024 The Karmada Authors.
```
XiShanYongYe-Chang: 2024 -> 2025

XiShanYongYe-Chang (commented):
Thanks~
Can you help fix the lint error?

pkg/controllers/binding/binding_predicate.go
XiShanYongYe-Chang: > When a binding object is deleted, its generation number automatically increments by 1, so we do not need to additionally check the DeletionTimestamp.

@rohan-019 I apologize for my previous mistake; we do indeed need this judgment, but the reason might be slightly different from what you mentioned.

Since we only checked the bindingSpec subsequently, it is necessary to determine the DeletionTimestamp here. 

Can we directly judge the difference between the old and new values, as the DeletionTimestamp will not disappear once it is set?

pkg/controllers/binding/binding_predicate.go
```diff
@@ -0,0 +1,98 @@
...
+//    - Critical spec fields change: requiredBy, clusters, gracefulEvictionTasks, resource, suspension, preserveResourcesOnDeletion, conflictResolution
+func NewResourceBindingPredicate() predicate.Predicate {
+    return predicate.Funcs{
+        CreateFunc: func(event.CreateEvent) bool {
+            return true
+        },
+        UpdateFunc: func(e event.UpdateEvent) bool {
+            if e.ObjectOld.GetGeneration() != e.ObjectNew.GetGeneration() {
+                return true
+            }
```
XiShanYongYe-Chang: I think here needs to be judged like this:

```go
if e.ObjectOld.GetGeneration() == e.ObjectNew.GetGeneration() {
                return false
}
```

pkg/controllers/binding/cluster_resource_binding_controller.go
rohan-019: Thanks for clarifying. 👍 I’ll keep NewResourceBindingPredicate() inside the For(&ClusterResourceBinding...) so it only affects bindings, and also retain the global WithEventFilter(predicate.GenerationChangedPredicate{}) so that it continues to work for OverridePolicy/ClusterOverridePolicy.

pkg/controllers/binding/binding_predicate.go
rohan-019: 1. Yes, Kubernetes bumps generation on delete, but in practice the informer may send the `DeletionTimestamp `update before the generation bump arrives.

2. The unit tests simulate this ordering, and without the explicit `DeletionTimestamp` check, deletion handling is missed → tests fail.

3. Keeping the extra check ensures we reliably catch the start of deletion, even if generation hasn’t bumped yet.

For the tests, I used the existing unit test suite in `binding_predicate_test.go.` Specifically, the case T`estResourceBindingPredicate_Update/deletion_timestamp_set` simulates a binding object whose DeletionTimestamp changes from `nil `→ `non-nil `while the generation has not yet bumped.
When I remove that condition, the test fails with:
```
--- FAIL: TestResourceBindingPredicate_Update/deletion_timestamp_set
    binding_predicate_test.go:256: Expected true, got false
```

pkg/controllers/binding/binding_predicate.go
rohan-019: We compare the generation of the old and new object to decide whether to trigger reconciliation. Returning true here means the object has changed in a meaningful way that requires handling. The original code checks if the generation has changed using `!=`, which is correct because a change in generation signals that the spec has been updated. Reversing this to return false when generations are equal would prevent necessary updates from being processed, causing reconciliation to be skipped, as seen in the failed test cases.


```
$ go test ./pkg/controllers/binding -v -run TestResourceBindingPredicate
I0926 09:53:10.187682    2124 defaults.go:54] default Karmada Image Version: v0.0.0-master
=== RUN   TestResourceBindingPredicate_Create
--- PASS: TestResourceBindingPredicate_Create (0.00s)
=== RUN   TestResourceBindingPredicate_Delete
--- PASS: TestResourceBindingPredicate_Delete (0.00s)
=== RUN   TestResourceBindingPredicate_Update
=== RUN   TestResourceBindingPredicate_Update/generation_change
    binding_predicate_test.go:256: Expected true, got false
=== RUN   TestResourceBindingPredicate_Update/deletion_timestamp_set
    binding_predicate_test.go:256: Expected true, got false
=== RUN   TestResourceBindingPredicate_Update/clusters_field_change
    binding_predicate_test.go:256: Expected true, got false
=== RUN   TestResourceBindingPredicate_Update/resource_field_change
    binding_predicate_test.go:256: Expected true, got false
=== RUN   TestResourceBindingPredicate_Update/conflict_resolution_change
    binding_predicate_test.go:256: Expected true, got false
=== RUN   TestResourceBindingPredicate_Update/no_significant_change
--- FAIL: TestResourceBindingPredicate_Update (0.00s)
    --- FAIL: TestResourceBindingPredicate_Update/generation_change (0.00s)
    --- FAIL: TestResourceBindingPredicate_Update/deletion_timestamp_set (0.00s)
    --- FAIL: TestResourceBindingPredicate_Update/clusters_field_change (0.00s)
    --- FAIL: TestResourceBindingPredicate_Update/resource_field_change (0.00s)
    --- FAIL: TestResourceBindingPredicate_Update/conflict_resolution_change (0.00s)
    --- PASS: TestResourceBindingPredicate_Update/no_significant_change (0.00s)
=== RUN   TestResourceBindingPredicate_Generic
--- PASS: TestResourceBindingPredicate_Generic (0.00s)
=== RUN   TestResourceBindingPredicate_ClusterResourceBinding
    binding_predicate_test.go:304: Expected ClusterResourceBinding update with clusters change to return true   
--- FAIL: TestResourceBindingPredicate_ClusterResourceBinding (0.00s)
FAIL
FAIL    github.com/karmada-io/karmada/pkg/controllers/binding   0.436s
FAIL
```

pkg/controllers/binding/binding_predicate.go
rohan-019: I kept it consistent as all related files follow the `<resource>_controller.go` style, like `binding_controller.go` and `cluster_resource_binding_controller.go.` The file I added, `binding_predicate.go`, fits this pattern.

rohan-019:
> Thanks~ Can you help fix the lint error?

I tried several times to fix lint errors using `golangci-lint run --fix pkg/controllers/binding/binding_predicate.go pkg/controllers/binding/cluster_resource_binding_controller.go` and also attempted a manual fix, but ended up with lint errors. Could you help me figure out what’s going wrong?

pkg/controllers/binding/binding_predicate.go
XiShanYongYe-Chang: > Kubernetes bumps generation on delete, but in practice the informer may send the DeletionTimestamp update before the generation bump arrives.

Thanks for the explanation. The changes to DeletionTimestamp and generation occur within the same update request, so the informer should not receive two separate events. During my testing, I only observed one event.

Of course, from a testing perspective, we can construct it this way, but it probably won't actually happen.

pkg/controllers/binding/binding_predicate.go
XiShanYongYe-Chang: Hi @rohan-019 The changes in generation are automatically handled by the apiserver and cannot be modified by the user:

https://github.com/kubernetes/kubernetes/blob/f884da7ff9ea583e2e86746dd4018bd797d6b617/staging/src/k8s.io/apiextensions-apiserver/pkg/registry/customresource/strategy.go#L178-L187

From the code, it can be observed that any change in the spec fields will result in an increment of the generation. If the comparison here determines that the new and old values are not equal, the subsequent logic will not have the opportunity to execute.

pkg/controllers/binding/binding_predicate.go
```diff
@@ -0,0 +1,90 @@
...
+*/
+
+package binding
+
+import (
+    "reflect"
+
+    "sigs.k8s.io/controller-runtime/pkg/event"
+    "sigs.k8s.io/controller-runtime/pkg/predicate"
+    "k8s.io/klog/v2"
```
XiShanYongYe-Chang:
```go
    "k8s.io/klog/v2"
	"sigs.k8s.io/controller-runtime/pkg/event"
	"sigs.k8s.io/controller-runtime/pkg/predicate"
```

pkg/controllers/binding/binding_predicate.go
```diff
@@ -0,0 +1,91 @@
...
+		!reflect.DeepEqual(oldSpec.GracefulEvictionTasks, newSpec.GracefulEvictionTasks) ||
+		!reflect.DeepEqual(oldSpec.Resource, newSpec.Resource) ||
+		!reflect.DeepEqual(oldSpec.Suspension, newSpec.Suspension) ||
+		!reflect.DeepEqual(oldSpec.PreserveResourcesOnDeletion, newSpec.PreserveResourcesOnDeletion) ||
+		oldSpec.ConflictResolution != newSpec.ConflictResolution
+}
+
+// NewResourceBindingPredicate returns an inline predicate for ResourceBinding and ClusterResourceBinding
+// to optimize event filtering and reduce unnecessary reconciliations
+func NewResourceBindingPredicate() predicate.Predicate {
```
XiShanYongYe-Chang: We currently do not need to set this method as an export method.

pkg/controllers/binding/binding_predicate.go
```diff
@@ -0,0 +1,91 @@
...
+
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+
+	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
+)
+
+// hasCriticalSpecFieldChanged checks if any critical spec fields have changed
+func hasCriticalSpecFieldChanged(oldObj, newObj interface{}) bool {
```
XiShanYongYe-Chang: "Critical" might not be the most appropriate term for this context; is there a more suitable word?

pkg/controllers/binding/cluster_resource_binding_controller.go
```diff
@@ -185,9 +185,9 @@ func (c *ClusterResourceBindingController) checkDirectPurgeOrphanWorks(ctx conte
 func (c *ClusterResourceBindingController) SetupWithManager(mgr controllerruntime.Manager) error {
 	return controllerruntime.NewControllerManagedBy(mgr).
 		Named(ClusterResourceBindingControllerName).
-		For(&workv1alpha2.ClusterResourceBinding{}).
+		For(&workv1alpha2.ClusterResourceBinding{}, builder.WithPredicates(NewResourceBindingPredicate())).
 		Watches(&policyv1alpha1.ClusterOverridePolicy{}, handler.EnqueueRequestsFromMapFunc(c.newOverridePolicyFunc())).
-		WithEventFilter(predicate.GenerationChangedPredicate{}).
+		WithEventFilter(NewResourceBindingPredicate()).
```
XiShanYongYe-Chang: The global event filter does not need to be modified.

pkg/controllers/binding/binding_predicate.go
```diff
@@ -0,0 +1,91 @@
...
+// to optimize event filtering and reduce unnecessary reconciliations
+func NewResourceBindingPredicate() predicate.Predicate {
+	return predicate.Funcs{
+		CreateFunc: func(event.CreateEvent) bool {
+			return true
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			if e.ObjectOld.GetGeneration() != e.ObjectNew.GetGeneration() {
+				return true
+			}
```
XiShanYongYe-Chang: This judgment will render our subsequent optimizations ineffective. You can test under what circumstances the generation of CR resources changes.

XiShanYongYe-Chang: We can directly remove this judgment because the global event filter has already checked the generation.

XiShanYongYe-Chang:
Hi @rohan-019, since this PR is the first one in our series, I have quite a few review comments. ^-^

rohan-019:
> Hi @rohan-019, since this PR is the first one in our series, I have quite a few review comments. ^-^

Hi @XiShanYongYe-Chang ,
I really appreciate your feedback, especially since this is the first PR in the series. I’ll go through all your comments carefully and make the necessary updates.

Thanks again for taking the time to review.

pkg/controllers/binding/binding_predicate.go
rohan-019: Changed to hasSignificantSpecFieldChanged

XiShanYongYe-Chang:
Hi @rohan-019, the newly added ut has failed.

pkg/controllers/binding/binding_controller.go
```diff
@@ -188,7 +187,7 @@ func (c *ResourceBindingController) SetupWithManager(mgr controllerruntime.Manag
 	return controllerruntime.NewControllerManagedBy(mgr).
 		Named(ControllerName).
 		For(&workv1alpha2.ResourceBinding{}).
-		WithEventFilter(predicate.GenerationChangedPredicate{}).
+		WithEventFilter(newResourceBindingPredicate()).
```
XiShanYongYe-Chang: The eventFilter added in For is specific to the objects set within `For`, whereas the eventFilter added in `WithEventFilter` is global, targeting all objects, including those in `For` and `Watch`. Moreover, its execution order precedes the eventFilters set within `For` and `Watch`.

We can modify it like this:
```go
-               For(&workv1alpha2.ResourceBinding{}).
-               WithEventFilter(newResourceBindingPredicate()).
+               For(&workv1alpha2.ResourceBinding{}, builder.WithPredicates(newResourceBindingPredicate())).
+               WithEventFilter(predicate.GenerationChangedPredicate{}).
```

pkg/controllers/binding/binding_predicate.go
```diff
@@ -0,0 +1,88 @@
...
+		!reflect.DeepEqual(oldSpec.GracefulEvictionTasks, newSpec.GracefulEvictionTasks) ||
+		!reflect.DeepEqual(oldSpec.Resource, newSpec.Resource) ||
+		!reflect.DeepEqual(oldSpec.Suspension, newSpec.Suspension) ||
+		!reflect.DeepEqual(oldSpec.PreserveResourcesOnDeletion, newSpec.PreserveResourcesOnDeletion) ||
+		oldSpec.ConflictResolution != newSpec.ConflictResolution
+}
+
+// newResourceBindingPredicate returns an inline predicate for ResourceBinding and ClusterResourceBinding
+// to optimize event filtering and reduce unnecessary reconciliations
+func newResourceBindingPredicate() predicate.Predicate {
```
XiShanYongYe-Chang: Given that it is called by both ResourceBinding and ClusterResourceBinding, it is recommended to name this function newBindingPredicate.

rohan-019:
> the newly added ut has failed.

Hey @XiShanYongYe-Chang ,
I initially removed the generation check based on your comment that the global event filter already handles it. However, after testing We found that both the new unit test (TestResourceBindingPredicate_Update/generation_change) and the related e2e test failed once this check was removed.

The reason is that once we override UpdateFunc with custom logic, the default GenerationChangedPredicate is no longer applied automatically. This means spec-only changes are not triggering reconciliation unless we explicitly handle the generation change inside our predicate.

To keep the behavior correct and align with the current tests, I’ve restored the != generation check in UpdateFunc.

XiShanYongYe-Chang:
> We found that both the new unit test (TestResourceBindingPredicate_Update/generation_change) and the related e2e test failed once this check was removed.

This test case is problematic because it is unlikely that only the generation will change while other fields remain unchanged. Therefore, it needs to be redesigned or removed.

XiShanYongYe-Chang:
> To keep the behavior correct and align with the current tests, I’ve restored the != generation check in UpdateFunc.

There's no need to do this.

rohan-019:
> > We found that both the new unit test (TestResourceBindingPredicate_Update/generation_change) and the related e2e test failed once this check was removed.
> 
> This test case is problematic because it is unlikely that only the generation will change while other fields remain unchanged. Therefore, it needs to be redesigned or removed.

That makes sense, as the Generation changes are handled globally, so this predicate doesn't need to check for them.
But instead of removing the generation check, I redesigned so that this predicate now focuses on catching edge cases where:
-Generation hasn't changed BUT
-Specific critical spec fields have changed (like RequiredBy, Clusters, etc.)

rohan-019:
> > To keep the behavior correct and align with the current tests, I’ve restored the != generation check in UpdateFunc.
> 
> There's no need to do this.

rohan-019:
Hey @XiShanYongYe-Chang,
Don't mind, I accidentally closed PR!!

pkg/controllers/binding/binding_predicate.go
```diff
@@ -0,0 +1,88 @@
...
+// newBindingPredicate returns an inline predicate for ResourceBinding and ClusterResourceBinding
+// to optimize event filtering and reduce unnecessary reconciliations
+func newBindingPredicate() predicate.Predicate {
+	return predicate.Funcs{
+		CreateFunc: func(event.CreateEvent) bool {
+			return true
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			// Also trigger on the start of deletion to ensure timely finalizer/cleanup reconciliation,
+			// even if a generation bump isn't observed yet by informers.
```
XiShanYongYe-Chang: This comment seems incorrect; the premise here is that a change in generation has occurred, and only then do we need to further determine what has changed.

XiShanYongYe-Chang: This comparison can be simplified to:
```go
e.ObjectOld.GetDeletionTimestamp() != e.ObjectNew.GetDeletionTimestamp() 
```

pkg/controllers/binding/binding_predicate_test.go
```diff
@@ -0,0 +1,306 @@
...
+			},
+			newObj: &workv1alpha2.ResourceBinding{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:       "test-rb",
+					Namespace:  "default",
+					Generation: 2,
+				},
+			},
+			expected: false,
+		},
```
XiShanYongYe-Chang: Although this case appears to be theoretically sound, it is highly unlikely to occur in reality, making it somewhat misleading. I recommend deleting it directly.

rohan-019:
Hi @XiShanYongYe-Chang 
I tried simplifying the deletion check from
`e.ObjectOld.GetDeletionTimestamp() == nil && e.ObjectNew.GetDeletionTimestamp() != nil`
to
`e.ObjectOld.GetDeletionTimestamp() != e.ObjectNew.GetDeletionTimestamp()`

as suggested, but this change caused failures in the e2e tests (e.g., migration_and_rollback_test.go timing out).

The reason is that the simplified form triggers reconciliation not only when deletion starts `(nil → non-nil)`, but also if the deletion timestamp is updated while the object is already marked for deletion. That extra trigger led to unexpected reconcile behavior and race conditions, which is why the e2e tests failed.

So it seems we need to keep the original, more restrictive check to ensure reconciliation only fires once at the beginning of deletion.

XiShanYongYe-Chang:
> The reason is that the simplified form triggers reconciliation not only when deletion starts `(nil → non-nil)`, but also if the deletion timestamp is updated while the object is already marked for deletion. That extra trigger led to unexpected reconcile behavior and race conditions, which is why the e2e tests failed.

Once the timestamp is set, it will not be modified again.

The E2E error should be just an occasional issue and likely not introduced by the current changes. Let's ignore it for now.

Can you help fix the lint errors and compress the commits into one?

pkg/controllers/binding/binding_predicate.go
```diff
@@ -0,0 +1,88 @@
...
+	"reflect"
+
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+
+	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
+)
+
+// hasSignificantSpecFieldChanged checks if any critical spec fields have changed
```
XiShanYongYe-Chang: Can you help update the comment info to align with the function name?

pkg/controllers/binding/binding_predicate_test.go
```diff
@@ -0,0 +1,288 @@
...
+	"testing"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+
+	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
+)
+
+func TestResourceBindingPredicate_Create(t *testing.T) {
```
XiShanYongYe-Chang: This test function not only tests ResourceBinding but also ClusterResourceBinding, making the name of this test function somewhat inappropriate.

pkg/controllers/binding/binding_predicate_test.go
```diff
@@ -0,0 +1,288 @@
...
+		},
+	}
+
+	createEventCRB := event.CreateEvent{Object: crb}
+	if !predicate.Create(createEventCRB) {
+		t.Error("Expected Create event for ClusterResourceBinding to return true")
+	}
+}
+
+func TestResourceBindingPredicate_Delete(t *testing.T) {
```
XiShanYongYe-Chang: ditto

XiShanYongYe-Chang (commented):
There are still some minor issues, but everything else is basically fine. 👍

pkg/controllers/binding/binding_predicate_test.go
XiShanYongYe-Chang: > This test function not only tests ResourceBinding but also ClusterResourceBinding, making the name of this test function somewhat inappropriate.

This comment seems to have been forgotten to be revised.

pkg/controllers/binding/binding_predicate_test.go
```diff
@@ -0,0 +1,288 @@
...
+		},
+	}
+
+	createEventCRB := event.CreateEvent{Object: crb}
+	if !predicate.Create(createEventCRB) {
+		t.Error("Expected Create event for ClusterResourceBinding to return true")
+	}
+}
+
+func TestBindingPredicate_Delete(t *testing.T) {
```
XiShanYongYe-Chang, line 58: This test function not only tests ResourceBinding but also ClusterResourceBinding, making the name of this test function somewhat inappropriate.

pkg/controllers/binding/binding_predicate_test.go
rohan-019: Really sorry for that @XiShanYongYe-Chang

pkg/controllers/binding/binding_predicate_test.go
rohan-019: > This test function not only tests ResourceBinding but also ClusterResourceBinding, making the name of this test function somewhat inappropriate.

I'm sorry for not getting back to you sooner. Can you please suggest an appropriate name!!

XiShanYongYe-Chang: Sorry, I made a mistake in my comment. The current name is fine. "Binding" can refer to both ResourceBinding and ClusterResourceBinding.

pkg/controllers/binding/binding_predicate_test.go
XiShanYongYe-Chang: It's okay.

XiShanYongYe-Chang (commented):
Thanks a lot~ 👍 
/lgtm

/cc @RainbowMango @zach593 @zhzhuang-zju

RainbowMango:
I guess this PR is part of #6780, so I updated the `Fixes to Part of` in the PR description.

pkg/controllers/binding/binding_predicate.go
```diff
@@ -0,0 +1,88 @@
...
+		},
+		DeleteFunc: func(event.DeleteEvent) bool {
+			// Ignore delete events; reconciliation is driven by generation bump and finalizers.
+			// Cleanup is handled by controller logic (e.g., work deletion).
+			return false
+		},
+		GenericFunc: func(event.GenericEvent) bool {
+			// Process generic events (rarely used)
+			return true
+		},
```
RainbowMango, line 86: Can you remind me what kind of generic event is used here?
PS: The comments say it is rarely used, but it still returns true.

pkg/controllers/binding/binding_predicate.go
```diff
@@ -0,0 +1,88 @@
...
+
+	// Check critical fields that require reconciliation
+	return !reflect.DeepEqual(oldSpec.RequiredBy, newSpec.RequiredBy) ||
+		!reflect.DeepEqual(oldSpec.Clusters, newSpec.Clusters) ||
+		!reflect.DeepEqual(oldSpec.GracefulEvictionTasks, newSpec.GracefulEvictionTasks) ||
+		!reflect.DeepEqual(oldSpec.Resource, newSpec.Resource) ||
+		!reflect.DeepEqual(oldSpec.Suspension, newSpec.Suspension) ||
+		!reflect.DeepEqual(oldSpec.PreserveResourcesOnDeletion, newSpec.PreserveResourcesOnDeletion) ||
+		oldSpec.ConflictResolution != newSpec.ConflictResolution
+}
```
RainbowMango, line 61: Can you explain why the following field changes can be ignored?
- ReplicaRequirements
- Components
- Failover

We should be very cautious when ignoring the changes.

XiShanYongYe-Chang: How can we determine which fields need to be focused on? There is a rule for this: whether the target field is involved in the processing of the binding controller. More specifically, it refers to which fields are used during the generation of work.

XiShanYongYe-Chang:
Hi @rohan-019 

PR #6796, opened 2025-09-29
Dependency conflict resolved for attached resourcebindings

### LINKED ISSUE #6000: Whether to forbid the same resource multi dependency distribution

Specify whether the same resource can be distributed by multiple resources in the [dependency distribution](https://karmada.io/docs/next/userguide/scheduling/propagate-dependencies/) feature.

**Why is this needed**:

According to the feedback received by users, some users do not involve the scenario where a single resource is distributed dependent by multiple resources, but some users will use the scenario, for example, sharing the same secret to pull images.

If allowed, it will bring some risks to users. We need to analyze these risks to think about whether we can resolve them in some ways or explicitly prohibit them.

Kexin2000:
**What type of PR is this?**

/kind feature

Computes effective dependency-related field values for attached ResourceBindings when the same dependent resource is referenced by multiple independent ResourceBindings.

How it works:
- Source of truth: referencing bindings listed in `Spec.RequiredBy`.
- Resolution rules:
  - ConflictResolution: any Overwrite → Overwrite; else Abort.
  - PreserveResourcesOnDeletion: any true → true; else false.
- Conflict detection: if either field has mixed values across referencing bindings, emit a single Warning event with reason `DependencyPolicyConflict`.
- Scope: only for dependency-generated ResourceBindings (`Spec.Placement == nil`); policy-owned bindings are not changed.

Fixes #6000  
Part of #6803

```release-note
`karmada-controller-manager`:
Compute effective field values for attached ResourceBindings when referenced by multiple ResourceBindings:
- ConflictResolution: any Overwrite → Overwrite; else Abort.
- PreserveResourcesOnDeletion: any true → true; else false.
Applies only to dependency-generated ResourceBindings; policy-owned bindings are unaffected.

### DIFF
--- a/pkg/dependenciesdistributor/dependencies_distributor.go
+++ b/pkg/dependenciesdistributor/dependencies_distributor.go
@@ -563,7 +563,28 @@ func (d *DependenciesDistributor) removeScheduleResultFromAttachedBindings(bindi
 		delete(attachedBindings[index].Labels, bindingLabelKey)
 		updatedSnapshot := deleteBindingFromSnapshot(bindingNamespace, bindingName, attachedBindings[index].Spec.RequiredBy)
 		attachedBindings[index].Spec.RequiredBy = updatedSnapshot
-		attachedBindings[index].Spec.PreserveResourcesOnDeletion = nil
+		// Only adjust strategy fields for dependency-generated bindings (Placement == nil).
+		if attachedBindings[index].Spec.Placement == nil {
+			// When no remaining referrers, clear strategy fields.
+			if len(updatedSnapshot) == 0 {
+				attachedBindings[index].Spec.ConflictResolution = ""
+				attachedBindings[index].Spec.PreserveResourcesOnDeletion = nil
+			} else {
+				// Recompute effective strategy from remaining RequiredBy.
+				rbs, err := d.resolveResourceBindingFromSnapshots(context.TODO(), updatedSnapshot)
+				if err != nil {
+					klog.Errorf("Failed to resolve for resourceBinding(%s/%s): %v", binding.Namespace, binding.Name, err)
+					errs = append(errs, err)
+					continue
+				}
+
+				_, effectiveCR := d.detectAndResolveConflictResolution(rbs)
+				attachedBindings[index].Spec.ConflictResolution = effectiveCR
+
+				_, effectivePreserve := d.detectAndResolvePreserveOnDeletion(rbs)
+				attachedBindings[index].Spec.PreserveResourcesOnDeletion = ptr.To(effectivePreserve)
+			}
+		}
 		if err := d.Client.Update(context.TODO(), attachedBindings[index]); err != nil {
 			klog.Errorf("Failed to update binding(%s/%s): %v", binding.Namespace, binding.Name, err)
 			errs = append(errs, err)
@@ -593,22 +614,25 @@ func (d *DependenciesDistributor) createOrUpdateAttachedBinding(ctx context.Cont
 			// Pre-fetch referencing ResourceBindings once and reuse for conflict checks.
 			rbs, err := d.resolveResourceBindingFromSnapshots(ctx, mergedRequiredBy)
 			if err != nil {
-				klog.Errorf("Failed to resolve referencing ResourceBindings for (%s): %v", bindingKey, err)
+				klog.Errorf("Failed to resolve referencing ResourceBindings for(%s): %v", bindingKey, err)
 				return err
 			}
 
-			crConflictDetected := d.conflictDetectedForConflictResolution(rbs)
-			preserveConflictDetected := d.conflictDetectedForPreserveOnDeletion(rbs)
+			// Compute effective policies from referencing ResourceBindings and detect conflicts.
+			crConflicted, effectiveCR := d.detectAndResolveConflictResolution(rbs)
+			preserveConflicted, effectivePreserve := d.detectAndResolvePreserveOnDeletion(rbs)
 
-			d.recordEventIfPolicyConflict(existBinding, crConflictDetected, preserveConflictDetected)
+			// Emit warning event if any conflict detected.
+			d.recordEventIfPolicyConflict(existBinding, crConflicted, preserveConflicted)
 
-			existBinding.Spec.ConflictResolution = attachedBinding.Spec.ConflictResolution
+			// Update with effective results.
+			existBinding.Spec.ConflictResolution = effectiveCR
+			existBinding.Spec.PreserveResourcesOnDeletion = ptr.To(effectivePreserve)
 		}
 
 		existBinding.Spec.RequiredBy = mergedRequiredBy
 		existBinding.Labels = util.DedupeAndMergeLabels(existBinding.Labels, attachedBinding.Labels)
 		existBinding.Spec.Resource = attachedBinding.Spec.Resource
-		existBinding.Spec.PreserveResourcesOnDeletion = attachedBinding.Spec.PreserveResourcesOnDeletion
 
 		if err := d.Client.Update(ctx, existBinding); err != nil {
 			klog.Errorf("Failed to update resourceBinding(%s): %v", bindingKey, err)
--- a/pkg/dependenciesdistributor/policy_utils.go
+++ b/pkg/dependenciesdistributor/policy_utils.go
@@ -47,8 +47,10 @@ func (d *DependenciesDistributor) resolveResourceBindingFromSnapshots(ctx contex
 	return rbs, nil
 }
 
-// conflictDetectedForConflictResolution checks if there is a conflict in ConflictResolution from referencing ResourceBindings.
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by Kexin2000** [comment]
Hi @XiShanYongYe-Chang, the PR is now ready for review. Please take a look when you are free. 😃

**Comment by XiShanYongYe-Chang** [comment]
Thanks~
/assign
/gemini review

File: pkg/dependenciesdistributor/policy_utils.go [hunk 1]
```diff
@@ -0,0 +1,133 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	policyv1alpha1 "github.com/karmada-io/karmada/pkg/apis/policy/v1alpha1"
+	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
+	"github.com/karmada-io/karmada/pkg/events"
+)
+
+// ResolveResourceBindingFromSnapshots resolves and returns the list of ResourceBindings
+// referenced by the given BindingSnapshots.
+func (d *DependenciesDistributor) ResolveResourceBindingFromSnapshots(ctx context.Context, requiredBy []workv1alpha2.BindingSnapshot) ([]*workv1alpha2.ResourceBinding, error) {
```
XiShanYongYe-Chang: This function may not need to be exported.

pkg/dependenciesdistributor/policy_utils.go
```diff
@@ -0,0 +1,133 @@
...
+		}
+		rbs = append(rbs, refRB)
+	}
+	return rbs, nil
+}
+
+// AggregateAndDetectConflictResolution aggregates the ConflictResolution across referenced ResourceBindings
+// and reports if there is a conflict (both Overwrite and Abort present).
+// Aggregation rule: any Overwrite -> Overwrite; else Abort. Empty/zero value counts as Abort.
+func (d *DependenciesDistributor) AggregateAndDetectConflictResolution(rbs []*workv1alpha2.ResourceBinding) (agg policyv1alpha1.ConflictResolution, conflicted bool, err error) {
```
XiShanYongYe-Chang: This function may not need to be exported, and the error return value may not be needed.

Is the naming of this function not quite appropriate? We probably don't have an aggregation operation.

pkg/dependenciesdistributor/policy_utils.go
```diff
@@ -0,0 +1,133 @@
...
+	} else {
+		agg = policyv1alpha1.ConflictAbort
+	}
+	return agg, conflicted, nil
+}
+
+// AggregateAndDetectPreserveOnDeletion aggregates the PreserveResourcesOnDeletion across referenced ResourceBindings
+// and reports if there is a conflict (both true and false present).
+// Aggregation rule: any true -> true; else false. Empty(nil) counts as false.
+func (d *DependenciesDistributor) AggregateAndDetectPreserveOnDeletion(rbs []*workv1alpha2.ResourceBinding) (agg bool, conflicted bool, err error) {
```
XiShanYongYe-Chang: ditto

RainbowMango:
@Kexin2000 Could you please resolve the conflicts?

pkg/dependenciesdistributor/policy_utils.go
Kexin2000: Done

pkg/dependenciesdistributor/policy_utils.go
Kexin2000: Done.

pkg/dependenciesdistributor/dependencies_distributor.go
```diff
@@ -563,7 +563,27 @@ func (d *DependenciesDistributor) removeScheduleResultFromAttachedBindings(bindi
...
+			// When no remaining referrers, clear strategy fields.
+			if len(updatedSnapshot) == 0 {
+				attachedBindings[index].Spec.ConflictResolution = ""
+				attachedBindings[index].Spec.PreserveResourcesOnDeletion = nil
+			} else {
+				// Recompute effective strategy from remaining RequiredBy.
+				rbs, err := d.resolveResourceBindingFromSnapshots(context.TODO(), updatedSnapshot)
+				if err != nil {
+					klog.Errorf("Failed to resolve for resourceBinding(%s/%s): %v", binding.Namespace, binding.Name, err)
+					return err
```
XiShanYongYe-Chang: It is recommended to use the append err method here and continue processing other binding objects using continue.

pkg/dependenciesdistributor/policy_utils.go
```diff
@@ -47,8 +47,10 @@ func (d *DependenciesDistributor) resolveResourceBindingFromSnapshots(ctx contex
 	return rbs, nil
 }
 
-// conflictDetectedForConflictResolution checks if there is a conflict in ConflictResolution from referencing ResourceBindings.
-func (d *DependenciesDistributor) conflictDetectedForConflictResolution(rbs []*workv1alpha2.ResourceBinding) bool {
+// detectAndResolveConflictResolution computes the effective ConflictResolution across referenced ResourceBindings
+// and reports if there is a conflict (both Overwrite and Abort present).
+// Resolution rule: any Overwrite -> Overwrite; else Abort. Empty/zero value counts as Abort.
+func (d *DependenciesDistributor) detectAndResolveConflictResolution(rbs []*workv1alpha2.ResourceBinding) (effective policyv1alpha1.ConflictResolution, conflicted bool) {
```
XiShanYongYe-Chang: Does the term "conflict" conflict with "ConflictResolution" in function naming?

In addition, the order of return values does not match the order of function naming.

pkg/dependenciesdistributor/policy_utils.go
```diff
@@ -59,16 +61,25 @@ func (d *DependenciesDistributor) conflictDetectedForConflictResolution(rbs []*w
...
+	}
+	return effective, conflicted
 }
 
-// conflictDetectedForPreserveOnDeletion checks if there is a conflict in PreserveResourcesOnDeletion from referencing ResourceBindings.
-func (d *DependenciesDistributor) conflictDetectedForPreserveOnDeletion(rbs []*workv1alpha2.ResourceBinding) bool {
+// detectAndResolvePreserveOnDeletion computes the effective PreserveResourcesOnDeletion across referenced ResourceBindings
+// and reports if there is a conflict (both true and false present).
+// Resolution rule: any true -> true; else false. Empty(nil) counts as false.
+func (d *DependenciesDistributor) detectAndResolvePreserveOnDeletion(rbs []*workv1alpha2.ResourceBinding) (effective bool, conflicted bool) {
```
XiShanYongYe-Chang: ditto

pkg/dependenciesdistributor/dependencies_distributor.go
```diff
@@ -590,25 +610,27 @@ func (d *DependenciesDistributor) createOrUpdateAttachedBinding(ctx context.Cont
 		// If the spec.Placement is nil, this means that existBinding is generated by the dependency mechanism.
 		// If the spec.Placement is not nil, then it must be generated by PropagationPolicy.
 		if existBinding.Spec.Placement == nil {
-			// Pre-fetch referencing ResourceBindings once and reuse for conflict checks.
```
XiShanYongYe-Chang, line 593: Is this comment unnecessary to remove?

pkg/dependenciesdistributor/policy_utils.go
Kexin2000: > Does the term "conflict" conflict with "ConflictResolution" in function naming?

 Yes
If there are two names of _conflict_ in the function name, it might be confusing and make the function name long.

> The order of return values does not match the order of function naming.

 Good point, I’ll update it

pkg/dependenciesdistributor/dependencies_distributor.go
Kexin2000: I deleted it by mistake, I’ll put it back

pkg/dependenciesdistributor/dependencies_distributor.go
```diff
@@ -593,22 +614,25 @@ func (d *DependenciesDistributor) createOrUpdateAttachedBinding(ctx context.Cont
...
+			// Compute effective policies from referencing ResourceBindings and detect conflicts.
+			crConflicted, effectiveCR := d.detectAndResolveConflictResolution(rbs)
+			preserveConflicted, effectivePreserve := d.detectAndResolvePreserveOnDeletion(rbs)
 
-			d.recordEventIfPolicyConflict(existBinding, crConflictDetected, preserveConflictDetected)
+			// Emit warning event if any conflict detected.
+			d.recordEventIfPolicyConflict(existBinding, crConflicted, preserveConflicted)
 
-			existBinding.Spec.ConflictResolution = attachedBinding.Spec.ConflictResolution
+			// Update with aggregation results.
```
XiShanYongYe-Chang: This review uses aggregation, which may not be entirely accurate.

XiShanYongYe-Chang (commented):
Thanks
Generally LGTM, only a comment suggestion.

Issue #6803, opened 2025-09-30
[Umbrella][Summer OSPP 2025]Resolving Conflicts with Dependency Distribution Policies


XiShanYongYe-Chang:
Currently, Karmada's dependency distribution feature supports the same resource being depended on by multiple resources, and it distributes according to these resources. For example, different Deployment resource objects share the same secret to pull images. 

However, this usage introduces a conflict issue: when no PropagationPolicy matches the dependent resources in the environment, the PropagationPolicy matched by the dependent resources will be synchronized when generating the resourceBinding object. If there are multiple dependent resources and the PropagationPolicy definitions matched by these resources are different, some policies related to distribution will conflict.

For Karmada versions 1.15 and earlier, the specific policy configurations that can cause conflicts include:

- PreserveResourcesOnDeletion 
- ConflictResolution 

Related issue: #6000 
Summer OSPP 2025 Project Link: https://summer-ospp.ac.cn/org/prodetail/255c40195?list=org&navpage=org

To facilitate the advancement of project development, we have broken down the following subtasks according to resource granularity.

- [ ] Proposal
- [ ] Display conflicts to users: send warning events.
- [ ] Docs
  - [ ] Explain to users how to resolve conflicts when they occur
  - [ ] If users do not actively resolve conflicts, the actions the system will take.
- [ ] Conflict Resolution: Specify conflict rules, with user-defined policies taking precedence.
- [ ] E2E
- [ ] Version Upgrade Document, User Notification
- [ ] API Change Consideration: Should the conflicting fields be introduced in ResourceBindingSpec's `RequiredBy` field?

XiShanYongYe-Chang:
@Kexin2000 Thank you for participating in the development of this project. All development work for the current project has been completed. Congratulations!

Hope you will continue to participate in open source and contribute to the Karmada project.

The remaining tasks will be tracked as issues, so I will close the current issue.

/close

PR #6808, opened 2025-09-30
Resolving Conflicts with Dependency Distribution Policies Proposal

### LINKED ISSUE #6000: Whether to forbid the same resource multi dependency distribution

Kexin2000:
**What type of PR is this?**

/kind documentation

- Proposal for Issue #6000 

Fixes #6000 
Part of #6803 

None.

NONE

### DIFF
--- a/docs/proposals/dependency-policy-conflict-resolution/dependency-policy-conflict-resolution.md
+++ b/docs/proposals/dependency-policy-conflict-resolution/dependency-policy-conflict-resolution.md
@@ -1,0 +1,360 @@
+---
+title: Resolving Conflicts with Dependency Distribution Policies
+authors:
+  - "@Kexin2000"
+reviewers:
+  - "@XiShanYongYe-Chang"
+approvers:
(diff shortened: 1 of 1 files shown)


RainbowMango:
I can help look at this. Please cc me once it is ready for review.

XiShanYongYe-Chang:
It's ready to review~
cc @RainbowMango

RainbowMango (approved):
/lgtm
/approve

Looks pretty good. Thanks.

PR #6812, opened 2025-10-03
Implement MaxAvailableComponentSets for general estimator

mszacillo:
**What type of PR is this?**
/kind feature

Implements MaxAvailableComponentSets method for general estimator. Currently, this implementation just relies on the cluster resource summary. We will need to file a follow-up task to implement `getMaximumSetsBasedOnResourceModels` which currently is a stub method.

Part of #6734 

Yes

```release-note
Implement MaxAvailableComponentSets for general estimator
```

### DIFF
--- a/pkg/estimator/client/general.go
+++ b/pkg/estimator/client/general.go
@@ -53,15 +53,6 @@ func (ge *GeneralEstimator) MaxAvailableReplicas(_ context.Context, clusters []*
 	return availableTargetClusters, nil
 }
 
-// MaxAvailableComponentSets returns the maximum number of complete multi-component sets (in terms of replicas) that each cluster can host.
-func (ge *GeneralEstimator) MaxAvailableComponentSets(
-	_ context.Context,
-	_ *ComponentSetEstimationRequest) ([]ComponentSetEstimationResponse, error) {
-	// Dummy implementation: return nothing for now
-	// TODO: Implement as part of #6734
-	return nil, nil
-}
-
 func (ge *GeneralEstimator) maxAvailableReplicas(cluster *clusterv1alpha1.Cluster, replicaRequirements *workv1alpha2.ReplicaRequirements) int32 {
 	//Note: resourceSummary must be deep-copied before using in the function to avoid modifying the original data structure.
 	resourceSummary := cluster.Status.ResourceSummary.DeepCopy()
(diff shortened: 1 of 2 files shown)


pkg/estimator/client/general.go
```diff
@@ -102,6 +93,130 @@ func (ge *GeneralEstimator) maxAvailableReplicas(cluster *clusterv1alpha1.Cluste
...
+
+	// Check pod constraint
+	allowedPods := getAllowedPodNumber(resourceSummary)
+	if allowedPods <= 0 {
+		return 0
+	}
+
+	// Upper bound by pods, if set pod requirement is 0 default to allowedPodNumber
+	if setPodNumber <= 0 {
+		return int32(allowedPods) // #nosec G115: integer overflow conversion int64 -> int32
```
mszacillo, line 125: This was an edge case I was unsure of how to handle. I can't think of cases where a CRD with components would have zero replicas (or pods) to schedule... unless the user explicitly wants the components to be scaled down to 0, and then scale up on a CRD update, which would defeat the purpose of the initial estimation.

pkg/estimator/client/general.go
```diff
@@ -102,6 +93,130 @@ func (ge *GeneralEstimator) maxAvailableReplicas(cluster *clusterv1alpha1.Cluste
...
+}
+
+// getMaximumSetsBasedOnClusterSummary checks aggregate resource limits.
+func getMaximumSetsBasedOnClusterSummary(resourceSummary *clusterv1alpha1.ResourceSummary, setCPU, setMem int64) int64 {
+	availCPU := availableQuantity(resourceSummary, corev1.ResourceCPU)
+	availMem := availableQuantity(resourceSummary, corev1.ResourceMemory)
+
+	switch {
+	case setCPU > 0 && setMem > 0:
+		return min64(availCPU/setCPU, availMem/setMem)
```
zhzhuang-zju:
```suggestion
		return min(availCPU/setCPU, availMem/setMem)
```

We can directly use the Go built-in function `min`, which was introduced in Go v1.21.0.

pkg/estimator/client/general.go
```diff
@@ -102,6 +93,130 @@ func (ge *GeneralEstimator) maxAvailableReplicas(cluster *clusterv1alpha1.Cluste
...
+	}
+	return allocatable.Value() // bytes for memory, count for pods, etc.
+}
+
+func min64(a, b int64) int64 {
+	if a < b {
+		return a
+	}
+	return b
+}
```
zhzhuang-zju:
```suggestion
```

pkg/estimator/client/general.go
```diff
@@ -102,6 +93,130 @@ func (ge *GeneralEstimator) maxAvailableReplicas(cluster *clusterv1alpha1.Cluste
...
+	allowedPods := getAllowedPodNumber(resourceSummary)
+	if allowedPods <= 0 {
+		return 0
+	}
+
+	// Upper bound by pods, if set pod requirement is 0 default to allowedPodNumber
+	if setPodNumber <= 0 {
+		return int32(allowedPods) // #nosec G115: integer overflow conversion int64 -> int32
+	}
+	maxSets := getAllowedPodNumber(resourceSummary) / setPodNumber
```
zhzhuang-zju:
```suggestion
	maxSets := allowedPods / setPodNumber
```

pkg/estimator/client/general.go
```diff
@@ -102,6 +93,130 @@ func (ge *GeneralEstimator) maxAvailableReplicas(cluster *clusterv1alpha1.Cluste
...
+	if num < maxSets {
+		maxSets = num
+	}
+
+	// Optional refinement with ResourceModels (if feature gate enabled)
+	if features.FeatureGate.Enabled(features.CustomizedClusterResourceModeling) && len(cluster.Status.ResourceSummary.AllocatableModelings) > 0 {
+		num, err := getMaximumSetsBasedOnResourceModels(cluster, components)
+		if err == nil && num < maxSets {
+			maxSets = num
+		}
```
zhzhuang-zju, line 157: When err is not nil, it's best to have logs for easier tracking.

pkg/estimator/client/general.go
```diff
@@ -102,6 +93,130 @@ func (ge *GeneralEstimator) maxAvailableReplicas(cluster *clusterv1alpha1.Cluste
...
+	}
+	maxSets := getAllowedPodNumber(resourceSummary) / setPodNumber
+
+	// Fallback to allowedPodNumber if there is no resourceRequirement
+	if setCPU == 0 && setMem == 0 {
+		return int32(maxSets) // #nosec G115: integer overflow conversion int64 -> int32
+	}
+
+	// Upper bound by aggregate CPU/Memory
+	num := getMaximumSetsBasedOnClusterSummary(resourceSummary, setCPU, setMem)
```
zhzhuang-zju: The function `getMaximumSetsBasedOnClusterSummary` calculates the maximumSets based on CPU and memory, which excludes other types of resources. Could we consider referencing or directly reusing the function `getMaximumReplicasBasedOnClusterSummary`, which hides the resource names and is more inclusive?

pkg/estimator/client/general.go
```diff
@@ -102,6 +93,130 @@ func (ge *GeneralEstimator) maxAvailableReplicas(cluster *clusterv1alpha1.Cluste
...
+	for i, cluster := range req.Clusters {
+		maxComponentSets := ge.maxAvailableComponentSets(cluster, req.Components)
+		responses[i] = ComponentSetEstimationResponse{Name: cluster.Name, Sets: maxComponentSets}
+	}
+	return responses, nil
+}
+
+func (ge *GeneralEstimator) maxAvailableComponentSets(cluster *clusterv1alpha1.Cluster, components []*workv1alpha2.Component) int32 {
+	// Compute per-set aggregate requirement (e.g. CPU, Memory)
+	setCPU, setMem, setPodNumber := perSetRequirement(components)
```
RainbowMango: Enumerating each resource type might not be a good idea, as we don't know the resource name exactly, like `CPU`, `Cpu`, and `cpu` all mean CPU resources; in addition, it is hard to support extended resources like `nvidia.com/gpu`, or customized resources like `bloomberg.com/gpu`.
We need a resource-type agnostic algorithm, a draft idea in my mind like:
```golang
	var availableSets int32
	resourceSummary := cluster.Status.ResourceSummary.DeepCopy()
	for {
		for _, component := range components {
			for requestedResource, requestedValue := range component.ReplicaRequirements {
				// let resourceSummary - (requestedResource * component.Replicas)
				// if no more resource left, break the loop
			}
		}

		// check whether to break in case of no resource left.

		// increase the count, and run next loop
		availableSets++
	}
```

mszacillo: That makes sense! I'll change this.

mszacillo:
I've updated the algorithm to be resource type agnostic - for unit calculation I just rely on the quantity's format. Please let me know if you have additional questions or concerns.

pkg/estimator/client/general_test.go
```diff
@@ -800,3 +800,209 @@ func TestMinimumModelIndex(t *testing.T) {
...
+						Allocated: corev1.ResourceList{
+							corev1.ResourcePods: resource.MustParse("10"),
+						},
+					},
+				},
+			},
+			expected: 0,
+		},
+		{
+			name: "empty component list should return 0",
```
mszacillo: I'm still deciding on this test case. It is possible for CRDs to have no replicas defined (such as a Kyverno clusterpolicy for instance). Since there are no resource requirements defined by this CRD though, we should just return pod limit.

mszacillo: Thinking about this more, I decided to return the max pod allowance instead. If the CRD does not define components or the components themselves do not have resources, then we should just return the max pod allowance for now. Perhaps there is a better way to do this -

pkg/estimator/client/general.go
```diff
@@ -102,6 +93,144 @@ func (ge *GeneralEstimator) maxAvailableReplicas(cluster *clusterv1alpha1.Cluste
...
+// podsInSet computes the total number of pods in the CRD
+func podsInSet(components []*workv1alpha2.Component) int64 {
+	var sum int64
+	for _, c := range components {
+		sum += int64(c.Replicas)
+	}
+	return sum
+}
+
+// perSetRequirement computes aggregate CPU/Memory demand of one set of components.
```
```suggestion
// perSetRequirement computes the aggregate resource(such as CPU, Memory, GPU, etc) demand of one set of components.
```
Just a nit.

pkg/estimator/client/general.go
```diff
@@ -102,6 +93,144 @@ func (ge *GeneralEstimator) maxAvailableReplicas(cluster *clusterv1alpha1.Cluste
...
+	}
+	maxSets = min(maxSets, podBound)
+
+	if features.FeatureGate.Enabled(features.CustomizedClusterResourceModeling) && len(cluster.Status.ResourceSummary.AllocatableModelings) > 0 {
+		num, err := getMaximumSetsBasedOnResourceModels(cluster, components)
+		if err == nil && num < maxSets {
+			maxSets = num
+		} else {
+			klog.Info(err.Error())
+		}
```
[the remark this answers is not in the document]
RainbowMango: +1
The refactored blocker looks great!

pkg/estimator/client/general.go
```diff
@@ -102,6 +93,144 @@ func (ge *GeneralEstimator) maxAvailableReplicas(cluster *clusterv1alpha1.Cluste
...
+			return 0 // no capacity for this resource
+		}
+
+		resBound := resAvail / req
+		if resBound < maxSets {
+			maxSets = resBound
+		}
+	}
+	maxSets = min(maxSets, podBound)
+
```
zhzhuang-zju: How about
```suggestion
	maxSets := podBound
	for resName, req := range perSet {
		if req <= 0 {
			continue
		}

		resAvail := available[resName]
		if resAvail <= 0 {
			return 0 // no capacity for this resource
		}

		resBound := resAvail / req
		if resBound < maxSets {
			maxSets = resBound
		}
	}

```

pkg/estimator/client/general.go
mszacillo: Ah thanks for catching!

RainbowMango: It's no big deal, but please let me know if you plan to make the updates.

mszacillo: Updated!

pkg/estimator/client/general_test.go
RainbowMango: +1

In addition, a resource with an empty component list, probably because it's not a workload. In that case, we should try to skip the replica/replicaSet estimation.

PR #6813, opened 2025-10-09
Reduce unnecessary watch events in the work-status-controller

XiShanYongYe-Chang:
**What type of PR is this?**

/kind cleanup

The `work-status-controller` only needs to watch the creation events of work objects and the updates of the `WorkApplied` status of work objects. It does not need to watch updates events for other fields of the work objects or deletion events.

Additionally, for the `karmada-agent`, it only needs to be concerned with the work of the current cluster.

Part of #6780

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/status/work_status_controller.go
+++ b/pkg/controllers/status/work_status_controller.go
@@ -543,18 +543,13 @@ func (c *WorkStatusController) getSingleClusterManager(cluster *clusterv1alpha1.
 
 // SetupWithManager creates a controller and register to controller manager.
 func (c *WorkStatusController) SetupWithManager(mgr controllerruntime.Manager) error {
-	ctrlBuilder := controllerruntime.NewControllerManagedBy(mgr).Named(WorkStatusControllerName).
+	return controllerruntime.NewControllerManagedBy(mgr).
+		Named(WorkStatusControllerName).
+		For(&workv1alpha1.Work{}, builder.WithPredicates(c.WorkPredicateFunc)).
 		WithOptions(controller.Options{
 			RateLimiter: ratelimiterflag.DefaultControllerRateLimiter[controllerruntime.Request](c.RateLimiterOptions),
-		})
-
-	if c.WorkPredicateFunc != nil {
-		ctrlBuilder.For(&workv1alpha1.Work{}, builder.WithPredicates(c.WorkPredicateFunc))
-	} else {
-		ctrlBuilder.For(&workv1alpha1.Work{})
-	}
-
-	return ctrlBuilder.Complete(c)
+		}).
+		Complete(c)
 }
 
 func (c *WorkStatusController) eventf(object *unstructured.Unstructured, eventType, reason, messageFmt string, args ...interface{}) {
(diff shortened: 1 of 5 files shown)


pkg/util/helper/predicate.go
```diff
@@ -68,6 +65,81 @@ func WorkWithinPushClusterPredicate(mgr controllerruntime.Manager) predicate.Fun
...
+			return false
+		},
+		DeleteFunc: func(deleteEvent event.DeleteEvent) bool {
+			return false
+		},
+		GenericFunc: func(event.GenericEvent) bool {
+			return false
+		},
+	}
+}
```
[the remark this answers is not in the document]
XiShanYongYe-Chang: Thanks, updated.

pkg/controllers/status/work_status_controller.go
```diff
@@ -543,18 +543,13 @@ func (c *WorkStatusController) getSingleClusterManager(cluster *clusterv1alpha1.
 
 // SetupWithManager creates a controller and register to controller manager.
 func (c *WorkStatusController) SetupWithManager(mgr controllerruntime.Manager) error {
-	ctrlBuilder := controllerruntime.NewControllerManagedBy(mgr).Named(WorkStatusControllerName).
+	return controllerruntime.NewControllerManagedBy(mgr).
```
CharlesQQ, line 546: Some update events also need to be ignored.

pkg/controllers/status/work_status_controller.go
```diff
@@ -543,18 +543,13 @@ func (c *WorkStatusController) getSingleClusterManager(cluster *clusterv1alpha1.
...
+		Named(WorkStatusControllerName).
```
CharlesQQ, line 547:
```
func (c *WorkStatusController) SetupWithManager(mgr controllerruntime.Manager) error {
	updateSyncIgnorePred := predicate.Funcs{ 
		UpdateFunc: func(updateEvent event.UpdateEvent) bool { return updateEvent.ObjectNew.GetResourceVersion() != updateEvent.ObjectOld.GetResourceVersion() }, 
	}
		Named(WorkStatusControllerName).
		For(&workv1alpha1.Work{}, builder.WithPredicates(c.WorkPredicateFunc, updateSyncIgnorePred)).
		WithOptions(controller.Options{
			RateLimiter: ratelimiterflag.DefaultControllerRateLimiter[controllerruntime.Request](c.RateLimiterOptions),
		}).
		Complete(c)
}
```

pkg/controllers/status/work_status_controller.go
XiShanYongYe-Chang: Actually, I'm a bit confused. If we don't want to care about periodically synchronized events, then do we really need to initiate periodic synchronization at all?

pkg/controllers/status/work_status_controller.go
XiShanYongYe-Chang: I understand the format you expect. In fact, both revisions are consistent, but there is one difference: the method you recommend is relatively harder to test.

XiShanYongYe-Chang:
The community has not reached a consensus on this matter. I'm so sorry. If new evidence emerges in the future that proves its value, we will reconsider it.
/close

Issue #6587, opened 2025-07-29
[lfx-mentorship-2025-Sept-Nov] Support TrainJob and SparkApplication in Karmada via Default Resource Interpreters


RainbowMango:
**CNCF LFX mentorship**
https://github.com/cncf/mentoring/tree/main/programs/lfx-mentorship/2025/03-Sep-Nov

LFX URL: https://mentorship.lfx.linuxfoundation.org/project/10cff3a7-3993-450f-a385-ebbcbf923805

Time Line: https://github.com/cncf/mentoring/tree/main/programs/lfx-mentorship/2025/03-Sep-Nov#timeline

Mentors: @zhzhuang-zju @RainbowMango 

**What would you like to be added**:
Develop default resource interpreters for [TrainJob](https://github.com/kubeflow/training-operator) and [SparkApplication](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator) workloads in Karmada. The task involves implementing interpreter logic, validating scheduling/propagation, and documenting the process of running these workloads on Karmada. Share the practice and integration details with the Kubeflow and Spark communities.

**Why is this needed**:
Refer to https://github.com/karmada-io/karmada/issues/5115, and https://github.com/karmada-io/karmada/pull/6535

**Iterated Tasks(draft, will update with details)**:
- [ ] Implement default(aka, built-in) resource interpreter in [third-party resourcecustomizations](https://github.com/karmada-io/karmada/tree/master/pkg/resourceinterpreter/default/thirdparty/resourcecustomizations)
    - [ ] For TrainJob
    - [ ] For SparkApplication
- [ ] Make a presentation at [the Karmada community meeting](https://github.com/karmada-io/karmada?tab=readme-ov-file#meeting).
    * Please feel free to add an agenda to [the meeting notes](https://docs.google.com/document/d/1y6YLVC-v7cmVAdbjedoyR5WL0-q45DBRXTvz5_I7bkA/edit).
    * By joining [the Google groups](https://groups.google.com/forum/#!forum/karmada), you will be able to edit the meeting notes. 
- [ ] Write documentation on the Karmada website to demonstrate the use case.
    - [ ] For TrainJob
    - [ ] For SparkApplication
- [ ] Make a presentation at the target community to acknowledge the use case.
    - [ ] Kubeflow community
    - [ ] Spark community

dikshant182004:
[...]

PR #6818, opened 2025-10-10
Implement default resource interpreter in third-party resourcecustomizations for SparkApplication

liaolecheng:
**What type of PR is this?**
/kind feature

Part of #[6587](https://github.com/karmada-io/karmada/issues/6587)

```release-note
`karmada-controller-manager`: Introduced built-in interpreter for SparkApplication.
```

### DIFF
--- a/pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/sparkoperator.k8s.io/v1beta2/SparkApplication/customizations.yaml
+++ b/pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/sparkoperator.k8s.io/v1beta2/SparkApplication/customizations.yaml
@@ -1,0 +1,342 @@
+apiVersion: config.karmada.io/v1alpha1
+kind: ResourceInterpreterCustomization
+metadata:
+  name: declarative-configuration-sparkapplication
+spec:
+  target:
+    apiVersion: sparkoperator.k8s.io/v1beta2
+    kind: SparkApplication
+  customizations:
+    healthInterpretation:
+      luaScript: >
+        function InterpretHealth(observedObj)
+          if not observedObj or
+             not observedObj.status or
+             not observedObj.status.applicationState or
+             not observedObj.status.applicationState.state then
+            return false
+          end
+        
+          -- Only the 'FAILED' state is considered unhealthy. All other states are treated
+          -- as healthy or recoverable.
+          local state = observedObj.status.applicationState.state
+          if state == 'FAILED' then
+            return false
+          end
+          return true
+        end
+    componentResource:
+      luaScript: |
+        local kube = require("kube")
+
+        local function isempty(s)
+          return s == nil or s == ''
+        end
+
+        -- Safe fetch of deeply nested table fields.
+        local function get(obj, path)
+          local cur = obj
+          for i = 1, #path do
+            if cur == nil then
+              return nil
+            end
+            cur = cur[path[i]]
+          end
+          return cur
+        end
+
+        -- Normalize possibly-string numbers with a default.
+        local function to_num(v, default)
+          if v == nil or v == '' then
+            return default
+          end
+          local n = tonumber(v)
+          if n ~= nil then
+            return n
+          end
+          return default
+        end
+
+        -- JSON-safe deep clone: strings/numbers/booleans/tables. Needed to prevent shared table references.
+        local function clone_plain(val, seen)
+          local tv = type(val)
+          if tv ~= "table" then
+            if tv == "string" or tv == "number" or tv == "boolean" or tv == "nil" then
+              return val
+            end
+            return nil
+          end
+          seen = seen or {}
+          if seen[val] then return nil end
+          seen[val] = true
+          local out = {}
+          for k, v in pairs(val) do
+            local tk = type(k)
+            if tk == "string" or tk == "number" then
+              local cv = clone_plain(v, seen)
+              if cv ~= nil then out[k] = cv end
+            end
+          end
+          seen[val] = nil
+          return out
+        end
+
+        local function apply_pod_template(pt_spec, requires)
+          if pt_spec == nil then
+            return
+          end
+
+          local nodeSelector = clone_plain(pt_spec.nodeSelector)
+          local tolerations  = clone_plain(pt_spec.tolerations)
+          local priority     = pt_spec.priorityClassName
+          local hardNodeAffinity = nil
+          if pt_spec.affinity and pt_spec.affinity.nodeAffinity then
+            hardNodeAffinity = clone_plain(pt_spec.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution)
+          end
+
+          -- Only create nodeClaim if there is content
+          if nodeSelector ~= nil or tolerations ~= nil or hardNodeAffinity ~= nil then
+            requires.nodeClaim = requires.nodeClaim or {}
+            requires.nodeClaim.nodeSelector = nodeSelector
+            requires.nodeClaim.tolerations  = tolerations
+            requires.nodeClaim.hardNodeAffinity = hardNodeAffinity
+          end
+
+          if not isempty(priority) then
+            requires.priorityClassName = priority
+          end
+        end
+
+        local MIN_MEMORY_OVERHEAD = "384m"
+        local JVM_DEFAULT_OVERHEAD_FACTOR = 0.10
+        local NON_JVM_DEFAULT_OVERHEAD_FACTOR = 0.40
+
+        local kube_unit_order = { B = 0, Ki = 1, Mi = 2, Gi = 3, Ti = 4, Pi = 5 }
+        local kube_unit_scale = { B = 1, Ki = 1024, Mi = 1024^2, Gi = 1024^3, Ti = 1024^4, Pi = 1024^5 }
+
+        local function parse_java_memory(java_mem_str)
+          local value_str, unit = tostring(java_mem_str):lower():match("^(%d+%.?%d*)([a-z]*)$")
+          local unit_map = {
+            b = "B",
+            kb = "Ki", k = "Ki",
+            mb = "Mi", m = "Mi",
+            gb = "Gi", g = "Gi",
+            tb = "Ti", t = "Ti",
+            pb = "Pi", p = "Pi",
+          }
+          local value = tonumber(value_str)
+          return value, unit_map[unit]
+        end
+        
+        local function convert_unit(value, from_unit, to_unit)
+          if from_unit == to_unit then return value end
+          local scale = kube_unit_scale[from_unit] / kube_unit_scale[to_unit]
+          return value * scale
+        end
+
+        local function smaller_unit(u1, u2)
+          return kube_unit_order[u1] < kube_unit_order[u2] and u1 or u2
+        end
+
+        local function calculate_total_memory_request(memory_str, overhead_str, memory_overhead_factor, min_memory_overhead)
+          local mem_value, mem_unit = parse_java_memory(memory_str)
+          local over_value, over_unit
+
+          if overhead_str then
+            over_value, over_unit = parse_java_memory(overhead_str)
+          else
+            local overhead_calc = mem_value * memory_overhead_factor
+            local min_over_value, min_over_unit = parse_java_memory(min_memory_overhead)
+
+            local smaller_u = smaller_unit(mem_unit, min_over_unit)
+            local overhead_calc_in_smaller = convert_unit(overhead_calc, mem_unit, smaller_u)
+            local min_over_in_smaller = convert_unit(min_over_value, min_over_unit, smaller_u)
+
+            over_value, over_unit = math.max(overhead_calc_in_smaller, min_over_in_smaller), smaller_u
+          end
+
+          local final_unit = smaller_unit(mem_unit, over_unit)
+          local mem_in_final = convert_unit(mem_value, mem_unit, final_unit)
+          local over_in_final = convert_unit(over_value, over_unit, final_unit)
+          local total_value = mem_in_final + over_in_final
+
+          return string.format("%f%s", total_value, final_unit)
+        end
+
+        function GetComponents(observedObj)
+          local components = {}
+
+          local job_type = get(observedObj, {"spec","type"})
+          local job = job_type:lower()
+          local is_jvm_job = (job == "java" or job == "scala")
+          local default_overhead_factor = is_jvm_job and JVM_DEFAULT_OVERHEAD_FACTOR or NON_JVM_DEFAULT_OVERHEAD_FACTOR
+          local memory_overhead_factor = to_num(get(observedObj, {"spec", "memoryOverheadFactor"}), default_overhead_factor)
+
+          -- Driver
+          local drv_replicas = 1  -- Spark Driver always has 1 instance
+          local drv_requires = {
+            resourceRequest = {}
+          }
+
+          local drv_cpu    = get(observedObj, {"spec","driver","cores"}) or 1
+          local drv_memory_str = get(observedObj, {"spec", "driver", "memory"}) or "1g"
+          local drv_overhead_str = get(observedObj, {"spec", "driver", "memoryOverhead"})
+          drv_requires.resourceRequest.cpu    = drv_cpu
+          drv_requires.resourceRequest.memory = calculate_total_memory_request(drv_memory_str, drv_overhead_str, memory_overhead_factor, MIN_MEMORY_OVERHEAD)
+
+          local drv_gpu = get(observedObj, {"spec","driver","gpu"})
+          if drv_gpu ~= nil then
+            local gpu_name = drv_gpu.name
+            local gpu_qty = to_num(drv_gpu.quantity, 0)
+            if not isempty(gpu_name) and gpu_qty > 0 then
+              drv_requires.resourceRequest[gpu_name] = gpu_qty
+            end
+          end
+    
+          apply_pod_template(get(observedObj, {"spec","driver"}), drv_requires)
+
+          local driverComponent = {
+            name = "driver",
+            replicas = drv_replicas,
+            replicaRequirements = drv_requires
+          }
+          table.insert(components, driverComponent)
+
+          -- Executor
+          local exe_replicas = to_num(get(observedObj, {"spec","executor","instances"}), 1)
+          local dynamic_enabled = get(observedObj, {"spec", "dynamicAllocation", "enabled"}) or false
+          if dynamic_enabled then
+            local initial_exec = to_num(get(observedObj, {"spec", "dynamicAllocation", "initialExecutors"}), 1)
+            local min_exec = to_num(get(observedObj, {"spec", "dynamicAllocation", "minExecutors"}), 1)
+            if initial_exec > exe_replicas then
+              exe_replicas = initial_exec
+            end
+            if min_exec > exe_replicas then
+              exe_replicas = min_exec
+            end
+          end
+
+          local exe_requires = {
+            resourceRequest = {}
+          }
+
+          local exe_cpu    = get(observedObj, {"spec","executor","cores"}) or 1
+          local exe_memory_str = get(observedObj, {"spec", "executor", "memory"}) or "1g"
+          local exe_overhead_str = get(observedObj, {"spec", "executor", "memoryOverhead"})
+          exe_requires.resourceRequest.cpu    = exe_cpu
+          exe_requires.resourceRequest.memory = calculate_total_memory_request(exe_memory_str, exe_overhead_str, memory_overhead_factor, MIN_MEMORY_OVERHEAD)
+
+          local exe_gpu = get(observedObj, {"spec","executor","gpu"})
+          if exe_gpu ~= nil then
+            local gpu_name = exe_gpu.name
+            local gpu_qty = to_num(exe_gpu.quantity, 0)
+            if not isempty(gpu_name) and gpu_qty > 0 then
+              exe_requires.resourceRequest[gpu_name] = gpu_qty
+            end
+          end
+
+          apply_pod_template(get(observedObj, {"spec","executor"}), exe_requires)
+
+          local executorComponent = {
+            name = "executor",
+            replicas = exe_replicas,
+            replicaRequirements = exe_requires
+          }
+          table.insert(components, executorComponent)
+ 
+          return components
+        end
+    statusAggregation:
+      luaScript: >
+        function AggregateStatus(desiredObj, statusItems)
+          if statusItems == nil then
+            return desiredObj
+          end
+          if desiredObj.status == nil then
+            desiredObj.status = {}
+          end
+
+          if #statusItems == 1 then
+            desiredObj.status = statusItems[1].status
+            return desiredObj
+          end
+
+          local statePriority = {
+            ["UNKNOWN"] = 0,
+            [""] = 0,
+            ["SUBMITTED"] = 1,
+            ["PENDING_RERUN"] = 1,
+            ["COMPLETED"] = 2,
+            ["SUSPENDING"] = 3,
+            ["SUSPENDED"] = 3,
+            ["RUNNING"] = 4,
+            ["RESUMING"] = 4,
+            ["SUCCEEDING"] = 4,
+            ["INVALIDATING"] = 5,
+            ["FAILING"] = 6,
+            ["FAILED"] = 6,
+            ["SUBMISSION_FAILED"] = 6,
+          }
+
+          local applicationState = {}
+          local executionAttempts = 0
+          local executorState = {}
+          local submissionAttempts = 0
+          local lastSubmissionAttemptTime = ""
+          local terminationTime = ""
+          local worstPriority = -1
+
+          for i = 1, #statusItems do
+            local currentStatus = statusItems[i].status
+            if currentStatus ~= nil then
+              if currentStatus.applicationState ~= nil then
+                local s = currentStatus.applicationState.state
+                local p = statePriority[s]
+                if p > worstPriority then
+                  worstPriority = p
+                  applicationState = currentStatus.applicationState
+                end
+              end
+              executionAttempts = executionAttempts + (currentStatus.executionAttempts or 0)
+              submissionAttempts = submissionAttempts + (currentStatus.submissionAttempts or 0)
+              if currentStatus.lastSubmissionAttemptTime and (lastSubmissionAttemptTime == "" or currentStatus.lastSubmissionAttemptTime > lastSubmissionAttemptTime) then
+                lastSubmissionAttemptTime = currentStatus.lastSubmissionAttemptTime
+              end
+              if currentStatus.terminationTime and (terminationTime == "" or currentStatus.terminationTime > terminationTime) then
+                terminationTime = currentStatus.terminationTime
+              end
+              if currentStatus.executorState ~= nil then
+                for exec, state in pairs(currentStatus.executorState) do
+                  executorState[exec] = state
+                end
+              end
+            end
+          end
+
+          desiredObj.status.applicationState = applicationState
+          desiredObj.status.executionAttempts = executionAttempts
+          desiredObj.status.executorState = executorState
+          desiredObj.status.submissionAttempts = submissionAttempts
+          desiredObj.status.lastSubmissionAttemptTime = lastSubmissionAttemptTime
+          desiredObj.status.terminationTime = terminationTime
+          return desiredObj
+        end
+    statusReflection:
+      luaScript: >
+        function ReflectStatus(observedObj)
+          local status = {}
+          if observedObj == nil or observedObj.status == nil then
+            return status
+          end
+
+          status.applicationState = observedObj.status.applicationState
+          status.driverInfo = observedObj.status.driverInfo
+          status.executorState = observedObj.status.executorState
+          status.sparkApplicationId = observedObj.status.sparkApplicationId
+          status.lastSubmissionAttemptTime = observedObj.status.lastSubmissionAttemptTime
+          status.submissionAttempts = observedObj.status.submissionAttempts
+          status.executionAttempts = observedObj.status.executionAttempts
+          status.submissionID = observedObj.status.submissionID
+          status.terminationTime = observedObj.status.terminationTime
+          return status
+        end
--- a/pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/sparkoperator.k8s.io/v1beta2/SparkApplication/customizations_tests.yaml
+++ b/pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/sparkoperator.k8s.io/v1beta2/SparkApplication/customizations_tests.yaml
@@ -1,0 +1,10 @@
+tests:
+  - desiredInputPath: testdata/desired-sparkapplication.yaml
+    statusInputPath: testdata/status-file.yaml
+    operation: AggregateStatus
+  - desiredInputPath: testdata/desired-sparkapplication.yaml
+    operation: InterpretComponent
+  - observedInputPath: testdata/observed-sparkapplication.yaml
+    operation: InterpretHealth
+  - observedInputPath: testdata/observed-sparkapplication.yaml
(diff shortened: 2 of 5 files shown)


pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/sparkoperator.k8s.io/v1beta2/SparkApplication/customizations.yaml
```diff
@@ -0,0 +1,272 @@
...
+  customizations:
+    healthInterpretation:
+      luaScript: >
+        function InterpretHealth(observedObj)
+          if observedObj and observedObj.status and observedObj.status.applicationState and observedObj.status.applicationState.state then
+            local state = observedObj.status.applicationState.state
+            return state ~= 'PENDING' and state ~= 'UNKNOWN'
+          end
+          return false
+        end
```
RainbowMango, line 27:
```suggestion
        function InterpretHealth(observedObj)
          if not observedObj or
             not observedObj.status or
             not observedObj.status.applicationState or
             not observedObj.status.applicationState.state then
            return false
          end
        
          -- Only the 'FAILED' state is considered unhealthy. All other states are treated
          -- as healthy or recoverable.
          local state = observedObj.status.applicationState.state
          if state == 'FAILED' then
            return false
          end
          
          return true
```

liaolecheng:
@RainbowMango I've revised the code, and it's now ready for review. I've been quite busy with other matters recently, but I’ll make sure to provide the test report (detailing how to test it) within the next couple of days. Thank you for your understanding and patience!

liaolecheng:
# Test Report

## E2E Testing

### Test Steps
1. Apply the SparkApplication CRD on the Karmada control plane.
2. Install SparkOperator on the member1 clusters respectively.
3. Submit the `spark-pi.yaml` file on the Karmada control plane. The file content is as follows:
````y
apiVersion: "sparkoperator.k8s.io/v1beta2"
kind: SparkApplication
metadata:
  name: spark-pi
  namespace: default
spec:
  type: Scala
  mode: cluster
  image: "spark:3.5.0"
  imagePullPolicy: Always
  mainClass: org.apache.spark.examples.SparkPi
  mainApplicationFile: "local:///opt/spark/examples/jars/spark-examples_2.12-3.5.0.jar"
  sparkVersion: "3.5.0"
  sparkUIOptions:
    serviceLabels:
      test-label/v1: 'true'
  restartPolicy:
    type: Never
  volumes:
    - name: "test-volume"
      hostPath:
        path: "/tmp"
        type: Directory
  driver:
    cores: 2
    memory: "512m"
    labels:
      version: 3.5.0
    serviceAccount: spark-operator-spark
    volumeMounts:
      - name: "test-volume"
        mountPath: "/tmp"
  executor:
    cores: 1
    instances: 2
    memory: "512m"
    labels:
      version: 3.5.0
    volumeMounts:
      - name: "test-volume"
        mountPath: "/tmp"
````
4. Submit the `spark-pi-pp.yaml` file on the Karmada control plane. The file content is as follows:
````y
apiVersion: policy.karmada.io/v1alpha1​
kind: PropagationPolicy​
metadata:​
  name: spark-pi-pp​
  namespace: default ​
spec:​
  resourceSelectors:​
    - apiVersion: sparkoperator.k8s.io/v1beta2 ​
      kind: SparkApplication                    ​
      name: spark-pi                             ​
  placement:​
    clusterAffinity:​
      clusterNames:​
        - member1​
````
### Test Result
1. Check the to verify the correctness of multi-component resource parsing.
image

2. Confirm that the `aggregatedStatus` correctly summarizes the health of resources across member clusters:
image

3. Check the spark-pi application on the Karmada control plane and member cluster respectively to verify the correctness of status parsing.
image

## Unit Testing
### Test Steps
1. Modify `testdata/desired-sparkapplication.yaml` to add GPU resources:
````y
apiVersion: "sparkoperator.k8s.io/v1beta2"
kind: SparkApplication
metadata:
  name: spark-pi
  namespace: default
spec:
  type: Scala
  mode: cluster
  image: "spark:3.5.0"
  imagePullPolicy: Always
  mainClass: org.apache.spark.examples.SparkPi
  mainApplicationFile: "local:///opt/spark/examples/jars/spark-examples_2.12-3.5.0.jar"
  sparkVersion: "3.5.0"
  sparkUIOptions:
    serviceLabels:
      test-label/v1: 'true'
  restartPolicy:
    type: Never
  volumes:
    - name: "test-volume"
      hostPath:
        path: "/tmp"
        type: Directory
  driver:
    cores: 2
    memory: "512m"
    gpu:
      name: "amd.com/gpu"
      quantity: 1 
    labels:
      version: 3.5.0
    serviceAccount: spark-operator-spark
    volumeMounts:
      - name: "test-volume"
        mountPath: "/tmp"
  executor:
    cores: 1
    instances: 2
    memory: "512m"
    gpu:
      name: "nvidia.com/gpu"
      quantity: 1
    labels:
      version: 3.5.0
    volumeMounts:
      - name: "test-volume"
        mountPath: "/tmp"
````
2. Use the following command to verify if GPU resources can be correctly interpreted:
````bash
karmadactl interpret -f customizations.yaml \
  ----desired-file testdata/desired-sparkapplication.yaml \
  --operation InterpretComponent
````
### Test Result
image

**Comment by zhzhuang-zju** [comment]
Thanks for your detailed test report!
/assign

File: pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/sparkoperator.k8s.io/v1beta2/SparkApplication/customizations.yaml [hunk 2]
```diff
@@ -0,0 +1,241 @@
...
+
+          local driverComponent = {
+            name = "driver",
+            replicas = drv_replicas,
+            replicaRequirements = drv_requires
+          }
+          table.insert(components, driverComponent)
+
+          -- Executor
+          local exe_replicas = to_num(get(observedObj, {"spec","executor","instances"}), 1)
```
zhzhuang-zju, line 206: The Spark Operator submits Spark applications to the cluster for execution via `spark-submit --args`. The initial number of executor replicas is determined by taking the **maximum value** among the following three arguments:

- `spark.dynamicAllocation.initialExecutors`
- `spark.dynamicAllocation.minExecutors`
- `spark.executor.instances`

These arguments correspond to the following fields in the `SparkApplication` API:

- **`spark.executor.instances`**  
  The value is taken from `spec.executor.instances` (preferred) or `spec.sparkConf["spark.executor.instances"]`.  
  The defaulting logic for `spec.executor.instances` is implemented here:  
  https://github.com/kubeflow/spark-operator/blob/8044286170695851c1c3b2b4f9ac357fc737c0fd/internal/webhook/sparkapplication_defaulter.go#L89-L94

- **`spark.dynamicAllocation.initialExecutors`**  
  The value is taken from `spec.dynamicAllocation.initialExecutors` (preferred) or `spec.sparkConf["spark.dynamicAllocation.initialExecutors"]`.

- **`spark.dynamicAllocation.minExecutors`**  
  The value is taken from `spec.dynamicAllocation.minExecutors` (preferred) or `spec.sparkConf["spark.dynamicAllocation.minExecutors"]`.

Note that `spark.dynamicAllocation.initialExecutors` and `spark.dynamicAllocation.minExecutors` **only take effect when dynamic allocation is enabled** (i.e., `spec.dynamicAllocation.enabled: true` or `spec.sparkConf["spark.dynamicAllocation.enabled"]`).

Therefore, the final initial executor count is the result of **collaborative resolution among multiple configuration sources**.

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/sparkoperator.k8s.io/v1beta2/SparkApplication/customizations.yaml
```diff
@@ -0,0 +1,241 @@
...
+          local exe_requires = {
+            resourceRequest = {}
+          }
+
+          local exe_cpu    = get(observedObj, {"spec","executor","cores"})
```
zhzhuang-zju: Like `exe_replicas`, the values of `exe_cpu`, `drv_cpu`, `exe_memory`, and `drv_memory` are also controlled by multiple parameters. The parameters configurable via `spec.sparkConf` can be found at: https://spark.apache.org/docs/latest/running-on-kubernetes.html.

liaolecheng:
@zhzhuang-zju Thank you for pointing this out. Previously, I only considered the static configuration of SparkApplication API fields. I will revise the solution to parse this logic properly.

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/sparkoperator.k8s.io/v1beta2/SparkApplication/customizations.yaml
zhzhuang-zju: @liaolecheng I think we can refer to how the Spark operator calculates resource requests for drivers and executors. However, it's puzzling that when calculating the resource quota for Spark pods, `spec.sparkConf` is completely ignored. I'm not sure what their considerations were, but let's start by referencing their implementation.

https://github.com/kubeflow/spark-operator/blob/8044286170695851c1c3b2b4f9ac357fc737c0fd/internal/webhook/resourcequota.go#L52

liaolecheng: I’ve looked into the workflow of the SparkOperator. Users typically submit the `spark-pi.yaml` file, and `spark-submit` is invoked by the controller. `sparkConf` aligns with some original parameters. Since the YAML file already has existing fields for configuration, configurations are typically not placed in `sparkConf`.What’s your take on this? https://www.kubeflow.org/docs/components/spark-operator/overview/

zhzhuang-zju: This is a good consideration. Perhaps for this reason, the spark-operator-webhook ignores the sparkConf configuration when calculating the resource requests for SparkApplications. As I mentioned in the discussion at https://github.com/karmada-io/karmada/pull/6818#discussion_r2434797880, we can start by referencing the implementation of the spark-operator-webhook.

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/sparkoperator.k8s.io/v1beta2/SparkApplication/customizations.yaml
```diff
@@ -0,0 +1,328 @@
...
+                  applicationState = currentStatus.applicationState
+                  completionTime = currentStatus.completionTime
+                  driverInfo = currentStatus.driverInfo
+                  executionAttempts = currentStatus.executionAttempts
+                  executorState = currentStatus.executorState
+                  submissionAttempts = currentStatus.submissionAttempts
+                  lastSubmissionAttemptTime = currentStatus.lastSubmissionAttemptTime
+                  submissionID = currentStatus.submissionID
+                  terminationTime = currentStatus.terminationTime
+              end
```
zhzhuang-zju, line 304: Currently, the implementation only aggregates the status of a SparkApplication from a single member cluster to the control plane, while the statuses from other member clusters are lost. Is this approach appropriate?

liaolecheng: In my opinion, under the current scheduling mechanism, a SparkApplication is only scheduled to one cluster. Additionally, these statuses (such as IDs and running states) pertain only to a single instance and do not align with aggregation semantics. What do you think about this?

zhzhuang-zju: Yes, that's how it appears from the current implementation. I've observed two approaches for implementing AggregatedStatus for multi-component workloads:

- Taking the status from a single cluster (simpler to implement but less scalable)
- Aggregating status from multiple clusters (more complex to implement but more scalable)

I'm actually hesitant between these two options myself. However, shouldn't we standardize this? Currently, different multi-component workloads are implemented differently. cc @RainbowMango @XiShanYongYe-Chang

XiShanYongYe-Chang: My choice is the second option, and in previous implementations, I think it should also be the second approach. That's why it's called "aggregate." I think we should try to aggregate and display information from different clusters as much as possible. If a certain field or element cannot be aggregated, I think it should be temporarily discarded for the current API.

liaolecheng: Perhaps none of these are particularly suitable for aggregation—like averaging times, summing counts, or using the last submission. If designers’ understanding of "aggregation" differs from users’, would that potentially lead to misunderstandings about the current status?

zhzhuang-zju: My thinking basically aligns with @XiShanYongYe-Chang.
For counter-type metrics, we can sum them up.
For timing-related metrics, we should take the timestamp with the latest time.
And for non-aggregatable fields like `sparkApplicationId`, we can temporarily discard them.
WDYT？ @RainbowMango

liaolecheng: I agree with that. Here’s how I’ve implemented it: If it’s a single cluster, we just replicate the data. For multiple clusters, we follow the rules we discussed—summing counter metrics, taking the latest timestamp for timing metrics, temporarily dropping non-aggregatable fields, and for status-related fields, take the worst status.

zhzhuang-zju: > If it’s a single cluster, we just replicate the data. For multiple clusters, we follow the rules we discussed—summing counter metrics, taking the latest timestamp for timing metrics, temporarily dropping non-aggregatable fields, and for status-related fields, take the worst status.

Totally agree

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/sparkoperator.k8s.io/v1beta2/SparkApplication/customizations.yaml
```diff
@@ -0,0 +1,365 @@
...
+                if currentStatus.terminationTime and (terminationTime == "" or currentStatus.terminationTime > terminationTime) then
+                    terminationTime = currentStatus.terminationTime
+                end
+              end
+            end
+          end
+
+          desiredObj.status.sparkApplicationId = sparkApplicationId
+          desiredObj.status.applicationState = applicationState
+          desiredObj.status.completionTime = completionTime
```
zhzhuang-zju: @liaolecheng Please confirm whether `status.completionTime` actually exists—it doesn't appear to be defined in the SparkApplication API.

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/sparkoperator.k8s.io/v1beta2/SparkApplication/customizations.yaml
```diff
@@ -0,0 +1,365 @@
...
+              if currentStatus ~= nil and currentStatus.applicationState ~= nil then
+                local s = currentStatus.applicationState.state
+                local p = statePriority[s]
+                if p > worstPriority then
+                  worstPriority = p
+                  applicationState = currentStatus.applicationState
+                end
+                executionAttempts = executionAttempts + (currentStatus.executionAttempts or 0)
+                submissionAttempts = submissionAttempts + (currentStatus.submissionAttempts or 0)
+                if currentStatus.lastSubmissionAttemptTime and (lastSubmissionAttemptTime == "" or currentStatus.lastSubmissionAttemptTime > lastSubmissionAttemptTime) then
```
zhzhuang-zju: When the length of `statusItems` is not 1, can we aggregate `status.executorState`? It is of type `map[string]string` and appears to be aggregatable.

zhzhuang-zju:
Great job~ others LGTM

zhzhuang-zju:
/lgtm
cc @RainbowMango

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/sparkoperator.k8s.io/v1beta2/SparkApplication/customizations.yaml
```diff
@@ -0,0 +1,367 @@
...
+
+          for _, u in ipairs(kube_units) do
+            if total_bytes >= u.threshold then
+              local scaled = total_bytes / u.threshold
+              if u.suffix == "Mi" then
+                return string.format("%d%s", math.floor(scaled), u.suffix)
+              end
+              if scaled == math.floor(scaled) then
+                return string.format("%d%s", scaled, u.suffix)
+              end
```
RainbowMango: Can you remind me why we need to handle `Mi` specially?

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/sparkoperator.k8s.io/v1beta2/SparkApplication/customizations.yaml
```diff
@@ -0,0 +1,367 @@
...
+
+          desiredObj.status.sparkApplicationId = sparkApplicationId
+          desiredObj.status.applicationState = applicationState
+          desiredObj.status.driverInfo = driverInfo
+          desiredObj.status.executionAttempts = executionAttempts
+          desiredObj.status.executorState = executorState
+          desiredObj.status.submissionAttempts = submissionAttempts
+          desiredObj.status.lastSubmissionAttemptTime = lastSubmissionAttemptTime
+          desiredObj.status.submissionID = submissionID
+          desiredObj.status.terminationTime = terminationTime
```
RainbowMango: All these fields are taken from the first of `statusItems`, right? Even when more than one status item in it.

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/sparkoperator.k8s.io/v1beta2/SparkApplication/customizations.yaml
liaolecheng: After applying a load factor, the memory values often become non-integer numbers. Since Kubernetes expects integer-based resource quantities, The function iterates from larger to smaller units (Ti → Gi → Mi → Ki):

1. If the byte value divides evenly by the current unit (e.g., exactly 2Gi), it returns that integer quantity.
2. If it doesn’t divide evenly, it tries the next smaller unit.
3. When it reaches Mi, it always returns the integer part (e.g., 512.8Mi → 512Mi) to ensure no decimals.
4. If the value is smaller than 1Mi, it falls back to Ki, rounding up to at least 1Ki.

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/sparkoperator.k8s.io/v1beta2/SparkApplication/customizations.yaml
liaolecheng: When there’s only one statusItem, all fields are taken directly from it, but when there are multiple items, the function aggregates them: it selects the worst applicationState, sums up attempts, takes the latest timestamps, merges executorState, and leaves other fields empty or unchanged, rather than simply using the first one.

RainbowMango: Oh, I see. 
But maybe we can return earlier when there is only one item. Like https://github.com/karmada-io/karmada/blob/dee69650b6e14ace8d7a451587368dd89a9adacd/pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/batch.volcano.sh/v1alpha1/Job/customizations.yaml#L123-L126

for better readability.

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/sparkoperator.k8s.io/v1beta2/SparkApplication/customizations.yaml
```diff
@@ -0,0 +1,367 @@
...
+
+          local total_memory_bytes = memory_bytes + overhead_bytes
+          return bytes_to_kube_quantity(total_memory_bytes)
+        end
+
+        function GetComponents(observedObj)
+          local components = {}
+
+          local job_type = get(observedObj, {"spec","type"})
+          local is_jvm_job = job_type:lower() == "java"
```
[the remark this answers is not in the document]
zhzhuang-zju: That's a good question. However, the Spark Operator itself is inconsistent when determining `is_jvm_job`:

- **ResourceQuota**: Only checks whether the type is `"java"`.  
  https://github.com/kubeflow/spark-operator/blob/d0cd834f6a134c79b09633d1fcb5b08ee9a1dd70/internal/webhook/resourcequota.go#L156-L169

- **Yunikorn**: Treats both `"java"` and `"scala"` types as JVM jobs.  
  https://github.com/kubeflow/spark-operator/blob/d0cd834f6a134c79b09633d1fcb5b08ee9a1dd70/internal/scheduler/yunikorn/resourceusage/memory.go#L28-L29

I'll deploy a SparkApplication locally to observe Spark's actual behavior.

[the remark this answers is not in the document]
zhzhuang-zju: @liaolecheng I agree with this comment. I deployed it locally and confirmed that Scala jobs are indeed recognized as JVM jobs.
sparkApplication
```yaml
apiVersion: sparkoperator.k8s.io/v1beta2
kind: SparkApplication
metadata:
  name: spark-pi
  namespace: default
spec:
  type: Scala
  mode: cluster
  image: docker.io/library/spark:4.0.0
  imagePullPolicy: IfNotPresent
  mainClass: org.apache.spark.examples.SparkPi
  mainApplicationFile: local:///opt/spark/examples/jars/spark-examples.jar
  arguments:
    - "5000"
  sparkVersion: 4.0.0
  driver:
    labels:
      version: 4.0.0
    memory: 5000m
    serviceAccount: spark-operator-spark
    securityContext:
      capabilities:
        drop:
          - ALL
      runAsGroup: 185
      runAsUser: 185
      runAsNonRoot: true
      allowPrivilegeEscalation: false
      seccompProfile:
        type: RuntimeDefault
  executor:
    labels:
      version: 4.0.0
    cores: 1
    memory: 512m
    securityContext:
      capabilities:
        drop:
          - ALL
      runAsGroup: 185
      runAsUser: 185
      runAsNonRoot: true
      allowPrivilegeEscalation: false
      seccompProfile:
        type: RuntimeDefault
```
resourcebinding
```yaml
apiVersion: work.karmada.io/v1alpha2
kind: ResourceBinding
metadata:
  annotations:
    policy.karmada.io/applied-placement: '{"clusterAffinity":{"clusterNames":["member4"]}}'
    propagationpolicy.karmada.io/name: spark-pia-9884cdb75
  creationTimestamp: "2025-11-05T02:39:22Z"
  finalizers:
  - karmada.io/binding-controller
  generation: 2
  labels:
    propagationpolicy.karmada.io/permanent-id: 667866a3-7039-4def-b792-74ffadcb2a3e
    resourcebinding.karmada.io/permanent-id: 12ecfd18-b7d3-4cff-9ab0-464a9262efae
  name: spark-pia-sparkapplication
  namespace: default
  ownerReferences:
  - apiVersion: sparkoperator.k8s.io/v1beta2
    blockOwnerDeletion: true
    controller: true
    kind: SparkApplication
    name: spark-pia
    uid: 8a29f9d7-53c0-4941-9599-9d9f5451e32f
  resourceVersion: "2621135"
  uid: 68f3d43d-b783-4538-8da6-56023f7f459a
spec:
  clusters:
  - name: member4
  components:
  - name: driver
    replicaRequirements:
      resourceRequest:
        memory: 7000Mi
    replicas: 1
```
driverpod
```yaml
    resources:
      limits:
        memory: 5500Mi
      requests:
        cpu: "1"
        memory: 5500Mi
```

liaolecheng:
cc @zhzhuang-zju

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/sparkoperator.k8s.io/v1beta2/SparkApplication/customizations.yaml
```diff
@@ -0,0 +1,342 @@
...
+          if desiredObj.status == nil then
+            desiredObj.status = {}
+          end
+
+          if #statusItems == 1 then
+            if statusItems[1].status ~= nil then
+              desiredObj.status = statusItems[1].status
+              return desiredObj
+            end
+          end
```
zhzhuang-zju, line 262:
```suggestion
          if #statusItems == 1 then
            desiredObj.status = statusItems[1].status
            return desiredObj
          end
```

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/sparkoperator.k8s.io/v1beta2/SparkApplication/customizations.yaml
```diff
@@ -0,0 +1,342 @@
...
+          local executionAttempts = 0
+          local executorState = {}
+          local submissionAttempts = 0
+          local lastSubmissionAttemptTime = ""
+          local terminationTime = ""
+          local worstPriority = -1
+
+          for i = 1, #statusItems do
+            local currentStatus = statusItems[i].status
+            if currentStatus ~= nil and currentStatus.applicationState ~= nil then
```
zhzhuang-zju:
```suggestion
            if currentStatus ~= nil then
                if currentStatus.applicationState ~= nil then
```

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/sparkoperator.k8s.io/v1beta2/SparkApplication/customizations.yaml
```diff
@@ -0,0 +1,342 @@
...
+          return value * scale
+        end
+
+        local function smaller_unit(u1, u2)
+          return kube_unit_order[u1] < kube_unit_order[u2] and u1 or u2
+        end
+
+        local function calculate_total_memory_request(memory_str, overhead_str, memory_overhead_factor, min_memory_overhead)
+          local mem_value, mem_unit = parse_java_memory(memory_str)
+          local over_value, over_unit
```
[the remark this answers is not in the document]
zhzhuang-zju: The default CPU for both the driver and executor is 1, and the default memory is 1Gi.

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/sparkoperator.k8s.io/v1beta2/SparkApplication/customizations.yaml
```diff
@@ -0,0 +1,342 @@
...
+          local total_value = mem_in_final + over_in_final
+
+          return string.format("%f%s", total_value, final_unit)
+        end
+
+        function GetComponents(observedObj)
+          local components = {}
+
+          local job_type = get(observedObj, {"spec","type"})
+          local job = job_type:lower()
```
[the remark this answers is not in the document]
liaolecheng: Since `spec.type` is a required field in the SparkApplication CRD, it should never be nil in a valid resource.

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/sparkoperator.k8s.io/v1beta2/SparkApplication/customizations.yaml
```diff
@@ -0,0 +1,342 @@
...
+
+            over_value, over_unit = math.max(overhead_calc_in_smaller, min_over_in_smaller), smaller_u
+          end
+
+          local final_unit = smaller_unit(mem_unit, over_unit)
+          local mem_in_final = convert_unit(mem_value, mem_unit, final_unit)
+          local over_in_final = convert_unit(over_value, over_unit, final_unit)
+          local total_value = mem_in_final + over_in_final
+
+          return string.format("%f%s", total_value, final_unit)
```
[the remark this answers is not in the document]
liaolecheng: Kubernetes can parse values like `563.200000Mi` because its resource.Quantity parser is designed to accept decimal representations as long as they resolve to a valid numeric value with a recognized unit.
```
 karmadactl interpret -f customizations.yaml --desired-file testdata/desired-sparkapplication.yaml --operation InterpretComponent
Executor memory request: 1536.000000Mi
---
# [1/1] components:
- name: driver
  replicaRequirements:
    resourceRequest:
        cpu: "2"
        memory: 896Mi
  replicas: 1
- name: executor
  replicaRequirements:
    resourceRequest:
        cpu: "1"
        memory: 1536Mi
  replicas: 3
```
`1825361100800m` is the normalized decimal representation of `1740.8Mi` produced by resource.Quantity.
It isn’t an error — Kubernetes converts the binary unit (Mi) into a smaller decimal unit (m, meaning 1/1000 of a byte) to preserve precision and eliminate floating-point fractions.
```
karmadactl interpret -f customizations.yaml --desired-file testdata/desired-sparkapplication.yaml --operation InterpretComponent
Executor memory request: 1740.800000Mi
---
# [1/1] components:
- name: driver
  replicaRequirements:
    resourceRequest:
        cpu: "2"
        memory: 896Mi
  replicas: 1
- name: executor
  replicaRequirements:
    resourceRequest:
        cpu: "1"
        memory: 1825361100800m
  replicas: 3
```

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/sparkoperator.k8s.io/v1beta2/SparkApplication/customizations.yaml
```diff
@@ -0,0 +1,342 @@
...
+          end
+        
+          -- Only the 'FAILED' state is considered unhealthy. All other states are treated
+          -- as healthy or recoverable.
+          local state = observedObj.status.applicationState.state
+          if state == 'FAILED' then
+            return false
+          end
+          return true
+        end
```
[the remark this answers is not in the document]
liaolecheng: What do you think about this suggestion? @RainbowMango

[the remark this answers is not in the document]
zhzhuang-zju: > The health check only considers FAILED as an unhealthy state. The SUBMISSION_FAILED state is also a terminal failure and should be treated as unhealthy, which would be consistent with the AggregateStatus logic in this same file.

~~+1. If SUBMISSION_FAILED occurs, it means the Spark operator encountered an issue when creating the Spark application via `spark-submit`. In this case, it's more reasonable to treat it as **unhealthy**.~~

I just noticed that SparkApplication supports retries, and it only transitions to `FAILED` after multiple retry attempts have failed—only then are `SUBMISSION_FAILED` or `FAILING` marked as `FAILED`. Therefore, `SUBMISSION_FAILED` or `FAILING` are not terminal states. From this perspective, it's reasonable to consider these states as healthy.

RainbowMango:
@zhzhuang-zju Do you have any further comments?

zhzhuang-zju:
> @zhzhuang-zju Do you have any further comments?

I have no further comments. Thanks again @liaolecheng for your hard work!

PR #6832, opened 2025-10-15
Reduce unnecessary watch events in cluster-status-controller

XiShanYongYe-Chang:
**What type of PR is this?**

/kind cleanup

The `cluster-status-controller` periodically processes the managed clusters (default interval: 10 seconds). Therefore, after a cluster is created or the `karmada-controller-manager` component is restarted, the controller will periodically collect the cluster status, and there is no need to pay attention to update events.

Part of #6780 

```release-note
NONE
```


XiShanYongYe-Chang:
The community has not reached a consensus on this matter. I'm so sorry. If new evidence emerges in the future that proves its value, we will reconsider it.
/close

PR #6849, opened 2025-10-17
Reduce unnecessary watch events in execution-status-controller

rohan-019:
**Description:**
Stop enqueueing final DeleteEvents for Work to avoid redundant reconciliations (deletion is handled via the controller’s finalizer). Add a new predicate buildWorkEventPredicate() that:

- Triggers on generation changes.
- Triggers when DeletionTimestamp transitions nil → non-nil (so finalizer logic runs).
- Ignores final DeleteEvents.

**Notes:**

- Replaced the generic GenerationChangedPredicate with this explicit predicate.
- If a custom WorkPredicateFunc is provided, ensure it doesn’t block the nil→non-nil deletion transition

**Linked Issue:**
part of #6780

### DIFF
--- a/pkg/controllers/execution/execution_controller.go
+++ b/pkg/controllers/execution/execution_controller.go
@@ -36,6 +36,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	clusterv1alpha1 "github.com/karmada-io/karmada/pkg/apis/cluster/v1alpha1"
@@ -139,10 +140,15 @@ func (c *Controller) SetupWithManager(mgr controllerruntime.Manager) error {
 			RateLimiter: ratelimiterflag.DefaultControllerRateLimiter[controllerruntime.Request](c.RateLimiterOptions),
 		})
 
+	// Ignore delete events; the deletion process will be handled by the deletion timestamp set.
+	deleteIgnorePred := predicate.Funcs{
+		DeleteFunc: func(event.DeleteEvent) bool { return false },
+	}
+
 	if c.WorkPredicateFunc != nil {
-		ctrlBuilder.For(&workv1alpha1.Work{}, builder.WithPredicates(c.WorkPredicateFunc))
+		ctrlBuilder.For(&workv1alpha1.Work{}, builder.WithPredicates(c.WorkPredicateFunc, deleteIgnorePred))
 	} else {
-		ctrlBuilder.For(&workv1alpha1.Work{})
+		ctrlBuilder.For(&workv1alpha1.Work{}, builder.WithPredicates(deleteIgnorePred))
 	}
 
 	return ctrlBuilder.Complete(c)


rohan-019:
Hi @XiShanYongYe-Chang, I hope you are doing well
Please take a look at this!!

pkg/controllers/execution/execution_controller.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2020 The Karmada Authors.
+Copyright 2025 The Karmada Authors.
```
XiShanYongYe-Chang: For existing files, we do not need to modify the time.

pkg/controllers/execution/execution_controller_test.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2024 The Karmada Authors.
+Copyright 2025 The Karmada Authors.
```
XiShanYongYe-Chang: ditto

pkg/controllers/execution/execution_controller.go
```diff
@@ -134,20 +135,43 @@ func (c *Controller) Reconcile(ctx context.Context, req controllerruntime.Reques
 // SetupWithManager creates a controller and register to controller manager.
 func (c *Controller) SetupWithManager(mgr controllerruntime.Manager) error {
 	ctrlBuilder := controllerruntime.NewControllerManagedBy(mgr).Named(ControllerName).
-		WithEventFilter(predicate.GenerationChangedPredicate{}).
```
XiShanYongYe-Chang, line 137: We don't need to remove the general filter, which has already accounted for changes in generation, including changes in deletion timestamps. Therefore, we don't need to alter the filtering for updates; we just need to ignore the deletion events.

pkg/controllers/execution/execution_controller.go
```diff
@@ -134,20 +135,43 @@ func (c *Controller) Reconcile(ctx context.Context, req controllerruntime.Reques
...
 		WithOptions(controller.Options{
 			RateLimiter: ratelimiterflag.DefaultControllerRateLimiter[controllerruntime.Request](c.RateLimiterOptions),
 		})
 
+	// Build the default predicate: allow generation changes and nil->non-nil deletion timestamp transitions; ignore delete events.
+	defaultPred := buildWorkEventPredicate()
```
XiShanYongYe-Chang: How about directly handling the existing `WorkWithinPushClusterPredicate`? For the agent(Pull) mode, a new predicate Funcs could be added.

XiShanYongYe-Chang (commented):
Thanks~

pkg/controllers/execution/execution_controller.go
rohan-019: Thanks for the feedback! I’ve reverted to keeping the existing `GenerationChangedPredicate`, since it already handles generation and deletion-timestamp changes.
Added only a simple `DeleteFunc` to ignore delete events.
For pull mode, it’s combined with `WorkPredicateFunc`; for push mode, with `WorkWithinPushClusterPredicate.`

pkg/controllers/execution/execution_controller.go
```diff
@@ -139,10 +140,18 @@ func (c *Controller) SetupWithManager(mgr controllerruntime.Manager) error {
...
+	}
+
 	if c.WorkPredicateFunc != nil {
-		ctrlBuilder.For(&workv1alpha1.Work{}, builder.WithPredicates(c.WorkPredicateFunc))
+		// Combine custom predicate with delete ignore
+		ctrlBuilder.For(&workv1alpha1.Work{}, builder.WithPredicates(c.WorkPredicateFunc, deleteIgnorePred))
 	} else {
-		ctrlBuilder.For(&workv1alpha1.Work{})
+		// Use push cluster predicate with delete ignore for karmada-controller-manager mode
+		pushClusterPred := helper.WorkWithinPushClusterPredicate(mgr)
```
XiShanYongYe-Chang: Why is this needed?

rohan-019: yes. like when `WorkPredicateFunc` is nil (default), the controller runs in karmada-controller-manager (push) mode and must only process work objects targeting clusters in **push** sync mode. `WorkWithinPushClusterPredicate` filters out works for clusters in **pull** mode, preventing the execution controller from attempting to sync resources that should be handled by agents.

XiShanYongYe-Chang: When `WorkPredicateFunc` is nil, the controller runs in karmada-agent (pull) mode, so here don't need this.

rohan-019: thanks for the clarification! You are right so the push-cluster predicate is not needed. I'll remove it.

pkg/controllers/execution/execution_controller.go
```diff
@@ -139,10 +140,17 @@ func (c *Controller) SetupWithManager(mgr controllerruntime.Manager) error {
 			RateLimiter: ratelimiterflag.DefaultControllerRateLimiter[controllerruntime.Request](c.RateLimiterOptions),
 		})
 
+	// Build predicates: ignore delete events, and handle push/pull cluster filtering
```
XiShanYongYe-Chang: How about this comment: Ignore delete events; the deletion process will be handled by the deletion timestamp set.

pkg/controllers/execution/execution_controller.go
```diff
@@ -139,10 +140,17 @@ func (c *Controller) SetupWithManager(mgr controllerruntime.Manager) error {
...
+	deleteIgnorePred := predicate.Funcs{
+		DeleteFunc: func(event.DeleteEvent) bool { return false },
+	}
+
 	if c.WorkPredicateFunc != nil {
-		ctrlBuilder.For(&workv1alpha1.Work{}, builder.WithPredicates(c.WorkPredicateFunc))
+		// Combine custom predicate with delete ignore for karmada-controller-manager (push) mode
```
XiShanYongYe-Chang: I think this comment can be removed.

pkg/controllers/execution/execution_controller.go
```diff
@@ -139,10 +140,17 @@ func (c *Controller) SetupWithManager(mgr controllerruntime.Manager) error {
...
+		ctrlBuilder.For(&workv1alpha1.Work{}, builder.WithPredicates(c.WorkPredicateFunc, deleteIgnorePred))
 	} else {
-		ctrlBuilder.For(&workv1alpha1.Work{})
+		// karmada-agent (pull) mode - just ignore delete events
```
XiShanYongYe-Chang: I think this comment can be removed.

pkg/controllers/execution/execution_controller_predicate_test.go
```diff
@@ -0,0 +1,61 @@
+/*
+Copyright 2025 The Karmada Authors.
```
XiShanYongYe-Chang: Maybe we don't need this test file.

XiShanYongYe-Chang (commented):
Thanks~

XiShanYongYe-Chang:
Thanks @rohan-019 LGTM
Can you help squash the commits into one?

XiShanYongYe-Chang:
Thanks @rohan-019 I'm sorry for the delay in response.
/lgtm
/cc @zach593  @RainbowMango

zach593:
/lgtm

Just one more thing: the title and commit message need to be updated. 
"execution-status-controller" → "execution-controller" should be fine.

pkg/controllers/execution/execution_controller.go
```diff
@@ -139,10 +140,15 @@ func (c *Controller) SetupWithManager(mgr controllerruntime.Manager) error {
 			RateLimiter: ratelimiterflag.DefaultControllerRateLimiter[controllerruntime.Request](c.RateLimiterOptions),
 		})
 
+	// Ignore delete events; the deletion process will be handled by the deletion timestamp set.
+	deleteIgnorePred := predicate.Funcs{
+		DeleteFunc: func(event.DeleteEvent) bool { return false },
+	}
+
```
CharlesQQ, line 147: // Ignore update events that have no change in the object's resource version.
	updateSyncIgnorePred := predicate.Funcs{
		UpdateFunc: func(updateEvent event.UpdateEvent) bool {
			return updateEvent.ObjectNew.GetResourceVersion() != updateEvent.ObjectOld.GetResourceVersion()
		},
	}

	if c.WorkPredicateFunc != nil {
		ctrlBuilder.For(&workv1alpha1.Work{}, builder.WithPredicates(c.WorkPredicateFunc, deleteIgnorePred, updateSyncIgnorePred))
	} else {
		ctrlBuilder.For(&workv1alpha1.Work{}, builder.WithPredicates(deleteIgnorePred,updateSyncIgnorePred))
	}

CharlesQQ: Some update events also need to be ignored.

XiShanYongYe-Chang: As long as an update event occurs, the RV should change, right?

CharlesQQ: yes, however, when the Informer cache is periodically synced, `rv` is not changed.

XiShanYongYe-Chang: Yes, you're right. Currently, our controllers are not set up for periodic synchronization. If we need to set it up later, I think we can consider it together.

XiShanYongYe-Chang:
Hi @rohan-019 

PR #6864, opened 2025-10-21
Add `EstimateComponentsPlugins` support to estimator framework

seanlaii:
**What type of PR is this?**
/kind feature

Add `EstimateComponentsPlugins` support to estimator framework to allow customizing scheduling logic for Components.

Part of #6734

```release-note
`karmada-scheduler-estimator`: Allow adding plugins in estimator for component scheduling.
`karmada-scheduler-estimator`: Deprecated the `Estimator` label value for `estimating_plugin_extension_point` in the `estimating_plugin_execution_duration_seconds` and `estimating_plugin_extension_point_duration_seconds` metrics. Use `estimate_replicas` instead; `Estimator` will be removed in a future release. To avoid double counting during the deprecation period, update PromQL to filter on `estimating_plugin_extension_point="estimate_replicas"` (or exclude `Estimator`).
```

### DIFF
--- a/pkg/estimator/server/framework/runtime/framework.go
+++ b/pkg/estimator/server/framework/runtime/framework.go
@@ -34,15 +34,20 @@ import (
 )
 
 const (
-	estimator = "Estimator"
+	// estimator is the legacy label for replica estimation metrics.
+	// Deprecated: Use estimateReplicasExtension instead. This will be removed in a future release.
+	estimator                   = "Estimator"
+	estimateReplicasExtension   = "estimate_replicas"
+	estimateComponentsExtension = "estimate_components"
 )
 
 // frameworkImpl implements the Framework interface and is responsible for initializing and running scheduler
 // plugins.
 type frameworkImpl struct {
-	estimateReplicasPlugins []framework.EstimateReplicasPlugin
-	clientSet               clientset.Interface
-	informerFactory         informers.SharedInformerFactory
+	estimateReplicasPlugins   []framework.EstimateReplicasPlugin
+	estimateComponentsPlugins []framework.EstimateComponentsPlugin
+	clientSet                 clientset.Interface
+	informerFactory           informers.SharedInformerFactory
 }
 
 var _ framework.Framework = &frameworkImpl{}
@@ -85,12 +90,16 @@ func NewFramework(r Registry, opts ...Option) (framework.Framework, error) {
 	estimateReplicasPluginsList := reflect.ValueOf(&f.estimateReplicasPlugins).Elem()
 	estimateReplicasType := estimateReplicasPluginsList.Type().Elem()
 
+	estimateComponentsPluginsList := reflect.ValueOf(&f.estimateComponentsPlugins).Elem()
+	estimateComponentsType := estimateComponentsPluginsList.Type().Elem()
+
 	for name, factory := range r {
 		p, err := factory(f)
 		if err != nil {
 			return nil, fmt.Errorf("failed to initialize plugin %q: %w", name, err)
 		}
 		addPluginToList(p, estimateReplicasType, &estimateReplicasPluginsList)
+		addPluginToList(p, estimateComponentsType, &estimateComponentsPluginsList)
 	}
 	return f, nil
 }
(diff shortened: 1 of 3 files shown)


pkg/estimator/server/framework/runtime/framework.go
```diff
@@ -119,7 +128,10 @@ func (frw *frameworkImpl) SharedInformerFactory() informers.SharedInformerFactor
 func (frw *frameworkImpl) RunEstimateReplicasPlugins(ctx context.Context, snapshot *schedcache.Snapshot, replicaRequirements *pb.ReplicaRequirements) (int32, *framework.Result) {
 	startTime := time.Now()
 	defer func() {
+		// Emit metrics with both old and new labels for backward compatibility
+		// TODO: Remove estimator label in a future release (deprecated)
 		metrics.FrameworkExtensionPointDuration.WithLabelValues(estimator).Observe(utilmetrics.DurationInSeconds(startTime))
+		metrics.FrameworkExtensionPointDuration.WithLabelValues(estimateReplicasExtension).Observe(utilmetrics.DurationInSeconds(startTime))
```
seanlaii, line 134: Hi @RainbowMango , 

In order to differentiate the metrics between `EstimateReplica` and `EstimateComponents` plugins, I decided to introduce a new metrics with `EstimateComponents` label and plan to use `EstimateReplica` to replace the old `Estimator` label.

Additionally, to avoid breaking change I decided to emit the same metrics twice but with different label. 
What do you think?  I am OK with reusing `Estimator` but `EstimateReplica` might be clearer.

pkg/estimator/server/framework/runtime/framework_test.go
```diff
@@ -252,3 +262,202 @@ func Test_frameworkImpl_RunEstimateReplicasPlugins(t *testing.T) {
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			r := make(Registry)
+			for _, p := range tt.plugins {
+				if err := r.Register(p.name, func(framework.Handle) (framework.Plugin, error) {
+					return p, nil
+				}); err != nil {
+					t.Fatalf("fail to register EstimateComponentsPlugin plugin (%s)", p.Name())
+				}
+			}
```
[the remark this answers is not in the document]
seanlaii: @gemini-code-assist looks like this has been fixed in https://go.dev/blog/loopvar-preview?utm_source=chatgpt.com. Could you double check?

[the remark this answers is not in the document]
seanlaii: @gemini-code-assist we are using 1.24.9: https://github.com/karmada-io/karmada/blob/master/go.mod#L3.

seanlaii:
cc @RainbowMango @mszacillo

zhzhuang-zju:
Thanks 
/assign

pkg/estimator/server/framework/runtime/framework.go
zhzhuang-zju: I'm okay with introducing `EstimateReplica` to replace `Estimator`. However, if we proceed with this change, the release note should include an additional section on the deprecation of the existing metrics.

RainbowMango: I'm OK with this way by introducing a new metric to deprecate the old one. 

+1 on the release note

In addition, I realized both the old and new label names are written in camel case, which is not recommended. 
I failed to find some documentation about the naming rule, but there is a metrics linter in the Prometheus project:
https://github.com/RainbowMango/client_golang/blob/e729ba11961abb3a50910c324221d06a35bdce4f/prometheus/testutil/promlint/validations/generic_name_validations.go#L74-L89

PS: I introduced this tool to Kubernetes several years ago, but I didn't get a chance to do the same thing in Karmada.  I think Karmada should adopt this tool as well.

seanlaii: Thanks for the info! Updated the label values with snake case and the release note.

RainbowMango: > In addition, I realized both the old and new label names are written in camel case, which is not recommended.

Hey, I need to correct my previous comment about the label format, according to the validation tools I linked above, only the label `name` should be written in `snake_case`. Actually, what we are talking about here is the `value`, I don't see any rules for it.  

I see you updated the label value to `snake_case`, which also looks fine to me.

PR #6875, opened 2025-10-26
[Feature] Add `ResourceQuota` plugin for multi-component scheduling

seanlaii:
**What type of PR is this?**
/kind feature

Add `ResourceQuota` plugin for multi-component scheduling.

Part of #6734

```release-note
`karmada-scheduler-estimator`: Add `ResourceQuota` plugin for multi-component scheduling.
```

### DIFF
--- a/pkg/estimator/server/framework/plugins/resourcequota/resourcequota.go
+++ b/pkg/estimator/server/framework/plugins/resourcequota/resourcequota.go
@@ -24,6 +24,7 @@ import (
 	"strings"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
 	corelisters "k8s.io/client-go/listers/core/v1"
@@ -42,31 +43,36 @@ const (
 	Name                   = "ResourceQuotaEstimator"
 	resourceRequestsPrefix = "requests."
 	resourceLimitsPrefix   = "limits."
+
+	// noQuotaConstraint represents the value when there is no quota constraint.
+	noQuotaConstraint = math.MaxInt32
 )
 
-// resourceQuotaEstimator is to estimate how many replica allowed by the ResourceQuota constrain for a given pb.ReplicaRequirements
-// Kubernetes ResourceQuota object provides constraints that limit aggregate resource consumption per namespace
+// resourceQuotaEstimator estimates how many replicas are allowed by the ResourceQuota constraint for a given pb.ReplicaRequirements.
+// Kubernetes ResourceQuota object provides constraints that limit aggregate resource consumption per namespace.
 // It is categorized into 3 types:
 // 1) compute resource including cpu, memory, hugepages-<size> and extended resource like gpu
 // 2) storage including pvc, ephemeral-storage etc.
 // 3) object count including count/services, count/secrets etc.
-// ResourceQuota supports 6 scope
+// ResourceQuota supports 6 scopes:
 // 1) Terminating
 // 2) NotTerminating
 // 3) BestEffort
 // 4) NotBestEffort
 // 5) PriorityClass
 // 6) CrossNamespacePodAffinity
-// For now, we only support ComputeResource (cpu, memory and extended resource) and PriorityClass scope, the reasons are
-// 1) pb.ReplicaRequirements only contains requested resources currently, we cannot determine if its QoS (BestEffort or not)
-// 2) storage and object count do not support scope selector
-// ToDo (@wengyao04): we can extend the resourceQuotaEstimator support if further requirements are needed
+// For now, we only support ComputeResource (cpu, memory and extended resource) and PriorityClass scope. The reasons are:
+// 1) pb.ReplicaRequirements only contains requested resources currently, we cannot determine its QoS (BestEffort or not)
+// 2) Storage and object count quotas do not support scope selector
+// 3) Pod quota is not included in this iteration to keep the initial implementation focused on compute resources
+// TODO (@wengyao04): we can extend the resourceQuotaEstimator support if further requirements are needed
 type resourceQuotaEstimator struct {
 	enabled  bool
 	rqLister corelisters.ResourceQuotaLister
 }
 
 var _ framework.EstimateReplicasPlugin = &resourceQuotaEstimator{}
+var _ framework.EstimateComponentsPlugin = &resourceQuotaEstimator{}
 
 // New initializes a new plugin and returns it.
 func New(fh framework.Handle) (framework.Plugin, error) {
@@ -86,11 +92,11 @@ func (pl *resourceQuotaEstimator) Name() string {
 	return Name
 }
 
-// Estimate the replica allowed by the ResourceQuota
+// Estimate estimates the replicas allowed by the ResourceQuota constraints.
 func (pl *resourceQuotaEstimator) Estimate(_ context.Context,
 	_ *schedcache.Snapshot,
 	replicaRequirements *pb.ReplicaRequirements) (int32, *framework.Result) {
-	var replica int32 = math.MaxInt32
+	var replica int32 = noQuotaConstraint
 	if !pl.enabled {
 		klog.V(5).Info("Estimator Plugin", "name", Name, "enabled", pl.enabled)
 		return replica, framework.NewResult(framework.Noopperation, fmt.Sprintf("%s is disabled", pl.Name()))
@@ -100,8 +106,9 @@ func (pl *resourceQuotaEstimator) Estimate(_ context.Context,
 
 	rqList, err := pl.rqLister.ResourceQuotas(namespace).List(labels.Everything())
 	if err != nil {
-		klog.Error(err, "fail to list resource quota", "namespace", namespace)
-		return replica, framework.AsResult(err)
+		klog.Error(err, "failed to list resource quota", "namespace", namespace)
+		// Conservative approach: return 0 replicas on error
+		return 0, framework.AsResult(err)
 	}
 	for _, rq := range rqList {
 		rqEvaluator := newResourceQuotaEvaluator(rq, priorityClassName)
@@ -113,8 +120,9 @@ func (pl *resourceQuotaEstimator) Estimate(_ context.Context,
 
 	var result *framework.Result
 	switch replica {
-	case math.MaxInt32:
-		result = framework.NewResult(framework.Noopperation, fmt.Sprintf("%s has no operation on input replicaRequirements", pl.Name()))
+	case noQuotaConstraint:
+		// No quota constraints found - resources are schedulable without restrictions
+		result = framework.NewResult(framework.Success, fmt.Sprintf("%s found no quota constraints", pl.Name()))
 	case 0:
 		result = framework.NewResult(framework.Unschedulable, fmt.Sprintf("zero replica is estimated by %s", pl.Name()))
 	default:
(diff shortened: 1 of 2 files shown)


seanlaii:
Hi @RainbowMango @zhzhuang-zju @mszacillo, please help take a look when you get a chance! Thank you!

zhzhuang-zju:
Thanks 
/assign

pkg/estimator/server/framework/plugins/resourcequota/resourcequota.go
zhzhuang-zju:
```suggestion
	var replica int32 = noQuotaConstraint
```

pkg/estimator/server/framework/plugins/resourcequota/resourcequota.go
```diff
@@ -123,9 +129,97 @@ func (pl *resourceQuotaEstimator) Estimate(_ context.Context,
...
+// extractComponentMetadata extracts namespace and priorityClassName from the first component.
+// It assumes all components in a set share the same namespace and priorityClassName,
+// which should be validated by the admission webhook.
+func (pl *resourceQuotaEstimator) extractComponentMetadata(components []pb.Component) (namespace string, priorityClassName string, err error) {
+	if len(components) == 0 {
+		return "", "", fmt.Errorf("components list is empty")
+	}
+
+	// Extract metadata from the first component.
+	// All components are expected to share the same namespace and priorityClassName.
```
zhzhuang-zju: I agree that all components should share the same namespace, but I have doubts about them all sharing the same `priorityClassName`. Different pods within the same component set may have varying functions and responsibilities, which could lead to differing requirements for priority classes.

zhzhuang-zju: I suggest that we can categorize a component set into multiple component sets based on `priorityClassName`, then calculate the available sets for each category, and finally take the smallest value.

seanlaii: Yeah, make sense. Thanks for the suggestion! Updated.
Additionally, if we have more scope in the future, we can add something like this:
```
For each ResourceQuota RQ:
    1. Determine which components match RQ's scopes
    2. Aggregate only those matched components
    3. Calculate max sets allowed by RQ
    4. Track minimum across all RQs
```

pkg/estimator/server/framework/plugins/resourcequota/resourcequota.go
seanlaii: Thanks for the review! Updated.

pkg/estimator/server/framework/plugins/resourcequota/resourcequota.go
```diff
@@ -155,45 +249,176 @@ func newResourceQuotaEvaluator(rq *corev1.ResourceQuota, priorityClassName strin
...
+// are ignored as they are not supported by the current implementation.
+//
+// We accumulate int64 values (instead of using Quantity.Add directly) for two reasons:
+// 1. Performance: int64 arithmetic is faster than Quantity operations which may trigger big.Decimal conversions
+// 2. Format control: we need CPU in milli-units but Memory/GPU in whole units, easier to control during final conversion
+func (e *resourceQuotaEvaluator) aggregateComponentRequirements(components []pb.Component) corev1.ResourceList {
+	resourceRequirements := map[corev1.ResourceName]int64{}
+
+	for _, component := range components {
+		if component.ReplicaRequirements.ResourceRequest == nil {
```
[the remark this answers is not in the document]
seanlaii: @gemini-code-assist replicarequirement is required according to the protobuff api.

pkg/estimator/server/framework/plugins/resourcequota/resourcequota.go
```diff
@@ -155,45 +249,176 @@ func newResourceQuotaEvaluator(rq *corev1.ResourceQuota, priorityClassName strin
...
+		}
+	}
+
+	// Check if it's an extended resource (e.g., nvidia.com/gpu)
+	if corev1helper.IsExtendedResourceName(resourceName) {
+		return true
+	}
+
+	return false
+}
```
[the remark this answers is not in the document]
seanlaii: @gemini-code-assist I have updated. Please review it again.

pkg/estimator/server/framework/plugins/resourcequota/resourcequota.go
```diff
@@ -123,77 +129,333 @@ func (pl *resourceQuotaEstimator) Estimate(_ context.Context,
...
+		klog.V(5).Info("Estimator Plugin", "name", Name, "enabled", pl.enabled)
+		return maxSets, framework.NewResult(framework.Noopperation, fmt.Sprintf("%s is disabled", pl.Name()))
+	}
+
+	if len(components) == 0 {
+		klog.V(5).Infof("%s: received empty components list", pl.Name())
+		return maxSets, framework.NewResult(framework.Noopperation, fmt.Sprintf("%s received empty components list", pl.Name()))
+	}
+
+	namespace := components[0].ReplicaRequirements.Namespace
```
RainbowMango, line 152: Here is the thing that I feel is not right. Similar to https://github.com/karmada-io/karmada/pull/6876#discussion_r2467771354.

The `namespace` might not have to be stored inside of `ReplicaRequirements`. Furthermore, can we assume all components share the same namespace? If yes, can we put the namespace directly under `MaxAvailableComponentSetsRequest`?

pkg/estimator/server/framework/plugins/resourcequota/resourcequota.go
```diff
@@ -123,77 +129,333 @@ func (pl *resourceQuotaEstimator) Estimate(_ context.Context,
...
+
+	rqList, err := pl.rqLister.ResourceQuotas(namespace).List(labels.Everything())
+	if err != nil {
+		klog.Error(err, "failed to list resource quota", "namespace", namespace)
+		return maxSets, framework.AsResult(err)
+	}
+
+	if len(rqList) == 0 {
+		klog.V(5).Infof("%s: no ResourceQuota found in namespace %s", pl.Name(), namespace)
+	}
```
RainbowMango, line 165: Is it a short path? Since no ResourceQuota is configured, we can end the estimation faster.

RainbowMango (commented):
Review is ongoing.

pkg/estimator/server/framework/plugins/resourcequota/resourcequota.go
zhzhuang-zju: > Additionally, if we have more scope in the future, we can add something like this:

Yes, that's exactly what I wanted.

pkg/estimator/server/framework/plugins/resourcequota/resourcequota.go
seanlaii: Yes, you are right. Thanks for the info! Updated it with early return.
Additionally, I also found that originally if there is no resource quota, the return is `noop`, which is not clear. Therefore, I change it to `Success` since the quota plugin did actually execute. I think this is clearer and won't affect the existing functionalities, but want to hear your opinion. Thanks!

pkg/estimator/server/framework/plugins/resourcequota/resourcequota.go
zhzhuang-zju: I observed the CRDs for `SparkApplication` and `FlinkDeployment`, and their different components do not explicitly support configuring the namespace. They only theoretically have the capability to configure the namespace because they include `corev1.PodTemplateSpec`.
- flink jobManager: https://github.com/apache/flink-kubernetes-operator/blob/3e3cb584a133c7ce0b310e4a1e5d986288fc0303/flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/spec/JobManagerSpec.java#L38-L47
- flink taskManager: https://github.com/apache/flink-kubernetes-operator/blob/3e3cb584a133c7ce0b310e4a1e5d986288fc0303/flink-kubernetes-operator-api/src/main/java/org/apache/flink/kubernetes/operator/api/spec/TaskManagerSpec.java#L42-L53
- spark driver and executor: https://github.com/kubeflow/spark-operator/blob/d0c8e69063e3eabf51cc5dc3052879cb5ef5a58b/api/v1beta2/sparkapplication_types.go#L419-L433

From the implementation of the Spark operator, it only focuses on the namespace of the `SparkApplication`. Additionally, I tried deploying a CR locally with namespaces set for both the driver and executor, and it ultimately failed to deploy. Therefore, `SparkApplication` does not support multi-components cross-namespace configurations.

For `FlinkDeployment`, I believe it is the same. @mszacillo @seanlaii, you are more knowledgeable about `FlinkDeployment`. Could you confirm this?

seanlaii: Thanks for the detailed experiment and explanation!
I am not familiar with `FlinkDeployment`, but for [`RayJob`, `RayCluster`, and `RayService`](https://github.com/ray-project/kuberay/tree/master/ray-operator/apis/ray/v1), they only support deploying to a single namespace.

pkg/estimator/server/framework/plugins/resourcequota/resourcequota.go
zhzhuang-zju: >  Therefore, I change it to Success since the quota plugin did actually execute.

Makes sense to me.

pkg/estimator/server/framework/plugins/resourcequota/resourcequota.go
```diff
@@ -123,77 +130,336 @@ func (pl *resourceQuotaEstimator) Estimate(_ context.Context,
...
+
+	namespace := components[0].ReplicaRequirements.Namespace
+	// Group components by priorityClassName since different pods may have different priorities.
+	componentsByPriority := pl.groupComponentsByPriority(components)
+
+	rqList, err := pl.rqLister.ResourceQuotas(namespace).List(labels.Everything())
+	if err != nil {
+		klog.Error(err, "failed to list resource quota", "namespace", namespace)
+		return maxSets, framework.AsResult(err)
+	}
```
RainbowMango, line 159: Should we return `maxSets` in case an internal error is found? That means the estimation will bypass the quota plugin.

pkg/estimator/server/framework/plugins/resourcequota/resourcequota.go
```diff
@@ -123,77 +130,336 @@ func (pl *resourceQuotaEstimator) Estimate(_ context.Context,
...
+
+		if groupMaxSets < maxSets {
+			maxSets = groupMaxSets
+		}
+
+		// Early exit if any priority group allows zero sets.
+		if maxSets == 0 {
+			break
+		}
+	}
```
RainbowMango: The current logic in the loop might have a critical flaw in how it evaluates quota consumption across priority classes.

By evaluating each `groupComponents` separately against the same, unmodified `rqList` (the same total available quota), this can lead to over-admission.

Consider a workload with two components (A and B) under different PriorityClasses, and the total quota allows only one such component. Evaluating A against the full quota might allow 1 set. Evaluating B against the same full quota might also allow 1 set. The loop takes the minimum (maxSets = 1) and admits the workload. However, running both A and B would require resources for two components, exceeding the quota.

RainbowMango: In addition, the current algorithm calculates the available replicas using availableResources / requestedResource, which is convenient. However, considering that in the future we will need to pass in a batch of components requiring reservation, we will first need to subtract these reserved resources from the availableResources before performing the evaluation. Therefore, we still need to develop an algorithm that decrements the Quota accordingly. See the draft idea at https://github.com/karmada-io/karmada/pull/6812#discussion_r2416607785.

In summary, although we don't need it right now, should we prepare for future extensibility at this time?

pkg/estimator/server/framework/plugins/resourcequota/resourcequota.go
zhzhuang-zju: > That means the estimation will bypass the quota plugin.

Not really. `framework.AsResult(err)` returns the error to the caller, and the caller should only trust the plugin's calculation result when no error has occurred. However, when an error does occur, it might be better to return 0—i.e., adopt a conservative stance on the available replicas in the face of an internal error.

pkg/estimator/server/framework/plugins/resourcequota/resourcequota.go
zhzhuang-zju: > By evaluating each groupComponents separately against the same, unmodified rqList (the same total available quota), this can lead to over-admission.

+1. The workflow described in https://github.com/karmada-io/karmada/pull/6875#discussion_r2464452640 is correct in principle. However, during implementation, the first step—`Determine which components match RQ's scopes`—was moved before the `For each ResourceQuota RQ` loop, so that components are grouped by `priorityClassName` upfront.

Take a ResourceQuota without any scope defined as an example: it should originally compute the number of available replica sets based on the entire `componentSet`. But due to this change, the quota now calculates the available sets separately for each `groupedComponents` and then takes the minimum across groups—effectively over-constraining the result.

> we will first need to subtract these reserved resources from the availableResources before performing the evaluation.

FYI, the `SubResource` function is specifically used to subtract these reserved resources from the `availableResource`. We can use it if needed.
https://github.com/karmada-io/karmada/blob/9902350ea9265ddc928d8d0f15822ede104ee471/pkg/util/resource.go#L75-L95

pkg/estimator/server/framework/plugins/resourcequota/resourcequota.go
seanlaii: Thanks for the suggestion!
Yes, the caller should only use the result if there is no error such as https://github.com/karmada-io/karmada/blob/master/pkg/estimator/client/accurate.go#L122-L125.
I agree with the conservative approach. Updated.

pkg/estimator/server/framework/plugins/resourcequota/resourcequota.go
seanlaii: Yeah, you are totally correct! Sorry, I implemented in a wrong way. I have updated the implementation so that it will now first iterate the `resource quota` and then do the filtering based on the priority class name. I have also added a test case to validate this case. 

Regarding the `reservation`, maybe we can have something like this:
```
func evaluateResourcesAgainstQuota(
    availableResources corev1.ResourceList,
    perSetRequirements corev1.ResourceList,
    reservedResources corev1.ResourceList) int32 {
    
    effectiveAvailable := availableResources.DeepCopy()
    if reservedResources != nil {
        effectiveAvailable.Sub(reserved)
    }
    
    filtered := filterConstrainedResources(effectiveAvailable, perSetRequirements)
    resource := util.NewResource(effectiveAvailable)
    allowed := resource.MaxDivided(filtered) 
    
    return int32(allowed)
}
```
Please feel free to correct me if my understanding is incorrect.
Thanks!

pkg/estimator/server/framework/plugins/resourcequota/resourcequota.go
RainbowMango: Thanks.

pkg/estimator/server/framework/plugins/resourcequota/resourcequota.go
RainbowMango: Yeah, it sounds even better than what I proposed before. 
What I was thinking is mostly for the estimation based on node resources, which can not go with the `divide` way.

pkg/estimator/server/framework/plugins/resourcequota/resourcequota.go
```diff
@@ -123,77 +131,327 @@ func (pl *resourceQuotaEstimator) Estimate(_ context.Context,
...
+	perSetRequirements corev1.ResourceList) int32 {
+	// Filter to only include resources constrained by this quota
+	filtered := filterConstrainedResources(availableResources, perSetRequirements)
+
+	if len(filtered) == 0 {
+		return noQuotaConstraint
+	}
+
+	// Create a Resource object with available quota
+	resource := util.NewResource(availableResources)
```
RainbowMango: Warning from my Goland IDE:
> Variable 'resource' collides with imported package name
> While legal, such declarations will make using the package exported identifiers impossible after the declaration or create confusion when reading the code.

pkg/estimator/server/framework/plugins/resourcequota/resourcequota.go
```diff
@@ -123,77 +131,327 @@ func (pl *resourceQuotaEstimator) Estimate(_ context.Context,
...
+		quotaApplies = true
+	} else {
+		quotaApplies = quotaAppliesToPriority(selectors, priorityClassName)
+	}
+
+	// If the quota applies, extract the free resources
+	if quotaApplies {
 		matchResource := matchingResources(resourceNames(rq.Status.Hard))
 		if len(matchResource) != 0 {
 			resource := calculateFreeResources(rq, matchResource)
```
RainbowMango: ditto. 
> Variable 'resource' collides with imported package name

pkg/estimator/server/framework/plugins/resourcequota/resourcequota.go
```diff
@@ -123,77 +131,327 @@ func (pl *resourceQuotaEstimator) Estimate(_ context.Context,
...
+		filteredRequirements := filterConstrainedResources(availableResources, replicaRequirements.ResourceRequest)
+
+		// If no resources are constrained by this quota, skip it
+		if len(filteredRequirements) == 0 {
+			continue
 		}
-		resource := util.NewResource(resourceList)
+
+		// Create a Resource object with available quota
+		resource := util.NewResource(availableResources)
```
RainbowMango: ditto.

pkg/estimator/server/framework/plugins/resourcequota/resourcequota.go
RainbowMango: ditto.

RainbowMango (commented):
/lgtm
Looks pretty good.

Just left some warnings raised by my IDE, some of them are for the legacy code. 
It would be great to get them fixed in this PR, but it shouldn't be a blocker. 

I will wait for a few hours, and then move it forward.

seanlaii:
> Just left some warnings raised by my IDE, some of them are for the legacy code.
It would be great to get them fixed in this PR, but it shouldn't be a blocker.

Sure, rename them. Thanks!

PR #6877, opened 2025-10-27
introduce node resource estimation as the default estimator plugin

zhzhuang-zju:
**What type of PR is this?**
/kind feature

The current estimator framework provides a plugin mechanism to calculate the available replicas/sets of a workload. However, the calculation based on node resources is not implemented as a plugin. Instead, it is performed separately after the plugin returns its result, and then compared with the plugin’s output.
I propose making node resource–based calculation a default plugin for the following reasons:
- It unifies the logic: the result returned by the plugins would be the final result.
- This capability was already considered when the estimator framework was initially designed, but it was never implemented. I believe it’s essential to deliver this enhancement now while introducing the new `EstimateComponentsPlugin`, to avoid increasing the cost and complexity of future modifications.
https://github.com/karmada-io/karmada/blob/a892fef437356429d7aa257efb22b8e3840b54eb/pkg/estimator/server/framework/interface.go#L184-L188
- It enables consistent monitoring through the plugin metrics

Part of #6734

```release-note
`karmada-scheduler-estimator`: Refactored the replica estimation logic by moving the node resource-based calculation into a dedicated, default plugin. 
```

### DIFF
--- a/pkg/estimator/server/estimate.go
+++ b/pkg/estimator/server/estimate.go
@@ -19,17 +19,12 @@ package server
 import (
 	"context"
 	"fmt"
-	"sync/atomic"
 	"time"
 
-	corev1 "k8s.io/api/core/v1"
 	utiltrace "k8s.io/utils/trace"
 
 	"github.com/karmada-io/karmada/pkg/estimator/pb"
-	nodeutil "github.com/karmada-io/karmada/pkg/estimator/server/nodes"
-	"github.com/karmada-io/karmada/pkg/util"
 	schedcache "github.com/karmada-io/karmada/pkg/util/lifted/scheduler/cache"
-	"github.com/karmada-io/karmada/pkg/util/lifted/scheduler/framework"
 )
 
 // EstimateReplicas returns max available replicas in terms of request and cluster status.
@@ -56,25 +51,7 @@ func (es *AccurateSchedulerEstimatorServer) EstimateReplicas(ctx context.Context
 	return maxAvailableReplicas, nil
 }
 
-func (es *AccurateSchedulerEstimatorServer) estimateReplicas(
-	ctx context.Context,
-	snapshot *schedcache.Snapshot,
-	requirements pb.ReplicaRequirements,
-) (int32, error) {
-	allNodes, err := snapshot.NodeInfos().List()
-	if err != nil {
-		return 0, err
-	}
-	var (
-		affinity    = nodeutil.GetRequiredNodeAffinity(requirements)
-		tolerations []corev1.Toleration
-	)
-
-	if requirements.NodeClaim != nil {
-		tolerations = requirements.NodeClaim.Tolerations
-	}
-
-	var res int32
+func (es *AccurateSchedulerEstimatorServer) estimateReplicas(ctx context.Context, snapshot *schedcache.Snapshot, requirements pb.ReplicaRequirements) (int32, error) {
 	replicas, ret := es.estimateFramework.RunEstimateReplicasPlugins(ctx, snapshot, &requirements)
 
 	// No replicas can be scheduled on the cluster, skip further checks and return 0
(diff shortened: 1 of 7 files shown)


pkg/estimator/server/framework/interface.go
```diff
@@ -182,7 +183,6 @@ func (s *Result) IsUnschedulable() bool {
 }
 
 // IsNoOperation returns true if "Result" is not nil and Code is "Nooperation".
-// TODO (wengyao04): we can remove it once we include node resource estimation as the default plugin in the future.
```
zhzhuang-zju, line 185: Although I am introducing the node resource estimation as the default plugin, I do not plan to delete the Nooperation plugin. This is because it decouples subsequent processing from whether the plugin is enabled or not.

RainbowMango:
@zhzhuang-zju Please resolve the conflicts.

zhzhuang-zju:
> @zhzhuang-zju Please resolve the conflicts.

done~

pkg/estimator/server/framework/interface.go
```diff
@@ -85,6 +85,7 @@ type EstimateComponentsPlugin interface {
 type Handle interface {
 	ClientSet() clientset.Interface
 	SharedInformerFactory() informers.SharedInformerFactory
+	Parallelism() int
```
RainbowMango, line 88: Can we add a comment on it?

zhzhuang-zju:
>  timed out waiting for command /home/runner/go/bin/karmadactl register 172.18.0.3:5443 --token 9slvgw.hbisjrgv7vq1z2wt --discovery-token-unsafe-skip-ca-verification=true --kubeconfig=/home/runner/.kube/member-e2e-b2tcg.config --cluster-name member-e2e-b2tcg --karmada-agent-image docker.io/karmada/karmada-agent:latest:

/retest

zhzhuang-zju:
> What do you think of the TODO item:

Ref to https://github.com/karmada-io/karmada/pull/6877#discussion_r2464382693

pkg/estimator/server/framework/interface.go
RainbowMango: The comment is saying once the node resource estimation becomes the default plugin, it is always enabled, then this method will always return true, thus useless.

zhzhuang-zju: Yes, but by design, every plugin should be pluggable.

pkg/estimator/server/framework/plugins/noderesource/noderesource.go
```diff
@@ -0,0 +1,109 @@
...
+
+// Name returns name of the plugin. It is used in logs, etc.
+func (pl *nodeResourceEstimator) Name() string {
+	return Name
+}
+
+// Estimate the replica allowed by the node resources for a given pb.ReplicaRequirements.
+func (pl *nodeResourceEstimator) Estimate(ctx context.Context,
+	snapshot *schedcache.Snapshot,
+	requirements *pb.ReplicaRequirements) (int32, *framework.Result) {
```
RainbowMango:
```suggestion
func (pl *nodeResourceEstimator) Estimate(ctx context.Context, snapshot *schedcache.Snapshot, requirements *pb.ReplicaRequirements) (int32, *framework.Result) {
```

pkg/estimator/server/framework/interface.go
RainbowMango: Fine, should we remove the TODO this time?

pkg/estimator/server/estimate.go
```diff
@@ -61,20 +56,6 @@ func (es *AccurateSchedulerEstimatorServer) estimateReplicas(
 	snapshot *schedcache.Snapshot,
 	requirements pb.ReplicaRequirements,
 ) (int32, error) {
```
RainbowMango:
```suggestion
func (es *AccurateSchedulerEstimatorServer) estimateReplicas(ctx context.Context, snapshot *schedcache.Snapshot, requirements pb.ReplicaRequirements) (int32, error) {
```

Issue #6841, opened 2025-10-16
Flaky E2E tests – tracking intermittent failures


zhzhuang-zju:
Our E2E tests have been failing randomly lately. These flaky failures make it difficult to determine if a genuine issue was introduced or if it’s just a test problem.

This issue is to collect and track intermittent E2E failures in one place. If you see a test fail and it passes on retry, please add a comment with:

- The test name
- A short error snippet or symptom
- Link to the CI run

Identified E2E test failures:

- [metrics presence testing](https://github.com/karmada-io/karmada/blob/3bdee54a58dd3bb026cfc04e58999a54c65325b9/test/e2e/suites/base/metrics_test.go#L67)
  failure e2e link: https://github.com/karmada-io/karmada/actions/runs/18395251404/job/52413766044?pr=4574#step:6:742
  - [ ] https://github.com/karmada-io/karmada/pull/4574, @zhzhuang-zju
Note: This is a newly introduced flaky failure from #4574, and it will be fixed together with #4574.

- [Application failover testing with purgeMode gracefully](https://github.com/karmada-io/karmada/blob/3bdee54a58dd3bb026cfc04e58999a54c65325b9/test/e2e/suites/base/failover_test.go#L357)
  failure e2e link: https://github.com/karmada-io/karmada/actions/runs/18553955121/job/52887389171#step:6:444
  - [x] https://github.com/karmada-io/karmada/pull/6845, @zhzhuang-zju   

- [Karmada Operator E2E test: PriorityClass configuration testing](https://github.com/karmada-io/karmada/issues/6842)
   - [x] https://github.com/karmada-io/karmada/pull/6843, @zhzhuang-zju        
- [Karmada instance status testing](https://github.com/karmada-io/karmada/actions/runs/18830730849/job/53721954046?pr=6878#step:6:53) 
   - [x] https://github.com/karmada-io/karmada/pull/6908 (@zhzhuang-zju )

- [Deploy workloads should be rejected in case of no enough quota left](https://github.com/karmada-io/karmada/pull/6876#issuecomment-3455446833)
   - [x] https://github.com/karmada-io/karmada/pull/6907 (@zhzhuang-zju )
   
- [Multi-Components: FederatedResourceQuota enforcement testing](https://github.com/karmada-io/karmada/actions/runs/19536609426/job/55932805279)
   -  [x]  https://github.com/karmada-io/karmada/pull/6941 (@zhzhuang-zju )

[...]

Issue #6819, opened 2025-10-12
Add Admission Control for Karmada Operator


jabellard:
**What would you like to be added**:

Add [**admission control**](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/) for the Karmada Operator :
- **Mutating webhook** to apply sensible defaults.
- **Validating webhook** to reject invalid specs.

To maintain backwards compatibility, registration and wiring of admission webhooks will be controlled via a flag that is off by default. As such, unless explicitly enabled when deploying the operator, the current behavior with both defaulting and validations running during reconciliation in the operator will be maintained.

**Why is this needed**:

- **Best practice**: [Admission webhooks](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/) are the Kubernetes-native way to enforce policy and defaults at the API boundary; they provide consistent behavior across all clients and earlier, clearer feedback.
- **Problems with in-controller defaulting/validation**:
  - **Unnecessary  reconciles & retries**: Invalid specs pass admission, then fail in reconcile, causing requeues, wasted compute, and noisy logs.
  - **Slower feedback loops**: Errors surface only after controller processing, increasing feedback latency for users and automation.
  - **Race/ordering issues**: Defaulting/validation inside the controller happens after the object is stored. If a user (or GitOps) sends rapid, conflicting updates, the controller can see transient, invalid specs and thrash before it settles. Admission stops bad states before they hit storage.

jabellard:
Hey @RainbowMango . Can you please take a look? I have a [WIP PR](https://github.com/karmada-io/karmada/pull/6820) for this that I'm working on. The key thing for this feature is that it will be opt-in to maintain backwards compatibility for users who want to continue to rely on the in-operator defaulting/validating behavior. However, given that using [admission control](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/) for these types of problems is a best practice, I think that should be the recommended approach moving forward. What do you think?

RainbowMango:
We don't need a separate component in addition to the `karmada-operator`, right? 
The admission webhook is good, and I remember we don't want to introduce another webhook component, so we implements the validation and mutating(default) inside the reconcile logic.

jabellard:
> We don't need a separate component in addition to the `karmada-operator`, right? The admission webhook is good, and I remember we don't want to introduce another webhook component, so we implements the validation and mutating(default) inside the reconcile logic.

[...]

PR #6908, opened 2025-11-05
operator: avoid unnecessary reconcile

zhzhuang-zju:
**What type of PR is this?**
/kind cleanup

Currently, default values for Karmada CRs are set during the operator’s reconcile loop. This introduces two issues:

1. **Unnecessary reconciliation**: When a Karmada CR is updated but becomes identical to its previous state after applying defaults, the controller still performs a full reconcile twice—once for the original change and again due to the update caused by setting defaults.

2. **Unstable CR status**: The operator sets the `Ready` condition of the Karmada CR to `False` at the start of each reconcile and updates it to `True` upon successful completion. For example, during initial deployment, the first reconcile succeeds and sets `Ready=True`. However, because the reconcile process itself updates the CR (by filling in default values), it triggers another reconcile, which resets `Ready=False`, causing flickering in the status. This also causes occasional flakiness in the operator's e2e tests.

To address this, I propose filtering out such "no-op" updates in the `UpdateEvent` handler—skipping reconciliation if the change is solely due to defaulting and results in no semantic difference.

Note that #6819 would also resolve this problem by moving defaulting logic into an admission controller. Therefore, I’ve left a TODO: once admission-based defaulting is implemented, this workaround can be removed.

Part of #6841

```release-note

```

### DIFF
--- a/operator/pkg/controller/karmada/controller.go
+++ b/operator/pkg/controller/karmada/controller.go
@@ -92,10 +92,16 @@ func (ctrl *Controller) Reconcile(ctx context.Context, req controllerruntime.Req
 		return ctrl.removeFinalizer(ctx, karmada)
 	}
 
-	if err := ctrl.updateKarmada(ctx, karmada); err != nil {
+	// Simulate a mutation webhook by setting defaults. Return and wait for the next reconciliation if spec is updated.
+	specUpdated, err := ctrl.updateKarmada(ctx, karmada)
+	if err != nil {
 		return controllerruntime.Result{}, err
 	}
+	if specUpdated {
+		return controllerruntime.Result{}, nil
+	}
 
+	// Simulate a validation webhook by validating the resource.
 	if err := ctrl.validateKarmada(ctx, karmada); err != nil {
 		klog.ErrorS(err, "Validation failed for karmada", "name", karmada.Name)
 		return controllerruntime.Result{}, nil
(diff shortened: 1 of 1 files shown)


zhzhuang-zju:
cc @jabellard @RainbowMango

operator/pkg/controller/karmada/controller.go
```diff
@@ -183,5 +183,17 @@ func (ctrl *Controller) onKarmadaUpdate(updateEvent event.UpdateEvent) bool {
...
+	}
+
+	// The spec has changed. We need to check if it's a semantic change or just defaulting.
+	// TODO(@zhzhuang-zju): If default values are set via an admission controller, the default() call below can be omitted.
+	newCopy := newObj.DeepCopy()
+	oldCopy := oldObj.DeepCopy()
+
+	operatorscheme.Scheme.Default(newCopy)
+	operatorscheme.Scheme.Default(oldCopy)
+	return !reflect.DeepEqual(newCopy.Spec, oldCopy.Spec)
```
RainbowMango: This is concerning as the mutation logic might bring a bug; it also should be validated.

zhzhuang-zju: If there is a bug, the `oldObj` will be validated during reconciliation. When `reflect.DeepEqual(newCopy.Spec, oldCopy.Spec)` evaluates to `true`, no new bug will be introduced.

PR #6912, opened 2025-11-05
Implementing maxAvailableComponentSets for resource models

mszacillo:
**What type of PR is this?**
/kind feature

This PR implements the maxAvailableComponentsSets within general estimator based on resource models. The packing algorithm is first-fit decreasing, which is a greedy approach to bin packing. To make the calculation of the maxSets more efficient we use binary-search for checking the max set number, and then attempting to fit all the number of replicas onto the existing nodes. 

Part of #6734

```release-note
`karmada-scheduler`: Implements `getMaximumSetsBasedOnResourceModels` in general estimator.
```

### DIFF
--- a/pkg/estimator/client/general.go
+++ b/pkg/estimator/client/general.go
@@ -19,7 +19,9 @@ package client
 import (
 	"context"
 	"fmt"
+	"maps"
 	"math"
+	"sort"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
@@ -54,7 +56,7 @@ func (ge *GeneralEstimator) MaxAvailableReplicas(_ context.Context, clusters []*
 }
 
 func (ge *GeneralEstimator) maxAvailableReplicas(cluster *clusterv1alpha1.Cluster, replicaRequirements *workv1alpha2.ReplicaRequirements) int32 {
-	//Note: resourceSummary must be deep-copied before using in the function to avoid modifying the original data structure.
+	// Note: resourceSummary must be deep-copied before using in the function to avoid modifying the original data structure.
 	resourceSummary := cluster.Status.ResourceSummary.DeepCopy()
 	if resourceSummary == nil {
 		return 0
@@ -149,8 +151,7 @@ func (ge *GeneralEstimator) maxAvailableComponentSets(cluster *clusterv1alpha1.C
 	}
 
 	if features.FeatureGate.Enabled(features.CustomizedClusterResourceModeling) && len(cluster.Status.ResourceSummary.AllocatableModelings) > 0 {
-		num, err := getMaximumSetsBasedOnResourceModels(cluster, components)
-		if err != nil {
+		if num, err := getMaximumSetsBasedOnResourceModels(cluster, components, podBound); err != nil {
 			klog.Warningf("Failed to get maximum sets based on resource models, skipping: %v", err)
 		} else if num < maxSets {
 			maxSets = num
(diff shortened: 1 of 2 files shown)


pkg/estimator/client/general.go
```diff
@@ -160,13 +160,197 @@ func (ge *GeneralEstimator) maxAvailableComponentSets(cluster *clusterv1alpha1.C
...
+			continue
+		}
+		if cap[k] < req {
+			return false
+		}
+	}
+	return true
+}
+
+func consumeAll(demand map[corev1.ResourceName]int64, cap map[corev1.ResourceName]int64) {
```
zhzhuang-zju: cc @mszacillo

zhzhuang-zju:
Thanks 
/assign

pkg/estimator/client/general.go
mszacillo: fixed!

pkg/estimator/client/general.go
```diff
@@ -160,13 +160,197 @@ func (ge *GeneralEstimator) maxAvailableComponentSets(cluster *clusterv1alpha1.C
...
-// It should refine the maximum sets based on cluster resource models, similar
-// to getMaximumReplicasBasedOnResourceModels but adapted to full component sets.
-func getMaximumSetsBasedOnResourceModels(_ *clusterv1alpha1.Cluster, _ []workv1alpha2.Component) (int64, error) {
-	// TODO: implement logic based on cluster.Spec.ResourceModels
-	// For now, just return MaxInt64 so it never reduces the upper bound.
-	return math.MaxInt64, nil
+// getMaximumSetsBasedOnResourceModels computes the maximum number of full sets that can be
+// placed on a cluster using the cluster's ResourceModels. It expands one set into individual component replicas
+// and performs a first-fit-decreasing placement onto model-grade nodes.
+// `upperBound` caps the search. We can this using the podBound (allowedPods / podsPerSet)
```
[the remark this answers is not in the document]
zhzhuang-zju: +1

pkg/estimator/client/general.go
```diff
@@ -160,13 +160,197 @@ func (ge *GeneralEstimator) maxAvailableComponentSets(cluster *clusterv1alpha1.C
...
+	minMap := convertToResourceModelsMinMap(cluster.Spec.ResourceModels)
+
+	// Build nodes for each grade index i
+	var nodes []modelNode
+	for i := 0; i < len(cluster.Spec.ResourceModels); i++ {
+		if i >= len(cluster.Status.ResourceSummary.AllocatableModelings) {
+			// Shouldn’t happen — status is malformed
+			return nil, fmt.Errorf("resource model/status mismatch: %d grades in spec, %d in status",
+				len(cluster.Spec.ResourceModels), len(cluster.Status.ResourceSummary.AllocatableModelings))
+		}
```
zhzhuang-zju:
```suggestion
	if len(cluster.Spec.ResourceModels) != len(cluster.Status.ResourceSummary.AllocatableModelings) {
		// Shouldn’t happen — status is malformed
		return nil, fmt.Errorf("resource model/status mismatch: %d grades in spec, %d in status",
			len(cluster.Spec.ResourceModels), len(cluster.Status.ResourceSummary.AllocatableModelings))
	}
	for i := 0; i < len(cluster.Spec.ResourceModels); i++ {
```

Move the length check out of the for loop to avoid redundant evaluations.

pkg/estimator/client/general.go
```diff
@@ -160,13 +160,197 @@ func (ge *GeneralEstimator) maxAvailableComponentSets(cluster *clusterv1alpha1.C
...
+			capTemplate[resName] = quantityAsInt64(mins[i])
+		}
+
+		// Append `count` identical nodes of this grade
+		for n := 0; n < count; n++ {
+			// Copy capTemplate to each node
+			capCopy := make(map[corev1.ResourceName]int64, len(capTemplate))
+			for k, v := range capTemplate {
+				capCopy[k] = v
+			}
```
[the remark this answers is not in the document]
zhzhuang-zju: > This map copying logic is repeated in a few places in this file 

Yes, but compared to extracting it into a helper method, I prefer using `util.Resource`. First, it's a struct and doesn't require repeated copying; second, it encapsulates many resource-related operations, which can simplify the code.
https://github.com/karmada-io/karmada/blob/29536305bcc4c965a3538241bc60fb468d24d377/pkg/util/resource.go#L29-L37

pkg/estimator/client/general.go
```diff
@@ -160,13 +160,197 @@ func (ge *GeneralEstimator) maxAvailableComponentSets(cluster *clusterv1alpha1.C
...
+	nodes, err := buildModelNodes(cluster)
+	if err != nil {
+		return -1, err
+	}
+	if len(nodes) == 0 {
+		return 0, nil
+	}
+
+	// Expand one set into per-replica resource maps
+	oneSet := expandReplicasOneSet(components)
```
zhzhuang-zju: I have concerns about expanding a `ComponentSet` down to the replica level. Since all replicas of the same component have identical resource requests, doing so would result in `oneSet` containing a large amount of redundant data when the replica count is high. Moreover, subsequent sorting operations would incur significant unnecessary overhead due to these duplicates.  

Could we instead optimize this by using a struct that carries both the component’s resource request and its replica count? Like:
https://github.com/karmada-io/karmada/blob/29536305bcc4c965a3538241bc60fb468d24d377/pkg/estimator/server/framework/plugins/noderesource/noderesource.go#L201-L208

pkg/estimator/client/general.go
```diff
@@ -160,13 +160,197 @@ func (ge *GeneralEstimator) maxAvailableComponentSets(cluster *clusterv1alpha1.C
...
+// demandScore is used to rank replicas in terms of their size and difficulty to schedule
+// Currently uses sum of values
+func demandScore(m map[corev1.ResourceName]int64) int64 {
+	var s int64
+	for _, v := range m {
+		if v > 0 {
+			s += v
+		}
+	}
+	return s
```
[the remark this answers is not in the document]
zhzhuang-zju: @gemini-code-assist Can we compute the score using  
`max(cpu_request / cluster_total_cpu, mem_request / cluster_total_mem, gpu_request / cluster_total_gpu, ...)`?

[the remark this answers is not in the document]
zhzhuang-zju: @mszacillo The first-fit decreasing algorithm is more effective than the first-fit algorithm. However, I haven't found a commonly used method for determining the decreasing order in multiple dimensions. I asked an AI, and the recommended approach is to use `score := max(cpu_request / cluster_total_cpu, mem_request / cluster_total_mem, gpu_request / cluster_total_gpu, ...)`. If you don't have a better solution, we can adopt this sorting method.

pkg/estimator/client/general.go
mszacillo: That's a fair suggestion! Lemme see what I can do.

pkg/estimator/client/general.go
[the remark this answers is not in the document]
mszacillo: Sure thing, I agree the current implementation definitely does have issues when resource values are of different scales, which can skew the result. Is there a reason why you prefer max over using a weighted sum?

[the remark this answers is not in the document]
zhzhuang-zju: Firstly, using a weighted sum requires setting weights, which can be somewhat subjective. Secondly, computing the max would be simpler. Of course, if a weighted sum proves to be more effective, then that would be a different matter.

zhzhuang-zju:
Thanks @mszacillo — this is a tough task!  
From a functional perspective, it is ready. But I have some concerns about the implementation:

1. Does the "binary search" here actually improve efficiency? 
Since each iteration fully evaluates whether all `upperBound` component sets can be accommodated by the nodes, the total computation isn’t reduced—in fact, it might even increase due to repeated full validations.

2. Given the high similarity between this logic and the `noderesource` plugin, can we unify their calculation logic and workflow?
The only difference appears to be the data representation: the `noderesource` plugin uses `schedulerframework.NodeInfo` to represent node resources, whereas this implementation uses resource models. If we convert these resource models into `schedulerframework.NodeInfo`, could we fully reuse the `noderesource` plugin’s code?

Admittedly, the `noderesource` plugin might not yet be complete—for example, it may lack component sorting. But unifying the logic would make future enhancements easier to apply consistently across both components. WDYT?
Also, let’s ask @RainbowMango and @seanlaii for their opinion.

mszacillo:
Thanks for the review @zhzhuang-zju !

> Does the "binary search" here actually improve efficiency?

Yes, binary search helps here. I implemented this method with the understanding that we are looking for an exact upperBound of possible component sets - rather than an approximation which would have been a lot simpler. 

We're looking for a max number of component sets between 0 and upperBound, with the knowledge that:

- If `S` sets are feasible, then all sets `s` <= `S` are also feasible
- If `S` sets are _not_ feasible, then all sets `s` >= `S` also also not feasible

We could naively loop through all possible set numbers [0, upperBound] and do the feasibility check, but given the rules listed above, binary search fits in quite well. And O(logN) < O(N).

> Since each iteration fully evaluates whether all upperBound component sets can be accommodated by the nodes, the total computation isn’t reduced—in fact, it might even increase due to repeated full validations.

Not sure I fully follow. By full validations you are referring to the feasibility fit check right? Even though each iteration fully validates feasibility, binary search reduces how many times the check is done. 

If we're discussing the feasibility check time complexity (assuming k = number of components), then we are sorting each time O(klogk), and then doing a greedy packing, which results in O(k * n * d), where n is nodes and d is the resources. Which given that number of components and resources is generally small, the complexity of this is generally going to depend on the number of nodes (larger clusters taking longer to check).  

That said! I'm realizing that we can move the sorting to be outside of the feasibility check, since the score we assign is dependent on the cluster. So it only needs to be computed once, and we can instead just set the replica count to match the # of sets we are running the feasibility check for. That would improve efficiency. 

> Given the high similarity between this logic and the noderesource plugin, can we unify their calculation logic and workflow?

I'm happy to unify calculation logic to make the code cleaner, but to me this sounds like a follow-up task. I could possibly even pick this up.

zhzhuang-zju:
> Not sure I fully follow. By full validations you are referring to the feasibility fit check right? Even though each iteration fully validates feasibility, binary search reduces how many times the check is done.

Binary search reduces the number of feasibility fit checks. However, is it possible to obtain the final result in a single feasibility fit check—for example, by assigning component sets one by one until we can no longer assign a complete component set?

> I'm happy to unify calculation logic to make the code cleaner, but to me this sounds like a follow-up task. I could possibly even pick this up.

Yes, you're absolutely right—this can be addressed as a follow-up task.

mszacillo:
> Binary search reduces the number of feasibility fit checks. However, is it possible to obtain the final result in a single feasibility fit check—for example, by assigning component sets one by one until we can no longer assign a complete component set?

Ah, I understand now. This was something I had considered early on before implementing, but the reason why I decided against it was because this breaks one of the assumptions of FFD, which requires the items to be globally sorted in descending order. That way you pack the largest items first and can fit in the smaller easier to pack items in the left over spaces. If you go set by set, this would be more of a First-Fit algorithm, which can lead to suboptimal packing. 

Generally the number of bins used by each algorithm is:

FF uses <= 1.7 * Optimal
FFD uses <= 11/9 * Optimal (which is a much tighter bound)

But if I'm being honest, we're working with relatively low dimensionality here (most cases will be 2D with only CPU + Memory defined). I spent some time trying to think of a case in which the per-set packing would result in a suboptimal scenario compared to FFD and couldn't. :) And the suboptimal packing really only starts to become a problem with higher dimension vector packing, where you'll see resource fragmentation. 

`TL;DR` - I was optimizing for a more correct estimate rather than efficiency. If we think this algorithm is too slow in practice perhaps we can change it to a per-set packing? What do you think?

zhzhuang-zju:
> I was optimizing for a more correct estimate rather than efficiency. If we think this algorithm is too slow in practice perhaps we can change it to a per-set packing? What do you think?

That makes sense—I see your point now. From an algorithmic perspective, your current implementation does yield more precise results, but the computational complexity is indeed a concern. I’d probably lean toward `per-set packing` for the following reasons:

- It’s simpler to implement and reason about.  
- The scheduling process itself is sequential—**one set at a time**—so the outcome of `per-set packing` may better align with the actual scheduling semantics (not sure, raise this as a point for discussion).

WDYT? @mszacillo @RainbowMango @seanlaii

zhzhuang-zju:
BTW, regarding `sorting` components, I also considered this when implementing the NodeResource plugin but eventually gave it up. The reason is that component resource requests are multi-dimensional, making it hard to definitively determine which one is "larger". Additionally, as bin packing progresses, the available resources on nodes keep changing—so would we need to continuously re-sort the components as well?

mszacillo:
> I’d probably lean toward per-set packing for the following reasons:

Understood - I agree with your points! I can change around the implementation to use a per-set method.

pkg/estimator/client/general.go
```diff
@@ -160,13 +160,255 @@ func (ge *GeneralEstimator) maxAvailableComponentSets(cluster *clusterv1alpha1.C
...
+
+	// Keep a working copy of model node capacities and place one full component set at a time
+	work := make([]modelNode, len(nodes))
+	for i := range nodes {
+		capCopy := make(map[corev1.ResourceName]int64, len(nodes[i].cap))
+		for k, v := range nodes[i].cap {
+			capCopy[k] = v
+		}
+		work[i] = modelNode{cap: capCopy}
+	}
```
zhzhuang-zju: Could you remind me why we need to copy `nodes` as `work` instead of using `nodes` directly?

pkg/estimator/client/general.go
```diff
@@ -160,13 +160,255 @@ func (ge *GeneralEstimator) maxAvailableComponentSets(cluster *clusterv1alpha1.C
...
+		avail := capacity[k]
+		if avail <= 0 {
+			return 0
+		}
+		bound := avail / req
+		if bound < limit {
+			limit = bound
+		}
+	}
+	if limit == math.MaxInt64 {
```
zhzhuang-zju, line 395: Reaching this branch usually happens when all demands are zero. In this case, would it be more appropriate to return `MaxInt64`? Because, in theory, a member cluster can indeed accommodate a very large number of workloads with zero resource demand.

pkg/estimator/client/general.go
```diff
@@ -160,13 +160,255 @@ func (ge *GeneralEstimator) maxAvailableComponentSets(cluster *clusterv1alpha1.C
...
+) (int64, error) {
+	if upperBound <= 0 {
+		return 0, nil
+	}
+
+	// Compressed one-set: per-kind (identical replicas grouped)
+	oneSetKinds := expandKindsOneSet(components)
+	if len(oneSetKinds) == 0 {
+		// No pods in a set -> nothing to schedule under models
+		return 0, nil
```
zhzhuang-zju: The same as https://github.com/karmada-io/karmada/pull/6912#discussion_r2548795838

pkg/estimator/client/general.go
mszacillo: Oops, not necessary. Forgot to clean this part up.

pkg/estimator/client/general.go
mszacillo: In this case no, because the helper method is used to determine maxFit of the component using the capacity of the cluster. We precheck cases in which the workloads have zero demand in the `maxAvailableComponentSets` method and return the podBound. Callers of `getMaximumSetsBasedOnResourceModels` will have resource requirements so it does not make sense to return maxInt here - that would throw off the calculation done by placeOneSet.

pkg/estimator/client/general.go
```diff
@@ -160,13 +160,253 @@ func (ge *GeneralEstimator) maxAvailableComponentSets(cluster *clusterv1alpha1.C
...
+
+	for i := range oneSetKinds {
+		oneSetKinds[i].score = demandScoreNormalized(oneSetKinds[i].dem, totals)
+	}
+	sort.Slice(oneSetKinds, func(i, j int) bool {
+		if oneSetKinds[i].score == oneSetKinds[j].score {
+			return demandSum(oneSetKinds[i].dem) > demandSum(oneSetKinds[j].dem)
+		}
+		return oneSetKinds[i].score > oneSetKinds[j].score
+	})
```
RainbowMango, line 196: @mszacillo, I'm not sure if I understand correctly.
Can you remind me why we need to sort the components here?

Does a component requesting scarce resources receive higher priority, meaning it gets placed first? If yes, what's the benefit?

RainbowMango:
@mszacillo can you help rebase this PR? As we introduced some E2E tests regarding multiple components scheduling, so that we can see the test results from here.

pkg/estimator/client/general.go
mszacillo: Generally its recommended to play larger and harder to place replicas first and fit smaller replicas into the remaining spaces, so that we can more efficiently pack them into the set of available nodes. It's a small optimization which provides better results than if we just packed replicas in random order.

mszacillo:
> can you help rebase this PR? As we introduced some E2E tests regarding multiple components scheduling, so that we can see the test results from here.

Of course! Rebased and pushed the changes.

RainbowMango:
Thanks, but the unit tests is failing:
```
=== Failed
=== FAIL: pkg/estimator/client TestGetMaximumSetsBasedOnResourceModels/Multi-component_JM/TM_across_grades_(CPU+Mem) (0.00s)
    general_test.go:556: 
        	Error Trace:	/home/runner/work/karmada/karmada/pkg/estimator/client/general_test.go:556
        	Error:      	Not equal: 
        	            	expected: 3
        	            	actual  : 2
        	Test:       	TestGetMaximumSetsBasedOnResourceModels/Multi-component_JM/TM_across_grades_(CPU+Mem)
    --- FAIL: TestGetMaximumSetsBasedOnResourceModels/Multi-component_JM/TM_across_grades_(CPU+Mem) (0.00s)

=== FAIL: pkg/estimator/client TestGetMaximumSetsBasedOnResourceModels (0.00s)

```

mszacillo:
Yes I am investigating this currently. Didn't want to comment until I root caused it, but seems like there is some nondeterminism after I changed the algorithm to use a per-set packing. I wonder if I can make the test less strict.

pkg/estimator/client/general.go
RainbowMango: OK, I see. Shall we add more comments to explain the intention?

mszacillo:
@RainbowMango The recent changes to `buildModelNodes` made the order of the nodes nondeterministic. I'll fix this and update the branch.

mszacillo:
@RainbowMango I added a sort for the grades to make sure they are ordered when constructing the nodes. This could be optimized, but do you know if there is validation for grades users can define? Do they have to start from 0 and increment by 1? Or can users define random numbered grades if they wanted - for example {2, 6, 19, 22}?

RainbowMango:
The minimum amount of resources from first grade should start with `0`, and the maximum amount of resources from the last grade should end with `math.MaxInt64`—no restriction on the increment between grades. 
https://github.com/karmada-io/karmada/blob/309c67701b6ff93069c685e8a9e7d2dfa06a00fa/pkg/apis/cluster/validation/validation.go#L198-L231

pkg/estimator/client/general.go
```diff
@@ -160,13 +160,257 @@ func (ge *GeneralEstimator) maxAvailableComponentSets(cluster *clusterv1alpha1.C
...
+	for _, grade := range grades {
+		tmpl, cnt := capsByGrade[uint(grade)], countByGrade[uint(grade)] // #nosec G115: integer overflow conversion int -> uint
+		if tmpl == nil || cnt == 0 {
+			continue
+		}
+		for range cnt {
+			capCopy := make(map[corev1.ResourceName]int64, len(tmpl))
+			for k, v := range tmpl {
+				capCopy[k] = v
+			}
```
```suggestion
			capCopy := maps.Clone(tmpl)
```
I just realized that now we have a built-in function for cloning maps since [Go1.21](https://go.dev/doc/go1.21#maps).

zhzhuang-zju:
Thanks 
/lgtm

PR #6919, opened 2025-11-09
feat: introduce asyncPriorityWorker in resourceDetector Processor

zach593:
**What type of PR is this?**
/kind feature

In the previous work, we added the priority queue featuregate for controllers driven by controller-runtime.
However, this capability could not take full effect, because some key controllers in controller-manager are driven by our self-developed asyncWorker.
Therefore, we need to implement the same mechanism in asyncWorker to achieve full coverage of the "priority queue" feature.

This PR adds priority queue support to asyncWorker and introduces the related capability in resourceDetector for processing resourceTemplate (activated via the priorityQueue featuregate).

This feature will be progressively rolled out to other controllers in upcoming updates.

Fixes #

test report:

I temporarily added time.Sleep(time.Second) in the detector and limited its concurrency to 1 to control the processing speed.
After adding 5,000 deployments and their corresponding propagation policies and completing synchronization, I restarted the controller.

image
image

It can be observed that while there was still a backlog in the queue, after manually modifying the corresponding object, the detector processed it very quickly.
As for the backlog of the `priority: 0` queue, it shouldn’t be that large — I suspect there might be an issue here, and I’ll check it later.

```release-note
After enabling the `ControllerPriorityQueue` feature gate, the asyncWorker uses a priority queue based implementation, which affects the processing order of items in the resource detector and causes the monitoring metric `workqueue_depth` to be split into multiple series.
```

### DIFF
--- a/pkg/util/worker.go
+++ b/pkg/util/worker.go
@@ -185,11 +185,20 @@ func (w *asyncWorker) AddWithOpts(opts AddOpts, item ...any) {
 		return
 	}
 
-	pq, ok := w.queue.(priorityqueue.PriorityQueue[any])
+	pq, ok := w.queue.(interface {
+		AddWithOpts(o priorityqueue.AddOpts, Items ...any)
+	})
 	if !ok {
 		klog.Warningf("queue is not priority queue, fallback to normal queue, queueName: %s", w.name)
 		for _, it := range item {
-			w.queue.Add(it)
+			switch {
+			case opts.After > 0:
+				w.queue.AddAfter(it, opts.After)
+			case opts.RateLimited:
+				w.queue.AddRateLimited(it)
+			default:
+				w.queue.Add(it)
+			}
 		}
 		return
 	}
--- a/pkg/util/worker_test.go
+++ b/pkg/util/worker_test.go
@@ -269,7 +269,30 @@ func assertUntil(maxDuration time.Duration, assertion func() error) error {
 }
 
 func Test_asyncWorker_AddWithOpts(t *testing.T) {
-	t.Run("addAfter", func(t *testing.T) {
+	t.Run("AddAfter, pq not enabled", func(t *testing.T) {
+		const name = "fake_node"
+		const duration = 1 * time.Second
+
+		worker := newTestAsyncWorker(nil, false)
+
+		start := time.Now()
+		worker.AddWithOpts(AddOpts{After: duration}, name)
+
+		item, _ := worker.queue.Get()
+		end := time.Now()
+
+		if name != item {
+			t.Errorf("Added Item: %v, want: %v", item, name)
+		}
+
+		elapsed := end.Sub(start)
+		if elapsed < duration {
+			t.Errorf("Added Item should be dequeued after %v, but the actually elapsed time is %v.",
+				duration.String(), elapsed.String())
+		}
+	})
+
+	t.Run("addAfter, pq enabled", func(t *testing.T) {
 		const name = "fake_node"
 		const duration = 1 * time.Second
 


RainbowMango:
> As for the backlog of the priority: 0 queue, it shouldn’t be that large — I suspect there might be an issue here, and I’ll check it later.

Yes, that's suspicious. You just updated one resource template, right? Can we add a log before enqueuing the items? So that we can know what items are in the queue.

zach593:
> It can be observed that while there was still a backlog in the queue, after manually modifying the corresponding object, the detector processed it very quickly.
> As for the backlog of the `priority: 0` queue, it shouldn’t be that large — I suspect there might be an issue here, and I’ll check it later.

Reason:

After creating the cluster using `hack/local-up-karmada.sh`, `karmada-apiserver` will contain some basic resources that support the cluster, such as CRDs and APIServices. These resources will enter the `resourceDetector` queue. However, when entering reconcile, they lack corresponding PP or CPP, resulting in an error and a second entry into the priority queue with a `ratelimit` option, this time with a priority of 0. 

https://github.com/karmada-io/karmada/blob/463fcca57d65e71eae02bd753bd610cb5b518433/pkg/detector/policy.go#L105-L107

The `ratelimit` option is implemented by calling `When` and adding `readyAt`. 

https://github.com/karmada-io/karmada/blob/463fcca57d65e71eae02bd753bd610cb5b518433/vendor/sigs.k8s.io/controller-runtime/pkg/controller/priorityqueue/priorityqueue.go#L144-L156

As seen in the `less` implementation, objects with `readyAt` are placed at the back of the queue, even if they might have higher priority. (before https://github.com/karmada-io/karmada/pull/6929)

https://github.com/karmada-io/karmada/blob/27ca69094a8a070098f7581da62f65beb8bc84b3/vendor/sigs.k8s.io/controller-runtime/pkg/controller/priorityqueue/priorityqueue.go#L367-L382

Therefore, although these items are placed at the end of the queue, their priority is 0. This results in a situation where, after modifying an object, it is immediately processed by the detector, but the queue metrics still appear to show a backlog of items with a priority of 0.

After https://github.com/karmada-io/karmada/pull/6929, this has already been fixed in the code shown below. After rebasing, the problem has not occurred again.

https://github.com/karmada-io/karmada/blob/463fcca57d65e71eae02bd753bd610cb5b518433/vendor/sigs.k8s.io/controller-runtime/pkg/controller/priorityqueue/priorityqueue.go#L421-L436

zach593:
Updated Test Report:

karmada-controller-manager configuration: 

image

Add a `time.Sleep(time.Second)` before logging in the resourceDetector:

image

After adding 5,000 deployments and their propagation policies and waiting for synchronization to complete, restart the controller.

And the result: 

image

RainbowMango:
> After https://github.com/karmada-io/karmada/pull/6929, this has already been fixed in the code shown below. After rebasing, the problem has not occurred again.

Nice!

pkg/util/worker.go
```diff
@@ -77,16 +108,28 @@ type Options struct {
 	KeyFunc            KeyFunc
 	ReconcileFunc      ReconcileFunc
 	RateLimiterOptions ratelimiterflag.Options
+	UsePriorityQueue   bool
```
RainbowMango, line 111: This option is only used to allow the legacy `AsyncWorker` in the system to gradually phase out and transition to `AsyncPriorityWorker`. Once the migration is fully completed, this parameter will be deleted, correct?

pkg/detector/detector.go
```diff
@@ -310,12 +311,17 @@ func (d *ResourceDetector) EventFilter(obj interface{}) bool {
 }
 
 // OnAdd handles object add event and push the object to queue.
-func (d *ResourceDetector) OnAdd(obj interface{}) {
+func (d *ResourceDetector) OnAdd(obj interface{}, isInitialList bool) {
```
RainbowMango: TODO: Need to clarify how the `isInitialList` is set.

pkg/detector/detector.go
```diff
@@ -1492,9 +1498,22 @@ func (d *ResourceDetector) applyReplicaInterpretation(object *unstructured.Unstr
 // Therefore, only set ResourceChangeByKarmada in lazy activation mode.
 // For more details, see: https://github.com/karmada-io/karmada/issues/5996.
 func (d *ResourceDetector) enqueueResourceTemplateForPolicyChange(key keys.ClusterWideKey, pref policyv1alpha1.ActivationPreference) {
-	if util.IsLazyActivationEnabled(pref) {
-		d.Processor.Add(keys.ClusterWideKeyWithConfig{ClusterWideKey: key, ResourceChangeByKarmada: true})
-		return
-	}
-	d.Processor.Add(keys.ClusterWideKeyWithConfig{ClusterWideKey: key})
+	enqueueKey := keys.ClusterWideKeyWithConfig{ClusterWideKey: key, ResourceChangeByKarmada: util.IsLazyActivationEnabled(pref)}
+	// Use of low priority here is based on the following reasons:
```
RainbowMango, line 1502: @zhzhuang-zju Please confirm if this breaks the intention of #6674.

pkg/util/worker.go
```diff
@@ -122,6 +179,24 @@ func (w *asyncWorker) AddAfter(item interface{}, duration time.Duration) {
 	w.queue.AddAfter(item, duration)
 }
 
+func (w *asyncWorker) AddWithOpts(opts AddOpts, item ...any) {
+	if item == nil {
+		klog.Warningf("Ignore nil item from queue")
+		return
+	}
+
+	pq, ok := w.queue.(priorityqueue.PriorityQueue[any])
```
RainbowMango: My Goland reported a warning regarding `priorityqueue.PriorityQueue[any]`:

Impossible interface assertion from `interface {...}` to `priorityqueue.PriorityQueue[any]`. Ambiguous: `AddRateLimited`, `NumRequeues`, `Forget` 
 Inspection info: Reports impossible interface-to-interface type assertions.

Checks for type assertions `v.(T)` and corresponding type-switch cases in which the static type `V` of `v` is the interface that cannot possibly implement the target interface `T`. This occurs when `V` and `T` contain methods with the same name but different signatures.
Example:
```golang
var v interface {
  Read()
}
_ = v.(io.Reader)
```
The `Read` method in `v` has a different signature than the `Read` method in `io.Reader`, so this assertion cannot succeed.

pkg/detector/detector.go
zhzhuang-zju: The assignment to `enqueueKey` here behaves as expected.

pkg/util/worker.go
zach593: right.

pkg/util/worker.go
zach593: Although I think this warning is due to Goland's insufficient static checking capabilities (after all, this line of code can return true or false correctly), I still made some implementation adjustments.

pkg/util/worker.go
RainbowMango: Got it.

pkg/detector/detector.go
RainbowMango: Got it.

pkg/util/worker.go
RainbowMango: Looks good.

RainbowMango (commented):
/lgtm
/gemini review

RainbowMango:
```
>>> git ls-remote --get-url origin
Error: failed linting charts: failed processing charts

------------------------------------------------------------------------------------------------------------------------
 ✖︎ karmada => (version: "0.0.0", path: "charts/karmada") > failed validating maintainer "chaosi-zju": 503 Service Unavailable
------------------------------------------------------------------------------------------------------------------------
failed linting charts: failed processing charts
Error: Process completed with exit code 1.
```
/retest

PR #6921, opened 2025-11-11
Add the resource UID to the binding.spec.resource in the dependencies-distributor

XiShanYongYe-Chang:
**What type of PR is this?**

/kind cleanup

In the dependencies-distributor scenario, we currently do not add the UID information of the resource to the `binding.spec.resource` field for the attached resources. However, the detector has [logic processing](https://github.com/karmada-io/karmada/blob/29536305bcc4c965a3538241bc60fb468d24d377/pkg/detector/detector.go#L811-L818) that could potentially cause repeated occurrences of this field, leading to unnecessary binding updates.

Fixes #

```release-note
NONE
```


RainbowMango (commented):
/assign

PR #3827, opened 2023-07-21
Make some cleanup in dependencies distributor

XiShanYongYe-Chang:
**What type of PR is this?**

/kind cleanup

**What this PR does / why we need it**:

Make some cleanup in the dependencies distributor:

- Change some unnecessary exposed member variables to internal variables.
- Delete some redundant code.
- Modify inappropriate method or variable names.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #6931, opened 2025-11-17
Fix the issue that attached resource changes were not synchronized to the cluster in dependencies distributor

### LINKED ISSUE #6924: Attached resource changes were not synchronized to the cluster in dependencies distributor
**What happened**:

In the dependency distributor scenario, workload resources depend on multiple resources of the same kind. When one of these resources changes, there may be a case of unsynchronization.

For example, a deployment depends on two ConfigMap resources, `cm-01` and `cm-02`. After the user modifies `cm-02`, Karmada does not synchronize this change to the member clusters.

Resource changes can be properly synchronized to the member cluster.

As the example suggests, multiple tests can be conducted.

In current logic, when a name mismatch is detected, it returns directly instead of continuing:

https://github.com/karmada-io/karmada/blob/e2c4b596a5da442fc0dbeab9f9063d8db8669208/pkg/dependenciesdistributor/dependencies_distributor.go#L234-L239

**Environment**:
- Karmada version:
- Others:

XiShanYongYe-Chang:
**What type of PR is this?**

/kind bug

Fixes #6924

```release-note
`karmada-controller-manager`: Fixed the issue that attached resource changes were not synchronized to the cluster in the dependencies distributor
```

### DIFF
--- a/pkg/dependenciesdistributor/dependencies_distributor.go
+++ b/pkg/dependenciesdistributor/dependencies_distributor.go
@@ -183,13 +183,18 @@ func (d *DependenciesDistributor) reconcileResourceTemplate(key util.QueueKey) e
 		return err
 	}
 
+	var errs []error
 	for i := range readonlyBindingList.Items {
 		binding := &readonlyBindingList.Items[i]
 		if !binding.DeletionTimestamp.IsZero() {
 			continue
 		}
 
-		matched := matchesWithBindingDependencies(resourceTemplateKey, binding)
+		matched, err := matchesWithBindingDependencies(resourceTemplateKey, binding)
+		if err != nil {
+			errs = append(errs, err)
+			continue
+		}
 		if !matched {
 			continue
 		}
@@ -198,48 +203,52 @@ func (d *DependenciesDistributor) reconcileResourceTemplate(key util.QueueKey) e
 		d.genericEvent <- event.TypedGenericEvent[*workv1alpha2.ResourceBinding]{Object: binding}
 	}
 
-	return nil
+	return utilerrors.NewAggregate(errs)
 }
 
 // matchesWithBindingDependencies tells if the given object(resource template) is matched
 // with the dependencies of independent resourceBinding.
-func matchesWithBindingDependencies(resourceTemplateKey *LabelsKey, independentBinding *workv1alpha2.ResourceBinding) bool {
+func matchesWithBindingDependencies(resourceTemplateKey *LabelsKey, independentBinding *workv1alpha2.ResourceBinding) (bool, error) {
 	dependencies, exist := independentBinding.Annotations[util.DependenciesAnnotationKey]
 	if !exist {
-		return false
+		return false, nil
 	}
 
 	var dependenciesSlice []configv1alpha1.DependentObjectReference
 	err := json.Unmarshal([]byte(dependencies), &dependenciesSlice)
 	if err != nil {
-		// If unmarshal fails, retrying with an error return will not solve the problem.
-		// It will only increase the consumption by repeatedly listing the binding.
-		// Therefore, it is better to print this error and ignore it.
 		klog.Errorf("Failed to unmarshal binding(%s/%s) dependencies(%s): %v",
 			independentBinding.Namespace, independentBinding.Name, dependencies, err)
-		return false
+		return false, err
 	}
 	if len(dependenciesSlice) == 0 {
-		return false
+		return false, nil
 	}
 
+	var errs []error
 	for _, dependency := range dependenciesSlice {
 		if resourceTemplateKey.GroupVersion().String() == dependency.APIVersion &&
 			resourceTemplateKey.Kind == dependency.Kind &&
 			resourceTemplateKey.Namespace == dependency.Namespace {
 			if len(dependency.Name) != 0 {
-				return dependency.Name == resourceTemplateKey.Name
+				if dependency.Name == resourceTemplateKey.Name {
+					return true, nil
+				}
+				continue
 			}
 			var selector labels.Selector
 			if selector, err = metav1.LabelSelectorAsSelector(dependency.LabelSelector); err != nil {
 				klog.Errorf("Failed to converts the LabelSelector of binding(%s/%s) dependencies(%s): %v",
 					independentBinding.Namespace, independentBinding.Name, dependencies, err)
-				return false
+				errs = append(errs, err)
+				continue
 			}
-			return selector.Matches(labels.Set(resourceTemplateKey.Labels))
+			if selector.Matches(labels.Set(resourceTemplateKey.Labels)) {
+				return true, nil
+			}
 		}
 	}
-	return false
+	return false, utilerrors.NewAggregate(errs)
 }
 
 // Reconcile performs a full reconciliation for the object referred to by the Request.
--- a/pkg/dependenciesdistributor/dependencies_distributor_test.go
+++ b/pkg/dependenciesdistributor/dependencies_distributor_test.go
@@ -369,15 +369,16 @@ func Test_reconcileResourceTemplate(t *testing.T) {
 	}
 }
 
-func Test_dependentObjectReferenceMatches(t *testing.T) {
+func Test_matchesWithBindingDependencies(t *testing.T) {
 	type args struct {
 		objectKey        *LabelsKey
 		referenceBinding *workv1alpha2.ResourceBinding
 	}
 	tests := []struct {
-		name string
-		args args
-		want bool
+		name    string
+		args    args
+		want    bool
+		wantErr bool
 	}{
 		{
 			name: "test custom resource",
@@ -398,7 +399,8 @@ func Test_dependentObjectReferenceMatches(t *testing.T) {
 					}},
 				},
 			},
-			want: true,
+			want:    true,
+			wantErr: false,
 		},
 		{
 			name: "test configmap",
@@ -419,7 +421,8 @@ func Test_dependentObjectReferenceMatches(t *testing.T) {
 					}},
 				},
 			},
-			want: true,
+			want:    true,
+			wantErr: false,
 		},
 		{
 			name: "test labels",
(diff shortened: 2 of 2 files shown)


pkg/dependenciesdistributor/dependencies_distributor.go
```diff
@@ -235,15 +235,23 @@ func matchesWithBindingDependencies(resourceTemplateKey *LabelsKey, independentB
...
 			}
 			var selector labels.Selector
 			if selector, err = metav1.LabelSelectorAsSelector(dependency.LabelSelector); err != nil {
 				klog.Errorf("Failed to converts the LabelSelector of binding(%s/%s) dependencies(%s): %v",
 					independentBinding.Namespace, independentBinding.Name, dependencies, err)
-				return false
+				// If convert LabelSelector failed, retrying with an error return will not solve the problem.
+				// It will only increase the consumption by repeatedly listing the binding.
+				// Therefore, it is better to print this error and ignore it.
+				continue
```
zhzhuang-zju, line 244: I agree. However, we should double-check: could a failure to convert LabelSelector to a selector cause intermittent test failures?

XiShanYongYe-Chang: Sorry, I didn't understand what you meant. Could you explain that in more detail?

zhzhuang-zju: I mean, if a failure to convert `LabelSelector` can be resolved by retrying, then we should return an error and trigger a retry. However, if the conversion failure is deterministic, then the current handling approach is perfectly reasonable.

RainbowMango (commented):
/assign

pkg/dependenciesdistributor/dependencies_distributor.go
RainbowMango: @XiShanYongYe-Chang I want to know if there is any historical reason for ignoring this error? (I see the `Unmarshal` above also ignored the error.)

If this is a persistent error, ignoring it will hide the real bug, it's better to expose the issue as early as possible. We can predict that even if we retry this error, the cost is just multiple attempts.

Can we add some validation checks in the webhook? If it can pass the validation, then this is likely an intermittent error, and we can retry with more confidence.

XiShanYongYe-Chang: > I want to know if there is any historical reason for ignoring this error? (I see the Unmarshal above also ignored the error.)

There was no particular historical reason; the initial revision of the PR was: #3827. The reason for the modification is as stated in the comment.

> If this is a persistent error, ignoring it will hide the real bug, it's better to expose the issue as early as possible. We can predict that even if we retry this error, the cost is just multiple attempts.

The ignore here is intended to avoid wasting subsequent useless reconciles, not to cover up any mistakes. Once an error occurs here, it indicates that the label selector was written incorrectly (for the unmarshal operation mentioned above, no error would occur unless the user modifies the annotationin the binding object). Retrying cannot resolve the issue and would waste the execution of the reconcile process.

> Can we add some validation checks in the webhook? 

I think it's feasible. We can modify [VerifyDependencies](https://github.com/karmada-io/karmada/blob/152433f7c493752653bf5e3928276f698de5fab3/pkg/util/interpreter/validation/validation.go#L28-L40) to perform correctness checks on the configv1alpha1.DependentObjectReference returned by the resource interpreter. If possible, we can create a new PR to address this issue separately.

Do we also need to validate the content of the unmarshal operation mentioned above? For example, validating the update operation of the binding object in karmada-webhook. This is done to prevent users from mistakenly modifying the annotation dependencies annotation `dependenciesAnnotationKey` key on the binding object. I'm not sure how much value this adds.

> If it can pass the validation, then this is likely an intermittent error, and we can retry with more confidence.

The error here is not an intermittent one; once it occurs, it will consistently be present. Therefore, I think that the current modification is more appropriate. Adding validation operations can further prevent the issue from arising.

RainbowMango: @zhzhuang-zju any thoughts?

zhzhuang-zju: My feeling is that handling errors inside a function—especially by silently ignoring them—smells bad. If an error truly needs to be ignored, it's better for the caller to make that decision explicitly.

For example:

```go
func matchesWithBindingDependencies(resourceTemplateKey *LabelsKey, independentBinding *workv1alpha2.ResourceBinding) bool {
	err := json.Unmarshal([]byte(dependencies), &dependenciesSlice)
	if err != nil {
		// If unmarshal fails, retrying with an error return will not solve the problem.
		// It will only increase the consumption by repeatedly listing the binding.
		// Therefore, it is better to print this error and ignore it.
		klog.Errorf("Failed to unmarshal binding(%s/%s) dependencies(%s): %v",
			independentBinding.Namespace, independentBinding.Name, dependencies, err)
		return false
	}
    // ...
}
```

This could be refactored to:

```go
func matchesWithBindingDependencies(resourceTemplateKey *LabelsKey, independentBinding *workv1alpha2.ResourceBinding) (bool, error) {
	err := json.Unmarshal([]byte(dependencies), &dependenciesSlice)
	if err != nil {
		return false, err
	}
    // ...
}

func (d *DependenciesDistributor) reconcileResourceTemplate(key util.QueueKey) error {
	matched, err := matchesWithBindingDependencies(resourceTemplateKey, binding)
	// The caller decides whether to ignore the error.
	if !matched || err != nil {
		continue
	}
}
```

Regarding whether to ignore the error: if we’re certain that retries won’t resolve the issue, then ignoring the error might be acceptable. Also, it would be good to reduce such errors by adding validation checks.

XiShanYongYe-Chang: > My feeling is that handling errors inside a function—especially by silently ignoring them—smells bad. If an error truly needs to be ignored, it's better for the caller to make that decision explicitly.

I don't think there's a big difference here. Whether it's the caller or the callee, neither is an externally exposed solution; both are internal processing logic, so I don't feel any particular issue with it.
In addition, from the perspective of the current function's functionality, the current change is also fine, and it actually better matches the function's name in terms of return value.
I'm not saying that the method you pointed out is bad. I think these are two different approaches, and neither is superior to the other.

XiShanYongYe-Chang: > Regarding whether to ignore the error: if we’re certain that retries won’t resolve the issue, then ignoring the error might be acceptable. 

This is certain; there will be no situation where success occurs this time and failure next time.

RainbowMango: > My feeling is that handling errors inside a function—especially by silently ignoring them—smells bad.

+1

> I don't think there's a big difference here. Whether it's the caller or the callee, neither is an externally exposed solution; both are internal processing logic, so I don't feel any particular issue with it.

+1

I agree with both of you, but in my opinion... 
If we can confirm that the controller can always set the correct value for the `resourcebinding.karmada.io/dependencies` annotation, then we should return an error here instead of ignoring it.

To prevent manual modification errors, we should add validation to the webhook.

XiShanYongYe-Chang: > If we can confirm that the controller can always set the correct value for the resourcebinding.karmada.io/dependencies annotation, 

It is certain that marshal and unmarshal will not produce errors without external interference.

> then we should return an error here instead of ignoring it.

I think it's acceptable to either return or not return an error. If we further enhance the validation logic, we can directly determine that this error will not occur, thus bypassing the need to handle `err` altogether.

XiShanYongYe-Chang: Putting aside the validation logic, if the label selector fails to be parsed and an error is returned directly, retries through reconciliation will continue to fail. As a result, the remaining dependent resources will no longer be able to match, which will also lead to the issue described in the problem.

XiShanYongYe-Chang: /cc @RainbowMango 
Does this logic need further processing?

RainbowMango: I will take a look tomorrow.

RainbowMango: Yes, I think so. 
The responsibility of the function `matchesWithBindingDependencies` tells if given resource template is required by a specific ResourceBinding; in case of any error arising, it shouldn't swallow it, it should return the error instead.

XiShanYongYe-Chang: I still maintain my viewpoint that not all errors necessarily require a rollback. If this is an occasional client access timeout error, then returning an error and retrying would be effective. However, since this error is bound to occur, retrying would be meaningless.

But in order to smoothly advance the current PR, let me make some adjustments to the PR.

pkg/dependenciesdistributor/dependencies_distributor.go
```diff
@@ -183,13 +183,18 @@ func (d *DependenciesDistributor) reconcileResourceTemplate(key util.QueueKey) e
 		return err
 	}
 
+	var errs []error
```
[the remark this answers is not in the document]
XiShanYongYe-Chang: I do not recommend pre-allocating space for maps; it is generally not a mistake in most cases.

pkg/dependenciesdistributor/dependencies_distributor.go
```diff
@@ -198,48 +203,52 @@ func (d *DependenciesDistributor) reconcileResourceTemplate(key util.QueueKey) e
...
 			if selector, err = metav1.LabelSelectorAsSelector(dependency.LabelSelector); err != nil {
 				klog.Errorf("Failed to converts the LabelSelector of binding(%s/%s) dependencies(%s): %v",
 					independentBinding.Namespace, independentBinding.Name, dependencies, err)
-				return false
+				errs = append(errs, err)
+				continue
+			}
+			if selector != nil && selector.Matches(labels.Set(resourceTemplateKey.Labels)) {
+				return true, nil
 			}
```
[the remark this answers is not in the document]
XiShanYongYe-Chang: ok

pkg/dependenciesdistributor/dependencies_distributor.go
XiShanYongYe-Chang: @RainbowMango updated.

RainbowMango (approved):
/lgtm
/approve

PR #6956, opened 2025-11-25
fix: semgrep-grpc-server-insecure-connection

rahulkhinchi-wq:
This pull request resolves the Semgrep security finding `grpc-server-insecure-connection` by preventing the creation of an insecure gRPC server instance.

## Issue

Semgrep reported that the code was creating a gRPC server without TLS credentials when `CertFile` or `KeyFile` were missing:
`grpc.NewServer()`

Without credentials, the server runs without encryption, which exposes the application to risks such as unencrypted traffic, message tampering, and possible man-in-the-middle attacks.

Rule details:

- **Rule ID:** grpc-server-insecure-connection  
- **File:** pkg/util/grpcconnection/config.go  
- **Line:** 74

## Fix

The function now returns an error if either `CertFile` or `KeyFile` is not provided, instead of silently creating an insecure server.

### Code change

```diff
-   return grpc.NewServer(), nil
+   return nil, fmt.Errorf("TLS credentials are required: both CertFile and KeyFile must be specified for secure gRPC server")
This enforces secure-by-default behavior and blocks insecure gRPC server instantiation.

Impact

Removes insecure default behavior.
Ensures that a secure TLS configuration is always provided.
Aligns gRPC server initialization with recommended security practices.


zhzhuang-zju:
> Semgrep reported that the code was creating a gRPC server without TLS credentials when `CertFile` or `KeyFile` were missing:  
> `grpc.NewServer()`

@rahulkhinchi-wq Thanks for reporting this. Could you share the full Semgrep report by email? (my email address: m17799853869@163.com, Using email is to avoid the accidental exposure of security issues.）
Did it flag any other issues?

Also, I'm a bit skeptical whether this is actually a bug. First, TLS configuration is optional by design. Second, the Karmada-maintained installation tools configure certificates and enable TLS by default.

PR #6965, opened 2025-11-27
enable asyncPriorityWorker in all controllers

zach593:
**What type of PR is this?**

/kind feature

After https://github.com/karmada-io/karmada/pull/6919 is merged, this PR enables the priority-queue functionality for all controllers that use async workers.
Once all controllers fully adopt priority queues, the karmada-controller-manager can still respond promptly to user updates even when a restart causes a backlog in the workqueue.
In theory, this makes rolling updates of the karmada-controller-manager nearly transparent to users, with no perceived downtime.

Fixes #

- Enabling this feature disables the rate-limiting workqueue’s QPS limits.
- controller-runtime has recently enabled the priority-queue capability by default and plans to release it in v0.23 (see https://github.com/kubernetes-sigs/controller-runtime/issues/2374). By the time Karmada reaches v1.17, we can consider promoting this capability to a beta feature and enabling it by default.

Test report:

A cluster was created using hack/local-up-karmada.sh. In this cluster, 5,000 Deployments and their corresponding PropagationPolicies were added and propagated to two member clusters, completing full synchronization.

For the controller-manager, a `time.Sleep(time.Second)` delay was injected at the beginning of each controller’s reconcile loop on top of this PR. The controller-manager was then deployed with the following configuration:

image

After restarting the controller-manager, the following was observed:

img_v3_02se_34facaa4-346e-4727-b9da-e93b3defcd5g

As shown in the screenshots, even while the queue still had a backlog, modifying any Deployment resulted in it being immediately propagated to the member clusters.

```release-note
`karmada-controller-manager`: The `ControllerPriorityQueue` feature gate now applies to all controllers, including async workers. Enable it with `--feature-gates=ControllerPriorityQueue=true`.
```

### DIFF
--- a/pkg/controllers/status/work_status_controller_test.go
+++ b/pkg/controllers/status/work_status_controller_test.go
@@ -36,6 +36,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/uuid"
 	dynamicfake "k8s.io/client-go/dynamic/fake"
 	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/utils/ptr"
 	controllerruntime "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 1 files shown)


pkg/controllers/status/work_status_controller.go
```diff
@@ -565,3 +568,28 @@ func (c *WorkStatusController) eventf(object *unstructured.Unstructured, eventTy
...
+	if deleted, ok := old.(cache.DeletedFinalStateUnknown); ok {
+		// This object might be stale but ok for our current usage.
+		old = deleted.Obj
+		if old == nil {
+			return
+		}
+	}
+	oldObj := old.(runtime.Object)
+	c.worker.Enqueue(oldObj)
+}
```
zach593, line 595: derived from `NewHandlerOnAllEvents()`

pkg/util/worker.go
```diff
@@ -192,10 +193,10 @@ func (w *asyncWorker) AddWithOpts(opts AddOpts, item ...any) {
...
 			switch {
-			case opts.After > 0:
-				w.queue.AddAfter(it, opts.After)
 			case opts.RateLimited:
 				w.queue.AddRateLimited(it)
+			case opts.After > 0:
+				w.queue.AddAfter(it, opts.After)
 			default:
 				w.queue.Add(it)
 			}
```
zach593, line 202: follow the implementation of `priorityQueueWrapper.AddWithOpts()` in controller-runtime.

XiShanYongYe-Chang: I'll post a link for easy viewing:

https://github.com/kubernetes-sigs/controller-runtime/blob/4b46eb04d57ff3bec4c3c05206c46af9aa647a24/pkg/internal/controller/controller.go#L550-L561

XiShanYongYe-Chang (commented):
Thanks a lot~
/lgtm

RainbowMango (approved):
/approve

PR #6977, opened 2025-11-29
feat: add InterpretSchedulingResult operation to ResourceInterpreter

### LINKED ISSUE #3459: Enable Interpreting Customized Scheduling Result in ResourceInterpreter as well
In the point `InterpretReplica` of `ResourceInterpreter`, enable interpreting customized scheduling result as well.  
Or implement a new point to do the same thing in `ResourceInterpreter`.

**Why is this needed**:
Consider about the two scenario as following:  
1. We have some clusters with the same distribution weight and approximate physical resources, and a lot of workload with `replicas = 1` (or the number of replicas can't be evenly divided by weight). We hope to schedule these "remain replicas" into different clusters equally. But up to now, karmada scheduler will only scheduling them into the same cluster if the clusters has the same distribution weight.  
1. When we retain some changes of replicas for workload in member cluster (for example, our HPA-controller works independently in member clusters), we may hope to sync the changes to controll plane, so that we will always have the consistent global perspective, for which it won't disturb our workload no matter when we stop retaining.  

For scenario 1, we may adjust the distribution algorithm to solve the problem, but it is not a general solution, and it doesn't work for scenario 2. We think it will be better to do something in `ResourceInterpreter`.

ljluestc:
This PR implements the feature requested in [GitHub issue #3459](https://github.com/karmada-io/karmada/issues/3459) to enable interpreting customized scheduling results in the ResourceInterpreter.

The issue describes two critical scenarios that need to be addressed:

1. **Remainder Distribution Problem**: When distributing replicas across clusters with equal weights, remainders from uneven division always go to the same cluster. For workloads with `replicas = 1` or when replica counts can't be evenly divided by cluster weights, Karmada currently schedules remainders to the same cluster consistently, leading to imbalanced resource utilization.

2. **Member Cluster Synchronization**: When HPA controllers work independently in member clusters, replica changes need to be synchronized back to the control plane to maintain a consistent global perspective and prevent workload disruption during cluster transitions.

This PR adds a new `InterpretSchedulingResult` operation to the ResourceInterpreter framework that allows customizing how replicas are distributed among clusters through:
- **Declarative Lua scripts** for custom scheduling logic
- **Webhook-based customization** for external scheduling services  
- **Third-party resource support** for extended scheduling capabilities
- **Full backward compatibility** with existing behavior

### **Which issue(s) this PR fixes**

Fixes #3459

### **Special notes for your reviewer**

This implementation provides a flexible framework for custom scheduling algorithms while maintaining full backward compatibility. The feature is opt-in and only activates when explicitly configured.

Key architectural decisions:
- Added new `InterpreterOperationInterpretSchedulingResult` operation type
- Extended `ResourceInterpreter` interface with `InterpretSchedulingResult()` method
- Implemented in all interpreter types (Default, Declarative, Webhook, Third-party)
- Added comprehensive test coverage including unit and integration tests

### **Does this PR introduce a user-facing change?**

```release-note
`ResourceInterpreter`: Added new `InterpretSchedulingResult` operation to enable customizing replica distribution across clusters. This allows:

1. Custom scheduling logic for distributing remainders across clusters instead of always assigning them to the same cluster
2. Synchronization of replica changes from member clusters back to control plane when HPA controllers work independently

Users can now configure custom scheduling behavior through:
- Lua scripts in ResourceInterpreterCustomization
- Webhook services for external scheduling logic
- Third-party resource interpreters

The feature is backward compatible and only activates when explicitly configured.
```

---

### **API Extensions**
- Added `InterpreterOperationInterpretSchedulingResult` operation type to `configv1alpha1`
- Extended `ResourceInterpreter` interface with `InterpretSchedulingResult(object, schedulingResult) ([]TargetCluster, error)` method
- Added `SchedulingResultInterpretation` type for declarative configurations
- Extended webhook request/response attributes to include scheduling results

### **Interpreter Implementations**
- **Default Interpreter**: Returns scheduling result unchanged (maintains backward compatibility)
- **Declarative Interpreter**: Supports Lua scripts for custom scheduling logic
- **Webhook Interpreter**: Enables external services to customize scheduling results
- **Third-party Interpreter**: Supports third-party resource scheduling customization

### **Lua VM Integration**
- Added `InterpretSchedulingResult()` method to Lua VM for script execution
- Implemented `ConvertLuaResultToTargetClusters()` for Lua-to-Go conversion
- Full Lua script support for complex scheduling algorithms


XiShanYongYe-Chang:
Hi @ljluestc, thank you for participating in Karmada.

I would like to know, apart from the issue you mentioned, what is your user scenario?

PR #6987, opened 2025-12-01
fix backoffQ of internal scheduling queue using the wrong sorting function

### LINKED ISSUE #6986: [Scheduler] backoffQ uses incorrect sorting function, potentially blocking ready-to-retry bindings
**What happened**:
I was reading source code of [scheduling_queue.go](https://github.com/karmada-io/karmada/blob/master/pkg/scheduler/internal/queue/scheduling_queue.go#L139) and noticed that `backoffQ` is initialized used the `Less` function:

```go
backoffQ: heap.NewWithRecorder(BindingKeyFunc, Less, metrics.NewBackoffBindingsRecorder())
```

The `Less` function (defined in [types.go](https://github.com/karmada-io/karmada/blob/master/pkg/scheduler/internal/queue/types.go#L64)) sorts bindings based on *Priority* and *Timestamp*(added to scheduling queue):

```go
func Less(bInfo1, bInfo2 *QueuedBindingInfo) bool {
	p1 := bInfo1.Priority
	p2 := bInfo2.Priority
	return (p1 > p2) || (p1 == p2 && bInfo1.Timestamp.Before(bInfo2.Timestamp))
}
```

However, `backoffQ` is designed to hold bindings that are waiting for a backoff period to expire. Bindings which have completed backoff are poped from `backoffQ` heap and will be moved to `ActiveQ` for scheduling. 
The flushing logic in `flushBackoffQCompleted` relies on the heap property to efficiently find expired bindings: 

```go
func (bq *prioritySchedulingQueue) flushBackoffQCompleted() {
    // ...
    for {
        bInfo, ok := bq.backoffQ.Peek()
        // ...
        // If the head of the heap is still backing off, we assume subsequent elements are too
        // and break the loop.
        if bq.isBindingBackingoff(bInfo) {
            break
        }
        // ...
    }
}
```

`backoffQ` should be sorted by Backoff Expiry Time. If `backoffQ` is sorted by priority, a high-priority binding with a long backoff duration could sit at the top of the heap. This would cause `flushBackoffQCompleted` to `break` early, blocking other lower-priority bindings that have already completed their backoff period from being moved to `activeQ`.

We can assume a test case:

1. Create a high-priority Binding A that fails scheduling and enters backoffQ with a long backoff duration (e.g., 10s).
2. Create a low-priority Binding B that fails scheduling and enters backoffQ with a short backoff duration (e.g., 1s).
3. Because A has higher priority, it will be at the top of backoffQ.
4. Wait for 2 seconds. Binding B is ready to retry, but flushBackoffQCompleted checks A, sees it's not ready, and stops. Binding B is blocked until A is ready.

We should implement a Less function for `backoffQ` that compares the backoff expiry time of two bindings.

**Environment**:
- Karmada version: master branch
- Others:

rayo1uo:
**What type of PR is this?**

/kind bug

[Scheduler] backoffQ uses incorrect sorting function, potentially blocking ready-to-retry bindings

Fixes #6986 

```release-note
karmada-scheduler: Fixes the bug in the backoff queue where the sorting function was incorrect, potentially causing high-priority items with long backoffs to block lower-priority items.
```

### DIFF
--- a/pkg/scheduler/internal/queue/scheduling_queue.go
+++ b/pkg/scheduler/internal/queue/scheduling_queue.go
@@ -22,6 +22,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog/v2"
+	"k8s.io/utils/clock"
 
 	"github.com/karmada-io/karmada/pkg/scheduler/internal/heap"
 	metrics "github.com/karmada-io/karmada/pkg/scheduler/metrics/queue"
@@ -132,13 +133,15 @@ func NewSchedulingQueue(opts ...Option) SchedulingQueue {
 
 	bq := &prioritySchedulingQueue{
 		stop:                          make(chan struct{}),
+		clock:                         clock.RealClock{},
 		bindingInitialBackoffDuration: options.bindingInitialBackoffDuration,
 		bindingMaxBackoffDuration:     options.bindingMaxBackoffDuration,
 		bindingMaxInUnschedulableBindingsDuration: options.bindingMaxInUnschedulableBindingsDuration,
 		activeQ:               NewActiveQueue(metrics.NewActiveBindingsRecorder()),
-		backoffQ:              heap.NewWithRecorder(BindingKeyFunc, Less, metrics.NewBackoffBindingsRecorder()),
 		unschedulableBindings: newUnschedulableBindings(metrics.NewUnschedulableBindingsRecorder()),
 	}
+
+	bq.backoffQ = heap.NewWithRecorder(BindingKeyFunc, bq.lessBackoffCompletedWithPriority, metrics.NewBackoffBindingsRecorder())
 
 	return bq
 }
@@ -158,6 +161,9 @@ type prioritySchedulingQueue struct {
 	// lock takes precedence and should be taken first,
 	// before any other locks in the queue.
 	lock sync.RWMutex
+
+	// clock is the clock used to get the current time.
+	clock clock.WithTicker
 
 	// binding initial backoff duration.
 	bindingInitialBackoffDuration time.Duration
@@ -217,7 +223,7 @@ func (bq *prioritySchedulingQueue) flushBackoffQCompleted() {
 // If this returns true, the binding should not be re-tried.
 func (bq *prioritySchedulingQueue) isBindingBackingoff(bindingInfo *QueuedBindingInfo) bool {
 	boTime := bq.getBackoffTime(bindingInfo)
-	return boTime.After(time.Now())
+	return boTime.After(bq.clock.Now())
 }
 
 // calculateBackoffDuration is a helper function for calculating the backoffDuration
@@ -247,6 +253,18 @@ func (bq *prioritySchedulingQueue) getBackoffTime(bindingInfo *QueuedBindingInfo
 	return backoffTime
 }
 
+// lessBackoffCompletedWithPriority is the function used by the backoffQ heap algorithm to sort bindings.
+// It sorts bindings based on their backoff completion time and priority.
+func (bq *prioritySchedulingQueue) lessBackoffCompletedWithPriority(bInfo1, bInfo2 *QueuedBindingInfo) bool {
+	bo1 := bq.getBackoffTime(bInfo1)
+	bo2 := bq.getBackoffTime(bInfo2)
+	if !bo1.Equal(bo2) {
+		return bo1.Before(bo2)
+	}
+	// If the backoff time is the same, sort the bindings as activeQ does.
+	return Less(bInfo1, bInfo2)
+}
+
 // flushUnschedulableBindingsLeftover moves bindings which stay in unschedulableBindings
 // longer than bindingMaxInUnschedulableBindingsDuration to activeQ.
 func (bq *prioritySchedulingQueue) flushUnschedulableBindingsLeftover() {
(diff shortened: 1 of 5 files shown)


pkg/scheduler/internal/queue/scheduling_queue.go
```diff
@@ -191,11 +197,19 @@ func (bq *prioritySchedulingQueue) Close() {
...
 
-// flushBackoffQCompleted moves all bindings from backoffQ which have completed backoff in to activeQ
+// flushBackoffQCompleted moves all bindings from backoffQ which have completed backoff to activeQ
 func (bq *prioritySchedulingQueue) flushBackoffQCompleted() {
 	bq.lock.Lock()
 	defer bq.lock.Unlock()
+	bindingsCompletedBackoff := bq.popAllBackoffQCompleted()
+	for _, bInfo := range bindingsCompletedBackoff {
+		bq.moveToActiveQ(bInfo)
+	}
```
whitewindmills, line 219: I don't get why you would specifically extract a slice(named `bindingsCompletedBackoff`) to store dequeued backoff-bindings, then move to activeQ

pkg/scheduler/internal/queue/scheduling_queue.go
```diff
@@ -191,11 +197,19 @@ func (bq *prioritySchedulingQueue) Close() {
...
+}
 
+// popAllBackoffQCompleted pops all bindings from backoffQ which have completed backoff.
+func (bq *prioritySchedulingQueue) popAllBackoffQCompleted() []*QueuedBindingInfo {
+	poppedBindingInfos := make([]*QueuedBindingInfo, 0, bq.backoffQ.Len()/2)
```
whitewindmills: what is the special meaning of taking half the length?

pkg/scheduler/internal/queue/scheduling_queue.go
rayo1uo: Thanks for reviewing! I refactored the `flushBackoffQCompleted` method by extracting a dedicated `popAllBackoffQCompleted` function, which facilitates the writing of unit tests and improves code testability. See `TestBackoffQueueOrdering` in `scheduling_queue_test.go`~

pkg/scheduler/internal/queue/scheduling_queue.go
rayo1uo: I just pre-allocate the slice with a capacity of half the queue length as a heuristic, assuming that typically only a subset of bindings in the backoff queue will be ready to move at any given time, thus reducing memory reallocations without over-allocating. If it's not necessary, I'll remove it.

pkg/scheduler/internal/queue/scheduling_queue.go
rayo1uo: > I don't get why you would specifically extract a slice(named `bindingsCompletedBackoff`) to store dequeued backoff-bindings, then move to activeQ

@whitewindmills Thanks for reviewing ❤️ ! I refactored the `flushBackoffQCompleted` method by extracting a dedicated `popAllBackoffQCompleted` function, which facilitates the writing of unit tests and improves code testability. See `TestBackoffQueueOrdering` function in `scheduling_queue_test.go`~

pkg/scheduler/internal/queue/scheduling_queue.go
whitewindmills: hhh, no big deal! but the number of dequeued bindings is usually unpredictable, so there is no need to deliberately control it‘s initial capacity.

pkg/scheduler/internal/queue/scheduling_queue.go
whitewindmills: if it's just for testing convenience, just return `bindingsCompletedBackoff` to unit tests, keep the orignal codes.

rayo1uo: > if it's just for testing convenience, just return `bindingsCompletedBackoff` to unit tests, keep the orignal codes.

Thank you. The revisions have been made.

pkg/scheduler/internal/queue/scheduling_queue.go
rayo1uo: > hhh, no big deal! but the number of dequeued bindings is usually unpredictable, so there is no need to deliberately control it‘s initial capacity.

Got! already remove it 👍

pkg/scheduler/internal/queue/scheduling_queue.go
```diff
@@ -213,11 +219,35 @@ func (bq *prioritySchedulingQueue) flushBackoffQCompleted() {
 	}
 }
 
+// popAllBackoffQCompleted pops all bindings from backoffQ which have completed backoff.
+func (bq *prioritySchedulingQueue) popAllBackoffQCompleted() []*QueuedBindingInfo {
```
whitewindmills: @rayo1uo could you add some comments explaining that this is for testing purposes? I tend to rename the func name and put it in the UT file and .

rayo1uo: @whitewindmills you mean put the function in ut file, right?

rayo1uo: @whitewindmills done for it, plz take an another look.

whitewindmills:
@rayo1uo please update the release note.
```
karmada-scheduler: Fixes the bug in the backoff queue where the sorting function was incorrect, potentially causing high-priority items with long backoffs to block lower-priority items.
```
/lgtm
/approve

PR #6997, opened 2025-12-04
refactor: unify maxAvailableComponentsSets calculation

zhzhuang-zju:
**What type of PR is this?**
/kind cleanup

Ref to https://github.com/karmada-io/karmada/pull/6912#issuecomment-3531756545
Given the high similarity in both algorithm and objective between the `noderesource` plugin and the resource model when computing `maxAvailableComponentsSets`, we can unify their computation logic.

Fixes #

```release-note

```

### DIFF
--- a/pkg/estimator/client/general.go
+++ b/pkg/estimator/client/general.go
@@ -19,7 +19,6 @@ package client
 import (
 	"context"
 	"fmt"
-	"maps"
 	"math"
 	"sort"
 
@@ -29,8 +28,17 @@ import (
 
 	clusterv1alpha1 "github.com/karmada-io/karmada/pkg/apis/cluster/v1alpha1"
 	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
+	"github.com/karmada-io/karmada/pkg/estimator"
+	"github.com/karmada-io/karmada/pkg/estimator/pb"
 	"github.com/karmada-io/karmada/pkg/features"
+	"github.com/karmada-io/karmada/pkg/util"
+	schedulerframework "github.com/karmada-io/karmada/pkg/util/lifted/scheduler/framework"
 )
+
+// maxPodsCountPerNode defines the maximum pods count per node.
+// This is the default and recommended value set by Kubernetes for nodes.
+// More details can be found at: https://kubernetes.io/docs/setup/best-practices/cluster-large/
+const maxPodsCountPerNode = 110
 
 // GeneralEstimator is the default replica estimator.
 func init() {
@@ -127,9 +135,9 @@ func (ge *GeneralEstimator) maxAvailableComponentSets(cluster *clusterv1alpha1.C
 		return int32(allowedPods) // #nosec G115: integer overflow conversion int64 -> int32
 	}
 
-	podBound := allowedPods / podsPerSet
+	podBound := int32(allowedPods / podsPerSet) // #nosec G115: integer overflow conversion int64 -> int32
 	if len(perSet) == 0 || allZero(perSet) {
-		return int32(podBound) // #nosec G115: integer overflow conversion int64 -> int32
+		return podBound
 	}
 
 	// Find limiting resource requirement, which will bound maxSet calculation
@@ -144,116 +152,48 @@ func (ge *GeneralEstimator) maxAvailableComponentSets(cluster *clusterv1alpha1.C
 			return 0 // no capacity for this resource
 		}
 
-		resBound := resAvail / req
+		resBound := int32(resAvail / req) // #nosec G115: integer overflow conversion int64 -> int32
 		if resBound < maxSets {
 			maxSets = resBound
 		}
 	}
 
 	if features.FeatureGate.Enabled(features.CustomizedClusterResourceModeling) && len(cluster.Status.ResourceSummary.AllocatableModelings) > 0 {
-		if num, err := getMaximumSetsBasedOnResourceModels(cluster, components, podBound); err != nil {
+		if num, err := getMaximumSetsBasedOnResourceModels(cluster, components, maxSets); err != nil {
 			klog.Warningf("Failed to get maximum sets based on resource models, skipping: %v", err)
 		} else if num < maxSets {
 			maxSets = num
 		}
 	}
 
-	return int32(maxSets) // #nosec G115: integer overflow conversion int64 -> int32
+	return maxSets
 }
 
 // getMaximumSetsBasedOnResourceModels computes the maximum number of full sets that can be
 // placed on a cluster using the cluster's ResourceModels. It expands one set into
 // replica kinds (demand + count) and performs a first-fit-decreasing placement onto model-grade nodes.
 // `upperBound` caps the search. We can set this using the podBound (allowedPods / podsPerSet)
-func getMaximumSetsBasedOnResourceModels(
-	cluster *clusterv1alpha1.Cluster,
-	components []workv1alpha2.Component,
-	upperBound int64,
-) (int64, error) {
-	if upperBound <= 0 {
-		return 0, nil
-	}
-
-	// Compressed one-set: per-kind (identical replicas grouped)
-	oneSetKinds := expandKindsOneSet(components)
-	if len(oneSetKinds) == 0 {
-		// If there are no pods to schedule, just return upperBound
-		return upperBound, nil
-	}
-
-	// Use cluster "available" totals (allocatable - allocated - allocating) for normalized scoring
-	// This reflects what the cluster can actually accept now
-	totals := availableResourceMap(cluster.Status.ResourceSummary)
-
-	for i := range oneSetKinds {
-		oneSetKinds[i].score = demandScoreNormalized(oneSetKinds[i].dem, totals)
-	}
-	sort.Slice(oneSetKinds, func(i, j int) bool {
-		if oneSetKinds[i].score == oneSetKinds[j].score {
-			return demandSum(oneSetKinds[i].dem) > demandSum(oneSetKinds[j].dem)
-		}
-		return oneSetKinds[i].score > oneSetKinds[j].score
-	})
-
-	//Build model nodes from Spec.ResourceModels and Status.AllocatableModelings
+func getMaximumSetsBasedOnResourceModels(cluster *clusterv1alpha1.Cluster, components []workv1alpha2.Component, upperBound int32) (int32, error) {
 	nodes, err := buildModelNodes(cluster)
 	if err != nil {
 		return -1, err
 	}
-	if len(nodes) == 0 {
-		return 0, nil
+
+	pbComponents := make([]pb.Component, 0, len(components))
+	for _, comp := range components {
+		pbComponents = append(pbComponents, pb.Component{
+			Name:                comp.Name,
+			Replicas:            comp.Replicas,
+			ReplicaRequirements: toPBReplicaRequirements(comp.ReplicaRequirements),
+		})
 	}
 
-	var sets int64
-	for sets < upperBound {
-		if !placeOneSet(oneSetKinds, nodes) {
-			break
-		}
-		sets++
-	}
-	return sets, nil
-}
-
-// placeOneSet attempts to place exactly ONE full set (all kinds with their per-set replica counts)
-// onto the provided working node capacities (in-place)
-// Returns true if successful
-func placeOneSet(orderedKinds []replicaKind, work []modelNode) bool {
-	for _, k := range orderedKinds {
-		remaining := k.count
-		if remaining <= 0 {
-			continue
-		}
-		// first-fit across nodes
-		for n := range work {
-			if remaining <= 0 {
-				break
-			}
-			fit := maxFit(work[n].cap, k.dem)
-			if fit <= 0 {
-				continue
-			}
-			place := fit
-			if place > remaining {
-				place = remaining
-			}
-			consumeMul(work[n].cap, k.dem, place)
-			remaining -= place
-		}
-		if remaining > 0 {
-			return false
-		}
-	}
-	return true
-}
-
-// modelNode holds remaining capacity for a given node across all resource types
-type modelNode struct {
-	cap map[corev1.ResourceName]int64
+	return estimator.NewSchedulingSimulator(nodes).SimulateScheduling(pbComponents, upperBound), nil
 }
 
 // buildModelNodes constructs identical nodes for each model grade using its Min vector,
 // repeated AllocatableModelings[grade].Count times. Grades are indexed directly.
-func buildModelNodes(cluster *clusterv1alpha1.Cluster) ([]modelNode, error) {
+func buildModelNodes(cluster *clusterv1alpha1.Cluster) ([]*schedulerframework.NodeInfo, error) {
 	if cluster == nil {
 		return nil, fmt.Errorf("nil cluster")
 	}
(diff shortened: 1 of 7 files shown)


seanlaii:
/assign
Thank you so much!

pkg/estimator/client/general.go
```diff
@@ -144,116 +147,58 @@ func (ge *GeneralEstimator) maxAvailableComponentSets(cluster *clusterv1alpha1.C
...
-	var sets int64
-	for sets < upperBound {
-		if !placeOneSet(oneSetKinds, nodes) {
-			break
+	pbComponents := make([]pb.Component, 0, len(components))
+	for _, comp := range components {
+		// Deep-copy so that pointer is not shared between goroutines
+		var cr *workv1alpha2.ComponentReplicaRequirements
+		if comp.ReplicaRequirements != nil {
+			cr = comp.ReplicaRequirements.DeepCopy()
```
seanlaii: Do we need to use DeepCopy here?

pkg/estimator/scheduling_simulator_components.go
```diff
@@ -0,0 +1,175 @@
...
+		}
+
+		for node.Allocatable.Allocatable(task.requiredResourcePerReplica) {
+			// Assign one replica to this node.
+			task.scheduleOnePod(node)
+			if task.done() {
+				// short path
+				return true
+			}
+		}
```
seanlaii: Minor Improvement: Maybe we can use `divide` to quickly calculate the available number of replica and minus the number instead of using while loop.

seanlaii (commented):
Thanks! The structure looks great. Overall LGTM, just left two minor comments.

pkg/estimator/client/general.go
zhzhuang-zju: That's a good question! Initially, I referred to the code snippet below, but it now seems unnecessary. 
https://github.com/karmada-io/karmada/blob/699a4b6f4e068d547394a667d831bb925e84802f/pkg/estimator/client/accurate.go#L104-L116

@mszacillo Can you help check if DeepCopy is needed here? I may have missed something.

pkg/estimator/scheduling_simulator_components.go
zhzhuang-zju: Good point! I can give it a try

zhzhuang-zju:
@seanlaii, please take a look again, thanks :)

pkg/estimator/client/general.go
zhzhuang-zju: Local testing didn't reveal any issues, so I think we can remove the deepcopy here for now. If @mszacillo has any concerns, we can continue the discussion.

seanlaii (approved):
/LGTM
Thanks!

RainbowMango (commented):
/assign

pkg/estimator/client/general.go
```diff
@@ -144,116 +147,52 @@ func (ge *GeneralEstimator) maxAvailableComponentSets(cluster *clusterv1alpha1.C
...
-	sort.Slice(oneSetKinds, func(i, j int) bool {
-		if oneSetKinds[i].score == oneSetKinds[j].score {
-			return demandSum(oneSetKinds[i].dem) > demandSum(oneSetKinds[j].dem)
-		}
-		return oneSetKinds[i].score > oneSetKinds[j].score
-	})
-
-	//Build model nodes from Spec.ResourceModels and Status.AllocatableModelings
-	nodes, err := buildModelNodes(cluster)
+func getMaximumSetsBasedOnResourceModels(cluster *clusterv1alpha1.Cluster, components []workv1alpha2.Component, upperSets int32) (int32, error) {
```
```suggestion
func getMaximumSetsBasedOnResourceModels(cluster *clusterv1alpha1.Cluster, components []workv1alpha2.Component, upperBound int32) (int32, error) {
```
I like `upperBound` better, because it is more straightforward.

pkg/estimator/client/general.go
```diff
@@ -144,116 +147,52 @@ func (ge *GeneralEstimator) maxAvailableComponentSets(cluster *clusterv1alpha1.C
...
+	nodes, err := getNodesAvailableResources(cluster)
```
RainbowMango: I like the original function name `buildModelNodes` as we don't just `get` nodes from the cluster object.

pkg/estimator/scheduling_simulator_components.go
```diff
@@ -0,0 +1,171 @@
...
+type SchedulingSimulator struct {
+	// nodes represent the available cluster nodes with their resource capacity
+	nodes []*schedulerframework.NodeInfo
+	// matchNode is a function to check if a component can be scheduled on a specific node
+	// based on node affinity, tolerations, and other scheduling constraints
+	matchNode func(nodeClaim *pb.NodeClaim, node *schedulerframework.NodeInfo) bool
+}
+
+// NewSchedulingSimulator creates a new scheduling simulator instance.
+func NewSchedulingSimulator(nodes []*schedulerframework.NodeInfo, matchNode func(nodeClaim *pb.NodeClaim, node *schedulerframework.NodeInfo) bool) *SchedulingSimulator {
```
RainbowMango: I've a question on the `matchNode` design.

Does the `SchedulingSimulator` really need the upper layer to decide how to match nodes?
I mean, it knows best how to match nodes and whether there is enough information to perform the match operation.

pkg/estimator/scheduling_simulator_components.go
```diff
@@ -0,0 +1,171 @@
...
+	sort.Slice(componentsCopy, func(i, j int) bool {
+		iMaxSets := totalResource.MaxDivided(componentsCopy[i].ReplicaRequirements.ResourceRequest)
+		jMaxSets := totalResource.MaxDivided(componentsCopy[j].ReplicaRequirements.ResourceRequest)
+		if iMaxSets == jMaxSets {
+			// Use tie-breaker when components have similar scheduling difficulty
+			return tieBreaker(componentsCopy[i], componentsCopy[j])
+		}
+		// Schedule components with fewer possible placements first (harder to place)
+		return iMaxSets < jMaxSets
+	})
```
RainbowMango: I'm not sure if it's worth sorting the components here, as to do so, we have to iterate over the nodes once.

I will take another look.

pkg/estimator/scheduling_simulator_components.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+// SimulateSchedulingFFD implements the First Fit Decreasing (FFD) algorithm to estimate
+// the maximum number of complete component sets that can be scheduled on the cluster.
+//
+// FFD Algorithm Steps:
+// 1. Sort components by their "difficulty to schedule" (decreasing order of resource constraints)
+// 2. For each complete set, try to schedule all components using first-fit strategy
+// 3. Continue until no more complete sets can be scheduled or upper limit is reached
+func (s *SchedulingSimulator) SimulateSchedulingFFD(components []pb.Component, upperSets int32) int32 {
```
RainbowMango: `upperSets ` ditto.

pkg/estimator/scheduling_simulator_components.go
```diff
@@ -0,0 +1,171 @@
...
+		if !s.scheduleComponent(component) {
+			return false
+		}
+	}
+
+	return true
+}
+
+func (s *SchedulingSimulator) scheduleComponent(component pb.Component) bool {
+	task := newSchedulingTask(component)
```
RainbowMango: I need to take another look to see if we can get rid of the scheduling task. As it seems a little bit complicated and builds an object in each loop (gc pressure).

pkg/estimator/scheduling_simulator_components.go
zhzhuang-zju: The `*schedulerframework.NodeInfo.node` obtained from the ResourceModel is nil, meaning there isn't sufficient information to determine the `matchNode` operation. In this case, skipping the match operation is also a viable option.

pkg/estimator/scheduling_simulator_components.go
zhzhuang-zju: > I will take another look.

Sure, below are the previous discussions, hope to help inform your decision.

https://github.com/karmada-io/karmada/pull/6912#issuecomment-3544974932
https://github.com/karmada-io/karmada/pull/6912#issuecomment-3546163585
https://github.com/karmada-io/karmada/pull/6912#discussion_r2554488343

pkg/estimator/scheduling_simulator_components.go
RainbowMango: Yes, that is exactly what I mean. That can be handled inside the `SchedulingSimulator`.

pkg/estimator/scheduling_simulator_components.go
RainbowMango: From the discussion so far, it seems you're more tend to use FF rather than FFD, right?

What I mean is that we should temporarily postpone implementing FFD because:
- There's no evidence demonstrating its effectiveness, and it introduces noticeable complexity.
- Even though our calculation here is precise, the order in which Pods are created in member clusters—and their scheduling order—is not guaranteed, so the effort we put in may not yield the expected results.

That said, it's still important to ensure a deterministic order for components—so that, given the same list of nodes, repeated scheduling yields consistent results.

However, I don't think we need to explicitly sort them; we can simply follow the order already defined in the ResourceBinding.

pkg/estimator/scheduling_simulator_components.go
RainbowMango: How does this sound? It doesn't reduce the number of object created, but looks more straightforward, I guess. 
```golang
func (s *SchedulingSimulator) scheduleComponent(component pb.Component) bool {
	requiredPerReplica := util.NewResource(component.ReplicaRequirements.ResourceRequest)
	requiredPerReplica.AllowedPodNumber = 1
	remaining := component.Replicas
	nodeClaim := component.ReplicaRequirements.NodeClaim

	for _, node := range s.nodes {
		if !s.matchNode(nodeClaim, node) {
			continue
		}

		allocatable := node.Allocatable.MaxDivided(requiredPerReplica.ResourceList())
		if allocatable == 0 {
			continue
		}

		if int64(remaining) < allocatable {
			allocatable = int64(remaining)
		}

		node.Allocatable.SubResource(requiredPerReplica.Clone().Multiply(allocatable))
		remaining -= int32(allocatable) // #nosec G115: xxx
		if remaining == 0 {
			return true
		}
	}

	return remaining == 0
}
```

In addition, the `requiredPerReplica.AllowedPodNumber = 1` not right, with this `node.Allocatable.MaxDivided` will return only `0` or `1`, that not what we expected.

pkg/estimator/scheduling_simulator_components.go
zhzhuang-zju: > From the discussion so far, it seems you're more tend to use FF rather than FFD, right?

Yes, primarily due to computational complexity and a lack of clear understanding of its necessity. As you mentioned, even if we perform highly accurate calculations on the control plane, we still cannot guarantee the actual scheduling order in member clusters.

> However, I don't think we need to explicitly sort them; we can simply follow the order already defined in the ResourceBinding.

Makes sense, will update soon~

pkg/estimator/scheduling_simulator_components.go
zhzhuang-zju: > How does this sound? It doesn't reduce the number of object created, but looks more straightforward, I guess.

Thanks, looks good!

> In addition, the requiredPerReplica.AllowedPodNumber = 1 not right, with this node.Allocatable.MaxDivided will return only 0 or 1, that not what we expected.

Not exactly. `requiredPerReplica.AllowedPodNumber = 1` means that each scheduling request consumes one pod slot that the environment can accommodate. It is `node.Allocatable.AllowedPodNumber` that represents the total number of pods the node can host.

pkg/estimator/scheduling_simulator_components.go
```diff
@@ -0,0 +1,118 @@
...
+// scheduleComponentSet attempts to schedule one complete set of all components.
+func (s *SchedulingSimulator) scheduleComponentSet(components []pb.Component) bool {
+	for _, component := range components {
+		if !s.scheduleComponent(component) {
+			return false
+		}
+	}
+
+	return true
+}
```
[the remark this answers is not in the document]
zhzhuang-zju: @gemini-code-assist This is a simulated scheduling; it won't retry upon failure, and the nodes used in each simulation are always up-to-date, so there won't be any issues.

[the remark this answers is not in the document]
zhzhuang-zju: @gemini-code-assist Still think it's unnecessary. I can clarify in the comments that simulated scheduling modifies node resources.

pkg/estimator/scheduling_simulator_components.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+}
+
+// SimulateSchedulingFF implements the First Fit(FF) algorithm to estimate
+// the maximum number of complete component sets that can be scheduled on the cluster.
+//
+// FF Algorithm Steps:
+// 1. For each complete set, try to schedule all components using first-fit strategy
+// 2. Continue until no more complete sets can be scheduled or upper limit is reached
+func (s *SchedulingSimulator) SimulateSchedulingFF(components []pb.Component, upperBound int32) int32 {
```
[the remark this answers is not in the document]
zhzhuang-zju: @gemini-code-assist ref to https://github.com/karmada-io/karmada/pull/6997#discussion_r2643107949, we decided use FF instead of FFD

zhzhuang-zju:
Thanks for your valuable comments @seanlaii @RainbowMango 
Please take a look.

pkg/estimator/client/general.go
```diff
@@ -144,116 +147,48 @@ func (ge *GeneralEstimator) maxAvailableComponentSets(cluster *clusterv1alpha1.C
...
-		resBound := resAvail / req
+		resBound := int32(resAvail / req) // #nosec G115: integer overflow conversion int64 -> int32
 		if resBound < maxSets {
 			maxSets = resBound
 		}
 	}
 
 	if features.FeatureGate.Enabled(features.CustomizedClusterResourceModeling) && len(cluster.Status.ResourceSummary.AllocatableModelings) > 0 {
-		if num, err := getMaximumSetsBasedOnResourceModels(cluster, components, podBound); err != nil {
+		if num, err := getMaximumSetsBasedOnResourceModels(cluster, components, maxSets); err != nil {
```
RainbowMango, line 162: PS: I just realized, using `maxSets` is better than `podBound`. Thanks.

pkg/estimator/client/general.go
```diff
@@ -267,12 +202,14 @@ func buildModelNodes(cluster *clusterv1alpha1.Cluster) ([]modelNode, error) {
...
-	capsByGrade := make(map[uint]map[corev1.ResourceName]int64, len(spec))
+	capsByGrade := make(map[uint]corev1.ResourceList, len(spec))
 	for _, m := range spec {
-		tmpl := make(map[corev1.ResourceName]int64, len(m.Ranges))
+		tmpl := make(corev1.ResourceList, len(m.Ranges))
 		for _, r := range m.Ranges {
-			tmpl[r.Name] = quantityAsInt64(r.Min)
+			tmpl[r.Name] = r.Min
 		}
+		// Pods resource is not modeled in ResourceModels, so we set it to MaxInt64 to avoid limiting scheduling.
```
RainbowMango: I don't get what the `limiting scheduling` means.

Why is pod number so important here?

pkg/estimator/scheduling_simulator_components.go
```diff
@@ -0,0 +1,120 @@
...
```
RainbowMango:
```suggestion
func (s *SchedulingSimulator) SimulateScheduling(components []pb.Component, upperBound int32) int32 {
```
I don't think it's worth taking the `FF` as part of the method name.
1. It's not an algorithm that can be understood just by the abbreviation `FF`. 
2. No plan to implement other algorithms, so no need to reserve extensibility.

pkg/estimator/client/general.go
zhzhuang-zju:
```go
kind: Node
spec:
  podCIDR: 10.10.0.0/24
  podCIDRs:
  - 10.10.0.0/24
  providerID: kind://docker/member1/member1-control-plane
status:
  allocatable:
    cpu: "4"
    pods: "110"
```
The number of pods a node can accommodate is inherently limited, and this value is one of the metrics considered in simulated scheduling. However, since the resource model does not impose any restriction on the number of pods per node, its `AllowedPodNumber` is set to `maxInt`, meaning pod count should not act as a limiting factor.

zhzhuang-zju: Improved the comments, hoping to help with understanding

pkg/estimator/client/general.go
```diff
@@ -267,12 +202,17 @@ func buildModelNodes(cluster *clusterv1alpha1.Cluster) ([]modelNode, error) {
...
+		tmpl := make(corev1.ResourceList, len(m.Ranges))
 		for _, r := range m.Ranges {
-			tmpl[r.Name] = quantityAsInt64(r.Min)
+			tmpl[r.Name] = r.Min
 		}
+		// The number of pods a node can accommodate is inherently limited, and this value is one of the metrics
+		// considered in simulated scheduling. However, since the resource model does not impose any restriction on the
+		// number of pods per node, its `AllowedPodNumber` is set to `maxInt`, meaning pod count should not act as a
+		// limiting factor.
+		tmpl[corev1.ResourcePods] = *resource.NewQuantity(math.MaxInt64, resource.DecimalSI)
```
```suggestion
		// The number of pods a node can accommodate is a critical metric in scheduling simulation. Each node must have
		// a pod capacity limit; otherwise, the scheduler cannot determine if a node can accommodate any pod. Kubernetes
		// sets a default pod limit of 110 per node, which we adopt here for consistency.
		//
		// Note: This virtual node uses a fixed pod capacity of 110 for each simulation. The actual pod count already
		// running on this virtual node is not considered in the calculation. This means the estimation may be inaccurate
		// and could potentially exceed 110 pods in practice. This is an inherent limitation of the ResourceModel mechanism,
		// as the estimation is approximative rather than precise.
		tmpl[corev1.ResourcePods] = *resource.NewQuantity(110, resource.DecimalSI)
```

zhzhuang-zju: thx~
done

RainbowMango (approved):
/lgtm
/approve

PR #7015, opened 2025-12-13
Add PBD and priority class support for Karmada operator

### LINKED ISSUE #6973: Priority Class and Pod Disruption Budget Support for Karmada Operator Deployments

Add support in the Karmada Operator Helm chart for configuring:

- A **priorityClassName** field for the operator Deployment  
- A **PodDisruptionBudget (PDB)** resource dedicated to the operator

These should be optional features configurable via `values.yaml`, disabled by default for backward compatibility.

Concretely:

1. Allow setting:
   ```yaml
   operator:
     priorityClassName: "system-cluster-critical"
   ```

2. Introduce a PDB template, disabled by default:
   ```yaml
   operator:
     pdb:
       enabled: true
       minAvailable: 1
   ```

---

**Why is this needed**:

1) **The operator is a critical control-plane component**  
   The Karmada Operator manages the lifecycle and configuration of the Karmada control planes. If it becomes unavailable, even momentarily, reconciliation, and component updates all stall. This places the operator in the same category as other mission‑critical system components, which typically include protections like disruption budgets and scheduling priority.

2) **Lack of PriorityClass can delay recovery or break upgrades**  
   During events such as node pressure, rolling OS updates, or autoscaler scale‑downs, the operator may have its pod evicted and consequently fail to reconcile Karmada instances

   Exposing `priorityClassName` ensures the operator is always scheduled ahead of standard workloads, consistent with Kubernetes best practices for core system components.

3) **No PDB exposes the operator to full eviction during maintenance**  
   Without a PodDisruptionBudget, a node drain can evict *all* operator pods simultaneously. This risks interrupting reconciliation loops

   A PDB ensures at least one instance stays available during voluntary disruptions such as node drains or cluster‑autoscaler actions.

4) **Improves resilience and cluster‑wide stability**  
   Environments with multiple nodes, autoscaling, or regular scheduled maintenance rely on predictable disruption handling. Adding PriorityClass + PDB support:
   - Reduces operational risk
   - Prevents subtle downtime
   - Aligns behavior with user expectations for high‑availability operator components.

5) **Aligns with best practices from upstream and ecosystem charts**  
   Many Kubernetes system charts (cert‑manager, Kyverno etc.) already provide support for PDBs and PriorityClasses. Adding support brings Karmada Operator to parity with modern operational standards.

---

### 1. Add optional `priorityClassName`
Example values:
```yaml
operator:
  priorityClassName: ""
```
Rendered into the Deployment only when non‑empty.

### 2. Add optional PodDisruptionBudget
Example values:
```yaml
operator:
  pdb:
    enabled: true
    minAvailable: 1
```
Rendered as a dedicated PDB manifest when enabled.

### 3. Documentation updates
- Document recommended PriorityClasses (e.g., system‑cluster‑critical).  
- Explain safe behavior when replicas = 1 (use `minAvailable: 1`).  
- Add example values and guidance for cluster operators.

---

## 🚀 Summary

Supporting PriorityClass and PDB in the Karmada Operator Helm chart:

- Improves scheduling guarantees during node pressure  
- Ensures at least one operator instance is always available  
- Aligns the chart with Kubernetes ecosystem best practices  
- Reduces operational risk and unexpected downtime  

These enhancements meaningfully improve Karmada’s resilience and bring the operator’s deployment model in line with expectations for critical cluster infrastructure.

jabellard:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**: Please see this issue: https://github.com/karmada-io/karmada/issues/6973

Fixes #6973

```release-note
karmada-operator Helm chart: Added support for configuring priority class and pod disruption budget config for Karmada operator deployment
```

### DIFF
--- a/charts/karmada-operator/templates/karmada-operator-deployment.yaml
+++ b/charts/karmada-operator/templates/karmada-operator-deployment.yaml
@@ -32,6 +32,9 @@ spec:
         {{- end }}
     spec:
       {{- include "karmada.operator.imagePullSecrets" . | indent 6 }}
+      {{- if .Values.operator.priorityClassName }}
+      priorityClassName: {{ .Values.operator.priorityClassName }}
+      {{- end }}
       containers:
       - name: {{ include "karmada.operator.fullname" . }}
         image: {{ template "karmada.operator.image" . }}
--- a/charts/karmada-operator/templates/karmada-operator-poddisruptionbudget.yaml
+++ b/charts/karmada-operator/templates/karmada-operator-poddisruptionbudget.yaml
@@ -1,0 +1,26 @@
+{{- if .Values.operator.pdb.enabled }}
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  name: {{ include "karmada.operator.fullname" . }}
+  namespace: {{ .Release.Namespace }}
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
+    app: {{ include "karmada.operator.fullname" . }}
+    {{- if .Values.commonLabels }}
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" $ ) | nindent 4 }}
+    {{- end }}
+  {{- if .Values.commonAnnotations }}
+  annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+  {{- end }}
+spec:
+  selector:
+    matchLabels: {{- include "common.labels.matchLabels" . | nindent 6 }}
+      app: {{ include "karmada.operator.fullname" . }}
+  {{- if .Values.operator.pdb.minAvailable | toString }}
+  minAvailable: {{ .Values.operator.pdb.minAvailable }}
+  {{- else if .Values.operator.pdb.maxUnavailable | toString }}
+  maxUnavailable: {{ .Values.operator.pdb.maxUnavailable }}
+  {{- else }}
+  {{- fail "pdb.enabled is true but neither pdb.minAvailable nor pdb.maxUnavailable is set" }}
+  {{- end }}
+{{- end }}
--- a/charts/karmada-operator/values.yaml
+++ b/charts/karmada-operator/values.yaml
@@ -111,3 +111,27 @@ operator:
     rollingUpdate:
       maxUnavailable: 0
       maxSurge: 50%
+
+  ## @param operator.priorityClassName Priority class name for the operator pods
+  ## This ensures the operator is scheduled ahead of standard workloads
+  ## Example: "system-cluster-critical"
+  ## When empty, no priorityClassName will be set
+  priorityClassName: ""
+
+  ## @param operator.pdb PodDisruptionBudget configuration for the operator
+  ## Ensures high availability during voluntary disruptions (node drains, cluster-autoscaler actions, etc.)
+  pdb:
+    ## @param operator.pdb.enabled Enable PodDisruptionBudget for the operator
+    enabled: false
+    ## @param operator.pdb.minAvailable Minimum number of pods that must be available
+    ## Can be an integer (e.g., 1) or a percentage (e.g., "50%")
+    ## Mutually exclusive with maxUnavailable
+    ## When both minAvailable and maxUnavailable are set, only minAvailable will be used
+    ## To have maxUnavailable be the value that is used instead, set minAvailable to an empty string
+    minAvailable: 1
+    ## @param operator.pdb.maxUnavailable Maximum number of pods that can be unavailable
+    ## Can be an integer (e.g., 1) or a percentage (e.g., "50%")
+    ## Mutually exclusive with minAvailable
+    ## When both minAvailable and maxUnavailable are set, only minAvailable will be used
+    ## To have maxUnavailable be the value that is used instead, set minAvailable to an empty string
+    maxUnavailable: ""


charts/karmada-operator/templates/karmada-operator-poddisruptionbudget.yaml
```diff
@@ -0,0 +1,24 @@
...
+  {{- end }}
+spec:
+  selector:
+    matchLabels: {{- include "common.labels.matchLabels" . | nindent 6 }}
+      app: {{ include "karmada.operator.fullname" . }}
+  {{- if .Values.operator.pdb.minAvailable }}
+  minAvailable: {{ .Values.operator.pdb.minAvailable }}
+  {{- else if .Values.operator.pdb.maxUnavailable }}
+  maxUnavailable: {{ .Values.operator.pdb.maxUnavailable }}
+  {{- end }}
```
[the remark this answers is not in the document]
jabellard: Fixed.

charts/karmada-operator/templates/karmada-operator-deployment.yaml
```diff
@@ -32,6 +32,9 @@ spec:
         {{- end }}
     spec:
       {{- include "karmada.operator.imagePullSecrets" . | indent 6 }}
+      {{- if .Values.operator.priorityClassName }}
+      priorityClassName: {{ .Values.operator.priorityClassName }}
```
seanlaii, line 36: nit:
```suggestion
      {{- with .Values.operator.priorityClassName }}
      priorityClassName: {{ . | quote }}
```

charts/karmada-operator/templates/karmada-operator-poddisruptionbudget.yaml
```diff
@@ -0,0 +1,26 @@
...
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  name: {{ include "karmada.operator.fullname" . }}
+  namespace: {{ .Release.Namespace }}
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
+    app: {{ include "karmada.operator.fullname" . }}
+    {{- if .Values.commonLabels }}
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" $ ) | nindent 4 }}
+    {{- end }}
```
seanlaii, line 11: nit:
```suggestion
    {{- with .Values.commonLabels }}
    {{- include "common.tplvalues.render" (dict "value" . "context" $) | nindent 4 }}
    {{- end }}
```

charts/karmada-operator/templates/karmada-operator-poddisruptionbudget.yaml
```diff
@@ -0,0 +1,26 @@
...
+  {{- if .Values.commonAnnotations }}
+  annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+  {{- end }}
```
seanlaii, line 14: nit:
```suggestion
  {{- with .Values.commonAnnotations }}
  annotations: {{- include "common.tplvalues.render" ( dict "value" . "context" $ ) | nindent 4 }}
  {{- end }}
```

charts/karmada-operator/templates/karmada-operator-poddisruptionbudget.yaml
```diff
@@ -0,0 +1,26 @@
+{{- if .Values.operator.pdb.enabled }}
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  name: {{ include "karmada.operator.fullname" . }}
+  namespace: {{ .Release.Namespace }}
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
+    app: {{ include "karmada.operator.fullname" . }}
+    {{- if .Values.commonLabels }}
```
seanlaii, line 9: Maybe we can provide `commonLabels` and `commonAnnotations` in `values.yaml`.

charts/karmada-operator/templates/karmada-operator-poddisruptionbudget.yaml
```diff
@@ -0,0 +1,26 @@
+{{- if .Values.operator.pdb.enabled }}
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  name: {{ include "karmada.operator.fullname" . }}
+  namespace: {{ .Release.Namespace }}
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
```
seanlaii, line 7: Tag standard labels here: https://github.com/bitnami/charts/blob/main/bitnami/common/templates/_labels.tpl for info.

charts/karmada-operator/templates/karmada-operator-poddisruptionbudget.yaml
```diff
@@ -0,0 +1,26 @@
...
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" $ ) | nindent 4 }}
+    {{- end }}
+  {{- if .Values.commonAnnotations }}
+  annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+  {{- end }}
+spec:
+  selector:
+    matchLabels: {{- include "common.labels.matchLabels" . | nindent 6 }}
+      app: {{ include "karmada.operator.fullname" . }}
+  {{- if .Values.operator.pdb.minAvailable | toString }}
```
seanlaii, line 19: I think we need to use `hasKey` since `null` value will be considered as `true`.
Maybe we can refer to [cert-manager](https://github.com/cert-manager/cert-manager/blob/master/deploy/charts/cert-manager/templates/poddisruptionbudget.yaml#L20).

jabellard: @seanlaii , thanks for the ref o the cert-manager approach and for thinking about the edge cases, but don't think we need to worry about this one. This would be a falsy value, so that block would not get evaluated.

charts/karmada-operator/templates/karmada-operator-poddisruptionbudget.yaml
jabellard: I'm just using the same labels used in the deployment to link it to the pbd.

charts/karmada-operator/templates/karmada-operator-poddisruptionbudget.yaml
jabellard: I'm just using the same labels used in the deployment to link it to the pbd.

charts/karmada-operator/templates/karmada-operator-poddisruptionbudget.yaml
seanlaii: Although users can use it, we currently don't provide it in https://github.com/karmada-io/karmada/blob/master/charts/karmada-operator/values.yaml.
We can add something like https://github.com/karmada-io/karmada/blob/master/charts/karmada/values.yaml#L12-L18.

charts/karmada-operator/templates/karmada-operator-poddisruptionbudget.yaml
seanlaii: Considering this case
```
minAvailable: null
maxAvailable: 1
```
the result will be
```
selector:
  matchLabels:
    minAvailable: 
```

jabellard: Really? Let me try it.

charts/karmada-operator/templates/karmada-operator-deployment.yaml
jabellard: You feel strongly about this? What's the benefit of changing it?

seanlaii: No, it is just more idiomatic. It's fine if you don't want to change it.
Normally, `if` is for condition check, and `with` is to check if the value is presented.
But I don't have strong opinion. The current behavior is correct.

charts/karmada-operator/templates/karmada-operator-poddisruptionbudget.yaml
jabellard: This expands the spoce of the PR, which is why I'd prefer to not add add it as part of this work. Would like to keep things as minimal as possible.

charts/karmada-operator/templates/karmada-operator-poddisruptionbudget.yaml
jabellard: @seanlaii , thanks for catching this. Will push update.

charts/karmada-operator/templates/karmada-operator-poddisruptionbudget.yaml
seanlaii: Sure, since we also don't have `commonAnnotations` in Karmada helm chart, I don't feel strongly on this. Feel free to ignore.

charts/karmada-operator/templates/karmada-operator-poddisruptionbudget.yaml
jabellard: Just tried. Handling this weird edge case comes at the cost of the code becoming a lot of unreadable and requiring use more advanced stuff. I'm unsure if it's worth the tradeoff. `hasKey` also will not catch this. You have other ideas?

seanlaii: Yeah, will probably need this
```
{{- $pdb := .Values.operator.pdb }}

{{- if and (hasKey $pdb "minAvailable") (ne $pdb.minAvailable null) (ne $pdb.minAvailable "") }}
minAvailable: {{ $pdb.minAvailable }}
{{- else if and (hasKey $pdb "maxUnavailable") (ne $pdb.maxUnavailable null) (ne $pdb.maxUnavailable "") }}
maxUnavailable: {{ $pdb.maxUnavailable }}
{{- end }}
```
I am OK to keep it to maintain the readability.
Could you add some comment in the `values.yaml` to inform the users? Thanks!

jabellard: Will add some instructions on how to have `maxUnavailable` be the field that takes precedence.

jabellard: Done

zhzhuang-zju:
Thanks
/assign

zhzhuang-zju:
@jabellard Thanks, I have no further comments. It would be perfect if you could squash your commits into one.

RainbowMango:
> @jabellard Thanks, I have no further comments. It would be perfect if you could squash your commits into one.

I rebased and squashed this; nothing has changed since the last review.

/lgtm

PR #6868, opened 2025-10-23
Bump Kubernetes dependencies from v1.33.2 to v1.34.1

RainbowMango:
**What type of PR is this?**

/kind cleanup

**What this PR does / why we need it**:

This PR bumps Kubernetes dependencies with minimal changes; https://github.com/karmada-io/karmada/issues/6862 tracks the rest of the tasks.

**Which issue(s) this PR fixes**:
Fixes #
Part of #6862

**Special notes for your reviewer**:
This PR introduces many Kubernetes changes; see #6862 for the full list, and each change is kept in a separate commit.

**Does this PR introduce a user-facing change?**:

```release-note
Kubernetes dependencies have been updated to v1.34.1.
```

PR #7023, opened 2025-12-17
karmadactl: fix the messy auto-completion suggestions

zhzhuang-zju:
**What type of PR is this?**
/kind bug

Recently, while using the auto-completion feature of `karmadactl get`, I noticed that the completion output is messy and unusable. For example, when typing `karmadactl get dep` and pressing Tab, the result appears as:

```bash
$ karmadactl get deployments\ \ \ \ \ \ \ \ \ \ \ \ \ \ \ \ \ \ \ \ \ \ \ \ \ \ \ \ \ \ \ \ deploy\ \ \ \ apps/v1\ \ \ \ \ \ \ \ \ \ \ \ \ \ \ \ \ \ \ \ \ \ \ \ \ \ \ true\ \ \ \ Deployment
```

The root cause lies in the `compGetResourceList` function, which calls `o.RunAPIResources()` to fetch the list of supported API resources from the environment. Normally, only the `name` column should be returned. This behavior is supposed to be controlled by the `o.PrintFlags.OutputFormat` setting within the `o.Complete` method. However, in `compGetResourceList`, the assignment to `o.PrintFlags.OutputFormat` occurs *after* `o.Complete`. As a result, the `OutputFormat` setting takes no effect, causing all columns—not just the `name`—to be returned during completion.

https://github.com/karmada-io/karmada/blob/ed620ac77196106ae497ed4ff211f5e3096dbec8/pkg/karmadactl/util/completion/completion.go#L255-L267

Fixes #

This issue affects not only `karmadactl get` but also other commands that rely on API resource completion, such as `karmadactl apply`.  
A backport to the `release-1.16` branch is required.

```release-note
`karmadactl`: Fixed the messy auto-completion suggestions for commands like 'get' and 'apply'.
```

### DIFF
--- a/pkg/karmadactl/util/completion/completion.go
+++ b/pkg/karmadactl/util/completion/completion.go
@@ -252,15 +252,15 @@ func compGetResourceList(restClientGetter genericclioptions.RESTClientGetter, cm
 	// TODO: Using karmadactlapiresources.CommandAPIResourcesOptions to adapt to the operation scope.
 	o := apiresources.NewAPIResourceOptions(streams)
 
-	if err := o.Complete(restClientGetter, cmd, nil); err != nil {
-		return nil
-	}
-
 	// Get the list of resources
 	o.PrintFlags.OutputFormat = ptr.To("name")
 	o.Cached = true
 	o.Verbs = []string{"get"}
 	// TODO:Should set --request-timeout=5s
+
+	if err := o.Complete(restClientGetter, cmd, nil); err != nil {
+		return nil
+	}
 
 	// Ignore errors as the output may still be valid
 	if err := o.RunAPIResources(); err != nil {


zhzhuang-zju:
@gemini-code-assist Thanks for your suggestions about the release note. Can you help fill it out for the PR description?

RainbowMango:
@zhzhuang-zju Do you know when and how this issue was introduced?

zhzhuang-zju:
> @zhzhuang-zju Do you know when and how this issue was introduced?

Yes, it was actually introduced during a [Kubernetes dependency upgrade](https://github.com/karmada-io/karmada/pull/6868). In this upgrade, the behavior of the `api-resources` command output changed—it was originally used only within `o.RunAPIResources()`, but was moved to be initialized in `o.Complete()`.

From a logical standpoint, parameter assignment should indeed occur before `o.Complete()`.

RainbowMango (approved):
Thanks for the clarification. 
Indeed, it is hard to catch changes like this during bump Kubernetes dependencies. 
/approve

PR #7038, opened 2025-12-22
skip cleanup rt and binding when the server could not find the requested resource

zhzhuang-zju:
**What type of PR is this?**
/kind bug

When deleting a PropagationPolicy or ClusterPropagationPolicy, the cleanup logic attempts to find all resources matched by the policy's resource selectors. If a selector targets a resource that doesn't exist (for example, a cluster-scoped resource with a namespace specified), the API server returns a "Not Found" error. Previously, this error would be propagated up, potentially blocking the policy from being finalized and deleted.

This PR fixes this by specifically handling `apierrors.IsNotFound` errors during the cleanup process. Now, when a resource is not found, an informational message is logged, and the process continues to the next resource, ensuring that policy deletion is not blocked by selectors pointing to non-existent resources.

Additionally, error logging has been added to the policy deletion handlers to improve traceability when failures occur.

Fixes #

```release-note
`karmada-controller-manager`: Fixed an issue where policy deletion could be blocked if a resource selector targeted a non-existent resource.
```

### DIFF
--- a/pkg/detector/detector.go
+++ b/pkg/detector/detector.go
@@ -1014,6 +1014,7 @@ func (d *ResourceDetector) ReconcilePropagationPolicy(key util.QueueKey) error {
 		policyID := propagationObject.Labels[policyv1alpha1.PropagationPolicyPermanentIDLabel]
 		claimMetadata := labels.Set{policyv1alpha1.PropagationPolicyPermanentIDLabel: policyID}
 		if err = d.handlePolicyDeletion(claimMetadata, propagationObject.Spec.ResourceSelectors, CleanupPPClaimMetadata); err != nil {
+			klog.Errorf("Failed to handle policy deletion for PropagationPolicy(%s), err: %v", nkey.NamespaceKey(), err)
 			return err
 		}
 		if controllerutil.RemoveFinalizer(propagationObject, util.PropagationPolicyControllerFinalizer) {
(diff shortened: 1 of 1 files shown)


zhzhuang-zju:
@RainbowMango @XiShanYongYe-Chang How does this PR description look? Is it clear and complete?

pkg/detector/detector.go
```diff
@@ -1123,6 +1125,10 @@ func (d *ResourceDetector) handlePolicyDeletion(claimMetadata labels.Set, resour
 			klog.Infof("Skip cleanup as API(%s, kind=%s) is not installed or has been removed", objRef.APIVersion, objRef.Kind)
 			continue
 		}
+		if apierrors.IsNotFound(err) {
+			klog.Infof("Skip cleanup as the server could not find the requested resource(%s, kind=%s, namespace=%s). This may be because a namespace was incorrectly specified for a cluster-scoped resource.", objRef.APIVersion, objRef.Kind, objRef.Namespace)
```
RainbowMango:
```suggestion
			klog.Infof("Skip cleanup as the server could not find the requested resource(%s, kind=%s, namespace=%s).", objRef.APIVersion, objRef.Kind, objRef.Namespace)
```

> This may be because a namespace was incorrectly specified for a cluster-scoped resource.

Is this the only reason for `not found`?

zhzhuang-zju: Yes, I haven't found any other causes for the "not found" error.

RainbowMango: The log is more like a comment that explains why we need to handle the error.
I don't think we need to explain the reason in logs, as this is a normal log, people don't have to pay attention to it.

zhzhuang-zju: Fixed, thx

PR #7044, opened 2025-12-25
docs(scheduler): scheduler-extensibility

CharlesQQ:
**What type of PR is this?**

Fixes #

```release-note

```

### DIFF
--- a/docs/proposals/scheduler-extensibility/README.md
+++ b/docs/proposals/scheduler-extensibility/README.md
@@ -1,0 +1,1474 @@
+---
+title: Enhancing Karmada Scheduler Extensibility to Support Customized Requirements
+authors:
+- "@charesQQ"
+reviewers:
+- TBD
+approvers:
+- TBD
+
+creation-date: 2025-12-25
+
+---
+
+# Enhancing Karmada Scheduler Extensibility to Support Customized Requirements
+
+## Summary
+
+While the current Karmada scheduler provides powerful multi-cluster scheduling capabilities, there are some limitations in meeting customized scheduling requirements when adopting Karmada in enterprise environments. This proposal aims to enhance the extensibility architecture of the Karmada scheduler to better support:
+
+1. **API Flexibility**: For example, specifying replica counts by IDC/region instead of by weight
+2. **AssignReplicas Requirements**: For example, average replica allocation with change minimization under specified IDC/region constraints, rather than score-based allocation
+3. **Advanced Scheduling Constraints**: For example, scheduling decisions that depend on service attributes and other profiling data
+
+This proposal presents an extensible scheduling framework enhancement that allows users to extend scheduler capabilities through a standardized plugin mechanism without modifying the core scheduler code. This design maintains the generality of the Karmada scheduler while meeting the customization needs of enterprise users.
+
+## Motivation
+
+### Background
+
+### Goals
+
+The goal of this proposal is to provide a **generic, extensible scheduling framework enhancement** that enables:
+
+1. **Plugin-based Extension**: Users can extend scheduler capabilities by implementing standard plugin interfaces without modifying core code
+2. **Backward Compatibility**: Existing scheduling configurations and behaviors remain unchanged, with new features enabled through optional mechanisms
+
+### Non-Goals
+
+The following items are out of scope for this proposal:
+
+1. **Scheduler Performance Optimization**: While the plugin mechanism considers performance, large-scale performance optimization is not the primary goal
+2. **Scheduling Algorithm Improvements**: This proposal focuses on extensibility architecture rather than specific scheduling algorithm improvements
+
+This proposal presents a **solution to add extension points and flexible API fields**. The core idea is to enrich APIs and add new extension points while maintaining the existing scheduling flow.
+
+### User Stories
+
+Based on the 4 types of strategies actually supported, we provide the following user stories:
+
+#### Story 1: Automatic Scheduling by Specified IDCs (Based on Strategy 1: Idcs)
+
+**Background**: I manage a microservice application that needs to be deployed in East China and North China IDCs, but I'm uncertain about the exact replica distribution between IDCs.
+
+**Requirements**:
+- Specify target IDCs: East China (idc-east) and North China (idc-north)
+- Let the scheduler automatically distribute replicas based on cluster resource utilization
+- Prioritize clusters with sufficient resources to achieve load balancing
+
+**Current Implementation**:
+
+```yaml
+apiVersion: policy.karmada.io/v1alpha1
+kind: PropagationPolicy
+metadata:
+  name: microservice-policy
+  annotations:
+    scheduler.karmada.io/replica-scheduling-strategy: |
+      {
+        "idcs": [
+          {"name": "idc-east"},
+          {"name": "idc-north"}
+        ]
+      }
+spec:
+  resourceSelectors:
+    - apiVersion: apps/v1
+      kind: Deployment
+      name: microservice-app
+  placement:
+    clusterAffinity:
+      labelSelector:
+        matchLabels:
+          env: production
+    replicaScheduling:
+      replicaSchedulingType: Divided
+  replicas: 30
+```
+
+**Scheduling Result Example**:
+- idc-east-cluster-1: 12 replicas (60% CPU utilization)
+- idc-east-cluster-2: 8 replicas (75% CPU utilization)
+- idc-north-cluster-1: 10 replicas (65% CPU utilization)
+
+**Alternative Using Karmada SpreadConstraints**:
+
+```yaml
+apiVersion: policy.karmada.io/v1alpha1
+kind: PropagationPolicy
+metadata:
+  name: microservice-policy-spread
+spec:
+  resourceSelectors:
+    - apiVersion: apps/v1
+      kind: Deployment
+      name: microservice-app
+  placement:
+    clusterAffinity:
+      labelSelector:
+        matchLabels:
+          env: production
+    spreadConstraints:
+      - spreadByLabel: topology.karmada.io/idc
+        maxGroups: 2
+        minGroups: 2
+    replicaScheduling:
+      replicaSchedulingType: Divided
+  replicas: 30
+```
+
+**Explanation**:
+- `spreadByLabel: topology.karmada.io/idc`: Distribute by IDC label dimension
+- `minGroups: 2`: Ensure deployment in at least 2 IDCs
+- `maxGroups: 2`: Limit deployment to at most 2 IDCs
+- Prerequisite: Clusters need labels `topology.karmada.io/idc`, for example:
+  - East China clusters: `topology.karmada.io/idc: idc-east`
+  - North China clusters: `topology.karmada.io/idc: idc-north`
+- The scheduler automatically distributes replicas between the two IDCs, with specific allocation determined by cluster resource conditions
+
+---
+
+#### Story 2: Specified IDCs with Replica Counts (Based on Strategy 2: SpecifiedIdcs)
+
+**Background**: I need to ensure my core business deploys 20 replicas in East China IDC and 10 replicas in North China IDC (to meet disaster recovery requirements).
+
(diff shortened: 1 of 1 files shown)


RainbowMango:
I can see a lot of interesting use cases. Will give it another look later.

docs/proposals/scheduler-extensibility/README_en.md
```diff
@@ -0,0 +1,1474 @@
...
+```
+
+**Explanation**:
+- `spreadByLabel: topology.karmada.io/idc`: Distribute by IDC label dimension
+- `minGroups: 2`: Ensure deployment in at least 2 IDCs
+- `maxGroups: 2`: Limit deployment to at most 2 IDCs
+- Prerequisite: Clusters need labels `topology.karmada.io/idc`, for example:
+  - East China clusters: `topology.karmada.io/idc: idc-east`
+  - North China clusters: `topology.karmada.io/idc: idc-north`
+- The scheduler automatically distributes replicas between the two IDCs, with specific allocation determined by cluster resource conditions
```
RainbowMango: It seems the alternative solution works. @CharlesQQ Have you tried it? Anything blocks you from this approach?

Personally, I would prefer this solution, but with limited changes:
- Set cluster region `.spec.Region` to all Clusters instead of labels for clarity. And use `spreadByField: region` accordingly. (optional)
- For the replica scheduling strategy, I'd prefer `dynamicWeight` for this use case.

CharlesQQ: From a dynamic scheduling perspective, this is feasible, and it can be perfectly replaced at the API level; However, the difference is that our scheduling strategy includes constraints based on allocation rate thresholds, which the native Karmada scheduler likely doesn't support, or perhaps only supports by estimator, but which we haven't used.

RainbowMango: Oh, I see.  I'm interested in it. 

Can you share some details regarding your `allocation rate thresholds`?
- How to set the thresholds for each cluster? 
- How is the allocation rate of a cluster estimated? Based on total usage across all nodes? Based on what kind of resources? Like CPU/Mem?
- How is the threshold taken into account in the scheduling process?

CharlesQQ: Okay, let me share our current approach:

1. We've customized a scheduling filter plugin based on an allocation rate threshold.

2. The cluster threshold is configured in another etcd instance. If a user modifies it, it will dynamically detect and reload the configuration in memory.

3. The allocation rate is calculated as follows: The `replicaRequirements.nodeClaim` is obtained from `rb/crb`. The denominator is the total resource amount of schedulable nodes calculated based on the `nodeClaim`. The numerator is the sum of pod requests with the same `nodeClaim`, i.e., `sum(pod.request(cpu/memory/gpu))/sum(node.allocatable(cpu/memory/gpu))`.

4. If the current allocation rate of `rb` on a scheduled cluster exceeds the threshold, the cluster is directly filtered out (if replicas already exist, they remain unchanged, and the score plugin will not add new replicas to that cluster).

RainbowMango: > sum(pod.request(cpu/memory/gpu))/sum(node.allocatable(cpu/memory/gpu))

Does that mean the allocation ratio is calculated separately for each resource specified in the requests?
Any resource allocation ratio exceeds the threshold, the filter plugin will reject the target cluster. For instance, if the CPU allocation ratio exceeds the threshold, even if there is enough memory, the target cluster still be rejected, right?

> We've customized a scheduling filter plugin based on an allocation rate threshold.

Is the allocation ratio calculated inside the plugin? It's pretty heavy as the plugin has to iterate through all nodes during each scheduling cycle.

CharlesQQ: > Is the allocation ratio calculated inside the plugin?

Yes

> It's pretty heavy as the plugin has to iterate through all nodes during each scheduling cycle.

The resource data for nodes and pods has already been loaded into memory via Informer. The computation does not involve I/O operations and is relatively fast (in the millisecond range), but it does consume some CPU.

RainbowMango: I know that means the scheduler would consume quite a large memory as it needs to cache all Nodes and Pods from all clusters.

CharlesQQ: > I know that means the scheduler would consume quite a large memory as it needs to cache all Nodes and Pods from all clusters.

That's it

RainbowMango: Thanks for the clarification. Opened https://github.com/karmada-io/karmada/discussions/7137 for tracking the threshold thing, as it is obviously out of the scope of this proposal.

docs/proposals/scheduler-extensibility/README_en.md
```diff
@@ -0,0 +1,1474 @@
...
+#### Story 2: Specified IDCs with Replica Counts (Based on Strategy 2: SpecifiedIdcs)
+
+**Background**: I need to ensure my core business deploys 20 replicas in East China IDC and 10 replicas in North China IDC (to meet disaster recovery requirements).
+
+**Requirements**:
+- East China IDC: 20 replicas
+- North China IDC: 10 replicas
+- Within each IDC, the scheduler automatically distributes based on cluster resources
+
+**Use Case Example**: A recommendation service depends on a third-party service that has 2x more instances in East China IDC than in North China IDC, while also requiring high availability across IDCs. For example, if the third-party service has 100 instances in East China and 50 in North China, the recommendation service needs to deploy in the same ratio (2:1) across both IDCs to ensure call efficiency and high availability while avoiding cross-IDC call latency.
```
RainbowMango: Now I'm looking at the second use story, which is interesting, it requires specifying weight preferences by `regions` instead of `clusters`. This might be a good candidate feature that we can do  .

My questions here are:
- Is the third-party service also managed by Karmada?
- Is there any particular reason that it has to have more instances in East China IDC than in North?

CharlesQQ: > Is the third-party service also managed by Karmada?

Yes

> Is there any particular reason that it has to have more instances in East China IDC than in North?

Basically, it depends on the location of the third-party dependencies. For example, if the dependent MySQL master node is located in the East IDC, the business will prefer to allocate more instances in that IDC for faster read and write speeds. At the same time, to ensure high availability between idcs, some instances will also be deployed to the North IDC.

RainbowMango: I see. That makes sense.

docs/proposals/scheduler-extensibility/README_en.md
```diff
@@ -0,0 +1,1474 @@
...
+---
+
+#### Story 3: Balanced Scheduling within Specified IDCs (Based on Strategy 3: SpecifiedBalancedIdcs)
+
+**Background**: My application requires high availability deployment with balanced replica distribution across all clusters within specified IDCs.
+
+**Requirements**:
+- East China IDC: 20 replicas, balanced across 4 clusters
+- North China IDC: 10 replicas, balanced across 2 clusters
+- Minimize replica count differences between clusters
```
RainbowMango: > balanced across 4 clusters
> balanced across 2 clusters

Here you don't limit the number of selected clusters, just aim to distribute replicas as evenly as possible across the chosen cluster. Is it right?

My question here is how do you handle a case where some clusters don't have sufficient resources?

CharlesQQ: > My question here is how do you handle a case where some clusters don't have sufficient resources?

Yes, this strategy only applies to Kubernetes clusters deployed on the public cloud. Once a cluster's resources exceed the capacity limit, an alert will be triggered. After determining that it is a genuine capacity issue, the SRE will intervene manually to scale up the system.

docs/proposals/scheduler-extensibility/README_en.md
```diff
@@ -0,0 +1,1474 @@
...
+- idc-east-cluster-1: 5 replicas
+- idc-east-cluster-2: 5 replicas
+- idc-east-cluster-3: 5 replicas
+- idc-east-cluster-4: 5 replicas
+- idc-north-cluster-1: 5 replicas
+- idc-north-cluster-2: 5 replicas
+
+---
+
+#### Story 4: Specified Clusters with Replica Counts (Based on Strategy 4: SpecifiedClusters)
```
RainbowMango: This story describes the scenario where people want to control replica assignments precisely; it is mainly used to migrate existing workloads to Karmada, but it can also be used in similar cases (pretty rare).

If Karmada supports this scenario with the PropagationPolicy where people can configure the assignments init, like:
```yaml
spec:
  placement:
    replicas:
      - bj-prod-cluster: 10
      - sh-prod-cluster: 8
      - gz-dr-cluster: 5
```
It looks like a static placement declaration, but a question just raised:
What do you expect from karmada-scheduler? Like, does it need to check the available resources on each member cluster?

docs/proposals/scheduler-extensibility/README_en.md
```diff
@@ -0,0 +1,1474 @@
...
+  replicas: 40
+```
+
+#### Story 5: Integration with CronHPA for Dynamic Scheduling (Custom Feature)
+
+**Background**: Our e-commerce system requires dynamic scaling during promotional periods and needs to propagate deployments to mixed clusters.
+
+**Requirements**:
+- Normal operation with low traffic: Deploy only in self-built IDCs
+- During promotions: Auto-scale using CronHPA, scheduling scaled replicas to cloud vendor clusters (mixed-type clusters)
```
RainbowMango: This scenario is somehow related to the [Extended Cluster Affinities](https://github.com/karmada-io/karmada/pull/7078) feature. @zhzhuang-zju @vie-serendipity. 

But my question here is: is it acceptable to scale the replicas in the self-built IDCs in case there are enough resources?  If not, why?

docs/proposals/scheduler-extensibility/README_en.md
```diff
@@ -0,0 +1,1474 @@
...
+
+---
+
+#### Story 6: NodeLabels Plugin Filters Clusters Without Matching Nodes (Custom Feature)
+
+**Background**: My application needs to run on nodes with specific CPU architectures (like ARM) or specific node labels.
+
+**Requirements**:
+- Only schedule to clusters with ARM nodes
+- Verify node labels before scheduling to avoid scheduling failures
```
RainbowMango: Do you need to check the available resources during the scheduling process?
If you only need a filter plugin to filter clusters with specified nodes according to the `NodeSelector`  configuration in workloads, it's relatively simple.

docs/proposals/scheduler-extensibility/README_en.md
CharlesQQ: > What do you expect from karmada-scheduler? Like, does it need to check the available resources on each member cluster?

Currently, this strategy is mainly used in service migration scenarios. The scheduling result is constrained by the  filtering strategy. If the available resources in the cluster are also filtered, it will have an impact.


In our scenario, the available resource strategy is placed in the assign replicas phase, not the filtering phase, therefore it will not...

CharlesQQ: > As we discussed earlier, my suggestion is that we extract some common requirements from these use cases and try to implement them directly in Karmada wherever possible

Yes, I think a new issue should be created to track this matter.

docs/proposals/scheduler-extensibility/README_en.md
CharlesQQ: This is a cluster that uses a time-sharing mechanism for online and offline services; when the online service is scaled down, the offline service is scaled up, and the cluster resources are always fully utilized.

Offline services are used for some offline task calculations.

docs/proposals/scheduler-extensibility/README_en.md
CharlesQQ: Yes, we've added a filtering plugin for NodeSelector; if it's not the SpecifiedClusters strategy, the available resources will be checked during the assign replicas phase.

PR #7049, opened 2025-12-26
feat: expose lua vm pool size as flag

vie-serendipity:
**What type of PR is this?**

/kind feature

We should consider to expose lua vm pool size to flags. So it can be adjusted. When `ConcurrentXXXXSyncs` is much more than 10, 10 lua vm might not be enough.

Fixes #

```release-note
1. `karmada-controller-manager`: Add new flags(--configurable-interpreter-lua-vm-pool-size and --third-party-interpreter-lua-vm-pool-size)
2. `karmada-agent`: Add new flags(--configurable-interpreter-lua-vm-pool-size and --third-party-interpreter-lua-vm-pool-size)
```

### DIFF
--- a/cmd/agent/app/agent.go
+++ b/cmd/agent/app/agent.go
@@ -272,7 +272,7 @@ func setupControllers(ctx context.Context, mgr controllerruntime.Manager, opts *
 	sharedFactory.Start(ctx.Done())
 	sharedFactory.WaitForCacheSync(ctx.Done())
 
-	resourceInterpreter := resourceinterpreter.NewResourceInterpreter(controlPlaneInformerManager, serviceLister)
+	resourceInterpreter := resourceinterpreter.NewResourceInterpreter(controlPlaneInformerManager, serviceLister, opts.ConfigurableInterpreterLuaVMPoolSize, opts.ThirdPartyInterpreterLuaVMPoolSize)
 	if err := resourceInterpreter.Start(ctx); err != nil {
 		return fmt.Errorf("failed to start resource interpreter: %w", err)
 	}
--- a/cmd/agent/app/options/options.go
+++ b/cmd/agent/app/options/options.go
@@ -78,6 +78,11 @@ type Options struct {
 	// KubeAPIBurst is the burst to allow while talking with karmada-apiserver.
 	KubeAPIBurst int
 
+	// ConfigurableInterpreterLuaVMPoolSize is the size of the pool for the configurable interpreter lua vm.
+	ConfigurableInterpreterLuaVMPoolSize int
+	// ThirdPartyInterpreterLuaVMPoolSize is the size of the pool for the third party interpreter lua vm.
+	ThirdPartyInterpreterLuaVMPoolSize int
+
 	ClusterCacheSyncTimeout metav1.Duration
 	// ResyncPeriod is the base frequency the informers are resynced.
 	// Defaults to 0, which means the created informer will never do resyncs.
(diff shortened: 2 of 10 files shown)


zhzhuang-zju:
Hi @vie-serendipity, does each operation of the resource interpreter currently occupy a separate Lua VM?

XiShanYongYe-Chang:
I'm glad to see the feedback on Lua pool size.

I would like to know more about the specific impact of the current default value of 10. Will it affect any particular speed?

XiShanYongYe-Chang:
> Hi @vie-serendipity, does each operation of the resource interpreter currently occupy a separate Lua VM?

I understand that all current operations should be shared, but custom interpreters and third-party interpreter are separate.

zhzhuang-zju:
> I understand that all current operations should be shared, but custom interpreters and third-party interpreter are separate.

Thanks for the clarification. Based on the current implementation, each Lua script execution acquires a VM from the VM pool and returns it after use. Therefore, it can be understood that both custom interpreters and third-party interpreters can each handle up to 10 concurrent resource interpretations simultaneously. And this concurrency limit could become a bottleneck for QPS.

https://github.com/karmada-io/karmada/blob/1b4e9c9d3c52c0fa0ade607ba9b8f55adde315ec/pkg/resourceinterpreter/customized/declarative/luavm/lua.go#L73-L79

vie-serendipity:
> I would like to know more about the specific impact of the current default value of 10. Will it affect any particular speed?

I wanted rb controller to create RB as quickly as possible, so I increased the concurrency, and it was helpful initially. But gradually, concurrency stopped being effective.

I identified that it was likely the size limitation of the Lua VM pool that was causing a significant amount of CPU-intensive operations. Obviously, increasing concurrency doesn't improve QPS for CPU-bound operations.

RainbowMango:
> I identified that it was likely the size limitation of the Lua VM pool that was causing a significant amount of CPU-intensive operations. Obviously, increasing concurrency doesn't improve QPS for CPU-bound operations.

Does any evidence show that the bottleneck is the size of the Lua VM pool?

vie-serendipity:
1. Flame graphs reveal that the revise operation is taking a significant amount of time, while other client activities are mostly IO operations. Theoretically, concurrency should be able to improve IO operations.
image

2. After I increased lua vm pool size, the RB controller QPS increases accordingly.

XiShanYongYe-Chang:
I feel that your evidence is somewhat convincing. 

Besides the revise operation, have there been any other resource interpreter operations that encountered bottlenecks?

> I wanted rb controller to create RB as quickly as possible

This likely refers to creating or updating work resources.

vie-serendipity:
> Besides the revise operation, have there been any other resource interpreter operations that encountered bottlenecks?

The detector, RB controller, and RB status controller all have this bottleneck, as they are all quite dependent on the interpreter's work.

> This likely refers to creating or updating work resources.

Yes, you're absolutely right. Sry for my mistake. I capture a new image.
image

XiShanYongYe-Chang:
I think it's fine. Let's see how @RainbowMango thinks. Tomorrow is the community meeting; should we just go ahead and talk about it?

vie-serendipity:
Thank you for the guidance of @RainbowMango in the community meeting, my assumption was unconfirmed.

Further investigation in the CPU Profile flame graph confirmed that the QPS of RB controller was not due to the creation and destruction of lua vm.

The following metrics indicate that despite the pool size being 10, there is no frequent creation and destruction happening.
image

image

However, increasing the Lua VM pool size did improve the performance. I doubt that it was due to hot locking, but the evidence shows that it is not the case.

image

vie-serendipity:
Finally, I think this might be related to `github.com/yuin/gopher-lua` implementation. I have temporarily achieved the desired outcome by adjusting the pool size.

I don't have much familiarity with Lua itself, I won't proceed with further problem localization in the short term. If the person who initially implemented this part had some insights, I would be very grateful to hear your troubleshooting approach.

RainbowMango:
> Finally, I think this might be related to github.com/yuin/gopher-lua implementation. I have temporarily achieved the desired outcome by adjusting the pool size.

That's a nice input and gives us a good reason to dig into it. 

> I don't have much familiarity with Lua itself, I won't proceed with further problem localization in the short term. If the person who initially implemented this part had some insights, I would be very grateful to hear your troubleshooting approach.

Don't worry, @XiShanYongYe-Chang is looking at it. I believe he will leave the findings here.

XiShanYongYe-Chang:
@vie-serendipity Thank you for providing some direction.

Issue #7050, opened 2025-12-26
Fix unsafe field access and division by zero in FlinkDeployment GetReplicas Lua script


XiShanYongYe-Chang:
The `GetReplicas` Lua script in `customizations.yaml` has some potential bugs that are not covered by the current test cases. The script can fail with a runtime error under certain conditions.

The `GetReplicas` script has the following issues:

1. Unsafe field access: It accesses nested fields like observedObj.spec.jobManager.replicas directly. This will cause an error if observedObj.spec or observedObj.spec.jobManager is nil. The GetComponents script uses a safe get function, and GetReplicas should do the same.

2. Potential division by zero: The script calculates tm_replicas with math.ceil(parallelism / observedObj.spec.flinkConfiguration['taskmanager.numberOfTaskSlots']). If taskmanager.numberOfTaskSlots is "0", this will cause a division-by-zero error. The GetComponents script correctly checks for task_slots == 0.

To ensure complete and robust test coverage, please add test cases that would expose these bugs. For example:

- A test case where spec.jobManager is missing.
- A test case where spec.flinkConfiguration['taskmanager.numberOfTaskSlots'] is "0".
Fixing the underlying Lua script is highly recommended. The implementation of GetComponents is a good reference for a more robust implementation.

ref https://github.com/karmada-io/karmada/pull/7047#discussion_r2647457842

Issue #6998, opened 2025-12-05
[Umbrella] Multi-components workload scheduling - phase ​​III


RainbowMango:
**Summary**:

This issue is used to track the iteration tasks of feature `Multi-components workload scheduling` described at #5115.
See [proposal](https://github.com/karmada-io/karmada/blob/master/docs/proposals/scheduling/multi-podtemplate-support/multiple-pod-template-support.md) for more details.

**Iteration Tasks -- Part-1: Bug fixes, enhancements, testing**:
- [x] Cluster failover does not work as expected with MultiplePodTemplateScheduling. ([#7065](https://github.com/karmada-io/karmada/issues/7065)) (@mszacillo, #7066)
- [x] refactor: unify maxAvailableComponentsSets calculation. (@zhzhuang-zju , #6997)

**Iteration Tasks -- Part-2: XXX**:
- Support multi-template applications scale sceanrios
  - [ ] Ensure multi-template applications have the opportunity to enter the rescheduling process.
  - [ ] During rescheduling, replica set estimation should take scheduled components into account.
    - For scale-up, only the incremental components/replicas need to be considered.
    - For scale-down, replica set estimation could be skipped.
  - [ ] If rescheduling fails, ensure the updated configuration is not propagated to member clusters.

RainbowMango:
@mszacillo @seanlaii @zhzhuang-zju Just opened this issue to track the tasks for this feature. Please leave your thoughts here if you think anything should be done in the next iteration.

seanlaii:
I can help add the RayService interpreter and the e2e tests for RayJob/RayCluster/RayService similar to https://github.com/karmada-io/karmada/pull/6914.

zhzhuang-zju:
[refactor: unify maxAvailableComponentsSets calculation](https://github.com/karmada-io/karmada/pull/6997) can be included in the scope of phase III tasks.

RainbowMango:
> I can help add the RayService interpreter and the e2e tests for RayJob/RayCluster/RayService similar to [#6914](https://github.com/karmada-io/karmada/pull/6914).

All right! Please link to this issue or leave a comment here once you have made it.

zhzhuang-zju:
Now that `Iteration Tasks — Part 1` has been completed, I can begin research and work on `Iteration Tasks — Part 2`.
/assign

zhzhuang-zju:
> For scale-up, only the incremental components/replicas need to be considered.

[...]

PR #7051, opened 2025-12-26
fix unsafe field in FlinkDeployment GetReplicas Lua script

FAUST-BENCHOU:
**What type of PR is this?**

Fixes https://github.com/karmada-io/karmada/issues/7050

```release-note

```

### DIFF
--- a/pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/flink.apache.org/v1beta1/FlinkDeployment/customizations.yaml
+++ b/pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/flink.apache.org/v1beta1/FlinkDeployment/customizations.yaml
@@ -123,8 +123,12 @@ spec:
 
           local jm_cpu    = get(observedObj, {"spec","jobManager","resource","cpu"})
           local jm_memory = get(observedObj, {"spec","jobManager","resource","memory"})
-          jm_requires.resourceRequest.cpu    = jm_cpu
-          jm_requires.resourceRequest.memory = kube.getResourceQuantity(jm_memory)
+          if jm_cpu ~= nil then
+            jm_requires.resourceRequest.cpu = jm_cpu
+          end
+          if jm_memory ~= nil then
+            jm_requires.resourceRequest.memory = kube.getResourceQuantity(jm_memory)
+          end
           apply_pod_template(pt_spec, jm_requires)
 
           local jobManagerComponent = {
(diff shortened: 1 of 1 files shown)


pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/flink.apache.org/v1beta1/FlinkDeployment/customizations.yaml
```diff
@@ -123,8 +123,12 @@ spec:
...
           local jm_cpu    = get(observedObj, {"spec","jobManager","resource","cpu"})
           local jm_memory = get(observedObj, {"spec","jobManager","resource","memory"})
-          jm_requires.resourceRequest.cpu    = jm_cpu
-          jm_requires.resourceRequest.memory = kube.getResourceQuantity(jm_memory)
+          if jm_cpu ~= nil then
+            jm_requires.resourceRequest.cpu = jm_cpu
+          end
+          if jm_memory ~= nil then
+            jm_requires.resourceRequest.memory = kube.getResourceQuantity(jm_memory)
+          end
```
XiShanYongYe-Chang, line 131: Is the effect of this modification the same as before it was made?

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/flink.apache.org/v1beta1/FlinkDeployment/customizations.yaml
```diff
@@ -167,63 +175,103 @@ spec:
           return components
         end
     replicaResource:
-      luaScript: >
```
XiShanYongYe-Chang, line 170: Both of these symbols should be acceptable.

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/flink.apache.org/v1beta1/FlinkDeployment/customizations.yaml
FAUST-BENCHOU: The effect is the same, as the logic is essentially just adding a check to ensure that the values are only assigned when they are not nil.

XiShanYongYe-Chang:
I don't think there's a big issue. Let's merge it first. If any problems arise later, we can continue to address them.

/lgtm
/approve

mszacillo:
Firstly, thanks for the fix @FAUST-BENCHOU!

@XiShanYongYe-Chang, apologies for not being able to review, I was away for the holidays. Now that we've released multiple pod template support in v1.16, I'm not sure if we need `GetReplicas` at all. That method was more of a stopgap so that Karmada could interpret the resource requirement of FlinkDeployments while we worked on multiple pod template support. But the method is not perfect, and will overestimate the total resource requirement. What do you think?

XiShanYongYe-Chang:
> apologies for not being able to review, I was away for the holidays

No problem, have a happy holiday. :)

> Now that we've released multiple pod template support in v1.16, I'm not sure if we need GetReplicas at all. That method was more of a stopgap so that Karmada could interpret the resource requirement of FlinkDeployments while we worked on multiple pod template support. But the method is not perfect, and will overestimate the total resource requirement. What do you think?

If that's the case, I think it might be better for us to remove the implementation of `GetReplicas`, wdyt?

RainbowMango:
> If that's the case, I think it might be better for us to remove the implementation of GetReplicas, wdyt?

Yes, but no rush. 
The `MultiplePodTemplatesScheduling` feature is now limited to the case that FlinkDeployment goes to a single cluster, they are not completely idempotent, which is exactly what we should consider for the following iterations(probably in #6998)

XiShanYongYe-Chang:
Thanks for the explanation.

PR #7052, opened 2025-12-28
Add interpreter test coverage for SidecarSet

Abhay349:
This PR adds  interpreter test coverage for SidecarSet.

### DIFF
--- a/pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/apps.kruise.io/v1alpha1/SidecarSet/testdata/aggregatestatus-test.yaml
+++ b/pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/apps.kruise.io/v1alpha1/SidecarSet/testdata/aggregatestatus-test.yaml
@@ -62,3 +62,16 @@ statusItems:
       readyPods: 1
 operation: AggregateStatus
 output:
+  aggregatedStatus:
+    apiVersion: apps.kruise.io/v1alpha1
+    kind: SidecarSet
+    metadata:
+      name: sample-sidecarset
+      namespace: test-sidecarset
+      generation: 1
+      labels:
+        app: sample
+    status:
+      matchedPods: 3
+      updatedPods: 3
+      readyPods: 3
--- a/pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/apps.kruise.io/v1alpha1/SidecarSet/testdata/interpretdependency-test.yaml
+++ b/pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/apps.kruise.io/v1alpha1/SidecarSet/testdata/interpretdependency-test.yaml
@@ -1,5 +1,7 @@
 # test case for interpreting dependency of SidecarSet
 # case1. SidecarSet with configmap and secret dependency
+# case2. SidecarSet with volume configmap and secret dependency
+# case3. SidecarSet with no dependencies
 
 name: "SidecarSet with configmap and secret dependency"
 description: "Test interpreting dependency of SidecarSet with configmap and secret dependency"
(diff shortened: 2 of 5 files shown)


XiShanYongYe-Chang:
Thanks
/assign

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/apps.kruise.io/v1alpha1/SidecarSet/testdata/aggregatestatus-test.yaml
```diff
@@ -1,5 +1,6 @@
 # test case for aggregating status of SidecarSet
 # case1. SidecarSet with two status items
+# case2. SidecarSet with single status item
```
XiShanYongYe-Chang: Based on the logic being tested, case 1 should be sufficient.

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/apps.kruise.io/v1alpha1/SidecarSet/testdata/interpretdependency-test.yaml
```diff
@@ -1,5 +1,8 @@
 # test case for interpreting dependency of SidecarSet
 # case1. SidecarSet with configmap and secret dependency
+# case2. SidecarSet with only configmap dependency
+# case3. SidecarSet with only secret dependency
```
XiShanYongYe-Chang: I understand that both of these cases can be covered by case 1, so they are actually redundant. In addition, there is logic in the tested code that retrieves references from the volume, and this part of the code is not covered by the tests. Could you help add some test cases for this?

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/apps.kruise.io/v1alpha1/SidecarSet/testdata/interpretreplica-test.yaml
```diff
@@ -1,5 +1,6 @@
 # test case for interpreting replica of SidecarSet
 # case1. SidecarSet replicas should always be 0
+# case2. SidecarSet replicas should be 0 even without spec
```
XiShanYongYe-Chang: Case 2 shouldn't be necessary, right?

XiShanYongYe-Chang (commented):
Thanks~

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/apps.kruise.io/v1alpha1/SidecarSet/testdata/interpretreplica-test.yaml
Ferryx349: Yeah, I initially added to explicitly cover the scenario where spec is absent, but since InterpretReplica always returns 0 regardless of spec presence, case 1 already covers this behavior. I’ll remove case 2 to avoid redundancy.

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/apps.kruise.io/v1alpha1/SidecarSet/testdata/interpretdependency-test.yaml
Ferryx349: I added them to  validate configmap and secret handling separately for clarity.
But I agree they don’t add extra coverage beyond case 1. I’ll remove them and add tests for volume-based dependencies instead.

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/apps.kruise.io/v1alpha1/SidecarSet/testdata/aggregatestatus-test.yaml
Ferryx349: I added the extra case initially to be thorough and cover both single and multiple status items as a learner.
But I agree that it doesn’t add additional coverage beyond case 1. I’ll remove it and keep the tests minimal.

PR #7066, opened 2025-12-31
Scheduler should detect change in replicas for multiple component workloads

### LINKED ISSUE #7065: Cluster failover does not work as expected with MultiplePodTemplateScheduling
**What happened**:

We were testing our internal fork which was recently rebased on top of the 1.16 release. As part of testing we did some cluster failovers with MultiplePodTemplatesScheduling feature set to true, but noticed that failover does not work as expected. Once workloads are evicted from their cluster, they do not get rescheduled. 

The root cause seems to be due to the existing implementation of [IsBindingReplicasChanged](https://github.com/karmada-io/karmada/blob/master/pkg/scheduler/scheduler.go#L409), which only takes spec.replicas into account. We need to also account for components.

Scheduler should correctly detect when replicas have been scaled up or down when using multiple component scheduling.

Attempt a cluster failover for a workload that has multiple components with the MultiplePodTemplatesScheduling feature set to true. This will be reproducible every time.

mszacillo:
**What type of PR is this?**
/kind bug

Fixes bug which prevents cluster failover for multiple component workloads when MultiplePodTemplatesScheduling is enabled. 

Fixes #7065

```release-note
`karmada-scheduler`: Fixes bug preventing multi-component workloads from being rescheduled during cluster failover events
```

### DIFF
--- a/pkg/util/binding.go
+++ b/pkg/util/binding.go
@@ -22,6 +22,7 @@ import (
 
 	policyv1alpha1 "github.com/karmada-io/karmada/pkg/apis/policy/v1alpha1"
 	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
+	"github.com/karmada-io/karmada/pkg/features"
 )
 
 // GetBindingClusterNames will get clusterName list from bind clusters field
@@ -38,6 +39,20 @@ func IsBindingReplicasChanged(bindingSpec *workv1alpha2.ResourceBindingSpec, str
 	if strategy == nil {
 		return false
 	}
+
+	// For multi-component workloads, trigger rescheduling when clusters are empty (e.g., after eviction).
+	// This is a temporary fix to ensure cluster failover works correctly.
+	// Limitation: This only handles the failover scenario where clusters are cleared.
+	// It does not detect component replica changes (e.g., scale up/down) or replica swaps between components.
+	// A complete solution requires changing how scheduling results are stored to support multi-template workloads,
+	// likely by extending TargetCluster to include per-component replica information.
+	// The comprehensive solution is tracked by: https://github.com/karmada-io/karmada/issues/6998
+	if features.FeatureGate.Enabled(features.MultiplePodTemplatesScheduling) && len(bindingSpec.Components) > 1 {
+		if len(bindingSpec.Clusters) == 0 {
+			return true
+		}
+	}
+
 	if strategy.ReplicaSchedulingType == policyv1alpha1.ReplicaSchedulingTypeDuplicated {
 		for _, targetCluster := range bindingSpec.Clusters {
 			if targetCluster.Replicas != bindingSpec.Replicas {


zhzhuang-zju:
/assign

I recall encountering a similar issue in scaling scenarios before. I’ll look into it further to see if it’s a common pattern.

pkg/util/binding.go
```diff
@@ -38,21 +38,41 @@ func IsBindingReplicasChanged(bindingSpec *workv1alpha2.ResourceBindingSpec, str
...
+}
+
+// GetSumofReplicasForComponents will get the sum of replicas for multi-component resources
+func GetSumOfReplicasForComponents(components []workv1alpha2.Component) int32 {
+	replicasSum := int32(0)
+	for _, component := range components {
+		replicasSum += component.Replicas
+	}
+	return replicasSum
+}
```
RainbowMango: A corner case would break down this:
Let's say a FlinkDeployment with two components:
- jobManager, replicas  1
- taskManager, replicas 3
If we swap the replicas between the two components, the total replicas doesn't change.

RainbowMango (commented):
I thought this was a simple issue, but it might involve how to record the scheduling results for multi-template workloads and how to detect changes and trigger re-scheduling.

I need a bit more time to look into it. 
Thanks @mszacillo for bringing this up, this is indeed a bug.

pkg/util/binding.go
mszacillo: Ah, good catch. Perhaps we can consider extending the ClusterInfo that is currently in the ResourceBinding?

```golang
// TargetCluster represents the identifier of a member cluster.
type TargetCluster struct {
	// Name of target cluster.
	Name string `json:"name"`
	// Replicas in target cluster
	// +optional
	Replicas int32 `json:"replicas,omitempty"`
}
```

This could be extended to include components.

RainbowMango: Yeah, that's exactly what I'm thinking too. In addition, I'm also thinking of deprecating `.spec.replicaRequirements` and `.spec.replicas` and replacing them with the `.spec.components`, as part of the multi-components workload scheduling feature(#6998).

Given that they are widely used across the whole codebase, it's a little bit challenging to do that in a compatible and smooth way.

zhzhuang-zju:
The previous behavior of multi-components scheduling was: among multiple candidate clusters, select the one with relatively sufficient resources for scheduling. Due to the absence of a rescheduling (re-entrant scheduling) path, the initial scheduling decision was never revised once made.

This fix aims to add a re-entrant scheduling path for multi-components scheduling, but the following points need further confirmation:

1. It’s clear that failover should trigger the scheduling process—but what about scaling operations (scale-up or scale-down)? For example, if cluster member1 was initially selected, but the workload later scales up such that member1 can no longer satisfy the resource requirements, what should the expected behavior be?

2. If scaling operations also trigger the scheduling process, I’ve identified an issue: when calculating `MaxAvailableComponentSets` each time, the previous scheduling result isn’t taken into account. For instance, during the first calculation, cluster member1 might support 1 component set, but during the second calculation, its available capacity drops to 0—because resources allocated by the prior scheduling decision are already occupying member1’s capacity.

WDYT? @mszacillo @RainbowMango

mszacillo:
> It’s clear that failover should trigger the scheduling process—but what about scaling operations (scale-up or scale-down)? For example, if cluster member1 was initially selected, but the workload later scales up such that member1 can no longer satisfy the resource requirements, what should the expected behavior be?

At least in our use-case, if someone is scaling up their workload, then Karmada should keep the application in the cluster that it is currently scheduled to. This is the current behavior when you use an aggregated replicaDivisionPreference. If for instance I scale up the parallelism of a FlinkDeployment currently comprising of {1 JM, 5 TMs} -> {1 JM, 6 TMs}, only the delta of 1 replica will be taken into account as part of [dynamicScaleUp](https://github.com/karmada-io/karmada/blob/b175217a30d9fffc4c14e0ef1a112cc9c9e676cc/pkg/scheduler/core/division_algorithm.go#L123). 

Ideally we should keep the behavior similar in the component case. 

> If scaling operations also trigger the scheduling process, I’ve identified an issue: when calculating MaxAvailableComponentSets each time, the previous scheduling result isn’t taken into account. For instance, during the first calculation, cluster member1 might support 1 component set, but during the second calculation, its available capacity drops to 0—because resources allocated by the prior scheduling decision are already occupying member1’s capacity.

Yeah this is the more complicated issue. :/ 

Thinking out loud, I wonder if we can update what the component estimator computes depending on mode:

If `Fresh` (or if it is Steady but being scheduled for the first time): Calculate maxSets 
If `Steady`: Calculate feasibility of the incremental change:
- Determine per-component delta (desired - currently scheduled to cluster)
- Evaluate quota against just that delta resource vector

For currently scheduled replicas, we'd need to reference the ClusterInfo as mentioned above. That said this would require a good amount of changes to how we call `MaxAvailableComponentSets`, since we just package `Components: spec.Components,` directly into the `ComponentSetEstimationRequest`.

zhzhuang-zju:
>  if someone is scaling up their workload, then Karmada should keep the application in the cluster that it is currently scheduled to.

Yeah, I agree with this strategy, as it best ensures business continuity.

However, if scaling operations (scale-up or scale-down) also retrigger the scheduling process, the current implementation cannot guarantee this behavior. I think this is also worth noting.

RainbowMango:
Great talk! 

Currently, scale-up and scale-down operations for multi-template applications bypass the scheduler and are automatically synced to member clusters by the controller. I think we need to improve this behavior, and I've documented my thoughts in issue #6998.

As for resolving the current failover issue, I believe a simple approach could work—for example:
```diff
diff --git a/pkg/util/binding.go b/pkg/util/binding.go                          
index 4b1dcdbf9..ba67e976c 100644                                               
--- a/pkg/util/binding.go                                                       
+++ b/pkg/util/binding.go                                                       
@@ -22,6 +22,7 @@ import (                                                      
                                                                                
        policyv1alpha1 "github.com/karmada-io/karmada/pkg/apis/policy/v1alpha1" 
        workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"     
+       "github.com/karmada-io/karmada/pkg/features"                            
 )
 
 // GetBindingClusterNames will get clusterName list from bind clusters field
@@ -38,6 +39,20 @@ func IsBindingReplicasChanged(bindingSpec *workv1alpha2.ResourceBindingSpec, str
        if strategy == nil {
                return false
        }
+
+       // For multi-component workloads, trigger rescheduling when clusters are empty (e.g., after eviction).
+       // This is a temporary fix to ensure cluster failover works correctly.
+       // Limitation: This only handles the failover scenario where clusters are cleared.
+       // It does not detect component replica changes (e.g., scale up/down) or replica swaps between components.
+       // A complete solution requires changing how scheduling results are stored to support multi-template workloads,
+       // likely by extending TargetCluster to include per-component replica information.
+       // The comprehensive solution is tracked by: https://github.com/karmada-io/karmada/issues/6998
+       if features.FeatureGate.Enabled(features.MultiplePodTemplatesScheduling) && len(bindingSpec.Components) > 1 {
+               if len(bindingSpec.Clusters) == 0 {
+                       return true
+               }
+       }
+
        if strategy.ReplicaSchedulingType == policyv1alpha1.ReplicaSchedulingTypeDuplicated {
                for _, targetCluster := range bindingSpec.Clusters {
                        if targetCluster.Replicas != bindingSpec.Replicas {

```

zhzhuang-zju:
> As for resolving the current failover issue, I believe a simple approach could work

+1
This approach is only triggered on failover, so it won't broaden the impact of the modification.

zhzhuang-zju:
Hi @mszacillo, in yesterday’s meeting, you mentioned this PR had been updated, but I don’t see the new changes yet. No rush, just wanted to sync up! 😄

mszacillo:
> Hi @mszacillo, in yesterday’s meeting, you mentioned this PR had been updated, but I don’t see the new changes yet. No rush, just wanted to sync up! 😄

Updated! I also removed the test cases since they are no longer relevant. We can add more unit tests once we implement the longer term strategy that @RainbowMango has documented.

zhzhuang-zju:
Thanks
/lgtm

Additionally, the `Does this PR introduce a user-facing change?:` part can be updated to:
```bash
`karmada-scheduler`: Fixed the issue that in failover scenarios, multi-component workloads were evicted from their clusters but failed to trigger rescheduling onto new clusters.
```

cc @RainbowMango

mszacillo:
Updated the comment!

zhzhuang-zju:
@mszacillo Should this fix be backported to previous branches?

RainbowMango:
> @mszacillo Should this fix be backported to previous branches?

It's @mszacillo's call :)

mszacillo:
> @mszacillo Should this fix be backported to previous branches?

@zhzhuang-zju I'd backport this to v1.16 release. v1.15 does not need it since multiplePodTemplateScheduling was not fully supported during that release. I can create a cherrypick PR.

Issue #6952, opened 2025-11-25
[Umbrella] Enhance third-party resource interpreter tests with output verification


XiShanYongYe-Chang:
/kind cleanup

We wish to enhance the testing framework for third-party resource interpreter customizations by adding support for verifying Lua script execution results against expected outputs.

In addition, the current test coverage for Lua script content is insufficient. Each operation of each resource has only one test case, which is not enough to fully cover the logic. Therefore, we also need to supplement the test cases for individual operations of individual resources.

Below, let's list all the subtasks that need to be completed for the current work item.

- [x] enhance resource interpreter test framework @zhzhuang-zju #6938 
- [ ] Complete test cases to cover the Lua script logic
  - [ ] apps.kruise.io
    - [ ] v1alpha1
      - [ ] AdvancedCronJob @XiShanYongYe-Chang #6991 
      - [ ] BroadcastJob
      - [ ] CloneSet
      - [ ] DaemonSet
      - [ ] SidecarSet
      - [ ] UnitedDeployment
    - [ ] v1beta1 - StatefulSet
  - [ ] argoproj.io - v1alpha1 - Workflow
  - [ ] flink.apache.org - v1beta1 - FlinkDeployment
  - [ ] helm.toolkit.fluxcd.io - v2beta1 - HelmRelease
  - [ ] kubeflow.org - v1 - TFJob
  - [ ] kustomize.toolkit.fluxcd.io - v1 - Kustomization
  - [ ] kyverno.io - v1
    - [ ] ClusterPolicy
    - [ ] Policy
  - [ ] source.toolkit.fluxcd.io
    - [ ] v1 - GitRepository
    - [ ] v1beta2
      - [ ] Bucket
      - [ ] HelmChart
      - [ ] HelmRepository
      - [ ] OCIRepository

If anyone is willing to get involved and complete some of these subtasks, we would be very grateful and welcome your participation. 

You can directly tell me in the comments which subtasks you want to complete, and I will assign them to you.

yaten2302:
Hi @XiShanYongYe-Chang , I would like to work on the `argoproj.io - v1alpha1 - Workflow` one. Kindly assign that to me :)
Also, how can I verify these tests locally?

XiShanYongYe-Chang:
> Hi @XiShanYongYe-Chang , I would like to work on the argoproj.io - v1alpha1 - Workflow one. Kindly assign that to me :)

Ok, thx~

> Also, how can I verify these tests locally?

[...]

PR #7070, opened 2026-01-04
Add complete interpreter test coverage for kubeflow.org - v1 - Notebook

FAUST-BENCHOU:
**What type of PR is this?**

Part of https://github.com/karmada-io/karmada/issues/6952

```release-note
NONE
```

### DIFF
--- a/pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/kubeflow.org/v1/Notebook/testdata/interpretstatus-test.yaml
+++ b/pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/kubeflow.org/v1/Notebook/testdata/interpretstatus-test.yaml
@@ -1,8 +1,11 @@
 # test case for interpreting status of Notebook
-# case1. Notebook: interpret status test
+# case1. Notebook: interpret status test with full status
+# case2. Notebook with nil status should return empty status
+# case3. Notebook with empty status should return empty status
+# case4. Notebook with partial status fields
 
-name: "Notebook: interpret status test"
-description: "Test interpreting status for Notebook"
+name: "Notebook: interpret status test with full status"
+description: "Test interpreting status for Notebook with complete status fields"
 observedObj:
   apiVersion: kubeflow.org/v1
   kind: Notebook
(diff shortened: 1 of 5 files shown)


pkg/resourceinterpreter/default/thirdparty/thirdparty_test.go
```diff
@@ -160,6 +160,10 @@ func getAllTestCases(t *testing.T, testDataDir string) TestStructure {
 				t.Fatal(fmt.Errorf("failed to decode file %s: %v", path, err))
 			}
 			test.Filepath = path
+			// Replace {{NOW_TIMESTAMP}} placeholders in Output with current time
+			if test.Output != nil {
+				test.Output = replaceTimestampPlaceholders(test.Output).(map[string]interface{})
```
XiShanYongYe-Chang: I understand that in most cases, comparisons should not be a problem.
Could the test script execution taking a slightly longer time lead to less consistent timing?

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/kubeflow.org/v1/Notebook/testdata/interpretstatus-test.yaml
```diff
@@ -38,3 +41,92 @@ observedObj:
...
+        reason: "ContainerCreating"
+        message: "Creating container"
+operation: InterpretStatus
+output:
+  status:
+    readyReplicas: 2
+    containerState:
+      waiting:
+        reason: "ContainerCreating"
+        message: "Creating container"
```
XiShanYongYe-Chang: Can you help add an empty line at the end of the file?

pkg/resourceinterpreter/default/thirdparty/thirdparty_test.go
FAUST-BENCHOU: I did consider adding some time tolerance, but that would require modifying deepequal, which could be quite a significant change.

XiShanYongYe-Chang: Comparing times can indeed be troublesome, or perhaps we could consider not comparing times at all. Still, it is necessary to modify the `deepEqual` function.

FAUST-BENCHOU: agree that, currently there aren't many Lua scripts that use the now time. We can directly specify fields that use the now time to exclude them in deepequal.

FAUST-BENCHOU:
@XiShanYongYe-Chang I've come up with a more general solution: we can retain the tagging concept and use tags like {{EXCLUDE}} to select the parts we don't want to compare, for example,
```          
message: "All notebooks are ready"
lastProbeTime: "{{EXCLUDE}}"
lastTransitionTime: "{{EXCLUDE}}"
```
it wont compare lastProbeTime and lastTransitionTime then.i think its better than defining Hard-coded fields.what do u think?

XiShanYongYe-Chang:
> it wont compare lastProbeTime and lastTransitionTime then.i think its better than defining Hard-coded fields.what do u think?

Good idea.

XiShanYongYe-Chang:
If it's ready for review, please cc me again.

FAUST-BENCHOU:
> If it's ready for review, please cc me again.如果准备好审核，请再次抄送给我。

@XiShanYongYe-Chang it's ok to review now,i'm just waiting for ci🙃

XiShanYongYe-Chang:
Thanks, can you help squash the commits into one?

FAUST-BENCHOU:
> Thanks, can you help squash the commits into one?谢谢，你能帮忙把这些提交合并成一个吗？

Sure.updated :)

PR #7074, opened 2026-01-04
Fixed the issue that watch connect cannot reflect resource from recovered clusters immediately

### LINKED ISSUE #6963: [Search] Search component cannot immediately reflect resources from recovered clusters in existing watch connections
## Background

We encountered this issue during chaos engineering tests with our deployment platform that uses Karmada's search component to aggregate Pod views from multiple Kubernetes clusters.

## Environment

- Karmada version: v1.14.5, K8s version: v1.19.3
- Two Kubernetes clusters: K8s1 (healthy) and K8s2 (subject to fault injection)
- Internal deployment platform uses informer to watch aggregated Pod resources through search component

## Steps to Reproduce

1. **Initial State**: Both K8s1 and K8s2 are healthy and registered with Karmada. The deployment platform has an active watch connection to search component showing aggregated Pods from both clusters.
2. **Fault Injection**: Inject network fault into K8s2 (drop all network packets), causing K8s2 to become `NotReady`.
3. **Search Component Behavior**: Search component stops the informer for K8s2 cluster (as designed in `controller.go:262-286`):

   ```go
   func (c *Controller) clusterAbleToCache(cluster string) (cls *clusterv1alpha1.Cluster, able bool, err error) {
       if !util.IsClusterReady(&cls.Status) {
           klog.Warningf("cluster %s is notReady try to stop this cluster informer", cluster)
           c.InformerManager.Stop(cluster)
           return
       }
   }
   ```
4. **Fault Recovery**: Remove network fault from K8s2. K8s2 becomes `Ready` again.
5. **Search Component Recovery**: Search component restarts the informer for K8s2 and begins caching resources.
6. **Resource Changes**: Create new Pods or modify existing Pods in K8s2.
7. **Issue Observed**: The deployment platform's existing watch connection **cannot see the new resources from K8s2**. The Pod list remains in the pre-fault state.
8. **Delayed Resolution**: After 5-10 minutes, the deployment platform finally sees the updated Pod list from K8s2.

### Why 5-10 Minutes Delay?

The delay is due to client-go's watch timeout mechanism:

- `MinWatchTimeout` is default to 5 minutes in client-go
- Actual timeout is randomized between `minWatchTimeout` and `2 * minWatchTimeout` (5-10 minutes)
- Only after watch timeout triggers reconnection, the new `Watch()` call includes K8s2

Reference: [client-go watch timeout](https://github.com/kubernetes/client-go/blob/46360b527ebcd4cb7e99ae1770ad9a613bb0f128/tools/cache/reflector.go#L55)

## Expected Behavior

When a previously unavailable cluster (K8s2) recovers and its resources are cached:

1. Existing watch connections should be notified about the cluster addition
2. Resources from the recovered cluster should be sent as `ADDED` events
3. Deployment platform should see K8s2 resources immediately, not after 5-10 minutes

## Impact

- **Service Visibility**: During the 5-10 minute window, the deployment platform has incomplete resource views
- **Operational Risk**: Operators may make incorrect decisions based on outdated information
- **Poor User Experience**: After cluster recovery, users expect immediate visibility into all resources
- **Inconsistent Views Across Replicas**: When the deployment platform has multiple replicas, each replica's watch connection times out at different random times (due to the randomized 5-10 minute timeout). This causes different replicas to see the updated Pod list at different moments, resulting in inconsistent responses. Users may observe Pod lists changing back and forth between requests as their requests are load-balanced across different replicas - some showing the old state (pre-recovery) and others showing the new state (post-recovery). This inconsistency severely impacts user experience and makes it difficult to trust the platform's data.

XiShanYongYe-Chang:
**What type of PR is this?**

/kind bug

ref #6963 

Fixes #6963 

```release-note
`karmada-search`: Fixed the issue that watch connect cannot reflect resources from recovered clusters immediately.
```

### DIFF
--- a/pkg/search/proxy/store/multi_cluster_cache.go
+++ b/pkg/search/proxy/store/multi_cluster_cache.go
@@ -61,6 +61,11 @@ type MultiClusterCache struct {
 	restMapper          meta.RESTMapper
 	// newClientFunc returns a dynamic client for member cluster apiserver
 	newClientFunc func(string) (dynamic.Interface, error)
+
+	activeWatchersLock sync.RWMutex
+	// activeWatchers tracks all active watch connections for each GVR
+	// key: GVR, value: list of active watch multiplexers
+	activeWatchers map[schema.GroupVersionResource][]*watchMuxWithInvalidation
 }
 
 var _ Store = &MultiClusterCache{}
@@ -72,6 +77,7 @@ func NewMultiClusterCache(newClientFunc func(string) (dynamic.Interface, error),
 		newClientFunc:       newClientFunc,
 		cache:               map[string]*clusterCache{},
 		registeredResources: map[schema.GroupVersionResource]struct{}{},
+		activeWatchers:      map[schema.GroupVersionResource][]*watchMuxWithInvalidation{},
 	}
 }
 
@@ -116,12 +122,14 @@ func (c *MultiClusterCache) UpdateCache(resourcesByCluster map[string]map[schema
 	}
 
 	// add/update cluster cache
+	clustersAdded := false
 	for clusterName, resources := range resourcesByCluster {
 		cache, exist := c.cache[clusterName]
 		if !exist {
 			klog.Infof("Add cache for cluster %v", clusterName)
 			cache = newClusterCache(clusterName, c.clientForClusterFunc(clusterName), c.restMapper)
 			c.cache[clusterName] = cache
+			clustersAdded = true
 		}
 		err := cache.updateCache(resources)
 		if err != nil {
@@ -129,6 +137,20 @@ func (c *MultiClusterCache) UpdateCache(resourcesByCluster map[string]map[schema
 		}
 	}
 	c.registeredResources = registeredResources
+
+	// Only invalidate watches when clusters are added (not removed)
+	// Cluster removal is already handled by cacher.Stop() -> terminateAllWatchers()
+	if clustersAdded {
+		// When a new cluster joins or a failed cluster recovers, the watch connection cannot perceive the change.
+		// As a result, resources in the cluster cannot be watched, leading to data inconsistency.
+		// Therefore, we actively disconnect to wait the client to initiate a new connection.
+		// Here not altering the watch behavior but simply terminating the connection early.
+		// The client will automatically re-establish the watch request, thereby obtaining the
+		// complete cluster resource information, including the new or recovered cluster.
+		// For details, please refer to the issue: https://github.com/karmada-io/karmada/issues/6963
+		klog.Infof("Cluster topology changed, invalidating all active watches to trigger reconnection")
+		c.invalidateAllWatches()
+	}
 	return nil
 }
 
@@ -347,7 +369,7 @@ func (c *MultiClusterCache) Watch(ctx context.Context, gvr schema.GroupVersionRe
 		accessor.SetResourceVersion(responseResourceVersion.String())
 	}
 
-	mux := newWatchMux()
+	mux := newWatchMuxWithInvalidation()
 	clusters := c.getClusterNames()
 	for i := range clusters {
 		cluster := clusters[i]
@@ -367,6 +389,9 @@ func (c *MultiClusterCache) Watch(ctx context.Context, gvr schema.GroupVersionRe
 		})
 	}
 	mux.Start()
+
+	// Register this watch so we can invalidate it when cluster topology changes
+	c.registerWatch(gvr, mux)
 	return mux, nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/search/proxy/store/multi_cluster_cache.go
```diff
@@ -500,6 +523,62 @@ func (c *MultiClusterCache) getClusterResourceVersion(ctx context.Context, clust
...
+		for _, mux := range watchers {
+			// Send invalidation event asynchronously to avoid blocking
+			go mux.Invalidate()
+		}
+	}
+
+	if totalWatches > 0 {
+		klog.Infof("Sent invalidation signal to %d active watch connections", totalWatches)
+	}
+}
```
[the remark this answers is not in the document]
XiShanYongYe-Chang: Thanks, updated.

XiShanYongYe-Chang:
Hi @NickYadance, can you help take a review?

RainbowMango (commented):
/assign

NickYadance:
> Hi @NickYadance, can you help take a review?

Looks good to me, thx @XiShanYongYe-Chang

XiShanYongYe-Chang:
Thanks @NickYadance

pkg/search/proxy/store/multi_cluster_cache.go
```diff
@@ -61,6 +61,11 @@ type MultiClusterCache struct {
 	restMapper          meta.RESTMapper
 	// newClientFunc returns a dynamic client for member cluster apiserver
 	newClientFunc func(string) (dynamic.Interface, error)
+
+	// activeWatchers tracks all active watch connections for each GVR
+	// key: GVR string representation, value: list of active watch multiplexers
+	activeWatchersLock sync.RWMutex
+	activeWatchers     map[string][]*invalidatableWatchMux
```
RainbowMango: The `invalidatableWatchMux` naming is confusing, the structure itself doesn't indicate it's used for holding an invalidatable watch mux.

RainbowMango: In addition, why not take `schema.GroupVersionResource` as the map key? That would make the code more readable.

pkg/search/proxy/store/multi_cluster_cache.go
```diff
@@ -116,19 +122,33 @@ func (c *MultiClusterCache) UpdateCache(resourcesByCluster map[string]map[schema
 	}
 
 	// add/update cluster cache
+	clustersAdded := false
+	addedClusters := []string{}
```
RainbowMango:
```suggestion
```
Doesn't have to introduce a variable just for logging, and we already have the cluster name logged once a new cache is added.

pkg/search/proxy/store/multi_cluster_cache.go
```diff
@@ -116,19 +122,33 @@ func (c *MultiClusterCache) UpdateCache(resourcesByCluster map[string]map[schema
...
 	for clusterName, resources := range resourcesByCluster {
 		cache, exist := c.cache[clusterName]
 		if !exist {
 			klog.Infof("Add cache for cluster %v", clusterName)
 			cache = newClusterCache(clusterName, c.clientForClusterFunc(clusterName), c.restMapper)
 			c.cache[clusterName] = cache
+			// Any cluster being added to cache (whether new or recovered) should trigger invalidation
+			// This is critical for cluster recovery scenarios where existing watch connections
+			// don't include the recovered cluster's resources
```
RainbowMango: As far as I was told, the existing watch connection will eventually(~5 min) receive the recovered cluster's resources, so this comment might not be entirely accurate.

pkg/search/proxy/store/multi_cluster_cache.go
```diff
@@ -116,19 +122,33 @@ func (c *MultiClusterCache) UpdateCache(resourcesByCluster map[string]map[schema
...
 			return err
 		}
 	}
 	c.registeredResources = registeredResources
+
+	// Only invalidate watches when clusters are added (not removed)
+	// Cluster removal is already handled by cacher.Stop() -> terminateAllWatchers()
+	if clustersAdded {
+		klog.Infof("Cluster topology changed (clusters added: %v), invalidating all active watches to trigger reconnection", addedClusters)
+		c.invalidateAllWatches()
```
RainbowMango, line 152: What bothers me is that I don't know the impact of interrupting a client's connection. How significant is it for the client?
For instance, if a client has already received some data from a healthy cluster, will it receive only the subsequent updates after re-establishing the watch, or will it get the full dataset again?

RainbowMango (changes requested):

pkg/search/proxy/store/multi_cluster_cache.go
XiShanYongYe-Chang: Updated with `watchMuxWithInvalidation`, wdyt.

> In addition, why not take schema.GroupVersionResource as the map key? That would make the code more readable.

Good suggestion.

pkg/search/proxy/store/multi_cluster_cache.go
XiShanYongYe-Chang: This should depend on the ResourceVersion parameter used when making the watch request. For the Reflector implementation in client-go, the default behavior is to only receive incremental changes. https://deepwiki.com/search/hpascaletargetrefworkloadkinda_e78b9389-089b-4a74-83f7-7370dc9976cf

In addition, we actually haven't changed the behavior of watch reconnections. It is still handled the same way as it was before. The current behavior simply terminates the watch request earlier than the default timeout, rather than waiting for the server to do so.

XiShanYongYe-Chang:
cc @RainbowMango

pkg/search/proxy/store/multi_cluster_cache.go
RainbowMango: OK. Please explain it in the comments.

And please explain in comments why it is needed to terminate the existing connections when a new cluster joins.

RainbowMango (changes requested):
Generally looks good, just request more detail comments.

PS: Please rebase this before the next push.

pkg/search/proxy/store/multi_cluster_cache.go
XiShanYongYe-Chang: Done

XiShanYongYe-Chang:
Updated, thanks @RainbowMango, can you help take a review again?

RainbowMango (approved):
/lgtm
/approve

PR #7076, opened 2026-01-05
update cherry-pick issue template

zhzhuang-zju:
**What type of PR is this?**
/cleanup

The current cherry-pick process is already well-established, and the format of its issue title has largely stabilized; therefore, we can fix this format in the issue template. 

Additionally, the script `hack/cherry_pick_pull.sh` has been enhanced to automatically copy the release note to the new PR. After testing across multiple PRs, this feature has proven stable, so the manual copying instructions can be removed from the issue template.

Fixes #

```release-note
NONE
```

### DIFF
--- a/.github/ISSUE_TEMPLATE/cherry-pick.md
+++ b/.github/ISSUE_TEMPLATE/cherry-pick.md
@@ -1,6 +1,7 @@
 ---
 name: CherryPick Track
 about: Track tasks when release branches need cherry-pick.
+title: 'Backport #<PR_ID>'
 labels: help wanted
 ---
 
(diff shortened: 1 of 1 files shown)


zhzhuang-zju:
> Once we enable the [cherrypicker](https://github.com/kubernetes-sigs/prow/tree/main/cmd/external-plugins/cherrypicker) Prow external plugin, we may no longer need this template.

That’s a good direction. I have one question: if conflicts occur during the cherry-pick, will it resolve them automatically?

RainbowMango:
No idea, haven't investigated it yet.

RainbowMango:
@zhzhuang-zju Here is an example that shows that the robot won't send the PR if there are any conflicts.
https://github.com/kubernetes/autoscaler/pull/8974#issuecomment-3711882573

zhzhuang-zju:
Thanks! That makes sense. If there are conflicts, we'll need to resolve them manually.

Issue #7014, opened 2025-12-12
Enhance ClusterAffinity to support overflow cluster resource pools for hybrid cloud scheduling


vie-serendipity:
**What would you like to be added**:

Enhance old API or introduce a new API, like PreferredClusterAffinities, to implement priority scheduling across multiple affinities and clusters. This indicates that when scheduling, a cluster's priority should be considered first, not just its resources or weight. If the high-priority cluster lacks resources, then lower-priority clusters should be considered subsequently.

**Why is this needed**:

In multi-clusters scenarios, scheduling shouldn't solely consider cluster's resources. Multi-clusters might have different priorities, considering cloud provider, cost-efficiency, cluster location, etc. Especially in hybrid cloud scenarios, on-premises clusters should be prioritized over cloud-based clusters. The scheduler should fill on-premises clusters first and then utilize the cloud resources.

**User Story**

1. As a user, I have an on-premises cluster as well as a cloud-based cluster, and I use FHPA for autoscaling. I want my workloads to prioritize using my on-premises cluster, and only utilize the cloud cluster when resources are insufficient to manage costs effectively. It's crucial that during scale-in, the prioritization is maintained, with cloud-based replicas being removed first.
2. As a user, I aim for cluster disaster recovery. Currently, using ClusterAffinities, workloads don't migrate back to the cluster once it's restored.

zhzhuang-zju:
Thanks! I’ve added this to next week’s meeting agenda—maybe we can discuss it together.

RainbowMango:
Great, thanks for bringing this up. Love to help enhance this based on more and more users' feedback.

yaten2302:
Hi @RainbowMango , was something decided in the community call for this issue? I'm new here, and have already raised 2 PRs - https://github.com/karmada-io/karmada/pull/7012 & https://github.com/karmada-io/karmada/pull/6989 :)

I would love to help in this issue, if it's accepted 👍 
Thanks

zhzhuang-zju:
Hi @yaten2302 Thank you for your active participation. This feature is still in the design phase. Once all related designs are finalized, I’ll publish the corresponding tasks, and you’ll be able to claim the task that you are interested in.

yaten2302:
Hu @zhzhuang-zju , sure, once you publish the corresponding tasks here, I'll pick them up :)
[...]

PR #7078, opened 2026-01-05
proposal(scheduler): Extended Cluster Affinities for Sequential Overflow Scheduling

zhzhuang-zju:
**What type of PR is this?**
/kind documentation

Karmada currently supports declaring a set of candidate clusters through `clusterAffinity`, or multiple sets of candidate clusters through `ClusterAffinities` (which combines multiple `clusterAffinity` terms in a specific order). However, in either approach, each `clusterAffinity` represents an independent, mutually exclusive cluster set during a single scheduling process—the scheduler ultimately selects only one cluster group defined by one `clusterAffinity` or its subset.

This model has limitations in hybrid cloud scenarios (such as coexistence of local data centers and public clouds). In practical use, local clusters typically serve as the preferred resource pool, while public cloud clusters act as extensions or backup resources. The two are not completely independent and mutually exclusive relationships, but should be **automatically used supplementarily based on priority when local resources are insufficient**.

To address this, this proposal introduces cascading cluster affinity scheduling to describe priority relationships between cluster groups. This mechanism will enable Karmada to better support workload scheduling in hybrid cloud environments and improve the deployment practicality of online applications in terms of elasticity.

Parts of #7014

```release-note

```


zhzhuang-zju:
cc @vie-serendipity @RainbowMango @XiShanYongYe-Chang

docs/proposals/scheduling/multi-scheduling-group/overflow-affinities/README.md
```diff
@@ -0,0 +1,226 @@
...
+
+| Replica Scheduling Strategy | Question                                                                                                        | Conclusion                                                                                                                                                                                                          |
+|-----------------------------|-----------------------------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Duplicated                  | Should distribution occur in all clusters, or only in the first-tier?                                           | Only distribute in first-tier. Because the primary cluster group can meet scheduling requirements                                                                                                                   |
```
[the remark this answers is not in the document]
zhzhuang-zju: > the workload is duplicated across all available clusters

Disagree. In `Duplicated` strategy, cluster available resources are not taken into account—meaning the primary cluster can fully satisfy the scheduling requirements, so there's no need to schedule workloads onto the extended clusters.

zhzhuang-zju:
Updated the `implementation` part, see https://github.com/karmada-io/karmada/pull/7078/changes#diff-5ea1bb478c9b042fe6052f25d0bd870c66743aa9b1ca6ebc6f687e5f415fb3a7R189-R205

I've implemented a simple demo locally, and it works well. Here is the link: https://github.com/karmada-io/karmada/compare/master...zhzhuang-zju:karmada:poc-cascade. Feel free to give it a try if you're interested.

RainbowMango (commented):
/assign

docs/proposals/scheduling/multi-scheduling-group/inherited-cluster-affinities/README.md
```diff
@@ -0,0 +1,242 @@
...
+### Non-Goals
+
+- Cluster utilization is an effective means of controlling cluster resource usage but is not within the scope of this proposal.
+
+## Proposal
+
+### User Stories
+
+#### Story 1
+As a user, I have a local cluster and a cloud cluster, and use FHPA for automatic scaling. To effectively control costs, I want my workloads to preferentially use the local cluster, and only use the cloud cluster when resources are insufficient. Critically, during the scaling process, the priority must remain unchanged, and cloud replicas must be removed first.
```
RainbowMango: @zhzhuang-zju @vie-serendipity Can you elaborate on the use story? Explain what kind of workload, and why it is so critical. 

Here is a good sample: https://github.com/karmada-io/karmada/pull/6685/changes#diff-dd49cd7489615634a312daf2d3c62e6ccc894ffd3d81650a640f39c886e6af0fR45-R51

docs/proposals/scheduling/cascading-scheduling/README.md
```diff
@@ -0,0 +1,242 @@
...
+
+This proposal maintains backward compatibility. It will introduce entirely new APIs, so systems built with previous versions of Karmada can be seamlessly migrated to the new version. Previous configurations (YAMLs) can be applied to the new version of Karmada without any behavior change.
+
+## Design Details
+
+### API change
+
+#### Approach 1:
+
+Extend the `ClusterAffinity` API by adding a `Supplements` field to describe supplementary cluster group configurations. This field allows users to define one or more alternative cluster groups for a single Affinity Group. When the primary cluster group has insufficient resources or is unavailable, the scheduler can automatically cascade to these supplementary cluster groups for workload deployment. Note: Supplements is an array that can set multiple tiers of extensible cluster groups, with scheduling priority decreasing as the tier level increases.
```
RainbowMango: This approach also introduces a new field to the `.spec.placement.clusterAffinities`, is it intended? Need to clarify the meanings of the existing feature [multiple scheduling group](https://github.com/karmada-io/karmada/tree/master/docs/proposals/scheduling/multi-scheduling-group).

docs/proposals/scheduling/multi-scheduling-group/overflow-affinities/README.md
```diff
@@ -0,0 +1,242 @@
...
+    AffinityStrategy *AffinityStrategy `json:"affinityStrategy,omitempty"`  
+}  
+  
+type AffinityStrategy struct {  
+    // Mode defines the scheduling mode  
+    // +kubebuilder:validation:Enum=Exclusive;Cascade
+    // +kubebuilder:default=Exclusive
+    // +optional  
+    Mode string `json:"mode,omitempty"`  
+}
```
RainbowMango, line 249: I wonder if there is a case that expects both modes. 
Like, use `Cascade` for the normal case as the user story describes in this proposal, and another user story(described in https://github.com/karmada-io/karmada/tree/master/docs/proposals/scheduling/multi-scheduling-group#as-a-user-i-want-to-deploy-applications-on-the-primary-cluster-and-leave-a-backup-cluster-for-disaster-recovery-cases) that is for backup purposes.

docs/proposals/scheduling/multi-scheduling-group/overflow-affinities/README.md
```diff
@@ -0,0 +1,242 @@
...
+          - cluster2
+          - cluster3
+    affinityStrategy:
+      mode: Cascade
+```
+
+#### Approach 3:
+
+Add a new API `PreferredClusterAffinities` at the same level as ClusterAffinities to declare cluster groups with priority relationships.
+
```
RainbowMango, line 282: We can't just introduce another field if the existing field can not cover some particular cases, except that the new field would deprecate the old one.

RainbowMango (commented):
I think this feature is essentially the [Multiple scheduling group](https://github.com/karmada-io/karmada/tree/master/docs/proposals/scheduling/multi-scheduling-group#as-a-user-i-want-to-deploy-applications-on-the-primary-cluster-and-leave-a-backup-cluster-for-disaster-recovery-cases), but extending the feature with more use cases. So, we have to take care of the existing feature; I don't mind extending it or even deprecating it.

The feature name could be `Inherited Cluster Affinities`. 
The feature description could be:
```
Extended the Multiple Scheduling Groups feature to support inherited 
cluster selection across affinity groups. When enabled, subsequent affinity 
groups automatically inherit cluster sets from previous groups, enabling 
progressive cluster expansion strategies.
```

docs/proposals/scheduling/multi-scheduling-group/overflow-affinities/README.md
zhzhuang-zju: Agree, we are very unlikely to go with this approach

docs/proposals/scheduling/cascading-scheduling/README.md
zhzhuang-zju: Thanks for the reminder~
Updated: https://github.com/karmada-io/karmada/pull/7078/changes#diff-0f8ee8822d634442d8951a9ba5f61fa7b55a5bdbf910cc947b4961efd77d0ad1R87-R89

docs/proposals/scheduling/multi-scheduling-group/inherited-cluster-affinities/README.md
zhzhuang-zju: Thanks for your suggestion. 
@vie-serendipity Could you help update the user story? Just leave a comment here—I'll sync it into the proposal.

docs/proposals/scheduling/multi-scheduling-group/overflow-affinities/README.md
zhzhuang-zju: For backup scenarios, would using `Exclusive` mode be sufficient?

zhzhuang-zju:
> I think this feature is essentially the [Multiple scheduling group](https://github.com/karmada-io/karmada/tree/master/docs/proposals/scheduling/multi-scheduling-group#as-a-user-i-want-to-deploy-applications-on-the-primary-cluster-and-leave-a-backup-cluster-for-disaster-recovery-cases), but extending the feature with more use cases. So, we have to take care of the existing feature; I don't mind extending it or even deprecating it.

Totally—this feature should serve as an extension to existing capabilities. I followed this principle when designing the API as well.

> The feature name could be Inherited Cluster Affinities.

I asked an AI, and it offered several good suggestions—I think we can consider them together.
- Inherited Cluster Affinities
- Tiered Cluster Affinities
- Progressive Cluster Affinities
- Additive Cluster Affinities

docs/proposals/scheduling/multi-scheduling-group/inherited-cluster-affinities/README.md
vie-serendipity: My suggestions are as follows; feel free to adopt or edit them.


As a user running GPU-based inference workloads in both my own IDC GPU cluster and a cloud GPU cluster, I need to keep GPU costs low so I only want to use cloud GPUs for extra peak traffic instead of baseline load. When traffic changes and FHPA scales my workloads, Karmada should always scale out in order from IDC to cloud and scale in in reverse order from cloud back to IDC, so that my GPU usage remains cost‑efficient.

zhzhuang-zju: Thanks @vie-serendipity 
Updated!

RainbowMango: Thanks.

docs/proposals/scheduling/multi-scheduling-group/overflow-affinities/README.md
```diff
@@ -0,0 +1,248 @@
...
+        - cluster1
+      supplements:
+        - clusterNames:
+          - cluster2
+          - cluster3
+```
+
+Since `ClusterAffinities` is a combination of multiple `ClusterAffinity` terms, it will also support using the `Supplements` field to declare tiered expansion relationships. In this case, the semantics are:
+- Each `ClusterAffinity` within `ClusterAffinities` can declare its own `Supplements` field, representing the extended cluster groups for that specific affinity group.
+- The relationship between different `ClusterAffinity` terms in `ClusterAffinities` remains mutually exclusive. The next `ClusterAffinity` will only be considered when both the base clusters and the extended clusters of the previous `ClusterAffinity` fail to satisfy the scheduling requirements.
```
RainbowMango, line 224: Just come up with another idea, we may introduce the `Supplements(place-holder name)`  field to `ClusterAffinityTerm`. 
That would be a clear extension of the existing feature. 
```yaml
  placement:
    clusterAffinities:
      - affinityName: "cost-sensitive"
        clusterNames:
          - local-idc-1
          - local-idc-2
        supplements:
          - clusterNames:
            - cloud-a-01
            - cloud-a-02
          - clusterNames:
            - cloud-b-01
            - cloud-b-02
```
What do you think?

zhzhuang-zju: This is a good improvement over `Approach 1`. It makes the API structure much clearer, whereas the previous version introduced a recursive `Supplements` field that was relatively complex. 

My only concern is that if a user only has one Affinity Group (primary + supplements), using `placement.clusterAffinity` might feel more natural. Of course, this is just my personal intuition.

zhzhuang-zju: Thanks, updated your idea as Approach 2

docs/proposals/scheduling/multi-scheduling-group/inherited-cluster-affinities/README.md
```diff
@@ -0,0 +1,290 @@
...
+	// AffinityName is the name of the cluster group.
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=32
+	// +required
+	AffinityName string `json:"affinityName"`
+
+	ClusterAffinity `json:",inline"`
+
+	// new added API field
+	Supplements []ClusterAffinity `json:"supplements,omitempty"`
```
RainbowMango:
```suggestion
// Supplements specifies additional cluster groups to supplement the primary
// cluster group when it does not have sufficient resources.
// The scheduler will combine clusters from both the primary group and 
// supplement groups for evaluation.
// +optional
Supplements []ClusterAffinity `json:"supplements,omitempty"`
```

Regarding the API names, I tend to the current `Supplements` over the following alternatives:
- Fallback 
- Secondary
- Backup
- Alternative
- Reserve

RainbowMango: @vie-serendipity I'd like to hear your thoughts about the API design, does the `approach 2` make sense to you? or any other comments?

vie-serendipity: The API name doesn't seem to be easy to understand directly. In addition, this introduces another level, which seems kind of complex.

RainbowMango: Which approach do you prefer? Or any other suggestion?

vie-serendipity: Actually I prefer approach 3. But it indeed cannot meet scenarios that require both modes.

docs/proposals/scheduling/multi-scheduling-group/overflow-affinities/README.md
```diff
@@ -0,0 +1,290 @@
...
+### Non-Goals
+
+- Cluster utilization is an effective means of controlling cluster resource usage but is not within the scope of this proposal.
+
+## Proposal
+
+### User Stories
+
+#### Story 1
+As a user running GPU-based inference workloads in both my own IDC GPU cluster and a cloud GPU cluster, I need to keep GPU costs low so I only want to use cloud GPUs for extra peak traffic instead of baseline load. When traffic changes and FHPA scales my workloads, Karmada should always scale out in order from IDC to cloud and scale in in reverse order from cloud back to IDC, so that my GPU usage remains cost‑efficient.
```
zhzhuang-zju, line 43: @zhengjr9 Hi, I noticed that you are also using GPU clusters and employing CronFHPA to scale your workloads. Do you have similar scenarios or requirements?

docs/proposals/scheduling/multi-scheduling-group/inherited-cluster-affinities/README.md
zhzhuang-zju: Thank you @RainbowMango and @vie-serendipity for the discussion. I agree that `supplements` could be confusing. Perhaps we could use a name that describes both the function and the structure, such as 
- `backupTier`
- `secondaryTier`. 

Using the word 'Tier' might better convey the hierarchical nature. Does this sound better?

zhzhuang-zju: I have discussed the API field with @vie-serendipity, and he also supports using `Tier` to emphasize its hierarchical nature. WDYT？ @RainbowMango

RainbowMango: I'll give it another look this week.

RainbowMango: Inspired by the `Tier`, I came up with some more alternative, here are the two I like most:

**First One--with AffinityTiers``**
```golang
// ClusterAffinityTerm selects a set of clusters.
type ClusterAffinityTerm struct {
    // AffinityName is the name of the cluster group.
    // +required
    AffinityName string `json:"affinityName"`

    ClusterAffinity `json:",inline"`

    // AffinityTiers represents an ordered list of cluster affinity tiers
    // within this affinity term. Each tier defines a group of candidate clusters.
    //
    // When the scheduler evaluates this term, it starts with the first tier.
    // If the clusters in the first tier have insufficient resources, the scheduler
    // progressively expands the candidate pool by including clusters from the
    // next tier, and so on until resource requirements are met.
    //
    // During scale-down, the scheduler shrinks in reverse order: clusters in
    // lower-priority tiers (later in the list) are released first.
    //
    // This is particularly useful for hybrid cloud scenarios where on-premise
    // clusters are preferred, and cloud clusters serve as overflow capacity.
    //
    // Note: AffinityTiers cannot co-exist with the inline ClusterAffinity fields.
    // +optional
    AffinityTiers []ClusterAffinityTier `json:"affinityTiers,omitempty"`
}

// ClusterAffinityTier represents a prioritized group of clusters within
// a ClusterAffinityTerm. Tiers are evaluated in order: earlier tiers have
// higher priority.
type ClusterAffinityTier struct {
    // Name is the name of this cluster affinity tier, used for identification
    // and observability.
    // +kubebuilder:validation:MinLength=1
    // +kubebuilder:validation:MaxLength=32
    // +required
    Name string `json:"name"`

    ClusterAffinity `json:",inline"`
}
```
An example of this design as follows:
```yaml
placement:
  clusterAffinities:
    - affinityName: hybrid-scheduling
      affinityTiers:
        - name: on-premise
          clusterNames: [local-1, local-2]
        - name: cloud-aws
          labelSelector:
            matchLabels:
              provider: aws
        - name: cloud-gcp
          labelSelector:
            matchLabels:
              provider: gcp
```

**Second One--with OverflowAffinities``**

```golang
type ClusterAffinityTerm struct {
    AffinityName string `json:"affinityName"`
    ClusterAffinity `json:",inline"`

    // OverflowAffinities defines additional cluster groups that the scheduler
    // can progressively include when the primary group (defined by the inline
    // ClusterAffinity) has insufficient resources. Groups are expanded in order
    // and contracted in reverse during scale-down.
    // Can only be used together with the inline ClusterAffinity (the inline
    // ClusterAffinity serves as the primary/preferred group).
    // +optional
    OverflowAffinities []OverflowClusterAffinity `json:"overflowAffinities,omitempty"`
}

type OverflowClusterAffinity struct {
    // Name identifies this overflow cluster group.
    // +required
    Name string `json:"name"`
    ClusterAffinity `json:",inline"`
}
```
An example of this would be:
```yaml
placement:
  clusterAffinities:
    - affinityName: hybrid-scheduling
      clusterNames: [local-1, local-2]
      overflowAffinities:
        - name: cloud-aws
          labelSelector:
            matchLabels:
              provider: aws
        - name: cloud-gcp
          labelSelector:
            matchLabels:
              provider: gcp
```

@zhzhuang-zju @vie-serendipity let me know which is your preference, or both are ok to you, or both are NOK.

zhzhuang-zju: > First One--with OverflowAffinities

I like this one

vie-serendipity: I think these two names are quite good, better than before.

I prefer overflowAffinities, the field name itself includes the explanation.

RainbowMango: OK. @zhzhuang-zju, please go with the `overflowAffinities` and update the proposal accordingly.

zhzhuang-zju: Great! Will kick it off soon

docs/proposals/scheduling/multi-scheduling-group/overflow-affinities/README.md
```diff
@@ -0,0 +1,303 @@
...
+
+### Risks and Mitigations
+
+This proposal maintains backward compatibility. It extends existing APIs by adding optional fields to support overflow cluster affinities, so systems built with previous versions of Karmada can be seamlessly migrated to the new version. Previous configurations (YAMLs) can be applied to the new version of Karmada without any behavior change.
+
+## Design Details
+
+### API change
+
+#### Approach 1:
```
RainbowMango: Please describe the approach we reached consensus on, and put the others in the alternatives section.

RainbowMango:
@zhzhuang-zju Please help update the proposal targeting for finnal version. And prioritize it as this should be included in the v1.18.

docs/proposals/scheduling/multi-scheduling-group/overflow-affinities/README.md
```diff
@@ -0,0 +1,307 @@
...
+
+During the scheduling process, the scheduler will first try to schedule the workload to member cluster `cluster1` as much as possible. If `cluster1` cannot accommodate all replicas, the remaining replicas will continue to be allocated among clusters `cluster2` and `cluster3`.
+
+If `cluster1` is unavailable, the scheduler will use the clusters in the extended cluster group (i.e., `cluster2` and `cluster3`) for scheduling. Note that when `cluster1` becomes available again, **workload migration will not be automatically triggered**; users need to explicitly use the `WorkloadRebalancer` resource to trigger rescheduling, or indirectly trigger it by adjusting workload replica counts.
+
+Each scheduling will **preferentially use the higher priority affinity group** under current conditions, i.e., the group with the earlier order in `ClusterAffinityTerm`.
+
+The overall process can be described as the following diagram:
+
+![overflow-cluster-affinity-scheduling](statics/overflow-cluster-affinity-scheduling.drawio.png)
```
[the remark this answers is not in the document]
zhzhuang-zju: The pr includes 3 files, and the image file is `statics/overflow-cluster-affinity-scheduling.drawio.png`

zhzhuang-zju:
> Please help update the proposal targeting for finnal version. And prioritize it as this should be included in the v1.18.

Done~ please take a look

docs/proposals/scheduling/multi-scheduling-group/overflow-affinities/README.md
```diff
@@ -0,0 +1,310 @@
...
+	// occurrences of the same cluster are ignored.
+	// +optional
+	OverflowAffinities []OverflowClusterAffinity `json:"overflowAffinities,omitempty"`
+}
+
+// OverflowClusterAffinity represents an overflow tier of candidate clusters.
+type OverflowClusterAffinity struct {
+	// AffinityName is the name of the cluster group.
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=32Expand commentComment on line R700Resolved
```
RainbowMango, line 86:
```suggestion
	// +kubebuilder:validation:MaxLength=32
```

docs/proposals/scheduling/multi-scheduling-group/overflow-affinities/README.md
```diff
@@ -0,0 +1,310 @@
...
+        clusterNames:
+          - cluster1
+        overflowAffinities:
+          - affinityName: "public-cloud"
+            clusterNames:
+              - cluster2
+              - cluster3
+```
+
+During the scheduling process, the scheduler will first try to schedule the workload to member cluster `cluster1` as much as possible. If `cluster1` cannot accommodate all replicas, the remaining replicas will continue to be allocated among clusters `cluster2` and `cluster3`.
```
RainbowMango, line 117: This sentence is a detailed description of the examples above, but it might bring misunderstandings:
> try to schedule `workload` to member cluster cluster1 `as much as possible`

try to schedule `replicas` of the Deployment, not the workload.

In addition, the behavior of the example yaml not for splitting replicas, but for duplicating. Please provide a real example here, not just for demonstrating the API usage.

docs/proposals/scheduling/multi-scheduling-group/overflow-affinities/README.md
```diff
@@ -0,0 +1,310 @@
...
+
+If `cluster1` is unavailable, the scheduler will use the clusters in the extended cluster group (i.e., `cluster2` and `cluster3`) for scheduling. Note that when `cluster1` becomes available again, **workload migration will not be automatically triggered**; users need to explicitly use the `WorkloadRebalancer` resource to trigger rescheduling, or indirectly trigger it by adjusting workload replica counts.
```
RainbowMango, line 119:
```suggestion
If `cluster1` is unavailable, the scheduler will use the clusters in the extended cluster group (i.e., `cluster2` and `cluster3`) for scheduling. Note that when `cluster1` becomes available again, **workload migration will not be automatically triggered to avoid replica flapping**; users need to explicitly use the `WorkloadRebalancer` resource to trigger rescheduling, or indirectly trigger it by adjusting workload replica counts.
```

docs/proposals/scheduling/multi-scheduling-group/overflow-affinities/README.md
```diff
@@ -0,0 +1,310 @@
...
+
+Each scheduling will **preferentially use the higher priority affinity group** under current conditions, i.e., the group with the earlier order in `ClusterAffinityTerm`.
+
+The overall process can be described as the following diagram:
+
+![overflow-cluster-affinity-scheduling](statics/overflow-cluster-affinity-scheduling.drawio.png)
+
+#### Alternatives
```
RainbowMango, line 127: 1. Put the alternatives at the end of this proposal
2. We should explain why the alternative was not acceptable.

https://github.com/karmada-io/karmada/blob/e918fb1ed53670e1827e7e6a5ff239ed017544cc/docs/proposals/proposal-template/proposal-template.md?plain=1#L123-L129

docs/proposals/scheduling/multi-scheduling-group/overflow-affinities/README.md
```diff
@@ -0,0 +1,310 @@
...
+3. `Score` Stage: The `ClusterAffinity` plugin needs to be adjusted to support Overflow cluster affinities. Specifically, it should assign different base scores to clusters based on
+the tier of the affinity group they belong to (e.g., 5000 for the first tier, 4000 for the second, and so on). This is to ensure that clusters in higher-priority groups are preferred during the subsequent cluster selection phase.
+   Note: The current scheduler framework constrains Score plugin return values to `[MinClusterScore(0), MaxClusterScore(100)]`. This narrow range is insufficient for overflow scheduling—a mere 0‑100 gap between tiers can easily be overshadowed by scores from other Score plugins, breaking the strict tier‑priority guarantee. Therefore, the `ClusterAffinity` plugin will intentionally return scores beyond `MaxClusterScore` in Overflow mode (e.g., 5000/4000/…). The large inter‑tier gap ensures that no combination of other plugin scores can cause a lower‑priority tier cluster to outrank a higher‑priority tier cluster.
```
RainbowMango, line 270: I don't think adjusting the score range is a good idea; this way essentially disables the other score plugins.  

It deserves to be discussed a little bit in the community meeting if there is no better solution.

docs/proposals/scheduling/multi-scheduling-group/overflow-affinities/README.md
```diff
@@ -0,0 +1,310 @@
...
+4. `Select` Stage: No changes are needed.
+5. `Assign` Stage: This stage needs to be adjusted to support Overflow cluster affinities. Specifically:
+   5.1 Group clusters based on their affinity group tier.
+   5.2 Iterate through the tiers in order, attempting to assign the remaining replicas. Clusters in each tier will be allocated as many replicas as possible until the
+   resources for that tier are exhausted or all replicas have been assigned.
+   5.3 Within each tier, the assignment logic will remain consistent with the current implementation.
+   5.4 The process concludes successfully once all replicas are assigned. If unassigned replicas remain after iterating through all tiers, the scheduling process fails.
```
RainbowMango, line 277:
```suggestion
   - 5.1 Group clusters based on their affinity group tier.
   - 5.2 Iterate through the tiers in order, attempting to assign the remaining replicas. Clusters in each tier will be allocated as many replicas as possible until the
   resources for that tier are exhausted or all replicas have been assigned.
   - 5.3 Within each tier, the assignment logic will remain consistent with the current implementation.
   - 5.4 The process concludes successfully once all replicas are assigned. If unassigned replicas remain after iterating through all tiers, the scheduling process fails.
```
format

docs/proposals/scheduling/multi-scheduling-group/overflow-affinities/README.md
```diff
@@ -0,0 +1,310 @@
...
+#### karmada-webhook
+
+Adding a validation for the `(Cluster)PropagationPolicy` resource to restrict that `spec.placement.clusterAffinities[i].overflowAffinities` can only be used together with the inline `ClusterAffinity` and must not be set independently.
+
+### Test Plan
+
+- All current testing should be passed, no breaking change would be involved by this feature.
+- Add new E2E tests to cover the feature, the scope should include:
+    * Workload propagating with scheduling type `Duplicated`.
+    * Workload propagating with scheduling type `Divided` (including scaling scenarios)
```
RainbowMango, line 288: A typical E2E test design should include expected behavior. Please help design more test cases in detail. Nowadays, AI can help to design the tests; all you need to feed the proposal and pick the tests.

docs/proposals/scheduling/multi-scheduling-group/overflow-affinities/README.md
```diff
@@ -0,0 +1,310 @@
...
+
+### Discussion Points
+
+**Should multi-cluster priority scheduling support working with `SpreadConstraints`?**
+
+Yes, they are independent APIs.
+
+**How does Overflow cluster affinities scheduling behave under different replica scheduling strategies?**
+
+The overflow scheduling logic is designed to be compatible with both `Duplicated` and `Divided` scheduling types. The key principle is that the scheduler will always attempt to fill replicas into the highest-priority tier first, regardless of the scheduling type.
```
RainbowMango, line 298: Do you mean, with `Duplicated`, the scheduler will not take the clusters from `OverflowAffinities`?

I remember that the scheduler will not check available resources with `Duplicated` mode (`CURRENTLY`), and all inference workload would use `Divided` mode. My question here is, why do we need to support `Duplicated`?

RainbowMango (commented):
Just one concern regarding the range of score plugin in the scheduler.

docs/proposals/scheduling/multi-scheduling-group/overflow-affinities/README.md
zhzhuang-zju: Thanks for pointing this out. I have no idea where this extra code came from

docs/proposals/scheduling/multi-scheduling-group/overflow-affinities/README.md
zhzhuang-zju: You are right. It is not a good idea to use `NormalizeScore` to raise the plugin score of `ClusterAffinity` above `MaxClusterScore`.

I checked the definition and usage of `NormalizeScore` in kube-scheduler. It is used to normalize the raw scores calculated by plugins for each node to a unified range [0, MaxNodeScore], ensuring that scores from different plugins are comparable.

https://github.com/kubernetes/kubernetes/blob/03779bbd00da25c7fcd03711ddfe466a3322e1d7/staging/src/k8s.io/kube-scheduler/framework/interface.go#L606-L611
https://github.com/kubernetes/kubernetes/blob/03779bbd00da25c7fcd03711ddfe466a3322e1d7/pkg/scheduler/framework/plugins/interpodaffinity/scoring.go#L257-L291

Therefore, I have changed to perform cluster sorting during the select phase.

zhzhuang-zju:
@RainbowMango Updated, please take a look

docs/proposals/scheduling/multi-scheduling-group/overflow-affinities/README.md
```diff
@@ -0,0 +1,320 @@
...
+
+#### karmada-scheduler
+
+The current scheduler only processes primary cluster group within `ClusterAffinityTerm`. So, it needs to adjust the scheduling logic to support Overflow cluster affinities scheduling.
+
+The details are as follows:
+1. The `schedule(Cluster)ResourceBindingWithClusterAffinities` scheduling entrypoint: No changes are needed.
+2. `Filter` Stage: The `ClusterAffinity` plugin needs to be adjusted to support Overflow cluster affinities. Specifically, a cluster passes the filter if it satisfies the scheduling conditions of any affinity group within a `ClusterAffinityTerm`.
+3. `Score` Stage: No changes are needed.
+4. `Select` Stage: The current `Select` stage prioritizes clusters with the highest scores. For overflow affinities scheduling, it needs to be adjusted to prioritize clusters belonging to the primary group and earlier overflow tiers over those in later tiers. This ensures that clusters are selected in strict tier order, so that replicas overflow to lower-priority tiers only after higher-priority tiers are fully utilized.
```
RainbowMango, line 142: I don't want to block this proposal before this, but I do want to double-check how to prioritize the clusters when reviewing the implementations later.
Another approach in my mind(and also talked to @zhzhuang-zju ) is extending to [ScoreList](https://github.com/karmada-io/karmada/blob/56157b6c4dca580a47c6b6640846b764fd2ee87a/pkg/scheduler/framework/interface.go#L240-L241) to allow it to hold the tiers.  Not sure which is better now.

RainbowMango (approved):
/lgtm
/approve

PR #7103, opened 2026-01-12
feat: avoid redundant process because of claim metadata change

vie-serendipity:
**What type of PR is this?**

/kind feature

This PR solves a problem where, when claim metadata is added to an object, the object should be processed next time because the update event will enqueue it again.

Every object will be handled by detector twice at least even though nothing changed. And accordingly every rb will be handled twice by rb controller. This should be optimized.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/detector/detector.go
+++ b/pkg/detector/detector.go
@@ -457,10 +457,15 @@ func (d *ResourceDetector) ApplyPolicy(object *unstructured.Unstructured, object
 		}
 	}()
 
-	policyID, err := d.ClaimPolicyForObject(object, policy)
+	policyID, claimed, err := d.ClaimPolicyForObject(object, policy)
 	if err != nil {
 		klog.Errorf("Failed to claim policy(%s/%s) for object: %s", policy.Namespace, policy.Name, object)
 		return err
+	}
+
+	if claimed {
+		klog.V(4).Infof("Claim metadata added to object(%s) for policy(%s/%s), object will be automatically requeued via update event", objectKey, policy.Namespace, policy.Name)
+		return nil
 	}
 
 	// If this Reconcile action is triggered by Karmada itself and the current bound Policy is lazy activation preference,
@@ -545,10 +550,15 @@ func (d *ResourceDetector) ApplyClusterPolicy(object *unstructured.Unstructured,
 		}
 	}()
 
-	policyID, err := d.ClaimClusterPolicyForObject(object, policy)
+	policyID, claimed, err := d.ClaimClusterPolicyForObject(object, policy)
 	if err != nil {
 		klog.Errorf("Failed to claim cluster policy(%s) for object: %s", policy.Name, object)
 		return err
+	}
+
+	if claimed {
+		klog.V(4).Infof("Claim metadata added to object(%s) for cluster policy(%s), object will be automatically requeued via update event", objectKey, policy.Name)
+		return nil
 	}
 
 	// If this Reconcile action is triggered by Karmada itself and the current bound Policy is lazy activation preference,
@@ -762,7 +772,7 @@ func (d *ResourceDetector) fetchClusterResourceBinding(ctx context.Context, crbN
 
 // ClaimPolicyForObject set policy identifier which the object associated with.
 // It will add policy labels and annotations to the object, and update the obj with the content returned by the Server, ensuring that the obj used when generating the binding is the latest.
-func (d *ResourceDetector) ClaimPolicyForObject(object *unstructured.Unstructured, policy *policyv1alpha1.PropagationPolicy) (string, error) {
+func (d *ResourceDetector) ClaimPolicyForObject(object *unstructured.Unstructured, policy *policyv1alpha1.PropagationPolicy) (string, bool, error) {
 	policyID := policy.Labels[policyv1alpha1.PropagationPolicyPermanentIDLabel]
 
 	objLabels := object.GetLabels()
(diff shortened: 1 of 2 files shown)


RainbowMango (commented):
/assign
Nice catch! @vie-serendipity . Could you please look at the failing tests?

vie-serendipity:
@RainbowMango 
When an object is selected by pp, it will be processed. The detector will label and annotate this object, after which the object is processed one more time.

I want it to be processed only once, so I proposed this PR. Only the objects with karmada labels and annotations will be processed. 

But in the scenario where ActivationPreference is lazy, if the first time it is not processed and resourceChangedByKarmada will be true on the second time, it will not be processed, so rb will not even be created. This is the reason why e2e test failed.

If I want to achieve my goal right now, I can actually set ActivationPreference to Lazy. Right?

RainbowMango:
> But in the scenario where ActivationPreference is lazy, if the first time it is not processed and resourceChangedByKarmada will be true on the second time, it will not be processed, so rb will not even be created. This is the reason why e2e test failed.

Do you mean which E2E test?
Is this one: https://github.com/karmada-io/karmada/actions/runs/20917907135/job/60096547228?pr=7103#step:6:197
```
• [FAILED] [420.054 seconds]
Karmadactl logs testing Test karmadactl logs for existing pod [BeforeEach] should get logs from a specific container in the pod
  [BeforeEach] /home/runner/work/karmada/karmada/test/e2e/suites/base/karmadactl_test.go:642
  [It] /home/runner/work/karmada/karmada/test/e2e/suites/base/karmadactl_test.go:670

  Captured StdOut/StdErr Output >>
  I0112 11:55:36.214755   56846 pod.go:60] Waiting for pod(karmadatest-n8gbz/pod-btv75) synced on cluster(member1)
  << Captured StdOut/StdErr Output

  [FAILED] Timed out after 420.000s.
  Expected
      <bool>: false
  to equal
      <bool>: true
  In [BeforeEach] at: /home/runner/work/karmada/karmada/test/e2e/framework/pod.go:68 @ 01/12/26 12:02:36.215

  Full Stack Trace
    github.com/karmada-io/karmada/test/e2e/framework.WaitPodPresentOnClusterFitWith({0xc000117579, 0x7}, {0xc000c0e0f0, 0x11}, {0xc000b2c0b7, 0x9}, 0x5654f08)
    	/home/runner/work/karmada/karmada/test/e2e/framework/pod.go:68 +0x634
    github.com/karmada-io/karmada/test/e2e/suites/base.init.func2.2({0xc000c0e0f0, 0x11}, {0xc000b2c0b7, 0x9})
    	/home/runner/work/karmada/karmada/test/e2e/suites/base/karmadactl_test.go:630 +0x6b
    github.com/karmada-io/karmada/test/e2e/suites/base.init.func2.3.1()
    	/home/runner/work/karmada/karmada/test/e2e/suites/base/karmadactl_test.go:655 +0xab7
```

> If I want to achieve my goal right now, I can actually set ActivationPreference to Lazy. Right?

No, I don't think so. 
1. `ActivationPreference` was not designed for this case
2. The `ActivationPreference` is an experimental feature, which might be removed in the future.

Please refer to the side effects from the `Warning` section of https://karmada.io/docs/next/userguide/scheduling/propagation-policy#activationpreference.

vie-serendipity:
No, I mean this one, which is about lazy_activation_policy_test.
https://github.com/karmada-io/karmada/actions/runs/21019184471/job/60430861139?pr=7103#step:6:2069

For karmadactl_test you refer to, I think that would be related to githu workflow environment. Because this time it passed.

RainbowMango:
I see, will look at it.

RainbowMango:
> But in the scenario where ActivationPreference is lazy, if the first time it is not processed and resourceChangedByKarmada will be true on the second time, it will not be processed, so rb will not even be created. This is the reason why e2e test failed.

I see. 
It is a tricky case and I don't have a clue about it so far. 

Maybe, the claimed operation should be ignored(skip reconciliation), no matter considering `ActivationPreference`.

zhzhuang-zju:
This PR seems to be stuck—I'll take a look.

zhzhuang-zju:
There are generally two approaches to skip redundant processing:

- **Terminate the current reconcile and defer the remaining steps to the next reconcile cycle.**  
  This is the approach adopted in this PR. However, it is currently affected by the lazy `ActivationPreference` mechanism and may fail in certain scenarios. I don't recommend modifying code related to lazy `ActivationPreference`, as it's difficult to assess its true impact.

- **Complete the current reconcile fully and skip the next one.**  
  Specifically, if only the `claimedMetadata` of the object has changed, skip the current reconcile. Currently, `eventfilter.SpecificationChanged` already filters out many changes that aren't related to the specification. How about extending `eventfilter.SpecificationChanged` to skip redundant reconciles triggered by changes to `claimedMetadata`? Note that `eventfilter.SpecificationChanged` is also used by other controllers, so the impact needs further evaluation.

RainbowMango:
> Specifically, if only the claimedMetadata of the object has changed, skip the current reconcile. Currently, eventfilter.SpecificationChanged already filters out many changes that aren't related to the specification. How about extending eventfilter.SpecificationChanged to skip redundant reconciles triggered by changes to claimedMetadata? Note that eventfilter.SpecificationChanged is also used by other controllers, so the impact needs further evaluation.

This approach is worth exploring further. Just a reminder, we have some logic where relying on the behavior, I mean, once a PropagationPolicy is removed, the detector will cleanup the claimed metadata from the resource template and expect a following reconcile to match another PropagationPolicy.

zhzhuang-zju:
Thanks for your reminder—this is indeed an issue. It could likely be resolved by distinguishing whether the operation is specifically adding `claimMetadata`. However, we still need to conduct a systematic analysis to avoid introducing unintended side effects.

vie-serendipity:
> It could likely be resolved by distinguishing whether the operation is specifically adding claimMetadata.

If there are no further changes to resources, the member cluster resources will lack claim metadata. I think we should maintain consistency; in addition, in some cases the claim metadata on resources is also helpful for troubleshooting.

RainbowMango:
I'm looking at issues tagged with milestone v1.17. And added an agenda to [today's meeting](https://docs.google.com/document/d/1BUZBDvtbtfU9MmOz7KVdvGiiCny-zz_34FuLlW4AGmM/edit?tab=t.0#heading=h.8d2qii1mb7w8), hopefully we can talk about it there.

PR #7104, opened 2026-01-13
feat: informer should sync cache before other controllers start

vie-serendipity:
**What type of PR is this?**

/kind feature

The cluster informer manager should sync the cache first to avoid directly fetching from the apiserver. This means detectors and other controllers have dependencies, so detectors should be placed in the controller-runtime cache case.

Avoid starting detectors and other controllers in a parallel way that causes the cache to be unavailable and other controllers directly query the apiserver.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/detector/detector.go
+++ b/pkg/detector/detector.go
@@ -194,6 +194,7 @@ func (d *ResourceDetector) discoverResources(ctx context.Context, period time.Du
 			d.InformerManager.ForResource(r, d.EventHandler)
 		}
 		d.InformerManager.Start()
+		d.InformerManager.WaitForCacheSync()
 	}, period, ctx.Done())
 }
 
--- a/pkg/detector/policy_test.go
+++ b/pkg/detector/policy_test.go
@@ -155,6 +155,9 @@ func Test_cleanPPUnmatchedRBs(t *testing.T) {
 			defer cancel()
 			fakeDynamicClient := dynamicfake.NewSimpleDynamicClient(scheme, tt.existingObject)
 			genMgr := genericmanager.NewSingleClusterInformerManager(ctx, fakeDynamicClient, 0)
+			genMgr.Lister(appsv1.SchemeGroupVersion.WithResource("deployments"))
+			genMgr.Start()
+			genMgr.WaitForCacheSync()
 			resourceDetector := &ResourceDetector{
 				Client:          fakeClient,
 				DynamicClient:   fakeDynamicClient,
@@ -290,6 +293,9 @@ func Test_cleanUnmatchedRBs(t *testing.T) {
 			defer cancel()
 			fakeDynamicClient := dynamicfake.NewSimpleDynamicClient(scheme, tt.existingObject)
 			genMgr := genericmanager.NewSingleClusterInformerManager(ctx, fakeDynamicClient, 0)
+			genMgr.Lister(appsv1.SchemeGroupVersion.WithResource("deployments"))
+			genMgr.Start()
+			genMgr.WaitForCacheSync()
 			resourceDetector := &ResourceDetector{
 				Client:          fakeClient,
 				DynamicClient:   fakeDynamicClient,
@@ -425,6 +431,9 @@ func Test_cleanUnmatchedCRBs(t *testing.T) {
 			defer cancel()
 			fakeDynamicClient := dynamicfake.NewSimpleDynamicClient(scheme, tt.existingObject)
 			genMgr := genericmanager.NewSingleClusterInformerManager(ctx, fakeDynamicClient, 0)
+			genMgr.Lister(appsv1.SchemeGroupVersion.WithResource("deployments"))
+			genMgr.Start()
+			genMgr.WaitForCacheSync()
 			resourceDetector := &ResourceDetector{
 				Client:          fakeClient,
 				DynamicClient:   fakeDynamicClient,
@@ -659,6 +668,9 @@ func Test_removeRBsClaimMetadata(t *testing.T) {
 			defer cancel()
 			fakeDynamicClient := dynamicfake.NewSimpleDynamicClient(scheme, tt.existingObject)
 			genMgr := genericmanager.NewSingleClusterInformerManager(ctx, fakeDynamicClient, 0)
+			genMgr.Lister(appsv1.SchemeGroupVersion.WithResource("deployments"))
+			genMgr.Start()
+			genMgr.WaitForCacheSync()
 			resourceDetector := &ResourceDetector{
 				Client:          fakeClient,
 				DynamicClient:   fakeDynamicClient,
@@ -884,6 +896,9 @@ func Test_removeCRBsClaimMetadata(t *testing.T) {
 			defer cancel()
 			fakeDynamicClient := dynamicfake.NewSimpleDynamicClient(scheme, tt.existingObject)
 			genMgr := genericmanager.NewSingleClusterInformerManager(ctx, fakeDynamicClient, 0)
+			genMgr.Lister(appsv1.SchemeGroupVersion.WithResource("deployments"))
+			genMgr.Start()
+			genMgr.WaitForCacheSync()
 			resourceDetector := &ResourceDetector{
 				Client:          fakeClient,
 				DynamicClient:   fakeDynamicClient,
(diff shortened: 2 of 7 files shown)


RainbowMango:
@vie-serendipity Can you elaborate on it? Which informer should be synced? And what's the side-effect?

vie-serendipity:
When there are a lot of workloads in the cluster, if the karmada controller manager restarts, the detector starts in parallel with the other controllers. This means that when a controller begins to reconcile on getting workloads, the detector's InformerManager may not have synced cache yet and will fall back to the API server. Frequent API server access increases the load of the API server.

Therefore, it's reasonable to sync cache before real controllers start to work. So I think we can put detector to case hasCache. That's what I do before.

```go
switch runnable := fn.(type) {
	case *Server:

	case hasCache:

	case webhook.Server:

	case warmupRunnable, LeaderElectionRunnable:
              
	default:
		return r.LeaderElection.Add(fn, nil)
}
```

After considering and testing, this will bring a really long start-up time, leading that we have to configure long initialDelaySeconds for liveness probe. This is not good.

I came up with another approach: block while getting or listing the objects from the SingleClusterInformerManager until its cache is synced, just like the client in controller-runtime does.

RainbowMango:
> When there are a lot of workloads in the cluster, if the karmada controller manager restarts, the detector starts in parallel with the other controllers. This means that when a controller begins to reconcile on getting workloads, the detector's InformerManager may not have synced cache yet and will fall back to the API server. Frequent API server access increases the load of the API server.

Do you mean the `controlPlaneInformerManager`?
https://github.com/karmada-io/karmada/blob/e6cee638c64f18808b1acf3324b5626f4c05d714/cmd/controller-manager/app/controllermanager.go#L859
I understand that the controllers widely use this informer manager, but the new informers are registered inthe  detector:
https://github.com/karmada-io/karmada/blob/fe621a2eb3d2a9c4be85b695ce4b1cf7dbf866e4/pkg/detector/detector.go#L186-L198

>  Frequent API server access increases the load of the API server.

How bad is it?

vie-serendipity:
> Do you mean the controlPlaneInformerManager?

Yes.

> How bad is it?

Not too bad. At about the 100k workload level, after a restart, some requests go directly to the API server, like 250 req/s. It seems like the API server should be able to handle this level of load, but I think it could be optimized.

image

RainbowMango:
> At about the 100k workload level, after a restart, some requests go directly to the API server, like 250 req/s. 

It sounds like you have a higher qps configuration, right? As the default QPS is 40.
```
      --kube-api-burst int                                                                                                                                                                                       
                Burst to use while talking with karmada-apiserver. (default 60)
      --kube-api-qps float32                                                                                                                                                                                     
                QPS to use while talking with karmada-apiserver. (default 40)
```

vie-serendipity:
> It sounds like you have a higher qps configuration, right? As the default QPS is 40.

Yes, I set qps and burst to 600 and 100 respectively.

RainbowMango:
The QPS setting reflects the API server's processing capacity, and what we can do is ensure that the total requests from all controllers stay below this threshold. I don't see a strong need for further optimization here.

vie-serendipity:
> what we can do is ensure that the total requests from all controllers stay below this threshold.

I'm not sure we're aligned.
The controller's requests should be watch requests and write requests (create/delete/update); all read requests (get/list) should come from the cache. But here a large number of read requests are going directly to the apiserver, which I believe is not expected. I believe the higher QPS setting is meant to allow the controller to make more write requests, not to perform direct reads.

RainbowMango:
Really? My understanding is that both read/write requests are counted for QPS.  @zach593 Can you confirm that?

vie-serendipity:
> Really? My understanding is that both read/write requests are counted for QPS.

I think you misunderstand me. I mean the controller's read requests should go to the cache (**read requests to cache** are not accounted for QPS) instead of directly to the API server (r**ead requests to API server** are accouted for QPS), while the **informer’s list/watch requests** are accouted for QPS.
The current problem is that the controller can't get data from the cache and falls back to the API server, directly reading and consuming QPS, which I think is not expected. The ideal behaviour is wait for cache synced and then read from cache.

RainbowMango:
Yeah, I get it. Thank you.

An idea in my mind that lets controllers who are using the informer wait until the cache gets synced, but I doubt it is worth doing that, as the QPS is still under control, not going beyond the API server's capacity. Doing so, would slow down the controller start time.

pkg/detector/detector.go
```diff
@@ -194,6 +194,7 @@ func (d *ResourceDetector) discoverResources(ctx context.Context, period time.Du
 			d.InformerManager.ForResource(r, d.EventHandler)
 		}
 		d.InformerManager.Start()
+		d.InformerManager.WaitForCacheSync()
```
zach593, line 197: Why is this wait needed here?

zach593:
> Really? My understanding is that both read/write requests are counted for QPS. @zach593 Can you confirm that?

It seems I don’t need to answer this anymore.

---

> After considering and testing, this will bring a really long start-up time, leading that we have to configure long initialDelaySeconds for liveness probe. This is not good.

Would you like to explain a bit why the hasCache version caused a longer startup time than this version? @vie-serendipity 

Or, can I understand it this way: the cache-sync time is roughly the same, but the hasCache approach affects the liveness probe result?

zach593:
I basically agree with your point that reconciliation should happen after the relevant caches have synced, but I don’t understand in what situation the dynamic client would be used. Do you mean this part?

https://github.com/karmada-io/karmada/blob/d58cbe1ea40ed09f13c26a5fa8492fe8f401ab42/pkg/detector/preemption.go#L228-L230

zach593:
Hi @RainbowMango , do you know the background of using the dynamic client as a backup in the past? I can see this pattern scattered across the repo, and some of the code paths are actually never reached.
Also, in our (ctrip) branch, these dynamic-client behaviors have already been removed as part of performance optimizations.

pkg/detector/detector.go
vie-serendipity: The implementation has some problems. I want to make sure whether this change is necessary.

RainbowMango:
@zach593 Good question! 

The pattern (use a dynamic client) was introduced by #205 when implementing the `binding-controller`, the controller needs to get the referencing resource template from the informer, see [the code here](https://github.com/karmada-io/karmada/pull/205/changes#diff-db0905daa82ac8fff994c62bc2d95b7367f92d467c57932bf0fdb9d51619f1f8R78-R82).

> Also, in our (ctrip) branch, these dynamic-client behaviors have already been removed as part of performance optimizations.

https://github.com/karmada-io/karmada/blob/2746a920308ed7dcccc070d8073f21ebb3b4de7c/pkg/controllers/binding/binding_controller.go#L127-L137

Maybe we don't need the dynamic client even in the binding controller, in case of `NotFound`, we can simply return with a retry (waiting for GC to remove the ResourceBinding or the cache gets synced) . 
@zach593, what do you think?

vie-serendipity:
> Would you like to explain a bit why the hasCache version caused a longer startup time than this version?

Sry for late reply. The hasCache version will wait at startup until all caches have finished syncing. In the current implementation, reconciliation can begin as soon as controller's watched resources have finished syncing.

vie-serendipity:
@zach593 

> Also, in our (ctrip) branch, these dynamic-client behaviors have already been removed as part of performance optimizations.

So for the scenario I mentioned, if the informer manager’s cache hasn’t synced and `get` returns an error, will it just rerurn err until the next reconcile loop?

zach593:
> Maybe we don't need the dynamic client even in the binding controller, in case of `NotFound`, we can simply return with a retry (waiting for GC to remove the ResourceBinding or the cache gets synced) .
> @zach593, what do you think?

@RainbowMango  I like this idea. I don’t think the dynamic client is necessary anywhere. For the binding controller, the retry approach is better than waiting for the cache to sync before returning resources, because it can temporarily skip unsynced items and keep processing others. For other scenarios, though, it may be simpler to just wait for the cache to sync before returning resources.

zach593:
> So for the scenario I mentioned, if the informer manager’s cache hasn’t synced and `get` returns an error, will it just rerurn err until the next reconcile loop?

@vie-serendipity Yes, but we also added a preload function for each informer, similar to what you did, so this error is not easy to trigger.

RainbowMango:
I‘m looking at the issue tagged with milestone v1.17. 
Added an agenda for this to today's [Scalability meeting](https://docs.google.com/document/d/1BUZBDvtbtfU9MmOz7KVdvGiiCny-zz_34FuLlW4AGmM/edit?tab=t.0#heading=h.8d2qii1mb7w8), let's talk about it there.

Issue #7112, opened 2026-01-15
Estimator: The quota plugin failed because namespace information is missing in the gRPC request


zhzhuang-zju:
**What happened**:

The scheduler estimator provides a resource quota plugin that calculates how many replicas a member cluster can accommodate based on the cluster’s quotas. To list the quotas of member clusters, the plugin requires a namespace. 
https://github.com/karmada-io/karmada/blob/d9abe6907035717bf1653fcbdb96ca2be171aecc/pkg/estimator/server/framework/plugins/resourcequota/resourcequota.go#L102-L105 

This namespace information comes from `bindingSpec.replicaRequirements.namespace`.  
https://github.com/karmada-io/karmada/blob/d9abe6907035717bf1653fcbdb96ca2be171aecc/pkg/estimator/client/accurate.go#L155

Tracing further back, we can find that this namespace originates from `PodTemplateSpec.Namespace`.  
https://github.com/karmada-io/karmada/blob/d9abe6907035717bf1653fcbdb96ca2be171aecc/pkg/util/helper/binding.go#L448-L458

However, `PodTemplateSpec.Namespace` is actually empty, causing the quota plugin to fail in correctly listing the quotas.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Karmada version:
- kubectl-karmada or karmadactl version (the result of `kubectl-karmada version` or `karmadactl version`):
- Others:

zhzhuang-zju:
cc @wengyao04

zhzhuang-zju:
https://github.com/karmada-io/karmada/blob/d9abe6907035717bf1653fcbdb96ca2be171aecc/pkg/estimator/client/interface.go#L41

My initial idea is to add a `namespace` parameter to this function.

RainbowMango:
I understand that this is a bug from Karmada upstream, that the default interpreter grabs the namespace from the `PodTemplate` instead of the resource templates' metadata. 

First things first. @seanlaii could you please confirm if this affects your environment? I don't believe that, as I remember you maintained a customized interpreter, right?

zhzhuang-zju:
I'm afraid the impact is not limited to the default interpreter. The `accuratePodRequirements` function is exposed as a library function and is used by many third-party resource interpreters when implementing `replicaResource`.
[...]

PR #7118, opened 2026-01-16
Add namespace interpretation for RayService interpreter

seanlaii:
**What type of PR is this?**
/kind feature

Interpret namespace from the RayService namespace.

Fixes #

```release-note

```

### DIFF
--- a/pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/ray.io/v1/RayService/customizations.yaml
+++ b/pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/ray.io/v1/RayService/customizations.yaml
@@ -92,6 +92,11 @@ spec:
             end
           end
 
+          -- Set namespace from the resource metadata
+          if requires ~= nil and desiredObj.metadata ~= nil and desiredObj.metadata.namespace ~= nil then
+            requires.namespace = desiredObj.metadata.namespace
+          end
+
           return totalReplicas, requires
         end
     replicaRevision:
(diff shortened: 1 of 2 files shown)


XiShanYongYe-Chang:
If the namespace is not handled, is the default behavior to use the same namespace as the resource?

zhzhuang-zju:
@seanlaii Is this PR intended to fix the issue #7112 related to the RayService resource?

pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/ray.io/v1/RayService/customizations.yaml
```diff
@@ -92,6 +92,11 @@ spec:
             end
           end
 
+          -- Set namespace from the resource metadata
```
zhzhuang-zju, line 95: The `requires` are obtained via `kube.accuratePodRequirements`. As mentioned in #7112, it fails to set the namespace correctly. Since multiple CRD resources use `kube.accuratePodRequirements`, this appears to be a common issue. It would be better to find a general solution.

XiShanYongYe-Chang: Are you saying that current resource interpretation practices all have this issue?

zhzhuang-zju: I'm afraid yes. However, the `namespace` is only used when the `features.ResourceQuotaEstimate` feature is enabled.

zhzhuang-zju:
@seanlaii Now that #7124 has been merged, this issue should have been resolved. Could you please help confirm?

seanlaii:
> @seanlaii Now that #7124 has been merged, this issue should have been resolved. Could you please help confirm?

Sorry for the late reply. I was on vacation.
I have confirmed that this issue has been fixed. The namespace will be interpreted from the object's namespace in detector.

PR #7124, opened 2026-01-21
Set bindingSpec.ReplicaRequirements.Namespace uniformly

### LINKED ISSUE #7112: Estimator: The quota plugin failed because namespace information is missing in the gRPC request

zhzhuang-zju:
**What type of PR is this?**
/kind bug

This PR fixes a bug in the scheduler estimator's resource quota plugin. The plugin was failing to correctly list resource quotas because the namespace information was missing from the gRPC request. This caused the plugin to list all ResourceQuota objects across all namespaces, instead of just the intended one.

The fix involves modifying `pkg/detector/detector.go` to ensure that `replicaRequirements.Namespace` is always set by using `object.GetNamespace()` when it's empty and the `ResourceQuotaEstimate` feature is enabled.

Fixes #7112

```release-note
`karmada-scheduler-estimator`: Fixed the issue where the resource quota plugin failed to list resource quotas due to a missing namespace in the gRPC request.
```

### DIFF
--- a/pkg/detector/detector.go
+++ b/pkg/detector/detector.go
@@ -1462,6 +1462,16 @@ func (d *ResourceDetector) applyReplicaInterpretation(object *unstructured.Unstr
 			klog.Errorf("Failed to customize replicas for %s(%s): %v", gvk, name, err)
 			return err
 		}
+
+		// Most interpreters (except webhook interpreter) cannot properly obtain the namespace because they extract
+		// information from PodTemplate, and advanced workloads' PodTemplates typically don't have a namespace set.
+		// Therefore, we uniformly check and set the namespace here.
+		// Note: The replicaRequirements.Namespace field is somewhat redundant and could be deprecated in the future,
+		// as the namespace can be obtained directly from ResourceBinding.
+		if features.FeatureGate.Enabled(features.ResourceQuotaEstimate) && replicaRequirements != nil && len(replicaRequirements.Namespace) == 0 {
+			replicaRequirements.Namespace = object.GetNamespace()
+		}
+
 		spec.Replicas = replicas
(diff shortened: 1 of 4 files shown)


zhzhuang-zju:
> Generally, it looks good, but do we need to clean up the non-functional namespace parser inside the resource interpreter?

I did a quick investigation, and `kube.accuratePodRequirements` is only used by the resource interpreters `replicaResource` and `InterpretComponent`. Among these, `InterpretComponent` is unaffected, and `replicaResource` will be fixed in this PR. Therefore, I think we can safely remove the previous non-functional namespace parser.

seanlaii:
Thanks for the fix!

PR #7127, opened 2026-01-23
Fix nil pointer panic in application failover controller

goyalpalak18:
### Problem summary
I found a crash scenario in the Application Failover controllers (`RBApplicationFailoverController` and `CRBApplicationFailoverController`) when `decisionConditions.tolerationSeconds` is not set.

- `TolerationSeconds` is an optional `*int32`
- The controllers dereference it without a nil check during failure detection
- When omitted, this leads to a nil pointer panic and a controller crash loop
- The CRD default of `300s` is not applied for resources created programmatically or migrated from older versions

This panic happens exactly during cluster failure scenarios, which makes it particularly risky.

---

### 1. Controller-side safety (defense in depth)
I added a nil check in both application failover controllers so `tolerationSeconds` is always initialized before use:

- `pkg/controllers/applicationfailover/rb_application_failover_controller.go`
- `pkg/controllers/applicationfailover/crb_application_failover_controller.go`

If `tolerationSeconds` is nil, it now defaults to **300 seconds**, which prevents panics even when webhook defaulting is skipped.

### 2. Webhook defaulting
I added a shared helper function to apply the same default during admission:

- New helper: `SetDefaultTolerationSeconds()` in `pkg/util/helper/policy.go`
- Called from:
  - `propagationpolicy/mutating.go`
  - `clusterpropagationpolicy/mutating.go`

This ensures newly created policies get the expected default value.

### 3. Unit tests
I added unit tests for the new helper to verify:
- Existing `tolerationSeconds` values are not overwritten
- The default of `300` is applied when the field is missing
- Other fields remain unchanged

---

## Why this fix is complete
- The controllers can no longer panic due to a nil `tolerationSeconds`
- Defaulting is consistent with the CRD-documented value of `300s`
- The change is small, low risk, and suitable for backporting
- The defaulting logic is covered by unit tests

This makes application failover more reliable during real cluster failure scenarios.

### DIFF
--- a/pkg/controllers/applicationfailover/crb_application_failover_controller.go
+++ b/pkg/controllers/applicationfailover/crb_application_failover_controller.go
@@ -125,6 +125,7 @@ func (c *CRBApplicationFailoverController) syncBinding(ctx context.Context, bind
 	// If tolerationSeconds is not set, use the default value of 300 seconds
 	// to avoid nil pointer dereference in detectFailure.
 	if tolerationSeconds == nil {
+		klog.Warningf("TolerationSeconds is nil for ClusterResourceBinding %s, expected to be set by webhook. Defaulting to 300 seconds.", binding.Name)
 		tolerationSeconds = ptr.To[int32](300)
 	}
 
--- a/pkg/controllers/applicationfailover/rb_application_failover_controller.go
+++ b/pkg/controllers/applicationfailover/rb_application_failover_controller.go
@@ -125,6 +125,7 @@ func (c *RBApplicationFailoverController) syncBinding(ctx context.Context, bindi
 	// If tolerationSeconds is not set, use the default value of 300 seconds
 	// to avoid nil pointer dereference in detectFailure.
 	if tolerationSeconds == nil {
+		klog.Warningf("TolerationSeconds is nil for ResourceBinding %s/%s, expected to be set by webhook. Defaulting to 300 seconds.", binding.Namespace, binding.Name)
 		tolerationSeconds = ptr.To[int32](300)
 	}
 


RainbowMango:
@mszacillo @XiShanYongYe-Chang Maybe you can take a look.

pkg/controllers/applicationfailover/crb_application_failover_controller.go
```diff
@@ -121,6 +122,11 @@ func (c *CRBApplicationFailoverController) detectFailure(clusters []string, tole
 func (c *CRBApplicationFailoverController) syncBinding(ctx context.Context, binding *workv1alpha2.ClusterResourceBinding) (time.Duration, error) {
 	key := types.NamespacedName{Name: binding.Name, Namespace: binding.Namespace}
 	tolerationSeconds := binding.Spec.Failover.Application.DecisionConditions.TolerationSeconds
+	// If tolerationSeconds is not set, use the default value of 300 seconds
+	// to avoid nil pointer dereference in detectFailure.
+	if tolerationSeconds == nil {
+		tolerationSeconds = ptr.To[int32](300)
```
mszacillo, line 129: I don't have an issue including an additional fallback for the tolerationSeconds if this is somehow unset, but your additional to the mutation webhook should be enough to safeguard against this field being empty. If we keep this, can we please add a warning log to notify the user that this field is not set, and that we expect it to be?

mszacillo:
@RainbowMango Another comment that you may have more context for. Is there a historical reason as to why we let tolerationSeconds be optional? Was this to make the use of the application failover API a bit easier for the user?

goyalpalak18:
Hey @mszacillo, thanks for the review!

I added that warning log you suggested—good call.

Regarding reproduction: It’s the former. This specifically happens when generating the policy programmatically (e.g., via client-go) without that field set, rather than being triggered by a specific version upgrade.

Ready for another look when you have a sec!

RainbowMango:
> @RainbowMango Another comment that you may have more context for. Is there a historical reason as to why we let tolerationSeconds be optional? Was this to make the use of the application failover API a bit easier for the user?

Yes, exactly.

RainbowMango:
> Regarding reproduction: It’s the former. This specifically happens when generating the policy programmatically (e.g., via client-go) without that field set, rather than being triggered by a specific version upgrade.

I'm not sure about it, just a guess here. 
The default rule is set on CRD. No matter how the policy is created, I think the field will be set with the default value. 

In other words, @goyalpalak18 Could you please tell us how you find this? Or how do you use this feature?

PS: Please don't get me wrong, this is not an objection, just want to try to find the root cause, as we usually trust the default rules in all the APIs. If there is any misunderstanding, we have to take some other actions on the other APIs.

goyalpalak18:
@RainbowMango Thanks for checking that! You are totally right—for normal API requests (like kubectl apply), the CRD default handles this perfectly.

I found that this panic surfaces specifically when the object is built programmatically (like in unit tests or internal controller logic in pkg/detector). Since those paths bypass the API server's admission chain, the default doesn't get applied, and TolerationSeconds can still be nil.

It’s just a defensive check to keep the controller from crashing if we hit one of those edge cases. The warning log I added should help us catch it if it does happen.

RainbowMango:
> I found that this panic surfaces specifically when the object is built programmatically (like in unit tests or internal controller logic in pkg/detector). Since those paths bypass the API server's admission chain, the default doesn't get applied, and TolerationSeconds can still be nil.

Thanks for the clarification. 
I will be shocked if some of the changes bypass the admission chain. 
@XiShanYongYe-Chang Could you please take another look? Verify it, probably?

RainbowMango:
> It’s just a defensive check to keep the controller from crashing if we hit one of those edge cases. The warning log I added should help us catch it if it does happen.

Yeah, I understand. But I hope to clarify if the risk is solid before moving forward. 
/hold

XiShanYongYe-Chang:
> @XiShanYongYe-Chang Could you please take another look? Verify it, probably?

Ok
/assign

XiShanYongYe-Chang:
Hi guys, when a user creates a resource through the client (client-go or kubectl, etc), this value will be set to the default value, so no panic will occur. The detailed analysis is as follows:

For `+kubebuilder:default=300` mechanism:

1. **Annotation to CRD Conversion**

The `+kubebuilder:default=300` annotation in Go code is automatically converted to the `default: 300` field in the generated CRD's OpenAPIV3Schema during the build process.

2. **Request Flow**

```
User Request → kube-apiserver → Admission Webhooks → Schema Validation
→ Default Value Filling (kube-apiserver) → Persistence to etcd → Response
```

3. **Key Points**

- **CRD-level**: Annotation becomes `default` field in CRD OpenAPIV3Schema
- **kube-apiserver-level**: Automatically applies defaults before persistence
- **Client-side**: No need to specify the field; kube-apiserver fills it automatically

> Thanks for checking that! You are totally right—for normal API requests (like kubectl apply), the CRD default handles this perfectly.

The above analysis is consistent with this conclusion.

> I found that this panic surfaces specifically when the object is built programmatically (like in unit tests or internal controller logic in pkg/detector). Since those paths bypass the API server's admission chain, the default doesn't get applied, and TolerationSeconds can still be nil.
> It’s just a defensive check to keep the controller from crashing if we hit one of those edge cases. The warning log I added should help us catch it if it does happen.

In a real-world environment, there is no way to construct a panic error.

A panic will occur when analyzed solely from the function's perspective. In this case, I personally feel that the benefits are not significant, and the code complexity is increased.

Issue #6858, opened 2025-10-20
AggregatedStatus in rb status may be deleted


spiritNO1:
**What happened**:
1、execution_controller updates member workload fail because of conflict, AppliedFailed is added in work's status.

2、After returns error, execution_controller retries, and updates member workload success.

3、But when setting Applied condition to AppliedSuccessful in function updateAppliedCondition, if the work object in informer cache is not updated, controller will not set AppliedSuccessful.

<img width="3400" height="1852" alt="Image" src="https://github.com/user-attachments/assets/08755538-6e0f-4009-8dfc-6a09f7fbd0d4" />

4、Because of GenerationChangedPredicate event filter, new event will not come if only work's status changes.

5、appliedMessage will be added to rb status, and the status from member workload will be empty.

<img width="3190" height="1866" alt="Image" src="https://github.com/user-attachments/assets/bfb976be-1bc1-4f42-a705-84ab159eeccd" />

6、Host workload's status will be empty.

**What you expected to happen**:
Host workload's status will not be effected.

**How to reproduce it (as minimally and precisely as possible)**:
Hard to reproduce

**Anything else we need to know?**:
I think maybe GenerationChangedPredicate can be removed from execution_controller?

**Environment**:
- Karmada version:
- kubectl-karmada or karmadactl version (the result of `kubectl-karmada version` or `karmadactl version`):
- Others:

RainbowMango:
@spiritNO1 Thanks for reporting this.  cc @XiShanYongYe-Chang for helping confirm this.

XiShanYongYe-Chang:
Hi @spiritNO1 Are you referring to the situation where, when updating the work status in step 3, the work get from the cache still has a status condition of "AppliedSuccessful"? Is that correct?

spiritNO1:
> Hi [@spiritNO1](https://github.com/spiritNO1) Are you referring to the situation where, when updating the work status in step 3, the work get from the cache still has a status condition of "AppliedSuccessful"? Is that correct?

Yes, in this time, cache has not modified applied condition to AppliedFailed

XiShanYongYe-Chang:
> Yes, in this time, cache has not modified applied condition to AppliedFailed

Thanks for response, then I understand that there may not be any problem, because one of the fields is different. The `LastTransitionTime` will be updated, which will also cause the `DeepEqual` check to fail, thus allowing the update to continue. How do you think?

spiritNO1:
[...]

PR #7128, opened 2026-01-25
feat: improve deployment creation performance by using informer cache…

### LINKED ISSUE #4225: Performance improvement for create deployment
Performance improvement for create deployment.

**Why is this needed**:
When I create a deployment on the karmada control plane and wait until all pods under the deployment are running (the image has been pulled to the local host in advance), then create the next deployment. Each deployment is delivered to two clusters with 10 pods in each cluster. In this way, 150 deployments (totaling  3000 pods) are created cyclically, The total time is about 3 minutes. It takes about 1 minute to create 150 deployments  (totaling  3000 pods) in a single cluster, It is expected that the performance of creating deployments in karmada is similar to that of a single cluster.

ljluestc:
**What type of PR is this?**

/kind feature

Performance improvement for create deployment.

Currently, `updateResourceStatus` in `pkg/controllers/status/common.go` performs a direct API `Get` and `Update` on the resource status for every Work status update. This bypasses the controller-runtime cache, leading to higher latency and increased load on the API server, especially when creating a large number of deployments concurrently.

This PR optimizes `updateResourceStatus` to attempt fetching the resource from the informer cache first.
1. It tries to get the object from the `InformerManager`.
2. If found, it calculates the new status and attempts to update the resource status via the API server.
3. If the update succeeds, it returns.
4. If the update fails with a conflict (optimistic locking failure) or if the object wasn't found in the cache, it falls back to the original behavior: fetching the latest object directly from the API server and retrying the update.

This "optimistic cache fetch" strategy significantly reduces the number of read operations against the API server during normal operation, improving performance for high-concurrency scenarios.

Fixes #4225

Verified with `go test ./pkg/controllers/status/...`. The retry logic ensures data consistency even if the cache is stale.

```release-note
NONE

### DIFF
--- a/pkg/controllers/status/common.go
+++ b/pkg/controllers/status/common.go
@@ -25,6 +25,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/tools/record"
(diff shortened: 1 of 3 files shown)


RainbowMango:
@ljluestc Thanks for bring this up.

> Currently, updateResourceStatus in pkg/controllers/status/common.go performs a direct API Get and Update on the resource status for every Work status update. This bypasses the controller-runtime cache, leading to higher latency and increased load on the API server, especially when creating a large number of deployments concurrently.

It would be great to have a way that can rely on the cache to update the status. 

Just show the code here and explain the history a little bit:
The referencing code here: https://github.com/karmada-io/karmada/blame/12ad34ccfbb8a6dd7fb0f6ab70cff79fbf2ddee6/pkg/controllers/status/common.go#L148-L213

This code was introduced when fixing #5285 (similar issue with #6858).

RainbowMango:
@ljluestc Please cc me once it is ready for review.

PR #7129, opened 2026-01-25
Fix graceful eviction race condition causing premature Work deletion

Aman-Cool:
# Fix: Graceful Eviction Race Condition

This change fixes a race condition between the graceful eviction controller and the status controller that could cause **Work objects to be deleted too early during failover**.

Previously, graceful eviction relied on `AggregatedStatus` without ensuring it had been updated for the current binding generation. Under certain controller execution orders, this could result in eviction tasks being removed before the new target cluster was ready, causing temporary workload unavailability.

The fix ensures eviction decisions are only made when the status is confirmed fresh for the current scheduling decision, restoring safe and deterministic failover behavior.

### DIFF
--- a/pkg/controllers/gracefuleviction/crb_graceful_eviction_controller.go
+++ b/pkg/controllers/gracefuleviction/crb_graceful_eviction_controller.go
@@ -83,6 +83,7 @@ func (c *CRBGracefulEvictionController) syncBinding(ctx context.Context, binding
 		scheduleResult: binding.Spec.Clusters,
 		observedStatus: binding.Status.AggregatedStatus,
 		hasScheduled:   binding.Status.SchedulerObservedGeneration == binding.Generation,
+		statusUpToDate: isStatusUpToDate(binding.Spec.Clusters, binding.Status.AggregatedStatus),
 	})
 	if reflect.DeepEqual(binding.Spec.GracefulEvictionTasks, keptTask) {
 		return nextRetry(keptTask, c.GracefulEvictionTimeout, metav1.Now().Time), nil
--- a/pkg/controllers/gracefuleviction/evictiontask.go
+++ b/pkg/controllers/gracefuleviction/evictiontask.go
@@ -30,6 +30,10 @@ type assessmentOption struct {
 	scheduleResult []workv1alpha2.TargetCluster
 	observedStatus []workv1alpha2.AggregatedStatusItem
 	hasScheduled   bool
+	// statusUpToDate indicates whether the observed status covers all scheduled clusters.
+	// This is a safeguard against race conditions where the graceful eviction controller
+	// processes a binding before the status controller has updated AggregatedStatus.
+	statusUpToDate bool
 }
 
 // assessEvictionTasks assesses each task according to graceful eviction rules and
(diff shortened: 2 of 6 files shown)


Aman-Cool:
This PR fixes a race where graceful eviction can act on stale status during failover, causing workloads to be deleted before the new cluster is ready. The change makes eviction wait for up-to-date status, preventing accidental downtime.

XiShanYongYe-Chang:
Hi @Aman-Cool Thank you for your feedback.
I would like to know how you encountered this issue? Or, could you give an example to illustrate this problem of competition? I'm a bit unclear on what premature deletion of work means.

Aman-Cool:
Hi @XiShanYongYe-Chang , thanks for taking a look.
I encountered this during a cluster failover scenario. When a Work is re-scheduled to a new target cluster, the graceful eviction controller may run before the status controller updates `AggregatedStatus` for the new binding generation. In that window, the eviction logic sees the old status and concludes the Work is safe to delete, even though the new target cluster hasn’t finished applying it yet.
By “premature deletion,” I mean the Work object is removed from the old cluster before the workload becomes ready on the new cluster, causing a short period of unavailability. The race depends on controller execution order, which makes it intermittent and hard to reproduce consistently.

Aman-Cool:
**Test Plan**

- Deploy a `Work` object bound to a member cluster.
- Trigger rescheduling by simulating member cluster failure or rebinding the `Work` to a different cluster.
- Ensure the graceful eviction controller runs before the status controller updates `AggregatedStatus` for the new binding generation.

**Expected behavior**
- *Before fix*: the `Work` may be deleted from the old cluster before the new target cluster reports ready, causing temporary workload unavailability.
- *After fix*: eviction is deferred until `AggregatedStatus` is confirmed fresh for the current binding generation; the `Work` remains available until the new cluster is ready.

**Verification**
- Repeatedly trigger failover and observe `Work` lifecycle, deletion timing, and status transitions via controller logs and object state.
- Confirm no premature deletion occurs during controller execution order variation.

XiShanYongYe-Chang:
Thanks @Aman-Cool Let me take a look at this.

By the way, may I ask what your company is? Are you currently using Karmada and the ClusterFailover feature?

Aman-Cool:
Thanks @XiShanYongYe-Chang for taking a look!
I’m contributing as an individual. I’m not running Karmada in production, but I’ve been evaluating the ClusterFailover flow and noticed this behavior while reasoning through controller ordering and failover scenarios.

XiShanYongYe-Chang:
Thanks, I get it~

XiShanYongYe-Chang:
Hi @Aman-Cool The situation you mentioned, I think we have already handled it in the allScheduledResourceInHealthyState function:

https://github.com/karmada-io/karmada/blob/b97876796aec8241929f54a0072656b7e09cfa6f/pkg/controllers/gracefuleviction/evictiontask.go#L95-L105

WDYT

Aman-Cool:
@XiShanYongYe-Chang ,That logic only handles the “not applied yet” case. The race happens when observed status exists but changes during eviction, which this PR fixes by preventing premature Work deletion.

XiShanYongYe-Chang:
> @XiShanYongYe-Chang ,That logic only handles the “not applied yet” case. The race happens when observed status exists but changes during eviction, which this PR fixes by preventing premature Work deletion.

I don't think so. When a specific statusItem is empty, it means the status on that cluster has not yet been collected, which is no different from the new code you have added. You can verify this through testing.

Aman-Cool:
@XiShanYongYe-Chang  ,Thanks for the pointer to `allScheduledResourceInHealthyState`.

My intent here is mainly defensive: to make it explicit that eviction should only act on status confirmed fresh for the current binding, and avoid reacting to transient status during rebinding.

If the existing logic already fully covers this, I’m happy to adjust or drop the change. Appreciate the guidance.

XiShanYongYe-Chang:
Thanks @Aman-Cool

XiShanYongYe-Chang:
Hi @Aman-Cool Would you like to drop this change?

Aman-Cool:
@XiShanYongYe-Chang ,Thanks for the discussion and for walking through the existing logic.
Given that `allScheduledResourceInHealthyState` already covers this case, I’m happy to drop the change to avoid adding redundant behavior. 
Thanks for taking the time to walk through this with me and for the guidance throughout the discussion. I really appreciate the context and patience.It was very helpful.
Appreciate the guidance.

Aman-Cool:
Closing this PR since we agreed to drop the change. Thanks again for the review and discussion.

XiShanYongYe-Chang:
Thanks~ @Aman-Cool

PR #7143, opened 2026-01-27
test(util): add unit tests for pkg/util/hash

goyalpalak18:
**What type of PR is this?**
/kind cleanup

---

### Description
I noticed that the `pkg/util/hash` package currently has **0% test coverage**.

This seemed risky because `DeepHashObject` is a foundational function used to generate deterministic names for resources (like PolicyNames). If this function were to behave inconsistently across Go versions or library updates, it could silently break resource propagation without us knowing.

---

### The Fix
I added a comprehensive `hash_test.go` file using the standard table-driven test pattern found elsewhere in the codebase.

Instead of just checking basic success, I ensured we cover the edge cases that matter for stability:
* **Nil handling:** Ensuring it doesn't panic on nil values.
* **Pointers vs Values:** Verifying they hash consistently.
* **State Safety:** confirming the hasher resets correctly between calls so state doesn't "leak" from one object to the next.

This brings the statement coverage for the package to **100%**.

---

### How to Test
1.  Navigate to the package directory.
2.  Run the tests with coverage enabled:

```bash
go test -v -cover ./pkg/util/hash/...

### DIFF
--- a/pkg/util/hash/hash_test.go
+++ b/pkg/util/hash/hash_test.go
@@ -1,0 +1,137 @@
+/*
+Copyright 2026 The Karmada Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package hash
+
+import (
+	"hash/fnv"
+	"testing"
+)
+
+func TestDeepHashObject(t *testing.T) {
+	type testStruct struct {
+		Name  string
+		Value int
+	}
+
+	tests := []struct {
+		name     string
+		obj1     interface{}
+		obj2     interface{}
+		wantSame bool
+	}{
+		{
+			name:     "identical structs produce same hash",
+			obj1:     testStruct{Name: "test", Value: 42},
+			obj2:     testStruct{Name: "test", Value: 42},
+			wantSame: true,
+		},
+		{
+			name:     "different structs produce different hash",
+			obj1:     testStruct{Name: "test1", Value: 42},
+			obj2:     testStruct{Name: "test2", Value: 42},
+			wantSame: false,
+		},
+		{
+			name:     "nil values produce same hash",
+			obj1:     nil,
+			obj2:     nil,
+			wantSame: true,
+		},
+		{
+			name:     "maps with same content produce same hash regardless of insertion order",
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by XiShanYongYe-Chang — COMMENTED** [review]
Thanks @goyalpalak18 
It can be seen that the content of this test is quite comprehensive. However, on the other hand, I feel it is a bit complex. Using so much code to cover the code under test might add some maintenance burden compared to the code being tested. From this perspective, the cost might be a bit too high.

**Comment by goyalpalak18** [comment]
Thanks for the review @XiShanYongYe-Chang!

I totally get where you're coming from—it does look like a lot of test code for such a small function. 

My main motivation was that since `DeepHashObject` drives resource naming (like `GeneratePolicyName`), stability is super critical here. If the underlying library (`kr/pretty`) changes or we hit a weird edge case, the hashes could silently shift and break resource propagation. I wanted these tests to act as a strict safety net against that.

That said, I'm happy to trim it down if you feel the maintenance cost is too high. Just let me know if you'd prefer I drop the complex nested cases and stick to the basics!

**Comment by XiShanYongYe-Chang** [comment]
> If the underlying library (kr/pretty) changes or we hit a weird edge case

Your concern is valid, but I think such drastic changes almost impossible, so I suggest we retain an appropriate level of testing to avoid adding to our maintenance burden.

**Comment by goyalpalak18** [comment]
@XiShanYongYe-Chang Done. I cut it down to just the 3 main test cases to keep it clean. Thanks for the feedback!

File: pkg/util/hash/hash_test.go [hunk 1]
```diff
@@ -0,0 +1,137 @@
+/*
+Copyright 2021 The Karmada Authors.
```
XiShanYongYe-Chang: 2021 -> 2026

XiShanYongYe-Chang:
Thanks
/assign
/gemini review

pkg/util/hash/hash_test.go
goyalpalak18: I've updated the copyright year to 2025. Thanks!

XiShanYongYe-Chang: Sorry, I replied incorrectly; it's already 2026 now.

XiShanYongYe-Chang:
Hi @goyalpalak18, can you help sign off on the commits like before?

goyalpalak18:
> Hi @goyalpalak18, can you help sign off on the commits like before?

Done! I have signed off on the commit.

pkg/util/hash/hash_test.go
XiShanYongYe-Chang: @goyalpalak18 This one seems to have been missed.

goyalpalak18: @XiShanYongYe-Chang Fixed the copyright year to 2026.

XiShanYongYe-Chang:
Thanks
lgtm
Can you help squash the commits into one?

goyalpalak18:
> Thanks lgtm Can you help squash the commits into one?

@XiShanYongYe-Chang I've squashed the commits into one.

PR #7152, opened 2026-01-31
fix: invalidate stale informer cache on cluster re-registration

goyalpalak18:
## Description

This PR addresses a critical race condition and cache invalidation issue where reusing a cluster name (common in disaster recovery, GitOps, or cloud auto-provisioning) caused the `InformerManager` to serve stale cache data from the previous cluster.

### The Problem
Previously, the `MultiClusterInformerManager` identified clusters solely by their name. If a cluster was unregistered and immediately a new physical cluster was registered with the *same name*, the system would return the existing (stopped or stale) informer manager.

This resulted in:
- **Silent failures:** Resources appeared "Applied" but didn't exist in the new cluster.
- **Data corruption:** Status updates were read from stale cache data.
- **Reconciliation loops:** Controllers would get stuck trying to update resources based on incorrect state.

## Solution

I have updated the `InformerManager` logic to track the **Cluster UID** alongside the cluster name to ensure identity verification.

**Key Implementation Details:**
1.  **UID Awareness:** Added `clusterUIDs` map to `MultiClusterInformerManager` to track the active UID for every registered cluster name.
2.  **Smart Invalidation:** Introduced `ForClusterWithUID`. When requested, it checks if the cached manager matches the provided UID.
    - If the UID matches: Returns the existing manager.
    - If the UID differs (name reuse): It stops the old manager, clears the stale cache, and initializes a fresh manager.
3.  **Controller Updates:** Updated `WorkStatusController` and `ClusterStatusController` to use `IsManagerExistWithUID` and `ForClusterWithUID`, ensuring they never operate on a stale client.

## Impact

This fix significantly improves reliability in dynamic multi-cluster environments. It ensures that Karmada correctly distinguishes between different physical clusters that share a name, preventing cross-cluster status corruption and ensuring accurate resource propagation.

## Testing

- **Unit Tests:** Updated `work_status_controller_test.go` and `cluster_status_controller_test.go` to mock UID changes and verify that the manager is correctly refreshed.
- **Manual Verification:**
    1. Register a cluster `cluster-a`.
    2. Deploy work to `cluster-a`.
    3. Unregister `cluster-a`.
    4. Register a *new* cluster with the name `cluster-a` (different IP/UID).
    5. Verified that the controller detects the UID mismatch and successfully syncs the new cache instead of returning stale data.

### DIFF
--- a/pkg/controllers/status/cluster_status_controller.go
+++ b/pkg/controllers/status/cluster_status_controller.go
@@ -217,7 +217,7 @@ func (c *ClusterStatusController) syncClusterStatus(ctx context.Context, cluster
 		// 	We should move this logic to both `execution-controller` and `work-status-controller`.
 		//  After that the 'initializeGenericInformerManagerForCluster' function as well as 'c.GenericInformerManager'
 		//  can be safely removed from current controller.
-		c.initializeGenericInformerManagerForCluster(clusterClient)
+		c.initializeGenericInformerManagerForCluster(clusterClient, cluster)
 
 		var conditions []metav1.Condition
 		conditions, err = c.setCurrentClusterStatus(clusterClient, cluster, &currentClusterStatus)
@@ -259,7 +259,7 @@ func (c *ClusterStatusController) setCurrentClusterStatus(clusterClient *util.Cl
 
 	if c.EnableClusterResourceModeling {
 		// get or create informer for pods and nodes in member cluster
-		clusterInformerManager, err := c.buildInformerForCluster(clusterClient)
+		clusterInformerManager, err := c.buildInformerForCluster(clusterClient, cluster)
 		if err != nil {
 			klog.ErrorS(err, "Failed to get or create informer for Cluster", "cluster", cluster.GetName())
 			// in large-scale clusters, the timeout may occur.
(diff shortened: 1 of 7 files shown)


goyalpalak18:
Hi  @XiShanYongYe-Chang @RainbowMango ,

I’ve addressed the stale cache issue by implementing UID tracking for the informer manager. This ensures we don't hit silent propagation failures or data corruption when a cluster name is reused.

Ready for your review!

RainbowMango:
> Manual Verification:
Register a cluster cluster-a.
Deploy work to cluster-a.
Unregister cluster-a.
Register a new cluster with the name cluster-a (different IP/UID).
Verified that the controller detects the UID mismatch and successfully syncs the new cache instead of returning stale data.

@goyalpalak18 Can this issue be reproduced with these steps?

Just out of curiosity, how do you find this?

goyalpalak18:
> > Manual Verification:
> > Register a cluster cluster-a.
> > Deploy work to cluster-a.
> > Unregister cluster-a.
> > Register a new cluster with the name cluster-a (different IP/UID).
> > Verified that the controller detects the UID mismatch and successfully syncs the new cache instead of returning stale data.
> 
> @goyalpalak18 Can this issue be reproduced with these steps?
> 
> Just out of curiosity, how do you find this?

Hey @RainbowMango, thanks for the review!

Yes, those reproduction steps definitely work.

As for how I found it—I was actually just digging through the MultiClusterInformerManager code to understand the lifecycle better. I noticed that ForCluster strictly checks for the existence of the cluster name in the map.

It occurred to me that if a cluster gets unregistered and immediately re-registered with the same name (which happens in some of our DR scenarios), we'd just silently hand back the old, stale manager because the name matches. Since the UID is unique even if the name is reused, tracking that seemed like the safest way to force a cache refresh.

RainbowMango:
Thanks for the clarification. I will take a look and try to reproduce it on my side. But it may take some time.

goyalpalak18:
> Thanks for the clarification. I will take a look and try to reproduce it on my side. But it may take some time.

Sounds good. Thanks for looking into it.

PR #7153, opened 2026-02-01
feat: optimize the mechanism of create or update dependencies-distribute resourcebinding

### LINKED ISSUE #7139: Analysis of queue backlog when  karmada-controller-manager startup in Large-Scale cluster
When Karmada cluster hosts more than 1000,000 resource templates, the proccessing speed of the karmada-controller-manager is relatively slow during the initial startup phase. This mainly involves tow queues: the resoruce detector queue and the dependencies-distributor queue.
#### resource detector queue
The resource detector queue primarily monitors resource template( such as workloads,PVCs,CongifMaps) in the cluaster and create\update\delete the corresponding ResourceBinding objescts.
When proccessing resource templates, the resource detector queue adds resource tempalates that cannot find a PropagationPolicy to a waiting cache list, and removes them when a PropagationPolicy is found. During this proccess, the untimely release of read-write locks leads to slower processing. The observed phenomennon is that the length of the resource detector queue hardly decreases in the first ten minutes after startup. Although increasing the number of concurrent workers can slightly alleviate the issue, the slow processing problem persists.

The metrics of detector queue at the startup phase:

Image

#### dependencies-distributor queue
The dependencies-distributor queue mainly monitors dependent resource templates (such as PVCs, ConfigMaps) in the cluster and creates, updates, and deletes the corresponding ResourceBinding objects.
The process workers of dependencies-distributor queue will find the ResourceBinding object A of the dependent resource according to the workload ResourceBinding object, and update the `RequiredBy` and `Label` of dependency information. When multiple workloads depend on the same resource, multiple workers will simultaneously update the same ResourceBinding object A, resulting in version conflict errors. These conflicts cause the tasks to re-enter the queue, leading to queue backlog. In this case, even if the number of concurrent workers is increased, more version conflicts and re-queueing will occur, and the queue processing speed remains slow. 
The metrics of dependencies-distributor queue at the startup phase:

Image

**Why is this needed**:
Topic 1:  The waiting cache list in detector controller must be exist?  or optimize the RWlock 
Topic 2:  If `RequiredBy` and `Label`  Is not changed，dependencies-distributor controller could skip update ResourceBinding Objects

LivingCcj:
**What type of PR is this?**
Feature

Fixes #7139

During the initial startup phase for karmada-controller-manager.  `workqueue_depth` of dependencies-distributor queue drop to 0 which need more than 20 minutes, the queue worker is 2.
image

After optimization，`workqueue_depth` of dependencies-distributor queue  drop to 0, which need 5 minutes， the queue worker Is 10.

image

At the same time, the number of version conflicts encountered when updating the dependencies distribute resourcebinding has also decreased significantly.

`karmada-controller-manager`: optimize the mechanism of create or update dependencies-distribute resourcebinding,


XiShanYongYe-Chang:
Thanks~
/assign

pkg/dependenciesdistributor/dependencies_distributor_test.go
```diff
@@ -3361,6 +3371,7 @@ func Test_createOrUpdateAttachedBinding(t *testing.T) {
 				t.Errorf("createOrUpdateAttachedBinding(), Client.Get() error = %v, wantErr %v", err, tt.wantErr)
 			}
 			if !reflect.DeepEqual(existBinding, tt.wantBinding) {
+				t.Logf("%v,%v", existBinding.Spec.ConflictResolution, tt.wantBinding.Spec.ConflictResolution)
```
XiShanYongYe-Chang: Is this a debug log?

XiShanYongYe-Chang: I also tried it, and the log line below does not carry the field names, which indeed makes it inconvenient to locate the issue. Can you help update with `%+v` for formatted output?

XiShanYongYe-Chang (commented):
Thanks, only a few minor comments, other parts lgtm.

Can you help also squash the commits into one?

pkg/dependenciesdistributor/dependencies_distributor_test.go
LivingCcj: Yea，the mentioned line is used for debug, that will remove that

RainbowMango (commented):
/assign

RainbowMango (approved):
/approve
Thanks.
Sorry for the delay.

PR #7163, opened 2026-02-02
Fix silent propagation failure due to stale informer cache in ObjectW…

Ady0333:
…atcher.Create

Replace cache-based existence check with direct API server create to prevent silent resource drift when informer cache is stale.

**What type of PR is this?**
/kind bug

This PR fixes an issue where ObjectWatcher.Create relied on the informer cache to determine whether a resource already existed on a member cluster.
If the cache was stale, the method could return success even when the resource was actually missing, causing silent propagation failures.

The fix changes the create path to always verify resource existence against the member cluster API server. It attempts to create the resource directly and handles AlreadyExists gracefully, ensuring the observed cluster state is always correct.

Tested locally with `make verify` and `make test`.
**MAKE TEST :**
image

```release-note
   NONE
```

### DIFF
--- a/pkg/util/objectwatcher/objectwatcher.go
+++ b/pkg/util/objectwatcher/objectwatcher.go
@@ -98,29 +98,23 @@ func (o *objectWatcherImpl) Create(ctx context.Context, clusterName string, desi
 		return err
 	}
 
-	fedKey, err := keys.FederatedKeyFunc(clusterName, desireObj)
+	clusterObj, err := dynamicClusterClient.DynamicClientSet.Resource(gvr).Namespace(desireObj.GetNamespace()).Create(ctx, desireObj, metav1.CreateOptions{})
 	if err != nil {
-		klog.Errorf("Failed to get FederatedKey %s, error: %v.", desireObj.GetName(), err)
+		if apierrors.IsAlreadyExists(err) {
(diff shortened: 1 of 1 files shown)


Ady0333:
@XiShanYongYe-Chang, this pr has been not reviewed from a very long time. I request you to review it once whenever you have time.

XiShanYongYe-Chang:
> If the cache was stale, the method could return success even when the resource was actually missing, causing silent propagation failures.

@Ady0333 Sorry, I didn't understand the question. Could you please explain it in more detail?

Ady0333:
@XiShanYongYe-Chang,
I was thinking about the case where the controller reads the resource from the informer cache and assumes it still exists, while the actual object might already be deleted in the member cluster. In that situation the method could succeed using the cached object, but the later propagation step might fail because the resource is already gone.

I might be misunderstanding the flow here, so please correct me if this situation cannot actually happen in this controller.

XiShanYongYe-Chang:
Okay, I understand what you mean. Actually, our event triggers are cache-triggered, so the situation you mentioned where something exists in the cache but not in reality will not occur.

XiShanYongYe-Chang:
Hi @Ady0333 do you have any more comments, or can we close this PR?

Ady0333:
> Hi @Ady0333 do you have any more comments, or can we close this PR?

Thanks for the clarification! That makes sense to me now. I don't have further comments we can close this PR.

PR #7169, opened 2026-02-03
fix: prevent nil pointer dereference in application failover controller

goyalpalak18:
### Description
I fixed a critical nil pointer dereference panic in the `applicationfailover` controller and validation webhook.

The `TolerationSeconds` field is a pointer (`*int32`), and while it usually has a default, I found that legacy objects or specific upgrade paths can leave it `nil`. Previously, the code dereferenced this pointer blindly, causing the `karmada-controller-manager` to crash-loop.

---

### Root Cause
The `detectFailure` and `syncBinding` functions assumed `TolerationSeconds` was always populated. I identified that direct dereferencing (`*tolerationSeconds`) without a check was the direct cause of the runtime panic.

---

### Proposed Changes
1. **Defensive Controller Logic:**
   I modified `rb_application_failover_controller.go` and `crb_application_failover_controller.go`. I added a nil check that falls back to the default `300s` if the field is missing, preventing the panic.

2. **Safe Validation:**
   I updated `pkg/util/validation/validation.go` to include a nil-check guard before validating the value.

3. **Tests:**
   I added unit tests to `rb_application_failover_controller_test.go`, `crb_application_failover_controller_test.go`, and `validation_test.go` to confirm that `nil` values are handled gracefully.

### DIFF
--- a/pkg/controllers/applicationfailover/crb_application_failover_controller.go
+++ b/pkg/controllers/applicationfailover/crb_application_failover_controller.go
@@ -121,6 +121,12 @@ func (c *CRBApplicationFailoverController) detectFailure(clusters []string, tole
 func (c *CRBApplicationFailoverController) syncBinding(ctx context.Context, binding *workv1alpha2.ClusterResourceBinding) (time.Duration, error) {
 	key := types.NamespacedName{Name: binding.Name, Namespace: binding.Namespace}
 	tolerationSeconds := binding.Spec.Failover.Application.DecisionConditions.TolerationSeconds
+	// Use default value if tolerationSeconds is nil to prevent nil pointer dereference.
+	// This can happen with objects created before the CRD default was added.
+	if tolerationSeconds == nil {
+		defaultTolerationSeconds := int32(300)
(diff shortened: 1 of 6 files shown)


RainbowMango:
Hi @goyalpalak18 is this the same issue as #7127?

goyalpalak18:
Hey @RainbowMango, yes it is.

I opened this one as a simplified replacement for #7127. Based on the feedback there, I dropped the webhook changes and just kept this defensive nil-check for the controller.

I meant to close #7127 earlier—I'll close it now so we can focus on this lighter fix.

XiShanYongYe-Chang:
Hi @goyalpalak18, first of all, thanks for your contribution.

As I commented earlier, I think this protection is a bit overdone, because logically speaking, when this function is called, this value will not be nil.

XiShanYongYe-Chang:
Hi @goyalpalak18 do you have any other comments or do you agree to close this PR?

goyalpalak18:
No other comments from my side. Let's close it. Thanks!

PR #7171, opened 2026-02-04
Fix application failover timer reset on controller restart

Ady0333:
Persist unhealthy detection timestamps to binding status to prevent the failover toleration timer from resetting when the controller restarts.

**What type of PR is this?**
/kind bug

The application failover controller previously tracked unhealthy detection timestamps only in memory. When the controller restarted, this state was lost, causing the failover toleration timer to reset and delaying failover by the full toleration period again.

This PR persists the first unhealthy detection timestamp in the binding status and restores it on reconciliation. As a result, the toleration timer continues correctly across controller restarts, ensuring failover happens within the expected time window.

This aligns failover behavior with real-world production conditions where controller restarts are common.

Tested locally with make verify and make test.

```release-note
NONE
```

### DIFF
--- a/pkg/apis/work/v1alpha2/binding_types.go
+++ b/pkg/apis/work/v1alpha2/binding_types.go
@@ -443,6 +443,12 @@ type ResourceBindingStatus struct {
 	// AggregatedStatus represents status list of the resource running in each member cluster.
 	// +optional
 	AggregatedStatus []AggregatedStatusItem `json:"aggregatedStatus,omitempty"`
+
+	// FailoverHistory records the timestamp when each cluster's workload first became
+	// unhealthy. Used by the application failover controller to track toleration periods
+	// across controller restarts.
+	// +optional
+	FailoverHistory map[string]metav1.Time `json:"failoverHistory,omitempty"`
 }
 
 // AggregatedStatusItem represents status of the resource running in a member cluster.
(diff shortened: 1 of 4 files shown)


Ady0333:
@whitewindmills and @zach593 ,

`make verify` fails locally due to gofmt diffs inside _go/pkg/mod/golang.org/toolchain (Go toolchain cache).
No repository files are modified.
CI should validate with the project toolchain.

XiShanYongYe-Chang:
Hi @Ady0333, thanks. First of all, thank you for your contribution.

However, from my perspective, this should not be considered a bug. We should be cautious about application migration behavior. When a component restarts, which is an extremely rare case, making it wait a little longer is not a big deal. In addition, the current PR introduces an additional API, which comes at a somewhat high cost.

I have a small suggestion. If you find a situation that might be an issue in the future, you can first submit an issue for discussion, and then we can decide how to fix it.

Ady0333:
Hi @XiShanYongYe-Chang, thanks for the feedback.

I understand your concern about behavior change and the additional API cost. That makes sense.

I’ll open an issue first to describe the scenario and discuss whether this is worth addressing, and we can decide the right approach there.

Please let me know if you’d prefer me to close this PR for now. Thanks for the guidance.

XiShanYongYe-Chang:
Hi @Ady0333 From my perspective, I really do not recommend moving forward with this issue, and it would be even better if we could base our decision on actual user feedback.

Ady0333:
Thanks for the clarification, @XiShanYongYe-Chang.
That makes sense. I’ll close this PR for now and wait for real user feedback before revisiting this.
Appreciate your time and guidance.

Issue #7064, opened 2025-12-31
[Umbrella] Workload Affinity Feature - phase I


RainbowMango:
**Summary**:

This issue is used to track the iteration tasks of feature `Workload Affinity` described at #4968.
See proposal(#6685) for more details.

**Iteration Tasks -- Part-1: workload affinity/anti-affinity**:
- [x] Finalize the proposal. (@mszacillo, #6685)
- [x] Introduce anti-affinity field to `PropagationPolicy`. (@mszacillo, #7131)
- [x] Introduce affinity groups fields to ResourceBinding as described at [this discussion](https://github.com/karmada-io/karmada/pull/6685#issuecomment-3717672003) (@RainbowMango, #7144)
- [x] Grab the affinity group from the resource template and populate it in ResourceBinding. (@dahuo98, #7166)
- [x] Introduce a new feature gate for this feature, started with `alpha` and disabled by default. (@mszacillo, #7148)
  * Feature gate name candidates: `WorkloadAffinity`, `SingleWorkloadAffinityGroup`.
- [x] karmda-scheduler: build a global index based on the affinity group identifier declared in ResourceBinding. (@mszacillo, @zhzhuang-zju , #7221)
- [x] Introduce a filter plugin for filter clusters by taking anti-affinity group into account. (@zhzhuang-zju, #7189)
- [x] Setup E2E tests. (@XiShanYongYe-Chang, #7193)
- [x] Add documentation (@RainbowMango, https://github.com/karmada-io/website/pull/985)
- [x] Others(bugfix/enhancements)
  * https://github.com/karmada-io/karmada/pull/7195
  * https://github.com/karmada-io/karmada/pull/7217
  * https://github.com/karmada-io/karmada/pull/7195
  * https://github.com/karmada-io/karmada/pull/7222

mszacillo:
Thanks @RainbowMango for populating the iteration tasks! 

Overall I think the tasks make sense, though I suspect building a global index in the scheduler could probably be broken down. But that can be done once we start working on it. 

Out of curiosity, could you elaborate on:

> Refactor the ResourceBinding API to define its own independent placement structure instead of reusing the one from PropagationPolicy, as their placement semantics and fields are not identical.

Was this a change that we've been meaning to do for a while, and this feature provides a good motivation for prioritizing it? I wasn't aware that this was an ongoing concern.

RainbowMango:
[...]

PR #7177, opened 2026-02-06
Extend scheduler filter plugin by introducing an extendable parameters

RainbowMango:
**What type of PR is this?**

/kind feature

- This PR introduces an optional `FilterPluginWithContext` interface to make filter parameters extensible without breaking existing plugins.
- Adds a `WorkloadAntiAffinity` filter plugin that implements FilterWithContext, registered only when WorkloadAffinity feature is enabled.

Fixes #
Part of #7064

```release-note
`karmada-scheduler`: Added optional `FilterPluginWithContext` (`FilterWithContext` method) for filter plugins, and introduced a WorkloadAntiAffinity filter plugin gated by the `WorkloadAffinity` feature gate.
```

### DIFF
--- a/pkg/scheduler/framework/plugins/workloadantiaffinity/workload_anti_affinity.go
+++ b/pkg/scheduler/framework/plugins/workloadantiaffinity/workload_anti_affinity.go
@@ -1,0 +1,84 @@
+/*
+Copyright 2026 The Karmada Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package workloadantiaffinity
+
+import (
+	"context"
+
+	"k8s.io/klog/v2"
+
+	clusterv1alpha1 "github.com/karmada-io/karmada/pkg/apis/cluster/v1alpha1"
+	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
+	"github.com/karmada-io/karmada/pkg/scheduler/framework"
+)
+
+const (
+	// Name is the name of the plugin used in the plugin registry and configurations.
+	Name = "WorkloadAntiAffinity"
+)
+
+// WorkloadAntiAffinity is a plugin that checks if a cluster matches the workload anti-affinity group constraint.
+// It separates workloads in the same anti-affinity group across different clusters.
+type WorkloadAntiAffinity struct{}
+
+var (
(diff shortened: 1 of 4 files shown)


pkg/scheduler/framework/plugins/workloadantiaffinity/workload_anti_affinity.go
```diff
@@ -0,0 +1,84 @@
...
+
+	antiAffinityGroup := bindingSpec.WorkloadAffinityGroups.AntiAffinityGroup
+
+	// Check if current cluster already has workloads with the same anti-affinity group
+	// if yes --> deny
+	klog.Infof("checking cluster: %s, with anti-affinity group: %s", cluster.GetName(), antiAffinityGroup)
+
+	// If cluster does not match anti-affinity group label, allow scheduling
+	return framework.NewResult(framework.Success)
+}
```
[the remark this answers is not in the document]
RainbowMango: Yes, it is intended for now.

RainbowMango:
cc @mszacillo 

PS: I tested it locally, it works as expected:
```
{"ts":1770375143358.3372,"caller":"workloadantiaffinity/workload_anti_affinity.go:80","msg":"checking cluster: member1, with anti-affinity group: app.group=backend","v":0}
```

RainbowMango:
@CharlesQQ Please take a look. This PR extends the filter plugin, but doesn't require any changes for your private plugins `now`. The previous interface will be deprecated under a deprecation policy.

pkg/scheduler/framework/runtime/framework.go
```diff
@@ -100,23 +100,35 @@ func (frw *frameworkImpl) RunFilterPlugins(
...
 
-func (frw *frameworkImpl) runFilterPlugin(
-	ctx context.Context,
-	pl framework.FilterPlugin,
-	bindingSpec *workv1alpha2.ResourceBindingSpec,
-	bindingStatus *workv1alpha2.ResourceBindingStatus,
-	cluster *clusterv1alpha1.Cluster,
-) *framework.Result {
+// runFilterPluginWithContext calls FilterPluginWithContext.FilterWithContext if available, otherwise falls back to FilterPlugin.Filter.
+func (frw *frameworkImpl) runFilterPluginWithContext(pl framework.FilterPlugin, filterCtx *framework.FilterContext) *framework.Result {
```
zhzhuang-zju, line 120: The `context` is usually the first parameter.
```suggestion
func (frw *frameworkImpl) runFilterPluginWithContext(filterCtx *framework.FilterContext, pl framework.FilterPlugin) *framework.Result {
```

mszacillo:
> PS: I tested it locally, it works as expected:

Nice! After this merges I'll rebase my existing index PR.

pkg/scheduler/framework/runtime/framework.go
RainbowMango: This context is not the one in stdlib.

RainbowMango:
> > PS: I tested it locally, it works as expected:
> 
> Nice! After this merges I'll rebase my existing index PR.

@mszacillo @zhzhuang-zju I merged this as I need to explore the implementation further based on this patch. 

@mszacillo, now you can rebase your PR, and I will look at the cache management part then.

PR #7181, opened 2026-02-09
fix: persist failover toleration timestamps across controller restarts

Ady0333:
### Summary

Fix application failover toleration timer being reset on controller restart.

The failover controllers relied on an in-memory map to track when a workload was first observed as unhealthy. On controller restarts (leader election switch, pod restart, rolling upgrade), this state was lost and the toleration timer restarted from zero, leading to unexpected failover delays in HA environments.

---

### Problem

The application failover controllers record unhealthy detection timestamps only in memory. When the controller restarts, this state is reinitialized and the original unhealthy start time cannot be recovered.

This causes the toleration timer to restart even if the workload has already been unhealthy for most of the configured toleration period. With frequent restarts, failover may be delayed indefinitely without any visible error.

---

### Fix

Persist the first unhealthy detection timestamp per cluster as an annotation on the binding.

During reconciliation:
- Restore timestamps from the annotation when in-memory state is empty (e.g. after restart)
- Use the restored timestamps for failover detection instead of starting from `now`
- Persist updated timestamps back to the annotation only when changes occur

The in-memory map remains the hot path for normal reconciliation, while the annotation is used only to recover state across restarts.

---

### Impact

- Prevents silent reset of the failover toleration timer
- Ensures failover triggers within the configured toleration window
- Improves reliability in HA and production deployments
- No API or CRD changes
- Backward compatible with existing bindings

---

### Tests

- Manually tested application failover with controller-manager restart
- Verified toleration timer continues from the original unhealthy timestamp
- Verified backward compatibility when annotation is not present

### DIFF
--- a/pkg/controllers/applicationfailover/common.go
+++ b/pkg/controllers/applicationfailover/common.go
@@ -113,6 +113,9 @@ func (m *workloadUnhealthyMap) deleteIrrelevantClusters(key types.NamespacedName
 // timestamps so that the toleration timer survives controller restarts.
 const workloadUnhealthyTimestampsAnnotation = "work.karmada.io/unhealthy-first-detected"
 
+// restoreFromAnnotation populates the in-memory map from persisted annotation data.
+// It is called at the beginning of syncBinding to recover timestamps after a controller restart.
+// If the in-memory map already has entries for the given key, this is a no-op.
 func (m *workloadUnhealthyMap) restoreFromAnnotation(key types.NamespacedName, annotations map[string]string) {
 	m.Lock()
 	defer m.Unlock()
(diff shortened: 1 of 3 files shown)


Ady0333:
@zhzhuang-zju could you please take a look when you get a chance?

This PR fixes a case where the application failover toleration timer was reset on controller restart.
Thanks in advance for your time!

XiShanYongYe-Chang:
Hi @Ady0333, First of all, thank you for contributing to the community.

I think that this PR and what #7171 described are essentially the same kind of thing. Therefore, my suggestion is to keep it the same.

Ady0333:
> Hi @Ady0333, First of all, thank you for contributing to the community.
> 
> I think that this PR and what #7171 described are essentially the same kind of thing. Therefore, my suggestion is to keep it the same.

Thanks for pointing that out. You’re right, this overlaps with my previous PR.

I’ll close this one to avoid duplication and follow the discussion on the earlier PR instead.  
Appreciate the review and guidance.

PR #7183, opened 2026-02-10
Fix ApplyReplica to support scale-up from zero when replicas field is missing

zhengjr9:
Remove the conditional check that prevented setting replicas when the field doesn't exist, allowing scale-up from zero.

**What type of PR is this?**
/kind bug

This PR fixes an issue where `ApplyReplica` failed to set the replicas field when scaling from zero. The original implementation only set the replica value when the field already existed in the resource spec (`ok == true`). This caused silent failures during scale-up operations where the replicas field was not present.

The fix removes the conditional check and always sets the desired replica value, regardless of whether the field exists. This ensures scale-up from zero works correctly.

Fixes the scale-up from zero failure when the replicas field is missing in resource spec.

- Updated `ApplyReplica` function to unconditionally set replicas field
- Updated test case `replicas field is not existing` to expect replicas=2 instead of 0
- All tests pass: `go test -v -run TestApplyReplica`

```release-note
`karmada-controller-manager`: Fixed CronFederatedHPA scale-up from zero failure when the replicas field is missing.
```

### DIFF
--- a/pkg/util/helper/unstructured.go
+++ b/pkg/util/helper/unstructured.go
@@ -56,6 +56,17 @@ func ApplyReplica(workload *unstructured.Unstructured, desireReplica int64, fiel
 	return nil
 }
 
+// ApplyReplicaAlways unconditionally applies the Replica value for the specific field.
+// Unlike ApplyReplica, this function sets the field regardless of whether it exists or not.
+// This is useful for scenarios like scaling from zero where the replicas field may be missing.
+func ApplyReplicaAlways(workload *unstructured.Unstructured, desireReplica int64, field string) error {
+	_, _, err := unstructured.NestedInt64(workload.Object, util.SpecField, field)
+	if err != nil {
+		return err
+	}
+	return unstructured.SetNestedField(workload.Object, desireReplica, util.SpecField, field)
+}
+
 // ToUnstructured converts a typed object to an unstructured object.
 func ToUnstructured(obj interface{}) (*unstructured.Unstructured, error) {
 	uncastObj, err := runtime.DefaultUnstructuredConverter.ToUnstructured(obj)
--- a/pkg/controllers/cronfederatedhpa/cronfederatedhpa_job.go
+++ b/pkg/controllers/cronfederatedhpa/cronfederatedhpa_job.go
@@ -203,7 +203,7 @@ func (c *ScalingJob) ScaleWorkloads(cronFHPA *autoscalingv1alpha1.CronFederatedH
 	}
 
 	if scale.Spec.Replicas != *c.rule.TargetReplicas {
-		if err := helper.ApplyReplica(scaleObj, int64(*c.rule.TargetReplicas), util.ReplicasField); err != nil {
+		if err := helper.ApplyReplicaAlways(scaleObj, int64(*c.rule.TargetReplicas), util.ReplicasField); err != nil {
 			klog.ErrorS(err, "CronFederatedHPA applies Replicas failed", "cronFederatedHPA", c.namespaceName, "namespace", cronFHPA.Namespace, "name", cronFHPA.Spec.ScaleTargetRef.Name)
 			return err
 		}
--- a/pkg/util/helper/unstructured_test.go
+++ b/pkg/util/helper/unstructured_test.go
@@ -156,6 +156,71 @@ func TestApplyReplica(t *testing.T) {
 	}
 }
 
+func TestApplyReplicaAlways(t *testing.T) {
+	testCases := []struct {
+		name             string
+		workload         *unstructured.Unstructured
+		replicas         int64
+		field            string
+		expectedReplicas int64
+		expectedErr      bool
+	}{
+		{
+			name: "replica with wrong type",
+			workload: &unstructured.Unstructured{
+				Object: map[string]interface{}{
+					"spec": map[string]interface{}{
+						"replicas": "some",
+					},
+				},
+			},
+			replicas:    2,
+			field:       "replicas",
+			expectedErr: true,
+		},
+		{
+			name: "replicas field is not existing - should set it",
+			workload: &unstructured.Unstructured{
+				Object: map[string]interface{}{
+					"spec": map[string]interface{}{},
+				},
+			},
+			replicas:         2,
+			field:            "replicas",
+			expectedReplicas: 2,
+			expectedErr:      false,
+		},
+		{
+			name: "apply replicas success",
+			workload: &unstructured.Unstructured{
+				Object: map[string]interface{}{
+					"spec": map[string]interface{}{
+						"replicas": int64(1),
+					},
+				},
+			},
+			replicas:         2,
+			field:            "replicas",
+			expectedReplicas: 2,
+			expectedErr:      false,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			if err := ApplyReplicaAlways(tc.workload, tc.replicas, tc.field); (err != nil) != tc.expectedErr {
+				t.Errorf("ApplyReplicaAlways expected error %v, but got %v", tc.expectedErr, err)
+			}
+			if !tc.expectedErr {
+				gotReplicas, _, _ := unstructured.NestedInt64(tc.workload.Object, "spec", tc.field)
+				if gotReplicas != tc.expectedReplicas {
+					t.Errorf("ApplyReplicaAlways expected replicas %d, but got %d", tc.expectedReplicas, gotReplicas)
+				}
+			}
+		})
+	}
+}
+
 func TestToUnstructured(t *testing.T) {
 	testCases := []struct {
 		name        string


zhzhuang-zju:
@zhengjr9 Thanks for the feedback. Could you share the reproduction steps? This would help us identify the issue.

zhengjr9:
@zhzhuang-zju Thanks for reviewing! Here's the detailed explanation of the issue:

### Root Cause

In the CronFederatedHPA scheduled scaling feature, when a workload has **zero replicas**, the Kubernetes API server returns a Scale object with an **empty spec**, as shown below:

```json
{
  "kind": "Scale",
  "apiVersion": "autoscaling/v1",
  "metadata": {
    "name": "t-mpts-mysql-cronhpa",
    "namespace": "app-t-mpts-gwjxq",
    "uid": "811ab552-045d-42d8-ab86-0b89ac581a5a",
    "resourceVersion": "1538036983",
    "creationTimestamp": "2026-01-28T09:50:34Z"
  },
  "spec": {}
}
```

Notice that \`spec.replicas\` is **missing** from the response when the current replica count is zero.

### Reproduction Steps

**Scenario**: Scale up a workload from zero using CronFederatedHPA

1. Create a workload (e.g., Deployment) and scale it to 0 replicas
2. Create a CronFederatedHPA to scale it up:
   ```yaml
   apiVersion: autoscaling.karmada.io/v1alpha1
   kind: CronFederatedHPA
   metadata:
     name: test-cronhpa
     namespace: default
   spec:
     scaleTargetRef:
       apiVersion: apps/v1
       kind: Deployment
       name: test-deployment
     rules:
     - name: scale-up
       schedule: "*/5 * * * *"
       targetReplicas: 2
   ```

3. When the cron job triggers, the controller executes \`ScaleWorkloads()\` in \`cronfederatedhpa_job.go\`

**Code Path** (\`pkg/controllers/cronfederatedhpa/cronfederatedhpa_job.go:167-219\`):

```go
func (c *ScalingJob) ScaleWorkloads(cronFHPA *autoscalingv1alpha1.CronFederatedHPA) error {
    // ...
    
    // Line 192: Get the Scale object
    scaleObj := &unstructured.Unstructured{}
    err = scaleClient.Get(ctx, targetResource, scaleObj)
    
    // Line 199: Convert to typed Scale
    scale := &autoscalingv1.Scale{}
    err = helper.ConvertToTypedObject(scaleObj, scale)
    
    // Line 206: Apply replicas - THIS FAILS
    if err := helper.ApplyReplica(scaleObj, int64(*c.rule.TargetReplicas), util.ReplicasField); err != nil {
        return err
    }
    
    // Line 210: Update the scale
    err := scaleClient.Update(ctx, targetResource, client.WithSubResourceBody(scaleObj))
}
```

**Problem (before fix)**:
- Line 192: \`scaleObj\` has \`spec: {}\` (no replicas field)
- Line 206: \`ApplyReplica()\` checks if \`spec.replicas\` exists
- Since the field doesn't exist (\`ok == false\`), the function returns without setting any value
- Line 210: Update call sends the unchanged Scale object
- **Result**: Scale-up silently fails

**After fix**:
- \`ApplyReplica()\` always sets the desired replica value, even if the field is missing
- The Scale object is correctly updated with \`spec.replicas: 2\`
- Scale-up succeeds

### Test Evidence

The test case \`"replicas field is not existing"\` in \`pkg/util/helper/unstructured_test.go\` validates this scenario:

```go
{
    name: "replicas field is not existing",
    workload: &unstructured.Unstructured{
        Object: map[string]interface{}{
            "spec": map[string]interface{}{}, // Empty spec, no replicas field
        },
    },
    replicas:         2,
    field:            "replicas",
    expectedReplicas: 2,  // Now correctly expects 2 instead of 0
    expectedErr:      false,
}
```

All tests pass: ✅ \`go test -v -run TestApplyReplica\`

Let me know if you need any additional information!

zhzhuang-zju:
@zhengjr9 Thanks, I'll give it a try based on your description. BTW, are you using CronFederatedHPA? If so, what kind of scenarios are you using it for?

zhengjr9:
@zhzhuang-zju Yes, we are using CronFederatedHPA in production.

**Use Case**: We use it for GPU-intensive workloads. We configure scheduled scaling rules to optimize GPU resource utilization:
- **Scale down to zero** during off-peak hours to save costly GPU resources
- **Scale up from zero** before peak hours to handle incoming workloads

This is particularly useful for:
- Machine learning training jobs that run on a schedule
- Batch processing workloads that only need GPU resources at specific times
- Development/testing environments where GPU resources are expensive and should be released when not in use

The bug we encountered was specifically in the **scale-up from zero** scenario, where the CronFederatedHPA would silently fail to scale the workload because the Scale object returned an empty `spec: {}` when replicas were at zero.

zhzhuang-zju:
@zhengjr9 Thank you for the detailed description. This use case of using CronFederatedHPA to optimize GPU resource utilization is very valuable. I hope to have more discussions with you when there's an opportunity.
Regarding this issue, I have also reproduced it locally. I will review your changes as soon as possible.

zhzhuang-zju:
> Connecting to github.com (github.com)|140.82.112.4|:443... connected.
HTTP request sent, awaiting response... 502 Bad Gateway or Proxy Error

Looks like a network issue
/retest

zhengjr9:
@zhzhuang-zju I would be very happy to have more discussions with you in the future, and I hope to contribute more to the Karmada community.

In AI scenarios, multi-cluster scheduling can unlock many powerful capabilities. For example, by abstracting a common GPU resource protocol through adapters, we can integrate with GPUs from different vendors and provide unified scheduling for both physical GPU compute and virtualized GPU resources. This is especially valuable for improving the bin-packing efficiency of virtualized GPU resources across multiple clusters, which is critical for optimizing costs in AI workloads.

zhzhuang-zju:
> I would be very happy to have more discussions with you in the future, and I hope to contribute more to the Karmada community.

You are very welcome. We look forward to future collaborations!

> In AI scenarios, multi-cluster scheduling can unlock many powerful capabilities. For example, by abstracting a common GPU resource protocol through adapters, we can integrate with GPUs from different vendors and provide unified scheduling for both physical GPU compute and virtualized GPU resources. This is especially valuable for improving the bin-packing efficiency of virtualized GPU resources across multiple clusters, which is critical for optimizing costs in AI workloads.

Indeed! We are also very keen to do more in AI scenarios, but we currently lack input from real-world scenarios. So, the scenarios you provided are very meaningful.

Regarding this particular fix, I agree with your changes. However, due to the wide scope of the `ApplyReplica` function, I am currently analyzing its impact. This might take some time.

zhzhuang-zju:
When `spec.replicas` is 0, its Scale object's `spec` field is indeed empty:
```bash
kubectl get --raw /apis/apps/v1/namespaces/default/deployments/nginx/scale
{"kind":"Scale","apiVersion":"autoscaling/v1","metadata":{"name":"nginx","namespace":"default","uid":"47747403-a967-4d20-a910-eb937dbeec74","resourceVersion":"3379159","creationTimestamp":"2026-02-10T07:54:33Z"},"spec":{},"status":{"replicas":0,"selector":"app=nginx"}}
```
As such, `(Cron)FederatedHPA` incorrectly skips updating the `replicas` of the Scale object using `ApplyReplica` in this scenario.

However, the `ApplyReplica` function is used in multiple places, and through analysis, its usage can be broadly categorized into two scenarios:
-   **Scenario 1: Conditional Application** - Applies the `replicas` value only if the field already exists. For example:
https://github.com/karmada-io/karmada/blob/48f669fbeb96684bf4a7654003c387d6a7dccd95/pkg/controllers/binding/common.go#L97-L105
-   **Scenario 2: Unconditional Application** - Applies the `replicas` value regardless of whether the field exists (as documented in this PR).

Given this, could we consider introducing a new function, for example, `ApplyReplicaAlways`, which would be used to set the corresponding field according to the input parameters, regardless of whether it exists or not? This would differentiate it from `ApplyReplica`, thereby avoiding unintended broad impact or introducing new side effects.

WDYT? @zhengjr9 @RainbowMango

zhengjr9:
@zhzhuang-zju Done! I've implemented the second approach as you suggested.

**Changes:**
- Restored `ApplyReplica` to original behavior (only updates existing fields)
- Added new function `ApplyReplicaAlways` for unconditional updates
- Updated CronFederatedHPA to use `ApplyReplicaAlways`

All tests pass ✅

PTAL when you have time. Thanks!

pkg/util/helper/unstructured.go
```diff
@@ -47,11 +47,27 @@ func ApplyReplica(workload *unstructured.Unstructured, desireReplica int64, fiel
...
-		}
+	if !ok {
+		return nil
+	}
+	err = unstructured.SetNestedField(workload.Object, desireReplica, util.SpecField, field)
+	if err != nil {
+		return err
+	}
+	return nil
+}
```
zhzhuang-zju: Function `ApplyReplica` could remain unchange

pkg/util/helper/unstructured.go
```diff
@@ -47,11 +47,27 @@ func ApplyReplica(workload *unstructured.Unstructured, desireReplica int64, fiel
...
+// Unlike ApplyReplica, this function sets the field regardless of whether it exists or not.
+// This is useful for scenarios like scaling from zero where the replicas field may be missing.
+func ApplyReplicaAlways(workload *unstructured.Unstructured, desireReplica int64, field string) error {
+	_, _, err := unstructured.NestedInt64(workload.Object, util.SpecField, field)
+	if err != nil {
+		return err
+	}
+	err = unstructured.SetNestedField(workload.Object, desireReplica, util.SpecField, field)
+	if err != nil {
+		return err
```
zhzhuang-zju:
```suggestion
	return unstructured.SetNestedField(workload.Object, desireReplica, util.SpecField, field)
```

zhzhuang-zju (commented):
Thanks, others LGTM
Please make sure to squash your commits after making change to make sure the PR is ready to get merged.

zhzhuang-zju:
> As such, (Cron)FederatedHPA incorrectly skips updating the replicas of the Scale object using ApplyReplica in this scenario.

The `FederatedHPA` also utilizes `helper.ApplyReplica`. However, unlike `CronFederatedHPA`, `FederatedHPA` does not support scaling up from zero.
https://github.com/karmada-io/karmada/blob/48f669fbeb96684bf4a7654003c387d6a7dccd95/pkg/controllers/federatedhpa/federatedhpa_controller.go#L318-L322
Therefore, although theoretically `FederatedHPA` could encounter the same issue, it's difficult to trigger in a production environment. For this reason, I think this PR can focus solely on `CronFederatedHPA`.

zhengjr9:
@zhzhuang-zju Done! I've addressed your review comments:

1. ✅ Restored `ApplyReplica` to original implementation (unchanged)
2. ✅ Simplified `ApplyReplicaAlways` return statement as suggested
3. ✅ Squashed into single commit

All tests pass. Ready for merge!

zhengjr9:
@zhzhuang-zju Thanks for the clarification! 

Agreed - this PR focuses solely on CronFederatedHPA as it's the practical production scenario where scale-from-zero is needed. FederatedHPA doesn't support scaling from zero, so the issue won't occur there.

The implementation is ready for your final review. 👍

zhzhuang-zju:
@zhengjr9 Thanks for your quick response!

I noticed a couple of minor points that would be great if addressed:
1. There are still two commits. It would be better if they could be squashed into a single commit.
2. The CI `DCO` check failed. You might need to add a sign-off by running the command `git commit -s -m "<commit Msg>"`.

zhengjr9:
@zhzhuang-zju Thanks for your review! 

Squashed into single commit with DCO sign-off.

RainbowMango:
@zhengjr9 Thanks for doing this and sharing your use case! It's great to see another use case for GPU resource optimization in real-world scenarios.

I'd love to invite you to share your experience at one of our [Karmada community meetings](https://github.com/karmada-io/karmada?tab=readme-ov-file#meeting).  This would be a nice opportunity for us to learn more about your use case and identify any gaps to guide future deployment. 

You can pick a convenient time and add your agenda to our [meeting notes](https://docs.google.com/document/d/1y6YLVC-v7cmVAdbjedoyR5WL0-q45DBRXTvz5_I7bkA/edit). (By joining[ the Google Groups](https://groups.google.com/forum/#!forum/karmada) you will be able to edit the meeting notes.)

In addition, you might not be in [the Karmada Adopter Group](https://github.com/karmada-io/community/tree/main/adopter-group) yet. We'd be delighted to have you join! It's a great way to connect with other Karmada users and share experiences. You can join by following [the instructions here](https://github.com/karmada-io/community/tree/main/adopter-group#how-to-join).

zhzhuang-zju:
/approve

Let's merge this PR now as we need to release the RC release today. Hope to keep in touch 😄

zhengjr9:
@zhzhuang-zju Thank you so much for the review and approval! I will continue to follow and contribute to the Karmada community.

zhzhuang-zju:
@zhengjr9 That's great! We’ll backport this fix to `release-1.14`, `release-1.15`, and `release-1.16`. The new versions will be out in late February, so you can use the fix then. Are you interested in helping out with the backport?

Also, I sent you an email just to connect. Looking forward to hearing from you.

zhengjr9:
@RainbowMango Thank you for the invitation! 

I'd be happy to join one of the upcoming Karmada community meetings. I will join the Google Groups and add my agenda to the meeting notes once I have a suitable time slot.

I'm also interested in joining the Karmada Adopter Group. I'll follow the instructions to submit our adoption story soon.

Looking forward to connecting with other Karmada users and contributing to the community!

zhengjr9:
@zhzhuang-zju Sure, I'd be happy to help with the backport! I will submit the PRs for release-1.14, release-1.15, and release-1.16.

Also, I'll check my email and reply to you soon.

PR #7203, opened 2026-02-15
fix: recompute FullyApplied condition inside retry loop to prevent st…

Ady0333:
Fix stale FullyApplied condition during concurrent rescheduling

**What type of PR is this?**

/kind bug

---

This PR fixes a status inconsistency where the `FullyApplied` condition could be incorrectly set to `True` during concurrent rescheduling.

In `AggregateResourceBindingWorkStatus` and `AggregateClusterResourceBindingWorkStatus`, the
`generateFullyAppliedCondition()` result was previously computed *outside* the
`retry.RetryOnConflict` → `UpdateStatus()` mutation closure.

Since `UpdateStatus()` internally re-fetches the latest binding before applying the mutation,
a concurrent scheduler update to `spec.clusters` could cause the condition to be evaluated
using a stale cluster list, while the status update was applied to a newer binding.

This could result in `FullyApplied=True` being written even though newly scheduled clusters
had not finished applying.

The fix moves the condition generation **inside** the mutation closure so it always evaluates
against the freshly re-fetched `binding.Spec`, ensuring correctness under concurrent updates.

---

- The change is minimal and localized to status update logic
- No behavior change outside concurrent rescheduling scenarios
- Existing tests pass; no unrelated files are touched

**Release-note**:
```NONE```

### DIFF
--- a/pkg/util/helper/workstatus.go
+++ b/pkg/util/helper/workstatus.go
@@ -71,13 +71,13 @@ func AggregateResourceBindingWorkStatus(
 		return err
 	}
 
-	fullyAppliedCondition := generateFullyAppliedCondition(binding.Spec, aggregatedStatuses)
-
 	var operationResult controllerutil.OperationResult
 	if err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
 		operationResult, err = UpdateStatus(ctx, c, binding, func() error {
 			binding.Status.AggregatedStatus = aggregatedStatuses
-			// set binding status with the newest condition
+			// set binding status with the newest condition, recompute using the
+			// refreshed binding.Spec to reflect any concurrent spec.clusters changes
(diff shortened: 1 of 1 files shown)


Ady0333:
@XiShanYongYe-Chang, the pr is ready please review it once you have time.

Happy to make any changes if needed!!

Ady0333:
Hello @XiShanYongYe-Chang 
Thanks for the clarification.

I see your point, if the informer cache is already updated when the controller processes the event, then the controller should not reach this path with a missing resource. In that case, it seems the NotFound scenario may indeed not occur in practice. 

Please let me know if you think this PR should be closed or if there is another situation where handling NotFound would still be useful.

XiShanYongYe-Chang:
I think we can close it.

PR #7221, opened 2026-02-26
introduces assigning binding list to avoid cache delay

zhzhuang-zju:
**What type of PR is this?**

/kind feature

This PR introduces a `BindingCache` within the `karmada-scheduler` to address a potential race condition and state inconsistency between successful API writes and Informer synchronization.

There is a non-negligible delay between when a ResourceBinding is successfully updated via the scheduler and when that update is reflected in the local Informer cache. If the scheduler processes subsequent ResourceBindings that depend on the state of the previously updated one during this synchronization window, it might make conflicting or incorrect decisions based on stale data.

The `BindingCache` acts as a temporary "source of truth" for committed but not yet synced ResourceBindings. It bridges this gap by providing the scheduler with immediate access to the most recently committed state, ensuring high consistency during rapid scheduling cycles.

Part of #7064

```release-note
`karmada-scheduler`: Improved scheduling consistency for workload affinity groups during rapid scheduling cycles by introducing a dedicated cache for recently committed ResourceBindings.
```

### DIFF
--- a/pkg/scheduler/cache/cache.go
+++ b/pkg/scheduler/cache/cache.go
@@ -17,13 +17,20 @@ limitations under the License.
 package cache
 
 import (
+	"maps"
+	"sync"
+
+	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/resourceversion"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
 
 	clusterv1alpha1 "github.com/karmada-io/karmada/pkg/apis/cluster/v1alpha1"
+	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
 	clusterlister "github.com/karmada-io/karmada/pkg/generated/listers/cluster/v1alpha1"
 	"github.com/karmada-io/karmada/pkg/scheduler/framework"
+	"github.com/karmada-io/karmada/pkg/util/names"
 )
 
 // Cache is an interface for scheduler internal cache.
@@ -36,18 +43,33 @@ type Cache interface {
 
 	// ResourceBindingIndexer returns the indexer for ResourceBindings, used for advanced scheduling logic.
 	ResourceBindingIndexer() cache.Indexer
+
+	// AssigningResourceBindings returns a cache for ResourceBindings that are in the "assigning" state.
+	// After the control plane generates scheduling results and commits them to the API server,
+	// there is a period before these results are fully propagated to and reflected in the member clusters.
+	// This cache stores these "assigning" bindings to provide the scheduler with a more accurate and
+	// immediate view of the intended state, enabling more precise decision-making during subsequent scheduling cycles.
+	AssigningResourceBindings() *AssigningResourceBindingCache
 }
 
 type schedulerCache struct {
 	clusterLister          clusterlister.ClusterLister
 	resourceBindingIndexer cache.Indexer
+
+	assigningResourceBindings *AssigningResourceBindingCache
 }
+
+// Check if our schedulerCache implements necessary interface
+var _ Cache = &schedulerCache{}
 
 // NewCache instantiates a cache used only by scheduler.
 func NewCache(clusterLister clusterlister.ClusterLister, resourceBindingIndexer cache.Indexer) Cache {
 	return &schedulerCache{
 		clusterLister:          clusterLister,
 		resourceBindingIndexer: resourceBindingIndexer,
+		assigningResourceBindings: &AssigningResourceBindingCache{
+			items: make(map[string]*workv1alpha2.ResourceBinding),
+		},
 	}
 }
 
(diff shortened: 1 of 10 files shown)


pkg/scheduler/scheduler.go
```diff
@@ -680,11 +680,14 @@ func (s *Scheduler) patchScheduleResultForResourceBinding(oldBinding *workv1alph
...
 
-	_, err = s.KarmadaClient.WorkV1alpha2().ResourceBindings(newBinding.Namespace).Patch(context.TODO(), newBinding.Name, types.MergePatchType, patchBytes, metav1.PatchOptions{})
+	result, err := s.KarmadaClient.WorkV1alpha2().ResourceBindings(newBinding.Namespace).Patch(context.TODO(), newBinding.Name, types.MergePatchType, patchBytes, metav1.PatchOptions{})
 	if err != nil {
 		klog.Errorf("Failed to patch schedule to ResourceBinding(%s/%s): %v", oldBinding.Namespace, oldBinding.Name, err)
 		return err
 	}
+	if features.FeatureGate.Enabled(features.WorkloadAffinity) {
+		s.schedulerCache.AssigningBindings().Add(result)
+	}
```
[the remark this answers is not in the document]
zhzhuang-zju: This is a good find. It doesn't affect the functionality itself, but it may leave residual bindings in the cache indefinitely.
Adding an expiration mechanism is one approach. Another is to put the `patch` operation and `Add` inside the `BindingCache` lock.

pkg/scheduler/framework/interface.go
```diff
@@ -75,6 +75,15 @@ type FilterContext struct {
...
 	ResourceBindingIndexer cache.Indexer
+
+	// AssigningBindingGetter is a getter for the ResourceBindings that have been successfully committed to the API server
+	// but are likely not yet visible in the Informer's local cache.
+	AssigningBindingGetter BindingGetter
+}
+
+// BindingGetter is an interface for getting ResourceBindings.
+type BindingGetter interface {
+	GetBindings() map[string]*workv1alpha2.ResourceBinding
```
RainbowMango: Remind me of the benefit of introducing the interface?

pkg/scheduler/framework/plugins/workloadaffinity/workload_affinity.go
```diff
@@ -87,32 +87,27 @@ func (p *WorkloadAffinity) FilterWithContext(filterCtx *framework.FilterContext)
...
-		if rb == nil || rb.Spec.WorkloadAffinityGroups == nil || rb.GetNamespace() != bindingSpec.Resource.Namespace {
-			continue
-		}
-
-		if rb.Spec.Clusters == nil {
-			continue
-		}
-
-		if rb.GetName() == bindingName {
+		if assignBinding, ok := assignBindings[names.NamespacedKey(rb.Namespace, rb.Name)]; ok && assignBinding.Generation > rb.Generation {
```
RainbowMango: I guess you want to make sure the assigning ResourceBinding is newer than the one from cache.
But why use `generation` instead of `resourceVersion`?

pkg/scheduler/cache/cache.go
```diff
@@ -36,18 +44,35 @@ type Cache interface {
 
 	// ResourceBindingIndexer returns the indexer for ResourceBindings, used for advanced scheduling logic.
 	ResourceBindingIndexer() cache.Indexer
+
+	// AssigningBindings returns a cache of ResourceBindings that have been successfully committed to the API server
+	// but are likely not yet visible in the Informer's local cache.
+	// This "temporary truth" store prevents the scheduler from making conflicting decisions
+	// in rapid succession before the Informer catches up.
+	AssigningBindings() *BindingCache
```
RainbowMango:
```suggestion
	AssigningResourceBindings() *BindingCache
```
I'd prefer the full name for clarity.

pkg/scheduler/cache/cache.go
```diff
@@ -36,18 +44,35 @@ type Cache interface {
...
 
 type schedulerCache struct {
 	clusterLister          clusterlister.ClusterLister
 	resourceBindingIndexer cache.Indexer
+
+	assigningBindings *BindingCache
+}
+
+// BindingCache acts as a buffer for the most recently updated ResourceBindings.
+type BindingCache struct {
```
RainbowMango:
```suggestion
type AssigningResourceBindingCache struct {
```
ditto.

In addition, please move down the definition to where its methods are declared.

pkg/scheduler/cache/cache.go
```diff
@@ -82,6 +107,54 @@ func (c *schedulerCache) Snapshot() Snapshot {
...
+		return
+	}
+
+	if ca.Generation <= binding.Generation {
+		delete(b.items, names.NamespacedKey(binding.Namespace, binding.Name))
+	}
+}
+
+// OnBindingDelete is called when a delete event for a ResourceBinding is received from the Informer.
+func (b *BindingCache) OnBindingDelete(binding *workv1alpha2.ResourceBinding) {
```
RainbowMango:
```suggestion
func (b *BindingCache) Delete(binding *workv1alpha2.ResourceBinding) {
```

pkg/scheduler/cache/cache.go
```diff
@@ -82,6 +107,54 @@ func (c *schedulerCache) Snapshot() Snapshot {
...
+	defer b.Unlock()
+
+	ca, ok := b.items[names.NamespacedKey(binding.Namespace, binding.Name)]
+	if !ok {
+		return
+	}
+
+	if ca.Generation <= binding.Generation {
+		delete(b.items, names.NamespacedKey(binding.Namespace, binding.Name))
+	}
```
RainbowMango: Can you explain this logic?

I think the item should exist until the ResourceBinding has been synced to member clusters, or make sure the informer cache gets the synced status.

pkg/scheduler/framework/interface.go
zhzhuang-zju: Just my thoughts:
1. The interface offers better extensibility, allowing more operations related to the bindingcache to be introduced via the interface later.
2. Align the operations of fetching bindings from the informer cache and from the bindingCache as much as possible. However, since the current scheduler is serial, this has little impact.

pkg/scheduler/framework/plugins/workloadaffinity/workload_affinity.go
zhzhuang-zju: Since the fields the scheduler cares about are under `spec`, using `generation` is feasible in effect.

pkg/scheduler/cache/cache.go
zhzhuang-zju: > or make sure the informer cache gets the synced status.

This logic will be called in the `UpdateFunc` of the event handler.
If the `Generation` of the binding in the `bindingcache` is less than or equal to the binding in the event handler, it means the informer cache has obtained the synced status, so it can be deleted from the `bindingcache`.

pkg/scheduler/cache/cache.go
```diff
@@ -17,15 +17,26 @@ limitations under the License.
...
 
 	clusterv1alpha1 "github.com/karmada-io/karmada/pkg/apis/cluster/v1alpha1"
+	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
 	clusterlister "github.com/karmada-io/karmada/pkg/generated/listers/cluster/v1alpha1"
 	"github.com/karmada-io/karmada/pkg/scheduler/framework"
+	"github.com/karmada-io/karmada/pkg/util/names"
 )
 
+// Check if our schedulerCache implements necessary interface
+var _ Cache = &schedulerCache{}
```
RainbowMango:
```diff
--- a/pkg/scheduler/cache/cache.go                                       
+++ b/pkg/scheduler/cache/cache.go                                       
@@ -34,9 +34,6 @@ import (                                               
        "github.com/karmada-io/karmada/pkg/util/names"                   
 )

-// Check if our schedulerCache implements necessary interface
-var _ Cache = &schedulerCache{}
-
 // Cache is an interface for scheduler internal cache.
 type Cache interface {
        AddCluster(cluster *clusterv1alpha1.Cluster)
@@ -63,6 +60,9 @@ type schedulerCache struct {
        assigningResourceBindings *AssigningResourceBindingCache
 }

+// Check if our schedulerCache implements necessary interface
+var _ Cache = &schedulerCache{}
+
 // NewCache instantiates a cache used only by scheduler.
 func NewCache(clusterLister clusterlister.ClusterLister, resourceBindingIndexer cache.Indexer) Cache {
        return &schedulerCache{
```
Move it to right after `type schedulerCache struct`.

pkg/scheduler/cache/cache.go
```diff
@@ -82,6 +105,80 @@ func (c *schedulerCache) Snapshot() Snapshot {
...
+		// should not happen since Kubernetes generates the resource versions, but log just in case
+		klog.Errorf("Failed to compare resource versions for ResourceBinding %s/%s: %v", binding.Namespace, binding.Name, err)
+		return
+	}
+
+	if compare == 1 {
+		return
+	}
+
+	condition := meta.FindStatusCondition(binding.Status.Conditions, workv1alpha2.FullyApplied)
```
RainbowMango: Try `meta.IsStatusConditionTrue()`?

pkg/scheduler/event_handler.go
```diff
@@ -200,6 +206,31 @@ func (s *Scheduler) onResourceBindingUpdate(old, cur any) {
 	metrics.CountSchedulerBindings(metrics.BindingUpdate)
 }
 
+// onResourceBindingDelete is called when a delete event for a ResourceBinding is received from the Informer.
+func (s *Scheduler) onResourceBindingDelete(obj any) {
+	if !features.FeatureGate.Enabled(features.WorkloadAffinity) {
+		return
+	}
```
RainbowMango: Sorry, I missed this from the previous review: I would prefer the following pattern:
```golang
	if features.FeatureGate.Enabled(features.WorkloadAffinity) {

		var binding *workv1alpha2.ResourceBinding
		switch t := obj.(type) {
		case *workv1alpha2.ResourceBinding:
			binding = t
		case cache.DeletedFinalStateUnknown:
			var ok bool
			binding, ok = t.Obj.(*workv1alpha2.ResourceBinding)
			if !ok {
				klog.Errorf("cannot convert to workv1alpha2.ResourceBinding: %v", t.Obj)
				return
			}
		default:
			klog.Errorf("cannot convert to workv1alpha2.ResourceBinding: %v", t)
			return
		}

		s.schedulerCache.AssigningResourceBindings().OnBindingDelete(binding)
	}
```
Ensure the whole block is protected and owned by the feature.

pkg/scheduler/scheduler.go
```diff
@@ -680,11 +680,14 @@ func (s *Scheduler) patchScheduleResultForResourceBinding(oldBinding *workv1alph
...
 	}
 
-	_, err = s.KarmadaClient.WorkV1alpha2().ResourceBindings(newBinding.Namespace).Patch(context.TODO(), newBinding.Name, types.MergePatchType, patchBytes, metav1.PatchOptions{})
+	result, err := s.KarmadaClient.WorkV1alpha2().ResourceBindings(newBinding.Namespace).Patch(context.TODO(), newBinding.Name, types.MergePatchType, patchBytes, metav1.PatchOptions{})
 	if err != nil {
 		klog.Errorf("Failed to patch schedule to ResourceBinding(%s/%s): %v", oldBinding.Namespace, oldBinding.Name, err)
 		return err
 	}
+	if features.FeatureGate.Enabled(features.WorkloadAffinity) {
+		s.schedulerCache.AssigningResourceBindings().Add(result)
```
RainbowMango, line 691: https://github.com/karmada-io/karmada/pull/7221/changes#r2857176604
This is resaonble concern, how about leave a comment or todo here?

zhzhuang-zju: Added, please take a look~

PR #7226, opened 2026-02-27
fix: recompute FullyApplied condition inside RetryOnConflict to avoid stale spec evaluation

Ady0333:
/kind cleanup

This PR fixes an issue where the `FullyApplied` condition of a `ResourceBinding` (and `ClusterResourceBinding`) could be computed using a stale snapshot of `binding.Spec.Clusters`.

Previously, `generateFullyAppliedCondition()` was executed before entering the `RetryOnConflict` loop. However, `UpdateStatus()` re-fetches the binding from the API server on each retry. If the scheduler updates `Spec.Clusters` between the initial read and a retry (for example when a new cluster is added by a PropagationPolicy update), the retry would write a condition that was evaluated against the old cluster set.

This could temporarily report `FullyApplied=True` even though a newly added cluster had not yet received or applied its `Work`.

To avoid this inconsistency, the condition is now recomputed inside the mutation closure so it is always evaluated against the latest `binding.Spec` fetched during the retry.

---

The change is intentionally minimal. The only modification is moving the call to `generateFullyAppliedCondition()` inside the retry mutation closure in:

- `AggregateResourceBindingWorkStatus`
- `AggregateClusterResourceBindingWorkStatus`

This ensures the condition is evaluated against the latest `Spec.Clusters` after each retry fetch.

---

**Fixes issue : #7227** 

---

```release-note
karmada-controller-manager: Fixed an issue where the FullyApplied condition of ResourceBinding could be incorrectly reported during RetryOnConflict retries when the cluster set changed.

### DIFF
--- a/pkg/util/helper/workstatus.go
+++ b/pkg/util/helper/workstatus.go
@@ -65,20 +65,17 @@ func AggregateResourceBindingWorkStatus(
 	if err != nil {
 		return err
 	}
-
 	aggregatedStatuses, err := assembleWorkStatus(workList.Items, binding.Spec.Resource)
 	if err != nil {
 		return err
 	}
-
-	fullyAppliedCondition := generateFullyAppliedCondition(binding.Spec, aggregatedStatuses)
 
 	var operationResult controllerutil.OperationResult
 	if err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
 		operationResult, err = UpdateStatus(ctx, c, binding, func() error {
 			binding.Status.AggregatedStatus = aggregatedStatuses
 			// set binding status with the newest condition
-			meta.SetStatusCondition(&binding.Status.Conditions, fullyAppliedCondition)
+			meta.SetStatusCondition(&binding.Status.Conditions, generateFullyAppliedCondition(binding.Spec, aggregatedStatuses))
 			return nil
 		})
 		return err
@@ -108,20 +105,17 @@ func AggregateClusterResourceBindingWorkStatus(
 	if err != nil {
 		return err
 	}
-
 	aggregatedStatuses, err := assembleWorkStatus(workList.Items, binding.Spec.Resource)
 	if err != nil {
 		return err
 	}
-
-	fullyAppliedCondition := generateFullyAppliedCondition(binding.Spec, aggregatedStatuses)
 
 	var operationResult controllerutil.OperationResult
 	if err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
 		operationResult, err = UpdateStatus(ctx, c, binding, func() error {
 			binding.Status.AggregatedStatus = aggregatedStatuses
 			// set binding status with the newest condition
-			meta.SetStatusCondition(&binding.Status.Conditions, fullyAppliedCondition)
+			meta.SetStatusCondition(&binding.Status.Conditions, generateFullyAppliedCondition(binding.Spec, aggregatedStatuses))
 			return nil
 		})
 		return err

### CODE REVIEW

**Comment by Ady0333** [comment]
@XiShanYongYe-Chang , please review this pr and let me know if any change is needed!

**Comment by XiShanYongYe-Chang** [comment]
Thanks @Ady0333 
/assign

Then, we can first submit an issue for discussion. After the discussion is clear, we will submit a PR.

**Comment by Ady0333** [comment]
> Thanks @Ady0333 /assign
> 
> Then, we can first submit an issue for discussion. After the discussion is clear, we will submit a PR.

Okay I will first submit an issue.

**Comment by Ady0333** [comment]
@XiShanYongYe-Chang 
I've updated the PR to recompute `aggregatedStatuses` inside the `RetryOnConflict` block as well, so both the aggregated status and `FullyApplied` condition are evaluated consistently on each retry.

Could you please take another look?

File: pkg/util/helper/workstatus.go [hunk 1]
```diff
@@ -61,24 +61,20 @@ func AggregateResourceBindingWorkStatus(
...
-	fullyAppliedCondition := generateFullyAppliedCondition(binding.Spec, aggregatedStatuses)
-
 	var operationResult controllerutil.OperationResult
-	if err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
+	if err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		workList, err := GetWorksByBindingID(ctx, c, binding.Labels[workv1alpha2.ResourceBindingPermanentIDLabel], true)
+		if err != nil {
+			return err
+		}
+		aggregatedStatuses, err := assembleWorkStatus(workList.Items, binding.Spec.Resource)
```
XiShanYongYe-Chang: I understand that this part does not need to be recalculated, right?

Ady0333: @XiShanYongYe-Chang 
Yes, you're right. That part does not strictly need to be recalculated. I moved it inside the retry block mainly for consistency with the condition computation, but I'm happy to move it back outside if you prefer.

XiShanYongYe-Chang: I think we can move only the parts that are necessary.

XiShanYongYe-Chang:
Can you help squash the commits into one commit in the meantime?

Ady0333:
> Can you help squash the commits into one commit in the meantime?

Yes sir ! I will squash the commits in the meantime.

pkg/util/helper/workstatus.go
XiShanYongYe-Chang: Hi @Ady0333 This seems to have been forgotten.

Ady0333: Hi @XiShanYongYe-Chang, sorry for the delay, I ran into some issues.

You're right, I accidentally moved more logic than necessary. I'll update the PR.
Thanks for pointing it out!

pkg/util/helper/workstatus.go
```diff
@@ -65,20 +65,18 @@ func AggregateResourceBindingWorkStatus(
...
 	if err != nil {
 		return err
 	}
 
-	fullyAppliedCondition := generateFullyAppliedCondition(binding.Spec, aggregatedStatuses)
-
 	var operationResult controllerutil.OperationResult
-	if err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
+	if err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		var err error
```
XiShanYongYe-Chang: I understand that there is no need to make any changes here.

Ady0333: Thanks for pointing that out! You're right, the change there isn’t necessary. I'll revert that part and update the PR.

pkg/util/helper/workstatus.go
```diff
@@ -65,20 +65,17 @@ func AggregateResourceBindingWorkStatus(
...
 	aggregatedStatuses, err := assembleWorkStatus(workList.Items, binding.Spec.Resource)
 	if err != nil {
 		return err
 	}
 
-	fullyAppliedCondition := generateFullyAppliedCondition(binding.Spec, aggregatedStatuses)
-
 	var operationResult controllerutil.OperationResult
-	if err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
+	if err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
XiShanYongYe-Chang: I understand that err doesn't need to redefine the variable, right?

Ady0333: Thanks for pointing that out. I've restored it to use the existing err variable.

pkg/util/helper/workstatus.go
```diff
@@ -108,20 +105,18 @@ func AggregateClusterResourceBindingWorkStatus(
...
+		var err error
```
XiShanYongYe-Chang: It seems like this was forgotten to be modified.

Ady0333: I have modified it now.

PR #5579, opened 2024-09-21
Add helm index for v1.10.0 and v1.11.0

chaosi-zju:
**What type of PR is this?**

/kind cleanup

**What this PR does / why we need it**:

Add helm index for v1.10.0 and v1.11.0

these two new version can't be searched by helm, as you see:

```shell
$ helm search repo karmada-charts -l
NAME                            CHART VERSION   APP VERSION     DESCRIPTION                      
karmada-charts/karmada          v1.9.0          latest          A Helm chart for karmada         
karmada-charts/karmada          v1.8.0          v1.1.0          A Helm chart for karmada         
karmada-charts/karmada          v1.7.1          v1.1.0          A Helm chart for karmada         
karmada-charts/karmada          v1.7.0          v1.1.0          A Helm chart for karmada         
karmada-charts/karmada          v1.6.2          v1.1.0          A Helm chart for karmada         
karmada-charts/karmada          v1.6.1          v1.1.0          A Helm chart for karmada         
karmada-charts/karmada          v1.6.0          v1.1.0          A Helm chart for karmada         
karmada-charts/karmada          v1.5.0          v1.1.0          A Helm chart for karmada         
karmada-charts/karmada          v1.4.0          v1.1.0          A Helm chart for karmada         
karmada-charts/karmada          v1.3.0          v1.1.0          A Helm chart for karmada         
karmada-charts/karmada          v1.2.2          v1.1.0          A Helm chart for karmada         
karmada-charts/karmada          v1.2.1          v1.1.0          A Helm chart for karmada         
karmada-charts/karmada          v1.2.0          v1.1.0          A Helm chart for karmada         
karmada-charts/karmada-operator v1.8.0          v1.1.0          A Helm chart for karmada-operator
```

**Which issue(s) this PR fixes**:

Fixes part of #5484

**Special notes for your reviewer**:

1. generate methods: https://github.com/karmada-io/karmada/pull/4051#discussion_r1324310397

2. test report on my local branch:

```shell
$ helm repo add my-branch-karmada https://raw.githubusercontent.com/chaosi-zju/karmada/chart-index/charts
"my-branch-karmada" has been added to your repositories

$ helm repo list        
NAME                            URL                                                                    
...
my-branch-karmada               https://raw.githubusercontent.com/chaosi-zju/karmada/chart-index/charts

$ helm search repo my-branch-karmada -l
NAME                                    CHART VERSION   APP VERSION     DESCRIPTION                      
[...]

PR #7247, opened 2026-03-03
CI: auto update helm index by github action

zhzhuang-zju:
**What type of PR is this?**
/kind cleanup

The `index` files for chart `karmada` and chart `karmada-operator` need to be updated with each new release. To reduce repetitive manual work, the helm update index operation is automated via GitHub Actions.
Trigger Methods:
- Automatically triggered upon the release of a new version
- Manually triggered

Fixes #

```release-note

```


zhzhuang-zju:
Local test results: https://github.com/zhzhuang-zju/karmada/pull/35
/cc @seanlaii @RainbowMango

.github/workflows/update-helm-index.yml
```diff
@@ -0,0 +1,101 @@
...
+        required: true
+        type: string
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  update-index-pr:
+    runs-on: ubuntu-latest
```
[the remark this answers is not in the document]
RainbowMango: +1

.github/workflows/update-helm-index.yml
```diff
@@ -0,0 +1,101 @@
...
+        description: 'The version to update (e.g., v1.17.0)'
+        required: true
+        type: string
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  update-index-pr:
```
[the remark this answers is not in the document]
zhzhuang-zju: This action should not be restricted only to the karmada repository; forked repositories can also be used for testing.

.github/workflows/update-helm-index.yml
```diff
@@ -0,0 +1,101 @@
+name: Update Helm Chart Index (Pull Request)
+env:
+  HELM_VERSION: v3.17.3
+on:
+  release:
+    types: [published]
```
seanlaii: To confirm my understanding: the current manual process is to wait until the release is fully published (with chart assets uploaded by `release.yml`) before updating `charts/index.yaml`. If that's the case, should the automation preserve the same ordering?
Currently, both this workflow and `release.yml` share the same trigger (`release: types: [published]`), so they run concurrently. There is no guarantee that `release.yml's release-charts` job finishes uploading the `.tgz` assets before this workflow attempts to download them, which might result in intermittent 404 failures on the curl step.
Maybe we should trigger this workflow after the Build Release workflow completes (e.g., using workflow_run)?

zhzhuang-zju: This is noteworthy. I will investigate how to ensure the sequential execution order of workflows, such as the `workflow_run` you mentioned.

zhzhuang-zju: @seanlaii Thanks for your advice. It’s done now. This CI will only run either manually or when the CI "Build Release" completes successfully.

RainbowMango (commented):
/assign

.github/workflows/update-helm-index.yml
```diff
@@ -0,0 +1,106 @@
...
+          # for `git describe --tags` in Makefile.
+          fetch-depth: 0
+      - name: Set Version
+        run: |
+          if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
+            echo "VERSION=${{ inputs.version }}" >> $GITHUB_ENV
+          else
+            LATEST_TAG=`gh api repos/karmada-io/karmada/releases/latest | jq -r '.tag_name'`
+            echo "VERSION=$LATEST_TAG" >> $GITHUB_ENV
+          fi
```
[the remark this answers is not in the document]
zhzhuang-zju: make sense

.github/workflows/update-helm-index.yml
```diff
@@ -0,0 +1,106 @@
...
+        run: |
+          git config --global user.name "github-actions"
+          git config --global user.email "actions@github.com"
+          BRANCH="auto-action-helm-index-${{ github.sha }}-${{ github.run_attempt }}"
+          git checkout -b $BRANCH
+          git add charts/index.yaml
+          if git diff --cached --quiet; then
+            echo "No changes to commit"
+            exit 0
+          fi
```
RainbowMango, line 101: When does this situation (`no changes to commit`) occur? And why does it still return 0? This is the error code indicating normal status, right?

.github/workflows/update-helm-index.yml
```diff
@@ -0,0 +1,106 @@
...
+          sed -i '/^  common:$/,/^  [a-zA-Z]/{/^  common:/d; /^  [a-zA-Z]/!d}' charts/index.yaml
+
+      - name: commit changes and create pull request
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          git config --global user.name "github-actions"
+          git config --global user.email "actions@github.com"
+          BRANCH="auto-action-helm-index-${{ github.sha }}-${{ github.run_attempt }}"
+          git checkout -b $BRANCH
```
RainbowMango, line 96: What's the name of the `BRANCH`? I didn't find the definition.

.github/workflows/update-helm-index.yml
```diff
@@ -0,0 +1,106 @@
...
+            --head "$BRANCH" \
+            --title "Update Helm Chart Index for ${{ env.VERSION }}" \
+            --body "$(cat <<EOF
+          /kind feature
+
+          Automated Helm chart index update for version \`${{ env.VERSION }}\`.
+
+          - Downloaded chart assets from release \`${{ env.VERSION }}\`.
+          - Merged new versions into \`charts/index.yaml\`.
+          - Cleaned up auto-generated \`common\` entries.
```
RainbowMango:
```suggestion
```

.github/workflows/update-helm-index.yml
```diff
@@ -0,0 +1,106 @@
...
+          # 0 indicates all history for all branches and tags.
+          # for `git describe --tags` in Makefile.
+          fetch-depth: 0
+      - name: Set Version
+        run: |
+          if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
+            echo "VERSION=${{ inputs.version }}" >> $GITHUB_ENV
+          else
+            LATEST_TAG=`gh api repos/karmada-io/karmada/releases/latest | jq -r '.tag_name'`
+            echo "VERSION=$LATEST_TAG" >> $GITHUB_ENV
```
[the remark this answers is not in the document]
RainbowMango: +1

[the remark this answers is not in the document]
zhzhuang-zju: This CI is auto-triggered only if:
- The triggering version matches the latest tag of the repo.
- It is a minor version.

.github/workflows/update-helm-index.yml
zhzhuang-zju: Since this CI can also be triggered manually—possibly for debugging purposes using an older version, in this case, I think we can treat it as normal

.github/workflows/update-helm-index.yml
zhzhuang-zju:
```yaml
BRANCH="auto-action-helm-index-${{ github.sha }}-${{ github.run_attempt }}"
```

RainbowMango: Oh, my god~ I can't believe how I missed that.

zhzhuang-zju:
# Helm Chart Investigation

Q: When merging an old version with `helm repo index`, will the entries in `index.yaml` be sorted by version?
A: Yes. According to local testing, they will be sorted by version automatically, so manual sorting is not required.

Q: What permissions are required for manually running a workflow?
A: [Write access to the repository is required to perform these steps.](https://docs.github.com/en/actions/how-tos/manage-workflow-runs/manually-run-a-workflow)

Q: Do we need to support patch versions?
A: Yes, we can support them.
First, we previously only updated for minor versions, but I did not find any specific reason for this. I only found one related [comment](https://github.com/karmada-io/karmada/pull/5579#discussion_r1770877101); perhaps @Poor12 can clarify?

The [official documentation](https://helm.sh/zh/docs/topics/chart_repository/#the-index-file) describes the index file as:
*A chart repository consists of packaged charts and a special file called index.yaml which contains an index of all of the charts in the repository.*

This does not restrict support to minor versions only, and patch versions are also officially maintained releases. Therefore, I think we should also include patch versions of the charts.

Second, the `index.yaml` files of multiple Helm charts I use include patch versions:

```bash
argo-cd                 https://argoproj.github.io/argo-helm
clowarden               https://cncf.github.io/clowarden/
prometheus-community    https://prometheus-community.github.io/helm-charts
```

Q: Should we remove `common`?
A: I think it is necessary to remove it.
If not removed, the output of `helm search repo my-karmada -l` will be:

```bash
$ helm search repo my-karmada -l
NAME                            CHART VERSION   APP VERSION     DESCRIPTION
my-karmada/common               2.27.0          2.27.0          A Library Helm Chart for grouping common logic ...
my-karmada/karmada              v1.16.0         v1.16.0         A Helm chart for karmada
```

The `common` chart is **not** generated by Karmada. If you need to check chart dependencies, you can use the command: `helm inspect chart my-karmada/karmada --version` or check them in `chart.yaml`, rather than keeping it in `index.yaml`.

/cc @RainbowMango @seanlaii If you agree, I can make the changes according to the above content.

RainbowMango:
> The common chart is not generated by Karmada. If you need to check chart dependencies, you can use the command: helm inspect chart my-karmada/karmada --version or check them in chart.yaml, rather than keeping it in index.yaml

OK.

RainbowMango:
> /cc @RainbowMango @seanlaii If you agree, I can make the changes according to the above content.

Can you clarify what changes you are going to make?

zhzhuang-zju:
> Can you clarify what changes you are going to make?

Currently, we only add the index when releasing minor versions. Moving forward, we could include the index for both minor and patch releases:
- For minor versions: The CI will add the index automatically.
- For patch versions: The CI should be triggered manually. Additionally, multiple patch versions can be provided as input at once

Since multiple patch versions are often released at the same time, automatic triggers would create too many PRs and likely lead to conflicts.

RainbowMango:
OK. I think it would be greate to support patch releases.

zhzhuang-zju:
Done~ And @RainbowMango, you can see the effect in the link: https://github.com/zhzhuang-zju/karmada/pull/39

seanlaii:
Thank you!

RainbowMango:
@zhzhuang-zju 
This PR brings a warning at https://securityscorecards.dev/viewer/?uri=github.com/karmada-io/karmada in `Token-Permissions` category, which shows:
```
Warn: topLevel 'contents' permission set to 'write': .github/workflows/update-helm-index.yml:18
Info: no jobLevel write permissions found
```
Please check the documentation here for remediation steps: https://github.com/ossf/scorecard/blob/c0df03399ddd8c3345b62ff3df813f3cbbe5bf42/docs/checks.md#token-permissions

zhzhuang-zju:
> Please check the documentation here for remediation steps: https://github.com/ossf/scorecard/blob/c0df03399ddd8c3345b62ff3df813f3cbbe5bf42/docs/checks.md#token-permissions

Thanks for the reminder! Will fix this soon

PR #7249, opened 2026-03-03
Fix/graceful eviction race condition

Ady0333:
**What type of PR is this?**
/kind bug

---

This PR adds optimistic locking to the graceful eviction controllers (both RB and CRB variants) when patching `Spec.GracefulEvictionTasks`.

Previously, the controller used `client.MergeFrom()` which generates a JSON merge patch that:
- Replaces the entire `GracefulEvictionTasks` array, and
- Does not include `metadata.resourceVersion`, so the API server performs no conflict detection.

If another controller (e.g., taint-manager or application-failover-controller) added a new eviction task via `Update()` between the graceful eviction controller’s `Get` and `Patch`, the new task could be silently overwritten.

By switching to `client.MergeFromWithOptions(..., client.MergeFromWithOptimisticLock{})`, the patch now includes the current `resourceVersion`. If the binding was modified concurrently, the API server returns a 409 Conflict, and the controller requeues and retries with the latest state.

This prevents silent eviction task loss under concurrent failover scenarios.

---

The change is minimal (two lines across RB and CRB controllers) and does not alter business logic.  
Existing error handling already propagates conflicts correctly, so no additional retry logic is required.

---

No

---

```release-note
karmada-controller-manager: Fixed an issue where concurrent updates to GracefulEvictionTasks could be silently overwritten, potentially leading to missed eviction tasks during failover.
```

### DIFF
--- a/pkg/controllers/gracefuleviction/crb_graceful_eviction_controller.go
+++ b/pkg/controllers/gracefuleviction/crb_graceful_eviction_controller.go
@@ -88,7 +88,7 @@ func (c *CRBGracefulEvictionController) syncBinding(ctx context.Context, binding
 		return nextRetry(keptTask, c.GracefulEvictionTimeout, metav1.Now().Time), nil
 	}
 
-	objPatch := client.MergeFrom(binding)
+	objPatch := client.MergeFromWithOptions(binding, client.MergeFromWithOptimisticLock{})
 	modifiedObj := binding.DeepCopy()
 	modifiedObj.Spec.GracefulEvictionTasks = keptTask
 	err := c.Client.Patch(ctx, modifiedObj, objPatch)
--- a/pkg/controllers/gracefuleviction/rb_graceful_eviction_controller.go
+++ b/pkg/controllers/gracefuleviction/rb_graceful_eviction_controller.go
@@ -88,7 +88,7 @@ func (c *RBGracefulEvictionController) syncBinding(ctx context.Context, binding 
 		return nextRetry(keptTask, c.GracefulEvictionTimeout, metav1.Now().Time), nil
 	}
 
-	objPatch := client.MergeFrom(binding)
+	objPatch := client.MergeFromWithOptions(binding, client.MergeFromWithOptimisticLock{})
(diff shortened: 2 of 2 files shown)


Ady0333:
@XiShanYongYe-Chang 
Just to clarify how this differs from the previous graceful-eviction fixes:

Earlier issues were local logic problems within a single controller (example: timeout leakage or stale condition handling). This PR addresses a concurrency issue where multiple controllers can update Spec.GracefulEvictionTasks simultaneously. By adding optimistic locking, we prevent silent overwrites during concurrent failover operations.

XiShanYongYe-Chang:
When reporting a bug, remember to first submit an issue, discuss and confirm it with the community members as a bug, and then submit a PR.

> Earlier issues were local logic problems within a single controller (example: timeout leakage or stale condition handling). This PR addresses a concurrency issue where multiple controllers can update Spec.GracefulEvictionTasks simultaneously. By adding optimistic locking, we prevent silent overwrites during concurrent failover operations.

When analyzing the problem, you need to combine the actual service. Currently, the controller does not encounter a scenario where the GracefulEvictionTasks field of the binding object is concurrently updated.

Ady0333:
Hi @XiShanYongYe-Chang,

Sorry for submitting the PR directly without first opening an issue, that was my mistake.

Regarding the concurrency case, my concern was a potential race between the graceful-eviction controller (removing tasks) and the taint-manager or failover controller (adding tasks), since they update the same field independently. I assumed they could overlap under certain failover conditions. If in actual service flow these updates don’t happen concurrently, then I understand the fix may be unnecessary. 

I’m happy to open an issue first for further discussion or close this PR if you prefer.

XiShanYongYe-Chang:
Thanks @Ady0333 ~
I think we can close this PR.

XiShanYongYe-Chang:
I have been organizing some to-do items recently; you might want to keep an eye on them.

Ady0333:
> I have been organizing some to-do items recently; you might want to keep an eye on them.

Yes sir 👍

PR #7255, opened 2026-03-06
Adopt openapi-gen changes

zhzhuang-zju:
**What type of PR is this?**
/kind feature

After upgrading the Kubernetes dependency to 1.35, due to the change in `openapi-gen`: it moves from a reflection‑based approach to OpenAPI model naming to a declarative approach.  

To ensure **all OpenAPI model names remain the same**, we need to:

1. Use `+k8s:openapi-model-package` to declare model package names.
2. When running the `openapi-gen` command, generate `zz_generated.model_name.go` files via the `--output-model-name-file` flag.

Fixes #

The model names are used in the OpenAPI spec as schema references.

If model package names are not declared, the OpenAPI model names will be the Go type names, for example:
`github.com~1karmada-io~1karmada~1pkg~1apis~1apps~1v1alpha1.WorkloadRebalancer`.

If `--output-model-name-file` is specified, the OpenAPI model names will be fully qualified model names, for example:
`com.github.karmada-io.karmada.pkg.apis.apps.v1alpha1.WorkloadRebalancer`.

Currently, many related tools (such as `hack/update-swagger-docs.sh` and the website API Reference generation tool) use the latter format.
Therefore, we need to specify `--output-model-name-file`.

```release-note

```


XiShanYongYe-Chang:
Thanks, could you briefly explain in the PR description why this needs to be done?

hack/update-codegen.sh
```diff
@@ -208,6 +210,12 @@ openapi-gen \
   "k8s.io/apimachinery/pkg/runtime" \
   "k8s.io/apimachinery/pkg/version"
 
+# Running command `openapi-gen --output-model-name-file "zz_generated.model_name.go"` generates a zz_generated.model_name.go file
+# for every input package containing the +k8s:openapi-model-package tag. This results in overwriting existing files in
+# the vendor directory, such as vendor/k8s.io/metrics/pkg/apis/metrics/v1beta1/zz_generated.model_name.go. To avoid
+# unnecessary modifications, changes within the vendor directory are discarded.
+git checkout -- vendor
```
zhzhuang-zju: **A known issue**: The OpenAPI model name generator attempts to write files to its dependency packages (e.g., Go modules).
**Issue link**: https://github.com/kubernetes/kube-openapi/issues/571
**Fixed by**: https://github.com/kubernetes/kube-openapi/pull/573, which adds the --readonly-pkg flag to skip read-only dependency packages. This fix was not backported to the k8s `release-1.35` branch.
The current implementation is a workaround but functional. If there are no other issues, I can update the relevant comments.

cc @RainbowMango @XiShanYongYe-Chang

RainbowMango: Great to know this issue has been fixed.

Another choice to update kube-openapi to `v0.0.0-20260304202019-5b3e3fdb0acf`, which included the fix.
Can you have a try?

zhzhuang-zju: Sure, will be back with the test result soon

RainbowMango: It should be ok, only one commit ahead of the current version `v0.0.0-20260127142750-a19766b6e2d4`.

zhzhuang-zju: done, thx~

zhzhuang-zju:
@RainbowMango All comments fixed, please take a look~

PR #7267, opened 2026-03-08
hack/util.sh: auto-detect macOS host IP instead of interactive prompt

### LINKED ISSUE #7257: [Enhancement] Automate macOS IP detection in local-up script
# Current Behavior
When running `hack/local-up-karmada.sh` on macOS, the script pauses execution and prompts the user to manually confirm or enter an IP address:
```
Detected that you are installing Karmada on macOS

It needs a Macintosh IP address to bind Karmada API Server(port 5443),
so that member clusters can access it from docker containers, please
input an available IP, default IP will be en0 inet addr if exists
[Enter for default 192.168.20.202]:
```

The script already attempts to auto-detect the IP via ipconfig getifaddr en0 and uses it as the default. However, it still halts execution waiting for user input even when a valid IP has been successfully detected. The user must manually press Enter to continue.

## Problems this causes

- Breaks fully automated setups and CI environments — any non-interactive execution will hang indefinitely waiting for input
- Creates a confusing experience for new developers who don't understand why they're being asked for an IP address
- Hardcodes en0 as the detection interface, which fails silently for developers on VPN or using non-standard network interfaces

## Files involved
- Script: `hack/local-up-karmada.sh`
- Function: `util::get_macos_ipaddress`  in  `hack/util.sh`

## Screenshot 
Image

Part of #7269

Krishiv-Mahajan:
**What type of PR is this?**
/kind bug

 The `util::get_macos_ipaddress` function in hack/util.sh unconditionally calls read to wait for user input, even when a valid IP was already auto-detected. This causes `hack/local-up-karmada.sh` to hang indefinitely in any non-interactive environment (CI pipelines, GitHub Actions, background scripts).

Additionally, the detection was hardcoded to en0 (WiFi only), which silently fails for developers on USB Ethernet, Thunderbolt, or VPN — resulting in an empty HOST_IPADDRESS and member clusters unable to reach the control plane.

**This PR replaces the interactive prompt with a layered auto-detection approach:**
If MAC_NIC_IPADDRESS is already set (e.g. by CI), use it directly
Detect the active interface via route get default instead of hardcoding en0
If the default route interface has no IP (VPN tunnel), scan en0–en3
Fall back to 127.0.0.1 with a clear warning if all detection fails

Fixes #7257 
Part of #7269

Tested end-to-end on macOS 14.3 (Apple Silicon) — ./hack/local-up-karmada.sh completed without any user input and all three member clusters reached Ready status. Linux and WSL2 code paths are unaffected (change is inside the existing darwin guard).

`hack/util.sh`: Fixed `util::get_macos_ipaddress` hanging for user input on macOS by replacing the interactive prompt with layered auto-detection from the active network interface, enabling non-interactive and CI usage.

### DIFF
--- a/hack/util.sh
+++ b/hack/util.sh
@@ -710,28 +710,55 @@ function util::get_wsl2_ipaddress() {
   util::verify_ip_address "${WSL2_HOST_IP_ADDRESS}"
 }
 
-# util::get_macos_ipaddress will get ip address on macos interactively, store to 'MAC_NIC_IPADDRESS' if available
-MAC_NIC_IPADDRESS=''
+# util::get_macos_ipaddress will auto-detect the host IP address on macOS, store to 'MAC_NIC_IPADDRESS' if available
+# The function tries the following in order:
+#   1. Use MAC_NIC_IPADDRESS if already set (e.g. by CI or the caller)
+#   2. Detect the active network interface via the routing table
+#   3. Scan common physical interfaces (en0-en3) as a fallback (e.g. when on VPN)
+#   4. Error return
+MAC_NIC_IPADDRESS=${MAC_NIC_IPADDRESS:-}
 function util::get_macos_ipaddress() {
   if [[ $(go env GOOS) = "darwin" ]]; then
-    tmp_ip=$(ipconfig getifaddr en0 || true)
     echo ""
     echo " Detected that you are installing Karmada on macOS "
     echo ""
     echo "It needs a Macintosh IP address to bind Karmada API Server(port 5443),"
-    echo "so that member clusters can access it from docker containers, please"
-    echo -n "input an available IP, "
-    if [[ -z ${tmp_ip} ]]; then
-      echo "you can use the command 'ifconfig' to look for one"
-      tips_msg="[Enter IP address]:"
-    else
-      echo "default IP will be en0 inet addr if exists"
-      tips_msg="[Enter for default ${tmp_ip}]:"
+    echo "so that member clusters can access it from docker containers"
+
+    # If already set by the caller (e.g. CI passing MAC_NIC_IPADDRESS env var), honour it
+    if [[ -n "${MAC_NIC_IPADDRESS:-}" ]]; then
+      util::verify_ip_address "${MAC_NIC_IPADDRESS}"
+      echo "Using provided IP address: ${MAC_NIC_IPADDRESS}"
+      return
     fi
-    read -r -p "${tips_msg}" MAC_NIC_IPADDRESS
-    MAC_NIC_IPADDRESS=${MAC_NIC_IPADDRESS:-$tmp_ip}
+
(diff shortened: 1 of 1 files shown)


zhzhuang-zju:
Thanks
/assign

hack/util.sh
```diff
@@ -710,28 +710,53 @@ function util::get_wsl2_ipaddress() {
...
 }
 
-# util::get_macos_ipaddress will get ip address on macos interactively, store to 'MAC_NIC_IPADDRESS' if available
-MAC_NIC_IPADDRESS=''
+# util::get_macos_ipaddress will auto-detect the host IP address on macOS, store to 'MAC_NIC_IPADDRESS' if available
+# The function tries the following in order:
+#   1. Use MAC_NIC_IPADDRESS if already set (e.g. by CI or the caller)
+#   2. Detect the active network interface via the routing table
+#   3. Scan common physical interfaces (en0-en3) as a fallback (e.g. when on VPN)
+#   4. Fall back to 127.0.0.1 as a last resort
```
zhzhuang-zju:
```suggestion
#   4. Error return
```
From your implementation, it will not fall back to 127.0.0.1 as it is not reachable from Docker containers

hack/util.sh
```diff
@@ -710,28 +710,53 @@ function util::get_wsl2_ipaddress() {
...
-    echo -n "input an available IP, "
-    if [[ -z ${tmp_ip} ]]; then
-      echo "you can use the command 'ifconfig' to look for one"
-      tips_msg="[Enter IP address]:"
-    else
-      echo "default IP will be en0 inet addr if exists"
-      tips_msg="[Enter for default ${tmp_ip}]:"
+    echo ""
+    echo " Detected that you are installing Karmada on macOS "
+    echo ""
```
zhzhuang-zju:
```suggestion
    echo ""
    echo " Detected that you are installing Karmada on macOS "
    echo ""
    echo "It needs a Macintosh IP address to bind Karmada API Server(port 5443),"
    echo "so that member clusters can access it from docker containers"
```

zhzhuang-zju (commented):
Thanks, others LGTM

zhzhuang-zju:
I'm not familiar with macOS. Maybe @zach593 could help take a look?

hack/util.sh
Krishiv-Mahajan: done

hack/util.sh
Krishiv-Mahajan: done

Krishiv-Mahajan:
> I'm not familiar with macOS. Maybe @zach593 could help take a look?

ok

zhzhuang-zju:
@Krishiv-Mahajan, please squash your commits, and then we can get this pr merged

zhzhuang-zju:
Thanks, let's move forward
/lgtm
/approve

Issue #7269, opened 2026-03-09
[LFX term 1] Enhance Karmada's Quick Start Experience and Incorporate macOS Support


zhzhuang-zju:
The initial interaction for many users with a project often begins with its "Quick Start" guide. Ensuring this guide is of high quality and up-to-date is crucial. Currently, the community offers several quick start methods, including a one-click setup script (hack/local-up-karmada.sh), installation tools such as Helm, karmadactl, and the operator, alongside learning tutorials on Killercoda.
Although the community's CI effectively maintains these installation methods, it lacks support for different operating systems, notably macOS. Furthermore, some Killercoda tutorials are outdated and do not reflect the latest features and best practices. This project aims to ease the entry barrier for new users by enhancing the quick start process, ensuring cross-platform compatibility, and updating educational content.

**Expected Outcome**:
  - Verify the feasibility of installing hack/local-up-karmada.sh on macOS environments, including both macOS ARM and Intel and address any issues that arise.
  - Implement a new CI workflow in GitHub Actions to test Karmada's installation and basic functionalities on macOS. GitHub Actions can select specific environments through runs-on, covering both macOS ARM and Intel.
  - Revise existing outdated scenarios on Killercoda and develop valuable new scenarios to guide users through current features. This includes:
    - Extracting common functions from various scenarios to reduce code duplication.
    - Updating the course information of existing scenarios by correcting outdated content and optimizing the material to enhance learning outcomes.
    - Adding new scenarios, such as:
      - A tutorial on FHPA usage.
      - An application-level failover guide.
      - Multi-component workload scheduling (support for workloads common in AI and big data).

zhzhuang-zju:
Hi @Krishiv-Mahajan, I‘ve reviewed your PR for task1. Overall, it looks good. I want to confirm:
- Is there any other pr for task1?
- Have you tested the final effect of your fix for task1? If so, can you show me the result? like the screenshot.

Krishiv-Mahajan:
@zhzhuang-zju I think the prs for task one are all done and I have tested the final effect of task one and the local setup is working smoothing 

<img width="1057" height="89" alt="Image" src="https://github.com/user-attachments/assets/a7ddb270-391a-4f6d-bcb9-387df1398f6f" />
this shows the result when docker is not working 

<img width="1057" height="833" alt="Image" src="https://github.com/user-attachments/assets/899d157a-e816-4008-bffe-3d4b5f6af440" />
[...]

PR #7274, opened 2026-03-11
fix: add curl fallback in deploy-k8s-metrics-server.sh for macOS comp…

### LINKED ISSUE #7273: hack/deploy-k8s-metrics-server.sh: wget not available on macOS by default, causing Quick Start failure
# What happened: 
Running  `hack/deploy-k8s-metrics-server.sh` on macOS fails immediately with` zsh: command not found: wget `because the script uses wget to download the metrics-server manifest, and wget is not installed on macOS by default.

# What you expected to happen: 
The script should download the metrics-server manifest and deploy it successfully on macOS without requiring any extra tool installations.

# How to reproduce it (as minimally and precisely as possible):
On a macOS machine (without wget installed via Homebrew)
Run `hack/deploy-k8s-metrics-server.sh` <kubeconfig> <context>
Script immediately fails at the wget download step

# Anything else we need to know?: 
`wget` is a GNU tool not bundled with macOS. The fix is to replace it with `curl`, which is available on macOS, Linux, and Windows and is already used consistently across other scripts in hack/ (e.g. util.sh, install-cli.sh).

# Suggested fix :

 In `hack/deploy-k8s-metrics-server.sh` first try curl , if it fails try using wget, if even that fails returns a clear error message
 
# Environment:

Karmada version: latest (master)
os: mac os Version 26.3 (25D125)

Part of #7269

Krishiv-Mahajan:
**What type of PR is this?**
/kind bug

`hack/deploy-k8s-metrics-server.sh` used wget to download the metrics-server manifest, but wget is not available on macOS by default. This caused the Quick Start script to fail immediately on macOS wit`h zsh: command not found: wget`. This PR adds curl as a fallback since curl is available on macOS, Linux, and Windows by default. `hack/install-cli.sh `already follows this pattern of preferring curl.

Fixes #7273
Part of #7269

`curl` and `wget `fetch the exact same file from the same URL. The only change is the tool used and the flag syntax (-O for wget → -o for curl). Behavior is identical on Linux where wget is typically pre-installed.

release`hack/deploy-k8s-metrics-server.sh`: Fixed Quick Start failure on macOS caused by missing `wget` by adding `curl` as a fallback.

### DIFF
--- a/hack/deploy-k8s-metrics-server.sh
+++ b/hack/deploy-k8s-metrics-server.sh
@@ -53,7 +53,16 @@ fi
 MEMBER_CLUSTER_NAME=$2
 
 # get deploy yaml
-wget https://github.com/kubernetes-sigs/metrics-server/releases/download/v0.6.3/components.yaml -O "${_tmp}/components.yaml"
+METRICS_SERVER_URL="https://github.com/kubernetes-sigs/metrics-server/releases/download/v0.6.3/components.yaml"
+# Prefer curl (available on macOS + Linux); fall back to wget on failure; fail with a clear error if both fail
+if command -v curl &>/dev/null && curl -fsSL "${METRICS_SERVER_URL}" -o "${_tmp}/components.yaml"; then
+  :
+elif command -v wget &>/dev/null && wget -qO "${_tmp}/components.yaml" "${METRICS_SERVER_URL}"; then
+  :
+else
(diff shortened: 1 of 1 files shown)


zhzhuang-zju:
Thanks, but I think this is an enhancement, rather than a bug
/remove-kind bug
/kind feature

/assign

zhzhuang-zju:
@Krishiv-Mahajan, please squash your commits into one so that we can merge them.

zhzhuang-zju:
Thanks
/lgtm
/approve

PR #7275, opened 2026-03-11
fix: correct SignerName in cert rotation CSR

### LINKED ISSUE #7256: cert_rotation_controller uses wrong SignerName causing agent CSR never approved during certificate rotation
**What happened**:
 During karmada-agent certificate rotation, the `cert_rotation_controller` submits CSRs with the wrong `SignerName`, which causes the `agent_csr_approving_controller` to filter them out and never approve them. This means certificate rotation silently fails.

  cert_rotation_controller should submit the CSR with SignerName = "kubernetes.io/kube-apiserver-client" (certificatesv1.KubeAPIServerClientSignerName), consistent with what agent_csr_approving_controller expects.

In [`pkg/controllers/certificate/cert_rotation_controller.go`](https://github.com/karmada-io/karmada/blob/105c6ef8e9e00440b5c65d38a592d25293531a79/pkg/controllers/certificate/cert_rotation_controller.go#L246), the `SignerName` constant is set to:

  ```go
  SignerName = "kubernetes.io/kube-apiserver-client-kubelet"
```
  However, in [`pkg/controllers/certificate/approver/agent_csr_approving.go`](https://github.com/karmada-io/karmada/blob/105c6ef8e9e00440b5c65d38a592d25293531a79/pkg/controllers/certificate/approver/agent_csr_approving.go#L151), the controller only processes and approves CSRs where 
```go
SignerName == certificatesv1.KubeAPIServerClientSignerName ("kubernetes.io/kube-apiserver-client")
```
  - The SetupWithManager predicate (lines 282, 287) filters out any CSR not using kubernetes.io/kube-apiserver-client.
  - The isAgentCSR / isSelfAgentCSR functions (line 197) also reject CSRs with a different signer.

  So the CSR submitted by cert_rotation_controller (signer: kubernetes.io/kube-apiserver-client-kubelet) is silently dropped by the approving controller and will never be approved. The rotation then times out waiting for a certificate that will never come.

**Environment**:
- Karmada version:
- Others:

hl8086:
**What type of PR is this?**

/kind bug

`cert_rotation_controller` was creating CSRs with `SignerName` set to `"kubernetes.io/kube-apiserver-client-kubelet"`, but `agent_csr_approving` only recognizes CSRs with `SignerName` equal to `"kubernetes.io/kube-apiserver-client"`. This mismatch caused certificate rotation CSRs to be silently ignored by the approver, resulting in rotation timeout and silent failure. This PR corrects the `SignerName` to align with the approver's expectation.

Fixes #7256

A regression test `TestSignerNameMatchesAgentCSRApprover` has been added to `cert_rotation_controller_test.go` to assert that the `SignerName` used when creating a rotation CSR always matches the value expected by `agent_csr_approving`, preventing future regressions.

```release-note
`karmada-agent`: Fixed the issue that certificate rotation CSRs were never auto-approved due to a SignerName mismatch between `cert_rotation_controller` and `agent_csr_approving`.

### DIFF
--- a/pkg/controllers/certificate/cert_rotation_controller.go
+++ b/pkg/controllers/certificate/cert_rotation_controller.go
@@ -53,8 +53,9 @@ const (
 	// CertRotationControllerName is the controller name that will be used when reporting events and metrics.
 	CertRotationControllerName = "cert-rotation-controller"
 
-	// SignerName defines the signer name for csr, 'kubernetes.io/kube-apiserver-client-kubelet' can sign the csr automatically
-	SignerName = "kubernetes.io/kube-apiserver-client-kubelet"
+	// SignerName defines the signer name for csr, 'kubernetes.io/kube-apiserver-client' is used
+	// to match the signer expected by the agent CSR approver (agent_csr_approving).
+	SignerName = certificatesv1.KubeAPIServerClientSignerName
 
 	// KarmadaKubeconfigName is the name of the secret containing karmada-agent certificate.
 	KarmadaKubeconfigName = "karmada-kubeconfig"
--- a/pkg/controllers/certificate/cert_rotation_controller_test.go
+++ b/pkg/controllers/certificate/cert_rotation_controller_test.go
@@ -28,6 +28,7 @@ import (
 	"testing"
 	"time"
 
+	certificatesv1 "k8s.io/api/certificates/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: pkg/controllers/certificate/cert_rotation_controller.go [hunk 1]
```diff
@@ -53,8 +53,9 @@ const (
 	// CertRotationControllerName is the controller name that will be used when reporting events and metrics.
 	CertRotationControllerName = "cert-rotation-controller"
 
-	// SignerName defines the signer name for csr, 'kubernetes.io/kube-apiserver-client-kubelet' can sign the csr automatically
-	SignerName = "kubernetes.io/kube-apiserver-client-kubelet"
+	// SignerName defines the signer name for csr, 'kubernetes.io/kube-apiserver-client' is used
+	// to match the signer expected by the agent CSR approver (agent_csr_approving).
+	SignerName = certificatesv1.KubeAPIServerClientSignerName
```
XiShanYongYe-Chang, line 58: ~Hi @hl8086 These two should be different, `-kubelet` is missing.~

~The label of this PR is not quite correct; it's not really a bug.~

Sorry, I saw the related issue. Ignore me

zhzhuang-zju:
Thanks, will take a look tomorrow
/assign
The DCO CI has failed. You can use the command `git commit -s -m ` to sign this pr off

pkg/controllers/certificate/cert_rotation_controller.go
zhzhuang-zju, line 58: https://github.com/karmada-io/karmada/blob/3aaefc07a226553167669643779b35b83babb646/pkg/karmadactl/register/register.go#L101
Side note: it would be great to also use `certificatesv1.KubeAPIServerClientSignerName` here, although it's out of scope for this specific issue

zhzhuang-zju (commented):
Thanks, others LGTM

To confirm, has this issue already been fixed in your environment?

pkg/controllers/certificate/cert_rotation_controller.go
hl8086: I haven't set up a full environment test yet, but I manually created a CSR with `SignerName = kubernetes.io/kube-apiserver-client` and successfully issued a new certificate, which confirms the signing flow works correctly with this SignerName.
I'd like to review the logic together first before running a full environment verification.

zhzhuang-zju: >I'd like to review the logic together first before running a full environment verification.

I have no further questions. If you also find no issues after your review, we can merge it.

zhzhuang-zju:
Thanks, as we don't find other issues, we can push this forward ~

/lgtm
/approve

PR #7279, opened 2026-03-12
fix: use sed -E instead of sed -r for macOS BSD sed compatibility

### LINKED ISSUE #7278: hack/update-estimator-protobuf.sh: sed -r flag incompatible with macOS BSD sed
# Fix `sed -r` → `sed -E` for macOS compatibility in `update-estimator-protobuf.sh`

Part of #7269

## Problem

Line 47 of `hack/update-estimator-protobuf.sh` uses `sed -r` to extract the `protoc` major version:

```bash
if [[ -z "$(which protoc)" || $(protoc --version | sed -r "s/libprotoc ([0-9]+).*/\1/g") -lt 3 ]];
```

`-r` is a GNU sed flag and is not recognized by BSD sed, which ships with macOS. Running this script on macOS produces:

```
sed: illegal option -- r
```

Since this is inside an `if` condition at the very start of the script, it aborts immediately — before even checking if `protoc` is installed.

## Fix

Replace `-r` with `-E`. The regex pattern itself is unchanged.

```bash
# Before (GNU sed only)
protoc --version | sed -r "s/libprotoc ([0-9]+).*/\1/g"

# After (BSD + GNU sed)
protoc --version | sed -E "s/libprotoc ([0-9]+).*/\1/g"
```

`-E` is the POSIX standard flag for extended regex and is supported by both BSD sed (macOS) and GNU sed (Linux, since 4.2 / 2012).

## Testing

Verified on macOS (Apple Silicon) — script no longer aborts with `illegal option` error.

Krishiv-Mahajan:
**What type of PR is this?**
/kind bug

`hack/update-estimator-protobuf.sh` uses` sed -r` to extract the `protoc` major version. `-r` is a GNU sed flag not recognized by BSD sed, which ships with macOS. This causes the script to abort immediately with` sed: illegal option -- r` on any fresh Mac. Replacing `-r `with `-E` fixes this — `-E `is POSIX standard and supported by both BSD sed (macOS) and GNU sed (Linux). No functional change.

Fixes #7278
Part of #7269

One character change — `-r `→ `-E` on line 47. Regex pattern is identical. No functional change on Linux.

NONE

### DIFF
--- a/hack/update-estimator-protobuf.sh
+++ b/hack/update-estimator-protobuf.sh
@@ -44,7 +44,7 @@ util:create_gopath_tree "${KARMADA_ROOT}" "${go_path}"
 export GOPATH="${go_path}"
 
 # https://github.com/kubernetes/kubernetes/blob/release-1.23/hack/update-generated-protobuf-dockerized.sh
-if [[ -z "$(which protoc)" || $(protoc --version | sed -r "s/libprotoc ([0-9]+).*/\1/g") -lt 3 ]]; then
+if [[ -z "$(which protoc)" || $(protoc --version | sed -E "s/libprotoc ([0-9]+).*/\1/g") -lt 3 ]]; then
   echo "Generating protobuf requires protoc 3.0.0-beta1 or newer. Please download and"
   echo "install the platform appropriate Protobuf package for your OS: "
   echo


zhzhuang-zju:
```bash
$ sed --help
Usage: sed [OPTION]... {script-only-if-no-other-script} [input-file]...
  -E, -r, --regexp-extended
                 use extended regular expressions in the script
                 (for portability use POSIX -E).
```
Thanks. According to the output from `sed --help`, `sed -r` and `sed -E` are functionally equivalent. However, for better cross-platform compatibility, `sed -E` is recommended. Therefore, I think this change is valid.

Additionally, @Krishiv-Mahajan, have you found any other places where `sed -r` is used? I did a quick search and found none. If you have found any, we can also replace them with `sed -E` at the same time.

Krishiv-Mahajan:
@zhzhuang-zju no I dont think there is `sed -r` is used anywhere else

PR #7287, opened 2026-03-14
Fix component-based scheduling to include single-component workloads

### LINKED ISSUE #7277: Multi-component accurate estimating not invoked when there's only one component in the workload
**What happened**:

We turned on multi-component scheduling and accurate schedule estimator. We're using volcano jobs. When we submit a volcano job with one task, it may be scheduled to a wrong member cluster. If there're multiple tasks, the scheduler works well.

After some investigation, we found, MultiplePodTemplatesScheduling is disabled if there is only one component. Here's the code snippet.

https://github.com/karmada-io/karmada/blob/3aaefc07a226553167669643779b35b83babb646/pkg/scheduler/core/estimation.go#L47-L49

https://github.com/karmada-io/karmada/blob/3aaefc07a226553167669643779b35b83babb646/pkg/scheduler/core/util.go#L79-L86

However, the `replicaRequirements` is always empty after we enabled MultiplePodTemplatesScheduling. Therefore, the job can be scheduled to the wrong cluster.

With MultiplePodTemplatesScheduling and accurate schedule estimator both on, volcano job can be scheduled to the correct cluster.

1. Using the resource interpreter pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/batch.volcano.sh/v1alpha1/Job/customizations.yaml .
2. Turn on both MultiplePodTemplatesScheduling and accurate schedule estimator
3. Submit a one replica volcano job. Alter the pod nodeAffinity to see the effect.

**Environment**:
- Karmada version: v1.16.1
- kubectl-karmada or karmadactl version (the result of `kubectl-karmada version` or `karmadactl version`): unused

seanlaii:
**What type of PR is this?**
/kind bug

When the `MultiplePodTemplatesScheduling` feature gate is enabled, the `Components` field is populated for all workloads, including **single-component** workloads. However, `IsBindingReplicasChanged` and `isMultiTemplateSchedulingApplicable` both required `len(Components) > 1 `(i.e., 2+ components) before entering the component-aware code paths. This meant **single**-component workloads were silently excluded from:

1. Failover rescheduling — When all clusters are evicted (clusters become empty), a single-component workload would not trigger rescheduling, leaving it stuck with no target clusters.

2. Component-based replica estimation — The scheduler would skip the `MaxAvailableComponentSets` estimation path for single-component workloads, falling back to the `Replica Scheduling` path even though `Components` was populated.

This contradicts the design of the Components field, which explicitly states it "is also capable of representing single-component workloads, such as Deployment."

This PR fixes both checks to use `len(Components) > 0`, ensuring consistent behavior for all component-based workloads regardless of the number of components. The change is fully guarded by the `MultiplePodTemplatesScheduling` feature gate, so existing behavior is unaffected when the gate is disabled.

Fixes #7277

```release-note
`karmada-scheduler`: Extended the MultiplePodTemplatesScheduling feature to support workloads with only one component.
```

### DIFF
--- a/pkg/scheduler/core/estimation.go
+++ b/pkg/scheduler/core/estimation.go
@@ -29,8 +29,8 @@ import (
 )
 
 // isMultiTemplateSchedulingApplicable checks if the given ResourceBindingSpec
-// meets the criteria for multi-template scheduling:
-//  1. The referenced resource holds multiple pod templates (multiple components).
+// meets the criteria for component-based scheduling:
+//  1. The referenced resource has at least one component populated.
 //  2. The placement configuration schedules the resource to exactly one cluster.
 //     This is currently determined by checking if spread constraints is set and requires exactly one cluster.
 //
@@ -44,7 +44,7 @@ func isMultiTemplateSchedulingApplicable(spec *workv1alpha2.ResourceBindingSpec)
 		return false
 	}
 
-	if len(spec.Components) < 2 {
+	if len(spec.Components) == 0 {
 		return false
 	}
 
--- a/pkg/scheduler/core/estimation_test.go
+++ b/pkg/scheduler/core/estimation_test.go
@@ -128,7 +128,7 @@ func Test_isMultiTemplateSchedulingApplicable(t *testing.T) {
 			want: false,
 		},
 		{
-			name: "spec with single component should not be applicable",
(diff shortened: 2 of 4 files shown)


RainbowMango (commented):
/assign

Issue #7252, opened 2026-03-05
[Test] Improve E2E test coverage for Karmada controller events


XiShanYongYe-Chang:
### Background

The current E2E test suite has very limited coverage of events emitted by Karmada controllers. As of now, only 4 out of 51 unique EventReason constants defined in `pkg/events/events.go` are verified in E2E tests (~7.8% coverage). The existing helper `framework/events.go` also only covers `core/v1` Event via CoreV1().Events().

This issue discusses two approaches to improve event test coverage and proposes a path forward.

### Approach 1: Dedicated Event Test File

  Create a new file test/e2e/suites/base/event_test.go, organized by EventReason, with shared resource setup reused across multiple event assertions.

  Pros

   - Centralized visibility: All event tests live in one place, making it straightforward to audit which EventReason constants are covered and which are not.
   - Efficient resource usage: A single resource setup (e.g., one Deployment + PropagationPolicy) can drive multiple event assertions, avoiding redundant creation and cleanup cycles.
   - Easy to extend: Adding coverage for a newly introduced EventReason has a clear home — no need to decide which existing test file to attach it to.
   - Prevents silent regressions: When a new EventReason is added to pkg/events/events.go, the gap in this file is immediately visible.

  Cons

   - Forced complexity for scenario-bound events: Some events are only produced under specific conditions (e.g., EvictWorkloadFromCluster* requires a failover scenario, DependencyPolicyConflict requires a conflict in dependency distribution). Triggering these from a generic setup
  introduces artificial complexity that is harder to maintain than simply placing the assertion next to the operation that triggers it.
   - Semantic separation: The event assertion is decoupled from the operation that produces it, making it less intuitive to understand why a given event is expected.
   
### Approach 2: Inline Event Assertions in Existing Test Cases

  Add `framework.WaitEventFitWith(...)` steps directly into existing test cases at the point where the relevant operation completes.

  Pros

   - Semantic accuracy: Event assertions sit immediately after the operation that triggers them, reflecting how users observe events in practice.
   - No extra setup cost: Fully reuses the resource lifecycle already established by the existing test case's BeforeEach / DeferCleanup.
   - Natural fit for scenario-bound events: Events that require specific conditions (failover, conflict, suspension) are already covered by the corresponding test cases.

  Cons

   - Scattered coverage: Event assertions are spread across 20+ test files with no single view of overall coverage. Identifying gaps requires grep across the entire suite.
[...]

PR #7300, opened 2026-03-18
test: add inline event assertions for resource template events

nXtCyberNet:
**What type of PR is this?**

/kind feature
/kind test

---

This PR adds inline event assertions to existing E2E test cases to improve event coverage for resource template events, as discussed in the related issue.

Following the community decision to use **Approach 2 (Inline Event Assertions)**, event checks are added directly inside existing test cases at the point where the relevant operation completes — no new test cases or files are introduced.

### `test/e2e/framework/events.go`

* Added `WaitEventV1FitWith` helper to support `events.k8s.io/v1` Event API alongside the existing `core/v1` helper (foundation for future migration)

### `test/e2e/suites/base/propagationpolicy_test.go`

Added inline event assertions to existing Deployment, Service, and Job propagation test cases:

* `ApplyPolicySucceed` — verified on Deployment, Service, Job
* `GetReplicasSucceed` — verified on Deployment, Job
* `ScheduleBindingSucceed` — verified on Deployment, Service, Job
* `SyncWorkSucceed` — verified on Deployment, Service, Job
* `AggregateStatusSucceed` — verified on Deployment, Job
* `ReflectStatusSucceed` — verified on Work object during Deployment propagation
* `InterpretHealthSucceed` — verified on Work object during Deployment propagation
* `SyncWorkloadSucceed` — verified on Work object during Deployment propagation

### `test/e2e/suites/base/overridepolicy_test.go`

Added inline event assertions to existing override policy test cases:

* `ApplyOverridePolicySucceed` — verified across OverridePolicy and OverrideRules test cases with Deployment

### `test/e2e/suites/base/preemption_test.go`

Added inline event assertions to existing preemption test cases:

* `PreemptPolicySucceed` — verified when high-priority PropagationPolicy preempts low-priority PropagationPolicy, when PropagationPolicy preempts ClusterPropagationPolicy, and when high-priority ClusterPropagationPolicy preempts low-priority ClusterPropagationPolicy

### `test/e2e/suites/base/dependenciesdistributor_test.go`

Added inline event assertions to existing dependency propagation test cases:

* `GetDependenciesSucceed` — verified in configmap automatic propagation tests with both PropagationPolicy and ClusterPropagationPolicy

---

Fixes part of  #7252

---

## How was this tested

All assertions were verified by running the E2E suite locally against a kind-based Karmada setup and confirming the expected events fire during the propagation flow.

```bash
export KUBECONFIG=$HOME/.kube/karmada.config
go test ./test/e2e/suites/base/... \
  -v \
  --ginkgo.focus="BasicCase|OverridePolicy|OverrideRules|Preemption|DependenciesDistributor" \
  --ginkgo.timeout=60m \
  -- \
  --host-context=karmada-host \
  --karmada-context=karmada-apiserver \
  --poll-timeout=300s
```

## Notes

* `GetDependenciesFailed`, `GetReplicasFailed`, and `ApplyPolicyFailed` are not covered in this PR as they require specific failure conditions that are difficult to construct reliably in E2E tests
* `events.k8s.io/v1` support added in `framework/events.go` is not yet used in assertions — it is foundation work for the planned migration from `GetEventRecorderFor` to `GetEventRecorder`

---

```release-note
NONE
```

### DIFF
--- a/test/e2e/suites/base/dependenciesdistributor_test.go
+++ b/test/e2e/suites/base/dependenciesdistributor_test.go
@@ -113,6 +113,12 @@ var _ = ginkgo.Describe("[DependenciesDistributor] automatically propagate relev
 							return true
 						})
 				})
+				ginkgo.By("checking GetDependenciesSucceed event on deployment", func() {
+					framework.WaitEventFitWith(kubeClient, deployment.Namespace, deployment.Name,
+						func(event corev1.Event) bool {
+							return event.Reason == events.EventReasonGetDependenciesSucceed
+						})
+				})
 
 				ginkgo.By("updating propagation policy's clusterNames", func() {
 					patch := []map[string]any{
@@ -206,6 +212,13 @@ var _ = ginkgo.Describe("[DependenciesDistributor] automatically propagate relev
 						func(*corev1.Secret) bool {
 							return true
 						})
+
+					ginkgo.By("checking GetDependenciesSucceed event on deployment", func() {
+						framework.WaitEventFitWith(kubeClient, deployment.Namespace, deployment.Name,
+							func(event corev1.Event) bool {
+								return event.Reason == events.EventReasonGetDependenciesSucceed
+							})
+					})
 				})
 
 				ginkgo.By("make the secret is not referenced by the deployment ", func() {
@@ -276,10 +289,16 @@ var _ = ginkgo.Describe("[DependenciesDistributor] automatically propagate relev
 						func(*appsv1.Deployment) bool {
 							return true
 						})
-
 					framework.WaitPVCPresentOnClustersFitWith(initClusterNames, pvc.GetNamespace(), pvc.GetName(),
 						func(*corev1.PersistentVolumeClaim) bool {
 							return true
+						})
+				})
+
+				ginkgo.By("checking GetDependenciesSucceed event on deployment", func() {
+					framework.WaitEventFitWith(kubeClient, deployment.Namespace, deployment.Name,
+						func(event corev1.Event) bool {
+							return event.Reason == events.EventReasonGetDependenciesSucceed
 						})
 				})
 
@@ -341,6 +360,12 @@ var _ = ginkgo.Describe("[DependenciesDistributor] automatically propagate relev
 					framework.WaitServiceAccountPresentOnClustersFitWith(initClusterNames, sa.GetNamespace(), sa.GetName(),
 						func(*corev1.ServiceAccount) bool {
 							return true
+						})
+				})
+				ginkgo.By("checking GetDependenciesSucceed event on deployment", func() {
+					framework.WaitEventFitWith(kubeClient, deployment.Namespace, deployment.Name,
+						func(event corev1.Event) bool {
+							return event.Reason == events.EventReasonGetDependenciesSucceed
 						})
 				})
 
@@ -680,6 +705,13 @@ var _ = ginkgo.Describe("[DependenciesDistributor] automatically propagate relev
 					framework.WaitConfigMapPresentOnClustersFitWith(initClusterNames, configMap.Namespace, configMapName,
 						func(*corev1.ConfigMap) bool {
 							return true
+						})
+				})
+
+				ginkgo.By("checking GetDependenciesSucceed event on deployment", func() {
+					framework.WaitEventFitWith(kubeClient, deployment.Namespace, deployment.Name,
+						func(event corev1.Event) bool {
+							return event.Reason == events.EventReasonGetDependenciesSucceed
 						})
 				})
 
--- a/test/e2e/suites/base/overridepolicy_test.go
+++ b/test/e2e/suites/base/overridepolicy_test.go
@@ -28,6 +28,7 @@ import (
 	"k8s.io/klog/v2"
 
 	policyv1alpha1 "github.com/karmada-io/karmada/pkg/apis/policy/v1alpha1"
+	"github.com/karmada-io/karmada/pkg/events"
 	"github.com/karmada-io/karmada/test/e2e/framework"
 	"github.com/karmada-io/karmada/test/helper"
 )
@@ -112,6 +113,13 @@ var _ = ginkgo.Describe("[OverridePolicy] apply overriders testing", func() {
 		})
 
 		ginkgo.It("deployment labelsOverride testing", func() {
+			ginkgo.By("checking ApplyOverridePolicySucceed event on deployment", func() {
+				framework.WaitEventFitWith(kubeClient, deploymentNamespace, deploymentName,
+					func(event corev1.Event) bool {
+						return event.Reason == events.EventReasonApplyOverridePolicySucceed
+					})
+			})
+
 			klog.Infof("check if deployment present on member clusters have correct labels value")
 			framework.WaitDeploymentPresentOnClustersFitWith(framework.ClusterNames(), deployment.Namespace, deployment.Name,
 				func(deployment *appsv1.Deployment) bool {
@@ -274,6 +282,13 @@ var _ = ginkgo.Describe("[OverridePolicy] apply overriders testing", func() {
 		})
 
 		ginkgo.It("deployment imageOverride testing", func() {
+			ginkgo.By("checking ApplyOverridePolicySucceed event", func() {
+				framework.WaitEventFitWith(kubeClient, deploymentNamespace, deploymentName,
+					func(event corev1.Event) bool {
+						return event.Reason == events.EventReasonApplyOverridePolicySucceed
+					})
+			})
+
 			klog.Infof("check if deployment present on member clusters have correct image value")
 			framework.WaitDeploymentPresentOnClustersFitWith(framework.ClusterNames(), deployment.Namespace, deployment.Name,
 				func(deployment *appsv1.Deployment) bool {
@@ -422,6 +437,13 @@ var _ = ginkgo.Describe("[OverridePolicy] apply overriders testing", func() {
 		})
 
 		ginkgo.It("deployment imageOverride testing", func() {
+			ginkgo.By("checking ApplyOverridePolicySucceed event", func() {
+				framework.WaitEventFitWith(kubeClient, deploymentNamespace, deploymentName,
+					func(event corev1.Event) bool {
+						return event.Reason == events.EventReasonApplyOverridePolicySucceed
+					})
+			})
+
 			klog.Infof("check if deployment present on member clusters have correct image value")
 			framework.WaitDeploymentPresentOnClustersFitWith(framework.ClusterNames(), deployment.Namespace, deployment.Name,
 				func(deployment *appsv1.Deployment) bool {
@@ -703,6 +725,13 @@ var _ = framework.SerialDescribe("OverridePolicy with nil resourceSelector testi
 		})
 
 		ginkgo.It("deployment imageOverride testing", func() {
+			ginkgo.By("checking ApplyOverridePolicySucceed event", func() {
+				framework.WaitEventFitWith(kubeClient, deploymentNamespace, deploymentName,
+					func(event corev1.Event) bool {
+						return event.Reason == events.EventReasonApplyOverridePolicySucceed
+					})
+			})
+
 			klog.Infof("check if deployment present on member clusters have correct image value")
 			framework.WaitDeploymentPresentOnClustersFitWith(framework.ClusterNames(), deployment.Namespace, deployment.Name,
 				func(deployment *appsv1.Deployment) bool {
@@ -788,6 +817,12 @@ var _ = ginkgo.Describe("[OverrideRules] apply overriders testing", func() {
 		})
 
 		ginkgo.It("deployment imageOverride testing", func() {
+			ginkgo.By("checking ApplyOverridePolicySucceed event", func() {
+				framework.WaitEventFitWith(kubeClient, deploymentNamespace, deploymentName,
+					func(event corev1.Event) bool {
+						return event.Reason == events.EventReasonApplyOverridePolicySucceed
+					})
+			})
 			klog.Infof("check if deployment present on member clusters have correct image value")
 			framework.WaitDeploymentPresentOnClustersFitWith(framework.ClusterNames(), deployment.Namespace, deployment.Name,
 				func(deployment *appsv1.Deployment) bool {
@@ -881,6 +916,7 @@ var _ = ginkgo.Describe("[OverrideRules] apply overriders testing", func() {
 				return true
 			})
 		})
+
 	})
 
 	ginkgo.Context("Deployment override specific images in container list", func() {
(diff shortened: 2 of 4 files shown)


test/e2e/framework/events.go
```diff
@@ -37,3 +38,18 @@ func WaitEventFitWith(kubeClient kubernetes.Interface, namespace string, involve
...
+		eventList, err := kubeClient.EventsV1().Events(namespace).List(context.TODO(), metav1.ListOptions{
+			FieldSelector: fields.OneTermEqualSelector("regarding.name", involvedObj).String(),
+		})
+		gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+		for _, e := range eventList.Items {
+			if fit(e) {
+				return true
+			}
+		}
+		return false
```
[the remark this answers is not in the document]
nXtCyberNet: its  because the k8s/event/v1 will use regarding.name instead of involvedObject.name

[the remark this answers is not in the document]
nXtCyberNet: events.k8s.io/v1 events are best-effort and field selector behavior can be inconsistent, so explicit iteration is preferred for better debugging and reliability in E2E tests

XiShanYongYe-Chang:
> Fixes https://github.com/karmada-io/karmada/issues/7252

This will close the target issue. Please adjust the wording: Fixes part of xx

test/e2e/framework/events.go
```diff
@@ -37,3 +38,20 @@ func WaitEventFitWith(kubeClient kubernetes.Interface, namespace string, involve
 		return slices.ContainsFunc(eventList.Items, fit)
 	}, PollTimeout, PollInterval).Should(gomega.Equal(true))
 }
+
+// WaitEventV1FitWith uses events.k8s.io/v1 to do the same thing as WaitEventFitWith
+func WaitEventV1FitWith(kubeClient kubernetes.Interface, namespace string, involvedObj string, fit func(eventsv1.Event) bool) {
```
XiShanYongYe-Chang: Support for eventsv1 can be added later.

XiShanYongYe-Chang:
Once this PR is ready, you can squash all the commits into one and then ask me to review it.

test/e2e/framework/events.go
nXtCyberNet: Okayy , I will remove it

nXtCyberNet: @XiShanYongYe-Chang  btw I didn’t use slices.ContainsFunc here because of the somewhat unpredictable nature of the events/v1 API (ordering, transient updates, etc.). does this issue still exist in newer versions, or is it safe to use ContainsFunc now? just asking for my knowledge, thanks.

XiShanYongYe-Chang: I don't quite understand why we can't use ContainsFunc.

nXtCyberNet:
> Once this PR is ready, you can squash all the commits into one and then ask me to review it.

Hi, I’ve squashed all the commits into one. The PR is ready for review, please take a look

XiShanYongYe-Chang:
Thanks
/assign

test/e2e/suites/base/overridepolicy_test.go
```diff
@@ -121,6 +122,12 @@ var _ = ginkgo.Describe("[OverridePolicy] apply overriders testing", func() {
 						deployment.GetLabels()["foo"] == "exist" &&
 						deployment.GetLabels()["app"] == "nginx"
 				})
+			ginkgo.By("checking ApplyOverridePolicySucceed event on deployment", func() {
+				framework.WaitEventFitWith(kubeClient, deploymentNamespace, deploymentName,
+					func(event corev1.Event) bool {
+						return event.Reason == events.EventReasonApplyOverridePolicySucceed
+					})
```
XiShanYongYe-Chang: e2e runs in parallel, and events exist for a period of time after being created. If we only check the reason without checking the corresponding resource name, can we effectively make a judgment?

nXtCyberNet: Hi, I believe this should not be an issue since all resources are created in their respective namespaces. Each namespace, along with its deployments, follows a consistent naming pattern, ensuring proper isolation.
Additionally, we are matching against InvolvedObject.Name, which is sufficiently unique, reducing the likelihood of false positive signals from unrelated resources.
That said, please let me know if I might be missing any edge cases or scenarios that should be considered.

XiShanYongYe-Chang: Yes, you are right

test/e2e/framework/events.go
XiShanYongYe-Chang: Delete this function first?

nXtCyberNet: Apologies, I overlooked that earlier. I’ve included it now

test/e2e/suites/base/dependenciesdistributor_test.go
```diff
@@ -113,6 +113,12 @@ var _ = ginkgo.Describe("[DependenciesDistributor] automatically propagate relev
 							return true
 						})
 				})
+				ginkgo.By("checking GetDependenciesSucceed event on deployment", func() {
```
XiShanYongYe-Chang, line 116: There are also dependency distribution scenarios for other resources in this test file. Why were these not tested together?

test/e2e/suites/base/overridepolicy_test.go
```diff
@@ -121,6 +122,12 @@ var _ = ginkgo.Describe("[OverridePolicy] apply overriders testing", func() {
 						deployment.GetLabels()["foo"] == "exist" &&
 						deployment.GetLabels()["app"] == "nginx"
 				})
+			ginkgo.By("checking ApplyOverridePolicySucceed event on deployment", func() {
```
XiShanYongYe-Chang: The check on an event can be moved before the check on resource distribution, as shown in line 116.

Can the same test be performed on ClusterOverridePolicy?

test/e2e/suites/base/preemption_test.go
```diff
@@ -87,6 +89,12 @@ var _ = ginkgo.Describe("[Preemption] propagation policy preemption testing", fu
 					framework.CreatePropagationPolicy(karmadaClient, highPriorityPolicy)
 					framework.WaitDeploymentPresentOnClusterFitWith(preemptingClusterName, deployment.Namespace, deployment.Name, func(*appsv1.Deployment) bool { return true })
 				})
+				ginkgo.By("checking PreemptPolicySucceed event on deployment", func() {
```
XiShanYongYe-Chang, line 92: Similarly, I think event checks can be moved before resource checks.

test/e2e/suites/base/propagationpolicy_test.go
```diff
@@ -157,59 +226,50 @@ var _ = ginkgo.Describe("[BasicCase] PropagationPolicy testing", func() {
 					return true
 				})
 
-			patch := []map[string]any{{"op": "replace", "path": "/spec/ports/0/port", "value": updateServicePort}}
```
XiShanYongYe-Chang, line 160: Why do we need to remove these codes? Same as below.

test/e2e/suites/base/propagationpolicy_test.go
```diff
@@ -97,16 +97,85 @@ var _ = ginkgo.Describe("[BasicCase] PropagationPolicy testing", func() {
...
+					})
+			})
 
-			framework.UpdateDeploymentReplicas(kubeClient, deployment, updateDeploymentReplicas)
-			framework.WaitDeploymentPresentOnClustersFitWith(framework.ClusterNames(), deployment.Namespace, deployment.Name,
-				func(deployment *appsv1.Deployment) bool {
-					return *deployment.Spec.Replicas == updateDeploymentReplicas
-				})
+			// resource template events — fired on the Deployment object itself
+			ginkgo.By("checking ApplyPolicySucceed event", func() {
```
XiShanYongYe-Chang, line 100: I think the event check here can be divided into two parts. The first part is the event for resource distribution, which can be checked before WaitDeploymentPresentOnClustersFitWith. The second part is the event for status collection, which can be checked after WaitDeploymentPresentOnClustersFitWith. WDYT

XiShanYongYe-Chang (commented):
Thanks a lot

XiShanYongYe-Chang:
Hi @nXtCyberNet The first PR may have a lot of reviews, but subsequent PRs should be much better if they are based on this template.

test/e2e/suites/base/overridepolicy_test.go
nXtCyberNet: Hi @XiShanYongYe-Chang, thank you for the feedback. I’ll move the event assertion before the resource distribution check in all affected It blocks. Currently, this PR is focused on the resource template subtask; I’ll add ApplyOverridePolicySucceed assertions to the ClusterOverridePolicy tests in another PR

test/e2e/suites/base/dependenciesdistributor_test.go
nXtCyberNet: You're right — those scenarios were missed earlier as each When block creates isolated deployments; I'll  added the dependency propagation assertions for all remaining resources.

test/e2e/suites/base/propagationpolicy_test.go
nXtCyberNet: One concern I want to raise before finalizing the grouping approach:

Since Kubernetes events are best-effort and not guaranteed to be delivered (they can be dropped under memory pressure or aggregated), is there a risk that WaitEventFitWith could cause a flaky test failure simply because an event was dropped rather than because the feature actually broke?

I noticed the existing event assertions in the codebase (e.g. WorkDispatching, DependencyPolicyConflict) make the same assumption. Just wanted to confirm this tradeoff is acceptable for this PR as well before proceeding — I can adjust the approach if needed.

nXtCyberNet: One follow-up thought — if events are dropped, wouldn’t the existing By block checks also fail for the same reason? In that case, the flakiness risk isn’t specific to WaitEventFitWith but applies to the current approach as well.

XiShanYongYe-Chang: Very good question.

>  I can adjust the approach if needed.

Do you have any good ideas?

XiShanYongYe-Chang: > One follow-up thought — if events are dropped, wouldn’t the existing By block checks also fail for the same reason? In that case, the flakiness risk isn’t specific to WaitEventFitWith but applies to the current approach as well.

@gemini-code-assist
Do you have any good suggestions?

XiShanYongYe-Chang: Based on the above analysis, I think that our current configuration with a longer timeout should be acceptable. If we find out later that events are being dropped, then we can consider the `Non-blocking checks:` again. WDYT @nXtCyberNet

XiShanYongYe-Chang: Kindly ping @nXtCyberNet 
Can we adjust the order of the inspection based on the first comments?

test/e2e/suites/base/overridepolicy_test.go
```diff
@@ -113,6 +114,12 @@ var _ = ginkgo.Describe("[OverridePolicy] apply overriders testing", func() {
 
 		ginkgo.It("deployment labelsOverride testing", func() {
 			klog.Infof("check if deployment present on member clusters have correct labels value")
+			ginkgo.By("checking ApplyOverridePolicySucceed event on deployment", func() {
+				framework.WaitEventFitWith(kubeClient, deploymentNamespace, deploymentName,
+					func(event corev1.Event) bool {
+						return event.Reason == events.EventReasonApplyOverridePolicySucceed
+					})
+			})
```
XiShanYongYe-Chang, line 121: If this logic is moved to the previous line of log, it would feel better.

pkg/events/events.go
XiShanYongYe-Chang, line 134: Why do we need to add some new events for the current PR?

nXtCyberNet: You're right, GetComponentsSucceed and GetComponentsFailed are out of scope for this PR. I noticed those event emissions were missing while working on the detector, but they aren't needed for the current assertions. I'll remove them and can submit them separately if needed.

test/e2e/suites/base/propagationpolicy_test.go
```diff
@@ -167,53 +274,6 @@ var _ = ginkgo.Describe("[BasicCase] PropagationPolicy testing", func() {
 		})
 	})
 
-	ginkgo.Context("Pod propagation testing", func() {
```
XiShanYongYe-Chang, line 170: Was this test mistakenly deleted?

other parts lgtm

XiShanYongYe-Chang: Could you please help remind me of any changes?

mohamedawnallah:
Thanks @nXtCyberNet for the PR!

I ran mutation testing on the line of code this PR exercises: https://github.com/karmada-io/karmada/blob/b15f5bd70ed3cf6e62bbea9126e1e1053d979030/pkg/detector/detector.go#L540

Happy path: Ran tests against the existing code — `ApplyPolicySucceed` passed in ~55 seconds.

Mutation test: Removed the target line from the source code so the test should fail. The mutation run failed at exactly 420.002 seconds (420s hardcoded `framework.PollTimeout`) on the `ApplyPolicySucceed` event check step.

Nice work. LGTM pending XiShanYongYe-Chang's comment: https://github.com/karmada-io/karmada/pull/7300#discussion_r3055197680

nXtCyberNet:
Hi @mohamedawnallah , thank you so much for the kind words and for taking the time to run mutation testing on the PR — that level of thoroughness is really appreciated and something I'll keep in mind for future contributions. Looking forward to making more meaningful PRs to the repo! 🙌

test/e2e/suites/base/propagationpolicy_test.go
XiShanYongYe-Chang: Hi @nXtCyberNet This use case was still wrongly deleted; we need keep it.

mohamedawnallah:
> Hi @mohamedawnallah , thank you so much for the kind words and for taking the time to run mutation testing on the PR — that level of thoroughness is really appreciated and something I'll keep in mind for future contributions. Looking forward to making more meaningful PRs to the repo! 🙌

Thanks @nXtCyberNet, PTAL at the comment again https://github.com/karmada-io/karmada/pull/7300#discussion_r3071818720. The PR merging is pending addressing that comment. Thanks

nXtCyberNet:
Hi, I believe the CI failures on Kubernetes v1.35.0 are due to environment instability rather than a code regression — I've opened issue #7388 to track this. If both of you agree, I'd love to investigate the flakiness further and work on a fix. Thanks!

test/e2e/suites/base/propagationpolicy_test.go
nXtCyberNet: sorry for the misunderstanding i have added that test  again!

mohamedawnallah:
Thanks @nXtCyberNet for addressing comment and having tracking issue documenting the observed flaky issues. It LGTM, just leaving space for @XiShanYongYe-Chang for any comments may have

XiShanYongYe-Chang:
Hi @nXtCyberNet can you help squash the commits into one?

mohamedawnallah:
@XiShanYongYe-Chang, I see @nXtCyberNet squashed the commits. Are we ready to merge?

XiShanYongYe-Chang:
Thanks for reminding @mohamedawnallah 
/approve

XiShanYongYe-Chang:
/retest
Leave lgtm to @mohamedawnallah for more check.

PR #7302, opened 2026-03-19
gracefuleviction: use optimistic lock in MergeFrom patch

### LINKED ISSUE #7290: [BUG]: Graceful eviction tasks silently dropped due to race condition - MergeFrom lacks optimistic locking
**What happened**:

I discovered a race condition in the graceful eviction controller that causes eviction tasks to be silently dropped during concurrent updates. The controller uses `client.MergeFrom()` without optimistic locking when updating `ResourceBinding.Spec.GracefulEvictionTasks`, which means it doesn't check for conflicts with other controllers (taint-manager and application failover controller) that write to the same field.

When the graceful eviction controller patches a binding based on a stale snapshot, it overwrites the entire `GracefulEvictionTasks` array—including tasks that were just added by other controllers. The Kubernetes API server applies the patch unconditionally because there's no `resourceVersion` in the diff.

I wrote tests to confirm this behavior. All three tests pass/fail as expected:
- `TestMergePatchLacksOptimisticLock` - confirms `MergeFrom()` produces JSON without resourceVersion
- `TestSyncBindingMergePatchDropsConcurrentEvictionTasks` - simulates the race, eviction task silently disappears
- `TestSyncBindingConcurrentWriteLoss_EndToEnd` - exercises real `syncBinding()` method, same result

---

Eviction tasks added by the taint-manager or application failover controller should never be lost. The graceful eviction controller should detect conflicts and retry with fresh data when concurrent updates occur.

---

**How to reproduce it**:

1. Deploy a workload to clusters A and B via ResourceBinding
2. Add a NoExecute taint to cluster B (simulating failure) - taint-manager adds an eviction task
3. Ensure the graceful eviction controller is processing the binding simultaneously (e.g., with an existing task from a prior operation completing)
4. Observe that the taint-manager's eviction task disappears from `Spec.GracefulEvictionTasks`
5. Verify the workload on cluster B is never evicted

Alternatively, you can increase the race window by adding latency between the graceful eviction controller's Get() and Patch() operations.

---

The issue affects both `rb_graceful_eviction_controller.go` and `crb_graceful_eviction_controller.go` at line 91-94. Both use the identical pattern without optimistic locking.

The fix is straightforward—wrap the patch in a `retry.RetryOnConflict` loop and use `client.MergeFromWithOptions(binding, client.MergeFromWithOptimisticLock{})` instead of bare `client.MergeFrom(binding)`. This ensures the API server rejects stale patches with 409 Conflict, and the controller retries with fresh data.

This is critical because the race occurs during cluster failures—exactly when graceful eviction must work correctly to evacuate workloads from tainted clusters.

---

**Environment**:
- Karmada version: Tested on main branch (commit hash available if needed)
- All 22 existing graceful eviction tests pass - no regressions
- Issue exists in both ResourceBinding and ClusterResourceBinding controllers

Ady0333:
Replace client.MergeFrom with client.MergeFromWithOptions using MergeFromWithOptimisticLock to include resourceVersion in the patch payload. This prevents the API server from applying a stale patch that could silently drop GracefulEvictionTasks added concurrently by the taint manager or application failover controller.

**What type of PR is this?**

/kind bug

---

This PR fixes a critical race condition in the graceful eviction controllers where eviction tasks are silently dropped during concurrent writes.

---

**Problem:**
The graceful eviction controllers (`rb_graceful_eviction_controller.go` and `crb_graceful_eviction_controller.go`) use `client.MergeFrom()` without optimistic locking to update `Spec.GracefulEvictionTasks`. This creates a JSON Merge Patch that omits `metadata.resourceVersion`, allowing the Kubernetes API server to apply patches unconditionally without conflict detection.

When three controllers write to the same field concurrently (taint manager, application failover controller, and graceful eviction controller), the last write wins without any conflict error, silently discarding tasks added by other controllers.

---

**Impact:**
- Workloads continue running on tainted/failing clusters that should have been evacuated
- Occurs during cluster failures — exactly when eviction matters most
- Lost eviction tasks cannot self-heal due to predicate filtering (only triggers when len(GracefulEvictionTasks) != 0)

---

**Solution:**
Replace `client.MergeFrom(binding)` with `client.MergeFromWithOptions(binding, client.MergeFromWithOptimisticLock{})`. This forces `metadata.resourceVersion` into the patch, enabling the API server to reject stale updates with HTTP 409 Conflict. The controller-runtime reconcile loop naturally retries with fresh data when conflicts occur.

---

Fixes #7290 

---

- **Minimal change:** Only 2 lines modified (1 per controller file, lines 91)
- **No retry logic needed:** Relies on controller-runtime's natural reconcile loop to handle 409 conflicts
- **All existing tests pass:** No test modifications required
- **Race scenario:** 
  1. Graceful eviction controller reads binding (RV=100), tasks = [taskA]
  2. Taint manager adds taskB via Update() → RV=101, tasks = [taskA, taskB]
  3. Before fix: MergeFrom patch succeeds (no RV check) → tasks = [], taskB lost
  4. After fix: Patch fails with 409 Conflict → reconcile loop retries with fresh data → taskB preserved

---

```release-note
Fixed a race condition where graceful eviction tasks could be silently dropped when multiple controllers concurrently modify the same ResourceBinding or ClusterResourceBinding, preventing workloads from being evacuated from tainted or failing clusters.
```

### DIFF
--- a/pkg/controllers/gracefuleviction/crb_graceful_eviction_controller.go
+++ b/pkg/controllers/gracefuleviction/crb_graceful_eviction_controller.go
@@ -88,7 +88,7 @@ func (c *CRBGracefulEvictionController) syncBinding(ctx context.Context, binding
 		return nextRetry(keptTask, c.GracefulEvictionTimeout, metav1.Now().Time), nil
 	}
 
-	objPatch := client.MergeFrom(binding)
+	objPatch := client.MergeFromWithOptions(binding, client.MergeFromWithOptimisticLock{})
 	modifiedObj := binding.DeepCopy()
 	modifiedObj.Spec.GracefulEvictionTasks = keptTask
 	err := c.Client.Patch(ctx, modifiedObj, objPatch)
--- a/pkg/controllers/gracefuleviction/rb_graceful_eviction_controller.go
+++ b/pkg/controllers/gracefuleviction/rb_graceful_eviction_controller.go
@@ -88,7 +88,7 @@ func (c *RBGracefulEvictionController) syncBinding(ctx context.Context, binding 
 		return nextRetry(keptTask, c.GracefulEvictionTimeout, metav1.Now().Time), nil
 	}
 
-	objPatch := client.MergeFrom(binding)
+	objPatch := client.MergeFromWithOptions(binding, client.MergeFromWithOptimisticLock{})
 	modifiedObj := binding.DeepCopy()
 	modifiedObj.Spec.GracefulEvictionTasks = keptTask
 	err := c.Client.Patch(ctx, modifiedObj, objPatch)


Ady0333:
Hello @XiShanYongYe-Chang , please review this pr once and let me know if any change is needed!!

XiShanYongYe-Chang:
Fix #7290

XiShanYongYe-Chang:
@Ady0333 would you like to do it? #7303

Ady0333:
> @Ady0333 would you like to do it? #7303

Yes @XiShanYongYe-Chang, I would like to do this.

PR #7306, opened 2026-03-19
Cherry pick of #7302 upstream release 1.15

Ady0333:
Cherry pick of #7302 on release-1.15.

#7302: Fix race condition in graceful eviction controllers

**Release note:**
```release-note
Fixed a race condition where graceful eviction tasks could be silently dropped when multiple controllers concurrently modify the same ResourceBinding or ClusterResourceBinding, preventing workloads from being evacuated from tainted or failing clusters.
```


Ady0333:
@XiShanYongYe-Chang , Cherry-pick for release -1.15

Issue #7321, opened 2026-03-23
Helm chart: v1beta2.custom.metrics.k8s.io APIService has raw {{ ca_crt }} placeholder in static-resources ConfigMap


rajsinghtech:
## What happened

When installing Karmada v1.17.0 via Helm with `metricsAdapter` enabled in components, the `karmada-static-resource` post-install Job fails because the `v1beta2.custom.metrics.k8s.io` APIService definition in the `karmada-static-resources` ConfigMap contains a raw `{{ ca_crt }}` template variable instead of the actual base64-encoded CA certificate.

The other APIServices in the same ConfigMap (`v1alpha1.cluster.karmada.io`, `v1beta1.metrics.k8s.io`, `v1beta1.custom.metrics.k8s.io`, `v1alpha1.search.karmada.io`) are rendered correctly with the real certificate.

The `karmada-static-resource` Job runs `kubectl apply -f /static-resources/` which hits the invalid YAML in the `karmada-aggregated-apiserver-apiservice.yaml` key and fails with:

```
error parsing /static-resources/karmada-aggregated-apiserver-apiservice.yaml: error converting YAML to JSON: yaml: invalid map key: map[interface {}]interface {}{"ca_crt":interface {}(nil)}
```

The Job exits with code 1.

## What you expected to happen

All APIService definitions in the `karmada-static-resources` ConfigMap should have the `{{ ca_crt }}` placeholder replaced with the actual base64-encoded CA certificate by the pre-install Job's init container (which runs `sed -i'' -e "s/{{ ca_crt }}/${karmada_ca}/g"` on the template files).

## How to reproduce

1. Install Karmada v1.17.0 via Helm with:
   ```yaml
   installMode: "host"
   certs:
     mode: "auto"
   components:
     - metricsAdapter
   ```
2. Observe the `karmada-static-resource` Job fails
3. Inspect the `karmada-static-resources` ConfigMap and note the `v1beta2.custom.metrics.k8s.io` APIService entry still has `{{ ca_crt }}` while others have the real cert

## Affected template files

- `charts/karmada/templates/_karmada_apiservice.tpl` - Defines all APIService resources including v1beta2.custom.metrics.k8s.io (line 55-70)
- `charts/karmada/templates/_helpers.tpl` - `karmada.apiserver.caBundle` helper (line 248-255) outputs the `{{ ca_crt }}` placeholder for `certs.mode: auto`
[...]

PR #7322, opened 2026-03-23
fix(helm): use | as sed delimiter for cert substitution

rajsinghtech:
## What type of PR is this?

/kind bug

The `sed` commands in the pre-install and pre-upgrade Jobs that substitute `{{ ca_crt }}` (and other cert placeholders) with actual base64-encoded certificate values use `/` as the delimiter. Since base64-encoded values can contain `/` characters, this breaks the `sed` substitution pattern, leaving raw `{{ ca_crt }}` placeholders in the `karmada-static-resources` ConfigMap.

When the `karmada-static-resource` Job subsequently runs `kubectl apply -f /static-resources/`, it fails with:

```
error converting YAML to JSON: yaml: invalid map key: map[interface {}]interface {}{"ca_crt":interface {}(nil)}
```

This is because `{{ ca_crt }}` is interpreted as a YAML flow mapping rather than a string value.

## How this PR fixes it

Switches all `sed` commands in `pre-install-job.yaml` and `pre-upgrade-job.yaml` from using `/` to `|` as the delimiter. The `|` character cannot appear in base64-encoded output (`A-Za-z0-9+/=`), making it a safe delimiter choice.

Fixes https://github.com/karmada-io/karmada/issues/7321

```release-note
Fixed Helm chart pre-install and pre-upgrade Jobs where certificate placeholder substitution could fail due to sed delimiter conflicts with base64-encoded values.
```

### DIFF
--- a/charts/karmada/templates/pre-install-job.yaml
+++ b/charts/karmada/templates/pre-install-job.yaml
@@ -453,20 +453,20 @@ spec:
           front_proxy_ca=$(base64 /opt/certs/front-proxy-ca.crt | tr -d '\r\n')
           front_proxy_client_crt=$(base64 /opt/certs/front-proxy-client.pem | tr -d '\r\n')
           front_proxy_client_key=$(base64 /opt/certs/front-proxy-client-key.pem | tr -d '\r\n')
-          sed -i'' -e "s/{{ print "{{ ca_crt }}" }}/${karmada_ca}/g" /opt/configs/cert.yaml
-          sed -i'' -e "s/{{ print "{{ ca_key }}" }}/${karmada_ca_key}/g" /opt/configs/cert.yaml
-          sed -i'' -e "s/{{ print "{{ crt }}" }}/${karmada_crt}/g" /opt/configs/cert.yaml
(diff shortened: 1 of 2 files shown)


XiShanYongYe-Chang:
Thank you for the quick fix.

zhzhuang-zju:
Thanks @rajsinghtech. 

> The sed commands in the pre-install and pre-upgrade Jobs that substitute {{ ca_crt }} (and other cert placeholders) with actual base64-encoded certificate values use / as the delimiter. Since base64-encoded values can contain / characters, this breaks the sed substitution pattern, leaving raw {{ ca_crt }} placeholders in the karmada-static-resources ConfigMap.

I'm fine with replacing `/` with `|`. It seems to have no side effects and can eliminate concerns about `/` breaking the `sed` substitution pattern.

However, I wonder if `/` actually breaks the `sed` substitution pattern. Based on the analysis in the comment https://github.com/karmada-io/karmada/pull/7185#discussion_r2939952157 and years of CI testing, this issue shouldn't occur. Please correct me if my analysis is wrong.

rajsinghtech:
Closing — can't confirm the root cause and the maintainers couldn't reproduce. The `|` delimiter change is defensive but not fixing a verified bug.

PR #7323, opened 2026-03-23
fix: recompute AggregatedStatus inside RetryOnConflict to avoid stale health evaluation

### LINKED ISSUE #7320: Stale AggregatedStatus persisted on conflict retry in binding-status-controller
## **What happened**:

I found a race condition in `pkg/util/helper/workstatus.go` where `aggregatedStatuses` is computed outside the `RetryOnConflict` loop. When a conflict retry occurs, the status update persists stale Work health data, overwriting fresher status written by concurrent controllers.

This causes the graceful eviction controller to read incorrect health status and potentially delete workloads from the old cluster before the replacement cluster is actually healthy.

---

## **What you expected to happen**:

On conflict retry, `aggregatedStatuses` should be recomputed from fresh Work objects so that `binding.Status.AggregatedStatus` always reflects the current cluster health state at the time of the status write.

---

## **How to reproduce it**:

1. Deploy a workload with failover enabled across 3 clusters (A, B, C)
2. Taint cluster C to trigger eviction
3. Scheduler adds replacement cluster D, updates `SchedulerObservedGeneration`
4. Simultaneously, binding-status-controller fetches Work statuses (captures D as healthy)
5. Cluster D's health degrades (e.g., pod fails readiness probe)
6. Conflict occurs between scheduler's update and status controller's update
7. On retry, `UpdateStatus()` re-fetches the binding but writes the stale `aggregatedStatuses` showing D as healthy
8. Graceful eviction controller reads stale D=healthy, removes eviction task for C
9. Result: C's workload deleted while D is actually unhealthy

---

## **Anything else we need to know?**:

The root cause is in `AggregateResourceBindingWorkStatus()` (lines 64-81) and `AggregateClusterResourceBindingWorkStatus()` (lines 108-121):
```go
// computed ONCE, outside retry
workList, err := GetWorksByBindingID(ctx, c, binding.Labels[...], true)
aggregatedStatuses, err := assembleWorkStatus(workList.Items, binding.Spec.Resource)

// retry loop uses stale aggregatedStatuses
if err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
    operationResult, err = UpdateStatus(ctx, c, binding, func() error {
        binding.Status.AggregatedStatus = aggregatedStatuses  // STALE on retry
        ...
    })
    return err
})
```

The fix is to move `GetWorksByBindingID()` and `assembleWorkStatus()` inside the retry closure, similar to how I fixed the `generateFullyAppliedCondition()` staleness in #7226.

---

## **Environment**:
- Karmada version: master (latest)
- Others: Affects all versions with graceful eviction enabled

Ady0333:
## **What type of PR is this?**

/kind bug

---

## **What this PR does / why we need it**:

This PR fixes a race condition in `pkg/util/helper/workstatus.go` where `AggregatedStatus` is computed outside the `RetryOnConflict` loop, causing stale cluster health data to be persisted on conflict retries.

---

## **The Problem:**

In `AggregateResourceBindingWorkStatus()` and `AggregateClusterResourceBindingWorkStatus()`, `GetWorksByBindingID()` and `assembleWorkStatus()` are called once before the retry loop. On conflict retry:
1. `UpdateStatus()` re-fetches the binding (fresh ResourceVersion)
2. The mutation closure writes stale `aggregatedStatuses` captured before the retry
3. This overwrites fresher health status written by concurrent controllers

---

## **Impact:**

During cluster failover with concurrent status updates, if a replacement cluster's health degrades between the initial read and conflict retry, the stale healthy status gets persisted. The graceful eviction controller may then prematurely remove eviction tasks, causing workloads to be deleted from the old cluster before the replacement is ready.

---

## **The Fix:**

Moved `GetWorksByBindingID()` and `assembleWorkStatus()` inside the retry closure to ensure fresh Work data on each retry attempt, following the same pattern as #7226.

## **Which issue(s) this PR fixes**:

Fixes #7320 

---

## **Special notes for your reviewer**:

Added 3 tests proving the bug exists and the fix works:
- `TestAggregateResourceBindingWorkStatus_RecomputesOnConflictRetry` - Simulates conflict with Work health degradation, verifies fresh status on retry
- `TestAggregateClusterResourceBindingWorkStatus_RecomputesOnConflictRetry` - Same for ClusterResourceBinding path
- `TestAggregateResourceBindingWorkStatus_MultiClusterFailover` - Multi-cluster failover scenario where health changes during conflict

All tests **fail before the fix** (demonstrating the bug) and **pass after the fix** (verifying the solution).

This follows the exact pattern as #7226 where we moved `generateFullyAppliedCondition()` inside the retry loop for the same reason.

---

## **Does this PR introduce a user-facing change?**:
```release-note
Fixed a race condition where stale cluster health status could be persisted during conflict retries in binding-status-controller, preventing incorrect graceful eviction decisions during cluster failover.
```

### DIFF
--- a/pkg/util/helper/workstatus.go
+++ b/pkg/util/helper/workstatus.go
@@ -61,17 +61,18 @@ func AggregateResourceBindingWorkStatus(
 	binding *workv1alpha2.ResourceBinding,
 	eventRecorder record.EventRecorder,
 ) error {
-	workList, err := GetWorksByBindingID(ctx, c, binding.Labels[workv1alpha2.ResourceBindingPermanentIDLabel], true)
-	if err != nil {
-		return err
-	}
-	aggregatedStatuses, err := assembleWorkStatus(workList.Items, binding.Spec.Resource)
-	if err != nil {
-		return err
-	}
-
 	var operationResult controllerutil.OperationResult
-	if err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
+	bindingID := binding.Labels[workv1alpha2.ResourceBindingPermanentIDLabel]
+	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		workList, err := GetWorksByBindingID(ctx, c, bindingID, true)
+		if err != nil {
+			return err
+		}
+		aggregatedStatuses, err := assembleWorkStatus(workList.Items, binding.Spec.Resource)
+		if err != nil {
+			return err
+		}
+
 		operationResult, err = UpdateStatus(ctx, c, binding, func() error {
 			binding.Status.AggregatedStatus = aggregatedStatuses
 			// set binding status with the newest condition
@@ -79,7 +80,8 @@ func AggregateResourceBindingWorkStatus(
 			return nil
 		})
 		return err
-	}); err != nil {
+	})
+	if err != nil {
 		eventRecorder.Event(binding, corev1.EventTypeWarning, events.EventReasonAggregateStatusFailed, err.Error())
 		return err
 	}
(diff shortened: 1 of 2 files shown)


Ady0333:
@XiShanYongYe-Chang 

I've created this draft PR to demonstrate the issue with concrete code and tests.

The test shows that when a conflict retry occurs and Work health changes between the initial read and retry, the stale status gets persisted. The test:
1. Starts with a Work showing Healthy status
2. Simulates a conflict on the first status update
3. Degrades the Work to Unhealthy during the retry window
4. Verifies that without the fix, stale Healthy status would be written (demonstrating the bug)

The fix moves the Work status assembly inside the retry closure, following the same pattern as #7226. This ensures each retry uses fresh data consistent with the re-fetched binding.

I'd appreciate your review on whether this scenario aligns with real-world failover concerns. The test proves the race is real—the question is whether the timing window matters in practice for graceful eviction decisions.

XiShanYongYe-Chang:
Hi @Ady0333 thanks~

As I explained in the issue, I understand what you mean. I think this does not constitute a problem in the current scenario, so I do not recommend modifying this logic.

Ady0333:
Thanks for reviewing @XiShanYongYe-Chang! I understand your position. While the race condition is technically present, if you don't see it causing issues in practice, I'll close this PR.

Feel free to reference this if the scenario comes up in the future.

PR #6597, opened 2025-08-03
Improve the performance of karmada-scheduler

LivingCcj:
**What type of PR is this?**
/kind feature

-->

**What this PR does / why we need it**:
The PR 
**Which issue(s) this PR fixes**:
related to issue #5202 

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
`karmada-scheduler`: Optimized the scheduler step,  invoking calculate `availableRelicas` is reduced to once, improving the performance.
```

PR #7327, opened 2026-03-25
[Proposal] Binding preemption Design for Karmada scheduler

seanlaii:
**What type of PR is this?**
/kind documentation
/kind feature

This PR adds a design proposal for binding-level preemption in the Karmada scheduler. 
Currently, priority-based scheduling only orders the queue — a high-priority binding that arrives after cluster resources are consumed by lower-priority bindings remains pending indefinitely. This proposal introduces preemption so the scheduler can evict lower-priority bindings to make room, scoped to **single-cluster Divided** scheduling in Phase 1 with estimator-based node-level simulation planned for Phase 2.

Fixes #

```release-note

```

### DIFF
--- a/docs/proposals/scheduling/binding-preemption/binding-preemption.md
+++ b/docs/proposals/scheduling/binding-preemption/binding-preemption.md
@@ -1,0 +1,765 @@
+---
+title: Binding Preemption for Karmada Scheduler
+
+authors:
+- "@seanlaii"
+
+reviewers:
+- "@RainbowMango"
+- "@XiShanYongYe-Chang"
+- "@zhzhuang-zju"
+- "@whitewindmills"
+
+approvers:
+
+creation-date: 2026-03-22
+---
+
+# Binding Preemption for Karmada Scheduler
+
+## Table of Contents
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+- [Background](#background)
+- [Design Overview](#design-overview)
+- [How to Enable Preemption](#how-to-enable-preemption)
+- [API Changes](#api-changes)
+- [Phase 1: Summary-Based Preemption](#phase-1-summary-based-preemption)
+- [Phase 2: Estimator-Based Preemption](#phase-2-estimator-based-preemption)
+- [Why ClusterAffinities Is Excluded](#why-clusteraffinities-is-excluded)
+- [Alternative Designs Considered](#alternative-designs-considered)
+- [Risks and Limitations](#risks-and-limitations)
+- [Backward Compatibility](#backward-compatibility)
+- [Rollout Plan](#rollout-plan)
+- [Test Plan](#test-plan)
+- [Observability](#observability)
+
+---
+
+## Summary
+
+This proposal introduces **binding-level preemption** to the Karmada scheduler, scoped to **single-cluster scheduling scenarios**. When a high-priority ResourceBinding cannot be scheduled on its target cluster due to insufficient resources, the scheduler may evict lower-priority bindings from that cluster to make room.
+
+The proposal is phased:
+- **Phase 1**: Summary-based preemption using aggregate replica arithmetic for victim selection.
+- **Phase 2**: Estimator-based preemption with node-level simulation for precise victim selection.
+
+Priority is sourced from the existing `SchedulePriority` mechanism in `PropagationPolicy`, resolved via Kubernetes `PriorityClass`, and propagated to `ResourceBindingSpec.SchedulePriority.Priority`.
+
+## Motivation
+
+Karmada v1.13 introduced priority-based scheduling (`PriorityBasedScheduling` feature gate), which orders the scheduling queue by priority. However, a high-priority binding that arrives after cluster resources are consumed by low-priority bindings will remain pending indefinitely. This is especially painful for batch/AI workloads where GPU resources are scarce. Binding preemption closes this gap.
+
+### Example: GPU Training Preemption
+
+Consider a Karmada environment with a single GPU cluster (`gpu-cluster`, 8 GPUs) where all workloads target this cluster via `MaxGroups == 1`:
+
+**Initial state** — the cluster is full with low-priority batch jobs:
+
+| Binding | Workload | Priority | GPUs | PriorityClass |
+|---|---|---|---|---|
+| batch-job-1 | Deployment (2 replicas, 2 GPU each) | 100 | 4 | `batch` (preemptionPolicy: Never) |
+| batch-job-2 | Deployment (2 replicas, 2 GPU each) | 100 | 4 | `batch` (preemptionPolicy: Never) |
+
+Total used: 8/8 GPUs. No capacity remaining.
+
+**A high-priority training job arrives**:
+
+```yaml
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: training-job
+spec:
+  replicas: 2
+  template:
+    spec:
+      containers:
+      - resources:
+          requests:
+            nvidia.com/gpu: 2
+---
+apiVersion: policy.karmada.io/v1alpha1
+kind: PropagationPolicy
+metadata:
+  name: training-policy
+spec:
+  schedulePriority:
+    priorityClassSource: KubePriorityClass
+    priorityClassName: critical-training  # priority: 1000, preemptionPolicy: PreemptLowerPriority
+  placement:
+    clusterAffinity:
+      clusterNames: [gpu-cluster]
+    spreadConstraints:
+    - spreadByField: Cluster
+      maxGroups: 1
+  resourceSelectors:
+  - apiVersion: apps/v1
+    kind: Deployment
+    name: training-job
+```
+
+**Key points in the PropagationPolicy**:
+- `priorityClassName: critical-training` — references the PriorityClass that has `preemptionPolicy: PreemptLowerPriority`. This is the opt-in.
+- `maxGroups: 1` — declares single-cluster intent. Without this, the binding could reroute to other clusters on retry, making preemption unnecessary. **Preemption requires explicit `MaxGroups == 1`.**
+
+**Without preemption**: `training-job` stays pending indefinitely — the cluster is full and priority only affects queue ordering, not resource allocation.
+
+**With preemption (this proposal)**:
+
+1. Scheduler detects `assignReplicas` failure for `training-job` (priority 1000).
+2. Scheduler finds lower-priority bindings on `gpu-cluster`: `batch-job-1` (100), `batch-job-2` (100).
+3. `training-job` needs 4 GPUs. Either batch job alone frees 4 GPUs — sufficient. The reprieve loop keeps the older job (`batch-job-1`) and selects `batch-job-2` as the victim.
+4. Scheduler evicts `batch-job-2` from `gpu-cluster` via GracefulEviction and records a claim.
+5. After `batch-job-2`'s resources are freed, `training-job` schedules successfully on `gpu-cluster`.
+6. `batch-job-2` reschedules to another cluster (if available) or stays pending.
+
+### Goals
+
+1. Define the preemption model for Karmada bindings in single-cluster scenarios.
+2. Reuse the existing `SchedulePriority` / `PriorityClass` mechanism for preemption ordering.
+3. Add a `PreemptionPolicy` field to `ResourceBindingSpec.SchedulePriority`.
+4. Implement preemption inside `Schedule()` when replica assignment fails.
+5. Integrate with GracefulEviction for victim eviction.
+6. Gate behind `PriorityBasedPreemptiveScheduling` (alpha, default off).
+
+### Non-Goals
+
+1. **Multi-cluster preemption**: Coordinated victim eviction across multiple clusters. Requires iterative recalculation.
+2. **Duplicated mode preemption**: `assignByDuplicatedStrategy` assigns full replicas to all candidates and never returns an error. The preemption trigger is unreachable.
+3. **StaticWeight preemption**: Divides by weight ratio without capacity checks. Never returns an error.
+4. **Component scheduling preemption**: Multi-component workloads (`len(Components) > 1`) bypass `assignFunc` in `AssignReplicas` entirely. Supporting them as preemptors requires a different trigger mechanism. See [Component Scheduling Path](#component-scheduling-path) for the future extension path.
+5. Preemption for ClusterResourceBinding.
+6. Gang preemption.
+7. Cross-scheduler preemption coordination.
+8. PodDisruptionBudget-aware preemption at the Karmada level.
+9. **ClusterAffinities preemption**: See [Why ClusterAffinities Is Excluded](#why-clusteraffinities-is-excluded).
+
+## Background
+
+### Current Scheduling Pipeline
+
+```
+ResourceBinding enqueued → Pop (priority-ordered) → findClustersThatFit (Filter)
+→ prioritizeClusters (Score) → selectClusters (SpreadConstraints + capacity)
+→ assignReplicas (Duplicated/Aggregated/StaticWeight/DynamicWeight) → Write result
+```
+
+There is no preemption phase. If scheduling fails, the binding enters `backoffQ` or `unschedulableBindings`.
+
+**Where capacity is checked**: Filter plugins do **not** check resource capacity. Capacity is first evaluated in `selectClusters` via `calAvailableReplicas` (which queries the general-estimator and/or scheduler-estimator for `AllocatableReplicas`), then checked again in `assignReplicas` via `dynamicDivideReplicas`.
+
+### Assignment Strategies and Error Behavior
+
+| Strategy | Checks capacity? | Returns `UnschedulableError`? |
+|---|---|---|
+| **Aggregated** | Yes (via `dynamicDivideReplicas`) | Yes |
+| **DynamicWeight** | Yes (via `dynamicDivideReplicas`) | Yes |
+| **StaticWeight** | No — divides by weight ratio | Never |
+| **Duplicated** | No — assigns full replicas to all | Never |
+
+`dynamicDivideReplicas` is the **only place** in the assignment flow that returns `UnschedulableError`. This is the preemption trigger point.
+
+### Existing Priority Scheduling
+
+`PropagationPolicy.spec.schedulePriority` specifies a `PriorityClassSource` and `PriorityClassName`. The ResourceDetector resolves this to a numeric priority written to `ResourceBindingSpec.SchedulePriority.Priority`. The priority queue orders bindings by priority descending, FIFO for ties.
+
+### Prior Art
+
+| System | Preemption unit | Trigger | Post-preemption coordination |
+|---|---|---|---|
+| **Kubernetes** | Pod on a node | PostFilter (all nodes rejected) | NominatedNodeName + in-memory nominator |
+| **Kueue** | Workload (atomic) | Insufficient ClusterQueue quota | preemptionExpectations (UID tracking) |
+| **Volcano** | Task within Job | Preempt/Reclaim actions | Transaction commit/rollback |
+
+Related proposals: [Binding Priority and Preemption](../binding-priority-preemption/README.md) (direct predecessor), [Policy Preemption](../policy-preemption/README.md), [Multiple Pod Templates Scheduling](../multi-podtemplate-support/multiple-pod-template-support.md).
+
+**Note**: Policy preemption (`PolicyPreemption` feature gate) resolves "which policy controls a resource" at the ResourceDetector level. Binding preemption (this proposal) resolves "which workload gets cluster resources when scarce" at the scheduler level. They are complementary and independent.
+
+## Design Overview
+
+```
+High-priority binding cannot schedule (assignReplicas fails)
+        │
+        ▼
+┌─── Preemption Trigger ───────────────────────────────────────────────┐
+│  Is preemption applicable?                                           │
+│  (workload? single-component? ClusterAffinity? MaxGroups==1?         │
+│   PreemptLowerPriority? feature gates enabled?)                      │
+└───────────────────────┬──────────────────────────────────────────────┘
+                        │ yes
+                        ▼
+┌─── Victim Selection ─────────────────────────────────────────────────┐
+│  1. Find lower-priority bindings on target cluster (via index)       │
+│  2. Check: evicting all candidates frees enough resources?           │
+│  3. Reprieve highest-priority candidates first (minimum victim set)  │
+│                                                                      │
+│  Phase 1: Aggregate replica arithmetic                               │
+│  Phase 2: Estimator gRPC → node-level simulation (preferred)         │
+│           Falls back to Phase 1 if estimator unavailable             │
+└───────────────────────┬──────────────────────────────────────────────┘
+                        │ victims found
+                        ▼
+┌─── Preemption Claims ────────────────────────────────────────────────┐
+│  Record in-memory claim on target cluster (10-min TTL)               │
+│  → Prevents re-preemption on retry                                   │
+│  → Reserves capacity (lower-priority bindings see reduced capacity)  │
+│  → Replaced (not accumulated) if target cluster changes              │
+└───────────────────────┬──────────────────────────────────────────────┘
+                        │
+                        ▼
+┌─── Preemption Execution ─────────────────────────────────────────────┐
+│  For each victim: GracefulEvictCluster (30s grace period)            │
+│  Preemptor enters unschedulableBindings (waits for resources)        │
+│  On retry: victim resources freed → preemptor schedules successfully │
+└──────────────────────────────────────────────────────────────────────┘
+```
+
+**Key concepts**:
+
+- **Preemption trigger**: `assignReplicas` failure (only Aggregated/DynamicWeight can fail).
+- **Victim selection**: "Remove all, then reprieve" greedy heuristic (same as Kubernetes `SelectVictimsOnNode`).
+- **Preemption claims**: In-memory capacity reservation (equivalent to Kubernetes' in-memory pod nominator).
+- **Preemption execution**: Existing GracefulEviction mechanism. Per-cluster, all-or-nothing (consistent with Kubernetes per-pod preemption).
+- **Cross-kind, cross-namespace**: Preemption is purely priority-based with no kind or namespace restrictions, consistent with Kubernetes.
+
+## How to Enable Preemption
+
+1. **Enable feature gates** on the Karmada scheduler: `--feature-gates=PriorityBasedScheduling=true,PriorityBasedPreemptiveScheduling=true`
+2. **Create PriorityClasses** with `preemptionPolicy: PreemptLowerPriority` (for preemptors) or `preemptionPolicy: Never` (for non-preemptors). Note: Kubernetes defaults `preemptionPolicy` to `PreemptLowerPriority` when unset, so omitting the field enables preemption.
+3. **Reference the PriorityClass** in PropagationPolicy via `schedulePriority.priorityClassSource: KubePriorityClass`.
+4. **Declare single-cluster intent** with `maxGroups: 1` in `placement.spreadConstraints`.
+
+## API Changes
+
+### Add `PreemptionPolicy` to `SchedulePriority`
+
+```go
+type PreemptionPolicy string
+
+const (
+    PreemptLowerPriority PreemptionPolicy = "PreemptLowerPriority"
+)
+
+type SchedulePriority struct {
+    Priority         int32            `json:"priority,omitempty"`
+    PreemptionPolicy PreemptionPolicy `json:"preemptionPolicy,omitempty"`
+}
+```
+
+- Defaults to unset (empty string), meaning the binding never preempts. No explicit "Never" constant is needed — the zero value naturally means "no preemption."
+- The ResourceDetector maps `PreemptionPolicy` from the PriorityClass's `preemptionPolicy` field when `PriorityBasedPreemptiveScheduling` is enabled. Kubernetes defaults `preemptionPolicy` to `PreemptLowerPriority` when unset, so any PriorityClass without explicit `preemptionPolicy: Never` enables preemption at the binding level. The feature gate (alpha, default off) provides the primary opt-in control.
+
+### New Constants and Feature Gate
+
+```go
+const (
+    EvictionProducerScheduler      = "Scheduler"
+    EvictionReasonBindingPreempted = "BindingPreempted"
+)
+
+// Feature gate (alpha, default off). Requires PriorityBasedScheduling.
+PriorityBasedPreemptiveScheduling featuregate.Feature = "PriorityBasedPreemptiveScheduling"
+```
+
+### Extended ScheduleResult
+
+```go
+type ScheduleResult struct {
+    SuggestedClusters []workv1alpha2.TargetCluster
+    // Non-nil when preemption was performed.
+    // When set, SuggestedClusters is nil — the preemptor is NOT yet scheduled.
+    PreemptionResult  *PreemptionResult
+}
+
+type PreemptionResult struct {
+    Cluster string
+    Victims []VictimBinding
+}
+
+type VictimBinding struct {
+    Namespace string
+    Name      string
+    Replicas  int32
+    Priority  int32
+}
+```
+
+## Phase 1: Summary-Based Preemption
+
+### Scope and Applicability
+
+Preemption is attempted only when **all** of the following hold:
+
+1. The binding is a workload (`IsWorkload() == true`).
+2. The binding is single-component (`len(Components) <= 1`). Multi-component workloads bypass `assignFunc` in `AssignReplicas` entirely — they cannot reach the preemption trigger. This is a **Phase 1 limitation**, not a permanent design constraint. See [Component Scheduling Path](#component-scheduling-path) for the extension path.
+3. The binding uses `ClusterAffinity`, not `ClusterAffinities`. See [Why ClusterAffinities Is Excluded](#why-clusteraffinities-is-excluded).
+4. The binding explicitly targets one cluster via `SpreadConstraints` (`MaxGroups == 1`).
+5. `assignReplicas` returns an error. The primary target is Aggregated mode with single-cluster scheduling. DynamicWeight also uses `dynamicDivideReplicas` internally, but with `MaxGroups == 1` it behaves identically to Aggregated.
+
+```go
+func isPreemptionApplicable(spec *workv1alpha2.ResourceBindingSpec) bool {
+    if !spec.IsWorkload() { return false }
+    if len(spec.Components) > 1 { return false }
+    if spec.Placement == nil { return false }
+    if spec.Placement.ClusterAffinities != nil { return false }
+    for _, sc := range spec.Placement.SpreadConstraints {
+        if sc.SpreadByField == policyv1alpha1.SpreadByFieldCluster &&
+            sc.MaxGroups == 1 {
+            return true
+        }
+    }
+    return false
+}
+```
+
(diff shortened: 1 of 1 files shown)


RainbowMango:
@seanlaii, I see this PR marked with `Draft`. Is there anything you want to update? Please also cc @whitewindmills once it's ready for review. 

PS: I can get back to this next week.

seanlaii:
> @seanlaii, I see this PR marked with `Draft`. Is there anything you want to update? Please also cc @whitewindmills once it's ready for review.
> 
> PS: I can get back to this next week.

Thanks for the info! @whitewindmills @RainbowMango  Please help review when you have a chance.
Also cc @shellfish007 to take a look if interested

shellfish007:
Thanks for the proposal! I'm still getting familiar with the codebase, so apologies if any of this is already covered somewhere. 
A couple of questions:

1. Cross-namespace preemption: Is it guaranteed that preemption victims are always ResourceBindings in the same namespace as the preemptor, or could a higher-priority workload preempt jobs from a different namespace? If cross-namespace preemption is possible, would it make sense to have a configurable option to restrict preemption within namespace boundaries? Thinking about multi-tenancy scenarios where teams could unintentionally (or intentionally) affect each other's workloads by simply raising their job priorities.
2. Namespace quota as a preemption trigger: It looks like preemption is only triggered when cluster-level resource assignment fails (assignReplicas). Would namespace quota exhaustion also be considered as a trigger? A higher-priority workload could remain pending not because the cluster is full, but because its namespace quota is exhausted — in that case preemption would never fire even if it should. Is this a known limitation or out of scope for now?                                              
                                                                                                                                                                                                                      
Happy to be corrected if either of these is already handled!

seanlaii:
> Thanks for the proposal! I'm still getting familiar with the codebase, so apologies if any of this is already covered somewhere. A couple of questions:
> 
> 1. Cross-namespace preemption: Is it guaranteed that preemption victims are always ResourceBindings in the same namespace as the preemptor, or could a higher-priority workload preempt jobs from a different namespace? If cross-namespace preemption is possible, would it make sense to have a configurable option to restrict preemption within namespace boundaries? Thinking about multi-tenancy scenarios where teams could unintentionally (or intentionally) affect each other's workloads by simply raising their job priorities.
> 2. Namespace quota as a preemption trigger: It looks like preemption is only triggered when cluster-level resource assignment fails (assignReplicas). Would namespace quota exhaustion also be considered as a trigger? A higher-priority workload could remain pending not because the cluster is full, but because its namespace quota is exhausted — in that case preemption would never fire even if it should. Is this a known limitation or out of scope for now?
> 
> Happy to be corrected if either of these is already handled!

Thanks for the great questions! These are both important considerations.

### Cross-namespace preemption
In the current design, `Preemption` is cross-namespace, consistent with `Kubernetes`. Also, `PriorityClass` is a `cluster-scoped` resource, so priority is not `namespace-bounded`. In Kubernetes, pod preemption can evict pods from any namespace on the same node, and the community's recommended approach for multi-tenancy protection is access control (e.g., admission webhooks restricting which `PriorityClasses` a namespace can use), rather than restricting preemption scope in the scheduler.

We follow the same approach here. That said, if namespace-scoped restriction becomes important based on real-world user feedback, it can be added as a configurable option (e.g., `preemptionScope: Namespace | Cluster`) in a follow-up without breaking changes. The current cross-namespace behavior would remain the default. I've added a section in the design doc to make this rationale explicit.

### Namespace quota as a preemption trigger
This is also a very good question! You're right, in the current design, preemption is only triggered by `cluster-level` resource exhaustion (`assignReplicas failure`), not by namespace quota exhaustion. This is consistent with Kubernetes, where `ResourceQuota` is enforced by an admission controller that is independent of the scheduler's preemption logic.

However, there is a subtle architectural difference worth noting:

- In Kubernetes: Quota is consumed at pod creation time (before scheduling), so the preemptor's quota is already reserved before preemption runs — there is no conflict between preemption and quota.
- In Karmada: `FederatedResourceQuota` is checked at binding patch time (during scheduling), so a theoretical edge case exists where preemption frees cluster resources but the preemptor's namespace quota is still insufficient.

This might be rare in practice (requires both cluster capacity and namespace quota to be exhausted simultaneously) and the system is self-healing (evicted victims reschedule, preemptor retries). Properly addressing it would require changes to the quota architecture (e.g., quota reservation at binding creation time), which is beyond the scope of this proposal.

I've added both topics to the Risks and Limitations section. Thanks again for the thoughtful review!

whitewindmills:
@seanlaii Thanks for the detailed proposal. Overall, I think the direction makes sense: extending priority-based scheduling from queue ordering to binding-level preemption is a real gap for scarce-resource workloads, especially GPU/batch scenarios.

Before moving this forward, I think we should clarify and tighten several points.

## Major Design Risks / Gaps

1. **Phase 1 victim accounting may be inaccurate.**

   The proposal currently selects victims mostly by replica count. However, `AllocatableReplicas` is calculated against the preemptor's `ReplicaRequirements`, while each victim may have very different per-replica resource requirements. Evicting 10 small victim replicas does not necessarily free enough resources for 10 preemptor replicas.

   I think Phase 1 should either:
   - convert each victim's `ReplicaRequirements * replicasOnCluster` into an equivalent number of preemptor replicas, or
   - explicitly document that Phase 1 only works reliably when victims and preemptors have comparable per-replica requirements.

2. **Claim accounting for equal-priority bindings needs clarification.**

   The current claim deduction only affects lower-priority bindings. This allows another binding with the same priority to consume the claimed capacity before the original preemptor is retried, especially because the preemptor may stay in `unschedulableBindings` until the periodic flush.

   I suggest claim deductions should apply to `priority <= claim.priority`, except for the claim owner itself. Otherwise, the claim does not fully protect the preemptor's nominated capacity.

3. **The `mayPreempt()` retry path may change behavior for unsupported scenarios.**

   Retrying `SelectBestClusters(..., InvalidReplicas)` before knowing whether the final selection is truly single-cluster may affect multi-cluster workloads that will later fail `preemptionEnabled()`. This may change error paths and diagnostics even when preemption is not actually supported.

   Can we make `mayPreempt()` stricter, for example by requiring `MaxGroups == 1` or `len(feasibleClusters) == 1` before bypassing the capacity check?

4. **The 5-minute preemption delay should probably be handled in Phase 1.**

   The proposal mentions that `ResourceSummary` changes do not currently requeue unschedulable bindings, so the preemptor may wait up to the 5-minute unschedulable flush interval. For preemption, especially GPU/batch workloads, this latency is quite visible.

   I would prefer adding ResourceSummary-change requeue as part of Phase 1, or at least making it an explicit required follow-up before enabling this feature beyond alpha.

5. **Partial-cluster eviction blast radius should be documented clearly.**

   `GracefulEvictCluster` removes the victim's assignment from the target cluster. If the preemptor only needs a small amount of capacity, we may still evict all victim replicas from that cluster. That is acceptable as a Phase 1 tradeoff, but it should be clearly documented as a limitation.

6. **Claim identity should use a stable binding identity.**

   The proposal uses `Kind/Namespace/Name` as the preemptor claim key. This may be insufficient for recreated workloads or same kind/name across API groups. I suggest using the `ResourceBinding` namespace/name plus UID, or the permanent-id label, as the claim identity.

7. **Preemption execution should handle patch failures more explicitly.**

   If all victim patches fail, the claim is still set and capacity is reserved for up to the TTL even though no victim is actually being evicted. I think `executePreemption` should return the number of successfully patched victims and an error. If zero victim patches succeed, the scheduler should clear the claim and treat the scheduling attempt as failed.

## Development Plan Suggestions

I suggest splitting the implementation into smaller, reviewable pieces:

1. API and feature gate:
   - add `PreemptionPolicy`
   - add generated artifacts
   - add `PriorityBasedPreemptiveScheduling`
   - clearly define dependency on `PriorityBasedScheduling`

2. Priority resolution:
   - map Kubernetes `PriorityClass.preemptionPolicy`
   - confirm whether `PropagationPolicy.spec.schedulePriority` should allow an explicit override

3. Scheduler Phase 1:
   - cluster-to-bindings index
   - applicability checks
   - victim filtering and selection
   - preemption claim store
   - claim-aware capacity calculation
   - `ScheduleResult.PreemptionResult`
   - scheduler main-loop handling and conditions

4. Eviction and observability:
   - `GracefulEviction` integration
   - events on preemptor and victims
   - scheduler metrics
   - clear handling of partial patch failures

5. Queue wake-up:
   - requeue relevant unschedulable bindings when `Cluster.Status.ResourceSummary` changes

6. Phase 2 should be a separate implementation milestone:
   - estimator `SelectVictims` API
   - pod-to-binding index
   - pod template label injection
   - estimator-side snapshot simulation

Phase 2 touches the estimator, interpreter path, and member-cluster pod indexing, so I would not bundle it with Phase 1 implementation.

## Test Plan Feedback

The proposed test plan is a good start. I think we should add coverage for these cases before considering the design complete:

1. Heterogeneous resources:
   - victim replicas require fewer resources than preemptor replicas
   - victim replicas require more resources than preemptor replicas
   - victim selection should not rely on raw replica count alone

2. Claim behavior:
   - lower-priority binding is blocked from consuming claimed capacity
   - equal-priority binding behavior is explicitly tested
   - higher-priority binding can override an existing claim
   - claim is cleared on successful scheduling
   - claim is cleared when the preemptor binding is deleted
   - claim replacement when the target cluster changes

3. Failure handling:
   - all victim patches fail
   - partial victim patch success
   - scheduler restart after preemption
   - claim TTL expiry and retry behavior

4. Unsupported scenarios:
   - no preemption for `Duplicated`
   - no preemption for `StaticWeight`
   - no preemption for `ClusterAffinities`
   - no preemption for multi-component preemptors
   - no preemption for `ClusterResourceBinding`
   - no preemption when feature gate is disabled
   - no preemption when `PreemptionPolicy` is `Never`

5. Queue and latency:
   - preemptor is requeued when victim resources are reflected in `ResourceSummary`
   - preemptor does not have to wait for the full unschedulable flush interval when resources are freed

6. Multi-tenancy and quota:
   - cross-namespace preemption behavior is covered by e2e or integration tests
   - quota rejection after preemption is documented and, if possible, covered by a recovery test

7. Observability:
   - event on preemptor when preemption starts
   - event on each victim
   - final successful scheduling event after resources are freed
   - metrics for attempts and victim count

I am supportive of the overall direction, but I think the Phase 1 resource accounting, claim semantics, requeue behavior, and patch-failure handling should be clarified before we proceed to implementation.

whitewindmills:
This is a complex and extensive proposal. Breaking it down into smaller, more manageable parts would likely expedite the review process.

_PS: To be honest, I often forget the previous context halfway through reading it._

RainbowMango (commented):
Start working on it. 
/assign

PS: 
I'm not sure if we can reach a consensus in release 1.18(DL: May 31st), but I do want to have it as soon as possible.

seanlaii:
@whitewindmills Thank you for the thorough review! I've addressed all seven points in the design document.

1. Phase 1 victim accounting
Good catch. The victim selection algorithm now operates on resource quantities instead of raw replica counts. The design sums freed resources across victims (`victim.ReplicaRequirements * victim.replicas`) and converts to preemptor-equivalent replicas using `calculateFittingReplicas` — which uses `floor` and `min` across resource types (the bottleneck resource determines how many preemptor replicas can fit). This is distinct from `calculateReplicasFromResources` (used for claim deductions, which uses ceil and max). The reprieve loop also operates on resource quantities, ensuring correct accounting when victims and preemptors have different per-replica sizes.

2. Equal-priority claim deductions
Agreed. I verified the Kubernetes source — `RunFilterPluginsWithNominatedPods` uses `priority >= nominatedPodPriority && UID != nominatedPodUID`. The design now aligns with this: `claim.priority >= currentPriority && claim.bindingKey != currentBindingKey`. This protects the preemptor's nominated capacity from both lower-priority and equal-priority bindings, while the self-exception ensures the preemptor is not penalized by its own claim when retrying.

3. `mayPreempt()` requires explicit `MaxGroups == 1`
Good suggestion — adopted. Both `mayPreempt()` (the `SelectBestClusters` retry gate) and `isPreemptionApplicable()` (the preemption gate) now require explicit `MaxGroups == 1` in `SpreadConstraints`. The runtime `selectedClusters == 1` fallback has been removed. Workloads **must** declare single-cluster intent explicitly to opt into preemption, consistent with component scheduling. Multiple clusters can still pass filters (the preemptor can have multiple candidates), but preemption is only attempted on the single cluster selected by SelectBestClusters. This avoids changing error paths for multi-cluster workloads.

4. 5-minute preemption delay
Agreed. This is being addressed in https://github.com/karmada-io/karmada/pull/7369, which adds `ResourceSummary` change detection to requeue preemptors immediately when victim resources are freed. Updated the design document to reference this PR.

5. Per-cluster eviction blast radius
Added a "Per-cluster eviction granularity" paragraph documenting that `GracefulEvictCluster` removes all victim replicas on the target cluster, even if the preemptor only needs partial capacity. This is consistent with Kubernetes where pod preemption is all-or-nothing per pod. Partial-replica eviction would require:
(1) a new API operation for graceful replica reduction,
(2) GracefulEviction controller support for scaling down, and
(3) a mechanism to prevent the victim's scheduler from scaling back up during the preemption window. This is deferred to a future proposal.

6. Claim identity
The claim key uses `Kind/Namespace/Name` from `spec.Resource` because `Schedule()` operates on `ResourceBindingSpec`, which does not carry binding metadata. Could you provide a specific scenario where this would be insufficient?
The cross-API-group collision (same Kind name in different API groups) is already prevented at the ResourceBinding layer — `GenerateBindingName(kind, name)` produces the same binding name for both, so the detector cannot create separate bindings for them. If they can't have separate bindings, they can't have separate claims. Using the binding's `ResourceBindingPermanentIDLabel` would require passing binding metadata into `Schedule()`, which is a larger refactor. I've added a "Claim identity" paragraph in the design document explaining this rationale.

7. Patch failure handling
Agreed. The design now specifies that `executePreemption` returns the eviction count. When zero victims are successfully evicted, the caller clears the preemption claim immediately to avoid holding a stale reservation for the full 10-minute TTL. The preemptor still enters `unschedulableBindings` and retries normally.

seanlaii:
> This is a complex and extensive proposal. Breaking it down into smaller, more manageable parts would likely expedite the review process.
> 
> _PS: To be honest, I often forget the previous context halfway through reading it._

Thanks for the feedback. I will try to break it down.

seanlaii:
> Start working on it. /assign
> 
> PS: I'm not sure if we can reach a consensus in release 1.18(DL: May 31st), but I do want to have it as soon as possible.

Thanks for the info and time! I have moved it to 1.19.

docs/proposals/scheduling/binding-preemption/binding-preemption.md
```diff
@@ -0,0 +1,995 @@
...
+
+The proposal is phased:
+- **Phase 1**: Summary-based preemption for single-cluster Divided scheduling with `ClusterAffinity` (the preemptor targets exactly one cluster).
+- **Phase 2**: Estimator-based preemption with node-level simulation for more precise victim selection.
+
+Priority is sourced from the existing `SchedulePriority` mechanism in `PropagationPolicy`, resolved via Kubernetes `PriorityClass`, and propagated to `ResourceBindingSpec.SchedulePriority.Priority`.
+
+## Motivation
+
+Karmada v1.13 introduced priority-based scheduling (`PriorityBasedScheduling` feature gate), which orders the scheduling queue by priority. However, a high-priority binding that arrives after cluster resources are consumed by low-priority bindings will remain pending indefinitely. This is especially painful for batch/AI workloads where GPU resources are scarce. Binding preemption closes this gap.
```
RainbowMango, line 52: By the way, are we ready to move this `PriorityBasedScheduling` to beta now? 

If not, what else do we need to do?
@seanlaii @whitewindmills

docs/proposals/scheduling/binding-preemption/binding-preemption.md
```diff
@@ -0,0 +1,995 @@
...
+| [Multiple Pod Templates Scheduling](../multi-podtemplate-support/multiple-pod-template-support.md) | Component scheduling — relevant for future component preemption. |
+| [Scheduler Estimator (521)](../521-scheduler-estimator/) | Estimator gRPC interface, relevant for Phase 2. |
+
+## Detailed Design
+
+### Scope: Single-Cluster Divided Mode
+
+Phase 1 preemption is attempted only when all of the following hold:
+1. The binding is a workload (`IsWorkload() == true`).
+2. The binding is single-component (`len(Components) <= 1`).
```
RainbowMango: Would you mind explaining why we restrict the number of components to 1 here?

I remember your use case that you don't need it to support multiple components, but this might be a limitation in the future. Personally, I don't want the following implementation to be based on this forever. For instance, what if a user runs Pytorch workloads with multiple components?

docs/proposals/scheduling/binding-preemption/binding-preemption.md
```diff
@@ -0,0 +1,995 @@
...
+3. The binding uses `ClusterAffinity` (single affinity), not `ClusterAffinities` (preference-ordered list).
```
RainbowMango: Again, I wonder to know the reason why it can't support `ClusterAffinities`.

docs/proposals/scheduling/binding-preemption/binding-preemption.md
whitewindmills: I think it’s up to user feedback.
I don't know if anyone has already applied it to production.

docs/proposals/scheduling/binding-preemption/binding-preemption.md
```diff
@@ -0,0 +1,995 @@
...
+type SchedulePriority struct {
+    Priority         int32            `json:"priority,omitempty"`
+    PreemptionPolicy PreemptionPolicy `json:"preemptionPolicy,omitempty"`
+}
+
+type PreemptionPolicy string
+
+const (
+    PreemptLowerPriority PreemptionPolicy = "PreemptLowerPriority"
+    PreemptNever         PreemptionPolicy = "Never"
```
RainbowMango: As we talked at the community meeting, we don't need the `Never` here, unset is the default behavior which means `never`.

RainbowMango (commented):
Review is still ongoing.

docs/proposals/scheduling/binding-preemption/binding-preemption.md
RainbowMango: @seanlaii is using it, but I don't know if he uses the same code as the upstream.

seanlaii: We are using this feature in production. But also open to feedback from other users.

docs/proposals/scheduling/binding-preemption/binding-preemption.md
seanlaii: Yes, missed it. Will update the document. Thanks!

docs/proposals/scheduling/binding-preemption/binding-preemption.md
seanlaii: Updated in https://github.com/karmada-io/karmada/pull/7327/changes#:~:text=%23%23%20Why-,ClusterAffinities,-Is%20Excluded.

docs/proposals/scheduling/binding-preemption/binding-preemption.md
seanlaii: The main reason is that I want to focus on replica scheduling in this proposal while maintaining the extendability to support component scheduling. The framework can be extend to support the component scheduling, and I am also happy to work on it, but I would prefer to put it in the other proposal as the current proposal is already a bit complex.
I have also updated the doc with more information in [**Component Scheduling Path**] : https://github.com/karmada-io/karmada/pull/7327/changes#:~:text=%23%23%23-,Component,-Scheduling%20Path.

seanlaii:
@RainbowMango @whitewindmills  Thanks for your time for the review. I have updated the document to make it shorter by removing some unnecessary information and implementation detail.

asarj:
Discussed offline in the [Karmada biweekly english meeting](https://docs.google.com/document/d/1y6YLVC-v7cmVAdbjedoyR5WL0-q45DBRXTvz5_I7bkA/edit?pli=1&tab=t.0#heading=h.b3hcamz4y2pj), will help out with the implementation of this proposal across two phases

asarj:
I attempted to one-shot phase 1's implementation as part of https://github.com/karmada-io/karmada/pull/7832, but I think it is way too big for reviews

If the maintainers agree with this approach, I intend to split the **phase 1** implementation from https://github.com/karmada-io/karmada/pull/7832 into 7 focused diffs in roughly this order:
1. Add the `PriorityBasedPreemptiveScheduling` feature gate
2. Add BindingPreemption API interface
3. Propagate `PriorityClass.preemptionPolicy` into karmada ResourceBindings
4. Add scheduler preemption result and claim store
5. Add Phase 1 victim selection and scheduler trigger
6. Execute preemption through graceful eviction
7. Add an E2E test and update the user-facing docs

asarj:
@whitewindmills thank you for all the review comments so far!

Discussed with @RainbowMango in the karmada english biweekly meeting, can you help give another review of this proposal and let us know if any further adjustments should be made? 

I think @seanlaii addressed all comments so far, but happy to work with the community to help tweak the proposal as needed as our team is very interested in having this feature upstream

RainbowMango:
Yes, please @whitewindmills. Let's see if we can start implementing this in release-1.20.

@asarj Sorry for letting this sit! I will try my best to help.

whitewindmills:
I will conduct a final review before merging. Thank you for your support.

docs/proposals/scheduling/binding-preemption/binding-preemption.md
RainbowMango: By the way. This feature has been promoted to beta in v1.19.0.

docs/proposals/scheduling/binding-preemption/binding-preemption.md
```diff
@@ -0,0 +1,765 @@
...
+metadata:
+  name: training-job
+spec:
+  replicas: 2
+  template:
+    spec:
+      containers:
+      - resources:
+          requests:
+            nvidia.com/gpu: 2
```
RainbowMango, line 81: @asarj, could you please share a full yaml for running a training job? Personally I want to know more about how you run the training job, and also there a chance that the yaml will be used in the E2E test.

docs/proposals/scheduling/binding-preemption/binding-preemption.md
```diff
@@ -0,0 +1,765 @@
...
+
+The preemption infrastructure (claims, victim selection, eviction) is designed to be reusable. The Phase 2 estimator path operates at the pod level, which naturally handles multi-component workloads. Extending preemption to multi-component requires:
+
+1. **A new trigger**: Extend `AssignReplicas` or add a pre-check to call `MaxAvailableComponentSets` (which already exists in `pkg/scheduler/core/estimation.go`) for multi-component workloads, returning an error when the cluster lacks capacity.
+2. **Relax the component check**: Remove `len(Components) > 1` guard in `isPreemptionApplicable`.
+3. **Extend claims**: Track per-component resources in `preemptionClaim.resourceNeed` for accurate claim deductions.
+
+Multi-component workloads can already be **victims** in Phase 2 — the estimator knows their actual pod-level resource consumption. In Phase 1, they are filtered out because their `Spec.Clusters` entries lack `Replicas`.
```
RainbowMango, line 642: This has been changed as we can now get replicas from the `Spec.Clusters` which was introduced in release-1.19. 


Refere to : https://github.com/karmada-io/karmada/blob/master/docs/CHANGELOG/CHANGELOG-1.19.md#multi-component-workload-scheduling-advances-phase-iv

RainbowMango (commented):
Hi @seanlaii, since I know you've moved on to a new role, I'd like to check how you'd prefer to proceed this PR:
- You address the comments yourself when you have time, or
- Let me or @asarj address the comments by a following PR after this PR.

Either way works for me. Just let us know which one is more convenient for you. Thanks a lot for starting this proposal! And sorry for letting this sit for SO LANG!

seanlaii:
> Hi @seanlaii, since I know you've moved on to a new role, I'd like to check how you'd prefer to proceed this PR:
> 
> * You address the comments yourself when you have time, or
> * Let me or @asarj address the comments by a following PR after this PR.
> 
> Either way works for me. Just let us know which one is more convenient for you. Thanks a lot for starting this proposal! And sorry for letting this sit for SO LANG!

Hi @RainbowMango ,

No worries! Thanks for the info! I prefer to let you or @asarj address the comments in a follow-up PR or this one.
Thanks!

docs/proposals/scheduling/binding-preemption/binding-preemption.md
asarj: Our training jobs leverage a internal CRD backed by a separate controller written in KubeFlow, so its original YAML contains environment-specific fields and would not be _directly_ reusable for a Karmada E2E test.

However, from Karmada’s point of view, the relevant shape is a single-replica workload with explicit resource requests. I think the standard Deployment example in this proposal is a better portable equivalent, but I also put together this example using 2 CPU and 2Gi memory, along with its PriorityClass and PropagationPolicy.
```yaml
apiVersion: scheduling.k8s.io/v1
kind: PriorityClass
metadata:
  name: high-priority-training
value: 1000
globalDefault: false
preemptionPolicy: PreemptLowerPriority
description: High-priority training workloads
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: cpu-training-job
  namespace: default
spec:
  replicas: 1
  selector:
    matchLabels:
      app: cpu-training-job
  template:
    metadata:
      labels:
        app: cpu-training-job
    spec:
      containers:
        - name: trainer
          image: busybox:1.36.0
          command: ["/bin/sh", "-c"]
          args:
            - |
              echo "Starting CPU training workload"
              while true; do
                sleep 30
              done
          resources:
            requests:
              cpu: "2"
              memory: 2Gi
            limits:
              cpu: "2"
              memory: 2Gi
---
kind: PropagationPolicy
metadata:
  name: cpu-training-policy
  namespace: default
spec:
  resourceSelectors:
    - apiVersion: apps/v1
      kind: Deployment
      name: cpu-training-job
  schedulePriority:
    priorityClassSource: KubePriorityClass
    priorityClassName: high-priority-training
  placement:
    clusterAffinity:
      clusterNames:
        - member1
    spreadConstraints:
      - spreadByField: cluster
        maxGroups: 1
    replicaScheduling:
      replicaDivisionPreference: Aggregated

```

The E2E test may need to adjust the resource requests based on the test cluster’s capacity, but the scheduling behavior should be the same. 

Would this work for what you had in mind?

docs/proposals/scheduling/binding-preemption/binding-preemption.md
asarj: Nice! Based on https://github.com/karmada-io/karmada/pull/7327#issuecomment-5763349030, I can help update the proposal to leverage this update or alternatively amend to it in a separate PR. Let me know which way is preferable!

docs/proposals/scheduling/binding-preemption/binding-preemption.md
RainbowMango: Thanks, that's very helpful!

> Our training jobs leverage a internal CRD backed by a separate controller written in KubeFlow,

I wonder whether the CRD produces a Deployment, Job, or Pod? Actually, I don't know what kind of workload is well used for training.

docs/proposals/scheduling/binding-preemption/binding-preemption.md
RainbowMango: Sounds great! I will run another round of review and merge this PR then you can help address the comments in a separate PR.

docs/proposals/scheduling/binding-preemption/binding-preemption.md
```diff
@@ -0,0 +1,765 @@
...
+
+```
+ResourceBinding enqueued → Pop (priority-ordered) → findClustersThatFit (Filter)
+→ prioritizeClusters (Score) → selectClusters (SpreadConstraints + capacity)
+→ assignReplicas (Duplicated/Aggregated/StaticWeight/DynamicWeight) → Write result
+```
+
+There is no preemption phase. If scheduling fails, the binding enters `backoffQ` or `unschedulableBindings`.
+
+**Where capacity is checked**: Filter plugins do **not** check resource capacity. Capacity is first evaluated in `selectClusters` via `calAvailableReplicas` (which queries the general-estimator and/or scheduler-estimator for `AllocatableReplicas`), then checked again in `assignReplicas` via `dynamicDivideReplicas`.
```
RainbowMango, line 151: TODO: the `assignReplicas` doesn't have to check resource capacity anymore, we have improved this in release-1.15 by #6597

docs/proposals/scheduling/binding-preemption/binding-preemption.md
```diff
@@ -0,0 +1,765 @@
...
+        ▼
+┌─── Preemption Trigger ───────────────────────────────────────────────┐
+│  Is preemption applicable?                                           │
+│  (workload? single-component? ClusterAffinity? MaxGroups==1?         │
+│   PreemptLowerPriority? feature gates enabled?)                      │
+└───────────────────────┬──────────────────────────────────────────────┘
+                        │ yes
+                        ▼
+┌─── Victim Selection ─────────────────────────────────────────────────┐
+│  1. Find lower-priority bindings on target cluster (via index)       │
```
RainbowMango, line 194: I wonder whether bindings without `SchedulePriority` are eligible victims.  My concern is in a multi-tenant scenario, one tenant enabling this preemption feature will require the other tenant to set the priority for their workloads. 

@asarj @seanlaii I'd like to hear your thoughts.

(By the way, this concern is based on preemption happening across namespaces, will talk about this below)

docs/proposals/scheduling/binding-preemption/binding-preemption.md
```diff
@@ -0,0 +1,765 @@
...
+└──────────────────────────────────────────────────────────────────────┘
+```
+
+**Key concepts**:
+
+- **Preemption trigger**: `assignReplicas` failure (only Aggregated/DynamicWeight can fail).
+- **Victim selection**: "Remove all, then reprieve" greedy heuristic (same as Kubernetes `SelectVictimsOnNode`).
+- **Preemption claims**: In-memory capacity reservation (equivalent to Kubernetes' in-memory pod nominator).
+- **Preemption execution**: Existing GracefulEviction mechanism. Per-cluster, all-or-nothing (consistent with Kubernetes per-pod preemption).
+- **Cross-kind, cross-namespace**: Preemption is purely priority-based with no kind or namespace restrictions, consistent with Kubernetes.
```
RainbowMango, line 225: I'd like to double-check whether cross-namespace preemption is really safe in Karmada, or whether we should restrict it to in-namespace in the alpha phase.

@asarj can you share your thoughts based on your use case?

docs/proposals/scheduling/binding-preemption/binding-preemption.md
```diff
@@ -0,0 +1,765 @@
...
+### Why Estimator-Based Preemption Is Needed
+
+Phase 1 preemption uses aggregate capacity data (derived from `ResourceSummary` via `calAvailableReplicas`) and cannot detect node-level constraints:
+
+- Preemptor needs 8 CPU with node affinity to `gpu-pool` nodes.
+- Victim A: 4 CPU on Node1 (gpu-pool). Victim B: 4 CPU on Node2 (cpu-pool).
+- Summary says "8 CPU freed" → preemption appears feasible.
+- Reality: only 4 CPU freed on gpu-pool nodes → preemptor still cannot schedule.
+
+The estimator has a per-node snapshot and can simulate pod removal to accurately determine feasibility.
```
RainbowMango, line 517: TODO: I'd suggest dropping `Phase 1` totally from this proposal as it can't make correct preemption decisions based on the ResourceSummary.

RainbowMango (commented):
@asarj i've done another round of review and left my comemt, could you please let me know if that make sense to you?

docs/proposals/scheduling/binding-preemption/binding-preemption.md
asarj: At least for our use case, our CRD produces a Pod

I do ,however, think that the Deployment manifest that both the proposal and I shared would allow us to E2E test the related fixtures when implementing the changes without requiring a custom CRD or other training operator

docs/proposals/scheduling/binding-preemption/binding-preemption.md
asarj: Awesome, sounds good. Will update this part once the proposal is merged

docs/proposals/scheduling/binding-preemption/binding-preemption.md
asarj: Got it, we can remove this part in the follow up PR

docs/proposals/scheduling/binding-preemption/binding-preemption.md
asarj: This is a great point. I think if we leverage Karmada's existing default priority of 0, then bindings without `SchedulePriority` should be eligible victims.

The way I think about it is that victim eligibility should be dependent on having lower priority and not the preemption policy.

If there exists a `ResourceBinding` with a nil `SchedulePriority`, then that should not be able to initiate preemption because opting into the `PreemptLowerPriority` preemption policy should be a prerequisite.

I will elaborate on use case in the comment thread for https://github.com/karmada-io/karmada/pull/7327#discussion_r4070030912

docs/proposals/scheduling/binding-preemption/binding-preemption.md
asarj: I elaborated a little bit in https://github.com/karmada-io/karmada/pull/7327#discussion_r4077178219, but to add:

Based on having this feature implemented internally in our Karmada fork, enabling one tenant for the preemption feature has not made other tenants want to do the same. Plus, our team strictly controls access for preemption to supported and critical use cases across business units, and we subsequently define the priority levels for certain workloads based on approved use cases

Having a in-namespace only implementation could work, but it would not serve well in the use case where member-cluster capacity is shared across namespaces (like in our use case)

If you think it's safer, perhaps in the follow up PR to this proposal we can work out what sensible guardrails would look like out of the box

docs/proposals/scheduling/binding-preemption/binding-preemption.md
asarj: Happy to drop Phase 1 here in the follow up PR, but following the theme of https://github.com/karmada-io/karmada/pull/7327#issuecomment-5297096068, would love to know how to split the implementation of this proposal (after follow up edits) in a way that keeps the PRs small and focused for reviewers and maintainer as we target v1.20 release for this

asarj:
> @asarj i've done another round of review and left my comemt, could you please let me know if that make sense to you?

Thanks @RainbowMango for the latest round of feedback! I've replied to all comments and they make sense to me. Once this proposal is approved and merged, I will create a new PR that amends the changes discussed here to make sure we are aligned on the idea before implementation

docs/proposals/scheduling/binding-preemption/binding-preemption.md
RainbowMango: > If you think it's safer, perhaps in the follow up PR to this proposal we can work out what sensible guardrails would look like out of the box

+1
Love to see more details on the preemption policy and how you manage the expectations from multiple tenants. 
It's a significant design trade-off and it deserves to be recorded.

docs/proposals/scheduling/binding-preemption/binding-preemption.md
RainbowMango: Will see along with the following PR.

RainbowMango (approved):
I rebased this PR just for making the CI happy.

As we discussed on above, we are gonna have a or several PRs to address the comments and further udpates. 

I'm going to move this forward now. 
/lgtm
/approve

PR #7340, opened 2026-03-31
FIX: use `%w` instead of `%v` to preserve `UnschedulableError` type chain

seanlaii:
**What type of PR is this?**
/kind bug

- `assignByDynamicStrategy` (assignment.go:220,230,238) wraps errors with `%v` instead of `%w`, which converts the error to a plain string and breaks the `*UnschedulableError` type chain.
- `handleErr` (scheduler.go:842) and `getConditionByError` (helper.go:118) use`errors.As` to detect `UnschedulableError`. Because `%v` discards the type, the check always fails, and bindings that are genuinely unschedulable (insufficient cluster replicas) are misrouted to `backoffQ` instead of `unschedulableBindings`.
- This causes unschedulable bindings to be retried with exponential backoff, wasting scheduler cycles and delaying scheduling of other work.

The fix changes `%v` to `%w` at all three call sites so `errors.As` can unwrap the error chain correctly.

**Error propagation chain**

dynamicDivideReplicas         — creates *UnschedulableError
  → dynamicScaleUp/Down/Fresh — passes through
    → assignByDynamicStrategy — wraps with %v ← BREAKS type chain
      → AssignReplicas         — passes through
        → Schedule             — wraps with %w (correct, but too late)
          → handleErr          — errors.As fails → wrong queue

Fixes #

With this fix, `UnschedulableError` is now correctly routed to `unschedulableBindings` instead of `backoffQ`.

However, the bindings only move out of `unschedulableBindings` in two ways:

- 5-minute timer flush (`flushUnschedulableBindingsLeftover`)
- event-driven requeue when updateCluster detects spec/label/deletion changes via `clusterReconcileWorker` → `enqueueAffectedBindings` → `Push()` → `moveToActiveQ()`.

Neither mechanism reacts to `ClusterStatus` changes. Since `ResourceSummary` and `Conditions` are in `ClusterStatus` (not ClusterSpec), these updates do not increment `Generation` and are silently dropped by the `updateCluster` switch statement. This means a low-priority binding placed in `unschedulableBindings` will wait up to 5 minutes to be retried when cluster resources become available.

I'd like to follow up with [this issue](https://github.com/karmada-io/karmada/issues/7344) to improve the requeue mechanism.

```release-note
`karmada-scheduler`: Fixed incorrect error type propagation that caused bindings with insufficient cluster replicas to be misrouted to `backoffQ` instead of `unschedulableBindings`, changing retry behavior from exponential backoff (1-10s) to timer-based retry (5 minutes).
```

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -629,7 +629,7 @@ func (s *Scheduler) scheduleResourceBindingWithClusterAffinities(rb *workv1alpha
 			firstErr = err
 		}
 
-		err = fmt.Errorf("failed to schedule ResourceBinding(%s/%s) with clusterAffiliates index(%d): %v", rb.Namespace, rb.Name, affinityIndex, err)
+		err = fmt.Errorf("failed to schedule ResourceBinding(%s/%s) with clusterAffiliates index(%d): %w", rb.Namespace, rb.Name, affinityIndex, err)
 		klog.Error(err)
 		s.recordScheduleResultEventForResourceBinding(rb, nil, err)
 		affinityIndex++
@@ -772,7 +772,7 @@ func (s *Scheduler) scheduleClusterResourceBindingWithClusterAffinities(crb *wor
 			firstErr = err
 		}
 
-		err = fmt.Errorf("failed to schedule ClusterResourceBinding(%s) with clusterAffiliates index(%d): %v", crb.Name, affinityIndex, err)
+		err = fmt.Errorf("failed to schedule ClusterResourceBinding(%s) with clusterAffiliates index(%d): %w", crb.Name, affinityIndex, err)
 		klog.Error(err)
 		s.recordScheduleResultEventForClusterResourceBinding(crb, nil, err)
 		affinityIndex++
--- a/pkg/scheduler/core/assignment.go
+++ b/pkg/scheduler/core/assignment.go
@@ -217,7 +217,7 @@ func assignByDynamicStrategy(state *assignState) ([]workv1alpha2.TargetCluster, 
 	if state.assignmentMode == Fresh {
 		result, err := dynamicFreshScale(state)
 		if err != nil {
-			return nil, fmt.Errorf("failed to do fresh scale: %v", err)
+			return nil, fmt.Errorf("failed to do fresh scale: %w", err)
 		}
 		return result, nil
 	}
@@ -227,7 +227,7 @@ func assignByDynamicStrategy(state *assignState) ([]workv1alpha2.TargetCluster, 
 		// We need to reduce the replicas in terms of the previous result.
 		result, err := dynamicScaleDown(state)
 		if err != nil {
-			return nil, fmt.Errorf("failed to scale down: %v", err)
+			return nil, fmt.Errorf("failed to scale down: %w", err)
 		}
 		return result, nil
 	} else if state.assignedReplicas < state.spec.Replicas {
@@ -235,7 +235,7 @@ func assignByDynamicStrategy(state *assignState) ([]workv1alpha2.TargetCluster, 
 		// First scheduling is considered as a special kind of scaling up.
 		result, err := dynamicScaleUp(state)
 		if err != nil {
-			return nil, fmt.Errorf("failed to scale up: %v", err)
+			return nil, fmt.Errorf("failed to scale up: %w", err)
 		}
 		return result, nil
 	}
--- a/pkg/scheduler/core/assignment_test.go
+++ b/pkg/scheduler/core/assignment_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package core
 
 import (
+	"errors"
 	"reflect"
 	"testing"
 
@@ -26,6 +27,7 @@ import (
 	policyv1alpha1 "github.com/karmada-io/karmada/pkg/apis/policy/v1alpha1"
 	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
 	"github.com/karmada-io/karmada/pkg/scheduler/core/spreadconstraint"
+	"github.com/karmada-io/karmada/pkg/scheduler/framework"
 	"github.com/karmada-io/karmada/pkg/util"
 	"github.com/karmada-io/karmada/test/helper"
 )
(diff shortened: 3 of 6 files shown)


seanlaii:
Hi @RainbowMango @zhzhuang-zju , could you help review the PR when you have a chance?
Additionally, if the follow-up makes sense to you, I can also put in a PR to improve it.

XiShanYongYe-Chang:
Thanks~
/assign

pkg/scheduler/scheduler.go
```diff
@@ -629,7 +629,7 @@ func (s *Scheduler) scheduleResourceBindingWithClusterAffinities(rb *workv1alpha
 			firstErr = err
 		}
 
-		err = fmt.Errorf("failed to schedule ResourceBinding(%s/%s) with clusterAffiliates index(%d): %v", rb.Namespace, rb.Name, affinityIndex, err)
+		err = fmt.Errorf("failed to schedule ResourceBinding(%s/%s) with clusterAffiliates index(%d): %w", rb.Namespace, rb.Name, affinityIndex, err)
```
zhzhuang-zju, line 632: Does this need to be changed? It seems that using %v and %w here has no impact.

pkg/scheduler/scheduler.go
```diff
@@ -772,7 +772,7 @@ func (s *Scheduler) scheduleClusterResourceBindingWithClusterAffinities(crb *wor
 			firstErr = err
 		}
 
-		err = fmt.Errorf("failed to schedule ClusterResourceBinding(%s) with clusterAffiliates index(%d): %v", crb.Name, affinityIndex, err)
+		err = fmt.Errorf("failed to schedule ClusterResourceBinding(%s) with clusterAffiliates index(%d): %w", crb.Name, affinityIndex, err)
```
zhzhuang-zju, line 775: ditto

zhzhuang-zju:
> Additionally, if the follow-up makes sense to you, I can also put in a PR to improve it.

Thanks @seanlaii, I think this is a reasonable proposal. It makes sense to me to reschedule Bindings that were previously marked as unschedulable when cluster resource changes are detected (e.g., changes in `ResourceSummary`). However, before proceeding with a PR, perhaps we should first define the specific trigger conditions and the complete workflow.

Asking @whitewindmills,  the code owner of the scheduler queue, to take a look as well.

XiShanYongYe-Chang (commented):
While tracing the error chain, I might have missed a place. Could you help me confirm this:

https://github.com/karmada-io/karmada/blob/da5b1668e9394c965c018f7bccc577df2b6d704f/pkg/scheduler/scheduler.go#L589

pkg/scheduler/scheduler.go
XiShanYongYe-Chang: +1
This error is used for recording and is not transmitted externally, so there is no need to use the %w.

whitewindmills:
>  It makes sense to me to reschedule Bindings that were previously marked as unschedulable when cluster resource changes are detected (e.g., changes in ResourceSummary). However, before proceeding with a PR, perhaps we should first define the specific trigger conditions and the complete workflow.

+1

XiShanYongYe-Chang:
Five minutes is indeed too long. I suggest creating a new issue or PR for the discussion so that the PR can be merged without any impact.

pkg/scheduler/scheduler.go
seanlaii: No, you're right, `%v` vs `%w` has no behavioral impact here. I just thought maybe we can change it to avoid potential issue in the future. But I can also revert it back to keep the PR focus.
WDYT?

seanlaii:
> I have a question: is there a way to find out whether all such issues have been resolved?

Good question! I have added three tests to verify the fix and prevent future regressions:                            

1. `TestHandleErr` (`scheduler_test.go`) — verifies that a wrapped `UnschedulableError` routes to `PushUnschedulableIfNotPresent` (not `PushBackoffIfNotPresent`). This directly tests the queue routing behavior that caused the starvation issue.
2. `Test_assignByDynamicStrategy_UnschedulableErrorPreserved` (`assignment_test.go`) — calls the real `assignByDynamicStrategy` with insufficient replicas and asserts `errors.As(*UnschedulableError)` succeeds on the returned error. If anyone changes `%w` back to `%v`, this test fails.
3. "wrapped unschedulable error" (`helper_test.go`) — adds a case to the existing `Test_getConditionByError` table with a wrapped `UnschedulableError`, verifying the condition reason is `Unschedulable` (not `SchedulerError`).

seanlaii:
>Thanks @seanlaii, I think this is a reasonable proposal. It makes sense to me to reschedule Bindings that were previously marked as unschedulable when cluster resource changes are detected (e.g., changes in ResourceSummary). However, before proceeding with a PR, perhaps we should first define the specific trigger conditions and the complete workflow.

> Five minutes is indeed too long. I suggest creating a new issue or PR for the discussion so that the PR can be merged without any impact.

Sounds good! I have created [this issue](https://github.com/karmada-io/karmada/issues/7344) with more information and update the information in this PR. Please take a look when you have a chance. Thanks!

pkg/scheduler/scheduler.go
zhzhuang-zju: >  But I can also revert it back to keep the PR focus.

Not necessary since there is no practical impact. I just wanted to confirm earlier.

zhzhuang-zju:
> I have added three tests to verify the fix and prevent future regressions:

@seanlaii Nice work! I have no other comments 
/lgtm
Can you fill in `Does this PR introduce a user-facing change?` part, as this is a bug fix? Thanks

XiShanYongYe-Chang:
> While tracing the error chain, I might have missed a place. Could you help me confirm this:
> 
> https://github.com/karmada-io/karmada/blob/da5b1668e9394c965c018f7bccc577df2b6d704f/pkg/scheduler/scheduler.go#L589

Wait a monet, please help confirm this comment.

seanlaii:
> > While tracing the error chain, I might have missed a place. Could you help me confirm this:
> > https://github.com/karmada-io/karmada/blob/da5b1668e9394c965c018f7bccc577df2b6d704f/pkg/scheduler/scheduler.go#L589
> 
> Wait a monet, please help confirm this comment.

Hi @XiShanYongYe-Chang , Sorry, I missed the comment earlier.
This is `klog.Errorf`, not `fmt.Errorf`, so it only prints the error message to the log. The original `err` is returned unchanged, so the error type chain is preserved.
Thanks!

seanlaii:
> /approve
> 
> but /hold waiting for a release note

Hi @RainbowMango , I have updated the release note. Thanks!

zhzhuang-zju:
/hold cancel

@seanlaii, are you willing to backport this fix to the previous branch?

seanlaii:
> /hold cancel
> 
> @seanlaii, are you willing to backport this fix to the previous branch?

Sure, will backport the fix.

zhzhuang-zju:
> Sure, will backport the fix.

Great, I created an issue to track the backport tasks, and assigned it to you

PR #7341, opened 2026-04-01
define top level permission for update-helm-index.yml

zhzhuang-zju:
**What type of PR is this?**
/kind cleanup

Ref to https://github.com/karmada-io/karmada/pull/7247#issuecomment-4161263507, this pr is to:
- define top-level permission for update-helm-index.yml (Verified via https://app.stepsecurity.io/secureworkflow)
- Remove the action `azure/setup-helm`, as this job requires write permissions, and we could minimize the use of actions.

Test report: https://github.com/zhzhuang-zju/karmada/actions/runs/23836850391/job/69482606137

Fixes #

```release-note

```

### DIFF
--- a/.github/workflows/update-helm-index.yml
+++ b/.github/workflows/update-helm-index.yml
@@ -15,8 +15,7 @@ on:
         type: string
 
 permissions:
-  contents: write
-  pull-requests: write
+  contents: read
 
 jobs:
   get-helm-versions:
@@ -48,6 +47,9 @@ jobs:
 
   update-index-pr:
     name: Update Index and Create PR
+    permissions:
+      contents: write
+      pull-requests: write
     needs: get-helm-versions
     if: needs.get-helm-versions.outputs.should_run == 'true'
     runs-on: ubuntu-22.04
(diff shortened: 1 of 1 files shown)


.github/workflows/update-helm-index.yml
```diff
@@ -61,12 +63,10 @@ jobs:
           ref: master
           # fetch-depth:
           # 0 indicates all history for all branches and tags.
-          # for `git describe --tags` in Makefile.
           fetch-depth: 0
       - name: Install Helm
-        uses: azure/setup-helm@dda3372f752e03dde6b3237bc9431cdc2f7a02a2 # v5.0.0
```
RainbowMango, line 67: > Remove the action azure/setup-helm, as this job requires write permissions, and we could minimize the use of actions.

Can you elaborate on why this action requires write permission? It is the action of installing tools, so I don't understand why it requires that. 

In addition, the action is still being used in the other workflows.

zhzhuang-zju: > Can you elaborate on why this action requires write permission? It is the action of installing tools, so I don't understand why it requires that.

The permission is set at the job level. Since the job `Update Index and Create PR` requires write permissions, the GitHub token used by every action within this job has write access. Therefore, it is not that the `azure/setup-helm` action itself needs write permissions; rather, write permissions are available within this job.

> In addition, the action is still being used in the other workflows.

Yes. However, my goal is not to remove the `azure/setup-helm` action, but to eliminate unnecessary actions within jobs that have write permissions.

RainbowMango (approved):
/lgtm
/approve

PR #7343, opened 2026-04-01
Validate critical initData values in operator init workflow

manmathbh:
### What type of PR is this?

/kind bug

### What this PR does / why we need it:

This PR adds validation for critical init data in the operator init workflow before those fields are used.

### Specifically, it now checks:

Spec.Components and Spec.Components.KarmadaAPIServer are present
Spec.HostCluster, Spec.HostCluster.Networking, and Spec.HostCluster.Networking.DNSDomain are present
DNSDomain is not empty
If any of these required fields are missing, the init flow now returns a clear error instead of continuing and failing later with less obvious behavior.

This makes operator initialization safer and easier to debug when the Karmada custom resource is incomplete or misconfigured.

### Special notes for your reviewer:

Main code change is in newRunData validation logic.
Added unit tests that cover invalid configuration cases:
missing components
missing KarmadaAPIServer component
missing host cluster
missing host cluster networking
missing dnsDomain
empty dnsDomain
Test report:
go test ./operator/pkg/...

### Does this PR introduce a user-facing change?:

```
 NONE
```

### DIFF
--- a/operator/pkg/init.go
+++ b/operator/pkg/init.go
@@ -157,7 +157,6 @@ func newRunData(opt *InitOptions) (*initData, error) {
 		return nil, fmt.Errorf("unexpected karmada invalid version %s", opt.KarmadaVersion)
 	}
 
-	// TODO: Verify whether important values of initData is valid
 	var address string
 	if opt.Karmada.Spec.Components.KarmadaAPIServer.ServiceType == corev1.ServiceTypeNodePort {
 		address, err = util.GetAPIServiceIP(remoteClient)


operator/pkg/init.go
```diff
@@ -178,11 +189,40 @@ func newRunData(opt *InitOptions) (*initData, error) {
...
+
+	remoteClient, err := util.BuildClientFromSecretRef(localClient, hostCluster.SecretRef)
+	if err != nil {
+		return nil, fmt.Errorf("error when creating cluster client to install karmada, err: %w", err)
+	}
+
+	return remoteClient, nil
+}
+
+func resolveDNSDomain(hostCluster *operatorv1alpha1.HostCluster) string {
```
zhzhuang-zju: Hi @manmathbh, can you remind me why we use the new function `resolveDNSDomain` to get the dnsDomain?

The operator sets a default value for `karmada.spec.hostcluster.networking.dnsDomain` when reconciling the Karmada CR for the first time. Therefore, we can actually just directly use `*opt.Karmada.Spec.HostCluster.Networking.DNSDomain`.
https://github.com/karmada-io/karmada/blob/c75c55636b97fb3b7c833a8e78bcc1ed04ca2c6e/operator/pkg/apis/operator/v1alpha1/defaults.go#L98-L110

manmathbh: Thanks @zhzhuang-zju  for calling this out. You’re right that in the normal reconcile flow, `dnsDomain` is defaulted, so direct access should work.

I added `resolveDNSDomain` mainly as a defensive safeguard for nil/partial objects (especially in tests or non-standard call paths) and to avoid nil-deref/staticcheck issues. If you’d prefer, I can simplify this and switch back to direct dereference based on the defaulting guarantee.

manmathbh: Hi @zhzhuang-zju, thanks for the reminder, this is a fair point.

You are right: in normal reconcile flow, spec.hostCluster.networking.dnsDomain is defaulted, so direct dereference is sufficient.
I originally added resolveDNSDomain as a defensive nil-safe path for tests/non-standard inputs and to avoid nil-deref warnings, but I agree that it adds unnecessary indirection here.

I will simplify this by removing resolveDNSDomain and using direct dnsDomain access based on the defaulting guarantee. The fail-fast behavior and clearer initData errors remain unchanged.

PTAL after I push the simplification.

zhzhuang-zju: Sorry for the late reply.

> I will simplify this by removing resolveDNSDomain and using direct dnsDomain access based on the defaulting guarantee.

I don't quite understand what's wrong with the current implementation.
Could you first explain the issue, as well as what exactly you mean by the simplification you mentioned?

manmathbh: Hi @zhzhuang-zju, thanks for following up, and no worries.

You’re right that the current implementation is functionally correct. There is no functional bug in `resolveDNSDomain` itself.

What I meant by “simplification” is only a code-clarity refactor:
- keep the same behavior,
- remove the extra helper indirection,
- read `dnsDomain` directly from `opt.Karmada.Spec.HostCluster.Networking.DNSDomain` (which is defaulted in normal reconcile flow).

So the concern is not correctness, only readability/maintainability and aligning with the defaulting contract.

If you prefer, I can keep the current implementation as-is (no further code change) and proceed with this PR in the current form.

zhzhuang-zju: > read dnsDomain directly from opt.Karmada.Spec.HostCluster.Networking.DNSDomain (which is defaulted in normal reconcile flow).

This appears to be how the current implementation works. My understanding is that no changes are needed.

Please correct me if I’m wrong.

zhzhuang-zju:
refer to https://github.com/karmada-io/karmada/pull/7343#discussion_r3079606280

/remove-kind bug
/kind cleanup

manmathbh:
@zhzhuang-zju Thanks for the review.
Since the current implementation looks fine per your last comment, could you /lgtm and /approve when you get a chance?

zhzhuang-zju:
> This appears to be how the current implementation works. My understanding is that no changes are needed.

Sorry for not making myself clear last time. What I mean is that there is no need to modify the code in the master branch. This is because the operator will [set default values](https://github.com/karmada-io/karmada/blob/master/operator/pkg/apis/operator/v1alpha1/defaults.go) and perform [necessary field validation](https://github.com/karmada-io/karmada/blob/master/operator/pkg/controller/karmada/validating.go) for Karmada CRs.

I agree that this approach can improve the robustness of the function. However, I do not recommend adding excessive nil checks in this case. Under a well-defined context, too many redundant nil checks will lead to duplicated code and reduce readability.

Therefore, my suggestion is to only remove the comment:  
`// TODO: Verify whether important values of initData is valid`  

The reason is that the key fields of `initData` have already been validated in `validateKarmada`.

manmathbh:
> > This appears to be how the current implementation works. My understanding is that no changes are needed.
> 
> Sorry for not making myself clear last time. What I mean is that there is no need to modify the code in the master branch. This is because the operator will [set default values](https://github.com/karmada-io/karmada/blob/master/operator/pkg/apis/operator/v1alpha1/defaults.go) and perform [necessary field validation](https://github.com/karmada-io/karmada/blob/master/operator/pkg/controller/karmada/validating.go) for Karmada CRs.
> 
> I agree that this approach can improve the robustness of the function. However, I do not recommend adding excessive nil checks in this case. Under a well-defined context, too many redundant nil checks will lead to duplicated code and reduce readability.
> 
> Therefore, my suggestion is to only remove the comment: `// TODO: Verify whether important values of initData is valid`
> 
> The reason is that the key fields of `initData` have already been validated in `validateKarmada`.

@zhzhuang-zju got it, thanks for clarifying.
I misunderstood earlier, validateKarmada already covers this and adding redundant checks hurts readability.
Let me strip this down to just removing the stale TODO.
Give me a moment.

zhzhuang-zju:
thanks
/lgtm
/approve

Issue #7344, opened 2026-04-01
Scheduler: add event-driven requeue for unschedulable bindings on cluster status changes


seanlaii:
**What would you like to be added**:

Add event-driven requeue for unschedulable bindings when cluster resource availability changes, modeled after the K8s scheduler's [`MoveAllToActiveOrBackoffQueue`](https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/backend/queue/scheduling_queue.go).

Proposed changes:
1. Add `MoveAllToActiveOrBackoffQueue(event string)` to the `SchedulingQueue` interface and `prioritySchedulingQueue`. It moves all bindings from `unschedulableBindings` to `activeQ` (or `backoffQ` if still backing off).
2. Call it from `updateCluster` (`event_handler.go`) when `ClusterStatus.ResourceSummary` or `ClusterStatus.Conditions` changes.
3. Call it from `addCluster` (`event_handler.go`) when a new cluster joins.

**Why is this needed**:

With `PriorityBasedScheduling` enabled, `handleErr` (`scheduler.go:844`) places bindings that fail with `UnschedulableError` (e.g., insufficient cluster replicas) into `unschedulableBindings`.

Today, bindings move out of `unschedulableBindings` in two ways:
1. **Timer-based**: `flushUnschedulableBindingsLeftover` checks every 30s and moves bindings older than 5 minutes back to `activeQ`.
2. **Event-driven** (spec changes only): When `updateCluster` detects a DeletionTimestamp, Labels, or Generation change, it triggers `clusterReconcileWorker` → `enqueueAffectedBindings` → `Push()` → `moveToActiveQ()`, which removes matching bindings from `unschedulableBindings`.

Neither mechanism reacts to `ClusterStatus` changes. `ResourceSummary` and `Conditions` are in `ClusterStatus`, not `ClusterSpec`, so these updates do not increment `Generation` and do not match any case in the `updateCluster` switch statement (`event_handler.go:299-309`).

**Example**: A low-priority binding fails scheduling due to insufficient replicas and enters `unschedulableBindings`. A high-priority job then completes, freeing resources. The cluster status controller updates `ResourceSummary` within ~10s, but `updateCluster` silently drops this status-only update. The low-priority binding waits up to 5 minutes in `unschedulableBindings` before the timer moves it back for retry.

Related: https://github.com/karmada-io/karmada/pull/7340

Chen-speculation:
[...]

PR #7369, opened 2026-04-08
scheduler: requeue unschedulable bindings on cluster status changes

Tej-Katika:
**What this PR does / why we need it**

When `PriorityBasedScheduling` is enabled, bindings that fail with `UnschedulableError` are placed in `unschedulableBindings` and only flushed back to `activeQ` by a 5-minute timer. This happens because `updateCluster` only reacts to `ClusterSpec` changes — `ClusterStatus` fields never increment `Generation` so they were silently dropped.

Two concrete problems this fixes:

1. **ResourceSummary / Conditions** — after a cluster frees capacity or transitions to `Ready`, affected bindings wait up to 5 minutes before retry instead of ~10 seconds.
2. **APIEnablements** — after a CRD is installed on a member cluster, bindings stuck with `"0/N clusters: API resource not found"` are never retried at all (workaround today is to manually patch `spec.rescheduleTriggeredAt`).

**Changes**

- `updateCluster`: adds a new `case` to the existing switch that calls `clusterReconcileWorker.Add(newCluster)` when `ResourceSummary`, `Conditions`, or `APIEnablements` change. This reuses the existing `reconcileCluster` → `enqueueAffectedBindings` → `priorityQueue.Push` → `moveToActiveQ` path, which already handles moving bindings out of `unschedulableBindings`.
- `addCluster`: adds `clusterReconcileWorker.Add(cluster)` so bindings stuck waiting for a new cluster are retried immediately on join rather than after 5 minutes.

No changes to the `SchedulingQueue` interface are needed.

**Which issue(s) this PR fixes**

Part of #7344

**Prerequisites**

#7340 (already merged) — fixes `%w` wrapping so `errors.As(err, &unschedulableErr)` correctly identifies the error, ensuring bindings land in `unschedulableBindings` rather than `backoffQ`.

**Special notes for your reviewer**

The new status-change `case` is intentionally placed after the generation/labels cases in the switch. If spec and status change simultaneously, the generation case fires and the existing path already requeues affected bindings — no double-trigger.

**Does this PR introduce a user-facing change?**

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/event_handler.go
+++ b/pkg/scheduler/event_handler.go
@@ -321,6 +321,9 @@ func (s *Scheduler) addCluster(obj any) {
 	if s.enableSchedulerEstimator {
 		s.schedulerEstimatorWorker.Add(cluster.Name)
 	}
+	if s.priorityQueue != nil {
+		s.priorityQueue.MoveAllToActiveQ()
+	}
 }
 
 func (s *Scheduler) updateCluster(oldObj, newObj any) {
@@ -350,6 +353,25 @@ func (s *Scheduler) updateCluster(oldObj, newObj any) {
 		// to the worker. Therefore, call Add func instead of Enqueue func.
 		s.clusterReconcileWorker.Add(oldCluster)
 		s.clusterReconcileWorker.Add(newCluster)
+	// Cluster status changes that may unblock previously unschedulable bindings:
+	//   - Ready condition transitions: a newly-Ready cluster becomes a scheduling
+	//     candidate; a cluster losing Ready may need its bindings re-evaluated.
+	//   - APIEnablements list changes while CompleteAPIEnablements is True: the
+	//     cluster's supported API set has shifted (e.g. a new CRD installed), so
+	//     bindings whose resource kind was previously unsupported may now fit.
+	//     The True guard avoids reacting to in-progress discovery snapshots.
+	//   - ResourceSummary changes: the cluster's allocatable resource pool moved,
+	//     so bindings that previously exceeded capacity may now fit.
+	// MoveAllToActiveQ flushes the unschedulable map back to the active queue
+	// without a full ResourceBinding scan; each binding is re-evaluated when
+	// popped from the active queue.
+	case util.ClusterConditionStatusChanged(oldCluster, newCluster, clusterv1alpha1.ClusterConditionReady) ||
+		(!equality.Semantic.DeepEqual(oldCluster.Status.APIEnablements, newCluster.Status.APIEnablements) &&
+			meta.IsStatusConditionTrue(newCluster.Status.Conditions, clusterv1alpha1.ClusterConditionCompleteAPIEnablements)) ||
+		!equality.Semantic.DeepEqual(oldCluster.Status.ResourceSummary, newCluster.Status.ResourceSummary):
+		if s.priorityQueue != nil {
+			s.priorityQueue.MoveAllToActiveQ()
+		}
 	}
 }
 
--- a/pkg/scheduler/event_handler_test.go
+++ b/pkg/scheduler/event_handler_test.go
@@ -32,6 +32,7 @@ import (
 	workv1alpha2 "github.com/karmada-io/karmada/pkg/apis/work/v1alpha2"
 	"github.com/karmada-io/karmada/pkg/features"
 	schedulercache "github.com/karmada-io/karmada/pkg/scheduler/cache"
+	internalqueue "github.com/karmada-io/karmada/pkg/scheduler/internal/queue"
 )
 
 func TestResourceBindingEventFilter(t *testing.T) {
@@ -148,55 +149,71 @@ func TestResourceBindingEventFilter(t *testing.T) {
 
 func TestAddCluster(t *testing.T) {
 	tests := []struct {
-		name                     string
-		enableSchedulerEstimator bool
-		obj                      any
-		expectedAdded            bool
-		expectedClusterName      string
+		name                         string
+		enableSchedulerEstimator     bool
+		obj                          any
+		priorityQueue                internalqueue.SchedulingQueue
+		expectedEstimatorAdded       bool
+		expectedEstimatorClusterName string
+		expectedMoveAllToActive      bool
 	}{
 		{
-			name:                     "valid cluster object with estimator enabled",
-			enableSchedulerEstimator: true,
-			obj:                      createCluster("test-cluster", 0, nil),
-			expectedAdded:            true,
-			expectedClusterName:      "test-cluster",
+			name:                         "valid cluster, estimator enabled, no priority queue",
+			enableSchedulerEstimator:     true,
+			obj:                          createCluster("test-cluster", 0, nil),
+			priorityQueue:                nil,
+			expectedEstimatorAdded:       true,
+			expectedEstimatorClusterName: "test-cluster",
+			expectedMoveAllToActive:      false,
 		},
 		{
-			name:                     "valid cluster object with estimator disabled",
+			name:                         "valid cluster, estimator enabled, with priority queue",
+			enableSchedulerEstimator:     true,
+			obj:                          createCluster("test-cluster", 0, nil),
+			priorityQueue:                &mockSchedulingQueue{},
+			expectedEstimatorAdded:       true,
+			expectedEstimatorClusterName: "test-cluster",
+			expectedMoveAllToActive:      true,
+		},
+		{
+			name:                     "valid cluster, estimator disabled, with priority queue",
 			enableSchedulerEstimator: false,
 			obj:                      createCluster("test-cluster-2", 0, nil),
-			expectedAdded:            false,
-			expectedClusterName:      "",
+			priorityQueue:            &mockSchedulingQueue{},
+			expectedEstimatorAdded:   false,
+			expectedMoveAllToActive:  true,
 		},
 		{
 			name:                     "invalid object type",
 			enableSchedulerEstimator: true,
 			obj:                      &corev1.Pod{},
-			expectedAdded:            false,
-			expectedClusterName:      "",
+			expectedEstimatorAdded:   false,
+			expectedMoveAllToActive:  false,
 		},
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			mockWorker := &mockAsyncWorker{}
+			estimatorWorker := &mockAsyncWorker{}
 			s := &Scheduler{
 				enableSchedulerEstimator: tt.enableSchedulerEstimator,
-				schedulerEstimatorWorker: mockWorker,
+				schedulerEstimatorWorker: estimatorWorker,
+				priorityQueue:            tt.priorityQueue,
 			}
 
 			s.addCluster(tt.obj)
 
-			if tt.expectedAdded {
-				assert.Equal(t, 1, mockWorker.addCount, "Worker Add should have been called once")
-				assert.Equal(t, tt.expectedClusterName, mockWorker.lastAdded, "Incorrect cluster name added")
+			if tt.expectedEstimatorAdded {
+				assert.Equal(t, 1, estimatorWorker.addCount, "Estimator worker Add should have been called once")
+				assert.Equal(t, tt.expectedEstimatorClusterName, estimatorWorker.lastAdded, "Incorrect cluster name added to estimator worker")
 			} else {
-				assert.Equal(t, 0, mockWorker.addCount, "Worker Add should not have been called")
-				assert.Nil(t, mockWorker.lastAdded, "No cluster name should have been added")
+				assert.Equal(t, 0, estimatorWorker.addCount, "Estimator worker Add should not have been called")
 			}
 
-			assert.Equal(t, 0, mockWorker.enqueueCount, "Worker Enqueue should not have been called")
-			assert.Nil(t, mockWorker.lastEnqueued, "No item should have been enqueued")
+			if tt.priorityQueue != nil {
+				mock := tt.priorityQueue.(*mockSchedulingQueue)
+				assert.Equal(t, tt.expectedMoveAllToActive, mock.moveAllToActiveCalled, "MoveAllToActive called state mismatch")
+			}
 		})
 	}
 }
@@ -207,8 +224,10 @@ func TestUpdateCluster(t *testing.T) {
 		enableSchedulerEstimator bool
 		oldObj                   any
 		newObj                   any
+		priorityQueue            internalqueue.SchedulingQueue
 		expectedEstimatorAdded   bool
 		expectedReconcileAdded   int
+		expectedMoveAllToActive  bool
 	}{
 		{
 			name:                     "valid cluster update with generation change",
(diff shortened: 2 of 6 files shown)


seanlaii (commented):
Hi @Tej-Katika , thanks for the PR!
/assign

seanlaii (commented):
One thing I'm slightly concerned about is the CPU cost when `ResourceSummary` changes frequently — since it updates every `10s` by default and `enqueueAffectedBindings` does a full scan of all bindings, we could end up with a lot of unnecessary work: listing all bindings, doing affinity matching, and requeueing bindings that will just be popped and short-circuited in `doScheduleBinding` without any actual scheduling needed.                                
                                                               
What do you think about adding a guard so we only trigger reconciliation when there are bindings that could actually benefit from it? Something like:                                                         

```
case !equality.Semantic.DeepEqual(oldCluster.Status.Conditions, newCluster.Status.Conditions) || !equality.Semantic.DeepEqual(oldCluster.Status.APIEnablements, newCluster.Status.APIEnablements):  
    s.clusterReconcileWorker.Add(newCluster)                                                            
case !equality.Semantic.DeepEqual(oldCluster.Status.ResourceSummary, newCluster.Status.ResourceSummary):
    if features.FeatureGate.Enabled(features.PriorityBasedScheduling) && s.priorityQueue.HasUnschedulableBindings() {                                                    
         s.clusterReconcileWorker.Add(newCluster)                                                        
    }     
```                                                    
This way `Conditions/APIEnablements` changes (low frequency, semantically significant) always trigger reconciliation, while `ResourceSummary` changes (high frequency, often just noise) only trigger it when there are unschedulable bindings that might actually benefit from rescheduling.

seanlaii:
Also, A few edge cases around switch-case ordering might be worth adding to guard against future regressions:
For example:                                                                                     

1. Generation + status change simultaneously — generation case should take precedence (expect 2 adds, not 3)                                                                                                  
2. DeletionTimestamp + status change simultaneously — deletion case should take precedence (expect 1 add)                                                                                                    
3. Identical non-nil status — DeepEqual returns true, should not trigger reconcile (expect 0 adds)
                                                                                                          
These would catch accidental reordering of the switch cases in the future.

pkg/scheduler/event_handler_test.go
```diff
@@ -146,55 +146,73 @@ func TestResourceBindingEventFilter(t *testing.T) {
...
+			expectedEstimatorAdded:       false,
+			expectedEstimatorClusterName: "",
+			expectedReconcileAdded:       false,
 		},
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			mockWorker := &mockAsyncWorker{}
+			estimatorWorker := &mockAsyncWorker{}
```
seanlaii, line 184: Could you elaborate the reason of validating estimator here?

seanlaii: It seems that this is not related to this PR, but it is a good enhancement. Maybe we can separate this to a new PR. WDYT?

seanlaii:
cc @RainbowMango @zhzhuang-zju to take a look as well. Thanks!

pkg/scheduler/event_handler_test.go
Tej-Katika: The original TestAddCluster already validated the estimator worker using a single mockWorker. When this PR introduced clusterReconcileWorker.Add(cluster) in addCluster, I needed a separate mock for each worker to independently assert their call counts — otherwise the counts would be conflated across both workers. Keeping the estimator assertions in the same test ensures the refactoring didn't accidentally break the existing estimator path while adding the reconcile path. It tests the full behavior of addCluster in one place rather than leaving the estimator path uncovered.

Tej-Katika: Yeah, that's a fair point. The worker split was structurally necessary to test the new reconcile call independently, but the new estimator else-branch assertions (checking addCount == 0 when estimator is disabled) are purely defensive additions that weren't in the original test. I can strip those back out to keep the diff tightly scoped to this PR. I'll revert the estimator side to match the original assertion style, and only keep what's needed to validate the new clusterReconcileWorker.Add behavior.

Tej-Katika:
ResourceSummary is updated by the cluster-status-controller on every sync cycle (~10s by default), so unconditionally running enqueueAffectedBindings on each update would be expensive in large-scale clusters with many bindings.
The proposed split makes sense. One thing to flag before implementing it: HasUnschedulableBindings() doesn't currently exist on the SchedulingQueue interface or prioritySchedulingQueue. We'd need to add it — something like:
// in SchedulingQueue interface:
 HasUnschedulableBindings() bool

// on prioritySchedulingQueue:
 func (bq *prioritySchedulingQueue) HasUnschedulableBindings() bool {
    bq.lock.RLock()
    defer bq.lock.RUnlock()
    return bq.unschedulableBindings.Len() > 0
  }
Alternatively, since s.priorityQueue is already nil when PriorityBasedScheduling is disabled (it's only initialized inside the feature gate check in New()), the guard could be written as:

case !equality.Semantic.DeepEqual(oldCluster.Status.ResourceSummary, newCluster.Status.ResourceSummary):
    if s.priorityQueue != nil && s.priorityQueue.HasUnschedulableBindings() {
        s.clusterReconcileWorker.Add(newCluster)
    }
This is equivalent to the feature gate check but avoids importing the features package in event_handler.go. Does adding HasUnschedulableBindings() to the SchedulingQueue interface sound reasonable, or would you prefer a different approach?
I'll update the PR with the split cases + the new method once you confirm the direction.

seanlaii:
> ResourceSummary is updated by the cluster-status-controller on every sync cycle (~10s by default), so unconditionally running enqueueAffectedBindings on each update would be expensive in large-scale clusters with many bindings. The proposed split makes sense. One thing to flag before implementing it: HasUnschedulableBindings() doesn't currently exist on the SchedulingQueue interface or prioritySchedulingQueue. We'd need to add it — something like: // in SchedulingQueue interface: HasUnschedulableBindings() bool

Adding `HasUnschedulableBindings()` sounds good to me. 

> if s.priorityQueue != nil

Make sense to me.

Tej-Katika:
cc @seanlaii @RainbowMango @zhzhuang-zju 

Please take a look. Thanks!

pkg/scheduler/event_handler.go
```diff
@@ -277,6 +277,7 @@ func (s *Scheduler) addCluster(obj any) {
 	if s.enableSchedulerEstimator {
 		s.schedulerEstimatorWorker.Add(cluster.Name)
 	}
+	s.clusterReconcileWorker.Add(cluster)
```
whitewindmills: `addCluster` is invoked for existing Cluster objects during informer initial cache population, not only for real new-cluster joins. This unconditional enqueue means a scheduler restart queues one cluster reconciliation per existing cluster; after cache sync each item scans all ResourceBindings/ClusterResourceBindings and may requeue matching bindings even though no cluster changed. In large installations this can turn startup into `O(clusters * bindings)` reconcile work. Consider gating this like the ResourceSummary path, for example only enqueueing when `s.priorityQueue != nil && s.priorityQueue.HasUnschedulableBindings()`, or otherwise distinguishing real post-start joins from initial informer replay.

pkg/scheduler/internal/queue/scheduling_queue.go
```diff
@@ -76,6 +76,9 @@ type SchedulingQueue interface {
 	// Len returns the length of activeQ.
 	Len() int
 
+	// HasUnschedulableBindings reports whether the unschedulableBindings sub-queue is non-empty.
```
whitewindmills: emm, precise, it is not a queue.

pkg/scheduler/event_handler.go
```diff
@@ -306,6 +307,13 @@ func (s *Scheduler) updateCluster(oldObj, newObj any) {
 		// to the worker. Therefore, call Add func instead of Enqueue func.
 		s.clusterReconcileWorker.Add(oldCluster)
 		s.clusterReconcileWorker.Add(newCluster)
+	case !equality.Semantic.DeepEqual(oldCluster.Status.Conditions, newCluster.Status.Conditions) ||
+		!equality.Semantic.DeepEqual(oldCluster.Status.APIEnablements, newCluster.Status.APIEnablements):
+		s.clusterReconcileWorker.Add(newCluster)
+	case !equality.Semantic.DeepEqual(oldCluster.Status.ResourceSummary, newCluster.Status.ResourceSummary):
+		if s.priorityQueue != nil && s.priorityQueue.HasUnschedulableBindings() {
+			s.clusterReconcileWorker.Add(newCluster)
```
whitewindmills: can we focus on unschedulableBindings instead of a full scan of all bindings?

pkg/scheduler/event_handler.go
Tej-Katika: Addressed in `843cd37b` — gated on `s.priorityQueue != nil && s.priorityQueue.HasUnschedulableBindings()` and dropped the unconditional reconcile-worker enqueue in favor of `MoveAllToActive()`.

pkg/scheduler/event_handler.go
Tej-Katika: Both the Conditions/APIEnablements case and the ResourceSummary case now call `MoveAllToActive()` directly, so
we no longer fan out into a full binding scan via `clusterReconcileWorker.Add` → `enqueueAffectedBindings`.

pkg/scheduler/internal/queue/scheduling_queue.go
```diff
@@ -76,6 +76,12 @@ type SchedulingQueue interface {
 	// Len returns the length of activeQ.
 	Len() int
 
+	// HasUnschedulableBindings reports whether the unschedulableBindings map is non-empty.
+	HasUnschedulableBindings() bool
```
whitewindmills: `MoveAllToActive` can be called safely, does we still need this?

Tej-Katika: Agreed. Since `MoveAllToActive()` is already safe to call when the unschedulable map is empty, the
`HasUnschedulableBindings()` guard is redundant. I'll drop the interface method and its implementation, and simplify the three call sites in `event_handler.go` to just `if s.priorityQueue != nil { s.priorityQueue.MoveAllToActive()}.`

Tej-Katika: Dropped `HasUnschedulableBindings()` and simplified the three call sites to `if s.priorityQueue != nil { s.priorityQueue.MoveAllToActive() }`. Updated the docstring on `MoveAllToActive` to note it's safe to call when the unschedulable map is empty.

Tej-Katika:
cc @seanlaii @RainbowMango @whitewindmills 

Please take a look. Thanks!

pkg/scheduler/event_handler.go
```diff
@@ -306,6 +309,15 @@ func (s *Scheduler) updateCluster(oldObj, newObj any) {
...
 		s.clusterReconcileWorker.Add(newCluster)
+	case !equality.Semantic.DeepEqual(oldCluster.Status.Conditions, newCluster.Status.Conditions) ||
+		!equality.Semantic.DeepEqual(oldCluster.Status.APIEnablements, newCluster.Status.APIEnablements):
+		if s.priorityQueue != nil {
+			s.priorityQueue.MoveAllToActive()
+		}
+	case !equality.Semantic.DeepEqual(oldCluster.Status.ResourceSummary, newCluster.Status.ResourceSummary):
+		if s.priorityQueue != nil {
+			s.priorityQueue.MoveAllToActive()
+		}
```
whitewindmills, line 374: we can merge those two cases.

pkg/scheduler/event_handler.go
```diff
@@ -306,6 +309,15 @@ func (s *Scheduler) updateCluster(oldObj, newObj any) {
 		// to the worker. Therefore, call Add func instead of Enqueue func.
 		s.clusterReconcileWorker.Add(oldCluster)
 		s.clusterReconcileWorker.Add(newCluster)
+	case !equality.Semantic.DeepEqual(oldCluster.Status.Conditions, newCluster.Status.Conditions) ||
```
whitewindmills: not all conditions need to be considered, I think we just need a `Ready` condition.

pkg/scheduler/event_handler.go
```diff
@@ -306,6 +309,15 @@ func (s *Scheduler) updateCluster(oldObj, newObj any) {
...
+		!equality.Semantic.DeepEqual(oldCluster.Status.APIEnablements, newCluster.Status.APIEnablements):
```
whitewindmills: For `APIEnablements`, we introduced a condition called `CompleteAPIEnablements` that indicates whether a cluster's `APIEnablements` are complete. Here, I think it will be a very important reference point.
https://github.com/karmada-io/karmada/blob/3fdff8de69b924d63428e374be69b6579311d52b/pkg/controllers/status/cluster_status_controller.go#L247-L260

pkg/scheduler/internal/queue/scheduling_queue.go
```diff
@@ -76,6 +76,10 @@ type SchedulingQueue interface {
 	// Len returns the length of activeQ.
 	Len() int
 
+	// MoveAllToActive moves all bindings from unschedulableBindings directly to activeQ.
+	// Safe to call when unschedulableBindings is empty; it is a no-op in that case.
```
whitewindmills:
```suggestion
	// Safe to call when unschedulableBindings are empty; it is a no-op in that case.
```

pkg/scheduler/internal/queue/scheduling_queue.go
```diff
@@ -76,6 +76,10 @@ type SchedulingQueue interface {
...
+	MoveAllToActive()
```
whitewindmills:
```suggestion
	MoveAllToActiveQ()
```

Tej-Katika:
@whitewindmills Corrected the errors and added the suggested fixes. Please take a look.

Thanks!

pkg/scheduler/event_handler.go
```diff
@@ -306,6 +323,12 @@ func (s *Scheduler) updateCluster(oldObj, newObj any) {
 		// to the worker. Therefore, call Add func instead of Enqueue func.
 		s.clusterReconcileWorker.Add(oldCluster)
 		s.clusterReconcileWorker.Add(newCluster)
+	case clusterConditionStatusChanged(oldCluster, newCluster, clusterv1alpha1.ClusterConditionReady) ||
+		clusterConditionStatusChanged(oldCluster, newCluster, clusterv1alpha1.ClusterConditionCompleteAPIEnablements) ||
```
whitewindmills: I think we cannot ignore this.
```suggestion
		（!equality.Semantic.DeepEqual(oldCluster.Status.APIEnablements, newCluster.Status.APIEnablements) && meta.IsStatusConditionTrue(newCluster.Status.Conditions, clusterv1alpha1.ClusterConditionCompleteAPIEnablements)) ||
```

pkg/scheduler/event_handler.go
```diff
@@ -277,6 +277,23 @@ func (s *Scheduler) addCluster(obj any) {
...
 	}
+	if s.priorityQueue != nil {
+		s.priorityQueue.MoveAllToActiveQ()
+	}
+}
+
+// clusterConditionStatusChanged reports whether the named condition's Status
+// differs between old and new. A missing condition is treated as having an
+// empty status, so first-time appearance also counts as a change.
+func clusterConditionStatusChanged(oldCluster, newCluster *clusterv1alpha1.Cluster, conditionType string) bool {
```
whitewindmills: These types of utility functions are recommended to be archived in the util directory.

Tej-Katika:
cc @whitewindmills

pkg/scheduler/event_handler.go
```diff
@@ -306,6 +309,13 @@ func (s *Scheduler) updateCluster(oldObj, newObj any) {
 		// to the worker. Therefore, call Add func instead of Enqueue func.
 		s.clusterReconcileWorker.Add(oldCluster)
 		s.clusterReconcileWorker.Add(newCluster)
+	case util.ClusterConditionStatusChanged(oldCluster, newCluster, clusterv1alpha1.ClusterConditionReady) ||
```
whitewindmills, line 368: Could you add some notes explaining why we're doing this?

whitewindmills:
LGTM
ping @seanlaii @RainbowMango

pkg/scheduler/event_handler.go
Tej-Katika: Added a comment block above this case walking through what each sub-condition catches (`Ready` transition, `APIEnablements` growth under `CompleteAPIEnablements`, `ResourceSummary` churn) and why `MoveAllToActiveQ` replaced the earlier full-scan path.

Tej-Katika:
cc @RainbowMango @seanlaii Please take a look. Thanks!

seanlaii (commented):
LGTM, could you help resolve the merge conflict? Thanks!

pkg/scheduler/event_handler.go
```diff
@@ -306,6 +309,25 @@ func (s *Scheduler) updateCluster(oldObj, newObj any) {
...
+	//     The True guard avoids reacting to in-progress discovery snapshots.
+	//   - ResourceSummary changes: the cluster's allocatable resource pool moved,
+	//     so bindings that previously exceeded capacity may now fit.
+	// MoveAllToActiveQ flushes the unschedulable map back to the active queue
+	// without a full ResourceBinding scan; each binding is re-evaluated when
+	// popped from the active queue.
+	case util.ClusterConditionStatusChanged(oldCluster, newCluster, clusterv1alpha1.ClusterConditionReady) ||
+		(!equality.Semantic.DeepEqual(oldCluster.Status.APIEnablements, newCluster.Status.APIEnablements) &&
+			meta.IsStatusConditionTrue(newCluster.Status.Conditions, clusterv1alpha1.ClusterConditionCompleteAPIEnablements)) ||
+		!equality.Semantic.DeepEqual(oldCluster.Status.ResourceSummary, newCluster.Status.ResourceSummary):
```
seanlaii, line 371: Would it make sense to only trigger `MoveAllToActiveQ` when allocatable resources increase?

Tej-Katika: The high-frequency churn in `ResourceSummary` is `Allocating/Allocated` (they move every scheduling cycle); `Allocatable` only shifts when nodes join/leave/resize. So the simple win is narrowing the trigger to `"Allocatable/AllocatableModelings` changed" — kills the per-cycle noise with a trivial `DeepEqual` on those sub-fields.

A strict increase-only check needs a per-resource-name map comparison + new-key handling + the resource-model grade path — and only additionally suppresses the rare decrease (node drained), which can't unblock bindings anyway. Since `MoveAllToActiveQ()` is a no-op on an empty unschedulable map and bindings are re-evaluated lazily on pop, an over-eager flush is cheap.

Tej-Katika:
> LGTM, could you help resolve the merge conflict? Thanks!

Updated!

whitewindmills:
leave LGTM to @seanlaii 
/approve
/kind feature

Tej-Katika:
cc @seanlaii — approved, just needs your LGTM. Fine to land as-is; happy to narrow the trigger in a follow-up if you'd prefer.

seanlaii (commented):
LGTM
Tag @RainbowMango for final pass. Thanks!

RainbowMango:
Thanks for the work here — the problem is real and clearly described. However, I'd like to share a concern about the overall direction of this approach.

Even today, having the scheduler react to cluster update events to reschedule associated ResourceBindings already makes me a little uneasy. This level of automation isn't necessarily what users want — rescheduling can be disruptive, and it happening implicitly in response to cluster-side changes takes control away from the user.

This PR pushes further in that direction: by making the scheduler sense cluster status changes (in particular ResourceSummary), it effectively turns rescheduling into something that can fire roughly every 10 seconds, since ResourceSummary is updated on every cluster status heartbeat and its values change continuously. In practice, this largely defeats the purpose of the 5-minute hold in unschedulableBindings and introduces a lot of churn.

Taking a longer-term view, I don't think the scheduler should be sensing cluster changes at all. Rescheduling of a ResourceBinding should be triggered explicitly by the user, or only under rules that the user has explicitly opted into — not inferred automatically from cluster state.

RainbowMango:
So I'd like us to step back and reconsider the underlying problem. 

Today, if a ResourceBinding fails to schedule due to insufficient resources, the worst case is that it waits up to 5 minutes before being retried. My question is: Is that actually an unacceptable outcome? 

- If we agree it isn't, then we may not need this cluster-status-driven path at all.
- And if 5 minutes is considered too long, a simpler and more self-contained solution might be to introduce a backoff mechanism with a longer interval, rather than coupling rescheduling to cluster status changes.

/hold 
for a while, for discussing the solution.

Tej-Katika:
Thanks @RainbowMango, yeah, you're right about `ResourceSummary`. The whole struct gets `DeepEqual'`d, and `Allocating/Allocated` shift basically every cycle, so that arm fires on nearly every heartbeat. No argument from me there.

Also worth saying upfront, since it changes how big a deal this is: the path only kicks in when `PriorityBasedScheduling` is on (it's behind the `s.priorityQueue != nil` check), and in that mode the 5-min `flushUnschedulableBindingsLeftover` timer already catches everything eventually. So nothing's actually stuck forever here. It's really just saving the wait from 5 min to a few seconds.

The bit I'd push back on is lumping it all together. The `ResourceSummary` (capacity) case is the noisy one, and I'm with you on it — 5 min is probably fine, and if it's not, backoff is the cleaner fix. But the other two feel different to me. When `Ready` flips to true, or `APIEnablements` finishes because a CRD finally got installed, the binding didn't "not fit" — it had nowhere to go at all, and now it does. Those are rare, one-off events, not churn. Having a freshly-Ready cluster sit idle for 5 min while bindings wait on it is the part that bugs me. And backoff doesn't help that case, it makes it worse — a longer interval just delays the retry that would've worked the second the cluster came up.

On the disruption side: this doesn't actually decide to reschedule anything. A binding is only in `unschedulableBindings` because the scheduler already tried to place it and failed. `MoveAllToActiveQ` just retries those sooner than the timer would, and it can't drag a healthy, already-placed binding into it. So it's changing when a retry that's already pending fires, not whether a workload moves.

So what I'm thinking of suggesting is to drop the `ResourceSummary` arm, keep just the two viability transitions, and spin the capacity case off into a separate backoff discussion. And if you'd rather the scheduler not react to even those long-term, I'm fine hiding it behind a feature gate, off by default.

Does that narrower scope sit ok with you, or do you see the viability ones as out of bounds for the scheduler too?

Tej-Katika:
@RainbowMango Thoughts?

RainbowMango:
I'd like to hear @seanlaii 's thoughts.

Tej-Katika:
@seanlaii Can you please provide your thoughts on this?

Tej-Katika:
@seanlaii Can you please provide your thoughts on this?

PR #7378, opened 2026-04-09
Add unit tests for scheduler util functions and ClusterAccessCredentialChanged

archy-rock3t-cloud:
**What type of PR is this?**

/kind cleanup

Add unit tests for scheduler util functions and ClusterAccessCredentialChanged.

```release-note
NONE
```


archy-rock3t-cloud:
Thanks for the review — removed the redundant length guard, `reflect.DeepEqual` handles the empty-slice case correctly since both `got` and `want` are non-nil empty slices from `make()`. /assign @xishanyongye-chang

PR #7392, opened 2026-04-15
test(e2e): add operator initData validation event test

### LINKED ISSUE #7252: [Test] Improve E2E test coverage for Karmada controller events

manmathbh:
/kind feature

This PR introduces an End-to-End (E2E) test for the Karmada operator to verify validation events. Specifically, it asserts that the operator correctly emits a ValidationError event when initialized with invalid initData up front.
Note: Following review feedback, the test was refactored to validate failure at creation time with proper nil-pointer safeguards, rather than mutating the object post-readiness.

**Part of #7252**

*Opening this initially as a Draft PR to leverage the CI pipeline's `kind` cluster for E2E validation.*

```release-note
NONE

### DIFF
--- a/test/e2e/suites/operator/crdtarball_test.go
+++ b/test/e2e/suites/operator/crdtarball_test.go
@@ -1,0 +1,112 @@
+/*
+Copyright 2026 The Karmada Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package e2e
+
+import (
+	"context"
+	"strings"
+	"time"
+
+	"github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/rand"
+
+	operatorv1alpha1 "github.com/karmada-io/karmada/operator/pkg/apis/operator/v1alpha1"
+	karmadacontroller "github.com/karmada-io/karmada/operator/pkg/controller/karmada"
+	operatorresource "github.com/karmada-io/karmada/test/e2e/framework/resource/operator"
+	"github.com/karmada-io/karmada/test/helper"
+)
+
+var _ = ginkgo.Describe("CRDTarball configuration testing", func() {
+	var karmadaName string
+
+	ginkgo.BeforeEach(func() {
+		karmadaName = KarmadaInstanceNamePrefix + rand.String(RandomStrLength)
+	})
+
+	ginkgo.AfterEach(func() {
+		operatorresource.DeleteKarmadaInstance(operatorClient, testNamespace, karmadaName)
+	})
+
+	ginkgo.Context("CRDTarball validation", func() {
+		ginkgo.It("should update Ready condition and emit Warning event when HTTP URL is invalid", func() {
+			ginkgo.By("create a Karmada instance with an invalid CRD tarball HTTP URL", func() {
+				karmada := helper.NewKarmada(testNamespace, karmadaName)
+				karmada.Spec.CRDTarball.HTTPSource.URL = "invalid-url"
+
+				err := operatorresource.CreateKarmadaInstance(operatorClient, karmada)
+				gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+			})
+
+			ginkgo.By("assert Ready condition becomes False with ValidationError reason", func() {
+				gomega.Eventually(func(g gomega.Gomega) {
+					updated, err := operatorClient.OperatorV1alpha1().Karmadas(testNamespace).Get(context.TODO(), karmadaName, metav1.GetOptions{})
+					g.Expect(err).ShouldNot(gomega.HaveOccurred())
+
+					var readyCond *metav1.Condition
+					for i := range updated.Status.Conditions {
+						if updated.Status.Conditions[i].Type == string(operatorv1alpha1.Ready) {
+							readyCond = &updated.Status.Conditions[i]
+							break
+						}
+					}
+
+					g.Expect(readyCond).ShouldNot(gomega.BeNil())
+					g.Expect(readyCond.Status).Should(gomega.Equal(metav1.ConditionFalse))
+					g.Expect(readyCond.Reason).Should(gomega.Equal(karmadacontroller.ValidationErrorReason))
+					g.Expect(readyCond.Message).Should(gomega.ContainSubstring(karmadacontroller.ErrInvalidCRDsRemoteURL.Error()))
+				}, pollTimeout, pollInterval).Should(gomega.Succeed())
+			})
+
+			ginkgo.By("assert a Warning ValidationError event is emitted", func() {
+				gomega.Eventually(func(g gomega.Gomega) bool {
+					events, err := hostClient.CoreV1().Events(testNamespace).List(context.TODO(), metav1.ListOptions{})
+					g.Expect(err).ShouldNot(gomega.HaveOccurred())
+
+					for _, event := range events.Items {
+						if event.InvolvedObject.Name == karmadaName &&
+							event.Type == corev1.EventTypeWarning &&
+							event.Reason == karmadacontroller.ValidationErrorReason &&
+							strings.Contains(event.Message, karmadacontroller.ErrInvalidCRDsRemoteURL.Error()) {
+							return true
+						}
+					}
+
+					return false
+				}, pollTimeout, pollInterval).Should(gomega.BeTrue())
+			})
+		})
+	})
+
+	ginkgo.Context("CRDTarball functionality", func() {
+		ginkgo.It("should deploy successfully when CRDTarball is not set (operator applies defaults)", func() {
+			ginkgo.By("create a Karmada instance with nil CRDTarball", func() {
+				karmada := helper.NewKarmada(testNamespace, karmadaName)
+				karmada.Spec.CRDTarball = nil
+
+				err := operatorresource.CreateKarmadaInstance(operatorClient, karmada)
+				gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+			})
+
+			ginkgo.By("assert Karmada becomes Ready with defaulted CRDTarball", func() {
+				operatorresource.WaitKarmadaReady(operatorClient, testNamespace, karmadaName, time.Now())
+			})
+		})
+	})
+})
--- a/operator/pkg/controller/karmada/validating.go
+++ b/operator/pkg/controller/karmada/validating.go
@@ -18,6 +18,7 @@ package karmada
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/url"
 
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by mohamedawnallah** [comment]
Thanks @manmathbh for the PR! Whenever you have it in a reviewable state, just ping me and I'll take a look as soon as I'm available

**Comment by manmathbh** [comment]
Hi @mohamedawnallah  thanks for checking in!

I have refactored the test based on the initial bot feedback:

- Added proper nil-pointer safeguards.
- Restructured the test to assert invalid initData during the initial object creation, rather than doing a post-readiness mutation (which better aligns with the PR's intent).

The CI suite is completely green. This is now ready for your review. PTAL!

File: hack/deploy-karmada-operator.sh [hunk 1]
```diff
@@ -70,4 +70,4 @@ kubectl --kubeconfig="${KUBECONFIG}" --context="${CONTEXT_NAME}" apply -f "${REP
 kubectl --kubeconfig="${KUBECONFIG}" --context="${CONTEXT_NAME}" apply -f "${REPO_ROOT}/operator/config/deploy/karmada-operator-deployment.yaml"
 
 # wait karmada-operator ready
-kubectl --kubeconfig="${KUBECONFIG}" --context="${CONTEXT_NAME}" wait --for=condition=Ready --timeout=30s pods -l app.kubernetes.io/name=karmada-operator -n ${KARMADA_SYSTEM_NAMESPACE}
+kubectl --kubeconfig="${KUBECONFIG}" --context="${CONTEXT_NAME}" rollout status deployment/karmada-operator --timeout=90s -n ${KARMADA_SYSTEM_NAMESPACE}
```
mohamedawnallah, line 73: Do we need to make those changes?

test/e2e/suites/operator/validating_test.go
```diff
@@ -0,0 +1,95 @@
...
+	"k8s.io/apimachinery/pkg/util/rand"
+
+	operatorv1alpha1 "github.com/karmada-io/karmada/operator/pkg/apis/operator/v1alpha1"
+	karmadacontroller "github.com/karmada-io/karmada/operator/pkg/controller/karmada"
+	operatorresource "github.com/karmada-io/karmada/test/e2e/framework/resource/operator"
+	"github.com/karmada-io/karmada/test/helper"
+)
+
+var _ = ginkgo.Describe("Karmada validation event testing", func() {
+	const validationErrorMessageSubstring = "invalid CRDs remote URL"
```
mohamedawnallah, line 36: Let's have a sentinel error perhaps with a concise name in https://github.com/karmada-io/karmada/blob/2ae3c07b397897fca831567c0523ec75f0f7c9bb/operator/pkg/controller/karmada/validating.go#L34 and we can reuse it here

test/e2e/suites/operator/validating_test.go
```diff
@@ -0,0 +1,95 @@
...
+	})
+
+	ginkgo.AfterEach(func() {
+		operatorresource.DeleteKarmadaInstance(operatorClient, testNamespace, karmadaName)
+	})
+
+	ginkgo.It("should emit warning event and set Ready condition false when CRD tarball HTTP URL is invalid", func() {
+		ginkgo.By("Step 1: create the Karmada instance with an invalid CRD tarball URL", func() {
+			karmada := helper.NewKarmada(testNamespace, karmadaName)
+			karmada.Spec.CRDTarball.HTTPSource.URL = "bad-url"
```
mohamedawnallah, line 51: nit:
```suggestion
			karmada.Spec.CRDTarball.HTTPSource.URL = "invalid-url"
```

mohamedawnallah (commented):
Thanks @manmathbh for the PR! Left some comments, otherwise LGTM

manmathbh:
Made changes based on review @mohamedawnallah. Please take a look ~

moawnallah (approved):
Thanks @manmathbh, LGTM!

XiShanYongYe-Chang:
Hi @manmathbh this formation `Fixes #xxx` will close the related issue, can you help update it `Part of #xxx`

test/e2e/suites/operator/validating_test.go
```diff
@@ -0,0 +1,89 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/rand"
+
+	operatorv1alpha1 "github.com/karmada-io/karmada/operator/pkg/apis/operator/v1alpha1"
+	karmadacontroller "github.com/karmada-io/karmada/operator/pkg/controller/karmada"
+	"github.com/karmada-io/karmada/test/e2e/framework"
+	operatorresource "github.com/karmada-io/karmada/test/e2e/framework/resource/operator"
+)
+
+var _ = ginkgo.Describe("Karmada validation event testing", func() {
```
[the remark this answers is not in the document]
XiShanYongYe-Chang: +1

XiShanYongYe-Chang (commented):
To be honest, I'm a bit hesitant right now about whether we need to create a separate test file specifically for testing events. I feel like it might not be very valuable.

manmathbh:
@XiShanYongYe-Chang honestly, fair point, I won't pretend a standalone event file is the cleanest design. My real goal here was to make sure the operator's error path has some coverage: if someone feeds it bad init data, does it actually surface the error in a way operators can see? The event + Ready=False check felt like the right way to confirm that.
All that said, I'm not precious about where the code lives. If folding it into an existing test makes more sense to you, just say the word and I'll move it. Thanks for pushing me to think about structure, not just coverage.

XiShanYongYe-Chang:
If the test is for operator verification, the test case should be designed completely. The test event becomes an additional test purpose. In this case, the task is not a part of #7252. 

If you plan to move forward with this pr, I recommend improving in this direction to provide a complete test case design.

manmathbh:
@XiShanYongYe-Chang, makes total sense. Wrong framing on my part, this shouldn't be under #7252 at all, it's operator verification.
Let me redesign, instead of a single "check if event fires" test, I'll build a proper validation suite that covers multiple invalid-init scenarios and for each one asserts both the Ready=False condition AND the surfaced error. The event assertion stays as a natural signal, not the headline.
I'll drop the #7252 reference and come back with a focused design. Sound right?

XiShanYongYe-Chang:
Let's ask the operator to help take a look~
cc @zhzhuang-zju

zhzhuang-zju:
First of all, thanks @manmathbh for the contribution.

In e2e testing, we generally focus on validating a specific feature or scenario end to end. I'd advise against integrating validations for different scenarios into a single test suite, as it makes the tests less cohesive.
So, if we're just adding a validation suite, I wouldn't recommend it, since UT is a much better fit for that purpose. However, if we are validating a feature—for example, `CRDTarball`—I think it's ok to include checks for invalid CRDTarball inputs as part of the process. E2E tests should always remain centered around a specific scenario or feature.

manmathbh:
@XiShanYongYe-Chang , @zhzhuang-zju the v1.35.0 e2e failure is a flake in migration_and_rollback_test.go:124, completely unrelated to CRDTarball. Operator tests are all green. Could someone re-trigger the failed job?

XiShanYongYe-Chang:
> /retest

I understand that this command is unrestricted

/retest

zhzhuang-zju:
/ok-to-test

@manmathbh Currently, the e2e test essentially only validates the basic CRDTarball HTTP URL. I don't think this is sufficient to qualify as a proper e2e test.

If you'd like this e2e test to be accepted, please consider adding some functional test scenarios, such as:
- If CRDTarball is not set, it should download the CRD files using the default value.
- If an incorrect CRDTarball is set, it should fail.
- If a correct CRDTarball is set, it should download the CRD files using the configured HTTP URL.

manmathbh:
@zhzhuang-zju, appreciate the clear direction  that was exactly what this needed.
Restructured into two focused contexts covering all three scenarios you laid out:

- validation: invalid URL  Ready=False + Warning event (what was already there, cleaned up)
- functionality: nil CRDTarball (defaults kick in) and valid URL (normal path), both asserting the operator reaches Ready

Squashed to one commit.
PTAL whenever you get a moment.

test/e2e/suites/operator/crdtarball_test.go
```diff
@@ -0,0 +1,116 @@
...
+
+			ginkgo.By("assert Karmada becomes Ready with defaulted CRDTarball", func() {
+				operatorresource.WaitKarmadaReady(operatorClient, testNamespace, karmadaName, time.Now())
+			})
+		})
+
+		ginkgo.It("should deploy successfully when CRDTarball HTTP URL is valid", func() {
+			InitializeKarmadaInstance(operatorClient, testNamespace, karmadaName)
+		})
+	})
```
zhzhuang-zju:
```suggestion
```

This part should be removed. 
I know this test is intended to verify that karmada instance will be deployed successfully when the CRDTarball HTTP URL is valid. However, `InitializeKarmadaInstance` uses a tricky workaround to skip downloading the CRD tarball. Therefore, this test cannot actually validate the scenario where the CRDTarball HTTP URL is legitimate.

zhzhuang-zju:
@manmathbh Now this E2E test has sufficient coverage.

manmathbh:
@zhzhuang-zju done, removed that one.
CI's all green now.

Issue #7062, opened 2025-12-30
Optimize ResourceBinding to Work synchronization throughput for large-scale Pod distribution scenarios (10000+ Pods).


Kevinz857:
**What would you like to be added**:

  Optimize ResourceBinding to Work synchronization throughput for large-scale Pod distribution scenarios (10000+ Pods).

  **Why is this needed**:

  ### Problem Description

  In large-scale resource distribution scenarios (e.g., distributing 10000+ Pods to multiple member clusters), we observed a significant bottleneck in the ResourceBinding to Work synchronization path.

  **Observed data during stress testing:**
  etcd object counts:
  - Pod:             12649
  - ResourceBinding: 12645  ✅ Almost caught up with Pods
  - Work:            6646   ❌ ~6000 lagging behind ResourceBindings

  ### Bottleneck Analysis

  We analyzed the complete data path and identified where the latency occurs:

  | Stage | Component | Latency | Status |
  |-------|-----------|---------|--------|
  | Pod → ResourceBinding | resource-detector | ~ms | ✅ Normal |
  | ResourceBinding → RB.spec.clusters | karmada-scheduler | ~ms | ✅ Normal |
  | **ResourceBinding → Work** | **binding-controller** | **seconds~minutes** | ⚠️ **Bottleneck** |

  ### Root Cause Analysis

  The binding-controller has several performance issues in the RB → Work path:

  1. **Synchronous Work Creation**
     - Each ResourceBinding reconcile blocks on Work creation
     - No decoupling between scheduling decision and persistence
     - Redundant reconciles when works are being created

  2. **Inefficient API Call Pattern**
     - Uses `controllerutil.CreateOrUpdate` which always does Get before Create
     - For new Work objects: 2 API calls (Get + Create) instead of 1
     - No fast-path to skip unchanged Work updates

  3. **Unnecessary Orphan Work Checks**
     - `removeOrphanWorks()` is called on **every reconcile**
     - Each check triggers a List API call via `GetWorksByBindingID()`
     - With 12645 RBs, this means 12645+ unnecessary List operations

  4. **Sequential Multi-Cluster Work Creation**
     - When distributing to N clusters, Work objects are created sequentially
     - For dual-cluster scenario: 2 sequential API calls instead of parallel

  5. **Excessive Event Recording**
     - Records 2 Events per successful sync (binding + workload)
     - In high-throughput scenarios, this creates significant API load

  ### API Call Analysis (Dual-Cluster Distribution)

[...]

PR #7416, opened 2026-04-18
[Proposal] scheduler estimator reservation

### LINKED ISSUE #6783: Stale Resource Assessment in karmada-scheduler-estimator Causes Incorrect Placement During Concurrent Aggregated Scheduling
**What happened**:
When two Deployments (Deploy1 and Deploy2), both using the Aggregated scheduling strategy, are created almost simultaneously and target the same cluster (Cluster A) which only has sufficient resources to run one of them, the Karmada control plane incorrectly schedules all replicas of both Deployments to Cluster A. As a result, one of the Deployments (or some of its Pods) fails to reach the Running state due to insufficient resources in Cluster A.

The root cause lies in the Karmada-scheduler-estimator's resource assessment mechanism. When processing rapidly consecutive scheduling requests, there exists a brief time window during which the scheduler-estimator's view of cluster resources does not accurately reflect real-time consumption. Specifically, when the scheduler evaluates the second request, the resources consumed by the first Deployment have not yet been deducted from the overall allocatable resources of Cluster A, leading to an overly optimistic scheduling decision.

The Karmada scheduler should accurately account for the resources consumed by previously scheduled but not yet realized workloads. After Deploy1 is successfully scheduled to Cluster A, the scheduler should recognize that Cluster A's remaining resources are insufficient for Deploy2. Consequently, Deploy2 should either be scheduled to another eligible cluster with sufficient resources (if specified in its PropagationPolicy) or its scheduling should fail with a clear error message indicating resource constraints.

1. Set up a Karmada environment with at least one member cluster (Cluster A) having limited resources (e.g., only enough CPU/memory to run one instance of your target Deployment).
2. Create two identical Deployment YAML files (e.g., deploy1.yaml and deploy2.yaml). Ensure that the sum of their resource requests (e.g., for CPU and memory) exceeds Cluster A's capacity, while a single Deployment's resource request remains within Cluster A's limits.
3. Create a PropagationPolicy for these Deployments configured with schedulingMode: Aggregated and clusterAffinity set to include only Cluster A.
4. Apply both Deployment manifests in very quick succession (e.g., using a script):
`kubectl apply -f deploy1.yaml && kubectl apply -f deploy2.yaml`
6. Check the scheduling results:
- Use kubectl get resourcebinding <deploy1-name> -o yaml and kubectl get resourcebinding <deploy2-name> -o yaml. You will observe that both Deployments are scheduled to Cluster A.
- Check the Pod status in Cluster A using kubectl get pods -l app=<deployment-label>. Pods from one Deployment will be in Pending state. Describing a pending Pod (kubectl describe pod <pod-name>) will show events with reasons like Insufficient cpu or Insufficient memory.

Maybe we can refer to the Kubernetes scheduler AssumePod mechanism to solve it. I haven't figured it out yet.

**Environment**:
- Karmada version: master
- kubectl-karmada or karmadactl version : NA
- Others:

XiShanYongYe-Chang:
**What type of PR is this?**

/kind feature
/kind documentation

ref https://github.com/karmada-io/karmada/issues/6783

Fixes #6783 

```release-note
NONE
```

### DIFF
--- a/docs/proposals/scheduling/estimator-reservation/README.md
+++ b/docs/proposals/scheduling/estimator-reservation/README.md
@@ -1,0 +1,684 @@
+---
+title: Multi-Cluster Workload Scheduling with Resource Reservation
+authors:
+- "@XiShanYongYe-Chang"
+reviewers:
+- "@RainbowMango"
+- "@GitHubxsy"
+- "@mszacillo"
+- "@zhzhuang-zju"
+- "@vie-serendipity"
+approvers:
+- "@RainbowMango"
+creation-date: 2026-04-15
+
+---
+
+# Multi-Cluster Workload Scheduling with Resource Reservation
+
+## Summary
+
+The karmada-scheduler processes the scheduling queue serially, handling one item at a time. The
+karmada-scheduler-estimator (hereinafter referred to as the Estimator) tracks only Pods that have
+been **bound to a node by the member cluster's kube-scheduler** (`pod.Spec.NodeName != ""`).
+After Karmada submits a scheduling decision, the corresponding Pods must traverse the full pipeline
+(`work distribution → member cluster Pod creation → member cluster kube-scheduler node binding`)
+before their resource consumption becomes visible in the Estimator's node model.
+
+As a result, when multiple workloads targeting the same cluster are scheduled in rapid succession,
+subsequent estimation queries all read the same "stale snapshot" — the resource consumption of
+earlier decisions is invisible to the Estimator — causing cluster resources to be over-committed
+and some workloads to fail due to insufficient resources.
+
+**Root cause**: There is an **architectural time gap** between when Karmada submits a scheduling
+decision (milliseconds) and when the Estimator observes that consumption (seconds, depending on
+work distribution and member cluster kube-scheduler node binding). Pods that have not yet been
+created or bound are excluded from the Estimator's node model by design, which is the fundamental
+cause of this problem.
+
+This proposal addresses the problem through two cooperative mechanisms:
+
+1. **Scheduler-side cache reservation for submitted-but-not-yet-landed resources**: After
+   karmada-scheduler successfully patches a ResourceBinding to karmada-apiserver, it immediately
+   records reservation information for the target ResourceBinding.
+2. **Estimator-side reservation deduction**: Extend the estimator gRPC interface to carry reserved
+   workload information in `MaxAvailableComponentSets` requests. Before
+   computing available replicas, the Estimator deducts the reserved workloads' resource consumption
+   from available nodes/quotas using the **existing First Fit (FF) algorithm**, then computes the
+   available replicas for the current request.
+
+## Motivation
+
+The karmada-scheduler processes the scheduling queue serially with a single worker, but its
+throughput far exceeds the rate at which member clusters synchronize resource state and the rate at
+which Pods are created and bound to nodes. When two ResourceBindings are scheduled in rapid
+succession, the following problem occurs:
+
+```
+Timeline  Scheduler Worker                          Estimator Snapshot (ClusterA)
+T1        Binding1: MaxAvailableReplicas → 10       Snapshot: 0 Pods bound, available = 10
+T2        Binding1: decision → assign 10 → ClusterA
+T3        Binding1: Patch karmada-apiserver ✓
+T4        Binding2: MaxAvailableReplicas → 10       Snapshot: still 0 Pods (Binding1's Pods not yet bound)
+T5        Binding2: decision → assign 10 → ClusterA
+T6        Binding2: Patch karmada-apiserver ✓       ClusterA is now over-committed!
+...
+T+N s     work distribution → member cluster creates Pods (Pending) → kube-scheduler binds to node
+          → Estimator observes NodeName, node snapshot finally updated
+```
+
+After completing the Patch at T3, the scheduler immediately processes Binding2. At this point,
+Binding1's Pods have not even been created in the member cluster, let alone bound to a node by
+kube-scheduler.
+
+The same cluster capacity is consumed by multiple scheduling decisions in succession, leading to
+over-commitment of cluster resources.
+
+This proposal introduces a **scheduler-side reservation cache + Estimator-side reservation
+deduction** mechanism to prevent over-commitment of cluster resources.
+
+### Goals
+
+- Prevent cluster resources from being over-committed across multiple consecutive scheduling
+  decisions targeting the same cluster.
+
+### Non-Goals
+
+- The Estimator produces node-selection results identical to kube-scheduler.
+- Address resource contention in `Duplicated` scheduling mode (that mode does not involve replica
+  count calculation).
+- Eliminate the propagation delay between the API Server and the Estimator's node snapshot.
+
+## Proposal
+
+### User Stories
+
+#### Story 1
+
+As a platform operator, I create two Deployments simultaneously with an `Aggregated` scheduling
+policy, and the target cluster can only accommodate one of them. I expect the second Deployment to
+be scheduled to another eligible cluster (if one exists), or to receive a clear resource-shortage
+error — not for both to be scheduled to the same cluster, leaving Pods silently Pending.
+
+## Design Details
+
+### Overall Architecture
+
+```
+karmada-scheduler (leader election, single active instance)
+│
+├── AssigningResourceBindingCache               [EXTENDED]
+│     Added: bindingKey → BindingReservation{entries, expiresAt}
+│
+├── calculateMultiTemplateAvailableSets()         [MODIFIED]
+│     Passes the current cluster's submitted-but-not-yet-landed reservations
+│     as reservedWorkloads in the ComponentSetEstimationRequest
+│
+├── patchScheduleResultForResourceBinding()      [MODIFIED]
+│   After successful Patch → delta = newReplicas - oldReplicas
+│   delta > 0  → Reserve(bindingKey, clusterName, entry)  // REPLACE semantics
+│   delta <= 0 → no-op
+│   cluster removed → ReleaseClusterReservation(bindingKey, clusterName)
+│
+event_handler.go                                [MODIFIED]
+ResourceBinding AggregatedStatus Healthy → ReleaseClusterReservation(bindingKey, clusterName)
+TTL GC fallback
+
+karmada-scheduler-estimator
+│
+├── gRPC MaxAvailableComponentSets()[MODIFIED: MaxAvailableComponentSetsRequest adds reservedWorkloads]
+│
+├── nodeResource plugin (EXTENDED)
+│     1. Obtain available nodes (cloned, Allocatable = Allocatable - Requested)
+│     2. Convert reservedWorkloads to Component list; consume node resources via FF algorithm
+│     3. Compute available replicas on the post-deduction node state
+│
+└── resourcequota plugin (EXTENDED)
+      1. Compute total resource consumption of reservedWorkloads (replicas × resourceRequest)
+      2. Subtract reserved consumption from ResourceQuota available capacity
+      3. Compute available replicas on the post-deduction quota state
+```
+
+### 1. Extend gRPC Interface (`pkg/estimator/pb/generated.proto`)
+
+Add `ReservedWorkload` message and a `reservedWorkloads` field to
+`MaxAvailableComponentSetsRequest` in the proto definition:
+
+```protobuf
+message ReservedWorkload {
+    // namespace is the namespace to which this reservation belongs.
+    // The resourcequota plugin uses this field to determine which ResourceQuota objects
+    // should have this reservation's resource consumption deducted from their available capacity.
+    optional string namespace = 1;
+    // components contains per-component replica and resource requirements.
+    repeated Component components = 2;
+}
+
+// MaxAvailableComponentSetsRequest extended: new reservedWorkloads field.
+message MaxAvailableComponentSetsRequest {
+    ...
+    // reservedWorkloads represents the scheduling commitments that have been
+    // patched to the API server but not yet reflected in the member cluster's real resource state.
+    repeated ReservedWorkload reservedWorkloads = 4;
+}
+```
+
+> After modifying the `.proto` file, regenerate the Go code:
+> `hack/update-estimator-protobuf.sh`.
+
+Also update the estimator client interface (`pkg/estimator/client/interface.go`). The
+`ReplicaEstimator` interface's `MaxAvailableComponentSets` method already uses
+`ComponentSetEstimationRequest` which is designed for field extension without signature changes.
+Add a `ReservedWorkloads` field to that request struct:
+
+```go
+// ComponentSetEstimationRequest extended: new ReservedWorkloads field.
+type ComponentSetEstimationRequest struct {
+    ...
+
+    // ReservedWorkloads carries in-flight scheduling commitments that have not yet been
+    // reflected in the estimator's node snapshot.
+    // +optional
+    ReservedWorkloads []pb.ReservedWorkload
+}
+```
+
+> `SchedulerEstimator` (`pkg/estimator/client/accurate.go`) implements this interface. It receives
+> `reservedWorkloads` from the scheduler and populates the corresponding gRPC request field
+> directly, without performing any cache read/write of its own.
+
+### 2. Scheduler Side: Extend `AssigningResourceBindingCache` (`pkg/scheduler/cache/cache.go`)
+
+`AssigningResourceBindingCache` already tracks ResourceBindings that have been "submitted to the
+API Server but not yet caught up by the Informer". Its lifecycle closely matches the reservation
+lifecycle required by this proposal, so we extend it directly rather than introducing a new cache
+structure.
+
+**Extension strategy: add an independent `reservations` field alongside the existing `items`
+field, with two separate lifecycle management paths:**
+
+- `items` (existing): stores full ResourceBinding objects, serving the `WorkloadAffinity` feature.
+  Deleted immediately when the Informer catches up and `FullyApplied = True`. Lifecycle logic is
+  **unchanged**.
+- `reservations` (new): stores incremental reservation data per ResourceBinding, serving this
+  proposal. Released via `ReleaseReservation()` when the ResourceBinding's AggregatedStatus
+  reaches Healthy, or via TTL GC as a fallback. Lifecycle is **independent of** `items`.
+
+```go
+// BindingReservation stores the submitted-but-not-yet-landed replica increments for a
+// ResourceBinding across its target clusters.
+type BindingReservation struct {
+    // entries: clusterName → ReservedWorkload
+    // The entry key is always clusterName.
+    entries   map[string]*ReservedWorkload
+    // ExpiresAt is the TTL deadline for this binding's reservation, set when the entry is
+    // written or updated (default: now + 5 minutes). Under normal conditions the reservation is
+    // proactively released when the ResourceBinding's AggregatedStatus for the corresponding
+    // cluster reaches Healthy via ReleaseReservation().
+    // However, if a downstream failure (e.g. persistent work distribution failure) prevents the
+    // Healthy state from ever being reached, the proactive release is never triggered.
+    // ExpiresAt serves as a safety net: a background GC() periodically scans and removes all
+    // expired entries, ensuring that stale reservations do not linger in the cache and skew
+    // subsequent scheduling estimates.
+    ExpiresAt time.Time
+}
+
+// ReservedWorkload represents the submitted-but-not-yet-landed scheduling reservation for a
+// specific cluster.
+type ReservedWorkload struct {
+    // Namespace is the namespace of the reserved workload. GetReservedWorkloads() copies this
+    // value into ReservedWorkload.Namespace so that the resourcequota plugin can determine
+    // which ResourceQuota objects should have this reservation deducted.
+    Namespace string
+    // Components contains per-component replica and resource requirements.
+    // Each Component.Replicas is set to the replica increment (delta).
+    Components []workv1alpha2.Component
+}
+
+// AssigningResourceBindingCache (extended)
+type AssigningResourceBindingCache struct {
+    ...
+
+    // reservations: bindingKey → *BindingReservation (new field, serves this proposal)
+    reservations map[string]*BindingReservation
+    // ttl is the lifetime of a reservation entry, used by Reserve() to compute ExpiresAt.
+    // Default 5 minutes; configurable via scheduler flag --reservation-ttl.
+    ttl          time.Duration
+}
+```
+
+**Why TTL is necessary.**
+Releasing a reservation depends on the entire downstream pipeline completing successfully. If a
+downstream failure prevents the workload from ever reaching `Healthy` state, the reservation can
+never be proactively released, continuously skewing subsequent scheduling estimates. Example
+failure scenarios (non-exhaustive):
+
+1. **Persistent Work distribution failure**: If the work controller cannot apply the manifest to
+   the member cluster (e.g. RBAC misconfiguration, member cluster unreachable, APIServer
+   connection timeout), the resource is never created in the member cluster, the ResourceBinding's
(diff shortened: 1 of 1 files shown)


XiShanYongYe-Chang:
cc @RainbowMango @mszacillo @GitHubxsy @zhzhuang-zju

docs/proposals/scheduling/estimator-reservation/README.md
```diff
@@ -0,0 +1,737 @@
...
+    Replicas int32 `json:"replicas,omitempty" protobuf:"varint,1,opt,name=replicas"`
+    // ReplicaRequirements contains the resource and scheduling constraints of the reserved
+    // workload (for single-template workloads).
+    // +optional
+    ReplicaRequirements ReplicaRequirements `json:"replicaRequirements,omitempty" protobuf:"bytes,2,opt,name=replicaRequirements"`
+    // Components contains per-component replica and resource requirements
+    // (for multi-template workloads). When non-empty, Replicas and ReplicaRequirements are not used.
+    // +optional
+    Components []Component `json:"components,omitempty" protobuf:"bytes,3,rep,name=components"`
+}
```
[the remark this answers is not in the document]
XiShanYongYe-Chang: Thanks.

docs/proposals/scheduling/estimator-reservation/README.md
```diff
@@ -0,0 +1,737 @@
...
+```
+
+**New methods:**
+
+| Method | Description |
+|--------|-------------|
+| `Reserve(bindingKey, clusterName string, entry *ReservationEntry)` | Called when delta > 0 after a successful Patch; refreshes ExpiresAt; writes with REPLACE semantics |
+| `ReleaseClusterReservation(bindingKey, clusterName string)` | Called immediately when a cluster is removed from the placement; deletes that cluster's reservation entry |
+| `ReleaseReservation(bindingKey string)` | Triggered with a delay of `reservationReleaseDelay` after FullyApplied; deletes all reservations for the binding |
```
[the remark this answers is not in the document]
XiShanYongYe-Chang: Thanks

docs/proposals/scheduling/estimator-reservation/README.md
```diff
@@ -0,0 +1,743 @@
...
+
+The resource interpreter provides a `Healthy` hook to determine whether a resource on the member
+cluster has reached a healthy state. By watching Work object status changes and checking the
+`health` field of each `status.manifestStatuses` entry: when all manifestStatuses of the Work
+corresponding to a given cluster are `Healthy`, the workload's Pods on that cluster are all
+Running+Ready, and it is safe to release the reservation.
+
+**Assessment:**
+
+From a safety standpoint, this option guarantees that releasing the reservation does not cause
```
mszacillo, line 406: Generally I agree that Option C seems to be the most desirable approach here, but it does introduce a potentially significant amount of pessimism into the scheduling, especially if workloads take a couple minutes to come up and be fully "healthy". For instance, if we're failing over hundreds of FlinkDeployments from one cluster to another, many of those applications can take 5+ minutes to fully become healthy. As the number of reservations starts to grow, the amount of double deductions we make from the cluster capacity starts to get really large as well (since most of these pods will be already be assigned to nodes anyways). 

I think its worth making this limitation very clear to users, and that its probably suggested to tune the ttl setting to be lower than the default 5 minutes. At least that is what I suspect we will have to do.

docs/proposals/scheduling/estimator-reservation/README.md
```diff
@@ -0,0 +1,743 @@
...
+The `nodeResource` plugin uses the First Fit (FF) algorithm to consume reserved resources at the
+**node level**. Implementation is in
+`pkg/estimator/server/framework/plugins/noderesource/noderesource.go`.
+
+**`Estimate` (single-template) flow:**
+
+1. Call `getNodesAvailableResources(snapshot)` to obtain a cloned node list. Each node's
+   `Allocatable` has already been adjusted to `Allocatable − Requested`, reflecting the true
+   available resources.
+2. If `reservedWorkloads` is non-empty, convert each entry to a `Component`, collect them into a
```
mszacillo, line 482: Is there anything preventing us from setting the component in the reservation item upon creation rather than converting it on the fly?

mszacillo (commented):
Just looked over this proposal and left a few thoughts. Thank you for writing this up, it is incredibly thorough! 

Out of curiosity, do we think its worth adding a note on how this feature will work with the node autoscaler support for estimator being discussed: https://github.com/karmada-io/karmada/issues/7375? The implementation is probably not too different from the other estimators discussed in this proposal, but perhaps worth mentioning.

docs/proposals/scheduling/estimator-reservation/README.md
XiShanYongYe-Chang: @mszacillo Thank a lot. 
You've pointed out the crux of the issue: during the period when the pod is binding until the workload status becomes healthy, the reserved resources are deducted in advance. This results in these resources not being used in a timely manner, and this period may last quite a while. Ultimately, the TTL mechanism is used as a fallback. 

I will include your feedback in the Risk section and remind users to pay attention to this.

At the same time, better methods are also welcomed to improve the efficiency

docs/proposals/scheduling/estimator-reservation/README.md
XiShanYongYe-Chang: You mean that reservations will fully use Components and no longer use separate Replicas and ReplicaRequirements, right? That's great! I've thought about it too, and there's actually no obstacle. 

I was wondering whether we should wait until later to deprecate Replicas and ReplicaRequirements and handle it all at once, or whether we should start using Components directly now. The former may be a bit easier to understand at the moment. Of course, I am now inclined towards the latter and would like to see what others think.

XiShanYongYe-Chang:
> Out of curiosity, do we think its worth adding a note on how this feature will work with the node autoscaler support for estimator being discussed: #7375? The implementation is probably not too different from the other estimators discussed in this proposal, but perhaps worth mentioning.

Thanks @mszacillo let me do it.

docs/proposals/scheduling/estimator-reservation/README.md
```diff
@@ -0,0 +1,743 @@
...
+- `MaxAvailableComponentSets`: populated in `ComponentSetEstimationRequest.ReservedWorkloads` and
+  forwarded to the gRPC request body by `SchedulerEstimator`.
+
+> Note: `calAvailableReplicas` is currently a plain function with no direct access to
+> `schedulerCache`. It can be refactored into a method on Scheduler, or the reservation data can
+> be threaded through as a parameter further up the call chain.
+
+### 4. Scheduler Side: Modify `patchScheduleResultForResourceBinding()` (`pkg/scheduler/scheduler.go`)
+
+After a successful API Server Patch, update the reservation based on
```
vie-serendipity, line 340: For now, patching and scheduling are synchronous. If we make scheduling asynchronous in the future to optimize speed, would it be more appropriate to reserve first and then patch?

Furthermore, would it be better to treat reserve as an interface—like the kube-scheduler?

docs/proposals/scheduling/estimator-reservation/README.md
```diff
@@ -0,0 +1,743 @@
...
+Trigger release precisely by detecting through the member cluster Pod informer that a Pod has been
+bound to a node.
+
+```
+Pros: Most accurate release timing; shortest protection window; no empirical values needed.
+Cons: Requires the Karmada control plane to directly observe member cluster Pod state — high
+      architectural invasiveness and implementation complexity.
+```
+
+**Option C: Watch Work status changes, check `manifestStatuses.health` (adopted)**
```
vie-serendipity, line 396: Is there a problem that this could cause duplicate recomputation of resources corresponding to rb?

docs/proposals/scheduling/estimator-reservation/README.md
XiShanYongYe-Chang: Very good advice, and I've heard you say that before. Is there any data support? I think we can add a agenda to the SIG-Scalability Meeting to discuss the optimization of this. WDYT

docs/proposals/scheduling/estimator-reservation/README.md
XiShanYongYe-Chang: You're right, and this is also the issue @mszacillo mentioned https://github.com/karmada-io/karmada/pull/7416#discussion_r3108043891. During this period, reserved resources will be subject to additional deductions.

docs/proposals/scheduling/estimator-reservation/README.md
mszacillo: If using replicas is more intuitive for now, then no problem! I don't have a super strong opinion, I just remembered some conversations about deprecating replicas, so I was unsure if we want to increase the scope of the eventual deprecation effort.

XiShanYongYe-Chang: Thanks, let me modify it. Using comments for estimators may be more intuitive

docs/proposals/scheduling/estimator-reservation/README.md
XiShanYongYe-Chang: @mszacillo Yesterday, I discussed this point with @RainbowMango, and we came up with a direction. We thought about monitoring changes in pods on the estimator side and checking whether they have been bound to nodes. We tried to find the association between pods and workloads, and when deducting reserved resources, we would perform a recheck. If we find that the pods corresponding to the reserved resources to be deducted have already been bound to nodes, we would skip the deduction of that resource. I'm not sure if I've explained this clearly. 

If this direction works, we can handle the deduction operations more precisely.

docs/proposals/scheduling/estimator-reservation/README.md
vie-serendipity: https://github.com/karmada-io/karmada/issues/7062
This issue also mentioned this, but I don’t know its progress. I’m not sure whether it was discussed in the community meeting.

XiShanYongYe-Chang: I don't recall that in my impression, but it's also fine to discuss it further. During the meeting, guys can delve deeper into the issue of asynchronous binding. If some relevant data can be provided as support, I think we can determine the optimization direction more quickly.

docs/proposals/scheduling/estimator-reservation/README.md
```diff
@@ -0,0 +1,743 @@
...
+- @RainbowMango
+- @GitHubxsy
+- @mszacillo
+- @zhzhuang-zju
+approvers:
+- @RainbowMango
+
+creation-date: 2026-04-15
+
+---
```
RainbowMango, line 15: > Error in user YAML: (<unknown>): found character that cannot start any token while scanning for the next token at line 3 column 3

Found this error when previewing the content. 


```suggestion
---
title: Multi-Cluster Workload Scheduling with Resource Reservation
authors:
- "@XiShanYongYe-Chang"
reviewers:
- "@RainbowMango"
- "@GitHubxsy"
- "@mszacillo"
- "@zhzhuang-zju"
approvers:
- "@RainbowMango"

creation-date: 2026-04-15

---
```
This might help, but not sure.

docs/proposals/scheduling/estimator-reservation/README.md
```diff
@@ -0,0 +1,743 @@
...
+
+This proposal introduces a **scheduler-side reservation cache + Estimator-side reservation
+deduction** mechanism to prevent over-commitment of cluster resources.
+
+### Goals
+
+- Prevent cluster resources from being over-committed across multiple consecutive scheduling
+  decisions targeting the same cluster.
+- Maintain in-flight reservation data in the karmada-scheduler cache; propagate it to the
+  Estimator via extended gRPC interfaces.
```
RainbowMango, line 85: Though the karmada-scheduler-estimator is the main estimator implementation, the scope should include the other two implementations.

docs/proposals/scheduling/estimator-reservation/README.md
```diff
@@ -0,0 +1,743 @@
...
+As a platform operator, I create two Deployments simultaneously with an `Aggregated` scheduling
+policy, and the target cluster can only accommodate one of them. I expect the second Deployment to
+be scheduled to another eligible cluster (if one exists), or to receive a clear resource-shortage
+error — not for both to be scheduled to the same cluster, leaving Pods silently Pending.
+
+#### Story 2
+
+As a developer running a load test, I submit a large number of workloads with the `Aggregated`
+policy in rapid succession. I expect the scheduler to correctly deduct the resources already
+committed but not yet landed when evaluating each new request.
```
RainbowMango, line 113: This story is duplicated with the first one.

docs/proposals/scheduling/estimator-reservation/README.md
RainbowMango: Thanks, @mszacillo, for the heads-up; this is a very important input. 

The idea I talked to @XiShanYongYe-Chang about yesterday was that the karmada-scheduler tries its best to provide the reservation list for the estimator. However, if the estimator has more detailed pod information, it can selectively ignore some items from the reservation list. 

The `karmada-scheduler-estimator` has pod information for the member clusters; if it can match this information to the reservation, it can theoretically be more accurate.

docs/proposals/scheduling/estimator-reservation/README.md
XiShanYongYe-Chang: thx, remove it.

docs/proposals/scheduling/estimator-reservation/README.md
XiShanYongYe-Chang: thx, I removed it.

docs/proposals/scheduling/estimator-reservation/README.md
XiShanYongYe-Chang: @mszacillo I update it~

docs/proposals/scheduling/estimator-reservation/README.md
XiShanYongYe-Chang: It works

docs/proposals/scheduling/estimator-reservation/README.md
mszacillo: Ah understood! Generally I like the idea. 

I assume you're referring to something like `podLister` which we use in the estimator when running [listDeploymentPods](https://github.com/karmada-io/karmada/blob/master/pkg/estimator/server/replica/replica.go#L90)? I notice that at the moment we only support this flow for [Deployments ](https://github.com/karmada-io/karmada/blob/master/pkg/estimator/server/replica/replica.go#L51) so we'd have to think about how to extend this to be generalizable to CRDs, which can be complicated. 

With FlinkDeployments, the operator will generate a Deployment + ReplicaSet for the JM pods, but the JM(s) will create TM pods natively without any intermediary K8s resources. Good news is that I believe all pods have a generic `app: <app_name>` label that mirrors the name of the workload, so perhaps pods could be listed directly based on that label + namespace. I think this implementation detail would differ for varying CRDs though.

XiShanYongYe-Chang: @mszacillo 
I have added a separate design regarding this key point, which proposes three methods. Please refer to it.

Hi @RainbowMango @vie-serendipity, can you help take a look at it as well.

docs/proposals/scheduling/estimator-reservation/README.md
```diff
@@ -0,0 +1,744 @@
...
+   the `Healthy` state is never reached.
+3. **Pods fail to run normally**: Pods have been created and bound to nodes but never reach
+   Running+Ready (e.g. image pull failure, CrashLoopBackOff, readiness probe persistent failure,
+   init container stuck). The `Healthy` condition cannot be satisfied, so the reservation lingers
+   in the cache. Note: in this scenario, the Estimator snapshot **already** reflects the Pod's
+   resource consumption (because `pod.Spec.NodeName != ""`), so the reservation is effectively
+   no longer necessary — TTL expiration cleans it up automatically.
+4. **Missing `Healthy` hook**: When no resource interpreter has a registered `Healthy` hook for
+   the resource type, the `health` field in `manifestStatuses` remains empty indefinitely, so the
+   release condition is never satisfied.
```
RainbowMango: This is NOT the truth. A workload's healthy status defaults to `healthy` in case no resource interpreter configured.

docs/proposals/scheduling/estimator-reservation/README.md
```diff
@@ -0,0 +1,744 @@
...
+reliance on a hard-to-calibrate empirical `Δt`, Option C's use of a clear business-semantic signal
+makes its behavior predictable and tuning-free, which provides greater engineering reliability.
+
+Implementation considerations:
+1. karmada-scheduler currently does not watch Work objects; adding this dependency requires
+   evaluating its impact on the scheduler's responsibility boundary.
+2. In multi-cluster scenarios, a single ResourceBinding corresponds to multiple Work objects; Work
+   health must be tracked per cluster.
+3. When a resource has no registered `Healthy` hook, the `health` value is empty; fall back to TTL
+   GC in that case.
```
RainbowMango: 1. It is indeed a big concern if this approach requires `karmada-scheduler` to start watching `Work` object, it will become a heavy burden for the scheduler.
2. Actually, we don't have to get the workloads' healthy status from `Work`, we can get enougf information from the legacy ResourceBinding.

docs/proposals/scheduling/estimator-reservation/README.md
```diff
@@ -0,0 +1,744 @@
...
+resources on exactly one node per replica, avoiding the broadcast deduction across all
+constraint-matching nodes, and is significantly more accurate than cluster-level aggregate
+deduction.
+
+**Reservation release and TTL.**
+Reservations are released per cluster after all manifestStatuses of the corresponding Work object
+reach `Healthy` state — using the workload being fully healthy (all Pods Running+Ready) as the
+release signal ensures no over-commitment is introduced. If a resource has no registered `Healthy`
+hook, causing the health value to remain empty indefinitely, the TTL (default 5 minutes) serves as
+a fallback to prevent stale entries from persisting in the cache.
```
RainbowMango: not truth.
ditto.

docs/proposals/scheduling/estimator-reservation/secondary-verification-en.md
```diff
@@ -0,0 +1,393 @@
...
+adjustedDeduction    = min(deltaResources, unreflectedResources)
+```
+
+## Three Candidate Proposals
+
+### Proposal A: Bottom-Up OwnerReferences Traversal (Recommended)
+
+#### Core Idea
+
+The Estimator already caches all node-bound Pods (complete objects including `OwnerReferences`). By traversing the OwnerReferences chain bottom-up from Pods, it can find Pods belonging to the reserved workload, calculate the already-reflected resource amount, and reduce the deduction accordingly.
```
RainbowMango: Traversing all pods in every scheduling cycle will significantly slow scheduling.
This is the biggest concern.

docs/proposals/scheduling/estimator-reservation/README.md
```diff
@@ -0,0 +1,744 @@
...
+## Future Extensions
+
+The current FF deduction uses a First Fit strategy (greedy in node order), which diverges from the
+LeastAllocated spreading strategy used by default in member cluster kube-schedulers. If more
+accurate simulation of actual scheduling behavior is needed in the future, a pluggable
+`NodeScorePlugin` extension point can be introduced in
+`pkg/estimator/server/framework/interface.go`, with built-in `LeastAllocatedScorer` (spreading)
+and `MostAllocatedScorer` (bin-packing) strategies selectable via an estimator startup flag to
+match the member cluster's actual scheduling configuration. This extension does not affect the core
+functionality of this proposal and can be delivered as an independent follow-on enhancement.
```
RainbowMango: I don't think this is an extension point of this proposal.

The ideas listed in `secondary-verfication-en.md` deserve to be listed here. (Please be brief.)

RainbowMango (commented):
Generally, it looks good to me. 
Just briefly summarize what we need to do with this proposal here(ensure we are all on the same page):
- Maintain a list of pending `ResourceBindings' in the `karmada-scheduler` cache
  - Append ResourceBinding to the list after its scheduling is completed.
  - Remove the ResourceBinding from the list when its status becomes healthy.
  - Set a TTL to ensure that the ResourceBinding won't stay in the list forever.
- When evaluating available replicas, the `karmada-scheduler` passes this list to the estimator.
- When the estimator evaluates resources, it first deducts the resources required by these pending ResourceBindings.

PS: Although several further optimization plans are listed in the `secondary-verification-en.md` file, none of them are ideal. So, this attempt will be excluded from this proposal. (A separate proposal is needed when there is a need.)

@mszacillo Michas, do you have any further comments or concerns?

docs/proposals/scheduling/estimator-reservation/README.md
XiShanYongYe-Chang: > If we make scheduling asynchronous in the future to optimize speed, would it be more appropriate to reserve first and then patch?

Hi @vie-serendipity, Sorry, I didn't understand your point earlier. What you said makes a lot of sense, but I think we might want to consider this issue again when we implement the asynchronous patch.

> Furthermore, would it be better to treat reserve as an interface—like the kube-scheduler?

The current reservation only operates on the scheduler's cache. It is more similar to the assumed step in kube-scheduler operations. If there is a need for handling similar reservation plugins in the future, we can further consider it.

docs/proposals/scheduling/estimator-reservation/README.md
XiShanYongYe-Chang: Thanks

docs/proposals/scheduling/estimator-reservation/secondary-verification-en.md
XiShanYongYe-Chang: +1

Regarding the design of this supplementary plan, I will create a separate PR to track it, so as not to affect the progress of the current main plan.

docs/proposals/scheduling/estimator-reservation/README.md
XiShanYongYe-Chang: Thanks~

XiShanYongYe-Chang:
All comments have been updated ~

docs/proposals/scheduling/estimator-reservation/README.md
```diff
@@ -0,0 +1,684 @@
...
+`AssigningResourceBindingCache` are not rebuilt from the informer resync. In the window between a
+restart and the point where member cluster Pods are Running and observed by the Estimator's node
+snapshot, new concurrent scheduling decisions may see stale cluster capacity, resulting in a brief
+over-commitment risk. As Pods come up one by one, the Estimator snapshot naturally reflects actual
+resource consumption and the situation self-heals. This is a low-probability edge case and
+represents an acceptable engineering trade-off.
+
+### Risks and Mitigations
+
+| Risk | Mitigation |
```
mszacillo, line 571: nit: Add risk mentioning how resources that take longer to become healthy can cause the scheduler to be overly pessimistic with estimations of the cluster's available capacity, since the resources pods have already been assigned to nodes but the reservation is not yet released.

mszacillo:
Did one more pass over the proposal, overall it looks good! Thank you for addressing all the comments @XiShanYongYe-Chang. I left a single nit comment, and will review the secondary verification proposal when ready. 

Let me know when you create the umbrella item for this proposal! I'd be happy to contribute.

docs/proposals/scheduling/estimator-reservation/README.md
XiShanYongYe-Chang: Thanks @mszacillo 
I updated the proposal.

RainbowMango (approved):
/lgtm
/approve

XiShanYongYe-Chang:
> Did one more pass over the proposal, overall it looks good! Thank you for addressing all the comments @XiShanYongYe-Chang. I left a single nit comment, and will review the secondary verification proposal when ready.
> 
> Let me know when you create the umbrella item for this proposal! I'd be happy to contribute.

Hi @mszacillo, I created an umbrella issue to track the task. Welcome to claim the tasks you are interested in. If there are any missing tasks, please feel free to point them out. Thank you. :)

PR #7422, opened 2026-04-20
Fix/webhook validation panic

### LINKED ISSUE #7260: fix: prevent nil pointer dereference in ValidateApplicationFailover
## **What happened**:

In `ValidateApplicationFailover()` (`pkg/util/validation/validation.go`),  
`DecisionConditions.TolerationSeconds` is defined as `*int32` (optional), but it is dereferenced without a nil check.

Current code:

if *applicationFailoverBehavior.DecisionConditions.TolerationSeconds < 0 {

If `TolerationSeconds` is nil, this dereference can cause a panic.

Within the same function, another optional pointer field (`GracePeriodSeconds`) is correctly guarded with a nil check before dereferencing, which makes the handling of `TolerationSeconds` inconsistent.

---

## **What you expected to happen**:

Optional pointer fields should be checked for nil before dereferencing.

The validation should follow the same pattern used for `GracePeriodSeconds`.

Example expected logic:

if applicationFailoverBehavior.DecisionConditions.TolerationSeconds != nil &&
   *applicationFailoverBehavior.DecisionConditions.TolerationSeconds < 0 {

---

## **How to reproduce it (as minimally and precisely as possible)**:

From the codebase:

1. Navigate to:

pkg/util/validation/validation.go

2. Locate the function:

ValidateApplicationFailover()

3. Around line ~271 the following dereference appears:

if *applicationFailoverBehavior.DecisionConditions.TolerationSeconds < 0 {

4. Since `TolerationSeconds` is defined as `*int32` and marked optional, passing a struct where this field is nil would cause a panic.

Example:

behavior := &policyv1alpha1.ApplicationFailoverBehavior{
    DecisionConditions: &policyv1alpha1.DecisionConditions{
        TolerationSeconds: nil,
    },
}

_ = ValidateApplicationFailover(behavior)

This dereference would panic because `TolerationSeconds` is nil.

5. In contrast, the same function correctly guards `GracePeriodSeconds`:

if applicationFailoverBehavior.GracePeriodSeconds != nil &&
   *applicationFailoverBehavior.GracePeriodSeconds <= 0 {

---

## **Anything else we need to know?**:

CRD defaults (`+kubebuilder:default=300`) may prevent this from happening in the webhook path during normal cluster operations. However, `ValidateApplicationFailover()` is a public validation function and could be called from tests, tooling, or future code paths where defaulting is not guaranteed.

Adding a nil check would improve consistency and safety.

---

## **Environment**:
- Karmada version: current `master` branch
- kubectl-karmada or karmadactl version (the result of `kubectl-karmada version` or `karmadactl version`): N/A
- Others: issue discovered during code review of validation logic

manmathbh:
/kind bug

This PR fixes a critical nil pointer dereference panic in the webhook validation logic for `ApplicationFailoverBehavior`.

### Root Cause:
In `ValidateApplicationFailover`, the `DecisionConditions.TolerationSeconds` field (which is an optional `*int32`) was being unconditionally dereferenced. If a user omitted this field, the webhook would panic and crash.

### What Changed:
- Added a defensive `nil` check before dereferencing `TolerationSeconds`, mirroring the existing safe pattern used for `GracePeriodSeconds` in the same function.
- Added a specific table-driven test case (`tolerationSeconds is nil`) to `validation_test.go` to ensure omitting this optional field passes validation without panicking.

Fixes #7260

The fix is highly localized and fully covered by unit tests.

```release-note
`karmada-webhook`: Fixed a panic in `ValidateApplicationFailover` that occurred when `TolerationSeconds` was omitted.

### DIFF
--- a/pkg/detector/detector.go
+++ b/pkg/detector/detector.go
@@ -636,11 +636,13 @@ func (d *ResourceDetector) ApplyClusterPolicy(object *unstructured.Unstructured,
 				bindingCopy.Spec.ReplicaRequirements = binding.Spec.ReplicaRequirements
 				bindingCopy.Spec.Replicas = binding.Spec.Replicas
 				bindingCopy.Spec.Components = binding.Spec.Components
+				bindingCopy.Spec.PropagateDeps = binding.Spec.PropagateDeps
 				bindingCopy.Spec.SchedulerName = binding.Spec.SchedulerName
 				bindingCopy.Spec.Placement = binding.Spec.Placement
 				bindingCopy.Spec.Failover = binding.Spec.Failover
 				bindingCopy.Spec.ConflictResolution = binding.Spec.ConflictResolution
 				bindingCopy.Spec.PreserveResourcesOnDeletion = binding.Spec.PreserveResourcesOnDeletion
+				bindingCopy.Spec.SchedulePriority = binding.Spec.SchedulePriority
 				bindingCopy.Spec.Suspension = util.MergePolicySuspension(bindingCopy.Spec.Suspension, policy.Spec.Suspension)
 				return nil
 			})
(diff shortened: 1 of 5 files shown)


XiShanYongYe-Chang:
@manmathbh According to the discussion in the issue, we do not need to fix this.

manmathbh:
> @manmathbh According to the discussion in the issue, we do not need to fix this.

Understood! Thanks @XiShanYongYe-Chang  for the clarification, I will close this out to keep the PR queue clean. Have a great week!

XiShanYongYe-Chang:
Thanks @manmathbh

Issue #7390, opened 2026-04-14
[Feature] Overflow Cluster Affinities Scheduling


zhzhuang-zju:
original issue: #7014 
Target release: release-1.18

Tasks:
- [ ] Proposal: https://github.com/karmada-io/karmada/pull/7078 (@zhzhuang-zju )
- [ ] API change: https://github.com/karmada-io/karmada/pull/7386 (@zhzhuang-zju )
- [ ] Scheduler/karmada-controller-manager Change (@zhzhuang-zju )
- [ ] ValidationWebhook: add a webhook for the `(Cluster)PropagationPolicy` resource to restrict that `spec.placement.clusterAffinities[i].overflowAffinities` can only be used together with the inline `ClusterAffinity` and must not be set independently.
- [ ] Necessary E2E tests

RainbowMango:
@zhzhuang-zju Just a reminder, do you need to update the iterated tasks to make sure they are consistent with the final proposal?

zhzhuang-zju:
Thank you for the reminder! Updated

Viscous106:
Hi @zhzhuang-zju, congrats on getting the scheduler change (#7423) and validation webhook (#7430) merged!

I read through the proposal at docs/proposals/scheduling/multi-scheduling-group/overflow-affinities/README.md and the merged scheduler logic. I'd like to take the remaining E2E tests sub-task and cover both scenarios from the test plan:

  - **Scale-out (overflow):** Create a Deployment whose replicas exceed primary-tier capacity and assert that excess replicas overflow tier-by-tier following the declared order in `overflowAffinities`.
  - **Scale-in (contraction):** Start with replicas spread across multiple tiers, scale the Deployment down, and assert that replicas are reclaimed in reverse tier order.

  My plan is to model the new tests after `test/e2e/suites/base/clusteraffinities_test.go` and use cluster-label manipulation (similar to `schedule_priority_test.go`) to make tier capacity deterministic. I'll use `Divided` + `dynamicWeight: AvailableReplicas` per the webhook validation rules.

  I'll hold off on opening a PR until you confirm the scope and approach. Happy to adjust if you'd like a different file layout or additional scenarios.

  cc @RainbowMango

zhzhuang-zju:
Thanks @Viscous106, you can take this task.
Actually, I've also been thinking about how to write this E2E test. Here are some my thoughts, hope they help:
[...]

PR #7423, opened 2026-04-21
feature: overflow cluster affinities

zhzhuang-zju:
**What type of PR is this?**
/kind feature

implement of overflow cluster affinities

Part of #7390

```release-note
`karmada-scheduler`: Extended cluster affinities to support overflow cluster resource pools for hybrid cloud scheduling.
```

### DIFF
--- a/pkg/scheduler/core/common.go
+++ b/pkg/scheduler/core/common.go
@@ -28,11 +28,11 @@ import (
 )
 
 // SelectClusters selects clusters based on the placement and resource binding spec.
-func SelectClusters(clustersScore framework.ClusterScoreList, placement *policyv1alpha1.Placement, spec *workv1alpha2.ResourceBindingSpec) ([]spreadconstraint.ClusterDetailInfo, error) {
+func SelectClusters(clustersScore framework.ClusterScoreList, placement *policyv1alpha1.Placement, spec *workv1alpha2.ResourceBindingSpec, status *workv1alpha2.ResourceBindingStatus) ([]spreadconstraint.ClusterDetailInfo, error) {
 	startTime := time.Now()
 	defer metrics.ScheduleStep(metrics.ScheduleStepSelect, startTime)
 
-	groupClustersInfo := spreadconstraint.GroupClustersWithScore(clustersScore, placement, spec, calAvailableReplicas)
+	groupClustersInfo := spreadconstraint.GroupClustersWithScore(clustersScore, placement, spec, status, calAvailableReplicas)
 	return spreadconstraint.SelectBestClusters(placement, groupClustersInfo, spec.Replicas)
 }
 
@@ -55,18 +55,7 @@ func AssignReplicas(clusters []spreadconstraint.ClusterDetailInfo, spec *workv1a
 	// the deprecation of binding.spec.replicas and binding.spec.ReplicaRequirements. After that, the check should be changed to
 	// len(spec.Components) == 1.
 	if (spec.Replicas > 0 || spec.ReplicaRequirements != nil) && len(spec.Components) <= 1 {
-		state := newAssignState(clusters, spec, status)
-		assignFunc, ok := assignFuncMap[state.strategyType]
-		if !ok {
-			// should never happen at present
-			return nil, fmt.Errorf("unsupported replica scheduling strategy, replicaSchedulingType: %s, replicaDivisionPreference: %s, "+
-				"please try another scheduling strategy", spec.Placement.ReplicaSchedulingType(), spec.Placement.ReplicaScheduling.ReplicaDivisionPreference)
-		}
-		assignResults, err := assignFunc(state)
-		if err != nil {
-			return nil, err
-		}
-		return removeZeroReplicasCluster(assignResults), nil
+		return assignWorkloadReplicas(clusters, spec, status)
 	}
 
 	// For non-workloads (e.g., Service, Config) and multi-component workloads (e.g., FlinkDeployment), propagate to all candidate clusters.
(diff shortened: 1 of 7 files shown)


pkg/scheduler/core/spreadconstraint/group_clusters.go
```diff
@@ -335,13 +340,14 @@ func (info *GroupClustersInfo) calcGroupScore(
...
 
-func (info *GroupClustersInfo) generateClustersInfo(clustersScore framework.ClusterScoreList, rbSpec *workv1alpha2.ResourceBindingSpec) {
+func (info *GroupClustersInfo) generateClustersInfo(clustersScore framework.ClusterScoreList, rbSpec *workv1alpha2.ResourceBindingSpec, rbStatus *workv1alpha2.ResourceBindingStatus) {
 	var clusters []*clusterv1alpha1.Cluster
 	for _, clusterScore := range clustersScore {
 		clusterInfo := ClusterDetailInfo{}
 		clusterInfo.Name = clusterScore.Cluster.Name
 		clusterInfo.Score = clusterScore.Score
 		clusterInfo.Cluster = clusterScore.Cluster
+		clusterInfo.Tier = getClusterTier(clusterScore.Cluster, rbSpec, rbStatus)
```
zhzhuang-zju, line 350: Related comment: https://github.com/karmada-io/karmada/pull/7078#discussion_r3108765991

@RainbowMango It would work to determine the cluster tier in either the `Select` stage or the `Prioritize` stage. In the end, the tier information still needs to be carried in `ClusterDetailInfo` so it can be passed to the `Assign` stage. I compared the two approaches, and the implementation effort is roughly the same.
I chose to implement it in the `Select` stage because the Prioritize stage returns scores from score plugins, and attaching tier information there does not feel semantically appropriate. WDYT?

RainbowMango (commented):
/assign

pkg/scheduler/core/spreadconstraint/group_clusters.go
```diff
@@ -79,6 +80,7 @@ type ZoneInfo struct {
 type ClusterDetailInfo struct {
 	Name  string
 	Score int64
+	Tier  int32
```
RainbowMango:
```suggestion
	// OverflowOrder represents the overflow priority level derived from cluster affinity matching.
	// A value of 0 means the cluster matches the primary ClusterAffinity (no overflow),
	// while higher values (1, 2, ...) correspond to successive OverflowAffinities.
	// A smaller value indicates higher scheduling priority.
	OverflowOrder int32
```
We didn't bring the `tier` concept eventually, so we should not use the name in code to avoid ambiguity.

pkg/scheduler/core/spreadconstraint/group_clusters.go
```diff
@@ -496,3 +502,32 @@ func isTopologyIgnored(placement *policyv1alpha1.Placement) bool {
 
 	return shouldIgnoreSpreadConstraint(placement)
 }
+
+// getClusterTier returns the tier of the cluster. A smaller tier value means the cluster has higher priority.
+func getClusterTier(cluster *clusterv1alpha1.Cluster, bindingSpec *workv1alpha2.ResourceBindingSpec, bindingStatus *workv1alpha2.ResourceBindingStatus) int32 {
```
RainbowMango:
```suggestion
func getClusterOverflowOrder(cluster *clusterv1alpha1.Cluster, bindingSpec *workv1alpha2.ResourceBindingSpec, bindingStatus *workv1alpha2.ResourceBindingStatus) int32 {
```

pkg/scheduler/framework/plugins/clusteraffinity/cluster_affinity.go
```diff
@@ -54,22 +54,32 @@ func (p *ClusterAffinity) Filter(
...
+	var affinities []*policyv1alpha1.ClusterAffinity
 	if bindingSpec.Placement.ClusterAffinity != nil {
-		affinity = bindingSpec.Placement.ClusterAffinity
+		affinities = append(affinities, bindingSpec.Placement.ClusterAffinity)
 	} else {
 		for index, term := range bindingSpec.Placement.ClusterAffinities {
 			if term.AffinityName == bindingStatus.SchedulerObservedAffinityName {
-				affinity = &bindingSpec.Placement.ClusterAffinities[index].ClusterAffinity
+				affinities = append(affinities, &bindingSpec.Placement.ClusterAffinities[index].ClusterAffinity)
+				// For non-workloads (e.g., Service, Config) and multi-component workloads (e.g., FlinkDeployment), will not use the affinities defined within `OverflowAffinities`.
```
RainbowMango: Why do you think non-workloads and multi-component workloads will not use this feature?

pkg/scheduler/core/common.go
```diff
@@ -76,3 +65,75 @@ func AssignReplicas(clusters []spreadconstraint.ClusterDetailInfo, spec *workv1a
 	}
 	return targetClusters, nil
 }
+
+func assignWorkloadReplicas(clusters []spreadconstraint.ClusterDetailInfo, spec *workv1alpha2.ResourceBindingSpec, status *workv1alpha2.ResourceBindingStatus) ([]workv1alpha2.TargetCluster, error) {
+	if supportOverflow(spec) {
```
RainbowMango: Here, we should select the workload with an overflow configuration and ensure this piece of code is used only for those instead of the legacy ones.

pkg/scheduler/framework/plugins/clusteraffinity/cluster_affinity.go
zhzhuang-zju: Simply put, for non-workloads, they will not fully consume the resources of a cluster group.

As for multi-component workloads, I don’t think we need to support overflow scheduling for this type of workload, since we currently only support scheduling them to a single member cluster.

pkg/scheduler/framework/plugins/clusteraffinity/cluster_affinity.go
```diff
@@ -54,22 +54,38 @@ func (p *ClusterAffinity) Filter(
...
 	cluster *clusterv1alpha1.Cluster,
 ) *framework.Result {
-	var affinity *policyv1alpha1.ClusterAffinity
+	var affinities []*policyv1alpha1.ClusterAffinity
 	if bindingSpec.Placement.ClusterAffinity != nil {
-		affinity = bindingSpec.Placement.ClusterAffinity
+		affinities = append(affinities, bindingSpec.Placement.ClusterAffinity)
 	} else {
 		for index, term := range bindingSpec.Placement.ClusterAffinities {
 			if term.AffinityName == bindingStatus.SchedulerObservedAffinityName {
```
RainbowMango:
```golang
		for index, term := range bindingSpec.Placement.ClusterAffinities {
			if term.AffinityName != bindingStatus.SchedulerObservedAffinityName {
				continue
			}

			affinities = append(affinities, &bindingSpec.Placement.ClusterAffinities[index].ClusterAffinity)

			// OverflowAffinities are designed for workloads that require actual compute resources
			// (e.g., CPU, memory) to run replicas. When the primary cluster group lacks capacity,
			// the scheduler overflows to additional cluster groups.
			// Non-workload resources (e.g., Service, ConfigMap) do not consume compute resources,
			// so overflow does not apply to them; they use only the primary cluster group.
			if bindingSpec.IsWorkload() {
				for i := range term.OverflowAffinities {
					affinities = append(affinities, &term.OverflowAffinities[i].ClusterAffinity)
				}
			}
		}
```

zhzhuang-zju: Thanks for the suggestion. I kept the `break` statement to avoid unnecessary subsequent iterations.

RainbowMango: Yeah, it makes sense.

RainbowMango (approved):
/lgtm
/approve

PR #7427, opened 2026-04-22
fix(karmadactl): make join and unjoin idempotent

### LINKED ISSUE #4779: Karmadactl join/unjoin is not idempotent
**What happened**:
When I use karmadactl to join a cluster, karmadactl exits abnormally due to environment problems. When I use karmadactl to join the same cluster again, an error message is displayed, indicating that the cluster already exists. （error message：the same cluster has been registered with name cluster1）. According to the code, the join is not idempotent and needs to support reentrant. The unjoin is similar.

Run the karmadactl join command to join the same cluster. A success message is returned.

When karmadactl is used to join a cluster, the karmadactl program is terminated. When karmadactl is used to join the same cluster again, an error is reported.

**Environment**:
- Karmada version:
- kubectl-karmada or karmadactl version (the result of `kubectl-karmada version` or `karmadactl version`): v1.10.0.preview2
- Others:

manmathbh:

/kind bug

This PR makes karmadactl join and unjoin idempotent for common retry scenarios.

**Problem summary:**
1. join failed on retries when the same cluster was already registered, even when the cluster name and cluster ID matched.
2. unjoin failed on retries when the cluster object had already been deleted, returning a not found error instead of treating it as already unjoined.

**What changed:**
1. Updated join flow to allow re-entry for the same cluster identity.
2. Updated cluster creation helpers to treat already exists as success when the existing cluster ID matches.
3. Updated unjoin flow to treat not found as success.
4. Updated cluster deletion helper to return success on not found.
5. Updated join unit tests to expect success for duplicate join of the same cluster identity.
6. Added unjoin unit test coverage for the not found idempotency case.

**Why this is needed:**
1. CLI operations should be safe to retry after partial failures or interrupted runs.
2. This aligns karmadactl behavior with Kubernetes idempotent API semantics for create and delete workflows.

Fixes #4779

**Test report:**
1. go test ./pkg/karmadactl/join -count=1
2. go test ./pkg/karmadactl/unjoin -count=1

Both passed locally.

**Scope notes:**
1. The change is intentionally localized to join and unjoin command paths plus their helper logic and tests.
2. No scheduler, metrics-adapter, or broad architectural behavior is modified.

```release-note
karmadactl: Made join and unjoin idempotent. Retrying join for an already-registered cluster with the same identity now succeeds, and retrying unjoin after the cluster object is already removed now succeeds.
```

### DIFF
--- a/pkg/karmadactl/join/join.go
+++ b/pkg/karmadactl/join/join.go
@@ -233,7 +233,7 @@ func (j *CommandJoinOption) RunJoinCluster(controlPlaneRestConfig, clusterConfig
 		return err
 	}
 
-	if err = registerOption.Validate(karmadaClient, false); err != nil {
+	if err = registerOption.Validate(karmadaClient, true); err != nil {
 		return err
 	}
 
--- a/pkg/karmadactl/join/join_test.go
+++ b/pkg/karmadactl/join/join_test.go
@@ -113,29 +113,23 @@ func TestRunJoinCluster(t *testing.T) {
 			joinOpts:            &CommandJoinOption{},
 			controlPlaneRestCfg: &rest.Config{},
 			clusterCfg:          &rest.Config{},
+			controlKubeClient:   fakeclientset.NewClientset(),
 			clusterKubeClient:   fakeclientset.NewClientset(),
 			karmadaClient:       fakekarmadaclient.NewClientset(),
 			clusterID:           types.UID(uuid.New().String()),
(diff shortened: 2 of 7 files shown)


manmathbh:
@chaunceyjiang @XiShanYongYe-Chang  had a commit message snag (bot flagged Fixes # in the body) , fixed that. 
Also tightened up the empty-ID edge case per Copilot's feedback and cleaned up the error wording.
Tests all pass, branch is clean.

Would be great to get eyes on this one, been ready for a bit.

PR #7432, opened 2026-04-25
fix: preserve application failover timer across ResourceUnknown transitions

### LINKED ISSUE #7431: [bug]: Application failover permanently suppressed when workload oscillates between Unhealthy and Unknown states
**What happened**:

When a workload oscillates between `ResourceUnhealthy` and `ResourceUnknown` states (e.g., during pod restarts, network blips, or karmada-agent interruptions), the application failover timer is silently reset on every transition to `Unknown`. This prevents failover from ever triggering, even when the workload remains genuinely unhealthy for periods exceeding `TolerationSeconds`.

The failover timer should accumulate unhealthy time across `Unhealthy → Unknown → Unhealthy` oscillations. Only a confirmed `ResourceHealthy` status should reset the timer. Failover should trigger when total unhealthy time exceeds `TolerationSeconds`, regardless of intermittent `Unknown` states.

1. Deploy a workload with application failover enabled and `TolerationSeconds: 60`

2. Make the workload unhealthy on cluster-A (e.g., crash the pod)
   - At T=0s: workload becomes `Unhealthy`, timer starts

3. At T=30s, restart the karmada-agent pod on cluster-A
   - Status collection is interrupted
   - Workload health becomes `ResourceUnknown`
   - **Bug:** Failover timer is deleted

4. At T=60s, karmada-agent recovers
   - Workload is still broken, becomes `Unhealthy` again
   - **Bug:** New timer starts at T=60s (not T=0s)

5. At T=90s, any infrastructure event causes another `Unknown` transition
   - Timer deleted again

**Result:** Failover never fires despite workload being unhealthy for 120+ seconds (exceeding the 60s tolerance).

### Root Cause

In `pkg/controllers/applicationfailover/common.go` line 119, `distinguishUnhealthyClustersWithOthers` treats `ResourceUnknown` identically to `ResourceHealthy`:

```go
// Line 119 - BUG: ResourceUnknown treated as healthy
if aggregatedStatusItem.Health == workv1alpha2.ResourceHealthy ||
    aggregatedStatusItem.Health == workv1alpha2.ResourceUnknown {
    others = append(others, cluster)  // Passed to deleteIrrelevantClusters as "healthyClusters"
}
```

The `others` slice is passed to `deleteIrrelevantClusters` whose parameter is named `healthyClusters`. This function **deletes the unhealthy timestamp** for any cluster in that list, resetting the failover timer.

### When ResourceUnknown Occurs

`ResourceUnknown` is set in two common situations (see `workstatus.go:186,196`):
1. `Applied=false` - karmada-agent cannot apply work to member cluster
2. `Applied=true` but no health status yet - before first collection or during gaps

Both happen constantly in production during:
- Rolling upgrades
- Pod restarts (especially karmada-agent)
- Network blips between controller and member clusters
- Member cluster API server restarts

### Impact

- **Silent failure:** No errors logged, failover simply never fires
- **Affects all workload types:** Both ResourceBinding and ClusterResourceBinding paths
- **Production-critical:** Any infrastructure with transient status gaps (i.e., everyone) cannot rely on application failover
- **Misleading test:** Existing test at `common_test.go:125-138` encodes the buggy behavior as correct

### Proposed Fix

In `pkg/controllers/applicationfailover/common.go` line 119:

```go
// Remove ResourceUnknown from the "healthy" classification
if aggregatedStatusItem.Health == workv1alpha2.ResourceHealthy {
    others = append(others, cluster)
}
```

With this fix:
- Unknown clusters are not added to `others`, so their timestamps are preserved
- Timer accumulates across `Unhealthy → Unknown → Unhealthy` oscillations
- Only confirmed `Healthy` resets the timer (matching the intent)

**Environment**:
- Karmada version: master (affects all versions with application failover)
- Others: N/A

Ady0333:
**What type of PR is this?**
/kind bug

---

Fixes a critical bug where application failover is permanently suppressed when workloads oscillate between Unhealthy and Unknown states.

---

Fixes #7431 

---

**Root cause:**
`distinguishUnhealthyClustersWithOthers` was treating `ResourceUnknown` the same as `ResourceHealthy`, causing the failover timer to be deleted on every Unknown transition (agent restarts, network blips, etc.).

---

**Changes:**
- Changed line 119 to only treat `ResourceHealthy` as healthy
- Unknown clusters now preserve their unhealthy timestamps
- Updated test to reflect correct behavior

---

**Does this PR introduce a user-facing change?**
```release-note
Fixed a critical bug where application failover would never trigger if workloads oscillated between Unhealthy and Unknown states due to infrastructure events.
```

### DIFF
--- a/pkg/controllers/applicationfailover/common.go
+++ b/pkg/controllers/applicationfailover/common.go
@@ -116,7 +116,7 @@ func distinguishUnhealthyClustersWithOthers(aggregatedStatusItems []workv1alpha2
 			unhealthyClusters = append(unhealthyClusters, cluster)
 		}
 
-		if aggregatedStatusItem.Health == workv1alpha2.ResourceHealthy || aggregatedStatusItem.Health == workv1alpha2.ResourceUnknown {
+		if aggregatedStatusItem.Health == workv1alpha2.ResourceHealthy {
 			others = append(others, cluster)
 		}
 	}
--- a/pkg/controllers/applicationfailover/common.go
+++ b/pkg/controllers/applicationfailover/common.go
@@ -106,9 +106,11 @@ func (m *workloadUnhealthyMap) deleteIrrelevantClusters(key types.NamespacedName
 	m.workloadUnhealthy[key] = unhealthyClusters
 }
 
-// distinguishUnhealthyClustersWithOthers distinguishes clusters which is in the unHealthy state(not in the process of eviction) with others.
+// distinguishUnhealthyClustersWithOthers distinguishes unhealthy clusters (not in the process of eviction) from healthy ones.
+// ResourceUnknown clusters are intentionally excluded from both lists to preserve their failover timers across transient
+// status collection gaps.
 func distinguishUnhealthyClustersWithOthers(aggregatedStatusItems []workv1alpha2.AggregatedStatusItem, resourceBindingSpec workv1alpha2.ResourceBindingSpec) ([]string, []string) {
-	var unhealthyClusters, others []string
+	var unhealthyClusters, healthyClusters []string
 	for _, aggregatedStatusItem := range aggregatedStatusItems {
 		cluster := aggregatedStatusItem.ClusterName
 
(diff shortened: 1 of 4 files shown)


Ady0333:
Hello @XiShanYongYe-Chang !!! Please review this PR and let me know if any changes are required...

Issue #7446, opened 2026-04-28
[Umbrella] Replace UpdateStatus() implementation and adjust RetryOnConflict() usage


zach593:
**What would you like to be added**:

Replace the current `helper.UpdateStatus()` implementation with a version that compares the newly computed status against the object instance fetched at the beginning of the current reconciliation, instead of re-fetching the object before doing the deep-equal check. That could potentially lead to reconciliation skew. This idea was suggested by @RainbowMango.

In addition, adjust how `RetryOnConflict()` is used in controllers:

- Status updates should no longer use a local `RetryOnConflict()` wrapper.
- Instead, `RetryOnConflict()` should wrap the outer reconcile flow, so that when a conflict happens, the controller retries the reconciliation using the latest object version.
- If the outer reconcile flow already uses `RetryOnConflict()`, individual operations inside the reconcile logic should not introduce another local `RetryOnConflict()` wrapper.

- [ ] cluster_status_controller
- [ ] cronfederatedhpa_job
- [ ] endpointslice_dispatch_controller
- [ ] execution_controller
- [ ] federated_resource_quota_enforcement_controller
- [ ] federated_resource_quota_status_controller
- [ ] mcs_controller
- [ ] remedy_controller
- [ ] service_import_controller
- [ ] work_status_controller

The changes to `helper.UpdateStatus()` should cover all mentioned controllers, but the changes to `RetryOnConflict()` may not be necessary for every controller.

**Why is this needed**:

Several issues have reported incorrect status convergence caused by controllers, for example:

- https://github.com/karmada-io/karmada/issues/6858
- https://github.com/karmada-io/karmada/issues/7227

A major cause is that many controllers currently use deep-equal checks together with informer cache in order to reduce unnecessary HTTP requests. However, this can lead to reconciliation skew.

When a controller enters reconciliation, it is operating on object version A. If it later needs to update status, the deep-equal comparison should still be made against version A, which is the object instance used to compute the new status. It should not be compared against version B or any newer version fetched later from informer cache or the API server.

The current `helper.UpdateStatus()` implementation re-gets the object before updating status. If the object has changed during reconciliation, this may lead to status comparison across different versions, which can break eventual consistency.

Also, `RetryOnConflict()` should be placed carefully. A `409 Conflict` during status update should not be retried only at the status-update level, because the controller needs to restart the whole reconciliation and recompute status from the latest object version.

[...]

Issue #7111, opened 2026-01-15
[Observability] Stop recording optimistic concurrency conflict errors that are retried in metrics


jabellard:
**What would you like to be added**:

The `work_sync_workload_duration_seconds` metric should not count Kubernetes conflict errors (HTTP 409) as failures. These are expected retriable errors in distributed systems and should not be recorded with `result="error"`.

**Why is this needed**:

## Problem: Conflict Errors Cause False SLO Violations

In our environment, this metric shows **15% error rate** (16.23x burn rate) causing critical alerts, but the system is actually healthy. Nearly all "errors" are Kubernetes conflict errors that the controller successfully retries.

### Sample Logs

```
E0115 02:56:14.736986 Failed to update resource(kind=Deployment, default/nginx-deployment)
in cluster test-cluster-region1, err: Operation cannot be fulfilled on
deployments.apps "nginx-deployment": the object has been modified

E0115 02:56:19.000869 Failed to update resource(kind=Deployment, default/nginx-deployment)
in cluster test-cluster-region1, err: Operation cannot be fulfilled on
deployments.apps "nginx-deployment": the object has been modified

[Pattern repeats - controller retries and eventually succeeds]
```

These are HTTP 409 Conflict errors from Kubernetes optimistic concurrency control - normal, expected, and automatically retried.

### Current Metrics

```promql
# Error rate: 15.8% (target is 99% success)
sum(rate(work_sync_workload_duration_seconds_count{result="error"}[5m])) /
sum(rate(work_sync_workload_duration_seconds_count[5m]))

# Burn rate: 16.23x (burning error budget 16x faster than allowed)
slo:current_burn_rate:ratio{sloth_slo="work-sync-workload-availability"}
```

Nearly all these "errors" are conflict errors that successfully retry.

### Why This Matters

**Conflict errors are not failures:**
- Temporary and automatically retried (controller succeeds on next attempt)
- Expected in distributed systems (Kubernetes optimistic concurrency control)
- Not user-facing (workloads eventually sync successfully)

**Impact:**
- False critical alerts (teams paged for "healthy" system)
- Misleading dashboards (85% availability shown when actual is >99%)
- Alert fatigue and wasted engineering time

**Current:** Metric measures retry rate ("Did first attempt succeed?")
**Expected:** Metric should measure availability ("Did workload eventually sync?")

---

## Proposed Change

**File:** `pkg/controllers/execution/execution_controller.go`

Skip recording the metric:

```go
[...]

PR #7447, opened 2026-04-28
fix(execution): centralize conflict retry and align status updates

### LINKED ISSUE #6858: AggregatedStatus in rb status may be deleted

zach593:
**What type of PR is this?**

/kind bug

This PR fixes conflict handling and status update behavior in `execution_controller`.

The current implementation may compare a status computed from the object observed at the beginning of reconciliation with a newer object version re-fetched later from the API server or informer cache. In the conflict path, this can cause `AppliedSuccessful` not to be written back correctly, and eventually lead to incorrect status aggregation such as the issue reported in `#6858`.

To address this, this PR:

- moves `RetryOnConflict()` to the outer `Reconcile()` flow, so a `409 Conflict` retries the whole reconciliation against the latest object version
- removes local conflict retry around individual workload sync operations
- introduces `helper.UpdateStatusFromExisting()` and uses it in `execution_controller` status updates
- updates status based on the object version already observed by the current reconciliation, instead of re-fetching it before the deep-equal check

Removing local conflict retries also avoids nested retries. As discussed in #7111, not every 409 Conflict should immediately result in a reconcile error. However, when both the outer reconcile flow and individual operations use RetryOnConflict(), a reconcile error may not be returned until both retry layers are exhausted. With the default retry settings, this can mean up to 5 * 5 = 25 conflicts before a reconcile error is surfaced.

Fixes #6858

Part of #7446

```release-note
karmada-controller-manager: Fixed an issue in the execution controller where status updates could be computed against a stale object version after conflicts, which might cause incorrect Work and aggregated status convergence.
```

### DIFF
--- a/pkg/controllers/execution/execution_controller.go
+++ b/pkg/controllers/execution/execution_controller.go
@@ -82,6 +82,17 @@ type Controller struct {
 func (c *Controller) Reconcile(ctx context.Context, req controllerruntime.Request) (controllerruntime.Result, error) {
 	klog.V(4).InfoS("Reconciling Work", "work", req.NamespacedName.String())
 
+	var result controllerruntime.Result
+	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		var err error
+		result, err = c.reconcile(ctx, req)
+		return err
+	})
+	return result, err
+}
+
+// reconcile performs the actual reconciliation logic for the Work.
+func (c *Controller) reconcile(ctx context.Context, req controllerruntime.Request) (controllerruntime.Result, error) {
 	work := &workv1alpha1.Work{}
 	if err := c.Client.Get(ctx, req.NamespacedName, work); err != nil {
 		// The resource may no longer exist, in which case we stop processing.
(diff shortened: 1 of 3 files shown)


zach593:
Closing this per the analysis at https://github.com/karmada-io/karmada/issues/7446#issuecomment-4497446182

PR #7457, opened 2026-05-01
Fix issue with installing Karmada operator CRDs via Helm - Phase 2

jabellard:
**What type of PR is this?**

/kind bug

The operator CRD is ~1.1MB, which exceeds the 1MB ConfigMap limit. The post-hook mechanism embeds the CRD into a ConfigMap to work around Helm not upgrading CRDs from `crds/`, but that hits a hard ceiling — and the CRD will only keep growing as the operator API surface expands. As that happens, any ConfigMap-based approach will eventually hit that limit.

This PR takes the same approach as cert-manager: place the CRD directly in `templates/` so Helm manages it as a regular resource. No ConfigMap, no Job, no size constraint.

cert-manager references:
- [Design doc explaining the rationale](https://github.com/cert-manager/cert-manager/blob/master/design/20240206.helm-resource-policy.md)
- [CRD template header](https://github.com/cert-manager/cert-manager/blob/master/make/_shared_new/helm/crd.template.header.yaml)
- [Build pipeline (crds.mk)](https://github.com/cert-manager/cert-manager/blob/master/make/_shared_new/helm/crds.mk)
- [Values with crds.keep](https://github.com/cert-manager/cert-manager/blob/master/deploy/charts/cert-manager/values.yaml)

To prevent `helm uninstall` from cascade-deleting all Karmada CRs, the CRD template carries the [`helm.sh/resource-policy: keep`](https://helm.sh/docs/howto/charts_tips_and_tricks/#tell-helm-not-to-uninstall-a-resource) annotation (controlled by `crds.keep`, default `true`). Same uninstall safety as the old `crds/` approach.

### Behavior Change

| Scenario | Before | After |
|---|---|---|
| `helm install` | Fails (ConfigMap > 1MB) | Works |
| `helm upgrade` | Fails (ConfigMap > 1MB | CRD updated |
| `helm uninstall` | CRD survives | CRD survives (`resource-policy: keep`) |
| `installCRDs=false` | Post-hook skipped | CRD template excluded |

Part of #7371

```release-note
`karmada-operator-chart`: Fixed issue with embedding the `Karmada` CRD, which has exceeded the config map size limit, into a config map used by the chart CRD installation/upgrade job, that prevented users from upgrading the chart.  A temporary patch was applied to release `v1.17.3`. To ensure a smooth upgrade, users must first upgrade to release `v1.17.3+` before upgrading to a `v.18.x` release.
```


seanlaii (commented):
The approach is sound and well-precedented. A couple of points worth noting:                                                                                                     

1. Upgrade path: For users upgrading from the hook-based chart, the CRD already exists (created by `kubectl apply --server-side`). Has the upgrade path been tested to confirm Helm correctly adopts the existing CRD as a managed template resource?
2. Orphaned hook resources: If a user's previous install/upgrade failed, the old hook resources (`ConfigMap, Job, RBAC, SA`) may still exist since `hook-delete-policy: hook-succeeded` only cleans up on success. Might be worth a note in the upgrade docs.
3. SSA trade-off: The old Job used `kubectl apply --server-side --force-conflicts`. The new approach uses Helm's standard 3-way merge, which is fine for the common case (single manager). Worth noting as an intentional trade-off.

Overall LGTM.

jabellard:
> The approach is sound and well-precedented. A couple of points worth noting:
> 
> 1. Upgrade path: For users upgrading from the hook-based chart, the CRD already exists (created by `kubectl apply --server-side`). Has the upgrade path been tested to confirm Helm correctly adopts the existing CRD as a managed template resource?
> 2. Orphaned hook resources: If a user's previous install/upgrade failed, the old hook resources (`ConfigMap, Job, RBAC, SA`) may still exist since `hook-delete-policy: hook-succeeded` only cleans up on success. Might be worth a note in the upgrade docs.
> 3. SSA trade-off: The old Job used `kubectl apply --server-side --force-conflicts`. The new approach uses Helm's standard 3-way merge, which is fine for the common case (single manager). Worth noting as an intentional trade-off.
> 
> Overall LGTM.

I did some E2E tests.

# E2E Tests

Upgrading chart using `release-1.15` branch (old approach)

```bash
helm upgrade --install karmada-operator charts/karmada-operator --namespace karmada-system --create-namespace
```

Get CRD metadata after successful upgrade:

```bash
kubectl  get crd karmadas.operator.karmada.io -oyaml | yq '.metadata' 
```

Output:
```yaml
annotations:
  controller-gen.kubebuilder.io/version: v0.18.0
creationTimestamp: "2026-05-03T16:45:41Z"
generation: 1
name: karmadas.operator.karmada.io
resourceVersion: "707"
uid: 1621d8e0-defe-47f1-8488-3209e659558f
```
CRD was installed using the job, so it does not have any of the Helm annotations or labels.

Upgrading chart using this branch (new approach):

```bash
helm upgrade --install karmada-operator charts/karmada-operator --namespace karmada-system --create-namespace 
```

Output
```text
Error: UPGRADE FAILED: Unable to continue with update: CustomResourceDefinition "karmadas.operator.karmada.io" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: missing key "app.kubernetes.io/managed-by": must be set to "Helm"; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "karmada-operator"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "karmada-system"
```

By default, Helm will not adopt the existing resource.

# Adopting Existing CRDs into Helm

When a CRD like `karmadas.operator.karmada.io` was created manually (outside of Helm), Helm will refuse to manage it because it lacks the required ownership metadata. Below are instructions for resolving this in both Helm 3 and Helm 4.

## Helm 3

Manually add the required labels and annotations so Helm can adopt the resource:

```bash
kubectl label crd karmadas.operator.karmada.io app.kubernetes.io/managed-by=Helm
kubectl annotate crd karmadas.operator.karmada.io meta.helm.sh/release-name=karmada-operator
kubectl annotate crd karmadas.operator.karmada.io meta.helm.sh/release-namespace=karmada-system
```

Then run the upgrade:

```bash
helm upgrade --install karmada-operator charts/karmada-operator --namespace karmada-system --create-namespace
```

After successful upgrade:
```bash
kubectl  get crd karmadas.operator.karmada.io -oyaml | yq '.metadata'   
```

Output:
```yaml
annotations:
  controller-gen.kubebuilder.io/version: v0.20.0
  helm.sh/resource-policy: keep
  meta.helm.sh/release-name: karmada-operator
  meta.helm.sh/release-namespace: karmada-system
creationTimestamp: "2026-05-03T16:45:41Z"
generation: 2
labels:
  app.kubernetes.io/managed-by: Helm
name: karmadas.operator.karmada.io
resourceVersion: "1756"
uid: 1621d8e0-defe-47f1-8488-3209e659558f
```

## Helm 4

Use the `--take-ownership` flag to have Helm skip the ownership annotation check and adopt the existing resource automatically:

```bash
helm upgrade --install karmada-operator charts/karmada-operator --namespace karmada-system --create-namespace --take-ownership
```

@seanlaii thanks for reviewing and making an excellent point regarding a required upgrade path for the users. Please check the E2E test results and the upgrade path above.

@RainbowMango , @zhzhuang-zju , what do you think is the best way to let users know about this? Release notes?

zhzhuang-zju:
@jabellard @seanlaii Thanks for the solution and the insightful discussion!

> what do you think is the best way to let users know about this? Release notes?

This is a common challenge we face when making significant changes. Here is how we generally handle it:

We first evaluate whether the upgrade can be automated via code or configuration.
*   **If yes:** The user doesn't need to do anything, and a standard release note is sufficient.
*   **If no:** We document the change and provide a workaround in the **Urgent Update Notes** section to ensure users are aware of it.

So, regarding this issue, since a fully automated upgrade isn't feasible, we will proceed by documenting it in the **Urgent Update Notes**.

**PS:** I actually attempted to solve this by adding a `pre-upgrade` job. However, after digging into the Helm source code, I found that it's not a viable solution. This is because Helm's ownership check for `existingResources` happens *before* the `pre-upgrade` hooks are executed during an upgrade.
https://github.com/helm/helm/blob/main/pkg/action/upgrade.go#L346

zhzhuang-zju:
> Orphaned hook resources: If a user's previous install/upgrade failed, the old hook resources (ConfigMap, Job, RBAC, SA) may still exist since hook-delete-policy: hook-succeeded only cleans up on success. Might be worth a note in the upgrade docs.

Good point! Our upgrade docs should also include instructions for users to identify and clean up these resources to avoid them being left behind permanently.

hack/update-crdgen.sh
```diff
@@ -35,5 +38,33 @@ controller-gen crd paths=./pkg/apis/remedy/... output:crd:dir=./charts/karmada/_
 controller-gen crd paths=./pkg/apis/work/... output:crd:dir=./charts/karmada/_crds/bases/work
 controller-gen crd paths=./pkg/apis/apps/... output:crd:dir=./charts/karmada/_crds/bases/apps
 controller-gen crd:generateEmbeddedObjectMeta=true paths=./examples/customresourceinterpreter/apis/... output:crd:dir=./examples/customresourceinterpreter/apis/
-controller-gen crd:generateEmbeddedObjectMeta=true paths=./operator/pkg/apis/operator/... output:crd:dir=./charts/karmada-operator/crds
+# Generate operator CRDs to a temp directory, then post-process into a Helm template.
+# This follows the cert-manager pattern: the generated metadata is discarded and replaced
+# with a Helm-controlled metadata block, keeping only the spec from controller-gen.
```
RainbowMango: By the way, the `generated metadata` is not discarded, echo [the content](https://github.com/karmada-io/karmada/blob/e669d0d8cb084523e9fd6ba552e3710c9c28fee6/charts/karmada-operator/crds/operator.karmada.io_karmadas.yaml#L4-L7) here
```yaml
metadata:
  annotations:
    controller-gen.kubebuilder.io/version: v0.20.0
  name: karmadas.operator.karmada.io
```

RainbowMango:
> what do you think is the best way to let users know about this? Release notes?

Basically, I agree with @zhzhuang-zju mentioned above. We can document it in the upgrade notes section in the release notes. (See [an example here](https://github.com/karmada-io/karmada/blob/master/docs/CHANGELOG/CHANGELOG-1.13.md#urgent-update-notes))

> So, regarding this issue, since a fully automated upgrade isn't feasible, 

Maybe there is a way that the user could be unaware of this change:
- in v1.17.x, we reduce the size of the CRD and add the required labels and annotations to it. 
- in v1.18, move CRD to templates

Do you think it's worth doing this?

zhzhuang-zju:
> Maybe there is a way that the user could be unaware of this change:
in v1.17.x, we reduce the size of the CRD and add the required labels and annotations to it.
in v1.18, move CRD to templates

I agree with this approach.

First, I have verified this upgrade path through testing; it is fully viable and can be performed transparently without user impact.

Second, as I mentioned in https://github.com/karmada-io/karmada/issues/7371#issuecomment-4385913719, the release-1.17 branch is not suitable for major mechanical changes. Therefore, a straightforward fix focused on reducing the CRD size is much more appropriate for this release. Adding the labels and annotations to facilitate a smooth upgrade makes this even better.

jabellard:
@zhzhuang-zju , @RainbowMango , @seanlaii 

Thanks. I like the idea of making this as seamless as possible for the users. We definitely want to make sure this is a great experience for them. So he's what I'll do based on your constructive feedback:
- Push PR that maintains the old approach but with the CRD size reduced to fit into the config map and with labels and annotations added as needed
- Backport that change to release 1.17
- Proceed with this change, which will be part of release 1.18. To make this a smooth transition for users, in the release notes, we'll have to let them know that before upgrading to release 1.18 that they will first need to upgrade to release 1.17.latest-patch-version

RainbowMango:
> Push PR that maintains the old approach but with the CRD size reduced to fit into the config map and with labels and annotations added as needed
Backport that change to release 1.17

Sounds good to me. Another approach also works if you don't want to revert anything from the master branch; that is, send the PR against the branch release-1.17 (not backporting) to fix the upgrading issue.

hack/update-crdgen.sh
jabellard: Fixed.

zhzhuang-zju:
Thanks @jabellard, I have no other comments besides filling out the release note section.
/lgtm

cc @seanlaii @RainbowMango to take another look

jabellard:
> Thanks @jabellard, I have no other comments besides filling out the release note section. /lgtm
> 
> cc @seanlaii @RainbowMango to take another look

I just added the note.

RainbowMango (approved):
/approve

PR #7460, opened 2026-05-02
fix(dependenciesdistributor): preserve Replicas and SchedulerName in attached bindings

### LINKED ISSUE #5929: rb created by ResourceDetector is overwriten by DependenciesDistributor
**What happened**:
When using Karmada, I discovered that in certain scenarios, the `ResourceBinding` created by the `ResourceDetector` has its replicas modified when updated by the `DependenciesDistributor`. The specific process is as follows:

1. First, create a PropagationPolicy with `propagateDeps: true`.
2. Create a Deployment with `replicas=0` and add a `persistentVolumeClaim` volume.
3. Update the annotations of the Deployment.
4. Update `replicas=5`.

Upon checking the logs, I found that the `spec.replicas` of the `ResourceBinding` changed in the sequence 0->5->0. Ultimately, this value was inconsistent with the Deployment.
I reviewed the relevant code and found that when the `DependenciesDistributor` updates the `ResourceBinding`, it does not set the replicas parameter for the `ResourceBinding`:

```go
// dependencies_distributor.go->buildAttachedBinding
    Resource: workv1alpha2.ObjectReference{
        APIVersion:      object.GetAPIVersion(),
        Kind:            object.GetKind(),
        Namespace:       object.GetNamespace(),
        Name:            object.GetName(),
        ResourceVersion: object.GetResourceVersion(),
    },
    RequiredBy: result,
},
```

I would like to confirm if this could be the cause of the phenomenon described above?

**Environment**:
- Karmada version: 1.9

manmathbh:
**What type of PR is this?**
/kind bug

When `propagateDeps: true` is set on a PropagationPolicy, the DependenciesDistributor creates attached ResourceBindings for dependent resources (ConfigMaps, Secrets, PVCs). However, `buildAttachedBinding` did not carry over the `Replicas` and `SchedulerName` fields from the independent binding. This caused these values to be reset during the reconcile cycle — ResourceDetector would correctly set replicas, then DependenciesDistributor would overwrite the binding spec with zeroed-out fields.

**Root cause**: `buildAttachedBinding` constructs a new `ResourceBindingSpec` but only copied `ConflictResolution` and `PreserveResourcesOnDeletion`, omitting `Replicas` and `SchedulerName`.

**Fix**:
- `buildAttachedBinding`: copy `Replicas` and `SchedulerName` from the independent binding
- `createOrUpdateAttachedBinding`: preserve `Replicas` and `SchedulerName` during updates

Fixes #5929

Tests updated in `Test_buildAttachedBinding` to verify `Replicas` and `SchedulerName` are carried over (72/72 tests pass).

```release-note
`karmada-controller-manager`: Fixed an issue where the DependenciesDistributor would reset the `Replicas` field on ResourceBindings when `propagateDeps` is enabled, causing replica counts to be lost during dependency propagation.
```

### DIFF
--- a/pkg/dependenciesdistributor/dependencies_distributor.go
+++ b/pkg/dependenciesdistributor/dependencies_distributor.go
@@ -633,6 +633,8 @@ func (d *DependenciesDistributor) createOrUpdateAttachedBinding(ctx context.Cont
 				// Update with effective results.
 				existBinding.Spec.ConflictResolution = effectiveCR
 				existBinding.Spec.PreserveResourcesOnDeletion = ptr.To(effectivePreserve)
+				existBinding.Spec.Replicas = attachedBinding.Spec.Replicas
+				existBinding.Spec.SchedulerName = attachedBinding.Spec.SchedulerName
 			}
 
 			existBinding.Spec.RequiredBy = mergedRequiredBy
@@ -782,8 +784,10 @@ func buildAttachedBinding(independentBinding *workv1alpha2.ResourceBinding, obje
 				ResourceVersion: object.GetResourceVersion(),
 			},
 			RequiredBy:                  result,
+			Replicas:                    independentBinding.Spec.Replicas,
 			PreserveResourcesOnDeletion: independentBinding.Spec.PreserveResourcesOnDeletion,
 			ConflictResolution:          independentBinding.Spec.ConflictResolution,
+			SchedulerName:               independentBinding.Spec.SchedulerName,
 		},
 	}
 }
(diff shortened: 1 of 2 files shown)


XiShanYongYe-Chang:
Hi @manmathbh, can you reproduce the issue in #5929?

manmathbh:
> Hi @manmathbh, can you reproduce the issue in #5929?

@XiShanYongYe-Chang  The issue reporter traced replicas going 0→5→0 in the binding spec and identified the missing field in `buildAttachedBinding`. The code gap is clear: `ConflictResolution` and `PreserveResourcesOnDeletion` are carried over from the independent binding, but `Replicas` and `SchedulerName` are not. Whether or not the exact 0→5→0 sequence is reproducible in every setup, the gap is real and the fix makes the function consistent.
I can add a targeted unit test demonstrating the fallback-to-zero behavior if that would help.

XiShanYongYe-Chang:
I understand that it's not like this. The reason why replicas are not set in the follow distribution is that for the resources being followed, it is not a workload resource, so it does not have a replicas field. Secondly, this was an issue that occasionally occurred in previous versions. After so many iterations of versions, I want to know whether this issue still exists. It might have already been fixed, which is why I asked you whether it can be reproduced in the environment. 

The test environment I refer to is not a ut test, but rather deploying a Karmada instance to perform functional testing.

manmathbh:
Thanks @XiShanYongYe-Chang — you're right that attached bindings are for non-workload resources and don't carry replicas. The original issue may have been resolved in a later version. I'll close this and re-investigate whether the issue described in #5929 is still reproducible on current master.

PR #7462, opened 2026-05-02
fix(scheduler): prevent overscheduling by accounting for in-flight assignments

### LINKED ISSUE #6783: Stale Resource Assessment in karmada-scheduler-estimator Causes Incorrect Placement During Concurrent Aggregated Scheduling

manmathbh:
**What type of PR is this?**
/kind bug

When two workloads with Aggregated scheduling are created concurrently and target the same cluster, the karmada-scheduler-estimator's resource snapshot is stale — it does not reflect resources consumed by the first workload when evaluating the second. This causes both to be placed on the same cluster, leading to resource starvation.

Karmada already has an `AssigningResourceBindingCache` (analogous to Kubernetes' `AssumePod`) that tracks bindings that have been scheduled but not yet applied on member clusters. However:
1. The cache was only populated when the `WorkloadAffinity` feature gate was enabled
2. The cached bindings were never consulted during replica division

**Fix** (3 changes):
1. **Unconditionally track in-flight bindings**: Remove `WorkloadAffinity` feature gate guard on `AssigningResourceBindings.Add()` so all scheduled bindings are tracked in the assigning cache.
2. **Wire assigning bindings into replica division**: Pass assigning bindings from the scheduler cache through `AssignReplicas` → `assignWorkloadReplicas` → `assignReplicasToClusters` → `newAssignState` → `buildAvailableClusters`.
3. **Deduct in-flight replicas**: In `buildAvailableClusters`, iterate through assigning bindings and subtract replicas already allocated to each cluster from available capacity. This prevents a second concurrent scheduling decision from claiming the same resources.

Fixes #6783

- All 450 scheduler tests pass
- Nil-safety added for `schedulerCache` in test and edge-case paths
- The `AssigningResourceBindingCache.OnBindingUpdate` automatically clears entries when informer catches up, preventing memory leaks

```release-note
`karmada-scheduler`: Fixed an issue where concurrent scheduling of workloads with Aggregated replica division could cause resource oversubscription by accounting for in-flight assignments during replica calculation.
```

### DIFF
--- a/pkg/scheduler/core/assignment.go
+++ b/pkg/scheduler/core/assignment.go
@@ -72,6 +72,11 @@ type assignState struct {
 	strategy   *policyv1alpha1.ReplicaSchedulingStrategy
 	spec       *workv1alpha2.ResourceBindingSpec
 
+	// assigningBindings holds bindings that have been scheduled but not yet
+	// reflected in member clusters. Their allocated replicas are deducted from
+	// available capacity to prevent overscheduling during concurrent placement.
+	assigningBindings map[string]*workv1alpha2.ResourceBinding
+
 	// fields below are indirect results
 	strategyType string
 
@@ -92,7 +97,7 @@ type assignState struct {
 	targetReplicas int32
 }
 
-func newAssignState(candidates []spreadconstraint.ClusterDetailInfo, spec *workv1alpha2.ResourceBindingSpec, status *workv1alpha2.ResourceBindingStatus) *assignState {
+func newAssignState(candidates []spreadconstraint.ClusterDetailInfo, spec *workv1alpha2.ResourceBindingSpec, status *workv1alpha2.ResourceBindingStatus, assigningBindings map[string]*workv1alpha2.ResourceBinding) *assignState {
 	var strategyType string
 
 	switch spec.Placement.ReplicaSchedulingType() {
@@ -119,7 +124,7 @@ func newAssignState(candidates []spreadconstraint.ClusterDetailInfo, spec *workv
 		expectAssignmentMode = Fresh
 	}
 
-	return &assignState{candidates: candidates, strategy: spec.Placement.ReplicaScheduling, spec: spec, strategyType: strategyType, assignmentMode: expectAssignmentMode}
+	return &assignState{candidates: candidates, strategy: spec.Placement.ReplicaScheduling, spec: spec, strategyType: strategyType, assignmentMode: expectAssignmentMode, assigningBindings: assigningBindings}
 }
 
 func (as *assignState) buildScheduledClusters() {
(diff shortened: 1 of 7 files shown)


XiShanYongYe-Chang:
Hi @manmathbh 

It's not a simple fix, and I've actually submitted a design document to deal with it: https://github.com/karmada-io/karmada/pull/7416

You can check whether similar work is being processed before you make any further contributions. And then can submit a PR.

manmathbh:
Thanks @XiShanYongYe-Chang — I'll close this and follow #7416. Happy to help with the implementation once the design is approved.

PR #7467, opened 2026-05-03
test(util/lifted): cover resource-name helpers

apshada:
Adds unit tests for the previously-uncovered helpers in `pkg/util/lifted/corev1helpers.go` and `resourcename.go`. Pure test additions, no production-code changes.


XiShanYongYe-Chang:
Hi @apshada thanks for your contribution.
The code in this folder was lifted from the upstream source, and there will be comments on the functions. Is this test consistent with the upstream?

PR #7469, opened 2026-05-03
test(e2e): add inline event assertions for binding object events

manmathbh:
**What type of PR is this?**
/kind feature

Part of #7252 — improving E2E test coverage for Karmada controller events.

Adds inline event assertions for two binding object events that were not
yet covered:

1. **`SyncScheduleResultToDependenciesSucceed`** — asserted in
   `dependenciesdistributor_test.go` after the configmap is auto-propagated
   with `propagateDeps: true`. This event is emitted by the
   DependenciesDistributor when it successfully syncs schedule results to
   dependent resources.

2. **`EvictWorkloadFromClusterSucceed`** — asserted in
   `workloadrebalancer_test.go` after graceful eviction tasks complete on
   the tainted cluster. This event is emitted by the graceful eviction
   controller after successfully evicting workloads from a failing cluster.

Both assertions follow the community meeting decision to use **Approach 2
(inline event assertions)** — placing assertions immediately after the
operation that triggers the event for semantic accuracy.

Part of #7252

cc @XiShanYongYe-Chang

This addresses two remaining binding object events from the
[subtask breakdown](https://github.com/karmada-io/karmada/issues/7252#issuecomment-2726196632).

```release-note
NONE
```

### DIFF
--- a/test/e2e/suites/base/dependenciesdistributor_test.go
+++ b/test/e2e/suites/base/dependenciesdistributor_test.go
@@ -119,6 +119,13 @@ var _ = ginkgo.Describe("[DependenciesDistributor] automatically propagate relev
 							return event.Reason == events.EventReasonGetDependenciesSucceed
 						})
 				})
+				ginkgo.By("checking SyncScheduleResultToDependenciesSucceed event on binding", func() {
+					bindingName := names.GenerateBindingName("Deployment", deployment.Name)
+					framework.WaitEventFitWith(kubeClient, deployment.Namespace, bindingName,
+						func(event corev1.Event) bool {
+							return event.Reason == events.EventReasonSyncScheduleResultToDependenciesSucceed
+						})
+				})
 
 				ginkgo.By("updating propagation policy's clusterNames", func() {
 					patch := []map[string]any{
@@ -219,6 +226,13 @@ var _ = ginkgo.Describe("[DependenciesDistributor] automatically propagate relev
 								return event.Reason == events.EventReasonGetDependenciesSucceed
 							})
 					})
+					ginkgo.By("checking SyncScheduleResultToDependenciesSucceed event on binding", func() {
+						bindingName := names.GenerateBindingName("Deployment", deployment.Name)
+						framework.WaitEventFitWith(kubeClient, deployment.Namespace, bindingName,
+							func(event corev1.Event) bool {
+								return event.Reason == events.EventReasonSyncScheduleResultToDependenciesSucceed
+							})
+					})
 				})
 
 				ginkgo.By("make the secret is not referenced by the deployment ", func() {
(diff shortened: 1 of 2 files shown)


test/e2e/suites/base/dependenciesdistributor_test.go
```diff
@@ -219,6 +219,13 @@ var _ = ginkgo.Describe("[DependenciesDistributor] automatically propagate relev
 								return event.Reason == events.EventReasonGetDependenciesSucceed
 							})
 					})
+					ginkgo.By("checking SyncScheduleResultToDependenciesSucceed event on binding", func() {
```
XiShanYongYe-Chang, line 229: Is a case randomly selected for the event test?

test/e2e/suites/base/workloadrebalancer_test.go
```diff
@@ -154,6 +155,13 @@ var _ = framework.SerialDescribe("workload rebalancer testing", func() {
 				framework.AssertBindingScheduledClusters(karmadaClient, namespace, deployBindingName, [][]string{targetClusters[1:]})
 				framework.WaitGracefulEvictionTasksDone(karmadaClient, namespace, deployBindingName)
 
+				ginkgo.By("checking EvictWorkloadFromClusterSucceed event on binding", func() {
```
XiShanYongYe-Chang: The main purpose of the current test file is not to test evictions. I suggest you handle it in the `failover_test.go` file.

XiShanYongYe-Chang (commented):
Thanks

test/e2e/suites/base/dependenciesdistributor_test.go
manmathbh: Thanks for the review @XiShanYongYe-Chang!
The `SyncScheduleResultToDependenciesSucceed` assertion is not random — it's in the 'configmap automatically propagation testing' test case because that's the only test where propagateDeps: true triggers the DependenciesDistributor to sync schedule results to dependencies. The event naturally fires as part of this flow.

test/e2e/suites/base/workloadrebalancer_test.go
manmathbh: Done @XiShanYongYe-Chang, moved the `EvictWorkloadFromClusterSucceed` assertion to `failover_test.go` in the 'purgeMode gracefully' context, right after the rescheduling is confirmed. The `workloadrebalancer_test.go` event has been removed

test/e2e/suites/base/dependenciesdistributor_test.go
XiShanYongYe-Chang: In the current file, other cases are also tested and distributed together.

manmathbh:
PTAL @XiShanYongYe-Chang, made changes based on your comments

test/e2e/suites/base/dependenciesdistributor_test.go
XiShanYongYe-Chang: Wouldn't it be better to check the other cases as well?

manmathbh:
@XiShanYongYe-Chang good catch, added `SyncScheduleResultToDependenciesSucceed` to the remaining propagation cases (secret, PVC, serviceAccount, and the second configmap case). 
All 5 propagation test cases now assert the event. Build and vet clean. PTAL.

XiShanYongYe-Chang:
Okay
/retest

XiShanYongYe-Chang (commented):
Thanks @manmathbh 
Sorry for the delay.
/lgtm
/approve

PR #7473, opened 2026-05-04
fix: remove incorrect PropagateDeps gate from application failover

### LINKED ISSUE #7458: [bug]: Application failover silently disabled when propagateDependencies is false
## What happened:

Application failover is silently disabled for namespaced ResourceBindings when `propagateDependencies` is not set to `true`, even when `failover.application` is fully configured. The workload stays on unhealthy clusters indefinitely with no error or warning.

Application failover should work whenever `spec.failover.application` is configured, regardless of the `propagateDependencies` setting. These are independent features - one controls dependency propagation (ConfigMaps/Secrets), the other controls health-based migration.

---

## How to reproduce it (as minimally and precisely as possible):

1. Create a Deployment with PropagationPolicy:
```yaml
spec:
  placement:
    clusterAffinity:
      clusterNames: [cluster-a, cluster-b]
  failover:
    application:
      decisionConditions:
        tolerationSeconds: 30
      purgeMode: Graciously
  # propagateDependencies NOT set (defaults to false)
```

2. Confirm workload is healthy on both clusters
3. Break the workload on cluster-a (set bad image, crash pods, etc.) - verify `AggregatedStatus` shows `Health: Unhealthy`
4. Wait more than 30 seconds
5. Check `kubectl get rb <name> -o yaml` - no `gracefulEvictionTasks` created, workload stays on unhealthy cluster-a
6. Add `propagateDependencies: true` to the policy - failover immediately starts working

---

## Anything else we need to know?:

**Root cause:** The controller's filter function incorrectly couples two unrelated features.

In `pkg/controllers/applicationfailover/rb_application_failover_controller.go` lines 239-241:
```go
if !rb.Spec.PropagateDeps {
    return false  // Blocks entire failover pipeline
}
```

This check causes `bindingFilter()` to return `false`, which makes the controller skip all failover logic (no unhealthy tracking, no eviction tasks, nothing).

**Proof it's a bug:** The cluster-scoped controller (`crb_application_failover_controller.go`) does NOT have this check - ClusterResourceBindings get failover without needing `PropagateDeps: true`. This proves the RB controller's gate is unintended divergence.

**Verified with test:**
Created a unit test that confirms the bug:
```go
// Test case: failover.application configured, health interpreter present, 
// PropagateDeps = false
rb := &workv1alpha2.ResourceBinding{
        Failover: &policyv1alpha1.FailoverBehavior{
            Application: &policyv1alpha1.ApplicationFailoverBehavior{...},
        },
        PropagateDeps: false,  // Default value
    },
}

result := controller.bindingFilter(rb)
// Expected: true (failover should activate)
// Actual: false (failover silently disabled)
```

Test confirms:
- ✅ `bindingFilter()` returns `false` when `PropagateDeps: false`
- ✅ CRB controller's filter does NOT have this gate
- ✅ Bug affects default policy configuration

---

## Impact:
- Silent failure - no error, no event, no log at any verbosity
- Affects default case - `propagateDependencies` defaults to `false`
- Production chaos drills fail - users think they're protected but aren't
- AI-generated policies would have failover configured but non-functional

**Simple fix:** Delete those 3 lines. PropagateDeps controls dependency distribution, not failover eligibility.

**Connection to agent-skills project:**
This perfectly illustrates why we need `karmada-create-policy` and `karmada-audit-policy` skills - there's currently no way for AI agents to know about this undocumented coupling. They'd generate policies with failover configured but silently broken.

---

## Environment:
- Karmada version: master (affects all versions with application failover)
- Others: Only affects namespaced ResourceBindings, ClusterResourceBindings work correctly

Ady0333:
## What type of PR is this?
/kind bug

---

Fixes #7458 - Removes incorrect coupling between `propagateDependencies` and 
application failover that was silently disabling failover for the default 
policy configuration.

**Root Cause:**
`bindingFilter()` in the namespaced controller incorrectly required 
`rb.Spec.PropagateDeps: true` to enable application failover, coupling two 
independent features:
- `propagateDependencies` - controls dependency distribution (ConfigMaps/Secrets)
- `failover.application` - controls health-based workload migration

**The Bug:**
```go
// Lines 239-241 (REMOVED)
if !rb.Spec.PropagateDeps {
    return false  // Silently disabled failover
}
```

This caused application failover to be completely inactive for namespaced 
ResourceBindings when `propagateDependencies` was not set (the default), 
even when `failover.application` was fully configured.

**Proof of Bug:**
The cluster-scoped controller (`crb_application_failover_controller.go`) does 
NOT have this check - ClusterResourceBindings get failover without needing 
`PropagateDeps: true`. This proves the RB controller's gate was unintended 
divergence, not deliberate design.

---

## Impact

**Before fix:**
- Silent failure (no error, event, or log)
- Affects default case (`propagateDependencies` defaults to `false`)
- Production chaos drills fail unexpectedly
- Users believe failover is configured but it doesn't work

**After fix:**
- Application failover works when configured, regardless of dependency propagation
- Matches CRB controller behavior (feature parity)
- Restores documented API semantics

---

## Testing

**New test:**
`TestRBApplicationFailoverController_bindingFilter_propagateDepsIndependent`

Creates a ResourceBinding with:
- `failover.application` configured
- `propagateDeps: false` (default value)
- Health interpreter available (Deployment)

Verifies `bindingFilter()` returns `true` (failover should activate).

This test would have failed before the fix, proving the bug existed.

**All tests passing:** ✅

---

```release-note
Fixed critical bug where application failover was silently disabled for 
namespaced ResourceBindings when propagateDependencies was not set to true
```

### DIFF
--- a/pkg/controllers/applicationfailover/rb_application_failover_controller.go
+++ b/pkg/controllers/applicationfailover/rb_application_failover_controller.go
@@ -236,8 +236,5 @@ func (c *RBApplicationFailoverController) bindingFilter(rb *workv1alpha2.Resourc
 		return false
 	}
 
-	if !rb.Spec.PropagateDeps {
-		return false
-	}
 	return true
 }


zhzhuang-zju:
Thanks @Ady0333. That's a good question. I looked through the previous PR records and found that this was discussed when the controller was first implemented.

https://github.com/karmada-io/karmada/pull/3461#discussion_r1184691442

In my opinion, it makes sense for application failover to have the propagateDependencies restriction. The logic is that if an application-level failover occurs, migrating the app to a new cluster while its dependencies remain in the old one would still cause issues for the application.

zhzhuang-zju:
Also found relevant information in the official documentation.

https://karmada.io/docs/next/userguide/failover/application-failover/#how-do-i-enable-the-feature

Ady0333:
Thank you @zhzhuang-zju for the clarification and the references!

I understand now - the PropagateDeps requirement is intentional to ensure dependencies migrate with the application. Looking at PR #3461, I can see this was a deliberate design decision to prevent the scenario where an app migrates but its dependencies (ConfigMaps, Secrets, etc.) remain in the old cluster, breaking the application. That makes sense.

My confusion came from the fact that when `propagateDeps: false`, the controller silently skips the binding in `bindingFilter()` with no indication to the user why failover isn't working. 

Would it make sense to add a Warning event when this happens? Something like:

```go
if !rb.Spec.PropagateDeps && hasFailoverConfigured {
    recorder.Event(rb, corev1.EventTypeWarning, 
        "FailoverSkipped", 
        "Application failover requires propagateDeps: true")
    return false
}
```

This way users would get a clear signal that their failover config is being ignored due to the PropagateDeps requirement, rather than having to debug why nothing is happening.

What do you think?

zhzhuang-zju:
Adding a relevant signal is a good idea, but I'm not sure about the format—`event` or just a log?
Additionally, I have another thought: is it really necessary to be this strict? If a resource disables the `InterpretDependency` operation, maybe we don't need to enforce restrictions on its `propagateDeps` configuration.

@Ady0333, I recommend not modifying the code until we clarify these points. Perhaps we could discuss this in the community meeting, if you're available.

Ady0333:
Thanks @zhzhuang-zju! I'd be happy to discuss this in the community meeting.

For the format - I was thinking an **Event** would be better than just a log because:
- Events are visible in `kubectl describe` - easier for users to discover
- They show up in monitoring/alerting systems
- Logs are often not checked unless something is already broken

For your question about the restriction - that's a great point. If a resource disables `InterpretDependency`, then maybe `propagateDeps` doesn't need to be enforced for that specific resource. The current blanket restriction might be overly strict.
I'll join to discuss this properly. In the meantime, I'll hold off on making any code changes until we align on the approach.

zhzhuang-zju:
> Thanks @zhzhuang-zju! I'd be happy to discuss this in the community meeting.

Great! Since this is not a bug, let's remove the `bug` label first.

/remove-kind bug

Ady0333:
> > Thanks @zhzhuang-zju! I'd be happy to discuss this in the community meeting.
> 
> Great! Since this is not a bug, let's remove the `bug` label first.
> 
> /remove-kind bug

You are absolutely right!!!

XiShanYongYe-Chang:
I found that similar issues may have been discussed twice, ref: https://github.com/karmada-io/karmada/issues/7474#issuecomment-4386469692

I also saw your discussion above. In fact, from the perspective of API design, this is expected. And the document also mentions this point.

Ady0333:
> I found that similar issues may have been discussed twice, ref: [#7474 (comment)](https://github.com/karmada-io/karmada/issues/7474#issuecomment-4386469692)
> 
> I also saw your discussion above. In fact, from the perspective of API design, this is expected. And the document also mentions this point.

Thank you @XiShanYongYe-Chang for the clarification.
You're right - the API comments (lines 331-336) clearly document this behavior. I should have read the documentation more carefully before filing the issue.
I'll close this since it's working as designed.

XiShanYongYe-Chang:
I'm sorry. I didn't mean to blame. Your findings are positive, and our documentation may be relatively simple, and it would be interesting to further supplement the documentation based on the behavior of the current code repository features.

Ady0333:
Thank you for the kind words, @XiShanYongYe-Chang!

Hearing "my contribtuions are positive" from you keeps me motivated for doing more good and perfect work in Karmada community 😄. You're right - the documentation is clear. This was my oversight in not reading it carefully enough before filing the issue. Thank you for taking the time to clarify.

XiShanYongYe-Chang:
Hi @Ady0333, do we still need this PR?

Ady0333:
No! I don't need this pr for now. I'll close this.

Issue #6926, opened 2025-11-14
Create a Best Practice Guide for Running Flink on Multi-Cluster with Karmada


RainbowMango:
**Description**

Recently, we've observed growing community interest and multiple user contributions around running `Apache Flink` on Karmada-managed multi-cluster environments. Several users have shared their experiences through talks, demos, and informal discussions. It's timely and valuable to consolidate this knowledge into an official `best practice guide`.

This initiative aims to:
- Provide clear, tested guidance for deploying and operating Flink workloads across multiple clusters using Karmada.
- Strengthen collaboration between the Karmada and Flink communities.
- Highlight Karmada’s capabilities in real-world stateful, distributed data processing scenarios.

**Proposed Actions**

1. Collect and showcase existing community content
   - Collect and archive relevant video recordings, blog posts, and presentations shared by users (e.g., KubeCon talks, meetups).
   - Add them to the Karmada community media or resource section.

2. Develop and publish a best practice document
   - Create a comprehensive guide covering architecture, deployment patterns, failover strategies, and performance considerations.
   - Publish it on the [Karmada official website](https://karmada.io).

3. `Cross-post to broader ecosystems`
   - Submit the best practice to the **CNCF blog** to reach a wider cloud-native audience.
   - Share the guide with the **Apache Flink community** (e.g., via Flink blog or mailing list) to encourage adoption and feedback.

4. `Identify gaps and plan enhancements`
   - Analyze current limitations in Karmada’s support for Flink (e.g., scheduling of stateful jobs, network topology awareness, observability integration).
   - Propose and track necessary feature enhancements or documentation improvements via follow-up issues.

## Next Steps

- [ ] Assign issue owner(s)
- [ ] Gather existing materials from community members
- [ ] Draft initial best practice outline
- [ ] Engage with the Flink community for technical review
- [ ] Schedule publication across target channels

We welcome input from both Karmada and Flink contributors on content scope, technical accuracy, and collaboration opportunities.

**Call for Participation**
We welcome contributions from community members, especially those with experience in:
- Flink deployment and operations.
- Karmada policy design and multi-cluster management.
- Technical writing and documentation.

Please comment below if you’re interested in contributing or have relevant resources to share!

PR #7478, opened 2026-05-06
Add comment/example for how to use retention for flink stateful failover

dahuo98:
**What type of PR is this?**
/kind documentation

This PR adds a section of comment in retention section of flink's resource interpreter. The comment shows what needs to be done if one wants to start flink app from latest checkpoint/savepoint during failover event. 

```release-note

```

### DIFF
--- a/pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/flink.apache.org/v1beta1/FlinkDeployment/customizations.yaml
+++ b/pkg/resourceinterpreter/default/thirdparty/resourcecustomizations/flink.apache.org/v1beta1/FlinkDeployment/customizations.yaml
@@ -335,3 +335,24 @@ spec:
           status.taskManager = observedObj.status.taskManager
           return status
         end
+    retention:
+      luaScript: |
+        -- For failing over flink applications, if initialSavepointPath is injected in member clusters, we will need to 
+        -- retain it to avoid being overwritten by karmada control plane
+        -- For example, the following script retains initialSavepointPath if resourcebinding.karmada.io/failover-jobid
+        -- label present, this label is set by defining StatePreservation in propagation policy 
+        ---------------------------- Code Example ----------------------------------------------------------------------
+        -- function Retain(desiredObj, observedObj)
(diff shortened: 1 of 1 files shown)


zhzhuang-zju:
Thanks! I'm wondering if we could document this scenario on the website instead of using code comments. For example, we could add it to a page like this: https://karmada.io/docs/next/userguide/failover/cluster-failover/#example-of-statepreservation-configuration
Also, let's ask @mszacillo to take a look, since he has a strong interest and expertise in using FlinkDeployment.

mszacillo:
@zhzhuang-zju Thanks for the comment! Totally agree, I think it's worth documenting under the statepreservation configuration feature, especially in the FlinkDeployment case. I believe we were planning on adding best-practices documentation specifically for the FlinkDeployment use-case, so it would be helpful to include this information there as well. (cc @RainbowMango )

As for this PR, I'm not opposed to having a simple default retain function that just retains the initialSavepointPath for the application in cases of FlinkDeployment - since it is critical to ensuring graceful recovery. As part of the best-practices for FlinkDeployment, we'd simply ask users to specifically use `resourcebinding.karmada.io/failover-jobid` as the state preservation label to trigger the retention logic.

RainbowMango:
> I believe we were planning on adding best-practices documentation specifically for the FlinkDeployment use-case, so it would be helpful to include this information there as well.

Yes! And the issue #6926 is tracking for this. 

By the way, it would be great to have a case study on cncf.io (Like [the RedNote case study](https://www.cncf.io/case-studies/xiaohongshu/)).   @mszacillo Are you interested in this? You would be the best expert to write this.

RainbowMango:
> As for this PR, I'm not opposed to having a simple default retain function that just retains the initialSavepointPath for the application in cases of FlinkDeployment - since it is critical to ensuring graceful recovery. As part of the best-practices for FlinkDeployment, we'd simply ask users to specifically use resourcebinding.karmada.io/failover-jobid as the state preservation label to trigger the retention logic.

Do you mean the default retention function always retains `.spec.job.initialSavepointPath` no matter `resourcebinding.karmada.io/failover-jobid` label present or not?

My concern is that the label is not reserved, so it is not appropriate to be coded here.

PR #7487, opened 2026-05-08
fix(graceful-eviction): retry on optimistic-lock conflicts when patching eviction tasks

### LINKED ISSUE #7483: [Umbrella] Eviction throughput collapses during cluster-failover storms due to optimistic-lock conflict between taint-manager and graceful-eviction controller
---
name: Bug Report
about: Eviction throughput collapses during cluster-failover storms due to optimistic-lock conflict between taint-manager and graceful-eviction controller
labels: kind/bug

---

### In-flight PRs

The two correctness fixes (eliminate the optimistic-lock conflict storm between the taint manager and the graceful-eviction controllers) plus the operator-facing tunable that lifts the hardcoded worker-pool ceiling.

1. Pure correctness — taint-manager half of the conflict pair.
    - [ ] `taint-manager`: retry on optimistic-lock conflicts when appending eviction tasks. (@RyanCopley, #7493)
2. Pure correctness — graceful-eviction-controller half of the conflict pair. Throughput improvement materializes when both (1) and (2) ship.
    - [ ] `graceful-eviction`: retry on optimistic-lock conflicts when patching eviction tasks. (@RyanCopley, #7494)
3. Operator tunable — removes the hardcoded `ConcurrentReconciles: 3` ceiling. Default bumps from 3 → 5 (matching the rest of the `Concurrent*` family); operators can set it higher to lift the eviction-rate ceiling on large clusters.
    - [ ] `karmada-controller-manager`: introduce `--concurrent-binding-eviction-syncs`. (@RyanCopley, #7495)

### Deferred — needs design discussion first

1. Decouple the eviction rate limiter from the generic `--rate-limiter-qps`. `NewGracefulEvictionRateLimiter` composes a `DynamicRateLimiter` driven by `--resource-eviction-rate` with the generic `DefaultControllerRateLimiter` via `MaxOf`, so the longer delay always wins — `--resource-eviction-rate=50` is silently capped at `--rate-limiter-qps`'s 10 RB/s default. This is a behavior change; happy to open a separate issue for the design discussion before any implementation.
    - [ ] [Design] Decouple `NewGracefulEvictionRateLimiter` from `--rate-limiter-qps`.

Will tick boxes as each PR merges.

**What happened**:

While discussing federation strategies at work, there was a build vs  buy discussion and I took the path of researching the "buy" decision. Karmada seems like a near perfect fit but it is facing elimination due to performance concerns at significant scale. I hope to actually contribute back the testing I did so that we can save ourselves more time internally.

i accomplished this though setting up a sample cluster locally and noticing the performance bottleneck and then using AI to iterate on profiling and resolving the issues.

---

Beginning with the AI generated stuff (sorry)

When a member cluster is tainted `NoExecute` and Karmada needs to redistribute its `ResourceBindings`, eviction throughput is far lower than the per-controller and apiserver capacity should allow. Two `karmada-controller-manager` controllers race to write the same field on the same object:

- `pkg/controllers/cluster/taint_manager.go` — `syncBindingEviction` / `syncClusterBindingEviction` **append** a `GracefulEvictionTask` to `Spec.GracefulEvictionTasks` via `Client.Update` with the cached `metadata.resourceVersion`.
- `pkg/controllers/gracefuleviction/{rb,crb}_graceful_eviction_controller.go` — `syncBinding` **removes** completed/expired tasks from the same field via `client.Patch(MergeFromWithOptimisticLock)`.

When the two land between each other's read and write, the apiserver returns `StatusReasonConflict`. Neither controller wraps its read-modify-write in `RetryOnConflict`, so the only recovery path is the controller-runtime workqueue's exponential backoff (5 ms → 10 ms → … → 1000 s).

Additionally, the `NoExecuteTaintManager`'s eviction worker pool size is **hardcoded to 3** at `cmd/controller-manager/app/controllermanager.go` with no flag to override, which caps eviction parallelism regardless of every other tuning knob (`--resource-eviction-rate`, `--rate-limiter-qps`, `--kube-api-qps`, …).

In a 5 K-binding lab on `release-1.17` (commit `af79ff9`, tag `v1.17.2`):

| Metric | Value |
|---|---|
| Sustained eviction rate at default config | **1.2 ResourceBindings / sec** |
| Time to redistribute a 23 K-RB cluster failure | **~5 hours (extrapolated)** |
| Retry rate on `resource-binding-graceful-eviction-controller` | 99.7 % (1072 / 1075 adds) |
| Retry rate on `binding-eviction` | 20 % |
| Average wall-clock per `binding-eviction` reconcile | 380 ms (mostly workqueue backoff) |
| CPU utilization during the storm | 8.83 % — workers asleep in backoff |

After patching the two controllers to wrap their read-modify-write in `retry.RetryOnConflict` and lifting the worker-pool ceiling, sustained throughput in the same lab rose to **~72 RB/s sustained / 108 RB/s peak** — roughly **60× the default**.

Eviction throughput during a cluster-failover storm should be bounded by `karmada-apiserver` capacity and the operator-tunable rate limits, not by an unflagged hardcoded worker count and an unhandled write-conflict between two in-tree controllers.

Specifically:

1. The `NoExecuteTaintManager`'s `syncBindingEviction` and `syncClusterBindingEviction` should retry inline on `apierrors.IsConflict(...)` instead of bouncing through workqueue backoff.
2. The `RBGracefulEvictionController` and `CRBGracefulEvictionController` should do the same on their `Patch` path.
3. The taint-manager's worker pool size should be operator-configurable via a flag, matching the existing `--concurrent-cluster-syncs` / `--concurrent-resourcebinding-syncs` / etc. naming convention.

1. Bring up a Karmada control plane plus three member clusters (kind is fine).
2. Push ~5 K identical `Deployments` + `PropagationPolicy` pairs (single-cluster placement, `replicas: 0` keeps the experiment to control-plane stress) into one namespace. Distribute so the bulk lands on one member cluster — e.g. `clusterAffinity: [member1, member2, member3]` with `spreadConstraints: maxGroups=1, minGroups=1` will tend to concentrate.
3. Stop the kubelet / docker container hosting the cluster that holds the bulk (e.g. `docker stop member2-control-plane`).
4. Sample `kubectl get rb -n <ns> -o json` every 1–2 s; track the count of bindings whose `.spec.clusters[0].name` is the failed cluster, and the count whose `.spec.gracefulEvictionTasks` is non-empty.
5. Observe sustained eviction rate ≪ 10 RB/s on the default `karmada-controller-manager` configuration.

A reference sampler and analysis script (~50 lines of bash + `jq`) are available on request.

The fix splits cleanly into three small PRs, all preserving existing behavior on the non-conflict path:

1. `fix(taint-manager)`: wrap `syncBindingEviction` / `syncClusterBindingEviction` in `retry.RetryOnConflict`. Pure correctness, no flags.
2. `fix(graceful-eviction)`: wrap `syncBinding` in both `{rb,crb}_graceful_eviction_controller.go` in the same pattern. Signature shifts to take `types.NamespacedName` so the binding can be refetched inside the retry closure. Pure correctness, no flags.
3. `feat(controller-manager)`: introduce `--concurrent-binding-eviction-syncs` (default 5, matching the rest of the `Concurrent*` family). Replaces the hardcoded `3` in `startClusterController`. UX/configurability change only.

A fourth, design-heavier change — decoupling `NewGracefulEvictionRateLimiter` from the global `--rate-limiter-qps` (the `MaxOf` composition silently caps eviction at the global QPS regardless of `--resource-eviction-rate`) — is worth a separate discussion and is intentionally not bundled here.

Each PR is self-contained, individually testable, and preserves the existing behavior on the non-conflict / single-writer path. I have local commits and a 5 K-binding bench-validated build ready to push as soon as we agree on the approach. Happy to break this up further if reviewers prefer.

**Environment**:
- Karmada version: `release-1.17` (commit `af79ff9`, tag `v1.17.2`); the same code paths exist unchanged on `master` as of commit `03b39ddb` and patches apply cleanly there.
- kubectl-karmada or karmadactl version: not relevant to the diagnosis.
- Others:
  - 4-node kind topology (1 host + 3 members), Kubernetes v1.34.0
  - 5 000 synthetic `Deployment` + `PropagationPolicy` pairs, `replicas: 0`, ~25 KiB each
  - Tuning attempted before patching: `--resource-eviction-rate=50`, `--rate-limiter-qps=200`, `--rate-limiter-bucket-size=1000`, `--kube-api-qps=200`, `--kube-api-burst=400` — none recovered throughput, confirming the bottleneck is in-controller (conflict storm + worker-pool ceiling), not apiserver-side.

RyanCopley:
**What type of PR is this?**

/kind bug

The `{RB,CRB}GracefulEvictionController` periodically patches `Spec.GracefulEvictionTasks` on bindings to remove completed or expired tasks. The patch uses `MergeFromWithOptimisticLock`, which encodes the cached `metadata.resourceVersion` and asks the apiserver to reject the write if the object has changed. The taint manager's binding-eviction worker writes the same field via `Update` with the same `resourceVersion` guard. When the two collide between each other's read and write the apiserver returns `StatusReasonConflict`, and there is no inline retry — the only recovery is the controller's exponential workqueue backoff, which starts at 5ms and grows to 1000s.

In a cluster-failover storm where the same binding cycles through both controllers many times in close succession, the conflict tax dominates wall-clock latency. A 5K-binding lab observed a **99.7% retry rate** on the `resource-binding-graceful-eviction-controller` workqueue (1072 retries across 1075 adds), with workers spending most of their time asleep in backoff rather than doing real work.

This change wraps the read-modify-patch in `retry.RetryOnConflict` for both the namespaced and cluster-scoped variants. The signature of `syncBinding` shifts from `(ctx, *Binding)` to `(ctx, types.NamespacedName)` because the binding object now lives inside the retry closure: each attempt does a fresh `Get` so the next patch's `MergeFromWithOptimisticLock` base carries the latest `resourceVersion`. `Reconcile` is simplified accordingly — the `apierrors.IsNotFound` short-circuit moves up to wrap the `syncBinding` call.

The eviction event is emitted only on the successful attempt (`EmitClusterEvictionEventForResourceBinding` / `…ForClusterResourceBinding` are now called with a nil error, since by the time we reach the loop the patch has succeeded).

Fixes #7483

This is the graceful-eviction-controller half of a conflict pair; the taint-manager half is in a sibling PR (#TODO link to PR1 once filed). Either PR is independently correct, but the throughput improvement only materializes when both ship — they are the two writers that collide on `Spec.GracefulEvictionTasks`. A third PR makes the eviction worker pool size configurable.

Tests:
- `TestRBGracefulEvictionController_syncBinding_retryOnConflict` and the `…CRB…` twin use `interceptor.Funcs` to inject two consecutive `Conflict` responses on `Patch` and assert the expired eviction task is still removed once the wrapper recovers.
- `TestRBGracefulEvictionController_syncBinding_notFoundSurfacedToReconcile` and the `…CRB…` twin verify that `NotFound` is surfaced from the closure so that `Reconcile` can short-circuit, since the closure no longer returns a no-op early.
- `go test -race ./pkg/controllers/gracefuleviction/...` is clean.

One heads-up: the pre-existing `TestCRBGracefulEvictionController_syncBinding` tolerance on the wall-clock-derived `retryAfter` value is loosened from `100ms` to `1s`. The original tolerance was already brittle — it compared a `metav1.Now()` captured at test setup against one captured inside the controller after fake-client setup work. The new `Get` inside the retry closure pushed the gap past `100ms` on slower hosts. `1s` is still well within the meaning of the assertion, which is "the controller computed roughly 3 minutes." Happy to factor this differently (e.g. inject a clock) if reviewers prefer.

Local benchmark on a 5K-binding kind lab failing one cluster, with both this PR and the taint-manager PR applied: sustained ~72 RB/s, peak 108 RB/s (vs. ~1.2 RB/s on stock controller-manager defaults). Methodology and raw CSV available on request.

```release-note
`karmada-controller-manager`: Fixed an optimistic-lock conflict storm in the resource-binding and cluster-resource-binding graceful-eviction controllers that caused eviction-task cleanup to stall during cluster-failover events. Both controllers now retry on conflict in-line instead of bouncing through the workqueue's exponential backoff.
```

### DIFF
--- a/pkg/controllers/gracefuleviction/crb_graceful_eviction_controller.go
+++ b/pkg/controllers/gracefuleviction/crb_graceful_eviction_controller.go
@@ -23,7 +23,9 @@ import (
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
 	controllerruntime "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
@@ -54,20 +56,11 @@ type CRBGracefulEvictionController struct {
 func (c *CRBGracefulEvictionController) Reconcile(ctx context.Context, req controllerruntime.Request) (controllerruntime.Result, error) {
 	klog.V(4).InfoS("Reconciling ClusterResourceBinding", "name", req.NamespacedName.String())
 
-	binding := &workv1alpha2.ClusterResourceBinding{}
-	if err := c.Client.Get(ctx, req.NamespacedName, binding); err != nil {
+	retryDuration, err := c.syncBinding(ctx, req.NamespacedName)
+	if err != nil {
 		if apierrors.IsNotFound(err) {
 			return controllerruntime.Result{}, nil
 		}
-		return controllerruntime.Result{}, err
-	}
-
-	if !binding.DeletionTimestamp.IsZero() {
-		return controllerruntime.Result{}, nil
-	}
-
-	retryDuration, err := c.syncBinding(ctx, binding)
-	if err != nil {
 		return controllerruntime.Result{}, err
 	}
 	if retryDuration > 0 {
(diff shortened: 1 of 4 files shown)


RyanCopley:
Okey dokey, I suspect the 1.35.0 failure is a flaky test, not my change

XiShanYongYe-Chang:
Hi @RyanCopley, I can retry the test.

PR #7517, opened 2026-05-17
fix(karmadactl): return error from DeleteConfirmation instead of calling os.Exit

### LINKED ISSUE #7516: fix(karmadactl/util): stop calling os.Exit in DeleteConfirmation, return error instead
`DeleteConfirmation()` in `pkg/karmadactl/util/deleteconfirmation.go` should return
`(bool, error)` instead of calling `os.Exit(1)` on stdin read failure, and use an
iterative loop instead of recursion for retrying invalid input.

**Files changed:**
- `pkg/karmadactl/util/deleteconfirmation.go` — core fix: signature change, `os.Exit` removed, recursion replaced with `for` loop, invalid-input feedback added
- `pkg/karmadactl/util/deleteconfirmation_test.go` — new file: 8 table-driven tests covering `y/n/yes/no`, uppercase variants, empty stdin error path, and invalid-then-valid retry flow
- `pkg/karmadactl/deinit/deinit.go` — caller updated to handle `(bool, error)` return
- `pkg/karmadactl/addons/init/disable_option.go` — caller updated to handle `(bool, error)` return

**Why is this needed**:
`DeleteConfirmation()` is called from two production code paths (`deinit` and
`addons disable`). Calling `os.Exit(1)` inside a library function causes three
concrete problems.

First, the error path is untestable. Any test that reaches the stdin read failure
kills the whole test process.

Second, callers have no chance to recover. When exit fires, `deinit` and
`addons disable` can't log, clean up, or surface a structured error. The process dies
with a bare `fmt.Println(err)` and nothing to indicate which command triggered it.

Third, it breaks in non-interactive environments. CI pipelines, piped stdin, and
scripted usage all cause `fmt.Scanln` to error immediately, so both commands fail
at the confirmation step unless the caller passes `--force`. That's a bad default
for tooling meant to be scriptable.

D3S-Gaurav:
**What type of PR is this?**
/kind bug
/kind cleanup

`DeleteConfirmation()` in `pkg/karmadactl/util/deleteconfirmation.go` called
`os.Exit(1)` when `fmt.Scanln` failed — in CI, piped stdin, or any
non-interactive environment. That's the wrong tool for a library function: callers
can't catch it, tests can't survive it, and the user gets no context about what
went wrong or which command triggered it.

Changes in this PR:

- Signature changes from `func DeleteConfirmation() bool` to
  `func DeleteConfirmation() (bool, error)` so callers can handle stdin failures
  instead of dying
- `os.Exit(1)` is replaced with `return false, fmt.Errorf("failed to read user
  confirmation: %w", err)`
- The recursive `default: return DeleteConfirmation()` retry is replaced with a
  `for` loop; it also now prints `"invalid input, please type (y)es or (n)o"` so
  users aren't left wondering why nothing happened
- Both callers (`deinit/deinit.go` and `addons/init/disable_option.go`) are
  updated to handle the `(bool, error)` return and propagate failures through
  cobra's `RunE`
- `deleteconfirmation_test.go` is new: 8 table-driven tests covering `y/n/yes/no`,
  uppercase variants, empty stdin (the previously unkillable error path), and an
  invalid-then-valid retry sequence

Fixes #7516

**Note for reviewer**:
The stdin error path had no test coverage before this change because hitting it
called `os.Exit(1)`, which killed the test binary. The new tests use `os.Pipe()`
to swap out stdin — no subprocess, no PTY needed.

A follow-up worth considering (out of scope here): inject `io.Reader` and
`io.Writer` into `DeleteConfirmation()` to remove the dependency on `os.Stdin`
and `os.Stdout` entirely.

```release-note
karmadactl: `deinit` and `addons disable` no longer exit the process silently
when stdin is unavailable (e.g. in CI or piped usage). They now return an error
that cobra surfaces to the user. Passing `--force` remains the supported way to
skip confirmation in non-interactive environments.
```

### DIFF
--- a/pkg/karmadactl/addons/init/disable_option.go
+++ b/pkg/karmadactl/addons/init/disable_option.go
@@ -68,8 +68,14 @@ func (o *CommandAddonsDisableOption) Validate(args []string) error {
 // Run start disable Karmada addons
 func (o *CommandAddonsDisableOption) Run(args []string) error {
 	fmt.Printf("Disable Karmada addon %s\n", args)
-	if !o.Force && !cmdutil.DeleteConfirmation() {
-		return nil
+	if !o.Force {
+		confirmed, err := cmdutil.DeleteConfirmation()
+		if err != nil {
+			return fmt.Errorf("failed to get user confirmation: %w", err)
+		}
+		if !confirmed {
+			return nil
+		}
 	}
 
 	var disableAddons = map[string]*Addon{}
--- a/pkg/karmadactl/deinit/deinit.go
+++ b/pkg/karmadactl/deinit/deinit.go
@@ -255,9 +255,15 @@ func removeLabels(node *corev1.Node, removesLabel string) {
 // Run start delete
 func (o *CommandDeInitOption) Run() error {
 	fmt.Println("removes Karmada from Kubernetes")
-	// delete confirmation,exit the delete action when false.
-	if !o.Force && !util.DeleteConfirmation() {
-		return nil
+	// delete confirmation, exit the delete action when user declines or input fails.
+	if !o.Force {
+		confirmed, err := util.DeleteConfirmation()
+		if err != nil {
+			return fmt.Errorf("failed to get user confirmation: %w", err)
+		}
+		if !confirmed {
+			return nil
+		}
 	}
 
 	if err := o.delete(); err != nil {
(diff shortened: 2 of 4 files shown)


D3S-Gaurav:
Hi @chaosi-zju,
I wanted to kindly follow up on this PR in case it’s pending review. This is my first contribution here, and I’d appreciate any feedback whenever convenient. Thank you!

Issue #7481, opened 2026-05-07
[Umbrella] Implementation of scheduler estimator assumption


XiShanYongYe-Chang:
In #7416 we propose a proposal to address the issue #6783, this is a big feature, and we can split it into the following subtasks to facilitate implementation and code review.

Fixed #6783 

- [x] Refine scheduling overcommit protection proposal @RainbowMango #7556

Implementation Task Breakdown

**Group A — gRPC Proto & Estimator Client**

- [x] T1 · Extend generated.proto (no deps) @XiShanYongYe-Chang #7491 
- [x] T2 · Extend ComponentSetEstimationRequest (deps: T1) @XiShanYongYe-Chang #7491
- [x] T3 · Wire AssumedWrokloads  into accurate client (deps: T1, T2) @mszacillo #7509 
- [x] T4 · Cluster-level assumption deduction in GeneralEstimator (deps: T1, T2) @XiShanYongYe-Chang #7502 #7568

**Group B — Estimator Server Framework**

- [x] T11 · Add assumedWrokloads to estimator server framework (deps: T1) @XiShanYongYe-Chang #7503 
- [x] T12 · noderesource plugin — FF-based deduction (deps: T11) @XiShanYongYe-Chang #7526
- [x] T13 · resourcequota plugin — quota-level deduction (deps: T11) @mszacillo #7538

**Group C — Scheduler**

- [x] T5 · Extend AssigningResourceBindingCache (no deps) @XiShanYongYe-Chang #7496 
- [x] T6 · ~Add --assumption-ttl scheduler flag (deps: T5)~
- [x] T7 · Thread AssumedWrokloads into calculateMultiTemplateAvailableSets (deps: T2, T5)  @XiShanYongYe-Chang #7505 
- [x] T8 · Write/release assumptions in patchScheduleResultForResourceBinding (deps: T5) @XiShanYongYe-Chang #7519 
- [x] T9 · Watch AggregatedStatus changes in ResourceBinding update handler (deps: T5) @XiShanYongYe-Chang  #7525
- [x] T10 · Start GC goroutine in Scheduler.Run() (deps: T5) @XiShanYongYe-Chang #7520 

**Group D — Tests & Docs**

- [x] T16 · E2E test 
  - [x] multi template (taking FlinkDeployment as an example) -- noderesource @XiShanYongYe-Chang #7551 
  - [x] multi template (taking FlinkDeployment as an example) -- resourcequota @XiShanYongYe-Chang #7551 
[...]

PR #7519, opened 2026-05-18
Write/Release assumptions in patchScheduleResultForResourceBinding

XiShanYongYe-Chang:
**What type of PR is this?**

/kind feature

Write/Release assumptions in patchScheduleResultForResourceBinding.

- Only resources of the workload type need to be calculated. Therefore, you only need to set the scheduler assumption cache for resources of the workload type.
- Generally, workload resources are namespace-scoped resources. Therefore, you do not need to process the scheduling result of ClusterResourceBinding.

In addition, we first handle multi-template workloads. Single-template resources, such as Deployments and StatefulSets, are not processed for now.

Part of #7481

The current implementation writes the complete components of workloads into the assumption cache after each scheduling cycle. This approach has several drawbacks:

**Over-deduction**: The assumption cache includes resources occupied by pods that are already steadily running in clusters. When the Estimator evaluates remaining cluster capacity, it calculates both the resource usage of actually running pods and the full resource deductions recorded in assumptions simultaneously. This results in double deduction of available resources for the same cluster, leading to overly conservative scheduling decisions afterwards.

The duration of this issue is restricted by TTL and cache release mechanisms. Further optimization and mitigation can be implemented on the Estimator side in follow-up work.

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -62,6 +62,7 @@ import (
 	"github.com/karmada-io/karmada/pkg/util/grpcconnection"
 	"github.com/karmada-io/karmada/pkg/util/helper"
 	utilmetrics "github.com/karmada-io/karmada/pkg/util/metrics"
+	"github.com/karmada-io/karmada/pkg/util/names"
 )
 
 // ScheduleType defines the schedule type of a binding object should be performed.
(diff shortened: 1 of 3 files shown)


pkg/scheduler/scheduler.go
```diff
@@ -690,11 +691,60 @@ func (s *Scheduler) patchScheduleResultForResourceBinding(oldBinding *workv1alph
...
+		}
+	}
+
+	// Write the full component footprint for every cluster in the new schedule result.
+	for clusterName := range newClusters {
+		assigningCache.Assume(bindingKey, clusterName, schedulercache.AssumedWorkload{
+			Namespace:  oldBinding.Spec.Resource.Namespace,
+			Components: oldBinding.Spec.Components,
+		})
+	}
```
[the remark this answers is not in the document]
XiShanYongYe-Chang: These tasks have been split into other PRs.

RainbowMango (commented):
/assign

pkg/scheduler/scheduler.go
```diff
@@ -690,11 +691,60 @@ func (s *Scheduler) patchScheduleResultForResourceBinding(oldBinding *workv1alph
...
+//   - Stale assumptions are released for clusters removed from the placement.
+//   - The full current Components are assumed for every cluster in the new schedule result.
+//     This is intentionally conservative, which may over-subtract during reschedules with
+//     no replica change, but ensures that newly created pods are visible to the estimator.
+//
+// Resources with no Components (e.g. non-workloads) are skipped.
+func (s *Scheduler) updateAssumptionsCache(oldBinding *workv1alpha2.ResourceBinding, scheduleResult []workv1alpha2.TargetCluster) {
+	if s.schedulerCache == nil || s.schedulerCache.AssigningResourceBindings() == nil {
+		return
+	}
```
RainbowMango, line 712: I think the two conditions(nil check) are unreachable. Please add a comment to explain why we need this, like to avoid false positive warnings.

pkg/scheduler/scheduler.go
```diff
@@ -690,11 +691,60 @@ func (s *Scheduler) patchScheduleResultForResourceBinding(oldBinding *workv1alph
...
+
+	// Skip resources that carry no Components (e.g. non-workloads).
+	if len(oldBinding.Spec.Components) == 0 {
+		return
+	}
```
RainbowMango:
```suggestion
	if !oldBinding.Spec.IsWorkload() {
		return
	}
```
Why not use the method if you want to skip the non-workload?

pkg/scheduler/scheduler.go
XiShanYongYe-Chang: I added a line of comment.

pkg/scheduler/scheduler.go
XiShanYongYe-Chang: It makes sense, I update it,

RainbowMango (approved):
/lgtm
/approve

PR #7526, opened 2026-05-19
FF-based assumption deduction within noderesource plugin

XiShanYongYe-Chang:
**What type of PR is this?**

/kind feature

Part of #7481

The deduction loop adds one SimulateScheduling(components, 1) call per assumed workload before the main estimation. Since upperBound=1, each call exits after a single FF pass — there is no repeated unmarshal.

The natural upper bound on assumed workload count is: throughput × TTL = 100 RB/min × 5 min = 500. A benchmark over 500 assumed workloads at node scales of 100 / 500 / 1000 shows the deduction adds approximately 30 ms per EstimateComponents call, with negligible sensitivity to cluster size. This is considered acceptable for a gRPC estimation request.

> Benchmark environment: Intel Xeon Gold 6278C @ 2.60GHz, 4 vCPUs, 15 GiB RAM. Results may vary on production hardware with more cores, as Go benchmarks are single-goroutine by default.

```release-note
NONE
```

### DIFF
--- a/pkg/estimator/server/framework/plugins/noderesource/noderesource.go
+++ b/pkg/estimator/server/framework/plugins/noderesource/noderesource.go
@@ -127,6 +127,8 @@ func getNodeAvailableResource(node *schedulerframework.NodeInfo) *util.Resource 
 
 // EstimateComponents estimates the maximum number of complete component sets that can be scheduled.
 // It returns the number of sets that can fit on the available node resources.
+// Before running the main estimation, it deducts any assumed (in-flight) workloads so that
+// pods already assigned but not yet reflected in cluster metrics are accounted for.
 func (pl *nodeResourceEstimator) EstimateComponents(_ context.Context, estCtx framework.ComponentEstimationContext) (int32, *framework.Result) {
 	if !pl.enabled {
 		return pl.disabledResult()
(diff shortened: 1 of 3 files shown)


pkg/estimator/server/framework/plugins/noderesource/noderesource.go
```diff
@@ -143,7 +145,27 @@ func (pl *nodeResourceEstimator) EstimateComponents(_ context.Context, estCtx fr
...
 	}
 
-	sets, err := estimator.NewSchedulingSimulator(nodes).SimulateScheduling(components, math.MaxInt32)
+	simulator := estimator.NewSchedulingSimulator(nodes)
+
+	// Deduct assumed (in-flight) workloads before running the main estimation.
+	// Each assumed workload represents pods that are being started but may not yet
+	// be reflected in the cluster's node resource accounting.
+	for _, assumed := range estCtx.AssumedWorkloads {
+		if len(assumed.GetComponents()) == 0 {
```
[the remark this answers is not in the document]
XiShanYongYe-Chang: I don't think it's necessary to do so.

pkg/estimator/server/framework/plugins/noderesource/noderesource.go
```diff
@@ -143,7 +145,27 @@ func (pl *nodeResourceEstimator) EstimateComponents(_ context.Context, estCtx fr
...
+			continue
+		}
+		placed, deductErr := simulator.SimulateScheduling(assumed.GetComponents(), 1)
```
RainbowMango:
```suggestion
		deductedSets, deductErr := simulator.SimulateScheduling(assumed.GetComponents(), 1)
```

pkg/estimator/server/framework/plugins/noderesource/noderesource.go
```diff
@@ -143,7 +145,27 @@ func (pl *nodeResourceEstimator) EstimateComponents(_ context.Context, estCtx fr
...
+		if deductErr != nil {
+			return 0, framework.AsResult(deductErr)
+		}
+		if placed == 0 {
+			// The assumed workload could not be placed in the simulation (e.g., resource
+			// fragmentation or tight capacity). Skip this deduction and continue.
+			klog.V(4).Infof("%s: could not deduct assumed workload (namespace=%s) from node resources", pl.Name(), assumed.GetNamespace())
+		}
```
RainbowMango:
```suggestion
		if deductedSets == 0 {
			// The assumed workload could not be placed in the simulation. This is expected when
			// the workload has already been admitted by the cluster and its resource usage is
			// already reflected in node.Requested, so no additional deduction is needed.
			klog.V(4).Infof("%s: assumed workload (namespace=%s) likely already reflected in node resources, skipping deduction", pl.Name(), assumed.GetNamespace())
		}
```

XiShanYongYe-Chang: Thanks~

RainbowMango (approved):
/lgtm
/approve

PR #7529, opened 2026-05-21
Distinguish event handlers registered for binding resources in the scheduler based on the type of the binding

XiShanYongYe-Chang:
**What type of PR is this?**

/kind cleanup

Compared with ClusterResourceBinding, ResourceBinding has some additional processing in the event handler. If the same function is still used for event processing, the differences will become larger and larger, which will bring more burdens or even errors to subsequent iterations. Therefore, the two functions are processed separately.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/event_handler.go
+++ b/pkg/scheduler/event_handler.go
@@ -61,8 +61,8 @@ func (s *Scheduler) addAllEventHandlers() {
 	_, err = clusterBindingInformer.AddEventHandler(cache.FilteringResourceEventHandler{
 		FilterFunc: s.resourceBindingEventFilter,
 		Handler: cache.ResourceEventHandlerFuncs{
-			AddFunc:    s.onResourceBindingAdd,
-			UpdateFunc: s.onResourceBindingUpdate,
+			AddFunc:    s.onClusterResourceBindingAdd,
+			UpdateFunc: s.onClusterResourceBindingUpdate,
 		},
 	})
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/scheduler/event_handler.go
```diff
@@ -120,33 +120,49 @@ func (s *Scheduler) resourceBindingEventFilter(obj any) bool {
...
+	}
+
 	if features.FeatureGate.Enabled(features.PriorityBasedScheduling) {
-		bindingInfo := newQueuedBindingInfo(obj)
-		if bindingInfo == nil {
-			// shouldn't happen
-			klog.Errorf("couldn't convert to QueuedBindingInfo %#v", obj)
+		s.priorityQueue.Push(newQueuedBindingInfoFromRB(rb))
+	} else {
+		key, err := cache.MetaNamespaceKeyFunc(obj)
```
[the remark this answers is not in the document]
XiShanYongYe-Chang: Updated.

pkg/scheduler/event_handler.go
```diff
@@ -120,113 +120,145 @@ func (s *Scheduler) resourceBindingEventFilter(obj any) bool {
...
 			return
 		}
-
 		s.queue.Add(key)
 	}
 	metrics.CountSchedulerBindings(metrics.BindingUpdate)
 }
 
 // onResourceBindingDelete is called when a delete event for a ResourceBinding is received from the Informer.
 func (s *Scheduler) onResourceBindingDelete(obj any) {
```
RainbowMango, line 240: This refactor is not necessary. 
We can't imply the entire function body is controlled by the feature gate, as it makes it harder to add future logic that should run regardless of whether the feature gate is enabled.

RainbowMango (changes requested):

pkg/scheduler/event_handler.go
XiShanYongYe-Chang: Okay, updated.

RainbowMango (approved):
/lgtm
/approve

PR #7544, opened 2026-05-23
fix: use os.IsNotExist for path existence check

QCsnakeSUFE:
**What type of PR is this?**

/kind bug

`os.Stat()` returns lookup-related errors such as `ENOENT`, `ENOTDIR`, or `EACCES`, but it does not return `EEXIST`.

The current implementation uses `os.IsExist(err)` after `os.Stat()` fails. This may incorrectly return `false` for errors that do not mean the path is absent, such as permission-denied errors.

This PR switches the check to `!os.IsNotExist(err)`, so the helper only returns `false` when the path is explicitly reported as non-existent.

NONE

NONE

```release-note

NONE

### DIFF
--- a/pkg/karmadactl/cmdinit/kubernetes/deploy.go
+++ b/pkg/karmadactl/cmdinit/kubernetes/deploy.go
@@ -415,7 +415,11 @@ func (i *CommandInitOption) initializeCommandLineArgs() {
 
 // initializeDirectory initializes a directory and makes sure it's empty.
 func initializeDirectory(path string) error {
-	if utils.IsExist(path) {
+	exist, err := utils.PathExists(path)
+	if err != nil {
+		return fmt.Errorf("failed to check path %q: %w", path, err)
+	}
+	if exist {
 		if err := os.RemoveAll(path); err != nil {
 			return err
 		}
@@ -518,7 +522,10 @@ func (i *CommandInitOption) prepareCRD() error {
 
 	for _, archive := range validation.CrdsArchive {
 		expectedDir := filepath.Join(i.KarmadaDataPath, archive)
-		exist, _ := utils.PathExists(expectedDir)
+		exist, err := utils.PathExists(expectedDir)
+		if err != nil {
+			return fmt.Errorf("failed to check path %q: %w", expectedDir, err)
+		}
 		if !exist {
 			return fmt.Errorf("lacking the necessary file path: %s", expectedDir)
 		}
--- a/pkg/karmadactl/cmdinit/utils/format_test.go
+++ b/pkg/karmadactl/cmdinit/utils/format_test.go
@@ -28,32 +28,6 @@ func stringInslice(target string, strArray []string) bool {
 	return slices.Contains(strArray, target)
 }
 
-func TestIsExist(t *testing.T) {
-	tests := []struct {
-		name string
-		path string
-		want bool
-	}{
-		{
-			name: "path exist",
-			path: "./",
-			want: true,
-		},
-		{
-			name: "path is not exist",
-			path: "for-a-not-exist-path" + randString(),
-			want: false,
-		},
-	}
-	for _, tt := range tests {
-		t.Run(tt.name, func(t *testing.T) {
-			if got := IsExist(tt.path); got != tt.want {
-				t.Errorf("IsExist() = %v, want %v", got, tt.want)
-			}
-		})
-	}
-}
-
 func TestStringToNetIP(t *testing.T) {
 	tests := []struct {
 		name string
--- a/pkg/karmadactl/cmdinit/utils/format.go
+++ b/pkg/karmadactl/cmdinit/utils/format.go
@@ -41,12 +41,6 @@ const (
 	MaxRespBodyLength = 1 << 20 // 1 MiB
 )
 
-// IsExist Determine whether the path exists
-func IsExist(path string) bool {
-	_, err := os.Stat(path)
-	return err == nil
-}
-
 // StringToNetIP String To NetIP
 func StringToNetIP(addr string) net.IP {
 	if ip := net.ParseIP(addr); ip != nil {
--- a/pkg/karmadactl/cmdinit/utils/util_test.go
+++ b/pkg/karmadactl/cmdinit/utils/util_test.go
@@ -137,3 +137,34 @@ func TestListFiles(t *testing.T) {
 		})
 	}
 }
+
+func TestPathExists(t *testing.T) {
+	tests := []struct {
+		name string
+		path string
+		want bool
+	}{
+		{
+			name: "path exist",
+			path: "./",
+			want: true,
+		},
+		{
+			name: "path is not exist",
+			path: "for-a-not-exist-path" + randString(),
+			want: false,
+		},
+	}
+	for _, tt := range tests {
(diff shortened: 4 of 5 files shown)

### CODE REVIEW

**Comment by zhzhuang-zju** [comment]
>os.Stat() returns lookup-related errors such as ENOENT, ENOTDIR, or EACCES, but it does not return EEXIST.

Hi @QCsnakeSUFE, can you elaborate more on this? If it doesn't return `EEXIST`, does that mean `os.IsNotExist` will always return false?

**Comment by QCsnakeSUFE** [comment]
Hi @zhzhuang-zju, thanks for your review!

Yes, I think the key point is about `os.IsExist` here. I assume you meant `os.IsExist` in your question. Because `os.Stat()` is a read-only lookup operation (calling fstatat under the hood), EEXIST is not the semantic error used to report the result of this check. EEXIST is normally associated with creation-style operations where the target already exists. Therefore, return `os.IsExist(err)` is not a meaningful way to classify stat failures here, and it makes the helper return false for errors that are not necessarily “path absent”.

This coincidence masks the problem for ENOENT (file not found), but incorrectly returns false for errors like EACCES (permission denied), causing the caller to treat the path as absent even though the existence check is actually inconclusive. By switching to `!os.IsNotExist(err)`, the helper returns false only when the stat error is classified as a not-exist error by Go.

A cleaner design would be the three-valued form used by ([k8s.io/utils/path.Exists](https://github.com/kubernetes/utils/blob/master/path/file.go#L56-L61)). 

Karmada also has a similar helper([`utils.PathExists`](https://github.com/karmada-io/karmada/blob/master/pkg/karmadactl/cmdinit/utils/util.go#162-173)). I didn't change `apiclient.Exists` to return `(bool, error)` in this PR mainly because I want to keep the API backward-compatible and keep the fix minimal. In my opinion, using `!os.IsNotExist(err)` is the safest minimal fix for the boolean-return API.

By the way, I also noticed another related helper, [`utils.IsExist`](https://github.com/karmada-io/karmada/blob/master/pkg/karmadactl/cmdinit/utils/format.go#L45-L48), which may be worth checking separately.

**Comment by QCsnakeSUFE** [comment]
Hi @zhzhuang-zju, just a gentle ping on this PR.

I have added more context above about why `os.IsExist(err)` is not meaningful for errors returned by `os.Stat()`, and why `!os.IsNotExist(err)` is used as a minimal backward-compatible fix for the current boolean API.

Please let me know if you think this should instead be changed to a `(bool, error)` helper, or if you prefer handling the related helpers in a separate PR.

**Comment by zhzhuang-zju** [comment]
Thanks @QCsnakeSUFE for your detailed explanation.

When using `os.Stat` together with `os.IsExist`, errors such as `permission denied` will be mistakenly treated as a non-existent file error. Is that right?

You’ve proposed two approaches:
- Switch to using `!os.IsNotExist(err)` as a minimal fix
- Refactor this function to return a `(bool, error)` helper

I'm in favor of the second approach. First, it won't take much work, since all upstream functions already handle errors. Second, it follows good programming practices: errors should be handled by upstream callers rather than being suppressed internally.
Also, you mentioned `utils.PathExists` and `utils.IsExist`. That’s a great point, and we can look at this from a holistic perspective. I suggest removing `utils.IsExist` and `apiclient.Exist`. These three utility functions do essentially the same thing, so we don’t need so many duplicate helpers.

**Comment by QCsnakeSUFE** [comment]
@zhzhuang-zju Yes, that's right. 

Thanks for the suggestion. I’ve updated this PR to use the existing `utils.PathExists` helper directly instead of keeping duplicate boolean helpers.

Changes made:
- Removed `apiclient.Exists`.
- Removed `utils.IsExist`.
- Updated the callers to use `utils.PathExists` and propagate non-not-exist stat errors instead of suppressing them.
- Updated the related unit test to cover `PathExists`.

I also noticed there is a similar `PathExists` helper under `operator/pkg/util`. I kept this PR focused on the `karmadactl` helpers and call sites discussed here to keep the diff small and reviewable. I can follow up separately if you’d like to consolidate the operator helper as well.

File: pkg/karmadactl/cmdinit/utils/format_test.go [hunk 1]
```diff
@@ -28,7 +28,7 @@ func stringInslice(target string, strArray []string) bool {
 	return slices.Contains(strArray, target)
 }
 
-func TestIsExist(t *testing.T) {
+func TestPathExist(t *testing.T) {
```
zhzhuang-zju: move this test function to the `pkg/karmadactl/cmdinit/utils/util_test.go`

pkg/karmadactl/cmdinit/kubernetes/deploy.go
```diff
@@ -415,7 +415,11 @@ func (i *CommandInitOption) initializeCommandLineArgs() {
 
 // initializeDirectory initializes a directory and makes sure it's empty.
 func initializeDirectory(path string) error {
-	if utils.IsExist(path) {
+	exist, err := utils.PathExists(path)
+	if err != nil {
+		return err
```
zhzhuang-zju: It would be better to wrap the `path` info in the err msg to help with troubleshooting.

zhzhuang-zju (commented):
Thanks @QCsnakeSUFE 

> I also noticed there is a similar PathExists helper under operator/pkg/util. I kept this PR focused on the karmadactl helpers and call sites discussed here to keep the diff small and reviewable. I can follow up separately if you’d like to consolidate the operator helper as well.

We don't need to account for functions under the operator directory; functions inside operator can be treated as a standalone unit.

QCsnakeSUFE:
@zhzhuang-zju Done. 

I moved the path existence test to `util_test.go` and wrapped the path in the returned error for easier troubleshooting.

pkg/karmadactl/cmdinit/kubernetes/deploy.go
```diff
@@ -415,7 +415,11 @@ func (i *CommandInitOption) initializeCommandLineArgs() {
 
 // initializeDirectory initializes a directory and makes sure it's empty.
 func initializeDirectory(path string) error {
-	if utils.IsExist(path) {
+	exist, err := utils.PathExists(path)
+	if err != nil {
+		return fmt.Errorf("failed to check path %q: %w", path, err)
```
zhzhuang-zju, line 420: I noticed you have wrapped the path in the returned error for easier troubleshooting. That's great!

Could you apply the same fix to other errors in this PR?

zhzhuang-zju (commented):
Thanks @QCsnakeSUFE, others LGTM

zhzhuang-zju:
@QCsnakeSUFE Thanks for pointing this out. This PR standardizes file existence check functions and prevents non-file-not-found errors from being suppressed.

But I don't think this is a bug. First, these file paths are created and managed by Karmada itself, so errors like EACCES are unlikely to occur. Second, if such errors (e.g., permission issues) did happen, they would fail at the `os.MkdirAll` stage. So I might need to change the PR label to `cleanup` and hold off backporting pending feedback. WDYT?

QCsnakeSUFE:
Thanks for the clarification, @zhzhuang-zju. I agree this change is better categorized as cleanup rather than a bug fix. I'm fine with changing the label and not backporting this unless maintainers think otherwise.

I also applied the same error wrapping to the other affected error returns for consistency. All CI checks are now passing.

PR #7552, opened 2026-05-25
feat(execution): only modify member cluster objects in execution controller

zach593:
**What type of PR is this?**

/kind cleanup 
/kind feature

This PR refines the responsibility boundary between the execution controller and the work-status controller. Previously, the work-status controller was responsible for both reflecting member cluster resource status back to Work objects AND detecting/reconciling drifts in member cluster resources (e.g., updating or recreating objects that had diverged from the desired state). This mixing of concerns made the codebase harder to reason about.

After this change, the execution controller takes over the write path — it directly watches member cluster resources via informers, detects meaningful drifts by comparing actual objects against version records (ignoring irrelevant fields like status and managedFields), and reconciles them. The work-status controller is simplified to focus solely on the read path — reflecting status from member cluster resources back to Work objects.

The result is a cleaner separation of concerns: the execution controller drives member cluster objects toward the desired state, while the work-status controller observes and reports their actual status.

Fixes #

Local benchmark results show that when restarting the controller-manager against an already-synced cluster, the time to drain the queue is virtually unchanged compared to before.

Test configuration:

image

- 20,000 Deployments, each with replicas split 1:1 across 2 member clusters.
- All controllers set to 100 workers, REST QPS capped at 500.
- GOMAXPROCS=8.

image

The queue drain takes approximately 3 minutes.

The informer startup for both 2 controllers no longer blocks during member cluster informer sync. This avoids the problem where a handful of resources being established could stall all workers and delay distribution or status collection for other resource types or member clusters. While the informer is still being established, the controller requeues after 60 seconds and retries.

```release-note
`karmada-controller-manager`: The execution controller now handles all member cluster object modifications, while the work-status controller is scoped to status collection only.
```


pkg/controllers/execution/execution_controller.go
```diff
@@ -396,3 +433,148 @@ func (c *Controller) eventf(object *unstructured.Unstructured, eventType, reason
...
+	if c.eventChannel == nil {
+		klog.V(5).InfoS("Ignore member cluster object event as event channel is not initialized")
+		return
+	}
+	clientObject, ok := object.(client.Object)
+	if !ok {
+		klog.ErrorS(nil, "Ignore event as object does not implement client.Object", "object", object)
+		return
+	}
+	c.eventChannel <- event.TypedGenericEvent[client.Object]{Object: clientObject}
```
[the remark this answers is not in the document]
zach593: Since the channel's backend is the controller's workqueue, blocking won't last long even if it happens. Between brief blocking and losing messages, I choose blocking here.

pkg/controllers/execution/execution_controller.go
```diff
@@ -396,3 +433,148 @@ func (c *Controller) eventf(object *unstructured.Unstructured, eventType, reason
...
+	workNS, nsExist := annotations[workv1alpha2.WorkNamespaceAnnotation]
+	_, nameExist := annotations[workv1alpha2.WorkNameAnnotation]
+	if !nsExist || !nameExist {
+		klog.V(5).InfoS("Skip resource missing Karmada work annotations",
+			"apiVersion", oldObj.GetAPIVersion(),
+			"kind", oldObj.GetKind(),
+			"namespace", oldObj.GetNamespace(),
+			"name", oldObj.GetName())
+		return
+	}
```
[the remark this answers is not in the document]
zach593: Once the member cluster object is created by the execution controller, these two annotations are guaranteed to exist. If an annotation is modified so that `curObj` cannot map to the corresponding work, it can still be mapped from `oldObj`. Therefore, searching for it in `curObj` is meaningless.

pkg/util/helper/cache.go
```diff
@@ -27,13 +27,75 @@ import (
...
+func EnsureInformerHandlersReady(
+	cluster *clusterv1alpha1.Cluster,
+	gvrTargets []schema.GroupVersionResource,
+	handler cache.ResourceEventHandler,
+	manager genericmanager.MultiClusterInformerManager,
+	clusterClientSetFunc util.NewClusterDynamicClientSetFunc,
+	kubeClientSet client.Client,
+	clusterClientOption *util.ClientOption,
+) (bool, error) {
+	singleClusterInformerManager, err := getSingleClusterManager(cluster, manager, clusterClientSetFunc, kubeClientSet, clusterClientOption)
```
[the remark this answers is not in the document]
zach593: That would actually require checking every single parameter, which would be too verbose. The existing caller code doesn't cause this problem.

pkg/controllers/execution/execution_controller.go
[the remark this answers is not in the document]
zach593: ditto, see https://github.com/karmada-io/karmada/pull/7552#discussion_r3334122314

RainbowMango (commented):
/assign

pkg/controllers/execution/execution_controller.go
```diff
@@ -128,11 +145,28 @@ func (c *Controller) Reconcile(ctx context.Context, req controllerruntime.Reques
...
+
+	informersSynced, err := helper.EnsureInformerHandlersReady(cluster, gvrTargets, c.getEventHandler(), c.InformerManager, c.ClusterClientSetFunc, c.Client, c.ClusterClientOption)
+	if err != nil {
+		klog.ErrorS(err, "Failed to ensure informers for Work.", "namespace", work.GetNamespace(), "name", work.GetName())
+		return controllerruntime.Result{}, err
+	}
+	if !informersSynced {
+		klog.V(4).InfoS("Member cluster informers are not synced yet.", "namespace", work.GetNamespace(), "name", work.GetName(), "cluster", cluster.Name)
+		return controllerruntime.Result{RequeueAfter: memberInformerNotSyncedRequeueAfter}, nil
```
RainbowMango: The `memberInformerNotSyncedRequeueAfter` is 60s, which means when a new cluster joins or a Work introduces a new GVR, the Work waits the full 60s before retrying.  For instance, I want to deploy my Deployment to one more cluster (never sync Deployment to that cluster before), I probably need to wait for 60s. That's concerning, I guess. 

What I'm thinking is:
1. Do we really need to wait for the informer sync?
2. If yes, can we give it a small retry interval? (By the way, do you know some data on how long it usually takes for an informer to get synced?)

pkg/controllers/execution/execution_controller.go
```diff
@@ -128,11 +145,28 @@ func (c *Controller) Reconcile(ctx context.Context, req controllerruntime.Reques
...
+	}
+
 	return c.syncWork(ctx, clusterName, work)
 }
 
 // SetupWithManager creates a controller and register to controller manager.
 func (c *Controller) SetupWithManager(mgr controllerruntime.Manager) error {
+	c.eventChannel = make(chan event.TypedGenericEvent[client.Object], 1024)
```
RainbowMango: Shall we add a comment on why `1024` was selected?

I bet the controller-runtime will consume the channel very fast(moving data from the channel to the queue), thus, the producer is unlikely to get blocked. So I don't see any problem here. Just wonder how the number comes out.

pkg/controllers/execution/execution_controller.go
```diff
@@ -396,3 +433,148 @@ func (c *Controller) eventf(object *unstructured.Unstructured, eventType, reason
...
+			"clusterName", clusterName)
+		return
+	}
+
+	// No meaningful content difference after pruning irrelevant fields.
+	if reflect.DeepEqual(oldCopy, curCopy) {
+		return
+	}
+
+	c.enqueueWorkload(curObj)
```
RainbowMango: Here enquing the `curObj`, and then parsing the Work key from the label. My question is: what if the label is removed mistakenly? The Work will be ignored from requing, right? 

In addition, can we just put the Work key into the queue?  That will save the parsing effort.

pkg/controllers/execution/execution_controller.go
```diff
@@ -396,3 +433,148 @@ func (c *Controller) eventf(object *unstructured.Unstructured, eventType, reason
...
+		return
+	}
+	c.eventChannel <- event.TypedGenericEvent[client.Object]{Object: clientObject}
+}
+
+// getEventHandler return callback function that knows how to handle events from the member cluster.
+func (c *Controller) getEventHandler() cache.ResourceEventHandler {
+	c.eventHandlerOnce.Do(func() {
+		if c.eventHandler == nil {
+			c.eventHandler = fedinformer.NewHandlerOnEvents(nil, c.onUpdate, c.onDelete)
```
RainbowMango: Can you explain why the `Add` event is ignored? Is there any trade-off?

Consider one corner case where the resource is updated unexpectly during karmada-controller-manager is gone and `karmada-controller-manager` will not notice it when starting.

pkg/controllers/execution/execution_controller.go
```diff
@@ -396,3 +433,148 @@ func (c *Controller) eventf(object *unstructured.Unstructured, eventType, reason
...
+		if old == nil {
+			return
+		}
+	}
+	oldObj, ok := old.(runtime.Object)
+	if !ok {
+		klog.ErrorS(nil, "Failed to assert old object as runtime.Object", "old", old)
+		return
+	}
+	c.enqueueWorkload(oldObj)
```
RainbowMango: Do we need to check where the object is managed by Karmada here?

cmd/agent/app/agent.go
```diff
@@ -345,12 +345,13 @@ func startClusterStatusController(ctx controllerscontext.Context) (bool, error)
...
-		ObjectWatcher:      ctx.ObjectWatcher,
-		InformerManager:    genericmanager.GetInstance(),
-		RateLimiterOptions: ctx.Opts.RateLimiterOptions,
+		Client:               ctx.Mgr.GetClient(),
+		EventRecorder:        ctx.Mgr.GetEventRecorderFor(execution.ControllerName), //nolint:staticcheck // Note: GetEventRecorderFor is deprecated in controller-runtime v0.23.0 in favor of GetEventRecorder. This changes event API from v1 events to events.k8s.io. We need to migrate carefully, especially considering the impact on users and RBAC permission changes in installation/deployment tools.
+		RESTMapper:           ctx.Mgr.GetRESTMapper(),
+		ObjectWatcher:        ctx.ObjectWatcher,
+		InformerManager:      genericmanager.GetInstance(),
+		RateLimiterOptions:   ctx.Opts.RateLimiterOptions,
+		ClusterClientSetFunc: util.NewClusterDynamicClientSetForAgent,
```
RainbowMango, line 354: Does the `karmada-agent` need `clusterClientOption`?

RainbowMango (commented):
Generally looks great! Thank you!

pkg/controllers/execution/execution_controller.go
zach593: In the propagation scenario, `Work` objects are the data source, while the resources in member clusters are downstream objects. When the execution controller starts, all `Work` objects will be reconciled once, which in turn causes all managed resources in member clusters to be refreshed. Therefore, even if a member-cluster resource is unexpectedly modified while `karmada-controller-manager` is down, the change will be corrected after the controller restarts.

pkg/controllers/execution/execution_controller.go
zach593: This is indeed a detail worth discussing. In fact, I think the non-blocking approach may even deserve a separate issue for further discussion.

With the current synchronous approach, a worker can get stuck here waiting for the corresponding member-cluster informer to become ready, and that informer may never become ready. In the worst case, only a small number of resources could occupy most or even all workers at the informer creation stage, which could significantly affect the controller's SLO.

Switching to a non-blocking approach clearly addresses this problem. What I am less certain about is the best retry interval. If the interval is too short, it could cause a retry storm. Even if there is no real impact, unusually high retry metrics could still alarm administrators. On the other hand, if the interval is too long, it introduces unnecessary waiting. One minute feels like a relatively balanced value to me.

As for exponential backoff, I don't think `informer not synced` should be treated as an error, so directly returning an error does not seem appropriate. `Requeue: true` has also been deprecated, while implementing a custom rate limiter just for this case seems unnecessarily complicated.

Another possible approach would be to simply `return controllerruntime.Result{}, nil` here, and then add the missing `onAdd` handling mentioned in this discussion:
https://github.com/karmada-io/karmada/pull/7552#discussion_r3694609806

I'd be interested to hear what you think.

BTW, in our clusters, it can take up to 3–4 minutes for the informer of a resource type to fully sync.

pkg/controllers/execution/execution_controller.go
zach593: updated.

pkg/controllers/execution/execution_controller.go
zach593: This references https://github.com/kubernetes-sigs/controller-runtime/blob/3be3f1bf2b2fcc6b5c9510d55c6a9972294653d0/pkg/source/source.go#L188

cmd/agent/app/agent.go
zach593: Yes, it does. I’ve wired it into the other relevant agent-side controller initializers as well.

pkg/controllers/execution/execution_controller.go
zach593: updated.

pkg/controllers/execution/execution_controller.go
zach593: updated.

pkg/controllers/execution/execution_controller.go
RainbowMango: Good point, shall we add a comment to explain it?

pkg/controllers/execution/execution_controller.go
zach593: After reconsidering, the `onAdd` approach I mentioned is not feasible. For a Work being propagated for the first time, the corresponding resource does not exist in the member cluster yet, so there will be no add event to re-enqueue the Work.

pkg/controllers/execution/execution_controller.go
zach593: done.

pkg/util/helper/cache.go
```diff
@@ -27,13 +27,75 @@ import (
...
+	clusterv1alpha1 "github.com/karmada-io/karmada/pkg/apis/cluster/v1alpha1"
+	"github.com/karmada-io/karmada/pkg/util"
 	"github.com/karmada-io/karmada/pkg/util/fedinformer/genericmanager"
 	"github.com/karmada-io/karmada/pkg/util/fedinformer/keys"
 	"github.com/karmada-io/karmada/pkg/util/restmapper"
 )
 
+// EnsureInformerHandlersReady ensures the member cluster informers for the given resources exist
+// and have the handler registered. It returns whether those informers have already synced without waiting.
+func EnsureInformerHandlersReady(
```
RainbowMango: This function registers informer and returns the sync status, it can't not gureentee the informer ready, so use `Ensure` might be misleading. 

```suggestion
// EnsureInformerHandlersReady ensures the member cluster informers for the given resources exist
// and have the handler registered. It returns whether those informers have already synced without waiting.
func RegisterInformersAndCheckSynced(
```

pkg/controllers/execution/execution_controller.go
```diff
@@ -128,11 +143,29 @@ func (c *Controller) Reconcile(ctx context.Context, req controllerruntime.Reques
...
+
+	informersSynced, err := helper.EnsureInformerHandlersReady(cluster, gvrTargets, c.getEventHandler(), c.InformerManager, c.ClusterClientSetFunc, c.Client, c.ClusterClientOption)
+	if err != nil {
+		klog.ErrorS(err, "Failed to ensure informers for Work.", "namespace", work.GetNamespace(), "name", work.GetName())
+		return controllerruntime.Result{}, err
+	}
+	if !informersSynced {
+		klog.V(4).InfoS("Member cluster informers are not synced yet.", "namespace", work.GetNamespace(), "name", work.GetName(), "cluster", cluster.Name)
+		// Do not return when informers are not synced; the following logic can proceed with the dynamic client.
+	}
```
RainbowMango, line 159: Here is a corner case:

- One Work (new GVR) is synced via the dynamic client, and the informer still needs several minutes to sync
- If the member cluster object is modified during this window, the change shows up as an `Add` event in the informer's initial list, but we will ignore that event.

What do you think of this?

pkg/util/helper/cache.go
zach593: Renamed to RegisterInformerHandlerAndCheckSynced because this function also registers the event handler.

pkg/controllers/execution/execution_controller.go
zach593: Updated.

- Added filtering for member-cluster resource Add events, requeueing Work only when the observed resource version differs from the recorded one.
- Added unit tests covering the new GVR informer scenario.
- Made a small performance optimization to `waitForCacheSync`.

pkg/util/work.go
```diff
@@ -79,6 +81,41 @@ func IsWorkContains(manifests []workv1alpha1.Manifest, targetResource schema.Gro
...
+func GetGVRsFromWork(restMapper meta.RESTMapper, work *workv1alpha1.Work) ([]schema.GroupVersionResource, error) {
+	var gvrTargets []schema.GroupVersionResource
+	gvrSet := map[schema.GroupVersionResource]bool{}
+	for _, manifest := range work.Spec.Workload.Manifests {
+		workload := &unstructured.Unstructured{}
+		if err := workload.UnmarshalJSON(manifest.Raw); err != nil {
+			klog.ErrorS(err, "Failed to unmarshal workload.",
+				"workNamespace", work.GetNamespace(),
+				"workName", work.GetName(),
+			)
```
RainbowMango, line 94: TODO: It feels a bit heavy to unmarshal the object just to parse the GVK. Just came up with an idea to introduce workload metadata in the Work API. I'm not sure it is worth doing.

pkg/util/helper/cache.go
```diff
@@ -27,13 +27,89 @@ import (
...
+// have already synced without waiting.
+func RegisterInformerHandlerAndCheckSynced(
+	cluster *clusterv1alpha1.Cluster,
+	gvrTargets []schema.GroupVersionResource,
+	handler cache.ResourceEventHandler,
+	manager genericmanager.MultiClusterInformerManager,
+	clusterClientSetFunc util.NewClusterDynamicClientSetFunc,
+	kubeClientSet client.Client,
+	clusterClientOption *util.ClientOption,
+) (bool, error) {
```
RainbowMango, line 52:
```suggestion
func RegisterInformerHandlerAndCheckSynced(cluster *clusterv1alpha1.Cluster, gvrTargets []schema.GroupVersionResource, handler cache.ResourceEventHandler, manager genericmanager.MultiClusterInformerManager, clusterClientSetFunc util.NewClusterDynamicClientSetFunc, kubeClientSet client.Client, clusterClientOption *util.ClientOption) (bool, error) 
```

pkg/util/fedinformer/genericmanager/single-cluster-manager.go
```diff
@@ -175,14 +175,30 @@ func (s *singleClusterInformerManagerImpl) WaitForCacheSyncWithTimeout(cacheSync
 }
 
 func (s *singleClusterInformerManagerImpl) waitForCacheSync(ctx context.Context) map[schema.GroupVersionResource]bool {
-	s.lock.Lock()
-	defer s.lock.Unlock()
```
RainbowMango, line 179: nice job!

Don't have to hold the lock at the beginning.

pkg/util/helper/cache.go
```diff
@@ -27,13 +27,89 @@ import (
...
+	return false, nil
+}
+
+func getSingleClusterManager(
+	cluster *clusterv1alpha1.Cluster,
+	manager genericmanager.MultiClusterInformerManager,
+	clusterClientSetFunc util.NewClusterDynamicClientSetFunc,
+	kubeClientSet client.Client,
+	clusterClientOption *util.ClientOption,
+) (genericmanager.SingleClusterInformerManager, error) {
```
RainbowMango, line 99:
```suggestion
func getSingleClusterManager(cluster *clusterv1alpha1.Cluster, manager genericmanager.MultiClusterInformerManager, clusterClientSetFunc util.NewClusterDynamicClientSetFunc, kubeClientSet client.Client, clusterClientOption *util.ClientOption) (genericmanager.SingleClusterInformerManager, error) {
```

RainbowMango (approved):
/lgtm
/approve

pkg/util/fedinformer/genericmanager/single-cluster-manager.go
zach593: Actually, `s.informerFactory.WaitForCacheSync` still holds a write lock, so I also optimized `RegisterInformerHandlerAndCheckSynced` to reduce the lock contention.

RainbowMango:
After this PR, the execution controller takes over all member cluster object modifications. When a resource is accidentally deleted in a member cluster, it is now recreated through the normal reconcile path, instead of the dedicated recreation path that used to live in the work-status controller.

So, [the CountCreateResourceToCluster  is now only called with  recreate=false](https://github.com/karmada-io/karmada/blob/82a9f4fa41fce763bb80ee90b1c9f84d624d08ff/pkg/controllers/execution/execution_controller.go#L363) . The  recreate  parameter and the  recreate  label of the  create_resource_to_cluster  metric become dead code and should be removed ( pkg/metrics/resource.go ).

@zach593 Let me know your thoughts.

zach593:
> After this PR, the execution controller takes over all member cluster object modifications. When a resource is accidentally deleted in a member cluster, it is now recreated through the normal reconcile path, instead of the dedicated recreation path that used to live in the work-status controller.
> 
> So, [the CountCreateResourceToCluster  is now only called with  recreate=false](https://github.com/karmada-io/karmada/blob/82a9f4fa41fce763bb80ee90b1c9f84d624d08ff/pkg/controllers/execution/execution_controller.go#L363) . The  recreate  parameter and the  recreate  label of the  create_resource_to_cluster  metric become dead code and should be removed ( pkg/metrics/resource.go ).
> 
> @zach593 Let me know your thoughts.

Good catch. There are a few cleanup items left over from this PR, including this dead code. I'll open a follow-up PR soon to address them together.

PR #7557, opened 2026-05-26
fix: avoid overwriting cluster summaries when node/pod listing fails

### LINKED ISSUE #7511: [bug]: cluster-status-controller overwrites NodeSummary/ResourceSummary with zeros on transient lister failures, causing scheduler to treat healthy cluster as zero capacity
## What happened

When `listNodes` or `listPods` returns `nil` or an error inside 
`setCurrentClusterStatus`, the controller silently overwrites 
`Cluster.Status.NodeSummary` and `Cluster.Status.ResourceSummary` 
with zero/empty values. The scheduler then reads zero allocatable capacity and stops placing new replicas on that cluster even though the cluster is healthy and `Ready=True`.

---

## What you expected to happen

When lister calls fail transiently (informer resync, brief RBAC 
revocation, control-plane upgrade), the last-known-good snapshot of 
`NodeSummary` and `ResourceSummary` should be preserved. The bad data 
should never reach the Cluster object.

---

## How to reproduce it

**Unit test proof** (added to `pkg/controllers/status/`):

The test simulates `listNodes` and `listPods` returning `nil` (the 
failure path) and asserts that the prior good snapshot is preserved.

**Result — both assertions fail:**
actual: &NodeSummary{TotalNum:0, ReadyNum:0}
actual: &ResourceSummary{Allocatable: nil, ...}

The prior good snapshot `(TotalNum:5, Allocatable: 10cpu/40Gi)` is 
silently replaced with zeros exactly what happens during a transient 
lister failure.

**Realistic trigger sequence:**
1. Federation of 3 clusters with `EnableClusterResourceModeling=true`
2. Workload with `replicaSchedulingType: Divided`, `weightPreference: dynamic`
3. Member cluster control-plane upgrade briefly restarts kube-apiserver
4. `listPods` returns `nil, err` (forbidden) during that window
5. `currentClusterStatus.ResourceSummary.Allocatable` written as zero map
6. Scheduler reads zero capacity, skips cluster for new replicas
7. Two reconciles later (~30-60s) the listers re-sync — but scheduling 
   decisions made in the bad window stick

---

## Root cause

`setCurrentClusterStatus` in 
`pkg/controllers/status/cluster_status_controller.go` swallows 
lister errors and overwrites unconditionally:

```go
nodes, err := listNodes(clusterInformerManager)
if err != nil {
    klog.ErrorS(err, "Failed to list nodes for Cluster", ...)
}  // err swallowed

pods, err := listPods(clusterInformerManager)
if err != nil {
    klog.ErrorS(err, "Failed to list pods for Cluster", ...)
}  // err swallowed

currentClusterStatus.NodeSummary = getNodeSummary(nodes)       // nodes may be nil
currentClusterStatus.ResourceSummary = getResourceSummary(nodes, pods)  // both may be nil
```

`getNodeSummary(nil)` returns `&NodeSummary{TotalNum:0, ReadyNum:0}`.
`getResourceSummary(nil, nil)` returns `&ResourceSummary{Allocatable: nil}`.

These are then written unconditionally to the Cluster object.

Note: the APIEnablements path 18 lines above in the same function 
already has an explicit `len(apiEnables) == 0` guard and a 
`ClusterConditionCompleteAPIEnablements` condition. The same defensive 
pattern was not applied to `NodeSummary`/`ResourceSummary`.

---

## Proposed fix

Preserve the last-known-good snapshot when the current collection 
returns empty or error mirroring the existing APIEnablements guard:

```go
nodes, errN := listNodes(clusterInformerManager)
pods, errP := listPods(clusterInformerManager)
if errN != nil || errP != nil || len(nodes) == 0 {
    klog.ErrorS(errors.Join(errN, errP),
        "Skipping ResourceSummary update for Cluster",
        "cluster", cluster.GetName(), "nodeCount", len(nodes))
    currentClusterStatus.NodeSummary = cluster.Status.NodeSummary
    currentClusterStatus.ResourceSummary = cluster.Status.ResourceSummary
} else {
    currentClusterStatus.NodeSummary = getNodeSummary(nodes)
    currentClusterStatus.ResourceSummary = getResourceSummary(nodes, pods)
}
```

Behavior is unchanged on the happy path. On the failure path, the 
last-good summary is preserved instead of being clobbered with zeros.

---

## Impact

- Scheduling correctness regression: healthy cluster temporarily 
  appears as zero capacity
- Silent: cluster stays `Ready=True`, no condition or event surfaces 
  the staleness
- Affects `--cluster-resource-modeling` (default for production 
  deployments using divided/dynamic-weight scheduling)
- Recoverable but lossy: next reconcile fixes the data, but any 
  scheduling decisions made in the bad window persist

---

## Environment

- Karmada version: master (affects all versions with 
  `EnableClusterResourceModeling`)
- Reproduced via unit test in 
  `pkg/controllers/status/cluster_status_controller_test.go`

Ady0333:
**What type of PR is this?**
  
  /kind bug
  /kind cleanup

---

  Skip overwriting NodeSummary and ResourceSummary in setCurrentClusterStatus when listNodes or listPods returns
   an error or a nil slice. The last-good values on currentClusterStatus (already deep-copied from
  cluster.Status) are preserved while the rest of the status (KubernetesVersion, APIEnablements, Ready) still
  refreshes that cycle.

  Without this, a transient lister failure or empty result would zero out NodeSummary.TotalNum and
  ResourceSummary.Allocatable, which the scheduler then reads as "cluster has no capacity" and routes replicas
  elsewhere.

  Before:
```go
  nodes, err := listNodes(clusterInformerManager)
  if err != nil {
      klog.ErrorS(err, "Failed to list nodes for Cluster", "cluster", cluster.GetName())
  }
  pods, err := listPods(clusterInformerManager)
  if err != nil {
      klog.ErrorS(err, "Failed to list pods for Cluster", "cluster", cluster.GetName())
  }
  currentClusterStatus.NodeSummary = getNodeSummary(nodes)
  currentClusterStatus.ResourceSummary = getResourceSummary(nodes, pods)

  if features.FeatureGate.Enabled(features.CustomizedClusterResourceModeling) {
      currentClusterStatus.ResourceSummary.AllocatableModelings = getAllocatableModelings(cluster, nodes, pods)
  }
```

  After:
```go
  nodes, nodesErr := listNodes(clusterInformerManager)
  if nodesErr != nil {
      klog.ErrorS(nodesErr, "Failed to list nodes, preserving previous summaries for cluster", "cluster",
  cluster.GetName())
  }
  pods, podsErr := listPods(clusterInformerManager)
  if podsErr != nil {
      klog.ErrorS(podsErr, "Failed to list pods, preserving previous summaries for cluster", "cluster",
  cluster.GetName())
  }
  if nodesErr == nil && podsErr == nil && nodes != nil && pods != nil {
      currentClusterStatus.NodeSummary = getNodeSummary(nodes)
      currentClusterStatus.ResourceSummary = getResourceSummary(nodes, pods)

      if features.FeatureGate.Enabled(features.CustomizedClusterResourceModeling) {
          currentClusterStatus.ResourceSummary.AllocatableModelings = getAllocatableModelings(cluster, nodes,
  pods)
      }
  }
```
---

  Fixes #7511

  - Minimal change, scoped to setCurrentClusterStatus.
  - No early return the caller still reaches updateStatusIfNeeded, so other status fields keep refreshing.
  - Covers both the error path and the nil, nil repro case from the issue.

  Does this PR introduce a user-facing change?

  `karmada-controller-manager`: Fixed a bug where transient node/pod listing failures could overwrite a
  cluster's `NodeSummary` and `ResourceSummary` with zeroed values, causing the scheduler to treat a healthy
  cluster as having zero capacity.

### DIFF
--- a/pkg/controllers/status/cluster_status_controller.go
+++ b/pkg/controllers/status/cluster_status_controller.go
@@ -269,20 +269,20 @@ func (c *ClusterStatusController) setCurrentClusterStatus(clusterClient *util.Cl
 			// if clusterInformerManager fails to be built, should be returned, otherwise, it may cause a nil pointer
 			return nil, err
 		}
-		nodes, err := listNodes(clusterInformerManager)
-		if err != nil {
-			klog.ErrorS(err, "Failed to list nodes for Cluster", "cluster", cluster.GetName())
+		nodes, nodesErr := listNodes(clusterInformerManager)
+		if nodesErr != nil {
+			klog.ErrorS(nodesErr, "Failed to list nodes, preserving previous NodeSummary for cluster", "cluster", cluster.GetName())
+		} else {
+			currentClusterStatus.NodeSummary = getNodeSummary(nodes)
 		}
-
-		pods, err := listPods(clusterInformerManager)
-		if err != nil {
-			klog.ErrorS(err, "Failed to list pods for Cluster", "cluster", cluster.GetName())
-		}
-		currentClusterStatus.NodeSummary = getNodeSummary(nodes)
-		currentClusterStatus.ResourceSummary = getResourceSummary(nodes, pods)
-
-		if features.FeatureGate.Enabled(features.CustomizedClusterResourceModeling) {
-			currentClusterStatus.ResourceSummary.AllocatableModelings = getAllocatableModelings(cluster, nodes, pods)
+		pods, podsErr := listPods(clusterInformerManager)
+		if podsErr != nil {
+			klog.ErrorS(podsErr, "Failed to list pods, preserving previous ResourceSummary for cluster", "cluster", cluster.GetName())
+		} else if nodesErr == nil {
+			currentClusterStatus.ResourceSummary = getResourceSummary(nodes, pods)
+			if features.FeatureGate.Enabled(features.CustomizedClusterResourceModeling) {
+				currentClusterStatus.ResourceSummary.AllocatableModelings = getAllocatableModelings(cluster, nodes, pods)
+			}
 		}
 	}
 	return conditions, nil


Ady0333:
Hello @XiShanYongYe-Chang ! Please review this pr.

XiShanYongYe-Chang:
Hi @Ady0333 
Sorry, I might not have time recently. I should start working on this PR in the next version, which will be early next month.

Ady0333:
> Hi @Ady0333 Sorry, I might not have time recently. I should start working on this PR in the next version, which will be early next month.

No worries @XiShanYongYe-Chang !!! Take your time...

Tej-Katika:
@Ady0333 Took a look at this since it's in an area I've been working around in, and a couple of questions I have: 

- The repro case in #7511 has `listNodes/listPods` returning `nil` with no error, and that's the case that zeroes the summary. But since `err == nil` there, I don't think `if err != nil { return }` actually catches it? So I'm not sure this prevents the overwrite the issue shows. The `err != nil` path also seems hard to reach, since `buildInformerForCluster` waits for the caches to sync before we list.

- From the thread it sounded like the idea was to skip overwriting the summaries rather than return — and returning makes the caller bail before `updateStatusIfNeeded`, so the `KubernetesVersion/APIEnablements/Ready` refresh gets skipped that cycle too. Since `currentClusterStatus` is already a deep copy of `cluster.Status`, would `return conditions, nil` on the error path work better here?

Might also be worth a small test that the prior summaries survive the failure path. What do you think?

Ady0333:
Good catches, thanks for looking @Tej-Katika .
  
  You're right that the repro in #7511 has nodes == nil with err == nil, so the if err != nil { return } doesn't
   actually cover that path. And returning early does skip updateStatusIfNeeded, which I didn't want either.

  Going to switch to your suggestion guard the assignments instead of returning, so prior
  NodeSummary/ResourceSummary are preserved (they're already on currentClusterStatus via the deep copy) and the
  rest of the status still refreshes:

```go
  if err == nil && nodes != nil {
      currentClusterStatus.NodeSummary = getNodeSummary(nodes)
      if pods != nil {
          currentClusterStatus.ResourceSummary = getResourceSummary(nodes, pods)
      }
  }
```

  (or splitting the two error checks similarly). Will also add a small unit test that the prior summaries
  survive the failure/nil path. Will push an update shortly.

PR #7566, opened 2026-05-27
Reduce unnecessary watch events in endpointslice-dispatch-controller

PrasannaMishra001:
**What type of PR is this?**

/kind cleanup

The endpointslice-dispatch-controller already cleans up via the `MCSEndpointSliceDispatchControllerFinalizer`, and the reconcile path handles the `DeletionTimestamp` transition through the update event. The final delete event therefore doesn't add anything, it just re-enqueues a work that no longer exists. This PR makes the predicate's `DeleteFunc` return false so those events are dropped.

Part of #6780

No new tests, the predicate functions are inline in `SetupWithManager` and aren't covered by existing tests.

```release-note
NONE
```


XiShanYongYe-Chang:
Hi @PrasannaMishra001 

PR #7570, opened 2026-05-28
feat: add scheduler empty-workload chart option

### LINKED ISSUE #7572: Expose scheduler empty-workload propagation in Helm values
## Problem

We hit a race condition during zero-replica transitions:

- the source workload is updated to `replicas=0`,
- while in parallel Karmada binding/work reconciliation may treat target changes as orphan cleanup and delete Work,
- resulting in non-deterministic behavior during transitions (Work deleted vs expected zero-replica propagation).

In this scenario we need scheduler support for empty workloads to keep propagation behavior consistent when replicas are zero.

Today the Helm chart does not expose `--enable-empty-workload-propagation`, so enabling the behavior requires post-deploy patching in infrastructure code (Terraform `null_resource` + `kubectl patch`) just to add one scheduler flag.

## Why this is painful

- **Race amplification**: operational recovery from this race depends on external patching instead of a chart value.
- **Operational fragility**: patch logic is outside Helm and can drift from chart state.
- **Upgrade risk**: scheduler rollout/upgrade can remove the patched arg until patch logic re-runs.
- **Extra permissions/complexity**: requires direct API patching and credential plumbing from IaC.
- **Harder troubleshooting**: desired scheduler behavior is not declarative in Helm values.

## Expected

Helm chart should provide a dedicated value, for example:

- `scheduler.enableEmptyWorkloadPropagation: false` (default)
- when enabled, render `--enable-empty-workload-propagation=true` in `karmada-scheduler` command args.

## Context

PR with implementation:
- https://github.com/karmada-io/karmada/pull/7570

If maintainers agree with this direction, we can iterate on naming/defaults as needed.

tamarubin:
**What type of PR is this?**

/kind feature

Adds a Helm chart value to expose scheduler empty-workload propagation:

- `scheduler.enableEmptyWorkloadPropagation` (default: `false`)
- when enabled, render `--enable-empty-workload-propagation=true` in `karmada-scheduler` command args
- document the new value in chart README

This avoids post-deploy patching (for example Terraform `null_resource` + `kubectl patch`) to set a single scheduler flag.

Fixes #7572

- Scope is intentionally minimal: one dedicated scheduler value for this flag.
- Agree that broader/flexible component command-line configuration (for example `extraFlags`) is useful as a follow-up capability.

```release-note
`karmada-chart`: Added `scheduler.enableEmptyWorkloadPropagation` Helm value (default `false`). When set to `true`, the chart renders `--enable-empty-workload-propagation=true` for `karmada-scheduler`.
```


jabellard:
Code changes generally look good.

/lgtm

/cc @zhzhuang-zju @seanlaii for another look. 

Also, as a follow up, @zhzhuang-zju maybe we should add more robust/flexible support for users to configure these type of settings.

Maybe by provide a `extraFlags` option for the components. What do you guys think?

jabellard:
Also, we need a note regarding this user facing change as part of the PR description.

@TamarRubin , can you please address that? Please take a look at his [PR](https://github.com/karmada-io/karmada/pull/7457) as an example.

a7i (approved):
/lgtm

PR #7585, opened 2026-05-29
update wrong memory format in the flinkdeployment-cr.yaml

XiShanYongYe-Chang:
**What type of PR is this?**

/kind cleanup

The `flinkdeployment-cr.yaml` use 100m (milli-bytes) of memory. As noted, memory should be specified in realistic units like Mi (e.g., 100Mi / 100M) rather than milli-bytes (m), which represent fractional bytes and are not meaningful for memory in Kubernetes.

Fixes #

```release-note
NONE
```

### DIFF
--- a/test/e2e/suites/base/schedule_multi_template_test.go
+++ b/test/e2e/suites/base/schedule_multi_template_test.go
@@ -65,15 +65,19 @@ var _ = ginkgo.Describe("[ScheduleMultiTemplate] schedule multi template resourc
 		return nil
 	}
 
-	// Helper function to verify component resource requirements
+	// Helper function to verify component resource requirements.
+	// Uses Cmp for quantity comparison to avoid mismatches caused by different
+	// internal representations (e.g. DecimalSI vs BinarySI) for equal values.
 	verifyComponentResources := func(component *workv1alpha2.Component, expectedReplicas int32, expectedCPU, expectedMemory string) {
 		gomega.Expect(component).ShouldNot(gomega.BeNil())
 		gomega.Expect(component.Replicas).Should(gomega.Equal(expectedReplicas))
 		gomega.Expect(component.ReplicaRequirements).ShouldNot(gomega.BeNil())
-		gomega.Expect(component.ReplicaRequirements.ResourceRequest).Should(gomega.Equal(corev1.ResourceList{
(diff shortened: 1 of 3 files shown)


test/e2e/suites/base/schedule_multi_template_test.go
```diff
@@ -65,15 +65,19 @@ var _ = ginkgo.Describe("[ScheduleMultiTemplate] schedule multi template resourc
...
 		gomega.Expect(component).ShouldNot(gomega.BeNil())
 		gomega.Expect(component.Replicas).Should(gomega.Equal(expectedReplicas))
 		gomega.Expect(component.ReplicaRequirements).ShouldNot(gomega.BeNil())
-		gomega.Expect(component.ReplicaRequirements.ResourceRequest).Should(gomega.Equal(corev1.ResourceList{
-			corev1.ResourceCPU:    resource.MustParse(expectedCPU),
-			corev1.ResourceMemory: resource.MustParse(expectedMemory),
-		}))
+		actualCPU := component.ReplicaRequirements.ResourceRequest[corev1.ResourceCPU]
+		actualMemory := component.ReplicaRequirements.ResourceRequest[corev1.ResourceMemory]
+		gomega.Expect(actualCPU.Cmp(resource.MustParse(expectedCPU))).Should(gomega.Equal(0),
```
zhzhuang-zju, line 77: How about:
```suggestion
		gomega.Expect(actualCPU.Equal(resource.MustParse(expectedCPU))).Should(gomega.BeTure,
```

XiShanYongYe-Chang: The Cmp method and the Equal method both perform semantic comparisons (ignoring differences in internal formats), with the distinction lying in their return values. At the underlying level, both methods first convert the two Quantities into a unified numerical form before performing the comparison. Therefore, both approaches are correct in their implementation and have no fundamental difference.

zhzhuang-zju: Yes, you're totally right. The Equal function is just a wrapper around `q.Cmp(v) == 0`, so both are fine with me. My only concern is that using `Equal` seems more common and easier to understand. This is just my suggestion, so feel free to take it or leave it.

Issue #7596, opened 2026-06-01
Performance Optimization planned for v1.19


RainbowMango:
## Goal
Track and deliver the performance optimizations planned for the 1.19 release.

### Highlight Enhancements
TBD

### Issues/PRs/Discussions
- [x] #6858 
- [x] https://github.com/karmada-io/karmada/pull/7552
- [x] https://github.com/karmada-io/karmada/pull/7863
- [x] https://github.com/karmada-io/karmada/pull/7807

RainbowMango:
Moved the left tasks to #7906 
- https://github.com/karmada-io/karmada/pull/7800
- https://github.com/karmada-io/karmada/pull/7803
- Lua interpreter VM pool size (#7049)
- Performance feedback (#7062)

PR #7602, opened 2026-06-04
perf(luavm): cache compiled Lua FunctionProto in RunScript to avoid r…

### LINKED ISSUE #7597: [perf] : cache compiled Lua FunctionProto in RunScript to avoid redundant parse+compile on every call

  Cache compiled Lua `FunctionProto` in the VM so scripts are parsed and compiled only once instead of on every
  `RunScript` call.

  Right now `RunScript` calls `l.DoString(script)` every single invocation even though the script never changes
  for a given customization. The fix is to compile once, store in a `sync.Map` keyed by script content, and
  reuse via `NewFunctionFromProto` + `PCall`. Behavior stays identical.

  Benchmarked the isolated parse+compile cost:

  | | ns/op | B/op | allocs/op |
  |---|---|---|---|
  | `DoString` (current) | 108,936 | 63,240 | 393 |
  | Cached proto | 1,827 | 2,888 | 10 |

  File: `pkg/resourceinterpreter/customized/declarative/luavm/lua.go` :- `RunScript()`

---

  **Why is this needed**:

  `RunScript` is called on every status event through `ReflectStatus`, `InterpretHealth`, `AggregateStatus` etc.
  In large fleets this is continuous unnecessary allocation and GC pressure. The script is immutable so
  recompiling it every time is pure waste.

  Sub-task of #7596.

qsxDree:
**What type of PR is this?**
/kind feature

Cache compiled Lua `FunctionProto` in the VM to avoid redundant parse+compile on every `RunScript` call.

`RunScript` is called on every status event through `ReflectStatus`, `InterpretHealth`, `AggregateStatus`, etc. In large fleets this causes continuous unnecessary allocation and GC pressure. Since scripts are immutable, recompiling every time is pure waste.

Fixes #
This fix caches compiled protos using `sync.Map` keyed by script content and reuses via `NewFunctionFromProto` + `PCall`. Behavior remains identical.

Fixes #7597
Part of #7596

```release-note
karmada-controller-manager`/`karmada-agent`: Optimized Lua script execution in the resource interpreter by caching compiled FunctionProto, improving performance for customized resource interpretation.
```

### DIFF
--- a/pkg/resourceinterpreter/customized/declarative/luavm/lua.go
+++ b/pkg/resourceinterpreter/customized/declarative/luavm/lua.go
@@ -60,8 +60,8 @@ func New(useOpenLibs bool, poolSize int) *VM {
 	return vm
 }
 
-// compileScript parses and compiles the Lua script into a FunctionProto, caching the result to avoid redundant compilation.
-func (vm *VM) compileScript(script string) (*lua.FunctionProto, error) {
+// loadOrCompileScript returns a cached FunctionProto for script, compiling it on first use.
+func (vm *VM) loadOrCompileScript(script string) (*lua.FunctionProto, error) {
 	if value, ok := vm.funcCache.Load(script); ok {
 		return value.(*lua.FunctionProto), nil
 	}
@@ -110,7 +110,7 @@ func (vm *VM) RunScript(script string, fnName string, nRets int, args ...any) ([
 	defer cancel()
 	l.SetContext(ctx)
 
-	proto, err := vm.compileScript(script)
+	proto, err := vm.loadOrCompileScript(script)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/resourceinterpreter/customized/declarative/luavm/lua.go
+++ b/pkg/resourceinterpreter/customized/declarative/luavm/lua.go
@@ -21,9 +21,12 @@ import (
 	"encoding/json"
 	"fmt"
 	"reflect"
+	"strings"
+	"sync"
 	"time"
 
 	lua "github.com/yuin/gopher-lua"
+	"github.com/yuin/gopher-lua/parse"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/klog/v2"
@@ -40,6 +43,8 @@ type VM struct {
 	// UseOpenLibs flag to enable open libraries. Libraries are disabled by default while running, but enabled during testing to allow the use of print statements.
 	UseOpenLibs bool
 	Pool        *fixedpool.FixedPool
+	// TODO(user): Implement a garbage collection or eviction policy for funcCache to prevent memory leaks if scripts are updated or dynamically generated.
+	funcCache sync.Map // map[string]*lua.FunctionProto
 }
 
 // New creates a manager for lua VM
@@ -53,6 +58,26 @@ func New(useOpenLibs bool, poolSize int) *VM {
 		func(a any) { a.(*lua.LState).Close() },
 		poolSize)
 	return vm
+}
+
+// compileScript parses and compiles the Lua script into a FunctionProto, caching the result to avoid redundant compilation.
+func (vm *VM) compileScript(script string) (*lua.FunctionProto, error) {
+	if value, ok := vm.funcCache.Load(script); ok {
+		return value.(*lua.FunctionProto), nil
+	}
+
+	chunk, err := parse.Parse(strings.NewReader(script), "<string>")
+	if err != nil {
+		return nil, err
+	}
+
+	proto, err := lua.Compile(chunk, "<string>")
+	if err != nil {
+		return nil, err
+	}
+
+	vm.funcCache.Store(script, proto)
+	return proto, nil
 }
 
 // NewLuaState creates a new lua state.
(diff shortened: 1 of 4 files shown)


qsxDree:
/cc @RainbowMango
Would like to request a review and let me know anywhere i went wrong, Thank you

RainbowMango:
Putting this into my queue. Thanks.
/assign

RainbowMango:
@qsxDree Yes, I like the approach. But I think the memory leak concerns should be addressed, as you mentioned on the comments:
> 	// TODO(user): Implement a garbage collection or eviction policy for funcCache to prevent memory leaks if scripts are updated or dynamically generated.

pkg/resourceinterpreter/customized/declarative/luavm/lua.go
```diff
@@ -55,6 +60,26 @@ func New(useOpenLibs bool, poolSize int) *VM {
 	return vm
 }
 
+// compileScript parses and compiles the Lua script into a FunctionProto, caching the result to avoid redundant compilation.
+func (vm *VM) compileScript(script string) (*lua.FunctionProto, error) {
```
zhzhuang-zju: How about:
```suggestion
func (vm *VM) loadOrCompileScript(script string) (*lua.FunctionProto, error) {
```

zhzhuang-zju (commented):
Hi @qsxDree, this change brings performance improvements. Please add a corresponding release note in the PR description, thanks.

qsxDree:
> @qsxDree Yes, I like the approach. But I think the memory leak concerns should be addressed, as you mentioned on the comments:
> 
> > // TODO(user): Implement a garbage collection or eviction policy for funcCache to prevent memory leaks if scripts are updated or dynamically generated.

to address the memory leak concern, i could implement the functionality of clearing the ```funcCache``` on config reload as we already watch ```ResourceInterpreterCustomization``` changes in ```configManager```. this will delete the stale compiled protos when scripts are updated.
should i implement this?
cc: @RainbowMango

qsxDree:
> Hi @qsxDree, this change brings performance improvements. Please add a corresponding release note in the PR description, thanks.

done. 
cc: @zhzhuang-zju

pkg/resourceinterpreter/customized/declarative/luavm/lua.go
qsxDree: yea that makes more sense. updated!

RainbowMango:
> to address the memory leak concern, i could implement the functionality of clearing the funcCache on config reload as we already watch ResourceInterpreterCustomization changes in configManager. this will delete the stale compiled protos when scripts are updated.
should i implement this?

Do you mean let [configManager ](https://github.com/karmada-io/karmada/blob/8349e2a5f5325673afde39f32b4736b3e7815042/pkg/resourceinterpreter/customized/declarative/configurable.go#L47) refresh the cache in LuaVM?

I'm not sure, but another idea in my mind is that,  taking group/version/kind, and operation type as the cache key, the cache will be re-compiled once the script changes.

qsxDree:
> > to address the memory leak concern, i could implement the functionality of clearing the funcCache on config reload as we already watch ResourceInterpreterCustomization changes in configManager. this will delete the stale compiled protos when scripts are updated.
> > should i implement this?
> 
> Do you mean let [configManager ](https://github.com/karmada-io/karmada/blob/8349e2a5f5325673afde39f32b4736b3e7815042/pkg/resourceinterpreter/customized/declarative/configurable.go#L47) refresh the cache in LuaVM?
> 
> I'm not sure, but another idea in my mind is that, taking group/version/kind, and operation type as the cache key, the cache will be re-compiled once the script changes.

sorry for my inexperience, but i don't get how having group/version/kind, and operation type as the cache key will enable to detect change in lua script. like none of those factors will change on modification of the script right?

RainbowMango:
>  like none of those factors will change on modification of the script right?

Yes, you are right. But note that each script belongs to a Group/Version/Kind AND operation. The cache workflow might like:
- Before executing a script, try to find the FunctionProto from the cache by the GVK And operation. 
- If hit the cache, we can also figure out if the script changed or not, as the previous script in the cach as well.

qsxDree:
> > like none of those factors will change on modification of the script right?
> 
> Yes, you are right. But note that each script belongs to a Group/Version/Kind AND operation. The cache workflow might like:
> 
>     * Before executing a script, try to find the FunctionProto from the cache by the GVK And operation.
> 
>     * If hit the cache, we can also figure out if the script changed or not, as the previous script in the cach as well.

okay, you mean have the gvk + operation type as key and the {string script, compiled proto} as the value, right?
then on cache hit we can check if the scripts match, if they do we proceed with the compiled proto, and if not we compile the new script and refresh the cache with it.

PS:
noticed that for ```InterpretDependecy``` a single GVK + Operation Type can have multiple Lua scripts. so another key of index (look_index) can be used to give each dependency script its own cache, would be 0 for the rest.

RainbowMango:
> okay, you mean have the gvk + operation type as key and the {string script, compiled proto} as the value, right?

Yes, that's the idea.

> noticed that for InterpretDependecy a single GVK + Operation Type can have multiple Lua scripts. so another key of index (look_index) can be used to give each dependency script its own cache, would be 0 for the rest.

Looks great, but one more thing needs to be confirmed: is the script index constant?

qsxDree:
> Looks great, but one more thing needs to be confirmed: is the script index constant?

the index is the position in the list for the current config snapshot. so it is constant for an unchanged config but not if it is updated.
 this has a minor effect on performance as some recompilation will have to be performed in case of updation to current config.
 
 /cc @RainbowMango

PR #7611, opened 2026-06-08
Fix mergeStatus overwriting existing manifest statuses

ManojLamani:
/kind bug

The current implementation of `mergeStatus` in `pkg/controllers/status/work_status_controller.go` ignores the existing `ManifestStatuses` slice and always replaces it with a single new entry.

When a `Work` object contains multiple manifests, status updates for one manifest can overwrite and discard previously recorded statuses for other manifests. This results in silent loss of status information during reconciliation.

This PR updates `mergeStatus` to preserve existing manifest statuses by matching entries using `Identifier.Ordinal`. If a matching status already exists, it is updated; otherwise, the new status is appended to the existing slice.

N/A

* Addresses the existing TODO in `mergeStatus`.
* Preserves status entries for multiple manifests within a single `Work` object.
* Verified by running the relevant controller tests after the change.

```release-note
NONE
```

### DIFF
--- a/pkg/clusterdiscovery/clusterapi/clusterapi.go
+++ b/pkg/clusterdiscovery/clusterapi/clusterapi.go
@@ -207,7 +207,8 @@ func (d *ClusterDetector) joinClusterAPICluster(clusterWideKey keys.ClusterWideK
 
 	clusterRestConfig, err := apiclient.RestConfig("", kubeconfigPath)
 	if err != nil {
-		klog.Fatalf("Failed to get cluster-api management cluster rest config. kubeconfig: %s, err: %v", kubeconfigPath, err)
+		klog.Errorf("Failed to get cluster-api management cluster rest config. kubeconfig: %s, err: %v", kubeconfigPath, err)
+		return err
 	}
 	opts := join.CommandJoinOption{
 		DryRun:           false,
--- a/pkg/controllers/status/work_status_controller.go
+++ b/pkg/controllers/status/work_status_controller.go
@@ -438,10 +438,14 @@ func (c *WorkStatusController) buildStatusIdentifier(work *workv1alpha1.Work, cl
 	return identifier, nil
 }
 
-func (c *WorkStatusController) mergeStatus(_ []workv1alpha1.ManifestStatus, newStatus workv1alpha1.ManifestStatus) []workv1alpha1.ManifestStatus {
-	// TODO(RainbowMango): update 'statuses' if 'newStatus' already exist.
-	// For now, we only have at most one manifest in Work, so just override current 'statuses'.
-	return []workv1alpha1.ManifestStatus{newStatus}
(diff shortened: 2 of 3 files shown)


pkg/clusterdiscovery/clusterapi/clusterapi.go
```diff
@@ -207,7 +207,8 @@ func (d *ClusterDetector) joinClusterAPICluster(clusterWideKey keys.ClusterWideK
 
 	clusterRestConfig, err := apiclient.RestConfig("", kubeconfigPath)
 	if err != nil {
-		klog.Fatalf("Failed to get cluster-api management cluster rest config. kubeconfig: %s, err: %v", kubeconfigPath, err)
+		klog.Errorf("Failed to get cluster-api management cluster rest config. kubeconfig: %s, err: %v", kubeconfigPath, err)
```
seanlaii, line 210: The change from `klog.Fatalf` to `klog.Errorf` + `return err` looks reasonable. `Fatalf` calls `os.Exit(1)` which would terminate the entire process. This is too aggressive for a reconcile path where the error iㄋ potentially transient and retriable. The current fix aligns with the error handling pattern used throughout the rest of this function.                                                                        
                                                                                                                                                                                                                
One minor suggestion: it would be helpful to call this out explicitly in the PR description as well, since it's a behavior change (process crash to graceful retry) that reviewers might want to be aware of  at a glance.

seanlaii (commented):
The logic looks correct, but currently a Work only ever contains a single manifest (`CreateOrUpdateWork` in `pkg/controllers/ctrlutil/work.go` hardcodes it). The original implementation with the simple override effectively produces the same result. Is there a concrete use case or plan to support multiple manifests per `Work`? If not, this change adds complexity for a scenario that doesn't exist today. It might be better to defer this until there's an actual requirement, at which point we'd also need to update `CreateOrUpdateWork`, the `execution controller`, and add proper test coverage for the multi-manifest path.

ManojLamani:
> The logic looks correct, but currently a Work only ever contains a single manifest (`CreateOrUpdateWork` in `pkg/controllers/ctrlutil/work.go` hardcodes it). The original implementation with the simple override effectively produces the same result. Is there a concrete use case or plan to support multiple manifests per `Work`? If not, this change adds complexity for a scenario that doesn't exist today. It might be better to defer this until there's an actual requirement, at which point we'd also need to update `CreateOrUpdateWork`, the `execution controller`, and add proper test coverage for the multi-manifest path.

Thanks for the review and the additional context. @seanlaii 

I based this change on the existing TODO in `mergeStatus` and the fact that the API allows multiple manifests in a Work object. However, I agree that the current implementation path creates Work objects with a single manifest, so the existing behavior is effectively sufficient today.

Given that full support for multiple manifests would require coordinated changes in work creation, execution, and test coverage, I'm happy to defer this change until there is a concrete requirement for that workflow.

PR #7613, opened 2026-06-08
taint manager eviction queue should not enqueue bindings with indefinite taint tolerations

### LINKED ISSUE #7612: TaintManager eviction queue should not enqueue workloads with indefinite taint tolerations
**What happened**:

We were doing some failover testing and noticed that the failover was taking longer than expected for all FlinkDeployments on the cluster undergoing eviction. We currently set the `resource-eviction-rate` to be 0.2, meaning we would expect FlinkDeployments to be drained at a rate of 1 workload every 5 seconds.

What we noticed however is that there were hundreds of other non-FlinkDeployment bindings (roles, secrets, configmaps, identities, etc) also being added to the eviction queue even though they are configured to have infinite taintTolerations for this eviction! This was throwing off the eviction rate of the FlinkDeployments by quite a large margin.  

This seems to have been introduced with dynamic rate limiting: https://github.com/karmada-io/karmada/pull/6675/changes/b733d826328080984e64a1b289ea78fab1953fc5#diff-67bee5e7ec3a123c6b42ab16c716b1a180a0c5fc46c7c65899377727c3df3499. This is a great feature but ideally the eviction queue should only include workloads that actually need eviction. 

Bindings for workloads with infinite tolerations should not be added to the taint manager eviction queue. The queue should only be populated by workloads that are actually undergoing eviction. At the moment you can have workloads that need to be evicted held up by a bunch of other resources which will ultimately be ignored by the taint manager. 

Schedule a bunch of resources using propagationpolicies which set a toleration for your NoExecute taint. If you trigger taint-based eviction, you will notice the eviction queue processing ALL workloads even though they tolerate the taint.

**Environment**:
- Karmada version: v1.17

mszacillo:
**What type of PR is this?**
/kind bug

Addresses #7612, issue described in detail there.

Fixes #7612

```release-note
`karmada-controller-manager`: taint-manager eviction queue should not enqueue bindings with indefinite taint tolerations
```

### DIFF
--- a/pkg/controllers/cluster/taint_manager.go
+++ b/pkg/controllers/cluster/taint_manager.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"time"
 
+	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/types"
@@ -119,6 +120,9 @@ func (tc *NoExecuteTaintManager) Reconcile(ctx context.Context, req reconcile.Re
 }
 
 func (tc *NoExecuteTaintManager) syncCluster(ctx context.Context, cluster *clusterv1alpha1.Cluster) (reconcile.Result, error) {
+	noExecuteTaints := helper.GetNoExecuteTaints(cluster.Spec.Taints)
+	now := time.Now()
+
 	// List all ResourceBindings which are assigned to this cluster.
 	rbList := &workv1alpha2.ResourceBindingList{}
 	if err := tc.List(ctx, rbList, client.MatchingFieldsSelector{
@@ -133,7 +137,7 @@ func (tc *NoExecuteTaintManager) syncCluster(ctx context.Context, cluster *clust
 			klog.ErrorS(err, "Failed to generate federated key for ResourceBinding", "gvk", rbList.Items[i].GetObjectKind().GroupVersionKind())
 			continue
 		}
-		tc.bindingEvictionWorker.Add(key)
+		tc.enqueueBinding(tc.bindingEvictionWorker, noExecuteTaints, rbList.Items[i].Annotations, key, now)
 	}
 
 	// List all ClusterResourceBindings which are assigned to this cluster.
@@ -150,9 +154,44 @@ func (tc *NoExecuteTaintManager) syncCluster(ctx context.Context, cluster *clust
 			klog.ErrorS(err, "Failed to generate federated key for ClusterResourceBinding", "gvk", crbList.Items[i].GetObjectKind().GroupVersionKind())
 			continue
 		}
-		tc.clusterBindingEvictionWorker.Add(key)
+		tc.enqueueBinding(tc.clusterBindingEvictionWorker, noExecuteTaints, crbList.Items[i].Annotations, key, now)
 	}
 	return controllerruntime.Result{RequeueAfter: tc.ClusterTaintEvictionRetryFrequency}, nil
+}
+
+// enqueueBinding computes the remaining toleration time for a binding against the
+// given taints and enqueues it to the provided worker appropriately:
+//   - Toleration expired or not tolerated: Add() immediately
+//   - Temporary toleration still active: AddAfter(remainingTime)
+//   - Indefinite toleration: skip entirely
+//
+// The worker must match the binding kind of key (bindingEvictionWorker for
+// ResourceBinding, clusterBindingEvictionWorker for ClusterResourceBinding).
+func (tc *NoExecuteTaintManager) enqueueBinding(worker util.AsyncWorker, taints []corev1.Taint, annotations map[string]string, key keys.FederatedKey, now time.Time) {
+	if len(taints) == 0 {
+		return
+	}
+
+	placement, err := helper.GetAppliedPlacement(annotations)
+	if err != nil || placement == nil {
+		worker.Add(key)
+		return
+	}
+
+	allTolerated, usedTolerations := helper.GetMatchingTolerations(taints, placement.ClusterTolerations)
+	if !allTolerated {
+		worker.Add(key)
+		return
+	}
+
+	minTolerationTime := helper.GetMinTolerationTimeWithCurrentTime(taints, usedTolerations, now)
+	switch {
+	case minTolerationTime == 0:
+		worker.Add(key)
+	case minTolerationTime > 0:
+		worker.AddAfter(key, minTolerationTime)
+	}
+	// minTolerationTime < 0 means indefinite toleration; skip.
 }
 
 // Start starts an asynchronous loop that handle evictions.
--- a/pkg/controllers/cluster/taint_manager_test.go
+++ b/pkg/controllers/cluster/taint_manager_test.go
@@ -26,6 +26,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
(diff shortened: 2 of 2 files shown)


pkg/controllers/cluster/taint_manager.go
```diff
@@ -119,6 +120,8 @@ func (tc *NoExecuteTaintManager) Reconcile(ctx context.Context, req reconcile.Re
 }
 
 func (tc *NoExecuteTaintManager) syncCluster(ctx context.Context, cluster *clusterv1alpha1.Cluster) (reconcile.Result, error) {
+    noExecuteTaints := helper.GetNoExecuteTaints(cluster.Spec.Taints)
+
 	// List all ResourceBindings which are assigned to this cluster.
 	rbList := &workv1alpha2.ResourceBindingList{}
 	if err := tc.List(ctx, rbList, client.MatchingFieldsSelector{
```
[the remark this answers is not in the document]
mszacillo: This is already covered by `Reconcile()`:

```
// short path to skip reconciliation if NoExecute taint eviction is disabled or no target taints on Cluster
f !tc.EnableNoExecuteTaintEviction || !helper.HasNoExecuteTaints(cluster.Spec.Taints) {
}
```

mszacillo:
@RainbowMango Tested in QA - confirmed workloads with indefinite taint tolerations are skipped from the eviction queue. For the test I included 5 FlinkDeployments, after toleration was used up the eviction workers processed the workloads at a 5 second pace:

 01:13:29: Toleration not yet expired
 01:13:34: Toleration expired — 3 evictions: (app-1, app-2, app-3)
 01:13:39: +5s — 2 evictions (app-4, app-5).
 
 Let me know if you have further questions!

RainbowMango:
OK. Putting it into my queue.

RainbowMango (commented):
/assign

pkg/controllers/cluster/taint_manager.go
```diff
@@ -128,6 +131,9 @@ func (tc *NoExecuteTaintManager) syncCluster(ctx context.Context, cluster *clust
 		return controllerruntime.Result{}, err
 	}
 	for i := range rbList.Items {
+		if toleratesAllTaintsIndefinitely(noExecuteTaints, rbList.Items[i].Annotations) {
```
RainbowMango: Here prevents the items that have indefinite tolerations from being enqueued, what about those with temporary tolerations? If we put them into the queue, they also affect the rate.

mszacillo: Good point! Apologies for the delay here, will take a look.

mszacillo:
> Here prevents the items that have indefinite tolerations from being enqueued, what about those with temporary tolerations? If we put them into the queue, they also affect the rate.

@RainbowMango I was thinking more about this over the weekend. Instead of using `toleratesAllTaintsIndefinitely` to filter out bindings with indefinite tolerations, we could also compute the remaining toleration time of the binding, and then either run `bindingEvictionWorker.Add()` or `tc.bindingEvictionWorker.AddAfter()`. So syncCluster() would also now check the binding's toleration time instead of entirely deferring that to syncBindingEviction(). 

For the following scenarios we'd:

1. Temporary toleration expired -> Add() immediately
2. Temporary toleration unexpired -> AddAfter(remainingTime) 
3. Indefinite toleration -> Skip entirely

What do you think?

RainbowMango:
Yes, I agree. In other words, only push ResourceBindings to `bindingEvictionWorker` for those that should be evicted.

mszacillo:
Hi @RainbowMango this should be ready for a re-review.

RainbowMango:
ok, will do.

pkg/controllers/cluster/taint_manager.go
```diff
@@ -150,11 +154,43 @@ func (tc *NoExecuteTaintManager) syncCluster(ctx context.Context, cluster *clust
...
 	}
 	return controllerruntime.Result{RequeueAfter: tc.ClusterTaintEvictionRetryFrequency}, nil
 }
 
+// enqueueBinding computes the remaining toleration time for a binding against the
+// given taints and enqueues it appropriately:
+//   - Toleration expired or not tolerated: Add() immediately
+//   - Temporary toleration still active: AddAfter(remainingTime)
+//   - Indefinite toleration: skip entirely
+func (tc *NoExecuteTaintManager) enqueueBinding(taints []corev1.Taint, annotations map[string]string, key keys.FederatedKey, now time.Time) {
```
RainbowMango: The `enqueueBinding()` is shared by both `ResourceBinding` and `ClusterResourceBinding`, but it always routes to `tc.bindingEvictionWorker` regardless of whether the key is a ClusterResourceBinding, which causes the ResourceBinding worker to attempt a Get with an empty namespace and silently fail, leaving CRBs unevicted.

RainbowMango: Is it possible to add one more parameter `worker util.AsyncWorker` to the `enqueueBinding()`?

mszacillo: Great point! I've updated the PR to address the change.

pkg/controllers/cluster/taint_manager_test.go
```diff
@@ -905,6 +906,332 @@ func Test_getPurgeMode(t *testing.T) {
...
+	if len(clusterBindingRecords) != 1 {
+		t.Fatalf("expected 1 ClusterResourceBinding enqueued to clusterBindingEvictionWorker, got %d", len(clusterBindingRecords))
+	}
+	if clusterBindingRecords[0].name != "no-toleration-crb" {
+		t.Errorf("expected no-toleration-crb enqueued, got %s", clusterBindingRecords[0].name)
+	}
+	if clusterBindingRecords[0].delay != 0 {
+		t.Errorf("ClusterResourceBinding with no toleration should have delay == 0 (immediate), got %v", clusterBindingRecords[0].delay)
+	}
+}
```
RainbowMango, line 1202: To fix the complexity issue.
```suggestion
func Test_syncCluster_enqueuesBindingsCorrectly(t *testing.T) {
	oneSecondAgo := metav1.Time{Time: time.Now().Add(-1 * time.Second)}

	cluster := &clusterv1alpha1.Cluster{
		ObjectMeta: metav1.ObjectMeta{
			Name: "test-cluster",
		},
		Spec: clusterv1alpha1.ClusterSpec{
			Taints: []corev1.Taint{
				{
					Key:       "cluster.karmada.io/maintenance",
					Effect:    corev1.TaintEffectNoExecute,
					TimeAdded: &oneSecondAgo,
				},
			},
		},
	}

	rbInfinite := &workv1alpha2.ResourceBinding{
		ObjectMeta: metav1.ObjectMeta{
			Name:      "infinite-rb",
			Namespace: "default",
			Annotations: map[string]string{
				"policy.karmada.io/applied-placement": `{"clusterAffinity":{"clusterNames":["test-cluster"]},"clusterTolerations":[{"key":"cluster.karmada.io/maintenance","operator":"Exists","effect":"NoExecute"}]}`,
			},
		},
		Spec: workv1alpha2.ResourceBindingSpec{
			Clusters: []workv1alpha2.TargetCluster{{Name: "test-cluster", Replicas: 1}},
		},
	}

	rbFinite := &workv1alpha2.ResourceBinding{
		ObjectMeta: metav1.ObjectMeta{
			Name:      "finite-rb",
			Namespace: "default",
			Annotations: map[string]string{
				"policy.karmada.io/applied-placement": `{"clusterAffinity":{"clusterNames":["test-cluster"]},"clusterTolerations":[{"key":"cluster.karmada.io/maintenance","operator":"Exists","effect":"NoExecute","tolerationSeconds":142}]}`,
			},
		},
		Spec: workv1alpha2.ResourceBindingSpec{
			Clusters: []workv1alpha2.TargetCluster{{Name: "test-cluster", Replicas: 1}},
		},
	}

	rbNoToleration := &workv1alpha2.ResourceBinding{
		ObjectMeta: metav1.ObjectMeta{
			Name:      "no-toleration-rb",
			Namespace: "default",
			Annotations: map[string]string{
				"policy.karmada.io/applied-placement": `{"clusterAffinity":{"clusterNames":["test-cluster"]},"clusterTolerations":[]}`,
			},
		},
		Spec: workv1alpha2.ResourceBindingSpec{
			Clusters: []workv1alpha2.TargetCluster{{Name: "test-cluster", Replicas: 1}},
		},
	}

	crbNoToleration := &workv1alpha2.ClusterResourceBinding{
		ObjectMeta: metav1.ObjectMeta{
			Name: "no-toleration-crb",
			Annotations: map[string]string{
				"policy.karmada.io/applied-placement": `{"clusterAffinity":{"clusterNames":["test-cluster"]},"clusterTolerations":[]}`,
			},
		},
		Spec: workv1alpha2.ResourceBindingSpec{
			Clusters: []workv1alpha2.TargetCluster{{Name: "test-cluster", Replicas: 1}},
		},
	}

	scheme := gclient.NewSchema()
	rbIndexerFunc := func(obj client.Object) []string {
		rb, ok := obj.(*workv1alpha2.ResourceBinding)
		if !ok {
			return nil
		}
		return util.GetBindingClusterNames(&rb.Spec)
	}
	crbIndexerFunc := func(obj client.Object) []string {
		crb, ok := obj.(*workv1alpha2.ClusterResourceBinding)
		if !ok {
			return nil
		}
		return util.GetBindingClusterNames(&crb.Spec)
	}

	innerClient := fake.NewClientBuilder().
		WithScheme(scheme).
		WithIndex(&workv1alpha2.ResourceBinding{}, indexregistry.ResourceBindingIndexByFieldCluster, rbIndexerFunc).
		WithIndex(&workv1alpha2.ClusterResourceBinding{}, indexregistry.ClusterResourceBindingIndexByFieldCluster, crbIndexerFunc).
		WithObjects(cluster, rbInfinite, rbFinite, rbNoToleration, crbNoToleration).
		Build()
	fakeClient := &gvkSettingClient{Client: innerClient}

	type enqueueRecord struct {
		name  string
		delay time.Duration
	}
	// recordEnqueues builds a worker that records enqueued keys into the given slice.
	recordEnqueues := func(records *[]enqueueRecord) *recordingWorker {
		return &recordingWorker{
			addAfterFunc: func(item any, delay time.Duration) {
				if k, ok := item.(keys.FederatedKey); ok {
					*records = append(*records, enqueueRecord{name: k.Name, delay: delay})
				}
			},
		}
	}
	var bindingRecords, clusterBindingRecords []enqueueRecord

	tc := &NoExecuteTaintManager{
		Client:                       fakeClient,
		EnableNoExecuteTaintEviction: true,
		bindingEvictionWorker:        recordEnqueues(&bindingRecords),
		clusterBindingEvictionWorker: recordEnqueues(&clusterBindingRecords),
	}

	_, err := tc.syncCluster(context.Background(), cluster)
	if err != nil {
		t.Fatalf("syncCluster() error = %v", err)
	}

	tests := []struct {
		bindingName string
		records     *[]enqueueRecord
		wantAction  string // "skip" (not enqueued), "immediate" (delay=0), "delayed" (delay>0)
	}{
		{bindingName: "infinite-rb", records: &bindingRecords, wantAction: "skip"},
		{bindingName: "finite-rb", records: &bindingRecords, wantAction: "delayed"},
		{bindingName: "no-toleration-rb", records: &bindingRecords, wantAction: "immediate"},
		// ClusterResourceBindings must be routed to the clusterBindingEvictionWorker,
		// not the bindingEvictionWorker.
		{bindingName: "no-toleration-crb", records: &clusterBindingRecords, wantAction: "immediate"},
	}
	for _, tt := range tests {
		t.Run(tt.bindingName, func(t *testing.T) {
			gotAction := "skip"
			for _, r := range *tt.records {
				if r.name != tt.bindingName {
					continue
				}
				if r.delay > 0 {
					gotAction = "delayed"
				} else {
					gotAction = "immediate"
				}
				break
			}
			if gotAction != tt.wantAction {
				t.Errorf("binding %s: enqueue action = %s, want %s", tt.bindingName, gotAction, tt.wantAction)
			}
		})
	}

	// No unexpected extra items should be enqueued.
	if len(bindingRecords) != 2 {
		t.Errorf("expected 2 ResourceBindings enqueued to bindingEvictionWorker, got %d", len(bindingRecords))
	}
	if len(clusterBindingRecords) != 1 {
		t.Errorf("expected 1 ClusterResourceBinding enqueued to clusterBindingEvictionWorker, got %d", len(clusterBindingRecords))
	}
}
```

RainbowMango (commented):
Generally looks good to me. 

I can try to fix the lint issues.

RainbowMango (approved):
/lgtm
/approve

PR #7623, opened 2026-06-12
fix: update cached scale target when scale target reference changes

### LINKED ISSUE #7622: bug: CronFHPAScaleTargetRefUpdates does not update cached scale target after change detection
### Problem

`CronFHPAScaleTargetRefUpdates()` detects when a `CronFederatedHPA` scale target changes, but it does not update the cached target reference stored in `cronFHPAScaleTargetMap`.

### Reproduction

Example sequence:

1. Register target A.
2. Update to target B.
3. Call the function again with target B.

Expected:

target A -> false
target B -> true
target B -> false

Actual:

target A -> false
target B -> true
target B -> true

### Root Cause

The cached target stored in `cronFHPAScaleTargetMap` is only initialized on the first call and is never updated when a new target is detected.

Because of this, the old target remains in the cache and every subsequent comparison continues to detect a difference.

### Expected Behavior

After detecting a target change, the cache should be updated to store the new target reference so future calls compare against the latest state.

### Verification

A unit test reproduces the issue:

1. Initialize the cache with target A.
2. Update to target B.
3. Call the function again with target B.

The test fails on the current implementation because the third call returns `true` instead of `false`.

After updating the cached target reference, the test passes successfully.

- Component: pkg/controllers/cronfederatedhpa
- Function: CronFHPAScaleTargetRefUpdates

rakshityadav1868:
fixes #7622 
What does this PR do?

This PR updates CronFHPAScaleTargetRefUpdates() to refresh the cached scale target reference when a target change is detected.

Without updating the cached value, subsequent calls continue comparing against the stale target reference and repeatedly report the same change.

Changes

* Update the cached scale target reference when a new target is detected.
* Add a regression test covering repeated calls with the same updated target reference.

### DIFF
--- a/pkg/controllers/cronfederatedhpa/cronfederatedhpa_handler.go
+++ b/pkg/controllers/cronfederatedhpa/cronfederatedhpa_handler.go
@@ -74,7 +74,12 @@ func (c *CronHandler) CronFHPAScaleTargetRefUpdates(cronFHPAKey string, scaleTar
 		return false
 	}
 
-	return !equality.Semantic.DeepEqual(origTarget, scaleTarget)
+	if !equality.Semantic.DeepEqual(origTarget, scaleTarget) {
+		c.cronFHPAScaleTargetMap[cronFHPAKey] = scaleTarget
+		return true
+	}
+
+	return false
 }
 
 // AddCronExecutorIfNotExist creates the executor for CronFederatedHPA if not exist
--- a/pkg/controllers/cronfederatedhpa/cronfederatedhpa_handler_test.go
+++ b/pkg/controllers/cronfederatedhpa/cronfederatedhpa_handler_test.go
@@ -98,6 +98,37 @@ func TestCronFHPAScaleTargetRefUpdates(t *testing.T) {
 	}
 }
 
+func TestCronFHPAScaleTargetRefUpdates_BugReproduction(t *testing.T) {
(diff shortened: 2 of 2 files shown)


pkg/controllers/cronfederatedhpa/cronfederatedhpa_handler.go
```diff
@@ -74,7 +74,12 @@ func (c *CronHandler) CronFHPAScaleTargetRefUpdates(cronFHPAKey string, scaleTar
 		return false
 	}
 
-	return !equality.Semantic.DeepEqual(origTarget, scaleTarget)
+	if !equality.Semantic.DeepEqual(origTarget, scaleTarget) {
+		c.cronFHPAScaleTargetMap[cronFHPAKey] = scaleTarget
```
ranxi2001, line 78: Thanks for addressing the repeated target-change detection. One concern: this updates the cache too early, because the cache acts as the marker that the target change has already been handled.

Suppose the target changes from A to B:

1. This line stores B in the cache.
2. `Reconcile` creates the new executor.
3. `updateRuleHistory` fails and `Reconcile` is retried.
4. The retry compares B with cached B and treats the target as unchanged.
5. `processCronRule` sees the existing executor and returns without retrying the failed status update.

The retry then succeeds while `ExecutionHistories` remains empty. I reproduced this by making the first `Status().Update` fail; the second `Reconcile` returned `nil` and made no second status update.

Could we update the cache only after all rules and status updates succeed, and add a regression test for this retry path?

PR #7624, opened 2026-06-13
fix: add systemNamespace SANs to cert hosts for custom namespace

### LINKED ISSUE #7599: Helm chart: deploying to a custom namespace (not karmada-system) breaks APIService discovery
### Problem

  When installing the Karmada Helm chart into a custom namespace (e.g. kkk) instead of the default karmada-system, the karmada-controller-manager fails with:

  no matches for kind "Cluster" in version "cluster.karmada.io/v1alpha1"

  ### Root Cause

  The chart registers APIService objects (e.g. v1alpha1.cluster.karmada.io) inside the karmada-apiserver with a service reference pointing to namespace: {{ .Values.systemNamespace }} (defaults to karmada-system).

  It also creates an ExternalName Service in karmada-system namespace (inside the karmada-apiserver) pointing to:
  karmada-aggregated-apiserver.{{ helm-release-namespace }}.svc.cluster.local

  The problem occurs because:
  1. The karmada-apiserver resolves the APIService backend via the host cluster's DNS
  2. The ExternalName Service in karmada-system namespace needs to exist in the host cluster for DNS resolution to work
  3. But the chart only creates this ExternalName Service inside the karmada-apiserver, not in the host cluster

  When using --namespace kkk, the host cluster has no karmada-system namespace, so the karmada-apiserver cannot route requests to the aggregated-apiserver.

 ###  Steps to Reproduce

  helm --namespace kkk upgrade -i karmada karmada-charts/karmada \
    --version=v1.18.0 \
    --create-namespace

  Then check controller-manager logs:
  kubectl logs deploy/karmada-controller-manager -n kkk

  ### Expected Behavior

  The chart should work correctly when deployed to any namespace, or clearly document that only karmada-system is supported.

  ### Possible Solutions

  1. Document the limitation: state that the chart must be installed in karmada-system namespace (or systemNamespace must match the Helm release namespace).
  2. Fix the templates: ensure the APIService service.namespace and the ExternalName Service are consistent with the actual deployment namespace, so custom namespaces work out of the box.
  3. Add validation: add a check that fails early if the release namespace doesn't match systemNamespace.

  ### Environment

  - Karmada version: v1.18.0
  - Helm chart version: v1.18.0
  - Kubernetes version: v1.35.2
  - Install method: Helm

Priyanshu-u07:
**What type of PR is this?**

/kind bug

This PR adds a `karmada.certHosts` helper in `_helpers.tpl` that conditionally appends systemNamespace SANs to the certificate hosts when `systemNamespace` differs from `.Release.Namespace`.

When deploying the Helm chart to a custom namespace (e.g. `kkk`) instead of` karmada-system`, the auto-generated TLS certificate only includes SANs for the release namespace (`*.kkk.svc`), but the karmada-apiserver verifies the aggregated-apiserver cert against `*.karmada-system.svc` (the `systemNamespace`). This causes a TLS verification failure and breaks APIService discovery.

Fixes #7599

Verified locally by deploying to `--namespace kkk`:

- **Before fix**: `x509: certificate is valid for *.kkk.svc, not karmada-aggregated-apiserver.karmada-system.svc`
- **After fix:** cert includes both `*.kkk.svc` and `*.karmada-system.svc`, no TLS errors, controller-manager starts without "no 
   matches for kind" errors.

```
`helm chart`: Fixed TLS certificate SAN mismatch when deploying to a custom namespace by adding systemNamespace SANs to certs.auto.hosts.
```


charts/karmada/values.yaml
```diff
@@ -144,6 +144,8 @@ certs:
       "*.etcd.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}",
       "*.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}",
       "*.{{ .Release.Namespace }}.svc",
+      "*.{{ .Values.systemNamespace }}.svc.{{ .Values.clusterDomain }}",
+      "*.{{ .Values.systemNamespace }}.svc",
```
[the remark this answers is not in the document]
Priyanshu-u07: systemNamespace is always defined with a default value of "karmada-system" in values.yaml

zhzhuang-zju: Will there be duplicate SAN entries if `Release.Namespace` is the same as `systemNamespace`?

zhzhuang-zju (commented):
Thanks @Priyanshu-u07

charts/karmada/values.yaml
Priyanshu-u07: Yes, there will be duplicate SAN entries in that case, but they're harmless.

zhzhuang-zju: Yes, it should be harmless. Additionally, `values.yaml` does not support deduplication via `if-else` logic.

@seanlaii @jabellard WDYT?

zhzhuang-zju: I have an idea: set the default value of `hosts` to empty in `values.yaml`, and handle its default value within the template. This way we can leverage `if-else` logic to eliminate duplicate definitions. @Priyanshu-u07 WDYT?

Priyanshu-u07: I agree with your idea. My plan would be:
- Add a `karmada.certHosts` helper in `_helpers.tpl` that adds `systemNamespace` SANs when `systemNamespace != .Release.Namespace`
- Use the helper in `pre-install-job.yaml` where `certs.auto.hosts` is used

Does this align with what you had in mind?

zhzhuang-zju: Yes, this aligns with what I envisioned.

Priyanshu-u07:
@zhzhuang-zju, Updated the implementation to use a `karmada.certHosts` helper and avoid duplicate SAN entries. PTAL.

charts/karmada/templates/_helpers.tpl
```diff
@@ -695,3 +695,19 @@ If none are set, outputs nothing.
   volumeMounts:
     {{- include "karmada.kubeconfig.volumeMount" .| nindent 4 }}
 {{- end -}}
+
+{{/*
+Build the certificate hosts list with systemNamespace SANs.
+When deploying to a namespace other than systemNamespace, the certificate
+must also cover *.{systemNamespace}.svc so that karmada-apiserver can verify
+the aggregated-apiserver's TLS cert when accessing APIService backends.
+See: operator/pkg/certs/certs.go L490-506 for the equivalent logic.
```
zhzhuang-zju: It would be better if we add some comments in `values.yaml`

Priyanshu-u07: Done

charts/karmada/values.yaml
```diff
@@ -139,6 +139,8 @@ certs:
     ## @param certs.auto.rootCAExpiryDays expiry of the root CA certificate in days, defaults to 3650 days (10 years)
     rootCAExpiryDays: 3650
     ## @param certs.auto.hosts hosts of the certificate
+    ## Note: When the release namespace differs from systemNamespace, SANs for
+    ## systemNamespace are automatically added by the karmada.certHosts helper.
```
zhzhuang-zju:
```suggestion
## Note: When the release namespace differs from systemNamespace, SANs for
## systemNamespace are automatically added by the karmada.certHosts helper, including:
## - "*.{{ .Values.systemNamespace }}.svc.{{ .Values.clusterDomain }}"
## - "*.{{ .Values.systemNamespace }}.svc"
```

charts/karmada/templates/_helpers.tpl
zhzhuang-zju: Adopt permanent links to avoid content relocation.
```suggestion
See: https://github.com/karmada-io/karmada/blob/f2b734140bd0ba5fa399ffa1d08a3b31ce78c099/operator/pkg/certs/certs.go#L490-L506 for the equivalent logic.
```

zhzhuang-zju (commented):
Thanks, others LGTM

charts/karmada/values.yaml
Priyanshu-u07: Done

charts/karmada/templates/_helpers.tpl
Priyanshu-u07: Done.

zhzhuang-zju:
Thanks
/lgtm
/approve

PR #7077, opened 2026-01-05
fix: ensure status updates are not skipped due to stale cache

zhy76:
Fixes #6858
fix: ensure status updates are not skipped due to stale cache

**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #6858

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
`karmada-controller-manager`: Fixed the issue that AggregatedStatus in rb status may be skipped update due to stale cache
```

PR #7632, opened 2026-06-15
Add warning comments to the UpdateStatus function

### LINKED ISSUE #6858: AggregatedStatus in rb status may be deleted

RainbowMango:
**What type of PR is this?**

/kind cleanup

This PR adds comments to the [UpdateStatus](https://github.com/karmada-io/karmada/blob/5b6c89cd672d59baea1120c266b278dde28d377f/pkg/util/helper/status.go#L39) helper in `pkg/util/helper/status.go`, clarifying its behavior when updating the status sub-resource. 

Fixes #6858
Fixes #7077

The concrete problem originally reported in #6858 — status updates being effectively lost — was traced back to stale-cache conflicts when the execution controller syncs workloads to member clusters. That root cause has been addressed by #7106, which retries on conflict in syncToClusters, so the reported symptom no longer reproduces.

Regarding the broader concern about [helper.UpdateStatus](https://github.com/karmada-io/karmada/blob/5b6c89cd672d59baea1120c266b278dde28d377f/pkg/util/helper/status.go#L39) itself: it fetches the object from the (potentially stale) informer cache and skips the write when the mutated status DeepEquals the cached copy. In theory this can skip a needed update. After re-reviewing all current callers, we could not find a reliable, reproducible path where this leads to a permanently lost status update in the codebase as it stands today.

Given that, rather than reworking the helper's semantics (which risks regressions such as dropping optimistic concurrency), we've chosen to document the precondition.

```release-note
NONE
```

### DIFF
--- a/pkg/util/helper/status.go
+++ b/pkg/util/helper/status.go
@@ -36,6 +36,19 @@ import (
 // Note: changes to any sub-resource other than status will be ignored.
 // Changes to the status sub-resource will only be applied if the object
 // already exist.
+//
+// WARNING: This helper fetches the object via the client's cache (an informer),
+// which may be stale. The desired status produced by MutateFn is compared
+// (DeepEqual) against this possibly-stale cached object, and the update is
+// skipped when they are equal. If the cache lags behind the API server, a
+// needed update can be silently skipped.
+//
+// Because of the above, callers MUST ensure the problem can eventually
+// converge even if a single update is skipped. In particular, a caller MUST
+// NOT filter out (ignore) reconcile events without also having a periodic
+// reconcile, otherwise a skipped update may never be retried and the status
+// stays stale forever.
+// See https://github.com/karmada-io/karmada/issues/6858 for the background.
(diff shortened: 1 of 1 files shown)


whitewindmills:
thanks
/lgtm
/approve

PR #7633, opened 2026-06-15
Fix metricsadapter: sort GetMetricBySelector results for deterministic output

ManojLamani:
**What type of PR is this?**

/kind cleanup

`GetMetricBySelector` aggregates custom metrics from member clusters into a
map (`sameMetrics`) and then iterates the map to build the response Items slice.
Go map iteration order is non-deterministic, so every call returns metrics in a
different order. This can cause inconsistent behavior in HPA controllers that
consume the custom metrics API and compare metric lists across calls.

This PR sorts the Items slice by `DescribedObject.Name` before returning,
producing a stable and predictable response. Resolves the existing TODO left
by @chaunceyjiang.

N/A

- Addresses the TODO comment in `GetMetricBySelector`:
  `// TODO(chaunceyjiang) The MetricValue items need to be sorted.`
- The sort key is `DescribedObject.Name`, which uniquely identifies a metric
  target within a namespace.
- No existing test file for this package; the change is a single `sort.Slice`
  call with no logic branching.

```release-note
`karmada-metrics-adapter`: Fixed the issue that `GetMetricBySelector` returned
custom metrics in non-deterministic order when aggregating results from multiple
member clusters.

### DIFF
--- a/pkg/metricsadapter/provider/custommetrics.go
+++ b/pkg/metricsadapter/provider/custommetrics.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"sort"
 	"strings"
 	"sync"
 
@@ -142,7 +143,6 @@ func (c *CustomMetricsProvider) GetMetricBySelector(ctx context.Context, namespa
 		if !ok {
 			break
 		}
-		// TODO(chaunceyjiang) The MetricValue items need to be sorted.
 		for _, metric := range metrics.Items {
 			// metrics is unique in one cluster, but it may exist in multiple clusters.
 			// for this situation, we need to add the value of all clusters.
@@ -155,6 +155,9 @@ func (c *CustomMetricsProvider) GetMetricBySelector(ctx context.Context, namespa
 	for _, metric := range sameMetrics {
 		metricValueList.Items = append(metricValueList.Items, metric)
 	}
+	sort.Slice(metricValueList.Items, func(i, j int) bool {
+		return metricValueList.Items[i].DescribedObject.Name < metricValueList.Items[j].DescribedObject.Name
+	})
 	if len(metricValueList.Items) == 0 {
 		return nil, provider.NewMetricNotFoundError(info.GroupResource, info.Metric)
 	}

### CODE REVIEW

**Comment by zhzhuang-zju** [comment]
thanks 
/assign

File: pkg/clusterdiscovery/clusterapi/clusterapi.go [hunk 1]
```diff
@@ -207,7 +207,8 @@ func (d *ClusterDetector) joinClusterAPICluster(clusterWideKey keys.ClusterWideK
 
 	clusterRestConfig, err := apiclient.RestConfig("", kubeconfigPath)
 	if err != nil {
-		klog.Fatalf("Failed to get cluster-api management cluster rest config. kubeconfig: %s, err: %v", kubeconfigPath, err)
+		klog.Errorf("Failed to get cluster-api management cluster rest config. kubeconfig: %s, err: %v", kubeconfigPath, err)
+		return err
```
zhzhuang-zju: This change seems unrelated to the main scope of this PR. I suggest submitting it as a separate PR.

zhzhuang-zju (commented):
Thanks @ManojLamani. I’m unsure whether this PR qualifies as a bugfix. Could you provide specific scenarios where functionality fails without this change?

Also, ask @jabellard to take a look; he is an expert on metrics.

ManojLamani:
Thank you for the clarification @zhzhuang-zju. 

You are right, this does not cause a hard functional failure. The core issue is that Go map iteration is non-deterministic, so GetMetricBySelector returns metrics in a different order on every call. While most HPA controllers iterate all items rather than relying on specific positions, an unstable API response makes the system harder to test and debug reliably. This is why the original author @chaunceyjiang left the TODO explicitly asking for sorting. And I've updated the kind bug to /kind cleanup to better reflect the nature of this change.

Thank you for your guidance.

Sure, @jabellard would appreciate your expertise on this. Please take a look when you get a chance!

pkg/clusterdiscovery/clusterapi/clusterapi.go
ManojLamani: Noted, I will remove this change from the current PR and submit it as a separate PR. Thank you!

zhzhuang-zju:
> And I've updated the kind bug to /kind cleanup to better reflect the nature of this change.

It doesn't seem to take effect. Let me help with this.
/kind-remove bug
/kind cleanup

ManojLamani:
Thank you @zhzhuang-zju for fixing the labels!

PR #7639, opened 2026-06-18
feat(helm): add extraCommandArgs support to remaining components

### LINKED ISSUE #7626: Add extraCommandArgs support to all Helm chart components

Add `extraCommandArgs` to all components in the Helm chart. Currently only `controllerManager` supports this. The following components are missing it:

- karmada-scheduler
- karmada-descheduler
- karmada-scheduler-estimator
- karmada-webhook
- karmada-agent
- karmada-aggregated-apiserver
- karmada-apiserver
- karmada-metrics-adapter
- karmada-search

Each component would get an `extraCommandArgs: {}` map in values.yaml that renders as `--key=value` args, following the existing pattern in `controllerManager`.

**Why is this needed**:

There's no way to pass arbitrary flags to most components without post-deploy patching. Every new flag request requires a chart change (e.g. #7572 / #7570 for a single scheduler flag). `extraCommandArgs` gives users a generic escape hatch so they don't need to wait for dedicated values to be added for each flag.

rakshityadav1868:
## What type of PR is this?

/kind feature

This PR adds `extraCommandArgs` support to all remaining Helm chart components:

- `karmada-scheduler`
- `karmada-descheduler`
- `karmada-scheduler-estimator`
- `karmada-webhook`
- `karmada-agent`
- `karmada-aggregated-apiserver`
- `karmada-apiserver`
- `karmada-metrics-adapter`
- `karmada-search`

Each component now exposes:

```yaml
extraCommandArgs: {}
```

Arguments are rendered as:

```yaml
- --<key>=<value>
```

and are appended after built-in and dedicated flags. This matches the existing `controllerManager` behavior and intentionally allows `extraCommandArgs` to act as the final override layer for single-value flags.

Documentation has also been added to `values.yaml` describing the precedence behavior and noting that some multi-value flags may not behave as simple overrides.

Fixes #7626

## Special notes for your reviewer

Validation performed:

```bash
helm lint charts/karmada
helm template charts/karmada >/dev/null
```

Additionally, rendering was verified with sample `extraCommandArgs` values (including `schedulerEstimator`) to ensure `--key=value` arguments are correctly appended to manifests.

### DIFF
--- a/charts/karmada/templates/_helpers.tpl
+++ b/charts/karmada/templates/_helpers.tpl
@@ -644,6 +644,79 @@ If none are set, outputs nothing.
 {{- end }}
 {{- end -}}
 
+{{- define "karmada.scheduler.extraCommandArgs" -}}
+{{- if .Values.scheduler.extraCommandArgs }}
+{{- range $key, $value := .Values.scheduler.extraCommandArgs }}
+- --{{ $key }}={{ $value }}
+{{- end }}
+{{- end }}
+{{- end -}}
+
+{{- define "karmada.descheduler.extraCommandArgs" -}}
+{{- if .Values.descheduler.extraCommandArgs }}
+{{- range $key, $value := .Values.descheduler.extraCommandArgs }}
+- --{{ $key }}={{ $value }}
+{{- end }}
+{{- end }}
+{{- end -}}
+
+{{- define "karmada.schedulerEstimator.extraCommandArgs" -}}
+{{- if .Values.schedulerEstimator.extraCommandArgs }}
+{{- range $key, $value := .Values.schedulerEstimator.extraCommandArgs }}
+- --{{ $key }}={{ $value }}
+{{- end }}
+{{- end }}
+{{- end -}}
+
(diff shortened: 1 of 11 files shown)


rakshityadav1868:
Hi @jabellard PTAL,

Addressing the bot suggestions:

For this PR, I intentionally kept the implementation consistent with the existing `controllerManager.extraCommandArgs` behavior and only extended the same pattern to the remaining components.

Since the goal of #7626 is feature parity across all components, I avoided introducing additional behavior changes such as bare flag support, deterministic ordering, or different rendering semantics, as those would also require revisiting the existing `controllerManager` implementation for consistency.

I'm happy to follow up with additional changes, but I wanted to keep this PR focused on the scope of #7626 and preserve the current chart behavior.

jabellard:
@rakshityadav1868 

Wow!!! I'm very happy to see this PR from you. This is a great feature that will benefit the community as a whole. I'll take a closer look soon.

rakshityadav1868:
hey @jabellard can u take a look !!

PR #7648, opened 2026-06-21
fix: close outFile before returning on copy error in Unpack and DeCompress

Ahmad-Faraj:
## What type of PR is this?

/kind bug

`Unpack` in `operator/pkg/util` and `DeCompress` in `pkg/karmadactl/cmdinit/utils` both open a file per tar entry inside a loop. When `ioCopyN` returns an error, the function returns without closing the open `*os.File`, leaking the file descriptor.

The successful `outFile.Close()` was also discarding its return value. On NFS and similar filesystems, write errors only surface at close time, so a failed write could return success with a truncated file.

None (self-found)

## Special notes for your reviewer

Both functions have the same bug; `Unpack` is in `operator/` and `DeCompress` in `karmadactl/cmdinit/utils/`.

No.

### DIFF
--- a/operator/pkg/util/util.go
+++ b/operator/pkg/util/util.go
@@ -153,9 +153,12 @@ func Unpack(file, targetPath string) error {
 				return err
 			}
 			if err := ioCopyN(outFile, tr); err != nil {
+				outFile.Close()
 				return err
 			}
-			outFile.Close()
+			if err := outFile.Close(); err != nil {
+				return err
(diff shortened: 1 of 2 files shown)


Ahmad-Faraj:
Friendly ping @zhzhuang-zju. Small fix for a file handle leak in Unpack and DeCompress, has been sitting for two weeks. Happy to adjust if anything looks off.

